PR #6258, opened 2018-11-14
[api] support webhooks from gitlab

adrianschroeter:
@ChrisBr just minimalistic enhancement to allow people using our internal gitlab. Not sure if makes sense here or in your github code ...


adrianschroeter:
it works, but test suite did not run yet :)

src/api/app/controllers/trigger_controller.rb
```diff
@@ -12,6 +12,11 @@ class TriggerController < ApplicationController
 
   def runservice
     auth = request.env['HTTP_AUTHORIZATION']
+
+    if request.env['HTTP_X_GITLAB_EVENT'] = 'Push Hook'
```
dmarcoux: I think you forgot an equal sign here in the `if`

adrianschroeter: yep, fixed

PR #6259, opened 2018-11-14
Fix rubocop violations

ChrisBr:
No idea how this went into master though ...

cc @bgeuken & @coolo maybe with the merging of master into staging-workflow?


ChrisBr:
> No idea how this went into master though ...

Ah, I think we recently changed the rubcop config which found is way back into staging-workflow now?

PR #6261, opened 2018-11-14
Add staging coverage

coolo:
Giving @bgeuken work to do :)


src/api/spec/models/staging/staging_project_spec.rb
```diff
@@ -69,24 +71,55 @@
...
+      before do
+        repository_arch.update(required_checks: ['check_1'])
+      end
+
+      it { expect(staging_project.overall_state).to eq(:testing) }
+    end
+
     context 'when there are pending checks' do
-      let!(:check) { create(:check, name: 'check_1', status_report: status_report, state: 'pending') }
+      let!(:check) { create(:check, name: 'check_1', status_report: published_report, state: 'pending') }
```
dmarcoux: [nitpick]

All `let!(:check) { ... }` are used directly, so they should all be in a `before` block. Except for the last one, which could be used on line 118 for `required_check: [check.name]`

coolo: why would you put the let! in a before block? The before is implicit

dmarcoux: I wasn't clear enough in what I meant. I meant that all the `let!(:check)` (except the last one) aren't needed since the variable `check` is never called. Instead, it should be like this:

```ruby
before do
  create(:check, name: 'check_1', status_report: published_report, state: 'pending')
end

it { expect(staging_project.overall_state).to eq(:testing) }
```

Using `let(:something)` is only when you need to call/use that `something` variable. Otherwise, it should be in a `before` block.

hennevogel: > Otherwise, it should be in a before block.

Says who? :smile:

dmarcoux: Just think about it, why having a variable if you don't use it? I would say the common sense says it.

hennevogel: before and let! are just two ways of doing

> force the method's invocation before each example.

See https://relishapp.com/rspec/rspec-core/docs/helper-methods/let-and-let

dmarcoux: But I guess it boils down to personal preference... this is why I put the [nitpick].

coolo: https://dzone.com/articles/rspec-let-vs-before:

**In nearly every situation, it is better to use let over before blocks.**

dmarcoux: It's nice to omit the part which is exactly my point

src/api/app/models/staging/staging_project.rb
```diff
@@ -105,11 +103,11 @@ def unassign_managers_group(managers)
     end
 
     def current_checks
-      relevant_checks + relevant_checks_for_architectures
```
coolo: now that checks and status_reports are gone, can we remove all these current_ and relevant_ prefixes?

coolo:
"Coverage increased (+0.03%) to 91.009%" - while master is at 91.225. There are specs left to write :)

src/api/app/models/staging/staging_project.rb
```diff
@@ -1,10 +1,8 @@
 module Staging
   class StagingProject < Project
     has_many :staged_requests, class_name: 'BsRequest', foreign_key: :staging_project_id, dependent: :nullify
-    has_many :status_reports, through: :repositories, inverse_of: :checkable
-    has_many :checks, through: :status_reports
+    has_many :status_reports_for_repositories, source: :status_reports, through: :repositories, inverse_of: :checkable
```
hennevogel: :thought_balloon: I don't really like the naming. I know `repo` vs. `arch` is what sets them apart technically but it's not what sets them apart in the workflow.

src/api/spec/models/staging/staging_project_spec.rb
```diff
@@ -11,9 +11,14 @@
   let(:staging_project) { staging_workflow.staging_projects.first }
 
   let!(:repository) { create(:repository, architectures: ['x86_64'], project: staging_project, name: 'staging_repository') }
+  let!(:repository_arch) { repository.repository_architectures.first }
+  let!(:architecture) { repository_arch.architecture }
   let!(:package) { create(:package_with_file, project: staging_project) }
 
-  let!(:status_report) { create(:status_report, checkable: repository) }
+  let!(:published_report) { create(:status_report, checkable: repository) }
+  let!(:build_report) { create(:status_report, checkable: repository_arch) }
```
hennevogel: :thought_balloon: those are way better names :smile:

ChrisBr: That makes sense

src/api/app/models/staging/staging_project.rb
coolo: These relationships shouldn't be public anyway - see https://github.com/coolo/open-build-service/commit/7e62af53e9e438ea32aeeec3b73c6cb616355e32

But I would prefer to do this as a followup

Issue #6263, opened 2018-11-14
Ruby Sass is deprecated and will be unmaintained as of 26 March 2019


dmarcoux:
I stumbled onto this while running *bundler* locally. We should take action soon.

```
Ruby Sass is deprecated and will be unmaintained as of 26 March 2019.

* If you use Sass as a command-line tool, we recommend using Dart Sass, the new
  primary implementation: https://sass-lang.com/install

* If you use Sass as a plug-in for a Ruby web framework, we recommend using the
  sassc gem: https://github.com/sass/sassc-ruby#readme

* For more details, please refer to the Sass blog:
  http://sass.logdown.com/posts/7081811
```

coolo:
We need https://github.com/twbs/bootstrap-rubygem/commit/7c6056eebb7a9fd8cac7845ca31bf62a81627df2 to be released

the rest is https://github.com/openSUSE/open-build-service/pull/6264

coolo:
Fixed in 99259053cc2dba78cc0ef110b7b6c0bdd07b5521

PR #6264, opened 2018-11-14
Replace sass-rails with sassc-rails

coolo:
ruby sass is dead, partly fixing #6263

### DIFF
--- a/dist/obs-bundled-gems.spec
+++ b/dist/obs-bundled-gems.spec
@@ -111,6 +111,11 @@ test -f %{buildroot}%_libdir/obs-api/ruby/2.5.0/gems/rack-%{rack_version}/rack.g
 # run gem clean up script
 /usr/lib/rpm/gem_build_cleanup.sh %{buildroot}%_libdir/obs-api/ruby/*/
 
+# work around sassc bug - and install libsass
+sassc_dir=$(ls -1d %{buildroot}%_libdir/obs-api/ruby/2.5.0/gems/sassc-2*)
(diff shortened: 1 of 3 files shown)


dmarcoux:
Just to be extra safe, I'll put the review app label. Just in case

coolo:
@ChrisBr good thing we implemented the OBS Package Build check, no? :)

ChrisBr:
> good thing we implemented the OBS Package Build check, no? :)

it was not the first time good :+1:

ChrisBr:
Means we need to require libsass?

ChrisBr:
But still wondering that the review app works and does not crash :thinking:

coolo:
the review is pretty much independent of the test suite IMO, so dismissing it isn't necessary

coolo:
I filed an issue upstream and added a workaround

coolo:
looks like it's working

ChrisBr:
> looks like it's working

Thanks for taking care. But could you please add some more details 7988e41e7d9131ea63de38414f0053aecd87249b and maybe even link to the issue. Thanks :heart:

coolo:
you mean URL within the spec file?

ChrisBr:
> you mean URL within the spec file?

No, in your commit message :smile:

coolo:
I know - but I dislike how github spams the other issue then, so I wanted to avoid that. :)

ChrisBr:
> I know - but I dislike how github spams the other issue then, so I wanted to avoid that. :)

But then at least add a description and write the repository to look for in the commit message :thinking:

ChrisBr:
LGTM, thanks @coolo

PR #6266, opened 2018-11-14
Add default_scope to StagingProject

ChrisBr:
because otherwise using ActiveRecord query methods will consider
all Projects and not only StagingProjects.
This causes some weird issues like returning Projects without
a Staging::Workflow associated.

### DIFF
--- a/src/api/app/models/staging/staging_project.rb
+++ b/src/api/app/models/staging/staging_project.rb
@@ -4,6 +4,8 @@ module Staging
     has_many :status_reports, through: :repositories, inverse_of: :checkable
     belongs_to :staging_workflow, class_name: 'Staging::Workflow'
 
+    default_scope { where.not(staging_workflow: nil) }
+
     def staging_identifier
       name[/.*:Staging:(.*)/, 1]
     end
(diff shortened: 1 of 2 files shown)


ChrisBr:
> Why do we want to keep staging projects without staging workflow? I mean, why are they even in the database?

StagingProject inherits from Project

ChrisBr:
@dmarcoux @DavidKang please have another look, needed to update the ``after_create`` as well, not sure about that though...

ChrisBr:
> what I why don't we have a validation that avoids that is happens. I think it is a better approach than the scope.

I don't get how this would solve the problem at hand? This is about the ActiveRecord query methods. If you do ``Staging::StagingProject.first`` ActiveRecord will create

```
SELECT  `projects`.* FROM `projects` WHERE (projects.id not in (0)) ORDER BY `projects`.`id` ASC LIMIT 1
```

This is because we use SingleTableInheritance. We could also introduce a new column ``type`` and store the class name in it (https://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html). However, this would raise other problems like a StagingProjet is basically just a normal project attached to a staging workflow. So if we store a type column and we delete the StagingWorkflow we also need to change the type of the project because in this very moment it changes to be a normal Project.

But I think adding the ``default_scope`` here is enough to distinct.

PR #6268, opened 2018-11-14
Request approver fixes

marcus-h:
- Fix permission check in BsRequest.approval_handling and testcase
- Fix approver specific code in BsRequestPermissionCheck

(Maybe we should also document the "cmd=cancelapproval" semantics in
a testcase...)


adrianschroeter:
Thanks!

PR #6271, opened 2018-11-15
Fix unknown view 'status' error

ChrisBr:
as Backend is not deployed yet we need to comment this code for now.

Kudos to @coolo for quickly fixing.

We can revert this patch when Backend deployed build_id for unpublished repositories.


coolo:
according to @adrianschroeter the backend is deployed right now

vpereira:
> according to @adrianschroeter the backend is deployed right now

there were some issues with restarting the code with systemd and the old code was till running. now its fixed.

ChrisBr:
Deployed and removed monkey patch ...

PR #6180, opened 2018-11-05
Move BsRequest#with_reviews to the review model

bgeuken:
This method is querying reviews. Thus it belongs to the review model.
Since it's also just a query method I converted it into a scope.

PR #6273, opened 2018-11-15
Fix N+1 queries for reviews in staging project model

dmarcoux:
Also use the `|` to join the arrays and remove the duplicates

### DIFF
--- a/src/api/app/models/staging/staging_project.rb
+++ b/src/api/app/models/staging/staging_project.rb
@@ -18,7 +18,7 @@ module Staging
     # The requests to review are requests which are not related to the staging project (unless they are also staged).
     # They simply need a review from the maintainers of the staging project.
     def requests_to_review
-      @requests_to_review ||= BsRequest.with_open_reviews_for(by_project: name)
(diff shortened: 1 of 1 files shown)


src/api/app/models/staging/staging_project.rb
```diff
@@ -35,7 +35,7 @@ def missing_reviews
       @missing_reviews = []
       attribs = [:by_group, :by_user, :by_project, :by_package]
 
-      (requests_to_review + staged_requests).uniq.each do |request|
+      (requests_to_review | staged_requests.includes(:reviews)).each do |request|
```
Ana06: :cupid:

coolo: @Ana06 you're just too easy :smile:

bgeuken: @dmarcoux I didn't realize that earlier... This only works when combined with fa178200e32a7cfb2897031d19bfb23ba21b176f

bgeuken:
This is now part of #6180

PR #6257, opened 2018-11-14
Fix StagingWorkflow#backlog

ChrisBr:
The assocation was wrong as it should only consider
* Requests in state :review
* with reviews in state :new and by_group of the staging managers
* with the target project of the distribution

Furthermore it changes the ``unassigned_reqeuests`` method to return an ActiveRecordCollection to be more efficient and make it easier to work with.

Additionally added a link to the GroupShow page which is currently used as Backlog.

PR #6206, opened 2018-11-08
Store StagingWorkflow in the backend

DavidKang:
StagingWorkflow is stored in the backend under `_project` directory with
the name `_staging_workflow`. In this special file, we store a minimal
info of the staging_workflow.

PR #6275, opened 2018-11-15
Staging workflow managers group

mdeniz:
Now a Managers group needs to be attached to the staging workflow in order to let them managing the assignation/exclusion of requests in the staging projects.

Now the Managers group assigned is shown in **Infos** sidebox:

image

When configuring the Staging you can assign a new group, only one will be assigned:

image

The field for selecting another group is an autocomplete one:

image

BTW, I have adapted some of the literals the user see. Now the **Staging Workflow** is referred as **Staging.** The edit action is to configure the Staging.
 
### DIFF
--- a/src/api/app/assets/javascripts/webui2/staging_workflow.js
+++ b/src/api/app/assets/javascripts/webui2/staging_workflow.js
@@ -7,3 +7,21 @@ function setSpinnersForDeletion() { // jshint ignore:line
     $(this).find('.delete-spinner').addClass('d-none');
   });
 }
+
+function autocompleteStagingManagersGroup() { // jshint ignore:line
+  $('#managers_title').autocomplete({
+    appendTo: '#assign-managers-group-modal-input',
+    source: $('#managers_title').data('autocompleteGroupsUrl'),
+    search: function() {
+      var icon = $('#assign-managers-group-search-icon i:first-child');
+      icon.addClass('d-none');
+      icon.next().removeClass('d-none');
+    },
+    response: function() {
+      var icon = $('#assign-managers-group-search-icon i:first-child');
+      icon.removeClass('d-none');
+      icon.next().addClass('d-none');
+    },
+    minLength: 2
+  });
+}
--- a/src/api/app/controllers/webui/staging/projects_controller.rb
+++ b/src/api/app/controllers/webui/staging/projects_controller.rb
@@ -12,6 +12,7 @@ module Webui
 
         staging_project_name = "#{@staging_workflow.project}:Staging:#{params[:staging_project_name]}"
         staging_project = @staging_workflow.staging_projects.build(name: staging_project_name)
+        staging_project.assign_managers_group(@staging_workflow.managers_group)
 
         if staging_project.valid? && staging_project.store
           flash[:success] = "Staging project with name = \"#{staging_project}\" was successfully created"
--- a/src/api/app/controllers/webui/staging/projects_controller.rb
+++ b/src/api/app/controllers/webui/staging/projects_controller.rb
@@ -50,7 +50,7 @@ module Webui
         return if @staging_workflow
 
         redirect_back(fallback_location: root_path)
-        flash[:error] = "StagingWorkflow with id = #{params[:staging_workflow_id]} doesn't exist"
+        flash[:error] = "Staging with id = #{params[:staging_workflow_id]} doesn't exist"
         return
       end
     end
--- a/src/api/app/controllers/webui/staging/projects_controller.rb
+++ b/src/api/app/controllers/webui/staging/projects_controller.rb
@@ -12,7 +12,6 @@ module Webui
 
         staging_project_name = "#{@staging_workflow.project}:Staging:#{params[:staging_project_name]}"
         staging_project = @staging_workflow.staging_projects.build(name: staging_project_name)
-        staging_project.assign_managers_group(@staging_workflow.managers_group)
 
         if staging_project.valid? && staging_project.store
           flash[:success] = "Staging project with name = \"#{staging_project}\" was successfully created"
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -20,6 +20,8 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
     staging_workflow = @project.build_staging
     authorize staging_workflow
 
+    staging_workflow.managers_group = Group.find_by(title: params[:managers_title])
+
     if staging_workflow.save
       flash[:success] = "Staging Workflow for #{@project.name} was successfully created"
       redirect_to staging_workflow_path(staging_workflow)
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -72,23 +72,11 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
   def update
     authorize @staging_workflow
 
-    begin
-      Staging::Workflow.transaction do
-        old_managers_group = @staging_workflow.managers_group
-        new_managers_group = Group.find_by(title: params[:managers_title])
-
-        @staging_workflow.staging_projects.each do |staging_project|
-          staging_project.unassign_managers_group(old_managers_group)
-          staging_project.assign_managers_group(new_managers_group)
-          staging_project.store
-        end
-
-        @staging_workflow.managers_group = new_managers_group
-        @staging_workflow.save!
-      end
-      flash[:success] = 'Managers group if staging was successfully assigned'
-    rescue
-      flash[:error] = "Sorry, something unexpected happended and the group couldn't be assigned to this staging"
+    @staging_workflow.managers_group = Group.find_by(title: params[:managers_title])
+    if @staging_workflow.save
+      flash[:success] = 'Managers group was successfully assigned'
+    else
+      flash[:error] = "Managers group couldn't be assigned: #{@staging_workflow.errors.full_messages.to_sentence}."
     end
 
     redirect_to edit_staging_workflow_path(@staging_workflow)
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -72,13 +72,25 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
   def update
     authorize @staging_workflow
 
-    @staging_workflow.managers = Group.find_by(title: params[:managers_id])
+    begin
+      Staging::Workflow.transaction do
+        old_managers_group = @staging_workflow.managers_group
+        new_managers_group = Group.find_by(title: params[:managers_title])
 
-    if @staging_workflow.save
-      flash[:success] = 'Managers group for Staging Workflow was successfully assigned'
-    else
-      flash[:error] = "Sorry the group couldn't be assigned to this Staging Workflow"
+        @staging_workflow.staging_projects.each do |staging_project|
+          staging_project.unassign_managers_group(old_managers_group)
+          staging_project.assign_managers_group(new_managers_group)
+          staging_project.store
+        end
+
+        @staging_workflow.managers_group = new_managers_group
+        @staging_workflow.save!
+      end
+      flash[:success] = 'Managers group if staging was successfully assigned'
+    rescue
+      flash[:error] = "Sorry, something unexpected happended and the group couldn't be assigned to this staging"
     end
+
     redirect_to edit_staging_workflow_path(@staging_workflow)
   end
 
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -23,10 +23,10 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
     staging_workflow.managers_group = Group.find_by(title: params[:managers_title])
 
     if staging_workflow.save
-      flash[:success] = "Staging Workflow for #{@project.name} was successfully created"
+      flash[:success] = "Staging for #{@project} was successfully created"
       redirect_to staging_workflow_path(staging_workflow)
     else
-      flash[:error] = "Staging Workflow for #{@project.name} couldn't be created"
+      flash[:error] = "Staging for #{@project} couldn't be created"
       render :new
     end
   end
@@ -61,10 +61,10 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
     @staging_workflow.staging_projects.where(id: params[:staging_project_ids]).destroy_all
 
     if @staging_workflow.destroy
-      flash[:success] = "Staging Workflow for #{@project.name} was successfully deleted."
+      flash[:success] = "Staging for #{@project} was successfully deleted."
       render js: "window.location='#{project_show_path(@project)}'"
     else
-      flash[:error] = "Staging Workflow for #{@project.name} couldn't be deleted: #{@staging_workflow.errors.full_messages.to_sentence}."
+      flash[:error] = "Staging for #{@project} couldn't be deleted: #{@staging_workflow.errors.full_messages.to_sentence}."
       render js: "window.location='#{staging_workflow_path(@staging_workflow)}'"
     end
   end
@@ -105,7 +105,7 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
     return if @staging_workflow
 
     redirect_back(fallback_location: root_path)
-    flash[:error] = "StagingWorkflow with id = #{params[:id]} doesn't exist"
+    flash[:error] = "Staging with id = #{params[:id]} doesn't exist"
     return
   end
 end
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -66,6 +66,19 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
     end
   end
 
+  def update
+    authorize @staging_workflow
+
+    @staging_workflow.managers = Group.find_by(title: params[:managers_id])
+
+    if @staging_workflow.save
+      flash[:success] = 'Managers group for Staging Workflow was successfully assigned'
+    else
+      flash[:error] = "Sorry the group couldn't be assigned to this Staging Workflow"
+    end
+    redirect_to edit_staging_workflow_path(@staging_workflow)
+  end
+
   private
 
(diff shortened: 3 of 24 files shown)


coolo:
hmm, 'stage' according to the [dictionary](https://www.merriam-webster.com/dictionary/stage) is a synonym for 'step'. But A B C aren't steps - the whole staging process is the step. 

If I read 'Add a stage' it sounds like it modifies the process, but it's not. It's just adding another door to the same stage.

coolo:
What was wrong with 'Project' as it reads in the existant dashboard anyway?

mdeniz:
I agree, "stage" is misleading. I will let it as Staging Project, no worries. Just wanted to try other name for that... :+1:

src/api/spec/controllers/webui/staging/workflows_controller_spec.rb
```diff
@@ -2,7 +2,10 @@
 
 RSpec.describe Webui::Staging::WorkflowsController do
   let(:user) { create(:confirmed_user, login: 'tom') }
+  let(:managers_group) {create(:group)}
+  let(:other_managers_group) {create(:group)}
   let(:project) { user.home_project }
+  let(:staging_workflow) { project.create_staging(managers: managers_group) }
```
ChrisBr: I think we should use a factor here instead and adapt the Factory as well. We do it like this everywhere.

See https://github.com/openSUSE/open-build-service/pull/6257/commits/d867d2953e4f25038f9ff9d08670074c2d609c8a

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -38,6 +40,7 @@ def show
     @ignored_requests = @staging_workflow.ignored_requests.first(5)
     @more_ignored_requests = @staging_workflow.ignored_requests.count - @ignored_requests.size
     @empty_projects = @staging_workflow.staging_projects.without_staged_requests
+    @managers = @staging_workflow.managers
```
ChrisBr: I don't like the naming managers. With this naming I would expect I get a list of Users but what I get is a single group. Doesn't make sense. So either renaming it to managers_group or really return the managers.

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -20,6 +20,8 @@ def create
     staging_workflow = @project.build_staging
     authorize staging_workflow
 
+    staging_workflow.managers = Group.find_by(title: params[:managers_id])
```
ChrisBr: [nitpick] Why ``managers_id`` and not ``title`` (or ``managers_title`` or ``group_title``

coolo:
Something else: this group needs to be written to the backend file as well (now that #6206 is in)

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -56,14 +59,38 @@ def destroy
...
+      flash[:success] = "Staging for #{@project.name} was successfully deleted."
       render js: "window.location='#{project_show_path(@project)}'"
     else
-      flash[:error] = "Staging Workflow for #{@project.name} couldn't be deleted: #{@staging_workflow.errors.full_messages.to_sentence}."
+      flash[:error] = "Staging for #{@project.name} couldn't be deleted: #{@staging_workflow.errors.full_messages.to_sentence}."
       render js: "window.location='#{staging_workflow_path(@staging_workflow)}'"
     end
   end
 
+  def assign_managers_group
```
ChrisBr: What about putting this in a dedicated controller as update action how we do already for most other controllers?

```
Staging::Project::GroupsController#update
```

src/api/app/models/group.rb
```diff
@@ -5,6 +5,7 @@
 # Groups have an arbitrary number of roles and users assigned to them.
 #
 class Group < ApplicationRecord
+  has_one :staging_workflow, foreign_key: :managers_id, dependent: :nullify
```
ChrisBr: As @hennevogel wrote in https://github.com/openSUSE/open-build-service/pull/6257/commits/d867d2953e4f25038f9ff9d08670074c2d609c8a, this is not needed.

src/api/app/models/staging/staging_project.rb
```diff
@@ -59,6 +59,17 @@ def overall_state
       @overall_state ||= state
     end
 
+    def assign_managers_group(managers)
+      role = Role.find_by_title!('maintainer')
+      return if relationships.find_by(group: managers, role: role)
```
ChrisBr: This means the group is already assigned? Then return ``true``, no? Or what does ``Relationship.add_group`` return?

src/api/app/controllers/webui/staging/workflows_controller.rb
ChrisBr: Same is true for the assocation

mdeniz:
> Something else: this group needs to be written to the backend file as well (now that #6206 is in)

I will do it in another PR... this is big enough :+1:

src/api/app/controllers/webui/staging/workflows_controller.rb
hennevogel: **Group**. *Noun*. Any collection or assemblage of persons or things; cluster; aggregation.

So manager**s** suggests this is a group :-) And please keep in mind that in the future this should be multiple groups and users.

src/api/app/models/group.rb
mdeniz: But if the admin removes the group? what happen with the staging workflow? it will have an id for a group that doesn't exist

src/api/app/controllers/webui/staging/workflows_controller.rb
mdeniz: Yes, sure I will change it

src/api/app/models/group.rb
hennevogel: True, `:nullify` makes sense. Did not think about this.

src/api/app/controllers/webui/staging/workflows_controller.rb
ChrisBr: @hennevogel  I don't get it ... It is a ``has_one`` assocation so plural does not make sense here, using plural for a belongs_to / has_one assocation violates convention.

src/api/app/models/staging/staging_project.rb
ChrisBr: As ``Relationship.add_group(self, managers, role, nil, true)`` also returns ``nil`` I'm fine.

src/api/app/controllers/webui/staging/workflows_controller.rb
hennevogel: This is a name. A name that should generate meaning if you read this code. This name is not there so you can guess what objects it returns. Who says that `staging_workflow.managers` does not return a comma separated list of names or any other data structure that you can cook up?

src/api/app/controllers/webui/staging/workflows_controller.rb
mdeniz: But this is updating the StagingWorkflow, not the Group.

src/api/app/controllers/webui/staging/workflows_controller.rb
ChrisBr: What? No, there are conventions you should follow for naming. And yes it should be possible from the name to know if you get one object or several (and even to know what type it is e.g. ``project`` -> object vs ``project_name`` -> string). 

> belongs_to associations must use the singular term. If you used the pluralized form in the above example for the author association in the Book model and tried to create the instance by Book.create(authors: @author), you would be told that there was an "uninitialized constant Book::Authors". This is because Rails automatically infers the class name from the association name. If the association name is wrongly pluralized, then the inferred class will be wrongly pluralized too.

https://edgeguides.rubyonrails.org/association_basics.html#the-belongs-to-association

src/api/app/controllers/webui/staging/workflows_controller.rb
ChrisBr: Then what about 

```
Staging::Project::ManagersController#update
```

mdeniz: Still we are updating the StagingWorkflow, not the Managers group

ChrisBr: Then 

``
Staging::Workflow::ManagersController#update
``

C'mon, either say you don't want to do it or come up with some other idea ...

mdeniz: I won't do it :heart:

mdeniz: `Staging::Workflow::AssignManagersController#create` ??

mdeniz: Nevermind, I will rename the `assign_managers_group` action to `update` and let it like it is... in the end is just updating one part of the `StagingWorkflow`

ChrisBr: Ok, that's also a solution and we can extend in the future.

src/api/app/controllers/webui/staging/workflows_controller.rb
ChrisBr: > be limited to one group in the future.

If this is the case, we have to change the whole assocation anyway in the future and everything what is using it.

> Which wont happen because we use

Yes, for this special case not. But does it mean just because we don't use this feature of Rails we ignore conventions? Conventions make it easier to work with the code. Do we now name controllers singular because we can hardcode the name in the routes? Of course not! This will cause somebody sometime to debug in the future because the association is not behaving as expected... Not mentioning that you don't get what you expect. Why are we ignoring conventions without having a reason?

But seems like I'm fight one's last stand, so ending this conversation then.

mdeniz: I think I will change it to `managers_group` to follow Rails convention and we can change it afterwards if turns into a polymorphic association with User and/or Group

src/api/app/assets/javascripts/webui2/staging_workflow.js
```diff
@@ -7,3 +7,21 @@ function setSpinnersForDeletion() { // jshint ignore:line
     $(this).find('.delete-spinner').addClass('d-none');
   });
 }
+
+function autocompleteStagingManagersGroup() { // jshint ignore:line
```
Ana06: Maybe we should disable this cop. We are adding the `ignore:line` everywhere! :weary: (not necessarily in this PR)

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -56,14 +59,39 @@ def destroy
     @staging_workflow.staging_projects.where(id: params[:staging_project_ids]).destroy_all
 
     if @staging_workflow.destroy
-      flash[:success] = "Staging Workflow for #{@project.name} was successfully deleted."
+      flash[:success] = "Staging for #{@project.name} was successfully deleted."
```
Ana06: [nitpick] `.name` not needed

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -56,14 +59,39 @@ def destroy
...
       render js: "window.location='#{project_show_path(@project)}'"
     else
-      flash[:error] = "Staging Workflow for #{@project.name} couldn't be deleted: #{@staging_workflow.errors.full_messages.to_sentence}."
+      flash[:error] = "Staging for #{@project.name} couldn't be deleted: #{@staging_workflow.errors.full_messages.to_sentence}."
```
Ana06: [nitpick] Same here: `.name` not needed

src/api/app/models/staging/staging_project.rb
```diff
@@ -59,6 +59,17 @@ def overall_state
...
+      Relationship.add_group(self, managers, role, nil, true)
+    end
+
+    def unassign_managers_group(managers)
+      role = Role.find_by_title!('maintainer')
+      relationships.find_by(group: managers, role: role).try(:destroy)
```
Ana06: I wonder if we should use `destroy!` instead

src/api/spec/factories/staging_workflow.rb
```diff
@@ -1,5 +1,8 @@
 FactoryBot.define do
   factory :staging_workflow, class: 'Staging::Workflow' do
+    project
+    association :managers, factory: :group, title: 'staging-workflow-managers'
```
Ana06: I think this should be `managers_group`

src/api/app/models/staging/staging_project.rb
```diff
@@ -59,6 +59,17 @@ def overall_state
       @overall_state ||= state
     end
 
+    def assign_managers_group(managers)
```
ChrisBr: I think we should do it in a callback. This way we don't need to do it always manually. We could check if the manager of the workflow changed and if yes we update the assocation.

Sth like
```ruby
before_save :update_staging_managers

def update_staging_managers
       managers = staging.staging_workflow.managers
       role = Role.find_by_title!('maintainer')
       return if relationships.find_by(group: managers, role: role)
       Relationship.add_group(self, managers, role, nil, true)
end
```

(untested)

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -12,6 +12,7 @@ def create
 
         staging_project_name = "#{@staging_workflow.project}:Staging:#{params[:staging_project_name]}"
         staging_project = @staging_workflow.staging_projects.build(name: staging_project_name)
+        staging_project.assign_managers_group(@staging_workflow.managers_group)
```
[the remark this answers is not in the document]
DavidKang: Make sense, I will move it in a `before_create`

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -56,14 +59,39 @@ def destroy
...
+      flash[:error] = "Staging for #{@project} couldn't be deleted: #{@staging_workflow.errors.full_messages.to_sentence}."
       render js: "window.location='#{staging_workflow_path(@staging_workflow)}'"
     end
   end
 
+  def update
+    authorize @staging_workflow
+
+    begin
+      Staging::Workflow.transaction do
```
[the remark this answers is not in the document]
DavidKang: :thinking:, we can try it :smile_cat:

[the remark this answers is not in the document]
DavidKang: I think this would not work because we are not modifying any staging_project directly, so the autosave will not be triggered.

src/api/app/assets/javascripts/webui2/staging_workflow.js
mdeniz: If we want to disable the cop is something for another PR. Please open one if you want :smile:

mdeniz:
@ChrisBr @Ana06 feedback implemented, can you please re-review it?

src/api/app/models/staging/workflow.rb
```diff
@@ -60,4 +61,21 @@ def create_staging_projects
...
+    new_managers_group = managers_group
+
+    staging_projects.each do |staging_project|
+      staging_project.unassign_managers_group(old_managers_group)
+      staging_project.assign_managers_group(new_managers_group)
+      staging_project.store
+    end
+
+    # FIXME: This assignation is need because after store a staging_project
+    # the object is reloaded and we lost the changes.
```
ChrisBr: Wondering why it is lost because when you reload it should be there, no?

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -72,16 +72,7 @@ def update
...
-        new_managers_group = Group.find_by(title: params[:managers_title])
-
-        @staging_workflow.staging_projects.each do |staging_project|
-          staging_project.unassign_managers_group(old_managers_group)
-          staging_project.assign_managers_group(new_managers_group)
-          staging_project.store
-        end
-
-        @staging_workflow.managers_group = new_managers_group
+        @staging_workflow.managers_group = Group.find_by(title: params[:managers_title])
```
ChrisBr: Why does it still need to be in a transaction and in a rescue? Why not doing it standard way?

```ruby
  @staging_workflow.managers_group = Group.find_by(title: params[:managers_title])
  if save
    ...
  else
    ...
  end
```

src/api/app/models/staging/workflow.rb
```diff
@@ -60,4 +61,21 @@ def create_staging_projects
...
     end
   end
+
+  def update_staging_projects_managers_group
+    return unless changes[:managers_group_id]
+
+    old_managers_group = Group.find(changes[:managers_group_id].first)
+    new_managers_group = managers_group
+
+    staging_projects.each do |staging_project|
```
ChrisBr: This could also happen in the StagingProject callback, no? So you just have to iterate over the projects and store?

mdeniz: No, the callback in `StagingProject` is only a  `before_create`, on update it won't be called

src/api/app/models/staging/workflow.rb
mdeniz: Tried in the console. See the managers_group_id value on top and on the bottom, it just changes back after a store of the first staging project:
image

Looks like the store of the `StagingProject` reloads the staging workflow and then the old managers groups is recovered...

src/api/app/models/staging/workflow.rb
ChrisBr: But why? I mean it could be also a ``before_save`` and we always update when it changed?

mdeniz:
@ChrisBr re-review it please :+1:

coolo:
So I setup an openSUSE:Factory on build-test with 2 rings and 2 staging projects and factory-staging group as reviewer.

I suggest you guys deploy your branch there and give it a try.

hennevogel:
> I suggest you guys deploy your branch there and give it a try.

because?

src/api/spec/models/staging/staging_project_spec.rb
```diff
@@ -3,7 +3,11 @@
 
 RSpec.describe Staging::StagingProject, vcr: true do
   let(:user) { create(:confirmed_user, login: 'tom') }
-  let(:staging_workflow) { create(:staging_workflow_with_staging_projects, project: user.home_project) }
+
+  let(:managers_group) { create(:group) }
+  let(:other_managers_group) { create(:group) }
+
+  let(:staging_workflow) { create(:staging_workflow_with_staging_projects, project: user.home_project, managers_group: managers_group) }
```
Ana06: I think we should move this to the factory.... I'll send a PR

hennevogel:
we have a functional development environment and an do this anywhere else, no need to fumble with obs-test...

coolo:
If you say so

PR #6282, opened 2018-11-15
Convert custom validations into Rails validations

bgeuken:
This implements validation of presence of the various possible
review assignees (package, project, group or user).
In addition this ensures that these validations always run and not
only during object creation.

### DIFF
--- a/src/api/app/models/review.rb
+++ b/src/api/app/models/review.rb
@@ -21,7 +21,14 @@ class Review < ApplicationRecord
   validates :reviewer, length: { maximum: 250 }
   validates :reason, length: { maximum: 65_534 }
 
-  validate :check_initial, on: [:create]
+  validates :user, presence: true, if: :by_user?, on: :create
+  validates :group, presence: true, if: :by_group?, on: :create
+  validates :project, presence: true, if: :by_project?, on: :create
+  validates :package, presence: true, if: :by_package?, on: :create
+  validates :by_project, presence: true, if: :by_package?
+
+  validate :review_assignment
+
   # Validate the review is not assigned to a review which is already assigned to this review
   validate :validate_non_symmetric_assignment
   validate :validate_not_self_assigned
@@ -56,6 +63,13 @@ class Review < ApplicationRecord
 
   before_validation :set_reviewable_association
   after_commit :update_cache
+
+  def review_assignment
+    errors.add(:unknown, 'no reviewer defined') unless by_user || by_group || by_project
+
+    return unless validate_reviewer_fields
+    errors.add(:base, 'it is not allowed to have more than one reviewer entity: by_user, by_group, by_project, by_package')
+  end
 
   def validate_non_symmetric_assignment
     return unless review_assigned_from && review_assigned_from == review_assigned_to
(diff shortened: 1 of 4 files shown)


bgeuken:
As a note, previously the validations for the review assignee were running on initial creation but not afterwards. So in theory there could be instances that have such reviews that would become invalid with this PR.
On our two major instances this is not the case. But it might make sense to provide a data migration. What we could do is to split reviews that have a by_group and a by_user review into two reviews.

src/api/app/models/review.rb
```diff
@@ -98,38 +119,6 @@ def assigned_reviewer
     self[:reviewer] || by_user || by_group || by_project || by_package
   end
 
-  def check_initial
-    # Validates the existence of references.
-    # NOTE: they can disappear later and the review should be still
```
coolo: isn't your change breaking exactly this NOTE? I.e. if we remove packages, we invalidate reviews ?

Ana06: Yes, @bgeuken you need to add `on: :create` to the`review_assignment` validation (as it was done in `check_initial` before) to keep the same behaviour. I agreee with @coolo that we may want to still display them.

src/api/app/models/review.rb
```diff
@@ -19,7 +19,14 @@ class NotFoundError < APIError
...
   validates :reason, length: { maximum: 65_534 }
 
-  validate :check_initial, on: [:create]
+  validates :user, presence: true, if: :by_user?
+  validates :group, presence: true, if: :by_group?
+  validates :project, presence: true, if: :by_project?
+  validates :package, presence: true, if: :by_package?
+
+  validate :review_assignment
+  validate :correctness_of_package, if: :by_package?
```
Ana06: project is always required now, this is not needed (it won't be executed if project is not there anyway)

src/api/app/models/review.rb
```diff
@@ -19,7 +19,14 @@ class NotFoundError < APIError
   validates :reviewer, length: { maximum: 250 }
   validates :reason, length: { maximum: 65_534 }
 
-  validate :check_initial, on: [:create]
+  validates :user, presence: true, if: :by_user?
+  validates :group, presence: true, if: :by_group?
+  validates :project, presence: true, if: :by_project?
```
Ana06: All this was not enforced before. You would need to make sure that `by_project` is set to nil if the project is removed :thinking:

src/api/app/models/review.rb
coolo: like 
```
irb(main):002:0> Review.where('by_package is not null and package_id is null').count
=> 3424

File: src/api/app/models/review.rb (hunk 3 above)
Comment (coolo) [reply src/api/app/models/review.rb]: this would break history

File: src/api/app/models/review.rb (hunk 1 above)
Comment (bgeuken) [reply src/api/app/models/review.rb]: Oh, that's the one thing I didn't check... what a mess

Comment (coolo) [reply src/api/app/models/review.rb]: well, we don't remove users and groups that often - otherwise you would see the same there

File: src/api/app/models/review.rb (hunk 3 above)
Comment (Ana06) [reply src/api/app/models/review.rb]: then this validation is wrong :rofl:

Comment (coolo) [reply src/api/app/models/review.rb]: I'm afraid bgeuken's idea just doesn't work.

**Comment by Ana06** [comment]
what is the state of this? :thinking:

File: src/api/app/models/review.rb (hunk 2 above)
Comment (bgeuken) [reply src/api/app/models/review.rb]: The difference is that this validation checks that `by_project` is set when `by_package` is not nil. The other validations are checking the object relations, and only on review creation.

But we can write `correctness_of_package` with the default rails validations.... *updated*

File: src/api/app/models/review.rb [hunk 4]
```diff
@@ -54,6 +61,16 @@ class NotFoundError < APIError
   before_validation :set_reviewable_association
   after_commit :update_cache
 
+  def review_assignment
+    # NOTE: they can disappear later and the review should be still
```
dmarcoux: This comment is now cut. It used to also have these 2 additional lines:

```
# usable to some degree (can be showed at least)
# But it must not be possible to create one with broken references
```

bgeuken: Hm, not sure why I added the comment again... Dropped it now.

IMO there is no reason or need to allow to remove by_* attributes after a review was created. They are duplications of the object references (user_id, group_id, ...) and should in my opinion stay. Even when the referenced object got deleted.
Otherwise I don't see any reason to have these non-relational references in first place.

src/api/app/models/review.rb
```diff
@@ -19,7 +19,14 @@ class NotFoundError < APIError
   validates :reviewer, length: { maximum: 250 }
   validates :reason, length: { maximum: 65_534 }
 
-  validate :check_initial, on: [:create]
+  validates :user, presence: true, if: :by_user?, on: :create
+  validates :group, presence: true, if: :by_group?, on: :create
+  validates :project, presence: true, if: :by_project?, on: :create
+  validates :package, presence: true, if: :by_package?, on: :create
+  validates :by_project, presence: true, if: :by_package?
```
Ana06: do we want to execute this always or only on create?

src/api/app/models/review.rb
```diff
@@ -19,7 +19,14 @@ class NotFoundError < APIError
   validates :reviewer, length: { maximum: 250 }
   validates :reason, length: { maximum: 65_534 }
 
-  validate :check_initial, on: [:create]
+  validates :user, presence: true, if: :by_user?, on: :create
```
Ana06: I think we should return custom error message for this validations

vpereira:
handled on #8005 Thank you @Ana06 and @bgeuken feel free to review it there if you want!

PR #6284, opened 2018-11-16
Move callback into the correct model

DavidKang:
`update_staging_workflow_on_backend` callback was moved into `staging_project` model.


hennevogel:
Hm why to StagingProject and not to StagingWorkflow?

DavidKang:
> Hm why to StagingProject and not to StagingWorkflow?

Because those callbacks are only when a StagingProject are created or destroyed

PR #6286, opened 2018-11-16
Add influxdb-rails

ChrisBr:
to collect performance metrics of the OBS rails app like
* request time
* view render time
* db query time

This information will be collection on metrics.o.o

The influxdb is already reachable on metrics.o.o so we could deploy it. Let me know so I can give you the credentials.

### DIFF
--- a/src/api/config/options.yml.example
+++ b/src/api/config/options.yml.example
@@ -183,6 +183,15 @@ default: &default
   #    persistent: true
   #    passive: true
 
+  # For sending application performance metrics to a influx time series database
+  # influxdb_database: rails
+  # influxdb_username: rails
+  # influxdb_password: 123456
+  # influxdb_hosts:
+  #  - https://domain.tld
+  # influxdb_port: 8086
+  # influxdb_ssl: true
+  # influxdb_retry: 10
 production:
   <<: *default
 
--- a/src/api/Gemfile
+++ b/src/api/Gemfile
@@ -76,6 +76,8 @@ gem 'data_migrate', '= 5.2.0'
 gem 'addressable'
 # for XML builder
 gem 'builder'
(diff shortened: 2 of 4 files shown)


src/api/config/options.yml.example
```diff
@@ -183,6 +183,14 @@ default: &default
   #    persistent: true
   #    passive: true
 
+  # For storing rails metrics in influxdb
```
hennevogel: This file is for Admins. I would suggest:

`# For sending application performance metrics to a influx time series database`

src/api/config/options.yml.example
```diff
@@ -183,6 +183,14 @@ default: &default
...
+  # https://github.com/influxdata/influxdb-rails
```
hennevogel: For admins this url doesn't really have much value as it does not explain any of the settings.

src/api/config/options.yml.example
```diff
@@ -183,6 +183,14 @@ default: &default
...
+  # influxdb_database: rails
+  # influxdb_username:
+  # influxdb_password:
+  # influxdb_hosts:
```
hennevogel: Please fill those with example data, so people at least have the chance to know what is expected.

ChrisBr:
> +1 the only thing I'm worried about is config.retry...

Why that? So you want to disable the retry or increase or decrease or...?

hennevogel:
See [Q: How does this gem handle an unreachable InfluxDB server?](https://github.com/influxdata/influxdb-rails#frequently-asked-questions) in the gems FAQ.

ChrisBr:
> See Q: How does this gem handle an unreachable InfluxDB server? in the gems FAQ.

Yup, I know, that's why I limited to 10 retries. I think it is not critical when we loose some of the data points if InfluxDB is not available.

PR #6288, opened 2018-11-16
Add staging project show page

saraycp:
Show staging project and its status and checks inside staging workflow.

It also makes the links inside overall state (staging workflow page) to point to the staging project.

show_staging_project_new


src/api/app/views/webui2/webui/staging/projects/show.html.haml
```diff
@@ -0,0 +1,36 @@
...
+            %tr
+              = render partial: 'status_items',
+                       locals: { staging_project: @staging_project, items: @staging_project.building_repositories, type: 'building_repositories' }
+            %tr
+              = render partial: 'status_items',
+                       locals: { staging_project: @staging_project, items: @staging_project.broken_packages, type: 'broken_packages' }
+            %tr
+              = render partial: 'checks',
+                       locals: { checks: @staging_project.status_reports.first.checks,
+                                 missing_checks: @staging_project.status_reports.first.missing_checks }
```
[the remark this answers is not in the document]
coolo: no, not at all. Every repository has its own status repo - as has every architecture in every repository.

So @staging_project.repositories.map { |r| r.current_status_report } +
    @staging.repository_architectures.map { |r| r.current_status_report }

And from this array all checks and missing_checks.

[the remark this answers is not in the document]
vpereira: > @staging.repository_architectures.map { |r| r.current_status_report }
you mean @staging_project? @coolo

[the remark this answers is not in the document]
coolo: yes, this was obviously pseudo code :)

src/api/app/views/webui2/webui/staging/projects/_checks.html.haml
```diff
@@ -0,0 +1,16 @@
...
+  Checks
+%td.p-2
+  - if checks.blank? && missing_checks.blank?
+    None
+  - else
+    %ul
+      - missing_checks.each do |name|
+        %li
+          %span.check-pending
+            "#{name} (Expected - Waiting for status to be reported)"
```
mdeniz: I think you don't need the double quotes

src/api/app/views/webui2/webui/staging/projects/show.html.haml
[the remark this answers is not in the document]
saraycp: @coolo can you please check again?

src/api/app/views/webui2/webui/staging/projects/_checks.html.haml
saraycp: done!

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -52,6 +58,10 @@ def set_staging_workflow
         flash[:error] = "Staging with id = #{params[:staging_workflow_id]} doesn't exist"
         return
       end
+
+      def set_bootstrap_views
```
coolo: [nitpick] shouldn't this be in the webui controller? I mean the implementation itself, not the before_action.

src/api/app/helpers/webui/staging/project_helper.rb
```diff
@@ -0,0 +1,9 @@
+module Webui::Staging::ProjectHelper
+  def icon_for_checks(checks, missing_checks)
+    return 'fa-eye text-info' if missing_checks.present?
+    return 'fa-check-circle text-primary' if checks.blank?
+    return 'fa-eye text-info' if checks.any? { |check| check.state == 'pending' }
```
coolo: there is a `pending?` function in check, which you can use

src/api/app/helpers/webui/staging/project_helper.rb
```diff
@@ -0,0 +1,9 @@
...
+    return 'fa-check-circle text-primary' if checks.all? { |check| check.state == 'success' }
```
coolo: so far we checked for failed? first, but your version would work as well - but please add a success? function to check and use it here

src/api/app/models/staging/staging_project.rb
```diff
@@ -100,6 +100,18 @@ def unassign_managers_group(managers)
       relationships.find_by(group: managers, role: role).try(:destroy!)
     end
 
+    def current_status_reports
+      repositories.map(&:current_status_report) + repository_architectures.map(&:current_status_report)
```
coolo: as current_status_report calls into the backend, I think you want to make sure it's done only once. So in one of the models it should be cached.

src/api/app/helpers/webui/staging/project_helper.rb
bgeuken: I would even try to run this on SQL level / via scopes and relations. But I am fine with doing this later, and in another PR.

src/api/app/models/staging/staging_project.rb
ChrisBr: cc @krauselukas maybe you sync with @saraycp about how to implement it. This is basically what we paired yesterday about ...

krauselukas: @ChrisBr thanks, for the hint

src/api/app/controllers/webui/staging/projects_controller.rb
hennevogel: And please rename this to `switch_to_webui2` so we will not forget this when we finishing the bootstrap story.

saraycp: @hennevogel  I can't do that because the original `switch_to_webui2` is already defined in the same controller. Instead of that, I have changed its name and added a comment in order to find it when we search in the future:

```
+  # NOTE: remove when bootstrap migration is done (related to switch_to_webui2)
+  def set_webui2_views
```

src/api/app/helpers/webui/staging/project_helper.rb
saraycp: In another PR then :smiley:

src/api/app/models/staging/staging_project.rb
```diff
@@ -100,6 +104,14 @@ def unassign_managers_group(managers)
       relationships.find_by(group: managers, role: role).try(:destroy!)
     end
 
+    def current_checks
+      relevant_checks + relevant_checks_for_architectures
+    end
+
+    def current_missing_checks
+      (relevant_status_reports + relevant_status_reports_for_architectures).map(&:missing_checks).flatten
```
saraycp: @ChrisBr do you have a better idea for missing_checks?

coolo: missing checks aren't in the database, they are calculcated. So I don't think there is a way

ChrisBr: Does this work when there is no relevant status report?

ChrisBr: But anyway, we can fix this in a follow up / when we merge the PR from @krauselukas. Would prefer to get this in ASAP.

src/api/app/controllers/webui/staging/projects_controller.rb
hennevogel: Also good :-)

src/api/app/models/staging/staging_project.rb
coolo: me too - as there is also #6261 pending also affected by this.

coolo: > Does this work when there is no relevant status report?

Good you ask, because it won't. That's why we created current_status_report on checkables. But with the current coverage, it's pretty easy to mess this up :(

But as already written: let's fix afterwards.

PR #6293, opened 2018-11-18
Editor based on bootstrap

hellcp:


Conan-Kudo:
@hellcp Unfortunately, it looks like the error is legitimate. The OBS PR package build can reproduce the error.

```
[  170s] Sass::SyntaxError: Undefined variable: "$font-family-monospace".
```

hellcp:
@coolo out of curiosity, that stylesheet was already compiled into application css, why list it twice?

hellcp:
Ah, because of javascript, I get to experience it first hand with a beautiful error

ChrisBr:
@hellcp I love it :heart_eyes: 

But I found a few issues:

bootstrap-file-editor

* If you need to vertically scroll, the line content gets into the line numbers (see gif)
* Does it make sense to make the save button red / danger?
* Currently the save button is also disabled when there is no change in the file

See review app
https://obs-reviewlab.opensuse.org/hellcp-editor/package/view_file/home:Admin/ruby/somefile.txt?expand=1

hellcp:
@ChrisBr whoops, forgot about default state for save button. But yes danger, either that or warning, because it's a button that is quite important and should be noticed.
I was actually gonna add a warning when closing tab with unsaved changes, but I forgot yesteday.

The issue with numbers is due to using $card-header-bg, which by default is transparent (and instead of overwriting it in colors.scss, we do that in file loaded after bootstrap, which is not really how that should be done :/)

hellcp:
> Ok, that means there is more refactoring needed to get this working?

Haha, no, I will just use scss function `opacify( mix( $body-bg, opacify($card-cap-bg, 1), alpha( $card-cap-bg ) * 100% ), 1 )`, which will give me non opatique version of default cap (header and footer) background, just to be 100% sure this won't happen again, and then just specify card properties in colors.scss to have the rest of stuff working properly.

hellcp:
Whoops, I forgot that alpha value is reverse of what opatique value is, that will be `opacify( mix( $body-bg, opacify($card-cap-bg, 1), 100% - (alpha( $card-cap-bg ) * 100%) ), 1 )` instead

hellcp:
I guess this time circleci is having a `let me click in the middle of </description> tag` kind of day :P

hellcp:
That's a dumb workaround, but it should work fine until you start using that helper in more places.

coolo:
Please rebase and squash your fixes. Other than LGTM

hellcp:
@coolo here you go :p

hellcp:
Hm, that was weird, my branch wanted to remove some .gitignore files, fixed now

hellcp:
@ChrisBr I don't have idea on how to fix it atm, but it's good if it doesn't save unchanged state.

ChrisBr:
> I don't have idea on how to fix it atm, but it's good if it doesn't save unchanged state.

I think my comment was wrong, it actually does safe. But at least the button is in disabled state. Anyway, I don't think this is blocking us from merging this PR, I can have a look in a follow up PR.

Thanks again for your help :heart: 

Merging ...

PR #6294, opened 2018-11-19
Fix update_staging_workflow_on_backend

ChrisBr:
Because the association is called staging and not staging workflow.


ChrisBr:
Duplicate of https://github.com/openSUSE/open-build-service/pull/6284

PR #6295, opened 2018-11-19
Remove unneeded call to set_buildinfo

Ana06:
It is a consuming call and it is being already done. :bowtie:

### DIFF
--- a/src/api/app/models/staging/staging_project.rb
+++ b/src/api/app/models/staging/staging_project.rb
@@ -95,8 +95,6 @@ module Staging
     end
 
     def build_state
-      set_buildinfo
-
       return :building if building_repositories.present?
       return :failed if broken_packages.present?
 


coolo:
shouldn't this function better know itself when to return without being expensive? Otherwise the model has to know that buildstate is not called without set_buildinfo before.

ChrisBr:
I agree with @coolo 

Another solution could be to just add a guard clause in ``set_buildinfo``

```
return if @broken_packages && @building_repositories
...
```

Ana06:
but `set_buildinfo` is private. The methods that should be used are `broken_packages` and `building_repositories`. :confused:

ChrisBr:
> but set_buildinfo is private. The methods that should be used are broken_packages and building_repositories.

Why does it matter if the method is private? I thought your main point is you don't want to do another backend call + calculations?

bgeuken:
IMO, we should split the code that collects the data from the backend from plugging this into the staging project. 
But I wouldn't spend to much time on this right now. Later on when the bigger chunks are in place we can think about how this should be refactored.

(Maybe do something similar than [LocalBuildResult::ForPackage](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/local_build_result/for_package.rb). Not sure yet what the best approach would be.)

Ana06:
This is still being called one extra time. I suggest getting this merge and then discuss about improvements. This solves the problem.

PR #6296, opened 2018-11-19
Remove staging assocation from Project

ChrisBr:
Meanwhile we introduced the dedicated Staging::StagingProject model, this seems to
be a leftover from this refactoring.


ChrisBr:
My assumption was wrong, we need this association

```
                                                      <- staging project A
main project -> staging workflow
                                                      <- staging project B                           
```

PR #6300, opened 2018-11-19
Add a factory for staging workflow manager groups

bgeuken:
We need to ensure that there are no conflicts between the
generated group titles, to be able to create multiple staging
workflow factories.


bgeuken:
Codecov and Hakiri got stuck again. This is just touching a feature branch and the changes (changing a FactoryBot factory) are unlikely to cause issues there...

=> merging

Issue #5974, opened 2018-10-02
Bootstrap: Wrong icon size for mobile devices on main page


bgeuken:
The icons on the news and latest updates sidebar are not scaling on smaller devices (screenshot was taken on my tablet (Samsung Galaxy S2 8", 1536x2048 pixels) ).

![screenshot_20181001-221705c](https://user-images.githubusercontent.com/968949/46335023-75ab0800-c626-11e8-8b74-dee9c4987eca.png)

dmarcoux:
We need to have a CSS media rule for this, so it will change the size of icons depending on the screen size. Something along those lines: https://stackoverflow.com/a/29299793

coolo:
I find https://stackoverflow.com/a/31352309 more practical.

ChrisBr:
JFYI: it works on my iPhone

![image](https://user-images.githubusercontent.com/3799140/46336693-dbe65980-c62b-11e8-8ae5-faf8d80b88c9.png)

eduardoj:
> JFYI: it works on my iPhone

In my android phone works too, without the icon problem @bgeuken pointed out.

saraycp:
Also fine in my phone. I think the problem is medium-size screens / tablets.

hennevogel:
It is never the device. The web is fluid. It is always just some [responsive breakpoint](https://responsivedesign.is/strategy/page-layout/defining-breakpoints/).

PR #6301, opened 2018-11-19
Change size icon for staging project deletion

bgeuken:
The delete icon was much smaller than the problem icon, which made
the table look a bit strange.
Since project deletion is one of the main actions on the edit view,
we increase it's size to make it more visible.

Before:

Before

After:

After

### DIFF
--- a/src/api/app/views/webui2/webui/staging/workflows/_problems.html.haml
+++ b/src/api/app/views/webui2/webui/staging/workflows/_problems.html.haml
@@ -5,7 +5,7 @@
 
 - if staging_project.problems.empty?
   .text-center
-    %i.fas.fa-check-circle.fa-2x.text-success
+    %i.fas.fa-check-circle.text-success
 - else
   %ul.list-group.list-group-flush
     = render partial: 'problems_item', locals: { staging_project: staging_project, index: 0...max_problems }


bgeuken:
What I still find a bit confusing / misleading is that the two icons and their color make them look like they represent two different actions for the same thing.

Should we make the 'problem' icon on the edit page a bit smaller. Or use different icon sets / colors? What do you think?

dmarcoux:
Somehow, your screenshots are not linked properly.

DavidKang:
https://obs-reviewlab.opensuse.org/bgeuken-staging-workflow_increase_icon_size/staging_workflows/1

bgeuken:
I reduced the size now.

We might want to update the icon sizes on the main page:wink:

mdeniz:
> We might want to update the icon sizes on the main pagewink

Are you talking about these ones?
image

I think that those on the main page are ok, they are meant to represent main things in a front-page. But for the rest of the actions/buttons we use the normal size.

bgeuken:
screenshot-2018-11-20 welcome - opensuse build service

and

screenshot-2018-11-20 welcome - opensuse build service 1

bgeuken:
Now, I remember I've opened an [issue](#5974) for that before

bgeuken:
Codecov and Hakiri got stuck again. This is just touching a feature branch and the changes (icon size) are unlikely to cause issues there...

=> merging

PR #6307, opened 2018-11-20
Fix StagingProject link

DavidKang:
StagingProject was using an incorrect `project name`


DavidKang:
This PR is wrong because the link is a todo

bgeuken:
Yes, but the link was also wrong. Anyway, I also had this patched in [PR6308](https://github.com/openSUSE/open-build-service/pull/6308/files#diff-709337929b85529474fa0f8ad6004a44R4)

PR #6308, opened 2018-11-20
Make the location of the staging project flexible

bgeuken:
Previously we required the location of the staging projects to be
under the 'Staging' namespace of the staging workflow project
(eg. my_workflow_project:Staging:A).
With this commit it will be possible to create the staging project
anywhere as long as the user has permissions to it's location.

### DIFF
--- a/src/api/app/controllers/webui/staging/projects_controller.rb
+++ b/src/api/app/controllers/webui/staging/projects_controller.rb
@@ -10,9 +10,8 @@ module Webui
 
       def create
         authorize @staging_workflow
-
-        staging_project_name = "#{@staging_workflow.project}:Staging:#{params[:staging_project_name]}"
-        staging_project = @staging_workflow.staging_projects.build(name: staging_project_name)
+        staging_project = @staging_workflow.staging_projects.build(name: params[:staging_project_name])
+        authorize(staging_project, :create?)
 
         if staging_project.valid? && staging_project.store
           flash[:success] = "Staging project with name = \"#{staging_project}\" was successfully created"
--- a/src/api/app/views/webui2/webui/staging/workflows/_create_staging_project.html.haml
+++ b/src/api/app/views/webui2/webui/staging/workflows/_create_staging_project.html.haml
@@ -6,8 +6,8 @@
           %h5.modal-title#create-staging-projects-modal-label Create Staging Project
         .modal-body
           .form-group
-            = label_tag(:staging_project_name, "#{staging_workflow.project}:Staging:")
-            = text_field_tag(:staging_project_name, nil, placeholder: 'Eg: A', class: 'form-control')
+            = label_tag(:staging_project_name, 'Please enter a project you have permissions to')
+            = text_field_tag(:staging_project_name, nil, placeholder: "Eg: #{staging_workflow.project}:Staging:A", class: 'form-control')
 
         .modal-footer
           = render partial: 'webui2/shared/dialog_action_buttons'
--- a/src/api/spec/controllers/webui/staging/projects_controller_spec.rb
+++ b/src/api/spec/controllers/webui/staging/projects_controller_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe Webui::Staging::ProjectsController do
   describe 'POST #create' do
     context 'a staging_project' do
       before do
-        post :create, params: { staging_workflow_id: staging_workflow.id, staging_project_name: 'C' }
+        post :create, params: { staging_workflow_id: staging_workflow.id, staging_project_name: 'home:tom:My:Projects' }
       end
 
       subject { staging_workflow }
@@ -20,18 +20,18 @@ RSpec.describe Webui::Staging::ProjectsController do
       it { expect(Project.count).to eq(4) }
       it 'create a new staging project' do
         subject.reload
-        expect(subject.staging_projects.map(&:name)).to match_array(['home:tom:Staging:A', 'home:tom:Staging:B', 'home:tom:Staging:C'])
+        expect(subject.staging_projects.map(&:name)).to match_array(['home:tom:Staging:A', 'home:tom:Staging:B', 'home:tom:My:Projects'])
       end
       it { expect(response).to redirect_to(edit_staging_workflow_path(subject)) }
       it { expect(flash[:success]).not_to be_nil }
       it 'assigns the managers group' do
-        expect(Staging::StagingProject.find_by_name('home:tom:Staging:C').groups).to contain_exactly(subject.managers_group)
+        expect(Staging::StagingProject.find_by_name('home:tom:My:Projects').groups).to contain_exactly(subject.managers_group)
       end
     end
 
     context 'an existent staging project' do
       before do
-        post :create, params: { staging_workflow_id: staging_workflow.id, staging_project_name: 'A' }
+        post :create, params: { staging_workflow_id: staging_workflow.id, staging_project_name: 'home:tom:Staging:A' }
       end
 
       subject { staging_workflow }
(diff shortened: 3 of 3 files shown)


src/api/app/views/webui2/webui/staging/workflows/_create_staging_project.html.haml
```diff
@@ -6,8 +6,8 @@
           %h5.modal-title#create-staging-projects-modal-label Create Staging Project
         .modal-body
           .form-group
-            = label_tag(:staging_project_name, "#{staging_workflow.project}:Staging:")
-            = text_field_tag(:staging_project_name, nil, placeholder: 'Eg: A', class: 'form-control')
+            = label_tag(:staging_project_name, 'Please enter a project you have permissions to')
```
bgeuken: Not sure if this label makes sense anymore. I need a second opinion here:smile:

bgeuken:
[Staging projects don't have to be under the name space of the workflow project](https://trello.com/c/V93w7ms0/953-staging-projects-dont-have-to-be-under-the-name-space-of-the-workflow-project)

src/api/app/views/webui2/webui/staging/workflows/_overall_state.html.haml
```diff
@@ -1,7 +1,7 @@
 %span.badge{ class: "state-#{staging_project.overall_state}" }
   %span.badge.badge-light
     -# TODO: Link to the staging project show page
-    = link_to(staging_project.staging_identifier, project_show_path(staging_project.staging_identifier))
+    = link_to(staging_project.staging_identifier, project_show_path(staging_project))
```
dmarcoux: Why not keep it like before? If the name doesn't fit the naming pattern, then it's just displayed. Otherwise, it's only the staging identifier as with the suggested pattern

dmarcoux:
I tried to test this locally and I got this error:

error

bgeuken:
@dmarcoux The exception doesn't look like it's related to my changes. Did you check if this happens in the staging-workflow branch?

dmarcoux:
The error also happens in the `staging-workflow` branch...

src/api/app/views/webui2/webui/staging/workflows/_overall_state.html.haml
bgeuken: You mean to move the change I made from the staging_identifier method to this link, and wherever else we need it? :thinking:

I think we can move this method to a helper and rename it. Essentially we want something that gives us the letter if it fits the default naming scheme and otherwise return the full name.

dmarcoux: I mean that the change to staging_identifier implies that you don't need to change it in other places, like this line.

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -9,9 +9,10 @@ class ProjectsController < WebuiController
 
       def create
         authorize @staging_workflow
+        staging_project = ::Staging::StagingProject.new(name: params[:staging_project_name])
+        authorize(staging_project, :create?)
 
-        staging_project_name = "#{@staging_workflow.project}:Staging:#{params[:staging_project_name]}"
-        staging_project = @staging_workflow.staging_projects.build(name: staging_project_name)
+        @staging_workflow.staging_projects << staging_project
```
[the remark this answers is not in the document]
bgeuken: Just double checked and apparently `<<` would save the record without running validations. I thought it be similar to `save` (not save the record, but keep the "invalid state").

I change it back


Just double checked...

bgeuken:
@ChrisBr Updated. Please check again

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -9,11 +9,10 @@ class ProjectsController < WebuiController
...
       def create
         authorize @staging_workflow
+        staging_project = @staging_workflow.staging_projects.build(name: params[:staging_project_name])
+        authorize(staging_project, :create?)
 
-        staging_project_name = "#{@staging_workflow.project}:Staging:#{params[:staging_project_name]}"
-        staging_project = @staging_workflow.staging_projects.build(name: staging_project_name)
-
-        if staging_project.valid? && staging_project.store
+        if staging_project.save && staging_project.store
```
DavidKang: The reason for using `valid?` is because `store` already do a `save!` and `write_to_backend`. So maybe is better use `write_to_backend` instead of `store`. :thinking:

src/api/app/models/staging/staging_project.rb
```diff
@@ -11,7 +11,7 @@ class StagingProject < Project
     before_create :add_managers_group
 
     def staging_identifier
-      name[/.*:Staging:(.*)/, 1]
+      name[/.*:Staging:(.*)/, 1] || name
```
hennevogel: Why do we need this at all?

src/api/app/views/webui2/webui/staging/workflows/_create_staging_project.html.haml
hennevogel: It doesn't really, what about making this an autocomplete for `Project.for_user|Project.for_group`?

src/api/app/models/staging/staging_project.rb
bgeuken: The identifier?

hennevogel: I would argue this should be default. But this is nitpicking...

src/api/app/views/webui2/webui/staging/workflows/_create_staging_project.html.haml
bgeuken: This project doesn't exist yet. This form would create it.

src/api/app/models/staging/staging_project.rb
bgeuken: The staging_identifier already does that. If you meant to overwrite the `name`... I wouldn't do that since this might create side effects. For example when storing the staging project we also store the `name`.

src/api/app/views/webui2/webui/staging/workflows/_create_staging_project.html.haml
hennevogel: Yes but you can only create projects below `Project.for_user|Project.for_group` :-)

src/api/app/models/staging/staging_project.rb
bgeuken: We just discussed this in person and concluded that we don't want to have a staging identifier. Since there is no hard rule to identify what the identifier would be.

Ana06: I find it fine like that. Otherwise it will return `nil` if it doesn't include `Staging` :confused:

src/api/app/views/webui2/webui/staging/workflows/_create_staging_project.html.haml
Ana06: I also find this confusing. What about: `Enter project name (you need permissions to create it)` or just `Enter project name` and if he doesn't have permissions he will get an error. :thinking:

src/api/app/models/staging/staging_project.rb
bgeuken: We dropped the staging identifier now.

bgeuken:
I've rebased the PR on the recent changes in our feature branch.

@hennevogel I would like to merge this and do the autocomplete part separately. Fine with you?

src/api/app/models/staging/staging_project.rb
Ana06: :angry:

PR #6317, opened 2018-11-20
Put user provided id in apostrophes

bgeuken:
It's a tiny change, but it makes it much clearer that OBS wasn't able to
fetch the id.

Before:

<status code="invalid_request">
  <summary>Could not assign requests  to home:Admin:Staging:A.</summary>
</status>

After:

<status code="invalid_request">
  <summary>Could not assign requests '' to home:Admin:Staging:A.</summary>
</status>

### DIFF
--- a/src/api/app/controllers/staging/staged_requests_controller.rb
+++ b/src/api/app/controllers/staging/staged_requests_controller.rb
@@ -21,7 +21,7 @@ class Staging::StagedRequestsController < ApplicationController
       render_error(
         status: 400,
         errorcode: 'invalid_request',
-        message: "Could not assign requests #{unassigned_requests.to_sentence} to #{@staging_project}."
+        message: "Could not assign requests '#{unassigned_requests.to_sentence}' to #{@staging_project}."
       )
     end
   end


src/api/app/controllers/staging/staged_requests_controller.rb
```diff
@@ -21,7 +21,7 @@ def create
       render_error(
         status: 400,
         errorcode: 'invalid_request',
-        message: "Could not assign requests #{unassigned_requests.to_sentence} to #{@staging_project}."
+        message: "Could not assign requests '#{unassigned_requests.to_sentence}' to #{@staging_project}."
```
coolo: these are numbers, no? Wouldn't it make more sense to have them in [] then?

bgeuken: Currently it would look like this:

```ruby
> [1,2,3].to_sentence
=> "1, 2, and 3"
```

Thinking about it... if people want to build tools around this it might be better to provide an output that is easier to parse.

bgeuken: We could show the plain array here, but I guess to make it easy to parse we would need to provide each missing number as an XML element. @coolo Am I right?

coolo: I don't think tools parse XML errors. If it's an error, it's for humans to understand IMO.

coolo: And I mean "count not assign requests [#{[1,2,3].to_sentense}]"

coolo: because "requests '1,2'" looks like the string 1,2 - not a list of numbers

coolo: That human can also understand the error message "Could not assign requests 1, 2, and 3 to ..", no? :)

I for one would be very confused why it put the numbers in quotes - as it's normally done to mark a literal. So scratch the to_sentense and go with "Could not assign requests (1, 2, 3) to .." or "Could not assign requests () to .."

coolo:
Wait a second - the render_error you're patching is in !empty? - why are you patching for the empty case then?

src/api/app/controllers/staging/staged_requests_controller.rb
bgeuken: You are ~~overengineering~~ overthinking this. This is just a tiny patch to make it clearer that the request number couldn't be parsed.

"Could not assign requests  to home:Admin:Staging:A." could mean anything.

"Could not assign requests '' to home:Admin:Staging:A." makes it clear that the request number couldn't be parsed / found. Be it because of a typo in the XML or whatnot.

bgeuken:
> @coolo commented an hour ago
> Wait a second - the render_error you're patching is in !empty? - why are you patching for the empty case then?

I explained it [here](https://github.com/openSUSE/open-build-service/pull/6317#discussion_r235145151) already.

coolo:
Not to me. The code is `if array.empty? render_ok else render array.to_sentense end`

So why do you care how an empty array looks in to_sentense? I just don't understand. 

And all I want you to do is use () instead of '' if you want to render arrays of numbers differently

bgeuken:
> coolo commented 2 hours ago
>
> And all I want you to do is use () instead of '' if you want to render arrays of numbers differently

I disagree, since this is an error message and not some method call developers have to use. But I'll change it so we can stop discussing dashes and parentheses:smiley:

PR #6318, opened 2018-11-20
Allow use of accesskey to trigger reload

bmwiedemann:
Allow use of accesskey to trigger reload
Because users are often waiting for status to update
and do not want to move their mouse to the tiny reload button

Uses Alt-Shift-R in Linux Firefox to trigger a reload

Only slightly tested with manual firebug patching on project/show and package/show pages (new/Beta style)

### DIFF
--- a/src/api/app/views/shared/_buildresult_box.html.haml
+++ b/src/api/app/views/shared/_buildresult_box.html.haml
@@ -7,7 +7,7 @@
   %ul{ id: "result_select_#{index}" }
     %li.selected
       %a{ href: '#', id: "result_select_link_#{index}_0", data: { id: "result_display_#{index}_0" } } Build Results
-      = sprite_tag('reload', title: 'Reload', onclick: "update_build_result_#{index}()", class: "result_reload", id: "result_reload_#{index}_0")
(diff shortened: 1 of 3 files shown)


bmwiedemann:
https://build.opensuse.org/package/live_build_log/OBS:Server:Unstable:TestGithub:PR-6318/obs-server:obs-api-testsuite-rspec/openSUSE_15.0/x86_64 had a failed test with

> [ 2787s]        expected to find text "Successfully branched package" in "500: Internal Server Error\nOooops, something went terribly wrong, our application crashed :-( We think it's best if you search if this issue is already reported on GitHub.\nSearch\nIn case of emergency...\nLet's be honest, we're both not perfect (yet!). But don't despair, as free software project we're commited to work together and collaborate. Here is how we do that.\n...we help each other out\nOn IRC and on the mailing list there is a vibrant community of developers, admins and users that support each other.\n...we tell us about problems\nIn the unfortunate event you've hit a bug in the OBS we depend on you to report them on GitHub.\n...we better ourselves\nYou have an idea how to make OBS better, faster, stronger? Those ideas we call feature and we track them in openFATE."

ChrisBr:
@Ana06 it is also in the new UI, no? 

But we should maybe add a hint to the UI explaining the shortcut.

bmwiedemann:
The usual way to hint this, is to have the "R" in "Reload" underlined, but <u>R</u> did not work there - might need CSS?

PR #6319, opened 2018-11-21
Improve style of overall state icon

bgeuken:
Add a margin to the staging project identifier to have a
clear seperation from the state type and the border of the outer
part of the "icon".

Before:
screenshot-2018-11-21 staging for home admin - open build service 1

After:
screenshot-2018-11-21 staging for home admin - open build service


bgeuken:
That's were I had a look first, but since this partial didn't follow the `@extend` pattern I added it in there.

I am fine with changing it. Though at some point we need to decide which way to go and than do it everywhere the same.

PR #6324, opened 2018-11-21
Staging workflow managers group improvements

dmarcoux:
We need to use `redirect_to` instead of `render` as otherwise, the value of `action_name` is `create` instead of `new`. This would bring a lot of issues in the rendering of the breadcrumbs.

Before:
before

After (with an error example in the flash):
managers-group

### DIFF
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -22,12 +22,18 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
 
     staging_workflow.managers_group = Group.find_by(title: params[:managers_title])
 
+    unless staging_workflow.managers_group
+      flash[:error] = "Managers Group #{params[:managers_title]} couldn't be found"
+      redirect_to new_staging_workflow_path(project_name: @project)
+      return
+    end
+
     if staging_workflow.save
       flash[:success] = "Staging for #{@project} was successfully created"
       redirect_to staging_workflow_path(staging_workflow)
     else
       flash[:error] = "Staging for #{@project} couldn't be created"
-      render :new
+      redirect_to new_staging_workflow_path(project_name: @project)
     end
   end
 
--- a/src/api/app/views/webui2/webui/staging/workflows/new.html.haml
+++ b/src/api/app/views/webui2/webui/staging/workflows/new.html.haml
@@ -4,21 +4,21 @@
   = render(partial: 'webui2/webui/project/tabs', locals: { project: @project })
   .card-body
     %h3= @pagetitle
-    .mb-3
+    %p
       Adding the staging capability to this project will enable you to group
       requests into staging projects and being able to see how they build together.
     .mb-3.w-50.m-auto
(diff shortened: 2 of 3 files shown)


dmarcoux:
[Known Bugs](https://trello.com/c/M2f4JqsF/954-known-bugs)

src/api/app/views/webui2/webui/staging/workflows/new.html.haml
```diff
@@ -4,21 +4,23 @@
   = render(partial: 'webui2/webui/project/tabs', locals: { project: @project })
   .card-body
     %h3= @pagetitle
-    .mb-3
+    %p
       Adding the staging capability to this project will enable you to group
       requests into staging projects and being able to see how they build together.
+    %p
+      Managers groups can only be created by an admin for now.
```
bgeuken: I would drop the 'for now'. It could be that this stays there forever and if it changes some day we would change this message.

And I would direct user towards a solution here, like 'Please contact an admin if don't have a group yet'.

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -22,12 +22,18 @@ def create
 
     staging_workflow.managers_group = Group.find_by(title: params[:managers_title])
 
+    if staging_workflow.managers_group.nil?
```
bgeuken: unless

src/api/app/views/webui2/webui/staging/workflows/new.html.haml
dmarcoux: Good suggestion. In addition to this, I would like to direct the user even more. What is the best way to contact an admin? IRC, our mailing list, something else?

bgeuken: Well, it depends on the instance. For OBS it's an email address, admin@opensuse.org. That one is stored in [Configuration[:admin_email]](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/configuration.rb#L36)

dmarcoux:
Please have another look.

src/api/app/views/webui2/webui/staging/workflows/new.html.haml
```diff
@@ -4,21 +4,25 @@
   = render(partial: 'webui2/webui/project/tabs', locals: { project: @project })
   .card-body
     %h3= @pagetitle
-    .mb-3
+    %p
       Adding the staging capability to this project will enable you to group
       requests into staging projects and being able to see how they build together.
+    %p
+      Managers groups can only be created by an admin. Please contact an
+      = Configuration[:admin_email] == 'unconfigured@openbuildservice.org' ? 'admin' : mail_to(Configuration[:admin_email], 'admin')
```
bgeuken: Let's move this to a helper

hennevogel:
codecov|hakiri hang. merging.

coolo:
> codecov|hakiri hang. merging.

isn't it time to remove these checks as required for staging-workflow? They obviously just don't report

coolo:
Did that now

PR #6329, opened 2018-11-21
Merge master into staging workflow branch

bgeuken:


bgeuken:
I got a few merge conflicts for files we didn't touch (backend files and a test file in dist/, see the merge commit message). Not sure why this would happen. Anyway, in all those cases I picked the change from master.

bgeuken:
@openSUSE/obs-frontend Please review

dmarcoux:
Hum... it's always a bit difficult to tell with so many changes. I guess that if the specs are green, everything should be fine, right?

dunno2

PR #6331, opened 2018-11-21
Avoid overlapping UI elements in revision diff

### LINKED ISSUE #6330: newUI rdiff has overlapping elements
obs-webui-bug

# Issue Description
new UI version of rdiff has overlapping 'float-right' elements
in both chromium and firefox

Expected Result
============
Overlapping elements should be avoided - e.g. by wrapping the left or right parts

Dropping the `btn-group` class seems to help. Adjusting div sizes from 10+2 to 9+3 could also be good then.

How to Reproduce
===============
https://build.opensuse.org/package/rdiff/home:bmwiedemann:reproducible:test/llvm6?opackage=llvm6&oproject=devel%3Atools%3Acompiler&rev=3

bmwiedemann:
Avoid overlapping UI elements in revision diff
by dropping btn-group class that prevents wrapping of individual buttons

Fixes #6330

### DIFF
--- a/src/api/app/views/webui2/webui/package/rdiff.html.haml
+++ b/src/api/app/views/webui2/webui/package/rdiff.html.haml
@@ -9,7 +9,13 @@
   = render partial: 'tabs', locals: { project: @project, package: @package }
   .card-body
     .row
-      .col-sm-10
+      .col-12
+        - if @filenames.present?
+          .btn-group.float-right
+            %button.btn.btn-sm.btn-outline-secondary#expand-diffs
+              Expand all
+            %button.btn.btn-sm.btn-outline-secondary#collapse-diffs
+              Collapse all
         - if @opackage
           %h3
             Difference Between Revision #{@rev}
@@ -17,13 +23,6 @@
         - else
           %h3
             Changes of Revision #{@rev}
-      - if @filenames.present?
-        .col-sm-2
-          .btn-group.float-right
-            %button.btn.btn-sm.btn-outline-secondary#expand-diffs
-              Expand all
-            %button.btn.btn-sm.btn-outline-secondary#collapse-diffs
-              Collapse all
     %br
     .row
       .col-sm-12


dmarcoux:
Thank you for the PR.

It looks a bit weird on a desktop:
comment

However, it seems to be fine on a mobile:
comment-2

bmwiedemann:
The desktop version still looks better than the screenshot in #6330 
We could still resize the divs from 8+2 to 7+3

dmarcoux:
We also want to keep the button group. I found a solution by adjusting the breakpoints of the button group from `small` to `medium`.

dmarcoux:
Here's a preview (sorry, it's a bit choppy with a gif...)
preview

bmwiedemann:
That looks good in your .gif, but when I apply it in my Firefox, it still looks as bad with overlapping elements in half of the sizes.

bmwiedemann:
The original screenshot was from chromium. Here are 2 different sizes from firefox with the "md" applied manually via F12

# wide window
obs-6330

# slim window
obs-6330b

dmarcoux:
I think that manually adding the class after the page is rendered doesn't produce the same result.

I just enabled the review app and added a [test case](https://obs-reviewlab.opensuse.org/bmwiedemann-rdiff/package/rdiff/home:Admin:branches:home:Admin:branches:home:Admin-test-test/ctris?opackage=ctris&oproject=home%3AAdmin%3Abranches%3Ahome%3AAdmin-test-test&rev=2), so please have a look and let me know if it's fine or not. You can login with `Admin` / `opensuse` to enable the Bootstrap version of the website.

bmwiedemann:
still bad here:
https://obs-reviewlab.opensuse.org/bmwiedemann-rdiff/package/rdiff/home:Iggy:branches:home:Admin/ruby?opackage=ruby&oproject=home:Admin&rev=2

dmarcoux:
I use the same font, but with the size 16. I can reproduce it now when using the size 20.

dmarcoux:
A possible fix would be to put the `h3` and the buttons by themselves, so both have the class `col-12`. This fixes the issue for sure. However, it's not using the available space in some cases (when the content of `h3` is short)

dmarcoux:
Please have another look and let me know what you think.

dmarcoux:
I will squash the commits

dmarcoux:
Thank you @bmwiedemann for your responsiveness!

PR #6332, opened 2018-11-21
Empty project link to staging project instead of project

saraycp:
Talking with @bgeuken we realized that inside Infos box there is a list of Empty Projects and in the old UI they pointed to the project show page. We consider they should point to the staging project instead.

### DIFF
--- a/src/api/app/views/webui2/webui/staging/workflows/_empty_projects_list.html.haml
+++ b/src/api/app/views/webui2/webui/staging/workflows/_empty_projects_list.html.haml
@@ -4,4 +4,4 @@
   %ul.pl-4
     - projects.sort.each do |project|
       %li
-        = link_to(project.staging_identifier, project_show_path(project.name))
+        = link_to(project.staging_identifier, staging_workflow_staging_project_path(staging_workflow, project.name))
--- a/src/api/app/views/webui2/webui/staging/workflows/_infos.html.haml
+++ b/src/api/app/views/webui2/webui/staging/workflows/_infos.html.haml
@@ -8,7 +8,7 @@
           %li= link_to managers.title, group_show_path(managers)
 
       %dt Empty projects:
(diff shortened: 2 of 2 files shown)


src/api/app/views/webui2/webui/staging/projects/show.html.haml
```diff
@@ -6,6 +6,8 @@
       = render(partial: 'webui2/webui/project/tabs', locals: { project: @staging_project })
       .card-body
         %h3= @pagetitle
+        %p.mt-3
+          Project: #{link_to(@pagetitle, project_show_path(@project.name))}
```
hennevogel: Isn't this the same as clicking on the **Overview** tab?

bgeuken:
How about using an icon instead? Something like we do [here](https://build.opensuse.org/package/binary/home:bgeuken/test_package/SLE_11_SP4/i586/_buildenv)

src/api/app/views/webui2/webui/staging/projects/show.html.haml
saraycp: yes it is, but is it obvious?

saraycp:
> How about using an icon instead? Something like we do [here](https://build.opensuse.org/package/binary/home:bgeuken/test_package/SLE_11_SP4/i586/_buildenv)

Better than my link! I'll change it... or should I removed it as @hennevogel proposes?

src/api/app/views/webui2/webui/staging/projects/show.html.haml
hennevogel: It has to be right? It's the only way people can reach `Project#show` by clicking.

I guess the reason it was there was because it was not possible to change the *tabs* from inside the engine.

saraycp:
@hennevogel @bgeuken extra link removed

PR #6333, opened 2018-11-21
Drop staging identifier

bgeuken:
As result of a discussion related to #6308 we decided to drop the
identifier, since there is no fixed naming scheme we could rely on.


dmarcoux:
I just removed the now useless specs for `staging_identifier`

bgeuken:
@dmarcoux Oh, right. Thank you :heart:

PR #6334, opened 2018-11-21
Add unique index to staging_requests_exclusions

eduardoj:
Ensures that a request can not be excluded twice for the same staging workflow in the database.


eduardoj:
> There is already a validation right?

Yes, in the model.

hennevogel:
So why do we need this index?

Ana06:
@hennevogel 

> So why do we need this index?

It is nice to double check things in the database level.  It doesn't hurt

eduardoj:
> So why do we need this index?

It can be needed for improving searches that uses both fields. Not needed for now. So dropping.

PR #6336, opened 2018-11-21
 Create excluded request API

Ana06:
Co-authored-by: Moisés Déniz Alemán <mdeniz@suse.com>

### DIFF
--- a/src/api/app/policies/staging/request_exclusion_policy.rb
+++ b/src/api/app/policies/staging/request_exclusion_policy.rb
@@ -1,0 +1,11 @@
+class Staging::RequestExclusionPolicy < ApplicationPolicy
+  def create?
+    group = record.staging_workflow.managers_group
+    user.groups_users.where(group: group).exists? ||
+      ProjectPolicy.new(user, record.staging_workflow.project).update?
+  end
+
+  def destroy?
+    create?
+  end
+end
--- a/src/api/config/routes.rb
+++ b/src/api/config/routes.rb
@@ -737,6 +737,12 @@ OBSApi::Application.routes.draw do
       resource :staged_requests, controller: 'staged_requests', only: [:create, :destroy], constraints: cons
     end
   end
+
+  controller 'staging/excluded_requests' do
+    post 'staging_excluded_requests/:number/:project_name' => :create, constraints: cons
+    delete 'staging_excluded_requests/:number' => :destroy, constraints: cons
+  end
+
   controller :source_attribute do
     get 'source/:project(/:package(/:binary))/_attribute(/:attribute)' => :show, constraints: cons
     post 'source/:project(/:package(/:binary))/_attribute(/:attribute)' => :update, constraints: cons, as: :change_attribute
--- a/src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
+++ b/src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
@@ -1,0 +1,136 @@
+require 'rails_helper'
+
+RSpec.describe Staging::ExcludedRequestsController, type: :controller, vcr: true do
+  let(:user) { create(:confirmed_user, login: 'user') }
+  let(:other_user) { create(:confirmed_user, login: 'other_user') }
+  let(:project) { user.home_project }
+  let(:staging_workflow) { create(:staging_workflow_with_staging_projects, project: project) }
+  let(:group) { staging_workflow.managers_group }
+  let(:manager) { create(:confirmed_user, login: 'manager', groups: [group]) }
+  let(:source_project) { create(:project, name: 'source_project') }
+  let(:target_package) { create(:package, name: 'target_package', project: project) }
+  let(:source_package) { create(:package, name: 'source_package', project: source_project) }
+  let(:review) { create(:review, by_group: group.title) }
+  let(:bs_request) do
+    create(:bs_request_with_submit_action,
+           creator: other_user,
+           target_project: project.name,
+           target_package: target_package.name,
+           source_project: source_project.name,
+           source_package: source_package.name,
+           reviews: [review])
+  end
+
+  before { login(manager) }
+
+  describe 'POST #create' do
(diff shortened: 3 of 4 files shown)


Ana06:
[Create an ignored request model](https://trello.com/c/R42utmXv/898-create-an-ignored-request-model)

src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+        before { subject }
+
+        it { expect(response).to have_http_status(400) }
+      end
+
+      it { expect { subject }.not_to(change { staging_workflow.request_exclusions.count }) }
+    end
+
+    context 'fails: invalid request exclusion' do
+      let(:project_without_staging) { user.home_project }
```
ChrisBr: Huh? https://github.com/openSUSE/open-build-service/pull/6336/files#diff-d6b6bb094c9652eb21cfdfa20e28ed77R7 here it has a staging workflow attached? How does this work?

src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+        it { expect(response).to have_http_status(400) }
+      end
+
+      it { expect { subject }.not_to(change { staging_workflow.request_exclusions.count }) }
+    end
+  end
+
+  describe 'DELETE #destroy' do
+    context 'succeeds' do
+      before { staging_workflow.request_exclusions.create(bs_request: bs_request, description: 'Bye bye') }
```
ChrisBr: Why not a factory?

src/api/app/policies/staging/request_exclusion_policy.rb
```diff
@@ -0,0 +1,17 @@
+class Staging::RequestExclusionPolicy < ApplicationPolicy
+  def initialize(user, record)
+    raise Pundit::NotAuthorizedError, 'request does not exist' unless record
+    @user = user
+    @record = record
+  end
+
+  def create?
+    group = @record.staging_workflow.managers_group
+    user.groups_users.where(group: group).exists? ||
```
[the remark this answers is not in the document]
coolo: because ApplicationPolicy has attr_reader :user, so @user and user are the same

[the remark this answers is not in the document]
coolo: as such the initialize function should just call super - to unconfuse you

src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
Ana06: it seems I didn't push the most recent changes :confounded:

src/api/config/routes.rb
```diff
@@ -737,6 +738,12 @@ def self.public_or_about_path?(request)
       resource :staged_requests, controller: 'staged_requests', only: [:create, :destroy], constraints: cons
     end
   end
+
+  controller 'staging/excluded_requests' do
+    post 'staging_excluded_requests/:number/:project_name' => :create, constraints: cons
```
ChrisBr: Why not moving this up inside the staging resources?

```ruby
scope module: 'staging' do
  resources :staging_projects, only: [:show], param: :name do
    resources :excluded_requests, only: [:create, :destroy]
  end
end
```

My two concerns are:
* the structure /:number/:project_name looks odd
* that we sprinkle the routes to several places

src/api/app/policies/staging/request_exclusion_policy.rb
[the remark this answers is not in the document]
ChrisBr: :man_facepalming: correct

src/api/config/routes.rb
Ana06: It didn't exist when I wrote this, I can move it. What other structure you propose? :confused:

coolo: I also fell over the :number/:project_name, but the more I thought about it, the more it appeared logic to me. So the question is who wins: logic or consistency :)

Because you ignore request X and if there is only one action, the target project is implicit. But if there are multiple target projects, you can filter through this where to ignore it. So required arguments before optional arguments - makes sense, no? :)

The problem is (and this is why it puzzled Chris and me): everywhere else the project is *the* argument.

ChrisBr: +1 for consistency :smile:

mdeniz: > Why not moving this up inside the staging resources?

Excluding request doesn't belong in any way to a staging project.

ChrisBr: It belongs to the staging scope for sure.

> Excluding request doesn't belong in any way to a staging project. It belongs to a staging workflow

I know, my code was not intended for c&p but for talking about it :smile: I still think that we should use resources and be consistent with the other routes (that was already my opinion a few weeks ago when we had discussion with no conclusion). 

I will approve the PR but I don't like that we have now different routes everywhere at all. I really don't like it.

Ana06: Always logic @coolo :wink:

ChrisBr: Thanks for not even moving it into the scope :angry: :angry: :angry:

mdeniz: Sorry, I merged it after you approved it... why did you approved it? hahaha
Sorry man

ChrisBr: Read my comment :thinking:

ChrisBr: > Because you ignore request X and if there is only one action, the target project is implicit. But if there are multiple target projects, you can filter through this where to ignore it. So required arguments before optional arguments - makes sense, no? :)

@coolo This is anyway not supported :smile: Currently the parameter is required and also we don't use the action in the controller to get the target_project ... so not logic :smile:

coolo: then it sounds like a plan not to require it - then Ana's route at least has a reason

PR #6337, opened 2018-11-21
Add disturl to Even::Build descendants payload

Werkov:
As a consumer of the Rabbit MQ events we would find it useful to identify builds by their disturl (the srcmd5 is complicated to match with our data because of packages with `_link`).

This is loosely based on aae1c09174ad6321d71bac5736376ff50b4bca7a.
I have no opportunity to test whether the disturl attribute I've added is initialized at the moment when the event is synthesized and it contains the value the resulting RPM will have. Hence any advice or review is more than welcomed.


src/api/app/models/event/build.rb
```diff
@@ -2,7 +2,7 @@ module Event
   class Build < Base
     self.description = 'Package has finished building'
     self.abstract_class = true
-    payload_keys :project, :package, :sender, :repository, :arch, :release, :readytime, :srcmd5,
+    payload_keys :project, :package, :sender, :repository, :arch, :release, :readytime, :srcmd5, :disturl
```
bgeuken: You are missing a comma here

Werkov: Thanks for the review. I've force pushed a fixed patch.

adrianschroeter:
The disturl only contains data which get transfered already. So this would be duplicate data, no?

coolo:
the disturl contains *expanded* srcmd5 - that's missing. But as it's basically the *only* thing additionally, I guess putting that in as extra is more useful.

Werkov:
@coolo Do you mean adding something like `expandedmd5` attribute? (And assemble the disturl by the receiver?)

coolo:
So you really need the disturl verbatim? But yes, as the disturl is just an aggregation of things already in the event, it looks a little questionable to put there.

adrianschroeter:
The :srcmd5 is not expanded? makes little sense to me to send unexpanded source reference since it is about a concrete build job. We should change this instead maybe.

Werkov:
So, I checked what would be the best name instead of `expandedmd5` into the [DISTURL synthesis](https://github.com/openSUSE/open-build-service/blob/b83f6b2bddd9bbaaaa9581fd8d6da36938babf81/src/backend/BSSched/BuildJob.pm#L774) and surprise, surprise, it's `srcmd5`. I like what @adrianschroeter suggests and with that interpretation, the current `srcmd5` value can be considered a bug (hence no need to pursue backwards compatibility (correct?)). I'll look into reworking my patch in that sense.

coolo:
good we checked :)

mlschroe:
So what's the point? The disturl contains exactly the same source identifier as the already present 'srcmd5'. From the buildjob generation code:
```
$binfo->{'srcmd5'} = $pdata->{'srcmd5'} if $pdata->{'srcmd5'};
$binfo->{'disturl'} = "obs://$obsname/$projid/$repoid/$pdata->{'srcmd5'}-$packid" if defined($obsname) && defined($packid);
```

coolo:
osc log won't show you any related info of build jobs - as the scheduler doesn't even know the link infos. That's what I mean with commit events - osc log is just an history of it :)

coolo:
No worries - this is probably the least understood part of OBS :)

PR #6339, opened 2018-11-22
WIP: Stop duplicating ApplicationPolicy#initialize in the subclasses

Ana06:
Use the parent method instead of override it with the same code. :bowtie:

### DIFF
--- a/src/api/app/policies/project_policy.rb
+++ b/src/api/app/policies/project_policy.rb
@@ -1,17 +1,17 @@
 class ProjectPolicy < ApplicationPolicy
   def initialize(user, record)
-    raise Pundit::NotAuthorizedError, 'record does not exist' unless record
+    require_record(record)
     @user = user
     @record = record
   end
 
   def create?
-    return false unless @user
+    require_user(user)
     @user.can_create_project?(@record.name)
   end
 
   def update?
-    return false unless @user
+    require_user(user)
     return false unless local_project_and_allowed_to_create_package_in?
     # The ordering is important because of the lock status check
     return true if @user.is_admin?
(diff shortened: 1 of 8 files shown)


ChrisBr:
> it is due to this line:

Exactly, it is because sometimes user is required and sometimes not. User is required always in the Application policy therefore we need to overwrite it. Might make sense to do it the other way around, yes.

src/api/app/policies/project_policy.rb
```diff
@@ -1,17 +1,17 @@
...
   def initialize(user, record)
-    raise Pundit::NotAuthorizedError, 'record does not exist' unless record
+    require_record(record)
     @user = user
     @record = record
   end
 
   def create?
-    return false unless @user
+    require_user(user)
```
coolo: there is a difference between return false and raising an exception. This is not create! but create?

src/api/app/policies/project_policy.rb
```diff
@@ -1,17 +1,17 @@
 class ProjectPolicy < ApplicationPolicy
   def initialize(user, record)
-    raise Pundit::NotAuthorizedError, 'record does not exist' unless record
+    require_record(record)
```
coolo: I think I'd prefer a call to super with flags if *not* required.

Ana06:
DO NOT MERGE: WIP

src/api/app/policies/project_policy.rb
Ana06: There is no difference. The gem will raise an exception if it return `false`. So it doesn't matter if you raise it first. Check: https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L70

src/api/app/policies/project_policy.rb
Ana06: how? :thinking:

Ana06:
It seems like I don't find time to finish this. I'll close for now.

PR #6340, opened 2018-11-22
Stop duplicating ApplicationPolicy#initialize in the Staging subclasses

Ana06:
Use the parent method instead of override it with the same code. :bowtie:


Ana06:
I close this until https://github.com/openSUSE/open-build-service/pull/6339 is finished :wink:

PR #6345, opened 2018-11-22
Make peek configurable

hennevogel:
Peek is helpful for OBS developers, that can do something about
performance. Not so much for OBS admins, so disable it by default.

### DIFF
--- a/src/api/app/controllers/application_controller.rb
+++ b/src/api/app/controllers/application_controller.rb
@@ -43,6 +43,7 @@ class ApplicationController < ActionController::Base
 
   # Don't show performance of database queries to users
   def peek_enabled?
+    return false if CONFIG['peek_enabled'] != 'true'
     User.current && (User.current.is_admin? || User.current.is_staff?)
   end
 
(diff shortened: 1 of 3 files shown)


hennevogel:
> is there now way to parse it to a boolean?

Sure it's a bool already. So we can also do

```ruby
return false unless CONFIG['peek_enabled'] == true
```

if that is more to your liking?

ChrisBr:
```
return false unless CONFIG['peek_enabled'] == true
```

That makes my brain explode :boom: ... ``return false if it is not true`` 

Isn't it the same as

```
return CONFIG['peek_enabled']
```

hennevogel:
Nope. Can be nil (no key), string (`peek_enabled = 'true'), any other garbage. And you want the line below to still be the deciding factor if it's enabled...

dmarcoux:
This: `ActiveModel::Type::Boolean.new.cast(CONFIG.fetch('peek_enabled', false))`

It would work when the key is not defined (it defaults to `false`) and [deals with strings like 'true', 'false', etc...](https://github.com/rails/rails/blob/5-2-1/activemodel/lib/active_model/type/boolean.rb)

We would need to put that in a method which receives the config key to fetch and the default value (`false` or `true`)

hennevogel:
Sounds like a (dangerous) `CONFIG` refactoring to me :cowboy_hat_face:

PR #6348, opened 2018-11-22
Add endpoint to list staging projects in a staging workflow

vpereira:


src/api/app/controllers/staging/staged_projects_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    if @project.staging
+      @staging_workflow = @project.staging
+      @staging_projects = @staging_workflow.staging_projects
+    else
+      render_error status: 400, errcode: 'project_has_no_staging_workflow'
+    end
+  end
+
+  def set_project
+    @project = Project.find_by!(name: params[:staging_project_name])
```
ChrisBr: It's not a staging_project, it's a normal project.

src/api/app/views/staging/staged_projects/index.xml.builder
```diff
@@ -0,0 +1,23 @@
+xml.staging_projects do
+  @staging_projects.each do |staging_project|
+    xml.staging_project(name: staging_project.name) do
+      render(partial: 'staging/staging_projects/staged_requests', locals: { staged_requests: staging_project.staged_requests,
```
ChrisBr: Can we not just leverage the show template instead of repeating every partial which is in it?

src/api/app/controllers/staging/staged_projects_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Staging::StagedProjectsController < ApplicationController
```
ChrisBr: Why StagedProjects? We already have app/controllers/staging/staging_projects_controller.rb ?

src/api/config/routes.rb
```diff
@@ -732,8 +732,9 @@ def self.public_or_about_path?(request)
 
   # StagingWorkflow API
   scope module: 'staging' do
-    resources :staging_projects, only: [:show], param: :name do
+    resources :staging_projects, only: [:show, :index], param: :name do
       get 'staged_requests' => 'staged_requests#index', constraints: cons
+      get 'staged_projects' => 'staged_projects#index', constraints: cons
```
ChrisBr: Why adding the route twice? (I know they're going to two different controllers but you're only adding one controller)

src/api/spec/controllers/staging/staged_projects_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  let(:staging_workflow) { create(:staging_workflow_with_staging_projects, project: project) }
+  let(:staging_project) { staging_workflow.staging_projects.first }
+
+  describe 'GET #index' do
+    context 'existing staging_workflow' do
+      before do
+        get :index, params: { staging_project_name: staging_workflow.project.name, format: :xml }
+      end
+
+      it { expect(response).to have_http_status(:success) }
```
ChrisBr: What about the testing the body as well?

src/api/app/controllers/staging/staged_projects_controller.rb
vpereira: the issue here is the routing. As the routes are today, it wouldn't work like that. It's not under ```StagingWorkflow``` and therefore, the route wouldn't accept the project_name as parameter. The route would be /staging/staging_projects. /staging/staging_projects would lead to show. If you look in the webui, we scoped everything under 'staging_workflow'.

hennevogel: Most clever would be to use one controller. But that already didn't happen for the other parts...

ChrisBr: The problem is that you miss the nested resource, you basically have two projects: main project and the staging project. You need to put it into a nested resource

``
resources :staging, only: [], param: 'main_project_name' do
 resources :staging_projects, only: [:index, :show], param: :name do
 end
end
``

It would produce routes like:

```
/staging/:staging_main_project_name/staging_projects
/staging/:staging_main_project_name/staging_projects/:staging_project_name
```

This is how rails suggests it
https://guides.rubyonrails.org/routing.html#nested-resources

Of course this requires that you always add the main project to the route but I don't think this is an issue and also Rails does it in their official documentation.

vpereira: > Most clever would be to use one controller. But that already didn't happen for the other parts...

now its the time to change, after will be hard. The "other parts" came before everything ;)

hennevogel: > now its the time to change

Well, depends. We are going to iterate over this for sure. So if can also be something for the next deliverable.

src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -1,7 +1,24 @@
 class Staging::StagingProjectsController < ApplicationController
   skip_before_action :require_login
 
+  before_action :set_main_project
```
DavidKang: This should be only for 'index', right?

src/api/config/routes.rb
```diff
@@ -731,10 +731,10 @@ def self.public_or_about_path?(request)
   end
 
   # StagingWorkflow API
-  scope module: 'staging' do
-    resources :staging_projects, only: [:show], param: :name do
-      get 'staged_requests' => 'staged_requests#index', constraints: cons
-      resource :staged_requests, controller: 'staged_requests', only: [:create, :destroy], constraints: cons
+  resources :staging, only: [], param: 'main_project_name' do
+    resources :staging_projects, only: [:index, :show], controller: 'staging/staging_projects', param: :name do
+      get 'staged_requests' => 'staging/staged_requests#index', constraints: cons
```
DavidKang: This shouldn't be part of `resources :staged_requests`?

vpereira:
[Create API endpoints for staging workflow show and index](https://trello.com/c/WdKJOubl/949-create-api-endpoints-for-staging-workflow-show-and-index)

src/api/config/routes.rb
ChrisBr: There is no resources but resource, with resources you don't have an index action. Changing to resources is not easily possible because it would create a delete route like this:

```
DELETE /staged_requests/:id
```

However, in the API we provide the id in the body. That's the reason why we use resource + additional index action.

PR #6350, opened 2018-11-22
Implement better error message in Staging::StageRequest

ChrisBr:
to let the user know what is going wrong:
* Request not found
* Backend exception
* Already branched


ChrisBr:
I also like the interface now better because it behaves more like an ActiveRecord. Currently it is only tested by the controller though.

PR #6352, opened 2018-11-22
API documentation for staged requests

bgeuken:

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -1930,7 +1930,6 @@ DELETE /status_reports/repositories/<project>/<repository>/required_checks/<chec
   Delete a required check
 XmlResult: required_checks
 
-
 == Staging
 
 === Excluded Requests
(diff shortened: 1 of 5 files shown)


hennevogel:
If you add an example, without a schema file it needs to be added in the [exception list](https://github.com/openSUSE/open-build-service/blob/master/src/api/test/functional/backend_test.rb#L14) in the test that is failing...

docs/api/api/api.txt
```diff
@@ -1944,6 +1943,21 @@ POST /staging_excluded_requests/<request>/<project>
...
 
+=== Staged Requests
+
+<staging_project>: The name of the staging project, e.g. home:user1:Staging:A
+
+GET /staging_projects/<staging_project>/staged_requests
+  Get all staged requests of a staging project.
+XmlResult: staged_requests
+
+POST /staging_projects/<staging_project>/staged_requests
```
coolo: what is missing here is the format expected. It's something with <number>1</number> from what I saw

vpereira: after the last changes, the routes look more like:

```
/staging/:staging_main_project_name/staging_projects/:staging_project_name/staged_requests(.:format)
```

bgeuken:
@vpereira Updated

PR #6356, opened 2018-11-22
Add missing pieces to Staging API

ChrisBr:
Adding missing pieces to the staging API:

* checks
* missing checks
* overall state
* obsolete requests

Furthermore it does DRY some of the partials.


ChrisBr:
> LGTM, not sure if we need tests thought thinking

It's only view changes, the methods I call in the view are already tested.

coolo:
That's a dangerous way to look at it - as you're talking API here. If the view is messed up, the client will see foobar. This is not like HTML views where the button may have the wrong colour.

ChrisBr:
> That's a dangerous way to look at it - as you're talking API here. If the view is messed up, the client will see foobar. This is not like HTML views where the button may have the wrong colour.

I will extend the specs later.

PR #6358, opened 2018-11-23
Staging workflow -> master

ChrisBr:
This is to setup a review app for the review meeting, do not merge.


docs/api/api/api.txt
```diff
@@ -1946,6 +1945,23 @@ POST /staging/<project>/excluded_requests/<request>?description=<description>
...
+<staging_workflow_project>: The name of the staging workflow project, e.g. home:user1
+<staging_project>: The name of the staging project, e.g. home:user1:Staging:A
+
+GET /staging/<staging_workflow_project>/staging_projects/<staging_project>/staged_requests
+  Get all staged requests of a staging project.
+XmlResult: staged_requests
+
+POST /staging/<staging_workflow_project>/staging_projects/<staging_project>/staged_requests
+  Add one or multiple bs_requests to the staged requests of a staging project.
+XmlBody: remove_staged_requests
```
coolo: why is this called remove_staged_requests in adding? Seems suboptimal :)

ChrisBr:
> @openSUSE/open-build-service do we want to get this merged today? thinking

yo, I'm currently have to do some other stuff but will rebase soon

ChrisBr:
Something here is really odd with, CircleCI is failing because of missing files. We should review this more deeply instead of rushing it into master ....

vpereira:
> Something here is really odd with, CircleCI is failing because of missing files. We should review this more deeply instead of rushing it into master ....

i had the same issue, after some rebase its gone. The css missing is related with webui2, no relation with staging-workflow. When I approved this PR was everything green here :-/

coolo:
80c01ff was all required to fix the test suites :)

src/api/app/assets/stylesheets/webui2/webui2.css.scss
```diff
@@ -38,6 +38,9 @@
 @import 'live_build_log';
 @import 'borders';
 @import 'diff';
+@import 'staging-workflow';
+@import 'repositories';
```
coolo: where is 'repositories' coming from?

coolo: damn. this is another mismerge of bce63b1ddb9651c7c53390be621cbe0d58e37881 - we have repositories twice now

vpereira:
@ChrisBr coolo fixed it already. I would suggest to close this PR

src/api/db/migrate/20181016103905_add_staged_request_id_to_bs_requests.rb
```diff
@@ -0,0 +1,8 @@
+class AddStagedRequestIdToBsRequests < ActiveRecord::Migration[5.2]
+  def change
+    change_table :bs_requests, bulk: true do |t|
+      t.integer :staging_project_id
```
Ana06: We should use `references` here.

PR #6366, opened 2018-11-26
 Use --auto-gen-config to generate .rubocop_todo.yml

Ana06:
The cops in the Metrics department by default get `Max` parameters generated in .rubocop_todo.yml, so that no offenses are reported the next time you run rubocop. The new `--auto-gen-only-exclude` command line option allow us to exclude the files instead. It will still change the maximum if the number of excluded files is higher than the exclude limit. This is specially useful when having only few files which increase the maximum considerably, as it happens in our case for many cops. :bowtie: We discussed like 1 year ago that it would be cool to have this in our code (when we documented and refactored our Rubocop process), but it was not implemented until the last Rubocop release: https://github.com/rubocop-hq/rubocop/pull/6329 :tada: 

I also removed  `dev:lint:rubocop:auto_correct` task, as it is a duplicate of `dev:lint:rubocop:auto_gen_config:all`


src/api/app/models/user.rb
```diff
@@ -102,6 +102,10 @@ class User < ApplicationRecord
   after_create :create_home_project, :track_create
   before_save :send_metric_for_beta_change, if: :in_beta_changed?
 
+  def self.in_beta?
+    where(in_beta: true)
```
eduardoj: Are these lines related to this pull request?

Ana06: not at all :eyes:

coolo:
The "I also removed dev:lint:rubocop:auto_correct" is not true though, is it?

Ana06:
@coolo 

> The "I also removed dev:lint:rubocop:auto_correct" is not true though, is it?

no

PR #6370, opened 2018-11-26
Preserve IDs of RepositoryArchitectures on _meta updates

### LINKED ISSUE #6341: Updating _meta drops required checks on RepositoryArchitecture
The RepositoryArchitecture model stores required checks for built repositories, but these required checks are dropped if updating _meta. As dropping required checks makes staging projects acceptable if they are not.

And as staging projects update _meta all the time, this is a critical issue for SLE Stagings.

coolo:
Fixes #6341

### DIFF
--- a/src/api/app/models/project/update_from_xml_command.rb
+++ b/src/api/app/models/project/update_from_xml_command.rb
@@ -282,24 +282,29 @@ class Project
       current_repo.save! if current_repo.changed?
     end
 
+    def check_for_duplicated_archs!(xml_archs)
+      taken_archs = {}
+
+      xml_archs.each do |archname|
+        if taken_archs.key?(archname)
+          raise SaveError, "double use of architecture: '#{archname}'"
+        end
+        taken_archs[archname] = 1
+      end
+    end
+
     def update_repository_architectures(current_repo, xml_hash)
       old_archs = {}
       current_repo.repository_architectures.each do |repoarch|
         old_archs[repoarch.architecture_id] = repoarch
       end
-      taken_archs = {}
+
+      xml_archs = xml_hash.elements('arch')
+      check_for_duplicated_archs!(xml_archs)
 
       position = 1
-      xml_hash.elements('arch') do |archname|
-        if taken_archs.key?(archname)
-          raise SaveError, "double use of architecture: '#{archname}'"
-        end
-        taken_archs[archname] = 1
-
-        unless Architecture.archcache.key?(archname)
-          raise SaveError, "unknown architecture: '#{archname}'"
-        end
-        arch = Architecture.archcache[archname]
+      xml_archs.each do |archname|
+        arch = Architecture.from_cache!(archname)
 
         if old_archs.key?(arch.id)
           old_archs[arch.id].update(position: position)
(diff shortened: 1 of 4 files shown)


coolo:
@ChrisBr I evaluated acts_as_list to write the loop a little tighter

src/api/app/models/project/update_from_xml_command.rb
```diff
@@ -282,22 +282,28 @@ def update_hostsystem(current_repo, xml_hash)
       current_repo.save! if current_repo.changed?
     end
 
-    def update_repository_architectures(current_repo, xml_hash)
-      # destroy architecture references
-      Rails.logger.debug "delete all repository architectures of repository '#{project.id}'"
-      RepositoryArchitecture.where('repository_id = ?', current_repo.id).delete_all
+    def check_for_duplicated_archs!(xml_archs)
+      taken_archs = {}
```
ChrisBr: ``xml_archs`` is an array I guess? Than this can be simplified with:

```ruby
def check_for_duplicated_archs!(architectures)
  duplicate_architectures = architectures - architectures.uniq 
  return if duplicate_architectures.empty

  raise SaveError, "double use of architecture: '#{duplicate_architectures.first}'"
end
```

src/api/app/models/project/update_from_xml_command.rb
```diff
@@ -282,22 +282,28 @@ def update_hostsystem(current_repo, xml_hash)
...
+    end
+
+    def update_repository_architectures(current_repo, xml_hash)
+      xml_archs = xml_hash.elements('arch')
+      check_for_duplicated_archs!(xml_archs)
+
+      archs = []
+      xml_archs.each_with_index do |archname, position|
+        arch = Architecture.from_cache!(archname)
+        current_repo.repository_architectures.find_or_create_by(architecture: arch).insert_at(position + 1)
```
ChrisBr: The ``+ 1`` is necessary because the loop starts with zero? You could do

```ruby 
xml_archs.each.with_index(1) do |archname, position|
```

coolo: ah, ok. The +1 is just to avoid touching one spec that hardcodes positions

coolo:
archname speaks to me - it's your fault :)

coolo:
rebased and polished

src/api/app/models/project/update_from_xml_command.rb
ChrisBr: Argh, it's removing both elements ... :cry: 

What about

```
%w(i586 i586) & %w(i586)
```

coolo: let me pull out the 'this is a little out of scope for this PR' hammer.

src/api/app/models/project/update_from_xml_command.rb
hennevogel: Please don't. Future us will spend cycles on trying to figure out why +1 is necessary...

src/api/app/models/project/update_from_xml_command.rb
coolo: the problem is that `%w(i586) & %w(i586)` has the same result without duplicating.

PR #6375, opened 2018-11-27
Migrate user home info to Boostrap

Ana06:
Create the layout for the user home page and implement the info partial. :bowtie: The involvement partial will be implemented in a separated PR. Until that is done only show this view in development environment.

image

image

### DIFF
--- a/src/api/app/helpers/webui/user_helper.rb
+++ b/src/api/app/helpers/webui/user_helper.rb
@@ -12,6 +12,16 @@ module Webui::UserHelper
     image_tag(url, width: size, height: size, alt: alt, class: opt[:css_class])
   end
 
+  def webui2_user_image_tag(user, css_class = nil, alt = nil)
+    alt ||= (user ? "#{user.name}'s avatar" : 'Avatar')
+    url = if user.try(:email) && ::Configuration.gravatar
+            "https://www.gravatar.com/avatar/#{Digest::MD5.hexdigest(user.email.downcase)}?s=500&d=wavatar"
+          else
+            'default_face.png'
+          end
+    image_tag(url, alt: alt, class: css_class, width: '100%')
+  end
+
   def _optional_icon(user, opt)
     if opt[:no_icon]
       ''
--- a/src/api/app/views/webui2/webui/user/_edit_dialog.html.haml
+++ b/src/api/app/views/webui2/webui/user/_edit_dialog.html.haml
@@ -1,0 +1,16 @@
+.modal.fade#edit-modal{ tabindex: -1, role: 'dialog', aria: { labelledby: 'branch-modal-label', hidden: true } }
+  .modal-dialog.modal-dialog-centered{ role: 'document' }
+    .modal-content
+      .modal-header
+        %h5.modal-title#branch-modal-label Edit your account
+      = form_for(user, url: user_save_path, method: 'post') do |f|
+        .modal-body
+          = f.hidden_field(:login, id: 'user')
+          .form-group
+            = f.label(:realname, 'Name:')
+            = f.text_field(:realname, class: 'form-control')
+          .form-group
+            = f.label(:email, 'Email:')
+            = f.text_field(:email, required: true, email: true, class: 'form-control')
+        .modal-footer
+          = render partial: 'webui2/shared/dialog_action_buttons'
--- a/src/api/app/views/webui2/webui/user/_info.html.haml
+++ b/src/api/app/views/webui2/webui/user/_info.html.haml
@@ -1,0 +1,63 @@
+= webui2_user_image_tag(user, 'home-avatar')
+
+%h4.home-usertitle
+  .home-username
+    = user.realname
+  .home-login
+    = user.login
+
+- unless user.is_nobody?
+  %p
+    = mail_to user.email, title: "Send mail to #{user.name}" do
+      %i.fas.fa-envelope
+      = user.email
+
+- if groups.any?
+  Member of the #{'group'.pluralize(groups.size)}
+  %ul
+    - groups.each do |group|
+      %li= link_to("#{group.title} (#{group.tasks})", group_show_path(group))
+
+- if role_titles.any?
+  Has the #{'role'.pluralize(role_titles.size)}
+  %ul
+    - role_titles.each do |title|
+      %li= title
+
+- if is_user
+  - if user.in_beta?
+    %p
+      .badge.badge-info
+        In public beta program
+    = link_to(user_save_path(user: { login: user.login, in_beta: 0 }), method: :post) do
+      %i.fas.fa-sign-out-alt
+      Leave public beta program
+  - else
+    = link_to(user_save_path(user: { login: user.login, in_beta: 1 }), method: :post) do
+      %i.fas.fa-sign-in-alt
+      Join public beta program
+  %br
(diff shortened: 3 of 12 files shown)


hellcp:
It would be nice to see some colours on icons in the sidebar. We should expose orange from bootstrap for RSS stuff too.

Ana06:
[User Info](https://trello.com/c/OETRo0Hx/979-user-info)

Ana06:
@hellcp 

> It would be nice to see some colours on icons in the sidebar. We should expose orange from bootstrap for RSS stuff too.

The colors wouldn't indicate anything here at all, right? If the colors don't provide a visual reinforcement, they are in my opinion an unnecessary distraction. I prefer to leave it like that.

hellcp:
@Ana06 I would disagree, considering that `leaving` for example is something that would be associated by `danger` or change by `warning` or `success` depending on kind of change (and changing password is certainly warning if you ask me). And rss is orange, because that's what we were made to expect.

Ana06:
@hellcp 

> considering that leaving for example is something that would be associated by danger or change by warning or success depending on kind of change (and changing password is certainly warning if you ask me). And rss is orange, because that's what we were made to expect.

With leaving you mean the `Leaving the public beta program` link? why is it associated with danger? It is a completely inoffensive action. You can click again to join. I am really against converting the links into a :rainbow: without a compelling reason.

dmarcoux:
The RSS feed icon should be orange for sure. It's always a white foreground on an orange background. For the others, I don't know... I don't have a strong opinion on that.

src/api/app/views/webui2/webui/user/_info.html.haml
```diff
@@ -0,0 +1,63 @@
...
+      %i.fas.fa-sign-in-alt
+      Join public beta program
+  %br
+  - if configuration.accounts_editable?(user)
+    - if account_edit_link.present?
+      = link_to account_edit_link do
+        %i.fas.fa-user-edit
+        Edit your account
+    - else
+      = link_to({ controller: 'user', action: 'save_dialog', user: user }, id: 'save_dialog', remote: true) do
```
DavidKang: Please use routes helper :smile_cat:

src/api/app/views/webui2/webui/user/_info.html.haml
```diff
@@ -0,0 +1,63 @@
...
+        %i.fas.fa-user-edit
+        Edit your account
+    %br
+  - if configuration.passwords_changable?(user)
+    = link_to({ controller: 'user', action: 'password_dialog', user: user }, id: 'password_dialog', remote: true) do
```
DavidKang: same here

src/api/app/views/webui2/webui/user/_info.html.haml
```diff
@@ -0,0 +1,63 @@
...
+      %li= link_to("#{group.title} (#{group.tasks})", group_show_path(group))
+
+- if role_titles.any?
+  Has the #{'role'.pluralize(role_titles.size)}
+  %ul
+    - role_titles.each do |title|
+      %li= title
+
+- if is_user
+  - if user.in_beta?
```
Ana06: I have moved this. Before everybody could see if a user is in the beta program. I think that only the user should see this. Maybe we should change it in the old UI as well.

src/api/app/views/webui2/webui/user/_info.html.haml
```diff
@@ -0,0 +1,63 @@
...
+    %p
+      .badge.badge-info
+        In public beta program
+    = link_to(user_save_path(user: { login: user.login, in_beta: 0 }), method: :post) do
+      %i.fas.fa-sign-out-alt
+      Leave public beta program
```
Ana06: I have move this links to be just after the information of if the user is in the beta or not

src/api/app/views/webui2/webui/user/_edit_dialog.html.haml
```diff
@@ -0,0 +1,16 @@
+.modal.fade#edit-modal{ tabindex: -1, role: 'dialog', aria: { labelledby: 'branch-modal-label', hidden: true } }
+  .modal-dialog.modal-dialog-centered{ role: 'document' }
+    .modal-content
+      .modal-header
+        %h5.modal-title#branch-modal-label Change Your Password
+      = form_for(user, url: user_save_path, method: 'post') do |f|
+        .modal-body
+          = f.hidden_field(:login, id: 'user')
+          .name
+            = f.label(:realname, 'Name:')
```
DavidKang: @Ana06, I think you are missing some classes. This should be something like:

```
 .form-group
   = f.label(:realname, 'Name:')
   = f.text_field(:realname, class: 'form-control')
```

Ana06:
@DavidKang 

> I agree with @dmarcoux, and regarding the others icons I also don't have a strong opinion, but maybe we can play with some colours if we plan to change the colour of the RSS icon.

@hellcp @dmarcoux @DavidKang can we do the colors try in another PR/issue? :pray: I guess it will involved some discussion and I would prefer to get this PR merged ASAP.

hellcp:
@Ana06 I wasn't trying to block it, bootstrap UI is certainly not finished yet, so there will be time and place to do all that.

src/api/app/views/webui2/webui/user/_edit_dialog.html.haml
Ana06: why is it not being followed? It seems to me like our pattern is wrong. It is not following Bootstrap...

Ana06: There is nothing about text fields in the pattern library :confused:

dmarcoux: @Ana06 and I cleared things up. It was marked as resolved, but it wasn't pushed yet. It's now all good.

dmarcoux:
Hakiri's warnings are addressed in another PR: https://github.com/openSUSE/open-build-service/pull/6381

src/api/app/helpers/webui/user_helper.rb
```diff
@@ -12,6 +12,16 @@ def user_image_tag(user, opt = {})
...
 
+  def webui2_user_image_tag(user, css_class = nil, alt = nil)
+    alt ||= (user ? "#{user.name}'s avatar" : 'Avatar')
+    url = if user.try(:email) && ::Configuration.gravatar
+            "https://www.gravatar.com/avatar/#{Digest::MD5.hexdigest(user.email.downcase)}?s=500&d=wavatar"
+          else
+            'default_face.png'
+          end
+    image_tag(url, alt: alt, class: css_class, width: '100%')
+  end
```
bgeuken: How about this:

* We put the `image_tag(...)` to the view
* Define a `gravatar_url` and `gravatar_tooltip` and use them in the `image_tag` call

src/api/app/views/webui2/webui/user/_edit_dialog.html.haml
ChrisBr: Is it the reason why the input fields are not aligned? Because that looks not nice :thinking: 

selection_012

src/api/app/helpers/webui/user_helper.rb
Ana06: why?

PR #6379, opened 2018-11-27
Project attributes on bootstrap

saraycp:

### DIFF
--- a/src/api/spec/bootstrap/features/webui/attributes_spec.rb
+++ b/src/api/spec/bootstrap/features/webui/attributes_spec.rb
@@ -1,0 +1,55 @@
+require 'browser_helper'
+
+RSpec.feature 'Bootstrap_Attributes', type: :feature, js: true do
+  let!(:user) { create(:confirmed_user) }
+  let!(:attribute_type) { create(:attrib_type, name: 'MyImageTemplates') }
+  let!(:attribute) { create(:attrib, project: user.home_project) }
+
+  context 'for a project' do
+    context 'without permissions' do
+      let!(:other_user) { create(:confirmed_user) }
+
+      scenario 'it is not possible to add an attribute, the link is not shown' do
+        login other_user
+
+        visit index_attribs_path(project: user.home_project_name)
+        expect(page).not_to have_content('Add a new attribute')
+      end
+    end
+
+    context 'with permissions' do
+      scenario 'remove attribute' do
+        login user
+
+        visit index_attribs_path(project: user.home_project_name)
+        click_link 'Delete attribute'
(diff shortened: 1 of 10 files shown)


saraycp:
[Attributes](https://trello.com/c/rOvBpFgC/963-attributes)

src/api/spec/bootstrap/features/webui/attributes_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      scenario 'remove attribute' do
+        login user
+
+        visit index_attribs_path(project: user.home_project_name)
+        click_link 'Delete attribute'
+        expect(find("#delete-attribute-modal-#{attribute.id}")).to have_text('Delete attribute?')
+        within("#delete-attribute-modal-#{attribute.id} .modal-footer") do
+          click_button('Delete')
+        end
+        sleep 1 # FIXME: Needed to avoid a flickering test.
```
bgeuken: IIRC you could add a finder for the flash message (or any other element that would appear when the request was processed) here. That way we would use the timeout capybara uses.
I am not 100% this would work here, but I would give it a try.

saraycp: I'll try!

src/api/spec/bootstrap/features/webui/attributes_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+          click_button('Delete')
+        end
+        expect(find('#flash')).to have_text('Attribute sucessfully deleted!')
+      end
+    end
+  end
+
+  context 'for a project with a package' do
+    let!(:package) do
+      create(:package, project_id: user.home_project.id)
```
Ana06: same here

src/api/spec/bootstrap/features/webui/attributes_spec.rb
saraycp: @bgeuken I replaced `sleep 1` by 
```expect(find('#flash')).to have_text('Attribute sucessfully deleted!')```
But it also fails in CircleCI :disappointed:

saraycp: Any other idea?

coolo: the problem is that find('#flash') is an instant return of nil and repeating the expect won't help. So you need expect something repeatable first.

`expect(page).to have_css('#flash');`
and *then* you can expect the flash to have text.

saraycp:
@Ana06  can you please check again?

PR #6385, opened 2018-11-28
Migrate Project#users page to bootstrap

eduardoj:
project_users

### DIFF
--- a/src/api/app/mixins/webui/manage_relationships.rb
+++ b/src/api/app/mixins/webui/manage_relationships.rb
@@ -14,11 +14,7 @@ module Webui::ManageRelationships
     rescue NotFoundError,
            Relationship::SaveError => e
       flash[:error] = e.to_s
-      if params[:webui2]
-        redirect_to users_path
-      else
-        redirect_to add_path(:add_person)
-      end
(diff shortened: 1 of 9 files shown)


src/api/app/mixins/webui/manage_relationships.rb
```diff
@@ -14,11 +14,7 @@ def save_person
     rescue NotFoundError,
            Relationship::SaveError => e
       flash[:error] = e.to_s
-      if params[:webui2]
-        redirect_to users_path
-      else
-        redirect_to add_path(:add_person)
-      end
+      redirect_to add_path(:add_person)
```
dmarcoux: It's like this now because package and project both have a Bootstrap version, right?

eduardoj: > It's like this now because package and project both have a Bootstrap version, right?

Yes.

eduardoj:
[Users](https://trello.com/c/78yGAxL5/967-users)

eduardoj:
@dmarcoux:

> The breadcrumbs are missing.

Right. Added. Please re-review.

PR #6392, opened 2018-11-29
Don't start repository architectures with position 1

coolo:
Webui starts with 0 and that's also the configured 'top'.
And as we rewrite the positions on update, there is no
migration to worry about either. It's just an offset,
but at least future developers won't have to worry if
the 1 is meaningful

### DIFF
--- a/src/api/spec/models/project/update_from_xml_command_spec.rb
+++ b/src/api/spec/models/project/update_from_xml_command_spec.rb
@@ -161,9 +161,8 @@ RSpec.describe Project::UpdateFromXmlCommand do
         )
         Project::UpdateFromXmlCommand.new(project).send(:update_repositories, xml_hash, false)
 
-        expect(repository_1.architectures.map(&:name).sort).to eq(['i586', 'x86_64'])
-        expect(repository_1.repository_architectures.where(position: 1).first.architecture.name).to eq('x86_64')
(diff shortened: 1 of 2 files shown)


coolo:
... and of course all developers will wonder forever why chrisbr put a top: 0 there (including his present himself :)

src/api/spec/models/project/update_from_xml_command_spec.rb
```diff
@@ -161,9 +161,8 @@
         )
         Project::UpdateFromXmlCommand.new(project).send(:update_repositories, xml_hash, false)
 
-        expect(repository_1.architectures.map(&:name).sort).to eq(['i586', 'x86_64'])
-        expect(repository_1.repository_architectures.where(position: 1).first.architecture.name).to eq('x86_64')
-        expect(repository_1.repository_architectures.where(position: 2).first.architecture.name).to eq('i586')
+        expect(repository_1.architectures.map(&:name)).to eq(['x86_64', 'i586'])
+        expect(repository_1.repository_architectures.map { |repoarch| repoarch.architecture.name }).to eq(['x86_64', 'i586'])
```
ChrisBr: Just wondering, are the two asserts now testing more or less the same?

coolo: I would say so - possibly it checks if architectures is sorted as well? But there is a 50:50 chance it's correct anyway :)

PR #6389, opened 2018-11-28
Improve User model efficiency

Ana06:
Improve efficiency of User model queries using pluck. Plucking the ids generate an extra query with the ids and there could potentially be a lot of ids. This make this query way less efficient. :bowtie:

PR #6406, opened 2018-11-30
Refactor change_review_state to be able to add new events to it

coolo:
This is a first step only as I need to avoid breaking the test suite - the code is typical OBS Spaghetti, so it will take a while to sort correctly.


src/api/app/models/bs_request.rb
```diff
@@ -738,6 +738,19 @@ def cancelapproval(opts)
...
 
+  def calculate_state_from_reviews
+    return :declined if reviews.any?(&:declined?)
+    return :review unless reviews.all?(&:accepted?)
+
+    # if it would be new, we need to check if all groups agree
+    bs_request_action_groups.each do |g|
+      if g.find_review_state_of_group == :review
+        return :review
+      end
```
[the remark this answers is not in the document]
coolo: let's come to this once we have working tests. For now I'm trying to reduce the spagetthi-factor, so that you can even see things like n+1 issues :)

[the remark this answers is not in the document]
bgeuken: You didn't touch [find_review_state_of_group](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/bs_request_action_group.rb#L173). So it was visible before your refactoring:wink:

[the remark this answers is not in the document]
bgeuken: The other thing that is strange is that you re-do what `find_review_state_of_group` does, returning `:review` or `:new`. 

btw, I refactored the method in #6407 now.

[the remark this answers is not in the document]
bgeuken: It also looks like this does some round trip. Calling a method of `bs_request_action_groups` which is then calling `bs_requests.*`.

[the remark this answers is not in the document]
coolo: For those not following 6407: I did not dig further into this as I propose to drop it all

src/api/app/models/bs_request.rb
```diff
@@ -806,41 +805,38 @@ def change_review_state(new_review_state, opts = {})
...
       history.create(p) if history
-
-      # pre-approved requests can be processed
-      BsRequestAutoAcceptJob.perform_later(id) if go_new_state == :new && approver
     end
   end
 
   def check_if_valid_review!(opts)
-    return unless !opts[:by_user] && !opts[:by_group] && !opts[:by_project]
+    return if opts[:by_user] || opts[:by_group] || opts[:by_project]
```
coolo: @Ana06 this is btw ssomething for your slides, because what happend rubocop rewrote

```
-    if !opts[:by_user] && !opts[:by_group] && !opts[:by_project]
-      raise InvalidReview.new
-    end
+    return unless !opts[:by_user] && !opts[:by_group] && !opts[:by_project]
+    raise InvalidReview.new
```

Turning rather logic code into a double negation - and I don't dislike double negations :)

src/api/app/models/event/request_reviews_done.rb
```diff
@@ -0,0 +1,29 @@
+module Event
+  class RequestReviewsDone < RequestStatechange
+    self.message_bus_routing_key = 'request.reviews_done'
```
Ana06: if this `self` needed? Shouldn't rubocop tell you something? :rage:

src/api/spec/models/bs_request_spec.rb
```diff
@@ -5,6 +5,7 @@
 # require real backend answers for projects/packages.
 # CONFIG['global_write_through'] = true
 
+# rubocop:disable Metrics/BlockLength
```
Ana06: what about following the advice instead of disabling it or disable it generally as this may happen in rspec generally? :thinking:

src/api/app/models/review.rb
```diff
@@ -229,6 +229,26 @@ def reviewable_by?(opts)
...
+  def change_state(new_state, comment)
+    return false if state == new_state && reviewer == User.current.login && reason == comment
+
+    self.reason = comment
+    self.state = new_state
+    self.reviewer = User.current.login
+    save!
+    Event::ReviewChanged.create(bs_request.notify_parameters)
+
+    arguments = { review: self, comment: comment, user_id: User.current.id }
```
Ana06: you can drop `.id`

src/api/app/controllers/request_controller.rb
```diff
@@ -26,7 +26,7 @@ class SaveError < APIError
   end
 
   def render_request_collection
-    # if all params areblank, something is wrong
+    # if all params are blank, something is wrong
```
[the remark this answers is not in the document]
Ana06: This comment is still not needed

Ana06:
BTW, you need to rebase :wink:

src/api/spec/models/bs_request_spec.rb
coolo: Disabling it for rspec generally sounds good - but outside of this PR.

src/api/app/controllers/request_controller.rb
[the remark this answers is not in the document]
coolo: took the diff out of the PR

src/api/app/models/event/request_reviews_done.rb
coolo: It's in all other events too - I won't start a difference here. If it's not necessary, it shouldn't be needed for none of the attributes in none of the events, so let's clean this up independently.

coolo:
rebased, please have another look

coolo:
could I have a 3rd pair of eyes please? this isn't trivial

src/api/app/models/event/request_reviews_done.rb
bgeuken: AFAIK, without the self Ruby couldn't tell if you want to define a new variable or assign it to the instance variable.

src/api/app/models/bs_request.rb
```diff
@@ -705,6 +705,15 @@ def cancelapproval(opts)
     approval_handling(nil, opts)
   end
 
+  def calculate_state_from_reviews
+    return :declined if reviews.any?(&:declined?)
```
bgeuken: I would define those as scopes and then call `reviews.declined.exists?` It's more performant since it's all happening in SQL. And it's more speaking.

src/api/app/models/bs_request.rb
```diff
@@ -773,30 +767,34 @@ def change_review_state(new_review_state, opts = {})
...
+        history = HistoryElement::RequestAllReviewsApproved
+        self.comment = 'All reviewers accepted request'
+        save!
+        # pre-approved requests can be processed
+        BsRequestAutoAcceptJob.perform_later(id) if approver
+      elsif new_request_state == :review
+        save!
+      elsif new_request_state == :declined
+        history = HistoryElement::RequestDeclined
+        save!
```
bgeuken: There is one subtle difference now which is that `change_review_state` saves the request. Before it didn't. Did you check that this doesn't have any side effect?

coolo: see line 790 of the original - I just moved it into the if and duplicated it, so the code flow is more obvious

src/api/app/models/event/review_changed.rb
```diff
@@ -0,0 +1,39 @@
...
+    def subject
+      "Request #{payload['number']} was reviewed (#{actions_summary})"
+    end
+
+    def expanded_payload
+      payload_with_diff
+    end
+
+    def custom_headers
+      super.merge(review_headers)
```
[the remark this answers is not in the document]
coolo: well, I didn't consider the extra class worth it for just this function. So I decided as you reviewed :)

So it depends on how much 'a bit' strange you find it, I'll happily introduce another Event::Review now that we have 2.

src/api/spec/models/bs_request_spec.rb
```diff
@@ -138,6 +139,57 @@
...
+    context 'with by_user review' do
+      before do
+        User.current = user
+        request.addreview(by_user: reviewer, comment: 'does it look ok?')
+      end
+
+      it 'raises exception on missing by_ paramter' do
+        expect do
+          request.change_review_state(:accepted)
+        end.to raise_error(BsRequest::InvalidReview)
```
bgeuken: That's more about personal taste, but I would prefer to write this in one line unless it exceeds the limit, like 
` expect { request.change_review_state(:accepted) }.to raise_error(BsRequest::InvalidReview) `

src/api/app/models/bs_request.rb
coolo: But we load the reviews into memory for the `reverse_each` before. So I kind of doubt it's more performant to go back into the DB for the declined ones. Shall we benchmark? :)

src/api/app/models/bs_request.rb
```diff
@@ -714,6 +714,21 @@ def calculate_state_from_reviews
     end
   end
 
+  def find_review_for_opts(opts)
+    reviews.reverse_each do |review|
```
bgeuken: Didn't test it and didn't check the docs, but this should work `reviews.reverse.find { |review| review.reviewable_by?(opts) }` and IIRC it returns nil if no match was found

src/api/app/models/bs_request.rb
```diff
@@ -714,6 +714,21 @@ def calculate_state_from_reviews
...
+      return review if review.reviewable_by?(opts)
+    end
+    nil
+  end
+
+  def supersede_request(history_arguments, superseded_opt)
+    self.state = :superseded
+    self.superseded_by = superseded_opt
+    history_arguments[:description_extension] = superseded_by.to_s
+    save!
```
bgeuken: I am nitpicking, but you could just use [update!](https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update-21) here.

src/api/app/models/bs_request.rb
```diff
@@ -725,73 +740,35 @@ def change_review_state(new_review_state, opts = {})
...
-        review.reviewer = User.current.login
-        review.save!
-        Event::ReviewChanged.create(notify_parameters)
-
-        history = nil
-        history = HistoryElement::ReviewAccepted if new_review_state == :accepted
-        history = HistoryElement::ReviewDeclined if new_review_state == :declined
-        history = HistoryElement::ReviewReopened if new_review_state == :new
-        history.create(review: review, comment: opts[:comment], user_id: User.current.id) if history
-      end
```
bgeuken: It took me quite a while to map all the removed code to it's new location. I hope I did not overlook any... Next time please split such changes into smaller chunks, aka. commits.
And some more details in the commit message would help.

src/api/app/models/bs_request.rb
coolo: argueing in one line about performance and in another suggesting .reverse instead of .reverse_each is somewhat inconsistent :)

But I can do that.

src/api/app/models/event/review_changed.rb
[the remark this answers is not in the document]
bgeuken: Good, then we leave it as is.

src/api/app/models/bs_request.rb
bgeuken: Yes, let's benchmark it

coolo: my pleasure :)

the problem I see is the 2nd select in your solution:

```
  Review Load (0.2ms)  SELECT `reviews`.* FROM `reviews` WHERE `reviews`.`bs_request_id` = 37948
  Review Exists (0.2ms)  SELECT  1 AS one FROM `reviews` WHERE `reviews`.`bs_request_id` = 37948 AND `reviews`.`state` = 'declined' LIMIT 1
```

But let's benchmark it on OBS data:

```
requests=BsRequest.all.to_a; puts Benchmark.measure { requests.each { |r| i=r.reviews.to_a; r.calculate_state_from_reviews_bjoern } }
```

vs
```
requests=BsRequest.all.to_a; puts Benchmark.measure { requests.each { |r| i=r.reviews.to_a; r.calculate_state_from_reviews_coolo } }
```

(obviously this requires a lot of memory in both situations, but we don't care about that - hence I'm loading all requests into memory before the measure to avoid this being a factor.

Now bjoern result: 743.018801  36.429906 779.448707 (946.578434)
Now coolo result: 412.326967  22.553010 434.879977 (510.203396)

An improvement of 46% if prefering ruby over database scope.

I understand that scopes are easier to read - but they are super expensive.

coolo: This discussion and https://github.com/openSUSE/open-build-service/pull/6389#issuecomment-444378744 make me believe that you guys really need to mistrust the database more. Just because it happens in the database doesn't make it fast.

coolo:
@Ana06 your review is outdated, please update

coolo:
anyone? :eyes:

Ana06:
OBS Package Build is failing, fell free to merge it when it works :wink:

coolo:
they fail now as rake disappeared, nothing to worry about - they built fine before

PR #6407, opened 2018-11-30
Refactor method of BsRequestActionGroup

bgeuken:
By moving all parts of the actual query to the SQL level we make this
method faster and solve some n + 1 issues as well.

### DIFF
--- a/src/api/app/models/bs_request_action_group.rb
+++ b/src/api/app/models/bs_request_action_group.rb
@@ -171,11 +171,8 @@ class BsRequestActionGroup < BsRequestAction
   end
 
   def find_review_state_of_group
-    bs_requests.each do |req|
-      req.reviews.each do |rev|
-        return :review if rev.state != :accepted
-      end
-    end
+    return :review if bs_requests.joins(:reviews).where.not('reviews.state': :accepted).exists?
+
     :new
   end
 


bgeuken:
We could also move this to a scope, but so far we only use this once. So I didn't do that.

coolo:
@adrianschroeter @hennevogel what do you think about just dropping this feature?

It was an [experiment](https://github.com/SUSE/hackweek/wiki/Improved-Factory-devel-project-submission-workflow) that failed, but led to the superior concept of staging projects, that sticked.

The hackweek was in August 2013, the last group request was created in January 2014.

src/api/app/models/bs_request_action_group.rb
```diff
@@ -171,11 +171,8 @@ def removerequest(opts)
   end
 
   def find_review_state_of_group
-    bs_requests.each do |req|
-      req.reviews.each do |rev|
-        return :review if rev.state != :accepted
-      end
-    end
+    return :review if bs_requests.joins(:reviews).where.not('reviews.state': :accepted).exists?
```
Ana06: or even:

`bs_requests.joins(:reviews).where.not('reviews.state': :accepted).exists? ? :review : :new`

bgeuken:
Closing. We have #6423 now for discussing whether we want to keep this feature or not.

PR #6410, opened 2018-11-30
Generalize DataTable inizalization so that it can be reused

Ana06:
This help us fixing the bug with input search field not being responsible in the whole application. I will add this to our pattern library and fix it in the rest of the code after this get merge. I don't do it in the same PR to help that this is merge fast and can be use in the new code.

### before
image

### after

image


src/api/app/views/webui2/webui/user/_involvement.html.haml
```diff
@@ -71,8 +71,5 @@
                     = link_to(project_name, project_show_path(project: project_name))
 
 = content_for :ready_function do
-  :plain
-    $('#involved-packages-table, #involved-projects-table, #owned-table').DataTable({
-      language: { search: '', searchPlaceholder: "Search..." }
-    });
+  initializeDataTable('#involved-packages-table, #involved-projects-table, #owned-table');
```
dmarcoux: The function has 2 parameters, but 3 are passed here. I guess this is the issue.

Ana06: no, I am passing one and the second is `null`... I'll :mag:

dmarcoux: Oh, I missed the quote, sorry!

src/api/app/views/webui2/webui/user/_involvement.html.haml
```diff
@@ -21,7 +21,7 @@
           %p.md-3
             #{user.name} is not involved in any packages
         - else
-          %table.table.table-striped.table-bordered#involved-packages-table
+          %table.responsible.table.table-striped.table-bordered#involved-packages-table
```
dmarcoux: `responsive` ??

Ana06: yes :see_no_evil:

Ana06: but if I write the correct one it also breaks. I think we have to remove the responsive thing. It has more bugs than what is doing correctly

hellcp: https://getbootstrap.com/docs/4.0/content/tables/#responsive-tables

dmarcoux: @hellcp It needs to be on the [container of the table](https://getbootstrap.com/docs/4.1/content/tables/#responsive-tables). But we don't use it in this case as the table doesn't overflow (so we don't need the scrollbar)

hellcp: @dmarcoux true, I didn't even look what kind of table that was :P

PR #6416, opened 2018-11-30
Add breakpoint for very larger screens

### LINKED ISSUE #6033: [beta] new UI can be too narrow
# Issue Description

Just like with GitHub, enforcing a maximum width on the main content of each page in the layout can cause things to be too squashed and waste a lot of valuable screen space.

How to Reproduce
===============
1. Join the beta program
2. Go to https://build.opensuse.org/package/show/systemsmanagement:sumaform:opensuse423/cloud-init
3. See something like this:

obs-width

4. Notice how certain lines are needlessly wrapped, and how much wasted space there is.

Expected Result
===============

Tweaking the width gives a much nicer result:

obs-width2

However, everyone's tastes / browsers / displays are different, so this should be adjustable.

bgeuken:
Kudos for providing the breakpoints goes to @hellcp

Fixes #6033

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/breakpoints.scss
+++ b/src/api/app/assets/stylesheets/webui2/breakpoints.scss
@@ -1,0 +1,5 @@
+@media (min-width: 1600px) {
+  .container {
+    max-width:1540px
+  }
+}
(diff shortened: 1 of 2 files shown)


src/api/app/assets/stylesheets/webui2/breakpoints.scss
```diff
@@ -0,0 +1,5 @@
+@media (min-width: 1600px) {
+ .container {
```
dmarcoux: The indentation should be 2 spaces.

hennevogel:
Did anybody check if this breaks pages? It isn't called **break**point without a reason :smile_cat:

PR #6417, opened 2018-11-30
Reduce padding in user and group table

### LINKED ISSUE #6049: [beta] feedback: excessive UI padding in Users tab (and maybe others)
In the new UI, the padding for .td seems excessive. You already have 150% line-heights, and then there is another 150%rem of padding (top 75% and bottom 75%). This wastes a lot of space and makes it more difficult to read. I would suggest having TH with current, extra padding, but reducing padding of all TD, like it seems it was done in the file list.

now

vs. 0 padding,

0padding

Maybe reducing padding to 25%?

PS. Thank you for the redesign. Overall, the UI feels a lot more modern. :+1:

DavidKang:
This should fix #6049.

### Before
screenshot_2018-11-30 users of test project home admin - open build service 1

### After
screenshot_2018-11-30 users of test project home admin - open build service

### DIFF
--- a/src/api/app/views/webui2/shared/user_or_groups_roles/_index.html.haml
+++ b/src/api/app/views/webui2/shared/user_or_groups_roles/_index.html.haml
@@ -15,7 +15,7 @@
 #involved-users{ data: data }
   - if users.present?
     %h3 User Roles
-    %table.responsive.table.table-striped.table-bordered.w-100#user-table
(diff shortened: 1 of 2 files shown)


DavidKang:
@dmarcoux could you review it again :smile_cat: 

screenshot_2018-11-30 users of test project home admin - open build service 2

PR #6422, opened 2018-11-30
Move meta related actions from ProjectController to its own CRUD controller

vpereira:
Move `save_meta` and `meta` from `ProjectController` to `Projects::MetaController`

- [X] write new code
- [X] adapt the rules to point to the new code
- [X] change the views
- [X] rewrite tests
- [x] cassettes
- [x] delete old code from `ProjectController`
- [x] refactor the `Projects::MetaController#update` action


src/api/app/views/webui2/webui/projects/meta/_breadcrumb_items.html.haml
```diff
@@ -0,0 +1,24 @@
+= render partial: 'webui/main/breadcrumb_items'
```
dmarcoux: These are the breadcrumbs from project, so it needs to be `= render partial: 'webui/project/breadcrumb_items'`

src/api/spec/controllers/webui/projects/meta_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    it { expect(response).to have_http_status(:success) }
+  end
+
+  describe 'POST #update' do
+    before do
+      login user
+    end
+
+    context 'with an nonexistent project' do
```
dmarcoux: with **a** nonexistent

src/api/app/views/webui2/webui/projects/meta/_breadcrumb_items.html.haml
```diff
@@ -0,0 +1,24 @@
+= render partial: 'webui/project/breadcrumb_items'
```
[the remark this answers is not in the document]
dmarcoux: The meta part should still be here though.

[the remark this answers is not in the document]
vpereira: meta3

it is there, no?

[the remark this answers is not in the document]
dmarcoux: I meant that it should be in this file, instead of being in: https://github.com/openSUSE/open-build-service/blob/b8a5640a4aecc95142e95803ff885b92e85cb7d9/src/api/app/views/webui2/webui/project/_breadcrumb_items.html.haml#L22-L24

src/api/app/controllers/webui/projects/meta_controller.rb
```diff
@@ -0,0 +1,64 @@
+module Webui
+  module Projects
+    class MetaController < WebuiController
+      require_dependency 'opensuse/validator'
+      before_action :set_project, only: [:show]
+      before_action :set_project_by_name, only: [:update]
+      after_action :verify_authorized, only: [:update]
+      def show
```
ChrisBr: Empty line missing

src/api/app/controllers/webui/projects/meta_controller.rb
```diff
@@ -0,0 +1,64 @@
+module Webui
+  module Projects
+    class MetaController < WebuiController
+      require_dependency 'opensuse/validator'
+      before_action :set_project, only: [:show]
+      before_action :set_project_by_name, only: [:update]
```
ChrisBr: Hm, what is the main difference? That set_project_by_name also respects remote projects, right? Maybe we should reflect that in the name? Because otherwise it is not clear why there are two different before filters...

src/api/app/controllers/webui/projects/meta_controller.rb
vpereira: WIP

src/api/app/controllers/webui/projects/meta_controller.rb
vpereira: @ChrisBr please suggest a better name.

vpereira: ```set_project_by_name``` is the same name used in the ```ProjectController``` so to keep the same name makes easier to identify a possibility to refactor it to use the same method in both. If you have a better name, please share with me.

Ana06: I think it is not about the name, but about having two actions that seem to be doing the same :thinking:

src/api/app/controllers/webui/projects/meta_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      def update
+        authorize @project, :update?
+        errors = []
+        errors = validate_request_and_set_error(errors)
+        errors = run_update(errors) if errors.empty?
+        status = if errors.empty?
+                   flash.now[:success] = 'Config successfully saved!'
+                   200
+                 else
+                   flash.now[:error] = errors.compact.join("\n")
```
Ana06: `compact` here is not needed

src/api/app/controllers/webui/projects/meta_controller.rb
ChrisBr: I just tried it and it seems to work with ``set_project`` as well. Which seems to be logical because why should ``show`` only work for local projects and ``update`` works for local and remote that doesn't make sense. So please change to:

```
before_action :set_project
```

And drop the ``only`` and ``set_project_by_name`` part.

src/api/app/services/meta_controller_service/meta_validator.rb
```diff
@@ -0,0 +1,22 @@
+module MetaControllerService
+  class MetaValidator
+    require_dependency 'opensuse/validator'
+
+    attr_reader :meta, :request_data, :errors
+
+    def initialize(params = {})
+      @meta = params[:meta] if params.key?(:meta)
```
dmarcoux: The `if` is not needed because if the key `meta` is not present in the `params` hash, it will return `nil`, which what the value of the instance variable `@meta` already is.

src/api/app/services/meta_controller_service/project_updater.rb
```diff
@@ -0,0 +1,21 @@
+module MetaControllerService
+  class ProjectUpdater
+    attr_reader :errors
+
+    def initialize(params = {})
+      @project = params[:project] if params.key?(:project)
```
dmarcoux: `if` unneeded again, same comment as above.

src/api/app/services/meta_controller_service/request_validator.rb
```diff
@@ -0,0 +1,24 @@
+module MetaControllerService
+  class RequestValidator
+    attr_reader :project, :request_data, :errors
+    def initialize(params = {})
+      @project = params[:project] if params.key?(:project)
```
dmarcoux: `if` unneeded again, same comment as above.

ChrisBr:
I very much like the approach of introducing service classes :heart: :+1:  But I would have some suggestions:

* I would prefer to not have dedicated helper methods but do everything in the action (at least for ``validate_request_and_set_error``). We could do the request validation in the ``ProjectUpdater`` (composition!)
* [nitpick] I would prefer to name the method on the classes ``validate`` and ``update`` instead of ``call``
*  Maybe we can have a helper method ``valid?`` instead of calling ``empty?``. I think especially when using the composition approach this does make sense.
* The name ``RequestValidator`` does not make much sense in this context, what about ``MetaValidator`` and the other one ``MetaXmlUpdater``

I imagine it could look something like this:

```ruby
module MetaControllerService
   class ProjectUpdater
     attr_reader :errors

     def initialize(project: nil, request_data = {}, validator = ::MetaControllerService::RequestValidator)
       @project = project
       @request_data = request_data
       @validator = validator.new(project: project, request_data: request_data)
     end

     def update
       return unless validator.valid?

       Project.transaction do
         errors = project.update_from_xml(request_data)[:error]
         project.store unless errors?
       end
     end

     def valid?
       return validator.errors unless validator.valid?
       errors.present?
     end

    private
    attr_writer :errors
    attr_reader :project, :request_data
   end
 end
```

```ruby
def update
  authorize @project, :update?
  updater = ::MetaControllerService::ProjectUpdater.new(project: @project, request_data: @request_data)
  updater.update  

  if updater.valid?
    flash.now[:success] = 'Config successfully saved!'
    200
  else
    flash.now[:error] = updater.errors.join("\n")
    400
  end
  switch_to_webui2
  render layout: false, status: status, partial: "layouts/#{view_namespace}/flash", object: flash
```

What do you think? Anyway, really great work :+1:

vpereira:
> I very much like the approach of introducing service classes   But I would have some suggestions:

Thank you for the extended review. Code examples helps me a lot! <3

One issue that I have doing like that is breaking the Single Responsibility principle, coupling again the classes. In another hand the ```ProjectUpdater``` would be a good place to make it. I will have a look, thank you!

ChrisBr:
> One issue that I have doing like that is breaking the Single Responsibility principle, coupling again the classes.

I don't see how this is violation Single Responsibility as this is a classic example of composition. Object orientation is all about sending messages between objects. As long as we only rely on the interface of the validator (``validate`` method) and pass in the validator as an instance variable, this does not violate Single Responsibility. On the contrary, this ensures that both ``ProjectUpdater`` and ``Validator`` keep their responsibility. And keeping the validator as instance variable ensures that you can also test in isolation like:

```ruby
  it 'is valid' do
    validator = double("validator", valid: true)
    updated = ProjectUpdater.new("foo", "bar" , validator)
    ...
  end

  it 'is invalid' do
    validator = double("validator", valid: false)
    updated = ProjectUpdater.new("foo", "bar" , validator)
    ...
  end
```

But I see your concern and to make this more clear, you could introduce an intermediate Service class which takes an updater and validator and composes them together like:

```ruby
class MetaService
  def initialize(updated = ProjectUpdated, validator = MetaValidator)
     # set instance variables
  end

  def save
    validator.valid?
    updated.update
    # etc etc.
  end
```

But honestly, I think this would be overkill and the initial approach with the two classes is sufficient for now.

This is just pseudo code to visualize my thoughts and is not intended to work right away.

vpereira:
please check it @ChrisBr

src/api/app/services/meta_controller_service/request_validator.rb
```diff
@@ -0,0 +1,25 @@
...
+  class RequestValidator
+    attr_reader :project, :request_data, :errors
+
+    def initialize(params = {})
+      @project = params[:project]
+      @request_data = params[:request_data]
+      @errors = []
+    end
+
+    def call
```
ChrisBr: See my other comment. I would prefer to change this method to a private ``validate`` method:

```ruby
class RequestValidator
  def valid?
    validate
    @errors.empty?
  end

  private

  def validate
    # ... the validation code from previous call method
  end
```

This has some advantages:
* keeping the interface as small as possible (one method less in the public interface)
* don't need to remember to call ``call`` before calling ``valid?``
* keeps the same behaviour than ActiveRecord

src/api/app/services/meta_controller_service/project_updater.rb
```diff
@@ -0,0 +1,30 @@
+module MetaControllerService
+  class ProjectUpdater
+    def initialize(project: nil, request_data: {}, validator_klass: ::MetaControllerService::RequestValidator)
+      @project = project
+      @request_data = request_data
+      @validator = validator_klass.new(project: project, request_data: request_data)
+    end
+
+    def call
+      @validator.call
```
ChrisBr: I would prefer to get rid of this additional ``call`` call. I think this should behave similar to ActiveRecord which also does not require an additional method call before calling ``valid?``. See my other comment.

src/api/app/services/meta_controller_service/project_updater.rb
```diff
@@ -0,0 +1,30 @@
+module MetaControllerService
+  class ProjectUpdater
+    def initialize(project: nil, request_data: {}, validator_klass: ::MetaControllerService::RequestValidator)
+      @project = project
+      @request_data = request_data
+      @validator = validator_klass.new(project: project, request_data: request_data)
+    end
+
+    def call
```
ChrisBr: [nitpick] I would prefer to use ``save`` here to keep same interface as ActiveRecord.

src/api/app/services/meta_controller_service/request_validator.rb
```diff
@@ -0,0 +1,25 @@
+module MetaControllerService
+  class RequestValidator
```
ChrisBr: I still don't think that the class name reflects what this class is doing: ``MetaValidator`` would make it more clear I think.

src/api/app/services/meta_controller_service/meta_validator.rb
```diff
@@ -0,0 +1,22 @@
+module MetaControllerService
+  class MetaValidator
```
ChrisBr: I don't think the class name reflects what this class is doing, it does not validate the Meta but the XML, so ``MetaXmlValidator`` might be a better name.

src/api/app/services/meta_controller_service/project_updater.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+
+      Project.transaction do
+        @errors = @project.update_from_xml(@request_data)[:error]
+        @project.store if valid?
+      end
+    end
+
+    def errors
+      @errors.is_a?(Array) ? @errors.join("\n") : @errors
```
ChrisBr: Why not always having an array? Just instantiate the error instance variable with an empty array and use ``<<`` instead.

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -14,15 +14,15 @@ class Webui::ProjectController < Webui::WebuiController
                                      :new_package, :new_package_branch, :incident_request_dialog, :release_request_dialog,
                                      :show, :linking_projects, :add_person, :add_group, :buildresult, :delete_dialog,
                                      :destroy, :remove_path_from_target, :rebuild_time, :packages_simple,
-                                     :requests, :save, :monitor, :toggle_watch, :meta,
+                                     :requests, :save, :monitor, :toggle_watch,
```
bgeuken: This looks like a cleanup of the last commit. Please squash them together

vpereira: it was a required part of my changes, yes. I squashed now the commits.

src/api/app/services/meta_controller_service/project_updater.rb
vpereira: we can hide this implementation detail out of the controller. So if its ok I would keep it where it is.

PR #6423, opened 2018-12-01
Drop grouped requests

coolo:
This was available for some time during OBS 2.5 development (autum 2013),
but got dropped again before release. It was superseded by staging projects
in early 2014


coolo:
Let's delete the requests from the OBS database - I'm very confident no other instance has these, as the api to generate them was never documented :)

Conan-Kudo:
Wait, we had the ability to have grouped SRs? Why was this never published feature?

That would have been useful for things like submitting swathes of packages as one SR, rather than hoping everything "came together" at once...

coolo:
you can submit multiple packages in one request - it's not a 'SR' then, but rather a multi action request.

The feature of grouping requests is that you change the group later on - which is not possible with MARs. See https://github.com/SUSE/hackweek/wiki/Improved-Factory-devel-project-submission-workflow and https://github.com/openSUSE/open-build-service/pull/6358 for the feature replacing it.

PR #6425, opened 2018-12-02
Don't waste time with expensive default scopes

coolo:
I happened to notice that / loads in 4 seconds on
IBS and looking, 99% is spent in the '6 last updated
packages' box because the default query limits
package names to .* and project names to .*
and times that are after 1970

So just express the default filters by lack of it instead
of creating artiticial ones


src/api/lib/statistics_calculations.rb
```diff
@@ -19,9 +21,12 @@ def self.packages(limit, timelimit, prj_filter, pkg_filter)
...
 
   def self.projects(limit, timelimit, prj_filter)
-    Project.where(updated_at: timelimit..Time.now).
-      where('name REGEXP ?', prj_filter).
-      order('updated_at DESC').
+    projects = Project.all
+    projects = projects.where(updated_at: timelimit..Time.now) if timelimit
+    if prj_filter
+      projects = projects.where('name REGEXP ?', prj_filter)
+    end
```
bgeuken: Nitpick: I would make this a guard clause, given that you already use one one line above.

src/api/lib/statistics_calculations.rb
```diff
@@ -19,9 +21,12 @@ def self.packages(limit, timelimit, prj_filter, pkg_filter)
...
+    projects.order('updated_at DESC').
       limit(limit).
       pluck(:name, :updated_at).
       map { |name, at| [at, name, :project] }
```
bgeuken: Nitpicking, and nothing you touched anyway... but this could be a ``map!`.

coolo: hmm, it returns that value - so I think map is perfectly fine

src/api/lib/statistics_calculations.rb
coolo: done

src/api/lib/statistics_calculations.rb
bgeuken: both do: https://ruby-doc.org/core-2.5.0/Array.html#method-i-map-21

coolo: ah, so it's not creating a new array

coolo: This is a much bigger story -> https://github.com/openSUSE/open-build-service/pull/6429 

(and I mentioned you in the commit log as revenche :)

PR #6427, opened 2018-12-02
Update grid for build results

bgeuken:
The build results for each repository are indented to
visually distinguish them from the repository they belong to.
Before we were doing this by adding an empty column. Instead we
now use bootstrap's grid column offset.


bgeuken:
For some reason the GitHub checks don't get updated, but the test suite passed: https://circleci.com/workflow-run/c571de5d-a4c7-4a5a-adf1-10b3217c31a9

PR #6428, opened 2018-12-02
Fix icon sizes and breakpoints of main page

bgeuken:
The icons were to large in proportion to the content they
belonged to. Reducing their size also required to drop some
tweaks done via padding and column length.
As a side effect the icons no longer overlap with other elements
for certain screen sizes.

Before

screenshot_2018-12-03 welcome - opensuse build service

After

screenshot_2018-12-02 welcome - open build service


bgeuken:
@dmarcoux Nice catch:thumbsup: I fixed it now. Please have another look

PR #6429, opened 2018-12-03
Use inline array functions for returned arrays

coolo:
Inspired by @bgeuken's comment:
 https://github.com/openSUSE/open-build-service/pull/6425#discussion_r238106145

The performance gain is most likely unmeasurable but it's still
a good default not to waste another array allocation

### DIFF
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -433,7 +433,7 @@ class Webui::ProjectController < Webui::WebuiController
       @localpackages[package.name] = 1
     end
 
-    @packagenames = @packagenames.flatten.uniq.sort
+    @packagenames = @packagenames.flatten.uniq.sort!
 
     ## Filter for PackageNames ####
     @packagenames.select! { |name| filter_matches?(name, @name_filter) } if @name_filter.present?
@@ -745,7 +745,7 @@ class Webui::ProjectController < Webui::WebuiController
     @linking_projects = @project.linked_by_projects.pluck(:name)
 
     reqs = @project.open_requests
-    @requests = (reqs[:reviews] + reqs[:targets] + reqs[:incidents] + reqs[:maintenance_release]).sort.uniq
+    @requests = (reqs[:reviews] + reqs[:targets] + reqs[:incidents] + reqs[:maintenance_release]).sort!.uniq
 
     @nr_of_problem_packages = @project.number_of_build_problems
   end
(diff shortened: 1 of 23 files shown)


src/api/app/controllers/webui/project_controller.rb
```diff
@@ -779,7 +779,7 @@ def find_maintenance_infos
 
     @is_maintenance_project = @project.is_maintenance?
     if @is_maintenance_project
-      @open_maintenance_incidents = @project.maintenance_incidents.pluck('projects.name').sort.uniq
+      @open_maintenance_incidents = @project.maintenance_incidents.pluck('projects.name').sort!.uniq!
```
bgeuken: There are some cases were the 'bang' version is returning nil when the object wasn't changed, [uniq](https://ruby-doc.org/core-2.5.0/Array.html#method-i-uniq-21) is one of them.

coolo: yeah, mean pitfall - will have to revert some cases :)

src/api/app/jobs/consistency_check_job.rb
```diff
@@ -234,24 +234,24 @@ def dir_to_array(xmlhash)
...
       array << e['name']
     end
-    array.sort
+    array.sort!
   end
 
   def hash_diff(a, b)
     # ignore the order inside of the hash
-    (a.keys.sort | b.keys.sort).each_with_object({}) do |diff, k|
+    (a.keys.sort! | b.keys.sort!).each_with_object({}) do |diff, k|
```
Ana06: [nitpick] I wonder if this is what we really want here, or this: `(a.keys | b.keys).sort!`

src/api/app/models/project.rb
```diff
@@ -1587,8 +1587,8 @@ def self.validate_repository_xml_attribute(request_data, project_name)
   end
 
   def get_removed_repositories(request_data)
-    new_repositories = request_data.elements('repository').map(&:values).flatten
-    old_repositories = repositories.all.map(&:name)
+    new_repositories = request_data.elements('repository').map!(&:values).flatten
+    old_repositories = repositories.map(&:name)
```
Ana06: why not `!` here?

coolo: This actually sounds like a job for pluck

coolo: But to answer your question:
```
undefined method `map!' for #<Repository::ActiveRecord_Associations_CollectionProxy:0x00000000085d93c0>

src/api/app/jobs/consistency_check_job.rb
coolo: I agree.

PR #6433, opened 2018-12-03
Improve User model efficiency - select

Ana06:
Use `select` instead of `pluck` as it generate a subquery with a select instead of adding all the ids to que query. Potentially there could be a lot of ids, which makes this query inefficient. :bowtie:

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -723,7 +723,7 @@ class User < ApplicationRecord
   def involved_patchinfos
     array = []
 
-    ids = PackageIssue.open_issues_of_owner(id).with_patchinfo.distinct.pluck(:package_id)
+    ids = PackageIssue.open_issues_of_owner(id).with_patchinfo.distinct.select(:package_id)
 
     Package.where(id: ids).find_each do |p|
       hash = { package: { project: p.project.name, name: p.name } }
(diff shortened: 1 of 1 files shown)


src/api/app/models/user.rb
```diff
@@ -752,7 +752,7 @@ def involved_patchinfos
   def user_relevant_packages_for_status
     role_id = Role.hashed['maintainer'].id
     # First fetch the project ids
-    projects_ids = involved_projects.pluck(:id)
+    projects_ids = involved_projects.select(:id)
     packages = Package.joins("LEFT OUTER JOIN relationships ON (relationships.package_id = packages.id AND relationships.role_id = #{role_id})")
     # No maintainers
     packages = packages.where([
```
coolo: as you're touching the function anyway, I would move the final pluck to the calling function (there is only one). As this function is user_relevant_packages_for_status (for whatever the user_ prefix is good for in User model :) and not suggesting it's returning ids.

Ana06: I agree, but that has nothing to  do with efficiency. :confused:

bgeuken: I would agree with @Ana06, but I know to many occasions where you demand similar additions from others. I am torn:laughing:

Ana06: `Demand != suggest`

I find great @coolo's suggestion but I will implement it in a different PR. It is out of the scope.

PR #6441, opened 2018-12-03
Implement InfluxDB BackendSubscriber

ChrisBr:
to send performance metrics of the Backend to our InfluxDB.

### DIFF
--- a/src/api/config/initializers/influxdb_custom_subscriber.rb
+++ b/src/api/config/initializers/influxdb_custom_subscriber.rb
@@ -1,0 +1,3 @@
+require_relative '../../lib/influxdb_obs/obs/middleware/backend_subscriber'
+
+ActiveSupport::Notifications.subscribe('obs.backend.process_response', InfluxDB::OBS::Middleware::BackendSubscriber.new('rails.backend', Rails.logger))
--- a/src/api/lib/backend/logger.rb
+++ b/src/api/lib/backend/logger.rb
@@ -1,3 +1,5 @@
+require_relative '../influxdb_obs/obs/normalizer/location_normalizer'
+
 module Backend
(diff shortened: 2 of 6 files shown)


src/api/config/initializers/influxdb_custom_subscriber.rb
```diff
@@ -0,0 +1,4 @@
+if CONFIG['influxdb_hosts'].present?
+  require Rails.root.join('lib', 'influxdb_obs', 'influxdb_obs').to_s
```
hennevogel: All files in `lib/` are autoloaded...

ChrisBr: Ah no? All code in app is autoloaded, lib is not autoloaded and either needs to be specified in application.rb or in the initializer.

https://stackoverflow.com/a/40019108

ChrisBr:
@hennevogel updated

src/api/config/initializers/influxdb_custom_subscriber.rb
```diff
@@ -0,0 +1,2 @@
+require_relative File.join('..', '..', 'lib', 'influxdb_obs', 'influxdb_obs')
```
hennevogel: That one eats strings....

src/api/lib/backend/backend.rb
```diff
@@ -12,6 +12,7 @@ module Api
 
 require 'benchmark'
 require 'api_exception'
+require_relative File.join('..', 'influxdb_obs', 'obs', 'normalizer', 'location_normalizer')
```
hennevogel: Why do we requires this here and not in the logger?

ChrisBr:
coveralls hanging ...

PR #6456, opened 2018-12-04
Fix project_destroy_path usage

### LINKED ISSUE #6449: [bootstrap] Deleting remote project does not work
In the new bootstrap view, deleting a remote project does crash with

```
No route matches [DELETE] "/project/destroy.openSUSE.org"
```

As you can see, it does create a wrong route `` "/project/destroy.openSUSE.org"``. Deleting normal projects works like expected.

dmarcoux:
The project_destroy_path doesn't have parameters, so passing the `project` variable isn't needed. This issue prevented the deletion of remote projects.

Fixes #6449


dmarcoux:
coveralls is stuck... merging

Issue #6464, opened 2018-12-04
[bootstrap] User request page


ChrisBr:
![selection_020](https://user-images.githubusercontent.com/3799140/49453578-b9b3b800-f7e3-11e8-9bf4-95440db79591.png)

No bugs but some small improvements:
* The number in the tab of open requests could be a label instead of parantheses
* I think default number of 25 requests per table might be too big as a table row has more space which required more spacing than before (see screenshot 25 fitted on one page).

Before

![selection_021](https://user-images.githubusercontent.com/3799140/49453680-e9fb5680-f7e3-11e8-83f8-3cbd3190a73e.png)

bgeuken:
Nice idea:smiley:

bgeuken:
I wrote a patch that would set the default on this page to 10, but now I have some doubts. Reducing this here will mostly affect power users and requires them to increase the requests shown per page whenever they access it. (Assuming they don't already do that by setting it from 25 to 50 or 100).

bgeuken:
We could free some space by moving the refresh icon somewhere else.

coolo:
I don't think 10 or 25 makes much of a difference. You normally don't work in this view, but get an overview here.

PR #6465, opened 2018-12-04
Style the request count on the user's tasks page with a badge

bgeuken:
This makes the page look nicer and makes the counter more visible for
users.

Part of #6464

Before

screenshot_2018-12-04 tasks for bgeuken - opensuse build service

After

screenshot_2018-12-04 tasks for admin - open build service


dmarcoux:
I think it would be nice to have this in the patterns library.

PR #6467, opened 2018-12-05
Fix 500 error when a user owns a project, but not a package

bgeuken:
When a user owns a project the owner search would return a touple of
[nil, 'some_project']. Our code was not checking for non-existant
packages and thus caused an error.

### DIFF
--- a/src/api/app/views/webui2/webui/user/_involvement.html.haml
+++ b/src/api/app/views/webui2/webui/user/_involvement.html.haml
@@ -66,7 +66,8 @@
               - owned.each do |package_name, project_name|
                 %tr
                   %td
-                    = link_to(package_name, package_show_path(package: package_name, project: project_name))
+                    - if package_name
+                      = link_to(package_name, package_show_path(package: package_name, project: project_name))
                   %td
                     = link_to(project_name, project_show_path(project: project_name))
 


bgeuken:
I am going to provide a test for this, but since this is breaking the user page (in bootstrap) I think we should patch this first.

src/api/app/views/webui2/webui/user/_involvement.html.haml
```diff
@@ -66,7 +66,8 @@
               - owned.each do |package_name, project_name|
                 %tr
                   %td
-                    = link_to(package_name, package_show_path(package: package_name, project: project_name))
+                    - if package_name
+                      = link_to(package_name, package_show_path(package: package_name, project: project_name))
```
coolo: https://api.rubyonrails.org/v5.2.1/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to_if anyone? :)

bgeuken: Good idea:thumbsup:

bgeuken: Either I am doing something wrong or `link_to_if` doesn't like path helpers. At least not if the condition is falsy.

After checking the [code](https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/actionview/lib/action_view/helpers/url_helper.rb#L439) I also noticed that this is not exactly the same. With `link_to_if` we would still render the `package_name` variable.

I am a bit in a hurry right now. I'll double check later on.

coolo: ah, I see. Sorry for the wrong lead

coolo:
So now coveralls *and* codecov hang? :(

https://coveralls.io/jobs/42909545 -> the upload works

bgeuken:
Regarding tests... I thought it would be enough to write a controller test and trigger the `render_views` flag. But the view code that caused the failure is in a separate tab. Testing this would require a feature test, or moving the code to a helper.

Given that this is just an edge case we (@ChrisBr and me) decided that it's not worth the effort.

=> merging

coolo:
well, not crashing on my account - even if not activating the tab - would be win :)

bgeuken:
If you mean that it started to crash for you again... Someone removed the monkey patch. I have to talk with our demolition squad^^

coolo:
No, I mean a smoke (controller) test would be already a win for myself. But let's assume that noone cares about this tab :)

PR #6468, opened 2018-12-05
Reduce the default number of requests shown on the user tasks page

### LINKED ISSUE #6464: [bootstrap] User request page

bgeuken:
In the new bootstrap UI the request table takes more space. Hence it's
better show less requests by default.

Fixes #6464


bgeuken:
@ChrisBr Please review

PR #6469, opened 2018-12-05
Update influxdb-rails gem

ChrisBr:
We need to use the beta branch because it is fixing an issue which
currently kills our InfluxDB server in production (adding the
path element as tag causes millions of tags which need to get
indexed).

### DIFF
--- a/src/api/config/initializers/influxdb_rails.rb
+++ b/src/api/config/initializers/influxdb_rails.rb
@@ -6,4 +6,6 @@ InfluxDB::Rails.configure do |config|
   config.influxdb_port     = CONFIG['influxdb_port']
   config.retry             = CONFIG['influxdb_retry']
   config.use_ssl           = CONFIG['influxdb_ssl']
+
+  config.time_precision = CONFIG['influxdb_time_precision']
 end
--- a/src/api/Gemfile
+++ b/src/api/Gemfile
@@ -77,7 +77,7 @@ gem 'addressable'
 # for XML builder
 gem 'builder'
 # to write the rails metrics directly into InfluxDB.
-gem 'influxdb-rails', '>=1.0.0.beta1'
+gem 'influxdb-rails', '>=1.0.0.beta2'
 
 group :development, :production do
   # to have the delayed job daemon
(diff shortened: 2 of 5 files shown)


src/api/config/initializers/influxdb_rails.rb
```diff
@@ -6,4 +6,6 @@
   config.influxdb_port     = CONFIG['influxdb_port']
   config.retry             = CONFIG['influxdb_retry']
   config.use_ssl           = CONFIG['influxdb_ssl']
+
+  config.time_precision = 'ms'
```
hennevogel: Why isn't this configurable?

ChrisBr: Changed!

ChrisBr:
We should wait until https://github.com/influxdata/influxdb-rails/pull/58 is merged!

src/api/config/initializers/influxdb_rails.rb
```diff
@@ -1,9 +1,11 @@
...
-  config.use_ssl           = CONFIG['influxdb_ssl']
+  config.influxdb_database   = CONFIG['influxdb_database']
+  config.influxdb_username   = CONFIG['influxdb_username']
+  config.influxdb_password   = CONFIG['influxdb_password']
+  config.influxdb_hosts      = CONFIG['influxdb_hosts'] || [] # default is otherwise localhost
+  config.influxdb_port       = CONFIG['influxdb_port']
+  config.retry               = CONFIG['influxdb_retry']
+  config.use_ssl             = CONFIG['influxdb_ssl']
+  config.time_precision      = CONFIG['influxdb_time_precision']
+  config.series_name_for_sql = CONFIG['influxdb_series_name_for_sql']
```
hennevogel: Why does this have to be configurable?

hennevogel: And what about defaults for all of those?

ChrisBr: All other ones have defaults except ``influxdb_series_name_for_sql`` because it is disabled by default.

hennevogel: k

ChrisBr:
I dropped a few commits because CircleCI was always timing out...

PR #6471, opened 2018-12-05
Migrate repository page to Boostrap

Ana06:
Design of the page has been changed:

- Show repositories in a nicer way, using the space more efficiently
- Integrated expert mode in the page.
- Integrate addition of additional paths the page.

image

TODO (to be done in a different PR):

- Add DoD Repository
- Add from a distribution page
- Tests
- Improve Javascript

Co-authored-by: @DavidKang and @Ana06

### DIFF
--- a/src/api/app/controllers/webui/repositories_controller.rb
+++ b/src/api/app/controllers/webui/repositories_controller.rb
@@ -18,6 +18,15 @@ class Webui::RepositoriesController < Webui::WebuiController
     @user_can_set_flags = policy(@project).update?
 
     switch_to_webui2 if @package
+
+    # TODO: Remove the `return unless` and the flash once this should be available to all beta users on production
+    return unless User.current.try(:in_beta?) && Rails.env.development?
+
+    flash[:notice] = 'We are currently migrating the project pages to Bootstrap. This page is only seen on the development environment.'
+
+    @repositories = @project.repositories.includes(:path_elements, :download_repositories)
+
+    switch_to_webui2
   end
 
   # GET project/add_repository/:project
--- a/src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
+++ b/src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
@@ -1,15 +1,21 @@
+- number_paths = 3
 .card-body
   - if repository.path_elements.present?
     Repository paths:
     %ul.list-unstyled.ml-3
-      - repository.path_elements.each do |path|
-        %li
-          %small.font-weight-bold
-            - if path.link.project == 'deleted'
-              %i.fas.fa-exclamation-circle.text-warning
-              Target repository has been removed
-            - else
-              #{path.link.project}/#{path.link.name}
+      - repository.path_elements.each_with_index do |path, index|
+        - if index < number_paths
+          = render partial: 'repository_path_item', locals: { project: project, repository: repository, path: path }
+        - else
+          - content_for :repository_path_item do
+            = render partial: 'repository_path_item', locals: { project: project, repository: repository, path: path }
+      - if repository.path_elements.size > number_paths
+        %details.repository-path
+          %summary.small
+            %span.more More
+            %span.less Less
+          = yield :repository_path_item
+
 .card-footer.text-center
   .row
     - if User.current.can_modify?(project) && !repository.is_dod_repository?
--- a/src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
+++ b/src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
@@ -1,0 +1,43 @@
+.card-body
+  - if repository.path_elements.present?
+    Repository paths:
+    %ul.list-unstyled.ml-3
+      - repository.path_elements.each do |path|
+        %li
+          %small.font-weight-bold
+            - if path.link.project == 'deleted'
+              %i.fas.fa-exclamation-circle.text-warning
+              Target repository has been removed
+            - else
+              #{path.link.project}/#{path.link.name}
+.card-footer.text-center
+  .row
+    - if User.current.can_modify?(project) && !repository.is_dod_repository?
+      .col
+        = link_to('#', title: 'Edit Repository',
+                       data: { toggle: 'modal', target: "#edit-repository-#{repository.id}" }) do
+          %i.fas.fa-edit.text-secondary
+      .col
+        = link_to('#', title: 'Add Repository Path',
+                       data: { toggle: 'modal', target: "#add-repository-path-#{repository.id}" }) do
+          %i.fas.fa-plus-circle.text-primary
+    .col
+      - url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
+      = link_to(url, title: 'Download Repository') do
+        %i.fas.fa-download.text-secondary
+
+    - if User.current.can_modify?(project)
+      .col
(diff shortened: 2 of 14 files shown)


Ana06:
[Repositories](https://trello.com/c/fJbKmTh9/968-repositories)

src/api/app/views/webui2/webui/repositories/_repository_entry.html.haml
```diff
@@ -0,0 +1,7 @@
+.col-12.col-md-6.col-lg-4
+  .card.mb-3.repository-card
+    .card-header
+      = link_to(repository, project_repository_state_path(project: project, repository: repository), class: 'font-weight-bold')
```
dmarcoux: The repository's name is missing in the path, so the path should be:

project_repository_state_path(project: project, repository: repository<b>.name</b>)

Otherwise the path is generated with the ID when it needs the name.

src/api/app/views/webui2/webui/repositories/_repository_entry.html.haml
```diff
@@ -0,0 +1,7 @@
...
+      .small
+        #{repository.architectures.pluck(:name).join(', ')}
```
dmarcoux: When no architectures are selected, nothing is displayed. Comparing with the old UI, we displayed back then `()`. Now, I believe we should display something along the lines of `No architecture selected`

src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
```diff
@@ -0,0 +1,49 @@
...
+        = link_to('#', alt: 'Edit Repository icon', title: 'Edit Repository',
+                       data: { toggle: 'modal', target: "#edit-repository-#{repository.id}" }) do
+          %i.fas.fa-edit.text-secondary
+      .col
+        = link_to('#', alt: 'Add Repository path icon', title: 'Add Path Repository',
+                       data: { toggle: 'modal', target: "#add-repository-path-#{repository.id}" }) do
+          %i.fas.fa-plus-circle.text-secondary
+    .col
+      - url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
+      = link_to(url, alt: 'Download Repository icon', title: 'Download Repository') do
```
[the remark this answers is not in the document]
DavidKang: `donwload_url` is from a @configuration, maybe you don't have it

DavidKang:
@dmarcoux, could you review it again? :smile_cat:

src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
```diff
@@ -0,0 +1,49 @@
...
+    .col
+      - url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
+      = link_to(url, title: 'Download Repository') do
+        %i.fas.fa-download.text-secondary{ alt: 'Download Repository icon' }
+
+    - if User.current.can_modify?(project)
+      .col
+        = link_to('#', title: 'Delete Repository',
+                       data: { toggle: 'modal', target: "#delete-repository-#{repository.id}" }) do
+          %i.fas.fa-trash.text-danger{ alt: 'Delete Repository icon' }
```
dmarcoux: It should be using the same icon as the other delete actions, so `fa-times-circle` instead of `fa-trash`

src/api/app/controllers/webui/repositories_controller.rb
```diff
@@ -17,7 +17,14 @@ def index
 
     @user_can_set_flags = policy(@project).update?
 
-    switch_to_webui2 if @package
```
coolo: so you're breaking the consistency of bootstrap pages for packages? I don't like that idea.

src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
Ana06: We have two different delete actions in this page and it would be nice to differentiate them. I would prefer to keep that icon

dmarcoux: We also have two different delete actions on the package show page and we still went with the same icon in both cases. This is because users expect a delete action when they see the `fa-times-circle` action (once they have seen it once). This is to be consistent

src/api/app/controllers/webui/repositories_controller.rb
dmarcoux: Good point! We can keep this line for now and once we have finished the repositories page for the projects, we will have only `switch_to_webui2` left.

src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
Ana06: I think the other icon is also really clear. User also expect the same icon to do the same thing, and here there are doing different things, deleting different things

Ana06: Anyway, it is change already. We can discuss this in another PR

src/api/app/controllers/webui/repositories_controller.rb
Ana06: exactly, good catch @coolo :wink: It is changed

ChrisBr:
I found an issue but it is out of scope / was already like this before:

image

There is the path multiple times listed, no idea what the difference is. We should show the difference or drop them.

Ana06:
@ChrisBr 

> I found an issue but it is out of scope / was already like this before

I add it as optional to the card, if we don't have time for it, I will create an issue. :wink:

src/api/app/views/webui2/webui/repositories/index.html.haml
```diff
@@ -16,27 +16,48 @@
...
+        - if User.current.is_admin?
+          %li.list-inline-item
+            = link_to('#', data: { toggle: 'modal', target: '#add-dod-repository-modal' }, class: 'nav-link') do
+              %i.fas.fa-plus-circle.text-success
+              Add DoD Repository
+            =# render partial: 'add_dod_repository_modal', locals: { project: @project }
 
+  .card.mb-3
+    .card-header.d-flex.justify-content-between
+      %h5 My Repositories
```
ChrisBr:
```suggestion
      %h5
        Repositories for #{@project}
```

ChrisBr:
selection_002

Two small things I noticed but we can also do that later on:

1) I think we should make all cards / boxes the same size to remove the whitespace I marked
2) I already talked with @Ana06 about using a label for the architecture in the same line as the repository name. One argument was that there can be many architectures. I think this is not often the case and maybe we can do sth similar like you did with the path: Just showing three architectures and then another label with "X more" which displays the hidden repositories in a mouse over (like we do with the build results). What do you think?

Ana06:
@ChrisBr 

> I think we should make all cards / boxes the same size to remove the whitespace I marked

I have to say we tried it yesterday and we didn't like it and we didn't find a nice way to do it really. Any proposal how to do this?

Ana06:
We agree this can be merged as the test failure is unrelated. @eduardoj will create a PR to skip those tests (and an issue to fix them :rofl:)

src/api/app/views/webui2/webui/repositories/index.html.haml
Ana06: Ouch I just saw the comment :see_no_evil: I'll change it in the following PR :wink:

PR #6446, opened 2018-12-03
Refactor and adopt pulse to bootstrap

hennevogel:
This was some basic functionality. A bit extended now to give some more information about what is happening in a Project.

peek 2018-12-05 11-07

PR #6472, opened 2018-12-05
Disable InstanceVariables in haml-lint

hennevogel:
I find [the arguments that are made](http://www.carlosramireziii.com/stop-using-instance-variables-in-partials.html) very weak. Instance variables exist for a
reason, to pass data to the higher layers. That is not a weakness to avoid in
the top most layer, that partials are. It's a strength you should embrace all
around.

I set up @user in my action in the controller, pass it to the view, which makes
use of it in the partial. All the arguments that are brought forward against
using them in partials, like coupling with the controller action etc., are also
true for using them in the view. You wouldn't stop using instance variables all
around, would you? So why in partials?

Another argument is that it makes partials reusable. Yes partials _can_ be
reusable, but is that the most important part of partials? I don't think so,
there are other, equally important, features of partials like abstraction. Why
would you make them way harder to use just to potentially make a partial
reusable? Even if it does not neet to be reused?

This is not a question of general style. It's a question of using the right
thing at the right time and shouldn't be linted at all IMHO.

### DIFF
--- a/src/api/.haml-lint.yml
+++ b/src/api/.haml-lint.yml
@@ -4,10 +4,7 @@ linters:
 
   # Apply lint in all partials, but exclude 'breadcrumb_items' partials
   InstanceVariables:
-    enabled: true
-    file_types: not_breadcrumb_items
-    matchers:
-      not_breadcrumb_items: \A_(?!breadcrumb_items).*\.haml\z
+    enabled: false
 
   RuboCop:
     enabled: true


hennevogel:
If you reuse partials, fine, use locals. But why do you think this is default and should be enforced?

ChrisBr:
> If you reuse partials, fine, use locals. But why do you think this is default and should be enforced?

1) When I want to reuse I have to refactor. Yes, f*** tomorrow but this a very easy way to avoid touching the code again.
2) I don't want to have several different ways how we use partials. We should use one way how to render partials.
3) I think this is just because of "comfort". I don't see any disadvantage of using locals except that you have to write slightly more code.

hennevogel:
> When I want to reuse I have to refactor. Yes, f*** tomorrow but this a very easy way to avoid touching the code again.

So where does this stop? Shouldn't everything be reusable then? Shouldn't views be reusable? What's so special about partials?

> I don't want to have several different ways how we use partials. We should use one way how to render partials.

But partials are just partials, they have no *single* function. They can do many different things. It's like saying every instance variable should be an array because there should be only one way how we use instance variables.

> I think this is just because of "comfort". I don't see any disadvantage of using locals except that you have to write slightly more code.

Well it's not only slightly more code. In #6446 for instance it would mean significant more code for partials that never will be reused. It also would make this way harder to change as you have to keep in mind and change every `locals` hash.

dmarcoux:
I agree with @ChrisBr on decoupling the controller's actions and the partials in regards to the instance variables. One concrete example of this is with Bootstrap. We currently have 2 versions of our UI and using local variables allow us to improve the naming of variables in the new Bootstrap partials and still not break the partials from the previous UI. Sure, we could refactor and do the changes in the partials from the previous UI, but this is not what we want to do right now, even though it's fairly simple, it's not part of the scope of the deliverable.

In the end, it gives us a bit more flexibility and can reduce the amount of refactoring involved if an instance variable changes (if it ever does... this is the question...). The partials don't need to be changed, except for their `render` calls.

I like the local variables approach a bit more, but I don't have a strong opinion on this.

Ana06:
I agree with @ChrisBr too :+1:

hennevogel:
> Sure, we could refactor and do the changes in the partials from the previous UI, but this is not what we want to do right now, even though it's fairly simple, it's not part of the scope of the deliverable.

And that is a reason to *enforce* this? If we don't lint for this you can still do it :-)

hennevogel:
> but right now, I am not convinced on why it is beneficial to do so

You can write non-reusable partials for abstracting view code¹  without passing the data needlessly through another hoop.

¹ I would actually argue that those are the majority of partials but I have not crunched any numbers.

ChrisBr:
> Shouldn't views be reusable?

By default, views are coupled to the action, yes.

> But partials are just partials, they have no single function. They can do many different things. It's like saying every instance variable should be an array because there should be only one way how we use instance variables.

That's comparing Äpfel mit Birnen :apple: :green_apple: I'm not saying that there is a single function for partials but using it with or without locals will have the exact same result. So we should decide which way we want to use (consistency!). And if I have to decide, I would decide for the way which gives us more flexibility (hint: locals).

hennevogel:
>> Shouldn't views be reusable?

> By default, views are coupled to the action, yes.

And by default, partials are never coupled to views? :-)

hennevogel:
[crunched it](https://gist.github.com/hennevogel/9fbad7874be11126fd4fbc030393e08c). 11 out of 49 (22%) partials in `webui2` are shared. So if we go with a default, we should enforce the usage of instance variables! Just give up already :clown_face:

src/api/.haml-lint.yml
```diff
@@ -4,10 +4,7 @@ linters:
 
   # Apply lint in all partials, but exclude 'breadcrumb_items' partials
```
dmarcoux: This comment should be removed if we decide to disable this linter.

hennevogel:
No one want's to fight me anymore? C'mon! :monkey:

hennevogel:
Another stupid side effect of this: Each level of partial involves another level of `locals` hash that you have to adopt if anything changes....

bgeuken:
@hennevogel 
> No one want's to fight me anymore? C'mon! monkey

I was busy with the sprint, others might have been as well:wink:

I didn't follow the full discussion, but here are my 2 cents:

Passing locals / collections to partials makes it clearer which variables are needed and which are not. It also pushes developers to think twice before adding yet another variable.
My main argument would be (trying to keep it short) that having to define lot's of instance variables in a controller or having to pass a lot variables via locals to a partial is a code smell.
Instead of trying to work around this developers should inspect why / if they need them all.

To name two examples:
* The project controller has a lot of these instance variables. Some of them are just storing [simple attribute checks](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/project.rb#L493). (one of these occurrences was fixed in #6430)
* The other example would be the pulse controller. When looking at the [update_pulse action](
https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/projects/pulse_controller.rb#L11) or thinking about what values / variables you would need to pass to your partial it becomes clear that much of the values / logic stored in the instance variables belong to a model.

hennevogel:
alright I give up. the internet wins again! :clown_face:

coolo:
Can we discuss [Cassettes Suck](https://blog.discogs.com/en/cassettes-suck) now? :)

PR #6476, opened 2018-12-05
Prevent exception in repositories breadcrumb

krauselukas:
project_repository_state_path needs a repository defined in order to find
the right route. When the breadcrumb is used in an view where
no repository is set, it will throw an exception. For example in
the add repository view, which will be added soonish.

### DIFF
--- a/src/api/app/views/webui2/webui/repositories/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui2/webui/repositories/_breadcrumb_items.html.haml
@@ -5,7 +5,7 @@
 - if current_page?(repositories_path)
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
     Repositories
-- elsif current_page?(project_repository_state_path)
+- elsif defined?(@repository) && current_page?(project_repository_state_path)
   %li.breadcrumb-item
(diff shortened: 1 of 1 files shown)


src/api/app/views/webui2/webui/repositories/_breadcrumb_items.html.haml
```diff
@@ -5,7 +5,7 @@
 - if current_page?(repositories_path)
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
     Repositories
-- elsif current_page?(project_repository_state_path)
+- elsif defined?(@repository) && current_page?(project_repository_state_path)
```
coolo: I count that as vote for #6472 :)

Ana06: we use instance variables in the breadcrumb. I think that is ok, but I would say this is an special case.

bgeuken: You take the one exception from the rule (that we consciously took) and use it as a counter-argument? Nice try:smiley:

PR #6482, opened 2018-12-06
[api] support releasing into maintenance_release without incident pro…

adrianschroeter:
…ject

Using readable timestamp as fallback.

Used in continous rebuild projects for container releases

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -996,6 +996,9 @@ class SourceController < ApplicationController
       multibuild_container = params[:package].gsub(/^.*:/, '')
     end
 
+    # uniq timestring for all targets
+    time_string = Time.now.utc.strftime('%Y%m%d%H%M%S')
+
     # specified target
     if params[:target_project]
       # we do not create it ourself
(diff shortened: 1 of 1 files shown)


adrianschroeter:
needed also for 15.1 containers in build.o.o now.

src/api/app/controllers/source_controller.rb
```diff
@@ -996,6 +996,9 @@ def package_command_release
       multibuild_container = params[:package].gsub(/^.*:/, '')
     end
 
+    # uniq timestring for all targets
+    time_string = Time.now.utc.strftime('%Y%m%d%H%M%S')
```
hennevogel: You're only using this once, why do you put this in a variable?

adrianschroeter: because it is used in a loop and must not change.

coolo: I guess this gets a little more obvious if you put that assignment before that loop then.

dmarcoux: It's also only needed if you get into the `else` on line 1007.

PR #6458, opened 2018-12-04
Add new package and branch package modal to project

krauselukas:
The new package and branch existing package view are basically
just a form. Since they need to be migrated to bootstrap,
they are a perfect fit for a modal dialog.

screenshot-2018-12-4 show opensuse factory - open build service 1
screenshot-2018-12-4 show opensuse factory - open build service

PR #6485, opened 2018-12-06
Be consistent in modal buttons

dmarcoux:
Use the same modal buttons as the other forms in which we create records

I realized that we missed that in the review of #6458. I found this while working on the subprojects page and comparing the modals...

So it's now like this:
modal-buttons-1

modal-buttons-2


dmarcoux:
coveralls is stuck

PR #6487, opened 2018-12-06
Be consistent in displaying counts

dmarcoux:
Use badges like other views

Before:
badges-before

After:
badges


dmarcoux:
coveralls is stuck again

PR #6491, opened 2018-12-06
Add project variable to all *_path calls in our breadcrumb partial

bgeuken:
While working on the bootstrap migration for the status tab I got
routing issues due to missing 'project' parameters.
The reason for that is that we explicitly set the 'project' for
some path helper calls, but not for all.

```ruby
No route matches {:action=>"index", :controller=>"webui/repositories",
 :format=>"html"}, missing required keys: [:project]

  - unless project.defines_remote_instance? || project.is_maintenance?
    %li.nav-item
      = tab_link('Repositories', repositories_path)
    %li.nav-item
      = tab_link('Monitor', project_monitor_path(project))
  %li.nav-item
```


hennevogel:
coveralls hangs. coverage is ok, [see ci](https://circleci.com/gh/bgeuken/open-build-service/884). merging.

PR #6492, opened 2018-12-06
Use badge for group tasks

dmarcoux:
To be consistent with other views

Before:
before

After:
after


hennevogel:
coveralls hangs. coverage is ok, see ci. merging.

PR #6505, opened 2018-12-07
Fix lookup issue due to conflicting namespaces

bgeuken:
The Rails lookup code otherwise has trouble to identify which
'Project' class or module is meant.

I started to get this error when accessing one of the project
pages (in development mode):

  "undefined method `find_by' for Webui::Project:Module"


coolo:
and suddenly Hakiri understood what you were talking about :)

bgeuken:
I would be quite sure that hakiri raised those issues before and got marked as false positives. It just pops up again, since the code changed.
Anyway, I'll review them.

coolo:
I'm just amused that hakiri complains about the views you didn't touch.

bgeuken:
They were all false positives. Hakiri was concerned that we directly show user provided data (params[:foo]) in the view, but we actually passed them through some methods before.

bgeuken:
@coolo Haliri checks the assignment of variables that are used in views.

PR #6506, opened 2018-12-07
Add pulse to breadcrumbs

bgeuken:


dmarcoux:
I believe it should be like what we did for the [meta breadcrumbs](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/views/webui2/webui/projects/meta/_breadcrumb_items.html.haml)

PR #6507, opened 2018-12-07
Render project_packages modals in the right location

dmarcoux:
This fixes the color of the headers from green to black, so it's now consistent with other modals

Before:
before-1
before-2

After:
after-1
after-2

### DIFF
--- a/src/api/app/views/webui2/webui/project/_project_packages.html.haml
+++ b/src/api/app/views/webui2/webui/project/_project_packages.html.haml
@@ -28,6 +28,3 @@
             = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#new-package-branch-modal' }) do
               %i.fas.fa-code-branch.text-primary
               Branch Existing Package
-
-      = render partial: 'new_package_modal', locals: { project: project }
-      = render partial: 'new_package_branch_modal', locals: { project: project, remote_projects: remote_projects }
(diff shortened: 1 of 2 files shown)


src/api/app/views/webui2/webui/project/_project_packages.html.haml
```diff
@@ -28,6 +28,3 @@
             = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#new-package-branch-modal' }) do
               %i.fas.fa-code-branch.text-primary
               Branch Existing Package
-
-      = render partial: 'new_package_modal', locals: { project: project }
-      = render partial: 'new_package_branch_modal', locals: { project: project, remote_projects: remote_projects }
```
eduardoj: `remote_projects` is not used in this partial any more. You can also remove `remote_projects` from the params in the [call](https://github.com/openSUSE/open-build-service/pull/6507/files#diff-1e13e4f606c0a3c55ae081f2b1c1cbd0R59)

dmarcoux: It is still [in use](https://github.com/openSUSE/open-build-service/blob/7a4d1dc2e7d0e6bdd571266b9c84e4f9059ad5df/src/api/app/views/webui2/webui/project/_new_package_branch_modal.html.haml#L18), but you are right that it can be removed in the line you linked.

dmarcoux:
@eduardoj Please have another look

PR #6508, opened 2018-12-07
Project log entry cleanup job

hennevogel:

### DIFF
--- a/src/api/app/models/project_log_entry.rb
+++ b/src/api/app/models/project_log_entry.rb
@@ -27,9 +27,8 @@ class ProjectLogEntry < ApplicationRecord
     entry
   end
 
-  # Delete old entries
-  def self.clean_older_than(date)
-    where('datetime < ?', date).delete_all
+  def self.cleanup
+    where(event_type: [:build_fail, :build_success]).where('datetime < ?', Time.zone.yesterday).delete_all
   end
 
   # Human readable message, based in the event class
--- a/src/api/app/controllers/webui/projects/pulse_controller.rb
+++ b/src/api/app/controllers/webui/projects/pulse_controller.rb
@@ -18,7 +18,7 @@ module Webui
         end
 
         pulse = @project.project_log_entries.where(datetime: range).order(datetime: :asc)
-        @builds = pulse.where(event_type: [:build_fail, :build_success])
+        @builds = pulse.where(event_type: [:build_fail, :build_success]).where(datetime: 24.hours.ago..Time.zone.now)
         @new_packages = pulse.where(event_type: :create_package)
         @deleted_packages = pulse.where(event_type: :delete_package)
         @branches = pulse.where(event_type: :branch_command)
--- a/src/api/app/jobs/cleanup_project_log_entries.rb
+++ b/src/api/app/jobs/cleanup_project_log_entries.rb
@@ -1,0 +1,7 @@
+class CleanupProjectLogEntries < ApplicationJob
+  queue_as :project_log_rotate
+
+  def perform
+    ProjectLogEntries.cleanup
+  end
+end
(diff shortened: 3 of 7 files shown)


src/api/app/models/project_log_entry.rb
```diff
@@ -27,9 +27,8 @@ def self.create_from(payload, created_at, event_model_name)
     entry
   end
 
-  # Delete old entries
-  def self.clean_older_than(date)
-    where('datetime < ?', date).delete_all
+  def self.cleanup
+    where(event_type: [:build_fail, :build_success]).where('datetime < ?', Time.zone.yesterday.beginning_of_day).find_each(&:delete)
```
coolo: why not delete_all? There can be a lot of them after all

coolo:
I find the concept of 'today' a little underdefined in a project with contributors in all places of the world. 

I'm already annoyed by https://metrics.opensuse.org/d/qjkd7uBmz/rails-metrics?orgId=1 defaulting to 'Today', which means basically no data available at the time I look at it. I find 'last 24 hours' a little more meaningful - no matter in what timezone you are.

src/api/app/models/project_log_entry.rb
hennevogel: From build.o.o

```ruby
ProjectLogEntry.where(event_type: [:build_fail, :build_success]).where('datetime > ?', Time.zone.yesterday.beginning_of_day).count
=> 45580
```

And we only started yesterday at 20:00...

hennevogel:
> I find the concept of 'today' a little underdefined in a project with contributors in all places of the world.

Good point. I change it.

src/api/app/models/project_log_entry.rb
coolo: How does this answer my question? delete_all is one SQL statement - your code is 45581 SQL statements.

hennevogel: Fear of locks :cowboy_hat_face: 

https://www.youtube.com/watch?v=RwpQuvNP0R8

Ana06: The only problem with `delete_all` is that it goes straight to the database skipping things like callbacks and dependent rules, but we don't have anything like that in those models :+1:

src/api/app/models/project_log_entry.rb
```diff
@@ -27,9 +27,8 @@ def self.create_from(payload, created_at, event_model_name)
     entry
   end
 
-  # Delete old entries
-  def self.clean_older_than(date)
-    where('datetime < ?', date).delete_all
+  def self.cleanup
+    where(event_type: [:build_fail, :build_success]).where('datetime < ?', Time.zone.yesterday.beginning_of_day).delete_all
```
Ana06: why is `beginning_of_day` needed? :thinking: `yesterday` returns an instance of the class `Date` which is already at the beginning of the day

hennevogel: Yeah I guess it's the same...

```
ProjectLogEntry.where(event_type: [:build_fail, :build_success]).where('datetime > ?', Time.zone.yesterday).to_sql
=> "SELECT `project_log_entries`.* FROM `project_log_entries` WHERE `project_log_entries`.`event_type` IN ('build_fail', 'build_success') AND (datetime > '2018-12-06')"
irb(main):025:0> ProjectLogEntry.where(event_type: [:build_fail, :build_success]).where('datetime > ?', Time.zone.yesterday.beginning_of_day).to_sql
=> "SELECT `project_log_entries`.* FROM `project_log_entries` WHERE `project_log_entries`.`event_type` IN ('build_fail', 'build_success') AND (datetime > '2018-12-06 00:00:00')"
```

src/api/app/models/project_log_entry.rb
hennevogel: Yeah so does `.delete`. I was worried about it taking too long but I guess this was unfounded.

src/api/app/models/project_log_entry.rb
Ana06: yes :eyes:

hennevogel:
coverage my ass...

PR #6509, opened 2018-12-07
Align links with users/groups page with other pages to be consistent

dmarcoux:
- Use nav-link classes for users/groups page
- Use the correct ID for 'Add Group' link
- Capitalize users/groups links correctly

Before:
before-links

After:
after-links


dmarcoux:
The specs should be fine now...

PR #6511, opened 2018-12-07
Fix autocomplete in group members' Add User modal

dmarcoux:
Before:
before-modal

After:
after-modal


dmarcoux:
@Ana06 The tests are already passing

PR #6515, opened 2018-12-07
Add badges to the tabs in the group show page

saraycp:
Badges are added to the tabs and part of the show view file is moved to a partial to make it more readable.


saraycp:
[Add badges to the tabs in the group show page](https://trello.com/c/35yuD4Od/996-add-badges-to-the-tabs-in-the-group-show-page)

PR #6516, opened 2018-12-07
Sync different sizes of our bootstrap tables

bgeuken:
We recently changed the table size for the user and groups tab
to make it consume less space. This commit adopts the other tables to
use the same table size everywhere.

Follow up of 46b35d8743e2fba48c38ce0ca4074ca0d459358b


bgeuken:
Related PR for our pattern library: https://github.com/openSUSE/obs-patterns/pull/46

bgeuken:
[Make sure all tables are using the same size](https://trello.com/c/q7yVEleL/997-make-sure-all-tables-are-using-the-same-size)

saraycp:
IMHO I like more the "Before" design, with more air around content, but... I agree it is a good way to optimize the space  :+1: 

@bgeuken please resolve conflicts.

PR #6518, opened 2018-12-07
Merge pulse actions into one

hennevogel:
Both requests (html & xhr) are somehow about show.


src/api/app/models/project_log_entry.rb
```diff
@@ -7,6 +7,15 @@ class ProjectLogEntry < ApplicationRecord
 
   validates :event_type, :datetime, :project_id, presence: true
 
+  scope :builds, -> { where(event_type: [:build_fail, :build_success]) }
+  scope :create_packages, -> { where(event_type: :create_package) }
```
Ana06: what about only one scope with a parameter for the event_types? :thinking:

hennevogel: That scope is called `.where` :wink:

src/api/app/controllers/webui/projects/pulse_controller.rb
```diff
@@ -2,13 +2,18 @@ module Webui
...
       def show
         switch_to_webui2
         @pulse = @project.project_log_entries.page(params[:page])
       end
 
-      def update_pulse
+      private
+
+      def set_pulse
+        return true unless request.xhr?
```
ChrisBr: You could do:

```
before_action :set_pulse, only: :xhr?

def xhr?
  request.xhr?
end
```

Maybe the helper method is not necessary but not sure about that

hennevogel: Nope, see [_normalize_callback_options](https://api.rubyonrails.org/classes/AbstractController/Callbacks/ClassMethods.html#method-i-_normalize_callback_options)...

hennevogel: there is `if: -> {...}` though...

src/api/app/models/project_log_entry.rb
Ana06: mm no.... what I am proposing is:

```Ruby
scope packages_with_type, -> (type) { where(event_type: type) }
```
I just don't see the point on having that many scope for such simple conditions. I would prefer to use `where` directly than the scopes. I don't see they provide any benefit.

hennevogel: I know this is what you proposing. That means:

```ruby
ProjectLogEntry.packages_with_type(:build_fail)
```
instead of

```ruby
ProjectLogEntry.where(packages_with_type :build_fail)
```

One everybody knows, one everybody has to look up at least once....

hennevogel: dropped the scope commit now...

hennevogel:
Hakiri hangs. Merging.

PR #6522, opened 2018-12-07
Cleanup classes that create ProjectLogEntries

hennevogel:
PLEs belong to a project so it makes not much sense to track
creation/deletion of Projects with PLEs.

Don't know what Event::Upload is. No events or PLEs like
that in production, we also show this nowhere.


coolo:
Event::Upload is created by the source server if a package is uploaded:
https://github.com/openSUSE/open-build-service/blob/master/src/backend/bs_srcserver#L587 

I *think* this is for initial setup

PR #6524, opened 2018-12-07
Add badges to the tabs in the user page

saraycp:
For consistency with the rest of the views I have also added badges to the tabs in the user page.

screenshot-2018-12-7 home of admin - open build service


saraycp:
[Add badges to the tabs in the user page](https://trello.com/c/LKZkLgHf/998-add-badges-to-the-tabs-in-the-user-page)

PR #6531, opened 2018-12-09
Feature job history

hennevogel:
Depends on #6519

### DIFF
--- a/src/api/app/controllers/webui/projects/rebuild_times_controller.rb
+++ b/src/api/app/controllers/webui/projects/rebuild_times_controller.rb
@@ -18,7 +18,12 @@ module Webui
           return
         end
         bdep = Backend::Api::BuildResults::Binaries.builddepinfo(@project.name, @repository, @arch)
-        jobs = Backend::Api::BuildResults::JobHistory.not_failed(@project.name, @repository, @arch, (@packages.size + @ipackages.size) * 3)
+        jobs = Backend::Api::BuildResults::JobHistory.for_repository_and_arch(project_name: @project.name,
+                                                                              repository_name: @repository,
+                                                                              arch_name: @arch,
+                                                                              filter: { limit: (@packages.size + @ipackages.size) * 3,
+                                                                                        code: ['succeeded', 'unchanged'] },
+                                                                              raw: true)
         unless bdep && jobs
           flash[:error] = "Could not collect infos about repository #{@repository}/#{@arch}"
           redirect_to controller: '/webui/project', action: :show, project: @project
--- a/src/api/app/models/package_build_status.rb
+++ b/src/api/app/models/package_build_status.rb
@@ -141,23 +141,22 @@ class PackageBuildStatus
 
     # first we check the lastfailures. This route is fast but only has up to
     # two results per package. If the md5sum does not match, we have to dig deeper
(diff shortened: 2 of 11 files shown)


src/api/app/models/package_build_status.rb
```diff
@@ -140,23 +140,22 @@ def check_everbuilt(srep, arch)
 
     # first we check the lastfailures. This route is fast but only has up to
     # two results per package. If the md5sum does not match, we have to dig deeper
-    hist = Xmlhash.parse(Backend::Api::BuildResults::JobHistory.last_failures(@pkg.project.name, @pkg.name, srep['name'], arch))
-    return unless hist
-    hist.elements('jobhist') do |jh|
-      if jh['verifymd5'] == @verifymd5 || jh['srcmd5'] == @srcmd5
+    jobhistory = @pkg.jobhistory(repository_name: srep['name'], arch_name: arch, filter: { code: 'lastfailures', limit: 2 })
```
hennevogel: Made the code do what the comment says now....

src/api/app/controllers/webui/projects/rebuild_times_controller.rb
```diff
@@ -18,7 +18,12 @@ def show
           return
         end
         bdep = Backend::Api::BuildResults::Binaries.builddepinfo(@project.name, @repository, @arch)
-        jobs = Backend::Api::BuildResults::JobHistory.not_failed(@project.name, @repository, @arch, (@packages.size + @ipackages.size) * 3)
+        jobs = Backend::Api::BuildResults::JobHistory.for_repository_and_arch(project_name: @project.name,
+                                                                              repository_name: @repository,
+                                                                              arch_name: @arch,
+                                                                              filter: { limit: (@packages.size + @ipackages.size) * 3,
```
vpereira: I would move the filter block to a private method and have the 3 as default parameter in the method, something like

```
filter: set_limit

... 

private
def set_limit(limit_offset=3)
  {  limit: (@packages.size + @ipackages.size) * limit_offset,
     code: ['succeeded', 'unchanged'] 
  }
end
```

src/api/app/models/package_build_status.rb
```diff
@@ -141,23 +141,22 @@ def check_everbuilt(srep, arch)
...
+    jobhistory.each do |entry|
+      if entry.verifymd5 == @verifymd5 || entry.srcmd5 == @srcmd5
```
vpereira: Why not:

```
  @everbuilt = true if entry.verifymd5 == @verifymd5 || entry.srcmd5 == @srcmd5
```
or move the whole equality test to its own method.

src/api/lib/backend/connection_helper.rb
```diff
@@ -103,6 +103,7 @@ def rename_params(params, renames)
     def expand_params(params, expand)
       expanded_params = []
       expand.each do |key|
+        next if params[key].is_a?(String)
```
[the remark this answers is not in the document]
hennevogel: This method expected an array. I just added this so it also works with strings.

src/api/app/controllers/webui/projects/rebuild_times_controller.rb
vpereira: for readability, I'm old I cannot cope with long lines

PR #6537, opened 2018-12-10
Move prjconf actions from project_controller to its own controller

vpereira:


src/api/app/controllers/webui/projects/project_configuration_controller.rb
```diff
@@ -0,0 +1,37 @@
+module Webui
+  module Projects
+    class ProjectConfigurationController < WebuiController
+      before_action :set_project
```
ChrisBr: What about ``verify_authorized``?
https://github.com/varvet/pundit#ensuring-policies-and-scopes-are-used

src/api/app/controllers/webui/projects/project_configuration_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      def show
+        sliced_params = params.slice(:rev)
+        sliced_params.permit!
```
ChrisBr: [nitpick] you could move that to ``params.slice(:rev).permit!``

src/api/app/services/project_configuration_service/project_configuration_updater.rb
```diff
@@ -0,0 +1,22 @@
...
+      @params[:user] = @user.login
+      @project.config.save(sliced_params, @params[:config])
+    end
+
+    private
+
+    def sliced_params
+      sliced_params = @params.slice(:user, :comment)
+      sliced_params.permit!
+      sliced_params.to_h
```
ChrisBr: Why is the ``to_h`` necessary?

src/api/app/controllers/webui/projects/project_configuration_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+        @content = @project.config.content(sliced_params.to_h)
+        switch_to_webui2
+        return if @content
+        flash[:error] = @project.config.errors.full_messages.to_sentence
+        redirect_to projects_path(nextstatus: 404)
+      end
+
+      def update
+        authorize @project, :update?
+        content = ::ProjectConfigurationService::ProjectConfigurationUpdater.new(@project, User.current, params).call
```
DavidKang: I like the idea of using services, but in this case, I don't see the value of it, this could be a private method in `ProjectConfigurationController`

src/api/spec/cassettes/Webui_ProjectController/GET_add_maintained_project_dialog/1_42_1.yml
```diff
@@ -0,0 +1,157 @@
+---
```
DavidKang: What happens with the old cassettes?

src/api/app/controllers/webui/projects/project_configuration_controller.rb
ChrisBr: This is not the content but the result. In general I would prefer to not rely on side effects of the Updater class (returning the result but expecting that it does change the @project object). We could return self and ask self about the result and the errors.

```
    def call
       @params[:user] = @user.login
       @saved = @project.config.save(sliced_params, @params[:config])
       self
     end

    def saved?
        @saved || false
    end

    def errors
        @project.config.errors.full_messages.to_sentence
    end
```

Then you can do:

```
result = ::ProjectConfigurationService::ProjectConfigurationUpdater.new(@project, User.current, params).call

if result.saved?
  ....
  flash.now[:error] = result.errors
else

end
```

src/api/app/services/project_configuration_service/project_configuration_updater.rb
```diff
@@ -0,0 +1,22 @@
+module ProjectConfigurationService
+  class ProjectConfigurationUpdater
+    def initialize(project, user, params)
+      @user = user
+      @project = project
+      @params = params
+    end
+
+    def call
+      @params[:user] = @user.login
```
ChrisBr: I think this should be part of the sliced_params method.

src/api/app/controllers/webui/projects/project_configuration_controller.rb
vpereira: the idea is to move away from private methods on controllers :)

src/api/app/services/project_configuration_service/project_configuration_updater.rb
vpereira: since rails 5.1, ActionController::Parameters no longer inherits from Hash. Therefore you have to call the ```.to_h``` if the other method is expecting  a hash.

src/api/app/controllers/webui/projects/project_configuration_controller.rb
ChrisBr: I also don't see much value in this particular service as there is not much logic into it and adds another layer of indirection. However, I'm fine with it leaving it like it is.

vpereira: IMO it will be easier to stub it to for instance, having a wrong or nil content.

src/api/app/services/project_configuration_service/project_configuration_updater.rb
vpereira:
```
irb(main):002:0> def foo(params: {})
irb(main):003:1>   puts params.inspect
irb(main):004:1> end
=> :foo
irb(main):005:0> foo(ActionController::Parameters.new(foo:'bar'))
        1: from (irb):5
ActionController::UnfilteredParameters (unable to convert unpermitted parameters to hash)
irb(main):006:0> 
```

vpereira: @chrisBR https://github.com/rails/rails/pull/28734 as reference

ChrisBr: Yes because it is not permitted

```
foo(params: ActionController::Parameters.new(foo:'bar').permit!)
# => <ActionController::Parameters {"foo"=>"bar"} permitted: true>
```

vpereira: please review it again @ChrisBr

Issue #6526, opened 2018-12-07
Dropdown menu link becomes completely green when :active and :hover


saraycp:
# Issue Description
Mouse over and click on one of the dropdown menu links shows everything green (combination of :active and :hover).

![home of admin open build service 2](https://user-images.githubusercontent.com/2581944/49663952-9ccfec80-fa50-11e8-8f9d-becfdc516fee.png)

How to Reproduce
===============
1. Open build.opensuse.org on a small screen (phone, table, narrowed browser) 
2. Open a project and click on the drop-down menu arrow.
3. Move your mouse over one of the menu options (not the active one)
4. Press mouse's button for a while.
5. You'll see everything green.

PR #6539, opened 2018-12-10
Show white text when clicking on dropdown menu link

saraycp:
When a dropdown menu link was clicked, the text became green on top of a
green background. Now text is set to white in that case (:active).

Fixes: #6526

Before:
home of admin open build service 2

After:
home of admin open build service 4

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/tabs-component.scss
+++ b/src/api/app/assets/stylesheets/webui2/tabs-component.scss
@@ -6,8 +6,14 @@ ul.nav.nav-tabs li.nav-item {
       color: $link-hover-color;
     }
 
-    &.dropdown-item.active {
+    &.dropdown-item.active,
+    &.dropdown-item:active {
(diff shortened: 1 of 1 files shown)


bgeuken:
How would it look like if we use a white badge when the background is green? Would you mind giving it a try?

coolo:
I read 'snow white' in the summary and wondered what winter special easter egg you're preparing :)

saraycp:
@Ana06 , squashed!

@coolo I was just thinking in something like the following... sorry, it's not my fault it's just OBS' Christmas spirit! :laughing: 

welcome open build service 1

bgeuken:
> @coolo I was just thinking in something like the following... sorry, it's not my fault it's just OBS' Christmas spirit!

Let's ship it!:laughing:

dmarcoux:
It should be in the beta to have even more people join!

PR #6540, opened 2018-12-10
Fix test for branch packages in modal

vpereira:
Fix the failing spec for Maintenance Workflow.

### DIFF
--- a/src/api/spec/bootstrap/features/webui/maintenance_workflow_spec.rb
+++ b/src/api/spec/bootstrap/features/webui/maintenance_workflow_spec.rb
@@ -32,7 +32,10 @@ RSpec.feature 'Bootstrap_MaintenanceWorkflow', type: :feature, js: true, vcr: tr
     sleep 1 # Needed to avoid a flickering test.
     expect(page).to have_text('Source')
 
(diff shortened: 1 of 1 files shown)


src/api/spec/bootstrap/features/webui/maintenance_workflow_spec.rb
```diff
@@ -32,7 +32,11 @@
     sleep 1 # Needed to avoid a flickering test.
     expect(page).to have_text('Source')
 
-    click_button('Accept')
+    within('#branch-modal .modal-footer') do
+      click_button('Accept')
+    end
+    sleep 1
```
coolo: what is that sleep good for? capybara will reload the page until it has the text.

vpereira: without sleep fails. with sleep passes.

coolo: this is a super dangerous myth that will lead to tons of sleeps everywhere. Rule of thumb: if you don't understand why a sleep helps, it's the wrong fix :)

PR #6502, opened 2018-12-07
Enable Bootstrap project show to all beta users

hennevogel:
Align the old UI with the new Bootstrap UI for specs:
- Update project feature specs
- Maintenance workflow feature specs
- Watchlist feature specs

These include fixes to existing feature specs to make them pass when
testing the new bootstrap UI. Specs that can't be written for both UIs
got ported to the spec/bootstrap/* directory.

Fix Submit as Update link condition

Co-authored-by: Björn Geuken <bgeuken@suse.de>
Co-authored-by: Dany Marcoux <dmarcoux@suse.com>
Co-authored-by: Eduardo Navarro <enavarro@suse.com>
Co-authored-by: Victor Pereira <vpereira@suse.com>

PR #6541, opened 2018-12-10
Write project comment spec for bootstrap

eduardoj:
This will fix comment tests for #6502.

### DIFF
--- a/src/api/spec/features/webui/comments_spec.rb
+++ b/src/api/spec/features/webui/comments_spec.rb
@@ -2,6 +2,7 @@ require 'browser_helper'
 
 RSpec.feature 'Comments', type: :feature, js: true do
   let(:user) { create(:confirmed_user, login: 'burdenski') }
+  let!(:comment) { create(:comment_project, commentable: user.home_project, user: user) }
 
   scenario 'can be created' do
     login user
@@ -13,13 +14,10 @@ RSpec.feature 'Comments', type: :feature, js: true do
   end
 
   scenario 'can be answered' do
-    skip_if_bootstrap
     login user
-    comment = create(:comment_project, commentable: Project.first, user: user)
     visit project_show_path(user.home_project)
-    # reply link contains an img element. Thus the regex
-    find('a', text: /.*Reply/).click
-    fill_in "reply_body_#{comment.id}", with: 'Reply Body'
+    find("#reply_link_id_#{comment.id}").click
+    fill_in("reply_body_#{comment.id}", with: 'Reply Body')
     click_button("add_reply_#{comment.id}")
 
     expect(page).to have_text('Reply Body')


dmarcoux:
Are you sure? Because we have the exact same spec in #6502 and it fails with this error:

```
 Comments when logged-in as a confirmed user can be answered
      Failure/Error: click_on('Reply')
      
      Capybara::ElementNotFound:
        Unable to find link or button "Reply"
      # ./spec/bootstrap/features/webui/projects/comments_spec.rb:15:in `block (3 levels) in <top (required)>'
```

eduardoj:
> Are you sure? ...

Fixed. I forgot to add the patch.

src/api/spec/bootstrap/features/webui/projects/comments_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  context 'when logged-in as a confirmed user' do
+    before do
+      login user
+    end
+
+    let!(:comment) { create(:comment_project, commentable: user.home_project, user: user) }
+
+    scenario 'can be answered' do
+      visit project_show_path(user.home_project)
+      find('#reply_link_id_1').click
```
dmarcoux: I believe it should be `find("#reply_link_id_#{comment.id}").click

eduardoj: Right. Changed.

src/api/spec/bootstrap/features/webui/projects/comments_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  context 'when logged-in as a confirmed user' do
+    before do
+      login user
+    end
+
+    let!(:comment) { create(:comment_project, commentable: user.home_project, user: user) }
+
+    scenario 'can be answered' do
+      visit project_show_path(user.home_project)
+      find("#reply_link_id_#{comment.id}").click
```
Ana06: I think this is the wrong approach, the button should be a button instead of a span.

dmarcoux: This won't change this line if it's a `button` or a `span`

Ana06: It will! The reason  why `click('Reply') doesn't find it, is that it is not a `button`. We don't need to use the complicated `id`, but to make that what looks like a button is a button. :bowtie:

dmarcoux: It's already a [button](https://github.com/openSUSE/open-build-service/blob/1fdd9038a0316c7cd6497b96dd753062258780fc/src/api/app/views/webui2/webui/comment/_links.html.haml#L4-L5) anyway. We can try with `Reply` as you suggest

src/api/spec/features/webui/comments_spec.rb
```diff
@@ -13,13 +14,10 @@
   end
 
   scenario 'can be answered' do
-    skip_if_bootstrap
```
Ana06: @eduardoj I think you haven't understood @dmarcoux proposal, if you remove this `skip` then you need to drop `src/api/spec/bootstrap/features/webui/projects/comments_spec.rb`

src/api/spec/bootstrap/features/webui/projects/comments_spec.rb
Ana06: I see this in the browser:

```
<span class="btn btn-sm btn-outline-secondary" id="reply_link_id_815556">
Reply
</span>
```

src/api/spec/features/webui/comments_spec.rb
```diff
@@ -13,13 +14,10 @@
...
     login user
-    comment = create(:comment_project, commentable: Project.first, user: user)
     visit project_show_path(user.home_project)
-    # reply link contains an img element. Thus the regex
-    find('a', text: /.*Reply/).click
-    fill_in "reply_body_#{comment.id}", with: 'Reply Body'
+    find("#reply_link_id_#{comment.id}").click
```
dmarcoux: And now because we have this spec for both the old and new UIs, we need to use the ID since the old UI has an image and this prevent `click_on('Reply')` from working.

src/api/spec/bootstrap/features/webui/projects/comments_spec.rb
dmarcoux: You don't have the latest version of master.

dmarcoux:
@Ana06 Please have another look

PR #6542, opened 2018-12-10
Remove 'close' icon from modal for package branching

bgeuken:
This align the package branch modal with the other modals.

Before:

screenshot from 2018-12-10 13-52-55

After:

screenshot from 2018-12-10 13-52-35


bgeuken:
I've updated the other two as well.

PR #6544, opened 2018-12-10
Update subheader for repository states

bgeuken:
Before:

screenshot-2018-12-10 opensuse build service

After:

screenshot from 2018-12-14 11-38-24

### DIFF
--- a/src/api/app/views/webui2/webui/package/binaries.html.haml
+++ b/src/api/app/views/webui2/webui/package/binaries.html.haml
@@ -1,7 +1,7 @@
 .card
   = render partial: 'tabs', locals: { package: @package, project: @project }
   .card-body
-    %h3 State of Repository #{@repository} for #{@project} / #{@package_name}
+    %h3 State of #{@repository} for #{@project} / #{@package_name}
     %ul.list-inline
       - if @configuration['download_url']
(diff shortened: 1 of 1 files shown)


bgeuken:
It's always on the same place, the right:wink:

You can compare it with [OBS:Server:Unstable](https://build.opensuse.org/package/binaries/OBS:Server:Unstable/obs-server/SLE_12_SP3). The icon is always the same (it belongs to the repository). The relevant information for these columns is the architecture.

bgeuken:
JFYI, I've dropped the architecture / repository "labels" and updated the screenshot.

coolo:
you're implying there is download link hidden in your screenshot?

adrianschroeter:
dropping the obsolete "Architecture:" string makes sense to me, but I had prefer to have the icon in front. (It is not always the same, it shows the repository state)

dmarcoux:
I also prefer to have the icon in front as it's consistent, always in the same spot. If it is after the text, it is then in a different position depending on the length of the text.

saraycp:
I also prefer the icon on the left.

bgeuken:
Alright, I've updated the PR and moved the icon back to the left. Thanks for the feedback

PR #6545, opened 2018-12-10
Add watchlists specs for Bootstrap

dmarcoux:
The HTML/CSS is too different to adapt the specs from the old UI


dmarcoux:
The changes need to be in #6502 because it will otherwise always fail since the project#show page is not rendered in Bootstrap. Closing this!

PR #6549, opened 2018-12-10
Not showing disabled build results on default

gsouzab:

### DIFF
--- a/src/api/app/models/local_build_result/for_package.rb
+++ b/src/api/app/models/local_build_result/for_package.rb
@@ -19,7 +19,7 @@ class LocalBuildResult
       backend_build_result.each do |result|
         result.elements('status').each do |status|
           results[status['package']] ||= []
-          if excluded?(status['code'])
+          if excluded_or_disabled?(status['code'])
             self.excluded_counter += 1
             next
           end
@@ -33,8 +33,9 @@ class LocalBuildResult
                            code: status['code'], state: result['state'], details: status['details'])
     end
 
-    def excluded?(status)
-      !show_all && status == 'excluded'
+    def excluded_or_disabled?(status)
+      return false if show_all
+      ['excluded', 'disabled'].include?(status)
     end
 
     def backend_build_result
--- a/src/api/app/views/webui/package/_buildstatus.html.haml
+++ b/src/api/app/views/webui/package/_buildstatus.html.haml
@@ -2,7 +2,7 @@
   .show_all_box
     = check_box_tag "show_all_#{@index}", true, @buildresults.show_all
     - label_message = @buildresults.excluded_counter.zero? ? 'Hide' :  "Show #{@buildresults.excluded_counter}"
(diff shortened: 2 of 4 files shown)


src/api/app/models/local_build_result/for_package.rb
```diff
@@ -33,8 +33,8 @@ def local_build_result(result, status)
                            code: status['code'], state: result['state'], details: status['details'])
     end
 
-    def excluded?(status)
-      !show_all && status == 'excluded'
+    def excluded_or_disabled?(status)
+      !show_all && (status == 'excluded' || status == 'disabled')
```
hennevogel: This is getting complicated. Do you have another idea how to do this?

src/api/app/models/local_build_result/for_package.rb
```diff
@@ -33,8 +33,12 @@ def local_build_result(result, status)
...
     end
 
-    def excluded?(status)
-      !show_all && status == 'excluded'
+    def excluded_or_disabled?(status)
+      if show_all
+	return false
+      end
+	 
+      status == 'excluded' || status == 'disabled'
```
Ana06: what about `status.in?['excluded', 'disabled']`? :wink:

src/api/app/models/local_build_result/for_package.rb
```diff
@@ -33,8 +33,12 @@ def local_build_result(result, status)
                            code: status['code'], state: result['state'], details: status['details'])
     end
 
-    def excluded?(status)
-      !show_all && status == 'excluded'
+    def excluded_or_disabled?(status)
+      if show_all
```
Ana06: it is nice to write short conditions in one line: `return false if show_all`

src/api/app/models/local_build_result/for_package.rb
gsouzab: Went with `['excluded', 'disabled'].include?(status)`

hennevogel:
@gsouzab now if you squash the two commits, we are done! Hooray :tada::balloon:

hennevogel:
You can also see proof that it works, here: 
https://obs-reviewlab.opensuse.org/gsouzab-disabled_build_results/package/show/home:Admin/ruby

gsouzab:
Hi @Ana06 and @hennevogel, I was struggling a bit with the squash but I think I got it.

Ana06:
@gsouzab will you squash them? If not I will do it for you so that we can merge this :wink:

gsouzab:
Hi @Ana06, I agree with you. Makes sense to have only one commit in this case.
I'll squash them up.

I also want to change the following messages from **excluded** to **excluded/disabled**:
captura de tela de 2018-12-18 12-57-35

Do you agree?

Ana06:
@gsouzab 

> I also want to change the following messages from excluded to excluded/disabled:
> Do you agree?

Good idea :wink:

PR #6552, opened 2018-12-10
Improve influxdb

ChrisBr:
* Improve calculation of the backend location, not needed to get the whole stacktrace anymore
* Enable SQL tracking, it is disabled by default


ChrisBr:
Finally it doesn't hang anymore :pray: :pray: :pray: The issue was that I set the influxdb host in a test and it tried to send to influx db infinitely and made CircleCI hang :cry: I now just mock the config in the one test. A fix for the influx db is provided in https://github.com/openSUSE/open-build-service/pull/6570.

Thanks to @coolo and @eduardoj for leading me to the right track.

PR #6562, opened 2018-12-11
Ordinalize the dates on the project pulse page

dmarcoux:
I discovered this amazing method today: https://guides.rubyonrails.org/active_support_core_extensions.html#ordinalize

Also put the dates in an ascending order and added the year

Before:
before

After:
after


coolo:
why are the dates going backward?

dmarcoux:
@coolo: Good point, it is more natural when it's ascending. I will switch them.

dmarcoux:
Updated. I also added the year

PR #6565, opened 2018-12-11
[backend] adapt maintenance hack in package sorting to take multibuil…

mlschroe:
…d into account

We use the incident number at the end of the package name to sort
maintenance projects.  This did not work for multibuild builds, as
they have a trailing ":flavor" at the end.


coolo:
can we have a test for that? :)

PR #6569, opened 2018-12-11
Correct OBS logo.  Retrieved from homepage and optimized.

bberberov:
This version is half the size of the original without any obvious differences.  No paths, colors or transforms were changed.  HTH.


ChrisBr:
@bberberov thanks a lot. Can you please provide a before / after screenshot. Thanks!

bberberov:
Not sure what you are looking for.

Before:
screenshot_2018-12-11 opensuse open-build-service

After:
screenshot_2018-12-11 bberberov open-build-service

coolo:
As it was added 3 years ago 'as reference' and has always been wrong (https://github.com/openSUSE/open-build-service/commit/51eb39291133bb3b76e9d00e7aadfad95dbc1447#diff-39e7d2f851d53754dabe18f2be37299d), perhaps this should just be deleted? :)

bberberov:
I can open it on [openSUSE/obs-landing](https://github.com/openSUSE/obs-landing) if you think they'll care.  It's only 8KB difference out of 1.34MB on the front page.

PR #6572, opened 2018-12-11
Fix table width for inherited packages

bgeuken:
When using the 'w-*' modifier in one column but not the other
this can mess up the size of the other columns. In this case
it was squeezing the second column to ~1% width, which caused the text
to be shown vertical.


dmarcoux:
We also had this issue in another page (I don't remember which one though). I will update the patterns library to specify this.

PR #6578, opened 2018-12-12
Fix merge conflicts in staging-workflow (again)

coolo:
I would like to improve performance of staging projects, but don't dare too as there are too many changes in staging-workflow conflicting. So I hope to get this merged before we can do so.


coolo:
I pushed 2 rather trivial commits to fix things.

coolo:
The coverage dropped from 91.2% to 90.72% during the last 2 sprints. This is not a good trend :(

PR #6586, opened 2018-12-12
Refactor ProjectConfigurationController and related specs

vpereira:
* Add new service for the show action
* Adapt the specs to stub the services instead of ask the backend
* Remove unused VCR cassettes


src/api/spec/controllers/webui/projects/project_configuration_controller_spec.rb
```diff
@@ -46,7 +55,10 @@
 
     context 'cannot save a project config' do
       before do
-        allow_any_instance_of(ProjectConfigFile).to receive(:save).and_return(nil)
+        allow(::ProjectConfigurationService::ProjectConfigurationUpdater).to receive(:new) {
+          -> { OpenStruct.new(saved?: false, errors: 'yay') }
+        }
+        # allow_any_instance_of(ProjectConfigFile).to receive(:save).and_return(nil)
```
dmarcoux: This should be removed, right?

src/api/app/controllers/webui/projects/project_configuration_controller.rb
```diff
@@ -5,12 +5,14 @@ class ProjectConfigurationController < WebuiController
...
-        @content = @project.config.content(sliced_params.to_h)
+        result = ::ProjectConfigurationService::ProjectConfigurationPresenter.new(@project, params).call
         switch_to_webui2
-        return if @content
-        flash[:error] = @project.config.errors.full_messages.to_sentence
-        redirect_to projects_path(nextstatus: 404)
+        if result.valid?
+          @content = result.config
+        else
+          flash[:error] = @project.config.errors.full_messages.to_sentence
```
dmarcoux: `@project.config.errors.full_messages.to_sentence` is exactly what `result.errors` returns

src/api/app/services/project_configuration_service/project_configuration_presenter.rb
```diff
@@ -0,0 +1,30 @@
+module ProjectConfigurationService
+  class ProjectConfigurationPresenter
+    attr_reader :config
+    def initialize(project, params)
+      @project = project
+      @params = params
```
dmarcoux: [nitpick]

I think calling `call` here would make it nicer to use. `call` could then be private.

vpereira: @dmarcoux can you reformulate your question. I didn't understand it.

dmarcoux: The `initialize` method would be like this:
```
def initialize(project, params)
  @project = project
  @params = params
  call
end
```

And then, you would never need to use `call` directly since it's always needed anyway to use `call`.

vpereira: no @dmarcoux IMO that's wrong. Having just call exposed, allows me to mock the service object giving a proc to it, as I'm doing in my specs.. that's the way to go.

dmarcoux: Good point with mocking :+1:

PR #6590, opened 2018-12-12
Improve Package#show performance

ChrisBr:
InfluxDB performance measuring showed that this view is a bottleneck on Package#show with
a mean render time of ~ 280 ms (bootstrap) and 160 ms (bentoo). Therefore we
introduce now collection caching.

We also need to introduce timeago_tag jquery gem to be able to cache the table
otherwise the timeago would be cached as well causing wrong results.

selection_011

https://metrics.opensuse.org/d/Gpp3B1Pik/package-show-dashboard

### DIFF
--- a/src/api/app/views/webui2/webui/package/_files_view.html.haml
+++ b/src/api/app/views/webui2/webui/package/_files_view.html.haml
@@ -8,32 +8,8 @@
           %th Changed
           %th Actions
       %tbody
-        - files.each_with_index do |file, index|
-          - can_be_removed = removable_file?(file_name: file[:name], package: package) && User.current.can_modify?(package)
-          %tr{ id: "file-#{valid_xml_id(file[:name])}" }
-            %td.text-word-break-all
-              - link_opts = { action: :view_file, project: project, package: package, filename: file[:name], expand: expand }
-              - unless is_current_rev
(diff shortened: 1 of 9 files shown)


ChrisBr:
I also removed an unnecessary backend call now:


|Backend API | Count | Mean | Median|
|-----------------|---------|---------|----------|
|Backend::Api::Sources::Package#service | 4 K | 127.53 ms | 14.00 ms|

src/api/app/views/webui2/webui/package/_files_view.html.haml
```diff
@@ -8,32 +8,8 @@
...
-                = link_to(file_url(project.name, package.name, file[:name], file[:srcmd5]), title: 'Download file') do
-                  %i.fas.fa-download.text-secondary
-              - if can_be_removed
-                = link_to('#', data: { toggle: 'modal', target: "#delete-file-modal-#{index}" },
-                          title: 'Delete file') do
-                  %i.fas.fa-times-circle.text-danger
-          - if can_be_removed
-            = render(partial: 'delete_file_dialog',
-                      locals: { project: project.to_param, package: package.to_param, filename: file[:name], file_id: index })
+        = render partial: 'file', collection: files, cached: true,
```
coolo: how does this caching relate to User.current as used by the partials?

PR #6591, opened 2018-12-12
Solve autocompletion issues

bgeuken:

### DIFF
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -56,11 +56,11 @@ class Webui::ProjectController < Webui::WebuiController
   end
 
   def autocomplete_projects
-    render json: Project.autocomplete(params[:term]).not_maintenance_incident.order(:name).pluck(:name)
+    render json: Project.autocomplete(params[:term]).not_maintenance_incident.pluck(:name)
   end
 
   def autocomplete_incidents
-    render json: Project.autocomplete(params[:term]).maintenance_incident.order(:name).pluck(:name)
+    render json: Project.autocomplete(params[:term]).maintenance_incident.pluck(:name)
   end
 
   def autocomplete_packages
@@ -73,7 +73,7 @@ class Webui::ProjectController < Webui::WebuiController
   end
 
   def autocomplete_repositories
-    render json: @project.repositories.pluck(:name)
+    render json: @project.repositories.order(:name).pluck(:name)
   end
 
   def users
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -91,9 +91,7 @@ class Project < ApplicationRecord
     where.not('name rlike ?', ::Configuration.unlisted_projects_filter) if ::Configuration.unlisted_projects_filter.present?
   }
   scope :remote, -> { where('NOT ISNULL(projects.remoteurl)') }
-  scope :autocomplete, lambda { |search|
-    where('lower(name) like lower(?)', "#{search}%").where.not('lower(name) like lower(?)', "#{search}%:%")
-  }
+  scope :autocomplete, ->(search) { where('lower(name) like lower(?)', "#{search}%") }
 
   # will return all projects with attribute 'OBS:ImageTemplates'
   scope :local_image_templates, lambda {
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -31,7 +31,7 @@ class Project < ApplicationRecord
   has_many :relationships, dependent: :destroy, inverse_of: :project
   has_many :packages, inverse_of: :project do
     def autocomplete(search)
-      where(['lower(packages.name) like lower(?)', "#{search}%"])
+      where(['lower(packages.name) like lower(?)', "#{search}%"]).limit(50)
     end
   end
 
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -31,7 +31,7 @@ class Project < ApplicationRecord
   has_many :relationships, dependent: :destroy, inverse_of: :project
   has_many :packages, inverse_of: :project do
     def autocomplete(search)
-      where(['lower(packages.name) like lower(?)', "#{search}%"]).limit(50)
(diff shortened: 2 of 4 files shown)


src/api/app/controllers/webui/project_controller.rb
```diff
@@ -66,14 +66,14 @@ def autocomplete_incidents
   def autocomplete_packages
     @project = Project.find_by(name: params[:project])
     if @project
-      render json: @project.packages.autocomplete(params[:term]).limit(50).pluck(:name)
+      render json: @project.packages.autocomplete(params[:term]).limit(50).order(:name).pluck(:name)
```
Ana06: move the order to the scope...

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -56,24 +56,24 @@ def index
   end
 
   def autocomplete_projects
-    render json: Project.autocomplete(params[:term]).not_maintenance_incident.order(:name).pluck(:name)
+    render json: Project.autocomplete(params[:term]).not_maintenance_incident.order(:name).limit(50).pluck(:name)
```
Ana06: I think that limit the results to 50 is not intuitive. If the user is looking for a package that is not in this 50, this is more confusing that helping. I would better not show anything if there are more than 50, than just show the first 50.

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -56,24 +56,24 @@ def index
...
-      render json: @project.packages.autocomplete(params[:term]).pluck(:name)
+      render json: @project.packages.autocomplete(params[:term]).limit(50).pluck(:name)
     else
       render json: nil
     end
   end
 
   def autocomplete_repositories
-    render json: @project.repositories.pluck(:name)
+    render json: @project.repositories.limit(50).pluck(:name)
```
Ana06: That means there are some repositories you can not select :see_no_evil:  (at least in the repository page). You need some criteria here.

Also, how likely is that we have more than 50 repositories in a project? :eyes:

src/api/app/models/project.rb
```diff
@@ -89,9 +89,7 @@ def autocomplete(search)
     where.not('name rlike ?', ::Configuration.unlisted_projects_filter) if ::Configuration.unlisted_projects_filter.present?
   }
   scope :remote, -> { where('NOT ISNULL(projects.remoteurl)') }
-  scope :autocomplete, lambda { |search|
-    where('lower(name) like lower(?)', "#{search}%").where.not('lower(name) like lower(?)', "#{search}%:%")
-  }
+  scope :autocomplete, ->(search) { where('lower(name) like lower(?)', "#{search}%") }
```
Ana06: Removing the not search for `home:`?!?!? But it is great! :angry:

src/api/app/controllers/webui/project_controller.rb
bgeuken: If you ask that nicely... :wink:

src/api/app/controllers/webui/project_controller.rb
bgeuken: It's an autocomplete. I doubt users get confused by getting a sub-list of possible matches. And I also doubt users want to scroll through hundreds of results.

And like @coolo I was confused when I entered 'hom' in the old search and didn't got any results.

src/api/app/models/project.rb
bgeuken: Could you also provide an argument? And it's actually a not search in sub-projects. So if I want to branch `OBS:Server:Unstable:Foo` I would have to enter `OBS:Server:Unstable` until the autocomplete offers me `OBS:Server:Unstable:Foo`

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -56,24 +56,24 @@ def index
...
-      render json: @project.packages.autocomplete(params[:term]).pluck(:name)
+      render json: @project.packages.autocomplete(params[:term]).limit(50).order(:name).pluck(:name)
     else
       render json: nil
     end
   end
 
   def autocomplete_repositories
-    render json: @project.repositories.pluck(:name)
+    render json: @project.repositories.limit(50).order(:name).pluck(:name)
```
Ana06: As said already, that means there are some repositories you can not select :see_no_evil:  (at least in the repository page). You need some criteria here.

Also, how likely is that we have more than 50 repositories in a project? :eyes:

src/api/app/controllers/webui/project_controller.rb
Ana06: Can you please move it? This is not its place. :bowtie:

src/api/app/controllers/webui/project_controller.rb
coolo: https://build.opensuse.org/project/show/home:computersalat:devel:openSUSE_Tools has 77 :)

But this isn't really autocomplete - this is just all :(

src/api/app/controllers/webui/project_controller.rb
bgeuken: Already did, but seems I overlooked this one since it's [defined](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/project.rb#L33) within `has_many :packages` definition of the project model:unamused:

src/api/app/controllers/webui/project_controller.rb
bgeuken: Ouch, we should rename this. Dropped the limit again.

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -56,24 +56,24 @@ def index
   end
 
   def autocomplete_projects
-    render json: Project.autocomplete(params[:term]).not_maintenance_incident.order(:name).pluck(:name)
+    render json: Project.autocomplete(params[:term]).not_maintenance_incident.limit(50).pluck(:name)
```
Ana06: I think that the limit should also be part of the scope, as it part of the autocomplete feature. Can you move it (everywhere) please please? :pray:

PR #6592, opened 2018-12-12
Remove OBS label from breadcrumbs

bgeuken:
On each page the breadcrumb starts with 'openSUSE Build Service'.
This text is taking away quite some space from the breadcrumbs and it
is not adding much value. Users already know that they are looking at
the OBS when they reached this page.
In case they don't we still have the OBS logo in the navbar.

The link to the main page is now part of the 'home' icon that is
part of each breadcrumb.

-------------------

Depending on which page you are this looks a bit different. That's why I added a couple of screenshots:

On the main page:

selection_004

On the other pages:

selection_005

selection_006

### DIFF
--- a/src/api/app/views/webui2/webui/main/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui2/webui/main/_breadcrumb_items.html.haml
@@ -1,9 +1,5 @@
 - home_title = @configuration ? @configuration['title'] : 'Open Build Service'
-- if current_page?(root_path)
-  %li.breadcrumb-item.active{ 'aria-current' => 'page' }
-    %i.fas.fa-home.mr-1
-    = home_title
-- else
-  %li.breadcrumb-item
-    %i.fas.fa-home.mr-1
-    = link_to home_title, root_path
+- unless current_page?(root_path)
+  %li.breadcrumb-anchor
+    = link_to(root_path, title: home_title) do
+      %i.fas.fa-home.mr-2
--- a/src/api/app/assets/stylesheets/webui2/breadcrumbs-component.scss
+++ b/src/api/app/assets/stylesheets/webui2/breadcrumbs-component.scss
@@ -8,4 +8,11 @@ ol.breadcrumb {
       color: $gray-600;
     }
   }
+
+  & li.breadcrumb-anchor {
+    font-size: 0.9rem;
+    & a {
+      color: $gray-600;
+    }
+  }
 }


src/api/app/views/webui2/webui/main/_breadcrumb_items.html.haml
```diff
@@ -5,5 +5,5 @@
     = home_title
 - else
   %li.breadcrumb-item
-    %i.fas.fa-home.mr-1
-    = link_to home_title, root_path
+    = link_to root_path do
```
dmarcoux: I would add a title here, perhaps `'Homepage'` or `home_title`

src/api/app/views/webui2/webui/main/_breadcrumb_items.html.haml
```diff
@@ -5,5 +5,5 @@
     = home_title
```
dmarcoux: It's somewhat weird now, because on the homepage, this is what we see:

homepage

And on the other pages:
project-show

I think `home_title` should also be removed here. There should also be a title as I said in my other comment.

bgeuken:
As you can see on the main page the 'Open Build Service' label is right next to the 'house' icon, while on the other pages the 'house' and the rest of the breadcrumb is divided by the '/' sign we use for dividing the breadcrumb parts.

IMO this is fine, since the house is the main page. Thus we don't have to divide the two elements on the main page, while on the other pages they represent different things.

@openSUSE/obs-frontend What do you think?

src/api/app/views/webui2/webui/main/_breadcrumb_items.html.haml
bgeuken: The whole point was to not add a title here. Though we could add a tooltip.

src/api/app/views/webui2/webui/main/_breadcrumb_items.html.haml
bgeuken: I added some reasoning in my other comment. I wanted to get some feedback first before changing this further.

src/api/app/views/webui2/webui/main/_breadcrumb_items.html.haml
dmarcoux: Sorry I wasn't clear, I was referring to the `title` attribute of the `a` tag generated by `link_to`

eduardoj: I'm fine with @bgeuken's approach. I would add the tooltip.

DavidKang:
@bgeuken, and if we remove the breadcrumb only on the main page?

bgeuken:
@DavidKang @dmarcoux @Ana06 Updated and I removed the breadcrumb from the main page as suggested by David. I've updated the screenshots as well. Have a look

PR #6593, opened 2018-12-12
Use a default icon for avatar loading error

### LINKED ISSUE #6183: [beta] Comments section squishes authors
Names of comment's authors are squished. Maybe move them where there are nick handles and use them there instead?

For example, see,

https://build.opensuse.org/package/show/devel:libraries:c_c++/boost

screen

dmarcoux:
Fixes #6183

Whenever an avatar is not loading somehow, we replace its `img` tag by a default icon. We keep the same classes as the avatar to be consistent.

I'm not entirely sure if this is the best approach, but I don't really have another in mind. If you validate this approach, I will move the Javascript to the assets and use it in the other pages where we also have avatars. This would be all done in this PR.

Example:
preview

### DIFF
--- a/src/api/app/views/webui2/webui/comment/_content.html.haml
+++ b/src/api/app/views/webui2/webui/comment/_content.html.haml
@@ -17,3 +17,9 @@
 - if level > 3
   - comment.children.each do |children|
     = render partial: 'webui2/webui/comment/content', locals: { comment: children, commentable: commentable, level: level + 1 }
+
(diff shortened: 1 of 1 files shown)


hennevogel:
http://en.gravatar.com/site/implement/images/

dmarcoux:
If Gravatar is blocked, then using a default image from their API won't work anyway, right? So it is not a possible solution.

PR #6596, opened 2018-12-12
Remove table info from Package datatable

ChrisBr:
because the number of packages is already included in the tab
and this removes some noise from the view.

selection_015

Old:

selection_016


bgeuken:
We might want to do the same on the user tasks page

ChrisBr:
> We might want to do the same on the user tasks page

Right! Added for the user page. For the tasks I need to investigate first how it works there ...

PR #6599, opened 2018-12-12
webui: use system monospace font for patch box again

jengelh:
Undo commit e31f2ca36b ("Use 'Liberation Mono' font for code
blocks"). I want my system's preference to be used again. Using
Liberation only there is also weird, because other places with
preformatted text, like the request review comment box, or the
package build log, still use the default monospace.

The "disappearing underscore" problem is caused by the line-height
directive, and *also* manifests in Chromium(!), starting with a
line-height of 105%. It turns out browsers have this weird idea that
the default line-height is "120%" and, as a result, any value less
than that shrinks the box. Throw out line-height for good.

### DIFF
--- a/src/api/app/assets/stylesheets/webui/application/cm2/suse.scss
+++ b/src/api/app/assets/stylesheets/webui/application/cm2/suse.scss
@@ -1,5 +1,5 @@
 .CodeMirror {
-  font-family: "Liberation Mono", monospace;
+  font-family: monospace;
   font-size: 9pt;
   height: auto;
 }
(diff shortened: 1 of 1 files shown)


coolo:
looks like I introduced the line height in https://github.com/openSUSE/open-build-service/commit/3951cf5333c45b22b77e733e3d68f5a90f4a4c1e 6 years ago without much of a reasoning, but I do remember the struggles with underscores.

It should be noted though that in webui2 there is a different setting of 1.2 introduced by @hellcp in a11cadfea8 - so you're fixing a dying UI anyway :)

hellcp:
We can always reset the value there, but I wanted override to be clear instead of hardcoding something dumb specifically for CM.

Line height is imprecise, but any other setting caused lines to either overlap or have a line between them (and I set background behind diff files, so it was kind of important to get the setting right so that part looked good).

PR #6600, opened 2018-12-13
Fix package file list cache

ChrisBr:
We need to use a custom cache key including the locals otherwise the cache will be stale.

Follow up of https://github.com/openSUSE/open-build-service/pull/6590

image

### DIFF
--- a/src/api/app/views/webui/package/_file.html.erb
+++ b/src/api/app/views/webui/package/_file.html.erb
@@ -11,12 +11,12 @@
   <td><span class="hidden"><%= file[:mtime] %></span><%= timeago_tag Time.at(file[:mtime].to_i) %>
   </td>
   <!-- limit download for anonymous user to avoid getting killed by crawlers -->
-  <td><%= if !User.current.is_nobody? || file[:size].to_i < (4 * 1024 * 1024)
+  <td><%= if !nobody || file[:size].to_i < (4 * 1024 * 1024)
             link_to sprite_tag('page_white_put', title: 'Download File'),
                     file_url(project.name, package.name, file[:name], file[:srcmd5])
           end %>
     <% if removable_file?(file_name: file[:name], package: package) %>
-        <% if User.current.can_modify? package %>
+        <% if can_modify %>
             <%= link_to sprite_tag('page_white_delete', title: 'Remove File'), {:action => :remove_file, :project => project.to_param,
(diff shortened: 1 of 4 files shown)


src/api/app/views/webui/package/_files_view.html.erb
```diff
@@ -9,7 +9,12 @@
       </tr>
       </thead>
       <tbody>
-        <%= render partial: 'file', collection: @files, cached: true, locals: { project: @project, package: @package, expand: @expand } %>
+        <%- file_locals = {
+          project: @project, package: @package, expand: @expand, is_current_rev: @is_current_rev,
+          can_modify: User.current.can_modify?(@package), nobody: User.current.is_nobody?
+        } %>
+        <%= render partial: 'file', collection: @files, cached: proc { |file| [file, file_locals].hash }, locals: file_locals
+        %>
```
coolo: about time the bento pages for packages and projects find their rest :(

ChrisBr: What do you mean?

coolo: the whole theory behind "let's keep bento and bootstrap alive in parallel for a while" was based on the assumption that every developer can convert one action a day. As this assumption failed, my recommendation would be to give up and make project and package pages bootstrap only (risking a mixed environment for everyone) - and continue with beta flag only for pages under sprint development.

But that's unrelated to this PR - it just makes every fix in views a little more complicated

bgeuken:
@ChrisBr Sorry, I saw this one to late. Though our packages are anyway in 'unresolvable' state right now.

Deployed state right now is 8536dcc77096dc55fd570ef14989830a40760df1

src/api/app/views/webui/package/_file.html.erb
```diff
@@ -11,12 +11,12 @@
   <td><span class="hidden"><%= file[:mtime] %></span><%= timeago_tag Time.at(file[:mtime].to_i) %>
   </td>
   <!-- limit download for anonymous user to avoid getting killed by crawlers -->
-  <td><%= if !User.current.is_nobody? || file[:size].to_i < (4 * 1024 * 1024)
+  <td><%= if !nobody || file[:size].to_i < (4 * 1024 * 1024)
```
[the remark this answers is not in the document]
ChrisBr: We can do that in a follow up as it is unrelated to this change!

PR #6610, opened 2018-12-13
We make Repositories quick

### LINKED ISSUE #6595: Ana is too polite to say, but the repositories flag code is horribly slow
I'm not Ana, but we're looking into it :)

coolo:
Stop using the flag model to calculate the flag matrix. Calculating the icon state per column is very wasteful as it does nr(repository)*nr(architecture)*15 SQL queries.

Replace it with one query for all flags, store it in a has and calculate the state in ruby - this is one SQL query and some hash lookups per icon.

This saves for openSUSE:Tools ~75% in render time (~30 seconds). :bowtie: :nerd_face: 

There are many other things that can be improved, make more efficient and refactor, but that will come in another PR. But first things first, now let focus on the big efficiency problem :wink: 

Co-authored-by: @Ana06 

For package view it's even worse 
(on basically empty database and only an import of meta of openSUSE:Tools):

**master:**
`Completed 200 OK in 48149ms (Views: 29610.1ms | ActiveRecord: 3936.5ms | Backend: 0.0ms)`

**our branch:**
`Completed 200 OK in 5817ms (Views: 5609.6ms | ActiveRecord: 14.1ms | Backend: 0.0ms)`

Fixes #6595


Ana06:
@openSUSE/open-build-service this is fixing an important efficiency issue. Please review so that we can merge it. :pray:

src/api/app/helpers/webui/repository_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def html_id_for_flag(flag_type, repository, architecture)
+    # repository and architecture can be nil
+    valid_xml_id("flag-#{flag_type}-#{repository}-#{architecture}")
+  end
+
+  # TODO: This should be private
+  def dig_flag(callback_function, flags, repository, architecture)
+    flag = send(callback_function, flags[:object], repository, architecture)
+    flag ||= send(callback_function, flags[:project], repository, architecture) if flags[:project]
```
ChrisBr: The callback function is always specified by us or is it possible for the user?

coolo: We can't make functions in modules private, but it's just a helper function for default_flag and effective_flag

saraycp:
The flags functionality is a bit complex for me to ensure that this refactoring is perfect. But I don't see any weird thing to comment... apart from GOOD WORK!! :tada: Thanks for working on this!!

Issue #6601, opened 2018-12-13
Staging dashboard no longer shows "Ready" requests


nilxam:
# Issue Description
Staging dashboard usually shows those `state=new` requests in the Ready block, seems a recent change breaks it, now those requests no longer show up on the dashboard.

eg. Ready block on https://build.opensuse.org/project/staging_projects/openSUSE:Leap:15.1

Expected Result
============
Those requests should show up there.

How to Reproduce
===============
1. Go to 'https://build.opensuse.org/project/staging_projects/openSUSE:Leap:15.1'
2. See `Ready` block, there is Empty but click the link, there are many.

PR #6611, opened 2018-12-13
Fix ready list in obs-factory staging dashboard

saraycp:
Since last refactor on this code, the Ready list was always empty. It
was taking request by reviews with 'new' state, whereas, in this case,
the review shouldn't be fixed to any state.

Fixes: #6601


src/api/app/models/bs_request.rb
```diff
@@ -80,6 +80,9 @@ class SaveError < APIError
   scope :with_target_project, lambda { |target_project|
     includes(:bs_request_actions).where('bs_request_actions.target_project': target_project)
   }
+  scope :new_with_reviews_for, lambda { |review_attributes|
+    where(state: 'new').where(id: Review.where(review_attributes).pluck(:bs_request_id)).includes(:reviews)
```
coolo: a select(:bs_request_id) would do and reduces this to one query instead of 2.

PR #6612, opened 2018-12-13
Add repository status to Monitor page columns

### LINKED ISSUE #6567: monitor page looses the icons in front of repository name
Those icons e.g. tell me things like

- "repository not created yet"
- "publishing"
- "published"

All this is lost after the migration to bootstrap.

Before:
example

Now:
now

ChrisBr:
Fix #6567

### DIFF
--- a/src/api/app/views/webui2/webui/project/monitor.html.haml
+++ b/src/api/app/views/webui2/webui/project/monitor.html.haml
@@ -21,9 +21,10 @@
                   = repo
             %tr
               %th
-              - @repohash.sort.each do |_repo, archlist|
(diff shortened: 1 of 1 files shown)


coolo:
I prefer small

ChrisBr:
I also prefer small, updated!

ChrisBr:
> small doesn't work here. You want fa-xs. Otherwise it looks like this

Hm, in chrome it worked for me :thinking: :cry: But thanks, changed...

ChrisBr:
selection_017

Added also a margin now.

PR #6613, opened 2018-12-13
Update influxdb-rails to master branch

ChrisBr:
To not store cache_hits and counts for render_collection series anymore.
These values have a high cardinality and therefore not usable as tags
which causes high load on metrics.o.o due to constantly indexing.

https://github.com/influxdata/influxdb-rails/pull/63

### DIFF
--- a/src/api/Gemfile
+++ b/src/api/Gemfile
@@ -77,7 +77,7 @@ gem 'addressable'
 # for XML builder
 gem 'builder'
 # to write the rails metrics directly into InfluxDB.
-gem 'influxdb-rails', '>=1.0.0.beta2'
+gem 'influxdb-rails', source: 'https://github.com/influxdata/influxdb-rails.git', branch: 'master', ref: '96b7eb18ee6b242b8f9cb99089e9b75c02f71bbe'
 # for client side time ago
 gem 'rails-timeago', '~> 2.0'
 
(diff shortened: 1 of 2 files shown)


coolo:
Or apply the patch in bundled-gems.spec - but the gem service doesn't like git it seems.

ChrisBr:
> Nice fix in the gem tada , but the packages seems to not have permissions to download gems:

Hm, that's weird, it should work as this was one of the reasons why we implemented the ``cpio`` mode based on bundler in the gem service :thinking: I will talk with @M0ses about that as we should fix it ...

> I am also not sure if we should download it. What about opening a new release wanted/needed issue in influxdb-rails?

I already talked with the maintainer but he wants to add more features before releasing a new version ...

ChrisBr:
> Hm, that's weird, it should work as this was one of the reasons why we implemented the cpio mode based on bundler in the gem service

Ok, I changed now to use an URL instead of git, this is anyway better because ``github:`` does not support SSL (https://bundler.io/guides/git.html#security). Furthermore I locked the commit now.

ChrisBr:
I debugged with @M0ses and it seems that the service container only has network to the geminabox container but not to the internet. We will change this next year but for now it is not supported ... Let's wait for the release next week, closing...

PR #6614, opened 2018-12-13
Webui2 helper

bgeuken:


src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -61,6 +61,7 @@ def format_projectname(prjname, login)
     prjname
   end
 
+  # bento_only
```
Ana06: I would add a `TODO delete when dropping bento` here (so that we now there is something to do at some point :wink: )

bgeuken: It's rather meant to be a tag that can be `grepped` for than a message for humans to read. So it doesn't have to explain much. And we all know that bento goes away soon anyway.

I documented the tag, and the steps needed after the UI migration is done, in our [wiki](https://github.com/openSUSE/open-build-service/wiki/How-to-work-in-Bootstrap-%28new-UI%29#things-to-do-after-the-migration).

Ana06: This is not something widely used. We need to find some agreement, otherwise it will be a nightmare to get rid of all this code. And I think that adding a TODO (with an explanation) is also a must. In case some of this code stays by accident, someone may eventually realize.

bgeuken: I would prefer to grep for 'bento_only' over 'TODO'. The latter would return additional results.

> In case some of this code stays by accident, someone may eventually realize.

That is why I would like to have some tag that can be parsed easily and is unlikely to be mistyped. Whether there is some additional explanation next to it I don't mind.

coolo: how about TODO: bento_only? :)

In 6610 Ana chose 'TODO: used by bento. Remove when dropping old UI.' (and I didn't care :)

bgeuken: Fine with me. Then `TODO: bento_only` ^^

Ana06: It is ok, but who is going to check it is done like that everywhere (it is not). Does everybody know this? Anyway, not a reason to block the PR.

PR #6619, opened 2018-12-14
Fix flag validation catching duplicates

### LINKED ISSUE #6615: wrong Flag validations break logic
We shouldn't allowed to have more than one Flag for the same repository, type (confusingly known as flag), architecture, project/package, independent from the status. We are even failing to catch the cases with the same status (which the validation is not allowing but it is only run on create). 

That is breaking the logic of the repositories page (remove flag doesn't expect another flag to exist), may be doing something different as what rendered (frontend takes the first but the backend may not) and it makes no sense to have more than one flag. :see_no_evil: 

There are many cases in our production instance were this happens (I would say that at least in some of them not intentionally), so this requires data migration.

coolo:
This has been broken since the introduction as
where('repo = ?', repo) for nil values compares with NULL,
which never exists. On top of that this also checked if
status was duplicated, which means you could have an
enabled and disabled status for the same field, which doesn't
make much sense :bowtie: 

We have a lot of duplicated flags by now, so we need
a data migration :eyes: 

Co-authored-by: @Ana06 

Fixes #6615

### DIFF
--- a/src/api/app/models/flag.rb
+++ b/src/api/app/models/flag.rb
@@ -24,19 +24,12 @@ class Flag < ApplicationRecord
     errors.add(:status, 'Status needs to be enable or disable') unless status && (status.to_sym == :enable || status.to_sym == :disable)
   end
 
-  validate :validate_duplicates, on: :create
+  validate :validate_duplicates
   def validate_duplicates
-    flag_exists = Flag.where(
-      'status = ? AND repo = ? AND project_id = ? AND package_id = ? AND architecture_id = ? AND flag = ?',
-      status,
-      repo,
-      project_id,
-      package_id,
-      architecture_id,
-      flag
-    ).exists?
-
-    errors.add(:flag, 'Flag already exists') if flag_exists
+    flags = Flag.where(repo: repo, project_id: project_id, flag: flag,
+                       package_id: package_id, architecture_id: architecture_id)
+    flags = flags.where.not(id: id) unless new_record?
+    errors.add(:flag, 'Flag already exists') if flags.exists?
   end
 
   def self.default_status(flag_name)
--- a/src/api/app/models/flag.rb
+++ b/src/api/app/models/flag.rb
@@ -24,13 +24,7 @@ class Flag < ApplicationRecord
     errors.add(:status, 'Status needs to be enable or disable') unless status && (status.to_sym == :enable || status.to_sym == :disable)
   end
 
-  validate :validate_duplicates
-  def validate_duplicates
-    flags = Flag.where(repo: repo, project_id: project_id, flag: flag,
-                       package_id: package_id, architecture_id: architecture_id)
-    flags = flags.where.not(id: id) unless new_record?
-    errors.add(:flag, 'Flag already exists') if flags.exists?
-  end
+  validates :flag, uniqueness: { scope: [:project_id, :package_id, :architecture_id, :repo] }
 
   def self.default_status(flag_name)
     FlagHelper.default_for(flag_name)
--- a/src/api/db/data/20181214100207_remove_duplicated_flags.rb
+++ b/src/api/db/data/20181214100207_remove_duplicated_flags.rb
@@ -1,0 +1,22 @@
+class RemoveDuplicatedFlags < ActiveRecord::Migration[5.2]
+  def up
+    attributes = [:flag, :repo, :architecture_id, :package_id, :project_id]
+    # whenever 2 flags are for the same 'thing' (in _meta), we remove all but one
+    last_flags = Flag.group([:status] + attributes).select('MAX(id) as id')
+    # we need to pluck the ids (even though it's a lot) as you can't delete
+    # from within group queries
+    Flag.where.not(id: last_flags).in_batches(&:delete_all)
+
+    # now comes the ugly part - we need to delete conflicting flags. If both
+    # 'enable' and 'disable' are set, the backend prefers disable - so we can
+    # safely remove the 'enable' for those where we have 2
+    Flag.group(attributes).select(attributes).having('count(id) > 1').each do |f|
+      Flag.where(flag: f.flag, repo: f.repo, architecture_id: f.architecture_id,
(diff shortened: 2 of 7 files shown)


coolo:
@Ana06 I double checked on production data again

Before:
frontend@53ecc24defcd:/obs/src/api> rails runner 'puts Flag.count'
431737

frontend@53ecc24defcd:/obs/src/api> rails runner 'puts Flag.count'
412655

(the migration runs about 4 minutes, possibly we should make the batches larger than the default 1000)

coolo:
no idea

Ana06:
@coolo, the co-author line has to be at the end. :wink: Now I am there :tada: 

@openSUSE/open-build-service please review!

src/api/app/models/flag.rb
```diff
@@ -24,19 +24,12 @@ def validate_custom_save
...
-      project_id,
-      package_id,
-      architecture_id,
-      flag
-    ).exists?
-
-    errors.add(:flag, 'Flag already exists') if flag_exists
+    flags = Flag.where(repo: repo, project_id: project_id, flag: flag,
+                       package_id: package_id, architecture_id: architecture_id)
+    flags = flags.where.not(id: id) unless new_record?
```
bgeuken: How about [validates :foo, uniqueness: true](https://guides.rubyonrails.org/active_record_validations.html#uniqueness)?

src/api/db/data/20181214100207_remove_duplicated_flags.rb
```diff
@@ -0,0 +1,24 @@
+class RemoveDuplicatedFlags < ActiveRecord::Migration[5.2]
+  def up
+    attributes = [:flag, :repo, :architecture_id, :package_id, :project_id]
+    # whenever 2 flags are for the same 'thing' (in _meta), we remove all but one
+    last_flags = Flag.group([:status] + attributes).select('MAX(id) as id')
+    # we need to pluck the ids (even though it's a lot) as you can't delete
+    # from within group queries
+    Flag.where.not(id: last_flags).in_batches do |batch|
+      Flag.where(id: batch.pluck(:id)).delete_all
```
bgeuken: Instead of running the query twice I would do this:
```ruby
Flag.where.not(id: last_flags).in_batches.delete_all
```

src/api/db/data/20181214100207_remove_duplicated_flags.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    # now comes the ugly part - we need to delete conflicting flags. If both
+    # 'enable' and 'disable' are set, the backend prefers disable - so we can
+    # safely remove the 'enable' for those where we have 2
+    Flag.group(attributes).select(attributes).having('count(id) > 1').each do |f|
```
bgeuken: I would process this [in batches](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each) as well.

src/api/db/data/20181214100207_remove_duplicated_flags.rb
```diff
@@ -0,0 +1,24 @@
...
+      Flag.where(flag: f.flag, repo: f.repo, architecture_id: f.architecture_id,
+                 package_id: f.package_id, project_id: f.project_id, status: 'enable').delete_all
```
bgeuken: It looks like you already have the flags you want. Why don't you just do this instead:
```ruby
flags = Flag.group(attributes).select(:id).having('count(id) > 1')
Flag.where(id: flags, status: 'enable').in_batches.delete_all
```
?

src/api/db/data/20181214100207_remove_duplicated_flags.rb
coolo: did you read the comment?

src/api/app/models/flag.rb
coolo: what's foo?

src/api/db/data/20181214100207_remove_duplicated_flags.rb
coolo: because it won't work to select ids of groups - there is always only one row per group returned. And that is not necessarly the enabled one.

src/api/app/models/flag.rb
bgeuken: Like this (untested):

```ruby
validates :repo, uniqueness: { scope: [:project_id, :package_id, :architecture_id]
```

src/api/db/data/20181214100207_remove_duplicated_flags.rb
bgeuken: But calling `last_flags` within the `Flag.where...` should execute the group query, or not? And if rails chains the two queries together you probably could prevent this by triggering the "group" query via `last_flags.pluck(:id)`.

coolo: can you imagine how many last_flags we have and how large that plucked array would be? There is a reason this is done in batches, because it's *a lot*

bgeuken: Okay Okay, you convinced me:wink:

src/api/db/data/20181214100207_remove_duplicated_flags.rb
bgeuken: Okay, I see

src/api/db/data/20181214100207_remove_duplicated_flags.rb
coolo: hmm, I don't think the way in batches works is compatible with groups - you wouldn't be able to fin d duplicates that are in between 2 batches.

src/api/app/models/flag.rb
coolo: I did this now - let's see

coolo:
@bgeuken I fixed your test - now all is green :)

Ana06:
Oh no! :green_heart: is gone...

Ana06:
:dancer:

PR #6620, opened 2018-12-14
Refactor Meta controller related specs

vpereira:
- Write specs for the ```ProjectUpdater``` object service
- Delete ```MetaController``` specs which were testing others components than the controller
- Remove unused cassettes


src/api/spec/services/meta_controller_service/project_updater_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      <project name="#{project.name}">
+      <title/><description/>
+      </project>
+      HEREDOC
+    end
+
+    let(:project_updater) { ::MetaControllerService::ProjectUpdater.new(project: project, request_data: Xmlhash.parse(meta)) }
+
+    describe '#new' do
+      it { expect(project_updater).not_to be_nil }
```
dmarcoux: This is testing that the constructor works? I don't really see why it's needed...

src/api/spec/services/meta_controller_service/project_updater_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      <path project="not-existent" repository="standard" />
+      </repository>
+      </project>
+      HEREDOC
+    end
+
+    let(:project_updater) { ::MetaControllerService::ProjectUpdater.new(project: project, request_data: Xmlhash.parse(meta)) }
+
+    describe '#new' do
+      it { expect(project_updater).not_to be_nil }
```
dmarcoux: Same thing about, I don't think it's needed to test `#new`

src/api/app/services/meta_controller_service/project_updater.rb
```diff
@@ -10,13 +10,14 @@ def call
...
         @errors = @validator.errors
-        return
+        return self
       end
 
       Project.transaction do
         @errors = @project.update_from_xml(@request_data)[:error]
         @project.store if valid?
       end
+      self
```
bgeuken: Please add a reason for the change to the commit message

src/api/spec/controllers/webui/projects/meta_controller_spec.rb
```diff
@@ -26,6 +26,10 @@
     context 'with a valid project' do
       context 'without a valid meta' do
         before do
+          allow(::MetaControllerService::ProjectUpdater).to receive(:new) {
+              -> { OpenStruct.new(valid?: false, errors: 'yada') }
+          }
+
```
bgeuken: Shouldn't this include some dropped VCR cassettes? :thinking:

src/api/spec/services/meta_controller_service/project_updater_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    describe '#call' do
+      it { expect { project_updater.call }.not_to raise_error }
```
bgeuken: This makes me wonder when this method would raise an exception and which one. I don't see any test for this.

src/api/spec/services/meta_controller_service/project_updater_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    describe '#valid?' do
+      let(:project_updater_call) { project_updater.call }
+      before do
+        project_updater_call
+      end
```
bgeuken: That's a bit repetitive. I would drop the `let`. IMO `let`s shouldn't contain method calls anyway (except constructors).

src/api/spec/services/meta_controller_service/project_updater_spec.rb
bgeuken: The constructor is written by us. So it makes sense to test it. But `not_to be_nil` could be anything. I would either be more specific or drop the test.

src/api/app/controllers/webui/projects/meta_controller.rb
```diff
@@ -12,8 +12,7 @@ def show
 
       def update
         authorize @project, :update?
-        updater = ::MetaControllerService::ProjectUpdater.new(project: @project, request_data: @request_data)
-        updater.call
+        updater = ::MetaControllerService::ProjectUpdater.new(project: @project, request_data: @request_data).call
 
         status = if updater.valid?
```
Ana06: why not using `@errors` in here? You can remove the updater variable and the changes to return self :nerd_face:

src/api/spec/services/meta_controller_service/project_updater_spec.rb
vpereira: > That's a bit repetitive. I would drop  if i can agree with thathe `let`. IMO `let`s shouldn't contain method calls anyway (except constructors).

I'm not sure, looks like a normal pattern to do something like:

```
let(:post_save_meta) { post :update, params: { project_name: 'nonexistent_project' }, xhr: true }
```

which normally for me looks good, and it's a method call. What do you think @bgeuken ? Said that I agree that in this situation is kind of superfluous.

src/api/spec/services/meta_controller_service/project_updater_spec.rb
vpereira: looking the code you will that is a transaction. https://github.com/vpereira/open-build-service/blob/master/src/api/app/services/meta_controller_service/project_updater.rb#L16

src/api/app/controllers/webui/projects/meta_controller.rb
vpereira: squashed @Ana06

src/api/app/services/meta_controller_service/project_updater.rb
vpereira: done, please check @bgeuken

src/api/spec/services/meta_controller_service/project_updater_spec.rb
bgeuken: Check that the created instance is of the expected class (the [project updater](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/services/meta_controller_service/project_updater.rb#L3) seems to handle this and that the request data was assigned.

src/api/app/services/meta_controller_service/project_updater.rb
bgeuken: Thanks. There are a few typos you might want to fix:-)

src/api/spec/services/meta_controller_service/project_updater_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    describe '#new' do
+      it { expect(project_updater).not_to be_nil }
+    end
+
+    describe '#call' do
+      it { expect { project_updater.call }.not_to raise_error }
+    end
+
+    describe '#valid?' do
+      let(:project_updater_call) { project_updater.call }
```
hennevogel: Is there a reason to sometimes do the `project_updater.call` directly and sometimes not?

src/api/spec/services/meta_controller_service/project_updater_spec.rb
bgeuken: IMO you can drop this test. The test for `valid?` already ensures that `call` does not raise an exception and since there isn't any particular exception we would expect here I don't see that point of having a test for this.

src/api/spec/services/meta_controller_service/project_updater_spec.rb
vpereira: > Might happen that a value comes from a method call and thus it's wrapped in a `let`.

if you see, ```call``` returns ```self``` and then i test the instances methods on that.. so i think its inline with what you are saying.

src/api/spec/services/meta_controller_service/project_updater_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      it { expect(project_updater).to be_valid }
```
bgeuken: This won't work because `project_updater_call` is never called. Just use `before` :wink:

src/api/app/controllers/webui/projects/meta_controller.rb
Ana06: @vpereira 

> I couldn't visualize what you are proposing.. could you please show me in code? :)

I can add a commit to this PR with the suggestion. Is that fine for you?

Ana06: There are several things here:

-  I don't understand what you gain making that `call` return  `self`. It has to be more expensive than returning `nil` and I don't see you get any improvement from it. Is it for the tests? If that is the case, I don't think you should change the code (change the tests to fit the code instead).
- You could return true or false and then you don't need to call `valid?`. Alternatively you could stop setting `@errors` in `call` (is not used) and leave the rest of the code as it was. [This refactoring would bring a small improvement, but if it only about the tests not related]
- `valid?` method is weird, why is it checking `@errors.blank?` [maybe also not the focus of the PR]

vpereira: @Ana06 thank you for your inputs. It's one way to design it. There is another way to have ```call``` as class method, which in turn, it calls ```new``` and return self anyway. 

For instance:

```
class CalculateVendorPayoutService
  def new(params)
    ...
  end

  def self.call(params)
    new(params)
  end
end
```

So because we used already this pattern in another place https://github.com/openSUSE/open-build-service/blob/master/src/api/app/services/project_configuration_service/project_configuration_updater.rb#L9  I think thats ok to keep like that.

src/api/spec/services/meta_controller_service/project_updater_spec.rb
vpereira: ops, done!

src/api/spec/services/meta_controller_service/project_updater_spec.rb
vpereira: nitpick, no @hennevogel ?

src/api/spec/services/meta_controller_service/project_updater_spec.rb
vpereira: I removed it completely @dmarcoux  and @bgeuken

src/api/spec/services/meta_controller_service/project_updater_spec.rb
vpereira: done @dmarcoux

src/api/app/services/meta_controller_service/project_updater.rb
vpereira: done @bgeuken, please review it again

src/api/spec/services/meta_controller_service/project_updater_spec.rb
vpereira: did it @bgeuken

src/api/spec/services/meta_controller_service/project_updater_spec.rb
vpereira: yup, because to test ```valid?``` I have to call ```call```

src/api/spec/services/meta_controller_service/project_updater_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    describe '#valid?' do
+      let!(:project_updater_call) { project_updater.call }
+
+      it { expect(project_updater).not_to be_valid }
+    end
+
+    describe '#errors' do
+      let!(:project_updater_call) { project_updater.call }
+
```
bgeuken: I would...
* remove the newline to have the `let`s grouped
* move `let!(:project_updater_call) { project_updater.call }` up to `context 'with nonexistent path'` since it's used in every `describe` block
* convert the `let` to a `before` block since it's just a method call. (IMO `subject` wouldn't be correct since we actually test `project_updater`, the object). It's just coincident that it returns the updater object. But I am splitting hairs, both `before` and `subject` would fit better than `let`.

That said, I give up:-)

vpereira: done @bgeuken. Please review

PR #6622, opened 2018-12-14
Remove expand_flags from project

coolo:
The function is only used in tests and it's a lot of complicated code to be only used in tests


coolo:
Extended in https://github.com/openSUSE/open-build-service/pull/6626 - but you can review this one first if you prefer

PR #6631, opened 2018-12-17
RFC: Use obs-bundled-gems in docker file

coolo:
We package them anyway, so the bundle install is quick

### DIFF
--- a/src/api/docker-files/Dockerfile
+++ b/src/api/docker-files/Dockerfile
@@ -27,7 +27,7 @@ USER frontend
 WORKDIR /obs/src/api
 
 # Refresh our bundle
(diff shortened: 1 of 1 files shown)


coolo:
The reason I use this workflow is because this way I can install the gems package in my tumbleweed and run rake task.

But what you ignore is that bundle install does more than downloading - it compiles quite some extensions. But I'm aware there are 2 sides of this medal, so I put it up for comments.

hennevogel:
> The reason I use this workflow is because this way I can install the gems package in my tumbleweed and run rake task.

What is keeping you from using docker-compose?

coolo:
it doesn't plug well in my editor

PR #6632, opened 2018-12-17
WIP circleci: Use precompiled assets in minitest too

coolo:


coolo:
ok, it reproducible hangs in FullTextSearchTest - had the same thing last time I tried, but I had hope it would have been the influxdb changes back then :)

When using ssh to go into circleci, there is nothing really going on, so I have no idea what's causing this.

Issue #6124, opened 2018-10-21
Reports API is showing irrelevant infos


coolo:
osc api /status_reports/published/openSUSE:Leap:15.1:Staging:C/images/reports/1539197354 (this is the current build id but any value will do including 'foobar') gives me:

```
<status_report uuid="1539197410">
  <check name="minimalx@smp_64" required="false">
    <state>success</state>
    <short_description/>
    <url>https://openqa.opensuse.org/tests/769440</url>
  </check>
  <check name="rescue_system" required="false">
    <state>success</state>
    <short_description/>
    <url>https://openqa.opensuse.org/tests/769441</url>
  </check>
  <check name="minimalx@uefi-staging" required="false">
    <state>success</state>
    <short_description/>
    <url>https://openqa.opensuse.org/tests/769442</url>
  </check>
  <check name="textmode" required="false">
    <state>success</state>
    <short_description/>
    <url>https://openqa.opensuse.org/tests/769443</url>
  </check>
  <check name="gnome" required="false">
    <state>success</state>
    <short_description/>
    <url>https://openqa.opensuse.org/tests/769444</url>
  </check>
  <check name="kde" required="false">
    <state>success</state>
    <short_description/>
    <url>https://openqa.opensuse.org/tests/769445</url>
  </check>
  <check name="cryptlvm" required="false">
    <state>success</state>
    <short_description/>
    <url>https://openqa.opensuse.org/tests/769439</url>
  </check>
</status_report>

```

That's a different id, Dude! Next to that it would be better if this was named build_id because that's what every other component names it.

PR #6636, opened 2018-12-17
Report the right status report

coolo:
:report_uuid is mandatory but wasn't used in the controller. Instead
it always returned .first - for unknown reason

Fixes https://github.com/openSUSE/open-build-service/issues/6124


coolo:
This kind of hints that the tests for status reports are a little lacking

PR #6642, opened 2018-12-17
Cache User#is_admin? & User#is_staff?

coolo:
is_admin? is called in many places, often 3 and more times
per action.

Co-authored-by: Ana María Martínez Gómez <ammartinez@suse.de>

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -385,11 +385,13 @@ class User < ApplicationRecord
   #####################
 
   def is_admin?
-    roles.where(title: 'Admin').exists?
+    return @is_admin unless @is_admin.nil?
+    @is_admin = roles.where(title: 'Admin').exists?
   end
 
   def is_staff?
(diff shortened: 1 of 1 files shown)


src/api/app/models/user.rb
```diff
@@ -385,11 +385,13 @@ def update_user_info_from_proxy_env(env)
   #####################
 
   def is_admin?
-    roles.where(title: 'Admin').exists?
+    return @is_admin unless @is_admin.nil?
+    @is_admin = roles.where(title: 'Admin').exists?
```
dmarcoux: This could also be `@is_admin ||= roles.where(title: 'Admin').exists?`.

Sorry, I didn't see this PR in time before it got merged.

Ana06: @dmarcoux no... if `@is_admin` is false, your code would look for the role again that is exactly what we want to avoid... :unamused:

PR #6646, opened 2018-12-17
Enable Bootstrap for user notifications

dmarcoux:
We forget to do this in our last deliverable


dmarcoux:
This was so much fun to fix and finding the right combination!!!!

PR #6647, opened 2018-12-17
Update styling of monitor filter

bgeuken:
* Give it more emphasive by removing the 'outline' style.
* Make it more explicit that this is an action by updating the
  button label.

screenshot from 2018-12-17 16-12-40


bgeuken:
I actually found the margin fine, once the button got more emphasize. So I didn't touch it

PR #6648, opened 2018-12-17
Migrate project index page to bootstrap

saraycp:
The content of the public projects' table is added by javascript. We keep it like
it was with small changes only: the table itself is written by the view not by JS,
we add "Search" placeholder and remove some unnecessary params.

Before:

public projects open build service 2

After:

public projects open build service

### DIFF
--- a/src/api/app/assets/javascripts/webui2/project.js
+++ b/src/api/app/assets/javascripts/webui2/project.js
@@ -1,0 +1,29 @@
+function renderProjectsTable(length) { // jshint ignore:line
+  length = length || 25;
+  var projects = mainProjects;
+  if (!$('#excludefilter').is(":checked"))
+    projects = projects.concat(exclProjects);
+  var projecturl = $("#projects-table-wrapper").data("url");
+  $("#projects-table").DataTable({
+    "data": projects,
+    "columns": [
+      {
+        "title": "Name",
+        "width": "60%",
+        "className": "text-word-break-all",
+        "render": function (obj, type, dataRow) {
+          var url = projecturl.replace(/REPLACEIT/, dataRow[0]);
+          return '<a href="' + url + '">' + dataRow[0] + '</a>';
+        }
+      },
+      {
+        "title": "Title",
+        "width": "40%",
+        "className": "text-nowrap"
+      }
+    ],
+    "pageLength": length,
+    "stateSave": true,
+    "language": { "search": '', "searchPlaceholder": "Search..." }
+  });
+}
--- a/src/api/app/views/webui2/webui/project/list.html.haml
+++ b/src/api/app/views/webui2/webui/project/list.html.haml
@@ -1,0 +1,52 @@
+:ruby
+  @pagetitle = 'Public Projects'
+.card
+  - if @important_projects.present?
+    .card-body
+      %h3 Main Projects
+      .table-responsive
+        %table.table.table-sm.table-striped.table-bordered
+          %thead
+            %tr
+              %th Name
+              %th Title
+          %tbody
+            - @important_projects.each do |project|
+              %tr
+                %td
+                  = link_to(project.first, action: :show, project: project.first)
+                %td
+                  #{project.second}
+  .card-body
+    %h3= @pagetitle
+    %ul.list-inline
+      - if @show_all
+        %li.list-inline-item.mr-3
+          = link_to(projects_path) do
+            %i.fas.fa-toggle-on.text-primary
+            Exclude #{::Configuration.unlisted_projects_filter_description}
+      - else
+        %li.list-inline-item.mr-3
+          = link_to(projects_path(show_all: true)) do
+            %i.fas.fa-toggle-off.text-primary
+            Include #{::Configuration.unlisted_projects_filter_description}
+      - unless User.current.is_nobody?
+        %li.list-inline-item
+          = link_to(new_project_path) do
+            %i.fas.fa-plus-circle.text-primary
+            Add New Project
+    - cachekey = Digest::MD5.hexdigest(@projects.join)
+    - projecturl = project_show_path('REPLACEIT')
+    #project-list
+      - if @projects.blank?
+        %p
+          %i No projects found
+      - else
+        #projects-table-wrapper{ 'data-url' => projecturl }
+          %table.responsive.table.table-sm.table-striped.table-bordered.w-100#projects-table
+          - content_for :head_javascript do
+            - cache ['list_public_arrays', cachekey] do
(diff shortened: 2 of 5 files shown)


src/api/app/assets/javascripts/webui2/project.js
```diff
@@ -0,0 +1,23 @@
+function renderProjectsTable(length) { // jshint ignore:line
+  length = (typeof length === "undefined") ? 25 : length;
```
bgeuken: Please use a [default parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) instead.

src/api/app/views/webui2/webui/project/list.html.haml
```diff
@@ -0,0 +1,51 @@
...
+            Exclude #{::Configuration.unlisted_projects_filter_description}
+        - else
+          = link_to(projects_path(show_all: true)) do
+            %i.fas.fa-toggle-off.text-primary
+            Include #{::Configuration.unlisted_projects_filter_description}
+        - unless User.current.is_nobody?
+          %span.float-right
+            = link_to(new_project_path) do
+              %i.fas.fa-plus-circle.text-primary
+              Create new project
```
dmarcoux: This should be `Add New Project` to be consistent with other Bootstrap pages (as a comparison, subprojects is using `Add New Subproject`)

src/api/app/views/webui2/webui/project/list.html.haml
```diff
@@ -0,0 +1,51 @@
...
+    - cachekey = Digest::MD5.hexdigest(@projects.join)
+    - projecturl = project_show_path('REPLACEIT')
```
dmarcoux: `REPLACEIT` is a leftover???

src/api/app/views/webui2/webui/project/list.html.haml
```diff
@@ -0,0 +1,51 @@
+:ruby
+  @pagetitle = 'Public Projects'
+.card
+  - if @important_projects.present?
+    .card-body
+      %h3 Main Projects
+      .table-responsive
+        %table.table.table-sm.table-striped.table-bordered
```
dmarcoux: I'm not sure if this table needs the search/pagination/information... I would add it. What do you think?

saraycp: It was a simple paragraph before, I thought it was better to have a list or even a table. I choosed a table because its content is quite similar to the table at the bottom. I also supposed it would only contain a few lines, so I kept it simple... but ... you make me doubt now
 :thinking:

src/api/app/assets/javascripts/webui2/project.js
saraycp: Changing the code to

```function renderProjectsTable(length = 25)```

raises this error:

```Unexpected token operator «=», expected punc «,». To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true)```

I think it is better to keep it like it is for the moment. Perhaps we will refactor this weird Javascript in the future.

**Comment by saraycp** [comment]
[Project index](https://trello.com/c/rBpV3Uf2/1000-project-index)

File: src/api/app/assets/javascripts/webui2/project.js (hunk 1 above)
Comment (coolo) [reply src/api/app/assets/javascripts/webui2/project.js]: hmm, so uglifier enforces to use old JS? Where would we set this harmony option?

File: src/api/app/views/webui2/webui/project/list.html.haml (hunk 3 above)
Comment (coolo) [reply src/api/app/views/webui2/webui/project/list.html.haml]: This is a pattern of mine to generate links in javascript by using a template generated from rails.

**Comment by coolo** [comment]
Within the production data seems to be a home project (excluding homes looks fine, you get a 2 columned table) that breaks the layout. That problem is also visible on https://build.opensuse.org/project?show_all=true, so just FYI.

screencapture-e120-suse-de-3000-project-2018-12-18-07_16_26

File: src/api/app/assets/javascripts/webui2/project.js (hunk 1 above)
Comment (Ana06) [reply src/api/app/assets/javascripts/webui2/project.js]: @coolo:

```
config.assets.js_compressor = Uglifier.new(harmony: true)
```

This typeof is really ugly... I would use `||` similarly as in ruby:

`length = length || 25`

File: src/api/app/views/webui2/webui/project/list.html.haml [hunk 5]
```diff
@@ -0,0 +1,52 @@
...
+          %thead
+            %tr
+              %th Name
+              %th Title
+          %tbody
+            - @important_projects.each do |project|
+              %tr
+                %td
+                  = link_to(project[0], action: :show, project: project[0])
```
Ana06: [nitpick] project.first

src/api/app/views/webui2/webui/project/list.html.haml
```diff
@@ -0,0 +1,52 @@
...
+                %td
+                  #{project[1]}
```
Ana06: [nitpick] project.second

src/api/app/views/webui2/webui/project/list.html.haml
bgeuken: We probably could use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) instead, but they are also not supported by IE. This sucks

src/api/app/assets/javascripts/webui2/project.js
bgeuken: > hmm, so uglifier enforces to use old JS? Where would we set this harmony option?

The problem is that Internet Explorer is not supporting ES6 as it seems. Unless we use a transpiler we have to stick with the old syntax.

saraycp:
> Within the production data seems to be a home project (excluding homes looks fine, you get a 2 columned table) that breaks the layout. That problem is also visible on https://build.opensuse.org/project?show_all=true, so just FYI.

@coolo, it is fixed now. When we have long names and title it breaks the line:

project_index_long_names

@dmarcoux please read my comments and let me know if you agree.
@Ana06 please have another look

hennevogel:
> The content of the public projects' table is added by javascript. We keep it like it was

Because?

coolo:
> > The content of the public projects' table is added by javascript. We keep it like it was
> 
> Because?

Because server side pagination is out of scope for the bootstrap migration :)

hennevogel:
> Because server side pagination is out of scope for the bootstrap migration :)

That is not true, if you can avoid migrating this 'pattern' to the new UI. The first "rule" we have for the migration scope is: Identify bad code patterns in the OLD UI and refactor them

PR #6649, opened 2018-12-17
Fix disappearing tabs on page zoom

### LINKED ISSUE #6597: [Bootstrap] Decrease in useability of the current Beta web UI
# Issue Description
I have switched to the BETA of OBS web interface and most things I don't care about, but I see clear and unequivocal worsening of the useability in replacing "tabs" in the top of each section on package pages and overview of requests pages.

In the previous version I could see immediately the state of my requests queues, now I have to use mouse to get the data.

In the previous version I could immediately go to all requests of each package, now I have to search for the tiny button in the corner of the page.

Expected Result
===============

Swift overview of the situation

Observed Result
===============

I have to search for tiny arrow with a mouse to get same information, and even to click on it.

How to Reproduce
===============

1. Go to https://build.opensuse.org/project/show/home:mcepl or https://build.opensuse.org/project/show/home:mcepl

2. See error

dmarcoux:
On Firefox, there is a 1 pixel difference on some page zoom levels (130% and up) between an element's offset height/width and scroll height/width. This causes the tabs to shrink when they should not.

Fixes #6597


src/api/app/assets/javascripts/webui2/tabs.js
```diff
@@ -27,7 +27,7 @@ function refreshTabs(tabList) {
 
 $.fn.hasOverflow = function() {
   var element = $(this)[0];
-  return (element.offsetHeight < element.scrollHeight) || (element.offsetWidth < element.scrollWidth);
+  return (element.offsetHeight < element.scrollHeight - 1) || (element.offsetWidth < element.scrollWidth - 1);
```
coolo: you heard of <= ? :)

dmarcoux: Obviously never, right? :man_facepalming:

dmarcoux: It's not the same though.

Because with `<=`, it will keep being true when the scroll and offset are the same, so this will shrink all the tabs even though it should not.

coolo: OK, let's agree on: this needs a little comment :)

PR #6651, opened 2018-12-17
Remove Refresh button from requests page in bootstrap

saraycp:
When we set some filters in requests' page and then click on Refresh,
all the filters get reset (back to default values). We also realized
that the Refresh functionality is not really needed. Therefore, we
decide to remove the button.

requests for pro1 open build service


saraycp:
[Remove Refresh button on requests page](https://trello.com/c/tbzueXJr/1001-remove-refresh-button-on-requests-page)

PR #6653, opened 2018-12-18
Refactor flag manipulation

coolo:
Several improvements:
 - Handle all flag manipulations through ajax, not only those within the matrix 
   (the current behaviour is *very* confusing)
 - Don't repeat the popover code all the time
 - Have one action for flag manipulation instead of 3

Avoiding the popover code increases the performance on large projects' repositories page:

master:
Completed 200 OK in 9097ms

refactor:
Completed 200 OK in 5678ms


coolo:
This includes #6626 to avoid merge conflicts, but it's independent from it.

src/api/app/assets/javascripts/webui2/repositories.js
```diff
@@ -1,22 +1,85 @@
 function setSpinnersForFlags() { // jshint ignore:line
-  $(document).on('click','.popover_flag_action', function() {
-    var flag = $(this).data('flag-id');
-    var icon = $('div[id="' + flag + '"] a');
-    icon.addClass('d-none');
-    icon.next().removeClass('d-none');
-    $('.popover').popover('hide');
-  });
+    $(document).on('click', '.popover_flag_action', function() {
+        var flag = $(this).data('flag-id');
```
coolo: this indenting was done by atom's beautifier. Does obs have a indentation policy on javascript?

src/api/app/controllers/webui/repositories_controller.rb
```diff
@@ -241,12 +244,42 @@ def remove_flag
...
+    flag_type = params[:flag]
+    if params[:command] == 'remove'
+      @main_object.flags.of_type(flag_type).where(repo: params[:repository], architecture: architecture).delete_all
+    elsif params[:command] == 'set-disable'
+      @main_object.flags.create(flag: flag_type, repo: params[:repository], architecture: architecture, status: 'disable')
+    elsif params[:command] == 'set-enable'
+      @main_object.flags.create(flag: flag_type, repo: params[:repository], architecture: architecture, status: 'enable')
+    end
+    # FIXME: This should happen in Flag or even better in Project
+    @main_object.store
```
coolo: This store is rather expensive for larger projects, but fixing the performance of _meta rendering benefits us in more places

src/api/app/views/webui2/shared/_repositories_flag_table_column.html.haml
```diff
@@ -1,19 +1,21 @@
...
   title = flag.status.capitalize
-  title += is_flag_set_by_user ? ' (set by user)' : ' (calculated)'
-  if user_can_set_flags
-    content = render(partial: 'shared/flag_popover',
-                     locals: { flag: flag, flags: flags, repository: repository, architecture: architecture,
-                               project: project, package: package })
-  end
   icon_class = flag.status == 'disable' ? 'fa-ban text-danger' : 'fa-check text-success'
-  icon_class += ' ml-0_6' if is_flag_set_by_user
+  data = { status: flag.status, flag: flag.flag, repository: repository,
```
coolo: next thing to do: this is really a helper function now - most is ruby (that's the reason I left the helper module empty and didn't remove it)

src/api/app/controllers/webui/repositories_controller.rb
```diff
@@ -241,12 +242,27 @@ def remove_flag
...
-        switch_to_webui2
         render 'change_flag'
       end
     end
   end
 
+  # POST flag/manipulate/:project(/:package)
+  # TODO when removing bento, remove the extra manipulate from the route, for
+  # now we need to avoid the clash with create_flag
+  def manipulate_flag
```
Ana06: `change_flag`? :thinking:

src/api/config/routes.rb
```diff
@@ -188,10 +188,12 @@ def self.public_or_about_path?(request)
       post 'project/create_image_repository' => :create_image_repository
 
       # Flags
+      post 'flag/manipulate/:project(/:package)' => :manipulate_flag, constraints: cons, as: 'manipulate_repository_flag'
+      # TODO bento_only
       put 'flag/:project(/:package)' => :toggle_flag, constraints: cons, as: 'toggle_repository_flag'
       post 'flag/:project(/:package)' => :create_flag, constraints: cons, as: 'create_repository_flag'
       delete 'flag/:project(/:package)/:flag' => :remove_flag, constraints: cons, as: 'remove_repository_flag'
-      post 'flag/:project(/:package)' => :manipulate_flag, constraints: cons, as: 'manipulate_repository_flag'
+      # end TODO
```
Ana06: I am not sure if this is clear, I would add the routes names to the first comment instead :thinking:

src/api/app/assets/javascripts/webui2/repositories.js
Ana06: I think it is 2 spaces :thinking:

src/api/app/controllers/webui/repositories_controller.rb
```diff
@@ -253,16 +253,32 @@ def remove_flag
...
+    flag_type = params[:flag]
+    if params[:command] == 'remove'
+      @main_object.flags.of_type(flag_type).where(repo: params[:repository], architecture: architecture).delete_all
+    elsif params[:command] == 'set-disable'
+      @main_object.flags.create(flag: flag_type, repo: params[:repository], architecture: architecture, status: 'disable')
+    elsif params[:command] == 'set-enable'
+      @main_object.flags.create(flag: flag_type, repo: params[:repository], architecture: architecture, status: 'enable')
     end
+    # FIXME: This should happen in Flag or even better in Project
+    #@main_object.store
```
Ana06: This is removed in the next commit... Commit that can be squashed?

src/api/app/controllers/webui/repositories_controller.rb
coolo: the problem in both cases is: how to diff between :project/:package and manipulate/:project

src/api/config/routes.rb
coolo: I just repeat the todo

src/api/app/assets/javascripts/webui2/repositories.js
```diff
@@ -12,11 +12,76 @@ function setRepoCheckbox() { // jshint ignore:line
...
       $form = $('#' + id + '_create');
     } else {
       $form = $('#' + id + '_destroy');
     }
     $form.submit();
   });
 }
+
+function takeOverData(target, source, field) {
+  target.data(field, source.data(field));
```
bgeuken: Hm, the function call is almost longer than the code it contains :thinking:

coolo: true - it evolved a little - and then I didn't bother to fold it back in.

coolo: But it avoids duplicating the field name.

src/api/app/assets/javascripts/webui2/repositories.js
```diff
@@ -12,11 +12,76 @@ function setRepoCheckbox() { // jshint ignore:line
...
+    option2 = '.flag-set-enable';
+  }
+
+  var clone = $('<div/>');
+  takeOverData(clone, $(this), 'repository');
+  takeOverData(clone, $(this), 'architecture');
+  takeOverData(clone, $(this), 'flag');
+
+  clone.append($(option1).html());
+  clone.append('<div class=".pt-2"/>');
```
bgeuken: That's already a class:wink:

src/api/app/controllers/webui/repositories_controller.rb
```diff
@@ -241,14 +244,47 @@ def remove_flag
...
 
+  def follow_change_flag_command(flag_type)
+    architecture = Architecture.from_cache!(params[:architecture]) unless params[:architecture].nil?
+
+    if params[:command] == 'remove'
+      @main_object.flags.of_type(flag_type).where(repo: params[:repository], architecture: architecture).delete_all
+    elsif params[:command] == 'set-disable'
+      @main_object.flags.create(flag: flag_type, repo: params[:repository], architecture: architecture, status: 'disable')
+    elsif params[:command] == 'set-enable'
+      @main_object.flags.create(flag: flag_type, repo: params[:repository], architecture: architecture, status: 'enable')
```
bgeuken: Or merge the two conditions into one. They are almost the same:wink:

coolo: I did, not sure you'll like my solution

coolo:
Together with #6682 this brings an improvement on repositories/openSUSE:Tools from 7s to 2s (remember, we've had >50 a week ago :)

Problem is that it still takes 10s before the page is finished as a lot of javascript is going on :(

src/api/app/controllers/webui/repositories_controller.rb
bgeuken: Fine with me. Note that in Ruby ^ and $ are marking start and end of line. \A and \z are for beginning and end of a string.
Not that it would matter here.

coolo: good to know

src/api/spec/support/shared_examples/features/boostrap_flag_tables.rb
```diff
@@ -65,20 +75,23 @@ def css_locator_for(repository, architecture)
     query_attributes[:flag] = flag_type
 
     disable_flag_field_for(repository: 'All', architecture: 'All')
-    expect(project.flags.reload.where(query_attributes.merge(status: :disable))).to exist
+    expect(project.flags.reload.where(query_attributes).pluck(:status)).to contain_exactly('disable')
```
Ana06: why is reload needed? b:confused:

coolo: because the project.flags are manipluated by selenium - and the rspec object doesn't know

coolo: I can try to leave it out

coolo:
rebased - let's see what broke :)

coolo:
that was unexpected -> nothing broke :)

src/api/spec/support/shared_examples/features/boostrap_flag_tables.rb
```diff
@@ -45,40 +55,43 @@ def css_locator_for(repository, architecture)
     query_attributes[:repo] = repository.name
 
     disable_flag_field_for(repository: repository.name, architecture: 'All')
-    expect(project.flags.reload.where(query_attributes.merge(status: :disable))).to exist
+    expect(project.flags.where(query_attributes.merge(status: :disable))).to exist
```
Ana06: I looks like I was right :nerd_face:

coolo: if it helps you :)

PR #6656, opened 2018-12-18
Update autocompletion queries and drop unused scopes

bgeuken:
* Drop unused project scopes
* Make the autocompletion queries more flexible.
  We used to only search for matches that would start with the search
  string. Since we now limit the search results, we be more flexible
  and include results that match with the search string at all.

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -85,8 +85,6 @@ class Project < ApplicationRecord
   scope :not_maintenance_incident, -> { where("kind <> 'maintenance_incident'") }
   scope :maintenance_incident, -> { where("kind = 'maintenance_incident'") }
   scope :maintenance_release, -> { where("kind = 'maintenance_release'") }
-  scope :home, -> { where("name like 'home:%'") }
-  scope :not_home, -> { where.not("name like 'home:%'") }
   scope :filtered_for_list, lambda {
     where.not('name rlike ?', ::Configuration.unlisted_projects_filter) if ::Configuration.unlisted_projects_filter.present?
   }
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -31,7 +31,7 @@ class Project < ApplicationRecord
   has_many :relationships, dependent: :destroy, inverse_of: :project
   has_many :packages, inverse_of: :project do
     def autocomplete(search)
-      where(['lower(packages.name) like lower(?)', "#{search}%"]).order(:name).limit(50)
+      where(['lower(packages.name) like lower(?)', "%#{search}%"]).order(:name).limit(50)
     end
   end
 
@@ -89,7 +89,7 @@ class Project < ApplicationRecord
     where.not('name rlike ?', ::Configuration.unlisted_projects_filter) if ::Configuration.unlisted_projects_filter.present?
   }
   scope :remote, -> { where('NOT ISNULL(projects.remoteurl)') }
-  scope :autocomplete, ->(search) { where('lower(name) like lower(?)', "#{search}%").order(:name).limit(50) }
+  scope :autocomplete, ->(search) { where('lower(name) like lower(?)', "%#{search}%").order(:name).limit(50) }
 
   # will return all projects with attribute 'OBS:ImageTemplates'
   scope :local_image_templates, lambda {


coolo:
I kind of like it:

bgeuken:
> Before you were showing shorter results (most similar to the search) first. Now you are not. This is potentially return to many things making the autocomplete unusable.

This was and still is returning results that match a given search string. The only difference now is that when searching for 'Unstable' I already get our 'OBS:Server:Unstable' project, and it's sub-projects, as search results.
Before it would only have matched with projects that start with 'Unstable'. Now let's assume there are a lot of them. Then our 'OBS:Server:Unstable' project wouldn't be part of that list.
But that's okay since this is an autocomplete. A user would just have to further specify the search string. And IMO this is an edge case as well.


Regarding the length of the result. That's just a side-effect of ordering it alphabetically. There is nothing that would guarantee that there aren't many long results that start with an 'a' and overshadow shorter results starting with a 'z'.

coolo:
Ana, it's 2:1 - your argument is valid, but still rather theoretic. While @bgeuken and me prefer the benefits this brings in *most* cases.

src/api/app/models/project.rb
```diff
@@ -31,7 +31,7 @@ class Project < ApplicationRecord
   has_many :relationships, dependent: :destroy, inverse_of: :project
   has_many :packages, inverse_of: :project do
     def autocomplete(search)
-      where(['lower(packages.name) like lower(?)', "#{search}%"]).order(:name).limit(50)
+      where(['lower(packages.name) like lower(?)', "%#{search}%"]).order(:name).limit(50)
```
ChrisBr: @bgeuken JFYI: There is a reason for only using the wildcard as postfix. If you ``%`` at the beginning of the query MySQL is not able to use an index anymore which will make the queries a lot slower. Unless there is really a requirement for this feature I would prefer to revert.

coolo: I can't find any proof that this matters. With or without prefix % it's using where on the packages_all_index index.

```
irb(main):006:0> search='coolo'; Project.find_by_name('openSUSE:Factory').packages.where(['lower(packages.name) like lower(?)', "%#{search}%"]).order(:name).limit(50).explain
=> EXPLAIN for: SELECT  `packages`.* FROM `packages` WHERE `packages`.`id` NOT IN (SELECT `packages`.`id` FROM `packages` WHERE `packages`.`project_id` IN (50315, 60546, 69200, 122039, 122041, 122042, 122043, 189310, 248287, 254068, 97989, 277778)) AND `packages`.`project_id` = 110 AND (lower(packages.name) like lower('%coolo%')) ORDER BY `packages`.`name` ASC LIMIT 50
+----+--------------+----------+-------+----------------------------+--------------------+---------+-------+-------+-----------------------------+
| id | select_type  | table    | type  | possible_keys              | key                | key_len | ref   | rows  | Extra                       |
+----+--------------+----------+-------+----------------------------+--------------------+---------+-------+-------+-----------------------------+
|  1 | PRIMARY      | packages | ref   | packages_all_index         | packages_all_index | 4       | const | 25140 | Using where; Using filesort |
|  2 | MATERIALIZED | packages | range | PRIMARY,packages_all_index | packages_all_index | 4       | NULL  |    25 | Using where; Using index    |
+----+--------------+----------+-------+----------------------------+--------------------+---------+-------+-------+-----------------------------+
2 rows in set (0.00 sec)

irb(main):007:0> search='coolo'; Project.find_by_name('openSUSE:Factory').packages.where(['lower(packages.name) like lower(?)', "#{search}%"]).order(:name).limit(50).explain
=> EXPLAIN for: SELECT  `packages`.* FROM `packages` WHERE `packages`.`id` NOT IN (SELECT `packages`.`id` FROM `packages` WHERE `packages`.`project_id` IN (50315, 60546, 69200, 122039, 122041, 122042, 122043, 189310, 248287, 254068, 97989, 277778)) AND `packages`.`project_id` = 110 AND (lower(packages.name) like lower('coolo%')) ORDER BY `packages`.`name` ASC LIMIT 50
+----+--------------+----------+-------+----------------------------+--------------------+---------+-------+-------+-----------------------------+
| id | select_type  | table    | type  | possible_keys              | key                | key_len | ref   | rows  | Extra                       |
+----+--------------+----------+-------+----------------------------+--------------------+---------+-------+-------+-----------------------------+
|  1 | PRIMARY      | packages | ref   | packages_all_index         | packages_all_index | 4       | const | 25140 | Using where; Using filesort |
|  2 | MATERIALIZED | packages | range | PRIMARY,packages_all_index | packages_all_index | 4       | NULL  |    25 | Using where; Using index    |
+----+--------------+----------+-------+----------------------------+--------------------+---------+-------+-------+-----------------------------+
2 rows in set (0.00 sec)

Comment (coolo) [reply src/api/app/models/project.rb]: (and there is nothing to revert - it's not merged yet :)

Comment (ChrisBr) [reply src/api/app/models/project.rb]: @coolo you're always online aren't you? :smile:  Hm ok, the lower seems to be the problem which anyway creates a temporary table :cry: 

```
irb(main):014:0> Project.where("name like lower('Main%')").explain
  Project Load (0.6ms)  SELECT `projects`.* FROM `projects` WHERE (projects.id not in (0)) AND (name like lower('Main%'))
=> EXPLAIN for: SELECT `projects`.* FROM `projects` WHERE (projects.id not in (0)) AND (name like lower('Main%'))
+----+-------------+----------+-------+-----------------------------+---------------------+---------+------+------+-----------------------+
| id | select_type | table    | type  | possible_keys               | key                 | key_len | ref  | rows | Extra                 |
+----+-------------+----------+-------+-----------------------------+---------------------+---------+------+------+-----------------------+
|  1 | SIMPLE      | projects | range | PRIMARY,projects_name_index | projects_name_index | 602     | NULL |    1 | Using index condition |
+----+-------------+----------+-------+-----------------------------+---------------------+---------+------+------+-----------------------+
1 row in set (0.00 sec)

irb(main):015:0> Project.where("lower(name) like lower('Main%')").explain
  Project Load (0.5ms)  SELECT `projects`.* FROM `projects` WHERE (projects.id not in (0)) AND (lower(name) like lower('Main%'))
=> EXPLAIN for: SELECT `projects`.* FROM `projects` WHERE (projects.id not in (0)) AND (lower(name) like lower('Main%'))
+----+-------------+----------+-------+---------------+---------+---------+------+------+-------------+
| id | select_type | table    | type  | possible_keys | key     | key_len | ref  | rows | Extra       |
+----+-------------+----------+-------+---------------+---------+---------+------+------+-------------+
|  1 | SIMPLE      | projects | range | PRIMARY       | PRIMARY | 4       | NULL |   18 | Using where |
+----+-------------+----------+-------+---------------+---------+---------+------+------+-------------+
1 row in set (0.00 sec)

```

coolo: I'm not always online - just if the Switch is taken by another family member :)

bgeuken: > @bgeuken JFYI: There is a reason for only using the wildcard as postfix. If you % at the beginning of the query MySQL is not able to use an index anymore which will make the queries a lot slower.

I doubt this is relevant (= noticeable by users) since we now limit the results to 50. Same goes for the more "loose" regex.
And the better usability should outweight this anyway.

PR #6657, opened 2018-12-18
Improve efficience of PackageController#buildresults

Ana06:
Pre-calculate in the controller if a build result is outdated (not in database but still temporarily in backend) instead of in the view, generating and SQL query per result. :nerd_face: 

### MASTER
real=47.93

### NEW CODE
real=21.93

--------------------------------------------------------------------


coolo:
efficience is a french word - in german it's Kostenwirksamkeit ;)

PR #6658, opened 2018-12-18
Migrate project creation page to bootstrap

saraycp:
The project creation form is re-designed using bootstrap and it is shown on a modal instead of a page.

The code for creation modal is unified for both projects and subprojects, they are very similar.

Before:

create new project open build service

After:

public projects open build service 5

### DIFF
--- a/src/api/app/views/webui2/webui/project/subprojects.html.haml
+++ b/src/api/app/views/webui2/webui/project/subprojects.html.haml
@@ -16,7 +16,7 @@
           %p
             %i This project has no subprojects
         - if User.current.can_modify?(@project)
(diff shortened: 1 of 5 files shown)


coolo:
please rebase

src/api/app/views/webui2/webui/project/subprojects.html.haml
```diff
@@ -16,15 +16,15 @@
           %p
             %i This project has no subprojects
         - if User.current.can_modify?(@project)
-          = link_to('#', data: { toggle: 'modal', target: '#add-new-subproject-modal' }, title: 'Add New Subproject', class: 'nav-link') do
+          = link_to('#', data: { toggle: 'modal', target: '#new-project-modal' }, title: 'Add New Subproject', class: 'nav-link') do
```
dmarcoux: The `title` is not needed here as we already have the text `Add New Subproject` in the link

PR #6659, opened 2018-12-18
Refactor removing excluded request

eduardoj:
Instead of generating html code for each modal that asks for removing a excluded reques,t only one modal is created. Arguments for the modal are passed with "data-*" attributes.

Following work of #6608.


coolo:
coveralls hangs again :(

PR #6660, opened 2018-12-18
Enable consider_all_requests_local in test suite

bgeuken:
This allows us to see the backtrace in CircleCI when a feature test fails, which makes debugging   feature tests much easier. :bowtie: 

Before (CircleCI artifact):

screenshot_2018-12-18 packages_triggering_package_rebuild_via_binaries_view png png image 1280 x 1024 pixels - scaled 7

After:

screenshot_2018-12-18 packages_triggering_package_rebuild_via_binaries_view png png image 1280 x 1024 pixels - scaled 9

### DIFF
--- a/src/api/config/environments/test.rb
+++ b/src/api/config/environments/test.rb
@@ -19,8 +19,7 @@ OBSApi::Application.configure do
   config.eager_load = ENV.fetch('EAGER_LOAD', '0') == '1'
 
   # Show full error reports and disable caching
-  # local requests don't trigger the global exception handler -> set to false
-  config.consider_all_requests_local = false
+  config.consider_all_requests_local = true
   config.action_controller.perform_caching = false
 
   # Configure public file server for tests with Cache-Control for performance.


src/api/config/environments/test.rb
```diff
@@ -20,7 +20,7 @@
 
   # Show full error reports and disable caching
   # local requests don't trigger the global exception handler -> set to false
-  config.consider_all_requests_local = false
+  config.consider_all_requests_local = true
```
coolo: the comment above that line looks a little misplaced now :)

The rails template just is this:

```
  # Show full error reports and disable caching.
  config.consider_all_requests_local       = true
  config.action_controller.perform_caching = false

  # Raise exceptions instead of rendering exception templates.
  config.action_dispatch.show_exceptions = false

```

And I bet it always has been - but this false has always been in here and noone questioned it so far. There are more puzzling diffs to rails default houghs.

   # Disable request forgery protection in test environment.
-  config.action_controller.allow_forgery_protection = false
+ config.action_controller.allow_forgery_protection = true

(it's interesting that the comment is the same :)

coolo:
What I wonder: will we still see the exception in circle's plain results?

coolo:
I guess it would help with

```
[ 2068s] Failures:
[ 2068s] 
[ 2068s]   1) Projects branching a package and select current revision
[ 2068s]      Failure/Error: expect(page).to have_text('Successfully branched package')
[ 2068s]        expected to find text "Successfully branched package" in "500: Internal Server Error\nOooops, something went terribly wrong, our application crashed :-( We think it's best if you search if this issue is already reported on GitHub.\nSearch\nIn case of emergency...\nLet's be honest, we're both not perfect (yet!). But don't despair, as free software project we're commited to work together and collaborate. Here is how we do that.\n...we help each other out\nOn IRC and on the mailing list there is a vibrant community of developers, admins and users that support each other.\n...we tell us about problems\nIn the unfortunate event you've hit a bug in the OBS we depend on you to report them on GitHub.\n...we better ourselves\nYou have an idea how to make OBS better, faster, stronger? Those ideas we call feature and we track them in openFATE."
[ 2068s]      # ./spec/features/webui/projects_spec.rb:413:in `block (3 levels) in <top (required)>'
[ 2068s] 
``` 

as well

bgeuken:
> @coolo
> What I wonder: will we still see the exception in circle's plain results?

It does, see: https://circleci.com/gh/bgeuken/open-build-service/1314#tests/containers/3

And I wouldn't know why not

src/api/config/environments/test.rb
bgeuken: Hard to say and probably something worth looking at. But I don't see why this would block the PR.

Regarding the deviation from the default Rails config... It might be similar to [public_file_server.enabled](https://github.com/openSUSE/open-build-service/pull/6616/files). In the Rails  4 to Rails 5 migration guide it was advertised, but the edge Rails guide says something different and is a bit more verbose.

As said, we could review them.

coolo: blocking the PR is only the misplaced comment saying the opposite to what you do

bgeuken: ahh :bulb:

coolo:
> And I wouldn't know why not

Because the exceptions are no longer thrown into rspec. The example you gave is not a code exception, but a rspec error.

PR #6664, opened 2018-12-19
Cache user_can_modify_project in repository page

coolo:
This function was called *twice* per repository and is not
exactly cheap:

```
  CACHE Flag Exists (0.0ms)  SELECT  1 AS one FROM `flags` WHERE `flags`.`project_id` = 2 AND `flags`.`flag` = 'lock' AND `flags`.`status` = 'enable' LIMIT 1
has_global_permission? change_project
  CACHE  (0.0ms)  SELECT `roles_static_permissions`.`role_id` FROM `roles_static_permissions` INNER JOIN `static_permissions` ON `static_permissions`.`id` = `roles_static_permissions`.`s tatic_permission_id` WHERE `static_permissions`.`title` = 'change_project'
running local permission check: user coolo, project home:Admin, permission 'change_project'
  CACHE Project Load (0.0ms)  SELECT  `projects`.* FROM `projects` WHERE (projects.id not in (0)) AND `projects`.`name` = 'home' ORDER BY `projects`.`name` DESC LIMIT 1
  CACHE Relationship Exists (0.0ms)  SELECT  1 AS one FROM `relationships` WHERE `relationships`.`project_id` = 2 AND `relationships`.`user_id` = 3 AND (role_id in (1,2)) LIMIT 1
  CACHE Relationship Exists (0.0ms)  SELECT  1 AS one FROM `relationships` INNER JOIN `groups` ON `groups`.`id` = `relationships`.`group_id` INNER JOIN `groups_users` ON `groups_users`.` group_id` = `groups`.`id` WHERE `relationships`.`project_id` = 2 AND `groups_users`.`user_id` = 3 AND (role_id in (1,2)) LIMIT 1

### DIFF
--- a/src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
+++ b/src/api/app/views/webui2/webui/repositories/_repository_card_content.html.haml
@@ -18,7 +18,7 @@
 
 .card-footer.text-center
   .row
-    - if User.current.can_modify?(project)
+    - if user_can_modify_project
       .col
         = link_to('#', title: 'Edit Repository',
                        data: { toggle: 'modal', target: "#edit-repository-#{repository.id}" }) do
@@ -32,7 +32,7 @@
       = link_to(url, title: 'Download Repository') do
         %i.fas.fa-download.text-secondary
 
-    - if User.current.can_modify?(project)
+    - if user_can_modify_project
       .col
         = link_to('#', title: 'Delete Repository',
                        data: { toggle: 'modal', target: "#delete-repository-#{repository.id}" }) do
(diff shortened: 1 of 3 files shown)


bgeuken:
> It looks cheap there (0.0) everything is cached stuck_out_tongue_closed_eyes but I guess it doesn't hurt to catch it before wink

I don't get this one. Why would you want to cache something twice?

coolo:
Oh man! There is a *huge* difference between caching a boolean result and caching SQL query results. Just because it doesn't hit the database, doesn't mean it's cheap. Want a benchmark again?

User.current=User.find_by_login('coolo')
project=Project.find_by_name('home:bgeuken')

Benchmark.measure { 100.times { User.current.can_modify?(project) } }
@real=1.4927424201741815

can_he=User.current.can_modify?(project); Benchmark.measure { 100.times { can_he } }
@real=1.2598000466823578e-05

That's 1.5 seconds for 100 calls - even though everything the database has to do is cached.

coolo:
circleci, where art thou?

coolo:
> I don't get this one. Why would you want to cache something twice?

It even has a name in rails: Russian doll caching

coolo:
> Well, russian doll caching is about having different layers of caches, for different layers of code / your app. Not for caching the same thing twice

caching different layers is exactly what this is about

PR #6675, opened 2018-12-19
Remove architecture and repository relationship from project and package

coolo:
The were only supporting the rather crude get_flags, which is only
used in one controller action, which queries architecture and
repositories anyway - so wasteful

### DIFF
--- a/src/api/.rubocop_todo.yml
+++ b/src/api/.rubocop_todo.yml
@@ -217,7 +217,7 @@ Metrics/BlockNesting:
 # Offense count: 87
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 1103
+  Max: 1105
 
 # Offense count: 239
 Metrics/CyclomaticComplexity:
--- a/src/api/app/controllers/webui/repositories_controller.rb
+++ b/src/api/app/controllers/webui/repositories_controller.rb
@@ -12,15 +12,15 @@ class Webui::RepositoriesController < Webui::WebuiController
   # GET project/repositories/:project
   def index
     return if switch_to_webui2
-    @build = @main_object.get_flags('build')
-    @debuginfo = @main_object.get_flags('debuginfo')
-    @publish = @main_object.get_flags('publish')
-    @useforbuild = @main_object.get_flags('useforbuild')
-    @architectures = @main_object.architectures.reorder('name').distinct
+    @architectures = Architecture.where(id: @project.repository_architectures.select(:architecture_id)).order(:name)
+    @repositories = @project.repositories.includes(:path_elements, :download_repositories)
+    repository_names = @repositories.pluck(:name)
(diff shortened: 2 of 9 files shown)


src/api/.rubocop_todo.yml
```diff
@@ -217,7 +217,7 @@ Metrics/BlockNesting:
 # Offense count: 87
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 1103
+  Max: 1105
```
Ana06: we have too many long classes! :see_no_evil:

src/api/app/controllers/webui/repositories_controller.rb
```diff
@@ -12,15 +12,15 @@ class Webui::RepositoriesController < Webui::WebuiController
   # GET project/repositories/:project
   def index
     return if switch_to_webui2
-    @build = @main_object.get_flags('build')
-    @debuginfo = @main_object.get_flags('debuginfo')
-    @publish = @main_object.get_flags('publish')
-    @useforbuild = @main_object.get_flags('useforbuild')
-    @architectures = @main_object.architectures.reorder('name').distinct
+    @architectures = Architecture.where(id: @project.repository_architectures.select(:architecture_id)).reorder('name').distinct
```
Ana06: why reorder? just order!

src/api/app/controllers/webui2/repositories_controller.rb
```diff
@@ -7,7 +7,7 @@ def webui2_index
 
     @user_can_set_flags = policy(@project).update?
 
-    @architectures = @project.architectures.reorder('name')
+    @architectures = Architecture.where(id: @project.repository_architectures.select(:architecture_id)).reorder('name').distinct
```
Ana06: why is the distinct needed? :confused:

src/api/test/unit/project_test.rb
```diff
@@ -18,6 +18,20 @@ def test_maintained_project_names
     assert_equal ['A', 'B', 'C'], project.maintained_project_names
   end
 
+  # TODO: bento_only
```
Ana06: where are you skipping the new UI? :confused:

src/api/app/controllers/webui/repositories_controller.rb
coolo: indeed

src/api/app/controllers/webui2/repositories_controller.rb
coolo: it's most likely not, I just wasn't sure

src/api/test/unit/project_test.rb
coolo: the new UI isn't using get_flags. The whole test (and the function tested) can be removed with bento - or as I wrote in the commit, we switch the old UI to the new methods as well.

coolo:
fixed the query

PR #6683, opened 2018-12-20
Avoid some more N+1 queries

coolo:


src/api/app/views/models/_project.xml.builder
```diff
@@ -19,7 +19,7 @@ xml.project(project_attributes) do
 
   my_model.render_relationships(xml)
 
-  repos = my_model.repositories.not_remote.sort { |a, b| b.name <=> a.name }
+  repos = my_model.repositories.preload(:download_repositories, :release_targets, :hostsystem, path_elements: :link).not_remote.sort { |a, b| b.name <=> a.name }
```
bgeuken: If you  find some time please also change the sort to [order](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-order).

coolo: good idea: 8cedff5

PR #6686, opened 2018-12-21
[WIP] Replace client-side by server-side pagination for project index

saraycp:
Playing around with [effective_datatables](https://github.com/code-and-effect/effective_datatables) gem to use server-side pagination instead of customized client-side pagination by javascript.

**Pagination:**

effective_datatables_pagination

**Search:**

effective_datatables_search

**Other actions:**

effective_datatables_actions

This is just WIP and not sure we are going to finally use this gem. If so, we should:
- change all the datatables in our code by _effective_datatables_
- improve design
- uninstall _jquery-datatables_ gem
- remove all the existing javascript and css files related to old datatables because requiring the gem files makes the old views mix both css, like in the following image:
show test_project open build service

    
**NOTE 1:** It goes through application controller (API), so we have to skip `validate_params` _before_action._ Otherwise, «effective_datatables» breaks. We should find a better solution.

**NOTE 2:** Please have a look to the `FIXME:` in the code. Any advice would help.


src/api/app/controllers/application_controller.rb
```diff
@@ -24,7 +24,8 @@ class ApplicationController < ActionController::Base
   # skip the filter for the user stuff
   before_action :extract_user
   before_action :shutup_rails
-  before_action :validate_params
+  # FIXME: efective_databases functionality breaks when this controller tries to validate params. There is a better way to do it?
```
coolo: monkey patch DatatablesController to inject a skip_before_action?

coolo:
this looks too good to be true :)

saraycp:
[Bootstrap Migration](https://trello.com/c/HhbQzp60/1399-bootstrap-migration)

saraycp:
[Use server side pagination for the Project#Index DataTable](https://trello.com/c/q7NL91TC/1616-use-server-side-pagination-for-the-projectindex-datatable)

saraycp:
I close this PR because another gem,  [ajax-datatables-rails](https://github.com/jbox-web/ajax-datatables-rails) seems to be better for us. Its default behavior  is closer to our needs,  whereas [effective_datatables](https://github.com/code-and-effect/effective_datatables) requires some patches to get what we need.

Some points against  `effective_datatables`:

- It includes a search box per column and we prefer one for everything.
- Search is case sensitive  and we need it to be insensitive.
- We don't need all the actions provided and it 's not easy to remove only some of them.

`ajax-datatables-rails`  tested in PR #6864.

coolo:
Told you - too good to be true :)

saraycp:
@coolo, you were right! :rofl:

PR #6700, opened 2019-01-02
Require a type in request action

coolo:
The baseclass is not a valid action. Fortunately it only happens in test cases

### DIFF
--- a/src/api/spec/controllers/webui/request_controller_spec.rb
+++ b/src/api/spec/controllers/webui/request_controller_spec.rb
@@ -13,20 +13,20 @@ RSpec.describe Webui::RequestController, vcr: true do
   let(:target_package) { create(:package_with_file, name: 'goal', project_id: target_project.id) }
   let(:source_project) { submitter.home_project }
   let(:source_project_fluffy) { submitter_with_group.home_project }
-  let(:source_package) { create(:package, name: 'ball', project_id: source_project.id) }
+  let(:source_package) { create(:package, :as_submission_source, name: 'ball', project: source_project) }
   let(:devel_project) { create(:project, name: 'devel:project') }
   let(:devel_package) { create(:package_with_file, name: 'goal', project: devel_project) }
-  let(:bs_request) { create(:bs_request, description: 'Please take this', creator: submitter.login) }
-  let(:bs_request_submit_action) do
-    create(:bs_request_action_submit, target_project: target_project.name,
-                                      target_package: target_package.name,
-                                      source_project: source_project.name,
-                                      source_package: source_package.name,
-                                      bs_request_id: bs_request.id)
+  let(:bs_request) do
+    create(:bs_request_with_submit_action,
+           description: 'Please take this', creator: submitter.login,
+           target_project: target_project.name,
+           target_package: target_package.name,
+           source_project: source_project.name,
+           source_package: source_package.name)
   end
   let(:request_with_review) do
-    create(:review_bs_request,
-           reviewer: reviewer,
+    create(:bs_request_with_submit_action,
+           review_by_user: reviewer,
            target_project: target_project.name,
            target_package: target_package.name,
            source_project: source_project.name,
@@ -67,8 +67,6 @@ RSpec.describe Webui::RequestController, vcr: true do
 
       before do
         login receiver
-        bs_request.bs_request_actions.delete_all
-        bs_request_submit_action
         get :show, params: { number: bs_request.number }
       end
 
@@ -80,7 +78,6 @@ RSpec.describe Webui::RequestController, vcr: true do
     context 'when there are no package maintainers' do
       before do
         login receiver
-        bs_request_submit_action
         get :show, params: { number: bs_request.number }
       end
 
@@ -101,7 +98,6 @@ RSpec.describe Webui::RequestController, vcr: true do
 
       shared_examples 'a full diff not requested for' do |file_name|
         before do
-          bs_request_submit_action
           get :show, params: { number: bs_request.number }
         end
 
@@ -142,7 +138,6 @@ RSpec.describe Webui::RequestController, vcr: true do
 
       shared_examples 'a full diff requested for' do |file_name|
         before do
-          bs_request_submit_action
           get :show, params: { number: bs_request.number, full_diff: true }
         end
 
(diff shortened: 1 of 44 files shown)


src/api/spec/factories/bs_requests.rb
```diff
@@ -23,7 +23,20 @@
         request.commenter = user.login
       end
       if request.bs_request_actions.none?
-        request.bs_request_actions << create(:bs_request_action, type: evaluator.type, source_project: evaluator.source_project)
+        if evaluator.type
+          request.bs_request_actions << create(:bs_request_action,
+                                               type: evaluator.type,
```
Ana06: why not `evaluator.type || set_bugowner` instead the `else`? :thinking:

coolo: I guess I will just require a type - and forbid to create a :bs_request

coolo: I did just that now - having a default in the parent factory just doesn't work out cleanly, so I changed the examples to create a :set_bugowner_request explicitly. Cleaner code.

src/api/app/views/staging/staging_projects/_missing_reviews.xml.builder
```diff
@@ -1,5 +1,5 @@
 builder.missing_reviews(count: count) do |missing_review|
   missing_reviews.each do |bs_request|
-    missing_review.entry(id: bs_request[:id], creator: bs_request[:by], state: bs_request[:state], package: bs_request[:package])
+    missing_review.entry(id: bs_request[:id], request: bs_request[:request], creator: bs_request[:by], state: bs_request[:state], package: bs_request[:package])
```
coolo: @vpereira this is a left over of me debugging something and I'll remove it - but this should be fixed independently of it.

Because the id of the review is super uninteresting - there is nothing you can do with it, there is no API for reviews (unfortunatly one might add), but what's missing is the request number.

src/api/spec/jobs/bs_request_action_webui_infos_job_spec.rb
```diff
@@ -45,11 +45,14 @@
...
-               source_package: source_package.name,
-               target_project: 'does-not-exist',
-               target_package: target_package.name)
+        request = build(:bs_request_with_submit_action,
+                        source_project: source_project.name,
+                        source_package: source_package.name,
+                        target_project: 'does-not-exist',
+                        target_package: target_package.name)
+        # avoid save!
+        request.save
```
coolo: this example is a little bizarre - and basically showing off why the PR got so large. Previously the actions were added behind the request's back. And as the really tricky validations are not in the action but in the request, this lead to all kind of 'impossible' requests created in rspec.

coolo:
@Ana06 please have a new look

vpereira:
would be nice to  have it merged before wed, when we will start a sprint working on the "Staging Workflow"

vpereira:
> you mean to avoid merging conflicts?

yes, its touching a lot of staging related specs, so would be better to have it on master before start it

src/api/spec/factories/bs_requests.rb
```diff
@@ -1,152 +1,116 @@
 FactoryBot.define do
   factory :bs_request do
     description { Faker::Lorem.paragraph }
-    state { 'new' }
+    state do |evaluator|
+      evaluator.reviews.present? ? :review : :new
```
Ana06: `[nitpick] evaluator.reviews.any? ? :review : :new`

src/api/spec/factories/project.rb
```diff
@@ -27,6 +27,12 @@
       project.write_to_backend
     end
 
+    trait :as_submission_source do
+      after(:create) do |project, _evaluator|
```
Ana06: [nitpick] `_` instead `_evaluator`

src/api/spec/jobs/bs_request_action_webui_infos_job_spec.rb
```diff
@@ -45,11 +45,14 @@
...
-               source_project: source_project.name,
-               source_package: source_package.name,
-               target_project: 'does-not-exist',
-               target_package: target_package.name)
+        request = build(:bs_request_with_submit_action,
+                        source_project: source_project.name,
+                        source_package: source_package.name,
+                        target_project: 'does-not-exist',
+                        target_package: target_package.name)
+        # avoid save!
```
Ana06: why? :confused: I have the comment this comment could be more explicit :unamused:

src/api/spec/models/staging/staging_project_spec.rb
```diff
@@ -159,6 +159,7 @@
   describe '#assign_managers_group' do
     context 'when the group wasn\'t assigned before' do
       before do
+        User.current = user
```
Ana06: why not `login(user)`? :thinking:

src/api/spec/models/staging/workflow_spec.rb
```diff
@@ -10,13 +11,18 @@
...
-           request_state: :review,
+           state: :review,
            target_project: project.name,
            target_package: target_package.name,
            source_project: source_project.name,
            source_package: source_package.name)
   end
 
+  before do
+    User.current = admin_user
```
Ana06: same here

src/api/spec/controllers/webui/request_controller_spec.rb
```diff
@@ -13,20 +13,20 @@
...
-  let(:bs_request) { create(:bs_request, description: 'Please take this', creator: submitter.login) }
-  let(:bs_request_submit_action) do
-    create(:bs_request_action_submit, target_project: target_project.name,
-                                      target_package: target_package.name,
-                                      source_project: source_project.name,
-                                      source_package: source_package.name,
-                                      bs_request_id: bs_request.id)
+  let(:bs_request) do
+    create(:bs_request_with_submit_action,
+           description: 'Please take this', creator: submitter.login,
```
Ana06: [nitpick] `.login` is not needed (and in the other 100 places where this happens).

src/api/spec/factories/project.rb
coolo: yeah, but often enough I still need it - and then it's easier to remove the _

src/api/spec/jobs/bs_request_action_webui_infos_job_spec.rb
coolo: the rspec test suite is *full* with invalid requests :(

src/api/spec/models/staging/staging_project_spec.rb
coolo: I guess I can do that. - but I found User.current in tons of model tests.

src/api/spec/controllers/webui/request_controller_spec.rb
coolo: See https://github.com/openSUSE/open-build-service/pull/6721 :)

coolo:
I take the login(user) changes into 6721 - need to rebase it anyway

PR #6708, opened 2019-01-05
Remove unused release_repository_dialog route

### LINKED ISSUE #6534: release_repository_dialog is not used
We should remove `release_repository_dialog` route. The action doesn't even exist in the controller.

ankitkataria:
- removes the `release_repository_dialog` routes from routes.rb as it is not
being used

Fixes #6534


ankitkataria:
Sorry about the force push, had to fix the commit message.

coolo:
No reason to be sorry - force pushes into development branches are the norm not the exception.

coolo:
that this breaks so easily is not a good sign for the test suite :(

PR #6714, opened 2019-01-07
Refactor architecture controller service

bgeuken:
Simplify how we bulk update architecture availability and
improve ruby dialect.


src/api/app/services/architectures_controller_service/architecture_updater.rb
```diff
@@ -1,16 +1,13 @@
...
+      @all_valid = false
     end
 
     def call
-      @archs.each do |name, value|
-        arch = Architecture.find_by(name: name)
-        arch.available = value
-        @all_valid &&= arch.save
-      end
+      @all_valid = @archs.to_h.map { |name, value| Architecture.where(name: name).update(available: value) }.flatten.all?(&:valid?)
```
Ana06: this was more efficient before... :sob: 

I would try to do less queries in the database instead... :thinking:

bgeuken: This table has as maximum 20 to 30 architectures. I doubt this has any relevant performance impact.

coolo: hmm, IMO it *are* less database queries.

I didn't check in the log, but before you had
SELECT * from architectures limit 1;
UPDATE architectures where id=X;

and now you have 
UPDATE architectures where name=X;

vpereira: right @coolo, @Ana06 could you please check it again and tell us concretely what are your concerns?

thank you

bgeuken: @coolo I think what @Ana06 means is that the `.all?(&:valid?)` causes another iteration, while before we would evaluate the validation result right away.

As a sidenote `all?` stops the iteration on the first falsy element.

vpereira: > @coolo I think what @Ana06 means is that the `.all?(&:valid?)` causes another iteration, while before we would evaluate the validation result right away.

but does it generates a new SQL query? looking in the log it doesnt look like. 

I think the step towards idiomatic code is the right decision here. Even if in a microbenchmarking the older code was faster, The number of architectures won't grow up in a way that we have to sacrifice code readability for speed.

coolo: I don't consider your code more readable to be honest.

For once: you call `valid?` on what exactly? On an architecture? On a string? On a hash element?

vpereira: > I don't consider your code more readable to be honest.

We can consider that more idiomatic code is more readable, since it's the kind of code that we are used to ready daily.

Mutation + ```&&=``` is for sure less idomatic ruby than method chaining.

> For once: you call `valid?` on what exactly? On an architecture? On a string? On a hash element?

The good thing about avoiding mutation and method chaining is that you can easily paste the one line in your console and remove the last call to see it

coolo: ... if it didn't had side effects.

vpereira: > ... if it didn't had side effects.

side effects and mutations aren't the same thing, but you know it.

Ana06: you are forgetting again Ruby... you go through the array many more times. I would leave it as it was before. I don't see the improvement.

What I meant with the queries is that it may be possible to update all the architectures at the same time.

vpereira: Hi @Ana06 
the other implementation that we came up with was:

```
...

def call
 @all_valid = @archs.to_h.map do |name, value|
                        arch = Architecture.find_by(name: name)	
                        arch.available = value	
                        arch.save
                      end
  self
 end

def valid?
 @all_valid.all?
end
```

Ana06: @vpereira I don't think you need to return `self`, but appart from that this suggestion LGTM :smile:

vpereira: leave my self alone

bgeuken: @vpereira @Ana06 I've updated the PR again. Sorry for the long delay.

 > Ana06 12 days ago Member
>
> What I meant with the queries is that it may be possible to update all the architectures at the same time.

We could do that with `update_all` but then the validations wouldn't run.

src/api/app/services/architectures_controller_service/architecture_updater.rb
```diff
@@ -1,21 +1,22 @@
...
+      @archs = params.require(:archs).permit!
+      @all_valid = false
     end
 
     def call
-      @archs.each do |name, value|
+      @all_valid = @archs.to_h.map do |name, value|
         arch = Architecture.find_by(name: name)
         arch.available = value
-        @all_valid &&= arch.save
```
Ana06: why did you remove this? This means going throug al the elements again in `.all?`

src/api/app/services/architectures_controller_service/architecture_updater.rb
```diff
@@ -1,21 +1,22 @@
...
-      @archs = archs
-      @all_valid = true
+    def initialize(params)
+      @archs = params.require(:archs).permit!
+      @all_valid = false
     end
 
     def call
-      @archs.each do |name, value|
+      @all_valid = @archs.to_h.map do |name, value|
```
Ana06: `to.h` + `map` is a not an efficient thing and you store objects instead of just a boolean in `@all_valid`... (you save one query to the database in exchange but...) as you say this efficiency is not important here, let's say this is a nitpick... but I still don't understand the benefit of the refactoring... (sorry @Ana06 of the past for [the contradiction](https://github.com/openSUSE/open-build-service/pull/6714#discussion_r246697320))

[for the @Ana06's of the future] I see a nice refactoring here once we update to Ruby 2.6 :full_moon:

PR #6720, opened 2019-01-08
Update influxdb-rails to beta 3

ChrisBr:
This will fix several issues and introduces new features:

Issues:
* Move count and cache_hits tags to values
* Do not track SCHEMA SQL queries

Features:
* Dynamic tags per request (we track for now ``anonymous``, ``beta`` and ``api`` tags).

### DIFF
--- a/src/api/app/controllers/application_controller.rb
+++ b/src/api/app/controllers/application_controller.rb
@@ -26,6 +26,7 @@ class ApplicationController < ActionController::Base
   before_action :shutup_rails
   before_action :validate_params
   before_action :require_login
+  before_action :set_influxdb_tags
 
   delegate :extract_user,
            :extract_user_public,
@@ -455,5 +456,16 @@ class ApplicationController < ActionController::Base
 
   def shutup_rails
     Rails.cache.silence! unless Rails.env.development?
+  end
+
+  def set_influxdb_tags
+    anonymous = User.current_login == User::NOBODY_LOGIN
+
+    InfluxDB::Rails.current.tags = {
+      beta: !anonymous && User.current.in_beta?,
+      anonymous: anonymous,
+      interface: :api,
+      volley: @skip_validation
+    }
   end
 end
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -18,6 +18,7 @@ class Webui::WebuiController < ActionController::Base
   before_action :check_anonymous
   before_action :require_configuration
   before_action :set_pending_announcement
+  before_action :set_influxdb_tags
   after_action :clean_cache
 
   # :notice and :alert are default, we add :success and :error
@@ -321,6 +322,17 @@ class Webui::WebuiController < ActionController::Base
   def set_pending_announcement
     return if Announcement.last.in?(User.current.announcements)
     @pending_announcement = Announcement.last
+  end
+
+  def set_influxdb_tags
+    anonymous = User.current_login == User::NOBODY_LOGIN
+
+    InfluxDB::Rails.current.tags = {
+      beta: !anonymous && User.current.in_beta?,
+      anonymous: anonymous,
+      interface: :webui,
+      volley: false # only relevant for API,
+    }
   end
 
   # NOTE: remove when bootstrap migration is done (related to switch_to_webui2)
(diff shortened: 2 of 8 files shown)


ChrisBr:
@hennevogel should we drop the data again from metrics.o.o?

src/api/app/controllers/application_controller.rb
```diff
@@ -456,4 +457,13 @@ def forward_from_backend(path)
   def shutup_rails
     Rails.cache.silence! unless Rails.env.development?
   end
+
+  def set_influxdb_tags
+    anonymous = User.current.present? && !User.current.is_nobody?
```
Ana06: [nitpick] `present?` is not needed

src/api/app/controllers/application_controller.rb
```diff
@@ -456,4 +457,13 @@ def forward_from_backend(path)
...
+    InfluxDB::Rails.current.tags = {
+      beta: anonymous && User.current.in_beta?,
```
Ana06: what happen if `User.current` is `nil`? This would break, you need `User.current.try(:in_beta?)`

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -323,6 +324,15 @@ def set_pending_announcement
     @pending_announcement = Announcement.last
   end
 
+  def set_influxdb_tags
+    anonymous = User.current.present? && !User.current.is_nobody?
```
Ana06: [nitpick] repeated code :see_no_evil:

src/api/app/controllers/application_controller.rb
coolo: aehm, what exactly is anonymous then? I find nobody more anonymous than !nobody (is that body then?)

src/api/app/controllers/application_controller.rb
coolo: anonymous is only true if there is a user (for whatever reason :)

src/api/app/controllers/application_controller.rb
ChrisBr: Jaja ... The conditional was wrong, fixed!

src/api/app/controllers/webui/webui_controller.rb
ChrisBr: I don't think it makes sense to abstract it now!

src/api/app/controllers/application_controller.rb
```diff
@@ -456,4 +457,14 @@ def forward_from_backend(path)
   def shutup_rails
     Rails.cache.silence! unless Rails.env.development?
   end
+
+  def set_influxdb_tags
+    anonymous = !User.current || (User.current && User.current.is_nobody?)
```
coolo: [nitpick] perhaps this is easier on the eyes:
```
anonymous = User.current_login == User::NOBODY_LOGIN

Comment (ChrisBr) [reply src/api/app/controllers/application_controller.rb]: Hm, we still need ``User.current || `` as it seems there are cases when User.current is nil.

**Comment by coolo** [comment]
may I suggest another tag for volleys? These mess up the average numbers alot - as they are only doing authentifications and then leave the rest to the backend.

File: src/api/app/controllers/application_controller.rb (hunk 4 above)
Comment (coolo) [reply src/api/app/controllers/application_controller.rb]: yes, but User.current_login handles that

**Comment by ChrisBr** [comment]
> may I suggest another tag for volleys? These mess up the average numbers alot - as they are only doing authentifications and then leave the rest to the backend.

Always appreciated but please specify a little bit more what you mean :smile:

**Comment by coolo** [comment]
git grep pass_to_backend

**Comment by ChrisBr** [comment]
Added the tags now also to ActiveJobs.

File: src/api/app/controllers/application_controller.rb [hunk 5]
```diff
@@ -456,4 +457,15 @@ def forward_from_backend(path)
   def shutup_rails
     Rails.cache.silence! unless Rails.env.development?
   end
+
+  def set_influxdb_tags
+    anonymous = User.current_login == User::NOBODY_LOGIN
```
Ana06: [nitpick] `User.curreny.try(:is_nobody?)` and written like that I think that the `anonymous variable is not needed

coolo: User.current.try(:is_nobody?) is false for User.current=nil - which we don't want.

PR #6721, opened 2019-01-08
 rspec: Simplify request factory parameters

coolo:
extending https://github.com/openSUSE/open-build-service/pull/6700 as review on it is delayed

### DIFF
--- a/src/api/spec/factories/bs_requests.rb
+++ b/src/api/spec/factories/bs_requests.rb
@@ -11,6 +11,13 @@ FactoryBot.define do
       create(:confirmed_user)
     end
 
+    initialize_with do
+      request = new(attributes)
+      # the sanitize part is a heavy requirement on a model
+      request.skip_sanitize
+      request
+    end
+
     reviews do |evaluator|
       ret = []
       ret << build(:review, by_project: evaluator.review_by_project) if evaluator.review_by_project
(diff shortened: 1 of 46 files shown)


src/api/spec/factories/bs_requests.rb
```diff
@@ -24,17 +31,42 @@
...
+
+      if attribs[:source_package].is_a?(Package)
+        attribs[:source_project] ||= attribs[:source_package].project.name
+        attribs[:source_package] = attribs[:source_package].name
+      end
+      if attribs[:target_package].is_a?(Package)
+        attribs[:target_project] ||= attribs[:target_package].project.name
+        attribs[:target_package] = attribs[:target_package].name
+      end
+      attribs[:type] = attribs[:type].to_s
```
Ana06: why the `to_s`? :thinking:

src/api/spec/factories/bs_requests.rb
```diff
@@ -66,6 +99,8 @@
     end
 
     after(:create) do |request, evaluator|
+      # the state will be overwritten by the constructor, so we need
+      # to set it this afterwards
```
Ana06: `it this`

src/api/spec/factories/bs_requests.rb
coolo: you don't want to know - really :)

But long story short: because the controller sets it as string as parsed out of XML. And it's really interesting how it breaks if you pass a symbol, but I find the rspec code easier to read if passing a symbol.

coolo:
fixed/extended the comments

PR #6722, opened 2019-01-08
Feeds controller refactoring

vpereira:
Refactor the action ```commits``` 

- Do Authorization with Pundit
- Extract controller logic to its own methods

### DIFF
--- a/src/api/app/policies/project_policy.rb
+++ b/src/api/app/policies/project_policy.rb
@@ -35,6 +35,10 @@ class ProjectPolicy < ApplicationPolicy
     user.can_modify?(record, true)
   end
 
+  def source_access?
+    User.current.is_admin? || !record.disabled_for?('sourceaccess', nil, nil)
+  end
+
   private
 
   def no_remote_instance_defined_and_has_not_remote_repositories?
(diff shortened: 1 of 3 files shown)


src/api/app/policies/project_policy.rb
```diff
@@ -35,6 +35,10 @@ def unlock?
     user.can_modify?(record, true)
   end
 
+  def source_access?
+    !User.current.is_admin? && record.disabled_for?('sourceaccess', nil, nil) ? false : true
```
coolo: isn't this a rather fancy way to implement the not operator ? :)

vpereira: ops, fixed, would you mind to check it again @coolo ?

src/api/app/policies/project_policy.rb
```diff
@@ -35,6 +35,10 @@ def unlock?
     user.can_modify?(record, true)
   end
 
+  def source_access?
+    !(!User.current.is_admin? && record.disabled_for?('sourceaccess', nil, nil))
```
coolo: f*** - you really try my limits here, no? :)

Because to me this is a pretty complicated way to write
`User.current.is_admin? || !record.disabled_for?('sourceaccess', nil, nil)`

vpereira: I'm not sure what you are talking about (now making sure that I saved the changes that I did with @DavidKang, one hour ago) :)

PR #6730, opened 2019-01-09
Improve InfluxDB

ChrisBr:
Several improvements for InfluxDB mostly concerning the (default) tags we sent. We now set default tags in the initializer making sure they're never empty (e.g. for DelayedJob processed in the background).

### DIFF
--- a/src/api/config/initializers/influxdb_rails.rb
+++ b/src/api/config/initializers/influxdb_rails.rb
@@ -15,7 +15,12 @@ InfluxDB::Rails.configure do |config|
   config.time_precision      = CONFIG['influxdb_time_precision']
   config.series_name_for_sql = 'rails.sql'
   config.tags_middleware = lambda do |tags|
-    # set a default location when e.g. using ActiveRecord outside of a request
-    { location: :raw }.merge(tags)
+    result = { beta: false, anonymous: true, interface: :none }.merge!(tags)
+    # TODO: workaround for https://github.com/influxdata/influxdb-rails/pull/64
+    result.reject! { |_, value| value.nil? || value == '' }
+    return result if result.key?(:method)
+
+    # set the default location for e.g. SQL calls outside a request
+    { location: :raw }.merge!(result)
   end
 end
--- a/src/api/config/initializers/influxdb_rails.rb
+++ b/src/api/config/initializers/influxdb_rails.rb
@@ -14,4 +14,8 @@ InfluxDB::Rails.configure do |config|
   config.use_ssl             = CONFIG['influxdb_ssl']
   config.time_precision      = CONFIG['influxdb_time_precision']
   config.series_name_for_sql = 'rails.sql'
+  config.tags_middleware = lambda do |tags|
+    # set a default location when e.g. using ActiveRecord outside of a request
+    { location: :raw }.merge(tags)
+  end
 end
--- a/src/api/app/controllers/application_controller.rb
+++ b/src/api/app/controllers/application_controller.rb
@@ -23,10 +23,10 @@ class ApplicationController < ActionController::Base
 
   # skip the filter for the user stuff
   before_action :extract_user
+  before_action :set_influxdb_tags
   before_action :shutup_rails
   before_action :validate_params
   before_action :require_login
-  before_action :set_influxdb_tags
 
   delegate :extract_user,
            :extract_user_public,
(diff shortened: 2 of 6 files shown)


src/api/config/initializers/influxdb_rails.rb
```diff
@@ -15,7 +15,12 @@
   config.time_precision      = CONFIG['influxdb_time_precision']
   config.series_name_for_sql = 'rails.sql'
   config.tags_middleware = lambda do |tags|
-    # set a default location when e.g. using ActiveRecord outside of a request
-    { location: :raw }.merge(tags)
+    # set a default tags
+    result = { beta: false, anonymous: true, interface: :none }.merge(tags)
+    result.reject! { |_, value| value.nil? || value == '' } # workaround for bug in influxdb-rails gem
```
Ana06: [nitpick] `value.nil? || value == ''` = `value.blank?`

[nitpick] maybe linking the bug and add a TODO so that this is fixed as some point? :thinking:

src/api/config/initializers/influxdb_rails.rb
```diff
@@ -15,7 +15,12 @@
...
+    return result if result.key?(:method)
+
+    # set the default location for e.g. SQL calls outside a request
+    { location: :raw }.merge(result)
```
Ana06: [nitpick] `merge!` would avoid create an extra hash

src/api/config/initializers/influxdb_rails.rb
```diff
@@ -14,4 +14,13 @@
   config.use_ssl             = CONFIG['influxdb_ssl']
   config.time_precision      = CONFIG['influxdb_time_precision']
   config.series_name_for_sql = 'rails.sql'
+  config.tags_middleware = lambda do |tags|
+    # set a default tags
+    result = { beta: false, anonymous: true, interface: :none }.merge(tags)
```
Ana06: [nitpick] `merge!` would avoid create an extra hash

src/api/config/initializers/influxdb_rails.rb
```diff
@@ -14,4 +14,13 @@
   config.use_ssl             = CONFIG['influxdb_ssl']
   config.time_precision      = CONFIG['influxdb_time_precision']
   config.series_name_for_sql = 'rails.sql'
+  config.tags_middleware = lambda do |tags|
+    # set a default tags
```
Ana06: _a default tags_? 

[Nitpick] To be honest, I don't think the comment is needed, I would remove it.

src/api/config/initializers/influxdb_rails.rb
ChrisBr: > [nitpick] value.nil? || value == '' = value.blank?

No, blank does also drop false but I want to keep false.

> [nitpick] maybe linking the bug and add a TODO so that this is fixed as some point? thinking

I will do as soon as I created the PR :smile:

ChrisBr:
@Ana06 updated!

ChrisBr:
@coolo so I played a little bit around and adding the request id as a field to InfluxDB could be a solution.

On the general dashboard we remove the tables and just have on table with the slowes N requests:

selection_020

This gets narrowed down when you zoom in. On this table we have links with the request id which send you to a dashboard with the request id as a variable:

image

There is just one problem: As the request id has a high cardinality we need to use a field. Query fields will be slow because these are not indexed. However, as we also use the time it already narrows down the possible data set. I think we need to try this out with real data if it works or not....

coolo:
hmm. can we store starttime and endttime of a request in a different (and as such smaller) table? Then we can use these to limit the queries in the big tables.

coolo:
... no, we can't. That would mean a join (and even on unindexed values) - which is just not supported.

So it would need to be part of the requests table and contained in the link as variable. Which leads to another problem: the *current* timeframe is appended in the link, which means we have the same problem there as we have now. We get to see the request data that fit the zoomed in request data.

Possibly influxdb/grafana is just not the right tool for such tracking :)

ChrisBr:
@coolo we can hand over the time in the link, see

https://github.com/grafana/grafana/issues/787

There is still just one small problem: In the table we could select the time from the request to create:

```
http://localhost:8000/d/PyWB-rLik/new-dashboard-copy?orgId=1&var-box=${__cell}&from=${__cell_0}&to=now
```

The ``__cell_0`` is the timestamp of the request. However, best would be if we can do something like ```
&from=${__cell_0 - 1h}&to=${__cell_0 + 1h}
```
so we catch everything what happened 1 hour before and 1 hour after the request. However, I haven't found a way yet to manipulate the data what we get from the cell :cry:

coolo:
why do you need an hour before and after? The request could very well know the first reported timestamp - and store it.

PR #6731, opened 2019-01-09
Set nobody when request is processed

ChrisBr:
When we authenticate a user, we set User.current = .
Howerver, after the request is processed we never set it to nil.
If the next request does not require authentication and skips the
extract_user before filter, we still have the user from the previous
request set in User.current.
We should clean up what we set in Thread.current.


ChrisBr:
Crossing fingers :crossed_fingers: nothing will break!

ChrisBr:
> Will this log the users out every time they visit a new page? thinking

No

coolo:
the minitest hack failed - but that's ok :)

PR #6732, opened 2019-01-09
Fix link for downloading a binary

bmwiedemann:
Without this patch,
https://build.opensuse.org/package/binary/home:bmwiedemann/git-deps/openSUSE_Leap_15.0/x86_64/git-deps-1-lp150.5.1.noarch.rpm
had

```html
<h3>
Detailed Information About git-deps-1-lp150.5.1.noarch.rpm
<a title="Download file" href="<a href=&quot;https://api.opensuse.org:443/build/home:bmwiedemann/openSUSE_Leap_15.0/x86_64/git-deps/git-deps-1-lp150.5.1.noarch.rpm&quot;>git-deps-1-lp150.5.1.noarch.rpm</a>"><i class='fas fa-download text-secondary'></i>
</a></h3>
```

### DIFF
--- a/src/api/app/views/webui2/webui/package/binary.html.haml
+++ b/src/api/app/views/webui2/webui/package/binary.html.haml
@@ -8,7 +8,7 @@
     %h3
       = @pagetitle
       - if @durl
-        = link_to(link_to(@filename, @durl), title: 'Download file') do
+        = link_to(@durl, title: 'Download') do
           %i.fas.fa-download.text-secondary
 
     %div


bmwiedemann:
@dmarcoux https://github.com/openSUSE/open-build-service/blob/master/src/api/app/helpers/webui/webui_helper.rb#L275 has similar syntax, except for the `do` - what is that intended to do?

dmarcoux:
The block (so everything under the `do`) is the content of the link, so what is displayed to the user. When there is a block, `link_to`'s first parameter is the URL of the link.

In this case, we don't need the filename. I will push the changes.

Issue #1771, opened 2016-05-02
Cannot login or manage users created with a space (OBS 2.6, 2.7RC)


Conan-Kudo:
On my private OBS 2.6 instance, a user has created a username with a space in it. Unfortunately, that user cannot log in with that user, and I can't (as Admin) delete it. It throws 404 errors every time.

Conan-Kudo:
This is still a problem with OBS 2.6.99 (OBS 2.7 RC).

bgeuken:
Duplicate or related with #1064

Ana06:
@openSUSE/open-build-service do we want to create users whose login has spaces? The main problem with that is that the home project would also have an space and we don't allow project names with spaces. So to support that we would have to replace the spaces somehow in the home project name. I think it would be much easier to just forbid that the name has spaces. It is anyway strange to have a login with spaces.

Ana06:
The problem with spaces is now different as when this was open, it is not possible anymore to create user with spaces (fails because of the problem with the project).

saraycp:
@Conan-Kudo please let us know if you still have problems with this and will try to find a solution.

Since 2.6 is no longer maintained and the login can't contain spaces anymore, we are going to close this issue. Please, feel free to re-open it if needed.

Issue #1064, opened 2015-08-27
Login to obs using username with dot not possible


c-hoeger:
whenever I create a user containing a dot, e.g. carsten.hoeger, I cannot login with that user.

linux:~ # osc api /person
<directory count="4">
 <entry name="Admin"/>
 <entry name="_nobody_"/>
 <entry name="choeger"/>
 <entry name="carsten.hoeger"/>
</directory>

login using choeger works, login using carsten.hoeger doesn’t.

Is that really the case or am I hitting a bug here?

==> /srv/www/obs/api/log/production.log <==
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.48] Started GET "/" for 172.16.210.1 at 2015-08-26 13:27:26 +0000
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.49] Processing by Webui::MainController#index as HTML
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.49] Anonymous request to /
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.50]   Rendered webui/main/_systemstatus.html.erb (2.2ms)
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.50]   Rendered shared/_sign_up.html.erb (1.0ms)
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.50]   Rendered webui/main/_sponsors.html.erb (0.0ms)
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.50]   Rendered webui/main/_news.html.erb (0.0ms)
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.51]   Rendered webui/main/index.html.erb within layouts/webui/webui (7.8ms)
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.51]   Rendered layouts/webui/_watch_and_search.html.erb (0.4ms)
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.51]   Rendered layouts/webui/_header.html.erb (0.8ms)
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.51]   Rendered layouts/webui/_personal_navigation.html.erb (1.2ms)
[6abee853-7581-4f09-8415-0b481354ca0f] [1457:17274.51]   Rendered layouts/webui/_flash.html.erb (0.0ms)
[...]

PR #6745, opened 2019-01-10
Do not allow user to have a login with invalid characters

Ana06:
At the moment creating a user whose login has certain characters fails with a
500 because the home project (created after saving the user) is invalid as the
validation is stricter there.

image

image

Fixes https://github.com/openSUSE/open-build-service/issues/1771

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -76,7 +76,7 @@ class User < ApplicationRecord
 
   validates :login,
             format: { with: Regexp.new("\\A(#{NOBODY_LOGIN}|[-+a-zA-Z0-9][-+\\w\\.]*)\\z"),
-                      message: 'must not contain invalid characters' }
+                      message: 'only allows letters, numbers and the following characters: _ - + . (. and _ not at the beginning)' }
   validates :login,
             length: { in: 2..100, allow_nil: true,
                       too_long: 'must have less than 100 characters',
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -74,9 +74,14 @@ class User < ApplicationRecord
   validates :login,
             uniqueness: { message: 'is the name of an already existing user' }
 
+  # returns the login validation patern and the errors message
+  def self.login_pattern
+    ['[-+a-zA-Z0-9][-+\w\.]*', 'only allows letters, numbers and the following characters: _ - + . (. and _ not at the beginning)']
+  end
+
   validates :login,
-            format: { with: Regexp.new("\\A(#{NOBODY_LOGIN}|[-+a-zA-Z0-9][-+\\w\\.]*)\\z"),
-                      message: 'only allows letters, numbers and the following characters: _ - + . (. and _ not at the beginning)' }
+            format: { with: Regexp.new("\\A(#{NOBODY_LOGIN}|#{User.login_pattern.first})\\z"),
+                      message: User.login_pattern.second }
   validates :login,
             length: { in: 2..100, allow_nil: true,
                       too_long: 'must have less than 100 characters',
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -75,7 +75,7 @@ class User < ApplicationRecord
             uniqueness: { message: 'is the name of an already existing user' }
 
   validates :login,
-            format: { with: %r{\A[\w \$\^\-\.#\*\+&'"]*\z},
+            format: { with: Regexp.new("\\A(#{NOBODY_LOGIN}|[-+a-zA-Z0-9][-+\\w\\.]*)\\z"),
                       message: 'must not contain invalid characters' }
   validates :login,
             length: { in: 2..100, allow_nil: true,
--- a/src/api/app/views/webui2/shared/_sign_up.html.haml
+++ b/src/api/app/views/webui2/shared/_sign_up.html.haml
@@ -6,14 +6,15 @@
 - else
   = form_tag({ controller: 'user', action: 'register', method: :post }, class: 'sign-up', autocomplete: 'off') do
     .form-group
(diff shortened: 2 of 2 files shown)


src/api/app/models/user.rb
```diff
@@ -75,7 +75,7 @@ class User < ApplicationRecord
             uniqueness: { message: 'is the name of an already existing user' }
 
   validates :login,
-            format: { with: %r{\A[\w \$\^\-\.#\*\+&'"]*\z},
+            format: { with: %r{\A[\w\$\^\-\.#\*\+&'"]*\z},
                       message: 'must not contain invalid characters' }
```
dmarcoux: Could we improve this message by including the invalid characters? This way, users know what to change. Otherwise, it's just a shot in the dark.

coolo: or have the same pattern on the HTML input?
https://www.w3schools.com/tags/att_input_pattern.asp

Ana06: or both :rofl: I do it :+1:

Ana06:
@dmarcoux @coolo please re-review it

src/api/app/models/user.rb
```diff
@@ -75,7 +75,7 @@ class User < ApplicationRecord
             uniqueness: { message: 'is the name of an already existing user' }
 
   validates :login,
-            format: { with: %r{\A[\w \$\^\-\.#\*\+&'"]*\z},
+            format: { with: Regexp.new("\\A(#{NOBODY_LOGIN}|[-+a-zA-Z0-9][-+\w\.]*)\\z"),
```
eduardoj: Nitpick: Shouldn't `\w\.` be `\\w\\.`?

src/api/app/models/user.rb
```diff
@@ -74,9 +74,14 @@ class User < ApplicationRecord
   validates :login,
             uniqueness: { message: 'is the name of an already existing user' }
 
+  # returns the login validation patern and the errors message
+  def self.login_pattern
+    ['[-+a-zA-Z0-9][-+\w\.]*', 'only allows letters, numbers and the following characters: _ - + t . (. and _ not at the beginning)']
```
dmarcoux: only allows letters, numbers and the following characters: _ - + **t** .

Hum... what is **t** doing there?

Ana06: I don't know... :confounded:

Ana06:
@openSUSE/open-build-service what do we want to do with the possible users affected? data migration? script? nothing?

Ana06:
@openSUSE/open-build-service what do we want to do with the possible users affected? data migration? script? nothing (the admin just fix manually those strange cases)?

In the case of a migration/script, we need to know what we want to do with those users. Two options:
1. assigning them a random login
2. removing them

I would personally don't do anything as those can of users don't even have a home. I would let admins to take care of this problem. In case we don't want this, I would go for a data migration that after a confirmation remove the users.

dmarcoux:
I prefer the data migration route, then sending an email to the user to let them know that their login changed.

Ana06:
@dmarcoux 

> I prefer the data migration route, then sending an email to the user to let them know that their login changed.


I don't think you can change the login of a user that easy (what will you use instead?, will the user agree (login can be something really personal)?, what do you do with possible dependencies with the name?) and I don't think it is worthwhile. Those user can not really use much in OBS. That's why I prefer to handle it manually. The migration would be in my opinion to get rid of the user.

dmarcoux:
@Ana06 Depending on the timestamp in `last_logged_in_at` for these users, I would be fine with simply removing them.

Ana06:
BTW, this also fixes https://github.com/openSUSE/open-build-service/issues/1064

Ana06:
@dmarcoux 

> @Ana06 Depending on the timestamp in `last_logged_in_at` for these users, I would be fine with simply removing them.

But the problem is still that if for some cases you don't remove then, the alternative should be done manually. We have in our instance 8 users with this problem. I think that handling this manually is acceptable. We could have a migration that ask for confirmation to remove the users or allow to cancel the migration, solve the problem and then run again. As said, those users are not really usable, they cna not have a home, so I don't think we need a complicated solution.

marcus-h:
On 2019-02-18 04:05:46 -0800, Ana María Martínez Gómez wrote:
> > @Ana06 Depending on the timestamp in `last_logged_in_at` for these users, I would be fine with simply removing them.
> 
> But the problem is still that if for some cases you don't remove then, the alternative should be done manually. We have in our instance 8 users with this problem. I think that handling this manually is acceptable. We could have a migration that ask for confirmation to remove the users or allow to cancel the migration, solve the problem and then run again. As said, those users are not really usable, they cna not have a home, so I don't think we need a complicated solution.
> 
Hmm this argument is IMHO quite "weak": such a user can do everything,
except working in her/his/its home project:)

(Otherwise, I have no real opinion on this... whatever we do, we should
document it in the release notes.)

Ana06:
I am not sure what is the state of this.... @openSUSE/open-build-service what should we do with this PR?

vpereira:
@Ana06  I would accept this PR as it is. We can investigate and define how to handle the 8 cases in another place and that's IMO unrelated with this PR.

dmarcoux:
Could we document this change in the [release notes](https://github.com/openSUSE/open-build-service/blob/master/ReleaseNotes-2.10) for `2.10`?

Having the data migration (or another strategy) in another PR is fine for me.

vpereira:
could you review it again @bgeuken ? Houw Houw Houw :gorilla: :gorilla:

Ana06:
@openSUSE/open-build-service I am not sure what the state of this PR is. I won't have time in the close feature to address the feedback, but if someone else would like to do it, please feel free to reopen it.

PR #6747, opened 2019-01-11
Experiment: patch gems

coolo:
Need to see it in obs


coolo:
ok, this doesn't really help us anyway as you can't use the new API in covered code as the test suite runs against bundler - which doesn't care what we patch in our rpms.

PR #6757, opened 2019-01-13
Ready requests take reviews into account

saraycp:
A request is «ready» when its state is «new» and it has some reviews and
at least one of them is in state «accepted».

### DIFF
--- a/src/api/app/models/staging/workflow.rb
+++ b/src/api/app/models/staging/workflow.rb
@@ -22,7 +22,7 @@ class Staging::Workflow < ApplicationRecord
     end
 
     def ready_to_stage
-      in_states('new')
+      in_states(:new)
     end
   end
 
--- a/src/api/app/models/staging/workflow.rb
+++ b/src/api/app/models/staging/workflow.rb
@@ -18,7 +18,7 @@ class Staging::Workflow < ApplicationRecord
   has_many :target_of_bs_requests, through: :project, foreign_key: 'staging_workflow_id' do
     def stageable
       managers_group_title = proxy_association.owner.managers_group.try(:title)
-      includes(:reviews).where(state: :review, reviews: { state: :new, by_group: managers_group_title })
+      includes(:reviews).where(state: :review, staging_project_id: nil, reviews: { state: :new, by_group: managers_group_title })
     end
 
     def ready_to_stage
--- a/src/api/app/models/staging/workflow.rb
+++ b/src/api/app/models/staging/workflow.rb
@@ -33,6 +33,7 @@ class Staging::Workflow < ApplicationRecord
   validates :managers_group, presence: true
 
   after_create :create_staging_projects
+  after_create :add_reviewer_group
   before_update :update_staging_projects_managers_group
 
   def unassigned_requests
@@ -76,5 +77,11 @@ class Staging::Workflow < ApplicationRecord
     # FIXME: This assignation is need because after store a staging_project
     # the object is reloaded and we lost the changes.
     self.managers_group = new_managers_group
+  end
+
+  def add_reviewer_group
+    role = Role.find_by_title('reviewer')
+    project.relationships.find_or_create_by(group: managers_group, role: role)
+    project.store
   end
 end
--- a/src/api/spec/models/staging/workflow_spec.rb
+++ b/src/api/spec/models/staging/workflow_spec.rb
@@ -33,15 +33,7 @@ RSpec.describe Staging::Workflow, type: :model do
       it { expect(subject).to be_empty }
     end
 
-    context 'with requests without reviews by the staging managers group' do
-      before { bs_request }
-
-      it { expect(subject).to be_empty }
-    end
-
     context 'with requests with reviews by the staging managers group' do
-      let!(:review) { create(:review, by_group: group.title, bs_request: bs_request) }
-
       it { expect(subject).to contain_exactly(bs_request) }
     end
 
--- a/src/api/spec/models/staging/workflow_spec.rb
+++ b/src/api/spec/models/staging/workflow_spec.rb
@@ -8,10 +8,10 @@ RSpec.describe Staging::Workflow, type: :model do
   let(:staging_project) { staging_workflow.staging_projects.first }
   let(:source_project) { create(:project, name: 'source_project') }
   let(:target_package) { create(:package, name: 'target_package', project: project) }
-  let(:source_package) { create(:package, name: 'source_package', project: source_project) }
+  let(:source_package) { create(:package, :as_submission_source, name: 'source_package', project: source_project) }
   let(:bs_request) do
     create(:bs_request_with_submit_action,
-           state: :review,
+           review_by_group: group, # FIXME: force to create a review for the group, because skip_sanitize is avoiding it
            target_package: target_package,
            source_package: source_package)
   end
@@ -84,16 +84,18 @@ RSpec.describe Staging::Workflow, type: :model do
       it { expect(subject).to be_empty }
     end
 
-    context 'with requests and some are in staging projects and some not' do
-      let!(:bs_request_2) do
+    context 'with request in state review whether they are in staging projects or not' do
+      let(:bs_request_2) do
         create(:bs_request_with_submit_action,
+               review_by_group: group, # FIXME: force to create a review for the group, because skip_sanitize is avoiding it
                target_package: target_package,
                source_package: source_package)
       end
 
       before do
-        bs_request.staging_project = staging_project
-        bs_request.save
+        bs_request
+        bs_request_2.change_review_state(:accepted, by_group: group.title)
+        bs_request_2.update!(staging_project: staging_project)
       end
 
       it { expect(subject).to contain_exactly(bs_request_2) }
--- a/src/api/spec/models/staging/workflow_spec.rb
+++ b/src/api/spec/models/staging/workflow_spec.rb
@@ -21,6 +21,11 @@ RSpec.describe Staging::Workflow, type: :model do
     staging_workflow
   end
 
+  context 'when created' do
+    let(:role) { Role.find_by_title('reviewer') }
+    it { expect(project.relationships.where(group: group, role: role)).to exist }
+  end
+
   describe '#unassigned_requests' do
     subject { staging_workflow.unassigned_requests }
 


saraycp:
[Ready staged requests takes into account the reviews](https://trello.com/c/2MHYM5ve/1011-ready-staged-requests-takes-into-account-the-reviews)

src/api/app/models/staging/workflow.rb
```diff
@@ -22,7 +22,8 @@ def stageable
     end
 
     def ready_to_stage
```
coolo: this now duplicates 90% of the `stagable` code - and taking that both scopes are called in one function each, which duplicates even more code... Perhaps this can be better expressed with a parameter to the state review?

coolo:
> A request is «ready» when its state is «new» and it has some reviews and at least one of them is in state «accepted».

I wonder whose definition this is and under what circumstances does it matter? Because:

- if the staging manager group is reviewer in the target project, there is no way to get it into 'new' without passing its review. So there is no gain.
- if the staging manager group is not reviewer in the target project, then the definition of 'backlog' is wrong and it makes me wonder how you envision that the request even *gets* the staging manager's review.

So all in all, that definition doesn't make sense to me.

vpereira:
> I wonder whose definition this is and under what circumstances does it matter? Because:
> 
> * if the staging manager group is reviewer in the target project, there is no way to get it into 'new' without passing its review. So there is no gain.
> * if the staging manager group is not reviewer in the target project, then the definition of 'backlog' is wrong and it makes me wonder how you envision that the request even _gets_ the staging manager's review.
> 
> So all in all, that definition doesn't make sense to me.

Hi @coolo, I think this defintion came from the Dashboard code itself.. So if it's not right, could you help us to define what means the state "Ready" ?

From the: https://github.com/openSUSE/open-build-service/wiki/Staging-Workflow#request-states

> ready: Requests which were once in the backlog, got their reviews accepted (by the many bots). The > ready requests have the state new. All ready requests can be found here: 
> https://build.opensuse.org/project/requests/openSUSE:Factory?state=new

coolo:
There you have your answer: for the documented staging workflow, requests in state new are 'ready' - which is exactly what master implements. So: 'there is no gain' with this PR.

coolo:
> But a new request arriving to openSUSE:Factory (openSUSE:Factory is target project), without any review associated, wouldn't it be in the "new" state? 

It's impossible to create a request without reviews for openSUSE:Factory - that's what the project reviewers are for.

vpereira:
> @coolo, thanks for your comments.
> 
> It seems the point is that we are not creating a review right after the BsRequest is created. I think something specific from factory creates reviews for each new BsRequest.
> 
> We are checking deeply.

AFAIK, that's because there is the factory-bot, a client code which assigns a new review to the incoming requests

coolo:
> It seems the point is that we are not creating a review right after the BsRequest is created. I think something specific from factory creates reviews for each new BsRequest.

https://openbuildservice.org/help/manuals/obs-reference-guide/cha.obs.request_and_review_ystem.html#id-1.7.9.5.7.4

saraycp:
@coolo  and @vpereira can you please have another look to this? The key now is: 
A new review was automatically created only if the request is sent to a project that already had a reviewer group associated. To get this, we set the reviewer group to the project right after the staging workflow is created.

coolo:
that makes much more sense now!

vpereira:
> @coolo and @vpereira can you please have another look to this? The key now is:
> A new review was automatically created only if the request is sent to a project that already had a reviewer group associated. To get this, we set the reviewer group to the project right after the staging workflow is created.

great! So just to review the workflow, to see if I've got it right:

- We create a staging workflow for a project (let's say openSUSE:Factory)
- One of the required steps will be to choose a group to add as reviewer for openSUSE:Factory
- As soon it is configured, all new requests to openSUSE:Factory will have automatically a review associated to this group

Is that right @saraycp ?

saraycp:
Exactly @vpereira. Regarding your sentence
> One of the required steps will be to choose a group to add as reviewer for openSUSE:Factory
we don't have to do it manually. It's done by an _after_create_ callback on staging workflow model. It really takes the staging_workflow's `managers_group` (it certainly exists) and add this group as reviewer to the project.

src/api/spec/controllers/webui/staging/workflows_controller_spec.rb
```diff
@@ -46,6 +47,8 @@
       it { expect(response).to redirect_to(staging_workflow_path(project.staging)) }
       it { expect(subject.managers_group).to eq(managers_group) }
       it { expect(flash[:success]).not_to be_nil }
+      it { expect(reviewer.group).to eq(managers_group) }
+      it { expect(reviewer).not_to be_nil }
```
bgeuken: Maybe it's just me I would find 
```ruby
it { expect(project.relationships.where(group: managers_group, role: Role.find_by_title('reviewer')).to exist }
```
easier to understand:thinking:

Though I would move this to the model tests. Since it's testing model behaviour.

src/api/spec/controllers/webui/staging/workflows_controller_spec.rb
```diff
@@ -46,6 +47,8 @@
       it { expect(response).to redirect_to(staging_workflow_path(project.staging)) }
       it { expect(subject.managers_group).to eq(managers_group) }
       it { expect(flash[:success]).not_to be_nil }
+      it { expect(reviewer.group).to eq(managers_group) }
```
bgeuken: That seems to be a duplication of the query you run in the let in line 38.

src/api/spec/controllers/webui/staging/workflows_controller_spec.rb
saraycp: Changed! But `.to exist` can not be used with an `ActiveRecord::AssociationRelation`. I've checked if there is one element.

src/api/spec/controllers/webui/staging/workflows_controller_spec.rb
saraycp: Changed as you proposed in the previous comment

src/api/app/models/staging/workflow.rb
```diff
@@ -18,11 +18,11 @@ def without_staged_requests
   has_many :target_of_bs_requests, through: :project, foreign_key: 'staging_workflow_id' do
     def stageable
       managers_group_title = proxy_association.owner.managers_group.try(:title)
-      includes(:reviews).where(state: :review, reviews: { state: :new, by_group: managers_group_title })
+      includes(:reviews).where(state: :review, staging_project_id: nil, reviews: { state: :new, by_group: managers_group_title })
```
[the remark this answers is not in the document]
saraycp: `staging_project_id` is in `BsRequest`. This is checking that the request is not associated to a staging project.

src/api/spec/models/staging/workflow_spec.rb
```diff
@@ -92,10 +84,7 @@
                source_package: source_package.name)
       end
 
-      before do
-        bs_request.staging_project = staging_project
```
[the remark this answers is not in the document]
saraycp: I check again that requests are «ready» whether they are associated to a staging project or not, as it was before, but using `bs_request_2` instead of `bs_request` which is always rejected because it is not in state `new`.

We also realized that the test was wrong. Now it is ok, «ready» requests are those which state `new` and all its reviews `accepted`. The solution is a bit tricky so please have another look and give some suggestions if needed. @Ana06 @bgeuken @coolo

src/api/spec/models/staging/workflow_spec.rb
```diff
@@ -23,22 +23,19 @@
     staging_workflow
   end
 
+  context 'when created' do
+    let(:role) { Role.find_by_title('reviewer') }
+    it { expect(project.relationships.where(group: group, role: role).size).to eq(1) }
```
coolo: [nitpick] the flag tests use the form `expect(project.relationships.where(group: group, role: role)).to exist`, which is a little easier.

src/api/spec/models/staging/workflow_spec.rb
```diff
@@ -83,21 +80,27 @@
...
                source_project: source_project.name,
                source_package: source_package.name)
       end
 
       before do
-        bs_request.staging_project = staging_project
-        bs_request.save
+        bs_request
+        bs_request_2.change_review_state(:accepted, by_group: group.title)
+        # A review by project is created by default. See BsRequest#apply_default_reviewers. We have to accept it too.
```
coolo: Create the source_package with :as_submission_source then this doesn't happen

src/api/spec/models/staging/workflow_spec.rb
```diff
@@ -83,21 +80,27 @@
...
+        bs_request_2.change_review_state(:accepted, by_project: source_project.name)
+        bs_request_2.staging_project = staging_project
+        bs_request_2.save
       end
 
       it { expect(subject).to contain_exactly(bs_request_2) }
+      it { expect(subject.first.state).to eq(:new) }
```
coolo: as you already verified that it's only bs_request_2, you're testing change_review_state here and not the staging controller. As such: I see no value of the 2nd expect

src/api/app/models/staging/workflow.rb
```diff
@@ -77,4 +78,11 @@ def update_staging_projects_managers_group
     # the object is reloaded and we lost the changes.
     self.managers_group = new_managers_group
   end
+
+  def add_reviewer_group
+    role = Role.find_by_title('reviewer')
+    return unless role
```
coolo: [nitpick] if that role does not exist, a lot of other code crashes. You're a little over-defensive here

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
```diff
@@ -97,7 +97,7 @@
             assert_select 'entry', 2
           end
           assert_select 'missing_reviews', 1 do
-            assert_select 'entry', 1
+            assert_select 'entry', 4
```
coolo: hmm, so it's reporting the staging manager group's reviews as missing reviews? That sounds like a bug.

src/api/spec/models/staging/workflow_spec.rb
saraycp: Good tip! :heart: Thanks!

src/api/spec/models/staging/workflow_spec.rb
saraycp: done!

src/api/spec/controllers/webui/staging/workflows_controller_spec.rb
saraycp: @bgeuken I think I did something wrong, I've tried again and it worked. :clap: Updated!

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
coolo: can you track this please? It's unrelated to this PR, but it shouldn't be dropped.

coolo: looks like there is no need as your test fails now for being 1 :)

src/api/spec/models/staging/workflow_spec.rb
```diff
@@ -79,16 +76,21 @@
...
+               source_package: source_package.name)
       end
 
       before do
-        bs_request.staging_project = staging_project
-        bs_request.save
+        bs_request
+        bs_request_2.change_review_state(:accepted, by_group: group.title)
+        bs_request_2.staging_project = staging_project
+        bs_request_2.save
```
bgeuken: Please always use `save!` in tests. So that we notice if some test preparation goes wrong.

You could also squash the two lines by using `update!`

src/api/spec/models/staging/workflow_spec.rb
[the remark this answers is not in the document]
bgeuken: Right now the test fails for me with `Review::NotFoundError`

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
saraycp: @coolo, after rebasing  [this PR](https://github.com/openSUSE/open-build-service/pull/6721), everything changed. Caused by `skip_sanitize` on [this line](https://github.com/coolo/open-build-service/blob/908bf1d850bbd9eb2dcfcccd8e3ef40c023f8601/src/api/spec/factories/bs_requests.rb#L17)

I'm going to fix all related with this PR first and then _missing_reviews._

src/api/spec/models/staging/workflow_spec.rb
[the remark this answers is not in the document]
saraycp: @bgeuken rebasing [this PR](https://github.com/openSUSE/open-build-service/pull/6721) is causing the failures... working on this...

src/api/spec/models/staging/workflow_spec.rb
```diff
@@ -79,16 +77,21 @@
...
-    context 'with requests and some are in staging projects and some not' do
-      let!(:bs_request_2) do
+    context 'with request in state review whether they are in staging projects or not' do
+      let(:bs_request_2) do
         create(:bs_request_with_submit_action,
-               target_package: target_package,
-               source_package: source_package)
+               state: :review,
+               review_by_group: group, # FIXME: force to create a review for the group, because skip_sanitize is avoiding it
+               target_project: project.name,
```
coolo: please don't revert my changes - aka don't pass .name but the object and don't repeat the projects, but let the bsrequest factory handle that.

I'm not sure what you mean with the FIXME, but I'm happy to have a look

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
coolo: the reviews are created in the factory, see https://github.com/coolo/open-build-service/blob/908bf1d850bbd9eb2dcfcccd8e3ef40c023f8601/src/api/spec/factories/bs_requests.rb#L24

PR #6759, opened 2019-01-14
Add method that allows copying of staging projects and (some of) their relations

bgeuken:

### DIFF
--- a/src/api/app/models/staging/staging_project.rb
+++ b/src/api/app/models/staging/staging_project.rb
@@ -9,6 +9,26 @@ module Staging
     after_destroy :update_staging_workflow_on_backend
     before_create :add_managers_group
 
+    def copy(new_project_name)
+      transaction do
+        new_project = deep_clone(include: [:flags], skip_missing_associations: true)
+        new_project.name = new_project_name
+        new_project.config.save({ user: User.current, comment: "Copying project #{name}" }, config.content)
+        new_project.save!
+
+        repositories.each { |repository| repository.copy_to(new_project) }
+
+        # We can't use deep_clone here because of an exception raised in Relationship#add_group
+        relationships.each do |relationship|
+          new_project.relationships.find_or_create_by!(relationship.slice(:role_id, :user_id, :group_id))
+        end
+
+        new_project.store
+
+        new_project
+      end
+    end
+
     def classified_requests
       requests = (requests_to_review | staged_requests.includes(:reviews)).map do |request|
         {
--- a/src/api/Gemfile
+++ b/src/api/Gemfile
@@ -80,6 +80,8 @@ gem 'builder'
 gem 'experimental-influxdb-rails', '>=1.0.0.beta5'
 # for client side time ago
 gem 'rails-timeago', '~> 2.0'
+# for copying objects with their relations
+gem 'deep_cloneable', '~> 2.4.0'
 
 group :development, :production do
   # to have the delayed job daemon
--- a/src/api/Gemfile.lock
+++ b/src/api/Gemfile.lock
@@ -125,6 +125,8 @@ GEM
     data_migrate (5.3.0)
       rails (>= 4.2)
     database_cleaner (1.7.0)
+    deep_cloneable (2.4.0)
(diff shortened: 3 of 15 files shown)


src/api/app/models/download_repository.rb
```diff
@@ -15,7 +15,7 @@ class DownloadRepository < ApplicationRecord
   def architecture_inclusion
     # Workaround for rspec validation test (validate_presence_of(:repository_id))
     return unless repository
-    return if repository.architectures.pluck(:name).include?(arch)
+    return if repository.repository_architectures.any? { |a| a.architecture.name == arch }
```
coolo: that shouts N+1 to me.

How about repository.repository_architectures.where(architecture: Architecture.where(name: arch)).exists?

that translates into e.g.
```
  RepositoryArchitecture Exists (0.8ms)  SELECT  1 AS one FROM `repository_architectures` WHERE `repository_architectures`.`repository_id` = 1853016 AND `repository_architectures`.`architecture_id` IN (SELECT `architectures`.`id` FROM `architectures` WHERE `architectures`.`name` = 'x86_64') LIMIT 1
```

while your code translates into:

```
  RepositoryArchitecture Load (0.5ms)  SELECT `repository_architectures`.* FROM `repository_architectures` WHERE `repository_architectures`.`repository_id` = 98 ORDER BY position
  Architecture Load (0.4ms)  SELECT  `architectures`.* FROM `architectures` WHERE `architectures`.`id` = 2 LIMIT 1
  Architecture Load (0.4ms)  SELECT  `architectures`.* FROM `architectures` WHERE `architectures`.`id` = 1 LIMIT 1
```

(and for completness the original code):
```
   (0.9ms)  SELECT `architectures`.`name` FROM `architectures` INNER JOIN `repository_architectures` ON `architectures`.`id` = `repository_architectures`.`architecture_id` WHERE `repository_architectures`.`repository_id` = 98 ORDER BY position

File: src/api/app/models/staging/staging_project.rb [hunk 2]
```diff
@@ -9,6 +9,18 @@ class StagingProject < Project
     after_destroy :update_staging_workflow_on_backend
     before_create :add_managers_group
 
+    def copy(new_project_name)
+      new_project = deep_clone(
+        include: [:flags, { repositories: [:path_elements, :repository_architectures, :download_repositories] }], skip_missing_associations: true
```
coolo: you can't use deep_clone for repositories either (assuming this is for real life staging projects).

Spot the Bs in https://build.opensuse.org/projects/openSUSE:Leap:15.1:Staging:B/meta

src/api/app/models/download_repository.rb
bgeuken: ` repository.repository_architectures.where(architecture: Architecture.where(name: arch)).exists? ` wouldn't work with deep_cloneable, because the related architectures are not necessarily saved yet (when cloning they are not).

coolo: then better stay with pluck but pluck the architecture_id

bgeuken:
["create staging project" from a template](https://trello.com/c/diFikqV6/1018-create-staging-project-from-a-template)

src/api/app/models/download_repository.rb
bgeuken: This code runs at a time when the data might not be saved yet. So we can not use `pluck` to fetch thing from the DB.

src/api/app/models/staging/staging_project.rb
coolo: I'm not saying it might, I'm saying it will.

src/api/app/models/download_repository.rb
bgeuken: That worked... though I used the architecture_id instead so that we don't fetch the architecture each time.

coolo: ah, I only checked the 'in memory' case.

src/api/app/models/staging/staging_project.rb
bgeuken: No, you are saying it could:wink:

I am having a look at it.

PR #6760, opened 2019-01-14
Show reviewers gravatar on requests in the staging workflow

krauselukas:
screenshot_2019-01-14_13-33-46

### DIFF
--- a/src/api/app/helpers/webui/staging/workflow_helper.rb
+++ b/src/api/app/helpers/webui/staging/workflow_helper.rb
@@ -43,27 +43,27 @@ module Webui::Staging::WorkflowHelper
     end
   end
 
-  def reviewers_gravatar(request)
+  def reviewers_gravatar(request, users_hash, groups_hash)
     missing_reviews = request[:missing_reviews]
     image_tags = []
     missing_reviews.each do |review|
       case review[:review_type]
       when 'by_group'
-        object = Group.find_by_title(review[:by])
+        object = groups_hash[review[:by]]
       when 'by_user'
-        object = User.find_by_login(review[:by])
+        object = users_hash[review[:by]]
       end
       image_tags << image_tag_for(object, size: 20)
     end
     image_tags
   end
 
-  def create_request_links(request)
+  def create_request_links(request, users_hash, groups_hash)
     css = 'ready'
     css = 'review' if request[:missing_reviews].present?
     css = 'obsolete' if request[:state].in?(BsRequest::OBSOLETE_STATES)
     link_content = [request[:package]]
-    link_content << reviewers_gravatar(request) if request[:missing_reviews].present?
+    link_content << reviewers_gravatar(request, users_hash, groups_hash) if request[:missing_reviews].present?
     content_tag(:span, class: "badge state-#{css}") do
       link_to(request_show_path(request[:number]), class: 'request') do
         safe_join(link_content)
@@ -71,14 +71,14 @@ module Webui::Staging::WorkflowHelper
     end
   end
 
-  def requests(staging_project)
+  def requests(staging_project, users_hash, groups_hash)
     number_of_requests = staging_project.classified_requests.size
 
     return 'None' if number_of_requests == 0
 
     requests_visible_by_default = 10
     requests_links = staging_project.classified_requests.map do |request|
-      create_request_links(request)
+      create_request_links(request, users_hash, groups_hash)
     end
 
     if number_of_requests <= requests_visible_by_default
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -48,6 +48,23 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
     @more_excluded_requests = @staging_workflow.excluded_requests.count - @excluded_requests.size
     @empty_projects = @staging_workflow.staging_projects.without_staged_requests
     @managers = @staging_workflow.managers_group
+
+    # as it is not expected that there are many groups (~30) we catch all of them. Otherwise use this instead:
+    # group_ids = Review.where(bs_request: BsRequest.where(staging_project: @staging_projects)).select('group_id').distinct
+    # Group.where(id: group_ids).each do |group|
+
+    # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
+    @groups_hash = {}
+    Group.find_each do |group|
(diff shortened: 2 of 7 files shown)


krauselukas:
In order to have a gravatar on a group, you first need to set an email address for the group. This can be done through the API. For example with Curl:
```
curl -u username:password -d "cmd=set_email&email=example@foo.de" -X POST 'http://localhost:3000/group/<group-title>'

```

src/api/app/helpers/webui/group_helper.rb
```diff
@@ -6,4 +6,14 @@ def group_management_label
       'Group Management'
     end
   end
+
+  def webui2_group_image_tag(group:, css_class: nil, alt: nil, title: nil)
+    alt ||= (group ? "#{group.title}'s avatar" : 'Avatar')
+    url = if group.try(:email) && ::Configuration.gravatar
+            "https://www.gravatar.com/avatar/#{Digest::MD5.hexdigest(group.email.downcase)}?s=500&d=wavatar"
```
coolo: size 500 might be a little tough if you display it as 20x20

src/api/app/helpers/webui/staging/workflow_helper.rb
```diff
@@ -43,6 +43,22 @@ def progress(staging_project)
     end
   end
 
+  def reviewers_gravatar(request)
+    missing_reviews = request[:missing_reviews]
+    image_tags = []
+    missing_reviews.each do |review|
+      case review[:review_type]
+      when 'by_group'
+        group = Group.find_by_title(review[:by])
```
coolo: This is super wasteful. You run a find_by per review - and mostly cover like 3-4 groups with that. You better preload those in the controller / model.

src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -417,5 +417,16 @@ def tab_link(label, path, active = false)
 
     link_to(label, path, class: html_class)
   end
+
+  def webui2_user_or_group_image_tag(user_or_group:, css_class: nil, alt: nil, title: nil, size: 500)
+    user_or_group.is_a?(Group) ? (alt ||= (user_or_group ? "#{user_or_group.title}'s avatar" : 'Avatar')) :
```
bgeuken: I would split this up to avoid duplication:
```ruby
name = user_or_group.is_a?(Group) ? user_or_group.title : user_or_group.name
alt ||= (user_or_group ? "#{name}'s avatar" : 'Avatar')
```

src/api/app/helpers/webui/staging/workflow_helper.rb
```diff
@@ -43,20 +43,44 @@ def progress(staging_project)
...
+    end
+    image_tags
+  end
+
+  def create_request_links(request)
+    css = 'ready'
+    css = 'review' if request[:missing_reviews].present?
+    css = 'obsolete' if request[:state].in?(BsRequest::OBSOLETE_STATES)
+    link_content = []
+    link_content << request[:package]
```
bgeuken: Please squash the two line into `link_content = [request[:package]]`

src/api/app/models/staging/staging_project.rb
```diff
@@ -66,13 +66,12 @@ def missing_reviews
             who = review.send(att)
             next unless who
 
-            @missing_reviews << { id: review.id, request: request.number, state: review.state.to_s, package: request.first_target_package, by: who }
+            @missing_reviews << { id: review.id, request: request.number, state: review.state.to_s, package: request.first_target_package, by: who, review_type: att.to_s }
```
Ana06: why `to_s`?

src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -417,5 +417,16 @@ def tab_link(label, path, active = false)
 
     link_to(label, path, class: html_class)
   end
+
+  def webui2_user_or_group_image_tag(user_or_group:, css_class: nil, alt: nil, title: nil, size: 500)
```
Ana06: why do you need a custom title? I would try to generalize instead... (similar to `alt`)

src/api/app/helpers/webui/staging/workflow_helper.rb
Ana06: you could also do it here but just with one/two queries

src/api/app/helpers/webui/staging/workflow_helper.rb
```diff
@@ -43,20 +43,43 @@ def progress(staging_project)
...
+    image_tags = []
+    missing_reviews.each do |review|
+      case review[:review_type]
+      when 'by_group'
+        group = Group.find_by_title(review[:by])
+        image_tags << webui2_user_or_group_image_tag(user_or_group: group, title: review[:by], size: 20)
+      when 'by_user'
+        user = User.find_by_login(review[:by])
+        image_tags << webui2_user_or_group_image_tag(user_or_group: user, title: review[:by], size: 20)
+      end
```
coolo: there is no else - as such you would miss by_project and by_package

bgeuken: We could define a default by_project and a default by_package icon that we show in such cases. But this can be done in a follow up PR.

krauselukas: @coolo 

> there is no else - as such you would miss by_project and by_package

@Ana06 and I will do this in an follow up PR

coolo:
As I said after your planning meeting: *requiring* an email for groups just for display reasons is not acceptable. I don't mind if using the gravatar in case the group has an email address, but there needs to be better (configurable per group involved in staging) fallback/overwrite than default_face.png 18 of the 32 groups you counted don't have an email - and none of those with email is involved in the factory staging process.

And then there is leap-reviewers which shares the email with @lnussel which makes showing his gravatar for the group rather puzzling

krauselukas:
@coolo 

> As I said after your planning meeting: _requiring_ an email for groups just for display reasons is not acceptable. I don't mind if using the gravatar in case the group has an email address, but there needs to be better (configurable per group involved in staging) fallback/overwrite than default_face.png 18 of the 32 groups you counted don't have an email - and none of those with email is involved in the factory staging process.
> 
> And then there is leap-reviewers which shares the email with @lnussel which makes showing his gravatar for the group rather puzzling

@hennevogel ping

coolo:
And OBS is using https://wordpress.org/plugins/wavatars/#description - if you wanted to show a random icon, you could just hash the group title.

PR #6762, opened 2019-01-14
Add WebUI/API endpoints, documentation and job to copy a staging project

dmarcoux:
copy
For the WebUI part, I have created a view to preview the copy of a staging project. Right now, it's quite basic and rough, but I prefer to gather feedback before going further (if we decide it's worth it). Anyway, my idea is to provide some information about what is to be copied instead of just copying a staging project without giving any feedback to the user (this would be so if we used a modal for example).

Let me know what you think. Here's how it looks:

Clickable icons to reach the `Copy` page, they are displayed when we are configuring the staging workflow (so after clicking the `Configure` on the staging workflow show page):
icon

The `Copy` page:
copy

### DIFF
--- a/src/api/app/jobs/staging_project_copy_job.rb
+++ b/src/api/app/jobs/staging_project_copy_job.rb
@@ -1,0 +1,7 @@
+class StagingProjectCopyJob < ApplicationJob
+  def perform(staging_workflow_project_name, original_staging_project_name, staging_project_copy_name)
+    staging_workflow_project = Project.find_by!(name: staging_workflow_project_name)
+    original_staging_project = staging_workflow_project.staging.staging_projects.find_by!(name: original_staging_project_name)
+    original_staging_project.copy(staging_project_copy_name)
+  end
+end
--- a/src/api/app/views/webui2/webui/staging/projects/preview_copy.html.haml
+++ b/src/api/app/views/webui2/webui/staging/projects/preview_copy.html.haml
@@ -1,0 +1,21 @@
+- @pagetitle = "Copy #{@staging_project.name}"
+
+.row
+  .col-xl-10
+    .card.mb-3
+      = render(partial: 'webui2/webui/project/tabs', locals: { project: @staging_project })
+      .card-body
+        %h3= @pagetitle
+        %p The following data will be copied:
+        %ul
+          %li Project Configuration
+          %li Repositories
+          %li Users and Groups
+
+        = form_tag(staging_workflow_staging_project_copy_path(@staging_workflow, @staging_project), method: :post) do
+          .form-group
+            = label_tag(:staging_project_copy_name, 'Name of the Staging Project Copy:')
+            = text_field_tag(:staging_project_copy_name, nil, placeholder: "Eg: #{@staging_project}-copy", class: 'form-control')
+
+          = submit_tag('Copy', class: 'btn btn-primary')
+          = link_to('Back', edit_staging_workflow_path(@staging_workflow), class: 'btn btn-secondary', role: 'button')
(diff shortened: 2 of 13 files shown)


dmarcoux:
["create staging project" from a template](https://trello.com/c/diFikqV6/1018-create-staging-project-from-a-template)

src/api/app/jobs/staging_project_copy_job.rb
```diff
@@ -0,0 +1,7 @@
+class StagingProjectCopyJob < ApplicationJob
+  def perform(staging_workflow_project_name, original_staging_project_name, staging_project_copy_name)
```
bgeuken: Nitpick: I would think about shortening the variable names.

bgeuken:
> Let me know what you think.

I agree, but well... I was part of this discussion:wink:

src/api/app/jobs/staging_project_copy_job.rb
dmarcoux: I agree that they are quite long, but I didn't find something shorter without losing clarity.

dmarcoux:
Specs are flickering hard...

Ana06:
@dmarcoux 

> Specs are flickering hard...

It seems related with Rantly and not with your code. I am on it.

dmarcoux:
Thank you @Ana06 for having a look.

src/api/app/jobs/staging_project_copy_job.rb
```diff
@@ -0,0 +1,7 @@
...
+    staging_workflow_project = Project.find_by!(name: staging_workflow_project_name)
+    original_staging_project = staging_workflow_project.staging.staging_projects.find_by!(name: original_staging_project_name)
+    original_staging_project.copy(staging_project_copy_name)
```
dmarcoux: FYI: If this fails, the user has no way to know what happened.

Is there something we can do about this?

bgeuken: I'll have a look into that

src/api/app/jobs/staging_project_copy_job.rb
bgeuken: How about this?:

staging_workflow_project_name -> workflow_name
original_staging_project_name -> original_name
staging_project_copy_name -> copy_name

src/api/app/views/webui2/webui/staging/projects/preview_copy.html.haml
```diff
@@ -0,0 +1,23 @@
...
+
+.row
+  .col-xl-10
+    .card.mb-3
+      = render(partial: 'webui2/webui/project/tabs', locals: { project: @staging_project })
+      .card-body
+        %h3= @pagetitle
+        %p The following data will be copied:
+        %ul
+          %li Attributes
```
bgeuken: Attributes are not copied

src/api/app/views/webui2/webui/staging/projects/preview_copy.html.haml
```diff
@@ -0,0 +1,23 @@
...
+          %li Groups
```
bgeuken: Let's combine this with user config: `Users & Groups`

src/api/app/jobs/staging_project_copy_job.rb
dmarcoux: I just pushed a commit for a possible solution. It's not working yet as explained in the commit, but we can discuss about this.

src/api/app/jobs/staging_project_copy_job.rb
dmarcoux: Hum... I prefer longer names, they are clearer. If someone else thinks we should go for the names you suggest, I'm fine with changing them.

src/api/app/jobs/staging_project_copy_job.rb
dmarcoux: I removed the commit from the PR. I will create another PR when this one is merged.

PR #6767, opened 2019-01-15
Refactor package controller

vpereira:
I just broke my last PR #6737, refactoring ```PackageController``` in smallers chunks of change.

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -33,7 +33,7 @@ class Webui::PackageController < Webui::WebuiController
 
   before_action :require_repository, only: [:binary, :binary_download]
   before_action :require_architecture, only: [:binary, :binary_download]
-
+  before_action :check_ajax, only: [:update_build_log, :devel_project, :buildresult, :rpmlint_result]
   # make sure it's after the require_, it requires both
   before_action :require_login, except: [:show, :linking_packages, :linking_packages, :dependency,
                                          :binary, :binaries, :users, :requests, :statistics, :commit,
@@ -44,6 +44,8 @@ class Webui::PackageController < Webui::WebuiController
   before_action :check_build_log_access, only: [:live_build_log, :update_build_log]
 
   before_action :check_package_name_for_new, only: [:save_new]
+
+  before_action :handle_parameters_for_rpmlint_log, only: [:rpmlint_log]
 
   prepend_before_action :lockout_spiders, only: [:revisions, :dependency, :rdiff, :binary, :binaries, :requests, :binary_download]
 
@@ -164,7 +166,7 @@ class Webui::PackageController < Webui::WebuiController
     @filename = File.basename(params[:filename])
 
     begin
-      @fileinfo = Backend::Api::BuildResults::Binaries.fileinfo_ext(@project, params[:package], @repository.name, @arch.name, @filename)
+      @fileinfo = Backend::Api::BuildResults::Binaries.fileinfo_ext(@project, @package_name, @repository.name, @arch.name, @filename)
     rescue Backend::Error => e
       flash[:error] = "File #{@filename} can not be downloaded from #{@project}: #{e.summary}"
       redirect_to controller: :package, action: :binaries, project: @project,
@@ -178,11 +180,14 @@ class Webui::PackageController < Webui::WebuiController
       return
     end
 
-    @durl = download_url_for_file_in_repo(@project, @package_name, @repository, @arch.name, @filename)
+    url_generator = ::PackageControllerService::URLGenerator.new(project: @project, package: @package_name,
+                                                                 user: User.current, arch: @arch,
+                                                                 repository: @repository, filename: @filename)
+    @download_url = url_generator.download_url_for_file_in_repo
 
     logger.debug "accepting #{request.accepts.join(',')} format:#{request.format}"
     # little trick to give users eager to download binaries a single click
-    redirect_to(@durl) && return if request.format != Mime[:html] && @durl
+    redirect_to(@download_url) && return if request.format != Mime[:html] && @download_url
 
     switch_to_webui2
   end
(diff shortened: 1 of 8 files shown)


vpereira:
please take a look again @hennevogel

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -980,8 +975,9 @@ def rpmlint_result
         end
       end
     end
-    repos.uniq.each do |repo_name|
-      @repo_list << [repo_name, valid_xml_id(elide(repo_name, 30))]
+
+    @repo_list = repos.uniq.collect do |repo_name|
```
Ana06: [nitpick]  using `collect!` save some memory :wink: and now that you are refactoring this you could go a bit further and get rid of repos, call it `@repo_list` instead and directly do `@repo_list.uniq!.collect!` which saves even more memory. :wink: :wink:

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -178,7 +180,10 @@ def binary
       return
     end
 
-    @durl = download_url_for_file_in_repo(@project, @package_name, @repository, @arch.name, @filename)
+    url_generator = ::PackageControllerService::URLGenerator.new(project: @project, package: @package,
```
Ana06: You need to use @package_name here, as it can be different from `@package` and it will point to the wrong url!!!!!

src/api/app/services/package_controller_service/url_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+      @package = params[:package]
+    end
+
+    def logger
+      Rails.logger
+    end
+
+    def rpm_url
+      get_frontend_url_for(controller: 'build') +
+        "/#{@project.name}/#{@repository.name}/#{@arch}/#{@package.name}/#{@filename}"
```
Ana06: [nitpick] `.name` is not needed

src/api/app/services/package_controller_service/url_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def file_available?(url, max_redirects = 5)
+      logger.debug "Checking url: #{url}"
+      uri = URI.parse(url)
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.open_timeout = 15
+      http.read_timeout = 15
+      response = http.head uri.path
+      if response.code.to_i == 302 && response['location'] && max_redirects > 0
```
Ana06: [nitpick] Rubocop don't like that `> 0`... :cry:  you should use `.positive?` instead.

src/api/app/services/package_controller_service/url_generator.rb
vpereira: Hi @Ana06 , I will think about it, I personally find confusing that we implement ```to_s``` in some models but not all, then it's not clear in the method using it, if it receives an object or a string.

src/api/app/services/package_controller_service/url_generator.rb
vpereira: not my code, i just moved it around.. funny that the ``` rake dev:lint:rubocop``` didn't complain at all..

src/api/app/controllers/webui/package_controller.rb
vpereira: ah actually, I always pursue immutability and Nil avoidance, and therefore I prefer it like that than this "micro optimization" (for memory, yes, speed, no)

src/api/app/controllers/webui/package_controller.rb
vpereira: @DavidKang can you please review it again?

src/api/app/services/package_controller_service/url_generator.rb
vpereira: done

DavidKang:
@vpereira, looks good :clap:, but could you squash the commits and change the first commit message to something like: 'Refactor PackageController'

vpereira:
> @vpereira, looks good , but could you squash the commits and change the first commit message to something like: 'Refactor PackageController'

yes sir! Doing it right now.. I'm just looking in the security issue being reported on the binary view

vpereira:
@hennevogel, 
regarding the Haikiri: Rails sanitize the params from ```link_to```

```
irb(main):002:0> helper.link_to "<script></script>", "/"
=> "<a href=\"/\">&lt;script&gt;&lt;/script&gt;</a>"
irb(main):003:0> 
```

so could you please mark it as false positive?

src/api/app/controllers/webui/package_controller.rb
Ana06: well memory allocation needs time... but it is a nitpick

src/api/app/controllers/webui/package_controller.rb
Ana06: exactly, this was introduced to handle multibuilds. Because of that it can not be removed :wink:

src/api/app/services/package_controller_service/url_generator.rb
Ana06: yes, also your code. There are many offenses and that is why Rubocop doesn't complain in this PR (because it is disabled in the TODO file). But this introduces another offense that can't be automatically corrected. This is something that should be avoided :smile: (but it is marked as nitpick anyway)

ChrisBr:
Why did we merge this when Hakiri failed? This broke master :cry:

coolo:
And your analysis of the harakiri warning is wrong. This is not about the link text, but about the URL.

vpereira:
@coolo no it's not,  it was just reduced to simplicity, the only parameter that we have there is params[:filename], which is being reduced to the basename as everywhere:  https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/package_controller.rb#L166 ... are you seeing something that i missed?

vpereira:
BTW @coolo we have the HTTP header [x-xss-protection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection) in place.

PR #6769, opened 2019-01-15
[DO NOT MERGE] Test install gems from other sources than rubygems.org

ChrisBr:
image

This is a test to make bundle_gems source service working with gems installed from other sources (like github) than rubygems.org.

DO NOT MERGE!

### DIFF
--- a/dist/obs-bundled-gems.spec
+++ b/dist/obs-bundled-gems.spec
@@ -29,6 +29,7 @@ Source0:        Gemfile
 Source1:        Gemfile.lock
 BuildRequires:  autoconf
 BuildRequires:  automake
+BuildRequires:  git-core
 BuildRequires:  cyrus-sasl-devel
 BuildRequires:  gcc
 BuildRequires:  gcc-c++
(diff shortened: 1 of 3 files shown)


ChrisBr:
Ok I debugged a little bit more with @M0ses today. The service works as intended and has internet connection. The problem is something different :cry: 

When using ``git: 'https://foo.bar'`` bundler does store the repository in ``ruby/2.5.0/bundler/gems/`` but not as a gem but the whole repository. The bundle gems service does only create the ``cpio`` archive from ``vendor/cache/``. And it seems like a ``bundle --local`` does always checkout from git :thinking: That's just a brief analysis but seems like this is not just some easy fix :cry: 

See https://github.com/openSUSE/obs-service-bundle_gems/blob/master/bundle_gems#L101

ChrisBr:
Updated the bundle_gems service. However, most likely we need to update our spec file as well as we currently only copy over ``*.gem`` files.

https://github.com/openSUSE/obs-service-bundle_gems/pull/11

PR #6770, opened 2019-01-15
Handle self-referencing repositories when cloning staging projects

bgeuken:
Follow up of #6759 (The first two commits).

This adds some code to handle staging projects that contain repositories that link to repositories of the project they belong to. In such cases we now clone the self-referencing link.

### DIFF
--- a/src/api/spec/cassettes/Staging_StagingProject/_copy/copies_the_repositories_and_it_s_relations.yml
+++ b/src/api/spec/cassettes/Staging_StagingProject/_copy/copies_the_repositories_and_it_s_relations.yml
@@ -1,0 +1,38 @@
+---
+http_interactions:
+- request:
+    method: put
+    uri: http://backend:5352/source/home:tom:Staging:C/_config?user=factory%20bot
+    body:
+      encoding: UTF-8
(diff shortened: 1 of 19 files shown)


src/api/spec/cassettes/Staging_StagingProject/_copy/copies_the_repositories_and_it_s_relations.yml
```diff
@@ -0,0 +1,38 @@
...
+      - no-cache
+      Connection:
+      - close
+      Content-Length:
+      - '164'
+    body:
+      encoding: UTF-8
+      string: |
+        <status code="404">
+          <summary>project 'home:tom:Staging:C' does not exist</summary>
```
coolo: globalwritethough must die

coolo: there were enough other problems

bgeuken: because there was one issue which is addressed in this PR... alright:wink:

PR #6201, opened 2018-11-07
Create request exclusion model (for ignored requests)

eduardoj:
Create request ignoration model, controller and views to create and delete ignored requests.

2018-11-08-ignored_requests

Things to do (which will be done in following PRs - some of them are not only specific of this feature):

- API
- link to this page
- Improve factories (the managers can for example be created by default in the factory)
- Add validation and index to ensure that a request can not be excluded twice.
- Consider if project is needed as a parameter and if we want to render a staging workflow without project (independent of the decision there would be changes to make)
- Add staging workflow to breadcrumb

PR #6774, opened 2019-01-16
Make it possible to bulk create and destroy excluded requests via API

eduardoj:


eduardoj:
[Exclude Request should be added with XML (API).](https://trello.com/c/lpKZnL9q/1010-exclude-request-should-be-added-with-xml-api)

src/api/db/structure.sql
```diff
@@ -1140,6 +1140,7 @@ CREATE TABLE `staging_request_exclusions` (
   `description` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
   `created_at` datetime NOT NULL,
   `updated_at` datetime NOT NULL,
+  `number` int(11) DEFAULT NULL,
```
ChrisBr: Why do we need to add the number now?

ChrisBr:
I'm note sure if I understand but why is this PR named 

> Make it possible to the API to create and destroy excluded requests

as this is just a refactoring, no? What is the feature you added?

eduardoj:
@ChrisBr:

> What is the feature you added?

Before excluded requests could only be created and destroyed in the API once at a time.

With this pull request xml like this will be accepted with the API for creating:

```
<excluded_requests>
  <request number='14' description='Hey!'/>
  <request number='15' description='What?'/>
</excluded_requests>
```
and like this for deleting excluded requests:
```
<requests>
  <number>14</number>
  <number>15</number>
</requests>
```

ChrisBr:
@eduardoj thanks for the explanation, I changed the title to 

> Make it possible to bulk create and destroy excluded requests via API

to reflect that.

src/api/app/controllers/staging/excluded_requests_controller.rb
```diff
@@ -1,54 +1,86 @@
...
     end
   end
 
   private
 
-  def set_staging_workflow
-    project = Project.get_by_name(params[:staging_main_project_name])
-    @staging_workflow = project.staging
-    return if @staging_workflow
+  def set_xml_hash
```
hennevogel: Bad naming, yes it's an `XmlHash` but what do you expect inside?

src/api/app/controllers/staging/excluded_requests_controller.rb
```diff
@@ -1,54 +1,86 @@
...
 
-    if request_exclusion.save
-      render_ok
-    else
+    if errors?
       render_error(
         status: 400,
         errorcode: 'invalid_request',
-        message: request_exclusion.errors.full_messages.to_sentence
+        message: "Excluding requests to #{@staging_workflow} failed: #{errors_list.join('. ')}."
```
hennevogel: :thought_balloon:  s/to/for/

src/api/app/controllers/staging/excluded_requests_controller.rb
```diff
@@ -1,54 +1,86 @@
...
+    [@xml_hash[:number]].flatten
   end
 
-  def set_request_exclusion
-    request = BsRequest.find_by!(number: params[:number])
-    @request_exclusion = request.request_exclusion
-    return if @request_exclusion
+  def requests_to_be_excluded
+    xml_hash_requests.map do |request|
+      bs_request = @staging_workflow.unassigned_requests.find_by_number(request[:number])
```
hennevogel: What if this is `nil`?

src/api/db/structure.sql
eduardoj: @ChrisBr, we followed @coolo's [suggestion](https://github.com/openSUSE/open-build-service/pull/6201#discussion_r233365761). It was easier to add the 'number' column.

src/api/app/controllers/staging/excluded_requests_controller.rb
eduardoj: @hennevogel:

> Bad naming, yes it's an `XmlHash` but what do you expect inside?

I expect inside to have an API request in XmlHash format.

Fine with changing `set_xml_hash` to `set_xml_hash_request`, and `@xml_hash` to `@xml_hash_request`?

src/api/app/controllers/staging/excluded_requests_controller.rb
eduardoj: @hennevogel:

> What if this is `nil`?

This case is handled by validations. There is also a [test](https://github.com/openSUSE/open-build-service/pull/6774/files#diff-d6b6bb094c9652eb21cfdfa20e28ed77R58) for this case.

src/api/app/controllers/staging/excluded_requests_controller.rb
hennevogel: yes, you expect a list of requests. requests_xml_hash or whatnot sounds fine to me...

eduardoj: Changed to requests_xml_hash.

coolo:
I don't like this API at all (just as the XML to select and unselect). To construct this XML you need a lot of code - while in most cases you ignore single requests. So why insist on requiring XML as input?

ChrisBr:
> I don't like this API at all (just as the XML to select and unselect). To construct this XML you need a lot of code - while in most cases you ignore single requests. So why insist on requiring XML as input?

So you would prefer like it was before that you can only exclude one request per api call? Same for select/unselect?

coolo:
> > I don't like this API at all (just as the XML to select and unselect). To construct this XML you need a lot of code - while in most cases you ignore single requests. So why insist on requiring XML as input?
> 
> So you would prefer like it was before that you can only exclude one request per api call? Same for select/unselect?

Yes, especially as select and unselect are very lengthy processes (as you ditched the idea of making it background jobs) and as such I want to show progress on the command line.

ChrisBr:
> Yes, especially as select and unselect are very lengthy processes (as you ditched the idea of making it background jobs) and as such I want to show progress on the command line.

Yo, background job is ditched FOR NOW but we still plan to do that in the (near) future. As this route is more flexible I would keep it, crafting the XML is anyway done by a bot / tool

PR #6794, opened 2019-01-18
 Add group icons to request show page 

krauselukas:
To be consistent with the staging workflow page

image

Co-authored by @Ana06

### DIFF
--- a/src/api/app/helpers/webui/group_helper.rb
+++ b/src/api/app/helpers/webui/group_helper.rb
@@ -6,4 +6,9 @@ module Webui::GroupHelper
       'Group Management'
     end
   end
+
+  def group_with_icon(group_title)
+    group = Group.find_by(title: group_title)
+    user_image_tag(group, alt: group_title) + ' ' + link_to(group_title, group_show_path(group))
+  end
 end
--- a/src/api/app/helpers/webui/user_helper.rb
+++ b/src/api/app/helpers/webui/user_helper.rb
@@ -1,4 +1,5 @@
 module Webui::UserHelper
+  # This method is migrated to Webui2 (and refactored) with the name: image_tag_for
   # @param [User] user object
   def user_image_tag(user, opt = {})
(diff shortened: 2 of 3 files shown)


src/api/app/helpers/webui/group_helper.rb
```diff
@@ -6,4 +6,9 @@ def group_management_label
       'Group Management'
     end
   end
+
+  def group_with_icon(group_title)
+    group = Group.find_by(title: group_title)
+    user_image_tag(group, alt: group_title) + ' ' + link_to(group_title, group_show_path(group))
```
bgeuken: Nitpick: IIRC you could write this as
```ruby
link_to() do
  user_image_tag
```

bgeuken: and I would consider using [image_tag_for](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/helpers/webui/webui_helper.rb#L421) since we use it for the other icons as well

Ana06: This is the old UI not the new one. :wink: and the link is only in the name

src/api/app/helpers/webui/user_helper.rb
```diff
@@ -1,10 +1,11 @@
 module Webui::UserHelper
+  # This method is migrated to Webui2 (and refactored) with the name: image_tag_for
   # @param [User] user object
   def user_image_tag(user, opt = {})
```
dmarcoux: If we start using this for groups too, we should use `user_or_group` instead of `user`.

Ana06: this is the old UI, we are going to drop it. We just added here so that the staging workflow is consistent and that we don't forget when migrating (the comment is there to remember you to use the correct method when migrating)

dmarcoux: Oh alright.

PR #6796, opened 2019-01-19
Drop airbrake/errbit in favour of sentry

hennevogel:
We are going to switch from errbit to sentry on our reference server because
sentry is a tad more sophisticated. This also means you can use this now with a
hosted version on https://sentry.io.
    
We do not keep errbit around in the bundle because of the size. Goodbye errbit,
you have been a true friend, good companion and helped us to track countless
errors! 💚


dmarcoux:
Is there something more which needs to be set on https://sentry.io/ before merging this?

hennevogel:
Have to talk with IBS people

src/api/app/controllers/application_controller.rb
```diff
@@ -27,6 +27,8 @@ class ApplicationController < ActionController::Base
   before_action :shutup_rails
   before_action :validate_params
   before_action :require_login
+  before_action :set_raven_context
```
ChrisBr: @hennevogel shouldn't we do the same in the webui controller?

vpereira:
@hennevogel any update on that?

hennevogel:
> @hennevogel any update on that?

No, @bugfinder wanted to deploy it differently. No progress as far as I can tell...

mlschroe:
How about you refactor the notification calls so that the code isn't sprinkled with Raven/Airbrake calls? This makes it easier when some new notification system pops up in the future. (Or even allow multiple systems to be used.)

hennevogel:
this is the second notification system in 13 years and even this change is now hanging around since 5 months. I'm not too worried about frequent change 😉

mlschroe:
But don't you admit that having the module name sprinkled around in the code is bad style? Aren't you a fan of object orientation?

coolo:
This has little to do with object orientation - this is changing a library. If the backend switched from XML::Structure to XML::Simple you would have to change one or two places as well.

mlschroe:
(Bad example, the XML::Structured calls are wrapped by BSUtil, so it's not hard to replace them. That's basically also what I am proposing: create a class that abstracts the error notification implementation.)

coolo:
I'm not a fan of such abstractions - they tend to add bugs themselves and make it harder to follow upstream receipes for specific problems. And then they only work for very comparable libraries - if the 2025 solution requires to add a callback into the Quantum processor, we will have to patch the abstraction, but it will no longer work with the 2019 one.

vpereira:
does this PR still make sense @hennevogel ? If you don't want to loose track of it, we can open an issue and link this closed PR as reference. What do you think?

hennevogel:
I still want us to switch. Let's leave this hanging around please. It doesn't cost anything :cowboy_hat_face:

hennevogel:
Sentry docker images by now are way more complicate. This would need more work...

PR #6802, opened 2019-01-21
rspec: Don't fake the request state

coolo:
I went too far with the bs_requests factory, reverting some of it


coolo:
@saraycp ?

saraycp:
@coolo  so... `sanitize`is not skipped in the factory anymore... :thinking: looks more reasonable! :heart:

PR #6803, opened 2019-01-21
[WIP] Add report model

bgeuken:
This is kind of a proof of concept. At least the payload part is similar to what we have for events, except that events don't have relations right now, eg. belongs to staging_project.
I think it's better to not extend the events in that way. Instead I created a report model to store results of jobs we run asynchronously.

We were talking about giving users an overview of what has been copied. This could be done via the payload in the [CopyStagingReport model](https://github.com/openSUSE/open-build-service/pull/6803/files#diff-1689407cb24a3477a992ee3a3d5e1ab0R12).
IIRC we had a similar issue with staging bs_requests. This could be solved the same way.

### DIFF
--- a/src/api/app/jobs/staging_project_copy_job.rb
+++ b/src/api/app/jobs/staging_project_copy_job.rb
@@ -1,7 +1,13 @@
 class StagingProjectCopyJob < ApplicationJob
+  queue_as :default
+
+  rescue_from(::ActiveRecord::ActiveRecordError) do |exception|
+    @original_staging_project.reports.create!(failure_message: exception.message)
+  end
+
   def perform(staging_workflow_project_name, original_staging_project_name, staging_project_copy_name)
     staging_workflow_project = Project.find_by!(name: staging_workflow_project_name)
-    original_staging_project = staging_workflow_project.staging.staging_projects.find_by!(name: original_staging_project_name)
-    original_staging_project.copy(staging_project_copy_name)
+    @original_staging_project = staging_workflow_project.staging.staging_projects.find_by!(name: original_staging_project_name)
+    @original_staging_project.copy(staging_project_copy_name)
   end
 end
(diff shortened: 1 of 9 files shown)


src/api/app/jobs/staging_project_copy_job.rb
```diff
@@ -2,6 +2,14 @@ class StagingProjectCopyJob < ApplicationJob
   def perform(staging_workflow_project_name, original_staging_project_name, staging_project_copy_name)
     staging_workflow_project = Project.find_by!(name: staging_workflow_project_name)
     original_staging_project = staging_workflow_project.staging.staging_projects.find_by!(name: original_staging_project_name)
-    original_staging_project.copy(staging_project_copy_name)
+
+    begin
```
dmarcoux: There is a built-in way to rescue from exception in ActiveJob. It is `rescue_from` (really original, I know!!!).

https://edgeguides.rubyonrails.org/active_job_basics.html#exceptions

bgeuken: Oh nice! I'll change it

hennevogel:
Why do you put this into the database? Isn't this just an email notification?

dmarcoux:
@hennevogel We aren't use on how we should let the user know this is done. Is an email notification enough? Should it be in OBS directly?

hennevogel:
> We aren't use on how we should let the user know this is done. Is an email notification enough? Should it be in OBS directly?

We currently have no notifications except email notifications. I guess that answers your scope.

> The problem I see with the email notification is that the user has no way to know if it will ever happen. If there was some kind of way to track jobs in the UI, at least the user could check there.

At some later stage you could make this a push/poll step in the UI workflow.  Like

1. Push job into queue
1. Show temporary page
1. Poll for status until it's done
1. Redirect to final output

Like forking a repo on github

screenshot from 2019-01-21 18-08-01

dmarcoux:
True, forking a repo on GitHub is a good example.

bgeuken:
We finished the staging workflow deliverable and work on the bootstrap migration for the next two sprints... Therefore closing

PR #6804, opened 2019-01-21
Several performance improvements for the Factory index page.

ChrisBr:
* Added caches for build_id and download_url
* Do not show checks for repositories without required checks

selection_027

https://obs-measure.opensuse.org/d/fzuOfHQik/request?orgId=1&var-request_id=08cb3ec0-8451-42aa-a019-f1fa9157a0ba&from=1547966082967&to=1547966137268&var-method=Webui::ObsFactory::StagingProjectsController%23index


src/api/app/models/repository_architecture.rb
```diff
@@ -10,6 +10,8 @@ class RepositoryArchitecture < ApplicationRecord
   validates :repository, uniqueness: { scope: :architecture }
 
   def build_id
+    return if required_checks.empty?
```
coolo: I find this problematic. There still is a build_id, so I would avoid calling the function, but not have the function fake there is no build_id

src/api/lib/backend/api/published.rb
```diff
@@ -7,15 +7,19 @@ class Published
...
 
       # Returns the build id for a repository
       # @return [String]
       def self.build_id(project_name, repository_name)
-        response = download_url_for_repository(project_name, repository_name, :status)
-        result = Xmlhash.parse(response).with_indifferent_access
-        result[:buildid]
+        Rails.cache.fetch("build_id-#{project_name}-#{repository_name}") do
+          response = download_url_for_repository(project_name, repository_name, :status)
+          result = Xmlhash.parse(response).with_indifferent_access
```
coolo: any idea what's the point of casting the hash to a with_indifferent_access to pick one element?
`Xmlhash.parse(response)['buildid']`

src/api/app/models/repository_architecture.rb
coolo: What I'd suggest is (untested):

```
index 4c43029e7..04f8efe80 100644
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1675,6 +1675,7 @@ class Project < ApplicationRecord
   end
 
   def status_reports(checkables)
+    checkables = checkables.reject { |checkable| checkable.required_checks.empty? }
     status_reports = Status::Report.where(checkable: checkables)
     result = {}
     status_reports.where(uuid: checkables.map(&:build_id)).find_each do |report|

Comment (ChrisBr) [reply src/api/app/models/repository_architecture.rb]: Makes sense.

File: src/api/app/models/project.rb [hunk 3]
```diff
@@ -1675,6 +1675,8 @@ def has_local_distribution(project_name, repository)
   end
 
   def status_reports(checkables)
+    checkables = checkables.reject { |checkable| checkable.required_checks.empty? }
```
Ana06: if we are speaking about performance, maybe this should be do in the SQL level :wink:

src/api/lib/backend/api/published.rb
```diff
@@ -17,8 +17,7 @@ def self.download_url_for_repository(project_name, repository_name, view = :publ
       def self.build_id(project_name, repository_name)
         Rails.cache.fetch("build_id-#{project_name}-#{repository_name}") do
           response = download_url_for_repository(project_name, repository_name, :status)
-          result = Xmlhash.parse(response).with_indifferent_access
-          result[:buildid]
+          (Xmlhash.parse(response) || {})['buildid']
```
Ana06: why do we need now `{}`? :confused:

src/api/app/models/project.rb
ChrisBr: The perf problem is the backend call.

src/api/lib/backend/api/published.rb
ChrisBr: Xmlhash can return nil which would crash then.

src/api/app/models/project.rb
coolo: and in SQL the required_checks does not even exist - as it's some JSON dump :(

Not sure how active record would handle that. But storing the required checks somewhere else is on the midterm TODO

src/api/lib/backend/api/published.rb
coolo: e.g. on 404s

src/api/app/models/project.rb
ChrisBr: Actually I tried it and it works

```
Repository.where.not(required_checks: nil)
```

so changed...

src/api/lib/backend/api/published.rb
Ana06: then this was already failing.... :confused:

src/api/app/models/project.rb
```diff
@@ -1689,7 +1689,8 @@ def status_reports(checkables)
   end
 
   def combined_status_reports
-    @combined_status_reports ||= status_reports(repositories) | status_reports(repository_architectures)
+    @combined_status_reports ||= status_reports(repositories.where.not(required_checks: nil)) |
```
Ana06: [nitpick] I would move `where.not(required_checks: nil)` to status_reports:

```
def status_reports(checkables)
    checkables = checkables.where.not(required_checks: nil)
```

so that it keep the same logic as before and don't duplicate code

coolo: Or just drop it - if the caches work, it doesn't matter too much and avoids the problem with bootstraping stagings' reports.

src/api/lib/backend/api/published.rb
coolo: aehm, just checked - 404s throw an exception. So parse basically can't return nil - unless the backend is broken, which I don't think we want to avoid optimize for

ChrisBr: Dropped

src/api/app/models/project.rb
ChrisBr: I already fixed the specs now :smile:

src/api/app/models/event/repo_build_started.rb
```diff
@@ -3,6 +3,19 @@ class RepoBuildStarted < Base
     self.message_bus_routing_key = 'repo.build_started'
     self.description = 'Repository (re)started building'
     payload_keys :project, :repo, :arch, :buildid
+    after_save :clear_cache
```
coolo: hmm, we also need this from repo_build_finished - as the buildid updates then.

PR #6806, opened 2019-01-21
Fix missing groups and users hash on staging

saraycp:
Issue introduced when improving the performance of the workflow
show page, which uses the same partials as the broken pages:
#6760

Co-authored-by: Ana María Martínez Gómez <ammartinez@suse.de>

### DIFF
--- a/src/api/app/controllers/webui/staging/projects_controller.rb
+++ b/src/api/app/controllers/webui/staging/projects_controller.rb
@@ -25,6 +25,9 @@ module Webui
       def show
         @staging_project = @staging_workflow.staging_projects.find_by(name: params[:project_name])
         @project = @staging_workflow.project
+
+        @groups_hash = ::Staging::Workflow.load_groups
+        @users_hash = ::Staging::Workflow.load_users(@staging_project)
       end
 
       def destroy
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -49,22 +49,8 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
     @empty_projects = @staging_workflow.staging_projects.without_staged_requests
     @managers = @staging_workflow.managers_group
 
-    # as it is not expected that there are many groups (~30) we catch all of them. Otherwise use this instead:
-    # group_ids = Review.where(bs_request: BsRequest.where(staging_project: @staging_projects)).select('group_id').distinct
-    # Group.where(id: group_ids).each do |group|
-
-    # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
-    @groups_hash = {}
-    Group.find_each do |group|
-      @groups_hash[group.title] = group
-    end
-
-    # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
-    @users_hash = {}
-    user_ids = Review.where(bs_request: BsRequest.where(staging_project: @staging_projects)).select('user_id').distinct
-    User.where(id: user_ids).each do |user|
-      @users_hash[user.login] = user
-    end
+    @groups_hash = ::Staging::Workflow.load_groups
+    @users_hash = ::Staging::Workflow.load_users(@staging_projects)
   end
 
   def edit
@@ -72,6 +58,9 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
 
     @project = @staging_workflow.project
     @staging_projects = @staging_workflow.staging_projects.includes(:staged_requests)
+
+    @groups_hash = ::Staging::Workflow.load_groups
+    @users_hash = ::Staging::Workflow.load_users(@staging_projects)
   end
 
   def destroy
--- a/src/api/app/models/staging/workflow.rb
+++ b/src/api/app/models/staging/workflow.rb
@@ -50,6 +50,31 @@ class Staging::Workflow < ApplicationRecord
     Backend::Api::Sources::Project.write_staging_workflow(project.name, User.current_login, render_xml)
   end
 
+  def self.load_groups
+    # as it is not expected that there are many groups (~30) we cache all of them. Otherwise use this instead:
+    # group_ids = Review.where(bs_request: BsRequest.where(staging_project: @staging_projects)).select('group_id').distinct
+    # Group.where(id: group_ids).each do |group|
+    #
+    # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
+    #
+    groups_hash = {}
+    # FIXME: try to use cache without failure. Like: Rails.cache.fetch(Group.all.cache_key) do
+    Group.find_each do |group|
+      groups_hash[group.title] = group
+    end
+    groups_hash
+  end
+
+  def self.load_users(staging_projects)
+    # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
+    users_hash = {}
+    user_ids = Review.where(bs_request: BsRequest.where(staging_project: staging_projects)).select('user_id').distinct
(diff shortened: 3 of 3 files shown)


src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -25,6 +25,10 @@ def create
       def show
         @staging_project = @staging_workflow.staging_projects.find_by(name: params[:project_name])
         @project = @staging_workflow.project
+        @groups_hash = {}
+        Group.find_each do |group|
```
ChrisBr: What about writing that into the cache in a function in group?

```rb
def group_cache
  Rails.cache.fetch(Group.all.cache_key) do
    result = {}
    Group.find_each do |group|
      result[group.title] = group
    end
    result
  end
end
```

Ana06: We should move this code to a place where it can be shared. It is now repeated. :sob:

dmarcoux:
The error also happens on the staging workflow edit page:

error

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -25,6 +25,10 @@ def create
       def show
         @staging_project = @staging_workflow.staging_projects.find_by(name: params[:project_name])
         @project = @staging_workflow.project
+        @groups_hash = {}
```
dmarcoux: `@users_hash` is also missing.

vpereira:
> The error also happens on the staging workflow edit page:

yes this one we catched already...

Ana06:
There is another option to fix this, much easier. The staging project edit and show pages do not look for the group and the user that much as the staging workflow page. We could look for the record in the view if not catched. :thinking:

src/api/app/controllers/webui/staging/concerns/groups_and_users_cache.rb
```diff
@@ -0,0 +1,32 @@
+module Webui
+  module Staging
+    module Concerns
+      module GroupsAndUsersCache
+        extend ActiveSupport::Concern
+
+        private
+
+        def cache_groups_and_users
```
ChrisBr: I think this method should be splitted into groups and users to only do one thing. Also it does not really do what it says because users is not really cached.

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -2,11 +2,13 @@ module Webui
   module Staging
     class ProjectsController < WebuiController
       layout 'webui2/webui'
+      include Webui::Staging::Concerns::GroupsAndUsersCache
 
       before_action :require_login
       before_action :set_staging_workflow
-      after_action :verify_authorized, except: :show
       before_action :set_webui2_views
+      before_action :cache_groups_and_users, only: :show
```
ChrisBr: You can do that in the concern as well.

```rb
  included do
    before_action :cache_groups_and_users, only: :show
  end
```

src/api/app/controllers/webui/staging/concerns/groups_and_users_cache.rb
```diff
@@ -0,0 +1,32 @@
...
+          Rails.cache.fetch(Group.all.cache_key) do
+            @groups_hash = {}
+            Group.find_each do |group|
+              @groups_hash[group.title] = group
+            end
+          end
+
+          # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
+          @users_hash = {}
+          user_ids = Review.where(bs_request: BsRequest.where(staging_project: @staging_project || @staging_projects)).select('user_id').distinct
```
ChrisBr: you could cache this as well, see my overall comment.

vpereira: we need to merge it today, maybe we can break it in two PRs @saraycp  @ChrisBr ?

ChrisBr: Fine for me...

src/api/app/controllers/webui/staging/concerns/groups_and_users_cache.rb
Ana06: splitting a method in two to always call both of them....

Ana06:
@saraycp and I are refactoring this again, give us 10 minutes :wink:

src/api/app/controllers/webui/staging/concerns/groups_and_users_cache.rb
Ana06: @vpereira 

> we need to merge it today, maybe we can break it in two PRs @saraycp @ChrisBr ?

how? :confused:

src/api/app/models/staging/workflow.rb
```diff
@@ -50,6 +50,31 @@ def write_to_backend
...
 
+  def self.load_groups
+    # as it is not expected that there are many groups (~30) we cache all of them. Otherwise use this instead:
+    # group_ids = Review.where(bs_request: BsRequest.where(staging_project: @staging_projects)).select('group_id').distinct
+    # Group.where(id: group_ids).each do |group|
+    #
+    # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
+    #
+    groups_hash = {}
+    # FIXME: try to use cache without failure. Like: Rails.cache.fetch(Group.all.cache_key) do
```
Ana06: What do you mean with "without failure"? What is exactly the problem? :confused:

saraycp: Using `Rails.cache.fetch(Group.all.cache_key)` always returns an empty hash :disappointed:

I simplified the code to make it work for the review. We can refactor afterwards.

saraycp: @Ana06 we can work on this thoroughly after the meeting if you want.

Ana06: Not sure what you tried. Can you please try the following code? I think it should work :wink:

```
Rails.cache.fetch(Group.all.cache_key) do
    groups_hash = {}
    Group.find_each do |group|
      groups_hash[group.title] = group
    end
    groups_hash
end

Comment (Ana06) [reply src/api/app/models/staging/workflow.rb]: If that doesn't work I would check if `cache_key` is correctly generated (in other words if it is always the same

Comment (saraycp) [reply src/api/app/models/staging/workflow.rb]: `each`is not the problem. I'll check `cache_key`

Comment (Ana06) [reply src/api/app/models/staging/workflow.rb]: @saraycp, do you have groups in your instance? That would explain that it is {}

Comment (saraycp) [reply src/api/app/models/staging/workflow.rb]: I do :smile:

File: src/api/app/controllers/webui/staging/workflows_controller.rb [hunk 7]
```diff
@@ -49,29 +49,18 @@ def show
...
-      @groups_hash[group.title] = group
-    end
-
-    # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
-    @users_hash = {}
-    user_ids = Review.where(bs_request: BsRequest.where(staging_project: @staging_projects)).select('user_id').distinct
-    User.where(id: user_ids).each do |user|
-      @users_hash[user.login] = user
-    end
+    @groups_hash = ::Staging::Workflow.load_groups
```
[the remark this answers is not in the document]
saraycp: yes, we can check again later. I was so sleepy to remember why :sweat_smile:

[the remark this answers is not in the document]
bgeuken: AFAIK the view rendering is part of the action, even if you don't explicitly call render :foo. So render_after is to late.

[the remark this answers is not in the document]
Ana06: yes, it would have to be a before filter (but this need some other things to be moved as well (which would also make sense)

PR #6809, opened 2019-01-21
Fix error in StagingProjectCopyJob

dmarcoux:
Setting the current user is needed in the job as it is performed asynchronously. Therefore, we now pass the current user id to the job to use it later.


dmarcoux:
@openSUSE/obs-frontend Please review.

PR #6815, opened 2019-01-22
Cache groups_hash

Ana06:
It didn't work in https://github.com/openSUSE/open-build-service/pull/6806 :relaxed:

### DIFF
--- a/src/api/app/models/staging/workflow.rb
+++ b/src/api/app/models/staging/workflow.rb
@@ -56,13 +56,13 @@ class Staging::Workflow < ApplicationRecord
     # Group.where(id: group_ids).each do |group|
     #
     # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
-    #
-    groups_hash = {}
-    # FIXME: try to use cache without failure. Like: Rails.cache.fetch(Group.all.cache_key) do
-    Group.find_each do |group|
-      groups_hash[group.title] = group
+    Rails.cache.fetch("groups_hash_#{Group.all.cache_key}") do
+      groups_hash = {}
+      Group.find_each do |group|
+        groups_hash[group.title] = group
+      end
+      groups_hash
     end
-    groups_hash
   end
 
   def self.load_users(staging_projects)


src/api/app/models/staging/workflow.rb
```diff
@@ -57,12 +57,13 @@ def self.load_groups
     #
     # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
     #
-    groups_hash = {}
-    # FIXME: try to use cache without failure. Like: Rails.cache.fetch(Group.all.cache_key) do
-    Group.find_each do |group|
-      groups_hash[group.title] = group
+    Rails.cache.fetch(Group.all.cache_key) do
```
coolo: you have to make sure no other code is using this cache_key to cache another return value, so better append some code path specific key

src/api/app/models/staging/workflow.rb
```diff
@@ -56,13 +56,13 @@ def self.load_groups
     # Group.where(id: group_ids).each do |group|
     #
     # TODO: Refactor this code using to_h when updating to Ruby 2.6 (performance improvement)
-    #
-    groups_hash = {}
-    # FIXME: try to use cache without failure. Like: Rails.cache.fetch(Group.all.cache_key) do
-    Group.find_each do |group|
-      groups_hash[group.title] = group
+    Rails.cache.fetch("groups_hash_#{Group.all.cache_key}") do
```
ChrisBr: Why is the prefix needed?

Ana06: [@coolo's comment](https://github.com/openSUSE/open-build-service/pull/6815#discussion_r249755960):

> you have to make sure no other code is using this cache_key to cache another return value, so better append some code path specific key

Ana06: I think it doesn't hurt

ChrisBr: I'm fine with coolo's explanation.

coolo: note that `['groups_hash', Group.all_cache_key] ` seems to be the stackoverflow preferred way to write this :)

PR #6822, opened 2019-01-23
OBS_Factory: Improve database queries

coolo:
Take selected_requests as number array and avoid duplicated queries for stagings without
no difference between open_requests and selected_requests

We still load the requests per project, but this is hard to fix without injecting into objects

```
Before: 8461ms (Views: 1742.5ms | ActiveRecord: 3279.2ms | Backend: 3.8ms)
After:  3054ms (Views: 821.8ms  | ActiveRecord: 689.0ms | Backend: 3.6ms)
```

(the backend is of course lobotomized in the dev environment)

### DIFF
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -89,7 +89,7 @@ class BsRequest < ApplicationRecord
   }
 
   before_save :assign_number
-  has_many :bs_request_actions, -> { includes([:bs_request_action_accept_info]) }, dependent: :destroy
+  has_many :bs_request_actions, dependent: :destroy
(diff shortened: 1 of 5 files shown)


coolo:
damn, that broke what is returned as selected_requests in JSON :(

Need to fix

ChrisBr:
> damn, that broke what is returned as selected_requests in JSON :(

To be fair: It improved the performance indeed :smile: 

https://obs-measure.opensuse.org/d/Gpp3B1Pik/action?orgId=1&var-Action=Webui::ObsFactory::StagingProjectsController%23index&var-beta=All

Issue #6778, opened 2019-01-17
[bootstrap] Project build status harder to parse visually


coolo:
Blocked and disabled are the same colour (on my monitor with my eyes), which makes it very hard to spot that there is a lot left to build

![screenshot_20190117_073557](https://user-images.githubusercontent.com/1067203/51300021-bdfe0280-1a2a-11e9-8e9c-947ea1b78af7.png)
![screenshot_20190117_073636](https://user-images.githubusercontent.com/1067203/51300022-c22a2000-1a2a-11e9-9666-76386a2af494.png)

PR #6828, opened 2019-01-23
Add background colour in buildresult for blocked state

DavidKang:
Fix issue #6778

screenshot_2019-01-23 show home admin perl-citrix - open build service 1


dmarcoux:
Hakiri seems to be stuck and since this is only changing CSS, I'll merge it.

PR #6829, opened 2019-01-23
Use an SQL query to check whether a package has a patchinfo or not

bgeuken:
This is much faster than iterating through an array and checking a
hashed backend result for each package.


src/api/app/controllers/webui/project_controller.rb
```diff
@@ -193,13 +193,7 @@ def new_release_request
...
     @bugowners_mail = @project.bugowner_emails
 
-    # An incident has a patchinfo if there is a package 'patchinfo' with file '_patchinfo', try to find that:
-    @has_patchinfo = false
-    if @packages.map { |p| p[0] }.include?('patchinfo')
-      Directory.hashed(project: @project.name, package: 'patchinfo').elements('entry') do |e|
-        @has_patchinfo = true if e['name'] == '_patchinfo'
-      end
-    end
+    @has_patchinfo = PackageKind.where(package_id: @project.package_ids, kind: 'patchinfo').exists?
```
vpereira: nice catch @bgeuken !

bgeuken: Probably, but we would still store this in a variable since it's used multiple times in the UI. So not sure we would gain much.

src/api/app/models/project.rb
```diff
@@ -179,6 +180,10 @@ def self.remote_image_template_from_xml(remote_project, image_template_project)
     project
   end
 
+  def patchinfos
```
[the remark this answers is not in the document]
vpereira: i disagree, project has patchinfos, that's how we see it. The only thing is that even if in the rails level we allow a project to have many patchinfos, we enforce it in different places that one project has one and only one patchinfo.

PR #6831, opened 2019-01-23
Filter out duplicated flags from user provided XML

### LINKED ISSUE #6823: New Flag validation is harmful
The new Flag uniqness validation leads to broken operations (eg. project copy) when the source had duplicates. Leaving inconstent data and blocked workflows.

Keep also in mind that we have plenty of frozen projects which must not get touched (neither in db nor in source), so a migration is no solution and leads to more inconsistent data.

Working with these project must be still possible.

I have disabled the check on production for now to unblock teams, so just P2 for now.

coolo:
This can happen either when posting a _meta (in this case an error could be good enough, but isn't really helpful) or when receiving _meta from local or remote backend. In this case we can't error out, but need to recover the situation

Fixes #6823

### DIFF
--- a/src/api/app/helpers/flag_helper.rb
+++ b/src/api/app/helpers/flag_helper.rb
@@ -40,6 +40,9 @@ module FlagHelper
     # translate the flag types as used in the xml to model name + s
     validate_type flagtype
 
+    # we need to catch duplicates - and prefer the last
+    flags_to_create = {}
+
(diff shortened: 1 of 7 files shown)


dmarcoux:
Could you please rebase to fix the conflicts? Otherwise, it looks good.

coolo:
done

PR #6834, opened 2019-01-24
Fix staging dashboard caching

coolo:
after_save hooks in derived classe aren't taken into account, so implement a hook in the base class that can overloaded

The users were happy about the performance, but noticed 3 days into the change that projects didn't move - so let's correctly invalidate caches


src/api/app/models/event/repo_published.rb
```diff
@@ -3,7 +3,6 @@ class RepoPublished < Base
     self.message_bus_routing_key = 'repo.published'
     self.description = 'Repository was published'
     payload_keys :project, :repo, :buildid
-    after_save :clear_cache
 
     private
```
DavidKang: @coolo, The next [private method](https://github.com/openSUSE/open-build-service/pull/6834/files#diff-80c5a72c12bac93b33994b23adcdd6d5R9) should be called `clear_caches` like the others?

coolo: yep, will fix

PR #6835, opened 2019-01-24
Refactor source diff code in package show view

bgeuken:
This commit moves business logic into helpers. We are about to migrate
this view to bootstrap. This change makes it easier to re-use the
existing code.

### DIFF
--- a/src/api/app/helpers/webui/request_helper.rb
+++ b/src/api/app/helpers/webui/request_helper.rb
@@ -129,4 +129,14 @@ module Webui::RequestHelper
     return "#{review[:by_project]} / #{review[:by_package]}" if review[:by_package]
     review[:by_user] || review[:by_group] || review[:by_project]
   end
+
+  def diff_data(action_type, sourcediff)
+    diff = (action_type == :delete ? sourcediff['old'] : sourcediff['new'])
+
+    { project: diff['project'], package: diff['package'], rev: diff['rev'] }
+  end
+
+  def diff_label(diff)
+    "#{diff['project']} / #{diff['package']} (rev #{diff['rev']})"
+  end
 end
--- a/src/api/app/views/webui/request/show.html.erb
+++ b/src/api/app/views/webui/request/show.html.erb
@@ -193,23 +193,14 @@
                   <% if sourcediff[:error] %>
                       <%= content_tag(:p, content_tag(:i, sourcediff[:error], class: :error)) %>
                   <% else %>
-                      <% if action[:type] == :delete
-                           t = sourcediff['old']
-                         else
-                           t = sourcediff['new']
-                         end
-                         project, package, rev = t['project'], t['package'], t['rev'] %>
-
+                      <% diff = diff_data(action[:type], sourcediff) %>
                       <% if action[:sourcediff].length != 1 %>
                           <h4>
-                            <%= sourcediff['new']['project'] %> / <%= sourcediff['new']['package'] %>
-                            (rev <%= sourcediff['new']['rev'] %>) -
-                            <%= sourcediff['old']['project'] %> / <%= sourcediff['old']['package'] %>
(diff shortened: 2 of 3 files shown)


src/api/app/helpers/webui/request_helper.rb
```diff
@@ -129,4 +129,13 @@ def reviewer(review)
     return "#{review[:by_project]} / #{review[:by_package]}" if review[:by_package]
     review[:by_user] || review[:by_group] || review[:by_project]
   end
+
+  def diff_data(action_type, sourcediff)
+    diff = (action_type == :delete ? sourcediff['old'] : sourcediff['new'])
+    diff.slice('project', 'package', 'rev').with_indifferent_access
```
Ana06: using `with_indifferent_access` when you only want to use it with symbols doesn't seems like a good idea...

src/api/app/views/webui/request/show.html.erb
```diff
@@ -193,23 +193,14 @@
...
-                            <%= sourcediff['new']['project'] %> / <%= sourcediff['new']['package'] %>
-                            (rev <%= sourcediff['new']['rev'] %>) -
-                            <%= sourcediff['old']['project'] %> / <%= sourcediff['old']['package'] %>
-                            (rev <%= sourcediff['old']['rev'] %>)
+                            <%= diff_label(sourcediff['new']) %> - <%= diff_label(sourcediff['old']) %>
                           </h4>
                       <% end %>
                       <%#TODO: Redo source/target link display %>
-                      <%= render(:partial => 'shared/sourcediff', locals: { filenames: sourcediff['filenames'], files: sourcediff['files'], :source => { project: project, package: package, :rev => rev }, :css_id => "diff_action_#{action_index}_#{action[:type]}_#{sourcediff_index}", :editor_width => '550px' }) %>
+                      <%= render(partial: 'shared/sourcediff', locals: { filenames: sourcediff['filenames'], files: sourcediff['files'], source: diff, css_id: "diff_action_#{action_index}_#{action[:type]}_#{sourcediff_index}", editor_width: '550px' }) %>
```
Ana06: Now source can be undefined, before it was always defined, but had `nil` values. This can crash in the partial :fire:

src/api/app/helpers/webui/request_helper.rb
bgeuken: Well, the sliced data is using strings as keys. And the render call expects symbols

src/api/app/views/webui/request/show.html.erb
bgeuken: There is always a project and a package when there is a sourcediff. So this should be fine.

src/api/app/helpers/webui/request_helper.rb
Ana06: all render calls are using symbols... so.. you should use symbols... otherwise Ruby need to convert your symbols to strings.... :sob:

src/api/app/views/webui/request/show.html.erb
Ana06: I don't understand what you mean... :confused:

src/api/app/helpers/webui/request_helper.rb
Ana06: `{ project: diff[:project], package: diff[:package], rev: diff[:rev] }`

src/api/app/views/webui/request/show.html.erb
Ana06: I read it wrong, sorry

src/api/app/helpers/webui/request_helper.rb
bgeuken: I doubt this has any relevant performance impact... but I don't want to have another bikeshed discussion. I changed it.

bgeuken:
@Ana06 updated. Please have another look:smiley:

PR #6838, opened 2019-01-24
Add skeleton for request show page in webui2

krauselukas:
The migration of the request view to bootstrap, requires
a basic skeleton

Co-authored-by: @bgeuken

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -125,6 +125,9 @@ class Webui::RequestController < Webui::WebuiController
 
     @comments = @bs_request.comments
     @comment = Comment.new
+
+    # TODO: Remove the statement after migration is finished
+    switch_to_webui2 if Rails.env.development?
   end
 
   def sourcediff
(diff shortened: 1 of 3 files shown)


krauselukas:
[Request#Show Skeleton](https://trello.com/c/A3V1iUDN/1028-requestshow-skeleton)

DavidKang:
@krauselukas, nice!, could you also add `switch_to_webui2` in the show action?

krauselukas:
@DavidKang done, but i added a statement, so that it only switches to webui2 in development

krauselukas:
@DavidKang i moved the switch_to_webui2 to the end of the show method, maybe you can check again :)

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -125,6 +125,9 @@ def show
 
     @comments = @bs_request.comments
     @comment = Comment.new
+
+    # TODO: Remove the statement after migration is finished
+    switch_to_webui2 if Rails.env.development?
```
dmarcoux: Didn't we do that before (using `if Rails.env.development?`) and it caused specs to fail when we removed it just before finishing the deliverable?

I think the condition should be `if Rails.env.development? || Rails.env.test?`

PR #6840, opened 2019-01-24
Add missing images extension to distribution icons

### LINKED ISSUE #6560: [bootstrap] Add repository from distribution missing images
selection_001
after
selection_002

As you can see, the images for RedHat, CentOS etc are missing. The only one which is there is the Kiwi image which is "ugly". I would vote for dropping this as well. This page needs anyway a redesign in the near future.

DavidKang:
Should fix #6560


DavidKang:
@Ana06, could you re-review it again please? :smile_cat: :pray:

PR #6841, opened 2019-01-24
Fix payload access in clear_caches

coolo:
Payload is actually a string hash not symbols, even though the payload_keys are listed as symbols. No suprise this went unnoticed :(


coolo:
Hopefully last words @DimStar77 :)

PR #6844, opened 2019-01-25
Switch to webui2 in test environment in request show

krauselukas:
We need to enable bootstrap in the test environment to
catch failing feature tests.

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -127,7 +127,7 @@ class Webui::RequestController < Webui::WebuiController
     @comment = Comment.new
 
     # TODO: Remove the statement after migration is finished
-    switch_to_webui2 if Rails.env.development?
+    switch_to_webui2 if Rails.env.development? || Rails.env.test?
   end
 
   def sourcediff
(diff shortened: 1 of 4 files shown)


dmarcoux:
It's not to prevent failing specs, but to catch failing specs early instead of just before finishing the deliverable.

The issue now is that since the page is not completely migrated to webui2, some specs will fail until we are done with the migration. Skipping them would bring us back to what we had before this PR, so it's not a viable solution. Someone has an idea?

DavidKang:
@dmarcoux, we can skip it only for Bootstrap using `skip_if_bootstrap` and in this case, we know that should add new tests for bootstrap.

krauselukas:
@dmarcoux allright, i changed the commit message and skiped the failing feature tests

dmarcoux:
[Enable skipped tests for Bootstrap](https://trello.com/c/wghw0C5E/1040-enable-skipped-tests-for-bootstrap)

dmarcoux:
As you can see with the link above, I created a card in our backlog to enable the skipped tests before we finish the deliverable.

DavidKang:
@coolo, thanks :smile_cat:

hennevogel:
I don't get it. We keep showing the old UI in production (fine) but then we delay fixing the tests/writing new ones with `skip_if_bootstrap` until the end of the deliverable again? Which was the horrible situation we wanted to get out of. I think you need to rethink this...

PR #6855, opened 2019-01-28
Don't repeat the same javascript behind every comment

coolo:
The error handler matches all comments, so repeating it is even harmful


dmarcoux:
Oh right, it's called over and over again since this view is called recursively.

coolo:
well, for the record: there is one. Putting this unconditionally into all pages (that could have comments at least).

dmarcoux:
True, but then we might have it when we don't need it. Anyway, it's fine as it is :slightly_smiling_face:.

PR #6858, opened 2019-01-28
 Get rid of api_obj

Ana06:
This will make @coolo happy :laughing:


coolo:
> This will make @coolo happy

I never said that - that takes more :)

PR #6859, opened 2019-01-28
Add coolo to git template

Ana06:
I have to write it every time I do something with coolo. :angry:


coolo:
My name at least doesn't require compose keys!

PR #6863, opened 2019-01-28
Merge fill_status_cache into monitor and split out differently

coolo:
This removes quite some nonsense that the code collected over the years

### DIFF
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -387,68 +387,20 @@ class Webui::ProjectController < Webui::WebuiController
   end
 
   def monitor
-    @legend = Buildresult::STATUS_DESCRIPTION
-
-    @name_filter = params[:pkgname]
-    @lastbuild_switch = params[:lastbuild]
-    if params[:defaults]
-      defaults = (begin
-                    Integer(params[:defaults])
-                  rescue ArgumentError
-                    1
-                  end) > 0
-    else
-      defaults = true
-    end
-    params['expansionerror'] = 1 if params['unresolvable']
-    monitor_set_filter(defaults)
-
-    find_opt = { project: @project, view: 'status', code: @status_filter,
-                 arch: @arch_filter, repository: @repo_filter }
-    find_opt[:lastbuild] = 1 if @lastbuild_switch.present?
-
-    @buildresult = Buildresult.find_hashed(find_opt)
-    if @buildresult.empty?
-      flash[:warning] = "No build results for project '#{@project}'"
-      redirect_to action: :show, project: params[:project]
-      return
-    end
-
-    unless @buildresult.key?('result')
+    unless (buildresult = monitor_buildresult)
       @buildresult_unavailable = true
       return
     end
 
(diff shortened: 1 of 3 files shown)


src/api/app/controllers/webui/project_controller.rb
```diff
@@ -606,52 +558,86 @@ def unlock
...
+
+    find_opt = { project: @project, view: 'status', code: @status_filter,
+                 arch: @arch_filter, repository: @repo_filter }
+    find_opt[:lastbuild] = 1 if @lastbuild_switch.present?
+
+    buildresult = Buildresult.find_hashed(find_opt)
+    if buildresult.empty?
+      flash[:warning] = "No build results for project '#{@project}'"
+      redirect_to action: :show, project: params[:project]
+      return
```
Ana06: This is now only returning to the previous method and afterward `@buildresult_unavailable = true` is set although you are going to redirect... :thinking:

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -606,52 +558,86 @@ def unlock
...
+    end
+
+    return unless buildresult.key?('result')
+    buildresult
+  end
+
+  def monitor_parse_buildresult(buildresult)
+    @packagenames = Set.new
```
Ana06: why a set? before we were not taking care of repetitions and sorting is less efficient now as you need to convert to an array....

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -387,68 +387,20 @@ def move_path
...
-    fill_status_cache
+    monitor_parse_buildresult(buildresult)
 
-    @localpackages = {}
-    @project.packages.each do |package|
-      @localpackages[package.name] = 1
+    # extract repos
+    repohash = {}
+    @statushash.each do |repo, arch_hash|
+      repohash[repo] = arch_hash.keys.sort!
```
Ana06: This is less efficient than before... you need to iterate once to get the keys and another onw to sort... and you are using this only twice... `sort!` in the first iteration it is more efficient

src/api/app/controllers/webui/project_controller.rb
coolo: Does it harm?

src/api/app/controllers/webui/project_controller.rb
coolo: we were taking care of repetitions. This set replaces

```
    @packagenames = @packagenames.flatten.uniq.sort!

    ## Filter for PackageNames ####
    @packagenames.select! { |name| filter_matches?(name, @name_filter) } if @name_filter.present?

src/api/app/controllers/webui/project_controller.rb
coolo: What first iteration? the arch_hash is a hash and the archs are its keys.

src/api/app/controllers/webui/project_controller.rb
Ana06: no, but it is not very elegant... let's say it is a nitpick...

PR #6170, opened 2018-11-02
Add webui2 version of kiwi images

hellcp:
New and better kiwi images, with all the bells and whistles

PR #6864, opened 2019-01-29
Server-side DataTable using ajax-datatables-rails gem

saraycp:
Playing around with [ajax-datatables-rails](https://github.com/jbox-web/ajax-datatables-rails) gem to use server-side pagination instead of customized client-side pagination by javascript in project index page.

The gem `ajax-datatables-rails` is proposed in the gem `jquery-datatables` [description](https://github.com/mkhairi/jquery-datatables/#server-side-processing). 

ajax_datatables_rails

Comparing with `effective_datatables` tested on PR #6686, `ajax-datatables-rails` looks simpler, has a search field for everything and allows case insensitive search.  The  default behaviour of `ajax-datatables-rails` is closer to our needs,  whereas `effective_datatables` requires some patches to get what we need.


src/api/app/assets/javascripts/webui2/project.js
```diff
@@ -1,27 +1,15 @@
-function renderProjectsTable(length) { // jshint ignore:line
-  length = length || 25;
-  var projects = mainProjects;
-  if (!$('#excludefilter').is(":checked"))
-    projects = projects.concat(exclProjects);
-  var projecturl = $("#projects-table-wrapper").data("url");
-  $("#projects-table").DataTable({
-    "data": projects,
+jQuery(document).ready(function() {
```
coolo: Still it's a good idea to only call this code when we have a projects-datatable in the page IMO

saraycp:
[Use server side pagination for the Project#Index DataTable](https://trello.com/c/q7NL91TC/1027-use-server-side-pagination-for-the-projectindex-datatable)

src/api/app/assets/javascripts/webui2/project.js
bgeuken: I am sure you have some good reasoning for that. Mind sharing?:-)

dmarcoux: It won't go through the DOM looking for something which isn't there.

coolo: There is *one* route that lists projects - and there are hundreds that haven't. But still you ask every single page to query all projects-tables.

saraycp: Do you propose to move this JS code to the view?

saraycp: @coolo @dmarcoux I just created a function in that file and call it from a `content_for :ready_function`. Please have a look :eyes:

saraycp:
@openSUSE/obs-frontend  please have a look :eyes:

src/api/app/assets/javascripts/webui2/project.js
```diff
@@ -1,27 +1,41 @@
...
-        "title": "Title",
-        "className": "text-nowrap w-25"
+function projectsDatatable() {
+  $('#projects-datatable').dataTable({
+    "processing": true,
+    "serverSide": true,
+    "ajax": {
+      "url": $('#projects-datatable').data('source'),
+      "data": function (d) {
+          d.all = $('#projects-datatable').data("all");
```
vpereira: ``` '' ``` vs ``` ""```:  every line you choose a different one :)

saraycp: @vpereira good catch!

src/api/app/assets/javascripts/webui2/project.js
bgeuken: I don't want to make this bigger as it is, but we had such discussions and concerns a couple of times by now, like isn't it bad to have `$(document).ready()` on multiple places, etc.

That's why I did a little [benchmark](https://jsperf.com/calling-selector-for-not-existing-element) so we get an idea how much time we loose when searching for a non-existing `$("#projects-table-wrapper")` element. I used my [home project](https://build.opensuse.org/project/show/home:bgeuken) for testing.
If I didn't miscalculate it takes about *0.32 microseconds* (not milliseconds!). I think it's fair to say that we don't have to worry about this:-)

> @coolo
> There is _one_ route that lists projects - and there are hundreds that haven't. But still you ask every single page to query all projects-tables.

The moment this becomes a problem we should think about modularizing our js code. But given the little amount of javascript we have right now, I don't this is an issue right now.

coolo: You're not wrong, one $() isn't the problem. But with so many other things this is about good common practises that avoid easy to avoid mistakes. 

See https://github.com/openSUSE/open-build-service/pull/6170/files#r232154799 - you end up having super-specific IDs that are global to the full application and no longer global on the view you're working on, because you never know what javascript code runs on your HTML. With having functions you can identify what code runs, with spread ready functions not so.

bgeuken: You would need a lot of $() to make this noticeably slow. And by that time there would be other things that would cause performance issues, like inefficiently and / or complicatedly written code.

Besides, the proper solution for that would still be to modularize the code (and rewriting it). What you suggests would mean to wrap thousands of selectors in thousands of functions. That doesn't sound very appealing to me.

The super-specific IDs can be tackled with writing generic code, modules, helpers.

> With having functions you can identify what code runs, with spread ready functions not so.

You would still have to mentally compile a list of function calls made by dozens of partials. The more you go down this road the more complex this gets. And it's not really, like you state, that you can not know which event handle is registered or which js function was called when the DOM was loaded.

In general I would say that writing code in a simple fashion first (unlike the example you posted) and than measuring and improving what is considerably slow is the right approach.

PR #6869, opened 2019-01-30
Refactor request show page and implement request diffing for bootstrap UI

bgeuken:
screenshot_2019-02-01 open build service

screenshot_2019-02-01 open build service 1

### DIFF
--- a/src/api/app/helpers/webui/package_helper.rb
+++ b/src/api/app/helpers/webui/package_helper.rb
@@ -84,19 +84,6 @@ module Webui::PackageHelper
     ::Cloud::UploadJob.new(filename: filename, arch: architecture).uploadable?
   end
 
-  def badge_for_diff_state(state)
-    case state
-    when 'added'
-      'badge-success'
-    when 'deleted'
-      'badge-danger'
-    when 'changed'
-      'badge-warning'
-    else
-      'badge-primary'
-    end
-  end
-
   def expand_diff?(filename, state)
     state != 'deleted' &&
       !filename.include?('/') &&
--- a/src/api/app/helpers/webui/package_helper.rb
+++ b/src/api/app/helpers/webui/package_helper.rb
@@ -88,6 +88,10 @@ module Webui::PackageHelper
     state != 'deleted' && !filename.include?('/') && (filename == '_patchinfo' || filename.ends_with?('.spec', '.changes'))
   end
 
+  def viewable_file?(filename)
+    !Package.is_binary_file?(filename) && !filename.include?('/')
+  end
+
   def calculate_revision_on_state(revision, state)
     result = revision.to_i
     result -= 1 if state == 'deleted'
--- a/src/api/app/helpers/webui/package_helper.rb
+++ b/src/api/app/helpers/webui/package_helper.rb
@@ -85,9 +85,7 @@ module Webui::PackageHelper
   end
 
   def expand_diff?(filename, state)
-    state != 'deleted' &&
-      !filename.include?('/') &&
-      (filename == '_patchinfo' || filename.ends_with?('.spec') || filename.ends_with?('.changes'))
+    state != 'deleted' && !filename.include?('/') && (filename == '_patchinfo' || filename.ends_with?('.spec', '.changes'))
   end
 
   def calculate_revision_on_state(revision, state)
--- a/src/api/app/helpers/webui/request_helper.rb
+++ b/src/api/app/helpers/webui/request_helper.rb
@@ -1,4 +1,7 @@
 module Webui::RequestHelper
+  include Webui::UserHelper
+  include Webui::WebuiHelper
+
   STATE_COLORS = {
     'new' => 'green',
     'accepted' => 'green',
@@ -147,4 +150,47 @@ module Webui::RequestHelper
       end
     end
   end
+
+  # rubocop:disable Style/FormatStringToken, Style/FormatString
+  def request_action_header(action, creator)
+    source_project_hash = { project: action[:sprj], package: action[:spkg] }
+
+    case action[:type]
+    when :submit
+      'Submit %{source_container} to %{target_container}' % {
+        source_container: project_or_package_link(source_project_hash),
+        target_container: project_or_package_link(project: action[:tprj], package: action[:tpkg])
+      }
+    when :delete
+      target_repository = link_to(action[:trepo], repository_index_path(project: action[:tprj], repository: action[:trepo])) if action[:trepo]
+      'Delete %{target_repository}%{target_container}' % {
+        target_repository: target_repository,
+        target_container: project_or_package_link(project: action[:tprj], package: action[:tpkg])
+      }
+    when :add_role, :set_bugowner
+      '%{creator} wants %{requester} to %{task} for %{target_container}' % {
+        creator: user_with_realname_and_icon(creator),
+        requester: requester_str(creator, action[:user], action[:group]),
+        task: creator_intentions(action[:role]),
+        target_container: project_or_package_link(project: action[:tprj], package: action[:tpkg])
+      }
+    when :change_devel
+      'Set the devel project to %{source_container} for %{target_container}' % {
+        source_container: project_or_package_link(source_project_hash),
+        target_container: project_or_package_link(project: action[:tprj], package: action[:tpkg])
+      }
+    when :maintenance_incident
+      source_project_hash.update(homeproject: creator)
+      'Submit update from %{source_container} to %{target_container}' % {
+        source_container: project_or_package_link(source_project_hash),
+        target_container: project_or_package_link(project: action[:tprj], package: action[:tpkg])
+      }
+    when :maintenance_release
+      'Release %{source_container} to %{target_container}' % {
+        source_container: project_or_package_link(source_project_hash),
+        target_container: project_or_package_link(project: action[:tprj], package: action[:tpkg])
+      }
+    end.html_safe
+  end
+  # rubocop:enable Style/FormatStringToken, Style/FormatString
 end
--- a/src/api/app/views/webui2/webui/package/_revision_diff_detail_buttons.html.haml
+++ b/src/api/app/views/webui2/webui/package/_revision_diff_detail_buttons.html.haml
@@ -7,7 +7,7 @@
   href: "#revision_details_#{index + 1}",
   title: 'Go down to the next diff' }
     %i.fas.fa-chevron-down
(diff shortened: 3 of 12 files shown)


src/api/app/helpers/webui/request_helper.rb
```diff
@@ -139,4 +142,47 @@ def diff_data(action_type, sourcediff)
   def diff_label(diff)
     "#{diff['project']} / #{diff['package']} (rev #{diff['rev']})"
   end
+
+  # rubocop:disable Style/FormatStringToken, Style/FormatString
+  def request_action_header(action, creator)
+    source_project_hash = { project: action[:sprj], package: action[:spkg] }
+
+    case action[:type].to_sym
```
coolo: I like this one - I split this into partials per action type, but this makes changes quite hard

src/api/app/helpers/webui/package_helper.rb
```diff
@@ -87,7 +87,7 @@ def uploadable?(filename, architecture)
   def expand_diff?(filename, state)
     state != 'deleted' &&
       !filename.include?('/') &&
-      (filename == '_patchinfo' || filename.ends_with?('.spec') || filename.ends_with?('.changes'))
+      (filename == '_patchinfo' || filename.ends_with?('.spec', '.changes'))
```
Ana06: [nitpick] now that it is shorter, you could write the whole condition in a single line

src/api/app/views/webui2/webui/request/_sourcediff_tab.html.haml
```diff
@@ -0,0 +1,37 @@
+%div
+  %b
+    = request_action_header(action, request_creator)
+  -# FIXME: request forwarding # here or decision box?
```
Ana06: ??? :eyes:

src/api/app/views/webui2/webui/request/_sourcediff_tab.html.haml
```diff
@@ -0,0 +1,37 @@
...
+  -# FIXME: make maintainer
```
Ana06: ?? :eyes:

src/api/app/views/webui2/webui/request/show.html.haml
```diff
@@ -1,15 +1,27 @@
...
     -# description content
 .card.mb-3
-  .card-header.d-flex.justify-content-between
-    %h5
-      Diff
+  .bg-light
+    %ul.nav.nav-tabs{ role: 'tablist' }
+      - @actions.each_with_index do |action, index|
+        %li.nav-item
+          = link_to(action[:name], "##{valid_xml_id(action[:name])}", class: "nav-link text-nowrap #{('active' if index == 0)}",
```
Ana06: `index.zero?` (if not you are introducing a Rubocop offense, Rubocop doesn't complain because we have many offenses, but the cop is enabled and agreed)

src/api/app/views/webui2/webui/request/show.html.haml
```diff
@@ -1,15 +1,27 @@
...
+            data: { toggle: 'tab' }, role: 'tab')
   .card-body
-    -# Diff content and build results
+    .tab-content
+      - @actions.each_with_index do |action, index|
+        .tab-pane.fade.show{ id: valid_xml_id(action[:name]), class: ('active' if index == 0) }
```
Ana06: same here

src/api/app/helpers/webui/package_helper.rb
```diff
@@ -103,9 +103,15 @@ def expand_diff?(filename, state)
...
+  def viewable_file?(filename)
+    !Package.is_binary_file?(filename) && !filename.include?('/')
+  end
+
+  def previous_revision(revision)
+    # rubocop:disable Style/RescueModifier
+    is_integer = (Integer(revision) rescue false)
+    # rubocop:enable Style/RescueModifier
+
+    revision.to_i - 1 if is_integer
```
Ana06: what about writing this as `revision.to_i - 1 if revision.present?`

src/api/app/views/webui2/webui/package/_revision_diff_detail_buttons.html.haml
```diff
@@ -7,9 +7,10 @@
   href: "#revision_details_#{index + 1}",
   title: 'Go down to the next diff' }
     %i.fas.fa-chevron-down
-  - link_options = { project: project, package: package, filename: filename, rev: calculate_revision_on_state(revision, state) }
-  - if !@linkinfo && (has_content && !Package.is_binary_file?(filename) && !filename.include?('/'))
-    = link_to package_view_file_path(link_options),
+  - revision = previous_revision(revision) if state == 'deleted'
```
Ana06: Why not using the `@last_rev` variable? instead of the `previous_revision` method? :thinking:

src/api/app/helpers/webui/request_helper.rb
bgeuken: tbh, I am still not 100% happy with it. But I guess it's good enough for now.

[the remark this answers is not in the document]
coolo: because of this insane webui_infos function

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -139,4 +142,47 @@ def diff_data(action_type, sourcediff)
   def diff_label(diff)
     "#{diff['project']} / #{diff['package']} (rev #{diff['rev']})"
   end
+
+  # rubocop:disable Style/FormatStringToken, Style/FormatString
```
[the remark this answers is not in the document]
bgeuken: One of them prefers `%<foo>`  over `%{foo}` but that causes argument errors. The other is configured to use `format` instead of `String#%`.
Since applying this would make the whole statement messier, I left it as is.

src/api/app/helpers/webui/request_helper.rb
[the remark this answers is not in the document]
bgeuken: To support both, strings and symbols. Does it hurt?

src/api/app/views/webui2/webui/request/_sourcediff_tab.html.haml
bgeuken: The whole feature is WIP as you know. That's two features that are part of the diff, but also require some work on the decision box.
I'll remove it.

src/api/app/views/webui2/webui/package/_revision_diff_detail_buttons.html.haml
bgeuken: That said, I would prefer this in a separate PR, since it's not clear yet if and how this can be solved differently.

src/api/app/helpers/webui/package_helper.rb
bgeuken: I re-arranged the code and dropped this method again. We could only run into this via request diffs. And this is taken care of differently now.

src/api/app/helpers/webui/request_helper.rb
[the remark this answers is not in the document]
Ana06: What causes argument errors? :confused:

src/api/app/helpers/webui/request_helper.rb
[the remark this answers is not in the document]
Ana06: I guess nobody will die because of it, but it is always a symbol and it was not converted before :confused: let say **nitpick**

src/api/app/views/webui2/webui/request/_sourcediff_tab.html.haml
Ana06: I just didn't understand the coment

src/api/app/helpers/webui/package_helper.rb
Ana06: but if it is present is always an integer... if not you are doing something wrong somewhere else....

src/api/app/helpers/webui/request_helper.rb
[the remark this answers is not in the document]
coolo: We'll come to it- the TODO is only 7 years old :)

```
  def webui_actions(opts = {})
    # TODO: Fix!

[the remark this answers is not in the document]
Comment (bgeuken) [reply src/api/app/helpers/webui/request_helper.rb]: Well, the fix is to remove the webui_actions and webui_infos methods and use active record relations. But I can refactor / rewrite everything in this PR:laughing:

File: src/api/app/helpers/webui/package_helper.rb (hunk 7 above)
Comment (bgeuken) [reply src/api/app/helpers/webui/package_helper.rb]: I rewrote the code around this (rendering the partials from the package diff and request diff controller). So this code is gone.

Though your statement is wrong "it" is not always an integer. Revisions can be hashes. That's why there is the `@linkinfo` variable.

File: src/api/app/helpers/webui/request_helper.rb (hunk 1 above)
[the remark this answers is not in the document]
Comment (coolo) [reply src/api/app/helpers/webui/request_helper.rb]: Well, I was just joking - because Ana's nitpick is right. :type in webui_infos is BsRequestAction#action_type which is a symbol for valid actions.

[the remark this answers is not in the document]
Comment (coolo) [reply src/api/app/helpers/webui/request_helper.rb]: it's still a nitpick though :)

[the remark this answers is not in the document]
Comment (bgeuken) [reply src/api/app/helpers/webui/request_helper.rb]: *shrug* I think you guys have the wrong focus:P

File: src/api/app/helpers/webui/request_helper.rb (hunk 9 above)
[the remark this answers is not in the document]
Comment (Ana06) [reply src/api/app/helpers/webui/request_helper.rb]: [nitpick] What about regenerating the todo instead of excluding it here with the idea of solving this at some point in the future?

File: src/api/app/views/webui2/webui/request/_sourcediff_tab.html.haml [hunk 10]
```diff
@@ -0,0 +1,31 @@
+%div
+  %b
+    = request_action_header(action, request_creator)
+  %hr
+- if action[:type] == :maintenance_incident && action[:releaseproject]
+  %i
+    Release in #{action[:releaseproject]}
+- action[:sourcediff].to_a.each do |sourcediff|
```
Ana06: why `to_a`?

bgeuken: It was before (in the old UI) wrapped in a *.present? check, which I didn't like...

```ruby
- if action[:sourcediff].present?
  - action[:sourcediff].each do |sourcediff|
```

I changed it now to `(action[:sourcediff] || []).each`. Hope you can live with it?:-)

PR #6870, opened 2019-01-30
Update default icon for comments

bgeuken:

### DIFF
--- a/src/api/app/views/webui2/webui/comment/_show.html.haml
+++ b/src/api/app/views/webui2/webui/comment/_show.html.haml
@@ -1,5 +1,5 @@
 .comments
-  - commentable.comments.without_parent.each do |comment|
+  - commentable.comments.without_parent.includes(:user, :children).each do |comment|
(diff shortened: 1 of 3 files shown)


dmarcoux:
I like this solution better!

src/api/app/views/webui2/webui/comment/_show.html.haml
```diff
@@ -1,5 +1,5 @@
 .comments
-  - commentable.comments.without_parent.each do |comment|
+  - commentable.comments.without_parent.includes(:user, :children).each do |comment|
```
Ana06: I think it would be even better to raise a query for every children but that loads user (It should be more efficient)

coolo:
comment/show is still used twice in the request/show - and afaik you don't plan to redo the page to show supersede comments differently.

bgeuken:
We currently migrate the request show page to the new UI. Let's see if that would be an issue there...

bgeuken:
Let's ship it. If it becomes a problem in the new view, we can take care of it

PR #6875, opened 2019-01-30
[api] support generic github issues

adrianschroeter:
write as

 gh#$project/$repo#$issue

eg

 gh#openSUSE/build#123

needs unfortunality a special hack, since github URLs expect /issues/
inside.

### DIFF
--- a/src/api/app/models/issue_tracker.rb
+++ b/src/api/app/models/issue_tracker.rb
@@ -32,6 +32,7 @@ class IssueTracker < ApplicationRecord
   def show_url_for(issue, html = nil)
     return unless issue
     url = show_url.gsub('@@@', issue)
+    url.gsub!(/(github#|gh#)/, '').gsub!('#', '/issues/') if kind == 'github'
     return "<a href=\"#{url}\">#{CGI.escapeHTML(show_label_for(issue))}</a>" if html
     url
   end
--- a/src/api/app/models/issue.rb
+++ b/src/api/app/models/issue.rb
@@ -83,6 +83,7 @@ class Issue < ApplicationRecord
   end
 
   def url
+    return issue_tracker.show_url_for(label) if issue_tracker.kind == 'github'
     issue_tracker.show_url.gsub('@@@', name)
   end
 
--- a/src/api/db/data/20190130115727_add_github_issue_tracker.rb
+++ b/src/api/db/data/20190130115727_add_github_issue_tracker.rb
@@ -1,0 +1,13 @@
+class AddGithubIssueTracker < ActiveRecord::Migration[5.1]
+  def up
+    IssueTracker.where(name: 'gh').first_or_create(description: 'Generic github tracker',
+                                                   kind: 'github',
+                                                   regex: '(?:gh|github)#(\w+\/\w+#\d+)',
+                                                   url: 'https://www.github.com',
+                                                   label: 'gh#@@@', show_url: 'https://github.com/@@@')
+  end
+
+  def down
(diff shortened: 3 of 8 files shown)


src/api/db/migrate/20190130115727_add_github_issue_tracker.rb
```diff
@@ -0,0 +1,13 @@
+class AddGithubIssueTracker < ActiveRecord::Migration[5.1]
```
hennevogel: This is a data migration and belongs to `db/data`

src/api/app/models/issue_tracker.rb
```diff
@@ -52,6 +52,9 @@ def valid_issue_name?(name)
   def show_url_for(issue, html = nil)
     return unless issue
     url = show_url.gsub('@@@', issue)
+    # generic github tracker needs special handling since they inject '/issues/' inside
+    # github handles issue/pull redirection on it's own luckily
+    url.gsub!(/\#/, '/issues/') if kind == 'github'
```
Ana06: Why need this to be a reg exp? :confused:

adrianschroeter:
On Sonntag, 3. Februar 2019, 10:11:24 CET Ana María Martínez Gómez wrote:
> Ana06 commented on this pull request.
> 
> > @@ -52,6 +52,9 @@ def valid_issue_name?(name)
> 
>    def show_url_for(issue, html = nil)
>      return unless issue
>      url = show_url.gsub('@@@', issue)
> +    # generic github tracker needs special handling since they inject
> '/issues/' inside +    # github handles issue/pull redirection on it's own
> luckily
> +    url.gsub!(/\#/, '/issues/') if kind == 'github'
> 
> Why need this to be a reg exp? :confused:

what is the alternative to replace the hash?

-- 

Adrian Schroeter
email: adrian@suse.de

coolo:
> what is the alternative to replace the hash?

If you pass a string, it's automatically converted to regexp. Aka `gsub('#', ...`

adrianschroeter:
On Sonntag, 3. Februar 2019, 13:18:12 CET Stephan Kulow wrote:
> > what is the alternative to replace the hash?
> 
> If you pass a string, it's automatically converted to regexp. Aka `gsub('#',
> ...`

okay, but does it make a difference?

-- 

Adrian Schroeter
email: adrian@suse.de

coolo:
> okay, but does it make a difference?

Well, it's a little easier on the eyes - but other than that: no

src/api/db/structure.sql
```diff
@@ -1435,6 +1435,7 @@ INSERT INTO `schema_migrations` (version) VALUES
 ('20181113095753'),
 ('20181201065026'),
 ('20190111130416'),
-('20190115131711');
+('20190115131711'),
+('20190130115727');
```
hennevogel: Data migrations do not write into the `schema_migrations` table, they write to `data_migrations` (which we do not track in the schema as it's instance specific). Just get rid of this chunk.

hennevogel:
@adrianschroeter are you going to move this to `db/data` or what?

Ana06:
what is the state of this? :thinking:

src/api/db/migrate/20190130115727_add_github_issue_tracker.rb
hennevogel: Moved it...

src/api/app/models/issue_tracker.rb
hennevogel: Changed to `'#'`

PR #6879, opened 2019-01-30
Use User login instead id for comment events

### LINKED ISSUE #4026: RabbitMQ sees users as ids
```
2017-10-16 16:02:15:979: Message published

Exchange:     pubsub
Routing keys: [<<"opensuse.obs.project.comment">>]
Payload: 
{"project":"home:coolo","commenters":[36,5],"commenter":5,"comment_body":"test"}
```

While the project and the comment are strings, the users are given as user ids. As the OBS API has no way to retrieve the user name from these IDs, they need to be mapped.

DavidKang:
closes #4026

### DIFF
--- a/src/api/app/models/comment.rb
+++ b/src/api/app/models/comment.rb
@@ -27,7 +27,7 @@ class Comment < ApplicationRecord
   end
 
   def create_notification(params = {})
-    params[:commenter] = user.id
+    params[:commenter] = user.login
     params[:comment_body] = body
     params[:commenters] = involved_users
 
@@ -52,15 +52,15 @@ class Comment < ApplicationRecord
   def involved_users
     users = Set.new
     users_mentioned = Set.new
-    Comment.where(commentable: commentable).find_each do |comment|
+    Comment.where(commentable: commentable).includes(:user).find_each do |comment|
       # take the one making the comment
-      users << comment.user_id
+      users << comment.user.login
       # check if users are mentioned (regexp borrowed from user model - with whitespace removed)
-      comment.body.scan(%r{@([\w\^\-\.#\*\+&'"]*)}).each do |user|
-        users_mentioned << user[0]
+      comment.body.scan(%r{@([\w\^\-\.#\*\+&'"]*)}).each do |user_login|
+        users_mentioned << user_login.first
       end
     end
-    users += User.where(login: users_mentioned.to_a).pluck(:id)
+    users += User.where(login: users_mentioned).pluck(:login)
     users.to_a
   end
 
--- a/src/api/app/models/event/comment_event.rb
+++ b/src/api/app/models/event/comment_event.rb
@@ -10,17 +10,17 @@ module Event
 
     def expanded_payload
       p = payload.dup
-      p['commenter'] = User.find(p['commenter'])
+      p['commenter'] = User.find_by(login: p['commenter'])
       p
     end
 
     def originator
-      User.find(payload['commenter'])
+      User.find_by(login: payload['commenter'])
     end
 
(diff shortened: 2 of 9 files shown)


src/api/app/models/comment.rb
```diff
@@ -50,18 +50,17 @@ def create_notification(params = {})
...
       # check if users are mentioned (regexp borrowed from user model - with whitespace removed)
-      comment.body.scan(%r{@([\w\^\-\.#\*\+&'"]*)}).each do |user|
-        users_mentioned << user[0]
+      comment.body.scan(%r{@([\w\^\-\.#\*\+&'"]*)}).each do |user_login|
+        users_mentioned << user_login.first
       end
     end
-    users += User.where(login: users_mentioned.to_a).pluck(:id)
-    users.to_a
+    users |= User.where(login: users_mentioned).pluck(:login)
```
Ana06: [nitpick] I see some refactoring here to improve SQL queries! :nerd_face: (I can send a PR)

src/api/app/models/comment.rb
```diff
@@ -50,18 +50,17 @@ def create_notification(params = {})
 
   # build an array of users, commenting or being mentioned on the commentable of this comment
   def involved_users
-    users = Set.new
-    users_mentioned = Set.new
+    users = []
```
Ana06: using set for users should be more efficient (the change in users_mentioned makes sense)

src/api/app/models/event/comment_for_project.rb
```diff
@@ -7,7 +7,7 @@ class CommentForProject < Base
     receiver_roles :maintainer, :bugowner, :watcher
 
     def subject
-      "New comment in project #{payload['project']} by #{User.find(payload['commenter']).login}"
+      "New comment in project #{payload['project']} by #{User.find_by_login(payload['commenter']).login}"
```
Ana06: `payload['commenter'])` only! :see_no_evil:

DavidKang: right!

src/api/app/models/comment.rb
coolo: I would guess the same - Sets are underused! :)

DavidKang:
@Ana06, because I wanted to receive some feedback and the tests need to be fixed.

Ana06:
@DavidKang  ping!

hennevogel:
@DavidKang are you going to address this or can we throw this back to the bug squad?

DavidKang:
@hennevogel, we can throw it back

bgeuken:
@DavidKang I've updated the commit (had to change the queries in [app/models/event/comment_event.rb](https://github.com/openSUSE/open-build-service/pull/6879/files#diff-2fb56c28b9bc87d652cfc47ff9f11139) ). Please have a look.

src/api/app/models/event/comment_event.rb
```diff
@@ -10,17 +10,17 @@ def self.included(base)
 
     def expanded_payload
       p = payload.dup
-      p['commenter'] = User.find(p['commenter'])
+      p['commenter'] = User.find_by(login: p['commenter'])
```
hennevogel: Hm this might mean a window of time where we have Events with old payload. I guess we need to deal with that in a data migration.

bgeuken: Good point

bgeuken:
@hennevogel Updated. Please review again.

bgeuken:
Adding the 'do not merge' label: This requires a data migration to be run. We should wait for the next maintenance window before merging.

Issue #6672, opened 2018-12-19
comments can't be linked


Ana06:
I would like to send a url that takes you to a comment in the project page, but it is not possible (or not intuitive because I didn't find out how to do it).

My first try was clicking in the time of the comment as I usually do in Github, but it does work. :sob: 

I want to link my comments! :pray:

PR #6885, opened 2019-01-31
Add anchor tag to comments

bgeuken:
This allows users to reference to comments by linking to them.

### DIFF
--- a/src/api/app/views/webui2/webui/comment/_content.html.haml
+++ b/src/api/app/views/webui2/webui/comment/_content.html.haml
@@ -5,7 +5,7 @@
     %p
       = link_to(comment.user, user_show_path(comment.user))
       wrote
-      %u{ title: l(comment.created_at.utc) }
+      = link_to("#comment-#{comment.id}", title: l(comment.created_at.utc), name: "comment-#{comment.id}") do
         #{time_ago_in_words(comment.created_at)} ago
     = render_as_markdown(comment)
     = render partial: 'webui2/webui/comment/links', locals: { comment: comment }


src/api/app/views/webui2/webui/comment/_content.html.haml
```diff
@@ -5,7 +5,7 @@
     %p
       = link_to(comment.user, user_show_path(comment.user))
       wrote
-      %u{ title: l(comment.created_at.utc) }
+      = link_to("#comment-#{comment.id}", title: l(comment.created_at.utc), name: "comment-#{comment.id}") do
```
dmarcoux: I find it a bit arbitrary to have the link on the timestamp.

What do you think about adding an icon next to the timestamp, like `%i.fas.fa-link.fa-sm{ title: 'Link to comment' }`? It would make this link a lot more obvious. I'm not so sure about the title, maybe it's not needed. Clicking the icon could also copy the link to the clipboard.

It would look like this:
suggestion

bgeuken: I find the icon a bit to noisy. For pages with many comments that would be a lot of icons.

I think users will figure out that this is a link and what it does. Just like they do with GitHub's coments.

coolo: I find the underlining hint enough that it's a link - underlining text that is *not* a link is actually very confusing and I wonder why it was done this way.

coolo:
Fixes https://github.com/openSUSE/open-build-service/issues/6672

src/api/app/views/webui2/webui/comment/_content.html.haml
Ana06: I do it in Github all the time, I like it as it is.

vpereira: > I find the icon a bit to noisy. For pages with many comments that would be a lot of icons.
> 
> I think users will figure out that this is a link and what it does. Just like they do with GitHub's coments.

noisy and why just there and not in all other links.. anything special about this link?

PR #6889, opened 2019-02-01
Fix flickering backend subscriber test

bgeuken:
Application and webui controller are both setting a request value
in the InfluxDB::Rails.current.values hash.
This was causing the test to fail randomly when run within our CI
(aka running multiple tests) because the request value got set.

### DIFF
--- a/src/api/spec/lib/influx_db/obs/middleware/backend_subscriber_spec.rb
+++ b/src/api/spec/lib/influx_db/obs/middleware/backend_subscriber_spec.rb
@@ -21,9 +21,8 @@ RSpec.describe InfluxDB::OBS::Middleware::BackendSubscriber do
 
     let(:result) do
       {
-        values: {
-          value: 2000
-        },
+        # :request might be nil. In that case we have to filter it out.
+        values: InfluxDB::Rails.current.values.reject { |_, v| v.nil? }.update(value: 2000),
         tags: {
           http_status_code: 200,
           http_method: 'GET',


bgeuken:
I can't say for sure that this would fix it. We will see. But in the last 10 test runs it didn't happen anymore.

coolo:
the CI failed :)

src/api/spec/lib/influx_db/obs/middleware/backend_subscriber_spec.rb
```diff
@@ -21,9 +21,8 @@
 
     let(:result) do
       {
-        values: {
-          value: 2000
-        },
+        # :request might be nil. In that case we have to filter it out.
+        values: InfluxDB::Rails.current.values.reject { |_, v| v.nil? }.update(value: 2000),
```
Ana06: why do you need to reject things? :confused:

bgeuken: [Apparently](https://circleci.com/gh/bgeuken/open-build-service/1749#tests/containers/2) there can be a nil value in the request. And the backend subscriber [filters](https://github.com/openSUSE/open-build-service/blob/master/src/api/lib/influxdb_obs/obs/middleware/backend_subscriber.rb#L38) nil values.

bgeuken:
> coolo commented an hour ago
> the CI failed :)

The [CircleCI tests](https://circleci.com/gh/bgeuken/open-build-service/tree/fix_flickering_test) look fine. You mean our package builds?

src/api/spec/lib/influx_db/obs/middleware/backend_subscriber_spec.rb
Ana06: why do you need to do it as well then?

bgeuken: Because this hash is used in the [test](https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/lib/influx_db/obs/middleware/backend_subscriber_spec.rb#L48) as expected parameter when `write_point` is called.

PR #6893, opened 2019-02-01
Migrate request history to bootstrap

krauselukas:
peek 2019-02-01 14-03

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/collapse-text.scss
+++ b/src/api/app/assets/stylesheets/webui2/collapse-text.scss
@@ -1,0 +1,37 @@
+/*
+Copyright (c) 2019 by Natalia (https://codepen.io/natonischuk/pen/QbGWBa)
+Fork of an original work by Natalia (https://codepen.io/natonischuk/pen/mJOdmJ)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in the
+Software without restriction, including without limitation the rights to use, copy,
+modify, merge, publish, distribute, sublicense, and/or sell copies of the Software,
+and to permit persons to whom the Software is furnished to do so, subject to the
+following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies
+or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+@mixin collapsedText($lineHeight: 1.2em, $lineCount: 1) {
+  overflow: hidden;
+  position: relative;
+  line-height: $lineHeight;
+  max-height: $lineHeight * $lineCount;
+  text-align: justify;
+  margin-right: -1em;
+  padding-right: 1em;
+  &:before {
+    content: ' ...';
+    position: absolute;
+    right: 0;
+    bottom: 0;
+  }
+}
--- a/src/api/app/assets/stylesheets/webui2/texts.scss
+++ b/src/api/app/assets/stylesheets/webui2/texts.scss
@@ -27,3 +27,7 @@
 .ml-0_6 {
   margin-left: 0.6rem !important;
 }
+
+@mixin uncollapsedText($lineHeight: 1.2em) {
+  line-height: $lineHeight;
+}
--- a/src/api/app/views/webui2/webui/request/_request_history.html.haml
+++ b/src/api/app/views/webui2/webui/request/_request_history.html.haml
@@ -1,0 +1,34 @@
+%ul.list-unstyled
+  %li.media
+    - user = User.find_by_login(bs_request.creator)
+    = image_tag_for(user, size: 32, custom_class: 'mr-3', width: nil)
+    .media-body
+      = render partial: 'request_history_heading', locals: { creator: bs_request.creator,
+                                                             created_at: bs_request.created_at }
+      created request
+      = render partial: 'request_history_body', locals: { text: bs_request.description }
+
+  - history.each do |element|
+    %li.media.mt-3
+      = image_tag_for(element.user, size: 32, custom_class: 'mr-3', width: nil)
+      .media-body
+        = render partial: 'request_history_heading', locals: { creator: element.user.login,
+                                                             created_at: element.created_at }
+        - if element.class == HistoryElement::RequestSuperseded
+          = link_to(element.description, request_show_path(element.description_extension))
+        - else
+          #{element.description}
+        = render partial: 'request_history_body', locals: { text: element.comment }
+
+= content_for :ready_function do
+  :plain
+
+    $('.req-description-box').on('click', '#req-description', function() {
+      var link = $(this).siblings('#show-full-req-description');
+      collapseHistory(this, link);
+    });
+
+    $('.req-description-box').on('click', '#show-full-req-description', function() {
+      var desc = $(this).siblings('#req-description');
+      collapseHistory(desc, this);
+    });
--- a/src/api/app/assets/javascripts/webui2/request.js
+++ b/src/api/app/assets/javascripts/webui2/request.js
@@ -204,3 +204,13 @@ function setupActionLink() { // jshint ignore:line
     return false;
   });
 }
+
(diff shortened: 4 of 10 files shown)


krauselukas:
[Request History](https://trello.com/c/X6v6gDtA/1033-request-history)

src/api/app/views/webui2/webui/request/_request_history.html.haml
```diff
@@ -0,0 +1,47 @@
...
+        .media-body
+          - unless bs_request.description.blank?
+            .req-description-box
+              - if bs_request.description.length > 100
+                = simple_format(bs_request.description, class: 'collapsed-history', id: 'req-description')
+                = link_to('', '###', id: "show-full-req-description", class: 'collapsed-link', title: 'show more')
+              - else
+                = simple_format(bs_request.description)
+
+  - history.each_with_index do |element, index|
```
Ana06: you need to include elements and users to avoid extra queries :bulb:

src/api/app/views/webui2/webui/request/_request_history.html.haml
```diff
@@ -0,0 +1,47 @@
...
+    %li.media.mt-3
+      = image_tag_for(element.user, size: 32, custom_class: 'mr-3', width: nil)
+      .media-body
+        %h5.mt-0.mb-1 #{element.user.login} (#{fuzzy_time(element.created_at)})
```
Ana06: [nitpick] repeated code.... :eyes: maybe a helper?

src/api/app/views/webui2/webui/request/_request_history.html.haml
```diff
@@ -0,0 +1,47 @@
...
+              .req-description-box
+                - if element.comment.length > 100
+                  = simple_format(element.comment, class: 'collapsed-history', id: 'req-description')
+                  = link_to('', '###', class: 'collapsed-link', id: 'show-full-req-description', title: 'show more')
+                - else
+                  = simple_format(element.comment)
+
+= content_for :ready_function do
+  :plain
+    $('.req-description-box').on('click', '#req-description', function() {
```
Ana06: why do you need the `#req-description` paramether?

Ana06: repeated code, you can move it to a method :wink:

vpereira:
> I think the arrow to close should be up as well, because if not you need to scroll down to close it...
@Ana06 
you don't need. just click on the whole div and it will collapse/uncollapse the text. The link is there just for those want to use it :)

src/api/app/views/webui2/webui/request/_request_history.html.haml
krauselukas: I think this is done in the request_controller through this line ```@history = @bs_request.history_elements.includes(:user)``` or do you mean something else?

src/api/app/views/webui2/webui/request/_request_history.html.haml
vpereira: > repeated code....  maybe a helper?

or partial if possible..

src/api/app/assets/stylesheets/webui2/texts.scss
```diff
@@ -27,3 +27,24 @@
 .ml-0_6 {
   margin-left: 0.6rem !important;
 }
+
+//inspired by https://codepen.io/natonischuk/pen/QbGWBa
```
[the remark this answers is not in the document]
krauselukas: done :D

src/api/app/views/webui2/webui/request/_request_history.html.haml
vpereira: or partial..

src/api/app/assets/stylesheets/webui2/collapse-text.scss
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright (c) 2019 by Natalia (https://codepen.io/natonischuk/pen/QbGWBa)
+Fork of an original work by Natalia (https://codepen.io/natonischuk/pen/mJOdmJ)
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in the
+Software without restriction, including without limitation the rights to use, copy,
+modify, merge, publish, distribute, sublicense, and/or sell copies of the Software,
+and to permit persons to whom the Software is furnished to do so, subject to the
+following conditions: The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software. THE SOFTWARE IS PROVIDED
```
coolo: As maintainer of cavil: Please format this like a MIT license is formatted. Aka: include some whitespace:

https://spdx.org/licenses/MIT.html#licenseText

src/api/app/views/webui2/webui/request/_request_history.html.haml
coolo: partials make it *much* slower!

src/api/app/assets/stylesheets/webui2/collapse-text.scss
krauselukas: done

src/api/app/views/webui2/webui/request/_request_history.html.haml
krauselukas: i moved the media heading and body to an partial, i don't have a strong opinion on that :sweat_smile:

src/api/app/views/webui2/webui/request/_request_history.html.haml
vpereira: @krauselukas the pattern ``` @mycollection.each { |e| render partial: 'foo', locals {} ``` is one that makes the partial usage slow, specially in the history section which can become huge and there is no way to support pagination.... I would suggest you to use partials but to pass the collection to render i.e ```= render partial: 'foo', collection: @my_collection```

some reading material:
https://medium.com/@coorasse/partial-rendering-performance-in-rails-101fdfb6ffb9
https://api.rubyonrails.org/classes/ActionView/PartialRenderer.html

krauselukas:
@vpereira @Ana06  i had a look into the links you provided, it makes totally sense. But in this case i'm not sure how to make use of it. I use the partial two times in this context, once for the creator and then for everything else. In both cases i need to access different attributes of the object i pass. I was able to work around it by passing different attributes through the locals. Not sure how this is possible if i use collections. For example for the creator i pass ```locals: {creator: bs_request.creator, created_at: bs_request.created_at }``` and for the history ```locals: { creator: element.user.login, created_at: element.created_at }```.  So in one case i need to access ```user.login``` and in the other case ```creator```. Is it possible without having a lot of statements in the partial itself? If not i would suggest to revert the changes, and doing it without partials, since both parts a slightly different.

src/api/app/views/webui2/webui/request/_request_history.html.haml
```diff
@@ -0,0 +1,34 @@
...
+  %li.media
+    - user = User.find_by_login(bs_request.creator)
+    = image_tag_for(user, size: 32, custom_class: 'mr-3', width: nil)
+    .media-body
+      = render partial: 'request_history_heading', locals: { creator: bs_request.creator,
+                                                             created_at: bs_request.created_at }
+      created request
+      = render partial: 'request_history_body', locals: { text: bs_request.description }
+
+  - history.each do |element|
```
DavidKang: IMO, this could be into a partial, but using `collection` instead of `each` [[1]](https://github.com/openSUSE/open-build-service/pull/6893/commits/e4ccc1a1a2572d0d35c8e4df30c07c7c86944822#r253272363)

src/api/app/views/webui2/webui/request/_request_history.html.haml
DavidKang: I agree with @Ana06, this could into a function to avoid repeated code

vpereira:
@krauselukas did you remove the permissions to the admins to push more code to it? I would like to finish this PR since it's blocking the other ones.. otherwise i will just copy your changes and open a new PR to update it.  Thank you!

src/api/app/views/webui2/webui/request/_request_history.html.haml
Ana06: I meant comments and users. I think comments are not loaded and used... :thinking:

src/api/app/assets/stylesheets/webui2/collapse-text.scss
Ana06: This was the license they provide. I think that @krauselukas just copied it over

src/api/app/views/webui2/webui/request/_request_history.html.haml
Ana06: @DavidKang 

> IMO, this could be into a partial, but using collection instead of each [1]

mm..? :confused: you still need `each` if you use a collection.... :confused:

src/api/app/assets/stylesheets/webui2/collapse-text.scss
coolo: The license doesn't change by formatting it nice

vpereira:
> this is not intuitive.... and goes again the freedom usability principle...
LOL what?

"The user control and freedom usability principle describes a system’s ability to
provide users with an efficient “escape route” in the event that the user operates the 
system incorrectly. This may be caused by selecting an option by mistake or entering the
wrong information. In either case, the system should provide “undo” and “redo”
functions (Nielsen, 1994)."

Tell me how, clicking in the div to expand and collapse a text (to read, no action is available) goes against it? @Ana06

vpereira:
superseded by #6914

PR #6901, opened 2019-02-03
Fix minor issues and improve user tab code

bgeuken:
* Add group icons to user & group tab
* Add 'send mail' links to groups table
* Prevent the invisible spinner from being rendered in the DOM. This was causing the table being re-rendered every 200ms. 

Before:

screenshot-2019-2-15 users of home admin - open build service 1

After:

screenshot-2019-2-15 users of home admin - open build service

### DIFF
--- a/src/api/app/views/webui2/shared/user_or_groups_roles/_list.html.haml
+++ b/src/api/app/views/webui2/shared/user_or_groups_roles/_list.html.haml
@@ -3,7 +3,7 @@
     mail_subject = "#{home_title} - #{project}"
     mail_subject += "/#{package}" if package
 - records.each do |record|
-  %tr{ id: "#{type}-#{valid_xml_id(record.to_s)}" }
+  %tr
     %td.align-middle
       = image_tag_for(record, size: 20)
       = link_to(display_name(record), user_or_group_show_path(record))
--- a/src/api/app/views/webui2/shared/user_or_groups_roles/_list.html.haml
+++ b/src/api/app/views/webui2/shared/user_or_groups_roles/_list.html.haml
@@ -5,8 +5,8 @@
 - records.each do |record|
   %tr{ id: "#{type}-#{valid_xml_id(record.to_s)}" }
     %td.align-middle
-      = user_with_realname_and_icon(record) if type == 'user'
-      = link_to(record.to_s, controller: 'groups', action: 'show', title: record.to_s) if type == 'group'
+      = image_tag_for(record, size: 20)
+      = link_to(display_name(record), user_or_group_show_path(record))
     - roles.each do |role|
       %td.align-middle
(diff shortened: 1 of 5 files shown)


saraycp:
@bgeuken is this finished? You still have "wip" and "test" commits. :thinking:

bgeuken:
@saraycp Sorry for the late reply. There are some failing tests that I have to fix (capybara doesn't properly handle the click event because there is a div that overlaps with the checkbox).
I put this on hold for now since it's just a side project. I will get back to it soon.

bgeuken:
@saraycp I found the issue. Would be great if you could review this when you have some time

src/api/app/views/webui2/shared/user_or_groups_roles/_list.html.haml
```diff
@@ -5,8 +5,8 @@
 - records.each do |record|
   %tr{ id: "#{type}-#{valid_xml_id(record.to_s)}" }
     %td.align-middle
-      = user_with_realname_and_icon(record) if type == 'user'
-      = link_to(record.to_s, controller: 'groups', action: 'show', title: record.to_s) if type == 'group'
+      = image_tag_for(record, size: 20)
+      = link_to(display_name(record), user_or_group_show_path(record))
```
saraycp: Do all groups have icons in production? :thinking:  otherwise it is going to fail..

src/api/app/views/webui2/shared/user_or_groups_roles/_list.html.haml
```diff
@@ -1,28 +1,27 @@
...
+    mail_subject = "#{home_title} - #{project}"
+    mail_subject += "/#{package}" if package
 - records.each do |record|
-  %tr{ id: "#{type}-#{valid_xml_id(record.to_s)}" }
+  %tr
     %td.align-middle
-      = user_with_realname_and_icon(record) if type == 'user'
-      = link_to(record.to_s, controller: 'groups', action: 'show', title: record.to_s) if type == 'group'
+      = image_tag_for(record, size: 20)
+      = link_to(display_name(record), user_or_group_show_path(record))
```
saraycp: Do all groups have icons in production? thinking otherwise it is going to fail..

bgeuken: We use an [identicon](https://en.wikipedia.org/wiki/Identicon) which is based on the email address. If the address is not set we [use](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/helpers/webui/webui_helper.rb#L415) a default icon.

saraycp: Ah! it's true! :+1:

PR #6902, opened 2019-02-04
Skip before_destroy of repository when we destroy a project

### LINKED ISSUE #3735: Project delete may modify project meta in backend before delete
Issue/Feature description
====================

Project restore may not restore the exact same project meta, if a repository builds against the same project. The restored project will point to "deleted" repository instead.

Expected result
============

Restore the same project meta, which existed when the project delete was executed..

How to Reproduce
===============
1. Create project A with repo B and C
2. Repo B need to have repo C in path
3. delete and undelete project.

Further information
===============
* This is caused by automatic store in Repository model, leading to a project meta PUT before DELETE operation

Example (The repository `home_Admin_project_abc_openSUSE_Leap_15.0` will be gone after `undelete`:
issue

DavidKang:
When we destroy a project we run the `before_action`
`cleanup_before_destroy`  from the project and also from the
repositories. Now when we destroy a project we skip the `before_action`
of the repositories.

Fix #3735

### DIFF
--- a/src/api/app/models/path_element.rb
+++ b/src/api/app/models/path_element.rb
@@ -1,7 +1,9 @@
 class PathElement < ApplicationRecord
+  # FIXME: This should be called parent
   belongs_to :repository, foreign_key: 'parent_id', inverse_of: :path_elements
   acts_as_list scope: [:parent_id]
 
+  # FIXME: This should be called repository
   belongs_to :link, class_name: 'Repository', foreign_key: 'repository_id', inverse_of: :links
 
   validates :link, :repository, presence: true
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -214,6 +214,8 @@ class Project < ApplicationRecord
 
     revoke_requests # Revoke all requests that have this project as source/target
     cleanup_packages # Deletes packages (only in DB)
+
+    repositories.each(&:mark_for_destruction)
   end
   private :cleanup_before_destroy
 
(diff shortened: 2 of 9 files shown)


src/api/app/models/path_element.rb
```diff
@@ -1,7 +1,9 @@
 class PathElement < ApplicationRecord
+  # FIXME: This should be called parent
```
Ana06: how are this FIXME related with the commit? :confused:

src/api/app/models/project.rb
```diff
@@ -332,31 +334,31 @@ def cleanup_linking_projects
...
   def cleanup_linking_targets
     # replace links to this projects with links to the "deleted" project
-    find_repos(:linking_target_repositories) do |link_rep|
-      link_rep.release_targets.includes(:target_repository).each do |rt|
-        next unless Repository.find(rt.repository_id).db_project_id == id
-        rt.target_repository = Repository.deleted_instance
-        rt.save
+    find_repos(:linking_target_repositories) do |linking_target_repository|
+      linking_target_repository.release_targets.includes(:target_repository).each do |release_target|
+        next unless release_target.link.db_project_id == id
```
Ana06: N+1 queries! :see_no_evil: (before as well, but now it is easier to fix....
you could even to this filtering in an SQL query in the previous line :bulb:

src/api/app/models/repository.rb
```diff
@@ -150,6 +150,7 @@ def cycles(arch)
   # or empty list
   def linking_repositories
     return [] if links.size.zero?
+    # FIXME: This is the same as using a `has_many through:` association
```
Ana06: [nitpick] not exactly... this returns an array...  but yes... it shouldn't be like that

dmarcoux: It's mostly to help us in the future (to hopefully refactor this)

src/api/app/models/path_element.rb
dmarcoux: It's mostly to help us in the future (to hopefully refactor this).

src/api/app/models/project.rb
dmarcoux: True about the N+1... we spent so much time digging and debugging while trying to understand this whole process, we didn't see it.

DavidKang:
@Ana06, @dmarcoux, could you review it again, please? :smile_cat:

dmarcoux:
I just had a look, it works. What did we forget before?

hennevogel:
@adrianschroeter can you have a look at the new test case?

src/api/app/models/repository.rb
Ana06: That is fine, but the comment could be more accurate :rofl: it is just a nitpick

src/api/app/models/path_element.rb
Ana06: That's fine, bit it is not related with the commit. IMO it should be in a different one, let's say it is a nitpick....

DavidKang:
@Ana06, the commit already has a description but I forget to add the issue :disappointed:

src/api/app/models/path_element.rb
bgeuken: It doesn't look like changing this would require a migration. So I would prefer to change the names instead of adding a fixme that will stay there forever.
And if there isn't time for this, let's open an issue. Maybe someone else wants to jump in.

src/api/app/models/repository.rb
bgeuken: Please just change this. We only use this at [one](https://github.com/openSUSE/open-build-service/search?q=linking_repositories&unscoped_q=linking_repositories). So this shouldn't take much time.

src/api/app/models/repository.rb
```diff
@@ -164,6 +165,7 @@ def is_local_channel?
 
   def linking_target_repositories
     return [] if targetlinks.size.zero?
+    # FIXME: This is the same as using a `has_many through:` association
```
bgeuken: same here

src/api/app/models/project.rb
```diff
@@ -332,31 +334,31 @@ def cleanup_linking_projects
...
   def cleanup_linking_targets
     # replace links to this projects with links to the "deleted" project
-    find_repos(:linking_target_repositories) do |link_rep|
-      link_rep.release_targets.includes(:target_repository).each do |rt|
-        next unless Repository.find(rt.repository_id).db_project_id == id
-        rt.target_repository = Repository.deleted_instance
-        rt.save
+    find_repos(:linking_target_repositories) do |linking_target_repository|
+      linking_target_repository.release_targets.includes(:target_repository, :link).each do |release_target|
+        next unless release_target.link.db_project_id == id
```
bgeuken: I think you could use a join here... Though I didn't check how much impact that would make here.

Ana06: a join? :confused:

PR #6904, opened 2019-02-04
Fix reviewable_by for by_package

coolo:
by_package always implies by_project as well - just the package name matching isn't good enough. See obs request 669918 where it was impossible to approve the sle-workerarounds reviews

CC @lnussel


lnussel:
when will this be live? blocks staging atm

coolo:
It is now (it wasn't me :)

hennevogel:
@lnussel and you can check [/about](https://build.opensuse.org/about) for deployed commit and last deploy time

PR #6907, opened 2019-02-04
Hide monitor table while loading it

Ana06:
To avoid the overflow until the DataTable with the scroll is loaded.

# Now:

image

# Before:

image

This is needed because loading the DataTable is slow, as it is creating two
tables and moving the content around. Another alternative could be to stop
using the scroll from DataTables and just add it to the table, but then the
package column would move as well when scrolling what makes the table unusable.
In this case, and taking into account that the page is only big (= slow) in few
cases, makes sense to prioritize ease of use over performance. The user waits a
bit, but at least he gets a nice table.

The double table that DataTable renders is also causing a bug (which disappears
when not using the scroll from DataTable). :bowtie:

### DIFF
--- a/src/api/app/assets/javascripts/webui2/project_monitor.js
+++ b/src/api/app/assets/javascripts/webui2/project_monitor.js
@@ -18,6 +18,13 @@ function setupProjectMonitor() { // jshint ignore:line
     }
   });
 
+  $('#table-spinner').addClass('d-none');
+  $('#project-monitor .obs-dataTable').removeClass('invisible');
+
+  $('#filter-button').on('click', function () {
+    $('#table-spinner').removeClass('d-none');
+  });
+
   $('#project-monitor-table').on('draw.dt', function () {
     setupPopover();
   });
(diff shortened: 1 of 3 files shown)


Ana06:
[Improve monitor firefox rendering](https://trello.com/c/qG8SLty6/1034-improve-monitor-firefox-rendering)

src/api/app/assets/stylesheets/webui2/datatables.scss
```diff
@@ -11,3 +11,7 @@ table.table-fixed {
   width: 100% !important;
   margin-left: 0 !important;
 }
+
+.slow-loading-table {
```
bgeuken: You could use bootstrap's [visibility classes](https://getbootstrap.com/docs/4.2/utilities/visibility/) instead and toogle the classes.

bgeuken: @Ana06 JFYI, I added this now to your commit since you are not here today. Hope that's okay:-)

PR #6916, opened 2019-02-05
Collapse/Expand the build results

dmarcoux:
@eduardoj and I tried two ways to achieve this.

We wanted to avoid rendering the partial on the server, then sending it to the client when a user refreshes the build results. This is the current state. We looked at multiple options and opted for [Mustache](https://mustache.github.io/) and gave it a fair try. However, it is simply so much changes/code to make this work and maintainable (especially this). It quickly became a mess and even though we would have liked for this to work, we decided to go with the second option.

So we are now with this solution as you can see in this PR. Instead of only sending the needed data to the client for it to be parsed/rendered with JavaScript, we still render the partial on the server and send it to the client. The changes are minimal, but the best would be to achieve what we aim for at first. It's something which we need to think about for `webui2`.

Anyway, here's a preview of how this works. Just before recording the GIF, I added the repository `openSUSE Leap 15.0`. I then collapse a repository and refresh the build results. The collapsed/expanded state is kept for all repositories and the build results are refreshed.

preview 

### DIFF
--- a/src/api/app/assets/javascripts/webui2/buildresult.js
+++ b/src/api/app/assets/javascripts/webui2/buildresult.js
@@ -18,6 +18,7 @@ function updateRpmlintResult(index) { // jshint ignore:line
 function updateBuildResult(index) { // jshint ignore:line
   var ajaxDataShow = $('#buildresult' + index + '-box').data();
   ajaxDataShow.show_all = $('#show_all_'+index).is(':checked'); // jshint ignore:line
+  ajaxDataShow.collapsedRepositories = $('.result div.collapse:not(.show)').map(function(_index, domElement) { return $(domElement).data('repository'); }).get();
   $('#build'+index+'-reload').addClass('fa-spin');
   $.ajax({
     url: $('#buildresult' + index + '-urls').data('buildresultUrl'),
--- a/src/api/app/assets/javascripts/webui2/buildresult.js
+++ b/src/api/app/assets/javascripts/webui2/buildresult.js
@@ -36,6 +36,15 @@ function updateBuildResult(index) { // jshint ignore:line
   });
 }
 
+function subscribeAjaxEvents(index) {
+  // TODO: Handle 'ajax:error'
+  $('#show_all_' + index).on('ajax:success', function(_event, data, _status, _xhr) {
+    $('#build' + index + ' .result').html(data);
+    // Since the element on which we bind the 'ajax:success' is rendered again on the line just above, it needs to be bound again...
+    subscribeAjaxEvents(index);
+  });
+}
+
 function updateArchDisplay(index) { // jshint ignore:line
   $('.rpmlint_arch_select_' + index).hide();
   $('#rpmlint_arch_select_' + index + '_' + $('#rpmlint_repo_select_' + index + ' option:selected').attr('value')).show();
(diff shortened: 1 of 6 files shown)


dmarcoux:
[Build Results](https://trello.com/c/rwZFAAsD/1032-build-results)

src/api/app/assets/javascripts/webui2/buildresult.js
```diff
@@ -35,6 +36,39 @@ function updateBuildResult(index) { // jshint ignore:line
   });
 }
 
+// TODO: Handle 'ajax:error', 'ajax:complete' and DRY binding code
+function subscribeAjaxEvents(index) {
+  $('#show_all_' + index).on('ajax:success', function(_event, data, _status, _xhr) {
+    $('#build' + index + ' .result').html(data);
+    // Since the element on which we bind this event is rendered again on the line just above, it needs to be bound again...
```
bgeuken: You can avoid this by attaching a generic event handler to the element that contains all the build results, e.g.
```js
$('.build-result-table').on('ajax:success', '.show-all', function() {
...
```

The index would get moved away to the data attribute of the `.show-all` element, if it's even needed. You probably could replace the `$('#build' + index + ' .result').html(data);` with something like `$(this).find('.build-result')` or `$(this).children`, sibling...

src/api/app/assets/javascripts/webui2/buildresult.js
```diff
@@ -35,6 +36,39 @@ function updateBuildResult(index) { // jshint ignore:line
   });
 }
 
+// TODO: Handle 'ajax:error', 'ajax:complete' and DRY binding code
+function subscribeAjaxEvents(index) {
+  $('#show_all_' + index).on('ajax:success', function(_event, data, _status, _xhr) {
```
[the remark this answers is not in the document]
dmarcoux: I know, but if we ever need to change this code, we don't need to check the documentation to see which parameters are available or not.

src/api/app/assets/javascripts/webui2/buildresult.js
```diff
@@ -35,6 +36,39 @@ function updateBuildResult(index) { // jshint ignore:line
...
+  $('button.build-refresh').on('ajax:before', function(_event) {
+    var button = $(this);
+    var buttonParams = button.data('params');
+    console.log(buttonParams);
+    // var serializedCollapsedRepositories = $('.result div.collapse:not(.show)').map(function(_index, domElement) { return $.param((({ repository }) => ({ repository }))($(domElement).data())); }).get().join('&');
+    // button.data('params', buttonParams + '&' + serializedCollapsedRepositories);
+    console.log(button.data);
+  });
+
+  // $('button.build-refresh').on('ajax:beforeSend', function(_event, _xhr, settings) {
```
vpereira: remove the commented out code + console.log

src/api/app/assets/javascripts/webui2/buildresult.js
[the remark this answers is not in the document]
dmarcoux: In the end, I won't be using this function at all.

src/api/app/assets/javascripts/webui2/buildresult.js
dmarcoux: In the end, I won't be using this function at all.

dmarcoux:
I'm closing this PR and will submit a new one when I got the UJS part done. It will just be a mess otherwise while I get this done.

PR #6917, opened 2019-02-05
Simplify updating request attributes

bgeuken:
and ensure that an exception is thrown if updating it fails. So
that we get aware of issues in our tests.


dmarcoux:
@bgeuken ping

saraycp:
LGTM again :smile:  Good idea to replace `update_attributes!` by `update_attributes` in order to avoid overwritten `BsRequest#save!` (which calls `sanitize!`) :+1:

PR #6921, opened 2019-02-06
Move check_ajax from sourcediff to a before_action

vpereira:
```check_ajax``` should be called in a before_action, since its not direct connected with the action ```sourcediff``` logic


saraycp:
@bgeuken @vpereira 

> looks like we have the same pattern in a couple of controllers. Would be nice if you could update them as well

I merge this PR and create a new one to update the rest.

saraycp:
New PR: #6923

PR #6922, opened 2019-02-06
[api] support releasing into maintenance_release without incident pro…

adrianschroeter:
…ject

Using readable timestamp as fallback.

Used in continous rebuild projects for container releases

consolidate mechanics of entire project and single package release
handling, fixing small unwanted differences

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -997,7 +997,7 @@ class SourceController < ApplicationController
     end
 
     # uniq timestring for all targets
-    time_string = Time.now.utc.strftime('%Y%m%d%H%M%S')
+    time_now = Time.now.utc
 
     # specified target
     if params[:target_project]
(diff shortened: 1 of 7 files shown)


src/api/app/controllers/source_controller.rb
```diff
@@ -996,6 +996,9 @@ def package_command_release
       multibuild_container = params[:package].gsub(/^.*:/, '')
     end
 
+    # uniq timestring for all targets
+    time_now = Time.now.utc
```
Ana06: why are you assigning it here if you only use it inside release_target_name in one line? :confused:

adrianschroeter:
@vpereira can you do me a favour and regenerate the cassettes?

(the spec suite is not working for me, neither plain nor via docker and we block opensuse images atm)

vpereira:
> (the spec suite is not working for me, neither plain nor via docker and we block opensuse images atm)

Hi @adrianschroeter!

I can do it, but maybe because you renamed some methods in the model, just changing the related specs (calling the new method names) will fix it?

adrianschroeter:
@vpereira  I would appriciate if you can do it. 

It is not pure renaming, it consolidates two methods into one.

Raising importance, since some really old version of this has been merged which caused some trouble.

src/api/app/controllers/source_controller.rb
vpereira: I think it's inside a loop iteration and because he wants to have all release targets with the same time, IMO is ok. Am I right @adrianschroeter ?

adrianschroeter:
On Freitag, 1. März 2019, 15:20:48 CET Victor Pereira wrote:
> vpereira commented on this pull request.
> 
> > @@ -996,6 +996,9 @@ def package_command_release
> 
>        multibuild_container = params[:package].gsub(/^.*:/, '')
>      end
> 
> +    # uniq timestring for all targets
> +    time_now = Time.now.utc
> 
> I think it's inside a loop iteration and because he wants to have all
> release targets with the same time, IMO is ok. Am I right @adrianschroeter
> ?

yes

-- 

Adrian Schroeter
email: adrian@suse.de

PR #6929, opened 2019-02-07
Only show request actions the user has permissions for

bgeuken:
When porting this to the new webui I accidentally ignored
the @is_target_maintainer variable.
This adds it to the new view now, which fixes that the
forward request checkbox and accept request box is shown
for users that don't have the required permissions.

### DIFF
--- a/src/api/app/views/webui2/webui/request/_decision_tab.html.haml
+++ b/src/api/app/views/webui2/webui/request/_decision_tab.html.haml
@@ -2,7 +2,7 @@
   = hidden_field_tag(:number, request_number)
   %p
     = text_area_tag(:reason, nil, placeholder: 'Please add a comment', rows: 4, class: 'w-100 form-control')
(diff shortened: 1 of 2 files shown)


bgeuken:
Seems it's time to enable the old tests for the new UI and port them over:/

src/api/app/views/webui2/webui/request/_decision_tab.html.haml
```diff
@@ -2,7 +2,7 @@
   = hidden_field_tag(:number, request_number)
   %p
     = text_area_tag(:reason, nil, placeholder: 'Please add a comment', rows: 4, class: 'w-100 form-control')
-  - if single_action_request
+  - if single_action_request && is_target_maintainer && state.in?(['new', 'review'])
```
hennevogel: Are we missing a policy?

PR #6940, opened 2019-02-08
Pass the package name as a parameter instead of the package object

### LINKED ISSUE #6847: Download link broken/unaivailable for _multibuild
In openSUSE:Factory:ARM:Live I am not able to download mutlibuild images.
It is the same as old closed issue https://github.com/openSUSE/open-build-service/issues/4985 apparently.

Expected Result
============
Download link should just work.

How to Reproduce
===============
Go to https://build.opensuse.org/package/binaries/openSUSE:Factory:ARM:Live/JeOS:JeOS-beagle/images and try to download the *.raw.xz file.
You get a 404 error.

Further Information
===============
* A link to an OBS instance showing the issue: https://build.opensuse.org/package/binaries/openSUSE:Factory:ARM:Live/JeOS:JeOS-beagle

eduardoj:
This way, in the case of multibuild, the url for downloading the binary contains the package and the build section, as needed.

Fix #6847.

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -1040,11 +1040,12 @@ class Webui::PackageController < Webui::WebuiController
   def edit; end
 
   def binary_download
+    package_name = params[:package]
     architecture = Architecture.find_by_name(params[:arch]).name
     filename = File.basename(params[:filename]) # Ensure it really is just a file name, no '/..', etc.
     repository = Repository.find_by_project_and_name(@project.to_s, params[:repository].to_s)
 
-    url_generator = ::PackageControllerService::URLGenerator.new(project: @project, package: @package,
+    url_generator = ::PackageControllerService::URLGenerator.new(project: @project, package: package_name,
                                                                  user: User.current, arch: architecture,
                                                                  repository: repository, filename: filename)
 


src/api/app/controllers/webui/package_controller.rb
```diff
@@ -1044,7 +1044,7 @@ def binary_download
     filename = File.basename(params[:filename]) # Ensure it really is just a file name, no '/..', etc.
     repository = Repository.find_by_project_and_name(@project.to_s, params[:repository].to_s)
 
-    url_generator = ::PackageControllerService::URLGenerator.new(project: @project, package: @package,
+    url_generator = ::PackageControllerService::URLGenerator.new(project: @project, package: @package_name,
```
DavidKang: I think you forgot to define `@package_name` with `@package_name = params[:package]`

eduardoj: Done. Is the `@` in `@package_name` needed?

DavidKang: nop, is not needed

eduardoj:
[Go! Go! Go! Buuuuuug Squad!!!!](https://trello.com/c/AnXFh9ob/1038-go-go-go-buuuuuug-squad)

PR #6945, opened 2019-02-08
Request Feature Tests

DavidKang:
Because of the migration to bootstrap we copy and adapted the feature
tests for request.

We need to create other PRs to extend the tests for request.

### DIFF
--- a/src/api/spec/bootstrap/features/webui/requests_spec.rb
+++ b/src/api/spec/bootstrap/features/webui/requests_spec.rb
@@ -5,23 +5,204 @@ RSpec.feature 'Bootstrap_Requests', type: :feature, js: true, vcr: true do
   let(:receiver) { create(:confirmed_user, login: 'titan') }
   let(:target_project) { receiver.home_project }
   let(:target_package) { create(:package, name: 'goal', project_id: target_project.id) }
+  let(:source_project) { submitter.home_project }
+  let(:source_package) { create(:package, name: 'ball', project_id: source_project.id) }
+  let(:bs_request) { create(:delete_bs_request, target_project: target_project, description: 'a long text - ' * 200, creator: submitter) }
+
+  RSpec.shared_examples 'expandable element' do
+    scenario 'expanding a text field' do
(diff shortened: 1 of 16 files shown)


src/api/spec/bootstrap/features/webui/requests_spec.rb
```diff
@@ -5,23 +5,204 @@
...
   let(:target_package) { create(:package, name: 'goal', project_id: target_project.id) }
+  let(:source_project) { submitter.home_project }
+  let(:source_package) { create(:package, name: 'ball', project_id: source_project.id) }
+  let(:bs_request) { create(:delete_bs_request, target_project: target_project, description: 'a long text - ' * 200, creator: submitter) }
+
+  RSpec.shared_examples 'expandable element' do
+    scenario 'expanding a text field' do
+      visit request_show_path(bs_request)
+      within(element) do
+        # FIXME: find a better way to check if the content is collapsed or not
```
bgeuken: I don't have the collapsable code in mind that we use here, but things that might work:
* Check for the CSS classes that apply the collapsed / uncollapsed state
* Use the visible option in the capybara finder. There is also a [visible?](https://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FElement:visible%3F) method that might work here.

dmarcoux: A collapsible element will always have the CSS class `collapse` and it won't have the CSS class `show` when collapsed. We could check if an element is collapsed if it has the CSS class `show`, something like this:

```
# Expanded element
expect(find(id: 'req-description')).to have_css('show')

# Collapsed element
expect(find(id: 'req-description')).not_to have_css('show')
```

bgeuken: Right, but IIRC we use a custom "collapsible" here

dmarcoux: Oh, then it doesn't apply.

PR #6947, opened 2019-02-08
Send emails as user nobody

### LINKED ISSUE #5351: [production][] undefined method `can_modify?' for nil:NilClass
threw this 6 times
<pre>
undefined method `can_modify?&#39; for nil:NilClass
</pre>

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5b4127641e40e80006f13a2e)

eduardoj:
In `source_access_check!` method in `BsRequestAction` a user is required to know if he has access to the source.

Fix #5351.

### DIFF
--- a/src/api/app/jobs/send_event_emails_job.rb
+++ b/src/api/app/jobs/send_event_emails_job.rb
@@ -2,7 +2,8 @@ class SendEventEmailsJob < ApplicationJob
   queue_as :mailers
 
   def perform
-    User.current = User.find_nobody!
+    # previous_user = User.current
+    # User.current = User.find_nobody!
     Event::Base.where(mails_sent: false).order(created_at: :asc).limit(1000).each do |event|
       subscribers = event.subscribers
 
@@ -20,6 +21,7 @@ class SendEventEmailsJob < ApplicationJob
         event.update_attributes(mails_sent: true)
       end
     end
+    # User.current = previous_user
     true
   end
 
--- a/src/api/app/jobs/send_event_emails_job.rb
+++ b/src/api/app/jobs/send_event_emails_job.rb
@@ -2,6 +2,7 @@ class SendEventEmailsJob < ApplicationJob
   queue_as :mailers
 
   def perform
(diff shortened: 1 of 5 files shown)


src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -2,6 +2,7 @@ class SendEventEmailsJob < ApplicationJob
   queue_as :mailers
 
   def perform
+    User.current = User.find_nobody!
```
Ana06: [nitpick] I think it would be better to consider the case the user is `nil` or ensure the user is always set instead of assigning it all over the code.... but it may be too late for this already... :sob:

eduardoj: > [nitpick] I think it would be better to consider the case the user is `nil` or ensure the user is always set instead of assigning it all over the code....

Do you mean to ensure the user is always set for all the jobs?

Can we ensure the user here in this job, and when needed, add it in all other jobs in other pull requests,... and then refactor the code if also needed?

hennevogel: > always set instead of assigning it all over the code.... but it may be too late for this already...

You are a couple of years to late. Authentication, session handling and authorization needs some serious refactoring. But not here.

Ana06: @hennevogel 

> You are a couple of years to late. Authentication, session handling and authorization needs some serious refactoring. But not here.

That is exactly what I meant....

eduardoj:
Add "DO NOT MERGE", as one last test is still failing.

Ana06:
@eduardoj 

> Add "DO NOT MERGE", as one last test is still failing.

we don't merge PR with failing tests.... :eyes:

coolo:
I say this and 5351 can be closed as 7168 was merged

eduardoj:
> I say this and 5351 can be closed as 7168 was merged

Closing this PR then.

PR #6965, opened 2019-02-12
[api] support rebuild and release triggers

adrianschroeter:
required for CR releases by automated testing and for openSUSE ports
co-maintainers

### DIFF
--- a/src/api/app/controllers/person/token_controller.rb
+++ b/src/api/app/controllers/person/token_controller.rb
@@ -7,25 +7,25 @@ module Person
     def index
       authorize @user, :show?
 
-      @list = @user.service_tokens
+      @list = @user.tokens
     end
 
     # POST /person/<login>/token
     def create
       authorize @user, :update?
 
-      pkg = nil
-      if params[:project] || params[:package]
-        pkg = Package.get_by_project_and_name(params[:project], params[:package])
-      end
-      @token = Token::Service.create(user: @user, package: pkg)
+      pkg = if params[:project] || params[:package]
+              Package.get_by_project_and_name(params[:project], params[:package])
+            end
+
+      @token = Token.token_type(params[:operation]).create(user: @user, package: pkg)
     end
 
     # DELETE /person/<login>/token/<id>
     def delete
       authorize @user, :update?
 
-      token = @user.service_tokens.find(params[:id])
+      token = @user.tokens.find(params[:id])
 
       render_error status: 404 && return unless token
 
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -1,53 +1,85 @@
 class TriggerController < ApplicationController
+  validate_action rebuild: { method: :post, response: :status }
+  validate_action release: { method: :post, response: :status }
   validate_action runservice: { method: :post, response: :status }
 
+  before_action :require_project_param, only: [:release]
+
+  before_action :extract_auth_from_request, :validate_auth_token, :require_valid_token
   #
-  # This controller is checking permission always only on the base of tokens
+  # Authentication happens with tokens, so no login is required
   #
   skip_before_action :extract_user
   skip_before_action :require_login
 
-  # github.com sends a hash payload
-  skip_before_action :validate_params, only: [:runservice]
+  # to get access to the method release_package
+  include MaintenanceHelper
 
-  def runservice
-    auth = request.env['HTTP_AUTHORIZATION']
+  include Trigger::Errors
 
-    if request.env['HTTP_X_GITLAB_EVENT'] == 'Push Hook'
-      auth = 'Token ' + request.env['HTTP_X_GITLAB_TOKEN']
-    end
+  def rebuild
+    Backend::Api::Sources::Package.rebuild(@pkg.project.name, @pkg.name)
+    render_ok
+  end
 
-    unless auth && auth[0..4] == 'Token' && auth[6..-1] =~ /^[A-Za-z0-9+\/]+$/
-      render_error errorcode: 'permission_denied',
-                   message: "No valid token found 'Authorization' header",
-                   status: 403
-      return
-    end
+  def release
+    matched_repo = false
+    @pkg.project.repositories.includes(:release_targets).each do |repo|
+      repo.release_targets.where(trigger: 'manual').each do |releasetarget|
+        release_target_repository_project = releasetarget.target_repository.project
+        unless policy(release_target_repository_project).update?
+          raise NoPermissionForTarget.setup('no_permission',
+                                            403, "no permission for target #{release_target_repository_project}")
+        end
+        target_package_name = @pkg.release_target_name
 
-    token = Token::Service.find_by_string(auth[6..-1])
-
-    unless token
-      render_error message: 'Token not found', status: 404
-      return
-    end
-
-    pkg = token.package || Package.get_by_project_and_name(params[:project].to_s, params[:package].to_s, use_source: true)
-    if pkg
-      # check if user has still access
-      unless token.user.is_active? && token.user.can_modify?(pkg)
-        render_error message: "no permission for package #{pkg.name} in project #{pkg.project.name}",
-                     status: 403,
-                     errorcode: 'no_permission'
-        return
+        # find md5sum and release source and binaries
+        release_package(@pkg, releasetarget.target_repository, target_package_name, repo, nil, nil, nil, true)
+        matched_repo = true
       end
     end
 
+    raise NoPermissionForPackage.setup('not_found', 404, "no repository from #{@pkg.project} could get released") unless matched_repo
+
+    render_ok
+  end
+
+  def runservice
     # execute the service in backend
-    path = pkg.source_path
-    params = { cmd: 'runservice', comment: 'runservice via trigger', user: token.user.login }
+    path = @pkg.source_path
+    params = { cmd: 'runservice', comment: 'runservice via trigger', user: User.session!.login }
     path << build_query_from_hash(params, [:cmd, :comment, :user])
     pass_to_backend(path)
 
-    pkg.sources_changed
+    @pkg.sources_changed
+  end
+
+  private
+
+  def require_project_param
+    render_error(message: 'Token must define the release package', status: 403, errorcode: 'no_permission') if params[:project].present?
+  end
+
+  def extract_auth_from_request
+    @token_extractor = ::TriggerControllerService::TokenExtractor.new(request).call
+  end
+
+  def validate_auth_token
+    raise InvalidToken unless @token_extractor.valid?
+  end
+
+  def require_valid_token
+    @token = @token_extractor.token
+
+    raise TokenNotFound unless @token
+
+    User.session = @token.user
+
+    raise NoPermissionForInactive unless User.session.is_active?
+
+    @pkg = @token.package || Package.get_by_project_and_name(params[:project].to_s, params[:package].to_s, use_source: true)
+
+    raise ActiveRecord::RecordNotFound unless @pkg
+    raise NoPermissionForPackage.setup('no_permission', 403, "no permission for package #{@pkg} in project #{@pkg.project}") unless policy(@pkg).update?
   end
 end
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -168,6 +168,7 @@ POST /person/<userid>/token
   cmd: create
   project: create a token for a specific project
   package: create a token for a specific package
+  operation: runservice(default), rebuild, release
 
 XmlResult: status
 
--- a/src/api/app/controllers/trigger/errors.rb
+++ b/src/api/app/controllers/trigger/errors.rb
@@ -1,0 +1,26 @@
+module Trigger::Errors
+  extend ActiveSupport::Concern
+
+  class NoPermissionForInactive < APIError
+    setup 403, 'no permission due to inactive user'
+  end
+
+  class TokenNotFound < APIError
+    setup 404, 'Token not found'
+  end
+
+  class InvalidToken < APIError
+    setup 'permission_denied',
+          403,
+          'No valid token found "Authorization" header'
+  end
+
+  class NoPermissionForPackage < APIError
+  end
+
+  class NoRepositoryCouldBeReleased < APIError
+  end
+
+  class NoPermissionForTarget < APIError
+  end
+end
--- a/src/api/app/models/token/rebuild.rb
+++ b/src/api/app/models/token/rebuild.rb
@@ -1,25 +1,6 @@
-class Token < ApplicationRecord
-  belongs_to :user
-  belongs_to :package, inverse_of: :tokens
-
-  has_secure_token :string
-
-  validates :user, presence: true
-
-  def token_name
-    self.class.token_name
-  end
-
-  def self.token_type(action)
-    case action
-    when 'rebuild'
-      Token::Rebuild
-    when 'release'
-      Token::Release
-    else
-      # default is Token::Service
-      Token::Service
-    end
+class Token::Rebuild < Token
+  def self.token_name
+    'rebuild'
   end
 end
 
--- a/src/api/app/models/token/release.rb
+++ b/src/api/app/models/token/release.rb
@@ -1,25 +1,6 @@
-class Token < ApplicationRecord
-  belongs_to :user
-  belongs_to :package, inverse_of: :tokens
-
-  has_secure_token :string
-
-  validates :user, presence: true
-
-  def token_name
-    self.class.token_name
-  end
-
-  def self.token_type(action)
-    case action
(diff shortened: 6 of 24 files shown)


src/api/app/controllers/trigger_controller.rb
```diff
@@ -1,16 +1,71 @@
...
   skip_before_action :require_login
 
-  # github.com sends a hash payload
-  skip_before_action :validate_params, only: [:runservice]
+  def rebuild
+    Backend::Api::Sources::Package.rebuild(@pkg.project.name, @pkg.name)
+    render_ok
+  end
+
+  # rubocop:disable Lint/NonLocalExitFromIterator
```
Ana06: why? :sob:

src/api/app/controllers/trigger_controller.rb
```diff
@@ -1,16 +1,71 @@
...
+      render_error message: 'Token must define the release package',
+                   status: 403,
+                   errorcode: 'no_permission'
+      return
+    end
+
+    matched_repo = false
+    User.current = @token.user
+    @pkg.project.repositories.each do |repo|
+      repo.release_targets.each do |releasetarget|
```
Ana06: N+1 queries! :see_no_evil: (performance improvement)

src/api/app/controllers/trigger_controller.rb
```diff
@@ -1,16 +1,71 @@
...
+        next unless releasetarget.trigger == 'manual'
```
Ana06: You can add this to the SQL query :wink: (performance improvement)

src/api/app/controllers/trigger_controller.rb
```diff
@@ -1,16 +1,71 @@
...
+        unless @token.user.can_modify?(releasetarget.repository.project)
```
Ana06: `repo.project`

You are using the project several time (some of them as `@pkg.project`). I think it is a good idea to assign the project to a variable before the loop :wink:

src/api/app/controllers/trigger_controller.rb
```diff
@@ -1,16 +1,71 @@
...
+          render_error message: "no permission for target #{releasetarget.repository.project.name}",
+                       status: 403,
+                       errorcode: 'no_permission'
+          return
+        end
+        target_package_name = @pkg.release_target_name
```
Ana06: This is the same every time you iterate, you should put it outside the iteration (small performance improvement)

src/api/app/controllers/trigger_controller.rb
```diff
@@ -1,16 +1,71 @@
...
+
+        # find md5sum and release source and binaries
+        release_package(@pkg, releasetarget.target_repository, target_package_name, repo, nil, nil, nil, true)
+        matched_repo = true
+      end
+    end
+    unless matched_repo
+      render_error message: "no repository from #{@pkg.project.name} could get released",
```
Ana06: [nitpick] `.name` is not needed

src/api/app/controllers/person/token_controller.rb
```diff
@@ -18,14 +18,16 @@ def create
       if params[:project] || params[:package]
         pkg = Package.get_by_project_and_name(params[:project], params[:package])
       end
-      @token = Token::Service.create(user: @user, package: pkg)
+      @token = Token::Service.create(user: @user, package: pkg) if params[:operation].blank? || params[:operation] == 'runservice'
```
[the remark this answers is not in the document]
Ana06: [nitpick] And it would be a good idea to move the `token_type` part to the token model as it is also used in https://github.com/openSUSE/open-build-service/pull/6965/files#diff-00232d8e1ee1f7438e6cbb963c74a038R82 and then it could be reused :smile:

src/api/app/controllers/trigger_controller.rb
```diff
@@ -19,35 +74,32 @@ def runservice
...
     unless auth && auth[0..4] == 'Token' && auth[6..-1] =~ /^[A-Za-z0-9+\/]+$/
       render_error errorcode: 'permission_denied',
-                   message: "No valid token found 'Authorization' header",
+                   message: 'No valid token found "Authorization" header',
                    status: 403
       return
     end
 
-    token = Token::Service.find_by_string(auth[6..-1])
+    token_type = Token::Service
```
Ana06: [nitpick] this is exactly was `case` is for :wink: 

```ruby
token_type = case request['action']
             when 'rebuild'
               Token::Rebuild
             when 'release'
               Token::Release
             else
               Token::Service
             end

File: src/api/app/controllers/trigger_controller.rb [hunk 9]
```diff
@@ -19,35 +74,32 @@ def runservice
...
+    return unless @pkg
 
-    # execute the service in backend
-    path = pkg.source_path
-    params = { cmd: 'runservice', comment: 'runservice via trigger', user: token.user.login }
-    path << build_query_from_hash(params, [:cmd, :comment, :user])
-    pass_to_backend(path)
+    # check if user has still access
+    unless @token.user.is_active? && @token.user.can_modify?(@pkg)
+      render_error message: "no permission for package #{@pkg.name} in project #{@pkg.project.name}",
```
Ana06: [nitpick] `.name` is not needed

src/api/app/models/token.rb
```diff
@@ -1,10 +1,14 @@
 class Token < ApplicationRecord
-  belongs_to :user, foreign_key: 'user_id', inverse_of: :service_tokens
+  belongs_to :user, foreign_key: 'user_id'
```
Ana06: why is `foreign_key` needed? :thinking:

src/api/app/models/token.rb
```diff
@@ -1,10 +1,14 @@
...
   belongs_to :package, inverse_of: :tokens
 
   has_secure_token :string
 
   validates :user, presence: true
+
+  def token_name
```
Ana06: why don't you just define it as an instance method in the subclass? :confused:

src/api/app/controllers/trigger_controller.rb
adrianschroeter: because we must escape from the loops with proper error message and don't want to have additional checking code just to get away with break.

src/api/app/controllers/trigger_controller.rb
adrianschroeter: this is really absolute minor here ... these are release jobs, they are expensive....

src/api/app/controllers/trigger_controller.rb
Ana06: ok

src/api/app/controllers/trigger_controller.rb
Ana06: well, this is raising a lot of extra queries and it is easy to solve:

```ruby
@pkg.project.repositories.includes(:release_targets).each do |repo|

File: src/api/app/models/token.rb (hunk 11 above)
Comment (Ana06) [reply src/api/app/models/token.rb]: I am not sure if it clear what I meant. You could define for example in `rebuild.rb`:

```
def token_name
  'rebuild'
end
```

without the `self`

**Comment by hennevogel** [comment]
@adrianschroeter can we help you to push this over the line somehow?

File: src/api/app/models/token.rb (hunk 11 above)
Comment (Ana06) [reply src/api/app/models/token.rb]: ^^

File: src/api/app/controllers/trigger_controller.rb (hunk 5 above)
Comment (Ana06) [reply src/api/app/controllers/trigger_controller.rb]: ^^

Comment (Ana06) [reply src/api/app/controllers/trigger_controller.rb]: This is the same every time you iterate, you should put it outside the iteration (small performance improvement)

File: src/api/app/models/token.rb (hunk 11 above)
Comment (Ana06) [reply src/api/app/models/token.rb]: his could be an instance method of the subclass. Any reason to do it like this needing to do `self.class.token_name`?

File: src/api/app/controllers/person/token_controller.rb [hunk 12]
```diff
@@ -7,25 +7,34 @@ class TokenController < ApplicationController
...
+
+      token_type = case params[:operation]
+                   when 'rebuild'
+                     Token::Rebuild
+                   when 'release'
+                     Token::Release
+                   else
+                     # default is Token::Service
+                     Token::Service
+                   end
```
bgeuken: This looks like it could be dried with the code [here](https://github.com/openSUSE/open-build-service/pull/6965/files#diff-00232d8e1ee1f7438e6cbb963c74a038R83)

vpereira: yup. Did it now.

src/api/app/controllers/person/token_controller.rb
```diff
@@ -7,25 +7,25 @@ class TokenController < ApplicationController
...
     # POST /person/<login>/token
     def create
       authorize @user, :update?
 
-      pkg = nil
-      if params[:project] || params[:package]
-        pkg = Package.get_by_project_and_name(params[:project], params[:package])
-      end
-      @token = Token::Service.create(user: @user, package: pkg)
+      pkg = if params[:project] || params[:package]
```
Ana06: [nitpick] this refactoring is unrelated

src/api/app/controllers/trigger_controller.rb
Ana06: this haven't been addressed

src/api/app/models/token.rb
Ana06: this  is also not addressed...

bgeuken: IIRC, this could also be a delegate to the class, but I think that's more of a nitpick

src/api/app/models/token.rb
```diff
@@ -1,10 +1,26 @@
...
+    self.class.token_name
+  end
+
+  def self.token_type(action)
```
Ana06: this can be returned directly in the methods.

src/api/app/controllers/trigger_controller.rb
```diff
@@ -1,53 +1,96 @@
...
 
-  def runservice
-    auth = request.env['HTTP_AUTHORIZATION']
+  # rubocop:disable Lint/NonLocalExitFromIterator
+  def release
+    matched_repo = false
+    User.session = @token.user
+    @pkg.project.repositories.includes(:release_targets).each do |repo|
+      repo.release_targets.where(trigger: 'manual').each do |releasetarget|
+        unless @token.user.can_modify?(releasetarget.repository.project)
```
bgeuken: If I am not mistaken, we want to check the `target_repository` of the release target instead: `releasetarget.target_repository.project`

adrianschroeter: right, this is indeed wrong

src/api/app/controllers/trigger_controller.rb
```diff
@@ -1,53 +1,96 @@
...
+    Backend::Api::Sources::Package.rebuild(@pkg.project.name, @pkg.name)
+    render_ok
+  end
 
-  def runservice
-    auth = request.env['HTTP_AUTHORIZATION']
+  # rubocop:disable Lint/NonLocalExitFromIterator
+  def release
+    matched_repo = false
+    User.session = @token.user
```
adrianschroeter: IMHO we should not do this, because any programming error or refactoring may lead to a security issue.
The token should only allow a certain operation, but we should prepare for side effects.

The problem coolo reported is actually a generic problem that we always try to write the package meta when using Package.source_changed. source_changed was supposed just to update database content which could get replayed from the source server.
However, this is beyond the topic of this pull request and a generic problem atm.

src/api/app/controllers/trigger_controller.rb
vpereira: @adrian 

so i have to check the permission on `releasetarget.target_repository.project` ?

what else is wrong? the call to `release_package` is right or there is something wrong there too?

src/api/app/controllers/trigger_controller.rb
vpereira: I think the issue isn't about the controller, but operations on models and backend that need an user.. I think thats horrible too, but I'm not sure we are able to avoid it now.

coolo: create a backlog item and continue :(

adrianschroeter:
On Freitag, 17. Mai 2019, 17:00:51 CEST Victor Pereira wrote:
> vpereira commented on this pull request.
> 
> 
> 
> 
> -  def runservice
> -    auth = request.env['HTTP_AUTHORIZATION']
> +  # rubocop:disable Lint/NonLocalExitFromIterator
> +  def release
> +    matched_repo = false
> +    User.session = @token.user
> +    @pkg.project.repositories.includes(:release_targets).each do |repo|
> +      repo.release_targets.where(trigger: 'manual').each do |releasetarget|
> +        unless @token.user.can_modify?(releasetarget.repository.project)
> 
> @adrian
> 
> so i have to check the permission on
> `releasetarget.target_repository.project` ?

yes! Just replace that line to

  unless @token.user.can_modify?(releasetarget.target_repository.project)

> what else is wrong? the call to `release_package` is right or there is
> something wrong there too?

should be fine and worked in the test suite stub from original
submission at least

-- 

Adrian Schroeter
email: adrian@suse.de

adrianschroeter:
On Freitag, 17. Mai 2019, 17:02:02 CEST Victor Pereira wrote:
> vpereira commented on this pull request.
> 
> 
> 
> 
> -  def runservice
> -    auth = request.env['HTTP_AUTHORIZATION']
> +  # rubocop:disable Lint/NonLocalExitFromIterator
> +  def release
> +    matched_repo = false
> +    User.session = @token.user
> 
> I think the issue isn't about the controller, but operations on models and
> backend that need an user.. I think thats horrible too, but I'm not sure we
> are able to avoid it now.

I agree that this is an independend problem. This pull request is not
changing anything here.


-- 

Adrian Schroeter
email: adrian@suse.de

src/api/app/controllers/trigger_controller.rb
vpereira: this is resolved, right?

src/api/app/models/token.rb
vpereira: we didn't want to change it before we have specs for that.

src/api/app/controllers/person/token_controller.rb
vpereira: yes, since we were refactoring other things, this nitpick isn't really relevant.

src/api/app/controllers/trigger_controller.rb
vpereira: AFAIR in your original change request, you handled it as nitpick. So, because Adrian didn't want to change it, I will keep it too (probably because the number of release_targets where trigger = manual won't end up in thousand lines iteration)

src/api/app/models/token.rb
vpereira: I'm not sure what would be the difference here, but because it was something that Adrian did, and I don't see why we should change it now, I would keep it (unless there is a strong argument to change it)

src/api/app/models/token.rb
vpereira: @Ana06 could be more specific, please?

src/api/app/controllers/trigger_controller.rb
vpereira: because it's a nitpick, I would let as it is, since the number of matches is really specific, I would keep it as @adrianschroeter did it and optimize it if necessary later... what do you think?

vpereira:
> There are still not two addressed changes. Apart from that LGTM

The two not addressed change requests are, as you wrote, nitpick, and I wouldn't change it now unless there is a strong reason to do it.

vpereira:
> I wrote two comments that are not nitpick:
> 
> [#6965 (comment)](https://github.com/openSUSE/open-build-service/pull/6965#discussion_r283311990)
> 
> [#6965 (comment)](https://github.com/openSUSE/open-build-service/pull/6965#discussion_r256834313)
> 
> But if it is needed to merge this, go ahead. I can send a PR afterwards to improve it.

great, if you want later when you are refactoring it, you can ping me and we can do it together. Thank you @Ana06 !

src/api/app/controllers/trigger_controller.rb
```diff
@@ -1,53 +1,85 @@
...
+    Backend::Api::Sources::Package.rebuild(@pkg.project.name, @pkg.name)
+    render_ok
+  end
 
-    token = Token::Service.find_by_string(auth[6..-1])
+  def release
+    matched_repo = false
+    @pkg.project.repositories.includes(:release_targets).each do |repo|
+      repo.release_targets.where(trigger: 'manual').each do |releasetarget|
+        release_target_repository_project = releasetarget.repository.project
```
bgeuken: This should be `releasetarget.target_repository.project`. releasetarget.repository would be the `repo` we are coming from

vpereira: could please take a look again @bgeuken ?

PR #6967, opened 2019-02-12
Add alert box for superseded request

dmarcoux:
My only concern with this is that we are repeating the information. As you can see in the screenshot below, we now display twice that the request is superseded. It is once in the alert box and then, once in the side links of the request. I believe this is something we should address when reworking this page more in depth (instead of only migrating and fixing major issues as we agreed upon).

preview

### DIFF
--- a/src/api/app/views/webui2/webui/request/_superseded_by_message.html.haml
+++ b/src/api/app/views/webui2/webui/request/_superseded_by_message.html.haml
@@ -1,0 +1,24 @@
+- if superseding.present? && superseded_by.nil? && diff_to_superseded.nil?
+  .alert.alert-secondary{ role: 'alert' }
+    %i.fa.fa-info-circle.text-info
+    This request supersedes:
+    - superseding.each do |bs_request|
+      = link_to("request #{bs_request.number}", request_show_path(number: bs_request.number))
+      = surround('(', ")#{', ' unless bs_request == superseding.last}") do
+        = link_to('Show diff', request_show_path(number: bs_request.superseded_by, diff_to_superseded: bs_request.number))
+- if superseded_by.present?
+  .alert.alert-secondary{ role: 'alert' }
+    %i.fa.fa-info-circle.text-info
+    This request is superseded by
+    = link_to("request #{superseded_by}", request_show_path(number: superseded_by))
+    = surround('(', ')') do
+      = link_to('Show diff', request_show_path(number: superseded_by, diff_to_superseded: number))
+- if diff_to_superseded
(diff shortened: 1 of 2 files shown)


dmarcoux:
[notification box for superseded diff is missing in superseding request](https://trello.com/c/TaqLPzzO/1042-notification-box-for-superseded-diff-is-missing-in-superseding-request)

dmarcoux:
Review link for a superseded request: https://obs-reviewlab.opensuse.org/dmarcoux-alert-superseded-request/request/show/4

src/api/app/views/webui2/webui/request/_superseded_by_message.html.haml
```diff
@@ -0,0 +1,24 @@
+- if superseding.present? && superseded_by.nil? && diff_to_superseded.nil?
+  .alert.alert-secondary{ role: 'alert' }
+    %i.fa.fa-info-circle.text-info
+    This request supersedes:
+    - superseding.each do |bs_request|
+      = link_to("request #{bs_request.number}", request_show_path(number: bs_request.number))
+      = surround('(', ")#{', ' unless bs_request == superseding.last}") do
```
vpereira:
```
 = surround('(', ")#{', ' unless bs_request == superseding.last}") do
```
there isn't a better way to do this? for me looks like cryptic

dmarcoux: I tried also with doing something like this:

```ruby
= surround('(', ')') do
  = link_to('Show diff', request_show_path(number: bs_request.superseded_by, diff_to_superseded: bs_request.number))
= ', ' unless bs_request == superseding.last
```

This would be a lot more readable, but there is then a space between the closing parenthesis and the comma.

vpereira: maybe a helper @dmarcoux ?

dmarcoux: I have the feeling that it is super specific and that it would be used only once.

DavidKang: I'm with @dmarcoux, this is something very specific

vpereira: I don't think the helpers are just about how many times it is being used, but as well about simplify the view logic. what we have now ```= surround('(', ")#{', ' unless bs_request == superseding.last}")``` is really easy to oversee while working with the view but at the same time hard to understand what's going on.

PR #6971, opened 2019-02-13
Add Superseded comments

bgeuken:
This PR adds comments of superseded requests, if there are any. To make this work I had update the comment code to allow having multiple comment sections in one document.

This also includes a few additional changes, like fixing bullet issues, refactoring the js code and cleaning up the views. 

screenshot_2019-02-13 request 15 - open build service

Link for review: https://obs-reviewlab.opensuse.org/bgeuken-superseded_comments/request/show/4


bgeuken:
In case you wonder... I kept the comment tabs now because they allow to switch between the comment sections quickly and require less scrolling than having a sparate card with superseded requests.

coolo:
https://obs-reviewlab.opensuse.org/bgeuken-superseded_comments/request/show/2 crashes when logging in as Admin. Perhaps now is a good time to reenable tests?

bgeuken:
[Comments](https://trello.com/c/z8nwbpmn/1030-comments)

bgeuken:
There is a WIP label:wink:

And the tests are enabled already...

coolo:
> And the tests are enabled already...

Damn. There goes my running gag.

bgeuken:
You still have VCR :clown_face: We won't take it from you

src/api/spec/bootstrap/features/webui/requests_spec.rb
```diff
@@ -23,6 +23,23 @@
   end
 
   context 'request show page' do
+    let!(:superseded_bs_request) do
+      request = create(:set_bugowner_request)
```
Ana06: what about doing this what a factory? or there is any reason why this is not possible? :confused:

src/api/spec/bootstrap/features/webui/requests_spec.rb
```diff
@@ -41,7 +41,18 @@
...
+        within 'li', text: 'Supersedes 1' do
+          find('a', text: '1').click
+        end
+        expect(page).to have_text('In state superseded')
+        within 'li', text: 'Superseded by 2' do
+          find('a', text: '2')
+        end
+      end
+
+      skip('The overview doesn\'t have the new collapse feature') do
```
Ana06: don't we miss a TODO to remove this test when dropping the old UI? :thinking:

src/api/app/views/webui2/webui/request/_request_comments.html.haml
```diff
@@ -0,0 +1,27 @@
...
+.card-body
+  .tab-content
+    .tab-pane.fade.show.active{ id: 'comments' }
+      = render partial: 'webui2/webui/comment/show', locals: { commentable: bs_request,
+        comment_counter_id: "#comment-counter-request-#{bs_request.number}" }
+    - if superseding
+      - superseding.includes(:comments).each do |superseded|
+        - if superseded.comments.any?
+          .tab-pane.fade.show{ id: "superseded-#{superseded.number}" }
+            = render partial: 'webui2/webui/comment/show', locals: { commentable: superseded,
```
dmarcoux: I like that we have the title on the tab (at line 12), but I think we should also write in the tab content that these comments are from the superseded request.

dmarcoux:
I will take over since @bgeuken is not here today.

src/api/app/views/webui2/webui/request/show.html.haml
```diff
@@ -35,13 +35,31 @@
...
-      .card-body#comments
-        = render partial: 'webui2/webui/comment/show', locals: { commentable: @bs_request }
+      .bg-light
+        %ul.nav.nav-tabs{ role: 'tablist' }
+          %li.nav-item
+            = link_to('#comments', class: 'nav-link text-nowrap active', data: { toggle: 'tab' }, role: 'tab') do
+              Comments for request #{@number}
+              %span.badge.badge-primary
+                = @comments.size
+          - if @superseding
```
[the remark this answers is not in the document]
dmarcoux: It is needed because `@superseding` can be `nil`, which would then make the next line fail.

src/api/app/views/webui2/webui/request/_request_comments.html.haml
dmarcoux: I gave it a try to see how it would look like. I'm not convinced and I think it's fine as it is now. I'm curious if users will be confused by the multiple comments tabs (when a request is superseded)

dmarcoux:
@bgeuken's remote is blocking pushes to this PR. I will have to open a new PR... sorry.

src/api/spec/bootstrap/features/webui/requests_spec.rb
bgeuken: That wouldn't work because the request model doesn't allow to initialize a request as superseded

bgeuken:
@dmarcoux Thanks for taking care of this

PR #6977, opened 2019-02-13
Provide compat monitor route

eduardoj:
The mandatory 'project' parameter for the project monitor page was introduced in f795712.

Routes like `/project/monitor?project=my_project` will now be redirected to `/project/monitor/my_project`, so users can keep the same monitor routes they already had, and will not get an 404 error.

Thanks to @darix for reporting it.

### DIFF
--- a/src/api/config/routes.rb
+++ b/src/api/config/routes.rb
@@ -143,13 +143,13 @@ OBSApi::Application.routes.draw do
         get 'package/rpmlint_log' => :rpmlint_log, constraints: cons
         get 'package/meta/:project/:package' => :meta, constraints: cons, as: 'package_meta'
         post 'package/save_meta/:project/:package' => :save_meta, constraints: cons, as: 'package_save_meta'
-        # compat route
+        # For backward compatibility
         get 'package/attributes/:project/:package', to: redirect('/attribs/%{project}/%{package}'), constraints: cons
         get 'package/edit/:project/:package' => :edit, constraints: cons, as: :package_edit
-        # compat routes
+        # For backward compatibility
         get 'package/repositories/:project/:package', to: redirect('/repositories/%{project}/%{package}'), constraints: cons
         get 'package/import_spec/:project/:package' => :import_spec, constraints: cons
-        # compat route
+        # For backward compatibility
         get 'package/files/:project/:package' => :show, constraints: cons
       end
     end
@@ -452,7 +452,7 @@ OBSApi::Application.routes.draw do
 
   ### /worker
   get 'worker/_status' => 'worker/status#index', as: :worker_status
-  get 'build/_workerstatus' => 'worker/status#index', as: :build_workerstatus # FIXME3.0: drop this compat route
+  get 'build/_workerstatus' => 'worker/status#index', as: :build_workerstatus # For backward compatibility
   get 'worker/:worker' => 'worker/capability#show'
   post 'worker' => 'worker/command#run'
 
--- a/src/api/config/routes.rb
+++ b/src/api/config/routes.rb
@@ -287,6 +287,12 @@ OBSApi::Application.routes.draw do
       post 'project/remove_person/:project' => :remove_person, constraints: cons
       post 'project/remove_group/:project' => :remove_group, constraints: cons
       get 'project/monitor/:project' => :monitor, constraints: cons, as: 'project_monitor'
+      # For backward compatibility
+      get 'project/monitor', to: redirect { |_path_parameters, request|
+        url_string = request.query_parameters.except(:project).to_param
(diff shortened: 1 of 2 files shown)


src/api/config/routes.rb
```diff
@@ -287,6 +287,10 @@ def self.public_or_about_path?(request)
       post 'project/remove_person/:project' => :remove_person, constraints: cons
       post 'project/remove_group/:project' => :remove_group, constraints: cons
       get 'project/monitor/:project' => :monitor, constraints: cons, as: 'project_monitor'
+      # compat route
+      get 'project/monitor', to: redirect { |_path_parameters, request|
+        "/project/monitor/#{request.query_parameters[:project]}?#{request.query_parameters.except(:project).to_param}"
```
Ana06: where is this route actually used? What about stop using it instead of redirecting? :eyes:

darix: What about not breaking people's bookmarks for no reason?

dmarcoux:
For some reasons, CircleCI didn't report back about `checkout_code` and `backend_test`. They both passed.

src/api/config/routes.rb
```diff
@@ -287,6 +287,10 @@ def self.public_or_about_path?(request)
       post 'project/remove_person/:project' => :remove_person, constraints: cons
       post 'project/remove_group/:project' => :remove_group, constraints: cons
       get 'project/monitor/:project' => :monitor, constraints: cons, as: 'project_monitor'
+      # compat route
```
dmarcoux: I find the comment `compat route` not so great. I would use something like `Backward compatible route`, but I don't want to block the PR for this.

src/api/config/routes.rb
Ana06: @darix 

> What about not breaking people's bookmarks for no reason?

This comment was for @eduardoj, not for you. Fixing the Bookmarks doesn't means we have to do it in this way. Sorry that this broke for you, but this is IMO not the correct solution. We can come up with a fast solution if this is a priority bug (BTW, where is the bug report?), but this shouldn't stay like that.

Ana06: I got wrong was this what about

src/api/config/routes.rb
Ana06: I would add a `TODO` to remove this route in the next big OBS release

eduardoj:
@Ana06:

> I would add a `TODO` to remove this route in the next big OBS release

`TODO` added. Note that in several other parts of the code it is used `FIXME`, `FIXME3.0`, or `FIXME: 3.0`.

src/api/config/routes.rb
eduardoj: @dmarcoux:

> I find the comment `compat route` not so great. I would use something like `Backward compatible route`, but I don't want to block the PR for this.

I added a commit to change this comment, and all others in the file, to `For backward compatibility`.

src/api/config/routes.rb
```diff
@@ -287,6 +287,10 @@ def self.public_or_about_path?(request)
       post 'project/remove_person/:project' => :remove_person, constraints: cons
       post 'project/remove_group/:project' => :remove_group, constraints: cons
       get 'project/monitor/:project' => :monitor, constraints: cons, as: 'project_monitor'
+      # For backward compatibility
+      get 'project/monitor', to: redirect { |_path_parameters, request|
+        "/project/monitor/#{request.query_parameters[:project]}?#{request.query_parameters.except(:project).to_param}"
```
dmarcoux: This leaves an extra `?` when there are no query parameters beside `project`. So if you go to `project/monitor?project=home:Admin`, you are redirected to `/project/monitor/home:Admin?`

eduardoj: Thanks. Fixed. Please have another look.

src/api/config/routes.rb
```diff
@@ -287,6 +287,12 @@ def self.public_or_about_path?(request)
       post 'project/remove_person/:project' => :remove_person, constraints: cons
       post 'project/remove_group/:project' => :remove_group, constraints: cons
       get 'project/monitor/:project' => :monitor, constraints: cons, as: 'project_monitor'
+      # For backward compatibility
+      get 'project/monitor', to: redirect { |_path_parameters, request|
+        url_string = request.query_parameters.except(:project).to_param
+        url_string = '?' << url_string unless url_string.empty?
+        "/project/monitor/#{request.query_parameters[:project]}?#{url_string}"
```
dmarcoux: There is still an extra`?` on this line. Please, remove it.

eduardoj: My fault. Thanks. Done.

PR #6981, opened 2019-02-14
DataTable rendering on Monitor page

saraycp:
The content of the cell is now processed and rendered by JavaScript.
This improves the performance of the DataTable initialization and it is
needed for the upcoming server-side pagination.

For example, with openSUSE:Tools (74 packages) there is a difference of
approximately 2 seconds (from ~ 5.3 to ~3.2 seconds)

There is an upcoming PR for server pagination.

### DIFF
--- a/src/api/app/assets/javascripts/webui2/project_monitor.js
+++ b/src/api/app/assets/javascripts/webui2/project_monitor.js
@@ -6,18 +6,74 @@ function setAllLinks(event) {
   $(this).closest('.dropdown-menu').find('input').prop('checked', event.data.checked);
 }
 
-function setupProjectMonitor() { // jshint ignore:line
+function statusCell(meta, statusHash, tableInfo, projectName, packageName) {
+  var info = tableInfo[meta.col - 1];
+  var repository = info[0];
+  var architecture = info[1];
+  var status = statusHash[repository][architecture][packageName] || {};
+  var code = status.code;
+  if (code === undefined) return null;
+
+  var klass = 'build-state-' + code;
+  var output = '<a ';
+  if (['succeeded', 'failed', 'building'].includes(code)) {
+    var url = '/package/live_build_log/' + projectName + '/' + packageName + '/' + repository + '/' + architecture;
+    output += 'href="' + url + '"';
+  } else {
+    var id = meta.row + '-' + meta.col;
+    output += 'href="javascript:void(0);" id="' + id + '"';
+
+    if (status.details !== undefined) {
+      if (code === 'scheduled') klass = 'text-warning';
+      output += ' data-content="' + status.details + '" data-placement="right" data-toggle="popover"';
+    }
+  }
+  output += ' class="' + klass + '">' + code + '</a>';
+  return output;
+}
+
+function initializeMonitorDataTable() {
+  var data = $('tbody').data();
+  var packageNames = data.packagenames;
+  var statusHash = data.statushash;
+  var tableInfo = data.tableinfo;
+  var projectName = data.project;
+
   initializeDataTable('#project-monitor-table', { // jshint ignore:line
     scrollX: true,
     fixedColumns: true,
     pageLength: 50,
     lengthMenu: [[10, 25, 50, 100, -1], [10, 25, 50, 100, "All"]],
+    data: packageNames,
     search: {
       regex: true,
       smart: false,
     },
-    columnDefs: [{ width: 150, targets: 0 }]
+    columnDefs: [
+      {
+        width: 150,
+        targets: 0,
+        className: 'text-left',
+        data: null,
+        render: function (packageName) {
+          var url = '/package/show/' + projectName + '/' + packageName;
+          return '<a href="' + url + '">' + packageName + '</a>';
+        }
+      },
+      {
+        targets: '_all',
+        data: null,
+        className: 'text-center',
+        render: function (packageName, type, row, meta) {
+          return statusCell(meta, statusHash, tableInfo, projectName, packageName);
+        }
+      }
+    ]
   });
+}
+
+function setupProjectMonitor() { // jshint ignore:line
+  initializeMonitorDataTable();
 
   $('#table-spinner').addClass('d-none');
   $('#project-monitor .obs-dataTable').removeClass('invisible');
--- a/src/api/app/helpers/webui/buildresult_helper.rb
+++ b/src/api/app/helpers/webui/buildresult_helper.rb
@@ -29,6 +29,7 @@ module Webui::BuildresultHelper
     end
   end
 
+  # NOTE: There is a JavaScript version of this method in project_monitor.js
   def webui2_arch_repo_table_cell(repo, arch, package_name, status = nil, enable_help = true)
     status ||= @statushash[repo][arch][package_name] || { 'package' => package_name }
     status_id = valid_xml_id("id-#{package_name}_#{repo}_#{arch}")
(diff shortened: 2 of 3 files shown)


src/api/app/helpers/webui/buildresult_helper.rb
```diff
@@ -29,6 +29,7 @@ def arch_repo_table_cell(repo, arch, package_name, status = nil, enable_help = t
     end
   end
 
+  # NOTE: There is a JavaScript version of this method in project_monitor.js
```
Ana06: This comment is here because we expect that the other code using this method (`build_status`) will move to use Javascript in the future. Then the already implemented method can be used.

src/api/app/assets/javascripts/webui2/project_monitor.js
```diff
@@ -6,17 +6,73 @@ function setAllLinks(event) {
...
+    }
+  }
+  output += ' class="' + klass + '">' + code + '</a>';
+  return output;
+}
+
 function setupProjectMonitor() { // jshint ignore:line
+  var data = $('tbody').data();
+  var packageNames = data.packagenames;
+  var statusHash =data.statushash;
```
eduardoj: Nitpick: missing space between `=` and `data`.

src/api/app/assets/javascripts/webui2/project_monitor.js
```diff
@@ -6,18 +6,75 @@ function setAllLinks(event) {
...
+  var status = statusHash[repository][architecture][packageName] || {};
+  var code = status.code;
+  if (code === undefined) { return null; }
+
+  var output = '<a ';
+  if (['succeeded', 'failed', 'building'].includes(code)) {
+    var url = '/package/live_build_log/' + projectName + '/' + packageName + '/' + repository + '/' + architecture;
+    output += 'href="' + url + '"';
+  }
+  else {
```
bgeuken: More of a personal preference... I would prefer `} else {`. I think it's also how we write this elsewhere

src/api/app/assets/javascripts/webui2/project_monitor.js
```diff
@@ -6,18 +6,75 @@ function setAllLinks(event) {
...
+    var id = meta.row + '-' + meta.col;
+    output += 'href="javascript:void(0);" id="' + id + '"';
+
+    if (status.details !== undefined) {
+      if (code === 'scheduled') klass = 'text-warning';
```
bgeuken: Please wrap this in `{}` like you did [above](https://github.com/openSUSE/open-build-service/pull/6981/files#diff-52aa359322a395d7dce7d088fe03e479R15)

Ana06: or do not wrap it above, as it is not needed....

src/api/app/assets/javascripts/webui2/project_monitor.js
Ana06: the linter does not complain.... I don't think style should be a topic we take care of. Although I like it more as you say as well...

src/api/app/assets/javascripts/webui2/project_monitor.js
```diff
@@ -6,18 +6,75 @@ function setAllLinks(event) {
...
+      output += ' data-content="' + status.details + '" data-placement="right" data-toggle="popover"';
+    }
+  }
+  var klass = 'build-state-' + code;
+  output += ' class="' + klass + '">' + code + '</a>';
+  return output;
+}
```
bgeuken: Have you considered to using JQuery to create the element? I think that would allow us to add add data attributes, classes and the href in a simpler way using JQuery methods (using `data()` and `addClass`).

Ana06: I am not sure what you mean, can you please elaborate?

bgeuken: It might have some performance impact to use JQuery here, but if it makes the code considerably more readable and doesn't make it to slow it should be worth it. We could measure it

Ana06: This would have performance consequences, I don't find it more readable and it is out of the scope of this PR. Remember this PR is needed to continue improving the efficiency of the page.

Ana06: I understand know. This would have performance consequences, I don't find it more readable, it may bring problems with escaping urls and it is out of the scope of this PR. Remember this PR is needed to continue improving the efficiency of the page.

dmarcoux:
[Improve monitor performance more](https://trello.com/c/15QGgiqX/1045-improve-monitor-performance-more)

Ana06:
> @bgeuken  added [Frontend](/openSUSE/open-build-service/labels/Frontend%20%F0%9F%91%BB)  [review-app](/openSUSE/open-build-service/labels/review-app)  labels [4 minutes ago](#event-2142072822)

It looks exactly the same as before

Ana06:
> @bgeuken  added [Frontend](/openSUSE/open-build-service/labels/Frontend%20%F0%9F%91%BB)  [review-app](/openSUSE/open-build-service/labels/review-app)  labels [4 minutes ago](#event-2142072822)

It looks exactly the same as before, it is just more efficient. If you want to try this, I would recommend you to "copy" openSUSE:Tools (or another similar project) via the interconnect and I guess you don't want to do this in the review-app (to automatize it)

src/api/app/assets/javascripts/webui2/project_monitor.js
dmarcoux: Instead of concatenating strings (which is what we do with `output` now), this is how you could do it with jQuery:

```javascript
$('<div>', { class: 'lala' });
// returns Object [ div.lala ]
```

[Documentation](https://api.jquery.com/jQuery/#jQuery2)

bgeuken:
No worries, I wasn't planning to do performance testing on the review app:laughing:

src/api/app/assets/javascripts/webui2/project_monitor.js
eduardoj: Is this `klass` variable ever used? Later in line 30 it is redefined.

src/api/app/assets/javascripts/webui2/project_monitor.js
Ana06: @dmarcoux

> could we try if using jQuery is really a hit in the performance

You can try it... But I still don't get what we gain with this. You can do it with jQuery and you can also do it with a string. What do we gain from changing it? This is already way much faster and it is needed to continue the performance improvement. I don't think this discussion belongs to this PR

bgeuken: > We can try it... But I still don't get what we gain with this.

We would gain more readable code :smiley:

I already collected some different ways how we could create the element and plan to create some benchmark for it. What I don't like about this code is that it mixes how the data is collected and how it is presented.

But I am fine doing this in a second step, given that the sprint is almost over.

src/api/app/assets/javascripts/webui2/project_monitor.js
```diff
@@ -6,18 +6,74 @@ function setAllLinks(event) {
...
+          var url = '/package/show/' + projectName + '/' + packageName;
+          return '<a href="' + url + '">' + packageName + '</a>';
+        }
+      },
+      {
+        targets: '_all',
+        data: null,
+        className: 'text-center',
+        render: function (packageName, type, row, meta) {
+          return statusCell(meta, statusHash, tableInfo, projectName, packageName);
```
bgeuken: nitpick: You probably could hand over the whole data set given that we seem to fetch all of them [above](https://github.com/openSUSE/open-build-service/pull/6981/files#diff-52aa359322a395d7dce7d088fe03e479R40). And if there is additional data we craft the data attribute differently.

Ana06: I am not sure what you mean. Can you please elaborate?

src/api/app/assets/javascripts/webui2/project_monitor.js
dmarcoux: > But I am fine doing this in a second step, given that the sprint is almost over.

Me too

PR #6986, opened 2019-02-15
Refactor Request#show controller and view

bgeuken:
- Reduce number of instance variables in the controller
- Avoid using webui_infos when its not necessary
- Move text to its own partial

Co-authored-by: Björn Geuken <bgeuken@suse.de>


bgeuken:
Follow up of #6956

bgeuken:
[Clean up instance variables in show action (request controller)](https://trello.com/c/OppkrgTZ/1043-clean-up-instance-variables-in-show-action-request-controller)

bgeuken:
I have fixed the test failures on my machine, but it seems we have some network issues that prevent me from pushing my changes.
Once this is solved I will do this.

PR #6987, opened 2019-02-15
Binaryid tracking

adrianschroeter:

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,8 +1,0 @@
-version: "2"
-exclude_patterns:
-  - "dist/"
-  - "src/api/test/"
-  - "src/api/spec/"
-  - "src/api/db/"
-  - "docs/api/restility"
-  - "src/api/app/assets/javascripts/webui"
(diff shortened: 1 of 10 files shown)


coolo:
@adrianschroeter I think our future selves will be glad if you put a little more motivation behind this commit into the log message. I assume it's obvious to *you* what a binary_id is, but it's not to me.

src/api/test/functional/channel_maintenance_test.rb
```diff
@@ -754,6 +754,7 @@ def test_large_channel_test
     # event handling
     UpdateNotificationEvents.new.perform
     get '/search/released/binary', params: { match: "repository/[@project = 'BaseDistro3' and @name = 'BaseDistro3_repo']" }
+puts @response.body
```
bgeuken: This looks like debugging code:wink:

bgeuken:
Just wanted to write the same^^

adrianschroeter:
yes, that is the reason why the dont merge me label is there and the commit has this particular name.

It is not supposed to be finished and more to come. Just a workaround because test suite broke locally again in a new interessting way.

src/api/test/functional/channel_maintenance_test.rb
coolo: just ignore the 'DROP ME' commit?

.codeclimate.yml
```diff
@@ -1,8 +0,0 @@
-version: "2"
```
hennevogel: Why do you delete the codeclimate configuration?

adrianschroeter:
On Montag, 18. Februar 2019, 19:30:20 CET Henne Vogelsang wrote:
> hennevogel requested changes on this pull request.
> 
> > @@ -1,8 +0,0 @@
> 
> -version: "2"
> 
> Why do you delete the codeclimate configuration?

because the test suites would not even start atm (silly one line too long 
error).

-- 

Adrian Schroeter
email: adrian@suse.de

dmarcoux:
Code Climate doesn't prevent the test suite from running. However, the `ci/circleci: checkout_code` check does.

adrianschroeter:
got merged into https://github.com/openSUSE/open-build-service/pull/7562

PR #6988, opened 2019-02-15
Fix decision box tabs header for reviews

dmarcoux:
The reviewer could still submit reviews, but it was misleading due to the wrong tab header. About the code changes, we use the same helper as the old UI.

Before:
before-fix

After:
after-fix


dmarcoux:
@openSUSE/obs-frontend Please review. It's a small change, but it's a deal breaker IMHO for the deliverable if we don't have this.

PR #6993, opened 2019-02-18
Make collapsible text generic and use it for request description

bgeuken:
* Make collapsible text generic so we can use it elsewhere.
* Make request description collapsible.
* Update font size of collapsed text. It's was to small (much smaller than the rest of the page).

### DIFF
--- a/src/api/app/views/webui2/webui/webui/_collapsible_text.html.haml
+++ b/src/api/app/views/webui2/webui/webui/_collapsible_text.html.haml
@@ -2,7 +2,7 @@
   .obs-collapsible-textbox
     - if text.length > threshold
       .obs-collapsible-text.obs-collapsed
-        = simple_format(text, class: 'm-0 p-0 small')
+        = simple_format(text, class: 'm-0 p-0')
       %a.obs-collapsible-link.obs-collapsed.text-center.d-block{ title: 'Show more' }
     - else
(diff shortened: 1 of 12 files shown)


bgeuken:
[Request Overview](https://trello.com/c/BDvEJdkV/1029-request-overview)

src/api/app/views/webui2/webui/webui/_collapsible_text.html.haml
```diff
@@ -0,0 +1,8 @@
+- if text.present?
+  .obs-collapsible-textbox
+    - if text.length > threshold
+      .obs-collapsible-text.obs-collapsed
+        = simple_format(text, class: 'm-0 p-0')
+      %a.obs-collapsible-link.obs-collapsed.text-center.d-block{ title: 'Show more' }
```
dmarcoux: [nitpick]

It looks a bit weird with the arrow in the center in some cases, but I'm not sure if there is a solution for all the cases. Anyway, I tried it with `text-right` instead of `text-center`. This is the result:

suggestion

bgeuken: I find it a bit to far away from the actual description in your screenshot.

Let's take it as is for now. We can still update this later on.

PR #7024, opened 2019-02-22
[api] validate project undelete operation for meta data

adrianschroeter:
supposed to prove that content get restored and that operations happen
atomar

This is showing that undelete happens in two steps atm, so the test case is supposed to fail atm


vpereira (approved):

vpereira:
@adrianschroeter looks like the minitest isn't passing.

adrianschroeter:
correct, this was just supposed to document the bug. I did not provide a fix, sorry.

eduardoj:
I rebased on master.

eduardoj:
I rebased on master again.

PR #7025, opened 2019-02-22
Make flash messages stick at the top of the viewpoint

bgeuken:
... and cleanup the flash partial a bit.

When the scrollbar is at the top of the page:

After scrolling down:

I increased the z-index of the flash. So all elements except the watchlist are currently in the back of the flash, when they overlap.

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/flash.scss
+++ b/src/api/app/assets/stylesheets/webui2/flash.scss
@@ -1,0 +1,4 @@
+#flash.sticky-top {
+  top: 1rem;
+  z-index: 1030;
+}
--- a/src/api/app/assets/stylesheets/webui2/flash.scss
+++ b/src/api/app/assets/stylesheets/webui2/flash.scss
@@ -2,3 +2,35 @@
   top: 1rem;
   z-index: 1030;
 }
+
+.alert-success {
+  @extend .alert-success;
+
+  i.fas {
+    @extend .fa-check-circle;
+  }
+}
+
+.alert-error {
+  @extend .alert-danger;
+
+  i.fas {
+    @extend .fa-exclamation-circle;
+  }
+}
+
+.alert-alert {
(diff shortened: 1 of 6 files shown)


hennevogel:
When it appears it moves the viewport down a bit which feels jumpy...

dmarcoux:
> When it appears it moves the viewport down a bit which feels jumpy...

I cannot reproduce this on Firefox 63.0.3.

src/api/app/assets/stylesheets/webui2/flash.scss
```diff
@@ -0,0 +1,36 @@
+#flash.sticky-top {
+  top: 1rem;
+  z-index: 1030;
+}
+
+.alert-success {
+  @extend .alert-success;
+
+  i.fas {
+    @extend .fa-check-circle;
```
dmarcoux: Great idea, I love it!

hennevogel:
> I cannot reproduce this on Firefox 63.0.3.

Add an ubuntu repo on the [repository page](https://obs-reviewlab.opensuse.org/bgeuken-sticky_flash/project/add_repository_from_default_list/hello),

bgeuken:
> **[Hennevogel](/hennevogel) ** commented [3 days ago](#issuecomment-466378538)
> When it appears it moves the viewport down a bit which feels jumpy...

That's because the flash container does not have a min-height. So when adding the flash text the rest of the page moves down.
Since adding a min-height would leave an ugly empty space, I made the flash "fixed-top". See screenshots above. Disadvantage is that it now overlaps with the top navbar.
The other option would be to have a separate flash container for ajax requests.

hennevogel:
I don't like it, it requires me to click all the time on the close button. I guess we should postpone this and live with the 'jumpy' part until we work on the general structure of the layout.

bgeuken:
I switched it back to the initial version.

PR #7032, opened 2019-02-25
Authorize move_path

saraycp:
Additionally turn the after_action around and maintain a whitelist instead of a blacklist. This requires new methods authorized by default.

new_package and it's dialog are just informal authorizations, so people do not even try to do this.

### DIFF
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -37,7 +37,12 @@ class Webui::ProjectController < Webui::WebuiController
   before_action :set_maintained_project, only: [:remove_maintained_project]
   before_action :check_ajax, only: [:buildresult, :edit_comment_form]
 
-  after_action :verify_authorized, only: [:save_new, :new_incident]
+  after_action :verify_authorized, except: [:index, :autocomplete_projects, :autocomplete_incidents, :autocomplete_packages, :autocomplete_repositories,
+                                            :users, :subprojects, :new, :new_incident, :incident_request_dialog, :release_request_dialog, :show,
+                                            :packages_simple, :linking_projects, :buildresult, :delete_dialog, :requests, :monitor, :status,
+                                            :maintained_projects, :add_maintained_project_dialog, :maintenance_incidents, :unlock_dialog,
+                                            :incident_request_dialog, :new_incident_request, :release_request_dialog, :new_release_request,
+                                            :remove_target_request, :remove_target_request_dialog, :toggle_watch, :edit_comment, :edit_comment_form]
 
   def index
     return if switch_to_webui2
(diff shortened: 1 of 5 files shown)


src/api/app/controllers/webui/project_controller.rb
```diff
@@ -115,9 +135,13 @@ def new_incident
     end
   end
 
-  def new_package; end
+  def new_package
+    authorize @project, :update?
```
vpereira: are you sure we should add it here? With that, if the user hasn't authorization, she will be redirected to the main page, changing how it works today: https://21838-2046192-gh.circle-artifacts.com/0/capybara/Projects_creating_packages_in_projects_not_owned_by_user__eg__global_namespace_as_non_admin_user.png

@hennevogel

hennevogel: Sure, you can't do anything on `new_package` anything without logging in or having access right?

vpereira: @hennevogel yes, however is the default behavior on obs to show all the views and do not hide anything from the user. Using the usability joker ("from the usability perspective..."), I think it's better to keep the user on the same screen instead of send him to the main index. What do you think @hennevogel ?

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -61,6 +61,21 @@ def index
     end
   end
 
+  def remove_role
+    authorize @project, :update?
+    super
```
hennevogel: I don't get it why don't you authorize in `Webui::ManageRelationships`?

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -61,6 +61,21 @@ def index
...
+
+  def save_group
+    authorize @project, :update?
+    super
+  end
+
+  def save_person
+    authorize @project, :update?
+    super
+  end
```
hennevogel: I don't get it why don't you authorize in `Webui::ManageRelationships`?

vpereira: I'm not doing anything in ```Webui::ManageRelationships``` that`s why I'm "overwriting" the methods in the controller were they are mixed in via ```include``` and calling authorize and super. Test suite was failing because of them and for me was better to do an authorization check there than white-list it.. what do you think @hennevogel ?

hennevogel: > I'm not doing anything in Webui::ManageRelationships

Clearly, but why? The `Webui::ManageRelationships` mixin brings controller methods. Why don't you add the authorization into the mixin?

vpereira: Hm, the ```lib``` is as generic as possible. Specific bits for this controller should be done in the controller itself, no? IMO the lib and authorization shouldn't be tight coupled.

hennevogel: > Hm, the mixing is as generic as possible. 

Nope

PR #7029, opened 2019-02-22
Migrate and re-design architectures view for bootstrap UI

bgeuken:
Now:

Before:

screenshot-2019-2-22 architecture configuration - open build service

Co-authored-by: Eduardo Navarro <enavarro@suse.com>

PR #7038, opened 2019-02-25
Toggle disabled state for buttons/checkboxes on architectures AJAX events

dmarcoux:
This is to prevent multiple clicks.

A tiny addition to #7029.

### DIFF
--- a/src/api/app/views/webui2/webui/architectures/index.html.haml
+++ b/src/api/app/views/webui2/webui/architectures/index.html.haml
@@ -26,6 +26,7 @@
 - content_for :ready_function do
   :plain
     $('#architectures').on('ajax:before ajax:complete', '.custom-control-input, .filter-button', function() {
+      $(this).prop('disabled', function(_, previousValue) { return !previousValue; });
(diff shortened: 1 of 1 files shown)


hennevogel:
Rails UJS knows [`disable-with`](https://thoughtbot.com/blog/a-tour-of-rails-jquery-ujs#disabling-links-and-buttons), which is what you want, right?

dmarcoux:
> Rails UJS knows [`disable-with`](https://thoughtbot.com/blog/a-tour-of-rails-jquery-ujs#disabling-links-and-buttons), which is what you want, right?

Yes, but it only works on buttons, which is not only what I want here. I also want to disable checkboxes when they are checked/unchecked.

hennevogel:
Are you [sure](https://github.com/rails/rails/blob/master/actionview/app/assets/javascripts/rails-ujs.coffee#L27)?

dmarcoux:
I tried it and it didn't work. I believe it's because we use Bootstrap's custom checkboxes.

hennevogel:
Alright...

PR #7039, opened 2019-02-25
Use ESLint

dmarcoux:
### Update
This is still WIP. As suggested by @vpereira, [DeepScan](https://deepscan.io/) might be an interesting solution for us. It seems to be an extra layer on top of ESLint.

<hr>

I want us to lint our JavaScript code and I know that we already _somewhat_ do it with `JSHint`. I believe it's important that we do it actively and with a good tool which is properly configured. For this purpose, I suggest that we use `ESLint`. Here's why:

ESLint is easily configurable, the documentation is much better than JSHint IMHO, it has a bigger community supporting it (and companies like AirBnb and Facebook) and good [integrations](https://eslint.org/docs/user-guide/integrations). We could also use plugins if we ever need too, ESLint is built to be extended.

I'm not done with this PR, but I prefer to have your opinion before going further with this. Do we want to adopt a JavaScript style guide such as the one from AirBnb or Google? Do we roll with something minimal and adjust it?

TODOs
- [ ] Push Docker image to registry
- [x] Define style guide and adapt ESLint config accordingly
- [x] Autofix errors when possible with `eslint --fix` (everything with a wrench is auto-fixable as described in the [rules](https://eslint.org/docs/rules/)) 
- [x] Generate list of TODOs (I believe it's not possible, but people seem to have workarounds for this https://github.com/eslint/eslint/issues/9844) **It's not possible, so I'll fix the errors instead since there isn't so many of them**
- [x] Check if caching could be enabled (see [Caching section](https://eslint.org/docs/user-guide/command-line-interface#caching))
- [ ] Probably forgot something... :clown_face:

---

To use ESLint, rebuild the Docker images by executing these commands from the repository's root:
```bash
docker build . --file src/api/Dockerfile.frontend-base --tag openbuildservice/frontend-base
docker-compose build frontend
```


dmarcoux:
@openSUSE/obs-frontend Would everybody be fine with using [AirBnb JavaScript style guide](https://github.com/airbnb/javascript)?

eduardoj:
@dmarcoux:

> @openSUSE/obs-frontend Would everybody be fine with using [AirBnb JavaScript style guide](https://github.com/airbnb/javascript)?

Fine for me. Thanks!

DavidKang:
@dmarcoux, fine for me :smile_cat:

.circleci/config.yml
```diff
@@ -68,8 +68,10 @@ jobs:
       - restore_cache:
           key: v4-lint
       - run:
-          name: Install jshint
-          command: sudo npm install -g jshint
+          name: Install eslint and its dependencies
+          command: |
+            sudo npm install -g eslint
+            npm info eslint-config-airbnb-base@latest peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g' | xargs sudo npm install --save-dev eslint-config-airbnb-base@latest
```
hennevogel: Why? :)

dmarcoux: Using the JavaScript style guide from AirBnb depends on some packages, it's described [here](https://www.npmjs.com/package/eslint-config-airbnb-base).

I know it's ugly, but if we had npm5+, we could use something a lot cleaner.

hennevogel: Those are our containers. We can have whatever we want :-)

dmarcoux: I will need to change the container to have a newer version of `npm`, hopefully it's new enough in Leap 42.3

hennevogel: I guess not but there is always [devel:languages:nodejs](https://build.opensuse.org/package/show/devel:languages:nodejs/nodejs10) :cowboy_hat_face:

bgeuken:
> @openSUSE/obs-frontend Would everybody be fine with using AirBnb JavaScript style guide?

I already thumbed it up, but in case this was overlooked:^^: Yes, good idea.:smiley:

vpereira:
@dmarcoux any update on that?

dmarcoux:
> @dmarcoux any update on that?

Sadly no. I prefer to keep this open. I will get to it one day. The scope grew since I need to update the Docker image. We will need to do it anyway one day since Leap 42.3 will be reaching EOL.

hennevogel:
We are now on 15.1...

dmarcoux:
I will get to it, when I have some free time in between deliverables. @vpereira suggested taking a look at [DeepScan](https://deepscan.io/). It might be a viable alternative.

dmarcoux:
So I updated this PR. Once we adjust the linters, we can merge this.

dmarcoux:
[Adapt the linter (rubocop, haml-lint etc.) settings](https://trello.com/c/iMNLZYty/1241-adapt-the-linter-rubocop-haml-lint-etc-settings)

dmarcoux:
I don't understand why `airbrake-js.js` still triggers this problem:

```
'console' is already defined as a built-in global variable. [Error/no-redeclare]
```

dmarcoux:
I don't think it will be possible to use ES6 syntax since the [uglifier used in Rails](https://github.com/lautis/uglifier) supports it only experimentally.

dmarcoux:
It's been almost a year since I started this and the last time I pushed changes is about 3 months ago. I'm closing this since I won't put time into this when it's not going to be merged due to a blocker: https://build.opensuse.org/request/show/732987

PR #7017, opened 2019-02-20
Migrate configuration page to Bootstrap

DavidKang:
We migrate the configuration page to Bootstrap.

screenshot_2019-02-21 configuration - open build service

PR #7041, opened 2019-02-26
Add create new user modal to manage user page

krauselukas:
User creation through the manage user page is part of the configuration area, which gets migrated to bootstrap. To be consistent with the other bootstrap views, i moved the signup form to a modal instead of having a seperate view.

screenshot-2019-2-26 manage users - open build service

### DIFF
--- a/src/api/app/views/webui2/shared/_sign_up_form_input.html.haml
+++ b/src/api/app/views/webui2/shared/_sign_up_form_input.html.haml
@@ -1,0 +1,14 @@
+.form-group
+  = label_tag 'login', 'Username:'
+  = text_field_tag 'login', nil, placeholder: 'Username', autocomplete: 'off', class: 'form-control'
+.form-group
+  = label_tag 'email', 'Email:'
+  = text_field_tag 'email', nil, placeholder: 'Email address', autocomplete: 'off', class: 'form-control'
+.form-group
+  = label_tag 'password', 'Password:'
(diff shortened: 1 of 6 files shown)


krauselukas:
[User creation page](https://trello.com/c/Xhj7xaov/1057-user-creation-page)

src/api/app/views/webui2/shared/_sign_up_form_input.html.haml
```diff
@@ -0,0 +1,14 @@
+.form-group
```
dmarcoux: We need to let our users know which fields are required. For an example to see how we did it before, please have a look at #7017.

src/api/app/views/webui2/webui/user/_register_user.html.haml
```diff
@@ -0,0 +1,11 @@
+.modal.fade#signup-modal{ tabindex: -1, role: 'dialog', aria: { labelledby: 'signup-modal-label', hidden: true } }
```
dmarcoux: I believe we shouldn't use a modal for this form. It's quite "heavy" for a modal. I know that we did use modals a lot in the pages we previously migrated, but I believe we took the wrong decision there. This should be on its own page. It has advantages such as better user feedback when something goes wrong. We don't have anything for that in our modals.

krauselukas: @dmarcoux okay, i will revert it to a detached view

krauselukas:
Deprecated by https://github.com/openSUSE/open-build-service/pull/7064

PR #7043, opened 2019-02-26
Migrate user edit page to bootstrap

bgeuken:

### DIFF
--- a/src/api/app/views/webui2/webui/user/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui2/webui/user/_breadcrumb_items.html.haml
@@ -1,9 +1,17 @@
-- if current_page?(users_path)
+- case action_name
+- when 'index'
   -# This page is only reachable from the configuration, so it's based on configuration
   = render partial: 'webui/configuration/breadcrumb_items'
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
     Manage Users
-- else
-  = render partial: 'webui/main/breadcrumb_items'
+- when 'edit'
+  = render partial: 'webui/configuration/breadcrumb_items'
+  %li.breadcrumb-item
+    = link_to('Manage users', users_path)
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
-    Home of #{@displayed_user}
+    = @displayed_user
+- when 'show'
+  = render partial: 'webui/main/breadcrumb_items'
+  - if action_name == 'show'
+    %li.breadcrumb-item.active{ 'aria-current' => 'page' }
+      Home of #{@displayed_user}
--- a/src/api/app/views/webui2/webui/user/edit.html.haml
+++ b/src/api/app/views/webui2/webui/user/edit.html.haml
@@ -1,0 +1,35 @@
+- @pagetitle = 'Edit User'
+
+.container
+  .card
+    .card-body
+      %h2
+        Edit User #{@displayed_user.login}
+      .row
+        .col-md-6
+          = form_for(@displayed_user, url: user_save_path, method: 'post') do |form|
+            = form.hidden_field(:login)
+            .form-group
+              = form.label(:realname, 'Name:')
+              = form.text_field(:realname, disabled: user_is_configurable(@configuration, @displayed_user), class: 'form-control')
+            .form-group
+              = form.label(:email, 'Email:')
+              %abbr.text-danger{ title: 'required' } *
+              = form.text_field(:email, required: true, email: true, disabled: user_is_configurable(@configuration, @displayed_user),
+                class: 'form-control')
+            .form-group
+              = form.collection_check_boxes(:role_ids, Role.global, :id, :title, {}) do |checkbox|
+                .custom-control.custom-checkbox
+                  = checkbox.check_box(class: 'custom-control-input')
+                  = checkbox.label(class: 'custom-control-label')
+            .form-group
+              = form.collection_radio_buttons(:state, User::STATES - ['subaccount'], :to_s, :to_s, class: 'custom-control-input') do |radio|
+                .custom-control.custom-radio.custom-control-inline
+                  = radio.radio_button(class: 'custom-control-input')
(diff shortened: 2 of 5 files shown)


src/api/app/views/webui2/webui/user/edit.html.haml
```diff
@@ -0,0 +1,34 @@
...
+        Editing User #{@displayed_user.login}
+      .row
+        .col-md-6
+          = form_for(@displayed_user, url: user_save_path, method: 'post') do |form|
+            = form.hidden_field(:login)
+            .form-group
+              = form.label(:realname, 'Name:')
+              = form.text_field(:realname, disabled: user_is_configurable(@configuration, @displayed_user), class: 'form-control')
+            .form-group
+              = form.label(:email, 'e-Mail:')
```
dmarcoux: It should be `Email` to be consistent with other views.

src/api/app/views/webui2/webui/user/edit.html.haml
```diff
@@ -0,0 +1,34 @@
+- @pagetitle = 'Edit User Data'
+
+.container
+  .card
+    .card-body
+      %h2
+        Editing User #{@displayed_user.login}
```
dmarcoux: To be consistent with other views `Edit User #{@displayed_user.login}`

src/api/app/views/webui2/webui/user/edit.html.haml
```diff
@@ -0,0 +1,34 @@
+- @pagetitle = 'Edit User Data'
```
dmarcoux: To be consistent with other views, we should remove `Data`.

src/api/app/views/webui2/webui/user/edit.html.haml
```diff
@@ -0,0 +1,34 @@
...
+
+.container
+  .card
+    .card-body
+      %h2
+        Editing User #{@displayed_user.login}
+      .row
+        .col-md-6
+          = form_for(@displayed_user, url: user_save_path, method: 'post') do |form|
```
dmarcoux: We need to let our users know which fields are required. For an example to see how we did it before, please have a look at #7017.

src/api/app/views/webui2/webui/user/_breadcrumb_items.html.haml
```diff
@@ -1,9 +1,18 @@
...
     Manage Users
-- else
-  = render partial: 'webui/main/breadcrumb_items'
+- when 'edit'
+  = render partial: 'webui/configuration/breadcrumb_items'
+  %li.breadcrumb-item
+    = link_to('Manage users', users_path)
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
-    Home of #{@displayed_user}
+    = @displayed_user
```
dmarcoux: For the groups, I'm doing this when we are on the `Edit Group` page.

`Home (link) > Manage Groups (link) > Group Name (link) > Edit`.

I think we should align here. Which way do you prefer?

bgeuken:
Note that I already enabled the tests for this view (it's not part of the configuration tabs). I had to do some updates and included some workaround for the custom form checkboxes. Let's see if the test suite now passes.

src/api/app/views/webui2/webui/user/_breadcrumb_items.html.haml
bgeuken: That's shown on the user's home page, like [here](https://build.opensuse.org/user/show/bgeuken). On the configuration page it would be `Configuration > Manage Users` and `Configuration > Manage Users > tux` (I changed it to edit instead of tux, aka the user name).
I think the configuration part is fine, since this is part of the admin configuration. Users would use a different view (dialog actually) to configure their account.

dmarcoux: Oh my bad, thank you for the explanation.

bgeuken:
[User edit page](https://trello.com/c/OHvZi8Zs/1056-user-edit-page)

src/api/spec/support/features/custom_forms_helper.rb
```diff
@@ -0,0 +1,11 @@
+module CustomFormsHelper
```
dmarcoux: It's good to have a workaround, but I believe it's unneeded. Even though custom Bootstrap checkboxes are not directly clickable, the label is.

The `check/uncheck` method have [the option `allow_label_click`](https://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FActions:check) which works with Bootstrap and non-Bootstrap checkboxes.

One example from one of our feature specs:

https://github.com/openSUSE/open-build-service/blob/2b5a83a9e27dbfc353b23df0cb35c4ad105fe73c/src/api/spec/features/webui/projects_spec.rb#L139-L140

bgeuken: :thumbsup: Nice. I dropped the commit.

bgeuken:
@dmarcoux Updated. Please review again

src/api/spec/bootstrap/features/webui/users/user_admin_edit_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  let(:admin_role_html_id) { "user_role_ids_#{Role.find_by(title: 'Admin').id}" }
+
+  scenario 'view user' do
+    login(admin)
+    visit user_edit_path(user: user.login)
+
+    expect(page).to have_field('Name:', with: 'John Doe')
+    expect(page).to have_field('Email:', with: 'john@suse.de')
+
+    expect(find_field('Admin', visible: false)).not_to be_checked
```
dmarcoux: We could loop here, it improves the readability:

```ruby
expect(find_field('confirmed', visible: false)).to be_checked

['Admin', 'unconfirmed', 'deleted', 'locked'].each do |field|
  expect(find_field(field, visible: false)).not_to be_checked
end
```

PR #7045, opened 2019-02-26
Migrate 'Manage Groups' pages to Bootstrap

dmarcoux:
Links for review:
- https://obs-reviewlab.opensuse.org/dmarcoux-webui2-manage-groups/groups
- https://obs-reviewlab.opensuse.org/dmarcoux-webui2-manage-groups/group/new
- https://obs-reviewlab.opensuse.org/dmarcoux-webui2-manage-groups/group/edit/review-team
- https://obs-reviewlab.opensuse.org/dmarcoux-webui2-manage-groups/group/show/review-team

group-index
group-create
group-edit
group-show


dmarcoux:
[Manage Groups](https://trello.com/c/qkefDNxt/1052-manage-groups)

hennevogel:
CRUD! YAY! 🙃

src/api/app/views/webui2/webui/groups/new.html.haml
```diff
@@ -0,0 +1,19 @@
...
+  .card-body
+    %h3= @pagetitle
+    = form_for(Group.new, url: group_create_path, method: :post) do |form|
+      .form-group
+        = form.label(:title)
+        %abbr.text-danger{ title: 'required' } *
+        = form.text_field(:title, class: 'form-control', required: true, autofocus: true)
+      .form-group
+        = label_tag(:members)
+        = text_field_tag('group[members]', @members, id: 'group_members', class: 'form-control tag-input', data: { source: user_autocomplete_path })
```
bgeuken: Use dashes in case of html ids and classes: `group-members`

src/api/spec/bootstrap/features/webui/groups_spec.rb
```diff
@@ -2,47 +2,74 @@
...
-    end
+    click_link('Edit', href: group_edit_title_path(group_1))
 
-    find('#group-members-tab').click
-    click_link 'Add User'
-    fill_in 'group_userid', with: 'eisendieter'
-    click_button('Accept')
-    expect(page).to have_content('eisendieter')
+    # Remove all users from tokenfield
+    2.times { first('button.tag-remove').click }
```
bgeuken: IIRC, you could use `find_all` instead

dmarcoux: `find` does find all elements, but clicking them doesn't work because it says something like `Ambiguous match, found 2 elements matching (...)`.

src/api/app/views/webui2/webui/groups/new.html.haml
dmarcoux: Following your suggestion still renders the `input` with the ID `group_members`. I will revert this change since it's confusing to have the view with `group-members` and the JavaScript with `group_members` in when fetching the element with jQuery.

dmarcoux:
@bgeuken Please have a look again, I commented on the suggestions you wrote.

src/api/app/views/webui2/webui/groups/new.html.haml
bgeuken: > Following your suggestion still renders the `input` with the ID `group_members`.

Not on my machine. May it be that you removed the id completely? Because in that case it uses what `text_field_tag` generates which is 'group_members'. In other words you can drop it, if you stick with the underscore version.

> I will revert this change since it's confusing to have the view with `group-members` and the JavaScript with `group_members`

Well, you would have to update any place where the id is used:smiley:

I don't think there is any style guide that would deviate from using hyphens as delimiter for html classes and ids. Similar like ruby using snake_case and js camelCase.

src/api/spec/bootstrap/features/webui/groups_spec.rb
bgeuken: [all / find_all](https://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Finders#all-instance_method) yields the elements. So you probably could write `
all('button.tag-remove').each { |element| element.click }` (*untested*)

src/api/app/assets/javascripts/webui2/tokenfield.js
```diff
@@ -0,0 +1,381 @@
+/* jQuery Tags Input Revisited Plugin
```
bgeuken: This file should go to [vendor/assets](https://guides.rubyonrails.org/asset_pipeline.html#asset-organization)

src/api/app/assets/stylesheets/webui2/tokenfield.scss
```diff
@@ -0,0 +1,124 @@
+// Adapted from https://github.com/underovsky/jquery-tagsinput-revisited
```
bgeuken: This one as well.

bgeuken:
Great that you already ported the tests :+1:

src/api/app/assets/javascripts/webui2/tokenfield.js
dmarcoux: I was wondering if we had something for this. Thank you for the pointer!

src/api/app/views/webui2/webui/groups/edit.html.haml
```diff
@@ -0,0 +1,15 @@
+:ruby
```
Ana06: [nitpick] As it is just one line, you can use `-` instead :wink:

src/api/app/views/webui2/webui/groups/index.html.haml
```diff
@@ -4,3 +4,28 @@
...
+      %thead
+        %tr
+          %th Group Name
+          %th Members
+          %th Action
+      %tbody
+        - @groups.each do |group|
+          %tr
+            %td
+              = link_to(group.title, group_show_path(title: group.title), id: group.title)
```
Ana06: [nitpick] `.title` is not needed (and in the rest of the places where it is converted to string or param neither

src/api/app/views/webui2/webui/groups/index.html.haml
```diff
@@ -4,3 +4,28 @@
...
   .card-body
+    %h3 Manage Groups
+    %table.responsive.table.table-sm.table-striped.table-bordered.w-100#manage-groups-table
+      %thead
+        %tr
+          %th Group Name
+          %th Members
+          %th Action
+      %tbody
+        - @groups.each do |group|
```
Ana06: N+1 queries, `includes(:users)`

src/api/app/views/webui2/webui/groups/index.html.haml
```diff
@@ -4,3 +4,28 @@
...
+          %tr
+            %td
+              = link_to(group.title, group_show_path(title: group.title), id: group.title)
+            %td
+              = safe_join(group.users.order('users.login ASC').map { |user| link_to(truncate(user.login, length: 24), user_show_path(user)) }, ', ')
```
Ana06: [nitpick] `ASC` is the default, so `order('users.login')` should be enough

src/api/app/views/webui2/shared/_group_members.html.haml
```diff
@@ -0,0 +1,20 @@
+- if policy(group).update?
+  %ul.list-inline
+    %li.list-inline-item
+      = link_to(group_edit_title_path(title: group.title), class: 'nav-link') do
+        %i.fas.fa-edit.text-secondary
+        Edit Group
+
+- if group.users.none?
```
Ana06: [nitpick] using `blank?` instead saves as a query to the database as it is already cached for `group.users`

src/api/app/assets/stylesheets/webui2/tokenfield.scss
coolo: we really need a 'how to correctly use open source' training :(

src/api/app/assets/javascripts/webui2/tokenfield.js
dmarcoux: Thinking about this, I think this should also be done for CodeMirror (so {javascripts,stylesheets}/cm2  and collapse-text.scss files).

src/api/app/views/webui2/webui/groups/index.html.haml
dmarcoux: It's already included in the controller

src/api/app/assets/javascripts/webui2/tokenfield.js
```diff
@@ -0,0 +1,381 @@
...
+ * https://github.com/underovsky/jquery-tagsinput-revisited
+ *
+ * Copyright (c) Krzysztof Rusnarczyk
+ * Licensed under the MIT license */
```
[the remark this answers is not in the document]
dmarcoux: I will put it in the header of this file.

dmarcoux:
@bgeuken I adjusted the tokenfield. I still use the same color, which is `badge-secondary` from our patterns. I couldn't find something a lighter blue without it looking weird. I tried a few `alert-*` colors, but they all looked weird. The tokens are now based on the badges, so they should be pretty similar to what we use in other views. I also replaced the remove icon by the one we use normally to remove/delete. I would like to set the inside color of the icon to white, but I will leave it like this for now.

Let me know what you think.

src/api/app/views/webui2/webui/groups/index.html.haml
Ana06: This is because the `order` that is done in the database level. I wonder, do we really want to order by name? I would let it by when it was added, so that it doesn't change the order after saving. If we want to keep this ordering, I would suggest to remove the `includes` from the controller, which saves us 2 database queries (this should only be done in the new UI, as in the old one the order is not there, but just moving it to the old view would make it)

bgeuken:
> @bgeuken I adjusted the tokenfield. I still use the same color, which is `badge-secondary` from our patterns.

secondary color sounds like the right thing to use. I guess it's just that the font size is much smaller than for badges.

I had a brief look; badge texts have `font-weight: 700;`.

With font-weight: inherited (aka. currently)


With font-weight: 700 (aka bootstrap badges):


I think it looks nicer with the changed font-weight and it's worth changing it. But I am not blocking the PR on this.

bgeuken:
The other thing I noticed is that the text of group members is quite smaller (like [here](https://obs-reviewlab.opensuse.org/dmarcoux-webui2-manage-groups/group/show/review-team)). IMO it should have the same font size we use elsewhere.

bgeuken:
And on wider screens the text fields get very wide:


On the other views we limited the size by adding a col-md-6.

dmarcoux:
Thank you @bgeuken and @Ana06 for the feedback. I'm going to address your requests today.

src/api/app/views/webui2/webui/groups/index.html.haml
dmarcoux: It's much better to order data alphabetically than by the creation date because it is constant. For example, the group 'Admins' will always be before 'OBS Developers'. Users would have to always use the search for finding a group because groups wouldn't be ordered. Nobody would remember after a few days when groups were created, so the order by creation date doesn't bring value at all.

src/api/app/views/webui2/webui/groups/new.html.haml
dmarcoux: I don't remember what I did last week, but it was clearly wrong. I guess I removed the `ID` and this is why it was still with the underscore. Anyway, I changed it now.

src/api/spec/bootstrap/features/webui/groups_spec.rb
dmarcoux: This worked, thank you!

src/api/app/views/webui2/webui/groups/index.html.haml
dmarcoux: I fixed the N+1 queries by adding `order(:login)` on the `users` association for the Group model. I don't see why we wouldn't want to order the users by login, so I believe it is fine. Let me know what you think.

dmarcoux:
@Ana06 and @bgeuken, please have another look.

bgeuken:
@dmarcoux One minor thing I noticed is that the group edit view now has the configuration tabs. That's different from the user edit view. We should align this one way or the other.

dmarcoux:
> @dmarcoux One minor thing I noticed is that the group edit view now has the configuration tabs. That's different from the user edit view. We should align this one way or the other.

The `Manage Groups` tab should be active for the group new/edit views, right? Is this what you mean?

src/api/app/models/group.rb
```diff
@@ -7,7 +7,7 @@
 class Group < ApplicationRecord
   has_one :staging_workflow, class_name: 'Staging::Workflow', foreign_key: :managers_group_id, dependent: :nullify
   has_many :groups_users, inverse_of: :group, dependent: :destroy
-  has_many :users, -> { distinct }, through: :groups_users
+  has_many :users, -> { distinct.order(:login) }, through: :groups_users
```
Ana06: nice solution! :+1: It is also affecting the old UI, but I think it is ok

Ana06: btw, `includes(:users)` (instead of `includes(:groups_users, :users)`) in the controller is enough. (although not necessarily part of this PR)

dmarcoux:
I need to fix errors in OBS Package Build.

src/api/app/models/group.rb
dmarcoux: Good catch, I will fix it.

hennevogel:
> I need to fix errors in OBS Package Build.

No you don't. This is not a required check.

dmarcoux:
> No you don't. This is not a required check.

I believe it's already fixed. It's just a matter of listing the directory under the packaged files.

bgeuken:
> dmarcoux commented 2 hours ago
> > @dmarcoux One minor thing I noticed is that the group edit view now has the configuration tabs. That's different from the user edit view. We should align this one way or the other.
>
> The Manage Groups tab should be active for the group new/edit views, right? Is this what you mean?

Sorry, I've overlooked this. What I meant is that the user edit has no configuration tab:



But the group edit does:



I would align the two. I have a small preference to not show the tab, because edit and create (for users) are basically sub-tabs. They are not directly connected to the configuration tab.

That said, I think this is something we can decide in a separate PR.

dmarcoux:
@bgeuken It's clearer now. I also prefer without tabs as the breadcrumbs are more prominent and tabs don't really bring value in this context. I'll do it in this PR.

PR #7046, opened 2019-02-26
Update Bootstrap styling for user notifications

eduardoj:
Before:

screenshot_2019-02-26 user notifications - open build service before

After:

screenshot_2019-02-27 user notifications - open build service_after_4

### DIFF
--- a/src/api/app/views/webui2/webui/users/subscriptions/index.html.haml
+++ b/src/api/app/views/webui2/webui/users/subscriptions/index.html.haml
@@ -1,32 +1,29 @@
 - @pagetitle = 'Notifications'
 
 .card
-  .card-header
+  .card-header.d-flex.justify-content-between
     %h5 Notifications
-  .card-body
-    = form_tag(user_notifications_path, method: :put, id: 'notification_form') do
-      - unless @groups_users.empty?
-        .card-block
-          %h3 Groups
-          %p You will receive emails from the checked groups
-          - @groups_users.each do |group_user|
-            .custom-control.custom-checkbox.custom-control-inline
-              = check_box_tag group_user.group, '1', group_user.email, class: 'custom-control-input', id: "checkbox-#{group_user.group}"
-              = label_tag group_user.group.title, nil, class: 'custom-control-label', for: "checkbox-#{group_user.group}"
-        %br
-      .card-block
-        %h3 Events
-        %p Choose from which events you want to get an email
-        %ul.list-group.list-group-flush
-        #subscriptions-form
-          = render partial: 'webui/subscriptions/subscriptions_form'
-        .card-body
-          = submit_tag 'Update', class: 'btn btn-primary'
-          = link_to('Reset to default', user_notifications_path(default_form: true), remote: true)
+  = form_tag(user_notifications_path, method: :put, id: 'notification_form') do
+    - unless @groups_users.empty?
+      .card-body
+        %h4.card-title Groups
+        %p You will receive emails from the checked groups
+        - @groups_users.each do |group_user|
+          .custom-control.custom-checkbox.custom-control-inline
+            = check_box_tag group_user.group, '1', group_user.email, class: 'custom-control-input', id: "checkbox-#{group_user.group}"
+            = label_tag group_user.group.title, nil, class: 'custom-control-label', for: "checkbox-#{group_user.group}"
+    .card-body
+      %h4.card-title Events
+      %p Choose from which events you want to get an email
+      #subscriptions-form
+        = render partial: 'webui/subscriptions/subscriptions_form'
+    .card-body
+      = submit_tag 'Update', class: 'btn btn-primary'
+      = link_to('Reset to default', user_notifications_path(default_form: true), remote: true)
 
 .card.mt-3
   .card-body
-    %h3 RSS Feed
+    %h4.card-title RSS Feed
     = form_tag(user_rss_token_path, id: 'generate_rss_token_form', method: :post) do
       %p
       - if @user.rss_token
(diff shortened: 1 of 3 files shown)


src/api/app/views/webui2/webui/users/subscriptions/index.html.haml
```diff
@@ -1,32 +1,32 @@
 - @pagetitle = 'Notifications'
 
-.card
-  .card-header
-    %h5 Notifications
+.card.mb-2
+  .card-header.d-flex.justify-content-between
+    %h3 Notifications
```
bgeuken: Let's keep this h5, like for the card headers on other pages

src/api/app/views/webui2/webui/users/subscriptions/index.html.haml
```diff
@@ -1,32 +1,32 @@
...
+              .custom-control.custom-checkbox.custom-control-inline
+                = check_box_tag group_user.group, '1', group_user.email, class: 'custom-control-input', id: "checkbox-#{group_user.group}"
+                = label_tag group_user.group.title, nil, class: 'custom-control-label', for: "checkbox-#{group_user.group}"
+      .card
         .card-body
+          %h4.card-title Events
+          %p Choose from which events you want to get an email
+          %ul.list-group.list-group-flush
+          #subscriptions-form
+            = render partial: 'webui/subscriptions/subscriptions_form'
```
bgeuken: There isn't any margin anymore between the checkbox fields and the button

eduardoj: Done. I moved the button to the parent card. I think int makes more sense.

screenshot from 2019-02-27 10-31-34_user_notifications_button

src/api/app/views/webui2/webui/users/subscriptions/index.html.haml
```diff
@@ -1,32 +1,32 @@
...
+                = check_box_tag group_user.group, '1', group_user.email, class: 'custom-control-input', id: "checkbox-#{group_user.group}"
+                = label_tag group_user.group.title, nil, class: 'custom-control-label', for: "checkbox-#{group_user.group}"
+      .card.mb-3
         .card-body
-          = submit_tag 'Update', class: 'btn btn-primary'
-          = link_to('Reset to default', user_notifications_path(default_form: true), remote: true)
+          %h4.card-title Events
+          %p Choose from which events you want to get an email
+          %ul.list-group.list-group-flush
+          #subscriptions-form
```
dmarcoux: The indentation is wrong here. Should the `div` be outside of the `ul`?

wrong-indentation

bgeuken:
This isn't touching the flash:wink:

dmarcoux:
> This isn't touching the flash :wink:

Well, IMHO it's part of this page's style, so while we're here, let's fix it.

bgeuken:
I didn't notice the additional boxes (card-bodies) before. I think we should slightly re-group / style them. But since we now decided to replace the form by "ajax checkboxes" I would postpone this change.

Sorry for misleading you

src/api/app/views/webui2/webui/users/subscriptions/index.html.haml
eduardoj: Something not related to the changes of this pull request, but I will have a look.

eduardoj:
@dmarcoux:

> The flash should be `success` instead of `notice`

Done.

src/api/app/views/webui2/webui/users/subscriptions/index.html.haml
dmarcoux: Almost every line of the view is changed somehow (indentation, new CSS classes, etc...). What would we need to do to have this related to this PR?

eduardoj: I preserved the relative indentation of the ` %ul.list-group.list-group-flush` and `#subscriptions-form` lines as it was before, because this PR was not intended to fix it.

eduardoj: Done. I've copied the work in progress from #7036, so now that Boostrap card css clases are used, the `ul` is no longer needed.

dmarcoux:
I feel like the page is using too many borders, it doesn't flow well and every card seems to be its own form, although it's one big form. I think we should have one card for everything under _Events_. It reminds me of this [tweet](https://twitter.com/steveschoger/status/897849211110273024).

eduardoj:
@dmarcoux:

> [...] I think we should have one card for everything under _Groups_ and _Events_.

In that case we should modify what is rendered by the `subscriptions_form.html.haml` partial. Or use another partial. As you have not complained with this styling in #7036 I understand that you prefer to have different sytling in both pages. I am wrong?

dmarcoux:
> In that case we should modify what is rendered by the `subscriptions_form.html.haml` partial. Or use another partial. As you have not complained with this styling in #7036 I understand that you prefer to have different sytling in both pages. I am wrong?

Don't get me wrong. I'm not complaining about anything. It's also not about who's right or wrong. It's about improving the final result. The views should be aligned and I believe we should clear out the views by having less lines/borders and separation since it's one form (so a single card for _Groups_ and _Events_).

eduardoj:
> [...] The views should be aligned and I believe we should clear out the views by having less lines/borders and separation since it's one form (so a single card for _Groups_ and _Events_).

Changed. I've used `card-body` classes to differentiate between each section: "Groups", "Events", and the update button.

dmarcoux:
> Changed. I've used card-body classes to differentiate between each section: "Groups", "Events", and the update button.

It's much better!

bgeuken:
[Notifications](https://trello.com/c/dN9wAiMq/1054-notifications)

PR #7048, opened 2019-02-26
Use redirect_to without custom nextstatus

### LINKED ISSUE #6933: ?nextstatus=404 renders old UI design
Expected Result
============
When ?nextstatus=404 is in the URL, the old design is shown even if you have activated the new UI design.

How to Reproduce
===============
1. Join the beta program
2. Open: https://build.opensuse.org/project/show/devel:languages:ruby:extensions?nextstatus=404 => renders old UI design
3. Open: https://build.opensuse.org/project/show/devel:languages:ruby:extensions => renders new UI design

dmarcoux:
Avoid unneeded code and use default redirection HTTP status code

Fixes #6933

_Note: I approved the CodeClimate issues as they are out of scope for this PR_


src/api/app/controllers/webui/project_controller.rb
```diff
@@ -50,9 +50,9 @@ def index
     @important_projects = Project.find_by_attribute_type(atype).where('name <> ?', 'deleted').pluck(:name, :title)
 
     if @spider_bot
-      render :list_simple, status: params[:nextstatus]
+      render :list_simple
     else
-      render :list, status: params[:nextstatus]
+      render :list
     end
   end
```
vpereira: one-liner` render @spider_bot.nil? ? :list :  :list_simple`

coolo:
I like how VCR makes changes to tests so much easier.
#globalwritethroughmustdie

dmarcoux:
[Address Feedback/Issues](https://trello.com/c/70G5o6TN/1048-address-feedback-issues)

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -155,8 +155,7 @@ def statistics
     end
 
     flash[:error] = "No statistics of a successful build could be found in #{@repository}/#{@arch}"
-    redirect_to controller: 'package', action: :binaries, project: @project,
-                package: @package, repository: @repository, nextstatus: 404
+    redirect_to(package_binaries_path(@project, @package, @repository))
```
bgeuken: Was the nextstatus used for anything, aka. should we redirect with status 404?

dmarcoux: `nextstatus` was used for rendering views and have a specific HTTP status code, which doesn't make sense in my opinion. Redirecting has to be with a 3xx HTTP status code. Rendering views with a 404 HTTP status code is weird, because the URL points to a valid page, not to an nonexistent page. Anyway, it was doing something like:

Go to package#statistics, statistics are not found for the package, redirect to package#binaries and render the view with a 404 status code.

---

`redirect_to` won't redirect with a status code which is not 3xx. This is taken from the [documentation](https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to):

>The status code can either be a standard HTTP Status code as an integer, or a symbol representing the downcased, underscored and symbolized description. Note that the status code must be a 3xx HTTP code, or redirection will not occur.

The user still ends up on the page, but it won't be treated as a redirection by the browser.

coolo: This is inherited from 7d7d77c804a6e984c34407e49ddcb6d6dd211fc8 and the idea is that non-existant packages create a 404 response. If you go to https://build.opensuse.org/project/show/ZZZ you get a 404 page, if you go to https://build.opensuse.org/package/show/home:coolo/ZZZ you end up on https://build.opensuse.org/project/show/home:coolo?nextstatus=404 - both URLs give a 404 status code, but one gives you the project. 

I guess a clear 404 is actually handled better from the browser - but then you would drop the redirects all together

dmarcoux: It's true that a clear 404 would be better. Instead of the redirects, I will raise an error like `Backend::NotFoundError` which is handled by `application_controller`:

https://github.com/openSUSE/open-build-service/blob/f2cbeec45076024c85684ab1bd126f35277cc823/src/api/app/controllers/application_controller.rb#L242-L244

I'm not so sure which exactly, I think it will depend on the cases.

hennevogel: `ApplicationController` is API. This is webui...

dmarcoux: Oops, then I'll raise `ActionController::RoutingError`

PR #7059, opened 2019-02-27
Use serverside pagination for rendering projects on the subproject tab

bgeuken:

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -238,10 +238,11 @@ class Project < ApplicationRecord
 
   def siblingprojects
     parent_name = parent.try(:name)
-    return [] unless parent_name
-    Project.where('name like (?) and name != (?)', "#{parent_name}:%", name).order(:name).select do |sib|
+    return Project.none unless parent_name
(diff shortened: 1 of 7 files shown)


bgeuken:
[Server Side Pagination Everywhere](https://trello.com/c/ChgVZ4d6/1047-server-side-pagination-everywhere)

src/api/app/models/project.rb
```diff
@@ -238,10 +238,11 @@ def subprojects
...
   def siblingprojects
     parent_name = parent.try(:name)
-    return [] unless parent_name
-    Project.where('name like (?) and name != (?)', "#{parent_name}:%", name).order(:name).select do |sib|
+    return Project.none unless parent_name
+    projects_id = Project.where('name like (?) and name != (?)', "#{parent_name}:%", name).order(:name).select do |sib|
       sib if parent_name == sib.possible_ancestor_names.first
-    end
+    end.pluck(:id)
+    Project.where(id: projects_id)
```
Ana06: I think we should add a `TODO` here. I think it is fine to merge the server pagination with this ugly fix, but we should refactor this.

bgeuken: I agree, but isn't it evident that this is ugly and should be changed? I don't think a fixme wouldn't help here much.

PR #7061, opened 2019-02-27
Use server side pagination for package list in project view

bgeuken:
We created a new endpoint for retrieving the package list, which
was needed for feeding it to datatable.

### DIFF
--- a/src/api/app/datatables/package_datatable.rb
+++ b/src/api/app/datatables/package_datatable.rb
@@ -1,11 +1,13 @@
 # NOTE: Folowing: https://github.com/jbox-web/ajax-datatables-rails#using-view-helpers
-class ProjectDatatable < AjaxDatatablesRails::ActiveRecord
+class PackageDatatable < AjaxDatatablesRails::ActiveRecord
   extend Forwardable
 
   def_delegator :@view, :link_to
-  def_delegator :@view, :project_show_path
+  def_delegator :@view, :package_show_path
+  def_delegator :@view, :time_ago_in_words
 
   def initialize(params, opts = {})
+    @project = opts[:project]
     @view = opts[:view_context]
     super
   end
@@ -14,34 +16,22 @@ class ProjectDatatable < AjaxDatatablesRails::ActiveRecord
     # Declare strings in this format: ModelName.column_name
     # or in aliased_join_table.column_name format
     @view_columns ||= {
-      name: { source: 'Project.name', cond: :like },
-      title: { source: 'Project.title', cond: :like }
+      name: { source: 'Package.name', cond: :like },
+      changed: { source: 'Package.updated_at', cond: :like }
     }
-  end
-
-  def show_all
-    @show_all ||= options[:show_all]
-  end
-
-  def projects
-    @projects ||= options[:projects]
   end
 
   # rubocop:disable Naming/AccessorMethodName
   def get_raw_records
-    if projects
-      projects
-    else
-      show_all ? Project.all : Project.filtered_for_list
-    end
+    @project.packages.order_by_name
   end
   # rubocop:enable Naming/AccessorMethodName
 
   def data
     records.map do |record|
       {
-        name: link_to(record.name, project_show_path(record)),
-        title: record.title
+        name: link_to(record.name, package_show_path(package: record, project: @project)),
+        changed: time_ago_in_words(Time.at(record.updated_at.to_i))
       }
     end
   end
--- a/src/api/app/views/webui2/webui/project/_project_packages.html.haml
+++ b/src/api/app/views/webui2/webui/project/_project_packages.html.haml
@@ -1,18 +1,12 @@
 .card-body
   - if packages.present?
     .obs-dataTable
-      %table.table.table-sm.table-fixed.table-striped.table-bordered#packages-table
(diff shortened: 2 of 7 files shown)


bgeuken:
[Server Side Pagination Everywhere](https://trello.com/c/ChgVZ4d6/1047-server-side-pagination-everywhere)

bgeuken:
[Server Side Pagination Everywhere](https://trello.com/c/ChgVZ4d6/1047-server-side-pagination-everywhere)

coolo:
how does this work with inherited packages? And if you move things to javascript, shouldn't you add functional tests?

src/api/app/datatables/package_datatable.rb
```diff
@@ -0,0 +1,38 @@
...
+  def view_columns
+    # Declare strings in this format: ModelName.column_name
+    # or in aliased_join_table.column_name format
+    @view_columns ||= {
+      name: { source: 'Package.name', cond: :like },
+      changed: { source: 'Package.updated_at', cond: :like }
+    }
+  end
+
+  # rubocop:disable Naming/AccessorMethodName
```
Ana06: I would exclude the datatables directory instead of doing it inline in all the files

src/api/app/datatables/package_datatable.rb
```diff
@@ -0,0 +1,38 @@
...
+  def get_raw_records
+    @project.packages.order_by_name
+  end
+  # rubocop:enable Naming/AccessorMethodName
+
+  def data
+    records.map do |record|
+      {
+        name: link_to(record.name, package_show_path(package: record, project: record.project)),
```
Ana06: using `@project` instead `record.project` should save us many sql queries :thinking:

src/api/app/views/webui2/webui/project/_project_packages.html.haml
```diff
@@ -1,18 +1,12 @@
 .card-body
```
Ana06: This partial is called for both inherited packages (`@ipackages`) and packages. You can not remove things from here that easily. Also the variable `package` given as parameter when calling this partial is not used anymore.

src/api/app/datatables/package_datatable.rb
bgeuken: Disabling the whole file because we want to ignore a cop in one line, or a few, sounds like overkill to me.

src/api/app/views/webui2/webui/project/_project_packages.html.haml
bgeuken: Oh, right. We forgot to split this up.

src/api/app/datatables/package_datatable.rb
bgeuken: Not really. This is taken care of by rails caching. The only query that happens here is loading the packages.

src/api/app/datatables/package_datatable.rb
Ana06: I mean to exclude `src/api/app/datatables` for `Naming/AccessorMethodName` in `.rubocop` (with a nice comment explaining why we are not going to finish this)

src/api/app/datatables/package_datatable.rb
Ana06: Even if Rails catch the SQL query, still Rails need to check it is cached, while you have already this info stored in a variable available here.

src/api/app/views/webui2/webui/project/_project_packages.html.haml
Ana06: you are right, there are two partials. I checked it wrong. Sorry.

But still you need to remove the `package` variable from:
` = render partial: 'project_packages', locals: { project: @project, packages: @packages }`

src/api/app/datatables/package_datatable.rb
bgeuken: Problem is that this would require me to convert all `Naming/AccessorMethodName` offenses we have right now and there are to many.

src/api/app/datatables/package_datatable.rb
bgeuken: I am very reluctant to do micro-optimizations, but since I know that this isn't leading us anywhere I change it...

src/api/app/views/webui2/webui/project/_project_packages.html.haml
bgeuken: Right now this partial still has a dependency on `@packages`. One could argue that we could, or even should, always show the package table. But I don't think we need to block this PR for that.

bgeuken:
@Ana06 When we triied this out we didn't remove the rubocop disable tags. So we didn't see the offense. I moved the test code to a [branch](https://github.com/bgeuken/open-build-service/tree/server_side_packages_rubocop) in case you want to play around with it and switched the code back to the initial version.

src/api/app/datatables/package_datatable.rb
Ana06: I think this should be straightforward (or there is a bug in Rubocop). But this is not that important, do not block this PR for this. I will take a look into it and send a PR if I find a solution. Thanks for the branch.

src/api/app/views/webui2/webui/project/_project_packages.html.haml
Ana06: ok

PR #7069, opened 2019-02-28
[ci] make osc tests more verbose

M0ses:


M0ses:
checking in kanku if it works

PR #7072, opened 2019-02-28
Don't let disabled projects/repos pass stagings as acceptable

coolo:
When osc crashes while selecting, sometimes we don't reenable the staging project and then it becomes acceptable - which is an error.

We discussed this in the release manager round and there is no valid use case of having disabled repositories *and* wanting it to be acceptable. Mapping this to 'building' is obviously a little fake, but unacceptable sounds too harsh too. So in building stagings the RM will look from time to time to see what's taking so long and she would notice then.

### DIFF
--- a/src/api/spec/models/obs_factory/staging_project_spec.rb
+++ b/src/api/spec/models/obs_factory/staging_project_spec.rb
@@ -379,6 +379,26 @@ RSpec.describe ObsFactory::StagingProject do
 
       it { expect(subject.build_state).to eq(:acceptable) }
     end
+
+    context 'with repositories' do
+      let!(:standard_repo) { create(:repository, project: staging_a) }
+
+      before do
+        allow_any_instance_of(ObsFactory::StagingProject).to receive(:building_repositories).and_return([])
+      end
+
+      context 'with disabled project' do
+        let!(:flag) { create(:build_flag, status: 'disable', project: staging_a) }
+        it { expect(staging_project_a.disabled_repositories).to contain_exactly(standard_repo) }
+        it { expect(staging_project_a.build_state).to eq(:building) }
+      end
+
+      context 'with disabled repo' do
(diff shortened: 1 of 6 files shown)


src/api/spec/models/obs_factory/staging_project_spec.rb
```diff
@@ -379,6 +379,26 @@
...
       it { expect(subject.build_state).to eq(:acceptable) }
     end
+
+    context 'with repositories' do
+      let!(:standard_repo) { create(:repository, project: staging_a) }
+
+      it 'reports building with disabled project' do
+        allow_any_instance_of(ObsFactory::StagingProject).to receive(:building_repositories).and_return([])
+        # overall state is not acceptable if nothing built
+        staging_a.flags.create(flag: 'build', status: 'disable')
```
bgeuken: Those three lines should be in a `before` block. And please use a factory for creating the flag: `create(:build_flag, status: 'disable', project: staging_a)`

src/api/spec/models/obs_factory/staging_project_spec.rb
```diff
@@ -379,6 +379,26 @@
...
+        expect(staging_project_a.disabled_repositories).to contain_exactly(standard_repo)
+        expect(staging_project_a.build_state).to eq(:building)
+      end
+
+      it 'reports building with disabled repo' do
+        allow_any_instance_of(ObsFactory::StagingProject).to receive(:building_repositories).and_return([])
+        # overall state is not acceptable if nothing built
+        staging_a.flags.create(flag: 'build', status: 'disable', repo: standard_repo.name)
```
bgeuken: same here

coolo:
better now?

bgeuken:
Yup, thanks.

coolo:
damn, now that it's deployed it's obvious: there is one repo everyone ignored - but is disabled all the time :)

So I need to rethink this.

Issue #7080, opened 2019-02-28
[production][package#show] Not Found


eduardoj:
package#show threw this 15 times
<pre>
Not Found
</pre>
[https://build.opensuse.org/package/show/openSUSE:Factory:ARM/ghc-imagesize-conduit](https://build.opensuse.org/package/show/openSUSE:Factory:ARM/ghc-imagesize-conduit)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5c77ffe89467720014b2d939)

hennevogel:
@dmarcoux fallout from the 404 thing?

bgeuken:
Looks like it:/

bgeuken:
It's actually not a regression. Just some noise in errbit. Users still see the same 404 view like before.

dmarcoux:
Thank you @bgeuken for taking care of it.

PR #7081, opened 2019-03-01
Migrate user configuration view to bootstrap

bgeuken:
The user actions dropdown got replaced with linked icons like we
usually do in the new UI.
Some of the actions, like confirming and locking users, got removed
since this is already covered by the user edit page which is linked
from within in this table.


bgeuken:
I already started working on the server side pagination, but I also have to take care of #7080. That's why I split this up

bgeuken:
[Manage Users](https://trello.com/c/sakdGS9h/1051-manage-users)

PR #7082, opened 2019-03-01
Fix regression with my previous staging attempt

coolo:
Staging projects have more disabled repositories than I thought, we forgot about the bootstrap_copy repository, which led to all staging projects in Factory being 'building' after my last change was deployed.

Now reverting this in parts and leaving the build_state unaffected, but forcing all checks of disabled repositories to 'pending' as such forcing the staging project to 'testing'

### DIFF
--- a/src/api/app/models/obs_factory/staging_project.rb
+++ b/src/api/app/models/obs_factory/staging_project.rb
@@ -44,7 +44,6 @@ module ObsFactory
     # @return [String] description of the Project object
     delegate :description, to: :project
 
-    delegate :checks, to: :project
     delegate :missing_checks, to: :project
 
     # Checks if the project is adi staging project
@@ -207,9 +206,23 @@ module ObsFactory
     end
 
     def build_state
-      return :building if building_repositories.present? || disabled_repositories.present?
+      return :building if building_repositories.present?
       return :failed if broken_packages.present?
       :acceptable
+    end
+
+    def checks
+      # for stagings we ignore 'old' checks in disabled repositories
+      repos = disabled_repositories
+      project.checks.each do |check|
+        next unless repos.include?(check.status_report.checkable)
+        # we can't simply ignore them as missing_checks is calculated
+        # independently, so we would be left without missing checks
+        # but without checks - so better mark them as 'old/pending',
+        # but only in memory, not in the database.
+        check.state = :pending
+        check.url = nil
+      end
     end
 
     def check_state
--- a/src/api/spec/controllers/webui/obs_factory/staging_projects_controller_spec.rb
+++ b/src/api/spec/controllers/webui/obs_factory/staging_projects_controller_spec.rb
@@ -298,6 +298,36 @@ RSpec.describe Webui::ObsFactory::StagingProjectsController, type: :controller, 
               'overall_state' => 'failed'
             )
           end
+
+          RSpec.shared_examples 'disabled checks' do
+            it 'returns testing' do
+              get :show, params: { project: factory, project_name: 'D' }, format: :json
+              expect(response).to have_http_status(:success)
+
+              json_hash = JSON.parse(response.body)
+              rewritten_hash = check.serializable_hash
+              rewritten_hash['state'] = 'pending'
+              rewritten_hash['url'] = nil
+
+              expect(json_hash).to include(
+                'missing_checks' => [],
+                'checks' => [rewritten_hash],
+                'overall_state' => 'testing'
+              )
+            end
+          end
+
+          context 'with disabled project' do
+            let!(:flag) { create(:build_flag, status: 'disable', project: factory_staging_d) }
+
+            it_behaves_like 'disabled checks'
+          end
+
+          context 'with disabled repository' do
+            let!(:flag) { create(:build_flag, status: 'disable', project: factory_staging_d, repo: images_repository.name) }
+
+            it_behaves_like 'disabled checks'
+          end
         end
 
         context 'published repo has succeeded check' do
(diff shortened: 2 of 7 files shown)


src/api/app/models/obs_factory/staging_project.rb
```diff
@@ -207,11 +206,25 @@ def attributes
...
-      return :building if building_repositories.present? || disabled_repositories.present?
+      return :building if building_repositories.present?
       return :failed if broken_packages.present?
       :acceptable
     end
 
+    def checks
+      # for stagings we ignore 'old' checks in disabled repositories
+      repos = disabled_repositories
+      project.checks.to_a.each do |check|
```
bgeuken: I don't think you need the `to_a`here. [Project#checks](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/project.rb#L1618) a query collection proxy

src/api/spec/controllers/webui/obs_factory/staging_projects_controller_spec.rb
```diff
@@ -298,6 +298,37 @@
               'overall_state' => 'failed'
             )
           end
+
+          RSpec.shared_examples 'disabled checks' do
+
+            it 'returns testing' do
```
bgeuken: Indentation is wrong

src/api/app/models/obs_factory/staging_project.rb
coolo: but what will the return of the each be then?

bgeuken: In case of `Status::Check.none` nothing (no iteration happens).
Otherwise the Status::Check objects that were found by the query.

And if we suspect that at some point we might have projects with hundreds of status checks a `find_each` would make sense.

coolo: the point is I want it to be *one* array that is returned - as I modify the elements. So a find_each is the opposite of what I want.

src/api/spec/controllers/webui/obs_factory/staging_projects_controller_spec.rb
coolo: Fixed

src/api/app/models/obs_factory/staging_project.rb
bgeuken: What difference does it make? The iterated elements (`check`) are the same. Only the "container" class changes:

```ruby
irb(main):008:0> report.checks.to_a.first
=> #<Status::Check id: 1, state: "pending", url: "https://foo", short_description: nil, name: "my_check", created_at: "2019-03-01 12:26:36", updated_at: "2019-03-01 12:26:36", status_reports_id: 2>
irb(main):009:0> report.checks.first
=> #<Status::Check id: 1, state: "pending", url: "https://foo", short_description: nil, name: "my_check", created_at: "2019-03-01 12:26:36", updated_at: "2019-03-01 12:26:36", status_reports_id: 2>
irb(main):010:0> report.checks.find_each.first
=> #<Status::Check id: 1, state: "pending", url: "https://foo", short_description: nil, name: "my_check", created_at: "2019-03-01 12:26:36", updated_at: "2019-03-01 12:26:36", status_reports_id: 2>
```

src/api/app/models/obs_factory/staging_project.rb
```diff
@@ -207,11 +206,25 @@ def attributes
...
+        next unless repos.include?(check.status_report.checkable)
+        # we can't simply ignore them as missing_checks is calculated
+        # independently, so we would be left without missing checks
+        # but without checks - so better mark them as 'old/pending',
+        # but only in memory, no in the database.
```
bgeuken: *not in the database.*

src/api/app/models/obs_factory/staging_project.rb
coolo: because find_each will cast to an array internally if not called with a block - and checks.first is even a SQL query as you access the AREL tree.

But what I want is an array, not a relation and not an enumerator.

src/api/app/models/obs_factory/staging_project.rb
```diff
@@ -207,11 +206,25 @@ def attributes
...
+        check.state = :pending
```
bgeuken: Doesn't this just mean that somewhere in our code we want to ignore the state? :thinking: I could imagine that changing it here will lead to unexpected side effects at some point in time, because noone would expect this.

src/api/app/models/obs_factory/staging_project.rb
coolo: So to answer my initial question in this thread:

```
Status::Check.limit(10).each { |a| a.state = 'pending' }.class
  Status::Check Load (1.6ms)  SELECT  `status_checks`.* FROM `status_checks` LIMIT 10
=> Array

Comment (coolo) [reply src/api/app/models/obs_factory/staging_project.rb]: And also 
```
irb(main):001:0> Status::Check.none.each { |a| a.state = 'pending' }.class
=> Array

coolo: So removed to_a

src/api/app/models/obs_factory/staging_project.rb
coolo: fixed

src/api/app/models/obs_factory/staging_project.rb
coolo: Yeah, perhaps introduce an explicit state instead? Make it 'disabled', but keep the stagin project in :testing?

src/api/app/models/obs_factory/staging_project.rb
bgeuken: Ah, I didn't realize that you were talking about the return value of the checks method, sorry.

src/api/app/models/obs_factory/staging_project.rb
bgeuken: I think I don't know enough about the actual problem yet to answer this.
But it might be one solution. Another option might be to move the filter you use here out of this method and apply it where / when we need it.

coolo: the whole status api is a little crude and leads to broken expectations - so I wouldn't want to extend this further. But you might have noticed that I didn't touch the staging namespace - as we need to reevaluate this, but on a larger scale. This works 'good enough' for now - as those working with the staging projects know what to look for.

bgeuken: Alright, let's see when we bump into this^^

PR #7087, opened 2019-03-01
Introduce AJAX in notifications and user notifications

eduardoj:
Ending work started in #7036.

### DIFF
--- a/src/api/app/controllers/webui/users/subscriptions_controller.rb
+++ b/src/api/app/controllers/webui/users/subscriptions_controller.rb
@@ -20,12 +20,15 @@ class Webui::Users::SubscriptionsController < Webui::WebuiController
       gu.save
     end
 
-    subscriptions_form.update!(params[:subscriptions])
+    subscriptions_form.update!(params[:subscriptions]) if params[:subscriptions]
     flash[:success] = 'Notifications settings updated'
   rescue ActiveRecord::RecordInvalid
     flash[:error] = 'Notifications settings could not be updated due to an error'
   ensure
-    redirect_to action: :index
+    respond_to do |format|
+      format.html { redirect_to action: :index }
+      format.js { render 'webui2/webui/users/subscriptions/update' }
+    end
   end
 
   private
--- a/src/api/spec/bootstrap/features/webui/users/users_notifications_settings_spec.rb
+++ b/src/api/spec/bootstrap/features/webui/users/users_notifications_settings_spec.rb
@@ -3,10 +3,6 @@ require 'browser_helper'
 RSpec.feature "User's notifications settings", type: :feature, js: true do
   let(:user_with_groups) { create(:user_with_groups, login: 'moi_wg') }
 
-  before do
-    skip_if_bootstrap
-  end
-
   scenario 'when a user is in some group' do
     login user_with_groups
     visit user_notifications_path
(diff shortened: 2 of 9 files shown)


eduardoj:
[Notifications](https://trello.com/c/dN9wAiMq/1054-notifications)

eduardoj:
Tests for user notifications in Bootstrap added. Please @openSUSE/open-build-service have a look.

src/api/spec/bootstrap/features/webui/users/users_notifications_settings_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    expect(page).to have_content 'You will receive emails from the checked groups'
+    # Using `visible: :all` for the Bootstrap version (as the checkboxes aren't found otherwise)
+    expect(page).to have_checked_field(group_title, visible: :all)
+    uncheck(group_title, allow_label_click: true)
+    expect(page).to have_content 'Notifications settings updated'
+    expect(page).to have_content 'You will receive emails from the checked groups'
+    # Using `visible: :all` for the Bootstrap version (as the checkboxes aren't found otherwise)
+    expect(page).to have_unchecked_field(group_title, visible: :all)
+  end
+end
```
hennevogel: Are you also testing somewhere that the right things happen apart from the checkbox status?

hennevogel:
Hm is this supposed to work also for users notifications?

eduardoj:
> Hm is this supposed to work also for [users notifications](https://obs-reviewlab.opensuse.org/eduardoj-use_ajax_in_notifications/user/notifications)?

Yes.

src/api/spec/bootstrap/features/webui/users/users_notifications_settings_spec.rb
eduardoj: > Are you also testing somewhere that the right things happen apart from the checkbox status?

No. This test is copied from the [spec/features/webui/users/users_notifications_settings_spec.rb
](https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/features/webui/users/users_notifications_settings_spec.rb) one and adapted to work with bootstrap. In case we should check for the right things happen apart from the checkbox status both tests should be modified.

hennevogel: A test not testing what really happens makes little sense. At least `expect` that the data changes in the way you want it to...

src/api/app/controllers/webui/users/subscriptions_controller.rb
```diff
@@ -20,12 +20,15 @@ def update
       gu.save
     end
 
-    subscriptions_form.update!(params[:subscriptions])
+    subscriptions_form.update!(params[:subscriptions]) if params[:subscriptions]
     flash[:success] = 'Notifications settings updated'
```
hennevogel: :thought_balloon: For this I also wonder if we could be a bit more specific...

src/api/spec/bootstrap/features/webui/users/users_notifications_settings_spec.rb
bgeuken: In the old feature test clicking the 'Update' button was causing a page reload. Since this is using ajax now, you would have to manually reload the page (visit user_notifications_path) to be sure that the new state is stored in the server

hennevogel: You can also just inspect the data...

bgeuken: Sure, but it has it's charm test this via the user interface given that it's a feature test.

Issue #6994, opened 2019-02-18
c&p of collapsible text can cause the text to collapse


bgeuken:
1. Go to the request show page of a request that has history elements (bootstrap enabled).
2. Uncollapse the text of one history element.
3. Select some text and copy it.

=> Once in a while this triggers a click event that causes the text to collapse.

hennevogel:
![](https://github.trello.services/images/mini-trello-icon.png) [Address Feedback/Issues](https://trello.com/c/70G5o6TN/1636-address-feedback-issues)

PR #7095, opened 2019-03-04
Do not collapse request history when selecting

Ana06:
In the request show page, the history elements which a long description used to collapse when selecting the text. The user needed most likely to open it again.

image

Fixes https://github.com/openSUSE/open-build-service/issues/6994

### DIFF
--- a/src/api/app/assets/javascripts/webui2/collapsible_text.js
+++ b/src/api/app/assets/javascripts/webui2/collapsible_text.js
@@ -1,9 +1,12 @@
 $(document).ready(function() {
   $('.obs-collapsible-textbox').on('click', function() {
-    var collapsibleLink = $(this).find('.obs-collapsible-link');
-    var showText = (collapsibleLink.attr('title') === 'Show more' ) ? 'Show less' : 'Show more';
+    var selectedText = document.getSelection().toString();
+    if(!selectedText) {
+      var collapsibleLink = $(this).find('.obs-collapsible-link');
+      var showText = (collapsibleLink.attr('title') === 'Show more' ) ? 'Show less' : 'Show more';
 
-    $(this).find('.obs-collapsible-text, .obs-collapsible-link').toggleClass('obs-collapsed obs-uncollapsed');
-    collapsibleLink.attr('title', showText);
+      $(this).find('.obs-collapsible-text, .obs-collapsible-link').toggleClass('obs-collapsed obs-uncollapsed');
+      collapsibleLink.attr('title', showText);
+    }
   });
 });


hennevogel:
this doesn't handle double clicks. I guess we have established that we should do this at all...

src/api/app/assets/javascripts/webui2/collapsible_text.js
```diff
@@ -1,9 +1,12 @@
 $(document).ready(function() {
   $('.obs-collapsible-textbox').on('click', function() {
```
dmarcoux: To handle double-clicks, this line could be:

`$('.obs-collapsible-textbox').on('click dblclick', function() {`

dmarcoux:
> I wonder if this should be done with custom Javascript

The code for this is slowly growing and somewhat not straightforward. I would gladly take a simpler solution, but I don't have one in mind right now.

coolo:
Well, I can only come back to my merge of comments and reviews - to give space for both.

hennevogel:
Now it doesn't work at all anymore and additionally looks weird...

peekhoamxz

Ana06:
It seems I understand wrong what @hennevogel meant with double click.. it should stay open... :walking_man:

Ana06:
@hennevogel @dmarcoux can we merge this? :smile:

hennevogel:
> The question would be, why people want to select this text. 

To copy it of course. For whatever reason you can dream up. Copying text is a basic feature of every web page. Do not disable it or make behave in any other way than a normal paragraph. Just don't.

Ana06:
@hennevogel 

> To copy it of course. For whatever reason you can dream up. Copying text is a basic feature of every web page. Do not disable it or make behave in any other way than a normal paragraph. Just don't.

should I make it work with double click as well? I don't think it worth the effort. It is possible to copy already.

hennevogel:
> should I make it work with double click as well?

If it's easily possible, sure. If not this "feature" should be removed.

dmarcoux:
I believe the way we collapse text will have to be improved. I don't like how it looks and we are having weird issues like this one.

bgeuken:
> dmarcoux commented a day ago
>> I wonder if this should be done with custom Javascript
> The code for this is slowly growing and is somewhat not straightforward. I would gladly take a simpler solution, but I don't have one in mind right now.

Me too. I would consider dropping the 'click the text to expand the text' feature. There seem to many edge cases (click, double-click, swipe?) and having an icon for that should be enough.

And the other issue, having only one line of text when the text is collapsed, we probably can only solve by dropping the [ellipsis](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) solution. Though this is something for another PR.

Ana06:
Taking into account that it doesn't seem straight forward to fix and that I think we need to find another solution for this (and @dmarcoux and @bgeuken as well), I would suggest we merge this as it is now and improve it in a more drastically way afterwards. At least people can copy while the perfect solution is implemented. @hennevogel are you fine with this?

PR #7101, opened 2019-03-04
Be consistent in Manage Groups/Users views

dmarcoux:
To compare the two options regarding the new/edit views, the Groups views have tabs and the Users views don't. I prefer with tabs, but I want your opinion on this. Once we decide, I will do it for both sets of pages. We still have one issue to fix for tabs as the `Manage Groups/Users` is not active when we are on the new/edit views.

Refer to the commits for details on what changed exactly. It's already written there.

Links for review:
- http://obs-reviewlab.opensuse.org/dmarcoux-consistent-manage-groups-users/groups
- http://obs-reviewlab.opensuse.org/dmarcoux-consistent-manage-groups-users/group/new
- http://obs-reviewlab.opensuse.org/dmarcoux-consistent-manage-groups-users/group/edit/review-team
- http://obs-reviewlab.opensuse.org/dmarcoux-consistent-manage-groups-users/users
- http://obs-reviewlab.opensuse.org/dmarcoux-consistent-manage-groups-users/user/register_user
- http://obs-reviewlab.opensuse.org/dmarcoux-consistent-manage-groups-users/user/signup
- http://obs-reviewlab.opensuse.org/dmarcoux-consistent-manage-groups-users/user/edit/Iggy

### DIFF
--- a/src/api/app/views/webui2/webui/groups/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui2/webui/groups/_breadcrumb_items.html.haml
@@ -1,20 +1,22 @@
-- if current_page?(groups_path)
+- case action_name
+- when 'index'
   -# This page is only reachable from the configuration, so it's based on configuration
   = render partial: 'webui/configuration/breadcrumb_items'
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
     Manage Groups
-- else
-  = render partial: 'webui/main/breadcrumb_items'
+- when 'new'
+  = render partial: 'webui/configuration/breadcrumb_items'
   %li.breadcrumb-item
-    = link_to 'Groups', groups_path
-  - if current_page?(group_new_path)
-    %li.breadcrumb-item.active{ 'aria-current' => 'page' }
-      Create Group
-  - elsif current_page?(group_show_path)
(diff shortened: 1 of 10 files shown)


bgeuken:
As mentioned earlier, I have a small preference for the edit views without tab.

> We still have one issue to fix for tabs as the Manage Groups/Users is not active when we are on the new/edit views.

That might be one reason to not have tabs. When we are on [the group edit page](https://obs-reviewlab.opensuse.org/dmarcoux-consistent-manage-groups-users/group/new) we currently have no active tab, since there is only 'Manage groups' but not 'Edit groups'.
We could make 'Manage groups' active in that case, but it would still link to a different page than the current one. Which would be unexpected.

A dropdown might help here. :thinking: Though I can't tell if that's a better solution or an overkill without seeing it in action.
The plain (aka tab-less) edit views have their charm as well:wink:

dmarcoux:
The dropdown would be overkill IMHO. Changing the link would be quite confusing as you said, so I agree that we shouldn't do that.

I will remove the tabs. The `Manage Groups/Users` pages are still accessible in the breadcrumbs.

dmarcoux:
@bgeuken, please have another look.

PR #7102, opened 2019-03-04
Reload apache instead of restarting it

eduardoj:
`restart_on_update` called `systemctl try-restart`, while
`service_del_postun -r` calls `systemctl force-reload`,
that is what it is expected to be done on each update of
the `obs-api` package.

Thanks to @bugfinder for providing the solution.

### DIFF
--- a/dist/obs-server.spec
+++ b/dist/obs-server.spec
@@ -601,7 +601,7 @@ fi
 %postun -n obs-api
 %insserv_cleanup
 %service_del_postun %{obs_api_support_scripts}
-%restart_on_update apache2
+%service_del_postun -r apache2
 
 %files
 %defattr(-,root,root)


eduardoj:
Marking as WIP, and 'DO NOT MERGE' as it is not tested yet.

dist/obs-server.spec
```diff
@@ -601,7 +601,7 @@ fi
 %postun -n obs-api
 %insserv_cleanup
 %service_del_postun %{obs_api_support_scripts}
-%restart_on_update apache2
+%restart_on_update_reload apache2
```
hennevogel: But that is not defined. `_restart_on_update_reload` is.

eduardoj: Done.

M0ses:
LGTM

eduardoj:
Changes tested. The reload of apache will work after the package is installed, not the first time it is installed.

Removing WIP, and 'DO NOT MERGE' labels. Ready to review.

PR #7103, opened 2019-03-04
Raise exception if TS guard file exists while running FullTextIndexJob

vpereira:
From time to time, ThinkingSphinx aborts keeping the guard file there
making impossible to rebuild the index. We want to have the exception on errbit to be able to track this issue

### DIFF
--- a/src/api/app/jobs/full_text_index_job.rb
+++ b/src/api/app/jobs/full_text_index_job.rb
@@ -1,9 +1,17 @@
+class ThinkingSphinx::GuardfileExistsError < StandardError
+  def message
+    "Guardfile #{File.join(ThinkingSphinx::Configuration.instance.indices_location, 'ts---all.tmp')} exists already"
+  end
+end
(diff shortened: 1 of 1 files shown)


src/api/app/jobs/full_text_index_job.rb
```diff
@@ -1,9 +1,17 @@
+class ThinkingSphinx::GuardfileExistsError < StandardError
+  def message
+    "Guardfile #{File.join(ThinkingSphinx::Configuration.instance.indices_location, '--all.tmp')} exists already"
```
eduardoj: [Nitpick]: s/--all.tmp/ts---all.tmp/

eduardoj: [Nitpick]: s/--all.tmp/ts---all.tmp/

vpereira: done

PR #7104, opened 2019-03-04
Enable tests for new bootstrap themed views

bgeuken:
We disabled them while working on the new views.


bgeuken:
Extracted from #7097 after I realized that it might be better to get the remaining tests right away

PR #7105, opened 2019-03-04
[backend] add 'nodiff' srcdiff option

mlschroe:
This still shows file states, but does not provide a textual diff.


coolo:
typo in the 2nd commit message

PR #7111, opened 2019-03-05
[api] fix crash on request creation with enabled notification

adrianschroeter:
requests getting default reviewers need to store the review first
before being able to access created_at in notification code.

### DIFF
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -295,7 +295,12 @@ class BsRequest < ApplicationRecord
     new = created_at ? nil : 1
     sanitize! if new && !@skip_sanitize
     super
-    notify if new
+    if new
+      reviews.each do |review|
+        review.save!
+      end
+      notify
+    end
   end
 
   def history_elements


src/api/app/models/bs_request.rb
```diff
@@ -295,7 +295,12 @@ def save!(args = {})
     new = created_at ? nil : 1
     sanitize! if new && !@skip_sanitize
     super
-    notify if new
+    if new
+      reviews.each do |review|
+        review.save!
```
coolo: I'm pretty sure this is what `super` does. And I'm also pretty sure we have tests that create requests and test the notifications. So what you're saying is we miss tests to create requests with default reviewers?

adrianschroeter:
On Dienstag, 5. März 2019, 10:57:03 CET Stephan Kulow wrote:
> coolo requested changes on this pull request.
> 
> > @@ -295,7 +295,12 @@ def save!(args = {})
> 
>      new = created_at ? nil : 1
>      sanitize! if new && !@skip_sanitize
>      super
> -    notify if new
> +    if new
> +      reviews.each do |review|
> +        review.save!
> 
> I'm pretty sure this is what `super` does. And I'm also pretty sure we have
> tests that create requests and test the notifications. So what you're
> saying is we miss tests to create requests with default reviewers?

we have that for sure, but maybe not with notifications.

just check the crashes in errbit of some certain internal system

-- 

Adrian Schroeter
email: adrian@suse.de

coolo:
I don't believe in your change though. We have quite some projects with default reviewers where people have notifications enabled. So there is something else

bgeuken:
> just check the crashes in errbit of some certain internal system

I didn't find the related exception. Could you give us some more context?

adrianschroeter:
solved the issue for the user by fixing his incident setup.

However, still a problematic handling creating invalid objects, but that needs to be fixed by further checks.

Will look into it later, since it is not a pressing issue anymore.

bgeuken:
@coolo pointed me to the errbit error... One culprit I could be that notify is not called via a callback after the save, but from within it.

@adrianschroeter Do you know if the reviews / review data he tried to save was valid?

adrianschroeter:
On Dienstag, 5. März 2019, 15:47:57 CET Björn Geuken wrote:
> @coolo pointed me to the errbit error... One culprit I could be that notify
> is not called via a callback after the save, but from within it.

well, that was inside of save of BsRequest or do you mean ActiveRecord?

> @adrianschroeter Do you know if the reviews / review data he tried to save
> was valid?

it turned out that the relation ships were all broken, since after solving the 
problem with not stored reviews we run into crashes due to not saved actions.

-- 

Adrian Schroeter
email: adrian@suse.de

PR #7117, opened 2019-03-05
Fix #6989. Disable Submit button autocomplete returns nothing

### LINKED ISSUE #6989: package/add_person/ is not bootstrap
If you add a user from the users tab, you get a popup - but if you put an invalid user account there, you end up on package/add_person, which is not bootstrap.

vpereira:
Closes #6989

### DIFF
--- a/src/api/app/mixins/webui/manage_relationships.rb
+++ b/src/api/app/mixins/webui/manage_relationships.rb
@@ -14,7 +14,7 @@ module Webui::ManageRelationships
     rescue NotFoundError,
            Relationship::SaveError => e
       flash[:error] = e.to_s
-      redirect_to add_path(:add_person)
+      switch_to_webui2 ? redirect_to(users_path) : redirect_to(add_path(:add_person))
       return
     end
     respond_to do |format|
@@ -33,7 +33,7 @@ module Webui::ManageRelationships
     rescue NotFoundError,
            Relationship::SaveError => e
       flash[:error] = e.to_s
-      redirect_to add_path(:add_group)
+      switch_to_webui2 ? redirect_to(users_path) : redirect_to(add_path(:add_group))
       return
     end
     respond_to do |format|


vpereira:
[Address Feedback/Issues](https://trello.com/c/70G5o6TN/1048-address-feedback-issues)

src/api/app/assets/javascripts/webui2/autocomplete.js
```diff
@@ -5,10 +5,19 @@ $(document).ready(function() {
...
+      create: function() {
+        // disable submit button
+        $('input.btn.btn-sm.btn-primary.px-4').prop("disabled", true);
+      },
       search: function() {
         $(this).prev().find('i').toggleClass('fa-search fa-spinner fa-spin');
       },
-      response: function() {
+      response: function(event, ui) {
+        if(ui.content.length > 0) {
```
dmarcoux: It's still possible to submit the form without an existing user. In the review app, we have the users `user_1` and `user_2`, so typing `u` in the autocomplete will return these 2 users and enable the submit button. Clicking on the submit button will take us to the `add_person` view.

Instead of checking if some users are returned, we need to check if one of the returned users is an exact match. If it is, then enable the submit button. If not, leave it disabled.

src/api/app/assets/javascripts/webui2/autocomplete.js
```diff
@@ -5,10 +5,19 @@ $(document).ready(function() {
...
       minLength: 2,
+      create: function() {
+        // disable submit button
+        $('input.btn.btn-sm.btn-primary.px-4').prop("disabled", true);
+      },
       search: function() {
         $(this).prev().find('i').toggleClass('fa-search fa-spinner fa-spin');
       },
-      response: function() {
+      response: function(event, ui) {
```
hennevogel: Guess this could move to the select event? Because this fires if I search for `Adm`. If I then click on Accept it tries to add the user `Adm`.

vpereira:
The only issue is that the user still able to submit the form pressing the enter key, no?

what do you think @dmarcoux @hennevogel ?

src/api/app/assets/javascripts/webui2/autocomplete.js
vpereira: I need both actually and change event. The select is for sure, the right event, but as soon as the user, delete a character or insert a new one, the button doesn't get disabled again. Using the change event. The response event as I implemented, should be able to enable/disable the button while you are typing.. not sure if it's an overkill or if depending how big is the list of possible entries, the interface will lag while iterating through the array with ```$.each```

hennevogel: I think you're overdoing it. This exercise is about helping the user to input the right things, it is **not** about keeping them from putting in the wrong thing. We have validations for that.

dmarcoux: The issue is that we have a modal for this. We don't have any way to validate what the user submits and then opening up the modal again if somehow the user input was invalid. If this form was on its own page, this would be way simpler. Modals are bad for forms, we should have modals only for confirm with Yes/No.

dmarcoux:
This is getting out of hand and proving that a modal shouldn't be use for this. I don't think we have the time to refactor this to finish the deliverable, but we will eventually need to get rid of this modal (and many more)

src/api/app/assets/javascripts/webui2/autocomplete.js
vpereira: it's not about validating the user input and avoiding bad data, but usability.

vpereira:
> This is getting out of hand and proving that a modal shouldn't be use for this. I don't think we have the time to refactor this to finish the deliverable, but we will eventually need to get rid of this modal (and many more)

You mean all of them?

dmarcoux:
I went with a completely different solution. Users will get feedback when they try to add a user which doesn't exist. We also avoid all the weird JavaScript errors with focus, change and all their other event friends.

dmarcoux:
@bgeuken and @hennevogel, please have another look.

hennevogel:
I don't get it. This is what we had before right?

dmarcoux:
No, it was redirecting to a view which isn't migrated to Bootstrap. You can try it in the review app and in production to see the difference.

hennevogel:
> No, it was redirecting to a view which isn't migrated to Bootstrap

Ah okay, yes. I still liked the the javascript helping but okay for me too :cowboy_hat_face:

dmarcoux:
Just rebased on master to fix the failing notifications specs.

bgeuken:
Sorry, I've overlooked this earlier

PR #7118, opened 2019-03-05
Fix notification tests for bootstrap UI

bgeuken:
This updates the notifications_spec after recent changes (notification
checkboxes use ajax now).
Since the notifications_spec covers both, the user and admin views for
notifications, we dropped spec for the user notification view.


src/api/spec/bootstrap/features/webui/notifications_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  RSpec.shared_examples 'updatable' do
+    scenario 'notifications' do
+      login user
+      visit path
+
+      expect(page).to have_content(title)
+      notification_field = find('.card-body h5', text: 'Package has failed to build').sibling('.form-inline')
+      ['Maintainer', 'Bugowner', 'Reader', 'Watching the project'].each do |label|
+        notification_field.check(label, allow_label_click: true)
+        expect(page).to have_css('#flash', text: 'Notifications settings updated')
```
dmarcoux: The flash appears on the first time a checkbox is checked, so if we don't close it before checking the following checkboxes, it doesn't really test if they are working. We should close the flash.

bgeuken: True, I changed it now.

dmarcoux: I think you forgot to push the changes.

bgeuken: Now:laughing:

dmarcoux: We should be more specific as there might be other buttons with the `data-dismiss` attribute. I suggest:

`find('#flash button[data-dismiss]').click`

bgeuken: Hm, if there are multiple ones capybara would fail with some "ambiguous element" error. But I can change it. Might even be that the selector is a bit faster with using the html id

dmarcoux: Oh right, it would have failed with the `ambiguous element` error.

PR #7125, opened 2019-03-06
Import request fixes

dbnicholson:
We had an older OBS instance running 2.3 that we finally upgraded last year. Unfortunately, the request migration failed pretty quickly because one of the users had been deleted. This adds a couple fixes and some logging to make debugging the process nicer. I didn't see any other tasks using `Rails.logger`, but it seems a lot nicer than using `puts`. Let me know if you'd prefer otherwise.

### DIFF
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -317,14 +317,13 @@ class BsRequest < ApplicationRecord
   def check_creator
     errors.add(:creator, 'No creator defined') unless creator
     # Allow admins to create requests for deleted or inactive users
-    unless User.current.is_admin?
-      user = User.not_deleted.find_by(login: creator)
-      # FIXME: We should run the authorization on controller level
-      raise APIError unless User.current.can_modify_user?(user)
-      errors.add(:creator, "Invalid creator specified #{creator}") unless user
-      return if user.is_active?
-      errors.add(:creator, "Login #{user.login} is not an active user")
-    end
+    return if User.current.is_admin?
+    user = User.not_deleted.find_by(login: creator)
+    # FIXME: We should run the authorization on controller level
+    raise APIError unless User.current.can_modify_user?(user)
+    errors.add(:creator, "Invalid creator specified #{creator}") unless user
+    return if user.is_active?
+    errors.add(:creator, "Login #{user.login} is not an active user")
   end
 
   def assign_number
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -316,12 +316,15 @@ class BsRequest < ApplicationRecord
 
   def check_creator
     errors.add(:creator, 'No creator defined') unless creator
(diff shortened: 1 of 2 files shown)


vpereira:
Hey @dbnicholson, 

I pushed some changes, could you please take a look?

a) rebased into master

b) fixed some rubocop issues

vpereira:
@bgeuken could you please review again, specifically my commit?

src/api/app/models/bs_request.rb
```diff
@@ -317,14 +317,13 @@ def skip_sanitize
...
     # Allow admins to create requests for deleted or inactive users
-    unless User.current.is_admin?
-      user = User.not_deleted.find_by(login: creator)
-      # FIXME: We should run the authorization on controller level
-      raise APIError unless User.current.can_modify_user?(user)
-      errors.add(:creator, "Invalid creator specified #{creator}") unless user
-      return if user.is_active?
-      errors.add(:creator, "Login #{user.login} is not an active user")
-    end
+    return if User.current.is_admin?
```
dbnicholson: Oh, that's nicer.

src/api/lib/tasks/delayed_job.rake
```diff
@@ -20,8 +20,7 @@ task(importrequests: :environment) do
       if r.save
         Rails.logger.info "Request ##{lastrq} imported"
       else
-        Rails.logger.error "Request ##{lastrq} could not be saved:\n%s" \
-                           % r.errors.full_messages.join("\n")
+        Rails.logger.error format("Request ##{lastrq} could not be saved:\n%<error>s", error: r.errors.full_messages.join("\n"))
```
dbnicholson: Also nicer (I'm a ruby noob for the most part). I'd personally continue the `error` argument on the next line, but it seems like long lines are acceptable here.

vpereira: We use [Rubocop](https://github.com/rubocop-hq/rubocop) to decide it for us :)

dbnicholson: I figured, but you did change the limit from the [default of 80](https://github.com/rubocop-hq/rubocop/blob/master/config/default.yml#L1681) to [150](https://github.com/openSUSE/open-build-service/blob/master/.rubocop.yml#L55).

I guess I need to widen my windows when working on OBS :smile:

vpereira: Well, if you don't want to, you can still run ```rake dev:lint:rubocop:auto_correct``` and it will fix it for you :)

bgeuken:
>  vpereira commented 3 hours ago
> @bgeuken could you please review again, specifically my commit?

Looks good:thumbsup:

src/api/lib/tasks/delayed_job.rake
dbnicholson: Interesting. Good to know!

PR #7127, opened 2019-03-06
Add upper-limit to range to avoid long running queries (boo#1125812)

vpereira:
Search ranges should have upper-limit to avoid long running queries and consequently leading to service degradation. Related with bug https://bugzilla.suse.com/show_bug.cgi?id=1125812


vpereira:
> I would move this inside `gethistory`, but won't block the PR for this.

hey, I did it already, I would move ```gethistory``` and other models to private, but that's is out of the scope of this PR. I will do it subsequently, thanks!

PR #7128, opened 2019-03-06
Show reason and requestor of review requests in the review request field

### LINKED ISSUE #7088: UI: Display message from review addition above review decision
Various tools utilized by the release team add various reviews for various reasons. When the selected reviewers come to the request it is not always clear why their review is being requested and as such what aspect of the request they should be reviewing. The problem is further aggravated on maintenance requests since the multi-action nature means even more people are involved in a single request and the reviews may be for individual actions.

Most recently with the rebuilt from scratch origin-manager the reason for adding the reviews is included as a message when the review is added. Other tools do this to some degree and could be enhanced as needed. Unfortunately it is all to easy to miss the message when performing the review since the web UI does not indicate which history item (with message) related to which review.

Long-story short, the most straight-forward solution would seem to be to display the message that was used when adding the review above the prompt for a decision on the review.

An example request where this would be handy: [request #680478 for openSUSE:Leap:15.1/pfstools](https://build.opensuse.org/request/show/680478).

image

Something more along the following:

image

From the XML one can see the information is all there just not presented.

```xml
<review state="new" when="2019-03-01T12:27:43" who="origin-manager" by_group="leap-reviewers">
  <comment>Changing to a lower priority origin.</comment>
</review>
```

This is a simple scenario, if I login with `jberry_factory` I would also get the tab for my maintainer review and for multi-action requests where reviews are added by package a single user may have multiple reviews.

This solution is generic and thus should be helpful outside of release team.

bgeuken:
Reviewers need this information to know what is expected from them, aka
why a review was assigned to them.

Fixes #7088

Old UI:

New UI:

### DIFF
--- a/src/api/app/helpers/webui/request_helper.rb
+++ b/src/api/app/helpers/webui/request_helper.rb
@@ -194,4 +194,8 @@ module Webui::RequestHelper
     end.html_safe
   end
   # rubocop:enable Style/FormatStringToken, Style/FormatString
+
+  def review_request_reason(bs_request, review)
+    bs_request.request_history_elements.where(description_extension: review[:id]).pluck(:comment).first.presence || 'No reason given'
+  end
 end
--- a/src/api/app/views/webui/request/show.html.erb
+++ b/src/api/app/views/webui/request/show.html.erb
@@ -253,6 +253,8 @@
         <% @my_open_reviews.each_with_index do |review, index| %>
             <div class="review_descision_display <%= 'hidden' if index != 0 %>" id="review_descision_display_<%= index %>">
               <%= form_tag(action: 'modify_review') do %>
+                  <p><%= user_with_realname_and_icon(review[:who], short: true) %> requested:</p>
+                  <p><%= simple_format(review_request_reason(@bs_request, review)) %></p>
                   <p><%= text_area_tag('comment', review[:reason], size: '80x2', style: 'width: 99%', class: 'review_comment', placeholder: 'Please comment on your decision') %></p>
 
                   <p>
--- a/src/api/app/models/review.rb
+++ b/src/api/app/models/review.rb
@@ -189,6 +189,7 @@ class Review < ApplicationRecord
     ret = _get_attributes
     # XML has this perl format, don't use that here
     ret[:when] = created_at
+    ret[:id] = id
     ret
   end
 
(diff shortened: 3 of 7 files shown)


src/api/app/views/webui/request/show.html.erb
```diff
@@ -253,6 +253,8 @@
         <% @my_open_reviews.each_with_index do |review, index| %>
             <div class="review_descision_display <%= 'hidden' if index != 0 %>" id="review_descision_display_<%= index %>">
               <%= form_tag(action: 'modify_review') do %>
+                  <p><%= user_with_realname_and_icon(review[:who], short: true) %> 'requested:'</p>
```
dmarcoux: The quotes around `requested:` should be removed

old-ui

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -194,4 +194,9 @@ def request_action_header(action, creator)
     end.html_safe
   end
   # rubocop:enable Style/FormatStringToken, Style/FormatString
+
+  def review_request_reason(bs_request, review)
+    bs_request.request_history_elements.where(description_extension: review[:id]).pluck(:comment).first ||
+      'No reason given'
```
dmarcoux: `'No reason given'` is never returned because even if the comment is an empty string, it is still truthy.

This then displays nothing in the old and new UI

old-ui
new-ui

dmarcoux: It should also be covered in the specs.

bgeuken: Fixed and added a helper spec for this.

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -194,4 +194,10 @@ def request_action_header(action, creator)
     end.html_safe
   end
   # rubocop:enable Style/FormatStringToken, Style/FormatString
+
+  def review_request_reason(bs_request, review)
+    reason = bs_request.request_history_elements.where(description_extension: review[:id]).pluck(:comment).first
+
+    reason.presence ? reason : 'No reason given'
```
dmarcoux: It can be reduced to `reason.presence || 'No reason given'`

bgeuken: You are right. Nice!

PR #7034, opened 2019-02-25
Migrate interconnect to Bootstrap

DavidKang:
Like the other pages, we also migrate the interconnect page to bootstrap.

screenshot_2019-02-28 interconnect - open build service

PR #7129, opened 2019-03-06
Use UJS instead of custom Javascript

eduardoj:
Follow work started in #7034.


eduardoj:
[Interconnect](https://trello.com/c/gfbQRNld/1053-interconnect)

src/api/app/views/webui2/webui/interconnects/_interconnect_button.html.haml
```diff
@@ -1,4 +1,5 @@
-.list-group-item
+- name_class = name.tr('.', '_')
```
Ana06: `.` is not the only character that we allow in project names that is not valid in a jQuery id. Check: https://learn.jquery.com/using-jquery-core/faq/how-do-i-select-an-element-by-an-id-that-has-characters-used-in-css-notation/

It may be a good idea to move this to a `helper`, something like `valid_jquery_id`, as you are calling it twice and it can be useful in other places (if we don't have one already, I remember we had one, but maybe it was in javascript and not in Rails and it is possible we dropped it)

hennevogel: The more interesting question is why we do this in the view and not wherever we gather the data?

Ana06: This is to generate the `id`, so I think it belongs to the view helper and not to the controller (if that is what you are suggesting @hennevogel)

hennevogel: The only purpose of gathering this data is showing it here right? Why would you gather it wrong and then create a helper in between to *"fix"* it?

eduardoj:
> [nitpick] A nice description would be appreciated

Added.

src/api/app/views/webui2/webui/interconnects/_interconnect_button.html.haml
eduardoj: I used another approach. Now the ids of the interconnect list elements, that are just the name of the projects, that can contain the caracters `+`, `.` and `:`, are used without any substitution. I  added the `valid_jquery_id` helper to create a valid jquery accessor to the interconnect list elements. 

Please @Ana06 and @hennevogel have a look again.

src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -422,5 +422,9 @@ def gravatar_icon(email, size)
   def home_title
     @configuration ? @configuration['title'] : 'Open Build Service'
   end
+
+  def valid_jquery_id(id)
+    id.gsub(/([:.+])/, '\\\\\\\\\1')
```
dmarcoux: I don't get the part with all the backslashes (`\`). Please enlighten me :slightly_smiling_face:

hennevogel: And this has nothing to do with jquery...

eduardoj: > I don't get the part with all the backslashes (`\`). Why are there so many? Please enlighten me :slightly_smiling_face:

I didin't find a simpler way to escape the characters (prepend them with a backslash). Short answer: backslashes must be escaped too.

So this is what I tried for `id = tizen.org`:

```
id.gsub(/([:.+])/, '\\1')        # $('.list-group-item#tizen\1org')
id.gsub(/([:.+])/, '\\\1')       # $('.list-group-item#tizen\1org')
id.gsub(/([:.+])/, '\\\\1')      # $('.list-group-item#tizen\1org')
id.gsub(/([:.+])/, '\\\\\1')     # $('.list-group-item#tizen\.org')
id.gsub(/([:.+])/, '\\\\\\1')    # $('.list-group-item#tizen\.org')
id.gsub(/([:.+])/, '\\\\\\\1')   # $('.list-group-item#tizen\\1org')
id.gsub(/([:.+])/, '\\\\\\\\1')  # $('.list-group-item#tizen\\1org')
id.gsub(/([:.+])/, '\\\\\\\\\1') # $('.list-group-item#tizen\\.org')
```
This last one was the one that worked.

This one: `$('.list-group-item#tizen\.org')`, without escaping the backslashes, seems not to be enough.

coolo: it might be easier to map them to _ in the output :)

eduardoj: > it might be easier to map them to _ in the output :)

It might be. But this way:
- The jquerys ids aren't escaped when rendering the page. Only the id selector is escaped when rendering a UJS response.
- As `_` is also allowed in the name of projects, cases like having `tizen_org` and `tizen.org` will not cause a problem.

dmarcoux: I agree with @coolo. Mapping any unallowed character to `_` is simpler.

eduardoj: > And this has nothing to do with jquery...

This is escaping the jquery id selectors, adding backslashes. Maybe the name of the helper can be changed. Are you referring to this, or to some other thing?

eduardoj: Applied @dmarcoux suggestion of using a `data-interconnect` attribute, instead of using the `id` attribute. No need to use a helper. Thanks!

dmarcoux:
The last paragraph of the first commit should be removed since it's now outdated:

>Add the `valid_jquery_id` helper for creating jquery accessors with especial
characters.

eduardoj:
> The last paragraph of the first commit should be removed since it's now outdated:
> 
> > Add the `valid_jquery_id` helper for creating jquery accessors with especial
> > characters.

Done.

src/api/app/views/webui2/webui/interconnects/_interconnect_button.html.haml
```diff
@@ -1,9 +1,9 @@
-.list-group-item
+.list-group-item{ data: { interconnect: "#{name}" } }
```
dmarcoux: The interpolation is not needed here.

eduardoj: Fixed.

dmarcoux:
@Ana06 and @hennevogel, please have another look.

PR #7132, opened 2019-03-07
Update exception we throw in case of 404

bgeuken:
That should have been part of 3ee580b44f87f060a0f1f68fd4797f6bb96935a0
but I overlooked it.


bgeuken:
Found via [errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5c77ffe89467720014b2d939)

PR #7138, opened 2019-03-07
Show recommends in package binary page

### LINKED ISSUE #6443: show recommends in package information screen
The binary view ("Details" on eg https://build.opensuse.org/package/binaries/openSUSE:Factory/0ad/standard) shows Requries and Provides along with the package that provide/require the symbols. It does not show recommends which would be useful for e.g. patterns that are used to fill the DVD content of products.

bgeuken:
We already fetch this data as part of the package_file variable. So
rendering it here won't cause any performance regression.
And since this data belongs to the package details it makes sense to
include this here.

Fixes #6443

On larger screens:

screenshot-2019-3-7 detailed information about ctris-0 42-13 1 x86_64 rpm - open build service

On smaller screens:

screenshot-2019-3-7 detailed information about ctris-0 42-13 1 x86_64 rpm - open build service 2

### DIFF
--- a/src/api/app/views/webui2/webui/package/_deps.html.haml
+++ b/src/api/app/views/webui2/webui/package/_deps.html.haml
@@ -56,3 +56,32 @@
           %tr
             %td{ colspan: '2' }
               %em No requires
+
+  .table-responsive.col-sm-12.col-md-6
+    %h3 Recommends
+    %table.table.table-striped.table-bordered.table-sm
+      %thead
+        %tr
+          %th.w-75 Symbol
+          %th.w-25 Provided by
+      %tbody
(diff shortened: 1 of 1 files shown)


bgeuken:
@mlschroe btw, I noticed that the backend request for this data can take a while, depending on the package. Would it be possible to cache this? I tried doing this on our side, but it looks like the cache key would have to depend on backend data / state...

The url would we query is `/build/:project/:repository/:arch/:package/:filename?view=fileinfo_ext`.

mlschroe:
I can't cache the "ext" request, as it does a lookup for all binaries in the repository search path providing a dependency. So it depends on data that can be hosted on a remote build service.

But I wonder if the expensive _ext query should really be the default and not an option.

PR #7142, opened 2019-03-07
Update rails stack because CVE-2018-16476

vpereira:
Hakiri complains about old version of ```activejob```. Updating rails fixes it https://hakiri.io/github/openSUSE/open-build-service/2.9/602d80f6fece13970fb3a4894d51424349fe3e9f/warnings?name=Information+Disclosure


coolo:
https://build.opensuse.org/project/show/OBS:Server:2.9 - have fun updating the packages.

vpereira:
since we are not using user input to ```perfom_later``` we can accept the risk and do not do this update.

PR #7143, opened 2019-03-07
Convert rss notification payload

### LINKED ISSUE #7134: [production][feeds#notifications] undefined method `realname' for nil:NilClass
feeds#notifications threw this 10 times
<pre>
undefined method `realname&#39; for nil:NilClass
</pre>
[https://build.opensuse.org/user/feed/DrpwJsUjm9rbXAEvJkKFncNK](https://build.opensuse.org/user/feed/DrpwJsUjm9rbXAEvJkKFncNK)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5c80db5fb7217300080441fe)

bgeuken:
In PR#6879 we converted the commenter and commenters entries of
event payloads. Since the RSS Notification payload is relying
on some of the event views and methods, this broke our RSS feeds.

This commit adds a data migration that converts existing RSS feeds.

Fixes #7134

### DIFF
--- a/src/api/db/data/20190307145633_migrate_rss_notification_payload.rb
+++ b/src/api/db/data/20190307145633_migrate_rss_notification_payload.rb
@@ -1,0 +1,29 @@
+class MigrateRssNotificationPayload < ActiveRecord::Migration[5.2]
+  def up
+    Notification::RssFeedItem.all.find_each { |notification| convert_payload(notification) }
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+
+  private
+
+  def convert_payload(notification)
+    return unless notification.event_type.in?(['Event::CommentForPackage', 'Event::CommentForProject', 'Event::CommentForRequest'])
+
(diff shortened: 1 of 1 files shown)


vpereira:
what are you missing here @bgeuken ?

bgeuken:
> what are you missing here @bgeuken ?

Well, this was a follow up from a regression we had because of a migration. I worked on this as demolition squad and handed it over afterwards.
 IIRC, because of a workaround (#7149) this became less urgent and stalled.

Regarding the code changes, the code is based on the migration from the [initial fix](https://github.com/openSUSE/open-build-service/commit/394769ab1230020cd59c743a970a03ba07a2ce38#diff-7f45477141e04b715fd2c628b05cff76). I didn't test it yet.

So missing parts are:

* Test the migration
* Double-check that there is no code left that could create the payload in the "old" format
* ???

hennevogel:
tested in on obs-test. works

PR #7144, opened 2019-03-07
Remove unneeded word and align filename to modal context

dmarcoux:
A small thing which I noticed... It's a bit redundant to have `New` in the modal's title.

createpackagemodal

### DIFF
--- a/src/api/app/views/webui2/webui/project/show.html.haml
+++ b/src/api/app/views/webui2/webui/project/show.html.haml
@@ -76,7 +76,7 @@
     = render partial: 'delete_project_dialog', locals: { project: @project }
 
     - if show_package_actions?
-      = render partial: 'new_package_modal', locals: { project: @project }
+      = render partial: 'create_package_modal', locals: { project: @project }
       = render partial: 'new_package_branch_modal', locals: { project: @project, remote_projects: @remote_projects }
 
     - if can_be_released?(@project, @packages, @open_release_requests, @has_patchinfo)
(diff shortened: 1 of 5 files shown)


src/api/app/views/webui2/webui/project/show.html.haml
```diff
@@ -76,7 +76,7 @@
     = render partial: 'delete_project_dialog', locals: { project: @project }
 
     - if show_package_actions?
-      = render partial: 'new_package_modal', locals: { project: @project }
+      = render partial: 'create_package_modal', locals: { project: @project }
```
bgeuken: This looks wrong:wink:

dmarcoux: What is wrong? I changed the partial filename to better reflect its content.

saraycp:
IMHO I don't see the point in changing the name of the partial. If it were a common view it would be something like `new.html` and not `create.html`. And it renders an empty form...  :thinking:  so for me is more like a _new*_ thing.

dmarcoux:
> IMHO I don't see the point in changing the name of the partial. If it were a common view it would be something like new.html and not create.html. And it renders an empty form... thinking so for me is more like a new* thing.

Unlike controller actions and their views, there isn't a 1-to-1 mapping around partials, so I find using the modal's title (or context) for the partial filename is a good solution.

src/api/app/views/webui2/webui/project/show.html.haml
bgeuken: Oh, in the GitHub UI it looked like the partial name stayed the same. I overlooked `-> $new_file` part.

Sorry for the noise

dmarcoux: No worries

PR #7151, opened 2019-03-08
Use a list for package binaries in binary view

bgeuken:
Otherwise they are hard to read when packages are
overflowing to the next line.

Before:
screenshot_2019-03-08 detailed information about ctris-0 42-15 1 x86_64 rpm - open build service 2

After:
screenshot_2019-03-08 detailed information about ctris-0 42-15 1 x86_64 rpm - open build service 1

### DIFF
--- a/src/api/app/views/webui2/webui/package/_deps.html.haml
+++ b/src/api/app/views/webui2/webui/package/_deps.html.haml
@@ -13,16 +13,18 @@
               %span.nowrap{ title: package_file['dep'] }
                 = truncate(package_file['dep'], length: 30)
             %td
-              - package_file.elements('requiredby') do |package_binary|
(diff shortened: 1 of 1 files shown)


bgeuken:
[Here](https://build.opensuse.org/package/binary/OBS:Server:2.9:Staging/ruby2.5/openSUSE_42.3/x86_64/ruby2.5-2.5.3-49.5.x86_64.rpm) is a real-life example which also has a larger amount of package binaries.

I think this is another place where it would come in handy to have some kind of expand / collapse mechanic.

dmarcoux:
What's up with CircleCI?

bgeuken:
@DavidKang Thanks David:heart:

PR #7158, opened 2019-03-11
Allow to delete a staging project via API

saraycp:
Also documents it in apidocs.

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -1947,6 +1947,10 @@ GET /staging/<staging_workflow_project>/staging_projects/<staging_project>
   Get the overall state of a staging project
 XmlResult: staging_project
 
+DELETE /staging/<staging_workflow_project>/staging_projects/<staging_project>
+  Delete a staging project
+XmlResult: status_ok
+
 POST /staging/<staging_workflow_project>/staging_projects/<staging_project>/copy/<staging_project_copy_name>
   Copy a staging project
 XmlResult: status_ok
--- a/src/api/config/routes.rb
+++ b/src/api/config/routes.rb
@@ -762,7 +762,7 @@ OBSApi::Application.routes.draw do
 
   # StagingWorkflow API
   resources :staging, only: [], param: 'main_project_name' do
-    resources :staging_projects, only: [:index, :show], controller: 'staging/staging_projects', param: :name do
+    resources :staging_projects, only: [:index, :show, :destroy], controller: 'staging/staging_projects', param: :name do
       post 'copy/:staging_project_copy_name' => :copy
 
       get 'staged_requests' => 'staging/staged_requests#index', constraints: cons
--- a/src/api/app/controllers/staging/staging_projects_controller.rb
+++ b/src/api/app/controllers/staging/staging_projects_controller.rb
@@ -2,6 +2,7 @@ class Staging::StagingProjectsController < ApplicationController
   before_action :require_login, except: [:index, :show]
 
   before_action :set_main_project
+  before_action :set_staging_project, only: [:show, :destroy]
 
   def index
     if @main_project.staging
@@ -12,9 +13,7 @@ class Staging::StagingProjectsController < ApplicationController
     end
   end
 
-  def show
-    @staging_project = @main_project.staging.staging_projects.find_by!(name: params[:name])
-  end
+  def show; end
 
   def copy
     authorize @main_project.staging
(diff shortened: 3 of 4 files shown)


saraycp:
[Allow to Destroy/Remove a Staging Project via the API](https://trello.com/c/2zIIt0hB/1076-allow-to-destroy-remove-a-staging-project-via-the-api)

DavidKang:
Looks good, but I think we need to provide a way to delete multiples staging project instead of doing it one by one.

docs/api/api/api.txt
```diff
@@ -1947,6 +1947,10 @@ GET /staging/<staging_workflow_project>/staging_projects/<staging_project>
   Get the overall state of a staging project
 XmlResult: staging_project
 
+DELETE /staging/<staging_workflow_project>/staging_projects/<staging_project>
```
coolo: why would I delete staging projects differently than regular projects? I guess I need a little context here (for the docs)

coolo:
> Looks good, but I think we need to provide a way to delete multiples staging project instead of doing it one by one.

For what user story?

docs/api/api/api.txt
saraycp: @coolo I don't know If I understood :thinking: Do you propose to have a similar route or use exactly the same for both regular and staging projects?

saraycp:
> >  Looks good, but I think we need to provide a way to delete multiples staging project instead of doing it one by one.
> 
> For what user story?

@coolo in the new UI we do it this way, we have a list of staging projects that we can select with checkboxes and then delete all of them. We should do the same for the API. Don't you think it is necessary?

docs/api/api/api.txt
coolo: Well, I propose you clarify in the documentation why it's necessary to have an extra route and what's breaking if I use the /source route to delete the project.

coolo: But I'm not even sure this is the case.

coolo:
> @coolo in the new UI we do it this way, we have a list of staging projects that we can select with checkboxes and then delete all of them. We should do the same for the API. Don't you think it is necessary?

I wonder how you envision the use of this API - if there is need to delete multiple projects from a client, that client can very well run a loop.

saraycp:
Ok, we have a few things to clarify.

First, we don't have to do a multiple deletion, this PR is about deleting a single staging project (SP). We got confused with the staging workflow (SW) deletion where we decide what to do with its SPs.

Second, do we need a new route to remove a SP if it is a project after all? Apparently not, but...  deleting with `/source/<project>` or `/staging/<staging_workflow_project>/staging_projects/<staging_project>` is slightly different. We have realized that only the SP class calls `after_destroy :update_staging_workflow_on_backend` ([line 218](https://github.com/openSUSE/open-build-service/blob/0ae5b0806ecbc1a84f1e222500c26e82a8ee5f1f/src/api/app/models/staging/staging_project.rb#L218)) that updates the SW xml. After deletion, it changes this

```
<workflow project=\"test_project\" managers=\"heroes\">
  <staging_project name=\"test_project:Staging:A\"/>
</workflow>
```
to this
```
<workflow project=\"test_project\" managers=\"heroes\"></workflow>
```

We can move the code that updates the SW on backend to the Project model, what I think is a bit untidy from the coding point of view. Or we can keep both routes, what I think is a bit risky if someone deletes the SP using the old route `/source/<project>` by accident, we will have an inconsistent SW on backend.

@coolo @hennevogel @DavidKang what do you think?

src/api/config/routes.rb
```diff
@@ -762,7 +762,7 @@ def self.public_or_about_path?(request)
 
   # StagingWorkflow API
   resources :staging, only: [], param: 'main_project_name' do
-    resources :staging_projects, only: [:index, :show], controller: 'staging/staging_projects', param: :name do
+    resources :staging_projects, only: [:index, :show, :destroy], controller: 'staging/staging_projects', param: :name do
```
Ana06: merge conflict :see_no_evil: https://github.com/openSUSE/open-build-service/pull/7161/commits/0dc1b55cf2866a6c39d7e2714d78d71dd4d811ff

docs/api/api/api.txt
Ana06: actually, what prevent you to use the normal project route? :thinking:

saraycp:
> actually, what prevent you to use the normal project route? 

@Ana06 nothing, like it is you can use both. But there is a slightly difference as I mentioned before.

coolo:
> @coolo @hennevogel @DavidKang what do you think?
OK, let's talk webui first. https://build.opensuse.org/project/show/TestLinux:Staging:A is a (staging) project and there is a 'Delete Project' button on it. Does this delete a project or a staging project? 

From my point of view: deleting a project, no matter if it's a maintenance project, a staging project, an incident project or a ... - it should always trigger the right things when deleted (or be blocked, which might be the right thing too).

docs/api/api/api.txt
vpereira: are you able to delete multiple projects via the ```/source``` route @coolo ?

saraycp:
> OK, let's talk webui first. https://build.opensuse.org/project/show/TestLinux:Staging:A is a (staging) project and there is a 'Delete Project' button on it. Does this delete a project or a staging project?

This deletes a regular project.

> From my point of view: deleting a project, no matter if it's a maintenance project, a staging project, an incident project or a ... - it should always trigger the right things when deleted (or be blocked, which might be the right thing too).

After some discussion we came to the same conclusion. So we are going to use only one endpoint to delete via API: `/source/<project>`. And we are going to move the callbacks specifically defined for staging projects to project model.

Superseded by: #7172

PR #7161, opened 2019-03-12
 Add API to create Staging Workflows & refactor staging routes

Ana06:
- Add API to create Staging Workflows, which was only possible to do in the WebUI before. :bowtie: 

- Refactor routes to simplify their definition the routes file and follow Rails defaults.

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -1933,11 +1933,16 @@ XmlResult: required_checks
 == Staging
   WARNING: This API is currently in beta and unstable
 
-=== Staging Projects
+=== Staging Workflow
 
 <staging_workflow_project>: The name of the staging workflow project, e.g. openSUSE:Factory
 <staging_project>: The name of the staging project, e.g. openSUSE:Factory:Staging:A
 <staging_project_copy_name>: The name of the staging project's copy, e.g. openSUSE:Factory:Staging:B
+
+POST /staging/<staging_workflow_project>/workflow
+  Create a staging workflow for a given project.
+XmlBody: create_staging_workflow
+XmlResult: status_ok
 
 GET /staging/<staging_workflow_project>/staging_projects
   Get the overall state of all staging projects belonging to a staging workflow project
--- a/src/api/app/controllers/staging/workflows_controller.rb
+++ b/src/api/app/controllers/staging/workflows_controller.rb
@@ -1,0 +1,38 @@
+class Staging::WorkflowsController < ApplicationController
+  before_action :require_login
+  before_action :set_project
+
+  def create
+    staging_workflow = @project.build_staging
+    authorize staging_workflow
+
+    staging_workflow.managers_group = Group.find_by!(title: xml_hash['managers'])
+
+    if staging_workflow.save
+      render_ok
+    else
+      render_error(
+        status: 400,
+        errorcode: 'invalid_request',
+        message: "Staging for #{@project} couldn't be created: #{staging_workflow.errors.to_sentence}"
+      )
+    end
+  end
+
+  private
+
+  def set_project
+    @project = Project.get_by_name(params[:staging_main_project_name])
+
+    return if @project
+    render_error(
+      status: 404,
+      errorcode: 'not_found',
+      message: "Project '#{params[:staging_workflow_project]}' not found."
+    )
+  end
+
+  def xml_hash
+    Xmlhash.parse(request.body.read) || {}
+  end
+end
--- a/src/api/app/controllers/staging/workflows_controller.rb
+++ b/src/api/app/controllers/staging/workflows_controller.rb
@@ -22,7 +22,7 @@ class Staging::WorkflowsController < ApplicationController
   private
 
   def set_project
-    @project = Project.get_by_name(params[:staging_main_project_name])
+    @project = Project.get_by_name(params[:staging_workflow_project])
 
     return if @project
     render_error(
--- a/src/api/spec/controllers/staging/workflows_controller_spec.rb
+++ b/src/api/spec/controllers/staging/workflows_controller_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe Staging::WorkflowsController, type: :controller, vcr: true do
     context 'with valid staging_project' do
       before do
         login user
-        post :create, params: { staging_main_project_name: project, format: :xml },
+        post :create, params: { staging_workflow_project: project, format: :xml },
                       body: "<workflow managers='#{group}'/>"
       end
 
@@ -22,7 +22,7 @@ RSpec.describe Staging::WorkflowsController, type: :controller, vcr: true do
     context 'invalid user' do
       before do
         login other_user
-        post :create, params: { staging_main_project_name: project, format: :xml },
+        post :create, params: { staging_workflow_project: project, format: :xml },
                       body: "<workflow managers='#{group}'/>"
       end
 
@@ -33,7 +33,7 @@ RSpec.describe Staging::WorkflowsController, type: :controller, vcr: true do
     context 'non-existent project' do
       before do
         login user
-        post :create, params: { staging_main_project_name: 'imaginary_project', format: :xml },
+        post :create, params: { staging_workflow_project: 'imaginary_project', format: :xml },
                       body: "<workflow managers='#{group}'/>"
       end
 
@@ -43,7 +43,7 @@ RSpec.describe Staging::WorkflowsController, type: :controller, vcr: true do
     context 'without group' do
       before do
         login user
-        post :create, params: { staging_main_project_name: project, format: :xml }
+        post :create, params: { staging_workflow_project: project, format: :xml }
       end
 
       it { expect(response).to have_http_status(:not_found) }
--- a/src/api/spec/controllers/staging/workflows_controller_spec.rb
+++ b/src/api/spec/controllers/staging/workflows_controller_spec.rb
@@ -1,0 +1,53 @@
+require 'rails_helper'
+
+RSpec.describe Staging::WorkflowsController, type: :controller, vcr: true do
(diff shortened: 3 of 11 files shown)


src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  before_action :require_login
+  before_action :set_project
+
+  def create
+    staging_workflow = @project.build_staging
+    authorize staging_workflow
+
+    staging_workflow.managers_group = Group.find_by(title: xml_hash)
+
+    unless staging_workflow.managers_group
```
coolo: in the API we usually just throw exceptions for generic problems.

docs/api/api/api.txt
```diff
@@ -1933,6 +1933,15 @@ XmlResult: required_checks
 == Staging
   WARNING: This API is currently in beta and unstable
 
+=== Staging Workflow
+
+<staging_workflow_project>: The project for which we want to create the staging workflow
+
+POST /staging/<staging_workflow_project>/workflow
+
+XmlBody: managers_group
```
coolo: That's not a good API. The next setting required would mean a complete new root node.

src/api/app/controllers/staging/workflows_controller.rb
hennevogel: We don't have to...

src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      return
+    end
+
+    if staging_workflow.save
+      render_ok
+    else
+      render_error(
+        status: 400,
+        errorcode: 'invalid_request',
+        message: "Staging for #{@project} couldn't be created"
```
hennevogel: Because?

src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    staging_workflow.managers_group = Group.find_by(title: xml_hash)
+
+    unless staging_workflow.managers_group
+      render_error(
+        status: 404,
+        errorcode: 'not_found',
+        message: "Managers Group '#{params[:managers_title]}' not found."
+      )
+      return
+    end
```
hennevogel: You can do this, as you do in other controllers, in a `before_filter`.

src/api/app/controllers/staging/workflows_controller.rb
coolo: Right, we can also duplicate code for no gain and increase code complexity.

coolo: BTW: I forgot to flag this as nit-pick.

Ana06: I kept the same structure that other staging API have. I don't have a preference, but we should be consistent...

src/api/app/controllers/staging/workflows_controller.rb
Ana06: It is the same message as in the WebUI. I agree that the reason should be there, but we also want to keep then consistent

src/api/app/controllers/staging/workflows_controller.rb
Ana06: I kept the same structure as in the WebUI code. There it is not in a `before_filter`

docs/api/api/api.txt
Ana06: I am not sure what you mean. What would make it a good API?

coolo: Well, 
```
<staging_workflow>
  <manager_group>test-staging</manager_group>
</staging_workflow>
```

Having the one setting you require *right now* in the root node doesn't allow any extension later.

src/api/app/controllers/staging/workflows_controller.rb
hennevogel: No we don't. The API is for programs.

src/api/app/controllers/staging/workflows_controller.rb
hennevogel: That's not a reason. Doing the right[tm] thing trumps consistency.

src/api/app/controllers/staging/workflows_controller.rb
hennevogel: Same here...

docs/api/api/api.txt
coolo: app/views/models/staging/_workflow.xml.builder uses `<workflow/>`

src/api/app/controllers/staging/workflows_controller.rb
Ana06: I mean that we have the same error message there. You were complaining about the error message, right?

src/api/app/controllers/staging/workflows_controller.rb
Ana06: @hennevogel 

> That's not a reason. Doing the right[tm] thing trumps consistency.

well... that moving this to a filter is the right thing is questionable. You can use a `before_filter` for this, but there is also not reason to use a `before_filter`. I am not repeating code or writing a huge method so that this seems to be needed. That's why I though that being consistent make sense in this case. But I can also change it...

docs/api/api/api.txt
Ana06: I get it, thanks! :cupid:

src/api/app/controllers/staging/workflows_controller.rb
Ana06: it makes sense to change... the code is in the application controller to use it! :smile:

src/api/app/controllers/staging/workflows_controller.rb
Ana06: Now that I am raising an exception as @coolo suggested, I don't think this is an issue anymore. Is it?

Ana06:
[Be able to create a Workflow via the API](https://trello.com/c/TEHt3853/1074-be-able-to-create-a-workflow-via-the-api)

docs/api/api/api.txt
```diff
@@ -1933,6 +1933,15 @@ XmlResult: required_checks
 == Staging
   WARNING: This API is currently in beta and unstable
 
+=== Staging Workflow
+
+<staging_workflow_project>: The project for which we want to create the staging workflow
+
+POST /staging/<staging_workflow_project>/workflow
+
+XmlBody: <workflow managers='group_title'/>
```
Ana06: is this the way we document those things? :confused:

src/api/spec/controllers/staging/workflows_controller_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+
+      it { expect(response).to have_http_status(:success) }
+      it { expect(project.staging).not_to be_nil }
+    end
+
+    context 'invalid user' do
+      before do
+        login other_user
+        post :create, params: { staging_main_project_name: project.to_s, format: :xml },
```
coolo: you replaced the param name everywhere else but not here?

docs/api/api/api.txt
```diff
@@ -1933,6 +1933,15 @@ XmlResult: required_checks
 == Staging
   WARNING: This API is currently in beta and unstable
 
+=== Staging Workflow
+
+<staging_workflow_project>: The project for which we want to create the staging workflow
+
+POST /staging/<staging_workflow_project>/workflow
```
coolo: You forgot to mention what it does, and I would move this into the 'Staging Projects' section - or rename that section, but don't make it 2.

src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  private
+
+  def set_project
+    @project = Project.get_by_name(params[:staging_workflow_project])
+
+    return if @project
+    render_error(
+      status: 404,
+      errorcode: 'not_found',
+      message: "Project '#{params[:project_name]}' not found."
```
coolo: where is :project_name coming from in here? I would just not repeat the name.

src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -0,0 +1,38 @@
+class Staging::WorkflowsController < ApplicationController
+  before_action :require_login
+  before_action :set_project
+
+  def create
+    staging_workflow = @project.build_staging
+    authorize staging_workflow
+
+    staging_workflow.managers_group = Group.find_by!(title: xml_hash['managers'])
```
hennevogel: What happens if this goes wrong?

docs/api/api/api.txt
```diff
@@ -1933,12 +1933,18 @@ XmlResult: required_checks
 == Staging
   WARNING: This API is currently in beta and unstable
 
-=== Staging Projects
+=== Staging Workflow
 
+<staging_workflow_project>: The project for which we want to create the staging workflow
```
coolo: I think mentioning the variable once will do

src/api/app/controllers/staging/workflows_controller.rb
coolo: https://github.com/openSUSE/open-build-service/pull/7161/files#diff-dd30d7dc99a89d91d1f36eeda51d9730R49

docs/api/api/api.txt
Ana06: what do you mean?

coolo: zoom out!

```
<staging_workflow_project>: The project for which we want to create the staging workflow
<staging_workflow_project>: The name of the staging workflow project, e.g. openSUSE:Factory
```

src/api/app/controllers/staging/workflows_controller.rb
Ana06: It ends in the application controller and returns a 404 with "Couldn't find Group"

PR #7162, opened 2019-03-12
Create staging project from existing project

### LINKED ISSUE #6236: Adding existing project as a staging project does not work
When I want to add an existing project as a staging project it is not allowed and I get the message ``Name already taken``. 

Is this what should be expected? I think if the project already exists and it is not added to an existing staging workflow already it should be possible. If it is already assigned to an existing staging workflow we should show an error.

dmarcoux:
Closes #6236

### DIFF
--- a/src/api/app/controllers/webui/staging/projects_controller.rb
+++ b/src/api/app/controllers/webui/staging/projects_controller.rb
@@ -10,16 +10,26 @@ module Webui
 
       def create
         authorize @staging_workflow
-        staging_project = @staging_workflow.staging_projects.build(name: params[:staging_project_name])
+
+        staging_project = Project.where.not(id: @staging_workflow.project_id)
+                                 .find_or_initialize_by(name: params[:staging_project_name])
         authorize(staging_project, :create?)
+
+        redirect_to edit_staging_workflow_path(@staging_workflow)
+
+        if staging_project.staging_workflow_id?
+          flash[:error] = "\"#{staging_project}\" is already assigned to a staging workflow"
+          return
+        end
+
+        staging_project.staging_workflow = @staging_workflow
 
         if staging_project.valid? && staging_project.store
           flash[:success] = "Staging project with name = \"#{staging_project}\" was successfully created"
-        else
-          flash[:error] = "#{staging_project} couldn't be created: #{staging_project.errors.full_messages.to_sentence}"
+          return
         end
 
-        redirect_to edit_staging_workflow_path(@staging_workflow)
+        flash[:error] = "#{staging_project} couldn't be created: #{staging_project.errors.full_messages.to_sentence}"
       end
 
       def show
--- a/src/api/app/models/concerns/staging_project.rb
+++ b/src/api/app/models/concerns/staging_project.rb
@@ -9,6 +9,7 @@ module StagingProject
     after_save :update_staging_workflow_on_backend, if: :staging_project?
     after_destroy :update_staging_workflow_on_backend, if: :staging_project?
     before_create :add_managers_group, if: :staging_project?
+    before_update :add_managers_group, if: proc { |project| project.staging_workflow_id_changed? && project.staging_workflow_id_was.nil? }
 
     scope :staging_projects, -> { where.not(staging_workflow: nil) }
   end
--- a/src/api/spec/controllers/webui/staging/projects_controller_spec.rb
+++ b/src/api/spec/controllers/webui/staging/projects_controller_spec.rb
@@ -29,6 +29,24 @@ RSpec.describe Webui::Staging::ProjectsController do
       end
     end
 
+    context 'an existent non-staging project' do
+      let!(:existent_project) { create(:project, name: "#{project}:new-staging", maintainer: user) }
+
+      before do
+        post :create, params: { staging_workflow_id: staging_workflow.id, staging_project_name: existent_project.name }
+      end
+
+      subject { staging_workflow }
+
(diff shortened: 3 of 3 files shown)


dmarcoux:
[Address Feedback](https://trello.com/c/K9U6WlG1/1081-address-feedback)

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -10,16 +10,27 @@ class ProjectsController < WebuiController
...
-        staging_project = @staging_workflow.staging_projects.build(name: params[:staging_project_name])
+
+        staging_project = ::Staging::StagingProject.unscoped
+                                                   .where.not(id: @staging_workflow.project_id)
+                                                   .find_or_initialize_by(name: params[:staging_project_name])
         authorize(staging_project, :create?)
 
+        redirect_to edit_staging_workflow_path(@staging_workflow)
+
+        if staging_project.staging_workflow_id?
```
Ana06: why do you need to check the id instead of using the relationship?

`if staging_project.staging_workflow`

dmarcoux: I only want to know if there's one, so I avoid fetching it if there's one by checking the ID instead of using the relationship.

Ana06: taking into account that we don't have a validation, that the `id` is set doesn't mean that the object exist, but we should have a validation anyway...

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -10,16 +10,27 @@ class ProjectsController < WebuiController
 
       def create
         authorize @staging_workflow
-        staging_project = @staging_workflow.staging_projects.build(name: params[:staging_project_name])
+
+        staging_project = ::Staging::StagingProject.unscoped
```
Ana06: At this point, the project doesn't have (suposedly) an staging working, so I think we should look in Project instead of doing `Staging::StagingProject.unscoped`

src/api/app/controllers/webui/staging/projects_controller.rb
Ana06: taking into account that we don't have a validation, that the `id` is set doesn't mean that the object exist

how can this be set if in the query we have ` .where.not(id: @staging_workflow.project_id)`?

src/api/app/controllers/webui/staging/projects_controller.rb
dmarcoux: If I use `Project` instead of `Staging::StagingProject`, I cannot assign a staging workflow later (on line 26) since the `Project` model doesn't have the `staging_workflow` association.

Ana06: It is possible to assign the `id`. The only difference between a Project and a StagingProject is that the second one has a `staging_workflow`. When you look for it in the first place you don't expect it (necessarily) to have a `staging_workflowp`.

I think those kind of queries will bring problems. We are also missing some validations that may make this query even more complicated.

dmarcoux:
I'll rebase on _master_ once #7188 is merged. This will solve the `StagingProject` vs `Project` discussion we had.

dmarcoux:
@Ana06, please have a look.

src/api/spec/controllers/webui/staging/projects_controller_spec.rb
```diff
@@ -29,6 +29,24 @@
...
+    context 'an existent non-staging project' do
+      let!(:existent_project) { create(:project, name: "#{project}:new-staging", maintainer: user) }
+
+      before do
+        post :create, params: { staging_workflow_id: staging_workflow.id, staging_project_name: existent_project.name }
+      end
+
+      subject { staging_workflow }
+
+      it { expect { subject }.not_to change(Project, :count) }
```
bgeuken: This would only make sense if `subject` would be the post request. Right now this just tests that creating a staging workflow factory is not creating a new project.

dmarcoux: True!

To not rewrite all the specs in this file, I'll go with `it { expect(Project.count).to eq(3) }`.

src/api/spec/controllers/webui/staging/projects_controller_spec.rb
```diff
@@ -29,6 +29,24 @@
       end
     end
 
+    context 'an existent non-staging project' do
+      let!(:existent_project) { create(:project, name: "#{project}:new-staging", maintainer: user) }
+
+      before do
+        post :create, params: { staging_workflow_id: staging_workflow.id, staging_project_name: existent_project.name }
```
Ana06: [nitpick] `.name` should be needed

src/api/app/models/concerns/staging_project.rb
```diff
@@ -9,6 +9,7 @@ module StagingProject
     after_save :update_staging_workflow_on_backend, if: :staging_project?
     after_destroy :update_staging_workflow_on_backend, if: :staging_project?
     before_create :add_managers_group, if: :staging_project?
+    before_update :add_managers_group, if: proc { |project| project.staging_workflow_id_changed? && project.staging_workflow_id_was.nil? }
```
Ana06: [nipick] equivalente to `{ |project| project.staging_workflow_id && project.staging_workflow_id_was.nil? }`

dmarcoux:
@bgeuken Please have another look

PR #7163, opened 2019-03-12
Just admins are allowed to run Repositories#create_flag

vpereira:
To keep inline with the api ```SourceController#project_command_set_flag```
just allow admin users to create flags _sourceaccess_ and _access_ (boo#1125782)

### DIFF
--- a/src/api/app/policies/flag_policy.rb
+++ b/src/api/app/policies/flag_policy.rb
@@ -1,0 +1,6 @@
+class FlagPolicy < ApplicationPolicy
+  # just admin are able to create sourceaccess and access flags
+  def create?
+    user.is_admin? || ['sourceaccess', 'access'].exclude?(record.flag)
+  end
+end
--- a/src/api/app/controllers/webui/repositories_controller.rb
+++ b/src/api/app/controllers/webui/repositories_controller.rb
@@ -189,8 +189,8 @@ class Webui::RepositoriesController < Webui::WebuiController
   # POST flag/:project(/:package)
   def create_flag
     authorize @main_object, :update?
-
     @flag = @main_object.flags.new(status: params[:status], flag: params[:flag])
+    authorize @flag, :create?
     @flag.architecture = Architecture.find_by_name(params[:architecture])
     @flag.repo = params[:repository] if params[:repository].present?
 


src/api/app/policies/flag_policy.rb
```diff
@@ -0,0 +1,8 @@
+class FlagPolicy < ApplicationPolicy
+  # just admin are able to create flag
+  def create_flag?
```
hennevogel: `create?` please

src/api/app/policies/flag_policy.rb
```diff
@@ -0,0 +1,8 @@
...
+    return true if user.is_admin?
+    return false if ['sourceaccess', 'access'].include?(record.flag)
```
hennevogel: If you turn this around you don't need the extra true.

src/api/app/policies/flag_policy.rb
```diff
@@ -0,0 +1,7 @@
+class FlagPolicy < ApplicationPolicy
+  # just admin are able to create flag
+  def create?
+    return true if user.is_admin?
+    return true unless ['sourceaccess', 'access'].include?(record.flag)
```
coolo:
```
return true unless X
```

at the end of a (ruby) function is basically long for 

```
!X
```

src/api/app/policies/flag_policy.rb
```diff
@@ -0,0 +1,7 @@
+class FlagPolicy < ApplicationPolicy
+  # just admin are able to create flag
```
coolo: this is false and misleading. Regular users can't *add* '*access' flags to existant projects. Creating them is allowed for everyone.

src/api/app/policies/flag_policy.rb
dmarcoux: The current code could return `nil`, but I don't see why it shouldn't be a boolean instead.

It could also be rewritten this way:

```
user.is_admin? || ['sourceaccess', 'access'].exclude?(record.flag)
```

vpereira:
@hennevogel could you please check it again?

coolo:
This is still wrong. Because the policy is not on the flag, but on the project. It should be add_flag? on package and project. You can create flags just fine when creating the objects.

hennevogel:
> Because the policy is not on the flag, but on the project.

This check is 2 lines above.

src/api/app/policies/flag_policy.rb
```diff
@@ -0,0 +1,6 @@
+class FlagPolicy < ApplicationPolicy
+  # just admin are able to create sourceaccess and access flags
+  def create?
+    user.is_admin? || ['sourceaccess', 'access'].exclude?(record.flag)
```
hennevogel: Cool, didn't know `.exclude`...

PR #7167, opened 2019-03-12
List excluded requests via the API

eduardoj:
Also add documentation in apidocs.

### DIFF
--- a/src/api/config/routes.rb
+++ b/src/api/config/routes.rb
@@ -769,6 +769,7 @@ OBSApi::Application.routes.draw do
       resource :staged_requests, controller: 'staging/staged_requests', only: [:create, :destroy], constraints: cons
     end
 
+    resources :excluded_requests, controller: 'staging/excluded_requests', only: [:index], constraints: cons
     resource :excluded_requests, controller: 'staging/excluded_requests', only: [:create, :destroy], constraints: cons
   end
 
--- a/src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
+++ b/src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 RSpec.describe Staging::ExcludedRequestsController, type: :controller, vcr: true do
+  render_views
+
   let(:user) { create(:confirmed_user, login: 'user') }
   let(:other_user) { create(:confirmed_user, login: 'other_user') }
   let(:project) { user.home_project }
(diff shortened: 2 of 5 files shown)


eduardoj:
[List excluded requests via the API](https://trello.com/c/3eZRxN8a/1078-list-excluded-requests-via-the-api)

src/api/config/routes.rb
```diff
@@ -769,6 +769,7 @@ def self.public_or_about_path?(request)
       resource :staged_requests, controller: 'staging/staged_requests', only: [:create, :destroy], constraints: cons
     end
 
+    resources :excluded_requests, controller: 'staging/excluded_requests', only: [:index], constraints: cons
```
Ana06: there is a conflict with https://github.com/openSUSE/open-build-service/pull/7161/commits/0dc1b55cf2866a6c39d7e2714d78d71dd4d811ff (just so that you take it into account if the other PR gets merge first)

src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
```diff
@@ -18,9 +20,22 @@
            review_by_group: group)
   end
 
-  before { login(manager) }
+  describe 'GET #index' do
```
Ana06: shouldn't you create excluded requests to be able to test this properly?

src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
```diff
@@ -18,9 +20,22 @@
...
+    before do
+      get :index, params: { staging_main_project_name: staging_workflow.project.name, format: :xml }
+    end
+
+    it { expect(response).to have_http_status(:success) }
+    it 'returns the excluded_requests xml' do
+      assert_select 'excluded_requests' do
+        assert_select 'request', 0
```
dmarcoux: I didn't see it in time, but this test is not enough. How are we sure it's working with this?

Ana06: that was exactly my previous comment, I don't understand why this got merged without that issue being discussed.

eduardoj: I will create another pull request to fix the test.

src/api/config/routes.rb
```diff
@@ -769,6 +769,7 @@ def self.public_or_about_path?(request)
...
     resource :excluded_requests, controller: 'staging/excluded_requests', only: [:create, :destroy], constraints: cons
```
Ana06: BTW, (unrelated to this PR) why is this route in singular?

src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
Ana06: thanks @eduardoj

src/api/config/routes.rb
eduardoj: > BTW, (unrelated to this PR) why is this route in singular?

For the `:destroy` route the id was not required (multiple excluded requests can be unexcluded at a time). And as the `:create` route doesn't need the id, it was decided to merge both routes in the same `resource` helper.

Now that the `:index` route is added, yes, you are right: `:index` and `:create` routes can be defined in the `resources` helper, and the `:destroy` route can be defined in the `resource` helper. I can do this in another pull request, after #7161 gets merged (the refactor of staging routes is done).

PR #7168, opened 2019-03-13
Introduce User.current_or_nobody

coolo:
There are way too many places requiring a user object but have to treat User.current being nil as a special case.

User.current_or_nobody always returns a user object and as such you can writer simpler code. I only changed examples though to validate the idea (including an alternative fix for #5351)

### DIFF
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -84,7 +84,7 @@ class Webui::WebuiController < ActionController::Base
     if CONFIG['kerberos_mode']
       kerberos_auth
     else
-      if User.current.nil? || User.current.is_nobody?
+      if User.current_or_nobody.is_nobody?
         render(text: 'Please login') && (return false) if request.xhr?
 
         flash[:error] = 'Please login to access the requested page.'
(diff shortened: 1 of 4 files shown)


dmarcoux:
It's less drastic as the previous attempt that @Ana06 commented about.

vpereira:
It is more or less what ```devise``` does, https://github.com/plataformatec/devise/wiki/How-To:-Create-a-guest-user

coolo:
The authorization part isn't even touched - I never said User.current should be removed. That was @Ana06

dmarcoux:
I understand your concerns @hennevogel, but @coolo is right, nothing in the authorization part is touched. I think the PR adds a better way to do something we already do it in multiple places. It wouldn't remove the need for `User.current`.

Ana06:
@coolo 

> The authorization part isn't even touched - I never said User.current should be removed. That was @Ana06

I said I would like it to be removed, not that we should do it now. I like the approach. It is closed enough to removing it without risking to break everything.

hennevogel:
> nothing in the authorization part is touched. 

Not yet. People for sure will use `User.current_or_nobody`.

coolo:
So your 'requested changes' is with the people not with this PR.

hennevogel:
> So your 'requested changes' is with the people not with this PR.

My requested change is to understand what is going on before merging this.

coolo:
I understand what is going on - can we continue?

hennevogel:
> I understand what is going on - can we continue?

Are we a bit feisty today? I like it! :cowboy_hat_face: 

Document it (at least in a comment above the method) or it did not happen.

coolo:
> Document it (at least in a comment above the method) or it did not happen.
Now we talk!

PR #7171, opened 2019-03-13
Replace user_name_with_icon by user_with_realname_and_icon

### LINKED ISSUE #7164: [production][request#show] undefined method `name' for nil:NilClass
request#show threw this 14 times
<pre>
undefined method `name&#39; for nil:NilClass
</pre>
[https://build.opensuse.org/request/show/684243](https://build.opensuse.org/request/show/684243)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5c878ae5f0a89200086524f2)

dmarcoux:
`user_name_with_icon` was introduced in f246132 when we should have instead used `user_with_realname_and_icon` since they both achieve the same result.

They really do the same thing...
same-result

### DIFF
--- a/src/api/app/datatables/user_configuration_datatable.rb
+++ b/src/api/app/datatables/user_configuration_datatable.rb
@@ -1,6 +1,6 @@
 class UserConfigurationDatatable < Datatable
   def_delegator :@view, :user_actions
-  def_delegator :@view, :user_name_with_icon
+  def_delegator :@view, :user_with_realname_and_icon
 
   def view_columns
     @view_columns ||= {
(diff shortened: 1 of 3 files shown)


dmarcoux:
The current changes will make the crashing stop... but it doesn't fix the root cause.

dmarcoux:
CircleCI passed: https://circleci.com/workflow-run/6c319bf3-9f51-4354-b1aa-fda881021087

I'm not sure if we should merge this without fixing the root cause... it's causing a lot of exceptions in production, but with the current changes, it's just hiding the problem. Instead of displaying the user and his/her icon, we display nothing.

dmarcoux:
I will still do this cleanup, but the fix for the issue is now in #7171. I will wait for that PR to be merged before rebasing.

dmarcoux:
I rebased on master following the merge of #7177.

dmarcoux:
CI is green: https://circleci.com/workflow-run/a36a8ab4-5002-4a58-8d66-44e32d60fc98

PR #7172, opened 2019-03-13
Set callbacks for staging project on project model

saraycp:
There where some callbacks specifically defined for staging projects
that have been moved to the project model. So they can be called even
if we perform some actions treating a staging project as regular
project.

For example, we can now delete a staging project via API using the same
endpoint that we use for any other kind of project `/source/<project>` and 
the right callbacks will be triggered.

Also updates apidocs.

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -16,6 +16,8 @@ class Project < ApplicationRecord
 
   before_destroy :cleanup_before_destroy
   after_destroy_commit :delete_on_backend
+  after_save :update_staging_workflow_on_backend
+  after_destroy :update_staging_workflow_on_backend
 
   after_save :discard_cache
   after_rollback :reset_cache
@@ -1704,6 +1706,13 @@ class Project < ApplicationRecord
 
   def calculate_missing_checks
     combined_status_reports.map(&:missing_checks).flatten
+  end
+
+  def update_staging_workflow_on_backend
+    return unless staging_workflow_id
+
+    staging_workflow = Staging::Workflow.find(staging_workflow_id)
+    staging_workflow.write_to_backend
   end
 end
 # rubocop:enable Metrics/ClassLength
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -283,7 +283,8 @@ XmlResult: status
 DELETE /source/<project>
 
   Deletes specified project. All packages of this project are deleted as if a
-  DELETE request were issued for each package.
+  DELETE request were issued for each package. If it is a staging project, its
+  associated staging workflow will be also updated in the backend.
 
 Parameters:
   force: If force = 1, the project is deleted even if repositories of other
--- a/src/api/app/models/staging/staging_project.rb
+++ b/src/api/app/models/staging/staging_project.rb
@@ -5,8 +5,6 @@ module Staging
 
     default_scope { where.not(staging_workflow: nil) }
 
-    after_save :update_staging_workflow_on_backend
-    after_destroy :update_staging_workflow_on_backend
     before_create :add_managers_group
 
     def copy(new_project_name)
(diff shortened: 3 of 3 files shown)


saraycp:
[Allow to Destroy/Remove a Staging Project via the API](https://trello.com/c/2zIIt0hB/1076-allow-to-destroy-remove-a-staging-project-via-the-api)

vpereira:
Hi @saraycp !

The task to update https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/source_project_controller.rb#L58 to use ```Pundit``` and have ```rspec``` coverage for that is out of the scope of this PR/card?

src/api/app/models/project.rb
```diff
@@ -16,6 +16,8 @@ class Project < ApplicationRecord
 
   before_destroy :cleanup_before_destroy
   after_destroy_commit :delete_on_backend
+  after_save :update_staging_workflow_on_backend
```
dmarcoux: Not having to create another API endpoint to delete a project is nice, but I don't like that we now have code specific to the staging workflow in the project... This model is already humongous.

vpereira: I wrote it somewhere here, we should avoid to have in controller or models   conditions like ```if(@prj.is_a?(StagingProject) ...``` or leak subclasses methods to it's parent class.. I would prefer to keep everything separated until we have a clear view of what is coming ahead

saraycp: I agree with you two, I don't like it either as I mentioned in [this comment](https://github.com/openSUSE/open-build-service/pull/7158#issuecomment-472073436).

However, if we keep it independent, the callbacks won't be triggered if we delete the staging project from the [project show page](https://build.opensuse.org/project/show/TestLinux:Staging:A) or from the old API endpoint `/source/<project>`. I mean, as a regular project.

I wonder if it's better to have that ugly code or to keep it independent and, in consequence , to deal with:

- providing 2 API endpoints (returning an error if someone tries to delete a staging project with the old endpoint)
- changing the link in the UI depending on the class.

@dmarcoux  @vpereira :thinking:

coolo: So let me risk being mob-murdered: How about redefine StagingProject not as subclass of Project, but as a ProjectRole. A Project can have one role - which is a polymorphic model and can attach it to e.g. staging workflows. And these role models have callsback as your staging project has.

hennevogel: > How about redefine StagingProject not as subclass of Project, but as a ProjectRole.

We already have this. It's called `Project.kind`

coolo: this just a string. I'm talking about a model that you can delegate to.

vpereira: hi @saraycp , For me, we should keep the callback in the StagingProject model, have its own endpoint as described in the deliverable and re-think it later. If necessary in the source controller we do something like:

````
 raise NoCookiesForYouError if @project.is_a?(StagingProject)
````

To avoid that people try to delete StagingProject through the false endpoint (which wont happen, people will use the  ```osc``` commands available)

IMO, This PRCDD (Pull Request Comment Driven Design) (TM) isn't the way to go. We should refine the features somewhere else.

coolo: remember you also need this is_a? in the view, And as they are after_save callbacks, you obviously also need them in every project editing happening.

vpereira: Hey @coolo!

> remember you also need this is_a? in the view, And as they are after_save callbacks, you obviously also need them in every project editing happening.

Yes, maybe, I'm not sure if it is needed, but maybe it is better to do something small like that, than start to redesign things before even ship the staging workflow

hennevogel: > start to redesign things before even ship the staging workflow

But this is exactly what we are doing here....

vpereira: > But this is exactly what we are doing here....

In the PR? absolutely, but it would be better if we don't...

hennevogel: > but it would be better if we don't...

So where else are you going to do this?

vpereira: as we are doing now, in our existing process (daily scrum) :)

saraycp:
I close this PR also, we are going to refactor everything. See you soon....

PR #7176, opened 2019-03-13
Fix "index" action tests for excluded requests

eduardoj:
Before it was only checked that an empty list was returned when there
weren't any exclusions. Now it is checked that with one exclusion a list
of one element is returned.

Ending work started in #7167.

### DIFF
--- a/src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
+++ b/src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
@@ -21,14 +21,28 @@ RSpec.describe Staging::ExcludedRequestsController, type: :controller, vcr: true
   end
 
   describe 'GET #index' do
+    let!(:request_exclusion_1) { create(:request_exclusion, bs_request: bs_request, staging_workflow: staging_workflow, description: 'Request 1') }
+    let(:source_package_2) { create(:package, name: 'source_package_2', project: source_project) }
+    let(:bs_request_2) do
+      create(:bs_request_with_submit_action,
+             creator: other_user,
+             target_package: target_package,
+             source_package: source_package,
+             review_by_group: group)
+    end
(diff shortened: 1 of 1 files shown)


src/api/spec/controllers/staging/excluded_requests_controller_spec.rb
```diff
@@ -21,14 +21,16 @@
   end
 
   describe 'GET #index' do
+    let!(:request_exclusion) { create(:request_exclusion, bs_request: bs_request, number: bs_request.number, staging_workflow: staging_workflow) }
```
Ana06: why do you need `number`?

eduardoj: `number` removed.

eduardoj:
> what about creating more than one? what about checking that the element is actually your excluded request? :wink:

I created the test based code written in ebb64db2c5e0. Should I create more than one test, and check that the element is actually the element meant to be in both tests, this one and staged requests?

Ana06:
@eduardoj 

> I created the test based code written in ebb64db. Should I create more than one test, and check that the element is actually the element meant to be in both tests, this one and staged requests?

no :rofl: I didn't meant you should write more than test, but to create more than one request for this test. To have a list you need at least 2 elements :wink:

eduardoj:
> @Ana06:
>
> no :rofl: I didn't meant you should write more than test, but to create more than one request for this test. To have a list you need at least 2 elements :wink:

I wanted to say "request" instead of "test". I modified my comment.

I added a second excluded request to this test, and checked for the number of elements returned in the xml, and for specific values of the descriptions. Staged request index test should be fixed too, then.

eduardoj:
[List excluded requests via the API](https://trello.com/c/3eZRxN8a/1078-list-excluded-requests-via-the-api)

Ana06:
Tests are working: https://circleci.com/gh/eduardoj/open-build-service/1504

PR #7178, opened 2019-03-13
 Create staging projects via API

DavidKang:
- Implemented API (added create action and route).
- Added controller and service tests.
- Added API documentation.

### DIFF
--- a/src/api/app/controllers/staging/staging_projects_controller.rb
+++ b/src/api/app/controllers/staging/staging_projects_controller.rb
@@ -19,8 +19,7 @@ class Staging::StagingProjectsController < ApplicationController
 
   def create
     authorize @staging_workflow
-    xml_request = (Xmlhash.parse(request.body.read) || {}).with_indifferent_access
-    result = ::Staging::StagingProjectCreator.new(xml_request[:staging_project], @staging_workflow).call
+    result = ::Staging::StagingProjectCreator.new(request.body.read, @staging_workflow).call
 
     if result.valid?
       render_ok
--- a/src/api/app/controllers/staging/staging_projects_controller.rb
+++ b/src/api/app/controllers/staging/staging_projects_controller.rb
@@ -2,6 +2,7 @@ class Staging::StagingProjectsController < ApplicationController
   before_action :require_login, except: [:index, :show]
 
   before_action :set_main_project
+  before_action :set_staging_workflow, only: :create
 
   def index
     if @main_project.staging
@@ -14,6 +15,22 @@ class Staging::StagingProjectsController < ApplicationController
 
   def show
     @staging_project = @main_project.staging.staging_projects.find_by!(name: params[:name])
+  end
+
+  def create
+    authorize @staging_workflow
+    xml_request = (Xmlhash.parse(request.body.read) || {}).with_indifferent_access
+    result = ::Staging::StagingProjectCreator.new(xml_request[:staging_project], @staging_workflow).call
+
+    if result.valid?
+      render_ok
+    else
+      render_error(
+        status: 400,
+        errorcode: 'invalid_request',
+        message: "Staging Projects for #{@main_project} failed: #{result.errors.join(' ')}"
+      )
+    end
   end
 
   def copy
@@ -45,5 +62,9 @@ class Staging::StagingProjectsController < ApplicationController
 
   def set_main_project
     @main_project = Project.find_by!(name: params[:staging_workflow_project])
+  end
+
+  def set_staging_workflow
+    @staging_workflow = @main_project.staging
   end
 end
--- a/src/api/app/controllers/staging/staging_projects_controller.rb
+++ b/src/api/app/controllers/staging/staging_projects_controller.rb
@@ -1,8 +1,9 @@
 class Staging::StagingProjectsController < ApplicationController
   before_action :require_login, except: [:index, :show]
-
   before_action :set_main_project
   before_action :set_staging_workflow, only: :create
+
+  validate_action create: { method: :post, request: :staging_project }
 
   def index
     if @main_project.staging
--- a/src/api/app/services/staging/staging_project_creator.rb
+++ b/src/api/app/services/staging/staging_project_creator.rb
@@ -1,0 +1,40 @@
+module Staging
+  class StagingProjectCreator
+    def initialize(staging_project_names, staging_workflow)
+      @staging_project_names = staging_project_names
+      @staging_workflow = staging_workflow
+    end
+
+    def call
+      staging_projects = []
+      staging_project_names.each do |name|
+        project = Project.find_or_initialize_by(name: name)
+
+        project_validator = StagingProjectValidator.new(project).call
+        if project_validator.valid?
+          project.staging_workflow_id = @staging_workflow.id
+          staging_projects << project
+        else
+          errors << project_validator.errors
+        end
+      end
+
+      staging_projects.each(&:store) if valid?
+      self
+    end
+
+    def errors
+      @errors ||= []
+    end
+
+    def valid?
+      errors.empty?
+    end
+
+    private
+
+    def staging_project_names
+      [@staging_project_names].flatten
+    end
+  end
+end
--- a/src/api/app/services/staging/staging_project_creator.rb
+++ b/src/api/app/services/staging/staging_project_creator.rb
@@ -10,19 +10,8 @@ module Staging
         errors << 'Staging projects are empty'
         return self
       end
-      staging_projects = Xmlhash.parse(@request_body).elements('staging_project').collect do |name|
-        project = Project.find_or_initialize_by(name: name)
-        project_validator = StagingProjectValidator.new(project).call
-
-        unless project_validator.valid?
-          errors << project_validator.errors
-          next
-        end
-
-        project.staging_workflow_id = @staging_workflow.id
-        project
-      end
-      staging_projects.each(&:store) if valid?
+      projects = staging_projects
+      projects.each(&:store) if valid?
       self
     end
 
@@ -34,10 +23,19 @@ module Staging
       errors.empty?
     end
 
-    private
+    def staging_projects
+      Xmlhash.parse(@request_body).elements('staging_project').collect do |name|
+        project = Project.find_or_initialize_by(name: name)
+        project_validator = StagingProjectValidator.new(project).call
 
-    def staging_project_names
-      [@staging_project_names].flatten
+        unless project_validator.valid?
+          errors << project_validator.errors
+          next
+        end
+
+        project.staging_workflow_id = @staging_workflow.id
+        project
+      end.compact
     end
   end
 end
--- a/src/api/app/services/staging/staging_project_creator.rb
+++ b/src/api/app/services/staging/staging_project_creator.rb
@@ -7,10 +7,10 @@ module Staging
 
     def call
       staging_projects = []
-      staging_project_names.each do |name|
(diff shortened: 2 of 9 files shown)


src/api/app/services/staging/staging_project_creator.rb
```diff
@@ -0,0 +1,47 @@
+module Staging
+  class StagingProjectCreator
+    def initialize(xml, staging_workflow)
+      @xml_hash = (Xmlhash.parse(xml) || {}).with_indifferent_access
```
hennevogel:
```ruby
Xmlhash.parse(nil)
        1: from (irb):1
TypeError (wrong argument type nil (expected String))
```

src/api/app/services/staging/staging_project_creator.rb
```diff
@@ -0,0 +1,47 @@
...
+      @staging_workflow = staging_workflow
+    end
+
+    def create
+      staging_projects = []
+      staging_project_names.each do |name|
+        project = Project.find_or_initialize_by(name: name)
+        if project.staging
+          errors << "Project \"#{project}\": is a main project of a staging workflow."
```
hennevogel: ~~Project \"#{project}\": is a main project of a staging workflow.~~

Project \"#{project}\": is has a staging already. Nested stagings are not supported.

src/api/app/services/staging/staging_project_creator.rb
```diff
@@ -0,0 +1,45 @@
+module Staging
+  class StagingProjectCreator
+    def initialize(staging_project_names, staging_workflow)
+      @staging_project_names = staging_project_names
+      @staging_workflow = staging_workflow
+    end
+
+    def create
```
bgeuken: Is there any reason to distinguish here between create and initialize? It looks like this is basically the same, or not?

src/api/app/services/staging/staging_project_creator.rb
```diff
@@ -0,0 +1,45 @@
...
+      staging_projects = []
+      staging_project_names.each do |name|
+        project = Project.find_or_initialize_by(name: name)
+
+        if project.valid? && project.staging.nil? && !project.staging_workflow_id?
```
bgeuken: Looks like this is duplicated in line 39. Would be nice if we could merge this.

src/api/app/services/staging/staging_project_creator.rb
```diff
@@ -0,0 +1,45 @@
...
+  class StagingProjectCreator
+    def initialize(staging_project_names, staging_workflow)
+      @staging_project_names = staging_project_names
+      @staging_workflow = staging_workflow
+    end
+
+    def create
+      staging_projects = []
+      staging_project_names.each do |name|
+        project = Project.find_or_initialize_by(name: name)
```
bgeuken: Is there any permission check that prevents users from creating projects they shouldn't be allowed to? :thinking:

bgeuken:
What's the reasoning for using a `service` for this? It looks like all the code there is build around model related stuff, creating objects, validating them, ....

src/api/app/services/staging/staging_project_creator.rb
DavidKang: Nop, you are right!

DavidKang:
@bgeuken, my reason for using a `service` for this is to abstract the logic from the controller. Also, this service could be reused if we want.

src/api/app/services/staging/staging_project_creator.rb
```diff
@@ -0,0 +1,40 @@
+module Staging
+  class StagingProjectCreator
+    def initialize(staging_project_names, staging_workflow)
+      @staging_project_names = staging_project_names
+      @staging_workflow = staging_workflow
+    end
+
+    def perform
```
vpereira: why don't call it ```call``` which is "standard" in the service objects pattern in ruby? I don't think that ```perform``` is a bad name, but if you look all the others services that we have, I think all of them have similar API. 

One advantage to have it named to ```call``` is to be able to stub them with simple procs as in https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/controllers/webui/projects/meta_controller_spec.rb#L29 making easier to get rid of VCR cassettes..

DavidKang: @vpereira, I will change the name of the method.

coolo: > why don't call it `call` which is "standard" in the service objects pattern in ruby? I don't think that `perform` is a bad name, but if you look all the others services that we have, I think all of them have similar API.

I see a mixed picture there:
```
coolo@f102#api>git grep 'def call' app | wc -l
7
coolo@f102#api>git grep 'def perform' app | wc -l
28
```

Looks like all the calls are @vpereira 's :)

coolo: https://hackernoon.com/service-objects-in-ruby-on-rails-and-you-79ca8a1c946e uses 'perform' while
https://medium.freecodecamp.org/service-objects-explained-simply-for-ruby-on-rails-5-a8cc42a5441f uses 'call' - I don't think there is something like a standard outside of OBS. And in OBS the standard I would say is 'perform'.

vpereira: > I see a mixed picture there:
> 
> ```
> coolo@f102#api>git grep 'def call' app | wc -l
> 7
> coolo@f102#api>git grep 'def perform' app | wc -l
> 28
> ```

@coolo looks like you are confusing jobs with services.. do you need any help?

```
vpereira@linux-y5ui:~/open-build-service/src/api/app/services> git grep "def call" | wc -l
6
vpereira@linux-y5ui:~/open-build-service/src/api/app/services> git grep "def perform" | wc -l
0
vpereira@linux-y5ui:~/open-build-service/src/api/app/services> 

```

coolo: Ah, that explains it. Thanks

vpereira: > https://hackernoon.com/service-objects-in-ruby-on-rails-and-you-79ca8a1c946e uses 'perform' while
> https://medium.freecodecamp.org/service-objects-explained-simply-for-ruby-on-rails-5-a8cc42a5441f uses 'call' - I don't think there is something like a standard outside of OBS. And in OBS the standard I would say is 'perform'.

@coolo is really hard to argument here without folling in the "argument from authority" fallacy. But let's try:

https://www.toptal.com/ruby-on-rails/rails-service-objects-tutorial
https://medium.freecodecamp.org/service-objects-explained-simply-for-ruby-on-rails-5-a8cc42a5441f
http://chrisholtz.com/blog/organize-your-app-with-service-objects/

coolo: I quote from there:

> The call method

> These services are invoked via the call() method. This method could be named anything - perform() > and process() are workable alternatives. However, I use call() because it is the same method name > used by ruby Proc and Lambda objects to invoke their content. This becomes more important when > applying functional programming principals to service objects. Also, call() can be invoked via 
> shortcut notation:

This gives enough reason.

DavidKang: @vpereira, changed!

vpereira:
can you please review it again @hennevogel  and see if your concerns were addressed, please?

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
```diff
@@ -139,4 +139,44 @@
...
+            <staging_project>#{project}:Staging:A</staging_project>
+            <staging_project>#{project}:Staging:E</staging_project>
+          </workflow>
+        XML
+      end
+
+      it { expect(subject).to have_http_status(:bad_request) }
+      it { expect { subject }.not_to change(Project, :count) }
+    end
+  end
```
hennevogel: What happens if I don't send a body?

bgeuken: This would have caused a failure. It's fixed now.

vpereira: the ```StagingProjectCreator``` won't be valid, and therefore, it will raise a ```bad_request```

src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -16,6 +17,22 @@ def show
     @staging_project = @main_project.staging.staging_projects.find_by!(name: params[:name])
   end
 
+  def create
+    authorize @staging_workflow
+    xml_request = (Xmlhash.parse(request.body.read) || {}).with_indifferent_access
+    result = ::Staging::StagingProjectCreator.new(xml_request[:staging_project], @staging_workflow).call
```
coolo: This is wrong. and you only failed to notice because all your tests have 2 staging projects.

The [] operator returns strings if there is only one element (and should only be used for attributes really)

```
irb(main):010:0> Xmlhash.parse('<workflow><staging_project>example2</staging_project></workflow>')['staging_project']
=> "example2"
```

Prefer elements if you want to ensure an array.

```
irb(main):011:0> Xmlhash.parse('<workflow><staging_project>example2</staging_project></workflow>').elements('staging_project')
=> ["example2"]
```

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
```diff
@@ -139,4 +139,51 @@
...
+        <<~XML
+          <workflow>
+            <staging_project>#{project}:Staging:C</staging_project>
+            <staging_project>#{project}:other_project</staging_project>
+          </workflow>
+        XML
+      end
+
+      it { expect(subject).to have_http_status(:success) }
+      it { expect { subject }.to change(Project, :count).by(1) }
```
coolo: IMO posting 2 names to create should either create 2 or error. Because if posting existing staging projects is fine, I would expect missing stagings to be removed (or at least unlinked - how do I unlink a project btw?).

DavidKang: @coolo, at the moment, is not possible to unlink a project from a staging, but you are right we should provide a way to unlink

dmarcoux:
We need to keep in mind to add the creation of the staging project to its history (#7214)

src/api/app/services/staging/staging_project_creator.rb
```diff
@@ -0,0 +1,43 @@
+module Staging
+  class StagingProjectCreator
+    def initialize(request_body, staging_workflow)
+      @request_body = request_body
+      @staging_workflow = staging_workflow
+    end
+
+    def call
+      if @request_body.blank?
```
coolo: That's a neat case of Test Driven Development, but it's hardly the only way to post invalid input.

src/api/app/services/staging/staging_project_creator.rb
```diff
@@ -0,0 +1,43 @@
...
+        errors << 'Staging projects are empty'
+        return self
+      end
+      staging_projects = Xmlhash.parse(@request_body).elements('staging_project').collect do |name|
```
coolo: make sure the controller adds a validation on the input - otherwise the parse here might return nil.

The good news: we already have a validator builtin. Use validate_action

vpereira: done, can you please check it @coolo

src/api/app/services/staging/staging_project_creator.rb
coolo: if you added XML validation, you don't need this anymore.

PR #7180, opened 2019-03-13
Update rails to version 5.1.6.2

hennevogel:


hennevogel:
We need to package the gems.

hennevogel:
Let's use the depfu PR...

PR #7188, opened 2019-03-14
Drop staging project inheritance

dmarcoux:
The Staging::StagingProjet model is now a concern which is included in the Project model. This will allow us to rely on the Project endpoints, avoid code duplication and remove the confusion around the two models.


dmarcoux:
Regarding the CodeClimate issues, we are not sure whether we should tackle them or not. We moved the code and it could be refactored more in depth, but it's not needed to drop the inheritance. It's more like a nice-to-have.

coolo:
What I wonder though: now every project has an overall_state - but it's really specific to staging. Perhaps prefix such non-private functions with staging_ ?

hennevogel:
>  we are not sure whether we should tackle them or not. 

It's like vacuum-cleaning your flat. You can postpone it, but at some point in time you will walk onto this piece of lego and then....

vpereira:
excellent job @saraycp @DavidKang and @dmarcoux !

vpereira:
> It's like vacuum-cleaning your flat. You can postpone it, but at some point in time you will walk onto this piece of lego and then....

I agree with @dmarcoux. If we have time to fix "unrelated code" then maybe would be better to invest it  on the source controller endpoint (authorization with pundit and rspecs coverage)?

coolo:
I'm not even sure it can be labeled 'fixing' - the functions are a little complex, but there are far bigger lego pieces in the code base (well, in this context the smaller ones hurt more, but you get the point).

And I would so love to see you trying a rspec for _work_flow file actually changing in the backend with cassette faking. :smiling_imp:

src/api/spec/models/project_spec.rb
```diff
@@ -630,5 +630,273 @@ def reset_project_in_backend
...
+          expect(path_elements_of_project.where(link: staging_project.repositories)).not_to exist
+          expect(path_elements_of_project.where(link: subject.repositories)).to exist
+        end
+
+        it 'renames the repository link if it contains a reference to the project' do
+          expect(PathElement.find_by(parent_id: subject.repositories, link: subject.repositories).link.name).to eq("home_#{user}_new_project_sles_12")
+        end
+      end
+    end
+  end
```
bgeuken: Let's move this to a separate spec in [spec/models/project](https://github.com/openSUSE/open-build-service/tree/master/src/api/spec/models/project). This spec file is already longer than it should be.

bgeuken:
I am a bit split about this one... I like that we re-arange the code around the staging project model. But I don't like that because we include the StagingProject module every project instance has all the methods (as in during runtime).
So codewhise it's still nicely separated, but not for the created objects themself.

The other issue I realized after reading up about `concerns` again (e.g. reading [this blog](https://medium.com/@carlescliment/about-rails-concerns-a6b2f1776d7d)) is that concerns usually provide some aspect / behaviour that can be shared (isn't bound to a particular class), like with the `checkable`. That doesn't seem to be the case here.

tldr;

I am not saying that we can not merge this or that I have a better solution at hand right now, but I think we need to define for our self how we solve such kind of problems in general. Like when do we use a mixin, a service, a concern or something else.

src/api/spec/models/project_spec.rb
saraycp: @bgeuken we already moved staging project tests to its own file.

saraycp:
[Drop staging project inheritance](https://trello.com/c/k7uVqaBs/1088-drop-staging-project-inheritance)

saraycp:
[Drop staging project inheritance](https://trello.com/c/k7uVqaBs/1088-drop-staging-project-inheritance)

PR #7201, opened 2019-03-15
Disable Metrics/BlockLength for specs

coolo:
Rspec.describe is one block and they tend to get long

@openSUSE/obs-frontend please vote


src/api/.rubocop_todo.yml
```diff
@@ -200,15 +199,15 @@ Lint/UriEscapeUnescape:
...
 Metrics/AbcSize:
   Max: 239
 
-# Offense count: 603
+# Offense count: 68
 # Configuration parameters: CountComments, ExcludedMethods.
 # ExcludedMethods: refine
 Metrics/BlockLength:
-  Max: 457
+  Max: 212
```
eduardoj: [Nitpick]: This Max value got changed automatically when rubocop with autocorrect was run?

src/api/.rubocop.yml
```diff
@@ -35,6 +35,7 @@ Style/UnneededInterpolation:
 Metrics/BlockLength:
   Exclude:
     - 'config/routes.rb'
+    - 'spec/**/*'
```
Ana06: I think a comment with the reason would be nice (just that this is because the `describe`)

coolo:
It's a little ironic that the OBS build failued due to the failure whose fix triggered the rubocop complain I'm 'fixing' here.

Issue #7204, opened 2019-03-15
Flickering test ./spec/bootstrap/features/webui/attributes_spec.rb:52


coolo:
As we don't reset attribute names for rspec (due to some of them required as seeds), we run into test data collusion.

I can reproduce the problem by running:

```
while rspec ./spec/bootstrap/features/webui/attributes_spec.rb:52; do echo; done
```

Resulting in:

```
Failures:

  1) Bootstrap_Attributes with an attribute for a project with a package add attribute with values
     Failure/Error: expect(page).to have_content('Attribute was successfully created.')
[...]

PR #7208, opened 2019-03-17
 Don't treat AttribNamespace and AttribType as static

coolo:
IMO the better fix for #7204


coolo:
In this case the missing circleci is due to me awaiting the test results before creating the PR.

coolo:
confirmed - rerunning the cirleci workflow triggers 'pending'

PR #7214, opened 2019-03-18
Add the creation of a staging project to its history

### LINKED ISSUE #7051: Staging Project History should include project creation
The history of `StagingProject` doe not include creation of the project. It should.

dmarcoux:
Closes #7051 

Screenshot_2019-04-01 Open Build Service

### DIFF
--- a/src/api/app/controllers/webui/staging/projects_controller.rb
+++ b/src/api/app/controllers/webui/staging/projects_controller.rb
@@ -26,6 +26,8 @@ module Webui
 
         if staging_project.valid? && staging_project.store
           flash[:success] = "Staging project with name = \"#{staging_project}\" was successfully created"
+          staging_project.create_project_log_entry(User.current)
+
           return
         end
 
@@ -34,6 +36,10 @@ module Webui
 
       def show
         @staging_project = @staging_workflow.staging_projects.find_by(name: params[:project_name])
+        @staging_project_log_entries = @staging_project.project_log_entries
+                                                       .where(event_type: [:staging_project_created, :staged_request, :unstaged_request])
+                                                       .includes(:bs_request)
+                                                       .order(datetime: :desc)
         @project = @staging_workflow.project
 
         @groups_hash = ::Staging::Workflow.load_groups
--- a/src/api/app/models/concerns/staging_project.rb
+++ b/src/api/app/models/concerns/staging_project.rb
@@ -129,6 +129,19 @@ module StagingProject
     staging_workflow_id.present?
   end
 
+  def create_project_log_entry(user)
+    project_log_entry = ProjectLogEntry.find_or_initialize_by(
+      project: self,
+      user_name: user.login,
+      event_type: :staging_project_created
+    )
+
+    return unless project_log_entry.new_record?
+
+    project_log_entry.datetime = Time.now
+    project_log_entry.save!
+  end
+
   private
 
   def cache_problems
--- a/src/api/app/views/webui2/webui/staging/projects/_history_elements.html.haml
+++ b/src/api/app/views/webui2/webui/staging/projects/_history_elements.html.haml
@@ -2,8 +2,8 @@
   - elements.each do |element|
     %li
       = element.event_type.humanize
-      = link_to("##{element.bs_request.number}", request_path(element.bs_request.number))
-      for package
-      %b= element.package_name
-      submitted by
-      %b= element.user_name
+      - case element.event_type
+      - when 'unstaged_request', 'staged_request'
+        = render(partial: 'unstaged_staged_request_history', locals: { history: element })
+      - when 'staging_project_created'
+        = render(partial: 'staging_project_created_history', locals: { history: element })
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -29,6 +29,10 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
     end
 
     if staging_workflow.save
+      staging_workflow.staging_projects.each do |staging_project|
+        staging_project.create_project_log_entry(User.current)
+      end
+
       flash[:success] = "Staging for #{@project} was successfully created"
       redirect_to staging_workflow_path(staging_workflow)
     else
--- a/src/api/app/views/webui2/webui/staging/projects/_staging_project_created_history.html.haml
+++ b/src/api/app/views/webui2/webui/staging/projects/_staging_project_created_history.html.haml
@@ -1,0 +1,3 @@
+%b #{time_ago_in_words(Time.at(history.datetime.to_i))} ago
+by
+%b= history.user_name
--- a/src/api/app/views/webui2/webui/staging/projects/_unstaged_staged_request_history.html.haml
+++ b/src/api/app/views/webui2/webui/staging/projects/_unstaged_staged_request_history.html.haml
@@ -1,0 +1,5 @@
+= link_to("##{history.bs_request.number}", request_path(history.bs_request.number))
+for package
+%b= history.package_name
+submitted by
+%b= history.user_name
--- a/src/api/app/views/webui2/webui/staging/projects/show.html.haml
+++ b/src/api/app/views/webui2/webui/staging/projects/show.html.haml
@@ -36,8 +36,6 @@
     .card.mb-3
       %h5.card-header History
       .card-body
-        = render partial: 'history_elements',
-                       locals: { elements: |
-                       @staging_project.project_log_entries.where(event_type: [:staged_request, :unstaged_request]).includes(:bs_request) } |
+        = render partial: 'history_elements', locals: { elements: @staging_project_log_entries }
   .col-xl-2
     = render partial: 'webui/staging/shared/legend'


dmarcoux:
[Address Feedback](https://trello.com/c/K9U6WlG1/1081-address-feedback)

dmarcoux:
I will rebase on _master_ once #7188 is merged and adapt the code (mostly moving it to the new files)

src/api/app/models/staging/staging_project.rb
```diff
@@ -8,6 +8,8 @@ class StagingProject < Project
     after_save :update_staging_workflow_on_backend
     after_destroy :update_staging_workflow_on_backend
     before_create :add_managers_group
+    after_create :create_project_log_entry
+    before_update :create_project_log_entry, if: proc { |staging_project| staging_project.staging_workflow_id_was.nil? }
```
bgeuken: You probably could drop the `.nil?` by using `unless`. It might even be possible to shorten this to `unless: :staging_workflow_id_was`

src/api/app/views/webui2/webui/staging/projects/_history_elements.html.haml
```diff
@@ -2,8 +2,14 @@
...
+      - when 'unstaged_request', 'staged_request'
+        = link_to("##{element.bs_request.number}", request_path(element.bs_request.number))
+        for package
+        %b= element.package_name
+        submitted by
+        %b= element.user_name
+      - when 'staging_project_created'
+        %b #{time_ago_in_words(Time.at(element.datetime.to_i))} ago
+        by
+        %b= element.user_name
```
bgeuken: That's more of a nitpick... I would prefer to move this to a helper. So we don't end up with a big switch condition in the view.

src/api/app/models/staging/staging_project.rb
dmarcoux: It's longer for sure, but it is also much clearer, so I prefer to leave it as it is.

src/api/app/views/webui2/webui/staging/projects/_history_elements.html.haml
dmarcoux: I was also considering if I should do it or not when I wrote the code. I will move it to a helper.

dmarcoux: Done, let me know what you think.

src/api/app/models/staging/staging_project.rb
```diff
@@ -225,5 +227,14 @@ def update_staging_workflow_on_backend
     def add_managers_group
       assign_managers_group(staging_workflow.managers_group)
     end
+
+    def create_project_log_entry
+      ProjectLogEntry.create!(
+        project: self,
+        user_name: User.current.login,
+        event_type: :staging_project_created,
```
dmarcoux: This is annoying in the specs when we need to create a few staging projects, but don't necessarily want to login (when testing _index_ actions for example)... It fails because `User.current` is `nil`

coolo: You can add a trait to the factory that adds skip_create. But if you also want the staging project in the backend, you better login in before.

dmarcoux: If possible, I would like to avoid skipping the callbacks. Let's see what the others think.

bgeuken: Solution 1. wouldn't work in other cases where `User.current` is needed. That's probably why we go with solution 2. in most of such cases.

The underlying problem is that we still have authorization code in some models. Until we solved this I would go with solution 2.

src/api/app/models/staging/staging_project.rb
bgeuken: IMO, as soon as someone knows that `if` / `unless` accepts symbols it's not. And then you have much more stuff to mentally parse.

dmarcoux:
I avoided changing all the specs by switching
```
User.current.login
```
to
```
User.current_or_nobody.login
```

I'm not a fan of this, but I don't see another way beside putting 30+ `login` calls across a bunch of spec files.

dmarcoux:
This `ProjectLogEntry` also needs to be displayd in the API, so add it after #7225 is merged.

coolo:
I don't like accepting nobody here - IMO accepting broken code for the sake of less noisy tests is wrong.

src/api/app/models/concerns/staging_project.rb
```diff
@@ -227,6 +229,15 @@ def add_managers_group
     assign_managers_group(staging_workflow.managers_group)
   end
 
+  def create_project_log_entry
+    ProjectLogEntry.create!(
+      project: self,
+      user_name: User.current_or_nobody.login, # User.current should be defined, but we don't want this to fail if it's somehow not
```
coolo: audit logs are pretty useless if they say nobody

coolo:
why is the coverage so low btw? not rendering views?

coolo:
@dmarcoux the more I think about it the more I come to this conclusion: having the model create audit logs / history is just as bad as models checking permissions. 

Basically every usage of User.current in model code is questionable. Another flag that this a bad idea is that your callback contains quite a big if.

coolo:
(And of course I tried, but the rspec is basically impossible to write sanely - but https://github.com/coolo/open-build-service/commit/da1211d0945de2b88b86a0500e11bb5ca24b6ce5 is at the edge of acceptable :)

src/api/app/models/concerns/staging_project.rb
dmarcoux: I agree... I tried it, but it's true that it's useless if they say nobody.

dmarcoux:
> @dmarcoux the more I think about it the more I come to this conclusion: having the model create audit logs / history is just as bad as models checking permissions.
> 
> Basically every usage of User.current in model code is questionable. Another flag that this a bad idea is that your callback contains quite a big if.

I agree. I need to read more on the subject. I don't want to merge bad code or having to "pollute" the specs with a bunch of `login` calls.

dmarcoux:
I ended up calling the method `create_project_log_entry` directly in the controller and passing the user to avoid the dependency on `User.current` in the model. I don't find this solution amazing because it's easy to forget to call the method, but at least, the model doesn't depend on `User.current` anymore and we don't need to change 50+ specs to "fix" them by calling `login` due to this dependency.

dmarcoux:
Ping @bgeuken :slightly_smiling_face:!

As you already had a look before, I want to be sure you had the chance to do so again before merging this PR.

src/api/app/helpers/webui/staging/project_helper.rb
```diff
@@ -6,4 +6,22 @@ def icon_for_checks(checks, missing_checks)
...
+    capture do
+      concat(link_to("##{history.bs_request.number}", request_path(history.bs_request.number)))
+      concat(' for package ')
+      concat(content_tag(:b, history.package_name))
+      concat(' submitted by ')
+      concat(content_tag(:b, history.user_name))
+    end
+  end
+
+  def staging_project_created_history(history)
```
hennevogel: Why a helper and not a partial?

coolo: helpers are much faster?

dmarcoux: They are both really small, so I think using a helper is fine in this case. I don't have a strong opinion on this, I'm also fine with using a partial so I  can change it if you think it's a blocker.

coolo: Lesson learned from refactoring repositories page: if the partial is small and repeated often, it should be a helper.

vpereira: "much faster" without benchmark is just a guess. I think the partial is easier to maintain, beside it's the "Rails way" https://apidock.com/rails/ActionView/Helpers/TextHelper/concat

vpereira:
> @dmarcoux the more I think about it the more I come to this conclusion: having the model create audit logs / history is just as bad as models checking permissions.
> 
> Basically every usage of User.current in model code is questionable. Another flag that this a bad idea is that your callback contains quite a big if.

I think the usage User.current in model is questionable too, in a pure MVC design, however we have different ways to interact with a model, than just via webui controller. Doing that in the controller, we will have to repeat the same code to WebUI, API and Background jobs (do we have any other entry point which do we have to cover?). So if we don't want to have that in the model, i.e background job, being fired by a callback, then an [Observers](https://github.com/rails/rails-observers) would make more sense to trigger Audit than direct on the controller.

coolo:
On 4/1/19 9:22 AM, Victor Pereira wrote:

> "much faster" without benchmark is just a guess. I think the partial is easier to maintain, beside it's the "Rails way"
> https://apidock.com/rails/ActionView/Helpers/TextHelper/concat
> 
How many more benchmarks do you have to see before you believe it?

Greetings, Stephan

coolo:
> On 4/1/19 9:22 AM, Victor Pereira wrote: "much faster" without benchmark is just a guess. I think the partial is easier to maintain, beside it's the "Rails way" https://apidock.com/rails/ActionView/Helpers/TextHelper/concat
> How many more benchmarks do you have to see before you believe it? Greetings, Stephan

c84084c24f4e7d0b9d46dd5d961b69582a23065c

coolo:
> I think the usage User.current in model is questionable too, in a pure MVC design, however we have different ways to interact with a model, than just via webui controller. Doing that in the controller, we will have to repeat the same code to WebUI, API and Background jobs (do we have any other entry point which do we have to cover?). So if we don't want to have that in the model, i.e background job, being fired by a callback, then an [Observers](https://github.com/rails/rails-observers) would make more sense to trigger Audit than direct on the controller.

That rails observers won't help you. They are just a different way to organize callbacks (and have been rightfully removed from rails) - you still have no User available in them. So you would need to rely on static variables just as well - which lead to the whole desaster.

What you need is an observer that is setup by controllers. The trick is to set this up without repeating code.

src/api/app/helpers/webui/staging/project_helper.rb
hennevogel: > I don't have a strong opinion on this

Me neither. If it's because of performance and it helps, so be it.

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -26,6 +26,8 @@ def create
 
         if staging_project.valid? && staging_project.store
           flash[:success] = "Staging project with name = \"#{staging_project}\" was successfully created"
+          staging_project.create_project_log_entry(User.current)
```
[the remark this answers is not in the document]
hennevogel: We already have a job `CreateProjectLogEntryJob`, but we can optimize this later.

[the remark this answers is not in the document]
dmarcoux: I agree with you and I had a look. To properly do this, I would need more time as I would like to refactor [this](https://github.com/openSUSE/open-build-service/blob/26cfc4414ed79aaf1b13ffe8f38074eb6c0e81ca/src/api/app/jobs/create_project_log_entry_job.rb#L5) to make it reusable for all types of `ProjectLogEntry`. Right now, `ProjectLogEntry.create_from` depends on having a request, which is not the case for our case here.

I will add a card to the staging workflow backlog to not forget this.

src/api/app/helpers/webui/staging/project_helper.rb
dmarcoux: I pushed changes to use a partial. Please have a look.

dmarcoux:
Please have a look again, I'm using partials now instead of methods in a helper.

coolo:
If the staging workflow is created, A and B are automatically created - and their creation is not recorded (talking about duplicating code).

But I wonder about something else now: will this history be filtered in any way (or cleaned up)?
Compare https://github.com/openSUSE/open-build-service/pull/6744#issuecomment-453145556

coolo:
BTW: the history lookup deserves an order call - they are displayed in rather interesting order.

coolo:
So as @vpereira wants to see benchmarks I went and took the 2 versions @dmarcoux coded.

The current PR code (partials):

```
Rendered webui2/webui/staging/projects/_history_elements.html.haml (418.7ms)
```

The previous version (helper)
```
Rendered webui2/webui/staging/projects/_history_elements.html.haml (227.4ms)
```

And this isn't with insane values of log entries - but with 401 entries. And without cleanup, this will be a *very* low number soonish.

src/api/app/controllers/webui/staging/projects_controller.rb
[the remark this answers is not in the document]
vpereira: great @dmarcoux!

dmarcoux:
I didn't expect this to be such a pain...

Regarding the cleanup, I agree it should be done. However, it will have to be in another PR since it also touches the other `ProjectLogEntry` records creating in other parts of the staging workflow.

dmarcoux:
> If the staging workflow is created, A and B are automatically created - and their creation is not recorded

Done

dmarcoux:
I created a [card](https://trello.com/c/SeW31Qda/1673-cleanup-a-staging-project-history-when-it-is-merged) regarding the cleanup of the history

coolo:
User.current strikes back - I like it :)

PR #7215, opened 2019-03-18
Complexity improvements

saraycp:
Tries to address code climate issued for  #7188


saraycp:
Once we checked these commits fix code climate issues, we close this PR and cherry-pick the commits to the original PR: #7188

PR #7219, opened 2019-03-18
Deprecate User.current

coolo:
I looked at the usage of User.current (by means of git grep) and seperated:
  - moved to User.pcurrent for usages that accept nobody
  - moved to User.rcurrent for usages that actually require a login
  
I also renamed User.current= to User.set_current plainly to support my git grep workflow.

But the idea is that the caller decides if nobody is acceptable or not - but he always gets a reply (or an acception if the controller failed to require login).

The function names are purely to be different to User.current, but I can imagine User.required and User.possibly_anonymous to make it more obvious what the code wants.

Once this flies, we can remove nobody from being set as current user and create an unnamed user object purely in memory (as long as it returns true to is_nobody? :)


src/api/app/models/user.rb
```diff
@@ -238,6 +238,7 @@ def self.pcurrent
   end
 
   def self.rcurrent
+    raise ArgumentError, 'Requiring user, but found nobodyg' unless current && !current.is_nobody?
```
adrianschroeter: typo .. "nobodyg"

coolo: That's his middle name!

coolo:
> I personally do not like `pcurrent` and `rcurrent`, thats kind of name that do not say anything and its easy to get confused by.. Would you mind to change it?

As I wrote: "The function names are purely to be different to User.current, but I can imagine 
 User.required and User.possibly_anonymous to make it more obvious what the code wants."

vpereira:
> As I wrote: "The function names are purely to be different to User.current, but I can imagine
> User.required and User.possibly_anonymous to make it more obvious what the code wants."

yup, but we dont approve comments, we approve code. So maybe you could add this proposition to the code. IMO ```current_or_nobody``` or ```possibly_nobody``` are better names

coolo:
I don't want your approval, I want your comments. You know what RFC stands for?

Anyway, if that's your only worry - I'm more interested in your suggestion for the requiring function. I can offer:
 - `User.current!`
 - `User.required`
 - `User.session`

For the other function, we have:
 - `User.current_or_nobody`
 - `User.possibly_nobody`
 - `User.anonymous_or_not` (ok, that's a joke)
 - `User.default_session`

coolo:
I went now with User.possibly_nobody and User.required - they can be `sed`ed though if someone comes up with a better idea :)

hennevogel:
Like we discussed in private already. I don't think we really have a good overview about authorization and now change one of the major conditions it depends on. I like that we catch things in the test suite with this, but I'm not sure that we do test all conditions.

I would feel better if we first move to pundit for authorization on the controller level, then do this.

coolo:
Well, I would feel better if we stopped using globalwritethrough before - but we can't have everything :)

hennevogel:
> but we can't have everything

But those two things we can definitely have!

coolo:
> I would feel better if we first move to pundit for authorization on the controller level, then do this.

I don't want to dismiss your feelings, but this change is actually less dangerous than moving authorization to the controller level. I mean the worst that can happen with this is that we throw an exception where previously nobody was returned.

hennevogel:
> I mean the worst that can happen with this is that we throw an exception where previously nobody was returned.

Or we allow something for nobody that before would throw an exception. No?

coolo:
> Or we allow something for nobody that before would throw an exception. No?

No, the opposite. User.current returned nobody - User.required never will. And that's the reason I reviewed the places where it's used instead of simply changing User.current as it was tried before.

hennevogel:
> No, the opposite. 

Your 🗣 in 🎅 👂

coolo:
The hakiri alarm is btw still false

vpereira:
@coolo among all the implementations alternatives, did you look into https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html ?

coolo:
No, I did not - and I really have no plan to refactor authentification, authorization and auditing all in one PR. This PR here is purely about splitting the API usage of what is known currently as magic User.current. So please stick to it.

And that ActiveSupport class is 'basically' syntactic sugar of what we're already doing. So it might be sensensible to refactor our custom code to use it - but not here.

src/api/app/controllers/application_controller.rb
```diff
@@ -460,14 +460,14 @@ def shutup_rails
   end
 
   def set_nobody
-    User.current = User.find_nobody!
+    User.set_current(User.find_nobody!)
```
vpereira: What's happening here if ```User.find_nobody!``` raises an exception?

src/api/app/controllers/application_controller.rb
```diff
@@ -460,14 +460,14 @@ def shutup_rails
...
   end
 
   def set_influxdb_data
-    anonymous = User.current_login == User::NOBODY_LOGIN
+    anonymous = User.possibly_nobody.is_nobody?
```
vpereira: ```anonymous?``` would be a better name for a boolean variable, no?

src/api/app/controllers/application_controller.rb
```diff
@@ -460,14 +460,14 @@ def shutup_rails
...
 
     InfluxDB::Rails.current.tags = {
-      beta: !anonymous && User.current.in_beta?,
+      beta: !anonymous && User.possibly_nobody.in_beta?,
```
vpereira: Can the nobody user be in beta?

src/api/app/controllers/build_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
       return
     end
 
-    if User.current.is_admin?
+    if User.possibly_nobody.is_admin?
```
vpereira: Is that possible that an user nobody is admin? Wouldn't a  ```User.current && User.current.is_admin?``` express better the business logic?

src/api/app/controllers/person_controller.rb
```diff
@@ -237,7 +237,7 @@ def change_my_password
   end
 
   def change_password(login, password)
-    unless User.current
+    unless User.required
```
vpereira: this kind of control shouldn't be done as, normally with others authentication framework, with a ```before_filter```?

src/api/app/controllers/application_controller.rb
coolo: What happened before if it raised an exception? I just changed the API of User model and shied away from too many logic changes in one PR (because I was told repeatedly that you guys prefer step by step approaches).

src/api/app/controllers/application_controller.rb
coolo: out of scope

src/api/app/controllers/application_controller.rb
coolo: No, it can't - but the current user can be nobody.

src/api/app/controllers/build_controller.rb
coolo: Not if User.current returns nobody - which is what I'm trying to make explicit with this PR. *Nothing* else. Refactor such code later to make sense: sure, but what you say is out of scope.

src/api/app/controllers/person_controller.rb
coolo: that question doesn't make sense to me (syntactically). This code should use a different methology to check users, but what is your question exactly?

coolo:
189a4f9d1befed7e34d72a2041d161a2234adf71 introduces User.admin_session? to simplify the use case of checking admins. This is again just a wrapper around existing logic

src/api/app/controllers/application_controller.rb
bgeuken: When I see a `foo?` somewhere I would expect it to be a method. I guess others would have the same "issue".

Anyway, it's not possible in Ruby to have `?` in variable names.

src/api/app/models/user.rb
```diff
@@ -216,29 +216,31 @@ def self.find_with_credentials_via_ldap(login, password)
...
     current || nobody
   end
 
-  def self.current=(user)
-    Thread.current[:user] = user
+  # Currently logged in user. Will thrown an exception if no user is logged in.
+  # So the controller needs to require login if using this (or models using it)
+  def self.required
+    raise ArgumentError, 'Requiring user, but found nobody' unless current && !current.is_nobody?
+    current
```
bgeuken: Conceptually I would find it nicer if `required` only takes care of ensuring that there is a user set and setting the user and all that happens elsewhere.
So I could write in my code:

```ruby
def index
  User.required!
  render xml: ::Cloud::Backend::UploadJob.all(::User.current, format: :xml)
end
```

It might even make sense to call this before calling the action or methd. But this seems to require some more refactoring.

bgeuken: I think this would also allow us to drop the `possibly_nobody` again.The `required!` method would take care of permitting access to nobody and `current` would take care of returning a user ('nobody' in case it's nil)

src/api/app/models/user.rb
```diff
@@ -216,29 +216,31 @@ def self.find_with_credentials_via_ldap(login, password)
...
   end
 
-  def self.current_login
-    current.try(:login) || NOBODY_LOGIN
+  # Currently logged in user or nil
+  def self.not_required
+    current if current && !current.is_nobody?
+  end
+
+  def self.set_current(user)
```
bgeuken: Not that I have a strong opinion about this, but in the ruby world the setter would usually be `current=`. We usually try to avoid `set_foo`, `do_bar`

src/api/app/models/user.rb
coolo: I'm actually fine renaming possibly_nobody back to current - once we drop assigning nobody to current. 

I really would like to take this refactoring step by step, because a lot can break if you screw it. So first step: deprecate User.current and start to instrument code if requiring a user. And I can't continue this refactoring in a parallel branch as every code change needs to follow this new paradigm.

In a controller your code looks good, but in a partial it's already problematic. But possibly it's enough. Later on pundit can also set this required! thing

src/api/app/models/user.rb
coolo: I know, basically once it's agreed to go this route I can revert this. But as long as this is in a parallel branch, I need a way to grep for User.current in rebased code :(

coolo: If you have a suggestion, we can also fixing this by removing current completely and replace it with User.tobedefined=

src/api/app/models/user.rb
bgeuken: That's why I didn't make it a change request. You mentioned that in the PR description that you plan a couple of more changes. It's just a bit hard to see where we (you actually:laughing:) are heading right now.
And I would like you to consider this early on:-)
But if the current `User.required` for a longer time I would have mixed feelings.

> In a controller your code looks good, but in a partial it's already problematic.

My thinking is that if we happen to require a non-nobody user in a partial, or any other method, we should ensure upfront that this is the case.
But it's true we might need to take some steps to get there.

src/api/app/models/user.rb
bgeuken: Okay, I see. I don't have a better name right now and if this changes anyway it can stay as is.
Though I would add a fixme so everyone knows and you don't have to answer this again and again.

bgeuken:
The rest seems to work fine:+1:

coolo:
> lib/feature_switch/feature.rb:5:in `active?'

I only grepped app/, will look into lib/ later on :)

bgeuken:
Then please make it a draft PR or add the WIP label. So we don't merge this by accident.

coolo:
OK, the file in lib/ is fine - I guess you didn't restart your server after switching branches.

But before this is merged, it needs to be rebased (and refixed), so it's WIP and it's not :)

src/api/app/models/user.rb
coolo: How about User.session= ?

vpereira: @coolo so as I understood, you will do then later a follow-up PR removing this ```set_current``` and moving it to ```self.current=``` right?

coolo: Unless I get upvotes for User.session= yes, that's the plan.

dmarcoux:
I don't feel confident in reviewing a PR with 222 files changed. I am sorry, but I would do it if it was split in smaller parts. I know it's often the same change, but it's still huge.

src/api/app/controllers/application_controller.rb
vpereira: @bgeuken a method fits well here.

vpereira:
Hi @coolo, 

TL;DR

**with great PR comes great responsibility** 

We are failing to understand what are you trying to achieve with this PR, specially if what you are doing is the best way to tackle the issue that we (maybe) have. For example the question "if we move to devise, will this whole issue go away?" is still open for us.

I personally approved this PR, however I didn't really thought through the possible side-effects that we will have to deal after this PR is merged and honestly, I'm not sure I will be able to do it in the next weeks. 

I think what we are missing is a background or "why this change is needed and what you want to achieve"  with that changes and if there is a possible way to do a so "intrusive change" in small steps.. Does that works for you?

coolo:
I squashed this now into one commit and tried to cherry pick units. Rebasing will continue :)

Ana06:
@coolo should we close this? :thinking:

coolo:
yeah, let's start from scratch

PR #7225, opened 2019-03-19
Add history to StagingProjectController#show view

vpereira:
Add history elements to the Staging Project show xml view

### DIFF
--- a/src/api/spec/controllers/staging/staging_projects_controller_spec.rb
+++ b/src/api/spec/controllers/staging/staging_projects_controller_spec.rb
@@ -105,6 +105,7 @@ RSpec.describe Staging::StagingProjectsController, type: :controller, vcr: true 
           assert_select 'broken_packages', 1 do
             assert_select 'entry', 2
           end
+          assert_select 'history', 1
         end
       end
     end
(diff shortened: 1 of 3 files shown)


vpereira:
[Add the history to the staging project show API](https://trello.com/c/F5N24rjI/1085-add-the-history-to-the-staging-project-show-api)

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
```diff
@@ -105,6 +105,7 @@
           assert_select 'broken_packages', 1 do
             assert_select 'entry', 2
           end
+          assert_select 'history', 1
```
dmarcoux: Is there a way to check if it's really what we expect? Now, it's only checking that there is a history element.

vpereira: Looking the other tests, I don't see value in testing active record and/or active view.. any suggestion @dmarcoux ?

dmarcoux: I'm not sure if we need to go further. It's true that this would test ActiveRecord / ActionView.

PR #7228, opened 2019-03-19
Swap buttons for copying staging projects

bgeuken:
In OBS, primary buttons are on the right.


bgeuken:
Fixed. Thanks for reviewing.

PR #7232, opened 2019-03-20
Fix active review tabs when there is no decision box

### LINKED ISSUE #7229: [beta UI] Multiple review request strangely displayed
Screenshot from 2019-03-19 15-39-29

Screenshot, while reviewing a submit request. When loading, it shows like this. At the bottom, there are two pending reviews I could potentially accept (opensuse-review-team and a project review from Staging)

On first sight, the two reviews being one right after the other is confusing - once switched using the 'tabs' above the review, only one stays left shown (and switches with the tab)

bgeuken:
When there is no decision box for the request (the user can not
accept or decline it) we have to mark the first review as
active (= visible).
Since we did not consider the index of the review list, all reviews
were visible when no the request decision box is hidden.

Fixes #7229

### DIFF
--- a/src/api/app/views/webui2/webui/request/show.html.haml
+++ b/src/api/app/views/webui2/webui/request/show.html.haml
@@ -50,7 +50,8 @@
                   My decision
             - @my_open_reviews.each_with_index do |review, index|
               %li.nav-item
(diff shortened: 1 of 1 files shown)


bgeuken:
A prepared request can be found [here](https://obs-reviewlab.opensuse.org/bgeuken-fix_issue_7229/request/show/1). Just login with user tux / opensuse.

PR #7169, opened 2019-03-13
2.9 download repository policy

vpereira:
Backport from PR #7165

PR #7235, opened 2019-03-20
Only allow admins to create sourceaccess/access repositories flags

Ana06:
Webui::RepositoriesController.create_flag allowed creating the access/sourceaccess flags, while in
SourceController#project_command_set_flag this is only allowed for admins.

This was fixed in master in: https://github.com/openSUSE/open-build-service/pull/7169/commits/5f89387c86be9bcefffe64e78125efa88e3c3c4f

### DIFF
--- a/src/api/app/controllers/webui/repositories_controller.rb
+++ b/src/api/app/controllers/webui/repositories_controller.rb
@@ -177,6 +177,7 @@ class Webui::RepositoriesController < Webui::WebuiController
     authorize @main_object, :update?
 
     @flag = @main_object.flags.new(status: params[:status], flag: params[:flag])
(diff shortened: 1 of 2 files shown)


Ana06:
we need to mark Hakiri as slved as we are patchig this directly on build.opensuse.org

dmarcoux:
The CircleCI check can be disregarded

bgeuken:
The title is a bit misleading, since this is only the case when the sourceaccess / access flag is set.

Ana06:
@bgeuken 

> The title is a bit misleading, since this is only the case when the sourceaccess / access flag is set.

I didn't want to change the title of the commit in master much

PR #7239, opened 2019-03-20
Add a task to generate staging workflow data for development environment

eduardoj:
Co-authored-by: Björn Geuken <bgeuken@suse.de>

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -136,6 +136,33 @@ namespace :dev do
       end
     end
   end
+
+  namespace :staging do
+    task data: :environment do
+      unless Rails.env.development?
+        puts "You are running this rake task in #{Rails.env} environment."
+        puts 'Please only run this task with RAILS_ENV=development'
+        puts 'otherwise it will destroy your database data.'
+        return
+      end
+
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+      timestamp = Time.now.to_i
+      maintainer = create(:confirmed_user, login: "maintainer_#{timestamp}")
+      User.current = maintainer
+      managers_group = create(:group, title: "managers_group_#{timestamp}")
+      staging_workflow = create(:staging_workflow_with_staging_projects, project: maintainer.home_project, managers_group: managers_group)
+      staging_workflow.managers_group.add_user(maintainer)
+
+      staging_workflow.staging_projects.each do |staging_project|
+        2.times { |i| request_for_staging(staging_project, maintainer.home_project, "#{staging_project.id}_#{timestamp}_#{i}") }
+      end
+
+      puts "**** Created staging workflow project: /staging_workflows/#{staging_workflow.id} ****"
+    end
+  end
+
   namespace :development_testdata do
     task create: :environment do
       unless Rails.env.development?
@@ -265,4 +292,21 @@ def copy_example_file(example_file)
     puts "Creating config/#{example_file} from config/#{example_file}.example"
     FileUtils.copy_file("#{example_file}.example", example_file)
   end
+end
+
+def request_for_staging(staging_project, maintainer_project, suffix)
+  requester = create(:confirmed_user, login: "requester_#{suffix}")
+  source_project = create(:project, name: "source_project_#{suffix}")
+  target_package = create(:package, name: "target_package_#{suffix}", project: maintainer_project)
+  source_package = create(:package, name: "source_package_#{suffix}", project: source_project)
+  request = create(
+    :bs_request_with_submit_action,
+    state: :new,
+    creator: requester,
+    target_package: target_package,
+    source_package: source_package,
+    staging_project: staging_project
+  )
+
+  request.reviews.each { |review| review.change_state(:accepted, 'Accepted') }
 end
--- a/src/api/.rubocop.yml
+++ b/src/api/.rubocop.yml
@@ -37,6 +37,7 @@ Metrics/BlockLength:
   Exclude:
     - 'config/routes.rb'
     - 'spec/**/*'
+    - '**/*.rake'
 
 ##################### Rails ##################################
 


eduardoj:
[Merge staging projects into Staging workflow main project](https://trello.com/c/xCnvFLBT/1017-merge-staging-projects-into-staging-workflow-main-project)

src/api/lib/tasks/dev.rake
```diff
@@ -136,6 +136,50 @@ namespace :dev do
...
+    source_project = create(:project, name: "source_project_#{suffix}")
+    target_package = create(:package, name: "target_package_#{suffix}", project: maintainer_project)
+    source_package = create(:package, name: "source_package_#{suffix}", project: source_project)
+
+    request = create(:bs_request_with_submit_action,
+      state: :new,
+      creator: requester,
+      target_package: target_package,
+      source_package: source_package,
+      description: 'BsRequest 1',
```
bgeuken: I guess we can drop this line. Now that we create multiple requests

eduardoj: Dropped.

src/api/lib/tasks/dev.rake
```diff
@@ -266,3 +293,20 @@ def copy_example_file(example_file)
...
+  request = create(
+    :bs_request_with_submit_action,
+    state: :new,
+    creator: requester,
+    target_package: target_package,
+    source_package: source_package,
+    staging_project: staging_project
+  )
+
+  request.reviews.each { |review| review.change_state(:accepted, 'Accepted') }
```
coolo: What's the point of this?

src/api/.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2019-03-15 06:48:15 +0000 using RuboCop version 0.65.0.
+# on 2019-03-20 15:24:35 +0000 using RuboCop version 0.65.0.
```
Ana06: why it is needed?

eduardoj: You can read the commit. I will paste it here:

With the addition of code in file `src/api/lib/tasks/dev.rake`, cop
Metrics/BlockLength complained about the length of the `dev` block.

Recreating the RuboCop TODO file with the
`dev:lint:rubocop:auto_gen_config:rails` task increments the length limit,
and then RuboCop doesn't complain.

src/api/lib/tasks/dev.rake
bgeuken: We needed a staging project that is in 'accepted' state. And for that all requests need to be accepted.

On second thought... we probably could / should handle this via a factory. :thinking:

src/api/.rubocop_todo.yml
coolo: Does this Metrics/BlockLength make sense at *any* point?

src/api/lib/tasks/dev.rake
coolo: You guys confuse requests and reviews at a level that scares me.

If requests are accepted, they should not be in stagings any longer. And if requests don't have reviews because none was added, the staging project is acceptable.

bgeuken: Where do you see an accepted request? ....

src/api/.rubocop_todo.yml
bgeuken: I would consider an iterator, like `map`, that contains a lot of code a code smell. For such cases the cop makes IMO sense.

We could do what rubocop [does](https://github.com/rubocop-hq/rubocop/commit/157b947473e32e6e79bedfa57798fdf9fc6a0650) and ignore this in case of rake files and gem specs...

coolo: But the limit is 457 lines atm - that's not a code smell, that's a code toilet

src/api/lib/tasks/dev.rake
coolo: Again: where are the reviews coming from you are accepting?

src/api/.rubocop_todo.yml
bgeuken: You were asking about the general sense of  Metrics/BlockLength:wink:

The interesting part is what maximum we would need after ignoring rake tasks and alike, as I proposed.

src/api/lib/tasks/dev.rake
bgeuken: Good question! There seems to be some code that creates them. Probably in the factories. Though I didn't find anything suspicious on first glance.

coolo: Is it by chance by_group manager_group? If so, blindly accepting them without actually selecting it in the staging project might be fine for tests, but is surely not cool for development environment.

vpereira: @coolo in that cause I think we just wanted to bring the overall_state to ```:acceptable``` to be able to "merge" staging projects and be able to work on our card.. that's not a generic staging dashboard setup rather a specific one.

@bgeuken maybe we should rename the rake task to reflect what it really is?

coolo: Yeah - and if you 'merge' staging projects you want a real staging project and not a fake one.

src/api/.rubocop_todo.yml
Ana06: @eduardoj I read the commit, but you shouldn't introduce new offenses... if this cop doesn't make sense for this file move it to `.rubocop.yml` (with a comment with the reason). If it is in the todo is because you plan to fix it and then you should fix it.

src/api/lib/tasks/dev.rake
bgeuken: Seems my message wasn't sent before:/

We started this rake task to have some test data that we can quickly set up on any machine. So that we have an easier time when switching the driver during our mob-programming session.
We don't claim that the data we create there are complete yet, and since we are using test factories to set this up, there might be some flaws we run into (which we are going to address).

But like @vpereira said the main goal here is to have a staging project in `:acceptable` state. This is needed so we can start implementing the `merge` feature.

src/api/.rubocop_todo.yml
eduardoj: I removed the commit that recreated the RuboCop TODO file, and I created one commit excluding the Metrics/BlockLength cop for `*.rake` files, as @bgeuken suggested. Please @Ana06 and @coolo, have a look again.

src/api/lib/tasks/dev.rake
coolo: And as I said: if you start implementing merge on this staging projects, you will run into very strange problems you won't have if you set up these staging projects properly

PR #7241, opened 2019-03-20
Drop unused IchainNotifier

hennevogel:
No idea if this ever worked, but it's commented since 2012...


coolo:
I remember it working - but I'm an elephant :)

PR #7243, opened 2019-03-20
Drop Messages

hennevogel:
We don't use Messages in any way in the application. This is just a
route to drop text into our database. We also do not use this in any
way in any production instance I've checked.


coolo:
I can't say I was aware of this model :)
7a5a2b69bea3be6804376bc9c2ddb19db9d7a3a2 commited it as 'frontend part', but it's unclear what part was missing, I'm not aware of any messages in the webui.

Comments prototype that never made it?

hennevogel:
Yeah sounds like a prototype. I though I remove it before you use it for passing around comments without notifications :clown_face:

src/api/db/migrate/20190320185348_drop_messages.rb
```diff
@@ -0,0 +1,5 @@
+class DropMessages < ActiveRecord::Migration[5.2]
+  def change
+    drop_table :messages
```
dmarcoux: This isn't reversible and Rubocop is not happy, so I guess we either ignore this in Rubocop or do it as suggested [here](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/ReversibleMigration):

```
# drop_table

# bad
def change
  drop_table :users
end

# good
def change
  drop_table :users do |t|
    t.string :name
  end
end
```

I don't have a strong opinion on this.

coolo: See db/migrate/20181201065026_drop_group_requests.rb

hennevogel: anfänger :smiley:

hennevogel:
give @adrianschroeter a chance to say his piece

dmarcoux:
Any news here?

bgeuken:
We should mention in the release notes that this model and the API endpoints got dropped. Just in case that someone was using this on his / her instance

adrianschroeter:
This is most likely really obsolete, but we have not declared it as deprecate yet.

So I like to propose our usual procedure here:
1. Remove it from api.txt file
2. Add a note in ReleaseNotes 2.10 that this is deprecated and will be dropped in future
3. remove the code in master after next release has been branched

Or does this code cause much additional work otherwise?

PR #7244, opened 2019-03-20
Drop Ratings

hennevogel:
We don't use this at all in the app. It's just a way to dump
text into our database. The latest Ratings in production are
from 2010.


hennevogel:
give @adrianschroeter a chance to say his piece

dmarcoux:
Any news here?

bgeuken:
We should mention in the release notes that this model and the API endpoints got dropped. Just in case that someone was using this on his / her instance

adrianschroeter:
This is most likely really obsolete, but we have not declared it as deprecate yet.

So I like to propose our usual procedure here:
1. Remove it from api.txt file
2. Add a note in ReleaseNotes 2.10 that this is deprecated and will be dropped in future
3. remove the code in master after next release has been branched

Or does this code cause much additional work otherwise?

PR #7245, opened 2019-03-20
Drop StatusCount model

hennevogel:
Use a local struct instead. Cheaper, easier, more flavor!

### DIFF
--- a/src/api/app/models/buildresult.rb
+++ b/src/api/app/models/buildresult.rb
@@ -85,9 +85,10 @@ class Buildresult
         state: state
       )
 
+      Struct.new('StatusCount', :code, :count)
       build.summary = []
       result['summary'].elements('statuscount').each do |count|
-        build.summary << StatusCount.new(code: count['code'], count: count['count'])
+        build.summary << Struct::StatusCount.new(count['code'], count['count'])
       end
 
       build.summary.sort! { |a, b| code2index(a.code) <=> code2index(b.code) }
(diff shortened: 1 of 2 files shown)


src/api/app/models/buildresult.rb
```diff
@@ -85,6 +85,7 @@ def self.summary(project_name)
         state: state
       )
 
+      StatusCount = Struct.new(:code, :count)
```
dmarcoux: To make Rubocop happy: `Struct.new('StatusCount', :code, :count)`, but I think this could be a hash too. The lines with `build.summary` would have to be changed too if we do that.

src/api/app/models/status_count.rb
```diff
@@ -1,4 +0,0 @@
-class StatusCount
-  include ActiveModel::Model
-  attr_accessor :code, :count
-end
```
bgeuken: 🙈

src/api/app/models/buildresult.rb
```diff
@@ -85,6 +85,7 @@ def self.summary(project_name)
...
       build.summary = []
       result['summary'].elements('statuscount').each do |count|
         build.summary << StatusCount.new(code: count['code'], count: count['count'])
```
bgeuken: [Looks](https://ruby-doc.org/core-2.5.0/Struct.html) like this could be written as ``StatusCount.new(count['code'], count['count'])` now:-)

src/api/app/models/buildresult.rb
hennevogel: thanks!

PR #7246, opened 2019-03-20
Authorize move_path

dmarcoux:
Additionally turn the after_action around and maintain a whitelist instead of a blacklist. This requires new methods authorized by default.

new_package and its dialog are just informal authorizations, so people do not even try to do this.

### DIFF
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -39,7 +39,13 @@ class Webui::ProjectController < Webui::WebuiController
 
   before_action :set_maintained_project, only: [:remove_maintained_project]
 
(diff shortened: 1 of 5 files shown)


dmarcoux:
Debugging this is tricky... I cannot run the Docker setup on the 2.9 branch. I rebuilt the frontend Docker image, but it still doesn't work. I might need to rebuild the frontend-base too... It just takes a while.

dmarcoux:
In the end, I had to clone the repository and start from scratch (so removing all Docker images/containers). It now works.

vpereira:
@dmarcoux looks like one rspec is broken (restarted the job to check if it wasn't flickering), looks like its giving a 500 error

dmarcoux:
@vpereira I know it's broken, I've been trying to fix this for 2 days now... I cannot reproduce it locally.

PR #7249, opened 2019-03-21
Allow for uploading multiple files to projects

### LINKED ISSUE #242: Upload multiple files in add file dialog
Using the web interface I need to upload the files for a new project. The offered add-file dialog can only add a single file.

Would be nice to have the capability to add multiple files or the content of a directory.

hellcp:
It ain't pretty, but it works (for now).

#TODO
JS:
- [x] Allow to rename files on file by file basis
- [x] Allow to add more URL fields  (including gh stuff)

Also already fixes #242, 6 years old issue at this point (and probably 12 yo by the time I finish this)

### DIFF
--- a/src/api/app/assets/javascripts/webui2/add_file.js
+++ b/src/api/app/assets/javascripts/webui2/add_file.js
@@ -1,4 +1,3 @@
-var enumerateLinks = 0;
 function createInputs(item) {
   var filelist = $(item.list);
   var li = document.createElement('li');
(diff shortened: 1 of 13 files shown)


marcus-h:
On 2019-03-20 19:20:26 -0700, Stasiek Michalski wrote:
> diff --git a/src/api/app/controllers/webui/package_controller.rb b/src/api/app/controllers/webui/package_controller.rb
> index f8800c3807b..694663ef6ef 100644
> --- a/src/api/app/controllers/webui/package_controller.rb
> +++ b/src/api/app/controllers/webui/package_controller.rb
> @@ -696,6 +697,13 @@ def save_file
>          unless services.save
>            errors << "Failed to add file from URL '#{file_url}'"
>          end
> +      elsif files.present?
> +        # We are getting uploaded files in an array
> +        files.each do |file|
> +          filename = file.original_filename if filename.blank?
> +          @package.save_file(file: file, filename: filename, comment: params[:comment])

Hmm just an idea: instead of a creating a commit for each file, it might
make sense to upload the files first and then do a single commit (which
includes the existing + uploaded files).

hellcp:
@marcus-h hm, yeah, good point

hennevogel:
@hellcp ping

hellcp:
Screenshot from 2019-08-16 02-56-56

hellcp:
Hm, apparently the package created for bootstrap tests has unmerged sources, which when expanded throw a vcr error

PR #7250, opened 2019-03-21
Update the manager group for a staging via the API

eduardoj:
And also add the documentation.

### DIFF
--- a/src/api/app/controllers/staging/workflows_controller.rb
+++ b/src/api/app/controllers/staging/workflows_controller.rb
@@ -1,6 +1,7 @@
 class Staging::WorkflowsController < ApplicationController
   before_action :require_login
   before_action :set_project
+  before_action :set_staging_workflow, only: :update
 
   def create
     staging_workflow = @project.build_staging
@@ -34,6 +35,22 @@ class Staging::WorkflowsController < ApplicationController
     staging_workflow.destroy!
   end
 
+  def update
+    authorize @staging_workflow
+
+    @staging_workflow.managers_group = Group.find_by!(title: xml_hash['managers'])
+
+    if @staging_workflow.save
+      render_ok
+    else
+      render_error(
+        status: 400,
+        errorcode: 'invalid_request',
+        message: "Staging #{@staging_workflow} couldn't be updated: #{@staging_workflow.errors.to_sentence}"
+      )
+    end
+  end
+
   private
 
   def set_project
@@ -45,6 +62,10 @@ class Staging::WorkflowsController < ApplicationController
       errorcode: 'not_found',
       message: "Project '#{params[:staging_workflow_project]}' not found."
     )
+  end
+
+  def set_staging_workflow
+    @staging_workflow = ::Staging::Workflow.find_by!(project: @project)
   end
 
   def xml_hash
--- a/src/api/app/models/staging/workflow.rb
+++ b/src/api/app/models/staging/workflow.rb
@@ -16,8 +16,8 @@ class Staging::Workflow < ApplicationRecord
   end
 
   has_many :target_of_bs_requests, through: :project, foreign_key: 'staging_workflow_id' do
-    def stageable
-      managers_group_title = proxy_association.owner.managers_group.try(:title)
+    def stageable(managers_group_title = nil)
+      managers_group_title ||= proxy_association.owner.managers_group.try(:title)
       includes(:reviews).where(state: :review, staging_project_id: nil, reviews: { state: :new, by_group: managers_group_title })
     end
 
@@ -34,7 +34,7 @@ class Staging::Workflow < ApplicationRecord
 
   after_create :create_staging_projects
   after_create :add_reviewer_group
-  before_update :update_staging_projects_managers_group
+  before_update :update_managers_group
 
   def unassigned_requests
     target_of_bs_requests.stageable.where.not(id: excluded_requests)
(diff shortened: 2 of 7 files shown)


eduardoj:
[As a project owner I want to update the staging manager group to stagings via the API](https://trello.com/c/oT7PAAZ9/1084-as-a-project-owner-i-want-to-update-the-staging-manager-group-to-stagings-via-the-api)

src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -32,6 +49,17 @@ def set_project
     )
   end
 
+  def set_staging_workflow
+    @staging_workflow = ::Staging::Workflow.find_by(project: @project)
```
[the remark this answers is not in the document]
eduardoj: Done.

coolo:
This is not handling any kind of migration - is this the same with the webui or is there no webui route yet?

eduardoj:
> This is not handling any kind of migration - is this the same with the webui or is there no webui route yet?

Why would it be needed a migration?

coolo:
> Why would it be needed a migration?

Just look at the code to see where the manager group is actually used. Storing it in the DB next to the workflow is pretty useless in itself - it's the after_create hooks that actually matter. And then there are all the requests pending.

src/api/spec/controllers/staging/workflows_controller_spec.rb
```diff
@@ -50,4 +52,27 @@
       it { expect(project.staging).to be_nil }
     end
   end
+
+  describe 'PUT #update' do
+    context 'with a valid managers group' do
```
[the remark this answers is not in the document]
eduardoj: @Ana06:

> what about a test for a group that doesn't exist and for a user without permissions?

Added. Please have a look again.

eduardoj:
@coolo, after we talked in person I added a `before_update` callback, for updating the review group of the staging workflow project. Please have a look again.

src/api/app/models/staging/workflow.rb
```diff
@@ -108,4 +109,15 @@ def add_reviewer_group
...
+    return unless changes[:managers_group_id]
+
+    old_managers_group = Group.find(changes[:managers_group_id].first)
+    new_managers_group = managers_group
+
+    role = Role.find_by_title('reviewer')
+    relationship = project.relationships.find_by(group: old_managers_group, role: role)
+    relationship.update(group: new_managers_group)
+  end
 end
```
coolo: This really should be one callback. There is already a before_update hook looking at the group. Basically what needs to be done on group update:

- Change the reviewer role in target project (make sure the project is stored to backend afterwards, I'm not sure this happens here)
- Change the maintainer role in all staging projects (happens in the other callback, but I'm not sure it stores the project either)
- Change all requests in backlog by adding review by_group: new and accepting review by_group: old (in that order)

vpereira: shouldn't we pack these "3 steps" in one transaction?

coolo: Note however that rollback needs special handling of the backend

eduardoj:
Please, have a look again. I will need some help with this, for sure.

bgeuken:
Seems fine to me. @coolo could you review again?
The transaction / rollback is missing, but I think that's fine for now. It's at least not a regression.

vpereira:
> Seems fine to me. @coolo could you review again?
> The transaction / rollback is missing, but I think that's fine for now. It's at least not a regression.

yes, it would be nice to have, but I don't see it as mandatory for this PR. Nice work @eduardoj !

src/api/app/models/staging/workflow.rb
```diff
@@ -101,11 +114,11 @@ def update_staging_projects_managers_group
     # FIXME: This assignation is need because after store a staging_project
     # the object is reloaded and we lost the changes.
     self.managers_group = new_managers_group
-  end
 
-  def add_reviewer_group
-    role = Role.find_by_title('reviewer')
-    project.relationships.find_or_create_by(group: managers_group, role: role)
+    # update reviewer group in staging project
```
coolo: the comment is wrong. this is not about the staging project but the project the workflow is for (do you have a name for that? We call it normally target project)

src/api/app/models/staging/workflow.rb
```diff
@@ -101,11 +114,11 @@ def update_staging_projects_managers_group
...
+    reviewer_role = Role.find_by_title('reviewer')
+    relationship = project.relationships.find_by(group: old_managers_group, role: reviewer_role)
+    relationship.update(group: new_managers_group)
```
coolo: Question: what happens if that group already exists? I see an uniqueness validation in relationships on role and group. I assume the update would fail and the old group would stay?

Ana06: yes, the old group would stay. I guess we need to accept the review if there is already one for that group :thinking:

src/api/app/models/staging/workflow.rb
eduardoj: @coolo:

> the comment is wrong. this is not about the staging project but the project the workflow is for (do you have a name for that? We call it normally target project)

Right. Based in our [wiki documentation](https://github.com/openSUSE/open-build-service/wiki/Staging-Workflow) I will rename it to "staging workflow project".

src/api/app/models/staging/workflow.rb
eduardoj: @coolo, @Ana06, modified. Please have a look again.

coolo:
The change is ok, but I wonder about the overlap between 'update manager group' and 'link projects to workflow' PRs. One is PUT to /workflow and one is POST to /staging_projects - and both have a very similiar looking XML in their body. But once with group and once with projects. This API doesn't make a consistent impression.

eduardoj:
@coolo:

> The change is ok, but I wonder about the overlap between 'update manager group' and 'link projects to workflow' PRs. One is PUT to /workflow and one is POST to /staging_projects - and both have a very similiar looking XML in their body. But once with group and once with projects. This API doesn't make a consistent impression.

So what do you propose instead of the current routes and bodies?

coolo:
I find one update route in the controller that will change manager groups **and** new staging projects easier to digest in the API.

PR #7262, opened 2019-03-22
Refactor state method for staging project

saraycp:
This code was a bit confusing. It had a lot of conditions that called
different methods, and some methods called each others. So it has been simplified.

### DIFF
--- a/src/api/app/models/concerns/staging_project.rb
+++ b/src/api/app/models/concerns/staging_project.rb
@@ -142,35 +142,21 @@ module StagingProject
   end
 
   def state
-    return :empty unless staged_requests.exists?
-    return :unacceptable if unacceptable_state?
-    return :review if review_state?
-    general_state
-  end
-
-  def unacceptable_state?
-    untracked_requests.present? || staged_requests.obsolete.present?
-  end
-
-  def review_state?
-    general_state == :acceptable && missing_reviews.present?
-  end
-
-  def general_state
-    build_state == :acceptable ? check_state : build_state
-  end
-
-  def build_state
-    return :building if building_repositories.present? || disabled_repositories.present?
-    return :failed if broken_packages.present?
-
+    return :empty if staged_requests.blank?
+    return :unacceptable if untracked_requests.present? || staged_requests.obsolete.exists?
+    bc_state = build_or_check_state
+    return bc_state if bc_state
+    return :review if missing_reviews.present?
     :acceptable
   end
 
-  def check_state
+  def build_or_check_state
+    # build_state
+    return :building if building_repositories.present? || disabled_repositories.present?
+    return :failed if broken_packages.present?
+    # check_state
     return :testing if missing_checks.present? || checks.pending.exists?
-    return :failed if checks.failed.exists?
-    return :acceptable
+    :failed if checks.failed.exists?
   end
 
   def set_buildinfo


src/api/app/models/concerns/staging_project.rb
```diff
@@ -142,35 +142,26 @@ def cache_problems
...
-  end
-
-  def unacceptable_state?
-    untracked_requests.present? || staged_requests.obsolete.present?
-  end
-
-  def review_state?
-    general_state == :acceptable && missing_reviews.present?
+    return :empty if staged_requests.blank?
+    return :unacceptable if untracked_requests.present? || staged_requests.obsolete.present?
```
Ana06: [nitpick] `staged_requests.obsolete` is active record and `present?` load it while `exists?` not

src/api/app/models/concerns/staging_project.rb
```diff
@@ -142,35 +142,26 @@ def cache_problems
...
+    return build_and_check_state if build_and_check_state
```
Ana06: This method is being called twice

src/api/app/models/concerns/staging_project.rb
```diff
@@ -142,35 +142,26 @@ def cache_problems
...
+    return :review if missing_reviews.present?
+    :acceptable
   end
 
-  def general_state
-    build_state == :acceptable ? check_state : build_state
+  def build_and_check_state
```
Ana06: [nitpick] Why not calling both methods directly instead of having a method to wrap them?

saraycp: Because doing it increased the complexity of the main method and Code Climate complained. 
We also thought about removing the two methods `build_state` and `check_state`and put all their code directly inside the wrapper. But we considered that, semantically, it's better like this.

Ana06: I am not sure if we should follow code climate advice in this case, I find this more complex to follow...

src/api/app/models/concerns/staging_project.rb
saraycp: Good catch!

src/api/app/models/concerns/staging_project.rb
saraycp: You are right! Something similar happened inside `build_and_check_state` so we have changed all a bit.

src/api/app/models/concerns/staging_project.rb
saraycp: @Ana06  We have changed it a bit. I think it's clearer now. Please have a look!

src/api/app/models/concerns/staging_project.rb
```diff
@@ -142,35 +142,22 @@ def cache_problems
...
-  end
-
-  def review_state?
-    general_state == :acceptable && missing_reviews.present?
-  end
-
-  def general_state
-    build_state == :acceptable ? check_state : build_state
+    return :empty if staged_requests.blank?
+    return :unacceptable if untracked_requests.present? || staged_requests.obsolete.exists?
```
vpereira: I'm wondering if now is better than before... Is that because of code climate? IMO The code climate cognitive complexity is definitely higher than before (introduced by the or condition). I would prefer to have it was before

src/api/app/models/concerns/staging_project.rb
```diff
@@ -142,35 +142,22 @@ def cache_problems
...
+    build_or_check_state = build_state || check_state
+    return build_or_check_state if build_or_check_state
```
hennevogel: Why don't we move all of this into one method? Are we checking `.build_state` independently from `check_state` somewhere?

vpereira: I don't know if I understood @hennevogel, but isn't better to hide all the "details" on methods and when reading the "state" method, you don't have to wrap your head around the details, but a quick glance on the method names, should make clear when the states are set?

Ana06: I am also not sure what @hennevogel means... Maybe it goes in the same direction as https://github.com/openSUSE/open-build-service/pull/7262#discussion_r268508491 :thinking:

src/api/app/models/concerns/staging_project.rb
```diff
@@ -142,35 +142,22 @@ def cache_problems
...
-
-  def review_state?
-    general_state == :acceptable && missing_reviews.present?
-  end
-
-  def general_state
-    build_state == :acceptable ? check_state : build_state
+    return :empty if staged_requests.blank?
+    return :unacceptable if untracked_requests.present? || staged_requests.obsolete.exists?
+    build_or_check_state = build_state || check_state
```
Ana06: `build_state` always return a value, which means that `check_state` will never be executed

coolo: It does not. Behind every return is an if (including the silent last one)

src/api/app/models/concerns/staging_project.rb
hennevogel: I mean make `build_or_check_state` a method. Not two method and a variable.

vpereira: @hennevogel ah yes, and preferably we see all return returning a ```:value```

src/api/app/models/concerns/staging_project.rb
saraycp: When we read this code on the mob-programming session, I realized most of us got confused because we had to jump to a different method on each condition. And even more tricky was that we called `general_state` at the end on the main method `state` but also from `review_state?` method. So the methods called each other :confounded: 

I really think it is clearer now, not so many confusing jumps.

But anyway, we can still improve it. And yes, code climate restrict it a lot... read the following comments.

src/api/app/models/concerns/staging_project.rb
saraycp: Exactly, in case `build_state` is not 'building' or 'failed' it will return with nil. Therefore, check_status will be called.

src/api/app/models/concerns/staging_project.rb
saraycp: I'm not sure if I understood.

@Ana06 suggested to get rid of the wrapper method `build_or_check_state` and not to call it twice (so, we assigned the condition to a variable).

@hennevogel, do you suggest to create the wrapper method again an call inside to the other two? Like this:

```
  def build_or_check_state                                                                               
    build_state || check_state                                                                                          
  end
```

Or do you propose that we have only one method with all the code inside? like:

```
  def build_or_check_state
    # build_state                                                                              
    return :building if building_repositories.present? || disabled_repositories.present?                 
    return :failed if broken_packages.present?                                                                  
    # check_state
    return :testing if missing_checks.present? || checks.pending.exists?                                 
    :failed if checks.failed.exists?                                                                     
  end 
```
(we thought about it but found it uglier, or semantically unclear)

@vpereira we can discuss later if we create a new method for each condition. Let's decide this before.

hennevogel: the latter

src/api/app/models/concerns/staging_project.rb
```diff
@@ -142,35 +142,21 @@ def cache_problems
...
-
-    :acceptable
-  end
-
-  def check_state
+    # check_state
     return :testing if missing_checks.present? || checks.pending.exists?
-    return :failed if checks.failed.exists?
-    return :acceptable
+    :failed if checks.failed.exists?
```
saraycp: Ok, I have done a mix of both suggestions. I created a single method called `build_or_check_state` and I also assigned it to a variable avoiding to call it twice.

@Ana06, @hennevogel can you have another look?

PS: I guess it's not a good idea to change the order of the lines, for this reason I don't join the two `failed`cases like:

```                              
    :failed if broken_packages.present? || checks.failed.exists?                                         
```

What do you think?

src/api/app/models/concerns/staging_project.rb
Ana06: I read it wrong, sorry

PR #7273, opened 2019-03-25
Remove symlink to foreman in Dockerfile

dmarcoux:
This isn't needed anymore as it produces the following error:
```
ln: failed to create symbolic link '/usr/bin/foreman': File exists
```


dmarcoux:
@bgeuken As we discussed on IRC.

Issue #7020, opened 2019-02-21
OBS WEBUI: User role selector on task list


Ana06:
From [openFATE](https://features.opensuse.org/323318):

> OBS Webui shall support filtering open tasks for a user by his workflow role there.
> 
> The role is defined via a group for now. Details for planing can be found in this trello card:
> 
> https://trello.com/c/H0fZCfFi/319-allow-filtering-of-requests-reviews-and-bugissues-by-user-groups
> 
> This request is just for tracking and connecting involved people.

hennevogel:
We have /group/show for this for a while...

coolo:
But it doesn't filter the group requests out of the others. So one group can hide all the other tasks. I have 339 tasks atm and basically no way to see what this is all about.

hennevogel:
Yeah `User.involved_p*` still uses the users groups too. We can get rid of that of course. That would help you, not so sure about other people that are only in groups that are not overwhelming.

coolo:
Not sure either - so I guess the feature here is to make this configurable: "support filtering"

hennevogel:
> I guess the feature here is to make this configurable

That's cheap. Let's decide.

coolo:
If I had to chose, I would pick github's grouping per repo. I.e. not one table to rule them all

PR #7277, opened 2019-03-26
Make group members tab the main tab on the group show view

bgeuken:
..., since this is the main content of a group.


dmarcoux:
I'm not sure if the `Group Members` tab is really the main tab here. Sadly, we don't have a way to track the usage of each tab. Other than this, it's all good.

bgeuken:
The group members are what defines a group. The other 3 tabs (reviews, requests, etc.) might be important, but are not about the group. They are, or might be, things that a group needs to work with. But for that we already have the `tasks` view.
And if that task view is currently missing some functionality, like only listing requests / reviews, for a certain group, we should extend it there.

coolo:
> And if that task view is currently missing some functionality, like only listing requests / reviews, for a certain group, we should extend it there.

yeah, right - https://github.com/openSUSE/open-build-service/issues/7020#issuecomment-465958330

The problem I see is you're technically right. For people looking for a group, they don't care about its tasks. But if *I* look at *my* tasks view, it gives me: 'factory-maintainers (326 tasks)' which links to the group. I expect it to link to 326 tasks, because *I* don't care who else is in that group (I basically know, but it's not the point here).

dmarcoux:
We could also change the default opened tabs by passing an anchor, so the link would for example be: https://build.opensuse.org/group/show/factory-auto#requests-in. In the end though, I believe it's just avoiding to fix the core issue.

hennevogel:
> Sadly, we don't have a way to track the usage of each tab. 

What are you missing? There is everything in place for you to generate and analyze data...

dmarcoux:
> What are you missing? There is everything in place for you to generate and analyze data...

I'm missing the time to do it.

bgeuken:
> coolo commented 18 hours ago

> And if that task view is currently missing some functionality, like only listing requests / reviews, for a certain group, we should extend it there.yeah, right - #7020 (comment)The problem I see is you're technically right. For people looking for a group, they don't care about its tasks. But if I look at my tasks view, it gives me: 'factory-maintainers (326 tasks)' which links to the group. I expect it to link to 326 tasks, because I don't care who else is in that group (I basically know, but it's not the point here).

I would prefer to address this problem instead carrying the current (as in bento UI) group show page over to the new UI.
It's not like the other tabs would be gone now... But we could move the other three tabs to a `/group/:group_title/tasks` page for now, if that helps. And in a later step we can then come up with a thought through design for task handling in general:-)

bgeuken:
> Sadly, we don't have a way to track the usage of each tab.

Setting up a metric for this shouldn't take much time. But they still wouldn't tell you the actual needs of our users, e.g. you might see which page are frequented a lot, but this won't tell us why. So we still need some user feedback, be it usability testing, direct comunication or other things.

dmarcoux:
Should we merge this or not?

coolo:
I just noticed the 'Backlog' link in staging dashboard links to groups/show:

https://build-test.opensuse.org/staging_workflows/1

PR #7282, opened 2019-03-27
[dist] Setting RAILS_ENV for delayed_job_monitor

srinidhibs:
RAILS_ENV defaults to development and running the command without
RAILS_ENV set to production, results in a backtrace.

```
$ sudo /usr/bin/bundle.ruby2.5 exec /srv/www/obs/api/script/delayed_job_monitor.rb
        26: from /srv/www/obs/api/script/delayed_job_monitor.rb:2:in `<main>'
        25: from /srv/www/obs/api/script/delayed_job_monitor.rb:2:in `require'
        24: from /srv/www/obs/api/config/environment.rb:24:in `<top (required)>'
        23: from /usr/lib64/ruby/gems/2.5.0/gems/railties-5.1.4/lib/rails/railtie.rb:185:in `method_missing'
        22: from /usr/lib64/ruby/gems/2.5.0/gems/railties-5.1.4/lib/rails/railtie.rb:185:in `public_send'
        21: from /usr/lib64/ruby/gems/2.5.0/gems/railties-5.1.4/lib/rails/application.rb:353:in `initialize!'
        20: from /usr/lib64/ruby/gems/2.5.0/gems/railties-5.1.4/lib/rails/initializable.rb:58:in `run_initializers'
        19: from /usr/lib64/ruby/2.5.0/tsort.rb:205:in `tsort_each'
        18: from /usr/lib64/ruby/2.5.0/tsort.rb:226:in `tsort_each'
        17: from /usr/lib64/ruby/2.5.0/tsort.rb:347:in `each_strongly_connected_component'
        16: from /usr/lib64/ruby/2.5.0/tsort.rb:347:in `call'
        15: from /usr/lib64/ruby/2.5.0/tsort.rb:347:in `each'
        14: from /usr/lib64/ruby/2.5.0/tsort.rb:349:in `block in each_strongly_connected_component'
        13: from /usr/lib64/ruby/2.5.0/tsort.rb:431:in `each_strongly_connected_component_from'
        12: from /usr/lib64/ruby/2.5.0/tsort.rb:350:in `block (2 levels) in each_strongly_connected_component'
        11: from /usr/lib64/ruby/2.5.0/tsort.rb:228:in `block in tsort_each'
        10: from /usr/lib64/ruby/gems/2.5.0/gems/railties-5.1.4/lib/rails/initializable.rb:59:in `block in run_initializers'
         9: from /usr/lib64/ruby/gems/2.5.0/gems/railties-5.1.4/lib/rails/initializable.rb:30:in `run'
         8: from /usr/lib64/ruby/gems/2.5.0/gems/railties-5.1.4/lib/rails/initializable.rb:30:in `instance_exec'
         7: from /usr/lib64/ruby/gems/2.5.0/gems/railties-5.1.4/lib/rails/application/finisher.rb:73:in `block in <module:Finisher>'
         6: from /usr/lib64/ruby/gems/2.5.0/gems/activesupport-5.1.4/lib/active_support/lazy_load_hooks.rb:49:in `run_load_hooks'
         5: from /usr/lib64/ruby/gems/2.5.0/gems/activesupport-5.1.4/lib/active_support/lazy_load_hooks.rb:49:in `each'
         4: from /usr/lib64/ruby/gems/2.5.0/gems/activesupport-5.1.4/lib/active_support/lazy_load_hooks.rb:50:in `block in run_load_hooks'
         3: from /usr/lib64/ruby/gems/2.5.0/gems/activesupport-5.1.4/lib/active_support/lazy_load_hooks.rb:65:in `execute_hook'
         2: from /usr/lib64/ruby/gems/2.5.0/gems/activesupport-5.1.4/lib/active_support/lazy_load_hooks.rb:60:in `with_execution_control'
         1: from /usr/lib64/ruby/gems/2.5.0/gems/activesupport-5.1.4/lib/active_support/lazy_load_hooks.rb:67:in `block in execute_hook'
/srv/www/obs/api/config/environments/development.rb:72:in `block (2 levels) in <top (required)>': uninitialized constant Bullet (NameError)
```

The last line confirms that `config/environments/development.rb` is being read. When `RAILS_ENV=production` is passed, the command runs without any errors:

```
$ sudo RAILS_ENV=production /usr/bin/bundle.ruby2.5 exec /srv/www/obs/api/script/delayed_job_monitor.rb
$ echo $?
0
```

### DIFF
--- a/dist/obs_api_delayed_jobs_monitor.cron
+++ b/dist/obs_api_delayed_jobs_monitor.cron
@@ -1,1 +1,1 @@
-*/10 * * * * RAILS_ENV=production /usr/bin/bundle.ruby2.5 exec /srv/www/obs/api/script/delayed_job_monitor.rb > /dev/null
+*/10 * * * * root RAILS_ENV=production /usr/bin/bundle.ruby2.5 exec /srv/www/obs/api/script/delayed_job_monitor.rb > /dev/null
--- a/dist/obs_api_delayed_jobs_monitor.cron
+++ b/dist/obs_api_delayed_jobs_monitor.cron
@@ -1,1 +1,1 @@
-*/10 * * * * /usr/bin/bundle.ruby2.5 exec /srv/www/obs/api/script/delayed_job_monitor.rb > /dev/null
+*/10 * * * * RAILS_ENV=production /usr/bin/bundle.ruby2.5 exec /srv/www/obs/api/script/delayed_job_monitor.rb > /dev/null


vpereira:
In production we have something like that in our ```.bashrc```:
```
api:~ # cat .bashrc 
export RAILS_ENV=production

API_ROOT=/srv/www/obs/api

function run_in_api
{
  # startproc only works for classical daemons
  chroot --userspec=wwwrun:www / /bin/bash -c "cd $API_ROOT && /usr/bin/bundle.ruby2.5 exec $*"
}
```

and we use it to run commands i.e:

```
run_in_api rails runner script/delayed_job_stats.rb
```

srinidhibs:
> In production we have something like that in our `.bashrc`:
> 

Thank you for sharing this!

But I don't think setting RAILS_ENV in /root/.bashrc will have any affect on how a cron job is invoked unless it is defined / set in the crontab.

vpereira:
> But I don't think setting RAILS_ENV in /root/.bashrc will have any affect on how a cron job is invoked unless it is defined / set in the crontab.

it won't not, I just wanted to point out why we never caught it in production - we run it often manually  while troubleshooting something

coolo:
No, we don't notice the problem because the file in production contains:
```
# disabled, cron.d entries would need a user name anyway and the command does not work
```

Someone forgot to file an issue :)

srinidhibs:
> No, we don't notice the problem because the file in production contains:
> 
> ```
> # disabled, cron.d entries would need a user name anyway and the command does not work
> ```
> 

Oh! You are right! I don't know how I missed this. 

> Someone forgot to file an issue :)

Shall I fix this then? 🙂

I'll add the `root` username for this cron job. Because `script/delayed_job_monitor.rb` invokes `service` command which is available only for `root` user.

coolo:
running things as root will break your permissions on log and lock files. I don't think this script is any useful :(

coolo:
A ruby script that calls a shell command to be run from a cron job to spit out:
```
service: no such service obsapidelayed
```

To then check for empty output to raise an exception from within that cron job. I suggest to plainly remove that script and its cron job.

coolo:
(ok, possibly the service still exists in 2.9, but the permission problem is still real)

dmarcoux:
Should we close this issue now that #7290 is merged? I would think so.

srinidhibs:
> (ok, possibly the service still exists in 2.9, but the permission problem is still real)

Yes. It exists in 2.9 and gets invoked just fine via cron. But I don't understand the concern about permissions. The script checks if a service is running and raises a `StandardError` that doesn't get written to `production.log`. Am I missing something obvious?

Although, while testing this script, I found another bug. `delayed_job_monitor.rb` runs `grep "active"` which also matches `inactive` and hence, the script doesn't raise any exception.

coolo:
It never worked, so it's useless. But this @dmarcoux this is 2.9 and 7290 is master

dmarcoux:
> It never worked, so it's useless. But this @dmarcoux this is 2.9 and 7290 is master

Oh right, my bad. I'm leaving this open then, but flagging it to 2.9.

coolo:
I vote for backporting #7290 and close this one - I'm still concerned what a root script does to the log files in /srv/obs.

@hennevogel @adrianschroeter ?

adrianschroeter:
yes, sounds to be save.

hennevogel:
Sorry @srinidhibs

PR #7284, opened 2019-03-27
Remove use_source from require_package

DavidKang:

### DIFF
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -260,7 +260,12 @@ class Webui::WebuiController < ActionController::Base
       begin
         @package = Package.get_by_project_and_name(@project.to_param, params[:package],
                                                    follow_project_links: true, follow_multibuild: true)
-      rescue APIException # why it's not found is of no concern :)
+      rescue APIException => exception
+        flash[:error] = if exception.is_a?(Package::ReadSourceAccessError)
+          "You don't have access to the sources of this package: \"#{params[:package]}\""
+        else
+          "Package \"#{params[:package]}\" not found in project \"#{params[:project]}\""
+        end
       end
     end
 
@@ -269,7 +274,6 @@ class Webui::WebuiController < ActionController::Base
     if request.xhr?
       render nothing: true, status: :not_found
     else
-      flash[:error] = "Package \"#{params[:package]}\" not found in project \"#{params[:project]}\""
       redirect_to project_show_path(project: @project, nextstatus: 404)
     end
   end
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -259,7 +259,7 @@ class Webui::WebuiController < ActionController::Base
     if params[:package].present?
       begin
         @package = Package.get_by_project_and_name(@project.to_param, params[:package],
-                                                   use_source: false, follow_project_links: true, follow_multibuild: true)
+                                                   follow_project_links: true, follow_multibuild: true)
       rescue APIException # why it's not found is of no concern :)
       end
     end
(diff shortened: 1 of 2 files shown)


src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -260,7 +260,12 @@ def require_package
       begin
         @package = Package.get_by_project_and_name(@project.to_param, params[:package],
                                                    follow_project_links: true, follow_multibuild: true)
-      rescue APIException # why it's not found is of no concern :)
+      rescue APIException => exception # why it's not found is of no concern :)
+        flash[:error] = if exception.is_a?(Package::ReadSourceAccessError) || exception.is_a?(Package::ReadAccessError)
+                          "You don't have access to the package \"#{params[:package]}\""
+                        else
+                          "Package \"#{params[:package]}\" not found in project \"#{params[:project]}\""
```
Ana06: are we sure that it was not found? couldn't be a different ''APIException"?

dmarcoux: We kept the same message as before, but we moved it up here so whenever it's not one of the access errors, we simply use it.

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -259,8 +259,13 @@ def require_package
     if params[:package].present?
       begin
         @package = Package.get_by_project_and_name(@project.to_param, params[:package],
-                                                   use_source: false, follow_project_links: true, follow_multibuild: true)
-      rescue APIException # why it's not found is of no concern :)
+                                                   follow_project_links: true, follow_multibuild: true)
+      rescue APIException => exception
+        flash[:error] = if exception.is_a?(Package::ReadSourceAccessError) || exception.is_a?(Package::ReadAccessError)
+          "You don't have access to the package \"#{params[:package]}\""
```
hennevogel: Please do not tell people that things are there when they [shouldn't know that](https://openbuildservice.org/help/manuals/obs-reference-guide/cha.obs.concepts.html#id-1.7.6.4.3). A 404 is enough.

dmarcoux: For a package with `sourceaccess` enabled, we display a link to the package in the packages table on the project show page, so users know it's there. They then click on the link and without those changes, it says `Package abc not found in project def`, even though it is listed in the packages table. We should not display the packages with `sourceaccess`, unless users can access it. It's just poor UX.

It's like if I go to a shop which has a sign displaying "SALE SALE SALE, COME BUY MY STUFF". I go in the shop and ask one of the employees "What is on sale?" and then the employee kicks me out of the shop telling me that there isn't a sale.

bgeuken: JFYI, sourceaccess means the access to the packages sources, like packages files, is restricted. The access flag would be for the whole package.

We should at least respond with a 404, if an Package::ReadAccessError was thrown. In case of Package::ReadSourceAccessError error it should be fine to mention the package AFAIK.

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -259,8 +259,13 @@ def require_package
     if params[:package].present?
       begin
         @package = Package.get_by_project_and_name(@project.to_param, params[:package],
-                                                   use_source: false, follow_project_links: true, follow_multibuild: true)
-      rescue APIException # why it's not found is of no concern :)
+                                                   follow_project_links: true, follow_multibuild: true)
+      rescue APIException => exception
+        flash[:error] = if exception.is_a?(Package::ReadSourceAccessError)
+          "You don't have access to the package \"#{params[:package]}\""
```
hennevogel: You don't have access to the **sources** of this package.

DavidKang: @hennevogel, changed

PR #7285, opened 2019-03-27
Merge staging project

bgeuken:


src/api/app/models/concerns/staging_project.rb
```diff
@@ -129,6 +129,20 @@ def staging_project?
...
+
+    staged_requests.each do |staged_request|
+      if staged_request.reviews.where(by_project: name).exists?
+        staged_request.change_review_state(:accepted, by_project: name, comment: "Merged from Staging Project #{name}.")
+      end
+      staged_request.change_state(newstate: 'accepted', comment: "Merged from Staging Project #{name}.")
+    end
+
+    @overall_state = nil
+    staged_requests.delete_all
```
hennevogel: You also need to remove the package sources/links in the staging project.

src/api/app/models/concerns/staging_project.rb
```diff
@@ -129,6 +129,20 @@ def staging_project?
     staging_workflow_id.present?
   end
 
+  def merge_staged_requests
+    return unless overall_state == :acceptable
```
hennevogel: As this changes data I would treat it like a validation and return false and an error on the instance.

bgeuken:
[Merge staging projects into Staging workflow main project](https://trello.com/c/xCnvFLBT/1017-merge-staging-projects-into-staging-workflow-main-project)

bgeuken:
@hennevogel I added the package removal. I also checked the cassettes. All of them are needed and generated on a single test run. With the additional branching code it's even more cassettes:unamused:

docs/api/api/api.txt
```diff
@@ -1960,6 +1960,11 @@ POST /staging/<staging_workflow_project>/staging_projects/<staging_project>/copy
   Copy a staging project
 XmlResult: status_ok
 
+POST /staging/<staging_workflow_project>/staging_projects/merge
+  Merge one or multiple staging projects
```
coolo: That's a rather strange wording. If I merge two staging projects, there is one afterwards. 'to merge == to combine or join together' (cambridge dictionary). But you don't merge staging projects here, you accept requests - and the staging projects number stays untouched - but they are empty afterwards.

synonyms for 'remove content' are:

1. clear
2. consume
3. deplete
4. discharge
5. drain
6. drink
7. dump
8. empty
9. escape
10. evacuate
11. exhaust
12. gut
13. leak
14. purge
15. run out
16. unload
17. vacate
18. decant
19. disgorge
20. ebb
21. eject
22. expel
23. leave
24. release
25. tap
26. unburden
27. void
28. flow out
29. make void
30. pour out
31. rush out
32. use up

coolo: I like disgorge ;)

Makes nice sentenses 'You can please review X so I can disgorge the staging projects?' :)

src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -20,22 +23,46 @@ def create
...
-    staging_workflow.destroy!
+    projects_to_merge.each { |project| StagingProjectMergeJob.perform_later(project.id, User.current.login) }
+    render_ok
   end
 
   private
 
+  def set_staging_projects
+    request_body = request.body.read
+    parsed_xml_hash = Xmlhash.parse(request_body) || {}
```
coolo: please add proper xml validation and remove the || {}

the API has a schema validator built in

src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -20,22 +23,46 @@ def create
...
+    @staging_projects = parsed_xml_hash['staging_project'] || []
```
coolo: No. Just use .elements if you need an array.

src/api/app/jobs/staging_project_merge_job.rb
```diff
@@ -0,0 +1,11 @@
+class StagingProjectMergeJob < ApplicationJob
+  queue_as :quick
```
coolo: You wish this was quick. But in practise this takes several minutes - and I'm not so sure you wouldn't jam the quick queue with that.

src/api/app/models/concerns/staging_project.rb
```diff
@@ -142,6 +169,7 @@ def cache_problems
   end
 
   def state
+    return :merging if Delayed::Job.where("handler LIKE '%StagingMergeProjectJob%'").exists?
```
coolo: This makes all staging projects for all staging workflows :merging if one of them is. Not an option

docs/api/api/api.txt
bgeuken: That was one of the open questions on my list. So far we called this feature Merge staging projects into Staging workflow main project' , but given that the overall_state required for this is called `acceptable` we might better call the action `accept` and the state `accepting`.

src/api/app/models/concerns/staging_project.rb
bgeuken: I know. I just wanted something that remotely works. So I could try out the stuff around this and write some tests.

src/api/app/jobs/staging_project_merge_job.rb
bgeuken: I don't think there is much we can do about right now. Except to not use delayed jobs. But then api requests might fail with a timeout.

bgeuken: Thinking about it... Using `perform_now` might do the trick. I'll check

src/api/app/controllers/staging/workflows_controller.rb
vpereira: @bgeuken 

you can do something like:

```
validate_action create: { method: :post, request: :staging_project }
```

bgeuken:
@coolo Thanks for the review. I did a couple of changes. Could you have another look?

bgeuken:
I think for now querying delayed jobs for knowing if a project is in accepting state is fine, since this is still beta / alpha. But in the long run we should replace all the state checks with a state machine.

src/api/app/jobs/staging_project_accept_job.rb
```diff
@@ -0,0 +1,10 @@
+class StagingProjectAcceptJob < ApplicationJob
+  queue_as :quick
+
+  def perform(project_id, user_login)
```
coolo: why the project by id and the user by login?

docs/api/api/api.txt
```diff
@@ -1948,6 +1948,11 @@ DELETE /staging/<staging_workflow_project>/workflow
   Delete a staging workflow associated to a given project.
 XmlResult: status_ok
 
+POST /staging/<staging_workflow_project>/workflow/accept_staging_projects
```
coolo: I guess it's a little late now, but why you want to accept multiple staging projects through the API? Especially if this runs in background, there is really no reason to force XML onto the API user, I can quicker and easier script a loop using a single project.

Because the posted XML does not conform to the XML schema (which would require a group attribute, as such I doubt your schema is right - where is it anyway?), so e.g.

`/staging/<staging_workflow_project>/staging_projects/<staging_project>/accept` would be the easier route.

src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -20,22 +24,50 @@ def create
...
+        status: 400,
+        errorcode: 'invalid_request',
+        message: "API request contains projects that are not in state acceptable: '#{projects_not_acceptable.join(', ')}'"
       )
       return
     end
-    authorize staging_workflow
 
-    staging_workflow.destroy!
+    projects_to_accept.each { |project| StagingProjectAcceptJob.perform_now(project.id, User.current.login) }
```
coolo: What's the difference between a job performed blocking and a service module? I find this rather strange.

src/api/app/models/concerns/staging_project.rb
```diff
@@ -155,6 +182,7 @@ def cache_problems
   end
 
   def state
+    return :accepting if Delayed::Job.where("handler LIKE '%job_class: StagingMergeProjectJob% project_id: #{id}%'").exists?
```
coolo: How can this be true if you use perform_now? 

> perform_now() Link
Performs the job immediately. The job is not sent to the queueing adapter but directly executed by blocking the execution of others until it's finished.

src/api/app/jobs/staging_project_merge_job.rb
hennevogel: > I don't think there is much we can do about right now. 

We can put it into it's own queue for instance. @eduardoj knows how to start one.

hennevogel:
@bgeuken rebased for you. Is there anything else we can do to help?

src/api/app/jobs/staging_project_accept_job.rb
bgeuken: I just picked it since it's an identifier for users. There isn't any special reason why I picked this and not the id.

docs/api/api/api.txt
hennevogel: For the validation check 0a586593e914b391c0a7f2c9cae6091ee0517fc2

src/api/app/models/concerns/staging_project.rb
bgeuken: Alright. I'll change it back to perfom_later. We can later than change the picked queue or  prioritize this differently.

docs/api/api/api.txt
bgeuken: True. We could adjust this. It's not ideal, but given this is still beta it shouldn't be a problem.

src/api/app/models/concerns/staging_project.rb
hennevogel: `handler` is just the job model serialized as haml. You can basically do 

```
job = YAML::load(blah.handler).job_data
```

for a hash with the job in it, that would contain the specific project. I guess this could also go neatly into the concern.

bgeuken:
Thanks:heart:

I pushed some changes as well (perform_now -> perform_later) using `--force-with-lease` option to avoid conflicts

docs/api/api/api.txt
hennevogel: No, for this release it's okay.

hennevogel:
@coolo please be aware that we need timely reviews for changes to this as we want to deliver this tomorrow at 11:00.

src/api/app/jobs/staging_project_merge_job.rb
hennevogel: I can also do this now, alright?

hennevogel: pushed.

bgeuken: Sure, thanks:thumbsup:

dist/systemd/obs-delayedjob-queue-staging.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description = Open Build Service DelayedJob Queue: staging
+BindsTo = obs-api-support.target
+
+[Service]
+Environment = "RAILS_ENV=production"
+User = wwwrun
+Group = www
+WorkingDirectory = /srv/www/obs/api
+ExecStart = /usr/bin/bundle.ruby2.5 exec script/delayed_job.api.rb --queue=consistency_check start -i 1060
```
DavidKang: @hennevogel, the queue should be `staging` and not `consistency_check`, right?

src/api/app/models/concerns/staging_project.rb
bgeuken: I don't think we can do this, since we need to find the correct delayed job first. What we probably could do is to save the DJ identifier and use it here to check if the job is finished, aka gone, or not.

But what we do right now (checking the job class and the project id) should also be fine.

dist/systemd/obs-delayedjob-queue-staging.service
hennevogel: true!

bgeuken:
@coolo One last review please:smiley_cat:

I've changed the route to `/staging/<staging_workflow_project>/staging_projects/<staging_project>/accept`. The rest is like before.

PR #7291, opened 2019-03-28
[frontend] Add validation for Issue.name

### LINKED ISSUE #1215: Make sure Issue.name is something valid
[Security issue from Hakiri](https://hakiri.io/projects/24a4980a727941/stacks/beceb185080072/builds/9f4cab3b56e087/warnings/d6fc3236809c77): Unescaped model attribute rendered inline in app/controllers/issues_controller.rb

srinidhibs:
This patch is extension of PR#4128 in order to fix test failures.

Fixes #1215

### DIFF
--- a/src/api/app/models/issue.rb
+++ b/src/api/app/models/issue.rb
@@ -10,6 +10,13 @@ class Issue < ApplicationRecord
   belongs_to :issue_tracker
   belongs_to :owner, class_name: 'User'
 
+  validate :name_validation, on: :create
+
+  def name_validation
+    return if issue_tracker.show_label_for(name).match?(issue_tracker.regex)
+    errors.add(:name, "with value \'#{name}\' does not match defined regex #{issue_tracker.regex}")
+  end
+
   scope :stateless, -> { where(state: nil) }
 
   def self.find_or_create_by_name_and_tracker(name, issue_tracker_name, force_update = nil)
@@ -48,9 +55,7 @@ class Issue < ApplicationRecord
   end
 
   def self.valid_name?(tracker, name)
-    # We only verify cve format atm. This should be done via a regexp definition in the
-    # tracker definition in future
-    !tracker.cve? || /^\d\d\d\d-\d+$/ =~ name
+    tracker.show_label_for(name).match?(tracker.regex)
   end
 
   after_create :fetch_issues
--- a/src/api/spec/factories/issue.rb
+++ b/src/api/spec/factories/issue.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :issue do
-    name { Faker::Lorem.words(5).join(' ') }
+    name { Faker::Base.numerify('##-####') }
 
     factory :issue_with_tracker do
       issue_tracker
--- a/src/api/spec/factories/issue_tracker.rb
+++ b/src/api/spec/factories/issue_tracker.rb
@@ -6,7 +6,7 @@ FactoryBot.define do
     url { Faker::Internet.url('example.com') }
     show_url { Faker::Internet.url('example.com') }
     regex { 'gh#(\d+)' }
-    label { Faker::Lorem.word }
+    label { 'gh#@@@' }
     issues_updated { Time.now }
   end
 end
--- a/src/api/spec/mixins/parse_package_diff_spec.rb
+++ b/src/api/spec/mixins/parse_package_diff_spec.rb
@@ -121,7 +121,7 @@ RSpec.describe ParsePackageDiff do
 
       let(:issue_tracker) { IssueTracker.first }
       let(:issue) { create(:issue, issue_tracker: issue_tracker) }
-      let(:deleted_issue) { create(:issue, name: 'deleted', issue_tracker: issue_tracker) }
+      let(:deleted_issue) { create(:issue, name: 1234, issue_tracker: issue_tracker) }
 
       subject { instance_with_parse_package_diff_support.sorted_filenames_from_sourcediff(package_diff).first['issues'] }
 
--- a/src/api/spec/models/issue_spec.rb
+++ b/src/api/spec/models/issue_spec.rb
@@ -21,4 +21,24 @@ RSpec.describe Issue do
       expect(issue_tracker).to have_received(:fetch_issues)
     end
   end
+
+  describe 'validate' do
+    let!(:issue_tracker) { create(:issue_tracker) }
+    let!(:issue_tracker_v1) { create(:issue_tracker, name: 'v1_tracker', regex: '([BD]-[\d]+)', label: '(B-@@@)') }
+    let!(:issue_tracker_cve) { create(:issue_tracker, name: 'cve_tracker', regex: '^(?:cve|CVE)-(\d\d\d\d-\d+)', label: 'CVE-@@@') }
+
+    let!(:issue) { create(:issue, name: '1234', issue_tracker: issue_tracker) }
+    let!(:issue_v1) { create(:issue, name: '1234', issue_tracker: issue_tracker_v1) }
+    let!(:issue_cve) { build(:issue, name: 'CVE-2019-12345', issue_tracker: issue_tracker_cve) }
+
+    it 'issue name should be valid' do
+      expect(issue).to be_valid
+    end
+    it 'V1 style should be valid name' do
+      expect(issue_v1).to be_valid
+    end
+    it 'CVE-XXXX-YYYY should be an invalid name' do
+      expect { issue_cve.save! }.to raise_error(ActiveRecord::RecordInvalid, /does not match defined regex/)
+    end
+  end
 end
--- a/src/api/test/unit/attrib_test.rb
+++ b/src/api/test/unit/attrib_test.rb
@@ -128,7 +128,8 @@ class AttribTest < ActiveSupport::TestCase
     attrib_type = AttribType.new(attrib_namespace: @namespace, name: 'AttribIssues')
     attrib = Attrib.new(attrib_type: attrib_type, project: Project.first)
 
-    attrib.issues << Issue.new(name: '12345')
+    bnc = IssueTracker.find_by_name('bnc')
+    attrib.issues << Issue.new(name: '12345', issue_tracker: bnc)
     assert_not attrib.valid?
     assert_equal ["can't have issues"], attrib.errors.messages[:issues]
     attrib_type.issue_list = true


src/api/app/models/issue.rb
```diff
@@ -10,6 +10,8 @@ class NotFoundError < APIError
   belongs_to :issue_tracker
   belongs_to :owner, class_name: 'User'
 
+  validates :name, format: { with: /\A[\d\-]+\z/, message: 'only digits and hyphens' }
```
coolo: This is wrong. `CVE-2018-2842` is a valid issue name (please make sure to extend the model test for this)

srinidhibs: Sorry about that! I simply copied the changes from PR#4128. I'm quite new to rspec and have tried my best to write a model test. Please suggest any improvements.

src/api/app/models/issue.rb
```diff
@@ -10,6 +10,8 @@ class NotFoundError < APIError
   belongs_to :issue_tracker
   belongs_to :owner, class_name: 'User'
 
+  validates :name, format: { with: /\A(?:cve-)?(?:\w)?[\d\-]+\z/i, message: 'can only contain digits and hyphens' }
```
coolo: Isn't the message wrong here now?

srinidhibs: I seem to have missed updating this PR. In case of CVE issues, does `name` contain the **CVE-** prefix? I ask this because, if I understood correctly while working on this PR, that `name` will not contain the prefix. I found that `IssueTracker.show_label_for` adds the prefix based on what is defined in `<label>` element of issue tracker:

```xml
  <issue-tracker>
    <name>cve</name>
    <kind>cve</kind>
    <description>CVE Numbers</description>
    <url>http://cve.mitre.org/</url>
    <show-url>http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-@@@</show-url>
    <regex>(?:cve|CVE)-(\d\d\d\d-\d+)</regex>
    <label>CVE-@@@</label>
    <enable-fetch>true</enable-fetch>
  </issue-tracker>
```

I also checked in the database of my private Build Service instance and I don't see **CVE-** as part of `name`:

```ruby
> IssueTracker.find_by_name('cve').issues.last
=> #<Issue id: 139607, name: "2019-9641", issue_tracker_id: 23, summary: nil, owner_id: nil, created_at: "2019-04-04 10:48:26", updated_at: "2019-04-04 10:48:26", state: nil>
```

Shall I remove the **CVE-** matcher from the regex?

coolo: Well, as we didn't have a validation before, it's unclear if the data is correct or not :(

But build.opensuse.org has 9231 cve issues that start with CVE- and 1 that starts with cve- - and 149T issues only having numbers.

The latest isn't that old:
=> #<Issue id: 635613, name: "CVE-0456-6807", issue_tracker_id: 44, summary: nil, owner_id: nil, created_at: "2018-10-30 15:21:10", updated_at: "2018-10-30 15:21:10", state: nil>

But it indeed looks wrong. so just erase it - and the invalid entries will need to be sorted out later.

coolo: On top of that: CVE-0456-6807 isn't even a valid cve ;)

coolo:
but this also means the cve- test needs to change *not* to accept cve-XXXX

src/api/app/models/issue.rb
bgeuken: I would also limit this to CVE only. There might be OBS instances with other issues trackers that we would otherwise invalidate.

```ruby
validates :name, format: { with: //, message: '' }, on: Proc.new { |issue| issue.issue_tracker.cve? }
```

src/api/spec/factories/issue.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :issue do
-    name { Faker::Lorem.words(5).join(' ') }
+    name { Faker::Base.numerify('##-####') }
```
bgeuken: That's a nice one:+1:

src/api/app/models/issue.rb
bgeuken: In build.o.o we have a few issue that contain `#`, `:` and whitespaces.

src/api/spec/factories/issue.rb
srinidhibs: Thank you! :slightly_smiling_face:

src/api/app/models/issue.rb
srinidhibs: But the name never contains `label` - at least, not based on **latest** code. This validation will be used when new issues are created, not for those values that are already in the database. Or am I missing something?

On my private instance, we already have problems like these:

```ruby
irb(main):001:0> Issue.where("name LIKE 'CVE%'").first.label
=> "CVE-CVE-2014-0138"
irb(main):002:0> Issue.where("name LIKE 'CVE%'").last.label
=> "CVE-CVE-2013-4449"
irb(main):025:0> Issue.where("name regexp '[a-zA-Z].*'").first
=> #<Issue id: 549, name: "index", issue_tracker_id: 15, summary: nil, owner_id: nil, created_at: "2013-01-25 22:16:04", updated_at: "2013-01-25 22:16:04", state: nil>
irb(main):026:0> Issue.where("name regexp '[a-zA-Z].*'").first.label
=> "bnc#index"
```

I believe, these should be fixed via a script like the `db/checker.rb` or via a migration. In newer code, the `name` of the issue is returned **without** the label (although, I'm not able to find exactly where this filter is being applied)

If `name` will not contain CVE, bnc, bsc, bgo, etc., labels anyway in newer versions of OBS, should there be a validation to accommodate these labels **if** the `name` contains them?

srinidhibs: > In build.o.o we have a few issue that contain `#`, `:` and whitespaces.
> 
> The regex values for them would look like:
> "(?:fate|Fate|FATE)\s*#\s*(\d+)"
> "(?:bnc|BNC|bsc|BSC|boo|BOO)\s*[#:]\s*(\d+)"
> ...

So essentially, something like this: :slightly_smiling_face: 

irb(main):029:0> union_re = Regexp.union(IssueTracker.pluck(:regex))
=> /boost\#\(\\d\+\)|bco\#\(\\d\+\)|RT\#\(\\d\+\)|deb\#\(\\d\+\)|fdo\#\(\\d\+\)|GCC\#\(\\d\+\)|bgo\#\(\\d\+\)|bio\#\(\\d\+\)|\(\?:Kernel\|K\|bko\)\#\(\\d\+\)|kde\#\(\\d\+\)|b\?lp\#\(\\d\+\)|Meego\#\(\\d\+\)|bmo\#\(\\d\+\)|\(\?:bnc\|BNC\|\)\\s\*\[\#:\]\\s\*\(\\d\{6,7\}\)|ITS\#\(\\d\+\)|i\#\(\\d\+\)|\(\?:fate\|Fate\|FATE\)\\s\*\#\\s\*\(\\d\+\)|rh\#\(\\d\+\)|bso\#\(\\d\+\)|sf\#\(\\d\+\)|\(\?:bxc\|Xamarin\)\#\(\\d\+\)|\(\?:cve\|CVE\)\-\(\\d\\d\\d\\d\-\\d\+\)|bxo\#\(\\d\+\)|obs\#\(\\d\+\)|build\#\(\\d\+\)|osc\#\(\\d\+\)|lf\#\(\\d\+\)/

coolo: The current validation would also run if you update the old issues, but possibly that's something we can fix - but I wouldn't make the old issues valid.

bgeuken: > The current validation would also run if you update the old issues, but possibly that's something we can fix

Right

> - but I wouldn't make the old issues valid.

We should at least document such a change, including some instruction to identify and fix affected issues, since there might be OBS instances out there that use this feature and have issues that would become invalid.

srinidhibs: > The current validation would also run if you update the old issues, but possibly that's something we can fix - but I wouldn't make the old issues valid.

Understood! I've now modified the validation to use a lambda to merge all regexes defined across all IssueTrackers (this should cover the scenario of private OBS instances having custom issue trackers and regex matches) along with "numbers and hyphens". 

Does this look better now? Or am I still missing something?

srinidhibs:
> but this also means the cve- test needs to change _not_ to accept cve-XXXX

I'm sorry, but I didn't understand this statement! Are you saying that an issue `name` of "CVE-2019-XXXX" - which actually means the full name is CVE-CVE-2019-XXXX along with the "CVE-" label - should not be accepted?

If yes, then I need help in figuring out where exactly is the `label` stripped off of `name` before it reaches the model? If I understand correctly, there is no direct route to **create** an issue. I tried to extend the test to check for invalid CVE name:

```diff
diff --git a/src/api/spec/models/issue_spec.rb b/src/api/spec/models/issue_spec.rb
index 8d64c9748..3d791a34c 100644
--- a/src/api/spec/models/issue_spec.rb
+++ b/src/api/spec/models/issue_spec.rb
@@ -28,9 +28,11 @@ RSpec.describe Issue, vcr: true do
     let!(:issue_v1) { create(:issue, name: 'B-1234', issue_tracker: issue_tracker) }
     let!(:issue_tracker_cve) { create(:issue_tracker, name: 'cve_tracker', regex: '(?:cve|CVE)-(\d\d\d\d-\d+)') }
     let!(:issue_cve) { create(:issue, name: 'CVE-2019-12345', issue_tracker: issue_tracker_cve) }
+    let!(:issue_invalid_cve) { create(:issue, name: 'CVE-CVE-2019-12345', issue_tracker: issue_tracker_cve) }
 
     it { expect(issue).to be_valid }
     it { expect(issue_v1).to be_valid }
     it { expect(issue_cve).to be_valid }
+    it { expect(issue_invalid_cve).not_to be_valid }
   end
 end
```

This test failed! This might be due to the new combined regex match, but I'm not sure.

src/api/app/models/issue.rb
coolo: I find this much worse than what we had before actually. The Issue name's validity depends on the tracker it's part of - matching another regexp doesn't count. So IMO what we have in valid_name? should be the validation.

srinidhibs: > I find this much worse than what we had before actually. The Issue name's validity depends on the tracker it's part of - matching another regexp doesn't count. So IMO what we have in valid_name? should be the validation.

I can't use `name.tracker` during validation as it returns an exception saying "unknown method `tracker` called on String" (I can share the exact error later today)

This is why I had to switch to matching string against all known regexes.

coolo: you need to use a custom validation, sadly. But you can.

srinidhibs: > you need to use a custom validation, sadly. But you can.

Yes, I was talking about [custom validation](https://guides.rubyonrails.org/active_record_validations.html#performing-custom-validations) itself. The problem I'm quoting is about accessing the issue tracker associated with the name. I'll give it another shot either tomorrow or over the weekend.

srinidhibs:
> but this also means the cve- test needs to change _not_ to accept cve-XXXX

I have now made changes so that CVE-XXXX-YYYY is not accepted as an issue `name` and also a model test exists to check the invalidity of this name.

src/api/app/models/issue.rb
srinidhibs: > The current validation would also run if you update the old issues, but possibly that's something we can fix - but I wouldn't make the old issues valid.

I've marked the validation to run **only** when a new Issue is **created**. This should handle existing invalid names that exist in the database. Hope this is acceptable!

srinidhibs: > I find this much worse than what we had before actually. The Issue name's validity depends on the tracker it's part of - matching another regexp doesn't count. So IMO what we have in valid_name? should be the validation.

I have done this now. Please check if this meets your expectations.

src/api/app/models/issue.rb
```diff
@@ -10,6 +10,21 @@ class NotFoundError < APIError
...
+  }
+
+  # rubocop:disable Lint/UselessAssignment, Lint/UnneededCopDisableDirective, Style/RedundantSelf
+  before_validation do
+    self.name = issue_tracker.show_label_for(self.name)
+  end
+
+  after_validation do
+    self.name = name[Regexp.new(self.issue_tracker.regex), 1]
+  end
```
bgeuken: The before / after validation changes look like a workaround. Instead, please fetch the label within the validation. This might mean that you need a custom validation method, e.g. something like this (untested):

```ruby
def name_validation
  return if issue_tracker.show_label_for(self.name) =~ Regexp.new(issue.issue_tracker.regex)
  errors.add(:name, $some_message)
end
```

That's what we also do in some of the other models. If you want to go one step further you could custom validator class as described in the article you posted. But I think the method above would already be good enough.

vpereira:
are we missing something in this PR?

bgeuken:
> are we missing something in this PR?

Just in case anyone else wonders. This part is missing: https://github.com/openSUSE/open-build-service/pull/7291#discussion_r280200231
and we agreed offline that Srinidhi solves this when he has some free cycles.

src/api/app/models/issue.rb
srinidhibs: I hope, this version meets your requirement.

srinidhibs: @bgeuken Do I need to make more changes to this PR?

DavidKang:
Hi @srinidhibs,
 
Could squash all the commits in one? please :smile_cat: 
All the commits, in the end, are about the same.

srinidhibs:
> Hi @srinidhibs,
> 
> Could squash all the commits in one? please smile_cat
> All the commits, in the end, are about the same.

Thank you for looking into this @DavidKang! I had squashed in the past, but had to make changes to meet review comments. Hence, left the individual commits so that I can back out changes easily if needed.

I can squash the commits if the overall changes are acceptable.

vpereira:
> Thank you for looking into this @DavidKang! I had squashed in the past, but had to make changes to meet review comments. Hence, left the individual commits so that I can back out changes easily if needed.
> 
> I can squash the commits if the overall changes are acceptable.

they are acceptable. Please squash it.

src/api/app/models/issue.rb
vpereira: @srinidhibs,

yes, looking the patchinfos, where the tracker is used, they are all done like https://build.opensuse.org/package/view_file/openSUSE:Maintenance:10206/patchinfo/_patchinfo?expand=1

srinidhibs:
> they are acceptable. Please squash it.

@vpereira @DavidKang Apologies for the delay in getting this done! I hope everything is up to your expectations. Please let me know if anything else needs to be improved.

vpereira:
> @vpereira @DavidKang Apologies for the delay in getting this done! I hope everything is up to your expectations. Please let me know if anything else needs to be improved.

great, it looks like you have to run the linter again

srinidhibs:
> > @vpereira @DavidKang Apologies for the delay in getting this done! I hope everything is up to your expectations. Please let me know if anything else needs to be improved.
> 
> great, it looks like you have to run the linter again @srinidhibs . A `rake dev:lint:rubocop:auto_correct` should do the trick

`1298 files inspected, no offenses detected`

Actually, the problem was that I had not rebased my branch with master. `docker-compose*` files were still pointing to `registry.opensuse.org/obs/server/unstable/container/sle12/sp3/containers/openbuildservice/frontend-base:latest` because of which CircleCI [run failed](https://circleci.com/gh/openSUSE/open-build-service/28191).

I've now rebased my branch with `master` and things look good!

hennevogel:
Hooray @srinidhibs :bouquet:

srinidhibs:
> Hooray @srinidhibs bouquet

Thank you @hennevogel @vpereira @bgeuken @DavidKang and @coolo for helping me out!! Really feels good to have contributed something! :-)

PR #7294, opened 2019-03-28
Remove `use_source` in require_package from WebuiController

dmarcoux:
Also improve flash message error and fix some tests

### DIFF
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -210,19 +210,22 @@ class Webui::WebuiController < ActionController::Base
     required_parameters :package
     params[:rev], params[:package] = params[:pkgrev].split('-', 2) if params[:pkgrev]
     @project ||= params[:project]
-    if params[:package].present?
-      begin
-        @package = Package.get_by_project_and_name(@project.to_param, params[:package],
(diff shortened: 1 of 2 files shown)


src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -210,19 +210,22 @@ def require_package
...
+                                                 follow_project_links: true, follow_multibuild: true)
+    rescue APIError => error
+      if [Package::Errors::ReadSourceAccessError, Authenticator::AnonymousUser].include?(error.class)
+        flash[:error] = "You don't have access to the sources of this package: \"#{params[:package]}\""
+        redirect_back(fallback_location: project_show_path(@project))
+        return
+      end
 
-    render nothing: true, status: :not_found
+      raise(ActiveRecord::RecordNotFound, 'Not Found') unless request.xhr?
```
vpereira: if you add the ```before_action check_ajax``` as in ```render_dialog``` this check can go away, no?

DavidKang: @vpereira, no because this is not only for ajax.

dmarcoux: So if this is resolved, then it's not a blocker anymore. Please have a look again @vpereira if something else has to be changed.

PR #7300, opened 2019-03-29
Revert "Swap accept review / request buttons on request show page"

### LINKED ISSUE #7255: [beta] webui: approve/disregard buttons have wrong order
Observed
============
The Decline button is on the left hand-side, which suggests it is the primary action.
buttons

Expected Result
============
The Approve button should be immediately left of the Decline button.

How to Reproduce
===============
1. Go to '....' a submit request where you can render a Disregard action

Further Information
===============
Regression caused by https://github.com/openSUSE/open-build-service/pull/7233

jengelh:
This reverts commit 7b9f86d255e3cd2fc7a32e19da1069bf2eccd9f9.
Focus after the textarea lands on Disregard, which is just weird.
Fixes #7255.

### DIFF
--- a/src/api/app/views/webui2/webui/request/_decision_tab.html.haml
+++ b/src/api/app/views/webui2/webui/request/_decision_tab.html.haml
@@ -16,12 +16,12 @@
           Forward submit request to
           #{project_or_package_link(project: forward[:project], package: forward[:package], short: true)}
     %hr
+  - if state.in?(['new', 'review']) && is_target_maintainer
+    - confirmation = (state == 'review' ? { confirm: 'Do you really want to approve this request, despite of open review requests?' } : {})
+    = submit_tag 'Accept request', name: 'accepted', class: 'btn btn-primary mr-2', data: confirmation
+    - unless is_author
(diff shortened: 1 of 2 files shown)


Ana06:
It seems there is still some discussion in #7255. Can you please add some before after screenshots so that we can discuss this? :pray:

bgeuken:
The screenshot would just show [this screenshot](https://user-images.githubusercontent.com/8861948/54754472-e699a880-4be3-11e9-89b6-52cefcbde729.png) the swapped buttons I belief.
The background of the previous change was that we usually have the primary button on the right, in OBS (new and old UI) and I wanted to streamline this. But obviously this breaks with user habits:/

Another option @coolo suggested would be to align buttons to the right when we have the primary buttons to the right. Though I suspect this wouldn't solve the issue here.

hennevogel:
We are basically breaking habbits to be consistent. That is okay for me.

jengelh:
>We are basically breaking habbits to be consistent.

The two Internet standards are: forward order on left-align/center, and reverse order on right-align. 7b9f86d does neither of the two - it does a totally inconsistent reverse-order left-aligned.

PR #7276, opened 2019-03-26
Staging Workflow Configure UX

DavidKang:
We changed a little bit the interface of the `Configuration` page.

Staging page:

Screenshot-2019-3-29 Open Build Service

Manage Staging:

Screenshot_2019-03-26 Open Build Service(1)

PR #7308, opened 2019-04-01
Address feedback: Staging Workflow Configure UX

saraycp:
This PR supersedes #7276, because I had some troubles trying to fetch it due to its branch name.

It addresses the feedback regarding the name length. Now, long strings will be broken in many lines.

Screenshot-2019-4-1 Open Build Service

It also adds an explanatory text on the creation modal "If the project already exists, it will simply be assigned to the Staging". 

Screenshot-2019-4-1 Open Build Service(3)

@vpereira @coolo @bgeuken  could you please have another look?


bgeuken:
@DavidKang @saraycp Nice work:thumbsup:

PR #7314, opened 2019-04-02
Add a color for the 'accepting' state of the staging workflow

bgeuken:


DavidKang:
This blue colour is too shiny for me, but we can change it afterwards. 
Right now I don't have a better proposal, but maybe this colour [#cbafcc](https://www.color-hex.com/color/cbafcc)?

bgeuken:
I had already changed it, but then I thought that this one a bit greyish. How about this one (ca57ce)?


@DavidKang What do you think?

DavidKang:
@bgeuken, at least for me is better than the blue colour :smile_cat:

PR #7320, opened 2019-04-02
Add the creation of a staging project to its history for the API

dmarcoux:
To follow up on #7214 and #7178.

### DIFF
--- a/src/api/app/services/staging/staging_project_creator.rb
+++ b/src/api/app/services/staging/staging_project_creator.rb
@@ -1,8 +1,9 @@
 module Staging
   class StagingProjectCreator
-    def initialize(request_body, staging_workflow)
+    def initialize(request_body, staging_workflow, user)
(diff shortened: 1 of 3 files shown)


src/api/app/services/staging/staging_project_creator.rb
```diff
@@ -1,17 +1,25 @@
...
     def call
       if @request_body.blank?
         errors << 'Staging projects are empty'
         return self
       end
-      projects = staging_projects
-      projects.each(&:store) if valid?
+
+      if valid?
+        projects = staging_projects
```
DavidKang: @dmarcoux, this line needs to be outside the statement the `if`, otherwise it will be always `valid`.

dmarcoux: Oh, you are right :+1:

PR #7330, opened 2019-04-03
[api] fix another add container special case

adrianschroeter:
When maintnenace updates went out, but container update was left out.
The next maintenancen update canidate will not find the container
candidates still using GA packages.

We come with update project already in that case and need to find the
GA versions via all project links.

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -961,22 +961,23 @@ class Package < ApplicationRecord
   def add_containers(opts = {})
     container_list = {}
 
-    # released maintenance updates first
-    origin_container(local: false).update_instance.binary_releases.where('ISNULL(obsolete_time)').find_each do |binary_release|
-      mc = binary_release.medium_container
-      container_list[mc] = 1 if mc
-    end
-    # GA versions second
-    origin_container(local: false).binary_releases.where('ISNULL(obsolete_time)').find_each do |binary_release|
-      mc = binary_release.medium_container
-      container_list[mc] = 1 if mc
+    # ensure to start with update project
+    update_pkg = origin_container(local: false).update_instance
+    # we need to take update project and all projects linking to into account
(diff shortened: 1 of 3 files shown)


src/api/app/models/package.rb
```diff
@@ -959,22 +959,23 @@ def is_local_link?
...
-    origin_container(local: false).binary_releases.where('ISNULL(obsolete_time)').find_each do |binary_release|
-      mc = binary_release.medium_container
-      container_list[mc] = 1 if mc
+    # ensure to start with update project
+    update_pkg = origin_container(local: false).update_instance
+    # we need to take update project and all projects linking to into account
+    update_pkg.project.expand_all_projects().each do |prj|
+    origin_package = prj.packages.find_by_name(update_pkg.name)
+      next unless origin_package
+      origin_package.binary_releases.where('ISNULL(obsolete_time)').find_each do |binary_release|
```
mschnitzer: If you just want to check if a value is `NULL`, you could write it like this:
```
origin_package.binary_releases.where(obsolete_time: nil)
```

vpereira:
@adrianschroeter could you please check it and see if everything is ok.

adrianschroeter:
I don't see any problem.
Thanks!

adrianschroeter:
any timeline on merging this? because the https://github.com/openSUSE/open-build-service/pull/7562 may depend on this.

maybe we could add the tests via another pull request? (IMHO rspec with cassettes are pretty useless here, since we would not take the real life data from backend into account btw)

vpereira:
> any timeline on merging this? because the #7562 may depend on this.
> 
> maybe we could add the tests via another pull request? (IMHO rspec with cassettes are pretty useless here, since we would not take the real life data from backend into account btw)

I'm merging it today. I was just adding a test case to get all checks happy!

PR #7332, opened 2019-04-04
Migrate maintenance incidents related actions to their own controller

dmarcoux:
What has been done:
- Move actions from `webui/project_controller.rb` to `webui/projects/maintenance_incidents_controller.rb`
- Adapt specs to follow changes (Move specs - Adapt them - Regenerate cassettes)


dmarcoux:
[Move maintenance incident related actions to its own controller](https://trello.com/c/0uuXkiQX/1098-move-maintenance-incident-related-actions-to-its-own-controller)

src/api/app/controllers/webui/projects/maintenance_incidents_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      before_action :require_login, only: [:create, :create_request]
+
+      after_action :verify_authorized, except: [:index, :create, :create_request]
+
+      def index
+        @incidents = @project.maintenance_incidents
+      end
+
+      def create
+        authorize @project, :update?
```
DavidKang: I thinkyou still need `project = Project.get_by_name(params[:ns])`, is different than the `set_project`

src/api/app/controllers/webui/projects/maintenance_incidents_controller.rb
```diff
@@ -0,0 +1,52 @@
+module Webui
+  module Projects
+    class MaintenanceIncidentsController < WebuiController
+      before_action :set_project, only: [:index, :create_request]
+      before_action :lockout_spiders, only: [:index]
+      before_action :require_login, only: [:create, :create_request]
+
+      after_action :verify_authorized, except: [:index, :create, :create_request]
```
bgeuken: We should verify this for `:create`

dmarcoux: I realized, it should also be the case for `create_request`. Login is required, but then we never check authorization.

PR #7337, opened 2019-04-05
Allow acceptance of requests when they are in 'accepting' state

bgeuken:
Because the overall_state method queries delayed jobs we need to
consider the 'accepting' state as one of the "mergable" states.


bgeuken:
We started working on a better solution for handling the 'accepting' state. But for now this is good enough to fix the bug.

bgeuken:
This change is now part of PR7373

PR #7339, opened 2019-04-05
Patchinfo refactor

saraycp:
- Rename controller actions as CRUD.
- Use more meaningful names for some methods.
- For the moment, read_patchinfo (now called extract_patchinfo_data) is called from a before_action.

We'll address code climate complaints in a different PR.


saraycp:
[Refactor patchinfo controller](https://trello.com/c/4JGl1Kkx/1094-refactor-patchinfo-controller)

src/api/app/controllers/webui/patchinfo_controller.rb
```diff
@@ -25,10 +25,11 @@ def create
     redirect_to edit_patchinfo_path(project: @project, package: @package)
   end
 
-  def updatepatchinfo
+  def fetch_issues
     authorize @project, :update?
 
-    Patchinfo.new.cmd_update_patchinfo(params[:project], params[:package])
+    message = 'updated via fetch_issues call'
+    Patchinfo.new.cmd_update_patchinfo(params[:project], params[:package], message)
```
bgeuken: :thinking: I would replace the `message` variable with the actual message. Given that it's only used once.

src/api/app/controllers/webui/patchinfo_controller.rb
```diff
@@ -22,32 +23,30 @@ def new_patchinfo
...
       redirect_to(controller: 'package', action: 'show', project: @project, package: @package) && return
     end
-
-    read_patchinfo
-    @binaries.each { |bin| @binarylist.delete(bin) }
+    redirect_to edit_patchinfo_path(project: @project, package: @package)
   end
 
-  def updatepatchinfo
+  def fetch_issues
```
dmarcoux: Are we really fetching issues here? It seems more like it's updating.

src/api/app/models/patchinfo.rb
```diff
@@ -199,14 +199,14 @@ def create_patchinfo(project, pkg_name, opts = {})
     { targetproject: @pkg.project.name, targetpackage: @pkg.name }
   end
 
-  def cmd_update_patchinfo(project, package)
+  def cmd_update_patchinfo(project, package, message = 'updated via updatepatchinfo call')
```
dmarcoux: In the message, `updatepatchinfo` should be `fetch_issues` since it was changed in the controller.

saraycp: `cmd_update_patchinfo` is called from both API and Webui controllers. In the API endpoint it's used like: `# POST /source/<project>/<package>?cmd=updatepatchinfo`and we don't want to change the name there. For this reason, we pass a different message using the name `fetch_issues`if it comes from the webui or `updatepatchinfo`if it comes from the API.

src/api/app/controllers/webui/patchinfo_controller.rb
saraycp: done, thanks!

src/api/app/controllers/webui/patchinfo_controller.rb
saraycp: We'll re-think this, maybe 'fetch' is not the best word.

src/api/app/assets/javascripts/webui/application/patchinfo.js.erb
```diff
@@ -51,6 +51,7 @@ $(document).on('click', 'a[data-add-issue]', function() {
   return false;
 });
 
+/* FIXME: it seems this function is not used, delete after we ensure it's true */
```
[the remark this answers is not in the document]
saraycp: Thanks! good to know that you didn't find any call to this either.

src/api/app/controllers/webui/patchinfo_controller.rb
```diff
@@ -22,32 +23,30 @@ def new_patchinfo
...
-    redirect_to action: 'edit_patchinfo', project: @project, package: @package
+    message = 'updated via fetch_issues call'
+    Patchinfo.new.cmd_update_patchinfo(params[:project], params[:package], message)
+    redirect_to edit_patchinfo_path(project: @project, package: @package)
   end
 
-  def edit_patchinfo
-    read_patchinfo
+  def edit
+    # TODO: check that @tracker has sense if it's comming from create (new_patchinfo) action
```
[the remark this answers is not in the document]
saraycp: That TODO is not part of this PR.

src/api/app/controllers/webui/patchinfo_controller.rb
saraycp: Changed to `update_issues`

saraycp:
@dmarcoux please have another look

PR #7369, opened 2019-04-08
Update flash messages that were using incorrect flash message type

### LINKED ISSUE #7124: Some flash messages have the wrong type
We have a few actions in which we use the wrong flash message type, mostly when creating/updating. It is currently using `:notice` while it should use `:success`. I've noticed that in a few Bootstrap-related PRs lately and instead of always pointing it out, we should fix it once and for all.

I expect flash messages to have the right type depending on their context. So when creating/updating, it should be `:success`, for informational messages `:notice` and `:error` when something bad happens.

Searching in all the controllers and changing the type, then adapting the specs should be what we need to do to close this issue.

bgeuken:
Fixes #7124


bgeuken:
The tests [passed](https://circleci.com/gh/bgeuken/open-build-service/tree/fix_issue_7124) but were not properly propergated here. Probably because they ran in my fork

PR #7373, opened 2019-04-08
Allow acceptance of requests when they are in 'accepting' state

saraycp:
The staging project's overall_state didn't change to the correct state after the delayed job.

While the delayed job exists, the overall_state is 'accepting', so we need the state to change from 'accepting' to 'accepted' right before the delayed job finish (we need to consider the 'accepting' state as one of the "mergable" states).

Since the accepting state relies on the existence of delayed jobs, we need to make sure the staging projects acceptance only happens via delayed jobs. So we moved accept method to the delayed job.


adrianschroeter:
just to be sure, this is not about classic requests, right? Accepting a former request already in accepted state is considered to be harmful there.

DavidKang:
@adrianschroeter, no, this is regarding the `StagingWorkflow` and the `accepting` is related to a `StagingProject`

coolo:
The description is very confusing. Why would you want to accept a staging project that already has a delayed job running accepting it?

coolo:
thanks

PR #7374, opened 2019-04-08
Migrate patchinfo#edit to Bootstrap

dmarcoux:
Improved UX:
- It is clear which fields are required and what is the minimal length
- The packager field is providing user auto-completion (which is now clearly marked as required across the Bootstrap UI, not only in patchinfo)
- Clarify help messages
- Buttons in the binaries section have a title to indicate what they do (in case someone isn't sure / doesn't know)

Screenshots:

Autocomplete changes for all Bootstrap views (as explained above)
autocomplete-other-views

Patchinfo - Bento (to compare)
patchinfo-bento

Patchinfo - Bootstrap
patchinfo-bootstrap

### DIFF
--- a/src/api/app/assets/javascripts/webui2/patchinfo.js
+++ b/src/api/app/assets/javascripts/webui2/patchinfo.js
@@ -1,0 +1,128 @@
+function setupPatchinfo() { // jshint ignore:line
+  addIssuesAjaxBefore();
+  addIssuesAjaxSuccess();
+  addIssuesAjaxComplete();
+  deleteIssueClick();
+  patchinfoBlockChange();
+  patchinfoBinariesEvents();
+}
+
+function addIssuesAjaxBefore() {
+  $('#add-issues').on('ajax:before', function() {
+    var issues = $('#issue_ids').val();
+    if (issues.length === 0) {
+      return false;
+    }
+
+    var element = $(this);
+    element.children('i.fas.fa-plus-circle').addClass('d-none');
+    element.children('i.fas.fa-spin').removeClass('d-none');
+
+    issues = $.unique(issues.replace(/ /g, '').split(','));
+
+    element.data('params', { issues: issues, project: element.data('project') });
+  });
+}
+
+function addIssuesAjaxSuccess() {
+  $('#add-issues').on('ajax:success', function(event, data) {
+    $('#ajax-error-message').text(data.error);
+    if (data.error !== '') {
+      return;
+    }
+
+    addIssues(data.issues);
+  });
+}
+
+function addIssues(issues) {
+  $.each(issues, function() {
+    var issueTracker = this[0];
+    var issueId = this[1];
+    var issueUrl = this[2];
+    var issueSum = this[3];
+
+    if ($('li#issue_' + issueTracker + '_' + issueId).length > 0) {
+      return;
+    }
+
+    addIssuesHtml(issueTracker, issueId, issueUrl, issueSum);
+  });
+}
+
+function addIssuesHtml(issueTracker, issueId, issueUrl, issueSum) {
+  var identifier = issueTracker + '_' + issueId;
+
+  $('#issues').append(
+    $('<div>', { id: 'issue_' + identifier, class: 'list-group-item flex-column align-items-start issue-element' }).append(
+      $('<div>', { class: 'd-flex w-100 mb-1' }).append(
+        $('<a>', { href: issueUrl, target: '_blank', rel: 'noopener' }).append(
+          $('<i>', { class: 'fa fa-bug text-danger' })
+        ).append(' ' + issueTracker + '#' + issueId),
+        $('<a>', { id: 'delete_issue_' + identifier, href: '#', title: 'Delete', class: 'ml-1' }).append(
+          $('<i>', { class: 'fas fa-times-circle text-danger' })
+        ),
+        $('<input>', { type: 'hidden', name: 'patchinfo[issueid][]', id: 'issueid_' + identifier, value: issueId, multiple: true }),
+        $('<input>', { type: 'hidden', name: 'patchinfo[issuetracker][]', id: 'issuetracker_' + identifier, value: issueTracker, multiple: true }),
+        $('<input>', { type: 'hidden', name: 'patchinfo[issueurl][]', id: 'issueurl_' + identifier, value: issueUrl, multiple: true }),
+        $('<input>', { type: 'hidden', name: 'patchinfo[issuesum][]', id: 'issueurl_' + identifier, value: issueSum, multiple: true })),
+      $('<small>', { class: 'text-muted' }).append(issueSum)
+    )
+  );
+}
+
+function addIssuesAjaxComplete() {
+  $('#add-issues').on('ajax:complete', function() {
+    var element = $(this);
+    element.children('i.fas.fa-spin').addClass('d-none');
+    element.children('i.fas.fa-plus-circle').removeClass('d-none');
+    $('#issue_ids').val('');
+  });
+}
+
+function deleteIssueClick() {
+  $(document).on('click', 'a[id^="delete_issue_"]', function(event) {
+    event.preventDefault();
+
+    $(this).parents('.issue-element').remove();
+  });
+}
+
+function patchinfoBlockChange() {
+  $('#patchinfo_block').change(function() {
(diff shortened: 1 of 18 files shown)


dmarcoux:
[Create patchinfo](https://trello.com/c/BSTVviHr/1091-create-patchinfo)

coolo:
pics or it didn't happen! :)

dmarcoux:
@coolo I will add at least one when this PR won't be a draft anymore.

dmarcoux:
What should I do with the CodeClimate issue? Should I address or approve it? I find splitting the function up doesn't really help. I would prefer to approve it.

coolo:
For what it's worth: I'm with codeclimat on this one. This is outside my personal climate:

```
          )
        );
      }
    });
  });
}
```

coolo:
(being introduced to programming in university with [Scheme]( https://en.wikipedia.org/wiki/Scheme_(programming_language)) didn't really help making me easy with closing parentheses)

saraycp:
@dmarcoux What about moving all the explanations right under the label (before the textarea)? Otherwise you can start writing and afterwards you read the explanation.

dmarcoux:
> @dmarcoux What about moving all the explanations right under the label (before the textarea)? Otherwise you can start writing and afterwards you read the explanation.

I believe it's a common pattern to have the help/explanation under the field.

dmarcoux:
@coolo Please have another look.

coolo:
I find this an improvement - not that I digged into the functions, but I feel now I could.

src/api/app/assets/javascripts/webui2/patchinfo.js
```diff
@@ -0,0 +1,118 @@
+function setupPatchinfo() { // jshint ignore:line
```
Ana06: [nitpick] We should consider alternatives which doesn't involved that much JavaScript

Ana06:
@saraycp do you have concerns with this PR (I am not completely sure if you are happy with [@dmarcoux's answer](https://github.com/openSUSE/open-build-service/pull/7374#issuecomment-484104075)) or can I click the merge button? :smile:

src/api/app/assets/javascripts/webui2/patchinfo.js
dmarcoux: I managed to reduce the amount of JS with rails-ujs, but it's true that this form depends a lot on JS. What could we do instead? For the AJAX/click/change events, I don't really see a way to avoid that. Regarding the `addIssues` function and appending HTML, this could be improved with templates like Mustache, but it's not something we do currently.

bgeuken:
One minor thing I noticed, the url to the edit page uses query parameters like `edit?package=patchinfo&project=home%3AAdmin`.
Since we just updated the patchinfo controller actions I would have thought that we also updated the webui routes.

bgeuken:
Seeing the [show version](#7416) of patchinfo I realized that both views look very different now. So when I, as a user, look at one [patchinfo](https://user-images.githubusercontent.com/1212806/56308571-e55a8d80-6147-11e9-95b3-e10cdedfdfdd.png) I might notice that the bug list at the bottom right is missing one bug.
After clicking on the 'Edit Patchinfo' link I'll have to parse the whole page again to find the area where I can add bugs.

I know this isn't the most visited view / most used workflow right now. So I just added this as a comment, but I think we should consider such things in future. If time permits it might even make sense to adjust the view already. But that's not my call :smile:

dmarcoux:
The layout of the patchinfo#show is nicer than before to display information, but it isn't for a form since it's better UX to go vertically through a form instead of going through it vertically, then horizontally, then vertically again. This would be the case if I used the same layout due to the `Fixed Bugs` and `Selected Binaries` boxes.

saraycp:
> @saraycp do you have concerns with this PR (I am not completely sure if you are happy with [@dmarcoux's answer](https://github.com/openSUSE/open-build-service/pull/7374#issuecomment-484104075)) or can I click the merge button? smile

@Ana06 @dmarcoux if it's a tendency, fine for me. I don't want to be old-fashioned :stuck_out_tongue_winking_eye:

bgeuken:
> Edit: As we discussed in our standup, we will eventually aim to get inline editing, but for now, it will stay as it is.

That's why I raised this issue, but didn't make it a change request:-)

src/api/app/views/webui2/webui/patchinfo/form/_issues.html.haml
```diff
@@ -0,0 +1,31 @@
+.form-group
+  = form.label(:issues, class: 'd-block')
+
+  %ul#issues
+    - issues.each do |issue|
```
bgeuken: When submitting an invalid patchinfo I get an `undefined method `each' for nil:NilClass` caused by this line.

vpereira:
Ok as User, there are some issues with that page that may or may not be addressed.

1) Nobody edit the issues summary since it comes from Bugzilla, and there are some internal tools which match the issues summary from patchinfo with bugzilla issues. AFAIK, L3 does it. What users can do is remove an issue or add new issues. 

2) The binaries section, could be hidden under "advanced options" if possible. I don't remember a case where we use it consistently.

3) Better align the fields to save some spaces and have description as text area with enough space to add long texts. Normally patchinfos are really verbose, with many issues (specially true for SLE updates) . The "old view" is actually "well designed" (thinking about maintenance team usability) 

4) It would be great to have the name of the package for which the patchinfo is being written. For example this incident https://build.suse.de/project/show/SUSE:Maintenance:10415 has 3 packages.

src/api/app/views/webui2/webui/patchinfo/edit.html.haml
```diff
@@ -0,0 +1,47 @@
+- @pagetitle = "Edit Patchinfo for #{@project}"
```
vpereira: Ok as User, there are some issues with that page that may or may not be addressed.

Nobody edit the issues summary since it comes from Bugzilla, and there are some internal tools which match the issues summary from patchinfo with bugzilla issues. AFAIK, L3 does it. What users can do is remove an issue or add new issues.

The binaries section, could be hidden under "advanced options" if possible. I don't remember a case where we use it consistently.

Better align the fields to save some spaces and have description as text area with enough space to add long texts. Normally patchinfos are really verbose, with many issues (specially true for SLE updates) . The "old view" is actually "well designed" (thinking about maintenance team usability)

It would be great to have the name of the package for which the patchinfo is being written. For example this incident https://build.suse.de/project/show/SUSE:Maintenance:10415 has 3 packages.

src/api/app/views/webui2/webui/patchinfo/form/_issues.html.haml
bgeuken: Fixed

bgeuken:
@dmarcoux JFYI, I fixed the [bug](https://github.com/openSUSE/open-build-service/pull/7374#discussion_r277573686) and the linter offense. I also rebased on master to solve the hakiri warning (caused by a missing nokogiri gem update)

src/api/app/views/webui2/webui/patchinfo/edit.html.haml
dmarcoux: Regarding editing the description of an issue, I kept the same functionality, just made it obvious to do it. It was previously possible to edit the description, but you had to know that clicking on the description would display an input which then allow you to edit the description. So for now, I would keep this functionality until we redesign this page/workflow (if we ever do it).

For the binaries section, I would not hide a section of a form. If this section is not needed in some patchinfo, the field is not required, so it can be ignored. This goes back to the possible redesign of this page/workflow.

Concerning this paragraph:

> Better align the fields to save some spaces and have description as text area with enough space to add long texts. Normally patchinfos are really verbose, with many issues (specially true for SLE updates) . The "old view" is actually "well designed" (thinking about maintenance team usability)

I'm all ears, but please, be clear on what is to be improved. What do you mean with better align the fields to save some spaces? The fields are fully using the available space (in comparison to the old view, which didn't at all). The `textarea` elements now have a few rows (thanks to @DavidKang for his commit). As for the old view is well designed, I basically took the form from the old view and kept the same functionality, then added a bunch of UX improvements (they're listed in the PR's description). Heck, the old view didn't even have any indication to show what is required (which is basic design IMHO).

About having the name of the package(s) for which the patchinfo is being written, the patchinfo is for a project, not for a package. The project will contain the patchinfo and packages. I would consider this feature for a redesign, not for this migration.

PR #7381, opened 2019-04-09
Fix the disabled state of CodeMirror editor buttons

### LINKED ISSUE #7371: File can be saved even if the button is greyed-out
# Issue Description
I opened the file editor in OBS and I was able to press 'Save' even if I didn't edit anything and the button was greyed-out.

Expected Result
============
I should not be able to press save when nothing has changed.

How to Reproduce
===============
1. Open the file editor
2. Don't change anything and press save

bgeuken:
This ensures that undo, redo and save buttons have the correct state
when changing the editor text and before / after saving it.

Fixes #7371

### DIFF
--- a/src/api/app/assets/javascripts/webui2/cm2/use-codemirror.js
+++ b/src/api/app/assets/javascripts/webui2/cm2/use-codemirror.js
@@ -2,6 +2,7 @@ var editors = new Array();
 
 var unstaged = false;
 
+// Show a dialog when there are unsave changes (redo or undo) and the user is about to leave the page
 window.onbeforeunload = function() {
   if (unstaged) {
     return true;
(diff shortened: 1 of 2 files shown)


src/api/app/assets/javascripts/webui2/cm2/use-codemirror.js
```diff
@@ -2,6 +2,7 @@ var editors = new Array();
 
 var unstaged = false;
 
+// Show a dialog when there are unsave changes (redo or undo) and the user is about to leave the page
```
dmarcoux: unsave**d**

src/api/app/views/webui2/shared/_editor.html.haml
```diff
@@ -9,14 +9,14 @@
     .sticky-top.bg-light{ id: "top_#{uid}" }
       .card-header.d-flex.justify-content-between
         .input-group.input-group-sm
-          %button.btn.btn-danger.btn-sm.disabled.save{ id: "save_#{uid}" }
+          %button.btn.btn-danger.btn-sm.save{ id: "save_#{uid}", disabled: true }
```
Ana06: why do you need to use disabled as prop instead of as class? :thinking:

bgeuken: Because that's how buttons are getting [disabled](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/disabled):smile:

bgeuken:

PR #7386, opened 2019-04-09
Deprecate Ratings

hennevogel:


hennevogel:
> AFAIK we need to keep the `*.xsd` files for the request validation.

Not really, at least we don't test for it. Anyway, added again.

PR #7389, opened 2019-04-10
Add validation to patchinfo in the Rails way

saraycp:
As Patchinfo is not an ApplicationRecord, we use ActiveModel::Model to
use Rails validators and set the required attr_accessors.

It also refactors patchinfo initialize method, passing data as argument.


saraycp:
[Refactor patchinfo controller](https://trello.com/c/4JGl1Kkx/1094-refactor-patchinfo-controller)

src/api/app/models/patchinfo.rb
```diff
@@ -37,14 +38,23 @@ class TrackerNotFound < APIError
   CATEGORIES = (CATEGORY_COLORS.keys << '').freeze
 
   attr_reader :document
+  attr_writer :data
+  attr_accessor :summary, :description, :packager
+
+  validates :summary, length: { minimum: 11, message: 'is too short (should have more than 10 signs)' }
```
hennevogel: s/signs/characters/

src/api/app/models/patchinfo.rb
```diff
@@ -247,4 +257,10 @@ def issues_by_tracker
     end
     issues_by_tracker
   end
+
+  def description_length
+    unless description && description.length > [summary.length, 50].max
```
hennevogel: I find this hard to understand and doubt it makes sense.

hennevogel:
Please also look into input [patterns](https://www.w3.org/TR/2009/WD-html5-20090825/forms.html#attr-input-pattern) for the UI...

src/api/app/models/patchinfo.rb
saraycp: done

src/api/app/models/patchinfo.rb
saraycp: @hennevogel Now we don't compare description length with summary length, only check if description is bigger than 50. Better now?

saraycp:
@hennevogel we have used `minlength` attribute instead of `pattern`. 

We also plan to keep the current views like they are and create new views and the controller for bootstrap.

PR #7390, opened 2019-04-10
Move amqp configuration options inside development section

DavidKang:

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -122,15 +122,22 @@ namespace :dev do
       Rake::Task['haml_lint'].invoke
     end
   end
+
   namespace :ahm do
     desc 'Configure the rails app to publish application health monitoring stats'
     task :configure do
+      unless Rails.env.development?
+        puts "You are running this rake task in #{Rails.env} environment."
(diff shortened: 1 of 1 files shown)


src/api/lib/tasks/dev.rake
```diff
@@ -122,15 +122,22 @@ namespace :dev do
...
+        return
+      end
+
       copy_example_file('config/options.yml')
       options_yml = YAML.load_file('config/options.yml') || {}
-      options_yml['amqp_namespace'] = 'opensuse.obs'
-      options_yml['amqp_options'] = { host: 'rabbit', port: '5672', user: 'guest', pass: 'guest', vhost: '/' }
-      options_yml['amqp_exchange_name'] = 'pubsub'
-      options_yml['amqp_exchange_options'] = { type: :topic, persistent: 'true', passive: 'true' }
+      options_yml[Rails.env]['amqp_namespace'] = 'opensuse.obs'
```
hennevogel: Why make this env dependent if you only allow one env?

coolo: everything in options.yml is env dependent (or is in 'default')

PR #7391, opened 2019-04-10
Add some DelayedJob instrumentation

hennevogel:
So we can know what's going on in the queue and how jobs perform.

### DIFF
--- a/src/api/config/clock.rb
+++ b/src/api/config/clock.rb
@@ -27,6 +27,10 @@ module Clockwork
     SendEventEmailsJob.perform_later
   end
 
+  every(30.seconds, 'instrument delayed job queue', :if => lambda { |_| CONFIG['influxdb_hosts'] }) do
+    InstrumentationDelayedJobQueue.perform_later
+  end
+
   every(49.minutes, 'rescale history') do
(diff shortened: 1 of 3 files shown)


src/api/config/clock.rb
```diff
@@ -27,6 +27,10 @@ module Clockwork
     SendEventEmailsJob.perform_later
   end
 
+  every(30.seconds, 'instrument delayed job queue') do
```
hennevogel: Not sure 30 seconds is the right frequency...

coolo: I would say it's fine. You do not fancy supergreat granuality - but rather trends for alarms. 

But I would make sure it's not trying to report anything to a not configured influxdb. a9ac08cec02481f89958ea74f6007076082118bf uses CONFIG['influxdb_hosts'] for that

hennevogel: I think this is taken care of in `InfluxDB::Rails`, let me check.

hennevogel:
autoload fails

src/api/config/clock.rb
hennevogel: See now...

hennevogel:
I might have a better idea

PR #7396, opened 2019-04-10
Add some DelayedJob instrumentation

hennevogel:
So we can know what's going on in the queue and how jobs perform.

### DIFF
--- a/src/api/config/initializers/influxdb_dj_subscriber.rb
+++ b/src/api/config/initializers/influxdb_dj_subscriber.rb
@@ -1,0 +1,36 @@
+if CONFIG['influxdb_hosts']
+  ActiveSupport::Notifications.subscribe(/active_job/) do |*args|
+    event = ActiveSupport::Notifications::Event.new(*args)
+    job = event.payload[:job]
+    exception = event.payload[:exception_object]
+
+    def write_queue_stats(state, job)
+      InfluxDB::Rails.client.write_point('active_job.queue',
+                                         tags: { state: state, job: job.class.name, queue: job.queue_name },
+                                         values: { value: 1 })
+    end
+
+    def write_performance_stats(job, event)
+      InfluxDB::Rails.client.write_point('active_job.performance',
+                                         tags: { job: job.class.name, queue: job.queue_name },
+                                         values: { value: event.duration })
+    end
+
+    if event.name == 'perform.active_job' && exception
+      InfluxDB::Rails.client.write_point('active_job.queue',
+                                         tags: { state: :failed, job: job.class.name, queue: job.queue_name },
+                                         values: { value: 1, id: job.job_id })
+      return true
+    end
+
+    case event.name
+    when 'enqueue.active_job'
+      write_queue_stats(:queued, job)
+    when 'perform_start.active_job'
+      write_queue_stats(:running, job)
+    when 'perform.active_job'
+      write_queue_stats(:succeeded, job)
+      write_performance_stats(job, event)
+    end
+  end
+end


src/api/config/initializers/influxdb_dj_subscriber.rb
```diff
@@ -0,0 +1,40 @@
+if CONFIG['influxdb_hosts']
+  ActiveSupport::Notifications.subscribe(/active_job/) do |*args|
+    event = ActiveSupport::Notifications::Event.new(*args)
+    job = event.payload[:job]
+    exception = event.payload[:exception_object]
+
+    def write_queue_stats(state, job)
+      puts "write_queue_stats #{state}"
```
coolo: why puts and not Rails.logger?

src/api/config/initializers/influxdb_dj_subscriber.rb
```diff
@@ -0,0 +1,40 @@
...
+      InfluxDB::Rails.client.write_point('active_job.queue',
+                                         tags: { state: state, job: job.class.name, queue: job.queue_name },
+                                         values: { value: 1 })
+    end
+
+    def write_performance_stats(job, event)
+      puts "write_performance_stats #{event.duration}"
```
coolo: same here

src/api/config/initializers/influxdb_dj_subscriber.rb
```diff
@@ -0,0 +1,40 @@
...
+      InfluxDB::Rails.client.write_point('active_job.performance',
+                                         tags: { job: job.class.name, queue: job.queue_name },
+                                         values: { value: event.duration })
+    end
+
+    if event.name == 'perform.active_job' && exception
+      InfluxDB::Rails.client.write_point('active_job.queue',
+                                         tags: { state: :failed, job: job.class.name, queue: job.queue_name, id: job.job_id },
```
coolo: careful! if you use the job id as a tag you're breaking all retention policies as you can no longer group data points.

src/api/config/initializers/influxdb_dj_subscriber.rb
```diff
@@ -0,0 +1,40 @@
...
+                                         values: { value: 1 })
+      return true
+    end
+
+    case event.name
+    when 'enqueue.active_job'
+      write_queue_stats(:queued, job)
+    # we do not use this currently, we use clockwork instead...
```
coolo: I don't think this comment is still true

hennevogel: Why?

src/api/config/initializers/influxdb_dj_subscriber.rb
hennevogel: unintentional

src/api/config/initializers/influxdb_dj_subscriber.rb
hennevogel: Retention policies are unbreakable :clown_face: This would generate non-downsampleable (cool I just invented a word!)  data, yes.

src/api/config/initializers/influxdb_dj_subscriber.rb
coolo: ok, let me rephrase: the comment is very confusing. I bet there are a lot of other things not used you can put here as comments - just to confuse reviewers :)

src/api/config/initializers/influxdb_dj_subscriber.rb
coolo: Just put it as additional value

src/api/config/initializers/influxdb_dj_subscriber.rb
hennevogel: better?

src/api/config/initializers/influxdb_dj_subscriber.rb
hennevogel: yeah ok...

src/api/config/initializers/influxdb_dj_subscriber.rb
```diff
@@ -0,0 +1,37 @@
...
+      InfluxDB::Rails.client.write_point('active_job.queue',
+                                         tags: { state: :failed, job: job.class.name, queue: job.queue_name },
+                                         values: { value: 1, id: job.job_id })
+      return true
+    end
+
+    case event.name
+    when 'enqueue.active_job'
+      write_queue_stats(:queued, job)
+    # when 'enqueue_at.active_job'
```
bgeuken: This looks like a leftover

hennevogel: I had a comment explaining why this is here but coolo didn't like it :clown_face:

coolo: But in the end @bgeuken and me have the same problem: it's rather puzzling to mention things not used.

hennevogel: removed

PR #7398, opened 2019-04-11
Update bundler version

coolo:
We used a very outdated version that e.g. causes the crash in the rubocop update, so I updated the obs package - but this needs to be aligned with Gemfile.lock (it's soon part of the containers)

### DIFF
--- a/dist/obs-bundled-gems.spec
+++ b/dist/obs-bundled-gems.spec
@@ -51,7 +51,7 @@ This package bundles all the gems required by the Open Build Service
 to make it easier to deploy the obs-server package.
 
 %define rake_version 12.3.2
-%define rack_version 2.0.6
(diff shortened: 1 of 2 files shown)


coolo:
ok, this is some mean chicken/egg problem. Our containers won't build with a mismatch between git and bundler package - and you can't update the git as it's using our containers to verify the bundle.

So you need to trust me here :(

coolo:
boy, is this hard. the bundled_gems service uses a hardcoded bundler version incompatible to this :(

coolo:
trying with 1.17.1 now

coolo:
ok, staying below 2 avoids the chicken

coolo:
Sorry for the self merge, but we need to align with the packages in OSU

PR #7399, opened 2019-04-11
Rename exeption objects to e

coolo:
This is a new default rubocop introduced with 0.67.2 and I agree with it

You can find details at https://rubocop.readthedocs.io/en/latest/cops_naming/#namingrescuedexceptionsvariablename

Please vote @openSUSE/obs-frontend

### DIFF
--- a/dist/openQA_mail_notification.rb
+++ b/dist/openQA_mail_notification.rb
@@ -25,8 +25,8 @@ def get_build_information(version)
     request = Net::HTTP::Get.new(uri.request_uri)
     response = http.request(request)
     JSON.parse(response.body)['jobs'].last
-  rescue Exception => ex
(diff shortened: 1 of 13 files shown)


eduardoj:
> Please vote @openSUSE/obs-frontend

I agree.

hennevogel:
And why is there no cop configuration?

coolo:
It's default - once rubocop releases a fixed version containing this cop, 0.67.2 was rejected due to 2 bugs, but the cop is real. We can have fixes for cops not yet released, no? :)

coolo:
It claims to support it - and also prints out Cs during progress, but doesn't change the code :)

And depfu sent a PR, that I closed due to a bug in that rubocop version

coolo:
I only cherry picked the fixes

PR #7402, opened 2019-04-11
Improve release target ng queries

bgeuken:
* Move check for release target extension into SQL 
* Refactored code of guess_release_target_from_package method and moved
  it into `release_target_ng`.
  * Instead of filtering the flag query result afterwards we do
    this as part of the SQL query.
  * The method was using a hash for mapping the queried results.
    This code got mostly removed. The remaining hashing is now part
    of the release target loop. So we spare us one loop.
  * Renamed varialbes: pkg -> package, rt_name -> release_target_name

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1397,12 +1397,10 @@ class Project < ApplicationRecord
     # One catch, currently there's only one patchinfo per incident, but things keep changing every
     # other day, so it never hurts to have a look into the future:
     package_count = 0
-    packages.where.not(id: global_patchinfo_package).select(:name, :id).each do |pkg|
+    packages.where.not(id: global_patchinfo_package).where("name like '%.%'").select(:name, :id).each do |pkg|
       # Current ui is only showing the first found package and a symbol for any additional package.
       break if package_count > 2
 
-      rt_name = pkg.name.split('.', 2).last
-      next unless rt_name
       # Here we try hard to find the release target our current package is build for:
       rt_name = guess_release_target_from_package(pkg, release_targets_ng)
 
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1381,6 +1381,7 @@ class Project < ApplicationRecord
     # First things first, get release targets as defined by the project, err.. incident. Later on we
     # magically find out which of the contained packages, err. updates are build against those release
     # targets.
+    repo_name_to_project_name = {}
     release_targets_ng = {}
     repositories.includes(:release_targets).each do |repo|
       repo.release_targets.each do |rt|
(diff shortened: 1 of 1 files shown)


bgeuken:
I didn't test the second commit yet. Hence the draft PR

src/api/app/models/project.rb
```diff
@@ -1391,25 +1392,22 @@ def release_targets_ng
...
-
-      rt_name = pkg.name.split('.', 2).last
-      next unless rt_name
-      # Here we try hard to find the release target our current package is build for:
-      rt_name = guess_release_target_from_package(pkg, release_targets_ng)
-
-      # Build-disabled packages can't be matched to release targets....
-      next unless rt_name
+      # Here we try to find the release target our current package is build for:
+      first_matching_build_target = package.flags.where(flag: :build, status: 'enable', repo: repo_name_to_project_name.keys).select(:repo).first
```
coolo: careful. not disabled doesn't mean enabled - enable is the default and doesn't have a flag entry then. But in case of incidents there seems to be explicit flags.

But https://build.opensuse.org/package/meta/openSUSE:Maintenance:9588/man-pages-posix.openSUSE_Leap_15.0_Update makes me wonder what significance the first repo has. The first seems to be _ports, but the release target 'from package' would be _Update.

coolo: You can read in every single comment just how frustrated sascha was about all this :)

coolo:
[random thought - pretty much out of context for this PR]

I'm afraid you can optimize this code as much as you want, it won't be enough unless this code is put at creation time and the results stored, so you can query them in one go.

We have 169 open incidents in opensuse and /projects/openSUSE:Maintenance/maintenance_incidents loads 'reasonably fast [for what it does]' (22 seconds)

But for SUSE we have (I'm not sure I'm allowed to say, but more :) and that page on IBS loads in 5.2 minutes and I'm sure you can halve that by doing smarter queries, but still noone will visit it.

Maintenance solved that by having their own DB with its own UI which is synced with OBS data (and do not use the OBS pages at all). As such I would dare to say: take the risk of introducing models that summarize existing data and are updated on changes (especially of build status).

Then you can also implement proper pagination as well, which is with the current code plain impossible as data is generated on the fly.

src/api/app/models/project.rb
bgeuken: > careful. not disabled doesn't mean enabled - enable is the default and doesn't have a flag entry then. But in case of incidents there seems to be explicit flags.

I found that strange as well. Though in the maintenance incident packages on OBS you can see that they explicitly enable the package flags.

coolo:
last week: maximum 2 minutes, this week: maximum 2.5 min. Exactly my point :)

bgeuken:
> /projects/openSUSE:Maintenance/maintenance_incidents loads 'reasonably fast [for what it does]' (22 seconds)

> this week: maximum 2.5 min. Exactly my point

Sure... exactly your point:wink:

bgeuken:
Closing in favor of #7529

PR #7406, opened 2019-04-12
Patchinfo refactor

saraycp:
Refactor Patchinfo controller and consequently model, views, javascript and specs.
    
- Moves patchinfo's XML handling to the model.
- Instantiates patchinfo object instead of using instance variables.
- Refactors update action mainly.
- Adapt webui views and javascript.
- Validates attributes in the model in the rails way.
- Simplifies part of the code.

   
### DIFF
--- a/src/api/app/models/patchinfo.rb
+++ b/src/api/app/models/patchinfo.rb
@@ -39,11 +39,15 @@ class Patchinfo
 
   attr_reader :document
   attr_writer :data
-  attr_accessor :summary, :description, :packager
+  attr_accessor :summary, :description, :packager, :category, :rating, :name,
+                :binaries, :version, :message, :relogin_needed,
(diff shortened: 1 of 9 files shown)


src/api/app/models/patchinfo.rb
```diff
@@ -257,4 +256,106 @@ def issues_by_tracker
...
+    self.zypp_restart_needed = !patchinfo_xml.value('zypp_restart_needed').nil?
+    if patchinfo_xml.value('stopped')
+      self.block = true
+      self.block_reason = patchinfo_xml.value('stopped')
+    end
+
+    self
+  end
+
+  def to_xml(project, package)
```
coolo: [nitpick] if the to_xml is in the model, you can (and possibly should) use CanRenderModel and builder file

PR #7408, opened 2019-04-15
Fix performance of group#show 

### LINKED ISSUE #7139: Groups#show page in Bootstrap is super slow
In the Bootstrap UI, the Groups#show page is super slow.

The maximum total time went from **43 ms** in the old UI to **24 seconds** in the Bootstrap UI. Everything is in our [Grafana instance](https://obs-measure.opensuse.org/d/qjkd7uBmz/performance-overview?refresh=5m&orgId=1). Search for `webui/groups/show` and you will see these numbers.

coolo:
The following call done in the group#show page is extremely slow (> 20 seconds):

```
BsRequest::FindFor::Group.new(params).all
```

because of the complex query that it generates. Refactor how the query is generated, writing in the Rail way improving the performance.

We now use the UNION feature of mysql, which allows to avoid the bottleneck of OR avoiding indices. The UNION joins 2 SELECT calls that are technically a OR of a join, but simplifies the way the query optimizer can use indices.

The same applies to FindFor::User, but this code is only used when calling the API with /request?view=collection&user=X.

benchmarks (on production clone of January, just counting the rails time without javascript):

wicked-maintainers:  7077ms -> 312ms
opensuse-review-team: 834ms -> 552ms
gnome-maintainers: 13885ms -> 1889ms

(gnome-maintainers is still a bit slow as it has a lot of reviews and the reviews trouble mariadb indexes, but we decided to stop here).

Fixes #7139

Co-authored-by: @Ana06
Co-authored-by: @andrii-suse

### DIFF
--- a/src/api/app/models/bs_request/find_for/group.rb
+++ b/src/api/app/models/bs_request/find_for/group.rb
@@ -8,14 +8,14 @@ class BsRequest
         group = ::Group.find_by_title!(group_title)
 
         # find requests where group is maintainer in target project
-        @relation, inner_or = extend_query_for_maintainer(group, @relation, roles, inner_or)
+        inner_or = extend_query_for_maintainer(group, roles, inner_or)
 
         if roles.empty? || roles.include?('reviewer')
           @relation = @relation.includes(:reviews).references(:reviews)
           # requests where the user is reviewer or own requests that are in review by someone else
           or_in_and = ["reviews.by_group=#{quote(group.title)}"]
 
-          @relation, inner_or = extend_query_for_involved_reviews(group, or_in_and, @relation, review_states, inner_or)
+          inner_or = extend_query_for_involved_reviews(group, or_in_and, review_states, inner_or)
         end
         if inner_or.empty?
           @relation.none
--- a/src/api/app/models/bs_request/find_for/group.rb
+++ b/src/api/app/models/bs_request/find_for/group.rb
@@ -4,24 +4,39 @@ class BsRequest
       include UserGroupMixin
 
       def all
-        inner_or = []
-        group = ::Group.find_by_title!(group_title)
+        query = union_query
+        @relation = if query.present?
+                      @relation.where("bs_requests.id IN (#{query})")
+                    else
+                      @relation.none
+                    end
+      end
 
-        # find requests where group is maintainer in target project
-        inner_or = extend_query_for_maintainer(group, roles, inner_or)
+      def all_count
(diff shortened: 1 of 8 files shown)


src/api/app/models/bs_request/find_for/group.rb
```diff
@@ -4,24 +4,39 @@ class Group < Base
...
-        @relation, inner_or = extend_query_for_maintainer(group, @relation, roles, inner_or)
+      def group
+        @group ||= ::Group.find_by_title!(group_title)
+      end
+
+      def union_query
+        query_parts = []
+        query_parts << bs_request_actions_query.to_sql if bs_request_actions_query
+        query_parts << reviews_query.to_sql if reviews_query
+        query_parts.join(' UNION ')
```
hennevogel: [active_record_union](https://github.com/brianhempel/active_record_union) looks pretty active to me.

Ana06: We checked it, but we thought it was not worthwhile to add a gem to use it just once. In addition, I am not sure if `union(nil)` works. Otherwise we will have to keep the `ifs` anyway. I can still give it a try.

Ana06:

src/api/app/models/bs_request/find_for/group.rb
Ana06: the gem is not working for this case: https://github.com/brianhempel/active_record_union/issues/23

Ana06:
I think that Hakiri warnings are all false positives. @openSUSE/open-build-service does anybody think differently? Please review :wink:

andrii-suse:
> I think that Hakiri warnings are all false positives.
They would be true positives if user input would go into them directly. As long as those subqueries (union_query, users_query and packages_query) are safe - result query will be safe as well, so it should be safe to ignore the warnings.

src/api/app/models/bs_request/find_for/group.rb
```diff
@@ -4,24 +4,39 @@ class Group < Base
...
+        query = union_query
+        @relation = if query.present?
+                      @relation.where("bs_requests.id IN (#{query})")
+                    else
+                      @relation.none
+                    end
+      end
+
+      def all_count
+        ActiveRecord::Base.connection.execute("SELECT COUNT(bs_request_id) FROM (#{union_query}) x").first.first
```
dmarcoux: What is the double `.first` for?

coolo: It returns a list of rows. We want the first element of the first row.

PR #7409, opened 2019-04-15
Correct wrong parameter in token api route documentation

krauselukas:
The documentation points to the usage of the userid in the api routes
for token creation, listing and deletion. But it should be the login (so the username).

Didn't checked the other routes. You can see the code here: https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/person/token_controller.rb#L39

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -154,13 +154,13 @@ Parameters:
   cmd: lock             # will lock the user and his home projects
   cmd: delete           # will mark the user as deleted and remove her home projects
 
-GET /person/<userid>/token
+GET /person/<login>/token
 
   Lists all authentication token for a user
 
 XmlResult: tokenlist
 
-POST /person/<userid>/token
+POST /person/<login>/token
 
   Create a new authentication token for this user. It may be limited for a specific package
   via optional project&package parameters.
@@ -171,7 +171,7 @@ POST /person/<userid>/token
 
 XmlResult: status
 
-DELETE /person/<userid>/token/<id>
+DELETE /person/<login>/token/<id>
 
   Delete a specific authentication token. The <id> is listed in GET call.
 


coolo:
`api/api.txt:  userid: user login name`

So using `<userid>` shouldn't be a problem. It's also consistent with `api/api.txt:GET /person/<userid>` - so downvote from me.

Ana06:
Line [121](https://github.com/openSUSE/open-build-service/pull/7409/files#diff-5d7068c489e190442136ca5db490770fL121) of `api/api.tx`t:

> <userid>: Id of user

I would say this is what needs to be updated.

krauselukas:
@coolo @Ana06 It was confusing to me, when i hear ```<userid>``` I'm thinking about the internal id, not the login alias username. In the routes.rb it is ```:login``` which is quite clear to me. Would be great to be consistent.
@ana06 changing the description in line 121 would already help, since it is the username not the id of the user :)

krauselukas:
@coolo @Ana06 I could replace ```<userid>``` with ```<username>``` in the whole api documentation. This would make it more obvious (at least for me). If there is no commitment regarding this, I would just change the description in line 121, and leave the rest as it is. :)

coolo:
then better use login - as this is already used in some parts. This is actually super inconsistent, so better clean this up on a mass scale :(

```
api/api/api.txt:  login: List available groups of this user.
api/api/api.txt:  userid: user login name, required for add_user and remove_user command
api/api/api.txt:  user: specify a user login name to list his packages
```

And it's all the same :(

krauselukas:
@coolo allright, sounds good

Ana06:
@krauselukas will you update the PR or should we close this?

krauselukas:
@ana06 sorry @Ana06 was busy with other things, i will close for now, and reopen as soon as i have time

PR #7416, opened 2019-04-16
Migrate patchinfo show view to Bootstrap

DavidKang:
We migrated the patchinfo show view to bootstrap.

- Delete patchinfo is a modal
- Blocked reason is shown as an alert message.
- Rating and category are shown as badges.

Screenshot_2019-04-22 Open Build Service(1)


src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
```diff
@@ -0,0 +1,32 @@
...
+      %i.fas.fa-info-circle.text-info
+      %span Submitted from
+      %span= user_with_realname_and_icon(packager)
+  - if patchinfo.version
+    %li
+      %i.fas.fa-info-circle.text-info
+      %span Version #{patchinfo.version}
+  - if patchinfo.relogin_needed
+    %li
+      %i.fas.fa-info-circle.text-warning
```
dmarcoux: When using `text-warning`, we should use the icon `fa-exclamation-triangle` to be consistent with the other views. See the [patterns](https://obs-patterns.netlify.com/atoms/icons.html) for details.

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
```diff
@@ -0,0 +1,32 @@
...
+      %span Relogin is needed
+  - if patchinfo.reboot_needed
+    %li
+      %i.fas.fa-info-circle.text-warning
```
dmarcoux: Same as above.

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
```diff
@@ -0,0 +1,32 @@
...
+      %span Reboot is needed
+  - if patchinfo.zypp_restart_needed
+    %li
+      %i.fas.fa-info-circle.text-warning
```
dmarcoux: Same as above.

src/api/app/views/webui2/webui/patchinfo/show.html.haml
```diff
@@ -0,0 +1,57 @@
...
+          = simple_format @patchinfo.message
+
+  .col-md-6
+    .card.mb-3
+      %h5.card-header Fixed bugs
+      .card-body
+        - if @patchinfo.issues.present?
+          - @patchinfo.issues.each do |issue|
+            .media
+              %i.pt-1.mr-1.fas.fa-bug.text-primary
```
dmarcoux: We need to add the `fa-bug` icon to the [patterns](https://obs-patterns.netlify.com/atoms/icons.html). Please send a PR for this :slightly_smiling_face:

src/api/app/helpers/webui/patchinfo_helper.rb
```diff
@@ -4,4 +4,27 @@ def patchinfo_bread_crumb(*args)
     args.insert(0, link_to(@package, package_show_path(project: @project, package: @package)))
     project_bread_crumb(*args)
   end
+
+  def patchinfo_header(patchinfo, package_names)
+    list = package_names.to_sentence
+    text = "#{patchinfo.category} update for #{truncate(list, length: 60)}"
```
dmarcoux: I believe the category should be displayed just like the rating is, so with a badge.

src/api/app/helpers/webui/patchinfo_helper.rb
```diff
@@ -4,4 +4,27 @@ def patchinfo_bread_crumb(*args)
...
+      header_subtitle(patchinfo.summary, text, list)
+    end
+  end
+
+  private
+
+  def header_title(patchinfo, text)
+    haml_tag(:h3, class: 'mb-0') do
+      haml_concat(patchinfo.summary.presence || text)
+      haml_tag(:span, 'blocked', class: 'badge badge-blocked') if patchinfo.block
```
dmarcoux: Adding a title like `Release of this patchinfo is blocked` would make it clearer what this `blocked` means.

src/api/app/helpers/webui/patchinfo_helper.rb
```diff
@@ -4,4 +4,27 @@ def patchinfo_bread_crumb(*args)
...
+      haml_tag(:span, patchinfo.rating, class: "badge badge-rating #{patchinfo.rating}")
```
dmarcoux: Adding a title like `Rating of this patchinfo` would make it obvious what this is about.

I wonder if rating could be renamed to Severity. But this is out of scope.

DavidKang:
@dmarcoux, I addressed your suggestions, could you review it again? :smile_cat:

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
```diff
@@ -0,0 +1,32 @@
+%ul.list-unstyled
+  - if patchinfo.block
```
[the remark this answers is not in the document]
dmarcoux: I agree. I would take this even further, we don't need a collapse for the block reason.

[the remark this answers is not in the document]
DavidKang: okey, I will remove the collapse.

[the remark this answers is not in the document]
DavidKang: @dmarcoux, you prefer to have the `blocked reason`  inside the `title` of the blocked badge? Maybe is too hidden

src/api/app/views/webui2/webui/patchinfo/show.html.haml
```diff
@@ -0,0 +1,57 @@
...
+  .col-md-6
+    .card.mb-3
+      %h5.card-header Fixed bugs
+      .card-body
+        - if @patchinfo.issues.present?
+          - @patchinfo.issues.each do |issue|
+            .media
+              %i.pt-1.mr-1.fas.fa-bug.text-danger
+              .media-body
+                - if issue[0] == 'CVE'
```
Ana06: [nitpick] you could write this as:
```ruby
link_text = issue[0] == 'CVE' ? issue[1] : "#{issue[0]}##{issue[1]}"
= link_to(link_text, issue[2], target: :_blank, rel: 'noopener', class: 'small')
```

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
```diff
@@ -0,0 +1,32 @@
...
+        %i.fas.fa-info-circle.text-info
+        %span Blocked Reason
+        %i.fas.fa-chevron-left.expander{ title: 'Show blocked reason' }
+        %i.fas.fa-chevron-down.collapser{ title: 'Hide blocked reason' }
+      .text-muted.collapse#blocked-reason-details
+        = patchinfo.block_reason.presence || 'No reason entered.'
+  - if packager
+    %li
+      %i.fas.fa-info-circle.text-info
+      %span Submitted from
```
dmarcoux: Submitted **by**

src/api/app/views/webui2/webui/patchinfo/show.html.haml
vpereira: > ```ruby
> link_text = issue[0] == 'CVE' ? issue[1] : "#{issue[0]}##{issue[1]}"
> ```

and have it in a helper

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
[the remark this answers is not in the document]
dmarcoux: The reason why a patchinfo update is blocked is one of the most important information. It should be prominent.

This isn't perfect, but here's a proposal. The spacing and positioning could be improved, but I think it would be already a step in the right direction:

proposal

This is the HTML:

```
<div class="mt-1">
  <span class="badge badge-blocked" title="Release of this patchinfo is blocked">blocked</span>
  <span class="badge badge-category security" title="Category of this patchinfo">security</span>
  <span class="badge badge-rating moderate" title="Rating of this patchinfo">moderate</span>
  
  <p class="text-danger m-0">This is causing issues! Wait!!! </p>
</div>
```

I'm not convinced though, because the top box with the h3, the badges, side information, bottom actions, summary and description is getting convoluted.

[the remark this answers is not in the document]
DavidKang: and this option?
Screenshot_2019-04-17 Open Build Service(2)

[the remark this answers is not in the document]
vpereira: hm I prefer the version from @dmarcoux , maybe somebody from maintenance could giv us some feedback here?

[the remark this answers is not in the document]
dmarcoux: Another proposal, using an alert message (which isn't dismissable)

another-proposal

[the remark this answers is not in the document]
Ana06: I like the alert message, it is consistent with the rest of the app.

src/api/app/views/webui2/webui/patchinfo/show.html.haml
```diff
@@ -0,0 +1,57 @@
...
+        %h5.card-header Message
+        .card-body
+          = simple_format @patchinfo.message
+
+  .col-md-6
+    .card.mb-3
+      %h5.card-header Fixed bugs
+      .card-body
+        - if @patchinfo.issues.present?
+          - @patchinfo.issues.each do |issue|
```
vpereira: Please add the issue "summary" here. It helps the Security Maintenance team to track, at one glance, which issues were fixed on that incident

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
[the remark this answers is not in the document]
vpereira: I think the reason in the flash message is bad, because the reason can become really lengthy and verbose

msmeissn:
looks good to me!

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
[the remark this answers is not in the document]
dmarcoux: If the reason is longer, it would still fit in the alert, so it wouldn't be an issue.

dmarcoux:
Great work @DavidKang!

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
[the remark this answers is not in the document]
vpereira: talking with our maintenance team, they prefer the first option presented by @dmarcoux , @hennevogel . Not sure if a multiline flash message is the right way to do it.

src/api/app/views/webui2/webui/patchinfo/show.html.haml
```diff
@@ -0,0 +1,60 @@
...
+
+  .col-md-6
+    .card.mb-3
+      %h5.card-header Fixed bugs
+      .card-body
+        - @patchinfo.issues.each do |issue|
+          .media
+            %i.pt-1.mr-1.fas.fa-bug.text-danger
+            .media-body
+              - link_text = issue[0] == 'CVE' ? issue[1] : "#{issue[0]}##{issue[1]}"
```
vpereira: could you please move it to a helper? I'm quite sure the security team prefer to read here ```CVE-2012-1234``` then ```cve#2012-1234```

DavidKang: done!

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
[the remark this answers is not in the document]
dmarcoux: Even if the maintenance team prefers my first proposal, I would still favor my second proposal since it's consistent with the rest of the Bootstrap UI (so I agree with @Ana06 here). On top of the consistency, the second proposal also clearly displays the block reason and makes the top box (with the h3, summary, etc..) more readable since there's isn't as much information in it.

Alerts with multiline content are fine, see the examples from [Bootstrap](https://getbootstrap.com/docs/4.3/components/alerts/#additional-content).

src/api/app/views/webui2/webui/patchinfo/show.html.haml
DavidKang: Screenshot_2019-04-17 Open Build Service(6)

src/api/app/views/webui2/webui/patchinfo/show.html.haml
DavidKang: I assume you are talking about this:
Screenshot_2019-04-17 Open Build Service(10)

This info is in `issue[3]`

src/api/app/assets/stylesheets/webui2/patchinfo.scss
```diff
@@ -0,0 +1,35 @@
+.patchinfo {
+  .badge{
```
bgeuken: There should be a whitespace:wink:

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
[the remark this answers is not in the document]
vpereira: @dmarcoux sorry, but i failed to find where we consistently show a  desired state set by the user as a sticky error flash message? I was looking in other sections we either have a legend, or a list with error icons or a table showing "broken".. could you point me where we have something similar with the "use setting the patchinfo to stop" and  "show a sticky error message" (possible multiline including references to external bugs) ?

src/api/app/views/webui2/webui/patchinfo/show.html.haml
vpereira: yes great!

src/api/app/assets/stylesheets/webui2/patchinfo.scss
DavidKang: Thanks! :smile_cat:

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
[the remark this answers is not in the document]
coolo: I think the basic misunderstanding is that @dmarcoux expected the block reason to be a problem with the patch. But in reality it's always just a halting state.

The patch itself is fine and tested, but the team waits a little longer before releasing to make it worth the extra overhead.

So an error flash is wrong, but I still like the idea of using a flash for this - as it's a rather important info.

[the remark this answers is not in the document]
coolo: (And IMO this is a pure misuse of this field, it should be a review on the release request, but then you have the problem that they appear as Tasks on your page as OBS doesn't allow to ignore reviews as tasks)

msmeissn:
We use the <stopped> tag to declare problems with the patch / patchinfo, either wiating for additional submisisons for feedback.
It is used when this patch is broken in some form usually, not for just holding it back.

coolo:
A lot of the stopped reasons are 'VUL-1 only', I wouldn't say that's a "problem with the patch". And even waiting for additional submissions is not an error. It's blocking the workflow, but it's not an error. As such it's not to be displayed as an error to the user, because the user didn't do anything wrong.

bgeuken:
> coolo commented 7 minutes ago
>
> A lot of the stopped reasons are 'VUL-1 only', I wouldn't  say that's a "problem with the patch". And even waiting for additional  submissions is not an error. It's blocking the workflow, but it's not an  error. As such it's not to be displayed as an error to the user,  because the user didn't do anything wrong.

+1

Another issue I see is that, so far, this area was reserved for notification messages and the area below (the tab) was used for the data related to the viewed object (package, project, patchinfo, ...). With this change we would start to mix them.

Just imagine a patchinfo in `blocked` state that has an error message after saving. A user would now see two read text boxes. It might even be that he / she doesn't notice this, because there was something red before.
And if it's noticed one would still have to mentally parse which of the two boxes is the actual error.
Not saying that this is a hard task for a user, but this is where a good design can help users to accomplish their job without having to think about such things ("Is this an error or is it in blocked state?").

vpereira:
There are some situation where I see the ```stopped``` flag being used:

a) Maintenance Team wants to prepare the patchinfo but the patches aren't ready
b) To bundle some less important issues to a package together and do a single release
c) Some dependency is missing, the incident is ready to be released and there are some coordination happening behind the scenes

in some extension, it is similar with a "review by project" done on Staging Workflow. Quite often, the idea is just to block the incident. So it is a "user defined state", not an error and therefore shouldn't be handled as such.

coolo:
I would rather compare it with excluded requests :)

vpereira:
> I would rather compare it with excluded requests :)

It's fair, however IMO with excluded requests, you want to ignore a request, while with the ```stopped``` flag you want to avoid an Incident being accidentally released.  Either way, none of them have a sticky error flash message.

src/api/app/views/webui2/webui/patchinfo/_side_elements.html.haml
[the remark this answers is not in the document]
dmarcoux: If we don't use an alert for this, it's just going to be hidden among the text of the top box. Since it's a state and it is important, I would prefer to use the alert. If you think it shouldn't be an error, than we can change it to a warning or even info.

Regarding other cases, I was thinking about the build results where we also use it to display the state in a sticky alert when it's building. We do it because it's an important information and shouldn't be overlooked by users.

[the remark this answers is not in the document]
bgeuken: > If we don't use an alert for this, it's just going to be hidden among the text of the top box.

When I checked this view in the reviewapp I first didn't recognize the flash error. It took me a bit to realize that this is probably because my brain is trained that this area is only important when I just fired an action / saved some data.
The second reason might be that it's at the edge of the viewport (I also scrolled down a little bit).
Might be that users would adjust to this, but I still think using a flash like appearance has to many downsides.

I think the core problem is that we want to give this some emphasize, but because we use many icons and colors it becomes tricky to do this.

vpereira:
> I propose to merge this now and continue getting feedback and discussing the design afterwards.  What do the rest think?

Since today we had a long discussion, where almost everyone is against using flash error to show this information, I would wait until @DavidKang is back. Maybe we should summarize the issue and discuss about the design proposals

DavidKang:
If the problem is the colour of the alert we can change it to be like in the request with the superseed info:
Screenshot_2019-04-18 Open Build Service


Another proposal is going back and having the block reason without the collapse.
Screenshot_2019-04-18 Open Build Service(1)

vpereira:
@DavidKang thank you for checking it today. I prefer going back and having the block reason without the collapse, keeping the consistency  as we do with other blocked workflows

dmarcoux:
> @DavidKang thank you for checking it today. I prefer going back and having the block reason without the collapse, keeping the consistency as we do with other blocked workflows

Let's do this, then merge this PR and iterate if needed.

bgeuken:
> If the problem is the colour of the alert we can change it to be like in the request with the superseed info:

The color was not the issue, but that we would have shown patchinfo related data in the area that is used for notifications.

Issue #7319, opened 2019-04-02
package#dependency view is not migrated to Bootstrap


dmarcoux:
# Issue Description
The package#dependency view is not migrated to Bootstrap

Expected Result
============
I expect the package#dependency view to be migrated to Bootstrap

How to Reproduce
===============
1. After joining the beta program, go to the package#binary view (one [example]( https://build.opensuse.org/package/binary/openSUSE:Factory:Staging:H/000release-packages:openSUSE-MicroOS-release/standard/x86_64/openSUSE-MicroOS-release-1-8.1.x86_64.rpm))
2. Click on one of the links under the `Required by` or `Provided by` columns
3. See the package#dependency view from the old UI

Further Information
===============
Reported by @coolo on IRC.

saraycp:
![](https://github.trello.services/images/mini-trello-icon.png) [[Last card] Address feedback/issues](https://trello.com/c/9QtsFies/1097-last-card-address-feedback-issues)

saraycp:
@coolo, @dmarcoux  PR  #7417 is my proposal, please have a look.

PR #7417, opened 2019-04-16
Migrate Package#dependency view to Bootstrap

saraycp:
Inside the details of a binary we find some links in the 'Required by' or 'Provided by' columns. The views to which those links point weren't migrated to Bootstrap.

The migrated view had a lot in common with the binary view, so part of the code has been moved to a partial. It also includes some refactoring in HTML and CSS.

Fixes: #7319

**Before:**

Detailed Informations About openSUSE MicroOS release dvd 1 7 2 i586 rpm   openSUSE Build Service

**After:**

Dependency information about openSUSE MicroOS release dvd 1 9 1 i586 rpm   Open Build Service(2)

### DIFF
--- a/src/api/app/views/webui2/webui/package/_fileinfo.html.haml
+++ b/src/api/app/views/webui2/webui/package/_fileinfo.html.haml
@@ -1,0 +1,37 @@
+%p
+  %strong Title:
+  = fileinfo.value(:summary)
+%p
+  %strong Description:
+  %br
+  - description = fileinfo.value(:description)
+  - if description
+    - description.split(/\n/).each do |line|
+      = line
+      %br
+  - else
+    %i No description set
+%p
+  %strong Version:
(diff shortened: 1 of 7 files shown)


saraycp:

src/api/app/views/webui2/webui/package/_fileinfo.html.haml
```diff
@@ -0,0 +1,38 @@
+%dl.row.mt-4
+  %dt.col-md-2.text-truncate Title
+  %dd.col-md-10= fileinfo.value(:summary)
```
coolo: jfyi: value() is converting the argument .to_s - so using strings is actually a little faster. But it really doesn't matter in views.

saraycp: @coolo, thanks! I've just opened a new PR for this: #7420

Issue #7257, opened 2019-03-21
[beta] webui: extend diff background marking to the right


jengelh:
Concerns the diff view as present on e.g. https://build.opensuse.org/request/show/687180

Current state:
![bk](https://user-images.githubusercontent.com/8861948/54755297-fdd99580-4be5-11e9-8148-38015ec1d857.png)

Is it possible to extend the pastel red/green background of diff lines towards the right edge, so that it does not look so staggered? E.g.
![bk2](https://user-images.githubusercontent.com/8861948/54755465-47c27b80-4be6-11e9-8b99-7d3171cc75e6.png)

jengelh:
It just occurred to me that the current practice actually makes useless trailing whitespace visible, so I'd leave it at that, unless the differ starts highlighting such totally differently (which is probably too much work).

Ana06:
@jengelh 

> It just occurred to me that the current practice actually makes useless trailing whitespace visible, so I'd leave it at that, unless the differ starts highlighting such totally differently (which is probably too much work).

I am not sure if I understand. Would you like to leave it as it is currently (in that case we should close the issue) or do you want to change it to highline the whole line and in that case what role play the spaces?

jengelh:
Now that I have had 27 days to work with and ponder about it, I find that having a switch to be able to use ways would be really cool.

hennevogel:
> I find that having a switch to be able to use both ways would be really cool.

No sorry we are not going to introduce settings for this. Let's close this.

PR #7422, opened 2019-04-17
Improve diff view

coolo:
Related to #7257

Bento:

bento

Before:

before

After:

after

### DIFF
--- a/src/api/app/assets/javascripts/webui2/cm2/use-codemirror.js
+++ b/src/api/app/assets/javascripts/webui2/cm2/use-codemirror.js
@@ -9,6 +9,12 @@ window.onbeforeunload = function() {
   }
 }
 
+function cm_mark_diff_lines() {
+  // as we can't use css to mark parents we use javascript to propagate diff lines
+  $('.cm-positive').parents('.CodeMirror-line').addClass('CodeMirror-positive-line');
+  $('.cm-negative').parents('.CodeMirror-line').addClass('CodeMirror-negative-line');
+}
+
 function use_codemirror(id, read_only, mode) {
   var codeMirrorOptions = {
     lineNumbers: true,
@@ -30,6 +36,10 @@ function use_codemirror(id, read_only, mode) {
   var textarea = $('#editor_' + id);
   var editor = CodeMirror.fromTextArea(document.getElementById("editor_" + id), codeMirrorOptions);
   editor.id = id;
+
+  cm_mark_diff_lines();
+  editor.on('scroll', cm_mark_diff_lines);
+
   if (!read_only) {
     editor.setSelections(editor);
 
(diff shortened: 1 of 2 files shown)


src/api/app/assets/javascripts/webui2/cm2/use-codemirror.js
```diff
@@ -9,6 +9,12 @@ window.onbeforeunload = function() {
   }
 }
 
+function cm_mark_diff_lines(cm) {
```
bgeuken: The parameter seems unused

coolo: It's passed in the event case and in line 70 and 79 are other examples of callbacks bearing unused parameters. But as I call the function also without event, I guess javascript doesn't care :)

src/api/app/assets/stylesheets/webui2/cm2/suse.scss
```diff
@@ -1,61 +1,69 @@
...
+  .CodeMirror-positive-line {
+    background: rgba($success, 0.1);
+  }
+
+  .CodeMirror-gutters {
+    // Prevent gutters from being opatique by adding background to foreground colors to create non-opatique version of cap color
+    background: opacify( mix( $card-bg, opacify( $card-cap-bg, 1 ), ( 1 - alpha( $card-cap-bg ) ) * 100% ), 1 );
+    border-right: $card-border-width solid $card-border-color;
+  }
+  .CodeMirror-guttermarker { color: $text-muted; }
```
dmarcoux: This can be grouped with the other color related rules above (all the `.cm-*`)

src/api/app/assets/stylesheets/webui2/cm2/suse.scss
```diff
@@ -1,61 +1,69 @@
...
-.cm-s-bootstrap .cm-hr {color: $secondary;}
-.cm-s-bootstrap .cm-link {color: $link-color;}
-
-.cm-s-bootstrap .cm-error {color: $danger;}
-
-.cm-s-bootstrap .cm-negative {
-  color: $danger;
-  background: rgba($danger, 0.1);
-}
+  .cm-header {color: $blue;}
```
dmarcoux: Some of these can be grouped and with a line to space it is out, it will be more readable.

src/api/app/assets/javascripts/webui2/cm2/use-codemirror.js
bgeuken: I would drop them there as well. But that's a different story:-)

> I guess javascript doesn't care

Yes, js is flexible in that regard. But why defining variables or parameters we never use:wink:

src/api/app/assets/stylesheets/webui2/cm2/suse.scss
coolo: but you realize I just removed the duplication? And I disagree with you - they are grouped by mode specifics and editor specifcs.

dmarcoux: Yes, I've seen that you use Sass like it should (before it was just CSS with some Sass variables).

Good point regarding the modes.

PR #7423, opened 2019-04-17
Catch unhandled backend errors and show a proper error message in the UI

bgeuken:
This will catch Backend::Error and Timeout::Error exeptions and
redirect users the previous page. The execption will be send to errbit.

Many of the webui controllers call store without worrying about
possible exceptions that could be raised. This change makes sure
that we have a generic solution for this.

### DIFF
--- a/src/api/app/controllers/webui/download_on_demand_controller.rb
+++ b/src/api/app/controllers/webui/download_on_demand_controller.rb
@@ -14,7 +14,7 @@ class Webui::DownloadOnDemandController < Webui::WebuiController
         @download_on_demand.save!
         @project.store
       end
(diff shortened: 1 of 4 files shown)


dmarcoux:
One typo in the commit message: `exception` instead of `execption`

vpereira:
unrelated with your PR but  looking the change, something jumped out of my monitor:

```
# FIXME: This is more than stupid. Why do we tell the user that something isn't found
  # just because there is some data missing to compute the request? Someone needs to read
  # http://guides.rubyonrails.org/active_record_validations.html
```

bgeuken:
I've updated the cases were we catched Backend::Error and redirect to root path. The more specific cases I left as it was.

PR #7425, opened 2019-04-17
WIP: Add service files through the api via token authentication

krauselukas:
Still need to check what is already covered by tests and where i need to add ones.


src/api/config/routes.rb
```diff
@@ -572,9 +572,12 @@ def self.public_or_about_path?(request)
     resources :architectures, only: [:index, :show, :update] # create,delete currently disabled
 
     ### /trigger
-    post 'trigger/runservice' => 'trigger#runservice'
+    post 'trigger/runservice' => 'services/trigger#runservice'
     post 'trigger/webhook' => 'services/webhooks#create'
 
+    ### /services
+    put 'services/add_service_file' => 'services/base#add_service_file'
```
bgeuken: We use RESTful routes:-) That means that the action, like 'add', is defined via the HTTP Verb, here PUT. So it's not necessary to have it as part of the URL.
Usually we would write this like `put 'services/service_file' => 'services/service_file#create'` or even drop the `_service` part -> `services/file`. That would mean that there would be a Service::ServiceFileController Service::FileController with a create action.

Some of our [webui controllers](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/projects/maintenance_incidents_controller.rb) be a good example for this.

krauselukas: @bgeuken makes sense, I will change it accordingly :)

coolo:
I'm not so sure I like it. The token is meant to be of security 2nd class - i.e. if it's lost, all the gaining user can do is trigger predefined services.

If you can use it also to create services, you can basically publish all kind of your software using my gpg key.

hennevogel:
>  The token is meant to be of security 2nd class

Well yes, it's meant to be authentication for one thing. Creating `_service` is broader than triggering but it's still only one thing basically. The alternative is to give out your password. What would you propose?

coolo:
I find it especially worrysome that it's possible to have global tokens basically replacing passwords to create _service files. This is only done during some workflow, right? So I would expect some one time token to be used for this.

hennevogel:
> So I would expect some one time token to be used for this.

One time is "easy", right @krauselukas?

krauselukas:
@coolo @hennevogel definitely possible and for the current needs sufficient. if both of you are happy with this approach i would go this way.

krauselukas:
I will close the PR and reopen as soon as I'm able to work on it :)

PR #7426, opened 2019-04-17
[WIP] Allow user to watch different types of items

bronzdoc:
Hi 👋 

Right now a user can only have a watchlist of projects...

This PR will add functionality that will allow a user to watch any "item" i.e projects, packages, requests, etc...

This is a WIP PR, still need to work on routes/controllers, refactor and specs, i will update accordingly.

### DIFF
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -411,22 +411,6 @@ class Webui::ProjectController < Webui::WebuiController
     @repoarray = repohash.sort
   end
 
-  def toggle_watch
-    if User.current.watches?(@project.name)
-      logger.debug "Remove #{@project} from watchlist for #{User.current}"
-      User.current.remove_watched_project(@project.name)
-    else
-      logger.debug "Add #{@project} to watchlist for #{User.current}"
-      User.current.add_watched_project(@project.name)
-    end
-
-    if request.env['HTTP_REFERER']
-      redirect_back(fallback_location: root_path)
-    else
-      redirect_to action: :show, project: @project
-    end
-  end
-
   def clear_failed_comment
     # FIXME: This should authorize destroy for all the attributes
     authorize @project, :update?
--- a/src/api/app/controllers/webui/watchlist_controller.rb
+++ b/src/api/app/controllers/webui/watchlist_controller.rb
@@ -1,0 +1,25 @@
+class Webui::WatchlistController < Webui::WebuiController
+   before_action :set_project, only: [:toggle_watch]
+
+  def toggle_watch
+    User.current.watches?(@project.name) ? unwatch_item : watch_item
+
+    if request.env['HTTP_REFERER']
+      redirect_back(fallback_location: root_path)
+    else
+      redirect_to action: project_show_path, project: @project
+    end
+  end
+
+  private
+
+  def watch_item
+    logger.debug "Add #{@project} to watchlist for #{User.current}"
+    User.current.add_watched_project(@project.name)
+  end
+
+  def unwatch_item
+    logger.debug "Remove #{@project} from watchlist for #{User.current}"
+    User.current.remove_watched_project(@project.name)
+  end
+end
--- a/src/api/app/models/concerns/watchable.rb
+++ b/src/api/app/models/concerns/watchable.rb
@@ -1,0 +1,8 @@
+module Watchable
+  extend ActiveSupport::Concern
+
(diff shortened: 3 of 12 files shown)


src/api/app/controllers/webui/watchlist_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+   before_action :set_project, only: [:toggle_watch]
+
+  def toggle_watch
+    User.current.watches?(@project.name) ? unwatch_item : watch_item
+
+    if request.env['HTTP_REFERER']
+      redirect_back(fallback_location: root_path)
+    else
+      redirect_to action: project_show_path, project: @project
+    end
```
hennevogel: I know you only move this but I don't think this makes sense. `redirect_back` already takes the `Referer` header into account.

src/api/app/models/concerns/watchable.rb
```diff
@@ -0,0 +1,8 @@
+module Watchable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :watched_items, as: :watchable, dependent: :destroy
+    has_many :users, through: :watched_items
```
hennevogel: `:users` is pretty generic, can you think of a better name?

src/api/db/migrate/20190415010809_create_watched_items.rb
```diff
@@ -1,6 +1,7 @@
 class CreateWatchedItems < ActiveRecord::Migration[5.2]
   def change
     create_table :watched_items, id: :integer do |t|
+      t.belongs_to :user
```
bgeuken: OBS is using integer for ids. That's because it was already around when Rails switched to bigint as default.
So will have to add `type: :integer` here.

src/api/app/models/user.rb
```diff
@@ -30,8 +30,8 @@ class User < ApplicationRecord
   # password yet. this is for backwards compatibility
   has_secure_password validations: false
 
-  has_many :watched_projects, dependent: :destroy, inverse_of: :user
-  has_many :watched_items, as: :watchable
+  has_many :watched_items
+  has_many :watched_projects, through: :watched_items, source: :watchable, source_type: 'Project'
```
bgeuken: Nice:thumbsup:

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -411,22 +411,6 @@ def monitor
...
-      User.current.add_watched_project(@project.name)
-    end
-
-    if request.env['HTTP_REFERER']
-      redirect_back(fallback_location: root_path)
-    else
-      redirect_to action: :show, project: @project
-    end
-  end
-
```
bgeuken: Please squash this commit and the following together. Since you move, and change, this code, it makes sense to keep this in one commit.

src/api/app/models/concerns/watchable.rb
bronzdoc: I made it like that since i'm making it just for users right now, so that's a reference to the users table, do we have other entities at the moment that should watch "watched_projects"? if so, can you provide an example so i have a bit more context? thanks! :)

src/api/db/migrate/20190415010809_create_watched_items.rb
bronzdoc: got it 👍

src/api/app/controllers/webui/project_controller.rb
bronzdoc: For sure! 🙂 
I'll squash the commits when i'm done working on this feature

bronzdoc:
Thanks for the review @hennevogel @bgeuken! will work on the changes you suggested and work in this feature in the upcoming days 🙂

src/api/app/models/concerns/watchable.rb
hennevogel: `Project`s for instance have people *owning* them through (I'm a bit embarrassed to say) a model called  `Relationship` (yes naming is a strong-suit of us ;-). I fear that people will misunderstand this. How about `has_many :watchers, through: :watched_items, class_name: 'User'`?

hennevogel:
@bronzdoc any update on this?

bronzdoc:
Hi @hennevogel 

No updates yet, have been a little busy. I hope to finish it soon...

hennevogel:
Okay let's close this. Please reopen if you would like to finish this. Thank you :bouquet:

PR #7427, opened 2019-04-18
Store the pagination count of DataTables that we use in bootstrap

bgeuken:
The pagination count will be store in local storage for 2 days.

### DIFF
--- a/src/api/app/assets/javascripts/webui2/datatables.js
+++ b/src/api/app/assets/javascripts/webui2/datatables.js
@@ -5,27 +5,25 @@
 //= require datatables/extensions/FixedColumns/fixedColumns.bootstrap4
 //= require datatables/extensions/FixedColumns/dataTables.fixedColumns
 
+var DEFAULT_DT_PARAMS = {
+  language: { search: '', searchPlaceholder: "Search..." },
+  stateSave: true,
+  stateDuration: 172800 // 2 days
+};
+
 function initializeDataTable(cssSelector, params){ // jshint ignore:line
-  var defaultParams = {
-    language: { search: '', searchPlaceholder: "Search..." },
-    stateSave: true,
-    stateDuration: 172800 // 2 days
-  };
-  var newParams = $.extend(defaultParams, params);
+  var newParams = $.extend({}, DEFAULT_DT_PARAMS, params);
   $(cssSelector).DataTable(newParams);
 }
 
 function initializeRemoteDatatable(cssSelector, params) { // jshint ignore:line
-  var defaultParams = {
-    language: { search: '', searchPlaceholder: "Search..." },
+  var defaultRemoteParams = {
     processing: true,
     serverSide: true,
     ajax: $(cssSelector).data('source'),
-    pagingType: 'full_numbers',
-    stateSave: true,
-    stateDuration: 172800 // 2 days
+    pagingType: 'full_numbers'
   };
-  var newParams = $.extend(defaultParams, params);
+  var newParams = $.extend(defaultRemoteParams, DEFAULT_DT_PARAMS, params);
 
   $(cssSelector).dataTable(newParams);
 }
--- a/src/api/app/assets/javascripts/webui2/datatables.js
+++ b/src/api/app/assets/javascripts/webui2/datatables.js
@@ -8,6 +8,8 @@
 function initializeDataTable(cssSelector, params){ // jshint ignore:line
   var defaultParams = {
     language: { search: '', searchPlaceholder: "Search..." },
+    stateSave: true,
+    stateDuration: 172800 // 2 days
   };
   var newParams = $.extend(defaultParams, params);
   $(cssSelector).DataTable(newParams);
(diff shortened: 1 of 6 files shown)


coolo:
I always found it quite useful if it remembered my filterings - and it was always quite obvious that it filtered as the filter string is set.

dmarcoux:
[Maintenance incidents](https://trello.com/c/4hbIZuK9/1095-maintenance-incidents)

src/api/app/assets/javascripts/webui2/datatables.js
```diff
@@ -19,7 +19,9 @@ function initializeRemoteDatatable(cssSelector, params) { // jshint ignore:line
     processing: true,
     serverSide: true,
     ajax: $(cssSelector).data('source'),
-    pagingType: 'full_numbers'
+    pagingType: 'full_numbers',
+    stateSave: true,
```
dmarcoux: Should we also do this in the function `initializeDataTable`, so just above this one?

bgeuken:
> stateSave stores not only the pagination count, but also the pagination position, filtering and sorting. If I made a search and came back to the same page in 2 days (when I had probably forgotten that I searched something 2 days ago), I would find it really unexpected that there are things filtered or that I am not in the first page.

I found that a nice addition, except the stored pagination position. That one is a bit confusing.

> What about keeping stateDuration default (1h 30 minutes) instead?

I found the default to short, e.g. it might happen that someone reviews the tasks in the morning and sets the page counter and when reviewing them again in the afternoon he / she doesn't want to redo this.
Though I don't have any strong opinion for one or the other.

src/api/app/assets/javascripts/webui2/datatables.js
bgeuken: True, if we use this we should do it on every DataTable

src/api/app/assets/javascripts/webui2/datatables.js
```diff
@@ -8,6 +8,8 @@
 function initializeDataTable(cssSelector, params){ // jshint ignore:line
   var defaultParams = {
```
Ana06: I think we should add default params to a variable that both methods share, as initializeTomoteDatatable is just extending them. This way we avoid repeating code and forgetting it in one of them (as has happened here) :wink:

bgeuken:
btw, I am going to update the pattern library once this is merged

src/api/app/assets/javascripts/webui2/datatables.js
bgeuken: @Ana06 Changed it:smiley:

src/api/app/assets/javascripts/webui2/datatables.js
```diff
@@ -5,22 +5,25 @@
...
 function initializeRemoteDatatable(cssSelector, params) { // jshint ignore:line
-  var defaultParams = {
-    language: { search: '', searchPlaceholder: "Search..." },
+  var defaultRemoteParams = {
     processing: true,
     serverSide: true,
     ajax: $(cssSelector).data('source'),
     pagingType: 'full_numbers'
   };
+  var defaultParams = $.extend(DEFAULT_DT_PARAMS, defaultRemoteParams);
```
Ana06: you can extend everything at once:

```JavaScript
var defaultParams = $.extend({}, DEFAULT_DT_PARAMS, defaultRemoteParams, params);
```

src/api/app/assets/javascripts/webui2/datatables.js
```diff
@@ -5,22 +5,25 @@
...
+  stateSave: true,
+  stateDuration: 172800 // 2 days
+};
+
 function initializeDataTable(cssSelector, params){ // jshint ignore:line
-  var defaultParams = {
-    language: { search: '', searchPlaceholder: "Search..." },
-  };
-  var newParams = $.extend(defaultParams, params);
+  var newParams = $.extend(DEFAULT_DT_PARAMS, params);
```
Ana06: The first parameter is modified (and returned) in `extend,` which means that `DEFAULT_DT_PARAMS` is modified. I guess that this is only a problem if we have more than one datatable per page.

To avoid problems I would write it:
```Javascript
var newParams = $.extend({}, DEFAULT_DT_PARAMS, params);
```

src/api/app/assets/javascripts/webui2/datatables.js
bgeuken: oh, nice:thumbsup:

src/api/app/assets/javascripts/webui2/datatables.js
```diff
@@ -5,23 +5,25 @@
...
-  var defaultParams = {
-    language: { search: '', searchPlaceholder: "Search..." },
+  var defaultRemoteParams = {
     processing: true,
     serverSide: true,
     ajax: $(cssSelector).data('source'),
     pagingType: 'full_numbers'
   };
-  var newParams = $.extend(defaultParams, params);
+  var newParams = $.extend(defaultRemoteParams, DEFAULT_DT_PARAMS, params);
```
Ana06: [nitpick] you don't need to assign this to a variable. You could use the first param directly.
[nitpick] I would use the following order as `defaultRemoteParams` could overwrite the default paramethers in `DEFAULT_DT_PARAMS`: `extend({}, DEFAULT_DT_PARAMS, defaultRemoteParams, params)`

coolo:
It can lead to strange behaviour - I left the page this morning when I had > 100 requests and now that I'm down to 40, it says there is no data. Took me a moment to realize I'm on page 5 of 4.

screencapture-build-opensuse-org-user-tasks-2019-05-05-18_49_28

PR #7430, opened 2019-04-18
WIP: Extract XML serializer and include it to the Comment model

vpereira:
It's a PoC of a general idea to extract non-core code to external modules and via composition include it back. I decided to do it with the serializers since it is already a common practice. Please do not merge

### DIFF
--- a/src/api/app/models/comment.rb
+++ b/src/api/app/models/comment.rb
@@ -18,7 +18,10 @@ class Comment < ApplicationRecord
   has_many :children, dependent: :destroy, class_name: 'Comment', foreign_key: 'parent_id'
 
   extend ActsAsTree::TreeWalker
+
   acts_as_tree order: 'created_at'
+
+  include Serializers::CommentSerializer
 
   scope :without_parent, -> { where(parent_id: nil) }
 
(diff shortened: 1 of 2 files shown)


vpereira:
> Not that I have anything against this practice but this piece of code in particular is a xml view ;-)

Hey, I don't know what you mean, there is no such thing as view in the model. What we do with a model is to serialize it via the method ```to_xml``` to be consumed in a view. Same approach is used by the [rails xml serializer](https://github.com/rails/activemodel-serializers-xml) which as a second step, we could look into and see if it fits our needs.

hennevogel:
> Hey, I don't know what you mean, there is no such thing as view in the model. 

We even pass `builder` into the model currently. There really is not need to do anything different than for any other view format. Gather what you need in the controller and write an XML view. No need for `.to_xml` or do we have `.to_html`?

vpereira:
> We even pass `builder` into the model currently. There really is not need to do anything different than for any other view format. Gather what you need in the controller and write an XML view. No need for `.to_xml` or do we have `.to_html`?

but AFAICS, in the https://github.com/openSUSE/open-build-service/blob/master/src/api/app/views/comments/index.xml.builder#L4 we are doing exactly that (calling ```to_xml```), no?

Issue #7216, opened 2019-03-18
Add a white background to flash messages when scrolling


mschnitzer:
**Is your feature request related to a problem? Please describe.**
I just discovered that the flash messages now follow the screen, when I scroll down, but I think it can be improved.

**Describe the solution you'd like**
Maybe don't make the background of the area between the start of the flash and to the top of the browser transparent. I would simply hide the text which is in between when scrolling. In my opinion it looks not good when you scroll down.

Please have a look on the screens. They should clarify what I mean (and yes, I used real gimp skills on these)

**Screenshot**

This is how it looks right now:
![service-running](https://user-images.githubusercontent.com/8287131/54532269-aa6a0c00-4987-11e9-8840-93b8d82e5035.png)

I would expect it to look like that:
![service-running-2](https://user-images.githubusercontent.com/8287131/54532554-4a279a00-4988-11e9-9270-8b2715834699.png)

Ana06:
I personally find it ok as it is already. think we are using Bootstrap classes, so this seems to be the default behavior.

Ana06:
I personally find it ok as it is already. I would rather stop it from moving, but I think we are already using Bootstrap classes, so this seems to be the default behavior.

Ana06:
I am personally not sure if it should stay when you scroll. @bgeuken it seems you implemented this, do you remember the reason for it?

saraycp:
I'm trying different options. Some GIFs to have a clearer idea:

**Right now**

![non_top_sticky_flash](https://user-images.githubusercontent.com/2581944/56495359-d3853b80-64f5-11e9-8fd1-45e36d112f7a.gif)

**Sticky on the top**

![top_sticky_flash](https://user-images.githubusercontent.com/2581944/56495442-092a2480-64f6-11e9-8711-ff4deac51cb8.gif)

**Sometimes we have more than one flash box. We should also decide how to do it.**

![2_top_sticky_flash](https://user-images.githubusercontent.com/2581944/56495553-645c1700-64f6-11e9-8fa9-fe6e81b21ffb.gif)

coolo:
:+1: - ignore the rare case of 2 flashes

saraycp:
@coolo, even if we ignore the case of 2 flashes, we also have the announcement box that looks a behaves like a flash box. Taking this into account, PR #7435 is my proposal.

Could you please have a look?

Result:

[...]

PR #7435, opened 2019-04-22
Effect on flash and announcement when scrolling

saraycp:
When scrolling, the flash and announcement boxes get stuck on the top and without space between each other.

Fixes: #7216

flash_on_scroll


saraycp:

Issue #7263, opened 2019-03-22
[beta] webui: reviewing irreviewable req leads to two boxes


jengelh:
https://build.opensuse.org/request/show/687568
if I attempt to approve a "superseded" request (which fails as it should), the so-reloaded page shows two info boxes:
![bk](https://user-images.githubusercontent.com/8861948/54823916-9e44be00-4ca9-11e9-91c1-8f1de42dc4d6.png)
The old webui would only show the failure box.

PR #7443, opened 2019-04-23
Fix flags in RequestController#modify_review

Ana06:
Render success flag only if there is not an error flag.

Fixes https://github.com/openSUSE/open-build-service/issues/7263


Ana06:

PR #7445, opened 2019-04-23
Use render partial to render a collection of comments

vpereira:
Move model rendering to the view, lift some view code to a presenter, remove ```to_xml``` from the model, since it's not model responsibility to know how to render a view.

### DIFF
--- a/src/api/spec/presenters/comments_controller_presenters/comment_presenter_spec.rb
+++ b/src/api/spec/presenters/comments_controller_presenters/comment_presenter_spec.rb
@@ -1,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe ::CommentsControllerPresenters::CommentPresenter do
+  let(:user) { create(:confirmed_user, login: 'tom') }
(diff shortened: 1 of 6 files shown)


src/api/spec/presenters/comments_controller_presenters/comment_presenter_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe ::CommentsControllerPresenters::CommentPresenter do
+  let(:user) { create(:confirmed_user, login: 'tom') }
+  let(:comment_project) { create(:comment_project, user: user) }
+  let(:comment_package) { create(:comment_package, user: user) }
+
+  describe 'obj is an user' do
```
bgeuken: I would use a more descriptive term here, like if `obj` is the comenter, I would write `Commenter is a ...`

vpereira: thanks @bgeuken, done!

PR #7461, opened 2019-04-25
Pagination maintenance

bgeuken:

PR #7466, opened 2019-04-25
Remove haml-linter project-wide

vpereira:
I personally see very small value added by the ```haml-linter``` and therefore I propose that we disable it. 


dmarcoux:
It's not powerful, but it does catch some errors, so I don't agree on disabling it.

vpereira:
> It's not powerful, but it does catch some errors, so I don't agree on disabling it.

which errors? The only issues that it catches is long lines related (which is customizable, yes) but beside that, nothing

bgeuken:
For me it's just the [datatable case](https://github.com/openSUSE/open-build-service/pull/7461/files#diff-4b151b10444bb0215d45155d08d0c202R11) were this is annoying. Because in that case there is no reasonable way to address this.
I wouldn't turn off all cops though.

dmarcoux:
haml-lint integrates with Rubocop, so it's definitely useful.

_Excerpt from haml-lint's README:_

>In addition to HAML-specific style and lint checks, it integrates with RuboCop

On top of this, we use it for much more than line length. For example, we use it to avoid using instance variables in all partials, beside the breadcrumb partials. Everything else is pretty much the [defaults](https://github.com/brigade/haml-lint/blob/2d69d92f60bf8cdc6c5d69b2953e8ad1dd9adf7b/config/default.yml), which we also use.

_Excerpt from haml-lint's README:_
> Any configuration loaded is automatically merged with the default configuration (see config/default.yml).

dmarcoux:
> For me it's just the [datatable case](https://github.com/openSUSE/open-build-service/pull/7461/files#diff-4b151b10444bb0215d45155d08d0c202R11) were this is annoying. Because in that case there is no reasonable way to address this.
> I wouldn't turn off all cops though.

It's annoying because it needs to be on multiple lines? I want to be sure I understand you correctly.

vpereira:
> haml-lint integrates with Rubocop, so it's definitely useful.
> 
> _Excerpt from haml-lint's README:_
> 
> > In addition to HAML-specific style and lint checks, it integrates with RuboCop
> 
> On top of this, we use it for much more than line length. For example, we use it to avoid using instance variables in all partials, beside the breadcrumb partials. Everything else is pretty much the [defaults](https://github.com/brigade/haml-lint/blob/2d69d92f60bf8cdc6c5d69b2953e8ad1dd9adf7b/config/default.yml), which we also use.
> 
> _Excerpt from haml-lint's README:_
> 
> > Any configuration loaded is automatically merged with the default configuration (see config/default.yml).

I don't see the point just because it integrates with Rubocop we should use it.. beside it, we are just using it on ```app/views/webui2/*``` which actually excludes 99% of our views..

dmarcoux:
We lint the new UI only because the other views will be removed once the Bootstrap migration is completed.

The point of `haml-lint` is to have consistent code style throughout our HAML views.

vpereira:
> We lint the new UI only because the other views will be removed once the Bootstrap migration is completed.

nope, we just lint views under ```webui2```. Not all "new UI views" are under ```webui2```. Please check the https://github.com/openSUSE/open-build-service/pull/7470 where I added more information

dmarcoux:
Oh right, the layouts for webui2 aren't linted. This is something we should do. I'll send a PR.

Could we have the discussions regarding the changes in one PR only? We have 3 now and it seems like the goal in the first place is to deal with linting (or not) long lines in HAML.

vpereira:
> Oh right, the layouts for webui2 aren't linted. This is something we should do. I'll send a PR.
> 
> Could we have the discussions regarding the changes in one PR only? We have 3 now and it seems like the goal in the first place is to deal with linting (or not) long lines in HAML.



> I suggest to close 2 PRs and only keep one.

The idea is to close 2 and just follow with one. This one sounds aggressive (who on earth wants to disable a Linter), but if you look closer you will see that we are just linting a minority of the whole haml views.

bgeuken:
> @dmarcoux commented an hour ago
> > For me it's just the [datatable case](https://github.com/openSUSE/open-build-service/pull/7461/files#diff-4b151b10444bb0215d45155d08d0c202R11) were this is annoying. Because in that case there is no reasonable way to address this.I wouldn't turn off all cops though.
> 
> It's annoying because it needs to be on multiple lines? I want to be sure I understand you correctly.

It's annoying because the `LineLength` cop complains about the length of the line and recommends to wrap this in a helper. But given that half of the length comes from html class definitions, which I wouldn't want to hide in a helper, and the other half comes from the `project_maintenance_incidents_path` helper, which already is a helper, the only remaining option to me is to break the line.
And this again causes the `MultiLine` cop to complain about breaking the lines.

So, long story short... one way or the other I have to disable a cop:-)

And as I wrote in my previous comment, I don't think this is a good reason to disable all cops. I am not even sure if I would want to disable any cop (MultiLine or LineLength). So far I only ran into this once. Which was annoying, but as long as I (we) don't run into this all the time I can live with it.

vpereira:
> So, long story short... one way or the other I have to disable a cop:-)

That

> So far I only ran into this once. Which was annoying, but as long as I (we) don't run into this all the time I can live with it.

Because we are just running the ```haml-linter``` in a small sets of pages, which are normally just created when we have conflicting views (under `webui2` directory) I would say that was a lucky case. I for example worked on two datatables with remote pagination, and I was 50% lucky 

To be lucky means:

* you are working on `webui` directory 
* the resource that you are working on has a short route name
* the resource that you are working on has some short selector name.

For the maintenance incidents, https://github.com/openSUSE/open-build-service/blob/master/src/api/app/views/webui2/webui/projects/maintenance_incidents/index.html.haml#L12 you ran into this issue and this funky line breaking IMO isn't better than have it in one line.

PR #7467, opened 2019-04-25
Consistent autocomplete style submit package

dmarcoux:
While submitting a package on OBS, I realized that the submit package modal and its autocomplete are not consistent with other modals/autocompletes.

See the commits for details on what has been done.

Before:
before

After:
after

### DIFF
--- a/src/api/app/views/webui2/webui/package/_submit_request_dialog.html.haml
+++ b/src/api/app/views/webui2/webui/package/_submit_request_dialog.html.haml
@@ -1,55 +1,57 @@
-.modal-dialog.modal-dialog-centered{ role: 'document' }
-  .modal-content
-    = form_tag({ controller: 'package', action: 'submit_request' }, method: 'post') do
-      .modal-header
-        %h5.modal-title Create Submit Request
-      .modal-body
-        %p
-          Do you want to submit #{package_link package, rev: revision}?
-        %p.font-italic
-          You are about to send a request to the owner of the original #{package ? 'package' : 'project'}. Please describe what you have done and
-          why they should apply your changes. The owner will be notified with the following details.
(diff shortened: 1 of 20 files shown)


src/api/app/views/webui2/webui/package/_submit_request_dialog.html.haml
```diff
@@ -1,55 +1,57 @@
...
+            Do you want to submit #{package_link package, rev: revision}?
+          %p.font-italic
+            You are about to send a request to the owner of the original #{package ? 'package' : 'project'}. Please describe what you have done and
+            why they should apply your changes. The owner will be notified with the following details.
+
+            = hidden_field_tag(:project, project)
+            = hidden_field_tag(:package, package)
+            = hidden_field_tag(:rev, revision)
+            .form-group
+              = label_tag(:sourceproject, 'From source project:')
```
vpereira: [nitpick] I know that you are not introducing it, but the html selectors shouldnt be snake case? i.e not `sourceproject` but `source_project` , not `targetproject`, `target_project`, not `targetpackage`, but `target_package`, would you mind to standardize it if you think that's right?

dmarcoux: This is a good point, but I would have to also change the bento view and the controller's action for this. I would prefer to not do it in this PR.

Fine with you?

vpereira: Ok

dmarcoux: Alright, since this was the one point to request changes, I will dismiss your review now.

PR #7468, opened 2019-04-25
Increase the LineLength for the haml-lint

vpereira:
Because bootstrap, datatables and how specific we are regarding the HTML selector names, the ```LineLength = 150```  is to short. I suggest 180

### DIFF
--- a/src/api/.haml-lint.yml
+++ b/src/api/.haml-lint.yml
@@ -1,6 +1,6 @@
 linters:
   LineLength:
-    max: 150
+    max: 180
 
   # Apply lint in all partials, but exclude 'breadcrumb_items' partials
   InstanceVariables:


vpereira:
> I find 150 already long enough. 250 is way too much, it doesn't fit on a single line inside the editor unless you use a tiny font (or have a 4K monitor).
> 
> Perhaps 180 could be an option... but to be honest, I wouldn't change this.

I changed it to 180. Still not enough if you look all the tables that use datatables but aren't under ```webui2```

vpereira:
@dmarcoux  can you explain why you choose the arbitrary value of 150, if the default is 80? https://github.com/brigade/haml-lint/blob/master/config/default.yml#L69 oderwise I don't see why we cannot set it to 180...

dmarcoux:
It's the same value as we defined for [Rubocop](https://github.com/openSUSE/open-build-service/blob/17503b9b1d300988e6b08764d21de5908ff0e696/.rubocop.yml#L55). As for why it is 150, you can ask David, he set it up :smile:. Seriously, I think 150 makes sense, 80 is way too short (and we're not in the 80s anymore with tiny screens)

dmarcoux:
Could we close this PR?

vpereira:
> Could we close this PR?

nope. we wanted to discuss about it in the standup yesterday, but you were not there.

PR #7474, opened 2019-04-25
haml-lint missing views

dmarcoux:
Following PRs from @vpereira regarding `haml-lint` (#7466, #7468, #7470), I realized that we didn't fully cover the views for the Bootstrap UI and both UIs (mailers, error pages, etc..). They are now covered with this PR, so they will be linted. Other views are for the bento UI only, which we don't want to lint since they will be removed once we complete the Bootstrap migration.

### DIFF
--- a/src/api/app/views/event_mailer/comment_for_project.html.haml
+++ b/src/api/app/views/event_mailer/comment_for_project.html.haml
@@ -1,5 +1,5 @@
 %p
-  Visit #{link_to(*[url_for(project: event['project'], controller: 'webui/project', action: :show, only_path: false, host: @host)]*2)}
+  Visit #{link_to(nil, project_show_url(event['project'], only_path: false, host: @host))}
 %p
   = event['commenter'].realname
   wrote in project #{event['project']}:
--- a/src/api/app/views/event_mailer/comment_for_request.html.haml
+++ b/src/api/app/views/event_mailer/comment_for_request.html.haml
@@ -1,5 +1,5 @@
 %p
-  Visit #{link_to(*[url_for(controller: 'webui/request', action: :show, number: event['number'], host: @host, only_path: false)]*2)}
+  Visit #{link_to(nil, request_show_url(event['number'], only_path: false, host: @host))}
 %p
   = event['commenter'].realname
   wrote in request #{event['number']}:
--- a/src/api/.haml-lint.yml
+++ b/src/api/.haml-lint.yml
@@ -42,7 +42,6 @@ exclude:
 - 'app/views/build/**/*'
 - 'app/views/comments/**/*'
 - 'app/views/distributions/**/*'
-- 'app/views/event_mailer/**/*'
 - 'app/views/exception_notifier/**/*'
 - 'app/views/group/**/*'
 - 'app/views/ichain_notifier/**/*'
--- a/src/api/.haml-lint.yml
+++ b/src/api/.haml-lint.yml
@@ -46,7 +46,7 @@ exclude:
 - 'app/views/exception_notifier/**/*'
 - 'app/views/group/**/*'
 - 'app/views/ichain_notifier/**/*'
-- 'app/views/layouts/**/*'
+- 'app/views/layouts/webui/**/*'
 - 'app/views/main/**/*'
 - 'app/views/message/**/*'
 - 'app/views/models/**/*'
--- a/src/api/.haml-lint.yml
+++ b/src/api/.haml-lint.yml
@@ -46,7 +46,6 @@ exclude:
 - 'app/views/group/**/*'
 - 'app/views/ichain_notifier/**/*'
 - 'app/views/layouts/webui/**/*'
-- 'app/views/main/**/*'
 - 'app/views/message/**/*'
 - 'app/views/models/**/*'
 - 'app/views/person/**/*'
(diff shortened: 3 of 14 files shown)


dmarcoux:
There is one offense left regarding an instance variable in the flash partial. The instance variable `@more_info` is only used in one action of the package controller, so I would prefer to get rid of it and somehow pass those details with the flash. As far as I know, there isn't a way to do this, so someone has an idea?

src/api/app/views/event_mailer/comment_for_project.html.haml
```diff
@@ -1,5 +1,7 @@
+- url = url_for(project: event['project'], controller: 'webui/project', action: :show, only_path: false, host: @host)
+
 %p
-  Visit #{link_to(*[url_for(project: event['project'], controller: 'webui/project', action: :show, only_path: false, host: @host)]*2)}
+  Visit #{link_to(*[url] * 2)}
```
vpereira: that's a trick that actually do not make the code better. That's just to make the linter happy.

src/api/app/views/event_mailer/comment_for_request.html.haml
```diff
@@ -1,5 +1,7 @@
+- url = url_for(controller: 'webui/request', action: :show, number: event['number'], host: @host, only_path: false)
+
 %p
-  Visit #{link_to(*[url_for(controller: 'webui/request', action: :show, number: event['number'], host: @host, only_path: false)]*2)}
+  Visit #{link_to(*[url] * 2)}
```
vpereira: that's a trick that actually do not make the code better. That's just to make the linter happy.

vpereira:
It still miss the  new bootstrap views created under ```webui```

src/api/app/views/event_mailer/comment_for_project.html.haml
dmarcoux: It's true... I'm guilty :smile:

vpereira: you are not guilty. The LineLength cop is :)

dmarcoux: I found a better solution, please have a look.

src/api/app/views/event_mailer/comment_for_request.html.haml
dmarcoux: I found a better solution, please have a look.

dmarcoux:
> It still miss the new bootstrap views created under `webui`

Do you mean under [src/api/app/views/webui/](https://github.com/openSUSE/open-build-service/tree/master/src/api/app/views/webui)? These views are for the bento UI only, so we don't lint them. We do lint the views under [src/api/app/views/webui2/](https://github.com/openSUSE/open-build-service/tree/master/src/api/app/views/webui2) though, since they are for the Bootstrap UI.

vpereira:
> Do you mean under [src/api/app/views/webui/](https://github.com/openSUSE/open-build-service/tree/master/src/api/app/views/webui)? These views are for the bento UI only, so we don't lint them. We do lint the views under [src/api/app/views/webui2/](https://github.com/openSUSE/open-build-service/tree/master/src/api/app/views/webui2) though, since they are for the Bootstrap UI.

Hey, I think there is a confusion here. Maybe from my part, I'm not sure, but AFAIK, in the ```webui``` we don't have just bento views. Views that have no conflicting with bento, but are new, still being created on webui directory. `Patchinfo` and `Maintained Projects` related views are those that I remember, because we worked on that on the last sprint. Interconnected had some views too, i.e src/api/app/views/webui/interconnects/new.html.haml.

bgeuken:
> Views that have no conflicting with bento, but are new, still being created on webui directory. Patchinfo and Maintained Projects related views are those that I remember, because we worked on that on the last sprint.

I thought that we create all bootstrap views in the webui2 namespace :thinking: If it's just the patchinfo and maintained projects views, I would move them over.

vpereira:
> I thought that we create all bootstrap views in the webui2 namespace  If it's just the patchinfo and maintained projects views, I would move them over.

Hm, are you going to move them over? Because there are probaly more, i.e: https://github.com/openSUSE/open-build-service/blob/master/src/api/app/views/webui/interconnects/new.html.haml  (I don't have the whole list of the views, but I'm sure that not all new views are under ```webui2```)

DavidKang:
I would not include `Interconnect` because it's already migrated (app/views/webui2/webui/interconnects/new.html.haml) neither`Patchinfo` that it also under `webui2`. Sometimes is true that we create new views under `webui` because of the refactoring like in `Patchinfo` and `Interconnect`, but the new views (bootstrap) should be under `webui2`. Even `StagingWorkflow` views is under `webui2`.

vpereira:
> I would not include `Interconnect` because it's already migrated (app/views/webui2/webui/interconnects/new.html.haml) neither`Patchinfo` that it also under `webui2`. Sometimes is true that we create new views under `webui` because of the refactoring like in `Patchinfo` and `Interconnect`, but the new views (bootstrap) should be under `webui2`. Even `StagingWorkflow` views is under `webui2`.

yup, my point not over moving it, but to enable ```haml-lint``` for all new views independent where they live

dmarcoux:
Did I miss some webui2 views in this PR? If not, please review :)

vpereira:
> Did I miss some webui2 views in this PR? If not, please review :)

As we discussed in the PR already, yes there are some news view missing, but I think we can accept this PR as it is. I just don't see a relation between that and the usefulness of the LineLength cop for haml

PR #7479, opened 2019-04-26
Add a 'merging' state to project model

bgeuken:
So far we were using the delayed jobs state for checking whether a
staging project is in "merging" state. But this has some drawbacks.
For example calling perform_now would start the project merge without
creating a delayed jobs instance.

This commit will introduce a 'merging' state that is used now to
reliably query the current state, which solves the mentioned issues.
In the far future we might even use a state machine for the
'overall_state' of a staging project. But this requires more drastic
changes and seem to be out of scope right now.

In addition we put the whole 'merging' code into a transaction to
avoid that a staging project ends up in a half-merged state.


bgeuken:
This should be deployed on a Wednesday, since it includes a migration.

coolo:
I find the use of transaction quite dangerous/racy. Because if 2 DJs can still run in parallel, they would accept requests in parallel before they notice they ran on the same things on commit. But the backend would receive possibly a mix or duplicated commits. 

I think I would prefer if creating the DJ would already update the state (creating the DJ + updating the state would be a safe transaction :) and then the DJ only accepts 'merging' SPs.

src/api/app/jobs/staging_project_accept_job.rb
```diff
@@ -3,23 +3,8 @@ class StagingProjectAcceptJob < ApplicationJob
...
-    staging_project.staged_requests.each do |staged_request|
-      if staged_request.reviews.where(by_project: staging_project.name).exists?
-        staged_request.change_review_state(:accepted, by_project: staging_project.name, comment: "Staging Project #{staging_project.name} got accepted.")
-      end
-      staged_request.change_state(newstate: 'accepted', comment: "Staging Project #{staging_project.name} got accepted.")
-      accepted_packages.concat(staged_request.bs_request_actions.map(&:target_package))
-    end
-    staging_project.packages.where(name: accepted_packages).find_each(&:destroy)
-    staging_project.staged_requests.delete_all
+    User.current = User.find_by(login: payload[:user_login])
```
hennevogel: `.find_by!`

src/api/app/models/concerns/staging_project.rb
```diff
@@ -36,6 +36,28 @@ def copy(new_project_name)
     end
   end
 
+  def accept
+    clear_memoized_data
+    return unless overall_state == :acceptable
+
+    transaction do
```
hennevogel: :thought_balloon: This screams deadlock to me...

hennevogel:
We could also just not call `StagingProjectAcceptJob.perform_now`. Let's close this for now until we come up with a solution that isn't so invasive.

Issue #7021, opened 2019-02-21
Repository.name validation


hennevogel:
Repository names currently have a high degree of freedom in the frontend. E.g. there can be spaces in the names or things like `*` and `;` or  `--`. All of those the backend can not handle. We should sync the validation, my guess is that the rules are similar to `Project.valid_name?`

saraycp:
I've created a draft PR for this: #7481.

Now the validation only allows signs like `-`, ` _`, `.` and `+`, getting names as complex as:
"devel-libraries_c_c++_openSUSE_Leap_42.3".

I've taken all the repository names we have in production (more than 1400 different names) and most of them match this new validation. Except around 40 of them, because they contain `spaces`. `\`, `ü` or `*` and I guess they shouldn't be considered as valid names anyway.

However, there is a test in [spec/models/repository_spec.rb](https://github.com/openSUSE/open-build-service/blob/7b24d1998932804a6a422a426c7a2736ffc6b06f/src/api/spec/models/repository_spec.rb#L17) that expect `fOO_-§$&!#+~()=?\\"` to be a valid name. But some of these characters are not even allowed by the backend (i.e. § is one on the range \000-\037 forbidden by the [backend](https://github.com/openSUSE/open-build-service/blob/7b24d1998932804a6a422a426c7a2736ffc6b06f/src/backend/BSVerify.pm#L66)) and almost none of them appear in the list of more than 1400 repository names from production.
  
Do we really have to allow all this kind of signs? Is it enough with the new validation?

hennevogel:
> (but are well supported by the backend)

This%20is%20good%20to%20know

coolo:
Compare https://download.opensuse.org/repositories/home:/coolo/openSUSE%20Leap%2015.1/ :)

PR #7481, opened 2019-04-26
Improve repository's name validation

saraycp:
Until now, the validation for the repository name didn't avoid:
  - spaces
  - characters like: * or ;
  - double hyphens or underscores

Please read the issue first, there are some cases I'm not sure if we have to validate.
Fixes: #7021

### DIFF
--- a/src/api/app/models/repository.rb
+++ b/src/api/app/models/repository.rb
@@ -22,9 +22,7 @@ class Repository < ApplicationRecord
   scope :remote, -> { where.not(remote_project_name: '') }
 
   validates :name, length: { in: 1..200 }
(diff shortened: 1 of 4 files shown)


hennevogel:
We would also need a data migration as we have repos that do not follow this currently. @mlschroe what's your opinion here? You can create all kind of weird [repo names](https://build-test.opensuse.org/projects/home:hennevogel/meta). Is this worth it?

coolo:
@mlschroe ?

coolo:
@mlschroe ?

hennevogel:
I guess no answer means it's not worth it.

PR #7484, opened 2019-04-26
Update maintenance incidents table

bgeuken:
* Merge release targets and build result column
* Make release target project linkable
* Add request creation time next to request flags

Before:

https://build.opensuse.org/projects/openSUSE:Maintenance/maintenance_incidents

After:

### DIFF
--- a/src/api/app/datatables/maintenance_incident_datatable.rb
+++ b/src/api/app/datatables/maintenance_incident_datatable.rb
@@ -26,8 +26,7 @@ class MaintenanceIncidentDatatable < Datatable
         category: category_cell(record, patchinfo),
         packages: packages_cell(record, release_targets_ng),
         info: info_cell(record, patchinfo),
-        release_targets: release_targets_cell(release_targets_ng),
-        build_results: build_results_cell(record, release_targets_ng)
+        release_targets: release_targets_cell(record, release_targets_ng)
       }
     end
   end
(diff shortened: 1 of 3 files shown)


bgeuken:
[Maintenance incidents](https://trello.com/c/4hbIZuK9/1095-maintenance-incidents)

vpereira:
@bgeuken the categories could be just recommended/security. I'm not sure why are we adding the "missing patchinfo" as category?

bgeuken:
I guess it's to raise that a patchinfo has to be added. It's also linked, so one can just click on it to generate one.

So, I don't know. But since it was already there and it seems useful I kept it.

src/api/app/datatables/maintenance_incident_datatable.rb
```diff
@@ -26,8 +26,7 @@ def data
         category: category_cell(record, patchinfo),
```
vpereira: Do it make sense to have in category, something that isn't a category?

bgeuken: So far it made sense for the maintenance team (bento UI). I am reluctant to change this without talking with them first.

Note that this is already part of the view, old and new UI.

vpereira: ah ok, sorry for the noise

PR #7488, opened 2019-04-27
Move set_sourceupdate_default to an explicit action in the controller

coolo:
The default is in the UI view, so it's all about the API controller and that one knows when to call it, no reason to 'fix it' in the check functions of the model

### DIFF
--- a/src/api/app/controllers/request_controller.rb
+++ b/src/api/app/controllers/request_controller.rb
@@ -152,8 +152,12 @@ class RequestController < ApplicationController
     end
 
     # cache the diff (in the backend)
-    @req.bs_request_actions.each do |a|
-      BsRequestActionWebuiInfosJob.perform_later(a)
+    @req.bs_request_actions.each do |action|
+      # cleanup implicit home branches.
+      # FIXME3.0: remove this, the clients should do this automatically meanwhile
+      action.set_sourceupdate_default(User.session!)
+      # cache the diff (in the backend)
+      BsRequestActionWebuiInfosJob.perform_later(action)
     end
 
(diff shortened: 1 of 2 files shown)


src/api/app/models/bs_request_action.rb
```diff
@@ -900,6 +891,12 @@ def set_target_associations
     self.target_project_object = Project.find_by_name(target_project)
   end
 
+  def set_sourceupdate_default(user)
+    return unless action_type.in?([:submit, :maintenance_incident])
+    return if sourceupdate
+    update(sourceupdate: 'cleanup') if target_project && user.branch_project_name(target_project) == source_project
```
bgeuken: Might be a matter of taste... I would combine the two if conditions

coolo: fine with me

src/api/app/controllers/request_controller.rb
```diff
@@ -152,8 +152,12 @@ def request_create
     end
 
     # cache the diff (in the backend)
-    @req.bs_request_actions.each do |a|
-      BsRequestActionWebuiInfosJob.perform_later(a)
+    @req.bs_request_actions.each do |action|
+      # cleanup implicit home branches.
+      # FIXME3.0: remove this, the clients should do this automatically meanwhile
+      action.set_sourceupdate_default(User.session!)
```
coolo: @bgeuken so here you have a use of session! that is not a pundit action

coolo:
I need this one to solve a codeclimate complaint with the User.current change - shall I merge the 2?

coolo:
merging

PR #7490, opened 2019-04-27
Don't store test package into backend

coolo:
Otherwise it requires a user, which we don't want here


hennevogel:
`commit_opts` the poor mans cassette

coolo:
at least it's honest! :)

PR #7491, opened 2019-04-27
Extend User's API

coolo:
Deprecate User.current in favor of split usage of User.possibly_nobody and User.required

The later will throw an exception if no user is logged in (controller needs to make sure it requires login), the former behaves like User.current

In addition introduce User.not_required that may return nil if no user is logged in and User.admin_session? as wrapper to simplify checks for admin user being logged in.

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -218,23 +218,30 @@ class User < ApplicationRecord
 
   # The currently logged in user (might be nil). It's reset after
   # every request and normally set during authentification
+  # Prefer possibly_nobody, session or session! - this function is scheduled
+  # for removal, it doesn't express the intention what to do in
+  # case of anonymous access well enough.
   def self.current
     Thread.current[:user]
-  end
-
-  def self.nobody
-    Thread.current[:nobody] ||= find_nobody!
   end
 
   # Currently logged in user or nobody user if there is no user logged in.
   # Use this to check permissions, but don't treat it as logged in user. Check
   # is_nobody? on the returned object
-  def self.current_or_nobody
+  def self.possibly_nobody
     current || nobody
   end
 
-  def self.session=(user)
-    Thread.current[:user] = user
+  # Currently logged in user. Will thrown an exception if no user is logged in.
+  # So the controller needs to require login if using this (or models using it)
+  def self.session!
+    raise ArgumentError, 'Requiring user, but found nobody' unless session
+    current
+  end
+
+  # Currently logged in user or nil
+  def self.session
+    current if current && !current.is_nobody?
   end
 
   def self.current_login
@@ -243,6 +250,11 @@ class User < ApplicationRecord
 
   def self.admin_session?
     current && current.is_admin?
+  end
+
+  # set the user as current session user (should be real user)
+  def self.session=(user)
+    Thread.current[:user] = user
   end
 
   def self.get_default_admin
@@ -868,6 +880,11 @@ class User < ApplicationRecord
   end
 
   private
+
+  def self.nobody
+    Thread.current[:nobody] ||= find_nobody!
+  end
+  private_class_method :nobody
 
   def password_validation
     return if password_digest || deprecated_password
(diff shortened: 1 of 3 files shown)


bgeuken:
There is a typo in the commit message:

> The lat**t**er will throw

coolo:
I also moved the admin_session? function in an extra PR

coolo:
On how to use these functions, look at https://github.com/openSUSE/open-build-service/pull/7219

src/api/app/models/user.rb
```diff
@@ -218,29 +218,39 @@ def self.find_with_credentials_via_ldap(login, password)
...
-    Thread.current[:user] = user
+  # Currently logged in user. Will thrown an exception if no user is logged in.
+  # So the controller needs to require login if using this (or models using it)
+  def self.required
+    raise ArgumentError, 'Requiring user, but found nobody' unless not_required
+    current
+  end
+
+  # Currently logged in user or nil
+  def self.not_required
```
coolo: I had second thoughts on these 2. I would by now prefer `User.session` for `User.not_required` and `User.session!` for `User.required`.

coolo:
I would really like to have a review here - as I need to rebase the real PR.

You guys wanted the big PR to be split up - and now you can't get the small PRs reviewed :(

src/api/app/models/user.rb
```diff
@@ -218,23 +218,28 @@ def self.find_with_credentials_via_ldap(login, password)
 
   # The currently logged in user (might be nil). It's reset after
   # every request and normally set during authentification
+  # DEPRECATED! Prefer possibly_nobody or required
```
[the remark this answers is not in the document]
coolo: well, I will make it private after the other PR is merged. This deprecation is just interim

src/api/app/models/user.rb
```diff
@@ -218,23 +218,28 @@ def self.find_with_credentials_via_ldap(login, password)
...
   end
 
-  def self.session=(user)
-    Thread.current[:user] = user
+  # Currently logged in user. Will thrown an exception if no user is logged in.
+  # So the controller needs to require login if using this (or models using it)
+  def self.session!
+    raise ArgumentError, 'Requiring user, but found nobody' unless session
+    current
+  end
```
bgeuken: Could we move this to to the pundit policies right away? The method doesn't seem to be used right now anyway. So this wouldn't require to much effort, right?

And on second thought... did this slip in? Given that it's unused...

src/api/app/models/user.rb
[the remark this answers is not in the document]
hennevogel: that is info that should be here IMHO

src/api/app/models/user.rb
coolo: Guys, you're killing me. When I sent a PR with the new API and 220 files using it, it was too big for you. Now I split the new API so you can review it in a special PR and now you're complaining the new API is unused?

I quote from a comment of 9 days ago: 
> On how to use these functions, look at #7219

That comment is slightly outdated now as I renamed User.required to User.session!, but it's still the same. Moving this to pundit policies is a per usage effort

As a matter of fact, many controllers already use `before_login :require_login`, so the User.session! usage there is pointless, but can be useful for the models used.

hennevogel: pundit? you have to explain that to me please

hennevogel: > Guys, you're killing me. 

I understand your frustration. Please keep in mind that this is a big change, it's not so easy to follow.

src/api/app/models/user.rb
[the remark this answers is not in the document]
coolo: Updated the comment

src/api/app/models/user.rb
bgeuken: > When I sent a PR with the new API and 220 files using it, it was too big for you

I reviewed that one as well :wink:

> and now you're complaining the new API is unused?

My ~~complain~~ question was if we could already move such things into pundit.
And given that this, and the other PR, are just interim changes it's natural that such questions come up:-)

bgeuken:
> You guys wanted the big PR to be split up - and now you can't get the small PRs reviewed :(

Apologies for being on vacation :tongue:

coolo:
> > You guys wanted the big PR to be split up - and now you can't get the small PRs reviewed :(
> 
> Apologies for being on vacation

You're not the one that complained in the big PR

PR #7492, opened 2019-04-27
Rename User.current= to User.session=

coolo:
In alignment of deprecating User.current, rename the assignment operator to session=


src/api/app/jobs/staging_project_copy_job.rb
```diff
@@ -1,7 +1,7 @@
 class StagingProjectCopyJob < ApplicationJob
   def perform(staging_workflow_project_name, original_staging_project_name, staging_project_copy_name, user_id)
     # This is needed as the job depends on the current user and without it, it will failed when performed later
-    User.current ||= User.find(user_id)
+    User.session = User.current || User.find(user_id)
```
coolo: I wonder why it would accept any other user happened to be set in the delayed job thread. IMO it should be a plain assingment. But it's out of scope for this PR

bgeuken: You are right, this looks wrong

hennevogel:
Phew...

PR #7493, opened 2019-04-27
Introduce User.admin_session?

coolo:
User.current.is_admin? may crash if User.current is not there, so it needs to be guarded, which is a lot of duplicated code

### DIFF
--- a/src/api/app/models/unregistered_user.rb
+++ b/src/api/app/models/unregistered_user.rb
@@ -8,7 +8,7 @@ class UnregisteredUser < User
   # Returns true if a user can register
   def self.can_register?
     # No registering if LDAP is on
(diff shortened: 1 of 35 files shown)


src/api/app/models/unregistered_user.rb
```diff
@@ -44,7 +44,7 @@ def self.can_register?
   def self.register(opts)
     can_register?
 
-    opts[:note] = nil unless User.current && User.current.is_admin?
+    opts[:note] = nil unless User.current && User.admin_session?
```
bgeuken: With User#admin_session? defined as `current && current.is_admin?` this doesn't make sense:wink:

coolo: hmm, not sure how I missed it

PR #7495, opened 2019-04-28
Refactor is_target_maintainer

coolo:
The old code was a little hard to parse

### DIFF
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -927,26 +927,8 @@ class BsRequest < ApplicationRecord
   end
 
   # Check if 'user' is maintainer in _all_ request targets:
-  def is_target_maintainer?(user = User.current)
-    has_target = false
-    is_target_maintainer = true
-    bs_request_actions.each do |a|
-      next unless a.target_project
-      if a.target_package
-        tpkg = Package.find_by_project_and_name(a.target_project, a.target_package)
-        if tpkg
-          has_target = true
-          is_target_maintainer &= user.can_modify?(tpkg)
-          next
-        end
-      end
-      tprj = Project.find_by_name(a.target_project)
-      if tprj
-        has_target = true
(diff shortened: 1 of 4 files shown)


src/api/app/models/bs_request.rb
```diff
@@ -927,26 +927,8 @@ def auto_accept
...
-      end
-      tprj = Project.find_by_name(a.target_project)
-      if tprj
-        has_target = true
-        is_target_maintainer &= user.can_modify?(tprj)
-      end
-    end
-    has_target && is_target_maintainer
+  def is_target_maintainer?(user)
+    bs_request_actions.map { |action| action.is_target_maintainer?(user) }.all?
```
bgeuken: wennschon dennschon:smiley:
```ruby
 bs_request_actions.all? { |action| action.is_target_maintainer?(user) }
```

coolo: Hmm, yesterday I tried and was somehow led to believe that .all? would return false on empty array. Today it does not - which is a problem

coolo: well, it's not - as we have no requests without actions. just the old `has_target` logic looked for it. But an action would target will return false on is_target_action? (I guess I could write specs for that)

coolo: done that now - sorry for the early morning brabling :)

src/api/spec/models/bs_request_action_spec.rb
```diff
@@ -143,5 +143,36 @@
...
+
+      context 'on home target' do
+        let(:another_user) { create(:confirmed_user) }
+        let(:bs_request) do
+          create(:set_bugowner_request, target_project: user.home_project)
+        end
+        let(:bs_request_action) { bs_request.bs_request_actions.first }
+
+        it 'is true for user' do
+          expect(bs_request_action).to be_is_target_maintainer(user)
```
coolo: that looks very strange, but that's how rubocop wrote it

PR #7499, opened 2019-04-29
Re-write build_succeed? method

bgeuken:

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1323,43 +1323,23 @@ class Project < ApplicationRecord
     Backend::Api::Build::Project.wipe_binaries(name)
   end
 
-  def build_succeeded?(repository)
-    states = {}
-    repository_states = {}
+  def build_succeeded?(repo_name)
+    begin
+      build_result = Xmlhash.parse(Backend::Api::BuildResults::Status.failed_results_summary(name, repo_name))
+    rescue Backend::NotFoundError
+      return false
+    end
 
-    br = Buildresult.find_hashed(project: name, view: 'summary')
-    # no longer there?
-    return false if br.empty?
+    return false if build_result.empty?
 
-    br.elements('result') do |result|
-      if result['repository'] == repository
-        repository_states[repository] ||= {}
-        result.elements('summary') do |summary|
-          summary.elements('statuscount') do |statuscount|
-            repository_states[repository][statuscount['code']] ||= 0
-            repository_states[repository][statuscount['code']] += statuscount['count'].to_i
-          end
-        end
-      else
-        result.elements('summary') do |summary|
-          summary.elements('statuscount') do |statuscount|
-            states[statuscount['code']] ||= 0
-            states[statuscount['code']] += statuscount['count'].to_i
-          end
-        end
+    build_result.elements('result').each do |result|
+      result.elements('summary').each do |summary|
+        # Since we query for failed, broken or unresolvable, a summary element
+        # with content means there was an "unsuccessful" build
+        return false if summary.present?
       end
     end
-    if repository_states.key?(repository)
-      return false if repository_states[repository].empty? # No buildresult is bad
-      repository_states[repository].each do |state, _|
-        return false if state.in?(['broken', 'failed', 'unresolvable'])
-      end
-    else
-      return false unless states.empty? # No buildresult is bad
-      states.each do |state, _|
-        return false if state.in?(['broken', 'failed', 'unresolvable'])
-      end
-    end
+
     true
   end
 
--- a/src/api/lib/backend/api/build_results/status.rb
+++ b/src/api/lib/backend/api/build_results/status.rb
@@ -48,8 +48,7 @@ module Backend
         # Lists failed package builds in a project
         # @return [String]
         def self.failed_results(project_name)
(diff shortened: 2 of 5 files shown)


src/api/app/models/project.rb
```diff
@@ -1323,43 +1323,18 @@ def reopen_release_targets
...
-      return false if repository_states[repository].empty? # No buildresult is bad
-      repository_states[repository].each do |state, _|
-        return false if state.in?(['broken', 'failed', 'unresolvable'])
-      end
-    else
-      return false unless states.empty? # No buildresult is bad
-      states.each do |state, _|
-        return false if state.in?(['broken', 'failed', 'unresolvable'])
+  def build_succeeded?(repo_name)
+    build_result = Buildresult.find_hashed(project: name, repository: repo_name, view: 'summary', code: [:failed, :broken, :unresolvable])
```
coolo: Please no longer use this function - it's just a wrapper around `result_swiss_knife`. Better add an explicit function to `Backend::Api::BuildResults`

bgeuken: It also converts the backend response to an `XmlHash`

coolo: Still just a wrapper

bgeuken: Which isn't a bad thing per se. We use it at many places and spares us from catching backen erorrs, converting the xml, etc. What's so bad about it?

coolo: I won't 'request changes', but note that I believe he's stronger than you: 
https://github.com/openSUSE/open-build-service/pull/5556#issuecomment-410986185

bgeuken: Ah, I though you want me to directly use the `swiss_army_knife` method.

bgeuken:
@coolo Updated. Please have another look

src/api/lib/backend/api/build_results/status.rb
```diff
@@ -47,8 +47,9 @@ def self.result_swiss_knife(project_name, params = {})
 
         # Lists failed package builds in a project
         # @return [String]
-        def self.failed_results(project_name)
-          response = http_get(['/build/:project/_result', project_name], params: { code: %w[failed broken unresolvable] }, expand: [:code])
+        def self.failed_results(project_name, params = {})
```
coolo: I'm only taking forward other's ideas, but this is not what was intended.

Rather an extra `self.failed_result_summary_for_repo(project_name, repository_name)`

what you're doing here is tempting, but is just swiss armying another backend function.

bgeuken: If the plan is to have individual method calls for certain query types, eg. one for '/build/:project/_result' and query parameter `view: :status` and one for '/build/:project/_result' and query parameter `view: :summary`, I can understand that.

But for every combination of API query parameters? That seems to much. For '/build/:project/_result' there would be architecture, repository, package, code, view and maybe more.
In the end we would end up adding a new layer, which wraps these method calls per API endpoints. Because for each API endpoint they would only differ by their params.

coolo: Well, the idea (and again, it's not mine, I'm argueing here for the sake of consistency) is to abstract how the backend wants it from how the API wants it. If you put `code: 'summary'` in your model, you're putting 'how the backend wants it' logic in API code.

And in the end, the backend does different things for different views, so having 2 functions in lib/backend doesn't sound that wrong.

bgeuken: > to abstract how the backend wants it from how the API wants it.

That sounds like a good plan.

> If you put code: 'summary' in your model, you're putting 'how the backend wants it' logic in API code.

No, that's how we want the data for this particular case: The minimal data to know if a package build failed, aka. matches 3 specific codes.

> And in the end, the backend does different things for different views, so having 2 functions in lib/backend doesn't sound that wrong.

The devil is in the detail. If I want an index view, with a query parameter (aka. filter), you don't need two different endpoints for this.
And in this case I want the build results for a specific project filtered by their `code` (being unresolvable, broken or failed). 

If you meant `view: 'summary` and not `code` then I already agreed with that one:
> If the plan is to have individual method calls for certain query types, eg. one for '/build/:project/_result' and query parameter view: :status and one for '/build/:project/_result' and query parameter view: :summary, I can understand that.

bgeuken: I changed parameter type from a hash to plain variables. Though I didn't add the `_for_repo` suffix to  the method name as suggested, since none of the others do that.

src/api/app/models/project.rb
```diff
@@ -1323,43 +1323,23 @@ def reopen_release_targets
...
-        result.elements('summary') do |summary|
-          summary.elements('statuscount') do |statuscount|
-            states[statuscount['code']] ||= 0
-            states[statuscount['code']] += statuscount['count'].to_i
-          end
-        end
-      end
+  def build_succeeded?(repo_name)
+    begin
+      build_result = Xmlhash.parse(Backend::Api::BuildResults::Status.failed_results_summary(name, repository: repo_name))
```
coolo: you didn't change the call to be a plain variable though

coolo: that means the function isn't covered, right?

PR #7516, opened 2019-05-01
Split number of request into incoming/outgoing #4602

### LINKED ISSUE #4602: Split number of requests into incoming/outgoing
As a maintainer of larger projects, like GNOME:Factory, the number of 'open requests' can become increasingly large, which in itself is not alarming.

This is, for example, a moment view into openSUSE:Factory:
screenshot from 2018-03-01 12-31-32

It would be helpful if the number of open requests could be split into how many are incoming (those are usually things the project maintainer have to worry for, to make things moving) and outgoing (somebody else has to review/accept them, I merely need to monitor them for declines/comments)

jesus-montano:
Add incoming_requests and outgoing_requests variables dividing `requests`
variable: reqs incidents and maintanance_release to incoming and targets and
reviews in the outgoing.
Use these variables in show, side_links and requests views.
change requests value as incomin+outgoing requests

Closes #4602

### DIFF
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -654,7 +654,9 @@ class Webui::ProjectController < Webui::WebuiController
     @linking_projects = @project.linked_by_projects.pluck(:name)
 
     reqs = @project.open_requests
-    @requests = (reqs[:reviews] + reqs[:targets] + reqs[:incidents] + reqs[:maintenance_release]).sort!.uniq
+    @incoming_requests = reqs[:incidents] | reqs[:maintenance_release]
+    @outgoing_requests = reqs[:targets] | reqs[:reviews]
+    @requests = @incoming_requests | @outgoing_requests
 
     @nr_of_problem_packages = @project.number_of_build_problems
   end
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -653,9 +653,8 @@ class Webui::ProjectController < Webui::WebuiController
     @ipackages = @project.expand_all_packages.find_all { |ip| !@packages.map { |p| p[0] }.include?(ip[0]) }
     @linking_projects = @project.linked_by_projects.pluck(:name)
 
-    reqs = @project.open_requests
-    @incoming_requests = reqs[:incidents] | reqs[:maintenance_release]
-    @outgoing_requests = reqs[:targets] | reqs[:reviews]
+    @incoming_requests = @project.incoming_requests
+    @outgoing_requests = @project.outgoing_requests
     @requests = @incoming_requests | @outgoing_requests
 
     @nr_of_problem_packages = @project.number_of_build_problems
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1253,17 +1253,17 @@ class Project < ApplicationRecord
 
   def outgoing_requests
     incidents = BsRequest.with_involved_projects(id)
-    .or(BsRequest.from_source_project(name))
-    .in_states(:new)
-    .with_types(:maintenance_incident)
-    .pluck(:number)
+                         .or(BsRequest.from_source_project(name))
+                         .in_states(:new)
+                         .with_types(:maintenance_incident)
+                         .pluck(:number)
 
     if is_maintenance?
       maintenance_release = BsRequest.with_target_subprojects(name + ':%')
-                      .or(BsRequest.with_source_subprojects(name + ':%'))
-                      .in_states(:new)
-                      .with_types(:maintenance_release)
-                      .pluck(:number)
+                                     .or(BsRequest.with_source_subprojects(name + ':%'))
+                                     .in_states(:new)
+                                     .with_types(:maintenance_release)
+                                     .pluck(:number)
     else
       maintenance_release = []
     end
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1237,6 +1237,39 @@ class Project < ApplicationRecord
     { reviews: reviews, targets: targets, incidents: incidents, maintenance_release: maintenance_release }
   end
 
+  def incoming_requests
+    reviews = BsRequest.where(id: BsRequestAction.bs_request_ids_of_involved_projects(id)).or(
+      BsRequest.where(id: BsRequestAction.bs_request_ids_by_source_projects(name)).or(
+        BsRequest.where(id: Review.bs_request_ids_of_involved_projects(id))
+      )
+    ).in_states(:review).distinct.order(priority: :asc, id: :desc).pluck(:number)
+
+    targets = BsRequest.with_involved_projects(id)
+                       .or(BsRequest.from_source_project(name))
+                       .in_states(:new).with_actions
+                       .pluck(:number)
+
+  end
+
+  def outgoing_requests
+    incidents = BsRequest.with_involved_projects(id)
+    .or(BsRequest.from_source_project(name))
+    .in_states(:new)
+    .with_types(:maintenance_incident)
+    .pluck(:number)
+
+    if is_maintenance?
+      maintenance_release = BsRequest.with_target_subprojects(name + ':%')
+                      .or(BsRequest.with_source_subprojects(name + ':%'))
+                      .in_states(:new)
+                      .with_types(:maintenance_release)
+                      .pluck(:number)
+    else
+      maintenance_release = []
+    end
+
+  end
+
   # for the clockworkd - called delayed
   def update_packages_if_dirty
     packages.dirty_backend_package.each(&:update_if_dirty)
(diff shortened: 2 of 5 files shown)


src/api/app/controllers/webui/project_controller.rb
```diff
@@ -655,7 +655,9 @@ def load_project_info
 
     reqs = @project.open_requests
     @requests = (reqs[:reviews] + reqs[:targets] + reqs[:incidents] + reqs[:maintenance_release]).sort!.uniq
-
+    @incoming_requests =  (reqs[:incidents] + reqs[:maintenance_release]).sort!.uniq
```
dmarcoux: Incoming requests are requests for which the target project is the project (`@project` in this context).

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -655,7 +655,9 @@ def load_project_info
...
+    @outgoing_requests = (reqs[:targets] +reqs[:reviews]).sort!.uniq
```
dmarcoux: Outgoing requests are requests for which the source project is the project (`@project` in this context)

src/api/app/views/webui2/webui/project/side_links/_requests.html.haml
```diff
@@ -9,4 +9,8 @@
     end
   %i.fa.fa-info-circle.text-info
   = link_to(path) do
-    = pluralize(requests.size, 'open request')
+    = pluralize(incoming_requests.size, 'Incoming request' )
+  %br
```
dmarcoux: Instead of using a `<br>`, these links should be in list items (`<li>`), so this partial should be:

```
%li
  link for incoming requests
%li
  link for outgoing requests
```

src/api/app/controllers/webui/project_controller.rb
jesus-montano: oh, then I mixed concepts and I have the splits upside down? 
or how can I fix it?

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -655,6 +655,8 @@ def load_project_info
 
     reqs = @project.open_requests
     @requests = (reqs[:reviews] + reqs[:targets] + reqs[:incidents] + reqs[:maintenance_release]).sort!.uniq
```
vpereira: if we still need  the variable ```@requests``` wouldn't be better to have it as ```@requests = @incoming_requests + @outgoing_requests``` to avoid to recalculate it again?

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -655,6 +655,8 @@ def load_project_info
...
+    @incoming_requests = (reqs[:incidents] + reqs[:maintenance_release]).sort!.uniq
```
Ana06: `sort` + `uniq` is equivalent to `|` which is way more efficient. In Ruby 2.6, you can also use `union` (but we are using 2.5). More info and metrics: http://anamaria.martinezgomez.name/2018/12/21/ruby2_6.html

src/api/app/controllers/webui/project_controller.rb
Ana06: Yes, I think we need `|` though as they could be duplicated.

src/api/app/views/webui2/webui/project/side_links/_requests.html.haml
```diff
@@ -1,12 +1,17 @@
 %li
   :ruby
-    if requests.size == 1
+    if incoming_requests.size == 1 || outgoing_requests.size == 1
```
Ana06: what happen if both are 1? I mean that you could have 2 requests.

why not just keeping using `requests` here?

vpereira: unless you are intentionally changing the logic, You should use either ```requests.size``` or ```incoming_requests.size + outgoing_requests.size  == 1```

Ana06: `incoming_requests.size + outgoing_requests.size == 1` could also be wrong, as the same one could be in both.

jesus-montano: then you recommended leave requests.size == 1 ?

jesus-montano:
I have done squash of my commits already if you can check it, would great

src/api/app/views/webui2/webui/project/side_links/_requests.html.haml
vpereira: > then you recommended leave requests.size == 1 ?

yes

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -654,7 +654,9 @@ def load_project_info
     @linking_projects = @project.linked_by_projects.pluck(:name)
 
     reqs = @project.open_requests
-    @requests = (reqs[:reviews] + reqs[:targets] + reqs[:incidents] + reqs[:maintenance_release]).sort!.uniq
+    @incoming_requests = reqs[:incidents] | reqs[:maintenance_release]
+    @outgoing_requests = reqs[:targets] | reqs[:reviews]
+    @requests = @incoming_requests + @outgoing_requests
```
vpereira: as @Ana06 said, would be better to have:

```
@requests = @incoming_requests | @outgoing_requests 
```
in this way you are selecting just the unique requests

jesus-montano:
can you check it again to know if it's fine or need another change?

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -654,7 +654,9 @@ def load_project_info
     @linking_projects = @project.linked_by_projects.pluck(:name)
 
     reqs = @project.open_requests
```
dmarcoux: This line returns all open requests, without any distinction between incoming and outgoing requests.

Incoming requests are requests for which the target project is the project (`@project` in this context).

Outgoing requests are requests for which the source project is the project (`@project` in this context).

So with this in mind, we need to filter them depending on those conditions.

src/api/app/controllers/webui/project_controller.rb
dmarcoux: Sure, but still, it's the core logic is broken. `@incoming_requests` doesn't contain incoming requests and `@outgoing_requests` doesn't contain outgoing requests. They're all mixed up. See [this comment](https://github.com/openSUSE/open-build-service/pull/7516#discussion_r283229697)

src/api/app/controllers/webui/project_controller.rb
vpereira: @dmarcoux maybe filtering them, creating two instance methods in ```Project```, one named ```incoming_requests``` and other named ```outgoing_requests```  ?

jesus-montano: let me see if I understood correctly. Do you want to rewrite the open_requests method in project.rb in models and divide it into incoming and outgoing?
incoming_requests should contain review and targets and outgoing should contain incidents and maintenance? Or what should be in each one?

dmarcoux: `open_requests` is used in other files, so I would instead create 2 methods, one for incoming requests and another for outgoing requests.

src/api/app/models/project.rb
```diff
@@ -1253,17 +1253,17 @@ def incoming_requests
...
     if is_maintenance?
       maintenance_release = BsRequest.with_target_subprojects(name + ':%')
-                      .or(BsRequest.with_source_subprojects(name + ':%'))
-                      .in_states(:new)
-                      .with_types(:maintenance_release)
-                      .pluck(:number)
+                                     .or(BsRequest.with_source_subprojects(name + ':%'))
+                                     .in_states(:new)
+                                     .with_types(:maintenance_release)
+                                     .pluck(:number)
```
vpereira: @jesus-montano try to run locally ```rake dev:lint:rubocop:auto_correct```  to fix those identations.. then please squash the commits related with that

jesus-montano: oh, great ty I will do it

src/api/app/models/project.rb
```diff
@@ -1237,6 +1237,39 @@ def open_requests
...
+                       .in_states(:new).with_actions
+                       .pluck(:number)
+
+  end
+
+  def outgoing_requests
+    incidents = BsRequest.with_involved_projects(id)
+                         .or(BsRequest.from_source_project(name))
+                         .in_states(:new)
+                         .with_types(:maintenance_incident)
```
dmarcoux: Outgoing requests are not necessarily maintenance incidents, so this should be removed.

src/api/app/models/project.rb
```diff
@@ -1237,6 +1237,39 @@ def open_requests
...
+                         .pluck(:number)
+
+    if is_maintenance?
+      maintenance_release = BsRequest.with_target_subprojects(name + ':%')
+                                     .or(BsRequest.with_source_subprojects(name + ':%'))
+                                     .in_states(:new)
+                                     .with_types(:maintenance_release)
```
dmarcoux: Same as above. Outgoing requests are not necessarily maintenance incidents, so this should be removed.

hennevogel:
@jesus-montano do you think you can finish this? Anything we can do to help you?

vpereira:
I'm closing it and tracking this issue and the ideas in the issue #4602

PR #7522, opened 2019-05-02
Apply background color in the diff section when switching between tabs of a request

### LINKED ISSUE #7500: New UI: diff highlighting gets partially removed when switchting between tabs
# Issue Description
When there's a submission with several packages, the diff highlighting gets partially removed when switching between tabs. This means, the color highlighting is there, but the background color is gone.

Expected Result
============
It should not get removed (even if I would say removing the background color from diffs looks better ;-) )

How to Reproduce
===============
1. Go to request 188253 on our internal build service
2. Switch between tabs)

eduardoj:
When there's a submission with several packages, the diff highlighting gets partially removed when switching between tabs. This means, the color highlighting is there, but the background color is gone. This PR fixes the issue.

Fixes #7500.

request_tabs_background

### DIFF
--- a/src/api/app/assets/javascripts/webui/application/request.js
+++ b/src/api/app/assets/javascripts/webui/application/request.js
@@ -203,7 +203,7 @@ function setupActionLink() { // jshint ignore:line
     $('#action_display_' + index).show();
     // It is necessary to refresh the CodeMirror editors after switching tabs to initialise the dimensions again.
     // Otherwise the editors are empty after calling show().
-    editors.forEach( function(editor) { editor.refresh(); });
+    editors.forEach( function(editor) { editor.refresh(); cm_mark_diff_lines(); });
     return false;
   });
 }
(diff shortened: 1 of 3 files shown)


vpereira:
would you mind to add a gif here @eduardoj ?

src/api/app/assets/javascripts/webui/application/request.js
```diff
@@ -203,7 +203,7 @@ function setupActionLink() { // jshint ignore:line
     $('#action_display_' + index).show();
     // It is necessary to refresh the CodeMirror editors after switching tabs to initialise the dimensions again.
     // Otherwise the editors are empty after calling show().
-    editors.forEach( function(editor) { editor.refresh(); });
+    editors.forEach( function(editor) { editor.refresh(); cmMarkDiffLines(); }); // jshint ignore:line
```
Ana06: why do you need to ignore the linter?

eduardoj: > why do you need to ignore the linter?

Otherwise I get an `'cmMarkDiffLines' is not defined` error with jshint, as written in the commit message.

The linter is defined to run in the dev.rake file this way: `jshint ./app/assets/javascripts/`. And `cmMarkDiffLines` is defined in `./app/assets/javascripts/webui2/cm2/use-codemirror.js`. Is it maybe a jshint false error?

eduardoj:
@vpereira, @Ana06, gif added. Please have a look again.

PR #7524, opened 2019-05-02
refactor project#index controller

vpereira:
To improve code quality and remove "hidden features" I'm pushing in this PR some changes:

* Remove the "special view" just for bots: that's not the way we make the bots happy. Using sitemaps is the right way. 

* Move variables to methods when is possible to simplify the logic and reduce visual polution in the actions 

* Remove unecessary instance variables

* Revisit some special logic that made sense in the paste, but it's not necessary anymore.


coolo:
but this way most spiders won't find packages

vpereira:
> but this way most spiders won't find packages

But the sitemaps are there for that right? I mean, there is not consistency or spider support to all endpoints..

hennevogel:
> but this way most spiders won't find packages

Why?

hennevogel:
And you do know that there is severe index punishment for spider specific pages right?

src/api/app/controllers/webui2/project_controller.rb
```diff
@@ -4,13 +4,9 @@ def webui2_index
     atype = AttribType.find_by_namespace_and_name!('OBS', 'VeryImportantProject')
     @important_projects = Project.find_by_attribute_type(atype).where('name <> ?', 'deleted').pluck(:name, :title)
 
-    if @spider_bot
-      render :list_simple
```
hennevogel: Are you also going to remove this action and views etc?

vpereira: everything, yes

src/api/app/controllers/webui2/project_controller.rb
```diff
@@ -27,6 +23,18 @@ def webui2_subprojects
 
   private
 
+  def show_all?
+    (params[:all].to_s == 'true')
+  end
+
+  def active_very_important_projects
+    Project.find_by_attribute_type(very_important_project_attribute).where('name <> ?', 'deleted')
```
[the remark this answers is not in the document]
vpereira: done, thank you @coolo

coolo:
> I find this 'move variables to functions' a little over the top, but I won't stand in the way

Especially the attribute that is only used in a one liner function itself doesn't really deserve a function.

vpereira:
> I find this 'move variables to functions' a little over the top, but I won't stand in the way

hey, It was the best way to have code isolated and being called just when it's necessary. As it was before, there some code that was just necessary for html responses, but were running always.

vpereira:
> Especially the attribute that is only used in a one liner function itself doesn't really deserve a function.

Why do you think that a variable is better than a method? Method can be tested, stubbed while testing, you can give better name, etc..

coolo:
but the very_important_project provides a functionality in itself - the attribute is not.

PR #7529, opened 2019-05-05
Replace release_targets_ng method with SQL queries

bgeuken:
Instead of (sql) querying the data first and then using hashes to map and evaluate the data as needed we now have two separate queries that fetch the incident packages and the release target project names individually.
As a result the code becomes much shorter and more readable.

### DIFF
--- a/src/api/app/helpers/webui/maintenance_incident_helper.rb
+++ b/src/api/app/helpers/webui/maintenance_incident_helper.rb
@@ -58,13 +58,13 @@ module Webui::MaintenanceIncidentHelper
     end
   end
 
-  def packages_cell(incident, release_targets_ng)
-    release_target = release_targets_ng.values.first
-    return if release_target[:packages].blank?
-    first_pkg = release_target[:packages].first
+  def packages_cell(incident)
+    packages = incident.packages_with_release_target.limit(2).pluck(:name)
+    return if packages.empty?
+    first_package = packages.first
     safe_join([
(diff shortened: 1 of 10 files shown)


src/api/app/helpers/webui/maintenance_incident_helper.rb
```diff
@@ -58,13 +58,15 @@ def category_cell(incident, patchinfo)
     end
   end
 
-  def packages_cell(incident, release_targets_ng)
-    release_target = release_targets_ng.values.first
-    return if release_target[:packages].blank?
-    first_pkg = release_target[:packages].first
+  def packages_cell(incident)
+    packages = incident.packages.joins(:flags).where(flags: { flag: :build, status: 'enable', repo: incident.release_targets.
```
coolo: this is a little a lot of AR for a helper IMO. Perhaps move this into a function of its own, either directly in the incident or a service class for incidents.

src/api/app/views/webui/projects/maintenance_incidents/_incident_table_entries.html.erb
```diff
@@ -1,16 +1,18 @@
 <% incidents.each do |incident| %>
     <% index = 0 %>
-    <% release_targets_ng = incident.release_targets_ng() %>
+    <% release_targets = incident.target_repositories.joins(:project).
```
coolo: Why do you need this N+1 query here? `target_repositories` should be prefetched for the full `incidents` array by the controller before going into the view.

coolo:
rebased as the obs build failure looked related

PR #7538, opened 2019-05-09
obsworker: use proxy only if it is enabled

cschneemann:
Use configured proxy only if it is also enabled to be used.


cschneemann:
> That seems to be the wrong logic. libproxy at least checks if PROXY_ENABLED is set to "no". It uses the proxy settings if PROXY_ENABLED is not set or has some different value.

So what should be changed here? It seems libproxy is not used in curl: https://github.com/curl/curl/pull/977

mlschroe:
It should be:
```
if [ "$PROXY_ENABLED" != "no" ]; then
```

cschneemann:
> It should be:
> 
> ```
> if [ "$PROXY_ENABLED" != "no" ]; then
> ```

true...

changed

coolo:
please squash these commits

PR #7550, opened 2019-05-13
Deprecate User.current

coolo:
Highlight the intention of the code accessing the current user. If there should be a user session, use User.session!, if you don't care use User.session - or if you want a fallback user use User.possibly_nobody


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -76,13 +76,13 @@ def modify_review
...
     @req = @bs_request.webui_infos(filelimit: diff_limit, tarlimit: diff_limit, diff_to_superseded: @diff_to_superseded)
-    @is_author = @bs_request.creator == User.current.login
+    @is_author = @bs_request.creator == User.possibly_nobody.login
     @is_target_maintainer = @req['is_target_maintainer']
 
     @my_open_reviews = @req['my_open_reviews']
     @other_open_reviews = @req['other_open_reviews']
-    @can_add_reviews = @bs_request.state.in?([:new, :review]) && (@is_author || @is_target_maintainer || @my_open_reviews.present?) && !User.current.is_nobody?
-    @can_handle_request = @bs_request.state.in?([:new, :review, :declined]) && (@is_target_maintainer || @is_author) && !User.current.is_nobody?
+    @can_add_reviews = @bs_request.state.in?([:new, :review]) && (@is_author || @is_target_maintainer || @my_open_reviews.present?)
```
coolo: in case you wonder: I removed the is_nobody? here as `@is_target_maintainer  || @is_author` kind of give it away it's not nobody

src/api/app/controllers/application_controller.rb
```diff
@@ -40,13 +40,13 @@ def authenticator
   end
 
   def pundit_user
-    return User.current unless User.current.is_nobody?
+    User.session! if User.session
```
Ana06: actually you want (or at least those two are equivalent):

```ruby
User.session
```

src/api/app/controllers/application_controller.rb
```diff
@@ -464,10 +464,10 @@ def set_nobody
   end
 
   def set_influxdb_data
-    anonymous = User.current_login == User::NOBODY_LOGIN
+    anonymous = !User.session
 
     InfluxDB::Rails.current.tags = {
-      beta: !anonymous && User.current.in_beta?,
+      beta: !anonymous && User.possibly_nobody.in_beta?,
```
Ana06: why not just `User.possibly_nobody.in_beta?` and getting rid of `anonymous`? :thinking:

src/api/app/controllers/webui/user_controller.rb
```diff
@@ -116,9 +116,9 @@ def register
       redirect_to controller: :user, action: :index
     else
       session[:login] = opts[:login]
-      User.session = User.find_by_login(session[:login])
-      if User.current.home_project
-        redirect_to project_show_path(User.current.home_project)
+      User.session = User.find_by!(login: session[:login])
```
Ana06: [nitpick] I agree with this change but is not in the correct commit

src/api/app/jobs/project_create_auto_cleanup_requests.rb
```diff
@@ -11,7 +11,7 @@ def perform
     return unless cleanup_days && cleanup_days > 0
 
     # defaults
-    User.session = User.current || User.find_by_login('Admin')
+    User.session = User.find_by!(login: 'Admin') unless User.session
```
Ana06: [nitpick]
```ruby
User.session ||= User.find_by!(login: 'Admin')
```
:bowtie:

src/api/app/jobs/staging_project_copy_job.rb
```diff
@@ -1,7 +1,7 @@
 class StagingProjectCopyJob < ApplicationJob
   def perform(staging_workflow_project_name, original_staging_project_name, staging_project_copy_name, user_id)
     # This is needed as the job depends on the current user and without it, it will failed when performed later
-    User.session = User.current || User.find(user_id)
+    User.session = User.find(user_id) unless User.session
```
Ana06: [nitpick] same here :bowtie:

src/api/app/models/package.rb
```diff
@@ -120,7 +120,7 @@ def self.check_access?(package)
   def self.check_cache(project, package, opts)
     @key = { 'get_by_project_and_name' => 1, :package => package, :opts => opts }
 
-    @key[:user] = User.current.cache_key if User.current
+    @key[:user] = User.session!.cache_key if User.session
```
Ana06: you don't need the `!`:
```Ruby
@key[:user] = User.session.cache_key if User.session
```

src/api/app/controllers/application_controller.rb
coolo: yes, you're right. this is a victim of several rounds of patching around with the API :)

src/api/app/controllers/application_controller.rb
coolo: this was protecting the User.current call - so it could use User.session just as well, but I guess your suggestion makes more sense.

src/api/app/controllers/webui/user_controller.rb
coolo: you mean I should cherry pick in an extra commit? I squashes my original commits as it became a mess rebasing, but I can cherry pick special cases.

src/api/app/jobs/project_create_auto_cleanup_requests.rb
coolo: I think this even was the original code more or less - but as I first had User.session= and only afterwards User.session, this went through phases. There are more such places where ||= is easier to read

src/api/app/controllers/request_controller.rb
```diff
@@ -152,8 +152,12 @@ def request_create
     end
 
     # cache the diff (in the backend)
-    @req.bs_request_actions.each do |a|
-      BsRequestActionWebuiInfosJob.perform_later(a)
+    @req.bs_request_actions.each do |action|
+      # cleanup implicit home branches.
+      # FIXME3.0: remove this, the clients should do this automatically meanwhile
```
dmarcoux: What does `3.0` refer to? OBS release?

src/api/app/controllers/source_controller.rb
```diff
@@ -259,13 +258,13 @@ def show_project_pubkey
...
   def delete_project_pubkey
-    params[:user] = User.current.login
+    params[:user] = User.session!.login
     path = pubkey_path
 
     # check for permissions
     upper_project = @prj.name.gsub(/:[^:]*$/, '')
     while upper_project != @prj.name && upper_project.present?
-      if Project.exists_by_name(upper_project) && User.current.can_modify?(Project.get_by_name(upper_project))
+      if Project.exists_by_name(upper_project) && User.session!.can_modify?(Project.get_by_name(upper_project))
```
dmarcoux: The line 226 in this file is now `User.possibly_nobody.can_modify?(...)`, while here it is `User.session!.can_modify?(...)`. Shouldn't they both use `possibly_nobody`? Also for other lines where we call `.can_modify?`?

src/api/app/controllers/request_controller.rb
coolo: Yeah - you will notice I just moved the comment though

src/api/app/controllers/source_controller.rb
coolo: In source controller all actions require login, so using possibly_nobody is indeed a smell. I reviewed that controller in detail and indeed all were wrong. Thanks for spotting

src/api/app/controllers/source_controller.rb
```diff
@@ -93,7 +93,7 @@ def delete_package
     tpkg = Package.get_by_project_and_name(@target_project_name, @target_package_name,
                                            use_source: false, follow_project_links: false)
 
-    unless User.current.can_modify?(tpkg)
+    unless User.session!.can_modify?(tpkg)
```
vpereira: @coolo,

why you decided for ```session!``` here and not ```possibly_nobody```?

src/api/app/controllers/source_controller.rb
```diff
@@ -133,7 +132,7 @@ def require_package
   end
 
   def verify_can_modify_target_package!
-    return if User.current.can_modify?(@package)
+    return if User.session!.can_modify?(@package)
```
vpereira: @coolo 

why ```session!``` and not ```possibly_nobody```?

src/api/app/controllers/source_attribute_controller.rb
```diff
@@ -43,8 +43,8 @@ def delete
     unless attrib
       raise ActiveRecord::RecordNotFound, "Attribute #{params[:attribute]} does not exist"
     end
-    unless User.current.can_create_attribute_in?(@attribute_container, @at)
-      raise ChangeAttributeNoPermission, "User #{user.login} has no permission to change attribute"
+    unless User.possibly_nobody.can_create_attribute_in?(@attribute_container, @at)
```
vpereira: @coolo 

why ```possibly_nobody``` and not ```session!``` ?

src/api/app/controllers/source_controller.rb
coolo: as I wrote above - the action requires login. So in the end User.session would be the correct form, no need to double check. But I would prefer if this was cleaned up one by one - as @bgeuken and you said earlier, User.session! should be a smell, that indicates a move to pundit, especially in controllers.

src/api/app/controllers/source_controller.rb
coolo: same as above

src/api/app/controllers/source_attribute_controller.rb
coolo: basically the same as above - so for consistency I better change to session! too

vpereira:
> Highlight the intention of the code accessing the current user. If there should be a user session, use User.session!, if you don't care use User.session - or if you want a fallback user use User.possibly_nobody

what I'm missing in this PR is the logic behind your decisions, i.e "in this call I decided to don't care about the session" or "in this method the fallback user is ok".. I'm asking because you changed more or less 80 files, taking the decision case by case or did you follow a general guideline? I.e: "For all writing attempt and audit log, ```session!``` is a MUST." or "every read attempt can fallback to nobody"..   could you please elaborate it to help us to understand how to decide which method should we call in which situation?

coolo:
The rules I tried to follow are:
- any access to .login should be a session! (audit, user actions, branches, home directory)
- any access to policies and .can_* functions should prefer possibly_nobody to simplify the code
- any check if there is a login at all should be User.session
- avoid mixing possibly_nobody and session! to access the same user object

The User.session function did not exist when I did the initial cleanup, nowadays I would write most functions that require login anyway using User.session

src/api/app/controllers/webui/user_controller.rb
vpereira: I think she wants that change as extra commit, since you are not deprecating the ```User.current``` here

src/api/app/controllers/source_controller.rb
vpereira: is it an exception for the reasoning 

"any access to policies and .can_* functions should prefer possibly_nobody to simplify the code"

because the action requires login and therefore we know that at this point of time, there will be a valid user other than nobody.. is that right, @coolo ?

coolo: Yes, so possibly using User.session is the proper thing to do. But in the end all 3 functions return the same user if login is secured otherwise. So it's just a question of a good policy on what to do.

src/api/app/controllers/webui/user_controller.rb
coolo: yeah - clarified offline. It's too much hassle to seperate though, so we stay on nitpick

src/api/app/controllers/request_controller.rb
dmarcoux: I did notice, but it didn't answer my question knowing this. Now I know, thank you.

src/api/app/controllers/person_controller.rb
```diff
@@ -36,12 +36,12 @@ def command
...
     user = User.find_by_login!(params[:login])
 
-    if user.login != User.current.login
+    if user != User.session
       logger.debug "Generating for user from parameter #{user.login}"
       render xml: user.render_axml(User.admin_session?)
     else
-      logger.debug "Generating user info for logged in user #{User.current.login}"
-      render xml: User.current.render_axml(true)
+      logger.debug "Generating user info for logged in user #{User.session.login}"
```
dmarcoux: This should be `User.session!.login`, right?

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -368,9 +366,9 @@ def set_influxdb_data
   end
 
   def set_influxdb_additional_tags
-    anonymous = User.current_login == User::NOBODY_LOGIN
+    anonymous = !User.session
     tags = {
-      beta: !anonymous && User.current.in_beta?,
+      beta: !anonymous && User.possibly_nobody.in_beta?,
```
dmarcoux: A really similar change happened in another file in this PR, but it was done differently:

From:

```
anonymous = User.current_login == User::NOBODY_LOGIN

  beta: !anonymous && User.current.in_beta?,
  anonymous: anonymous,
  interface: :api
}
```

To:

```
  beta: User.possibly_nobody.in_beta?,
  anonymous: !User.session,
  interface: :api
}
```

I suggest to do it the same way.

vpereira: i like that @dmarcoux

vpereira: the ```User.session``` does already the check if the current is nobody @coolo

dmarcoux: I just copy pasted what was done by @coolo, so credits go to him. It's making it easier to understand what I'm referring to.

src/api/app/controllers/person_controller.rb
coolo: I actually changed it to session because the if is already making sure there is a session, but in general you are right

src/api/app/controllers/webui/webui_controller.rb
coolo: Right. @Ana06 pointed out the one case and I missed the other case

coolo:
I force pushed to align the 2 places @dmarcoux pointed out - now awaiting @Ana06's blessing

src/api/app/helpers/webui/user_helper.rb
```diff
@@ -61,7 +61,7 @@ def user_and_role(user, role = nil, options = {})
 
     # It's necessary to concat icon and $variable and don't use string interpolation!
     # Otherwise we get a new string and not an ActiveSupport::SafeBuffer
-    if User.current.is_nobody?
+    if !User.session
```
Ana06: [nitpick] I would flip this `if` so that the condition is not negated

PR #7552, opened 2019-05-13
Align usage of flash messages for groups/users_controller

### LINKED ISSUE #7546: Adding a user to a group even if he/she is already a member ends up in an exception
# Issue Description
When trying to add a user to a group even he/she is already part of that group will throw an exception in the WebUI.

Expected Result
============
A propper error message is shown.

How to Reproduce
===============
1. Create a group
2. Add a user
3. Add the same user again

dmarcoux:
The create action is called synchronously, so we redirect instead of rendering JavaScript. This differs from other actions (update and destroy) since they are called asynchronously. The flash messages now also contain the user/group whenever needed.

Closes #7546

### DIFF
--- a/src/api/spec/controllers/webui/groups/users_controller_spec.rb
+++ b/src/api/spec/controllers/webui/groups/users_controller_spec.rb
@@ -5,13 +5,6 @@ RSpec.describe Webui::Groups::UsersController do
   let(:user) { create(:confirmed_user) }
   let(:admin) { create(:admin_user) }
 
-  RSpec.shared_examples 'response for non existing user or group' do
-    it 'reports an error' do
-      expect(response).to have_http_status(:not_found)
-      expect(flash[:error]).not_to be_nil
-    end
-  end
-
   describe 'POST create' do
     before do
       login(admin)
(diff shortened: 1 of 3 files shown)


src/api/spec/controllers/webui/groups/users_controller_spec.rb
```diff
@@ -5,57 +5,49 @@
...
 
       subject! do
-        post :create, params: { group_title: group.title, user_login: user.login }, format: :js
+        post :create, params: { group_title: group.title, user_login: user.login }
       end
 
       it 'reports the error' do
-        expect(response).to have_http_status(:bad_request)
-        expect(flash[:error]).not_to be_nil
+        expect(flash[:error]).to eq("Couldn't add user '#{user}' to group '#{group}': User User already has this group")
```
dmarcoux: The duplication of `User` in the second part of the message is coming from the controller when we call `errors.full_messages.to_sentence` on a `GroupUser` instance. See the output below, especially `@messages`. Since there's the `:user` key, it's added to the message which already contains `User`.

```
 @base=#<GroupsUser:0x00007f872811b050 group_id: 1, user_id: 5, created_at: nil, email: true, id: nil>,
 @details={:user=>[{:error=>"User already has this group"}]},
 @messages={:user=>["User already has this group"]}>
```

dmarcoux:
[Address Feedback](https://trello.com/c/KybHivS4/1686-address-feedback)

PR #7557, opened 2019-05-14
Validate patchinfo summary, not description, in both webui and XML

saraycp:
According to the usage and the feedback coming from users, it seems that
only the summary field is compulsory. So we ensure summary is present
from webui and XML.


hennevogel:
@saraycp what's the status here?

hennevogel:
guess we close this @saraycp?

PR #7486, opened 2019-04-26
Reduce webui_infos usage in request controller

coolo:
Move the important parts of webui_info into the controller, most work is still in webui_actions for now though

PR #7560, opened 2019-05-14
Display review box only for requests which are new or in review

### LINKED ISSUE #7442: Request in superseeded state shouldn't have approve button
It is not possible to change the state of a request which is not in review state (for example superseded). In those case there shouldn't be an approve button or it should be disabled:

Related to:

image

Related to https://github.com/openSUSE/open-build-service/issues/7263

dmarcoux:
Closes #7442

Try it in the review app by submitting a request to the project `home:Iggy` for the package `ruby` by superseding the existing request (Do it [here](https://obs-reviewlab.opensuse.org/dmarcoux-issue-7442/package/show/home:Admin/ruby)). Then add a review for the current user (I suggest using `Admin`). You should see the review tab, but as soon as you supersede the request, the review form won't be visible anymore.

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -96,7 +96,6 @@ class Webui::RequestController < Webui::WebuiController
     @is_author = @bs_request.creator == User.possibly_nobody.login
 
     @is_target_maintainer = @bs_request.is_target_maintainer?(User.session)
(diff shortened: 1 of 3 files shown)


dmarcoux:

coolo:
I disagree (as it also conflicts with what I did in #7486) - the view should just not show this box, the reviews are still valid from the model's point of view.

dmarcoux:
I pushed another set of changes. This time, it's displaying the review box only for requests which are new or in review.

PR #7569, opened 2019-05-15
Add skeleton for search views in Bootstrap

dmarcoux:
This is to prevent us from blocking each other while we work on search views.

### DIFF
--- a/src/api/app/views/webui2/webui/search/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui2/webui/search/_breadcrumb_items.html.haml
@@ -1,0 +1,8 @@
+= render partial: 'webui/main/breadcrumb_items'
+
+%li.breadcrumb-item.active{ 'aria-current' => 'page' }
+  - case action_name
+  - when 'index'
+    Search
+  - when 'owner'
+    Search Owners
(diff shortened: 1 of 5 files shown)


dmarcoux:
The specs will obviously fail if this is available for the test environment, so I will make it for the development environment only.

src/api/app/views/webui2/webui/search/_breadcrumb_items.html.haml
```diff
@@ -0,0 +1,9 @@
+= render partial: 'webui/main/breadcrumb_items'
+
+- case action_name
+- when 'index'
+  %li.breadcrumb-item.active{ 'aria-current' => 'page' }
```
eduardoj: This line, that is the same as line 8, can be moved before the `case` statement. But this can be done later on.

dmarcoux: Good point, I did it now.

src/api/app/views/webui2/webui/search/_breadcrumb_items.html.haml
```diff
@@ -0,0 +1,8 @@
+= render partial: 'webui/main/breadcrumb_items'
+
+%li.breadcrumb-item.active{ 'aria-current' => 'page' }
+  - case action_name
```
dmarcoux: This could be simplified with a `if`, because the only different part is `Owners`. However, this is just too much in my opinion. So I keep it as it is.

PR #7573, opened 2019-05-16
Migrate Image Template to Bootstrap

DavidKang:
We migrate Image Template view to Bootstrap

#### Before
Screenshot_2019-05-16 Open Build Service(1)

#### After
Screenshot_2019-05-16 Open Build Service(3)

### DIFF
--- a/src/api/app/assets/javascripts/webui2/application.js
+++ b/src/api/app/assets/javascripts/webui2/application.js
@@ -48,4 +48,5 @@
 //= require webui2/project_monitor.js
 //= require webui2/collapsible_text
 //= require webui2/patchinfo.js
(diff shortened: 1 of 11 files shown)


hennevogel:
why the old icons?

DavidKang:
Because they are coming from a file

bgeuken:
There seems to be a problem when the appliance icons are not available:



It looks similar for the old UI, except that the text in the icon field does not overflow. That's why I didn't block the PR. But it would be nice if we could catch the not found errors and render a default image instead.
IIRC we do something like that for the comments already.

DavidKang:
@bgeuken, could you review it again :smile_cat:

PR #7576, opened 2019-05-16
Increase contrast on default text

### LINKED ISSUE #7254: [beta] webui: increase contrast on default text
Currently, the beta webui, by way of bootstrap, declares a color `#212529` that is used for the body, and color `#495057` for e.g. `<textarea>`, and these choices suffer from the problems previously fixed in #4608. Please change all occurrences of these two strings to `#000000`.

eduardoj:
The colors used by Bootstrap for the generic text (text-body) and for form controls are not #000000. Changing it to #000000 the contrast on default text is increased.

Fixes #7254.

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/bootstrap_variables/colors.scss
+++ b/src/api/app/assets/stylesheets/webui2/bootstrap_variables/colors.scss
@@ -1,4 +1,5 @@
 // Override Bootstrap's default colors
+$body-color: #000000;
 $body-bg: #f6f6f6;
 $gray-100: #f8f9fa;
 $gray-200: #f0f0f0;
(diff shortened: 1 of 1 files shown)


eduardoj:

coolo:
it's #24292e in github for some reason as well. sounds like UI designers in general dislike black (at least now, I remember different days :)

dmarcoux:
I would not merge this. Read this article: https://ianstormtaylor.com/design-tip-never-use-black/

eduardoj:
Closing, as it will not be the proper solution to address the issue, even if the issue is still valid.

eduardoj:
@pdostal, please can you have a look.

PR #7581, opened 2019-05-17
Make sure the packages factory sets a user for backend log

coolo:
Now let the real fun start with user handling - everything else was just prep work :)

I put the exception outside the global_write_through so it's checking not just in real life, but also in the test suite

### DIFF
--- a/src/api/spec/controllers/worker/command_controller_spec.rb
+++ b/src/api/spec/controllers/worker/command_controller_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe Worker::CommandController do
   let!(:project) do
     create(:project_with_repository, maintainer: user) do |project|
       project.store
-      project.packages.create(attributes_for(:package))
+      create(:package, project: project)
     end
   end
 
--- a/src/api/spec/factories/packages.rb
+++ b/src/api/spec/factories/packages.rb
@@ -5,7 +5,9 @@ FactoryBot.define do
     title { Faker::Book.title }
     description { Faker::Lorem.sentence }
 
-    after(:create, &:write_to_backend)
+    after(:build) do |package|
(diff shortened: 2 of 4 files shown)


coolo:
This is just a first draft - the packages model uses the User.session in multiple places and we could move all of them out now and rely on controllers/factories. So the actions creating/deleting a package would check the policies and then set the user and the model gets dumbed down.

coolo:
And of course the same applies to the project model, but first let's agree on a way to go.

src/api/spec/factories/packages.rb
```diff
@@ -5,7 +5,9 @@
     title { Faker::Book.title }
     description { Faker::Lorem.sentence }
 
-    after(:create, &:write_to_backend)
```
coolo: btw: henne, in 245dfbf7b97 you added this, but the model's after_save hook already makes sure this is called. So it's not needed and just repeats the function (which obviously is meaningless in the VCR case). I just spotted it while adding the after(:build) hook

src/api/spec/controllers/worker/command_controller_spec.rb
```diff
@@ -15,7 +15,9 @@
   let!(:project) do
     create(:project_with_repository, maintainer: user) do |project|
       project.store
-      project.packages.create(attributes_for(:package))
+      package = project.packages.build(attributes_for(:package))
+      package.commit_user = user
+      package.save
```
hennevogel: Please make use of the package factory

src/api/spec/factories/packages.rb
hennevogel: If you say so :cowboy_hat_face:

coolo:
@hennevogel ping

PR #7583, opened 2019-05-17
rspec: Avoid home projects for users created in factories

coolo:


src/api/spec/factories/users.rb
```diff
@@ -5,10 +5,29 @@
...
     end
 
+    trait :with_home do
+      avoid_home_project { false }
+    end
+
+    to_create do |user, evaluator|
+      before = Configuration.allow_user_to_create_home_project
+      if evaluator.avoid_home_project
+        Configuration.update(allow_user_to_create_home_project: false)
```
hennevogel: Why don't you make this default in the suite setup, instead of changing it here all the time?

coolo: You can still have 2 users in the same test - one with and one without home.

hennevogel: Sure and the one `:with_home` can change the default during creation, no?

vpereira:
@coolo  can you rebase with master and fix it here too? https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/controllers/trigger_controller_spec.rb

coolo:
@hennevogel ping

hennevogel:
@coolo sorry, a bit busy with other stuff. maybe later today

hennevogel:
@coolo[bot] rebase

coolo:
I can rebase!

Issue #7564, opened 2019-05-15
[production][image_templates#index] You cannot call create unless the parent is saved


bgeuken:
image_templates#index threw this 46 times
<pre>
You cannot call create unless the parent is saved
</pre>
[https://build.opensuse.org/image_templates](https://build.opensuse.org/image_templates)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5cd07c6a6e12e0000d300d73)

```ruby

/GEM_ROOT/gems/activerecord-5.2.3/lib/active_record/associations/collection_association.rb:352→ _create_record
--
/GEM_ROOT/gems/activerecord-5.2.3/lib/active_record/associations/has_many_association.rb:124→ _create_record
/GEM_ROOT/gems/activerecord-5.2.3/lib/active_record/associations/association.rb:190→ create
/GEM_ROOT/gems/activerecord-5.2.3/lib/active_record/associations/collection_proxy.rb:350→ create
/PROJECT_ROOT/app/models/user.rb:125→ create_home_project



```

bgeuken:
The template page is accessible again. So I'll remove the p2 label

coolo:
where does this come from? Hopefully not from the User.current changes?

dmarcoux:
> where does this come from? Hopefully not from the User.current changes?

No because they aren't deployed yet. It happened during our planning meeting. I tried to access the page while being logged-in with my user and it was stuck loading.

coolo:
but the exception in errbit happened the last time 17 hours ago - so it's not the same problem

bgeuken:
The reason for the failure is that the home projects base on the user's `login` but the validations for both differ. User names that start with `.`, `:` or `_` cause the home project creation to crash.

We could tighten the user login validation, but that would log out LDAP users with a name that matches.
In build.o.o we have 12 users with a name that starts with `_`. Though their last login is years ago.
The other option would be to add a valid character to the home project name, if needed. Disadvantage here would be that the home project name would become a bit less predictable and we might have to adapt some code.

PR #7586, opened 2019-05-17
Fix user home creation for users with certain login names

bgeuken:
It's possible that user login names start with '.', ':' or '_'. This
causes a problem when creating home projects, because they can not
contain a ':' followed by one of aboves characters, like 'home:.tux'
in case of user login '.tux'.

On the other hand we support authentication via LDAP, Kerberos and alike
and we have some older user records start with a '_' in some OBS
instances.
Therefor we keep above user records valid and add a prefix for home
projects of such users, e.g. 'home:project.tux'.

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -374,7 +374,11 @@ class User < ApplicationRecord
   end
 
   def home_project_name
-    "home:#{login}"
+    if login.starts_with?('.', ':', '_')
+      "home:project#{login}"
+    else
+      "home:#{login}"
+    end
   end
 
   def home_project
--- a/ReleaseNotes-2.10
+++ b/ReleaseNotes-2.10
@@ -59,7 +59,9 @@ Bugfixes:
   * In WebUI, only admins are allowed to create sourceaccess/access repositories flags.
   * Added missing authorization to move repository path in Webui::ProjectController.
   * Require sourceaccess by default in `require_package`.
-
+  * Users with a login that started with '.', ':' or '_' would get an error when creating
+    their home project. This got solved, though such user home projects are now of the format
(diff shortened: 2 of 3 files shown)


src/api/app/models/user.rb
```diff
@@ -374,7 +374,11 @@ def render_axml(watchlist = false)
   end
 
   def home_project_name
-    "home:#{login}"
+    if login.starts_with?('.', ':', '_')
+      "home:project#{login}"
```
bgeuken: If anyone knows a better, valid char (@ doesn't work) or a better solution in general, please speak up:smiley:

coolo: I wouldn't waste a lot of thought on this - as long it's possible at all. But a : in the middle of the project name is a problem in itself - as as `home:project:bgeuken` would be subproject of `home:project` - just as a login of `bgeuken:mudder` would result in `home:bgeuken:mudder` and you got access in it.

As such using `home:user_#{id}` if login is not resulting in a valid project name might be an alternative

hennevogel:
I fail to understand why we workaround this at all?

bgeuken:
Because in case of authentication via LDAP, Kerberos and alike we can not control the format of user logins. So we can either disallow such user login names or we use a workaround like this, see [issue 7564](https://github.com/openSUSE/open-build-service/issues/7564#issuecomment-493044731).

But given that you closed the issue I assume we go with disallowing these logins and document this?

hennevogel:
> But given that you closed the issue I assume we go with disallowing these logins and document this?

Yes, just because it's possible with LDAP doesn't mean we can support it.

bgeuken:
Alright, closing

PR #7587, opened 2019-05-17
Fix image template description for remote templates

bgeuken:
Since we use XMLHash to parse image templates of remote instances,
we get a string like '{}' when such an template description is empty.

This commit ensures that we don't show such strings in the UI.

Before:

After:


src/api/app/views/webui/image_templates/index.html.haml
```diff
@@ -23,7 +23,9 @@
                 %span.name.break-words
                   = link_to_if(!remote, title_or_name(template), package_show_path(project: project, package: template))
                 %span.description.grey.break-words
-                  = template.description
+                  -# Remote templates with empty description
+                  - unless template.description == '{}'
```
hennevogel: Hm, shouldn't you catch this wherever you calculate the string?

bgeuken: Fair enough :thumbsup:

Issue #7253, opened 2019-03-21
[beta] webui: use standard monospace font


jengelh:
I just joined the beta webui programme on build.opensuse.org, and now find what coolo had already hinted at in https://github.com/openSUSE/open-build-service/pull/6599#issuecomment-446847192 that the beta webui is using a second set of CSS files that also need adjustment, that is, specifying "font-family: monospace", overrding SFMono and so on.

hennevogel:
![](https://github.trello.services/images/mini-trello-icon.png) [Address Feedback](https://trello.com/c/KybHivS4/1686-address-feedback)

dmarcoux:
Since the PR fixing this was reverted, I'm re-opening this issue.

dmarcoux:
Closing this after discussing with @eduardoj. He created a new issue to better reflect what we aim to achieve regarding fonts. This will include a monospace font.

jengelh:
But I don't want any (arbitrary) monospace font someone else chose, I want the page to use the preference set forth in the browser settings.

coolo:
look for other means to achieve it, e.g. by following
https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use#w_custom-fonts

Just because you have proper fonts setup, doesn't mean websites like OBS can rely on it to be fine in general.

PR #7589, opened 2019-05-17
Revert "Override Bootstrap monospace fonts with 'monospace'"

eduardoj:
Reverts openSUSE/open-build-service#7570.

Related to issue openSUSE/open-build-service#7253.

We prefer to define a fixed group of fonts than to rely on the font configured in the browser. This way the aspect of the page will be more homogeneous.

The user that wants to override the font defined in the web page can also configure the browser accordingly. For users of firefox see [this page](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use#w_custom-fonts).


hennevogel:
The real fix is to come up with a set of fonts that we want to use and also deliver them...

eduardoj:

PR #7597, opened 2019-05-19
features/webui/repositories_spec: Stub request to opensuse.org

coolo:


src/api/spec/bootstrap/features/webui/repositories_spec.rb
```diff
@@ -14,6 +14,11 @@
   describe 'Repositories' do
     before do
       login admin_user
+
+      fake_distribution_body = File.open(Rails.root.join('test/fixtures/backend/distributions.xml')).read
+
+      stub_request(:get, 'https://api.opensuse.org/public/distributions.xml').
+        to_return(status: 200, body: fake_distribution_body, headers: {})
```
bgeuken: This change would mean that we can drop, or shorten, a couple of cassettes, or not?

coolo: 2:

```
spec/cassettes/Bootstrap_Repositories/Repositories/add/delete_repository_from_distribution.yml:    uri: https://api.opensuse.org/public/distributions.xml
spec/cassettes/Bootstrap_Repositories/Repositories/add_repository_from_distribution.yml:    uri: https://api.opensuse.org/public/distributions.xml

bgeuken: Then why don't you update them in the same commit?

coolo: Well, because killing them all is my target - but if you wish, I can arrange it

coolo: removing them is enough actually

PR #7599, opened 2019-05-19
Make the staging_project#copy action a little more secure

coolo:
I disabled the code in production to avoid abuses of this rather big hole

### DIFF
--- a/src/api/app/controllers/staging/staging_projects_controller.rb
+++ b/src/api/app/controllers/staging/staging_projects_controller.rb
@@ -36,6 +36,9 @@ class Staging::StagingProjectsController < ApplicationController
   def copy
     authorize @main_project.staging
 
+    project = Project.where(name: params[:staging_project_copy_name]).first_or_initialize
+    authorize project, :create_new?
+
     StagingProjectCopyJob.perform_later(params[:staging_workflow_project], params[:staging_project_name], params[:staging_project_copy_name], User.session!.id)
     render_ok
   end
--- a/src/api/app/policies/project_policy.rb
+++ b/src/api/app/policies/project_policy.rb
@@ -39,6 +39,10 @@ class ProjectPolicy < ApplicationPolicy
     User.admin_session? || !record.disabled_for?('sourceaccess', nil, nil)
   end
 
+  def create_new?
+    record.new_record? && create?
+  end
+
   private
 
   def no_remote_instance_defined_and_has_not_remote_repositories?
--- a/src/api/spec/controllers/staging/staging_projects_controller_spec.rb
+++ b/src/api/spec/controllers/staging/staging_projects_controller_spec.rb
@@ -115,6 +115,7 @@ RSpec.describe Staging::StagingProjectsController do
     let(:staging_workflow_project) { staging_workflow.project.name }
     let(:original_staging_project_name) { staging_workflow.staging_projects.first.name }
     let(:staging_project_copy_name) { "#{original_staging_project_name}-copy" }
+    let(:existingproject) { create(:project) }
     let(:params) do
       {
         staging_workflow_project: staging_workflow_project,
@@ -137,6 +138,14 @@ RSpec.describe Staging::StagingProjectsController do
                                                                                                            original_staging_project_name,
                                                                                                            staging_project_copy_name,
                                                                                                            user.id)
+    end
+
+    it 'does not allow unauthoried access' do
+      post :copy, format: :xml, params: params.merge(staging_project_copy_name: 'newtoplevelproject')
+      expect(response).to have_http_status(:forbidden)
+
+      post :copy, format: :xml, params: params.merge(staging_project_copy_name: existingproject.name)
+      expect(response).to have_http_status(:forbidden)
     end
   end
 


bgeuken:
Good idea to catch this upfront before queuing it via DJ.

Though there doesn't seem to be a security issue. When copying to a location that already exists, this would fail with an "Name has already been taken" error.
And when copying to a project location a user doesn't have permission for, this would fail with:
"No permission to modify project 'newtoplevelproject' for user 'permitted_user'".

If you have instructions to expose this, please send it via mail or tell us tomorrow in the office:-)

coolo:
Which just proves my point that *relying* on policies in controllers is dangerous. 

But I have bad news for you. Remember https://github.com/openSUSE/open-build-service/pull/7492#discussion_r279208640 ? Now calc 1+1

src/api/app/policies/project_policy.rb
```diff
@@ -39,6 +39,10 @@ def source_access?
     User.admin_session? || !record.disabled_for?('sourceaccess', nil, nil)
   end
 
+  def create_new?
+    record.new_record? && create?
```
dmarcoux: I believe this should go in `create?`. What is the reason for having `create_new?`?

coolo: hmm, fair question. create is create not update. But other controllers use Project.new, which always makes this true, so this needs to be looked at as well.

coolo:
please fix as needed. don't wait for me

bgeuken:
I'll take care of it

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
```diff
@@ -138,6 +139,14 @@
                                                                                                            staging_project_copy_name,
                                                                                                            user.id)
     end
+
+    it 'does not allow unauthoried access' do
```
hennevogel: Please do a policy test instead. You then can make sure the policy is applied in this controller with [`after_action :verify_authorized`](https://github.com/varvet/pundit#ensuring-policies-and-scopes-are-used) and if you really must, test here that this [`after_action`](http://matchers.shoulda.io/docs/v3.1.1/Shoulda/Matchers/ActionController.html#use_after_action-instance_method) is called.

coolo: I find this really questionable. Because the policy did its job and the after_action did its job. Still the developer didn't realize *which* policies to apply. And there a test case can help the review process.

src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -36,6 +36,18 @@ def create
   def copy
     authorize @main_project.staging
 
+    @project = Project.where(name: params[:staging_project_copy_name]).first_or_initialize
+    authorize @project, :create?
+
+    unless @project.new_record?
```
coolo: but wasn't @dmarcoux 's point that this is true for all creates?

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
```diff
@@ -138,6 +139,16 @@
                                                                                                            staging_project_copy_name,
                                                                                                            user.id)
     end
+
+    it 'does not allow unauthorized access' do
+      post :copy, format: :xml, params: params.merge(staging_project_copy_name: 'newtoplevelproject')
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it 'does not allow to copying into existing projects' do
```
coolo: this is strange grammar

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
hennevogel: > Still the developer didn't realize which policies to apply.

Can't save yourself from stupidity (you end up with what we have right now). This should have been catched during code-review, it has not been. The solution is to (help you) to review more carefully.

src/api/app/controllers/staging/staging_projects_controller.rb
bgeuken: It's true for all creations of staging projects. It's not like anyone would be able / authorized to do that.

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
bgeuken: As long as there are two authorization calls in the controller `verify_authorized` won't help much. The solution seems to be to have a staging project policy.

bgeuken: Or we create a project policy for the `copy` action. :thinking:

hennevogel:
> Which just proves my point that relying on policies in controllers is dangerous.

This has nothing to do with where you check permissions. If you do it wrongly or not at all you will get into trouble...

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
hennevogel: > As long as there are two authorization calls in the controller verify_authorized won't help much.

Like I said, you can't technically guard yourself from everything.

> The solution seems to be to have a staging project policy.

True, the check in the second `authorize` should move in there :+1:

bgeuken: > Like I said, you can't technically guard yourself from everything.

But in this special case (having two authorize calls in one action) it seems to make sense to have 

> True, the check in the second authorize should move in there :+1:

I would say that either we move the whole authorization code there and only authorize it once or we live with the duplicated call for now (and use the same project policy).

I would go with the latter for now. Given that staging projects are currently no model of their own but a module added to a project.

src/api/app/controllers/staging/staging_projects_controller.rb
dmarcoux: > but wasn't @dmarcoux 's point that this is true for all creates?

Yes, it is. I would put this in the policy. It is not specific to the creation of a staging project. If someone tries to create a project which already exists, he/she shouldn't be authorized to do so.

bgeuken: > Edit: Actually, this is a validation. It doesn't belong in the policy.

Exactly:-)

dmarcoux: Could we move this validation in the job creation? So it fails whenever trying to create a copy job with a project which already exists?

coolo:
> > Which just proves my point that relying on policies in controllers is dangerous.
> 
> This has nothing to do with where you check permissions. If you do it wrongly or not at all you will get into trouble...

The more important is to duplicate the safety net - either by having a 2nd line of defense in the model (questionable to get/keep sanely) or have proper controller tests that verify the expectations of the developer. But pundit/policies are no magic bullet.

src/api/app/controllers/staging/staging_projects_controller.rb
bgeuken: You mean like:

```ruby
  project = Project.where(name: params[:staging_project_copy_name]).first_or_initialize
  raise SomeException, project.errors.full_messages.to_sentence unless project.valid?
```

and then catch it in the UI? Or do you have a service in mind?

hennevogel:
>>> Which just proves my point that relying on policies in controllers is dangerous.
>>>
>> This has nothing to do with where you check permissions. If you do it wrongly or not at all you will get into trouble...
>>
> The more important is to duplicate the safety net

Why?

> either by having a 2nd line of defense in the model (questionable to get/keep sanely)

We already have this and it's killing us because it's too hard to foresee the consequences. Models calling other Models, passing on sessions, exceptions in the control flow etc. etc.

> or have proper controller tests that verify the expectations of the developer. But pundit/policies are no magic bullet. 

Nothing is a magic bullet. But 

1. put all the "who can do what" code into a policy
2. test the :hankey: out of the policy
3. make as sure as you can that policies are applied in controllers

is a sane way forward. And if we have to duplicate code or call other policies, so be it. Clarity trumps in this area.

src/api/app/controllers/staging/staging_projects_controller.rb
dmarcoux: I wouldn't use a service for this.

Perhaps a [callback](https://guides.rubyonrails.org/v5.2/active_job_basics.html#available-callbacks), so it doesn't even reach the perform. Raising an exception in there and catching it in the controller should be fine.

vpereira: Service is the way to avoid leaking authentication and authorization to other layers. That's how the best practices say to do so.

bgeuken: We already use services for similar cases in the staging project controller, like the `create` action. So I am leaning towards services, if I would have to pick one of the two.

Though I was a bit surprised to find pundit policy calls in the `StagingProjectValidator` service.

bgeuken:
I've extracted the security patch to #7617 so this isn't blocked by the other changes.

coolo:
rebased - IMO it's good to go.

src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -34,7 +36,17 @@ def create
   end
 
   def copy
-    authorize @main_project.staging
+    @project = Project.where(name: params[:staging_project_copy_name]).first_or_initialize
```
vpereira: maybe move this ```@project``` assignment to a `before_action`

coolo: this shouldn't even have a `@` IMO - so I could move the whole authorize into a before filter. but taking that it's only 2 lines more I wonder if it's worth it

vpereira:
@coolo would you mind to make it green again? 

@hennevogel are your change requests already addressed? Could you please review it again if necessary?

coolo:
not this week

coolo:
rebased again - green locally

hennevogel:
> rebased again - green locally

Are you going to address the feedback?

coolo:
No. You can add other tests, but I won't remove any

hennevogel:
> No. You can add other tests, but I won't remove any

Okay I guess someone (@openSUSE/obs-frontend) has to help @coolo out with policies...

vpereira:
@coolo would be ok if I open a new PR with this code and then we start a new discussion there ?

coolo:
Go ahead - most of what happened here has been cherry picked into other PRs anyway :)

PR #7602, opened 2019-05-20
Introduce User.run_as to avoid leaking sessions

coolo:


src/api/app/models/user.rb
```diff
@@ -865,6 +865,16 @@ def send_metric_for_beta_change
     RabbitmqBus.send_to_bus('metrics', "user.#{channel} value=1")
   end
 
+  def run_as
```
hennevogel: Nice idea to `yield` but this needs to be unit tested.

src/api/app/jobs/consistency_check_job.rb
```diff
@@ -74,7 +74,7 @@ def fix
   end
 
   def init
-    User.session ||= User.get_default_admin
+    User.session = User.get_default_admin
```
hennevogel: What does this have to do with `User.run_as`?

src/api/app/models/project.rb
```diff
@@ -1075,23 +1075,22 @@ def repositories_linking_project(tproj)
 
   # called either directly or from delayed job
   def do_project_copy(params)
-    # set user if nil, needed for delayed job in Package model
-    User.session ||= User.find_by!(login: params[:user])
-
-    check_write_access!
+    User.find_by!(login: params[:user]).run_as do
```
hennevogel: :thought_balloon: I don't like that this is happening here, it should be in the controller IMHO. I realize that you are just refactoring it...

hennevogel:
In general again: Please split commits into feature + adoption + other fixes.

src/api/app/jobs/consistency_check_job.rb
coolo: well, not so much - more with leaking sessions :)

src/api/app/models/user.rb
coolo: I added some, if you have other cases in mind that need testing, let me know

src/api/app/jobs/consistency_check_job.rb
```diff
@@ -4,8 +4,11 @@
 class ConsistencyCheckJob < ApplicationJob
   queue_as :consistency_check
 
-  def perform(fix = nil)
-    init
+  def perform
+    User.get_default_admin.run_as { _perform(nil) }
```
hennevogel: :thought_balloon: I don't like this. I mean I get what you are doing but can't you just wrap everything in `perform` into a `.run_as`?

hennevogel:
You're also aware of `lib/tasks/delayed_job.rake` right? Just checking...

src/api/app/jobs/consistency_check_job.rb
hennevogel: Alright we talked offline and going to refactor this later...

PR #7605, opened 2019-05-20
send package status to a redis server if configured

mlschroe:


vpereira:
hi @mlschroe does it need any change in the UI, or can we take advantage of that in the frontend?

hennevogel:
Sounds like an awesome opportunity to start documentation! :smile:

mlschroe:
Not so fast, lets first see if it works or not.

coolo:
shall we add a redis container to the ci?

PR #7608, opened 2019-05-20
 Move do_project_copy to the active job completely

coolo:
There is no reason to implement the job in the model just to differ perform_now and perform_later. This simplifies the setup and makes the permission check unique

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -685,15 +685,15 @@ class SourceController < ApplicationController
       end
     end
 
+    job_params = params.slice(
+      :cmd, :user, :comment, :oproject, :withbinaries, :withhistory, :makeolder, :makeoriginolder, :noservice
+    ).permit!.to_h
+    job_params[:user] = User.session!.login
+
     if params.key?(:nodelay)
-      @project.do_project_copy(params)
+      ProjectDoProjectCopyJob.perform_now(@project.id, job_params)
       render_ok
     else
-      job_params =
-        params.slice(
-          :cmd, :user, :comment, :oproject, :withbinaries, :withhistory, :makeolder, :makeoriginolder, :noservice
-        ).permit!.to_h
-
       ProjectDoProjectCopyJob.perform_later(@project.id, job_params)
       render_invoked
     end
--- a/src/api/app/jobs/project_do_project_copy_job.rb
+++ b/src/api/app/jobs/project_do_project_copy_job.rb
@@ -1,7 +1,41 @@
 class ProjectDoProjectCopyJob < ApplicationJob
   queue_as :quick
 
+  attr_accessor :project, :params
   def perform(project_id, params)
-    Project.find(project_id).do_project_copy(params)
+    self.project = Project.find(project_id)
+    self.params = params
+    do_project_copy
+  end
+
+  private
+
+  def do_project_copy
+    User.find_by!(login: params[:user]).run_as do
+      # copy entire project in the backend
+      begin
+        path = "/source/#{CGI.escape(project.name)}"
+        path << Backend::Connection.build_query_from_hash(params,
+                                                          [:cmd, :user, :comment, :oproject, :withbinaries, :withhistory,
+                                                           :makeolder, :makeoriginolder, :noservice])
+        Backend::Connection.post path
+      rescue Backend::Error => e
+        logger.debug "copy failed: #{e.summary}"
+        # we need to check results of backend in any case (also timeout error eg)
+      end
+      _update_backend_packages
+    end
+  end
+
+  def _update_backend_packages
+    # restore all package meta data objects in DB
+    backend_pkgs = Xmlhash.parse(Backend::Api::Search.packages_for_project(project.name))
+    backend_pkgs.elements('package') do |package|
+      pkg_name = package['name']
+      pkg = project.packages.where(name: pkg_name).first_or_initialize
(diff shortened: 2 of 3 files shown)


coolo:
this needs to be rebased once #7602 is merged - it's a followup on @hennevogel's remarks.

src/api/app/jobs/project_do_project_copy_job.rb
```diff
@@ -1,7 +1,41 @@
...
+  end
+
+  def _update_backend_packages
+    # restore all package meta data objects in DB
+    backend_pkgs = Xmlhash.parse(Backend::Api::Search.packages_for_project(project.name))
+    backend_pkgs.elements('package') do |package|
+      pname = package['name']
+      p = project.packages.where(name: pname).first_or_initialize
+      p.update_from_xml(Xmlhash.parse(Backend::Api::Sources::Package.meta(project.name, pname)))
+      p.save! # do not store
```
coolo: I just moved this code - but the comment is a little strange if you know
https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/package.rb#L68

hennevogel: Guess it's older than the callback :-)

coolo: the callback is from 2012 - at least :)

vpereira: any possibility to have a spec for that?

coolo: Not with VCR

vpereira: don't get it. You can mock all backend calls, since we are not testing the backend, but a job class.

coolo: Possibly. But this is tested in the minitest suite

coolo: (as you can see with in the current failures :)

src/api/app/jobs/project_do_project_copy_job.rb
```diff
@@ -1,7 +1,41 @@
...
+      _update_backend_packages
+    end
+  end
+
+  def _update_backend_packages
+    # restore all package meta data objects in DB
+    backend_pkgs = Xmlhash.parse(Backend::Api::Search.packages_for_project(project.name))
+    backend_pkgs.elements('package') do |package|
+      pname = package['name']
+      p = project.packages.where(name: pname).first_or_initialize
```
vpereira: do we have to use `p` here?

src/api/app/jobs/project_do_project_copy_job.rb
```diff
@@ -1,7 +1,41 @@
...
+    do_project_copy
+  end
+
+  private
+
+  def do_project_copy
+    User.find(params[:user]).run_as do
+      # copy entire project in the backend
+      begin
+        path = "/source/#{CGI.escape(project.name)}"
```
vpereira: what do you think about using a helper to build the url? `File.join` for example would be an option, no?

src/api/app/controllers/source_controller.rb
```diff
@@ -685,15 +685,15 @@ def project_command_copy
       end
     end
 
+    job_params = params.slice(
+      :cmd, :user, :comment, :oproject, :withbinaries, :withhistory, :makeolder, :makeoriginolder, :noservice
+    ).permit!.to_h
+    job_params[:user] = User.session.id
```
vpereira: would be possible to have the `nobody id` here?

vpereira:
@coolo are you still working on that? I see that there is a broken test 

@hennevogel do you want to review it again?

src/api/app/jobs/project_do_project_copy_job.rb
coolo: `File.join` is about the file system and this is not windows vs unix - these are URLs. next to that: this code is in around 1000 places within obs.

The backend lib uses placeholders like `http_post(['/source/:project', project_name]`, but this API is not exported and putting this call into the backend library is *way* outside of the scope of this PR.

src/api/app/controllers/source_controller.rb
coolo: `User.session` never returns nobody - it would crash on nobody. But nobody is excluded from the controller. But I can make `User.session!` to make it more obvious

src/api/app/jobs/project_do_project_copy_job.rb
vpereira: then please use `URL.join` if you prefer https://apidock.com/ruby/URI/join/class

coolo: OBS uses in 0 places `URL.join`. What makes this place special?

coolo:
ping? As this is just implementing what Henne suggested, I will not continue here.

vpereira:
> ping? As this is just implementing what Henne suggested, I will not continue here.

Yes, looks like we are just missing a go from @hennevogel

src/api/app/jobs/project_do_project_copy_job.rb
vpereira: > OBS uses in 0 places `URL.join`. What makes this place special?

It's not really an argument, the only question that matters is if it's a better pattern or not.

vpereira:
I created an issue to track the missing spec https://github.com/openSUSE/open-build-service/issues/8189

PR #7609, opened 2019-05-21
Move method calls to callbacks

bgeuken:


coolo:
there is something about the bs_request model you will have to learn - and I wonder if I tell you or leave you the fun to discover it yourself :)

bgeuken:
> there is something about the bs_request model you will have to learn - and I wonder if I tell you or leave you the fun to discover it yourself :)

:construction_worker: (didn't find an indiana jones emoji)

PR #7610, opened 2019-05-21
Migrate "Search Owners" Page to Bootstrap

dmarcoux:
Bento:
search-owners-bento

Bootstrap:

On desktop:
search-owners

On mobile:
search-owners-mobile

### DIFF
--- a/src/api/app/views/webui2/webui/search/_results_owner.html.haml
+++ b/src/api/app/views/webui2/webui/search/_results_owner.html.haml
@@ -1,0 +1,38 @@
+.mt-4#search-results
+  %h4 Results:
+  - results.each do |result|
+    .search_result.mt-3
+      %h6
+        Responsible for
+        #{project_or_package_link project: result.project, package: result.package, short: false}
+        in #{result.rootproject}
+
+      %ul.list-unstyled
+        - users_and_groups = result.users.merge(result.groups) { |_role, user, group| user.concat(group) }
+        - users_and_groups.each do |role, records|
+          %li.mt-3
+            :ruby
+              emails = records.map(&:email).reject(&:blank?).join(', ')
+            = role.capitalize.pluralize
+            - if emails.present? && User.session
+              = mail_to(emails) do
+                %i.far.fa-envelope.ml-1{ title: "Send email to #{role.pluralize(records.size)}" }
+            .d-flex.flex-wrap.flex-column.flex-md-row
+              - records.each do |record|
+                .card.m-1.p-2.search-result
+                  .row.no-gutters
+                    .col-auto
+                      = link_to(user_or_group_show_path(record)) do
+                        = image_tag_for(record, size: 60, custom_class: 'align-self-center shadow-lg border border-secondary rounded border-2')
+                    .col
+                      .card-body.p-0.pl-2
+                        %ul.list-unstyled
+                          %li{ title: record.name }
+                            = truncate(record.name, length: 40)
+                          - if record.is_a?(User) && record.realname.present?
+                            %li.small
+                              = record.login
+                          %li.float-right
+                            - if record.email.present? && User.session
+                              = mail_to(record.email) do
+                                %i.far.fa-envelope{ title: "Send email to #{record.class.name.downcase}" }
--- a/src/api/app/views/webui2/webui/search/owner.html.haml
+++ b/src/api/app/views/webui2/webui/search/owner.html.haml
@@ -4,3 +4,47 @@
   = render partial: 'tabs'
   .card-body
     %h3 Search for people responsible for package:
+    .d-flex.justify-content-center
+      = form_tag(search_owner_path, method: :get, class: 'my-3 w-75') do
+        %input{ name: 'owner', type: 'hidden', value: '1' }
+        .form-group.d-flex
+          = text_field_tag('search_text', params[:search_text], placeholder: 'Search', autofocus: true, required: true, class: 'form-control')
+          %button.btn.btn-primary.ml-1{ type: 'submit', title: 'Search' }
+            %i.fa.fa-search
+        .form-row
+          .d-flex.flex-wrap.flex-column.flex-sm-row
+            .form-group.form-inline
+              .custom-control.custom-switch.ml-1.mr-4
+                = hidden_field_tag('devel', 'off', id: 'nodevel')
+                = check_box_tag('devel', 'on', @owner_devel.nil? || @owner_devel == '1', id: 'devel', class: 'custom-control-input')
+                %label.custom-control-label{ for: 'devel' } Follow Devel Projects
+
+            .form-group.form-inline
+              - limit_options = options_for_select([['Newest', '1'], ['Oldest', '-1'], ['All', '0']], @owner_limit)
+              = select_tag('limit', limit_options, class: 'custom-select border-0 pl-0 w-auto')
+              %label#limit-label{ for: 'limit' }
+                = 'package'.pluralize(@owner_limit.to_i.abs)
+    -# TODO: Ajaxify this section, so whenever changing the filters above, it fetches and refreshes the results
+    - if @results.present?
+      = render partial: 'results_owner', locals: { results: @results }
+
+      -# FIXME: This is also broken in Bento: https://github.com/openSUSE/open-build-service/issues/7616
+      - if User.session
+        %ul.nav
+          %li.nav-item
+            = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#request-bugowner-change-modal' }) do
+              %i.fas.fa-exchange-alt.text-primary
+              Request bugowner change
+        - projects_and_packages = @results.map { |result| { project: result.project, package: result.package } }
+        = render partial: 'request_bugowner_change_dialog', locals: { projects_and_packages: projects_and_packages }
+
+= content_for(:ready_function) do
+  :plain
+    $('#limit').change(function() {
+      if ($(this).val() === '0') {
+        $('#limit-label').text('packages');
+      }
+      else {
+        $('#limit-label').text('package');
+      }
+    });
--- a/src/api/app/assets/stylesheets/webui2/borders.scss
+++ b/src/api/app/assets/stylesheets/webui2/borders.scss
@@ -1,3 +1,7 @@
 .border-dotted {
   border-style: dotted !important;
 }
+
+.border-2 {
+  border-width: 2px !important;
+}
--- a/src/api/app/assets/stylesheets/webui2/search.scss
+++ b/src/api/app/assets/stylesheets/webui2/search.scss
@@ -1,0 +1,5 @@
+#search-results {
+  .search-result {
+    min-width: 22rem;
+  }
+}
--- a/src/api/app/assets/stylesheets/webui2/webui2.css.scss
+++ b/src/api/app/assets/stylesheets/webui2/webui2.css.scss
@@ -56,6 +56,7 @@
 @import 'groups';
 @import 'patchinfo';
 @import 'image_templates';
+@import 'search';
 
 html {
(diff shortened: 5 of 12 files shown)


dmarcoux:
[Owner Search](https://trello.com/c/T3WpUd5G/1107-owner-search)

dmarcoux:
I'm looking into the specs...

DavidKang:
@dmarcoux, looks good!, One thing that I realize is that the radio button of `ignore/following project` is not select by default. Is that on purpose?

dmarcoux:
> @dmarcoux, looks good!, One thing that I realize is that the radio button of `ignore/following project` is not select by default. Is that on purpose?

It was like this in the Bento UI, but I forgot to pick a default radio button when I migrated the view. I will change it so that it ignores devel projects by default.

msmeissn:
some form of list would be good

vpereira:
I personally know how in some teams interact with this information: They search for a packager, they copy their login information to use in another system (i.e bugzilla). They do it 20 times or more a day. I'm wondering if with the solution proposed here, are we improving their user experience?

dmarcoux:
> some form of list would be good

What do you mean exactly? The maintainers/bugowners are displayed one after another. Sure, it's now different, but the avatar is bigger and the font size for the login/name is also increased. The distinction between the users and groups is also gone. The groups were previously displayed as plain text (so without any link) above the users.

dmarcoux:
> I personally know how in some teams interact with this information: They search for a packager, they copy their login information to use in another system (i.e bugzilla). They do it 20 times or more a day. I'm wondering if with the solution proposed here, are we improving their user experience?

We agreed on porting this with minor changes (so the same information is displayed). The changes regarding the list of users/groups is also what we discussed.

vpereira:
I'm wondering if for the users wouldn't be better something like that:

Screenshot_2019-05-21_14-12-12

mschnitzer:
Agreed to some suggestions via Thumbsup (see above).

Regarding the mobile version: I would put a line break after each item.

dmarcoux:
> Agreed to some suggestions via Thumbsup (see above).
> 
> Regarding the mobile version: I would put a line break after each item.

The items are in a flex container, so they are displayed on the same line whenever they fit.

dmarcoux:
> > > Agreed to some suggestions via Thumbsup (see above).
> > > Regarding the mobile version: I would put a line break after each item.
> > 
> > 
> > The items are in a flex container, so they are displayed on the same line whenever they fit.
> 
> @dmarcoux I know, but it looks a bit weird though. As the items are not aligned, it looks a bit chaotic.

True, I will check if I can do something about this.

dmarcoux:
Let me know what you think. I did some changes.

The users all take the same space now. They also display in a column on devices with small viewports. The name and login can be copied (the link to the user/group page was moved to the image). There is a link to send an email to all maintainers/bugowners or to a single maintainer/bugowner. These links aren't displayed to anonymous users. I replaced the radio buttons by a switch for the `Follow Devel Projects` filter and by a dropdown for the package filter (Newest, Oldest, All).

saraycp:
Now it looks a lot nicer! :heart:

I just suggest removing the blue border around the avatar. IMHO, it looks better like we did in the groups page.
groups page

And I wonder if having the link in the image is intuitive. :thinking:

dmarcoux:
Regarding the border around the image, I did a bit of reading and found this [post](https://ux.stackexchange.com/questions/20697/how-to-indicate-that-an-image-is-a-link-too) on how to indicate that an image is a link too. There needs to be a difference between an image with a link and one without a link. This is where the border comes into play. The problem with how we did it on the group members page is that it's a bit tricky to copy the name (if you want too), the login is also between parentheses (it's annoying to copy, you need to remove them instead of simply double-clicking).

saraycp:
> Regarding the border around the image, I did a bit of reading and found this post on how to indicate that an image is a link too. There needs to be a difference between an image with a link and one without a link. This is where the border comes into play. It's also important to take the mobile users into account. Only having a hover state (so a different cursor) wouldn't cut it because there is no such thing as a hover state on mobile devices.

I understand, but the link in the image doesn't convince me anyway. And, in addition, I think the link makes it uglier.

> The problem with how we did it on the group members page is that it's a bit tricky to copy the name (if you want to), the login is also between parentheses (it's annoying to copy, you need to remove them instead of simply double-clicking).

When I compared with the groups page, I was talking about the image without the border, not the text.
But... now you have the name an the login in two different lines... I guess you're not having that problem anymore. Am I wrong?

src/api/app/views/webui2/webui/search/_results_owner.html.haml
```diff
@@ -0,0 +1,41 @@
...
+                        = image_tag_for(record, size: 60, custom_class: 'align-self-center shadow-lg border border-secondary rounded border-2')
+                    .col
+                      .card-body.p-0.pl-2
+                        %ul.list-unstyled
+                          - name = record.try(:realname) || record.name
+                          %li{ title: name }
+                            = truncate(name, length: 40)
+                          - if record.class.name == 'User'
+                            %li.small
+                              = record.login
```
Ana06: If the `realname` is not set you are render `login` twice

src/api/app/views/webui2/webui/search/_results_owner.html.haml
```diff
@@ -0,0 +1,41 @@
...
+                          %li.float-right
+                            - if record.email && User.session
```
Ana06: `email.present?` as email could be empty

src/api/app/views/webui2/webui/search/owner.html.haml
```diff
@@ -4,3 +4,47 @@
   = render partial: 'tabs'
   .card-body
     %h3 Search for people responsible for package:
+    .d-flex.justify-content-center
+      = form_tag(search_owner_path, method: :get, class: 'my-3 w-75') do
+        %input{ name: 'owner', type: 'hidden', value: '1' }
+        %input{ name: 'package', type: 'hidden', value: '0' }
+        %input{ name: 'project', type: 'hidden', value: '0' }
```
Ana06: why this 0 value?

src/api/app/views/webui2/webui/search/owner.html.haml
```diff
@@ -4,3 +4,47 @@
...
+    -# TODO: Ajaxify this section, so whenever changing the filters above, it fetches and refreshes the results
+    - if @results.present?
+      = render partial: 'results_owner', locals: { results: @results }
+      - if User.session
+        %ul.nav
+          %li.nav-item
+            = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#request-bugowner-change-modal' }) do
+              %i.fas.fa-exchange-alt.text-primary
+              Request bugowner change
+        - projects = @results.map { |result| { project: result.project, package: result.package } }
```
Ana06: calling projects to an array of hashes with keys project and package is confusing

src/api/app/views/webui2/webui/search/_results_owner.html.haml
```diff
@@ -0,0 +1,41 @@
+.mt-4#search-results
+  %h4 Results:
+  - results.each do |result|
+    .search_result.mt-3
+      %h6
+        Responsible for package
```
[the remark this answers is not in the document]
dmarcoux: It's like this because of my local setup. I put the root project as the same project containing the package.

In production, it is more like:

`Responsible for package bat for openSUSE set in package utilities / bat`

src/api/app/views/webui2/webui/search/_results_owner.html.haml
```diff
@@ -0,0 +1,41 @@
...
+        \#{params['search_text']}
+        for #{result.rootproject} set in
+        \#{project_or_package_link project: result.project, package: result.package, short: false}
+
+      %ul.list-unstyled
+        - records = result.users.merge(result.groups) { |_role, user, group| user + group }
+        - records.each do |role_with_records|
```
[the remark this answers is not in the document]
dmarcoux: Good suggestion, but I will use another name for the block to not over-shadow `records`.

src/api/app/views/webui2/webui/search/_results_owner.html.haml
```diff
@@ -0,0 +1,41 @@
...
+  - results.each do |result|
+    .search_result.mt-3
+      %h6
+        Responsible for package
+        \#{params['search_text']}
+        for #{result.rootproject} set in
+        \#{project_or_package_link project: result.project, package: result.package, short: false}
+
+      %ul.list-unstyled
+        - records = result.users.merge(result.groups) { |_role, user, group| user + group }
```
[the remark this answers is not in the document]
dmarcoux: I guess you meant `concat`

src/api/app/views/webui2/webui/search/_results_owner.html.haml
```diff
@@ -0,0 +1,41 @@
...
+        - records.each do |role_with_records|
+          %li.mt-3
+            :ruby
+              role = role_with_records.first
+              emails = role_with_records.second.map(&:email).reject(&:blank?).join(', ')
```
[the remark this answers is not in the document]
dmarcoux: I find this too small for a helper.

src/api/app/views/webui2/webui/search/_results_owner.html.haml
dmarcoux: Good catch, I'll add a check if `realname` is present.

src/api/app/views/webui2/webui/search/owner.html.haml
dmarcoux: True... I'll find a better name

src/api/app/views/webui2/webui/search/owner.html.haml
```diff
@@ -4,3 +4,47 @@
...
+                = hidden_field_tag('devel', 'off', id: 'nodevel')
+                = check_box_tag('devel', 'on', @owner_devel.nil? || @owner_devel == '1', id: 'devel', class: 'custom-control-input')
+                %label.custom-control-label{ for: 'devel' } Follow Devel Projects
+
+            .form-group.form-inline
+              - limit_options = options_for_select([['Newest', '1'], ['Oldest', '-1'], ['All', '0']], @owner_limit)
+              = select_tag('limit', limit_options, class: 'custom-select border-0 pl-0 w-auto')
+              %label#limit-label{ for: 'limit' }
+                = 'package'.pluralize(@owner_limit.to_i.abs)
+    -# TODO: Ajaxify this section, so whenever changing the filters above, it fetches and refreshes the results
```
dmarcoux: I would like to do this once this PR is merged as it's not a requirement to migrate this view.

src/api/app/views/webui2/webui/search/owner.html.haml
dmarcoux: I thought it was needed due to the `set_parameters` callback in the `search_controller`, but it isn't. I will remove them, except for `owner`, it's needed. We would need to refactor the `search_controller` to get rid of this.

src/api/app/views/webui2/webui/search/owner.html.haml
dmarcoux: I went with projects_and_packages. It's longer, but it is what it is...

src/api/app/views/webui2/webui/search/_results_owner.html.haml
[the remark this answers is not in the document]
dmarcoux: I will remove the first part, so it would be `Responsible for package utilities / bat`.

dmarcoux:
@Ana06, please have another look.

@msmeissn, @vpereira and @mschnitzer, anything to say following the changes?

mschnitzer:
@dmarcoux lgtm from my side

PR #7617, opened 2019-05-22
Prevent the StagingProjectCopyJob from using existing User#session

bgeuken:
This made it possible to create projects with different user permissions
than intended.


bgeuken:
This just became obsolete with #7602

PR #7628, opened 2019-05-25
Migrate appliances to Leap 15.1

vpereira:
Things that changed by default in 15.1:

- mysql listeing on localhost
- no user nobody available (necessary for installing npm modules)

The PR is done, I'm able to build the images locally and so on. However would be good if @hennevogel could try it out and if it's ok upload the images to the registry

### DIFF
--- a/docker-files/base/docker-bootstrap-15.1.sh
+++ b/docker-files/base/docker-bootstrap-15.1.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+for option in "$@"
+do
+  case "${option}" in
+  frontend)
+    zypper -n install --no-recommends --replacefiles \
+      sphinx \
+      nodejs10 npm10 \
+      mariadb-client \
+      mysql-devel \
+      sqlite3-devel \
+      git-core \
+      ruby2.5-devel cyrus-sasl-devel openldap2-devel libxml2-devel zlib-devel libxslt-devel \
+      perl-XML-Parser \
+      libffi-devel autoconf \
+      ruby2.5-rubygem-bundler
+    ;;
+
+  backend)
+    zypper ar -f https://download.opensuse.org/repositories/Cloud:/Tools/openSUSE_Leap_15.1/Cloud:Tools.repo
+    zypper ar -f https://download.opensuse.org/repositories/devel:/languages:/python/openSUSE_Leap_15.1/devel:languages:python.repo
+    zypper --gpg-auto-import-keys refresh
+    zypper -n install --no-recommends --replacefiles \
+      inst-source-utils \
+      obs-server obs-signd \
+      obs-service-download_src_package obs-service-download_files \
+      obs-service-download_url \
+      obs-service-format_spec_file obs-service-kiwi_import \
+      perl-Devel-Cover perl-Diff-LibXDiff \
+      python3-setuptools \
+      python3-ec2uploadimg \
+      aws-cli \
+      osc
+      #
+      # Azure client disabled temporarily till python3-pydocumentdb 2.0.1 can
+      # build for Leap 15.1
+      # azure-cli
+    ;;
+
+  memcached)
+    zypper -n install --no-recommends --replacefiles memcached
+    ;;
+
+  *)
+    echo "Error: possible options are: frontend|backend|memcached"
+    exit
+    ;;
+
+  esac
+done
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,5 @@
 CONTAINER_USERID = %x(id -u).freeze
-VERSION = '42.3'.freeze
+VERSION = '15.1'.freeze
 
 namespace :docker do
   desc 'Build our development environment'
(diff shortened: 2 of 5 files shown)


docker-files/base/docker-bootstrap-15.1.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+for option in "$@"
+do
+  case "${option}" in
+  frontend)
+    zypper -n install --no-recommends --replacefiles \
```
coolo: this won't fly. We want the containers to build in OBS - but OBS parses the Dockerfile to understand what packages are needed and zypper commands are magic. Hiding them in shell scripts won't work.

vpereira: Our development environment we are building and uploading it to dockerhub. It's not planned to do it in this PR, but we can and should do it later. So I would handle it as nitpick

coolo: I won't because in test development we build in OBS - and we should stop using the old way and you make it *worse*.

vpereira: why do we make it worse using dockerhub @coolo

coolo: for once: because SUSE as a company moves away from docker (towards podman and it's own registry). But that's not the main problem  - the main problem is the unifaction of test containers and dev containers. And those live in obs.

vpereira: You didn't answer why I'm making it technically worse with this PR, updating it from 42.3 to 15.1 @coolo.

src/api/Dockerfile.frontend-base
```diff
@@ -2,6 +2,8 @@ FROM openbuildservice/base
 
 RUN /root/bin/docker-bootstrap.sh frontend
 
+# necessary for npm install -g
+RUN useradd -r --shell /dev/null nobody
```
coolo: and don't: just install it. zypper -n in system-user-nobody

docker-files/base/docker-bootstrap-15.1.sh
coolo: well, ok - you're not making it worse (beside the obvious duplication of most files), but you're not improving the situation either.

coolo:
What I'm basically asking for is exploding the bootstrap.sh into the 3 files where it's called. It's 3 unique set of instructions called exactly 3 times with different arguments. So the benefit of this script is 0 - and by having the instructions in the Dockerfiles (which really shouldn't be duplicated but be renamed) you make it possible for OBS to detect dependencies.

It still won't work within OBS without more changes, but at least we get there.

src/api/Dockerfile.frontend-base
vpereira: a great, done

docker-files/base/docker-bootstrap-15.1.sh
vpereira: Hey @coolo. Thank you for the clarification. Actually I'm not duplicating any file, I'm adding support to 15.1 without remove the support to 42.3. That's necessary until we are sure that everything is working properly, then we can remove all 42.3 files as soon as we stick with 15.1. Is that ok for you or do you think that we have to radically drop support for 42.3 loosing the change to compare both environments if extra debug is needed?

vpereira: @coolo, 42.3 goes EOL in the end of the month. How moving our development environment to the last maintained leap doesn't improve the situation?

coolo:
Actually, I wonder what blocks the usage of the test containers for development. I see nothing - the only thing we'd require is a tiny container build locally to set the userid

vpereira:
> What I'm basically asking for is exploding the bootstrap.sh into the 3 files where it's called. It's 3 unique set of instructions called exactly 3 times with different arguments. So the benefit of this script is 0 - and by having the instructions in the Dockerfiles (which really shouldn't be duplicated but be renamed) you make it possible for OBS to detect dependencies.
> 

For your initial comments I wasn't able to get this information. Please keep in mind that I didn't introduce any new file structure, I just ported what we have to 15.1. I'm not introducing any new change with this PR, and if we close it, we won't be improving anything. Regarding SUSE topics, I would prefer that you discuss it off-line with our PO. 

> It still won't work within OBS without more changes, but at least we get there.

As you can see today, our development environment is using dockerhub already. I'm not introducing anything new here. As I told you in your fist comment, I agree that we should do it later but I don't agree that we have to change everything at the same time.

coolo:
Let's say I don't believe in later.

docker-files/base/docker-bootstrap-15.1.sh
coolo: If you want to keep the 42.3 containers, tag them on dockerhub

coolo:
https://github.com/openSUSE/open-build-service/pull/7631 is my version of what I thought is the goal of the container story

PR #7630, opened 2019-05-27
Test circleci on Leap 15.1 containers

coolo:
In the light of the discussion in #7628 I went forward to setup OBS project


hennevogel:
Please let's do this step after we are done with the 2.10 release. We have enough other stuff on our plate right now.

vpereira:
@coolo I think this one we don't need since we run the CI in the same OS that we run in production (i.e SLE based) so can I close it or do you see another reason to do it?

coolo:
leap 15.1 *is* sle based.

PR #7631, opened 2019-05-27
Rebase dev containers on OBS built containers

coolo:
Don't duplicate containers on dockerhub and obs - but just use the obs containers for both testing and development.

The Rakefile also contains provisioning for deployment, I'm not sure how to handle this.


hennevogel:
Please let's do this step after we are done with the 2.10 release. We have enough other stuff on our plate right now.

vpereira:
@hennevogel looks like the openSUSE:Leap:42.3 was disabled on `OBS:Server:Unstable`. I tested this PR  locally now and worked flawless. Can we merge it? Otherwise we cannot build our docker containers locally anymore.

vpereira:
i rebased it with master

src/api/docker-files/Dockerfile
```diff
@@ -3,11 +3,19 @@
 # contained rails app generating files in the git checkout with
 # some strange user...
 
-FROM openbuildservice/frontend-base
+FROM registry.opensuse.org/obs/server/unstable/container/leap151/containers/openbuildservice/frontend-base
```
hennevogel: The container on docker hub is "just" a gem cache and rebuilds every time we commit. The container on OBS doesn't.

hennevogel:
@coolo  I would also appreciate if you could stick to smaller pull requests in the future that change one thing....

src/api/docker-files/Dockerfile
hennevogel: Ah you do already, sorry

src/api/docker-files/Dockerfile
```diff
@@ -3,11 +3,19 @@
...
 ARG CONTAINER_USERID
 
+# install foreman from rpms, we only need it in docker
+RUN zypper ar http://download.opensuse.org/repositories/devel:/languages:/ruby:/extensions/openSUSE_Leap_15.1/ dlre
+RUN zypper -n --gpg-auto-import-keys ref
+RUN zypper -n in -C 'rubygem(foreman)' 'rubygem(mailcatcher)'
+
+# for lint task
+RUN npm install -g jshint
```
hennevogel: Why do you move this here instead?

vpereira: what would be the alternative  @hennevogel ?

hennevogel: Leave it where it was, in the frontend-base container

hennevogel: All of us rebuild this container all the time.

coolo: that would require jshint to exist as obs package - you can't npm install on obs

DavidKang: @hennevogel, so all the lines 9-15 should be part of frontend-base, right?

vpereira: @DavidKang we will need the following repository `devel:languages:nodejs` to be able to install `jshint`

hennevogel: That package is broken, all of the packages on OBS are...

hennevogel: Guess we have to life with this here now until we find a solution...

PR #7633, opened 2019-05-27
Rename api_exception to api_error and add specs for that

vpereira:
Because I'm using the APIerror.setup in the `trigger_controller` in a different way than before, and because I would like to refactor this class in the future,  I just added some spec for the class and then rename the file  from api_exception to api_error.


vpereira:
> I wouldn't be suprised if you can remove half the require lines now - and rely on autoload there as well

me neither. I just didn't want to squash so many changes in one PR.

PR #7634, opened 2019-05-27
fix permission handling for tokens on packages via project links

adrianschroeter:
We need to validate the project permissions in that case
for "rebuild" operations instead of the package permissions.

move some checks from generic method in the operation specific ones
therefore

### DIFF
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -11,6 +11,8 @@ class TriggerController < ApplicationController
   #
   skip_before_action :extract_user
   skip_before_action :require_login
+  skip_before_action :validate_params # new gitlab versions send other data as parameters,
+  # which which we may need to ignore here. Like the project hash.
 
   # to get access to the method release_package
   include MaintenanceHelper
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -17,12 +17,22 @@ class TriggerController < ApplicationController
 
   include Trigger::Errors
 
+  def check_rebuild_permission!
+    return if User.session.can_modify_project?(@prj)
+    return if @pkg.project == @prj && policy(@pkg).update?
+
+    raise NoPermissionForPackage.setup('no_permission', 403, "no permission for package #{@pkg} in project #{@prj}")
+  end
+
   def rebuild
-    Backend::Api::Sources::Package.rebuild(@pkg.project.name, @pkg.name)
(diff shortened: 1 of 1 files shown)


adrianschroeter:
just coded without any testing yet ...

src/api/app/controllers/trigger_controller.rb
```diff
@@ -77,9 +85,19 @@ def require_valid_token
...
-    @pkg = @token.package || Package.get_by_project_and_name(params[:project].to_s, params[:package].to_s, use_source: true)
-
-    raise ActiveRecord::RecordNotFound unless @pkg
-    raise NoPermissionForPackage.setup('no_permission', 403, "no permission for package #{@pkg} in project #{@pkg.project}") unless policy(@pkg).update?
+    if @token.package
+      @pkg = @token.package
+      @prj = @pkg.project
+    else
+      @prj = Project.get_by_name(params[:project])
+      if @token.class == Token::Rebuild
```
coolo: so the rebuild token is for a project and not for a package?

adrianschroeter: it can be for a source package object

but it can be also global and then you need to check it for project where the package build container lives.

As discussed face to face: It would be better to have a package build container object and use policies for that. This would also help in other controllers and operations. But it is outside of the scope here...

vpereira:
I will fix the linter and see if i can write a test for that. Thanks @adrianschroeter

vpereira:
@adrianschroeter the tests are failing with a 500 error, probably some error was introduced.. could you please try it locally and see if you can tackle the issue down?

adrianschroeter:
On Freitag, 31. Mai 2019, 09:43:52 CEST Victor Pereira wrote:
> @adrianschroeter the tests are failing with a 500 error, probably some error
> was introduced.. could you please try it locally and see if you can tackle
> the issue down?

this is fixed now. And also the new gitlab breakage

-- 

Adrian Schroeter
email: adrian@suse.de

vpereira:
@adrianschroeter  now looks good! Before approve it, there is anything that we should update in the documentation (i.e https://openbuildservice.org/help/manuals/obs-admin-guide/obs.cha.administration.html#id-1.6.7.9.9.4) ?

adrianschroeter:
@vpereira well, we could extend it, but all of it is still valid

vpereira:
> @vpereira well, we could extend it, but all of it is still valid

Ok if you have time and think that's necessary, please extend it.. accepting and merging

PR #7640, opened 2019-05-28
Fix Save button on editing Kiwi image

eduardoj:
When we migrated to the new UI, the variable isOutdatedUrl was not passed, and this resulted in a 'Save' button that didn't trigger any action.


hennevogel:
How about a feature spec?

eduardoj:
> How about a feature spec? That at least changing _any_ attribute through the interface works.

Added. Please review again.

eduardoj:
> This spec makes 28 connections to the backend? Did you run this multiple times maybe?

Nops. I ran it only once.

Anyway, I checked again, dropping the cassette and creating it again, and obtained the same number of requests: 27.

PR #7641, opened 2019-05-28
hot fix to get basic functionality back for now

adrianschroeter:
seems there is more, but no time right now.

many clases plus methods plus global variables are not really
easy to debug

### DIFF
--- a/src/api/app/models/owner_search/assignee.rb
+++ b/src/api/app/models/owner_search/assignee.rb
@@ -44,7 +44,8 @@ module OwnerSearch
       # no filter defined, so do not check for roles and just return container
       return owner if @rolefilter.empty?
       # lookup in package container
-      extract_from_container(owner, pkg, @rolefilter)
+      matched = extract_from_container(owner, pkg, @rolefilter)
+      return matched if matched
 
       # eventually fallback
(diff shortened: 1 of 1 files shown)


bugfinder:
IBS added as api-15-owner_hotfix.patch

coolo:
it would be cool to understand the problem you're trying to fix - because as you can see on the red marks here, you broke tests.

coolo:
This is about the empty result of `iosc bugowner -v bash`, do I guess right?

bugfinder:
looks like the patch broken more than it fixed, reverted for now.

coolo:
@adrianschroeter your "hot fixed regressions" patched 10 minutes before you leave the office are pretty annoying. Especially for things that aren't regressions at all.

211b125291d201636425d88441c9c43220c02655 is your commit from 2014 and added a filter to only return confirmed accounts. While it seems your whole stunt was about subaccounts not returned.

So if the requirements changed meanwhile, the code should change too - that's not exactly the definition of regression. I'll create a PR for what you seem to be after - returning subaccounts.

coolo:
-> https://github.com/openSUSE/open-build-service/pull/7642

hennevogel:
Love is in the air...

PR #7642, opened 2019-05-28
Return bugowners even if they are a subaccount

coolo:
For SLE we have a lot of team accounts that are bugowners for packages.

Those accounts are not confirmed (they are not employees), but they are in state :subaccount - but they still are a valid answer if looking for a owner of a package

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -66,10 +66,12 @@ class User < ApplicationRecord
   has_and_belongs_to_many :announcements
   has_many :commit_activities
 
+  scope :confirmed, -> { where(state: 'confirmed') }
   scope :all_without_nobody, -> { where('login != ?', NOBODY_LOGIN) }
   scope :not_deleted, -> { where.not(state: 'deleted') }
   scope :not_locked, -> { where.not(state: 'locked') }
   scope :with_login_prefix, ->(prefix) { where('login LIKE ?', "#{prefix}%") }
+  scope :active, -> { confirmed.or(User.where(state: :subaccount, owner: User.confirmed)) }
 
   scope :list, lambda {
     all_without_nobody.includes(:owner).select(:id, :login, :email, :state, :realname, :owner_id, :updated_at, :ignore_auth_services)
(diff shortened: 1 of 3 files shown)


coolo:
@bugfinder I applied this now instead and now both bugowner calls succeed. I hope that helps

adrianschroeter:
the accounts were confirmed.

you need to check the state of the owner instead.

adrianschroeter:
and the return value of the container result is still ignored on purpose?

coolo:
well, I don't know what you expect. I only know the code you wrote :)

hennevogel:
I have a suggestion for you: How about you document this part/feature and the reasons it works like it does in our [developer wiki](https://github.com/openSUSE/open-build-service/wiki/Owner-search)? Then we can look up what has been discussed today, when we talked about this again in 2024 :cowboy_hat_face:

bugfinder:
applied by coolo manually on IBS (api-internal)
added to IBS package now by oertel

coolo:
extended the logic to what @adrianschroeter said

src/api/app/models/user.rb
```diff
@@ -66,10 +66,12 @@ class User < ApplicationRecord
   has_and_belongs_to_many :announcements
   has_many :commit_activities
 
+  scope :confirmed, -> { where(state: 'confirmed') }
   scope :all_without_nobody, -> { where('login != ?', NOBODY_LOGIN) }
   scope :not_deleted, -> { where.not(state: 'deleted') }
   scope :not_locked, -> { where.not(state: 'locked') }
   scope :with_login_prefix, ->(prefix) { where('login LIKE ?', "#{prefix}%") }
+  scope :active, -> { confirmed.or(User.where(state: :subaccount, owner: User.confirmed)) }
```
hennevogel: Wasn't there a problem with using scopes within scopes a couple of weeks ago?

coolo: Jein - the problem was that calling functions from within a scope uses that scope as default scope. Not problematic here

dmarcoux:
@adrianschroeter Is this fine for you as it is now?

coolo:
I guess @adrianschroeter no longer cares

PR #7645, opened 2019-05-29
Status Monitor Skeleton in Bootstrap

saraycp:
The skeleton is already migrated and includes a partial for each of the three sections on Status Monitor. We now can work on this in parallel.

Screenshot-2019-5-29 Open Build Service

### DIFF
--- a/src/api/app/controllers/webui/monitor_controller.rb
+++ b/src/api/app/controllers/webui/monitor_controller.rb
@@ -36,6 +36,7 @@ class Webui::MonitorController < Webui::WebuiController
       @workers_sorted = workers.sort_by { |a| a[0] } if workers
       @available_arch_list = Architecture.available.order(:name).pluck(:name)
     end
+    switch_to_webui2 if Rails.env.development?
   end
 
   def update_building
--- a/src/api/app/views/webui2/webui/monitor/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui2/webui/monitor/_breadcrumb_items.html.haml
@@ -1,0 +1,3 @@
+= render partial: 'webui/main/breadcrumb_items'
+
+%li.breadcrumb-item.active{ 'aria-current' => 'page' } Monitor
--- a/src/api/app/views/webui2/webui/monitor/_events.html.haml
+++ b/src/api/app/views/webui2/webui/monitor/_events.html.haml
@@ -1,0 +1,1 @@
+%h2 Statistical Plots
--- a/src/api/app/views/webui2/webui/monitor/_lights.html.haml
+++ b/src/api/app/views/webui2/webui/monitor/_lights.html.haml
@@ -1,0 +1,1 @@
+%h2 Server Status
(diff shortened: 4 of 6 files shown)


saraycp:
[Monitor Page Skeleton](https://trello.com/c/WBt0fzZg/1154-monitor-page-skeleton)

src/api/app/controllers/webui/monitor_controller.rb
```diff
@@ -36,6 +36,7 @@ def index
       @workers_sorted = workers.sort_by { |a| a[0] } if workers
       @available_arch_list = Architecture.available.order(:name).pluck(:name)
     end
+    switch_to_webui2
```
DavidKang: We need to add `switch_to_webui2 if Rails.env.development?`, because we are starting this migration.

saraycp: Done! Thanks!

src/api/app/assets/stylesheets/webui2/monitor.scss
```diff
@@ -0,0 +1,168 @@
+.builderbox {
```
hennevogel: How do you know you need all this?

src/api/app/assets/javascripts/webui2/monitor.js
```diff
@@ -0,0 +1,10 @@
+function monitor_ready() { // jshint ignore:line
```
hennevogel: How do you know you need this function?

vpereira:
> The skeleton is already migrated and includes a partial for each of the three sections on Status Monitor. We now can work on this in parallel.

some screenshot?

src/api/app/assets/javascripts/webui2/monitor.js
saraycp: Taking into account that we don't plan to change many things, I guess we are going to keep most of this code, maybe with some slight differences.

src/api/app/assets/stylesheets/webui2/monitor.scss
saraycp: Taking into account that we don't plan to change many things, I guess we are going to keep most of this code, maybe with some slight differences.

saraycp:
> :ghost: :skull: :moon: *Boo*strap!

@hennevogel uuuhh :fearful: :smile:

I think that most of the CSS and JS code is going to keep like it is. But I can remove it from the skeleton and add only the final version later... :thinking: 

@vpereira screenshot added, but it's almost empty anyway.

src/api/app/assets/javascripts/webui2/monitor.js
hennevogel: Let's only take over what we actually are going to use please. We have to maintain this code.

saraycp: :+1: done!

saraycp:
Ok, @hennevogel, without futurology now. Can you please review again?

PR #7646, opened 2019-05-29
Bs request action (refactoring model and specs)

vpereira:
The `BsRequestAction` is one of the models with lowest test coverage. Looking in the code, I was able to see some issues with the specs and the cassettes. After making the `User.current` private, some specs should be updated and the cassettes refreshed. Beside it I expanded the test coverage on the `BsRequestAction` model and refactored it little bit. Later on, @hennevogel found out that this test didn't need the VCRs at all.

### DIFF
--- a/src/api/spec/models/bs_request_action_spec.rb
+++ b/src/api/spec/models/bs_request_action_spec.rb
@@ -1,13 +1,13 @@
 require 'rails_helper'
-# WARNING: If you change tests make sure you uncomment this line
-# and start a test backend. Some of the BsRequestAction methods
-# require real backend answers for projects/packages.
-# CONFIG['global_write_through'] = true
 
-RSpec.describe BsRequestAction, vcr: true do
+RSpec.describe BsRequestAction do
   let(:user) { create(:confirmed_user, login: 'request_user') }
 
-  context 'encoding of sourcediffs' do
+  before do
+    allow(User).to receive(:session!).and_return(user)
+  end
+
+  context 'encoding of sourcediffs', vcr: true do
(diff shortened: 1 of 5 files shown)


vpereira:
@hennevogel could you review it?

src/api/spec/models/bs_request_action_spec.rb
```diff
@@ -1,13 +1,13 @@
...
-# require real backend answers for projects/packages.
-# CONFIG['global_write_through'] = true
 
-RSpec.describe BsRequestAction, vcr: true do
+RSpec.describe BsRequestAction do
   let(:user) { create(:confirmed_user, login: 'request_user') }
 
-  context 'encoding of sourcediffs' do
+  before do
+    allow(User).to receive(:session!).and_return(user)
```
hennevogel: :thought_balloon: I wonder why we mock this instead of setting `User.current`...

vpereira: well since `User.current=` is now private, we would have to start to do nasty things, which would probably lead to the question: "Why not just mock `session!` ?".. If you don't mind I would like to keep it like that.

hennevogel: Sorry I've meant `User.session`, this is the right "replacement".

coolo:
Your pitch doesn't match "Coverage decreased (-0.06%) to 90.721%"

vpereira:
@coolo not sure where it came from. Comparing the assets (code coverage) from this branch and master, the coverage in this method increased from 71% to 73%

PR #5308, opened 2018-07-05
Disable Cassettes for backend

coolo:
Let's start the VCR discussion with something real

PR #7655, opened 2019-05-31
Disable vcr for Issue model and remove unecessary cassettes

vpereira:
We are trying to reduce the usage of vcr as much as possible. The model Issue didn't need it enabled and therefore, we disabled it, and removed the unecessary cassettes.


coolo:
this is pretty time consuming. Basically all model tests I didn't flag with backend in https://github.com/openSUSE/open-build-service/pull/5308/commits/fee146b6fbe652aadd69264c61c4495cf1ebde11 don't need cassettes

PR #7656, opened 2019-05-31
Migrate Server Status to Bootstrap on status monitor page

saraycp:
In the old code there were some JavaSscript that showed some hidden text
when hovering an element with class 'scheduler_status'. However, it's
impossible to find that class in the code, so it seems that
functionality is not used anymore.

Screenshot-2019-5-31 Open Build Service


saraycp:
[Monitor Page: Server Status](https://trello.com/c/H1uHf1Gg/1150-monitor-page-server-status)

saraycp:
@DavidKang thanks for your advice. Can you have another look?

PR #7662, opened 2019-06-03
[backend] fix cross architecture import jobs

adrianschroeter:
(aka ExportFilter).  a72e0962d0e introduced a new validation for the job
arch, but we wrote the origin arch into the job since years.

Changing this to the target arch now


adrianschroeter:
nah, leads to wrong import names ....

adrianschroeter:
fixing it the other way around, by accepting the "wrong" arch on import jobs.

PR #7668, opened 2019-06-04
Fix code duplication offense in relationship

coolo:
codeclimate said it takes 2 hours to fix, so I thought I try

### DIFF
--- a/src/api/app/mixins/webui/manage_relationships.rb
+++ b/src/api/app/mixins/webui/manage_relationships.rb
@@ -1,5 +1,40 @@
 module Webui::ManageRelationships
-  def load_obj
+  def save_person_or_group(what)
+    authorize main_object, :update?
+    begin
+      Relationship.add_role(main_object, Role.find_by_title!(params[:role]), check: true, user: params[:userid], group: params[:groupid]) # report error on duplicate
+      main_object.store
+    rescue NotFoundError,
+           Relationship::SaveError => e
+      flash[:error] = e.to_s
+      if switch_to_webui2
+        redirect_to(users_path)
+      elsif what == :person
+        redirect_to(add_path(:add_person))
+      else
+        redirect_to(add_path(:add_group))
+      end
+      return
+    end
+    respond_to do |format|
+      format.js { render json: {}, status: :ok }
+      format.html do
+        success_str = what == :person ? "user #{params[:userid]}" : "group #{params[:groupid]}"
+        flash[:success] = "Added #{success_str} with role #{params[:role]}"
+        redirect_to users_path
+      end
+    end
+  end
+
+  def save_person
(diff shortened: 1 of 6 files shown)


coolo:
OK, that code complexity check due to renaming the function was a little unexpected - continue later :)

src/api/app/models/relationship/add_role.rb
```diff
@@ -0,0 +1,48 @@
+class Relationship::AddRole
```
vpereira: can you unit test it?

src/api/app/mixins/webui/manage_relationships.rb
```diff
@@ -1,50 +1,48 @@
...
-      format.js { render json: {}, status: :ok }
-      format.html do
-        flash[:success] = "Added group #{params[:groupid]} with role #{params[:role]}"
-        redirect_to users_path
-      end
-    end
+    save_person_or_group(:group)
+  end
+
+  def load_obj
```
vpereira: ```load_obj``` is a real bad name.. if you are already making the code better, could you please rename it too?

coolo: actually I only moved it to private - it's only used in another function.

vpereira: yes I know it, but to rename it would be great too.

src/api/app/models/relationship/add_role.rb
coolo: it's basically tested from the unit tests running add_person and add_group, but I added some rather minimal test.

src/api/app/models/relationship/add_role.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  private
+
+  attr_accessor :package_or_project, :role, :user, :group, :check
+
+  def duplicate?
+    return unless package_or_project.relationships.where(user: user, group: group, role: role).exists?
+    raise SaveError, 'Relationship already exists' if check
+    debug_str = user ? "user #{user.login}" : "group #{group.title}"
```
vpereira: can you remove this debug code, please? I know it was already there, but if you could remove it would be great!

coolo: why?

vpereira: because to have a debug here is arbitrary. Why this method and not the others? We should have a way to introspect but to have arbitrary debugging is a smell.

coolo: hmm, that's true for the other 140 debug statements as well. So I consider this a nitpick - and if would have been easy to mention it in your first review. So I don't want to reopen this

vpereira: fair enough.

PR #7687, opened 2019-06-05
Pass default params to cloud upload jobs DataTable

eduardoj:
With the migration from the old UI to the new UI the default parameters to initialize the cloud upload jobs DataTable were not included. Now they are added, and also it is added the 'responsive' parameter.


eduardoj:
Link for reviewing: https://obs-reviewlab.opensuse.org/eduardoj-fix_upload_jobs_table/cloud/upload/

PR #7692, opened 2019-06-06
Fix behaviour of triggering multibuild containers

adrianschroeter:
It is okay to trigger rebuild of a multibuild container when access to
main container got approved

Also specify explicit behaviour on service and release actions


vpereira:
@adrianschroeter i retriggered again the rspec build.. if it fails again, could you please rebase it with master?

adrianschroeter:
seems to be good now

PR #7693, opened 2019-06-06
Refactor method release. Extract code to external method

vpereira:
Extract code from controller to a new method to simplify the controller.


coolo:
you need to push to your branch for circleci to pickup the PR :(

vpereira:
> you need to push to your branch for circleci to pickup the PR :(

yes. I wait until is green to open a PR, then you open a PR and circleci doesn't report to GH, so I have to re-run the whole workflow again..

PR #7697, opened 2019-06-06
Migrate new project view to bootstrap

### LINKED ISSUE #7278: Project#new view is not migrated to Bootstrap
# Issue Description
We migrated the Project views to the Bootstrap UI, but Project#new view is still using the old UI.

Expected Result
============
The Project#new view should be migrated to Bootstrap or deleted. We already use modals in the Project#subprojects view to do the same (creating subprojects is like creating projects). We need to align these views. I believe we should get rid of the modals.

How to Reproduce
===============
1. Go to https://build.opensuse.org/project/new
2. See the old UI

Further Information
===============
I found this out when clicking on the `Create Home` link at the top right.

project-new-bootstrap

eduardoj:
Fixes #7278.

Before:

Screenshot from 2019-06-06 17-51-20_project_new_before

After:

Screenshot from 2019-06-06 17-56-20_project_new_after

### DIFF
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -104,6 +104,8 @@ class Webui::ProjectController < Webui::WebuiController
     @project.name = params[:name] if params[:name]
 
     @show_restore_message = params[:restore_option] && Project.deleted?(params[:name])
+
+    switch_to_webui2
   end
 
   def new_package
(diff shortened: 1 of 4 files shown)


eduardoj:
[Address Feedback Sprint 59](https://trello.com/c/8I2pIzPH/1159-address-feedback-sprint-59)

PR #7699, opened 2019-06-07
[dist] obsdodup starts after obsapisetup

M0ses:
otherwise the following errors could occur:

```
Jun 07 12:14:21 localhost systemd[1]: Started OBS dodup, updates download on demand metadata.
Jun 07 12:14:22 localhost bs_dodup[1414]: BSConfig.pm did not return a true value at /usr/lib/obs/server/BSConfiguration.pm line 27.
Jun 07 12:14:22 localhost bs_dodup[1414]: BEGIN failed--compilation aborted at /usr/lib/obs/server/BSConfiguration.pm line 27.
Jun 07 12:14:22 localhost bs_dodup[1414]: Compilation failed in require at /usr/lib/obs/server/bs_dodup line 19.
Jun 07 12:14:22 localhost bs_dodup[1414]: BEGIN failed--compilation aborted at /usr/lib/obs/server/bs_dodup line 19.
Jun 07 12:14:22 localhost systemd[1]: obsdodup.service: Main process exited, code=exited, status=255/n/a
Jun 07 12:14:22 localhost systemd[1]: obsdodup.service: Unit entered failed state.
Jun 07 12:14:22 localhost systemd[1]: obsdodup.service: Failed with result 'exit-code'.
```

### DIFF
--- a/dist/systemd/obsdodup.service
+++ b/dist/systemd/obsdodup.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=OBS dodup, updates download on demand metadata
 After=network.target
+After=obsapisetup.service
 
 [Service]
 EnvironmentFile=/etc/sysconfig/obs-server


M0ses:
Not tested in live environment

adrianschroeter:
depending on obsapisetup is not a good idea, it may not even run on the same machine ...

dodup should be able to run without the need to have an api running at all since it is backend, so maybe we need to be more tolerant on missing config or have better defaults?

coolo:
then obsapisetup shouldn't create BSConfig.pm :)

coolo:
Actually according to grep obsstoragesetup does that

M0ses:
@coolo: Your argument is valid: We _should_ start after obsstoragesetup/obsapisetup, so "After" should be the right statement/option.

But I see no need to start before obsapisetup has finished

@adrianschroeter : As far as I understood from the docs - "After" would not depend on starting  obsstoragesetup/obsapisetup - but if they are enabled  - they would be considered

adrianschroeter:
after should be fine indeed.

PR #7700, opened 2019-06-07
Fix value of worker log directory variable (OBS_LOG_DIR)

eduardoj:
Trying to setup a worker with rpm packages in a bare metal machine resulted in failures creating the log files. The default directory for the log files of workers should be the same as for the other OBS services.

Another reason to change to the default directory is that there is no code to create this log subdirectory.

Some code of the `obsworker` script was also cleaned up.

### DIFF
--- a/dist/obsworker
+++ b/dist/obsworker
@@ -424,10 +424,10 @@ case "$1" in
                 WORKERID="${HOSTNAME}:$I"
             fi
             R=$OBS_WORKER_DIRECTORY/root_$I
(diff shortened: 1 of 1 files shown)


eduardoj:
@vpereira:

> looks like you are changing the indentation in some places.. if that's intentional, maybe would you like to mention it in the commit message?

In the second commit ("Cleanup code") it says in the last line:
```
- Unify indentation to 4 characters
```
Is that what you are referring to?

Issue #7551, opened 2019-05-13
Submit request form is not pre-filled with target project


bgeuken:
As far as I remember this form used to be filled out with the target project. This was at least the case when the source project was branched from another one.

How to Reproduce
===============
1. Branch a project
2. Edit the sources
3. Submit the project

=> The target project field is empty

hennevogel:
There is one more place on `/package/rdiff`

danidoni:
> There is one more place on `/package/rdiff`

I've taken care of this in #9029

PR #7719, opened 2019-06-10
Fix submit request input fields

DavidKang:
Target project and package weren't pre-filled.


dmarcoux:
This is about fixing the issue #7551, right? It should be in the commit message.

PR #7727, opened 2019-06-11
Use content_for with running ruby code

eduardoj:
As the `content_for` helper is not inserting any content directly in the call, use `-` instead of `=` for calling the helper.

Also remove the parenthesis on all the calls of `content_for` with the `:ready_function` parameter, as it is more often used without parenthesis in the rest of the code.


dmarcoux:
You are right!

We also use `:plain` in some cases and in others, we don't. I think we should align this too in this PR.

eduardoj:
@dmarcoux:

> We also use `:plain` in some cases and in others, we don't. I think we should align this too in this PR.

I only found one case using `:plain` in an one-liner. Done. :+1:

dmarcoux:
Adding the review app to make sure the JavaScript is still working (it should)

PR #7740, opened 2019-06-12
Show New Image link on the main page

DavidKang:
Now the`new image` link is also visible for none logged users.


dmarcoux:
None logged user is not right. I suggest: anonymous users.

DavidKang:
Because is the current behaviour in `bento`. I know that if the user is not logged I can't interact too much with the image template (only going to the projects), but the feature will be visible for those potentials future users. And if we are not interested in this feature we could always disable it behind the Feature Switch.

PR #7741, opened 2019-06-12
Add 'cloud_upload_azure` to the default configuration

DavidKang:
We added `cloud_upload_azure` as `false` in the default configuration of
`feature switch`

### DIFF
--- a/src/api/lib/feature_switch/obs_repository.rb
+++ b/src/api/lib/feature_switch/obs_repository.rb
@@ -5,7 +5,8 @@ module Feature
     class ObsRepository < YamlRepository
       DEFAULTS = {
         image_templates: true,
-        cloud_upload: false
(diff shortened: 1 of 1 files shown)


dmarcoux:
Why is it needed?

vpereira:
> Why is it needed?
> 
> The commit description is not really useful, it's almost the same as the commit message.

I think as default we should have the feature enable. So I'm fine with that, another question: Do we have to update any documentation? Looking [here](https://openbuildservice.org/help/manuals/obs-admin-guide/obs.cha.installation_and_configuration.html), nothing is said about globally enable or disable it

DavidKang:
I will close this PR in favour of [this approach](https://github.com/openSUSE/open-build-service/pull/7759)

PR #7746, opened 2019-06-13
Handle exception in `Feature.active?` method

DavidKang:
We decide to use the `dig` method to make the `Feature.active?` more robust
if the configuration isn't in a proper format.

### DIFF
--- a/src/api/spec/lib/feature_switch/feature_spec.rb
+++ b/src/api/spec/lib/feature_switch/feature_spec.rb
@@ -1,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe Feature do
+  let(:user) { create(:confirmed_user, :in_beta, login: 'Tom') }
+  subject { Feature.active?('bootstrap') }
+
+  context 'with beta users' do
+    before do
+      User.session = user
+      Feature.instance_variable_set(:@perform_initial_refresh_for_user, true)
+    end
+
+    context 'when the file has "beta" key' do
+      it { expect(subject).to be_truthy }
+    end
+
+    context 'when the file doesn\'t have "beta" key' do
+      before do
+        allow(YAML).to receive(:load_file).and_return({})
+      end
+
+      it { expect(subject).to be_falsey }
+    end
+  end
+end
--- a/ReleaseNotes-2.10
+++ b/ReleaseNotes-2.10
@@ -42,6 +42,7 @@ Frontend:
    feature, switch database.yml to utf8mb4 encoding
  * Admins can write Terms of Services, via the API, and they will be shown in the WebUI to
    users unless they acknowledge them.
+ * Added `beta` environment in 'config/feature.yml' to toggle features in the beta program.
 
 Backend:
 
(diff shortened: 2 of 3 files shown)


src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe Feature do
+  let(:user) { build(:confirmed_user, login: 'Tom') }
+  subject { Feature.active?('bootstrap') }
+
+  context 'with beta users' do
+    before do
+      allow(User).to receive(:possibly_nobody).and_return(user)
```
hennevogel: Why don't you use the login helper?

src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      allow(user).to receive(:in_beta?).and_return(true)
```
coolo: more importantly: why didn't you create a user with :in_beta trait :)

src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe Feature do
+  let(:user) { build(:confirmed_user, login: 'Tom') }
```
dmarcoux: Is there a reason for not creating this user?

DavidKang: not, really

src/api/spec/lib/feature_switch/feature_spec.rb
DavidKang: This helper is not accessible  under the spec/lib directory. So I decided to use the next aproach:
```
User.session = user
```

src/api/spec/lib/feature_switch/feature_spec.rb
DavidKang: changed

src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe Feature do
+  let(:user) { create(:confirmed_user, :in_beta, login: 'Tom') }
```
vpereira: Why do we have to call ```create``` and not just use ```build``` we dont need the user in the db.

src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  subject { Feature.active?('bootstrap') }
+
+  context 'with beta users' do
+    before do
+      User.session = user
```
vpereira: I don't see an advantage in set the ```session``` against stub it with ```allow``` any reason why?

src/api/spec/lib/feature_switch/feature_spec.rb
dmarcoux: True, I was wrong in my previous comment about this.

vpereira: ok, but I prefer to see this PR merged. so lets resolve it and approve as soon as everything is green.

PR #7751, opened 2019-06-13
Refactor feature_spec to avoid saving information in the db if not necessary

vpereira:
Avoid creating user in the db if necessary. Prefer to stub authentiction methods, than setting globals

### DIFF
--- a/src/api/spec/lib/feature_switch/feature_spec.rb
+++ b/src/api/spec/lib/feature_switch/feature_spec.rb
@@ -1,12 +1,12 @@
 require 'rails_helper'
 
 RSpec.describe Feature do
-  let(:user) { create(:confirmed_user, :in_beta, login: 'Tom') }
+  let(:user) { build(:confirmed_user, :in_beta, login: 'Tom') }
   subject { Feature.active?('bootstrap') }
 
   context 'with beta users' do
     before do
-      User.session = user
+      allow(User).to receive(:possibly_nobody).and_return(user)
       Feature.instance_variable_set(:@perform_initial_refresh_for_user, true)
     end
 
--- a/src/api/spec/lib/feature_switch/feature_spec.rb
+++ b/src/api/spec/lib/feature_switch/feature_spec.rb
@@ -1,12 +1,12 @@
 require 'rails_helper'
 
-RSpec.describe Feature do
+RSpec.describe Feature, type: :model do
   let(:user) { build(:confirmed_user, :in_beta, login: 'Tom') }
   subject { Feature.active?('bootstrap') }
 
   context 'with beta users' do
     before do
-      allow(User).to receive(:possibly_nobody).and_return(user)
+      login(user)
       Feature.instance_variable_set(:@perform_initial_refresh_for_user, true)
     end
 


src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -1,12 +1,12 @@
 require 'rails_helper'
 
 RSpec.describe Feature do
-  let(:user) { create(:confirmed_user, :in_beta, login: 'Tom') }
+  let(:user) { build(:confirmed_user, :in_beta, login: 'Tom') }
   subject { Feature.active?('bootstrap') }
 
   context 'with beta users' do
     before do
-      User.session = user
```
coolo: prefer use `user.run_as` then?

vpereira: That's a valid question. In a unit testing like that, I prefer to stub everything that isn't the Object Under Test. But definitely ```User#run_as``` is better than ```User.session```

dmarcoux: We would need to do `user.run_as` in both examples since it needs a block. I prefer `User.session = user`, but both options are fine.

coolo: stubing is very expensive in the future - as you're not testing the model, but a thin line of code and couple your tests very much to the code *below* your line.

My bible here: https://martinfowler.com/articles/mocksArentStubs.html#TestIsolation

Stubbing should always be last resort IMO

hennevogel: > I prefer to stub everything that isn't the Object Under Test

what's your reasoning for this?

vpereira: It really should be in our CONTRIBUTING guide to avoid this "Reductio ad Martinfowlerum". 

We are stubbing a method of a valid object, so we aren't mocking anything. **The user isn't important at all and to have to set it in the first place is the real coupling.**  Read it loud: "To test if the OBS is reading the file ```features.yml``` in the proper way, we have to set a ```User.session```"... IMO the user here is just a barrier to reach the code path that I want to test.  

> what's your reasoning for this?

 Because *UNIT TESTING* is a level of software testing where individual units/ components of a software are tested.  What is a unit? A unit is the smallest testable part of any software. It usually has one or a few inputs and usually a single output. 

If we don't make all the conditions deterministic to the point where we can just test what we want, we are doing again *INTEGRATION TESTING*, which  is a level of software testing where individual units are combined and tested as a group and we try to expose faults in the interaction between integrated units.

coolo: We can discuss the different schools of testing forth and back - I linked the fowler article exactly to avoid that, because he explains your *and* my point of view pretty clearly.

The real problem here is that the code under testing sucks: it 's a global function relying on global variables. And IMO to test it, you need to set these f*** global variables.

But perhaps it's better to actually move this *out* of the global function and where we want it: depending on the user object.

Because what `Feature.is_active?(:bootstrap)` means is: Should the _current user_ have bootstrap. So IMO it should actually express this and use `User.possibly_nobody.sees_feature?(:bootstrap)`.

coolo: ok, that won't work as we use multiple global feature functions.

vpereira: you can see here ```https://github.com/mgsnova/feature/blob/master/lib/feature.rb``` and in ```lib/feature_switch/obs_repository.rb```

coolo: https://github.com/openSUSE/open-build-service/pull/7759

coolo: > I prefer to try to pick the right tool to the right job.

Obviously what's right is subject of your school   :)

vpereira: > Obviously what's right is subject of your school :)

@coolo,

Not really. Convince me that to save the user to db and run a sql query to check if the user is in beta - which must be because we are creating it from a Factory (which should be unit tested too BTW) - is better than use a stub, and I'm full in your side.

coolo: it is better because you don't have to think about what to stub how, but can just use a standard mechanism in all tests. What you're trying to save here are sub-microsecond wins at the cost of considerable man minutes. And not only once, this would happen every time you write a test.

And writing tests shouldn't be an art - it should avoid regressions. At times invest a little more to save a little more in test suite executation, but the huge benefit of rspec tests is that we can run them in parallel - and concentrate on getting actual features implemented instead of arguing about build vs create.

vpereira: > And writing tests shouldn't be an art - it should avoid regressions. At times invest a little more to save a little more in test suite executation, but the huge benefit of rspec tests is that we can run them in parallel - and concentrate on getting actual features implemented instead of arguing about build vs create.

unit test should avoid regression in the unit being tested. Not cache an error somewhere else,  because a) somewhere else has already unit test and b) we have integration tests (we have too the minitests and the feature tests

So in the case here we have:

a ) User being created by a factory (The factory has a unit test, testing the trait, if it doesnt', it should be tested)

b) User model spec where you test  if the method ```in_beta?``` works as expected. 

to have to exercise all the stack, all components, every single time IMO doesn't make sense. For that we have already the minitest suite.

coolo: https://martinfowler.com/articles/mocksArentStubs.html#TestIsolation

hennevogel: Okay let me try to summarize:

1. building vs creating the factory is not something we argue about. If you can avoid `create`, great!
2. we argue about stubbing `User.possibly_nobody` vs. using the `login` or `User.session = user` method because some of us think stubbing should be a last resort.

I've pushed a commit with a proposal, what do you think?

vpereira: I personally don't like it. It's not a model and doing that way, IMO is a step towards a even tighter coupling. Please close this PR and continue on  #7759

hennevogel: >  It's not a model and doing that way, IMO is a step towards a even tighter coupling.

How so?

vpereira: your not just coupling the lib with an user via ```User.session = user``` but now with the model spec infrastructure

PR #7753, opened 2019-06-13
Display user email on user#show view only to logged-in users

### LINKED ISSUE #7744: anoymous webui shows email address of users
(both without login)
https://build-test.opensuse.org/user/show/bgeuken shows me an email address,
https://build.opensuse.org/user/show/bgeuken does not.

Showing email addresses is ... let's say retro anyway, but at least it should require a login.

dmarcoux:
Fixes #7744 

This is testable locally:
1. Add `return true` as the first line in [switch_to_webui2?](https://github.com/openSUSE/open-build-service/blob/e896f17e38d3ce5438e907486e0ff69acd4c7f3f/src/api/app/controllers/webui/webui_controller.rb#L305)
2. Visit http://localhost:3000/user/show/Admin as an anonymous user
3. Do not see the email of _Admin_ (as you can see in the screenshot below)

issue-7744

### DIFF
--- a/src/api/app/views/webui2/webui/user/_info.html.haml
+++ b/src/api/app/views/webui2/webui/user/_info.html.haml
@@ -11,7 +11,7 @@
       %span.badge.badge-secondary.align-top
         = title.upcase
 
-    - unless user.is_nobody?
+    - unless User.possibly_nobody.is_nobody?
       %p
         = mail_to user.email, title: "Send mail to #{user.name}" do
           %i.fas.fa-envelope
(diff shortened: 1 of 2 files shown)


src/api/app/views/webui2/webui/user/_info.html.haml
```diff
@@ -11,7 +11,7 @@
       %span.badge.badge-secondary.align-top
         = title.upcase
 
-    - unless user.is_nobody?
+    - unless User.possibly_nobody.is_nobody?
```
coolo: that's very long for `if User.session`

PR #7755, opened 2019-06-13
Fix bug on Request bugowner change link

### LINKED ISSUE #7616: "Request bugowner change" link doesn't work
# Issue Description
While migrating the `Search Owners` page to Bootstrap, I realized that the `Request bugowner change` link doesn't work. The modal enforces submitting either a user or a group. However, the controller requires both. There are hidden fields submitting the user and the group whenever they aren't filled up, but they contain the value `"False"` so it's failing since this isn't a valid user/group.

Expected Result
============
I expect that a request to change the bugowner will be created.

How to Reproduce
===============
1. Go to the `Search Owners` page
2. Enter a package (for example `ctris`)
3. Click on the `Request bugowner change` link
4. Fill the modal by typing in a user and a description
5. Submit the modal
6. See the error `Couldn't find Group 'False'`

saraycp:
In the Search Owner page there is a link to request a new bugowner for the package. The new bugowner can be a user or a group, not both. However, the action was always requiring both fields by mistake. Now only one is required and sent.

Fixes #7616 in both old and new UI.

Reuses the same JavaScript function used to add a reviewer to a requet, which has been slightly improved.

Screenshot-2019-6-13 Open Build Service

To test it locally:

1. Choose a package from your home project.
2. Go to the package's User tab.
3. Add a user as bugowner.
4. Go to the Search Owner page (http://localhost:3000/search/owner)
5. Search by the name of the package selected in step 1.
6. You should see some results. Now click one link 'Request bugowner change'.
7. Fill in the form.
8. A new request is created.

### DIFF
--- a/src/api/app/assets/javascripts/webui2/request.js
+++ b/src/api/app/assets/javascripts/webui2/request.js
@@ -58,15 +58,15 @@ function setupRequestDialog() { // jshint ignore:line
 
 function requestAddReviewAutocomplete() { // jshint ignore:line
   $('.modal').on('shown.bs.modal', function() {
-    $('.hideable input:not(:visible)').removeAttr('required');
+    $('.hideable input:not(:visible)').attr('disabled', true);
   });
 
   $('#review_type').change(function () {
     $('.hideable').addClass('d-none');
-    $('.hideable input:not(:visible)').removeAttr('required');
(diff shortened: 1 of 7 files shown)


coolo:
I'm just confused by the use of a checkbox icon for bullet lists. It looks like I can deselect, but I can't.

But this is not changed in this PR.

PR #7759, opened 2019-06-14
Rework the feature selection

coolo:
- Put production defaults into code, so left over configuration   files from earlier versions don't interfere with new features  in new releases (current config files left over will still overrule though). There were defaults for some features but not for all
- Do not monkey patch Feature class, but use our custom config loading class to merge beta features if configured
- Provide a beta switching function in Feature and use it from login

### DIFF
--- a/src/api/config/feature.yml
+++ b/src/api/config/feature.yml
@@ -1,30 +1,26 @@
-production:
-  features: &default
-    image_templates: true
-    kiwi_image_editor: false
-    cloud_upload: false
-    cloud_upload_azure: false
-    bootstrap: false
+# (keep aligned with lib/feature_switch/obs_repository.rb)
+#production:
+#  features:
+#   image_templates: true
+#   cloud_upload: false
+#   cloud_upload_azure: false
+#   bootstrap: false
+#   kiwi_image_editor: false
 
 development:
   features:
-    <<: *default
     kiwi_image_editor: true
     cloud_upload: true
     cloud_upload_azure: true
-    bootstrap: false
 
 test:
   features:
-    <<: *default
     kiwi_image_editor: true
     cloud_upload: true
     cloud_upload_azure: true
-    bootstrap: false
 
 beta:
   features:
-    <<: *default
     kiwi_image_editor: true
     cloud_upload: true
     cloud_upload_azure: true
--- a/src/api/config/feature.yml
+++ b/src/api/config/feature.yml
@@ -1,4 +1,3 @@
-# (keep aligned with lib/feature_switch/obs_repository.rb)
 #production:
 #  features:
 #   image_templates: true
--- a/src/api/lib/authenticator.rb
+++ b/src/api/lib/authenticator.rb
@@ -225,6 +225,7 @@ class Authenticator
     end
 
     User.session = @http_user
+    Feature.use_beta_features(@http_user.in_beta?)
 
     if @http_user.state == 'confirmed'
       Rails.logger.debug "USER found: #{@http_user.login}"
@@ -247,6 +248,7 @@ class Authenticator
   def load_nobody
     @http_user = User.find_nobody!
     User.session = @http_user
+    Feature.use_beta_features(false)
     @user_permissions = Suse::Permission.new(@http_user)
   end
 end
--- a/src/api/lib/feature_switch/obs_repository.rb
+++ b/src/api/lib/feature_switch/obs_repository.rb
@@ -5,17 +5,13 @@ module Feature
     class ObsRepository < YamlRepository
       DEFAULTS = {
         image_templates: true,
-        cloud_upload: false
-      }.freeze
+        cloud_upload: false,
+        cloud_upload_azure: false,
+        bootstrap: false,
+        kiwi_image_editor: false
+      }.with_indifferent_access
 
-      # Read given file, perform erb evaluation and yaml parsing
-      #
-      # @param file_name [String] the file name fo the yaml config
-      # @return [Hash]
-      #
-      def read_file(file_name)
-        super.with_indifferent_access
-      end
+      attr_accessor :use_beta_features
 
       # Extracts active features from given hash
       #
@@ -23,7 +19,12 @@ module Feature
       # @param selector [String] uses the value for this key as source of feature data
       #
       def get_active_features(data, selector)
-        data[@environment]['features'] = DEFAULTS.merge(data[@environment]['features'])
+        data[selector] ||= {}
+        features = data[selector].fetch('features', {})
+        if use_beta_features && data.dig('beta', 'features')
+          features.merge!(data['beta']['features'])
+        end
+        data[selector]['features'] = DEFAULTS.merge(features)
         super
       end
     end
--- a/src/api/lib/feature_switch/obs_repository.rb
+++ b/src/api/lib/feature_switch/obs_repository.rb
@@ -3,6 +3,7 @@ module Feature
     # ObsRepository for active and inactive features based on YamlRepository having default values for each key in OBS
     #
     class ObsRepository < YamlRepository
+      # remember to update comments in config/features.yml if changing
       DEFAULTS = {
         image_templates: true,
         cloud_upload: false,
--- a/src/api/spec/lib/feature_switch/feature_spec.rb
+++ b/src/api/spec/lib/feature_switch/feature_spec.rb
@@ -1,25 +1,35 @@
 require 'rails_helper'
 
 RSpec.describe Feature do
-  let(:user) { create(:confirmed_user, :in_beta, login: 'Tom') }
-  subject { Feature.active?('bootstrap') }
+  subject { Feature.active?(:bootstrap) }
+
+  let(:is_beta) { false }
+
+  before do
+    allow(YAML).to receive(:load).and_return(yaml)
+    Feature.use_beta_features(is_beta)
+    Feature.refresh!
+  end
+
+  context 'with true test settings' do
+    let(:yaml) { { 'test' => { 'features' => { 'bootstrap': true } } } }
+    it { is_expected.to be_truthy }
+  end
+
+  context 'with false test settings' do
+    let(:yaml) { { 'test' => { 'features' => { 'bootstrap': false } } } }
+    it { is_expected.to be_falsey }
+  end
 
   context 'with beta users' do
-    before do
-      User.session = user
-      Feature.instance_variable_set(:@perform_initial_refresh_for_user, true)
-    end
+    let(:is_beta) { true }
+    let(:yaml) { { 'beta' => { 'features' => { 'bootstrap': true } } } }
 
-    context 'when the file has "beta" key' do
-      it { expect(subject).to be_truthy }
-    end
+    it { is_expected.to be_truthy }
 
     context 'when the file doesn\'t have "beta" key' do
-      before do
-        allow(YAML).to receive(:load_file).and_return({})
-      end
-
-      it { expect(subject).to be_falsey }
+      let(:yaml) { {} }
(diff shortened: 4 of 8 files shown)


src/api/config/feature.yml
```diff
@@ -1,30 +1,20 @@
-production:
-  features: &default
-    image_templates: true
-    kiwi_image_editor: false
-    cloud_upload: false
-    cloud_upload_azure: false
-    bootstrap: false
+# production defaults are in lib/feature_switch/obs_repository.rb
+# you can overwrite them here though
```
hennevogel: no you can't. as we will update the package and you're changes are gone...

also expecting admins to change ruby code to configure the app is not what we want.

coolo: the config file is config(noreplace) in the package and that's where admin should configure it. I'm not saying admins should change ruby code - only to read if they care. If the comment is misleading, do you have suggestions how to make it clearer?

coolo: repeat the defaults as comment?

coolo: I did that now - please check if that fits your expectations

src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -1,22 +1,25 @@
...
-  subject { Feature.active?('bootstrap') }
+  subject { Feature.active?(:bootstrap) }
 
   context 'with beta users' do
-    before do
-      User.session = user
-      Feature.instance_variable_set(:@perform_initial_refresh_for_user, true)
+    let(:beta) do
+      Feature.use_beta_features(true)
+      Feature.refresh!
```
vpereira: why assign it with let if you don't need it? Please add it to the ```before``` block

src/api/lib/feature_switch/obs_repository.rb
```diff
@@ -3,27 +3,30 @@ module Repository
     # ObsRepository for active and inactive features based on YamlRepository having default values for each key in OBS
     #
     class ObsRepository < YamlRepository
+      # defaults need to be strings as we merge with yaml - which doesn't know symbols
       DEFAULTS = {
-        image_templates: true,
-        cloud_upload: false
+        'image_templates' => true,
```
vpereira: please use https://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html and then we can keep it as symbol. We can use symbol in the YAML and the ruby YAML parse is able to parse it too without problems... said that I think HashWithIndifferentAccess better.

src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -1,22 +1,25 @@
 require 'rails_helper'
+require 'yaml'
```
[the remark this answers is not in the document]
coolo: right. I was fighting allow(YAML) and thought it helps - but it didn't

src/api/spec/lib/feature_switch/feature_spec.rb
coolo: I looked for rspec guidance and it was preffered over functions - because that's what it is, a function that i need in 2 before blocks

vpereira: not sure where you are searching for rspec guidance. The better specs say:

"When you have to assign a variable instead of using a before block to create an instance variable, use let. Using let the variable lazy loads only when it is used the first time in the test and get cached until that specific test is finished. "  

which value are you assiging to beta? The method calls belong to ```before``` block

coolo: they are in 2 before blocks and so I needed a function - if you put them in the context's before block they are run in the wrong order. https://stackoverflow.com/a/17778760/11191489 - but I'll use a method to unconfuse you

src/api/lib/feature_switch/obs_repository.rb
```diff
@@ -3,27 +3,30 @@ module Repository
...
       # Extracts active features from given hash
       #
       # @param data [Hash] hash parsed from yaml file
       # @param selector [String] uses the value for this key as source of feature data
       #
       def get_active_features(data, selector)
-        data[@environment]['features'] = DEFAULTS.merge(data[@environment]['features'])
+        data[selector] ||= {}
+        features = data[selector].fetch('features', {})
+        if use_beta_features && data.dig('beta', 'features')
```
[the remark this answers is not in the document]
coolo: no? because the order of the merge calls matter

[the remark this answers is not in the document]
vpereira: you assign ```features``` 3 times and merge it with itself. I'm sure we can write it better.. maybe add a unit test for the method and we can try to improve it later?

src/api/spec/lib/feature_switch/feature_spec.rb
vpereira: I'm not confused, I just want clean code. Please don't mark the reviews as resolved. If they are still not.

src/api/lib/feature_switch/obs_repository.rb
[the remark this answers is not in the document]
coolo: I extended the test

src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -1,25 +1,35 @@
...
 RSpec.describe Feature do
-  let(:user) { create(:confirmed_user, :in_beta, login: 'Tom') }
-  subject { Feature.active?('bootstrap') }
+  subject { Feature.active?(:bootstrap) }
+
+  let(:is_beta) { false }
+
+  before do
+    allow(YAML).to receive(:load).and_return(yaml)
+    Feature.use_beta_features(is_beta)
```
vpereira: Why is better than stub? Why not create an user and pass it as argument?  Just in case!!!

https://martinfowler.com/articles/mocksArentStubs.html#TestIsolation

src/api/config/feature.yml
```diff
@@ -1,30 +1,26 @@
...
-    cloud_upload_azure: false
-    bootstrap: false
+# (keep aligned with lib/feature_switch/obs_repository.rb)
+#production:
+#  features:
+#   image_templates: true
+#   cloud_upload: false
+#   cloud_upload_azure: false
+#   bootstrap: false
+#   kiwi_image_editor: false
```
hennevogel: Configuration happens in `.yml` files, period. This just won't fly, sorry.

src/api/lib/authenticator.rb
```diff
@@ -225,6 +225,7 @@ def check_extracted_user
     end
 
     User.session = @http_user
+    Feature.use_beta_features(@http_user.in_beta?)
```
hennevogel: I also do not like this here too much as the feature selection is now leaking into app code and spec code.

src/api/config/feature.yml
coolo: that's all good - it's just the defaults that is ruby. If admins want to configure, yaml is the way to go

src/api/lib/authenticator.rb
coolo: yeah, because one global state (active features) is intermixed with another global state (active user) - and feature gem wasn't made for that. It's actually 2 different feature sets that the user should pick. But that would mean duplicating the features API from within the user model.

Would you prefer that?

src/api/spec/lib/feature_switch/feature_spec.rb
```diff
@@ -1,25 +1,35 @@
 require 'rails_helper'
 
 RSpec.describe Feature do
-  let(:user) { create(:confirmed_user, :in_beta, login: 'Tom') }
-  subject { Feature.active?('bootstrap') }
+  subject { Feature.active?(:bootstrap) }
+
+  let(:is_beta) { false }
```
vpereira: isn't that just a different way of Mock?

coolo: it's a variable depending on context

src/api/lib/authenticator.rb
hennevogel: I prefer what we have. Extending the gem in `lib/`.

src/api/config/feature.yml
hennevogel: This is a file for admins. You tell them to keep it in sync with the ruby file.

src/api/spec/lib/feature_switch/feature_spec.rb
vpereira: which turns out to be exactly the same as stub the method ```is_beta?``` from the user, no?

src/api/lib/authenticator.rb
vpereira: We are coupling now the authenticator lib with the beta feature.

src/api/config/feature.yml
vpereira: where do you tell them @hennevogel ?

hennevogel: You don't tell them to keep configuration in sync with some source code file.

src/api/lib/authenticator.rb
hennevogel: > We are coupling now the authenticator lib with the beta feature.

What does this have to do with the authenticator?

src/api/spec/lib/feature_switch/feature_spec.rb
coolo: you confuse me.

This PR decouples the whole thing from the user model - it's adding an explicit switch to be called from controllers. It's basically a policy on what features to use - the features don't peek into a global variable anymore, but take what was told at the beginning.

src/api/config/feature.yml
coolo: that sync comment is for developers - I move it to the source file instead to keep in sync with the config file. That also makes more sense.

src/api/lib/authenticator.rb
vpereira: > What does this have to do with the authenticator?

This PR wiill couple it with the features. I'm just saying what it is doing, not that is right.

src/api/spec/lib/feature_switch/feature_spec.rb
vpereira: Actually you confuse me, I'm trying to understand your reasoning. In the PR https://github.com/openSUSE/open-build-service/pull/7751 I understood that you want always to test the models interacting with the SUT, and that tests should avoid regression. So if somebody delete the method `User#is_beta?` this test should be able to catch it. Will this spec be able to catch it and avoid this regression? Or using your [Bible](https://martinfowler.com/articles/mocksArentStubs.html#TestIsolation)  as reference: _"If you introduce a bug to a system with mockist testing, it will usually cause only tests whose SUT contains the bug to fail._"

coolo: but the Feature library has no relation to User model - that's what the whole thing is fixing.

If User#is_beta? exists or not does not matter to the feature switch - as long as someone calls use_beta_features with true at one point. It might be a global instance setting in the future - the feature library won't care.

Stubbing means hiding the layer *below* the code under test - but User is not there. So no, I don't need to create artificial dependencies in tests to avoid stubbing.

hennevogel:
handled in #7880 now... 

```


                               _                        
                            .-" l                       
                          .'   /                        
           _.------------'    :                         
      .-"""                    "-.                      
    .'                            "-.                   
 _.-'    __                          "-.                
(---"""  "        _                     "-.             
 """"""""-----.-"" `.                      `.           
               \ `.__;________               `.         
                `.:           """"---...       `.       
                  `                     """--..  \--""-.
                                              :   `.   :
                                               \    .-" 
                                                "-._l   
```

PR #7762, opened 2019-06-14
Make bootstrap default

DavidKang:
For 2.10 release, we plan to set the Bootstrap interface by default. For that, we had to adapt some part of the code base:
 - Put bootstrap as default in the feature switch
 - Replace support file `feature_bootstrap` by `feature_bento`
 - Adapt tests
 - Adapt Circle-CI configuration


.circleci/config.yml
```diff
@@ -192,7 +192,10 @@ jobs:
       - run: *bootstrap_old_test_suite
       - run:
           name: Run spider
-          command: cd src/api; bundle exec rake test:spider
+          command: |
+            cd src/api
+            export BENTO=1
+            bundle exec rake test:spider
```
hennevogel: Hm, why do we run the spider test with bento?

DavidKang: Because we never run the spider test with Bootstrap. Shall I adapt the spider test for it?

hennevogel: What would you need to adapt?

coolo: I'm interested in what needs to be adopted for that.

DavidKang: not much, for sure the checking of the [logo](ttps://github.com/openSUSE/open-build-service/blob/master/src/api/test/functional/webui/spider_test.rb#L108)

hennevogel: I'm all for having the spider test run with what we will use most. Not necessarily in this PR..

hennevogel:
Here we go!!!! :rocket:

Issue #7673, opened 2019-06-04
POST to _attribute does not save to backend


coolo:
Those calls update the attribute in the database, but don't leave a history entry in the backend:
`iosc api -X POST /source/SUSE:SLE-12-SP5:GA/_attribute/OSRT:Config -T sp5.config`

adrianschroeter:
2.10 backport please since this is data corruption. thanks

PR #7763, opened 2019-06-14
Fix attrib storage on changed value

adrianschroeter:
The backend got not update on changed attribute values leading to a broken history

Fixes https://github.com/openSUSE/open-build-service/issues/7673

### DIFF
--- a/src/api/app/models/attrib.rb
+++ b/src/api/app/models/attrib.rb
@@ -34,7 +34,7 @@ class Attrib < ApplicationRecord
            :validate_issues,
            :validate_allowed_values_for_attrib_type
 
-  after_commit :write_container_attributes, on: [:create, :destroy]
+  after_commit :write_container_attributes, on: [:create, :destroy, :update]
 
   #### Class methods using self. (public and then private)
   def self.find_by_container_and_fullname(container, fullname)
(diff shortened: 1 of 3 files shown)


src/api/test/functional/attributes_test.rb
```diff
@@ -379,9 +379,29 @@ def test_create_attributes_project
     assert_response :success
     assert_xml_tag(tag: 'revisionlist')
     revision = Xmlhash.parse(@response.body).elements('revision').last
-    assert_equal 'tom', revision['user']
+    assert_equal 'king', revision['user']
     srcmd5 = revision['srcmd5']
 
+    # check history on changed value only
```
coolo: sorry, but the test is called test_create_attributes_project - testing updates from within is the reason why developers loath the minitests :(

src/api/app/models/attrib.rb
```diff
@@ -34,7 +34,7 @@ class Attrib < ApplicationRecord
            :validate_issues,
            :validate_allowed_values_for_attrib_type
 
-  after_commit :write_container_attributes, on: [:create, :destroy]
+  after_commit :write_container_attributes, on: [:create, :destroy, :update]
```
coolo: this is still called explicitly in webui - and I had a reason back then, but I don't remember

vpereira: probably because it wasn't being called here?

src/api/test/functional/attributes_test.rb
vpereira: well, even if I agree with you, the same test destroy attributes to.. so I guess it's a general issue in the minitest test suite.. handling it as  nitpick.

PR #7768, opened 2019-06-15
Rework the feature selection

coolo:
- Do not monkey patch Feature class, but use our custom config
  loading class to merge beta features if configured
- Provide a beta switching function in Feature and use it from
  User.session=

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -244,9 +244,10 @@ class User < ApplicationRecord
     current && current.is_admin?
   end
 
-  # set the user as current session user (should be real user)
+  # set the user as current session user (should be real user or nil)
   def self.session=(user)
     Thread.current[:user] = user
+    Feature.use_beta_features(user.try(:in_beta) || false)
   end
 
   def self.get_default_admin
--- a/src/api/lib/feature_switch/feature.rb
+++ b/src/api/lib/feature_switch/feature.rb
@@ -1,17 +1,7 @@
 module Feature
-  @perform_initial_refresh_for_user = true
-
-  def self.active?(feature)
-    if User.possibly_nobody.in_beta?
-      # caching the feature.yml file, this is basically taken from Feature.active_features
-      if @auto_refresh || @perform_initial_refresh_for_user || (@next_refresh_after && Time.now > @next_refresh_after)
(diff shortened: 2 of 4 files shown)


coolo:
This is the same approach but possibly more acceptable - I left the cleanup of the defaults in config/features.yml for a later moment and made the feature switch when setting the user session. This leaks less, but makes the Feature gem a clear dependency of the user model.

And @vpereira before you ask: the unit test only tests the Feature class as used by user, not the integration into the User model. That is at the moment only implicitly tested in controller tests e.g. for kiwi_image that verify you get a 404 if you don't have the feature enabled.

vpereira:
> And @vpereira before you ask: the unit test only tests the Feature class as used by user, not the integration into the User model. That is at the moment only implicitly tested in controller tests e.g. for kiwi_image that verify you get a 404 if you don't have the feature enabled.

But does it make sense to couple the ```User.session=``` with ```Feature``` if an user just use the API?

src/api/app/models/user.rb
```diff
@@ -244,9 +244,10 @@ def self.admin_session?
     current && current.is_admin?
   end
 
-  # set the user as current session user (should be real user)
+  # set the user as current session user (should be real user or nil)
   def self.session=(user)
     Thread.current[:user] = user
+    Feature.use_beta_features(user.try(:in_beta) || false)
```
vpereira: We use ```User.session``` everywhere, specially models, model specs and API. I don't think this WebUI specific setting should be necessary for them..

coolo: what makes you think features are webui specific? It's just a question of time til (staging) API enables some features only for some users.

But even if not: The use_beta_features function only resets some variables. And User.session= is the right place IMO

hennevogel:
handled in #7880 now... 

```


                               _                        
                            .-" l                       
                          .'   /                        
           _.------------'    :                         
      .-"""                    "-.                      
    .'                            "-.                   
 _.-'    __                          "-.                
(---"""  "        _                     "-.             
 """"""""-----.-"" `.                      `.           
               \ `.__;________               `.         
                `.:           """"---...       `.       
                  `                     """--..  \--""-.
                                              :   `.   :
                                               \    .-" 
                                                "-._l   
```

Issue #7749, opened 2019-06-13
It should be forbidden to add relationships with nobody 


coolo:
Nobody doesn't do any good, so relationships need to be forbidden.

saraycp:
@coolo, could you please give an example?

coolo:
This should be impossible:
![](https://user-images.githubusercontent.com/1067203/59433298-82a10d80-8de9-11e9-9c12-f7eddf56ef18.png)

Nobody should have no relationship and no group membership (actually it shouldn't exist, but that's a too far stretched goal for now)

Relationship already validates:
```
validates :user, presence: {
```

But it needs to also validate that user is no nobody.

PR #7777, opened 2019-06-17
Avoid having associations with user _nodody_

saraycp:
For now, we ban _nobody_ to have a relationship with project or package (Relationship model) and also to be a member of a group or be a reviewer for a request.

A custom validation was added because it's going to be checked in more than one model.

Fixes: #7749

### DIFF
--- a/src/api/app/models/relationship.rb
+++ b/src/api/app/models/relationship.rb
@@ -36,6 +36,8 @@ class Relationship < ApplicationRecord
     message: 'User and group can not exist at the same time'
   }, if: proc { |relationship| relationship.group.present? }
 
+  validate :allowed_user
+
   # don't use "is not null" - it won't be in index
   scope :projects, -> { where.not(project_id: nil) }
   scope :packages, -> { where.not(package_id: nil) }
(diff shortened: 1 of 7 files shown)


src/api/app/models/relationship.rb
```diff
@@ -126,6 +128,13 @@ def check_global_role
     errors.add(:base,
                "global role #{role.title} is not allowed.")
   end
+
+  # NOTE: Adding a normal validation, the error doesn't reach the view due to
```
dmarcoux: So this explains why we cannot include `AllowedUserValidator` in this model? Just making sure I understand correctly.

dmarcoux: So this explains why we cannot include `AllowedUserValidator` in this model? Just making sure I understand correctly.

saraycp: Exactly!

saraycp: Adding the validation to this model it apparently do nothing (doesn't associate the user to the project/package) and shows a success flash message instead of the error.

PR #7778, opened 2019-06-17
[dist] Fixing our spec file license tag.

adrianschroeter:
OBS source can be either used under GPL version 2.0 _or_ under
GPL version 3.0 terms.

This is not a re-licensing, just a fix of the former tagging in spec
files. (GPL v2.0 only and v3.0 only is basically not fullfillable, due
to the fact that v2.0 and v3.0 are incompatible)

Checked with SUSE legal dept.

### DIFF
--- a/dist/obs-api-testsuite-rspec.spec
+++ b/dist/obs-api-testsuite-rspec.spec
@@ -20,7 +20,7 @@ Name:           obs-api-testsuite-rspec
 Version:        2.10~pre
 Release:        0
 Summary:        The Open Build Service -- RSpec test suite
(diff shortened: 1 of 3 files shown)


adrianschroeter:
@alexandergraul JFYI

coolo:
> GPL v2.0 only and v3.0 only is basically not fullfillable, due
> to the fact that v2.0 and v3.0 are incompatible

Well, not within the same binary - but you can very well have scripts as GPL3 and other scripts as GPL2 and share code as LGPL and javascript as Apache-2.0 and css as MIT... licenses are a mess :)

PR #7787, opened 2019-06-19
2.10 ReleaseNotes update

adrianschroeter:
- added backend parts
- moved systemd files to generic parts
- removed mentioning of /public routes. They are anyway defined as
  private and not part of official api.


adrianschroeter:
There is currently the "Terms of Service" api mentioned, which is still not documented in the api and was still crashing last time I looked.
 (so, either removing the line or fix it)

hennevogel:
> There is currently the "Terms of Service" api mentioned, which is still not documented in the api

It's here now #7789

> and was still crashing last time I looked.

You where using it wrong, uploading a unstructured text file instead of an XML file with content :)

saraycp:
@adrianschroeter according to the messages in your commits, I guess you want to introduce ReleaseNotes-2.11, which is missing. I can add it outside the `docs/` directory and check if it stops failing.

However, you also updated the text, so maybe we should backport those changes to branch `2.10`.

Am I right?

adrianschroeter:
@saraycp ReleaseNotes-2.11 got created now.
@henne that is the point, docu still lacking. come one, should be easy to do ....

DavidKang:
@adrianschroeter, which docu is missing? Terms of Service?

adrianschroeter:
On Dienstag, 2. Juli 2019, 08:54:43 UTC David Kang wrote:
> @adrianschroeter, which docu is missing? Terms of Service?

yep

-- 

Adrian Schroeter
email: adrian@suse.de

hennevogel:
@adrianschroeter #7789 is merged since 12 days

PR #7797, opened 2019-06-20
Disable W3C mode for Capybara in the openqa tests

vpereira:
Backport from necessary change to make openQA work again


dmarcoux:
Duplicate of #7795

PR #7798, opened 2019-06-20
Break lines on words, in the collapsible text component

### LINKED ISSUE #7796: word-break CSS is over-zealous
The package view in the OBS web UI has CSS setting `.obs-collapsible-textbox` to `word-break: break-all`.  This is way too over-zealous, as seen by this screenshot:

image

In contrast, changing it to `word-break: break-word` works very nicely:

image

and should still deal correctly with long package names and other long words.

dmarcoux:
Fixes #7796

I believe this will be fine since it's used only for comments and in descriptions of packages, projects, patchinfos and requests.


eduardoj:
> LGTM. WIP is frozen.

I retriggered WIP adding a coma in the title of the pull request.

PR #7802, opened 2019-06-21
Fix comments word break, use hyphens instead

### LINKED ISSUE #7653: Comments are not wrapped properly.
# Issue Description
In most languages, you would expect that text is wrapped on word boundaries rather than on character boundaries -- but OBS comments now are wrapped on character boundaries resulting in hard-to-read comments.

This boils down to a `word-break: break-all;` CSS rule.

Expected Result
============
Text should be broken on word boundaries.

How to Reproduce
===============
1. Make a comment on a SR.
2. Note that the text is wrapped at character boundaries instead of word boundaries.

issue example

saraycp:
Using `word-break: break-all;` the comment looks nice when we have long words, but it's true that breaking the word at any character is not a good idea. It looks like this:

Screenshot-2019-6-21  home Admin(4)

If we remove the `word-break: break-all;` CSS rule, the comment exceeds the area and looks like follows: 

Screenshot-2019-6-21  home Admin(1)

So, taking into account that with or without `word-break: break-all;` the result is not ideal... I propose to remove that rule and use `hyphens: auto` instead. It doesn't break the normal words but add a hyphen to the long ones.

Screenshot-2019-6-21  home Admin(2)

Fixes #7653

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/comments.scss
+++ b/src/api/app/assets/stylesheets/webui2/comments.scss
@@ -1,6 +1,6 @@
 #comments {
   .media .media-body {
-    @extend .text-word-break-all;
+    hyphens: auto;
 
     pre, code {
       @extend .p-2;


saraycp:
[Address Feedback (Sprint 60)](https://trello.com/c/h3HkOuTp/1171-address-feedback-sprint-60)

PR #7812, opened 2019-06-25
Allow to specify commit messages when changing files

### LINKED ISSUE #3179: Allow to specify commit messages when changing files
As a user, I'd like to specify commit messages when I change any file. (like on GitHub)

rhabacker:
Fixes #3179  (for 2.9 branch)

- The commit input field is disabled by default or by pressing undo
image

- After changing any text in the edit field, the commit input field can be edited
image

- After pressing **Save** the message is added to the revision
image

### DIFF
--- a/src/api/app/assets/javascripts/webui/application/cm2/use-codemirror.js
+++ b/src/api/app/assets/javascripts/webui/application/cm2/use-codemirror.js
@@ -26,10 +26,13 @@ function use_codemirror(id, read_only, mode) {
     editor.on('change', function (cm) {
       var changed = true;
       cm.updateHistory(cm);
-      if (cm.historySize().undo > 0)
+      if (cm.historySize().undo > 0) {
         $("#save_" + id).removeClass('inactive');
-      else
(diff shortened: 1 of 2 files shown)


rhabacker:
> Hakiri — Security warnings were found. 

This result is not related to this patch

rhabacker:
> The Travis CI build failed 
> Failures:
> 1) Projects creating packages in projects not owned by user, eg. global namespace as non-admin user

This error message does not seem to be related to this patch either.

hennevogel:
Sorry but we do not backport features...

hennevogel:
Can you please submit this to master?

rhabacker:
> Thank you for your contribution! bouquet
> 
> But sorry but we do not backport features, can you submit this to master instead? TIA!

Which is then ported back to 2.10 ? How long will it take for the public obs to update to 2.10 before this annoying problem is fixed?

hennevogel:
Let's see if we will back-port this to 2.10

Issue #7378, opened 2019-04-09
Please show the full remote project name in submissions


mschnitzer:
**Is your feature request related to a problem? Please describe.**
We often have the issue that maintainers submit from https://build.opensuse.org to our internal instance of the build service. Right now, the source project name is shortened.

For example, the project `openSUSE.org:network:samba:STABLE` becomes `openSUSE.org:n...:samba:STABLE`.

**Describe the solution you'd like**
Could you please show the full remote or even better: link to the remote build service so that I don't have to always handcraft the URL's myself when I want to go there? Otherwise we first need to click on a file, then get the 404 error message (as files from remote cannot be shown) and then we need to c&p the project name link.

**Describe alternatives you've considered**
-

**Additional context**
-

mschnitzer:
@adrianschroeter @hennevogel this is becoming more and more an issue. Could you please address this for the WebUI? I still need to work around on this many times which is quite annoying.

hennevogel:
Not right now, sorry.

PR #7846, opened 2019-07-02
Show link to remote projects and packages in requests

krauselukas:
Screenshot-2019-7-2 Request 12 (new)
If a request comes through an interconnect it is necessary to create the link manually or to search for the package/project on the other OBS instance in order to see the origin.

Related to https://github.com/openSUSE/open-build-service/issues/7378

### DIFF
--- a/src/api/app/helpers/webui/webui_helper.rb
+++ b/src/api/app/helpers/webui/webui_helper.rb
@@ -328,12 +328,44 @@ module Webui::WebuiHelper
                      class: 'project', title: opts[:project])
   end
 
+  def link_to_remote_project(remote_prj, opts)
+    remote_sub_prj = resolve_remote_project_name(opts[:project])
+    base_url = URI.join(remote_prj.remoteurl, '/').to_s
+    out = opts[:short] ? ''.html_safe : 'project '.html_safe
+    out + link_to(opts[:project], "#{base_url}project/show/#{remote_sub_prj}",
+                  class: 'project', title: opts[:project])
+  end
+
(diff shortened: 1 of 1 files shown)


saraycp:
@krauselukas please rebase this commit with master, now the flickering tests are skipped.

src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -328,12 +328,44 @@ def link_to_project(prj, opts)
                      class: 'project', title: opts[:project])
   end
 
+  def link_to_remote_project(remote_prj, opts)
+    remote_sub_prj = resolve_remote_project_name(opts[:project])
+    base_url = URI.join(remote_prj.remoteurl, '/').to_s
+    out = opts[:short] ? ''.html_safe : 'project '.html_safe
+    out + link_to(opts[:project], "#{base_url}project/show/#{remote_sub_prj}",
```
hennevogel: How do you know this works? This is the URL to the *API* which is not necessarily the same as the web ui url.

krauselukas: yes, it works with api.opensuse.org, but with some interconnects it might not work. Only thing I have in mind would be adding a column to the database with the webui url and using this one if available. Would that make sense?

hennevogel: It's already in `Configuration.obs_url` and there is already interconnect support for `Configuration`.

krauselukas:
I think that would be already sufficient for now https://github.com/openSUSE/open-build-service/pull/7889

PR #7847, opened 2019-07-02
Fix flickering test for 'kiwi-repositories-list' element

saraycp:
~~The element is still not there, use `find` to wait for it.~~

### DIFF
--- a/src/api/spec/bootstrap/features/webui/kiwi/images_spec.rb
+++ b/src/api/spec/bootstrap/features/webui/kiwi/images_spec.rb
@@ -75,6 +75,8 @@ RSpec.feature 'Bootstrap_Kiwi_Images', type: :feature, js: true, vcr: true do
       find('#image-name')
       click_link('Software')
 
+      wait_for_ajax
+
       within('#kiwi-repositories-list') do
         expect(page).to have_text("#{repository}@#{project}")
       end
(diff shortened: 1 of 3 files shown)


coolo:
I don't think so. within() waits just as much as find does. I think the problem is that the element is there, but the content is not filled as the ajax call is still pending.

So I'm afraid you need to avoid within as long as the page is not fully loaded. Not sure how to recognize if it's loaded though

saraycp:
@coolo you are right. It wasn't clear for me if `within` waited or not, at least I didn't find it in the official documentation. But I found it in other places, [here](https://www.coursera.org/lecture/photo-tourist-web-app-capstone/capybara-waits-K7l96) for example.

Now I'm having a look at [this article](https://thoughtbot.com/blog/automatically-wait-for-ajax-with-capybara).

saraycp:
The element is still not there, `within` waits, but only the maximum time set in the Capybara configuration: `Capybara.default_max_wait_time = 6`. However, it seems some Ajax calls are taking longer than that, so, following [this approach](https://thoughtbot.com/blog/automatically-wait-for-ajax-with-capybara) now it should wait for it.

saraycp:
We are going to skip the failing tests until we find the right solution for them.
PR to skip: #7857

coolo:
waiting for ajax is kind of retro - we have this everywhere in openqa where we don't have capybara. In capybara you really want to `expect` the proper state. I don't understand the page under testing well enough, but I guess 'the proper state' means having repositories in the list

PR #7850, opened 2019-07-03
Update smoke tests gems

eduardoj:
Using old versions of this gems prevent our smoke tests in openQA to be run.

See: https://openqa.opensuse.org/group_overview/62


dist/t/Gemfile.lock
```diff
@@ -47,4 +47,4 @@ DEPENDENCIES
   selenium-webdriver
 
 BUNDLED WITH
-   1.13.7
+   1.17.1
```
hennevogel: Sure this wont break it?

eduardoj: > Sure this wont break it?

Checking...

eduardoj: It doesn't break it. Checked running smoke tests locally without problem.

Added a commit to change the repository to add, from openSUSE Leap 42.3 to openSUSE Leap 15.1.

PR #7851, opened 2019-07-03
Fix saving of project meta if containing current user as maintainer

### LINKED ISSUE #7844: flags are ignored on project meta creation
As @lethliel found out we loose flags on project creation. I am relative sure that this is a regression and it should be fixed before 2.10.

Please note that this is only the case when creating a project, not when updating an existing one.

How to Reproduce
===============
```
1. osc api -X DELETE /source/home:adrianSuSE:test   # just to be sure
2. # cat w
<project name="home:adrianSuSE:test">
  <title>AdrianSuSE's Home Project</title>
  <description>My playground, just for me</description>
  <person userid="adrianSuSE" role="bugowner"/>
  <person userid="adrianSuSE" role="maintainer"/>
<build><disable/></build>

  <repository name="openSUSE_Leap_15.1">
    <path project="openSUSE:Leap:15.1" repository="standard"/>
    <arch>x86_64</arch>
  </repository>
</project>
3. osc api /source/home:adrianSuSE:test/_meta -T w
4. osc api /source/home:adrianSuSE:test/_meta

```
the `<build><disable/></build>` part is missing.

Further Information
===============

coolo:
If the uploaded XML contains the same relationship that the
source controller wants to create (current user as maintainer),
we get an invalid project and active record aborts the save!.

It took quite a while to debug this as we expected save! to throw
an exception for invalid projects. As this behaviour started
with rails 5.2.1 to trigger, we believe it is a Rails bug.

Fixes #7844

### DIFF
--- a/src/api/app/controllers/source_project_meta_controller.rb
+++ b/src/api/app/controllers/source_project_meta_controller.rb
@@ -70,7 +70,10 @@ class SourceProjectMetaController < SourceController
         project = Project.new(name: @project_name)
         project.update_from_xml!(@request_data)
         # FIXME3.0: don't modify send data
-        project.relationships.build(user: User.session!, role: Role.find_by_title!('maintainer'))
+        maintainer_role = Role.find_by_title!('maintainer')
+        unless project.relationships.any? { |relationship| relationship.user == User.session! && relationship.role == maintainer_role }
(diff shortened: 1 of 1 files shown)


Ana06:
Just in case someone is curious what the Rails bug is: https://github.com/rails/rails/issues/36591

PR #7852, opened 2019-07-03
[ci] wait for outdated repos while testing build

M0ses:
Without this commit, the test cases in

0030_create_and_build_package.ts

will break, even if the result is outdated.

This is a regression that was introduced with #9b9f3a873d .
As we changed from "osc r" to "osc r -v" the output has changed,
but the regex to detect outdated repos was never fixed

With this fix we ensure, not to check as long as we have
outdated results.

### DIFF
--- a/dist/t/osc/0030_create_and_build_package.ts
+++ b/dist/t/osc/0030_create_and_build_package.ts
@@ -107,6 +107,7 @@ sub wait_for_buildresults {
     my $recalculation = 0;
     my @result        = `osc r -v`;
     for my $line (@result) {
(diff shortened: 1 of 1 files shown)


DavidKang:
@M0ses, please, could you rebase with master to fix/skip the flickering test.

M0ses:
@DavidKang : There should be no conflict with master and I'm quite sure that the failed tests are a "false positive"

anyway: I retriggered the package build in obs testing project and if it fails we should merge anyway as I`m nearly 100% sure that this is a false positive.

PR #7858, opened 2019-07-04
Add migration for in_rollout field for users

saraycp:
Needed for Rollout Program, used to move new features out of Beta
Program but make it available to users by batches.


src/api/db/migrate/20190704072437_add_column_in_rollout_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddColumnInRolloutToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :in_rollout, :boolean, default: false, index: true
```
hennevogel: Now new users won't be `in_rollout` right?

PR #7588, opened 2019-05-17
Add contributors graph to user show

coolo:
Add a graph showing the contributions/activities of the last year, for now we show what's in the database:
 - requests created
 - comments made
 - reviews accepted/declined

We fine tuned the spread of the colours to look a little clever, taking 50th, 80th and 95th percentiles of the activity values.

screencapture-localhost-3000-user-show-coolo-2019-05-17-13_33_04
screencapture-localhost-3000-user-show-ammartinez-2019-05-17-13_33_14

(She picked the colours)

Fixes #6892

Issue #6892, opened 2019-02-01
Contributions graph in user profile page


Ana06:
It would nice to have a contributions graph in the user profile page in OBS, similar to what I have [in my profile](https://github.com/Ana06) in Github:

![image](https://user-images.githubusercontent.com/16052290/52124489-a37e7880-2629-11e9-8dd9-268de0794a59.png)

It is a nice way to check what someone did and to recognise contributions, which may also increased incontributions and the use of OBS. :bulb: 

This may need some design to decide what exactly need to be counted and how to display it.

Related to https://github.com/openSUSE/open-build-service/issues/931 and https://github.com/openSUSE/open-build-service/issues/428

jayvdb:
This would be very helpful for me.  I have company support to do some OBS packaging, but my boss wants some evidence I did stuff, and my accepted task list isnt public.

I see xhr happening on `/user/tasks` and `/project/requests/foo` , which might be suitable for building client side analytics.  Is there any documentation for those, or a better API for that type of info from the backend?

coolo:
@Ana06 forced me to put this here:

![screencapture-localhost-3000-user-show-coolo-2019-05-14-16_54_08](https://user-images.githubusercontent.com/1067203/57708150-27a0bd00-7669-11e9-9d39-8ea14ea872a7.png)
![screencapture-localhost-3000-user-show-dimstar-2019-05-14-16_54_47](https://user-images.githubusercontent.com/1067203/57708163-2d969e00-7669-11e9-987c-b163075db3c9.png)

PR #7862, opened 2019-07-04
Single day contributions

coolo:
The 'rosa dots' left the user curious what she did that specific day. So we solve that by expanding the day on click and provide details about the requests created, reviewed and the commits done that day. 

For comments we only show the number.

screencapture-localhost-3000-user-show-ammartinez-2019-07-04-14_00_53

screencapture-localhost-3000-user-show-coolo-2019-07-04-13_58_17


Ana06:
@vpereira 

> Not sure through which channel you are collecting user feedback, maybe would be better if they could create issues allowing us to discuss about the possible solutions and implementation details.

This is part of https://github.com/openSUSE/open-build-service/issues/6892, which as you see was open for a lot of time without not much feedback. But it is also allowed to provide feedback in the PRs.

> Would be possible to add test cases for the brand new methods?

we can write some tests, but most of this is the rendering in the view and we don't test views.

src/api/.haml-lint.yml
```diff
@@ -32,6 +32,10 @@ linters:
       - Style/IfUnlessModifier
       - Style/Next
       - Style/WhileUntilModifier
+    exclude: 
+      # the linter has trouble understanding the if/else of the view
+      - 'app/views/webui2/webui/user/_activity_date_commits.html.haml'
```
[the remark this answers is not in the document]
Ana06: there is a bug in haml-lint because it complains about something which doesn't happen. In my opinion, we should report it (we need to report it @coolo!) and exclude it until it is fixed instead of using work arounds.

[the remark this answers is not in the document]
vpereira: @Ana06 even if that's a fact, which you are probably right, the views have more logic than normal.. so would that possible to extract it to helpers?

vpereira:
> we can write some tests, but most of this is the rendering in the view and we don't test views.

@Ana06 

Not the methods from `User::Contributions`, right?

vpereira:
> @vpereira
> 
> > Not sure through which channel you are collecting user feedback, maybe would be better if they could create issues allowing us to discuss about the possible solutions and implementation details.
> 
> This is part of #6892, which as you see was open for a lot of time without not much feedback. But it is also allowed to provide feedback in the PRs.
> 
> > Would be possible to add test cases for the brand new methods?
> 
> we can write some tests, but most of this is the rendering in the view and we don't test views.

well the feedbacks happened on #7588, which we iterated extensively in our sprint review. But what concerns me is the "The 'rosa dots' left the user curious what she did that specific day." feedback which we are not seeing in any known channel. No action is required here from @coolo, just a comment.

coolo:
> > we can write some tests, but most of this is the rendering in the view and we don't test views.
> 
> @Ana06
> 
> Not the methods from `User::Contributions`, right?

For the record: I agree that this could use some coverage. Testing the user::contributions itself doesn't make so much sense though - it's only useful to understand that the view renders what we expect.

So I rather break with "we don't test views" than to test user::contributions in itself.

vpereira:
> For the record: I agree that this could use some coverage. Testing the user::contributions itself doesn't make so much sense though - it's only useful to understand that the view renders what we expect.
> 
> So I rather break with "we don't test views" than to test user::contributions in itself.

I'm not sure how "Would be possible to add test cases for the brand new methods?" could be interpreted as "please test views". 

The request changes are:

- Too much logic and ruby in the views - could we extract it to helpers and avoid just excluding it from the linter? 

- Add specs for the new methods added. No view test needed

coolo:
> * Add specs for the new methods added. No view test needed

Ok, not sure if you are intentionally not reading what I write, but the new functions are just for the view. Testing them in a unit test is a lot of work for very little gain - each of them are one liners.

If we move the code into helpers, we can make it a helper test - and I think that's a good way to cover them.

vpereira:
> Ok, not sure if you are intentionally not reading what I write, but the new functions are just for the view. Testing them in a unit test is a lot of work for very little gain - each of them are one liners.

I just don't see a point in discussing why testing is important. The method `commits_done_per_day` isn't a one-liner, and if you extract code from your views and you write a method where some transformation is happening, a test case will be necessary too.

adrianschroeter:
> We need to have this functionality optional for the OBS admin. There may be legal reasons why you can not enable such a functionality. So please don't merge without the admin config option.

Just clarified it, the config option must also include the already merge pink box.

hennevogel:
To clarify, @coolo  please put this behind an OPS switch in `config/options.yml`. Something like

```
# Disable contribution graph on the user home page
# user_contributions_on_home_page=false
```

Ana06:
@vpereira 

> so would be possible to extract some of that logic to helpers?

how do you propose to do this? The problem is that this is not really logic, but just rendering the information in a concrete format. @coolo I suggest we write a test for `activities_per_date`. @vpereira  would that be fine for you? is there anything else to address?


@adrianschroeter

> We need to have this functionality optional for the OBS admin. There may be legal reasons why you can not enable such a functionality. So please don't merge without the admin config option.

Although I agree that this should be done, it is not related to this PR. This PR is about single day contributions. The graph is already there and allowing to configure per instance if it should be shown, it is just a different feature. Adding more things into this PR, makes it more difficult to review.

vpereira:
> how do you propose to do this? The problem is that this is not really logic, but just rendering the information in a concrete format. @coolo I suggest we write a test for `activities_per_date`. @vpereira would that be fine for you? is there anything else to address?

The suggestion, is to refactor the view `_activity_date.html.haml` and others if necessary.. the section 
``` - count = activities_per_date[:requests_created].size - max_count
    - activities_per_date[:requests_created][0..(max_count - 1)].each do |number|
```
is almost the same, just changing the hash key, so it can be definitely refactored to one helper.. the same apply to other section too... please check the changed view and try to refactor repeated code to a helper when possible.. https://dev.to/wuz/stop-trying-to-be-so-dry-instead-write-everything-twice-wet-5g33

vpereira:
> To clarify, @coolo please put this behind an OPS switch in `config/options.yml`. Something like
> 
> ```
> # Disable contribution graph on the user home page
> # user_contributions_on_home_page=false
> ```

Maybe a different PR for that, which to stay in the safe side, should be merged before this one.

coolo:
> Maybe a different PR for that, which to stay in the safe side, should be merged before this one.

I can wait.

Ana06:
@vpereira 
 
> The suggestion, is to refactor the view _activity_date.html.haml and others if necessary.. the section
> 
>     - activities_per_date[:requests_created][0..(max_count - 1)].each do |number|
> 
> is almost the same in different areas, and just changing the hash key we could probably refactor it to a helper.... the same maybe applies to other section too... please check the changed view and try to refactor repeated code to a helper when possible.. https://dev.to/wuz/stop-trying-to-be-so-dry-instead-write-everything-twice-wet-5g33

I agree with you that repeating code is a bad idea (it is also one of the Rails principles). The problem here is that the code is not really the same. We render different things in slightly different ways and I don't think that changing one of them necessarily implies modifying the rest. I do no know how we should move it to helpers.

@openSUSE/open-build-service what happens with this PR? https://github.com/openSUSE/open-build-service/pull/7940 is already merged and this is a really cool feature in which @coolo and I put a lot of effort. It would be nice to merge it. Is there something else blocking it?  I can write a test for `activities_per_date` and rebase if this is going to be merged, but I do not want to invest more time on it with the feeling it will stay here for months (or even never be merged).

vpereira:
> @openSUSE/open-build-service what happens with this PR? #7940 is already merged and this is a really cool feature in which @coolo and I put a lot of effort. It would be nice to merge it. Is there something else blocking it? I can write a test for `activities_per_date` and rebase if this is going to be merged, but I do not want to invest more time on it with the feeling it will stay here for months (or even never be merged).

Hi @Ana06 there is no reason why we wouldn't merge it, however there are some change requests open and a merge conflict to be solved. So as usual, those issues should be tackled before merge it.

saraycp:
@coolo  @Ana06 There is already a merged PR #7940 to make the Contribution Graph optional. Feel free to rebase and go on with this.

src/api/app/views/webui2/webui/user/_activity.html.haml
```diff
@@ -21,4 +21,12 @@
                 classname = activity_classname(activity, percentiles)
 
               %td{ title: "#{current_day} - #{activity} contributions", class: classname }
-                &nbsp;
+                - if activity > 0
+                  = link_to(user_show_path(user, params: { date: current_day }), class: 'activity-link') do
+                    &nbsp;
```
vpereira: couldn't we use `table { empty-cells: show; }` and remove the `&nbsp` @coolo ?

coolo: I can try. I wasn't aware of that css

coolo: works

coolo: No, I was wrong. Without the &nbsp; you can't click the links. And if I add &nbsp; only to links, it will render the table very imbalanced. Will revert that commit

vpereira: ok, sorry for that.

coolo: No need to be sorry, I learned a new trick (that is just not applicable here :)

Ana06:
@openSUSE/open-build-service what is the state of this PR? Why hasn't it been merged?

coolo:
This would still benefit from some test coverage on the commits part

coolo:
Rebased and tested it - it still works.

While the test coverage could be improved, I don't really fancy investing there for a feature that is disabled in production.

hennevogel:
@vpereira are you okay with no tests?

vpereira:
> @vpereira are you okay with no tests?

I'm not okay with no tests, but if you are ok @hennevogel you can merge it without problems!

hennevogel:
alright, I'll write some tests...

danidoni:
I would like to merge this, If you give me access to this branch, I'll push some commits addressing @vpereira's feedback. Let's see if we can merge it soon.

danidoni:
> I would like to merge this, If you give me access to this branch, I'll push some commits addressing @vpereira's feedback. Let's see if we can merge it soon.

I was using https acess for the remote, that was why I couldn't push before :man_facepalming:

Ana06:
@danidoni great to see you are working on this :cupid: Looking forward to see it in build.opensuse.org!

dmarcoux:
> @danidoni great to see you are working on this cupid Looking forward to seeing it in build.opensuse.org!

You won't see this on build.opensuse.org since it's disabled due to legal reasons...

danidoni:
> @danidoni great to see you are working on this cupid Looking forward to seeing it in build.opensuse.org!

You're welcome!

src/api/app/controllers/webui/users_controller.rb
```diff
@@ -38,7 +38,9 @@ def show
     # move back to the monday before (make it up to 53 weeks)
     @first_day -= (@first_day.cwday - 1)
 
-    @activity_hash = User::Contributions.new(@displayed_user, @first_day).activity_hash
+    @activity_hash = UserYearlyContribution.new(@displayed_user, @first_day).call
```
vpereira, line 41: maybe change the name to something expressing the time range (in the case yearly)

src/api/app/services/user_daily_contribution.rb
```diff
@@ -0,0 +1,45 @@
...
+    user.requests_created.where('date(created_at) = ?', date).pluck(:number)
+  end
+
+  def number_of_comments_for_date
+    user.comments.where('date(created_at) = ?', date).count
+  end
+
+  def number_of_reviews_done_per_day
+    ReviewsFinder.new.completed_by_reviewer(user)
+                 .where('date(reviews.created_at) = ?', date)
```
vpereira, line 28: I'm not sure if we need to cast the date here, but would be nice if we are consistent. So either we cast `date` everywhere or we don't. Bonus would be a spec for that.

vpereira:
Why I cannot see the contribution charts in the mobile version? Didn't we discuss to have it responsive too @saraycp @DavidKang ?

src/api/app/controllers/webui/users_controller.rb
danidoni: Done

vpereira:
@danidoni please don't forget to squash the commits :muscle:

DavidKang:
@vpereira 
> Why I cannot see the contribution charts in the mobile version? Didn't we discuss to have it responsive too @saraycp @DavidKang ?

AFAIK, the decision was taken in this PR https://github.com/openSUSE/open-build-service/pull/10271, maybe you could found more info in the Trello card :slightly_smiling_face:

vpereira:
> AFAIK, the decision was taken in this PR #10271, maybe you could found more info in the Trello card

love our vibrant community! Thanks @DavidKang :)

saraycp (changes requested):
When I click on a pink square, the whole page gets refreshed and it doesn't go back to the right tab (Contributions). Could we avoid this?

single_contrib

src/api/app/services/user_daily_contribution.rb
danidoni: I've added the casts in all places where needed.

src/api/app/services/user_daily_contribution.rb
```diff
@@ -0,0 +1,46 @@
+class UserDailyContribution
+  attr_accessor :user, :date
+
+  def initialize(user, date)
+    @user = user
+    @date = date
+    @date = date.to_date if date.respond_to?(:to_date)
```
danidoni, line 7: This ensures we deal with a date not a date time

danidoni:
> @danidoni please don't forget to squash the commits muscle



> When I click on a pink square, the whole page gets refreshed and it doesn't go back to the right tab (Contributions). Could we avoid this?
> 
> single_contrib

Let's see what I can do about that

danidoni:
> @danidoni please don't forget to squash the commits muscle

I'll address the tab issue and then squash them all

danidoni:
> > When I click on a pink square, the whole page gets refreshed and it doesn't go back to the right tab (Contributions). Could we avoid this?
> 
> Let's see what I can do about that

Well, we talked offline with @saraycp and we agreed on fix that behaviour in another PR, so she will dismiss her `change request` so we can merge this.

vpereira (approved):
Thanks @danidoni for improving this code a little bit. I guess we will have to live with that.

PR #7865, opened 2019-07-04
Add tab indicator in the codemirror stylesheet

### LINKED ISSUE #7849: No tab display in web editor
When editing with the web editor, it is currently not displayed whether a file uses spaces or tabs, as may be the case with shell scripts, for example.

hellcp:
Fixes #7849


eduardoj:
To make it easier to review here is a direct link to an example of a file with tabs in the review app: https://obs-reviewlab.opensuse.org/hellcp-patch-1/package/view_file/home:Admin/test_tabs/test_tabs.txt

PR #7868, opened 2019-07-05
Create Rollout Program setup

saraycp:
When we want to move a feature out of the Beta Program because it's ready to go live, we can enable it for all the users at the same time or little by little to batches/groups of users. In case we choose the second option, we are talking about the **Rollout Program**.

We are going to run some tasks every couple of days/weeks to add batches of users to the Rollout Program and let them use the new feature.

We'll move all the users out of the Rollout Program every time we're going to publish a new Rollout Feature and then we will start adding them. New users coming from that time on, are going to be in rollout by default.


saraycp:
[Turn on bootstrap in batches](https://trello.com/c/NBR883Q8/1178-turn-on-bootstrap-in-batches)

src/api/spec/factories/users.rb
```diff
@@ -4,6 +4,7 @@
     realname { Faker::Name.first_name }
     sequence(:login) { |n| "user_#{n}" }
     password { 'buildservice' }
+    in_rollout { false }
```
hennevogel: isn't this the default?

src/api/spec/factories/users.rb
```diff
@@ -49,6 +54,13 @@
           create(:service_token, user: user)
         end
       end
+
+      factory :non_recently_logged_user do
```
hennevogel: Do we really need a factory for this if we only use it once?

vpereira: I think we can use this factory elsewhere too. I would keep it

src/api/spec/factories/users.rb
DavidKang: No, by default is `true`, because we wanted to have the new users  into the **Rollout Program**

vpereira: @hennevogel 

IMO to be deterministic the factory shouldn't rely on the default, unless the default state itself is being tested.

hennevogel: Why do we turn it around in the factory then?

src/api/spec/factories/users.rb
hennevogel: F**K THE FUTURE SIGN

src/api/spec/factories/users.rb
DavidKang: We don't want to run the rollout program in all the tests. If we want to test the rollout program IMO there should be a specific test for that.
Right now, if we don't add `in_rollout: false` in the factory all the tests will be running in the `new UI`, because the enabled feature in the rollout is `Bootstrap`. Apart from that we also would need to modify the tests for `ImageTemplate`, `CloudUpload` because we are overwritten the `Feature.run_with_desactivated` with the values that are in the [feature.yml](https://github.com/openSUSE/open-build-service/pull/7868/files/f14000af0c46a7a101356b0449706ef89d012fa6#diff-c340e923ae2721e6d51a1a5da72dc438R17) file.

src/api/spec/factories/users.rb
vpereira: is that a blocking for you @hennevogel ?

src/api/spec/factories/users.rb
vpereira: @hennevogel is that ok, or do you want to see a change here?

hennevogel: we've discussed this offline. @DavidKang is pushing a new version.

vpereira:
@hennevogel could you review it again?

src/api/spec/controllers/cloud/upload_jobs_controller_spec.rb
```diff
@@ -47,14 +47,16 @@
         HEREDOC
       end
 
+      before do
+        allow(Feature).to receive(:active?).with(:cloud_upload).and_return(true)
```
eduardoj: Nitpick: This `allow` sentence, that returns `true`, is repeated three more times later: lines 51, 94, 147 and 192. They can be removed if we add this line in the parent `before` section, after line 36.

DavidKang:
We discuss offline and we decide to close this PR and use a different gem

PR #7872, opened 2019-07-07
Initial changes to make OBS work on Fedora

Conan-Kudo:
This PR contains the initial changes necessary to get OBS to build+install on Fedora following the Fedora Packaging Guidelines.

The bulk of the basic changes were ported from @OlegGirko's OBS 2.9 port to Fedora, adapted for OBS 2.10/master. A couple of changes were added by myself as well.


Conan-Kudo:
@adrianschroeter, @hennevogel, @hellcp, this is the initial set of changes to make OBS more portable.

Conan-Kudo:
@adrianschroeter Could we get this backported into OBS 2.10?

mmohring:
@Conan-Kudo this is super cool.

OlegGirko:
Thank you @Conan-Kudo for starting this process.
I have some remarks about this PR.

* You've lost prefixes with square brackets for some of my commits' subjects (did you use `git am`?):

6ccc9397 [api] Use bundler_ext instead of bundler if Gemfile.in file present.
c559ae86 [api] Log server name to test log for backend log messages.
5b5fd8ea [backend] Make OBS_BACKEND_DATA_DIR variable configuration parameter.
b002cdca [backend] Set proper pathnames when installing backend files.
3b837ee8 [api] Use absolute pathnames for apidocs in "config/environment.rb".
b17ad944 [api] Make api log directory configurable.

* I was not intending to upstream first two commits:

6ccc9397 [api] Use bundler_ext instead of bundler if Gemfile.in file present.
c559ae86 [api] Log server name to test log for backend log messages.

The first one if Fedora-specific hack, so I was thinking that it should be a patch in Fedora source rpm. But if you think that it can be accepted upstream, let it be there.

The second one is not related to porting at all. I was using it to make my logs more understandable when debugging failed tests. Actually, this one would be beneficial for upstream, but it probably deserves a separate PR.

* Commits that are related to porting to Fedora, but not included in this PR:

OlegGirko/open-build-service@54487b7d [dist] Use configuration variables instead of hardcoded paths.
OlegGirko/open-build-service@c1ce740e [dist] Make tests respect INITDDIR and SBINDIR configuration parameters.

Although `INITDDIR` variable is not used anymore, making `BINDIR` and `SBINDIR` configurable would be nice. However, this is not essential for porting to Fedora anymore.

OlegGirko/open-build-service@2ffa7846 [backend] Fix cleanup in BSSched::EventSource::Directory test.

This is a bug, and this makes tests on Fedora fail (at least for OBS 2.9). Probably deserves a separate PR.

OlegGirko/open-build-service@41dfee13 [api] Specify mysqld pidfile explicitly for api test.

Not including this one causes tests on Fedora fail (for OBS 2.9).

OlegGirko/open-build-service@ef88e4cb [dist] Use python2 instead of python as interpreter.

There are more scripts that use `#!/usr/bin/python` (without explicit version). Having these scripts cause rpm build fail.

OlegGirko/open-build-service@44425400 [frontend] Require mini_mime gem explicitly.
OlegGirko/open-build-service@a4a1a811 [frontend] Require minimal version of sanitize, webmock, vcr gems.

Not including these two cause dependency problems in Fedora. The first one makes essential gem not included, the second one makes outdated gems that come with Fedora included (I build newer versions of these gems in my OBS project).

OlegGirko/open-build-service@11f464e3 [backend] Make chekconstraints in repserver not fail on empty payload.

This is also a bug causing tests fail. Probably deserves a separate PR.

OlegGirko/open-build-service@fa2bd3dc [ci] Make test backend crash more prominent.
OlegGirko/open-build-service@6890a383 [ci] Actually wait for scheduler when test backend is already started.

These two commits fix a serious race condition that was causing test failures for OBS 2.9. Probably need a separate PR.

Conan-Kudo:
@OlegGirko I did use `git am`... Ugh, I forgot about that. I'll fix that...

I left out the rubygem changes, because I didn't know how to evaluate porting those forward...

Conan-Kudo:
@OlegGirko As for the `python` interpreter thing, I'd rather port it to Python 3, since we're dropping the Python 2 stack right now.

OlegGirko:
No, `git am` is not very good idea. `git rebase` is much better.

Porting gem changes is straightforward:
1. rebase,
2. resolve conflicts.

There will definitely be conflicts in `Gemfile.lock`, so making manual changes equivalent to ones in original commits are necessary.

Anyway,let's wait for developers' reaction to this PR and see whether it will be accepted. Then I can prepare more PRs.

Conan-Kudo:
> The first one makes essential gem not included, the second one makes outdated gems that come with Fedora included (I build newer versions of these gems in my OBS project).

Why not just submit pull requests to update the gems in Fedora? That's generally a better strategy than forking rubygem packages...

Conan-Kudo:
> The first one makes essential gem not included, the second one makes outdated gems that come with Fedora included (I build newer versions of these gems in my OBS project).

@OlegGirko Why not just submit pull requests to update the gems in Fedora? That's generally a better strategy than forking rubygem packages...

OlegGirko:
> Why not just submit pull requests to update the gems in Fedora?

Look [here](https://obs.infoserver.lv/project/monitor/obs:server:fedora:2.9) and see how many packages are in this project. Most of them are Ruby gems.
If you think about making OBS included in Fedora, think about getting all these gems included in Fedora first. I'm not strong enough to do it, so I prefer using my own OBS server to build packages for myself (and others adventurous enough to use my packages) because it's easier for me. I don't even have a login in Fedora infrastructure.

src/api/config/application.rb
```diff
@@ -2,8 +2,14 @@
 
 require 'rails/all'
 
-# Assets should be precompiled for production (so we don't need the gems loaded then)
-Bundler.require(*Rails.groups(assets: ['development', 'test']))
+gemfile_in = File.expand_path('../Gemfile.in', __dir__)
```
hennevogel: Sorry but this distro packaging specific thing should be in a distro specifc patch in a distro specific package. We do can't support this in any way.

Conan-Kudo: So then, would you be okay with me deleting SUSE-isms that aren't useful for everyone else from OBS? For example, the creation of rcFOO links that don't work anywhere else and are completely unnecessary now that we're in systemd era?

Conan-Kudo: But yeah, I can remove that patch from this list...

Conan-Kudo: Dropped.

hennevogel: > So then, would you be okay with me deleting SUSE-isms that aren't useful for everyone else from OBS? For example, the creation of rcFOO links that don't work anywhere else and are completely unnecessary now that we're in systemd era?

Sure, I would be fine with this.

Conan-Kudo: Done.

adrianschroeter:
On Montag, 8. Juli 2019, 11:56:33 UTC Henne Vogelsang wrote:
> hennevogel commented on this pull request.
> 
> > @@ -2,8 +2,14 @@
> 
>  require 'rails/all'
> 
> -# Assets should be precompiled for production (so we don't need the gems
> loaded then) -Bundler.require(*Rails.groups(assets: ['development',
> 'test']))
> +gemfile_in = File.expand_path('../Gemfile.in', __dir__)
> 
> > So then, would you be okay with me deleting SUSE-isms that aren't useful
> > for everyone else from OBS? For example, the creation of rcFOO links that
> > don't work anywhere else and are completely unnecessary now that we're in
> > systemd era?
> Sure, I would be fine with this.

nah, we want to keep them.

But you can do if clauses testing for defined suse_version around it. So it 
won't be used anywhere else.

-- 

Adrian Schroeter
email: adrian@suse.de

Conan-Kudo:
@adrianschroeter That's done in https://github.com/openSUSE/open-build-service/pull/7872/commits/1fc9576e7dcbe14668217f8950551beb90e5a7ef. I also dropped `rcobs-api-support`, since `service(8)` can't handle systemd targets anyway...

hennevogel:
@M0ses can you please review all the Makefile changes?

Conan-Kudo:
@M0ses I've rebased this PR against current master...

Makefile.include
```diff
@@ -1,8 +1,15 @@
 INSTALL=/usr/bin/install
 OBS_BACKEND_PREFIX=/usr/lib/obs/server
+OBS_BACKEND_DATA_DIR=/srv/obs
+OBS_BACKEND_LOG_DIR=$(OBS_BACKEND_DATA_DIR)/log
+OBS_BACKEND_SERVICE_LOG_DIR=$(OBS_BACKEND_DATA_DIR)/service/log
 OBS_DOCUMENT_ROOT=/srv/www/obs
 OBS_API_PREFIX=$(OBS_DOCUMENT_ROOT)/api
+OBS_API_LOG_DIR=$(OBS_API_PREFIX)/log
```
M0ses: Introducing path variables to make the installation more consistent makes sense to me.

Makefile.include
```diff
@@ -1,8 +1,15 @@
...
 OBS_APIDOCS_PREFIX=$(OBS_DOCUMENT_ROOT)/docs
+OBS_SRCSERVER_PORT=5352
+OBS_REPOSERVER_PORT=5252
+OBS_SERVICESERVER_PORT=5152
```
M0ses: So far I see no reason to make ports configurable.

dist/Makefile
```diff
@@ -17,36 +17,74 @@ install: install_obsapisetup install_apache install_project_update install_logro
 
 install_obsapisetup: system_dirs
 	$(INSTALL) -m 755 setup-appliance.sh $(DESTDIR)$(OBS_BACKEND_PREFIX)/setup-appliance.sh
+	sed -i \
+	    -e 's|wwwrun\.www|$(APACHE_USER).$(APACHE_GROUP)|g' \
+	    -e 's|/srv/www/obs/api|$(OBS_API_PREFIX)|g' \
+	    -e 's|/srv/www/obs/overview|$(OBS_DOCUMENT_ROOT)/overview|g' \
+	    -e 's|/srv/obs|$(OBS_BACKEND_DATA_DIR)|g' \
+	    $(DESTDIR)$(OBS_BACKEND_PREFIX)/setup-appliance.sh
```
M0ses: setup-appliance.sh is a shell script and the values should be sourced from a config file or better detect the OS and set the according variables.

PLEASE SEE  ALSO https://github.com/M0ses/open-build-service/blob/fedora_packaging/dist/setup-appliance.sh#L603

dist/obs-server.spec
```diff
@@ -336,6 +336,16 @@ export DESTDIR=$RPM_BUILD_ROOT
...
+
+%if 0%{?suse_version}
+systemd_services="$(basename --multiple --suffix .service %{buildroot}%{_unitdir}/*.service)"
+for systemd_service in $systemd_services; do
+    if [[ $systemd_service != *"@"* ]]; then
+        ln -sf /usr/sbin/service %{buildroot}%{_sbindir}/rc${systemd_service}
+    fi
+done
+%endif
+
```
M0ses: This makes absolute sense to me. It`s a distro specific functionalty and should go into specfile.
+1

docs/api/validate_xml.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
M0ses: +1

src/api/Makefile
```diff
@@ -29,8 +37,8 @@ docs:
 	cp -av ../../docs/api/api $(DESTDIR)$(OBS_APIDOCS_PREFIX)
 	cp -av ../../docs/api/html/* $(DESTDIR)$(OBS_APIDOCS_PREFIX)/api/html/
 	ln -sf $(OBS_APIDOCS_PREFIX)/api/ $(DESTDIR)$(OBS_API_PREFIX)/public/schema
-	echo 'CONFIG["apidocs_location"] ||= File.expand_path("../docs/api/html/")' >> $(DESTDIR)$(OBS_API_PREFIX)/config/environment.rb
-	echo 'CONFIG["schema_location"] ||= File.expand_path("../docs/api/")' >> $(DESTDIR)$(OBS_API_PREFIX)/config/environment.rb
+	echo 'CONFIG["apidocs_location"] ||= File.expand_path("$(OBS_APIDOCS_PREFIX)/api/html/")' >> $(DESTDIR)$(OBS_API_PREFIX)/config/environment.rb
+	echo 'CONFIG["schema_location"] ||= File.expand_path("$(OBS_APIDOCS_PREFIX)/api/")' >> $(DESTDIR)$(OBS_API_PREFIX)/config/environment.rb
```
M0ses: I`m not 100% sure but I guess this will not work - single quotes in bash prevent expansion of variables.

src/api/Makefile
```diff
@@ -44,14 +52,18 @@ config: prepare_dirs
 	echo "# This is to prevent fdupes from hardlinking" >> $(DESTDIR)$(OBS_API_PREFIX)/config/database.yml
 	echo "# This is to prevent fdupes from hardlinking" >> $(DESTDIR)$(OBS_API_PREFIX)/config/options.yml
 	echo "# This is to prevent fdupes from hardlinking" >> $(DESTDIR)$(OBS_API_PREFIX)/config/thinking_sphinx.yml
+	sed -i \
+	    -e 's|5352|$(OBS_SRCSERVER_PORT)|' \
+	    $(DESTDIR)$(OBS_API_PREFIX)/config/options.yml \
+	    $(DESTDIR)$(OBS_API_PREFIX)/config/options.yml.example
```
M0ses: This looks dangerous to me - changing arbitrary numbers in a config file may lead to unexpected results!

src/api/script/start_test_backend
```diff
@@ -106,11 +106,12 @@ srcsrv = Thread.new do
     begin
       line = srcsrv_out.gets
       break if line.nil?
-      logger.debug line.strip unless line.empty?
+      logger.debug 'bs_srcserver: ' + line.strip unless line.empty?
```
M0ses: all of the debug output in this file looks good to me
+1

src/backend/Makefile
```diff
@@ -22,12 +22,31 @@ install: prepare_dirs install_data_dirs bs_config
...
 	ln -sf /usr/lib/build $(DESTDIR)$(OBS_BACKEND_PREFIX)/build 
+	sed -i \
+	    -e 's|/srv/www/obs/api|$(OBS_API_PREFIX)|' \
+	    -e 's|/srv/obs|$(OBS_BACKEND_DATA_DIR)|' \
+	    -e 's|5352|$(OBS_SRCSERVER_PORT)|' \
+	    -e 's|5252|$(OBS_REPOSERVER_PORT)|' \
+	    -e 's|5152|$(OBS_SERVICESERVER_PORT)|' \
+	    $(DESTDIR)$(OBS_BACKEND_PREFIX)/bs_* \
+	    $(DESTDIR)$(OBS_BACKEND_PREFIX)/*.pm \
+	    $(DESTDIR)$(OBS_BACKEND_PREFIX)/*.pm.template \
```
M0ses: replacing arbitrary strings in our daemons/template config with a sed command is IMHO a "no go"

M0ses:
Sorry for my late response on this PR, I created my own branch were I tried get working installation on Fedora 31. I also include all of the non-controversial commits (or at least this is my plan ;-))

https://github.com/M0ses/open-build-service/tree/fedora_packaging

there I created an install script

https://github.com/M0ses/open-build-service/blob/fedora_packaging/dist/install_rh.sh

which I can use in my KankuFile for Fedora:

https://github.com/M0ses/open-build-service/blob/fedora_packaging/KankuFile.fedora

but I think it could also be used in vagrant with some slight modifications/enhancements.

I also started packaging missing packages here

https://build.opensuse.org/project/show/home:M0ses:branches:OBS:Server:Unstable

So far at least the backend is coming up properly, but I see the following issues:

* gpg needs the option `--files-are-digests` which is not available in fedoras gpg (so I had replace/overwrite) - this should go upstream.
* frontend is not running because of several issues
  * mod_passenger included in fedora crashes with coredump on loading
  * so far I could not get a working installation of passenger when trying to install with bundler.

M0ses:
> > Why not just submit pull requests to update the gems in Fedora?
> 
> Look [here](https://obs.infoserver.lv/project/monitor/obs:server:fedora:2.9) and see how many packages are in this project. Most of them are Ruby gems.
> If you think about making OBS included in Fedora, think about getting all these gems included in Fedora first. I'm not strong enough to do it, so I prefer using my own OBS server to build packages for myself (and others adventurous enough to use my packages) because it's easier for me. I don't even have a login in Fedora infrastructure.

@OlegGirko @Conan-Kudo 

I think we should go with obs-bundle-gems here:
https://build.opensuse.org/package/view_file/OBS:Server:Unstable/obs-server/_service:obs_scm:obs-bundled-gems.spec?expand=1

Conan-Kudo:
> gpg needs the option --files-are-digests which is not available in fedoras gpg (so I had replace/overwrite) - this should go upstream.

We should have this, we've been carrying the patch in Fedora since 2015 at least... I see it here: https://src.fedoraproject.org/rpms/gnupg2/blob/master/f/gnupg-2.2.18-file-is-digest.patch

We use `obs-signd` for COPR, so it's definitely used.

Conan-Kudo:
> I think we should go with obs-bundle-gems here:
https://build.opensuse.org/package/view_file/OBS:Server:Unstable/obs-server/_service:obs_scm:obs-bundled-gems.spec?expand=1

At least initially, I think it's fine to use the bundled-gems source service. It's going to take a while for me to get all the rubygems in Fedora...

Makefile.include
OlegGirko: Ports should be configurable because they may conflict with other services.
For example, Fedora uses port 5252 for `milter-greylist`.

src/api/Makefile
OlegGirko: It works because these substitutions are performed by `make`, not `sh`.

src/api/Makefile
OlegGirko: It causes no problems for me so far.
But making these files generated from templates would be more preferable way.

OlegGirko:
@M0ses:

> I also started packaging missing packages here

I already have all missing packages for OBS 2.9 in my [OBS project](https://obs.infoserver.lv/project/show/obs:server:fedora:2.9). It works pretty well: the OBS server itself is running on Fedora.

But I didn't port OBS to Fedora 2.10 yet.

> * gpg needs the option --files-are-digests which is not available in fedoras gpg (so I had replace/overwrite) - this should go upstream.

Also, `gpg2` in Fedora is 2.2, and it's incompatible with 2.1 in the way it stores private keys. It uses its own private key storage and doesn't allow to specify public key file. I solve this problem by `obs-signd-gnupg2` package that has separate build of `gpg2` 2.1 that doesn't conflict with Fedora's `gpg2` 2.2.

> * mod_passenger included in fedora crashes with coredump on loading

This is because SELinux prevents `mod_passenger` from working. Either switch SELinux to permissive mode, or write a policy to make `mod_passenger` work.

Or pester somebody in Fedora Project to finally fix SELinux policy to allow `mod_passenger` working.

> * so far I could not get a working installation of passenger when trying to install with bundler

Don't use bundler. Use Fedora packages.

> I think we should go with obs-bundle-gems here

Bundlilng is against Fedora packaging policy and should be used when there is no other way.
This is not the case for OBS in Fedora: as you see, I can run properly packaged OBS without bundled gems.

Conan-Kudo:
> Also, gpg2 in Fedora is 2.2, and it's incompatible with 2.1 in the way it stores private keys. It uses its own private key storage and doesn't allow to specify public key file. I solve this problem by obs-signd-gnupg2 package that has separate build of gpg2 2.1 that doesn't conflict with Fedora's gpg2 2.2.

@mlschroe mentioned to me once that obs-signd is supposed to be compatible with gnupg2 v2.2. The fact that Fedora COPR is able to sign packages with signd seems to imply this is true.

OlegGirko:
@Conan-Kudo:
> > Also, gpg2 in Fedora is 2.2, and it's incompatible with 2.1 in the way it stores private keys. It uses its own private key storage and doesn't allow to specify public key file. I solve this problem by obs-signd-gnupg2 package that has separate build of gpg2 2.1 that doesn't conflict with Fedora's gpg2 2.2.
> 
> @mlschroe mentioned to me once that obs-signd is supposed to be compatible with gnupg2 v2.2. The fact that Fedora COPR is able to sign packages with signd seems to imply this is true.

No, it's not compatible with gpg2 2.2. Attempt to create a private key results in error `gpg did not create private key`. This happens because `signd` creates params file for batch key generation containing `%secring` directive and expects secret keyring to appear there after the key is created, but gpg2 2.2 just ignores `%secring` directive and uses its own private key storage.

OlegGirko:
> No, it's not compatible with gpg2 2.2. Attempt to create a private key results in error `gpg did not create private key`. This happens because `signd` creates params file for batch key generation containing `%secring` directive and expects secret keyring to appear there after the key is created, but gpg2 2.2 just ignores `%secring` directive and uses its own private key storage.

Oops, seems like I was using old version of `signd`. It was adapted to gpg 2.2 in openSUSE/obs-sign@2d62ca6d8f86daed231f59a3eb4a5095e030227e.

Conan-Kudo:
The version of signd in Fedora _is_ a bit old for git master OBS, I've sent a PR to update it: https://src.fedoraproject.org/rpms/obs-signd/pull-request/3

OlegGirko:
I mean, I was using old version of `signd` when I encountered the problem with missing secring. I've just checked and found that current version in Fedora (although, not the latest one) doesn't have this problem.

M0ses:
> We should have this, we've been carrying the patch in Fedora since 2015 at least... I see it here: https://src.fedoraproject.org/rpms/gnupg2/blob/master/f/gnupg-2.2.18-file-is-digest.patch
> 
> We use `obs-signd` for COPR, so it's definitely used.

@Conan-Kudo 

Hmm, I stumbled over it when I tried to build obs-signd for fedora

```
[  102s] gpg: invalid option "--files-are-digests"
```

SEE here:

https://build.opensuse.org/package/live_build_log/home:M0ses:testing/obs-signd/Fedora_31/x86_64

I don`t know why it works for COPR but it looks like OBS is choosing the right package:

```
# ~/obs/home:M0ses:testing/obs-signd> osc buildinfo -d |grep gnupg
....
added gnupg2@Fedora:31/standard because of (direct):gpg
```

Conan-Kudo:
@M0ses Apparently the flag is `--file-is-digest`, not `--files-are-digests`. Anyway, you should get a fresh obs-signd package from Fedora adapted for this if you turn on the updates repo in your OBS project.

M0ses:
@Conan-Kudo 

Please have a look into 

https://github.com/openSUSE/obs-sign/blob/master/gnupg-1.4.7-files_are_digests.patch

https://build.opensuse.org/package/view_file/security:privacy/gpg2/gnupg-2.2.8-files-are-digests.patch?expand=1

the option used by obs-signd is "--files-are-digests"

Conan-Kudo:
@M0ses The Fedora package for `obs-signd` has a patch to change the parameter: https://src.fedoraproject.org/rpms/obs-signd/blob/master/f/0001-Rename-option-files-are-digests-to-file-is-digest.patch

Conan-Kudo:
@M0ses I've shipped an updated `obs-signd` to Fedora 31 as an update that is compatible with OBS git master: https://bodhi.fedoraproject.org/updates/FEDORA-2020-396d51fdbe

You should be able to use this with OBS with no problems.

M0ses:
@Conan-Kudo 

THX - I`m still working on a prototype for Fedora 31 based on obs-server unstable, but to make some progress with you PR I will split out the (for me) uncontroversial parts

Beginning with:
https://github.com/openSUSE/open-build-service/pull/8928

PR #7881, opened 2019-07-10
Create rake tasks for rollout program

DavidKang:
When we want to move a feature out of the Beta Program because it's ready to go live, we can enable it for all the users at the same time or little by little to batches/groups of users. In case we choose the second option, we are talking about the Rollout Program.

We are going to run some tasks every couple of days/weeks to add batches of users to the Rollout Program and let them use the new feature.

We'll move all the users out of the Rollout Program every time we're going to publish a new Rollout Feature and then we will start adding them. New users coming from that time on, are going to be in rollout by default.

Co-authored-by: Saray Cabrera Padrón scabrerapadron@suse.de
Co-authored-by: Eduardo Navarro enavarro@suse.com

### DIFF
--- a/src/api/lib/tasks/rollout.rake
+++ b/src/api/lib/tasks/rollout.rake
@@ -1,0 +1,37 @@
+# rubocop:disable Rails/SkipsModelValidations
+namespace :rollout do
+  desc 'Move all the users to rollout program'
+  task :all_on do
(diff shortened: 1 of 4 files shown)


saraycp:
[Add flipper gem & initial setup](https://trello.com/c/QUlAB2wh/1183-add-flipper-gem-initial-setup)

PR #7885, opened 2019-07-10
Remove Image Templates from FeatureToggle

DavidKang:
As a finished feature, `image template` shouldn't be behind the FeatureToggle.

### DIFF
--- a/ReleaseNotes-2.11
+++ b/ReleaseNotes-2.11
@@ -42,4 +42,5 @@ Bugfixes:
 Intentional changes:
 ====================
 
-
+  Features enabled by default:
+    * Image Template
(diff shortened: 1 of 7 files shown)


hennevogel:
Release notes are missing

DavidKang:
@hennevogel, could you review it again? :smile_cat:

ReleaseNotes-2.11
```diff
@@ -42,4 +42,5 @@ Bugfixes:
 Intentional changes:
 ====================
 
-
+  * Next features are not going to be enabled/disabled by configuration anymore and they will be now available for all users:
```
hennevogel: The fact they they are not configured anymore is not really important. Only that it's there now by default. Something like

Features enabled by default:
- Image Templates

DavidKang: done!

PR #7887, opened 2019-07-10
Add the Flipper groups registry to its initializer

saraycp:
Flipper.register doesn't store anything in the database, so we need to
set the groups when initializing to make it persistent.


saraycp:
[Add flipper gem & initial setup](https://trello.com/c/QUlAB2wh/1183-add-flipper-gem-initial-setup)

PR #7889, opened 2019-07-11
Dont shorten request source path text in webui2

krauselukas:
Looking at the shortend/trimmed source path of a request
makes it sometimes difficult to identify the source, especially
if its through an interconnect (since its not a link to the source
in this case). Bootstrap handles long text quite well in this case.
So trimming the text after 40 characters is not necessary.

Before:
Screenshot-2019-7-11 Request 12 (new)(1)

After:
Screenshot-2019-7-11 Request 12 (new)

Related to #7378

This would be sufficient solution for now, until https://github.com/openSUSE/open-build-service/pull/7846 is ready :)

### DIFF
--- a/src/api/app/helpers/webui/request_helper.rb
+++ b/src/api/app/helpers/webui/request_helper.rb
@@ -153,7 +153,7 @@ module Webui::RequestHelper
 
   # rubocop:disable Style/FormatStringToken, Style/FormatString
   def request_action_header(action, creator)
-    source_project_hash = { project: action[:sprj], package: action[:spkg] }
+    source_project_hash = { project: action[:sprj], package: action[:spkg], trim_to: nil }
 
     case action[:type]
     when :submit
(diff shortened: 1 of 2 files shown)


mschnitzer:
@krauselukas thank you for the implementation. This was quite annoying when dealing with remotes :-/

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -152,8 +152,8 @@ def hidden_review_payload(review)
   end
 
   # rubocop:disable Style/FormatStringToken, Style/FormatString
-  def request_action_header(action, creator)
-    source_project_hash = { project: action[:sprj], package: action[:spkg] }
+  def request_action_header(action, creator, trim_to = 40)
```
hennevogel: I don't get it. We use this helper once, without trimming. Why make this an option?

hennevogel: @krauselukas ping

krauselukas: @hennevogel back than I didn't wanted to change it in the old UI, but now since its dropped it doesn't make sense, I will change it :)

hennevogel:
coveralls hangs. everything is fine, see codecov. merging

PR #7890, opened 2019-07-11
Bootstrap to Flipper

saraycp:
Uses Flipper instead of Feature to handle Bootstrap feature.


src/api/app/controllers/application_controller.rb
```diff
@@ -89,6 +89,13 @@ def self.validate_action(opt)
 
   protected
 
+  def feature_enabled?(feature)
+    return if Flipper.enabled?(feature.to_sym, User.possibly_nobody) || Flipper.enabled?(feature.to_sym)
```
DavidKang: Why `User.possibly_nobody`? IMO we should user `User.session` that is not taking into account `_nobody_` user.

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -378,4 +385,8 @@ def set_influxdb_additional_tags
   def set_webui2_views
     prepend_view_path('app/views/webui2')
   end
+
+  def enabled_feature?(feature)
```
hennevogel: `feature_enabled` & `enabled_feature` confuse me..

src/api/spec/lib/feature_toggle/flipper_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    context 'with user both in beta and rollout' do
+      let(:user) { create(:confirmed_user, :in_beta, login: 'Tom') }
+
+      before do
+        User.session = user
+      end
+
+      it { expect(subject).to be_truthy }
```
DavidKang: This shouldn't be falsey? :thinking:

src/api/app/controllers/application_controller.rb
```diff
@@ -89,6 +89,13 @@ def self.validate_action(opt)
...
+    render file: Rails.root.join('public/404'), status: :not_found, layout: false
+  end
+
+  # TODO: remove when all the migration from Feature to Flipper is finished.
+  # It'll be replaced by feature_enabled?
   def feature_active?(feature)
     return if Feature.active?(feature)
     render file: Rails.root.join('public/404'), status: :not_found, layout: false
```
vpereira: why it cannot return simply boolean without render?

saraycp:
[Move bootstrap from the old feature gem to flipper](https://trello.com/c/nhbtzglL/1184-move-bootstrap-from-the-old-feature-gem-to-flipper)

src/api/app/controllers/application_controller.rb
DavidKang: @vpereira, We use this method as a `before_action` into different controllers like `image_template` to make them not accessible.

saraycp: Exactly, useful when someone goes directly to a URL that only exists if the feature is active. But if the user is not allowed to access to the feature, he/she will see the 404.

src/api/app/controllers/application_controller.rb
saraycp: done!

saraycp:
@hennevogel please have another look.

Issue #7679, opened 2019-06-04
Problems adding additional repotype


rhabacker:
I tried to add an additional reptype "rpmdb" to bs_publish to allow direct access to Windows Installer and portable binary packages in rpm containers (e.g. https://build.opensuse.org/package/view_file/windows:mingw:win32/mingw32-dbus-1/mingw32-dbus-1/mingw32-dbus-1-dbus-1-installer.spec?expand=1) and got lost in the details of the huge publish() function (https://github.com/openSUSE/open-build-service/blob/f085f2d6e928342d5479b0b4967b4e34c74f06f8/src/backend/bs_publish#L1612).

While creating and deleting entire repositories is very easy (since only two additional small functions are required), this cannot be said for installing, removing and updating packages implemented in a single 1100-line function that covers all cases and repository types together. It's also not easy to see which part of this function is for which repo type, which is important to add new repo types completely.

To facilitate maintenance and further expansion, it would be nice if this function were refactoed. A variant would be: 
 
1. a common setup function for all repo types
2. single functions for publishing parts and repo types, similar to creating and deleting repositories
3. a common post-setup function.

This would make adding new repository types much easier.

dmarcoux:
@mlschroe, please have a look.

rhabacker:
If someone like to know how the "rpmdb" support is intended to work: 

1. The project need to have set Repotype: rpmdb in the project config

2. publishing a repository: a rpm database is created in the published directory.

3. unpublishing a repository: the database and all installed files are deleted. 

3. publishing new rpm packages:  they are installed with rpm -i using the published directory as database root dir.
Note: This case relates to packages, which are in the list to publish but not installed in the rpmdb.

4. unpublishing packages:  package to unpublish are removed by running rpm -e <package> with the published directory as database root dir.
Note: The "unpublishing a package" case relates to packages,  which are installed but not in the list of packages to publish.

5. updating packages: The related package is updated with rpm -U <package> 
Note: This case relates to packages, which are installed and are in the list of packages to publish, but with a different version.

Explained with the initial example in the repository after publishing there are the following files/dirs
```
 .../
  openSUSE_Leap_42.3/
    dbus-1-1.12.14-bin.7z
    dbus-1-1.12.14-setup.exe
    dbus-1-1.12.14-debug.7z
    dbus-1-1.12.14-src.7z
    var/ # rpmdb

```

rhabacker:
Is there any progress with this issue ? 

[...]

PR #7891, opened 2019-07-11
bs_publish: move out repository create stuff into separate function

rhabacker:
This refactoring provides the counterpart of deleterepo() to increase code readability.

see #7679

Please note that this is my first obs post in Perl and I am not 100% sure if I have always used the correct Perl syntax.


src/backend/bs_publish
```diff
@@ -1505,6 +1505,74 @@ sub clean_blobdir {
   rmdir($blobdir);
 }
 
+sub createrepo {
+  my ($projid, $repoid, $data) = @_;
+  my $prp = "$projid/$repoid";
+  print "    creating repository\n";
+  my ($extrep, $stageservers) = BSUrlmapper::get_extrep_stageservers($prp);
+  return unless $extrep;
```
adrianschroeter: you don't  create repo metadata anymore, if no stage server is defined?
This is not wanted....

rhabacker: You suggest to use `BSUrlmapper::get_extrep($prp)` here, as done in the updated patch ?

mlschroe:
Hey, where's the `xrepoid` handling? You can't simply delete such things...

mlschroe:
Oh, wait, it's there. The inline comments confused me.

mlschroe:
Please put `$extrep` in the data hash instead of calculating it again.

mlschroe:
And your `createrepo` is actually not the counterpart of `deleterepo`, the counterpart of `deleterepo` is in fact the `publish` function. Your `createrepo` is more like a `createmetadata` function.

rhabacker:
> And your `createrepo` is actually not the counterpart of `deleterepo`, the counterpart of `deleterepo` is in fact the `publish` function. Your `createrepo` is more like a `createmetadata` function.

There is a set of createrepo_xxx() and deleterepo_xxx() function called by this new function. Does this mean that these functions are also only creating meta data and should be renamed to createmetadata_xxx and deletemetadata_xxx ?

rhabacker:
Furthermore there are 
```
   createpatterns_ymp($extrep, $projid, $xrepoid, $data, $patterntype{'ymp'});
   deletepatterns_ymp($extrep, $projid, $xrepoid, $data);
```
inside these functions. Does your statement apply also to this ?

PR #7905, opened 2019-07-15
Add scopes in_beta and in_rollout to User model

vpereira:
Added scopes to filter users in beta and rollout and improved rails dialect for scope `all_without_nobody`


saraycp:
Useful, thanks! :bouquet:  Waiting for the checks to be merged!

PR #7915, opened 2019-07-16
Make more clear what the input is for

### LINKED ISSUE #7863: Improve commit message input with editor
It's not really clear what the input is for. It's also in quite an ugly place, pushing undo/redo buttons to the right.
Screenshot from 2019-07-04 15-11-28

A modal that opens after clicking on the save button?
An expander?
Something that isn't visible while editing, but only shows up after you are done with the edits.

Something like github maybe? Although it's annoying you need to scroll down for it...
Screenshot from 2019-07-04 15-14-53

danidoni:
Remove the "go to line" input control because it doesn't provide any additional
value, users can search for the line number using the browser anyway.

Fixes #7863

Screenshot from 2019-07-16 12-00-49

### DIFF
--- a/src/api/app/views/webui2/shared/_editor.html.haml
+++ b/src/api/app/views/webui2/shared/_editor.html.haml
@@ -7,12 +7,8 @@
 %div{ class: "card #{'editable' unless style[:read_only]}" }
   - unless style[:read_only]
     .sticky-top.bg-light{ id: "top_#{uid}" }
-      .card-header.d-flex.justify-content-between
-        .input-group.input-group-sm
-          %button.btn.btn-danger.btn-sm.save{ id: "save_#{uid}", disabled: true }
-            %i.fa.fa-save
-            Save
-          %input.form-control.form-control-sm{ id: "comment_#{uid}", type: 'text', size: '40', disabled: true }
+      .card-header
+        .d-flex.justify-content-between
           .btn-group.pl-2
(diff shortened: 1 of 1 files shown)


DavidKang:
@danidoni, could you add a screenshot? please :smile_cat:

src/api/app/views/webui2/shared/_editor.html.haml
```diff
@@ -7,21 +7,18 @@
...
-              %i.fa.fa-undo
-              Undo
-            %button.btn.btn-secondary.btn-sm{ id: "redo_#{uid}", onclick: "editors[#{uid}].Redo(this);", disabled: true }
-              %i.fa.fa-redo
-              Redo
-        %button.btn.btn-warning.btn-sm{ data: { 'target': "#settings#{uid}", 'toggle': 'modal' } }
-          %i.fa.fa-cog
+      .card-header
+        .d-flex.justify-content-between
+          .input-group.input-group-sm
```
DavidKang: Theses classes are needed? :thinking:. We don't have any input here

DavidKang: Theses classes are needed? :thinking:. We don't have any input here

danidoni: They are not. I'll remove them!

PR #7922, opened 2019-07-16
Change favicon in live log dynamically

hellcp:
Changes favicon to match the current build status
Screenshot from 2019-07-16 20-12-41
Screenshot from 2019-07-16 20-12-46
Screenshot from 2019-07-16 20-12-54


src/api/app/views/webui2/webui/package/live_build_log.html.haml
```diff
@@ -35,5 +35,8 @@
       = render partial: 'live_build_log_controls'
 
 :javascript
+  faviconFinished = '#{path_to_image('favicon-warning.ico')}'
+  faviconRunning = '#{path_to_image('favicon-info.ico')}'
+  faviconPaused = '#{path_to_image('favicon.ico')}'
```
danidoni: How about passing all those paths as new parameters in the LiveLog constructor below? That way, there's no need to define global variables and the jslinter will not complain.

Please, remember to make the same change in the previous `live_build_log.html.erb`.

hellcp: Good point, sounds saner than the current way :smile:

hennevogel: @hellcp do you think you find the time to go over it?

hennevogel:
coveralls hangs, see codecov. merging.

hennevogel:
@hellcp awesome!

Issue #7799, opened 2019-06-20
Single CodeMirror instances are too small


dmarcoux:
# Issue Description
Following the changes in #7790, all CodeMirror instances are rather small. This isn't an issue on a page which has multiple instances, such as the request overview page. However, for the package file page, which has a single CodeMirror instance, this is quite annoying.

Expected Result
============
I expect a file to be completely displayed in a CodeMirror instance when it's the only one on a page.

How to Reproduce
===============

1. Be part of the beta program
2. Go to a page which has only one CodeMirror instance (example: https://build.opensuse.org/package/view_file/openSUSE:Factory/helm-mirror/helm-mirror.spec?expand=1)
3. See small CodeMirror instance

Further Information
===============

Bento:
![Screenshot_2019-06-20 openSUSE Build Service](https://user-images.githubusercontent.com/1102934/59875551-55340100-93a1-11e9-841c-5f2599f3194e.png)

Bootstrap:
![bootstrap](https://user-images.githubusercontent.com/1102934/59875559-582ef180-93a1-11e9-8de1-506e0bfd6ced.png)

dmarcoux:
This is also an issue on pages where multiple CodeMirror instances are present. However for pages which have multiple CodeMirror instances in read-only mode, like the request overview page (request#show), we shouldn't use CodeMirror to display diffs. We should use something like [diff2html](https://diff2html.xyz/).

hellcp:
The reason for this is https://github.com/openSUSE/open-build-service/pull/7790/files#diff-43ba76d11ecad92709bacf239b9e8c92L4 however I can tell that this won't work well if you want a few instances of codemirror with scrolling. The easiest solution I can see is adding h-100 to every instance where the editor should be full-height

dmarcoux:
> The reason for this is https://github.com/openSUSE/open-build-service/pull/7790/files#diff-43ba76d11ecad92709bacf239b9e8c92L4

Yep.

> The easiest solution I can see is adding h-100 to every instance where the editor should be full-height

This should work on pages which only have one instance. We could also apply `h-auto`.

dmarcoux:
I tried both proposals on `package#view_file` view, which has one CodeMirror instance. Here's [one example](https://build.opensuse.org/package/view_file/home:dmarcoux/rustup.rs/rustup.rs.spec?expand=1).

This is how the CodeMirror instance is displayed based on the class we use (in addition to `CodeMirror` and `cm-s-bootstrap`):
[...]

PR #7924, opened 2019-07-16
Minor editor fixes

hellcp:
Taking a little less space to perform the same action for submitting revisions:
Screenshot from 2019-07-16 23-07-36

Automatic height for editor, where necessary

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/cm2.scss
+++ b/src/api/app/assets/stylesheets/webui2/cm2.scss
@@ -8,3 +8,8 @@
 .short-progress {
     height: 5px;
 }
+
+.wide-edit {
+    min-width: 300px;
+    max-width: 1vw;
+}
--- a/src/api/app/assets/stylesheets/webui2/cm2.scss
+++ b/src/api/app/assets/stylesheets/webui2/cm2.scss
@@ -13,3 +13,7 @@
     min-width: 300px;
     max-width: 1vw;
 }
+
+.big-editor {
+    height: auto;
+}
--- a/src/api/app/views/webui2/shared/_editor.html.haml
+++ b/src/api/app/views/webui2/shared/_editor.html.haml
@@ -16,8 +16,23 @@
             %button.btn.btn-secondary.btn-sm{ id: "redo_#{uid}", onclick: "editors[#{uid}].Redo(this);", disabled: true }
               %i.fa.fa-redo
               Redo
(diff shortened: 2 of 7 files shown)


DavidKang:
@hellcp, thanks got your contribution.

Regarding to the proposal of the commit message, I found it a little bit confusing having 2 button that do the same in different place and at the same time they are too near from each other. Maybe the principal save button should be a dropdown called `Commit changes`. What do you think about it?

hennevogel:
> Taking a little less space to perform the same action for submitting revisions

And that is better why?

hellcp:
Editor's role is to edit text files, every other action is secondary.

hennevogel:
> Editor's role is to edit text files, every other action is secondary.

I don't agree. Committing the changes is equally important. Emphasizing adding a commit message will help collaboration in the app. I think the previous version is better and more clear.

coolo:
I'm with @hellcp on this one. The currently deployed version (I'm not sure what else went into master since) takes a lot of space for the message.

hennevogel:
the space is endless. saving space is not a goal and it especially doesn't trump usability and workflow concerns.

hellcp:
Screenshot from 2019-07-17 16-02-46
@danidoni maybe to enforce it better, just make the entire save button a dropdown, although that will require 2 clicks to save instead of 1. The issue with yet another button in the area will be width on mobile.

Unless obviously the button doesn't contain text and just an icon, then I would be fine with that:
Screenshot from 2019-07-17 16-07-45

DavidKang:
@hellcp, like @hennevogel says, the space here is not a problem. At the same time also avoid making two clicks for adding a commit message and also "encourage/remember" the user to add a message before saving. That is something good, from my point of view.

The approach, of having a single dropdown, and not with the text "Save", but maybe better "Commit changes" could be an option:

image

eduardoj:
@hellcp, thanks for the pull request! You second commit (812b9d64e2b306357fcbc7b7080fa7c9c4373c32) should fix #7799. We can benefit if we separate the commits in two pull requests. Most of the comments of this pull request are related to 90b7740237e9e0ba4e0d2211a0c3918528ddf6c2. Could you create another pull request, only with 812b9d64e2b306357fcbc7b7080fa7c9c4373c32? 

If you let me I can create the pull request with this commit for you.

hennevogel:
Alright, let's close this. Thanks for proposing this, even though we didn't take it in the end. Keep em coming :-)

PR #7934, opened 2019-07-17
Fix pre-filling of the Submit Request Form

### LINKED ISSUE #7551: Submit request form is not pre-filled with target project

danidoni:
After branching and submitting a package, the form was not pre-filled with the
project and package.

This fix, pre-fills the Submit when hitting the Project overview's
"Submit package" link.

Fixes #7551

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -10,6 +10,7 @@ class Webui::PackageController < Webui::WebuiController
 
   before_action :set_project, only: [:show, :index, :users, :linking_packages, :dependency, :binary, :binaries,
                                      :requests, :statistics, :commit, :revisions, :submit_request_dialog,
+                                     :branch_diff_info,
                                      :add_person, :add_group, :rdiff, :save_new,
                                      :save, :delete_dialog,
                                      :remove, :add_file, :save_file, :remove_file, :save_person,
@@ -20,6 +21,7 @@ class Webui::PackageController < Webui::WebuiController
 
   before_action :require_package, only: [:show, :linking_packages, :dependency, :binary, :binaries,
                                          :requests, :statistics, :commit, :revisions, :submit_request_dialog,
+                                         :branch_diff_info,
                                          :add_person, :add_group, :rdiff,
                                          :save, :save_meta, :delete_dialog,
                                          :remove, :add_file, :save_file, :remove_file, :save_person,
@@ -35,6 +37,7 @@ class Webui::PackageController < Webui::WebuiController
   before_action :check_ajax, only: [:update_build_log, :devel_project, :buildresult, :rpmlint_result]
   # make sure it's after the require_, it requires both
   before_action :require_login, except: [:show, :index, :linking_packages, :linking_packages, :dependency,
+                                         :branch_diff_info,
                                          :binary, :binaries, :users, :requests, :statistics, :commit,
                                          :revisions, :rdiff, :view_file, :live_build_log,
                                          :update_build_log, :devel_project, :buildresult, :rpmlint_result,
@@ -540,6 +543,10 @@ class Webui::PackageController < Webui::WebuiController
       return false
     end
     true
+  end
+
+  def branch_diff_info
+    switch_to_webui2
   end
 
   private :check_package_name_for_new
--- a/src/api/app/controllers/webui2/package_controller.rb
+++ b/src/api/app/controllers/webui2/package_controller.rb
@@ -25,4 +25,20 @@ module Webui2::PackageController
                                                       repository: @repository,
                                                       architecture: params[:arch]).results
   end
+
+  def webui2_branch_diff_info
+    linked_package = @package.backend_package.links_to
+    if linked_package
+      target_project = linked_package.project.name
+      target_package = linked_package.name
+      description = @package.commit_message(target_project, target_package)
+    end
+
+    render json: {
+      'targetProject': defined?(target_project) ? target_project : '',
+      'targetPackage': defined?(target_package) ? target_package : '',
+      'description': defined?(description) ? description : '',
+      'cleanupSource': @project.branch? # We should remove the package if this request is a branch
+    }
+  end
 end
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1535,6 +1535,12 @@ class Project < ApplicationRecord
     @missing_checks ||= calculate_missing_checks
   end
 
+  # This is not what makes a Package a branch, we only use this to prefill the submit request
+  # dialog in the UI. Please do not rely on this!
+  def branch?
+    name.include?(':branches:') # Rather ugly decision finding...
+  end
+
   private
 
   def discard_cache
--- a/src/api/app/assets/javascripts/webui2/request.js
+++ b/src/api/app/assets/javascripts/webui2/request.js
@@ -53,7 +53,31 @@ function setupRequestDialog() { // jshint ignore:line
     updateSupersedeAndDevelPackageDisplay();
   });
 
+  prefillSubmitRequestForm();
+
   updateSupersedeAndDevelPackageDisplay();
+}
+
+/*
+  This prefills the dialog with data coming from the package prefill endpoint.
+  The actual url has to be set in the link that shows the modal (coming as `e.relatedTarget`),
+  as a url data attribute.
+ */
+function prefillSubmitRequestForm() {
+  $('#submit-request-modal').on('show.bs.modal', function(e) {
+    $.ajax({
+      url: $(e.relatedTarget).data().url,
+      dataType: 'json',
+      contentType: 'application/json; charset=utf-8',
+      accept: 'application/json',
+      success: function (e) {
+        $('#targetpackage').attr('value', e.targetPackage);
+        $('#targetproject').attr('value', e.targetProject);
+        $('#description').val(e.description);
+        $('#sourceupdate').attr('checked', e.cleanupSource);
+      }
+    });
+  });
 }
 
 function requestAddReviewAutocomplete() { // jshint ignore:line
--- a/src/api/app/views/webui2/webui/package/_submit_request_dialog.html.haml
+++ b/src/api/app/views/webui2/webui/package/_submit_request_dialog.html.haml
@@ -19,30 +19,30 @@
               = text_field_tag(:sourceproject, project.name, disabled: true, class: 'form-control text-truncate')
             .form-group
               = render partial: 'webui/autocomplete', locals: { html_id: 'targetproject', label: 'To target project:',
-                                                                disabled: params[:readonly], value: target_project,
+                                                                disabled: params[:readonly], value: nil,
                                                                 data: { source: autocomplete_projects_path,
                                                                         requests_url: request_list_small_path,
                                                                         develpackage_url: package_devel_project_path } }
               - if params[:readonly]
-                = hidden_field_tag(:targetproject, target_project)
+                = hidden_field_tag(:targetproject, nil)
 
             .form-group
               = label_tag(:targetpackage, 'To target package:')
-              = text_field_tag(:targetpackage, target_package, disabled: params[:readonly], class: 'form-control')
+              = text_field_tag(:targetpackage, nil, disabled: params[:readonly], class: 'form-control')
               - if params[:readonly]
-                = hidden_field_tag(:targetpackage, target_package)
+                = hidden_field_tag(:targetpackage, nil)
 
             .form-group
               = label_tag(:description, 'Description:')
               %abbr.text-danger{ title: 'required' } *
-              ~ text_area_tag(:description, description, size: '40x3', class: 'form-control', required: true)
+              ~ text_area_tag(:description, nil, size: '40x3', class: 'form-control', required: true)
 
             .form-group.d-none#supersede-display
               = label_tag(:supersede_requests, 'Supersede requests:')
               #supersede-requests
 
             .custom-control.custom-checkbox#sourceupdate-display
-              = check_box_tag(:sourceupdate, 'cleanup', cleanup_source, class: 'custom-control-input')
+              = check_box_tag(:sourceupdate, 'cleanup', nil, class: 'custom-control-input')
               = label_tag(:sourceupdate, 'Remove local package if request is accepted', class: 'custom-control-label')
 
             %p.d-none#devel-project-warning
--- a/src/api/app/views/webui2/webui/package/bottom_actions/_submit_package.html.haml
+++ b/src/api/app/views/webui2/webui/package/bottom_actions/_submit_package.html.haml
@@ -1,4 +1,8 @@
 %li.list-inline-item
-  = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#submit-request-modal' }) do
+  = link_to('#',
+      class: 'nav-link',
+      data: { toggle: 'modal',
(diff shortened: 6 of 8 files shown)


danidoni:
[Address Feedback (Sprint 61)](https://trello.com/c/ZdIgWr3T/1179-address-feedback-sprint-61)

vpereira:
Hey @danidoni looks like you need to refresh the VCR. Sadly I couldn't find a wiki page to provide you. Maybe @eduardoj can help you to do it tomorrow.

danidoni:
> Hey @danidoni looks like you need to refresh the VCR. Sadly I couldn't find a wiki page to provide you. Maybe @eduardoj can help you to do it tomorrow.

Thanks! Let me take a look.

vpereira:
Hi @danidoni, I created a wiki to try to document how we work with *VCR*. https://github.com/openSUSE/open-build-service/wiki/Testing-with-VCR feel free to extend it if you feel that some bit of information was missing.

Something that is important to note is: As soon as you enabled the `global_write_through` you should run just the specific specs, limiting to specific line or context for example, that you want to have the cassette refreshed or generated. If you run for example the whole spec, you will endup updating cassettes that shouldn't be touched (which looks like its the case here)

danidoni:
> That's an awful lot of new cassettes and requests. Something isn't right.

Yes, that's what I'm afraid of... This is the outcome of setting the VCR configuration to `:new-episodes` to record the new requests.

src/api/app/controllers/webui2/package_controller.rb
```diff
@@ -25,4 +28,11 @@ def webui2_statistics
                                                       repository: @repository,
                                                       architecture: params[:arch]).results
   end
+
+  private
+
+  def backend_linked_project_package(package)
+    lt = package.backend_package.links_to
```
vpereira: maybe we could come up with a better name than `lt` here? `link_to_package` ?

src/api/app/controllers/webui2/package_controller.rb
```diff
@@ -1,5 +1,8 @@
 module Webui2::PackageController
   def webui2_show
+    @tprj, @tpkg = backend_linked_project_package(@package)
+    @tprj = params[:targetproject] if params[:targetproject] # allow to override by parameter
+    @tpkg = params[:targetpackage] if params[:targetpackage] # allow to override by parameter
```
vpereira: With the strong typing hat on: Are we storing in `@tprj` and `@tpkg` information of the same type?

danidoni:
Ouch, I closed the PR by mistake. Reopening

src/api/app/controllers/webui2/package_controller.rb
danidoni: Sure, I'm returning the project and package names, so what about `project_and_package_names`?

src/api/app/controllers/webui2/package_controller.rb
danidoni: Yes, both are strings if they are present

src/api/app/controllers/webui2/package_controller.rb
vpereira: great!

src/api/app/controllers/webui2/package_controller.rb
vpereira: ok

src/api/app/controllers/webui2/package_controller.rb
danidoni: Fixed

danidoni:
> I'm missing more things. This is how it looked in bento:
> 
> Screenshot from 2019-08-26 15-15-51

I'll check it out now.

danidoni:
For the record, @vpereira and me spoke offline and agreed that the generation of all those changes in the cassettes are indeed expected, because the feature specs visit the package show page, and pick up the new code that is calling the backend.

I'll check the ones that are returning 404 because I'm pretty sure those are wrong.

danidoni:
Updated the cassettes that were recorded with the 404 error codes. @vpereira check it out again please.

vpereira:
> Updated the cassettes that were recorded with the 404 error codes. @vpereira check it out again please.

for me it's ok, just @hennevogel mentioned that there are more fields that we can pre-fill https://github.com/openSUSE/open-build-service/pull/7934#pullrequestreview-279583704

hennevogel:
Any news here?

danidoni:
> Any news here?

Yes, I'm working on a commit to address the rest of the fields that were pre-filled in the old UI. I'll push it soon.

src/api/app/models/project.rb
```diff
@@ -1568,6 +1568,10 @@ def missing_checks
     @missing_checks ||= calculate_missing_checks
   end
 
+  def branch?
+    name.include?(':branches:')
```
hennevogel: Please also 'port' over the comment ;-) Otherwise people might start to use this for other things.

src/api/spec/cassettes/Packages/editing_package_files/editing_an_existing_file.yml
```diff
@@ -148,22 +150,94 @@ http_interactions:
...
+      Accept-Encoding:
+      - identity
+      Accept:
+      - "*/*"
+      User-Agent:
+      - Ruby
+  response:
+    status:
+      code: 404
+      message: _service  no such file
```
danidoni: I don't know how to fix this... Not the haziest clue. Any help would be very appreciated.

coolo: Upvote #5308 to ease test development would be my first advise

src/api/spec/cassettes/Packages/editing_package_files/editing_an_existing_file.yml
```diff
@@ -756,16 +1132,16 @@ http_interactions:
       Connection:
       - close
       Content-Length:
-      - '122'
+      - '309'
     body:
       encoding: UTF-8
       string: |
-        <status code="404">
-          <summary>_service: no such file</summary>
```
coolo: but you can see here that the 404 was there before - you just changed the order. And this is what puzzles me: tons of your  vcr additions are from 2018. Did you edit that or just fight with vcr madness?

src/api/spec/cassettes/Packages/editing_package_files/editing_an_existing_file.yml
vpereira: if it was a 404 before then it should be ok...

src/api/app/controllers/webui2/package_controller.rb
```diff
@@ -1,6 +1,11 @@
 module Webui2::PackageController
   def webui2_show
+    @tprj, @tpkg = project_and_package_names(@package)
```
hennevogel: Please do not use these awful variable names.

src/api/app/controllers/webui2/package_controller.rb
```diff
@@ -1,6 +1,11 @@
...
+    @tprj = params[:targetproject] if params[:targetproject] # allow to override by parameter
+    @tpkg = params[:targetpackage] if params[:targetpackage] # allow to override by parameter
```
hennevogel: Why? There is no one making use of this feature in our app.

src/api/app/controllers/webui2/package_controller.rb
```diff
@@ -25,4 +30,11 @@ def webui2_statistics
                                                       repository: @repository,
                                                       architecture: params[:arch]).results
   end
+
+  private
+
+  def project_and_package_names(package)
```
hennevogel: This belongs into `Package`.

src/api/app/models/project.rb
```diff
@@ -1535,6 +1535,10 @@ def missing_checks
     @missing_checks ||= calculate_missing_checks
   end
 
+  def branch?
```
hennevogel:
```suggestion
  # This is not what makes a Package a branch, we only use this to prefill the submit request
  # dialog in the UI. Please do not relay on this!
  def branch?
```

src/api/spec/cassettes/Packages/editing_package_files/editing_an_existing_file.yml
danidoni: > Upvote #5308 to ease test development would be my first advise

Hm, that would make sense, yeah...

danidoni: > if it was a 404 before then it should be ok...

Its true, didn't notice...

src/api/spec/cassettes/Packages/editing_package_files/editing_an_existing_file.yml
danidoni: Just fighting...

src/api/app/controllers/webui2/package_controller.rb
danidoni: I have some improvements here, but I wanted to fix the tests before introducing more changes

src/api/app/controllers/webui2/package_controller.rb
danidoni: Not even while using it from `osc`?

hennevogel: Using the webui from osc? Never happens.

danidoni: ok!

src/api/app/controllers/webui2/package_controller.rb
danidoni: Yes, so true.

dmarcoux:
Regarding the Rubocop offenses, `Style/StringLiterals` is dead simple. For `Lint/UselessAssignment`, I would go for this instead of using `||=`:

```
defined? target_project ? target_project : ''
```

Same thing for the package and description.

src/api/app/controllers/webui2/package_controller.rb
```diff
@@ -27,18 +27,21 @@ def webui2_statistics
...
     end
-    cleanup_source = @project.branch? # We should remove the package if this request is a branch
 
     render json: {
-      'targetProject': target_project ||= "",
-      'targetPackage': target_package ||= "",
-      'description': description ||= "",
+      'targetProject': target_project,
+      'targetPackage': target_package,
+      'description': description,
```
vpereira: I just fail to see why these temporary variable is better than call ` linked_package.project.name`,  `linked_package.name` and `@package.commit_message(target_project, target_package)`  direct in the render?

danidoni: Because using `linked_package.project.name` and friends should only be done when there is actually a `linked_package`. I could move the render inside the if and provide an else branch with another render with blank values, but then I'll be repeating the render block. It's a matter of taste which version to use I guess as both versions have pros and cons.

dmarcoux:
Regarding the `Style/UnneededCondition`, it's a false positive in this case. I will report it to upstream.

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -10,6 +10,7 @@ class Webui::PackageController < Webui::WebuiController
 
   before_action :set_project, only: [:show, :index, :users, :linking_packages, :dependency, :binary, :binaries,
                                      :requests, :statistics, :commit, :revisions, :submit_request_dialog,
+                                     :prefill,
```
hennevogel: :thought_balloon: I really don't like this naming. Should include some reference to what is actually prefilled. It's not the package...

danidoni: Hm, good point.

src/api/app/controllers/webui2/package_controller.rb
vpereira: yes, sure. But I think there are better ways to achieve it, other than initialize an empty string. Either you encapsulate this logic in method or render what you have and make sure in the view you validate if the data exists before to try to render it.

i.e: Add keys to a hash and render the hash as it is, and in the view you do a `hash.fetch(:key, $DEFAULT_VALUE)`

src/api/app/controllers/webui/package_controller.rb
danidoni: How about `:prefill_submit_request_form`?

hennevogel: Hm, maybe something more generic like `:branch_diff_info`?

src/api/app/controllers/webui2/package_controller.rb
```diff
@@ -25,4 +25,24 @@ def webui2_statistics
...
+  def webui2_prefill
+    target_project = ''
+    target_package = ''
+    description = ''
+    cleanup_source = @project.branch? # We should remove the package if this request is a branch
+    linked_package = @package.backend_package.links_to
+    if linked_package
+      target_project = linked_package.project.name
+      target_package = linked_package.name
+      description = @package.commit_message(target_project, target_package)
```
dmarcoux: It looks like the description isn't filled. I just tried it again to be sure in the review app.

To reproduce this:
- Go to https://obs-reviewlab.opensuse.org/danidoni-submit-request-form-is-not-pre-filled-with-target-project/package/show/home:Admin:branches:home:Admin/ruby
- See the `I improved the text` in the `Latest Revision` box. (this should be the description)
- Click on `Submit package`
- The project and package fields are pre-filled. The description is not.

hennevogel: This is based on `.changes` files, no on commit messages.

dmarcoux: We should rename the method from the `Package` model then. It's literally called `commit_message`.

hennevogel: Feel free to refactor this, afterward :)

dmarcoux: https://github.com/openSUSE/open-build-service/issues/8306

danidoni: I'll check it out

src/api/app/controllers/webui/package_controller.rb
danidoni: I like it

src/api/app/controllers/webui2/package_controller.rb
dmarcoux: Don't bother with this now, we can tackle the issue later and it doesn't have to be in this deliverable.

src/api/app/controllers/webui2/package_controller.rb
danidoni: We've found some other way to refactor this without initializing the variables. Will this work for you @vpereira ?

src/api/app/controllers/webui2/package_controller.rb
```diff
@@ -25,4 +25,21 @@ def webui2_statistics
...
+    end
+
+    render json: {
+      'targetProject': defined?(target_project) ? target_project : '',
```
vpereira: is that better than `linked_package ? linked_package.project.name : ' '` ?

danidoni:
Squashed and rebased

src/api/app/controllers/webui2/package_controller.rb
danidoni: Yes, I need the `target_project` and `target_package` for the `description`, so there is no reason to not use them there...

PR #7939, opened 2019-07-18
Add obs api support target to tests

eduardoj:
This will help us to run migration tests from 2.9 to 2.10 in kanku, and check for the status of systemd units.

Co-authored-by: Frank Schreiner <schreiner@suse.de>


eduardoj:
This is already checked in `t/0060-check_required_services.ts`. Closing.

PR #7940, opened 2019-07-18
Make the contribution graph optional

saraycp:
In config/options.yml 'contribution_graph' is set to false by default.
It should be decided by instance if it is enabled or not.

### DIFF
--- a/src/api/app/controllers/webui/user_controller.rb
+++ b/src/api/app/controllers/webui/user_controller.rb
@@ -22,7 +22,8 @@ class Webui::UserController < Webui::WebuiController
     @role_titles = @displayed_user.roles.global.pluck(:title)
     @account_edit_link = CONFIG['proxy_auth_account_page']
 
-    return unless switch_to_webui2
+    return unless switch_to_webui2 && (CONFIG['contribution_graph'] != :off)
+
     @last_day = Time.zone.today
 
     @first_day = @last_day - 52.weeks
--- a/src/api/app/views/webui2/webui/user/show.html.haml
+++ b/src/api/app/views/webui2/webui/user/show.html.haml
@@ -13,9 +13,10 @@
                                                           configuration: @configuration,
                                                           account_edit_link: @account_edit_link }
   .col-12.col-md-8.col-lg-9
-    = render partial: 'webui2/webui/user/activity', locals: { activity_hash: @activity_hash,
-                                                              first_day: @first_day,
-                                                              last_day: @last_day }
+    - unless CONFIG['contribution_graph'] == :off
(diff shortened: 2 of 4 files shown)


saraycp:
[Hide contribution box behind an OPS flag or make it user configurable](https://trello.com/c/xOiftzdL/1190-hide-contribution-box-behind-an-ops-flag-or-make-it-user-configurable)

src/api/app/controllers/webui/user_controller.rb
```diff
@@ -22,7 +22,8 @@ def show
     @role_titles = @displayed_user.roles.global.pluck(:title)
     @account_edit_link = CONFIG['proxy_auth_account_page']
 
-    return unless switch_to_webui2
+    return unless switch_to_webui2 && CONFIG['contribution_graph'].present?
```
hennevogel: Hm I think for the rest of the switches we use `:on`|`:off` as values.

hennevogel: I have also change the docu in the wiki now...

saraycp: @hennevogel done using `:on | :off`. I also fixed one failing test.

src/api/app/controllers/webui/user_controller.rb
```diff
@@ -22,7 +22,8 @@ def show
     @role_titles = @displayed_user.roles.global.pluck(:title)
     @account_edit_link = CONFIG['proxy_auth_account_page']
 
-    return unless switch_to_webui2
+    return unless switch_to_webui2 && (CONFIG['contribution_graph'] == :on)
```
hennevogel: Man I'm making a mess. this should be unless it's `:off` (so it's on by default). I'm so sorry...

saraycp: First of all, do we need this `:on` by default in `config/options.yml`? I have it `:off`.

On the other hand, I would say this line is ok :thinking: We should evaluate the lines after this condition (setting instance variables for the graph) only if we are in Bootstrap and we have the graph ON; otherwise it should return.

hennevogel: It should be commented by default in options.yml. Like

```
# Disable the contribution graph on the user home page
# contribution_graph = :off 
```

saraycp: done!

saraycp:
Finally, the requirements are:

- without `contribution_graph` option (commented or not present) => Shows contribution graph
- with `contribution_graph` option set to `:on` =>  Shows contribution graph
- with `contribution_graph` option set to `:off` => Does NOT show contribution graph

Please @hennevogel, have another look :bouquet:

Issue #7938, opened 2019-07-18
After installing obs 2.10 with zypper I get 'OBS API Autosetup is not enabled in sysconfig, skipping'


rhabacker:
I installed obs 2.10 on openSUSE_Leap_15.1 with zypper, based on the described procedure of https://github.com/openSUSE/open-build-service /blob/master/README.md. 

The 3rd step is to call the setup wizard:

`/usr/lib/obs/server/setup-appliance.sh`

When executing the command, the error message given in the title appears and the setup is aborted.

PR #7941, opened 2019-07-18
[doc] "setup-appliance.sh --force" in README.md

M0ses:
Fixes: #7938

Fixed our documentation to include the --force option for setup-appliance.sh
in our README.md.

Otherwise the installation as described in README.md will fail with an error
message like this:

```OBS API Autosetup is not enabled in sysconfig, skipping```

"--force" overwrite the variable "OBS_API_AUTOSETUP" taken from
/etc/sysconfig/obs-server to "yes".

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ After finishing the installation of your base system, follow these steps:
 3. Run our setup wizard
 
    ```shell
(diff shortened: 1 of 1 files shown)


hennevogel:
You should rebase on master to make CI work...

PR #7954, opened 2019-07-22
Enable Sorbet

vpereira:
Initials step necessary to enable Sorbet.

What was did until now and what is the scope of this PR:

Do the initial enabling of Sorbet and see how it impacts OBS in production.

For now I followed the steps described in: https://sorbet.org/docs/adopting

No type checking is enabled. 

The only file manually updated was the `Gemfile`

A new directory `sorbet` was added to the project.

After run `srb tc`, it touched every single ruby file adding the header `# typed: false`. So don't get distracted by the number of files changed. 

TODO:
- add the sorbet directory to the rpm spec file


hennevogel:
@vpereira if you add the "do not merge"-label, please explain why.

vpereira:
It is a experience, and therefore do not merge it. If you want to play with that, ping me.

vpereira:
Not really progressing. We can try it later or even wait until ruby 3.0 is there.

PR #7958, opened 2019-07-23
Make editor big where needed

### LINKED ISSUE #7799: Single CodeMirror instances are too small

eduardoj:
Commit taken from @hellcp 's #7924 pull request.

Fixes #7799.

### DIFF
--- a/src/api/app/views/webui2/shared/_editor.html.haml
+++ b/src/api/app/views/webui2/shared/_editor.html.haml
@@ -1,6 +1,6 @@
 :ruby
   save ||= {}
-  style ||= { read_only: false }
+  style ||= { read_only: false, big_editor: true }
   uid ||= next_codemirror_uid
   content_for(:head_style, codemirror_style(style))
 
@@ -31,7 +31,7 @@
   = text_area_tag("editor_#{uid}", text, cols: '0', rows: '0', data: data, class: 'form-control')
 
   - unless style[:read_only]
-    .sticky-bottom.bg-light{ id: "bottom_#{uid}" }
+    .bg-light{ id: "bottom_#{uid}" }
       .card-footer
         .d-flex.justify-content-end
           %p.small.text-muted
@@ -50,4 +50,4 @@
   = render partial: 'shared/editor_modal', locals: { uid: uid }
 
 - content_for :ready_function do
-  use_codemirror(#{uid}, #{style[:read_only]}, '#{mode}');
+  use_codemirror(#{uid}, #{style[:read_only]}, '#{mode}', #{style[:big_editor]});
--- a/src/api/app/assets/javascripts/webui2/cm2/use-codemirror.js
+++ b/src/api/app/assets/javascripts/webui2/cm2/use-codemirror.js
@@ -15,7 +15,7 @@ function cmMarkDiffLines(id) {
   $('#revision_details_' + id + ' .cm-negative').parents('.CodeMirror-line').addClass('CodeMirror-negative-line');
 }
 
-function use_codemirror(id, read_only, mode) {
+function use_codemirror(id, read_only, mode, big_editor) {
   var codeMirrorOptions = {
     lineNumbers: true,
     matchBrackets: false,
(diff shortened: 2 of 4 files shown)


vpereira:
@coolo do you want to take a look and see if it's solving your issue?

coolo:
https://obs-reviewlab.opensuse.org/eduardoj-make_editor_bigger_where_needed works great - but is bento only :)

eduardoj:
> https://obs-reviewlab.opensuse.org/eduardoj-make_editor_bigger_where_needed works great - but is bento only :)

Not now :)

danidoni:
> nitpick: I would just change the `big_editor` name for something more technical i.e: `fullscreen_editor` or `no_scrollbar_editor`

Are we using the big editor only for showing files (not editing)? If that's the case, why not using `view_only_editor`?

coolo:
> > nitpick: I would just change the `big_editor` name for something more technical i.e: `fullscreen_editor` or `no_scrollbar_editor`
> 
> Are we using the big editor only for showing files (not editing)? If that's the case, why not using `view_only_editor`?

Because there are other view only editors in requests that should not be big

coolo:
https://obs-reviewlab.opensuse.org/eduardoj-make_editor_bigger_where_needed/package/view_file/home:Admin/ruby/somefile.txt?expand=1 fails my tests.

Going there and scrolling with the mouse still scrolls only milimeters, but not the text. And Ctrl-F won't find you the 'Hidden' I've hidden at the end of the text.

coolo:
if I move the `height: auto` to the .CodeMirror it works. On the div it's toothless

vpereira:
> if I move the `height: auto` to the .CodeMirror it works. On the div it's toothless

won't that reintroduce that https://github.com/openSUSE/open-build-service/commit/a386a035108da5e4360cdd472b103174337b8516 ?

coolo:
https://build.opensuse.org/projects/openSUSE:Leap:15.2/meta is affected as well, btw

> > if I move the `height: auto` to the .CodeMirror it works. On the div it's toothless
> 
> won't that reintroduce [a386a03](https://github.com/openSUSE/open-build-service/commit/a386a035108da5e4360cdd472b103174337b8516) ?

That performance problem is about requests/show - let's worry about that one afterwards.

hellcp:
Yeah, a better way to approach this was to probably keep that `height:auto` and add a class for `height: 300px`, because that's only used once or twice in all the views. Big editor appears everywhere but the pages that include multiple editors, which are review and diff views.

hennevogel:
Just don't use `big_editor: true` on `BsRequest#show`...

eduardoj:
The `height: auto` was moved to the` .CodeMirror`. @hennevogel, @hellcp, @coolo, @vpereira, @danidoni, would you mind taking another look?

src/api/app/views/webui2/shared/_editor.html.haml
```diff
@@ -1,6 +1,6 @@
 :ruby
   save ||= {}
-  style ||= { read_only: false }
+  style ||= { read_only: false, big_editor: false }
```
hennevogel: Hm, we only have one page where this should be false, `BsRequest#show`. How about turning this default around?

coolo:
https://obs-reviewlab.opensuse.org/eduardoj-make_editor_bigger_where_needed/package/view_file/home:Admin/ruby/somefile.txt?expand=1 works now.

And I think I see a case of having the Save button at the top there :)

src/api/app/views/webui2/shared/_editor.html.haml
eduardoj: I agree. I will change the default.

eduardoj: Done. Should I merge all the commits into one?

hennevogel: Yeah sounds like a good idea :)

PR #7962, opened 2019-07-23
Enable Bootstrap feature in dev:development_testdata_create task

eduardoj:
This way we can see the Bootstrap UI in our review app.


eduardoj:
This task is failing in line [235](https://github.com/openSUSE/open-build-service/blob/e7fc3b1d6341859de9f61df16308d42c208f880c/src/api/lib/tasks/dev.rake#L235). That was why I put the `Flipper.enable` sentence just at the beggining of the task.

Issue #7916, opened 2019-07-16
Broken sorting in project monitor page


andreas-schwab:
Sorting by an arch column in a project monitor page does not group the unresolvable entries together.

On https://build.opensuse.org/project/monitor/devel:tools, sort by the second column, there are two groups of unresolvable entries at the top and at the bottom.

DavidKang:
@andreas-schwab, thanks for reporting the issue.

PR #7966, opened 2019-07-24
Fix sorting bug on project monitor page

saraycp:
When sorting by a certain column, the cells with the same content weren't always placed together. See _unresolvable_ cells in the image:

sorting_wrong

The problem is that Datatables sorts by the content of the cell, which is not a simple value ('unresolvable', 'failed'...) but an HTML tag to be displayed. So, it's trying to sort by values like "<a href=...".

Now we specify the value to be displayed and the value to sort/filter by.

Useful links:
- https://datatables.net/manual/data/orthogonal-data
- https://datatables.net/examples/advanced_init/html5-data-attributes.html

Fixes: #7916

### DIFF
--- a/src/api/app/assets/javascripts/webui2/project_monitor.js
+++ b/src/api/app/assets/javascripts/webui2/project_monitor.js
@@ -8,6 +8,7 @@ function statusCell(meta, statusHash, tableInfo, projectName, packageName) {
   var architecture = info[1];
   var status = statusHash[repository][architecture][packageName] || {};
   var code = status.code;
+  var cellContent = {};
   if (code === undefined) return null;
 
   var klass = 'build-state-' + code;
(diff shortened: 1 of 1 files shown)


src/api/app/assets/javascripts/webui2/project_monitor.js
```diff
@@ -61,7 +66,14 @@ function initializeMonitorDataTable() {
         data: null,
         className: 'text-center',
         render: function (packageName, type, row, meta) {
-          return statusCell(meta, statusHash, tableInfo, projectName, packageName);
+          var cellContent = statusCell(meta, statusHash, tableInfo, projectName, packageName);
+          if (cellContent === null) return null;
+          // Value to display in the cell
+          if (type === 'display') {
```
eduardoj: Nitpick: this `if` can also be a onliner, like the one two lines above.

saraycp: done!

PR #7968, opened 2019-07-24
Configuration updates for users do not need to be propagated to the backend

danidoni:
The changes to the Configuration regarding user properties do not need to be
sent to the backend.

Doing so, we skip the creation of new cassettes involving backend configuration
calls.

This will make some of the extra cassettes being recorded in https://github.com/openSUSE/open-build-service/pull/7551 to not happen anymore.

### DIFF
--- a/src/api/spec/bootstrap/features/webui/maintained_projects_spec.rb
+++ b/src/api/spec/bootstrap/features/webui/maintained_projects_spec.rb
@@ -7,11 +7,18 @@ RSpec.feature 'Bootstrap_MaintainedProjects', type: :feature, js: true, vcr: tru
   let(:maintenance_project) { create(:maintenance_project, name: 'maintenance_project', target_project: openSUSE_project) }
 
   describe 'index page' do
-    scenario 'without login' do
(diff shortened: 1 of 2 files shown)


src/api/spec/bootstrap/features/webui/maintained_projects_spec.rb
```diff
@@ -7,11 +7,20 @@
...
   describe 'index page' do
-    scenario 'without login' do
-      visit projects_project_maintained_projects_path(project_name: maintenance_project.name)
-      expect(page).to have_text('Maintained Projects')
-      expect(page).not_to have_selector('#new-maintenance-project-modal')
-      expect(page).not_to have_selector('#delete-maintained-project-modal')
+    context 'without login' do
+      before do
+        # The maintenance project factory needs a user logged in, so we fake it just for the creation
+        User.session = admin_user
```
coolo: this is an anti-pattern. Use admin_user.run_as {}

danidoni: Oh, thanks!

PR #7975, opened 2019-07-25
Fix updater notes to 2.10

adrianschroeter:
- correct commands to avoid failures
 - we did not recommend to use zypper services yet
 - drop lines only needed to 2.9 in 2.10 section
 - recommend to reboot instead just to restart some frontend parts only

### DIFF
--- a/dist/README.UPDATERS
+++ b/dist/README.UPDATERS
@@ -1,20 +1,25 @@
 For Updaters from OBS 2.9 to OBS 2.10
 =====================================
 
-Note: * Update from OBS 2.5 should also work, but is untested.
-        A direct update from OBS 2.4 or older will not work.
-        It's recommended to wait for all pending delayed jobs to be processed before
-        creating a DB dump or running the migrations.
-        To see the number of pending delayed jobs run
-        `(cd /srv/www/obs/api; bundle exec rails r -e production "puts Delayed::Job.count")`
+Note: Update from OBS 2.5 should also work, but it is not fully tested.
+      A direct update from OBS 2.4 or older will not work.
+      It's recommended to wait for all pending delayed jobs to be processed before
+      creating a DB dump or running the migrations.
+      To see the number of pending delayed jobs run
+
+        cd /srv/www/obs/api
+        bin/bundle exec bin/rails r -e production "puts Delayed::Job.count" RAILS_ENV=production
 
 
   1) The format of the OBS options.yml is now distinguishing between Rails environments.
-     Please run following command: '(cd /srv/www/obs/api/; rake migrate_options_yml)'
+     Please run following commands: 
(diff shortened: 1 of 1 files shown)


dist/README.UPDATERS
```diff
@@ -25,36 +30,24 @@ Note: * Update from OBS 2.5 should also work, but is untested.
...
-        systemctl restart obs-api-support.target
-        systemctl restart memcached
-
-  9) Enable and start the new services
-
-        systemctl enable obsservicedispatch
-        systemctl start obsservicedispatch
-
-     obsservicedispatch must run on the host where the src server is running.
+  7) Reboot your system to restart all OBS components
```
eduardoj: Fine with require a reboot. If we don't support upgrading without rebooting, we should also remove [this lines](https://github.com/openSUSE/open-build-service/blob/60f9ce6aa0ec62537c2dc9712cbe236551920df3/dist/obs-server.spec#L609-L613) from `obs-server.spec` file.

adrianschroeter:
On Donnerstag, 25. Juli 2019, 10:48:09 CEST Eduardo J. wrote:
> eduardoj commented on this pull request.
> 
> 
> 
> 
>          chown -R wwwrun.www /srv/www/obs/api/log
>          chown -R wwwrun.www /srv/www/obs/api/tmp
> 
> -  8) Restart following services in this order
> -
> -        systemctl restart apache2
> -        systemctl restart obs-api-support.target
> -        systemctl restart memcached
> -
> -  9) Enable and start the new services
> -
> -        systemctl enable obsservicedispatch
> -        systemctl start obsservicedispatch
> -
> -     obsservicedispatch must run on the host where the src server is
> running. +  7) Reboot your system to restart all OBS components
> 
> Fine with require a reboot. If we don't support upgrading without rebooting,
> we should also remove [this
> lines](https://github.com/openSUSE/open-build-service/blob/60f9ce6aa0ec6253
> 7c2dc9712cbe236551920df3/dist/obs-server.spec#L609-L613) from
> `obs-server.spec` file.

fine with me as well.

-- 

Adrian Schroeter
email: adrian@suse.de

dist/README.UPDATERS
```diff
@@ -25,36 +30,24 @@ Note: * Update from OBS 2.5 should also work, but is untested.
...
 
-# Needed in case of a ruby update
-#
-# 5) Change to ruby2.5
-#
-#       edit /etc/apache2/conf.d/mod_passenger.conf:
-#
-#       PassengerRuby "/usr/bin/ruby.ruby2.5"
+    NOTE: You may need to switch also your base distribution to a newer version. Adapt all repositories for that
+          and do an
```
vpereira: do a zypper up

PR #7977, opened 2019-07-25
 Don't throw an exception on a rss notification when a target project can not be found

### LINKED ISSUE #7094: [production][feeds#notifications] home:marxin:branches:games
feeds#notifications threw this 2 times
<pre>
home:marxin:branches:games
</pre>
[https://build.opensuse.org/user/feed/zRFTiGbGQKCx7dd6VUuSpY8g](https://build.opensuse.org/user/feed/zRFTiGbGQKCx7dd6VUuSpY8g)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5c7a0e1d7c8af20008465952)

eduardoj:
Prevent notification feeds to fail on notifications with deleted projects.

Fixes #7094.

### DIFF
--- a/src/api/app/models/event/request.rb
+++ b/src/api/app/models/event/request.rb
@@ -38,53 +38,8 @@ module Event
       { 'X-OBS-Review-By_Project' => payload['by_project'] }
     end
 
-    def headers_for_actions
-      ret = {}
-      payload['actions'].each_with_index do |a, index|
-        if payload['actions'].length == 1 || index.zero?
-          suffix = 'X-OBS-Request-Action'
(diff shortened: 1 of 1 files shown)


eduardoj:
@hennevogel:

> Please don't change the email notifications, it's enough to skip the diff.

Done.

coolo:
The reason it's still happening is you filter target projects, but branches are source projects.

The crash is about https://build.opensuse.org/request/show/699560 - and as you can see the target is alive and kicking

PR #7983, opened 2019-07-29
[dist] start obsapisetup after mariadb

M0ses:
setup-appliance.sh does datatbase actions. The service "mariadb" needs
to be started before.

mariadb has the

ExecStartPre=/usr/lib/mysql/mysql-systemd-helper  upgrade

condition, which may result in delayed upstart of mariadb after package
upgrade.

HINT: needs to be backported to 2.10 also


hennevogel:
@M0ses please also submit this to 2.10

PR #7985, opened 2019-07-29
[dist] fix for upgrade obs-api 2.9 to 2.10

M0ses:
From 2.9 to 2.10 obsapidelayed.service changed to
obs-api-support.target. The restart and enabling of obs-api-support.target
was tried in the post script. This is to late as the obsapidelayed.service
is already deinstalled at this point and the systemctl command will return
an error.

This change ensures that "systemctl is-enabled/is-active" will be executed
in the %pre section while upgrading the package and the state will be stored
files in the %_run_dir. In the %post section we check for this files and
try to enable/start obs-api-support.target.

NEEDS BACKPORT TO 2.10

### DIFF
--- a/dist/obs-server.spec
+++ b/dist/obs-server.spec
@@ -567,6 +567,16 @@ getent passwd obsapidelayed >/dev/null || \
   /usr/sbin/useradd -r -s /bin/bash -c "User for build service api delayed jobs" -d /srv/www/obs/api -g www obsapidelayed
 %service_add_pre %{obs_api_support_scripts}
 
+# On upgrade keep the values for the %post script
+if [ "$1" == 2 ]; then
+  systemctl --quiet is-enabled obsapidelayed.service && touch %{_rundir}/enable_obs-api-support.target
+  if systemctl --quiet is-active  obsapidelayed.service;then
+    touch %{_rundir}/start_obs-api-support.target
+    systemctl stop    obsapidelayed.service
+    systemctl disable obsapidelayed.service
+  fi
+fi
+
 %post -n obs-common
 %service_add_post obsstoragesetup.service
 %{fillup_and_insserv -n obs-server}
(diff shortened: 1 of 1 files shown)


dist/obs-server.spec
```diff
@@ -567,6 +567,15 @@ getent passwd obsapidelayed >/dev/null || \
   /usr/sbin/useradd -r -s /bin/bash -c "User for build service api delayed jobs" -d /srv/www/obs/api -g www obsapidelayed
 %service_add_pre %{obs_api_support_scripts}
 
+# On upgrade keep the values for the %post script
+if [ "$1" == 2 ]; then
+  systemctl --quiet is-enabled obsapidelayed.service && touch %{_rundir}/enable_obs-api-support.target
+  if systemctl --quiet is-active  obsapidelayed.service;then
+    touch %{_rundir}/start_obs-api-support.target
+    systemctl stop               obsapidelayed.service
```
darix: shouldnt you also disable it?

dist/obs-server.spec
```diff
@@ -567,6 +567,15 @@ getent passwd obsapidelayed >/dev/null || \
   /usr/sbin/useradd -r -s /bin/bash -c "User for build service api delayed jobs" -d /srv/www/obs/api -g www obsapidelayed
 %service_add_pre %{obs_api_support_scripts}
 
+# On upgrade keep the values for the %post script
+if [ "$1" == 2 ]; then
+  systemctl --quiet is-enabled obsapidelayed.service && touch %{_rundir}/enable_obs-api-support.target
```
darix: wouldnt you need to touch 2 files?

1. is_enabled 
2. is_active

M0ses: done in the next line if "is-active"

dist/obs-server.spec
M0ses: Shouldn't this be done automatically if the service is removed?

darix: uhm dont think so. we normally only remove services on uninstall.

M0ses: done

hennevogel:
@M0ses are you also going to submit this to 2.10?

Issue #7944, opened 2019-07-19
./obssrcserver: No such file or directory


rhabacker:
On openSUSE_Leap_15.1 after install obs I did run 
`/usr/lib/obs/server/setup-appliance.sh --force` 

and got the following errors:
```
/usr/lib/obs/server/setup-appliance.sh: line 641: ./obssrcserver: No such file or directory
/usr/lib/obs/server/setup-appliance.sh: line 642: ./obsrepserver: No such file or directory
```
Looking into the mentioned location shows: 
```
  if [[ $GPG_KEY_CREATED == 1 ]];then
    pushd .
    # avoid systemctl
    cd /etc/init.d
    ./obssrcserver reload
    ./obsrepserver reload
    popd
  fi
```
but unfortunally there is no related sysv init service in /etc/init.d
```
ls /etc/init.d
boot.d  obssignd  rc0.d  rc1.d  rc2.d  rc3.d  rc4.d  rc5.d  rc6.d  rcS.d
```
because the related services are real systemd services.

PR #7986, opened 2019-07-29
[dist] changed reload of SysV init to systemd

M0ses:
As our src/rep server nowadays use native systemd services we need to
use "systemctl restart" instead of trying to reload via the old SysV
init scripts.

Fixes: #7944


mlschroe:
Dumb question: shouldn't that be `reload` instead of `restart`?

M0ses:
@mlschroe : fixed

M0ses:
BACKPORT TO 2.10 needed

hennevogel:
> BACKPORT TO 2.10 needed

You should send a pull request to 2.10 then :)

M0ses:
As you could do also :) - it was only meant as a gentle reminder ...

PR #7990, opened 2019-07-29
Make CircleCI use SLE12SP4 instead of SLE12SP3

DavidKang:
Now that docker images for SLE12SP4 are available, we should use them in CircleCI.


hennevogel:
@DavidKang please rebase

DavidKang:
@hennevogel, done!

Issue #7984, opened 2019-07-29
bs_service: cannot write to rundir '/srv/obs/run'


rhabacker:
After installing obs from http://download.opensuse.org/repositories/OBS:/Server:/2.10/openSUSE_15.1/ the service obsservice could not be started. 

Expected Result
============
The service should start instead

How to Reproduce
===============
1. open ssh shell to obs server
2. run as root
`rcobsservice start`
3. run 
```
rcobsservice status

● obsservice.service - OBS source service server
   Loaded: loaded (/usr/lib/systemd/system/obsservice.service; enabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Mon 2019-07-29 12:24:50 CEST; 1h 7min ago
 Main PID: 2671 (code=exited, status=13)

Jul 29 12:24:50 lxserver29 systemd[1]: Started OBS source service server.
Jul 29 12:24:50 lxserver29 systemd[1]: obsservice.service: Main process exited, code=exited, status=13/n/a
Jul 29 12:24:50 lxserver29 systemd[1]: obsservice.service: Unit entered failed state.
Jul 29 12:24:50 lxserver29 systemd[1]: obsservice.service: Failed with result 'exit-code'.
```
Running the related service by hand with 

` /usr/lib/obs/server/bs_service --logfile src_service.log`

returns the message

`cannot write to rundir '/srv/obs/run'`

Further Information
===============
Changing  the following lines in  /usr/lib/obs/server/BSStdServer.pm
from 
```
   mkdir_p($rundir);
-  die("cannot write to rundir '$rundir'\n") unless POSIX::access($rundir, POSIX::W_OK);
```
to
```
   mkdir_p($rundir);
+  my $login = getpwuid($<);
+  die("cannot write to rundir '$rundir' as user '$login'\n") unless POSIX::access($rundir, POSIX::W_OK);
```
returns 

`cannot write to rundir '/srv/obs/run' as user 'obsservicerun`

The reason is that /srv/obs/run does not have required access rights for user 'obsservicerun'
```
ls /srv/obs -al | grep run 
drwxr-xr-x  2 obsrun        obsrun  4096 Jul 29 13:36 run
```
* Exact server version (in case a private instance is used)
obs-api-2.10-lp151.10.1.noarch
obs-service-download_url-0.1.1-lp151.2.1.noarch
obs-bundled-gems-2.10-lp151.10.1.x86_64
[...]

PR #7994, opened 2019-07-30
BSStdServer.pm: give more hints about the reason why we can't write in rundir

rhabacker:
see #7984


mlschroe:
I tweaked the patch a bit so that it does the getpwuid only if the access is denied. See pull #7998

PR #7996, opened 2019-07-30
setup-appliance.sh: Fix failure to find systemd services

### LINKED ISSUE #7995: setup-appliance.sh fail to find systemd services
Running setup-appliance.sh from an OBS 2.10 installation on openSUSE 15.1 shows several error messages like
```
obsdodup: unknown service
```
The reason is that setup-appliance.sh uses `chkconfig` to determine the state of recommended and optional services, which fails for systemd services like obsdodup.

Additional informations
obs-api-2.10-lp151.10.1.noarch
obs-service-download_url-0.1.1-lp151.2.1.noarch
obs-bundled-gems-2.10-lp151.10.1.x86_64
obs-service-verify_file-0.1.1-lp151.2.1.noarch
obs-common-2.10-lp151.10.1.noarch
obs-api-deps-2.10-lp151.10.1.x86_64
obs-worker-2.10-lp151.10.1.noarch
obs-signd-2.5.3-lp151.2.1.x86_64
obs-server-2.10-lp151.10.1.noarch

rhabacker:
With the new function check_service() System V and systemd services can
be queried via a uniform interface.

fixes #7995

### DIFF
--- a/dist/setup-appliance.sh
+++ b/dist/setup-appliance.sh
@@ -447,8 +447,11 @@ function check_recommended_backend_services {
   RECOMMENDED_SERVICES="obsdodup obsdeltastore obssigner obssignd obsservicedispatch"
 
   for srv in $RECOMMENDED_SERVICES;do
-    STATE=$(chkconfig $srv|awk '{print $2}')
-    if [[ $STATE != on ]];then
+    ENABLED=`systemctl is-enabled $srv`
+    if [[ "$ENABLED" == "enabled" ]];then
+        ACTIVE=`systemctl is-active $srv`
+        [[ "$ACTIVE" == "active" ]] || systemctl start $srv
+    else
       ask "Service $srv is not enabled. Would you like to enable it? [Yn]" "y"
       case $rv in
         y|yes|Y|YES)
@@ -472,8 +475,11 @@ function check_optional_backend_services {
   OPTIONAL_SERVICES="obswarden obsapisetup obsstoragesetup obsworker obsservice"
 
   for srv in $OPTIONAL_SERVICES;do
-    STATE=$(chkconfig $srv|awk '{print $2}')
-    if [[ $STATE != on ]];then
+    ENABLED=`systemctl is-enabled $srv`
+    if [[ "$ENABLED" == "enabled" ]];then
+        ACTIVE=`systemctl is-active $srv`
+        [[ "$ACTIVE" == "active" ]] || systemctl start $srv
+    else
       ask "Service $srv is not enabled. Would you like to enable it? [yN]" $DEFAULT_ANSWER
       case $rv in
         y|yes|Y|YES)
--- a/dist/setup-appliance.sh
+++ b/dist/setup-appliance.sh
@@ -23,10 +23,16 @@ function check_unit {
 
   echo "Checking unit $srv ..."
 
-  logline "Enabling $srv"
-  execute_silently systemctl enable $srv
-  if [[ $? -gt 0 ]];then
-    logline "WARNING: Enabling $srv daemon failed."
+  ENABLED=`systemctl is-enabled $srv 2>/dev/null`
+  if [[ "$ENABLED" != "enabled" ]]; then
+    systemctl enable $srv
+    if [[ $? -gt 0 ]];then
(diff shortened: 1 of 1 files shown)


dist/setup-appliance.sh
```diff
@@ -427,6 +427,28 @@ function ask {
...
+            ACTIVE=`systemctl is-active $srv`
+            [[ "$ACTIVE" == "active" ]] || systemctl start $srv
+        fi
+    fi
+    if [[ $STATE != on ]];then
+      ask "Service $srv is not enabled. Would you like to enable it? [Yn]" "y"
+      case $rv in
+        y|yes|Y|YES)
+          systemctl enable $srv
+          systemctl start $srv
```
Conan-Kudo: `systemctl enable --now $srv`

dist/setup-appliance.sh
```diff
@@ -427,16 +427,34 @@ function ask {
...
+    if [[ $STATE == "" ]]; then
+        ENABLED=`systemctl is-enabled $srv`
+        if [[ "$ENABLED" == "enabled" ]];then
+            STATE=on
+            ACTIVE=`systemctl is-active $srv`
+            [[ "$ACTIVE" == "active" ]] || systemctl start $srv
+        fi
+    fi
+    if [[ $STATE != on ]];then
+      ask "Service $srv is not enabled. Would you like to enable it? [Yn]" "y"
```
M0ses: The user should only be asked for RECOMMENDED_SERVICES or OPTIONAL_SERVICES, 
but not for REQUIRED_SERVICES

dist/setup-appliance.sh
```diff
@@ -427,16 +427,34 @@ function ask {
 
 }
 ###############################################################################
+function check_service {
+    srv=$1
+    STATE=$(chkconfig $srv 2>/dev/null|awk '{print $2}')
```
M0ses: I'm wondering if we really should use `chkconfig` here. 
Nowadays all supported OS'es should have `systemctl` available.
Opinions?

rhabacker: The service obssignd seems not to have been ported to systemd yet (see 2.10: https://build.opensuse.org/package/show/OBS:Server:2.10:Staging/obs-signd master: https://github.com/openSUSE/obs-sign/blob/master/dist/obs-signd.spec
), but is contained in the RECOMMENDED_SERVICES (see https://github.com/openSUSE/open-build-service/blob/master/dist/setup-appliance.sh#L447). 

After an installation of obs 2.10 SystemV services seem to be generated in systemctl
```
systemctl status obssignd
● obssignd.service - LSB: start the gpg sign daemon
   Loaded: loaded (/etc/init.d/obssignd; generated; vendor preset: disabled)
   Active: active (running) since Fri 2019-08-02 09:47:04 CEST; 2 weeks 3 days ago
     Docs: man:systemd-sysv-generator(8)
    Tasks: 3 (limit: 4915)
   CGroup: /system.slice/obssignd.service
           ├─1942 /usr/bin/perl /usr/sbin/signd -f
           └─2158 gpg-agent --homedir /srv/obs/gnupg --use-standard-socket --daemon

Aug 02 09:47:04 host systemd[1]: Starting LSB: start the gpg sign daemon...
Aug 02 09:47:04 host obssignd[1804]: Starting gpg sign daemon (signd): ..done
Aug 02 09:47:04 host systemd[1]: Started LSB: start the gpg sign daemon.
```
but I can't tell if this is always the case. It would be better to convert this service to a real systemd service first from my opinion.

Conan-Kudo: This is always the case. That said, there is a systemd service for it, it's just not installed on the openSUSE packaging (it is for the Fedora packaging).

dist/setup-appliance.sh
rhabacker: I added this switch to all systemctl enable calls, see commit  445bdc0

rhabacker:
The commits from this merge request has been superseeded by the commits https://github.com/openSUSE/open-build-service/commit/94023b05a063870bf04ee808e4680302a3da62c9 and https://github.com/openSUSE/open-build-service/commit/a90c85f2dfe12eb9bc48b7e066629ca08171c240, so that this mr can be closed as resolved.

PR #8005, opened 2019-07-31
Convert custom validations into Rails validations

vpereira:
This implements validation of presence of the various possible review assignees (package, project, group or user).

In addition this ensures that these validations always run and not
only during object creation.

As a note, before this PR the validations for the review assignee are running on initial creation but not afterwards. ~~So in theory, after this PR there could be instances that have such reviews that would become invalid~~ we don't have that issue in our instances

NOTE:

Because the old PR #6282 was from end 2018 and there was some change requests pending and long dicussion, I reopened it here. Credits go to @bgeuken


src/api/app/models/review.rb
```diff
@@ -57,6 +64,13 @@ class NotFoundError < APIError
   before_validation :set_reviewable_association
   after_commit :update_cache
 
+  def review_assignment
+    errors.add(:unknown, 'no reviewer defined') unless by_user || by_group || by_project
+
+    return unless validate_reviewer_fields
```
vpereira: does it here make sense? 

Looking the possible values to make `validate_reviewer_fields` the possible combinations would return true, which looks wrong to me:

```
by_user: true by_group: false by_project: false by_package: true
by_user: false by_group: true by_project: true by_package: true
by_user: false by_group: true by_project: false by_package: true
by_user: true by_group: true by_project: false by_package: true
by_user: true by_group: true by_project: true by_package: true
by_user: true by_group: true by_project: false by_package: false
by_user: false by_group: true by_project: true by_package: false
by_user: true by_group: true by_project: true by_package: false
by_user: true by_group: false by_project: true by_package: true
by_user: true by_group: false by_project: true by_package: false
```

@coolo could you help me here?

src/api/db/data/20190116105222_sanitize_reviews.rb
```diff
@@ -0,0 +1,36 @@
+class SanitizeReviews < ActiveRecord::Migration[5.2]
```
vpereira: so this here is probably to bring the reviews to a valid state, if this PR gets merge. This data migration isn't necessary for b.o.o and b.s.d but maybe some customer is using multiple review targets. Not sure if It will work as expected.

src/api/app/models/review.rb
coolo: note that returning true is an error condition. So read the boolean expression as
```
  def validate_reviewer_fields
    return (by_group || by_project || by_package) if by_user
    return (by_project || by_package) if by_group
    false
  end
```

The function name implies that true is good, so that is the only problem I see :)

vpereira: hi @coolo, yes I didn't push my changes, sorry... I just need a new name for this method. Maybe `invalid_reviewers?`  

BTW I still think that the cases where the `validate_reviewer_fields` returns false, still don't match with the wanted logic right? 

I mean both

```
by_user: false by_group: false by_project: true by_package: true
by_user: false by_group: true by_project: false by_package: false
```

should be valid (and return false, but they returning true)

coolo: by_package _requires_ by_project to be set as well.

Your 2nd line is only having by_group and as such is fine

coolo: Basically the only valid options are:

```
by_user: true by_group: true by_project: false by_package: false
by_user: false by_group: true by_project: false by_package: false
by_user: false by_group: true by_project: true by_package: false
by_user: false by_group: true by_project: true by_package: true
```

Every other combination is invalid.

src/api/app/models/review.rb
```diff
@@ -270,6 +268,7 @@ def set_reviewable_association
     self.group = Group.find_by(title: by_group)
   end
 
+  # A review can be by one and only one of folowing options: by_user, by_group, by_project, by_package
```
coolo: wrong for by_package

src/api/app/models/review.rb
vpereira: Awesome, thank you!

coolo: I had a bug in my initial comment that by_group was too often true - hope that didn't confuse you :)

src/api/db/data/20190116105222_sanitize_reviews.rb
```diff
@@ -0,0 +1,36 @@
...
+  # Find reviews that have multiple review targets (e.g. by_user and by_group)
+  # and split them up. So that reviews are per review target.
+  def up
+    Review.where.not(by_group: nil, by_user: nil, by_project: nil).find_each do |review|
+      # The package / project review
+      review.dup.update!(by_group: nil, group_id: nil, by_user: nil, user_id: nil)
+      # The group review
+      review.dup.update!(by_user: nil, user_id: nil, by_project: nil, project_id: nil, by_package: nil, package_id: nil)
+      # Te user review
```
danidoni: Typo: `Te` -> `The`

src/api/app/models/review.rb
```diff
@@ -21,7 +21,14 @@ class NotFoundError < APIError
   validates :reviewer, length: { maximum: 250 }
   validates :reason, length: { maximum: 65_534 }
 
-  validate :check_initial, on: [:create]
+  validates :user, presence: true, if: :by_user?, on: :create
```
dmarcoux: Should these validations be only on creation? I feel like I'm either missing something or `on: :create` should be removed.

src/api/db/data/20190116105222_sanitize_reviews.rb
vpereira: done!

src/api/app/models/review.rb
vpereira: Hi Dani, yes, The PR was already like that, but yes, I think it shouldn't be just in the `:create` @coolo or @hennevogel what do you think? Should be possible to change the initial value anyway (i.e change a review from `by_project` to `by_user`) ?

coolo: No, but it's possible to delete a user - and then review should stay intact. `:by_user` stays, `user` might be dead in old reviews.

Not so much of a problem with users (but technically a possiblity, think DSVGO :), but with project and package it happens *every day*

vpereira: @coolo ok, but then the following scenario would be possible when we do not limit the validation to `create`:

User create a review by project. Project is deleted. The review cannot be accepted, revoked, whatsoever, since the validation may fail.. right? So should we limit the validation just for creation?

vpereira: hm from comment looks like it should work:

```
 # The authoritative storage are the by_ attributes as even when a record (project, package ...) got deleted
  # the review should still be usable, however, the entity association is nullified
```

coolo: What's the point of creating a new PR if you repeat the discussion from the old one? :)

https://github.com/openSUSE/open-build-service/pull/6282#discussion_r234075246

vpereira: But now I have the solution, based in your comment, without any BIAS and less noise.

if the valdiation `by_package` isn't limited to `on_create`, when you delete got deleted, you cannot change the state of the review anymore. I added some test cases to validate it. So looks like it make sense to have the validation just `on: :create` what do you think @coolo

src/api/app/models/review.rb
```diff
@@ -288,7 +268,8 @@ def set_reviewable_association
     self.group = Group.find_by(title: by_group)
   end
 
-  def validate_reviewer_fields
+  # A review can be by one and only one of folowing options: by_user, by_group or by_project
```
dmarcoux: Typo: fol**l**owing

src/api/db/data/20190116105222_sanitize_reviews.rb
```diff
@@ -0,0 +1,36 @@
...
+      review.dup.update!(by_group: nil, group_id: nil, by_user: nil, user_id: nil)
+      # The group review
+      review.dup.update!(by_user: nil, user_id: nil, by_project: nil, project_id: nil, by_package: nil, package_id: nil)
+      # The user review
+      review.update!(by_group: nil, group_id: nil, by_project: nil, project_id: nil, by_package: nil, package_id: nil)
+    end
+    Review.where.not(by_group: nil, by_user: nil).find_each do |review|
+      # The group review
+      review.dup.update!(by_user: nil, user_id: nil)
+      # Te user review
```
dmarcoux: Typo: T**h**e

src/api/db/data/20190116105222_sanitize_reviews.rb
```diff
@@ -0,0 +1,36 @@
...
+      review.update!(by_group: nil, group_id: nil)
+    end
+    Review.where.not(by_group: nil, by_project: nil).find_each do |review|
+      # The group review
+      review.dup.update!(by_project: nil, project_id: nil, by_package: nil, package_id: nil)
+      # The package / project review
+      review.update!(by_group: nil, group_id: nil)
+    end
+    Review.where.not(by_user: nil, by_project: nil).find_each do |review|
+      # Te user review
```
dmarcoux: Typo: T**h**e

src/api/db/data/20190116105222_sanitize_reviews.rb
dmarcoux: I'm not really sure if we need this. I would drop it.

hennevogel:
Hakiri is not complaining about anything on it's page.... :man_shrugging:

PR #8007, opened 2019-08-01
[dist] enable obs-api-support.target on update 

M0ses:


hennevogel:
@M0ses why shouldn't we merge?

M0ses:
It doesn`t work ATM WIP

M0ses:
Doing last check in kanku before ready to merge

M0ses:
IMHO ready to merge

hennevogel:
Sounds like something you want to push to 2.10 too, right?

M0ses:
@hennevogel : Correct, I will prepare a PR for 2.10 and add you as reviewer

PR #8019, opened 2019-08-02
Fix CodeMirror height on read only mode

### LINKED ISSUE #8018: Viewing spec files through a letterbox is not good
# Issue Description
Viewing spec files is difficult when it is limited to 300px. On a 1080p monitor that is 1/3 of the usable screen space.
On a 1440p monitor it's the difference between seeing 15 lines vs 65 lines at a time.

Expected Result
============
The data you want to see should fill the screen as much as possible.

How to Reproduce
===============
1. Go to https://build.opensuse.org/package/view_file/shells/zsh/zsh.spec?expand=1
2. Realise viewing only 15 lines at a time is a bad idea ;)

Further Information
===============
Can be overridden by removing this style.
.CodeMirror {
    height: 300px;
}

DavidKang:
When we are on `read only` mode we didn't include `big_editor`
parameter. That was causing that the editor was rendering with an
incorrect height.

Fix #8018

### DIFF
--- a/src/api/app/views/webui2/shared/_editor.html.haml
+++ b/src/api/app/views/webui2/shared/_editor.html.haml
@@ -1,11 +1,12 @@
 :ruby
   save ||= {}
-  style ||= { read_only: false, big_editor: true }
+  options = { read_only: false, big_editor: true }
+  options.merge!(style) if defined?(style)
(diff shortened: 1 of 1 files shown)


DavidKang:
Close in favour of #8020

PR #8030, opened 2019-08-05
Add optional: enable new UI theme

lrupp:
At least in one appliance, the old UI was still present.
Thanks to Victor for the tip with "Flipper[:bootstrap].enable" - I guess this might help others as well.

### DIFF
--- a/dist/README.UPDATERS
+++ b/dist/README.UPDATERS
@@ -43,13 +43,18 @@ Note: * Update from OBS 2.5 should also work, but is untested.
         chown -R wwwrun.www /srv/www/obs/api/log
         chown -R wwwrun.www /srv/www/obs/api/tmp
 
-  8) Restart following services in this order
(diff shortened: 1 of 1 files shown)


hennevogel:
Those are instructions for the released version, this won't work in 2.10. In 2.10 bootstrap is default and the feature flipper doesn't exist.

This will also not work for the next release, we will have moved bootstrap out of beta/rollout in master by then. 

Are you deploying master?

vpereira:
Ah true,  @hennevogel. Any suggestion where @lrupp should add this important bit of information?

hennevogel:
> Ah true, @hennevogel. Any suggestion where @lrupp should add this important bit of information?

We don't really have any place to document deployments of master. How to do this in general is documented [in the wiki](https://github.com/openSUSE/open-build-service/wiki/Feature-Toggles) already. If this was about master, it doesn't really make sense.

PR #8034, opened 2019-08-06
Improve readability by removing some small class modifiers

danidoni:
Improve readability by removing some small class modifiers

Some of the pages of the project have readability issues because the text is too
small. We removed the CSS classes to increase the size and improve on that.

Screenshot_2019-08-07_11-50-28

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/collapsible-text.scss
+++ b/src/api/app/assets/stylesheets/webui2/collapsible-text.scss
@@ -1,8 +1,8 @@
 @mixin collapsible($maxHeight: 4.5em) {
   .obs-collapsible-textbox {
-    word-break: break-word;
     padding-right: 1em;
-    text-align: justify;
+    hyphens: auto;
+    max-width: 600px;
 
     .obs-collapsible-text {
       overflow: hidden;
--- a/src/api/app/assets/stylesheets/webui2/collapsible-text.scss
+++ b/src/api/app/assets/stylesheets/webui2/collapsible-text.scss
@@ -2,7 +2,7 @@
   .obs-collapsible-textbox {
     padding-right: 1em;
     hyphens: auto;
-    max-width: 600px;
+    max-width: $seventy-five-chars;
 
     .obs-collapsible-text {
       overflow: hidden;
--- a/src/api/app/assets/stylesheets/webui2/bootstrap_variables/fonts.scss
+++ b/src/api/app/assets/stylesheets/webui2/bootstrap_variables/fonts.scss
@@ -1,0 +1,1 @@
+$seventy-five-chars: 75ch;
--- a/src/api/app/assets/stylesheets/webui2/comments.scss
+++ b/src/api/app/assets/stylesheets/webui2/comments.scss
@@ -2,6 +2,10 @@
   .media .media-body {
     hyphens: auto;
 
+    p:not(:first-child) {
+      max-width: 600px;
+    }
+
     pre, code {
       @extend .p-2;
       @extend .bg-light;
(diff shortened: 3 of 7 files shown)


dmarcoux:
I pushed again to the PR to add an extra line in `src/api/app/assets/stylesheets/webui2/comments.scss` between the new rule we added and the existing ones

hennevogel:
Hm, why did you squash all those nice commits that explained *why* you where doing something :cry:

vpereira:
ah nice @dmarcoux so I think we are good to go, right?

vpereira:
Because we were all in the mob programming, who can review it? @coolo ?

src/api/app/assets/stylesheets/webui2/collapsible-text.scss
```diff
@@ -1,8 +1,8 @@
 @mixin collapsible($maxHeight: 4.5em) {
   .obs-collapsible-textbox {
-    word-break: break-word;
     padding-right: 1em;
-    text-align: justify;
+    hyphens: auto;
+    max-width: 600px;
```
vpereira: is that a bad pattern to add a comment here explaining why we went for 600px? Our reasoning was to be around 65-80 chars long.

danidoni:
> Hm, why did you squash all those nice commits that explained _why_ you where doing something cry

Oh sorry, I thought those WIP comments were just noise...

src/api/app/assets/stylesheets/webui2/collapsible-text.scss
danidoni: I don't see how putting a comment explaining something would be a bad pattern... For me, you're good to go and write it just there

vpereira: hm maybe I just didn't express myself well.. yes comments are not bad, but we don't do it in css so I'm not sure if just this arbitrary value among all  other arbitrary values in our CSS should be commented.

vpereira:
[Project: Review font classes / tags and line length](https://trello.com/c/u5dCsQyq/1204-project-review-font-classes-tags-and-line-length)

src/api/app/assets/stylesheets/webui2/collapsible-text.scss
dmarcoux: We could use a Sass variable for this and it would have 2 benefits:

- It documents itself (if the variable has a name like: `80_characters_width`)
- We are consistent in all places where we use the variable

vpereira: I liked it. Just the variable name should be something else, since it cannot start with a number.

danidoni: `eighty_characters_width` maybe?

vpereira: Actually with this calculator http://www.unitconversion.org/typography/characters-x-to-pixels-x-conversion.html, 600px are 75 chars :)

vpereira: maybe it varies with font-size, or font-family?

hennevogel: > is that a bad pattern to add a comment here explaining why we went for 600px?

No it's cool. Better would be to also have a single commit with a good commit description that you can blame :cowboy_hat_face:

hennevogel:
The coverage looks good but the [upload](https://circleci.com/gh/danidoni/open-build-service/178) failed. Merging.

PR #8036, opened 2019-08-07
Adjust typography on Rpmlint and Build Result tabs

saraycp:
We changed the Rpmlint and Build Results partials for the package's build result box which appears in the request page as well.

We decided to change the font family of the Rpmlint result, that was in a `<pre>`tag, in order to reduce the number of different font families and increase readability.

**Before:**
Screenshot-2019-8-7 Request 8 (new)(5)
Screenshot-2019-8-7 Request 8 (new)(4)

**After:**

Screenshot-2019-8-7 Request 8 (new)(2)
Screenshot-2019-8-7 Request 8 (new)(3)

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/build-results.scss
+++ b/src/api/app/assets/stylesheets/webui2/build-results.scss
@@ -1,5 +1,5 @@
 .rpmlint-result {
-  font-size: .7rem;
+  font-size: .95em;
   max-height: 35rem;
 }
 
--- a/src/api/app/views/webui2/webui/package/_buildstatus.html.haml
+++ b/src/api/app/views/webui2/webui/package/_buildstatus.html.haml
@@ -1,5 +1,5 @@
 - unless buildresults.excluded_counter.zero? && !buildresults.show_all
-  .mt-3.small.custom-control-checkbox
+  .mt-3.custom-control-checkbox
     = check_box_tag("show_all_#{index}", true, buildresults.show_all, class: 'custom-control-input d-none',
                     onchange: "updateBuildResult('#{index}')")
     - label_message = buildresults.excluded_counter.zero? ? 'Hide' : "Show #{buildresults.excluded_counter}"
@@ -9,7 +9,7 @@
 - buildresults.results.each_pair do |package, results|
   %h5.text-primary.mt-3.mb-3= package
   .mb-2#package-buildstatus
-    .small.px-3
+    .px-3
       - if results.present?
         - previous_repo = nil
         - results.each do |result|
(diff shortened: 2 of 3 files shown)


DavidKang:
Just one thing, on the package view, the selectors of rpmlint are stacked
Screenshot_2019-08-07 perl-Citrix

krauselukas:
> Just one thing, on the package view, the selectors of rpmlint are stacked
> Screenshot_2019-08-07 perl-Citrix

oh true, do you see an issue in that? or can we leave it as it is?

krauselukas:
> There is an issue in the build results on the package show:
> 
> Screenshot from 2019-08-07 15-08-26
> 
> Somehow, it happens only with this truck icon because it's a bit larger than the other icons like the cog and calendar.

@dmarcoux will have a look

krauselukas:
> There is an issue in the build results on the package show:
> 
> Screenshot from 2019-08-07 15-08-26
> 
> Somehow, it happens only with this truck icon because it's a bit larger than the other icons like the cog and calendar.

@dmarcoux we added some offset, this fixed the problem
Screenshot-2019-8-7 ctris

dmarcoux:
It's much better now!

I realized that it's also a bit of an issue on mobile when the different sections/boxes are stacked:

mobile-stacked-view

To be honest, it was already an issue before as you can see in the screenshot. It's just a bit worse now.

krauselukas:
@dmarcoux we reduced the whitespace for xl screens, we got rid of the offsets and added a margin instead. We also adjusted the column length for certain screen sizes. 

There is still an issue in medium sized screens, which should be takled in another PR. The solution could be making it one column instead of two.

Screenshot-2019-8-7 Request 4 (review)(3)

src/api/app/assets/stylesheets/webui2/build-results.scss
```diff
@@ -1,6 +1,7 @@
 .rpmlint-result {
-  font-size: .7rem;
+  font-size: .9em;
   max-height: 35rem;
+  font-family: $font-family-sans-serif;
```
vpereira: in that case, because we want to simulate the output of a shell, IMO it would be ok to have a different font stack like:  

`font-family: Inconsolata, Consolas, monospace`

What do you think @hennevogel ? Do you want to have it look like a shell output?

vpereira: however the original font-stack, `font-family: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace; ` is as well good!

hennevogel: Hm, I'm not sure that we want to simulate output of a shell. You are right, it's kind of a log. Maybe we need a component for things like this?

saraycp:
We moved back to monospace font family, but instead of having a tiny font size like the following

tiny

we changed it to something bigger. We propose `0.95em` because with the combination `monospace -1em/16px` the text seems even bigger than the text in the description. Now we think it's more homogeneous.

rmlint_monospace

What do you think @vpereira @hennevogel ?

**BTW, we'd suggest having `0.95em` also for the editor!! :nerd_face:**

saraycp:
[Request: Review font classes / tags and line length](https://trello.com/c/gAEO50ru/1206-request-review-font-classes-tags-and-line-length)

PR #8038, opened 2019-08-07
Do not duplicate checkout volume

hennevogel:
It's already in docker-compose.yml, no need to specify it a second time in the
override file.


dmarcoux:
The `OBS Package Build` failure doesn't matter for this PR.

PR #8039, opened 2019-08-07
Remove strong tag from repository tables

danidoni:
The name of the project in the repository table is not important enough to be that strong. So we decided to remove it

Before:
before

After:
after


hennevogel:
coveralls is having problems today, merging. see codecov.

PR #8042, opened 2019-08-07
Change project and package grid

DavidKang:
We gave more space on buildresult section.

### Before
Screenshot_2019-08-07 home Admin(1)

### After
Screenshot_2019-08-07 home Admin

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/project-package-show-grid.scss
+++ b/src/api/app/assets/stylesheets/webui2/project-package-show-grid.scss
@@ -19,7 +19,7 @@
     "build-results"
     "comments";
   @include media-breakpoint-up(lg) {
-    grid-template-columns: 1fr 350px;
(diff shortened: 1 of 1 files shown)


hennevogel:
> We gave more space on buildresult section.

Why? (Asking for future you ;-)

saraycp:
@dmarcoux we have just added an offset in #8042, so the architecture and the status do not overlap anymore. Don't you think that could solve the problem?

I like @DavidKang's idea, we don't need so many space for the cards on the left and we do for the build result box.

dmarcoux:
I also think the changes introduced in this PR are nice. However, to be sure that this PR doesn't cause issue, I would wait for #8036 to be merged and then rebase this PR on master to double-check that everything is fine.

vpereira:
I rebased it and it looks good in mobile reasonable sizes

PR #8044, opened 2019-08-07
Fixed manual for initial interconnect setup, added note about docker

digitaltom:
The interconnect button name changed, and on a default SUSE system, you need to make sure your user is in the docker group before running  `rake docker build`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -208,7 +208,7 @@ make sure you run all this from a filesystem that supports this.
    git submodule update
    ```
 
-5. Build your development environment with (make sure your user has access to Docker):
+5. Build your development environment with:
 
     ```
     rake docker:build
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -208,7 +208,7 @@ make sure you run all this from a filesystem that supports this.
    git submodule update
    ```
 
-5. Build your development environment with:
+5. Build your development environment with (make sure your user has access to Docker):
```
hennevogel: That's mean, it's like saying: "Make sure you sacrificed the right :goat: " without telling you which :goat:...

And that you have to take care of this is mentioned in the linked openSUSE docu https://en.opensuse.org/SDB:Docker

digitaltom: ok makes sense, i dropped it.

hennevogel: You saved a lot of :goat: :goat: :goat: :smile:

PR #8048, opened 2019-08-08
Show the error raised when editting status comment

saraycp:
In the status tab/page of a project there is a list of failing packages. A maintainer user can add a comment to each of the failing packages. However, if there were any trouble while adding the comment, the error wasn't visible.

Now the error is visible inside the comment cell.

error_status


dmarcoux:
I just noticed there's a typo in the commit message (in `editing`). Well, it's too late now.

Issue #8047, opened 2019-08-07
[production][source#package_command] undefined method `get_by_name' for #<Package::ActiveRecord_Associations_CollectionProxy:0x00007f5...


hennevogel:
source#package_command threw this 12 times
<pre>
undefined method `get_by_name&#39; for #&lt;Package::ActiveRecord_Associations_CollectionProxy:0x00007f4f9c83b330&gt;
Did you mean?  get_value
</pre>
[https://api.opensuse.org/source/Cloud:OpenStack:Rocky/python-XStatic-jQuery?cmd=branch](https://api.opensuse.org/source/Cloud:OpenStack:Rocky/python-XStatic-jQuery?cmd=branch)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5d48c1ff2834890008a647d1)

PR #8051, opened 2019-08-08
Use find_by instead of get_by_name

DavidKang:
If a project uses a develproject setting, resolve_devel_package method
crashes, because it's using get_by_name that is not defined on the
collection proxy.

Fixes:  #8047


dmarcoux:
There are only some RuboCop offenses to fix

PR #8053, opened 2019-08-08
Add space between parent table and subproject table

DavidKang:
`@parentproject` was not used in the new UI, then the class `mt-3` was never applied.

### Before
Screenshot_2019-08-08 Open Build Service(1)

### After
Screenshot_2019-08-08 Open Build Service


dmarcoux:
This page should be redesigned. I wouldn't bother with trying to improve the current state.

DavidKang:
@dmarcoux, I agree with you that the page needs a redesign. That's the reason I didn't put more effort into it

PR #8055, opened 2019-08-08
Add dev environment setup for macOS to the contribution guide

krauselukas:
People who are using a Mac for development, sometimes struggling to
setup OBS on there machines.


hennevogel:
Can you move this to the [wiki](https://github.com/openSUSE/open-build-service/wiki)? I don't think this is interesting enough to be added here.

PR #8057, opened 2019-08-08
Typography - Project Repositories

dmarcoux:
This page needs a redesign, but for now this will already improve the view a bit. The `Repositories Flags` section is already addressed in #8039.

Reorganize hierarchy of repository paths and put emphasis on a repository's architectures.

Before:
project-repositories-before

After:
project-repositories


hennevogel:
I like it. Another idea: Use `<ol>` instead of `<ul>` for the path's. The order matters very much...

dmarcoux:
> I like it. Another idea: Use `<ol>` instead of `<ul>` for the path's. The order matters very much...

I added a commit for this. Good idea!

PR #8058, opened 2019-08-08
Show URL the same way as on Package#show

DavidKang:
The URL of the project was too small and inconsistent to other sections like package#show view.

### Before:
Screenshot_2019-08-08 home Admin(3)

### After:
Screenshot_2019-08-08 home Admin(2)

### DIFF
--- a/src/api/app/views/webui2/webui/project/show.html.haml
+++ b/src/api/app/views/webui2/webui/project/show.html.haml
@@ -11,8 +11,8 @@
       .col-md-8
         %h3#project-title
           = @project.title.presence || @project
-          - if @project.url.present?
-            %small= link_to(@project.url, @project.url)
+        - if @project.url.present?
+          = link_to(@project.url, @project.url, class: 'mb-3 d-block')
         #description-text
           - if @project.description.blank?
             %i No description set


src/api/app/views/webui2/webui/project/show.html.haml
```diff
@@ -11,8 +11,8 @@
       .col-md-8
         %h3#project-title
           = @project.title.presence || @project
-          - if @project.url.present?
-            %small= link_to(@project.url, @project.url)
+        - if @project.url.present?
+          = link_to(@project.url, @project.url, class: 'mb-3 d-block')
```
dmarcoux: Does the link need to be a block element? We do the same somehow in package#show, but I don't understand why it's needed.

DavidKang:
@dmarcoux,

> The commit message is wrong. mb-3 is added, not mt-3.

Right!, thanks :smile_cat: 

> In comparison to package#show, which looks better without the big gap.

I don't have this gap, could you look if it's your description?

dmarcoux:
Oh right, my bad, there's an extra paragraph in my description. Sorry for bothering you with this :smile:!

dmarcoux:
I removed `d-block` and it renders just the same. I believe we should remove it. Same for the package#show view while we're at it.

hennevogel:
> > The commit message is wrong. mb-3 is added, not mt-3.
> Right!, thanks smile_cat

I also don't like this for another reason. It's like adding the patch to the headline. I can see that you added classes. Tell me what you did: `Show URL the same way as on Package#show`

DavidKang:
@dmarcoux, this is what happens when I remove d-block class

Screenshot_2019-08-08 perl-Citrix(1)

Issue #8012, opened 2019-08-01
DoD does not fetch latest python-certifi package


seife:
# Issue Description
SLES12-SP4, own OBS instance, KIWI image build
After SLE-Module-Public-Cloud got an updated python-requests, this is pulled into the build
The new python-requests depends on python-certifi >= 2017.4.17
python-certifi is available in version 2018.4.16 from SLE-Module-Public-Cloud also.

But it does not get picked up.

Reproducer: https://build.opensuse.org/package/show/home:seife:kiwitest/SLES_12SP4_liveiso1 rev 3

My guess is, that this is an error in BSSolv, because the newer python-certifi version 2018... is noarch, the old 2015.xxx is x86_64 and BSSolv is probably doing the sorting wrong or something like that.

Unfortunately, debugging BSSolv problems in OBS is next to impossible for anyone not mls :-)

mmohring:
Hi, I think this issue is the same as: #4121 . Newer versions are not fetched, resulting in a resolv error. When you switch off DoD, and fetch missing pkgs by hand, and call rescan of the repo, it works. This means the DoD pkg dependency does not produce the same result as the normal repo scan.

mmohring:
I looked into the old entry, it is from Nov 8, 2017 and got never fixed. The problem now is that the public OBS from SUSE is now hit by this bug, and you cannot use the workaround I had described. Gets time now to fix this finally.....

seife:
#4121 is probably the same issue:
42.3 GA had `mysql-community-server-5.6.36-25.1.x86_64.rpm`, while the update repo has first `mysql-community-server-errormessages-5.6.37-27.1.x86_64.rpm`, but then switches to `mysql-community-server-errormessages-5.6.38-30.1.noarch.rpm`

So it is again a x86_64 -> noarch switch, that perl-BSSolv cannot handle properly.
Unfortunately, debugging perl-BSSolv is beyond my capabilities.

mmohring:
My guess would be that `<archfilter>list of archs</archfilter>` from a DoD descriptor is not considered when looking for the highest version from a DoD repo. The question is if it can happend that a package can have two subpackages with the same name and arch noarch as well as a processor arch. If this cannot happend then it must look for the highest version in the all repos in the archfilter list.

mmohring:
I have now created a patch in #8064. Please review it.

seife:
MR #8064 fixes the issue in my private OBS instance.

seife:
[...]

Issue #4121, opened 2017-11-08
[backend] DoD package resolution failure


mmohring:
Issue/Feature description
====================
Name resolution fails for a recently updated package in openSUSE Leap 42.2 and 42.3.

Expected result
============
Should behave as if the packages were not loaded by DoD but when present in :full directory.

How to Reproduce
===============
1. recently, mysql-community-server was patched in openSUSE:Leap:42.{2,3}:Update
2. obs-server has a build requirment to mysql-community-server
3. If you build obs-server against openSUSE:Leap:42.{2,3}:Update, it results in a expansion error

Further information
===============
* Error message: mysql-community-server-errormessages = 5.6.38 needed by mysql-community-server
* I use a private instance of git master as new as b.o.o

If I download the mysql-community-server-errormessages.rpm into :full and remove the DoD from our local openSUSE:Leap:42.{2,3}:Update.xml and run a rescan repository, the expansion error goes away.

mmohring:
I am using:

`
    <download arch="x86_64" url="http://download.opensuse.org/update/leap/42.2/oss" repotype="rpmmd">
      <archfilter>x86_64,noarch</archfilter>
    </download>
`

and

`
    <download arch="x86_64" url="http://download.opensuse.org/update/leap/42.3/oss" repotype="rpmmd">
      <archfilter>x86_64,noarch</archfilter>
    </download>
`

as download repositories.

mmohring:
I have noticed this with other updates, dito. E.g. java-1.8 security update vs. mozilla-nss.

mmohring:
@mls this is an issue for you i guess. Do you need more info ? Can you provide some info how DoD generates different package dependencies as the normal binary package parser does ? Does it use the repo info to generate the :full.solv database ?

seife:
Hi Martin,
in order to have @mlschroe be able to reproduce and fix the issue, you probably need to provide exact instructions on how to reproduce the issue on a plain, clean obs install.

See https://lists.opensuse.org/opensuse-buildservice/2016-11/msg00020.html and https://lists.opensuse.org/opensuse-buildservice/2016-11/msg00024.html to see how the similar issue #2296 was solved :-)

seife:
[...]

PR #8064, opened 2019-08-09
[backend] fix wrong highest version computation for DoD repos

mmohring:
This fix relates to issue  #4121, and potential duplicate #8012. Update repositories can contain multiple versions of a package, and a package can change its architecture over the lifetime. Make sure the highest version is picked independent of the package architecture when computing the DoD cache and cache file.

### DIFF
--- a/src/backend/bs_dodup
+++ b/src/backend/bs_dodup
@@ -370,7 +370,7 @@ sub cmppkg {
 }
 
 sub addpkg {
-  my ($cache, $p, $archfilter) = @_; 
+  my ($cache, $p, $archfilter, $doddata) = @_; 
 
   return unless $p->{'location'} && $p->{'name'} && $p->{'arch'};
   return if $archfilter && !$archfilter->{$p->{'arch'}};
@@ -378,7 +378,8 @@ sub addpkg {
     return if grep {$p->{'name'} =~ /^$_/s } @$BSConfig::dodupblacklist;
   }
   $p->{'path'} = delete $p->{'location'};
-  my $key = "$p->{'name'}.$p->{'arch'}";
+  # Index the cache with the DoD architecure (scheduler) name. not the package architecture name, a package can change its architecture between multiple versions
+  my $key = "$p->{'name'}.$doddata->{'arch'}";
   return if $cache->{$key} && cmppkg($cache->{$key}, $p) > 0;   # highest version only
   $cache->{$key} = $p; 
 }
@@ -393,7 +394,7 @@ sub parsemetadata {
       $archfilter->{$_} = 1 unless delete $archfilter->{"-$_"};
     }
   }
-  Build::Repo::parse($doddata->{'repotype'}, $file, sub { addpkg($cache, $_[0], $archfilter) }, 'addselfprovides' => 1, 'normalizedeps' => 1, 'withchecksum' => 1, 'testcaseformat' => 1);
+  Build::Repo::parse($doddata->{'repotype'}, $file, sub { addpkg($cache, $_[0], $archfilter, $doddata) }, 'addselfprovides' => 1, 'normalizedeps' => 1, 'withchecksum' => 1, 'testcaseformat' => 1);
   $baseurl =~ s/\/$//;
   $cache->{'/url'} = $baseurl;
   BSUtil::store("$file.parsed", $file, $cache);


mmohring:
I have tested it with my local install, which has a lot of DoD repos (and especially use the Update repositories which can have multiple versions of a package). Running the patch resulted in downloading a lot of .noarch packages for the respective scheduler arch. It caused also a lot of rebuilds, because now a lot of packages with arch noarch are detected as the highest version. Lots of strange build errors are gone.

mmohring:
@mlschroe can you please take a look.

mmohring:
I can think of two scenarios: 

- A package is build for two or more architectures. It is published in a multiarch repo. At least one subpackage is noarch. Then the release counts of the .noarch subpackage could be different, but the version counts are the same if this package is build by OBS and both builds succeed.
- A package is build, and published for two archs. The the package is updated with a newer version. The package produces subpackages with arch and noarch. Now one of builds for the two architectures fails. Then we have the old version of the arch and noarch package of the failed build as well as the new version of the package on the other arch and noarch published. I this case the DoD would download the newer version of the .noarch subpackages, and not both.

This means that DoD misses the capability to handle multiversion packages for some cases. If you assume that multiversions of a packages should not be in an "official" update repo, then the presented fix is fine. Anyway, current DoD implementation cannot handle multiversions of packages.

mmohring:
@hennevogel you added the labels and the label bug to one of the issues. Any news ?

seife:
@mmohring this fixes my problem, too. I had to remove the cloud_update dod repo (that had caused the problems on my side) and restart obsdodup to get it to pick up the change.
A plain "obs_admin --rescan-dod SLES12-SP4" was not enough. Or I was not patient enough ;-)

Thanks! 👍

mmohring:
> @mmohring this fixes my problem, too. I had to remove the cloud_update dod repo (that had caused the problems on my side) and restart obsdodup to get it to pick up the change.
> A plain "obs_admin --rescan-dod SLES12-SP4" was not enough. Or I was not patient enough ;-)
> 
> Thanks! +1

The plan "obs_admin --rescan-dod SLES12-SP4" does not help because DoD thinks the meta data and the meta data cache are not changed and thus does not do a recalculation (manual deletion of the files helps). If the upstream repo changes, the recomputation will take place and trigger the package downloads.

hennevogel:
> @hennevogel you added the labels and the label bug to one of the issues. Any news ?

If you mean reviewing this PR, not really my cup of tea, sorry.

mmohring:
> > @hennevogel you added the labels and the label bug to one of the issues. Any news ?
> 
> If you mean reviewing this PR, not really my cup of tea, sorry.

So we are waiting for Michael to review it ?

seife:
Well, the fix seems not complete.
Without it:

```
> osc buildinfo -d images x86_64|grep python-requests
  <error>unresolvable: nothing provides python-certifi &gt;= 2017.4.17 needed by python-requests</error>
added python-requests@SLES12-SP4/cloud_update because of python2-kiwi:python-requests
added python-py@SLES12-SP4/cloud_update because of python-requests:python-py
added python-chardet@SLES12-SP4/cloud_update because of python-requests:python-chardet &gt;= 3.0.2
added python-urllib3@SLES12-SP4/cloud_update because of python-requests:python-urllib3 &gt;= 1.21.1
```

With the patch:

```
 osc buildinfo -d images x86_64|grep python-requests
  <bdep name="python-requests" version="2.11.1" release="6.28.1" arch="noarch" project="SLES12-SP4" repository="standard" />
added python-requests@SLES12-SP4/standard because of python2-kiwi:python-requests
added python-requests@SLES12-SP4/standard because of python2-kiwi:python-requests
```

So now it is using python-requests from the (lower priority) non-update repo, which is wrong.

And today the resolver failed differently: it took autoyast2 from SLES12-SP4/standard (wrong), but autoyast2-installation from SLES12-SP4/server_update (correct), but then these two did not match and thus requirements were not met.

```
  <error>unresolvable: nothing provides autoyast2-installation = 3.2.31 needed by autoyast2</error>
...
added autoyast2-installation@SLES12-SP4/server_update because of patterns-sles-base:autoyast2-installation
...
added autoyast2@SLES12-SP4/standard because of patterns-sles-base:autoyast2
```

So something is broken differently now.

seife:
This on top of #8064 seems to help. The .src.rpm in the repos seem to confuse the calculation.

```
--- bs_dodup.orig       2019-08-12 14:31:20.000000000 +0000
+++ bs_dodup    2019-08-15 06:40:46.496797017 +0000
@@ -380,6 +380,7 @@
   $p->{'path'} = delete $p->{'location'};
   # Index the cache with the DoD architecure (scheduler) name. not the package architecture name, a package can change its architecture between multiple versions
   my $key = "$p->{'name'}.$doddata->{'arch'}";
+  $key = "$p->{'name'}.$p->{'arch'}" if $p->{arch} eq "src";
   return if $cache->{$key} && cmppkg($cache->{$key}, $p) > 0;   # highest version only
   $cache->{$key} = $p; 
 }
```

seife:
Thinking about it: the .src.rpm will always have the same version as either .noarch.rpm or .${arch}.rpm.
If it comes last, it will be the one added to the %cache and thus in the doddata file.
But it will not be usable for installation later. => `nothing provides autoyast2-installation = 3.2.31 needed by autoyast2`.
So just skipping the addition of .src.rpm to %cache and normalizint arch to scheduler arch will fix that, but still this seems like just a workaround.

The real bug is IMVHO in BSSolv not correctly resolving. Unfortunately @mlschroe does not comment on that ;-)

marcus-h:
On 2019-08-15 00:28:19 -0700, Stefan Seyfried wrote:
<SNIP>
> The real bug is IMVHO in BSSolv not correctly resolving.
>
See [1] (if the arches mismatch, the version is ignored. I cannot say
if this is a bug or a feature, though).

[1] https://github.com/openSUSE/open-build-service/issues/4121#issuecomment-519683672

mmohring:
@seife what you are suggesting is:

- Use the original doddup, but with your .src.rpm ignore
- Fix perl-BSSolv/BSSolv.xs according to the comment in #4121

seife:
No, original dodup without changes.
Then .src gets sorted in %cache as "foo.src", noarch as "foo.noarch"...

src (and nosrc) is ignored by BSSolv already.
The comment in #4121 makes BSSolv use the correct (IMO) resolution, by just looking for the highest version, regardless of arch.

mmohring:
Hi, update on the situation: I am using my original patch without any erl-BSSolv/BSSolv.xs fix. The src.rpm and nosrc.rpm are filtered away in my settings because I use archfilters. In one of my comments I had written scenarios when we might need multiple versions of a package. You added another scenarion when you might also need multiple versions of a package, but in this case the arch and the noarch version. My solution would be to not fix perl-BSSolv/BSSolv.xs but to let DoD behave like a "normal" repo, where you also do not need a fix.

mmohring:
One last on :full: it usually has no version numbers inside. Also, there is currently only one :full per scheduler arch. So DoD should exactly do that, like a normal non DoD repo. Since DoD is not done in perl-BSSolv/BSSolv.xs, the fix should be done in the DoD code.

seife:
Martin, the bs_dodup does not download anything. It just indexes the repomd for bs_sched and bs_srcserver to then download it and populate :full 

If you now filter out things in bs_dod (which is what your patch does, it only stores the highest version, whatever arch it is, even .src or .nosrc as $name.$schedulerach), you are actually taking away options for the solver to decide later what things to download.

The problem with the current unpatched (neither dod nor BSSolv.xs) code is, that bs_dod puts in all package versions (foo.x86_64, foo.noarch, foo.src, foo.ppc64le in my case) into doddata.
Later in bs_sched/bs_srcserver, BSSolv.xs simply does not consider foo.noarch at all (that's the code @marcus-h mentioned), and this leads to the deps being non-resolvable.

Your solution is to just not store everything in doddata, but only the highest version.
This works, but is fragile (my autoyast2 nonresolvable problem came from "src" arch being stored as the only highest version, and being unusable for solving, and see my comment here https://github.com/openSUSE/open-build-service/issues/8012#issuecomment-521219436 about possible problems if multiple arches have different versions, and noarch is better for one arch (x86_64), but not the other (ppc64le). These things are not easily solvable in a generic way in bs_dod, but are necessary for a universal solution.

The solver in BSSolv OTOH must be able to solve this anyway (not sure if it really can do it yet, but zypper can, and BSSolv should be at least as good as zypper :-) ), so my patch to BSSolv.xs to just not skip over noarch looks much better IMVHO. I'm not sure why it was added, there is no comment in BSSolv.xs and it was present from the first version in github, but it seems obviously incorrect considering the failures we see here.

mmohring:
Hi Stefan, I know what you say. It maybe that my wording was too sloppy. I did not talk about bs_dod. Nor did I say my patch is good. I talked about that DoD should mimic the bahaviour of a normal repo. The problem is now, as you say, that normal repos are *push* by the builds, and DoD is *pull*, by the dependencies. I know that bs_dodup does not load the packages. Did I say that ? And in the end, what counts, is that it just works (TM).

mmohring:
> The problem with the current unpatched (neither dod nor BSSolv.xs) code is, that bs_dod puts in all package versions (foo.x86_64, foo.noarch, foo.src, foo.ppc64le in my case) into doddata.
> Later in bs_sched/bs_srcserver, BSSolv.xs simply does not consider foo.noarch at all (that's the code @marcus-h mentioned), and this leads to the deps being non-resolvable.

I understood the code like that that only those pkgs are considered later which are listed in archfilter. Do you have archfilter not set ?

mmohring:
> If you now filter out things in bs_dod (which is what your patch does, it only stores the highest version, whatever arch it is, even .src or .nosrc as $name.$schedulerach), you are actually taking away options for the solver to decide later what things to download.

What I was saying is that usually, only one version of a package is stored in :full. The files in :full are not versioned, not for a normal repo nor for a DoD repo. If the same version or a higher version of a package with .noarch or a different arch overwrites another one, shit happens also for a non DoD repo. So if there is multiple versions, it happens that for a normal repo they overwrite each other (last build rules) or in case of DoD, it might happen that a different arch overwrites another one. What I wanted to say is that both cases are not handled.

And last but not least, I do not have your repo meta data for SLE 12 SP4 at hand, So I cannot check if what happens is *normal* or not.

seife:
I do not have archfilter set.

:full repos are already split into x86_64, ppc64le, ..., so the only thing that can happen is, that a noarch overwrites an $arch package if it is newer (normal built :full) or if it has a higher version (dod repo), so this is fine.

The problem we are hitting here is that without the BSSolv fix, the packages are not even considered for download if noarch has a higher version than $arch, so the dod repo will never even populate/build its :full repo.

IF the x86_64 python-certifi-2016.x gets overwritten by the noarch python-certifi-2017.x in dod :full, then everything is fine, mission accomplished ;-)

a ppc64le python-certifi will not overwrite the x86_64 one, because it is `/srv/obs/build/SLES12-SP4/cloud_update/$ARCH/:full/`, only that $ARCH/:full contains both $ARCH and noarch rpms.

(Paths etc above are from memory, I'm off on vacation for next week, so can only check on my installations when I return to the office on August 26)

mlschroe:
[back from vacation]

I prefer to have this fixed in perl-BSSolv. That way we have all architectures in the dod data and may be able to do redhat style builds in the (far) future.

(To be honest, I can't remember why we do that highest version filtering in bs_dodup at all, maybe to somewhat reduce the size of the solv file)

(We need to fix perl-BSSolv anyway because the code in there is clearly incorrect.)

mmohring:
> [back from vacation]
> 
> I prefer to have this fixed in perl-BSSolv. That way we have all architectures in the dod data and may be able to do redhat style builds in the (far) future.
> 
> (To be honest, I can't remember why we do that highest version filtering in bs_dodup at all, maybe to somewhat reduce the size of the solv file)
> 
> (We need to fix perl-BSSolv anyway because the code in there is clearly incorrect.)

Hi Michael, I thought that you were on vacation. I had headaches because DoD did not work correctly for so long. So its good to know where the problem is. So you do the official fix in perl-BSSolv. We test it. If all works I drop this pull request and we can close the bug entries.

mmohring:
> > I prefer to have this fixed in perl-BSSolv. That way we have all architectures in the dod data and may be able to do redhat style builds in the (far) future.
If I understand it correctly, we would still need the arch filters if we want to be able to do the redhat multiarch builds. So it means we cannot drop archfilter then.

adrianschroeter:
latest perl-BSSolv should cover this problem already.

Issue #8072, opened 2019-08-09
BsRequest#show diff display height


hennevogel:
To work around #7752 we have removed redrawing of CodeMirror to fit height of the file we display. 

We need to get rid of this workaround and provide the same functionality again. Preferably by using something else than the heavy codemirror for displaying diffs.

Issue #7752, opened 2019-06-13
Request#show performance issue


hennevogel:
Visiting a request with many individual diffs for files is very slow, so slow that the browser warns you about a slow running script.

hennevogel:
After looking into performance my suspicion is that this problem is in the [function](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/assets/javascripts/webui2/packages.js#L17-L22) that refreshes the editors.

hennevogel:
Guess I was wrong. Removing this whole function locally didn't change anything.

coolo:
and the performance is worse than with bento? I didn't notice it

coolo:
The numbers are obvious, the graphs only point out that bootstrap relayouts. I assume that's because it's responsive? 

bento:
![Screenshot_20190613_185505](https://user-images.githubusercontent.com/1067203/59451974-03253580-8e0d-11e9-8184-f990cc55f7ac.png)

bootstrap:
![Screenshot_20190613_185636](https://user-images.githubusercontent.com/1067203/59451981-07515300-8e0d-11e9-9941-3959d602af16.png)

DavidKang:
@hennevogel, I guess is because of this [function](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/views/webui2/webui/request/show.html.haml#L96-L100)

dmarcoux:
To reproduce this locally:

1. Branch the package `kernel-source` from Factory with the OBS interconnect, so `openSUSE.org:openSUSE:Factory`.
2. Submit a request to another project.
3. Go on the request#show page of this request
4. The page is slow

dmarcoux:
[I found the culprit](https://github.com/openSUSE/open-build-service/blob/ed1d1e56b53065af0eb2f8d96ce9bc4698012df5/src/api/app/views/webui2/shared/_editor.html.haml#L56-L57)

dmarcoux:
![](https://github.trello.services/images/mini-trello-icon.png) [Address Feedback (Sprint 60)](https://trello.com/c/h3HkOuTp/1171-address-feedback-sprint-60)

PR #8073, opened 2019-08-09
Use a "big_editor" for SR/MR diff views

sknorr:
This change because ...:
* Having multiple scrollbars on the same page is annoying
* In most cases, the diffs to the .spec/.changes files are not that
  long but seeing just ~20 lines of a file's diff at a time is really
  limiting the ability to get an overview of the changes
* You can fold the diffs if you don't want to scroll


hennevogel:
Sorry, this doesn't work. See #7752 why. We have an issue to track this at #8072 and already gave it the second highest priority. 

Thank you for your contribution anyway @sknorr :bouquet:

sknorr:
Too bad. But happy you're working on it.

PR #8074, opened 2019-08-09
Use 'robohash' style of default Gravatars instead of 'wavatar'

sknorr:
Robohash looks (IMO) more modern and is less pixelated.

For comparison: 

* Wavatar example: https://www.gravatar.com/avatar/00000000000000000000000000000000?d=wavatar&f=y&size=500
* Robohash example: https://www.gravatar.com/avatar/00000000000000000000000000000000?d=robohash&f=y&size=500


sknorr:
Thanks!

hennevogel:
thank you for your contribution @sknorr :bouquet:

PR #8082, opened 2019-08-12
Improve constrast of added/removed lines in CodeMirror

dmarcoux:
Somehow WAVE still seems to flag the new version as not having a good enough contrast. I double-checked on WebAIM and it's good. Refer to the links below.

Black on red: https://webaim.org/resources/contrastchecker/?fcolor=212529&bcolor=F5C2C7
Black on green: https://webaim.org/resources/contrastchecker/?fcolor=212529&bcolor=B2E0B2

Before:
codemirror-before

After:
codemirror-after

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/cm2/bootstrap.scss
+++ b/src/api/app/assets/stylesheets/webui2/cm2/bootstrap.scss
@@ -40,18 +40,18 @@
   .cm-error {color: $danger;}
 
   .cm-negative {
-    color: $danger;
+    color: inherit;
   }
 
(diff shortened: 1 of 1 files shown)


dmarcoux:
For comparison purposes:

With an alpha of `0.1`:
codemirror-0-1

With an alpha of `0.2`:
codemirror-0-2

My proposal with `0.3`: (it is already added in the PR's description)
codemirror-after

hennevogel:
Can we close this in favor or #8084?

dmarcoux:
> Can we close this in favor or #8084?

Yep, closing this.

PR #8083, opened 2019-08-12
Update colors for obs_green, bootstrap-info and bootstrap-success

krauselukas:
The colors had a bad contrast in many areas of the OBS
webui

**Before**
Screenshot-2019-8-12 ctris(1)

**After**
Screenshot-2019-8-12 ctris(2)

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/bootstrap_variables/colors.scss
+++ b/src/api/app/assets/stylesheets/webui2/bootstrap_variables/colors.scss
@@ -6,25 +6,22 @@ $gray-300: #e9ecef;
 $gray-400: #aeaeae;
 $gray-500: #878c92;
 
-$obs_green: #008000;
-$obs_blue: #069;
-$obs_yellow: #f0ad4e;
-$obs_cyan: #007a7c;
-
-$primary: $obs_green;
-$secondary: $obs_blue;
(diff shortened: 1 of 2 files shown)


dmarcoux:
I like it. I wonder if we need multiple greens, like we have `primary` and `success` which are the same now. I don't have an issue with this. However, we still have `$green` which is another green we use sometimes. Should we remove it along with all the other non-obs colors?

Let's gather some more feedback. @openSUSE/obs-frontend, what do you think?

hennevogel:
I think it also works better with the darker header/footer.

dmarcoux:
@krauselukas and I removed the different tones of green. It created inconsistencies since sometimes we used `$obs_green` and in other cases, we used `$green`. We now define `$green` and use it everywhere (for `$primary` and whenever needed in our CSS). Same thing for the other colors like `$obs_blue`, `$obs_yellow` and the rest.

krauselukas:
[Address Wave Recommendations Regarding Contrast/Colors](https://trello.com/c/iZGln7WX/1200-address-wave-recommendations-regarding-contrast-colors)

dmarcoux:
Coveralls is still running after one hour... and changes to stylesheets won't affect OBS Package Build, so I'm merging this.

PR #8084, opened 2019-08-12
Use coderay instead codemirror to present read-only diffs

### LINKED ISSUE #8072: BsRequest#show diff display height

vpereira:
To offer an options  to the users to check a set of one or more diffs, we were using CodeMirror, a javascript based Editor.

However when the request has more than a few diffs, it starts to get
slow. As drop-in solution I'm proposing to go  with [CodeRay](https://github.com/rubychan/coderay).

Fixes #8072

How it looks:

Screenshot_2019-08-14_11-59-44

Benchmark loading 50 diffs in one request:

With CodeMirror:
```
 Completed 200 OK in 61ms (Views: 3.8ms | ActiveRecord: 7.4ms | Backend: 17.4ms)
```
With CodeRay (with tables or div, no real performance difference among them):
```
Completed 200 OK in 47ms (Views: 0.9ms | ActiveRecord: 5.7ms | Backend: 17.3ms)
```

TODO:
- [x] probably fix some test case
- [x] adapt the css to our needs
- [x] add to the changes for 2.11

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/coderay.scss
+++ b/src/api/app/assets/stylesheets/webui2/coderay.scss
@@ -1,0 +1,131 @@
+.CodeRay {
+  background-color:#fff;
+  border: 1px solid silver;
+  color: black;
+}
+.CodeRay pre {
+  margin: 0px;
+}
+
+span.CodeRay { white-space: pre; border: 0px; padding: 2px; }
+
+table.CodeRay { border-collapse: collapse; width: 100%; padding: 2px; }
+table.CodeRay td { padding: 2px 4px; vertical-align: top; }
+
+.CodeRay .line-numbers {
+  background-color: hsl(180,65%,90%);
+  color: gray;
+  text-align: right;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  user-select: none;
+}
+.CodeRay .line-numbers a {
+  background-color: hsl(180,65%,90%) !important;
+  color: gray !important;
+  text-decoration: none !important;
+}
+.CodeRay .line-numbers pre {
+  word-break: normal;
+}
+.CodeRay .line-numbers a:target { color: blue !important; }
+.CodeRay .line-numbers .highlighted { color: red !important; }
+.CodeRay .line-numbers .highlighted a { color: red !important; }
+.CodeRay span.line-numbers { padding: 0px 4px; }
+.CodeRay .line { width: 100%; }
+.CodeRay .code { width: 100%; }
+
+.CodeRay .debug { color: white !important; background: blue !important; }
+
+.CodeRay .annotation { color:#007 }
+.CodeRay .attribute-name { color:#b48 }
+.CodeRay .attribute-value { color:#700 }
+.CodeRay .binary { color:#549 }
+.CodeRay .binary .char { color:#325 }
+.CodeRay .binary .delimiter { color:#325 }
+.CodeRay .char { color:#D20 }
+.CodeRay .char .content { color:#D20 }
+.CodeRay .char .delimiter { color:#710 }
+.CodeRay .class { color:#B06; font-weight:bold }
+.CodeRay .class-variable { color:#369 }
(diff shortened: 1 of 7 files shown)


hennevogel:
I think you need to 'port' the [.spec](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/assets/stylesheets/webui/application/cm2/suse.scss#L48) support we have.

src/api/app/assets/stylesheets/webui2/coderay.scss
```diff
@@ -0,0 +1,132 @@
+.CodeRay {
+  background-color: #efefef;
+  font-family: Monaco, "Courier New", "DejaVu Sans Mono", "Bitstream Vera Sans Mono", monospace;
```
hennevogel: Is this based on the other monospace fonts we use?

vpereira: @hennevogel I adapted to use the `font-family` to exactly what we use with code mirror. I did change the  `background-color` to white.. My first try was similar to github where Comment and Code block have different colors.

vpereira: I re-generated the default stylesheet and now it's using the default `font-stack` (the screenshot was updated). Please take a look @hennevogel

src/api/app/helpers/application_helper.rb
```diff
@@ -1,3 +1,8 @@
 # Methods added to this helper will be available to all templates in the application.
 module ApplicationHelper
+  def render_diff(content)
+    # rubocop:disable Rails/OutputSafety
+    CodeRay.scan(content, :diff).div(line_numbers: :table, css: :class).html_safe
```
hennevogel: Hmmmmmm. Diffs are user input right?

hennevogel:
Long lines make the diff div overflow. 

Screenshot from 2019-08-13 10-40-37

hennevogel:
```
osc -A https://obs-reviewlab.opensuse.org/vpereira-use_coderay copypac openSUSE.org:openSUSE:Factory kernel-source home:Admin
osc -A https://obs-reviewlab.opensuse.org/vpereira-use_coderay sr home:Admin kernel-source MaintenanceProject -m "blah"
```

https://obs-reviewlab.opensuse.org/vpereira-use_coderay/request/show/2

src/api/app/helpers/application_helper.rb
dmarcoux: Yes

vpereira: @hennevogel your comments is regarding user input validation x marking html as safe?

hennevogel: yeah, if I can create whatever diff I want  and you mark it as html safe, can I inject XSS?

vpereira: but did you try to inject it? I tried it and in a `pre` tag didn't work. I tried too to close the `pre` tag as `</pre><script>alert('foo')</script>` and didn't work too.  Any idea @hennevogel ?

vpereira:
@hennevogel does it look better now?

vpereira:
@hennevogel could you please take another look?

src/api/app/helpers/application_helper.rb
hennevogel: nope, if you tried, all is fine.

hennevogel:
> There are no line numbers displayed anymore

Ah that's not really true. There are, but they are about the patched file and not the diff 'file'. nice!

hennevogel:
So the only thing left is measure coderay+table vs. coderay+span.

vpereira:
> So the only thing left is measure coderay+table vs. coderay+span.

I did it already, I even updated the PR description... there is - performance wise, no difference.

PR #8085, opened 2019-08-12
Improve constrast of red/green for build results in striped tables

dmarcoux:
The new version makes the page pass the WAVE contrast checks. The changes are minor since the contrast was almost already fine. It just needed a small boost :wink:

Before:
project-monitor-before

After:
project-monitor-after


hennevogel:
coveralls hangs, see codecov. merging.

PR #8086, opened 2019-08-12
Improve contrast for all states of build results in striped tables

dmarcoux:
Before:
project-monitor-before

After:
project-monitor-after


dmarcoux:
Merging now since CSS changes don't affect `OBS Package Build` and `coverage/coveralls`

Issue #8068, opened 2019-08-09
User.last_logged_in_at isn't updated


hennevogel:
This is supposedly fixed in #1704 but still we have users where this doesn't happen. For instance the user `favogt` who logs in daily through the UI...

PR #8098, opened 2019-08-14
Update last_logged_in_at when a user logs in via proxy

saraycp:
When users logged in via proxy `last_logged_in_at` and `login_failure_count` were not updated. Now we do it every time we check the user.

We also set RabbitMQ messages when needed.

Fixes: #8068

### DIFF
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -177,9 +177,7 @@ class Webui::WebuiController < ActionController::Base
       User.session = user_checker.find_or_create_user!
 
       if User.session!.is_active?
-        User.session!.update_user_info_from_proxy_env(request.env)
-        User.session!.mark_login!
-        send_login_information_rabbitmq(:success)
+        User.session!.update_login_values(request.env)
       else
         User.session!.count_login_failure
         session[:login] = nil
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -178,7 +178,10 @@ class Webui::WebuiController < ActionController::Base
 
       if User.session!.is_active?
         User.session!.update_user_info_from_proxy_env(request.env)
+        User.session!.mark_login!
+        send_login_information_rabbitmq(:success)
       else
+        User.session!.count_login_failure
         session[:login] = nil
         User.session = User.find_nobody!
         send_login_information_rabbitmq(:disabled) if previous_user != User.possibly_nobody.login
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -281,7 +281,7 @@ class User < ApplicationRecord
       mark_login!
       self
     else
-      update_attributes!(login_failure_count: login_failure_count + 1)
+      count_login_failure
       nil
     end
   end
@@ -860,6 +860,10 @@ class User < ApplicationRecord
 
   def mark_login!
     update_attributes(last_logged_in_at: Time.now, login_failure_count: 0)
+  end
+
+  def count_login_failure
+    update_attributes(login_failure_count: login_failure_count + 1)
   end
 
   def send_metric_for_beta_change
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -387,22 +387,6 @@ class User < ApplicationRecord
     "#{home_project_name}:branches:#{branch}"
   end
 
-  # updates users email address and real name using data transmitted by authentication proxy
-  def update_user_info_from_proxy_env(env)
-    proxy_email = env['HTTP_X_EMAIL']
-    if proxy_email.present? && email != proxy_email
-      logger.info "updating email for user #{login} from proxy header: old:#{email}|new:#{proxy_email}"
-      self.email = proxy_email
-      save
-    end
-    return unless env['HTTP_X_FIRSTNAME'].present? && env['HTTP_X_LASTNAME'].present?
-    realname = env['HTTP_X_FIRSTNAME'].force_encoding('UTF-8') + ' ' + env['HTTP_X_LASTNAME'].force_encoding('UTF-8')
-    return unless self.realname != realname
-
-    self.realname = realname
-    save
-  end
-
   #####################
   # permission checks #
   #####################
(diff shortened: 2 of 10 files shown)


hennevogel:
Hm but this isn't thread save right?

src/api/app/models/user.rb
```diff
@@ -862,6 +862,10 @@ def mark_login!
     update_attributes(last_logged_in_at: Time.now, login_failure_count: 0)
   end
 
+  def count_login_failure
+    update_attributes(login_failure_count: login_failure_count + 1)
```
eduardoj: This has changed from `update_attributes!` to `update_attributes`. Was this change intentional?

src/api/app/models/user.rb
```diff
@@ -859,7 +843,38 @@ def combined_rss_feed_items
...
+    # updates user's email and real name using data transmitted by authentication proxy
+    self.email = env['HTTP_X_EMAIL'] if env['HTTP_X_EMAIL'].present?
+    self.realname = proxy_realname if proxy_realname
+
+    self.last_logged_in_at = Time.zone.today
+    self.login_failure_count = 0
+
+    # If the values stay the same after the assignations above, nothing will be updated in database.
+    # So, we won't consider this is a new 'Log in' process.
+    return if changes.blank?
```
hennevogel: Isn't that taken care of by `ActiveRecord`?

```ruby
irb(main):002:0> u = User.last
  User Load (0.9ms)  SELECT  `users`.* FROM `users` ORDER BY `users`.`id` DESC LIMIT 1
=> #<User id: 3, created_at: "2019-08-01 10:58:53", updated_at: "2019-08-01 11:09:09", last_logged_in_at: "2019-08-01 10:58:53", login_failure_count: 0, login: "hans", email: "hvogel@hennevogel.de", realname: "", password_digest: nil, deprecated_password: "d7f33cef468756cde99e31f6b8256505", deprecated_password_hash_type: "md5", deprecated_password_salt: nil, adminnote: nil, state: "confirmed", owner_id: nil, ignore_auth_services: false, in_beta: false, in_rollout: true>
irb(main):006:0> u.email               
=> "hvogel@hennevogel.de"
irb(main):007:0> u.email = "hvogel@hennevogel.de"
=> "hvogel@hennevogel.de"
irb(main):008:0> u.save 
   (0.6ms)  BEGIN
  User Exists (0.8ms)  SELECT  1 AS one FROM `users` WHERE `users`.`login` = 'hans' AND `users`.`id` != 3 LIMIT 1
   (0.5ms)  COMMIT
=> true
irb(main):009:0> u.email = "hvogel@hennevogel.com"
=> "hvogel@hennevogel.com"
irb(main):010:0> u.save
   (0.5ms)  BEGIN
  User Exists (0.9ms)  SELECT  1 AS one FROM `users` WHERE `users`.`login` = 'hans' AND `users`.`id` != 3 LIMIT 1
  User Update (6.2ms)  UPDATE `users` SET `updated_at` = '2019-08-15 12:11:22.589734', `email` = 'hvogel@hennevogel.com' WHERE `users`.`id` = 3
   (4.7ms)  COMMIT
=> true
```

No `UPDATE` call in the first save....

src/api/app/services/webui_controller_service/user_checker.rb
```diff
@@ -19,6 +19,7 @@ def find_or_create_user!
       if user.exists?
         User.find_by!(login: user_login)
       else
+        # This will end up in a before_action(on: :create) that updates last_logged_in_at.
```
hennevogel: Will it? When?

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -177,10 +177,11 @@ def check_user
       User.session = user_checker.find_or_create_user!
 
       if User.session!.is_active?
-        User.session!.update_user_info_from_proxy_env(request.env)
+        User.session!.update_login_values(request.env)
       else
         session[:login] = nil
         User.session = User.find_nobody!
+        User.session!.count_login_failure
```
hennevogel: Hm, I don't get it. `User.session!` will just always raise here right?

src/api/app/models/user.rb
saraycp: It was intentional for consistence, because values like `last_logged_in_at`, `login_failure_count`and other login values are updated in different places of the code and we don't use `save!` (or `update_attributes!` which indeed  calls `save!`).

And I don't see the point of raising an exception if those values didn't update correctly :thinking: Do you think it is?

Examples:

```
  def mark_login!                                                                                                                                        
    update_attributes(last_logged_in_at: Time.now, login_failure_count: 0)                                                                               
  end
```

```
  def update_user_info_from_proxy_env(env)                                                                                                            
    proxy_email = env['HTTP_X_EMAIL']                                                                                                                 
    if proxy_email.present? && email != proxy_email                                                                                                   
      logger.info "updating email for user #{login} from proxy header: old:#{email}|new:#{proxy_email}"                                               
      self.email = proxy_email                                                                                                                        
      save                                                                                                                                        
 [...]
```

src/api/app/models/user.rb
saraycp: Yes, in fact I'm based on what you explain when I call save in line 877. That `save` won't perform an UPDATE in database when the value doesn't changed. However, I really need to know (before saving) if there are any changes and only if so I send the message to RabbitMQ and write in the log if email is modified.

I could rewrite this like:

```
  def update_login_values(env)
    # updates user's email and real name using data transmitted by authentication proxy
    self.email = env['HTTP_X_EMAIL'] if env['HTTP_X_EMAIL'].present?
    self.realname = proxy_realname if proxy_realname

    self.last_logged_in_at = Time.zone.today
    self.login_failure_count = 0

    if changes.any?
      # At this point some login value changed, so a successful log in is tracked
      RabbitmqBus.send_to_bus('metrics', 'login,access_point=webui value=1')
    end

    if changes.keys.includes?('email')
      logger.info "updating email for user #{login} from proxy header: old:#{email}|new:#{env['HTTP_X_EMAIL']}"
    end

    save
  end
```

But using `return if changes.blank?` (equivalent to `save && return if changes.blank?` when no UPDATE performed) I avoid going through lines that I'm not going to execute.

src/api/app/services/webui_controller_service/user_checker.rb
saraycp: Sorry, there is a typo in the comment. It should have been `before_validation(on: :create)` not `before_action(on: :create)`. And following the trace:  `User.create_user_with_fake_pw!` calls `create!`, that calls `before_validation(on: :create)` which updates `last_logged_in_at` among others.

[create_user_with_fake_pw!](https://github.com/openSUSE/open-build-service/blob/fb6f6202ea5c161607dd24f459a0cb9292b0b6dc/src/api/app/models/user.rb#L170)

[before_validate](https://github.com/openSUSE/open-build-service/blob/fb6f6202ea5c161607dd24f459a0cb9292b0b6dc/src/api/app/models/user.rb#L148)

src/api/app/models/user.rb
eduardoj: Fine for me, then. :+1:

src/api/app/controllers/webui/webui_controller.rb
saraycp: First, I've moved `User.session!.count_login_failure` two lines up because we need `User.session` to be the real user and not `nobody`. If I'm not wrong, at this point the real user is there but its state is not `confirmed`/not active. Shouldn't we consider it as an error?

src/api/app/models/user.rb
saraycp: Forget about the first comment. I have rewritten it and I think is more clear now. Please have a look @hennevogel

hennevogel:
@saraycp I think this is good to go, can you give it one last rebase?

saraycp:
@hennevogel rebased and "WIP" removed.

PR #8103, opened 2019-08-14
Typography - patchinfo show

DavidKang:
We removed the `small` class from summary and block reason. This information was too attenuated because of the small font size and and light gray colour.

#### Before
Screenshot_2019-08-16 Open Build Service

#### After
Screenshot_2019-08-16 Open Build Service

### DIFF
--- a/src/api/app/helpers/webui/patchinfo_helper.rb
+++ b/src/api/app/helpers/webui/patchinfo_helper.rb
@@ -22,7 +22,7 @@ module Webui::PatchinfoHelper
   private
 
   def header_title(patchinfo, text, list)
-    haml_tag(:h3, class: 'mb-0') do
+    haml_tag(:h3) do
       haml_tag(:span, text, title: list)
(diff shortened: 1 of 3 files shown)


dmarcoux:
The first element which caught my eyes on this page is the block reason, but it should be the header. I suggest changing the color and especially the font-weight of the block reason so the hierarchy is respected.

vpereira:
> The first element which caught my eyes on this page is the block reason, but it should be the header. I suggest changing the color and especially the font-weight of the block reason so the hierarchy is respected.

Hm, if i remember correct we changed this to exactly give more emphasis on the block reason, right? Look the what the @dmarcoux from the past said https://github.com/openSUSE/open-build-service/pull/7416#discussion_r276271014

dmarcoux:
Well, this Dany from the past didn't know much about typography for the web.

So I would say the hierarchy should be the following:

1. Header
2. Block reason
3. ?
4. Profit

PR #8104, opened 2019-08-14
Enable RFS in bootstrap for responsive font-sizes

krauselukas:
Just a preview for you to see the webui with responsive font-sizes enabled. In this example it scales down to 14px at max (default is to stop at 16px, which needs to be discussed). It starts the scaling at large screens (960px width). Check it out in the review app :)

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/bootstrap_variables/fonts.scss
+++ b/src/api/app/assets/stylesheets/webui2/bootstrap_variables/fonts.scss
@@ -1,1 +1,10 @@
 $seventy-five-chars: 75ch;
+
+// enable responsive fonts in bootstrap (RFS)
+$enable-responsive-font-sizes: true;
+
+// stop scaling down at a font-size of 16px (default 1.25rem)
+$rfs-base-font-size: 1rem;
+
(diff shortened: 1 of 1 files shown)


krauselukas:
Improve the readability of the webui by scaling
the font-sizes depending on the screen size.
The proportions between headings and regular text will be
more harmonic on smaller screens.

Based on research the rfs scaling factor is set
to 2.5 in order to be close to the "major second"
scale, on small screens (360px width).
The rfs-base-font-size is set to 1rem (16px) to
not scale further down, since it's the recommended
font-size for regular text across all devices and screen sizes
to keep the readability.

I compared the recommended font-sizes from this article <https://typecast.com/blog/a-more-modern-scale-for-web-typography> with the "major second" scale here <https://type-scale.com/>.

Screenshot-2019-8-16 Type Scale - A Visual Calculator

They are very close, so I used the outcome to set the rfs scaling factor accordingly:

Welcome   Open Build Service
Public Projects   Open Build Service

krauselukas:
[Fluid/Responsive Web Typography](https://trello.com/c/IqTltQzy/1216-fluid-responsive-web-typography)

PR #8105, opened 2019-08-14
Add more status indicators for finished build in live build log

hellcp:
Screenshot from 2019-08-14 20-22-32
Screenshot from 2019-08-14 21-07-44
Helps to tell if the builds have succeeded without opening the tabs


src/api/app/assets/javascripts/webui2/live_build_log.js
```diff
@@ -70,12 +72,24 @@ $.extend(LiveLog.prototype, {
...
+    else if (status === 'succeeded') {
+      this.status.html('Build succeeded');
+      this.indicatorStatus('succeeded');
+      this.faviconStatus(this.faviconSucceeded);
+    }
+    else {
+      this.status.html('Build finished');
+      this.indicatorStatus('finished');
+      this.faviconStatus(this.faviconFinished);
+    }
```
danidoni: Hey! Looks good!

Just only one thing. Looks like [CodeClimate](https://codeclimate.com/github/openSUSE/open-build-service/pull/8105) is complaining about this code being too similar and a refactor might be desirable... 

What about adding a function that returns different js objects based on the value of `status` and then assign those objects to `this.status`, `this.indicatorStatus` and `this.faviconStatus`? Doing it that way the similar code CodeClimate is complaining about is isolated into one function...

Something like this:
```js
finish: function(status) {
   ...
   var statusDetails = this.buildStatusDetails(status);
   this.status.html(statusDetails.content);
   this.indicatorStatus(statusDetails.indicator);
   this.faviconStatus(statusDetails.favicon);
   ...
```

PR #8106, opened 2019-08-15
fix crash in owner search

adrianschroeter:
a binary which got published via a snapshoted repo, but the
original container got removed later ....


vpereira:
@adrianschroeter are you ok with the proposed changes?

adrianschroeter:
sure (more complex to read, but I don't mind)

PR #8109, opened 2019-08-15
Use hyphens: auto, just when supported

### LINKED ISSUE #8100: hyphens: auto usage on chrome
We started to use `hyphens: auto` while [caniuse.com](https://caniuse.com/#search=hyphens) says support for chrome is only partial, especially the auto part. This broke for instance comments on chrome.

Screenshot from 2019-08-14 15-21-18

vpereira:
Unfortunately the hyphens: auto, doesn't work on Chrome desktop properly
therefore we are falling back to word-break: break-word when hyphens
aren't supported

We can see in the following screenshots how it looks like:

On Firefox:

Screenshot_2019-08-15_11-31-29-firefox

On Chrome:

With the alternative to `word-break: break-word`: (`word-break: normal;overflow-wrap: anywhere`) 
suggested in https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

Screenshot_2019-08-15_11-30-19-word-break-normal

On Chrome:

With the `word-break: break-word`

Screenshot_2019-08-15_11-28-31-word-break

So even if `word-break: break-word` is deprecated it is a temporary solution while Chrome doesn't support it well.

Fixes #8100

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/comments.scss
+++ b/src/api/app/assets/stylesheets/webui2/comments.scss
@@ -1,6 +1,16 @@
 #comments {
   .media .media-body {
-    hyphens: auto;
+
+    // hyphens isn't supported on Desktop chrome
+    // https://caniuse.com/#search=hyphens
+    @supports(hyphens:auto) {
+      hyphens: auto;
+    }
+    @supports not(hyphens: auto) {
+      // break-word is depreacted. But works on Chrome well
+      // https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
+      word-break: break-word;
+    }
 
     p:not(:first-child) {
       max-width: $seventy-five-chars;


src/api/app/assets/stylesheets/webui2/comments.scss
```diff
@@ -1,6 +1,16 @@
...
+
+    // hyphens isn't supported on Desktop chrome
+    // https://caniuse.com/#search=hyphens
+    @supports(hyphens:auto) {
+      hyphens: auto;
+    }
+    @supports not(hyphens: auto) {
+      // break-word is depreacted. But works on Chrome well
+      // https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
+      word-break: break-word;
```
dmarcoux: As we discussed together, this is a gamble. Will Chrome support `hyphens: auto` before `word-break: break-word` is not supported anymore? I prefer to find another solution.

vpereira: Hi @dmarcoux 

Gambling is one way to see it. I see it as reducing the numbers of devices which won' t be able to see the page correctly to the minimum as possible.

What would be the alternative? Stop to use `hyphens: auto` since it's not supported by all major browsers and stick with the `break-all` and show it badly in all browsers? Not sure if it's better.

hennevogel: And maybe we can find the long term solution **after** we have fixed the page for people? :cowboy_hat_face:

vpereira: I agree with that @hennevogel

dmarcoux: I would use `break-all` for Chrome and if Chrome users complain, well, they can use Firefox loll

dmarcoux: Let's re-open https://github.com/openSUSE/open-build-service/issues/7796

vpereira: I don't see a problem to use temporarily `word-break: break-word` if that's the best technical solution.. Following your reasoning, because we have the `support` guard, if Chrome stops to support the `break-word` before it supports fully `hyphens`, then the user can still move to Firefox anyway..

vpereira:
All the tests are passing, just circleci isn't reporting back to github https://circleci.com/workflow-run/968c111e-27ac-4748-b3d7-eeb9c640e048

PR #8112, opened 2019-08-15
Set max line-length for obs description at root page

krauselukas:
To be consistent with the maximum line-length for text at other places
of obs.

**Before**
Screenshot-2019-8-15 Open Build Service(1)

**After**
Screenshot-2019-8-15 Open Build Service

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/card-component.scss
+++ b/src/api/app/assets/stylesheets/webui2/card-component.scss
@@ -8,6 +8,14 @@
     h1, h2, h3, h4, h5, h6 {
       color: $primary;
     }
+
+    .description {
+      max-width: $seventy-five-chars;
+    }
+
+    #proceed-list {
+      max-width: $seventy-five-chars;
+    }
   }
 
   .nav-tabs .nav-link.active {
--- a/src/api/app/views/webui2/webui/main/index.html.haml
+++ b/src/api/app/views/webui2/webui/main/index.html.haml
@@ -1,7 +1,7 @@
 - @pagetitle = 'Welcome'
 
 .row
-  .col-md-8
+  .col-md-8.col-lg-7
     .card.mb-3
       .card-body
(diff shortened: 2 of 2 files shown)


dmarcoux:
There's now a big gap above the `Status Monitor` icon. I would suggest also apply this width to the icons' container. It would look like this:

description-icon-width

Then the next step is to resize the container because now it's just too big for no reason. This involves changing the layout.

Another proposal is to bring the icons to the right side of the text on desktop. On smaller devices, the icons would be below the text as they are already right now. I don't know if it would be looking good and make sense, it's just an idea.

krauselukas:
> There's now a big gap above the `Status Monitor` icon. I would suggest also apply this width to the icons' container. It would look like this:
> 
> description-icon-width
> 
> Then the next step is to resize the container because now it's just too big for no reason. This involves changing the layout.
> 
> Another proposal is to bring the icons to the right side of the text on desktop. On smaller devices, the icons would be below the text as they are already right now. I don't know if it would be looking good and make sense, it's just an idea.

@dmarcoux thanks, I will check how it looks

krauselukas:
@dmarcoux I changed the layout a bit, and gave the boxes on the right side a bit more space on large screens.

**Before**
Screenshot-2019-8-15 Open Build Service(3)

**After**
Screenshot-2019-8-15 Open Build Service(2)

hennevogel:
> I changed the layout a bit

These things don't make much sense if you are not thinking about what this page should do for people. Please stop doing this unless we redesign the page.

dmarcoux:
@krauselukas Did you try with the next to the text? So they fill the gap? It might be nicer.

We either leave a gap or redesign, but this would be later, because it's out of the scope for this deliverable.

dmarcoux:
Let's go with the gap for now. We'll redesign the homepage in another deliverable.

PR #8113, opened 2019-08-15
Typography - Search

dmarcoux:
Stick to default font for results' description in the search (same thing for the search issues page)

Before:
search-before

After:
search-after

---

Improve text hierarchy in the search owners page. For the user cards, we stick with the emphasis on the user names, so this explains why the user logins are smaller. 

Before:
search-owners-before

After:
search-owners-after


dmarcoux:
[Address typography - Search](https://trello.com/c/8g4BUdJ9/1212-address-typography-search)

krauselukas:
@dmarcoux LGTM, but maybe you could add the "why" part to the commit message (I think there should be one right?)

dmarcoux:
@krauselukas Please have another look.

PR #8123, opened 2019-08-18
Remove unused images and update some

hellcp:
Searched through the codebase to remove the images that were no longer used for anything.

Added some more icons and updated the old ones to fill out the distributions view with icons:
Screenshot from 2019-08-18 05-46-02

Added current openSUSE sponsors to the images directory, so they can get an update on the reference server (and updated the AMD logo) ;)


hellcp:
Basically nothing, some images are larger, and the list is up to date with the state of [openSUSE sponsors](https://en.opensuse.org/Sponsors).

hennevogel:
> the list is up to date with the state 

@adrianschroeter is that okay?

src/api/app/views/webui2/webui/repositories/distributions.html.haml
```diff
@@ -7,8 +7,8 @@
     %h3.mb-3 Add Repositories to #{@project}
```
dmarcoux: I suggest we remove the `.mb-3` class since the images are now a bit taller (from 26px to 40px) and the spacing between this header and the header of the first distribution is already good with the `.pt-3` class.

hellcp: Yup, seems fine to me

hellcp:
The sponsors are in the reference server configuration, to actually set them you need to create a data that will be displayed by the https://github.com/openSUSE/open-build-service/blob/master/src/api/app/views/webui2/webui/main/_sponsors.html.haml in the rails configuration. It's not documented anywhere as far as I have seen and there is no default config for that in the repo.

hellcp:
Yeah, it seems that default config of sponsors was removed with https://github.com/openSUSE/open-build-service/commit/f5435b148d22c3ac973161718f07b1a58074acf5 and now only the reference server configuration has that set.

Issue #8116, opened 2019-08-16
New user interface: hard to distinguish success/failure


msvec:
# Issue Description
In the new interface it is rather hard to distinguish between "succeeded" and "failed" as the two colors are rather close or the background makes it hard to distinguish between the two.

For example in this overview I don't see any difference, unless I expect the screen closely:
https://build.opensuse.org/project/monitor/LibreOffice:6.3

dmarcoux:
Here's an example:
![Screenshot_20190819_162846](https://user-images.githubusercontent.com/1102934/63273635-9dd34200-c29e-11e9-862d-88af2e292ee3.png)

@msvec If you don't mind answering, do you happen to be colorblind? I am asking because I believe some tones of green and red can be seen as the same for certain types of colorblindness.

msvec:
@dmarcoux Actually I am partly :) E.g. in your screenshot I can see the distinction, but it's rather small and not easy to spot on the first view as it was in the design before.

vpereira:
hi @msvec , thank you for your issue, sorry for the inconvenience. Could you tell me if this issue used to happen in the old UI too? I mean, was literally ported from the old UI, with more padding and bigger fonts, but I suspect that both wouldn't affect it.

dmarcoux:
@msvec and @vpereira I added a screenshot from the old UI in my [first comment](https://github.com/openSUSE/open-build-service/issues/8116#issuecomment-522601887) to compare.

msvec:
Actually in the previous design it was easier to differentiate the red / green colors. I can't tell what exactly is causing the difference, though.

dmarcoux:
The tones of red and green changed in the new UI. They seem to be worse for certain colorblindness.

krauselukas:
@msvec hi, could you maybe compare the screenshots in this PR https://github.com/openSUSE/open-build-service/pull/8146 and tell us if the changes brings improvement for you? :) Unfortunately the review app doesn't show this page, please let me know if the screenshots are not sufficient to tell if it helps or not.

hennevogel:
https://obs-reviewlab.opensuse.org/krauselukas-issue_8116/project/monitor/home:Admin

msvec:
It's not really better, the old UI (as shown in the screenshot above) in comparison feels easier to differentiate between the two colors.

krauselukas:
[...]

PR #8131, opened 2019-08-19
WIP Improve constrast on category for striped tables 

DavidKang:
#### Before
Screenshot_2019-08-19 Open Build Service(4)

#### After
Screenshot_2019-08-19 Open Build Service(3)


vpereira:
Probably the same issue but in different table https://github.com/openSUSE/open-build-service/issues/8116. Is that possible to go with a global solution for the issue?

PR #8132, opened 2019-08-19
Adjust typography and elements size and position in group show page

saraycp:
Adjusts typography

**Before:**

Screenshot-2019-8-19 Open Build Service(2)

**After:**
Screenshot-2019-8-20 Open Build Service(6)

And also sets the same width for all the boxes:

**Before:**

group_members_beforee

**After:**

group_members_after

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/groups.scss
+++ b/src/api/app/assets/stylesheets/webui2/groups.scss
@@ -1,5 +1,14 @@
 #group-users {
   .group-user {
-    min-width: 26rem;
+    width: 19.7rem;
+
+    @include media-breakpoint-up(xl) {
+      width: 21.7rem;
+    }
+
+    @include media-breakpoint-up(xxl) {
+      width: 22.4rem;
+    }
   }
 }
+
--- a/src/api/app/helpers/webui/user_or_groups_roles_helper.rb
+++ b/src/api/app/helpers/webui/user_or_groups_roles_helper.rb
@@ -1,9 +1,13 @@
 module Webui::UserOrGroupsRolesHelper
   def display_name(object)
     if object.is_a?(User) && object.realname.present?
-      "#{object.name} (#{object.login})"
+      content_tag :span do
+        concat "#{object.name} ("
+        concat content_tag(:i, object.login)
+        concat ')'
+      end
     else
-      object.name
+      content_tag(:span, object.name)
     end
   end
 
--- a/src/api/app/views/webui2/webui/groups/_group_members.html.haml
+++ b/src/api/app/views/webui2/webui/groups/_group_members.html.haml
@@ -25,11 +25,10 @@
                   class: 'group-member-removal float-right', title: 'Remove member from group') do
                   %i.fas.fa-sm.fa-times-circle.text-danger
               .card-body.ml-2
-                %b
-                  = link_to(user_show_path(user)) do
-                    = display_name(user)
+                = link_to(user_show_path(user)) do
+                  = display_name(user)
 
-                .custom-control.custom-checkbox
+                .custom-control.custom-checkbox.mt-2
                   = check_box_tag("maintainer_#{user.login}", true, group.maintainer?(user), class: 'custom-control-input group-maintainership',
                     name: :maintainer, disabled: !write_access, data: { method: :patch,
                     url: group_user_path(group_title: group.title, user_login: user.login), remote: true })
--- a/src/api/app/views/webui2/webui/groups/_group_members.html.haml
+++ b/src/api/app/views/webui2/webui/groups/_group_members.html.haml
@@ -16,15 +16,15 @@
     .d-flex.flex-wrap
       - group.users.each do |user|
         .card.m-1.p-2.group-user
+          - if write_access
+            = link_to(group_user_path(group_title: group.title, user_login: user.login), data: { method: :delete, remote: true },
+              id: user.login, class: 'group-member-removal text-right', title: 'Remove member from group') do
+              %i.fas.fa-sm.fa-times-circle.text-danger
           .row.no-gutters
-            .col-md-2
+            .col-md-3
               = image_tag_for(user, size: 80, custom_class: 'align-self-center')
-            .col-md-10
-              - if write_access
-                = link_to(group_user_path(group_title: group.title, user_login: user.login), data: { method: :delete, remote: true },
-                  class: 'group-member-removal float-right', title: 'Remove member from group') do
-                  %i.fas.fa-sm.fa-times-circle.text-danger
-              .card-body.ml-2
+            .col-md-9
+              .card-body.py-1
                 = link_to(user_show_path(user)) do
                   = display_name(user)
 


saraycp:
[Address typography - Group Show](https://trello.com/c/I99IAu5p/1215-address-typography-group-show)

src/api/app/assets/stylesheets/webui2/groups.scss
```diff
@@ -2,4 +2,9 @@
   .group-user {
     min-width: 26rem;
   }
+
+  a.user-realname {
```
dmarcoux: Applying a `max-width` to the `.group-user` class above would achieve the same effect. It's also easier to follow what is affecting the width since it would be all under one class.

src/api/app/views/webui2/webui/groups/_group_members.html.haml
```diff
@@ -25,11 +25,10 @@
...
                   %i.fas.fa-sm.fa-times-circle.text-danger
               .card-body.ml-2
-                %b
-                  = link_to(user_show_path(user)) do
-                    = display_name(user)
+                = link_to(user_show_path(user), class: 'user-realname') do
+                  = display_name(user)
 
-                .custom-control.custom-checkbox
+                .custom-control.custom-checkbox.mt-3
```
dmarcoux: I believe `.mt-3` isn't needed.

src/api/app/views/webui2/webui/groups/_group_members.html.haml
```diff
@@ -25,11 +25,10 @@
                   class: 'group-member-removal float-right', title: 'Remove member from group') do
                   %i.fas.fa-sm.fa-times-circle.text-danger
               .card-body.ml-2
-                %b
-                  = link_to(user_show_path(user)) do
-                    = display_name(user)
+                = link_to(user_show_path(user), class: 'user-realname') do
+                  = display_name(user)
```
dmarcoux: I wonder if we should split the name and the login. Could one of them be in italic for example?

src/api/app/views/webui2/webui/groups/_group_members.html.haml
saraycp: I believe we really need some more space between the name and the roles list. Maybe `mt-2` is enough. See the difference:

**No margin:**
Screenshot-2019-8-20 Open Build Service

**mt-2:**
Screenshot-2019-8-20 Open Build Service(1)

**mt-3:**
Screenshot-2019-8-20 Open Build Service(2)

src/api/app/assets/stylesheets/webui2/groups.scss
saraycp: That was the first thing I tried but we'll have the same problem, boxes with different widths, unless we set `min-width`and `max-width` to the same value (equivalent to setting `width`).

But it's true my solution is not nice, let me try something different ...

src/api/app/views/webui2/webui/groups/_group_members.html.haml
saraycp: Sometimes we only show the name, sometimes name and login... and introducing some italic text here... :thinking: I would say it's going to be less homogeneous than it was before.

src/api/app/views/webui2/webui/groups/_group_members.html.haml
dmarcoux: I prefer `mt-2` or even `mt-1` would be enough in this case.

src/api/app/assets/stylesheets/webui2/groups.scss
dmarcoux: Having boxes with different width is fine. If some are smaller, we can fit more on one row.

What is the goal for having the boxes all with the same width? Because if this is what we want here, setting the max-width on an element inside the `.group-user` container is not obvious. We might as well set the `width` to a specific value. I would go against this, I believe we should take advantage of the flexibiliy provided by the flexbox (we use `.d-flex` on the container of those `.group-user` containers)

src/api/app/views/webui2/webui/groups/_group_members.html.haml
dmarcoux: Alright, no worries.

src/api/app/assets/stylesheets/webui2/groups.scss
saraycp: > What is the goal for having the boxes all with the same width?

Let's say... harmony? For me it looks chaotic with different widths, but perhaps it's a very personal feeling. So, I'll keep it like it was before.

src/api/app/views/webui2/webui/groups/_group_members.html.haml
saraycp: Wait, I assumed you wanted to split in different lines. I gave it a try with italic but in the same line and I like it. What do you think @dmarcoux?

Screenshot-2019-8-20 Open Build Service(4)

It also looks nice in
Screenshot-2019-8-20 Open Build Service(3)

dmarcoux: I like it :+1: !

saraycp:
@dmarcoux please have another look and read the comments again.

src/api/app/helpers/webui/user_or_groups_roles_helper.rb
```diff
@@ -1,9 +1,13 @@
 module Webui::UserOrGroupsRolesHelper
   def display_name(object)
     if object.is_a?(User) && object.realname.present?
-      "#{object.name} (#{object.login})"
+      content_tag :span do
+        concat "#{object.name} ("
+        concat content_tag(:i, object.logini.to_s)
```
dmarcoux: Typo in `object.login`.

`.to_s` isn't needed since `login` is already a string.

src/api/app/helpers/webui/user_or_groups_roles_helper.rb
```diff
@@ -1,9 +1,13 @@
...
+        concat ')'
+      end
     else
-      object.name
+      content_tag(:span, object.name.to_s)
```
dmarcoux: `.to_s` isn't needed since `name` is already a string.

src/api/app/assets/stylesheets/webui2/groups.scss
hennevogel: Boxes, the same size are way easier to process because I don't have to notice the change in 'rhythm

dmarcoux: Good point. Let's set the width on the `.group-user` container then.

saraycp:
I finally fixed the same width for all the boxes. But I adapted the width value to the different breakpoints in order to make the most of the space.

Please check the Gifs in the updated description.

src/api/app/assets/stylesheets/webui2/groups.scss
```diff
@@ -1,5 +1,27 @@
...
+    .group-member-removal {
+      position: absolute;
+      top: 0;
+      right: 0.5rem;
+    }
+  }
+}
+
+@include media-breakpoint-up(xl) {
+  #group-users {
```
dmarcoux: `#group-users` doesn't need to be repeated in the media queries.

It can be like this:
```
#group-users {
  .group-user {
    width: 19.7rem;
  }

  @include media-breakpoint-up(xl) {
    width: 21.7rem;
  }

  @include media-breakpoint-up(xxl) {
    width: 22.4rem;
  }
}

```

src/api/app/views/webui2/webui/groups/_group_members.html.haml
```diff
@@ -17,19 +17,20 @@
       - group.users.each do |user|
         .card.m-1.p-2.group-user
           .row.no-gutters
```
dmarcoux: This row will be empty if a user doesn't have write access, so rendering it isn't needed in this case.

saraycp: Sure! I'll move the condition out so the row won't be there if no icon is shown.

Anyway, the solution is not ideal, but this is the explanation:

The problem here is the position of the (X) icon to delete the member. 

Before, the icon was inside the second column and it didn't stay in the top right corner in small screens. And now that the box is fitting better for smaller screens it looks even uglier. Example:

Screenshot-2019-8-21 Open Build Service

My first approach was putting the icon nomatter-where inside `.group-user` and playing with absolute position to fix it to the top right corner. However, elements with absolute position are not detected by Capybara.

Options I had:
a) Find some workaround to make the element clickable for Capybara.
b) Set the icon in its own row even if it takes its space on the top.

I choosed b). Do you have a better idea @dmarcoux?

Screenshot-2019-8-21 Open Build Service(2)

src/api/app/views/webui2/webui/groups/_group_members.html.haml
```diff
@@ -16,20 +16,19 @@
     .d-flex.flex-wrap
       - group.users.each do |user|
         .card.m-1.p-2.group-user
+          - if write_access
+            = link_to(group_user_path(group_title: group.title, user_login: user.login), data: { method: :delete, remote: true },
+              id: user.login, class: 'group-member-removal d-inline text-right ', title: 'Remove member from group') do
```
dmarcoux: It's a nice solution :+1:

We can remove the class `d-inline` since a link is already an inline element.

saraycp: Of course! I only forgot to delete it after testing different things.

src/api/app/views/webui2/webui/groups/_group_members.html.haml
```diff
@@ -16,20 +16,19 @@
     .d-flex.flex-wrap
       - group.users.each do |user|
         .card.m-1.p-2.group-user
+          - if write_access
+            = link_to(group_user_path(group_title: group.title, user_login: user.login), data: { method: :delete, remote: true },
+              id: user.login, class: 'group-member-removal text-right ', title: 'Remove member from group') do
```
dmarcoux: There's one unneeded space at the end of the classes

dmarcoux:
Good job!

PR #8135, opened 2019-08-20
project_monitor.js: sort unfinished status codes before finished codes

andreas-schwab:


dmarcoux:
Hey @andreas-schwab,

Thank you for submitting this PR. Could you explain why this is needed? Could you also add screenshots to preview what this changes?

andreas-schwab:
It doesn't make sense having to scroll through hundreds of empty rows.

hennevogel:
>  empty rows

Hm, when do you run into a project that has hundreds of packages that don't have a build status code?

andreas-schwab:
You see it in any larger project.

hennevogel:
> You see it in any larger project.

I don't think this does what you think it does. If a package has no build results for any repo/arch (if it's build disabled) it's not shown on the monitor page at all. If a package has a result for any repo/arch, the whole row is shown. So there are no empty rows, rows always contain something.

If you sort a **column**, empty cells are at the bottom (with or without your change). 

So what is it you actually try to achieve? :)

andreas-schwab:
> So there are no empty rows, rows always contain something.

Nope.

> If you sort a column, empty cells are at the bottom (with or without your change).

Nope.

PR #8136, opened 2019-08-20
Fix clearfix class in login form

eduardoj:
We want to use the `clearfix` CSS class as it is described in the [Bootstrap documentation](https://getbootstrap.com/docs/4.3/utilities/clearfix/):

> Easily clear floats by adding `.clearfix` **to the parent element**.

It is also included a little refactoring related to the submit button.

### DIFF
--- a/src/api/app/views/layouts/webui2/_login_form.html.haml
+++ b/src/api/app/views/layouts/webui2/_login_form.html.haml
@@ -19,5 +19,4 @@
           = label_tag(:password, 'Password', class: 'text-light')
           = password_field_tag(:password, nil, placeholder: 'Password', required: true, class: 'form-control')
         .clearfix
-          .float-right
-            = submit_tag('Log In', class: 'btn btn-success')
+          = submit_tag('Log In', class: 'btn btn-success float-right')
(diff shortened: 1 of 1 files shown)


src/api/app/views/layouts/webui2/_login_form.html.haml
```diff
@@ -12,12 +12,11 @@
...
-        .float-right
-          = submit_tag('Log In', class: 'btn btn-success')
         .clearfix
+          .form-group
+            = label_tag(:username, 'Username', class: 'text-light')
+            = text_field_tag(:username, nil, placeholder: 'User Name', required: true, class: 'form-control')
+          .form-group
+            = label_tag(:password, 'Password', class: 'text-light')
+            = password_field_tag(:password, nil, placeholder: 'Password', required: true, class: 'form-control')
+          = submit_tag('Log In', class: 'btn btn-success float-right')
```
dmarcoux: `.clearfix` is only for the submit button, so I believe it should be above this line. The other form controls are not floating.

eduardoj: Right! Done. Please review again.

PR #8137, opened 2019-08-20
Typography - User/Group Tasks

dmarcoux:
Deal with long words properly in all browsers

Before:
before

After:
after

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/datatables.scss
+++ b/src/api/app/assets/stylesheets/webui2/datatables.scss
@@ -26,6 +26,6 @@ table.dataTable.table-sm {
 
   td { 
     white-space: normal !important;
-    word-break: break-all;
+    @extend .text-break;
   }
 }


dmarcoux:
[Address typography - User / Group Tasks](https://trello.com/c/fPhoPCCt/1217-address-typography-user-group-tasks)

src/api/app/assets/stylesheets/webui2/datatables.scss
```diff
@@ -26,6 +26,18 @@ table.dataTable.table-sm {
 
   td { 
     white-space: normal !important;
-    word-break: break-all;
+
+
+    // hyphens isn't supported on Desktop chrome
```
DavidKang: We are using the same code in more places/files. Maybe is a good idea to extend/inherit it. We could do someting like:

``` scss

%hyphens-support {
    // hyphens isn't supported on Desktop chrome
    // https://caniuse.com/#search=hyphens
    @supports(hyphens: auto) {
      hyphens: auto;
    }

    @supports not(hyphens: auto) {
      // break-word is deprecated. But works on Chrome well
      // https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
      word-break: break-word;
    }
}

td {
 @extend %hyphens-support;
}

Comment (dmarcoux) [reply src/api/app/assets/stylesheets/webui2/datatables.scss]: Good suggestion! It looks like it might not be enough, see my [comment](https://github.com/openSUSE/open-build-service/pull/8133#issuecomment-523025213).

However, we definitely need to stop repeating ourselves. I won't merge this PR until we have something fully working.

Comment (dmarcoux) [reply src/api/app/assets/stylesheets/webui2/datatables.scss]: In the end, I'll go with only `break-word`, so this won't be needed.

**Comment by dmarcoux** [comment]
@DavidKang Please have another look

File: src/api/app/assets/stylesheets/webui2/datatables.scss [hunk 2]
```diff
@@ -26,6 +26,9 @@ table.dataTable.table-sm {
 
   td { 
     white-space: normal !important;
-    word-break: break-all;
+
+    // break-word is deprecated, but it works on all major browsers, unlike hyphens
+    // https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
+    word-break: break-word;
```
DavidKang: @dmarcoux, what about using [.text-break](https://getbootstrap.com/docs/4.3/utilities/text/#word-break) class from bootstrap?

PR #8138, opened 2019-08-20
Remove 'btn-sm' class from filter buttons

eduardoj:
**Before**

Screenshot from 2019-08-20 15-25-03

**After**

Screenshot from 2019-08-20 15-25-26


eduardoj:
[Address typography - Configuration](https://trello.com/c/A1DIEngF/1214-address-typography-configuration)

PR #8146, opened 2019-08-21
Remove zebra striped rows from datatables 

### LINKED ISSUE #8116: New user interface: hard to distinguish success/failure

krauselukas:
People with limited color vision have issues to read
colored text/items on the stronger gray of the zebra striped
tables.

Fixes #8116

**Before**
Screenshot-2019-8-21 Open Build Service(1)

**After**
Screenshot-2019-8-21 Open Build Service

I would suggest to make the zebra effect a bit lighter. Incase people still having issues to read the content, we still can remove it completly.

Here some interesting article regarding this topic:
https://alistapart.com/article/zebrastripingdoesithelp/

I added a hover effect to the datatables, this makes it easier to keep track of the currently selected row.

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/bootstrap_variables/tables.scss
+++ b/src/api/app/assets/stylesheets/webui2/bootstrap_variables/tables.scss
@@ -1,2 +1,0 @@
-// For better contrast in tables using .table-striped
-$table-accent-bg: $gray-300;
--- a/src/api/app/assets/stylesheets/webui2/webui2.css.scss
+++ b/src/api/app/assets/stylesheets/webui2/webui2.css.scss
@@ -16,7 +16,6 @@
 @import 'bootstrap_variables/colors';
 @import 'bootstrap_variables/spacers';
 @import 'bootstrap_variables/fonts';
-@import 'bootstrap_variables/tables';
 @import 'bootstrap';
 @import 'bootstrap-modal';
 @import "font-awesome-sprockets";
--- a/src/api/app/views/webui2/shared/_patchinfos_table.html.haml
+++ b/src/api/app/views/webui2/shared/_patchinfos_table.html.haml
@@ -1,4 +1,4 @@
-%table.responsive.table.table-sm.table-bordered.w-100#open-patchinfos-table{ title: 'Running Maintenance Updates' }
+%table.responsive.table.table-sm.table-bordered.table-hover.w-100#open-patchinfos-table{ title: 'Running Maintenance Updates' }
   %thead
     %tr
       %th Project
--- a/src/api/app/views/webui2/shared/_patchinfos_table.html.haml
+++ b/src/api/app/views/webui2/shared/_patchinfos_table.html.haml
@@ -1,4 +1,4 @@
-%table.responsive.table.table-sm.table-striped.table-bordered.w-100#open-patchinfos-table{ title: 'Running Maintenance Updates' }
+%table.responsive.table.table-sm.table-bordered.w-100#open-patchinfos-table{ title: 'Running Maintenance Updates' }
   %thead
     %tr
       %th Project
--- a/src/api/app/views/webui2/shared/_requests_table.html.haml
+++ b/src/api/app/views/webui2/shared/_requests_table.html.haml
@@ -1,4 +1,4 @@
-%table.responsive.requests-datatable.table.table-sm.table-striped.table-bordered.w-100{ data: { source: source_url,
+%table.responsive.requests-datatable.table.table-sm.table-bordered.w-100{ data: { source: source_url,
   'page-length': local_assigns[:page_length] }, id: id }
   %thead
     %tr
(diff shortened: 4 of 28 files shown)


src/api/app/assets/stylesheets/webui2/bootstrap_variables/tables.scss
```diff
@@ -1,2 +1,6 @@
 // For better contrast in tables using .table-striped
-$table-accent-bg: $gray-300;
+$table-accent-bg: $gray-100;
+
+table.dataTable tbody tr:hover {
```
hennevogel: Why not [`.table-hover`](https://getbootstrap.com/docs/4.3/content/tables/#hoverable-rows)?

dmarcoux:
hover won't work on mobile.

src/api/app/assets/stylesheets/webui2/bootstrap_variables/tables.scss
hennevogel: Why not [`.table-hover`](https://getbootstrap.com/docs/4.3/content/tables/#hoverable-rows)?

krauselukas: > Hover won't work on mobile.

@dmarcoux I will check if there is a solution for mobile

krauselukas:
> Hover won't work on mobile.

@dmarcoux I will check if there is a solution for mobile

src/api/app/assets/stylesheets/webui2/bootstrap_variables/tables.scss
krauselukas: @hennevogel not sure what you mean exactly, there is not class ```.table-hover```?

dmarcoux: Instead of having this custom rule, `.table-hover` can be used.

krauselukas: Ah okay, get it, okay but I will wait until we agree on having it, since i need to add this class to many tables :D

krauselukas:
I removed the zebra stripes completly, since it doesn't really provides a benifit regarding the readability, and it makes it even harder to distinguish between the elements for people with limited color vision.

I moved the hover part out of this PR and will open another one for this.

**Tables without the zebra stripes**
Screenshot-2019-8-22 Open Build Service(1)
Screenshot-2019-8-22 This is a test project

coolo:
>  it doesn't really provides a benifit regarding the readability

Do you have any proof of that? Because I for one find it much easier to follow a coloured line

krauselukas:
@coolo check out those articles: <https://alistapart.com/article/zebrastripingdoesithelp/> and the follow up <https://alistapart.com/article/zebrastripingmoredataforthecase/>

It's kind of a difficult topic, it seems like it slightly influences the readability, but its a lot about personal preferences. The improvement of the readability is small (or in some cases not even measureable) and on the other hand it makes the experience for another group of people worse. Since the improvement through the zebra stripes is rather small for most of the people, I would go the other way and make it better for them who are suffering from it.
I will also open a PR which introduces highlighting of the row on hover, this should support keeping track of the current position inside the table :)

coolo:
Aehm, the articles linked spent a lot of effort to conclude with:

```
The results of the three studies conducted to date suggest that the safest option 
is to shade the alternating, individual rows of your table with a single color. 
```

So while I understand there are people who have a problem with the stripes, I can read very well out of the linked articles that I'm not alone with having a problem in removing it.

coolo:
I actually found the light option the best - also for reading, not as a compromise :)

hennevogel:
@coolo that is why we are going to bring it back on hover...

adrianschroeter:
@krauselukas good idea ... could you also bring back the former meaning of the inverted background? (I guess it was only dropped because of the new strips).

The inverted background was used on all states which are still in flux (supposed to change). These are scheduled/building/finished/blocked/dispatched/signing out of my head (but there used to be a variable to define these).

This is really helpful in some more complex projects on the monitor page, it takes atm additional time to group these in mind (and having the active background only on "blocked" state is quite meaningless)

krauselukas:
> @krauselukas good idea ... could you also bring back the former meaning of the inverted background? (I guess it was only dropped because of the missing contrast with the new stripes).
> 
> The inverted background was used on all states which are still in flux (supposed to change). These are scheduled/building/finished/blocked/dispatched/signing out of my head (but there used to be a variable to define these).
> 
> This is really helpful in some more complex projects on the monitor page, it takes atm additional time to group these in mind (and having the active background only on "blocked" state is quite meaningless)

@adrianschroeter yes I can bring it back, I will check how much work it is, and depending on that adding it to this PR or to another one

krauselukas:
@hennevogel @adrianschroeter I will tackle the background for states that are still in flux in another PR

PR #8147, opened 2019-08-21
Bundle update to allow us to move to rails 6

vpereira:
Some dependencies were still locked in to work just with rails <= 5.9.. 

After update Capybara however one test started to flicker hard, so I disabled it for now. If it gets approved i will open an issue to investigate how to rewrite this test to make it pass again.


src/api/Gemfile.lock
```diff
@@ -52,10 +52,10 @@ GEM
       activerecord (>= 3.0.0)
     addressable (2.6.0)
       public_suffix (>= 2.0.2, < 4.0)
-    airbrake (8.0.1)
-      airbrake-ruby (~> 3.0)
-    airbrake-ruby (3.1.0)
-      tdigest (= 0.1.1)
+    airbrake (9.4.3)
```
hennevogel: Please check if this still works? You can use the OBS Test app on our errbit instance.

vpereira: https://errbit-opensuse.herokuapp.com/apps/58c94e2eeb6526000a000000 works!

PR #8152, opened 2019-08-21
[ci] capybara tests more flexible

M0ses:
This PR contains changes to hand over an specific hostname
to the frontend functional tests so we can use the test suite on an
jump host to test our installation from the outside

### DIFF
--- a/dist/t/spec/support/capybara.rb
+++ b/dist/t/spec/support/capybara.rb
@@ -29,7 +29,7 @@ if hostname.empty?
   hostname = ipaddress
 end
 
-Capybara.app_host = ENV['SMOKETEST_HOST'].nil? ? "https://#{hostname}" : "http://localhost:3000"
+Capybara.app_host = ENV.fetch('SMOKETEST_HOST', "https://#{hostname}")
 
 RSpec.configure do |config|
   config.include Capybara::DSL


dist/t/spec/support/capybara.rb
```diff
@@ -29,6 +29,11 @@
   hostname = ipaddress
 end
 
+if ENV['OBS_HOST']
+  hostname = ENV['OBS_HOST']
+end
+
+
 Capybara.app_host = ENV['SMOKETEST_HOST'].nil? ? "https://#{hostname}" : "http://localhost:3000"
```
hennevogel: `ENV['SMOKETEST_HOST']` is already what you want...

hennevogel:
```suggestion
Capybara.app_host = ENV.fetch("SMOKETEST_HOST", "https://#{hostname}")
```

M0ses:
Nice.
Thanks for your advice.
But I would like to add for consideration that now the interface for users changed.

formerly

`SMOKETEST_HOST=1 bundle exec rspec`

could be used to test `http://localhost:3000`

which now would be

`SMOKETEST_HOST=http://localhost:3000 bundle exec rspec`

I'm not sure if we would like to break the interfaces for testers here.

hennevogel:
> I'm not sure if we would like to break the interfaces for testers here.

[We](https://github.com/os-autoinst/os-autoinst-distri-obs/blob/master/tests/rspec_webui_tests.pm#L22) are the only testers, I'm sure we don't mind ;-)

M0ses:
@hennevogel THX for your approval

Merging because "OBS Package Build" tests seems to fail for other reasons.

PR #8155, opened 2019-08-22
Remove work around for libsass

hennevogel:
This has been addressed upstream.


dist/obs-bundled-gems.spec
```diff
@@ -145,6 +141,9 @@ find %{buildroot} -type f -print0 | xargs -0 grep -l /usr/bin/env | while read f
   chmod a-x $file
 done
 
+# Remove rpaths from files
+chrpath --delete %{buildroot}%_libdir/obs-api/ruby/*/gems/sassc-2.1.0-x86_64-linux/lib/sassc/libsass.so
```
coolo: I would glob the version number as well

hennevogel:
@coolo please have a look again

PR #8165, opened 2019-08-22
Use icons in project monitor page and highlight more relavant states

krauselukas:
The icons will help people who are not able to perceive
the colors, which are indicating the build state, to
recognize the state of the packages easier and faster.

** Before **
Screenshot-2019-8-22 Open Build Service(4)
Screenshot-2019-8-22 home Admin(1)

** After **
Screenshot-2019-8-22 Open Build Service(5)

Screenshot-2019-8-22 home Admin

Related to https://github.com/openSUSE/open-build-service/issues/8116

If this PR receives positiv feedback I could also distinguish between the state broken, failed etc. by adding new icons to our patterns and maybe adding the icons to the build results in the show view as well :)

### DIFF
--- a/src/api/app/assets/javascripts/webui2/project_monitor.js
+++ b/src/api/app/assets/javascripts/webui2/project_monitor.js
@@ -25,7 +25,7 @@ function statusCell(meta, statusHash, tableInfo, projectName, packageName) {
       output += ' data-content="' + status.details + '" data-placement="right" data-toggle="popover"';
     }
   }
-  output += ' class="' + klass + '">' + code + '</a>';
+  output += ' class="' + klass + '">' + '<i class="fa"></i>' + ' ' + code + '</a>';
 
   cellContent.display = output;
   cellContent.value = code;
--- a/src/api/app/assets/stylesheets/webui2/build-results.scss
+++ b/src/api/app/assets/stylesheets/webui2/build-results.scss
@@ -20,14 +20,17 @@
 
 .build-state-succeeded {
   color: $green;
+  & i.fa { @extend .fa-check; }
 }
 
 .build-state-building {
   color: $blue;
+  & i.fa { @extend .fa-cog; }
 }
 
 .build-state-scheduled {
   color: $cyan;
+  & i.fa { @extend .fa-calendar-alt; }
 }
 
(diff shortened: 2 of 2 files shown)


coolo:
If you deemphasize the text of the state I'd suggest to use 4-5 letter abrevation of the state next to a unique icon, so that they align and are easier to overview. The way you made it now is very noisy

vpereira:
The issue regarding colorblindness was that the user wasn't able to distinguish between red or green, but the text (succeeded/failure) could be read, right?

coolo:
The text has always been different but 8116 is about distinguishing the states without reading.

I do agree with the low investment part - but while a redesign is really needed, it's far from easy (not sure how far you guys are with designs, but Ana and I tried for 2 days to come up with a sound proposal covering all use cases and failed). So having a good solution in place might be worth it.

On that note: Adrian mentioned in that other issue that only jobs *in progress* should have a background color, so be careful with that measure.

dmarcoux:
> Adrian mentioned in that other issue that only jobs _in progress_ should have a background color, so be careful with that measure.

@adrianschroeter: Could you explain the reasoning behind this?

krauselukas:
> > Adrian mentioned in that other issue that only jobs _in progress_ should have a background color, so be careful with that measure.
> 
> @adrianschroeter: Could you explain the reasoning behind this?

@dmarcoux have a look into https://github.com/openSUSE/open-build-service/pull/8170 in the old webui a background color used to indicate a state that supose to change, like building, finished etc.

I will not invest more time into this PR for now, since this view needs a bit more planning :)

adrianschroeter:
On Montag, 26. August 2019, 12:04:28 CEST Dany Marcoux wrote:
> > Adrian mentioned in that other issue that only jobs _in progress_ should
> > have a background color, so be careful with that measure.
> @adrianschroeter: Could you explain the reasoning behind this?

It is useful to spot easily moving packages for a certain repo/arch on the 
monitor page. This can easily explain other errors like unresolvables there.

This is one of the major reasons when you look at that page, if you have 
projects with plenty of dependencies between the packages.

We had this in the old theme and it was a key feature there, I notice
it quite heavily in some of the projects that the background in blocked
state has no meaning anymore...

-- 

PR #8168, opened 2019-08-23
Make sure language exists

### LINKED ISSUE #8167: [production][package#show] Symbol or String expected, but NilClass given.
package#show threw this 3 times
<pre>
Symbol or String expected, but NilClass given.
</pre>
[https://build.opensuse.org/package/show/devel:gcc/gcc33](https://build.opensuse.org/package/show/devel:gcc/gcc33)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5d5f9c259df855000dda726e)

hellcp:
Fixes #8167

Skips over coderay entirely if no language is input, to save on having additional processing for that, and additional classes coderay would insert for text.

### DIFF
--- a/src/api/lib/obsapi/markdown_renderer.rb
+++ b/src/api/lib/obsapi/markdown_renderer.rb
@@ -1,4 +1,5 @@
 require 'uri'
+require 'cgi'
 
 module OBSApi
   class MarkdownRenderer < Redcarpet::Render::HTML
@@ -35,10 +36,11 @@ module OBSApi
         link = URI.join(::Configuration.obs_url, link)
       rescue URI::InvalidURIError
       end
-      "<a href='#{link}'#{title}>#{content}</a>"
+      "<a href='#{link}'#{title}>#{CGI.escape_html(content)}</a>"
     end
 
     def block_code(code, language)
+      language ||= :plaintext
       CodeRay.scan(code, language).div(css: :class)
     end
   end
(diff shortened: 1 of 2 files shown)


src/api/lib/obsapi/markdown_renderer.rb
```diff
@@ -39,7 +39,11 @@ def link(link, title, content)
     end
 
     def block_code(code, language)
-      CodeRay.scan(code, language).div(css: :class)
+      if language
+        CodeRay.scan(code, language).div(css: :class)
+      else
+        "<pre><code>#{code}</code></pre>"
```
eduardoj: Thanks for the fast patch!

This would not escape html characters, when there is no language detected. So scripts could be injected.

As described [here](https://github.com/vmg/redcarpet/issues/308#issuecomment-23780412), I would go for something like:

```ruby
def block_code(code, language)
  language ||= :plaintext
  CodeRay.scan(code, language).div(css: :class)
end
```

hellcp: The issue for me is that the difference between plaintext styling of `<pre><code><code></pre>` and coderay's plaintext `<div class="CodeRay"><div class="code"><pre></pre></div></div>`, which is a real waste in comparison.

hennevogel: waste of what?

hellcp: I would really rather escape that manually than rely on coderay for this, just to skip the coderay for non-language codeblocks.

PR #8174, opened 2019-08-24
Move methods only used in its own class to private (II)

eduardoj:
There is no need to make public all the methods of a class, part II.

Continue the work started in #7980.

### DIFF
--- a/src/api/app/models/buildresult.rb
+++ b/src/api/app/models/buildresult.rb
@@ -57,12 +57,6 @@ class Buildresult
     AVAIL_STATUS_VALUES.keys.map(&:to_s)
   end
 
-  def self.code2index(code)
-    index = AVAIL_STATUS_VALUES[code.to_sym]
-    return index if index
-    raise ArgumentError, "code '#{code}' unknown #{AVAIL_STATUS_VALUES.inspect}"
-  end
-
   def self.final_status?(status)
     status.in?(['succeeded', 'failed', 'unresolvable', 'broken', 'disabled', 'excluded'])
   end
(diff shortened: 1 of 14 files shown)


eduardoj:
> I normally think it's good, however i'm just wondering if all those methods are being indirectly tested, specially the one that you removed the test case because it was moved to private. Did you check the spec coverages if those methods are covered indirectly by tests @eduardoj ?

I only removed those lines of the test because I removed a method (`get_html`), that was not used.

In the rest of the cases, moving methods to private shouldn't affect the coverage results.

vpereira:
> I only removed those lines of the test because I removed a method (`get_html`), that was not used.
> 
> In the rest of the cases, moving methods to private shouldn't affect the coverage results.

Yes, sorry if i wasn't clear. I'm not saying that this PR affects the coverage, I'm wondering that you were able to move those methods around without change, or delete, any test, which means that those methods were not directly tested  and we don't know if they are indirectly tested too.

src/api/app/models/buildresult.rb
```diff
@@ -98,4 +92,12 @@ def self.summary(project_name)
 
     local_build_results
   end
+
+  private
+
+  def code2index(code)
```
vpereira: is that code being used anywhere?

eduardoj: Yep, in [this line](https://github.com/openSUSE/open-build-service/blob/25aafabde00ca52eeabde5c756a2628535da42ca/src/api/app/models/buildresult.rb#L94).

vpereira: argh, you are right my grep gave me the following output:

```
src/api/app/models/buildresult.rb
60:  def self.code2index(code)
94:      build.summary.sort! { |a, b| code2index(a.code) <=> code2index(b.code) }
```

and for me it looked like it was being used in the same method, but it's not!

PR #8176, opened 2019-08-26
[WIP] Remove bento views

dmarcoux:
I'm testing a few things in preparation of our next deliverable.


hennevogel:
Screenshot from 2019-08-26 13-43-10

YES

coolo:
wow, you give up?

dmarcoux:
This was only one way of achieving this I had in mind before the planning. We discussed another way in our planning and we'll go with it, so this PR is now closed.

Issue #8183, opened 2019-08-27
Horizontal scroll on monitor page shows the info behind a highlighted row


saraycp:
# Issue Description
Horizontal scroll on monitor page shows the info behind a highlighted row.

![Show home SCabreraPadron   openSUSE Build Service(3)](https://user-images.githubusercontent.com/2581944/63776122-0a72c000-c8e1-11e9-8aac-22fad441d8cc.png)

Expected Result
============
Use an opaque color for highlighted rows, instead of Bootstrap semi-transparent color. 

How to Reproduce
===============
1. Go to 'https://build.opensuse.org/project/monitor/OBS:Server:Unstable'
2. Move horizontal scroll a bit
3. Mouse over package names
4. See error: in rows where the state cells are not empty, the info is shown behind the package name.

Further Information
===============
https://getbootstrap.com/docs/4.0/content/tables/#hoverable-rows

vpereira:
with chrome if I'm not logged in I see other problems:

![Screenshot_2019-08-27_16-24-44](https://user-images.githubusercontent.com/37418/63779656-36913f80-c8e7-11e9-8787-c43c6c7e0fd6.png)
![Screenshot_2019-08-27_16-24-23](https://user-images.githubusercontent.com/37418/63779658-3729d600-c8e7-11e9-8af2-9f5e1d496da9.png)

vpereira:
in a smaller screen I see like that with the mouse over the package name:

![Screenshot_2019-08-27_16-25-30](https://user-images.githubusercontent.com/37418/63779778-693b3800-c8e7-11e9-9954-58aee8235266.png)

hennevogel:
This is because of the [FixedColumns](https://datatables.net/extensions/fixedcolumns/) extension to datatables. The sticky colums is basically another table rendered on top of the other.

hennevogel:
If you move your mouse to the right (the second table beneath the first column) you get the other row highlighted.

hennevogel:
Here is what you can do until we arrive at a redesign of this page:

0. remove the FixedColumns extension for datatables, it's causing too much trouble for little gain
1. Hide the link to this page on everything below the `md` breakpoint inside our app
2. if people arrive in this page from outside of our app: do not render the table, tell them to enable 'desktop mode' in their browser

vpereira:
I would suggest something different to end with this wide table with horizontal scroll:

[...]

PR #8184, opened 2019-08-27
Fix issue with semi-transparent table highlight

saraycp:
When doing horizontal scrolling in the project monitor page and then mouse over, some of the cells show the info behind. This happens because Bootstrap set a semi-transparent color as background and we have an special feature in the monitor page: two tables, one for the package names and the other one for the states which hides behind the first one.

Making the row highlight color opaque, it doesn't look broken.

Fixes:  #8183

**Before:**

63779778 693b3800 c8e7 11e9 9954 58aee8235266 png  PNG Image  1267 × 688 pixels

**After:**

63784290 ba9af580 c8ee 11e9 8494 1ddb4292f755 png  PNG Image  527 × 673 pixels

### DIFF
--- a/src/api/app/assets/stylesheets/webui2/datatables.scss
+++ b/src/api/app/assets/stylesheets/webui2/datatables.scss
@@ -24,8 +24,12 @@ table.dataTable.table-sm {
     bottom: 0.3em;
   }
 
-  td { 
+  td {
     white-space: normal !important;
     @extend .text-break;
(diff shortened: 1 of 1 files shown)


vpereira:
circleci didn't report back to github, but it was everything green https://circleci.com/workflow-run/b3587731-3fcd-4684-b878-5dc2a4ea6521

PR #8140, opened 2019-08-20
Typography user show

DavidKang:
We applied some changes regarding the contrast in the username. Apart from that, we grouped the email with the other elements at the bottom.

#### Before
Screenshot_2019-08-20 Home of Admin(1)

#### After
Screenshot_2019-08-23 Home of Admin(2)

**Beta Program Tooltip:**
Screenshot_2019-08-23 Home of Admin(5)

PR #8193, opened 2019-08-28
Add test for beta program switch for Bootstrap

saraycp:
The original test was skipped in case of Bootstrap (see PR #8140), so a new one specific for Bootstrap is needed.

### DIFF
--- a/src/api/spec/bootstrap/features/webui/users/profile_spec.rb
+++ b/src/api/spec/bootstrap/features/webui/users/profile_spec.rb
@@ -1,0 +1,26 @@
+require 'browser_helper'
+
+RSpec.feature 'User Profile', type: :feature, js: true do
+  let!(:user) { create(:confirmed_user) }
+
+  before do
+    login user
+    visit user_show_path(user)
+  end
+
+  scenario 'public beta program' do
+    within('#beta-form') do
+      find('.custom-control-label').click
+    end
+    expect(page).to have_text("User data for user '#{user.login}' successfully updated.")
+    expect(find('#beta-switch', visible: false)).to be_checked
+    expect(user.reload.in_beta).to be_truthy
+
+    within('#beta-form') do
+      find('.custom-control-label').click
+    end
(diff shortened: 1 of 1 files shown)


src/api/spec/bootstrap/features/webui/users/profile_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'browser_helper'
+
+RSpec.feature 'User Profile', type: :feature, js: true do
```
hennevogel: Why is this a feature and not a controller spec?

saraycp: I wanted to check the JS was running correctly and it really changed the `switch`after clicking. And that implies a change in database.

dmarcoux: To cover the beta program switch, not only what is on the controller.

hennevogel: That is really costly just for checking a javascript oneliner with no logic that can break...

```
$('#beta-switch').on('click', function (){ $('#beta-form').submit(); });
```

dmarcoux: Until the day one of the IDs changes and we forget to update the JS. Shit happens.

hennevogel: If you think this is likely and this is a good trade off, okay.

dmarcoux: I prefer to have this spec and not have to think about it by knowing it's covered.

saraycp: > I prefer to have this spec and not have to think about it by knowing it's covered.

I think the same. Maybe in some other cases it's too much (when a simple JS to add cool effect) but in this case, we are switching on/off a very important feature.

PR #8196, opened 2019-08-28
Introduces client side javascript error tracking

hennevogel:
As a developer you might be interested in exceptions that occur in the
javascript that is running in the users browser. Separate this with another
errbit configuration key because this is getting exposed publicly.

### DIFF
--- a/src/api/app/assets/javascripts/webui2/airbrake-js.js.erb
+++ b/src/api/app/assets/javascripts/webui2/airbrake-js.js.erb
@@ -1,0 +1,25 @@
+//= require airbrake-js-client
+var errbit_id = "<%= CONFIG['errbit_javascript_api_key'] %>";
+var errbit_host = "<%= CONFIG['errbit_javascript_host'] %>";
+
+if (errbit_id) {
+  var airbrake = new airbrakeJs.Client({
+    projectId: 1,
+    projectKey: errbit_id,
+    host: errbit_host,
+    environment: 'production',
+  });
+
+  window.onerror = function (message, file, line, col, error) {
+    var promise = airbrake.notify(error);
+    promise.then(function(notice) {
+      if (notice.id) {
+        // console.log('Notified errbit. Notice:', notice.id);
+        console.log('Notified errbit.');
+      } else {
+        // console.log('Notifying errbit failed. Reason:', notice.error);
+        console.log('Notifying errbit failed!');
+      }
+    });
+  };
+};
--- a/ReleaseNotes-2.11
+++ b/ReleaseNotes-2.11
@@ -27,7 +27,8 @@ Features
 ========
 
 Generic:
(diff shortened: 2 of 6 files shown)


hennevogel:
there is debug code... 🙈🙉🙊

src/api/app/assets/javascripts/webui2/airbrake-js.js.erb
```diff
@@ -0,0 +1,25 @@
...
+
+  window.onerror = function (message, file, line, col, error) {
+    var promise = airbrake.notify(error);
+    promise.then(function(notice) {
+      if (notice.id) {
+        // console.log('Notified errbit. Notice:', notice.id);
+        console.log('Notified errbit.');
+      } else {
+        // console.log('Notifying errbit failed. Reason:', notice.error);
+        console.log('Notifying errbit failed!');
```
danidoni: Shouldn't we let the exception bubble up if the errbit notification failed? We're swallowing the error trace otherwise right?

hennevogel: No this after the error and it's processing. See the [window.onerror](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror) docu.

Screenshot from 2019-08-29 12-50-37

hennevogel: And users can't really fix the error if something goes wrong when sending it to errbit.

danidoni: > No this after the error and it's processing. See the [window.onerror](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror) docu.
> 
> Screenshot from 2019-08-29 12-50-37

Alright! :+1:

danidoni: > And users can't really fix the error if something goes wrong when sending it to errbit.

You are right, my concerns were more on the line of the devs trying to debug stuff that does not work and this swallowing the error traces, but as you posted before, everything is ok, so :coffee: :dog: :fire:

src/api/app/assets/javascripts/webui2/airbrake-js.js.erb
```diff
@@ -0,0 +1,25 @@
...
+    projectKey: errbit_id,
+    host: errbit_host,
+    environment: 'production',
+  });
+
+  window.onerror = function (message, file, line, col, error) {
+    var promise = airbrake.notify(error);
+    promise.then(function(notice) {
+      if (notice.id) {
+        // console.log('Notified errbit. Notice:', notice.id);
```
dmarcoux: I believe we should remove those `console.log` calls if they are commented.

PR #8220, opened 2019-08-30
Remove Bento helpers, 1st part

eduardoj:
We want to get rid of the helpers used only in the old UI.

For the sake of ease this pull request only takes in account helpers from `a` till `m` located under `src/api/app/helpers/webui/`.

Some refactoring was also done:  ~~move methods only used in its class to private, or~~ move a helper to its corresponding helper class.


eduardoj:
[Remove bento helpers](https://trello.com/c/aaNWKJZf/1240-remove-bento-helpers)

eduardoj:
Marking as "DO NOT MERGE" until fix spider tests are fixed.

vpereira:
waiting #8220

src/api/app/helpers/webui/maintenance_incident_helper.rb
```diff
@@ -100,6 +56,43 @@ def release_targets_cell(incident)
 
   private
```
coolo: AFAICS rails does not support private functions in helpers - the nature of mixing modules into controllers plainly makes them private functions to every controller.

So IMO you just pollute git blame without much of a gain here.

eduardoj: > AFAICS rails does not support private functions in helpers - the nature of mixing modules into controllers plainly makes them private functions to every controller.

Not what I wanted to achieve. I will remove that part. Thanks for the hint.

eduardoj:
Spider tests are fixed! :tada: Removing "DO NOT MERGE" label.

src/api/app/helpers/webui/buildresult_helper.rb
```diff
@@ -1,34 +1,4 @@
...
-
-      if enable_help && status['code']
-        concat(' ')
-        concat(sprite_tag('help', title: Buildresult.status_description(status['code'])))
-      end
-    end
-  end
-
   # NOTE: There is a JavaScript version of this method in project_monitor.js
   def webui2_arch_repo_table_cell(repo, arch, package_name, status = nil, enable_help = true)
```
hennevogel: What about renaming this one? :-)

eduardoj: Ops. I'll do it.

eduardoj: Done.

src/api/app/helpers/webui/monitor_helper.rb
```diff
@@ -4,4 +4,15 @@ def self.print_statistics_array(array)
     array ||= []
     '[' + array.map { |time, value| "[#{time * 1000}, #{value}]" }.join(',') + ']'
   end
+
+  def icon_for_daemon(state)
```
hennevogel: The commit message says you are moving `kiwi_image_breadcrumb` not this.

eduardoj: Thanks! Fixed.

src/api/spec/helpers/webui/buildresult_helper_spec.rb
```diff
@@ -1,100 +0,0 @@
-require 'rails_helper'
-
-RSpec.describe Webui::BuildresultHelper do
-  include Webui::WebuiHelper
-  let(:project) { create(:project) }
-  let(:package) { create(:package, project: project) }
-  let(:arch) { 'i586' }
-  let(:repo) { create(:repository, project: project, architectures: [arch]) }
-
-  describe '#arch_repo_table_cell' do
```
dmarcoux: This helper is still present. We should update these specs instead of removing them.

eduardoj: I will update the specs.

eduardoj: Updated. Please re-review.

dmarcoux:
@eduardoj is not available today and tomorrow, so I will take over.

dmarcoux:

dmarcoux:
@hennevogel Again please? :smile:

dmarcoux:
I just realized that the `group_with_icon` helper is needed. I dropped the commit removing it. The helper is used in the partial which we need to migrate to Bootstrap. See the issue @eduardoj created: https://github.com/openSUSE/open-build-service/issues/8223.

PR #8224, opened 2019-08-30
 Run spider test in the new UI

DavidKang:
As we plan to remove bento, we need to run the tests in the new UI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -191,7 +191,9 @@ jobs:
       - run: *bootstrap_old_test_suite
       - run:
           name: Run spider
-          command: cd src/api; bundle exec rake test:spider
+          command: |
+            cd src/api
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -191,7 +191,12 @@ jobs:
       - run: *bootstrap_old_test_suite
       - run:
           name: Run spider
-          command: cd src/api; bundle exec rake test:spider
+          command: |
+            # TODO: Remove this when switch_to_webui2 is removed
+            export BOOTSTRAP=1
```
hennevogel: Can we start doing this inside the specs?

coolo: nitpick: spider test is a minitest :)

DavidKang:
@coolo and @hennevogel, could you review it again? please.

PR #8234, opened 2019-09-02
[dist] obs-bundled-gems.spec: fix build on ppc by dropping

bugfinder:
the useless "-march=native" compile flag unsupported on power when building rubygem-sassc


hennevogel:
They desperately need help getting their [build](https://github.com/sass/sassc-ruby/blob/master/ext/extconf.rb) of libsass in order or allow to use the system libs...

bugfinder:
at least in sassc-2.1.0 the libsass sources are byte-identical to the sources of libsass-3.6.1,
the only differences are removed files but no changed code at all. So patching in a "use-system-libraries" option is IMHO the way to go there

PR #8238, opened 2019-09-02
Remove Bento helpers, 3rd part

eduardoj:
Remove helpers only used in the old UI, from `s` till `w`, located under `src/api/app/helpers/webui/`.


eduardoj:
[Remove bento helpers](https://trello.com/c/aaNWKJZf/1240-remove-bento-helpers)

src/api/spec/helpers/webui/webui_helper_spec.rb
```diff
@@ -297,14 +164,6 @@
...
 
-  describe '#repo_status_icon' do
-    skip('Please add some tests')
-  end
-
-  describe '#tab' do
-    skip('Please add some tests')
-  end
-
   describe '#render_dialog' do
```
krauselukas: nitpick: I think the ```#render_dialog``` can be removed as well, at least I couldn't find something in the webui_helper that seemed to be related

eduardoj: Added. Thanks!

src/api/spec/helpers/webui/webui_helper_spec.rb
```diff
@@ -63,27 +63,6 @@
     end
   end
 
-  describe '#repo_status_icon' do
```
dmarcoux: This helper is still present, so we should update the specs instead of removing them.

eduardoj: Bento helper `repo_status_icon` and its tests were removed in 1a0eabefd1aab039f61e43415eea1a578e904393. Newer helper `webui2_repo_status_icon` was renamed to `repo_status_icon` in ef35b3ab2b06d2c53bf8a4ec7495956372225f13. That is why "the helper is still present". Do we really need a test for actual renamed to `repo_status_icon` helper, that has this code?:
```
  def repo_status_icon(status)
    REPO_STATUS_ICONS[status] || 'eye'
  end
```

dmarcoux: It's true that it's pretty basic code... I guess it's fine.

PR #8239, opened 2019-09-02
Add feature spec for Factory dashboard

vpereira:
The idea isn't to test the functionality per se but to make sure that removing bento won't break it. Therefore I'm testing for bento assets and bento layout elements

### DIFF
--- a/src/api/spec/bootstrap/features/webui/factory_dashboard_spec.rb
+++ b/src/api/spec/bootstrap/features/webui/factory_dashboard_spec.rb
@@ -1,0 +1,129 @@
+require 'browser_helper'
+
+RSpec.feature 'FactoryDashboard', type: :feature, js: true, vcr: true do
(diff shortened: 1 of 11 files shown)


saraycp:
I don't have a strong opinion, but maybe we could also check the following.

In a view like https://build.opensuse.org/project/staging_projects/openSUSE:Leap:15.2
- add a request and check it's in the table and also shown inside Info in Backlog section.
- when we have many openqa errors they are shown correctly as collapsed list
- check the 'A' is shown on the left side.

In a view like https://build.opensuse.org/project/staging_projects/openSUSE:Leap:15.2/A
- when errors, they are shown in the correct cell
- the icons are there and correspond to the state (when errors or not)

In a view like https://build.opensuse.org/project/dashboard/openSUSE:Leap:15.2:
- The staging project list is not empty

...

PR #8255, opened 2019-09-03
Replace bento view by bootstrap views for comments

DavidKang:
The coexistence between bento and bootstrap is no longer necessary. We
started integrating controllers and views


dmarcoux:
I wouldn't fix the CodeClimate issues. I approved them.

PR #8266, opened 2019-09-04
Replace Bento by Bootstrap for search

eduardoj:


eduardoj:
[Webui::SearchController](https://trello.com/c/7esWWLHi/1256-webuisearchcontroller)

src/api/app/views/webui/search/_results.html.haml
```diff
@@ -1,25 +1,26 @@
...
-    .search_result
-      %h6.data-title
-        = sprite_tag(rtype, class: rtype, title: rtype.humanize)
-        = project_or_package_link project: project, package: package, short: false, trim_to: nil
-        = content_tag :span, result.sphinx_attributes, style: 'display:none'
+.mt-4#search-results
+  %h4 Results:
+  - results.each do |result|
+    :ruby
+      if result.is_a? Project
```
danidoni: Shouldn't we use `duck typing` instead of static class types for branching? Something like ```ruby
if result.respond_to?(:project_name)
```

Comment (eduardoj) [reply src/api/app/views/webui/search/_results.html.haml]: I agree. And this code can be changed after this pull request. But the goal of this pull request is only replacing Bento code with the already existing Bootstrap code.

Definitely a review in this direction should be done for the whole application. I counted 124 ocurrences of  `is_a?` in the actual code inside `src/api/app/`:

```bash
> grep -r is_a? src/api/app/ | wc -l
124
>
```

danidoni: Ok, I see

PR #8267, opened 2019-09-04
 Replace bento views by bootstrap views for groups

DavidKang:


dmarcoux:
Some specs are failing due to a wrong path.

DavidKang:
@dmarcoux, could you review it again? please :smile_cat:

dmarcoux:
I have 2 questions:

- The shared views under `src/api/app/views/webui2/shared/user_or_groups_roles` are going to be moved when we move the users controller?
- It's not possible to delete a group? Try it here: https://obs-reviewlab.opensuse.org/davidkang-bootstrapmerge-groups/groups

Everything else looks good.

DavidKang:
> The shared views under src/api/app/views/webui2/shared/user_or_groups_roles are going to be moved when we move the users controller?

These views are already moved to `/webui/shared`. We have a card for remove the `webui2/webui/shared` directory when we finish to move all the controllers and views.

> It's not possible to delete a group? Try it here: https://obs-reviewlab.opensuse.org/davidkang-bootstrapmerge-groups/groups

I looked in the bento UI, and at least via webui we can't delete or change the name of the group.

PR #8269, opened 2019-09-04
Minimal bento elements for Staging

saraycp:
Since we have migrated almost all our views to Bootstrap and they are well tested, we started to remove most of the bento related elements.

However, the Staging dashboard is not migrated to Bootstrap and we decided to preserve a minimal amount of bento elements to keep the dashboard working. They are moved to `obs_factory` namespace in views (including the layout), assets, controllers, etc.

Most of the old CSS, Javascript and images/icon files has been removed, for that reason, `icons._sprite.png` was updated and now it only contains the remaining icons and the sponsor ones. To do so, I had to modify the task (sprite.rake) used to generate this image based on the icon files. It looked like this before:

before

and now it looks like:

icons_sprite

To test this PR, please, follow [this instructions](https://github.com/openSUSE/open-build-service/wiki/Factory-Dashboard) to create content locally or try out on the review-app. The views to test are:

- [main dashboard page](https://obs-reviewlab.opensuse.org/saraycp-minimal_bento_for_staging//project/dashboard/openSUSE:Factory).
- [staging projects page](https://obs-reviewlab.opensuse.org/saraycp-minimal_bento_for_staging/project/staging_projects/openSUSE:Factory).
- [one staging project page](https://obs-reviewlab.opensuse.org/saraycp-minimal_bento_for_staging/project/staging_projects/openSUSE:Factory/A).


hennevogel:
In the last commit you also remove the sponsor images again.

saraycp:
@hennevogel

> One general question: Why do you move `application.css.erb` to `application.css.erb` if there is no erb in it?

Good catch, updated!

> In the last commit you also remove the sponsor images again.

We had sponsor images in two different places: inside `assets/icons/`directory and inside `assets/images/sponsor/` directory. AFAIK, we are only using those in `assets/icons/`(now moved to `assets/images/icons`) that are used to generate `icons_sprite.png`. So I removed all the logos inside `assets/images/sponsor/`despite it contained some logos that weren't in `assets/icons/`, but I checked they weren't used in production either.

saraycp:
> ~You miss a change in `config/initializer/assets.rb`.~ You changed the extension of the `application.css.erb` to `application.scss`, so the assets will not be precompiled. If I'm not wrong the correct extension should be `.css.scss`

You are right, thanks. Updated!

DavidKang:
> You are wrong, manifest files don't have other preprocessor extensions.

ok, sorry @saraycp

saraycp:
Ok, @hennevogel, moved back `application.css.scss` to `application.scss` and also rebased.
Please :eyes:

hennevogel:
Sorry now you removed the wrong extension. This should be `application.css`

saraycp:
oh sorry @hennevogel :sweat_smile: Updated!

PR #8275, opened 2019-09-05
Refactor SessionController

vpereira:
After the migration to boostrap, there was some room for improvement like:

- Make the RabbitMQ notifications async
- Extract logic of the SessionController#create to a service
- Adapt and extend test suite


src/api/app/controllers/webui/session_controller.rb
```diff
@@ -8,35 +8,30 @@ def new
...
+    SessionControllerNotificationJob.perform_later('login,access_point=webui value=1')
     redirect_on_login
   end
 
   def destroy
-    Rails.logger.info "Logging out: #{session[:login]}"
-
     reset_session
-    RabbitmqBus.send_to_bus('metrics', 'logout,access_point=webui value=1')
+    SessionControllerNotificationJob.perform_later('logout,access_point=webui value=1')
```
hennevogel: This doesn't make much sense, pushing to the bus is not costly. And the resulting data only makes sense if it's connected to the time where it happened, not the time where DJ has gone through it's queue.

src/api/app/services/session_controller_service/session_creator.rb
```diff
@@ -0,0 +1,23 @@
+module SessionControllerService
```
hennevogel: This is a lot of glue code for `User.find_with_credentials`. Are you planing to move code in from that end?

src/api/app/controllers/webui/session_controller.rb
vpereira: it's updating a counter.. just it, right? I'm not sure why the time should matter.. even if it's not "costly" the session creation shouldn't be affected by that. Normal applications always off-load main thread with logging actions.

src/api/app/services/session_controller_service/session_creator.rb
vpereira: I didn't clean my whole stash. Now it's done :)

src/api/app/controllers/webui/session_controller.rb
hennevogel: > I'm not sure why the time should matter..

Because this is sent to a timeseries database. You correlate it with other data from this DB, logs, etc. Timestamps matter very much.

> IMO it's a good pattern to off-load pure logging actions from main thread..

I don't agree. You wouldn't do this with `Rails.logger` would you?

src/api/app/services/session_controller_service/session_creator.rb
hennevogel: > I didn't clean my whole stash. 

Not sure what you mean by that, can you elaborate?

> The idea was to clean up some logic from the SessionController#create

Sure thing, this is good. But that is only half of the logic right? The rest is behind `User.find_with_credentials`, my question was if you want to tackle this too?

src/api/app/controllers/webui/session_controller.rb
```diff
@@ -1,47 +1,48 @@
...
     redirect_on_logout
   end
 
   private
 
+  def session_creator
+    @session_creator = SessionControllerService::SessionCreator.new(params.slice(:username, :password))
+  end
+
+  def check_user_active
```
hennevogel: This sounds like an extension to `SessionControllerService.valid?`

src/api/app/controllers/webui/session_controller.rb
```diff
@@ -1,47 +1,48 @@
...
-      return
-    end
-
-    User.session = user
-    session[:login] = user.login
-    Rails.logger.debug "Authenticated as user '#{user.login}'"
-    RabbitmqBus.send_to_bus('metrics', 'login,access_point=webui value=1')
-
+    User.session = @session_creator.user
+    session[:login] = @session_creator.user.login
```
hennevogel: Hm, those two lines look like things that should happen in `SessionControllerService::SessionCreator` to me.

src/api/app/controllers/webui/session_controller.rb
vpereira: > Because this is sent to a timeseries database. You correlate it with other data from this DB, logs, etc. Timestamps matter very much.

We are sending it to a Remote server and we are aggregating the "+1", so I don't see how we are able to correlate it with something else.

> I don't agree. You wouldn't do this with Rails.logger would you?

if we are just counting events why not? `Thread.new { Rails.logger("+1") }

Would I make you happy if I pass the timestamp around?

src/api/app/controllers/webui/session_controller.rb
vpereira: can be, I will look on that

src/api/app/controllers/webui/session_controller.rb
vpereira: The `User.session` for sure, however I think don't think it belongs there. The `session` is a global from the controller  and I'm not sure how would be a good way to manipulate it without being too hacky.

src/api/app/controllers/webui/session_controller.rb
vpereira: not sure exactly what you mean, could you elaborate?

src/api/app/services/session_controller_service/session_creator.rb
vpereira: > Not sure what you mean by that, can you elaborate?
> 
> >

Ah I had some changes on my stash that I didn't push, specially the changes on the controller.. may bad.

vpereira: Your question was if would make sense to move the whole find_with_credentials* to a service? Maybe, should it happen in this PR? not necessarily.

vpereira: > This is a lot of glue code for `User.find_with_credentials`. Are you planing to move code in from that end?

if you look the new `create` method it is straight forward, no more nested conditions of `if xxx; redirect and return` which in my opinion can be better handled with `before_filters`

vpereira: just look the file https://github.com/openSUSE/open-build-service/blob/7cba5648875dbfda6ccd91b868d8e5695c17610b/src/api/app/controllers/webui/session_controller.rb#L14 the method is just straight forward, easy to understand without hide important information. Just beautiful :)

vpereira:
as discussed offline with @hennevogel I'm going to keep the rabbitmq code in the controller.

PR #8281, opened 2019-09-05
Replace Bento views by Bootstrap views for patchinfo

dmarcoux:
Patchinfo in the review app: https://obs-reviewlab.opensuse.org/dmarcoux-bootstrap-patchinfo/patchinfo/show/home:Admin/patchinfo


src/api/app/views/webui/patchinfo/show.html.haml
```diff
@@ -1,86 +1,53 @@
...
+
+  .col-md-6
+    .card.mb-3
+      %h5.card-header Fixed bugs
+      .card-body
+        - @patchinfo.issues.each do |issue|
+          .media
+            %i.pt-1.mr-1.fas.fa-bug.text-danger
+            .media-body.mb-3
+              = patchinfo_issue_link(issue[1], issue[0], issue[2])
```
vpereira: would be fine in the near feature if we actually could work with a structure or class instead of  ```issue[n]``` which doesn't give a lot of information about what we are trying to achieve there.

Issue #8279, opened 2019-09-05
[production][public_key#key_dialog] undefined method `remove_dialog_tag' for #<#<Class:0x00007f926800ea40>:0x00007f92599ac6d8>
Did yo...


hennevogel:
public_key#key_dialog threw this 57 times
<pre>
undefined method `remove_dialog_tag&#39; for #&lt;#&lt;Class:0x00007f926800ea40&gt;:0x00007f92599ac6d8&gt;
Did you mean?  remove_flag
</pre>
[https://build.opensuse.org/projects/openSUSE:Maintenance:6922/public_key/key_dialog](https://build.opensuse.org/projects/openSUSE:Maintenance:6922/public_key/key_dialog)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5d6fca494d64e7000df42928)

coolo:
It's mostly spiders hitting this. I assume they got the URL from the bento UI?

dmarcoux:
> It's mostly spiders hitting this. I assume they got the URL from the bento UI?

It looks like this is what's happening. This `key_dialog` action is the old `GPG Key / SSL Certificate` modal. We used to have actions for rendering those modals in Bento, but we don't anymore. This issue will be fixed once we remove this route [here](https://github.com/openSUSE/open-build-service/blob/54d6c1c82d26972ec6a4f7f24baadf45ec32121f/src/api/config/routes.rb#L337-L340). It will be done soon as we are already tackling this in our current deliverable.

DavidKang:
handled in PR #8313

PR #8282, opened 2019-09-05
bring back remove_dialog_tag helper method

vpereira:
bring method back, since we still have some old views using it


dmarcoux:
This was closed for [this reason](https://github.com/openSUSE/open-build-service/issues/8279#issuecomment-528342687).

PR #8294, opened 2019-09-07
Remove Gitlab webhook event restriction during authentication.

matthewdva:
When validating Gitlab webhook triggers the authentication token
is only valid for "Push Hook" events.  Gitlab has multiple events
that could be valid.  Choosing to accept the Gitlab authentication
header does not need to be dependent on the webhook event.

This single line patch only checks for the http header X-GITLAB-TOKEN
and removes the dependency on the X-GITLAB-EVENT.

### DIFF
--- a/src/api/app/services/trigger_controller_service/token_extractor.rb
+++ b/src/api/app/services/trigger_controller_service/token_extractor.rb
@@ -10,7 +10,8 @@ module TriggerControllerService
     end
 
     def extract_auth_token
-      @auth_token = if @http_request.env['HTTP_X_GITLAB_EVENT'] == 'Push Hook'
+      events = ['Push Hook', 'Tag Push Hook', 'Merge Request Hook']
+      @auth_token = if events.any? { |event| event == @http_request.env['HTTP_X_GITLAB_EVENT'] }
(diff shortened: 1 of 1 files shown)


adrianschroeter:
it is okay to extend the hooks by a list, but we should avoid to run on any call, even when it can not trigger a source change.

matthewdva:
@adrianschroeter I refactored the update based on your suggestion.  I see some of the required tests are now failing.  I do not see where my change is causing the failed test.  Please advise if I need to make further adjustments.

vpereira:
> @adrianschroeter I refactored the update based on your suggestion. I see some of the required tests are now failing. I do not see where my change is causing the failed test. Please advise if I need to make further adjustments.

Hi @matthewdva!
 
The Hakiri found some security issue in your branch. Probably your gems are out of date. Rebasing with master should fix it

matthewdva:
> 
> 
> > @adrianschroeter I refactored the update based on your suggestion. I see some of the required tests are now failing. I do not see where my change is causing the failed test. Please advise if I need to make further adjustments.
> 
> Hi @matthewdva!
> 
> The Hakiri found some security issue in your branch. Probably your gems are out of date. Rebasing with master should fix it

@vpereira Thank you for the tip.  All the test pass again.

Issue #8303, opened 2019-09-09
Bento Icons for Images


hennevogel:
Some [images](https://build.opensuse.org/image_templates) used bento icons and are now broken.

PR #8307, opened 2019-09-10
Recover missing icon for image template view

saraycp:
Most of the icons showed in [image template view](https://build.opensuse.org/image_templates) are taken from the `_icon`file inside each package. However, the default icon is taken from the assets and it was removed in a previous PR, so we have recovered it.

The default icon is used directly, so no need to re-generate icons_sprite.png.

Fixes: #8303


hennevogel:
This fixes #8303 right?

saraycp:
> This fixes #8303 right?

:+1:  Added in the description.

dmarcoux:
> This fixes #8303 right?

Yes. It's in the PR's description, but not the commit message.

PR #8310, opened 2019-09-10
Move Webui::Users::AnnouncementsController to Bootstrap

vpereira:
Move the create action to bootstrap.

### DIFF
--- a/src/api/app/controllers/webui/users/announcements_controller.rb
+++ b/src/api/app/controllers/webui/users/announcements_controller.rb
@@ -9,6 +9,5 @@ class Webui::Users::AnnouncementsController < Webui::WebuiController
     else
       flash.now[:error] = "Couldn't find Announcement"
     end
-    switch_to_webui2
   end
 end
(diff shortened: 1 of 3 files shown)


vpereira:
[Webui::UserController](https://trello.com/c/zOHXNyMW/1261-webuiusercontroller)

src/api/app/controllers/webui/users/announcements_controller.rb
```diff
@@ -9,6 +9,5 @@ def create
     else
       flash.now[:error] = "Couldn't find Announcement"
     end
-    switch_to_webui2
```
dmarcoux: This should be kept until we get rid of it completely (which we will do later).

vpereira:
if you want to create a Announcement via osc:

```
osc -A  https://obs-reviewlab.opensuse.org/vpereira-move_announcements_controllers_to_bootstrap/announcements  -X POST -d '<announcement><title>foo</title><content>bar</content></announcement' /announcements
```

src/api/app/controllers/webui/users/announcements_controller.rb
vpereira: why if it's not necessary? It's just necessary if it needs the webui2 views in the view path...

dmarcoux:
I fixed the command in @vpereira's comment.

PR #8312, opened 2019-09-10
Merge all the possible appdata metadata files

hellcp:
To verify this feature
- Build a package containing *.metainfo.xml
- Check if the outputted appdata.xml in metadata folder contains the entry for the component

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -1813,7 +1813,7 @@ sub publish {
 	my $updateinfo = readxml("$r/$bin", $BSXML::updateinfo, 1) || {};
 	push @updateinfos, @{$updateinfo->{'update'} || []};
       }
-      if ($bin =~ /[-.]appdata.xml$/) {
+      if ($bin =~ /[-.](appdata|metainfo)\.xml$/) {
         # collect application data
         my $appdataxml = readstr("$r/$bin", 1) || '';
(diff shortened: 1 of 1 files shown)


mlschroe:
"*.metainfo.xml" is not supported in many other places in OBS. I'm not sure we should change the filename, it's a SUSE specific thing anyway.

mlschroe:
I.e. the brp script should export the metainfo.xml file from the rpm package as packagename-appdata.xml.

Conan-Kudo:
@mlschroe Why? `.appdata.xml` has been the legacy name for several years now, and `.metainfo.xml` has replaced it in the AppStream spec. Same goes for the install path (`/usr/share/appdata` vs `/usr/share/metainfo`).

mlschroe:
It's fine to use `.metainfo.xml` in the packages, but that doesn't mean that the intermediate file used to transport the information to the publisher has to have the same name. There are quite a few places in the OBS code that check for `appdata.xml`. E.g.

- the worker download code used when building products
- the publisher
- the reposerver calculating the binary information
- the scheduler channel code used for maintenance updates
- the scheduler patchinfo code used for maintenance updates

and so on.

adrianschroeter:
closing for now, feel free to reopen if you have a compatible approach.

PR #8325, opened 2019-09-11
Remove Bento only methods/code in request controller

eduardoj:
Now that we are only using the new web interface, we get rid of the Bento code and methods in the request controller.

Adapt the request controller tests.


eduardoj:
[Webui::RequestController](https://trello.com/c/u9p1Dun0/1255-webuirequestcontroller)

PR #8346, opened 2019-09-13
[api] fix editing of issue tracker entries

adrianschroeter:
eg. via

` # osc api -e /issue_trackers/$tracker
`
we need to backport in case 2.10 is also affected.

Also all other places in test suite where "put ...., params: $body_data"
is used needs to be checked if we broke the api there as well.
needs to become raw_put again...

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -97,7 +97,7 @@ POST /issue_tracker/<name>
 
   Create new issue tracker.
 
-XmlResult: issue_tracker
+XmlResult: status
 
 DELETE /issue_tracker/<name>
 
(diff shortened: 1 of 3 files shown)


hennevogel:
Maybe you can explain what the problem actually is that you are fixing?

adrianschroeter:
that it isn't working basically and all changes get ignored when use as described in the api. Just run the adapted test suite without the fix to see that (sending it as www encoded params was never intended in the api).

hennevogel:
So you do want to get rid of the (partial) json API?

hennevogel:
How about this?

hennevogel:
@vpereira please have another look

hennevogel:
@adrianschroeter this changes the API now, see the included chunk for `api.txt`. I think it makes sense, are you okay with that?

adrianschroeter:
@hennevogel seems you forgot to commit the chunk for api.txt?

If it is just about the drop of .json, I am fine with it (it seems to not usable anyway so far at least for writing).

hennevogel:
Indeed I have. pushed now

PR #8354, opened 2019-09-16
Remove actions for Bento modals in user_controller.rb

dmarcoux:
`render_dialog` is for Bento modals which we don't need anymore. Their routes and specs were also removed.


dmarcoux:
[Webui::UserController](https://trello.com/c/zOHXNyMW/1261-webuiusercontroller)

PR #8356, opened 2019-09-16
Remove unneeded cassettes

dmarcoux:
These cassettes should have been removed when merging the Bootstrap feature specs under `spec/features/`.


dmarcoux:
Somehow Hakiri is stuck, but since I'm removing cassettes, it doesn't matter.

PR #8358, opened 2019-09-17
Make the spider test crawl the sitemap (as good bots do)

coolo:
We no longer have a way to list all projects nor a way to list
all packages in projects without having javascript due to use
of ajax. This made the spider test a game of luck and certain
packages were only found if they were listed in requests or
rebuild graphs (as it happened in Eduardo's PR for some reason).

But to have reproducible result, it's important to crawl all
pages to find also those we don't normally test in rspec

### DIFF
--- a/src/api/test/functional/webui/spider_test.rb
+++ b/src/api/test/functional/webui/spider_test.rb
@@ -7,6 +7,28 @@ require 'nokogiri'
 ENV['BOOTSTRAP'] = '1'
 
 class Webui::SpiderTest < Webui::IntegrationTest
+  def ignore_link?(link)
+    return true if link =~ %r{/mini-profiler-resources}
+    # that link is just a top ref
+    return true if link.end_with?('/package/rdiff')
+    # admin can see even the hidden
+    return true if link.end_with?('/package/show/HiddenRemoteInstance')
+    return true if link =~ %r{/package/show/SourceprotectedProject}
+    # this is crashing (bug)
+    return true if link =~ %r{/package/show/UseRemoteInstance}
+    return true if link.end_with?('/project/show/HiddenRemoteInstance')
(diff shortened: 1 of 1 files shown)


hennevogel:
Not that any real spider would use our sitemap, but good idea 😊

src/api/test/functional/webui/spider_test.rb
```diff
@@ -7,6 +7,28 @@
 ENV['BOOTSTRAP'] = '1'
 
 class Webui::SpiderTest < Webui::IntegrationTest
+  def ignore_link?(link)
+    return 1 if link =~ %r{/mini-profiler-resources}
```
hennevogel: nitpick: `return true...`

src/api/test/functional/webui/spider_test.rb
```diff
@@ -7,6 +7,28 @@
...
+    # that link is just a top ref
+    return 1 if link.end_with?('/package/rdiff')
+    # admin can see even the hidden
+    return 1 if link.end_with?('/package/show/HiddenRemoteInstance')
+    return 1 if link =~ %r{/package/show/SourceprotectedProject}
+    # this is crashing (bug)
+    return 1 if link =~ %r{/package/show/UseRemoteInstance}
```
coolo: Note that this bug is real, but I didn't report it. You can see it in action if you follow any package name on https://build-test.opensuse.org/project/monitor/OBS:Server:Unstable?arch_x86_64=1&broken=1&defaults=0&repo_SLE_12_SP4=1

src/api/test/functional/webui/spider_test.rb
coolo: Yeah, that happens if you mix perl and ruby between day and night :)

coolo: fixed

PR #8359, opened 2019-09-17
 Consolidate duplicated methods from project and package controller 

krauselukas:
Due to the migration from bento to bootstrap, certain
methods turned unnecessary and can be removed.

### DIFF
--- a/src/api/app/mixins/webui/manage_relationships.rb
+++ b/src/api/app/mixins/webui/manage_relationships.rb
@@ -7,19 +7,19 @@ module Webui::ManageRelationships
     rescue NotFoundError,
            Relationship::AddRole::SaveError => e
       flash[:error] = e.to_s
-      return redirect_to(users_path)
+      return redirect_to(custom_users_path)
     end
     respond_to do |format|
       format.js { render json: {}, status: :ok }
       format.html do
         success_str = what == :person ? "user #{params[:userid]}" : "group #{params[:groupid]}"
         flash[:success] = "Added #{success_str} with role #{params[:role]}"
-        redirect_to users_path
+        redirect_to custom_users_path
       end
     end
   end
 
-  def users_path
+  def custom_users_path
     url_for(action: :users, project: @project, package: @package)
   end
 
@@ -55,7 +55,7 @@ module Webui::ManageRelationships
         else
           flash[:success] = "Removed group '#{params[:groupid]}'"
         end
-        redirect_to users_path
+        redirect_to custom_users_path
       end
     end
   end
--- a/src/api/app/mixins/webui/manage_relationships.rb
+++ b/src/api/app/mixins/webui/manage_relationships.rb
@@ -1,14 +1,4 @@
 module Webui::ManageRelationships
-  def redirect_after_save
-    if switch_to_webui2
-      redirect_to(users_path)
-    elsif what == :person
-      redirect_to(add_path(:add_person))
-    else
-      redirect_to(add_path(:add_group))
-    end
-  end
-
   def save_person_or_group(what)
     authorize main_object, :update?
     begin
@@ -17,7 +7,7 @@ module Webui::ManageRelationships
     rescue NotFoundError,
            Relationship::AddRole::SaveError => e
       flash[:error] = e.to_s
-      return redirect_after_save
+      return redirect_to(users_path)
     end
     respond_to do |format|
       format.js { render json: {}, status: :ok }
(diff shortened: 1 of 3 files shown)


coolo:
Better not:

https://github.com/openSUSE/open-build-service/search?q=add_path&unscoped_q=add_path

coolo:
at least your tests caught you

dmarcoux:
The mixin depends on these methods, but doesn't define them, even though they are only used in the context of this mixin. The only difference between these methods is that the ones in `package_controller.rb` have `package: @package`. Checking if `@package` is not `nil` and then adding it can be done in the mixin. It would remove the duplication.

dmarcoux:
One more thing, I just realized that `add_path` is only used in the Bento UI. So this can be removed, don't bother with refactoring this.

coolo:
But the bootstrap tests fail after he removed it?

dmarcoux:
I just updated my comment above with the cause of the failures. I'm still trying to see why it's redirecting.... it's weird.

coolo:
High technology invented in 2012 :)

krauselukas:
Seems like I opened the box of the pandora :D

dmarcoux:
> High technology invented in 2012 :)

I'm trying to find when path/route helpers were introduced in Rails... Probably after 2012.. lol

coolo:
Well, @dmarcoux is right about add_path though :)

And you can still merge users_path into the mixin

coolo:
Important read here: https://en.wikipedia.org/wiki/Duck_typing

dmarcoux:
> Important read here: https://en.wikipedia.org/wiki/Duck_typing

Quack :duck:

dmarcoux:
Oh I see a fervent adapt of [rubber duck debugging](https://en.wikipedia.org/wiki/Rubber_duck_debugging)!

krauselukas:
@dmarcoux okay, I think its ready for another review ;)

src/api/app/mixins/webui/manage_relationships.rb
```diff
@@ -29,6 +23,10 @@ def save_person_or_group(what)
     end
   end
 
+  def users_path
+    url_for(action: :users, project: @package ? @project : @project.name, package: @package || nil)
```
coolo: what do you think ` || nil` does to the value? :)

krauselukas: why easy if there is also a comlicated solution :D thanks, long day ;)

src/api/app/mixins/webui/manage_relationships.rb
```diff
@@ -1,12 +1,6 @@
 module Webui::ManageRelationships
-  def redirect_after_save(what)
-    if switch_to_webui2
-      redirect_to(users_path)
-    elsif what == :person
-      redirect_to(add_path(:add_person))
-    else
-      redirect_to(add_path(:add_group))
-    end
+  def redirect_after_save
```
coolo: nitpick: that function is only used once and a one liner

krauselukas: done

src/api/app/mixins/webui/manage_relationships.rb
```diff
@@ -29,6 +19,10 @@ def save_person_or_group(what)
     end
   end
 
+  def users_path
```
dmarcoux: Please rename this to something else to not overshadow the path helper method with the same name.

src/api/app/mixins/webui/manage_relationships.rb
```diff
@@ -29,6 +19,10 @@ def save_person_or_group(what)
...
+    url_for(action: :users, project: @package ? @project : @project.name, package: @package)
```
dmarcoux: Passing `@project` or `@project.name` generates the same URL, so this ternary operator isn't needed.

src/api/app/mixins/webui/manage_relationships.rb
krauselukas: done

src/api/app/mixins/webui/manage_relationships.rb
krauselukas: done

krauselukas: Yeah :)

krauselukas: Hmm weired doesn't show here that it is outdated

krauselukas: Upps something got lost sorry

dmarcoux:
Good job @krauselukas!

krauselukas:
@dmarcoux merci :)

PR #8372, opened 2019-09-18
Remove switch_to_webui2 method

DavidKang:
Since we are done with the integration of the new UI, this method is not
needed anymore.

### DIFF
--- a/src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
+++ b/src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
@@ -5,9 +5,6 @@ module Webui
         before_action :require_login
         before_action :set_azure_configuration
 
-        # GET /cloud/azure/configuration
-        def show; end
-
         # PATCH/PUT /cloud/azure/configuration
         def update
           if @azure_configuration.update(permitted_params)
--- a/src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
+++ b/src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
@@ -6,9 +6,7 @@ module Webui
         before_action :set_azure_configuration
 
         # GET /cloud/azure/configuration
-        def show
-          switch_to_webui2
-        end
+        def show; end
 
         # PATCH/PUT /cloud/azure/configuration
         def update
--- a/src/api/app/controllers/webui/cloud/ec2/configurations_controller.rb
+++ b/src/api/app/controllers/webui/cloud/ec2/configurations_controller.rb
@@ -5,7 +5,6 @@ module Webui
         before_action :require_login
 
         def show
-          switch_to_webui2
           @ec2_configuration = User.session!.ec2_configuration || User.session!.create_ec2_configuration
           @aws_account_id = CONFIG['aws_account_id']
         end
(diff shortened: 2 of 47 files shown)


src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
```diff
@@ -6,9 +6,7 @@ class ConfigurationsController < WebuiController
         before_action :set_azure_configuration
 
         # GET /cloud/azure/configuration
-        def show
-          switch_to_webui2
-        end
+        def show; end
```
dmarcoux: An empty action isn't needed because of [Rails' default](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-by-default-convention-over-configuration-in-action)

There are a few cases of this in the PR. I didn't comment on every single case to avoid spamming the same comment over and over.

src/api/app/controllers/webui/kiwi/images_controller.rb
```diff
@@ -65,7 +65,7 @@ def edit
         @is_edit_software_action = params[:section] == 'software'
 
         respond_to do |format|
-          format.html { switch_to_webui2 }
+          format.html
```
coolo: I'm curious: what's the default respond? Is this creating 406 then?

coolo:
Note you're conflicting with htps://github.com/openSUSE/open-build-service/pull/8359 - so possibly you guys want to merge

dmarcoux:
I suggest rebasing once that PR from @krauselukas is merged.

src/api/app/controllers/webui/kiwi/images_controller.rb
DavidKang: The default format is html. So I removed it

src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
DavidKang: @dmarcoux, I removed some of them. The others, rubocop complains, because they are called in a `before_action` and it's need to be explicitly defined.

dmarcoux: What was the RuboCop offense? I'm curious.

DavidKang: https://rails.rubystyle.guide/#lexically-scoped-action-filter

src/api/app/controllers/webui/cloud/ec2/configurations_controller.rb
```diff
@@ -5,7 +5,6 @@ class ConfigurationsController < WebuiController
         before_action :require_login
 
         def show
-          switch_to_webui2
           @ec2_configuration = User.session!.ec2_configuration || User.session!.create_ec2_configuration
           @aws_account_id = CONFIG['aws_account_id']
```
hennevogel: Those two lines can go right?

src/api/app/controllers/webui/cloud/upload_jobs_controller.rb
```diff
@@ -8,14 +8,12 @@ class UploadJobsController < WebuiController
       before_action :set_upload_job, only: :destroy
 
       def index
-        switch_to_webui2
         @upload_jobs = ::Cloud::Backend::UploadJob.all(User.session!)
```
hennevogel: Same here...

src/api/app/controllers/webui/cloud/ec2/configurations_controller.rb
dmarcoux: No, they are still executed. They would need to be removed if it would have been `return if switch_to_webui2`.

Issue #8259, opened 2019-09-03
Incorrect appdata icons archive being generated by the publisher


doccaz:
OBS is generating the icons archive for appdata in the repodata directory with the name "app-icons.tar.gz" (more specifically, bs_publisher does that).

The expected behavior is that the bs_publisher component should deal with each case accordingly, e.g. generate app-icons.tar.gz for SLE12 and lower, and appdata-icons.tar.gz for SLE15 and newer. This is necessary because since SLE15/Leap 15, the libzypp appdata plugin checks for "appdata-icons.tar.gz":
https://github.com/DimStar77/openSUSE-appstream/blob/master/AsHelper.py

If it cannot find the proper icons archive, ALL appdata for that repository is ignored, and thus GNOME Software won't show the packages as available.

I currently have a customer that is using a private OBS instance (version 2.9.5-51.4) to generate packages automatically for GNOME Software (with a few thousand clients running on SLED 12 SP4). They cannot migrate these to SLE 15 due to this problem.

The package brp-extract-appdata was broken for SLE15+ as well, I proposed a patch already: https://github.com/openSUSE/brp-extract-appdata/issues/2

doccaz:
PR closed, thanks!

PR #8379, opened 2019-09-18
creating extra appdata-icons.xml.gz for SLE15+ (issue #8259)

doccaz:
Here is my proposed patch to work around the app-icons vs appdata-icons issue with SLE15+. The effect is minimal (only another file published with the same contents as app-icons.xml.gz). These files are usually small, and will allow for the repositories to be used with GNOME Software on both SLE12/Leap 42.3 and Leap 15/SLE15+.

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -647,7 +647,7 @@ sub createrepo_rpmmd {
   }
 
   # add appdata
-  unlink("$extrep/repodata/$_") for grep {/appdata\.xml/ || /app-icons/} @oldrepodata;
+  unlink("$extrep/repodata/$_") for grep {/appdata\.xml/ || /app-icons/ || /appdata-icons/} @oldrepodata;
   if (%{$data->{'appdatas'} || {}}) {
     create_appdata_files("$extrep/repodata", $data->{'appdatas'});
     if (-e "$extrep/repodata/appdata.xml") {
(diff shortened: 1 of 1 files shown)


DavidKang:
@doccaz, thanks for your contribution :bouquet:

doccaz:
Just found out that the extra *-appdata-icons.tar.gz file I created is NOT being deleted by the publisher when it updates the repository. So I now have a couple extra appdata-icons files in the directory... Trying to find where I missed it in the code.

DavidKang:
@mlschroe or @M0ses, could you have a look to this PR? please :smile_cat:

src/backend/bs_publish
```diff
@@ -658,7 +658,11 @@ sub createrepo_rpmmd {
     if (-e "$extrep/repodata/app-icons.tar") {
       print "    adding app-icons.tar to repodata\n";
       qsystem($modifyrepo_bin, "$extrep/repodata/app-icons.tar", "$extrep/repodata", @legacyargs) && die("    modifyrepo failed: $?\n");
-      unlink("$extrep/repodata/app-icons.tar");
+      qsystem("mv $extrep/repodata/app-icons.tar $extrep/repodata/appdata-icons.tar") && die("    copy appdata-icons failed: $?\n");
```
mlschroe: You shouldn't call qsystem() with "mv xxx yyy" as one string, instead you should use 'mv', 'xxx', 'yyy'.
But in this specific case you shouldn't call qsystem() at all; just use perl's rename() function.

doccaz: Done. Sorry for the (huge) delay. Too much happening at the same time.

adrianschroeter:
can you squash the two commits into one? Will merge right afterwards :)

adrianschroeter:
I have manualy cherry-picked now to avoid the additional merge commit.

thanks!

PR #8386, opened 2019-09-19
Remove unneeded project#packages_simple action

dmarcoux:
The view wasn't migrated to Bootstrap and it's not accessible without knowing the URL, so this can be removed.

The Bento version of the view was removed in 0991846.


coolo:
this was for spiders I assume?

coolo:
https://github.com/openSUSE/open-build-service/commit/93b980323f5e0af33d5f0f94a038e49668324431 - when noscript tag was still a thing :)

Issue #8314, opened 2019-09-11
Remove GetFlags mixin


dmarcoux:
This mixin is used in the Bento UI only. It cannot be simply removed since some unit tests also depend on it. The unit tests need to be adapted.

coolo:
https://www.youtube.com/watch?v=3_Af1qeouMc

krauselukas:
Fixed by https://github.com/openSUSE/open-build-service/pull/8387

PR #8387, opened 2019-09-19
Remove GetFlags mixin and deprecated methods from flag model

krauselukas:
The GetFlags mixin was only used by the bento ui and
can be removed.

Related to https://github.com/openSUSE/open-build-service/issues/8314


dmarcoux:
There are also a few methods the `Flag` model which can be removed since they are only used in Bento and in this `GetFlags` mixin. They are marked with the comment `# TODO: used by bento. Remove when dropping old UI.`

krauselukas:
> There are also a few methods the `Flag` model which can be removed since they are only used in Bento and in this `GetFlags` mixin. They are marked with the comment `# TODO: used by bento. Remove when dropping old UI.`

@dmarcoux yeah I saw them already, will take care of them as well :)

src/api/.rubocop.yml
```diff
@@ -36,6 +36,13 @@ Style/MultilineWhenThen:
 Style/UnneededInterpolation:
   Enabled: false
 
+# Disable this cop for the minitest, because it will be moved to rspec anyway and
+# a couple of outdated bento ui methods need to be moved to the minitest's
+# in order to get removed from the main's project codebase
+Metrics/ClassLength:
+  Exclude:
+    - 'test/unit/project_test.rb'
```
coolo: rather disable in the file - as done in 4 other tests

src/api/test/unit/project_test.rb
```diff
@@ -18,18 +18,47 @@ def test_maintained_project_names
...
+
+    the_flags
+  end
+
   # TODO: bento_only
   def _project_flags(project, flag)
     architectures = Architecture.where(id: project.repository_architectures.select(:architecture_id)).reorder('name').distinct
     repository_names = project.repositories.pluck(:name)
-    project.get_flags(flag, repository_names, architectures)
+    get_flags(project, flag, repository_names, architectures)
```
coolo: get rid of the complete test instead. There is no point in testing a function that only exists in the test itself.

src/api/test/unit/project_test.rb
```diff
@@ -18,18 +18,47 @@ def test_maintained_project_names
...
   end
 
   # TODO: bento_only
   def _package_flags(package, flag)
     architectures = Architecture.where(id: package.project.repository_architectures.select(:architecture_id)).reorder('name').distinct
     repository_names = package.project.repositories.pluck(:name)
-    package.get_flags(flag, repository_names, architectures)
+    get_flags(package, flag, repository_names, architectures)
```
coolo: get rid of the complete test instead

src/api/test/unit/project_test.rb
```diff
@@ -122,7 +151,7 @@ def test_flags_inheritance
     package2.store
 
     # the all 'disable' build flag shall overwrite the project setting
-    package2.get_flags('build', ['SLE_11_SP4'], Architecture.all).each do |repo|
+    get_flags(package2, 'build', ['SLE_11_SP4'], Architecture.all).each do |repo|
```
coolo: same here. Drop all testing of the local function get_flags.

We have a rspec for Flag::SpecifiedFlags and that's what is used.

src/api/test/unit/project_test.rb
krauselukas: @coolo I thought the specified_flags rspec test is not covering all the scenarios, but yeah I looked again, and it seems its covering the inheritance of the flags (even though a bit simpler). So I will remove all this stuff.

src/api/app/models/flag.rb
```diff
@@ -38,16 +38,12 @@ def arch
...
-    repo_flag = main_object.flags.find_by('flag = ? AND repo = ? AND architecture_id IS NULL', flag, repo)
-    arch_flag = main_object.flags.find_by('flag = ? AND repo IS NULL AND architecture_id = ?', flag, architecture_id)
-    same_flag = main_object.flags.find_by('flag = ? AND repo = ? AND architecture_id = ?', flag, repo, architecture_id)
-    if main_object.is_a?(Package)
-      same_flag ||= main_object.project.flags.find_by('flag = ? AND repo = ? AND architecture_id = ?', flag, repo, architecture_id)
-      repo_flag ||= main_object.project.flags.find_by('flag = ? AND repo = ? AND architecture_id IS NULL', flag, repo)
-      arch_flag ||= main_object.project.flags.find_by('flag = ? AND repo IS NULL AND architecture_id = ?', flag, architecture_id)
-      all_flag ||= main_object.project.flags.find_by('flag = ? AND repo IS NULL AND architecture_id IS NULL', flag)
-    end
+    pkg_or_prj_flags = main_object.is_a?(Package) ? main_object.project.flags : main_object.flags
```
coolo: careful. this is not the equivalent code. repo_flag for packag is package.flag || project.flag - you set it to package.flag

coolo: but the good news: this function is not used anymore - if you cleaned up the unit tests

krauselukas: Was still used by Specified::Flags rspec test, but I adapted the test, so the default_status method is gone now :)

src/api/spec/models/flag/specified_flags_spec.rb
```diff
@@ -16,8 +16,8 @@
     end
 
     it 'has default status on effective flag' do
-      expect(subject.effective_flag(nil, nil).status).to eq(Flag.default_status(flag_type))
-      expect(subject.effective_flag(repository_name, architecture_name).status).to eq(Flag.default_status(flag_type))
+      expect(subject.effective_flag(nil, nil).status).to eq(subject.default_flag(nil, nil).status)
```
coolo: Sorry Lukas, I don't get it. `Flag.default_status(flag_type)` is just a one liner calling `FlagHelper.default_for(flag_name)` . The function you replace it with is a much more complex variant to call the same helper

krauselukas: Puhh right, this Flag stuff confuses me, I changed it to use the helper directly, thanks for the hint

coolo: https://www.youtube.com/watch?v=LeyofQK6tRw

PR #8388, opened 2019-09-19
Remove rebuild_time from project controller filters

dmarcoux:
Hitting this action redirects to the show action in `projects/rebuild_times_controller.rb`. The filters are therefore not needed in `project_controller.rb`


dmarcoux:
I approved the Code Climate issue with this message:

`We cannot do anything about filters being similar in two controllers.`

dmarcoux:
I had to rebase to resolve conflicts, so please @DavidKang, have another look.

coolo:
> I approved the Code Climate issue with this message:
> 
> `We cannot do anything about filters being similar in two controllers.`

PR #8406, opened 2019-09-20
Move UserController#change_password to UsersController#change_password

vpereira:
- We adapted the routes, now the change_password is accessible under an
user i.e /users/$USER/change_password
- Adapted the change password dialog to use the new route
- The method change_password requires login
- Added more specs to cover other scenarios

How to test it:

- Login as User
- Go to the $USER profile
- Change your password
- Logout
- Login with new password


danidoni:
>     * The method change_password requires login
Oops, you mean you could change the password without being logged in?

vpereira:
> Oops, you mean you can right now change the password without being logged in?

No, if you look the `change_password` method, it sets the user to `User.session!` which will just explode if it's _nobody_.

danidoni:
> > Oops, you mean you can right now change the password without being logged in?
> 
> No, if you look the `change_password` method, it sets the user to `User.session!` which will just explode if it's _nobody_.

Oh, good to know! Thanks!

PR #8412, opened 2019-09-20
Prevent showing "Add user as maintaner" checkbox of a delete request

### LINKED ISSUE #7297: 'Add X as maintainer' displayed for !submit requests
I first spotted it in add_role requets, where it's just funny (I didn't try what happens if you accept the add_role request *and* check the box).

But https://build.opensuse.org/request/show/686832 is a delete request for a project - and adding the requesting user as maintainer won't work. I will try now though :)

This is an issue in both UIs:

Bento:
checkbox-old

Bootstrap:
checkbox

### LINKED ISSUE #8403: [production][request#changerequest] undefined method `can_be_modified_by?' for nil:NilClass
request#changerequest threw this 0 times
<pre>
undefined method `can_be_modified_by?&#39; for nil:NilClass
</pre>
[https://build.opensuse.org/request/changerequest](https://build.opensuse.org/request/changerequest)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5d841bed5efab9000bf4823b)

eduardoj:
Fixes #7297 and #8403.

The issue happens, for example, when a user receives a delete request from the Admin bot to remove a project to free resources on our always crowded server. It doesn't make much sense to ask for making the requestor to be maintainer of a project that is going to be deleted in the same request.

To verify the changes:
- Go to https://obs-reviewlab.opensuse.org/eduardoj-fix_8403/request/show/2
- Log in with the user 'Iggy'.
- Verify there isn't any "Add user as maintaner" checkbox at the end of the page.

**Update:**
The pull request was modified to take in account the cases where the requests were not of the type `submit`.

### DIFF
--- a/src/api/app/views/webui/request/_decision_tab.html.haml
+++ b/src/api/app/views/webui/request/_decision_tab.html.haml
@@ -3,7 +3,7 @@
   %p
     = text_area_tag(:reason, nil, placeholder: 'Please add a comment', rows: 4, class: 'w-100 form-control')
   - if single_action_request && is_target_maintainer && state.in?(['new', 'review'])
-    - if state.in?(['new', 'review']) && !action[:creator_is_target_maintainer]
+    - if !action[:creator_is_target_maintainer] && action[:type] == :submit
       .custom-control.custom-checkbox
         = check_box_tag('add_submitter_as_maintainer_0', "#{action[:tprj]}_#_#{action[:tpkg]}", false, class: 'custom-control-input')
         %label.custom-control-label{ for: 'add_submitter_as_maintainer_0' }


vpereira:
> This checkbox doesn't make sense for add_role, set_bugowner, maintenance_incident and maintenance_request actions either. This is just useful for single action submit requests - and for nothing else.

agree, however it has nothing to do with the bug fixed in this PR #8403 could that be done in another PR @eduardoj @coolo ?

eduardoj:
@coolo:

> This checkbox doesn't make sense for add_role, set_bugowner, maintenance_incident and maintenance_request actions either. This is just useful for single action submit requests - and for nothing else.

What about requests of type change_devel? Does it make sense this checkbox for this request type?

coolo:
> What about requests of type change_devel? Does it make sense this checkbox for this request type?

You can construct cases - but I'd still say no. The point of a devel project is that it's maintained elsewhere.

coolo:
> agree, however it has nothing to do with the bug fixed in this PR #8403. Could that be done in another PR @eduardoj @coolo ?

It very much has to do with this. Because #8403 is just a duplicate of https://github.com/openSUSE/open-build-service/issues/7297#issuecomment-477900733 and I select the checkbox from time to time just to annoy you :)

eduardoj:
@vpereira, @coolo, thanks for your feedback. Please have a look again.

PR #8416, opened 2019-09-21
Remove UserController

vpereira:
To accomplish it we did:

- Moved home action to WebuiController and adapted the routes
- Removed the old and unused routes '/home/my_work' and '/home/my_project'
- Removed the UserController

Note:

Since nobody access the following routes, and they were already marked as "old routes" since 2.9, I dropped them. Should I add a note to the release notes?

The routes: 

```
  get 'home/my_work' => :home
  get 'home/list_my' => :home
  get 'home/home_project' => :home_project
```

For me it would be ok  to remove the '/home' endpoint too and instead use the `/users/$USER` directly. What do you think @hennevogel ? I deprecated it,  removing it from the partials `_logged_in_user_navigation.html.haml` and `obs_factory/_personal_navigation.html.haml` but it still accessible.

### DIFF
--- a/src/api/app/views/layouts/webui/_logged_in_user_navigation.html.haml
+++ b/src/api/app/views/layouts/webui/_logged_in_user_navigation.html.haml
@@ -1,5 +1,5 @@
 %li.nav-item
-  = link_to(home_path, id: 'link-to-user-home', class: 'nav-link') do
+  = link_to(user_path(User.session!), id: 'link-to-user-home', class: 'nav-link') do
     = User.session!.login
 %li.nav-item
   - tasks = User.session!.tasks
--- a/ReleaseNotes-2.11
+++ b/ReleaseNotes-2.11
@@ -51,3 +51,9 @@ Intentional changes:
     * Kiwi Image Editor
     * Cloud upload for EC2
     * Cloud upload for Azure
+ 
+ * Removed the following routes:
+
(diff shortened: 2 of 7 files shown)


coolo:
I can tell you that home/my_work used to be my 'browser homepage' - but I assume most people don't care as much :)

vpereira:
> I can tell you that home/my_work used to be my 'browser homepage' - but I assume most people don't care as much :)

Looking the logs, just bots are crawling it. Today we have the "My Tasks" which probably makes more sense that this "my_work".

dmarcoux:
I would add a note in the release notes about the deleted routes.

src/api/app/views/layouts/webui/_logged_in_user_navigation.html.haml
```diff
@@ -1,5 +1,5 @@
 %li.nav-item
-  = link_to(home_path, id: 'link-to-user-home', class: 'nav-link') do
+  = link_to(user_path(User.session!.login), id: 'link-to-user-home', class: 'nav-link') do
```
dmarcoux: `.login` isn't needed.

vpereira: done

vpereira:
> I would add a note in the release notes about the deleted routes.

done!

PR #8420, opened 2019-09-24
Move user specific urls under scope 'my'

vpereira:
Some routes are user specific and they are bounded with User.session! therefore instead of force it to be under /users, we can create the routes like:

- /my/tasks
- /my/notifications
...

### DIFF
--- a/src/api/app/controllers/webui/users/rss_tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/rss_tokens_controller.rb
@@ -13,7 +13,7 @@ module Webui
           flash[:success] = 'Successfully generated your RSS feed url'
           User.session!.create_rss_token
         end
-        redirect_back(fallback_location: user_notifications_path)
+        redirect_back(fallback_location: my_notifications_path)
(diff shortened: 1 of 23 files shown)


vpereira:
> I like the idea, but it's somewhat going against the PR #8416 we merged yesterday.  We removed `/home` (it's only a redirection now). If we merge this PR, we should bring back `/home` and maybe even add `/home/project`, which would be the `project#show` of the user's home project.
> 
> What do you think?

The idea of the PR from yesterday was to get rid of the `UserController`, the action `home` was just moved to `WebuiController` without any further change. 

`/home/project` I don't think that make sense. To keep it consistent, maybe `/my/home` and `/my/profile` ? However there is a specific difference between what I did in this PR: All routes are not "permanent links", i.e they are based on the `User.session!`, different then `UsersController#show`

vpereira:
it's all green https://circleci.com/workflow-run/f9a1a1a8-6d0a-432b-b719-1f7a8f2b6a3b

PR #8424, opened 2019-09-24
Move leftover bootstrap feature tests to the main feature tests

krauselukas:
While removing the bento ui, a few feature test haven't
been moved from the seperate bootstrap folder to the main's
feature test folder.


hennevogel:
We don't use `MonitorPage` anymore?

krauselukas:
> We don't use `MonitorPage` anymore?

@hennevogel I couldn't find a place where its used, let's see, if its used the test's should fail right :D

krauselukas:
@hennevogel but yeah, its strange that it is not used anymore, so I will double check if a feature test got accidentally removed :)

krauselukas:
@hennevogel I created a seperate commit for removing the unused support module

PR #8425, opened 2019-09-24
Split routes.rb in smaller files (webui/api)

vpereira:
To make easier to navigate through the routes files, split it in two
domains:
  - webui (config/routes/webui_routes.rb)
  - api (config/routes/api_routes.rb)

There are some room for improvement, like I would like to have a base constrain with the bare minimum contrains that apply for webui and api and go specifics in the respective files.. We can do it however in a new iteration.


src/api/config/routes/routes_helper.rb
```diff
@@ -0,0 +1,44 @@
+
```
dmarcoux: This line should be removed.

src/api/config/routes/api_routes.rb
```diff
@@ -0,0 +1,408 @@
...
+    get 'issue_trackers/issues_in' => 'issue_trackers#issues_in'
+    resources :issue_trackers, only: [:index, :show, :create, :update, :destroy] do
+      resources :issues, only: [:show] # Nested route
+    end
+
+    ### /tag
+
+    # routes for tagging support
+    #
+    # get 'tag/_all' => 'tag',
```
dmarcoux: Should we remove those commented routes? I guess it's fine to do it.

dist/obs-server.spec
```diff
@@ -831,6 +831,9 @@ usermod -a -G docker obsservicerun
 
 /srv/www/obs/api/config/boot.rb
 /srv/www/obs/api/config/routes.rb
+/srv/www/obs/api/config/routes/api_routes.rb
```
hennevogel: You should add the whole directory, not just the files in it.

adrianschroeter: nah, it must not overwrite %config definitions for other files inside

coolo: I don't think config/routes should be %confg

src/api/config/routes/api_routes.rb
vpereira: I would remove it.

src/api/config/routes/routes_helper.rb
vpereira: done

src/api/config/routes/api_routes.rb
dmarcoux: Yes, I'm fine with this!

dist/obs-server.spec
vpereira: done @hennevogel , looks package is building successfully.

hennevogel:
@openSUSE/obs-frontend any other comments about this?

PR #8429, opened 2019-09-24
Allow for uploading multiple files to packages

hellcp:
Supersedes #7249 

To verify this feature

- Create a project and a package
- Visit add_file view
- Upload multiple files, add multiple empty files and add some links
- Submit
- Multiple files should be uploaded to the package

|Before|After|
|---|---|
|Screenshot Before|Screenshot After|

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -1,11 +1,9 @@
 require_dependency 'api_error'
 require 'builder/xchar'
-require 'builder'
 require 'rexml/document'
 require_dependency 'has_relationships'
 require_dependency 'opensuse/validator'
 require_dependency 'authenticator'
-require 'digest'
 
 # rubocop: disable Metrics/ClassLength
 class Package < ApplicationRecord
@@ -1281,6 +1279,7 @@ class Package < ApplicationRecord
     params = {}
     params[:comment] = opt[:comment] if opt[:comment]
     params[:user] = User.session!.login
+    params[:rev] = opt[:rev]
     Backend::Api::Sources::Package.write_file(project.name, name, opt[:filename], content, params)
 
     # KIWI file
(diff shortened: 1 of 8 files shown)


src/api/app/models/package.rb
```diff
@@ -1365,6 +1367,39 @@ def save_file(opt = {})
     sources_changed(wait_for_update: ['_aggregate', '_constraints', '_link', '_service', '_patchinfo', '_channel'].include?(opt[:filename]))
   end
 
+  def save_files(opt = {})
```
vpereira: I would prefer to see this kind of code - that has nothing to do with persistence - happening in another level, i.e a service object, being called in the controller. IMO `User.session` should leak as less as possible to the model layer..

vpereira:
great stuff @hellcp !

src/api/app/models/package.rb
hellcp: Yeah, it would probably be the best to do that in the controller, which would allow to also commit `_service` in the same filelist if correctly implemented.

adrianschroeter:
please make sure that this will result in a single commit ...

hellcp:
> please make sure that this will result in a single commit ...

It almost does, service models don't have support for custom rev, which is why that wasn't included yet

hellcp:
Before anybody actually uses that preview app, I gotta note, everything works, except:
* `_service` and `_service:whatever` files get overwritten no matter what you do
* services aren't done in the same commit

I have a bunch of service model reworks ahead of me and I didn't have the time to approach that properly

the rest *should work*

adrianschroeter:
I am relative sure that you can not overwrite _service:* files from UI. They just get replaced by service run.

hellcp:
This was superseded

PR #8437, opened 2019-09-25
Display the 'Staging' tab only for authorized users

### LINKED ISSUE #8409: Show StagingWorkflow#new only for authorized users
If there is no `StagingWorkflow` for a project the `Staging` tab is shown for everybody anyway, regardless if they would be allowed to create it. This shouldn't be the case.

dmarcoux:
When the project isn't a staging project.

The other case is when the 'Staging' tab is active. It doesn't rely on authorization.

Fixes #8409

### DIFF
--- a/src/api/app/views/webui/project/_tabs.html.haml
+++ b/src/api/app/views/webui/project/_tabs.html.haml
@@ -33,7 +33,7 @@
       %li.nav-item
         = tab_link('Pulse', project_pulse_path(project))
       - active = controller_path.starts_with?('webui/staging')
-      - if !project.staging_project? || active
+      - if (!project.staging_project? && policy(Staging::Workflow.new(project: project)).create?) || active
         %li.nav-item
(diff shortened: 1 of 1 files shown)


dmarcoux:

coolo:
the staging tab is quite confusing. https://build.opensuse.org/project/show/openSUSE:Leap:15.2 shows a Staging tab if logged in, but not if logged in. And that tab shows the work flow.

But if I log out, I can still access https://build.opensuse.org/staging_workflows/136 and it will be on 15.2's Staging tab. But that tab disappears if you switch to Overview.

dmarcoux:
It's true that it's quite confusing.

To be honest, I never understood why we decided to go with this. I would prefer that we always display the `Staging` tab. Whenever the project is a staging workflow project or a staging project, the tab would link to the staging workflow show page. Whenever it's not the case, the tab would be hidden for users which cannot create the staging workflow. Otherwise, clicking the tab would link to the creation of the staging workflow.

What do you think?

PR #8439, opened 2019-09-25
Use File.read to force the new routes files to be reloaded

vpereira:
In development environment, after a file change, the routes.rb were
reloaded but the files under config/routes don't. In that way, the `File.read` will run always when `routes.rb` gets reloaded.

How to test it:

- run it locally (yes you have to checkout the branch)
- change a controller, i.e open the webui/project_controller#show and add an instance variable `foo = true` go to your browser and try to reload the application.. all routes should be still accessible.

### DIFF
--- a/src/api/config/routes/api_routes.rb
+++ b/src/api/config/routes/api_routes.rb
@@ -1,5 +1,3 @@
-require_relative 'routes_helper'
-
 OBSApi::Application.routes.draw do
   cons = RoutesContrains::CONS
 
--- a/src/api/config/routes.rb
+++ b/src/api/config/routes.rb
@@ -1,8 +1,15 @@
+require_relative 'routes/routes_helper'
+
+class ActionDispatch::Routing::Mapper
+  def draw(routes_name)
+    instance_eval(File.read(Rails.root.join("config/routes/#{routes_name}_routes.rb")))
+  end
+end
+
 OBSApi::Application.routes.draw do
   mount Peek::Railtie => '/peek'
-  require_relative 'routes/routes_helper'
-  require_relative 'routes/webui_routes'
-  require_relative 'routes/api_routes'
+  draw :webui
+  draw :api
 end
 
 OBSEngine::Base.subclasses.each(&:mount_it)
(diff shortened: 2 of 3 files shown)


src/api/config/application.rb
```diff
@@ -18,6 +18,7 @@ class Application < Rails::Application
     # config.autoload_paths += %W(#{config.root}/extras)
     config.eager_load_paths << Rails.root.join('lib', 'backend')
 
+    # Make sure all routes are loaded and automatically reloaded (specially in development)
```
dmarcoux: I wouldn't change this file. It looks like you tested something and forgot to remove those changes.

krauselukas: [nitpick] don't know if this comment is useful at this place

src/api/config/routes/api_routes.rb
```diff
@@ -1,4 +1,4 @@
-require_relative 'routes_helper'
+require File.join(Rails.root, 'config', 'routes', 'routes_helper')
 
 OBSApi::Application.routes.draw do
```
dmarcoux: This shouldn't be needed since it's already in `config/routes.rb`. Same thing for `webui_routes.rb`.

I tried it and it works.

src/api/config/routes/api_routes.rb
```diff
@@ -1,4 +1,4 @@
-require_relative 'routes_helper'
+require File.join(Rails.root, 'config', 'routes', 'routes_helper')
```
dmarcoux: This can be replaced by `require_relative 'routes/routes_helper'` at the top of `config/routes.rb`. Same thing for `webui_routes.rb`.

vpereira: did you try it? `require_relative` and `instance_load` doesn't work well. hint try it locally.

vpereira: Hi @dmarcoux , no unfortunately it cannot be replaced. Because of the `instance_eval`, the basepath for the `require_relative` cannot be inferred. Or did you try it locally and worked for you?

dmarcoux:
This [comment](https://github.com/openSUSE/open-build-service/pull/8439/files#r328184471) wasn't addressed.

PR #8445, opened 2019-09-26
[dist] chrpath in obs-bundled-gems.spec no longer mandatory

M0ses:
The new version of obs-service-bundle_gems has the
"force_ruby_platform" enabled. This leads to the following error
while build obs-bundle-gems:

```
[  396s] + chrpath --delete '/home/abuild/rpmbuild/BUILDROOT/obs-bundled-gems-2.11~alpha.20190925T165156.cb40d16c84-8857.1.x86_64/usr/lib64/obs-api/ruby/*/gems/sassc-*-x86_64-linux/lib/sassc/libsass.so'
[  396s] open: No such file or directory
[  396s] elf_open: Invalid argument
```

because "libsass.so" no longer gets packaged.

This patch simply returns a "true" value if chrpath fails so the %install
section does not break anymore.

This PR is a follow up for

https://github.com/openSUSE/obs-service-bundle_gems/pull/14

@bugfinder - Thanks for preparing

### DIFF
--- a/dist/obs-bundled-gems.spec
+++ b/dist/obs-bundled-gems.spec
@@ -142,9 +142,6 @@ find %{buildroot} -type f -print0 | xargs -0 grep -l /usr/bin/env | while read f
   chmod a-x $file
 done
 
-# Remove rpaths from files
-chrpath --delete %{buildroot}%_libdir/obs-api/ruby/*/gems/sassc-*-x86_64-linux/lib/sassc/libsass.so
-
 %files
 %_libdir/obs-api
 


vpereira:
couldn't we just remove the line `chrpath --delete '/home/abuild/rpmbuild/BUILDROOT/obs-bundled-gems-2.11~alpha.20190925T165156.cb40d16c84-8857.1.x86_64/usr/lib64/obs-api/ruby/*/gems/sassc-*-x86_64-linux/lib/sassc/libsass.so` ?

dist/obs-bundled-gems.spec
```diff
@@ -142,8 +142,10 @@ find %{buildroot} -type f -print0 | xargs -0 grep -l /usr/bin/env | while read f
   chmod a-x $file
 done
 
-# Remove rpaths from files
-chrpath --delete %{buildroot}%_libdir/obs-api/ruby/*/gems/sassc-*-x86_64-linux/lib/sassc/libsass.so
+# Remove rpaths from files (legacy)
+# if service bundle_gems ran with force_ruby_platform
```
hennevogel: I really don't think we need to cover for this case. The service from now on will always run with `force_ruby_platform=true`...

Let's just remove this `chrpath` call and be done with it.

dist/obs-bundled-gems.spec
```diff
@@ -142,8 +142,10 @@ find %{buildroot} -type f -print0 | xargs -0 grep -l /usr/bin/env | while read f
...
+# the "|| true" prevents package build from breaking
+chrpath --delete %{buildroot}%_libdir/obs-api/ruby/*/gems/sassc-*-x86_64-linux/lib/sassc/libsass.so || true
```
eduardoj: Now that the bundle_gems service run always with `force_ruby_platform true` maybe we can get rid of all the `chrpath` line. Can we try removing the `chrpath` line?

vpereira: sounds good to me!

M0ses: I`m ok with that if you are

PR #8447, opened 2019-09-26
 Allow accessing staging project show page without being logged in 

### LINKED ISSUE #7353: Staging project show page requires login
https://build.opensuse.org/staging_workflows/6 is accessible without login, but
https://build.opensuse.org/staging_workflows/6/staging_projects/TestLinux:Staging:A requires login. There is no reason for that.

krauselukas:
There is no reason to require being logged in to obs, in order
to see the content of the staging project show page.

Fixes #7353

### DIFF
--- a/src/api/app/controllers/webui/staging/projects_controller.rb
+++ b/src/api/app/controllers/webui/staging/projects_controller.rb
@@ -1,7 +1,7 @@
 module Webui
   module Staging
     class ProjectsController < WebuiController
-      before_action :require_login
+      before_action :require_login, except: :show
       before_action :set_staging_workflow
       after_action :verify_authorized, except: :show
 


krauselukas:

src/api/spec/features/webui/staging/projects_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'browser_helper'
```
hennevogel: A feature spec to test behavior of a `before_action` is a bit overkill. Have a look at [`use_before_action`](http://matchers.shoulda.io/docs/v4.1.2/Shoulda/Matchers/ActionController.html#use_before_action-instance_method) like we do in the other controller tests.

krauselukas: @hennevogel I removed the feature test, I know it was a bit overkill for this scenarion, was more the idea of getting the feature test's started for the staging. But I might tackle this in another PR, with things that a more worth to test.
This is kind of hard to test in the controller, ```use_before_action``` seems to check for the ```before_action``` globally in the controller which of course is there, but I can't test for the ```except``` of a certain method. Since we are not really testing for this in other controller test's, I would just leave it out.

hennevogel:
Coveralls hangs. Coverage is fine, see codecov.

PR #8448, opened 2019-09-26
Make sure that if config/routes/* files are changed, they are reloaded

vpereira:
Before this changes, if you do a change in a route in `config/routes/*`, in the development environment, the server won't reload it. 

With this change, you are able to add a new route to i.e `config/routes/webui_routes` and it will be reloaded automatically.


src/api/config/application.rb
```diff
@@ -119,5 +120,9 @@ class Application < Rails::Application
         # See Rails::Configuration for more options
       end
     end
+
+    config.after_initialize do
```
hennevogel: Why both?

Issue #7355, opened 2019-04-06
Wrong error message during staging request


coolo:
```POST https://api-test.opensuse.org/staging/openSUSE:Factory/staging_projects/openSUSE:Factory:Staging:A/staged_requests

<status code="invalid_request">
  <summary>Assigning requests to openSUSE:Factory:Staging:A failed: Request '292' does not exist or target_project is not 'openSUSE:Factory'.</summary>
</status>

```

But that's wrong (and having an error message that says 'this or that' sounds strange in any case - better (not for this bug but in general IMO would be "No request 292 found with target project openSUSE:Factory")

```
Request: #292

  submit:       home:coolo:fakedevel/doxygen@1 -> openSUSE:Factory


Message:
update from Factory

State:   review     2019-04-05T13:17:15 coolo
Comment: Being evaluated by staging project "openSUSE:Factory:Staging:B"

Review:  new        Project: openSUSE:Factory:Staging:B                2019-04-05T13:17:15 coolo                 
  Being evaluated by staging project "openSUSE:Factory:Staging:B"
         accepted   Group: factory-staging                             2019-04-05T06:49:27 coolo                 
  Picked openSUSE:Factory:Staging:B

History: 2019-04-05T13:17:15 coolo        Review got accepted
         2019-04-05T13:17:15 coolo        Request got a new review request
         2019-04-05T06:49:27 coolo        Request created
```

I assume it's looking for the review by_group as well?

PR #8450, opened 2019-09-26
Show a proper error message when excluding a request

### LINKED ISSUE #8355: Exclude requests which are `ready`
In the staging workflow, it should be possible to exclude requests which are `ready`.

To reproduce, create a staging workflow for a project (`home:Admin` for example) and have a request open for this project.

Then run this command (by adapting it to your project name and request ID):
```
osc api staging/PROJECT_NAME/excluded_requests -X POST -d "<excluded_requests><request number='REQUEST_ID' description='invalid request' /></excluded_requests>"
```

This will fail:
```
Excluding requests for #<Staging::Workflow:0x00007f1eea77f410> failed: Request : Number can't be blank and Request is not a number.
```

eduardoj:
Make API and web UI consistent with the same behaviour and error messages.

Fixes #8355.

### DIFF
--- a/src/api/app/models/staging/request_excluder.rb
+++ b/src/api/app/models/staging/request_excluder.rb
@@ -3,15 +3,7 @@ class Staging::RequestExcluder
   attr_accessor :requests_xml_hash, :staging_workflow
 
   def create
-    requests_to_be_excluded.map do |request|
-      bs_request = staging_workflow.unassigned_requests.find_by_number(request[:number])
-      request_excluded = Staging::RequestExclusion.new(bs_request: bs_request,
-                                                       number: bs_request.try(:number),
-                                                       description: request[:description],
-                                                       staging_workflow: staging_workflow)
-
-      errors << "Request #{request_excluded.bs_request_id}: #{request_excluded.errors.full_messages.to_sentence}." unless request_excluded.save
-    end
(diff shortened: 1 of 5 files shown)


eduardoj:

src/api/app/models/staging/request_excluder.rb
```diff
@@ -34,11 +26,36 @@ def valid?
...
     [requests_xml_hash[:number]].flatten.map(&:to_i)
   end
+
+  def valid_request?(bs_request, request_number)
+    return true if bs_request && bs_request.staging_project.nil?
+
+    errors << if bs_request
+                "Request #{request_number} could not be excluded because is staged in: #{bs_request.staging_project}"
+              else
+                "Request #{request_number} doesn't exist or it doesn't belong to this project"
```
dmarcoux: It would be nice to provide a more accurate error message to the user.

To make this possible, we need to retrieve the request and if it exists, check if it belongs to the project. Right now, we only retrieve the request if it belongs to the project, so it's preventing us from knowing if this request simply doesn't exist.

eduardoj: True. I will take care of #7355. Can we accept this pull request then, as it is only fixing #8355? Or do you think it is needed to fix #8355 in this same pull request?

hennevogel:

PR #8452, opened 2019-09-26
Move binary specs to their own file

hennevogel:
The Webui::PackageController spec is over 1600 lines long, it accurately
reflects what is in Webui::PackageController but we do not have to mimic the
length of the Class in the spec. Let's move out some specs for actions that
belong together so that at least the spec becomes a reasonable, manageable size.


vpereira:
> I also disabled VCR on all specs that did not need the backend, that is the reason for the massive cassette drop.

but sounds like it still need the VCR files https://build.opensuse.org/package/live_build_log/OBS:Server:Unstable:TestGithub:PR-8452/obs-server:obs-api-testsuite-rspec/SLE_15/x86_64

hennevogel:
Codecov hangs. Coveralls is fine. Merging.

PR #8455, opened 2019-09-27
Fixes #8286: Return XML if staging workflow was successfully deleted

### LINKED ISSUE #8286: Delete route /staging/home:mstrigl/workflow returns nothing
# Deleting a staging workflow returns nothing
Using the API to delete a staging workflow for a project works, but returns nothing.
DELETE https://api.opensuse.org/staging/home:mstrigl/workflow 

Expected Result
============
This should be returned according to the documentation:
```
<status code="ok">
    <summary>Ok</summary>
</status>

```
How to Reproduce
===============
1. create a staging
2. osc api -X DELETE /staging/home:mstrigl/workflow
3. See no return

vpereira:
If an User deletes a staging workflow via API, it should return the "OK" XML as every other endpoint.

Fixes: #8286

### DIFF
--- a/src/api/app/controllers/staging/workflows_controller.rb
+++ b/src/api/app/controllers/staging/workflows_controller.rb
@@ -25,6 +25,7 @@ class Staging::WorkflowsController < ApplicationController
     authorize @staging_workflow
 
     @staging_workflow.destroy!
+    render_ok
   end
 
   def update


vpereira:

src/api/app/controllers/staging/workflows_controller.rb
```diff
@@ -25,6 +25,7 @@ def destroy
     authorize @staging_workflow
 
     @staging_workflow.destroy!
+    render_ok
```
DavidKang: I think we should also render something if we the staging could not be destroyed.

PR #8456, opened 2019-09-27
Refactor routes for webui/patchinfo

vpereira:
- Make it resourceful
- Adapt specs
- Adapt views


vpereira:
I will close it for now, because it needs some thinking about the routes

PR #8460, opened 2019-09-27
Remove staged request without specify a staging project

### LINKED ISSUE #8369: Deleting requests from staging projects without specifying the staging project is not possible
# Deleting requests from staging projects without specifying the staging project is not possible.
Since one request can only be linked to one staging project an API call like this should be possible:

`DELETE /staging/<staging_workflow_project>/staging_projects/staged_requests/<rqid>`

Expected Result
============
Delete requests without specifying the staging project.

How to Reproduce
===============
It is simply not possible.

Further Information
===============
* I have a workaround in osc for this at the moment.

DavidKang:
We added a alternative route to remove a single staged request without
the need of specifying a staging project.

Fix #8369


src/api/config/routes/api_routes.rb
```diff
@@ -278,6 +278,7 @@
       get 'staged_requests' => 'staged_requests#index', constraints: cons
       resource :staged_requests, only: [:create, :destroy], constraints: cons
     end
+    delete 'staging_projects/staged_requests/:request_id' => :destroy, constraints: cons, controller: 'staged_requests', as: 'single_staged_request'
```
krauselukas: I think this new endpoint should be added to the API docs

krauselukas: [nitpick] Since it is an api endpoint, and not used in the webui the ```as: 'single_staged_request``` shouldn't be necessary, right?

docs/api/api/api.txt
```diff
@@ -1985,6 +1985,7 @@ XmlResult: status_ok
 
 <staging_workflow_project>: The name of the staging workflow project, e.g. home:user1
 <staging_project>: The name of the staging project, e.g. home:user1:Staging:A
+<request>: The number of the request, e.g. 12345
```
coolo: Careful. everwhere else in this document we talk about request ids when it talks about the user given number - and about 'number' if we mean a count of requests.

docs/api/api/api.txt
```diff
@@ -1995,8 +1996,12 @@ POST /staging/<staging_workflow_project>/staging_projects/<staging_project>/stag
 XmlBody: remove_staged_requests
 XmlResult: status_ok
 
+DELETE /staging/<staging_workflow_project>/staging_projects/staged_requests/<request>
+  Remove a bs_request from the staged requests of a staging project.
+XmlResult: status_ok
+
 DELETE /staging/<staging_workflow_project>/staging_projects/<staging_project>/staged_requests
```
hennevogel: Why do we need both routes? Why not only one route, without `StagingProject`, where I send XML with one or more requests in it?

DavidKang: @hennevogel, I applied the changes that you suggest.

docs/api/api/api.txt
```diff
@@ -1995,8 +1995,8 @@ POST /staging/<staging_workflow_project>/staging_projects/<staging_project>/stag
 XmlBody: remove_staged_requests
 XmlResult: status_ok
 
-DELETE /staging/<staging_workflow_project>/staging_projects/<staging_project>/staged_requests
-  Remove one or multiple bs_requests to the staged requests of a staging project.
+DELETE /staging/<staging_workflow_project>/staged_requests
+  Remove one or multiple bs_requests from the staged requests of a staging project.
```
hennevogel: > Remove one or multiple bs_requests from the staged requests of a staging project.

Remove requests from the staging projects they are staged in.

src/api/app/controllers/staging/staged_requests_controller.rb
```diff
@@ -62,21 +63,33 @@ def xml_hash
   end
 
   def set_staging_project
-    @staging_project = Project.get_by_name(params[:staging_project_name])
+    @staging_project = @staging_workflow.staging_projects.find_by(name: params[:staging_project_name])
+    return if @staging_project
+    render_error(
+      status: 404,
+      errorcode: 'not_found',
+      message: "Staging Project '#{params[:staging_project_name]}' doesn't exist or it doesn't belong to the Staging: '#{@staging_workflow.project}'"
```
hennevogel: The second part of this message is not true anymore. 

> "Staging Project '#{params[:staging_project_name]}' doesn't exist or it doesn't belong to the Staging: '#{@staging_workflow.project}'"

"Staging Project '#{params[:staging_project_name]}' not found in in Staging: '#{@staging_workflow.project}'"

src/api/app/controllers/staging/staged_requests_controller.rb
```diff
@@ -31,12 +33,11 @@ def create
   end
 
   def destroy
-    authorize @staging_project, :update?
+    authorize @staging_workflow, :update?
```
hennevogel: :thought_balloon: I guess at some point this needs to be more fine grained again and move into `::Staging::StageRequests`

src/api/app/models/staging/stage_requests.rb
```diff
@@ -102,4 +105,28 @@ def branch_package(bs_request_action)
...
+                "Can't change staged requests '#{request.number}': Project '#{staging_project}' is being accepted."
+              end
+    return false
+  end
+
+  def missing_deleted_packages(package_names, result)
+    not_deleted_packages = package_names - result.pluck(:name)
+    return if not_deleted_packages.empty?
+
+    errors << "Could not delete packages #{not_deleted_packages.to_sentence}."
```
hennevogel: Why and what can I do about  it?

src/api/app/models/staging/stage_requests.rb
```diff
@@ -102,4 +105,28 @@ def branch_package(bs_request_action)
...
+  end
+
+  def missing_unassigned_requests(request_numbers, requests)
+    not_unassigned_requests = request_numbers - requests.pluck(:number).map(&:to_s)
+    return if not_unassigned_requests.empty?
+
+    errors << "Requests with number #{not_unassigned_requests.to_sentence} not found."
```
hennevogel: Found where? While doing what?

src/api/app/controllers/staging/staged_requests_controller.rb
DavidKang: true!

docs/api/api/api.txt
coolo: why is it saying bs_request anyway? I'm painfully aware why the class is named this way, but in api documentation it should be request, no?

src/api/app/models/staging/stage_requests.rb
```diff
@@ -19,17 +19,22 @@ def create
...
-    result = staging_project.packages.where(name: package_names).destroy_all
-    not_deleted_packages = package_names - result.pluck(:name)
+    not_deleted_packages = {}
 
     requests.each do |request|
-      add_review_for_unstaged_request(request)
+      staging_project = request.staging_project
+      next unless unstageable?(request, staging_project)
+
+      staging_project.staged_requests.delete(request)
```
hennevogel: Wouldn't it be better if this would be the last thing we do in this loop?

src/api/app/models/staging/stage_requests.rb
```diff
@@ -117,8 +121,45 @@ def add_review_for_staged_request(request)
...
+                "Can't change staged requests '#{request.number}': Project '#{staging_project}' is being accepted."
+              end
+    return false
+  end
+
+  def missing_packages(not_deleted_packages)
+    return if not_deleted_packages.empty?
+
+    message = not_deleted_packages.map { |staging_project, package_names| "'#{package_names.flatten.to_sentence}' from '#{staging_project}'" }
+    errors << "The following packages couldn't be deleted: #{message.to_sentence}"
```
hennevogel: Hm under which conditions can this happen?

DavidKang: I just wanted to be sure that it was controlled if something strange happens. Are you suggesting that this is not needed?

hennevogel: No I want the error message to explain what is going on and what people can do about it.

DavidKang:
@hennevogel, could you review it again?

PR #8474, opened 2019-09-30
Refactor routes patchinfo

vpereira:
Refactor `Patchinfo` related routes to a singular resource, with one exception:  `show` 
The missing route is going to be tackled later by some redirects rules because it follows a special routes naming. It would be done in a follow-up PR


vpereira:
Done @krauselukas!  Rebased on master and squashed some related commits.

PR #8475, opened 2019-09-30
Link to the staging workflow for the 'Staging' tab in the context of a staging project

### LINKED ISSUE #7357: Staging project don't have staging tab
If I come from the [staging workflow](https://build-test.opensuse.org/staging_workflows/1), the [staging project info](https://build-test.opensuse.org/staging_workflows/1/staging_projects/openSUSE:Factory:Staging:A) is a tab in the project and I can switch to other tabs from there. But if I do, the Staging tab is gone and I can't switch back.

dmarcoux:
Closes #7357

### DIFF
--- a/src/api/app/views/webui/project/_tabs.html.haml
+++ b/src/api/app/views/webui/project/_tabs.html.haml
@@ -33,9 +33,17 @@
       %li.nav-item
         = tab_link('Pulse', project_pulse_path(project))
       - active = controller_path.starts_with?('webui/staging')
-      - if (!project.staging_project? && policy(Staging::Workflow.new(project: project)).create?) || active
+      - if project.staging_project?
+        %li.nav-item
+          = tab_link('Staging', staging_workflow_path(project.staging_workflow), active)
+      - elsif policy(project.staging || Staging::Workflow.new(project: project)).create?
         %li.nav-item
           = tab_link('Staging', new_staging_workflow_path(project: project), active)
+      - elsif project.staging.persisted?
+        -# If Staging::Workflow.new(...) is executed above, project.staging will be a new object.
+           It will be persisted if the project is really a staging workflow project.
+        %li.nav-item
+          = tab_link('Staging', staging_workflow_path(project.staging), active)
     %li.nav-item.dropdown
       %a.nav-link.dropdown-toggle{ href: '#', 'data-toggle': 'dropdown', 'role': 'button', 'aria-expanded': 'false', 'aria-haspopup': 'true' }
       .dropdown-menu.dropdown-menu-right


dmarcoux:

hennevogel:
At least until we want nested staging workflows! :cowboy_hat_face:

dmarcoux:
@hennevogel Please have another look since the code changed a bit.

src/api/app/views/webui/project/_tabs.html.haml
```diff
@@ -33,9 +33,15 @@
...
         = tab_link('Pulse', project_pulse_path(project))
       - active = controller_path.starts_with?('webui/staging')
-      - if (!project.staging_project? && policy(Staging::Workflow.new(project: project)).create?) || active
+      - if project.staging_project?
+        %li.nav-item
+          = tab_link('Staging', staging_workflow_path(project.staging_workflow), active)
+      - elsif policy(Staging::Workflow.new(project: project)).create?
         %li.nav-item
           = tab_link('Staging', new_staging_workflow_path(project: project), active)
+      - elsif (staging = Staging::Workflow.find_by(project_id: project.id))
```
dmarcoux: Using `project.staging` doesn't work here because the association returns a `Staging::Workflow` even when there isn't one for the project. It returns an empty `Staging::Workflow`, except for the `project_id` field which contains the value of `project.id`. I don't understand why it does this to be honest.

hennevogel: Hmmm :thinking:

coolo: I can't reproduce:

```irb(main):002:0> Project.find_by(name: 'home:coolo').staging
=> nil

Comment (dmarcoux) [reply src/api/app/views/webui/project/_tabs.html.haml]: This happens after `policy(Staging::Workflow.new(project: project)).create?` is executed. To reproduce it in the view:

1. Add `binding.pry` above `- if project.staging_project?` (line 36 in the changed file)
2. Go to the project#show page of a project with a staging workflow.
3. Execute `project.staging` in `pry`, see the `Staging::Workflow` as expected:
```
#<Staging::Workflow:0x00007f2e3a6a8208 id: 3, project_id: 28, created_at: Wed, 25 Sep 2019 11:19:06 UTC +00:00, updated_at: Wed, 25 Sep 2019 11:19:06 UTC +00:00, managers_group_id: 4>
```
4. Execute `policy(Staging::Workflow.new(project: project)).create?` in `pry`
5. Execute again `project.staging` in `pry`, see the `Staging::Workflow`:
```
#<Staging::Workflow:0x00007f2e3b9a1d50 id: nil, project_id: 28, created_at: nil, updated_at: nil, managers_group_id: nil>
```

Kudos to @DavidKang for pointing me to the cause (the policy). I still don't see why it does this though.

coolo: I see - `Staging::Workflow::new(project:project)` is binding the new object to the `belongs_to` association. But then make it simple and use project.staging_workflow_id instead of running a N+1 query (where N is fortunately 1 :)

dmarcoux: `project.staging_workflow_id` is for the `staging_projects` association of a `Staging::Workflow`, so I cannot use it to check if a project has a staging workflow.

coolo: right. How about `project.staging.id` then? :)

coolo: But even that will cause an extra query if you don't prefetch

dmarcoux: I ended up going with `persisted?`.

hennevogel: Can you file an issue for the policy please?

PR #8477, opened 2019-09-30
Use symbol instead of string to access broken_packages hash in staging project

### LINKED ISSUE #7351: Unresolvables in 'images' repo should only be errors if finished building
As long as the 'standard' repo is building, unresolvables in any repo aren't worth reporting.

The old dashboard shows a green checkmark, the staging workflow shows 2 unresolvables for products. Please fix.

krauselukas:
While migrating the old staging workflow to the new one, the
broken_packages hash moved from strings to symbols for the keys.
One of the checks still uses strings to access the hash, which
led to wrong results.

Fixes #7351

### DIFF
--- a/src/api/app/models/concerns/staging_project.rb
+++ b/src/api/app/models/concerns/staging_project.rb
@@ -194,7 +194,7 @@ module StagingProject
       add_building_repositories(result) if building
     end
 
-    @broken_packages.reject! { |package| package['state'] == 'unresolvable' } if @building_repositories.present?
+    @broken_packages.reject! { |package| package[:state] == 'unresolvable' } if @building_repositories.present?
   end
 
   def add_broken_packages(result, building)


dmarcoux:
The commit message is hard to follow. Too many `while` and `which`. Especially this sentence, it's looooooooonnnng:

>One of the checks, which sorts out packages from the hash which are in state unresolvable while there are still repositories in the build process, still used strings to access the hash, which led to an wrong result.

I suggest going straight to the point with something like:

>This change should have happened while migrating code from the old staging workflow to the new one. The hash is using symbols.

krauselukas:
> The commit message is hard to follow. Too many `while` and `which`. Especially this sentence, it's looooooooonnnng:
> 
> > One of the checks, which sorts out packages from the hash which are in state unresolvable while there are still repositories in the build process, still used strings to access the hash, which led to an wrong result.
> 
> I suggest going straight to the point with something like:
> 
> > This change should have happened while migrating code from the old staging workflow to the new one. The hash is using symbols.

@dmarcoux I shortend the commit message :)

krauselukas:

Issue #7359, opened 2019-04-06
Staging projects always end in 100% building


coolo:
https://build-test.opensuse.org/staging_workflows/1 shows E in 100% building, but there is nothing building.

For obs_factory I fixed that in bf36f59fad287ce55ad9669eb5243daa8d572753, but the staing workflow was left in the broken state we see now. I didn't touch it as the current solution in obs_factory isn't perfect either - it needs an extra state.

krauselukas:
@coolo do I see it right, the bug is basically showing the overall state of the a staging project in state building if there are disabled repositories, and introducing a new state for this scenario is more like a feature?

coolo:
the bug is that it's impossible to get green stagings with bootstrap_copy disabled. The feature is ignoring checks for disabled repositories. one way or the other. For obs_factory I stay in testing, but this is confusing as well - but better than what we have with staging workflow.

krauselukas:
What was the reason to keep the state in building if repositories are disabled?

Imagine the following scenario. Your staging project includes one standard repo, which is used to build the packages against. Additionally there is one image repo, which is used to build images including the mentioned packages and to test them through the status api in openQA. When there are still broken or unresolvable packages, you don't want always to build new images and run the tests in openQA or to see outdated results of the checks. So you disable the image repository. 
But you don't just want the overall state of the staging project to switch to acceptable since you want to prevent people accidentally forgetting about the status checks and accepting the requests.

Difficulties?

There are certain cases where a staging project always has an disabled repository, not just temporary. So simply checking if there is a disabled repo, and setting the overall state to something else than acceptable is not the final solution.

What now?

Simply setting the overall state of an staging project to 'building' if there is an disabled repo, makes the staging workflow unusable. So for now I will revert it (see https://github.com/openSUSE/open-build-service/pull/8479), and ignore the disabled repositories. But for the future this needs a solution.

@coolo please correct me, if I express something wrong :)

PR #8479, opened 2019-09-30
Don't set overall state of staging project to building if disabled repos are present

krauselukas:
Keeping the overall state of a staging project in building
if there is a disabled repository, makes the staging workflow
unusable.

Related to #7359


krauselukas:

PR #8485, opened 2019-10-01
Add missing XmlBody to DELETE staged requests

dmarcoux:


dmarcoux:
Oh right, the apidocs aren't working in the review app.

docs/api/api/api.txt
```diff
@@ -1997,6 +1997,7 @@ XmlResult: status_ok
 
 DELETE /staging/<staging_workflow_project>/staging_projects/<staging_project>/staged_requests
   Remove one or multiple bs_requests to the staged requests of a staging project.
+XmlBody: remove_staged_requests
```
eduardoj: Good catch. Curious that this same file is used in line 1995 for the POST request.

dmarcoux: Yes... I was thinking if we should rename the file, even though it's the same structure for both.

Issue #8488, opened 2019-10-01
Repotype: staticlinks creates duplicate file name links


Vogtinator:
Currently some kiwi appliances like `openSUSE-MicroOS.x86_64-16.0.0-kvm-and-xen-Snapshot20190926.qcow2` generate symlinks like

```
openSUSE-MicroOS.x86_64-kvm-and-xen.qcow2
openSUSE-MicroOS.x86_64-kvm-and.qcow2
openSUSE-MicroOS.x86_64-kvm.qcow2
openSUSE-MicroOS.x86_64-and.qcow2
```

AFAICT that's not caused by the regex itself, but rather that `createrepo_staticlinks` iterates over the links it just created and creates links to them...

vmoutoussamy:
Just for the record, JeOS is affected on https://download.opensuse.org/distribution/leap/15.1/jeos/ and https://download.opensuse.org/tumbleweed/appliances/

eduardoj:
Handled in #8490.

PR #8490, opened 2019-10-01
[backend] avoid multiple static link creation

adrianschroeter:
issue obs#8488

Don't create static links for links already.
But keep working for directories (for product build names without
matching name)


adrianschroeter:
https://github.com/openSUSE/open-build-service/issues/8488

PR #8499, opened 2019-10-02
Call explicitly name on a class

### LINKED ISSUE #8232: [production][comments#create] undefined method `find' for nil:NilClass
comments#create threw this 1 times
<pre>
undefined method `find&#39; for nil:NilClass
</pre>
[https://build.opensuse.org/comments](https://build.opensuse.org/comments)"

[See this exception on Errbit](http://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5d6cf51bc7b48d000db006ec)

eduardoj:
This way we make sure that a string is returned, instead of a class.

Errbit is receiving exceptions where the parameters receive something like this:

```
"commentable_type" => "+HCX3IyBjrtZ6Z3UQhVYa8OT3v7eye52V1jbcILfgWtlzj7...",
```

This should fix #8232.

### DIFF
--- a/src/api/app/views/webui/comment/_comment_field.html.haml
+++ b/src/api/app/views/webui/comment/_comment_field.html.haml
@@ -1,5 +1,5 @@
 = form_for(comment, action: 'create', method: :post, remote: true, html: { class: 'new-comment-form' }) do |f|
-  = hidden_field_tag :commentable_type, commentable.class
+  = hidden_field_tag :commentable_type, commentable.class.name
   = hidden_field_tag :commentable_id, commentable.id
   = f.hidden_field :parent_id, value: comment.parent_id
   ~ f.text_area :body, rows: '4', placeholder: 'Add a new comment (markdown markup supported)', required: true,


hennevogel:
Hm, how can this lead to calling `find` on a `nil:NilClass`?

eduardoj:
@hennevogel:

> Hm, how can this lead to calling `find` on a `nil:NilClass`?

The error `undefined method 'find' for nil:NilClass` is thrown in `app/controllers/webui/comments_controller.rb`, in line 51.

These are lines 50 and 51:
```
commentable = [Project, Package, BsRequest].find { |klass| klass.name == params[:commentable_type] }                                              
@commented = commentable.find(params[:commentable_id])
```

In the case that `params[:commentable_type]` is something like `"+HCX3IyBjrtZ6Z3UQhVYa8OT3v7eye52V1jbcILfgWtlzj7..."`, `comentable` will be assigned to `nil` (line 50).

And in line 51 an error will be thrown as a result of calling `nil.find()`.

The point is that `comentable_type` should have a string with the name of the class (`"BsRequest"`) instead of a class name (`BsRequest`).

hennevogel:

PR #8501, opened 2019-10-02
Append ' ago' to the time duration in the package change column

danidoni:
We use the same convention in some other places.

**Before**
image

**After**
image


hennevogel:

PR #8515, opened 2019-10-05
Switch back to upstream influxdb-rails

hennevogel:
The latest beta covers everything we need, let's use it.


hennevogel:

PR #8521, opened 2019-10-07
Avoid duplication in broken packages list

### LINKED ISSUE #7362: Staging project show duplicates broken packages per arch
```
 Broken packages	btrfsprogs (failed), build (failed), perl-Glib (failed), btrfsprogs (failed), build (failed), orc (failed), perl-Glib (failed)
```

Both btrsprogs link to the same page (https://build-test.opensuse.org/package/show/openSUSE:Factory:Staging:A/btrfsprogs in my case), so without the architecture given it's rather useless duplication.

saraycp:
The same package appeared more than once because it failed with
different architectures. However, the list is not making distinction by
architecture and the link always points to the same page. So, we will
show each package only once.

Fixes #7362

### DIFF
--- a/src/api/app/models/concerns/staging_project.rb
+++ b/src/api/app/models/concerns/staging_project.rb
@@ -197,7 +197,7 @@ module StagingProject
     result.elements('status') do |status|
       code = status.get('code')
 
-      if code.in?(['broken', 'failed']) || (code == 'unresolvable' && !building)
+      if @broken_packages.none? { |p| p[:package] == status['package'] } && broken_code?(code, building)
         @broken_packages << { package: status['package'],
                               project: name,
                               state: code,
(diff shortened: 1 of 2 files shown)


src/api/app/models/concerns/staging_project.rb
```diff
@@ -197,13 +197,15 @@ def add_broken_packages(result, building)
...
       code = status.get('code')
 
-      if code.in?(['broken', 'failed']) || (code == 'unresolvable' && !building)
-        @broken_packages << { package: status['package'],
-                              project: name,
-                              state: code,
-                              details: status['details'],
-                              repository: result['repository'],
-                              arch: result['arch'] }
+      unless @broken_packages.map { |p| p[:package] }.include?(status['package'])
```
coolo: Instead of always mapping the array into another array, how about 

```suggestion
      unless @broken_packages.none? { |p| p[:package] == status['package'] }
```

saraycp: @coolo much better, thanks!

PR #8523, opened 2019-10-07
Let editor insert spaces instead of tabs

### LINKED ISSUE #8504: Mixed tabs and spaces are growing in OBS
# Issue Description
Many *.spec and *.changes files contains mixed tabs and spaces. I found when editing files on OBS website, Tab key always insert tabs. But usually we don't use tabs in packaging.

# How to Reproduce
1. Login to OBS.
2. Open any text file in OBS web interface.
3. Press the tab key.

# Expected Result
Pressing Tab key insert spaces, maybe 2 spaces or 4 spaces.

# Actual Result
Pressing Tab key insert tabs.

# Further Information
See https://build.opensuse.org/package/view_file/games:tools/steam/steam.spec?expand=1
image

guoyunhe:
This PR changes the default CodeMirror editor behavior.

Before: when pressing Tab key, insert a tab indention.

After: when pressing Tab key, insert two spaces indention.

It should fix #8504

### DIFF
--- a/src/api/app/assets/javascripts/webui/cm2/use-codemirror.js
+++ b/src/api/app/assets/javascripts/webui/cm2/use-codemirror.js
@@ -23,14 +23,27 @@ function use_codemirror(id, read_only, mode, big_editor) {
     mode: mode,
     theme: "bootstrap"
   };
+
   if (read_only) {
     codeMirrorOptions['readOnly'] = true;
   }
   else {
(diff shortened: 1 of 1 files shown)


saraycp:
@guoyunhe thanks for your contribution!

In fact, with your proposal it looks much better even when removing tabs. See before:

tab_on_editor

and after:

tab_with_spaces

PR #8536, opened 2019-10-07
Show build logs for finished builds in monitor page

### LINKED ISSUE #8503: Cannot click on 'finished' state in a project monitor
When you are on a project monitor, you can click on a number of package states: 'building', 'succeeded' and 'failed', but you cannot on 'finished'. 
Whereas it could use the same link as 'building', etc.

eduardoj:
Include 'finished' into the states of a package in the monitor page that link to the build logs.

Fixes #8503.


eduardoj:
After having a look in master at the same function `getlogfile`, also `signing` state can be added (thanks @coolo). Please have a look again.

PR #8538, opened 2019-10-08
Refactored ProjectController#status and dependencies to Projects::Status::Controller.

vpereira:
To reduce the complexity of the `ProjectController`, I moved the `status` action and dependencies to it's own controller. Now `status` is a resource under `projects`.

- Controllers, specs and views were adapted accordingly.


vpereira:
to test it: go to `/projects/:project_name/status` and interact with this view.

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -336,7 +336,7 @@ def clear_failed_comment
     flash.now[:success] = 'Cleared comments for packages'
```
coolo: clear failed comment is about the same view, so it would make sense to move as well

vpereira: yes, however not in this PR. But it's in my TODO list too, thanks @coolo

src/api/app/views/webui/projects/status/show.html.haml
```diff
@@ -76,7 +76,7 @@
                     = summary
                     %br
                 %td{ data: { package_name: package['name'] } }
-                  = render partial: 'status_comment', locals: { package_name: package['name'], editable: package['firstfail'].present?,
+                  = render partial: 'webui/project/status_comment', locals: { package_name: package['name'], editable: package['firstfail'].present?,
```
DavidKang: That partial is used in other places? If not, we could move it to the new directory

vpereira:
> Testing locally, when a build fails and I go to Status tab and click on the Filter button, I get an error on app/views/webui/projects/status/show.html.haml:79.
> 
> It can't find the partial `_status_comment`.

could you review it again @saraycp ?

src/api/app/views/webui/projects/status/show.html.haml
vpereira: yes, its used in another view under project. So I would keep it linked instead of have a copy of that for now.

Issue #8516, opened 2019-10-06
Build results: "unresolvable", "broken" and "failed" links green on hover


cboltz:
# Issue Description

In the build results (both the box on the right side of a package page, and on the monitor page), the links for "failed", "broken" and "unresolvable" are displayed as red text. However, when you hover these links, they become green (and underlined). Given that the red color indicates a failure, the links should not become green on hover.

Expected Result
============
The link text should stay red (and get underlined) if you hover them.

How to Reproduce
===============
1. Go to https://build.opensuse.org/package/show/home:cboltz/diffuse (or any package page with a build failure)
2. In the "build results" box, over your mouse over the "unresolvable" links
3. See the link becoming green

You'll see the same behaviour on https://build.opensuse.org/project/monitor/home:cboltz (I assume it's using the same CSS, therefore reporting it in the same bugreport.)

AshNaz87:
I have completed the fix, but don't have access rights

hennevogel:
hey @AshNaz87 you don't have access rights for what exactly? See our documentation on [how to contribute code](https://github.com/openSUSE/open-build-service/blob/master/CONTRIBUTING.md#how-to-contribute-code) :smile: Looking forward to your pull request!

dmarcoux:
Fixed in #9132

PR #8547, opened 2019-10-08
Fix issue #8516. Build state failed link should remain the same color…

AshNaz87:
… on hover

### DIFF
--- a/src/api/app/assets/stylesheets/webui/build-results.scss
+++ b/src/api/app/assets/stylesheets/webui/build-results.scss
@@ -42,6 +42,10 @@
 
 .build-state-unresolvable, .build-state-broken, .build-state-failed {
   color: $red;
+
+  &:hover {
+    color: $red;
+  }
 }
 
 .build-state-disabled {


DavidKang:
Hi @AshNaz87, thanks for your contribution :smile_cat:.

Just a few things, I think your branch is not updated with master. That the reason why the linters are complaining. Regarding the title of the commit, please doesn't add the issue number, instead of that you could add it at the end of the commit description like `fix #XXXX`. And if is possible when you made visual changes, it will be good if you provide some screenshots.

coolo:
I don't think the index has anything to do with the hover

saraycp:
> I don't think the index has anything to do with the hover

True! we wrote the wrong issue number. :+1:

DavidKang:
@AshNaz87, do you need help? :smile_cat:

AshNaz87:
@DavidKang Are the steps straightforward in terms of setting up?

Thing the task itself is not so bad

hennevogel:
move to #9132 

sorry this took so long :disappointed:

PR #8548, opened 2019-10-09
Move log writing operation to a background job

vpereira:
Move log writing operation to a background job.

Non-critical operations, like log writing, shouldn't influence in the user experience.

### DIFF
--- a/src/api/app/controllers/webui/staging/projects_controller.rb
+++ b/src/api/app/controllers/webui/staging/projects_controller.rb
@@ -23,8 +23,7 @@ module Webui
 
         if staging_project.valid? && staging_project.store
           flash[:success] = "Staging project with name = \"#{staging_project}\" was successfully created"
-          staging_project.create_project_log_entry(User.session!)
-
+          CreateProjectLogEntryJob.perform_later(project_log_entry_payload(staging_project), staging_project.created_at.to_s, staging_project.class.name)
           return
         end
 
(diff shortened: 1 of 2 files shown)


vpereira:
[Create ProjectLogEntry records in a job](https://trello.com/c/zZnrnVmb/1295-create-projectlogentry-records-in-a-job)

coolo:
So you create a DB entry to avoid creating a DB entry?

vpereira:
> So you create a DB entry to avoid creating a DB entry?

argh, `perform_now` is what I wanted.

src/api/app/controllers/webui/staging/projects_controller.rb
```diff
@@ -79,7 +78,7 @@ def preview_copy
       def copy
         authorize @staging_workflow
 
-        StagingProjectCopyJob.perform_later(@staging_workflow.project.name, params[:staging_project_project_name], params[:staging_project_copy_name], User.session!.id)
+        StagingProjectCopyJob.perform_now(@staging_workflow.project.name, params[:staging_project_project_name], params[:staging_project_copy_name], User.session!.id)
```
hennevogel: That makes little sense, why do you change this?

vpereira: yes I was thinking about it too. Actually I wanted to go back to `perform_later`, but I've got sidetracked by another issue. `perform_later` is less less intrusive. The logging has low priority and shouldn't block other jobs to execute..

PR #8552, opened 2019-10-09
Add weekly-digest configuration

hennevogel:
We enabled https://github.com/apps/weekly-digest for our repo, let's configure it.


hennevogel:
We discuss in #8546 still...

PR #8558, opened 2019-10-10
Switch to Thinking::Sphinx real-time indexes

### LINKED ISSUE #8541: Searching for a recent created project doesn't return any match
# Issue Description
Search indexes seems not to be updated just after every object (project, package, ...) has been created. This results in no matches searching by their names, just after they have been created.

Now it is needed to wait for the reindexing proccess, that occurs each hour, for that project to be found in the search.

Expected Result
============
Searching for the project that has been just created, returns it as match, without waiting so much as for the next hourly reindex.

How to Reproduce
===============
1. Got to https://build.opensuse.org/
1. Create a project.
1. Search by the name of that project in the search box placed in the upper right corner.
1. See no matches.

Further Information
===============
* Thinking Sphinx: https://freelancing-gods.com/thinking-sphinx/
* Definition of our indexes: https://github.com/openSUSE/open-build-service/tree/master/src/api/app/indices/
* Enabled matches_regex for MySql by jamespearson · Pull Request #36800: https://github.com/rails/rails/pull/36800

eduardoj:
Instead of reindexing every hour with SQL-backed indexes, we now use real-time indices, that allow us to search for a just created or updated project or package.

Real-time indices are updated with callbacks, so we don't need an external process for updating indices.

Fixes #8547.

After this is fully tested we should also remove some remainig code, like the `FullTextIndexJob` job, the entry to reindex in the `config/clock.rb` file, the `obs-delayedjob-queue-sphinx-indexing.service` systemd service, ...

For a detailed explanation of what has been done, please read carefully the commits' descriptions.


eduardoj:
Marking as "DO NOT MERGE" because both tests suites must be adapted.

src/api/Procfile
```diff
@@ -1,5 +1,5 @@
 web: bundle exec rails server -b 0.0.0.0
 delayed: bundle exec script/delayed_job.api.rb run
-clock: bundle exec clockworkd --log-dir=log -l -c config/clock.rb run 
-search: bundle exec rake ts:rebuild NODETACH=true
+clock: bundle exec clockworkd --log-dir=log -l -c config/clock.rb run
+search: (sleep 60 && bundle exec rake ts:rt:index) & bundle exec rake ts:rebuild NODETACH=true && fg
```
hennevogel: `ts:rt:index` is not taken care of by `ts:rebuild`?

saraycp: The normal sequence for `rebuild` is:
```
task :rebuild => [
    :stop, :clear, :configure, 'ts:sql:index', :start, 'ts:rt:index'
]
```

However, running just `ts:rebuild` causes errors at this point. Only if we add `NODETACH=true` it works. And adding that makes the sequence stop in `start`, it never runs  `ts:rt:index`.

eduardoj:
We have updated the pull request, and now is ready to review. Please have a look at the updated descriptions of the pull request and commits.

To test the functionality you can use the review-app. Create a package or a project. After that you will find them in the results of a search.

Removing the "DO NOT MERGE" label.

src/api/app/models/package.rb
```diff
@@ -1457,6 +1460,28 @@ def last_build_reason(repo, arch, package_name = nil)
     PackageBuildReason.new(data)
   end
 
+  # Methods for Thinking Sphinx real-time indexes
```
hennevogel: Can you isolate them in some file please instead of working around the ClassLength cop.

src/api/app/models/package.rb
```diff
@@ -1457,6 +1460,28 @@ def last_build_reason(repo, arch, package_name = nil)
...
+  def package_issues_ids
+    package_issues.pluck(:issue_id)
+  end
+
+  def devel_packages?
+    develpackages.exists?
+  end
+
+  def backend_packages?
+    Package.find_by_sql("SELECT * FROM backend_packages WHERE package_id = #{id}").present?
```
hennevogel: This is a [cache](https://github.com/openSUSE/open-build-service/wiki/The-BackendPackage-Cache), you know this right?

src/api/app/models/project.rb
```diff
@@ -1559,6 +1561,36 @@ def self.very_important_projects_with_categories
     end
   end
 
+  # Methods for Thinking Sphinx real-time indexes
```
hennevogel: Same here...

src/api/spec/features/webui/search_spec.rb
```diff
@@ -1,223 +1,226 @@
-require 'sphinx_helper'
+require 'browser_helper'
```
hennevogel: But why? It was nicely isolated in that helper. Now you have `thinking_sphinx_test_init` all over the place.

src/api/Procfile
hennevogel: :thought_balloon:  I really don't like this....

src/api/spec/features/webui/search_spec.rb
eduardoj: They way RSpec helper methods are isolated is [documented](https://relishapp.com/rspec/rspec-core/docs/helper-methods/define-helper-methods-in-a-module#include-a-module-in-only-some-example-groups).

I have already checked, and calling `thinking_sphinx_test_init` in a test that doesn't have `thinking_sphinx: true` results in an `undefined local variable or method 'thinking_sphinx_test_init'` error.

In other words, `thinking_sphinx_test_init` is only available in tests with `thinking_sphinx: true`.

src/api/Procfile
eduardoj: I don't like it either. Any other alternative?

ReleaseNotes-2.11
```diff
@@ -29,6 +29,8 @@ Features
 Generic:
  * You can track javascript errors that occur in the users browser with errbit, enable
    it in the API configuration file options.yml.
+ * Search was improved: searching for a just created or updated project or package will
```
dmarcoux: I suggest this:

`Search was improved: New and updated projects/packages are indexed directly instead of once every hour.`

eduardoj: Thanks. Done.

src/api/app/models/package.rb
eduardoj: Done. Now they are isolated in `app/mixins/package_sphinx.rb`.

eduardoj: However adding the two `include` calls made `app/models/package.rb` bigger, so I changed the `Max` limit of the `Metrics/ClassLength` cop.

src/api/Procfile
eduardoj: In production this `ts:rt:index` job is run after the `:start` job because `NODETACH` is NOT set to `true`. In production we are running it in detached mode.

src/api/app/models/package.rb
eduardoj: I changed this line to `backend_package.present?`, making use of the `has_one` relationship.

Being a cache, do you see it could be harmfull to use this query using real-time indices?

src/api/Procfile
hennevogel: Okay, then how about you create an `thinking_sphinx` issue and link it here?

src/api/app/models/package.rb
hennevogel: Well potentially, if the case doesn't exist during your `backend_package.present?` call, it will be [created](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/package.rb#L1107).

src/api/spec/features/webui/search_spec.rb
eduardoj: The code to init and stop `ThinkingSphinx::Test` was moved from the tests to the `spec/support/thinking_sphinx.rb` file, using [hooks filtered by symbols](https://relishapp.com/rspec/rspec-core/v/3-9/docs/hooks/filters#filtering-hooks-using-symbols).

src/api/Procfile
eduardoj: Done: https://github.com/pat/thinking-sphinx/issues/1150

coolo:
I don't think it's fixing #8547 - so what's the actual issue fixed? your branch is named fix_8541 - that might give a hint :)

hennevogel:
#8541 was the issue fixed.

saraycp:
I've just updated the description, we wrote the wrong number.

coolo:
Note that with this a running searchd is now a requirement. If you create/update projects and searchd doesn't run, you get a 500 now - it wasn't a problem before. Struggled with that in the CI of openSUSE-release-tools

hennevogel:
> If you create/update projects and searchd doesn't run, you get a 500 now - it wasn't a problem before. 

This was always a problem.

coolo:
It was most likely a problem to find the projects afterwards through the webui, but for our CI there was no requirement on searchd in the API before.

PR #8566, opened 2019-10-11
Set review for staged requests if request state new or review

### LINKED ISSUE #8522: Unstaging revoked requests gives 403
Testing #8507 on build-test I noticed that unstaging a request not in `review` gives a misleading response. It returns a 403, but funnily enough still unstage it from the staging project. 

```
coolo@f102#~>tosc request revoke 1760 -m oops
Result of change request state: ok
coolo@f102#~>tosc ps request delete openSUSE:Factory -s openSUSE:Factory:Staging:A -r 1760
Server returned an error: HTTP Error 403: Forbidden
The request is not in state new or review
coolo@f102#~>tosc -d ps request delete openSUSE:Factory -s openSUSE:Factory:Staging:A -r 1760
DELETE https://api-test.opensuse.org/staging/openSUSE:Factory/staging_projects/openSUSE:Factory:Staging:A/staged_requests
```

vpereira:
Fixes #8522 

Just update the reviews in a request if the request state is `new` or `review`

Add a spec to test to cover the case where request state is `revoked`

### DIFF
--- a/src/api/app/models/staging/staged_requests.rb
+++ b/src/api/app/models/staging/staged_requests.rb
@@ -29,7 +29,7 @@ class Staging::StagedRequests
     not_deleted_packages = package_names - result.pluck(:name)
 
     requests.each do |request|
-      add_review_for_unstaged_request(request)
+      add_review_for_unstaged_request(request) if request.state.in?([:new, :review])
 
       ProjectLogEntry.create!(
         project: staging_project,
--- a/src/api/spec/controllers/staging/staged_requests_controller_spec.rb
+++ b/src/api/spec/controllers/staging/staged_requests_controller_spec.rb
@@ -155,6 +155,20 @@ RSpec.describe Staging::StagedRequestsController do
         it { expect(staging_project.packages).to be_empty }
         it { expect(staging_project.staged_requests).to be_empty }
       end
+
+      context 'with revoked request' do
+        before do
+          login user
+          bs_request.state = :revoked
+          bs_request.save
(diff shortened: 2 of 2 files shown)


src/api/app/models/staging/staged_requests.rb
```diff
@@ -29,7 +29,7 @@ def destroy
     not_deleted_packages = package_names - result.pluck(:name)
 
     requests.each do |request|
-      add_review_for_unstaged_request(request)
+      add_review_for_unstaged_request(request) if [:review, :new].include?(request.state.to_sym)
```
coolo: there is tons of code like `req.state.in?([:review, :new])` and `@bs_request.state.in?([:new, :review])` so I don't think you need .to_sym, but should prefer `.in?` for consistency.

And btw, I found this gem:
```
app/models/bs_request.rb:    unless state == :review || (state == :new && state == :new)
```

vpereira: > there is tons of code like `req.state.in?([:review, :new])` and `@bs_request.state.in?([:new, :review])` so I don't think you need .to_sym, but should prefer `.in?` for consistency.

Yup, i will do it like that, thanks!

> And btw, I found this gem:
> 
> ```
> app/models/bs_request.rb:    unless state == :review || (state == :new && state == :new)
> ```
:pray: TGIF :pray:

vpereira: done @coolo

PR #8576, opened 2019-10-14
Add configuration breadcrumbs for User management

vpereira:
Before:
Screenshot_2019-10-14_13-35-42

After:
Screenshot_2019-10-14_13-48-32

### DIFF
--- a/src/api/app/views/webui/users/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui/users/_breadcrumb_items.html.haml
@@ -1,17 +1,18 @@
-= render partial: 'webui/main/breadcrumb_items'
 - case action_name
   - when 'index'
+    = render partial: 'webui/configuration/breadcrumb_items'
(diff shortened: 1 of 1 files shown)


src/api/app/views/webui/users/_breadcrumb_items.html.haml
```diff
@@ -1,18 +1,17 @@
-= render partial: 'webui/main/breadcrumb_items'
 - case action_name
   - when 'index'
+    = render partial: 'webui/configuration/breadcrumb_items'
     %li.breadcrumb-item.active{ 'aria-current' => 'page' }
       Manage Users
   - when 'new'
+    = render partial: (User.session ? 'webui/configuration/breadcrumb_items' : 'webui/main/breadcrumb_items')
```
dmarcoux: Is the ternary operator needed? I believe it should always be `'webui/configuration/breadcrumb_items'`.

PR #8579, opened 2019-10-15
Align usage of icons for projects/packages in the UI

### LINKED ISSUE #8480: reviews by_project don't have an icon
https://build.opensuse.org/staging_workflows/136 has zaz selected into N (thanks to @eduardoj for the fix) but the request has a review by_package and is as such orange, but you can't see why.

dmarcoux:
To follow what we already do in the 'Latest Updates' section on the homepage.

Fixes #8480

On the staging workflow page, I couldn't use `text-warning`, so I went with `text-dark` instead.
staging-workflow-after
On the request page:
request-after

As a reference, this is what we already do in the `Latest Updates` section:
latest-updates

### DIFF
--- a/src/api/app/helpers/webui/staging/workflow_helper.rb
+++ b/src/api/app/helpers/webui/staging/workflow_helper.rb
@@ -43,19 +43,22 @@ module Webui::Staging::WorkflowHelper
     end
   end
 
-  def reviewers_gravatar(request, users_hash, groups_hash)
+  def reviewers_icon(request, users_hash, groups_hash)
     missing_reviews = request[:missing_reviews]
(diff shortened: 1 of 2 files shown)


coolo:
I wouldn't necessarly call https://user-images.githubusercontent.com/1102934/66823195-07419b00-ef46-11e9-9abf-450900bd5c10.png 'aligned'. 

Looking at https://build.opensuse.org/staging_workflows/136 I wonder: gibt's die auch in schön?

dmarcoux:
@coolo: Instead of being snide, I wish you would simply phrase your comment**s** with a constructive/friendly tone.

PR #8582, opened 2019-10-16
Staging workflow link instead of branch packages

### LINKED ISSUE #7343: Staging a request branches the package
Staging a request shouldn't branch the package, but create a _link with a fixed revision. Updating a branch with a different revision is much harder and more noisy. You basically need to delete the complete source - while with _links you simply exchange one file.

vpereira:
Fixes #7343 

The code is working, however it probably belongs to somewhere else (maybe a service object, maybe in the `Package` class. The refactoring could be done later.


src/api/app/models/staging/staged_requests.rb
```diff
@@ -84,6 +85,34 @@ def requests
...
   end
 
+  def link_package(bs_request_action)
+    request = bs_request_action.bs_request
+
+    source_package = Package.get_by_project_and_name!(bs_request_action.source_project,
+                                                      bs_request_action.source_package)
+
+    # get the base revision from the source package
+    baserev = source_package.backend_package.srcmd5
```
coolo: This is not the submitted revision but HEAD

src/api/app/models/staging/staged_requests.rb
```diff
@@ -84,6 +85,34 @@ def requests
...
+
+    # create the package in the staging project
+    link_package = Package.create!(project: staging_project,
+                                   name: bs_request_action.source_package)
```
coolo: Does `create!` fail if it already exists? Because that can happen on supersedes

src/api/app/models/staging/staged_requests.rb
```diff
@@ -84,6 +85,34 @@ def requests
...
+
+    # add link file to the project, linking to the source project baserev
+    link_package.save_file(filename: '_link', file: "<link project=\"#{source_package.project.name}\" baserev=\"#{baserev}\"></link>")
```
coolo: what does the baserev to do with this?

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
```diff
@@ -69,7 +69,7 @@
       end
 
       it { expect(response).to have_http_status(:bad_request) }
-      it { expect(staging_project.packages.pluck(:name)).to match_array([target_package.name]) }
+      it { expect(staging_project.packages.pluck(:name)).to match_array([source_package.name]) }
```
coolo: No, we indeed expect the target_package.name here

src/api/app/models/staging/staged_requests.rb
vpereira: I was just mimicking what I saw in the `_link` file and it worked, package can be built. Any suggestion?

src/api/app/models/staging/staged_requests.rb
vpereira: hm supersedes isn't an issue, if you unstage the old issue and stage the new one? Both steps could be packed together in the `osc ps` command. In another hand would be ok to keep the package if it exists and just update the `_link` file?

coolo: IMO a supersede is a single action (especially as long as you don't stop the scheduler from looking at the project after that unstage) - and beging able to update the _link file is main reason why branch is so harmful

src/api/app/models/staging/staged_requests.rb
coolo: In what _link file btw?

src/api/app/models/staging/staged_requests.rb
hennevogel: > would be ok to keep the package if it exists and just update the _link file?

Yes, that is the idea.

src/api/app/models/staging/staged_requests.rb
vpereira: fixed, could you look into it again @coolo ?

src/api/app/models/staging/staged_requests.rb
```diff
@@ -84,6 +85,35 @@ def requests
...
+    # get the revision from the source package
+    source_rev = bs_request_action.source_rev
+
+    # find or create the package in the staging project
+    link_package = Package.find_or_create_by!(project: staging_project, name: bs_request_action.target_package)
+
+    # add or overwrite "_link", linking to the source project baserev
+    Backend::Api::Sources::Package.write_link(staging_project.name,
+                                              link_package.name,
+                                              User.session!, "<link project=\"#{source_package.project.name}\" rev=\"#{source_rev}\"></link>")
```
coolo: You forgot the package here and we need to fixate the `vrev` - because the `source_rev` needs to be expanded `srcmd5` and quite some packages won't work if the start with revision 1.

Just as in the code you're replicating: https://github.com/openSUSE/openSUSE-release-tools/blob/master/osclib/stagingapi.py#L1132

src/api/app/models/staging/staged_requests.rb
```diff
@@ -84,6 +85,44 @@ def requests
...
+  def link_package(bs_request_action)
+    request = bs_request_action.bs_request
+
+    source_package = Package.get_by_project_and_name!(bs_request_action.source_project,
+                                                      bs_request_action.source_package)
+
+    # we need the revision and md5 and make it sticky in the _link
+    # packages in staging a locked in a specific version
+    backend_package_information = Directory.hashed(project: source_package.project.name,
+                                                   package: source_package.name, expand: 1)
```
coolo: we're getting closer. But let me repeat what I wrote in chat: you forgot the (optional) revision

coolo: And I'd suggest you just use `source_package.dir_hash`

src/api/app/models/staging/staged_requests.rb
```diff
@@ -84,6 +85,46 @@ def requests
...
+
+    # we need the revision and md5 and make it sticky in the _link
+    # packages in staging a locked in a specific version
+    query_options = { expand: 1 }
+    query_options[:rev] = bs_request_action.source_rev if bs_request_action.source_rev
+
+    backend_package_information = source_package.dir_hash(query_options)
+
+    # get the revision from the source package
+    source_rev = backend_package_information['vrev']
```
coolo: Whatever the vrev is (I don't think there is any documentation about it :), it's not the source revision :)

vpereira: done

vpereira: vrev is the "version" revision, it is also incremented as rev, but it gets reset if the version of the package changes

vpereira: done

src/api/app/models/staging/staged_requests.rb
```diff
@@ -84,6 +85,44 @@ def requests
...
   end
 
+  def link_package(bs_request_action)
+    request = bs_request_action.bs_request
+
+    source_package = Package.get_by_project_and_name!(bs_request_action.source_project,
+                                                      bs_request_action.source_package)
+
+    # we need the revision and md5 and make it sticky in the _link
+    # packages in staging a locked in a specific version
```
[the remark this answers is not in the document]
vpereira: The comments were important to me to keep track of what I was trying to achieve. Now that it's ready, I removed them.

src/api/spec/cassettes/Staging_StagedRequestsController/POST_create/with_valid_staging_project/1_2_5_4.yml
```diff
@@ -734,106 +25,21 @@ http_interactions:
...
+    uri: http://backend:5352/source/home:permitted_user:Staging:A/target_package/_link?user=permitted_user
     body:
       encoding: UTF-8
-      string: |
-        <workflow project="home:permitted_user" managers="staging-workflow-managers-11">
-          <staging_project name="home:permitted_user:Staging:A"/>
-          <staging_project name="home:permitted_user:Staging:B"/>
-        </workflow>
+      string: <link project="source_project" package="source_package" rev="d41d8cd98f00b204e9800998ecf8427e"
+        vrev=""></link>
```
coolo: you wondered if there it's possible to have empty vrevs, didn't you? Looks like you even got a test case for it

PR #8593, opened 2019-10-20
Fix smoke tests

M0ses:
Without this patch only a part of the string 'openSUSE.org:openSUSE:Tools'
gets passed to the backend, which quits with a 404 from the remote side.

This behavior is non-determenistic.

SEE ALSO:
teamcapybara/capybara#1890

Fix taken from here:

teamcapybara/capybara#1890 (comment)

This patch

* adds a sleep after clicking "Branch existing project" before entering the strings
  to give jQuery the time to complete the animation
* sends the strings char by char to avoid race conditions with autocompletion

### DIFF
--- a/dist/t/spec/features/0040_package_spec.rb
+++ b/dist/t/spec/features/0040_package_spec.rb
@@ -43,6 +43,7 @@ RSpec.describe "Package" do
       click_link('Home Project')
     end
     click_link('Branch Existing Package')
+    page.evaluate_script('$.fx.off = true;') # Needed to disable javascript animations to avoid race conditions
     within('#new-package-branch-modal') do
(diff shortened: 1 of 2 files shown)


coolo:
don't go this route. Disable animations and wait for popups to appear instead

eduardoj:
> don't go this route. Disable animations and wait for popups to appear instead

Yes, something like [this](https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/features/webui/search_spec.rb#L27).

M0ses:
Working on your suggestions - needs to be squashed when done

M0ses:
commits squashed, ready to merge from my POV

PR #8596, opened 2019-10-21
Disable use of concurrency in asset compilation

coolo:
sassc is not thread safe and there are tons of crashes while building packages. Unfortunately #8555 was merged even though the problem was noted there, so applying workaround from the upstream issue.

### DIFF
--- a/src/api/config/application.rb
+++ b/src/api/config/application.rb
@@ -91,6 +91,11 @@ module OBSApi
     # avoid a warning
     I18n.enforce_available_locales = true
 
+    config.assets.configure do |env|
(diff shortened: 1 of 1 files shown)


coolo:
I ran 21 packages with each 12 precompilations and I had 18/21 build failures. With this patch added it's 0/21

vpereira:
nice @coolo!

To be fair before #8555 the probably was already happening, not just so often as now.

hennevogel:
Cool. In #8555 we just mis-attributed this crash to the old sassc issue....

PR #8598, opened 2019-10-22
Refactor BranchPackage model

vpereira:
That's a first iteration on that "model". I want to break it in smaller classes, add specs to that and maybe in the near future merge it with `StagedRequests#link_package`

In This PR:

- Add new specs
- extract the `dryreport` and errors to its own modules
- extract the `set_target_project` method to its own class

### DIFF
--- a/src/api/app/models/branch_package.rb
+++ b/src/api/app/models/branch_package.rb
@@ -533,22 +533,10 @@ class BranchPackage
   end
 
   def set_target_project
-    if params[:target_project]
-      @target_project = params[:target_project]
-      @auto_cleanup = ::Configuration.cleanup_after_days if params[:autocleanup] == 'true'
-    else
-      if params[:request]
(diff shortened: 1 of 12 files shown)


src/api/app/models/branch_package.rb
```diff
@@ -539,23 +530,7 @@ def lookup_incident_pkg(p)
...
-          end
-        else
-          builder.package(project: p[:link_target_project], package: p[:package]) do
-            builder.devel(project: p[:copy_from_devel].project.name, package: p[:copy_from_devel].name) if p[:copy_from_devel]
-            builder.target(project: @target_project, package: p[:target_package])
-          end
-        end
-      end
-    end
+    BranchPackage::DryRun::Report.new(@packages, @target_project).to_xml
```
danidoni: :+1:

PR #8622, opened 2019-10-28
Refactor commit_message to generate_file_diff

### LINKED ISSUE #8306: Refactor instance method 'commit_message' in the Package model
The instance method `commit_message` in the `Package` model has a misleading name. It has nothing to do with commit messages. It's parsing _changes_ files. Renaming the method should be enough to remove the confusion here.

ngsikai:
Hello!

Performed the refactoring as suggested in #8306, changing method name of `commit_message` to `generate_file_diff` as it is closer semantically to what that method is doing.

To verify this feature

- Run the regression spec
Closes #8306


vpereira:
Hey @ngsikai thank you for your submission. 

If you see, in our CI you have some rspec errors, which probably are green for you locally. That's because we use VCR to replay HTTP traffic between frontend and backend. We have some information [here](https://github.com/openSUSE/open-build-service/wiki/Testing-with-VCR), please check it and see if you are able to generate the VCR for the new test case that you created

ngsikai:
Hey @vpereira ,

Thanks for sharing on how to fix the spec errors. I am unfortunately having problems starting up the backend service to capture the new VCR response. 

When I run `docker-compose up backend` locally,

I am faced with an error like this:

Screen Shot 2019-10-29 at 2 42 21 PM

Would like to get some advice on what I am doing wrong/ what I should be doing to start it up successfully. Did a submodule init and submodule update within `src/backend` to pull in the build submodule but that raised different errors as well.

dmarcoux:
Hey @ngsikai,

After you did the submodule init, what were the errors? We'll do our best to help you out on this one.

hennevogel:
@ngsikai are you by any chance on macos? If yes you need a [case sensitive filesystem](https://github.com/openSUSE/open-build-service/wiki/Setup-an-OBS-Development-Environment-on-macOS).

ngsikai:
@dmarcoux @hennevogel I am indeed on macos and do not have a case sensitive filesystem. I give the linked solution a shot then 👍 Thanks for the suggestion 🙏

Issue #8623, opened 2019-10-28
Can't create projects on build-test


coolo:
I wanted to get back to test staging workflow, but I'm blocked by this issue - is there some migration missing?

```
I, [2019-10-28T10:08:05.243033 #18557]  INFO -- : [f3ab4478-f70c-4216-af3c-76882e0ec30b] [18557:99349.01] method=PUT path=/source/openSUSE:Factory:Staging:adi:217/_meta format=xml controller=SourceProjectMetaController action=update status=500 error='ThinkingSphinx::SphinxError: index 'project_core' does not support INSERT (enabled=1) - REPLACE INTO project_core (id, `name`, `title`, `description`, `sphinx_internal_id`, `sphinx_internal_class`, `sphinx_deleted`, `project_id`, `attrib_type_ids`, `issue_ids`, `linked_count`, `activity_index`, `links_to_other`, `is_devel`, `updated_at`) VALUES (1629, 'openSUSE:Factory:Staging:adi:217', '', '', 407, 'Project', 0, 407, (), (), 0, 0, 0, 0, 0)' duration=142.14 view=0.00 db=77.51 params={"project"=>"openSUSE:Factory:Staging:adi:217"} host=149.44.161.60 time=2807238.56 backend=0 user=coolo
F, [2019-10-28T10:08:05.249137 #18557] FATAL -- : [f3ab4478-f70c-4216-af3c-76882e0ec30b] [18557:99349.02]   
F, [2019-10-28T10:08:05.249213 #18557] FATAL -- : [f3ab4478-f70c-4216-af3c-76882e0ec30b] [18557:99349.02] ThinkingSphinx::SphinxError (index 'project_core' does not support INSERT (enabled=1) - REPLACE INTO project_core (id, `name`, `title`, `description`, `sphinx_internal_id`, `sphinx_internal_class`, `sphinx_deleted`, `project_id`, `attrib_type_ids`, `issue_ids`, `linked_count`, `activity_index`, `links_to_other`, `is_devel`, `updated_at`) VALUES (1629, 'openSUSE:Factory:Staging:adi:217', '', '', 407, 'Project', 0, 407, (), (), 0, 0, 0, 0, 0)):
F, [2019-10-28T10:08:05.249249 #18557] FATAL -- : [f3ab4478-f70c-4216-af3c-76882e0ec30b] [18557:99349.02]   
[...]

PR #8627, opened 2019-10-28
Rebuild thinking sphinx indexes through a data migration

krauselukas:
After switching from SQL-backed indexes to real-time indexes
running 'ts:rebuild' is needed in order to index the already
existing packages and projects.

Related to https://github.com/openSUSE/open-build-service/issues/8623


vpereira:
I think we need to document it somehow. As I understood, today if you update your local OBS instance you will have to run the `rake ts:rebuild`. If you are deploying a new instance you won't need it. So I'm not sure if any migration is needed other than document it.

saraycp:
We discussed during the stand up that it might not make sense to run a data migration for instances created from scratch. We have to find another way to run `ts:rebuild`automatically but only once (the first time real-time indexes are deployed).

coolo:
why would any migration run on instances created from scratch? I don't get it

PR #8635, opened 2019-10-29
Refactor look_incident_package method to its own class

vpereira:
Refactoring the `BranchPackage` "method"

I extracted one more method in a class, beside it:

- Migrate usage of REXML to Nokogiri 
- Extract and Refactor methods
- Add spec for the new class
- Adapt branch_package

### DIFF
--- a/src/api/app/models/branch_package.rb
+++ b/src/api/app/models/branch_package.rb
@@ -500,6 +500,7 @@ class BranchPackage
 
   def lookup_incident_pkg(p)
     return unless p[:package].is_a?(Package)
+    return if p[:link_target_project].maintenance_projects.empty?
     BranchPackage::LookupIncidentPackage.new(p).package
   end
 
--- a/src/api/app/models/branch_package.rb
+++ b/src/api/app/models/branch_package.rb
@@ -500,32 +500,7 @@ class BranchPackage
 
   def lookup_incident_pkg(p)
     return unless p[:package].is_a?(Package)
-    @obs_maintenanceproject ||= AttribType.find_by_namespace_and_name!('OBS', 'MaintenanceProject')
-    @maintenance_projects ||= Project.find_by_attribute_type(@obs_maintenanceproject)
-    incident_pkg = nil
-    p[:link_target_project].maintenance_projects.each do |mp|
-      # no defined devel area or no package inside, but we branch from a release are: check in open incidents
-
-      # only approved maintenance projects
-      next unless @maintenance_projects.include?(mp.maintenance_project)
-
-      answer = Backend::Api::Search.incident_packages(p[:link_target_project].name, p[:package].name, mp.maintenance_project.name)
-      data = REXML::Document.new(answer)
-      data.elements.each('collection/package') do |e|
-        ipkg = Package.find_by_project_and_name(e.attributes['project'], e.attributes['name'])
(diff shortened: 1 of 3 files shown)


src/api/app/models/branch_package/lookup_incident_package.rb
```diff
@@ -0,0 +1,53 @@
...
+      # only approved maintenance projects
+      next unless maintenance_projects.include?(mp.maintenance_project)
+      # extract possible packages
+      possible_packages(mp.maintenance_project)
+    end.flatten
+    pkg = nil
+    # choose the last one based on the project id in the name openSUSE:Maintenance:1234
+    # can we use the MaintenanceIncident#incident_id ?
+    possible_packages.each do |possible_package|
+      pkg = possible_package if pkg.nil? || possible_package.project.name.gsub(/.*:/, '').to_i > pkg.project.name.gsub(/.*:/, '').to_i
```
vpereira: maybe here instead of gsub of the name, check the `incident_id` ? What do you think @coolo ?

src/api/app/models/branch_package/lookup_incident_package.rb
```diff
@@ -0,0 +1,53 @@
...
+    incident_packages = Backend::Api::Search.incident_packages(@link_target_project.name, @package.name,
+                                                               maintenance_project.name)
+    Nokogiri::XML(incident_packages)
+  end
+
+  private
+
+  # TODO: there is not a better way to find it?
+  def incident?(package)
+    package.project.is_maintenance_incident? && package.project.is_unreleased?
```
vpereira: There is a better way to do it? Maybe delegate it?

src/api/app/models/branch_package/lookup_incident_package.rb
```diff
@@ -0,0 +1,52 @@
...
+  def package
+    possible_packages = @link_target_project.maintenance_projects.collect do |mp|
+      # only approved maintenance projects
+      next unless maintenance_projects.include?(mp.maintenance_project)
+      # extract possible packages
+      possible_packages(mp.maintenance_project)
+    end.flatten
+    pkg = nil
+    # choose the last one based on the incident number (incremental sequence)
+    possible_packages.each do |possible_package|
```
danidoni: I would use `possible_packages.find { |p| ... }` here

src/api/app/models/branch_package.rb
```diff
@@ -500,32 +500,8 @@ def find_packages_to_branch
 
   def lookup_incident_pkg(p)
```
danidoni: :thought_balloon: I would try to use something a bit more descriptive as variable name

vpereira: Yes, I can do it, to change this method wasn't my goal. Specially that looks like `p` is the thing in this whole `BranchPackage` class

src/api/app/models/branch_package/lookup_incident_package.rb
vpereira: why?

vpereira: is that a change request or an idea?

danidoni: Nevermind, I got it wrong

PR #8636, opened 2019-10-30
Add filter for spec and test directory

vpereira:
We should exclude the `test` and `spec` directories from simplecov coverage calculation

We can accomplish it using `add_filter` as explained here https://rubydoc.info/gems/simplecov/SimpleCov/Configuration#add_filter-instance_method

### DIFF
--- a/src/api/spec/support/coverage.rb
+++ b/src/api/spec/support/coverage.rb
@@ -11,5 +11,9 @@ SimpleCov.start 'rails' do
   add_filter '/lib/templates/'
   add_filter '/lib/memory_debugger.rb'
   add_filter '/lib/memory_dumper.rb'
+
+  # filter spec directory not in sync with test/test_helper
+  add_filter 'spec'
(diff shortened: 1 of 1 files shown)


src/api/spec/support/coverage.rb
```diff
@@ -11,5 +11,9 @@
   add_filter '/lib/templates/'
   add_filter '/lib/memory_debugger.rb'
   add_filter '/lib/memory_dumper.rb'
+
+  # filter spec directory not in sync with test/test_helper
```
hennevogel: I don't understand the comment. What is not in sync?

vpereira: if you open the file, you will see two lines above, the following comment:

"NOTE: Keep filters in sync with test/test_helper.rb"

vpereira:
nothing changed, so I guess those directories are already being black-listed.

PR #8639, opened 2019-10-30
Accept a parameter to remove exclusion from requests before staging

### LINKED ISSUE #8525: Allow to stage an excluded request in one step
If I exclude a request, I do so to ease managing the backlog. But once I have free staging projects, I would like to give it a 2nd chance. To do so right now, I need to:

1. remove from excluded requests

1. stage into project of choice

because if I skip to step 2, I'll get:
```
Assigning requests to openSUSE:Factory:Staging:B failed: Request 114 not found in Staging for project openSUSE:Factory.
```

(which is yet another misleading error, I'd file an issue for this but I prefer if this is solved within the context of this very issue)

I would prefer if I could ACK that it's currently excluded (to avoid races between 2 staging managers, where one excludes a request from backlog and another selects it into a project) and have `add --unignore`

danidoni:
Stage request action now accepts the new parameter `remove_exclusion`

Right now, you can't stage an excluded request. You have to:

1. remove the exclusion first
2. stage the request afterwards

This opens a time window where maintainers can compete to stage the
same request.

This PR implements the acceptance of a new parameter during the
request staging to do 1. and 2. in the same action.

Fixes #8525

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -2012,7 +2012,9 @@ GET /staging/<staging_workflow_project>/staging_projects/<staging_project>/stage
 XmlResult: staged_requests
 
 POST /staging/<staging_workflow_project>/staging_projects/<staging_project>/staged_requests
-  Add one or multiple bs_requests to the staged requests of a staging project.
+  Add one or multiple bs_requests to the staged requests of a staging project
+  remove_exclusion: removes the request from exclusion request before staging it.
+
 XmlBody: remove_staged_requests
 XmlResult: status_ok
 
@@ -2023,6 +2025,5 @@ XmlResult: status_ok
 
 == Internal only routes
 
-/public shall not be used in any tools, it is for OBS remote support only and may
-change or disappear at any time. The route is only working when anonymous mode is enabled.
+   /public shall not be used in any tools, it is for OBS remote support only and may change or disappear at any time. The route is only working when anonymous mode is enabled.
 
--- a/src/api/app/controllers/staging/staged_requests_controller.rb
+++ b/src/api/app/controllers/staging/staged_requests_controller.rb
@@ -15,22 +15,21 @@ class Staging::StagedRequestsController < Staging::StagingController
   def create
     authorize @staging_project, :update?
 
-    result = ::Staging::StagedRequests.new(
+    if params[:remove_exclusion]
+      ::Staging::RequestExcluder
+        .new(requests_xml_hash: { number: @request_numbers },
+             staging_workflow: @staging_workflow)
+        .destroy!
+    end
+
+    ::Staging::StagedRequests.new(
       request_numbers: @request_numbers,
       staging_workflow: @staging_workflow,
       staging_project: @staging_project,
       user_login: User.session!.login
-    ).create
+    ).create!
 
-    if result.valid?
-      render_ok
-    else
-      render_error(
-        status: 400,
-        errorcode: 'invalid_request',
-        message: "Assigning requests to #{@staging_project} failed: #{result.errors.to_sentence}."
-      )
-    end
+    render_ok
   end
 
   def destroy
--- a/src/api/app/models/staging/excluded_request_not_found.rb
+++ b/src/api/app/models/staging/excluded_request_not_found.rb
@@ -1,0 +1,3 @@
+class Staging::ExcludedRequestNotFound < APIError
+  setup 'invalid_request'
+end
--- a/src/api/app/models/staging/staged_requests.rb
+++ b/src/api/app/models/staging/staged_requests.rb
@@ -3,6 +3,7 @@ class Staging::StagedRequests
   attr_accessor :request_numbers, :staging_project, :staging_workflow, :user_login
 
   def create
+    add_requests_excluded_errors
     add_request_not_found_errors
     requests.each do |request|
       bs_request_action = request.bs_request_actions.first
@@ -16,6 +17,13 @@ class Staging::StagedRequests
     result.each { |request| add_review_for_staged_request(request) }
 
     self
+  end
+
+  def create!
+    create
+
+    return if valid?
+    raise Staging::ExcludedRequestNotFound, errors.to_sentence
   end
 
   def destroy
@@ -65,8 +73,14 @@ class Staging::StagedRequests
     @not_removed_packages ||= {}
   end
 
+  def add_requests_excluded_errors
+    excluded_request_numbers.map do |request_number|
+      errors << "Request #{request_number} currently excluded from project #{request_target_project}. Use --remove-exclusion if you want to force this action."
+    end
+  end
+
   def add_request_not_found_errors
-    not_found_requests.each do |request_number|
+    not_found_request_numbers.each do |request_number|
       errors << if BsRequest.exists?(number: request_number)
                   "Request #{request_number} not found in Staging for project #{request_target_project}"
                 else
@@ -79,8 +93,12 @@ class Staging::StagedRequests
     staging_workflow.project
   end
 
-  def not_found_requests
+  def not_found_request_numbers
     request_numbers - requests.pluck(:number).map(&:to_s)
+  end
+
+  def excluded_request_numbers
+    staging_workflow.request_exclusions.where(number: @request_numbers).pluck(:number).map(&:to_s)
   end
 
   def requests
--- a/src/api/spec/controllers/staging/staged_requests_controller_spec.rb
+++ b/src/api/spec/controllers/staging/staged_requests_controller_spec.rb
@@ -19,6 +19,7 @@ RSpec.describe Staging::StagedRequestsController do
            target_package: target_package,
            source_package: source_package,
            description: 'BsRequest 1',
+           number: 1,
            review_by_group: group)
   end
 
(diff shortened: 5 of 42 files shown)


src/api/app/controllers/staging/staged_requests_controller.rb
```diff
@@ -15,6 +15,19 @@ def index
...
     authorize @staging_project, :update?
 
+    if params[:remove_exclusion] == '1'
+      result = ::Staging::RequestExcluder.new(requests_xml_hash: { number: @request_numbers },
+                                              staging_workflow: @staging_workflow).destroy
+      unless result.valid?
+        render_error(
+          status: 400,
+          errorcode: 'invalid_request',
+          message: "Error while unexcluding requests: #{result.errors.join(' ')}"
```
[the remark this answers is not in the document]
danidoni: Yeah

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
```diff
@@ -99,6 +99,45 @@
       it { expect(staging_project.staged_requests).to include(bs_request) }
       it { assert_select 'status[code=ok]' }
     end
+
+    context 'with an excluded request', vcr: true do
+      context 'when not providing the remove exclusion parameter' do
+        before do
+          Staging::RequestExclusion.new(bs_request: bs_request,
```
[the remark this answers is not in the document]
danidoni: I don't know. I'll try to build one

src/api/app/controllers/staging/staged_requests_controller.rb
```diff
@@ -15,6 +15,19 @@ def index
   def create
     authorize @staging_project, :update?
 
+    if params[:remove_exclusion] == '1'
```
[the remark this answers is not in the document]
danidoni: Ok, I'll just check for parameter presence

hennevogel:
Listen to codeclimate! ;-)

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
```diff
@@ -99,6 +99,65 @@
...
+    end
+
+    context 'when providing the remove exclusion parameter' do
+      context 'and not having any excluded request' do
+        before do
+          login user
+          post :create,
+               params: { staging_workflow_project: staging_workflow.project.name,
+                         staging_project_name: staging_project.name,
+                         remove_exclusion: true,
```
saraycp: The same as before, `true`or `1`?

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
```diff
@@ -99,6 +99,65 @@
...
+
+        it { expect(response).to have_http_status(:success) }
+        it { expect(staging_project.packages.pluck(:name)).to match_array([target_package.name]) }
+        it { expect(staging_project.staged_requests).to include(bs_request) }
+        it { assert_select 'status[code=ok]' }
+      end
+    end
+
+    context 'when providing the remove exclusion parameter' do
+      context 'and not having any excluded request' do
```
saraycp: What about merging this with the context in 122? We could have something like:

```
context 'when providing the remove exclusion parameter'
  context 'having any excluded request'
  context 'not having any excluded request'
```
And maybe try to share a single before block for both.

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
```diff
@@ -99,6 +99,65 @@
...
+        before do
+          create(:request_exclusion,
+                 bs_request: bs_request,
+                 number: bs_request.number,
+                 staging_workflow: staging_workflow)
+          login user
+          post :create,
+               params: { staging_workflow_project: staging_workflow.project.name,
+                         staging_project_name: staging_project.name,
+                         remove_exclusion: true,
```
saraycp: I wonder if it's better to set '1' instead of 'true' :thinking:  All in all, '1' is the value in the query string.

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
danidoni: @hennevogel and I discussed this yesterday and agreed on not caring what the content of the variable is as long as its "truthy", to be consistent with the rest of the API.

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
danidoni: I'll look at it

danidoni: Ah, I remember now. It will be a bit strange, if we show every context in the hierarchy we will have this:
```
context 'with an excluded request'
  context 'when providing the remove exclusion parameter'
    context 'and not having any excluded request'
```
To me its kind of strange having this excluded-but-not-excluded use case.
What do you think?

danidoni:
[Allow to stage an excluded request in one step](https://trello.com/c/yXQcl7VA/1309-allow-to-stage-an-excluded-request-in-one-step)

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
saraycp: Ok I understand it's more complex than I thought. What do you think of switching the last two contexts?

This remains the same
```
context 'with an excluded request'
  context 'when not providing the remove exclusion parameter'
  context 'when providing the remove exclusion parameter'
```

```
context 'without an excluded request'  # Before: context 'and not having any excluded request'
  context 'when providing the remove exclusion parameter' 
```

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
vpereira: however as @saraycp say, consistence is the key. Everywhere else, AFAICR, we use 1. It is expected to have 1 here too.

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
danidoni: As we talked offline, we are just checking parameter presence, so we don't really care what's inside that parameter

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
danidoni: I'll use a `'1'` in the specs and documentation, but just check for parameter presence in the controller

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
danidoni: We agreed offline to not switching the contexts

danidoni:
> To be consistent, please change the parameter from true to 1.

Done

danidoni:
> You are missing api documentation.

Added API Documentation

danidoni:
@vpereira @hennevogel please review again

vpereira:
> @vpereira @hennevogel please review again

Now would be nice if you could squash the commits. For example the specs related could  be squashed in one commit.

src/api/app/models/staging/staged_requests.rb
```diff
@@ -81,6 +90,10 @@ def not_found_requests
     request_numbers - requests.pluck(:number).map(&:to_s)
   end
 
+  def excluded_requests
```
coolo: I find this a very unfortunate function name taking that it does not return requests, but strings of numbers. And I find it unfortunate to call such an expensive query in a loop - is this codeclimate speaking? :)

docs/api/api/api.txt
```diff
@@ -2012,7 +2012,9 @@ GET /staging/<staging_workflow_project>/staging_projects/<staging_project>/stage
 XmlResult: staged_requests
 
 POST /staging/<staging_workflow_project>/staging_projects/<staging_project>/staged_requests
-  Add one or multiple bs_requests to the staged requests of a staging project.
+  Add one or multiple bs_requests to the staged requests of a staging project
+  remove_exclusion: If remove_exclusion is set (any value will work) it will allow the removal of the exclusions over a request before staging a request.
```
coolo: If I read the code right, a value of 0 will not 'work'. the api.txt is generally bad in giving valid values - so in line with other paragraphs in the same document I find: `remove_exclusion: removes the request from exclusion before staging` to be enough.

danidoni:
> > @vpereira @hennevogel please review again
> 
> Now would be nice if you could squash the commits. For example the specs related could be squashed in one commit.

Of course!

src/api/app/models/staging/staged_requests.rb
danidoni: You are right. On it

danidoni: Fixed, check it out, please.

docs/api/api/api.txt
danidoni: If you put a '0' in it, it would behave as if you put a '1' in it. That's what I wanted to say with `any value will work`. But let's use your suggestion.

coolo: ah, I always forget that ruby is not perl :)

coolo:
```
~>perl -e 'my $a="0"; print "no\n" unless $a;'
no
```

danidoni:
@coolo check it out again please

src/api/app/controllers/staging/staged_requests_controller.rb
```diff
@@ -15,22 +15,27 @@ def index
...
-      render_ok
-    else
-      render_error(
-        status: 400,
-        errorcode: 'invalid_request',
-        message: "Assigning requests to #{@staging_project} failed: #{result.errors.to_sentence}."
-      )
-    end
+    render_ok
+  rescue Staging::ExcludedRequestNotFound => e
```
coolo: this is all api, no? then better not use `StandardError` but ` APIError`, then you don't need to rescue at all. Just

```
class ExcludedRequestNotFound < APIError
   setup 'invalid_request'
end
```

And it will render a 400 - and if you throw it, you give it the message

vpereira: and why you care here but not https://github.com/openSUSE/open-build-service/pull/8695 @coolo ? That's what I was suggesting - specially because the code there happens in `show` and `accept`

src/api/app/controllers/staging/staged_requests_controller.rb
```diff
@@ -15,22 +15,27 @@ def index
...
+
+    ::Staging::StagedRequests.new(
       request_numbers: @request_numbers,
       staging_workflow: @staging_workflow,
       staging_project: @staging_project,
       user_login: User.session!.login
-    ).create
+    ).create!
 
-    if result.valid?
```
vpereira: as the code was before, was perfect to be extracted to a service...

src/api/app/controllers/staging/staged_requests_controller.rb
danidoni: I've applied the changes @coolo suggested. I need to close this PR now. @victor, @coolo we can review error managing later in another issue.

danidoni:
@coolo please, can you review it again?

src/api/app/models/staging/excluded_request_not_found.rb
```diff
@@ -0,0 +1,9 @@
+class Staging::ExcludedRequestNotFound < APIError
+  attr_reader :message
```
coolo: you really don't need all that. APIError already brings that. Including the render_error if caught in application_controller. All you need to have is `setup 'invalid_request'` in the class.

src/api/app/controllers/staging/staged_requests_controller.rb
victor: please don’t mention-spam. Thanks

danidoni: Oops sorry, my bad. I wanted to reach @vpereira

src/api/app/models/staging/excluded_request_not_found.rb
vpereira: @coolo is that the change request or just a comment? Please help @danidoni to get this PR done.

coolo: Overriding initialize of APIError to is basically just making a StandardError again - and I said previously this is wrong.

danidoni: Done

PR #8643, opened 2019-10-31
Return a value in the method check_for_update_project

vpereira:
Looking through the code in the method `check_for_update_project` I realized that we were changing `p` however we weren't returning it therefore all the changes aren't having any effect.


hennevogel:
`p` is an entry in the `@package` hash outside this method. No need to return it..

Issue #8527, opened 2019-10-07
Split history information out of /staging/$PRJ/staging_projects 


coolo:
The API route `/staging/openSUSE:Factory/staging_projects` is one that needs to be fetched often as it's the one to list the selections and problems of a staging workflow. It contains tons of history information though that is very seldomly needed. Either make this an optional part of the XML or split it into an extra route

coolo:
Same for `/staging/openSUSE:Factory/staging_projects/openSUSE:Factory:Staging:A`

coolo:
I guess you can easily drop the history from `/staging/openSUSE:Factory/staging_projects` completely - wanting all history of all staging projects would be so exceptional, that you can easily write a loop

PR #8644, opened 2019-10-31
[WIP] Split staging history

saraycp:
When retrieving staging projects via API all the related history was shown although it was seldomly needed. Now, there is a new route to request the history (and some other extra information soonish) for a concrete staging project: `staging/<project>/staging_projects/<staging_project>/detail`.

Fixes: #8527


saraycp:
This PR proposed a solution with two options: default output and detailed output (which will include history and probably status and requests in the future).

I close this PR and will propose another solution allowing any combination of the 3 blocks of extra information.

PR #8645, opened 2019-10-31
Introduces sphinx start task

hennevogel:
Move the logic of when to build indexes while starting sphinx from the systemd unit
to a rake tasks so it's more robust and easier to understand.


hennevogel:
This should fix openqa. systemd was not able to start the `ExecStart` line anymore...

coolo:
> This should fix openqa. systemd was not able to start the `ExecStart` line anymore...

'should fix' is a signal for the need of staging projects - just saying :)

PR #8657, opened 2019-11-04
Simplify API staging projects results

### LINKED ISSUE #8527: Split history information out of /staging/$PRJ/staging_projects

### LINKED ISSUE #8534: No cheap way to get the list of staging projects
`/staging/openSUSE:Leap:15.2/staging_projects` includes tons of informations for all staging projects, but there is no way to just get the list of staging projects.

I'd suggest that /staging_projects just returns an index and for details you need to query .../staging_projects/...:A

saraycp:
When retrieving staging projects via API, only a minimal set of information is going to be shown. Any extra information like history, status or even staged requests can be ordered optionally by setting the corresponding parameter in the URL.

Examples:
  - `staging/<project>/staging_projects?requests=1&status=1`
  - `staging/<project>/staging_projects/<staging_project>?requests=1&status=1&history=1`

This fixes #8527 and fixes #8534

### DIFF
--- a/src/api/spec/controllers/staging/staging_projects_controller_spec.rb
+++ b/src/api/spec/controllers/staging/staging_projects_controller_spec.rb
@@ -86,29 +86,134 @@ RSpec.describe Staging::StagingProjectsController do
         untracked_request.change_review_state(:accepted, by_group: staging_workflow.managers_group.title)
         bs_request.change_review_state(:accepted, by_group: staging_workflow.managers_group.title)
         logout
-        get :show, params: { staging_workflow_project: staging_workflow.project.name, staging_project_name: staging_project.name, format: :xml }
       end
 
-      it { expect(response).to have_http_status(:success) }
+      context 'without requesting extra information' do
+        before do
+          get :show, params: { staging_workflow_project: staging_workflow.project.name, staging_project_name: staging_project.name, format: :xml }
+        end
 
-      it 'returns the staging_project name xml' do
-        assert_select 'staging_project' do
-          assert_select 'staged_requests', 1 do
-            assert_select 'request', 3
+        it { expect(response).to have_http_status(:success) }
+
+        it { expect(response.body).not_to include("<staging_project name=\"#{staging_project.name}\" state=") }
+
+        it 'returns the staging_project default xml' do
+          assert_select 'staging_project' do
+            assert_select 'staged_requests', 0
+            assert_select 'untracked_requests', 0
+            assert_select 'requests_to_review', 0
+            assert_select 'missing_reviews', 0
+            assert_select 'broken_packages', 0
+            assert_select 'history', 0
           end
-          assert_select 'untracked_requests', 1 do
-            assert_select 'request', 1
+        end
+      end
+
+      context 'with staged requests' do
+        before do
+          get :show, params: { staging_workflow_project: staging_workflow.project.name, staging_project_name: staging_project.name,
+                               requests: 1, format: :xml }
+        end
+
+        it { expect(response).to have_http_status(:success) }
+
+        it { expect(response.body).not_to include("<staging_project name=\"#{staging_project.name}\" state=") }
+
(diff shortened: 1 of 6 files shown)


src/api/spec/controllers/staging/staging_projects_controller_spec.rb
```diff
@@ -86,29 +86,105 @@
         untracked_request.change_review_state(:accepted, by_group: staging_workflow.managers_group.title)
         bs_request.change_review_state(:accepted, by_group: staging_workflow.managers_group.title)
         logout
-        get :show, params: { staging_workflow_project: staging_workflow.project.name, staging_project_name: staging_project.name, format: :xml }
       end
 
-      it { expect(response).to have_http_status(:success) }
+      context 'with staged requests' do
```
dmarcoux: We should also test at least once with `request: 0`, `history: 0` or `status: 0`.

dmarcoux:
I wonder if we should just check for the presence of the query string parameters instead of checking if it's `0` or not. I don't really see an added value in checking for the parameters' values.

src/api/spec/controllers/staging/staging_projects_controller_spec.rb
saraycp: After a short discussion, we decided that any value means "true". No value or a missing parameter means "false".
I'll adjust the conditions...

coolo:
(as discussed in chat - just for completness):

this does not fix #8534 as the overall state is part of the output even with no option provided - and this is still calculating everything.

saraycp:
> (as discussed in chat - just for completness):
> 
> this does not fix #8534 as the overall state is part of the output even with no option provided - and this is still calculating everything.

Exactly! As it is, it simplifies the output but still takes a long time because of the `overall_state`. I've added a WIP and I'll also make that value optional.

However, I wonder if it makes sense to hide the state when listing all the staging projects. Are only the names enough information? @coolo @openSUSE/obs-frontend

coolo:
please disable the overall state if status is falsey. I have use cases to only need the names (when I need to create a new adi project, I need to exclude all existing ones) and I have use cases where I just need the selected requests and don't care about the state

saraycp:
@coolo @dmarcoux  can you have another look?

Now the `state` xml attribute (overall state) is no longer shown by default, only when requesting `status` information.

The tests have been adjusted to check if the `state` attribute is included in the xml or not.

Apidocs has also been updated accordingly.

coolo:
```I, [2019-11-06T08:02:16.006975 #26551]  INFO -- : [a83654f7-cb23-4977-aa4d-a16108d87a38] [26551:6.34] method=GET path=/staging/openSUSE:Factory/staging_projects format=xml controller=Staging::StagingProjectsController action=index status
=200 duration=79.83 view=59.99 db=7.22 params={"staging_workflow_project"=>"openSUSE:Factory"}
```

Much faster now :)

saraycp:
Only staging projects' names in the output:
```
duration=79.83 view=59.99 db=7.22 
 ```
Staging projects' names and overall state in the output:
```
duration=7237.45 view=5793.19 db=1367.26
```
Everything (name, overall state, staged requests, history) in the output:
```
duration=23374.02 view=19766.09 db=3564.14
```

Definitely it's much more faster! :rocket:

coolo:
It will take some time to adopt the plugin to use this API, but it should prove itself very useful.

coolo:
Don't wait for OBS - it's under heavy load

coolo:
https://github.com/openSUSE/openSUSE-release-tools/pull/2295/commits/aceb9565fd953db2c92d31557ff6beb59d979fa7 adopts some commands to the new API and it's super fast now. And maps easily to python arguments as well

PR #8659, opened 2019-11-05
Refactor project model arrays

vpereira:
Refactored out some array building tasks, substituting it for enumerator operations and added specs for it.

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -766,11 +766,7 @@ class Project < ApplicationRecord
   end
 
   def expand_all_repositories
-    all_repositories = repositories.to_a
-    repositories.each do |repository|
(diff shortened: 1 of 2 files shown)


src/api/app/models/project.rb
```diff
@@ -766,11 +766,7 @@ def find_package(package_name, check_update_project = nil, processed = {})
   end
 
   def expand_all_repositories
```
danidoni: :thought_balloon: What about renaming this method to `expanded_repositories` to avoid the same name with the `expand_all_repositories` method below?

vpereira: Yes, kind of ugly, but I don't really want to change any public API in this PR, thank you!

PR #8660, opened 2019-11-05
Take local links into account when staging/unstaging requests

vpereira:
When staging a request, we have to take into account if there are more than one spec in the target package, and if there is, we should create the linking among them accordingly 

TODO:

- [x] make the destroy more robust, keep track from errors when deleting linked packages
- [ ] add specs

### DIFF
--- a/src/api/app/models/staging/staged_requests.rb
+++ b/src/api/app/models/staging/staged_requests.rb
@@ -91,6 +91,10 @@ class Staging::StagedRequests
     source_package = Package.get_by_project_and_name!(bs_request_action.source_project,
                                                       bs_request_action.source_package)
 
+    # it is possible that target_package doesn't exist
+    target_package = Package.get_by_project_and_name(bs_request_action.target_project,
+                                                     bs_request_action.target_package)
+
     query_options = { expand: 1 }
     query_options[:rev] = bs_request_action.source_rev if bs_request_action.source_rev
 
@@ -102,10 +106,16 @@ class Staging::StagedRequests
 
     link_package = Package.find_or_create_by!(project: staging_project, name: bs_request_action.target_package)
 
-    Backend::Api::Sources::Package.write_link(staging_project.name,
-                                              link_package.name,
-                                              User.session!,
-                                              "<link project=\"#{source_package.project.name}\" package=\"#{source_package.name}\" rev=\"#{package_rev}\" vrev=\"#{source_vrev}\"></link>")
+    create_link(staging_project.name, link_package.name, User.session!, project: source_package.project.name,
+                                                                        package: source_package.name, rev: package_rev,
+                                                                        vrev: source_vrev)
+    # for multispec packages, we have to create local links to the main package
+    if target_package.present?
+      target_package.find_project_local_linking_packages.each do |local_linking_package|
+        linked_package = Package.find_or_create_by!(project: staging_project, name: local_linking_package.name)
+        create_link(staging_project.name, linked_package.name, User.session!, package: target_package.name, cicount: 'copy')
+      end
+    end
 
     ProjectLogEntry.create!(
       project: staging_project,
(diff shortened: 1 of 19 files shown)


src/api/app/models/staging/staged_requests.rb
```diff
@@ -91,6 +91,9 @@ def link_package(bs_request_action)
     source_package = Package.get_by_project_and_name!(bs_request_action.source_project,
                                                       bs_request_action.source_package)
 
+    target_package = Package.get_by_project_and_name!(bs_request_action.target_project,
```
coolo: Make sure you spec this for 'if the target package does not yet exist'

coolo: Actually on second thought: the target project has no business here at all. All spec files are required to be linked - if the submission introduces new ones, those as well.

vpereira: @coolo `target_project` or `target_package` ?

coolo: `target_project` - the target_package is the name, but within the staging project.

vpereira: Let's see, if my rational is right:

I'm using the `target_project` + `target_package` here, because in the `curl` package on `openSUSE:Factory` (which will normally will be our `target_project` ) is where I'm going to find the local links, i.e for curl will be the `curl-mini`

coolo: the point is *it will be* - if you submit curl-ng to openSUSE:Factory, the staging project still needs to linkpac curl-ng-mini to curl-ng

Same as submitting curl-now-as-mutlibuild should *not* create a curl-mini in the staging project. The target project's local links have *zero* influence on the local links in staging projects.

vpereira: I couldn't really follow your example, but The last sentence was key to me. 

"The target project's local links have zero influence on the local links in staging projects." is the opposite that we were assuming. Could you please describe me (maybe as issue, if a comment isn't the best place for it) how exactly those links should be built? Today I'm just able to reverse engineer it https://build.opensuse.org/package/show/openSUSE:Factory/curl-mini

hennevogel: @coolo well, creating new local links (in staging or target) based on the existence of build descriptions in the source is not a route I want to go down with this. This is too distro specific for my taste. That you would have to do on your own with some bot or something, sorry.

hennevogel: This will throw an exception for submissions of packages that would be new in `target_project`. It shouldn't.

coolo: I don't believe that local links are more distribution specific than staging projects in general. But that also means I can't use the accept command of a staging project, but need to do everything from the client. That's a pity after all the work put into accepting.

hennevogel: @coolo let's discuss this offline please

coolo: sure

src/api/app/models/staging/staged_requests.rb
```diff
@@ -129,15 +141,15 @@ def add_review_for_unstaged_request(request, staging_project)
...
   end
 
-  def remove_packages(request, staging_project)
-    package_names = request.bs_request_actions.pluck(:target_package)
-    staging_project_packages = staging_project.packages.where(name: package_names)
-    staging_project_packages.each do |package|
-      next if package.destroy
+  def remove_packages(staging_project_packages)
+    staging_project_packages.collect do |package|
+      (package.find_project_local_linking_packages | [package]).collect { |pkg| pkg unless pkg.destroy }
```
coolo: A destroy in a double nested collect function? I can't say this is clear code

src/api/app/models/staging/staged_requests.rb
coolo: I actually wonder when "let's discuss this offline please" became "I leave for vaction, let's merge this anyway".

vpereira: Probably, he assumed that because it's not in production, we could iterate through it again until we get it right, and it shouldn't block our deliverable (because a specification change/misunderstanding in the day that we actually should finish the deliverable) 

If you could take part in the planning meetings, or find a way to spec it in a proper way, maybe we could avoid those issues..  PR based development, where you specify the feature when the PR is ready, probably won't work.

coolo: let's discuss this offline please

PR #8666, opened 2019-11-06
Disable 15.0 and 15.1 as valid distributions

coolo:
We only care for openSUSE Leap 15.2 going forward

Co-authored-by: @lnussel


lnussel:
:+1:

PR #8667, opened 2019-11-06
Run sphinx after OBS api setup is done

eduardoj:
This way we prevent Sphinx to throw errors of connection to mariadb when the setup of OBS is not finished. This happens the first time the appliance is booted.


eduardoj:
OBS Package Build tests are frozen in `blocked` state. Merging.

PR #8671, opened 2019-11-07
Create Sphinx start Rails task for development environment

eduardoj:
Provide a way to reindex Sphinx when it is started in `NODETACHED` mode. This is only needed in development environment.

The check for rebuilding the configuration of the index was modified to include the condition of no existence of the `config/*.sphinx.conf` file. That is the case of the development environment.

This is a workaround until pat/thinking-sphinx#1150 is fixed.

### DIFF
--- a/src/api/lib/tasks/sphinx.rake
+++ b/src/api/lib/tasks/sphinx.rake
@@ -8,8 +8,31 @@ namespace :sphinx do
       Rake::Task['ts:start'].invoke
     end
   end
+
+  desc 'Start the sphinx daemon for the development environment'
+  task start_for_development: :environment do
+    if index_to_build?
+      Rake::Task['ts:clear'].invoke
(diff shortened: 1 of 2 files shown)


src/api/lib/tasks/sphinx.rake
```diff
@@ -8,8 +8,31 @@ namespace :sphinx do
...
     end
   end
+
+  desc 'Start the sphinx daemon for the development environment'
+  task start_for_development: :environment do
+    if index_to_build?
+      Rake::Task['ts:clear'].invoke
+      Rake::Task['ts:configure'].invoke
+      t = Thread.new do
+        counter = 0
```
danidoni: :thought_balloon:  What about renaming this to `retries`?

eduardoj: Done.

Issue #8524, opened 2019-10-07
Excluded requests should be shown as package names


coolo:
In the backlog I see the package name of excluded requests:
![screencapture-build-test-opensuse-org-staging-workflows-1-2019-10-07-13_53_28](https://user-images.githubusercontent.com/1067203/66309489-169f6380-e90a-11e9-9398-8577cbecd397.png)

But on the 'excluded requests' page I have to enter request numbers and are presented those:

![screencapture-build-test-opensuse-org-staging-workflows-1-excluded-requests-2019-10-07-13_53_42](https://user-images.githubusercontent.com/1067203/66309501-1b641780-e90a-11e9-98b7-bce0829d2072.png)

Don't know about you, but I'm bad with numbers. Especially as they have 6 digits as on build.*

PR #8676, opened 2019-11-07
Replace request number by package name in excluded requests

saraycp:
In both webui and API, excluded requests list shows the package name, before there was only the request number.

Fixes: #8524

How to test?

```
osc api -X GET '/staging/<project>/excluded_requests'
```
or in webui

```
/staging_workflows/<staging_workflow_id>/excluded_requests
```

### DIFF
--- a/src/api/app/datatables/excluded_request_datatable.rb
+++ b/src/api/app/datatables/excluded_request_datatable.rb
@@ -10,7 +10,7 @@ class ExcludedRequestDatatable < Datatable
 
   def view_columns
     @view_columns ||= {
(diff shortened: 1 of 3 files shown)


coolo:
You're still required to enter a number to add excludes in the webui, but it's to be seen how often people actually use this. So far we relied purely on the command line for this, so displaying it properly is more important.

saraycp:
> You're still required to enter a number to add excludes in the webui, but it's to be seen how often people actually use this. So far we relied purely on the command line for this, so displaying it properly is more important.

@coolo, I opened this issue so we don't forget it: #8677

PR #8691, opened 2019-11-09
[backend] add example for global Macros to BSConfig.pm

computersalat:
To prevent other users from reviewing the code to find out how to define Macros on a global base, that you don't need to define them in every `OBS Project`, I added an example to the `BSConfig.pm.template`


adrianschroeter:
it is not a good practice to put macros inside of BSConfig.pm, but hey ...

Issue #8686, opened 2019-11-08
[production][staging_projects#show] undefined method `staging_projects' for nil:NilClass


coolo:
staging_projects#show threw this 0 times
<pre>
undefined method `staging_projects&#39; for nil:NilClass
</pre>
[https://api.opensuse.org/staging/openSUSE:Leap:15.1/staging_projects/O](https://api.opensuse.org/staging/openSUSE:Leap:15.1/staging_projects/O)"

[See this exception on Errbit](http://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5dc590e4cd4ef600088339ac)

coolo:
api-test does not crash, but returns a strange error:

```
coolo@cleopatra#osc-plugin-factory>tosc api /staging/openSUSE:Factory/staging_projects/O
Server returned an error: HTTP Error 404: Not Found
Couldn't find Project with [WHERE (projects.id not in (0)) AND `projects`.`staging_workflow_id` = ? AND `projects`.`name` = ?]

saraycp:
> api-test does not crash, but returns a strange error:
> 
> ```
> coolo@cleopatra#osc-plugin-factory>tosc api /staging/openSUSE:Factory/staging_projects/O
> Server returned an error: HTTP Error 404: Not Found
> Couldn't find Project with [WHERE (projects.id not in (0)) AND `projects`.`staging_workflow_id` = ? AND `projects`.`name` = ?]
> ```

This seems not to be related to this issue, I'll create a new issue just for this.

PR #8695, opened 2019-11-11
Add error message for non-existent staging project

saraycp:
Before, it was raising an exception when the staging project didn't exist. 

```
osc api -X GET '/staging/openSUSE:Factory/staging_projects/openSUSE:Factory:Staging:Fake'                                             
      
=>                                                                                
Couldn't find Project with [WHERE (projects.id not in (0)) AND `projects`.`staging_wor
kflow_id` = ? AND `projects`.`name` = ?]
```

Now, it shows a proper error.

```
osc api -X GET '/staging/openSUSE:Factory/staging_projects/openSUSE:Factory:Staging:Fake' 

=>
Staging Project openSUSE:Factory:Staging:Fake not found.
```

Fixes: #8700 and also fixes: #8686

### DIFF
--- a/src/api/app/controllers/source/errors.rb
+++ b/src/api/app/controllers/source/errors.rb
@@ -94,4 +94,8 @@ module Source::Errors
   class NotLocked < APIError; end
 
   class InvalidFlag < APIError; end
+
+  class StagingProjectNotFound < APIError
+    setup 'not_found', 404
+  end
 end
--- a/src/api/app/controllers/source/errors.rb
+++ b/src/api/app/controllers/source/errors.rb
@@ -95,6 +95,10 @@ module Source::Errors
 
   class InvalidFlag < APIError; end
 
+  class StagingWorkflowNotFound < APIError
+    setup 'not_found', 404
+  end
+
   class StagingProjectNotFound < APIError
     setup 'not_found', 404
   end
--- a/src/api/app/controllers/staging/staging_projects_controller.rb
+++ b/src/api/app/controllers/staging/staging_projects_controller.rb
@@ -71,6 +71,8 @@ class Staging::StagingProjectsController < Staging::StagingController
   end
 
   def set_staging_project
+    raise StagingWorkflowNotFound, "Staging Workflow for project \"#{@project.name}\" does not exist." unless @project.staging
+
     @staging_project = @project.staging.staging_projects.find_by(name: params[:staging_project_name])
     return if @staging_project
     raise StagingProjectNotFound, "Staging Project \"#{params[:staging_project_name]}\" does not exist."
--- a/src/api/app/controllers/staging/staging_projects_controller.rb
+++ b/src/api/app/controllers/staging/staging_projects_controller.rb
@@ -1,8 +1,11 @@
 class Staging::StagingProjectsController < Staging::StagingController
+  include Source::Errors
+
   before_action :require_login, except: [:index, :show]
   before_action :set_project
   before_action :set_staging_workflow, only: :create
   before_action :set_options, only: [:index, :show]
+  before_action :set_staging_project, only: [:show, :accept]
 
   validate_action create: { method: :post, request: :staging_project }
 
@@ -19,9 +22,7 @@ class Staging::StagingProjectsController < Staging::StagingController
     end
   end
 
-  def show
-    @staging_project = @project.staging.staging_projects.find_by!(name: params[:staging_project_name])
-  end
+  def show; end
 
   def create
     authorize @staging_workflow
@@ -46,10 +47,9 @@ class Staging::StagingProjectsController < Staging::StagingController
   end
 
   def accept
-    staging_project = Project.find_by!(name: params[:staging_project_name])
-    authorize staging_project, :update?
+    authorize @staging_project, :update?
 
-    if staging_project.overall_state != :acceptable
+    if @staging_project.overall_state != :acceptable
       render_error(
         status: 400,
         errorcode: 'invalid_request',
@@ -57,7 +57,7 @@ class Staging::StagingProjectsController < Staging::StagingController
       )
       return
     end
-    StagingProjectAcceptJob.perform_later(project_id: staging_project.id, user_login: User.session!.login)
+    StagingProjectAcceptJob.perform_later(project_id: @staging_project.id, user_login: User.session!.login)
     render_ok
   end
 
@@ -68,5 +68,11 @@ class Staging::StagingProjectsController < Staging::StagingController
     [:requests, :history, :status].each do |option|
       @options[option] = params[option].present?
     end
+  end
+
+  def set_staging_project
+    @staging_project = @project.staging.staging_projects.find_by(name: params[:staging_project_name])
+    return if @staging_project
+    raise StagingProjectNotFound, "Staging Project \"#{params[:staging_project_name]}\" does not exist."
   end
 end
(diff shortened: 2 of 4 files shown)


dmarcoux:
It's all good, I pushed a commit to test the changes instead of commenting about adding a spec. 

The only thing left is to squash the commits.

coolo:
But why does build.opensuse.org throw a 500 then?

coolo:
Oh, possibly because I'm admin on the one instance and normal user on the other?

saraycp:
> It's all good slightly_smiling_face
> 
> I pushed a commit to test the changes instead of commenting about adding a spec.
> 
> The only thing left is to squash the commits.

Thanks @dmarcoux, squashed!

coolo:
Ah, no. So sorry @saraycp but this is not fixing the issue - I just mixed up two things in that issue.

```
coolo@f102#api>tosc api -X GET '/staging/openSUSE:Factory/staging_projects/openSUSE:Factory:Staging:Fake'  
Couldn't find Project with [WHERE (projects.id not in (0)) AND `projects`.`staging_workflow_id` = ? AND `projects`.`name` = ?]
coolo@f102#api>tosc api -X GET '/staging/home:coolo/staging_projects/openSUSE:Factory:Staging:Fake'  
Server returned an error: HTTP Error 500: Internal Server Error
```

You fix the rather strange error message in here, but the issue is about the crash on checking the route on a project without staging workflow. Just didn't realize that I mixed 15.1 and 15.2 when trying obs :)

saraycp:
> Ah, no. So sorry @saraycp but this is not fixing the issue - I just mixed up two things in that issue.

No worries, I'll create another issue just for this.

saraycp:
Fixes: #8700

src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -20,7 +20,14 @@ def index
   end
 
   def show
-    @staging_project = @project.staging.staging_projects.find_by!(name: params[:staging_project_name])
```
vpereira: Hm, isn't that better to raise `ActiveRecord::RecordNotFound` and catch it on `application_controller` ? That's, IMO, the "OBS way" to handle it.

coolo: It's already caught there. That's why you get a 404 and not a 500. It's just that `RecordNotFound` renders this rather strange error. So to fix it 'the OBS way', you would have to raise e.g. InvalidProjectParameters

vpereira: @coolo I'm fine with that, even before if we go DRY and use the same code doing that for `show` and `accept`, or do you think that's better to do the error handling in every single action?

coolo: I don't care either way. I'm just saying that your review comment makes little sense as the `find_by!` that @saraycp patches out is doing exactly what you're suggesting - including the creation of the issue.

So custom error message is all that's required - either through render_error or raise. I prefer raise, but exceptions aren't everyone's darling.

saraycp: @vpereira I understand your point, the DRYer the better. But this is an exceptional case and  we just want to customize the message at this point for now.

I've force-pushed already to use `raise` instead of `render_error`, halfway between both ideas.

vpereira: > don't care either way. I'm just saying that your review comment makes little sense as the `find_by!` that @saraycp patches out is doing exactly what you're suggesting

it was a rhetoric question @coolo ;)

@saraycp I don't see why we should move away from the `bang` finder methods.. any reason?  I had the impression, based in past reviews, that it was the rule here.. BTW I think the action `accept` has the same problem right? So maybe a `before_action` taking care of that for both methods - raising something like https://github.com/openSUSE/open-build-service/blob/00d38f203dd9469c132a3abb3de00fb2525946ad/src/api/app/controllers/source/errors.rb would be, IMO,  a better approach

coolo: I personally find the proposed solution pretty brilliant - no need for a different exception. And the bang method throws that error the issue is about - you can't pass that message to the exception as @saraycp does here, so I see no way not to move away from them *and* having custom errors.

coolo:
8704 is for the webui and requires different mechanics.

saraycp:
> 8704 is for the webui and requires different mechanics.

Exactly! That was Webui, I just had to redirect a set a flash message when something wrong happend.

This PR is related to API and I planned to provide a solution taking into account your suggestions, @vpereira. For this we are not in a hurry, so I add WIP.

saraycp:
@vpereira, I've added a commit trying to follow your proposal. Am I in the right direction?

A few things to have into account:

- The `find_by!` we had in `accept` weren't exactly the same as the one in `show` action. I've unified and moved them to a `before_action`. I hope it doesn't make any difference.
- Calling `find_by!`, as usual, doesn't allow us to customize the error message. So I have to call `find_by` instead and raise an exception with a customized message.
- I created a new exception in error.rb and raise it when the staging project doesn't exist. *

* I created and exception but I still don't see the benefit of doing this over raising an already existent exception like 'ActiveRecord::RecordNotFound'.

src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -69,4 +69,10 @@ def set_options
       @options[option] = params[option].present?
     end
   end
+
+  def set_staging_project
+    @staging_project = @project.staging.staging_projects.find_by(name: params[:staging_project_name])
```
coolo: now also raise  *NotFound `unless @project.staging` to fix #8686 - you're so close :)

src/api/app/controllers/source/errors.rb
```diff
@@ -94,4 +94,8 @@ class ProjectCopyNoPermission < APIError
   class NotLocked < APIError; end
 
   class InvalidFlag < APIError; end
+
+  class StagingProjectNotFound < APIError
+    setup 'not_found', 404
```
coolo: btw: if you just setup 404, the error will be `staging_project_not_found` - and this is basically the only benefit of not using the generic `ActiveRecord` exception. The class name of the exception becomes the user facing error.

src/api/app/controllers/staging/staging_projects_controller.rb
saraycp: Just to clarify: it raised a 500 error when the project did exist but it wasn't associated to a staging workflow. If the project didn't exist, it was raising a proper 404 error.

Now the first case is also raising a 404. Please have another look, @coolo.

saraycp:
Fixes: #8700 and also fixes: #8686

src/api/app/controllers/source/errors.rb
coolo: Really, leaving the 'not_found' improves the API usage - a little bit. API errors follow e.g. this format:

```
osc -d api -X PUT -T Rakefile /source/GNOME:Factory/gnome-shell/Rakefile
...
<status code="put_file_no_permission">
   <summary>Insufficient permissions to store file in package gnome-shell, project GNOME:Factory</summary></status>
```

This maps to `raise PutFileNoPermission, "Insufficient permissions to store file in package #{@package_name}, project #{@project_name}"`

So with 'not_found', it says code='not_found', without it it says code='staging_workflow_not_found' or code='staging_project_not_found' - and the client has an easy way to differ. And you have to type less.

But I don't think it's important enough to hold it back more.

PR #8717, opened 2019-11-14
[dist] require newer build script

adrianschroeter:
or backend services will fail to load


adrianschroeter:
ignored unrelated ruby packaging issues

PR #8719, opened 2019-11-14
Fix package build updating sassc and rake

coolo:
They don't like to play well with each other, so at least update them to their latest version.

For sassc we need to disable the cpu specific optimization to avoid running into issue 146 again


coolo:
To be combined with https://build.opensuse.org/request/show/748746

PR #8728, opened 2019-11-18
Drop no longer used attributes created and updated from the documentaion

### LINKED ISSUE #8725: project xml schema mentions non existant attributes `created` and `updated`
# Issue Description

The API [documentation/schema](https://build.opensuse.org/apidocs/project.rng) of the project xml element mentions the optional `created` and `updated` attributes.

I have tried setting them (both via the webUI and the API), but they are not retained by OBS, it appears to forget them immediately.

Expected Result
============

I would have guessed that the `created` and `updated` elements would be set automatically and writing them would result in an error. Or that OBS rejects invalid dates (it currently does not). It is also odd that the value written to the field is simply discarded.

Further Information
===============
* A link to an OBS instance showing the issue: build-test.opensuse.org

dcermak:
These attributes have been removed in 2007 and shouldn't be used anymore.
This fixes #8725.

### DIFF
--- a/docs/api/api/package.rng
+++ b/docs/api/api/package.rng
@@ -46,12 +46,6 @@
       <optional>
         <attribute name="project"/>
       </optional>
-      <optional>
-        <attribute name="created"/>
-      </optional>
-      <optional>
-        <attribute name="updated"/>
-      </optional>
 
       <element ns="" name="title">
         <text/>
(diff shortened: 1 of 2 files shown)


coolo:
I thought you also wanted it to drop from package.rng?

dcermak:
@coolo Done!

PR #8736, opened 2019-11-18
Prevent staging package twice

### LINKED ISSUE #8730: It shouldn't be possible to stage the same package name twice
As packages are linked, they overwrite each other - so having 2 requests for the same target package is plainly impossible. Still the API doesn't catch it and happily tracks 2 requests for the same package in the same staging project.

eduardoj:
Having two requests for the same target package makes no sense.

Fix #8730.


src/api/app/models/staging/staged_requests.rb
```diff
@@ -120,7 +125,7 @@ def link_package(bs_request_action)
 
     package_rev = backend_package_information['srcmd5']
 
-    link_package = Package.find_or_create_by!(project: staging_project, name: bs_request_action.target_package)
```
coolo: We originally had the find_or_create to make superseding work - but as requests don't supersede each other, Eduardo and I discussed that errors are easier to handle from the client's side.

PR #8742, opened 2019-11-19
Stage all requests, but only link submit requests

### LINKED ISSUE #8738: Delete requests aren't staged
*Any* request that is POSTED to staged_requests should be staged afterwards (unless an error is returned) - delete requests are silently ignored at the moment. For now only submit requests are linked, but all other types (like e.g. also change_devel or set_bugowner) should be staged as well and be accepted together with the staging project.

This is a followup for #8737 and as such P3

coolo:
Fixes #8738

### DIFF
--- a/src/api/app/models/staging/staged_requests.rb
+++ b/src/api/app/models/staging/staged_requests.rb
@@ -11,9 +11,17 @@ class Staging::StagedRequests
         link_package(bs_request_action)
       end
       # TODO: implement delete requests
+      ProjectLogEntry.create!(
+        project: staging_project,
+        user_name: user_login,
+        bs_request: request,
+        event_type: :staged_request,
+        datetime: Time.now,
+        package_name: bs_request_action.target_package
+      )
+      staging_project.staged_requests << request
+      add_review_for_staged_request(request)
     end
-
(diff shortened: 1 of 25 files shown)


src/api/app/models/staging/staged_requests.rb
```diff
@@ -11,10 +11,18 @@ def create
         link_package(bs_request_action)
       end
       # TODO: implement delete requests
```
eduardoj: Is there anything left to do to implement delete requests? Does this comment still apply?

coolo: Well, at least it's to be decided - because the client still does this:
https://github.com/openSUSE/openSUSE-release-tools/blob/0d0290f1e6a4fc53f76d377a8b5686a714751e69/osclib/stagingapi.py#L1098

coolo: (and there are quite some [corner cases](https://github.com/openSUSE/openSUSE-release-tools/issues/1899) where this fails, so I would actually leave it in the client)

coolo:
I finally managed to convince the test suite to like me - so please someone review

PR #8743, opened 2019-11-19
Rename number in excluded and staged requests

### LINKED ISSUE #8732: Excluded request api is having 'number' - everywhere it's 'id'
docs/api/api/create_excluded_requests.xml has
```
<excluded_requests>
  <request number='33' description='invalid request'/>
</excluded_requests>
```

For the sake of consistency, this should be `id=`

eduardoj:
For the sake of consistency 'number' attribute was renamed to 'id', and 'number' element to 'request'.

Fixes #8732.


docs/api/api/remove_staged_requests.xml
```diff
@@ -1,4 +1,4 @@
 <requests>
-  <number>7</number>
-  <number>13</number>
+  <request>7</request>
```
coolo: same here

docs/api/api/delete_excluded_requests.xml
```diff
@@ -1,3 +1,3 @@
 <excluded_requests>
-  <number>33</number>
+  <request>33</request>
```
coolo: this should be `<request id='33'/>` if you want to be consistent

eduardoj: I only saw the initial version of [your comment](https://github.com/openSUSE/open-build-service/issues/8732#issuecomment-555484334). On it.

eduardoj: Done.

docs/api/api/remove_staged_requests.xml
eduardoj: Done.

PR #8751, opened 2019-11-20
Display correctly delete requests

### LINKED ISSUE #8739: Staging workflow doesn't render delete requests as promised
I added some delete requests to https://build.opensuse.org/staging_workflows/136/staging_projects/openSUSE:Leap:15.2:Staging:C (hacky hack), but they are rendered like submit requests - not as the legend promises.

DavidKang:
Delete requests weren't rendered as specified in the legend.

Fix #8739.

#### Before
Screenshot_2019-11-20 Open Build Service(3)

Screenshot_2019-11-20 Open Build Service(2)

#### After
Screenshot_2019-11-20 Open Build Service(8)

Screenshot_2019-11-20 Open Build Service(7)

### DIFF
--- a/src/api/app/assets/stylesheets/webui/staging-workflow.scss
+++ b/src/api/app/assets/stylesheets/webui/staging-workflow.scss
@@ -88,7 +88,6 @@ ul .table-list-group-item {
     @extend .w-50;
 
     span.badge {
-
       margin-left: 10px;
 
       a.request {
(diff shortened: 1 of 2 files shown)


coolo:
hmm, delete requests can have reviews as well (or be revoked/obsolete). So I think changing the background for them is somewhat false. But we can make that a different issue

DavidKang:
>  delete requests can have reviews as well (or be revoked/obsolete)

Uhmm, good point.

@coolo, something like this would be fine?

Screenshot_2019-11-20 Open Build Service(5)
Screenshot_2019-11-20 Open Build Service(4)

That the meaning of delete request is the line-through the name of the package.

coolo:
that's how you can find it on https://build.opensuse.org/project/staging_projects/openSUSE:Factory - so fine with me.

PR #8752, opened 2019-11-20
Fix collapse on problems list

### LINKED ISSUE #8749: Expandable/collapsible problem list has not functional on staging workflow
# Issue Description
On staging workflow page, if there is any problem then it will be listed at the *Problems* section, there is a "see more" button to expanding problems in case too many problems, however clicked that "see more" button nothing happened.

Expected Result
============
Expanding and collasping function at the problems section should works to show more/less issues.

How to Reproduce
===============
1. Go to https://build.opensuse.org/staging_workflows/136
2. Click on 'See X more' button at the Problems section

For example, it has `See 8 more`... in Staging:N which point to https://build.opensuse.org/staging_workflows/136#hidden-item-openSUSE_Leap_15.2_Staging_N
1. Scroll down to 'Staging:N' entry
2. Click on 'See 8 more'

Further Information
===============

DavidKang:
We have project names with `.`, that was causing the issue. Now apart
from replacing `:` in the project name we also substitute `.` with `_`.

Fix #8749.

### DIFF
--- a/src/api/app/views/webui/staging/workflows/_problems.html.haml
+++ b/src/api/app/views/webui/staging/workflows/_problems.html.haml
@@ -1,6 +1,6 @@
 :ruby
   max_shown = 5
-  project_class = staging_project.name.tr(':', '_')
+  project_class = staging_project.name.tr(':.', '_')
   item_project_class = 'hidden-item-' + project_class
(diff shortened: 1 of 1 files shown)


coolo:
`valid_xml_id` replaces much more btw

PR #8755, opened 2019-11-21
bs_dispatch: Fix log output

rhabacker:
1. fix entry without timestamp
2. print requested uri on getting rpc errors


adrianschroeter:
thanks!

PR #8760, opened 2019-11-21
Force accept for staging projects

coolo:
Took out the extra check in the background job and added extra checks to the policy


src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -46,23 +46,25 @@ def copy
     render_ok
   end
 
+  class StagingProjectNotAcceptable < APIError; end
```
saraycp: What about moving this to `app/controllers/source/errors.rb` as we agreed on previous PRs?

coolo: hmm, did we? putting staging errors into source::errors sounds wrong to me. But I can create a staging/errors.rb and pull your exception from it. You're refering to https://github.com/openSUSE/open-build-service/pull/8695#discussion_r344836772 or something else?

saraycp: Yes, I'm referring to that conversation. Now that you say it :thinking:  it's true it sounds wrong having this in source/errors.rb, but it already contains erros like: `WrongRouteForStagingWorkflow`, `StagingWorkflowNotFound` and `StagingProjectNotFound`.

I'd suggest to have a generic error file for all of them instead of creating one per controller... What do you think?

coolo: Well, the trend is different :)

`git grep 'class.*APIError'` - there are already 3 error.rb

saraycp: In controllers directory I see only trigger/errors.rb and source/errors.rb. However, for now I would say let's create *staging/errors.rb* and move `class StagingProjectNotAcceptable < APIError; end` into it. I'll create an issue to move the rest of the staging-related errors to that file.

coolo: done

PR #8763, opened 2019-11-22
Add validation for xml on exclude requests

### LINKED ISSUE #8756: [production][excluded_requests#create] undefined method `[]' for nil:NilClass
excluded_requests#create threw this 0 times
<pre>
undefined method `[]&#39; for nil:NilClass
</pre>
[https://api.opensuse.org/staging/openSUSE:Leap:15.2/excluded_requests](https://api.opensuse.org/staging/openSUSE:Leap:15.2/excluded_requests)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5dd65d998f79af000dbf522a)

DavidKang:
At the moment we only check if the xml is present or is well formed.
The schema validation is planned for the future.

Fix #8756.

### DIFF
--- a/src/api/app/controllers/staging/excluded_requests_controller.rb
+++ b/src/api/app/controllers/staging/excluded_requests_controller.rb
@@ -1,7 +1,8 @@
 class Staging::ExcludedRequestsController < Staging::StagingController
-  before_action :require_login, except: [:index]
+  before_action :require_login, except: :index
   before_action :set_project
-  before_action :set_staging_workflow, :set_requests_xml_hash
+  before_action :set_staging_workflow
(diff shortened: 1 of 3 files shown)


coolo:
Suse::Validator.validate is really easy to plug into and with tools like https://www.freeformatter.com/xsd-generator.html creating a schema takes a minute.

DavidKang:
> Suse::Validator.validate is really easy to plug into and with tools like https://www.freeformatter.com/xsd-generator.html creating a schema takes a minute.

That makes things easier. We will use it hopefully soon. Thanks :smiley_cat:

PR #8769, opened 2019-11-22
Replace codemirror-rails gem as it's dead (and our CM is very outdated)

coolo:
It's a little sad as the asset gem is so easy to manage - next step would be to get pnm/yarn into the containers


coolo:
https://obs-reviewlab.opensuse.org/coolo-new_codemirror/package/view_file/home:Admin/eee/test.xml?expand=1 looks good to me

DavidKang:
@coolo, these files need to be added in the `manifests` or `initializer/assets.rb` for production? Otherwise LGTM

coolo:
No, they replace 100% the gem - i.e. hook into the asset pipeline as before. And as you can check on the review app, it just works.

PR #8772, opened 2019-11-23
Expect dates as `as_json` expects them

coolo:
Increases compatibility with rails6


coolo:
too bad this is not compatible with rails 5 :)

PR #8776, opened 2019-11-25
WIP: Port to Rails 6

coolo:
Unfortunately 6.0.1 is buggy and breaks minitest


coolo:
hakiri runs brakeman 3.7.2 while 4.7 is released to support rails 6. the hakiri warnings sound like spam :(

coolo:
that hakiri service seems to be dead. So brakeman within circleci ?

vpereira:
We are just closing it and handling it on #9210. Thank you for your help @coolo  :+1:

PR #8778, opened 2019-11-26
Display an icon based on the state of a check

### LINKED ISSUE #8741: Indicate failed and succeeded checks with more than text
The checks in the staging project page are just a plain list and it takes quite some parsing to realize if some failed - add colours and/or icons please.
 
screencapture-build-opensuse-org-staging-workflows-136-staging-projects-openSUSE-Leap-15-2-Staging-F-2019-11-19-07_05_32

dmarcoux:
Closes #8741

Preview:
issue-8741-2nd

It's not directly related to the issue, but I'm wondering if we should sort the checks by state... Any thoughts?

### DIFF
--- a/src/api/app/assets/stylesheets/webui/staging-workflow.scss
+++ b/src/api/app/assets/stylesheets/webui/staging-workflow.scss
@@ -140,3 +140,8 @@ ul .table-list-group-item {
     }
   }
 }
+
+li.missing-check {
+  margin-left: 1.25rem;
+  list-style: disc;
+}
--- a/src/api/app/helpers/webui/staging/project_helper.rb
+++ b/src/api/app/helpers/webui/staging/project_helper.rb
@@ -6,4 +6,10 @@ module Webui::Staging::ProjectHelper
     return 'fa-check-circle text-primary' if checks.all?(&:success?)
     'fa-exclamation-circle text-danger'
   end
+
+  def icon_for_check(check)
+    return 'fa-check-circle text-primary' if check.success?
+    return 'fa-eye text-info' if check.pending?
+    'fa-exclamation-circle text-danger'
+  end
 end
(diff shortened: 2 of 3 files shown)


DavidKang:
@dmarcoux, what about if we replace the bullets with the icons?

coolo:
About your question: I think grouping by state makes sense as on the #index page we only show failed checks, having the same set on the #show page may make it easier to recognize.

dmarcoux:
> @dmarcoux, what about if we replace the bullets with the icons?

Here's a preview of how this would look. I removed the list `<ul>` has `padding-left 0; list-style: none`. I added some missing checks to have a better overview of the changes.

issue-8741-david-suggestion

coolo:
now missing checks miss some bullet, no?

coolo:
(can we have a transparent icon there? :)

dmarcoux:
> now missing checks miss some bullet, no?

This is intended. It's always the same text for a missing check (beside the name of the check), so I felt like adding an icon doesn't bring value here.

> (can we have a transparent icon there? :)

Hum... what do you mean? Is it to align the missing checks with the checks?

coolo:
> Hum... what do you mean? Is it to align the missing checks with the checks?

right.

dmarcoux:
I just pushed a new set of changes and updated the screenshot in the PR's description. Let me know what you think.

coolo:
the icon for pending is a little larger than the rest, but I guess that's only noticable as your check names are very similiar. So I'm good

dmarcoux:
> the icon for pending is a little larger than the rest, but I guess that's only noticable as your check names are very similiar. So I'm good

I also noticed that when making sure this aligns well. It's exactly 2 pixels larger. I ended up leaving this as it is.

Issue #8762, opened 2019-11-22
Package is used by following packages as devel package


saraycp:
It happens in a delayed job: delayed_job#ProjectCreateAutoCleanupRequests

It seems the error is raised when trying to remove a package when another another package is pointing to it as devel package

[See in Errbit.](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5dccedc003a7750008ba9106)

PR #8782, opened 2019-11-26
Prevent an exception to be raised in auto cleanup

saraycp:
The performance of the auto cleanup delayed job failed because the check of dependencies raised an exception. So the delayed job was retried again and again.

The exception was caused by a project with a develpackage inside, that couldn't be removed.

Now, instead of raising an exception, the dependencies are correctly checked.

Fixes: #8762

### DIFF
--- a/src/api/spec/factories/attribs.rb
+++ b/src/api/spec/factories/attribs.rb
@@ -48,6 +48,12 @@ FactoryBot.define do
       values { [build(:attrib_value, value: Faker::Lorem.sentence)] }
     end
 
+    factory :auto_cleanup_attrib do
+      attrib_type { AttribType.find_by_namespace_and_name!('OBS', 'AutoCleanup') }
+
+      values { [build(:attrib_value, value: 14)] }
+    end
+
     factory :attrib_with_default_value do
       attrib_type { create(:attrib_type_with_default_value) }
     end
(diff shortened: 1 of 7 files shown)


src/api/spec/jobs/project_create_auto_cleanup_requests_job_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  describe '#perform' do
+    let(:admin) { create(:admin_user, login: 'Admin') }
+    let(:project) { create(:project, name: 'ProjectA') }
+    let(:attribute) { create(:auto_cleanup_attrib, project: project) }
+
+    subject { ProjectCreateAutoCleanupRequests.new.perform }
+
+    before do
+      allow(::Configuration).to receive(:cleanup_after_days).and_return(3)
```
coolo: so there was a timebomb built in. It worked 3 days - and now fails :)

https://github.com/openSUSE/open-build-service/issues/8826

src/api/spec/factories/attribs.rb
```diff
@@ -48,6 +48,12 @@
       values { [build(:attrib_value, value: Faker::Lorem.sentence)] }
     end
 
+    factory :auto_cleanup_attrib do
+      attrib_type { AttribType.find_by_namespace_and_name!('OBS', 'AutoCleanup') }
+
+      values { [build(:attrib_value, value: 14)] }
```
coolo: the bug is there - OBS:Autocleanup contains a timestamp. And '14' was parsed on 2019-11-28 as a date in the past (11-14) while today it's parsed as a date in the future (12-14)

As such the test case is flickering - just very slow, it will work again on 15th :)

Issue #8784, opened 2019-11-26
Staging managers can't unstage unless also target maintainers


coolo:
All I wanted to do (as staging manger) was unstage request:

`DELETE https://api.suse.de/staging/SUSE:SLE-15-SP2:GA/staging_projects/SUSE:SLE-15-SP2:GA:Staging:C/staged_requests`

but it failed with:
```
<status code="update_staging/workflow_not_authorized">
   <summary>You are not authorized to update this Staging/workflow.</summary>
</status>

coolo:
So this is specific to `Staging::StagedRequestsController#destroy` - the policy call there is wrong.

PR #8789, opened 2019-11-27
Use correct policies in StagedRequestController

DavidKang:
The users of the manager group could unstaged the requests if they
weren't added as maintainers. Now we use a specific policy for staged an
unstaged a request.

Fix: #8784

### DIFF
--- a/src/api/app/policies/staging/staged_request_policy.rb
+++ b/src/api/app/policies/staging/staged_request_policy.rb
@@ -1,0 +1,11 @@
+class Staging::StagedRequestPolicy < ApplicationPolicy
+  def create?
+    group = record.managers_group
+    user.groups_users.where(group: group).exists? ||
+      ProjectPolicy.new(user, record.project).update?
+  end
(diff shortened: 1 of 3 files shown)


src/api/app/policies/staging/staged_request_policy.rb
```diff
@@ -0,0 +1,15 @@
+class Staging::StagedRequestPolicy < ApplicationPolicy
+  def initialize(user, record)
+    super(user, record, user_optional: true)
```
coolo: If user is optional, then you will have to catch it being nil

coolo: but I think I would just leave out the constructor and rely on `require_login` in the controller to catch it

DavidKang: @coolo, could you review again?  please :smiley_cat:

coolo:
thx for reminding

Issue #8787, opened 2019-11-27
Staging dashboard redundant project name


lnussel:
Looking at the staging dashboard of 15.2 the screen is pretty busy and certainly costs brain power to digest. More than the old one.
![staging](https://user-images.githubusercontent.com/722232/69709243-59590d00-10fd-11ea-8c41-b237077a6e58.png)

There is a headline stating "Staging for openSUSE Leap 15.2". But then all rows in the table have openSUSE:Leap:15.2 again. Looks redundant, eg "A" should be sufficient.

DavidKang:
@lnussel, we added the full name of the project on purpose, because it is possible to add as `staging project` other projects that are not a subproject of `openSUSE:Leap:15.2`.

coolo:
Can you make a special case for the staging projects as we use it? How about a staging nick name for every staging?

hennevogel:
Well we can certainly cover the case of subprojects and use only `Staging:A`. I doubt it helps much with the busy factor of this page...

coolo:
Well, I think it does a big difference: 
 
Before:
![screencapture-file-space-Downloads-Staging-SP2-html-2019-11-27-15_52_53](https://user-images.githubusercontent.com/1067203/69733800-38f57680-112e-11ea-8114-4e31eb509432.png)

With taken the old approach:
![screencapture-file-suse-coolo-Downloads-Staging-for-SUSE-SLE-15-SP2-GA-SUSE-Internal-OBS-Instance-html-2019-11-27-15_52_22](https://user-images.githubusercontent.com/1067203/69733808-401c8480-112e-11ea-9db0-a1c64cb518a1.png)

lnussel:
absolutely :+1:

hennevogel:
We can just display the tile/description and you can set there whatever you want.

Issue #8680, opened 2019-11-08
Strange sort order in staging workflow list


coolo:
I can't say what the projects in https://build-test.opensuse.org/staging_workflows/1 are sorted after.

creation time? I have B, adi:5, adi:6, ... adi:223, adi:1, adi:3, adi:4, ..., adi:220

The Factory staging dashboard sorts by state, but I guess sorting by name would be acceptable (as it would allow async loading of the state)

DimStar77:
I am an active user of the 'sort by state' - as it gives me a very quick overview of where what kind of work is needed (e.h. pending reviews only, failures, which I need to bring to the attention of maintainers or possibly can solve by merging stagings)

so, if in any way possible, keeping the sorting we currently have, that would be much appreciated

fcrozat:
For SLE, since we don't use adi these days, I'm wondering if we should sort by state (Green on top, Red at the bottom) or just sort alphabetically, since we don't have a lot of stagings anyway.
Both have pro and cons, I'm not sure which one is best.

If sorting by state, it should bring the "actionnable staging" at the top, so, Green THEN Review THEN Red (or somthing like repo-checker or packagelist red), then TEST/Building.

gyr:
I like the current sort by state, for the reason already mentioned in previous comments, it groups the stagings by action that needs to be taken.

dmarcoux:
To answer @coolo's question, the staging projects on the staging workflow page are not sorted. In the end, it is the same as sorting by creation time since we obviously never change the creation time.

coolo:
At least the sorting order is stable atm - so let's call it a feature to give it a proper sorting

PR #8803, opened 2019-11-28
WIP: Add categories for staging projects

coolo:
We need to seperate letter projects from adi projects and shorten how they are displayed.

Current state of affairs:

screencapture-f102-suse-de-3000-staging-workflows-136-2019-11-29-07_04_45

### DIFF
--- a/src/api/app/views/webui/staging/workflows/_overall_state.html.haml
+++ b/src/api/app/views/webui/staging/workflows/_overall_state.html.haml
@@ -1,6 +1,6 @@
 %span.badge{ class: "state-#{staging_project.overall_state}" }
   %span.badge.badge-light
-    = link_to(staging_project.name, staging_workflow_staging_project_path(staging_workflow, staging_project.name))
+    = link_to(staging_project.staging_nick, staging_workflow_staging_project_path(staging_workflow, staging_project.name))
   %span
     %br
     = staging_project.overall_state
--- a/src/api/app/views/webui/staging/workflows/_overall_state.html.haml
+++ b/src/api/app/views/webui/staging/workflows/_overall_state.html.haml
@@ -1,8 +1,9 @@
-%span.badge{ class: "state-#{staging_project.overall_state}" }
-  %span.badge.badge-light
+%div{ class: "overall-state staging-project state-#{staging_project.overall_state}" }
+  .letter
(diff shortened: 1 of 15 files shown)


coolo:
The goal is to fix #8787 and #8680

src/api/app/views/webui/staging/workflows/_overall_state.html.haml
```diff
@@ -1,8 +1,9 @@
...
-  %span
-    %br
-    = staging_project.overall_state
-    %br
+%div{ class: "overall-state staging-project state-#{staging_project.overall_state}" }
+  .letter
+    = link_to(staging_project.staging_nick, staging_workflow_staging_project_path(staging_workflow, staging_project.name))
+  .state
+    - state = staging_project.overall_state
+    - state = 'broken' if state == :unacceptable
```
coolo: Possibly we can rename the whole state - 'empty' and 'acceptable' are the only states that are interesting to the client.

src/api/app/views/webui/staging/workflows/show.html.haml
```diff
@@ -16,14 +16,27 @@
...
+
+- content_for :ready_function do
+  :plain
+    $('.staging-projects-datatable').DataTable({
+      responsive: true,
+      paging: false,
+      ordering: false,
+      searching: false,
+      info: false,
+      "columnDefs": [ { "width": "6em", "targets": 0 } ]
```
coolo: here I need help. somehow you need to define a width or it will take randomly 30%. the package list is defined to take 50%, but I didn't find a setting so datatable will take the minimum required for the first column

But with 6em the staging workflows we have in production look good, but if you don't define categories and use random project names, it may require a lot more than that (and will line break in funny ways)

coolo:
This is ready for review, but not necessarly ready for merging - there is no way to add the categories atm beside the rails console.

I would add a default category for `.*:Staging:(?<nick>.*)` so the user has something to edit - but having routes to edit them could be a 2nd PR. But I don't want to invest there before the path was agreed upon.

coolo:
Any opinions on the direction I took here from the team?

hennevogel:
> The goal is to fix https://github.com/openSUSE/open-build-service/issues/8787 and https://github.com/openSUSE/open-build-service/issues/8680

Both are fixed in another way..

PR #8804, opened 2019-11-28
Add schema of the GET /source/<project>/<package>/_history route

dcermak:
The schema of the `GET /source/<project>/<package>/_history` route is not defined. I have tried my best to guess it from some API calls.

@adrianschroeter: I need the info what the attribute `vrev` of the `<revision>` element is.

### DIFF
--- a/docs/api/api/revisionlist.rng
+++ b/docs/api/api/revisionlist.rng
@@ -30,21 +30,28 @@
 
           <attribute name="vrev">
             <a:documentation>
-              !FIXME!
+              The vrev is maintained by the server and ensures a strictly monotone
+              increasing number for a given version. It consists of the version parsed
+              from the the build description and the checkin counter. The checkin counter
+              gets reset to zero if the new version did not exist yet. Together with the
+              build counter this forms the version-release of the resulting binary.
             </a:documentation>
             <data type="nonNegativeInteger"/>
           </attribute>
 
           <element ns="" name="srcmd5">
             <a:documentation>
-              MD5 hash of this revision
+              MD5 hash of this revision as identifier. Note that this is the sum over the
+              unexpanded sources (in case a link exists).
             </a:documentation>
             <text/>
           </element>
 
           <element ns="" name="version">
             <a:documentation>
-              Version (e.g. 1.5.0) of the package at this revision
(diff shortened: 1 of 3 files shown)


docs/api/api/revisionlist.rng
```diff
@@ -30,21 +30,27 @@
...
+              unexpanded sources (in case a link exists).
             </a:documentation>
             <text/>
           </element>
 
           <element ns="" name="version">
             <a:documentation>
-              Version (e.g. 1.5.0) of the package at this revision
+              Version (e.g. 1.5.0) of the package at this revision parsed from the source.
+              It may not be set if the package is linking to another package instance.
```
dcermak: Does this mean that this element is optional or that it might be empty?

docs/api/api/revisionlist.rng
```diff
@@ -0,0 +1,96 @@
...
+            </a:documentation>
+            <data type="nonNegativeInteger"/>
+          </attribute>
+
+          <attribute name="vrev">
+            <a:documentation>
+              The vrev is maintained by the server and ensures a strictly monotone
+              increasing number for a given version. It consists of the version parsed
+              from the the build description and the checkin counter. The checkin counter
+              get reset to zero if the new version did not exist yet. Together with the
```
dcermak:
```suggestion
              gets reset to zero if the new version did not exist yet. Together with the
```

docs/api/api/revisionlist.rng
```diff
@@ -0,0 +1,96 @@
...
+              build counter you get the version-release of the resulting binary.
```
dcermak:
```suggestion
              build counter this forms the version-release of the resulting binary.
```

docs/api/api/revisionlist.rng
adrianschroeter: it would be set to "unknown"

adrianschroeter: yes, it would be better to be optional, but ... history ;)

dcermak: I've added that information to the docs.

adrianschroeter:
the rspec error was unrelated

PR #8808, opened 2019-11-28
Align db/structure with what dev containers spit out

coolo:
Mariadb changed the output format since switching to code15


coolo:
fun, circleci still runs on code12

PR #8810, opened 2019-11-28
WIP: Restrict to add only one bugowner per project

### LINKED ISSUE #3011: Disallow multiple bugowner via "add_role" request
Issue/Feature description
====================
add_role request actions allow to add an additional bugowner in a package or project. bugowners should be only a single definition. (We could handle multiple, but then we would need a definition of the default owner).

set_bugowner is the supposed action to set the bugowner (removes all entities and adds the new one).

Noticed by openSUSE distro people @coolo and Leonardo 

Expected result
============

We should either fail if add_role for bugowner is used and target has already a defintion.

Alternative solution would be to change the action type automatically to set_bugowner.

How to Reproduce
===============
1. Create a package with bugowner role
2. create an "add_role" with "bugowner" role request
3. multiple entries exists in target after accept.

DavidKang:
It doesn't make sense to have more than one bugowner per project. Also
in the WebUI we replace the select box by radio buttons, because there
are only two options available.

This restriction is also applied for the project's meta via WebUI and
API.

Screenshot_2020-04-02 home Admin

Fix #3011


coolo:
> If the project already has a bugowner assigned we only show the option to request a maintainer role.

This one is most likely not what we want, because *replacing* the bugowner is still a thing. We have set_bugowner requests for that - so the webui wouldn't support this at all after this?

DavidKang:
> This one is most likely not what we want, because replacing the bugowner is still a thing. We have set_bugowner requests for that - so the webui wouldn't support this at all after this?

Yesterday I realized that I was removing this possibility :slightly_smiling_face:. In the end, I will use the `set_bugowner` method. Thanks :smile_cat:.

hennevogel:
@DavidKang what's up here?

DavidKang:
@hennevogel, I didn't had time for it, but I would like to finish it.

DavidKang:
I don't have time right now, and this PR is too old. I will close it for now.

PR #8815, opened 2019-11-29
Pick less offensive colours for staged requests

coolo:
screencapture-f102-suse-de-3000-staging-workflows-136-2019-11-29-12_02_09

### DIFF
--- a/src/api/app/assets/stylesheets/webui/staging-workflow.scss
+++ b/src/api/app/assets/stylesheets/webui/staging-workflow.scss
@@ -1,3 +1,7 @@
+$request-ready-color: $success;
+$request-review-color: $secondary;
+$request-obsolete-color: $dark;
+
 .legend {
   ul {
     @extend .ml-2;
@@ -9,6 +13,16 @@
         height: 1rem;
         @extend .border;
         @extend .border-dark;
+      }
+
+      .request-ready {
+        background-color: $request-ready-color;
+      }
+      .request-review {
+        background-color: $request-review-color;
+      }
+      .request-obsolete {
+        background-color: $request-obsolete-color;
       }
     }
   }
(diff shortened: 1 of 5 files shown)


coolo:
Oh? There is a discussion forum 'with the development team' other than here where the development happens?

coolo:
btw, if you sticked to your colour patterns, we might not have this discussion. Part of the problem is that you mix bootstrap background colours with your own foreground colours.

vpereira:
> btw, if you sticked to your colour patterns, we might not have this discussion. Part of the problem is that you mix bootstrap background colours with your own foreground colours.

that's why it's import to talk with the development team. We had a sprint about colors x usability, involving customers with some level of vision impairment and we used the tool https://wave.webaim.org/, to guide us through the process.

coolo:
I doubt that sprint involved the staging dashboard's colours.

coolo:
the review app bot is broken :(

eduardoj:
@coolo:

> the review app bot is broken :(

Fixed!

vpereira:
@coolo could you please validate this PR with WAVE (https://wave.webaim.org) ?

coolo:
I fixed the review apps to have a proper staging workflow by default, resulting in:

https://wave.webaim.org/report#/https://obs-reviewlab.opensuse.org/coolo-fix_colours/staging_workflows/openSUSE:Factory

Warnings about low contrast.

coolo:
It should be noted that this is not a regression as the master dashboard has white on orange. But I'll try to adopt.

coolo:
I picked more boring colours - and there are no more complaints.

vpereira:
yes, no issues at WAVE and the colours are ok.

PR #8827, opened 2019-12-02
Store a timestamp of the past in the cleanup attribute

### LINKED ISSUE #8826: Auto cleanup requests job broken (test suite)
```
Failure/Error: expect(project.target_of_bs_request_actions.where(type: 'delete').count).to eq(1)

  expected: 1
       got: 0

  (compared using ==)
./spec/jobs/project_create_auto_cleanup_requests_job_spec.rb:23:in `block (4 levels) in <top (required)>'

### PR DESCRIPTION
**Pull request by coolo**
Fixes #8826

### DIFF
--- a/src/api/spec/factories/attribs.rb
+++ b/src/api/spec/factories/attribs.rb
@@ -51,7 +51,7 @@ FactoryBot.define do
     factory :auto_cleanup_attrib do
       attrib_type { AttribType.find_by_namespace_and_name!('OBS', 'AutoCleanup') }
 
-      values { [build(:attrib_value, value: 14)] }
+      values { [build(:attrib_value, value: (Time.now - 14.days).to_s)] }
     end
 
     factory :attrib_with_default_value do

### CODE REVIEW

File: src/api/spec/factories/attribs.rb [hunk 1]
```diff
@@ -51,7 +51,7 @@
     factory :auto_cleanup_attrib do
       attrib_type { AttribType.find_by_namespace_and_name!('OBS', 'AutoCleanup') }
 
-      values { [build(:attrib_value, value: 14)] }
+      values { [build(:attrib_value, value: (Time.now - 14.days).to_s)] }
```
eduardoj: This line changes what is stored in the database and in the backend. But the test only needs what is stored in the database.

Ideally the cassettes should be also updated. There is still a '14' stored in there ([here](https://github.com/openSUSE/open-build-service/blob/b740a8ba99ec7404f914ba3a0820c919dd4f2fd1/src/api/spec/cassettes/ProjectCreateAutoCleanupRequests/_perform/with_project_without_dependencies/sets_a_deletion_request_on_the_project.yml#L50) and [here](https://github.com/openSUSE/open-build-service/blob/b740a8ba99ec7404f914ba3a0820c919dd4f2fd1/src/api/spec/cassettes/ProjectCreateAutoCleanupRequests/_perform/with_devel_package_inside_the_project/does_not_create_a_deletion_request.yml#L50)). But this works.

eduardoj:
Thanks for reporting and for the pull request.

src/api/spec/factories/attribs.rb
coolo: You can update the cassette every 5 minutes - because as soon it's updated, the timestamp written is already outdated

PR #8832, opened 2019-12-03
Refactor code related to issue models 

eduardoj:
We realized that the issues' models had unused code. In order to have a better codebase we decided to cleanup the unused code and refactor.

Have a look at the commits' descriptions for a more detailed explanation.

### DIFF
--- a/src/api/app/models/issue.rb
+++ b/src/api/app/models/issue.rb
@@ -49,10 +49,6 @@ class Issue < ApplicationRecord
     end
   end
 
-  def self.valid_name?(tracker, name)
-    tracker.show_label_for(name).match?(tracker.regex)
-  end
-
   def fetch_issues
     IssueTrackerFetchIssuesJob.perform_later(issue_tracker.id)
   end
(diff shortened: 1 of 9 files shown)


src/api/app/models/issue.rb
```diff
@@ -4,6 +4,7 @@ class Issue < ApplicationRecord
   class NotFoundError < APIError
     setup 'issue_not_found', 404, 'Issue not found'
   end
+  class InvalidName < APIError; end
```
vpereira: please move all the error classes to `app/models/issue/errors.rb` as we have for the others?

coolo: a concern for 4 lines? Isn't that a little over the top?

vpereira: all errors classes. Can be more in the future. It's matter of code organization and structure.

eduardoj: Done. Issue errors were moved to its own class. Thanks!

PR #8835, opened 2019-12-04
Move BsRequest error classes to its own module

vpereira:
Refactor the errors classes from `BsRequest`  to a related specific error module


coolo:
IMO the exceptions from app/models/bs_request_permission_check.rb belong in the same namespace

PR #8837, opened 2019-12-04
Add specs for ChannelBinary and related factories

vpereira:
Add specs for the model `ChannelBinary`

### DIFF
--- a/src/api/spec/models/channel_binary_spec.rb
+++ b/src/api/spec/models/channel_binary_spec.rb
@@ -1,0 +1,34 @@
+require 'rails_helper'
+
+RSpec.describe ChannelBinary, vcr: true do
+  let(:user) { create(:confirmed_user, login: 'tux') }
+  let(:project) { create(:project, name: 'projectX') }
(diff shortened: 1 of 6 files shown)


src/api/spec/models/channel_binary_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  describe '.find_by_project_and_package' do
+    let!(:channel_binary) do
+      create(:channel_binary, name: 'foo', project: project, architecture: repository.architectures.first,
+                              repository: repository, channel_binary_list: channel_binary_list, package: package.name)
+    end
+    let!(:maintenance_project) { create(:maintenance_project, target_project: project) }
+
+    it { expect(ChannelBinary.find_by_project_and_package(project.name, package.name)).not_to be_empty }
+    it { expect(ChannelBinary.find_by_project_and_package(project.name, package.name)).to include(channel_binary) }
```
eduardoj: This line checking from `include(channel_binary)` implies that it is not empty. So I would remove the line 22.

Same case for lines 31 and 32.

PR #8838, opened 2019-12-04
Move errors from BsRequestPermissionCheck to BsRequest::Errors

vpereira:
Follow-up from #8835


src/api/app/models/bs_request_permission_check.rb
```diff
@@ -1,41 +1,5 @@
...
-    setup 403
-  end
-
-  class ReviewChangeStateNoPermission < APIError
-    setup 403
-  end
-
-  class ReviewNotSpecified < APIError
-  end
+  include BsRequest: Errors
```
coolo: that extra space looks wrong

PR #8841, opened 2019-12-05
Use brakeman in circleci instead of relying on dead hakiri

coolo:
Background: https://brakemanscanner.org/blog/2018/06/28/brakeman-has-been-acquired-by-synopsys/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,6 +233,10 @@ jobs:
     steps:
       - attach_workspace:
          at: .
+      - run:
+          name: Patch brakeman in
+          command: |
+            sed -i -e 's,# BRAKEMAN,gem "brakeman",' src/api/Gemfile
       - run: *install_dependencies
       - run:
           name: Run brakeman
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -226,6 +226,21 @@ jobs:
           path: src/api/coverage_results
           destination: raw_coverage
 
+  brakeman:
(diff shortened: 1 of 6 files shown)


coolo:
(one more prerequire for rails 6)

coolo:
brakeman complains here that we allow checks to have urls. And there is indeed a potential problem as we don't do any verification on these urls. Shall we?

Conan-Kudo:
I don't think you guys can pull Brakeman in as a dependency. It's been licensed CC-BY-NC-SA since June 2018. At least, this will make things ugly for packaging OBS in Fedora, where this license is banned.

coolo:
There interest to use a real and uptodate security scanner is small anyway - as you can see in the lack of any comments in the last 2 months.

krauselukas:
@coolo sorry that this didn't got more attention, there is definitely interest in getting something that is more up-to-date and works with Rails 6. One solution regarding the licence problem, could be to not ship the gem and only integrate it to our circle ci setup (we only need it available in this repo right?). Otherwise we could also think about using another application, maybe from the GitHub Marketplace, which also gets the job done, or even uses brakeman in the background. I will also do some research, since I continued with the rails 6 upgrade last week.

krauselukas:
I added a little chapter to our wiki to show how to handle false positives with brakeman for now :)
https://github.com/openSUSE/open-build-service/wiki/Development-Environment-Tips-&-Tricks#brakeman

PR #8848, opened 2019-12-06
Refactor BsRequestAction#create_expand_package

vpereira:
Refactor `create_expand_package` little bit and adding specs to it

### DIFF
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -400,6 +400,7 @@ class BsRequestAction < ApplicationRecord
     Package.source_path(self.source_project, self.source_package)
   end
 
+  # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/CyclomaticComplexity
   # rubocop:disable Metrics/PerceivedComplexity
   def create_expand_package(packages, opts = {})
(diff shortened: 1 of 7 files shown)


src/api/app/models/project.rb
```diff
@@ -1550,6 +1550,14 @@ def self.very_important_projects_with_categories
     end
   end
 
+  def build_results
+    project_state.get_elements("/resultlist/result[@project='#{name}'')]")
+  end
+
+  def project_state
+    REXML::Document.new(Backend::Api::BuildResults::Status.version_releases(name))
```
coolo: avoid REXML in any refactorings - we have 2 other xml parsers in use. This is even more important if it's unclear what is returned here in the code calling.

vpereira: Nokogiri is much faster, however to refactor the code increased some rubocop metrics score (Metrics/AbcSize, Metrics/CyclomaticComplexity and Metrics/PerceivedComplexity).. but I believe other changes that I will do in the code will be able to reduce it again

src/api/app/models/bs_request_action.rb
```diff
@@ -482,20 +480,20 @@ def create_expand_package(packages, opts = {})
...
                                     'did not finish the build yet'
           end
 
           # all versrel are the same
           versrel[repo] ||= {}
-          result.get_elements('status').each do |status|
+          result.search('status').each do |status|
             package = status.attributes['package']
-            vrel = status.attributes['versrel']
+            vrel = status.attributes['versrel'].value if status.attributes['versrel']
```
coolo: No point in having the assignment conditional if you `next` in the next line if it didn't happen

src/api/app/models/bs_request_action.rb
```diff
@@ -400,6 +400,7 @@ def source_cleanup
     Package.source_path(self.source_project, self.source_package)
   end
 
+  # rubocop:disable  Metrics/AbcSize
```
coolo: if you need to touch it once more, remove that space please

vpereira: done!

PR #8858, opened 2019-12-10
Prevent duplicate issues in patchinfo

### LINKED ISSUE #8798: Patchinfo allows duplicated issues
# Issue Description
Creating or editing a patchinfo allows the same issue (kde#123456 i. e.) to be introduced more than once.
Screenshot from 2019-11-27 14-09-57

Expected Result
============
Entering an already entered issued should not be allowed.

How to Reproduce
===============
1. Create a project.
1. Click on 'Create Patchinfo'.
1. Scroll down to 'Issues'.
1. Enter `kde#123456` and click on 'Add issues'.
1. Repeat the previous step a couple of times.
1. Click on the 'Save' button.

Either the issue should not be allowed to be entered after clicking in 'Add issues', or clicking in 'Save' should result in an error, with some message like 'Issue duplicated'.

Further Information
===============

dmarcoux:
Closes #8798

Submitting at once the same issue multiple times by separating its instance with commas does remove the duplicates which were entered by the user. However, as described in the issue, it was possible to add the same issue multiple times by adding them one by one. These changes prevent this.

Nothing changed visually... so no screenshots.

### DIFF
--- a/src/api/app/assets/javascripts/webui/patchinfo.js
+++ b/src/api/app/assets/javascripts/webui/patchinfo.js
@@ -20,6 +20,19 @@ function addIssuesAjaxBefore() {
 
     issues = $.unique(issues.replace(/ /g, '').split(','));
 
+    var currentIssues = $('[name="patchinfo[issueid][]"]').map(function(_, element) {
+      var issueId = element.id;
+      if (issueId.startsWith('issueid_cve')) {
+        return issueId.replace('issueid_cve_', '');
+      } else {
+        return issueId.replace('issueid_', '').replace('_', '#');
+      }
+    }).toArray();
+
+    issues = issues.filter(function(issue) {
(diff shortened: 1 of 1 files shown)


coolo:
Let's see if obs-bot works for you, it failed on all other PRs so far :(

dmarcoux:
> Let's see if obs-bot works for you, it failed on all other PRs so far :(

Nope, still broken... Hum... what is happening?

hennevogel:
>> Let's see if obs-bot works for you, it failed on all other PRs so far :(

> Nope, still broken... Hum... what is happening?

image_from_ios

vpereira:
I'm not sure if in some scenarios the security team won't want to have repeated CVEs in the patchinfo metada, but I guess they don't use this entrypoint anyway.

PR #8861, opened 2019-12-11
Extract code blocks from create_expand_package to separated methods

vpereira:
Codeclimate pointed out that the method `BsRequestAction#create_expand_package` is the most complext method. So to try to reduce it, I extracted some code blocks to methods

### DIFF
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -400,7 +400,6 @@ class BsRequestAction < ApplicationRecord
     Package.source_path(self.source_project, self.source_package)
   end
 
-  # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/CyclomaticComplexity
   # rubocop:disable Metrics/PerceivedComplexity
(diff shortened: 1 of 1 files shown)


src/api/app/models/bs_request_action.rb
```diff
@@ -443,21 +442,21 @@ def create_expand_package(packages, opts = {})
...
-      elsif tprj.try(:is_maintenance_incident?) && is_maintenance_release?
-        # fallback, how can we get rid of it?
-        data = Directory.hashed(project: tprj.name, package: ltpkg)
-        e = data['linkinfo']
-        tpkg = e['package'] if e
-      else
-        # we need to get rid of it again ...
-        tpkg = tpkg.gsub(/#{Regexp.escape(suffix)}$/, '') # strip distro specific extension
-      end
+      tpkg = if target_package
```
danidoni: Can you extract this big if also?

danidoni: We agreed on not doing it in this PR

PR #8862, opened 2019-12-11
Remove FullTextIndexJob and its queue

dmarcoux:
This job should have been removed in the PR #8558 since the indexing is now in real-time.


dmarcoux:
> There is more to it. We also have a [DJ queue](https://github.com/openSUSE/open-build-service/blob/master/dist/systemd/obs-delayedjob-queue-sphinx-indexing.service) that you should drop

Removed!

PR #8908, opened 2020-01-03
Show action column when user is logged in

### LINKED ISSUE #8681: [production][] n is undefined
threw this 369 times
<pre>
n is undefined
</pre>
[https://build.opensuse.org/package/users/X11:Unity/vala-panel-appmenu](https://build.opensuse.org/package/users/X11:Unity/vala-panel-appmenu)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5d69314124569a000813582d/problems/5da986477fe6e0000de07ab7)

eduardoj:
Datatable javascript was raising an exception because the action column was missing the column header.

The action column is now rendered when a user is logged in.

Fixes #8681.

### DIFF
--- a/src/api/app/views/webui/shared/user_or_groups_roles/_index.html.haml
+++ b/src/api/app/views/webui/shared/user_or_groups_roles/_index.html.haml
@@ -21,7 +21,8 @@
           %td Username
           - roles.each do |role|
             %td= role.title.capitalize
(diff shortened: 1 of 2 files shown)


eduardoj:
OBS Package Build are frozen. Merging.

PR #8909, opened 2020-01-03
Prevent staging a request with errors

### LINKED ISSUE #8730: It shouldn't be possible to stage the same package name twice

eduardoj:
In #8736 an error was thrown when a second request for the same target package wanted to be added, preventing the creation of the linked package. But the request was still staged.

Now both, the linked package and the staged request are not created.

Fixes #8730.

### DIFF
--- a/src/api/app/models/staging/staged_requests.rb
+++ b/src/api/app/models/staging/staged_requests.rb
@@ -5,23 +5,7 @@ class Staging::StagedRequests
   def create
     add_requests_excluded_errors
     add_request_not_found_errors
-    requests.each do |request|
-      bs_request_action = request.bs_request_actions.first
-      if bs_request_action.is_submit?
-        link_package(bs_request_action)
-      end
-      # TODO: implement delete requests
-      ProjectLogEntry.create!(
-        project: staging_project,
-        user_name: user_login,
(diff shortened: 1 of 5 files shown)


DavidKang:
@danidoni, could you review it again? :smile_cat:

src/api/app/models/staging/staged_requests.rb
```diff
@@ -195,4 +172,32 @@ def link_xml(opts = {})
...
+      package_name: bs_request_action.target_package
+    )
+    staging_project.staged_requests << request
+    add_review_for_staged_request(request)
+  end
+
+  def package_can_be_linked?(request, target_package)
+    return true unless Package.find_by(project: staging_project, name: target_package)
+
+    errors << "Can't stage request '#{request.number}': package '#{target_package}' already exists in '#{staging_project}'."
```
danidoni: This is changing the state of the error from inside a method that looks like a checker method... I would put the error filling part outside. You know, I could call `package_can_be_linked?` several times and this would be changing the object state...

PR #8914, opened 2020-01-06
Refactor create expand package

vpereira:
Follow-up PR of #8861 and #8848 

Cleaned the method `create_expand_package` little bit leading to a reduction in the `Metrics/PerceivedComplexity` score (From 73 to 65) and `Metrics/CyclomaticComplexity` score (from 67 to 61). Both are still far from threshold (45) allowed by Rubocop.

### DIFF
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -463,7 +463,7 @@ class BsRequestAction < ApplicationRecord
 
       # overwrite target if defined
       tprj = Project.get_by_name(target_project) if target_project
-      raise UnknownTargetProject, 'target project does not exist' unless tprj || is_maintenance_release?
+      raise UnknownTargetProject unless tprj || is_maintenance_release?
 
       # do not allow release requests without binaries
       if is_maintenance_release? && pkg.is_patchinfo? && data && !opts[:ignore_build_state]
@@ -548,7 +548,7 @@ class BsRequestAction < ApplicationRecord
             new_packages << pkg
             next
           elsif !is_maintenance_incident? && !is_submit?
-            raise UnknownTargetPackage, 'target package does not exist'
+            raise UnknownTargetPackage
           end
         end
       end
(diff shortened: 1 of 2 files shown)


src/api/app/models/bs_request_action/errors.rb
```diff
@@ -7,7 +7,10 @@ class RemoteSource < APIError
     setup 'remote_source', 400, 'No support for auto expanding from remote instance. You need to submit a full specified request in that case.'
   end
   class RemoteTarget < APIError; end
-  class InvalidReleaseTarget < APIError; end
+  class InvalidReleaseTarget < APIError
+    setup 'invalid_release_target', 400, 'Can not release to a maintenance incident project'
+  end
+  class MultipleReleaseTargets < APIError; setup 'Multiple release target projects are not supported'; end
```
eduardoj: A `400` error code should be added here too. Isn't it?

vpereira: just if it's used in the controllers, right?

src/api/app/models/bs_request_action.rb
```diff
@@ -798,6 +781,23 @@ def set_sourceupdate_default(user)
...
+      # skip excluded patchinfos
+      status = pkg.project.project_state.search("/resultlist/result[@repository='#{repo.name}' and @arch='#{firstarch.name}']").first
+
+      next if status && (s = status.search("status[@package='#{pkg.name}']").first) && s.attributes['code'].value == 'excluded'
+
+      raise BuildNotFinished, "patchinfo #{pkg.name} is broken" if s.attributes['code'].value == 'broken'
+
+      check_maintenance_release(pkg, repo, firstarch)
+    end
+    true
```
eduardoj: This always returns true. It should only return true if it finds a patchinfo.

vpereira: not true. did you check the original code? it will assign true if no exception is raised.

src/api/app/models/bs_request_action/errors.rb
eduardoj: So then in line 11 for `InvalidReleaseTarget`, the `400` can also be removed? Or is it another use case?

src/api/app/models/bs_request_action.rb
eduardoj: Fixed. Thanks!

PR #8918, opened 2020-01-07
add jira issue kind

adrianschroeter:
(no fetcher yet)


adrianschroeter:
not used in api code yet, but needed to be able to write correct issue kind in updateinfo

PR #8926, opened 2020-01-09
cleanup project meta template

adrianschroeter:
- Ubuntu repo seems not to be used anymore (was used for travis-ci)
- update to current versions
- move to the right place (dist is for packaged files)


adrianschroeter:
grr

PR #8930, opened 2020-01-09
Fix the API docs involving the directory schema

dcermak:
- The directory element can contain the `vrev` attribute
- The example for the `/source/<project>/<package>` route is a `_meta` and not the list of files as would be expected.

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -507,7 +507,7 @@ Parameters:
   repository: string, for view=info: parse buildinfo for this repository, optinal
   product: string, limit the product view to a given product
 
-XmlResult: package directory.xsd
+XmlResult: directory_filelist directory.xsd
 
 
 GET /source/<project>/<package>/_meta
--- a/docs/api/api/directory.xsd
+++ b/docs/api/api/directory.xsd
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" 
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
   elementFormDefault="qualified">
 
   <xs:annotation>
@@ -20,7 +20,9 @@
         <xs:element ref="linkinfo" minOccurs="0" maxOccurs="unbounded"/>
         <xs:element ref="serviceinfo" minOccurs="0" maxOccurs="unbounded"/>
       </xs:sequence>
+      <xs:attribute name="name" type="xs:string"/>
       <xs:attribute name="rev" type="xs:string"/>
+      <xs:attribute name="vrev" type="xs:string"/>
       <xs:attribute name="srcmd5" type="xs:string"/>
       <xs:attribute name="count" type="xs:int"/>
     </xs:complexType>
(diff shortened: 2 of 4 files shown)


adrianschroeter:
Thanks, but you introduced new directory_filelist.xml without new schema file. This is the reason for the testcase failure. 

We could either extend current directory.xml or we introduce an additional schema. directory_source_listing might be a good name here.

dcermak:
> Thanks, but you introduced new directory_filelist.xml without new schema file. This is the reason for the testcase failure.
> 
> We could either extend current directory.xml or we introduce an additional schema. directory_source_listing might be a good name here.

Not sure if introducing another schema makes sense here, as the `directory.xsd` **is** the correct schema already, the example is just plain wrong.

adrianschroeter:
On Dienstag, 14. Januar 2020, 09:24:55 CET Dan Čermák wrote:
> > Thanks, but you introduced new directory_filelist.xml without new schema file. This is the reason for the testcase failure.
> > 
> > We could either extend current directory.xml or we introduce an additional schema. directory_source_listing might be a good name here.
> 
> Not sure if introducing another schema makes sense here, as the `directory.xsd` **is** the correct schema already, the example is just plain wrong.

The example is complete valid for a different route (/source/$project).

-- 

dcermak:
Adrian Schröter <notifications@github.com> writes:

> On Dienstag, 14. Januar 2020, 09:24:55 CET Dan Čermák wrote:
>> > Thanks, but you introduced new directory_filelist.xml without new schema file. This is the reason for the testcase failure.
>> > 
>> > We could either extend current directory.xml or we introduce an additional schema. directory_source_listing might be a good name here.
>> 
>> Not sure if introducing another schema makes sense here, as the `directory.xsd` **is** the correct schema already, the example is just plain wrong.
>
> The example is complete valid for a different route (/source/$project).
>

Would a symlink work for the schema? Then at least the two directory.xsd
schemas couldn't diverge.

dcermak:
Adrian Schröter <notifications@github.com> writes:

> On Dienstag, 14. Januar 2020, 09:24:55 CET Dan Čermák wrote:
>> Not sure if introducing another schema makes sense here, as the `directory.xsd` **is** the correct schema already, the example is just plain wrong.
>
> The example is complete valid for a different route (/source/$project).

Oh and the example is also invalid for that route, /source/$project
gives you the list of packages in $project, but the example is a
package's _meta. That doesn't even conform to the directory.xsd schema.

dcermak:
@adrianschroeter How should we proceed here?

adrianschroeter:
Just extend the existing one, because it is used for all routes.

Or split it into multiple, and create a special one for source directory listings, but that needs also code updates or the test suite will continue to fail.

hennevogel:
@dcermak whats the status here?

dcermak:
> @dcermak whats the status here?

I honestly don't know how to fix the test failures: the current example for the source listing is the wrong one and extending it won't work (the example is a package _meta file which cannot obey the directory.xsd schema).

hennevogel:
> I honestly don't know how to fix the test failures

Just say so, @openSUSE/obs-frontend  are here to help :-)

dcermak:
Hm, no idea what the issue is here. @openSUSE/obs-frontend help?

Issue #7376, opened 2019-04-09
Requests manually accepted need to be handled


coolo:
If an user accepts a staged request manually, it needs to be treated as untracked request.

coolo:
because it's no longer part of the staging project, but the sources are in it. User errors happen

coolo:
If you remove accepted requests from staging projects automatically, you can simplify #7373 actually - just accepting will do.

coolo:
I think the easier and more logic approach is to forbid accepting requests in staging. Because in the long run we also want to get rid of the by_project review to stop requests from being ready.

DimStar77:
> I think the easier and more logic approach is to forbid accepting requests in staging. Because in the long run we also want to get rid of the by_project review to stop requests from being ready.

As an RM I sometimes NEED to accept a request; would I 'just' need to unselect it first before accepting (that would be acceptable - in the old dashboard I basically also had to unselect it to not mess up things)

coolo:
Well, the options are:
- auto-unselecting on accept
- require manual unselect to make it force acceptable

Basically: we require force accept on backlog and on normal accept on ready - but we don't require force accept on staged requests, because you can always put them back into the backlog first. And it's the smaller mess IMO

fcrozat:
Fine with me

coolo:
as #8941 is merged, this should be resolved

PR #8941, opened 2020-01-13
Handle Stage Request manually accepted

DavidKang:


vpereira:
@coolo Any idea if we should do something similar for other final states, like revoked?

coolo:
And about your question: revoked requests *usually* get a request superseding it, so we keep them in the staging project for a while - the staging project will be unacceptable in this case.

vpereira:
> So this is the opposite of the agreement in #7376, right?

hey @coolo ,

Reading through the comments, the only agreed thing was that it should be handled. Otherwise there was some discussion going on where none of the developers were involved, right? We will revisit what was proposed there, but in the meantime, any disadvantage on this PR approach?

coolo:
You implemented *the other* option from https://github.com/openSUSE/open-build-service/issues/7376#issuecomment-559467756 - it's not so bad, but the downside is that you unselect without doing all the actions that the client would do on unselect.

vpereira:
From the linked issue, I read the following:

```
Well, the options are:

auto-unselecting on accept
require manual unselect to make it force acceptable
Basically: we require force accept on backlog and on normal accept on ready - but we don't require force accept on staged requests, because you can always put them back into the backlog first. And it's the smaller mess IMO
```

which we understood as "auto-unselecting" was actually the first option and the less intrusive. 

> You implemented _the other_ option from [#7376 (comment)](https://github.com/openSUSE/open-build-service/issues/7376#issuecomment-559467756) - it's not so bad, but the downside is that you unselect without doing all the actions that the client would do on unselect.

Any other action from the client that comes to your mind that we should implement here?

coolo:
atm this is mainly limited to removing adi projects if they become empty. I don't think that should be done on the API level. And it's optional. 

But you don't know what will happen in the future, that's why I proposed to block accepting (in 2 comments before the one you quoted). The options were only listed to show Dimstar what the alternative is.

PR #8945, opened 2020-01-14
[api] fix /trigger/release to release all projects

seife:
Triggering package release with a token via /trigger/release path did
only release one project, because .any? returns after first successful
iteration.

See https://lists.opensuse.org/opensuse-buildservice/2020-01/msg00048.html for the gory details.

### DIFF
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -28,7 +28,12 @@ class TriggerController < ApplicationController
   def release
     raise NoPermissionForPackage.setup('no_permission', 403, "no permission for package #{@pkg} in project #{@pkg.project}") unless policy(@pkg).update?
 
-    matched_repo = @pkg.project.repositories.includes(:release_targets).any? { |repo| matched_repo?(repo) }
+    matched_repo = false
+    @pkg.project.repositories.includes(:release_targets).each do |repo|
+      if matched_repo?(repo) then
+        matched_repo = true
+      end
+    end
 
     raise NoPermissionForPackage.setup('not_found', 404, "no repository from #{@pkg.project} could get released") unless matched_repo
 


src/api/app/controllers/trigger_controller.rb
```diff
@@ -28,7 +28,12 @@ def rebuild
   def release
     raise NoPermissionForPackage.setup('no_permission', 403, "no permission for package #{@pkg} in project #{@pkg.project}") unless policy(@pkg).update?
 
-    matched_repo = @pkg.project.repositories.includes(:release_targets).any? { |repo| matched_repo?(repo) }
+    matched_repo = FALSE
+    @pkg.project.repositories.includes(:release_targets).each do |repo|
+      if matched_repo?(repo) then
+        matched_repo = TRUE
+      end
```
hennevogel: I fear this will solve your problem but introduce another one. If you have more than one repo with `release_targets` you have to make sure they don't overwrite `matched_repo` which we base the next `raise` on.

This needs more thought.

src/api/app/controllers/trigger_controller.rb
```diff
@@ -28,7 +28,12 @@ def rebuild
   def release
     raise NoPermissionForPackage.setup('no_permission', 403, "no permission for package #{@pkg} in project #{@pkg.project}") unless policy(@pkg).update?
 
-    matched_repo = @pkg.project.repositories.includes(:release_targets).any? { |repo| matched_repo?(repo) }
+    matched_repo = FALSE
+    @pkg.project.repositories.includes(:release_targets).each do |repo|
+      if matched_repo?(repo) then
+        matched_repo = TRUE
```
hennevogel: do not use FALSE/TRUE constants, they are deprecated. You can just use `false` and `true`.

seife: ok, updating...

src/api/app/controllers/trigger_controller.rb
seife: I think it is still correct. The raise is "no repo from project can be released". If at least one can be released, then it will not raise.
IOW: this is no change to the previous behaviour. Just that before not everything was released.

Or am I understanding this totally wrong? can release() be called multiple times in parallel? is this code not thread safe or something like that? only then could `matched_repo = false` overwrite anything. As you might guess, I know nothing about RoR (and am not really interested in learning ;-)

hennevogel: > If at least one can be released, then it will not raise.

No as long as the last repo will release it will not raise. No matter what happened before.

seife: Yes, and that's totally what was intended IMHO. If at least one repo can be released, then it will not raise.
Before it was the same. But it was "if at least and at most one repo was released, it did not raise". It did not raise even though 6 of my 7 configured repos were not released ;-)

Certainly a more detailed return code might be useful, but this is not in scope for this fix and a major change in behaviour.

adrianschroeter: seife is right regarding the wanted behaviour.

The code is IMHO quite bad atm since matched_repo? is doing the active release. It looks more like a passive method to me and makes the code hard to read.

seife: Well, I tried to do the minimal-invasive fix, with the additional constraint that I know not much about RoR. I think this is good enough for backporting to 2.10 (will apply it to my build anyway).

A rewrite would certainly be good, don't ask how I cursed about the one line I removed and replaces until I understood what it does ;-)

hennevogel: > If at least one repo can be released, then it will not raise.

That is not true. Both `matched_repo?` and `MaintenanceHelper.release_package` will raise if you don't have permissions....

hennevogel: Let me see if I can push something half-sane...

seife: > That is not true. Both `matched_repo?` and `MaintenanceHelper.release_package` will raise if you don't have permissions....

But is that really different to the previous broken behaviour?

I think if you do not have permissions on the "source" repo, then you won't get there because f line 29.
If you do not have permissions on the release target, it will raise later.

It did this with the old code, I know this because I added the permissions one after another. Do you think it will now not raise if you have permissions on some of the repos but not on others?

But is this even possible?
Ah yes. Different repos could have different target projects with different permissions. What would "osc release" do then? And how can we model this in the simple "release via token trigger" API?

hennevogel: giphy-downsized-large

hennevogel: BTW `matched_repo?` has the same problem...

hennevogel: `osc release` is [`SourceController.package_command_release`](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/source_controller.rb#L1016), doesn't handle this either....

hennevogel:
See #8947 for another idea...

hennevogel:
#8947 is merged now

PR #8946, opened 2020-01-14
Unstage declined requests

### LINKED ISSUE #8757: Unstaging a declined request needs to change reviews
The client can't fix that through the API as you can't add/remove reviews on declined requests, but it still makes sense to align the state of the request as long as we have reviews *and* staged_request relationship.

Blocks #8754

DavidKang:
When a request is on a declined state, we can't touch the reviews. For that, we reopen it temporally to manage the reviews.

Fixes #8757

TODO:
- [x] Tests

### DIFF
--- a/src/api/spec/controllers/staging/staged_requests_controller_spec.rb
+++ b/src/api/spec/controllers/staging/staged_requests_controller_spec.rb
@@ -424,25 +424,60 @@ RSpec.describe Staging::StagedRequestsController do
         before do
           login user
           bs_request.change_state(newstate: 'declined', user: user.login, comment: 'Fake comment')
-          delete :destroy, params: { staging_workflow_project: staging_workflow.project.name, format: :xml },
-                           body: "<requests><request id='#{bs_request.number}'/></requests>"
         end
 
-        it { expect(response).to have_http_status(:success) }
-        it { expect(staging_project.packages).to be_empty }
-        it { expect(staging_project.staged_requests).to be_empty }
-        it { expect(bs_request.reviews.where(by_group: group.title, state: 'new')).to be_present }
-        it { expect(bs_request.reviews.where(by_project: staging_project.name, state: 'new')).to be_empty }
-
-        context 'when the declined request was unstaged and reopened' do
+        context 'when is unstaged by the same user' do
           before do
-            login user
-            bs_request.change_state(newstate: 'new', user: user.login, comment: 'Fake comment')
+            delete :destroy, params: { staging_workflow_project: staging_workflow.project.name, format: :xml },
+                             body: "<requests><request id='#{bs_request.number}'/></requests>"
           end
 
-          it { expect(bs_request.state).to eq(:review) }
+          it { expect(response).to have_http_status(:success) }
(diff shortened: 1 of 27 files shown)


DimStar77:
not sure I'd want that - at least in ADIs I often have a group of stuff relying on each other, and a decline means I need a fixed submission incoming - that should go to the same adi group the original thing was in. In case of 'Project based Stagings (think full KDE Stack, Full Qt update) it's the same thing - one request being declined means the full group is only ready when a fix is incoming for that group

vpereira:
Hey @DimStar77 any idea how to handle it? Another solution would be to avoid decline a request if it's staged (look the issue that we are trying to fix #8757) but then it wouldn't be in line with the #8941

DimStar77:
> Another solution would be to avoid decline a request if it's staged

Clearly not an option - legal as well as the regular source review must be able to decline a request irrespective of it being staged or not.

coolo:
declining and accepting are clearly 2 different beasts. For a declined request, the manual **unstage** needs to manipulate the reviews. The declining action itself should not touch it

vpereira:
> Clearly not an option - legal as well as the regular source review must be able to decline a request irrespective of it being staged or not.
@DimStar77 

I don't really know how you are doing staging, but I was assuming that you just stage a request after those checks are done, no?

coolo:
https://stephan.kulow.org/mermaid.html happens all in parallel

DavidKang:
Well I push a second commit where the reviews are handled during the unstage process. The only thing is that I reopen the request to be able manage the reviews. I reopen the request, because I don't like the option to skip validations on reviews.

DavidKang:
> https://stephan.kulow.org/mermaid.html happens all in parallel

thanks

coolo:
That will do. Thanks

src/api/spec/controllers/staging/staged_requests_controller_spec.rb
```diff
@@ -418,6 +419,21 @@
         it { expect(staging_project.packages).to be_empty }
         it { expect(staging_project.staged_requests).to be_empty }
       end
+
+      context 'with declined request' do
+        before do
+          login user
+          bs_request.change_state(newstate: 'declined', user: user.login, comment: 'Fake comment')
+          delete :destroy, params: { staging_workflow_project: staging_workflow.project.name, format: :xml },
+                           body: "<requests><request id='#{bs_request.number}'/></requests>"
```
coolo: see the "test case" in the feature request. The interesting part is if reopening the request leaves the staging project untouched and the request will appear in backlog.

I expect this to be the case, but it would be good to have this future safe.

DavidKang: @coolo, what do you mean with untouched?
I extended the test, reopening a declined request that was unstaged.

coolo: untouched == it stays empty after reopen if it was empty before. Just as your test verifies

coolo:
By this the 'untracked' state should be impossible to reach.

PR #8947, opened 2020-01-14
Fixes releasing for multiple targets via trigger

hennevogel:
Triggering package release via a token did only release one project,
because .any? does not continue after first successful evaluation.

Also remove superflous permission check for target, release_package does that
already. And throw a 404 if the project has no release targets.

Co-authored-by: Stefan Seyfried <stefan.seyfried@sap.com>

Supersedes #8945

### DIFF
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -3,7 +3,7 @@ class TriggerController < ApplicationController
   validate_action release: { method: :post, response: :status }
   validate_action runservice: { method: :post, response: :status }
 
-  before_action :require_project_param, only: [:release]
+  before_action :disallow_project_param, only: [:release]
 
   before_action :extract_auth_from_request, :validate_auth_token, :require_valid_token
   #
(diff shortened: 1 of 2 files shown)


src/api/app/controllers/trigger_controller.rb
```diff
@@ -28,9 +28,12 @@ def rebuild
...
     raise NoPermissionForPackage.setup('no_permission', 403, "no permission for package #{@pkg} in project #{@pkg.project}") unless policy(@pkg).update?
 
-    matched_repo = @pkg.project.repositories.includes(:release_targets).any? { |repo| matched_repo?(repo) }
+    manual_release_targets = @pkg.project.release_targets.where(trigger: 'manual')
+    raise NoPermissionForPackage.setup('not_found', 404, "#{@pkg.project} has no release targets that are triggered manually") unless manual_release_targets.any?
 
-    raise NoPermissionForPackage.setup('not_found', 404, "no repository from #{@pkg.project} could get released") unless matched_repo
+    manual_release_targets.each do |release_target|
+      release_package(@pkg, release_target.target_repository, @pkg.release_target_name, release_target.repository, nil, nil, nil, true)
+    end
```
seife: Will this still report an error if no package could be released? (not that I really care, but it would be a change in behaviour)

hennevogel: the first package that encounters an exception (can only be missing permissions or something unforseen) will stop the action. see the test.

PR #8950, opened 2020-01-16
[backend] fix package lock check on wipe event

adrianschroeter:


mlschroe:
Nope, that's wrong.

adrianschroeter:
fixing just the wipe check bug for now.

Will create an independend issue for the lock behaviour inconsistency between backend and api.

Issue #8951, opened 2020-01-16
Url for staging tab is for creating a "new" staging


fcrozat:
Url used for staging is (for IBS, with latest deployement):
https://build.suse.de/staging_workflows/new?project=SUSE%3ASLE-15-SP2%3AGA

(and it causes a redirect to https://build.suse.de/staging_workflows/SUSE:SLE-15-SP2:GA )

it should probably a direct link to staging, if it was already created.

coolo:
it's worse if you're not maintainer. The tab will link to .../9 which gives you a 404

coolo:
My problem is a different one - and fixed by 8952 (and monkey patched). @fcrozat 's problem is still there

vpereira:
there is no bug here@fcrozat the link is working as expected, even though I agree that the route is kind of awkward. What do you think @hennevogel ?

coolo:
"working as expected"? who expected the link to go to new if there is already staging workflow?

vpereira:
it's doing what was proposed in the PR the code path is correct, so the code is working as expected. The solution was took here https://github.com/openSUSE/open-build-service/pull/6109, so it is working as expected because it behaves exactly was the developer wanted to. It's not a bug.

coolo:
the [comment](https://github.com/openSUSE/open-build-service/pull/8437#issuecomment-536552829) of @dmarcoux at least is in line with my own expectation. It's not a problem (and clearly not a regression), but IMO worth changing.

vpereira:
@coolo I'm not saying the opposite either https://github.com/openSUSE/open-build-service/issues/8951#issuecomment-575130827 my point is that it's exactly what you said. It's not a problem or bug. But it's ugly. Why this differentiation is important?  Priority.

coolo:
sure.

PR #8952, opened 2020-01-16
Point Staging link to the Staging Workflow project

vpereira:
Change the link in the `_tabs.html.haml` to use the project name instead of id in the url


coolo:
Fixes #8951 ?

PR #8957, opened 2020-01-16
Sort staging projects in staging workflow show page

### LINKED ISSUE #8680: Strange sort order in staging workflow list

eduardoj:
Use the same sort method that was used before for obs_factory.

Fixes #8680.

Screenshot from 2020-01-16 15-12-45

### DIFF
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -42,6 +42,7 @@ class Webui::Staging::WorkflowsController < Webui::WebuiController
   def show
     @project = @staging_workflow.project
     @staging_projects = @staging_workflow.staging_projects.includes(:staged_requests).reject { |project| project.overall_state == :empty }
+                                         .sort_by! { |project| project_weight(project) }
     @unassigned_requests = @staging_workflow.unassigned_requests.first(5)
     @more_unassigned_requests = @staging_workflow.unassigned_requests.count - @unassigned_requests.size
     @ready_requests = @staging_workflow.ready_requests.first(5)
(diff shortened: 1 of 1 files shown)


DimStar77:
Without something like https://github.com/openSUSE/open-build-service/pull/8803 this new sort is quite 'bad'

in Factory I now have a list of all acceptable adi stagings, followed by adi stagings needing reviews in front of all letter stagings.

DimStar77:
on the old obs_factory dashboard this worked as we had two tabs - Letter and adi - so very similar to the new category concept proposed by @coolo

nilxam:
> Without something like #8803 this new sort is quite 'bad'
> 
> in Factory I now have a list of all acceptable adi stagings, followed by adi stagings needing reviews in front of all letter stagings.

I'd argue the same...

vpereira:
@fcrozat could you please tell us how it looks from your side?

coolo:
SLE only has one category of stagings, so the lack of categories doesn't hurt there

hennevogel:
we can certainly talk about something like categories. but not here. I'll approach you peepz for a meeting.

vpereira:
> SLE only has one category of stagings, so the lack of categories doesn't hurt there

sure, but would be good to have a feedback from him too. If the PR didn't make sense for both use cases that we have, we should rething the whole sorting.

fcrozat:
> @fcrozat could you please tell us how it looks from your side?

looks ok.

PR #8961, opened 2020-01-16
 Show 0% of tests done where there are no checks 

### LINKED ISSUE #8794: Staging workflow show testing 100% when no openQA status is available
When staging_workflow is waiting for openQA status " openqa:gnome (Expected - Waiting for status to be reported) ", the main page shows "testing 100 % " instead of "0%"

eduardoj:
The [original piece of code](https://github.com/openSUSE/open-build-service/blob/675c1238d2c467f6720f9dd8e78f22c3841454d7/src/api/app/presenters/obs_factory/staging_project_presenter.rb#L133) returns 0% when there are no checks.

Also refactor a little piece of code.

Fixes #8794.

### DIFF
--- a/src/api/app/helpers/webui/staging/workflow_helper.rb
+++ b/src/api/app/helpers/webui/staging/workflow_helper.rb
@@ -23,11 +23,12 @@ module Webui::Staging::WorkflowHelper
   end
 
   def testing_progress(staging_project)
-    notdone = staging_project.checks.pending.size
-    allchecks = staging_project.checks.size + staging_project.missing_checks.size
+    return 0 if staging_project.checks.size.zero?
 
-    return 100 if allchecks == 0
-    100 - notdone * 100 / allchecks
+    not_done = staging_project.checks.pending.size + staging_project.missing_checks.size
+    all_checks = staging_project.checks.size + staging_project.missing_checks.size
+
+    100 - not_done * 100 / all_checks
   end
 
   def progress(staging_project)


src/api/app/helpers/webui/staging/workflow_helper.rb
```diff
@@ -23,10 +23,11 @@ def review_progress(staging_project)
   end
 
   def testing_progress(staging_project)
-    notdone = staging_project.checks.pending.size
     allchecks = staging_project.checks.size + staging_project.missing_checks.size
+    return 0 if allchecks == 0
+
+    notdone = staging_project.checks.pending.size
 
-    return 100 if allchecks == 0
```
vpereira: What's going happen if `notdone =  0`?

src/api/app/helpers/webui/staging/workflow_helper.rb
```diff
@@ -23,10 +23,11 @@ def review_progress(staging_project)
...
     100 - notdone * 100 / allchecks
```
vpereira: shouldn't we add parenthesis here to make sure that we are getting the right results?

src/api/app/helpers/webui/staging/workflow_helper.rb
```diff
@@ -23,10 +23,11 @@ def review_progress(staging_project)
   end
 
   def testing_progress(staging_project)
-    notdone = staging_project.checks.pending.size
     allchecks = staging_project.checks.size + staging_project.missing_checks.size
+    return 0 if allchecks == 0
```
vpereira: why are you returning 0 instead of 100 now?

src/api/app/helpers/webui/staging/workflow_helper.rb
eduardoj: It will be returned 0.

src/api/app/helpers/webui/staging/workflow_helper.rb
eduardoj: Adding the parenthesis where exactly? The formula looks good to me as it is.

src/api/app/helpers/webui/staging/workflow_helper.rb
coolo: testing_progress shouldn't even be called in that situation - as the staging project would be in a different overall state. So the returned value doesn't really make a difference - I guess. But I agree with @vpereira - 100 makes sense *in case the function is called*.

If you check a2b7c8392a79c63cc88d3dc09d2e2786eb5ae16c when testing_percentage was introduced, the 0 was about the situation when there were no openqa jobs. And *that* situation meant openQA did not yet start to test. But checks that are not there, should be missing - to be relevant.

I hope that clarifies a little.

src/api/app/helpers/webui/staging/workflow_helper.rb
coolo: I would rather think 100 - 0, aka 100. And this is exactly what the bug is all about.

In the situation described in the ticket, there are 0 checks and 1 missing check. That means allchecks = 1, notdone = 0.

vpereira: as @coolo said, If I'm not wrong, it will be like:

```100 - notdone * 100 / allchecks``` is like

```100 - ((notdone * 100) / allchecks)```

which will be in the end ```100 - 0 / allchecks```, no?

src/api/app/helpers/webui/staging/workflow_helper.rb
vpereira: if you want to have ```100 - ((notdone * 100) / allchecks)``` then you are right.. I thought you wanted to have ```((100 - notdone) * 100) / allchecks```

src/api/app/helpers/webui/staging/workflow_helper.rb
eduardoj: You are both right. :man_facepalming: I will fix this.

coolo: ¡Cálculo de puntos antes del cálculo de líneas! :smile:

eduardoj:
I pushed another version of the method. It returns `0` when there are no checks. Please @vpereira, @coolo, have another look.

coolo:
3 authors, 2 reviewers for 3 lines - math is hard! :)

PR #8965, opened 2020-01-17
[backend] allow publishing of empty repository

adrianschroeter:
First part as a proposal. Usecase are esp :Update repositories which must be there from the beging but can be empty. Otherwise QA checks for the distro my fail.

This rule allows the publisher to create a repo neverthless even though there are now binaries. Second part would be to trigger the publish from scheduler if that flag is set.

I think @bugfinder will love that feature and be done just by setting up the publish rule in prjconf altogether with the rest....

### DIFF
--- a/src/backend/BSSched/PublishRepo.pm
+++ b/src/backend/BSSched/PublishRepo.pm
@@ -172,6 +172,9 @@ sub prpfinished {
     $seen_binary = {};
   }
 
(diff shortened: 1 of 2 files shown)


mlschroe:
Note also that doing it this way has the undesired side effect that projects building against :Update will also publish empty repositories. So you'll need to use it inside some `%if` that checks '%_project'.

PR #8971, opened 2020-01-17
Fix staging tab link

### LINKED ISSUE #8951: Url for staging tab is for creating a "new" staging

hennevogel:
It should only point to new_staging_workflow_path if there isn't a staging
workflow.

Fixes #8951

### DIFF
--- a/src/api/app/views/webui/project/_tabs.html.haml
+++ b/src/api/app/views/webui/project/_tabs.html.haml
@@ -36,14 +36,15 @@
       - if project.staging_project?
         %li.nav-item
           = tab_link('Staging', staging_workflow_path(project.staging_workflow.project), active)
-      - elsif policy(project.staging || Staging::Workflow.new(project: project)).create?
-        %li.nav-item
-          = tab_link('Staging', new_staging_workflow_path(project: project), active)
-      - elsif project.staging.persisted?
-        -# If Staging::Workflow.new(...) is executed above, project.staging will be a new object.
+      - elsif project.staging && project.staging.persisted?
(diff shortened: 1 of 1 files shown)


src/api/app/views/webui/project/_tabs.html.haml
```diff
@@ -36,14 +36,15 @@
...
-        %li.nav-item
-          = tab_link('Staging', new_staging_workflow_path(project: project), active)
-      - elsif project.staging.persisted?
-        -# If Staging::Workflow.new(...) is executed above, project.staging will be a new object.
+      - elsif project.staging && project.staging.persisted?
+        -# If Staging::Workflow.new(...) is executed below, project.staging will be a new object.
            It will be persisted if the project is really a staging workflow project.
         %li.nav-item
           = tab_link('Staging', staging_workflow_path(project), active)
+      - elsif policy(project.staging || Staging::Workflow.new(project: project)).create?
```
DavidKang: `project.staging`, shouldn't be necessary, right?

hennevogel: True

hennevogel: removed

PR #8977, opened 2020-01-20
split addjobhist into two parts

sjamgade:
the job history hash can be reused for storing stats.


mlschroe:
Wouldn't it be easier to call makejobhist() from addjobhist()?

sjamgade:
I am splitting it to reuse the returned jobhist hash for storing hash

PR #8978, opened 2020-01-20
Add Query objects to Package model

vpereira:
In the quest of reduce the complexity and size of controllers and models, we introduced long time ago services objects, which reside in `app/services`, now following the design patterns, we are introducing the query object pattern, which the implementation will be located under `app/queries`.

The idea is to extract complex SQL queries or scopes to it's own PORO class, where we can better manage and test, removing some code which the implementation isn't so relevant to the models or controllers. 

The reasoning behind it is to use the battle tested design patterns to try to keep the base sane and avoid 700+ lines of models/controllers. 

References: 
https://www.martinfowler.com/eaaCatalog/queryObject.html
https://mkdev.me/en/posts/how-to-use-query-objects-to-refactor-rails-sql-queries 
https://github.com/infinum/rails-handbook/blob/master/Design%20Patterns/Query%20Objects.md

### DIFF
--- a/src/api/app/queries/packages_finder.rb
+++ b/src/api/app/queries/packages_finder.rb
@@ -1,0 +1,84 @@
+class PackagesFinder
+  def initialize(relation = Package.all)
+    @relation = relation
+  end
+
+  def by_package_and_project(package, project)
+    @relation.where(name: package, projects: { name: project }).includes(:project)
+  end
+
+  def find_by_attribute_type(attrib_type, package = nil)
+    # One sql statement is faster than a ruby loop
+    # attribute match in package or project
+    sql = build_sql_find_by_attribute(package)
+    finder_arguments = if package
+                         [sql, attrib_type.id.to_s, attrib_type.id.to_s, package]
+                       else
+                         [sql, attrib_type.id.to_s, attrib_type.id.to_s]
+                       end
+    find_package(finder_arguments)
+  end
+
+  def find_by_attribute_type_and_value(attrib_type, value, package = nil)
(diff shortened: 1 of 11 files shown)


src/api/app/queries/packages_finder.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def dirty_backend_packages
+    @relation.joins('left outer join backend_packages on backend_packages.package_id = packages.id').
+      where('backend_packages.package_id is null')
+  end
+
+  private
+
+  def find_package(args)
+    Package.find_by_sql(args).keep_if { |p| Package.check_access?(p) }
```
DavidKang: Could you change `p` by something more explicit like `package`?

src/api/spec/queries/packages_finder_spec.rb
```diff
@@ -0,0 +1,80 @@
+require 'rails_helper'
+
+RSpec.describe PackagesFinder, vcr: true do
+  describe '#by_package_and_project' do
+    let(:package_finder) { PackagesFinder.new }
```
DavidKang: `package_finder` is not used in the test. the same happens in the others `describe`

src/api/app/queries/packages_finder.rb
DavidKang: we talked off-line.

vpereira: I don't see a problem to have single letter for some short block. i.e no issue with `0.upto(10).each { |i| i * 10 }` so if it's ok I would keep like that.

PR #8979, opened 2020-01-20
Introduces full staging workflow factory

hennevogel:
A project with a staging, staging projects and staged requests

### DIFF
--- a/src/api/spec/factories/staging_workflow.rb
+++ b/src/api/spec/factories/staging_workflow.rb
@@ -20,7 +20,36 @@ FactoryBot.define do
         letters = [*'A'..'Z']
         new_staging_projects_count.times do |index|
           letter = letters[index + staging_workflow.staging_projects.count]
-          staging_workflow.staging_projects << create(:staging_project, name: "#{project.name}:Staging:#{letter}")
+          staging_workflow.staging_projects << create(:staging_project, name: "#{staging_workflow.project.name}:Staging:#{letter}", maintainer: staging_workflow.managers_group)
+        end
+      end
+
+      factory :staging_workflow_with_sources do
+        transient do
+          bs_request_count { 10 }
+        end
+
+        after(:create) do |staging_workflow, evaluator|
+          target_packages = create_list(:package, evaluator.bs_request_count, project: staging_workflow.project)
+          source_project = create(:project)
+          target_packages.each do |package|
+            create(:package_with_file, name: package.name, project: source_project)
+          end
+          target_packages.each do |target_package|
+            create(:bs_request_with_submit_action, target_project: staging_workflow.project,
+                                                   target_package: target_package,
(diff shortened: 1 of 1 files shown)


src/api/spec/factories/staging_workflow.rb
```diff
@@ -20,9 +20,42 @@
...
+            create(:package_with_file, name: package.name, project: source_project)
+          end
+          target_packages.each do |target_package|
+            create(:bs_request_with_submit_action, target_project: staging_workflow.project,
+                                                   target_package: target_package,
+                                                   source_package: source_project.packages.find_by(name: target_package.name)
+                                                   )
+          end
+          target_packages.each_slice(staging_workflow.staging_projects.count) do |packages|
+            old_session = User.session
```
coolo: that looks a lot like `staging_workflow.managers_group.users.first.run_as &block`

src/api/spec/factories/staging_workflow.rb
```diff
@@ -20,9 +20,42 @@
...
+            User.session = staging_workflow.managers_group.users.first
+            packages.each do | package |
+              ::Staging::StagedRequests.new(
+                request_numbers: [package.target_of_bs_requests.first.number],
+                staging_workflow: staging_workflow,
+                staging_project: staging_workflow.staging_projects.sample,
```
coolo: why slice if you later on take a sample anyway? Sounds like one loop is enough.

src/api/spec/factories/bs_requests.rb
```diff
@@ -101,7 +101,6 @@
     after(:build) do |request, evaluator|
       # Monkeypatch to avoid errors caused by permission checks made
       # in user and bs_request model
-      User.session = evaluator.creating_user
```
coolo: I'm afraid this will break quite some tests. we'll see

src/api/spec/factories/staging_workflow.rb
```diff
@@ -20,9 +20,42 @@
...
+        end
+
+        after(:create) do |staging_workflow, evaluator|
+          target_packages = create_list(:package, evaluator.bs_request_count, project: staging_workflow.project)
+          source_project = create(:project)
+          target_packages.each do |package|
+            create(:package_with_file, name: package.name, project: source_project)
+          end
+          target_packages.each do |target_package|
+            create(:bs_request_with_submit_action, target_project: staging_workflow.project,
```
coolo: depending on your goals here, you may want to spam some reviews in the requests. The way you have it, all projects are building/acceptable. Most likely your packages will end being `excluded` as they have no spec file.

So throwing some reviews and checks into the projects will make it more colourful.

src/api/spec/factories/staging_workflow.rb
hennevogel: Yeah I was playing with distributing it evenly before...

hennevogel:
This is now waiting for #8988

PR #8981, opened 2020-01-21
Remove obs_factory 

vpereira:
Now that Staging workflow is here to stay, we should remove the good old "Factory Dashboard"

@coolo for historical reasons you should merge it :)


vpereira:
at least code climate is happy about it :)

coolo:
I know I said it's no longer used, but I forgot about https://build.opensuse.org/project/dashboard/openSUSE:Factory and
https://build.opensuse.org/project/dashboard/openSUSE:Leap:15.2

and while the latter was only accessed by googlebot yesterday, Factory was looked at by @DimStar77 suprisingly often. So I assume we need to provide an alternative for him.

This alternative shouldn't be within OBS IMO, but it needs time. So please don't remove yet

vpereira:
@DimStar77 any reason why you are still visiting the old dashboard?

vpereira:
@coolo what is the difference between https://build.opensuse.org/staging_workflows/openSUSE:Leap:15.2 and https://build.opensuse.org/project/dashboard/openSUSE:Leap:15.2 ?

vpereira:
> need an alternative project dashboard first (somewhere else)

could be a redirect ?

coolo:
> @coolo what is the difference between https://build.opensuse.org/staging_workflows/openSUSE:Leap:15.2 and https://build.opensuse.org/project/dashboard/openSUSE:Leap:15.2 ?

One is showing the staging projects and one is showing basically everything else.

DimStar77:
> @DimStar77 any reason why you are still visiting the old dashboard?

yes (or I'd not do it; save assumption :) )

It's the only place to give me a one-shot overview of:
* Rebuild state of the Rings and their success state (e.g. ring0 must be rebuilt in order to be able to rebase staging projects)
* Build-Progress of openSUSE:Factory/standard, openSUSE:Factory/images and :Live incl. the number of failures in an easily digestable format)
* Overview of building/testing/published snapshot version (to verify if the things move to where I'd expect them to move)
* openQA result summary (granted, here the most interesting is the fact that it arrived in openQA)

Of course, all the information provided on this dashboard *CAN* be collected in different places - the having a one-stop-place for all the information is worth a lot

vpereira:
@coolo I fail to find a feature request to the staging dashboard. Could you help us out with that?

hennevogel:
> I fail to find a feature request to the staging dashboard. 

We have talked about this in the past. You can not remove this, sorry.

vpereira:
> We have talked about this in the past. You can not remove this, sorry.

How we cannot remove it? We re-implemented the whole staging workflow to get rid of it..

hennevogel:
> How we cannot remove it?

**it** being the [dashboard](https://build.opensuse.org/project/dashboard/openSUSE:Leap:15.2): We explicitly excluded the dashboard from the staging workflow. They have not much to do with staging, they are for a distribution. We are not sure yet how to replace them.

hennevogel:
giphy

PR #8984, opened 2020-01-21
Package model cleanup

vpereira:
Refactor `Package` model improving the ruby dialect

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -534,11 +534,11 @@ class Package < ApplicationRecord
     # issues introduced by local changes
     if is_link?
       query = { cmd: :linkdiff, onlyissues: 1, linkrev: :base, view: :xml }
-      new = parse_issues_xml(query)
-      (issue_change.keys + new.keys).uniq.each do |key|
+      new_issues = parse_issues_xml(query)
+      (issue_change.keys + new_issues.keys).uniq.each do |key|
         issue_change[key] ||= {}
-        issue_change[key].merge!(new[key]) if new[key]
-        issue_change['kept'].delete(new[key]) if issue_change['kept'] && key != 'kept'
+        issue_change[key].merge!(new_issues[key]) if new_issues[key]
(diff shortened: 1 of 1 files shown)


src/api/app/models/package.rb
```diff
@@ -141,14 +141,9 @@ def self.check_cache(project, package, opts)
...
-      prj = project
-    else
-      return if Project.is_remote_project?(project)
-      prj = Project.get_by_name(project)
-    end
-    raise UnknownObjectError, "#{project}/#{package}" unless prj
-    prj
+    return project if project.is_a?(Project)
+    return if Project.is_remote_project?(project)
+    Project.get_by_name(project)
```
vpereira: `Project.get_by_name` returns already a ` UnknownObjectError` if nothing is found. So this custom exception was never raised

src/api/app/models/package.rb
```diff
@@ -224,12 +219,8 @@ def self.exists_by_project_and_name(project, package, opts = {})
   end
 
   def self.exists_on_backend?(package, project)
-    begin
-      answer = Backend::Connection.get(Package.source_path(project, package))
-      return true if answer
-    rescue Backend::Error
-      # ignored
-    end
+    !Backend::Connection.get(Package.source_path(project, package)).nil?
```
vpereira: I was normally used to use the `!!` technique to make every value boolean but Rubocop doesn't like. Instead it recommends https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/DoubleNegation instead

src/api/app/models/package.rb
```diff
@@ -540,11 +525,11 @@ def find_changed_issues
     # issues introduced by local changes
     if is_link?
       query = { cmd: :linkdiff, onlyissues: 1, linkrev: :base, view: :xml }
-      new = parse_issues_xml(query)
```
vpereira: `new` was a bad name for a variable

PR #8985, opened 2020-01-21
Remove not needed/used "responders" ruby gem

eduardoj:
The gem was only used in the tests, as rails 4.2 was introduced.

### DIFF
--- a/src/api/Gemfile
+++ b/src/api/Gemfile
@@ -41,8 +41,6 @@ gem 'sanitize'
 gem 'pundit'
 # for password hashing
 gem 'bcrypt'
-#
-gem 'responders', '~> 3.0'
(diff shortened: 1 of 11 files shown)


src/api/Gemfile
```diff
@@ -41,8 +41,6 @@ gem 'sanitize'
 gem 'pundit'
 # for password hashing
 gem 'bcrypt'
-#
-gem 'responders', '~> 3.0'
```
vpereira: Lets wait until rails-assets.org is back. Since we are removing gem, to make sure that the package is building would be a good thing

eduardoj: rails-assets.org is back. Everything is green.

PR #8986, opened 2020-01-22
Refactor relationship model

vpereira:
Extract query to a query object and refactor methods improving ruby dialect and removing repetitions.

### DIFF
--- a/src/api/app/models/relationship.rb
+++ b/src/api/app/models/relationship.rb
@@ -78,8 +78,7 @@ class Relationship < ApplicationRecord
     # {projecs: [p1,p2], whitelist: { u1: [p1], u2: [p1,p2], u3: [p2] } }
     forbidden_projects = Rails.cache.fetch('forbidden_projects') do
       forbidden_projects_hash = { projects: [], whitelist: {} }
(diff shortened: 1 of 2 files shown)


src/api/app/queries/relationships_finder.rb
```diff
@@ -0,0 +1,16 @@
+class RelationshipsFinder
+  def initialize(relation = Relationship.all)
+    @relation = relation
+  end
+
+  def foo
```
DavidKang: foo?

src/api/app/models/relationship.rb
```diff
@@ -119,6 +121,20 @@ def self.with_groups_and_roles
 
   private
 
+  class << self
+    def add_role(obj, role, opts = {})
+      Relationship::AddRole.new(obj, role, opts).add_role
+    end
+
+    def cache_sequence
```
DavidKang: I think these two methods and the constante `RELATIONSHIP_CACHE_SEQUENCE`, make more sense to be on the 8de0cd6b9095e8376c8d75fc6ed98430dcb93522 commit

PR #8988, opened 2020-01-22
Save the bs_request factory always as the creator

hennevogel:
Do not set the session in a factory. Especially if you don't reset it afterward! This is something you should do inside your test. I ran into weird permission problems in another factory because User.session changed to some random user...


coolo:
763 failures is an improvement over the 788 failures we saw in the other PR. Props!

hennevogel:
boom

PR #8993, opened 2020-01-22
Remove responders also from the lockfile

hennevogel:
Missed in #8985


src/api/Gemfile.lock
```diff
@@ -320,15 +320,12 @@ GEM
     rainbow (3.0.0)
     rake (13.0.1)
     rantly (2.0.0)
-    rbtree3 (0.5.0)
+    rbtree3 (0.6.0)
```
eduardoj: Not needed after rebasing.

PR #9007, opened 2020-01-28
[backend] support milestone publishing using macros

adrianschroeter:
Makes it possible to define project global milestones in prjconf


bugfinder:
are you planning to use the added parameter $config later on ?

PR #9010, opened 2020-01-28
Refactor monitor controller into services

vpereira:
That's something that I had on my stash from last week and I would like to push it before start to work on the new deliverable.

Basically I started some refactoring on the `MonitorController` extracting some code to service objects. There are still some actions (like the `MonitorController#events` or partially the `MonitorController#index`) which should be extracted too but, I will do them in another PR later.

### DIFF
--- a/src/api/app/services/monitor_controller_service/building_information_updater.rb
+++ b/src/api/app/services/monitor_controller_service/building_information_updater.rb
@@ -27,7 +27,7 @@ module MonitorControllerService
     end
 
     def workers_hash(b, delta)
-      b.slice('project', 'repository', 'package', 'arch', 'starttime').merge('delta' => delta)
+      b.slice('project', 'repository', 'package', 'arch', 'starttime').merge('delta' => delta.to_s)
     end
 
     def calculate_delta(starttime)
--- a/src/api/app/services/monitor_controller_service/building_information_updater.rb
+++ b/src/api/app/services/monitor_controller_service/building_information_updater.rb
@@ -23,7 +23,7 @@ module MonitorControllerService
     end
 
     def update_workers
-      @workers = Hash[@worker_status.elements('building').collect { |b| [worker_id(b), workers_hash(b, calculate_delta(b['starttime'].to_i))] }]
+      @workers.merge!(Hash[@worker_status.elements('building').collect { |b| [worker_id(b), workers_hash(b, calculate_delta(b['starttime'].to_i))] }])
     end
 
     def workers_hash(b, delta)
--- a/src/api/app/services/monitor_controller_service/building_information_updater.rb
+++ b/src/api/app/services/monitor_controller_service/building_information_updater.rb
@@ -1,0 +1,50 @@
+module MonitorControllerService
+  class BuildingInformationUpdater
+    attr_accessor :workers
+
+    def initialize
+      @workers = {}
+    end
+
+    def call
+      initialize_workers
+      update_workers
+      self
+    end
+
+    private
+
+    def worker_status
+      @worker_status ||= Xmlhash.parse(WorkerStatus.hidden.to_xml)
+    end
+
+    def initialize_workers
+      @workers = Hash[worker_status.elements('idle').collect { |b| [worker_id(b), {}] }]
+    end
+
+    def update_workers
+      @workers = Hash[@worker_status.elements('building').collect { |b| [worker_id(b), workers_hash(b, calculate_delta(b['starttime'].to_i))] }]
+    end
+
+    def workers_hash(b, delta)
+      b.slice('project', 'repository', 'package', 'arch', 'starttime').merge('delta' => delta)
+    end
+
+    def calculate_delta(starttime)
+      delta = (Time.now - Time.at(starttime)).round
+      delta = 5 if delta < 5
+      delta = max_time if delta > max_time
+      delta = (100 * Math.sin(Math.acos(1 - (Float(delta) / max_time)))).round
+      delta = 100 if delta > 100
+      delta
+    end
+
+    def worker_id(b)
+      b['workerid'].gsub(%r{[:./]}, '_')
+    end
+
+    def max_time
+      4.hours.to_i
+    end
+  end
+end
--- a/src/api/app/services/monitor_controller_service/status_history_fetcher.rb
+++ b/src/api/app/services/monitor_controller_service/status_history_fetcher.rb
@@ -1,0 +1,32 @@
+module MonitorControllerService
+  class StatusHistoryFetcher
+    def initialize(key, range)
+      @key = key
+      @range = user_range(range)
+    end
+
+    def call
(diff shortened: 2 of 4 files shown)


src/api/app/services/monitor_controller_service/building_information_updater.rb
```diff
@@ -0,0 +1,50 @@
...
+      Hash[['project', 'repository', 'package', 'arch', 'starttime'].collect { |k| [k, b[k]] }].merge('delta' => delta)
+    end
+
+    def calculate_delta(starttime)
+      delta = (Time.now - Time.at(starttime)).round
+      delta = 5 if delta < 5
+      delta = max_time if delta > max_time
+      delta = (100 * Math.sin(Math.acos(1 - (Float(delta) / max_time)))).round
+      delta = 100 if delta > 100
+      delta
```
vpereira: something that I noticed is that `delta` is the only value in the xml which isn't encoded as string https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/monitor_controller.rb#L55, we should probably change it. What do you think @hennevogel ?

src/api/spec/services/monitor_controller_service/building_information_updater_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it { expect(subject).not_to be_empty }
+    it { expect(subject).to have_key('build01_1') }
+    it { expect(subject['build01_1']).to be_empty }
+    it { expect(subject).to have_key('build01_2') }
+    it { expect(subject['build01_2']).to have_key('delta') }
+    it { expect(subject['build01_2']['delta']).to eq(100) }
+    it { expect(subject['build01_3']).to have_key('delta') }
+    it { expect(subject['build01_3']['delta']).to eq(48) }
+    it { expect(subject['build01_4']).to have_key('delta') }
+    it { expect(subject['build01_4']['delta']).to eq(66) }
```
vpereira: I just wanted to test different branches from `calculate_delta` private method, that's why 3 different tests on that.

src/api/app/services/monitor_controller_service/building_information_updater.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def worker_id(b)
+      b['workerid'].gsub(%r{[:./]}, '_')
+    end
+
+    def max_time
+      4 * 3600
```
danidoni: What about defining a constant?

src/api/app/services/monitor_controller_service/building_information_updater.rb
```diff
@@ -0,0 +1,50 @@
...
+    def initialize_workers
+      @workers = Hash[worker_status.elements('idle').collect { |b| [worker_id(b), {}] }]
+    end
+
+    def update_workers
+      @workers.merge!(Hash[@worker_status.elements('building').collect { |b| [worker_id(b), workers_hash(b, calculate_delta(b['starttime'].to_i))] }])
+    end
+
+    def workers_hash(b, delta)
+      Hash[['project', 'repository', 'package', 'arch', 'starttime'].collect { |k| [k, b[k]] }].merge('delta' => delta)
```
danidoni: Looks like this is filtering some keys and discarding the rest, then merging the `delta`. If so, you can use `Hash#slice` to filter `b` and then `merge`

src/api/app/services/monitor_controller_service/building_information_updater.rb
vpereira: because it get evaluated when the code is loaded which isn't really what we want for each calculation that we have.

src/api/app/services/monitor_controller_service/building_information_updater.rb
vpereira: I don't see any advantage on that. Do you want to benchmark it?

src/api/app/services/monitor_controller_service/building_information_updater.rb
vpereira:
```
vpereira@redstar:~> cat c.rb
FOO = 2/0

puts "bar"
vpereira@redstar:~> ruby c.rb 
	1: from c.rb:1:in `<main>'
c.rb:1:in `/': divided by 0 (ZeroDivisionError)
vpereira@redstar:~> 
```

src/api/app/services/monitor_controller_service/building_information_updater.rb
hennevogel: Hm, everything in XML is a string right?

src/api/app/services/monitor_controller_service/building_information_updater.rb
hennevogel: I have another problem. Those are magic numbers. What are those? Seconds I guess. Minutes? And why not use `14400` in 36 then? :-)

src/api/app/services/monitor_controller_service/building_information_updater.rb
danidoni: No, I don't want to benchmark it, its just that to me it feels easier to read like this:
```
def workers_hash(b, delta)
  b.slice('project', 'repository', 'package', 'arch', 'starttime').merge('delta' => delta)
end
```

vpereira: great @danidoni, updated, thanks!

src/api/app/services/monitor_controller_service/building_information_updater.rb
vpereira: I went with `4.hours.to_i`

src/api/app/services/monitor_controller_service/building_information_updater.rb
vpereira: yes, i thought so :) but I'm not sure if I change I will break someone consuming it

hennevogel: types can't be transferred over the wire.

src/api/app/services/monitor_controller_service/status_history_fetcher.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    private
+
+    def user_range(range)
+      [upper_range_limit, range].min
+    end
+
+    def upper_range_limit
+      24 * 365
```
hennevogel: Same magic numbers here. This time I have no idea what this would mean. 24 years? I also don't understand why we support arbitrary ranges if the interface only ever asks for 

Screenshot from 2020-01-29 12-52-39

hennevogel: I totally understand if you don't want to go into this so deep...

vpereira: I would love definitely to work on that but not in this PR.  I think that was just an arbitrary value chosen to make sure that `[upper_range_limit, user_range].min` works as expected.

PR #9011, opened 2020-01-29
[backend] add support for buildflags:obsgendiff for diff generation

mlschroe:
If the "obsgendiff" build flag is set and the repository has at
least one release target, we get all published files with the
suffix ".obsgendiff" from the first target.

It is up to the build job to do the diff and create new .obsgendiff
files.


adrianschroeter:
hm, not sure if the suffix is a good name, since it is actually no diff. The diff gets created in addition, but the needed file is more a complete state, like the .packages file.

adrianschroeter:
what about *.obsaggregate ?

mlschroe:
It's data for obsgendiff.

mlschroe:
I talked with Adrian and we'll merge this as it so we can start playing with this feature. We might rename to file in some upcoming commit, though.

PR #9013, opened 2020-01-30
WorkerStatus refactoring

vpereira:
Related with https://github.com/openSUSE/open-build-service/pull/9010

Refactoring and improvements being done in the model `WorkerStatus`:

- Focus on the method `WorkerStatus.hidden`
- Added spec for it
- Extracted sql query to a query object
- Instead of loop through all entries on worker_status trying to match for a name, filter the results using "XML XPATH"
- Extract small methods from it

### DIFF
--- a/src/api/app/models/worker_status.rb
+++ b/src/api/app/models/worker_status.rb
@@ -1,25 +1,45 @@
 class WorkerStatus
-  def self.hidden
-    mydata = Rails.cache.read('workerstatus')
-    ws = Nokogiri::XML(mydata || Backend::Api::BuildResults::Worker.status).root
-    prjs = {}
-    ws.css('building').each do |b|
-      prjs[b['project']] = 1
+  class << self
+    def hidden
+      mydata = Rails.cache.read('workerstatus')
+      ws = Nokogiri::XML(mydata || Backend::Api::BuildResults::Worker.status).root
+      # remove information about projects which are not visible to the current user
+      hide_project_information(ws)
     end
-    names = {}
-    # now try to find those we have a match for (the rest are hidden from you
-    Project.where(name: prjs.keys).pluck(:name).each do |n|
-      names[n] = 1
+
+    private
+
+    def hide_project_information(worker_status_root)
+      worker_status = worker_status_root
+      prjs = initialize_projects(worker_status)
+      xpath_string = generate_xpath(project_names(prjs.keys))
+      worker_status.xpath(xpath_string).each do |b|
+        b['project'] = '---'
+        b['repository'] = '---'
+        b['package'] = '---'
+      end
+      worker_status
     end
-    ws.css('building').each do |b|
-      # no prj -> we are not allowed
-      next if names.key?(b['project'])
-      Rails.logger.debug "workerstatus2clean: hiding #{b['project']} for user #{User.possibly_nobody.login}"
-      b['project'] = '---'
-      b['repository'] = '---'
-      b['package'] = '---'
+
+    # use Xpath to filter the projects
+    def generate_xpath(names)
+      return '//building' if names.empty?
+      # find all entries where project isn't in names
+      "//building[not(@project=#{xpath_list(names)})]"
     end
-    ws
+
+    # from array of project names generates a xpath sequence like ('name1', 'name2', 'nameX')
+    def xpath_list(names)
+      format("(#{names.collect { '\'%s\'' }.join(',')})", *names)
+    end
+
+    def initialize_projects(ws)
+      Hash[ws.css('building').collect { |b| [b['project'], 1] }]
+    end
+
+    def project_names(prjs)
+      ProjectNamesFinder.new(prjs).call
+    end
   end
 
   def update_workerstatus_cache
--- a/src/api/app/queries/project_names_finder.rb
+++ b/src/api/app/queries/project_names_finder.rb
@@ -1,0 +1,10 @@
+class ProjectNamesFinder
+  def initialize(projects, relation = Project.all)
+    @relation = relation
+    @projects = projects
+  end
+
+  def call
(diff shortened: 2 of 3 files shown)


src/api/app/models/worker_status.rb
```diff
@@ -1,25 +1,45 @@
...
     end
-    names = {}
-    # now try to find those we have a match for (the rest are hidden from you
-    Project.where(name: prjs.keys).pluck(:name).each do |n|
-      names[n] = 1
+
+    # use Xpath to filter the projects
+    def generate_xpath(names)
+      return '//building' if names.empty?
+      "//building[not(@project=#{xpath_list(names)})]"
```
vpereira: We try to find all `building` entries where the project doesn't meet one of the elements of the list

http://saxon.sourceforge.net/saxon7.9/expressions.html#sequence

src/api/app/models/worker_status.rb
```diff
@@ -1,25 +1,45 @@
...
-      # no prj -> we are not allowed
-      next if names.key?(b['project'])
-      Rails.logger.debug "workerstatus2clean: hiding #{b['project']} for user #{User.possibly_nobody.login}"
-      b['project'] = '---'
-      b['repository'] = '---'
-      b['package'] = '---'
+
+    # from array generates a xpath list like ('name1', 'name2', 'nameX')
+    def xpath_list(names)
+      format("(#{names.collect { |_f| '\'%s\'' }.join(',')})", *names)
```
coolo: I would have expected some higher level of paranoia from you when it comes to quoting :)

vpereira: :)

I assumed that because there is no user input involved and it will actually handle well-formatted project name, it would be ok.. argh this "_f"  :angry: 

I would love to have some help to improve this code, I failed to find some ruby magic one-liner to transform an array of strings to this XPATH sequence. Any idea @coolo ?

hennevogel: Paranoia about what? That the xpath query fails? :scream:

src/api/app/models/worker_status.rb
```diff
@@ -1,25 +1,46 @@
 class WorkerStatus
-  def self.hidden
-    mydata = Rails.cache.read('workerstatus')
-    ws = Nokogiri::XML(mydata || Backend::Api::BuildResults::Worker.status).root
-    prjs = {}
-    ws.css('building').each do |b|
-      prjs[b['project']] = 1
+  class << self
+    def hidden
+      mydata = Rails.cache.fetch('workerstatus') do
```
eduardoj: Any reason to change from a `Rails.cache.read` call to a `Rails.cache.fetch` one?

vpereira: Yes unless @coolo says differently, I don't see a point to don't put the information in cache, using `fetch`.

coolo: hmm, my knowledge about Rails.cache is not uptodate - but doesn't using `.fetch` implicitly write without expiry? Before it was reading and if the data is expired it reads from backend.

Having outdated cache data in /build/_workerstatus isn't really the point.

vpereira: > hmm, my knowledge about Rails.cache is not uptodate - but doesn't using `.fetch` implicitly write without expiry? Before it was reading and if the data is expired it reads from backend.
> 
> Having outdated cache data in /build/_workerstatus isn't really the point.

Hey @coolo thank you for the quick analysis.

I wondering if we are really going to have an outdated cache with something like that being executed by clockworkd:

```
  every(30.seconds, 'status.refresh') do
    # this should be fast, so don't delay
    WorkerStatus.new.update_workerstatus_cache
  end
```

I think it would be better to have the fetch expiring the key every 30 seconds (!!) and have just the `StatusHistory` generation running out of the main thread with a delayed job? 

In the clockwork which overwrites the same cache entry.

coolo: but that writes into the cache - and that's the reason `hidden` is only reading. So it's always clear what expiry date applies?

vpereira: > hmm, my knowledge about Rails.cache is not uptodate - but doesn't using `.fetch` implicitly write without expiry?

unless you are setting the `expire_in`, as we do

vpereira: > but that writes into the cache - and that's the reason `hidden` is only reading. So it's always clear what expiry date applies?

I just think it's a waste of resources keep generating a StatusHistory and caching this information, every 30 seconds. I understand that StatusHistory is necessary to show some statistics (probably 30 seconds resolution is an overkill) but definitely if nobody is accessing the site you don't need to keep invalidating the caching every 30 seconds and asking the backend for new data...

coolo: well, the /monitor graph can show an hour - so 120 points don't appear overkill to me.

vpereira: @eduardoj I rolled back this change, please review it again

vpereira: > well, the /monitor graph can show an hour - so 120 points don't appear overkill to me.

The meaning of the word overkill has nothing to do with what the graph is capable to show @coolo. Overkill is generate an excessive amount of data if it's not necessary to any business critical decision or actively invalidate the cache every 30 seconds regardless of user access or not, to save some milliseconds in the first hit in the monitor page. These whole statistics monitoring things should be done at https://obs-measure.opensuse.org/ anyway..

coolo: If build.opensuse.org was the only instance, I'd agree - but it's not. And requiring a grafana installation to see how the scheduler backlog is sounds a little harsh.

And this is not to save time for hitting /monitor - this is about showing a graph about the last hour's events. Grafana light.

Issue #8999, opened 2020-01-27
OBS publishes AppStream repodata in RPM repos as 'appdata.xml.gz' and 'appdata-icons.tar.gz' instead of correct names


Conan-Kudo:
# Issue Description
When OBS is configured to produce AppStream repodata for RPM repos in projects, it appends files to the repository metadata index as `appdata.xml.gz` and `appdata-icons.tar.gz`. This behavior does not match what `appstream-builder(1)` and `appstream-generator(1)` do, and package managers and tools that attempt to locate AppStream repodata in RPM repos fail to find it in OBS-generated repositories because of this.

Expected Result
============
OBS produces repositories with `appstream.xml.gz` and `appstream-icons.tar.gz` files instead of `appdata.xml.gz` and `appdata-icons.tar.gz`.

Further Information
===============
* This can be observed on the public OBS with the Tumbleweed repo: http://download.opensuse.org/tumbleweed/repo/oss/repodata/

doccaz:
Don't forget that SLE uses those filenames in the libzypp backend. Maybe create links/hard links when generating these files (and the ones required by the standard) to avoid duplicates? See issue #8259 .

Conan-Kudo:
@doccaz Can we fix libzypp and the appdata stuff in in SLE 15 SP2/Leap 15.2 (and of course TW) so that the standard names work? It's really bizarre that it's broken in this manner...

hellcp:
That should be fairly easy to fix in the plugin, it's a [pretty basic python script](https://github.com/DimStar77/openSUSE-appstream/blob/master/AsHelper.py#L74) loaded by another helper somewhere in libzypp (it is as much of a mess as you could imagine)

doccaz:
Now we need to decide what to do with the OBS publisher. Since we need compatibility with previous names for earlier releases, I vote to just add the "new" appstream catalog/icon files. What do you think?

hellcp:
That's fine, although it would be good to decide on a deadline where OBS will stop supporting earlier naming. If we do this for 15.2/15 SP2 and continue supporting this for 12 SP5, that gives us 7 years from now until the 12 SP5 EOL. Does 1. Jan 2027 sound good?

Conan-Kudo:
Could we make it a prjconf flag to have the legacy names for AppStream repodata in addition to the standard ones?

mlschroe:
[...]

PR #9019, opened 2020-01-30
adding appstream.xml and appstream-icons to match appstream-generator requirements

doccaz:
Fixing publisher code to also generate the appstream.xml and appstream-icons.gz files, to match what appstream-generator and everything that uses appdata look for (like Gnome Software) in recent releases.
We're also changing the way Leap 15.2 and Tumbleweed look for appdata to match this, as proposed in https://github.com/DimStar77/openSUSE-appstream/pull/2

As discussed in https://github.com/openSUSE/open-build-service/issues/8999 .

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -659,7 +659,10 @@ sub createrepo_rpmmd {
     if (-e "$extrep/repodata/appdata.xml") {
       print "    adding appdata.xml to repodata\n";
       qsystem($modifyrepo_bin, "$extrep/repodata/appdata.xml", "$extrep/repodata", @legacyargs) && die("    modifyrepo failed: $?\n");
-      unlink("$extrep/repodata/appdata.xml");
+      rename("$extrep/repodata/appdata.xml", "$extrep/repodata/appstream.xml");
+      print "    adding appstream.xml to repodata\n";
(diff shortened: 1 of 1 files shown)


adrianschroeter:
You are aware that the publisher is not creating the factory repository of the products?
So for the particular repo it is the wrong code.

It is also potential incompatible, but maybe okay. Are we sure that old code streams and tooling is not stumbling over the new file? We may need a configuration option otherwise ...

Please prefix your commit with [backend] tag.

thanks!

adrianschroeter:
Does it make to have both files at the same time at all?

Otherwise I would go with a config option instead...

doccaz:
> You are aware that the publisher is not creating the factory repository of the products?
> So for the particular repo it is the wrong code.
I didn't know that. Who creates the appdata for factory?

> Please prefix your commit with [backend] tag.
Done.

doccaz:
> Does it make to have both files at the same time at all?
> 
Yes, it already does a copy of appdata to appease both SLE12 and SLE15. What transpired from issue #8999 is that the filenames are different for the latest appstream specification.

> Otherwise I would go with a config option instead...
Someone else suggested adding a prjconf option to generate appdata (and the required format). I don't know my way around the UI code, though.

mlschroe:
What "latest appstream specification"? Such a thing does not really exist for the repository metadata.

PR #9028, opened 2020-02-03
Remove obs_factory excluded files in linters

eduardoj:
After removing obs_factory from the code base, exclusions on linter configuration files regarding obs_factory files are not needed any more.

I only searched for the "obs_factory" string in the code base. Probably the example in [lib/engines/](https://github.com/openSUSE/open-build-service/tree/master/src/api/lib/engines) should also be changed.


coolo:
lib/engines was not in use since obs_factory was merged - officially at least.

eduardoj:
Thanks @coolo. I added a commit to remove that code. Please @vpereira, have a look again.

PR #9031, opened 2020-02-03
Refactor project model

vpereira:
In this PR, I really didn't try to refactor methods, instead I went for organizing it

- Pack all class methods together
- Pack all private methods together
- Extract repeated SQL queries to query objects
- Added specs for it

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1182,16 +1182,14 @@ class Project < ApplicationRecord
 
   def open_requests_with_project_as_source_or_target
     # Includes also requests for packages contained in this project
-    rel = BsRequest.where(state: [:new, :review, :declined]).joins(:bs_request_actions)
-    rel = rel.where('bs_request_actions.source_project = ? or bs_request_actions.target_project = ?', name, name)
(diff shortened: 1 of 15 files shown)


src/api/app/models/project.rb
```diff
@@ -125,77 +120,364 @@ def staging_history
...
-  def self.deleted?(project_name)
-    return false if find_by_name(project_name)
+    # NOTE: This has to cover project name validations in src/backend/BSVerify.pm (verify_projid)
+    def valid_name?(name)
+      return false unless name.is_a?(String)
+      return false if name == '0'
+      return false if name =~ /:[:\._]/
+      return false if name =~ /\A[:\._]/
+      return false if name.end_with?(':')
+      return true  if name =~ /\A[-+\w\.:]{1,200}\z/
```
danidoni: :thought_balloon: I would find super useful to have those patterns as constants, with names explaining what they match

danidoni: Ok, I remembered this is  not a refactoring but a reorganization. So feel free to discard my comment.

PR #9034, opened 2020-02-04
Make the container fluid

DavidKang:
~~To remove the jumping effect of the containers depending on the size of
the viewport. We made the container fluid for all the viewports with
width < 1540px.~~

We made the container responsive for all the viewports with
width < 1540px. So the container doesn't take anymore different
max-width depending on the viewport, profiting better the horizontal
space.


hennevogel:
jumping effect?

DavidKang:
@hennevogel, I changed the description. I hope now is better explained.

hennevogel:
I still don't understand but okay ;-)

PR #9037, opened 2020-02-05
[backend] generate _statistics file when the result is no-change

sjamgade:
this patch moves some of the code around into a new function . It adds a
few lines for the real work of creating file when there is no change.
The code for the new function `updatestatsfromfile` add stats to the
passed in hash after reading it from file. The function is not
responsible for checking the existence of file and its format.

### DIFF
--- a/src/backend/BSSched/BuildJob.pm
+++ b/src/backend/BSSched/BuildJob.pm
@@ -544,11 +544,12 @@ sub jobfinished {
     }
     # Add a comment to logfile from last real build
     BSUtil::appendstr("$dst/logfile", "\nRetried build at ".localtime(time())." returned same result, skipped\n");
+    my $jobhist = makejobhist($info, $status, $js, 'unchanged');
+    addbuildstats($jobdatadir, $dst, $jobhist) if $all{'_statistics'};
     unlink("$gdst/:logfiles.fail/$packid");
     rename($meta, "$gdst/:meta/$packid") if $meta;
     unlink($_) for @all;
     rmdir($jobdatadir);
-    my $jobhist = makejobhist($info, $status, $js, 'unchanged');
     addjobhist($gctx, $prp, $jobhist);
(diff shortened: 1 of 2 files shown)


src/backend/bs_worker
```diff
@@ -3258,6 +3258,11 @@ sub dobuild {
     if (($buildinfo->{'reason'} || '') eq "rebuild counter sync") {
       $ret = 0;
     } else {
+      my $statsfile = "$buildroot/.mount/.build.packages/OTHER/_statistics";
+      updatestatsfromfile($stats, $statsfile) if (-e $statsfile);
+      $stats->{'times'}->{'total'}->{'time'} = { 'unit' => 's', '_content' => (time() - $starttime) };
```
mlschroe: I think you should keep this in sync with the $ret == 0 case and do this only
`if ($vm =~ /(xen|kvm|zvm|emulator|pvm|openstack)/)`

Also, I would not be surprised if you need that rename() done in the other cases to get the _statistics file in the correct place.

src/backend/BSSched/BuildJob.pm
```diff
@@ -544,11 +544,12 @@ sub jobfinished {
     }
     # Add a comment to logfile from last real build
     BSUtil::appendstr("$dst/logfile", "\nRetried build at ".localtime(time())." returned same result, skipped\n");
+    my $jobhist = makejobhist($info, $status, $js, 'unchanged');
+    addbuildstats($jobdatadir, $dst, $jobhist) if ($all{'_statistics'});
```
mlschroe: the parens are not needed if the `if` is a statement modifier. Can you please remove them so that the code fits the style of the rest of the build service?

I.e. just use `addbuildstats($jobdatadir, $dst, $jobhist) if $all{'_statistics'};`

Same for the other `if` modifiers in the rest of the commit.

PR #9038, opened 2020-02-05
Replace "Your Home" by "Profile"

saraycp:
To avoid confusion between the concepts "Your Home" and "Your Home Project",
the first one has been renamed to "Profile".

The icon for "Profile" is now a user icon instead of a house, which
is commonly used for the main page (a.k.a landing or home page).

Welcome   Open Build Service(2)


coolo:
Why the 'Your'? Who's profile would it be if not mine?

saraycp:
> Why the 'Your'? Who's profile would it be if not mine?

Yes, I also had the doubt and @vpereira suggested the same. Changed!

PR #9043, opened 2020-02-06
[dist] Migrate SUSE-specific rcFOO link creation from Makefile to spec

Conan-Kudo:
This logic only makes sense on SUSE Linux distributions, where the
packaging policy indicates this preference. The links don't work
at all on Fedora or any other distribution.

### DIFF
--- a/dist/obs-server.spec
+++ b/dist/obs-server.spec
@@ -403,6 +403,16 @@ make
 export DESTDIR=$RPM_BUILD_ROOT
 export OBS_VERSION="%{version}"
 DESTDIR=%{buildroot} make install
+
+%if 0%{?suse_version}
+systemd_services="$(basename --multiple --suffix .service %{buildroot}%{_unitdir}/*.service) $(basename --multiple --suffix .target %{buildroot}%{_unitdir}/*.target)"
+for systemd_service in $systemd_services; do
+    if [[ $systemd_service != *"@"* ]]; then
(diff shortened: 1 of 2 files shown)


dist/obs-server.spec
```diff
@@ -876,7 +892,7 @@ usermod -a -G docker obsservicerun
 %{_unitdir}/obs-delayedjob-queue-releasetracking.service
 %{_unitdir}/obs-delayedjob-queue-staging.service
 %{_unitdir}/obs-sphinx.service
-%{_sbindir}/rcobs-api-support
+%if 0%{?suse_version}
 %{_sbindir}/rcobs-clockwork
```
eduardoj: It seems `rcobs-api-support` got removed. Any reason?

Conan-Kudo: This was previously discussed in the original PR, but it doesn't work because the `rcFOO` stuff only works with `.service` files. It's not supposed to be used with `.target` and `.timer` files.

M0ses: Uhh,  I have to disagree. Basically it seems to work also for targets:

```
nextcloud:~ # rcredis status
* redis.target - Redis target allowing to start/stop all redis@.service instances at once
```

Conan-Kudo: It shouldn't... But I guess I can put it back since it clearly seems to...

Conan-Kudo: Restored.

M0ses:
Failing circleci tests seems not to be related to this PR

PR #9045, opened 2020-02-06
[backend] support container architecture variants

adrianschroeter:
Esp. to avoid armv6 and v7 conflicts

The kiwi/umoci/docker/podman tooling seems not to support it yet, but that are independend bugs.
Could not test it for real therefore yet.


Vogtinator:
> The kiwi/umoci/docker/podman tooling seems not to support it yet, but that are independend bugs.
Could not test it for real therefore yet.

The way I dealt with that in docker-publisher (which pushes from `api.opensuse.org` to `registry-1.docker.io`) is by using the arch of the binary as source for the variant. Doing it like this here would need too much changes for a simple workaround though AFAICT

PR #9051, opened 2020-02-07
Use the responsive namespace instead of harcoding it

danidoni:


danidoni:
We talked offline and this is not really needed. So closing it

PR #9055, opened 2020-02-07
Fix virtual hosts apache subdirectory in spec file

eduardoj:
This way OBS installed in SUSE and openSUSE distributions, and also the OBS appliances will find again the OBS apache configuration file.

openQA tests for OBS appliances are failing. Applying this patch would fix them.


M0ses:
THX for your fix

PR #9057, opened 2020-02-07
Fix partial editor option hash defaults

### LINKED ISSUE #8018: Viewing spec files through a letterbox is not good

Conan-Kudo:
We've setup some defaults, but only if you do not pass a `style` local to the
editor partial. Also set defaults if they are not passed in with `style`.

Fixes #8018 for v2.10

(cherry picked from commit 91815b1829d7459a9beb4e184392898a3ff0a2e5)

### DIFF
--- a/src/api/app/views/webui2/shared/_editor.html.haml
+++ b/src/api/app/views/webui2/shared/_editor.html.haml
@@ -1,6 +1,7 @@
 :ruby
   save ||= {}
-  style ||= { read_only: false, big_editor: true }
+  style ||= {}
+  style.reverse_merge!(read_only: false, big_editor: true)
   uid ||= next_codemirror_uid
   content_for(:head_style, codemirror_style(style))
 


Conan-Kudo:
@hennevogel @adrianschroeter Can this be pulled in to 2.10 soon, along with #9039?

Conan-Kudo:
@adrianschroeter @hennevogel Rebased on current 2.10 HEAD to make it mergeable.

(cc: @hellcp)

hennevogel:
coverage is fine, see codecov.

PR #9060, opened 2020-02-08
Refactor WorkerStatus model

vpereira:
Among the changes we did:

- Substituted XmlParse for Nokogiri
- Use when possible XPATH search instead of loop through objects
- Extracted repeated code to a new method
- Improved ruby dialect

### DIFF
--- a/src/api/app/models/architecture.rb
+++ b/src/api/app/models/architecture.rb
@@ -16,6 +16,7 @@ class Architecture < ApplicationRecord
 
   #### Scopes (first the default_scope macro if is used)
   scope :available, -> { where(available: 1) }
+  scope :unavailable, -> { where(available: 0) }
 
   #### Validations macros
   validates :name, uniqueness: true
--- a/src/api/app/models/worker_status.rb
+++ b/src/api/app/models/worker_status.rb
@@ -44,77 +44,80 @@ class WorkerStatus
 
   def update_workerstatus_cache
     # do not add hiding in here - this is purely for statistics
-    ret = Backend::Api::BuildResults::Worker.status
-    wdata = Xmlhash.parse(ret)
+    backend_status = Backend::Api::BuildResults::Worker.status
+    wdata = Nokogiri::XML(backend_status)
     @mytime = Time.now.to_i
-    @squeues = {}
-    Rails.cache.write('workerstatus', ret, expires_in: 3.minutes)
+    @squeues = Hash.new(0)
+
+    Rails.cache.write('workerstatus', backend_status, expires_in: 3.minutes)
+
     StatusHistory.transaction do
-      wdata.elements('blocked') do |e|
-        save_value_line(e, 'blocked')
+      ['blocked', 'waiting'].each do |state|
+        wdata.search("//#{state}").each { |e| save_value_line(e, state) }
       end
-      wdata.elements('waiting') do |e|
-        save_value_line(e, 'waiting')
-      end
-      wdata.elements('partition') do |p|
-        p.elements('daemon') do |daemon|
-          parse_daemon_infos(daemon)
-        end
-      end
+      wdata.search('partition/daemon').each { |daemon| parse_daemon_infos(daemon) }
       parse_worker_infos(wdata)
-      @squeues.each_pair do |key, value|
-        StatusHistory.create(time: @mytime, key: key, value: value)
-      end
+      @squeues.each_pair { |key, value| StatusHistory.create(time: @mytime, key: key, value: value) }
     end
-    ret
+
+    backend_status
   end
 
   private
 
   def add_squeue(key, value)
-    @squeues[key] ||= 0
     @squeues[key] += value.to_i
   end
 
   def parse_daemon_infos(daemon)
-    return unless daemon['type'] == 'scheduler'
-    arch = daemon['arch']
-    # FIXME2.5: The current architecture model is a gross hack, not connected at all
-    #           to the backend config.
-    a = Architecture.find_by_name(arch)
-    if a
-      a.available = true
-      a.save
-    end
-    queue = daemon.get('queue')
+    return unless daemon.attributes['type'].value == 'scheduler'
+
+    queue = daemon.at_xpath('queue')
     return unless queue
-    ['high', 'next', 'med', 'low'].each { |key| add_squeue("squeue_#{key}_#{arch}", queue[key]) }
+
+    architecture_name = daemon.attributes['arch'].value
+
+    daemon_architecture(architecture_name)
+
+    ['high', 'next', 'med', 'low'].each do |key|
+      s_key = squeue_key([key, architecture_name])
+      add_squeue(s_key, queue.attributes[key].value)
+    end
+  end
+
+  def squeue_key(key_parts)
+    generic_key_generation(key_parts.prepend('squeue'))
+  end
+
+  def generic_key_generation(key_parts)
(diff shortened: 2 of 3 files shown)


src/api/app/models/worker_status.rb
```diff
@@ -44,88 +44,91 @@ def project_names(prjs)
...
     # FIXME2.5: The current architecture model is a gross hack, not connected at all
     #           to the backend config.
-    a = Architecture.find_by_name(arch)
-    if a
-      a.available = true
-      a.save
-    end
-    queue = daemon.get('queue')
+    queue = daemon.at_xpath('queue')
     return unless queue
```
vpereira: Can that happen that this XML https://build.opensuse.org/apidocs/workerstatus.xml in the `partition > daemon` section comes without queue?  @adrianschroeter ?

src/api/app/models/worker_status.rb
```diff
@@ -44,88 +44,91 @@ def project_names(prjs)
...
   end
 
   def parse_worker_infos(wdata)
     allworkers = {}
     workers = {}
     ['building', 'idle', 'dead', 'down', 'away'].each do |state|
-      wdata.elements(state) do |e|
-        id = e['workerid']
-        if workers.key?(id)
-          Rails.logger.debug 'building+idle worker'
```
vpereira: Kind of superfluous logging call. Removed.

src/api/app/models/worker_status.rb
```diff
@@ -44,88 +44,91 @@ def project_names(prjs)
...
   def add_squeue(key, value)
-    @squeues[key] ||= 0
     @squeues[key] += value.to_i
   end
 
   def parse_daemon_infos(daemon)
-    return unless daemon['type'] == 'scheduler'
-    arch = daemon['arch']
+    return unless daemon.attributes['type'].value == 'scheduler'
     # FIXME2.5: The current architecture model is a gross hack, not connected at all
```
vpereira: does this comment make any sense? 2.5 "not fixed", 2.10, "not fixed".. so I guess this comment can go away and if somebody think that it is important, please open a GH issue..

src/api/app/models/worker_status.rb
```diff
@@ -44,88 +44,83 @@ def project_names(prjs)
...
-          next
-        end
-        workers[id] = 1
-        key = state + '_' + e['hostarch']
-        allworkers["building_#{e['hostarch']}"] ||= 0
-        allworkers["idle_#{e['hostarch']}"] ||= 0
-        allworkers["dead_#{e['hostarch']}"] ||= 0
-        allworkers["down_#{e['hostarch']}"] ||= 0
-        allworkers["away_#{e['hostarch']}"] ||= 0
-        allworkers[key] = allworkers[key] + 1
```
vpereira: do we need to create a `StatusHistory` for those which weren't found? For example if you run the same spec on master, it will create 3 more `StatusHistory` elements for (dead, down, away) even though, no dead, down or always workers were found (but they are being manually initialized with 0 in the master )

src/api/app/models/worker_status.rb
```diff
@@ -44,88 +44,83 @@ def project_names(prjs)
...
 
   def add_squeue(key, value)
-    @squeues[key] ||= 0
     @squeues[key] += value.to_i
   end
 
   def parse_daemon_infos(daemon)
-    return unless daemon['type'] == 'scheduler'
-    arch = daemon['arch']
-    # FIXME2.5: The current architecture model is a gross hack, not connected at all
```
coolo: And where do you create architectures now?

src/api/app/models/worker_status.rb
coolo: You don't, but dropping them makes scaling a lttle more complex - same for plotting. But give it a try

vpereira: ok let's see how it goes, thanks @coolo

src/api/app/models/worker_status.rb
vpereira: Looking in master we don't create it here https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/worker_status.rb#L79 @coolo, and probably we won't create `Achitecture` while calculating the statistics, but I understand that I made it dependent of an existing `Architecture`, but mainly because my assumption is if there is a worker for a specific architecture, this architecture should exist in the DB, already, no?

coolo: But it's set to available here

vpereira: here too `Architecture.find_by_name(arch).tap { |a| a.update(available: true) }`

vpereira: @coolo did the last answer, answered your question about making the Architecture available?

coolo: Well, yes - but now you find and update the architecture in a loop instead of once. Doesn't appear ideal to me.

vpereira: > Well, yes - but now you find and update the architecture in a loop instead of once. Doesn't appear ideal to me.

ah, yes well, Rails provides a SQL query cache, used to cache the results of database queries for the duration of the request. So while we are in the same request we are good: https://guides.rubyonrails.org/caching_with_rails.html#sql-caching

vpereira: I changed that little bit @coolo. I added a scope for `Architecture` named `unavailable` and just update the objects if necessary.. and I do everything outside the loop, to avoid confusion.

src/api/app/models/architecture.rb
```diff
@@ -16,6 +16,7 @@ class Architecture < ApplicationRecord
 
   #### Scopes (first the default_scope macro if is used)
   scope :available, -> { where(available: 1) }
+  scope :unavailable, -> { where.not(id: available) }
```
coolo: is this a subquery? Why not plainly `where(available: 0)` ?

vpereira: Yes it was a sub-query. Changed.

src/api/app/models/worker_status.rb
```diff
@@ -4,33 +4,25 @@ def hidden
...
       prjs = initialize_projects(worker_status)
-      xpath_string = generate_xpath(project_names(prjs.keys))
-      worker_status.xpath(xpath_string).each do |b|
-        b['project'] = '---'
-        b['repository'] = '---'
-        b['package'] = '---'
+      prj_names = project_names(prjs.keys)
+
+      worker_status.css('building').each do |b|
+        next if prj_names.include?(b['project'])
```
vpereira: CSS selector is much slower than plain  ruby, even though looking from straight Big O point of view,  this each + check if list contains element should be worse. https://gist.github.com/vpereira/18ff44783b18176926d3ecd2dfabc481

src/api/app/models/worker_status.rb
```diff
@@ -4,33 +4,25 @@ def hidden
...
-    def generate_xpath(names)
-      return '//building' if names.empty?
-      # find all entries where project isn't in names
-      "//building[not(@project=#{xpath_list(names)})]"
-    end
-
-    # from array of project names generates a xpath sequence like ('name1', 'name2', 'nameX')
-    def xpath_list(names)
-      format("(#{names.collect { '\'%s\'' }.join(',')})", *names)
+    def hide_project_information(prj)
```
danidoni: I would add a bang at the end to signal that this method changes `prj` instead of returning a modified copy

PR #9078, opened 2020-02-12
Get rid of systemV

eduardoj:
As systemd is being used to manage OBS services, systemV method calls are no longer needed. Removing systemV code and compatibility packages (as insserv-compat) makes it more clear OBS services handling.

Labelling as "DO NOT MERGE" untill the changes in the service scripts are tested. The pull request was created to receive feedback about removing systemV from the OBS codebase.

### DIFF
--- a/dist/obs-server.spec
+++ b/dist/obs-server.spec
@@ -188,7 +188,7 @@ BuildRequires: obs-server-macros
 
 %if 0%{?suse_version:1}
 BuildRequires:  fdupes
-PreReq:         %insserv_prereq permissions pwdutils
+Requires(pre):  shadow
 %endif
 
 %if 0%{?suse_version:1}
@@ -227,9 +227,6 @@ Provides:       obs-source_service = %version
 
 Recommends:     obs-service-download_url
 Recommends:     obs-service-verify_file
-%if 0%{?suse_version} >= 1550
-Requires:       insserv-compat
-%endif
 
 BuildRequires:  systemd-rpm-macros
 
@@ -262,9 +259,6 @@ Group:          Productivity/Networking/Web/Utilities
 Requires:       util-linux >= 2.16
 # the following may not even exist depending on the architecture
 Recommends:     powerpc32
-%if 0%{?suse_version} >= 1550
-Requires:       insserv-compat
-%endif
 
 %description -n obs-worker
 This is the obs build host, to be installed on each machine building
@@ -276,7 +270,7 @@ Summary:        The Open Build Service -- base configuration files
 Group:          Productivity/Networking/Web/Utilities
 %if 0%{?suse_version}
 Requires(pre):  shadow
-PreReq:         %fillup_prereq
+Requires(pre):  %fillup_prereq
 %endif
 
 %description -n obs-common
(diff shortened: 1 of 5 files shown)


eduardoj:
All the tests done passed, confirming that the services involved were succesfully reloaded:
- manually upgrading the obs-server and obs-worker packages
- manually reinstalling the same packages
- and also `obs-server` and `upgrade-lastest` kanku tests, thanks to @M0ses.

Removing the "DO NOT MERGE" label. The pull request is ready to review.

dist/obs-server.spec
```diff
@@ -188,7 +188,6 @@ BuildRequires: obs-server-macros
 
 %if 0%{?suse_version:1}
 BuildRequires:  fdupes
-PreReq:         %insserv_prereq permissions pwdutils
```
bugfinder: actually, yes "permissions" can go as well, as "%run_permissions" call  was removed in commit d0c101da0986cd4d9df6e0f9754cf202e4f882bb from 2013 ...
but then the cleanup needs to be completed by removing

 %verifyscript -n obs-server
 %verify_permissions

as well ...

and "pwdutils" should be kept, but replaced by "shadow" as it's still used here:
%triggerin -n obs-server -- docker
usermod -a -G docker obsservicerun

so this should IMHO be:
-PreReq:         %insserv_prereq permissions pwdutils
+Requires(pre): shadow

dist/obs-server.spec
```diff
@@ -276,7 +269,6 @@ Summary:        The Open Build Service -- base configuration files
 Group:          Productivity/Networking/Web/Utilities
 %if 0%{?suse_version}
 Requires(pre):  shadow
-PreReq:         %fillup_prereq
```
bugfinder: no, this is still needed for the %fillup_only call,
should probably be replaced by
Requires(pre): %fillup_prereq

dist/obsworker
```diff
@@ -352,8 +335,8 @@ case "$1" in
             NUM=`ls -d /sys/devices/system/cpu/cpu[0-9]* | wc -l`
         fi
         if [ "--zvm" == "$vmopt" ]; then
-            check_vmcp || rc_status -v
-            create_initrd $OBS_VM_KERNEL $OBS_VM_INITRD || rc_status -v
+            check_vmcp
+            create_initrd $OBS_VM_KERNEL $OBS_VM_INITRD
```
bugfinder: you're loosing the state here, should probably just be

check_vmcp || exit 1
create_initrd .... || exit 1

or something else to fail if intermediate steps have failed

dist/obs-server.spec
eduardoj: Thanks! `%verify*` lines removed, and suggested patch applied.

dist/obs-server.spec
eduardoj: Done.

dist/obsworker
eduardoj: Done. I substituted these lines as you proposed and added a comment in the affected commit.

Please, re-review it again.

eduardoj:
"OBS Package Build" tests failed after merging the "Update rack" pull request. I only rebased with master. Waiting for the tests to pass to merge the pull request.

PR #9081, opened 2020-02-17
Change top navigation menu

DavidKang:
We limited the max-width of the top navigation and also changed the menus on the top navigation.

Screenshot_2020-02-17 Open Build Service(2)

### DIFF
--- a/src/api/app/assets/stylesheets/webui/responsive_ux.scss
+++ b/src/api/app/assets/stylesheets/webui/responsive_ux.scss
@@ -10,13 +10,27 @@ body.responsive-ux {
   body.responsive-ux {
     // START LAYOUT
     #navigation {
-      .menu-options { font-size: initial; }
-      .offcanvas-collapse { right: .5rem; }
-      .watchlist-collapse { right: 8rem; }
+      .menu-options {
+        font-size: initial;
 
-      .watchlist-collapse, .offcanvas-collapse {
+        #link-to-user-home { order: 2; }
+        #logout-link { order: 3; }
+        .nav-link {
+          order: 1;
+          i { vertical-align: middle; }
+          &:last-child { padding-right: 0 !important; }
+        }
+      }
+
+      .toggler {
+        display: flex !important;
+        .nav-link { color: #f8f9fa !important; }
+      }
+
+      .watchlist-collapse {
         position: absolute;
         top: 49px;
+        right: 8rem;
         bottom: initial;
         left: initial;
         width: initial;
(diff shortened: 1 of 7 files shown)


dmarcoux:
Whenever there is room available, so on most devices except smaller ones like phones, I suggest that we display all items directly in the top navigation bar. So if I want to logout, I could directly click `Logout` instead of having to click on my avatar, then on `Logout`.

DavidKang:
@dmarcoux, something like this:

Screenshot_2020-02-17 Open Build Service

Honestly, I don't like to have the logout at this level.

dmarcoux:
I like it!

I'm fine with having the `Logout` on this level, but completely on the right side. If not, where should we put it? As before in a menu under the avatar, together with `Profile`?

DavidKang:
> where should we put it? As before in a menu under the avatar, together with Profile?

@dmarcoux, I don't have a clear solution for that. For now, I can live with the logout at the end.

danidoni:
> > where should we put it? As before in a menu under the avatar, together with Profile?
> 
> @dmarcoux, I don't have a clear solution for that. For now, I can live with the `logout` at the right.

What about hiding the buttons under something that slides left when you move the mouse over, like for example, the profile. Then you have a one-click solution and is not "polluting" the top bar. The downside is that maybe its not super evident, but maybe with an arrow/left-chevrons suggesting something can be pulled off...

dmarcoux:
I don't agree that it's "polluting" the top bar to have the buttons there. I understand that we need to think about different strategies on devices with smaller viewports, but on desktop, displaying a small amount of icons is fine.

hennevogel:
I dont like it either. It's too cramped and gives logout a weight that it does not deserve.

Especially as the off canvas menu provides the possibility to extend the navigation beyond this...

danidoni:
> I don't agree that it's "polluting" the top bar to have the buttons there. I understand that we need to think about different strategies on devices with smaller viewports, but on desktop, displaying a small amount of icons is fine.

Okay!

vpereira:
Screenshot_2020-02-19_23-10-49

I don't like to use the "GH does like that" joker but my point is more like, it's doable, we use GH daily and it's really not a problem to have it there 

Gitlab does the same:

Screenshot_2020-02-19_23-14-15

Trello, Basecamp, Gmail, you name it.. all of them do it as we do, so it's pretty much a standard to have it as we have now.

DavidKang:
@vpereira, after the planning meeting this PR is no valid anymore. So I will close it.

PR #9094, opened 2020-02-19
WIP: Configure guardrails and remove Hakiri

krauselukas:


coolo:
What's the gain? https://www.guardrails.io/docs/en/tools uses the same old brakeman 4.3 not supporting rails 6

PR #9116, opened 2020-02-21
[backend] add 'cacheonly' option for diffing sources

mlschroe:
We'll use this in the API to defer source diffing to a delay job,
so that the WebUI does not clog up all source server slots.

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -1035,6 +1035,7 @@ class Webui::PackageController < Webui::WebuiController
 
   def get_diff(project, package, options = {})
     options[:view] = :xml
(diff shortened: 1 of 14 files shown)


vpereira:
@mlschroe ```undefined method `is_nobody?' for nil:NilClass```

adrianschroeter:
@vpereira my bug, hopefully fixed now.

mlschroe:
@adrianschroeter you need to add the `cachonly` option to the list of accepted options in the source_diff method (`lib/backend/api/sources/package.rb`)

vpereira:
> @vpereira my bug, hopefully fixed now.

Ok, but the spider is failing @adrianschroeter normally it means that some route is raising an exception..

PR #9117, opened 2020-02-21
[backend] fix oracle to match multiple cpu flags

lethliel:
Instead of accepting only one match of cpu flags while
constraints validation all flags must be present in the
worker

### DIFF
--- a/src/backend/BSDispatcher/Constraints.pm
+++ b/src/backend/BSDispatcher/Constraints.pm
@@ -115,7 +115,7 @@ sub oracle {
     if ($constraints->{'hardware'}->{'cpu'}) {
       return 0 unless $worker->{'hardware'}->{'cpu'};
       my %workerflags = map {$_ => 1} @{$worker->{'hardware'}->{'cpu'}->{'flag'} || []};
-      return 0 unless grep {$workerflags{$_}} @{$constraints->{'hardware'}->{'cpu'}->{'flag'} || []};
+      return 0 if grep {!$workerflags{$_}} @{$constraints->{'hardware'}->{'cpu'}->{'flag'} || []};
     }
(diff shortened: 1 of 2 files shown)


src/backend/BSDispatcher/Constraints.pm
```diff
@@ -114,8 +114,9 @@ sub oracle {
     return 0 if $constraints->{'hardware'}->{'physicalmemory'} && getmbsize($constraints->{'hardware'}->{'physicalmemory'}) > $memory;
     if ($constraints->{'hardware'}->{'cpu'}) {
       return 0 unless $worker->{'hardware'}->{'cpu'};
-      my %workerflags = map {$_ => 1} @{$worker->{'hardware'}->{'cpu'}->{'flag'} || []};
-      return 0 unless grep {$workerflags{$_}} @{$constraints->{'hardware'}->{'cpu'}->{'flag'} || []};
+      for my $flag (@{$worker->{'hardware'}->{'cpu'}->{'flag'} || []}) {
+	  return 0 unless grep {$flag} @{$constraints->{'hardware'}->{'cpu'}->{'flag'} || []};
+      }
```
mlschroe: Bah! How about:
```
my %workerflags = map {$_ => 1} @{$worker->{'hardware'}->{'cpu'}->{'flag'} || []};
return 0 if grep {!$workerflags{$_}} @{$constraints->{'hardware'}->{'cpu'}->{'flag'} || []};
```

mlschroe: And what on earth is `{$flag}` supposed to do? You didn't test this, did you? ;-)

lethliel:
@mlschroe fixed

PR #9118, opened 2020-02-21
Use flex in build results elements

eduardoj:
This way we prevent the elements to wrap when there is still enough space, or to overlap. This change is applied both to the build results of a project and a package.

**Before:**
Screenshot from 2020-02-21 14-02-27_build_results_project_before

**After:**
Screenshot from 2020-02-21 14-02-51_build_results_project_after


eduardoj:
[BuildResults](https://trello.com/c/47jMdsfn/1333-buildresults)

PR #9127, opened 2020-02-24
Enable responsive option and change DOM options for Datatables

vpereira:
With this PR I suggest two changes to be done in Datatables to improve the responsive usability.

1-) Enable https://datatables.net/extensions/responsive/
2-) Change the DOM options order and remove the select box for https://datatables.net/reference/option/lengthChange

Before:

Screenshot_2020-02-24_15-33-30

Screenshot_2020-02-24_15-33-57

After:

Screenshot_2020-02-24_15-32-03

Screenshot_2020-02-24_15-32-33

### DIFF
--- a/src/api/app/assets/javascripts/webui/datatables.js
+++ b/src/api/app/assets/javascripts/webui/datatables.js
@@ -13,7 +13,8 @@ var DEFAULT_DT_PARAMS = {
     info: "page _PAGE_ of _PAGES_ (_TOTAL_ records)",
     infoFiltered: ""
   },
-
+  responsive: true,
+  dom: 'ftpi',
   pageLength: 25,
   stateSave: true,
   pagingType: 'full',


src/api/app/assets/javascripts/webui/datatables.js
```diff
@@ -13,7 +13,8 @@ var DEFAULT_DT_PARAMS = {
     info: "page _PAGE_ of _PAGES_ (_TOTAL_ records)",
     infoFiltered: ""
   },
-
+  responsive: true,
+  dom: 'ftpi',
```
hennevogel: Why?

vpereira:

src/api/app/assets/javascripts/webui/datatables.js
vpereira: hey @hennevogel 

if you look the documentation https://datatables.net/reference/option/dom, you will see that I'm changing the search elements. "ftpi" translates to:

f: filter
t: table
p: pagination
i: table summary

It controls which search elements are going to be displayed and in which order. So "ftpi" is different than "ftip" which would show the "table summary" before the "pagination". In the case I removed the "l" and reodered the "i" and "p" as we did in the Projects#index. I think the default is "fltip"

hennevogel: I understand what you are doing. But why do you remove the length?

vpereira: > I understand what you are doing. But why do you remove the length?

why have it in first place?

hennevogel: To have the option to have many things on one page. For overview I would guess.

hennevogel: I don't understand, this has little to do with search.

vpereira: > To have the option to have many things on one page. For overview I would guess.

we had it because it was default in datatables. We didn't decide to have it.

PR #9134, opened 2020-02-26
Authorize against the linked project instead of the package's project

### LINKED ISSUE #6289: Trigger rebuild button broken for staging projects
Note that this is not related to the new staging workflow, but broken for all project links most likely.

In https://build.opensuse.org/package/live_build_log/openSUSE:Leap:15.1:Staging:A/kdebase4-workspace/standard/x86_64 I have a 'Trigger rebuild' button (as I'm maintainer in openSUSE:Leap:15.1:Staging I should have). But once I click this button, I get an error that I'm not allowed to retrigger it.

I assume because it's an inherited package. But I do *have* the rights:

```
POST https://api.opensuse.org/build/openSUSE:Leap:15.1:Staging:A?cmd=rebuild&arch=x86_64&repository=standard&package=kdebase4-workspace
ok
```

this makes the webui basically unusable for staging manager tasks.

danidoni:
When we're triggering the build from a linked project, the Pundit Policy checker
for the package fetches the package's project, not the one we're rebuilding
from.

This PR detects if we're triggering the build from a linked project and then
passes that project to Pundit to perform the authorization.

Fixes #6289

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -700,7 +700,14 @@ class Webui::PackageController < Webui::WebuiController
   end
 
   def trigger_rebuild
-    authorize @package, :update?
+    # When we're in a linked project, the package's project points to some other
+    # project, not the one we're triggering the build from.
+    # Here we detect that, and if so, we authorize against the linked project.
+    if @project != @package.project
+      authorize @project, :update?
+    else
+      authorize @package, :update?
+    end
 
     allowed_params = [:project, :package, :repository, :arch]
 


src/api/app/controllers/webui/package_controller.rb
```diff
@@ -700,7 +700,14 @@ def abort_build
   end
 
   def trigger_rebuild
-    authorize @package, :update?
+    # When we're in a linked project, the package's project points so some other
```
dmarcoux: Did you mean "**t**o some other" ??

danidoni: Yes

danidoni:
Squashed

PR #9136, opened 2020-02-26
Enable datatable responsive plugin as default

vpereira:
Enable the responsive extension to be able to control datatables in different query medias https://datatables.net/extensions/responsive/


vpereira:

PR #9139, opened 2020-02-26
Refactor PackageController#trigger_rebuild to a Service Object

vpereira:
Move PackageCotnroller#trigger_rebuild logic to a Service object. 

Reuse the same service and authorization on the TriggerController#rebuild

### DIFF
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -20,8 +20,9 @@ class TriggerController < ApplicationController
   include Trigger::Errors
 
   def rebuild
-    check_rebuild_permission!
-    Backend::Api::Sources::Package.rebuild(@prj.name, @pkg_name)
+    rebuild_trigger = PackageControllerService::RebuildTrigger.new(package: @pkg, project: @prj, params: params)
+    authorize rebuild_trigger.policy_object, :update?
+    rebuild_trigger.rebuild?
     render_ok
   end
 
@@ -48,13 +49,6 @@ class TriggerController < ApplicationController
   end
 
   private
-
-  def check_rebuild_permission!
-    return if User.session.can_modify_project?(@prj)
-    return if @pkg.project == @prj && policy(@pkg).update?
-
-    raise NoPermissionForPackage.setup('no_permission', 403, "no permission for package #{@pkg} in project #{@prj}")
-  end
 
   def prepare_path_for_runservice
     path = @pkg.source_path
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -700,22 +700,14 @@ class Webui::PackageController < Webui::WebuiController
   end
 
   def trigger_rebuild
-    # When we're in a linked project, the package's project points to some other
-    # project, not the one we're triggering the build from.
-    # Here we detect that, and if so, we authorize against the linked project.
-    if @project != @package.project
-      authorize @project, :update?
-    else
-      authorize @package, :update?
-    end
+    rebuild_trigger = PackageControllerService::RebuildTrigger.new(package: @package, project: @project, params: params)
+    authorize rebuild_trigger.policy_object, :update?
 
-    allowed_params = [:project, :package, :repository, :arch]
-
-    if @package.rebuild(params.slice(*allowed_params).permit!.to_h)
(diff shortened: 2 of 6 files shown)


src/api/app/controllers/webui/package_controller.rb
```diff
@@ -703,19 +703,14 @@ def trigger_rebuild
...
     # Here we detect that, and if so, we authorize against the linked project.
-    if @project != @package.project
-      authorize @project, :update?
-    else
-      authorize @package, :update?
-    end
-
-    allowed_params = [:project, :package, :repository, :arch]
+    rebuild_trigger = PackageControllerService::RebuildTrigger.new(package: @package, project: @project, params: params)
+    authorize rebuild_trigger.fetch_policy, :update?
```
vpereira: I would like to move the policy check to the service, but because of the `verify_authorized` https://github.com/varvet/pundit/blob/6be4621538f916124f76b57460c8cc46597a3b6a/lib/pundit.rb#L192, it isn't  working

src/api/app/controllers/trigger_controller.rb
```diff
@@ -20,8 +20,9 @@ class TriggerController < ApplicationController
   include Trigger::Errors
 
   def rebuild
-    check_rebuild_permission!
-    Backend::Api::Sources::Package.rebuild(@prj.name, @pkg_name)
+    rebuild_trigger = PackageControllerService::RebuildTrigger.new(package: @pkg, project: @prj, params: params)
+    authorize rebuild_trigger.fetch_policy, :update?
+    rebuild_trigger.rebuild?
     render_ok
```
vpereira: we are able to give the user a feedback if rebuild fails.. should we?

adrianschroeter: the rebuild trigger, you mean?

it should never fail, if permissions are right. Not sure if it is worth to handle such a case ... you could notify the admin that he has a problem though.

src/api/app/services/package_controller_service/rebuild_trigger.rb
```diff
@@ -0,0 +1,32 @@
...
+      @package = options[:package]
+      @project = options[:project]
+      @params = options[:params]
+    end
+
+    def rebuild?
+      @package.rebuild(@params.slice(*allowed_params).permit!.to_h)
+    end
+
+    def fetch_policy
```
hennevogel: I don't like this name. How about 

```suggestion
    def policy_object
```

src/api/app/controllers/webui/package_controller.rb
hennevogel: You know you can skip this right? Just checking...

src/api/app/services/package_controller_service/rebuild_trigger.rb
```diff
@@ -0,0 +1,32 @@
...
+      return @project if @project != @package.project
```
hennevogel: Please also "port" the documentation why we do this. It's very non-obvious...

src/api/app/controllers/webui/package_controller.rb
vpereira: for sure I know that we can skip it, but should I?

src/api/app/controllers/trigger_controller.rb
vpereira: notify the admin is IMO much worse than give the user a feedback. The user should receive an error message and if the user thinks that is necessary, forward the message to the admin. To generate airbrake  notifications with - eventually trash - is IMO just wrong.

src/api/app/services/package_controller_service/rebuild_trigger.rb
vpereira: fixed

danidoni:
If the refactor doesn't change the request, the cassettes shouldn't re-record... So those changes are not really needed. Did they slipped in unnoticed?

danidoni:
If the refactor doesn't change the request, the cassettes shouldn't re-record... So those changes are not really needed. Did they slipped in unnoticed?

src/api/app/controllers/webui/package_controller.rb
hennevogel: Good question. We can talk if you want to!

hennevogel: postponed to later

PR #9140, opened 2020-02-27
Allow authenticated users to see latest_updates card on the main page.

matthewdva:
The "latest_updates" card is only displayed if allow_anonymous configuration
option is allowed.  This patch allows authenticated users to get the
"latest_updates" card even if allow_anonymous is not allowed.

### DIFF
--- a/src/api/Gemfile.lock
+++ b/src/api/Gemfile.lock
@@ -251,7 +251,7 @@ GEM
     mousetrap-rails (1.4.6)
     mysql2 (0.5.2)
     nio4r (2.5.2)
-    nokogiri (1.10.5)
+    nokogiri (1.10.8)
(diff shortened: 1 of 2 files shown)


hennevogel:
Sorry but we only backport fixes for serious problems into the release branches.

matthewdva:
Thank you for the information related to back ports.  I apologize if this was covered in the contribute guide and I missed it.  Do you automatically back port  if the change meets the "serious" criteria?

hennevogel:
Oh don't worry, sorry this wasted some of your cycles. You are right, it is **not** covered in the contribution guide. I'll add it!

We don't do automatic backports no. Because the "serious" criteria is actually a conversation between the team, not something you can automate :smile:

PR #9141, opened 2020-02-27
Allow authenticated users to see latest_updates card on the main page.

matthewdva:
The "latest_updates" card is only displayed if allow_anonymous configuration
option is allowed.  This patch allows authenticated users to get the
"latest_updates" card even if allow_anonymous is not allowed.


hennevogel:
Thanks! :bouquet:

PR #9143, opened 2020-02-27
[dist] enable xforward in setup-appliance.sh

M0ses:
Without this patch starting apache in a setup with setup-appliance.sh runs into the following error:

```
Feb 27 08:09:55 wizard-install start_apache2[5138]: AH00526: Syntax error on line 86 of /etc/apache2/vhosts.d/obs.conf:
Feb 27 08:09:55 wizard-install start_apache2[5138]: Invalid command 'XForward', perhaps misspelled or defined by a module not included in the server configuration
```


hennevogel:

PR #9153, opened 2020-02-27
Fixed logout redirection when anonymous access is not allowed.

matthewdva:
If anonymous access is not allowed and a user logs out the
user is redirected back to the referring page prior to logout.  This
redirection produces a error since anything other than the root path
is now unavailable to the user.

This patch will redirect the use to the root path on logout if anonymous
access is not allowed.


hennevogel:
keep em coming @matthewdva :smile:  :bouquet:

PR #9158, opened 2020-02-28
Fix table width when tab is switched in small screens

vpereira:
Both tables have the same column widths defined with column.width option.
However table in “Tab 2” has column widths different from the table in “Tab 1”.

After click we have to adjust the columns

Before:

Screenshot_2020-02-28_13-03-54

After:

Screenshot_2020-02-28_13-05-41

### DIFF
--- a/src/api/app/assets/javascripts/webui/requests_table.js
+++ b/src/api/app/assets/javascripts/webui/requests_table.js
@@ -1,4 +1,9 @@
 $(document).ready(function() {
+  $('#requests').find('a[data-toggle="tab"]').on('shown.bs.tab', function(){
+    $($.fn.dataTable.tables(true)).DataTable()
+       .columns.adjust();
+  });
+
   $('.requests-datatable').each(function(){
     // 1. Create DataTable
     var dataTableId = $(this).attr('id');


vpereira:

src/api/app/assets/javascripts/webui/requests_table.js
```diff
@@ -1,4 +1,9 @@
 $(document).ready(function() {
+  $('a[data-toggle="tab"]').on('shown.bs.tab', function(){
```
hennevogel: Hm on all tabs? This sounds a bit overkill.

vpereira: ```$('a[data-toggle="tab"]').length => 4``` sounds ok to me. We don't have other tabs in the view.

hennevogel: But this is javascript is not scoped to the page, it's global.

vpereira: oh because the compiled assets right? ok i will scope it better

vpereira: done

PR #9167, opened 2020-03-02
Fix missing public key from project directory.

matthewdva:
Problem: 
   When a global signing key is defined and BSConfig::forceprojectkeys is 
not set, the default public key defined in BSConfig::keyfile is not placed
in the project/projid.pkg directory.  

This creates a problem for local builds.  Packages are signed using the default
key. When the packages are used during local builds the verification fails because there
isn't a public key to validate the packages from that project.

The default public key is correctly added to the repository during
publishing.

This patch mimics the process the publish routine uses to place the default
pubkey file during project creation.

### DIFF
--- a/src/backend/bs_srcserver
+++ b/src/backend/bs_srcserver
@@ -1893,6 +1893,14 @@ sub putproject {
   if ($BSConfig::forceprojectkeys) {
     my ($sk) = getsignkey({}, $projid);
     createkey({ %$cgi, 'comment' => 'autocreate key' }, $projid) if $sk eq '';
+  } elsif ($BSConfig::keyfile) {
+    if (-e $BSConfig::keyfile) {
+       my ($pk) = readstr($BSConfig::keyfile);
+       mkdir_p("$projectsdir/$projid.pkg") || die("creating $projectsdir/$projid.pkg: $!\n");
+       writestr("$projectsdir/$projid.pkg/_pubkey", undef, $pk);
+    } else {
+      print "WARNING: configured default sign key $BSConfig::keyfile does not exist.\n";
+    }
   }
 
   my %except = map {$_ => 1} qw{title description person group url attributes};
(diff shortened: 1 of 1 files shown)


mlschroe:
I don't really like to do it this way. Is this about osc? If yes, we need to fix osc understanding of what the public key is.

matthewdva:
This is an osc issue.  However I have looked at OSC and am unclear how to isolate the fix in osc alone, if at all.  In this scenario, all projects are signed with the global key.  However the source server only writes the global public key to published repositories when the forceprojectkeys is not set.  OSC will traverse parents in the source service to find a public key for a project.  However there is no parent public key in this scenario.

The public key is defined in BSConfig.pm but is not constrained to be located within the source server tree.  The global public key needs to be available via the source service at some level.

The only way I can think to adjust osc to get the public key would be to have it grab the key from the published repo.  I don't think this is viable option for a few reasons.

Changing osc to get the key from a published repository will break getting keys from DOD projects since DOD projects do not publish repositories.  Since a project can have multiple repositories, some enabled / active and other that could be stale other keys, osc should check all possible repositories to ensure it gets all the available public keys.  This will have osc making a lot more requests.

Changing osc behavior to address this scenario will ultimately slow osc and generate more load.

I admit the proposed solution is not the best approach.  However it may be the least risky solution.  The root issue is when using a global  signing key, a parent pubkey is not available.  A better approach would simply ensure that the global pubkey is the top most pubkey in the source service.  I have not researched how to make this change.  I suspect this change poses more risk that the current proposal.  

I sincerely respect your response to my proposal.  This a very mature product any backend changes pose a risk to the stability of the product.  If this still is not the correct approach, I trust you to identify the correct solution.

Thank you for your time and consideration.

mlschroe:
Osc is supposed to use the_keyinfo method to get the pubkey used for signing a project.
i.e. `osc api /source/<project>/_keyinfo`

I'll talk with the osc guys about switching osc to do this.

lethliel:
I will fix this in osc.
The issue there is: https://github.com/openSUSE/osc/issues/781

PR #9169, opened 2020-03-02
Increase size of search box and center it in smaller screens

vpereira:

Before:
Screenshot_2020-03-02_12-36-04

After:
Screenshot_2020-03-02_12-35-35


vpereira:

PR #9171, opened 2020-03-02
Make HTTPS communication between services possible

andrewshadura:
When workers and services are located in different networks, we don’t want to send traffic unencrypted and open to anyone to listen to.

This change makes it possible to easily opt in to HTTPS communication by changing the service URLs in BSConfig.


mlschroe:
I think you must also symlink BSSSL.pm into the `worker` directory.

andrewshadura:
True.

andrewshadura:
Argh. I have accidentally pushed an unrelated branch but then deleted and re-pushed it, but it has closed the pull request.

andrewshadura:
Please see #9189 for a new PR.

PR #9178, opened 2020-03-04
Fix brakeman complain on user model

vpereira:
The construction ` Project.where("name LIKE '#{home_project_name}:%'")` even though safe in the context being used, should be avoided.

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -644,7 +644,7 @@ class User < ApplicationRecord
     save!
 
     # wipe also all home projects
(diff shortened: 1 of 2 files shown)


coolo:
Some quotes to mark 'where' as keyword would have been nice - "Escape where query on" sounds pretty jodaish if read as plain english :)

vpereira:
> Some quotes to mark 'where' as keyword would have been nice - "Escape where query on" sounds pretty jodaish if read as plain english :)

the geek version from https://www.youtube.com/watch?v=ytotGK6YQ40

PR #9179, opened 2020-03-04
Fix SQL injection vector

danidoni:
Since the introduction of the [brakeman](https://github.com/openSUSE/open-build-service/pull/8841) webhook check, #9048 fails due to some potential SQL injection vulnerability.

This PR addresses that security issue in order to allow the PR to pass


vpereira:
it is already being handled on #9178

PR #9184, opened 2020-03-04
Use brakeman instead of unmaintained hakiri for 2.10

krauselukas:
We moved to brakeman, to check the codebase for security
vulnerabilities, on the open-build-service master branch (>2.10).
In order to get rid of hakiri for the whole project, we need
to backport those changes, which only affect the CI setup.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,6 +231,24 @@ jobs:
           path: src/api/coverage_results
           destination: raw_coverage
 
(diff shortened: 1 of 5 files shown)


krauselukas:
Hakiri is complaining about CVE-2020-5247 related to the Puma Server. We only use Puma for developement and testing, not production. So I think this can be ignored. Hakiri will be dropped after this gets merged.

PR #9188, opened 2020-03-05
Actions for main#index

dmarcoux:
Before with actions and generic links all over the page.
before

Now, the actions are in their own menu and generic links in the existing menu where we already have `Profile` and other links. The icons don't have a color anymore as more colors didn't fit well on a dark background. The menu with generic links is also available to anonymous users whenever links make sense for such users (like `All Projects` or `Status Monitor`). The button for the actions menu is hidden whenever it's empty or for anonymous users. Depending on the role of the user, the actions and the generic links will vary to follow the user's permissions.

------
Actions
Large devices:
actions

Small devices:
actions-small

-----

Generic menu
Large devices:
menu

Small devices:
menu-small

### DIFF
--- a/src/api/app/assets/javascripts/webui/responsive_ux.js
+++ b/src/api/app/assets/javascripts/webui/responsive_ux.js
@@ -3,14 +3,27 @@ $(function () {
     $('a[data-toggle="offcanvas"]').toggleClass('active');
     $('.offcanvas-collapse').toggleClass('open');
     $('.watchlist-collapse').removeClass('open');
+    $('.actions-collapse').removeClass('open');
     $('a[data-toggle="watchlist"]').removeClass('active');
+    $('a[data-toggle="actions"]').removeClass('active');
   });
 
   $('[data-toggle="watchlist"]').on('click', function () {
     $('a[data-toggle="watchlist"]').toggleClass('active');
     $('.watchlist-collapse').toggleClass('open');
     $('.offcanvas-collapse').removeClass('open');
+    $('.actions-collapse').removeClass('open');
     $('a[data-toggle="offcanvas"]').removeClass('active');
+    $('a[data-toggle="actions"]').removeClass('active');
+  });
+
+  $('[data-toggle="actions"]').on('click', function () {
+    $('a[data-toggle="actions"]').toggleClass('active');
+    $('.actions-collapse').toggleClass('open');
+    $('.offcanvas-collapse').removeClass('open');
+    $('.watchlist-collapse').removeClass('open');
+    $('a[data-toggle="offcanvas"]').removeClass('active');
+    $('a[data-toggle="watchlist"]').removeClass('active');
   });
 
   $('.access-modal').on('show.bs.modal', function () {
--- a/src/api/app/views/layouts/webui/responsive_ux/_bottom_navigation.html.haml
+++ b/src/api/app/views/layouts/webui/responsive_ux/_bottom_navigation.html.haml
@@ -15,17 +15,17 @@
             .badge.badge-primary.align-text-top= tasks
           %br
           .small Tasks
-      %li.nav-item.border-left.border-gray-500
-        = link_to('javascript:void(0)', class: 'nav-link text-light p-1', alt: 'Menu', data: { toggle: 'offcanvas' }) do
-          %i.fas.fa-bars.fa-lg
-          %br
-          .small Menu
     - else
       - unless kerberos_mode? || !can_sign_up?
         %li.nav-item.border-right.border-gray-500
           = sign_up_link(css_class: 'nav-link text-light p-1')
       %li.nav-item
         = log_in_link(css_class: 'nav-link text-light p-1')
+    %li.nav-item.border-left.border-gray-500
+      = link_to('javascript:void(0)', class: 'nav-link text-light p-1', alt: 'Menu', data: { toggle: 'offcanvas' }) do
+        %i.fas.fa-bars.fa-lg
+        %br
+        .small Menu
 
 - if User.session
   = render partial: "layouts/#{responsive_namespace}/watchlist"
--- a/src/api/app/views/layouts/webui/responsive_ux/_bottom_navigation.html.haml
+++ b/src/api/app/views/layouts/webui/responsive_ux/_bottom_navigation.html.haml
@@ -15,6 +15,12 @@
             .badge.badge-primary.align-text-top= tasks
           %br
           .small Tasks
+      - if content_for?(:actions)
+        %li.nav-item.border-left.border-gray-500
+          = link_to('javascript:void(0)', class: 'nav-link text-light p-1', alt: 'Actions', data: { toggle: 'actions' }) do
+            %i.fas.fa-ellipsis-v.fa-lg
+            %br
+            .small Actions
     - else
       - unless kerberos_mode? || !can_sign_up?
         %li.nav-item.border-right.border-gray-500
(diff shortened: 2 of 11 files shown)


dmarcoux:
[Contextual Links](https://trello.com/c/bLctMUJL/1336-contextual-links)

src/api/app/views/layouts/webui/responsive_ux/_user_navigation.html.haml
```diff
@@ -17,6 +17,23 @@
...
+        %i.fas.fa-list.fa-lg.mr-2
+        All Projects
+    %li.nav-item
+      = link_to(new_project_path, class: 'nav-link') do
+        %i.fas.fa-plus-square.fa-lg.mr-2
+        Create Project
+    %li.nav-item
+      = link_to(image_templates_path, class: 'nav-link') do
+        %i.fas.fa-compact-disc.fa-lg.mr-2
+        New Image
```
hennevogel: Create Project and New Image sound like "actions" to me.

src/api/app/assets/javascripts/webui/responsive_ux.js
```diff
@@ -3,14 +3,27 @@ $(function () {
...
   $('[data-toggle="watchlist"]').on('click', function () {
     $('a[data-toggle="watchlist"]').toggleClass('active');
     $('.watchlist-collapse').toggleClass('open');
     $('.offcanvas-collapse').removeClass('open');
+    $('.contextual-actions-collapse').removeClass('open');
     $('a[data-toggle="offcanvas"]').removeClass('active');
+    $('a[data-toggle="contextual-actions"]').removeClass('active');
+  });
+
+  $('[data-toggle="contextual-actions"]').on('click', function () {
```
dmarcoux: I'm on the fence if I should DRY this or not. What do you think?

danidoni: Go for it!

src/api/app/views/layouts/webui/responsive_ux/_navigation.html.haml
```diff
@@ -14,12 +14,18 @@
             %i.fas.fa-tasks.fa-lg
             - if tasks.positive?
               %span.badge.badge-primary.align-text-top= tasks
-        .toggler{ data: { toggle: 'offcanvas' } }
-          = image_tag_for(User.session, size: 40, custom_class: 'rounded-circle bg-light')
+        - if content_for?(:contextual_actions)
+          .toggler
+            = link_to('javascript:void(0)', class: 'nav-link text-light', alt: 'Contextual Actions', data: { toggle: 'contextual-actions' }) do
+              %i.fas.fa-th.fa-lg
```
saraycp: What about using `ellipsis-v` instead? I think nowadays most of the people is used to the three dots icon to find extra actions on the current element. It would look like:

3dots

src/api/app/views/layouts/webui/responsive_ux/_bottom_navigation.html.haml
```diff
@@ -11,15 +11,19 @@
           %i.fas.fa-tasks.fa-lg
           - if tasks.positive?
             .badge.badge-primary.align-text-top= tasks
-      %li.nav-item.border-left.border-gray-500
-        = link_to('javascript:void(0)', class: 'nav-link text-light', alt: 'Menu', data: { toggle: 'offcanvas' }) do
-          %i.fas.fa-bars.fa-lg
+      - if content_for?(:contextual_actions)
+        %li.nav-item.border-left.border-gray-500
+          = link_to('javascript:void(0)', class: 'nav-link text-light', alt: 'Contextual Actions', data: { toggle: 'contextual-actions' }) do
+            %i.fas.fa-th.fa-lg
```
saraycp: What about using `ellipsis-v` instead? I think nowadays most of the people is used to the three dots icon to find extra actions on the current element. It would look like:

3dots

dmarcoux: Good idea, I've changed the icon for both navigation (small and large devices).

src/api/app/views/webui/main/_contextual_actions.html.haml
```diff
@@ -0,0 +1,18 @@
+- content_for :contextual_actions do
```
hennevogel: I don't think they are in context of the main page only. Those should just always be there no?

hennevogel: Thinking about it... How about

- move `Configuration` to the avatar menu, it's a place not an action
- show `New Image` only in the `Project` actions (would probably imply to move away from branching later)
- show `Create Status Message` in the main page actions (if you're an Admin)
- move `Create Project` to the `Configuration` actions. Creating top level projects is only possible for admins anyway...

dmarcoux: >     * move `Configuration` to the avatar menu, it's a place not an action

I agree.

>     * show `New Image` only in the `Project` actions (would probably imply to move away from branching images later)

Alright, this will be done in a following PR for project#show (it's the next in line).

>     * show `Create Status Message` in the main page actions

This is already the case, so nothing has to change for this.

>     * move `Create Project` to the `Configuration` actions. Creating top level projects is only possible for admins anyway...

Users can create projects which aren't top level as its explained with this message above the form on the page:

`All users have a homes project by default. Yours is home:user_1, and you can also create sub projects inside. New projects outside of your home project can get requested via email.`

With this in mind, I would keep `Create Project` on the main page actions.

hennevogel: sorry for the back and forth...

dmarcoux: No worries, please have another look.

dmarcoux:
I've rebased on master to follow the changes from @saraycp's PR and the screenshots have been updated.

src/api/app/views/layouts/webui/responsive_ux/_nobody_navigation.html.haml
```diff
@@ -1,2 +1,22 @@
-= sign_up_link(css_class: 'nav-link text-light text-center p-1 toggler')
-= log_in_link(css_class: 'nav-link text-light text-center p-1 toggler')
+.toggler
+  = sign_up_link(css_class: 'nav-link text-light text-center p-1')
+.toggler
+  = log_in_link(css_class: 'nav-link text-light text-center p-1')
+.toggler
+  = link_to('javascript:void(0)', class: 'nav-link text-light', alt: 'Generic Links', data: { toggle: 'offcanvas' }) do
+    %i.fas.fa-user-circle.fa-2x
```
saraycp: Sorry, I didn't realized before this icon was added when nobody is logged in.
Why don't we use a "hamburguer" menu icon instead? It's content is not related to a user at all, as we can see in this image:
Screenshot_2020-03-10 Open Build Service

saraycp: In fact, I would say that when logged in, the gravatar should be replaced by the "hamburger" menu also. Since it contains links like "Status Monitor" and not only user-related things. But it can be done in another PR.

dmarcoux: I agree! I'll do it now :)

dmarcoux: So it's okay to address those 2 changes in a separate PR?

saraycp: sure!

PR #9192, opened 2020-03-05
Align navigation items adding always icons and texts for consistency and clarity

saraycp:
In top and bottom navigation bars, having only icons wasn't clear enough, so we add texts under them.

For consistency, we also add icons to those which didn't have, like _Sign Up_ and _Log In_.

Therefore, the bars height and elements sizes and spacing had to be adapted.

### No login:

compare_small_nobody

###  Login:

compare_small_logged

### No login:

compare_large_nobody

### Login:

compare_large_logged

### DIFF
--- a/src/api/app/assets/stylesheets/webui/responsive_ux.scss
+++ b/src/api/app/assets/stylesheets/webui/responsive_ux.scss
@@ -1,3 +1,9 @@
+// Variables
+// TODO: move to app/assets/stylesheets/webui/bootstrap_variables/sizes.scss
+$top-bar-height: 57px;
+$bottom-bar-height: 69px;
+$md-top-bar-height: 69px;
+
 body.responsive-ux {
   // open space to not overlap top navigation with content
   padding-top: 3.5rem !important;
@@ -6,16 +12,25 @@ body.responsive-ux {
   @import 'responsive_ux/build-result';
   @import 'responsive_ux/sponsors';
 
-  .build-results .sticky-top { top: 57px; }
+  .build-results .sticky-top { top: $top-bar-height; }
 }
 
 // FIXME: move to somewhere else?
 
 @include media-breakpoint-up(md) {
+
   body.responsive-ux {
     // START LAYOUT
+    .flash-and-announcement.sticky-top { top: $md-top-bar-height + 4px; }
+    #top_0.sticky-top { top: $md-top-bar-height; }
+
+    #navigation {
+      .fixed-top { height: $md-top-bar-height; }
+      .navbar-brand img { height: 42px; }
+    }
+
     .watchlist-collapse, .offcanvas-collapse {
-      top: 57px; /* Height of navbar */
+      top: $md-top-bar-height; /* Height of navbar */
       bottom: 0;
       width: 380px;
 
--- a/src/api/app/assets/stylesheets/webui/responsive_ux/_layout.scss
+++ b/src/api/app/assets/stylesheets/webui/responsive_ux/_layout.scss
@@ -1,17 +1,19 @@
 .border-gray-500 { border-color: $gray-500 !important; }
 .navbar-dark .navbar-toggler { color: $gray-300; }
-.flash-and-announcement.sticky-top { top: 60px; }
-#top_0.sticky-top { top: 57px; }
+.flash-and-announcement.sticky-top { top: $top-bar-height + 16px; }
+#top_0.sticky-top { top: $top-bar-height; }
 
 // NAVIGATION
 
 #navigation {
-  .fixed-top { height: 57px; }
+  .fixed-top { height: $top-bar-height; }
   .toggler {
-    @extend .ml-3;
+    @extend .ml-2;
     @extend .d-none;
     @extend .d-md-block;
     cursor: pointer;
+    &.text-center { min-width: 62px; }
+    &:first-of-type { @extend .ml-3;  }
   }
 
   .navbar-nav .nav-link {
@@ -21,18 +23,26 @@
 
   .menu-right { cursor: pointer; }
   .menu-options { font-size: 1.1rem }
+
+  .navbar-brand img { height: 30px; }
 }
 
-#bottom-navigation { height: 58px; }
+#bottom-navigation { height: $bottom-bar-height; }
 
 #navigation, #bottom-navigation {
   a.nav-link {
+    .small {
+      line-height: 1rem;
+      padding-top: 0.25rem;
+    }
+
     &.active {
       background-color: darken($dark, 10%);
       @extend .rounded;
     }
   }
 }
+
 // BREADCRUMBS
 
 .breadcrumb {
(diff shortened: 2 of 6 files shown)


src/api/app/assets/stylesheets/webui/responsive_ux.scss
```diff
@@ -6,16 +12,25 @@ body.responsive-ux {
...
+  .build-results .sticky-top { top: $top-bar-height; }
 }
 
 // FIXME: move to somewhere else?
 
 @include media-breakpoint-up(md) {
+
   body.responsive-ux {
     // START LAYOUT
+    .flash-and-announcement.sticky-top { top: $md-top-bar-height + 16px; }
```
krauselukas: @saraycp to me the 16px extra space is a bit much, maybe going with 8px? :)

src/api/app/assets/stylesheets/webui/responsive_ux.scss
```diff
@@ -1,3 +1,9 @@
+// Variables
+// TODO: move to app/assets/stylesheets/webui/bootstrap_variables/sizes.scss
```
krauselukas: @saraycp is there are reason to leave it here with the TODO comment, or is it a left over?

src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -368,18 +368,38 @@ def responsive_namespace
 
   def sign_up_link(css_class: nil)
     return unless can_sign_up?
+
     if proxy_mode?
-      link_to('Sign Up', sign_up_params[:url], class: css_class)
+      link_to(sign_up_params[:url], class: css_class) do
+        link_content('Sign Up', css_class, 'fa-signature')
```
hennevogel: Hm, this is the first time I have seen this icon in this context. Not sure this helps. How about `user-plus`?

Screenshot from 2020-03-06 11-36-44

DavidKang:
I still don't see the icons for login and sign up. The icons don't bring any value, in fact, they are polluting, trying to put icons that could never fit because we don't need them. IMO we are over using the icons. And regarding consistency,
at least for me, the login and sign up has nothing  to do with the "button actions" of the navbars.

src/api/app/helpers/webui/webui_helper.rb
saraycp: In fact, that was my first idea... Let's see if we finally use icons for Log In and Sign up.

src/api/app/assets/stylesheets/webui/responsive_ux.scss
saraycp: The idea is to move this part to the correct file as soon as we move it out of the beta program.

src/api/app/assets/stylesheets/webui/responsive_ux.scss
saraycp: You are right, and I think even 4px is enough. Updated!

src/api/app/helpers/webui/webui_helper.rb
saraycp: I've already replaced the signature icon with `user-plus` one and updated the screenshots on the PR's description.

saraycp:
@DavidKang, @danidoni, @vpereira  My idea was giving the same aspect to the bar, not only regarding icons and texts but also height of the bar.

Another thing to have into account is that soonish there could be another action in the bar, the Menu one, which will include links to `All projects`, `Status Monitor` ... that should be visible even when nobody is logged in.

Yours is also a good reasoning, but I think I'm going to leave it like it is for now.

DavidKang:
@saraycp, I'm fine with your decision 🙂

src/api/app/assets/stylesheets/webui/responsive_ux/_layout.scss
```diff
@@ -21,18 +23,26 @@
...
+      padding-top: 0.25rem;
+    }
+
     &.active {
       background-color: darken($dark, 10%);
       @extend .rounded;
     }
   }
 }
+
```
danidoni: Is this intended?

src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -368,18 +368,38 @@ def responsive_namespace
...
     return unless can_sign_up?
+
     if proxy_mode?
-      link_to('Sign Up', sign_up_params[:url], class: css_class)
+      link_to(sign_up_params[:url], class: css_class) do
+        link_content('Sign Up', css_class, 'fa-user-plus')
+      end
     else
-      link_to('Sign Up', '#', class: css_class, data: { toggle: 'modal', target: '#sign-up-modal' })
+      link_to('#', class: css_class, data: { toggle: 'modal', target: '#sign-up-modal' }) do
```
danidoni: I think we use `'javascript:void(0)'` instead of `#`

danidoni: Well, you use it [here](https://github.com/openSUSE/open-build-service/pull/9192/files#diff-10c13e183f1904c9cf862bd0a498e5fbR6) so you already know it :+1:

src/api/app/assets/stylesheets/webui/responsive_ux/_layout.scss
saraycp: Sorry, I don't know what you mean.

src/api/app/helpers/webui/webui_helper.rb
saraycp: I just moved code. However, AFAIK, the `'javascript:void(0)'` was added to avoid an unwanted jump/scrolling when opening an offcanvas. In case of _Log In_ and _Sign up_ we open a modal so it is not needed.

src/api/app/assets/stylesheets/webui/responsive_ux/_layout.scss
vpereira: @danidoni  means the space on the line 45 @saraycp

saraycp: So yes! it's intended, I added an empty line before and after each comment.

src/api/app/assets/stylesheets/webui/responsive_ux.scss
```diff
@@ -1,3 +1,9 @@
...
+$top-bar-height: 57px;
```
dmarcoux: This is a small, but great change! :+1:

src/api/app/helpers/webui/webui_helper.rb
danidoni: Ok

src/api/app/assets/stylesheets/webui/responsive_ux/_layout.scss
danidoni: Ok!

PR #9193, opened 2020-03-06
Add group to role request dialog

quatran:


krauselukas:
@quatran Congrats for your first PR to the project! :tada: Let's see on monday how to fix the tests and if I remember correctly, we already have something in the codebase to auto-complete group names :) maybe we can use it here as well, let's see

danidoni:
Congrats on your first PR! :clap:

src/api/app/views/webui/request/_add_role_request_dialog.html.haml
```diff
@@ -18,9 +18,24 @@
...
+                  %a{ 'id' => 'user-tab', 'data-toggle' => 'tab', 'href' => '#user', 'class' => 'nav-link text-nowrap active',
+                   'role' => 'tab', 'aria-controls' => 'user', 'aria-selected' => 'true', 'onclick' => 'deleteGroupInput()' }
+                    For user
+                %li.nav-item
+                  %a{ 'id' => 'group-tab', 'data-toggle' => 'tab', 'href' => '#group', 'class' => 'nav-link text-nowrap',
+                   'role' => 'tab', 'aria-controls' => 'group', 'aria-selected' => 'false', 'onclick' => 'deleteUserInput()' }
+                    For groups
+              %div{ 'class' => 'tab-content' }
+                %div{ 'id' => 'user', 'class' => 'tab-pane fade show active', 'role' => 'tab', 'aria-labelledby' => 'user-tab' }
+                  = text_field_tag(:user, nil, id: 'user-text-field', class: 'form-control', placeholder: 'Enter username',
```
dmarcoux: As @krauselukas said, we have an autocomplete component which would be perfect for this PR. It's listed in our [patterns library](https://obs-patterns.netlify.com/molecules/autocomplete.html).

For a concrete example of how it's used for groups, have a look [here](https://github.com/openSUSE/open-build-service/blob/cc1d6484273fbb98bd2d96ca52917dd052e3bd97/src/api/app/views/webui/shared/user_or_groups_roles/_add_group_dialog.html.haml#L8-L9). For users, it's almost the same and you can check [here](https://github.com/openSUSE/open-build-service/blob/cc1d6484273fbb98bd2d96ca52917dd052e3bd97/src/api/app/views/webui/shared/user_or_groups_roles/_add_user_dialog.html.haml#L8-L9).

src/api/app/views/webui/request/_add_role_request_dialog.html.haml
```diff
@@ -18,9 +18,24 @@
...
+                  onclick: 'deleteGroupInput()')
+                %div{ 'id' => 'group', 'class' => 'tab-pane fade', 'role' => 'tab', 'aria-labelledby' => 'group-tab' }
```
dmarcoux: This is really not a big deal and I don't know if you're new to HAML or not, but just in case you are, the `id` and `class` properties can be written directly after the tag instead of being passed in the attributes (so inside the curly brackets). This would result in:

`%div.tab-pane.fade#group{ 'role' => 'tab', ... }`.

One more trick, the `div` is so commonly used that it can be skipped! So it results in:

`.tab-pane.fade#group{ 'role' => 'tab', ... }`

Voilà!

quatran: Thanks for the useful tips. I'm fairly new to HAML :)

src/api/app/assets/javascripts/webui/request.js
```diff
@@ -79,25 +79,17 @@ function prefillSubmitRequestForm() {
...
 }
 
-function requestAddReviewAutocomplete() { // jshint ignore:line
+function requestAddAutocomplete(element) { // jshint ignore:line
   $('.modal').on('shown.bs.modal', function() {
     $('.hideable input:not(:visible)').attr('disabled', true);
   });
 
-  $('#review_type').change(function () {
+  $(element).change(function () {
```
hennevogel: Element is the most general base class from which all element objects (i.e. objects that represent elements) in a Document inherit.  -> https://developer.mozilla.org/en-US/docs/Web/API/Element

src/api/spec/features/webui/requests_spec.rb
```diff
@@ -101,6 +105,11 @@
         click_link 'Request role addition'
         within('#add-role-modal') do
           find(:id, 'role').select('Maintainer')
+          find(:id, 'role_type').select('User')
```
krauselukas, line 108: @quatran hey, the PR looks good to me so far, I have one suggestion. You already adjusted this feature test for the request role addition of a User. Maybe you could take over this logic and do the same for the group :)  If you need help, let me know. But first step would be to see how its done here for the User.

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -334,16 +334,16 @@ def forward_request_to(fwd)
...
-      target_project: params[:project],
-      target_package: params[:package],
-      source_project: params[:devel_project],
-      source_package: params[:devel_package] || params[:package],
-      target_repository: params[:repository],
-      person_name: params[:user],
-      group_name: params[:group],
-      role: params[:role],
-      type: type.to_s
-    }
```
vpereira, line 347: can you explain why is it better than before?

quatran: We pass the params of user or group only if they are present. Otherwise it would result into a crashing of the follow up functions which uses `request_action_attributes`

PR #9196, opened 2020-03-09
[dist] initial version of perl-OBS-XML

M0ses:
Create an extra packge for XML::Structured DTD's from BSXML name OBS::XML


M0ses:
For the record: reviewed by @mlschroe

PR #9201, opened 2020-03-11
Create actions menu for project#show

dmarcoux:
On large devices:
actions-large

On small devices:
actions-small

### DIFF
--- a/src/api/app/views/webui/project/responsive_ux/_actions.html.haml
+++ b/src/api/app/views/webui/project/responsive_ux/_actions.html.haml
@@ -1,0 +1,32 @@
+- content_for :actions do
+  - if (bugowners_mail.present? && !CONFIG['bugzilla_host'].nil?) || User.session
+    - if bugowners_mail.present? && !CONFIG['bugzilla_host'].nil?
+      = render partial: 'webui/project/responsive_ux/actions/report_bug', locals: { bugowners_mail: bugowners_mail, project: project }
+    - if project.is_locked? && User.possibly_nobody.can_modify?(project, true)
+      = render partial: 'webui/project/responsive_ux/actions/unlock_project', locals: { project: project }
+    - elsif User.possibly_nobody.can_modify?(project)
+      = render partial: 'webui/project/responsive_ux/actions/modify_project', locals: { project: project,
+                                                                          packages: packages,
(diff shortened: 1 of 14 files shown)


dmarcoux:
[Contextual Links](https://trello.com/c/bLctMUJL/1336-contextual-links)

DavidKang:
@dmarcoux, we have a side link(patchinfo presente), that does the same as 'show patchinfo'. So maybe se could remove one of them. What do you think about that?

dmarcoux:
> @dmarcoux, we have a side link(patchinfo presente), that does the same as 'show patchinfo'. So maybe se could remove one of them. What do you think about that?

Good point. I will remove the `Show Patchinfo` link and it also allows me to get rid of the CSS for properly aligning the icon stacks since it's the only icon stack.

dmarcoux:
I wonder however if the side links should be in the actions menu. What do you think?

hennevogel:
> I wonder however if the side links should be in the actions menu.

Those links are more information/status then links. I would leave them alone.

src/api/app/views/webui/project/responsive_ux/_actions.html.haml
```diff
@@ -0,0 +1,32 @@
...
+      = render partial: 'webui/project/responsive_ux/actions/delete_project', locals: { project: project }
+    - elsif !project.is_locked?
+      = render partial: 'webui/project/responsive_ux/actions/request_role_addition_and_deletion', locals: { project: project }
+  - elsif project.key_info.present?
+    = render partial: 'webui/project/responsive_ux/actions/gpg_key_ssl_certificate', locals: { project: project }
+
+  %li.nav-item
+    = link_to(image_templates_path, class: 'nav-link') do
+      %i.fas.fa-compact-disc.fa-lg.mr-2
+      New Image
```
krauselukas: @dmarcoux New Image currently appears in the "Actions on this page" section of the actions menu. To me that's more a general action. But not sure if its necessary at this point to introduce an additional area in the action menu for such things. I'm okay with leaving it here

dmarcoux: I was also a bit divided on this one and in the end, I think it's okay to leave it with the others.

PR #9202, opened 2020-03-11
[api] fix build for mariadb-10.4 using the current user

bugfinder:
use the current user instead of root for the mysql test install


hennevogel:
Either there is something missing or changed for SLE15. See the [test builds](https://build.opensuse.org/package/show/OBS:Server:Unstable:TestGithub:PR-9202/obs-server).

bugfinder:
yes, SLE-15 GA on OBS is even worse. working on that.

PR #9203, opened 2020-03-11
Implement in-place editing in project overview page

danidoni:
Use in-place editing for the project's

- Title
- Url
- Description

We've been using modals to edit the project's title, url and description. We
think we have better tools for that. In-place editing is one of the ways to do
it.

With this PR we introduce a new way of updating some of the project's
information from the projects overview page.

**NOTE**: Still missing/not working:

- [x] Cancel button
- [x] Make it work with model validations
- [x] Refresh only the affected area instead of the whole screen
- [ ] Visual feedback when switching from editing to read only

**Before**
Screenshot of the project monitor

**After**
Screenshot of the project monitor

Co-authored-by: Saray Cabrera Padrón scabrerapadron@suse.de

### DIFF
--- a/src/api/app/assets/javascripts/webui/application.js
+++ b/src/api/app/assets/javascripts/webui/application.js
@@ -55,3 +55,5 @@
 //= require webui/monitor.js
 //= require webui/airbrake-js
 //= require webui/responsive_ux
+//= require webui/in_place_editing.js
+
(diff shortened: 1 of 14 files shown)


danidoni:
[Modals](https://trello.com/c/cq2te0jh/1332-modals)

saraycp:
Superseded by #9238

PR #9205, opened 2020-03-11
[api] fix build for mariadb-10.4

bugfinder:
build using the current user and try not to break older mariadb
revisions in the process where socket auth was not loaded by default.


bugfinder:
next try ...

PR #9206, opened 2020-03-11
Add "notifiable" polymorphic fields to table "notifications"

eduardoj:
This commit is one of a serie of commits to make notifications a polymorphic association. Later on we will adapt the existing code, and remove the fields that will be not needed, to make a smooth
transition.


dmarcoux:

PR #9208, opened 2020-03-11
Distinguish between subscriptions and notifications

krauselukas:
Since the users tasks are moved to a more notification's
like workflow, we should distinguish between notifications
and the subscriptions to certain events which might end up
as notifications.


danidoni:
Fix test failures and you're good to go

krauselukas:
As discussed, this might be technically correct, but probably not the right thing to do from a user perspective. The solution is somewhere in between.

PR #9214, opened 2020-03-12
Reduce the gap between flash and top navigation bar

saraycp:
This was overlooked in #9192.


dmarcoux:
I had a look and it doesn't seem to change anything...

I have this in my browser's inspector:
flash

saraycp:
> I had a look and it doesn't seem to change anything...

@dmarcoux  I see it correctly locally and in `review-app`. 

From xs to md viewports it should be `top: 61px` and from md on it should be `top: 73px;`.
Take into account I'm just adjusting the gap between the flash box and the top navigation bar. And the top navigation bar has different heights depending on the screen size.

Show home user_3   Open Build Service

PR #9220, opened 2020-03-12
Write data into the new polymorphic fields from the notifications table

eduardoj:
We want to store notification data in the newly created polymorphic fields from the notifications table. We don't modify yet the existing way of reading the notifications table.

### DIFF
--- a/src/api/app/jobs/send_event_emails_job.rb
+++ b/src/api/app/jobs/send_event_emails_job.rb
@@ -26,12 +26,31 @@ class SendEventEmailsJob < ApplicationJob
 
   def create_rss_notifications(event)
     event.subscriptions.each do |subscription|
-      Notification::RssFeedItem.create(
+      notification_params = {
         subscriber: subscription.subscriber,
         event_type: event.eventtype,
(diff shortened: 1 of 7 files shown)


eduardoj:

src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -26,12 +26,31 @@ def perform
...
+    end
+  end
+
+  def notification_dynamic_params(event)
+    case event.eventtype
+    when 'ReviewWanted', 'RequestCreate', 'RequestStatechange'
+      {
+        notifiable_type: 'BsRequest',
+        notification_id: event.payload['id'],
+        bs_request_oldstate: (event.payload['oldstate'] if event.eventtype == 'RequestStatechange')
```
hennevogel: BTW I also think you need to say state, otherwise you will not be able to tell which state it switched to **at the time** of the Notification. Only at which state it is **now**...

eduardoj: Ops, you're right. Pull request is in the way.

PR #9224, opened 2020-03-13
Add request state field to notifications table

eduardoj:
At the time we access the state, it could already have been changed. We need to store this field when the notification is created.

We store the "state" only when it is needed, for notifications coming from events with type "RequestStatechange". The other request and comment notifications don't make use of this "state" field.

Related to #9206 and #9220.

### DIFF
--- a/src/api/app/jobs/send_event_emails_job.rb
+++ b/src/api/app/jobs/send_event_emails_job.rb
@@ -37,20 +37,29 @@ class SendEventEmailsJob < ApplicationJob
   end
 
   def notification_dynamic_params(event)
-    case event.eventtype
-    when 'ReviewWanted', 'RequestCreate', 'RequestStatechange'
-      {
+    if event.eventtype == 'RequestStatechange'
+      return {
         notifiable_type: 'BsRequest',
         notification_id: event.payload['id'],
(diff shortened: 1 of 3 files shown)


src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -38,12 +38,18 @@ def create_rss_notifications(event)
...
         notifiable_type: 'BsRequest',
         notification_id: event.payload['id'],
-        bs_request_oldstate: (event.payload['oldstate'] if event.eventtype == 'RequestStatechange')
-      }.compact
+        bs_request_state: event.payload['state'],
+        bs_request_oldstate: event.payload['oldstate']
+      }
+    when 'ReviewWanted', 'RequestCreate'
+      {
+        notifiable_type: 'BsRequest',
```
hennevogel: :thought_balloon:  Can't you just use a guard clause inside the `when`?

eduardoj: Sure. On it.

eduardoj: Done. Please re-review.

eduardoj:

PR #9225, opened 2020-03-14
Add polymorphic association between BsRequest, Comment and Notification Model

krauselukas:
In order to query the comments and request related to a
notification, the polymorphic association needs to be
indicated in the models.

Related to #9206 and #9220.


krauselukas:

PR #9230, opened 2020-03-16
Add polymorphic relationship between notifications users and groups

vpereira:
Add the basic Polymorphic relationship between `Notification`, `User` and `Group`

no migration is needed


src/api/spec/models/notification_spec.rb
```diff
@@ -10,4 +10,18 @@
     it { expect(subject.class).to eq(delete_package_event.class) }
     it { expect(subject.payload).to eq(delete_package_event.payload) }
   end
+
+  describe 'relationship with users' do
```
hennevogel: :thought_balloon:  I don't see much sense in those kinds of tests, there is no logic involved and I'm pretty sure that the `ActiveRecord` test suite covers associations :-)

PR #9233, opened 2020-03-16
Notifications cleanup

vpereira:
We are going to have a Job which from time to time will clean Notifications older than 3 months.

In this PR we handle the Notifications cleanup

### DIFF
--- a/src/api/spec/factories/notification.rb
+++ b/src/api/spec/factories/notification.rb
@@ -6,5 +6,14 @@ FactoryBot.define do
   end
 
   factory :rss_notification, parent: :notification, class: 'Notification::RssFeedItem' do
+    transient do
+      stale { false }
+    end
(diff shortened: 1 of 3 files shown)


src/api/spec/factories/notification.rb
```diff
@@ -6,5 +6,14 @@
   end
 
   factory :rss_notification, parent: :notification, class: 'Notification::RssFeedItem' do
+    transient do
+      stale { false }
+    end
+    after(:create) do |notification, evaluator|
+      if evaluator.stale
+        notification.created_at = 6.months.ago if evaluator.stale
```
eduardoj: `evaluator.stale` is being checked twice. It seems like a left over, isn't it?

vpereira: done

PR #9235, opened 2020-03-16
Fix notifiable id fields in notifications table

eduardoj:
Due to a typo `notification_id` field was introduced when trying to add data to the notifications table. The correct name should have been `notifiable_id`.


eduardoj:

PR #9236, opened 2020-03-16
Validate notifications at creation time

vpereira:
Notification validatation rules:

apply only to users
  - by notification creation, check if the user logged in the last 3 months and if not,  don't create it

apply only to groups
  - by notification creation if any user in the group logged in in the last 3 months and if not, don't create it


vpereira:
I will do this check on the job

PR #9237, opened 2020-03-16
Add method User#away? and update User factory dead_user

vpereira:
Add method User#away? to check if user didn't login in the last 3 months and add User factory `dead_user` to create a "away" confirmed user 

Extracted from #9237


src/api/app/models/user.rb
```diff
@@ -279,6 +279,10 @@ def self.find_by_login!(login)
     raise NotFoundError, "Couldn't find User with login = #{login}"
   end
 
+  def away?
```
danidoni: :thought_balloon: What about using `dead?` here?

PR #9238, opened 2020-03-17
In-place edition for project basic information

saraycp:
This is our suggestion:
- in-place edition for the **project**'s basic information
- instead of editing each form element, we propose to edit the whole form (including Title, URL and Description)
- we use Rails UJS
- we add some JavaScript animations for the transitions from content to form and the other way around

It looks like:

in-place-edition

Validations are going to be shown in the form as well:

in-place-edition-validation

Something similar will be done for packages, after we have a quite stable version for projects.

Please take into account that we tried to add little commits to make it easier for you to follow our reasoning. They will be squashed before becoming a real PR.

### DIFF
--- a/src/api/app/assets/javascripts/webui/responsive_ux.js
+++ b/src/api/app/assets/javascripts/webui/responsive_ux.js
@@ -47,3 +47,21 @@ function toggleTabs(tabLinkContainerId) { // jshint ignore:line
     currentTabLink.addClass('active');
   });
 }
+
+function setFormValidation(element, messages) { // jshint ignore:line
+  element.addClass('is-invalid');
+  element.after("<div class='invalid-feedback'>"+ messages + "</div>");
+}
+
+function resetFormValidation() { // jshint ignore:line
+  $('.in-place-editing form .is-invalid').each(function(){
+    $(this).toggleClass('is-invalid');
+    $(this).siblings('.invalid-feedback').remove();
+  });
+}
+
+function scrollToInPlace() { // jshint ignore:line
+  $('html, body').animate({
+    scrollTop: $('.in-place-editing').offset().top - 73 // header height + 16px
+  }, 500);
+}
--- a/src/api/app/assets/javascripts/webui/responsive_ux.js
+++ b/src/api/app/assets/javascripts/webui/responsive_ux.js
@@ -34,6 +34,11 @@ $(function () {
     $('#login').focus();
     $('#username').focus();
   });
+
+  $('#navigation .nav-item').on('click', function (){
+    $('.actions-collapse').removeClass('open');
+    $('a[data-toggle="actions"]').removeClass('active');
+  });
 });
 
 /* bootstrap's tabs javascript doesn't remove the active class
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -10,7 +10,7 @@ class Webui::ProjectController < Webui::WebuiController
                                        :new_release_request, :new_package, :edit_comment]
 
   before_action :set_project, only: [:autocomplete_repositories, :users, :subprojects,
-                                     :new_package,
+                                     :new_package, :edit,
                                      :show, :buildresult,
                                      :destroy, :remove_path_from_target,
                                      :requests, :save, :monitor, :toggle_watch,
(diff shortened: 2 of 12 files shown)


src/api/app/views/webui/project/_basic_info.html.haml
```diff
@@ -0,0 +1,29 @@
+- if flipper_responsive?
+  .basic-info.mb-3{ class: 'in-place-edition' }
```
hennevogel: The word is `editing`, edition is something else.

src/api/app/views/webui/project/_edit.html.haml
```diff
@@ -14,3 +14,10 @@
     %a.cancel.btn.btn-outline-danger.px-4
       Cancel
     = submit_tag('Update', class: 'btn btn-primary px-4')
+
+:javascript
+  $(document).ready(function() {
+    $('.cancel').on('click', function(){
+      location.reload();
```
hennevogel: Hm, I don't understand. Why?

src/api/app/assets/javascripts/webui/responsive_ux.js
```diff
@@ -47,3 +47,15 @@ function toggleTabs(tabLinkContainerId) { // jshint ignore:line
...
+  element.addClass('is-invalid');
+  element.after("<div class='invalid-feedback'>"+ messages + "</div>");
+}
+
+function resetFormValidation() { // jshint ignore:line
+  $('.in-place-edition form .is-invalid').each(function(){
+    $(this).toggleClass('is-invalid');
+    $(this).siblings('.invalid-feedback').remove();
+  });
+}
```
hennevogel: Hm, isn't this something that happens automatically with UJS?

src/api/app/views/webui/project/edit.js.haml
```diff
@@ -0,0 +1 @@
+$('.in-place-edition').html("#{escape_javascript(render(partial: 'webui/project/edit', locals: { project: @project }))}");
```
hennevogel: Why do you use the `.haml` pre-processor? There is not haml in here.

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -226,7 +226,7 @@ def restore
   end
 
   def edit
-    authorize @project, :update?
+    authorize @project, :edit?
```
hennevogel: Why? Do you expect this this to be different?

BTW you can also skip the method argument and it'll be derived from the controller method name you call this in.

src/api/app/views/webui/project/edit.js.haml
```diff
@@ -1 +1,7 @@
-$('.in-place-edition').html("#{escape_javascript(render(partial: 'webui/project/edit', locals: { project: @project }))}");
+:plain
```
hennevogel: You would not have to do this if you just skip the `.haml` pre-processor.

src/api/app/views/webui/project/_basic_info.html.haml
saraycp: Oh, good to know! I replace it in the code and as soon as we squash the commits I'll correct the commit's texts as well.

src/api/app/views/webui/project/_edit.html.haml
saraycp: This was a first attempt, but in the next commit this is replaced by a normal `remote: true` link to use UJS.

src/api/app/assets/javascripts/webui/responsive_ux.js
saraycp: The validations weren't shown, so I had to handle it manually. But let me research a bit more...

src/api/app/views/webui/project/edit.js.haml
saraycp: True! There's no reason to use `.haml` to only render a partial. I'll use `.erb` instead.
For both edit and show.

src/api/app/controllers/webui/project_controller.rb
saraycp: I just created `:edit?` to be consistent with names, but it's true it's exactly the same as using `:update?`. I'll remove the one I created.

src/api/app/views/webui/project/edit.js.haml
saraycp: Ok, `*.js.haml` has been converted to `*.js.erb` and the `:plain`has been removed for both edit and show.

saraycp:
@hennevogel I've addressed your feedback, except the one related to [validations](https://github.com/openSUSE/open-build-service/pull/9238#discussion_r394257811), I'm doing more research in that regard.

PR #9239, opened 2020-03-17
Add title field to notifications table

eduardoj:
We want to store the subject of an event in the notifications table when a notification is inserted in the table, so we add this new `title` field.


eduardoj:
> [...]
> 
> The migration and the code using the new `title` field needs to be deployed separately, so in different PRs.

You are right. Done. This pull request contains now only the migration.

dmarcoux:
Before merging the following PR for this, let's be sure the migration is live in production.

dmarcoux:
@eduardoj @DavidKang The migration is now live.

eduardoj:

PR #9243, opened 2020-03-17
Fix inserting comments and requests, and fill title in notifications

eduardoj:
Fix storing `comments` and `bs_requests` in notifications. The complete name of the class, starting with `Event::`, was needed.

The `title` field of notifications is filled with the result of calling the `subject` method of the event.

### DIFF
--- a/src/api/app/jobs/send_event_emails_job.rb
+++ b/src/api/app/jobs/send_event_emails_job.rb
@@ -43,6 +43,7 @@ class SendEventEmailsJob < ApplicationJob
       return {
         notifiable_type: 'BsRequest',
         notifiable_id: event.payload['id'],
+        title: event.subject,
         bs_request_state: event.payload['state'],
         bs_request_oldstate: event.payload['oldstate']
       }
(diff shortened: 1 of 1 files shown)


src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -37,26 +37,37 @@ def create_rss_notifications(event)
   end
 
   def notification_dynamic_params(event)
-    if event.eventtype == 'RequestStatechange'
+    if event.eventtype == 'Event::RequestStatechange'
       return {
         notifiable_type: 'BsRequest',
```
danidoni: You can extract this return block and the others below into methods. Doing so you'll reduce the method line numbers and fix the Code Climate issues

eduardoj: We refactored the code. Now Code Climate is passing. Please have a look again.

src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -39,29 +39,28 @@ def create_rss_notifications(event)
...
-    if event.eventtype == 'RequestStatechange'
-      return {
-        notifiable_type: 'BsRequest',
-        notifiable_id: event.payload['id'],
-        bs_request_state: event.payload['state'],
-        bs_request_oldstate: event.payload['oldstate']
-      }
-    end
+    return {} unless event.eventtype.in?(['Event::RequestStatechange',
+                                          'Event::RequestCreate',
```
danidoni: You can extract this array into a constant. Maybe something like `EVENTS_TO_NOTIFY`

eduardoj: Done. Please re-review.

eduardoj:

PR #9244, opened 2020-03-17
WIP Convert action navigation into a sidebar 

DavidKang:
We made the action navigation bar always visible for xl and xxl breakpoints. With this change, the actions are always visible and available.

Screenshot_2020-03-17 home Admin

### DIFF
--- a/src/api/app/assets/stylesheets/webui/responsive_ux.scss
+++ b/src/api/app/assets/stylesheets/webui/responsive_ux.scss
@@ -2,32 +2,32 @@
 // TODO: move to app/assets/stylesheets/webui/bootstrap_variables/sizes.scss
 $top-bar-height: 3.5625rem;
 $bottom-bar-height: 3.5rem;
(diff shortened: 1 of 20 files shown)


hennevogel:
I like it, but I'm not sure we should change it right now. We haven't even "tested" the menus on the top right for a while with people.

hennevogel:
I have the feeling for those viewport sizes we should have a full menu on the left. What do you think?

danidoni:
> I have the feeling for those viewport sizes we should have a full menu on the left. What do you think?

That's a good idea

DavidKang:
> I have the feeling for those viewport sizes we should have a full menu on the left. What do you think?

The idea behind this PR, is to move into this direction, having a menu on the left. I started with actions because it was something that I felt was needed. 

So, having a full menu sounds good :smile_cat:

DavidKang:
For the moment, we are not going to this direction. I will close this PR for now.

PR #9247, opened 2020-03-18
Rename bs_request_status column to bs_request_state in notifications table

eduardoj:
The original name of the field in the `bs_requests` table ~was~ is `state`, not `status`.


hennevogel:
Hm, as in `BsRequest` right?

eduardoj:
> Hm, as in `BsRequest` right?

Yes.

eduardoj:

PR #9249, opened 2020-03-19
Read RSS feeds also from new polymorphic notification columns

eduardoj:
As part of our proccess of migrating notifications to new polymorphic columns with zero downtime, here we read RSS feeds also from the new polymorphic columns.

### DIFF
--- a/src/api/app/jobs/send_event_emails_job.rb
+++ b/src/api/app/jobs/send_event_emails_job.rb
@@ -17,21 +17,23 @@ class SendEventEmailsJob < ApplicationJob
         next
       end
 
-      begin
-        create_rss_notifications(event)
-        EventMailer.event(subscribers, event).deliver_now
-      rescue StandardError => e
-        Airbrake.notify(e, event_id: event.id)
-      ensure
-        event.update_attributes(mails_sent: true)
-      end
+      create_notifications(event)
+      send_email(subscribers, event)
     end
     true
   end
 
   private
 
-  def create_rss_notifications(event)
+  def send_email(subscribers, event)
+    EventMailer.event(subscribers, event).deliver_now
+  rescue StandardError => e
+    Airbrake.notify(e, event_id: event.id)
+  ensure
+    event.update_attributes(mails_sent: true)
+  end
+
+  def create_notifications(event)
     return unless event.eventtype.in?(EVENTS_TO_NOTIFY)
 
     event.subscriptions.each do |subscription|
@@ -50,6 +52,8 @@ class SendEventEmailsJob < ApplicationJob
     }
 
     Notification::RssFeedItem.create(notification_params.merge!(notification_dynamic_params(event)))
+  rescue StandardError => e
+    Airbrake.notify(e, event_id: event.id)
   end
 
   def notification_dynamic_params(event)
--- a/src/api/app/jobs/send_event_emails_job.rb
+++ b/src/api/app/jobs/send_event_emails_job.rb
@@ -32,22 +32,27 @@ class SendEventEmailsJob < ApplicationJob
   private
 
   def create_rss_notifications(event)
-    event.subscriptions.each do |subscription|
-      notification_params = {
-        subscriber: subscription.subscriber,
-        event_type: event.eventtype,
-        event_payload: event.payload,
-        subscription_receiver_role: subscription.receiver_role
-      }
+    return unless event.eventtype.in?(EVENTS_TO_NOTIFY)
 
-      next if subscription.subscriber && subscription.subscriber.away?
-      Notification::RssFeedItem.create(notification_params.merge!(notification_dynamic_params(event)))
+    event.subscriptions.each do |subscription|
+      create_rss_notification(event, subscription)
     end
   end
 
-  def notification_dynamic_params(event)
-    return {} unless event.eventtype.in?(EVENTS_TO_NOTIFY)
+  def create_rss_notification(event, subscription)
+    return if subscription.subscriber && subscription.subscriber.away?
 
+    notification_params = {
+      subscriber: subscription.subscriber,
+      event_type: event.eventtype,
+      event_payload: event.payload,
+      subscription_receiver_role: subscription.receiver_role
+    }
+
+    Notification::RssFeedItem.create(notification_params.merge!(notification_dynamic_params(event)))
+  end
+
+  def notification_dynamic_params(event)
     dynamic_params = { notifiable_id: event.payload['id'], notifiable_type: 'BsRequest', title: notification_title(event.subject) }
 
     if event.eventtype == 'Event::RequestStatechange'
--- a/src/api/app/views/notifications/_group_action.text.erb
+++ b/src/api/app/views/notifications/_group_action.text.erb
@@ -1,0 +1,1 @@
+- <%= action.inspect %>
--- a/src/api/app/views/webui/feeds/notifications.rss.builder
+++ b/src/api/app/views/webui/feeds/notifications.rss.builder
@@ -9,14 +9,24 @@ xml.rss version: '2.0' do
 
     @notifications.each do |notification|
       xml.item do
-        xml.title notification.event.subject
-        # We render the event_mailer view to 'description'
-        xml.description render(
-          template: "event_mailer/#{notification.event.template_name}",
-          layout: false,
-          formats: :text,
-          locals: { event: notification.event.expanded_payload }
-        )
+        if notification.notifiable
+          xml.title notification.title
+          xml.description render(
+            template: "notifications/#{notification.template_name}",
+            layout: false,
+            formats: :text,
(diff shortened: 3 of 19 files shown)


eduardoj:

src/api/db/data/20200319095810_remove_old_rss_notifications.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveOldRssNotifications < ActiveRecord::Migration[5.2]
+  def up
+    Notification.where(notifiable_id: nil).or(Notification.where(notifiable_type: nil)).delete_all
```
hennevogel: Hm, this is time sensitiive right? If I run this in the next release I will not have any `Notifications` anymore

src/api/app/views/notifications/_group_action.text.erb
```diff
@@ -0,0 +1 @@
+- <%= action.inspect %>
```
danidoni: Is this `.inspect` intended?

src/api/app/views/webui/feeds/notifications.rss.builder
```diff
@@ -9,13 +9,12 @@ xml.rss version: '2.0' do
 
     @notifications.each do |notification|
       xml.item do
-        xml.title notification.event.subject
-        # We render the event_mailer view to 'description'
+        xml.title notification.title
         xml.description render(
-          template: "event_mailer/#{notification.event.template_name}",
+          template: "notifications/#{notification.template_name}",
```
hennevogel: Hm, do you ever expect those to be different from the mailer content? Why duplicate all of this?

src/api/app/views/notifications/_group_action.text.erb
eduardoj: Yes. These notification templates were copied from `app/views/event_mailer/` to `app/views/notifications/`, and adapted to use the new polymorphic notification columns. In [this case](https://github.com/openSUSE/open-build-service/blob/d61545cf84f10b93b299a5d4e9e8992164789963/src/api/app/views/event_mailer/_group_action.text.erb#L1) it was not needed to adapt anything, and only a copy was made.

danidoni: okay

src/api/db/data/20200319095810_remove_old_rss_notifications.rb
eduardoj: Upgrading to the next release will delete all notification records. But new notification records will be created thanks to the code changes made in pull requests, previously merged (#9243 and friends). So yes, you will not have any  _old_ `Notifications`, but you will receive notifications from events that happen since  the upgrade, and later on.

hennevogel: I don't think you should do that. You should handle the `Notification`s that are there and that are relevant.

src/api/app/views/webui/feeds/notifications.rss.builder
eduardoj: They will be not so different. At least by now the only difference is that the diff of a submit request is not show in the body of a RSS feed item.

Another reason to duplicate templates was not to touch mailer templates. Using the same templates for both mails and RSS feed items would pretty probably mean to change variables in the templates, and create some logic to fill this variables with different objects, depending on rendering mails for an event, or rendering a RSS feed item from a notification.

What do you think?

src/api/db/data/20200319095810_remove_old_rss_notifications.rb
eduardoj: What do you mean by _handle_ and _relevant_?

hennevogel: * handle == set the notifiable
* relevant == newer than 100 days

src/api/app/views/webui/feeds/notifications.rss.builder
hennevogel: Hm maybe I don't understand something. The RSS feeds now use the mailer templates right? Why change this now?

src/api/db/data/20200319095810_remove_old_rss_notifications.rb
eduardoj: This would mean to create a migration of the notifications newer than 100 days, for users that have logged in the last 100 days and that have their RSS feed enabled. And they would be 10 notification registers maximum per user. Keeping in mind that emails woin't be affected, and only users that use RSS feeds, worst case scenario is that the users won't receive last 10 notifications through their RSS feeds. I don't think the work invested in doing this migration is worth it. Is this really needed?

src/api/app/views/webui/feeds/notifications.rss.builder
eduardoj: > Hm maybe I don't understand something. The RSS feeds now use the mailer templates right?

Yes.

> Why change this now?

This has been changed to use other templates to not touch the mailer templates, and to substitute the variables inside the templates with the values retrieved from the new polymorphic columns from notifications.

src/api/db/data/20200319095810_remove_old_rss_notifications.rb
hennevogel: >  worst case scenario is that the users won't receive last 10 notifications through their RSS feeds

Often people use RSS feeds for automation, yes I think this would not be nice. Doesn't have to be in this PR though...

src/api/app/views/webui/feeds/notifications.rss.builder
hennevogel: Ah, you are getting the data from other places. I understand. Okay

src/api/db/data/20200319095810_remove_old_rss_notifications.rb
eduardoj: Ok, I will modify this pull request, and I will substitute the deletion of data with a migration. Marking this pull request as WIP till this is done.

src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -32,22 +32,27 @@ def perform
...
-      notification_params = {
-        subscriber: subscription.subscriber,
-        event_type: event.eventtype,
-        event_payload: event.payload,
-        subscription_receiver_role: subscription.receiver_role
-      }
-
-      next if subscription.subscriber && subscription.subscriber.away?
-      Notification::RssFeedItem.create(notification_params.merge!(notification_dynamic_params(event)))
+      create_rss_notification(event, subscription)
```
hennevogel: Can you move this glue code into the outer loop and out of the rescue? Wrapping your head around singular/pluralized method names is error prone. Like

```ruby
Event::Base.where(mails_sent: false).order(created_at: :asc).limit(1000).each do |event|
  ...
  if event.eventtype.in?(EVENTS_TO_NOTIFY)
    event.subscriptions.each do |subscription|
       create_rss_notification(event, subscription)
    end
  end
end
```

src/api/app/views/webui/feeds/notifications.rss.builder
```diff
@@ -9,14 +9,24 @@ xml.rss version: '2.0' do
...
+        if notification.notifiable
+          xml.title notification.title
+          xml.description render(
+            template: "notifications/#{notification.template_name}",
+            layout: false,
+            formats: :text,
+            locals: { notification: notification }
+          )
+        else
+          # TODO: This 'else' clause will be removed after all records from notifications are migrated
```
hennevogel: Please mention `responsive_ux` somehow so we don't forget this.

eduardoj: Done.

src/api/app/jobs/send_event_emails_job.rb
eduardoj: Code climate complained when the code was moved outside, without using a method. We changed the method name to `create_notifications` and refactored some more code. Please have a look again.

src/api/db/data/20200319095810_remove_old_rss_notifications.rb
eduardoj: We discussed offline and decided not to run a migration in this pull request. So for rss feeds, data will be read from the old `event_payload` field only when it is necessary.

src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -10,44 +10,48 @@ class SendEventEmailsJob < ApplicationJob
 
   def perform
     Event::Base.where(mails_sent: false).order(created_at: :asc).limit(1000).each do |event|
+      event.update_attributes(mails_sent: true)
```
hennevogel: If something goes wrong below, you can never get back the event. Please do this as last action.

eduardoj: Changed back to the `ensure` inside the `send_email` method.

PR #9250, opened 2020-03-19
[backend] bs_dodup: support xz compressed Packages files

mlschroe:


adrianschroeter:
should also go to 2.10 ...

PR #9251, opened 2020-03-19
bs_repserver: allow more than to components in appliance build numbers

lnussel:
WSL applicance for TW currently has three components as it also embeds
the TW version number.


adrianschroeter:
k, but please use the [backend] commit prefix

Vogtinator:
Did a quick test, appears to work for WSL, JeOS and LXC.

PR #9253, opened 2020-03-19
[backend] document commitfilelist command

adrianschroeter:

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -832,7 +832,6 @@ POST /source/<project>/<package>
   linkdiff: for server side diff of a linked or branched package
   servicediff: shows the changes of the service run
   commit: commit files in upload revision
-  commitfilelist: commit defined files in upload revision
   deleteuploadrev: delete all uploaded sources which are not committed yet
   copy: copy package sources from another package
   undelete: undelete the package
@@ -880,9 +879,22 @@ Parameters:
   onlyissues: used to limit to issues (for diff commands)
   setrelease: define a specific release tag when used with "release" command. Setting it to "-" strips
               the release string.
-  withvalidate: activate sha validation code
   withvrev: copy also the vrev counter of the revision
 
+POST /source/<project>/<package>?cmd=commitfilelist
+
+  creating a new commit including defined files already uploaded to the repository.
+  This can be done via PUT using rev=repository which avoids a commit.
+
+  This requires a directory xml providing file name and md5 as body.
+  It is recommended to use the withvalidate parameter and including the 
+  sha256 sum including "sha256:" prefix in "hash" attribute for security reason.
+
+Parameters:
+  withvalidate: activate sha validation code (optional, but recommended)
+  comment: comment for history, optional
+
+XmlBody: directory
 
 
 GET /source/<project>/<package>/<binary>/_attribute/<attribute>
--- a/docs/api/api/directory.xsd
+++ b/docs/api/api/directory.xsd
@@ -40,6 +40,14 @@
       <xs:attribute name="originproject" type="xs:string"/>
       <xs:attribute name="available" type="xs:boolean"/>
       <xs:attribute name="recommended" type="xs:boolean"/>
+      <xs:attribute name="hash" type="xs:string">
+        <xs:annotation>
+          <xs:documentation>
+          a checksum for the save verification of the file. The checksum method
+          needs to be prefixed, eg sha256:......
+          </xs:documentation>
+        </xs:annotation>
+      </xs:attribute>
     </xs:complexType>
   </xs:element>
 


marcus-h:
On 2020-03-19 07:28:08 -0700, Adrian Schröter wrote:
> diff --git a/docs/api/api/api.txt b/docs/api/api/api.txt
> index 5aa83f8297f..7f1d054adbc 100644
> --- a/docs/api/api/api.txt
> +++ b/docs/api/api/api.txt

> @@ -880,9 +879,20 @@ Parameters:
>    onlyissues: used to limit to issues (for diff commands)
>    setrelease: define a specific release tag when used with "release" command. Setting it to "-" strips
>                the release string.
> -  withvalidate: activate sha validation code
>    withvrev: copy also the vrev counter of the revision
>
> +POST /source/<project>/<package>?cmd=commitfilelist
> +
> +  commit defined files in upload revision

Hmm the "upload revision" part is be a bit "confusing" (the files
have to be part of the repo not part of the upload rev).
Maybe

commit defined files if they are present in the repository

?

docs/api/api/api.txt
```diff
@@ -880,9 +879,20 @@ Parameters:
...
   setrelease: define a specific release tag when used with "release" command. Setting it to "-" strips
               the release string.
-  withvalidate: activate sha validation code
   withvrev: copy also the vrev counter of the revision
 
+POST /source/<project>/<package>?cmd=commitfilelist
+
+  commit defined files in upload revision
+
+  Requires a directory xml providing file name and md5 as body.
```
dcermak:
```suggestion
  Requires a directory xml providing file names and md5 hashes as the body.
```

docs/api/api/api.txt
```diff
@@ -880,9 +879,20 @@ Parameters:
...
+  It is recommended to use withvalidate parameter and send also additional the
```
dcermak:
```suggestion
  It is recommended to use the withvalidate parameter and to additionally send the
```

docs/api/api/api.txt
```diff
@@ -880,9 +879,20 @@ Parameters:
...
-  withvalidate: activate sha validation code
   withvrev: copy also the vrev counter of the revision
 
+POST /source/<project>/<package>?cmd=commitfilelist
+
+  commit defined files in upload revision
+
+  Requires a directory xml providing file name and md5 as body.
+  It is recommended to use withvalidate parameter and send also additional the 
+  sha256 in "hash" attribute for security reason.
```
dcermak:
```suggestion
  sha256 hash in the "hash" attribute of each directory entry for security reasons.
```

docs/api/api/directory.xsd
```diff
@@ -36,6 +36,7 @@
       <xs:attribute name="name" type="xs:string"/>
       <xs:attribute name="size" type="xs:string"/>
       <xs:attribute name="md5" type="xs:string"/>
+      <xs:attribute name="hash" type="xs:string"/>
```
dcermak: This is probably out of scope here, but I'd suggest to call this parameter `sha256hash` for clarity.

docs/api/api/api.txt
```diff
@@ -880,9 +879,20 @@ Parameters:
   onlyissues: used to limit to issues (for diff commands)
   setrelease: define a specific release tag when used with "release" command. Setting it to "-" strips
               the release string.
-  withvalidate: activate sha validation code
   withvrev: copy also the vrev counter of the revision
 
+POST /source/<project>/<package>?cmd=commitfilelist
+
+  commit defined files in upload revision
```
dcermak: This sounds like this API call would result in a commit being created, but I thought that wasn't the case here?

adrianschroeter: it creates a commit, yes

docs/api/api/directory.xsd
adrianschroeter: yeah, unfortunatly this would break the api ... we could just introduce this in addition.

adrianschroeter:
New version pushed, I hope this makes it better understandable now?

marcus-h:
On 2020-03-19 08:11:41 -0700, Adrian Schröter wrote:
> New version pushed, I hope this makes it better understandable now?
>
Just a minor remark:

> +  This can be done via PUT them using rev=upload.

Hmm "rev=repository" is probably "better" for this use case.

docs/api/api/directory.xsd
mlschroe: No, the value is prefixed with the hash name, e.g. `sha256:abcdef...`. That's why it's just named `hash`.

dcermak: Oh! That definitely belongs into the the documentation of that attribute then, because I would've just dumped the sha256 hash there.

docs/api/api/api.txt
dcermak: Ah, in that case we should also mention the parameter `comment` here.

docs/api/api/directory.xsd
adrianschroeter: right, added

docs/api/api/api.txt
adrianschroeter: right, added

adrianschroeter:
> On 2020-03-19 08:11:41 -0700, Adrian Schröter wrote: New version pushed, I hope this makes it better understandable now?
> Just a minor remark:
> + This can be done via PUT them using rev=upload.
> Hmm "rev=repository" is probably "better" for this use case.

right, changed

adrianschroeter:
ignoring css webui rspec failure, seems to be invalid

PR #9256, opened 2020-03-19
[backend] Make HTTPS communication between services possible

mlschroe:
When workers and services are located in different networks, we don’t
want to send traffic unencrypted and open to anyone to listen to.

Make it possible to easily opt in to HTTPS communication by changing
the service URLs in BSConfig.

This is based on a pull request created by Andrej Shadura
<andrew.shadura@collabora.co.uk>. Thanks a lot!


andrewshadura:
I think I found lots of other places in the code where this breaks, but ultimately I ran into the fact workers listen on HTTP so they need a reverse proxy of some sort anyway.
When I find time I will submit an additional patch with other bits if they still apply (mine is based on an earlier version of the backend).

mlschroe:
Ok, thanks. I merged this pull requests now so you can use it as base for your next pull request.

PR #9248, opened 2020-03-19
Create User Notification View

krauselukas:
### Desktop
Peek 2020-03-19 17-01

### Mobile

Peek 2020-03-19 17-02

PR #9263, opened 2020-03-20
Add notification factories

danidoni:
We need several factories to support the different use cases for the
notifications controller specs

### DIFF
--- a/src/api/spec/factories/notification.rb
+++ b/src/api/spec/factories/notification.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :notification do
-    type { 'Notification' }
+    type { 'Notification::RssFeedItem' }
     event_type { 'FakeEventType' }
     event_payload { { fake: 'payload' } }
     subscription_receiver_role { 'owner' }
@@ -13,12 +13,12 @@ FactoryBot.define do
       bs_request_oldstate { :new }
     end
 
-    trait :creation do
+    trait :request_created do
       event_type { 'Event::RequestCreated' }
       association :notifiable, factory: :bs_request_with_submit_action
     end
 
-    trait :review do
+    trait :review_wanted do
       event_type { 'Event::ReviewWanted' }
       association :notifiable, factory: :bs_request_with_submit_action
     end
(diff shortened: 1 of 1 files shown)


src/api/spec/factories/notification.rb
```diff
@@ -1,8 +1,42 @@
...
+    title { Faker::Lorem.sentence }
+    delivered { false }
+
+    trait :state_change do
+      event_type { 'Event::StateChange' }
+      association :notifiable, factory: :bs_request_with_submit_action
+      bs_request_oldstate { :new }
+    end
+
+    trait :creation do
```
hennevogel:
```suggestion
    trait :request_created do
```

src/api/spec/factories/notification.rb
```diff
@@ -1,8 +1,42 @@
...
+      event_type { 'Event::RequestCreated' }
+      association :notifiable, factory: :bs_request_with_submit_action
+    end
+
+    trait :review do
```
hennevogel:
```suggestion
    trait :review_wanted do
```

src/api/spec/factories/notification.rb
```diff
@@ -1,8 +1,42 @@
 FactoryBot.define do
   factory :notification do
+    type { 'Notification' }
     event_type { 'FakeEventType' }
```
hennevogel: This should be one of the real event types I think.

src/api/spec/factories/notification.rb
danidoni: ok

src/api/spec/factories/notification.rb
danidoni: ok

src/api/spec/factories/notification.rb
danidoni: Let me check

hennevogel: That's not an event type. That's a `Notification` type.

danidoni: Ah you were talking about the `event_type` field... I thought you were talking about the `type` field :man_facepalming:

danidoni: I've set a real event type.

danidoni:
I'm closing this one and just push the factory into https://github.com/openSUSE/open-build-service/pull/9248. That way I can add also the specs for the notification controller there.

PR #9270, opened 2020-03-23
Simplify the test package to be build on openQA to avoid flickering

vpereira:
Use a simpler package on the openQA job. The idea of those tests
are to test the integration between backend and frontend and not the
unix toolchain.

Often tests were failing because of the package
building takes too long


dist/t/spec/fixtures/hello_world.spec
```diff
@@ -0,0 +1,27 @@
+Name:       hello_world
+Version:    1
+Release:    1
+Summary:    Most simple RPM package
+License:    FIXME
```
coolo: make this CC0-1.0 :)

vpereira: ;)

Issue #9285, opened 2020-03-25
Meta data gets not adapted after deleting user


adrianschroeter:
A deleted user is still visible in webui and meta data after deleting him. 

A real life example is "mrdocs", still listed eg in "graphics" project, but not via /person/mrdocs anymore.

In other cases this might be a DSGVO problem (IMHO also a regression, but not 100%)

hennevogel:
This is a more a feature (as in missing functionality) than a bug. Let me see what we do already and what we can do.

hennevogel:
- [ ] remove all `Relationship`
- [ ] remove all attrib_namespace_modifiable_bies
- [ ] remove all attrib_type_modifiable_bies
- [ ] remove all comments
- [ ] remove all commit_activities
- [ ] remove all group_maintainers
- [ ] remove all groups_users
- [ ] remove all history_elements
- [ ] remove all messages
- [ ] remove all reviews
- [ ] remove all status_messages
- [X] delete the home project
- [X] blank `User.mail`
- [X] blanke `User.realname`

adrianschroeter:
On Mittwoch, 25. März 2020, 17:24:01 CET wrote Henne Vogelsang:
> - [ ] remove all `Relationship`
> - [ ] remove all attrib_namespace_modifiable_bies
> - [ ] remove all attrib_type_modifiable_bies

not needed.

> - [ ] remove all comments

that would be good, yes

> - [ ] remove all commit_activities
> - [ ] remove all group_maintainers
> - [ ] remove all groups_users

this is no needed

> - [ ] remove all history_elements

that must not happen.

> - [ ] remove all messages

unsure if needed

> - [ ] remove all reviews

must not happen.

> - [ ] remove all status_messages
> - [X] delete the home project
> - [X] blank `User.mail`
> - [X] blanke `User.realname`
> 
> 
> 
> 


-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

hennevogel:
Maybe you can explain **why** something is not needed in your opinion? Because for instance `Relationship` is the use case you are asking for. `Relationship` is the connection between `User` and `Package` or `Project`.

dmarcoux:
Any progress on this because it's affecting some requests like:
[...]

PR #9288, opened 2020-03-25
[frontend] avoid visibility of deleted users

adrianschroeter:
initial unfinished input for issue 9285

Could not test it due to some webmock crash, but turns the behaviour from opt-out to opt-in of deleted users.

use default scrope to not include them

FIXME: re-creating of existing user most likely fail in a bad way not.
       but test suite just dies with webmock, so I can not test it


src/api/app/models/user.rb
```diff
@@ -650,6 +652,9 @@ def delete!
     # or another user could take over the identity which can open security
     # issues (other infrastructur and systems using repositories)
 
+    # keep object copy for cleanup
+    myself = self.dup
```
hennevogel: This makes little sense. We never delete `self`

src/api/app/models/user.rb
```diff
@@ -661,6 +666,10 @@ def delete!
       project.destroy
     end
 
+    # wipe affected caches
+    myself.involved_projects.each do |prj| prj.reset_cache; end
+    myself.involved_packages.each do |pkg| pkg.reset_cache; end
```
hennevogel: This does nothing to resolve your problem. There will still be a `Relationship` between User and Project...

hennevogel:
Can we continue the discussion in #9285 instead?

adrianschroeter:
On Mittwoch, 25. März 2020, 17:14:42 CET wrote Henne Vogelsang:
> 
> 
> 
> > @@ -650,6 +652,9 @@ def delete!
>      # or another user could take over the identity which can open security
>      # issues (other infrastructur and systems using repositories)
>  
> +    # keep object copy for cleanup
> +    myself = self.dup
> 
> This makes little sense. We never delete `self`

 not, but setting a different state, which can have effects.

> 
> > @@ -661,6 +666,10 @@ def delete!
>        project.destroy
>      end
>  
> +    # wipe affected caches
> +    myself.involved_projects.each do |prj| prj.reset_cache; end
> +    myself.involved_packages.each do |pkg| pkg.reset_cache; end
> 
> This does nothing to resolve your problem. There will still be a `Relationship` between User and Project...

sure, no problem, this won't be visible anymore.

It just enforces that all xml via api gets regenerated and the user
will be missing there.

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

hennevogel:
>>  This makes little sense. We never delete `self`

> not, but setting a different state, which can have effects.

Not here, not then. `self` is already loaded...

> It just enforces that all xml via api gets regenerated and the user will be missing there.

No it wont. It will reset the cache and the next call to `Project.to_axml` will render the still existing `Relationship` into the cache again.

adrianschroeter:
On Mittwoch, 25. März 2020, 18:57:56 CET wrote Henne Vogelsang:
> >>  This makes little sense. We never delete `self`
> 
> > not, but setting a different state, which can have effects.
> 
> Not here, not then. `self` is already loaded...
> 
> > It just enforces that all xml via api gets regenerated and the user will be missing there.
> 
> No it wont. It will reset the cache and the next call to `Project.to_axml` will render the still existing `Relationship` into the cache again.

nope, because the users with deleted states are filtered due to
the default_scope. At least they should, could test yet as I said.

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

hennevogel:
Several features break because of this change. See the failing tests...

PR #9289, opened 2020-03-25
Add strong_migrations to our development environment

vpereira:
Install https://github.com/ankane/strong_migrations and white list all
migrations before `20200318123203`

### DIFF
--- a/src/api/config/initializers/strong_migrations.rb
+++ b/src/api/config/initializers/strong_migrations.rb
@@ -1,0 +1,4 @@
+# since its a timestamp and not a regular number, disable the cop
+# rubocop:disable Style/NumericLiterals
+StrongMigrations.start_after = 20200318123203
+# rubocop:enable Style/NumericLiterals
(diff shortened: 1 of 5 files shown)


src/api/config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,2 @@
+# adding the number spaced by underscores to make rubocop happy
+StrongMigrations.start_after = 20_200_318_123_203
```
vpereira: I would prefer to just disable the cop here. It isn't a "large" number but a timestamp

vpereira:
Looking the spec https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/db/data/remove_duplicate_repositories_spec.rb#L15, talking with @coolo he explained us that https://github.com/openSUSE/open-build-service/commit/b288c877d14ac7183064967db932371883652bd2 is part of 2.10 - so if we don't support migrations from 2.9 to 2.11, you can just remove the migrations and replace it with a new initial migration

PR #9299, opened 2020-03-27
[backend] include count of cumulative build failures in notifications

sjamgade:
this count will help track long failing packages

### DIFF
--- a/src/api/app/models/event/build.rb
+++ b/src/api/app/models/event/build.rb
@@ -3,7 +3,7 @@ module Event
     self.description = 'Package has finished building'
     self.abstract_class = true
     payload_keys :project, :package, :sender, :repository, :arch, :release, :readytime, :srcmd5,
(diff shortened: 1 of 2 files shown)


src/api/app/models/event/build.rb
```diff
@@ -3,7 +3,7 @@ class Build < Base
     self.description = 'Package has finished building'
     self.abstract_class = true
     payload_keys :project, :package, :sender, :repository, :arch, :release, :readytime, :srcmd5,
-                 :rev, :reason, :bcnt, :verifymd5, :hostarch, :starttime, :endtime, :workerid, :versrel, :previouslyfailed
+                 :rev, :reason, :bcnt, :verifymd5, :hostarch, :starttime, :endtime, :workerid, :versrel, :previouslyfailed, :cumulative_failcount
```
hennevogel: **nitpick**: I don't like the name, count is always cumulative right? :smile:

sjamgade: correct.
Should be resolved now :-)

PR #9301, opened 2020-03-27
Don't send emails to groups without an email

### LINKED ISSUE #9114: [production][] SMTP To address may not be blank: []
threw this 116 times
<pre>
SMTP To address may not be blank: []
</pre>

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5e1e477f34c7c50008957a59)

eduardoj:
The corner case of sending a mail to only a group, that hasn't an email defined, was not taken into account.

This patch makes the code not to send an email in this case.

Fixes #9114.

### DIFF
--- a/src/api/app/mailers/event_mailer.rb
+++ b/src/api/app/mailers/event_mailer.rb
@@ -21,6 +21,9 @@ class EventMailer < ActionMailer::Base
     subscribers = subscribers.to_a
     return if subscribers.empty?
 
+    recipients = subscribers.map(&:display_name).reject(&:blank?)
+    return if recipients.empty?
+
     set_headers
     begin
       locals = { event: e.expanded_payload }
@@ -33,7 +36,6 @@ class EventMailer < ActionMailer::Base
 
     template_name = e.template_name
     orig = e.originator
-    tos = subscribers.map(&:display_name)
 
     if orig
       orig = orig.display_name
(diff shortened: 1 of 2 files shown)


src/api/app/mailers/event_mailer.rb
```diff
@@ -21,6 +21,9 @@ def event(subscribers, e)
     subscribers = subscribers.to_a
     return if subscribers.empty?
 
+    tos = subscribers.map(&:display_name).reject(&:blank?)
```
vpereira: Shouldn't we test this condition on `src/api/spec/mailers/event_mailer_spec.rb`

dmarcoux: What is `tos`?

danidoni: I'm working on specs for this change

danidoni: `tos` is the plural of `to` addresses :stuck_out_tongue:

dmarcoux: Ishhhh, could we have a better name for this?

danidoni: Yeah, I'll change it

vpereira:
@danidoni please squash those commits together

danidoni:
> I had to look at the code changes to understand what this does since the commit message is not very clear. Could we reword the commit message to something like `Don't send emails to groups without an email`?

Sure

PR #9306, opened 2020-03-30
[backend] fix setrelease for single medium products

adrianschroeter:
these do not have a medium number behind -Media


adrianschroeter:
needs a rerun

PR #9322, opened 2020-04-02
[backend] solve double match special case for setrelease

adrianschroeter:
openSUSE Leap 15.2 product releases use "-BuildX.Y" as setrelease.
This leads to double match which breaks the name and adds unwanted X.Y.Z
result.

### DIFF
--- a/src/backend/bs_repserver
+++ b/src/backend/bs_repserver
@@ -2524,8 +2524,12 @@ sub copybuild {
       if (defined($setrelease)) {
 	$setrelease =~ s/^-?/-/; # "-" will drop the release tag
 	$setrelease =~ s/-?$//;  # drop leading "-", it depends on the format
-	$nbin =~ s/-([^-]+)(-Media\d?(?:\..*?)?)$/$setrelease$2/; # kiwi product builds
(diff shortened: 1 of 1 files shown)


src/backend/bs_repserver
```diff
@@ -2524,8 +2524,12 @@ sub copybuild {
       if (defined($setrelease)) {
 	$setrelease =~ s/^-?/-/; # "-" will drop the release tag
 	$setrelease =~ s/-?$//;  # drop leading "-", it depends on the format
-	$nbin =~ s/-([^-]+)(-Media\d?(?:\..*?)?)$/$setrelease$2/; # kiwi product builds
-	$nbin =~ s/-Build[\d.]+\./$setrelease./; # kiwi appliance builds
+        # we must ensure to run only one of the following, because people
+        # overwrite -BuildX.Y part with -BuildV.W (using invalid numbers)
```
Vogtinator, line 2528: Explanation is wrong - using any numbers whatsoever for the build is fine - as long as it is not `BuildX.Y-Media`, which matches the product because of the `-Media` and the appliance because of `BuildX.`, but the `Y-` is missing because of the `-`

Vogtinator:
Testing seems to be successful so far.

PR #9327, opened 2020-04-02
Remove unneeded JS dependencies

dmarcoux:
While working on the [Webpacker card](https://trello.com/c/Gg2gJcwH/1346-try-webpacker-again), I've found out that we have a few JS dependencies which we don't use anymore. Let's remove them.


dmarcoux:

PR #9328, opened 2020-04-02
[backend] Add a config option to enable the X-Registry-Supports-Signa…

mlschroe:
…tures extension

The extension needs an update to the registry's apache config, so
always enabling it is dangerous.

### DIFF
--- a/src/backend/BSConfig.pm.template
+++ b/src/backend/BSConfig.pm.template
@@ -284,6 +284,10 @@ our $publish_containers = [
    '.*'             => ['local'],
 ];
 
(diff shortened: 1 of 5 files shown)


src/backend/BSConfig.pm.template
```diff
@@ -284,6 +284,8 @@ our $publish_containers = [
    '.*'             => ['local'],
 ];
 
+our $local_registry_signatures_extension = 1;
```
adrianschroeter, line 287: A comment here would be nice ... esp. the for hint for the apache config update for updaters...

mlschroe:
Just for you...

PR #9330, opened 2020-04-02
[backend] fix container variant handling

mlschroe:
The variant is not in the image config, but only in the containerinfo
data.


mlschroe:
Sigh. Sigh. Sigh.

PR #9344, opened 2020-04-07
[backend] fix dying workers in cleanup_chroot mode

adrianschroeter:
it is racy by design, since a job can get aborted at any time in parallel.

The method name (rm_rf) is a bit questionable, since "rm -rf" would never return an error.

### DIFF
--- a/src/backend/bs_worker
+++ b/src/backend/bs_worker
@@ -205,8 +205,10 @@ sub cleanup_job {
   # if this is a chroot like case, get rid of it after the build
   if ((!$vm || $vm eq 'lxc' || $vm eq 'docker' || $vm eq 'nspwan') && $cleanup_chroot && -d $buildroot) {
       printf "Removing buildroot $buildroot";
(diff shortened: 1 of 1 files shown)


dmarcoux:
There's a typo in the commit message. It should be `dying` instead of `dieing`.

mlschroe:
I don't understand this commit, a delete in parallel will not leave the dir in place.

adrianschroeter:
On Dienstag, 7. April 2020, 13:28:19 CEST wrote Michael Schroeder:
> I don't understand this commit, a delete in parallel will not leave the dir in place.

no, but if that removes the directory it will die. And if this is the master worker
the entire worker is dead.

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

adrianschroeter:
because the other process might have re-created and put stuff inside already

PR #9345, opened 2020-04-07
[dist] fix db:setup in setup-appliance.sh

M0ses:
fixes problem with rake command db:setup:

```
** Invoke environment
** Execute strong_migrations:safety_assured
rails aborted!
Set SAFETY_ASSURED=1 to run this task in production
```


M0ses:
OBS built successfully - will merge

hennevogel:
Hm, you successfully avoided the failure but why did that happen?

PR #9346, opened 2020-04-07
[backend]pass --ccache param if usecache:packid is present as buildflag

sjamgade:
this buildflag is to be manually added in project config as

BuildFlags: useccache:package-name

along with this ccache has be included as BuildRequires in the spec file
for ccache to be installable.

If buildflag is added but ccache is not present it will result in error

### DIFF
--- a/src/backend/bs_worker
+++ b/src/backend/bs_worker
@@ -3097,6 +3097,7 @@ sub dobuild {
   my $config = BSRPC::rpc("$server/getconfig", undef, "project=$projid", "repository=$repoid", @configpath);
   writestr("$buildroot/.build.config", undef, $config);
 
+  my $bconf;
   if ($deltamode) {
     # replace @mopt@ in specfile
     my $spec = readstr("$srcdir/deltagen.spec");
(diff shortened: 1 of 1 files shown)


src/backend/bs_worker
```diff
@@ -3116,6 +3117,15 @@ sub dobuild {
       writestr("$srcdir/deltagen.spec", undef, $spec);
     }
   }
+  $bconf = $bconf || Build::read_config($buildinfo->{'arch'}, "$buildroot/.build.config");
+
+  my $useccache;
+  if (exists $bconf->{'buildflags:useccache'}) {
+      if (grep /^useccache:$packid$/s, @{$bconf->{'buildflags'} || []}) {
```
mlschroe, line 3124: Please use `/^useccache:\Q$packid\E$/` so that it also works if the packid contains a `+` sign

src/backend/bs_worker
```diff
@@ -3219,6 +3229,7 @@ sub dobuild {
   push @args, '--debug' if $buildinfo->{'debuginfo'};
   push @args, '--arch', $arch;
   push @args, '--jobs', $jobs if $jobs;
+  push @args, '--ccache' if $useccache;
```
mlschroe, line 3232: Ok, I'm confused now. Why is this --ccache and not --pkgccache?

mlschroe:
Why can't ccache be automatically added when the buildjob is created? You just need to add a couple of lines to BSSched/BuildJob.pm

PR #9361, opened 2020-04-09
[backend] filter out ccache archive while listing package binaries

sjamgade:

### DIFF
--- a/src/backend/bs_repserver
+++ b/src/backend/bs_repserver
@@ -1010,6 +1010,7 @@ sub getbinarylist {
     @bins = grep {!/^::import::/} @bins if $cgi->{'noimport'};
     @bins = BSRepServer::Containertar::add_containers(@bins) if grep {/\.containerinfo$/} @bins;
     @bins = filtersources(@bins) if $cgi->{'nosource'} || -e "$reporoot/$prp/$arch/$packid/.nosourceaccess";
+    @bins = grep {$_ ne "_ccache.tar.gz"} @bins unless $cgi->{'withccache'};
     for (sort @bins) {
       next if %binaries && !$binaries{$_};
       if (/\.tar$/ && ! -e "$reporoot/$prp/$arch/$packid/$_") {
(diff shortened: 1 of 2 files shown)


src/backend/bs_repserver
```diff
@@ -1010,6 +1010,7 @@ sub getbinarylist {
     @bins = grep {!/^::import::/} @bins if $cgi->{'noimport'};
     @bins = BSRepServer::Containertar::add_containers(@bins) if grep {/\.containerinfo$/} @bins;
     @bins = filtersources(@bins) if $cgi->{'nosource'} || -e "$reporoot/$prp/$arch/$packid/.nosourceaccess";
+    @bins = grep {!/^_ccache.tar.gz$/} @bins unless $cgi->{'withccache'};
```
mlschroe, line 1013: Hmm, the `.` in the regexp seem to need escaping.

Come to think of it: why do you use a regexp at all in this case? It could be simply:
```
@bins = grep {$_ ne '_ccache.tar.gz'} @bins unless $cgi->{'withccache'};
```

src/backend/bs_worker
```diff
@@ -3075,7 +3075,24 @@ sub dobuild {
...
+  }
+  my $server = $buildinfo->{'srcserver'} || $srcserver;
+  my $config = BSRPC::rpc("$server/getconfig", undef, "project=$projid", "repository=$repoid", @configpath);
+  writestr("$buildroot/.build.config", undef, $config);
+
+
+  my $bconf = Build::read_config($buildinfo->{'arch'}, "$buildroot/.build.config");
+  my $useccache=0;
+  if (exists $bconf->{'buildflags:useccache'}) {
+      if (grep /^useccache:\Q$packid\E$/, @{$bconf->{'buildflags'} || []}) {
```
mlschroe, line 3091: Having said the above for _ccache.tar.gz, this could also be
```
if (grep {$_ eq "useccache:$packid"} @{$bconf->{'buildflags'} || []}) {
```
Just saying...

Oh, and should we rename the flag to `pkgccache` to match the build option? Just a thought...

PR #9365, opened 2020-04-09
Display the number of unread notifications in the top-level navigation

dmarcoux:
We don't rely on a cache for this since it's only a simple count.

Preview on desktop and mobile:
desktop
mobile

In case you wonder, this would fit 999 999 notifications before it breaks. It's even more on desktop, so we should be alright!
mobile-edge-case

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -802,12 +802,17 @@ class User < ApplicationRecord
     to_s
   end
 
+  # TODO: Remove once responsive_ux is out of beta
   def tasks
     Rails.cache.fetch("requests_for_#{cache_key}") do
       declined_requests.count +
         incoming_requests.count +
         involved_reviews.count
     end
+  end
+
+  def unread_notifications
+    notifications.not_marked_as_done.size
   end
 
   def watched_project_names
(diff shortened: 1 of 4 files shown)


dmarcoux:
[S - As a user I want to see in the navigation bar if there is a new notification](https://trello.com/c/fOHHrLr2/1362-s-as-a-user-i-want-to-see-in-the-navigation-bar-if-there-is-a-new-notification)

src/api/app/models/user.rb
```diff
@@ -810,6 +811,10 @@ def tasks
     end
   end
 
+  def unread_notifications
+    notifications.not_marked_as_done.size
```
eduardoj, line 815: Couldn't it be better to do a `.count` instead of a `.size`?

dmarcoux: `.size` calls `.count` or `.length` depending on the context. See this from [Stack Overflow](https://stackoverflow.com/questions/6083219/activerecord-size-vs-count).

eduardoj: It seems that not, it is not better. Is the same: `.size` also throws a `SELECT COUNT(*) ...`.

dmarcoux: I want a `SELECT COUNT(*)` here and this is what `.size` does in this context. `.count` also does it.

PR #9376, opened 2020-04-14
Allow all hosts in ActionDispatch::HostAuthorization

hennevogel:
There is a new middleware in Rails 6 which guards from DNS
rebinding attacks. We are not too worried about this attack
in development.


hennevogel:
Let's see how far we come if we just set this in the one place where this is currently making problems...

PR #9378, opened 2020-04-14
WIP: Filter notifications by project

eduardoj:
Show a clickable list of projects to filter notifications by project.
Adapt models used by notifications to be able to filter by project.

### DIFF
--- a/src/api/app/controllers/webui/users/notifications_controller.rb
+++ b/src/api/app/controllers/webui/users/notifications_controller.rb
@@ -1,10 +1,20 @@
 class Webui::Users::NotificationsController < Webui::WebuiController
-  before_action :require_login
   MAX_PER_PAGE = 300
+  VALID_NOTIFICATION_TYPES = ['read', 'reviews', 'comments', 'requests', 'unread'].freeze
+
+  before_action :require_login
+  before_action :check_param_type, :check_param_project, only: :index
 
   def index
     notifications_for_subscribed_user = NotificationsFinder.new.for_subscribed_user
-    @notifications = NotificationsFinder.new(notifications_for_subscribed_user).for_notifiable_type(params[:type])
+
+    @projects = NotificationProjects.new(NotificationsFinder.new(notifications_for_subscribed_user).for_notifiable_type('unread')).call
+
+    @notifications = if params[:project]
+                       NotificationsFinder.new(notifications_for_subscribed_user).for_project_name(params[:project])
+                     else
+                       NotificationsFinder.new(notifications_for_subscribed_user).for_notifiable_type(params[:type])
+                     end
     @notifications = params['show_all'] ? show_all : @notifications.page(params[:page])
   end
 
@@ -21,6 +31,20 @@ class Webui::Users::NotificationsController < Webui::WebuiController
   end
 
   private
+
+  def check_param_type
+    return if params[:type].nil? || VALID_NOTIFICATION_TYPES.include?(params[:type])
+
+    flash[:error] = 'Filter not valid.'
+    redirect_to my_notifications_path
+  end
+
+  def check_param_project
+    return unless params[:project] == ''
+
+    flash[:error] = 'Filter not valid.'
+    redirect_to my_notifications_path
+  end
 
   def show_all
     total = @notifications.size
--- a/src/api/app/helpers/webui/webui_helper.rb
+++ b/src/api/app/helpers/webui/webui_helper.rb
@@ -402,10 +402,22 @@ module Webui::WebuiHelper
     end
   end
 
-  def user_notification_link(link_text, new_filter, filter: 'inbox')
+  def user_notification_link(link_text, filter_item)
     css_class = 'list-group-item list-group-item-action'
-    css_class += ' active' if new_filter == filter || (filter.nil? && new_filter == 'inbox')
-    link_to(link_text, my_notifications_path(type: new_filter), class: css_class)
+    css_class += ' active' if notification_filter_active?(filter_item)
+    link_to(link_text, my_notifications_path(type: filter_item[:type], project: filter_item[:project]), class: css_class)
+  end
+
+  private
+
+  def notification_filter_active?(filter_item)
+    if params[:project].present?
+      filter_item[:project].try(:name) == params[:project]
+    elsif params[:type].present?
+      filter_item[:type] == params[:type]
+    else
+      filter_item[:type] == 'unread'
+    end
   end
 end
 
--- a/src/api/app/helpers/webui/notification_helper.rb
+++ b/src/api/app/helpers/webui/notification_helper.rb
@@ -8,4 +8,11 @@ module Webui::NotificationHelper
       link_to('Show all', my_notifications_path(parameters), class: 'btn btn-sm btn-secondary ml-2')
     end
   end
+
+  def project_list(projects)
+    hash_projects = Hash.new(0)
+    projects.each { |project| hash_projects[project] += 1 }
+
+    hash_projects.sort_by(&:last).reverse
+  end
 end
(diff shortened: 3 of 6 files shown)


src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -402,10 +402,14 @@ def link_content(text, css_class, icon)
...
   end
 
-  def user_notification_link(link_text, new_filter, filter: 'inbox')
+  def user_notification_link(link_text, new_filter)
     css_class = 'list-group-item list-group-item-action'
-    css_class += ' active' if new_filter == filter || (filter.nil? && new_filter == 'inbox')
-    link_to(link_text, my_notifications_path(type: new_filter), class: css_class)
+    if (new_filter[:type].present? && params[:project].nil? && new_filter[:type] == params[:type]) ||
+       (new_filter[:project].present? && new_filter[:project] == params[:project]) ||
+       (params[:type].nil? && params[:project].nil? && new_filter[:type] == 'inbox')
```
coolo, line 409: any way to make this easier to parse for humans?

eduardoj: Done. Please have a look again.

src/api/app/models/review.rb
```diff
@@ -66,6 +66,8 @@ class NotFoundError < APIError
   before_validation :set_reviewable_association
   after_commit :update_cache
 
+  delegate :notifiable_projects, to: :bs_request
```
hennevogel, line 69: Why?

src/api/app/models/notification.rb
```diff
@@ -16,6 +16,21 @@ class Notification < ApplicationRecord
 
   scope :stale, -> { where('created_at < ?', 3.months.ago) }
 
+  def self.projects_with_count
```
hennevogel, line 19: `_with_count`? What are you trying to express?

src/api/app/models/bs_request.rb
```diff
@@ -1072,6 +1072,10 @@ def event_parameters
     notify_parameters
   end
 
+  def notifiable_projects
```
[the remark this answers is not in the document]
eduardoj: Creating `notifiable_projects` method in both `bs_request`, `comment` and `review` models permits to call `notifiable_projects` within a notifiable, allowing to retrieve the projects of a notifiable. [Here](https://github.com/openSUSE/open-build-service/pull/9378/files#diff-3fea9b194ee535ea7887130c45fc9988R32) for filtering, and [here](https://github.com/openSUSE/open-build-service/pull/9378/files#diff-8d970507aa562596ea2113061d254ed2R24) to retrieve the list of projects to filter for.

src/api/app/models/review.rb
eduardoj: Explained before: creating the `notifiable_projects` method in a review, together with the `notifiable_projects` in `comment` and `bs_request` allows us to retrieve the projects of a notifiable, independant of what notifiable_type it is.

src/api/app/models/notification.rb
eduardoj: `self.projects_with_count` will return a hash, where the keys are projects, and the values are the number of ocurrences each project happens. So `_with_count` refers to the number or ocurrences of each project.

eduardoj:

src/api/app/models/bs_request.rb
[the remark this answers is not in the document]
hennevogel: I see

src/api/app/models/notification.rb
hennevogel: So never mind what data structure this returns, what is this about? Notifications grouped by project?

eduardoj: Yes, kind of.

eduardoj: So, instead of `projects_with_count`, would it be better `projects_with_number_of_notifications`?

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -2,7 +2,43 @@ class Webui::Users::NotificationsController < Webui::WebuiController
...
+    redirect_back(fallback_location: root_path)
+  end
+
+  private
+
+  def notifications_by_project
+    @notifications = Notification.with_notifiable.unread.includes([:notifiable])
+                                 .for_subscribed_user(User.session)
+                                 .select do |notification|
+                                   notification if notification.notifiable.notifiable_projects.include?(params[:project])
```
danidoni, line 32: I think just using `notification.notifiable.notifiable_projects.include?(params[:project])` is enough, `select` will detect the `true` and choose the element from the collection

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -2,7 +2,43 @@ class Webui::Users::NotificationsController < Webui::WebuiController
...
+                                 end
+    return if @notifications.present?
+
+    flash[:error] = 'Filter not valid.'
```
danidoni, line 36: I would split the logic regarding fetching the `@notifications` from the logic regarding the `flash[:error]`

danidoni: `notifications_by_project` looks like a _getter_ method and making it side-effectful with the `flash` and the `redirect_to` is a bit misleading to me

dmarcoux:
We'll need to use `NotificationsFinder` which was introduced in #9399. Let me know if I can help.

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -19,21 +55,12 @@ def index
...
-    end
-  end
-
-  def update
-    notification = User.session.notifications.find(params[:id])
-    authorize notification, policy_class: NotificationPolicy
-
-    if notification.toggle(:delivered).save
-      flash[:success] = "Successfully marked the notification as #{notification.unread? ? 'unread' : 'read'}"
     else
```
danidoni, line 61: The same than before here, this method is getting data and the `flash` and the `redirect_to` are too much responsibilities. I would move that out

src/api/app/controllers/webui/users/notifications_controller.rb
eduardoj: True. Applied.

src/api/app/controllers/webui/users/notifications_controller.rb
eduardoj: Logic splitted. We created `before_action` methods. Please have a look again.

eduardoj:
> We'll need to use `NotificationsFinder` which was introduced in #9399. Let me know if I can help.

We modified the code to use  `NotificationsFinder`. Please have a look.

src/api/app/controllers/webui/users/notifications_controller.rb
eduardoj: This code was also moved.

src/api/app/controllers/webui/users/notifications_controller.rb
danidoni: Okay

src/api/app/models/notification.rb
hennevogel: How about `projects_with_notifications`?

hennevogel: And do the rest of the hash magic in a view helper?

eduardoj: 'projects_with_count' was renamed to `projects`, and the hash magic code was extracted to a view helper. Please review it again.

eduardoj:
Afer discussing it offline we agreed to try another approach. Marking as WIP.

eduardoj:
It will be tried another approach, including a creation of a relational table between notification and projects. Closing this PR.

PR #9379, opened 2020-04-14
SubscriptionsController update route

vpereira:
Add a custom name for the update route, to be able to use `/subscriptions/update` without id and keep the `my_subscriptions_path` pointing to the `SubscriptionsController#index`

How to test it:

Visit the Profile -> Change your notifications (update your notifications should work as expected)

Notifications -> Actions -> Manage Subscriptions (should bring you to your subscriptions index, and there you could update your notifications)

### DIFF
--- a/src/api/config/routes/webui_routes.rb
+++ b/src/api/config/routes/webui_routes.rb
@@ -292,7 +292,7 @@ OBSApi::Application.routes.draw do
 
       resources :subscriptions, only: [:index], controller: 'webui/users/subscriptions', as: :my_subscriptions do
         collection do
-          put 'update'
+          put 'update', as: :update
         end
       end
 
--- a/src/api/config/routes/webui_routes.rb
+++ b/src/api/config/routes/webui_routes.rb
@@ -293,7 +293,6 @@ OBSApi::Application.routes.draw do
       resources :subscriptions, only: [:index], controller: 'webui/users/subscriptions', as: :my_subscriptions do
         collection do
           put 'update'
-          get 'update'
         end
       end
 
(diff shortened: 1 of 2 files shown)


hennevogel:
> to be able to use /subscriptions/update without id

Hm why?

src/api/config/routes/webui_routes.rb
```diff
@@ -292,8 +292,7 @@
 
       resources :subscriptions, only: [:index], controller: 'webui/users/subscriptions', as: :my_subscriptions do
         collection do
-          put 'update'
-          get 'update'
+          put 'update', as: :update
         end
       end
```
hennevogel, line 297:
```suggestion
           resources :subscriptions, only: [], controller: 'webui/users/subscriptions' do
        collection do
          get '/', action: :index, as: :my
        end
        member do
          put 'update'
        end
      end
```

leads to
```
Prefix               Verb     URI Pattern                               Controller#Action
my_subscriptions     GET      /my/subscriptions(.:format)               webui/users/subscriptions#index
subscription         PUT      /my/subscriptions/:id/update(.:format)    webui/users/subscriptions#update
```

hennevogel: `EventSubscriptionForm::update` oh my. Why straightforward if you can also do it complicated... Nevermind me.

vpereira:
[S(M) - As a user I want to be able to reach the Notification configuration from the Notifications page](https://trello.com/c/I82a25aE/1359-sm-as-a-user-i-want-to-be-able-to-reach-the-notification-configuration-from-the-notifications-page)

PR #9388, opened 2020-04-15
Add ShellCheck scripts linter

eduardoj:
We want to benefit from having a linter that checks our script files (bash or sh).

Initially it is only run on only one file. The second commit of this pull request is an example of including a file into the list of files to be checked by the linter.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,9 @@ jobs:
       - <<: *frontend_base
     steps:
       - checkout
+      - run:
+          name: Run scripts linters
+          command: make -C dist scripts_linters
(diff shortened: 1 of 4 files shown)


eduardoj:
With this initial configuration even adding a new bash file in the repository it will not be parsed by the linter. It should be explicitly included in the `dist/run_shellcheck.sh` file.

Are you fine with this approach, @openSUSE/obs-backend?

dmarcoux:
It's been 2 months since we pinged @openSUSE/obs-backend and we didn't get a reply. I'll assume that nobody is against this. We can start to target more files in upcoming PRs.

PR #9389, opened 2020-04-15
Add pagination to notifications

saraycp:
Notifications are now paginated.

A `Show all` link is provided to display the notifications all at once. But there is a limit set by `MAX_PER_PAGE`. In case of exceeding this limit, a flash message will be shown and the results will be paginated with `MAX_PER_PAGE` notifications per page.

show_all_notifications

### DIFF
--- a/src/api/app/helpers/webui/notification_helper.rb
+++ b/src/api/app/helpers/webui/notification_helper.rb
@@ -1,0 +1,11 @@
+module Webui::NotificationHelper
+  def link_to_all
+    parameters = params[:type] ? { type: params[:type] } : {}
+    if params['show_all'] # already showing all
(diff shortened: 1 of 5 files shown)


src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -0,0 +1,11 @@
+module Webui::NotificationHelper
+  def show_all_link
```
danidoni, line 2: What about `link_to_all`? I think its easier to understand

saraycp: Yes, changed!

saraycp:
@krauselukas @danidoni  I rebased with master and changed the helper name as suggested.
Can you have another look?

PR #9391, opened 2020-04-15
Add task to create a bunch of notifications

saraycp:
Add notifications of different types to the database.

Run this task with `rails "dev:notifications:data[3]"` replacing 3 with any number to indicate how many times you want this code to be executed.

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -170,6 +170,59 @@ namespace :dev do
     end
   end
 
+  # Run this task with: rails "dev:notifications:data[3]"
+  # replacing 3 with any number to indicate how many times you want this code to be executed.
+  namespace :notifications do
+    task :data, [:repetitions] => :environment do |_t, args|
+      unless Rails.env.development?
+        puts "You are running this rake task in #{Rails.env} environment."
+        puts 'Please only run this task with RAILS_ENV=development'
+        puts 'otherwise it will destroy your database data.'
+        return
+      end
+
+      args.with_defaults(repetitions: 1)
+      repetitions = args.repetitions.to_i
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+
+      admin = User.where(login: 'Admin').first
+      home_admin = admin.home_project
+      unless home_admin
+        home_admin = create(:project, name: admin.home_project_name)
+        create(:relationship, project: home_admin, user: admin, role: Role.hashed['maintainer'])
+      end
+      requestor = User.where(login: 'Requestor').first || create(:confirmed_user, login: 'Requestor')
+      User.session = requestor
+      requestor_project = create(:project, name: "requestor_project_#{Faker::Lorem.word}")
+
+      repetitions.times do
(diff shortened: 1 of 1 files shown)


src/api/lib/tasks/dev.rake
```diff
@@ -170,6 +170,57 @@ namespace :dev do
...
+
+      unless home_admin
+        home_admin = create(:project, name: admin.home_project_name)
+        create(:relationship, project: home_admin, user: admin, role: Role.hashed['maintainer'])
+      end
+
+      branched_project = Project.where(name: requestor.branch_project_name('home:Admin')).first ||
+                         create(:project, name: requestor.branch_project_name('home:Admin'))
+      admin_package = Package.where(name: 'admin_package').first ||
+                      create(:package_with_file, name: 'admin_package', project: home_admin, file_content: 'from admin home')
```
krauselukas, line 197: @saraycp how about adding a timestamp or a random number to the package name? this way we could run the task a few times if a few more fake notifications are needed, without having always the same package name :)

saraycp: Good idea!

saraycp: @krauselukas I folloed your suggestion and added different names. But also improved to make it easier to generate many notification at a time.

Run this task with `rails "dev:notifications:data[3]"` replacing 3 with any number to indicate how many times you want this code to be executed.

src/api/lib/tasks/dev.rake
```diff
@@ -170,6 +170,60 @@ namespace :dev do
...
+      unless home_admin
+        home_admin = create(:project, name: admin.home_project_name)
+        create(:relationship, project: home_admin, user: admin, role: Role.hashed['maintainer'])
+      end
+      requestor = User.where(login: 'Requestor').first || create(:confirmed_user, login: 'Requestor')
+      requestor_project = create(:project, name: "requestor_project_#{Time.now.to_i}")
+
+      repetitions.times do |repetition|
+        timestamp = Time.now.to_i
+        admin_package = create(:package_with_file, name: "admin_package_#{timestamp}_#{repetition}", project: home_admin, file_content: 'from admin home')
```
hennevogel, line 200: Why not use faker? like `Faker::Lorem.word`

saraycp: Much better, updated!

src/api/lib/tasks/dev.rake
```diff
@@ -170,6 +170,59 @@ namespace :dev do
...
+          :bs_request_with_submit_action,
+          creator: requestor,
+          target_project: home_admin,
+          target_package: admin_package,
+          source_project: requestor_project,
+          source_package: requestor_package
+        )
+
+        # Will create a notification (ReviewWanted event) for this review.
+        User.session = requestor
```
eduardoj, line 213: This assignment could be moved outside the loop, couldn't it?

saraycp: sure thing

saraycp: @eduardoj  done!

Issue #9352, opened 2020-04-08
Wrong file name for files with same content/hash


ismail:
https://build.opensuse.org/package/show/hardware:boot/raspberrypi-firmware shows two "raspberry-firmware.changes" file, identical in size. But in reality there are two different files:

sisyphus ~/work/obs/hardware:boot/raspberrypi-firmware > sha1sum *.changes
8b484e2bf0ad6dabdfe12eaab73e403e73d1a4bd  raspberrypi-firmware-config.changes
8b484e2bf0ad6dabdfe12eaab73e403e73d1a4bd  raspberrypi-firmware.changes

... with same content & hash. Probably this results in showing the wrong filename for the -config variant.

ismail:
This looks fine now, did something change on the server side?

krauselukas:
Thanks for the report @ismail, I can reproduce it. Could be related to the caching, need to check it first though

ismail:
There seems to be a wide-spread issue with numbers not updating which smells like a caching issue. Like my tasks is stuck at "11" even though I have some hundred SRs in the queue.

danidoni:
![](https://github.trello.services/images/mini-trello-icon.png) [Cache bugs](https://trello.com/c/lZxR7wrx/1369-cache-bugs)

PR #9392, opened 2020-04-15
Use cache version to fetch cached users tasks

krauselukas:
Since Ruby on Rails introduced recycable cache keys, which
we enabled with the migration to Rails 6.0, we need to pass
the cache_version in combination with the cache_key (which
now is a stable value) in order to receive the most
recent data.

Related to #9352

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -804,7 +804,7 @@ class User < ApplicationRecord
 
   # TODO: Remove once responsive_ux is out of beta
   def tasks
-    Rails.cache.fetch("requests_for_#{cache_key}") do
+    Rails.cache.fetch("requests_for_#{cache_key}", version: cache_version) do
       declined_requests.count +
         incoming_requests.count +
         involved_reviews.count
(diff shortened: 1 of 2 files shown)


coolo:
wow, that would apply to basically all cache_key usages then - and the [explained benefit]( https://blog.bigbinary.com/2019/08/06/rails-adds-support-for-recyclable-cache-keys.html) of this feature does not apply to memcache as memcache does not overflow.

krauselukas:
> wow, that would apply to basically all cache_key usages then - and the [explained benefit](https://blog.bigbinary.com/2019/08/06/rails-adds-support-for-recyclable-cache-keys.html) of this feature does not apply to memcache as memcache does not overflow.

@coolo yes we will check the other places where caching is going on. not sure if disabling the cache versioning is an option (maybe there is no way around it in the future). we will do some research regarding this :)

krauselukas:
Since we are showing wrong numbers for the users tasks currently, and its a known issue, this should be merged. After the research we can still revert this, in case we are not going with the recyclable cache keys

danidoni:
[Cache bugs](https://trello.com/c/lZxR7wrx/1369-cache-bugs)

PR #9393, opened 2020-04-15
Adjust pagination in some pages

saraycp:
Some pages use [kaminari](https://github.com/amatsuda/kaminari) for pagination. However, they weren't taking the right generated views for them.

We have to specify that the views are inside `webui` with `views_prefix: 'webui'`. Otherwise, kaminari would try to find them in  `app/views/kaminari` instead of `app/views/webui/kaminari`.

**Before:**

Screenshot_2020-04-15 Open Build Service(2)

**After:**

Screenshot_2020-04-15 Open Build Service(3)

### DIFF
--- a/src/api/app/views/webui/package/revisions.html.haml
+++ b/src/api/app/views/webui/package/revisions.html.haml
@@ -17,8 +17,11 @@
             - else
               %i Revision #{revision} not found
       - unless params['show_all']
-        = paginate @revisions
-        = link_to('Show all', action: 'revisions', project: @project, package: @package, show_all: 1)
+        = paginate @revisions, views_prefix: 'webui'
+        - unless @revisions.total_pages == 1
+          .text-center= link_to('Show all',
+                                package_view_revisions_path(project: @project, package: @package, show_all: 1),
+                                class: 'btn btn-sm btn-secondary')
     - else
       %p
         %i No commits exists yet.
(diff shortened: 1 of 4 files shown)


src/api/app/views/webui/search/_results_issue.html.haml
```diff
@@ -21,4 +21,4 @@
         - else
           = truncate(result.description, length: 80)
   - if per_page
-    = paginate results
+    = paginate results, , views_prefix: 'webui'
```
dmarcoux, line 24: What is the second comma for?

saraycp: Ups, it's a typo I probably introduced right after testing the view. Thanks!!

src/api/app/views/webui/package/revisions.html.haml
```diff
@@ -17,8 +17,10 @@
             - else
               %i Revision #{revision} not found
       - unless params['show_all']
-        = paginate @revisions
-        = link_to('Show all', action: 'revisions', project: @project, package: @package, show_all: 1)
+        = paginate @revisions, views_prefix: 'webui'
+        .text-center= link_to('Show all',
```
dmarcoux, line 21: We should add `.mt-3`, otherwise `Show all` is right under the last revision.

preview

dmarcoux: It's fine though when the pagination controls are present.

preview

So I believe the issue is actually displaying `Show All` when it isn't needed. We should not display `Show All` if there are no pagination controls.

saraycp: You are right, @dmarcoux! Now it is not displayed if there's only one page to show.

It could even be improved adding a `Show less` button when displaying all. But I think for this PR. which is focused on the appearance, is enough.

PR #9400, opened 2020-04-16
Show Project monitor's repository dropdown's topmost options

### LINKED ISSUE #9350: Project Monitor z-index issue
In beta there is a problem with the z-index of the selectors on the project monitor page. I can't reach the `All` and `None` helpers

## Beta
Screenshot from 2020-04-08 12-42-58

## Not-Beta
Screenshot from 2020-04-08 12-43-20

danidoni:
When there are many repositories to render in the Repositories dropdown the
topmost options were being rendered behind the top navbar.

Raising the z-index of the dropdown's content, we make them visible again.

Fixes #9350

**Before**
image

**After**
image

### DIFF
--- a/src/api/app/assets/stylesheets/webui/application.scss
+++ b/src/api/app/assets/stylesheets/webui/application.scss
@@ -59,6 +59,7 @@
 @import 'image_templates';
 @import 'search';
 @import 'monitor';
+@import 'project-monitor';
 @import 'cloud_upload';
 
 @import 'responsive_ux';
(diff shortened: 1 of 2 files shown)


dmarcoux:
The issue is that the navigation is fixed at the top of the page with `.fixed-top` and this class sets a `z-index` of `1030`, which is more than all dropdowns in Bootstrap (they have `1000`). This is documented in the [official docs](https://getbootstrap.com/docs/4.1/layout/overview/#z-index).

So to fix this, we could change the `z-index` of all dropdowns to be more than `1030`, but as the official documentation says:

> We don’t encourage customization of these individual values; should you change one, you likely need to change them all.

I would strongly be against this. In the end, we have this issue because the project monitor page is using dropdowns in a quite questionable way. This page needs to be redesigned and changing the z-index of dropdowns is not going to fix this poor design.

danidoni:
As we talked offline, I'll fix all the dropdowns present in that page and we'll rethink this page to avoid this problem in the future.

danidoni:
@dmarcoux I've fixed the other three dropdowns. Review it again please

PR #9407, opened 2020-04-16
Add perltidy linter and configuration

eduardoj:
We want to benefit from having a linter that checks our perl files.

Initially it is only run on only one file. The second commit of this pull request is an example of fixing the offenses of the linter for only one more perl file.

### DIFF
--- a/.circleci/conditional_config.yml
+++ b/.circleci/conditional_config.yml
@@ -25,6 +25,10 @@ parameters:
     type: boolean
     default: false
 
+  run-perl-linter:
+    type: boolean
+    default: false
+
 images:
   - &common_frontend_config
     user: frontend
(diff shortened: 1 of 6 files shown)


eduardoj:
With this initial configuration adding a new perl file under the already excluded subdirectories (`src/backend/` is one of them) will have no effect for the linter. It should be explicitly included in the `dist/run_shellcheck.sh` file.

Are you fine with this approach, @openSUSE/obs-backend?

dmarcoux:
What is the benefit of linting perl files if we don't lint `src/backend`? This is where most of the perl files are.

eduardoj:
> What is the benefit of linting perl files if we don't lint `src/backend`? This is where most of the perl files are.

The idea is to include subdirectories as they get linted, progressively, and not to require using the linter for every new file now.

dmarcoux:
I searched a bit to see if there's the concept of a TODO file like `RuboCop`, but it seems like it's not the case. So including sub-directories as they get linted is the right approach.

eduardoj:
This current approach of introducing perltidy using Test::PerlTidy doesn't scale. The files to be checked by the linter can only have the following extensions: pl, pm, PL, t. See [this line](https://metacpan.org/release/SHLOMIF/Test-PerlTidy-20210709/source/lib/Test/PerlTidy.pm#L141). Closing.

PR #9412, opened 2020-04-17
Disable cache versioning

krauselukas:
Since we lately experienced some issues related to the newly
introduced cache versioning, we should disable it for now and
first adjust/check our codebase accordingly.

### DIFF
--- a/src/api/config/application.rb
+++ b/src/api/config/application.rb
@@ -87,6 +87,10 @@ module OBSApi
 
     config.action_controller.perform_caching = true
 
+    # Don't use cache versioning for now
+    config.active_record.cache_versioning = false
+    config.active_record.collection_cache_versioning = false
+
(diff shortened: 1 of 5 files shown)


hennevogel:
So this config option is going away later in Rails?

krauselukas:
> So this config option is going away later in Rails?

@hennevogel I couldn't find anything mentioning that this will be dropped, I checked the 6.1 Milestones etc.

src/api/config/application.rb
```diff
@@ -87,6 +87,12 @@ class Application < Rails::Application
 
     config.action_controller.perform_caching = true
 
+    # FIXME: Check all the places in the codebase where we cache content
```
krauselukas, line 90: @hennevogel so maybe not marking as fixme?

hennevogel: yeah just disable it! :)

PR #9413, opened 2020-04-17
Add counter to notifications creation rake task

krauselukas:
Add some improvements to the notifications rake tasks.

- The Faker::Lorem.word starts to repeat words after a short while. This leads to validation errors, when creating the packages in the rake task. So both the current time and a counter has been used to get non-repeated names.
- Add a reason (comment) when creating a review, so they will have some text to show as excerpt.
- Ensure admin user has notifications via web enabled in the configuration.
- Create notifications for more than one project to have some options in the filter by project.
- Generates notifications for event type 'RequestStatechange'.

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -186,39 +186,53 @@ namespace :dev do
       require 'factory_bot'
       include FactoryBot::Syntax::Methods
 
-      admin = User.where(login: 'Admin').first
-      home_admin = admin.home_project
-      unless home_admin
-        home_admin = create(:project, name: admin.home_project_name)
-        create(:relationship, project: home_admin, user: admin, role: Role.hashed['maintainer'])
-      end
+      # Users
+      admin = User.where(login: 'Admin').first || create(:admin_user, login: 'Admin')
+      subscribe_to_all_notifications(admin)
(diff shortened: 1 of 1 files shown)


saraycp:
@vpereira, @krauselukas I took advantage of this PR to add some other improvements to the rake task. They are detailed in the description.

Regarding the proposal to use `create_list`:
> I don't know but you know, but the "right" way to create a list of objects with FactoryBot is using `create_list`

I tried different ways but it seems it's very useful for tests, where we always start from an empty database. However, it's not so useful for this rake task, since there isn't an easy way to avoid that `create_list` repeats names.

src/api/lib/tasks/dev.rake
```diff
@@ -186,39 +186,53 @@ namespace :dev do
       require 'factory_bot'
       include FactoryBot::Syntax::Methods
 
+      # Users
       admin = User.where(login: 'Admin').first
```
DavidKang, line 190: What happens if the user Admin doesn't exist? I will suggest something like you do in [line 192](https://github.com/openSUSE/open-build-service/pull/9413/files#diff-d1bd1d1c8aec8f6a9617aa0dbf3a8e01R192).

saraycp: I assume it's a strange case, but it can happen. Updating...

krauselukas:
@saraycp nice! :)

PR #9420, opened 2020-04-20
excludebuild support in build config

adrianschroeter:
The opposite of onlybuild to blacklist builds of specific packages on project level.


src/backend/BSSched/Checker.pm
```diff
@@ -847,8 +865,15 @@ sub checkpkgs {
     }
 
     # check if this package is excluded by prjconf whitelist
```
mlschroe, line 867: Please make that comment `prjconf black/whitlist`

PR #9426, opened 2020-04-21
Preview on notification

saraycp:
According to the type, the notifications list will also show an excerpt
of the description (requests), body (comments) or reason (reviews).

Screenshot_2020-04-21 Open Build Service

On `RequestStatechange` notifications, both old and new states are included in the title.

Screenshot from 2020-04-21 13-10-10

### DIFF
--- a/src/api/app/presenters/notification_presenter.rb
+++ b/src/api/app/presenters/notification_presenter.rb
@@ -31,4 +31,18 @@ class NotificationPresenter < SimpleDelegator
       'Comment'
     end
   end
+
+  def excerpt
+    text =  case @model.notifiable_type
+            when 'Request'
+              @model.notifiable.description
+            when 'Review'
+              @model.notifiable.reason
+            when 'Comment'
+              @model.notifiable.body
+            else
+              ''
+            end
+    text.to_s.truncate(100)
+  end
 end
--- a/src/api/app/models/event/request_statechange.rb
+++ b/src/api/app/models/event/request_statechange.rb
@@ -6,7 +6,7 @@ module Event
     receiver_roles :source_maintainer, :target_maintainer, :creator, :reviewer, :source_watcher, :target_watcher
 
     def subject
-      "Request #{payload['number']} changed to #{payload['state']} (#{actions_summary})"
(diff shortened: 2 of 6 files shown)


src/api/app/presenters/notification_presenter.rb
```diff
@@ -31,4 +31,18 @@ def notification_badge
       'Comment'
     end
   end
+
+  def excerpt
```
danidoni, line 35: You can move each one of the `when`'s into the `@model` as a `excerpt` method and here just forward the `excerpt` to the `@model`

saraycp: @model is always a `Notification`'s object. Do you mean moving the `excerpt` method to that model? In that case, it will be identical, with the same `case` inside.

Or do you mean moving it to each notifiable model (Comment, Review, Request) ?
In such case, I don't know if it is an advantage since it's only used for notifications :thinking:

dmarcoux: I prefer if it stays here in the presenter since it's logic only for views dealing with notifications.

danidoni: > `@model` is always a `Notification`'s object. Do you mean moving the `excerpt` method to that model? In that case, it will be identical, with the same `case` inside.
> 
> Or do you mean moving it to each notifiable model (Comment, Review, Request) ?
> In such case, I don't know if it is an advantage since it's only used for notifications thinking

I meant on the notifiable object. But It's just a :thought_balloon:

saraycp: In fact, if we move `excerpt` to Notification model, applying the same logic, we'd have to do the same with the rest of the methods in the NotificationPresenter.

saraycp: Ok, so I think I can leave it in the NotificationPresenter. But I've simplified the code a bit, using `notifiable_type`instead of `event_type`. In case you want to have another look, @danidoni

danidoni: Okay!

PR #9427, opened 2020-04-21
Do not show download links when no published artifacts exist

### LINKED ISSUE #3151: Handle redirection to not existing repo better
Issue/Feature description
====================

* Clicking on "Go to download repository" can lead to a 404 error when the repository never gots published. This is the case the case for the default branch repositories for example.

Expected result
============

Either not showing the link when not available or have a general warning when publish is disabled that the content might not exists or would be outdated.

danidoni:
When the build is not published, the path on the download server is not present
yet, so the link points to a non-existent resource.

This PR avoids rendering the link when we're certain we have no published
artifacts yet.

Fixes #3151

### DIFF
--- a/src/api/.rubocop.yml
+++ b/src/api/.rubocop.yml
@@ -94,6 +94,10 @@ Metrics/BlockLength:
     - 'test/test_consistency_helper.rb'
     - 'test/unit/package_test.rb'
 
+Metrics/ClassLength:
+  Exclude:
+    - 'app/controllers/webui/package_controller.rb'
+
 ##################### Rails ##################################
 
 Rails:
--- a/src/api/.rubocop.yml
+++ b/src/api/.rubocop.yml
@@ -94,10 +94,6 @@ Metrics/BlockLength:
     - 'test/test_consistency_helper.rb'
     - 'test/unit/package_test.rb'
 
-Metrics/ClassLength:
-  Exclude:
-    - 'app/controllers/webui/package_controller.rb'
-
 ##################### Rails ##################################
 
 Rails:
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -192,7 +192,6 @@ class Webui::PackageController < Webui::WebuiController
   def binaries
     @repository = params[:repository]
     @package_name = params[:package]
-    @state_presenter = Repositories::StatePresenter.new(@project, @repository, @configuration)
 
     results_from_backend = Buildresult.find_hashed(project: @project, package: @package_name, repository: @repository, view: ['binarylist', 'status'])
     if results_from_backend.empty?
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -192,6 +192,7 @@ class Webui::PackageController < Webui::WebuiController
   def binaries
     @repository = params[:repository]
     @package_name = params[:package]
+    @state_presenter = Repositories::StatePresenter.new(@project, @repository, @configuration)
 
     results_from_backend = Buildresult.find_hashed(project: @project, package: @package_name, repository: @repository, view: ['binarylist', 'status'])
     if results_from_backend.empty?
--- a/src/api/app/controllers/webui/repositories_controller.rb
+++ b/src/api/app/controllers/webui/repositories_controller.rb
@@ -114,7 +114,9 @@ class Webui::RepositoriesController < Webui::WebuiController
   end
 
   # GET project/repository_state/:project/:repository
-  def state; end
+  def state
+    @state_presenter = Repositories::StatePresenter.new(@project, @repository, @configuration)
+  end
 
   # POST /project/create_dod_repository
   def create_dod_repository
--- a/src/api/app/presenters/repositories/state_presenter.rb
+++ b/src/api/app/presenters/repositories/state_presenter.rb
@@ -1,0 +1,11 @@
+module Repositories
+  class StatePresenter < SimpleDelegator
+    attr_reader :published_repository_exist, :download_area_url
+
+    def initialize(project, repository, configuration)
+      download_url = configuration['download_url']
+      @published_repository_exist = Backend::Api::Published.published_repository_exist?(project.to_s, repository.to_s)
+      @download_area_url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
+    end
+  end
+end
--- a/src/api/app/presenters/repositories/state_presenter.rb
+++ b/src/api/app/presenters/repositories/state_presenter.rb
@@ -3,7 +3,10 @@ module Repositories
     attr_reader :published_repository_exist, :download_area_url
 
     def initialize(project, repository, configuration)
+      @published_repository_exist = false
       download_url = configuration['download_url']
+      return unless download_url
+
       @published_repository_exist = Backend::Api::Published.published_repository_exist?(project.to_s, repository.to_s)
       @download_area_url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
     end
--- a/src/api/app/views/webui/package/binaries.html.haml
+++ b/src/api/app/views/webui/package/binaries.html.haml
@@ -4,7 +4,8 @@
     %h3 State of #{@repository} for #{@project} / #{@package_name}
     %ul.list-inline
       - if @configuration['download_url']
-        = render partial: 'webui/shared/download_repository_link', locals: { presenter: @state_presenter }
+        = render partial: 'webui/shared/download_repository_link',
+          locals: { presenter: Repositories::StatePresenter.new(@project, @repository, @configuration) }
       - if @buildresults.present? && policy(@package).update?
         %li.list-inline-item
           = link_to('#', data: { toggle: 'modal', target: '#delete-all-binaries-modal' }, title: 'Delete all built binaries') do
--- a/src/api/app/views/webui/package/binaries.html.haml
+++ b/src/api/app/views/webui/package/binaries.html.haml
@@ -4,8 +4,7 @@
     %h3 State of #{@repository} for #{@project} / #{@package_name}
     %ul.list-inline
       - if @configuration['download_url']
-        = render partial: 'webui/shared/download_repository_link', locals: { project: @project, repository: @repository,
-                                                                       download_url: @configuration['download_url'] }
+        = render partial: 'webui/shared/download_repository_link', locals: { presenter: @state_presenter }
       - if @buildresults.present? && policy(@package).update?
         %li.list-inline-item
           = link_to('#', data: { toggle: 'modal', target: '#delete-all-binaries-modal' }, title: 'Delete all built binaries') do
--- a/src/api/app/views/webui/shared/_download_repository_link.html.haml
+++ b/src/api/app/views/webui/shared/_download_repository_link.html.haml
@@ -1,5 +1,8 @@
-- url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
-%li.list-inline-item
-  = link_to(url, title: 'Go to download repository', class: 'nav-link') do
-    %i.fas.fa-download
-    Go to download repository
+- if presenter.published_repository_exist
+  %li.list-inline-item
+    = link_to(presenter.download_area_url, title: 'Go to download repository', class: 'nav-link') do
+      %i.fas.fa-download
+      Go to download repository
+- else
+  %li.list-inline-item
+    There are no published packages
--- a/src/api/lib/backend/api/published.rb
+++ b/src/api/lib/backend/api/published.rb
@@ -20,6 +20,11 @@ module Backend
           Xmlhash.parse(response)['buildid']
         end
       end
+
+      def self.published_repository_exist?(project_name, repository_name)
+        response = http_get(['/published/:project/:repository', project_name, repository_name])
+        !Xmlhash.parse(response).empty?
+      end
     end
   end
 end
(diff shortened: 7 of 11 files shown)


src/api/app/views/webui/shared/_download_repository_link.html.haml
```diff
@@ -1,5 +1,11 @@
 - url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
-%li.list-inline-item
-  = link_to(url, title: 'Go to download repository', class: 'nav-link') do
-    %i.fas.fa-download
-    Go to download repository
+- result = Buildresult.summary(project)
+- published = result[repository.name].select {|s| s.state != "unpublished" }
```
hennevogel, line 3: ## Flags
- A `Flag` is an on/off switch for *something*. Like
  - allow (`status=enable`) publishing (`flag=publish`) for the `Architecture` x86_64  (`architecture_id=1`)
  - do not (`status=disable`) build (`flag=build`) for a specific `Repository`  (`repo=openSUSE_Tumbleweed`)
  - allow (`status=enable`) building  (`flag=build`) for a specific `Repository`  (`repo=openSUSE_Tumbleweed`) in the `Architecture`  x86_64  (`architecture_id=1`)
- More specific `Flags` overwrite unspecific flags. In the example above the third `Flag` overwrites the second.
- `Flags` have global defaults ([FlagHelper::TYPES](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/helpers/flag_helper.rb#L6))

## Project
- A Project **can** have many `Flags`
- If a `Project` does not have a `Flag` for a specific thing, the global default is taken.

## Package
- A `Package` **can** have many `Flag`s
- If a `Package` does not have a `Flag` for a specific thing, the `Flag`s of its `Project` are taken into account.

danidoni: Okay, I've changed the code to use `Flag`s

src/api/app/views/webui/shared/_download_repository_link.html.haml
```diff
@@ -1,5 +1,10 @@
 - url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
-%li.list-inline-item
-  = link_to(url, title: 'Go to download repository', class: 'nav-link') do
-    %i.fas.fa-download
-    Go to download repository
+- publish_disabled = project.flags.find_by_flag_and_status('publish', 'disable').present?
```
hennevogel, line 2: And what if a more specific flag on any package in this project overwrites this?

danidoni: I've been thinking into this. The only use case where the download could be completely disabled by the packages is that all packages have the publishing disabled. Because if only one package has publishing enabled, the download page will be present. So I'll loop over the project's packages and mark as `publish_disabled` if the project publishing is disabled or if all the project's packages have the publishing disabled.

adrianschroeter:
On Donnerstag, 23. April 2020, 08:44:40 CEST wrote Dani Donisa:
> @danidoni commented on this pull request.
> 
> 
> 
> > @@ -1,5 +1,10 @@
>  - url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
> -%li.list-inline-item
> -  = link_to(url, title: 'Go to download repository', class: 'nav-link') do
> -    %i.fas.fa-download
> -    Go to download repository
> +- publish_disabled = project.flags.find_by_flag_and_status('publish', 'disable').present?
> 
> I've been thinking into this. The only use case where the download could be completely disabled by the packages is that all packages have the publishing disabled. Because if only one package has publishing enabled, the download page will be present. So I'll loop over the project's packages and mark as `publish_disabled` if the project publishing is disabled or if all the project's packages have the publishing disabled.

the current state of the flag is not saying that there was not a publish before.

=> publish disable is not removing files from former publish.

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

danidoni:
> the current state of the flag is not saying that there was not a publish before. => publish disable is not removing files from former publish.

Yes, I am aware of that, but I don't know any other mechanism. Do we have a way to know if we currently have published artifacts?

adrianschroeter:
On Donnerstag, 23. April 2020, 10:05:06 CEST wrote Dani Donisa:
> > the current state of the flag is not saying that there was not a publish before. => publish disable is not removing files from former publish.
> 
> Yes, I am aware of that, but I don't know any other mechanism. Do we have a way to know if we currently have published artifacts?

checking the /published route ?

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

danidoni:
> checking the /published route ?

I'm using the Backend API to check for the download area presence. Thanks!

src/api/app/views/webui/shared/_download_repository_link.html.haml
```diff
@@ -1,5 +1,10 @@
 - url = "#{download_url}/#{project.to_s.gsub(/:/, ':/')}/#{repository}"
-%li.list-inline-item
-  = link_to(url, title: 'Go to download repository', class: 'nav-link') do
-    %i.fas.fa-download
-    Go to download repository
+- download_area_exist = Backend::Api::Published.download_area_exist?(project, repository)
```
vpereira, line 2: view doing a query direct to the backend?

danidoni:
> Please views don't query backend.
> 
> The best solution here is to have a presenter doing it. Helpers shouldn't query the backend either. Make sure that before call thing it the authorization and permissions check where used.

So true, fixing.

src/api/app/views/webui/shared/_download_repository_link.html.haml
danidoni: Moved into a presenter

danidoni:
> Please views don't query backend.
> 
> The best solution here is to have a presenter doing it. Helpers shouldn't query the backend either. Make sure that before call thing it the authorization and permissions check where used.

As the presenter is being filled in the `state` action in the controller (see [here](https://github.com/openSUSE/open-build-service/pull/9427/commits/fd53f3455f8ca5f2e811c4510976c0b1d01c2232#diff-d27248312946013db3c39839d2291dfaR117)), authorization [is already checked](https://github.com/openSUSE/open-build-service/pull/9427/commits/fd53f3455f8ca5f2e811c4510976c0b1d01c2232#diff-d27248312946013db3c39839d2291dfaR7)

src/api/app/controllers/webui/repositories_controller.rb
```diff
@@ -113,7 +113,14 @@ def destroy
   end
 
   # GET project/repository_state/:project/:repository
-  def state; end
+  def state
+    state_presenter = Repositories::StatePresenter.new(@project, @repository, @configuration)
+    render 'state', locals: {
```
hennevogel, line 118: Why do you explicit call render here instead of passing on the presenter as instance variable?

src/api/lib/backend/api/published.rb
```diff
@@ -20,6 +20,13 @@ def self.build_id(project_name, repository_name)
           Xmlhash.parse(response)['buildid']
         end
       end
+
+      def self.download_area_exist?(project, repository)
```
hennevogel, line 24:
```suggestion
      def self.published_repository_exist?(project, repository)
```

src/api/app/controllers/webui/repositories_controller.rb
danidoni: To avoid using instance variables, the locals provide a bit of checking for typos and everything else

src/api/lib/backend/api/published.rb
danidoni: I like it

src/api/app/controllers/webui/repositories_controller.rb
hennevogel: Hm, we don't avoid instance variables anywhere else. And I really think we should guard against typos by reviews and/or tests...

danidoni: Okay!

src/api/app/presenters/repositories/state_presenter.rb
```diff
@@ -0,0 +1,11 @@
+module Repositories
+  class StatePresenter < SimpleDelegator
+    attr_reader :download_area_exist, :download_area_url
+
+    def initialize(project, repository, configuration)
+      download_url = configuration['download_url']
+      @download_area_exist = Backend::Api::Published.download_area_exist?(project, repository)
```
hennevogel, line 7: You have renamed this

danidoni: Yeah... Fixed

src/api/app/presenters/repositories/state_presenter.rb
```diff
@@ -4,7 +4,7 @@ class StatePresenter < SimpleDelegator
 
     def initialize(project, repository, configuration)
       download_url = configuration['download_url']
-      @download_area_exist = Backend::Api::Published.download_area_exist?(project, repository)
+      @download_area_exist = Backend::Api::Published.published_repository_exist?(project, repository)
```
hennevogel, line 7: I think you should also rename the instance variable. There is not such thing as `download area` used in our terminology...

danidoni: Yeah, changing

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/ClassLength
```
hennevogel, line 1: You should exclude the file in the config

hennevogel:
@danidoni time to squash it I guess :+1:

src/api/lib/backend/api/published.rb
```diff
@@ -20,6 +20,13 @@ def self.build_id(project_name, repository_name)
           Xmlhash.parse(response)['buildid']
         end
       end
+
+      def self.published_repository_exist?(project, repository)
```
DavidKang, line 24: To be consistent, I would pass as parameters only the name of the project and the name of the repository. Then the next [2 lines](https://github.com/openSUSE/open-build-service/pull/9427/files#diff-51b1b7f7dd955063df614ebf0b46831eR25-R26) wouldn't be needed.

danidoni: Done

src/api/app/presenters/repositories/state_presenter.rb
```diff
@@ -0,0 +1,11 @@
+module Repositories
+  class StatePresenter < SimpleDelegator
+    attr_reader :published_repository_exist, :download_area_url
+
+    def initialize(project, repository, configuration)
+      download_url = configuration['download_url']
```
hennevogel, line 6: This can be `nil`. See https://obs-reviewlab.opensuse.org/danidoni-render_download_repository_link_only_if_published_artifacts_exist/project/repository_state/home:Admin/openSUSE_Leap_15.2

danidoni: Ok, I'll make `@published_repository_exist` to be false when there is no `download_url` configured.

vpereira: should everything happen in the `initialize` method?

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -157,6 +157,7 @@ def binary
   def binaries
     @repository = params[:repository]
     @package_name = params[:package]
```
vpereira, line 159: view or helper call a presenter. Not the controller. https://www.rubyguides.com/2019/09/rails-patterns-presenter-service/

src/api/app/presenters/repositories/state_presenter.rb
vpereira: should everything happen in the initialize method?

src/api/app/views/webui/package/binaries.html.haml
```diff
@@ -4,8 +4,7 @@
     %h3 State of #{@repository} for #{@project} / #{@package_name}
     %ul.list-inline
       - if @configuration['download_url']
-        = render partial: 'webui/shared/download_repository_link', locals: { project: @project, repository: @repository,
-                                                                       download_url: @configuration['download_url'] }
+        = render partial: 'webui/shared/download_repository_link', locals: { presenter: @state_presenter }
```
vpereira, line 7: you should instantiate the presenter here, not in the controller

src/api/app/presenters/repositories/state_presenter.rb
danidoni: Why not? You would delay the evaluation in the calling methods if the computation can change from call to call, but I'm doing it because the data passed as configuration won't change, ever. So there is no reason to lazy load it. 

You could say it's a matter of taste. And you would be right. I have no problem in changing it if there is a technical reason, as you know :)

src/api/app/controllers/webui/package_controller.rb
danidoni: Fixed

src/api/.rubocop.yml
```diff
@@ -69,6 +69,10 @@ Metrics/BlockLength:
     - 'spec/**/*'
     - '**/*.rake'
 
+Metrics/ClassLength:
+  Exclude:
+    - 'app/controllers/webui/package_controller.rb'
```
dmarcoux, line 74: I wonder why this is needed now. This controller has been huge for already a loooooong time. I would prefer to increase the value of `Max` here instead:

https://github.com/openSUSE/open-build-service/blob/41aa41acd7b3ab38b4661d6dd5bfcad6029d1c35/src/api/.rubocop_todo.yml#L189-L190

We don't create huge classes so this will be tackle in the future whenever we split up that controller.

PR #9428, opened 2020-04-21
Create notifications_projects join table

eduardoj:
Retrieving the projects of a notification, requires querying for comments, requests and reviews. We want to avoid n+1 queries instead of going through the polymorphic associations notifiable and notifiable.commentable for comments.


eduardoj:

eduardoj:
> we want to avoid n+1 queries instead of going through the polymorphic associations `notifiable` and `notifiable.commentable` for comments.

Changed. Please, re-review.

dmarcoux:
I already approved, but having a better name for the table would be nice. Someone has an idea? `notified_projects`?

danidoni:
> I already approved, but having a better name for the table would be nice. Someone has an idea? `notified_projects`?

I like that name :+1:

eduardoj:
This creation table code will be used in another pull request, with code that will use this table. Closing.

PR #9429, opened 2020-04-21
Drop gem airbrake-js-client

dmarcoux:
We could never really act on JS errors reported since they are quite cryptic and whenever rails-assets is down, our CI is blocked.


dmarcoux:
We might have to change something in our Errbit instance. @hennevogel, could you have a look since you're the admin.

hennevogel:
:sob:

PR #9405, opened 2020-04-16
Configure multiple notification channels

DavidKang:
Nowadays, we have three kinds of channels (rss, email and web), but they aren't configurable.
With this PR we want to provide to the user the possibility to manage their notification channels.

### Large
Screenshot_2020-04-16 Open Build Service(1)

### Medium
Screenshot_2020-04-17 Open Build Service

### Small
Screenshot_2020-04-17 Open Build Service(1)

PR #9430, opened 2020-04-21
Adapt Job to handle new notification channels

DavidKang:
Following with the PR #9405. This PR affects `send_event_emails_job` job (a job for creating the notifications and send email) and user notifications view.

We also added a data migration to duplicate the existent notifications to create Notification::WebItem records and also create subscriptions for `rss` and `web`.

### DIFF
--- a/src/api/app/controllers/webui/users/notifications_controller.rb
+++ b/src/api/app/controllers/webui/users/notifications_controller.rb
@@ -3,7 +3,7 @@ class Webui::Users::NotificationsController < Webui::WebuiController
   MAX_PER_PAGE = 300
 
   def index
-    notifications_for_subscribed_user = NotificationsFinder.new.for_subscribed_user
+    notifications_for_subscribed_user = User.session.notifications.for_web
     @notifications = NotificationsFinder.new(notifications_for_subscribed_user).for_notifiable_type(params[:type])
     @notifications = params['show_all'] ? show_all : @notifications.page(params[:page])
   end
--- a/src/api/app/jobs/send_event_emails_job.rb
+++ b/src/api/app/jobs/send_event_emails_job.rb
@@ -4,11 +4,7 @@ class SendEventEmailsJob < ApplicationJob
   def perform
     Event::Base.where(mails_sent: false).order(created_at: :asc).limit(1000).each do |event|
       subscribers = event.subscribers
-
-      if subscribers.empty?
-        event.update_attributes(mails_sent: true)
-        next
-      end
+      event.update(mails_sent: true) if subscribers.empty?
 
       NotificationCreator.new(event).call
       send_email(subscribers, event)
@@ -19,6 +15,7 @@ class SendEventEmailsJob < ApplicationJob
   private
 
   def send_email(subscribers, event)
+    return if subscribers.empty?
     EventMailer.event(subscribers, event).deliver_now
   rescue StandardError => e
     Airbrake.notify(e, event_id: event.id)
--- a/src/api/app/models/event_subscription/find_for_event.rb
+++ b/src/api/app/models/event_subscription/find_for_event.rb
@@ -6,7 +6,7 @@ class EventSubscription
       @event = event
     end
 
-    def subscriptions
+    def subscriptions(channel = :instant_email)
       receivers_and_subscriptions = {}
 
       event.class.receiver_roles.flat_map do |receiver_role|
@@ -14,8 +14,9 @@ class EventSubscription
         receivers = event.send("#{receiver_role}s")
         receivers = filter_and_convert_groups_without_emails_to_users(receivers)
 
+        options = { eventtype: event.eventtype, receiver_role: receiver_role, channel: channel }
         # Find the default subscription for this eventtype and receiver_role
-        default_subscription = EventSubscription.defaults.find_by(eventtype: event.eventtype, receiver_role: receiver_role)
+        default_subscription = EventSubscription.defaults.find_by(options)
 
         receivers.each do |receiver|
           # Prevent multiple enabled subscriptions for the same subscriber & eventtype
@@ -23,7 +24,7 @@ class EventSubscription
           next if receivers_and_subscriptions[receiver].present? || receiver == event.originator
 
           # Try to find the subscription for this receiver
-          receiver_subscription = EventSubscription.for_subscriber(receiver).find_by(eventtype: event.eventtype, receiver_role: receiver_role)
+          receiver_subscription = EventSubscription.for_subscriber(receiver).find_by(options)
 
           if receiver_subscription.present?
             # Use the receiver's subscription if it exists
--- a/src/api/app/services/notification_creator.rb
+++ b/src/api/app/services/notification_creator.rb
@@ -28,6 +28,7 @@ class NotificationCreator
   def create_notification_per_subscription(subscription, channel)
     return unless create_notification?(subscription.subscriber, channel)
     params = subscription.parameters_for_notification.merge!(@event.parameters_for_notification)
+    # TODO: Replace by Notification when we remove Notification::RssFeedItem class
     notification = Notification::RssFeedItem.find_or_create_by!(params) # avoid duplication
     notification.update("#{channel}": true)
   end
--- a/src/api/app/services/notification_creator.rb
+++ b/src/api/app/services/notification_creator.rb
@@ -5,6 +5,7 @@ class NotificationCreator
                       'Event::CommentForProject',
                       'Event::CommentForPackage',
                       'Event::CommentForRequest'].freeze
+  CHANNELS = [:web, :rss].freeze
 
   def initialize(event)
     @event = event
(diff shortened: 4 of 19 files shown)


src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -3,22 +3,20 @@ class SendEventEmailsJob < ApplicationJob
...
+      event.update_attributes(mails_sent: true) if email_subscribers.empty?
 
-      if subscribers.empty?
-        event.update_attributes(mails_sent: true)
-        next
-      end
-
-      NotificationCreator.new(event).call
-      send_email(subscribers, event)
+      NotificationCreator.new(event, :web).call
```
danidoni, line 9: Why are we calling the notification creation for `:web` and `:rss` here? I don't understand :thinking:

DavidKang: This PR is the second part of #9405, So the user will have different channels for the same `subscription`. The `NotificationCreator` will create the notification if the user has a subscription for the channel `rss` or `web`.

src/api/db/data/20200420080753_generate_web_notifications.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  private
+
+  def generate_web_notifications
+    Notification::RssFeedItem.all.each do |rss_notification|
+      duplicate = rss_notification.dup
+      duplicate.type = 'Notification::WebItem'
+      # to avoid duplicated web notifications
+      Notification::WebItem.find_or_create_by(duplicate.attributes.except('id'))
```
hennevogel, line 18: :thought_balloon:  Man I still don't like all of this data duplication...

src/api/db/data/20200420080753_generate_web_notifications.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+  end
+
+  def generate_event_subscripitions
+    event_subscriptions = EventSubscription.where(channel: :instant_email).where.not(user_id: nil)
+    event_subscriptions.each do |event_subscription|
+      create_subscription_for_channel(event_subscription, :rss)
+      create_subscription_for_channel(event_subscription, :web)
+    end
+  end
```
hennevogel, line 28: What about the global defaults? `EventSubscription.where(channel: :instant_email).where(user_id: nil)`

src/api/app/jobs/send_event_emails_job.rb
danidoni: Okay

DavidKang:
We talked offline, and we decided to go into another direction. For that reason I put this PR as WIP.

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -3,7 +3,7 @@ class Webui::Users::NotificationsController < Webui::WebuiController
   MAX_PER_PAGE = 300
 
   def index
-    notifications_for_subscribed_user = NotificationsFinder.new.for_subscribed_user
+    notifications_for_subscribed_user = NotificationsFinder.new(Notification::WebItem.all).for_subscribed_user
```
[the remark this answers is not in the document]
vpereira: if there is no way to scope it or pass a `where` there, maybe a solution would be to wrap it with `LazyObject.new { Notification::WebItem.all }`

src/api/app/models/notification/web_item.rb
```diff
@@ -0,0 +1,2 @@
+class Notification::WebItem < Notification
```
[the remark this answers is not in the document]
DavidKang: This class was removed

src/api/db/data/20200420080753_generate_web_notifications.rb
DavidKang: The migration has changed and don't duplicate the Notification.

src/api/db/data/20200420080753_generate_web_notifications.rb
```diff
@@ -0,0 +1,35 @@
...
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+
+  private
+
+  def generate_event_subscripitions
+    event_subscriptions = EventSubscription.where(channel: :instant_email).where.not(user_id: nil)
+    event_subscriptions.each do |event_subscription|
+      create_subscription_for_channel(event_subscription, :rss)
```
hennevogel, line 16: Hm. I think we only should do this if the subscription user has a `Token::Rss`

src/api/app/services/notification_creator.rb
```diff
@@ -5,23 +5,36 @@ class NotificationCreator
...
   private
 
-  def create_notification_per_subscription(subscription)
-    return if subscription.subscriber && subscription.subscriber.away?
+  def create_notification_per_subscription(subscription, channel)
+    return unless create_notification?(subscription.subscriber, channel)
     params = subscription.parameters_for_notification.merge!(@event.parameters_for_notification)
-    Notification::RssFeedItem.find_or_create_by!(params) # avoid duplication
+    notification = Notification::RssFeedItem.find_or_create_by!(params) # avoid duplication
+    notification.update_attributes("#{channel}": true)
```
vpereira, line 32: https://github.com/openSUSE/open-build-service/issues/9441 <- `update_attributes` is getting deprecated. Could you use `update` here?

src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -4,11 +4,7 @@ class SendEventEmailsJob < ApplicationJob
   def perform
     Event::Base.where(mails_sent: false).order(created_at: :asc).limit(1000).each do |event|
       subscribers = event.subscribers
-
-      if subscribers.empty?
-        event.update_attributes(mails_sent: true)
-        next
-      end
+      event.update_attributes(mails_sent: true) if subscribers.empty?
```
vpereira, line 7: please don't use `update_attributes`

src/api/app/models/event_subscription/find_for_event.rb
```diff
@@ -6,33 +6,34 @@ def initialize(event)
...
           next if receivers_and_subscriptions[receiver].present? || receiver == event.originator
 
           # Try to find the subscription for this receiver
-          receiver_subscription = EventSubscription.for_subscriber(receiver).find_by(eventtype: event.eventtype, receiver_role: receiver_role)
+          receiver_subscription = EventSubscription.for_subscriber(receiver).find_by(options)
 
           if receiver_subscription.present?
             # Use the receiver's subscription if it exists
-            if receiver_subscription.enabled?
+            if receiver_subscription.enable?
```
vpereira, line 31: Why not `enabled` ? `enable` would be more a verb, or a method to set a state and `enabled` a state. So if i understood it right, you are asking for the `state` so `if subscription.enabled?` would make more sense, no?

src/api/app/services/notification_creator.rb
```diff
@@ -5,23 +5,36 @@ class NotificationCreator
                       'Event::CommentForProject',
                       'Event::CommentForPackage',
                       'Event::CommentForRequest'].freeze
+  CHANNELS = [:web, :rss].freeze
```
vpereira, line 8: for email we don't create a notification but instead just send the email, right?

src/api/db/data/20200420080753_generate_web_notifications.rb
vpereira: cannot we enforce the relationship between Notification and an user and then get rid of this last where?

src/api/app/services/notification_creator.rb
DavidKang: yes, we just send the email

src/api/app/models/event_subscription/find_for_event.rb
DavidKang: Good point, thanks :smile_cat:

src/api/app/jobs/send_event_emails_job.rb
DavidKang: Done

src/api/app/services/notification_creator.rb
DavidKang: Done

src/api/app/models/event_subscription/find_for_event.rb
DavidKang: I applied the changes in the PR #9405

src/api/db/data/20200420080753_generate_web_notifications.rb
DavidKang: Done, we don't create subscriptions if the user doesn't have rss token

src/api/app/services/notification_creator.rb
```diff
@@ -5,23 +5,36 @@ class NotificationCreator
...
+  def create_notification_per_subscription(subscription, channel)
+    return unless create_notification?(subscription.subscriber, channel)
     params = subscription.parameters_for_notification.merge!(@event.parameters_for_notification)
-    Notification::RssFeedItem.find_or_create_by!(params) # avoid duplication
+    notification = Notification::RssFeedItem.find_or_create_by!(params) # avoid duplication
+    notification.update("#{channel}": true)
+  end
+
+  def create_notification?(subscriber, channel)
+    return false if subscriber && subscriber.away?
```
hennevogel, line 37: :thought_balloon: why are we guarding for subscriber being `nil`? Can't happen right?

src/api/app/services/notification_creator.rb
```diff
@@ -5,23 +5,36 @@ class NotificationCreator
...
-    return if subscription.subscriber && subscription.subscriber.away?
+  def create_notification_per_subscription(subscription, channel)
+    return unless create_notification?(subscription.subscriber, channel)
     params = subscription.parameters_for_notification.merge!(@event.parameters_for_notification)
-    Notification::RssFeedItem.find_or_create_by!(params) # avoid duplication
+    notification = Notification::RssFeedItem.find_or_create_by!(params) # avoid duplication
+    notification.update("#{channel}": true)
+  end
+
+  def create_notification?(subscriber, channel)
```
hennevogel, line 36: :thought_balloon:  all of this sounds like something that should be in `Subscription`

src/api/app/services/notification_creator.rb
DavidKang: As we create notifications with a DelayedJob, a subscriber could be deleted on the meantime.

src/api/app/services/notification_creator.rb
DavidKang: IMO, I think the responsibility to know if we should create a notification should be on `NotificationCreator`. But we can change it later.

PR #9431, opened 2020-04-21
Refactor ConsistencyCheckJob job

vpereira:
This PR looks big, but actually it's not.

It's basic a refactoring of two rake tasks from [delayed_job.rake](https://github.com/openSUSE/open-build-service/blob/master/src/api/lib/tasks/delayed_job.rake#L33), `check_project` and `fix_project`

The changes are:

- Removed two checks: if there are packages and projects with invalid names. We have name validation for both, on frontend and backend. 
- Added spec for the the `ConsistencyCheckJob` class itself
- Extracted auxiliary code to services 
- Added spec to the services
- Changed the `ConsistencyCheckJob` class without change it's API
- Improved documentation

### DIFF
--- a/src/api/app/jobs/consistency_check_job.rb
+++ b/src/api/app/jobs/consistency_check_job.rb
@@ -1,61 +1,39 @@
 require 'api_error'
-require 'xmlhash'
 
 class ConsistencyCheckJob < ApplicationJob
   queue_as :consistency_check
 
   def perform
-    User.get_default_admin.run_as { _perform(nil) }
+    User.get_default_admin.run_as { _perform(false) }
   end
 
-  def check_one_project(project, fix)
-    unless Project.valid_name?(project.name)
-      @errors << "Invalid project name #{project.name}\n"
-      if fix
-        # just remove it, the backend won't accept it anyway
-        project.commit_opts = { no_backend_write: 1 }
-        project.destroy
-      end
-      return
-    end
-    @errors << package_existence_consistency_check(project, fix)
-    @errors << project_meta_check(project, fix)
+  def check_one_project(project, fix = false)
+    package_existence_consistency_check(project, fix)
+    project_meta_check(project, fix)
+    @errors.flatten.reject(&:empty?).join("\n")
   end
 
-  def _perform(fix)
-    @errors = project_existence_consistency_check(fix)
-    Project.find_each(batch_size: 100) { |project| check_one_project(project, fix) }
-    return if @errors.empty?
-    @errors = "FIXING the following errors:\n" << @errors if fix
-    Rails.logger.error('Detected problems during consistency check')
-    Rails.logger.error(@errors)
-
-    AdminMailer.error(@errors).deliver_now
-    nil
+  # method called by the rake task `fix_project`
+  def fix_project(project)
+    User.get_default_admin.run_as { check_project(project, true) }
   end
 
-  # for manual fixing by admin via rails command
-  def fix_project
-    User.get_default_admin.run_as { check_project(true) }
-  end
-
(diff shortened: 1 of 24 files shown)


src/api/app/jobs/consistency_check_job.rb
```diff
@@ -9,53 +8,32 @@ def perform
   end
 
   def check_one_project(project, fix)
-    unless Project.valid_name?(project.name)
```
vpereira, line 12: All projects in the database have a valid name. This check isn't necessary.

src/api/app/jobs/consistency_check_job.rb
```diff
@@ -9,53 +8,32 @@ def perform
...
+    @errors.flatten.reject(&:empty?).join("\n")
   end
 
   def _perform(fix)
-    @errors = project_existence_consistency_check(fix)
+    @errors << project_existence_consistency_check(fix)
     Project.find_each(batch_size: 100) { |project| check_one_project(project, fix) }
     return if @errors.empty?
-    @errors = "FIXING the following errors:\n" << @errors if fix
-    Rails.logger.error('Detected problems during consistency check')
```
vpereira, line 30: The only instrumentation used in production is errbit. Random log without a strategy or reasoning behind is just noise.

src/api/app/jobs/consistency_check_job.rb
```diff
@@ -9,53 +8,32 @@ def perform
...
     nil
   end
 
-  # for manual fixing by admin via rails command
-  def fix_project
-    User.get_default_admin.run_as { check_project(true) }
-  end
-
-  def check_project(fix = nil)
-    if ENV['project'].blank?
```
vpereira, line 43: this check should be done on the rake task, not be buried inside the job class.

src/api/app/jobs/consistency_check_job.rb
```diff
@@ -1,61 +1,37 @@
...
-    @errors << project_meta_check(project, fix)
+  def check_one_project(project, fix = false)
+    package_existence_consistency_check(project, fix)
+    project_meta_check(project, fix)
+    @errors.flatten.reject(&:empty?).join("\n")
   end
 
-  def _perform(fix)
-    @errors = project_existence_consistency_check(fix)
+  def _perform(fix = false)
```
vpereira, line 16: I don't like the name `_perform` . Any idea?

danidoni: What about `check_projects_in_batches`?

src/api/app/jobs/consistency_check_job.rb
```diff
@@ -77,52 +53,30 @@ def fix
...
-      project_list_backend = dir_to_array(Xmlhash.parse(Backend::Api::Sources::Project.list))
-    rescue Backend::NotFoundError
-      # project disappeared ... may happen in running system
-      return ''
-    end
+  def project_existence_consistency_check(fix = false)
+    project_consistency_checker = ConsistencyCheckJobService::ProjectConsistencyChecker.new.call
 
-    diff = project_list_api - project_list_backend
+    diff = project_consistency_checker.diff_frontend_backend
```
vpereira, line 77: here we have some ugly code repetition that I still have to get rid of.

src/api/app/jobs/consistency_check_job.rb
vpereira: IMO it cannot be `check_*` because it can be used to "fix" the errors too

danidoni: Then if this is doing two different things, maybe its better to split the behaviour, one checker and one fixer...

PR #9433, opened 2020-04-22
Create and backfill notified_projects table

dmarcoux:


dmarcoux:

src/api/db/data/20200421121610_backfill_notified_projects.rb
```diff
@@ -0,0 +1,25 @@
...
+        when 'Package'
+          notification.projects << notification.notifiable.commentable.project
+        when 'BsRequest'
+          notification.projects << notification.notifiable.commentable.target_project_objects.uniq
+        end
+      when 'Review'
+        notification.projects << notification.notifiable.bs_request.target_project_objects.uniq
+      end
+    end
+  end
```
saraycp, line 20: :thinking: We could have problems if we run this data migration without deploying the changes that associate notifications and projects (right after notification's creation).

The reason is that since the data migration is run and until these changes are deployed,  new notifications will be created but won't associate projects to them. 

If we do this the other way around (deploying the code first and running this migration afterwards) the problem will be that we'll try to insert duplicated data. Maybe this could be solved by emptying `notified_projects` table as first step in this data migration.

We can talk about this offline in case my explanation is not clear and decide what to do afterwards.

dmarcoux: We could always rerun the data migration, it won't create duplicate records anyway. However, we need to rescue from `ActiveRecord::RecordNotUnique`

saraycp: As we discussed offline, possible duplication exceptions are going to be handled properly. 

Running the migration twice is not possible as it is tied to versions (in db/data_schema.rb), therefore the `down` part is going to be prepared to empty the whole table in case we need to rollback and migrate again (what we'll try to avoid).

src/api/app/models/notification.rb
```diff
@@ -2,6 +2,9 @@ class Notification < ApplicationRecord
   belongs_to :subscriber, polymorphic: true
   belongs_to :notifiable, polymorphic: true
 
+  has_many :notified_projects, dependent: :destroy
+  has_many :projects, through: :notified_projects
```
saraycp, line 6: Testing, I tried to remove a notification and realized this raised an error. The problem seems to be that the `dependent: :destroy` should be included in the second line. Something like:

```
  has_many :notified_projects
  has_many :projects, through: :notified_projects, dependent: :destroy
```

This way it works fine and the record in the join table gets removed after a notification is removed.

src/api/app/models/project.rb
```diff
@@ -91,6 +91,9 @@ def staging_history
 
   has_one :staging, class_name: 'Staging::Workflow', inverse_of: :project, dependent: :destroy
 
+  has_many :notified_projects, dependent: :destroy
+  has_many :notifications, through: :notified_projects
```
saraycp, line 95: See [previous comment](https://github.com/openSUSE/open-build-service/pull/9433/files#r413780164).

src/api/app/models/notification.rb
dmarcoux: It's dropping also removing the `project` with your suggestion. I don't believe it's what we want. The current version removes the associated records in `notified_projects`. What is the error you got?

saraycp: This is the error:

```
  NotifiedProject Destroy (1.3ms)  DELETE FROM `notified_projects` WHERE `notified_projects`.`` IS NULL                                                                                       
   (0.6ms)  ROLLBACK                                                                                                                                                                          
        1: from (irb):8                                                                                                                                                                       
ActiveRecord::StatementInvalid (Mysql2::Error: Unknown column 'notified_projects.' in 'where clause')
```

dmarcoux: Hum...we will need an ID because we're using a `has_many through:` association... See the docs [here](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-create_table).
```
:id

    Whether to automatically add a primary key column. Defaults to true. Join tables for ActiveRecord::Base.has_and_belongs_to_many should set it to false.

    A Symbol can be used to specify the type of the generated primary key column.
```

dmarcoux: I've adapted the migration and now it works as excepted.

eduardoj:
Test fixed. Please review.

src/api/app/services/notified_projects.rb
```diff
@@ -0,0 +1,26 @@
...
+    when 'BsRequest'
+      @notifiable.target_project_objects.uniq
+    when 'Comment'
+      case @notifiable.commentable_type
+      when 'Project'
+        @notifiable.commentable
+      when 'Package'
+        @notifiable.commentable.project
+      when 'BsRequest'
+        @notifiable.commentable.target_project_objects.uniq
```
vpereira, line 20: if `target_project_objects` is a relation, wouldn't be better to use `distinct` here?

eduardoj: True. Changed.

src/api/db/data/20200421121610_backfill_notified_projects.rb
```diff
@@ -0,0 +1,13 @@
+class BackfillNotifiedProjects < ActiveRecord::Migration[6.0]
+  def up
+    Notification.where.not(notifiable_id: nil, notifiable_type: nil).find_each do |notification|
```
vpereira, line 3: if we validate `notifiable_id` and `notifiable_type` we don't have to add this `where.not` here right?

src/api/app/services/notified_projects.rb
```diff
@@ -0,0 +1,26 @@
+class NotifiedProjects
+  def initialize(notification)
+    @notification = notification
+    @notifiable = @notification.notifiable
+  end
+
+  def call
+    return [] unless @notifiable
```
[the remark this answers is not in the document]
eduardoj: `@notification.notifiable` will return `nil` if:
- `@notification.notifiable_type` is `nil`, or
- `@notification.notifiable_id` is `nil`, or
- `@notification.notifiable_type` and`@notification.notifiable_id` are both `nil`

The code `return [] unless @notifiable` catches all these cases. But maybe it is not so readable.

And true, `Project.none` is the right thing to return. Patch applied.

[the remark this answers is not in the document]
vpereira:
```
irb(main):003:0> BsRequest.first.notifications
=> #<ActiveRecord::Associations::CollectionProxy []>
irb(main):004:0> puts "foo" unless BsRequest.first.notifications
=> nil
```

src/api/app/models/notified_project.rb
```diff
@@ -0,0 +1,4 @@
+class NotifiedProject < ApplicationRecord
+  belongs_to :notification
```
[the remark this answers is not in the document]
eduardoj: > don't we have to validate it or can we have NotifiedProject without projects or notifications?

This is already taken in account with the `null: false` in the migration of the table creation. Do you think a validation is still needed?

> Important to think about the case where project and notifications get destroyed too

This is handled with `dependent: detroy` lines both in `app/models/notification.rb` and in `/app/models/project.rb` files.

src/api/app/services/notified_projects.rb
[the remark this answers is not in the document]
danidoni: > `@notification.notifiable` will return `nil` if:
> 
>     * `@notification.notifiable_type` is `nil`, or
> 
>     * `@notification.notifiable_id` is `nil`, or
> 
>     * `@notification.notifiable_type` and`@notification.notifiable_id` are both `nil`
> 
> 
> The code `return [] unless @notifiable` catches all these cases. But maybe it is not so readable.
> 
> And true, `Project.none` is the right thing to return. Patch applied.

Just for the record, it will also return nil if `@notification.notifiable_id` points to a non existent record...

src/api/db/data/20200421121610_backfill_notified_projects.rb
vpereira: discussed with Saray, I'm fine with that!

PR #9435, opened 2020-04-22
[backend] automatically add the ccache package if useccache is set

mlschroe:
The buildflags:useccache substitute can be used in the project
config to add different packages.


mstsky115:
Hi, I want to know how to set useccache for obs-server physical machine. I set BuildFlags: useccache:myPackage in the Project Config TAB of the webpage. But when I looked at the build log, I noticed that ccache was not used to build the package. You can help me？

PR #9447, opened 2020-04-24
Notifications filtered by projects

saraycp:
This PR is a follow-up to #9433.

The notifications can be now filtered by the projects associated to them.
To do so, a list of project names and a the number of notifications per project are displayed.

Screenshot_2020-04-25 Open Build Service

### DIFF
--- a/src/api/app/controllers/webui/users/notifications_controller.rb
+++ b/src/api/app/controllers/webui/users/notifications_controller.rb
@@ -1,10 +1,20 @@
 class Webui::Users::NotificationsController < Webui::WebuiController
-  before_action :require_login
   MAX_PER_PAGE = 300
+  VALID_NOTIFICATION_TYPES = ['read', 'reviews', 'comments', 'requests', 'unread'].freeze
+
+  before_action :require_login
+  before_action :check_param_type, :check_param_project, only: :index
 
   def index
     notifications_for_subscribed_user = User.session.notifications.for_web
-    @notifications = NotificationsFinder.new(notifications_for_subscribed_user).for_notifiable_type(params[:type])
+    @notifications = if params[:project]
+                       NotificationsFinder.new(notifications_for_subscribed_user).for_project_name(params[:project])
+                     else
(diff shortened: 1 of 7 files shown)


src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -1,10 +1,23 @@
...
+    @notifications = if params[:project]
+                       NotificationsFinder.new(notifications_for_subscribed_user).for_project_name(params[:project])
+                     else
+                       NotificationsFinder.new(notifications_for_subscribed_user).for_notifiable_type(params[:type])
+                     end
+
+    @projects = Project.joins(:notifications)
+                       .where(notifications: { subscriber: User.session, delivered: false })
+                       .order('name desc').group(:name).count
+                       .sort_by(&:last).reverse.to_h
```
vpereira, line 19: wouldn't be better to just call `reverse.to_h` and fire the query when it's actually needed and have `@projects` as actually an activerecord collection?

danidoni, line 19: My only improvement is to move this four lines into a method with a descriptive name so we can know at a quick glance what this is doing.

saraycp: @vpereira this has been addressed as discussed offline. Thanks for your suggestion! :bouquet:

saraycp: @danidoni thanks! It's been moved to the method `projects_for_filter`.

saraycp:
I had to do some little changes when rebasing. Could you have another look? @vpereira and @danidoni 

Pay special attention to [this file](https://github.com/openSUSE/open-build-service/pull/9447/files#diff-3fea9b194ee535ea7887130c45fc9988R59) and  [this method](https://github.com/openSUSE/open-build-service/pull/9447/files#diff-3fea9b194ee535ea7887130c45fc9988R59).

PR #9449, opened 2020-04-27
Track syntax changes in db/structure.sql

dmarcoux:
When I ran the last migrations with `bundle exec rake db:migrate`, those changes happened.


hennevogel:
ran where?

dmarcoux:
> ran where?

In my local development environment with Docker.

hennevogel:
> In my local development environment with Docker.

Which is different to what we CI/use in production and those syntax changes are mariadb version dependent....

hennevogel:
It's also why the test fails :smile:

PR #9453, opened 2020-04-28
[api] add_containers: pick right container in maintenance_release pro…

adrianschroeter:
…jects

we have multiple candidates there and we need to take the incident counter
or time stamp suffix into account.

Should also speed up things as we sort candidates earlier that way.


src/api/app/models/package.rb
```diff
@@ -907,14 +907,22 @@ def add_containers(opts = {})
       next unless origin_package
       origin_package.binary_releases.where(obsolete_time: nil).find_each do |binary_release|
         mc = binary_release.medium_container
-        container_list[mc] = 1 if mc
+        if mc
+          mc_update_project = mc.project.update_instance
+          # pick only one and the highest container.
+          identifier = mc_update_project.name + '/' + mc.name
```
vpereira, line 913: would you mind to break it in smaller methods?

vpereira (dismissed):

src/api/app/models/package.rb
hennevogel: @adrianschroeter anything we can help you with?

adrianschroeter:
On Montag, 27. Juli 2020, 18:58:12 CEST wrote Henne Vogelsang:
> 
> 
> 
> > @@ -907,14 +907,22 @@ def add_containers(opts = {})
>        next unless origin_package
>        origin_package.binary_releases.where(obsolete_time: nil).find_each do |binary_release|
>          mc = binary_release.medium_container
> -        container_list[mc] = 1 if mc
> +        if mc
> +          mc_update_project = mc.project.update_instance
> +          # pick only one and the highest container.
> +          identifier = mc_update_project.name + '/' + mc.name
> 
> @adrianschroeter anything we can help you with?

well, if you want to refactor it just do it. It is in production
since the creation and people can work since then. But please be wise
picking method names since distributing the code makes it harder 
to read (esp. when misleading method names get picked as in some
other cases).

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

eduardoj:
I rebased on master.

src/api/app/models/package.rb
hennevogel: I just stared at this a bit but found no obvious way to split this. Let's move all of this out into a concern...

hennevogel:
No idea how to test this 0.1% better, merging...

PR #9456, opened 2020-04-28
Show the correct number of unread notifications

### LINKED ISSUE #9443: Wrong Notification count
Screenshot from 2020-04-24 10-42-44

```
<about>
  <title>Open Build Service API</title>
  <description>API to the Open Build Service</description>
  <revision>2.11~alpha.20200421T113048.7be98971ab</revision>
  <last_deployment>2020-04-21 12:50:04 +0200</last_deployment>
</about>
```

There is data

```ruby
irb(main):005:0> User.find_by(login: 'hennevogel').notifications.count
=> 2
```
But the wrong kind...
```ruby
irb(main):006:0> User.find_by(login: 'hennevogel').notifications.where.not(type: 'Notification::RssFeedItem').count
=> 0
```

eduardoj:
Before we were counting all unread notifications. Now we only count those that have a notifiable.

Fixes #9443.

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -813,7 +813,7 @@ class User < ApplicationRecord
   end
 
   def unread_notifications
-    NotificationsFinder.new(notifications.for_web).unread.size
+    NotificationsFinder.new(notifications.for_web).unread_with_notifiable.size
   end
 
   def watched_project_names
--- a/src/api/app/queries/notifications_finder.rb
+++ b/src/api/app/queries/notifications_finder.rb
@@ -11,6 +11,10 @@ class NotificationsFinder
     @relation.where(delivered: false)
   end
 
+  def unread_with_notifiable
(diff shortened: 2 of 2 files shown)


eduardoj:
Rebased on master to avoid triggering the SourcePackageFinder flickering test.

hennevogel:
Why do we still have Notifications without notifiable?

DavidKang:
Because in the past, we created `Notification` for `BuildFail` and `ServiceFail` events.
When we started with notifications, we excluded these events, but we haven't decided what to do with them (BuildFail and ServiceFail).

So, I would say that we should make another data migration to delete the Notifications with event_type `Event::BuildFail` and `Event::ServiceFail`.

hennevogel:
So we can close this and you do the migration instead right?

DavidKang:
@hennevogel, then I will close this PR.

PR #9470, opened 2020-04-29
Disable builds flag when accepting a StagingProject

### LINKED ISSUE #7367: Build disable accepting staging projects
As accepting requests not only takes a while but also changes sources within the staging project, build disable the staging project before accepting.

Once it's empty, it should be build disabled anyway - so doing it upfront saves the scheduler from recalculating scheduled jobs during accepting phase.

DavidKang:
To not re-trigger the builds when we are accepting a
StagingProject, we disable all the build flags.

Fix #7367


src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -52,6 +52,14 @@ def accept
 
     # check general state
     raise StagingProjectNotAcceptable, 'Staging project is not in state acceptable.' unless can_accept?
+
+    # Disabling build for all repositories and architectures.
+    build_flag = @staging_project.flags.find_or_initialize_by(flag: 'build', repo: nil, architecture_id: nil)
+    build_flag.update(status: 'disable')
```
danidoni, line 58: What if you move this into `StagingProject#disable!` or something like this?

src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -52,6 +52,14 @@ def accept
...
+
+    # Remove all the build flags enabled by the user.
+    @staging_project.flags.where(flag: 'build', status: 'enable').destroy_all
```
danidoni, line 61: :thought_balloon: This could be a `StagingProject#clear_flags!`

coolo: remembere that StagingProject is just a concern included into all projects.

src/api/app/controllers/staging/staging_projects_controller.rb
danidoni: We've been talking offline and we've decided to not apply this suggestion.

src/api/app/controllers/staging/staging_projects_controller.rb
danidoni: We've been talking offline and we've decided to not apply this suggestion.

PR #9478, opened 2020-04-30
Add autocomplete functionality to Staging::ExcludedRequests controller

vpereira:
Add method autocomplete to model Staging::Workflow

The method look exclusively for unassigned requests (backlog)
and filter by id containing specific digits

To see it working, either you change the value of `minLength` in the `autocomple.js` to 1 or you create more than 10 requests

### DIFF
--- a/src/api/app/models/staging/workflow.rb
+++ b/src/api/app/models/staging/workflow.rb
@@ -82,6 +82,10 @@ class Staging::Workflow < ApplicationRecord
     users_hash
   end
 
+  def autocomplete(num)
+    unassigned_requests.where('CAST(bs_requests.number AS CHAR) LIKE ?', "%#{num}%")
+  end
+
   private
 
   def create_staging_projects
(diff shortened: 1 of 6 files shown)


vpereira:
[M - As a staging manager I want to pick excluded requests from an autocomplete interface of the requests from the backlog](https://trello.com/c/DjbqHBsF/1385-m-as-a-staging-manager-i-want-to-pick-excluded-requests-from-an-autocomplete-interface-of-the-requests-from-the-backlog)

src/api/app/models/staging/workflow.rb
```diff
@@ -82,6 +82,10 @@ def self.load_users(staging_projects)
     users_hash
   end
 
+  def autocomplete(num)
+    unassigned_requests.where('CAST(bs_requests.id AS CHAR) LIKE ?', "%#{num}%")
```
DavidKang, line 86: I have a concern, regarding this. Here you are searching by bs_request.id, and in the [controller](https://github.com/openSUSE/open-build-service/pull/9478/files#diff-b778c1c1eeae49afc99a16fa4a269f14R59) you returns the number of the requests that could or not the same as the request number.

vpereira: updated, could you please take a look @DavidKang ?

dmarcoux:
The autocomplete does work, but it's not providing a lot of value in my opinion... It would be much nicer if we could at least see the package name or an excerpt of the request's description.

PR #9486, opened 2020-05-04
Add missing API docu

### LINKED ISSUE #8117: API Documentation is missing information on DELETE requests from Staging
# API Docs is missingt the body example how the data should look like
# when deleting requests from a Staging

https://build.opensuse.org/apidocs/index.html#187

Expected Result
============
A documentation of what the body of the transmitted data should look like to delete requests. 
Like for adding requests it is: 

```
<requests>
<number>7</number>
<number>13</number>
</requests>

```
How to Reproduce
===============
1. Go to ''https://build.opensuse.org/apidocs/index.html#187
2. Look what you should submit 
3. Fail to find :)

quatran:


quatran:
[S - Add missing API docu](https://trello.com/c/9ADeaXWP/1382-s-add-missing-api-docu)

docs/api/api/api.txt
```diff
@@ -1953,6 +1953,7 @@ XmlResult: status_ok
 
 DELETE /staging/<staging_workflow_project>/workflow
   Delete a staging workflow associated to a given project.
+XmlBody: delete_staging_workflow
```
dmarcoux, line 1956: This is wrong. This endpoint doesn't need a body. However, there's the `with_staging_projects` option which isn't documented.

krauselukas:
@quatran Could you please add a little description of what and why you did those changes to the commit message. Also reference the issue in the commit message as well :)

docs/api/api/api.txt
dmarcoux: I looked at the other `DELETE` endpoints and how we describe them. We don't refer to a `XmlBody` since it's not needed. We could be clearer by writing something like `XmlBody: none` or `XmlBody: empty`, but if we go ahead with this change, we need to be consistent and do it for all `DELETE` endpoints.

dmarcoux: For an example of how we describe parameters, see [this](https://github.com/openSUSE/open-build-service/blob/c6b95b130c0f2032393a324cd10a147062f3292a/docs/api/api/api.txt#L1677)

dmarcoux:
So the issue #8117 has already been fixed in my previous PR #8485. I forgot to link the issue in my commit message at the time, so this explains the confusion now.

In the end, documenting the `with_staging_projects` parameter is all that we need to do for the `DELETE /staging/<staging_workflow_project>/workflow` endpoint.

docs/api/api/api.txt
```diff
@@ -1953,6 +1953,11 @@ XmlResult: status_ok
 
 DELETE /staging/<staging_workflow_project>/workflow
   Delete a staging workflow associated to a given project.
+
+Parameters: 
+  with_staging_projects
+
+XmlBody: none
```
dmarcoux, line 1960: I don't think we should do this. Otherwise, as I said in my previous comments, we need to be consistent and do it for all `DELETE` endpoints.

quatran: ok. i will add 'XmlBody: none' to the all other 'DELETE' endpoints.

docs/api/api/api.txt
```diff
@@ -1953,6 +1953,11 @@ XmlResult: status_ok
 
 DELETE /staging/<staging_workflow_project>/workflow
   Delete a staging workflow associated to a given project.
+
+Parameters: 
+  with_staging_projects
```
dmarcoux, line 1958: The description is missing.

PR #9489, opened 2020-05-04
[backend] ccache archive no longer has .gz suffix

sjamgade:
drop the suffix as the archive is no longer compressed but compression
is handled by ccache itself


sjamgade:
PR for build script. https://github.com/openSUSE/obs-build/pull/560

PR #9490, opened 2020-05-04
Fix notification creation rake task for development

krauselukas:
- Fixes the following:

* When the admin home project does not exist, calling the
  'create_and_assign_project' with 'admin.home_project' fails
  since it is nil.
* The 'create_and_assign_project' returned the relationship,
  not the project

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -193,7 +193,7 @@ namespace :dev do
       User.session = requestor
 
       # Projects
-      admin_home_project = admin.home_project || create_and_assign_project(admin.home_project, admin)
+      admin_home_project = admin.home_project || create_and_assign_project(admin.home_project_name, admin)
       requestor_project = Project.find_by(name: 'requestor_project') || create_and_assign_project('requestor_project', requestor)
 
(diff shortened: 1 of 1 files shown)


src/api/lib/tasks/dev.rake
```diff
@@ -416,8 +416,9 @@ def request_for_staging(staging_project, maintainer_project, suffix)
 end
 
 def create_and_assign_project(project_name, user)
-  create(:project, name: project_name)
-  create(:relationship, project: project_name, user: user, role: Role.hashed['maintainer'])
+  project = create(:project, name: project_name)
```
dmarcoux, line 419: This isn't a blocker, but it could be rewritten with [tap](https://ruby-doc.org/core-2.7.1/Object.html#method-i-tap):

```
create(:project, name: project_name).tap do |project|
  create(:relationship, project: project, user: user, role: Role.hashed['maintainer'])
end
```

krauselukas: @dmarcoux thats nice, thanks, haven't seen the ```tap``` yet :)

krauselukas: @dmarcoux applied

PR #9495, opened 2020-05-05
Move repositories actions to Actions menu

saraycp:
Under beta program, the general actions related to repositories are
moved to the Actions menu.

When not in beta, the actions are displayed in the same place, however
the code has been moved to a partial for simplicity.


saraycp:
@DavidKang, thanks. Fixed. Could you have another look?

saraycp:
@dmarcoux, partial renamed!

dmarcoux:
The top card (with the tabs) is now quite empty with the actions moved in the `Actions` menu. I suggest combining the first card with the second. So the repositories defined in the second card would be in the first card.

suggestion

saraycp:
> The top card (with the tabs) is now quite empty with the actions moved in the `Actions` menu. I suggest combining the first card with the second. So the repositories defined in the second card would be in the first card. It would also remove the repetition of `Repositories for my_project`, which doesn't bring any value.

@dmarcoux now the two sections/cards are combined, as you suggested. I decided to modify this only for beta users.

Since this changes increased the code complexity I finally refactored the view code. There is a new commit for that, which can be squashed once it is approved.

PR #9496, opened 2020-05-05
Move attribute actions to Actions menu

saraycp:
Under beta program, the general actions related to attribute are
moved to the Actions menu.

### DIFF
--- a/src/api/app/views/webui/attribute/responsive_ux/_index_actions.html.haml
+++ b/src/api/app/views/webui/attribute/responsive_ux/_index_actions.html.haml
@@ -1,5 +1,5 @@
 - content_for :actions do
   %li.nav-item
-  = link_to(new_attribs_path(project: project.name, package: package), title: 'Add a new attribute') do
+  = link_to(new_attribs_path(project: project.name, package: package), title: 'Add Attribute') do
     %i.fas.fa-plus-circle.fa-lg.mr-2
-    Add a new attribute
+    Add Attribute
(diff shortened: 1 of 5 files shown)


eduardoj:
There are some attributes that can be edited, like "OBS:Issues" or "OBS:RejectRequests". So we can "Add an issue", or "Add a value". Should we also add these actions to the Actions menu, or not?

Editing OBS:Issues now:

Screenshot from 2020-05-06 08-56-47

saraycp:
> There are some attributes that can be edited, like "OBS:Issues" or "OBS:RejectRequests". So we can "Add an issue", or "Add a value". Should we also add these actions to the Actions menu, or not?

I think that view is going to need a bit of refactoring, since it's using coccon's `link_to_add_association` associating the form and the link.
I propose to do it in a different PR.

saraycp:
@dmarcoux, partial renamed!

src/api/app/views/webui/attribute/responsive_ux/_index_actions.html.haml
```diff
@@ -0,0 +1,5 @@
+- content_for :actions do
+  %li.nav-item
+  = link_to(new_attribs_path(project: project.name, package: package), title: 'Add attribute') do
+    %i.fas.fa-plus-circle.fa-lg.mr-2
+    Add attribute
```
dmarcoux, line 5: `attribute` should be capitalized, so it would be `Add Attribute` (same for the title on line 3).

dmarcoux:
There's a typo in the last commit message: attribu**r**e

src/api/app/views/webui/attribute/responsive_ux/_index_actions.html.haml
saraycp: True! Updated.

PR #9501, opened 2020-05-06
[backend] add withccache option to getbinarylist call

sjamgade:
this option is forwarded to repserver which does the main job

### DIFF
--- a/src/backend/bs_srcserver
+++ b/src/backend/bs_srcserver
@@ -4083,7 +4083,7 @@ sub getbinarylist {
   my $view = $cgi->{'view'};
   my $nosource = $cgi->{'nosource'};
   my $reposerver = $BSConfig::partitioning ? BSSrcServer::Partition::projid2reposerver($projid) : $BSConfig::reposerver;
-  my @args = BSRPC::args($cgi, 'view', 'nosource', 'withmd5', 'binary', 'module');
+  my @args = BSRPC::args($cgi, 'view', 'nosource', 'withmd5', 'binary', 'module', 'withccache');
   if ($view && ($view eq 'cache' || $view eq 'cpio' || $view eq 'solv' || $view eq 'solvstate')) {
     # do not check arch in interconnect mode
     my $proj = checkprojrepoarch($projid, $repoid, undef, 1);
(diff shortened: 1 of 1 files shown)


sjamgade:
I am not very sure I have covered all the cases (with all possible values for various queryparams and their codepaths). But did do some basic testing with `osc`

mlschroe:
I'm not really convinced that osc should use this, as old build service versions do not support the withccache parameter. I'd prefer a second request that just tries to get the ccache file.

But that's something the osc people should decide. Marco, any thoughts?

mlschroe:
I'm not convinced osc should use this, as old obs versions do that understand that parameter. I think it would be better to get the ccache file with a second request.

Also note that this should IMHO not be tied to --oldpackages in osc.

(And the API needs to forward the new parameter, but I have not checked this.)

But that's something the osc folks need to decide. Marco, any thoughts?

mlschroe:
(Bah, sorry about the duplicate comment. github seems to be not up to its usual self.)

mlschroe:
Ok, I somewhat misunderstood the intent: I thought this was for using ccache in osc build, bit it was done for osc getbinaries. (But who would want to get the ccache with the rpms?)

sjamgade:
Sorry, I will word my commit message better next time.
This is just to allow downloading ccache. 
the OSC's pr adds `--withccache` param to `osc getbinaries ...` so the archive is downloaded only when `--withccache` option is preset. 

The idea is avoid having a extra route just for ccache. Also because it is more like a buildartifact, just like rpm.  We could do have an extra route just like `jobhistory,buildlog...` but those arent build artifacts.

mlschroe:
But you do not need a new route, you can simply use the "getbinary" route to get the ccache tar ball.

sjamgade:
osc is using `getbinary` to download, this patch allow listing of ccache archive. which osc does to get a list of binaries to download.

lethliel:
The osc part is fine and can be merged after this PR is merged. 

Any objections merging this @mlschroe

lethliel:
merged as discussed in Stand-Up

PR #9506, opened 2020-05-06
ConsistencyCheck job refactoring

vpereira:
To avoid breaking the code which is running today and at the same time improve the code, we decided to move the old `ConsistencyCheck` under a different namespace, call it via `rake` as usual, avoiding touch the old code.

To be able to refactor the `ConsistencyCheck` class we broke the job in services and are adding specs for that as necessary.

As soon as the new code is working as it should, we can drop the old code.


vpereira:
Please do not merge it. It's a WIP

src/api/app/models/project.rb
```diff
@@ -103,7 +103,10 @@ def staging_history
   scope :filtered_for_list, lambda {
     where.not('name rlike ?', ::Configuration.unlisted_projects_filter) if ::Configuration.unlisted_projects_filter.present?
   }
+
   scope :remote, -> { where('NOT ISNULL(projects.remoteurl)') }
+  scope :local, -> { where.not('NOT ISNULL(projects.remoteurl)') }
```
krauselukas, line 108:
```suggestion
  scope :local, -> { where('ISNULL(projects.remoteurl)') }
```

src/api/app/services/consistency_check_job_service/backend_package_importer.rb
```diff
@@ -0,0 +1,39 @@
...
+      @errors << delete_error_message
+    end
+
+    private
+
+    def delete_error_message
+      "DELETED in backend due to invalid data #{@project.name}/#{@package.name}"
+    end
+
+    def create_package_frontend
```
krauselukas, line 25: @vpereira Just a suggestion, I would name it `create_package_in_frontend`, makes it more clear on the first read

src/api/app/services/consistency_check_job_service/backend_project_importer.rb
```diff
@@ -0,0 +1,37 @@
...
+    rescue ActiveRecord::RecordInvalid
+      delete_source
+      @errors << "DELETED #{@project} on backend due to invalid data"
+    rescue Backend::NotFoundError
+      @errors << "specified #{@project} does not exist on backend"
+    end
+
+    private
+
+    def create_project_frontend
```
krauselukas, line 23: same here `create_project_in_frontend`, but like I already said, only a suggestion :)

PR #9509, opened 2020-05-06
Move package branch dialog to dedicated view

krauselukas:
The package branch dialog deserves it's own view, since
there are to quite some options which lead to a pretty
packed and unclear modal.


krauselukas:
[[M] Review Modals for Package#show](https://trello.com/c/bMLbySQv/1401-m-review-modals-for-packageshow)

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -373,6 +373,11 @@ def branch_diff_info
     }
   end
 
+  def branch_dialog
```
dmarcoux, line 376: We don't have a modal/dialog anymore, so this action name is misleading. I suggest naming this action `branch` and renaming the `branch` action below to `create_branch` (maybe we can find a better name). Then we need to adapt the routes a bit.

We can either have one route `branch` with the `GET` going to `package#branch` and the `POST` going to `package#create_branch`. This is what Rails does for `index` and `create`. They both point to the same route, but depending on the HTTP verb, the action varies.

The other option is to have a separate route, so two routes just like the actions, so `GET branch` and `POST create_branch`.

damianatorrpm:
If it is on it's own page maybe 'more options' should be expanded.

krauselukas:
@hennevogel I brought back the ```More options``` area, I agree that this brings in unnecessary complexity that many people don't need. I reworked a bit the hierarchy on it, added some padding and removed the strong tags from the form labels.
@damianatorrpm moving the things to the right side doesn't really bring an advantage from my point of view. On desktop sized screens we have both a lot of space on the right and on the bottom.

krauselukas:
I would ignore codeclimate, the problems are not introduced by this PR (only gets triggered because I renamed a method) and fixing it is a big refactor.

src/api/app/views/webui/package/branch.html.haml
```diff
@@ -0,0 +1,28 @@
+:ruby
+  @pagetitle = "Branch #{@package}"
+
+.card.mb-3
```
hennevogel, line 4: Inside containers you normally use [the grid](https://getbootstrap.com/docs/4.4/layout/grid/) to place things, not cards.

src/api/app/views/webui/package/branch.html.haml
```diff
@@ -0,0 +1,28 @@
...
+      %p
+        -# FIXME: get rid of this helper
+        = project_or_package_link(project: branch_project, short: true)
+      = form_tag({ controller: :package, action: :create_branch }, method: :post) do
+        %details
+          %summary More options
+          .p-2
+            = hidden_field_tag :linked_project, @project.name
+            = hidden_field_tag :linked_package, @package.name
+            = render partial: 'webui/shared/package_branch_form', locals: { show_project_field: true,
```
hennevogel, line 23: Using the grid also will make it possible to avoid overly long input fields.

src/api/app/views/webui/package/branch.html.haml
krauselukas: @hennevogel I applied the grid system, but I kept the card. We use them almost everywhere. Not sure if I understood correctly, do you want to get rid of the way we use the cards to structure the view or was it more about the missing grid?

src/api/app/views/webui/package/branch.html.haml
krauselukas: It's fixed, the applied grid keeps the input fields in place :)

src/api/app/views/webui/package/branch.html.haml
hennevogel: You now have a wild mixture of card and grid. The grid is for the layout, the card is styling the content. Build a grid, fill it with content.

vpereira: @hennevogel I see that in many PRs we are going with `card.mb-3` and you are against it. Could you please explain us how you want it to avoid everyone doing their own interpretation of what you are saying?

hennevogel: we talked about this offline. grid first, then put content onto it is our new motto! :smile:

dmarcoux:
@hennevogel Please have another look if changes are still needed.

PR #9512, opened 2020-05-07
Implement the in-place package editing

danidoni:
This PR changes the existing modal to change the description and title
of the package to an in-place editing feature, like the one found in
Project.

**Before**
image

**After**
image

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -5,13 +5,13 @@ class Webui::PackageController < Webui::WebuiController
   include Webui::ManageRelationships
   include BuildLogSupport
 
-  before_action :set_project, only: [:show, :index, :users, :dependency, :binary, :binaries, :requests, :statistics, :revisions,
+  before_action :set_project, only: [:show, :edit, :update, :index, :users, :dependency, :binary, :binaries, :requests, :statistics, :revisions,
                                      :new, :branch_diff_info, :rdiff, :create, :save, :remove, :add_file, :save_file,
                                      :remove_file, :save_person, :save_group, :remove_role, :view_file, :abort_build, :trigger_rebuild,
                                      :trigger_services, :wipe_binaries, :buildresult, :rpmlint_result, :rpmlint_log, :meta, :save_meta, :files,
                                      :binary_download]
 
-  before_action :require_package, only: [:show, :dependency, :binary, :binaries, :requests, :statistics, :revisions,
+  before_action :require_package, only: [:edit, :update, :show, :dependency, :binary, :binaries, :requests, :statistics, :revisions,
                                          :branch_diff_info, :rdiff, :save, :save_meta, :remove, :add_file, :save_file,
                                          :remove_file, :save_person, :save_group, :remove_role, :view_file, :abort_build, :trigger_rebuild,
                                          :trigger_services, :wipe_binaries, :buildresult, :rpmlint_result, :rpmlint_log, :meta, :files, :users,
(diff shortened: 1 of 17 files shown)


danidoni:
[[M] In place editing for basic Package information](https://trello.com/c/IW4ak8d6/1396-m-in-place-editing-for-basic-package-information)

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -236,7 +236,7 @@ def require_package
     return if params[:package].blank?
 
     begin
-      @package = Package.get_by_project_and_name(@project.to_param, params[:package],
+      @package = Package.get_by_project_and_name(@project.to_param, params[:package_name] || params[:package],
```
danidoni, line 239: I feel this change is not right... I'm trying to come up with an alternative, because the form is settings the package parameters in the `:package` param, not the package name.

src/api/config/routes/webui_routes.rb
```diff
@@ -51,9 +51,14 @@
 
     resources :package, only: [:index], controller: 'webui/package', constraints: cons
 
+    controller 'webui/package' do
+      get 'package/edit/:project/:package' => :edit, constraints: cons, as: 'edit_package'
+      patch 'package/update' => :update, constraints: cons
```
vpereira, line 56: would be possible to make it more resourceful and bring it inside of project?

danidoni: I've been looking into this and before that we need to refactor a lot of existing routes. For now I'll let them as they are

src/api/app/controllers/webui/webui_controller.rb
danidoni: Fixed

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -849,6 +881,19 @@ def package_params
     params.require(:package).permit(:name, :title, :description)
   end
 
+  def package_details_params
```
dmarcoux, line 884: This is a code smell to me, having a second strong parameters method. We should only have `package_params`.

src/api/app/views/webui/package/responsive_ux/actions/_edit_package.html.haml
```diff
@@ -0,0 +1,4 @@
+%li.nav-item
+  = link_to(edit_package_path(project, package), class: 'nav-link', remote: true) do
+    %i.fas.fa-edit.fa-lg.mr-1
+    Edit package
```
dmarcoux, line 4:
```suggestion
    Edit Package
```

dmarcoux:
The commits should also be squashed.

src/api/app/controllers/webui/package_controller.rb
danidoni: I've explained why I opted to have a second parameter method below. I know its a smell, but if I touch this now, this PR will go wildly out of scope.

danidoni: I can refactor this in another PR

danidoni:
> The commits should also be squashed.

Done

src/api/app/controllers/webui/package_controller.rb
dmarcoux: Sure

danidoni:
> Validation errors are not displayed in the form. For example, when I add a very long title nothing happens.

I've been looking into this and to fix it I have to also do the refactoring @dmarcoux was asking for, so... Let's go refactoring :D

danidoni:
> > Validation errors are not displayed in the form. For example, when I add a very long title nothing happens.
> 
> I've been looking into this and to fix it I have to also do the refactoring @dmarcoux was asking for, so... Let's go refactoring :D

I'll push a quick & dirty fix and tackle the refactoring in another PR.

saraycp:
> I'll push a quick & dirty fix and tackle the refactoring in another PR.

Ok!

PR #9517, opened 2020-05-07
Introduce feature beta tests

eduardoj:
We want to run tests in the beta environment. For that we define a new subdirectory where every test located there will be executed in the beta environment: `spec/features/beta/`.

We also added an extra CircleCI container for running the beta feature tests in our CI environment.

We adapted all the feature tests to pass within the beta environment.


hennevogel:
hm, how do you keep this in sync?

src/api/spec/support/shared_examples/features/bootstrap_user_tab.rb
```diff
@@ -52,7 +52,12 @@ def toggle_checkbox(html_id)
     end
 
     scenario 'Add non existent user' do
-      click_link('Add User')
+      # TODO: Remove once responsive_ux is out of beta.
```
hennevogel, line 55: Why don't you copy this shared example into a beta sub directory too?

hennevogel:
> hm, how do you keep this in sync?

@eduardoj and I talked about this offline and I understand now :smile:

src/api/spec/support/shared_examples/features/bootstrap_user_tab.rb
eduardoj: Copied an adapted to use `click_menu_link` method calls only. Please have a look.

src/api/spec/features/beta/webui/attributes_spec.rb
```diff
@@ -64,12 +64,14 @@
...
 
           visit index_attribs_path(project: user.home_project_name)
-          expect(page).not_to have_content('Add a new attribute')
+          expect(page).not_to have_content('Add Attribute')
         end
       end
 
       context 'with permissions' do
         scenario 'remove attribute' do
+          skip('Remove dialog is not working')
```
hennevogel, line 73: I don't get it. Why skip it instead of fixing it?

hennevogel: Or at least have a FIXME and an issue

src/api/spec/features/beta/webui/login_spec.rb
```diff
@@ -7,17 +7,19 @@
 
   scenario 'login with home project shows a link to it' do
     login user
-    within('#personal-navigation') do
-      expect(page).to have_link('Home Project')
+    click_link('Places')
+    within('#navigation') do
+      expect(page).to have_link('Your Home Project')
```
hennevogel, line 12: Why not do the same as in 5eede45cd8e3be7e434e3ec066916018c49100c0

src/api/spec/features/beta/webui/attributes_spec.rb
DavidKang: The fix was merged. I forgot to remove this. Thanks.

src/api/spec/features/beta/webui/login_spec.rb
DavidKang: Good catch!

DavidKang: Done.

src/api/spec/features/beta/webui/attributes_spec.rb
DavidKang: Removed.

.circleci/config.yml
```diff
@@ -105,9 +105,15 @@ jobs:
             cd src/api
             pickfile=log/pick.$CIRCLE_NODE_INDEX.list
 
-            # single out db specs
-            circleci tests glob 'spec/**/*_spec.rb' | grep -v 'spec/db/' > spec.list
-            circleci tests split --total $CIRCLE_NODE_TOTAL --split-by=timings < spec.list > $pickfile
+            # reserve the latest node for beta
```
dmarcoux, line 108: It is also for `db` specs as stated on line 113.

eduardoj: In this `if` statement we use the last container to run only the beta tests. In the `else` section we exclude both, `beta` and `db` specs.

The `db` specs are run in container 0, see line 122.

PR #9522, opened 2020-05-08
Hide build results for delete requests

### LINKED ISSUE #9514: [production][package#buildresult] undefined method `buildresult' for nil:NilClass
package#buildresult threw this 54 times
<pre>
undefined method `buildresult&#39; for nil:NilClass
</pre>
[https://build.opensuse.org/package/buildresult?project=openSUSE%3AFactory%3AStaging%3Aadi%3A41&amp;package=&amp;index=0&amp;show_all=false](https://build.opensuse.org/package/buildresult?project=openSUSE%3AFactory%3AStaging%3Aadi%3A41&amp;package=&amp;index=0&amp;show_all=false)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5eb00e00f48ca5000dee35da)

saraycp:
When we are in a request for package deletion, it makes no sense to
build the package which no longer exists.

Using 'defined?' to ckeck if the project or package exist is not a good
idea because it always returns truthy (returns "local-variable") even if they are nil.
Therefore, 'defined?' has been replaced by '||='.

Even if it is not going to display build results, just the text "No build results available" I decided to give it the same aspect: a card with the `Build Results` tab inside.

Fixes #9514 

Screenshot_2020-05-08 Request 385 (review)(1)

### DIFF
--- a/src/api/app/views/webui/shared/_buildresult_box.html.haml
+++ b/src/api/app/views/webui/shared/_buildresult_box.html.haml
@@ -1,51 +1,68 @@
 :ruby
   index ||= ''
+  project ||= nil
+  package ||= nil
+  bs_request ||= nil
+  buildable = package || project
 
-  # For a request staged in a staging project, we display the Build Results / RPM Lint from the staging project instead
-  is_staged_request = defined?(bs_request) && !bs_request.staging_project_id.nil?
-  project = bs_request.staging_project.name if is_staged_request
+- if buildable
+  :ruby
+    # For a request staged in a staging project, we display the Build Results / RPM Lint from the staging project instead
+    is_staged_request = bs_request && !bs_request.staging_project_id.nil?
+    project = bs_request.staging_project.name if is_staged_request
 
-  ajax_data = {}
-  ajax_data['project'] = h(project) if defined?(project)
-  ajax_data['package'] = h(package) if defined?(package)
-  ajax_data['index'] = h(index) if defined?(index)
+    ajax_data = {}
+    ajax_data['project'] = h(project) if project
+    ajax_data['package'] = h(package) if package
+    ajax_data['index'] = h(index) if index
 
-.card
-  .bg-light{ id: "buildresult#{index}-urls", data: { buildresult_url: defined?(package) ? package_buildresult_path : project_buildresult_path } }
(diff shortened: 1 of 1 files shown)


src/api/app/views/webui/shared/_buildresult_box.html.haml
```diff
@@ -1,51 +1,68 @@
...
   index ||= ''
+  project ||= nil
+  package ||= nil
+  bs_request ||= nil
+  buildable = package || project
 
-  # For a request staged in a staging project, we display the Build Results / RPM Lint from the staging project instead
-  is_staged_request = defined?(bs_request) && !bs_request.staging_project_id.nil?
-  project = bs_request.staging_project.name if is_staged_request
+  if buildable
```
krauselukas, line 8: @saraycp I think you can move the content of this if block inside the next one in line 19

src/api/app/views/webui/shared/_buildresult_box.html.haml
```diff
@@ -1,51 +1,69 @@
...
+    # For a request staged in a staging project, we display the Build Results / RPM Lint from the staging project instead
+    is_staged_request = bs_request && !bs_request.staging_project_id.nil?
+    project = bs_request.staging_project.name if is_staged_request
 
-  ajax_data = {}
-  ajax_data['project'] = h(project) if defined?(project)
-  ajax_data['package'] = h(package) if defined?(package)
-  ajax_data['index'] = h(index) if defined?(index)
+    ajax_data = {}
+    ajax_data['project'] = h(project) if project
```
danidoni, line 14: :thought_balloon: All this logic can go into a Presenter, doing so, you remove all the logic from the view

src/api/app/views/webui/shared/_buildresult_box.html.haml
saraycp: @krauselukas updated!

src/api/app/views/webui/shared/_buildresult_box.html.haml
saraycp: I wouldn't even call this view if the package or project are missing. We can handle more things from the controller...

In fact, I started to refactor this code but it was taking a lot of time. So I decided to simplify and just solve the issue.

vpereira: in P2 we first try to solve the problem.

PR #9534, opened 2020-05-11
Changes to obsworker script to fix build with zVM worker 

azouhr:
Hi,

this change is to fix the creation of mkinitrd within the obsworker script when z/VM is used as build worker. I also added a variable ZVM_VOLUME_SWAP (also already used within build-vm-zvm) to make it easier to customize the worker for different z/VM setups.

In addition, I found that you have to tell z/VM to use IUCV to the local guest for console messages. Else, the send commands used in build-vm-zvm will only work as long as some real console is attached.


adrianschroeter:
can you squash the commits to one please?

azouhr:
pull request is broken, I will close it and try to do it right.

PR #9536, opened 2020-05-11
Move project release request modal to its own page

dmarcoux:
preview


dmarcoux:

src/api/app/views/webui/project/release_request.html.haml
```diff
@@ -0,0 +1,22 @@
+- @pagetitle = 'Create Maintenance Release Request'
+
+.card.mb-3
```
hennevogel, line 3: We should not use `card` outside of the grid. The grid is for laying out the page. Cards are for styling content. Build the grid, put the content in it.

dmarcoux: So I've came to the realization that we do misuse the `card` component. It's true that we should build the layout with a grid/flexbox. The card component is a container, but it's not meant to be used to create a layout. It's for content and it's pretty clear when looking at the card component documentation in https://getbootstrap.com/docs/4.4/components/card/#content-types

I guess I just didn't know much about the grid/flexbox whenever we started the Bootstrap migration. Then never went back to check what the card component actually is.

dmarcoux: For now, we'll create the new views with the grid and eventually, we'll have to refactor our views to stop misusing the card component.

PR #9538, opened 2020-05-11
WIP: Status message and Announcement refactoring

vpereira:
### Background
StatusMessage and Announcements could be represented as the same thing: A communication channel between Administrators and Users. What differ them is just one thing: The Severity.

### TODO:

- [x] Drop title from `Announcement`
- [x] Rename content field to message
- [x] Adapt `StatusMessage` model to support announcement severity
- [x] Make the `StatusMessageController#create` support Announcement creation too (prototype)
- [x] Create `StatusMessageController#new`, add route and initial view
- [x] Add support to scope (group of users which will see the announcement)
- [ ] Update Annotation schema on changed model
- [ ] Move view from modal dialog to regular view
- [ ] Remove the `safety_assured` from the migration
- [ ] Add a data migration to handle old Announcements
- [x] Adapt XML API(endpoints and schemas)
- [ ] Refactor prototype controller code
- [x] Fix specs
- [ ] Adapt https://github.com/openSUSE/open-build-service/wiki/Announcements

### How to test it:

Log as Admin, click on "Create StatusMessage". If the message has severity = `announcement` then the message will be created as `Announcement`

You can test it going straight to `/status_messages/new`, the view still not completely done and not linked anywhere

### Out of scope:

- Support multiple Announcements at the same time

## Functionality:
status_message_to_announcement


vpereira:
[[L] As an admin I want to be able to announce something inside the app only for beta users](https://trello.com/c/TjHv1Ehz/1409-l-as-an-admin-i-want-to-be-able-to-announce-something-inside-the-app-only-for-beta-users)

PR #9550, opened 2020-05-14
Do not blindly read an uploaded file into memory

marcus-h:
Currently, when uploading a file, its entire content is read into
memory and then passed to Package.verify_file! (even if verify_file!
never uses the content). Instead of always reading the file into
memory, it should be only read if its content is used in
Package.verify_file!.
For this, we now pass a "readable" object to Package.verify_file!.
The "readable" object defers the call to request.raw_post.to_s until
its "read" method is called. Moreover, Package.verify_file! only
calls the "read" method if it is going to use the uploaded file's
content.

Note: it is still the case that large files are read into memory.


vpereira:
Hi @marcus-h, thank you for your PR. I will take a look on that tomorrow-

marcus-h:
On 2020-05-14 22:54:15 -0700, Victor Pereira wrote:
> @vpereira requested changes on this pull request.
>
> Probably calling `content.to_io` would be better than pass it to a proc in the controller?
>
Hmm what is "content" in this case? (What do we pass to the
Package.verify_file! method?) If there is a more rails/ruby-ish way, I'm
all for it:)

vpereira:
> Hmm what is "content" in this case? (What do we pass to the Package.verify_file! method?) If there is a more rails/ruby-ish way, I'm all for it:)


Hi, it is an ```ActionDispatch::Http::UploadedFile```

You can create one like 

```foo = ActionDispatch::Http::UploadedFile.new(filename: 'foo.jpg', type: 'image/jpeg', tempfile: Tempfile.new)```

and, ```foo.to_io``` will return an [IO](https://ruby-doc.org/core-2.7.1/IO.html)

**Comment by marcus-h** [comment]
On 2020-05-15 02:24:30 -0700, Victor Pereira wrote:
> > Hmm what is "content" in this case? (What do we pass to the Package.verify_file! method?) If there is a more rails/ruby-ish way, I'm all for it:)
>
>
> Hi, it is an ```ActionDispatch::Http::UploadedFile```
>
> You can create one like
>
> ```foo = ActionDispatch::Http::UploadedFile.new(filename: 'foo.jpg', type: 'image/jpeg', tempfile: Tempfile.new)```
>
> and, ```foo.to_io``` will return an [IO](https://ruby-doc.org/core-2.7.1/IO.html)
>
Hmm I'm not sure if I got you right but doesn't this add a lot more
complexity to the controller?

tmp = Tempfile.new
body_stream = request.body
IO.copy_stream(body_stream, tmp)
tmp.rewind
content = ActionDispatch::Http::UploadedFile.new(tempfile: tmp)
Package.verify_file(..., content)
# if no rewind is supported we are screwed up... :/
body_stream.rewind if body_stream.respond_to?(:rewind)

If something like that is preferred, I would suggest something like

body_stream = request.body
Package.verify_file(..., body_stream)
# if no rewind is supported we are screwed up... :/
body_stream.rewind if body_stream.respond_to?(:rewind)

The issue with this code is that it relies on the presence of a "rewind"
method (the rewind is crucial as otherwise we would send the "wrong" data
to the backend (see download_request in the application_controller.rb)).
The "advantage" of the request.raw_post approach is that the code in
ActionDispatch::Request takes care of the absence of a "rewind" method.

**Comment by vpereira** [comment]
> Hmm I'm not sure if I got you right but doesn't this add a lot more
complexity to the controller?

@marcus-h 

No, note that we have already the object `ActionDispatch::Http::UploadedFile` in the method `Package.verify_file` so just calling `to_io` instead of pass it's path to `File.open` and call `read` would actually simply the code, no?

**Comment by marcus-h** [comment]
On 2020-05-15 08:51:48 -0700, Victor Pereira wrote:
> > Hmm I'm not sure if I got you right but doesn't this add a lot more
> complexity to the controller?
>
> @marcus-h
>
> No, note that we have already the object `ActionDispatch::Http::UploadedFile` in the method `Package.verify_file` so just calling `to_io` instead of pass it's path to `File.open` and call `read` would actually simply the code, no?
>
Ah now I got it (maybe;) )! I think we are talking about different cases:)

Case 1: the uploaded file is part of a request with content-type
        multipart/form-data.

This case occurs if a file is uploaded via the webui. While processing/parsing
the multipart/form-data request, an ActionDispatch::Http::UploadedFile instance
is created, which is eventually passed to Package.verify_file! (caller:
Package.save_file).


Case 2: else

This case can occur if a file is uploaded, for instance, via curl:
curl -u Admin:opensuse \
    http://localhost:3000/source/home:Admin/foo/somefile \
    -X PUT -T somefile -H "Content-Type: application/octet-stream"

That is, Package.verify_file! is called from SourceController.update_file
where we have no ActionDispatch::Http::UploadedFile instance available.


In Case 1, everything is fine (we could do your "to_io" change but then we
also have to call "rewind" afterwards).

Does this make sense?:)

**Comment by vpereira** [comment]
> That is, Package.verify_file! is called from SourceController.update_file where we have no ActionDispatch::Http::UploadedFile instance available.

As I see, scenario 2,  we are handling a string which could be handled with https://ruby-doc.org/stdlib-2.5.1/libdoc/stringio/rdoc/StringIO.html and then we have in both cases, an IO therefore we could handle both scenarios in the same way... what do you think @marcus-h ?

**Comment by marcus-h** [comment]
On 2020-05-16 02:54:43 -0700, Victor Pereira wrote:
> > That is, Package.verify_file! is called from SourceController.update_file where we have no ActionDispatch::Http::UploadedFile instance available.
>
> As I see, scenario 2,  we are handling a string which could be handled with https://ruby-doc.org/stdlib-2.5.1/libdoc/stringio/rdoc/StringIO.html and then we have in both cases, an IO therefore we could handle both scenarios in the same way... what do you think @marcus-h ?
>
Hmm we cannot use a StringIO because we would need the string at initialization
time. That is, we would always have to call "request.raw_post". This defeats
the purpose of my patch.
What my patch does is the following: it defers the call to "request.raw_post"
to the moment where the "read" method is called on the Proc (and the "read"
method is only called if we really need to verify the content).

**Comment by vpereira** [comment]
@marcus-h looking again this code, instead of call `to_s` in the http request body and wrap it in a callable object to be consumed later, we could pass the httprequest around, no https://apidock.com/rails/v5.2.3/ActionDispatch/Request/body ? It is already an IO, so IMO is the same  as your proc { "string" }, right?

**Comment by marcus-h** [comment]
On 2020-05-19 06:30:51 -0700, Victor Pereira wrote:
> @marcus-h looking again this code, instead of call `to_s` in the http request body and wrap it in a callable object to be consumed later, we could pass the httprequest around, no https://apidock.com/rails/v5.2.3/ActionDispatch/Request/body ? It is already an IO, so IMO is the same  as your proc { "string" }, right?
>
Yes, it will work as long as we are able to "rewind" the body afterwards:

body_io = request.body
Package.verify_file!(body_io)
body_io.rewind if body_io.respond_to(:rewind)  # could be part of Package.verify_file!

If the backing io class does not support the "rewind" method we are
screwed up. Though, I have no clue how likely/unlikely this is in the
future (note: I'm just saying this because the code in
ActionDispatch::Request::raw_post has a specific check for this (otherwise
I wouldn't care:) )).

**Comment by vpereira** [comment]
> Yes, it will work as long as we are able to "rewind" the body afterwards:

@marcus-h maybe it was stupid from my side, but looking the https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/request.rb#L324, if we don't call the `to_s` the `raw_post` is already good to go. So removing the `to_s` in the original code an slightly changing the verify_file we could fix it.. what do you think?

**Comment by marcus-h** [comment]
On 2020-05-19 09:21:46 -0700, Victor Pereira wrote:
> > Yes, it will work as long as we are able to "rewind" the body afterwards:
>
> @marcus-h maybe it was stupid from my side, but looking the https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/request.rb#L324, if we don't call the `to_s` the `raw_post` is already good to go. So removing the `to_s` in the original code an slightly changing the verify_file we could fix it.. what do you think?
>
Sounds good. I just removed "to_s" call (I wasn't sure if it was there
on purpose (maybe for some testing code/whatever), hence I kept it).

What do we have to change in Package.verify_file! ?

**Comment by marcus-h** [comment]
On 2020-05-20 01:04:19 -0700, Victor Pereira wrote:
> > @@ -361,7 +361,9 @@ def update_file
>        @pack.save
>      end
>
> -    Package.verify_file!(@pack, params[:filename], request.raw_post.to_s)
> +    readable = proc { request.raw_post }
>
> @marcus-h what I meant ist, the `request.raw_post` is already an readable.

No, it is a String (the contents of the uploaded file).

> So you could just pass it to the `verify_file!` method, instead of wrap it in a proc.
>
> >
>      # file is an ActionDispatch::Http::UploadedFile and Suse::Validator.validate
>      # will call to_s therefore we have to read the content first
> -    content = File.open(content.path).read if content.is_a?(ActionDispatch::Http::UploadedFile)
> +    content = File.open(content.path) if content.is_a?(ActionDispatch::Http::UploadedFile)
>
> not necessary to call `File.open` here.  https://api.rubyonrails.org/v6.0.2.1/classes/ActionDispatch/Http/UploadedFile.html has a method `open`,

We could do that but then we would need to "rewind" again because
ActionDispatch::Http::UploadedFile's "open" just delegates to the "open"
method of the underlying tempfile (Tempfile) and Tempfile.open exhibits
the following behavior:

marcus@linux:~> irb.ruby2.5
irb(main):001:0> require 'tempfile'
=> true
irb(main):002:0> tmp = Tempfile.new
=> #<Tempfile:/tmp/20200520-3083-xuitgj>
irb(main):003:0> tmp.write('foobar')
=> 6
irb(main):004:0> tmp.rewind
=> 0
irb(main):005:0> tmp2 = tmp.open
=> #<File:/tmp/20200520-3083-xuitgj>
irb(main):006:0> tmp.read(3)
=> "foo"
irb(main):007:0> tmp2.read(3)
=> "bar"
irb(main):008:0>
marcus@linux:~>

> but you still need to read it no?

It is read in the subsequent line:

+    content = content.read if content.respond_to?(:read)

**Comment by vpereira** [comment]
@marcus-h don't you think we have a similar problem here too https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/patchinfo_controller.rb#L46 ? We initialize a patchinfo class and then call the `to_xml` on that, ask if it is valid and then verify the content. The object is however already initialized and in the memory, then we verify the file.. we here at least use the file after that, but there is no size constrain on the content  too, right?

**Comment by marcus-h** [comment]
On 2020-05-20 06:01:36 -0700, Victor Pereira wrote:
> @marcus-h don't you think we have a similar problem here too https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/patchinfo_controller.rb#L46 ? We initialize a patchinfo class and then call the `to_xml` on that, ask if it is valid and then verify the content. The object is however already initialized and in the memory, then we verify the file.. we here at least use the file after that, but there is no size constrain on the content  too, right?
>
Yes. It is still possible that we read a _large_ file into memory (for
instance, an artificially large _patchinfo file).

My goal is just to avoid reading the file into memory if we are not
going to verify it...

**Comment by hennevogel** [comment]
Hey peepz, what's the status here?

**Comment by vpereira** [comment]
@hennevogel we have to rebase it , then it can be merged

**Comment by marcus-h** [comment]
On 2020-07-27 17:04:07 +0000, Victor Pereira wrote:
> @hennevogel we have to rebase it , then it can be merged
>
I just forced pushed a new (untested) version. Main changes:
- a second readable is used in the PackageService::FileVerifier
  class (this is needed because the PackageService::SchemaVerifier
  is called unconditionally)
- PackageService::FileVerifier.content_readable is renamed to
  PackageService::FileVerifier.content and only called if needed. Also,
  it now caches the content (to avoid, for instance, multiple reads).

**Comment by marcus-h** [comment]
On 2020-07-28 02:21:36 +0200, Marcus Hüwe wrote:
> On 2020-07-27 17:04:07 +0000, Victor Pereira wrote:
> > @hennevogel we have to rebase it , then it can be merged
> >
> I just forced pushed a new (untested) version. Main changes:
> - a second readable is used in the PackageService::FileVerifier
>   class (this is needed because the PackageService::SchemaVerifier
>   is called unconditionally)

I just pushed yet another version with s/readable/defer_content/ (in
this case readable was a misnomer because it had no "read" method:) ).

File: src/api/app/controllers/source_controller.rb [hunk 1]
```diff
@@ -371,7 +371,9 @@ def update_file
       end
     end
 
-    Package.verify_file!(@pack, params[:filename], request.raw_post)
+    readable = proc { request.raw_post }
+    readable.define_singleton_method(:read) { call }
+    Package.verify_file!(@pack, params[:filename], readable)
```
hennevogel, line 376: It's probably me being stupid but why do you add all this code instead of passing the request object into `Package.verify_file!`? The `request` object already supports all the things you want, no?

marcus-h:
On 2020-07-28 08:16:26 -0700, Henne Vogelsang wrote:
> > @@ -371,7 +371,9 @@ def update_file
>        end
>      end
>
> -    Package.verify_file!(@pack, params[:filename], request.raw_post)
> +    readable = proc { request.raw_post }
> +    readable.define_singleton_method(:read) { call }
> +    Package.verify_file!(@pack, params[:filename], readable)
>
> It's probably me being stupid but why do you add all this code instead of passing the request object into `Package.verify_file!`? The `request` object already supports all the things you want, no?
>
Yes, that's also possible. Instead of passing the "big" request object,
I used the readable approach for the following reasons:

- Package.verify_file! expects a "content"; IMHO, the request object is
  "more" than content
- "content" can also be an ActionDispatch::Http::UploadedFile instance
  whose backing file is opened and then read (see
  PackageService::FileVerifier); the read part could be shared with the
  readable approach from above
- calling "read" on some content sounds "natural"

That is, just personal taste and no technical reasons. If you prefer,
I'm perfectly fine with passing the request object to Package.verify_file!:)

hennevogel:
> Yes, that's also possible. Instead of passing the "big" request object,

Well you're not passing the object around. Just the reference to it right?

> - Package.verify_file! expects a "content"; IMHO, the request object is
>  "more" than content

It expects `content` because someone wrote it to expect `content`. If you change it to expect the request, it expects the request :smile: 

> - "content" can also be an ActionDispatch::Http::UploadedFile instance
>  whose backing file is opened and then read (see
>  PackageService::FileVerifier); the read part could be shared with the
>  readable approach from above

As far as I can see we basically have three cases.

1. The thing we are trying to save is an empty string
1. The thing we are trying to save is a string
1- The thing we are trying to save is an tempfile

- We know when we need to verify: Once the filename we are going to save to is in the (combined) list of file names to verify.
- If we need to verify we need the 'content', no matter what it is and where it comes from.

Sounds to me like the service should not decide if it should verify, the caller should do that. The service should also not decide how to "get" the content, the caller should do that.

hennevogel:
@marcus-h if you don't want to do all that now, I understand. Then let's push your version and leave this for later...

marcus-h:
On 2020-07-30 14:05:11 -0700, Henne Vogelsang wrote:
> > Yes, that's also possible. Instead of passing the "big" request object,
>
> Well you're not passing the object around. Just the reference to it right?
>
Yes, of course. Sorry for the confusion! What I meant was that the request
object has several methods, references etc., which could be "misused" by
future code. Or the other way around: if I read code like
Package.verify_file!(request), I would assume that verify_file! might do
more than just reading the content - but that's just personal taste:)

> > - Package.verify_file! expects a "content"; IMHO, the request object is
> >  "more" than content
>
> It expects `content` because someone wrote it to expect `content`. If you change it to expect the request, it expects the request :smile:
>
Simply renaming it to "request" is probably misleading from the other
callers' POV (which just pass content (ActionDispatch::Http::UploadedFile)).

> > - "content" can also be an ActionDispatch::Http::UploadedFile instance
> >  whose backing file is opened and then read (see
> >  PackageService::FileVerifier); the read part could be shared with the
> >  readable approach from above
>
> As far as I can see we basically have three cases.
>
> 1. The thing we are trying to save is an empty string
> 1. The thing we are trying to save is a string
> 1- The thing we are trying to save is an tempfile
>
In essence, yep.

> - We know when we need to verify: Once the filename we are going to save to is in the (combined) list of file names to verify.
> - If we need to verify we need the 'content', no matter what it is and where it comes from.
>
Yep.

> Sounds to me like the service should not decide if it should verify, the caller should do that. The service should also not decide how to "get" the content, the caller should do that.
>
That is, you are proposing something like this?

if Package.valid_filename!(params[:filename])
    and Package.content_validation?(@pack, params[:filename])
  Package.verify_file!(@pack, params[:filename], request.raw_post)
end

or

# this is the caller from Package.save_file:

def save_file(opt = {})
  ...
  if Package.valid_filename!(opt[:filename])
      and Package.content_validation?(self, opt[:filename])
    content = File.open(content.path).read if content.is_a?(ActionDispatch::Http::UploadedFile)
    Package.verify_file!(self, opt[:filename], content)
  end
  ...
end

where Package.valid_filename! raises an exception if the passed filename
is invalid and Package.content_validation? returns true if content
validation is needed.

hennevogel:
Yeah essentially. Not quite so python though :wink: 

```ruby
def save_file(opt = {})
  ...
  valid_filename!(opt[:filename])
  if content_needs_validation?(opt[:filename])
    content = File.open(content.path).read if content.is_a?(ActionDispatch::Http::UploadedFile)
    Package.verify_file!(self, opt[:filename], content)
  end
  ...
end

def valid_filename!
  ...
end
def content_needs_validation?
  ...
end
```

This would move a couple of things back to `Package` but it's way more readable IMHO.

For me all of this points to the fact that all of those validations belong to the `Backend::File` in the end...

PR #9549, opened 2020-05-14
Move 'Create Package' modal to its own page

dmarcoux:
With some refactoring... :)

Refer to the commit messages for details.

preview-view

preview-errors

PR #9551, opened 2020-05-14
Move 'Create Package' modal to its own page

dmarcoux:
Refactor the controllers/routes to make this RESTful and follow Rails conventions. The routes aren't using resources since it's too much work for now with the huge amount of routes in the package_controller.

Note: All `FIXME` in the `package_controller.rb` are going to be addressed in another PR. I split the changes in multiple PRs to make this easier to review.

preview-view


dmarcoux:

dmarcoux:
@vpereira If you want to review again, this is the first PR taken out of #9549. Others will follow once this once is approved and merged.

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -337,19 +338,23 @@ def rdiff
...
-    @package.description = params[:description]
-    if params[:source_protection]
-      @package.flags.build(flag: :sourceaccess, status: :disable)
-    end
-    if params[:disable_publishing]
-      @package.flags.build(flag: :publish, status: :disable)
-    end
+  def new
+    # FIXME: Use the package policy for this
+    authorize @project, :update?
```
dmarcoux, line 343: I use `authorize @project, :update?` since this is what we were using in the project_controller action `new_package`.

vpereira: I will just look into the `sourceaccess` thing.. We have a `source_access?` policy which isn't included in `update?`, but TBH update looks the right one to me too

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -991,23 +1000,24 @@ def set_remote_linkinfo
...
+    if Package.exists_by_project_and_name(@project.name, package_name)
+      flash[:error] = "Package '#{package_name}' already exists in project '#{@project}'"
+      redirect_to action: :new, project: @project
       return false
     end
+    # FIXME: This should be a Pundit policy
     unless User.possibly_nobody.can_create_package_in?(@project)
       flash[:error] = "You can't create packages in #{@project.name}"
-      redirect_to controller: :project, action: :new_package, project: @project
+      redirect_to action: :new, project: @project
```
vpereira, line 1020: Possible improvement here: because all conditions are basically doing the same thing: updating `flash[:error]`, redirecting to `:new` action we could extract it to a service or method, collection the errors and if error exists, set the error in the flash and redirect...

src/api/app/controllers/webui/package_controller.rb
dmarcoux: This is temporary, it's going to be the `create?` method which I'm going to create in the package policy in a following PR.

src/api/app/controllers/webui/package_controller.rb
dmarcoux: This will all be gone in a following PR to address the FIXMEs.

PR #9552, opened 2020-05-14
Authorize package creation with a Pundit policy

dmarcoux:
Refactoring `can_create_package_in?` out of existence.

Following up on #9551.

### DIFF
--- a/src/api/app/models/bs_request_permission_check.rb
+++ b/src/api/app/models/bs_request_permission_check.rb
@@ -385,10 +385,10 @@ class BsRequestPermissionCheck
         @write_permission_in_this_action = true
       end
     else
-      if @target_project && User.session!.can_create_package_in?(@target_project, true)
+      if @target_project && PackagePolicy.new(User.session!, Package.new(project: @target_project), true).create?
         @write_permission_in_target = true
       end
-      if @target_project && User.session!.can_create_package_in?(@target_project, ignore_lock)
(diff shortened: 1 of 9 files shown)


src/api/app/policies/package_policy.rb
```diff
@@ -1,4 +1,18 @@
 class PackagePolicy < ApplicationPolicy
+  # FIXME: Using more than 2 arguments is considered a code smell.
+  def initialize(user, record, ignore_lock = false)
```
dmarcoux, line 3: I could have probably worked around this code smell (so having another argument, `ignore_lock`) by [using a context as explained in Pundit's README](https://github.com/varvet/pundit#additional-context), but this would have involved even more refactoring. I decided this was already good enough for this PR.

dmarcoux:

src/api/app/models/bs_request_permission_check.rb
```diff
@@ -385,10 +385,10 @@ def set_permissions_for_action(action, new_state = nil)
         @write_permission_in_this_action = true
       end
     else
-      if @target_project && User.session!.can_create_package_in?(@target_project, true)
+      if @target_project && PackagePolicy.new(User.session!, Package.new(project: @target_project), true).create?
```
dmarcoux, line 388: I cannot use `Pundit.policy(...)` here since there is a third argument in the call.

PR #9554, opened 2020-05-14
Always verify an uploaded file in case of a _pattern package

marcus-h:
As of now, if a file is PUTed to a _pattern package, the file is
only verified if the _pattern package was non-existent. Instead, the
the file should be always verified (regardless whether the _pattern
package was created or already existed).
Since @pack is not set in check_permissions_for_file in case of a
_pattern package (which is perfectly fine), it is set up in
update_file (the code also takes a potential race into account).
Hence, @pack is always defined when it is passed to
Package.verify_file!.
The new behavior is documented in a testcase.


src/api/app/controllers/source_controller.rb
```diff
@@ -355,10 +355,19 @@ def update_file
...
+                                             follow_project_links: false)
+        @pack = Package.new(name: '_pattern', title: 'Patterns',
+                            description: 'Package Patterns')
+        @prj.packages << @pack
+        @pack.save
+      else
+        @pack = Package.get_by_project_and_name(@project_name, @package_name,
+                                                follow_project_links: false)
+        # very unlikely... (actually this should be a 400 instead of 404)
+        raise RemoteProjectError, 'Cannot modify a remote package' if @pack.nil?
```
vpereira, line 369: in the functional test you are testing for a 400, so does this comment still make sense @marcus-h ?

marcus-h:
On 2020-05-15 01:06:23 -0700, Victor Pereira wrote:
> @vpereira requested changes on this pull request.
>
>
>
> > -    if @package_name == '_pattern' && !Package.exists_by_project_and_name(@project_name, @package_name, follow_project_links: false)
> -      @pack = Package.new(name: '_pattern', title: 'Patterns', description: 'Package Patterns')
> -      @prj.packages << @pack
> -      @pack.save
> +    if @package_name == '_pattern'
> +      if !Package.exists_by_project_and_name(@project_name, @package_name,
> +                                             follow_project_links: false)
> +        @pack = Package.new(name: '_pattern', title: 'Patterns',
> +                            description: 'Package Patterns')
> +        @prj.packages << @pack
> +        @pack.save
> +      else
> +        @pack = Package.get_by_project_and_name(@project_name, @package_name,
> +                                                follow_project_links: false)
> +        # very unlikely... (actually this should be a 400 instead of 404)
> +        raise RemoteProjectError, 'Cannot modify a remote package' if @pack.nil?
>
> in the functional test you are testing for a 400, so does this comment still make sense @marcus-h ?
>
Yes, the "raise RemoteErrorProject" codepath is not triggered by the test (the
test is supposed to test the _pattern specific code in Package.verify_file!).
The RemoteErrorProject stuff is just to account for a highly unlikely race:

- Package.exists_by_project_and_name returns true (hence, we end up in the
  else block)
- in between another request is processed that turns the project denoted by
  @project_name into a remote project
- Package.get_by_project_and_name returns nil (since it is a remote project)

About the comment: a RemoteProjectError results in 404 but in this particular
situation a 400 would be more suitable (actually, 409 would fit best).

vpereira:
@adrianschroeter why do we have to stay compatible with API <= 2.0? Can't we handle "_pattern" as a regular package? IMO To have some package being automatically created when a file is uploaded is weird.

vpereira:
how to bypass it as it is Today:


```
# make sure package doesn't exist
(base) vpereira@linux-qouy:~/open-build-service> osc api -X DELETE /source/home:vpereirabr:Test/_pattern
<status code="ok">
  <summary>Ok</summary>
</status>
```

```
# try to add a pattern with invalid content
(base) vpereira@linux-qouy:~/open-build-service> osc api -X PUT -d "xoxoxoxo" /source/home:vpereirabr:Test/_pattern/obs-test
pattern validation error: 1:1: FATAL: Start tag expected, '<' not found
```

```
# try to add a pattern with valid content
(base) vpereira@linux-qouy:~/open-build-service> osc api -X PUT -d "<pattern xmlns="http://novell.com/package/metadata/suse/pattern"><name>OBS</name></pattern>" /source/home:vpereirabr:Test/_pattern/obs-test
<revision rev="1">
  <srcmd5>297dbe6d7042c2115b3d11d7589ad6ee</srcmd5>
  <time>1590144031</time>
  <user>vpereirabr</user>
  <comment></comment>
  <requestid/>
</revision>
```

```
# try to add a pattern with invalid content
(base) vpereira@linux-qouy:~/open-build-service> osc api -X PUT -d "xoxoxoxo" /source/home:vpereirabr:Test/_pattern/obs-test
<revision rev="2">
  <srcmd5>ec4cdae9df76d3ca3f919c1d40b0c745</srcmd5>
  <time>1590144033</time>
  <user>vpereirabr</user>
  <comment></comment>
  <requestid/>
</revision>
(base) vpereira@linux-qouy:~/open-build-service> osc api  '/source/home:vpereirabr:Test/_pattern/obs-test'
xoxoxoxo
```

PR #9558, opened 2020-05-15
Download 1st ccache archive for linked packages in maintenance prjs

sjamgade:
Maintenance projects have linked packages. These packages could shared
their ccache archive for initial builds allowing faster rebuilds.

releasename is added as param to job file allowing worker to know that
it should use ccache archive of the (release)named packages part of the
projects mentioned in its own repository paths. Here the first hit
wins.

### DIFF
--- a/src/backend/bs_worker
+++ b/src/backend/bs_worker
@@ -2548,6 +2548,49 @@ sub getbinaries {
   return @meta;
 }
 
+sub getlinkedccache {
+  my ($buildinfo, $odir) = @_;
+
+  mkdir_p($odir) || die("mkdir_p $odir: $!\n");
+  for my $other (@{$buildinfo->{'path'}}) {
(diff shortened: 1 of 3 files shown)


src/backend/bs_worker
```diff
@@ -2548,6 +2548,51 @@ sub getbinaries {
   return @meta;
 }
 
+sub getlinkedccache {
+  my ($buildinfo, $odir) = @_;
+
+  mkdir_p($odir) || die("mkdir_p $odir: $!\n");
+  my $pathidx = 1;
+  my $other = $buildinfo->{'path'}->[$pathidx];
+  while ($other) {
```
mlschroe, line 2557:
```
for my $other (@{$buildinfo->{'path'}}) {
...
next if $oprojid eq $buildinfo->{'project'} && $orepoid eq $buildinfo->{'repository'}:
```

PR #9564, opened 2020-05-18
[backend] limit reading jobhistory to 1000 entries

sjamgade:
jobhistory can be really large (imagine all of project history, since
the beginning of time). If some package rebuild failed and was never
rebuilt(successfully) since ages(searching for that package needs entire
jobhistory(which could be in gbs) to be read) we will have huge memory
consumption.

In otherwords, package build failed now and a prior rebuild was
Gigabytes ago (or not all) then we will have memory issues.

### DIFF
--- a/src/backend/bs_repserver
+++ b/src/backend/bs_repserver
@@ -2164,7 +2164,10 @@ sub notify_jobresult {
         $failcnt += 1;
         return 1;
     };
-    BSFileDB::fdb_getall_reverse("$reporoot/$prpa/:jobhistory", $BSXML::jobhistlay, undef, $filter);
+    # each line is around <200 chars (biggest can be pacakage name)
+    # 1000 * 200 ~= ~1.6 MB
+    # setting it to 1000 skip over jobhistory of !$package
+    BSFileDB::fdb_getall_reverse("$reporoot/$prpa/:jobhistory", $BSXML::jobhistlay, 1000, $filter);
     $ninfo{'successive_failcount'} = $failcnt;
     BSNotify::notify('BUILD_FAIL', \%ninfo);
   }


mlschroe:
No, that's not a good way to do it. You're discarding the output of fdb_getall_reverse() anyway, so your filter should never return 1 or -1. Instead it should only return 0 (continue with scan) or -2 (stop scanning).

sjamgade:
In fdb_getall_reverse() `@res` will be populated (even for return 0, there is no way to throw away result, so it is left it to 1) and -2/-1, have the same effect except the last row, which is a few bytes.

mlschroe:
The code reads
```
my $f = $filter->($r);
next unless $f; 
```
so returning 0 will not push anything on `@res`.

hennevogel:
*\*cough*\* pagination *\*cough*\*

:smile:

mlschroe:
Your filter should read:
```
    my $filter = sub {
        return 0 if $_[0]->{'package'} ne $ninfo{'package'};
        return -2 if $_[0]->{'code'} ne 'failed';
        return ++$failcnt >= 100 ? -2 : 0;
    };   
```

mlschroe:
@hennevogel what are you talking about? This is about calculating the count of failed packages. It has nothing to do with pagination.

sjamgade:
okay. understood. I am thankful for you explanation. 

but now that I have grown wiser that jobhistory file could be in GBs. Is it worth  traversing the whole jobhistory file to find 2nd successive failed build. Scanning 1GB at 32k bytes speed can take quite some time. (given that I am able to read perl in a logical and consistent way)

ismail:
Can you please merge this PR as-is since it fixes a critical regression? And as we talked in the daily stand-up a better way is to record per-package failure information in package directory. That can be another PR.

mlschroe:
I'd prefer to merge #9526

ismail:
> I'd prefer to merge #9526 

That’ll fix this issue as well, works for me :)

mlschroe:
Obsoleted with the merge of  #9526

PR #9565, opened 2020-05-18
Remove before_action check package name for new

dmarcoux:
Validations belong in the model, not in a before_action in a controller.

The check `Package.valid_name?` was just a duplicate, `validate :valid_name` already does it.

As for `unique?`, it checks for uniqueness in local and remote projects. This makes `validates :project_id, uniqueness: { scope: :name }` obsolete since it only checks for uniqueness in local projects.


danidoni:
LGTM

Issue #9040, opened 2020-02-06
display bill of materials


lnussel:
Any image build in OBS produces an additional ".packages" file. That file lists the contained rpm packages with their name, version, disturl and most importantly license. That information is very valuable to a class of users that have a whitelist of free software licenses they are allowed to run on their infrastructure.

OBS unfortunately doesn't seem to have a handler for that file so there's just a generic view as used for rpm packages, eg.:

https://build.opensuse.org/package/binary/openSUSE:Containers:Tumbleweed/tumbleweed-busybox-image.20200117174618/containers/aarch64/tumbleweed-busybox-image.aarch64-1.0.0-Build5.242.packages

How about showing a nice table with the actual file content there?

hennevogel:
we have SBOM support by now which is the de-facto standard for this.

lnussel:
ok but that information is not shown either?
https://build.opensuse.org/projects/openSUSE:Factory/packages/openSUSE-MicroOS:kvm-and-xen-sdboot/repositories/images/binaries only has download links, no "Details" like it's available for packages

hennevogel:
I don't think people want to use OBS for viewing/verifying/working with SBOM files. There are tons of already established dedicated tools for this, lot of them are open source.

PR #9571, opened 2020-05-19
[backend] do not publish .packages files anymore

adrianschroeter:
These files are just intermediate files for the .report files.
The .report files get not published either yet, but we should
not confuse users with internal build artefacts.

Better creating something useful in future if needed.

### DIFF
--- a/ReleaseNotes-2.11
+++ b/ReleaseNotes-2.11
@@ -48,6 +48,7 @@ Intentional changes:
 
 Backend:
  * .sha256 files get gpg signed in detached mode now
+ * .packages files get not anymore published
  * Switched from OR to AND when checking CPU flags in _constraints file
 
 Webui:
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -1994,9 +1994,6 @@ sub publish {
 	  $p = $bin;
           $kiwimedium = "$arch/$1" if !$2 && -e "$r/$1.packages";
           $p = mapsleimage($config, "$reporoot/$prp/$arch/:repo", $rbin, $p, $sleimage) if $sleimage;
-        } elsif ($bin =~ /\.packages$/) {
-	  $p = $bin;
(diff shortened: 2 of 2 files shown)


vmoutoussamy:
Vagrant and JeOS images will appreciate this change : ) We got some report from confused users on these files.

dcermak:
Thank you @adrianschroeter!

bear454:
This breaks Public Cloud publishing, which relies on .packages for dependency version checks.

adrianschroeter:
On Dienstag, 26. Mai 2020, 17:38:01 CEST wrote James Mason:
> This breaks Public Cloud publishing, which relies on .packages for dependency version checks.

hm, but this is an internal build artefact inside of OBS. It is just
a base for .report files in the end.

However, couldn't you get it via the api? 

I do not really like to see them published to the user, since 
it is not usefull, and is an internal piece where no guarantees
are given...

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

aosthof:
@adrianschroeter While we could adjust to a future solution - e.g. the .report file - our current image publishing automation relies on the .packages files and taking them away just broke our process. As we need to do a general image refresh end of this week the absence of the .package files is not acceptable.

FYI, @bugfinder already reverted the change for the Devel:, home and SUSE: projects in IBS which should give us some breathing room again - thanks!

lnussel:
The bill of materials is valuable information. Who gets confused by that? Please bring it back. See also  #9040

rjschwei:
We have tools and processes that are broken and basically means we will not get openSUSE images published in AWS, Azure and GCE. vs. people that are being confused.

We should find a solution that meets all our needs and revert this change until such a solution is found and all tools can be adjusted accordingly.

Vogtinator:
Yes, I also like to have the .packages files (or an equivalent) available for easy checking.

PR #9573, opened 2020-05-19
Remove old auth proxy opensuse page

krauselukas:
This html template seems not to be used anywhere in OBS 2.11.
Good chance that I'm not aware of how this is still used, but by greping through the codebase, I couldn't find any hints that this is still used :)


krauselukas:
> +1 You can also remove this line from `.dockerignore`:
> 
> ```
> src/auth-proxy/
> ```

@eduardoj done :)

hennevogel:
@mlschroe @adrianschroeter are we still using this somewhere?

adrianschroeter:
well, it is still running, but a verions indepdend of this one. you can drop it here.

PR #9575, opened 2020-05-20
Fix beta breadcrumb

DavidKang:
After introducing Bootstrap 4.5.0, the breadcrumb icons weren't aligned
anymore.

### Before
Screenshot_2020-05-20 testing

### After
Screenshot_2020-05-20 testing(1)


hennevogel:
we should have some rules about depfu bootstrap updates...

DavidKang:
> we should have some rules about depfu bootstrap updates...

We should add a review-app

PR #9577, opened 2020-05-20
[backend] Speedup jobhistory route by 4-6x

sjamgade:
This patch allows fdb_getall_reverse to return undecoded line to filter.
This raw line can be filtered out earlier without the need of decoding
over the layout.

This is useful for large and long maintained projects. Projects
(Factory, Devel..)  have a long build history, decoding every
line in the jobhistory file only to be filtered out based on filter
conditions (package type, result code, buildtimes) is expensive. This
patch facilitates skipping those costs reach the result faster.

Tested on IBS/SUSE:FACTORY:HEAD/standard/i586/MozillaThunderbird
without patch:
100 results: 6 sec
1000 results: 53 sec

with patch:
100 results: 1 sec
1000 results: 13 sec

### DIFF
--- a/src/backend/BSFileDB.pm
+++ b/src/backend/BSFileDB.pm
@@ -51,6 +51,12 @@ sub decode_line {
   return $r;
 }
 
+sub encode_field {
+    my ($str) = @_;
+    $str =~ s/([\000-\037%|=\177-\237])/sprintf("%%%02X", ord($1))/ge;
+    return $str;
+}
+
 sub encode_line {
   my ($r, $lay) = @_;
   my @line;
@@ -229,9 +235,15 @@ sub fdb_getall_reverse {
       @l = split("\n", $buf, -1);
     }
     for (reverse @l) {
-      my $r = decode_line($_, $lay);
+      my $r = $_;
+      $r = decode_line($_, $lay) if defined($lay);
       if ($filter) {
-        my $f = $filter->($r);
+        my $f;
+        if (defined($lay)) {
+          $f = $filter->($r);
+        } else {
+          $f = $filter->(split('\|', $r));
+        }
 	next unless $f;
 	if ($f < 0) {
 	  push @res, $r if $f == -1;
--- a/src/backend/bs_repserver
+++ b/src/backend/bs_repserver
@@ -3154,24 +3154,30 @@ sub getjobhistory {
   if ($cgi->{'code'} && @{$cgi->{'code'}} == 1 && $cgi->{'code'}->[0] eq 'lastfailures') {
     return getlastfailures($cgi, $projid, $repoid, $arch);
   }
+  my $idx = 0;
+  my %indices = map {$_ => $idx++} @$BSXML::jobhistlay;
+  my $packidx = $indices{'package'};
+  my $codeidx = $indices{'code'};
+  my $endtimeidx = $indices{'endtime'};
   if ($cgi->{'package'} && $cgi->{'code'}) {
-    my %packid = map {$_ => 1} @{$cgi->{'package'}};
+    my %packid = map {BSFileDB::encode_field($_) => 1} @{$cgi->{'package'}};
     my %code = map {$_ => 1} @{$cgi->{'code'}};
-    $filter = sub {$packid{$_[0]->{'package'}} && $code{$_[0]->{'code'}}};
+    $filter = sub {$packid{$_[$packidx]} && $code{$_[$codeidx]}};
   } elsif ($cgi->{'package'}) {
-    my %packid = map {$_ => 1} @{$cgi->{'package'}};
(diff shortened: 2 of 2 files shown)


src/backend/bs_repserver
```diff
@@ -3154,24 +3154,33 @@ sub getjobhistory {
   if ($cgi->{'code'} && @{$cgi->{'code'}} == 1 && $cgi->{'code'}->[0] eq 'lastfailures') {
     return getlastfailures($cgi, $projid, $repoid, $arch);
   }
+  my $idx = 0;
+  my %indices = map {$_ => $idx++} @$BSXML::jobhistlay;
+  my $packidx = %indices{'package'};
+  my $codeidx = %indices{'code'};
```
mlschroe, line 3160: That's not good perl. It should probably be `$indices{'packages'}`

src/backend/bs_repserver
```diff
@@ -3154,24 +3154,33 @@ sub getjobhistory {
...
+  my $endtimeidx = %indices{'endtime'};
   if ($cgi->{'package'} && $cgi->{'code'}) {
     my %packid = map {$_ => 1} @{$cgi->{'package'}};
     my %code = map {$_ => 1} @{$cgi->{'code'}};
-    $filter = sub {$packid{$_[0]->{'package'}} && $code{$_[0]->{'code'}}};
+    $filter = sub {$packid{$_[$packidx]} && $code{$_[$codeidx]}};
```
mlschroe, line 3165: But the filter is only called with one argument, so what is `$_[...]` supposed to do?

And what about the decoding? If you want to skip the decoding you need to encode the values in the filter.

mlschroe:
So we need FileDB::encode_field:
```
sub encode_field {
  my ($str) = @_;
  $str =~ s/([\000-\037%|=\177-\237])/sprintf("%%%02X", ord($1))/ge;
  return $str;
}
```

With that we do:
```
  my %packid = map {BSFileDB($_) => 1} @{$cgi->{'package'}};
```
and with that we do the matching in the filter.

sjamgade:
ah now I understand what you mean. thanks for clarifying

src/backend/BSFileDB.pm
```diff
@@ -59,7 +65,7 @@ sub encode_line {
   ($al) = splice(@lay, -2, 2) if $lay->[-1] eq '[]';
   push @line, defined($r->{$_})  ? $r->{$_} : '' for @lay;
   push @line, @{$r->{$al} || []} if $al;
-  s/([\000-\037%|=\177-\237])/sprintf("%%%02X", ord($1))/ge for @line; 
+  @line = map {encode_field($_)} @line;
```
mlschroe, line 68: Please don't call encode_field here. Perl is a bit slow in regard to function calls, unlike languages like C. Just leave the code duplicated.

src/backend/bs_repserver
```diff
@@ -3154,24 +3154,33 @@ sub getjobhistory {
...
-    my %packid = map {$_ => 1} @{$cgi->{'package'}};
+    my %packid = map {BSFileDB::encode_field($_) => 1} @{$cgi->{'package'}};
     my %code = map {$_ => 1} @{$cgi->{'code'}};
-    $filter = sub {$packid{$_[0]->{'package'}} && $code{$_[0]->{'code'}}};
+    $filter = sub {$packid{$_[$packidx]} && $code{$_[$codeidx]}};
   } elsif ($cgi->{'package'}) {
-    my %packid = map {$_ => 1} @{$cgi->{'package'}};
-    $filter = sub {$packid{$_[0]->{'package'}}};
+    my %packid = map {BSFileDB::encode_field($_) => 1} @{$cgi->{'package'}};
+    $filter = sub {$packid{$_[$packidx]}};
```
mlschroe, line 3168: I still don't get how this can work if the filter is called with the complete string. Don't you want to split the line into fields before calling the filter function?

sjamgade: I might be missing something. 
But I have added `my @line = split('\|', $_);` to the endtime filter (couple of lines below).
Then `@line` gets passed to other filters

src/backend/BSFileDB.pm
sjamgade: okay dropped the change

src/backend/bs_repserver
mlschroe: But only if the endtime filter is used...

Oh wait, there's a bogus `|| 1` in the code! A leftover from some testing, I think.

So your code works, but only because of that bug...

mlschroe: Anyway, it's IMHO bad style. I think the BSFileDB code should do the split if $lay is undef.

sjamgade:
Merging as discussed in standup

PR #9590, opened 2020-05-22
Add announcement functionality to status message

saraycp:
We want to announce something inside the app only for certain groups of users. For example, only for beta users, for admins, etc. Sometimes the announcement is so important that we want it to be visible all over the web site and force the user to acknowledge. But sometimes it is enough to display the message inside the `News` box (status messages).

This functionality is a kind of mix between the already-existing `Announcement` and `StatusMessage` functionalities, so we have decided to merge them in `StatusMessage`.

When an admin creates a new status message, he can choose the new severity `announcement`. This kind of status messages will be displayed in the News box but also as a flash message on the top of every page. They need acknowledgement by the user. From now on, he/she can also choose the **communication scope**, to define who is allowed to see the status message.

status_message_to_announcement_II

Example 1:  Admin creates a normal status message (only visible in the News box) that will be displayed for beta users only.

Example 2:  Admin creates an _announcement status message_ (visible in the News box and on the top) that will be displayed for admin users only.

This PR is the first of a series. In this case we add all the new functionality to Status Message.
The following PRs will migrate the already existing data in `announcements` table to `status_messages` table; as well as drop Announcement feature keeping API backward compatibility.

This PR is a rework from #9538


src/api/app/controllers/status_messages_controller.rb
```diff
@@ -17,34 +14,26 @@ def show
...
-      raise PermissionDeniedError, 'message cannot be deleted, you have not sufficient permissions'
-    end
-
     StatusMessage.find(params[:id]).delete
     render_ok
   end
+
+  private
+
+  # TODO: make it more robust
```
vpereira, line 33: I will improve it here and add some specs around it

src/api/app/views/webui/main/_status_messages.html.haml
```diff
@@ -10,7 +10,6 @@
     - if User.admin_session?
       - unless flipper_responsive?
         .card-footer
-          = link_to('#', 'data-toggle': 'modal', 'data-target': '#status-message-modal') do
+          = link_to(new_status_message_path, class: 'nav-link') do
             %i.fas.fa-plus-circle.text-primary
-            Add status message
-      = render partial: 'add_status_message_modal'
+            Add Status Message
```
dmarcoux, line 15: To have consistent naming in links, let's use `Create Status Message`

src/api/app/views/webui/status_messages/new.html.haml
```diff
@@ -0,0 +1,24 @@
+- @pagetitle = 'New Status Message'
```
dmarcoux, line 1: To have a consistent naming, let's use `Create Status Message`

src/api/app/views/webui/status_messages/new.html.haml
```diff
@@ -0,0 +1,24 @@
...
+.row
+  .col
+    .card
+      .card-body
```
dmarcoux, line 5: We should use the grid here.

So something like:
```
.col-12
  the title
.col-12
  the paragraph
.col-12.col-md-9.col-lg-6
  the form
```

src/api/app/views/webui/status_messages/new.html.haml
```diff
@@ -0,0 +1,24 @@
...
+              Severity:
+              %abbr.text-danger{ title: 'required' } *
+            %p With "announcement" severity the message will be displayed on the top and all over the web site and requires acknowledgement.
+            - options = StatusMessage.severities.keys.map { |severity| [severity.humanize, severity] }
+            = f.select(:severity, options_for_select(options), {}, class: 'custom-select')
+          .form-group
+            = f.label(:communication_scope, 'Communication Scope:')
+            - options = StatusMessage.communication_scopes.keys.map { |scope| [scope.humanize, scope] }
+            = f.select(:communication_scope, options_for_select(options), {}, class: 'custom-select')
+          = f.submit 'Accept', class: 'btn btn-primary'
```
dmarcoux, line 24: The action of this form is to create, so we should name the button `Create`

src/api/app/views/webui/status_messages/new.html.haml
```diff
@@ -0,0 +1,24 @@
...
+          .form-group
+            = f.label(:message) do
+              Message:
+              %abbr.text-danger{ title: 'required' } *
+            = f.text_area(:message, rows: 3, required: true, class: 'form-control')
+          .form-group
+            = f.label(:severity) do
+              Severity:
+              %abbr.text-danger{ title: 'required' } *
+            %p With "announcement" severity the message will be displayed on the top and all over the web site and requires acknowledgement.
```
dmarcoux, line 17: There's a comma missing.

```suggestion
            %p With "announcement" severity, the message will be displayed on the top and all over the web site and requires acknowledgement.
```

src/api/app/models/status_message.rb
```diff
@@ -1,12 +1,49 @@
...
+  belongs_to :user # TODO: rename as creator
+  has_many :status_message_acknowledgements, dependent: :destroy
+  has_many :users, through: :status_message_acknowledgements
+
   validates :user, :severity, :message, presence: true
+
   scope :alive, -> { where(deleted_at: nil).order('created_at DESC') }
+  scope :announcements, -> { alive.where(severity: 'announcement') }
+
+  enum severity: { information: 0, green: 1, yellow: 2, red: 3, announcement: 4 }
```
dmarcoux, line 11: Green, yellow and red aren't severities, they are colors. I know it was already like this, but if we're refactoring status messages, we should change this. Yellow could be Warning, Green is just like Information (we don't need it), Red is for issues/problems, Announcement is the same as information.

src/api/app/views/webui/status_messages/new.html.haml
vpereira: I think we tried that, but it totally broke the view. From the discussion about the "grid" I thought the grid was to stop using straight card, but instead using `.row > .col > .card` no?

src/api/app/models/status_message.rb
vpereira: Totally agree. @hennevogel what do you think?

src/api/app/views/webui/status_messages/new.html.haml
saraycp: I think I understand now. I believe @dmarcoux suggests to add these columns inside the `card-body` not replacing what we had. I'm pushing a new commit changing this.

src/api/app/models/status_message.rb
saraycp: I also agree but I don't think it's time to do it. I wouldn't say we are refactoring Status Message but adding extra functionality. And since we have to keep backward compatibility it won't be fast.

We have caught many things that can be improved regarding status message code and we wrote them down for the future. Things like: renaming severity, moving status message from `main` to its own namespace, renaming `user_id` to `creator_id`, etc.
But I believe none of them are our goal now.

src/api/app/views/webui/status_messages/new.html.haml
saraycp: done!

src/api/app/views/webui/status_messages/new.html.haml
saraycp: done!

src/api/app/views/webui/main/_status_messages.html.haml
saraycp: done!

src/api/app/views/webui/status_messages/new.html.haml
saraycp: done!

src/api/db/migrate/20200522151615_backfill_add_communication_scope_to_status_message.rb
```diff
@@ -0,0 +1,12 @@
+class BackfillAddCommunicationScopeToStatusMessage < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def up
+    StatusMessage.unscoped.in_batches do |relation|
+      # rubocop:disable Rails/SkipsModelValidations
+      relation.update_all communication_scope: 0
+      # rubocop:enable Rails/SkipsModelValidations
+      sleep(0.01)
```
hennevogel, line 9: Hm, why sleep?

src/api/spec/controllers/webui/webui_controller_spec.rb
```diff
@@ -135,4 +135,168 @@ def edit
       end
     end
   end
+
+  describe '#current_announcement' do
```
hennevogel: This should be a unit test of `StatusMessage.visible_for_current_user?` instead

src/api/app/controllers/status_messages_controller.rb
```diff
@@ -17,34 +14,27 @@ def show
...
+    authorize status_message
+    status_message.delete
     render_ok
   end
+
+  private
+
+  # TODO: make it more robust
+  def validate_status_message
+    Suse::Validator.validate(:status_message, request.raw_post)
```
hennevogel, line 35: Not quite sure but doesn't this happen automatically in `ApplicationController.validate_xml_request`?

src/api/db/migrate/20200522151615_backfill_add_communication_scope_to_status_message.rb
saraycp: When I tried to run a migration to add `communication_scope` with default `0`, `strong_migrations` proposed this solution to avoid downtime. It is exactly what `strong_migrations`, I only added rubocop enable/disable.

src/api/spec/controllers/webui/webui_controller_spec.rb
saraycp: Both `current_announcement` and `visible_for_current_user?`disappear in a more recent commit. We are working on squashing commits right now. I'll let you know when it is ready to review again.

src/api/db/migrate/20200522151615_backfill_add_communication_scope_to_status_message.rb
hennevogel: hm okay...

src/api/app/views/webui/main/_status_message.html.haml
```diff
@@ -12,4 +12,4 @@
           = link_to(status_message_path(status_message), method: :delete, title: 'Remove status message') do
             %i.fas.fa-times-circle.text-danger
       .mt-2.mb-0
-        = render_as_markdown(status_message.message)
+        = render_as_markdown(truncate(status_message.message, length: 255))
```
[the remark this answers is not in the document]
saraycp: True, we decided to show the whole text in the News box. independently of its length.

src/api/app/controllers/webui/status_messages_controller.rb
```diff
@@ -26,4 +26,15 @@ def destroy
 
     redirect_to(controller: 'main', action: 'index')
   end
+
+  def acknowledge
+    status_message = StatusMessage.find(params[:id])
+    unless status_message.acknowledge!
+      RabbitmqBus.send_to_bus('metrics', "user.acknowledged_status_message status_message_id=#{status_message.id}")
```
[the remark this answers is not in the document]
saraycp: True :+1:

src/api/app/controllers/webui/status_messages_controller.rb
```diff
@@ -26,4 +26,15 @@ def destroy
 
     redirect_to(controller: 'main', action: 'index')
   end
+
+  def acknowledge
```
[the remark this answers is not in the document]
saraycp: Added!

src/api/app/views/webui/main/_add_status_message_modal.html.haml
```diff
@@ -4,12 +4,13 @@
       .modal-header
         %h5.modal-title Add Status Message
       .modal-body
-        = form_tag({ controller: 'status_messages', action: 'create' }, method: 'post') do
+        = form_tag(status_messages_path, method: 'post') do
```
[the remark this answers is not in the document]
saraycp: True :+1:

src/api/app/views/webui/status_messages/new.html.haml
```diff
@@ -0,0 +1,19 @@
+- @pagetitle = 'New Status Message'
+.row
+  .col
+    .card
+      .card-body
+        %h3= @pagetitle
+        = form_for(StatusMessage.new, url: status_messages_path, method: 'post') do |f|
```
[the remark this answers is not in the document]
saraycp: True :+1:

src/api/app/views/webui/status_messages/new.html.haml
dmarcoux: @saraycp Yes, this is what I meant. The example from my previous comment should be done under `.card-body`, so starting on line 6. Sorry @vpereira, I wasn't clear enough!

src/api/app/models/status_message.rb
dmarcoux: @saraycp Aright no worries. It's true that it's out of scope. I'm fine with this.

src/api/app/controllers/status_messages_controller.rb
vpereira: just uncomment and run the specs should make us sure

src/api/app/views/webui/status_messages/new.html.haml
vpereira: it was addressed!

src/api/db/migrate/20200522151615_backfill_add_communication_scope_to_status_message.rb
vpereira: https://www.rubydoc.info/gems/strong_migrations/0.6.2

src/api/app/models/status_message.rb
```diff
@@ -1,12 +1,51 @@
...
     self.deleted_at = Time.now
     save
   end
+
+  def acknowledge!
+    users << User.session!
+  end
+
+  def self.newest_announcement_for_current_user
+    announcement = StatusMessage.announcements.find_by(communication_scope: StatusMessage.communication_scopes_for_current_user)
```
vpereira, line 35: I would prefer to raise an exception here than return nil. Nil checking is a bad smell.

src/api/app/views/layouts/webui/webui.html.haml
```diff
@@ -55,6 +55,7 @@
     .container-xxl.sticky-top.flash-and-announcement.text-word-break-all
       - unless @hide_announcement_notification
         #announcement= render partial: 'layouts/webui/announcement', locals: { pending_announcement: @pending_announcement }
```
vpereira, line 57: this should probably be deleted, no?

src/api/app/models/status_message.rb
vpereira: I will refactor it later.

src/api/app/controllers/status_messages_controller.rb
vpereira: I did the test, and commenting the line out makes the test fail. So in the controller layer, there is no. I see that we have it in another controllers and services too, so i would keep it.

src/api/app/views/layouts/webui/_status_message_announcement.html.haml
```diff
@@ -0,0 +1,12 @@
+- if current_announcement
+  .row.justify-content-center#status-message-announcement
+    .col-12
+      .alert.fade.show.alert-notice
+        .row
+          .col-md-10
+            %i.fa.fa-bullhorn.text-info
+            = truncate(current_announcement.message, length: 80)
```
DavidKang, line 8: Why are you truncating the message?

saraycp: I was afraid it was to long for a flash. But it's true the user won't be able to read the whole text before accepting it (if he/she doesn't realize it is completely displayed in the News box). @hennevogel what do you think?

src/api/app/models/status_message.rb
```diff
@@ -1,12 +1,51 @@
...
   def delete
     self.deleted_at = Time.now
     save
   end
+
+  def acknowledge!
+    users << User.session!
+  end
+
+  def self.newest_announcement_for_current_user
```
hennevogel, line 34:
```suggestion
  def self.latest_for_current_user
```

src/api/app/models/status_message_acknowledgement.rb
```diff
@@ -0,0 +1,9 @@
+class StatusMessageAcknowledgement < ApplicationRecord
+  belongs_to :status_message
+  belongs_to :user
+
+  validates :status_message, presence: true
+  validates :user, presence: true
+
+  validates :status_message_id, uniqueness: { scope: :user_id, message: 'The user has already acknowledged the status message' }
```
hennevogel, line 8:
```suggestion
  validates :status_message_id, uniqueness: { scope: :user_id, message: 'You have already acknowledged the message' }
```

src/api/app/views/layouts/webui/_status_message_announcement.html.haml
hennevogel, line 8: Why truncate and how can I see the rest of the message? What about markdown?

hennevogel: We need at least the ability to include links.

src/api/app/views/webui/status_messages/acknowledge.js.haml
```diff
@@ -0,0 +1,5 @@
+- if flash[:error]
```
hennevogel, line 1: You are not using haml inside this partial, this is ruby. Should be `acknowledge.js.erb`

src/api/app/views/webui/status_messages/new.html.haml
```diff
@@ -0,0 +1,28 @@
+- @pagetitle = 'Create Status Message'
+.row
+  .col
+    .card
+      .card-body
+        .row
+          .col-12
+            %h3= @pagetitle
+          .col-12
+            %p With "announcement" severity, the message will be displayed on the top and all over the web site and requires acknowledgement.
```
hennevogel, line 10: This should be closer to the select. Like https://getbootstrap.com/docs/4.5/components/forms/#help-text

src/api/app/views/status_messages/_status_message.xml.builder
```diff
@@ -0,0 +1,7 @@
+builder.status_message(id: status_message.id) do |m|
+  m.message status_message.message
+  m.user status_message.user.login
+  m.severity status_message.severity
+  m.scope status_message.communication_scope
+  m.created_at status_message.created_at
+end
```
hennevogel, line 7: Why do you change from attributes to elements?

src/api/app/views/webui/status_messages/acknowledge.js.haml
DavidKang: IMO, they are using HAML. or I'm wrong?

src/api/app/models/status_message.rb
```diff
@@ -31,7 +31,7 @@ def acknowledge!
     users << User.session!
   end
 
-  def self.newest_announcement_for_current_user
+  def self.latest_for_current_user
```
saraycp, line 34: You should also rename it where we use it:
spec/models/status_message_spec.rb
app/controllers/webui/webui_controller.rb

src/api/app/views/layouts/webui/_status_message_announcement.html.haml
saraycp: If we don't truncate, then we can render as markdown again. Yes!.

hennevogel: You worry too much :smile:

src/api/app/views/status_messages/_status_message.xml.builder
vpereira: because element is the right way to do. The rule of thumb is that every instance method or attribute should be element with exception of id, which is a metadata.

src/api/app/views/webui/status_messages/acknowledge.js.haml
saraycp: We are using haml. We could use `<% if ` instead of `- if`, does it have any benefit?

src/api/app/views/webui/status_messages/new.html.haml
saraycp: I like the idea

src/api/spec/controllers/status_messages_controller_spec.rb
```diff
@@ -42,7 +42,7 @@
     let(:request_xml) do
       <<~XML
         <status_messages>
-          <message severity="1">New message was sent!</message>
+          <message severity="information">New message was sent!</message>
```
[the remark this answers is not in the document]
krauselukas: If I see correctly 1 is green in the enum

src/api/app/views/webui/status_messages/acknowledge.js.haml
hennevogel: ups. did not see this. sorry

src/api/spec/controllers/status_messages_controller_spec.rb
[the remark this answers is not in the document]
saraycp: I don't think it affects, but I changed it anyway. Thanks!

src/api/app/views/status_messages/_status_message.xml.builder
hennevogel: I would tend to agree, but you now changed the complete API. I can't upload the XML anymore I've always uploaded and get the same result. Can we try to avoid this?

src/api/app/controllers/status_messages_controller.rb
```diff
@@ -1,50 +1,38 @@
 class StatusMessagesController < ApplicationController
-  class PermissionDeniedError < APIError
-    setup 403
-  end
-
-  class CreatingMessagesError < APIError; end
+  skip_before_action :require_login, only: [:show, :index]
```
krauselukas, line 2: @saraycp should we really skip this? I think we almost require to be logged in for all api endpoints (besides the /about)

saraycp: In the WebUI, non-logged-in users (nobody) can see certain kind of announcements.
AFAIR, we tried to do the same via API by adding this line.

However, now that we decided no to restrict the XML output according to communication scope, we can remove this line, otherwise non-logged-in users could see status messages published for other groups.

What do you think @vpereira and @hennevogel ?

krauselukas: Ok, I just thought we limit the usage of the API as much as possible to authenticated users, I don't see a problem in making those information's available to everyone.

krauselukas: Since its more a political question, its not a blocker to me :D those informations are public anyway, everyone can sign up.

adrianschroeter:
On Donnerstag, 28. Mai 2020, 17:07:46 CEST wrote Lukas Krause:
> @krauselukas commented on this pull request.
> 
> 
> 
> > @@ -1,50 +1,38 @@
>  class StatusMessagesController < ApplicationController
> -  class PermissionDeniedError < APIError
> -    setup 403
> -  end
> -
> -  class CreatingMessagesError < APIError; end
> +  skip_before_action :require_login, only: [:show, :index]
> 
> Since its more a political question, its not a blocker to me :D those informations are public anyway, everyone can sign up.

not necessary in all instances. Eg. not in build.suse.com where you need to get approved.

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

krauselukas:
I just realized one thing, if the announcement is for 'all_users' I have no way of closing the flash as user nobody.
Screenshot_2020-05-28 Open Build Service

saraycp:
> I just realized one thing, if the announcement is for 'all_users' I have no way of closing the flash as user nobody.

Yes, it is like this on purpose. Closing means acknowledging, and we store the acknowledgements in database. This way we know if the user already accepted without using cookies. When there is no user (nobody) it makes no sense to store that association in database.

It's true that an 'all_users' announcement will be visible forever unless a new  'all_users' announcement is created or the admin removes it. But I think it's the only way if we are not going to use cookies.

krauselukas:
> > I just realized one thing, if the announcement is for 'all_users' I have no way of closing the flash as user nobody.
> 
> Yes, it was made like this on purpose. Closing means acknowledging, and we store the acknowledgements in database. This way we know if the user already accepted without using cookies. When there is no user (nobody) it makes no sense to store that association in database.
> 
> It's true that an 'all_users' announcement could keep visible for long time until the admin removes it or a new 'all_users' announcement replaces it. But I think it's the only way if we are not going to use cookies.

I understand, but could be annoying since we also use the sticky top, often I don't log in if I just want to check quickly for a package or other read only things, especially on mobile. If I cannot close I feel haunted, could it maybe be a combination of cookies and writing to database? Or maybe for now, not having announcements for user nobody. I guess its enough for now if it only appears in the news box for not logged in users.

src/api/app/views/status_messages/_status_message.xml.builder
vpereira: I'm not sure. First we changed it anyway adding new elements. Second XML or HTML, that's just views.

src/api/app/controllers/status_messages_controller.rb
hennevogel: We are not trying to hide anything from anyone. We just don't want to bother people with stuff they don't need to see in the UI

In the API this is about data manipulation with some client. The API is not a UI. And @krauselukas is right, we usually only do API calls authenticated. Stick to that.

hennevogel:
> If I cannot close I feel haunted

For now we can also be cool with you feeling haunted :smile:

krauselukas:
> > If I cannot close I feel haunted
> 
> For now we can also be cool with you feeling haunted smile
> 
> ghost ghost ghost ghost ghost ghost ghost ghost ghost ghost

Fine then :+1: 
https://media.giphy.com/media/51Uiuy5QBZNkoF3b2Z/giphy.gif

src/api/app/controllers/status_messages_controller.rb
saraycp: Line dropped!

src/api/app/views/status_messages/_status_message.xml.builder
vpereira: and as you explained [here](https://github.com/openSUSE/open-build-service/pull/9590#discussion_r431958661), XML API isn't an UI, so because this endpoint isn't exposed by osc, we can change it something that make sense

hennevogel: I'm not saying you can't do it or that it's impossible. I'm asking you to be a bit more backward compatible. That would only require us to change back to attributes right?

vpereira: > I'm not saying you can't do it or that it's impossible. I'm asking you to be a bit more backward compatible. That would only require us to change back to attributes right?

wrong. 

The XML view looks like:

```
<message severity="0">
a single sample message. contains some text ... attributes are optional and ignored when sent (put) to the api. the format of this example will be used as a convenient way to put a message.
</message>
```

After this PR it will look like:

```
<status message id=XXX>
  <message>blabla</message>
  <severity>information</severity>
  <scope>all_users/scope>
</status_message>
```

After your proposal it would look like:

```
<message severity="0" scope="0" id="XXX">blabla</message>
```

which IMO is wrong specially when you look into the 'index' route where you have

```
<status messages count='x'>
<message severity="0" scope="0">blabla</message>
<message severity="0" scope="0">blabla</message>  
</status_messages>
```

but the correct should be:

```
<status messages count='x'>
<status_message id="1" >...</status_message>
<status_message id="2">...</status_message>  
</status_messages>
```

p.s the id will be necessary in the destroy, which I didn't want to change now, but as all API destroy endpoints it should get the id from the XML and not from the URL

https://build.opensuse.org/apidocs/index#174

hennevogel: Okay, did not know this was this untidy. Let's declare this a breaking change because we clean up the `StatusMessage` API :+1: 

Please add something to the release notes about this.

saraycp: @hennevogel Release notes updated

PR #9592, opened 2020-05-22
[backend] initial dedicated source publishing implementation

adrianschroeter:
Allows to specify an rsync module where sources gets published.
The structure is in $rsync/$project/$package/$srcmd5/ directory.

This can be derived from DISTURL from the binaries.

### DIFF
--- a/src/backend/BSConfig.pm.template
+++ b/src/backend/BSConfig.pm.template
@@ -171,6 +171,11 @@ our $relsync_pool = {
 #for example "--whole-file" if network faster than disk
 #our $rsync_extra_options = "--whole-file";
 
+# optional source publishing to rsync server
+#our $sourcepublish_sync = 'rsync://127.0.0.1/sources';
+# optional filter to publish only some areas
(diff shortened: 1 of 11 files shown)


src/backend/BSConfig.pm.template
```diff
@@ -171,6 +171,11 @@ our $relsync_pool = {
 #for example "--whole-file" if network faster than disk
 #our $rsync_extra_options = "--whole-file";
 
+# optional source publishing to rsync server
+#our $publishsource_sync = 'rsync://127.0.0.1/sources';
+# optional filter to publish only some areas
+#our $publishsource_filter = [ "openSUSE:.*", "SUSE:.*" ];
```
hennevogel, line 177: We really should expose those to `/configuration` in the future...

adrianschroeter: it is a perl regexp, this shouldn't be in /configuration.

if you need a check in frontend we can introduce a request route.

PR #9609, opened 2020-05-27
[backend] fix warning in missing dbgsplit handling

adrianschroeter:
seems to be a typo from the beginning, right?


mlschroe:
i think this is supposed to be
```
$stageserver .= $dbgsplit if $dbgsplit && $stageservers;
```

The stageserver test is for compatibility. See commit adfebff94f7de5df5c4cf2cdf32f512df98f1935

PR #9611, opened 2020-05-27
[backend] do not reconstruct containers in the root filesystem

mlschroe:
Instead, use the uploaddir in our partition.


adrianschroeter:
rspec tests seem to be flaky and independ of the change

PR #9612, opened 2020-05-27
Refactor request_controller

dmarcoux:
See commit messages for details.

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -322,7 +322,10 @@ class Webui::RequestController < Webui::WebuiController
   def set_package
     return unless params.key?(:package_name)
 
-    @package = Package.find_by(name: params[:package_name])
+    @package = Package.get_by_project_and_name(params[:project_name], params[:package_name],
+                                               use_source: false, follow_project_links: true, follow_multibuild: true)
+  rescue APIError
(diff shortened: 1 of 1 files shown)


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -11,20 +11,19 @@ class Webui::RequestController < Webui::WebuiController
   after_action :verify_authorized, only: [:create]
 
   def create
-    request = BsRequest.new(bs_request_params)
-    authorize request, :create?
-    begin
-      request.save!
-    rescue APIError => e
-      flash[:error] = e.message
+    @request = BsRequest.new(bs_request_params)
```
dmarcoux, line 14: This also allows us to access the newly created request in an `after_action` callback. One example in which this is useful is when we need to supersede one/multiple request(s). We first need to create a request, then mark the superseded requests as such by the newly created request.

vpereira: `request` is already used by rails in the controller, as `response`. I would avoid to call it `@request`, even if ugly `@bs_request` would be more appropriate

dmarcoux:
Please have another look. In the end, I had to keep relying on `rescue` since `BsRequest` and `BsRequestAction` models are sometimes raising instead of using validations. At least, we now get error messages from validations.

PR #9618, opened 2020-05-28
WIP: Announcement with status message

vpereira:
Keep the Announcement XML API, but use StatusMessage as model


vpereira:
i will open it again when the main PR is merged

Issue #9615, opened 2020-05-27
Add a new API to copy single files


aplanas:
**Is your feature request related to a problem? Please describe.**

I am implementing a new PoC that explore git as an alternative [tool](https://github.com/aplanas/obsgit) to track the history of an OBS source project. This tool have two primitives: export a package from OBS to git, and import a package from git to OBS. For now is also using OBS as a file store for binary files (`tar.gz` for example).

I would like to copy a file form a project/package into a different project/package, and avoid download the binary and upload it again in a different place, as is implemented today.

**Describe the solution you'd like**

A new API that can copy a single file from one project/package to another project/package under a different file name.

**Describe alternatives you've considered**

As today, I am downloading and uploading the package, wasting bandwidth and strategies in OBS to optimize the storage.

aplanas:
There is a `BSSrcrep::copyfiles` that maybe can be leveraged.

hennevogel:
@mlschroe maybe you can put @aplanas on the right path?

marcus-h:
On 2020-05-27 07:31:51 -0700, Alberto Planas wrote:
> **Is your feature request related to a problem? Please describe.**
>
> I am implementing a new PoC that explore git as an alternative [tool](https://github.com/aplanas/obsgit) to track the history of an OBS source project. This tool have two primitives: export a package from OBS to git, and import a package from git to OBS. For now is also using OBS as a file store for binary files (`tar.gz` for example).
>
> I would like to copy a file form a project/package into a different project/package, and avoid download the binary and upload it again in a different place, as is implemented today.
>
You could copy the entire package/all files into the target package
(without making an actual commit) and then simply commit a new filelist
to the target package (this a creates a new commit).

Example:
Let's suppose you want to copy the "osc.spec" file from the
openSUSE:Factory/osc package to the home:Marcus_H/copy_file package.

(1) Retrieve the current filelist of the home:Marcus_H/copy_file package

marcus@linux:~> curl -u Marcus_H https://api.opensuse.org/source/home:Marcus_H/copy_file
Enter host password for user 'Marcus_H':
<directory name="copy_file" rev="1" vrev="1" srcmd5="5db8a391c8f3173348d0c1653b765ad0">
  <entry name="existing_file" md5="3858f62230ac3c915f300c664312c63f" size="6" mtime="1590602954"/>
</directory>
marcus@linux:~>


(2) Get the (latest) revision of the openSUSE:Factory/osc package and
[...]

PR #9619, opened 2020-05-28
[backend] Support a source copy without creating a commit

marcus-h:
By passing the "rev=repository" query parameter to the "copy"
route, the source files of the origin package are just made available
in the target package (for instance, for later processing). No commit
to the target package is made during this operation (see the addrev
with "rev=repository" semantics).
For a potential use case, see [1].

[1] https://github.com/openSUSE/open-build-service/issues/9615

### DIFF
--- a/src/backend/bs_srcserver
+++ b/src/backend/bs_srcserver
@@ -3397,7 +3397,7 @@ sub update_link_in_service {
 
 sub sourcecopy {
   my ($cgi, $projid, $packid) = @_;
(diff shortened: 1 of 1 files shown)


mlschroe:
Must that really be a regexp match?

marcus-h:
On 2020-05-28 04:38:05 -0700, Michael Schroeder wrote:
> Must that really be a regexp match?
>
No, fixed/changed:)

PR #9620, opened 2020-05-28
Improve ruby dialect on StatusHelper#parse_status

vpereira:


src/api/app/helpers/webui/projects/status_helper.rb
```diff
@@ -40,42 +39,45 @@ def parse_status(package)
...
       comments_to_clear << package['failedcomment']
     end
-
     { summary: outs, sortkey: sortkey, icon_type: icon, comments_to_clear: comments_to_clear }
   end
+  # rubocop:enable Rails/OutputSafety
+
+  private
+
+  def age_from_devel(package)
```
hennevogel, line 80: :thought_balloon: Why the parameter if package is always the same hash?

PR #9613, opened 2020-05-27
Drop old announcements feature from WebUI

saraycp:
This is the second part of a series of PRs that started with  #9590.

The goal is merging Announcement into StatusMessage, adding the possibility to publish messages to certain groups of users and, depending on the severity, show it in the News box only or on the top of the page all over the website.

Co-authored-by: Victor Pereira <vpereira@suse.com>

PR #9624, opened 2020-05-29
Dump announcement into status message

saraycp:
As announcement functionality has been merged to status message, we also dump the existing data from the old to the new table.

This PR is part of a series, the previous ones were: #9590 and #9613


hennevogel:
please explain why you add the do not merge label

PR #9625, opened 2020-05-29
[backend] use releasename for comparison to useccache when available

sjamgade:

### DIFF
--- a/src/backend/bs_worker
+++ b/src/backend/bs_worker
@@ -3133,7 +3133,9 @@ sub dobuild {
   my $bconf = Build::read_config($buildinfo->{'arch'}, "$buildroot/.build.config");
   my $useccache=0;
   if (exists $bconf->{'buildflags:useccache'}) {
-      if (grep {$_ eq "useccache:$packid"} @{$bconf->{'buildflags'} || []}) {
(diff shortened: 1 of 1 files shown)


src/backend/bs_worker
```diff
@@ -3133,7 +3133,9 @@ sub dobuild {
   my $bconf = Build::read_config($buildinfo->{'arch'}, "$buildroot/.build.config");
   my $useccache=0;
   if (exists $bconf->{'buildflags:useccache'}) {
-      if (grep {$_ eq "useccache:$packid"} @{$bconf->{'buildflags'} || []}) {
+      my $opackid = $packid;
+      $opackid = $buildinfo->{'releasename'} if ($buildinfo->{'releasename'});
+      if (grep {$_ eq "useccache:$opackid"} @{$bconf->{'buildflags'} || []}) {
```
mlschroe, line 3138: I think this should be `$_ eq "useccache:$opackid" || $_ eq "useccache:$packid"`

sjamgade: so do you want to allow putting in both ?
And not check for releasename only in case it is specified ?

PR #9626, opened 2020-05-29
Always verify an uploaded file in case of a _pattern package

vpereira:
This is just the rework of #9554

As of now, if a file is PUTed to a _pattern package, the file is
only verified if the _pattern package was non-existent. Instead, the
the file should be always verified (regardless whether the _pattern
package was created or already existed).
Since @pack is not set in check_permissions_for_file in case of a
_pattern package (which is perfectly fine), it is set up in
update_file (the code also takes a potential race into account).
Hence, @pack is always defined when it is passed to
Package.verify_file!.
The new behavior is documented in a testcase.

Co-authored-by: Marcus Hüwe <suse-tux@gmx.de>

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -355,10 +355,19 @@ class SourceController < ApplicationController
     # _pattern was not a real package in former OBS 2.0 and before, so we need to create the
     # package here implicit to stay api compatible.
     # FIXME3.0: to be revisited
-    if @package_name == '_pattern' && !Package.exists_by_project_and_name(@project_name, @package_name, follow_project_links: false)
-      @pack = Package.new(name: '_pattern', title: 'Patterns', description: 'Package Patterns')
-      @prj.packages << @pack
-      @pack.save
+    if @package_name == '_pattern'
+      if !Package.exists_by_project_and_name(@project_name, @package_name,
+                                             follow_project_links: false)
+        @pack = Package.new(name: '_pattern', title: 'Patterns',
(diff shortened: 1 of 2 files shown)


marcus-h:
LGTM.
(Minor: maybe we should keep the FIXME.)

vpereira:
> LGTM. (Minor: maybe we should keep the FIXME.)

done

dmarcoux:
This is what I've done to test this:

```
# Start from clean state
dosc api -X DELETE /source/home:Admin:test/_pattern
# Try with an invalid pattern, it fails
dosc api -X PUT -d "xoxoxoxo" /source/home:Admin:test/_pattern/obs-test
# Try with a valid pattern, it works
dosc api -X PUT --file=/path/to/my/pattern_file /source/home:Admin:test/_pattern/obs-test
# Try with an invalid pattern, it fails again
dosc api -X PUT -d "xoxoxoxo" /source/home:Admin:test/_pattern/obs-test
# Check the pattern, it should still be valid (not xoxoxoxo)
dosc api '/source/home:Admin:test/_pattern/obs-test'


pattern_file:

<pattern xmlns="http://novell.com/package/metadata/suse/pattern">
  <name>OBS</name>
  <summary>My summary</summary>
  <description>My description</description>
  <category>My category</category>
</pattern>
```

With the changes included in this PR, I get this error when trying to `PUT` an invalid pattern (whether it exists or not)
```
pattern validation error: 1:1: FATAL: Start tag expected, '<' not found
```

This is expected. Previously, it would work and this was wrong.

Issue #9617, opened 2020-05-27
DEPRECATION WARNING: ActionView::Base instances must implement `compiled_method_container` or use the class s method `with_empty_template_cache` for constructing an ActionView::Base instance that has an empty cache. 


krauselukas:
DEPRECATION WARNING: ActionView::Base instances must implement `compiled_method_container` or use the class method `with_empty_template_cache` for constructing an ActionView::Base instance that has an empty cache. (called from render_xml at /srv/www/obs/api/app/mixins/can_render_model.rb:8)

This gets thrown many times when running the consistency_check job.

PR #9627, opened 2020-05-29
Silence deprecation warnings in production

hennevogel:
They have no value in production, the admin can't do anything
about them.

### DIFF
--- a/src/api/config/environments/production.rb
+++ b/src/api/config/environments/production.rb
@@ -27,7 +27,7 @@ OBSApi::Application.configure do
   config.action_mailer.delivery_method = :sendmail
   config.action_mailer.perform_caching = false
 
-  config.active_support.deprecation = :log
+  config.active_support.deprecation = :silence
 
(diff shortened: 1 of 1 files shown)


hennevogel:
I have the **unfounded** suspicion that this is the base of some of the sync problems we're seeing...

hennevogel:
The problem currently is #9617 which happens **every** time we render a model to XML...

vpereira:
> The problem currently is #9617 which happens **every** time we render a model to XML...

it should fix it https://github.com/openSUSE/open-build-service/issues/9617#issuecomment-636117615 (untested code :))

saraycp:
> it should fix it [#9617 (comment)](https://github.com/openSUSE/open-build-service/issues/9617#issuecomment-636117615) (untested code :))

Even if there is a possible solution for #9617 and this PR is a temporary solution, I'm going to merge it to avoid handling monkey patches in production. These days we are deploying frequently and this make things easier.

PR #9617 is still open, so we can do follow-up there.

PR #9616, opened 2020-05-27
Request delegation

adrianschroeter:
[api] support request delegation to package origin project.
    
    Adds an option to let the server delegate the request for
    a target project to the project where the packages is actual
    living.
    
    This works when source has no package link, eg from submitting
    from a base distro project to another one.
    
    This is needed for Jump submissions in first place (jsc#OBS-59)

PR #9640, opened 2020-06-02
Fix version on data migration

saraycp:
Taking advantage of this data migration has not been run in
production yet, we fix its version to a non-future date.

Related to #9616


hennevogel:
Sorry this slipped through and created so much confusion...

PR #9646, opened 2020-06-02
Update rack version in obs-bundled-gems.spec file

DavidKang:
We need to update the rack version to make obs-bundled-gems works


hennevogel:
rack was part of the rails update you did?

DavidKang:
When I run the bundle update, It also updated the rack gem

PR #9650, opened 2020-06-03
Remove unneeded script

dmarcoux:
Nobody is using this.


dmarcoux:
Oh well, I was wrong. @hennevogel and @saraycp use it. We'll keep it then.

PR #9670, opened 2020-06-04
Ccache relasename and first archive

sjamgade:
Commit 1:
use releasename too to decide ccache as buildrequirement
this is in continuation to d5352ff

Commit 2:
store the first _ccache.tar for unchanged results
it is better to store ccache archive even in case of unchanged builds if
it is the first archive created.
this is a corner case - ccache is enabled, but the built result is same as
last successful result when ccache was not enabled.


src/backend/BSSched/BuildJob.pm
```diff
@@ -546,6 +546,11 @@ sub jobfinished {
     BSUtil::appendstr("$dst/logfile", "\nRetried build at ".localtime(time())." returned same result, skipped\n");
     my $jobhist = makejobhist($info, $status, $js, 'unchanged');
     addbuildstats($jobdatadir, $dst, $jobhist) if $all{'_statistics'};
+
+    my $ccachetar = "$jobdatadir/_ccache.tar";
+    my $occachetar = "$dst/_ccache.tar";
+    rename($ccachetar, "$dst/_ccache.tar") if -e $ccachetar && !-e $occachetar;
+
```
mlschroe, line 553: You should also use `$occachetar` in the rename() call. And maybe use $all{'_ccache.tar'} instead of the -e?

Issue #9671, opened 2020-06-04
apidocs shows most_active instead of most_active_(projects|packages)


perlpunk:
# Issue Description

https://build.opensuse.org/apidocs/index#164 shows:
`GET /statistics/most_active?type=<type>&limit=<limit>`

Expected Result
============
It should instead document `most_active_projects` and `most_active_packages`

Also the schema and examples for it are outdated. For example, there is no `update_count` in the result.

perlpunk:
Created #9672

DavidKang:
Handled in PR #9672

PR #9672, opened 2020-06-04
Fix apidocs for most_active*

perlpunk:
Fixes issue #9671

https://build.opensuse.org/apidocs/index#164 shows:
`GET /statistics/most_active?type=<type>&limit=<limit>`

Now it documents `most_active_projects` and `most_active_packages`

Also I updated the schema / examples, as they didn't reflect the actual results.


docs/api/api/most_active_packages.xsd
```diff
@@ -0,0 +1,38 @@
...
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+
+  <xs:element name="most_active_packages">
+    <xs:annotation>
+      <xs:documentation>
+        List of most active packages or projects in the openSUSE build
+        service ordered activity.
+      </xs:documentation>
+    </xs:annotation>
+    <xs:complexType>
```
perlpunk, line 11: I think this need to be fixed. We only have one type here now.

perlpunk:
I'm not sure what is failing here: https://app.circleci.com/pipelines/github/openSUSE/open-build-service/596/workflows/95509e4f-d727-4ad3-b139-92b9d1f3c216/jobs/44729/tests

docs/api/api/most_active_packages.xsd
DavidKang: Hi @perlpunk, this is because you need to adapt the tests, since you added 2 new schemas and removed `most_active`. Have a look [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/test/functional/backend_test.rb#L18).

perlpunk: thanks, I had missed that one when grepping the sources :)

PR #9675, opened 2020-06-04
Autocorrect (some) RuboCop TODOs - RSpec/{ExpectActual,ScatteredLet}

dmarcoux:


src/api/spec/controllers/webui/patchinfo_controller_spec.rb
```diff
@@ -56,6 +31,29 @@ def do_proper_post_save
         <binary>fake_binary_001</binary>
       </patchinfo>')
   end
+  let(:other_user) { create(:confirmed_user, :with_home, login: 'gilberto') }
```
DavidKang, line 34: IMO, those 3 `let` were fine on top, we should only move `do_proper post_save` method.

dmarcoux: I agree.

PR #9677, opened 2020-06-05
Specify case_sensitive on uniqueness validations

### LINKED ISSUE #9439: Deprecation Warning Uniqueness validator
```
DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison
 in Rails 6.1. To continue case sensitive comparison on the :name attribute in Status::Check model,
 pass `case_sensitive: true` option explicitly to the uniqueness validator. (called from update at
 /srv/www/obs/api/app/controllers/status/checks_controller.rb:21)
```

```
DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison
 in Rails 6.1. To continue case sensitive comparison on the :flag attribute in Flag model, pass
 `case_sensitive: true` option explicitly to the uniqueness validator. (called from block in store at
 /srv/www/obs/api/app/models/project.rb:786)

```

saraycp:
Uniqueness validation has been `case_sensitive: true` by default. From Rails 6.1 on, the default will be false so we have to specify the opposite to avoid deprecation warnings and malfunction in the future.

The concept can be a bit confusing, but `case_sensitive: true` means: the validation is going to check if an object with the same value and the **same capitalization** already exists.

Example:
If we have this validation in the Project's model
```validates :name, uniqueness: { case_sensitive: true }```
This means a project with "lorem" name and another project with "LOREM" name can be both created. Because the validation is taking capitalization into account.

Some has been considered case insensitive for the following reasons.
- it makes sense according to the values and usage,
- they contain fixed values added by developers and not users,
- they create indexes in database that crash with equal values using
different capitalization.

Closes #9439

### DIFF
--- a/src/api/app/models/architecture.rb
+++ b/src/api/app/models/architecture.rb
@@ -19,7 +19,7 @@ class Architecture < ApplicationRecord
   scope :unavailable, -> { where(available: 0) }
 
   #### Validations macros
-  validates :name, uniqueness: true
+  validates :name, uniqueness: { case_sensitive: false }
   validates :name, presence: true
 
   #### Class methods using self. (public and then private)
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -32,6 +32,7 @@ class BsRequestAction < ApplicationRecord
   validates :type, presence: true
   validates :type, uniqueness: {
     scope: [:target_project, :target_package, :bs_request_id],
+    case_sensitive: false,
     conditions: -> { where.not(type: ['add_role', 'maintenance_incident']) }
   }
 
--- a/src/api/app/models/download_repository.rb
+++ b/src/api/app/models/download_repository.rb
@@ -4,7 +4,7 @@ class DownloadRepository < ApplicationRecord
   belongs_to :repository
 
   validates :repository, presence: true
-  validates :arch, uniqueness: { scope: :repository_id }, presence: true
+  validates :arch, uniqueness: { scope: :repository_id, case_sensitive: false }, presence: true
   validate :architecture_inclusion
   validates :url, presence: true, format: { with: /\A[a-zA-Z]+:.*\Z/ } # from backend/BSVerify.pm
   validates :repotype, presence: true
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

File: src/api/app/models/architecture.rb [hunk 1]
```diff
@@ -19,7 +19,7 @@ class Architecture < ApplicationRecord
   scope :unavailable, -> { where(available: 0) }
 
   #### Validations macros
-  validates :name, uniqueness: true
+  validates :name, uniqueness: { case_sensitive: true }
```
hennevogel, line 22: This should be insensitive

src/api/app/models/bs_request_action.rb
```diff
@@ -32,6 +32,7 @@ class BsRequestAction < ApplicationRecord
   validates :type, presence: true
   validates :type, uniqueness: {
     scope: [:target_project, :target_package, :bs_request_id],
+    case_sensitive: true,
```
hennevogel, line 35: This should be insensitive

src/api/app/models/download_repository.rb
```diff
@@ -4,7 +4,7 @@ class DownloadRepository < ApplicationRecord
   belongs_to :repository
 
   validates :repository, presence: true
-  validates :arch, uniqueness: { scope: :repository_id }, presence: true
+  validates :arch, uniqueness: { scope: :repository_id, case_sensitive: true }, presence: true
```
hennevogel, line 7: This should be insensitive

src/api/app/models/flag.rb
```diff
@@ -19,7 +19,7 @@ class Flag < ApplicationRecord
 
   validate :validate_custom_save
 
-  validates :flag, uniqueness: { scope: [:project_id, :package_id, :architecture_id, :repo] }
+  validates :flag, uniqueness: { scope: [:project_id, :package_id, :architecture_id, :repo], case_sensitive: true }
```
hennevogel, line 22: This should be insensitive

saraycp:
@hennevogel apart from your suggestions: _Architecture#name, BsRequestAction#type, DownloadRepository#arch, Flag#flag_; I also added the following as case insensitive:
_Kiwi::Profile#name, StaticPermission#title_.
Please have another look.

PR #9678, opened 2020-06-05
Move files-related actions to their own controller

DavidKang:
This PR is the first of a series of Pr's to move all the actions related to `file` into `FilesController`.


src/api/app/controllers/webui/packages/files_controller.rb
```diff
@@ -3,10 +3,73 @@ module Packages
     class FilesController < Packages::MainController
       before_action :set_project
       before_action :set_package
+      after_action :verify_authorized, only: :create
```
hennevogel, line 6: Why only if you also authorize :new?

DavidKang: True, thanks :smiley:

dmarcoux:
There's a typo in the PR's title.  It should be `its` instead of `it's`. If you aren't sure why, please read this: https://www.grammarly.com/blog/its-vs-its/

src/api/app/controllers/webui/packages/files_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+          errors << e.message
+        rescue Backend::Error => e
+          errors << Xmlhash::XMLHash.new(error: e.summary)[:error]
+        rescue StandardError => e
+          errors << e.message
+        end
+
+        if errors.empty?
+          message = "The file '#{filename}' has been successfully saved."
+          # We have to check if it's an AJAX request or not
```
dmarcoux, line 51: This comment is quite useless. I would remove it. Same for the one on line 60.

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -359,72 +359,6 @@ def trigger_services
...
-        redirect_back(fallback_location: root_path, error: message)
-        return
-      end
-    end
-
-    status ||= 200
-    render layout: false, status: status, partial: 'layouts/webui/flash', object: flash
-  end
-
   def remove_file
```
dmarcoux, line 362: This action and `view_file` are going to be moved in another PR?

src/api/app/controllers/webui/packages/files_controller.rb
DavidKang: This code was only moved to this controller. I plan to refactor the whole `create` action in another PR.

src/api/app/controllers/webui/package_controller.rb
DavidKang: yes

dmarcoux:
I've just re-read the PR title and using `its` would be correct if we were talking about only one action. However, since we're talking about multiple actions, it should be "Move actions related to files to their own controller`. Sorry for not noticing this earlier :see_no_evil:

PR #9681, opened 2020-06-05
Remove word-break break-all from flash messages and announcement

krauselukas:
The break-all style just breaks the lines between letters. This
makes the annoucements unpleasent to read.
Leaving the default and breaking the lines between the words
is the better option.

**Before**
Screenshot_2020-06-05 openSUSE Build Service

**After**
Screenshot_2020-06-05 Open Build Service

### DIFF
--- a/src/api/app/views/layouts/webui/webui.html.haml
+++ b/src/api/app/views/layouts/webui/webui.html.haml
@@ -51,7 +51,7 @@
           .col-auto.ml-auto#personal-navigation
             = render partial: 'layouts/webui/personal_navigation'
 
-    .container-xxl.sticky-top.flash-and-announcement.text-word-break-all
+    .container-xxl.sticky-top.flash-and-announcement.text-break
       = render partial: 'layouts/webui/status_message_announcement', locals: { current_announcement: @current_announcement }
       #flash= render partial: 'layouts/webui/flash', object: flash
     .modal.fade#modal{ tabindex: '-1', role: 'dialog', aria: { labelledby: 'modalLabel', hidden: true } }


hennevogel:
Screenshot from 2020-06-05 16-42-10

krauselukas:
@hennevogel arghh okay, that I overlooked

hennevogel:
Screenshot from 2020-06-05 16-44-36

hennevogel:
https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/

saraycp:
@krauselukas could you also

src/api/app/assets/stylesheets/webui/texts.scss
```diff
@@ -18,6 +18,10 @@
   word-break: break-all;
 }
 
+.text-word-break-word {
```
DavidKang, line 21: I would use https://getbootstrap.com/docs/4.5/utilities/text/#word-break, instead of adding a custom class.

krauselukas:
> @krauselukas could you also distribute the columns differently? 11 and 1 would be better.
> 
> https://github.com/openSUSE/open-build-service/blob/9c01cd45928c0a997baf4e1bddfa6897462ee108/src/api/app/views/layouts/webui/_status_message_announcement.html.haml#L6
> 
> https://github.com/openSUSE/open-build-service/blob/9c01cd45928c0a997baf4e1bddfa6897462ee108/src/api/app/views/layouts/webui/_status_message_announcement.html.haml#L8

As discussed, lets tackle this in another PR, since this is not the original issue :)

PR #9690, opened 2020-06-08
Project status controller service

vpereira:
Move some repeated code to single fetcher services.

### DIFF
--- a/src/api/app/controllers/webui/projects/status_controller.rb
+++ b/src/api/app/controllers/webui/projects/status_controller.rb
@@ -210,30 +210,19 @@ module Webui
       end
 
       def status_gather_attributes
-        project_status_attributes(@status.keys, 'OBS', 'ProjectStatusPackageFailComment') do |package, value|
+        ProjectStatusControllerService::ProjectStatusFailCommentFinder.call(@status.keys).each do |package, value|
           @status[package].failed_comment = value
         end
 
         return unless @include_versions || @limit_to_old
 
-        project_status_attributes(@status.keys, 'openSUSE', 'UpstreamVersion') do |package, value|
+        ProjectStatusControllerService::OpenSUSEUpstreamVersionFinder.call(@status.keys).each do |package, value|
           @status[package].upstream_version = value
         end
-        project_status_attributes(@status.keys, 'openSUSE', 'UpstreamTarballURL') do |package, value|
+
+        ProjectStatusControllerService::OpenSUSEUpstreamTarballURLFinder.call(@status.keys).each do |package, value|
           @status[package].upstream_url = value
         end
-      end
-
-      def project_status_attributes(packages, namespace, name)
-        ret = {}
-        at = AttribType.find_by_namespace_and_name(namespace, name)
-        return unless at
-
-        attribs = at.attribs.where(package_id: packages)
-        AttribValue.where(attrib_id: attribs).joins(:attrib).pluck('attribs.package_id, value').each do |id, value|
-          yield id, value
-        end
-        ret
       end
 
       def project_status_set_version(p)
--- a/src/api/app/services/project_status_controller_service/attrib_values_finder.rb
+++ b/src/api/app/services/project_status_controller_service/attrib_values_finder.rb
@@ -1,0 +1,31 @@
+module ProjectStatusControllerService
+  class AttribValuesFinder
+    attr_reader :packages, :name, :namespace
+
+    def initialize(packages, namespace, name)
+      @packages = packages
+      @name = name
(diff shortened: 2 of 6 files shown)


src/api/app/services/project_status_controller_service/project_status_failure_command_finder.rb
```diff
@@ -0,0 +1,7 @@
+module ProjectStatusControllerService
+  class ProjectStatusFailureCommandFinder < AttribValuesFinder
```
dmarcoux, line 2: This is the only class which differs from the third parameter passed to the `new` method in `call`. To be consistent with the other classes, it should be `ProjectStatusPackageFailCommentFinder`.

src/api/app/services/project_status_controller_service/attrib_values_finder.rb
```diff
@@ -0,0 +1,31 @@
...
+    attr_accessor :packages, :name, :namespace
+
+    def initialize(packages, namespace, name)
+      @packages = packages
+      @name = name
+      @namespace = namespace
+    end
+
+    def self.call
+      raise NotImplementedError
```
dmarcoux, line 12: We should avoid using `NotImplementedError` whenever we want to document that subclasses need to implement this.

Interesting read on this: http://chrisstump.online/2016/03/23/stop-abusing-notimplementederror/

This is an example from our codebase: https://github.com/openSUSE/open-build-service/blob/70d920d52767236d4202b252be49a0320ff81d85/src/api/app/controllers/webui/request_controller.rb#L320-L322

I've done this recently when refactoring the request controller.

src/api/app/services/project_status_controller_service/project_status_failure_command_finder.rb
dmarcoux, line 2: This is the only class which differs from the third parameter passed to the `new` method in `call`. To be consistent with the other classes, it should be `ProjectStatusPackageFailCommentFinder`.

src/api/app/services/project_status_controller_service/attrib_values_finder.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    def attribute_values
+      AttribValue.where(attrib_id: package_ids_by_attribute_type).joins(:attrib).pluck('attribs.package_id, value')
+    end
+
+    private
+
+    def attribute_type
+      AttribType.find_by_namespace_and_name(@namespace, @name)
```
dmarcoux, line 22: This could be memoized since it might be called twice in `package_ids_by_attribute_type`.

src/api/app/services/project_status_controller_service/attrib_values_finder.rb
vpereira: We have many more examples in our application doing the opposite. So I would assume your solution rather an exception than the rule ;)

BTW  https://refactoring.guru/design-patterns/abstract-factory/ruby/example#lang-features

Even thou I agree that NotImplemented is probably not done for that, I don't like to the let the developer in the dark here.. so are you fine with a custom error class saying to implement it?

src/api/app/services/project_status_controller_service/attrib_values_finder.rb
vpereira: please check it again @dmarcoux

src/api/app/services/project_status_controller_service/attrib_values_finder.rb
dmarcoux: I'm fine with a custom error class which says to implement the abstract method.

src/api/app/controllers/webui/projects/status_controller.rb
```diff
@@ -210,30 +210,19 @@ def status_gather_requests
       end
 
       def status_gather_attributes
-        project_status_attributes(@status.keys, 'OBS', 'ProjectStatusPackageFailComment') do |package, value|
+        ProjectStatusControllerService::ProjectStatusFailCommentFinder.call(@status.keys).each do |package, value|
```
dmarcoux, line 213: This class is missing. I think it wasn't included when you pushed.

src/api/app/services/project_status_controller_service/attrib_values_finder.rb
dmarcoux: The code I linked to in my previous comment should also be changed once we agreed on the solution.

vpereira: @dmarcoux could you please take a look and tell me if is that ok?

PR #9708, opened 2020-06-09
Move to zeitwerk

krauselukas:
Zeitwerk is the new default code loading engine for Rails 6.0
and above. The classic code loading mechanism will be dropped in
the future, so we have to adapt the codebase.


krauselukas:
[Start looking into Zeitwerk](https://trello.com/c/fXyklYB4/1450-start-looking-into-zeitwerk)

krauselukas:
> Please adapt the Inflector to support names like: URLGenerator, XMLParser, MySQLAdapter
> 
> source: https://github.com/fxn/zeitwerk#inflection

@vpereira this got adressed, please check again :)

krauselukas:
The codeclimate issues are all old issue, its only gets triggered since I moved code from A to B. So I would mute them since its not the scope of the PR to resolve them

krauselukas:
I added the `DO NOT MERGE` because I will test this first in our test production instance

krauselukas:
Tested in production mode, works well.

PR #9716, opened 2020-06-10
Change click_button call to follow changes in files_view partial

dmarcoux:
We forgot to change this in #9678


DavidKang:
@dmarcoux, thanks :heart:

PR #9736, opened 2020-06-11
Remove gem coffee-rails

dmarcoux:
We don't have any CoffeeScript files.

There are some CoffeeScript-related files for CodeMirror, but it's a mode for editing CoffeeScript which is completely unrelated.


dmarcoux:

PR #9738, opened 2020-06-11
Replace sanitize with rails-html-sanitizer

dmarcoux:
This allows us to remove the gem "sanitize" since it's now unused.

We don't remove anymore the content of tags which are removed, so the inner text stays. It could be done, but it would involve using `Loofah` which is what `rails-html-sanitizer` builds upon. I prefer to keep it simple and keeping the inner text is fine.

I've used the same tags as `Sanitize::Config::RESTRICTED[:elements] + ['pre']`.


dmarcoux:

dmarcoux:
You can test this by writing some HTML in a comment on the project overview page.

Example: `<pre><code class="xml">blablabla 123</code></pre>`

PR #9739, opened 2020-06-11
Auto-correct RuboCop offenses in app/models for Style/SafeNavigation

dmarcoux:


dmarcoux:
After some discussions in the team, we decided to not apply changes from the `Style/SafeNavigation` cop. I will revert #9727, then create a follow-up PR to disable this cop.

PR #9741, opened 2020-06-11
Do not fail when acknowledging twice

### LINKED ISSUE #9714: [production][status_messages#acknowledge] Validation failed: Status message You have already acknowledged the message
status_messages#acknowledge threw this 5 times
<pre>
Validation failed: Status message You have already acknowledged the message
</pre>
[https://build.opensuse.org/my/status_messages/440](https://build.opensuse.org/my/status_messages/440)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5edb21a30348cc0008d291ef)

danidoni:
Now an exception is raised when acknowledging an already acknowledged status message.

This PR makes OBS ignore a re-acknowledging. OBS will show an error to the user only when the failure is something else.

Fixes #9714

### DIFF
--- a/src/api/app/controllers/webui/status_messages_controller.rb
+++ b/src/api/app/controllers/webui/status_messages_controller.rb
@@ -33,13 +33,16 @@ class Webui::StatusMessagesController < Webui::WebuiController
 
   def acknowledge
     status_message = StatusMessage.find(params[:id])
-    if status_message.acknowledge!
-      RabbitmqBus.send_to_bus('metrics', "user.acknowledged_status_message status_message_id=#{status_message.id}")
-    else
-      flash.now[:error] = "Could not accept status message: #{status_message.errors.full_messages.to_sentence}"
-    end
+    collect_metrics(status_message) if status_message.acknowledge!
+
     respond_to do |format|
       format.js { render controller: 'status_message', action: 'acknowledge' }
     end
+  end
+
+  private
+
+  def collect_metrics(status_message)
+    RabbitmqBus.send_to_bus('metrics', "user.acknowledged_status_message status_message_id=#{status_message.id}")
   end
 end
--- a/src/api/app/models/status_message.rb
+++ b/src/api/app/models/status_message.rb
@@ -28,7 +28,14 @@ class StatusMessage < ApplicationRecord
   end
 
   def acknowledge!
+    return false if acknowledged?
+
     users << User.session!
+    true
+  end
+
+  def acknowledged?
+    users.include?(User.session!)
   end
 
   def self.latest_for_current_user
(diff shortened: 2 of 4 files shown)


src/api/app/models/status_message.rb
```diff
@@ -29,6 +29,16 @@ def delete
 
   def acknowledge!
     users << User.session!
```
krauselukas, line 31:
```suggestion
unless users.include?(User.session!)
    users << User.session!
```

krauselukas: Then you dont need the `elsif` in the controller

danidoni: Yeah I already had this in consideration when thinking on the solution. The point is the code in the suggestion is checking the association against the database every time the user clicks on the `Got it` button. The code in the PR will only make the check against the association when we're not in the happy path.

But thanks anyway @krauselukas !

src/api/app/controllers/webui/status_messages_controller.rb
```diff
@@ -35,7 +35,7 @@ def acknowledge
     status_message = StatusMessage.find(params[:id])
     if status_message.acknowledge!
       RabbitmqBus.send_to_bus('metrics', "user.acknowledged_status_message status_message_id=#{status_message.id}")
-    else
+    elsif !status_message.acknowledged?
       flash.now[:error] = "Could not accept status message: #{status_message.errors.full_messages.to_sentence}"
```
dmarcoux, line 39: The flash won't have any error message when `status_message.acknowledge!` is `false` due to a user already acknowledged. We could add an error to `status_message` so this would be addressed.

src/api/app/models/status_message.rb
```diff
@@ -29,6 +29,16 @@ def delete
...
+  rescue ActiveRecord::RecordInvalid => e
+    return false if users.include?(User.session!)
```
dmarcoux, line 33: This could be `return false if acknowledged?`

src/api/app/controllers/webui/status_messages_controller.rb
dmarcoux: Oh wait, this won't happen because we're checking again if the status message is acknowledged.

src/api/app/models/status_message.rb
danidoni: Yes

src/api/app/models/status_message.rb
vpereira: I don't get it @danidoni , the user should click just once on the "Got it", so I don't see the issue with @krauselukas suggestion

danidoni: Mmm you are both right, I got confused somehow. I'm addressing it now.

danidoni: Check it out again, please!

PR #9742, opened 2020-06-11
Address deprecation warning regarding scoping

### LINKED ISSUE #9442: Deprecation warning class methods will no longer inherit scoping in Rails 6.1
```
W, [2020-04-24T08:31:40.950539 #10697]  WARN -- : [22db7b98-1756-40d5-a325-3609cecb9d71] [10697:315.39] 
```

saraycp:
Having the following warning:

"[...] Class level methods will no longer inherit scoping
from `<scope name>` in Rails 6.1.
To continue using the scoped relation, pass it into the block directly.
To instead access the full set of models, as Rails 6.1 will, use
`<model>.unscoped`, or `<model>.default_scoped` if a model has default scopes.
[...]"

It is addressed by adding .unscoped or .default_scoped when needed.

Closes #9442

This PR addresses two of the three deprecation warnings included in the issue. As soon as this gets approved, I can apply the same fix to other similar existing scopes.

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -75,7 +75,7 @@ class User < ApplicationRecord
   scope :not_deleted, -> { where.not(state: 'deleted') }
   scope :not_locked, -> { where.not(state: 'locked') }
   scope :with_login_prefix, ->(prefix) { where('login LIKE ?', "#{prefix}%") }
-  scope :active, -> { confirmed.or(User.where(state: :subaccount, owner: User.confirmed)) }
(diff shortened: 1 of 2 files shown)


dmarcoux:
8 specs are failing in the `rspec` test suite. It seems to be related to the changes.

saraycp:
> 8 specs are failing in the `rspec` test suite. It seems to be related to the changes.

@dmarcoux this new issue #9748 explains the reason of the failures. As I haven't find a solution for one of the three deprecation warnings yet, I decided to take it out of this commit.

Please have another look.

Issue #9748, opened 2020-06-15
Deprecation Warning on complex scoped relations


saraycp:
As a follow-up of #9442 issue, where:

```
DEPRECATION WARNING: Class level methods will no longer inherit scoping from 
`for_project` in Rails 6.1. To continue using the scoped relation, 
pass it into the block directly. 
To instead access the full set of models, as Rails 6.1 will, use `BsRequest.unscoped`, 
or `BsRequest.default_scoped` if a model has default scopes. 
(called from initialize at /srv/www/obs/api/app/models/bs_request/find_for/base.rb:4)
```

You can get this kind of message when running something like `BsRequest.for_project(Project.first)`  or `Package.dirty_backend_package`.

## Further information

When running  `BsRequest.for_project(Project.first)` it ends up in this [line](https://github.com/openSUSE/open-build-service/blob/27e9536d07393f7a78d12973bb37da9ac3d93270/src/api/app/models/bs_request/find_for/base.rb#L4) where `BsRequest.with_actions` raises a deprecation warning.

Beforehand, it seems the fix is as simple as using `BsRequest.unscoped.with_actions` [in this line](https://github.com/openSUSE/open-build-service/blob/775d8ca6a4f4af96aa04b8d186b865e64a1ab281/src/api/app/models/bs_request/find_for/base.rb#L4).

However, `@relation` becomes a chain of queries inside [app/models/bs_request/find_for/query.rb](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/bs_request/find_for/query.rb) and adding `unscoped` makes it break/leak because it always loose the previously chained queries. [Lines 8 to 10](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/bs_request/find_for/query.rb#L8-L10) are the problematic ones.

Something similar happens with  `Package.dirty_backend_package`, starting in this [line](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/queries/packages_finder.rb#L2).

## Discussions on the Internet related to the topic

- https://github.com/rails/rails/pull/32380
- https://github.com/rails/rails/pull/35280
- https://www.redmine.org/issues/32911
- https://www.reddit.com/r/rails/comments/cok18p/deprecation_warning_scopes/
- https://github.com/rails/rails/issues/38881

PR #9750, opened 2020-06-15
Prevent deprecation warnings on scopes

saraycp:
Other scope definitions are affected by the deprecation warnings related to scoping.

Message:

```
```
Follow-up of #9742

Co-authored-by:  Quang Tran <quang.tran@suse.com>

PS: this skips BsRequest and Package scopes, there is an specific issue for them: #9748

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -77,7 +77,7 @@ class User < ApplicationRecord
   scope :with_login_prefix, ->(prefix) { where('login LIKE ?', "#{prefix}%") }
   scope :active, -> { confirmed.or(User.unscoped.where(state: :subaccount, owner: User.unscoped.confirmed)) }
   scope :staff, -> { joins(:roles).where('roles.title = ?', 'Staff') }
-  scope :not_staff, -> { where.not(id: User.staff.pluck(:id)) }
+  scope :not_staff, -> { where.not(id: User.unscoped.staff.pluck(:id)) }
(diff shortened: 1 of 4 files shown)


src/api/app/models/user.rb
```diff
@@ -77,7 +77,7 @@ class User < ApplicationRecord
   scope :with_login_prefix, ->(prefix) { where('login LIKE ?', "#{prefix}%") }
   scope :active, -> { confirmed.or(User.unscoped.where(state: :subaccount, owner: User.unscoped.confirmed)) }
   scope :staff, -> { joins(:roles).where('roles.title = ?', 'Staff') }
-  scope :not_staff, -> { where.not(id: User.staff.pluck(:id)) }
+  scope :not_staff, -> { where.not(id: User.unscoped.staff.pluck(:id)) }
```
danidoni, line 80: Shouldn't this be `default_scoped`?

krauselukas: But this model does not have a default scope, so I think no

saraycp: There isn't any `default_scope`defined in User model, so `unscoped` is correct here.

danidoni: Okay, I understood the `To instead access the full set of models, as Rails 6.1 will, use `BsRequest.unscoped`, 
or `BsRequest.default_scoped` if a model has default scopes.` as `either use unscoped or default_scoped`

PR #9756, opened 2020-06-15
Introduce cache versioning

eduardoj:
Before cache versioning, `cache_key` contained the key and the version.

Introducing cache versioning makes `cache_key` to contain only the key. In order to have the same content as before, we have to use `cache_key_with_version`.

This is a follow up of #9392.

Have a look at this [blog entry](https://blog.bigbinary.com/2019/08/06/rails-adds-support-for-recyclable-cache-keys.html) as reference.

### DIFF
--- a/src/api/config/application.rb
+++ b/src/api/config/application.rb
@@ -85,8 +85,7 @@ module OBSApi
 
     config.action_controller.perform_caching = true
 
-    # Don't use cache versioning for now
-    config.active_record.cache_versioning = false
+    config.active_record.cache_versioning = true
     config.active_record.collection_cache_versioning = false
 
     config.action_controller.action_on_unpermitted_parameters = :raise
--- a/src/api/app/models/binary_release.rb
+++ b/src/api/app/models/binary_release.rb
@@ -184,7 +184,7 @@ class BinaryRelease < ApplicationRecord
   end
 
   def to_axml(_opts = {})
-    Rails.cache.fetch("xml_binary_release_#{cache_key}") { render_xml }
+    Rails.cache.fetch("xml_binary_release_#{cache_key_with_version}") { render_xml }
   end
 
   def identical_to?(binary_hash)
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -365,15 +365,15 @@ class BsRequest < ApplicationRecord
 
   def to_axml(opts = {})
     if opts[:withfullhistory]
-      Rails.cache.fetch("xml_bs_request_fullhistory_#{cache_key}") do
+      Rails.cache.fetch("xml_bs_request_fullhistory_#{cache_key_with_version}") do
         render_xml(withfullhistory: 1)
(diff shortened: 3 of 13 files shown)


eduardoj:
[Enable cache versioning (again)](https://trello.com/c/l7nCB7Cf/1453-enable-cache-versioning-again)

src/api/config/application.rb
```diff
@@ -85,8 +85,8 @@ class Application < Rails::Application
 
     config.action_controller.perform_caching = true
 
-    # Don't use cache versioning for now
-    config.active_record.cache_versioning = false
+    # Use cache versioning
```
dmarcoux, line 88: [nitpick]

I would remove this comment since it's pretty obvious.

dmarcoux:
Just to be sure, there are 3 instances of `User.session!.cache_key`. Should these be also changed to `User.session!.cache_key_with_version`?

- https://github.com/openSUSE/open-build-service/blob/bbdab04f9c286cdcfec0bad546419f654845e115/src/api/app/models/package.rb#L128
- https://github.com/openSUSE/open-build-service/blob/bbdab04f9c286cdcfec0bad546419f654845e115/src/api/app/views/webui/users/tasks/index.html.haml#L7
- https://github.com/openSUSE/open-build-service/blob/bbdab04f9c286cdcfec0bad546419f654845e115/src/api/app/views/webui/users/tasks/index.html.haml#L22

src/api/config/application.rb
eduardoj: Done.

eduardoj:
> Just to be sure, there are 3 instances of `User.session!.cache_key`. Should these be also changed to `User.session!.cache_key_with_version`?

True! Thanks!

hennevogel:
Why?

eduardoj:
> Why?

Why not?

hennevogel:
> Why not?

Because there are no problems with caching?

eduardoj:
> Because there are no problems with caching?

True, there are currently no problems with caching. Cache versioning is an improved caching concept, introduced in Rails 5.2, enabled by default in Rails 6. Citing the blog post referencing in de PR description "Large applications frequently need to invalidate their cache because cache store has limited memory. We can optimize cache storage and minimize cache miss using recyclable cache keys."

So this pull request doesn't change the current behaviour, but enables cache versioning either for future use cases, or to change current cases.

hennevogel:
> So this pull request doesn't change the current behaviour, but enables cache versioning either for future use cases, or to change current cases.

You should explain this  in ce8daaee58a6f4bfa834cff95ed5e36ca8452fc7

eduardoj:
> > So this pull request doesn't change the current behaviour, but enables cache versioning either for future use cases, or to change current cases.
> 
> You should explain this in [ce8daae](https://github.com/openSUSE/open-build-service/commit/ce8daaee58a6f4bfa834cff95ed5e36ca8452fc7)

Description of the commit changed. Please review again.

PR #9757, opened 2020-06-15
Use distinct on activerecord instead of uniq on Array

vpereira:
Instead of `map.uniq` in the array level, use DISTINCT on name to reduce the architectures in the database level.

### DIFF
--- a/src/api/app/controllers/webui/main_controller.rb
+++ b/src/api/app/controllers/webui/main_controller.rb
@@ -31,17 +31,12 @@ class Webui::MainController < Webui::WebuiController
 
   def gather_busy
     busy = []
-    archs = Architecture.where(available: 1).map(&:worker).uniq
-    archs.each do |arch|
-      starttime = Time.now.to_i - 168.to_i * 3600
-      rel = StatusHistory.where("time >= ? AND \`key\` = ?", starttime, 'building_' + arch)
-      values = rel.pluck(:time, :value).collect! { |time, value| [time.to_i, value.to_f] }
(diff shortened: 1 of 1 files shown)


src/api/app/controllers/webui/main_controller.rb
```diff
@@ -31,17 +31,12 @@ def index
 
   def gather_busy
     busy = []
-    archs = Architecture.where(available: 1).map(&:worker).uniq
-    archs.each do |arch|
-      starttime = Time.now.to_i - 168.to_i * 3600
-      rel = StatusHistory.where("time >= ? AND \`key\` = ?", starttime, 'building_' + arch)
-      values = rel.pluck(:time, :value).collect! { |time, value| [time.to_i, value.to_f] }
+    starttime = Time.now.to_i - 168.to_i * 3600
```
hennevogel, line 34:
```suggestion
    starttime = (Time.now - 7.days).to_i
```

```ruby
irb(main):011:0> (Time.now - 7.days).to_i
=> 1591645202
irb(main):012:0> Time.now.to_i - 168.to_i * 3600
=> 1591645216
irb(main):013:0> (Time.now - 7.days).to_i
=> 1591645217
irb(main):014:0> Time.at(Time.now.to_i - 168.to_i * 3600)
=> 2020-06-08 19:40:45 +0000
irb(main):015:0> Time.at((Time.now - 7.days).to_i)
=> 2020-06-08 19:40:50 +0000
```
magic numbers :)

vpereira: yup, much better, thanks!

PR #9758, opened 2020-06-15
Fix Rubocop complains for ConsistencyCheckJob

vpereira:
A PR was merged before rebased with master, where some rubocop cops were
enabled


src/api/.rubocop.yml
```diff
@@ -115,6 +117,11 @@ RSpec/FilePath:
   Exclude:
     - 'spec/controllers/webui/package_controller/binaries.spec'
 
+# We disabled this cop because named subject isn't a better construction then
+# using let
+RSpec/NamedSubject:
+  Enabled: false
```
dmarcoux, line 123: If the cop is now disabled, its TODOs have to be removed: https://github.com/openSUSE/open-build-service/blob/bbdab04f9c286cdcfec0bad546419f654845e115/src/api/.rubocop_todo.yml#L873-L974

src/api/.rubocop.yml
```diff
@@ -115,6 +117,11 @@ RSpec/FilePath:
   Exclude:
     - 'spec/controllers/webui/package_controller/binaries.spec'
 
+# We disabled this cop because named subject isn't a better construction then
+# using let
+RSpec/NamedSubject:
```
dmarcoux, line 122: If the cop is now disabled, its TODOs have to be removed: https://github.com/openSUSE/open-build-service/blob/bbdab04f9c286cdcfec0bad546419f654845e115/src/api/.rubocop_todo.yml#L873-L974

vpereira: ok let me push it too

vpereira: done

src/api/app/services/consistency_check_job_service/project_meta_checker.rb
```diff
@@ -28,8 +28,8 @@ def backend_meta
     end
 
     # transform hash to array, compare it and transform it back to hash
-    def hash_diff(a, b)
-      difference = a.size > b.size ? a.to_a - b.to_a : b.to_a - a.to_a
+    def hash_diff(array_a, array_b)
+      difference = array_a.size > array_b.size ? array_a.to_a - array_b.to_a : array_b.to_a - array_a.to_a
```
vpereira, line 32: I really don't see any reasoning why change it. For one-liner, same rule as for blocks should be used.. to tricky rubocop didn't make the code cleaner.

src/api/.rubocop.yml
```diff
@@ -115,6 +117,11 @@ RSpec/FilePath:
   Exclude:
     - 'spec/controllers/webui/package_controller/binaries.spec'
 
+# We disabled this cop because named subject isn't a better construction then
```
saraycp, line 120:
```suggestion
# We disabled this cop because named subject isn't a better construction than
```

src/api/app/services/consistency_check_job_service/project_meta_checker.rb
saraycp: The only difference is the naming of the variables, right?

vpereira: yes

PR #9764, opened 2020-06-16
validate OBS_WORKER_SECURITY_LEVEL as variable when use it for hostlabel

banuchka:
Check **OBS_WORKER_SECURITY_LEVEL** as variable before append empty value to worker hostlabels


DavidKang:
Hi @banuchka, thanks for your contribution!

dmarcoux:
Regarding the issues in the `ci/circleci: checkout_code` check, rebasing the branch on master should fix them. We had some issues directly on master which are now addressed.

dist/obsworker
```diff
@@ -74,7 +74,7 @@ if [ -n "$OBS_WORKER_WIPE_AFTER_BUILD" ]; then
     OBS_WIPE_AFTER_BUILD="--wipeafterbuild"
 fi
 
-if [ -n "OBS_WORKER_SECURITY_LEVEL" ]; then
+if [ -n "$OBS_WORKER_SECURITY_LEVEL" ]; then
```
vpereira, line 77: @dmarcoux that's the kind issue that ShellCheck would catch, therefore the PR https://github.com/openSUSE/open-build-service/pull/9388 should be merged

dmarcoux: I agree that ShellCheck is useful. However, it's only targeting one file in that PR. It's better than nothing, but still quite limited.

PR #9768, opened 2020-06-16
Force association chain reload

### LINKED ISSUE #9763: "Show link diff" button missing
Recently,  the show link diff button (which usually appears below "Links to */*") has disappeared from the options at build.opensuse.org. 

I was part of the beta program when I noticed it, however leaving the beta program has no effect and the button still doesn't appear.

danidoni:
Looks like when we
removed the `foreign_key: :package_id` from the
`has_one :backend_package`
here (https://github.com/openSUSE/open-build-service/pull/9701), we hit an Rails/ActiveRecord
bug (https://github.com/rails/rails/issues/38709).

This PR is a workaround.

Until we find another workaround or the issue gets fixed we are gonna
hit a bit of a penalty due to the association reload.

As an alternative we can revert https://github.com/openSUSE/open-build-service/pull/9701.

Fixes #9763

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -797,10 +797,13 @@ class Webui::PackageController < Webui::WebuiController
   def set_linkinfo
     return unless @package.is_link?
 
-    linked_package = @package.backend_package.links_to
+    # https://github.com/rails/rails/issues/38709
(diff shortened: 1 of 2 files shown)


danidoni:
I'll close this one, see https://github.com/openSUSE/open-build-service/pull/9772

PR #9769, opened 2020-06-16
Adapt custom routes helper to zeitwerk conventions

krauselukas:
In order to have the routes splitted into api and webui
routes, a couple of helper classes got created and added
to the eager_load_paths in the past. The folder that contains
those helper classes, contains the route definitions as well.
This has the side effect, that zeitwerk derives constants
from those (so 'ApiRoutes' and 'WebuiRoutes'). But there
is no module/class that corresponds to those constant names.

Also the custom routes helper defining multiple classes
in one file which doesn't work with Zeitwerk anymore (it doesn't
break the codebase, since the file that contains the declarations
was still required manually in the routes.rb).

In order to fix it, the custom routes helper got moved to the
'app/lib' folder, which gets eager_loaded/auto_loaded
by default and classes/files got splited accordingly.


dmarcoux:
I've just pushed a commit for suggestions. I figured it would be easier than having to refer to files which weren't previously changed.

PR #9774, opened 2020-06-17
Clean up notifications with Notification.cleanup

dmarcoux:
Instead of keeping at most 10 notifications for groups and active users, we keep notifications if they were created up to 3 months ago.

There are no specs for this job, but it's one line and that one line is covered in the specs for the Notification model.

### DIFF
--- a/src/api/app/jobs/cleanup_notifications.rb
+++ b/src/api/app/jobs/cleanup_notifications.rb
@@ -1,5 +1,5 @@
 class CleanupNotifications < ApplicationJob
   def perform
-    Notification::RssFeedItem.cleanup
+    Notification.cleanup
   end
 end


dmarcoux:

hennevogel:
Why do we delete notifications after 3 months? :)

hennevogel:
```
Loading production environment (Rails 6.0.3.1)
irb(main):001:0> Notification.count
=> 7731

dmarcoux:
We could keep them for a longer timespan or even forever. It's maybe a bit too much to never delete them since this will just keep on growing and growing. It might be premature optimization to even have a cleanup job for this, but at the same time, a notification which is a few months old doesn't bring much value...

PR #9777, opened 2020-06-17
Remove Notification::RssFeedItem

dmarcoux:
We don't need this class. The only difference from its parent is the `cleanup` class method and it isn't used anymore.


dmarcoux:

danidoni:
Fix the specs and then LGTM!

PR #9783, opened 2020-06-18
Add method .projects for Backend::Api::Search

vpereira:
We need a better way to search for a set of projects on backend. Therefore the `Backend::Api::Search` now supports it. passing an array of names to the method `.projects` it will create a xpath and search for it on backend.

### DIFF
--- a/src/api/spec/lib/backend/api/search_spec.rb
+++ b/src/api/spec/lib/backend/api/search_spec.rb
@@ -1,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe Backend::Api::Search, vcr: true do
+  describe '.projects' do
+    context 'with projects' do
+      let(:project_names) { create_list(:project, 2) { |prj, i| prj.name = "foo_#{i}" }.map(&:name) }
+
+      subject { Backend::Api::Search.projects(project_names) }
+
+      it { expect(Nokogiri::XML(subject).xpath('//collection//project').count).to eq(2) }
+    end
+
+    context 'no projects' do
+      subject { Backend::Api::Search.projects([]) }
+
+      it { expect(Nokogiri::XML(subject).xpath('//collection//project').count).to eq(0) }
+    end
+
+    context 'with unexistent projects' do
+      subject { Backend::Api::Search.projects(['xaa', 'xee']) }
+
+      it { expect(Nokogiri::XML(subject).xpath('//collection//project').count).to eq(0) }
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


src/api/spec/lib/backend/api/search_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'rails_helper'
+
+RSpec.describe Backend::Api::Search, vcr: true do
+  describe '.projects' do
+    context 'with projects' do
+      # rubocop:disable RSpec/LetSetup
+      let!(:projects) { 2.times.collect { |i| create(:project, name: "foo_#{i}") } }
```
dmarcoux, line 7: If this isn't called, it should be a in `before` block, right?

src/api/spec/lib/backend/api/search_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'rails_helper'
+
+RSpec.describe Backend::Api::Search, vcr: true do
+  describe '.projects' do
+    context 'with projects' do
+      # rubocop:disable RSpec/LetSetup
```
dmarcoux, line 6: I would replace those 2 `let` with:

`let(:project_names) {  2.times.collect { |i| create(:project, name: "foo1_#{i}") }.map(&:name) }`

or

`let(:project_names) { create_list(:project, 2) { |prj, i| prj.name = "foobar_#{i}"}.map(&:name) }`

vpereira: true, done!

src/api/spec/lib/backend/api/search_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe Backend::Api::Search, vcr: true do
+  describe '.projects' do
+    context 'with projects' do
+      let(:project_names) { 2.times.collect { |i| create(:project, name: "foo_#{i}") }.map(&:name) }
```
hennevogel, line 6: https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#building-or-creating-multiple-records

dmarcoux: It's one of the suggestions from my [previous comment](https://github.com/openSUSE/open-build-service/pull/9783#discussion_r442133313). @vpereira told that it didn't work with `create_list` due to the backend. I wonder if it's because it should be `let!` instead of `let`.

PR #9784, opened 2020-06-18
Stop enqueueing CleanupNotifications job

dmarcoux:
Following the discussion we had in #9774 and offline, we agreed that this job isn't needed since there aren't many notifications.


dmarcoux:

PR #9788, opened 2020-06-18
Ignore type column for Notification model

dmarcoux:
We are going to remove this column from the notifications table. This is to avoid a downtime.


dmarcoux:
[(S) Drop `Notification.type`](https://trello.com/c/FvQGaHYS/1460-s-drop-notificationtype)

dmarcoux:
Oh right, we ignore the column, but then it doesn't a default value so it's obviously failing. :see_no_evil:

I'll have to rework my approach!

Issue #9779, opened 2020-06-18
database migration still broken


M0ses:
Here from the logs:

```
rails aborted!
StandardError: An error has occurred, all later migrations canceled:

Mysql2::Error: Table 'api_production.announcements' doesn't exist
/srv/www/obs/api/db/data/20200528135241_dump_announcements_into_status_messages.rb:13:in `up'
bin/rails:4:in `require'
bin/rails:4:in `<main>'

Caused by:
ActiveRecord::StatementInvalid: Mysql2::Error: Table 'api_production.announcements' doesn't exist
/srv/www/obs/api/db/data/20200528135241_dump_announcements_into_status_messages.rb:13:in `up'
bin/rails:4:in `require'
bin/rails:4:in `<main>'

Caused by:
Mysql2::Error: Table 'api_production.announcements' doesn't exist
/srv/www/obs/api/db/data/20200528135241_dump_announcements_into_status_messages.rb:13:in `up'
bin/rails:4:in `require'
bin/rails:4:in `<main>'
Tasks: TOP => db:migrate:with_data
(See full trace by running task with --trace)
```

obs version (current master):

```
obs-server:/srv/www/obs/api # rpm -q obs-api
obs-api-2.11~alpha.20200617T224243.916617bcb4-lp151.10499.1.noarch
```

This is a follow-up to #9762

M0ses:
Please let me know if you need a VM for testing/debugging

hennevogel:
Again this script of yours is doing something funky that is not supported by rails...

M0ses:
@hennevogel  

* "this script of yours" - this is the script that does the setup of our appliance since I joined the team!
* "rake db:migrate:with_data" should be reentrant as it was since ages (for me at least since 2015)

lrupp:
> Table 'api_production.announcements' doesn't exist

Which script is supposed to create this table?

hennevogel:
The `setup-appliance.sh` script :point_right:  https://github.com/openSUSE/open-build-service/blob/master/dist/setup-appliance.sh#L280

hennevogel:
Sounds like this did not run. The other issues was also in this area. I'm telling you. There is a script problem :smile_cat:

M0ses:
Different VM / same problem:

```
Jun 18 14:25:05 obs-server-vdi setup-appliance.sh[1784]: Initialize OBS api database (first time only)
[...]

PR #9790, opened 2020-06-18
Remove Rake::Task monkey patch

dmarcoux:
`redefine_task` was used to redefine the `db:setup` task. We prefer to be explicit about having to run `db:structure:load:with_data db:seed` and to avoid monkey patching Rake::Task.

While looking into #9779 and talking with @hennevogel and @saraycp about it, we found this nice :monkey: patch.

Sadly, we cannot use `db:structure:load:with_data`. For some obscure reasons, specs fail due to missing seeds, even if seeds are added after calling `db:structure:load:with_data`. We can still remove the monkey patch.

~~To take data migrations into account when setting up a new database, we need to run `db:structure:load:with_data`, not `db:structure:load`. This will prevent issues like #9779 if somehow data migrations are ran against a new database. It shouldn't happen, but if it does, it's simply going to do nothing since the [data_schema.rb](https://github.com/openSUSE/open-build-service/blob/d42d8310aa25b18637b77831e6816a9f4adf837d/src/api/db/data_schema.rb) is also loaded when creating the database.~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ aliases:
 
   - &create_test_db
     name: Create database
-    command: cd src/api; bundle exec rake db:create db:setup RAILS_ENV=test
+    command: cd src/api; bundle exec rake db:create db:structure:load db:seed RAILS_ENV=test
 
   - &bootstrap_old_test_suite
     name: Setup application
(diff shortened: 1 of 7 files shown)


dmarcoux:
Sadly, we cannot use `db:structure:load:with_data`. For some obscure reasons, specs fail due to missing seeds, even if seeds are added after calling `db:structure:load:with_data`.

PR #9792, opened 2020-06-19
Mark notifications read or unread asynchronously

danidoni:
Mark notifications read or unread asynchronously.

Now, when you are reviewing your notifications and you tick one as read, the whole page reloads and you lose the point in the page you were working on. This change allows you to keep working in place.


danidoni:
[(M) Avoid full page reloads for marking Notifications as read/unread](https://trello.com/c/wVvdPjua/1466-m-avoid-full-page-reloads-for-marking-notifications-as-read-unread)

PR #9794, opened 2020-06-19
Switch to apparition capybara driver

hennevogel:
This driver is being developed by the current developer of
Capybara and will attempt to keep up to date with new Capybara releases.
We expect this to be more stable.

Esecially as it uses CDP to communicate with Chrome, thereby obviating the need
for chromedriver.


krauselukas:
https://github.com/openSUSE/open-build-service/pull/9794/files#diff-3017e4be0352c93e16a87addfa46ca38L5
Would it make sense to also change the `javascript_driver` to `apparition? not sure what the `:desktop` setting actually means

hennevogel:
@krauselukas it's the driver registered below.

krauselukas:
> @krauselukas it's the driver registered below.

Ahh okay, haven't seen it, okay :)

hennevogel:
Maybe later...

PR #9795, opened 2020-06-19
Run brakeman in checkout_code CI step

hennevogel:
brakeman is a linter, we should run it where we run all the other linters too.
Firing up a container is too expensive just for this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,6 +83,9 @@ jobs:
             - src/api/tmp/rubocop_cache_rails_dir
       - run: rm -rf src/api/tmp/rubocop*
       - run:
+          name: Run brakeman
+          command: sudo gem install --no-format-executable brakeman; brakeman --rails6 -p src/api
+      - run:
           name: Setup application
           command: cd src/api; bundle exec rake dev:prepare assets:precompile RAILS_ENV=test FORCE_EXAMPLE_FILES=1
       - persist_to_workspace:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -82,6 +82,9 @@ jobs:
             - src/api/tmp/rubocop_cache_root_dir
             - src/api/tmp/rubocop_cache_rails_dir
       - run: rm -rf src/api/tmp/rubocop*
+      - run:
+          name: Run brakeman
+          command: sudo gem install --no-format-executable brakeman; cd src/api; brakeman --rails6
```
dmarcoux, line 87: I've looked at the brakeman documentation and the option `rails6` is not documented. It runs `brakeman` with some Rails 6 specific checks?

Brakeman can receive the path as an option with `-p / --path`. Doing `cd src/api` is also fine.

eduardoj: Shouldn't we include running brakeman as another task/command inside the [`dev:lint` task](https://github.com/openSUSE/open-build-service/blob/24145637ba72265db98f118e424e0bfd2d9e7e5a/src/api/lib/tasks/dev.rake#L63)?

hennevogel: > Rails 6 specific checks

```
brakeman --help|grep rails6
    -6, --rails6                     Force Rails 6 mode
```

> Shouldn't we include running brakeman as another task

Can't as it's not there normally, only on CI.

hennevogel:
The brakeman CI task is missing for a reason :wink:

hennevogel:
merging and disabling brakeman

PR #9796, opened 2020-06-19
Run spider test only after functional tests

hennevogel:
Only run the spider test of everything else is green. Will make this test run
less frequent and the whole cycle faster a bit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -310,7 +310,9 @@ workflows:
             - checkout_code
       - spider:
           requires:
-            - checkout_code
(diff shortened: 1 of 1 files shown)


vpereira:
@hennevogel because the spider fails really fast, for example, when we have unreachable route due a 500 error.. isn't that better to just run the all the other tests IF the spider ran successfully?

hennevogel:
@vpereira it doesn't fail fast and it also doesn't really tell you what is going on if it fails. It's more like a last line of defense :smile:

PR #9801, opened 2020-06-22
Ignore type column for Notification model

dmarcoux:
We are going to remove this column from the notifications table. This is to avoid a downtime.


dmarcoux:
[(S) Drop `Notification.type`](https://trello.com/c/FvQGaHYS/1460-s-drop-notificationtype)

dmarcoux:
Ignoring columns in a model with STI is not supported. See https://github.com/rails/rails/issues/34344

We will have to drop this column in our maintenance window on Thursday this week.

PR #9803, opened 2020-06-22
Add counter to notification filter links

krauselukas:
Currently we only show the amount of notifications,
for the project filter links.
In order to enable the user to get a quick overview
about the amount of pending notifications,
we now show a badge with the amount also for notifications
related to review, comments, request and overall unread.

### DIFF
--- a/src/api/app/controllers/webui/users/notifications_controller.rb
+++ b/src/api/app/controllers/webui/users/notifications_controller.rb
@@ -14,6 +14,7 @@ class Webui::Users::NotificationsController < Webui::WebuiController
                      end
 
     @projects_for_filter = projects_for_filter
+    @notifications_count = notifications_count
 
     @notifications = params['show_all'] ? show_all : @notifications.page(params[:page])
   end
@@ -61,5 +62,10 @@ class Webui::Users::NotificationsController < Webui::WebuiController
            .where(notifications: { subscriber: User.session, delivered: false, web: true })
            .order('name desc').group(:name).count # this query returns a sorted-by-name hash like { "home:b" => 1, "home:a" => 3  }
            .sort_by(&:last).reverse.to_h # this sorts the hash by amount: { "home:a" => 3, "home:b" => 1 }
+  end
+
+  def notifications_count
+    counted_notifications = NotificationsFinder.new(User.session.notifications.for_web).unread.group(:notifiable_type).count
+    counted_notifications.merge!('unread' => User.session.unread_notifications)
   end
 end
--- a/src/api/app/helpers/webui/notification_helper.rb
+++ b/src/api/app/helpers/webui/notification_helper.rb
@@ -9,17 +9,10 @@ module Webui::NotificationHelper
     end
   end
 
-  def filter_by_type_link(link_text, filter_item)
-    link_to(link_text, my_notifications_path(filter_item), class: filter_css(filter_item))
-  end
-
-  def filter_by_project_link(link_text, amount, filter_item)
-    badge_color = notification_filter_active?(filter_item) ? 'badge-light' : 'badge-primary'
+  def filter_notification_link(link_text, amount, filter_item)
     link_to(my_notifications_path(filter_item), class: filter_css(filter_item)) do
-      capture do
-        concat(link_text)
-        concat(tag.span(amount, class: "badge #{badge_color} align-text-top ml-2"))
-      end
+      concat(link_text)
+      concat(tag.span(amount, class: "badge #{badge_color(filter_item)} align-text-top ml-2")) if amount && amount.positive?
     end
   end
 
@@ -39,5 +32,9 @@ module Webui::NotificationHelper
     else
       filter_item[:type] == 'unread'
     end
+  end
+
+  def badge_color(filter_item)
+    notification_filter_active?(filter_item) ? 'badge-light' : 'badge-primary'
   end
 end
--- a/src/api/app/views/webui/users/notifications/index.html.haml
+++ b/src/api/app/views/webui/users/notifications/index.html.haml
@@ -13,17 +13,17 @@
 
       .card-body.collapse#filters
         .row.list-group-flush
-          = filter_by_type_link('Unread', type: 'unread')
-          = filter_by_type_link('Read', type: 'read')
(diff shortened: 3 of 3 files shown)


src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -62,4 +63,8 @@ def projects_for_filter
            .order('name desc').group(:name).count # this query returns a sorted-by-name hash like { "home:b" => 1, "home:a" => 3  }
            .sort_by(&:last).reverse.to_h # this sorts the hash by amount: { "home:a" => 3, "home:b" => 1 }
   end
+
+  def notifications_count
+    Notification.where(subscriber: User.session, web: true, delivered: false).group(:notifiable_type).count
```
dmarcoux, line 68: I suggest using [NotificationsFinder](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/queries/notifications_finder.rb) to re-use what we already have.

This would then be:

`NotificationsFinder.new(User.session.notifications.for_web).unread.group(:notifiable_type).count`

krauselukas:
[(S) All Notification filters should show the number of Notifications that would show. Like the Project filter does.](https://trello.com/c/12bMfc1w/1463-s-all-notification-filters-should-show-the-number-of-notifications-that-would-show-like-the-project-filter-does)

src/api/app/controllers/webui/users/notifications_controller.rb
krauselukas: @dmarcoux I addressed it, please check again :) also @DavidKang as you suggested, I'm adding the amount of unread notifications to the hash now, instead of doing a separate call in the view

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -9,17 +9,10 @@ def link_to_all
...
 
-  def filter_by_type_link(link_text, filter_item)
-    link_to(link_text, my_notifications_path(filter_item), class: filter_css(filter_item))
-  end
-
-  def filter_by_project_link(link_text, amount, filter_item)
-    badge_color = notification_filter_active?(filter_item) ? 'badge-light' : 'badge-primary'
+  def filter_notification_link(link_text, amount, filter_item)
     link_to(my_notifications_path(filter_item), class: filter_css(filter_item)) do
-      capture do
```
krauselukas, line 19: Not sure why we use the `capture` block here. Maybe there is something I'm not aware of, but from my point of view this is not necessary

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -9,17 +9,10 @@ def link_to_all
     end
   end
 
-  def filter_by_type_link(link_text, filter_item)
-    link_to(link_text, my_notifications_path(filter_item), class: filter_css(filter_item))
-  end
-
-  def filter_by_project_link(link_text, amount, filter_item)
-    badge_color = notification_filter_active?(filter_item) ? 'badge-light' : 'badge-primary'
+  def filter_notification_link(link_text, amount, filter_item)
```
krauselukas, line 12: Codeclimate reported a high cognitive complexity here, so I splitted it up a bit

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -40,4 +33,8 @@ def notification_filter_active?(filter_item)
       filter_item[:type] == 'unread'
     end
   end
+
+  def determine_badge_color(filter_item)
```
danidoni, line 37: I would use just `badge_color`.

src/api/app/helpers/webui/notification_helper.rb
danidoni: I don't see any Erb code being rendered inside the capture block so I also think this is not really needed.

src/api/app/helpers/webui/notification_helper.rb
krauselukas: done

krauselukas: yes, now you said it, the naming felt bad :D

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -62,4 +63,9 @@ def projects_for_filter
            .order('name desc').group(:name).count # this query returns a sorted-by-name hash like { "home:b" => 1, "home:a" => 3  }
            .sort_by(&:last).reverse.to_h # this sorts the hash by amount: { "home:a" => 3, "home:b" => 1 }
   end
+
+  def notifications_count
+    counted_notifications = NotificationsFinder.new(User.session.notifications.for_web).unread.group(:notifiable_type).count
+    counted_notifications.merge!('unread' => User.session.unread_notifications)
```
dmarcoux, line 69: Since `'unread'` is always present, this has the side-effect of displaying a badge even when there are no unread notifications.

unnecessary-badge

krauselukas: Ah right, going to address this

krauselukas: I'm checking now for `zero?` in the helper @dmarcoux

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -9,17 +9,10 @@ def link_to_all
...
-  def filter_by_project_link(link_text, amount, filter_item)
-    badge_color = notification_filter_active?(filter_item) ? 'badge-light' : 'badge-primary'
+  def filter_notification_link(link_text, amount, filter_item)
     link_to(my_notifications_path(filter_item), class: filter_css(filter_item)) do
-      capture do
-        concat(link_text)
-        concat(tag.span(amount, class: "badge #{badge_color} align-text-top ml-2"))
-      end
+      concat(link_text)
+      concat(tag.span(amount, class: "badge #{badge_color(filter_item)} align-text-top ml-2")) if amount && !amount.zero?
```
dmarcoux, line 15:
```suggestion
      concat(tag.span(amount, class: "badge #{badge_color(filter_item)} align-text-top ml-2")) if amount && amount.positive?
```

krauselukas: !not_done ;) @dmarcoux

PR #9806, opened 2020-06-22
Do not run data migrations in db:verify

dmarcoux:
Data migrations have nothing to do with `db/structure.sql`

This avoids issues like in #9801 right now.

### DIFF
--- a/src/api/db/migrate/001_initial_database.rb
+++ b/src/api/db/migrate/001_initial_database.rb
@@ -269,6 +269,8 @@ class InitialDatabase < ActiveRecord::Migration[4.2]
       t.string   'admin_email',                                     default: 'unconfigured@openbuildservice.org'
     end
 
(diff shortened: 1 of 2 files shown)


dmarcoux:
This is failing because the `data_migrations` is not created by a migration, but from `db:migration:with_data`.

See upstream:
1. https://github.com/ilyakatz/data-migrate/blob/e3f7ffa9fd58a44958df6e766cdb18ce98fba663/tasks/databases.rake#L7
2. https://github.com/ilyakatz/data-migrate/blob/7b36ac50dda0ae932ac54c7ba6f7b27eeb25fef8/lib/data_migrate/data_migrator.rb#L62

This has been discussed upstream in https://github.com/ilyakatz/data-migrate/issues/105.

dmarcoux:
After looking more into this, it's going to be a rabbit hole and I'm not going into this now.

PR #9808, opened 2020-06-22
Drop mailcatcher from Procfile

krauselukas:
Mailcatcher is not used anymore in the dev environment

Last usage got removed by https://github.com/openSUSE/open-build-service/pull/9807


vpereira:
@krauselukas can you drop it from the image too? Looks like it is installed in the containers

krauselukas:
> @krauselukas can you drop it from the image too? Looks like it is installed in the containers

@vpereira yes try to find where this gets installed

krauselukas:
@vpereira dropped in the image by this https://build.opensuse.org/request/show/816405

PR #9813, opened 2020-06-23
Simplify rake db:* calls

dmarcoux:
Following up on the changes from #9790

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ aliases:
 
   - &create_test_db
     name: Create database
-    command: cd src/api; bundle exec rake db:create db:structure:load db:seed RAILS_ENV=test
+    command: cd src/api; bundle exec rake db:setup RAILS_ENV=test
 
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -40,7 +40,7 @@ aliases:
 
   - &create_test_db
     name: Create database
-    command: cd src/api; bundle exec rake db:create db:structure:load db:seed RAILS_ENV=test
+    command: cd src/api; bundle exec rake db:setup RAILS_ENV=test
```
hennevogel, line 43: `db:setup` does not include create 

https://github.com/rails/rails/blob/master/activerecord/lib/active_record/railties/databases.rake#L340

dmarcoux: Somebody should tell that to its description which literally says: `Creates the database, (...)`

dmarcoux: The CI would also not work if the database wasn't created.

PR #9816, opened 2020-06-23
Remove check_database script and DB::Checker

dmarcoux:
The only reference to the `check_database` script is from the `2.4` release notes and the RuboCop TODOs...

`DB::Checker` can also be deleted since it isn't used anymore due to the removal of the `check_database` script.


hennevogel:
kill em all

PR #9822, opened 2020-06-23
Remove RoleTest

dmarcoux:
Those 2 tests are clearly unneeded.

`test_something` is not really testing something, maybe if saving/destroying works in ActiveRecord.

`test_role` is checking if validations work.


src/api/test/unit/role_test.rb
```diff
@@ -1,15 +0,0 @@
...
-
-class RoleTest < ActiveSupport::TestCase
-  def test_something
-    norole = Role.create(title: 'norole')
-    norole.title = 'thisrole'
-    norole.save
-    norole.destroy
-  end
-
-  def test_role
```
dmarcoux, line 11: Should we test this with `validates_uniqueness_of` from `shoulda-matchers`? 

I don't really see benefits in doing so, but I'm interesting in knowing what you think.

hennevogel: Testing that rails features work makes little sense. The rails test suite tests that...

PR #9829, opened 2020-06-25
Show the flash message after marking a notification as read/unread

dmarcoux:
Without this change, the flash message is only shown after reloading the page.


src/api/app/views/webui/users/notifications/_update.js.erb
```diff
@@ -1,2 +1,4 @@
 $('#filters').html("<%= j(render partial: 'notifications_filter', locals: { projects_for_filter: projects_for_filter, notifications_count: notifications_count }) %>");
 $('#notifications-list').html("<%= j(render partial: 'notifications_list', locals: { notifications: notifications }) %>");
+$('#flash').html("<%= escape_javascript(render(layout: false, partial: 'layouts/webui/flash', object: flash)) %>");
+
```
eduardoj, line 4: Extra empty line at the end of the file.
```suggestion
```

PR #9833, opened 2020-06-25
Autofocus on first field for new project or new package

danidoni:
Autofocus on the first field on the new project form or new package form

### DIFF
--- a/src/api/app/views/webui/project/_form.html.haml
+++ b/src/api/app/views/webui/project/_form.html.haml
@@ -7,11 +7,11 @@
       .input-group-prepend
         %span.input-group-text
           #{namespace}:
-      = form.text_field(:name, required: true, class: 'form-control')
+      = form.text_field(:name, required: true, class: 'form-control', autofocus: true)
   - else
     = form.label(:name, 'Project Name:')
     %abbr.text-danger{ title: 'Required' } *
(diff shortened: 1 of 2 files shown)


src/api/app/views/webui/project/_form.html.haml
```diff
@@ -7,7 +7,7 @@
       .input-group-prepend
         %span.input-group-text
           #{namespace}:
-      = form.text_field(:name, required: true, class: 'form-control')
+      = form.text_field(:name, required: true, class: 'form-control', autofocus: true)
```
dmarcoux, line 10: This `autofocus: true` should also be on this line 14 since this text field is only displayed when there is a namespace (so for subprojects).

danidoni: Right!

src/api/app/views/webui/project/_form.html.haml
```diff
@@ -7,9 +7,9 @@
...
   - else
-    = form.label(:name, 'Project Name:')
+    = form.label(:name, 'Project Name:', autofocus: true)
```
dmarcoux, line 12: `autofocus: true` needs to be on the text field

danidoni: So true, changing

PR #9835, opened 2020-06-25
Migrate AttribNamespacePolicy tests from Minitest to RSpec

dmarcoux:
We want to move away from Minitest.


src/api/app/models/attrib_namespace_modifiable_by.rb
```diff
@@ -1,5 +1,5 @@
 class AttribNamespaceModifiableBy < ApplicationRecord
-  belongs_to :attrib_namespaces
+  belongs_to :attrib_namespace
```
dmarcoux, line 2: This was wrong, but never used, so nobody noticed.

PR #9836, opened 2020-06-25
Chunk binaryversion requests when querying a remote server

mlschroe:


srinidhibs:
Thank you so much @mlschroe for this patch! This has helped us cut down on the build time of a **very** large product ISO using remote repositories by many hours - if I'm right, more than 4-6 hours!

I just have one question - do I also need the `cleandiffcache` function and associated code on the local OBS instance?

PR #9838, opened 2020-06-25
Create the data_migrations table unless it exists

hennevogel:
This is a cosmetic migration. We have this table in structure.sql and it
was most likely created by the data_migrate gem on every deployment
implicitely. Let's be explicit about this.

Also include the new mariadb defaults into the structure file.


hennevogel:
Do not merge until we see the package building results please

dmarcoux:
It's time to [normalize all the things](https://github.com/openSUSE/open-build-service/blob/e3301af481f6b38eaf481b77732aec9882ed3ef8/src/api/script/compare_structure_sql.sh)! This is the crazy option.

The other option is the use the same Docker image for MariaDB in CircleCI and in our Docker environment. This is the sane option.

hennevogel:
Okay the second time in a week that the dev-env on openSUSE, master on SLE is biting...

src/api/db/migrate/20200625142212_create_data_migrations.rb
```diff
@@ -0,0 +1,11 @@
+# rubocop:disable Rails/CreateTableWithTimestamps
+class CreateDataMigrations < ActiveRecord::Migration[6.0]
+  def up
+    return if ActiveRecord::Base.connection.table_exists? 'data_migrations'
```
dmarcoux, line 4: This can be replaced by `if_not_exists: true` in the `create_table` statement below.

https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-create_table

hennevogel: nice

hennevogel:
Let's sort this out after we sort out `structure.sql`...

PR #9852, opened 2020-06-29
Migrate from structure.sql to schema.rb

dmarcoux:
We don't use any database-specific features. This will simplify maintenance since we don't have to deal with all the tiny details of a SQL dump (like changes between various MariaDB versions)

Remove custom database rake tasks (see commit description for details on this)

With the introduction of schema.rb, some RuboCop cops are now active and linting the schema for potential offenses. One of them(Rails/UniqueValidationWithoutIndex) is registering offenses and to keep changes as small as possible, we don't want to address the offenses in this PR. We could have disabled the cops, but they do provide value. In the end, regenerating the RuboCop TODOs makes the CI pass and allows us to tackle those offenses in a separate PR.


dmarcoux:
Just want @hennevogel to have a look before merging since we talked about this.

PR #9862, opened 2020-07-01
[backend] add support for $BSConfig::certfile to configure a default …

mlschroe:
…cert file

We already have BSConfig::keyfile if forceprojectkeys is false.
We also need a default cert to build things like kernel packages.


coolo:
An update to the build submodule slipped in. I guess you better create a specific commit for that

coolo:
you even removed it

mlschroe:
Thanks!

PR #9873, opened 2020-07-03
Clean up old or irrelevant notifications

danidoni:
We delete old or outdated notifications for Requests and Comments.
When someone adds a Review the notifications page will show a
notification for a Request instead.

The notifications page can get a bit crowded with notifications
sometimes.
Most of those notifications would be no longer relevant because the
latest will have the most up to date information. With this change the
notifications page will only show useful and up to date data about
your Requests.


danidoni:

src/api/spec/services/notification_service/notifier_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'rails_helper'
+
+RSpec.describe NotificationService::Notifier do
+  let(:user_bob) { create(:confirmed_user, login: 'bob') }
+  let(:user_kim) { create(:confirmed_user, login: 'kim') }
+  let(:commenter) { create(:confirmed_user, login: 'ann') }
+
+  let(:create_bob_subscription) { create(:event_subscription_comment_for_project, receiver_role: 'maintainer', user: user_bob, channel: :rss) }
```
danidoni, line 8: Rubocop complained about the previous `let!` statements that weren't referenced anywhere else in the spec. I've made it explicit in those cases were they were necessary, but I don't fully like how it looks like. The thing is the alternatives I have (use the factories inline where necessary) are very much zero DRY. Any ideas?

src/api/app/models/review.rb
```diff
@@ -250,7 +250,7 @@ def matches_user?(user)
 
   def event_parameters(params = {})
     params = params.merge(_get_attributes)
-    params[:id] = id
+    params[:id] = bs_request.id
```
hennevogel, line 253: You can't just change the event to suit your needs. You will have to do this traversal in `Notification`.

Also requests are identified by number.

src/api/app/queries/outdated_notifications_finder/bs_request.rb
```diff
@@ -0,0 +1,10 @@
+class OutdatedNotificationsFinder::BsRequest
+  def initialize(scope, notifiable)
+    @scope = scope
+    @notifiable = notifiable
+  end
+
+  def call
+    @scope.where(notifiable_type: 'BsRequest', notifiable_id: @notifiable.id)
```
hennevogel, line 8: All of this code, just for a simple `.where`?

danidoni: At first this was going to be something that covered several models, and the idea was to start with the ones we knew and plug in the rest, but in the end we have two. I'll try to make it simpler

src/api/app/models/review.rb
danidoni: > You can't just change the event to suit your needs. You will have to do this traversal in `Notification`.

I think I'm not following

vpereira: one thing that henne said @danidoni  is that we don't use `bs_request.id` but `bs_request.number`

src/api/app/services/notification_service/notifier.rb
```diff
@@ -0,0 +1,52 @@
...
+      Airbrake.notify(e, event_id: @event.id)
+    end
+
+    private
+
+    def create_notification_per_subscription(subscription, channel)
+      return unless create_notification?(subscription.subscriber, channel)
+
+      klass = channel == :web ? 'Web' : 'RSS'
+      "NotificationService::#{klass}Channel".constantize.new(subscription, @event).call
```
vpereira, line 33: any other way to do this construction without the constantize?

src/api/app/services/notification_service/notifier.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def create_notification?(subscriber, channel)
+      return false if subscriber&.away?
+      return false if channel == :rss && !subscriber.try(:rss_token)
```
vpereira, line 38: you could follow the example from line 37 and use `subscriber&.rss_token`, right?

src/api/app/services/notification_service/web_channel.rb
```diff
@@ -0,0 +1,18 @@
...
+      @subscription = subscription
+      @event = event
+    end
+
+    def call
+      params = @subscription.parameters_for_notification.merge!(@event.parameters_for_notification)
+      notification = Notification.find_by(notifiable_type: params[:notifiable_type], notifiable_id: params[:notifiable_id], web: true)
+      OutdatedNotifications::Destroyer.new(@subscription.subscriber, notification).call if notification
+
+      notification = Notification.create(params)
```
vpereira, line 13: if will explode if notification is nil.. `Notification.create!` or just call `.projects` if  `notification` exists

src/api/app/services/notification_service/notifier.rb
danidoni: Yes, I wanted to try a small change and see if Code Climate accepted it. I'll address the other one as well.

src/api/app/services/notification_service/notifier.rb
danidoni: Can you explain why shouldn't I use the constantize?

src/api/app/services/notification_service/web_channel.rb
danidoni: We are making heavy changes to this section, but I will take it into account, thanks!

src/api/app/models/review.rb
danidoni: Ummm good point, I forgot about it, thanks!

danidoni: > You can't just change the event to suit your needs. You will have to do this traversal in `Notification`.

Once I have a Notification, created from the event payload set here, I no longer have access to the request to fetch the id and fix it there.

And as we have set the `notifiable_type` to `BsRequest` [here](https://github.com/openSUSE/open-build-service/pull/9873/files#diff-ac49bda8ff4dfaeb507282adc2cea1f3R27), the current `notifiable_id` is the one for the `Review`, but the type is not, so the `notifiable` association is broken...

Those `event_parameters` are only used to build the event, so I think its safe to make the change here.

src/api/app/services/notification_service/notifier.rb
vpereira: http://gavinmiller.io/2016/the-safesty-way-to-constantize/

danidoni: I see the problem.

I agree that using the `constantize` on data coming from the user is dangerous, but in this case the input is clearly limited to `Web` or `RSS`, so I see no problem here.

Anyway, I'm gonna rewrite it to avoid using the `constantize`

vpereira: Software engineering is programming over time. It's not so hard to write code that is correct today, the problem is the code calling this method tomorrow. Nobody review manually every called method to make sure that the parameter passed won't compromise it and to rewrite it safely using `constantize` is a PITA. Please avoid it.

src/api/app/services/notification_service/notifier.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      true
+    end
+
+    def notifiable_exists?
+      # We need this check because the notification is created in a delayed job.
+      # So the notifiable object could have been removed in the meantime.
+      notifiable_type = @event.parameters_for_notification[:notifiable_type]
+      notifiable_id = @event.parameters_for_notification[:notifiable_id]
+      notifiable_type.constantize.exists?(notifiable_id)
```
vpereira, line 49: the same here.

src/api/app/services/notification_service/web_channel.rb
```diff
@@ -0,0 +1,33 @@
...
+      notification = Notification.create(@parameters_for_notification)
+      notification.projects << NotifiedProjects.new(notification).call
+      notification.update(web: true)
+    end
+
+    private
+
+    def finder_class
+      type = @parameters_for_notification[:notifiable_type]
+      "OutdatedNotificationsFinder::#{type}".constantize
```
vpereira, line 26: same here

src/api/app/services/notification_service/notifier.rb
danidoni: Changed, check it out again, please

danidoni:
> please get a rid of constantize, which is normally possible RCE vector, almost like `eval`

Done

src/api/app/models/review.rb
hennevogel: > And as we have set the notifiable_type to BsRequest here

You don't *have* to do this right?

> Those event_parameters are only used to build the event, so I think its safe to make the change here.

The content of `Events` are send over the message bus. People might do something with it. If we *must* we can can change this, we never promised this to be stable. I also don't think you can actually do something with the review ID, but I haven't checked.

danidoni: I'll think about this and see if I can make a proposal with a better implementation

PR #9875, opened 2020-07-03
Add /package/dependency route to robots.txt

### LINKED ISSUE #9872: Require login for package/dependency
This route is very expensive and shouldn't be given for free to web spiders.

vpereira:
The route `/package/dependency` is too expensive to be hit by crawlers. 

Closes #9872


coolo:
I don't think the crawler that affects us is caring for your robts.txt

See https://programmer.ink/think/5d65baa285c87.html for the base of this

vpereira:
> See https://programmer.ink/think/5d65baa285c87.html for the base of this

sure, but it should go to the `robots.txt` for those that respect it

Issue #9876, opened 2020-07-03
Expensive webui routes


hennevogel:
Sometimes we have spiders that go through the webui routes that are not easily identified as spiders. They have today hit some expensive routes in the `Webui::PackageController`. We should review why those are expensive and what we still can allowed without login.

- dependency
- revisions
- rdiff
- commit

PR #9877, opened 2020-07-03
Require auth for expensive routes

hennevogel:


hennevogel:
We are looking for a more fine grained solution in #9876

hennevogel:
We want to merge this so we don't have to monkey patch all deployments.

hennevogel:
This is done in #9881 now

PR #9878, opened 2020-07-03
Redesign notifications view to be more compact

saraycp:
A notification has now a more compact aspect. Specially useful on small screens. Using icons, colors and avatars makes it easier to understand at a glance.

- An icon represents each kind of notification: request or comment.
- Review notifications are going to be replaced by requests
  notifications with "in review" state.
- Display the state of the request clearly.
- Display involved users.
- Link to the comments box and not the comment itself.

The last design version proposed:

Screenshot_2020-07-08 Open Build Service(2)


src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -16,6 +16,81 @@ def filter_notification_link(link_text, amount, filter_item)
...
+
+    capture do
+      tag.i(class: 'fas fa-comments mr-1')
+    end
+  end
+
+  def request_state(notification)
+    state = if notification.notifiable_type == 'BsRequest'
+              notification.notifiable.state
+            elsif notification.notifiable_type == 'Review'
```
vpereira, line 30: what will happen if notifiable_type is neither `BsRequuest` or `Review`?

src/api/app/assets/stylesheets/webui/project-package-show-grid.scss
```diff
@@ -6,7 +6,7 @@
   grid-area: build-results;
 }
 
-.comments {
+#comments-list {
```
vpereira, line 9: not possible to keep it as `comments` ?

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -16,6 +16,81 @@ def filter_notification_link(link_text, amount, filter_item)
...
+      'secondary'
+    when :declined, :revoke
+      'danger'
+    when :superseded
+      'warning'
+    when :accepted
+      'success'
+    when /requires review/
+      'dark'
+    end
```
vpereira, line 53: no default?

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -16,6 +16,81 @@ def filter_notification_link(link_text, amount, filter_item)
...
+                  "##{notification.notifiable.number}"
+                when 'Event::ReviewWanted'
+                  "##{notification.notifiable.bs_request.number}"
+                when 'Event::CommentForRequest'
+                  "Request ##{notification.notifiable.commentable.number}"
+                when 'Event::CommentForProject'
+                  notification.notifiable.commentable.name
+                when 'Event::CommentForPackage'
+                  commentable = notification.notifiable.commentable
+                  "#{commentable.project.name} / #{commentable.name}"
```
vpereira, line 68: no default?

src/api/app/assets/stylesheets/webui/project-package-show-grid.scss
saraycp: At this point I need an anchor to let the notification point to the beginning of the comments box.
As there is another `#comments` in the same view (see `card-body`), I needed to use another ID to go to the top of the card (including card header).

src/api/app/helpers/webui/notification_helper.rb
saraycp: I'm not sure we need it, since the options are defined by us. I don't see how we can get a different value here.

src/api/app/helpers/webui/notification_helper.rb
saraycp: Ok, added an `else`clause, so any other state that would appear in black.

src/api/app/helpers/webui/notification_helper.rb
saraycp: We have slightly modified the code, so it does nothing in case it's not `BsRequest`. Review part is dropped since it's going to disappear soon.

src/api/app/models/notification.rb
```diff
@@ -2,6 +2,8 @@ class Notification < ApplicationRecord
   MAX_RSS_ITEMS_PER_USER = 10
   MAX_RSS_ITEMS_PER_GROUP = 10
 
+  self.ignored_columns = ['type']
```
DavidKang, line 5: This line of code, shouldn't be here, right? If I'm not wrong we removed it in a previous PR.

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
```diff
@@ -17,23 +17,26 @@
...
+            .actions.ml-auto.align-self-start
+              - unless params[:type] == 'read'
+                = link_to(my_notification_path(id: notification),
+                          method: :put, class: 'btn btn-sm btn-outline-success px-3', title: 'Mark as "Read"') do
                   %i.fas.fa-check
-            - unless notification.unread?
-              .actions.align-self-end.align-self-sm-start.pl-3.pt-3.pt-sm-0
-                = link_to(my_notification_path(id: notification, type: params[:type], project: params[:project]),
-                        method: :put, class: 'btn btn-sm btn-outline-success px-4', title: 'Mark as "Unread"', remote: true) do
+              - unless notification.unread?
```
danidoni, line 31: Can we swap the conditions and make them positive instead of negative?

DavidKang:
The reviews have the same "look and feel" than a comment.

Screenshot_2020-07-07 Open Build Service

If we are going to not show the reviews notifications anymore. I suggest to exclude them from the notification list.

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
DavidKang: yes, maybe we could refactor it a little bit the `actions` block, because the only differences are the `title` and the `icon`

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -16,6 +18,75 @@ def filter_notification_link(link_text, amount, filter_item)
...
+                end
+
+    link_to(link_text, notification.link_to_notification_target, class: 'mx-1 text-word-break-all')
+  end
+
+  def avatars(notification)
```
hennevogel, line 72: This *really* looks like a partial.

saraycp: True! I already moved as much as I could to a partial. Please have another look, @hennevogel

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
saraycp: Moved to a helper and refactored, @danidoni

src/api/app/models/notification.rb
saraycp: Dropped!

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
```diff
@@ -17,23 +17,18 @@
...
-                  %i.fas.fa-undo
+          .list-group-item.px-2.grid-container
+            .notifiable
+              = notification_icon(n.notifiable_type)
+              = notification_link(notification)
+              = request_state(notification)
+            .actions.ml-auto.align-self-start
+              = read_unread_link(notification)
+            - if notification.kind_of_request
+              %p.kind.text-word-break-all.m-0= notification.kind_of_request
```
hennevogel, line 28: This is displaying up to *50* times `submit requestor_project/package_1594135627_1`. I doubt this is what you want.

If the request has one action, use `actions_summary`. If it has more than one action print something like "Multiple Actions".

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -16,6 +18,98 @@ def filter_notification_link(link_text, amount, filter_item)
...
+    end
+  end
+
+  def notification_link(notification)
+    link_text = case notification.event_type
+                # TODO: add 'Event::ReviewWanted' when converted to BsRequest notification
+                when 'Event::RequestStatechange', 'Event::RequestCreate'
+                  "##{notification.notifiable.number}"
+                when 'Event::CommentForRequest'
+                  "Request ##{notification.notifiable.commentable.number}"
```
hennevogel, line 61: Either `#7` or `Request #7`. Not both.

Screenshot from 2020-07-08 14-11-53

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -16,6 +18,98 @@ def filter_notification_link(link_text, amount, filter_item)
...
+                  "#{commentable.project.name} / #{commentable.name}"
+                end
+
+    link_to(link_text, notification.link_to_notification_target, class: 'mx-1 text-word-break-all')
+  end
+
+  def read_unread_link(notification)
+    title, icon = notification.unread? ? ['Mark as "Read"', 'fa-check'] : ['Mark as "Unread"', 'fa-undo']
+
+    link_to(my_notification_path(id: notification),
```
hennevogel, line 75: You omitted the `remote: true` part which breaks this.

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -37,4 +131,11 @@ def notification_filter_active?(filter_item)
   def badge_color(filter_item)
     notification_filter_active?(filter_item) ? 'badge-light' : 'badge-primary'
   end
+
+  def reviewers_tags(review)
+    return user_avatar(User.find_by(login: review.by_user)) if review.by_user
+    return user_avatar(Group.find_by(title: review.by_group)) if review.by_group
+    return simulated_avatar('fa-archive text-warning', "Package #{review.by_project}/#{review.by_package}") if review.by_package
+    return simulated_avatar('fa-cubes text-secondary', "Project #{review.by_project}") if review.by_project
+  end
```
hennevogel, line 140: All the methods in this helper that use `capture` are partials. Is there a reason we using helper methods instead?

hennevogel:
Screenshot from 2020-07-08 14-19-44

Design wise I think the emphasis is too much on the "mark as read" button. First and foremost we want people to review their notifications. Also there are too many font sizes and weights going on. Links, label, text and muted text all look different which makes this rather busy.

I also think you could separate request meta data and request content better. Something like

notification

saraycp:
> notification

Your proposal splits the content in 4 lines having a lot of space on the left side of the button.

This is another alternative @DavidKang  and I propose:

Here the button goes to the bottom (less prominent) but also the avatars and date.

In desktop version, we keep the same structure, but unified text colors (all of them are black) but we use different `font-weight` to make a clear difference between the kind for request's text (_submit home:Admin_) and the excerpt:

Screenshot_2020-07-08 Open Build Service(2)

What do you think about this approach? @hennevogel

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
```diff
@@ -17,23 +17,18 @@
...
-                  %i.fas.fa-check
-            - unless notification.unread?
-              .actions.align-self-end.align-self-sm-start.pl-3.pt-3.pt-sm-0
-                = link_to(my_notification_path(id: notification, type: params[:type], project: params[:project]),
-                        method: :put, class: 'btn btn-sm btn-outline-success px-4', title: 'Mark as "Unread"', remote: true) do
-                  %i.fas.fa-undo
+          .list-group-item.px-2.grid-container
+            .notifiable
+              = notification_icon(n.notifiable_type)
+              = notification_link(notification)
```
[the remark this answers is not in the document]
saraycp: Replacing `#7` by `Request #7` as suggested [here](https://github.com/openSUSE/open-build-service/pull/9878#discussion_r451495233) makes the link larger. Do you think is it enough?

Using a bigger font-size is another option, but I don't think it's necessary.

src/api/app/helpers/webui/notification_helper.rb
saraycp: `Request #7`then.

src/api/app/helpers/webui/notification_helper.rb
saraycp: Sure, we probably messed this up when rebasing.

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
saraycp: We now display something like: `Multiple Actions on project home:Admin` when we have more than one action.

src/api/app/helpers/webui/notification_helper.rb
saraycp: The complex part here is displaying as "avatars" many different things: sometimes real avatars, sometimes simple icons and sometimes  the counter (`+N`). I need the helpers you mention because the are called from many places. 

Apart from that, I can't convert them to partials because I need to call them inside the `request_tags` helper. Which is the one that gives us an array of things we can really use.

hennevogel: I'm not talking only about the helper above. I'm talking about all the helpers that use some form of `capture`,

saraycp: We have moved as much as we could to the view. Now can be easier to understand.

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -16,6 +18,78 @@ def filter_notification_link(link_text, amount, filter_item)
...
+                when 'Event::CommentForRequest'
+                  "Request ##{notification.notifiable.commentable.number}"
+                when 'Event::CommentForProject'
+                  notification.notifiable.commentable.name
+                when 'Event::CommentForPackage'
+                  commentable = notification.notifiable.commentable
+                  "#{commentable.project.name} / #{commentable.name}"
+                end
+
+    link_to(link_text, notification.link_to_notification_target, class: 'mx-1 text-word-break-all')
```
hennevogel, line 49: This can go do the view if `notification_link` just returns `link_text`

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -16,6 +18,78 @@ def filter_notification_link(link_text, amount, filter_item)
...
+  end
+
+  def read_unread_link(notification)
+    title, icon = notification.unread? ? ['Mark as "Read"', 'fa-check'] : ['Mark as "Unread"', 'fa-undo']
+
+    link_to(my_notification_path(id: notification),
```
hennevogel, line 55: This can go to the view if `read_unread_link` just returns the array of `title` and `icon`.

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -16,6 +18,78 @@ def filter_notification_link(link_text, amount, filter_item)
...
+            method: :put, class: 'btn btn-sm btn-outline-success px-3', title: title, remote: true) do
+      concat(tag.i(class: "fas #{icon}"))
+    end
+  end
+
+  # A limited number of avatars are going to be displayed
+  def request_tags(request)
```
hennevogel, line 62: This really sounds like a scope or method in `BsRequest` that tells you the involved people instead. 

So instead of building an array of image tags in a helper, you would ask the request for the involved people and create the HTML in a loop.

I'm pretty sure the `BsRequest` method will be useful for other things too.

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -37,4 +111,11 @@ def notification_filter_active?(filter_item)
   def badge_color(filter_item)
     notification_filter_active?(filter_item) ? 'badge-light' : 'badge-primary'
   end
+
+  def reviewers_tags(review)
+    return user_avatar(User.find_by(login: review.by_user)) if review.by_user
+    return user_avatar(Group.find_by(title: review.by_group)) if review.by_group
+    return simulated_avatar('fa-archive text-warning', "Package #{review.by_project}/#{review.by_package}") if review.by_package
+    return simulated_avatar('fa-cubes text-secondary', "Project #{review.by_project}") if review.by_project
```
hennevogel, line 119: Review by package and review by project mean users too. You could fetch them from the `Review.users_and_groups_for_review`.

src/api/app/presenters/notification_presenter.rb
```diff
@@ -45,4 +34,18 @@ def excerpt
             end
     text.to_s.truncate(100)
   end
+
+  def kind_of_request
+    return unless @model.notifiable_type == 'BsRequest'
+
+    request = @model.notifiable
+    return "Multiple Actions on project #{request.bs_request_actions.first.target_project}" if request.bs_request_actions.size > 1
```
hennevogel, line 42:
```suggestion
    return "Multiple actions for project #{request.bs_request_actions.first.target_project}" if request.bs_request_actions.size > 1
```

src/api/app/presenters/notification_presenter.rb
```diff
@@ -45,4 +34,18 @@ def excerpt
...
+
+    BsRequest.actions_summary(@model.event_payload)
+  end
+
+  def commenters
```
hennevogel, line 47: This you can do with the commentable directly no? Like `commentable.comments.where('updated_at >= ?', @model.unread_date)...`

Sounds like an additional `commenters` scope to me.

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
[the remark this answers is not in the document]
hennevogel: No this is good for now

src/api/app/helpers/webui/notification_helper.rb
saraycp: That was my initial plan and at some point I thought it wasn't a good idea. I'll give it a second thought.

src/api/app/helpers/webui/notification_helper.rb
saraycp: We already have a `NotificationPresenter#link_to_notification_target` to get the paths we need for the link. So I'm merging these two methods to take both the link text and the link path in one shot. Have a look to `NotificationPresenter#notifiable_link`.

src/api/app/helpers/webui/notification_helper.rb
saraycp: Everything has been moved to the view.

src/api/app/presenters/notification_presenter.rb
saraycp: done

src/api/app/presenters/notification_presenter.rb
saraycp: True! Changed!

src/api/app/helpers/webui/notification_helper.rb
hennevogel: Have you forgot to push this code? `NotificationPresenter` is unchanged.

src/api/app/helpers/webui/notification_helper.rb
saraycp: If I'm not wrong, the idea here is to replace this "simulated avatar":
Screenshot_2020-07-10 Open Build Service
by the avatars of the users and groups related to that project.

This can end up with a huge list of avatars that are going to be hidden anyway because we are displaying only 6.

So if we keep what we have and show the "simulated avatar" (display project/package icon) we sum up what we want to say: "_Hey you, people involved in this project/package. Have a look!"_. And there is a high probability they are visible. In the other way around, we'll just see a couple of new faces only.

hennevogel: Okay we can also postpone this and find the middle ground. I'm fine with creating something in the backlog

src/api/.rubocop_todo.yml
```diff
@@ -285,6 +285,7 @@ Metrics/ModuleLength:
   Exclude:
     - 'app/helpers/flag_helper.rb'
     - 'app/helpers/maintenance_helper.rb'
+    - 'app/helpers/webui/notification_helper.rb'
```
hennevogel, line 288: I guess by now you can remove this again or?

saraycp: good catch

src/api/app/presenters/notification_presenter.rb
hennevogel: I meant scope inside the `Project`, `Package` or `BsRequest` but okay. Something for later I guess...

src/api/app/helpers/webui/notification_helper.rb
hennevogel: You or me?

hennevogel: I created it https://trello.com/c/HMSq1L3N

src/api/app/models/review.rb
```diff
@@ -58,6 +58,7 @@ class NotFoundError < APIError
   scope :bs_request_ids_of_involved_users, ->(user_ids) { where(user_id: user_ids).select(:bs_request_id) }
 
   scope :declined, -> { where(state: :declined) }
+  scope :open, -> { where(state: :new) }
```
hennevogel, line 61: Hm why do you invent a new term for the scope?

src/api/app/assets/images/request_icon.svg
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" standalone="no"?>
```
hennevogel, line 1: Please add the license information we have talked about

saraycp: We already had a `LICENCE.md` inside `app/assets/images/icons`, so we have moved the icon to that sub-directory and added the text in that existing file.

src/api/app/models/review.rb
saraycp: "Open reviews" is a term we use more often when talking about this. Moreover, renaming the scope to `new`will make us use it like `Review.new`, what can be even more confusing.

So it's been changed to `in_state_new`.

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -1,4 +1,6 @@
 module Webui::NotificationHelper
+  MAX_VISIBLE_AVATARS = 6
```
hennevogel, line 2: This is unused now

src/api/app/helpers/webui/notification_helper.rb
saraycp: This code has been refactored. Maybe is not exactly what you requested, but the code has been simplified a lot.
We can refactor this in another round.

PR #9883, opened 2020-07-07
Create data migration to adapt review and duplicated notifications 

krauselukas:
Instead of having dedicated notification for `Event::ReviewWanted`,
we will from now on unify this with the request notifications.
The association to the review will be dropped. Therefore we
need to associate the review notifications to the corresponding
request through a data migration.
Also, we are going to show only the latest state of a request
or comment through the notifications. The data migration
needs to remove duplicated/outdated notifications that already
exist.

DO NOT MERGE: The data migration will cause a downtime, we need to wait for the maintenance window. This should
be merged close to it.

### DIFF
--- a/src/api/db/data/20200702145811_adapt_review_and_duplicated_notifications.rb
+++ b/src/api/db/data/20200702145811_adapt_review_and_duplicated_notifications.rb
@@ -1,0 +1,59 @@
+class AdaptReviewAndDuplicatedNotifications < ActiveRecord::Migration[6.0]
+  def up
+    transform_review_notifications
+    remove_outdated_notifications
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+
+  private
+
+  # associate the review notifications to the corresponding bs_request, since
+  # they won't be treated separately anymore
+  def transform_review_notifications
+    Notification.where("notifiable_type = 'Review'").find_each do |notification|
+      notification.update(notifiable_id: notification.notifiable.bs_request.id, notifiable_type: 'BsRequest')
+    end
+  end
+
+  # only keep the latest comment and request notifications related to the same notifiable and subscriber
+  def remove_outdated_notifications
+    Notification.find_each do |notification|
+      next if delete_notification_without_notifiable(notification)
+
+      notifications = outdated_notifications(notification)
+
+      notifications.destroy_all
+    end
+  end
+
+  def outdated_notifications(notification)
+    return fetch_comment_notifications(notification) if notification.notifiable_type == 'Comment'
+
+    fetch_bsrequest_notifications(notification)
+  end
+
+  def fetch_comment_notifications(notification)
+    notifiable = notification.notifiable
+    Notification.for_web.where(notifiable_type: 'Comment', subscriber_id: notification.subscriber_id)
+                .joins('JOIN comments ON notifications.notifiable_id = comments.id')
+                .where(comments: { commentable_type: notifiable.commentable_type,
+                                   commentable_id: notifiable.commentable_id }).order(id: :desc).offset(1)
+  end
+
+  def fetch_bsrequest_notifications(notification)
+    Notification.for_web.where(notifiable_id: notification.notifiable_id,
+                               notifiable_type: notification.notifiable_type,
+                               subscriber_id: notification.subscriber_id).order(id: :desc).offset(1)
+  end
+
+  # make sure the notifiable exist, otherwise delete the notification since it refers to nothing
+  def delete_notification_without_notifiable(notification)
+    return false if notification.notifiable
+
+    notification.destroy
+    true
+  end
+end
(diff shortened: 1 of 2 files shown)


krauselukas:

hennevogel:
Have you checked how long the downtime will be?

krauselukas:
> Have you checked how long the downtime will be?

@hennevogel not yet, I will check :+1:

src/api/spec/db/data/adapt_review_and_duplicated_notifications_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      let(:second_comment_notification) do
+        create(:notification, :comment_for_request, notifiable: comment_request02,
+                                                    subscriber: subscriber, web: true)
+      end
+
+      before do
+        first_request_notification
+        second_request_notification
+        first_comment_notification
+        second_comment_notification
```
DavidKang, line 55: I think we could use `let!` instead of calling them inside the `before` block

krauselukas: I cannot use `let!` on `first_request_notification` and `first_comment_notification` because rubocop doesn't allow it since I don't refer to them in the `it` block's. But I need them to be in the database, and to create them in the right order with the `second` notifications. Mixing up `let!` and calling the other from the `before` block makes it hard to keep them in order and is hard to read IMO. That why I used the `let` for all of them and call them in order from the `before` block

DavidKang: okey, thanks ^^

krauselukas:
@hennevogel @eduardoj I forgot to take the subscriber into account for the comment notifications, please check again

krauselukas:
@eduardoj @hennevogel I had to add a check to make sure that the notifiable is not `nil` since this lets the data migration explode. We have the rare occurrence of it (seems to be just hiccups), but need to be handled here. I simply delete the notifications if the corresponding notifiable is missing.

src/api/db/data/20200702145811_adapt_review_and_duplicated_notifications.rb
```diff
@@ -0,0 +1,64 @@
...
+      next if delete_notification_without_notifiable(notification)
+
+      notifications = outdated_notifications(notification)
+
+      next if notifications.count <= 1
+
+      # check for the highest id, because the 'created_at' date can be the same for multiple notifications,
+      # since a delayed_job is creating them in batches
+      newest_notification = notifications.first
+      notifications.where.not(id: newest_notification.id).destroy_all
```
danidoni, line 33: I think you can use `notifications.reject { |n| n.id == newest_notification.id }.destroy_all` and avoid querying the DB again with the `where`.

danidoni: You can try also this
```ruby
newest_notification = notifications.shift
notifications.destroy_all
```

src/api/spec/db/data/adapt_review_and_duplicated_notifications_spec.rb
danidoni: I found the same issue in my PR, the only thing I did was to rename the `let`s so they were verbs and looked like an action (e.g. `create_first_request_notification`)

src/api/db/data/20200702145811_adapt_review_and_duplicated_notifications.rb
krauselukas: Hmmm, `shift` doesnt work on `ActiveRecord_Relation` objects, the `reject` returns an array and you can't do the `destroy_all` on it. I will check if what we can do here

krauselukas: I know just iterate over the already fetched records, and call `.destroy`. This avoids querying the two times. `destroy_all` does the same, it calls `destroy` on every record. Couldn't find a more elegant solution that doesn't cause more complexity.

krauselukas: okay, we changed it so we use `.offset(1)` on the queries in `fetch_comment_notifications` and fetch_request_notifications`. This way we do not include the newest notification into the collection, and we can simply call `destroy_all` on them

krauselukas: https://guides.rubyonrails.org/v6.0/active_record_querying.html#limit-and-offset

src/api/db/data/20200702145811_adapt_review_and_duplicated_notifications.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    fetch_bsrequest_notifications(notification)
+  end
+
+  def fetch_comment_notifications(notification)
+    notifiable = notification.notifiable
+    Notification.for_web.where(notifiable_type: 'Comment', subscriber_id: notification.subscriber_id)
+                .joins('JOIN comments ON notifications.notifiable_id = comments.id')
+                .where(comments: { commentable_type: notifiable.commentable_type,
+                                   commentable_id: notifiable.commentable_id }).order(id: :desc).offset(1)
```
danidoni, line 43: :clap: Nice trick!

src/api/db/data/20200702145811_adapt_review_and_duplicated_notifications.rb
danidoni: Awesome!

PR #9884, opened 2020-07-07
Add last_seen_at column to notifications table

krauselukas:
In order to show the avatars of the users involved in the notifiable of the notification since the user last time checked the notification, we need to save the date and time the user lastly did so.

In other words:

One of the improvements in notifications is unifying/merging notifications affecting the same notifiable. Keeping all the information in one single line.

When many users write comments in my project, instead of receiving many notifications, I'll receive only one notification summarizing what's new since last time I checked.

Among others, we need to track who are the new commenters. To do so, we decided to store the date we last checked the notifications `last_seen_at` and then iterate over the list of comments created from that time to the present in order to take the needed information.

Related to https://github.com/openSUSE/open-build-service/pull/9873 & 
https://github.com/openSUSE/open-build-service/pull/9878

### DIFF
--- a/src/api/db/migrate/20200703110429_add_last_seen_at_to_notifications.rb
+++ b/src/api/db/migrate/20200703110429_add_last_seen_at_to_notifications.rb
@@ -1,0 +1,5 @@
+class AddLastSeenAtToNotifications < ActiveRecord::Migration[6.0]
+  def change
+    add_column :notifications, :last_seen_at, :datetime
+  end
+end
--- a/src/api/app/models/notification.rb
+++ b/src/api/app/models/notification.rb
@@ -44,6 +44,7 @@ end
 #  delivered                  :boolean          default(FALSE)
 #  event_payload              :text(65535)      not null
 #  event_type                 :string(255)      not null
+#  last_seen_at               :datetime
 #  notifiable_type            :string(255)      indexed => [notifiable_id]
 #  rss                        :boolean          default(FALSE)
 #  subscriber_type            :string(255)      indexed => [subscriber_id]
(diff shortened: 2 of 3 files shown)


krauselukas:

src/api/db/migrate/20200703110429_add_last_seen_at_to_notifications.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastSeenAtToNotifications < ActiveRecord::Migration[6.0]
+  def change
+    add_column :notifications, :last_seen_at, :datetime
```
eduardoj, line 3:
```suggestion
    add_column :notifications, :last_seen_at, :datetime, precision: 6
```

Not sure if we need to add the precision. :thinking:

krauselukas: yes let's be explicit here, currently we don't rely to much on the precision, but doesn't hurt

vpereira: unless you want to do stuff  like:

```user.save
'UPDATE `users` SET `updated_at` = '2016-01-18 23:59:59.999999' WHERE `users`.`id` = 1'
```

we don't need precision at all

saraycp: @eduardoj and I have been talking about this and did some verification. Rails 6 add `precision(6)` to datetime fields created as `timestamps` and not to those created just as `datetime`.
In this case, then, I think we don't need it.

PR #9885, opened 2020-07-07
Display selected project name for mobile

### LINKED ISSUE #9882: Notification filter box doesn't show which project we are filtering
For small devices, the notification filter box show us what we are filtering: 'Inbox", 'Requests', etc... But when we filter by project it shows 'Inbox' instead of the name of the project.

quatran:
For small devices the notification filter box now displays the project
name instead of 'Inbox'.

Fixes #9882  

**Before**
issue_9982_before

**After**
issue_9982_after

### DIFF
--- a/src/api/app/views/webui/users/notifications/index.html.haml
+++ b/src/api/app/views/webui/users/notifications/index.html.haml
@@ -8,7 +8,7 @@
     .card.mb-3
       %strong.d-block.d-md-none.p-3{ data: { toggle: 'collapse', target: '#filters' },
                                   aria: { expanded: true, controls: 'filters' } }
-        Filtered by: #{(params[:type] || 'Inbox').humanize}
+        Filtered by: #{(params[:type] || params[:project] || 'Inbox').humanize}
         %i.float-right.mt-1.fa.fa-chevron-down
       .card-body.collapse#filters
         = render partial: 'notifications_filter', locals: { projects_for_filter: @projects_for_filter, notifications_count: @notifications_count }
--- a/src/api/app/views/webui/users/notifications/index.html.haml
+++ b/src/api/app/views/webui/users/notifications/index.html.haml
@@ -8,7 +8,7 @@
     .card.mb-3
       %strong.d-block.d-md-none.p-3{ data: { toggle: 'collapse', target: '#filters' },
                                   aria: { expanded: true, controls: 'filters' } }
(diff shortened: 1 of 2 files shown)


src/api/app/views/webui/users/notifications/index.html.haml
```diff
@@ -8,7 +8,7 @@
     .card.mb-3
       %strong.d-block.d-md-none.p-3{ data: { toggle: 'collapse', target: '#filters' },
                                   aria: { expanded: true, controls: 'filters' } }
-        Filtered by: #{(params[:type] || 'Inbox').humanize}
+        Filtered by: #{(params[:type] || params[:project] || 'Inbox').humanize}
```
hennevogel, line 11: You should never render unescaped user input on the page.

Humanize titleizes too. Turns `openSUSE:Factory` into `OpenSUSE Factory`

hennevogel: https://obs-reviewlab.opensuse.org/quatran-issue_9882_notification_filter_box/my/notifications?project=/hans/

hennevogel: I'm also rather sure you can XSS this once you can around the constraints

https://obs-reviewlab.opensuse.org/quatran-issue_9882_notification_filter_box/my/notifications?project=%3Cscript%3Ealert(%27XSS%27)%3C/script%3E

eduardoj:
I added a couple of commits, to make the controller validate the input of the `project` parameter, as it was already done with the `type` parameter. @hennevogel, this should be enough for preventing XSS. @danidoni, could you also review this again?

@quatran, do you mind having a look?

eduardoj:
@hennevogel:

> This seems a bit complicated to me. Isn't this rather a matter of saving the project we filter by in an instance variable (if we find it) then use that instance variable to render the name?

The main change we propose is that the project is validated against `@projects_for_filter`, and this way we don't need to instantiate the project requiring an aditional query to the database. Yes, we use `params[:project]` in the view, but now it is validated, and moreover we ensure the name belongs to the project filters.

Wouldn't your approach be more expensive?

hennevogel:
> Wouldn't your approach be more expensive?

No, as one database query has no real performance impact while executing this code. This code however has significant impact on maintenance. Especially as it's *implicitly* guarding against some serious security concerns.

PR #9886, opened 2020-07-07
Handle multibuild in owner search

### LINKED ISSUE #9837: /search/owner?binary=<binary-name> does not view groups as maintainers/bugowners
# /search/owner?binary=<binary-name> does not return maintainer/bugowner groups

Expected Result
============
/search/owner?binary=<binary-name> should also return groups

How to Reproduce
===============
1. isc api /search/owner?binary=petsc
2. compare with isc maintainer SUSE:SLE-15:Update petsc

Further Information
===============
* This is reproducible on internal BS

hennevogel:
When searching the backend for binaries it often returns
the src package. Those can belong to any of the multibuild
package container names.

Fixes #9837

### DIFF
--- a/src/api/app/models/owner_search/assignee.rb
+++ b/src/api/app/models/owner_search/assignee.rb
@@ -94,6 +94,7 @@ module OwnerSearch
 
       package_name = b['package']
       package_name.gsub!(/\.[^.]*$/, '') if prj.is_maintenance_release?
+      package_name = Package.striping_multibuild_suffix(package_name)
       pkg = prj.packages.find_by_name(package_name)
 
       return false if pkg.nil? || pkg.is_patchinfo?


src/api/app/models/owner_search/assignee.rb
```diff
@@ -94,6 +94,7 @@ def parse_binary_info(b, prj)
 
       package_name = b['package']
       package_name.gsub!(/\.[^.]*$/, '') if prj.is_maintenance_release?
+      package_name = Package.striping_multibuild_suffix(package_name)
```
vpereira, line 97: It could for sure be extracted to a method, where you give the` b['package']` and `pkg` and returns the results of `prj.packages.find_by_name`

hennevogel: All of this should be refactored deeply, yes.

PR #9896, opened 2020-07-09
[backend] add memoization of files via the $conf{'memoize'} file

sjamgade:
The master process can memoize files and provide them to the forked
children, if they havent changed since the last time they were changed.

The children or any other process can request memoization by writing its
path to $conf{'memoize'} file as stored in the master process. The
children do not run periodic check so only master process will be doing
the periodic calls.
The maxsize of any memoized file can be controlled via
$conf{'memoize_max_size'}. It does not have default value but this value
is mandatory for memoization to continue.


marcus-h:
On 2020-07-09 04:50:42 -0700, Sumit Jamgade wrote:
> The master process can memoize files and provide them to the forked
> children, if they havent changed since the last time they were changed.
>
> The children or any other process can request memoization by writing its
> path to $conf{&#39;memoize&#39;} file as stored in the master process.

Out of curiousity, what's your concrete use case in the OBS itself? (Currently,
there are no users of this feature).

> diff --git a/src/backend/BSStdServer.pm b/src/backend/BSStdServer.pm
> index eceebc40496..e64d4a01a09 100644
> --- a/src/backend/BSStdServer.pm
> +++ b/src/backend/BSStdServer.pm

<SNIP>

> +sub do_memoize {
> +  my ($fn, $msize) = @_;
> +  if (-e "$fn") {

return unless -e "$fn"; would save a level of indention:)

> +    my $file;
> +    BSUtil::lockopen($file, '<', "$fn");
> +    rename($fn, "$fn.read") || die("rename $fn $fn.read $!\n");

Just two thoughts:
- the rename might break your locking scheme (assumption: processes that
  write to the file also need an exclusive lock on it)
- hmm waiting for a lock in a periodic handler? (this might be OK (depending
  on how this feature is used...))

sjamgade:
This is first part of the PR and I am working on the second part.
Thank for the input.

mlschroe (approved):

PR #9898, opened 2020-07-09
Notification last seen at

danidoni:
Copy the last_seen_at over to the new notification upon cleanup

We need the last_seen_at field being copied over the new notifications
so we can show the avatars of the people commenting on a request since
the last time you marked a notification as read.

### DIFF
--- a/src/api/app/services/notification_service/web_channel.rb
+++ b/src/api/app/services/notification_service/web_channel.rb
@@ -9,19 +9,22 @@ module NotificationService
     def initialize(subscription, event)
       @subscription = subscription
       @event = event
-      @parameters_for_notification = @subscription.parameters_for_notification
-                                                  .merge!(@event.parameters_for_notification)
+      @parameters_for_notification = subscription_parameters.merge(event_parameters).merge!(web: true)
     end
 
     def call
-      # Destroy older notifications
+      return nil unless @subscription.present? && @event.present?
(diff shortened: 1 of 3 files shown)


src/api/app/services/notification_service/web_channel.rb
```diff
@@ -1,27 +1,28 @@
...
-      'Comment' => OutdatedNotificationsFinder::Comment
-    }.freeze
+    ALLOWED_FINDERS = { 'BsRequest' => OutdatedNotificationsFinder::BsRequest,
+                        'Comment' => OutdatedNotificationsFinder::Comment }.freeze
 
     def initialize(subscription, event)
       @subscription = subscription
       @event = event
       @parameters_for_notification = @subscription.parameters_for_notification
                                                   .merge!(@event.parameters_for_notification)
```
vpereira, line 11: this first `merge!` can return `nil` no?  I think the first `merge` should be without `!`

danidoni: You are right, @vpereira. I've added some guards also. Please check it out.

danidoni:
> please add a description to the PR

Done!

danidoni:

danidoni:
> LGTM, just squash a bit the commits, and I will approve

Yessir! o7

PR #9906, opened 2020-07-13
Add missing extension to image

saraycp:


saraycp:
Duplicated from #9905

DavidKang:
Then, let's close it. Thanks @saraycp

PR #9913, opened 2020-07-14
[dist] fix test case 0110-check_signd.ta for obs < 2.11

M0ses:
Re-written 0110-check_signd.ta in perl.

This patch checks if the function "create_sign_cert" exists in
/usr/lib/obs/server/setup-appliance.sh which was introduced in 2.11.
We do not compare version of obs-server package to be more flexible
if this feature gets backported.

Without this patch our test cases for 2.10 are failing.


M0ses:
Tested successful in 2.10 and unstable

Issue #6311, opened 2018-11-20
Introduce theming per user and instance


hellcp:
### Is your feature request related to a problem? Please describe.
I heard @simotek complaining about lack of dark theme option in OBS. At the same time @Conan-Kudo would like to be able to use company theme / project theme for instances of OBS. It would be nice to provide users with option to switch between themes that are available on the instance, but also allow instance admins to set default theme

### Describe the solution you'd like
Previously it was done with bratwurst, which wasn't a neat solution. However to properly setup theming both stylesheets, layouts and some icons have to be replacable. I would suggest creating a directory where themes would reside (like /src/api/app/themes?) with directories:
* assets
  * stylesheets
  * images
* views
  * layouts

(layouts because the most important part of theming will probably be header and footer, although would still be nice if other stuff was expandable, so for example replacing elements in views/packages to include more links in show, for it to better integrate with infrastructure of instance owners)

In addition, this would be a nice way to make webui3 and beyond, because we wouldn't be relying on adding switch_to_webui2 to every view we want to switch, instead opting for existing solution (will also be useful when updating bootstrap 4 to bootstrap 5 etc.)

### Describe alternatives you've considered
Make people switch themes by overriding existing code and files (that only works with instance based theming though).

### Additional context
Example of quite well executed theming in similar platform would have to go to pagure, although one of differences I would like to see there would be including scss compiling, because we might want to have some of our bootstrap usage expanded:
https://pagure.io/pagure/blob/master/f/pagure/themes

hennevogel:
Themeing that requires code changes doesn't really make sense IMHO. If you can do this, you can do this already right now, as Ruby on Rails allows for this to easily happen. We already do it currently with the bootstrap beta feature and it's a no-brainer. 

What I would find more interesting is a way to overwrite some of the sass variables (and maybe even introduce more narrow scopes for them) so you can influence the colors from inside the app.

hellcp:
@hennevogel you have to consider how other projects already have themes of their own, where not just variables are used. Heck, even we do with chameleon.

What I would really want is for themes to be drag and drop into a single directory, without overriding what's already in code.

Conan-Kudo:
So there are two main use-cases I see here:

* Having a corporate theme instance wide: This, I imagine is easy for OBS to support now.

* Having a per-user setting for swapping to something like a dark theme or high contrast theme (for vision issues)

guoyunhe:
[...]

PR #9929, opened 2020-07-19
[WIP] Chameleon theme

guoyunhe:
As you may know, we made a bootstrap theme for all openSUSE websites. It will unify the look of websites and add a megamenu to navigate between all sites.

Light Mode:
(need tweaking)

Dark Mode:

图片


DavidKang:
Hi @guoyunhe, first of all, thanks for your contribution. Having the dark/light mode could be something cool!
One thing that I personally don't like, and it's my personal opinion, is the `megamenu`. The content of this `megamenu` is very tied to openSUSE. Also the content of this menu is not in the app context.

guoyunhe:
Hi @DavidKang , the megamenu is optional. I will make a configuration option so it only appears in openSUSE Build Serivce site.

> Also the content of this menu is not in the app context.

The content is similar to what we used to have with bento theme (a long time ago). It provides a way to navigate between all openSUSE sites.

DavidKang:
> I will make a configuration option so it only appears in openSUSE Build Service site.

Nice!

> The content is similar to what we used to have with bento theme (a long time ago). It provides a way to navigate between all openSUSE sites.

Is true, that in the past we had something similar. I' m still not convinced about the menu, but I understand your reason for it.

Another thing to think about, is if we are going to introduce this `megamenu` we need to align the colour of the menu and the icons in order to be consistent. For example, almost all the icons in this menu are `outline` and in the rest of the app are the opposite.

DavidKang:
Sorry, I close I by accident

hennevogel:
I would also suggest you think again about mixing application navigation into the "openSUSE" theme. Because unlike the web applications openSUSE installs (wiki, forums, wordpress etc.), many of the applications we maintain (OBS, OSEM, openQA, cooverview etc.) are generic applications that we ship to many people outside of the context of openSUSE.

Which currently would mean we would have double the effort to maintain navigation. One time for openSUSE, one time for everyone else. That's not something many of those projects will be willing to carry I'm afraid.

It's somehow the equivalent of adding this bar to upstream wordpress, which they will never agree to for the same reasons.

I would suggest you do not mix application navigation into your "project navigation bar". People then can just carry the maintenance of the button that triggers it and account for the extra pixels on top in their layout.

guoyunhe:
WordPress and MediaWiki supports theming so we can easily add the megamenu without mixing code into the generic apps. But OBS doesn't provide such feature.

I am moving the megamenu part into a config option. So by default here will not be code of megamenu in the app codebase. And only build.opensuse.org instance will add it.

The PR is still work in progress.

hennevogel:
> WordPress and MediaWiki supports theming so we can easily add the megamenu without mixing code into the generic apps. But OBS doesn't provide such feature.

Yeah, we have a feature open for that for a long time #6311 but we didn't really have the need for it yet.

> I am moving the megamenu part into a config option. So by default here will not be code of megamenu in the app codebase. And only build.opensuse.org instance will add it.

I'm not really worried about what is displayed or not. We of course can figure this out.

I'm worried about the time when *we* are going to change the app navigation in the future. Then *we* have to take two styles of navigation into account and make this work. Means double the amount of work.

guoyunhe:
> I'm worried about the time when we are going to change the app navigation in the future. Then we have to take two styles of navigation into account and make this work. Means double the amount of work.

The megamenu is an individual popup, independent to the navbar of OBS or any other sites. Here is no need to do any extra work for it. It just works out of box, no matter how the site itself changes. We have been using it for wikis, news and doc websites. Once megamenu is updated, it automatically applied to all sites because the script is hosted in one place on static.opensuse.org .

build.opensuse.org just need the following configuration to make it work:

```ruby
CONFIG['navbar_extra'] = '<button class="navbar-toggler megamenu-toggler" type="button" data-toggle="collapse" data-target="#megamenu" aria-expanded="true">
  <svg width="1em" height="1em" viewBox="0 0 16 16" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
    <path fill-rule="evenodd" d="M1 2.5A1.5 1.5 0 0 1 2.5 1h3A1.5 1.5 0 0 1 7 2.5v3A1.5 1.5 0 0 1 5.5 7h-3A1.5 1.5 0 0 1 1 5.5v-3zM2.5 2a.5.5 0 0 0-.5.5v3a.5.5 0 0 0 .5.5h3a.5.5 0 0 0 .5-.5v-3a.5.5 0 0 0-.5-.5h-3zm6.5.5A1.5 1.5 0 0 1 10.5 1h3A1.5 1.5 0 0 1 15 2.5v3A1.5 1.5 0 0 1 13.5 7h-3A1.5 1.5 0 0 1 9 5.5v-3zm1.5-.5a.5.5 0 0 0-.5.5v3a.5.5 0 0 0 .5.5h3a.5.5 0 0 0 .5-.5v-3a.5.5 0 0 0-.5-.5h-3zM1 10.5A1.5 1.5 0 0 1 2.5 9h3A1.5 1.5 0 0 1 7 10.5v3A1.5 1.5 0 0 1 5.5 15h-3A1.5 1.5 0 0 1 1 13.5v-3zm1.5-.5a.5.5 0 0 0-.5.5v3a.5.5 0 0 0 .5.5h3a.5.5 0 0 0 .5-.5v-3a.5.5 0 0 0-.5-.5h-3zm6.5.5A1.5 1.5 0 0 1 10.5 9h3a1.5 1.5 0 0 1 1.5 1.5v3a1.5 1.5 0 0 1-1.5 1.5h-3A1.5 1.5 0 0 1 9 13.5v-3zm1.5-.5a.5.5 0 0 0-.5.5v3a.5.5 0 0 0 .5.5h3a.5.5 0 0 0 .5-.5v-3a.5.5 0 0 0-.5-.5h-3z"/>
  </svg>
</button>'
CONFIG['head_extra'] = ''
CONFIG['body_before'] = '<div id="megamenu" class="megamenu collapse"></div>'
CONFIG['body_after'] = '<script src="https://static.opensuse.org/chameleon-3.0/dist/js/chameleon.js"></script>'
```

hellcp:
I was working on a fairly simple theming similar to what is done for TSP like https://github.com/openSUSE/travel-support-program/blob/master/config/initializers/config.rb#L5-L9

That way everybody can have it their way ;)

I do have that stashed somewhere, so if you want it, I can undig it

hellcp:
I submitted #9952 with the theming support, it should be fairly easy to implement this alongside the existing theme now

PR #9931, opened 2020-07-19
Add support to collect exceptions on influxdb

vpereira:
The influxdb rails gem try to save the whole exception stack when
something happens. The way that we are doing here, we are incremeting
a counter always when Airbrake.notify is called

Kind of related with #9930 

With this change we are able to track number of exceptions on Grafana, under `rails.exceptions`


hennevogel:
We found another way to do this for now. A "fix" for influxdb-rails is on it's way (influxdata/influxdb-rails#121)

PR #9935, opened 2020-07-20
[api] fix change time in reviews (when attriute)

adrianschroeter:
It was never updated which made the information in most
cases identical to request creation time always.

The attribute is supposed same as the when attriute
of the request itself, where it is the last time the
request changed.

### DIFF
--- a/src/api/app/models/review.rb
+++ b/src/api/app/models/review.rb
@@ -146,7 +146,7 @@ class Review < ApplicationRecord
     r.reviewer = r.creator = hash.delete('who')
     r.reason = hash.delete('comment')
     begin
-      r.created_at = Time.zone.parse(hash.delete('when'))
+      r.updated_at = Time.zone.parse(hash.delete('when'))
     rescue TypeError
       # no valid time -> ignore
     end
@@ -159,7 +159,7 @@ class Review < ApplicationRecord
   def _get_attributes
     attributes = { state: state.to_s }
     # old requests didn't have who and when
-    attributes[:when] = created_at.strftime('%Y-%m-%dT%H:%M:%S')
+    attributes[:when] = updated_at.strftime('%Y-%m-%dT%H:%M:%S')
     attributes[:who] = reviewer if reviewer
     attributes[:by_group] = by_group if by_group
     attributes[:by_user] = by_user if by_user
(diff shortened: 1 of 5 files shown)


hennevogel:
Not sure this does what you want it to do. `Review.updated_at` will change every time that the record changes. No matter if this is for a `state` change or not. Let's say because we delete a `Group` and `Review.group_id` gets nullified.

In the model we look up `history_element`s to decide when it was accepted/declined. Maybe we should add a new field?

adrianschroeter:
On Dienstag, 28. Juli 2020, 17:32:28 CEST wrote Henne Vogelsang:
> Not sure this does what you want it to do. `Review.updated_at` will change every time that the record changes. No matter if this is for a `state` change or not. Let's say because we delete a `Group` and `Review.group_id` gets nullified.

deleting a group must indeed not touch the timestamp, but it is also only
stored as text string therefore to be independend. If there is another vector
causing this we need to solve this indeed.

IMHO we should get rid of the belongs_to :group there.

(on the other hand, removing a group could maybe seen as very seldom special case no worth 
 to be taken care off....)

> In the model we look up `history_element`s to decide when it was accepted/declined. Maybe we should add a new field?

I like to avoid that if possible...

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

hennevogel:
>> Maybe we should add a new field?

> I like to avoid that if possible...

Because?

hennevogel:
Let me show you what I mean. Commit incoming...

adrianschroeter:
fine with me (even though we could drop update_at now since we have no usage for it anymore)

adrianschroeter:
hm, and what is with old requests where the data was still valid? 
Maybe we need a migration to rescue the data from update_at that otherwise we also break the data on the older requests now...

hennevogel:
> hm, and what is with old requests where the data was still valid?

In the data migration we copy over the date from `updated_at`. Until this data migration is run we cover this in the reader of the new attribute. It does `changed_state_at || updated_at`.

eduardoj:
I rebased on master.

hennevogel (approved):

PR #9943, opened 2020-07-22
[api] support plain release request

adrianschroeter:
This is basically the same functionality as release command for
packages, but it allows to involve review workflows.

A very much stripped down version of maintenance_release action,
not relying on incident setups.

OBS-51

### DIFF
--- a/docs/api/api/request.rng
+++ b/docs/api/api/request.rng
@@ -64,7 +64,8 @@
           <value>set_bugowner</value>
           <value>maintenance_incident</value>
           <value>maintenance_release</value>
-          <value>group</value>
+          <value>release</value>
+          <value>group</value> <!-- should not be used anymore -->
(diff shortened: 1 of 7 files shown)


hennevogel:
@adrianschroeter do you need help with the linter?

adrianschroeter:
On Montag, 27. Juli 2020, 18:44:23 CEST wrote Henne Vogelsang:
> @adrianschroeter do you need help with the linter?

I would actually just disable the check here...

-- 

Adrian Schroeter
email: adrian@suse.de

Maxfeldstraße 5                         
90409 Nürnberg 
Germany

Vogtinator:
It should be supported to release into a repository which has more arches than the source - that's what we're doing for openSUSE.

hennevogel (approved):

PR #9944, opened 2020-07-22
Remove session tracking in proxy mode

hennevogel:
This was an attempt to measure logins with proxy authentication based
on User.session. But User.session is a thread variable that is reset
after each request. So it's basically impossible to check based on this.
The user migh just have authenticated or we might just be on a new
thread ¯\_(ツ)_/¯

It also doesn't make much sense to measure :unauthenticated and
:disabled without measuring the rest of the case.

So let's stop measuring this at all until we find a better way.

For non-proxy auth this is still measured in Webui::SessionController.

### DIFF
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -169,7 +169,6 @@ class Webui::WebuiController < ActionController::Base
 
   def check_user
     @spider_bot = request.bot?
-    previous_user = User.possibly_nobody.login
     User.session = nil # reset old users hanging around
 
     user_checker = WebuiControllerService::UserChecker.new(http_request: request, config: CONFIG)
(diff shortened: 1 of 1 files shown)


src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -197,12 +195,6 @@ def check_user
     User.session = User.find_by_login(session[:login]) if session[:login]
 
     User.session ||= User.possibly_nobody
-
-    if !User.session
```
krauselukas, line 201: @hennevogel why did you remove this `if/else` block? doesn't seem to be related to the proxy mode

hennevogel: You are right, this code would also run without proxy mode. But it still would generate "wrong" data.

Once `User.session` is false we run an unauthenticated request, but that isn't data we want to track. And `previous_user` being different from `User.session` does *not* mean a log in happened. 

Let me split those to another commit...

hennevogel: Did not find a way to split this more intelligent so I made a new commit message. Please check again.

krauselukas: okay understand :+1:

PR #9946, opened 2020-07-23
[backend] try to catch disk full errors in bs_worker tasks

adrianschroeter:
Like downloading sources or binary packages.


mlschroe:
```
`$ex = 3 if $@ =~ /no space left/i;
```

PR #9952, opened 2020-07-24
Add theming support

### LINKED ISSUE #6311: Introduce theming per user and instance

hellcp:
This is a simple theming implementation, with a sample openSUSE theme (which shouldn't be used in production, I will leave it up to @guoyunhe to implement the remaining bits and pieces)

To verify this feature

- Change theme variable in config from `default` to `openSUSE`
- See a different look of OBS

No screenshots because you really shouldn't use the openSUSE theme yet

Fixes #6311


danidoni:
Nice! Thanks for the contribution!

Conan-Kudo:
@hellcp Any chance you'd rebase this to get it landed now?

hellcp:
Now is probably the worst time, because of the Bootstrap 5 update. I think that Henne wanted to talk about the implementation of this at some point too, since he doesn't agree with replicating the structure of the views directory in the themes. That's probably a discussion we need to have at some point though, and it's unlikely this would be merged without agreeing to some kind of a compromise

PR #9957, opened 2020-07-27
Fix rss notifications creation

### LINKED ISSUE #9900: [production][] undefined method `rss_token' for #<Group:0x000055e6c10b71a8>
threw this 3 times
<pre>
undefined method `rss_token&#39; for #&lt;Group:0x000055e6bed4dd20&gt;
</pre>

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5f0721943ed333000cb1261f)

DavidKang:
Group model doesn't have `rss_token`, using `try` will return `nil` if the method doesn't exist.

Fix #9900

### DIFF
--- a/src/api/app/services/notification_service/notifier.rb
+++ b/src/api/app/services/notification_service/notifier.rb
@@ -41,8 +41,7 @@ module NotificationService
     end
 
     def create_notification?(subscriber, channel)
-      return false if subscriber&.away?
-      return false if channel == :rss && !subscriber&.rss_token
+      return false if subscriber.nil? || subscriber.away? || (channel == :rss && !subscriber.try(:rss_token))
       return false unless notifiable_exists?
 
       true


src/api/app/services/notification_service/notifier.rb
```diff
@@ -41,8 +41,7 @@ def create_notification_per_subscription(subscription, channel)
     end
 
     def create_notification?(subscriber, channel)
-      return false if subscriber&.away?
-      return false if channel == :rss && !subscriber&.rss_token
+      return false if subscriber.nil? || subscriber.away? || (channel == :rss && !subscriber.try(:rss_token))
```
DavidKang, line 44: This is a small fix for a bigger one that should be handled in a different PR. We created an issue for that https://github.com/openSUSE/open-build-service/issues/9958

PR #9959, opened 2020-07-28
[api] introduce "manual" service mode

adrianschroeter:
It is just an alias for "disabled" on the server side, but osc may
make a difference here in future.

It is the better name for the todays usage of "disabled", while
"disabled" may mean in future really the original documented purpose
of keeping it disabled, just for reference.

### DIFF
--- a/docs/api/api/obs.rng
+++ b/docs/api/api/obs.rng
@@ -86,12 +86,13 @@
 
   <define ns="" name="service-modes">
     <choice>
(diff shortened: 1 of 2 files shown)


adrianschroeter:
Still in discussion here: https://github.com/openSUSE/osc/pull/826 and here:
 https://github.com/openSUSE/obs-docu/pull/154

PR #9960, opened 2020-07-28
Add a session identifier to influxdb

vpereira:
Having a session concept in the influxdb allows us to monitor
how many simultaneous users are accessing the
frontend

Example of panel in Grafana:

Screenshot_2020-07-28_12-15-36


vpereira:
there are still somethings to figure out. I will close it for now

PR #9967, opened 2020-07-30
bs_worker log output improvement

adrianschroeter:
newline after finished removing


src/backend/bs_worker
```diff
@@ -204,11 +204,12 @@ sub cleanup_job {
...
   # if this is a chroot like case, get rid of it after the build
   if ((!$vm || $vm eq 'lxc' || $vm eq 'docker' || $vm eq 'nspwan') && $cleanup_chroot && -d $buildroot) {
-      printf "Removing buildroot $buildroot";
+      printf "Removing buildroot $buildroot...";
       eval {
         rm_rf($buildroot);
         mkdir_p($buildroot);
-      }
+      };
+      printf "\n";
```
mlschroe, line 212: Please do not use printf if there is no `%` in the string. (Yes, I know this is copy and paste from above where it is also wrong.)

PR #9983, opened 2020-08-06
Keep the filter when marking a notification read or unread 

### LINKED ISSUE #9908: Marking a notification as read/unread drops filters
Peek 2020-07-13 12-08

danidoni:
Instead of picking the notification type from the `params` variable,
this PR makes the notification type explicitly available as a partial
local variable and function parameters.

Before this PR, the user was losing the selected notifiable type from the
notifications filter when marking a notification as 'read'.

This was due to the notifications/update remote partial not getting
the `params[:type]` when rendering.

Peek 2020-08-07 13-02

It's also fixed in small devices

Peek 2020-08-06 16-13

Fixes #9908

### DIFF
--- a/src/api/app/views/webui/users/notifications/index.html.haml
+++ b/src/api/app/views/webui/users/notifications/index.html.haml
@@ -9,8 +9,8 @@
       %strong.d-block.d-md-none.p-3{ data: { toggle: 'collapse', target: '#filters' },
                                   aria: { expanded: true, controls: 'filters' } }
         Filtered by: #{params[:type]&.humanize || @filtered_project || 'Unread'}
-        %i.float-right.mt-1.fa.fa-chevron-down
+        %i.float-right.mt-1.fa.fa-chevron-down#notifications-dropdown-trigger
       .card-body.collapse#filters
-        = render partial: 'notifications_filter', locals: { projects_for_filter: @projects_for_filter, notifications_count: @notifications_count }
+        = render partial: 'notifications_filter', locals: { filter: @notifications_filter }
   .col-md-8.col-lg-9#notifications-list
-    = render partial: 'notifications_list', locals: { notifications: @notifications }
+    = render partial: 'notifications_list', locals: { notifications: @notifications, selected_filter: @notifications_filter.selected_filter }
--- a/src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
+++ b/src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
@@ -1,0 +1,84 @@
+require 'browser_helper'
+
+RSpec.describe 'User notifications', type: :feature, js: true do
+  let!(:user) { create(:confirmed_user, :in_beta) }
+
+  describe 'when having no notifications' do
+    context 'when accessing the notifications page' do
+      before do
+        login user
+        visit my_notifications_path
+      end
+
+      it 'shows no notifications' do
+        expect(page).to have_text('There are no notifications')
+      end
+    end
+  end
+
+  describe 'when having notifications' do
+    let!(:notification_for_projects_comment) { create(:web_notification, :comment_for_package, subscriber: user) }
+    let!(:another_notification_for_projects_comment) { create(:web_notification, :comment_for_package, subscriber: user) }
+    let(:notifiable) { notification_for_projects_comment.notifiable }
+    let(:another_notifiable) { another_notification_for_projects_comment.notifiable }
+    let(:project) { notifiable.commentable.project }
+
+    context 'when clicking on the Comments filter' do
+      before do
+        login user
+        visit my_notifications_path
+      end
+
+      it 'shows all unread comment notifications' do
+        find('#notifications-dropdown-trigger').click if mobile?
+        within('#filters') { click_link('Comments') }
+        expect(page).to have_text(notifiable.commentable.name)
+      end
+
+      context 'when marking a comment notification as read' do # rubocop:todo RSpec/NestedGroups
+        before do
+          find('#notifications-dropdown-trigger').click if mobile?
+          within('#filters') { click_link('Comments') }
+          click_link("update-notification-#{notification_for_projects_comment.id}")
+        end
+
+        it 'keeps the Comments filter' do
+          wait_for_ajax
+
+          find('#notifications-dropdown-trigger').click if mobile?
+          expect(find('.list-group-item.list-group-item-action.active')).to have_text('Comment')
+        end
+      end
+    end
+
+    context 'when clicking on the project filter' do
+      before do
+        login user
+        project.notifications << notification_for_projects_comment
+        project.notifications << another_notification_for_projects_comment
+        visit my_notifications_path
+      end
+
+      it 'shows all unread project notifications' do
+        find('#notifications-dropdown-trigger').click if mobile?
+        within('#filters') { click_link(project.name) }
+        find('#notifications-dropdown-trigger').click if mobile?
+        expect(find('.list-group-item.list-group-item-action.active')).to have_text(project.name)
+      end
+
+      context 'when marking a project notification as read' do # rubocop:todo RSpec/NestedGroups
+        before do
+          find('#notifications-dropdown-trigger').click if mobile?
+          within('#filters') { click_link(project.name) }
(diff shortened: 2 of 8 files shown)


src/api/spec/features/webui/users/notifications_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'browser_helper'
+
+RSpec.describe 'User notifications', type: :feature, js: true, beta: true do
```
saraycp, line 3: _Beta_ is set for this whole test, so I'd suggest to move it to `spec/features/beta`. All the tests in that directory has beta enabled by default, so there is no need to add `beta: true`.

That behavior comes from: 
https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/support/beta.rb

danidoni:
> Testing in review-app, it seems we have the same problem when filtering by project name.
> 
> Could that be solved in a similar way?

I broke something with the refactor, I'll take a look

danidoni:
> > Testing in review-app, it seems we have the same problem when filtering by project name.
> > Could that be solved in a similar way?
> 
> I broke something with the refactor, I'll take a look

Fixed

src/api/app/models/notifications_filter.rb
```diff
@@ -0,0 +1,11 @@
+class NotificationsFilter
```
danidoni, line 1: I'm not sure where to put this model

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'browser_helper'
+
+RSpec.describe 'User notifications', type: :feature, js: true, beta: true do
```
eduardoj, line 3: As @saraycp pointed out, after moving the test into `spec/features/beta/`, you don't need to define the `beta: true` metadata.

src/api/app/models/notifications_filter.rb
dmarcoux: This looks like a presenter. A quite basic one, but still a presenter.

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
danidoni: True, I'll remove it

src/api/app/models/notifications_filter.rb
danidoni: Hmm, yes I thought about it before, but I wasn't sure. I'll try again.

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
danidoni: Done

src/api/app/models/notifications_filter.rb
danidoni: I've renamed the class as a presenter

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    let(:project) { notifiable.commentable.project }
+
+    context 'when clicking on the Comments filter' do
+      before do
+        login user
+        visit my_notifications_path
+      end
+
+      it 'shows all unread comment notifications' do
+        find(class: 'fa-chevron-down').click if mobile?
```
dmarcoux, line 33: This seems quite fragile. Could we have a selector which is more specific, something like `#notifications-filter-desktop i.fa-chevron-down`?

danidoni: Yes, good idea, I did that before, no reason to not doing it now.

danidoni: Even better, I'll add an id to the dropdown control

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    let(:project) { notifiable.commentable.project }
+
+    context 'when clicking on the Comments filter' do
+      before do
+        login user
+        visit my_notifications_path
+      end
+
+      it 'shows all unread comment notifications' do
+        find(id: '#notifications-dropdown-trigger').click if mobile?
```
dmarcoux, line 33: To be consistent with other feature specs: 
```suggestion
        find('#notifications-dropdown-trigger').click if mobile?
```

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        within('#filters') { click_link('Comments') }
+        expect(page).to have_text(notifiable.commentable.name)
+      end
+
+      context 'when marking a comment notification as read' do
+        before do
+          find(id: '#notifications-dropdown-trigger').click if mobile?
+          within('#filters') { click_link('Comments') }
+          find_link(id: format('update-notification-%d', notification_for_projects_comment.id)).click
```
dmarcoux, line 42: This should work, right?
```suggestion
          click_link("update-notification-#{notification_for_projects_comment.id}")
```

danidoni: Yyyesss

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        visit my_notifications_path
+      end
+
+      it 'shows all unread comment notifications' do
+        find('#notifications-dropdown-trigger').click if mobile?
+        within('#filters') { click_link('Comments') }
+        expect(page).to have_text(notifiable.commentable.name)
+      end
+
+      context 'when marking a comment notification as read' do # rubocop:todo RSpec/NestedGroups
```
danidoni, line 38: What about disabling this at project level? I find it too tight

dmarcoux: I would be fine with setting the maximum level of nesting to `2`.

https://rubocop.readthedocs.io/projects/rspec/en/latest/cops_rspec/#rspecnestedgroups

danidoni: Then let's do it

.rubocop.yml
```diff
@@ -31,6 +32,11 @@ Layout/DotPosition:
 Layout/EndAlignment:
   EnforcedStyleAlignWith: variable
 
+#################### RSpec ###########################
+
+RSpec/NestedGroups:
```
dmarcoux, line 37: I don't mind having this in the PR, but it should be in a different commit. It's not related to the notification filters.

danidoni: Yes, I'll do it in another PR, because including RSpec cops is uncovering a lot of issues. I'm going back to having it disabled locally from that specific spec.

src/api/app/views/webui/users/notifications/index.html.haml
```diff
@@ -9,8 +9,9 @@
...
                                   aria: { expanded: true, controls: 'filters' } }
         Filtered by: #{params[:type]&.humanize || @filtered_project || 'Unread'}
-        %i.float-right.mt-1.fa.fa-chevron-down
+        %i.float-right.mt-1.fa.fa-chevron-down#notifications-dropdown-trigger
       .card-body.collapse#filters
-        = render partial: 'notifications_filter', locals: { projects_for_filter: @projects_for_filter, notifications_count: @notifications_count }
+        = render partial: 'notifications_filter', locals: { filter: @notifications_filter }
   .col-md-8.col-lg-9#notifications-list
-    = render partial: 'notifications_list', locals: { notifications: @notifications }
+    - notifications_list_locals = { notifications: @notifications, selected_filter: @notifications_filter.selected_filter }
```
krauselukas, line 16: @danidoni its maybe nitpicking, but I would skip the step of assigning this to a variable, and just pass it directly to the locals

danidoni: Yes, fixed

krauselukas:
@danidoni LGTM, just squash the last commit and I will approve

PR #9992, opened 2020-08-07
[backend] allow building helm chart

sjamgade:
the mediaType change is as expected by helm documented here:
https://helm.sh/docs/topics/registries/
and
https://github.com/helm/helm/blob/d6f6184351f7aecd9d0efdfb8aba7da5664031eb/internal/experimental/registry/constants.go

### DIFF
--- a/src/backend/BSPublisher/Registry.pm
+++ b/src/backend/BSPublisher/Registry.pm
@@ -542,9 +542,12 @@ sub push_containers {
       close $tarfd if $tarfd;
 
       # put manifest into repo
+      # helm has a static set of mediatypes supported
+      # https://github.com/helm/helm/blob/d6f6184351f7aecd9d0efdfb8aba7da5664031eb/internal/experimental/registry/constants.go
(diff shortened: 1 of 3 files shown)


src/backend/BSPublisher/Registry.pm
```diff
@@ -542,9 +542,12 @@ sub push_containers {
       close $tarfd if $tarfd;
 
       # put manifest into repo
+      # helm has a static set of mediatypes supported
+      # https://github.com/helm/helm/blob/d6f6184351f7aecd9d0efdfb8aba7da5664031eb/internal/experimental/registry/constants.go
+      my $mediaType = "application/vnd.oci.image.manifest.v1+json" if (($containerinfo->{'type'} || '') eq 'helm');
```
mlschroe, line 547: So helm refuses to fetch charts without this change?

sjamgade: yup

PR #10008, opened 2020-08-12
[backend] support memory per job constraint

adrianschroeter:
To increase the memory need per used job on the worker. This can be
potentionally racy, because parallel builds may happened on different
times so the need of memory might differ depending on build speed.

### DIFF
--- a/src/backend/BSDispatcher/Constraints.pm
+++ b/src/backend/BSDispatcher/Constraints.pm
@@ -111,6 +111,7 @@ sub oracle {
     my $memory = ($worker->{'hardware'}->{'memory'} || 0);
     my $swap = ($worker->{'hardware'}->{'swap'} || 0);
     return 0 if $constraints->{'hardware'}->{'memory'} && getmbsize($constraints->{'hardware'}->{'memory'}) > ( $memory + $swap );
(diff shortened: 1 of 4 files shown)


src/backend/BSDispatcher/Constraints.pm
```diff
@@ -111,6 +111,7 @@ sub oracle {
     my $memory = ($worker->{'hardware'}->{'memory'} || 0);
     my $swap = ($worker->{'hardware'}->{'swap'} || 0);
     return 0 if $constraints->{'hardware'}->{'memory'} && getmbsize($constraints->{'hardware'}->{'memory'}) > ( $memory + $swap );
+    return 0 if $constraints->{'hardware'}->{'memory'} && getmbsize($constraints->{'hardware'}->{'memoryperjob'} * ($worker->{'hardware'}->{'jobs'} || 0)) > ( $memory + $swap );
```
mlschroe, line 114: copy/pasto: 'memory' is supposed to be 'memoryperjob'

Also: zero is a bad default for `$worker->{'hardware'}->{'jobs'}`, as this basically disables the check. Make this 1 instead.

adrianschroeter: right, fixed.

Do you have a better idea for the name btw? it is IMHO not so readable with three words connected, but underscore would break the scheme...

marxin:
@adrianschroeter One small note, can you please document the newly added param here:
https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.build_job_constraints.html
?

PR #10012, opened 2020-08-13
Restrict status messages showing on the rss feed

### LINKED ISSUE #9926: Control communication scope for RSS feed.
# Issue Description

We can create a status message with a specific communication scope. So just a certain group of people can see the message.

Right now, that restriction is not applied to RSS feed. So a status message restricted to **Logged in users** in OBS, can be read via the public RSS feed:  https://build.opensuse.org/main/news.rss

Expected Result
============

Apply communication scope restrictions to RSS feed as well.

danidoni:
The RSS feed is showing all status messages, ignoring its scope, so
everybody sees everything. This PR enforces the propper roles so the
RSS feed shows appropriate content to each user.

Fixes #9926

### DIFF
--- a/src/api/app/models/status_message.rb
+++ b/src/api/app/models/status_message.rb
@@ -6,6 +6,8 @@ class StatusMessage < ApplicationRecord
   validates :user, :severity, :message, presence: true
 
   scope :announcements, -> { order('created_at DESC').where(severity: 'announcement') }
+  scope :messages_for_me, -> { where(communication_scope: communication_scopes_for_current_user) }
+  scope :newest, -> { order('created_at DESC') }
 
   enum severity: { information: 0, green: 1, yellow: 2, red: 3, announcement: 4 }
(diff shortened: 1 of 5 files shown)


src/api/app/models/status_message.rb
```diff
@@ -6,6 +6,8 @@ class StatusMessage < ApplicationRecord
   validates :user, :severity, :message, presence: true
 
   scope :announcements, -> { order('created_at DESC').where(severity: 'announcement') }
+  scope :messages_for_me, -> { where(communication_scope: communication_scopes_for_current_user) }
```
hennevogel, line 9:
```suggestion
  scope :for_current_user, -> { where(communication_scope: communication_scopes_for_current_user) }
```

danidoni:
> Not sure this makes much sense as most of the times nobody will be logged in when accessing the feed....

I think it does. Now the RSS channel gets every single status message we have in the database. With this PR, when you're not logged in, you get messages allowed for non logged users, and when you are logged in, you get the messages allowed per your role.

PR #10021, opened 2020-08-18
Move footer legal to bottom

DavidKang:
### Before
Screenshot_2020-08-18 Open Build Service(1)

### After
Screenshot_2020-08-20 Open Build Service

### DIFF
--- a/src/api/app/assets/stylesheets/webui/responsive_ux/layout.scss
+++ b/src/api/app/assets/stylesheets/webui/responsive_ux/layout.scss
@@ -79,18 +79,19 @@
 
 // FOOTER
 
-#footer {
-  strong { color: $gray-700; }
-
+#footer, #footer-legal {
   & a {
     color: $gray-700;
     &:hover { color: $gray-900; }
   }
+}
 
+#footer {
+  strong { color: $gray-700; }
   li { @extend .my-2; }
+}
 
-  #footer-legal {
-    font-size: .9rem;
-    p { color: $gray-700; }
-  }
+#footer-legal {
+  font-size: .9rem;
+  p { color: $gray-700; }
 }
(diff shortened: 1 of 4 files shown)


dmarcoux:
I prefer the "Before" version since it feels less blocky with a more subtle separation, but let's see what others think.

saraycp:
> I prefer the "Before" version since it feels less blocky with a more subtle separation, but let's see what others think.

Without the sponsors section, the _Before_ version looks quite good. With sponsors looks weird for me. Maybe moving them somewhere else?

I see a few advantages on the white background:
- The footer goes unnoticed with the _Before_ version.  Both white background or a more prominent separation line can do the trick.
- Some of the sponsors have a white background, so they look better on top of white.

dmarcoux:
>     * Some of the sponsors have a white background, so they look better on top of white.

The logos with a white background should have a transparent background instead, making them look good on all backgrounds.

saraycp:
> The logos with a white background should have a transparent background instead, making them look good on all backgrounds.

Yes, we can do it as well, then we have to find a better way to display the sponsor section.

dmarcoux:
In the following screenshot, I've changed the border color by switching the class `border-gray-300` to `border-gray-400`. Let me know what you think of this.

footer separation with border-gray-400

saraycp:
I think the darker border makes it clearer.

Apart from that, what do you think about moving the sponsors inside the footer and the `Open Build Service (OBS)..` to the end of the page?

My proposal draft (needs adjusting the style a bit) is:
sponsor_into_footer

dmarcoux:
I like your proposal @saraycp. Let's do that and use the darker border.

DavidKang:
@dmarcoux, I applied the changes :smile_cat:

dmarcoux (approved):

PR #10007, opened 2020-08-12
Implement left-top-top navigation design for the responsive_ux layout

dmarcoux:
The `responsive_ux` layout didn't change on small devices, so this explains why previews below are only for large devices.

Before:
before

After:
after

PR #10027, opened 2020-08-19
Prevent content from breaking the responsive-ux layout

dmarcoux:
This is something which was somewhat addressed in #10007. The tabs didn't break the layout anymore, but other elements still do. Limiting the width of specific elements could turn out to be a long process. It's better and simpler to tackle this issue for the content container.

Just like before, tabs are not overflowing:
tabs

As for the other elements, like the diffs on the request show page, they were overflowing. It's now fixed.
Before:
request-show-before
After:
request-show-after


DavidKang (approved):

saraycp (approved):
It works! :rocket:

PR #10028, opened 2020-08-19
Use real-time specific Sphinx Rake tasks

dmarcoux:
The generic tasks are also dealing with SQL-backed indices, which we don't use.

We'll have to be sure that SQL-backed indices are really gone from our reference server / IBS.

### DIFF
--- a/ReleaseNotes-2.11
+++ b/ReleaseNotes-2.11
@@ -29,8 +29,8 @@ Features
 Generic:
  * Search was improved: New and updated projects/packages are indexed directly instead
    of once every hour. As configuration files for Thinking Sphinx were changed, running
-   a 'ts:rebuild' task is needed after deployment, as user wwwrun:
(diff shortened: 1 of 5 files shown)


dmarcoux:
[find solution for TS indexing issues](https://trello.com/c/zALynfru/1448-find-solution-for-ts-indexing-issues)

hennevogel:
Hm, how is changing to the `:rt:` namespace making sure that SQL-backed indices are really gone?

dmarcoux:
According to the quote in the PR's description, we'll have to delete our index and binlog files to ensure that the SQL-backed indices are gone. The changes from this PR are only ensuring that we're not creating SQL-backed indices.

hennevogel:
> we'll have to delete our index and binlog files

We have in `ReindexSphinxDb` right?

dmarcoux:
> We have in `ReindexSphinxDb` right?

It's from `2015`. We switched to real-time indices in October 2019.

hennevogel:
> It's from 2015.

Then we should have this also now or not? At least in the release notes.

vpereira (approved):

PR #10038, opened 2020-08-20
Correct Style/IfUnlessModifier offenses in controllers

DavidKang:
This PR is the first of a series of PR's to tackle Style/IfUnlessModifier offenses

PR #10040, opened 2020-08-20
Correct Style/IfUnlessModifier offenses - Part III

DavidKang:
This PR is the continuation of #10038


dmarcoux:
There's a typo in the last commit message. It should be `mixins` instead of `mixings`.

PR #10046, opened 2020-08-24
[backend] BSPublisher/Container.pm allow to specify repo

bugfinder:
along with the project in BSConfig::publish_containers


mlschroe:
You should also update BSConfig.pm.template

PR #10065, opened 2020-08-24
Stay away from nested transactions

hennevogel:
No idea why this operation is in a transaction. But `Project.store`
runs in a nested transation then. If something goes wrong with `.store`
exceptions are ignored.

### DIFF
--- a/src/api/app/models/branch_package.rb
+++ b/src/api/app/models/branch_package.rb
@@ -90,6 +90,12 @@ class BranchPackage
 
   private
 
+  def add_request_cloned_attribute(project)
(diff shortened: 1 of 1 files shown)


hennevogel:
See https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html#module-ActiveRecord::Transactions::ClassMethods-label-Nested+transactions

hennevogel:
I have the suspicion that this often "ate" creation failures...

dmarcoux (approved):

PR #10071, opened 2020-08-25
Append ThinkingSphinx callbacks to Package and Project models

### LINKED ISSUE #9436: [production][search#index] Record IDs found by Sphinx but not by ActiveRecord : 5259490, 5250055
search#index threw this 40 times
<pre>
Record IDs found by Sphinx but not by ActiveRecord : 681698, 682295, 682304
</pre>
[https://build.opensuse.org/search?name=1&amp;package=1&amp;page=4&amp;project=1&amp;search_text=rpm-packaging](https://build.opensuse.org/search?name=1&amp;package=1&amp;page=4&amp;project=1&amp;search_text=rpm-packaging)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5e8279f957ce55000f1b5b83)

dmarcoux:
Deleted packages/projects are now removed from Sphinx indices. Before, we were missing the `after_destroy` callback from
`ThinkingSphinx::ActiveRecord::Callbacks::DeleteCallbacks`. This resulted in having deleted packages/projects in our Sphinx indices. The error was `ThinkingSphinx::Search::StaleIdsException` with the message `Record IDs found by Sphinx but not by ActiveRecord : 123, 456, 789`.

See upstream:
https://github.com/pat/thinking-sphinx/blob/df2746b5679c8009527ac87b761357d536d48c05/lib/thinking_sphinx/callbacks.rb#L10
https://github.com/pat/thinking-sphinx/blob/7622e5e40fe8a79a3329ddc78d54390cfe7a7f02/lib/thinking_sphinx/callbacks/appender.rb#L16

This fixes #9436.

**TODO after merging this PR:**
We will have to cleanup the indices and to do so, we need to rebuild them.
```shell
# We need to remove the existing indices
cd to_our_application
rm tmp/binlog db/sphinx
run_in_api bundle exec rake ts:rt:rebuild 
```

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -4,6 +4,7 @@ require 'rexml/document'
 
 # rubocop: disable Metrics/ClassLength
 class Package < ApplicationRecord
+  include AppendSphinxCallbacks
   include FlagHelper
   include Flag::Validations
   include CanRenderModel
@@ -11,7 +12,6 @@ class Package < ApplicationRecord
   include Package::Errors
   include HasRatings
   include HasAttributes
-  include PopulateSphinx
   include PackageSphinx
 
   BINARY_EXTENSIONS = ['.0', '.bin', '.bin_mid', '.bz', '.bz2', '.ccf', '.cert',
@@ -69,7 +69,6 @@ class Package < ApplicationRecord
   before_destroy :remove_devel_packages
 
   after_save :write_to_backend
-  after_save :populate_sphinx, if: -> { name_previously_changed? || title_previously_changed? || description_previously_changed? }
   after_rollback :reset_cache
 
   # The default scope is necessary to exclude the forbidden projects.
--- a/src/api/app/mixins/populate_sphinx.rb
+++ b/src/api/app/mixins/populate_sphinx.rb
@@ -1,7 +1,0 @@
-module PopulateSphinx
-  def populate_sphinx
-    ThinkingSphinx::RealTime::Callbacks::RealTimeCallbacks.new(
-      self.class.name.underscore.to_sym
-    ).after_save(self)
-  end
-end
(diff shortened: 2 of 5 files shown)


dmarcoux:
[Find solution for TS indexing issues](https://trello.com/c/zALynfru/1448-find-solution-for-ts-indexing-issues)

src/api/spec/models/project_spec.rb
```diff
@@ -768,4 +768,11 @@ def reset_project_in_backend
 
     it { expect(project.project_state).not_to be_nil }
   end
+
+  describe 'sphinx' do
```
hennevogel, line 772: We have `RSpec.describe FullTextSearch`...

dmarcoux: I was looking for exactly that :slightly_smiling_face:

dmarcoux:
The specs I added are also passing on master, so without my changes. I really don't get it since I've experienced the opposite in my manual testing. This is what I've done:

1. I created a package in the web UI.
2. I checked if the package is in the indices. It was.
3. I deleted the package in the web UI.
4. I checked if the package is in the indices. It was not.

Without the changes from this PR, the package was still in the indices in the 4th step.

src/api/app/models/package.rb
```diff
@@ -21,6 +20,15 @@ class Package < ApplicationRecord
                        '.taz', '.tb2', '.tbz', '.tbz2', '.tgz', '.tlz', '.txz', '.ucode',
                        '.xpm', '.xz', '.z', '.zip', '.ttf'].freeze
 
+  ThinkingSphinx::Callbacks.append(self, behaviours: [:real_time]) do |package|
+    # Index package only if its name, title or description changed
+    if package.name_previously_changed? || package.title_previously_changed? || package.description_previously_changed?
+      [package]
```
vpereira, line 26: can you move those callbacks to a concern? like https://api.rubyonrails.org/classes/ActiveSupport/Concern.html ?

vpereira (changes requested):

vpereira (approved):

src/api/app/models/package.rb
dmarcoux: Sure!

PR #10076, opened 2020-08-26
Allow staff users to create/update/destroy status messages

dmarcoux:
We often have the need to create (and sometimes destroy) status messages on our reference server. Allowing staff users to do this is much better than turning everybody into an admin.

### DIFF
--- a/src/api/app/controllers/webui/status_messages_controller.rb
+++ b/src/api/app/controllers/webui/status_messages_controller.rb
@@ -1,14 +1,13 @@
 class Webui::StatusMessagesController < Webui::WebuiController
-  # permissions.status_message_create
-  before_action :require_admin, only: [:destroy, :create]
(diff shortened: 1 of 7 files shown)


src/api/app/controllers/webui/status_messages_controller.rb
```diff
@@ -1,14 +1,13 @@
...
+    authorize StatusMessage
+  end
 
   def create
-    # TODO: make use of permissions.status_message_create
-    status_message = StatusMessage.new(user: User.session!,
-                                       message: params[:status_message][:message],
-                                       severity: params[:status_message][:severity],
-                                       communication_scope: params[:status_message][:communication_scope])
+    status_message = authorize StatusMessage.new(status_message_params)
```
hennevogel, line 10: Cool, didn't know about chaining. :+1:

hennevogel (approved):

PR #10094, opened 2020-08-28
Cleanup dev-env

hennevogel:
Some random cleanups for our dev-env


docker-compose.selenium.yml
```diff
@@ -1,14 +0,0 @@
-version: "2.1"
```
krauselukas, line 1: Hmm, I found it very useful sometimes for debugging. The visual output helped with nasty feature tests.

krauselukas: But still possible to create screenshots in the tests, so I guess I could live without it

vpereira (approved):

PR #10097, opened 2020-08-31
Show download link for files that exceed 4 megabytes

### LINKED ISSUE #10096: Unable to download some tarballs from webui when logged-in
# Issue Description
I am unable to download _some_ source tarballs from webui when logged-in.

Expected Result
============
Download icon should be shown.

How to Reproduce
===============
1. Login to build.o.o
2. Go to: https://build.opensuse.org/package/show/openSUSE:Factory:ARM/efl
3. No download icon for `efl-1.24.3.tar.xz`

Further Information
===============
* https://build.opensuse.org/package/show/openSUSE:Factory:ARM/efl

krauselukas:
In the past we showed the download link for files with
a size of less than 4 megabytes also to non authenticated users.
Now we show the link only to authenticated users no matter which
sizes the file has. A forgotten if statement led to the problem
that we didn't showed the download link to anyone anymore when
the file exceeds the 4 megabytes.

Fixes #10096

### DIFF
--- a/src/api/app/views/webui/package/_file.html.haml
+++ b/src/api/app/views/webui/package/_file.html.haml
@@ -13,9 +13,8 @@
   / limit download for anonymous user to avoid getting killed by crawlers
   - unless nobody
     %td.text-center
-      - if file[:size].to_i < 4.megabytes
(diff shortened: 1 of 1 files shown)


hennevogel (approved):
oops

PR #10101, opened 2020-08-31
[api][backend] support an explicit marker to generate diffs from usin…

adrianschroeter:
…g obsgendiff

release targets may have a trigger="obsgendiff" to specify the final
release area. This will be used as base to generate the diffs (aka
ChangeLogs) from.

The typical use case is a setup like

 Build Area => Candidate Area => Final Release Area

The Build Area would have a release target with trigger=manual pointing
to Candidate Area. And another release target pointing to Final Release
Area using trigger=obsgendiff


mlschroe:
The backend part looks ok.

adrianschroeter:
this has nothing to do with reference server.

I would fix the conflicts, but only if someone is promissing to merge it in time (otherwise it is just wasted time till next schema conflict)

hennevogel (approved):
For the frontend part...

PR #10102, opened 2020-09-01
Remove wizardtemplate.spec

dmarcoux:
We don't use this file. It's been untouched for 12 years...

It used to be part of a wizard workflow. :mage:


dist/obs-server.spec
```diff
@@ -931,7 +931,6 @@ usermod -a -G docker obsservicerun
 %{__obs_api_prefix}/db/data
 %{__obs_api_prefix}/db/migrate
 %{__obs_api_prefix}/db/seeds.rb
-%{__obs_api_prefix}/files/wizardtemplate.spec
```
hennevogel, line 934: Breaks the package build...

hennevogel (changes requested):

dist/obs-server.spec
dmarcoux: The package build is failing due to those errors:
```
RPM build errors:
     File listed twice: /usr/lib/obs/server/BSConfig.pm
     Directory not found: /home/abuild/rpmbuild/BUILDROOT/obs-server-2.11~alpha.20200901T093837.fab91c9552-2.1.x86_64/srv/www/obs/api/files
```

hennevogel: Yeah, you remove `files` directory but we explicitly package that somewhere...

dmarcoux: Oh right, I forgot to remove that directory from the spec file.

dmarcoux:
@hennevogel Could you have another look please?

hennevogel (approved):

PR #10103, opened 2020-09-01
Show confirmation dialog on status message delete

### LINKED ISSUE #10087: Confirm deletion of status messages
Deleting a status message doesn't require a confirmation. All destructive actions should require confirmation to prevent user errors.

status-message

As you can see in the screenshot above, clicking on the delete icon should trigger a confirmation modal. We already have a few modals doing exactly the same on other pages. The comments deletion should be the closest one in terms of implementation.

OleksandrOrlov:
There were no confirmation while removing a status message. To prevent
accidental removing, all the destructive actions should contain a
confirmation.

The commit adds the appropriate modal dialog.

Fixes #10087

To verify this feature

1. Log in to local instance of OBS;
2. Create new status message (if there are no existing ones on the Home Page);
3. Click 'X' (Delete status message) icon for the status message.

Expected result:
"Delete status message?" confirmation dialog is shown. After clicking "Delete", the dialog should disappear, status message should be deleted.

Here is the screenshot of how it looks:
delete_status_message

### DIFF
--- a/src/api/app/views/webui/main/_status_message.html.haml
+++ b/src/api/app/views/webui/main/_status_message.html.haml
@@ -9,7 +9,8 @@
         #{time_ago_in_words(status_message.created_at)} ago
       - if policy(status_message).destroy?
         .float-right
-          = link_to(status_message_path(status_message), method: :delete, title: 'Remove status message') do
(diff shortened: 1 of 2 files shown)


saraycp (approved):
Nice first contribution! :clap:

dmarcoux (approved):

PR #10126, opened 2020-09-07
[backend] obsworker executation without detaching

adrianschroeter:
"obsworker run" is the same as "start" but is not detaching.
Needed for container deployments


eduardoj:
There is something already done in this PR: #9271.

adrianschroeter:
good spot, taking the older one

PR #10187, opened 2020-09-22
Render flash message now in notifications#update

dmarcoux:
Using `flash[:some_key]` will keep the flash for the next request, so this means that when we set the flash in an action called through an AJAX request, the flash will still be displayed when we reload the page.


OleksandrOrlov (approved):
LGTM

eduardoj (approved):

PR #10189, opened 2020-09-22
Add ability to mark multiple notifications as read/unread

krauselukas:
When dealing with a big amount of notifications, it is very
time consuming to mark them one by one.
Therefore a way to update multiple notifications at once
is required.

**Desktop**
Peek 2020-09-30 15-23

**Mobile**
Peek 2020-09-30 15-24

This PR is work in progress since I want to implement a way that enables the user to literally mark everything as read (not just the items that are paginated in the notification list)

### DIFF
--- a/src/api/app/assets/javascripts/webui/notification.js
+++ b/src/api/app/assets/javascripts/webui/notification.js
@@ -1,0 +1,32 @@
+function setSelectAllCheckbox() {
+  $('#select-all-notifications').change(function() {
+    var checkboxes = $(this).closest('form').find('input[type=checkbox]');
+    checkboxes.prop('checked', $(this).is(':checked'));
+  });
+}
+
+function setCheckboxCounterAndSubmitButton() {
+  var amountBoxesChecked = 0;
+  $("input[id^='notification_ids_']").each(function() {
+    if($(this).is(':checked')){
+      amountBoxesChecked += 1;
+    }
+  });
+
+  if(amountBoxesChecked <= 0) {
+    $('#done-button').addClass('disabled');
+    $('#select-all-label').text('Select all');
+  } else {
+    $('#done-button').removeClass('disabled');
+    $('#select-all-label').text(amountBoxesChecked + ' selected');
+  }
+}
+
+function handleNotificationCheckboxSelection() { // jshint ignore:line
+  setCheckboxCounterAndSubmitButton();
+  setSelectAllCheckbox();
+  $('input[type="checkbox"]').change(function() {
+    setCheckboxCounterAndSubmitButton();
+  });
+}
+
--- a/src/api/app/assets/stylesheets/webui/notifications_redesign/notifications.scss
+++ b/src/api/app/assets/stylesheets/webui/notifications_redesign/notifications.scss
@@ -1,5 +1,6 @@
 // TODO: As soon as notifications_redesign feature is 100% rolled out,
 // all the 'body.notifications-redesign' wrappers can be removed.
+$notifications-filter-box-height: 3.5rem;
 
 body.notifications-redesign  {
   .avatars-counter, .simulated-avatar {
@@ -26,6 +27,9 @@ body.notifications-redesign  {
       &.show { border-top: 1px solid $gray-300;}
     }
   }
+  #notification-action-bar {
+    &.sticky-top { top: $top-navigation-height; }
+  }
 }
 
 @include media-breakpoint-up(md) {
@@ -41,12 +45,19 @@ body.notifications-redesign  {
     #notifications-filter-desktop {
       &.show { border-top: 1px solid $gray-300; }
       &.sticky-top { top: $top-navigation-height; }
+      height: $notifications-filter-box-height;
+      // To not overlap with the notification action bar
+      z-index: calc(#{$zindex-sticky} + 1);
+
       #filters {
         max-height: 100vw; overflow: auto;
         -webkit-box-shadow: 2px 3px 5px rgba(0,0,0,.2);
         -moz-box-shadow: 2px 3px 5px rgba(0,0,0,.2);
         box-shadow: 2px 3px 5px rgba(0,0,0,.2);
       }
+    }
+    #notification-action-bar {
+      &.sticky-top { top: calc(#{$top-navigation-height} + #{$notifications-filter-box-height}); }
     }
   }
 
--- a/src/api/app/controllers/webui/users/notifications_controller.rb
+++ b/src/api/app/controllers/webui/users/notifications_controller.rb
@@ -7,26 +7,25 @@ class Webui::Users::NotificationsController < Webui::WebuiController
   before_action :check_feature_toggle
 
   def index
-    @notifications = fetch_notifications
+    @notifications = paginated_notifications
     @filtered_project = Project.find_by(name: params[:project])
     @notifications_filter = notifications_filter
   end
 
   def update
-    notification = User.session.notifications.find(params[:id])
-    authorize notification, policy_class: NotificationPolicy
+    notifications = fetch_notifications.where(id: params[:notification_ids])
 
-    if notification.toggle(:delivered).save
-      flash.now[:success] = "Successfully marked the notification as #{notification.unread? ? 'unread' : 'read'}"
-    else
-      flash.now[:error] = "Couldn't mark the notification as #{notification.unread? ? 'read' : 'unread'}"
+    # rubocop:disable Rails/SkipsModelValidations
+    unless notifications.update_all('delivered = !delivered')
+      flash.now[:error] = "Couldn't mark the notifications as #{notifications.first.unread? ? 'read' : 'unread'}"
     end
+    # rubocop:enable Rails/SkipsModelValidations
 
     respond_to do |format|
       format.html { redirect_to my_notifications_path }
       format.js do
         render partial: 'update', locals: {
-          notifications: fetch_notifications,
+          notifications: paginated_notifications,
           notifications_filter: notifications_filter
         }
       end
@@ -74,12 +73,16 @@ class Webui::Users::NotificationsController < Webui::WebuiController
   end
 
   def fetch_notifications
-    notifications_for_subscribed_user = User.session.notifications.for_web
-    notifications = if params[:project]
-                      NotificationsFinder.new(notifications_for_subscribed_user).for_project_name(params[:project])
-                    else
-                      NotificationsFinder.new(notifications_for_subscribed_user).for_notifiable_type(params[:type])
-                    end
+    notifications_for_subscribed_user = NotificationsFinder.new(policy_scope(Notification))
+    if params[:project]
+      notifications_for_subscribed_user.for_project_name(params[:project])
+    else
+      notifications_for_subscribed_user.for_notifiable_type(params[:type])
+    end
+  end
+
+  def paginated_notifications
+    notifications = fetch_notifications
     params[:page] = notifications.page(params[:page]).total_pages if notifications.page(params[:page]).out_of_range?
     params[:show_all] ? show_all(notifications) : notifications.page(params[:page])
   end
--- a/src/api/app/policies/notification_policy.rb
+++ b/src/api/app/policies/notification_policy.rb
@@ -1,6 +1,7 @@
 class NotificationPolicy < ApplicationPolicy
-  def update?
-    return true if record.subscriber_type == 'User' && record.subscriber_id == user.id
-    return true if record.subscriber_type == 'Group' && record.subscriber_id.include?(user.groups.ids)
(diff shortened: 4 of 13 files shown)


krauselukas:
[Mark multiple notifications as read/unread](https://trello.com/c/UIZa6hDL/1538-mark-multiple-notifications-as-read-unread)

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -13,13 +13,10 @@ def index
...
-    notification = User.session.notifications.find(params[:id])
-    authorize notification, policy_class: NotificationPolicy
+    notifications = policy_scope(Notification).where(id: params[:notification_ids])
 
-    if notification.toggle(:delivered).save
-      flash.now[:success] = "Successfully marked the notification as #{notification.unread? ? 'unread' : 'read'}"
-    else
-      flash.now[:error] = "Couldn't mark the notification as #{notification.unread? ? 'read' : 'unread'}"
+    unless notifications.update_all('delivered = !delivered')
+      flash[:error] = "Couldn't mark the notifications as #{notifications.first.unread? ? 'read' : 'unread'}"
```
dmarcoux, line 19: We need to keep `.now` to not reintroduce the issue fixed in #10187.

```suggestion
      flash.now[:error] = "Couldn't mark the notifications as #{notifications.first.unread? ? 'read' : 'unread'}"
```

src/api/config/routes/webui_routes.rb
```diff
@@ -306,7 +306,8 @@
     scope :my do
       resources :tasks, only: [:index], controller: 'webui/users/tasks', as: :my_tasks
 
-      resources :notifications, only: [:index, :update], controller: 'webui/users/notifications', as: :my_notifications
+      resources :notifications, only: [:index], controller: 'webui/users/notifications', as: :my_notifications
```
dmarcoux, line 309: What is the reason behind this change?

src/api/app/assets/javascripts/webui/notification.js
```diff
@@ -0,0 +1,32 @@
...
+      amountBoxesChecked += 1;
+    }
+  });
+
+  if(amountBoxesChecked <= 0) {
+    $('#done-button').addClass('d-none');
+    $('#select-all-label').text('Select all');
+  } else {
+    $('#done-button').removeClass('d-none');
+    $('#select-all-label').text('selected ' + amountBoxesChecked);
```
dmarcoux, line 21: It reads much better like this:

```suggestion
    $('#select-all-label').text(amountBoxesChecked + ' selected');
```

src/api/app/controllers/webui/users/notifications_controller.rb
krauselukas: Ahh, sorry this happend during the rebasing, was not by purpose

src/api/config/routes/webui_routes.rb
krauselukas: To the plural `resources` I must pass the id to the path helper, which doesn't work out with the form and the checkboxes

krauselukas: But I guess there is a more elegant way, so I will have another look

src/api/app/controllers/webui/users/notifications_controller.rb
dmarcoux: No worries, it happens :slightly_smiling_face:

danidoni (dismissed):

danidoni (approved):

danidoni (approved):

src/api/app/assets/stylesheets/webui/notifications_redesign/notifications.scss
```diff
@@ -1,19 +1,21 @@
 // TODO: As soon as notifications_redesign feature is 100% rolled out,
 // all the 'body.notifications-redesign' wrappers can be removed.
+$notifications-filter-box-height: 56px;
```
dmarcoux, line 3: I suggest using a responsive unit:

```suggestion
$notifications-filter-box-height: 3.5rem;
```

krauselukas:
This is ready for another review

DavidKang:
Nice @krauselukas!, I just have a question and is more for curiosity. Why the checkbox is on the second line, instead of been next to the notification icon (first line)? It was intended to be like that?

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
```diff
@@ -9,37 +9,49 @@
...
+                - if notification.notifiable_type == 'BsRequest'
+                  = image_tag('icons/request-icon.svg', height: 18, title: 'Request notification')
+                  = link_to(notification.notifiable_link[:text], notification.notifiable_link[:path], class: 'mx-1 text-word-break-all')
+                  %span.badge{ class: "badge badge-#{request_badge_color(notification.notifiable.state)}" }
+                    = notification.notifiable.state
+                - else
+                  %i.fas.fa-comments{ title: 'Comment notification' }
+                  = link_to(notification.notifiable_link[:text], notification.notifiable_link[:path], class: 'mx-1 text-word-break-all')
+              .action.ml-auto.align-self-end.align-self-md-start
+                - title, icon = notification.unread? ? ['Mark as "Read"', 'fa-check'] : ['Mark as "Unread"', 'fa-undo']
```
danidoni, line 40: I think you can use [this helper](https://github.com/openSUSE/open-build-service/pull/10220/files#diff-249c5d6caddd60654f20780971650cecR53).

danidoni (changes requested):

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
dmarcoux: This button and the one rendered by the helper are a bit different. This button only has an icon while the helper button has some text. There's also an extra class (`px-3`) on the `link_to`.

The helper could be adapted for sure, this just has to be kept in mind.

krauselukas:
> Nice @krauselukas!, I just have a question and is more for curiosity. Why the checkbox is on the second line, instead of been next to the notification icon (first line)? It was intended to be like that?

@DavidKang the idea was to have some distance to the link, so you don't click on the link instead of the checkbox by accident. But as soon as this PR is rebased with the new layout from @saraycp this might get some fine tuning

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
krauselukas: but this reminds me, that I have to adapt the helper a bit, since the notification#update method changed a bit

danidoni (approved):

danidoni:
LGTM! Nice work @krauselukas!

krauselukas:
> LGTM! Nice work @krauselukas!

Thanks, sorry I forgot to reuse the helper you mentioned, I will push once more

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -74,12 +73,16 @@ def notifications_count
   end
 
   def fetch_notifications
-    notifications_for_subscribed_user = User.session.notifications.for_web
-    notifications = if params[:project]
-                      NotificationsFinder.new(notifications_for_subscribed_user).for_project_name(params[:project])
-                    else
-                      NotificationsFinder.new(notifications_for_subscribed_user).for_notifiable_type(params[:type])
-                    end
+    notifications_for_subscribed_user = policy_scope(Notification)
```
dmarcoux, line 76: To remove the duplication in the `if/else` below.

```suggestion
    notifications_for_subscribed_user = NotificationsFinder.new(policy_scope(Notification))
    if params[:project]
      notifications_for_subscribed_user.for_project_name(params[:project])
    else
      notifications_for_subscribed_user.for_notifiable_type(params[:type])
    end
```

src/api/app/policies/notification_policy.rb
```diff
@@ -1,6 +1,7 @@
 class NotificationPolicy < ApplicationPolicy
-  def update?
-    return true if record.subscriber_type == 'User' && record.subscriber_id == user.id
-    return true if record.subscriber_type == 'Group' && record.subscriber_id.include?(user.groups.ids)
+  class Scope < Scope
+    def resolve
+      scope.for_web.where(subscriber_type: 'User', subscriber_id: user.id).or(scope.where(subscriber_type: 'Group', subscriber_id: user.groups.ids))
```
dmarcoux, line 4: Could we use `NotificationsFinder` here? The `where(...)` is doing the same as `NotificationsFinder#for_subscribed_user`:

https://github.com/openSUSE/open-build-service/blob/ab37c65a7276d93a377a1cc0a5a33e87aba161d5/src/api/app/queries/notifications_finder.rb#L22-L25

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
```diff
@@ -9,40 +9,50 @@
...
+      .card-body
+        .text-center
+          %span.ml-3= page_entries_info notifications, entry_name: 'notification'
+          = link_to_all unless notifications.total_pages == 1 && params['show_all'].nil?
+        .list-group.list-group-flush.mt-3
+          - notifications.each do |n|
+            - notification = NotificationPresenter.new(n)
+            .list-group-item.px-0.px-md-1.py-3
+              .row
+                .col-auto
```
dmarcoux, line 27: To reduce the unneeded gap between the checkbox and the notification's content, I would suggest this:

```suggestion
                .col-auto.pr-0
```

danidoni (approved):

krauselukas:
@danidoni sorry, I realized that the params passed in the helper method are slightly different, I dropped the changes. It needs more adjustements in order to reuse the helper. I think its not worth, the code will lose readability...

src/api/config/routes/webui_routes.rb
dmarcoux: I've found a better way to have this. It's much more explicit:

```suggestion
      resources :notifications, only: [:index], controller: 'webui/users/notifications', as: :my_notifications do
        collection do
          # We allow updating multiple notifications in a single HTTP request
          put :update
        end
      end
```

With this change, the path helper `my_notification_path` will have to be changed to `my_notifications_path`.

danidoni:
> @danidoni sorry, I realized that the params passed in the helper method are slightly different, I dropped the changes. It needs more adjustement in order to reuse the helper. I think its not worth, the code will lose readability...

Okay, nvm

dmarcoux (changes requested):

krauselukas:
@dmarcoux I applied you suggestions, thanks! :)

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
```diff
@@ -23,40 +23,62 @@
...
+      it_behaves_like 'keeps the Comments filter'
     end
 
     context 'when clicking on the project filter' do
       before do
-        login user
         project.notifications << notification_for_projects_comment
         project.notifications << another_notification_for_projects_comment
-        visit my_notifications_path
+        visit current_path
```
dmarcoux, line 81: I don't understand this change. Where is `current_path` coming from?

krauselukas: I can also keep `my_notifications_path` if that is more clear

dmarcoux: Yes, I think it's clearer to keep `my_notifications_path` and even adding a comment.

krauselukas: ok

krauselukas: done

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
```diff
@@ -23,39 +23,63 @@
...
+
+      it_behaves_like 'keeps the Comments filter'
     end
 
     context 'when clicking on the project filter' do
       before do
-        login user
         project.notifications << notification_for_projects_comment
         project.notifications << another_notification_for_projects_comment
+        # need to load the page again in order to have the notifcations
```
dmarcoux, line 81:
```suggestion
        # need to load the page again in order to have the notifications
```

dmarcoux (approved):

src/api/spec/features/beta/webui/users/notifications_controller_spec.rb
krauselukas: Arghh :D

krauselukas: fixed

PR #10192, opened 2020-09-22
Truncate multi-line notification excerpts to the first line

eduardoj:
In case of multi-line descriptions of notifications, joining the first lines of notifications' descriptions is confusing.

Truncate the excerpt to the first line, showing three dots at the end of the excerpt. This provides a more understandable message.

### DIFF
--- a/src/api/app/presenters/notification_presenter.rb
+++ b/src/api/app/presenters/notification_presenter.rb
@@ -1,4 +1,7 @@
 class NotificationPresenter < SimpleDelegator
+  TRUNCATION_LENGTH = 100
+  TRUNCATION_ELLIPSIS_LENGTH = 3 # `...` is the default ellipsis for String#truncate
+
   def initialize(model)
     @model = model
     super(@model)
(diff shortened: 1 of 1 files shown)


src/api/app/presenters/notification_presenter.rb
```diff
@@ -37,7 +45,7 @@ def excerpt
             else
               ''
             end
-    text.to_s.truncate(100)
+    truncate_to_first_new_line(text.to_s)
```
dmarcoux, line 48: `truncate_to_first_new_line` could be replaced with:
```suggestion
    truncation_ellipsis_length = 3
    truncation_index = (text.index("\n").try(:+, truncation_ellipsis_length) || TRUNCATION_LENGTH).clamp(0, TRUNCATION_LENGTH)
    text.truncate(truncation_index)
```

eduardoj: Thanks for the suggestion, but:
- I find the alternative not so easy to read, and
- I have done some profiling, and your code seems to be between 30% and 40% slower.

I can show you what I have profiled offline.

dmarcoux (approved):

PR #10194, opened 2020-09-23
Design of individual notifications

dmarcoux:
Before:
before

After:
after

### DIFF
--- a/src/api/app/helpers/webui/request_helper.rb
+++ b/src/api/app/helpers/webui/request_helper.rb
@@ -115,54 +115,72 @@ module Webui::RequestHelper
     "#{diff['project']} / #{diff['package']} (rev #{diff['rev']})"
   end
 
-  # rubocop:disable Style/FormatString
+  # rubocop:disable Style/FormatString, Metrics/PerceivedComplexity
   def request_action_header(action, creator)
-    source_project_hash = { project: action[:sprj], package: action[:spkg], trim_to: nil }
+    # FIXME: Remove assignments once we don't call anymore this helper with keys on the right side of the assignment
+    #        (those keys come from bs_request.webui_actions method)
+    action[:source_project] ||= action[:sprj]
+    action[:source_package] ||= action[:spkg]
+    action[:target_project] ||= action[:tprj]
+    action[:target_package] ||= action[:tpkg]
+    action[:target_repository] ||= action[:trepo]
+    action[:person_name] ||= action[:user]
+    action[:group_name] ||= action[:group]
 
-    case action[:type]
+    source_project_hash = { project: action[:source_project], package: action[:source_package], trim_to: nil }
+    target_project_hash = { project: action[:target_project], package: action[:target_package] }
+
+    case action[:type].to_sym
     when :submit
       'Submit %{source_container} to %{target_container}' % {
         source_container: project_or_package_link(source_project_hash),
-        target_container: project_or_package_link(project: action[:tprj], package: action[:tpkg])
+        target_container: project_or_package_link(target_project_hash)
       }
     when :delete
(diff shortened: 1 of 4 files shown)


dmarcoux:
[Design Notifications](https://trello.com/c/kmoGbSHk/1534-design-notifications)

src/api/app/presenters/notification_presenter.rb
```diff
@@ -40,13 +40,18 @@ def excerpt
...
+  def request_actions
     request = @model.notifiable
-    return "Multiple actions for project #{request.bs_request_actions.first.target_project}" if request.bs_request_actions.size > 1
+    request_action = request.bs_request_actions.first
+    return "Multiple actions for project #{request_action.target_project}" if request.bs_request_actions.size > 1
 
-    BsRequest.actions_summary(@model.event_payload)
+    # This is to avoid relying on bs_request.webui_actions, which is a method we need to remove
+    request_action_hash = request_action.serializable_hash.merge(type: request_action.type).with_indifferent_access
+    Webui::WebuiController.helpers.request_action_header(request_action_hash, request.creator)
```
dmarcoux, line 54: It's quite inconsistent how we display requests and their actions throughout the application. For example, in the project#pulse page, we have an icon for each request state. We don't do that anywhere else. `BsRequest.actions_summary` is also used to describe actions, but only in events (so for `app/models/event/`). In the end, I went with the helper method `request_action_header` since it's the one we use on request#show.

dmarcoux:
I feel like we have an information overload now in the notifications. The `Filter` box is eating up a lot of useful space.

danidoni:
> I feel like we have an information overload now in the notifications. The `Filter` box is eating up a lot of useful space. It's out of scope for this PR, but I believe we should rework the layout of the notifications page.

To me, it feels a bit weird, close to the Places menu on the left.

Maybe we can move those filters as dropdowns at the top, like the Status, Architectures and Repositores filters in the Monitor section of the Project.

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -117,48 +117,66 @@ def diff_label(diff)
...
       }
     when :delete
-      target_repository = "repository #{link_to(action[:trepo], repositories_path(project: action[:tprj], repository: action[:trepo]))} for " if action[:trepo]
+      target_repository = if action[:target_repository]
+                            link_to_repository = link_to(action[:target_repository],
+                                                         repositories_path(project: action[:target_project],
+                                                                           repository: action[:target_repository]))
+                            "repository #{link_to_repository} for "
+                          else
+                            ""
```
danidoni, line 146: This is triggering a rubocop offense.

```suggestion
                            ''
```

danidoni (changes requested):

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -115,54 +115,72 @@ def diff_label(diff)
     "#{diff['project']} / #{diff['package']} (rev #{diff['rev']})"
   end
 
-  # rubocop:disable Style/FormatString
+  # rubocop:disable Style/FormatString, Metrics/PerceivedComplexity
```
dmarcoux, line 118: `Metrics/PerceivedComplexity` is addressed once the `FIXME` I added is dealt with.

krauselukas (commented):
I like the changes, I agree with you, the places menu in combination with the filter box is not ideal on certain screen sizes

hennevogel:
Here is why this feels like information overload

notification-design

Green: Where does this happen?
Red: What is happening there?
Blue: State

saraycp:
Superseded by: #10216. Where a redesign is proposed.

The refactoring proposed in [this commit](https://github.com/openSUSE/open-build-service/pull/10194/commits/4313ad33bdbc6de0ea22c777f952aae0788fec6c) is really interesting. We can create a new PR just for that.

PR #10195, opened 2020-09-23
[backend] introduce publish flag to publish tracking report files

adrianschroeter:
RepoType: withreports   will also publish OBS internal .report files.

We will also keep publishing .packages files for now with that flag,
but we plan to remove that part when testing infrastructure has been adapted.

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -2153,12 +2153,15 @@ sub publish {
           $kiwimedium = "$arch/$1" if !$2 && -e "$r/$1.packages";
 	  $p = mapsleimage($sleimagedata, "$reporoot/$prp/$arch/:repo", $rbin, $p) if $sleimagedata;
         } elsif ($bin =~ /\.packages$/) {
+          # FIXME2.11: to be removed
+	  next unless $config->{'publishflags:withreports'};
           $p = $bin;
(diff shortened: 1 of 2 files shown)


mlschroe:
Why is this not a publish flag?

adrianschroeter:
right, it should be.

mlschroe:
What is `$config->{'withreports'}`? Why do you think that this can be set? Do you mean `$config->{'publishflags:withreports'}`?

src/backend/bs_publish
```diff
@@ -2153,12 +2153,15 @@ sub publish {
...
+	  next unless $config->{'withreports'};
           $p = $bin;
 	  $p = mapsleimage($sleimagedata, "$reporoot/$prp/$arch/:repo", $rbin, $p) if $sleimagedata;
         } elsif ($bin =~ /^(.*)\.report$/) {
 	  # collect reports
 	  $kiwireport{"$arch/$1"} = readxml("$r/$rbin", $BSXML::report, 1) if $do_packtrack || $do_sourcepublish;
-	  next;
+	  $p = $bin;
+	  next unless $config->{'withreports'};
 	} elsif ($bin =~ /^(.*)\.index$/) {
```
mlschroe, line 2165: Shouldn't you swap those two lines so that they look like the .packages case?

mlschroe (approved):

PR #10200, opened 2020-09-23
Replace `require_login` with Pundit in Cloud::Azure::ConfigurationController

### LINKED ISSUE #10083: Replace `require_login` controller filters with Pundit policies
Some controllers rely solely on Pundit and others on custom legacy code like [require_login](https://github.com/openSUSE/open-build-service/blob/92b16b72e57e483f7085441a2a39c5864c4ed28f/src/api/app/lib/authenticator.rb#L69-L71) to authorize various actions throughout the application.

We want to rely completely on Pundit to use a widely adopted authorization system instead of our own custom solution.

intrip:

Tackles [Cloud::Azure::ConfigurationController](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/cloud/azure/configurations_controller.rb).

Ref #10083


src/api/app/controllers/concerns/webui/rescue_handler.rb
```diff
@@ -14,7 +14,9 @@ module Webui::RescueHandler
                       when 'create_branch?' then 'create_branch'
                       else exception.try(:query)
                       end
-      message = if pundit_action && exception.record
+      message = if exception.reason
```
intrip, line 17: In order to keep the current flash message I propose to use the [latest version of Pundit exception](https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L24-L41) and pass a `reason`.

Sadly `Pundit` didn't publish a newer release yet so I had to patch `NotAuthorizedError ` in order to achieve this.

src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
```diff
@@ -2,8 +2,13 @@ module Webui
   module Cloud
     module Azure
       class ConfigurationsController < WebuiController
-        before_action :require_login
         before_action :set_azure_configuration
+        # TODO: Remove this when we'll refactor kerberos_auth
+        before_action :kerberos_auth
```
intrip, line 7: [As you pointed out](https://github.com/openSUSE/open-build-service/issues/10083#issuecomment-696264073) we need to deal with the `kerberos_mode`. 
I'm not completely sure yet how to deal with `kerberos_mode`, this is just an idea, what's you're take?

**Please note that I didn't test this part locally**

src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
```diff
@@ -27,7 +32,15 @@ def destroy
         private
 
         def set_azure_configuration
-          @azure_configuration = User.session!.azure_configuration || User.session!.create_azure_configuration
+          @azure_configuration = User.session ? find_or_create_azure_notification : ::Cloud::Azure::Configuration.new
```
intrip, line 35: This is a chicken-egg problem: we need `User` to fetch his `@azure_configuration` but we need the `@azure_configuration` in the `Policy`.

src/api/app/policies/application_policy.rb
```diff
@@ -1,7 +1,10 @@
 class ApplicationPolicy
   attr_reader :user, :record
 
+  PLEASE_LOGIN = 'Please login to access the requested page.'.freeze
+
   def initialize(user, record, opts = {})
+    ensure_logged_in!(user, opts)
```
intrip, line 7: The idea is to extend the Base class to deal with anonymous user in a backward compatible way.
Later on we could remove the check below (after making sure that the API works well) and further refactor the code in order to default `:ensure_logged_in ` to `true`.

src/api/spec/factories/azure_configuration.rb
```diff
@@ -3,5 +3,11 @@
     user { create(:user) }
     application_id { nil }
     application_key { nil }
+
+    trait :skip_encrypt_credentials do
+      after(:build) do |config|
+        config.define_singleton_method(:encrypt_credentials) {}
```
intrip, line 9: In order to avoid contacting the `backend` in the specs

src/api/spec/cassettes/Webui_Cloud_Azure_ConfigurationsController/DELETE_destroy/1_3_1.yml
```diff
@@ -1,124 +0,0 @@
----
```
intrip, line 1: I'm not totally sure that's correct, I've tried to follow the [Guidelines](https://github.com/openSUSE/open-build-service/wiki/Testing-with-VCR)

intrip:
@dmarcoux I've created a first PR, there's a few questions/notes which I've written down as comment and might require further discussions.
Could you take a peek and LMK your take? Thanks! 💚

intrip:
@dmarcoux I've just noticed that we have a failure in CI, this PR might require quite a few changes so I'd postpone checking this to after your review 😄

vpereira:
> @dmarcoux I've just noticed that we have a failure in CI, this PR might require quite a few changes so I'd postpone checking this to after your review  /cc @vpereira

I just rerun the failed tests and it is green again - sometimes it happens - https://app.circleci.com/pipelines/github/openSUSE/open-build-service/1597/workflows/2fd08176-a998-42e5-bd00-f3c046814ba3

intrip:
> > @dmarcoux I've just noticed that we have a failure in CI, this PR might require quite a few changes so I'd postpone checking this to after your review  /cc @vpereira
> 
> I just rerun the failed tests and it is green again - sometimes it happens - but it was unrelated with your changes. https://app.circleci.com/pipelines/github/openSUSE/open-build-service/1597/workflows/2fd08176-a998-42e5-bd00-f3c046814ba3

@vpereira thanks! Might be a flaky test 🤷‍♂️ .

danidoni (dismissed):

src/api/app/policies/application_policy.rb
vpereira: I really liked this approach, thank you!

src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
vpereira: looking into the method `kerberos_auth` it should work as "filter". So i guess its ok

vpereira:
@intrip, I have just one question:

looking the method `WebuiController#require_login` we have the whole "else" case which I don't see it implemented anywhere. specially the redirect destination if the proxy_auth_mod is enabled/disabled... any idea if it will be handled anywhere else?

intrip:
> @intrip, I have just one question:
> 
> looking the method `WebuiController#require_login` we have the whole "else" case which I don't see it implemented anywhere. specially the redirect destination if the proxy_auth_mod is enabled/disabled... any idea if it will be handled anywhere else?

@vpereira good catch 👍  !
I completely overlooked [the else part](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/webui_controller.rb#L61-L72) indeed therefore the redirects are no longer correct moreover the `xhr` answer is different 😞.

Maybe you have already an idea on how we could deal with that 🤔 ?

What about extending the  [rescue handler](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/concerns/webui/rescue_handler.rb#L5-L27) in order to deal with the redirects in case of Anonymous user:

```diff
# In ApplicationPolicy
-   PLEASE_LOGIN = 'Please login to access the requested page.'.freeze

  def ensure_logged_in!(user, opts)
-    raise Pundit::NotAuthorizedError, reason: PLEASE_LOGIN if opts[:ensure_logged_in] && (user.nil? || user.is_nobody?)
+    raise Pundit::NotAuthorizedError, reason: :anonymous_user if opts[:ensure_logged_in] && (user.nil? || user.is_nobody?)
  end

# In Webui::RescueHandler
      message = if exception.reason
-                exception.reason
+                # I'd put it into a separate method
+                case exception.reason
+                when :anonymous_user
+                  'Please login to access the requested page.'.freeze
+                else
+                  'Sorry, you are not authorized to perform this action.'
+                end

+    # I'd put it into a separate method
+    redirect_path = case exception.try(:reason)
+                               when :anonymous_user
+                               mode = CONFIG['proxy_auth_mode'] || :off
+                               if mode == :off
+                                 new_session_path
+                               else
+                                 root_path
+                               end
      if request.xhr?
        render json: { error: message }, status: 400
      else
        flash[:error] = message
-       redirect_back(fallback_location: root_path)
+       if redirect_path
+         redirect_to redirect_path
+       else
+         redirect_back(fallback_location: root_path)
+       end
      end
```

Related to [xhr part](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/webui_controller.rb#L61) I propose to migrate it to JSON as you can see from the snipped above, is it possible or is gonna break something?

Thanks!

vpereira:
> @vpereira good catch  !
> I completely overlooked [the else part](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/webui_controller.rb#L61-L71) therefore the redirects are no longer correct and the `xhr` answer is different .

Hey @intrip!

Maybe we should add some specs to see if the behavior still the same? Basically testing to where the user is being redirected when `proxy_auth_mode` is on or off, what do you think?

intrip:
> > @vpereira good catch  !
> > I completely overlooked [the else part](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/webui_controller.rb#L61-L71) therefore the redirects are no longer correct and the `xhr` answer is different .
> 
> Hey @intrip!
> 
> Maybe we should add some specs to see if the behavior still the same? Basically testing to where the user is being redirected when `proxy_auth_mode` is on or off, what do you think?

@vpereira 

Definitely agree in adding the tests, I'll add them.

You prefer to wait for the (probably) failing tests before discussing about/writing the implementation?

vpereira:
> @vpereira
> 
> Definitely agree in adding the tests, I'll add them.
> 
> You prefer to wait for the (probably) failing tests before discussing about/writing the implementation?

Yes, maybe they are not failing at all and case is closed :)

src/api/app/controllers/concerns/webui/rescue_handler.rb
```diff
@@ -3,27 +3,18 @@ module Webui::RescueHandler
...
-      message = if pundit_action && exception.record
-                  "Sorry, you are not authorized to #{pundit_action} this #{exception.record.class}."
-                else
-                  'Sorry, you are not authorized to perform this action.'
-                end
+      message = unauthorized_message(exception)
+      redirect_path = unauthorized_path(exception)
+
       if request.xhr?
         render json: { error: message }, status: 400
```
intrip, line 10: In case of anonymous `xhr` requests we return a `json` (instead of a `text`) response.
I see test passes so I'm supposing this is not a breaking change 😄 
Wdyt?

src/api/app/controllers/concerns/webui/rescue_handler.rb
```diff
@@ -52,5 +43,55 @@ class MissingParameterError < RuntimeError; end
       logger.debug "#{exception.class.name} #{exception.message} #{exception.backtrace.join('\n')}"
       render file: Rails.root.join('public/404'), status: 404, layout: false, formats: [:html]
     end
+
+    private
+
+    UNAUTHORIZED_MESSAGE = 'Sorry, you are not authorized to perform this action.'.freeze
+    def unauthorized_message(exception)
+      pundit_action = action_for_query(exception.query)
```
intrip, line 51: The `try` was removed: looks like it's not necessary since we are dealing with `Pundit::NotAuthorizedError` data type.

src/api/app/controllers/concerns/webui/rescue_handler.rb
```diff
@@ -52,5 +43,55 @@ class MissingParameterError < RuntimeError; end
...
+      if exception.reason
+        message_for_reason(exception.reason)
+      elsif pundit_action && exception.record
+        "Sorry, you are not authorized to #{pundit_action} this #{exception.record.class}."
+      else
+        UNAUTHORIZED_MESSAGE
+      end
+    end
+
+    def action_for_query(query) # rubocop:disable Metrics/CyclomaticComplexity
```
intrip, line 62: The `Metrics/CyclomaticComplexity` appeared after moving the code. 🤷‍♂️

intrip:
@vpereira 
I've added the specs for both redirect and `xhr` as discussed. I had to introduce a few code changes in order to make the redirect working as expected. Could you please take another look? Thanks!

src/api/app/controllers/concerns/webui/rescue_handler.rb
vpereira: i ignored the codeclimate issue. We can deal with that later.

dmarcoux:
Sorry @intrip, I was in holiday, I'm back now. I will have a look at the PR today :slightly_smiling_face:

src/api/config/initializers/patch_pundit_exception.rb
```diff
@@ -0,0 +1,28 @@
...
+#  - commit: https://github.com/varvet/pundit/commit/973b63b396c2a98099caf5eefd1c6841416eddfa
+#  - #master: https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L24-L41
+#
+# TODO: Remove this after a new Pundit release.
+module Pundit
+  class NotAuthorizedError < Error
+    attr_reader :query, :record, :policy, :reason
+
+    def initialize(options = {})
+      if options.is_a? String # rubocop:disable Style/MethodCallWithArgsParentheses
```
vpereira, line 14: hey @intrip could you please drop this `rubocop:disable` and use the `option.is_a?(String)` ?

vpereira (approved):
Hi @intrip well done! 

There is just a `rubocop:disable` that should be removed since we always add the parenthesis for method call (sometimes it looks non-ruby, we have some exceptions for that, but its a convention that we try to follow)

thanks!

intrip:
@vpereira thanks for the thorough review! I've just pushed the `rubocop` changes.

intrip:
> Sorry @intrip for the delay, I was in holiday, I'm back now. I will have a look at the PR today 🙂

@dmarcoux no worries at all, I hope holidays went well 😄

src/api/app/policies/application_policy.rb
dmarcoux: I wonder if this is really needed since in the following line, we're already handling anonymous users or not, if we pass the option `user_optional`. It's not checking for `user.is_nobody?` though. This could be changed in the following line.

Was `ensure_logged_in!` introduced to prevent potential errors with the `reason` option for `Pundit::NotAuthorizedError`? Is this what you meant with backward compatible?

dmarcoux (changes requested):
Great work, I don't have any major change to request. I want to clear out some things before approving.

src/api/config/initializers/patch_pundit_exception.rb
```diff
@@ -0,0 +1,28 @@
+# This monkey patch allows custom messages for a Pundit error.
+#
+# This is an exact copy of Pundit #master.
+# Refs:
+#  - commit: https://github.com/varvet/pundit/commit/973b63b396c2a98099caf5eefd1c6841416eddfa
+#  - #master: https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L24-L41
```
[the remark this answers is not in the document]
intrip: Very good point indeed!
I've update the comment.

src/api/app/controllers/webui/cloud/azure/configurations_controller.rb
```diff
@@ -2,8 +2,13 @@ module Webui
   module Cloud
     module Azure
       class ConfigurationsController < WebuiController
-        before_action :require_login
         before_action :set_azure_configuration
+        # TODO: Remove this when we'll refactor kerberos_auth
+        before_action :kerberos_auth
+        before_action :authorize_azure_configuration
+        # TODO: Remove this when the migration of require_login is completed
```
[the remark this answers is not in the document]
intrip: Apologies for the misleading comment: what I meant is that we'll eventually move the `after_action` to the `Webui::WebuiController` (as you wrote above).
I've removed the misleading comment, we can deal with that operation later on and I think you agree that a comment is not really required in order to explain that intent.

src/api/app/policies/application_policy.rb
intrip: Changing the line below and relying on `:user_optional` flag has an impact on all the already existing Policies which relies on it (pretty much in the API), since I wanted to reduce the scope of the code to check in this PR I've opted for introducing a new flag.
The idea is that eventually we'll migrate all the codebase to the new flag, but we can do this more gradually.

Moreover we introduced the `reason` attribute and we should (probably) also extend the API `rescue_handler` to deal with it.

intrip:
@dmarcoux thank you for the quick review!
I've answered all your qs, there is a comment which I don't see from the conversation for some reason, I'll link it here just incase you don't see it as well https://github.com/openSUSE/open-build-service/pull/10200/files#r496756830 🤷‍♂️

src/api/app/policies/application_policy.rb
dmarcoux: The `reason` attribute on `Pundit::NotAuthorizedError` is a purely additive one. Everything will keep behaving like it did before, even if the `reason` attribute is set. To stay backward compatible, we can set the `reason` attribute while still setting an exception message. It could be done like this, on the line 8 of this file (so directly below this comment):
```
    raise Pundit::NotAuthorizedError(reason: ANONYMOUS_USER), 'must be logged in' unless user || opts[:user_optional]
```

It's not needed though since nothing relies on the exception's message, so I would just go for this:
```
    raise Pundit::NotAuthorizedError(reason: ANONYMOUS_USER) unless user || opts[:user_optional]
```

`RescueHandler` (for the API) will simply ignore the `reason` attribute on `Pundit::NotAuthorizedError`. It's relying on the `record` and `query` (if present). See below: https://github.com/openSUSE/open-build-service/blob/bbdaa4907875ea3fb76c3813b15e2e72a6ddb945/src/api/app/controllers/concerns/rescue_handler.rb#L46-L68

`Webui::RescueHandler` will handle the `reason` attribute whenever it's provided with the changes introduced by this PR in `Webui::RescueHandler#unauthorized_message`. This is the right approach.

Beside those rescue handlers, there are two controllers which also rescue from `Pundit::NotAuthorizedError` and they are both not relying on the exception message. Why are they not using the rescue handlers? This is a question for another time...

So to conclude, this is why I would be against introducing another method to handle anonymous users.

intrip: @dmarcoux I agree with you that `reason` is not an issue 👍 .

My main concern was that if we change the `:user_optional` check in order to deal with anonymous user the existing logic changes for all the `ApplicationPolicy` subclasses, in particular all the Policies which have `:user_optional` not set (which is the default) will start raising a `NotAuthorizedError` when `user.is_nobody?` and that's what I meant with *backward compatibility*.

Since I was there I took a look and tried to figure out if the above was an issue therefore I've checked all the actions which do both `skip_before_action :require_login` and call `authorize` in order to figure out if any of them wouldn't work with the new change: looks like all of them will work well with this change therefore I've just pushed your proposal.
I hope I was clear this time and sorry for omitting some details.

*Side note*: some specs are failing now but It's related to the fact that the anonymous check is in the `initializer`, if you approve the changes I'll fix them 😄

intrip: @dmarcoux I've edited the message above.

dmarcoux: We have two approaches. The first one, we go with the backward compatible way and later deal with it. This limits the scope of this PR. The second way is dealing with all policies requiring changing due to the behavior of the `ApplicationPolicy`. This increases the scope of this PR a lot. I underestimated the increase of this approach.

Let's go with your first proposal. Sorry for the back and forth, I thought it wouldn't be so problematic to go with the second approach. I was wrong.

intrip: @dmarcoux No worries at all! I also didn't like the idea of adding another flag but reducing the scope of the PR is a good trade-off.
Was good that you pointed out the issue so that we had the opportunity to dig deeper into that option as well 👍 .

I've just rolled back the last changes.

dmarcoux (approved):
Great work and collaboration @intrip!

PR #10206, opened 2020-09-25
[backend] support "nofailedpackages" publishing flag

mlschroe:
This makes publishing fail if a packge build has not succeeded.


adrianschroeter (approved):

srinidhibs:
Thank you @mlschroe for this feature! This should help optimize our Pipelines where we run long running `while` loops to check if all packages built successfully or not.

PR #10207, opened 2020-09-25
Only use GET for search index endpoint

### LINKED ISSUE #10196: Search on the 404 page doesn't work
Issue Description
=============
The search button on the 404 page doesn't react on clicks and after a hit on the `return`-key the search forwards to a 403 error page.

Expected Result
============
The search button and search should behave normally.

How to Reproduce
===============
1. Go to not existing URL. For example https://build.opensuse.org/notexistingsite
2. Type something in the search box
3. Click on the 'Search' buttons.
4. See that nothing happens
5. Hit the `return` key.
6. See the 403 error

Further Information
===============
Here is a screencapture:
obs-404-search

krauselukas:
The search#index endpoint doesnt modify or
store and any data. Therefore we shouldn't allow
or use POST request to access the resource.
This will also prevent issues with our 404 page
which is a static html file that doesn't include a
CSRF token (which would be require for POST).

Fixes #10196

### DIFF
--- a/src/api/app/views/layouts/webui/_search.html.haml
+++ b/src/api/app/views/layouts/webui/_search.html.haml
@@ -1,2 +1,2 @@
-= form_tag(search_path(project: 1, package: 1, name: 1), class: 'form-inline') do
+= form_tag(search_path(project: 1, package: 1, name: 1), method: :get, class: 'form-inline') do
   %input.form-control.w-100{ 'aria-label': 'Search', name: 'search_text', placeholder: 'Search', type: 'text' }/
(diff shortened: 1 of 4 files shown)


src/api/app/controllers/webui/search_controller.rb
```diff
@@ -2,6 +2,7 @@ class Webui::SearchController < Webui::WebuiController
   before_action :set_attribute_list
   before_action :set_tracker_list
   before_action :set_parameters, except: :issue
+  skip_before_action :verify_authenticity_token, only: [:index]
```
hennevogel, line 5: Why is this a `POST` route at all? We are not changing/storing any data?

hennevogel (changes requested):

src/api/app/controllers/webui/search_controller.rb
krauselukas: :facepalm: true

krauselukas: its allows both POST and GET, I will check

krauselukas: @hennevogel I changed all the POST' to GET', no idea why we used POST', couldn't find a reason for it

hennevogel (approved):

hennevogel:
Hmmm.

PR #10214, opened 2020-09-28
Do not pack development and test gems

### LINKED ISSUE #10222: Pack into obs-server only the gems needed for production
We need to pack into the obs-server package only the gems needed for production.

Now we're packaging the gems for production, development and test environments. Removing them from the packaging will yield a smaller package.

danidoni:
This change makes the [obs-server](https://build.opensuse.org/package/show/OBS:Server:Unstable/obs-server) package to pack only the gems needed for production, we skip the `:development` and `:test` gem groups.

Fixes #10222

### DIFF
--- a/dist/obs-api-testsuite-rspec.spec
+++ b/dist/obs-api-testsuite-rspec.spec
@@ -52,7 +52,7 @@ popd
 
 pushd src/api
 # configure to the bundled gems
-bundle --local --path %_libdir/obs-api/ --without development test
+bundle --local --path %_libdir/obs-api/ --without development
 
 ./script/prepare_spec_tests.sh
 
--- a/dist/obs-api-testsuite-rspec.spec
+++ b/dist/obs-api-testsuite-rspec.spec
@@ -52,7 +52,7 @@ popd
 
 pushd src/api
 # configure to the bundled gems
-bundle --local --path %_libdir/obs-api/ --without development
+bundle --local --path %_libdir/obs-api/
 
 ./script/prepare_spec_tests.sh
 
--- a/dist/obs-api-testsuite-rspec.spec
+++ b/dist/obs-api-testsuite-rspec.spec
@@ -52,7 +52,7 @@ popd
 
 pushd src/api
 # configure to the bundled gems
-bundle --local --path %_libdir/obs-api/
+bundle --local --path %_libdir/obs-api/ --without development test
 
 ./script/prepare_spec_tests.sh
 
--- a/dist/obs-bundled-gems.spec
+++ b/dist/obs-bundled-gems.spec
@@ -127,7 +127,7 @@ bundle config build.ffi --enable-system-libffi
 bundle config build.nokogiri --use-system-libraries
 bundle config build.sassc --disable-march-tune-native
 
-bundle --local --path %{buildroot}%_libdir/obs-api/
+bundle --local --path %{buildroot}%_libdir/obs-api/ --without development test
 
 # test that the rake and rack macros is still matching our Gemfile
 test -f %{buildroot}%_libdir/obs-api/ruby/%{__obs_ruby_version}/gems/rake-%{rake_version}/rake.gemspec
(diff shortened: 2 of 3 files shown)


vpereira:
@danidoni is this failure expected https://build.opensuse.org/package/live_build_log/OBS:Server:Unstable:TestGithub:PR-10214/obs-server/SLE_15_SP2/x86_64 ?

danidoni:
> @danidoni is this failure expected https://build.opensuse.org/package/live_build_log/OBS:Server:Unstable:TestGithub:PR-10214/obs-server/SLE_15_SP2/x86_64 ?

No, I'm currently looking at what's happening there, but ATM I have no idea.

dist/obs-api-testsuite-rspec.spec
```diff
@@ -52,7 +52,7 @@ popd
 
 pushd src/api
 # configure to the bundled gems
-bundle --local --path %_libdir/obs-api/
+bundle --local --path %_libdir/obs-api/ --without development test
```
eduardoj, line 55:
```suggestion
bundle --local --path %_libdir/obs-api/ --without development
```

I have not checked it, but most probably, to run the test suite we still need to build the gems of the environment test.

eduardoj:
> > @danidoni is this failure expected https://build.opensuse.org/package/live_build_log/OBS:Server:Unstable:TestGithub:PR-10214/obs-server/SLE_15_SP2/x86_64 ?
> 
> No, I'm currently looking at what's happening there, but ATM I have no idea.

Having a look at the log, it seems that we run the minitest tests before building the package, with `sh -x ./script/api_minitest.sh`. Most likely we should still need the test environment for building the packages.

dist/obs-api-testsuite-rspec.spec
danidoni: Aaah, good catch... You're right

danidoni:
> > > @danidoni is this failure expected https://build.opensuse.org/package/live_build_log/OBS:Server:Unstable:TestGithub:PR-10214/obs-server/SLE_15_SP2/x86_64 ?
> > 
> > 
> > No, I'm currently looking at what's happening there, but ATM I have no idea.
> 
> Having a look at the log, it seems that we run the minitest tests before building the package, with `sh -x ./script/api_minitest.sh`. Most likely we should still need the test environment for building the packages.

Yeah, you are totally right. I've accepted your suggestion. Let's see how it works now.

coolo:
Whatever you bundle in obs-server.spec doesn't matter - the gems come from obs-bundled-gems. And if you need the test gems in there to run the test suite while building, you will need to create subpackages in obs-bundled-gems so you can cherry pick which to build against and which to run against.

Which of course opens a big door of mess if you never test what you run. E.g. you will never know if your production env will run without a given gem - I mean before deployment.

danidoni:
> Whatever you bundle in obs-server.spec doesn't matter - the gems come from obs-bundled-gems. And if you need the test gems in there to run the test suite while building, you will need to create subpackages in obs-bundled-gems so you can cherry pick which to build against and which to run against.

Yes, I understand. As we have several specfiles, I'll change only the one in `obs-bundled-gems` to remove the unneeded gems from production, let the `dist/obs-api-testsuite-rspec.spec` and the `dist/obs-server.spec` as it was before.

hennevogel:
What @coolo is asking:

Where do you test that booting the application with

`bundle... --without development test`

still works?

danidoni:
> What @coolo is asking:
> 
> Where do you test that booting the application with
> 
> `bundle... --without development test`
> 
> still works?

Ah... Nowhere... I see the problem :thinking:

coolo:
It's not the only problem though - but I'll let the "OBS Package Build" tell you the details once it's synced 😄

danidoni:
> It's not the only problem though - but I'll let the "OBS Package Build" tell you the details once it's synced smile

Right now we build three packages, obs-server, obs-bundled-gems and obs-api-testsuite-rspec. As we share the Gemfile, it is impossible to use different gems in each of the three builds.

So this PR is currently impossible to finish unless we do what @coolo says and split the package in three.

I'm closing it.

PR #10216, opened 2020-09-28
Notifications accurate redesign

saraycp:
In this new round of redesign, we've tried to avoid having the notifications' information spread all over the area. So we've put together the information which answers to these two questions respectively: Where does the notification happen? What is happening there? The goal is making it easier to understand at a glance.

**Improvements:**

- We ensure we display relevant information like source and target project/package when needed.
- Change the position of the elements to group them logically.
- Specify the kind of requests in the link's text. Example: _"Add Role Request # 222"_
- The link's text for comments is changed from project/package name or request Id to something more clear like: _"Comment on Package"_ or _"Comment on Add Role Request"_. Needed to distinguish where the comment happened, especially in comments on requests since the action summary is not as verbose as before.
- Less variety of font styles

**Before (desktop version)**

desktop_before

**After (desktop version)**

desktop_after

**Before (mobile version)**

mobile_before

**After (mobile version)**

mobile_after

### DIFF
--- a/src/api/app/assets/stylesheets/webui/notifications_redesign/notifications.scss
+++ b/src/api/app/assets/stylesheets/webui/notifications_redesign/notifications.scss
@@ -2,24 +2,6 @@
 // all the 'body.notifications-redesign' wrappers can be removed.
 
 body.notifications-redesign  {
-  .notifications-grid-container {
-    display: grid;
-    grid-template-columns: 1fr auto;
-    grid-template-rows: auto;
-    grid-template-areas:
-      "notifiable notifiable"
-      "content content"
-      "who-when action";
-    column-gap: 0.5rem;
-    row-gap: 0.25rem;
-    align-items: start;
-
-    .notifiable { grid-area: notifiable; }
-    .who-when { grid-area: who-when; }
-    .action { grid-area: action; }
-    .content { grid-area: content; }
-  }
-
   .avatars-counter, .simulated-avatar {
     display: inline-block;
     text-align: center;
@@ -50,14 +32,6 @@ body.notifications-redesign  {
   body.notifications-redesign {
     #notifications-filter-desktop {
       .collapse { display: block !important; }
-    }
-
-    .notifications-grid-container {
-      min-height: 6.25rem;
-      grid-template-columns: 1fr auto 3rem;
-      grid-template-areas:
-        "notifiable who-when action"
-        "content content content";
     }
   }
 }
--- a/src/api/app/helpers/webui/notification_helper.rb
+++ b/src/api/app/helpers/webui/notification_helper.rb
@@ -28,6 +28,28 @@ module Webui::NotificationHelper
     end
   end
 
+  def action_description(notification)
+    case notification.event_type
+    when 'Event::RequestStatechange', 'Event::RequestCreate', 'Event::ReviewWanted', 'Event::CommentForRequest'
+      source_and_target(notification)
+    when 'Event::CommentForProject'
+      "#{notification.notifiable.commentable.name}"
+    when 'Event::CommentForPackage'
+      commentable = notification.notifiable.commentable
+      "#{commentable.project.name} / #{commentable.name}"
+    end
+  end
+
+  def source_and_target(notification)
+    capture do
+      if notification.source.present?
+        concat(tag.span(notification.source))
+        concat(tag.i(nil, class: 'fas fa-long-arrow-alt-right text-info mx-2'))
+      end
+      concat(tag.span(notification.target))
+    end
+  end
+
   private
 
   def css_for_filter_link(filter_item, selected_filter)
--- a/src/api/app/presenters/notification_presenter.rb
+++ b/src/api/app/presenters/notification_presenter.rb
@@ -12,17 +12,16 @@ class NotificationPresenter < SimpleDelegator
   def notifiable_link
     case @model.event_type
     when 'Event::RequestStatechange', 'Event::RequestCreate', 'Event::ReviewWanted'
-      { text: "Request ##{@model.notifiable.number}",
+      { text: "#{type_of_action(@model.notifiable)} Request ##{@model.notifiable.number}",
         path: Rails.application.routes.url_helpers.request_show_path(@model.notifiable.number, notification_id: @model.id) }
     when 'Event::CommentForRequest'
-      { text: "Request ##{@model.notifiable.commentable.number}",
+      { text: "Comment on #{type_of_action(@model.notifiable.commentable)} Request",
         path: Rails.application.routes.url_helpers.request_show_path(@model.notifiable.commentable.number, notification_id: @model.id, anchor: 'comments-list') }
     when 'Event::CommentForProject'
-      { text: @model.notifiable.commentable.name,
+      { text: 'Comment on Project',
         path: Rails.application.routes.url_helpers.project_show_path(@model.notifiable.commentable, notification_id: @model.id, anchor: 'comments-list') }
     when 'Event::CommentForPackage'
-      commentable = @model.notifiable.commentable
-      { text: "#{commentable.project.name} / #{commentable.name}",
+      { text: 'Comment on Package',
         path: Rails.application.routes.url_helpers.package_show_path(package: @model.notifiable.commentable,
                                                                      project: @model.notifiable.commentable.project,
                                                                      notification_id: @model.id,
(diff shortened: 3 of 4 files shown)


src/api/app/presenters/notification_presenter.rb
```diff
@@ -73,11 +63,44 @@ def avatar_objects
...
+  end
+
+  def request_summary(bs_request)
+    return "#{bs_request.bs_request_actions.first.target_project}" if bs_request.bs_request_actions.size > 1
+
+    bs_request_action = bs_request.bs_request_actions.first
+
+    case bs_request_action.type
+    when 'submit', 'maintenance_incident', 'maintenance_release', 'release change_devel'
+      "#{bs_request_action.source_project} - #{bs_request_action.target_project}"
```
hennevogel, line 101: hm, I think we misunderstood each other. this needs to show which package.

saraycp: True! Code updated. Please have another look, @hennevogel

src/api/app/assets/stylesheets/webui/notifications_redesign/notifications.scss
```diff
@@ -7,19 +7,29 @@ body.notifications-redesign  {
...
-    align-items: start;
 
     .notifiable { grid-area: notifiable; }
-    .who-when { grid-area: who-when; }
-    .action { grid-area: action; }
+    .actions { grid-area: actions; }
     .content { grid-area: content; }
   }
 
+  .notifications-who-where-grid-container {
```
hennevogel, line 20: Is there a reason you don't use the bootstrap grid? 

Why don't you stack the avatars and the test on mobile?

saraycp: @hennevogel, I was able to get rid of the custom grids.
The avatar's area and the text area are displayed in two different rows when their content is too large.

eduardoj:
There were conflicts after another pull request was merged. I rebased and solved the conflicts.

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
```diff
@@ -17,29 +17,31 @@
...
-              = render partial: 'notification_avatars', locals: { avatar_objects: notification.avatar_objects }
-              %span.d-inline-block.text-nowrap.ml-2 #{time_ago_in_words(notification.created_at)} ago
-
+          .list-group-item.px-0.px-md-2.container
+            .row
+              .col
+                - if notification.notifiable_type == 'BsRequest'
+                  = image_tag('icons/request-icon.svg', height: 18, title: 'Request notification')
+                  = link_to(notification.notifiable_link[:text], notification.notifiable_link[:path], class: 'mx-1 text-word-break-all')
+                  %small.text-nowrap #{time_ago_in_words(notification.created_at)} ago
```
krauselukas, line 26: The time is now only shown on bs_request notifications, not on the comments, is this intended? @saraycp

saraycp: Oops! I overlooked that. I've just added it for comments as well. I'm going to update the screenshots.

dmarcoux (commented):
The more I look at the notifications view, even before the changes from this PR, the more I think that the notification excerpt overloads the view with information. The excerpt doesn't bring much value, especially on mobile. I would be in favor of removing the excerpt entirely. What do you think?

danidoni:
> The more I look at the notifications view, even before the changes from this PR, the more I think that the notification excerpt overloads the view with information. The excerpt doesn't bring much value, especially on mobile. I would be in favor of removing the excerpt entirely. What do you think?

I'm in

src/api/app/presenters/notification_presenter.rb
```diff
@@ -73,11 +63,46 @@ def avatar_objects
...
+
   private
 
   def render_without_markdown(content)
     # Initializes a Markdown parser, if needed
     @remove_markdown_parser ||= Redcarpet::Markdown.new(Redcarpet::Render::StripDown)
     ActionController::Base.helpers.sanitize(@remove_markdown_parser.render(content.to_s))
   end
+
+  def kind_of_request(bs_request)
```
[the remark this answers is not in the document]
saraycp: What we call action summary in other parts of the code returns something like "Submit packageA to packageB".

Now we need to split that text into two parts:
- "Submit" what I called _kind_of_request_.
- "packageA to packageB" what I called _request_summary_.


What about using  _type_of_action_ instead _kind_of_request_?
:thinking: Now I need to find a good name to replace _request_summary_.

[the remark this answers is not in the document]
dmarcoux: I like _type_of_action_!

saraycp:
> > The more I look at the notifications view, even before the changes from this PR, the more I think that the notification excerpt overloads the view with information. The excerpt doesn't bring much value, especially on mobile. I would be in favor of removing the excerpt entirely. What do you think?
> 
> I'm in

There was a discussion about this in the past. It seems some users rely on the expert to really understand what they have to do or what's going on.

We decided to drop the excerpt in the mobile version for design reasons, however, removing it from the desktop views was considered too much.

eduardoj:
> The more I look at the notifications view, even before the changes from this PR, the more I think that the notification excerpt overloads the view with information. The excerpt doesn't bring much value, especially on mobile. I would be in favor of removing the excerpt entirely. What do you think?

Now that showing excerpts is implemented, we can easily remove them now, and also add them again later, if necessary. I don't have a strong opinion.

I suppose it will depend on the user case:
- there will be users that don't need to read an excerpt, because the want to see everything what has happened in a project or a package the maintain, for example; and
- there will be users that need need to read an excerpt, because they can not afford to go through all the notifications, and the rest of the information that is displayed not contained in the excerpt, is not enough.

I don't see an issue now, on keeping the excerpts. So I would leave them as they are, by now.

DavidKang:
@saraycp, Nice work. Just a few comments. 

#### Desktop:
Screenshot_2020-09-29 Open Build Service(1)

IMO, the `excerpt` is to near of the avatar and package name

#### Mobile
Screenshot_2020-09-29 Open Build Service

In general, I feel the content is too compact and needs a little bit of space.

Apart from that, nice work :smile_cat:

saraycp:
> IMO, the `excerpt` is to near of the avatar and package name

> In general, I feel the content is too compact and needs a little bit of space.

Hey! Thanks for your comments, @DavidKang. Exactly that, playing with the spaces, is what I'm doing, instead of using different font styles or dropping the excerpt.

saraycp:
PR and screenshots updated. I've added a new commit with the adjustments to include more space between the elements and make the info look less crowded. The commit could be squashed as soon it is reviewed.

vpereira (approved):
Great work @saraycp probably some users want to see the excerpt and others not. Maybe in a second moment have a way to show/hide them would be needed.

dmarcoux:
Between 700px and 760px, the left/right margins are a bit too much. It looks a bit weird:

margin-notifications

It's not a blocker, I've just noticed this.

src/api/app/helpers/webui/notification_helper.rb
```diff
@@ -28,6 +28,28 @@ def request_badge_color(state)
...
+      "#{notification.notifiable.commentable.name}"
+    when 'Event::CommentForPackage'
+      commentable = notification.notifiable.commentable
+      "#{commentable.project.name} / #{commentable.name}"
+    end
+  end
+
+  def source_and_target(notification)
+    capture do
+      if notification.source.present?
```
krauselukas, line 45: Hmm, this only work with a NotificationPresenter object right? Why did you moved all the logic to the helpers?

saraycp: Because I needed to add the arrow icon in the middle. But I'm working on dropping `source` and `target` from the presenter and just have all this logic in the helper. But it is not a blocker, it's something I can in another PR.

dmarcoux:
The conflict is due to the merge of #10220: https://github.com/openSUSE/open-build-service/pull/10220/files#diff-1fc6031128e8767883b805cc36ea1d95

src/api/app/helpers/webui/notification_helper.rb
krauselukas: ok

saraycp:
There are some details that can be improved:

- [Excessive margins](https://github.com/openSUSE/open-build-service/pull/10216#issuecomment-701389508) noticed by @dmarcoux 
- Split source and target in two independent lines with a down arrow and use different sizes for project and package. As suggested by @danidoni 
Screenshot_2020-09-30 Open Build Service(9)

However, I would like to merge this PR how it is and do the improvements in a new one. This way we can rebase with #10189 as soon as possible.

dmarcoux (approved):

danidoni (approved):

Issue #9279, opened 2020-03-24
[beta] Need a way to mark a request notification done from the request


coolo:
If I have a request in notifications, I can visit it or mark it as done - but not both. So I always have to take 2 steps, work on the request and then mark the notification as done. That's quite frustrating.

coolo:
As reference: github's toolbar within PRs - when coming from notification:

![image](https://user-images.githubusercontent.com/1067203/77473643-0f1b3a80-6e16-11ea-822d-8e31b228f0d4.png)

PR #10220, opened 2020-09-28
Show Notification Toolbar when opening notification from Notifications List Page

OleksandrOrlov:
The PR adds a toolbar on the top of the page, which adds ability to mark notification as read from Request, Project and Package Pages.

The feature was requested in #9279

Steps to test the feature:
1. Log in as Admn that has notifications;
2. Proceed to the Notifications List Page;
3. Click on the title of any of the noifications;
4. When presented with the Request/Package/Project Page, focus on the top of the page.

Expected result:
Notification Toolbar is shown on the top, just below the Top Nav.

Here are the screenshots of how it looks:

**Desktop**

**Mobile**

### DIFF
--- a/src/api/app/assets/stylesheets/webui/notification_toolbar.scss
+++ b/src/api/app/assets/stylesheets/webui/notification_toolbar.scss
@@ -1,0 +1,9 @@
+.notification-toolbar.sticky-top {
+  z-index: $zindex-fixed;
+  top: $top-navigation-height;
+}
+
+.notification-toolbar ~ .container-xxl .flash-and-announcement {
+  top: $top-navigation-height + $notification-toolbar-height + .25rem !important;
+  z-index: calc(#{$zindex-fixed} - 5);
+}
--- a/src/api/app/queries/notifications_finder.rb
+++ b/src/api/app/queries/notifications_finder.rb
@@ -44,4 +44,10 @@ class NotificationsFinder
   def for_project_name(project_name)
     unread.joins(:projects).where(projects: { name: project_name })
   end
+
+  def for_subscribed_user_by_id(notification_id)
+    return unless User.session && notification_id
+
+    self.class.new.for_subscribed_user.find_by(id: notification_id)
+  end
 end
--- a/src/api/app/views/layouts/webui/_notification_toolbar.html.haml
+++ b/src/api/app/views/layouts/webui/_notification_toolbar.html.haml
@@ -1,0 +1,5 @@
+.notification-toolbar.sticky-top.navbar.navbar-light.bg-light.border-bottom.border-gray-400
+  = link_to(my_notifications_path, class: 'btn btn-sm btn-outline-secondary', title: 'Back to notifications') do
+    %i.fas.fa-arrow-left
+    Back to notifications
+  = mark_as_read_or_unread_button(notification)
--- a/src/api/app/assets/stylesheets/webui/application.scss
+++ b/src/api/app/assets/stylesheets/webui/application.scss
@@ -62,6 +62,7 @@
 
 @import 'responsive_ux/responsive_ux';
 @import 'notifications_redesign/notifications';
+@import 'notification_toolbar';
 
 html {
     overflow-y: scroll !important;
--- a/src/api/app/assets/stylesheets/webui/responsive_ux/responsive_ux.scss
+++ b/src/api/app/assets/stylesheets/webui/responsive_ux/responsive_ux.scss
@@ -3,6 +3,7 @@
 $top-navigation-height: 3.5625rem;
 $bottom-navigation-height: 3.5rem;
 $left-navigation-width: 16rem;
+$notification-toolbar-height: 3rem;
 
 body.responsive-ux {
   @import 'build-result';
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -104,6 +104,9 @@ class Webui::PackageController < Webui::WebuiController
 
     @comments = @package.comments.includes(:user)
     @comment = Comment.new
+
+    @current_notification = NotificationsFinder.new.for_subscribed_user_by_id(params[:notification_id])
+
     @services = @files.any? { |file| file[:name] == '_service' }
 
     respond_to do |format|
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -132,6 +132,9 @@ class Webui::ProjectController < Webui::WebuiController
     @has_patchinfo = @project.patchinfos.exists?
     @comments = @project.comments
     @comment = Comment.new
+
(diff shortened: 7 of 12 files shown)


src/api/app/assets/javascripts/webui/notification_toolbar.js
```diff
@@ -0,0 +1,7 @@
+$(window).scroll(function () {
+    if (this.scrollY > 100) {
+        $('.notification-controls').addClass('sticking');
```
krauselukas, line 3: @OleksandrOrlov is this a left over, I cannot find the `notification-controls` class in the html code

src/api/app/assets/stylesheets/webui/notification_toolbar.scss
```diff
@@ -0,0 +1,8 @@
+.notification-toolbar.sticky-top {
+  z-index: calc(#{$zindex-modal} - 10);
```
krauselukas, line 2: @OleksandrOrlov I think you need to check the z-index again, it overlaps with the watchlist on mobile 
Screenshot_2020-09-29 Request 8 (declined)

src/api/app/assets/javascripts/webui/notification_toolbar.js
OleksandrOrlov: You're right. This is not needed here anymore. Thanks for the finding. Sorry:)

src/api/app/views/layouts/webui/_notification_toolbar.html.haml
```diff
@@ -0,0 +1,11 @@
+- if notification
```
danidoni, line 1: I would move the condition [outside](https://github.com/openSUSE/open-build-service/pull/10220/files#diff-45250954406833ba3815d964f479ed8bR48). So if we don't have the notification we don't even call the partial rendering.

src/api/app/views/layouts/webui/_notification_toolbar.html.haml
```diff
@@ -0,0 +1,11 @@
...
+  .notification-toolbar.sticky-top.navbar.navbar-light.bg-light.border-bottom.border-gray-400
+    - title, icon = notification.unread? ? ['Mark as read', 'fa-check'] : ['Mark as unread', 'fa-undo']
```
danidoni, line 3: And maybe move this logic into a helper

danidoni (changes requested):

src/api/app/queries/notifications_finder.rb
```diff
@@ -44,4 +44,10 @@ def for_notifiable_type(type = 'unread')
   def for_project_name(project_name)
     unread.joins(:projects).where(projects: { name: project_name })
   end
+
+  def current_notification(notification_id)
+    return unless User.session && notification_id
+
+    self.class.new.for_subscribed_user.where(id: notification_id).first
```
dmarcoux, line 51: I suggest using `find_by(...)` instead of `where(...).first`. They differ a bit since `first` orders the records, then return the first one. As for `find_by`, it returns the first record without ordering the records. In our case, we don't need to order the records since there's only one possible notification with the ID we're looking for. We could use `find`, but it will raise an error if a notification cannot be found for that ID. It might happen if someone is trying to be funny and pass us some garbage.

The subtle differences between the methods are described [here](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html).

```suggestion
    self.class.new.for_subscribed_user.find_by(id: notification_id)
```

src/api/app/assets/stylesheets/webui/notification_toolbar.scss
OleksandrOrlov: Fixed.

src/api/app/queries/notifications_finder.rb
OleksandrOrlov: Changed, thanks!

src/api/app/views/layouts/webui/_notification_toolbar.html.haml
```diff
@@ -0,0 +1,11 @@
...
+    - update_path = my_notification_path(id: notification.id)
+    = link_to(my_notifications_path, class: 'btn btn-sm btn-outline-secondary', title: 'Back to notifications') do
+      %i.fas.fa-arrow-left
+      Back to notifications
+    = link_to(update_path, id: format('update-notification-%d', notification.id),
+                          method: :put, class: 'btn btn-sm btn-outline-success', title: title) do
+      %i.fas{ class: "#{icon}" }
+      #{title}
```
dmarcoux, line 11: Just a tiny nitpick...

```suggestion
      = title
```

src/api/app/queries/notifications_finder.rb
```diff
@@ -44,4 +44,10 @@ def for_notifiable_type(type = 'unread')
   def for_project_name(project_name)
     unread.joins(:projects).where(projects: { name: project_name })
   end
+
+  def current_notification(notification_id)
```
dmarcoux, line 48: `current_notification` is a bit too generic, making this method somewhat cryptic.

How about `for_subscribed_user_by_id`? It's longer, but much clearer. It would read like `NotificationsFinder.new.for_subscribed_user_by_id(12345)`.
```suggestion
  def for_subscribed_user_by_id(notification_id)
```

src/api/app/views/layouts/webui/_notification_toolbar.html.haml
OleksandrOrlov: Moved outside.

src/api/app/views/layouts/webui/_notification_toolbar.html.haml
OleksandrOrlov: Moved to helper. Not sure about the name for the function, I've chosen `title_and_icon_for_read_button`. If you have better name in mind, please let me know.

src/api/app/views/layouts/webui/_notification_toolbar.html.haml
OleksandrOrlov: Done.

src/api/app/queries/notifications_finder.rb
OleksandrOrlov: Makes sense, since it could be used not only for the case when it is required for the current notification. Your name is more general and reflects of what it does inside. Thank you.

src/api/app/views/layouts/webui/_notification_toolbar.html.haml
danidoni: You can move the `link_to` also and the maybe you can rename the helper to `link_for_notification` or something like this

OleksandrOrlov: Done.

dmarcoux (approved):

dmarcoux:
@danidoni Please have another look if there's still something to change

danidoni (approved):

danidoni:
> @danidoni Please have another look if there's still something to change

:ship: :it:

PR #10227, opened 2020-09-29
Update BS Request with AJAX call

OleksandrOrlov:
Update BS request with AJAX call
    
The commit introduces asynchronous update for BS Requests.
    
This was implemented in scope of adding notification toolbar feature (#10220).
Previously, when the request got updated (e.g. approved/rejected), the page was refreshed and notification toolbar was removed, because it relies on 'notification_id' query parameter, but that parameter got removed after action was made on BS request.

To test the feature:

1. Log in as user that has notifications;
2. Proceed to the Notifications List Page;
3. Open Request notification which is in "review" or "declined" state;
4. Scroll the Request Page to the bottom;
5. Press "Decline"/"Approve"/"Reopen" button.

Expected result:
The page is updated asynchronously.  (Overview, Decision, Request History sections are updated).

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -125,6 +125,11 @@ class Webui::RequestController < Webui::WebuiController
     user = User.session # might be nil
     @my_open_reviews = reviews.select { |review| review.matches_user?(user) }
     @can_add_reviews = @bs_request.state.in?([:new, :review]) && (@is_author || @is_target_maintainer || @my_open_reviews.present?)
+
+    respond_to do |format|
+      format.html
+      format.js { render_request_update }
+    end
   end
 
   def sourcediff
@@ -336,5 +341,20 @@ class Webui::RequestController < Webui::WebuiController
     opt['role'] = params[:role]
     opt['type'] = type.to_s
     opt
+  end
+
+  def render_request_update
+    render partial: 'update', locals: {
+      is_target_maintainer: @is_target_maintainer,
+      is_author: @is_author,
+      bs_request: @bs_request,
+      history: @history,
+      can_add_reviews: @can_add_reviews,
+      package_maintainers: @package_maintainers,
+      can_handle_request: @can_handle_request,
+      my_open_reviews: @my_open_reviews,
+      show_project_maintainer_hint: @show_project_maintainer_hint,
+      actions: @actions
+    }
   end
 end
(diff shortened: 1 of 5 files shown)


src/api/app/views/layouts/webui/responsive_ux/_unread_notifications.html.haml
```diff
@@ -0,0 +1,5 @@
+= link_to(my_notifications_path, class: 'nav-link text-light p-0 w-100', alt: 'Notifications') do
+  %i.fas.fa-bell
+  - unless User.session.unread_notifications.zero?
+    %span.badge.badge-primary.align-text-top= User.session.unread_notifications
+  %span.d-block Notifications
```
danidoni, line 5: Add a newline here, please

OleksandrOrlov: Yep, sorry. There was a PR merged recently and I didn't make rebase. Now the issue is gone. Thanks for the notice:)

danidoni (approved):

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -337,4 +342,20 @@ def request_action_attributes(type)
     opt['type'] = type.to_s
     opt
   end
+
+  def render_request_update
+    render partial: 'update', locals: {
```
DavidKang, line 347: Is there any benefit to have this method only for render partial?

OleksandrOrlov: If I will put it to `show` method, then linter will complain, that there are too many lines in the method. Could you please tell me if you have any other idea how can I solve that issue? Thank you.

DavidKang: I see now, `show` action is a little big :sweat_smile:. Maybe doing it in other way, moving all those instance variables to a method like for example: `load_dependencies`. But right now, I don't have a better solution for the issue, sorry :crying_cat_face:

dmarcoux (approved):

src/api/app/views/webui/request/_update.js.erb
```diff
@@ -0,0 +1,11 @@
+$('#decision').html("<%= j(render partial: 'decision_tab', locals: { request_number: request_number,
```
dmarcoux, line 1: Just one thing before merging, could we use `escape_javascript` instead of `j`?

dmarcoux (changes requested):
I've found an issue. When accepting a request, the decision stays even though it should be hidden. This isn't an issue when declining or reopening a request. This happens because we always render the `decision_tab` partial in the new `update.js.erb` partial.

decision-box

OleksandrOrlov:
@dmarcoux I've fixed the issue. Thank you for the attention! Could you please review when you'll have a time.

dmarcoux (approved):

PR #10242, opened 2020-10-01
Publish build events to the metrics bus

hennevogel:
There are also some interesting times and tags in the build events.
Let's measure them.


danidoni (dismissed):

hennevogel:
have some more changes anyway 😁

PR #10244, opened 2020-10-01
[api] change request state to badge

### LINKED ISSUE #10238: Display request state with a badge on request#show
The request state is displayed between parentheses, such as `Request 123 (some state)`. This is how it looks right now:

request-show-before

We sometimes display the state as described above, but in other cases, like in the notifications, we display the state of a request in a badge. It looks much nicer. I suggest doing the same

In order to be consistent throughout the application, we should always display the request state in a badge. I quickly copy-pasted this in the view to show how it would look like:
request-show-after

We should re-use what we use in the notifications, [request_badge_color](https://github.com/openSUSE/open-build-service/blob/8ca526bbfb3ddafe37f68e9ac518190ec7225956/src/api/app/helpers/webui/notification_helper.rb#L16-L29) is the helper we're using in the notifications.

hendrixfan:
Hi there,

this fixes #10238. It shows the state in request#show as a badge to make it more consistent looking.

### DIFF
--- a/src/api/spec/features/beta/webui/requests_spec.rb
+++ b/src/api/spec/features/beta/webui/requests_spec.rb
@@ -86,7 +86,8 @@ RSpec.describe 'Bootstrap_Requests', type: :feature, js: true, vcr: true do
         visit request_show_path(bs_request)
         click_button 'Accept'
 
-        expect(page).to have_text("Request #{bs_request.number} (accepted)")
+        expect(page).to have_text("Request #{bs_request.number}")
+        expect(find('span.badge.badge-success')).to have_text('accepted')
         expect(page).to have_text('In state accepted')
       end
     end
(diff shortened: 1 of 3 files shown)


src/api/spec/features/beta/webui/requests_spec.rb
```diff
@@ -86,7 +86,7 @@
         visit request_show_path(bs_request)
         click_button 'Accept'
 
-        expect(page).to have_text("Request #{bs_request.number} (accepted)")
```
danidoni, line 89: If you remove this `(accepted)` string, then you should add another expectation checking the presence of the accepted badge markup and class

danidoni (changes requested):

danidoni:
Hi @hendrixfan! Thanks for your contribution!

danidoni (approved):

PR #10249, opened 2020-10-02
Show notifications counter on top of bell icon

OleksandrOrlov:
Notifications counter badge shifted bell icon when appeared/disappeared
or when amount of digits was changed.

The commit adjusts CSS styles to show the badge on top of icon, so that
the position of icon is not changed when counter is updated.

How to test the changes:
1. Log in as any user that has notifications;
2. Proceed to the Notifications Page;
3. Mark several notifications as "Read", so that amount of digits in notifications counter badge is changed (or mark all as read o make the badge disappear);
4. Focus on the bell icon.

Expected result:
Bell icon does not change its position.

Here are screenshots of how it looks:

**Before**
before_counter

**After**
1_digit_counter

2_digit_counter

3_digit_counter

4_digit_counter

### DIFF
--- a/src/api/app/assets/stylesheets/webui/notifications_redesign/notifications.scss
+++ b/src/api/app/assets/stylesheets/webui/notifications_redesign/notifications.scss
@@ -69,3 +69,16 @@ body.notifications-redesign  {
     }
   }
 }
+
+.notifications-counter {
+  position: relative;
+  margin: 0 auto;
+  width: 6rem;
+  .badge {
+    position: absolute;
+    top: -0.4rem;
+    left: 3.1rem;
+    border: 0.1rem solid $dark;
+    padding: 0.1rem 0.3rem;
+  }
+}


danidoni (approved):
Awesome!

DavidKang:
Should this change also be applied to bottom navigation?
Screenshot_2020-10-05 Open Build Service

dmarcoux:
> Should this change also be applied to bottom navigation?
> Screenshot_2020-10-05 Open Build Service

Yes, it should.

dmarcoux (changes requested):
As pointed out by @DavidKang, let's also do this for the bottom navigation.

OleksandrOrlov:
@dmarcoux I have created another PR to make notification counter in the bottom panel to be also rendered as partial and to allow counter updating with AJAX, as it was already done for the counter in the Top panel. After the PR will be merged, I'll provide the proper style here, as it will be the same for the top and for the bottom counters.

dmarcoux (approved):

PR #10255, opened 2020-10-05
Update notifications counter in Bottom Nav with AJAX

OleksandrOrlov:
Notifications counter did not update in real-time, when marking a
notification as read/unread. Only when page was refreshed, it was
updated.

The commit changes the behavior to use AJAX and update the counter every
time when action on notification is performed, so that there is no need
to update the page to see the actual amount of unread notifications in
the counter.

Peek 2020-10-05 16-04

### DIFF
--- a/src/api/app/views/layouts/webui/responsive_ux/_unread_notifications_counter.html.haml
+++ b/src/api/app/views/layouts/webui/responsive_ux/_unread_notifications_counter.html.haml
@@ -1,4 +1,4 @@
-= link_to(my_notifications_path, class: 'nav-link text-light p-0 w-100', alt: 'Notifications') do
+.notifications-counter
   %i.fas.fa-bell
   - unless User.session.unread_notifications.zero?
     %span.badge.badge-primary.align-text-top= User.session.unread_notifications
--- a/src/api/app/views/webui/users/notifications/_update.js.erb
+++ b/src/api/app/views/webui/users/notifications/_update.js.erb
@@ -1,5 +1,5 @@
 $('#filters').html("<%= j(render partial: 'notifications_filter', locals: { filter: notifications_filter }) %>");
 $('#notifications-list').html("<%= j(render partial: 'notifications_list', locals: { notifications: notifications, selected_filter: notifications_filter.selected_filter }) %>");
 $('#flash').html("<%= escape_javascript(render(layout: false, partial: 'layouts/webui/flash', object: flash)) %>");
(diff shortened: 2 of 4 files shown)


danidoni (approved):

src/api/app/views/layouts/webui/responsive_ux/_unread_notifications_counter.html.haml
```diff
@@ -1,4 +1,4 @@
-= link_to(my_notifications_path, class: 'nav-link text-light p-0 w-100', alt: 'Notifications') do
+.notifications-counter
```
dmarcoux, line 1: The link isn't in this partial anymore since the CSS classes are slightly different depending on where we render this partial, right? I'm asking to be sure I understand the reasoning behind this change.

src/api/app/views/webui/users/notifications/_update.js.erb
```diff
@@ -1,5 +1,6 @@
 $('#filters').html("<%= j(render partial: 'notifications_filter', locals: { filter: notifications_filter }) %>");
 $('#notifications-list').html("<%= j(render partial: 'notifications_list', locals: { notifications: notifications, selected_filter: notifications_filter.selected_filter }) %>");
 $('#flash').html("<%= escape_javascript(render(layout: false, partial: 'layouts/webui/flash', object: flash)) %>");
-$('#notifications-counter').html("<%= escape_javascript(render partial: 'layouts/webui/responsive_ux/unread_notifications_counter') %>");
+$('#top-notifications-counter').html("<%= escape_javascript(render partial: 'layouts/webui/responsive_ux/unread_notifications_counter') %>");
```
dmarcoux, line 4: The lines 4 and 5 can be combined:

```suggestion
$('#top-notifications-counter, #bottom-notifications-counter').html("<%= escape_javascript(render partial: 'layouts/webui/responsive_ux/unread_notifications_counter') %>");
```

src/api/app/views/layouts/webui/responsive_ux/_unread_notifications_counter.html.haml
OleksandrOrlov: Yep, this is because their classes differs a bit.

src/api/app/views/webui/users/notifications/_update.js.erb
OleksandrOrlov: Thanks for the hint:)

OleksandrOrlov: Done.

dmarcoux (approved):

PR #10257, opened 2020-10-06
Disable buttons properly and show spinner on submit in notifications list

### LINKED ISSUE #10248: Notification: Mark as Read button is not really disabled
The look and feel of the button is like a disabled one, but it still working, making requests to the server.

Screenshot_2020-10-02 openSUSE Build Service(2)

krauselukas:
The button on the notification page was disabled through
a CSS class, which only lead to a disabled look, but still
allowed to submit the form.

Fixes #10248

There was no feedback to the user when marking notifications
as read and unread. When marking a bigger amount of notifications
it can take a bit to update the view, which might lead to additional
submit attempts. By showing a spinner and disabling the button
we can prevent this from happening.

### DIFF
--- a/src/api/app/assets/javascripts/webui/notification.js
+++ b/src/api/app/assets/javascripts/webui/notification.js
@@ -14,10 +14,10 @@ function setCheckboxCounterAndSubmitButton() {
   });
 
   if(amountBoxesChecked <= 0) {
-    $('#done-button').addClass('disabled');
+    $('#done-button').prop('disabled', true);
     $('#select-all-label').text('Select all');
   } else {
-    $('#done-button').removeClass('disabled');
+    $('#done-button').prop('disabled', false);
     $('#select-all-label').text(amountBoxesChecked + ' selected');
   }
 }
--- a/src/api/app/assets/javascripts/webui/notification.js
+++ b/src/api/app/assets/javascripts/webui/notification.js
@@ -29,4 +29,3 @@ function handleNotificationCheckboxSelection() { // jshint ignore:line
     setCheckboxCounterAndSubmitButton();
   });
 }
-
--- a/src/api/app/views/webui/users/notifications/_notification_action_bar.html.haml
+++ b/src/api/app/views/webui/users/notifications/_notification_action_bar.html.haml
@@ -6,5 +6,5 @@
           = check_box_tag('select-all-notifications', 1, false, class: 'custom-control-input')
           %label.custom-control-label.align-middle.mr-4#select-all-label{ for: 'select-all-notifications' } Select All
           - button_text = type == 'read' ? "Mark as 'Unread'" : "Mark as 'Read'"
-          = button_tag(type: 'submit', class: 'btn btn-sm btn-outline-success px-3 disabled', id: 'done-button') do
+          = button_tag(type: 'submit', class: 'btn btn-sm btn-outline-success px-3', id: 'done-button', disabled: 'disabled') do
             = button_text
--- a/src/api/app/views/webui/users/notifications/_notification_action_bar.html.haml
+++ b/src/api/app/views/webui/users/notifications/_notification_action_bar.html.haml
@@ -5,6 +5,9 @@
         .custom-control.custom-checkbox
           = check_box_tag('select-all-notifications', 1, false, class: 'custom-control-input')
           %label.custom-control-label.align-middle.mr-4#select-all-label{ for: 'select-all-notifications' } Select All
-          - button_text = type == 'read' ? "Mark as 'Unread'" : "Mark as 'Read'"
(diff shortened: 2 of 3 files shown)


src/api/app/assets/javascripts/webui/notification.js
```diff
@@ -30,3 +30,9 @@ function handleNotificationCheckboxSelection() { // jshint ignore:line
   });
 }
 
+function setSpinnerAndDisableButtons() { // jshint ignore:line
```
dmarcoux, line 33: I believe this function could be replaced with [automatic disabling from Rails unobtrusive JavaScript](https://guides.rubyonrails.org/working_with_javascript_in_rails.html#automatic-disabling). This implies setting on the button the data attribute `data-disable-with: 'Marking as "Read"...'` (or Unread depending on the case)

krauselukas: ahh cool, I will have a look

krauselukas: @dmarcoux @eduardoj I applied the ujs way, it know disables the button and show a spinner instead of the text, let me know what you think :)

src/api/app/views/webui/users/notifications/_notification_action_bar.html.haml
```diff
@@ -6,5 +6,7 @@
           = check_box_tag('select-all-notifications', 1, false, class: 'custom-control-input')
           %label.custom-control-label.align-middle.mr-4#select-all-label{ for: 'select-all-notifications' } Select All
           - button_text = type == 'read' ? "Mark as 'Unread'" : "Mark as 'Read'"
-          = button_tag(type: 'submit', class: 'btn btn-sm btn-outline-success px-3 disabled', id: 'done-button') do
+          = button_tag(type: 'submit', class: 'btn btn-sm btn-outline-success px-3', id: 'done-button', disabled: 'disabled',
+                       'data-disable-with': "<i class='fas fa-spinner fa-spin'></i>".html_safe) do
```
dmarcoux, line 10: It looks a bit weird with only a spinner, but it's not a blocker.

dmarcoux:
All check marks are spinning when clicking on one and it doesn't work anymore. I marked one notification as read, then it stopped working. 

all-checkmarks-are-spinning

eduardoj (dismissed):
@dmarcoux:

> All check marks are spinning when clicking on one and it doesn't work anymore. I marked one notification as read, then it stopped working.

Right. That needs to be fixed.

src/api/app/views/webui/users/notifications/_notification_action_bar.html.haml
krauselukas: @dmarcoux I agree, the spinner does now only show up on the button that got clicked

krauselukas:
> All check marks are spinning when clicking on one and it doesn't work anymore. I marked one notification as read, then it stopped working.
> 
> all-checkmarks-are-spinning

@eduardoj @dmarcoux Fixed, please have another look, I forgot to commit a change

krauselukas:
@eduardoj @dmarcoux sorry, please wait with another review, there is something wrong again...

krauselukas:
@eduardoj @dmarcoux  Now its ready, I fixed a rubocop offence and applied it wrongly

dmarcoux:
For the button with the check mark, replacing the icon with the spinner would look better. It's already something we do in the autocomplete, like on the users page of a project/package, when adding a user/group. 

For the `Mark as 'Read'` button, I'm not sure if we show the loading spinner on the left or right of the button text normally. Could you check this? This is to be consistent.

krauselukas:
> For the button with the check mark, replacing the icon with the spinner would look better. It's already something we do in the autocomplete, like on the users page of a project/package, when adding a user/group.
> 
> For the `Mark as 'Read'` button, I'm not sure if we show the loading spinner on the left or right of the button text normally. Could you check this? This is to be consistent.

We don't really have the combination with text and spinner in a button (at least I couldn't find one), but yes with checkboxes for example with show it to the right, I can adopt this

DavidKang:
> We don't really have the combination with text and spinner in a button (at least I couldn't find one), but yes with checkboxes for example with show it to the right, I can adopt this

There is something similiar in https://obs-reviewlab.opensuse.org/krauselukas-fixdisabled_button/architectures, when you click on 'all' and 'none' buttons

eduardoj:
> > All check marks are spinning when clicking on one and it doesn't work anymore. I marked one notification as read, then it stopped working.
> 
> Right. That needs to be fixed.

This got addressed. Dismissing my previous "Request changes" review.

krauselukas:
@dmarcoux the spinner is now right aligned to the button text. For the buttons that are marking a single notification, there is now only the spinner without the icon

dmarcoux (approved):

PR #10258, opened 2020-10-06
Replace `require_login` with Pundit in Webui::Users::SubscriptionsController

intrip:

Tackles `Webui::Users::SubscriptionsController`

Ref #10083

### DIFF
--- a/src/api/app/controllers/webui/users/subscriptions_controller.rb
+++ b/src/api/app/controllers/webui/users/subscriptions_controller.rb
@@ -1,11 +1,14 @@
 class Webui::Users::SubscriptionsController < Webui::WebuiController
-  before_action :require_login
+  # TODO: Remove this when we'll refactor kerberos_auth
+  before_action :kerberos_auth
+
+  after_action :verify_authorized
 
   def index
+    @subscriptions_form = authorize(subscriptions_form(default_form: params[:default_form]))
+
     @user = User.session!
     @groups_users = @user.groups_users
-
-    @subscriptions_form = subscriptions_form(default_form: params[:default_form])
 
     respond_to do |format|
       format.html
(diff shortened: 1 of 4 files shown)


src/api/app/policies/event_subscription_policy.rb
```diff
@@ -0,0 +1,31 @@
+class EventSubscriptionPolicy < ApplicationPolicy
+  def initialize(user, record, opts = {})
+    super(user, record, opts.merge(ensure_logged_in: true))
+  end
+
+  def index?
+    record_of_user_and_group?(record)
```
intrip, line 7: This code path is not used (yet) but the idea is to add meaningful policies that could be reused across the codebase, in order to achieve this I've added what (I think) is the expected logic for this Policy.

Please double check this because I've just inferred this from the codebase without any prior domain knowledge 🤷‍♂️

intrip:
@dmarcoux could you please review this? 🙏

src/api/app/controllers/webui/users/subscriptions_controller.rb
```diff
@@ -1,5 +1,9 @@
 class Webui::Users::SubscriptionsController < Webui::WebuiController
-  before_action :require_login
+  # TODO: Remove this when we'll refactor kerberos_auth
+  before_action :kerberos_auth
+  before_action -> { authorize(EventSubscription.new) }
```
dmarcoux, line 4: Even though this controller is named `SubscriptionsController`, it's working with `EventSubscription::Form`, not `EventSubscription` (at least not directly).

Checking the `subscriber` of the `EventSubscription::Form` object seems like the right thing to do here. If `subscriber` is `nil` or the currently logged-in user, then they are authorized.

src/api/app/controllers/webui/users/subscriptions_controller.rb
```diff
@@ -1,29 +1,38 @@
...
     User.session!.groups_users.each do |gu|
       gu.email = params[gu.group.title] == '1'
       gu.save
     end
 
-    subscriptions_form.update!(params[:subscriptions]) if params[:subscriptions]
-    flash.now[:success] = 'Notifications settings updated'
-  rescue ActiveRecord::RecordInvalid => e
-    flash.now[:error] = "Notifications settings could not be updated due to an error: #{e.message}"
-  ensure
```
intrip, line 26: I've removed the `ensure` catch all error logic because it led into a double render error when `Pundit::NotAuthorizedError` was raised within this action.

src/api/app/controllers/webui/users/subscriptions_controller.rb
intrip: Good suggestion, makes completely sense! I've introduced a `EventSubscription::FormPolicy`. In order to achieve that I had to introduce a few extra changes, LMK your your take on this 🙂

intrip:
@dmarcoux I've pushed some changes to follow-up your suggestion

src/api/app/controllers/webui/users/subscriptions_controller.rb
```diff
@@ -1,29 +1,38 @@
 class Webui::Users::SubscriptionsController < Webui::WebuiController
-  before_action :require_login
+  # TODO: Remove this when we'll refactor kerberos_auth
+  before_action :kerberos_auth
+
+  after_action :verify_authorized
 
   def index
+    @subscriptions_form = subscriptions_form(default_form: params[:default_form])
```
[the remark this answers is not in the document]
intrip: Good to know, TIL!
I've just committed the suggestions.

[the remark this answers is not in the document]
intrip: Well, just noticed the suggestion miss the line deletion, I'll fix it manually and push again

intrip:
@dmarcoux I've just pushed the suggestion, thanks!

dmarcoux (approved):
Nice work :)

PR #10264, opened 2020-10-07
Replace `require_login` with Pundit in Webui::Users::RssTokensController

intrip:

Tackles `Webui::Users::RssTokensController`

Ref #10083

### DIFF
--- a/src/api/app/controllers/webui/users/rss_tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/rss_tokens_controller.rb
@@ -1,18 +1,21 @@
 module Webui
   module Users
     class RssTokensController < WebuiController
-      before_action :require_login
+      # TODO: Remove this when we'll refactor kerberos_auth
+      before_action :kerberos_auth
+
+      after_action :verify_authorized
 
       def create
-        token = User.session!.rss_token
-        if token
+        token = authorize(Token::Rss.find_or_initialize_by(user: User.session))
+        if token.persisted?
           flash[:success] = 'Successfully re-generated your RSS feed url'
           token.regenerate_string
-          token.save
         else
           flash[:success] = 'Successfully generated your RSS feed url'
-          User.session!.create_rss_token
         end
+        token.save
+
         redirect_back(fallback_location: my_subscriptions_path)
       end
     end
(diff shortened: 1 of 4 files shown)


src/api/spec/controllers/webui/users/rss_tokens_controller_spec.rb
```diff
@@ -4,18 +4,20 @@
...
-    before do
-      login(user)
+    it_behaves_like 'require logged in user' do
+      let(:method) { :post }
+      let(:action) { :create }
     end
 
     context 'with a user with an existent token' do
-      let(:last_token) { user.create_rss_token.string }
+      let!(:last_token) { user.create_rss_token.string }
```
intrip, line 13: I think there was a little bug in the spec, I've updated the `let!` and the flash message expectation.

intrip:
@dmarcoux could you take a peek 🙏 ?

danidoni (approved):

src/api/app/controllers/webui/users/rss_tokens_controller.rb
```diff
@@ -1,20 +1,30 @@
...
-      before_action :require_login
+      # TODO: Remove this when we'll refactor kerberos_auth
+      before_action :kerberos_auth
+
+      after_action :verify_authorized
 
       def create
-        token = User.session!.rss_token
-        if token
+        token = User.session ? find_or_build_rss_token : Token::Rss.new
```
dmarcoux, line 10: The following line achieves the same, but without an extra method: 

```suggestion
        token = authorize(Token::Rss.find_or_initialize_by(user: User.session))
```

dmarcoux (changes requested):

src/api/app/controllers/webui/users/rss_tokens_controller.rb
intrip: Neat improvement! Recently working with Rails 3.2 led me into forgetting some features 😅

intrip:
@dmarcoux suggestion applied, thanks!

src/api/app/controllers/webui/users/rss_tokens_controller.rb
dmarcoux: Hehehe, I know the feeling.... I once had to maintain a Rails `2.3` application using Ruby `1.8.7` in 2015-2016.... I'm having nightmares again

dmarcoux (approved):

PR #10266, opened 2020-10-07
Make sidebar scrollable if it does not fit into viewport

OleksandrOrlov:
Sidebar items were hidden over the viewport boundaries when sidebar does
not fit into the screen.

The commit adds CSS styles to show scrollbar when the items in sidebar
go over the viewport boundaries.

To verify the feature:

1. Log in to the application;
2. Decrease browser's height, so that sidebar items go out of viewport boundaries;
3. Focus on the sidebar.

Expected result:
Scrollbar appears. All the previous behavior remains same (e.g. scrollbar is not moving when scrolling the content area).

Here are screenshots of how it looks:

**Before**
before

**After**
after

### DIFF
--- a/src/api/app/assets/stylesheets/webui/responsive_ux/grid.scss
+++ b/src/api/app/assets/stylesheets/webui/responsive_ux/grid.scss
@@ -38,9 +38,12 @@
       display: block;
 
       #left-navigation {
-        position: fixed;
+        position: sticky;
         top: $top-navigation-height;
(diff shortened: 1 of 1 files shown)


dmarcoux (approved):

danidoni (approved):

OleksandrOrlov:
As soon as I have 2 approves, I'm merging it:)

PR #10268, opened 2020-10-07
Feature: saved replies

arharovets:
Hello everyone,

this PR introduces saved replies feature with functionality similar to [GitHub saved replies](https://docs.github.com/en/github/writing-on-github/working-with-saved-replies). A user can create/update/remove replies, and use any of those replies to fill any comment input field.

To verify this feature
- create sample data (including a project with packages, right now this feature only works there)
- go to the root page, click 'Saved replies' link
Screenshot_2020-10-07 Open Build Service
- create several replies with title and body
- go to the package overview page
Screenshot_2020-10-07 Alone on a Wide, Wide Sea
- click on the dropdown list and choose a reply you would like to fill the comment input field with. When you click it, reply text should appear in the input field.

This is a work in progress, so, at the moment, it looks ugly.

### DIFF
--- a/src/api/app/controllers/webui/comment_snippets_controller.rb
+++ b/src/api/app/controllers/webui/comment_snippets_controller.rb
@@ -3,7 +3,7 @@ class Webui::CommentSnippetsController < Webui::WebuiController
   before_action :check_displayed_user
 
   def index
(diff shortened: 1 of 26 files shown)


dmarcoux:
We won't be introducing this feature in OBS, at least, not now.

PR #10263, opened 2020-10-07
Move the actions present in the user's profile page to the left menu

danidoni:
We want to move all user's profile page action links into the left menu, inside the Actions area.

Those actions are, **Edit your account** and **Change your password**

This is how it looks like on small screens:
Peek 2020-10-08 17-30

This is how it looks like on large screens:
image

PR #10269, opened 2020-10-08
Move actions to sidebar on the configuration page

danidoni:
**Create Users from the Actions menu**

image

**Create Groups from the Actions menu**

image

### DIFF
--- a/src/api/app/views/webui/users/index.html.haml
+++ b/src/api/app/views/webui/users/index.html.haml
@@ -17,10 +17,17 @@
             Actions
       %tbody
 
-    .pt-4
-      = link_to(new_user_path(pagetitle: 'create user', submit_btn_text: 'create')) do
-        %i.fas.fa-plus-circle.text-primary
-        Create User
+    - if feature_enabled?(:responsive_ux)
+      - content_for :actions do
+        %li.nav-item
+          = link_to(new_user_path(pagetitle: 'create user', submit_btn_text: 'create'), class: 'nav-link') do
+            %i.fas.fa-lg.mr-2.fa-plus-circle
+            Create User
+    - else
+      .pt-4
+        = link_to(new_user_path(pagetitle: 'create user', submit_btn_text: 'create')) do
+          %i.fas.fa-plus-circle.text-primary
+          Create User
 
 - content_for :ready_function do
   initializeUserConfigurationDatatable("#{::Configuration.ldap_enabled?}");
--- a/src/api/app/views/webui/groups/index.html.haml
+++ b/src/api/app/views/webui/groups/index.html.haml
@@ -17,10 +17,17 @@
             %td
               = safe_join(group.users.map { |user| link_to(truncate(user.login, length: 20), user_path(user), title: user.login) }, ', ')
 
-    .pt-4
-      = link_to(group_new_path) do
-        %i.fas.fa-plus-circle.text-primary
-        Create Group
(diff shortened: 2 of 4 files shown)


danidoni:

src/api/app/views/webui/users/index.html.haml
```diff
@@ -18,9 +18,14 @@
       %tbody
 
     .pt-4
-      = link_to(new_user_path(pagetitle: 'create user', submit_btn_text: 'create')) do
-        %i.fas.fa-plus-circle.text-primary
-        Create User
+      - if feature_enabled?(:responsive_ux)
+        - content_for :actions do
+          %li.nav-item
+            = link_to_action('Create User', new_user_path(pagetitle: 'create user', submit_btn_text: 'create'), icons: 'fa-plus-circle fa-lg mr-2')
```
DavidKang, line 24: @danidoni, maybe you could also move the other classes inside the helper `fa-lg mr-2`.

src/api/app/views/webui/users/index.html.haml
```diff
@@ -18,9 +18,14 @@
       %tbody
 
     .pt-4
-      = link_to(new_user_path(pagetitle: 'create user', submit_btn_text: 'create')) do
-        %i.fas.fa-plus-circle.text-primary
-        Create User
+      - if feature_enabled?(:responsive_ux)
+        - content_for :actions do
+          %li.nav-item
```
DavidKang, line 23: I think line should be removed, right?

src/api/app/views/webui/users/index.html.haml
DavidKang: I think it would easier for future changes and to have all the icons with the same size and margin. The same happens in your other PR https://github.com/openSUSE/open-build-service/pull/10263, these classes are repeated.

src/api/app/helpers/webui/responsive_ux/actions_helper.rb
```diff
@@ -0,0 +1,11 @@
+module Webui::ResponsiveUx::ActionsHelper
+  def link_to_action(label, url, args)
```
dmarcoux, line 2: This is nice to avoid repetition, but for something as simple as a link with an icon and some text, I would be against introducing this change. I prefer repeating some simple code over an abstraction like this helper method.

src/api/app/views/webui/users/index.html.haml
danidoni: Yes, you are right!

src/api/app/views/webui/users/index.html.haml
danidoni: You are right, again!

src/api/app/helpers/webui/responsive_ux/actions_helper.rb
danidoni: After thinking for a while I've decided to not use the helper method

src/api/app/views/webui/users/index.html.haml
danidoni: I went with not using the helper method as it introduced some dependencies in the rest of branches I'm working on. This is simpler

src/api/app/views/webui/users/index.html.haml
danidoni: I've decided to not use a helper method, so this won't be needed.

dmarcoux (approved):

PR #10270, opened 2020-10-08
Prevent exception on empty URL in markdown text

### LINKED ISSUE #10259: [production][comments#create] bad argument (expected URI object or URI string)
comments#create threw this 31 times
<pre>
bad argument (expected URI object or URI string)
</pre>
[https://build.opensuse.org/comments](https://build.opensuse.org/comments)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5f74ccb34fdf3e0016e17c8b)

krauselukas:
An empty URL part in markdown text leads to an exception
in the mardown renderer. When a comment
includes such an empty URL string, it ends up with a
code 500. This happens when people post code examples
which contain the same syntax as a markdown link.
Instead of throwing an exception, we can just render
the original content incase of an empty URL part.

Kudos to @eduardoj

Fixes #10259

### DIFF
--- a/src/api/lib/obsapi/markdown_renderer.rb
+++ b/src/api/lib/obsapi/markdown_renderer.rb
@@ -30,6 +30,10 @@ module OBSApi
 
     # unfortunately we can't call super (into C) - see vmg/redcarpet#51
     def link(link, title, content)
+      # A return value of nil will not output any data
+      # the contents of the span will be copied verbatim
+      return nil if link.blank?
+
       title = " title='#{title}'" if title.present?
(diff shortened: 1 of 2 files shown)


krauselukas:
The PR is WIP since I want to check for a more elegant solution here

eduardoj (approved):

danidoni (approved):

krauselukas:
coverage/coveralls is hanging, going to merge

PR #10271, opened 2020-10-08
User profile layout redesign

saraycp:
New layout for the user profile page. Distributing _Involved Projects/Packages_ and _Contribution Graph_ in tabs when both are present.

**No tabs when Contribution Graph is disabled:**

Screenshot_2020-10-08 Profile of Admin

**Tabs when Contribution Graph is enabled:**

Screenshot_2020-10-08 Profile of Admin(2)

**No tabs on small viewports no matter if the Contribution Graph is desabled or enabled**

Screenshot_2020-10-08 Profile of Admin(1)

### DIFF
--- a/src/api/app/views/webui/user/user_profile_redesign/_involvement_and_activity.html.haml
+++ b/src/api/app/views/webui/user/user_profile_redesign/_involvement_and_activity.html.haml
@@ -1,0 +1,38 @@
+.card{ class: CONFIG['contribution_graph'] == :off ? 'd-block' : 'd-block d-md-none' }
+  -# Header that replaces the tabs when the contribution graph is not displayed (in small viewports or when not enabled).
+  .card-header
+    %h5 Involved Projects and Packages
+  .card-body
+    = render partial: 'webui/user/involvement', locals: { user: user,
+                                                          owned: owned,
+                                                          involved_projects: involved_projects,
+                                                          involved_packages: involved_packages }
+
+-# Displaying tabs in medium-large viewports when the Contribution Graph is enabled
+- unless CONFIG['contribution_graph'] == :off
+  .card.d-none.d-md-block
+    -# Tabs
+    .bg-light
+      %ul.nav.nav-tabs.pt-2.px-3.flex-nowrap.collapsible{ 'role': 'tablist' }
+        %li.nav-item
+          = link_to('#involved-projects-and-packages', id: 'involved-projects-and-packages-tab', class: 'nav-link active text-nowrap',
+                    data: { toggle: 'tab' }, role: 'tab', aria: { controls: 'involved-projects-and-packages', selected: true }) do
+            Involved Projects/Packages
+        %li.nav-item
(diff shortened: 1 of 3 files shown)


src/api/app/assets/stylesheets/webui/user_profile_redesign/user_profile.scss
```diff
@@ -0,0 +1,5 @@
+// TODO: As soon as user-profile-redesign feature is 100% rolled out,
+// all the 'body.user-profile-redesign' wrappers can be removed.
+
+body.user-profile-redesign  {
+}
```
vpereira, line 5: add a new line, please

src/api/app/assets/stylesheets/webui/application.scss
```diff
@@ -62,6 +62,7 @@
 
 @import 'responsive_ux/responsive_ux';
 @import 'notifications_redesign/notifications';
+@import 'user_profile_redesign/user_profile';
```
DavidKang, line 65: I would avoid add an "empty" css file for the future. I would add it when we really need it

saraycp: Alright, I'm taking out the corresponding commit, I'll add it in the next PR if needed.

src/api/app/assets/stylesheets/webui/user_profile_redesign/user_profile.scss
saraycp: Done. But I've taken out the corresponding commit, I'll add it to the next PR.

src/api/app/views/webui/user/user_profile_redesign/_involvement_and_activity.html.haml
```diff
@@ -0,0 +1,38 @@
...
+        .tab-pane.fade.show.active#involved-projects-and-packages{ role: 'tabpanel', aria: { labelledby: 'involved-projects-and-packages-tab' } }
+          = render partial: 'webui/user/involvement', locals: { user: user,
+                                                                owned: owned,
+                                                                involved_projects: involved_projects,
+                                                                involved_packages: involved_packages }
+
+        .tab-pane.fade#contributions{ role: 'tabpanel', aria: { labelledby: 'contributions-tab' } }
+          = render partial: 'webui/user/activity', locals: { activity_hash: activity_hash,
+                                                             first_day: first_day,
+                                                             last_day: last_day }
```
hennevogel, line 38: Why is the logic, what get's shown and what not, done with CSS *after* you render everything?

saraycp: A bit of explanation can be useful here.

When **Contribution Graph** is **off**, the tabs and their contents are not rendered. So that part is ok (line 11 on).

However, when **Contribution Graph** is **on**, the tabs should be displayed in large viewports. But in small ones, we should only display a common card (lines 1-10). As we discussed offline and after doing a small change, the code will stay the same in this case, using CSS to display or hide the tabs.

vpereira (approved):

PR #10278, opened 2020-10-09
Replace `require_login` with Pundit in Webui::Users::NotificationsCon…

intrip:

Tackles `Webui::Users::NotificationsController`

Ref #10083

### DIFF
--- a/src/api/app/policies/notification_policy.rb
+++ b/src/api/app/policies/notification_policy.rb
@@ -1,5 +1,11 @@
 class NotificationPolicy < ApplicationPolicy
   class Scope < Scope
+    def initialize(user, scope)
+      raise Pundit::NotAuthorizedError, reason: ApplicationPolicy::ANONYMOUS_USER if user.nil? || user.is_nobody?
+
+      super(user, scope)
+    end
+
     def resolve
       NotificationsFinder.new(scope).for_subscribed_user(user)
     end
--- a/src/api/app/controllers/webui/users/notifications_controller.rb
+++ b/src/api/app/controllers/webui/users/notifications_controller.rb
@@ -2,9 +2,12 @@ class Webui::Users::NotificationsController < Webui::WebuiController
   MAX_PER_PAGE = 300
   VALID_NOTIFICATION_TYPES = ['read', 'reviews', 'comments', 'requests', 'unread'].freeze
 
-  before_action :require_login
+  # TODO: Remove this when we'll refactor kerberos_auth
+  before_action :kerberos_auth
   before_action :check_param_type, :check_param_project, only: :index
   before_action :check_feature_toggle
+
+  after_action :verify_policy_scoped
 
   def index
     @notifications = paginated_notifications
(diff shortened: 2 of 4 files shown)


src/api/app/policies/notification_policy.rb
```diff
@@ -1,4 +1,16 @@
...
+    super(user, record, opts.merge(ensure_logged_in: true))
+  end
+
+  def index?
+    true
+  end
+
+  def update?
+    true
+  end
```
intrip, line 12: I'm not totally convinced of this approach thus I'd like to further discuss it: the idea is that since we introduced `verify_policy_scoped` in the controller checking for the `subscriber` in `NotificationPolicy#update` it's redundant and also leads to a slowdown of the related endpoint (trade-off), moreover we also have tests in the controller to ensure we cannot update other user notifications. On the other hand if we look at the `NotificationPolicy#update` per-se we don't see that constraint and therefore is misleading.

intrip:
@dmarcoux could you take a look here? I've followed an approach to authorize collections but I'd like to know your take on this.

dmarcoux:
We are already [scoping](https://github.com/varvet/pundit#scopes) the notifications to the currently logged-in user in the `fetch_notifications` method https://github.com/openSUSE/open-build-service/blob/6e1e25f6aed699c00bc9fcf8d1ff0090bc62ba70/src/api/app/controllers/webui/users/notifications_controller.rb#L75-L82

This is how we handled authorization for collections of `Notification` in `NotificationPolicy`. `fetch_notifications` is then called in the `index` and `update` actions, through `paginated_notifications` or directly.

Now, to get rid of `require_login`, I wonder if only setting up the `initialize` method as you did is enough. I believe it would be, but I'm not sure how it behaves for anonymous users. This would have to be tested. If `initialize` is enough, the methods `index?` and `update?` in `NotificationPolicy` won't be needed since we're scoping the notifications as I described above.

intrip:
@dmarcoux 
Agree with you about the scoping , I've also added `verify_policy_scoped` and [a test](https://github.com/intrip/open-build-service/blob/refactor_require_login_user_notifications/src/api/spec/controllers/webui/users/notifications_controller_spec.rb#L132-L143) to make sure we won't introduce regressions on that side (the `index` was already covered).

Related to `require_login` I think that the `initialize` check is enough: I've added both [index tests](https://github.com/intrip/open-build-service/blob/refactor_require_login_user_notifications/src/api/spec/controllers/webui/users/notifications_controller_spec.rb#L30-L33) and [update tests](https://github.com/intrip/open-build-service/blob/refactor_require_login_user_notifications/src/api/spec/controllers/webui/users/notifications_controller_spec.rb#L110-L113) to ensure that everything works as expected for anonymous users.
My main concerns were on the readability of the `NotificationPolicy`: having 2 methods (`update?`, `index?`) that always returns `true` is misleading.
I took a deeper look at `Pundit` implementation, since `authorize` [requires (or infers) a `query`](https://github.com/varvet/pundit/blob/v2.1.0/lib/pundit.rb#L216) I went for a little different approach: we just [initialize the policy in a filter](https://github.com/intrip/open-build-service/blob/refactor_require_login_user_notifications/src/api/app/controllers/webui/users/notifications_controller.rb#L7), this allow us to get rid of both the `index?` and `update?` methods in the policy.

Could you please review the changes?

dmarcoux:
Using a closed scope is a great idea.

dmarcoux (approved):

PR #10286, opened 2020-10-13
Replace `require_login` with Pundit in Webui::UsersController

intrip:

Tackles `Webui::UsersController`

Ref #10083

### DIFF
--- a/src/api/app/controllers/webui/users_controller.rb
+++ b/src/api/app/controllers/webui/users_controller.rb
@@ -1,9 +1,13 @@
 class Webui::UsersController < Webui::WebuiController
-  before_action :require_login, only: [:index, :edit, :destroy, :update, :change_password, :edit_account]
+  # TODO: Remove this when we'll refactor kerberos_auth
+  before_action :kerberos_auth, only: [:index, :edit, :destroy, :update, :change_password, :edit_account]
+  before_action :authorize_user, only: [:index, :edit, :destroy, :update, :change_password, :edit_account]
   before_action :require_admin, only: [:index, :edit, :destroy]
   before_action :check_displayed_user, only: [:show, :edit, :update, :edit_account]
   before_action :role_titles, only: [:show, :edit_account, :update]
   before_action :account_edit_link, only: [:show, :edit_account, :update]
+
+  after_action :verify_authorized, only: [:index, :edit, :destroy, :update, :change_password, :edit_account]
 
   def index
     respond_to do |format|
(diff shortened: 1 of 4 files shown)


src/api/app/controllers/webui/users_controller.rb
```diff
@@ -179,6 +183,12 @@ def set_all_filters_if_none_is_set(filters, filter_keys)
     end
   end
 
+  def authorize_policy_initialize
+    policy(User)
+    # TODO: Remove when we'll do the authorization checks with Pundit
+    skip_authorization
```
intrip, line 189: This may look a bit odd so let me explain the reasons: Currently in [UserPolicy](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/policies/user_policy.rb#L2-L8) we have some authorization rules used for the `api` which differs for the ones that we need for the `webui`. Adding the authorization rules for the `webui` in this PR (whilst still keeping the other rules for the `api`) would've been a bit complicated and out of scope.

The `skip_authorization` is needed because we don't call `authorize` yet (the logged in check is in `initialize`)

intrip:
@dmarcoux could you take a peek? 🙏

src/api/app/controllers/webui/users_controller.rb
dmarcoux: Could we instead have a user policy for webui and another for the api? I would find this easier to understand when looking at this again in a few weeks/months. We already do this for controllers. The api controllers are directly under `src/api/app/controllers/` while the webui controllers are under `src/api/app/controllers/webui/`. So the policies for the api would be under `src/api/app/policies/` and the ones for webui would be under `src/api/app/policies/webui/`.

What do you think?

intrip: Agree! We'll have to differentiate the 2 policies logic at some point and using different classes looks pretty straightforward 👍  

For the sake of this refactor is fine if we don't port the authorization logic from the controllers into the `webui` policy yet (I'd just port the `require_login` logic for now)?

dmarcoux: It's fine to just port the `require_login` logic for now.

intrip: @dmarcoux I've created a separate policy now, could you please check?

dmarcoux (approved):

dmarcoux:
Nice job :slightly_smiling_face:

PR #10287, opened 2020-10-13
Prevent rendering of carriage return character in xml views

### LINKED ISSUE #10272: 0x13 characters in Edit description of a package via webinterface
When changing the description of a package via "Edit description" in the webinterface the resulting XML contains` &#13;`. This wastes space and looks ugly. Please strip these characters.

Workflow:
* Open Edit description
* Open spec file in another tab
* Copy text from spec file description to package description

krauselukas:
When editing the description of a package/project in the webui,
the text_area in the form add's a sequence of '\r\n' to indicate
a new line. The carriage return indicated by the '\r' gets
rendered in the corresponding xml output for example in the meta
view or through the API. In order to prevent this from happening
we can replace this sequence with a simple '\n' which is the
standard in most recent unix environments to indicate a new line.

Fixes #10272

### DIFF
--- a/src/api/app/views/models/_package.xml.builder
+++ b/src/api/app/views/models/_package.xml.builder
@@ -1,6 +1,6 @@
 xml.package(name: my_model.name, project: my_model.project.name) do
   xml.title(my_model.title)
-  xml.description(my_model.description)
+  xml.description(my_model.description.gsub(/\r\n/, "\n"))
   xml.releasename(my_model.releasename) if my_model.releasename
 
   xml.devel(project: my_model.develpackage.project.name, package: my_model.develpackage.name) if my_model.develpackage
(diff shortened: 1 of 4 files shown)


danidoni (approved):

hennevogel (changes requested):
So we are going to have all descriptions wrong in the database and just display them selectively?

krauselukas:
> How about making sure that filling a HTML form does not produce database entries with HTML entities in it? :)

@hennevogel My idea was to do this as an follow up. Cleaning up faulty entries in the database will involve a big data migration. So maybe for now we clean this on render, and as follow up prevent new faulty data and maybe a cleanup in the end.

krauselukas:
I should have mentioned this here, sorry

hennevogel:
And it sounds to me as sending `\r\n` is the norm with browser. So the question is: Why is this rendered as html entity when we display the data?

krauselukas:
> > Cleaning up faulty entries in the database will involve a big data migration.
> 
> How big? What are the numbers?

So it would be 409049 descriptions that need to be checked

krauselukas:
The root cause seems to be a issue with Nokogiri that escapes the carriage return https://github.com/sparklemotion/nokogiri/issues/1356 I will tackled the problem differently in another PR. Closing...

PR #10288, opened 2020-10-13
Move "Add Group Member" action next to heading of group#show page.

danidoni:
While managing members of a group, this is how it looked like before:

ksnip_20201013-144105

And this is how it looks like now:
image

### DIFF
--- a/src/api/app/views/webui/groups/_group_members.html.haml
+++ b/src/api/app/views/webui/groups/_group_members.html.haml
@@ -1,11 +1,12 @@
 - write_access = policy(group).update?
 
 - if write_access
-  %ul.list-inline
-    %li.list-inline-item
-      = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#add-group-user-modal' }) do
-        %i.fas.fa-plus-circle.text-primary
-        Add Member
+  - unless feature_enabled?(:responsive_ux)
+    %ul.list-inline
+      %li.list-inline-item
+        = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#add-group-user-modal' }) do
+          %i.fas.fa-plus-circle.text-primary
+          Add Member
   = render partial: 'add_group_user_modal', locals: { group: group }
 
 - if group.users.blank?
--- a/src/api/app/views/webui/groups/show.html.haml
+++ b/src/api/app/views/webui/groups/show.html.haml
@@ -32,7 +32,11 @@
   .card-body
     .tab-content
       .tab-pane.show.active#group-members{ aria: { controls: 'group-members' }, role: 'tabpanel' }
-        %h3 Group Members
+        %h3
+          Group Members
+          - if feature_enabled?(:responsive_ux)
+            = link_to('#', data: { toggle: 'modal', target: '#add-group-user-modal' }, title: 'Add Member') do
+              %i.fas.fa-xs.fa-plus-circle.text-primary
         = render(partial: 'group_members', locals: { group: @group })
 
       .tab-pane#reviews-in{ aria: { controls: 'reviews-in' }, role: 'tabpanel' }


danidoni:

dmarcoux (commented):
Adding a member to the group is fine in the context of the page. I don't believe it belongs in the left/bottom navigation. I would even go further and integrate this better in the UI with something like those:

This one is nice if groups tend to only have a few members.
example

This one is nice if groups tend to have a lot of members. Scrolling the page to be able to add a new member is not nice UX.
example-icon-next-to-heading

danidoni:
> Adding a member to the group is fine in the context of the page. I don't believe it belongs in the left/bottom navigation. I would even go further and integrate this better in the UI with something like those:
> 
> This one is nice if groups tend to only have a few members.
> example
> 
> This one is nice if groups tend to have a lot of members. Scrolling the page to be able to add a new member is not nice UX.
> example-icon-next-to-heading

Ok, you have a point on not moving the actions to the left menu, but I think now is not the best time to introduce yet another visual style in the application, because doing it right means applying this style throught all the application. And that is not the purpose of this PR.

Do you agree?

dmarcoux:
The referenced card is about moving actions to the sidebar, but we should only do so when it makes sense. If it doesn't, it's better to leave it as it is. The best is to make it right. It's a small task, especially if we go with the second option in my previous comment.

Aligning the other actions which were mistakenly moved to the left/bottom navigation is also something we should do. It's not time-consuming.

src/api/app/views/webui/groups/_group_members.html.haml
```diff
@@ -1,11 +1,12 @@
...
 - if write_access
-  %ul.list-inline
-    %li.list-inline-item
-      = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#add-group-user-modal' }) do
-        %i.fas.fa-plus-circle.text-primary
-        Add Member
+  - unless feature_enabled?(:responsive_ux)
+    %ul.list-inline
+      %li.list-inline-item
+        = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#add-group-user-modal' }) do
```
dmarcoux, line 7: This should fix the failures in the feature specs.

```suggestion
        = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#add-group-user-modal' }, title: 'Add Member') do
```

dmarcoux:
The title of the PR seems to be wrong. The changes are about the `Manage Groups` and `Manage Users` pages, no?

danidoni:
> The title of the PR seems to be wrong. The changes are about the `Manage Groups` and `Manage Users` pages, no?

Mmm well, true, I use to go to that page from the Staging Project dashboard, clicking on the managers link. So I keep thinking about it as the Staging Manager Group Members.

> It's also not working. Is everything pushed?

Let me check

dmarcoux:
This PR's title and description remain confusing. I would rephrase this as something like `Move "Add Member" action next to heading of group#show page`.

dmarcoux:
The commit also needs to be rephrased, it's still referring to staging project members, but this is about group members.

src/api/app/views/webui/groups/show.html.haml
```diff
@@ -32,7 +32,11 @@
   .card-body
     .tab-content
       .tab-pane.show.active#group-members{ aria: { controls: 'group-members' }, role: 'tabpanel' }
-        %h3 Group Members
+        %h3
+          Group Members
+          - if feature_enabled?(:responsive_ux)
+            = link_to('#', id: 'add-group-member', data: { toggle: 'modal', target: '#add-group-user-modal' }) do
```
dmarcoux, line 38: Setting the title `'Add Member'` instead of an id on this link would make the changes to the feature specs unneeded. It would also improve usability a tiny bit on desktop.

saraycp (approved):
Regarding these two options, I prefer the second one. So this PR looks good to me.

> example

> example-icon-next-to-heading

PR #10289, opened 2020-10-13
Fix TypeError in Backend::Logger

dmarcoux:
This should have been included in #10283. It was overlooked in the review process.


krauselukas (approved):

hennevogel:
Who is making sure we are setting `extended_backend_log` also in the test environment?

PR #10292, opened 2020-10-13
Add biography for user profile

saraycp:
We would like to allow the users to write a short biography about themself. The bio will be displayed in the redesigned profile page... coming soon. This PR only includes the migration.

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -952,6 +952,7 @@ end
 #
 #  id                            :integer          not null, primary key
 #  adminnote                     :text(65535)
+#  biography                     :string(255)      default("")
 #  deprecated_password           :string(255)      indexed
 #  deprecated_password_hash_type :string(255)
 #  deprecated_password_salt      :string(255)
--- a/src/api/app/models/unregistered_user.rb
+++ b/src/api/app/models/unregistered_user.rb
@@ -71,6 +71,7 @@ end
 #
 #  id                            :integer          not null, primary key
 #  adminnote                     :text(65535)
+#  biography                     :string(255)      default("")
 #  deprecated_password           :string(255)      indexed
 #  deprecated_password_hash_type :string(255)
 #  deprecated_password_salt      :string(255)
--- a/src/api/db/migrate/20201013143657_add_biography_to_users.rb
+++ b/src/api/db/migrate/20201013143657_add_biography_to_users.rb
@@ -1,0 +1,10 @@
+class AddBiographyToUsers < ActiveRecord::Migration[6.0]
+  def up
+    add_column :users, :biography, :string
+    change_column_default :users, :biography, ''
+  end
+
+  def down
+    remove_column :users, :biography
+  end
+end
--- a/src/api/db/migrate/20201014105103_backfill_add_biography_to_users.rb
+++ b/src/api/db/migrate/20201014105103_backfill_add_biography_to_users.rb
@@ -1,0 +1,12 @@
+class BackfillAddBiographyToUsers < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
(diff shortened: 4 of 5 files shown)


src/api/app/models/user.rb
```diff
@@ -952,6 +952,7 @@ def hash_string(value)
 #
 #  id                            :integer          not null, primary key
 #  adminnote                     :text(65535)
+#  bio                           :text(65535)
```
vpereira, line 955: A short biography is typically four sentences, approximately 150 to 200 words or less in length.

vpereira (changes requested):

src/api/app/models/user.rb
saraycp: That's true, it makes no sense to store a text for this. I've just changed to String, 255 characters, what I think is not a bad number for a bio. But we can play afterwards with validations in case we decide to allow less characters for design decitions.

vpereira (approved):

src/api/db/migrate/20201013143657_add_bio_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddBioToUsers < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :bio, :string
```
dmarcoux, line 3: It would be nice to have an empty string as a default value so we don't have to check for `nil` in the code.
```suggestion
    add_column :users, :bio, :string
    change_column_default :users, :bio, ''
```

We cannot set the default value directly in the `add_column` statement because our MariaDB version will block writes to a table when adding a column with a default value. See [`Adding a column with a default value` in strong_migrations' documentation](https://github.com/ankane/strong_migrations/tree/1e133224579fe708fd01324504cd4e8fc3f29e71#adding-a-column-with-a-default-value).

dmarcoux (changes requested):

src/api/db/migrate/20201013143657_add_bio_to_users.rb
saraycp: I doubted between "bio" and "biography," so second one is ok for me too.

However, setting an empty string as default value doesn't prevent us to do some check, since we don't want to display an empty `<p></p>`. I wonder what's the benefit. :thinking:

dmarcoux: It's one less check to do if we don't have to deal with `nil`.

saraycp: is it some best practice? Otherwise I don't see the point to add `change_column_default` and a backfill migration if I'm going to use `.blank?` anyway.

dmarcoux: Working with `nil` is for sure more error-prone. Even its creator said it was [an error to invent this concept](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/). He calls `nil` (or `null`) his billion dollar mistake.

We don't need to have the default value, it's just to avoid potential errors later.

saraycp: Thanks for your explanation, @dmarcoux :+1:

saraycp: @dmarcoux updated replacing "bio" by "biography" and setting empty string as default value.

dmarcoux:
There's a typo in the commit message: biogr**a**phy

saraycp:
> There's a typo in the commit message: biogr**a**phy

Thanks, corrected!

dmarcoux (approved):

krauselukas:
Will deploy this tomorrow morning

PR #10294, opened 2020-10-13
Unescape carriage return char in nokogiri xml output

### LINKED ISSUE #10272: 0x13 characters in Edit description of a package via webinterface

krauselukas:
The carriage return character '\r' gets escaped by
nokogiri/libxml which causes the character to be
rendered in the xml output.

https://github.com/sparklemotion/nokogiri/issues/1356

Since this is a valid xml char we can workaround this problem
by unescaping it again.

Fixes #10272

### DIFF
--- a/src/api/config/initializers/nokogiri_builder.rb
+++ b/src/api/config/initializers/nokogiri_builder.rb
@@ -10,8 +10,8 @@ module ActionView
       def call(template)
         'xml = ::Nokogiri::XML::Builder.new { |xml|' +
           template.source +
-          "}.to_xml :indent => 2, :encoding => 'UTF-8',
-            :save_with => Nokogiri::XML::Node::SaveOptions::NO_DECLARATION | Nokogiri::XML::Node::SaveOptions::FORMAT;"
+          "}.to_xml(:indent => 2, :encoding => 'UTF-8',
+            :save_with => Nokogiri::XML::Node::SaveOptions::NO_DECLARATION | Nokogiri::XML::Node::SaveOptions::FORMAT).gsub('&#13;', '\r')"
       end
     end
   end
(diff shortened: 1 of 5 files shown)


dmarcoux (approved):
It would be nice to have a spec covering this.

eduardoj (approved):

krauselukas:
@dmarcoux I added a little test, its only checking at one point, but I think other tests at least covering that it doesn't brake things and the new one checks for a regression

dmarcoux (approved):

stoecker:
I would have simply deleted it as it serves no purpose...

eduardoj:
> I would have simply deleted it as it serves no purpose...

@stoecker , could you elaborate? With 'it', are you referring to carriage return characters '\r'?

stoecker:
> @stoecker , could you elaborate? With 'it', are you referring to carriage return characters '\r'?

Yes. The XML is perfectly valid without \r and dropping them saves space.

eduardoj:
> Yes. The XML is perfectly valid without \r and dropping them saves space.

I agree that the XML is perfectly valid without \r, and dropping the will save also space. But dropping the \r characters will make what now is shown like this:

Screenshot from 2020-10-14 14-10-37

... to be shown like this:

Screenshot from 2020-10-14 14-14-15

So, preserving the \r characters has sense for some descriptions. Am I missing something else?

stoecker:
Do you mix \r and \n? \r serves no purpose, the \n is used for text flow.

PR #10295, opened 2020-10-13
User profile page redesign

saraycp:
This time we try to improve the user profile page adjusting positions, margins/paddings and sizes of the different elements. Paying special attention to the avatar and displaying related information together.

The Contribution Graph is now embedded in a tab.
The same happens with the list(s) of projects/packages where the user is involved. How we display this list(s) is going to be changed, but not in this PR.

Screenshot_2020-10-13 Profile of Admin(1)

**Before:**

user_profile_before

**After:**

user_profile_redesign_info_2

The animation shows the user biography, however, the field will be really added to this PR as soon as [this one](https://github.com/openSUSE/open-build-service/pull/10292) is merged.

TODO:

- [x] Add the real `biography` field to the view, allowing markdown on it.
- [x] Merge this PR with the one that redesigns the `Involved Projects/Packages` tab content.

- Make fields like `biography`, `name` and `email` editable inline when that's possible. In another PR.

**NOTE:** This PR requires to enable the feature flag in production and IBS. Running `Flipper.enable(:user_profile_redesign, :beta) in Rails console.`

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -7,6 +7,7 @@ class User < ApplicationRecord
   # Keep in sync with states defined in db/schema.rb
   STATES = ['unconfirmed', 'confirmed', 'locked', 'deleted', 'subaccount'].freeze
   NOBODY_LOGIN = '_nobody_'.freeze
+  MAX_BIOGRAPHY_LENGTH_ALLOWED = 250
 
   # disable validations because there can be users which don't have a bcrypt
   # password yet. this is for backwards compatibility
@@ -102,6 +103,7 @@ class User < ApplicationRecord
   validate :password_validation
   validates :password, length: { minimum: 6, maximum: ActiveModel::SecurePassword::MAX_PASSWORD_LENGTH_ALLOWED }, allow_nil: true
   validates :password, confirmation: true, allow_blank: true
+  validates :biography, length: { maximum: MAX_BIOGRAPHY_LENGTH_ALLOWED }
 
   before_save :send_metric_for_beta_change, if: :in_beta_changed?
   after_create :create_home_project, :track_create
--- a/src/api/app/views/webui/user/_edit_dialog.html.haml
+++ b/src/api/app/views/webui/user/_edit_dialog.html.haml
@@ -12,5 +12,11 @@
           .form-group
             = f.label(:email, 'Email:')
             = f.text_field(:email, required: true, email: true, class: 'form-control')
+          - if feature_enabled?(:user_profile_redesign)
+            .form-group
+              = f.label(:biography) do
+                Biography:
+                %small.form-text Max. #{User::MAX_BIOGRAPHY_LENGTH_ALLOWED} characters.
+              = f.text_area(:biography, rows: 6, class: 'form-control', maxlength: User::MAX_BIOGRAPHY_LENGTH_ALLOWED)
         .modal-footer
           = render partial: 'webui/shared/dialog_action_buttons', locals: { submit_tag_text: 'Update' }
--- a/src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
+++ b/src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
@@ -1,0 +1,66 @@
+.card.mb-3
+  .card-body
+    .row.mb-3.mb-md-0
+      .col-4.col-sm-2.col-md-12.text-center
+        = image_tag_for(user, size: 200)
+      .col-8.col-sm-10.col-md-12.pl-0.p-md-3
+        %h4#home-realname
+          = user.realname
+        %h5.text-muted#home-login
+          = user.login
+
+    - role_titles.each do |title|
+      %span.badge.badge-secondary
+        = title.upcase
+
+    .mt-4
+      - if User.session
+        = mail_to(user.email, title: "Send mail to #{user.name}", class: 'd-block') do
+          %i.fas.fa-envelope.mr-1
+          = user.email
+
+      - if policy(user).see_rss?
+        = link_to(user_rss_notifications_path(token: user.rss_token.string, format: 'rss'),
+                                              title: 'RSS Feed for Notifications', class: 'd-block') do
+          %i.fas.fa-rss.mr-1
+          RSS for Notifications
+
+    - if groups.any?
+      .h5.mt-4.mb-0 Member of the #{'group'.pluralize(groups.size)}
+      %ul.list-group.list-group-flush
+        - groups.each do |group|
+          %li.list-group-item.d-flex.justify-content-between.align-items-center.p-2
+            = link_to(group.title, group_show_path(group))
+            %span.badge.badge-primary
+              #{group.tasks} #{'task'.pluralize(group.tasks)}
+
+    - if is_user
+      .mt-4
+        = form_tag(user_path, id: 'beta-form', method: :patch) do
+          .custom-control.custom-switch
+            = hidden_field_tag 'user[in_beta]', false
+            = check_box_tag('user[in_beta]', !user.in_beta, user.in_beta, class: 'custom-control-input', id: 'beta-switch')
+            = label_tag 'Public Beta Program', nil, class: 'custom-control-label', for: 'beta-switch'
+            = hidden_field_tag('user[login]', user.login)
+            %i.fa.fa-question-circle.text-info{ data: { placement: 'top', toggle: 'popover', html: 'true',
+                                                        content: 'Join the <strong>beta program</strong> to try the latest ' + |
+                                                                 'features we develop and give us feedback on them before they go live.' } } |
+
+      .mt-4
+        - if configuration.accounts_editable?(user)
+          = render partial: 'webui/user/edit_account', locals: { account_edit_link: account_edit_link }
+        - if configuration.passwords_changable?(user)
+          = render partial: 'webui/user/change_password'
+        - if feature_enabled?(:responsive_ux)
+          - content_for :actions do
+            %li.nav-item
+              = link_to(my_subscriptions_path, class: 'nav-link', title: 'Change Your Notifications') do
+                %i.fas.fa-lg.mr-2.fa-bell
+                %span.nav-item-name Change Your Notifications
+        - else
+          = link_to(my_subscriptions_path, class: 'd-block') do
+            %i.fas.fa-bell
+            Change Your Notifications
+
+:javascript
+  switchBeta();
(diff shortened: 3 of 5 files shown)


danidoni:
My personal feeling about the profile on bigger screens is that the information is too narrow. Now it is maybe at 1/4 of the available width? It feels compressed. I would allow maybe 1/3 of the available width.

saraycp:
> My personal feeling about the profile on bigger screens is that the information is too narrow. Now it is maybe at 1/4 of the available width? It feels compressed. I would allow maybe 1/3 of the available width.

Taking this into account, I adjusted the distribution of the column in different breakpoints, so now we don't have a narrow left column anymore.

Thanks @danidoni also for making me aware of the feature tests problems. Due to a missing condition, the actions were displayed twice when `responsive_ux` feature flag was enabled but `user_profile_redesign` not.

src/api/app/policies/user_policy.rb
```diff
@@ -6,4 +6,8 @@ def update?
   def show?
     user.can_modify_user?(record)
   end
+
+  def see_rss?
```
dmarcoux, line 10: This policy method could be replaced by `Token::RssPolicy#create?` since a user can only see their RSS feed if they are allowed to create a token for it.

https://github.com/openSUSE/open-build-service/blob/75a5c61d7d2eb870d35bbf721aece09b0b803da4/src/api/app/policies/token/rss_policy.rb#L7-L9

src/api/app/policies/user_policy.rb
```diff
@@ -6,4 +6,8 @@ def update?
...
+    user.login == record.login && record.rss_token
```
OleksandrOrlov, line 11: @saraycp Am I correctly understood, that it may be changed like this?

```suggestion
    Token::RssPolicy.new(user, record).create?
```

OleksandrOrlov, line 11: @dmarcoux @saraycp Am I correctly understood it may be like this?

```suggestion
    Token::RssPolicy.new(user, record.rss_token).create? if record.rss_token
```

src/api/app/policies/user_policy.rb
saraycp: I'm not really sure if you ask for getting rid of  `UserPolicy#see_rss?` or refactoring it.
After some discussion with @OleksandrOrlov and @danidoni , we propose a refactoring on the method, which is now called `rss_access?` in order to avoid including policy's code in the view. Something that is already done in some other policies.

Let us know if you'd propose something different.

src/api/app/models/user.rb
```diff
@@ -7,6 +7,7 @@ class User < ApplicationRecord
   # Keep in sync with states defined in db/schema.rb
   STATES = ['unconfirmed', 'confirmed', 'locked', 'deleted', 'subaccount'].freeze
   NOBODY_LOGIN = '_nobody_'.freeze
+  MAX_BIOGRAPHY_LENGTH_ALLOWED = 200
```
hennevogel, line 10: :thought_balloon:  If you limit me like this you need to A) tell me and B) help me in the interface to keep the limit

hennevogel: 200 is also a *very* harsh limit for markdown formatted text

saraycp: true!

krauselukas: this could be a start https://www.w3schools.com/tags/att_input_maxlength.asp

krauselukas (commented):
Besides whats already mentioned, no additional remarks, good stuff @saraycp

src/api/app/models/user.rb
saraycp: > 200 is also a _very_ harsh limit for markdown formatted text

Is 250 or 255 enough for you? :innocent: :bouquet: I understand your point, but we used `string` type for the database field. In case we decide to extend it, we'll need another migration.

saraycp: As discussed offline, we are going to keep `biography`as string for now. We can extend it in the near future.

saraycp:
For the moment, we have extended the model validation to allow 250 instead of 200 characters (we can extend it even more in the future).

I also added text in the form to warn the user about the limit:
Screenshot_2020-10-16 Profile of Admin

src/api/app/views/webui/user/_edit_dialog.html.haml
```diff
@@ -12,5 +12,10 @@
           .form-group
             = f.label(:email, 'Email:')
             = f.text_field(:email, required: true, email: true, class: 'form-control')
+          .form-group
+            = f.label(:biography) do
+              Biography:
+              %small.form-text The biography can not contain more than 250 characters.
```
krauselukas, line 18:
```suggestion
              %small.form-text.text-muted The biography can not contain more than 250 characters.
```

krauselukas: It would make it a little less dominant

krauselukas: the description is a bit overkill in my opinion, its already clear that this is related to the biography, so maybe its enough to write something like "max. 250 characters" or "Must be 1-250 characters long"

src/api/app/policies/user_policy.rb
dmarcoux: I meant that this policy method is unneeded.

Thinking more about this, we don't even need a policy for this since the user will be able to see the `RSS for Notifications` link if they have a RSS token. In the view, instead of checking `policy(user).rss_access?`, we can check `user.rss_token`. If it's present, we show the `RSS for Notifications` link.

src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
```diff
@@ -0,0 +1,68 @@
...
+      %span.badge.badge-secondary
+        = title.upcase
+
+    .mt-4
+      - if User.session
+        = mail_to(user.email, title: "Send mail to #{user.name}", class: 'd-block') do
+          %i.fas.fa-envelope.mr-1
+          = user.email
+
+      - if policy(user).rss_access?
```
dmarcoux, line 24: To follow up on my [previous comment](https://github.com/openSUSE/open-build-service/pull/10295#discussion_r507557696), here's how we could check if we need to show the `RSS for Notifications` link or not.

```suggestion
      - if if user.rss_token && @is_displayed_user
```

src/api/app/views/webui/user/_edit_dialog.html.haml
saraycp: I like to see it `text-muted`. However, if remember correctly, in the past we had the same discussion when we added this kind of small texts in other forms. At that time we decided not to make it grey, so we don't have too many font sizes/colors/weights... 

The configuration page is a good example:

Screenshot_2020-10-19 Open Build Service

saraycp: Then the result could be like:

Screenshot_2020-10-19 Profile of scp

src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
saraycp: Thanks. Done!

src/api/app/views/webui/user/_edit_dialog.html.haml
krauselukas: @saraycp Ah okay, I forgot this, fine for me then! :+1:

src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
```diff
@@ -0,0 +1,68 @@
+.card.mb-3
+  .card-body
+    .row.mb-3.mb-md-0
+      .col-4.col-sm-2.col-md-12.text-center
+        = image_tag_for(user, size: 100)
```
krauselukas, line 5: @saraycp whats the reason behind the small image tag? in the previous info box it was 500px

krauselukas: just talking about the big viewports

saraycp: Now that you say it... maybe it was a drastic change. So I've updated to 200. You can compare:

Size 100:

Screenshot_2020-10-19 Profile of Admin(5)

Size 200:

Screenshot_2020-10-19 Profile of Admin(7)

src/api/app/policies/user_policy.rb
DavidKang: IMO, I think the policy is needed, because other users could visit your profile page.

src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
krauselukas: better :+1:

saraycp:
**NOTE:** This PR requires to enable the feature flag in production and IBS. Running `Flipper.enable(:user_profile_redesign, :beta) in Rails console.`

src/api/app/views/webui/user/_edit_dialog.html.haml
```diff
@@ -12,5 +12,11 @@
           .form-group
             = f.label(:email, 'Email:')
             = f.text_field(:email, required: true, email: true, class: 'form-control')
+          - if feature_enabled?(:user_profile_redesign)
+            .form-group
+              = f.label(:biography) do
+                Biography:
+                %small.form-text Max. 250 characters.
```
dmarcoux, line 19:
```suggestion
                %small.form-text Max. #{User::MAX_BIOGRAPHY_LENGTH_ALLOWED} characters.
```

src/api/app/views/webui/user/_edit_dialog.html.haml
```diff
@@ -12,5 +12,11 @@
...
+              = f.text_area(:biography, rows: 6, class: 'form-control', maxlength: 250)
```
dmarcoux, line 20:
```suggestion
              = f.text_area(:biography, rows: 6, class: 'form-control', maxlength: #{User::MAX_BIOGRAPHY_LENGTH_ALLOWED})
```

src/api/app/views/webui/user/_edit_dialog.html.haml
saraycp: Done

src/api/app/views/webui/user/_edit_dialog.html.haml
saraycp: Done

vpereira (approved):

krauselukas (approved):

PR #10300, opened 2020-10-14
Move the Create Status Message back to the News box

danidoni:
The Create Status Message action is best placed near its context, the News box,
where the status messages live.

This is how it looked like before:

image

And this is how it looks like now:

image

### DIFF
--- a/src/api/app/views/webui/main/_status_messages.html.haml
+++ b/src/api/app/views/webui/main/_status_messages.html.haml
@@ -2,6 +2,9 @@
   .card-header.d-flex.justify-content-between
     %h5
       News
+      - if feature_enabled?(:responsive_ux) && policy(StatusMessage.new).create?
+        = link_to(new_status_message_path, title: 'Create Status Message') do
+          %i.fas.fa-xs.fa-plus-circle.text-primary
     .float-right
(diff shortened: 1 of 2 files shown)


danidoni:

src/api/app/views/webui/main/_status_messages.html.haml
```diff
@@ -7,7 +7,7 @@
         %i.fa.fa-rss
   .list-group
     = render partial: 'status_message', collection: status_messages
-    - if policy(StatusMessage.new).create? && !feature_enabled?(:responsive_ux)
+    - if policy(StatusMessage.new).create? # && !feature_enabled?(:responsive_ux)
```
dmarcoux, line 10: Everything after `#` should be removed, right?

danidoni: Oh yeah, i forgot that :facepalm:

dmarcoux:
To be consistent with all the `Create ABC` actions for lists, so just like other PRs, we could go with this:

consistent-action

danidoni:
> To be consistent with all the `Create (...)` actions for lists, so just like other PRs, we could go with this:
> 
> consistent-action

Agree! Fixed

dmarcoux (approved):

PR #10303, opened 2020-10-14
Use history icon for 'Job History' link

dmarcoux:
This is to avoid using the same `fa-list` icon as the `All Projects` link in the `Places` links.

Before:
before

Now:
now


dmarcoux:
[Revisit the icons for the left navigation](https://trello.com/c/H4yvAsgS/1550-revisit-the-icons-for-the-left-navigation)

danidoni (approved):

PR #10304, opened 2020-10-14
Prevent duplication for involved projects/packages

saraycp:
If the Contribution Graph is enabled, the 'Involved Projects/Packages' list can be displayed inside a simple card or inside a tab according to the viewport size.

Instead of rendering the same partial ('_involvement.html.haml') twice and using CSS to show/hide one or the other according to the viewport size, now we inject the content of the partial in the right place using JavaScript.

Preventing duplication is needed because it brings conflicts when selecting other elements in the view by JavaScript.

Follow-up of: #10271

### DIFF
--- a/src/api/app/assets/javascripts/webui/user_profile.js
+++ b/src/api/app/assets/javascripts/webui/user_profile.js
@@ -1,0 +1,7 @@
+function moveInvolvementToContainer() { // jshint ignore:line
+  var container = $('#involvement-and-activity');
+  if ($('#involvement-and-activity > .tab-content:visible').length > 0)
+    container = $('.tab-pane#involved-projects-and-packages');
+
+  container.prepend($('#involvement'));
+}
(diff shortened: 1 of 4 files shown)


src/api/app/assets/javascripts/webui/user_profile.js
```diff
@@ -0,0 +1,7 @@
+function placeInvolvement() { // jshint ignore:line
```
dmarcoux, line 1: I know naming things is hard... I understand what this function does, but only after looking at its code.

Do you find something like `moveInvolvementToContainer` clearer?

src/api/app/assets/javascripts/webui/user_profile.js
```diff
@@ -0,0 +1,7 @@
...
+  var parent = $('#involvement-and-activity');
```
dmarcoux, line 2: `parent` is the container of `#involvement`, right? If so, I would name this variable `container`.

src/api/app/assets/javascripts/webui/user_profile.js
eduardoj: Sure. Renamed to `moveInvolvementToContainer`.

dmarcoux (approved):

PR #10307, opened 2020-10-15
Move Create User action next to the heading of the Manage Users page

danidoni:
The action is now on the right of the Manage User's heading

This is how it looked like before:
image

image

### DIFF
--- a/src/api/app/views/webui/users/index.html.haml
+++ b/src/api/app/views/webui/users/index.html.haml
@@ -3,7 +3,11 @@
 .card.mb-3
   = render partial: 'webui/configuration/tabs'
   .card-body
-    %h3= @pagetitle
+    %h3
+      = @pagetitle
+      - if feature_enabled?(:responsive_ux)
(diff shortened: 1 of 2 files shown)


danidoni:

src/api/app/views/webui/users/index.html.haml
```diff
@@ -3,7 +3,11 @@
 .card.mb-3
   = render partial: 'webui/configuration/tabs'
   .card-body
-    %h3= @pagetitle
+    %h3
+      = @pagetitle
+      - if feature_enabled?(:responsive_ux)
+        = link_to(new_user_path(pagetitle: 'create user', submit_btn_text: 'create'), title: 'Create User', id: 'create-user') do
```
dmarcoux, line 9: The `id` isn't needed, so we could remove it.

```suggestion
        = link_to(new_user_path(pagetitle: 'create user', submit_btn_text: 'create'), title: 'Create User') do
```

danidoni: Done

PR #10312, opened 2020-10-16
Service name validation

vpereira:
Refactor Service model:
  - extract name validation to an external class
  - Add spec for the method `Service.verify_xml!`
  - Refresh specs

### DIFF
--- a/src/api/app/models/service/name_validator.rb
+++ b/src/api/app/models/service/name_validator.rb
@@ -1,0 +1,25 @@
+class Service
+  class NameValidator
+    attr_accessor :name
+
+    def initialize(name)
+      @name = name
+    end
+
+    def valid?
+      valid_name?
+    end
+
+    private
+
+    def valid_name?
+      return false unless name.is_a?(String)
+      return false if name.length > 200 || name.blank?
+      return false if /^[_.]/.match?(name)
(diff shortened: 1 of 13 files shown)


src/api/app/models/service/name_validator.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    private
+
+    def valid_name?
+      return false unless name.is_a?(String)
+      return false if name.length > 200 || name.blank?
+      return false if /^[_.]/.match?(name)
+      return false if /::/.match?(name)
+      return true if /\A\w[-+\w.:]*\z/.match?(name)
+
```
vpereira, line 21: The Validation being made is partially not necessary, since the libxml2 do already some validity checks

Examples:

```
=> "<services><service name=\"<>\" mode=\"trylocal\" /><service name=\"update_source\" mode=\"disabled::foo\" /></services>"
irb(main):007:0> Xmlhash.parse(xml)
=> nil

"<services><service name=Foo::Bar mode=\"trylocal\" /><service name=\"update_source\" mode=\"disabled::foo\" /></services>"
irb(main):011:0> Xmlhash.parse(xml)
=> nil
```

danidoni (approved):

Issue #4861, opened 2018-04-13
Searching for Packages/Projects or Owners in an out of the box OBS instance results in an exception


eduardoj:
Issue/Feature description
====================
When searching for Packages/Projects or Owners in an out of the box OBS instance results in an exception. The reason is that the `OBS:OwnerRootProject` flag is not set for any project, in a new setup.

Expected result
============
A warning message saying that `OBS:OwnerRootProject` flag for a project must be set.

How to Reproduce
===============
1. When I search for Packages/Projects or Owners in a newly setup OBS instance
2. an exception is thrown.

ChrisBr:
Why is it a bug? Is just missing configuration, no?

eduardoj:
Right. Removing "Bug" label.

PR #10313, opened 2020-10-16
User profile involved projects and packages revamp

eduardoj:
Before:

Peek 2020-10-16 10-48-User-Profile-Involved-old

After:

Peek 2020-10-19 17-13-User-Profile-Involved-2

## TODO
- [x] Write specs
- [ ] Reset filters / clear all button
- [x] Iterator improvements
- [x] Fix crashes
- [ ] Address UX Feedback
  - [ ] Filters should show their state
  - [ ] Complicated filters should indicate that their are active
  - [ ] Too many borders...

### DIFF
--- a/src/api/app/controllers/webui/users_controller.rb
+++ b/src/api/app/controllers/webui/users_controller.rb
@@ -11,12 +11,23 @@ class Webui::UsersController < Webui::WebuiController
   end
 
   def show
-    @iprojects = @displayed_user.involved_projects.pluck(:name, :title)
-    @ipackages = @displayed_user.involved_packages.joins(:project).pluck(:name, 'projects.name as pname')
-    @owned = @displayed_user.owned_packages
     @groups = @displayed_user.groups
     @role_titles = @displayed_user.roles.global.pluck(:title)
     @account_edit_link = CONFIG['proxy_auth_account_page']
+
+    if Flipper.enabled?(:user_profile_redesign, User.possibly_nobody)
+      attribute_type = AttribType.find_by_name!('OBS:OwnerRootProject')
+      @owner_root_project_exists = Project.find_by_attribute_type(attribute_type).exists?
+
+      filters = adjust_filters
+
+      @involved_items = @displayed_user.involved_items(filters)
+      @involved_items_as_owner = @displayed_user.involved_items_as_owner(filters) if @owner_root_project_exists
+    else
+      @iprojects = @displayed_user.involved_projects.pluck(:name, :title)
+      @ipackages = @displayed_user.involved_packages.joins(:project).pluck(:name, 'projects.name as pname')
+      @owned = @displayed_user.owned_packages
+    end
 
     return if CONFIG['contribution_graph'] == :off
 
@@ -135,6 +146,32 @@ class Webui::UsersController < Webui::WebuiController
   end
 
   private
+
+  def adjust_filters
+    filters = params.slice(:search_text, :involved_projects, :involved_packages,
+                           :role_maintainer, :role_bugowner, :role_reviewer, :role_downloader, :role_reader)
+    filters[:role_owner] = params[:role_owner] if @owner_root_project_exists && params.key?(:role_owner)
+
+    filters[:search_text] = filters[:search_text]&.strip
+    @filters = filters.dup
+
+    filter_keys = [:involved_projects, :involved_packages]
+    set_all_filters_if_none_is_set(filters, filter_keys)
+
+    filter_keys = [:role_maintainer, :role_bugowner, :role_reviewer, :role_downloader, :role_reader]
+    filter_keys << :role_owner if @owner_root_project_exists
+    set_all_filters_if_none_is_set(filters, filter_keys)
+
+    filters
+  end
+
+  def set_all_filters_if_none_is_set(filters, filter_keys)
+    return unless (filters.keys.map(&:to_sym) & filter_keys).empty?
+
+    filter_keys.each do |filter|
+      filters[filter] = 1
+    end
+  end
 
   def create_params
     {
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -677,6 +677,27 @@ class User < ApplicationRecord
     owned
   end
 
+  def involved_items_as_owner(filters = {})
+    OwnerSearch::Owned.new.for(self)
+                      .map { |owner| owned_item(owner, filters) }
+                      .reject(&:nil?)
+  end
+
+  def involved_items(filters = {})
+    involved_items = if filters[:role_owner]
+                       involved_items_as_owner(filters)
+                     else
+                       []
+                     end
+
+    roles_to_filter = filters.keys.select { |key| key != 'role_owner' && key =~ /^role_/ }.map { |key| Role.hashed[key.delete_prefix('role_')] }
+
+    involved_items.concat(related_items(Project, roles_to_filter, filters)) if filters.key?(:involved_projects)
+    involved_items.concat(related_items(Package, roles_to_filter, filters)) if filters.key?(:involved_packages)
+
+    involved_items.uniq
+  end
+
   # lists reviews involving this user
   def involved_reviews(search = nil)
     result = BsRequest.by_user_reviews(id).or(
@@ -945,6 +966,31 @@ class User < ApplicationRecord
                   else
                     UserBasicStrategy.new
                   end
+  end
+
+  def owned_item(owner, filters = {})
+    item = owned_project_or_package(owner, filters)
+
+    return if item.nil?
+    return if filters[:search_text].present? && !item.to_s.match?(Regexp.escape(filters[:search_text]))
+
+    item
+  end
+
+  def owned_project_or_package(owner, filters = {})
+    if owner.package.present?
+      owner.package if filters.key?(:involved_packages)
+    elsif filters.key?(:involved_projects)
+      owner.project
+    end
+  end
+
+  def related_items(project_or_package_class, roles_to_filter, filters = {})
+    related_items = project_or_package_class.related_to_user(id).where(relationships: { role_id: roles_to_filter }).or(
+      project_or_package_class.related_to_group(group_ids).where(relationships: { role_id: roles_to_filter })
+    )
+    related_items = related_items.where('name LIKE ?', "%#{filters[:search_text]}%") if filters[:search_text].present?
+    related_items
   end
 end
 
--- a/src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
+++ b/src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
@@ -1,0 +1,71 @@
+- roles = Role.local_roles
+#involvement
+  .card.sticky-top.border-0#involved-filters
+    .card-body
+      = form_with(model: displayed_user, method: :get, local: true, class: 'form-inline justify-content-between') do
+        .form-group
+          %input.form-control{ 'aria-label': 'Search', name: 'search_text', placeholder: 'Search', type: 'text', value: filters[:search_text] }
+        %span.dropdown.form-group#involved-project-package
+          %button.btn.btn-outline-secondary.dropdown-toggle.form-control{ data: { toggle: :dropdown }, type: :button }
+            %span.caret
+            Projects / Packages
+          .dropdown-menu.keep-open
+            - ['projects', 'packages'].each do |projects_packages|
+              - involved_html_id = "involved_#{projects_packages}"
+              .dropdown-item
+                .custom-control.custom-checkbox
+                  = check_box_tag involved_html_id, 1, filters[involved_html_id].present?, class: 'custom-control-input'
+                  %label.custom-control-label{ for: involved_html_id }
+                    = projects_packages.capitalize
+        %span.dropdown.form-group#involved-role
+          %button.btn.btn-outline-secondary.dropdown-toggle.form-control{ data: { toggle: :dropdown }, type: :button }
+            %span.caret
+            Role
+          .dropdown-menu.keep-open
+            - if owner_root_project_exists
+              .dropdown-item
+                .custom-control.custom-checkbox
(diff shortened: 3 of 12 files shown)


src/api/app/controllers/webui/users_controller.rb
```diff
@@ -11,13 +11,37 @@ def index
...
+        [:involved_projects, :involved_packages].each do |filter|
+          filters[filter] = 1
+        end
+      end
+      if !filters.key?(:role_owner) && !filters.key?(:role_maintainer) && !filters.key?(:role_bugowner) &&
+         !filters.key?(:role_reviewer) && !filters.key?(:role_downloader) && !filters.key?(:role_reader)
+        [:role_owner, :role_maintainer, :role_bugowner, :role_reviewer, :role_downloader, :role_reader].each do |filter|
+          filters[filter] = 1
+        end
+      end
```
danidoni, line 35: How about moving this code into a method? Something like `filters_for_involved_items` or else.

src/api/app/controllers/webui/users_controller.rb
```diff
@@ -11,13 +11,37 @@ def index
...
-    @owned = @displayed_user.owned_packages
     @groups = @displayed_user.groups
     @role_titles = @displayed_user.roles.global.pluck(:title)
     @account_edit_link = CONFIG['proxy_auth_account_page']
 
+    if Flipper.enabled?(:user_profile_redesign, User.possibly_nobody)
+      filters = params.slice(:search_text, :involved_projects, :involved_packages,
+                             :role_owner, :role_maintainer, :role_bugowner, :role_reviewer, :role_downloader, :role_reader)
+
+      filters[:search_text] = filters[:search_text]&.strip
```
vpereira, line 22: Please extract this logic to a testable object, either a method or a service object.

src/api/app/models/user.rb
```diff
@@ -675,6 +675,46 @@ def owned_packages
     owned
   end
 
+  def owned_item(owner, filters = {})
```
vpereira, line 678: could you please add specs to these new methods?

src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
```diff
@@ -0,0 +1,71 @@
+- roles = Role.local_roles
+#involvement
+  .card.sticky-top#involved-filters
+    .card-body
+      = form_with(model: displayed_user, method: :get, local: true, class: 'form-inline justify-content-between') do
+        .form-group
+          %input.form-control{ 'aria-label': 'Search', name: 'search_text', placeholder: 'Search', type: 'text', value: filters[:search_text] }
+        %span.dropdown.form-group#involved-project-package
+          %button.btn.btn-outline-secondary.dropdown-toggle.form-control{ data: { toggle: :dropdown }, type: :button }
+            %span.caret
```
vpereira, line 10: probably an option to reset the filter is necessary?

vpereira (changes requested):

src/api/app/controllers/webui/users_controller.rb
eduardoj: Codeclimate was complaining about it. I was moving the code already, to `adjust_filters`. It should be solved. Please have a look again.

src/api/app/controllers/webui/users_controller.rb
eduardoj: Codeclimate was complaining. I was working on it. I moved this logic to `adjust_filters`. Please take a look again.

src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
eduardoj: Do you mean like options `All` and `None` used in the filters in the project monitor page?

Screenshot from 2020-10-16 12-17-26

We decided not to include them in the initial mockup, because there were not so many items in the filter.

src/api/app/models/user.rb
```diff
@@ -675,6 +675,46 @@ def owned_packages
...
+    end
+  end
+
+  def involved_items_as_owner(filters = {})
+    items = []
+
+    OwnerSearch::Owned.new.for(self).each do |owner|
+      new_item = owned_item(owner, filters)
+      items << new_item unless new_item.nil? || (filters[:search_text].present? && !new_item.to_s.match?(Regexp.escape(filters[:search_text])))
+    end
```
danidoni, line 692: This iteration can be split into several small steps:

We can get the `new_items` with a `map`:
```ruby
OwnerSearch::Owned.new.for(self).map{ |o| owned_item(o, filters) }
```

Then we can `reject` nils with:
```ruby
.reject(&:nil?)
```

and then we can reject the second part of the condition with:

```ruby
.reject { |new_item| filters[:search_text].present? && !new_item.to_s.match?(Regexp.escape(filters[:search_text])) }
```

If I'm not wrong, if we don't have a `:search_text` filter, the second part of the condition will never happen, so we could skip this last step if we have no `filters[:search_text]`:

```ruby
items = OwnerSearch::Owned.new.for(self)
  .map{ |o| owned_item(o, filters) } # Transform to owned items
  .reject(&:nil?)
if filters[:search_text].present?
  # Filter by the search terms only if we have the filters set
  items.reject! { |new_item| !new_item.to_s.match?(Regexp.escape(filters[:search_text])) }
end
items 
```

src/api/app/models/user.rb
eduardoj: Sure. On it.

src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
vpereira: I'm talking about "Involved Package and Projects"

Once the filter is applied by the user then a button called ‘Reset Filters’ or ‘’Clear All’’ should be given next to the filters so that the user can quickly make changes. If we don't have it, we will force the user to either refresh the page (assuming that the filter will be erased) or user have to empty the search and click on "Filter results"

 Github does the same:

Screenshot_2020-10-16_12-35-48

src/api/app/models/user.rb
vpereira: I don't know if this snippet works, but looks like right and definitely more readable.

src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
eduardoj: Now I get it. A button called ‘Reset Filters’ or ‘’Clear All’’ was not included in the initial mockup. Probably something we forgot to add.

src/api/app/controllers/webui/users_controller.rb
```diff
@@ -136,6 +144,28 @@ def change_password
...
+    filters[:search_text] = filters[:search_text]&.strip
+    @filters = filters.dup
+
+    if !filters.key?(:involved_projects) && !filters.key?(:involved_packages)
+      [:involved_projects, :involved_packages].each do |filter|
+        filters[filter] = 1
+      end
+    end
+    if !filters.key?(:role_owner) && !filters.key?(:role_maintainer) && !filters.key?(:role_bugowner) &&
+       !filters.key?(:role_reviewer) && !filters.key?(:role_downloader) && !filters.key?(:role_reader)
```
danidoni, line 160: You can also replace this condition with:
```ruby
roles = %i(role_owner role_maintainer role_bugowner role_reviewer role_downloader role_reader)
if (filters.keys & roles).empty?
```

When there is no intersection between the `filters.keys` array, that means the filters.keys array does not include any of the roles. Then we set the filters.

WDYT?

src/api/app/models/user.rb
```diff
@@ -675,6 +675,30 @@ def owned_packages
     owned
   end
 
+  def involved_items_as_owner(filters = {})
+    items = []
+
+    OwnerSearch::Owned.new.for(self).each do |owner|
```
hennevogel, line 681: This crashes if you don't have the attribute

Screenshot from 2020-10-16 13-26-51

hennevogel:
Some UX feedback. 

## Filters should show their state

Peek 2020-10-16 13-31

## Complicated filters should indicate that their are active

Peek 2020-10-16 13-37

## [Too many borders...](https://twitter.com/steveschoger/status/897849211110273024)

Screenshot from 2020-10-16 13-40-38

src/api/app/models/user.rb
eduardoj: That [is just missing configuration, no?](https://github.com/openSUSE/open-build-service/issues/4861#issuecomment-417377553)

hennevogel: Nope. You can set this attribute on some project or you can not.

src/api/app/controllers/webui/users_controller.rb
eduardoj: That was nice. I applied your suggestion and moved the duplicated code into `set_all_filters_if_none_is_set`. Please have a look again.

src/api/app/models/user.rb
eduardoj: Then let's not display an option to filter by "Owner" in the roles filter, when there isn't any project with the `OBS:OwnerRootProject` attribute set. On it.

src/api/app/models/user.rb
eduardoj: That helped too. I used `map` as you proposed. I also moved the condittion of the `search_text` filter into the `owned_item` method. This prevents looping twice over the items. Please have a look again.

src/api/app/models/user.rb
eduardoj: It should be fixed now. Please have a look again.

src/api/app/models/user.rb
```diff
@@ -675,6 +675,25 @@ def owned_packages
...
+  def involved_items_as_owner(filters = {})
+    OwnerSearch::Owned.new.for(self)
+                      .map { |owner| owned_item(owner, filters) }
+                      .reject(&:nil?)
+  end
+
+  def involved_items(filters = {})
+    involved_items = []
+
+    involved_items = involved_items_as_owner(filters) if filters[:role_owner]
```
vpereira, line 687: you could do this code as:

```
involved_items = if filters[:role_owner]
                              involved_items_as_owner(filters)
                            else
                              []
                           end
```

src/api/app/models/user.rb
```diff
@@ -944,6 +963,27 @@ def hash_string(value)
                     UserBasicStrategy.new
                   end
   end
+
+  def owned_item(owner, filters = {})
+    item = if !owner.package.nil?
```
vpereira, line 968: `if owner.package.present?`

src/api/app/models/user.rb
eduardoj: Thanks. Applied.

src/api/app/models/user.rb
eduardoj: Also applied. :+1:

eduardoj:
@hennevogel, I removed a couple of borders. Please tell me if it is enough.

Screenshot from 2020-10-19 12-41-16

src/api/app/controllers/webui/users_controller.rb
```diff
@@ -136,6 +147,32 @@ def change_password
 
   private
 
+  def adjust_filters
```
danidoni, line 150: We're gonna try to move this to a model, so its easier to test

src/api/spec/controllers/webui/users_controller_spec.rb
```diff
@@ -84,6 +84,71 @@
...
+
+        context 'and we have selected a filter for a role' do
+          it 'shows only those items with the role' do
+            get :show, params: { login: user.login, role_owner: 1 }
+            expect(assigns(:involved_items)).to contain_exactly(some_project, some_package)
+            expect(assigns(:involved_items_as_owner)).to contain_exactly(some_project, some_package)
+          end
+        end
+
+        context 'and we have selected a filter for a bugowner' do
```
krauselukas, line 138: I guess the bugowner part is work in progress right? The rest regarding the specs looks good :+1:

krauselukas:
> @hennevogel, I removed a couple of borders. Please tell me if it is enough.
> 
> Screenshot from 2020-10-19 12-41-16

@eduardoj to me it looks much better now

src/api/app/controllers/webui/users_controller.rb
vpereira: service object or presenter. IMO, This information has nothing to do with a model.

eduardoj:
Regarding these two points from "Some UX feedback":
- Filters should show their state
- Complicated filters should indicate that their are active

Badges were added for each dropdown, with the number of the options that were selected in each case.

I updated the animation in the description. Please have a look, for the last time.

krauselukas (commented):
Looks good to me, maybe you can give the commit with the spec a better description

src/api/app/controllers/webui/users_controller.rb
```diff
@@ -11,13 +11,24 @@ def index
...
+    if Flipper.enabled?(:user_profile_redesign, User.possibly_nobody)
+      attribute_type = AttribType.find_by_name!('OBS:OwnerRootProject')
+      @owner_root_project_exists = Project.find_by_attribute_type(attribute_type).exists?
+
+      filters = adjust_filters
+
+      @involved_items = @displayed_user.involved_items(filters)
+      @involved_items_as_owner = @displayed_user.involved_items_as_owner(filters) if @owner_root_project_exists
+    else
+      @iprojects = @displayed_user.involved_projects.pluck(:name, :title)
```
dmarcoux, line 27: Is this a new product from Apple? If not, I would go with something more explicit, although longer, like `@involved_projects`. Same for `@ipackages` which would be `@involved_packages`.

hennevogel: This is for the old views, no need to change this. It'll be dropped once we roll this out.

dmarcoux: Oh, I totally overlooked that. :facepalm:

krauselukas (approved):
Since we planned on moving the methods that do not belong to the controller later on (https://github.com/openSUSE/open-build-service/issues/10324) this looks good to me :+1:

PR #10314, opened 2020-10-16
Refactor Service::NameValidator#valid_name?

vpereira:
Refactor `Service::NameValidator#valid_name?` method

### DIFF
--- a/src/api/app/models/service/name_validator.rb
+++ b/src/api/app/models/service/name_validator.rb
@@ -13,13 +13,17 @@ class Service
     private
 
     def valid_name?
-      return false unless name.is_a?(String)
-      return false if name.length > 200 || name.blank?
-      return false if /^[_.]/.match?(name)
-      return false if /::/.match?(name)
-      return true if /\A\w[-+\w.:]*\z/.match?(name)
+      return false unless name.present? || name.is_a?(String)
+      return false if name.length > 200
 
-      false
+      case name
+      when /^[_.]/, /::/
+        false
+      when /\A\w[-+\w.:]*\z/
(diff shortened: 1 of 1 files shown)


src/api/app/models/service/name_validator.rb
```diff
@@ -15,11 +15,15 @@ def valid?
     def valid_name?
       return false unless name.is_a?(String)
```
vpereira, line 16: maybe we should invert both lines? before we check if it's `name.blank? || name.size > 200` and then we check the class?

danidoni: I would replace this type check with a `respond_to?(:blank?)`

vpereira: > I would replace this type check with a `respond_to?(:blank?)`

I'm not sure, why?

```
[1] pry(main)> nil.blank?
=> true
```

danidoni: I'm talking about the `unless name.is_a?(String)`.
We use `duck typing` we don't really care if its really a string or not...
Do you agree?

vpereira: but everything answers to blank no?

```
[2] pry(main)> 1.respond_to?(:blank?)
=> true
```

vpereira: IMO @danidoni they check if it's string to avoid any XML injection attack.

danidoni: That makes sense

danidoni (approved):

PR #10320, opened 2020-10-19
Service in podman

M0ses:
* replace docker containers by podman
* Performance improvement in our test cases approx 20-30%
* 2 new packages system-user-obsrun/system-user-obs-servicerun


M0ses:
Not resolvable in SLE 12

```
nothing provides sysuser-shadow,
nothing provides sysuser-tools
```

should I do an extra PR for this change?

IMHO:

O:S:U should not care about old distributions

@adrianschroeter ?

M0ses:
@adrianschroeter - 

I think I fixed all the queried points (node_modules/SLE 12 compability)

Merging - as already discussed

PR #10321, opened 2020-10-19
Allow inline edition for user profile page

saraycp:
When the authentication is handled by a proxy, biography is the only attribute editable. Otherwise, biography, name and email can be edited.
    
**Without proxy:**

wo_proxy

**With proxy:**

with_proxt

Co-authored-by: Oleksandr Orlov <oorlov@suse.com>

DO NOT MERGE until #10295 is merged.

### DIFF
--- a/src/api/app/controllers/webui/users_controller.rb
+++ b/src/api/app/controllers/webui/users_controller.rb
@@ -1,7 +1,9 @@
 class Webui::UsersController < Webui::WebuiController
-  before_action :require_login, only: [:index, :edit, :destroy, :update, :change_password]
+  before_action :require_login, only: [:index, :edit, :destroy, :update, :change_password, :edit_account]
   before_action :require_admin, only: [:index, :edit, :destroy]
-  before_action :check_displayed_user, only: [:show, :edit, :update]
+  before_action :check_displayed_user, only: [:show, :edit, :update, :edit_account]
+  before_action :role_titles, only: [:show, :edit_account, :update]
+  before_action :account_edit_link, only: [:show, :edit_account, :update]
 
   def index
     respond_to do |format|
@@ -15,8 +17,6 @@ class Webui::UsersController < Webui::WebuiController
     @ipackages = @displayed_user.involved_packages.joins(:project).pluck(:name, 'projects.name as pname')
     @owned = @displayed_user.owned_packages
     @groups = @displayed_user.groups
-    @role_titles = @displayed_user.roles.global.pluck(:title)
-    @account_edit_link = CONFIG['proxy_auth_account_page']
 
     return if CONFIG['contribution_graph'] == :off
 
@@ -70,6 +70,12 @@ class Webui::UsersController < Webui::WebuiController
 
   def edit; end
 
+  def edit_account
+    respond_to do |format|
+      format.js
+    end
+  end
+
   def update
     unless User.admin_session?
       if User.session! != @displayed_user || !@configuration.accounts_editable?(@displayed_user)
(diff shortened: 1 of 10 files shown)


src/api/app/controllers/webui/users_controller.rb
```diff
@@ -79,24 +85,24 @@ def update
...
-      @displayed_user.update_globalroles(Role.global.where(id: params[:user][:role_ids])) unless params[:user][:role_ids].nil?
-    end
-
-    begin
-      @displayed_user.save!
-      flash[:success] = "User data for user '#{@displayed_user.login}' successfully updated."
-    rescue ActiveRecord::RecordInvalid => e
-      flash[:error] = "Couldn't update user: #{e.message}."
+    respond_to do |format|
+      if @displayed_user.save
```
dmarcoux, line 92: This block could be simplified like this:

```ruby
      format.any(:html, :js) do
        if @displayed_user.save
          flash[:success] = "User data for user '#{@displayed_user.login}' successfully updated."
        else
          flash[:error] = "Couldn't update user: #{@displayed_user.errors.full_messages.to_sentence}."
        end

        redirect_back(fallback_location: user_path(@displayed_user)) if request.format.symbol == :html
      end
```

The `flash[:error]` is the same for both cases, this is the only difference.

saraycp: I think your proposal is not exactly what we need, as for `js` format we need `flash.now`. I've refactored the code a bit, please have a look.

krauselukas:
@saray https://github.com/openSUSE/open-build-service/pull/10295 got merged, you can rebase this PR

src/api/app/controllers/webui/users_controller.rb
dmarcoux: Oh right, this is true!

src/api/app/policies/user_policy.rb
```diff
@@ -6,4 +6,8 @@ def update?
   def show?
     user.can_modify_user?(record)
   end
+
+  def rss_access?
```
dmarcoux, line 10: This should be removed as it's not needed.

saraycp: Thanks! Something went wrong when resolving conflicts in the rebase, because that was removed in the previous PR.
Anyway, I've reset the branch with master and cherry-picked the new commit. Everything should be ok now.

dmarcoux (approved):

src/api/app/views/webui/user/user_profile_redesign/_edit_account_form.html.haml
```diff
@@ -0,0 +1,28 @@
...
+  .form-group
+    = f.text_field(:realname, readonly: behind_proxy, placeholder: 'Name', class: 'form-control')
+  .form-group
+    = f.text_field(:login, readonly: true, class: 'form-control')
+  .form-group
+    - role_titles.each do |title|
+      %span.badge.badge-secondary
+        = title.upcase
+  .form-group
+    = f.text_area(:biography, rows: 6, placeholder: 'Biography', maxlength: 250, class: 'form-control')
```
dmarcoux, line 16:
```suggestion
    = f.text_area(:biography, rows: 6, placeholder: 'Biography', maxlength: User::MAX_BIOGRAPHY_LENGTH_ALLOWED, class: 'form-control')
```

src/api/app/views/webui/user/user_profile_redesign/_edit_account_form.html.haml
```diff
@@ -0,0 +1,28 @@
...
+    = f.label(:biography, class: 'd-block text-right') do
+      %small.form-text Max. 250 characters.
```
dmarcoux, line 18:
```suggestion
      %small.form-text Max. #{User::MAX_BIOGRAPHY_LENGTH_ALLOWED} characters.
```

src/api/app/views/webui/user/user_profile_redesign/_edit_account_form.html.haml
saraycp: Pushed!

src/api/app/views/webui/user/user_profile_redesign/_edit_account_form.html.haml
saraycp: Pushed!

krauselukas (approved):

PR #10327, opened 2020-10-20
Query autocomplete finders

vpereira:
Improve and extend the autocomplete finders that we have


src/api/app/queries/autocomplete_finder_base.rb
```diff
@@ -0,0 +1,7 @@
+class AutocompleteFinderBase
```
dmarcoux, line 1: We're starting to have more and more query objects, I think we could reorganize those `AutocompleteFinder` classes.

How about having the classes like this:
- src/api/app/queries/autocomplete_finder.rb
- src/api/app/queries/autocomplete_finder/package.rb
- src/api/app/queries/autocomplete_finder/project.rb

The parent class (`src/api/app/queries/autocomplete_finder.rb`) would be pretty much the same, except the name changes from `AutocompleteFinderBase` to `AutocompleteFinder`.

The children classes would be like `AutocompleteFinder::Package < AutocompleteFinder`.

src/api/app/queries/autocomplete_finder_base.rb
```diff
@@ -0,0 +1,7 @@
...
+  def initialize(relation, search_criteria, limit = 50)
```
dmarcoux, line 2: What do you think about `limit` being a keyword argument? It would look like this and make this a bit clearer when the `AutocompleteFinder` classes are initialized.

```suggestion
  def initialize(relation, search_criteria, limit: 50)
```

dmarcoux (approved):

PR #10336, opened 2020-10-22
Create system-user-obsrun and system-user-obsservicerun packages

andreas-schwab:


M0ses (approved):
Thanks Andreas for your contribution - I was investigating at exactly the the same problem ATM

PR #10337, opened 2020-10-22
Bugfix/user logging

hennevogel:
Several fixes for production logs that will make them more readable.


hennevogel:
Now if we get https://github.com/rails/rails/commit/a115a4cc796e55f2a86e36e92780fa74cd1c4c33 some day, we'll have clean logs...

PR #10338, opened 2020-10-22
Remove / to about routing malarky

hennevogel:
No need to route `/` and `/public/` to controllers that redirect to AboutController#index.
Just route to AboutController#index directly.


hennevogel:
If you're worried about the removed `/404` route, [don't be](https://obs-measure.opensuse.org/d/Gpp3B1Pik/measurements-per-action?orgId=1&var-Action=MainController%23notfound&var-beta=All&from=now-1y&to=now). 8 of those 11 was me testing it :smile_cat:

vpereira (approved):

PR #10339, opened 2020-10-23
Change Delayed::Worker.default_log_level to debug

hennevogel:
We want DJ logs to only appear as debug logs. And ActiveJob logs
as info logs.


hennevogel:
Sorry about the confusion...

PR #10345, opened 2020-10-26
Refactor Group model

vpereira:
Error Message if user not found:

Screenshot_2020-11-04_10-27-21

Existing user:

Screenshot_2020-11-04_10-27-52

### DIFF
--- a/src/api/app/models/group.rb
+++ b/src/api/app/models/group.rb
@@ -1,5 +1,3 @@
-require 'api_error'
-
 # The Group class represents a group record in the database and thus a group
 # in the ActiveRbac model. Groups are arranged in trees and have a title.
 # Groups have an arbitrary number of roles and users assigned to them.
@@ -34,7 +32,9 @@ class Group < ApplicationRecord
   alias_attribute :name, :title
 
   def self.find_by_title!(title)
-    find_by_title(title) || (raise NotFoundError, "Couldn't find Group '#{title}'")
+    find_by!(title: title)
+  rescue ActiveRecord::RecordNotFound => e
+    raise e, "Couldn't find Group '#{title}'", e.backtrace
   end
 
   def update_from_xml(xmlhash)
@@ -57,8 +57,9 @@ class Group < ApplicationRecord
         GroupMaintainer.create(user: user, group: self)
       end
     end
+
     cache.each do |login_id, _|
-      GroupMaintainer.where('user_id = ? AND group_id = ?', login_id, id).delete_all
+      delete_user(GroupMaintainer, login_id, id)
     end
 
     # update user list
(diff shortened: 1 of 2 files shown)


src/api/app/models/group.rb
```diff
@@ -34,7 +32,9 @@ class Group < ApplicationRecord
   alias_attribute :name, :title
 
   def self.find_by_title!(title)
-    find_by_title(title) || (raise NotFoundError, "Couldn't find Group '#{title}'")
+    find_by!(title: title)
+  rescue ActiveRecord::RecordNotFound => e
```
vpereira, line 36: any better way to change the message of the raised exception?

dmarcoux: I don't think so.

I've found this [StackOverflow thread](https://stackoverflow.com/questions/2823748/how-do-i-add-information-to-an-exception-message-in-ruby) which is providing more info.

src/api/app/models/group.rb
```diff
@@ -34,7 +32,9 @@ class Group < ApplicationRecord
...
+    raise e, "Couldn't find Group '#{title}'"
```
dmarcoux, line 37: To follow up on my previous comment,  we should also include the backtrace like this:

```suggestion
    raise e, "Couldn't find Group '#{title}'", e.backtrace
```

src/api/app/models/group.rb
dmarcoux: I don't think so.

I've found this [StackOverflow thread](https://stackoverflow.com/questions/2823748/how-do-i-add-information-to-an-exception-message-in-ruby) which is providing more info.

hennevogel: I take it you checked how the API is handling `ActiveRecord::RecordNotFound` right?

vpereira: @hennevogel 

```
rescue_from Backend::NotFoundError, ActiveRecord::RecordNotFound do |exception|
      render_error message: exception.message, status: 404, errorcode: 'not_found'
    end
```
Is there something missing? Any test that you want to see?

hennevogel: Nope, all is good. I was just not sure.

src/api/app/models/group.rb
```diff
@@ -80,16 +81,13 @@ def update_from_xml(xmlhash)
...
+      delete_user(GroupsUser, login_id, id)
     end
   end
 
   def add_user(user)
-    return if users.find_by_id(user.id) # avoid double creation
-
-    gu = GroupsUser.create(user: user, group: self)
-    gu.save!
+    return if users.find_by_id(user.id) || GroupsUser.create!(user: user, group: self)
```
dmarcoux, line 90: Is `return if` really needed here? It could be removed unless I'm missing something.

dmarcoux (approved):

src/api/app/models/group.rb
vpereira:
```
(base) vpereira@linux-qouy:~/open-build-service> dosc api '/group/xxxx'
WARNING: SSL certificate checks disabled. Connection is insecure!

Couldn't find Group 'xxxx'
```

src/api/app/models/group.rb
vpereira: done!

dmarcoux (approved):

PR #10346, opened 2020-10-26
[backend] Include the expanded buildconfig inside buildenv

### LINKED ISSUE #10020: Include prjconf in _buildenv
When working on reproducible builds for openSUSE, I try to verify that binaries published by OBS, really correspond to the source code using my `nachbau` script. For that, I grab the old binaries listed in _buildenv into a `deps` dir and do a `osc build -p deps` 
This regularly breaks when prjconf gets updated in certain ways.
E.g. atm I get `unresolvable: nothing provides libpython3_8-1_0 = 3.8.5 needed by python38-base, (got version 3.8.3-4.2)`
maybe because 
```
osc meta prjconf openSUSE:Factory
FileProvides: /usr/bin/python3 python38-base
```

I would like to have a way to build with the prjconf used for the old build.
This at least needs a reference to the prjconf revision in `_buildenv`
and a way to use that revision later.

It could also include a full copy of the old prjconf either in `_buildenv` or a separate `_prjconf` .
And if OBS/osc would allow builds with a custom prjconf, that might also be a nice addition for debugging and testing

I could use old revisions from
https://github.com/bmwiedemann/openSUSE/blob/master/packages/_/_project/_config
but so far, finding the right version is not trivial and using it for OBS-resolution is not possible.

eduardoj:
This allows in reproducible builds to build with the prjconf used for the old build.

Fixes #10020.

Co-authored-by: Sumit Jamgade <sjamgade@suse.de>

### DIFF
--- a/src/backend/BSXML.pm
+++ b/src/backend/BSXML.pm
@@ -537,6 +537,7 @@ our $buildinfo = [
 	'versrel',
 	'bcnt',
 	'release',
+        'config',
 	'debuginfo',
 	'constraintsmd5',
       [ 'prjconfconstraint' ],
--- a/src/backend/bs_worker
+++ b/src/backend/bs_worker
@@ -2624,7 +2624,7 @@ sub readbuildenv {
   $buildinfo->{'hasbuildenv'} = 1;
   # work around buildenv generation bug
   $_->{'name'} =~ s/\.rpm$// for @{$bi->{'bdep'} || []};
-  for (qw{error versrel bcnt release bdep}) {
+  for (qw{error versrel bcnt release bdep config}) {
     $buildinfo->{$_} = $bi->{$_} if defined $bi->{$_};
   }
   for ('versrel', 'bcnt', 'release') {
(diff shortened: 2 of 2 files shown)


src/backend/bs_worker
```diff
@@ -3128,7 +3133,13 @@ sub dobuild {
...
   }
   my $server = $buildinfo->{'srcserver'} || $srcserver;
-  my $config = BSRPC::rpc("$server/getconfig", undef, "project=$projid", "repository=$repoid", @configpath);
+  my $config;
+  if (defined $buildinfo->{'config'}) {
+    $config = $buildinfo->{'config'};
+  } else {
+    $config = BSRPC::rpc("$server/getconfig", undef, "project=$projid", "repository=$repoid", @configpath);
+  }
+  $buildinfo->{'outbuildinfo'}->{'config'} = $config; # the bs_worker code sets $outbuildinfo->{'config'} to the build config $config
```
adrianschroeter, line 3142: this is bs_worker code, you readbuildenv method instead?

eduardoj: I removed the comment. It created more confusion than clarity.

src/backend/BSXML.pm
```diff
@@ -513,6 +513,7 @@ our $sourceinfolist = [
 our $buildinfo = [
     'buildinfo' =>
 	'project',
+        'config',
```
mlschroe, line 516: Could you please move that down a bit so that it comes somewhere after the `[]` line? The build config tends to be quite big and contains newlines, so we want an element and not an attribute here.

Sorry for not catching this before. The rest is fine.

mlschroe (changes requested):

src/backend/BSXML.pm
eduardoj: I moved `config` down, just before `debuginfo`. I hope it is fine there.

mlschroe (approved):

PR #10353, opened 2020-10-28
Introduce group measurements

eduardoj:
Measure the number of groups ~every time a group is created or destroyed~, ~hourly~, each five minutes.

Tḧe number of groups changes not so often. However, we need to measure frequently to avoid graphic panels with "No data" in Grafana.

### DIFF
--- a/src/api/config/clock.rb
+++ b/src/api/config/clock.rb
@@ -27,6 +27,10 @@ module Clockwork
     SendEventEmailsJob.perform_later
   end
 
+  every(5.minutes, 'measurements') do
+    MeasurementsJob.perform_later
+  end
+
   every(49.minutes, 'rescale history') do
     StatusHistoryRescalerJob.perform_later
   end
--- a/src/api/app/jobs/measurements_job.rb
+++ b/src/api/app/jobs/measurements_job.rb
@@ -1,0 +1,7 @@
+class MeasurementsJob < ApplicationJob
+  queue_as :quick
+
+  def perform
+    RabbitmqBus.send_to_bus('metrics', "group count=#{Group.count}")
+  end
+end


danidoni (approved):

src/api/app/models/group.rb
```diff
@@ -201,6 +204,10 @@ def involved_projects_ids
     ### all projects where user is maintainer
     Relationship.projects.where(group_id: id, role_id: role.id).distinct.pluck(:project_id)
   end
+
+  def send_metric_for_group
+    RabbitmqBus.send_to_bus('metrics', "group count=#{Group.count}")
```
hennevogel, line 209: - Potentially this event will happen on year apart
- Potentially this event will not happen for a year

How can I know how many groups are there yesterday?

hennevogel (changes requested):

src/api/app/models/group.rb
eduardoj: Yep. With this approach, after introducing this change in the code, there will be no data until a group is created or destroyed. The advantage, however, is that there is no need to query periodically for data that has not been changed. To prevent this from happening, could we introduce kind of a "data migration" to store the first measurement, after this code is deployed? :thinking:

> How can I know how many groups are there yesterday?

Once the first measurement is stored, `fill(previous)` from Grafana should provide the value required.

hennevogel: > The advantage, however, is that there is no need to query periodically for data that has not been changed. 

How is this an advantage?

> Once the first measurement is stored, fill(previous) from Grafana should provide the value required.

Only if my time range includes the previous point.

vpereira: @hennevogel it wouldn't be so useful to every hour, have a cronjob pushing `Group.count` to influxdb. This for sure would fit "measures that don't make sense and we should drop to save space". 

Every measure should tell us a use case or be used to tell a story. The scenario that we saw as necessary was: "The number of groups are changing in an unexpected way, let's alert if it grows or shrinks abruptly". 

Similar scenario we imagined for users in beta.  Here, we want to know from time to time (i.e after a new feature is released under beta), if the user are moving from/to beta. That's would help us to understand if user are happy with beta, or moving away from that because of the new feature. To have cronjob pushing aggregated data about Group/Users, blindly sounds, IMO,  unnecessary

eduardoj: We still measure each time a group is created or deleted, but we introduced a monthly clockwork entry to ensure the following:
 1. we are going to have a first measurement as soon as this code is deployed for the first time, and
 1. there is at least one measurement per month.

What do you think about this solution?

vpereira: I still don't see what we want to monitor with that. Possible scenario: If an admin is spamming the application with new groups or deleting wrongly a huge lot of groups, as demolition squad,  i want to know about it ASAP and not after a month..

hennevogel (approved):

src/api/config/clock.rb
```diff
@@ -27,6 +27,10 @@ module Clockwork
     SendEventEmailsJob.perform_later
   end
 
+  every(5.minutes, 'measurements') do
+    MeasurementsJob.perform_later
```
hennevogel, line 31: Now you definitely need to add a timestamp to your measurement because there is no guarantee that the queue is empty.

hennevogel (changes requested):

src/api/config/clock.rb
eduardoj: Adding a timestamp to the measurement would not help to have measurements in time, to prevent possible delays caused by triggering jobs attached to not empty queues. However, I will try using a clockwork thread, instead of a delayed job.

hennevogel: Ah no, it'll just be whenever the queue runs. okay

hennevogel (approved):

PR #10354, opened 2020-10-28
Change measurement of the number of beta users

eduardoj:
Before we measured if a user left or joined the beta program, that didn't provide the total amount of beta users.

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -105,7 +105,6 @@ class User < ApplicationRecord
   validates :password, confirmation: true, allow_blank: true
   validates :biography, length: { maximum: MAX_BIOGRAPHY_LENGTH_ALLOWED }
 
-  before_save :send_metric_for_beta_change, if: :in_beta_changed?
   after_create :create_home_project, :track_create
 
   alias flipper_id id
@@ -898,11 +897,6 @@ class User < ApplicationRecord
     end
 
     save
-  end
-
-  def send_metric_for_beta_change
-    channel = (in_beta? ? 'joined_beta' : 'left_beta')
-    RabbitmqBus.send_to_bus('metrics', "user.#{channel} value=1")
   end
 
   def run_as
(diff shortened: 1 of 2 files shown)


src/api/app/models/user.rb
```diff
@@ -901,8 +901,7 @@ def update_login_values(env)
   end
 
   def send_metric_for_beta_change
-    channel = (in_beta? ? 'joined_beta' : 'left_beta')
-    RabbitmqBus.send_to_bus('metrics', "user.#{channel} value=1")
+    RabbitmqBus.send_to_bus('metrics', "user in_beta=#{User.in_beta.count},count=#{User.count}")
```
hennevogel, line 904: Your `in_beta` tag will have a very high cardinality if the number that `User.in_beta.count` returns is changing often.

A tag key should have not have many possible tag values.

danidoni (approved):

hennevogel (changes requested):

src/api/app/models/user.rb
eduardoj: As far as I understood, the syntax to add tags would be just after the metric:

```
metric_name,tag1=value_tag1,tag2=value_tag2 field1=value_field1,field2=value_field2
```

The idea was to store the value in a field, and not in a tag. So we don't have tags in this case, but two fields: `in_beta` and `count`. Is that correct?

hennevogel: https://docs.influxdata.com/influxdb/v1.8/write_protocols/line_protocol_tutorial/

hennevogel: 'metrics' is not the metric name. It's not part of what you send to influxdb. It's the bus you publish this on.

saraycp: Using real numbers, the line protocol could look like:
`user in_beta=5,count=100`

Where the metric name is `user`, there are no tags and the fields are `in_beta` and `count`. This way we don't have a problems with cardinality. But maybe you consider there is a better approach for this.

hennevogel: Ah yes, sorry I was wrong. :man_facepalming: 

Then you just have the same "problems" as in #10353

hennevogel (changes requested):

src/api/app/models/user.rb
eduardoj: We fixed this, performing the measurements every 5 minutes. Could you have a look again?

hennevogel (approved):

PR #10363, opened 2020-10-30
Measure the amount of users logged in in the last 5 minutes

danidoni:
This PR adds a new field in the users measurements to track the amount of logged in users in the last 5 minutes

### DIFF
--- a/src/api/app/models/notification.rb
+++ b/src/api/app/models/notification.rb
@@ -22,7 +22,7 @@ class Notification < ApplicationRecord
   end
 
   def any_user_in_group_active?
-    !subscriber.users.recently_seen.empty?
+    subscriber.users.seen_since(3.months.ago).any?
   end
 
   def template_name
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -71,6 +71,7 @@ class User < ApplicationRecord
   scope :with_email, -> { where.not(email: [nil, '']) }
 
   scope :recently_seen, -> { where('last_logged_in_at > ?', 3.months.ago) }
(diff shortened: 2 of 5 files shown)


danidoni:
[Have more AHM measurements about users and groups](https://trello.com/c/TSUAuQJk/1560-have-more-ahm-measurements-about-users-and-groups)

src/api/app/models/user.rb
```diff
@@ -71,6 +71,7 @@ class User < ApplicationRecord
   scope :with_email, -> { where.not(email: [nil, '']) }
 
   scope :recently_seen, -> { where('last_logged_in_at > ?', 3.months.ago) }
+  scope :seen_in_interval, ->(since_time, until_time) { where(last_logged_in_at: since_time..until_time) }
```
hennevogel, line 74: If you simplify this like

```suggestion
  scope :seen_since, ->(since) { where('last_logged_in_at > ?'), since }
```

You can replace `recently_seen` and drop the extra method in the job.

hennevogel (changes requested):

src/api/app/models/user.rb
danidoni: Yup!

src/api/app/models/notification.rb
```diff
@@ -22,7 +22,7 @@ def user_active?
   end
 
   def any_user_in_group_active?
-    !subscriber.users.recently_seen.empty?
+    !subscriber.users.seen_since(3.months.ago).empty?
```
vpereira, line 25: This is the same as  `subscriber.users.seen_since(3.months.ago).any?` , no?

danidoni: True, addressing...

vpereira (approved):

PR #10330, opened 2020-10-21
Refactor status history rescaler job

vpereira:
Break the StatusHistoryRescalerJob in two and do the statistics using ActiveRecord

PR #10365, opened 2020-10-30
Revert "Refactor status history rescaler job"

hennevogel:
Reverts openSUSE/open-build-service#10330


hennevogel:
This has lead to problems we are investigating.

vpereira:
Added a spec to make sure that the we are re-scaling the data following the "specification"

- For records older than a month, re-scale the data in 12 hours slots
- For records older than 7 days, re-scale the data in 6 hours slots
- For records older than 1 day, re-scale the data in 1 hour slots
- For records older than 2 hours, re-scale the data in 4 minutes slots
- Records newer than 2 hours, stay untouched.

vpereira (approved):

PR #10368, opened 2020-10-31
Improve spec for StatusHistoryRescalerJob

vpereira:
- Add more cases to make sure that the scale is working as expected

### DIFF
--- a/src/api/spec/jobs/status_history_rescaler_job_spec.rb
+++ b/src/api/spec/jobs/status_history_rescaler_job_spec.rb
@@ -3,32 +3,54 @@ require 'rails_helper'
 RSpec.describe StatusHistoryRescalerJob, type: :job do
   include ActiveJob::TestHelper
 
-  describe '#rescale' do
-    before do
-      Timecop.freeze(2010, 7, 12)
+  let(:status_histories) { StatusHistory.where(key: 'busy_x86_64').order(:time) }
 
-      now = Time.now.to_i - 2.days
-      StatusHistory.transaction do
-        1000.times do |i|
-          StatusHistory.create(time: now + i, key: 'idle_x86_64', value: i)
-        end
+  describe '#rescale' do
+    context 'newer than 2 hours records' do
+      before do
+        5.times { |i| StatusHistory.create(time: (Time.now.utc - i.seconds).to_i, key: 'busy_x86_64', value: 10 * i) }
       end
 
-      StatusHistory.create(time: Time.now.to_i, key: 'busy_x86_64', value: 100)
+      subject! { StatusHistoryRescalerJob.perform_now }
+
+      it 'keeps the StatusHistory with key = busy_x86_64' do
+        expect(status_histories.count).to eq(5)
+        expect(status_histories.first.value).to eq(40)
+        expect(status_histories.last.value).to eq(0.0)
+      end
     end
 
-    after do
-      Timecop.return
+    context 'newer than 9 days' do
+      before do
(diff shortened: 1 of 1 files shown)


src/api/spec/jobs/status_history_rescaler_job_spec.rb
```diff
@@ -3,32 +3,52 @@
...
+        2.times { |i| StatusHistory.create(time: 9.months.ago + i.minutes, key: 'busy_x86_64', value: 10 * i) }
+        2.times { |i| StatusHistory.create(time: 3.months.ago + i.minutes, key: 'busy_x86_64', value: 10 * i) }
+        2.times { |i| StatusHistory.create(time: 3.hours.ago + i.minutes, key: 'busy_x86_64', value: 10 * i) }
+      end
 
-    it { expect(StatusHistory.count).to eq(2) }
+      subject! { StatusHistoryRescalerJob.perform_now }
+
+      it 'reduces the records older than 1 month to 1' do
+        expect(StatusHistory.count).to eq(4)
```
vpereira, line 46: here it get's resampled as expected:
all records older than one month are going to be reduced.
 
We have 2 slots over 1 month with 2 records each. Total 4 records that get reduced to 2 records (Records within the same time slot, get merged together in one new record). So in the end we have 4 records.  2 reduced and 2 intact (because they were created in the last 24 hours)

src/api/spec/jobs/status_history_rescaler_job_spec.rb
```diff
@@ -3,32 +3,52 @@
...
+        expect(status_histories.first.value).to eq(40)
+        expect(status_histories.last.value).to eq(0.0)
+      end
     end
 
-    after do
-      Timecop.return
+    context 'newer than 9 days' do
+      before do
+        2.times { |i| StatusHistory.create(time: 9.days.ago + i.minutes, key: 'busy_x86_64', value: 10 * i) }
```
vpereira, line 25: Here just 2 records are older than 7 days, so they get reduced to 1 (both are merged together).

src/api/spec/jobs/status_history_rescaler_job_spec.rb
```diff
@@ -3,32 +3,52 @@
...
-      Timecop.freeze(2010, 7, 12)
-
-      now = Time.now.to_i - 2.days
-      StatusHistory.transaction do
-        1000.times do |i|
-          StatusHistory.create(time: now + i, key: 'idle_x86_64', value: i)
-        end
+    context 'newer than 2 days events' do
+      before do
+        5.times { |i| StatusHistory.create(time: (Time.now.utc - i.seconds).to_i, key: 'busy_x86_64', value: 10 * i) }
```
vpereira, line 11: Those records are kept intact. So no reduction is expected. Therefore the number of records doesn't change

src/api/spec/jobs/status_history_rescaler_job_spec.rb
```diff
@@ -3,32 +3,54 @@
...
+        2.times { |i| StatusHistory.create(time: 3.hours.ago + i.minutes, key: 'busy_x86_64', value: 10 * i) }
+      end
+
+      subject! { StatusHistoryRescalerJob.perform_now }
+
+      it 'reduces the records older than 7 days to 1' do
+        expect(StatusHistory.count).to eq(3)
+      end
+
+      it { expect(status_histories.first.value).to eq(5.0) }
```
vpereira, line 35: the two reduced records had the following values: 0 and 10, so the average is 5. Therefore the new created record has value = 5.

danidoni (approved):

PR #10371, opened 2020-11-02
Add statistics for backend Workers

vpereira:
Add statistics for the workers, similar with what we have in the frontend but those are for Grafana

TODO:

- [x] Create metrics for backend workers
- [x] Create new metrics job
- [x] Test locally with local grafana

Screenshot_2020-11-03_09-39-09

Tried locally using https://github.com/openSUSE/open-build-service/wiki/Site-Reliability#development-environment

Built locally some packages and could do it being updated on my local grafana instance

### DIFF
--- a/src/api/app/jobs/backend_measurements_job.rb
+++ b/src/api/app/jobs/backend_measurements_job.rb
@@ -1,0 +1,27 @@
+class BackendMeasurementsJob < ApplicationJob
+  queue_as :quick
+
+  def perform
+    return unless CONFIG['amqp_options']
+
+    RabbitmqBus.send_to_bus('metrics', "workers #{worker_status_fields}")
+
+    worker_jobs = WorkerStatusService::JobsStatisticsFetcher.call
+    worker_jobs[:waiting].each do |wj|
+      RabbitmqBus.send_to_bus('metrics', "waiting_jobs,#{jobs_fields(wj)}")
+    end
+    worker_jobs[:blocked].each do |wj|
+      RabbitmqBus.send_to_bus('metrics', "blocked_jobs,#{jobs_fields(wj)}")
+    end
+  end
+
+  private
+
+  def worker_status_fields
+    WorkerStatus.statistics.map { |k, v| "#{k}=#{v}" }.join(',')
+  end
+
+  def jobs_fields(job_line)
+    job_line.map { |k, v| "#{k}=#{v}" }.join(' ')
+  end
+end
--- a/src/api/app/models/worker_status.rb
+++ b/src/api/app/models/worker_status.rb
@@ -7,6 +7,10 @@ class WorkerStatus
       hidden_projects(ws)
     end
 
+    def statistics
+      WorkerStatusService::WorkerStatisticsFetcher.call
+    end
+
     private
 
     def hidden_projects(worker_status_root)
--- a/src/api/app/services/worker_status_service/jobs_statistics_fetcher.rb
+++ b/src/api/app/services/worker_status_service/jobs_statistics_fetcher.rb
@@ -1,0 +1,33 @@
+module WorkerStatusService
+  class JobsStatisticsFetcher
+    attr_reader :worker_status
+
+    class << self
+      def call
+        new.statistics
+      end
+    end
+
+    def initialize
+      @worker_status = Hash.from_xml(backend_worker_status)
+    end
+
+    def waiting
+      @worker_status['workerstatus']['waiting']
+    end
+
+    def blocked
+      @worker_status['workerstatus']['blocked']
+    end
+
+    def statistics
+      { waiting: waiting, blocked: blocked }
+    end
+
+    private
+
+    def backend_worker_status
+      Rails.cache.read('workerstatus') || Backend::Api::BuildResults::Worker.status
+    end
+  end
+end
--- a/src/api/app/services/worker_status_service/worker_statistics_fetcher.rb
+++ b/src/api/app/services/worker_status_service/worker_statistics_fetcher.rb
@@ -1,0 +1,37 @@
+module WorkerStatusService
+  class WorkerStatisticsFetcher
+    attr_reader :worker_status
+
+    class << self
+      def call
+        new.statistics
+      end
+    end
+
+    def initialize
+      @worker_status = Nokogiri::XML(backend_worker_status)
+    end
+
+    def statistics
+      { building: building, idle: idle, total: total }
+    end
+
+    def total
+      building + idle
+    end
+
+    def building
+      worker_status.search('//building').size
+    end
+
+    def idle
+      worker_status.search('//idle').size
+    end
+
+    private
+
+    def backend_worker_status
+      Rails.cache.read('workerstatus') || Backend::Api::BuildResults::Worker.status
+    end
+  end
+end
--- a/src/api/config/clock.rb
+++ b/src/api/config/clock.rb
@@ -27,6 +27,10 @@ module Clockwork
     SendEventEmailsJob.perform_later
   end
 
+  every(1.minute, 'backend_measurements') do
+    BackendMeasurementsJob.perform_later
+  end
+
   every(5.minutes, 'measurements') do
     MeasurementsJob.perform_later
   end
--- a/src/api/spec/models/worker_status_spec.rb
+++ b/src/api/spec/models/worker_status_spec.rb
@@ -124,4 +124,58 @@ RSpec.describe WorkerStatus do
       it { expect { subject }.to change { Architecture.available.count }.by(3) }
     end
   end
+
+  describe '#statistics' do
+    subject { WorkerStatus.statistics }
+
+    context 'when the xml response is empty' do
+      let(:xml_response) { '<workerstatus clients="0"></workerstatus>' }
+
+      it { expect(subject).not_to be_nil }
+      it { expect(subject).to include(:building, :idle, :total) }
+    end
+
+    context 'it generates a json with count for building and queued jobs' do
+      let(:xml_response) do
+        <<-HEREDOC
+        <workerstatus clients="285">
+          <idle workerid="build24/7" hostarch="x86_64"/>
+          <building repository="openSUSE_11.3_Update" arch="x86_64" project="home:enzokiel" package="android-sdk" starttime="1289838671" workerid="build03/1" hostarch="x86_64"/>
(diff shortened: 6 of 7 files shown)


src/api/app/models/worker_status.rb
```diff
@@ -7,6 +7,13 @@ def hidden
       hidden_projects(ws)
     end
 
+    def statistics
+      mydata = Rails.cache.read('workerstatus')
+      ws = Nokogiri::XML(mydata || Backend::Api::BuildResults::Worker.status)
+      { building: ws.search('//building').size, waiting: ws.search('//waiting').size,
+        blocked: ws.search('//blocked').size, idle: ws.search('//idle').size }
```
eduardoj, line 14: For `building` and `idle` is fine to count the ocurrences in the xml with `size`. For `waiting` and `blocked`, shouldn't we sum the `jobs` attributes?

hennevogel: Don't calculate it when recording. I would send it as `waiting` with the `arch` as tag and the `jobs` as field.

hennevogel: summarizing you can do in the visualization

vpereira: hm, i read it too late @hennevogel :rofl: 

If you want like that I would separate it in another PR and in this PR just keep the workers (idle x building) and then do another PR to handle jobs.

vpereira:
@eduardoj and @hennevogel I removed the waiting and blocked to handle in a different PR and different code

src/api/app/jobs/backend_measurements_job.rb
```diff
@@ -0,0 +1,17 @@
+BackendMeasurementsJob < ApplicationJob
+  queue_as :quick
+
+  def perform
+    return unless CONFIG['amqp_options']
+
+    RabbitmqBus.send_to_bus('metrics', "workers #{worker_status_fields}")
```
vpereira, line 7: I added the backend measurements metrics to a different job, mainly because:
- To avoid a huge job with many responsibilities
- In case we have some backend communication problem, it doesn't affect the other metrics
- Because the information about workers and jobs are high volatile and dynamic we should pull it as often as we can

src/api/config/clock.rb
```diff
@@ -27,6 +27,10 @@ module Clockwork
     SendEventEmailsJob.perform_later
   end
 
+  every(3.minutes, 'backend_measurements') do
+    BackendMeasurementsJob.perform_later
```
vpereira, line 31: 3 minutes is how long it takes to expire the backend cached data.

src/api/app/models/worker_status.rb
vpereira: @hennevogel could you please take a look again? I added the waiting and jobs fields.

src/api/app/services/worker_status_service/jobs_statistics_fetcher.rb
```diff
@@ -0,0 +1,33 @@
...
+    attr_reader :ws
+
+    class << self
+      def call
+        new.statistics
+      end
+    end
+
+    def initialize
+      @ws = Hash.from_xml(worker_status)
```
vpereira, line 12: https://apidock.com/rails/v6.0.0/Hash/from_xml/class maybe better than the Xmlhash gem?

src/api/app/jobs/backend_measurements_job.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    worker_jobs = WorkerStatusService::JobsStatisticsFetcher.call
+    worker_jobs[:waiting].each do |wj|
+      RabbitmqBus.send_to_bus('metrics', "waiting_jobs,#{jobs_fields(wj)}")
```
vpereira, line 11: I'm not sure if that should land at `metrics` or `backend_metrics`. What do you think @hennevogel ?

src/api/app/jobs/backend_measurements_job.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+    worker_jobs[:blocked].each do |wj|
+      RabbitmqBus.send_to_bus('metrics', "blocked_jobs,#{jobs_fields(wj)}")
```
vpereira, line 14: I'm not sure if that should land at metrics or backend_metrics. What do you think @hennevogel ?

src/api/app/jobs/backend_measurements_job.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  private
+
+  def worker_status_fields
+    WorkerStatus.statistics.map { |k, v| "#{k}=#{v}" }.join(',')
+  end
+
+  def jobs_fields(job_line)
+    job_line.map { |k, v| "#{k}=#{v}" }.join(' ')
```
vpereira, line 25: This line is generating the following InfluxDB line protocol: `arch=i586 jobs=5`, so arch is a tag, job is a field.

src/api/app/services/worker_status_service/jobs_statistics_fetcher.rb
dmarcoux: Nokogiri might also be an option here.

vpereira: Nokogiri is an option, however, to use Nokogiri, to transform the data manually to hash is, IMO, more error prone than using a Rails core method to do that

coolo: > https://apidock.com/rails/v6.0.0/Hash/from_xml/class maybe better than the Xmlhash gem?

define 'better' here. XmlMini is a pure ruby implementation and unless you plan to add the 3rd libxml using gem, it's for sure super slow: https://www.rubydoc.info/docs/rails/ActiveSupport/XmlMini

So what bothers you with xmlhash that you feel the need to pull the handbrake?

dmarcoux: Let's pull in the `libxml-ruby` gem then. I prefer to rely on a gem which is actively maintained and widely used.

We want to eventually remove the `xmlhash` gem anyway.

vpereira: > unless you plan to add the 3rd libxml using gem, it's for sure super slow: https://www.rubydoc.info/docs/rails/ActiveSupport/XmlMini

You mean, maintained 3rd library https://github.com/xml4r/libxml-ruby vs unmaintained 3rd library https://github.com/coolo/xmlhash? We all know the answer. But this PR isn't about it, but about sending backend metrics to our monitoring system. Do you want to help me to review this PR @coolo ? :)

src/api/app/jobs/backend_measurements_job.rb
hennevogel: We don't have a `backend_metrics` exchange on rabbit and I don't think we need to introduce one.

src/api/app/services/worker_status_service/jobs_statistics_fetcher.rb
hennevogel: Before anyone exchanges or introduces anything, one should know what impact it has on performance right? :)

src/api/app/services/worker_status_service/jobs_statistics_fetcher.rb
```diff
@@ -0,0 +1,33 @@
+module WorkerStatusService
+  class JobsStatisticsFetcher
```
hennevogel, line 2: :thought_balloon: I think this belongs into the backend library. It's getting data from the backend and molding it into a consumable form.

src/api/app/services/worker_status_service/worker_statistics_fetcher.rb
```diff
@@ -0,0 +1,37 @@
+module WorkerStatusService
```
hennevogel, line 1: :thought_balloon: same here, this is what we introduced the backend library for.

src/api/config/clock.rb
hennevogel: It's a bit more complicated I think. There is this tick in `clockwork`

```
every(30.seconds, 'status.refresh') do
  # this should be fast, so don't delay
  WorkerStatus.new.update_workerstatus_cache
end
```

That method *always* fetches the status from the backend, writes it into the cache and creates `StatusHistory`s from the data. That cache is then used throughout the app. So you practically you have a cache that refreshed every 30 seconds and expires (if no one refreshes) in 3 minutes.

src/api/app/services/worker_status_service/worker_statistics_fetcher.rb
```diff
@@ -0,0 +1,37 @@
...
+  class WorkerStatisticsFetcher
+    attr_reader :ws
```
hennevogel, line 3:
```suggestion
    attr_reader :worker_status
```

src/api/app/services/worker_status_service/jobs_statistics_fetcher.rb
```diff
@@ -0,0 +1,33 @@
...
+    attr_reader :ws
```
hennevogel, line 3:
```suggestion
    attr_reader :worker_status
```

hennevogel (changes requested):

src/api/app/services/worker_status_service/jobs_statistics_fetcher.rb
vpereira: we could do it later, lets experiment with this code without change the backend lib

src/api/config/clock.rb
vpereira: True, so we can make it 30 seconds or 1 minute if you want. Because we want to monitor it, lower delay among the points would be better.. are you ok with 1 minute here @hennevogel ?

src/api/app/services/worker_status_service/worker_statistics_fetcher.rb
vpereira: done

src/api/config/clock.rb
vpereira: I reduce it to `1.minute`

hennevogel: I would prefer if we could unify this to 

```
- refresh the cache every 60 seconds
- create StatusHistory every 60 seconds
- send metrics every 60 seconds
```

preferably in one job. Would make all this this way more clear and understandable IMHO.

vpereira:
Rework on the job are needed

PR #10374, opened 2020-11-03
[dist] in obs-server.spec prereq shadow for system-user packages

bugfinder:
Also move preinstall scripts to the same section in the
specfile where all other pre scripts are


M0ses (approved):
LGTM

PR #10381, opened 2020-11-03
Only show log in option if no user is logged in

### LINKED ISSUE #10326: "Log In" link is shown on Create User screen even though Admin is already logged in
Expected Result
============
Log in link should not be shown for the logged in user.

How to Reproduce
===============
1. Log in to the app as Admin;
2. Go to Configuration -> Manage Users -> Create User;
3. Focus on the Create User form.

Actual result:
"Log in" link is shown. Please, see screenshot below.
logged_in_user

jcronenberg:
Only show the "or Log In" option if no user is logged in

Fixes #10326 

Before Screenshot:
before
After Screenshots:
after
after_logged_out

### DIFF
--- a/src/api/app/views/webui/shared/_sign_up.html.haml
+++ b/src/api/app/views/webui/shared/_sign_up.html.haml
@@ -24,5 +24,6 @@
         class: 'form-control', required: true)
     = hidden_field_tag 'register', 'true'
     = submit_tag submit_btn_text, class: 'btn btn-primary'
-    or
-    = link_to('Log In', new_session_path)
+    - unless User.session
+      or
+      = link_to('Log In', new_session_path)


src/api/app/views/webui/shared/_sign_up.html.haml
```diff
@@ -24,5 +24,6 @@
         class: 'form-control', required: true)
     = hidden_field_tag 'register', 'true'
     = submit_tag submit_btn_text, class: 'btn btn-primary'
-    or
-    = link_to('Log In', new_session_path)
+    - if User.session.nil?
```
dmarcoux, line 27: We tend to always check only for `User.session`, so without the `.nil?`. `User.session` is either a `User` or `nil`. This line could be written like this:

```suggestion
    - unless User.session
```

dmarcoux:
Regarding the commit message, `Log In` is not an option, but it's a link. I suggest something like:

Display "Log In" link in sign_up partial unless a user is logged in

dmarcoux (approved):

PR #10385, opened 2020-11-04
[api] fix behaviour of not() operation in xpath query

bugfinder:
not($attribute) is basically an existens check as defined by w3

real test case is

/search/released/binary/id?match=[starts-with(repository/@project,"project")+and+not(@medium)]

but we do not have iso media in fixtures yet

patch created by adrian, applied in IBS since Mach 2019

### DIFF
--- a/src/api/lib/xpath_engine.rb
+++ b/src/api/lib/xpath_engine.rb
@@ -601,18 +601,24 @@ class XpathEngine
     # Note that this can result in bloated SQL statements, so some trust in the query optimization
     # capabilities of your DBMS is neeed :-)
 
-    if expr.first.in?([:child, :attribute])
-      # for incorrect writings of not(@name) as existens check
-      # we used to support it :/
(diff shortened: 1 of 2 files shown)


adrianschroeter (approved):

src/api/lib/xpath_engine.rb
```diff
@@ -601,18 +601,24 @@ def xpath_func_not(root, expr)
     # Note that this can result in bloated SQL statements, so some trust in the query optimization
     # capabilities of your DBMS is neeed :-)
 
-    if expr.first.in?([:child, :attribute])
-      # for incorrect writings of not(@name) as existens check
-      # we used to support it :/
-      @condition_values_needed = 2 if expr.first == :attribute
+    case expr.first
+    when :attribute
+      # existens check to an attribute
```
danidoni, line 606:
```suggestion
      # existence check to an attribute
```

danidoni (approved):

PR #9642, opened 2020-06-02
[api] allow to enforce maintenance request staging

adrianschroeter:
To create or to enforce staged maintenace submissions as part of
maintenance_incident request actions.

This can be used either on demand when creating the request by
user or can be enforced in the maintenance project by creating
OBS:EnforceIncidentRequestStaging attribute

Still waiting from feedback from maintenance team. So not merging before I hear from their experience, but to ensure that we do not forget about it.

PR #10386, opened 2020-11-04
[api] maintenance_release request action default reviewer change

bugfinder:
maintenance_release request actions takes default reviewers only from package
and update project. Not from GA project anymore.

patch from adrian, applied in IBS since July 2019


dmarcoux:
This is a duplicate of #9642

bugfinder:
ah duplicate, thanks. overlooked it ... closing here

bugfinder:
actually I picked the wrong patch ... will redo

PR #10387, opened 2020-11-04
[api] Alternative maintenance workflow

bugfinder:
- opt-in for channel branching
- patchinfo manual release support using timestamps

patch from adrian, applied in IBS since Sep 2019


adrianschroeter:
that one lacks the attribute creation ... will add it.

adrianschroeter (approved):

hennevogel (changes requested):
You migration file name does not match your migration class name...

PR #10389, opened 2020-11-04
Pass the association to the partial instead of call it two times

vpereira:
Master: 
	Avg (3 Requests): Duration: 1087.9ms | Allocations: 426595

Passing the association to the partial:
	Avg (3 Requests): Duration: 908.8ms | Allocations: 302319


eduardoj (approved):

vpereira:
[Speed up Webui::RequestController#show](https://trello.com/c/AwkFtKWd/1525-speed-up-webuirequestcontrollershow)

PR #10421, opened 2020-11-11
Introduce OpenAPI documentation with Swagger-UI

eduardoj:
This pull request introduces a proof of concept of documenting the current API with an OpenAPI 3.0 specification and Swagger-UI.

This pull request does not pretend to provide a replacement of the current documentation, found under https://build.opensuse.org/apidocs . This is more an attempt to find API endpoints where the OpenAPI 3.0 specification and Swagger-UI could not be useful for the current OBS API.

Different endpoints, or groups of them, will be added in different commits. The `/about` endpoint commit is provided as an example.

To test what has been introduced, you can have a look at the new documentation page here: 

1. Start the development environment
1. Go to http://localhost:3000/apidocs/swagger_ui

Screenshot from 2020-11-11 09-28-59


eduardoj:
> Review app will appear here: http://obs-reviewlab.opensuse.org/eduardoj-introduce_openapi_documentation

To see this from the review-app:
1. Go to https://obs-reviewlab.opensuse.org/eduardoj-introduce_openapi_documentation/apidocs/swagger_ui
2. Enter https://obs-reviewlab.opensuse.org/eduardoj-introduce_openapi_documentation/apidocs/OBS-v2.9.yaml in the text box at the top
3. Click "Explore"

krauselukas:
@eduardoj looks good :)

src/api/public/apidocs/OBS-v2.9.yaml
```diff
@@ -0,0 +1,114 @@
...
+      responses:
+        '200':
+          description: OK. The request has succeeded.
+          content:
+            application/xml:
+              schema:
+                $ref: '#/components/schemas/group'
+components:
+  schemas:
+    about:
```
lagartoflojo, line 57: As your schema grows, having all definitions in one file can get annoying.
You can also include external files, like so:

https://github.com/SUSE/happy-customer/blob/master/glue/public/api_docs/swagger/swagger.yaml#L141

src/api/public/apidocs/OBS-v2.9.yaml
```diff
@@ -0,0 +1,114 @@
+openapi: 3.0.0
+servers: []
+info:
+  description: Open Build Service API
```
lagartoflojo, line 4: You can use the description field to give an intro to the API.
Some basic questions can be answered here, like what's the purpose of the API? When would I use it? Does it need authentication? Are there any conventions? Is it versioned, and if so, how do I specify the version?

You can even use markdown for this, for example:
https://github.com/SUSE/happy-customer/blob/master/glue/public/api_docs/swagger/swagger.yaml#L4

gionaufal:
I don't know if you considered [rswag](https://github.com/rswag/rswag), but we use in one of our projects and it works fine. The problem I see for you is that you would either have to migrate your current tests to this rswag way, or write redundant tests to create the yamls. And sure, it is yet another dependency.

In our case we introduced at the start of the project, so we didn't have this problem of redundancy.

src/api/public/apidocs/OBS-v2.9.yaml
abitrolly: The link gives 404. Is it a private repo?

lagartoflojo: The repo is private, but the referenced Swagger file is public:
https://scc.suse.com/connect/v4/swagger.yaml

eduardoj:
Purpose of this pull request fulfilled. Thanks everybody for your comments!

abitrolly:
@eduardoj what do you think about generating the docs automatically with this tool https://github.com/yukihirop/r2-oas ?

dcermak:
> @eduardoj what do you think about generating the docs automatically with this tool https://github.com/yukihirop/r2-oas ?

Looks cool, just keep in mind that this will not be able to do much for routes that are passed through from the backend (e.g. `/build/`). That of course doesn't mean we shouldn't use it.

abitrolly:
@dcermak what is wrong with the `/build/` paths? I am not a Rails expert, but it looks like api and backend routes are both exposed in the same way https://github.com/openSUSE/open-build-service/blob/master/src/api/config/routes.rb

dcermak:
Anatoli Babenia <notifications@github.com> writes:

> @dcermak what is wrong with the `/build/` paths? I am not a Rails expert, but it looks like api and backend routes are both exposed in the same way https://github.com/openSUSE/open-build-service/blob/master/src/api/config/routes.rb
>

I am also not a Rails expert and must confess that I don't know how much
information this tool is able to extract, but it will not be able to
provide you with a lot of insight about the backend routes as these are
fully implemented in Perl and not in Ruby.

abitrolly:
@dcermak didn't notice that. Perl, indeed.

https://github.com/openSUSE/open-build-service/tree/master/src/backend

eduardoj:
Let's give it another try in #10571.

PR #10427, opened 2020-11-11
Show request action name in the page title if only one action exists

### LINKED ISSUE #1655: page title for requests should be more informative
For example, the HTML page `<title>` for https://build.opensuse.org/request/show/379272 is:

```
Request 379272 - openSUSE Build Service
```

but that really doesn't tell the user much at all.  A better title would be

```
Cloud:OpenStack:Master / openstack-neutron - Request 379272 - openSUSE Build Service
```

jcronenberg:
This changes the title that gets displayed for a request to include the action name if only one action exists for that request. Otherwise it only displays the request number.

Fixes #1655

To verify this feature:
1. Start sphinx `bundle exec rake ts:start`
2. Create development test data `bundle exec rake dev:development_testdata:create`
3. Open rails console `bundle exec rails c`
4. Create another action for a request e.g. like this `BsRequestAction.create(bs_request_id: 7, type: "submit", target_project: Project.first, target_package: Package.first, source_project: Project.second, source_package: Package.second)`
5. Start the rails server `bundle exec rails s -b 0.0.0.0`
6. Visit the requests page to see one with multiple actions (http://localhost:3000/request/show/7) and another without multiple actions (e.g. http://localhost:3000/request/show/8)

Before with single action:
before_single

After with single action:
after_single

After with multiple actions:
after_multiple

### DIFF
--- a/src/api/app/views/webui/request/show.html.haml
+++ b/src/api/app/views/webui/request/show.html.haml
@@ -2,6 +2,8 @@
   content_for(:meta_title, "Request #{@bs_request.number} (#{@bs_request.state})")
   content_for(:meta_description, @bs_request.description)
   content_for(:meta_image, gravatar_url(User.find_by_login(@bs_request.creator).email))
+  @pagetitle = "Request #{@bs_request.number}"
+  @pagetitle += ": #{@actions.first[:name]}" if @actions.count == 1
 
 - if @not_full_diff && User.session
   = render partial: 'webui/shared/truncated_diff_hint', locals: { path: request_show_path(number: @bs_request.number, full_diff: true) }
@@ -10,7 +12,6 @@
                                                      number: @bs_request.number,
                                                      diff_to_superseded: @diff_to_superseded,
                                                      superseding: @bs_request.superseding }
-- @pagetitle = "Request #{@bs_request.number}"
 .card.mb-3
   .card-header.d-flex.justify-content-between
     %h5


dmarcoux (approved):

src/api/app/views/webui/request/show.html.haml
```diff
@@ -10,7 +10,11 @@
                                                      number: @bs_request.number,
                                                      diff_to_superseded: @diff_to_superseded,
                                                      superseding: @bs_request.superseding }
-- @pagetitle = "Request #{@bs_request.number}"
+:ruby
```
krauselukas, line 13: @jcronenberg there is already a `:ruby` block in line 1, you could just move your code to this block to save the additional `:ruby`

krauselukas: besides that it looks good to me :+1:

dmarcoux: Oh true, good point :+1:

src/api/app/views/webui/request/show.html.haml
```diff
@@ -2,6 +2,10 @@
   content_for(:meta_title, "Request #{@bs_request.number} (#{@bs_request.state})")
   content_for(:meta_description, @bs_request.description)
   content_for(:meta_image, gravatar_url(User.find_by_login(@bs_request.creator).email))
+  @pagetitle = "Request #{@bs_request.number}"
+  if @actions.count == 1
```
vpereira, line 6: Prefer to use a clause guard when it is possible:

```@pagetitle += ": #{@actions.first[:name]}" if @actions.count == 1```

I think rubocop doesnt lint it on haml, but it would be a cop violation on another area of our codebase

vpereira (changes requested):

dmarcoux (approved):

krauselukas (approved):

PR #10428, opened 2020-11-11
Set transparent background to the sponsors' logos

saraycp:
We recently re-designed the sponsors' section (it is still part of the beta program).

Some logos had a white background since we moved them to a grey area they look ugly. So we have made transparent the logo background.

**Before:**

Screenshot_2020-11-11 openSUSE Build Service

**After:**

Screenshot_2020-11-11 Open Build Service(3)

This can't be tested in `review-app` since there we use the default icons, not the real ones.
To test it locally, you can drop/comment [these lines](https://github.com/openSUSE/open-build-service/blob/master/src/api/config/environments/development.rb#L72-L86) in the configuration file.


dmarcoux (approved):

dmarcoux:
The CircleCI workflow has to be restarted. The issues are totally unrelated to the changes. I cannot do it since it's not my fork.

krauselukas (approved):

Issue #820, opened 2014-11-21
Creating internal OBS user fails when logging via LDAP, role "User" is non-existent


tuminoid:
When logging in with LDAP for first time, OBS internal user is being created. In `application_controller.rb` in function `extract_ldap_user` is code:

```
        newuser.adminnote = "User created via LDAP"
        user_role = Role.find_by_title("User")
        newuser.roles << user_role
```

which fails with error: 

```
[f4b8330f-6a76-46dd-8daa-7bf55f37ea1a] [12725:4.60]   Role Load (0.4ms)  SELECT `roles`.* FROM `roles` WHERE `roles`.`title` = 'User' LIMIT 1
[f4b8330f-6a76-46dd-8daa-7bf55f37ea1a] [12725:4.61]    (0.2ms)  BEGIN
[f4b8330f-6a76-46dd-8daa-7bf55f37ea1a] [12725:4.61]    (0.1ms)  ROLLBACK
[f4b8330f-6a76-46dd-8daa-7bf55f37ea1a] [12725:4.61] Completed 500 Internal Server Error in 322ms
```

as there is no role called `User`:

```
MariaDB [api_production]> select * from roles;
+----+------------+-----------+--------+
| id | title      | parent_id | global |
+----+------------+-----------+--------+
|  1 | Admin      |      NULL |      1 |
|  2 | maintainer |      NULL |      0 |
|  3 | bugowner   |      NULL |      0 |
|  4 | reviewer   |      NULL |      0 |
|  5 | downloader |      NULL |      0 |
|  6 | reader     |      NULL |      0 |
+----+------------+-----------+--------+
6 rows in set (0.00 sec)
```

PR #10429, opened 2020-11-12
Create data migration to remove role "User"

dmarcoux:
At the commit https://github.com/openSUSE/open-build-service/commit/4204a68c21941b477b61dbfaa6ba6ac330d2a4d4 in the database seeds, the role "User" is associated to the "Admin" user, but not to the "nobody" user.
From this, I'm guessing that this role was used only on "real" users. We can also see that in the commit https://github.com/openSUSE/open-build-service/commit/e523b55785b0714fa658868fee629056b64c1bf7. Users are created, then are assigned the role "User".

In the commit https://github.com/openSUSE/open-build-service/commit/ae9bba8076117282e03bc47665d26c92e9e23c70, the role "User" was commented out in the database seeds.
Why? I don't know, but it might be related to the next paragraph.

In the commit https://github.com/openSUSE/open-build-service/commit/5ae5084e9e63469c9303d6f1baf1af0d7a9f72e7, the role "User" was completely removed from the database seeds.
According to Adrian in the [issue 820](https://github.com/openSUSE/open-build-service/issues/820), the role "User" was dropped.

With all this in mind, I believe we can safely remove the role "User" from our production database.

### DIFF
--- a/src/api/db/data/20201112083506_remove_role_user.rb
+++ b/src/api/db/data/20201112083506_remove_role_user.rb
@@ -1,0 +1,10 @@
+class RemoveRoleUser < ActiveRecord::Migration[6.0]
+  def up
+    Role.destroy_by(title: 'User')
+  end
+
+  def down
+    # While we can recreate the role, we cannot recreate associations between users and the role "User" (so records in RolesUser)
+    Role.create(title: 'User', global: true)
+  end
+end
--- a/src/api/app/models/role.rb
+++ b/src/api/app/models/role.rb
@@ -65,7 +65,7 @@ class Role < ApplicationRecord
   end
 
   def self.global_roles
-    ['Admin', 'User']
+    ['Admin']
   end
 
   def self.ids_with_permission(perm_string)
--- a/src/api/db/data_schema.rb
+++ b/src/api/db/data_schema.rb
@@ -1,2 +1,2 @@
 # encoding: UTF-8
-DataMigrate::Data.define(version: 20200528135241)
+DataMigrate::Data.define(version: 20201112083506)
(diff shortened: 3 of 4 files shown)


dmarcoux:
[What's going on with "User" role?](https://trello.com/c/5D2Xtpwn/1565-whats-going-on-with-user-role)

dmarcoux:
@adrianschroeter Since you were involved in the issue #820, what do you think of this?

src/api/db/data/20201112083506_remove_role_user.rb
```diff
@@ -0,0 +1,14 @@
+class RemoveRoleUser < ActiveRecord::Migration[6.0]
+  def up
+    role = Role.find_by(title: 'User')
```
vpereira, line 3: in Rails 6 you can simplify it with a `Role.destroy_by(title: 'User')` <3

vpereira (changes requested):

adrianschroeter:
That issue is IMHO independ.

The "User" role was never used, it was just a template in initial rails setup IIRC. So it should be good to go away.

src/api/db/data/20201112083506_remove_role_user.rb
dmarcoux: Cool, I didn't know about that method.

dmarcoux: Please have another look :)

krauselukas (approved):
also couldn't find any usage of it :crossed_fingers:

saraycp (approved):

vpereira (approved):

PR #10432, opened 2020-11-12
Refactor PackageController#find_last_req method

vpereira:
* Use guard clause instead of long condition blocks
* Extract complex ActiveRecord query to external method

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -831,26 +831,26 @@ class Webui::PackageController < Webui::WebuiController
   end
 
   def find_last_req
-    if @oproject && @opackage
-      last_req = BsRequestAction.where(target_project: @oproject,
-                                       target_package: @opackage,
-                                       source_project: @package.project,
-                                       source_package: @package.name).order(:bs_request_id).last
-      return unless last_req
+    return if @oproject.blank? || @opackage.blank?
 
-      last_req = last_req.bs_request
-      if last_req.state != :declined
-        return # ignore all !declined
-      end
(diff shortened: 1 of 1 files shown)


eduardoj (approved):

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -831,26 +831,24 @@ def check_package_name_for_new
...
-        when: last_req.updated_at,
-        comment: last_req.comment
-      }
-    end
-    nil
+    last_req = find_last_bs_request_not_declined
+
+    return if last_req.blank?
+
+    { id: last_req.number, decliner: last_req.commenter,
```
coolo, line 840: aehm, do you realize why it says decliner here? Hint: this is not about not_declined

vpereira: ähhm was meinst du @coolo 👼

coolo (changes requested):
The code checks the last request and only returns it if it was declined. Check where this is used:

```
    The previous request
    = link_to(request['id'], request_show_path(project, package, request['id']))
    was declined #{fuzzy_time_string(request['when'])} by
    = user_with_realname_and_icon(request['decliner'], short: true)
    with the following message:
    %pre= request['comment']
```

coolo (commented):
looks more correct now

coolo (commented):
Can't I just undo 'request changes?'

danidoni:
> Can't I just undo 'request changes?'

Can't you dismiss the review?

coolo:
thx

eduardoj (approved):

PR #10433, opened 2020-11-12
Add preview for the comment markdown

### LINKED ISSUE #10019: Implement a preview for comments fields
With the comments fields supporting markdown (MD), which seems to be not very standardised generally speaking, it would be very helpful if we could **preview our comments** before actually submitting them (like on GH). It is easy for a MD formatted text to come out looking very wrong due to a simple typo, for instance, and having a Preview box will catch the error before the text is submitted.

GH like "Preview" boxes alongside comments fields.

Actually writing a comment into a MD editor locally and then copying it into the comments field, but this can get tiring soon.

OleksandrOrlov:
It is easy for a MD formatted text to come out looking wrong due to a
typo.

The PR adds Preview tab for comment section, so that it is possible
to view the formatted with markdown comment before saving it.

Fixes #10019

To verify this feature

1. Log in to the application;
2. Go to any page, where comments section exists (e.g. select "Your Home Project" in the sidebar);
3. Type any comment with markdown in "Write" tab of comment section;
4. Select "Preview" Tab.

Expected result:
The comment is shown with rendered markdown. Please, see the GIF below on how it looks:
Peek 2020-11-12 15-10

### DIFF
--- a/src/api/app/assets/javascripts/webui/comment.js
+++ b/src/api/app/assets/javascripts/webui/comment.js
@@ -56,4 +56,26 @@ $(document).ready(function(){
     if (!closest.hasClass('collapsed'))
       closest.trigger('click');
   });
+
+  $('.comments-list').on('click', '.preview-comment-tab:not(.active)', function (e) {
+      var commentContainer = $(e.target).closest('[class*="-comment-form"]');
+      var commentBody = commentContainer.find('.comment-field').val();
+      var commentPreview = commentContainer.find('.comment-preview');
+      if (commentBody) {
+        $.ajax({
+          method: 'POST',
+          url: commentContainer.data('previewCommentUrl'),
+          dataType: 'json',
+          data: { 'comment[body]': commentBody },
+          success: function(data) {
+            commentPreview.html(data.markdown);
+          },
+          error: function() {
+            commentPreview.html('Error loading markdown preview');
+          }
+        });
+      } else {
+        commentPreview.html('Nothing to preview');
+      }
+  });
 });
--- a/src/api/app/assets/javascripts/webui/comment.js
+++ b/src/api/app/assets/javascripts/webui/comment.js
@@ -14,7 +14,17 @@ function updateCommentCounter(selector, count) {
   $(selector).text(parseInt(oldValue) + count);
 }
 
+function validateForm(e) {
+  var submitButton = $(e.target).closest('[class*="-comment-form"]').find('input[type="submit"]');
+  submitButton.prop('disabled', !$(e.target).val());
+}
+
 $(document).ready(function(){
+  // Disable submit button if textarea is empty and enable otherwise
+  $('.comments-list').on('keyup', '.comment-field', function(e) {
+    validateForm(e);
+  });
+
   $('.comments-list').on('keyup click', '.comment-field', function() {
     resizeTextarea(this);
   });
--- a/src/api/app/views/webui/comment/_reply.html.haml
+++ b/src/api/app/views/webui/comment/_reply.html.haml
@@ -18,8 +18,8 @@
 - if User.session
   .collapse{ id: "reply_form_of_#{comment.id}" }
     = render(partial: 'webui/comment/comment_field', locals: { form_method: :post,
-     comment: comment.children.new, commentable: commentable })
+     comment: comment.children.new, commentable: commentable, element_suffix: "reply_for_#{comment.id}" })
   - if policy(comment).update?
     .collapse{ id: "edit_form_of_#{comment.id}" }
       = render(partial: 'webui/comment/comment_field', locals: { form_method: :put,
-       comment: comment, commentable: commentable })
+       comment: comment, commentable: commentable, element_suffix: "edit_#{comment.id}" })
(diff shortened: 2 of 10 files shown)


vpereira:
Hi @OleksandrOrlov I have some considerations:

1) Going to a Requerst, I couldn't find a way to preview my comments
2) Going to Admin home's project and adding a comment to an existing user, breaks the preview:

Screenshot_2020-11-12_15-31-07

Screenshot_2020-11-12_15-33-30

OleksandrOrlov:
> Hi @OleksandrOrlov I have some considerations:
> 
> 1. Going to a Requerst, I couldn't find a way to preview my comments
> 2. Going to Admin home's project and adding a comment to an existing user, breaks the preview:
> 
> Screenshot_2020-11-12_15-31-07
> 
> Screenshot_2020-11-12_15-33-30

It is happened, because in [ajax call](https://github.com/openSUSE/open-build-service/blob/d2a2cd036dcae2908dd4a37ebd368640468b3dc3/src/api/app/assets/javascripts/webui/comment.js#L67) I'm using absolute path, but in the review app, which is appended directly to the host name. In review app we also using one more part in the path (oleksandrorlov-10019_preview_comments_markdown/), so the path is wrong... 

I guess I have to change the approach in the ajax call, but right now have no idea how to do this...

dmarcoux:
A way to have the correct route for the AJAX request would be to get this value from the path helpers, so `preview_comments_path` in this case. It needs to be passed to the frontend somehow, which can then use that route. This could be stored in a hidden field or a data attribute on an element.

OleksandrOrlov:
@vpereira @dmarcoux the wrong path is fixed. Now it works on review-app.

src/api/app/views/webui/comment/_reply.html.haml
```diff
@@ -18,8 +18,9 @@
...
   .collapse{ id: "reply_form_of_#{comment.id}" }
     = render(partial: 'webui/comment/comment_field', locals: { form_method: :post,
-     comment: comment.children.new, commentable: commentable })
+     comment: comment.children.new, commentable: commentable, element_suffix: "reply_for_#{comment.id}" })
   - if policy(comment).update?
     .collapse{ id: "edit_form_of_#{comment.id}" }
       = render(partial: 'webui/comment/comment_field', locals: { form_method: :put,
-       comment: comment, commentable: commentable })
+       comment: comment, commentable: commentable, element_suffix: "edit_#{comment.id}" })
+      %button.btn.cancel.btn-outline-gray-dark Cancel
```
dmarcoux, line 26: Is this expected, it doesn't look right:

cancel-button-missing-style

OleksandrOrlov: Oups! It will be in another PR and with correct style:) Sorry, removed the button. Thank you.

eduardoj:
On an empty comment, being in the "Write" tab and then clicking "Update", results in this warning:

Screenshot from 2020-11-12 17-04-12

But in the same empty comment, clicking in the "Preview" tab and then "Update", results in no warnings.

OleksandrOrlov:
> On an empty comment, being in the "Write" tab and then clicking "Update", results in this warning:
> 
> Screenshot from 2020-11-12 17-04-12
> 
> But in the same empty comment, clicking in the "Preview" tab and then "Update", results in no warnings.

@eduardoj Thanks for pointing me out. The tooltip is shown automatically by browser for the text area as 'required' property is set for it, but I cannot do the same for the "Preview" tab, as it is div.

I addressed the issue by disabling the button if the text area is empty and enabling it when at least one character is entered. Could you please review if the solution fits your expectations.

saraycp (changes requested):
Good job, @OleksandrOrlov !
My only concern here is about style. The tabs for Write and Preview look a bit weird for me. I think there are two reasons:
- It is difficult to see the tabs if they are on top of white background. We usually have them on top of gray.
- The content of the tab has no padding.

preview_comments

I also find weird that the area is not a box (no borders on the left and on the right). 
Screenshot_2020-11-13 A Time to Kill

src/api/app/assets/javascripts/webui/comment.js
```diff
@@ -14,7 +14,22 @@ function updateCommentCounter(selector, count) {
   $(selector).text(parseInt(oldValue) + count);
 }
 
+function validateForm(e) {
+  var submitButton = $(e.target).closest('[class*="-comment-form"]').find('input[type="submit"]');
+  if ($(e.target).val().length > 0) {
```
eduardoj, line 19: Could this `if {} else {}` be changed to this:

```suggestion
    submitButton.prop('disabled', !$(e.target).val());
```

OleksandrOrlov: Thank you! This looks really nice. I applied the suggested change:)

saraycp (approved):

saraycp:
> My only concern here is about style. The tabs for Write and Preview look a bit weird for me. [...]

Much better now, @OleksandrOrlov Thanks!

eduardoj (approved):

PR #10446, opened 2020-11-16
Add Cancel button to comment section

OleksandrOrlov:
It is not clear how to collapse comment section once it is expanded
while editing or replying.

The PR adds "Cancel" button for this purpose.

To verify this feature

1. Log in to the application;
2. Go to any page, where comments section exists (e.g. select "Your Home Project" in the sidebar);
3. Create a comment if there are no ones;
4. Press "Reply" or "Edit" buttons;
5. When the comment section is expanded, press "Cancel" button.

Expected result:
Comment section is collapsed.

**Before**
comment_before

**After**
after_comment

### DIFF
--- a/src/api/app/assets/javascripts/webui/tabs.js
+++ b/src/api/app/assets/javascripts/webui/tabs.js
@@ -31,7 +31,7 @@ $.fn.hasOverflow = function() {
   return (element.offsetHeight < element.scrollHeight - 1) || (element.offsetWidth < element.scrollWidth - 1);
 };
 
-$.fn.collapse = function(){
+$.fn.collapseCollapsible = function(){
   var collapsibleElements = this;
   
   return collapsibleElements.each(function() {
@@ -43,5 +43,5 @@ $.fn.collapse = function(){
 };
 
 $(document).ready(function() {
-  $('ul.collapsible').collapse();
+  $('ul.collapsible').collapseCollapsible();
 });
--- a/src/api/app/views/webui/comment/_comment_field.html.haml
+++ b/src/api/app/views/webui/comment/_comment_field.html.haml
@@ -18,8 +18,12 @@ class: "#{form_method}-comment-form" }, data: { preview_comment_url: preview_com
         class: 'w-100 form-control comment-field'
       .tab-pane.fade{ id: "preview_#{element_suffix}", role: 'tabpanel', 'aria-labelledby': 'preview-comment-tab' }
         .comment-preview.border-bottom.border-gray-300.my-3
-      - case form_method
-      - when :post
-        = f.submit 'Add comment', class: 'btn btn-primary mb-3', data: { disable_with: 'Creating comment...' }, disabled: true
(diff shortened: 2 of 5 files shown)


src/api/app/views/webui/comment/_comment_field.html.haml
```diff
@@ -18,8 +18,12 @@ class: "#{form_method}-comment-form" }, data: { preview_comment_url: preview_com
         class: 'w-100 form-control comment-field'
       .tab-pane.fade{ id: "preview_#{element_suffix}", role: 'tabpanel', 'aria-labelledby': 'preview-comment-tab' }
         .comment-preview.border-bottom.border-gray-300.my-3
-      - case form_method
-      - when :post
-        = f.submit 'Add comment', class: 'btn btn-primary mb-3', data: { disable_with: 'Creating comment...' }, disabled: true
-      - when :put
-        = f.submit 'Update comment', class: 'btn btn-primary mb-3', data: { disable_with: 'Updating comment...' }
+      .comment-controls.mb-3.float-right
```
dmarcoux, line 21: What is the reason for aligning the buttons to the right?

OleksandrOrlov: It could be discussed for sure. I just wanted to make it more consistent with the other forms where Cancel and Submit buttons are exist. For example:
delete_project
edit_project

As I can see, we do not have one rule for the position of the buttons, and in some cases when only one button is shown, it is aligned by left side. But in all cases with two buttons (at least those, that I've found), we aligning them to the right with the Cancel button first. 

WDYT, should I revert the buttons back to the left side and show them in the order: "Update Comment" "Cancel"?

dmarcoux: It should be consistent. It's not great that we aren't consistent with that in the whole application. With this change, it just looks weird because the `Reply`, `Edit` and `Delete` buttons are aligned to the left, but then the `Update` and `Cancel` buttons are aligned to the right.

I would revert this change, so keep all the buttons aligned to the left to be consistent on this page.

OleksandrOrlov: I've reverted the buttons back to the left.

dmarcoux (approved):

src/api/app/assets/javascripts/webui/tabs.js
```diff
@@ -31,7 +31,7 @@ $.fn.hasOverflow = function() {
   return (element.offsetHeight < element.scrollHeight - 1) || (element.offsetWidth < element.scrollWidth - 1);
 };
 
-$.fn.collapse = function(){
+$.fn.collapseCollapsible = function(){
```
OleksandrOrlov, line 34: I've renamed the function as it overrides `collapse` Bootstrap's function. Thanks @eduardoj for pointing me out:)

eduardoj (approved):

PR #10448, opened 2020-11-17
Fix spider test

bugfinder:
also skip links to show home:Iggy/ToBeDeletedTestPack


saraycp (approved):
Thanks @bugfinder

hennevogel:
Hm the real solution would be to handle deleted packages in the contributions feature. Did anyone create an issue for this?

PR #10453, opened 2020-11-19
Move 'Add Maintained Project' action next to Maintained Projects Header

danidoni:
**Before this change:**

**After this change:**

### DIFF
--- a/src/api/app/views/webui/projects/maintained_projects/index.html.haml
+++ b/src/api/app/views/webui/projects/maintained_projects/index.html.haml
@@ -6,7 +6,12 @@
 .card
   = render partial: 'webui/project/tabs', locals: { project: @project }
   .card-body
-    %h3 Maintained Projects
+    %h3
+      Maintained Projects
+      - if update_policy && feature_enabled?(:responsive_ux)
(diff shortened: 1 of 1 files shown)


danidoni:

eduardoj (approved):

src/api/app/views/webui/projects/maintained_projects/index.html.haml
```diff
@@ -6,7 +6,12 @@
 .card
   = render partial: 'webui/project/tabs', locals: { project: @project }
   .card-body
-    %h3 Maintained Projects
+    %h3
+      Maintained Projects
+      - if feature_enabled?(:responsive_ux)
```
dmarcoux, line 11: We should also check for `update_policy` here, no?

So it would be:
```suggestion
      - if update_policy && feature_enabled?(:responsive_ux)
```

danidoni: Oh yes, I forgot

dmarcoux (approved):
All good once the commits are squashed

Issue #10323, opened 2020-10-20
User profile page, "Involved Packages and Projects" missing clear filters option


vpereira:
Once the filter is applied by the user then a button called ‘Reset Filters’ or ‘’Clear All’’ should be given next to the filters so that the user can quickly make changes. If we don't have it, we will force the user to either refresh the page (assuming that the filter will be erased) or user have to empty the search and click on "Filter results"

dmarcoux:
![](https://github.trello.services/images/mini-trello-icon.png) [Address issues for the involved projects/packages feature](https://trello.com/c/BPrPcOcO/2061-address-issues-for-the-involved-projects-packages-feature)

coogor:
Sorry to say, but this new overview is a mess! I really struggle to find my packages - old overview was much more productive

hennevogel:
@coogor thanks for your feedback. Can you be in any way specific? What exactly is less productive and why?

coogor:
Sure.
1) When I enter 'my profile' I get an unsorted list of projects, including description. First I need to scroll down to look for the right project, second I know what in the projects, so the description is a waste of space
2) When I filter, e.g. for gnuhealth, I get exactly one package presented: Application:ERP:GNUHealth:Factory / gnuhealth
The same package is as well in A:E:G:3.2/3.4/3.6, but this is not displayed
3) The search for Tryton gives exactly Zero Results (oops....)
So, I need to juggle with filters without getting a proper result. Before I had a sorted list of involved projects, and on another page a list of packages. One click and I was where I needed to go.
Now I need a bunch of clicks to not get a result.
Does that explain? Otherwise I'm happy to explain further...

eduardoj:
Hi, thanks for the clarification of the issues.

> Sure.
> 
> 1. When I enter 'my profile' I get an unsorted list of projects, including description. First I need to scroll down to look for the right project, second I know what in the projects, so the description is a waste of space

Before introducing this redesign, the title of the project was displayed. Packages have no title, so nothing related to the package was displayed.

In an attempt to provide some context also for packages, the description for the packages was added. By the contrary, the whole description of each package should not be displayed. This will be fixed.

I will also revise the sorting.

> 2. When I filter, e.g. for gnuhealth, I get exactly one package presented: Application:ERP:GNUHealth:Factory / gnuhealth
>        The same package is as well in A:E:G:3.2/3.4/3.6, but this is not displayed

[...]

PR #10456, opened 2020-11-19
Clear filter options for Profile Involved Projects/Packages search

vpereira:
Screenshot_2020-11-19_14-41-22

Screenshot_2020-11-20_13-07-27

Screenshot_2020-11-20_13-07-11


vpereira:
https://trello.com/c/4cq6Kw2Y/1576-add-option-to-clear-filters-in-user-profile-involved-packages

vpereira:
[Add option to clear filters in user profile "involved packages"](https://trello.com/c/4cq6Kw2Y/1576-add-option-to-clear-filters-in-user-profile-involved-packages)

src/api/app/helpers/webui/user_helper.rb
```diff
@@ -1,4 +1,24 @@
 module Webui::UserHelper
+  def display_filters?(filters)
+    return false if filters.empty?
+    return false if filters.size == 1 && filters.include?('search_text')
+
+    true
+  end
+
+  def filters_to_humans(filters)
```
danidoni, line 9: :thought_balloon: I guess you want to say `filters_to_human` (singular) like [rails does](https://api.rubyonrails.org/classes/ActionView/Helpers/NumberHelper.html#method-i-number_to_human)?

vpereira: actually I'm accepting suggestions for this method name. Any better idea @danidoni ?

danidoni: Now its better

src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
```diff
@@ -50,6 +50,13 @@
           %button.btn.btn-primary.form-control#filter-button{ type: :submit }
             Apply Filter
 
+  - if display_filters?(filters.keys)
+    .card-body.pt-0
+    %li.list-inline-item
```
saraycp, line 55: Can this `li` be dropped and display only the `link_to`?

saraycp (changes requested):

src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
```diff
@@ -50,6 +50,12 @@
           %button.btn.btn-primary.form-control#filter-button{ type: :submit }
             Apply Filter
 
+    - if display_filters?(filters.keys)
```
saraycp, line 53: Having another look, I'd suggest moving all this code to the helper.
In the view we can simply call one helper method, for example: `= clear_filters_link(filters.keys)`.
And inside the helper method do all the checks and return a `span` containing the link or just a link.

vpereira: hi @saraycp, i did move the code to a helper as suggested, i just let the check if we have to display the filters or not in the partial to make easier to understand the logic while reading the partial

src/api/app/helpers/webui/user_helper.rb
```diff
@@ -1,4 +1,36 @@
...
+      end
+    end
+  end
+
+  def display_filters?(filters)
+    return false if filters.empty?
+    return false if filters.size == 1 && filters.include?('search_text')
+
+    true
+  end
```
saraycp, line 19: I think we can get rid of this method and use a simplified condition in the view. See comment there.

src/api/app/helpers/webui/user_helper.rb
```diff
@@ -1,4 +1,36 @@
...
+
+  def display_filters(filters)
```
saraycp, line 21: I would move this to `private`.

src/api/app/helpers/webui/user_helper.rb
```diff
@@ -1,4 +1,36 @@
 module Webui::UserHelper
+  def display_applied_filters(filters)
+    tag.span(class: 'd-block ml-4 mb-3') do
+      link_to(User.session!) do
+        tag.i(class: 'fas fa-trash-alt') do
```
saraycp, line 5: Screenshot_2020-11-20 Profile of Admin

Now there is some missing space between the icon and the link.

src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
saraycp: We can get rid of that helper method and use the following condition in the view instead:

`- unless filters.keys == ['search_text']`

AFAIK filter is never empty. If we don't apply any filter or search text, the `filters` variable contains `{ "search_text"=>nil }`.

So that condition will be true if there are no filters applied or if the only filter is some text, `{ "search_text"=>"some text" }`.

In the rest of the cases, the `filters` variable will have more than one key, so you can display the link.

vpereira: let me check

vpereira: right, could you please check it one more time @saraycp ?

src/api/app/helpers/webui/user_helper.rb
vpereira: fixed. I updated the screenshots to the last version.

src/api/app/helpers/webui/user_helper.rb
vpereira: done!

src/api/app/helpers/webui/user_helper.rb
vpereira: hm, i prefer to keep the method but add the suggested logic to it.

src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
saraycp: Ok. I would have moved it to the view, but I can understand that the method name is more meaningful than the condition itself. :+1:

saraycp (approved):

dmarcoux:
This PR closes the issue #10323, right?

PR #10457, opened 2020-11-19
Move Add Maintenance Incidents actions next to the Maintenance Incidents Header

danidoni:
## How it looks like now?
image

### DIFF
--- a/src/api/app/views/webui/projects/maintenance_incidents/index.html.haml
+++ b/src/api/app/views/webui/projects/maintenance_incidents/index.html.haml
@@ -1,5 +1,12 @@
 - @pagetitle = 'Maintenance Incidents'
 
+- if policy(@project).create? && feature_enabled?(:responsive_ux)
+  - content_for :actions do
+    %li.nav-item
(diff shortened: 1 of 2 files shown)


danidoni:

krauselukas (commented):
Shouldn't be the action in the sidebar in this case? Since the add action is related to the maintenance incident and the the view context are the incidents itself?

krauselukas (approved):
Sorry, I thought again, I think its correct looking at other examples

krauselukas (changes requested):
as discussed with @danidoni my original concern was correct, so the action belongs to the sidebar

DavidKang:
Not sure,  but this PR https://github.com/openSUSE/open-build-service/pull/10453, should also be reworked to be consistent, right?
I mean, if you are going to move it to the sidebar.

danidoni:
> Not sure, but this PR #10453, should also be reworked to be consistent, right?
> I mean, if you are going to move it to the sidebar.

Yes, that's the next step

danidoni:
> Not sure, but this PR #10453, should also be reworked to be consistent, right?
> I mean, if you are going to move it to the sidebar.

Taken care of here #10458

krauselukas (approved):

PR #10458, opened 2020-11-20
Move the 'Add Project to Maintain' action to the left sidebar

danidoni:
As this is a page action and not a contextual action it should be in the left Actions Sidebar.

## Before
image

## Now
image


vpereira (approved):

dmarcoux:
I disagree with the changes. This action is contextual and the context is that box where `Maintained Projects` is. It's the same thing we did for adding members to a group, like here:

example

PR #10461, opened 2020-11-20
Add characters count for User Biography

OleksandrOrlov:
There is a limit of characters that can be typed in Biography textarea.
After the limit is reached, no characters can be entered. But there is
no visual hint on how many characters remaining.

The commit adds a counter that updates in real-time to show to the user
on how many characters remaining.

How to test the feature:

1. Login to the application;
2. Go to User Profile Page;
3. Press "Edit Your Account" in sidebar menu;
4. Type any characters to Biography textarea.

Expected result:
Once textarea become in focus, the counter is appeared in the bottom-right corner of the textarea. It shows how many characters remaining when typing.

**When responsive_ux is disabled**
old_dialog

**When responsive_ux is enabled**
new_dialog

### DIFF
--- a/src/api/app/assets/javascripts/webui/user_profile.js
+++ b/src/api/app/assets/javascripts/webui/user_profile.js
@@ -5,3 +5,16 @@ function moveInvolvementToContainer() { // jshint ignore:line
 
   container.prepend($('#involvement'));
 }
+
+function updateCharactersCount(e) { // jshint ignore:line
+  var maxLength = $(e.target).attr('maxlength');
+  var currentLength = $(e.target).val().length;
+  var remainingCount = maxLength - currentLength;
+
+  // Change font color to "danger" when low amount of characters left
+  $('#bio-chars-counter').toggleClass('text-danger', remainingCount < 10);
+
+  // Make text singular or plural based on the amount of characters
+  var plural = (remainingCount === 1) ? '' : 's';
+  $('#bio-chars-counter').css("visibility", "visible").text(remainingCount + " character" + plural + " remaining");
+}
--- a/src/api/app/assets/stylesheets/webui/user.scss
+++ b/src/api/app/assets/stylesheets/webui/user.scss
@@ -32,3 +32,7 @@
     text-decoration: none;
   }
 }
+
+#bio-chars-counter {
+  min-height: 1.2rem;
+}
--- a/src/api/app/views/webui/user/_edit_dialog.html.haml
+++ b/src/api/app/views/webui/user/_edit_dialog.html.haml
@@ -16,7 +16,7 @@
             .form-group
               = f.label(:biography) do
                 Biography:
-                %small.form-text Max. #{User::MAX_BIOGRAPHY_LENGTH_ALLOWED} characters.
               = f.text_area(:biography, rows: 6, class: 'form-control', maxlength: User::MAX_BIOGRAPHY_LENGTH_ALLOWED)
+              %small.form-text.float-right#bio-chars-counter
         .modal-footer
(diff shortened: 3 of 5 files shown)


src/api/app/assets/javascripts/webui/user_profile.js
```diff
@@ -5,3 +5,30 @@ function moveInvolvementToContainer() { // jshint ignore:line
...
+function updateCharactersCount(e) {
+  var maxLength = $(e.target).attr('maxlength');
+  var currentLength = $(e.target).val().length;
+  var remainingCount = maxLength - currentLength;
+
+  // Change font color to "danger" when low amount of characters left
+  var style = window.getComputedStyle(document.body);
+  var defaultFontColor = style.getPropertyValue('color');
+  var dangerFontColor = style.getPropertyValue('--danger');
+  $('#bio-chars-counter').css('color', (remainingCount < 10) ? dangerFontColor : defaultFontColor);
```
dmarcoux, line 18: This could be replaced by [.toggleClass()](https://api.jquery.com/toggleclass/) with the `text-danger` class.

It would be like this:
```suggestion
  $('#bio-chars-counter').toggleClass('text-danger', remainingCount < 10);
```

OleksandrOrlov: Done. Thank you for the suggestion!

dmarcoux (approved):

src/api/app/assets/javascripts/webui/user_profile.js
```diff
@@ -5,3 +5,27 @@ function moveInvolvementToContainer() { // jshint ignore:line
...
+
+  // Change font color to "danger" when low amount of characters left
+  $('#bio-chars-counter').toggleClass('text-danger', remainingCount < 10);
+
+  // Make text singular or plural based on the amount of characters
+  var plural = (remainingCount === 1) ? '' : 's';
+  $('#bio-chars-counter').css("visibility", "visible").text(remainingCount + " character" + plural + " remaining");
+}
+
+$(document).ready(function(){
```
krauselukas, line 22: @OleksandrOrlov usually we try to call the javascript functions from the corresponding view, to avoid the overhead of calling the code everywhere. Its not wrong, but this way you can also avoid interference, incase you used the same html id in another view. If you grep for example for the `moveInvolvementToContainer()` function, you will see how it called from the haml file

OleksandrOrlov: Good point! Thank you for the note. I've moved the JS to haml file.

I've moved it to `show` file, as the same javascript is needed for both `_edit_dialog` and `_edit_account_form`.

src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -341,6 +341,7 @@ def feature_enabled?(feature)
   def feature_css_class
     css_classes = []
     css_classes << 'responsive-ux' if feature_enabled?(:responsive_ux)
+    css_classes << 'user-profile-redesign' if feature_enabled?(:user_profile_redesign)
```
eduardoj, line 344: If this new CSS class is not used, I would not introduce it. Is it needed?

OleksandrOrlov: I added this as a separate commit, as this is not directly related to the feature, but it should be fixed. I added this after discussion with Saray. I can move it to another PR, but I thought that this one line may be added here as a separate commit.

OleksandrOrlov: In order to avoid any confusion, I've removed the commit from this PR and will introduce the change in another PR.

eduardoj (approved):

krauselukas (approved):

PR #10468, opened 2020-11-23
Add maintenance incidents actions next to the maintenance incidents heading

danidoni:

**Before**
image

**After**
image


danidoni:

PR #10472, opened 2020-11-24
Set body class if user profile redesign is enabled

OleksandrOrlov:
The commit adds the appropriate class to the body tag when
user_profile_redesign feature is enabled.


dmarcoux (approved):

saraycp (approved):
Since we started working with feature flags, we added the corresponding classes to the `body`.
This is especially justified when we have specific CSS rules for the current feature flag. And we have realized that in this case, we do have specific CSS rules.

eduardoj (approved):

PR #10475, opened 2020-11-24
Move the 'Add Member' action to the bottom of the group members list

danidoni:
As we're moving contextual actions close to their context, the 'Add Member'
action should go below the members list.

**Before**

**After**

### DIFF
--- a/src/api/app/views/webui/groups/_group_members.html.haml
+++ b/src/api/app/views/webui/groups/_group_members.html.haml
@@ -1,14 +1,5 @@
 - write_access = policy(group).update?
 
-- if write_access
-  - unless feature_enabled?(:responsive_ux)
-    %ul.list-inline
-      %li.list-inline-item
-        = link_to('#', class: 'nav-link', data: { toggle: 'modal', target: '#add-group-user-modal' }) do
-          %i.fas.fa-plus-circle.text-primary
-          Add Member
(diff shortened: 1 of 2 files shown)


src/api/app/views/webui/groups/_group_members.html.haml
```diff
@@ -38,6 +37,14 @@
                     Maintainer
                   %i.fas.fa-spinner.fa-spin.d-none
 
+  - if write_access && feature_enabled?(:responsive_ux)
```
dmarcoux, line 40: I would simplify this view by removing the `feature_enabled?` check and remove the other block on line 3. The `render partial: ...` call on line 9 would then be moved under this block on line 40.

saraycp (commented):
Why is this action considered contextual? I'm not against but I would like to know the reasoning behind.

src/api/app/views/webui/groups/_group_members.html.haml
danidoni: Okay!

danidoni:
> Why is this action considered contextual? I'm not against but I would like to know the reasoning behind.

Because it affects something we have on the page (so it has a context), right?

danidoni:

dmarcoux (approved):

PR #10479, opened 2020-11-25
Move "Add Group" to the bottom of the group list

danidoni:
We are moving the "Add Group" action next to the group list because both
elements are related, so they should be next to each other.

**Before**
Screenshot 2020-11-25 at 15 13 25

**After**
Screenshot 2020-11-25 at 15 11 12

### DIFF
--- a/src/api/app/views/webui/groups/index.html.haml
+++ b/src/api/app/views/webui/groups/index.html.haml
@@ -5,9 +5,6 @@
   .card-body
     %h3
       = @pagetitle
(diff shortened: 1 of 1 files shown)


danidoni:

danidoni:
You can review the changes [here](https://obs-reviewlab.opensuse.org/danidoni-move-add-group-to-the-bottom-of-the-group-list/groups)

dmarcoux (approved):

eduardoj (approved):

PR #10483, opened 2020-11-25
Remove users/groups actions from sidebar

dmarcoux:
Before (responsive_ux)
before

Now (responsive_ux)
now

Now (without responsive_ux, it didn't change)
non-responsive-ux


src/api/app/views/webui/shared/user_or_groups_roles/_index.html.haml
```diff
@@ -61,7 +55,7 @@
                  locals: { records: groups, roles: roles, type: 'group',
                            user_is_maintainer: user_is_maintainer,
                            project: project, package: package }
-  - if !feature_enabled?(:responsive_ux) && user_is_maintainer
+  - if user_is_maintainer
     %ul.list-inline
```
krauselukas, line 59: @dmarcoux the PR still contains the `nav-link` class and the inline list

dmarcoux: So we're using `.pt-4` now?

danidoni: Yes

dmarcoux: Please have another look

danidoni (approved):

src/api/app/views/webui/shared/user_or_groups_roles/_index.html.haml
danidoni: Great

saraycp (changes requested):
The current behaviour hides the whole area when there are no groups. That should be fixed to keep the headings `User Roles` and `Group Roles` always visible. It's easier to understand with images:

**When there are some groups to list:**

Screenshot_2020-11-26 Open Build Service

**When there aren't any groups to list:**

Screenshot_2020-11-26 Open Build Service(1)

It would be nice if you could fix it. This also happens with the users list.

dmarcoux:
@saraycp I didn't change anything in regards to the issues you wrote about. To be completely honest, I don't believe that this page with its current design is worth investing more time into. I would rather merge this PR as it is.

src/api/app/views/webui/shared/user_or_groups_roles/_index.html.haml
```diff
@@ -12,12 +12,6 @@
...
 
-- if feature_enabled?(:responsive_ux) && user_is_maintainer
-  = render partial: 'webui/shared/user_or_groups_roles/responsive_ux/actions'
-  = render partial: 'webui/shared/user_or_groups_roles/add_user_dialog', locals: { project: project.to_param, package: package.to_param }
-  = render partial: 'webui/shared/user_or_groups_roles/delete_dialog'
-  = render partial: 'webui/shared/user_or_groups_roles/add_group_dialog', locals: { project: project.to_param, package: package.to_param }
-
 #involved-users{ data: data }
   - if users.present?
     %h3 User Roles
```
saraycp, line 17:
```suggestion
  %h3 User Roles
  - if users.present?
```

dmarcoux:
I flipped the lines, it should be fine now: fixed

saraycp (approved):
> I flipped the lines, it should be fine now:

It is! Thanks for applying my suggestion.

PR #10485, opened 2020-11-25
Remove 'Add Attribute' action from sidebar

dmarcoux:
Before (responsive_ux)
before

Now (responsive_ux)
now

Now (without responsive_ux, nothing changed beside using an inline list instead of a paragraph)
non-responsive-ux


src/api/app/views/webui/attribute/index.html.haml
```diff
@@ -39,9 +38,10 @@
       %p
         %em No attributes set
 
-    - if can_update_container && !feature_enabled?(:responsive_ux)
-      %p
-        = link_to(new_attribs_path(project: @project.name, package: @package), title: 'Add Attribute') do
-          %i.fas.fa-plus-circle.text-primary
-          Add Attribute
+    - if can_update_container
+      %ul.list-inline
```
dmarcoux, line 42: This change is to be consistent with other pages.

krauselukas (approved):

PR #10507, opened 2020-11-30
[backend] global timeout for services

M0ses:
This patch ensures that the BSConfig::service_timeout confiured on the
src_server is also used by servicedispatcher and service server.

### DIFF
--- a/src/backend/bs_service
+++ b/src/backend/bs_service
@@ -116,7 +116,10 @@ sub rm_rf {
 }
 
 sub run_source_update {
-  my ($cgi, $projid, $packid) = @_;
+  my ($cgi, $projid, $packid, $timeout) = @_;
+
+  # just as fallback for older servicedispatchers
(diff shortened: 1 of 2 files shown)


mlschroe:
Why do we need the timeout in the event?

src/backend/bs_service
```diff
@@ -305,7 +307,7 @@ my $dispatches = [
 
   '/service' => \&list_service,
   '/serverstatus' => \&BSStdServer::serverstatus,
-  '!- POST:/sourceupdate/$project/$package' => \&run_source_update,
+  '!- POST:/sourceupdate/$project/$package $timeout:intnum?' => \&run_source_update,
```
mlschroe, line 310: This should be `num`, not `intnum`. We do not want negative timeouts.

mlschroe (changes requested):

mlschroe (approved):

PR #10511, opened 2020-12-01
[api] fix place sourcediff in request diffing

adrianschroeter:
The sourcediff need to be part of the action where they belong to.

Not only the very fist action item.

(jsc#OBS-71)

### DIFF
--- a/src/api/app/controllers/request_controller.rb
+++ b/src/api/app/controllers/request_controller.rb
@@ -187,7 +187,7 @@ class RequestController < ApplicationController
         builder = Nokogiri::XML::Builder.new
         action.render_xml(builder)
         xml_request.add_child(builder.doc.root.to_xml)
-        xml_request.at_css('action').add_child(action_diff)
+        xml_request.at_xpath("//request/action[last()]").add_child(action_diff)
       else
         diff_text += action_diff
       end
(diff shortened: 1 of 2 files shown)


src/api/app/controllers/request_controller.rb
```diff
@@ -187,7 +187,7 @@ def request_command_diff
         builder = Nokogiri::XML::Builder.new
         action.render_xml(builder)
         xml_request.add_child(builder.doc.root.to_xml)
-        xml_request.at_css('action').add_child(action_diff)
+        xml_request.at_xpath("//request/action[last()]").add_child(action_diff)
```
danidoni, line 190:
```suggestion
        xml_request.at_xpath('//request/action[last()]').add_child(action_diff)
```

danidoni (changes requested):

danidoni:
I think this change is already in master [here](https://github.com/adrianschroeter/open-build-service/blob/b83eef7075e69dfca7628ac317b24d11ef10a73d/src/api/app/controllers/request_controller.rb#L190), am I wrong?

Can you try to rebase this PR against master and see if this change is still relevant?

hennevogel:
It's already there...

PR #10512, opened 2020-12-01
[backend] repserver: also log the build type of finished jobs

mlschroe:


hennevogel:
Can we also add this to the events?

mlschroe:
Of course! Just add support for a "buildtype" element to the API.

hennevogel:
> Of course! Just add support for a "buildtype" element to the API.

Okay, here or do you want to do it together in another PR?

hennevogel:
If it's in the payload, I think this should be it.

hennevogel:
You saw I pushed a commit right? :-)

PR #10522, opened 2020-12-03
[dist] drop rpath from mysql2.so and move operations to install

saraycp:
After applying the changes to introduce the new minor release 2.10.8, we realized these changes introduced by Rüdiger Oertel are also necessary.

> Rework obs-bundled-gems.spec to do all compile and install in the install section of the specfile as future rpm versions
> will clean buildroot at the start of the install section and bundle does not properly decouple compile and install section.

Adaptation of 086e86379 by Rüdiger Oertel <ro@suse.de>. In the current version, we also needed to require `chrpath` for openSUSE 15.2.


hennevogel:
`checkout_code` succeeded, otherwise the rest would not run. Some GH hickup. See the CI logs. Merging

PR #10525, opened 2020-12-03
Add a checklist for deployment consideration to the PR template

saraycp:
Some PRs introduce changes that require special actions in the reference server. Actions that sometimes should be performed at a specific point of time.

This checklist can help us to identify if this PR needs special attention.

-----------------

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,19 @@
+_Replace this line with a description of your changes. Please follow the suggestions in the comment below._
+
+-----------------
+
+If this PR requires any particular action or consideration before deployment,
+please check the reasons or add your own to the list:
+
+* [ ] Requires a database migration that can cause downtime. Postpone deployment until maintenance window[1].
+* [ ] Contains a data migration that can cause temporary inconsistency, so should be run at a specific point of time.
+* [ ] Changes some configuration files (e.g. options.yml), so the changes have to be applied manually in the reference server.
+* [ ] A new Feature Toggle[2] should be enabled in the reference server.
+* [ ] Proper documentation or announcement has to be published upfront since the introduced changes can confuse the users.
+
+[1] https://github.com/openSUSE/open-build-service/wiki/Deployment-of-build.opensuse.org#when-there-are-migrations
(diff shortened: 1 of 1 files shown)


danidoni (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,20 @@
...
+
+If this PR requires any particular action or consideration before deployment,
+please check the reasons or add your own to the list:
+
+* [ ] Requires a schema migration that can cause downtime. Postpone deployment until maintenance window[1].
+* [ ] Contains a data migration that should be run at a specific point of time.
+* [ ] Changes some configuration files (e.g. options.yml), so the changes have to be applied manually in the reference server.
+* [ ] A new Feature Toggle[2] should be enabled in the reference server.
+* [ ] Proper documentation or announcement has to be published upfront since the introduced changes can confuse the users.
+* [ ] Requires running any task or action that can cause some temporary inconsistency or downtime. Postpone deployment until maintenance window.
```
krauselukas, line 13: I think you are talking about a rake task here that modifies existing data right? I would rephrase it a bit, maybe:

krauselukas:
```suggestion
* [ ] Requires transformation of existing data through e.g. a rake task, which might cause temporary inconsistency and therefore downtime. Postpone deployment until maintenance window.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,20 @@
+_Replace this line with a description of your changes. Please follow the suggestions in the comment below._
+
+-----------------
+
+If this PR requires any particular action or consideration before deployment,
+please check the reasons or add your own to the list:
+
+* [ ] Requires a schema migration that can cause downtime. Postpone deployment until maintenance window[1].
```
krauselukas, line 8: I would use `database migration` or `db migration`. To me `schema migration` sounds a bit generic :)

saraycp: good point, yes

.github/PULL_REQUEST_TEMPLATE.md
saraycp: Well, there is already a item about data migrations, I'll try to extend it a bit and drop this one.

PR #10546, opened 2020-12-10
Add job to delay the published repositories cleanup

saraycp:
After accepting maintenance release requests, the corresponding published repositories were removed inmediately. This broke some automated maintenance tests because the repositories were removed before the tests finished.

We have moved the cleanup to a delayed job, so the repositories will be removed some specific time after they are published.

-----------------

### DIFF
--- a/src/api/app/jobs/published_repositories_cleanup_job.rb
+++ b/src/api/app/jobs/published_repositories_cleanup_job.rb
@@ -1,0 +1,6 @@
+class PublishedRepositoriesCleanupJob < ApplicationJob
+  def perform(source_project_name)
+    # cleanup published binaries to save disk space on ftp server and mirrors
+    Backend::Api::Build::Project.wipe_published_locked(source_project_name)
+  end
+end
--- a/src/api/app/models/bs_request_action_maintenance_release.rb
+++ b/src/api/app/models/bs_request_action_maintenance_release.rb
@@ -59,8 +59,7 @@ class BsRequestActionMaintenanceRelease < BsRequestAction
 
       next if cleaned_projects[sprj]
 
-      # cleanup published binaries to save disk space on ftp server and mirrors
-      Backend::Api::Build::Project.wipe_published_locked(sprj)
+      maintenance_release_cleanup(sprj)
       cleaned_projects[sprj] = 1
     end
     opts[:projectCommit] = {}
@@ -167,6 +166,14 @@ class BsRequestActionMaintenanceRelease < BsRequestAction
         end
       end
     end
+  end
+
+  # Delaying removal of published repositories for accepted maintenance release requests,
+  # gives some margin to the automated maintenance tests to be finished.
+  def maintenance_release_cleanup(project_name)
+    delay = CONFIG['maintenance_release_repositories_lifetime']
+    set_options = delay ? { wait: delay.seconds } : {}
+    PublishedRepositoriesCleanupJob.set(set_options).perform_later(project_name)
   end
 
   #### Alias of methods
--- a/src/api/config/options.yml.example
+++ b/src/api/config/options.yml.example
@@ -204,6 +204,10 @@ default: &default
   #   icon: sponsor_abc
   #   url: https://www.example.com
 
+  # Lifetime for repositories published after accepting maintenance release requests.
+  # Default: 2 days (172800 seconds).
+  # maintenance_release_repositories_lifetime: 172800
+
 production:
   <<: *default
 
--- a/src/api/test/functional/maintenance_test.rb
+++ b/src/api/test/functional/maintenance_test.rb
@@ -901,6 +901,10 @@ class MaintenanceTests < ActionDispatch::IntegrationTest
   end
 
   def test_create_maintenance_project_and_release_packages
+    # FIXME: https://github.com/rails/rails/issues/37270
+    (ActiveJob::Base.descendants << ActiveJob::Base).each(&:disable_test_adapter)
+    ActiveJob::Base.queue_adapter = :inline
+
     # Backup
     system("for i in #{Rails.root}/tmp/backend_data/projects/BaseDistro2.0.pkg/*.rev; do cp $i $i.backup; done")
     # the birthday of J.K.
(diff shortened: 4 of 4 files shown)


src/api/app/models/bs_request_action_maintenance_release.rb
```diff
@@ -59,8 +59,9 @@ def per_request_cleanup(opts)
 
       next if cleaned_projects[sprj]
 
-      # cleanup published binaries to save disk space on ftp server and mirrors
-      Backend::Api::Build::Project.wipe_published_locked(sprj)
+      # Waiting 2 days before removing published repositories for accepted release requests,
+      # gives some margin to the automated maintenance tests to be finished.
+      PublishedRepositoriesCleanupJob.set(wait: 2.days).perform_later(sprj)
```
danidoni, line 64: :thought_balloon: What about moving the `2.days` into a constant?

adrianschroeter:
On Freitag, 11. Dezember 2020, 14:45:58 CET Dani Donisa wrote:
> @danidoni commented on this pull request.
> 
> 
> 
> > @@ -59,8 +59,9 @@ def per_request_cleanup(opts)
>  
>        next if cleaned_projects[sprj]
>  
> -      # cleanup published binaries to save disk space on ftp server and mirrors
> -      Backend::Api::Build::Project.wipe_published_locked(sprj)
> +      # Waiting 2 days before removing published repositories for accepted release requests,
> +      # gives some margin to the automated maintenance tests to be finished.
> +      PublishedRepositoriesCleanupJob.set(wait: 2.days).perform_later(sprj)
> 
> :thought_balloon: What about moving the `2.days` into a constant?

an option in options.yml would be nice

-- 

src/api/app/models/bs_request_action_maintenance_release.rb
saraycp: The value has been moved to options.yml and use seconds instead of days, in case some instance needs to configure a shorter delay (minutes, hours...).

src/api/config/options.yml.example
```diff
@@ -203,6 +203,10 @@ default: &default
   #   description: We are the best at ABC
   #   icon: sponsor_abc
   #   url: https://www.example.com
+  #
+  # Set the waiting time until the published repositories can be removed.
+  # The default value is 2 days (172800 seconds).
+  removal_delay: 172800
```
danidoni, line 209: Great!

src/api/app/models/bs_request_action_maintenance_release.rb
danidoni: :+1:

src/api/config/options.yml.example
coolo: The name and description of this option sound scary. That this is only about accepting maintenance release requests is far from obvious

danidoni: Maybe `published_repositories_cleanup_job_delay`?

coolo: I would claim in about all OBS installations, this option is not relevant as no maintenance release requests are seen. But every installation has published repositories *not* cleaned up by this setting.

saraycp:
@coolo, @danidoni please have another look. I renamed the option, made it optional and also adapted the tests.

src/api/app/jobs/published_repositories_cleanup_job.rb
```diff
@@ -0,0 +1,8 @@
+class PublishedRepositoriesCleanupJob < ApplicationJob
+  self.queue_adapter = :async if Rails.env.test?
```
hennevogel, line 2: Why are you doing this? We have `config.active_job.queue_adapter = :inline` in the test environment.

src/api/test/functional/maintenance_test.rb
```diff
@@ -1473,9 +1473,10 @@ def test_create_maintenance_project_and_release_packages
     run_publisher
     # published binaries from incident got removed?
     get "/published/#{incident_project}/BaseDistro3/i586/package-1.0-1.i586.rpm"
-    assert_response 404
+    assert_response 200
```
hennevogel, line 1476: How can this now be 200 if there is no delay in removing (as before)?

hennevogel (changes requested):

src/api/test/functional/maintenance_test.rb
saraycp: As soon as I move the cleanup to a job, the files are found (so 200 is returned), no matter if the job is performed with or without delay. Even if the queue adapter is `:inline`.

Apart from that, my plan was to emulate the case as similar to production as possible. For that, if I set a delay like in `PublishedRepositoriesCleanupJob.set(wait: 3.seconds).perform_later(project_name)`, I need to use `:async` as queue adapter. Otherwise, I'll get an error like:

```
NotImplementedError (Use a queueing backend to enqueue jobs in the future.
```

https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/InlineAdapter.html
https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/AsyncAdapter.html

src/api/app/jobs/published_repositories_cleanup_job.rb
saraycp: Explained in https://github.com/openSUSE/open-build-service/pull/10546#discussion_r556042023

src/api/test/functional/maintenance_test.rb
hennevogel: > As soon as I move the cleanup to a job, the test finds the files

Why?

saraycp: Well, I've finally decided not to change the queue adapter, so it will keep using `:inline` since it's the one in the setup.
In fact, now that the option `# maintenance_release_repositories_lifetime: 172800` is commented/optional, it won't be used in CircleCI tests, so I emulate the same locally.

Regarding why the cleanup job find the files if there isn't any waiting time and `:inline`is set up... I'm still trying to figure out.

coolo:
> @coolo, @danidoni please have another look. I renamed the option, made it optional and also adapted the tests.

Much clearer option name now

src/api/test/functional/maintenance_test.rb
```diff
@@ -901,6 +901,10 @@ def test_manual_branch_with_extend_names
   end
 
   def test_create_maintenance_project_and_release_packages
+    # FIXME: https://github.com/rails/rails/issues/37270
+    (ActiveJob::Base.descendants << ActiveJob::Base).each(&:disable_test_adapter)
+    ActiveJob::Base.queue_adapter = :inline
+
```
saraycp, line 907: I finally had to use this trick to force `:inline` adapter, as in many other tests. Now the tests pass everywhere.

danidoni (approved):
Once the checks are green. This is good to go

hennevogel (approved):

PR #10549, opened 2020-12-11
Adds a unique index on the name column on the status_checks table

danidoni:
This PR addresses the [cop regarding an unique index missing](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsuniquevalidationwithoutindex) detected by [this depfu PR](https://github.com/openSUSE/open-build-service/pull/10545)

### DIFF
--- a/src/api/app/models/status/check.rb
+++ b/src/api/app/models/status/check.rb
@@ -59,17 +59,18 @@ end
 # Table name: status_checks
 #
 #  id                :integer          not null, primary key
-#  name              :string(255)
+#  name              :string(255)      indexed => [status_reports_id]
 #  short_description :string(255)
 #  state             :string(255)
 #  url               :string(255)
 #  created_at        :datetime         not null
 #  updated_at        :datetime         not null
-#  status_reports_id :integer          indexed
+#  status_reports_id :integer          indexed => [name], indexed
 #
 # Indexes
 #
-#  index_status_checks_on_status_reports_id  (status_reports_id)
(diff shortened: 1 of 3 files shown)


krauselukas (changes requested):
@danidoni If I'm not mistaken the validation here  https://github.com/openSUSE/open-build-service/blob/a37e0015fc7df83c642c16593f40e4c20fdb140b/src/api/app/models/status/check.rb#L19 is scoped. The name of a status check itself is not uniq. I think you have to do whats described here https://github.com/openSUSE/open-build-service/blob/a37e0015fc7df83c642c16593f40e4c20fdb140b/src/api/app/models/status/check.rb#L19

danidoni:
> @danidoni If I'm not mistaken the validation here
> 
> https://github.com/openSUSE/open-build-service/blob/a37e0015fc7df83c642c16593f40e4c20fdb140b/src/api/app/models/status/check.rb#L19
> is scoped. The name of a status check itself is not uniq. I think you have to do whats described here https://guides.rubyonrails.org/active_record_validations.html#uniqueness so to create a unique index on both columns.

You are so right! I've added the index for both columns. Please, check it out again.

krauselukas (approved):
LGTM

danidoni:
We need to wait for the next maintenance window to deploy this: Thursday, December 17th, 2020

dmarcoux:
Adding an index is safe with MariaDB, we don't need to wait for a maintenance window.

dmarcoux:
Reverted in #10545

PR #10567, opened 2020-12-18
Remove mentions of openFATE on error pages

### LINKED ISSUE #10566: open build service 404 page mentions obsolete "openFATE"
# Issue Description
The 404 page of build.opensuse.org shows

```
 You have an idea how to make OBS better, faster, stronger? Those ideas we call feature and we track them in openFATE.
```

linking to https://features.opensuse.org/feature/new which is obsolete

Expected Result
============
Probably openFATE is not used anymore but also github issues? I suggest to simply link to the github issues instead or not even ask the user to make the distinction between feature request or bug anymore.

How to Reproduce
===============
1. Go to https://build.opensuse.org/project/show/foo
2. In section "we better ourselves" find the outdated text
3. See the fate page loaded with the message about a deprecated service

Further details
======
Similar to https://github.com/openSUSE/obs-landing/issues/237

okurz:
openFATE has been discontinued and should no longer be linked to as
there is no archive.

This commit is replacing the outdated references with the same text as
used in https://github.com/openSUSE/obs-landing/pull/276 referencing
github issues as the correct place for features as well as used for
issue reports.

Fixes #10566

### DIFF
--- a/src/api/public/402.html
+++ b/src/api/public/402.html
@@ -96,7 +96,7 @@
         <div class="col-md-4">
           <h4>...we better ourselves</h4>
           <p>
-            You have an idea how to make OBS better, faster, stronger? Those ideas we call <em>feature</em> and we track them in <a href="https://features.opensuse.org/feature/new">openFATE</a>.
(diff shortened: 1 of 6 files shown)


danidoni (approved):

dmarcoux:
I just noticed, the backend submodule was changed in this PR... This should not be included.

PR #10591, opened 2021-01-05
Split OpenAPI documentation in several files

eduardoj:
Splitting the documentation in several files will make it easier to introduce or change the documentation in the near future.

To test what has been introduced, you can have a look at the new documentation page here:

- Start the development environment
- Go to http://localhost:3000/apidocs-new
 
-----------------

If this PR requires any particular action or consideration before deployment, please check the reasons or add your own to the list:


krauselukas:
@eduardoj I get an error in the review app 
Screenshot_2021-01-05 Swagger UI

eduardoj:
@krauselukas:

> @eduardoj I get an error in the review app https://obs-reviewlab.opensuse.org/eduardoj-split_openapi/apidocs-new/#/

I will add a commit using relative routes. That should solve the problem. In the last force-push I have only rebased.

eduardoj:
@krauselukas, I added a commit with relative paths. It seems to work. Could you please check agagin?

krauselukas (approved):

PR #10602, opened 2021-01-07
Add links to schemas used for body validation

eduardoj:
These schemas (.xsd and .rng files) are currently used to validate both requests and responses.
    
For now we are defining OpenAPI schemas only to create examples, and we are not using OpenAPI schemas yet to validate bodies of requests or responses.

For example, this was the link added to the description of the response of the `/about` route:

Screenshot from 2021-01-07 15-49-03

-----------------

### DIFF
--- a/src/api/public/apidocs-new/components/responses/succeeded.yaml
+++ b/src/api/public/apidocs-new/components/responses/succeeded.yaml
@@ -1,4 +1,7 @@
-description: OK. The request has succeeded.
+description: |
+  OK. The request has succeeded.
+
+  XML Schema used for body validation: [status.xsd](../schema/status.xsd)
 content:
   application/xml; charset=utf-8:
     schema:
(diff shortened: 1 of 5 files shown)


src/api/public/apidocs-new/components/responses/succeeded.yaml
```diff
@@ -1,4 +1,7 @@
-description: OK. The request has succeeded.
+description: |
+  OK. The request has succeeded.
+
+  Schema used for body validation: [status.xsd](../schema/status.xsd)
```
hennevogel, line 4:
```suggestion
  XML Schema used for body validation: [status.xsd](../schema/status.xsd)
```

hennevogel (approved):

src/api/public/apidocs-new/components/responses/succeeded.yaml
eduardoj: Applied.

danidoni (approved):

PR #10603, opened 2021-01-07
Validate presence of project on Staging::Workflow.

### LINKED ISSUE #7175: Missing validation for project in Staging:Workflow
From `src/api/app/models/staging/workflow.rb`:

```ruby
 parent = Project.find_or_initialize_by(name: "#{project.name}:Staging:#{letter}")
```

which is called in an after create. But it is not compulsory to have a project (at least there is not validation) which causes:

`undefined method `name' for nil:NilClass`

Solution: Add the validation and not allow to have an staging workflow without a project and fixed everything this breaks.

bjoernalbers:
This fixes #7175.

I've added one validation with corresponding tests.

### DIFF
--- a/src/api/spec/models/staging/workflow_spec.rb
+++ b/src/api/spec/models/staging/workflow_spec.rb
@@ -20,6 +20,11 @@ RSpec.describe Staging::Workflow, type: :model do
     staging_workflow
   end
 
+  describe 'validations' do
+    it { is_expected.to validate_presence_of(:managers_group) }
+    it { is_expected.to validate_presence_of(:project) }
+  end
+
   context 'when created' do
     let(:role) { Role.find_by_title('reviewer') }
 
--- a/src/api/app/models/staging/workflow.rb
+++ b/src/api/app/models/staging/workflow.rb
@@ -30,7 +30,7 @@ class Staging::Workflow < ApplicationRecord
   has_many :request_exclusions, class_name: 'Staging::RequestExclusion', foreign_key: 'staging_workflow_id', dependent: :destroy
   has_many :excluded_requests, through: :request_exclusions, source: :bs_request
 
-  validates :managers_group, presence: true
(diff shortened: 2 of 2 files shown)


danidoni (approved):

src/api/spec/models/staging/workflow_spec.rb
```diff
@@ -20,6 +20,11 @@
     staging_workflow
   end
 
+  describe 'validations' do
+    it { is_expected.to validate_presence_of(:managers_group) }
+    it { is_expected.to validate_presence_of(:project) }
```
hennevogel, line 25: We usually don't test the (presence of) validations. Only if they contain logic.

dmarcoux:
I wonder if we would need a data migration in addition to the validation. There might be staging workflows without a valid project. I didn't check...

hennevogel:
Indeed there are, but they are not accessible anymore anyway as we route through the project. I don't think we need to do anything about them in this PR.

src/api/spec/models/staging/workflow_spec.rb
vpereira: hm, i like the tests to make sure that the validation exists and wasn't disabled (i.e some disable it to debug something else, and forget to enable it back) or removed by mistake.

bjoernalbers: If it's not possible to create a staging workflow w/o associated project (via the UI) there might be no other way to test the validation.
Plus these tests are quite cheap, no need to hit the db either.

vpereira (approved):

vpereira:
> Indeed there are, but they are not accessible anymore anyway as we route through the project. I don't think we need to do anything about them in this PR.

In production:

```
 > Staging::Workflow.where(project_id: nil).first
 => 7
```

vpereira:
Thank you for your contribution @bjoernalbers!

PR #10612, opened 2021-01-12
Add default value to Channel#disabled, to avoid the three-state boolean

vpereira:
We have the issue of [Three-state boolean](https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem) in different places of our database. We should address them with:

- database migration adding a default
- data migration to handle production
- a test case if necessary

-----------------

### DIFF
--- a/src/api/db/migrate/20210112091454_change_channel_disabled_default.rb
+++ b/src/api/db/migrate/20210112091454_change_channel_disabled_default.rb
@@ -1,0 +1,7 @@
+class ChangeChannelDisabledDefault < ActiveRecord::Migration[6.0]
+  # rubocop:disable Rails/ReversibleMigration
+  def change
+    change_column_default :channels, :disabled, false
+  end
+  # rubocop:enable Rails/ReversibleMigration
+end
--- a/src/api/db/data/20210112091948_add_channel_default.rb
+++ b/src/api/db/data/20210112091948_add_channel_default.rb
@@ -1,0 +1,11 @@
+class AddChannelDefault < ActiveRecord::Migration[6.0]
+  def up
+    # rubocop:disable Rails/SkipsModelValidations
+    Channel.where(disabled: nil).update_all(disabled: false)
+    # rubocop:enable Rails/SkipsModelValidations
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/src/api/db/schema.rb
+++ b/src/api/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_12_09_105103) do
+ActiveRecord::Schema.define(version: 2021_01_12_091454) do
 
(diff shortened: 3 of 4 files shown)


vpereira:
Added "DO NOT MERGE" tag because in our code base three state boolean is often done because people think in XML:

element is there and enabled, element is there and disabled, element is not there

TODO:

- [ ] check if we have enough test coverage in this area

coolo:
No idea what I have to do with channels, I never touched that.

vpereira:
> No idea what I have to do with channels, I never touched that.

Nothing with Channels, but a maybe a caveat with XML and booleans being `nil` ?  :)

src/api/db/migrate/20210112091454_change_channel_disabled_default.rb
```diff
@@ -0,0 +1,7 @@
+class ChangeChannelDisabledDefault < ActiveRecord::Migration[6.0]
+  # rubocop:disable Rails/ReversibleMigration
+  def change
+    change_column_default :channels, :disabled, false
```
dmarcoux, line 4: Could this be written like that?

```suggestion
    change_column_default :channels, :disabled, from: nil, to: false
```

adrianschroeter:
this model just follows the XML definition, so the three state exists there.

We (api and backend code) do not make a difference indeed atm, on the other side it might change. Not sure if it is worth the effort here. Why is the three state a problem for you?

src/api/db/migrate/20210112091454_change_channel_disabled_default.rb
vpereira: It definitely could be written like that to "make rubocop happy". However I don't see a scenario where we are going rollback from treating boolean as boolean and therefore having just two states - true or false - to something where we introduce a third state: `nil`.  I would prefer to disable this Cop and rely on `strong_migrations`

vpereira:
> this model just follows the XML definition, so the three state exists there.
> 
> We (api and backend code) do not make a difference of not set or disabled indeed atm, on the other side it might change. Not sure if it is worth the effort here. Why is the three state a problem for you?

Hi @adrianschroeter, thank you for your input. That's was exactly the kind of information that i was looking for!

I was having a related issue in another area, where I was doing a ` != TRUE` comparison. In SQL when you do a != TRUE, any NULL values will not be included in the results. This is because the NULL value represents an unknown value and not false. So i was getting an unexpected results. 

I then searched through our code base and found some occurrences of this "issue".  I think would be great to fix it, however there is no pressure on that matters, for sure.

adrianschroeter:
On Dienstag, 12. Januar 2021, 16:25:48 CET Victor Pereira wrote:
> > this model just follows the XML definition, so the three state exists there.
> > 
> > We (api and backend code) do not make a difference of not set or disabled indeed atm, on the other side it might change. Not sure if it is worth the effort here. Why is the three state a problem for you?
> 
> Hi @adrianschroeter, thank you for your input. That's was exactly the kind of information that i was looking for!
> 
> I was having a related issue in another area, where I was doing a ` != TRUE` comparison. In SQL when you do a != TRUE, any NULL values will not be included in the results. This is because the NULL value represents an unknown value and not false. So i was getting an unexpected results. 

Usually NULL/nil represents the default behaviour which is in the code in our logic.

While true or false overwrites it.

-- 

PR #10613, opened 2021-01-12
Add Tasks link to the top navigation for users not in the beta program

### LINKED ISSUE #10611: Task Counter Went Away
The task counter is not visible anymore, just an unimportant button with now highlight of work a packager has to do.

dmarcoux:
This includes a counter. All of this is displayed when the notifications_redesign feature is not enabled, so for users not in the
beta program.

The Tasks link in the Places menu is not changed.

Fixes #10611

Example with a lot of tasks (so yes, it scales):
it-scales

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -804,7 +804,6 @@ class User < ApplicationRecord
     to_s
   end
 
-  # TODO: Remove once responsive_ux is out of beta
   def tasks
     Rails.cache.fetch("requests_for_#{cache_key_with_version}") do
       declined_requests.count +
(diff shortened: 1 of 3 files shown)


src/api/app/models/user.rb
```diff
@@ -804,7 +804,7 @@ def to_param
     to_s
   end
 
-  # TODO: Remove once responsive_ux is out of beta
+  # TODO: Remove once notifications_redesign is out of beta
```
hennevogel, line 807: nitpick: *if* we replace tasks. I don't think this comment makes much sense...

dmarcoux: Yeah... I wasn't sure. I'll remove it completely.

dmarcoux: As for the `Tasks` link in the top navigation, it's only shown if the `notifications_redesign` feature isn't enabled. If the tasks stay, then the changes in this PR will have to be adapted. Time will tell...

hennevogel (approved):

PR #10618, opened 2021-01-13
WIP: Update rails to rails 6.1

vpereira:
Update Rails from 6.0.34 to 6.1.1

More information: https://guides.rubyonrails.org/6_1_release_notes.html


dmarcoux:
I also found this in the official guides: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-0-to-rails-6-1

vpereira:
still working in progress

src/api/config/application.rb
```diff
@@ -13,7 +13,7 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded.
 
     # Enable rails version 6.0 defaults
```
dmarcoux, line 15: We should change this comment to something like:

`Enable rails version X defaults`

Or we update it every time....

vpereira: yes, rails tries to update it with their update tool, but unfortunately we have too many customization on this file.

vpereira:
We should handle this issue https://bigbinary.com/blog/migrating-existing-session-cookies-while-upgrading-to-rails-4-1-and-above but we have to change some custom initializers to be able change it to `:json`

vpereira:
Looks like we have some blockers:

- https://github.com/rails/rails/issues/41126
- https://github.com/rails/rails/issues/41091

PR #10624, opened 2021-01-14
[backend] publish all ChangeLog.* files

adrianschroeter:

-----------------

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -2171,6 +2171,10 @@ sub publish {
 	  # collect virt-builder index parts
 	  $kiwiindex .= readstr("$r/$bin", 1) . "\n";
 	  next;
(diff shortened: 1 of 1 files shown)


ikapelyukhin:
:shipit:

src/backend/bs_publish
```diff
@@ -2171,6 +2171,9 @@ sub publish {
 	  # collect virt-builder index parts
 	  $kiwiindex .= readstr("$r/$bin", 1) . "\n";
 	  next;
+	} elsif ($bin =~ /^ChangeLog./) {
```
eduardoj, line 2174:
```suggestion
	} elsif ($bin =~ /^ChangeLog\./) {
```

mlschroe (changes requested):
NAK unless you add some description about why you think this is needed.

mlschroe (approved):

PR #10628, opened 2021-01-14
Remove unneeded path parameter for creating an issue tracker from apidocs documentation

eduardoj:
The `POST` route for issue trackers doesn't require an issue tracker id in path parameters. And the `create` method in the controller doesn't make use of any path parameters.

This is how issue_trackers routes looks like now:

```
issue_tracker_issue GET    /issue_trackers/:issue_tracker_id/issues/:id(.:format) issues#show                                                            
     issue_trackers GET    /issue_trackers(.:format)                              issue_trackers#index                                                   
                    POST   /issue_trackers(.:format)                              issue_trackers#create                                                  
      issue_tracker GET    /issue_trackers/:id(.:format)                          issue_trackers#show                                                    
                    PATCH  /issue_trackers/:id(.:format)                          issue_trackers#update                                                  
                    PUT    /issue_trackers/:id(.:format)                          issue_trackers#update                                                  
                    DELETE /issue_trackers/:id(.:format)                          issue_trackers#destroy                                                 
```

Also replace `bnc` as an example of ids with `<id>`, in the comments, for the description of the route related to the controller method.

-----------------

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -47,6 +47,12 @@ XmlResult: issue_trackers
 
 <name>: Issue tracker name
 
+POST /issue_trackers
+
+  Create an issue tracker.
+
+XmlResult: status
+
 GET /issue_trackers/<name>
 
   Read issue tracker data.
@@ -56,12 +62,6 @@ XmlResult: issue_tracker
 PUT /issue_trackers/<name>
 
   Update issue tracker data.
-
-XmlResult: status
-
-POST /issue_trackers/<name>
-
-  Create new issue tracker.
 
 XmlResult: status
 
--- a/src/api/app/controllers/issue_trackers_controller.rb
+++ b/src/api/app/controllers/issue_trackers_controller.rb
@@ -38,8 +38,8 @@ class IssueTrackersController < ApplicationController
 
   # GET /issue_trackers/<name>
   def show
-    @issue_tracker = IssueTracker.find_by_name(params[:id])
-    render_error(status: 404, errorcode: 'not_found', message: "Unable to find issue tracker '#{params[:id]}'") && return unless @issue_tracker
+    @issue_tracker = IssueTracker.find_by_name(params[:name])
+    render_error(status: 404, errorcode: 'not_found', message: "Unable to find issue tracker '#{params[:name]}'") && return unless @issue_tracker
 
     respond_to do |format|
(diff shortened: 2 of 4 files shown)


docs/api/api/api.txt
```diff
@@ -47,6 +47,12 @@ XmlResult: issue_trackers
 
 <name>: Issue tracker name
 
+POST /issue_trackers
+
+  Create new issue tracker.
```
dmarcoux, line 52: Creating something already implies that it's new. I suggest going with:

```suggestion
  Create an issue tracker.
```

src/api/app/controllers/issue_trackers_controller.rb
```diff
@@ -46,7 +46,7 @@ def create
     end
   end
 
-  # PUT /issue_trackers/bnc
+  # PUT /issue_trackers/<id>
```
dmarcoux, line 49: This is a name, not an id. Same for the other change in this file.

```suggestion
  # PUT /issue_trackers/<name>
```

dmarcoux (changes requested):
The routes are expecting an id because this is the default for resources in Rails. However, it is really a name. We are retrieving the issue trackers with _IssueTracker.**find_by_name**(params[:id])_, so it's clearly a name.

I suggest renaming the default param from `id` to `name` since this is what it really is. This involves changing the routes. We would need to add `param: :name` to the resources defined in https://github.com/openSUSE/open-build-service/blob/6bf4751cebcd518bbeaaafc98b5113c0ec230842/src/api/config/routes/api_routes.rb#L74

We can then clear out the inconsistencies in the code and only use `param[:name]`. In the end, I wonder if those changes are worth the time. I understand if you don't want to go this far.

dmarcoux:
Instead of `not needed` in the commit message, I suggest using `unneeded`

eduardoj:
@dmarcoux, I applied some of your sugestions in the existing commit, and added a couple of commits more:
- also following your suggestion, to use only `param[:name]`
- the last commit, to remove a line that was not used.

> We can then clear out the inconsistencies in the code and only use `param[:name]`. In the end, I wonder if those changes are worth the time. I understand if you don't want to go this far.

Let's do it, and hope for the best! :facepalm:

Please have a look again.

danidoni (approved):

dmarcoux (approved):

PR #10629, opened 2021-01-14
Adjust the rollout tasks

saraycp:
Most of the rollout tasks like "all_on" or "from_beta" excluded the staff users. Semantically speaking it was confusing. As we no longer need that restriction, the staff are not excluded anymore.

We also add tasks to move the anonymous user (_nobody_) into or out of the rollout program. Something that is usually needed at the beginning or end of the rollout process.

This PR comes with an update of the developer documentation: [Steps for Rolling Out](https://github.com/openSUSE/open-build-service/wiki/Feature-Toggles-(Flipper)#steps-for-rolling-out)

-----------------

* [x] Proper documentation or announcement has to be published upfront since the introduced changes can confuse the users.

### DIFF
--- a/src/api/lib/tasks/rollout.rake
+++ b/src/api/lib/tasks/rollout.rake
@@ -1,35 +1,29 @@
 # rubocop:disable Rails/SkipsModelValidations
+
+# Follow the recommendations for the rollout in:
+# https://github.com/openSUSE/open-build-service/wiki/Feature-Toggles#steps-for-rolling-out
+
 namespace :rollout do
   desc 'Move all the users to rollout program'
   task all_on: :environment do
     User.all_without_nobody
-        .not_staff
         .where(in_rollout: false).in_batches.update_all(in_rollout: true)
   end
 
   desc 'Move all the users out of the rollout program'
   task all_off: :environment do
     User.all_without_nobody
-        .not_staff
         .where(in_rollout: true).in_batches.update_all(in_rollout: false)
   end
 
   desc 'Move the users already in beta to rollout program'
   task from_beta: :environment do
-    User.not_staff
-        .where(in_beta: true, in_rollout: false).in_batches.update_all(in_rollout: true)
-  end
-
-  desc 'Move the members of groups to rollout program'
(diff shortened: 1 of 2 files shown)


src/api/lib/tasks/rollout.rake
```diff
@@ -1,47 +1,49 @@
 # rubocop:disable Rails/SkipsModelValidations
+
+# USAGE NOTES:
```
dmarcoux, line 3: I think we should point to the wiki page below instead of having these usage notes:

https://github.com/openSUSE/open-build-service/wiki/Feature-Toggles-%28Flipper%29#steps-for-rolling-out

This will remove the documentation duplication. It prevents us from having outdated documentation here or in the wiki.

saraycp: I decided not to do so because the Wiki documentation is so changeable that, sooner or later, we'll end up with a broken link.

dmarcoux: I also thought about that when writing my comment. Let's see what others think. I'm fine if we leave it like this.

danidoni: I think that if we have a wiki, we have to maintain it. If it has broken links, we fix them. But if we add documentation in the code then we have two different sources, and then we have to maintain both (otherwise we'll have documentation saying different things, which IMO is worse that having no documentation).
So for me, +1 for having it on the wiki.

vpereira: I think as well that we should link to the Wiki, however we should rename this page. Flipper is implementation detail and shouldn't be in the title. The title should be something like "Features rollout" or "How to rollout new features"

dmarcoux: Alright, so it seems like we agree on having a link to the wiki.

saraycp: Comment updated pointing to the wiki. Thanks for giving your opinion. :bouquet:

dmarcoux (approved):

danidoni (approved):

PR #10633, opened 2021-01-15
Document person show endpoint

danidoni:
The OpenAPI documentation for the `person#show` action.

-----------------

### DIFF
--- a/src/api/public/apidocs-new/paths/person.yaml
+++ b/src/api/public/apidocs-new/paths/person.yaml
@@ -1,0 +1,40 @@
+get:
+  summary: List all people.
+  description: >
+    List all people.
+  security:
+    - basic_authentication: []
+  parameters:
+    - in: query
+      name: prefix
+      schema:
+        type: string
+      required: false
+      description: A prefix to filter the people to look for
+      example: 'Adm'
+  responses:
+    '200':
+      description: >
+        OK. The request has succeeded.
+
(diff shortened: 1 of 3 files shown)


src/api/public/apidocs-new/paths/person.yaml
```diff
@@ -0,0 +1,27 @@
+get:
+  summary: List all persons.
```
dmarcoux, line 2: The plural of _person_ is **people**, not _persons_.

dmarcoux (commented):
There is a typo in the commit message: _endpoin**t**_ instead of _endpoing_

src/api/public/apidocs-new/paths/person.yaml
danidoni: So true :facepalm: Fixed

src/api/public/apidocs-new/paths/person.yaml
```diff
@@ -0,0 +1,27 @@
+get:
+  summary: List all people.
+  description: >
+    List all persons.
```
dmarcoux, line 4: Power to the people! :wink:

danidoni: True also, fixed

danidoni:

eduardoj (approved):
LGTM. I would merge the commits into the first one.

dmarcoux (approved):

PR #10635, opened 2021-01-15
Document '/distributions' API routes with OpenAPI

jcronenberg:
This adds the "distributions" group to the new OpenAPI documentation.
Note that I found the old documentation to be at times outdated. So I tried to also update the documentation with more up to date information.

To verify the new addition to the documentation:
- Start the development environment
- Go to http://localhost:3000/apidocs-new/

distributions

### DIFF
--- a/src/api/public/apidocs-new/paths/distributions.yaml
+++ b/src/api/public/apidocs-new/paths/distributions.yaml
@@ -1,0 +1,82 @@
+get:
+  summary: List all distributions.
+  description: |
+    List all distributions that can be build against.
+    This will not list distributions that are available via [interconnect](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.concepts.html#id-1.5.10.3.5).
+    Check /distributions/include_remotes for this.
+  security:
+    - basic_authentication: []
+  responses:
+    '200':
+      $ref: '../components/responses/distributions.yaml'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+  tags:
+    - Distributions
+
+# FIXME: This endpoint doesn't work. It should be fixed before we document it.
+#        It's not documented in the non-Swagger API documentation.
+#        PUT /distributions would do the same (while also deleting all existing distributions). This is confusing!
+#
+# post:
+#   summary: Create distributions.
+#   description: |
+#     Create one or more distributions.
+
+#     This is only for admins.
+#   security:
+#     - basic_authentication: []
+#   requestBody:
+#     description: Distribution definition
+#     content:
+#       application/xml; charset=utf-8:
+#         schema:
+#           $ref: '../components/schemas/distributions.yaml'
+#   responses:
+#     '200':
+#       $ref: '../components/responses/distributions.yaml'
+#     '401':
+#       $ref: '../components/responses/unauthorized.yaml'
+#   tags:
+#     - Distributions
+
+put:
+  summary: Create distributions while deleting all existing distributions.
+  description: |
+    Create one or more distributions. Existing distributions will be deleted.
+
+    This is only for admins.
+  security:
+    - basic_authentication: []
+  requestBody:
+    description: Distributions definition
+    content:
+      application/xml; charset=utf-8:
+        schema:
+          $ref: '../components/schemas/distributions.yaml'
+  responses:
+    '200':
+      description: |
+        OK. The Request has succeded.
+
+        XML Schema used for body validation: [distributions.rng](../schema/distributions.rng)
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+          example:
+            code: 'ok'
+            summary: 'Ok'
+    '400':
+      description: Bad Request.
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+          example:
+            code: 'validation_failed'
+            summary: 'distributions validation error: 40:0: ERROR: Expecting an element name, got nothing'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+  tags:
+    - Distributions
--- a/src/api/public/apidocs-new/paths/distributions_distribution_id.yaml
+++ b/src/api/public/apidocs-new/paths/distributions_distribution_id.yaml
@@ -1,0 +1,64 @@
+get:
+  summary: Show a distribution.
+  description: |
+    Show a distribution that can be build against.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/distribution_id.yaml'
+  responses:
+    '200':
+      description: |
+        OK. The request has succeded.
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/distributions.yaml'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      description: Not found.
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+          example:
+            code: 'not_found'
+            summary: "Couldn't find Distribution with 'id'=0"
+  tags:
+    - Distributions
+
+delete:
+  summary: Delete a distribution.
+  description: |
+    Delete a distribution.
+
+    This is only for admins.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/distribution_id.yaml'
+  responses:
+    '200':
+      description: |
+        OK. The request has succeded.
+      content:
+        application/xml; charset=utf-8:
+          schema:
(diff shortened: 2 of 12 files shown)


src/api/public/apidocs-new/components/schemas/distribution_hash.yaml
```diff
@@ -0,0 +1,42 @@
+type: object
+properties:
+  id:
+    type: integer
```
jcronenberg, line 4: Right now if you request a distribution with an id e.g. /distributions/16626 you get a response that includes the following:  
```<id type="integer">16626</id>```
Adding this `type` attribute however seems to be impossible with OpenAPI as it doesn't allow integers to have xml attributes like an object. And an object can't have a value. As @eduardoj pointed out this `type` attribute is unnecessary anyway as the API doesn't really need to specify the types.

**Comment by jcronenberg** [comment]
[Distributions](https://trello.com/c/JYUW5VdQ/1600-distributions)

File: src/api/public/apidocs-new/components/schemas/distribution_hash.yaml (hunk 1 above)
Comment (hennevogel) [reply src/api/public/apidocs-new/components/schemas/distribution_hash.yaml:4]: This is the case because we 'just' render the pure object in the controller. Instead of writing a view.

File: src/api/public/apidocs-new/paths/distributions.yaml [hunk 2]
```diff
@@ -0,0 +1,117 @@
...
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+  tags:
+    - Distributions
+
+put:
+  summary: Write all base distributions hosted on this OBS instace.
+  description: |
+    This Endpoint can be used to overwrite the list of all distributions hosted on this OBS instance.
+    Note this overwrites everything, it can't be used to add an additional distribution
```
hennevogel, line 82: I don't understand. You want to say that this is for updating the attributes of all distributions but you can't add a new one?

src/api/public/apidocs-new/paths/distributions.yaml
```diff
@@ -0,0 +1,117 @@
+get:
+  summary: List all base distributions hosted on this OBS instance.
+  description: |
+    Get a list of all base distributions hosted on this OBS instace.
```
hennevogel, line 4:
```suggestion
    List all distributions that can be build against.
```

src/api/public/apidocs-new/paths/distributions.yaml
```diff
@@ -0,0 +1,117 @@
...
+    This will not list Distributions hosted remotely.
```
hennevogel, line 5:
```suggestion
    This will not list distributions that are available via [interconnect](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.concepts.html#id-1.5.10.3.5).
```

src/api/public/apidocs-new/paths/distributions_distribution_id.yaml
```diff
@@ -0,0 +1,58 @@
+get:
+  summary: Get data of one base distribution hosted on this OBS instance.
+  description: |
+    Get data of one base distribution hosted on this OBS instance.
```
hennevogel, line 4:
```suggestion
    Show a distribution that can be build against.
```

src/api/public/apidocs-new/paths/distributions.yaml
```diff
@@ -0,0 +1,117 @@
+get:
+  summary: List all base distributions hosted on this OBS instance.
```
hennevogel, line 2:
```suggestion
  summary: List all distributions.
```

src/api/public/apidocs-new/paths/distributions_distribution_id.yaml
```diff
@@ -0,0 +1,58 @@
+get:
+  summary: Get data of one base distribution hosted on this OBS instance.
```
hennevogel, line 2:
```suggestion
  summary: Show a distribution.
```

src/api/public/apidocs-new/paths/distributions_distribution_id.yaml
```diff
@@ -0,0 +1,58 @@
...
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+          example:
+            code: 'not_found'
+            summary: "Couldn't find Distribution with 'id'=0"
+  tags:
+    - Distributions
+
+delete:
+  summary: Delete one base distribution hosted on this OBS instance.
```
hennevogel, line 30:
```suggestion
  summary: Delete a distribution.
```

src/api/public/apidocs-new/paths/distributions_distribution_id.yaml
```diff
@@ -0,0 +1,58 @@
...
+  description: |
+    Delete one base distribution hosted on this OBS instance.
```
hennevogel, line 32:
```suggestion
    Delete a distribution.
```

src/api/public/apidocs-new/paths/distributions_include_remotes.yaml
```diff
@@ -0,0 +1,74 @@
+get:
+  summary: List all base distributions.
+  description: |
+    Get a list of all base distributions hosted on this OBS instace and also remotely.
```
hennevogel, line 4:
```suggestion
    List all distributions that can be build against, including the one provided by the interconnect.
```

src/api/public/apidocs-new/paths/distributions_include_remotes.yaml
```diff
@@ -0,0 +1,74 @@
+get:
+  summary: List all base distributions.
```
hennevogel, line 2:
```suggestion
  summary: List all distributions.
```

hennevogel (changes requested):

src/api/public/apidocs-new/paths/distributions.yaml
jcronenberg: Well i guess you can of course add a new one but in that process you overwrite all existing ones. The point I wanted to make was that even if the xml contains the existing distributions it still overwrites them and e.g. gives them a new id.

jcronenberg: I guess just writing that it overwrites everything would be better because, yes in theory you could of course add an additional distribution.

src/api/public/apidocs-new/paths/distributions_include_remotes.yaml
```diff
@@ -0,0 +1,74 @@
...
+    '200':
+      description: |
+        OK. The request has succeeded.
+
+        XML Schema used for body validation: [distributions.rng](../schema/distributions.rng)
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/distributions.yaml'
+          example:
```
krauselukas, line 17: @jcronenberg this is a pretty big example, since you have this 200 response two times (also in the `GET /distributions`), with the same example, I would move it to its own file in the responses folder and then reference it. A little less repeated code :)

src/api/public/apidocs-new/paths/distributions_distribution_id.yaml
```diff
@@ -0,0 +1,58 @@
+get:
+  summary: Show a distribution.
+  description: |
+    Show a distribution that can be build against.
+  security:
```
krauselukas, line 5: @jcronenberg you are missing the parameter part here, you have to add the `distribution_id`

krauselukas: here an example https://github.com/openSUSE/open-build-service/blob/3853a4ce3d7311846529308bbed1d587d4ba9781/src/api/public/apidocs-new/paths/person_login_token.yaml#L6

src/api/public/apidocs-new/paths/distributions_distribution_id.yaml
```diff
@@ -0,0 +1,58 @@
...
+            code: 'not_found'
+            summary: "Couldn't find Distribution with 'id'=0"
+  tags:
+    - Distributions
+
+delete:
+  summary: Delete a distribution.
+  description: |
+    Delete a distribution.
+  security:
```
krauselukas, line 33: same here, you the parameter part is missing

krauselukas:
@jcronenberg besides the last two comments, it look good to me. Nice job!

hennevogel (approved):

hennevogel (changes requested):
You dropped the update route on purpose?

jcronenberg:
> You dropped the update route on purpose?

I dropped the "POST /distributions/<distribution_id>" route because it doesn't exist as far as I can tell. But I just saw that there is now a "POST /distributions" route which I overlooked.

dmarcoux:
I took over since @jcronenberg is at school now. I pushed changes and added a FIXME comment which includes the documentation for the _POST /distributions_ endpoint. Let me know what you think.

krauselukas (approved):

hennevogel (approved):

PR #10637, opened 2021-01-15
Add OpenAPI documentation for request endpoints

krauselukas:

-----------------


krauselukas:
[Request](https://trello.com/c/Pxazglqu/1599-request)

src/api/public/apidocs-new/paths/request_request_diff.yaml
```diff
@@ -0,0 +1,62 @@
...
+  description: |
+    Get the diff for all by the request affected packages
+  security:
+    - basic_authentication: []
+  parameters:
+    - in: query
+      name: diff_to_superseded
+      type: integer
+      description: | 
+        Diff relative to a given superseded request. State the id of the corrsponding superseded request.
```
danidoni, line 12:
```suggestion
        Diff relative to a given superseded request. State the id of the corresponding superseded request.
```

src/api/public/apidocs-new/paths/request_request_diff.yaml
```diff
@@ -0,0 +1,62 @@
...
+      example: 10401
+    - in: query
+      name: view
+      schema:
+        type: string
+        enum:
+          - xml
+      description: Set this parameter to xml in order to receive a structered diff instead of plain text.
```
danidoni, line 20:
```suggestion
      description: Set this parameter to xml in order to receive a structured diff instead of plain text.
```

src/api/public/apidocs-new/paths/request_request_diff.yaml
krauselukas: done

src/api/public/apidocs-new/paths/request_request_diff.yaml
krauselukas: done

dmarcoux:
We should add a note to the endpoints which are running expensive operations. The endpoint `POST /request/{id}?cmd=diff` is for sure one of them. I believe `GET /request?view=collection` could also be marked as such.

src/api/public/apidocs-new/paths/request_request_diff.yaml
```diff
@@ -0,0 +1,63 @@
+post:
+  summary: Get the diff for all by the request affected packages
```
dmarcoux, line 2: I don't understand this sentence. Is this `Get the diff for all packages affected in the request`? I'm not sure.

krauselukas: You are right, the sentence is a bit off, will change it

krauselukas:
> We should add a note to the endpoints which are running expensive operations. The endpoint `POST /request/{id}?cmd=diff` is for sure one of them. I believe `GET /request?view=collection` could also be marked as such.

Yes, might be a  good idea to mention it, since its also mentioned in the general description of the API

hennevogel:
> Yes, might be a good idea to mention it, since its also mentioned in the general description of the API

No please. Calling an inexpensive operation 300 times concurrently is wrong the same way. Let's keep this simple.

src/api/public/apidocs-new/paths/request_request_diff.yaml
krauselukas: I applied your suggestion :)

src/api/public/apidocs-new/paths/request_request_id.yaml
```diff
@@ -25,3 +25,163 @@ get:
...
+  responses:
+    '200':
+      description: OK
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+          example:
+            code: ok
+            summary: Ok
```
eduardoj, line 155:
```suggestion
      $ref: '../components/responses/succeeded.yaml'
```

src/api/public/apidocs-new/paths/request_request_id.yaml
```diff
@@ -185,3 +185,50 @@ post:
                 summary: Couldn't find request with id '120'
   tags:
     - Request
+
+put:
+  summary: Modify a given request.
+  description: |
+    Modify a given request. **NOTE:** Certain parts of a request can only be changed by admins.
```
eduardoj, line 192: We should add the reference to the schema: "XML Schema used for body validation: request.rng".

src/api/public/apidocs-new/paths/request_request_id.yaml
```diff
@@ -185,3 +185,50 @@ post:
...
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/request.yaml#/RequestId'
+  requestBody:
+    description: |
+      Request XML
+      **NOTE:** You cannot pass the request id in the xml. You can pass the history in the xml, but it will be simply ignored.
+    required: true
```
eduardoj, line 201: We should add the reference to the schema: "XML Schema used for body validation: request.rng".

src/api/public/apidocs-new/paths/request_request_id.yaml
```diff
@@ -232,3 +232,36 @@ put:
...
+  responses:
+    '200':
+      description: OK
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+          example:
+            code: ok
+            summary: Ok
```
eduardoj, line 252:
```suggestion
      $ref: '../components/responses/succeeded.yaml'
```

src/api/public/apidocs-new/paths/request.yaml
```diff
@@ -0,0 +1,16 @@
...
+  description: Get a simple directory listing of all requests
+  security:
+    - basic_authentication: []
+  responses:
+    '200':
+      description: OK
+      content:
+        application/xml:
+          schema:
+            $ref: '../components/schemas/request/directory.yaml'
```
eduardoj, line 12: We could use the existing `../components/schemas/directory.yaml`.

src/api/public/apidocs-new/paths/request.yaml
```diff
@@ -7,10 +7,85 @@ get:
...
           schema:
             $ref: '../components/schemas/request/directory.yaml'
     '401':
       $ref: '../components/responses/unauthorized.yaml'
   tags:
     - Request
+
+post:
+  summary: Create a new request
+  description: Create a new request
```
eduardoj, line 20: We can remove 'new'. It sounds redundant.

eduardoj (changes requested):
Huge work! Thanks!

src/api/public/apidocs-new/paths/request.yaml
krauselukas: @eduardoj there is a slight difference between those schemas, the request one has the an additonal `count` attribute

eduardoj: The `components/schemas/directory.yaml` schema doesn't require to define the `count` attribute. So that schema can still be used.

In this case, we could provide an example without the `count` attribute:

```yaml
       content:
        application/xml:
          schema:
            $ref: '../components/schemas/request/directory.yaml'
          example:                                                                                                                                       
            entry:                                                                                                                                       
              - name: Some text                                                                                                                          
```

krauselukas: ahh okay, thanks

krauselukas:
@eduardoj I addressed your review, please have another look :)

src/api/public/apidocs-new/components/parameters/request.yaml
```diff
@@ -0,0 +1,82 @@
...
+  name: withfullhistory
+  schema:
+    type: string
+  description: Includes both, request and review history in the results.
+  example: 1
+Limit:
+  in: query
+  name: limit
+  schema:
+    type: int
```
eduardoj, line 65:
```suggestion
    type: integer
```
`int` seems not to be a basic data type.

eduardoj (approved):
> @eduardoj I addressed your review, please have another look :)

Looks good to me. Thanks a lot!

PR #10643, opened 2021-01-18
Add general documentation to API new documentation

dmarcoux:
Is something missing?

Preview:
preview-api-doc

### DIFF
--- a/src/api/public/apidocs-new/OBS-v2.10.50.yaml
+++ b/src/api/public/apidocs-new/OBS-v2.10.50.yaml
@@ -8,7 +8,18 @@ tags:
   - name: Groups
 
 info:
-  description: Open Build Service API
+  description: |
+    The _Open Build Service API_ is a XML API.
+
+    To authenticate, use [HTTP basic authentication](https://en.wikipedia.org/wiki/Basic_access_authentication) by passing the _Authorization_ header in the form of `Authorization: Basic <credentials>`.
+
+    There is no API versioning as there is no need for it right now.
+
+    Only rudimentary rate limiting is implemented, so please be gentle when using the API concurrently, especially with potentially expensive operations.
+    In case of abuse, we will limit/remove your access.
+
+    For command-line users, we recommend using [osc](https://github.com/openSUSE/osc) with its _api_ command to interact with the API.
+    It's as simple as this example: `osc api /about` (_about_ is one of the endpoints documented below)
   version: '2.10.50'
   title: Open Build Service API
   contact:


dmarcoux:
[General Documentation](https://trello.com/c/rW4CFRPw/1593-general-documentation)

eduardoj:
> Is something missing?

Nop from my side. It's what we have now.

eduardoj (approved):

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -8,7 +8,18 @@ tags:
...
 
 info:
-  description: Open Build Service API
+  description: |
+    The _Open Build Service API_ is a XML API.
+    This documentation is based on the _OpenAPI 3.0.0_ specification and powered by _Swagger_.
+
+    To authenticate, use [HTTP basic authentication](https://en.wikipedia.org/wiki/Basic_access_authentication) by passing the _Authorization_ header in the form of `Authorization: Basic <credentials>`.
+
+    We don't have any versioning.
```
adrianschroeter, line 17: sure we have.

Or do you really speak about some complete new api I never heard about?

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -8,7 +8,18 @@ tags:
...
+
+    We don't have rate limiting, so be conscious about how you use the API.
+    We do monitor the API, so abuse will lead to a warning, then a ban if it continues.
```
adrianschroeter, line 20: Why do you write this so public? We have some rate limiting, but we shouldn't motivate more people to missuse it by promoting it...

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -8,7 +8,18 @@ tags:
...
+
+    For command-line users, we recommend using [osc](https://github.com/openSUSE/osc) with its _api_ command to interact with the API.
```
adrianschroeter, line 22: what is "its _api_" ? another one?

adrianschroeter (changes requested):

src/api/public/apidocs-new/OBS-v2.10.50.yaml
dmarcoux: My intention is to point users to `osc api`. I use the same terms as the manpage for `osc` and the output of `osc --help`, which are both listing `api` as one of the commands available for `osc`. I could add an example to make it clearer. Would this help?

src/api/public/apidocs-new/OBS-v2.10.50.yaml
dmarcoux: I'm writing about the existing API. Could you point me to the versioning that we have? Because I don't see any implementation of that.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
dmarcoux: It's common to have rate limiting in APIs, so we should clearly state what we have.

I don't see any rate limiting in place for the API, at least in Rails application. Is it implemented in the backend?

saraycp: I aggre with @adrianschroeter here. I don't think we need to state this so clearly, it could tempt people to missuse it.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
adrianschroeter: check for the /source/ routes, you can specify the revision via the "rev" CGI parameter. This works for everything which is considered to be sources, this includes also meta data like project and package _meta files or attributes.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
adrianschroeter: ah, misreaded it, sorry. Yes, an example would help and maybe point out that only the path part is needed unlike with curl. An easy example could be

 osc api /about

src/api/public/apidocs-new/OBS-v2.10.50.yaml
hennevogel: The problem we frequently see, like yesterday evening for instance, is that good-will users that don't know the conditions and assume that we care about guarding against this, trigger grave problems for everyone. Just by doing things that make sense. Documenting this will help avoid some of those situations. Documenting the conditions will not turn them from good-will people into people with bad intentions.

On the other hand, if you have someone that comes to the API with bad intentions, not knowing the conditions will not stop them to be bad. It will also not make it in any way harder for them to do so.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
dmarcoux: I am referring to API versioning. The only version of the API which is available, is the latest one. This is why I'm writing that we don't have any versioning. To be extra clear, I'll add _API_ in front of _versioning_.

What you are referring to, in the _/sources/_ routes, is about requesting a specific version of a source.

dmarcoux:
@adrianschroeter Let me know if your concerns were addressed. I believe that they are.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
vpereira: @dmarcoux check this https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/application_controller.rb#L23

```
 def add_api_version
    response.headers['X-Opensuse-APIVersion'] = (CONFIG['version']).to_s
  end
```

vpereira: for me it looks like a version. We add it to every API request...

dmarcoux: I did see that. But it's not API versioning. It's telling clients what the version of the API, which is always referring to the latest version deployed in production.

If we had API versioning, it could be something like below. It's just an example:
- https://build.opensuse.org/v1/about (old version which isn't changed, except on rare occasions like fixing a security vulnerability)
- https://build.opensuse.org/v2/about (current version which introduced breaking changes from v1, not actively developed)
- https://build.opensuse.org/v3/about (new version which will introduce breaking changes from v2, actively developed and not released)

adrianschroeter: k, we do kind of "versioning" by introducing alternative routes where needed. But this is only needed in rare cases.

Anyway, I would not mention something on the frontpage what is not there. It is imho way to prominent there and we should focus on what is there. Maybe giving a highlevel overview what are the /source /build or /attribute areas for example.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
vpereira: why we don't have rate limiting in the first place? it could be configured by the administrator. IMO a better solution than document it.

dmarcoux: I'm all for rate limiting, but this is out of scope for this PR.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
dmarcoux: API versioning is quite common nowadays and it's good to be clear that our API doesn't have this. As for the documentation specific to endpoints like _/source_ and others, we're going to tackle that in the future. It's going to be like the endpoints which are already documented with Swagger.

adrianschroeter:
On Dienstag, 19. Januar 2021, 13:14:28 CET Dany Marcoux wrote:
> @dmarcoux commented on this pull request.
> 
> 
> 
> > @@ -8,7 +8,18 @@ tags:
>    - name: Groups
>  
>  info:
> -  description: Open Build Service API
> +  description: |
> +    The _Open Build Service API_ is a XML API.
> +    This documentation is based on the _OpenAPI 3.0.0_ specification and powered by _Swagger_.
> +
> +    To authenticate, use [HTTP basic authentication](https://en.wikipedia.org/wiki/Basic_access_authentication) by passing the _Authorization_ header in the form of `Authorization: Basic <credentials>`.
> +
> +    We don't have any versioning.
> 
> API versioning is quite common nowadays and it's good to be clear that our API doesn't have this. As for the documentation specific to endpoints like _/source_ and others, we're going to tackle that in the future. It's going to be like the endpoints which are already documented with Swagger.

then please turn it around into something positive: We have a long time stable api which does not require api versioning.

-- 

dmarcoux:
Writing that we don't have API versioning is not implying that the API isn't stable. I've rephrased the sentence about API versioning:

_We don't have any API versioning as we don't have a need for it right now._

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -8,7 +8,19 @@ tags:
   - name: Groups
 
 info:
-  description: Open Build Service API
+  description: |
+    The _Open Build Service API_ is a XML API.
+    This documentation is based on the _OpenAPI 3.0.0_ specification and powered by _Swagger_.
```
hennevogel, line 13: Who cares? :smile:

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -8,7 +8,19 @@ tags:
...
+
+    To authenticate, use [HTTP basic authentication](https://en.wikipedia.org/wiki/Basic_access_authentication) by passing the _Authorization_ header in the form of `Authorization: Basic <credentials>`.
+
+    We don't have any API versioning as we don't have a need for it right now.
+
+    We don't have rate limiting, so be conscious about how you use the API.
```
hennevogel, line 19:
```suggestion
    We only implement rudimentary rate limiting, so please be gentle when using the API concurrently, especially with potentially expensive operations. In case of abuse we will limit/remove your access.
```

hennevogel (changes requested):

src/api/public/apidocs-new/OBS-v2.10.50.yaml
dmarcoux: I guess _rudimentary_ is the right word here... I like the second part of your suggestion.

dmarcoux:
Please have another look @hennevogel and @adrianschroeter

hennevogel (approved):

krauselukas (commented):
@dmarcoux LGTM, I would just rephrase the commit message a bit, maybe something like `Add general documentation to new OpenAPI docs`

krauselukas (approved):

adrianschroeter:
I would drop my review in the github box, but no clue how to do that ... just ignore it please.

PR #10646, opened 2021-01-18
Document person command action

danidoni:
The OpenAPI documentation for the `person#command` action.

You can test it in [our review-app](https://obs-reviewlab.opensuse.org/danidoni-document-person-command-action/apidocs-new/#/Person/post_person)

-----------------

### DIFF
--- a/src/api/public/apidocs-new/components/responses/unknown_command_error.yaml
+++ b/src/api/public/apidocs-new/components/responses/unknown_command_error.yaml
@@ -1,0 +1,17 @@
+description: |
+  Bad request.
+
+  XML Schema used for body validation: [status.xsd](../schema/status.xsd)
+content:
+  application/xml; charset=utf-8:
+    schema:
+      $ref: '../schemas/api_response.yaml'
+    examples:
+      Unknown Command:
+        value:
+          code: unknown_command
+          summary: "Allowed command is 'register'."
+      Bad Request:
+        value:
+          code: err_register_save
+          summary: "Could not save the registration, details: Login is the name of an already existing user and Password can't be blank."
--- a/src/api/public/apidocs-new/components/schemas/unregistered_person.yaml
+++ b/src/api/public/apidocs-new/components/schemas/unregistered_person.yaml
@@ -1,0 +1,16 @@
+type: object
+properties:
+  login:
+    type: string
+  realname:
+    type: string
+  email:
+    type: string
+  password:
+    type: string
+  note:
+    type: string
+  state:
+    type: string
+xml:
+  name: unregisteredperson
--- a/src/api/public/apidocs-new/paths/person.yaml
+++ b/src/api/public/apidocs-new/paths/person.yaml
@@ -38,3 +38,33 @@ get:
       $ref: '../components/responses/unauthorized.yaml'
   tags:
     - Person
+
+post:
+  summary: Executes a command on the person endpoint.
+  description: |
+    Allows executing command on the person endpoint.
+
+    As of now, the only command allowed is 'register'.
+  parameters:
+    - in: query
+      name: cmd
+      schema:
+        type: string
+        enum: [register]
+      required: true
+      description: The command to execute
+      example: register
+  requestBody:
+    description: The data for the person to register.
+    required: true
+    content:
+      application/xml:
+        schema:
+          $ref: '../components/schemas/unregistered_person.yaml'
+  responses:
+    '200':
+      $ref: '../components/responses/succeeded.yaml'
+    '400':
+      $ref: '../components/responses/unknown_command_error.yaml'
+  tags:
+    - Person


src/api/public/apidocs-new/paths/person.yaml
```diff
@@ -0,0 +1,33 @@
+post:
+  summary: Executes a command on the person endpoint.
+  description: >-
+    Allows executing command on the person endpoint.
+    As of now, the only command allowed is 'register', even if it says that the only allowed command is 'change_password'.
```
danidoni, line 5: See [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/person_controller.rb#L29) for details.

danidoni:

dmarcoux:
@danidoni There is a conflict

danidoni:
> @danidoni There is a conflict

Fixed, thanks!

src/api/public/apidocs-new/paths/person.yaml
dmarcoux: I know that this PR is only about documenting the current state of this endpoint, but I would be in favor of changing this line in the command endpoint:

https://github.com/openSUSE/open-build-service/blob/cc9879e0541feb708cbcf053d0433eb8d1197dc5/src/api/app/controllers/person_controller.rb#L33

It would be `raise UnknownCommandError, "Allowed commands are 'register'"`. We could get rid of this confusing part `even if it says that the only allowed command is 'change_password'` in the documentation and code.

hennevogel: Yes, do the right thing. This doesn't cost anything...

danidoni: Changed. The error message will be fixed [here](https://github.com/openSUSE/open-build-service/pull/10666)

src/api/public/apidocs-new/paths/person.yaml
```diff
@@ -38,3 +38,35 @@ get:
...
+  parameters:
+    - in: query
+      name: cmd
+      schema:
+        type: string
+      required: true
+      description: The command to execute
+      example: register
+  requestBody:
+    description: Optional description in *Markdown*
```
eduardoj, line 58: Leftover?

src/api/public/apidocs-new/paths/person.yaml
```diff
@@ -38,3 +38,35 @@ get:
...
+    As of now, the only command allowed is 'register'.
+  security:
+    - basic_authentication: []
+  parameters:
+    - in: query
+      name: cmd
+      schema:
+        type: string
+      required: true
+      description: The command to execute
```
eduardoj, line 55: It should be clear here, in the definition of the parameter, the possible options this parameter could have. Adding something like "The only command allowed is 'register'.", could be enough. Using an enum type, with only one value, would also help.

src/api/public/apidocs-new/paths/person.yaml
```diff
@@ -38,3 +38,35 @@ get:
       $ref: '../components/responses/unauthorized.yaml'
   tags:
     - Person
+
+post:
+  summary: Executes a command on the person endpoint.
+  description: >-
+    Allows executing command on the person endpoint.
+    As of now, the only command allowed is 'register'.
```
eduardoj, line 46:
```suggestion
    Allows executing command on the person endpoint.
    
    As of now, the only command allowed is 'register'.
```
These two lines show as one in the Swagger UI. Adding an empty line between them will fix it.

src/api/public/apidocs-new/components/responses/unknown_command_error.yaml
```diff
@@ -0,0 +1,19 @@
...
+content:
+  application/xml; charset=utf-8:
+    schema:
+      $ref: '../schemas/api_response.yaml'
+    examples:
+      unknown_command_error:
+        value:
+          code: unknown_command
+          summary: "Allowed commands are 'register'."
+        summary: Unknown command error
```
eduardoj, line 14:
```suggestion
      Unknown Command:
        value:
          code: unknown_command
          summary: "Allowed commands are 'register'."
```
Nitpick: I find adding "error" or "_error" kind of redundant, in this context. We can also get rid of the last `summary` tag if we change the name of the example.

src/api/public/apidocs-new/components/responses/unknown_command_error.yaml
```diff
@@ -0,0 +1,19 @@
...
+      bad_request_error:
+        value:
+          code: err_register_save
+          summary: "Could not save the registration, details: Login is the name of an already existing user and Password can't be blank."
+        summary: Bad Request Error
```
eduardoj, line 19:
```suggestion
      Bad Request:
        value:
          code: err_register_save
          summary: "Could not save the registration, details: Login is the name of an already existing user and Password can't be blank."
```
Nitpick: same as above.

eduardoj (changes requested):
Only a couple of minor issues, and nitpicks. :+1:

src/api/public/apidocs-new/paths/person.yaml
```diff
@@ -38,3 +38,35 @@ get:
...
+      required: true
+      description: The command to execute
+      example: register
+  requestBody:
+    description: Optional description in *Markdown*
+    required: true
+    content:
+      application/xml:
+        schema:
+          $ref: '../components/schemas/unregisteredperson.yaml'
```
eduardoj, line 63:
```suggestion
          $ref: '../components/schemas/unregistered_person.yaml'
```
Nitpick: adding a `_` will improve readibility.

src/api/public/apidocs-new/paths/person.yaml
danidoni: Totally. I'll change it

src/api/public/apidocs-new/paths/person.yaml
danidoni: I've used the union type

danidoni:
> Only a couple of minor issues, and nitpicks. +1

Fixed

src/api/public/apidocs-new/paths/person.yaml
danidoni: I mean `enum`, not `union`

danidoni:
I've removed the requirement to log in due to [this](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/person_controller.rb#L11)

saraycp (approved):
LGTM

src/api/public/apidocs-new/components/responses/unknown_command_error.yaml
```diff
@@ -0,0 +1,17 @@
...
+  XML Schema used for body validation: [status.xsd](../schema/status.xsd)
+content:
+  application/xml; charset=utf-8:
+    schema:
+      $ref: '../schemas/api_response.yaml'
+    examples:
+      Unknown Command:
+        value:
+          code: unknown_command
+          summary: "Allowed commands are 'register'."
```
dmarcoux, line 13:
```suggestion
          summary: "Allowed command is 'register'."
```

danidoni: True, changing

dmarcoux (approved):

eduardoj (approved):

src/api/public/apidocs-new/components/schemas/unregistered_person.yaml
```diff
@@ -0,0 +1,16 @@
+type: object
+properties:
+  login:
+    type: string
+  realname:
+    type: string
+  email:
+    type: string
```
dmarcoux, line 8: It would be nice to have some value example here as it allows users to try it out in the web interface without having to type anything. Something like this:

https://github.com/openSUSE/open-build-service/blob/822926b893b11957161e9d7413a05a1fe85e1f1a/src/api/public/apidocs-new/components/schemas/configuration.yaml#L51-L53

dmarcoux: I can take care of this.

danidoni: Good catch!

dmarcoux: The follow-up PR is #10692

PR #10648, opened 2021-01-18
Update the list of crawlers

hennevogel:
This was never done after the initial commit. Should be done periodically somehow...

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -68,7 +68,7 @@ class Webui::PackageController < Webui::WebuiController
   end
 
   def show
-    if request.bot?
+    if @spider_bot
       params.delete(:rev)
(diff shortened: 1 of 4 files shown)


danidoni (approved):

danidoni:
I guess the test failures are something unrelated to the changes, can you re-run them?

dmarcoux:
Where is the list coming from? I'm just curious... :slightly_smiling_face:

hennevogel:
`rake voight_kampff:import_user_agents`

vpereira:
could you re-run the feature specs @hennevogel ? Unfortunately I couldn't do it myself.

hennevogel:
Seems like this has an effect on the feature spec.

hennevogel:
@danidoni @dmarcoux I think you want to re-review this

dmarcoux (approved):

PR #10657, opened 2021-01-20
Document person register endpoint

danidoni:
The OpenAPI documentation for the `person#register` action.

You can test it in [our review-app](https://obs-reviewlab.opensuse.org/danidoni-document-person-register-endpoint/apidocs-new/#/Person/post_person_register)

-----------------

### DIFF
--- a/src/api/public/apidocs-new/OBS-v2.10.50.yaml
+++ b/src/api/public/apidocs-new/OBS-v2.10.50.yaml
@@ -58,7 +58,8 @@ paths:
 
   /person:
     $ref: 'paths/person.yaml'
-
+  /person/register:
+    $ref: 'paths/person_register.yaml'
   /person/{login}/token:
     $ref: 'paths/person_login_token.yaml'
 
--- a/src/api/public/apidocs-new/components/schemas/unregisteredperson.yaml
+++ b/src/api/public/apidocs-new/components/schemas/unregisteredperson.yaml
@@ -1,0 +1,16 @@
+type: object
+properties:
+  login:
+    type: string
+  realname:
+    type: string
+  email:
+    type: string
+  password:
+    type: string
+  note:
+    type: string
+  state:
+    type: string
+xml:
+  name: unregisteredperson
--- a/src/api/public/apidocs-new/paths/person.yaml
+++ b/src/api/public/apidocs-new/paths/person.yaml
@@ -2,8 +2,6 @@ get:
   summary: List all people.
   description: >
     List all people.
-  security:
-    - basic_authentication: []
   parameters:
     - in: query
       name: prefix
@@ -34,7 +32,5 @@ get:
               - name: 'user_2'
               - name: 'user_3'
               - name: 'Requestor'
-    '401':
-      $ref: '../components/responses/unauthorized.yaml'
   tags:
     - Person


danidoni:

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -38,6 +38,9 @@ paths:
   /person:
     $ref: 'paths/person.yaml'
 
+  /person/register:
```
krauselukas, line 41: @danidoni the path is `/person?cmd=register`

krauselukas (dismissed):

src/api/public/apidocs-new/OBS-v2.10.50.yaml
krauselukas: the parameter also need to be added to the `person_register.yaml`

dmarcoux:
I wonder if it's needed to document this endpoint since the `command` endpoint documented in #10646 is already doing the exact same thing.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
danidoni: I'm talking about [this](https://github.com/openSUSE/open-build-service/blob/master/src/api/config/routes/api_routes.rb#L23) one...

danidoni:
> I wonder if it's needed to document this endpoint since the `command` endpoint documented in #10646 is already doing the exact same thing.

Well, it's there and it's working (I mean, it's not a fake one). If ppl use it we have to document it. Another think would be to somehow alias it to the #10646 ...

src/api/public/apidocs-new/OBS-v2.10.50.yaml
krauselukas: ohh sorry for the noise, overlooked that there are two endpoint for the same thing.....

krauselukas:
You could also just keep what you already documented, and note that its deprecated like its mentioned in the comment here https://github.com/openSUSE/open-build-service/blob/master/src/api/config/routes/api_routes.rb#L23

krauselukas (approved):

saraycp (changes requested):
Part of this documentation was introduced in a previous PR:
https://github.com/openSUSE/open-build-service/blob/b20f631b04be2e2a8b5b45e60b91e19a6f271e1c/src/api/public/apidocs-new/paths/person_register.yaml#L16

There, a `401` response was documented. However, authorization is not required for this endpoint. Any kind of user, including `nobody` can register a new user.

danidoni:
> Part of this documentation was introduced in a previous PR:
> 
> https://github.com/openSUSE/open-build-service/blob/b20f631b04be2e2a8b5b45e60b91e19a6f271e1c/src/api/public/apidocs-new/paths/person_register.yaml#L16
> 
> There, a `401` response was documented. However, authorization is not required for this endpoint. Any kind of user, including `nobody` can register a new user.

I've removed the requirement to log in due to [this](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/person_controller.rb#L11)

Thanks @saraycp !

saraycp (approved):

PR #10658, opened 2021-01-20
Document 'person/{login}/token' get method

saraycp:
Document get method for 'person/{login}/token' endpoint using OpenAPI.

You can go to http://localhost:3000/apidocs-new/ to check it out.

Co-authored-by: Eduardo Navarro <enavarro@suse.com>

-----------------

If this PR requires any particular action or consideration before deployment,
please check the reasons or add your own to the list:

* [ ] Requires a database migration that can cause downtime. Postpone deployment until maintenance window[1].
* [ ] Contains a data migration that can cause temporary inconsistency, so should be run at a specific point of time.
* [ ] Changes some configuration files (e.g. options.yml), so the changes have to be applied manually in the reference server.
* [ ] A new Feature Toggle[2] should be enabled in the reference server.
* [ ] Proper documentation or announcement has to be published upfront since the introduced changes can confuse the users.

[1] https://github.com/openSUSE/open-build-service/wiki/Deployment-of-build.opensuse.org#when-there-are-migrations
[2] https://github.com/openSUSE/open-build-service/wiki/Feature-Toggles-%28Flipper%29#you-want-real-people-to-test-your-feature

PR #10661, opened 2021-01-20
Document the person login action

danidoni:
The OpenAPI documentation for the `person#login` action.

You can test it in [our review-app](https://obs-reviewlab.opensuse.org/danidoni-document-person-login-action/apidocs-new/#/Person/post_person__user_login__login)

-----------------

### DIFF
--- a/src/api/public/apidocs-new/components/parameters/user_login.yaml
+++ b/src/api/public/apidocs-new/components/parameters/user_login.yaml
@@ -1,0 +1,7 @@
+in: path
+name: user_login
+schema:
+  type: string
+required: true
+description: User login
+example: admin
--- a/src/api/public/apidocs-new/paths/person_login.yaml
+++ b/src/api/public/apidocs-new/paths/person_login.yaml
@@ -1,0 +1,18 @@
+post:
+  summary: Perform a login for some person
+  description: >
+    This is a fake login endpoint (it always responds back with an ok).
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/user_login.yaml'
+  responses:
+    '200':
+      description: >
+        OK. The request has succeeded.
+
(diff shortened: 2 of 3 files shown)


src/api/public/apidocs-new/components/parameters/user_login.yaml
```diff
@@ -0,0 +1,7 @@
+in: path
+name: user_login
+schema:
+  type: string
+required: true
+description: User login
```
hennevogel, line 6: User or person? We should make up our mind.

hennevogel (changes requested):

src/api/public/apidocs-new/paths/person_login.yaml
```diff
@@ -0,0 +1,18 @@
+post:
+  summary: Perform a login for some person
```
dmarcoux, line 2: This line says that it's performing a login, but then the description says that it's a fake login endpoint. Should we even document this endpoint? I would not since it seems quite useless to any API user.

dmarcoux (changes requested):
The summary says that the endpoint performs a login, but then the description says that it's a fake login endpoint.

We should remove this code, not document it. This endpoint was added in October 2013 in the commit 8145e40 and was never changed  since then (it was only moved around). It seems quite useless to any API user.

src/api/public/apidocs-new/components/parameters/user_login.yaml
saraycp: I think I'm going to use "person" and "people" for consistency in the descriptions and this parameter as well.
@danidoni  what about renaming this file to simply `login`? The description can be something like `Person login`.

I'll do the same in #10658.

danidoni:
I'm removing the endpoint [here](https://github.com/openSUSE/open-build-service/pull/10667). This PR does not make sense any more.

PR #10662, opened 2021-01-20
Remove configuration schedulers route for inexistent code

dmarcoux:
In the commit d759517, schedulers was added to the configuration resource in the routes. Its corresponding action was never created in ConfigurationsController. It could also never work since schedulers is not one of the default RESTful actions. It cannot be simply passed to the `only` option like that.

### DIFF
--- a/src/api/config/routes/api_routes.rb
+++ b/src/api/config/routes/api_routes.rb
@@ -6,7 +6,7 @@ OBSApi::Application.routes.draw do
 
     resources :about, only: :index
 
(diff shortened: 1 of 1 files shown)


dmarcoux:
[Configuration](https://trello.com/c/vL9fQtTx/1595-configuration)

danidoni (approved):

dmarcoux:
I pushed again to improve the commit description. I added a bit more details.

eduardoj (approved):

PR #10666, opened 2021-01-21
Fix misleading error message

danidoni:
This PR fixes the misleading error message thrown in [this endpoint](https://github.com/openSUSE/open-build-service/pull/10646)

-----------------

### DIFF
--- a/src/api/app/controllers/person_controller.rb
+++ b/src/api/app/controllers/person_controller.rb
@@ -24,13 +24,12 @@ class PersonController < ApplicationController
     render_ok # just a dummy check for the webui to call (for now)
   end
 
-  # Returns a list of all users (that optionally start with a prefix)
   def command
     if params[:cmd] == 'register'
       internal_register
       return
     end
-    raise UnknownCommandError, "Allowed commands are 'change_password'"
+    raise UnknownCommandError, "Allowed command is 'register'"
   end
 
   def get_userinfo


danidoni:

src/api/app/controllers/person_controller.rb
```diff
@@ -30,7 +30,7 @@ def command
       internal_register
       return
     end
-    raise UnknownCommandError, "Allowed commands are 'change_password'"
+    raise UnknownCommandError, "Allowed commands are 'register'"
```
dmarcoux, line 33: Since there's only one command allowed, I suggest this:
```suggestion
    raise UnknownCommandError, "Allowed command is 'register'"
```

dmarcoux (changes requested):
The comment just above the `command` action is also wrong. I suggest removing it completely.

https://github.com/openSUSE/open-build-service/blob/574979c9f7b42e09bb7b6746ca91df6f4cfbe253/src/api/app/controllers/person_controller.rb#L27

danidoni:
> The comment just above the `command` action is also wrong. I suggest removing it completely.
> 
> https://github.com/openSUSE/open-build-service/blob/574979c9f7b42e09bb7b6746ca91df6f4cfbe253/src/api/app/controllers/person_controller.rb#L27

Changed

src/api/app/controllers/person_controller.rb
danidoni: Fixed

dmarcoux (approved):

krauselukas (approved):

PR #10669, opened 2021-01-21
Correct 'person/<userid>/token' post method in apidocs

saraycp:
The post method does not use the 'cmd' parameter.

The 'project' is only used in combination with the 'package'. This action will never create a token for a specific project, only for a package.

-----------------

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -165,9 +165,8 @@ POST /person/<userid>/token
   Create a new authentication token for this user. It may be limited for a specific package
   via optional project&package parameters.
 
-  cmd: create
-  project: create a token for a specific project
-  package: create a token for a specific package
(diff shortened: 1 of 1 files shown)


docs/api/api/api.txt
```diff
@@ -165,9 +165,8 @@ POST /person/<userid>/token
   Create a new authentication token for this user. It may be limited for a specific package
   via optional project&package parameters.
 
-  cmd: create
-  project: create a token for a specific project
-  package: create a token for a specific package
+  project: together with package, used to create a token for a specific package
+  package: together with project, used create a token for a specific package
```
dmarcoux, line 169:
```suggestion
  package: together with project, used to create a token for a specific package
```

saraycp: fixed

eduardoj (approved):

PR #10672, opened 2021-01-21
Document 'person/{login}/group' endpoint

saraycp:
Using OpenAPI.

-----------------

### DIFF
--- a/src/api/public/apidocs-new/OBS-v2.10.50.yaml
+++ b/src/api/public/apidocs-new/OBS-v2.10.50.yaml
@@ -71,6 +71,8 @@ paths:
     $ref: 'paths/person.yaml'
   /person/register:
     $ref: 'paths/person_register.yaml'
+  /person/{login}/group:
+    $ref: 'paths/person_login_group.yaml'
   /person/{login}/token:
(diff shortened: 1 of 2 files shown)


src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -58,9 +58,10 @@ paths:
 
   /person:
     $ref: 'paths/person.yaml'
-
   /person/{login}/token:
     $ref: 'paths/person_login_token.yaml'
+  /person/{login}/group:
+    $ref: 'paths/person_login_group.yaml'
```
eduardoj, line 64:
```suggestion
  /person/{login}/group:
    $ref: 'paths/person_login_group.yaml'
  /person/{login}/token:
    $ref: 'paths/person_login_token.yaml'
```
Nitpick: I would order the endpoints alphabetically.

eduardoj (approved):

src/api/public/apidocs-new/OBS-v2.10.50.yaml
saraycp: good practice, done

PR #10677, opened 2021-01-22
Document person get_userinfo action

danidoni:
The OpenAPI documentation for the `person#get_userinfo` action.

-----------------

### DIFF
--- a/src/api/public/apidocs-new/components/schemas/person.yaml
+++ b/src/api/public/apidocs-new/components/schemas/person.yaml
@@ -1,0 +1,22 @@
+type: object
+properties:
+  login:
+    type: string
+    example: user_1
+  email:
+    type: string
+    example: unconfigured@openbuildservice.org
+  realname:
+    type: string
+    example: User1
+  state:
+    type: string
+    example: confirmed
+  globalrole:
+    type: string
+  ignore_auth_services:
+    type: boolean
+  watchlist:
+    $ref: 'watchlist.yaml'
+xml:
+  name: person
--- a/src/api/public/apidocs-new/paths/person_login.yaml
+++ b/src/api/public/apidocs-new/paths/person_login.yaml
@@ -1,3 +1,36 @@
+get:
+  summary: Get details about a person
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/login.yaml'
+  responses:
+    '200':
+      description: >
+        OK. The request has succeeded.
+
+        XML Schema used for body validation: [user.rng](../schema/user.rng)
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/person.yaml'
+          example:
+            login: Admin
+            email: root@localhost
+            realname: OBS Instance Superuser
+            state: confirmed
+            globalrole: Admin
+            ignore_auth_services: false
+            watchlist:
+              - project:
+                  name: "home:Admin"
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      $ref: '../components/responses/person/not_found.yaml'
+  tags:
+    - Person
+
 post:
   summary: Perform changes on a registered person
   security:
@@ -36,6 +69,6 @@ post:
     '403':
       $ref: '../components/responses/update_user_not_authorized.yaml'
     '404':
-      $ref: '../components/responses/not_found.yaml'
+      $ref: '../components/responses/person/not_found.yaml'
   tags:
     - Person
--- a/src/api/public/apidocs-new/components/responses/person/not_found.yaml
+++ b/src/api/public/apidocs-new/components/responses/person/not_found.yaml
@@ -5,7 +5,7 @@ description: |
 content:
   application/xml; charset=utf-8:
     schema:
-      $ref: '../schemas/api_response.yaml'
+      $ref: '../../schemas/api_response.yaml'
     example:
       code: not_found
       summary: "Couldn't find User with login = <login>"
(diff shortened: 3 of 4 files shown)


src/api/public/apidocs-new/paths/person_login.yaml
```diff
@@ -0,0 +1,37 @@
+get:
+  summary: Get details about a person
+  security:
+    - basic_authentication: []
+  parameters:
+    - in: path
+      name: login
```
saraycp, line 7: You could refer to [login.yaml](https://github.com/openSUSE/open-build-service/blob/master/src/api/public/apidocs-new/components/parameters/login.yaml) instead.

danidoni:

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -69,10 +69,13 @@ paths:
...
   /person:
     $ref: 'paths/person.yaml'
-  /person/register:
-    $ref: 'paths/person_register.yaml'
+  /person/{login}:
+    $ref: 'paths/person_login.yaml'
   /person/{login}/group:
     $ref: 'paths/person_login_group.yaml'
+  /person/register:
+    $ref: 'paths/person_register.yaml'
```
eduardoj, line 77:
```suggestion
  /person/register:
    $ref: 'paths/person_register.yaml'
  /person/{login}:
    $ref: 'paths/person_login.yaml'
  /person/{login}/group:
    $ref: 'paths/person_login_group.yaml'
```

Nitpick: I would list before fixed routes, and then routes with parameters. What do you think?

src/api/public/apidocs-new/paths/person_login.yaml
```diff
@@ -0,0 +1,37 @@
...
+            state: confirmed
+            globalrole: Admin
+            ignore_auth_services: false
+            watchlist:
+              - project:
+                  name: "home:Admin"
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      $ref: '../components/responses/notfound.yaml'
```
eduardoj, line 35:
```suggestion
      $ref: '../components/responses/not_found.yaml'
```
Using a not found response file works only in this case, because of the `summary` string in the example. Unfortunately, the component can not be reused, because the message in the example will be not always the same.

I'm for not using a component file here.

In case of using a file, I would also rename it adding an underscore.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
danidoni: I was going for alphabetical ordering, but I like your idea

src/api/public/apidocs-new/paths/person_login.yaml
danidoni: I would use a component and assume the message is just an example.

saraycp: I understand both points of view. The example's `summary` is not appropriate for all the `not_found` errors we can get. However, the example can be overwritten when defining the errors... but we can't ensure that everyone overwrites the example.

What about using a generic `summary` like: 
```xml
<summary> <summary> </summary>
```
?

Then, if someone forgets to overwrite the example, we won't have an inappropriate `summary`.

As plan B, we can also remove the `summary` from the file. In the worst case, we'll have a missing `summary`, not an inappropriate one.

saraycp: Sorry, I assumed the example can be overwritten, but it can not. I was mixing up `schemas` with `responses`.

We have decided offline to use some specific not_found files inside a proper namespace.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -71,8 +71,11 @@ paths:
     $ref: 'paths/person.yaml'
   /person/register:
     $ref: 'paths/person_register.yaml'
+  /person/{login}:
+    $ref: 'paths/person_login.yaml'
   /person/{login}/group:
     $ref: 'paths/person_login_group.yaml'
+
```
dmarcoux, line 78: Is this intentional? I would remove it.

src/api/public/apidocs-new/components/schemas/person.yaml
```diff
@@ -0,0 +1,18 @@
+type: object
+properties:
+  login:
+    type: string
+  email:
+    type: string
```
dmarcoux, line 6: It would be nice to have some value example here as it allows users to try it out in the web interface without having to type anything. Something like this:

https://github.com/openSUSE/open-build-service/blob/822926b893b11957161e9d7413a05a1fe85e1f1a/src/api/public/apidocs-new/components/schemas/configuration.yaml#L51-L53

src/api/public/apidocs-new/OBS-v2.10.50.yaml
saraycp: removed

src/api/public/apidocs-new/components/schemas/person.yaml
saraycp: I added examples for the main attributes

dmarcoux:
There is a conflict in _src/api/public/apidocs-new/paths/person_login.yaml_

src/api/public/apidocs-new/paths/person_login.yaml
```diff
@@ -1,3 +1,36 @@
...
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/login.yaml'
+  responses:
+    '200':
+      description: >
+        OK. The request has succeeded.
+
+        XML Schema used for body validation: [person.xsd](../schema/person.xsd)
```
dmarcoux, line 12: The link to _person.xsd_ is broken. I couldn't find that file in the project.

danidoni: Fixed

eduardoj (approved):

dmarcoux (approved):

PR #10680, opened 2021-01-22
Check if the user exist prior to check for user write policies

danidoni:
If we send a non-existent user, it's write policy check fail because we don't have the `@user` filled in.

Moving the policy check a bit down, we're covered by the user existence check.

### DIFF
--- a/src/api/app/controllers/person_controller.rb
+++ b/src/api/app/controllers/person_controller.rb
@@ -263,6 +263,6 @@ class PersonController < ApplicationController
   private
 
   def set_user
-    @user = User.find_by(login: params[:login])
+    @user = User.find_by!(login: params[:login])
   end
 end
--- a/src/api/spec/controllers/person_controller_spec.rb
+++ b/src/api/spec/controllers/person_controller_spec.rb
@@ -72,6 +72,16 @@ RSpec.describe PersonController do
       login user
     end
 
+    context 'when asking for a non-existing login' do
+      before do
+        post :post_userinfo, params: { login: 'nonexistent', cmd: 'change_password', format: :xml }
+      end
+
+      it 'does not change the password' do
+        expect(old_password_digest).to eq(user.reload.password_digest)
+      end
+    end
+
     context 'when using default authentication' do
       before do
         request.env['RAW_POST_DATA'] = 'password_has_changed'


src/api/app/controllers/person_controller.rb
```diff
@@ -41,12 +41,12 @@ def get_userinfo
   end
 
   def post_userinfo
-    authorize @user, :update?
-
     login = params[:login]
     # just for permission checking
     User.find_by_login!(login)
 
+    authorize @user, :update?
```
hennevogel, line 48: Should `set_user` catch this?

hennevogel (changes requested):

src/api/app/controllers/person_controller.rb
vpereira: I don't understand this change either @danidoni. Do you have a scenario where it is behaving wrongly, failing or throwing some exception? To be in the safe side, you could use `find_by!` on `set_user` but TBH, the original code LGTM.

danidoni: Yes, in the code before this change you could send a `login` parameter for a non-existent user (say `"captainreynolds"`). The `set_user` will fill `@user` with `nil`, and then the `authorize @user, :update?` will fail with 
```Pundit::NotDefinedError (unable to find policy `NilClassPolicy` for `nil`)```.
I will try to come up with a failing test.

Comment (danidoni) [reply src/api/app/controllers/person_controller.rb:48]: > Should `set_user` catch this?

No, the `set_user` [is not "explosive"](https://github.com/openSUSE/open-build-service/blob/a07fc3ff491c00984049f2261b59a5ffd52f69b6/src/api/app/controllers/person_controller.rb#L265), it will just return `nil`.

Comment (hennevogel) [reply src/api/app/controllers/person_controller.rb:48]: Well *why* isn't it "explosive"?

Comment (danidoni) [reply src/api/app/controllers/person_controller.rb]: I've changed it to throw an error, and returning the policy code where it was before.

File: src/api/spec/controllers/person_controller_spec.rb [hunk 2]
```diff
@@ -72,6 +72,16 @@
       login user
     end
 
+    context 'when asking for a non-existing login' do
+      before do
+        post :post_userinfo, params: { login: 'nonexistent', cmd: 'change_password', format: :xml }
```
hennevogel, line 77: I don't understand this spec. What are you testing?

hennevogel (changes requested):

src/api/spec/controllers/person_controller_spec.rb
danidoni: I'm triggering an action that uses the `set_user` before_action with a non-existent user to asses that aborts it.

dmarcoux (changes requested):
Instead of changing `set_user`, I would suggest changing both methods in [UserPolicy](https://github.com/openSUSE/open-build-service/blob/2ab3b55a532b5faba9527f2c647653c335efe847/src/api/app/policies/user_policy.rb) to handle nonexistent users. [Testing a policy](https://github.com/openSUSE/open-build-service/blob/2ab3b55a532b5faba9527f2c647653c335efe847/src/api/spec/policies/user_policy_spec.rb) is a lot easier and you won't have to turn around the bush by doing a POST only to test authorization.

danidoni:
> Instead of changing `set_user`, I would suggest changing the [UserPolicy](https://github.com/openSUSE/open-build-service/blob/2ab3b55a532b5faba9527f2c647653c335efe847/src/api/app/policies/user_policy.rb) to handle nonexistent users. [Testing a policy](https://github.com/openSUSE/open-build-service/blob/2ab3b55a532b5faba9527f2c647653c335efe847/src/api/spec/policies/user_policy_spec.rb) is a lot easier and you won't have to turn around the bush by doing a POST only to test authorization.

Huh, good idea. I'll try that...

danidoni:
The policy is working as expected so this PR makes no sense now

PR #10681, opened 2021-01-22
Document person post_userinfo action

danidoni:
Document `person#post_userinfo` action.

-----------------

### DIFF
--- a/src/api/public/apidocs-new/paths/person_login.yaml
+++ b/src/api/public/apidocs-new/paths/person_login.yaml
@@ -1,0 +1,41 @@
+post:
+  summary: Perform changes on a registered person
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/login.yaml'
+    - in: query
+      name: cmd
+      schema:
+        type: string
+        enum: [change_password,lock,delete]
+      required: true
+      description: "The command to execute against the provided person."
+      example: change_password
+  requestBody:
+    description: "The password in plain text when the cmd param is 'change_password'."
+    required: false
+    content:
+      text/plain:
+        schema:
+          type: string
+  responses:
+    '200':
+      description: >
+        OK. The request has succeeded.
+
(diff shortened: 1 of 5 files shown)


danidoni:

src/api/public/apidocs-new/paths/person_login.yaml
```diff
@@ -0,0 +1,44 @@
+post:
+  summary: Perform changes on a registered person
+  security:
+    - basic_authentication: []
+  parameters:
+    - in: path
```
krauselukas, line 6: @danidoni the login parameter is already defined in `../parameters/login.yaml`

src/api/public/apidocs-new/paths/person_login.yaml
```diff
@@ -0,0 +1,44 @@
...
+        XML Schema used for body validation: [api_response.xsd](../schema/api_response.xsd)
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '400':
+      $ref: '../components/responses/bad_request_error.yaml'
+    '404':
```
krauselukas, line 41: @danidoni there is also a possible 403 response https://github.com/openSUSE/open-build-service/blob/fd0900485928814ac792e5d758c5af126b7cd32a/src/api/app/controllers/person_controller.rb#L54

src/api/public/apidocs-new/paths/person_login.yaml
danidoni: Yes!

src/api/public/apidocs-new/paths/person_login.yaml
danidoni: Added a 403 error, but the code I've got from my tests is 'update_user_not_authorized' instead of 'change_password_no_permission'... I will do some tests later

krauselukas (approved):

PR #10683, opened 2021-01-24
Introduce saved replies feature

rubhanazeem:
Introduce the saved replies feature.

-----------------

### DIFF
--- a/src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
+++ b/src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
@@ -40,10 +40,19 @@
               = link_to(my_subscriptions_path, class: 'nav-link', title: 'Change Your Notifications') do
                 %i.fas.fa-lg.mr-2.fa-bell
                 %span.nav-item-name Change Your Notifications
+            %li.nav-item
+              = link_to('#', title: 'Saved replies', class: 'nav-link') do
+                %i.fas.fa-mail-bulk
+                Saved replies
+
         - else
           = link_to(my_subscriptions_path, class: 'd-block') do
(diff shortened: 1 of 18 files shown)


vpereira:
Well done :+1: 

Please squash the "Fix lint" commits if possible!

src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
```diff
@@ -40,10 +40,19 @@
               = link_to(my_subscriptions_path, class: 'nav-link', title: 'Change Your Notifications') do
                 %i.fas.fa-lg.mr-2.fa-bell
                 %span.nav-item-name Change Your Notifications
+            %li.nav-item
+              = link_to(saved_replies_path, title: 'Saved replies', class: 'nav-link') do
+                %i.fas.fa-mail-bulk
```
saraycp, line 45: The action here looks a little different from the rest, just because some Bootstrap classes are missing: `fa-lg` and `mr-2`.
Simply add them taking the other list items as an example.

saraycp (changes requested):

src/api/app/views/webui/user/user_profile_redesign/_info.html.haml
rubhanazeem: Done. Thanks a lot for the review. 😄

saraycp:
@rubhanazeem thank you for your work and for addressing our suggestions.

As we don't plan to introduce this feature in our code base right now, I'm going to close this PR. It was nice to see how you tackled this, that was the goal! :bouquet:

rubhanazeem:
Yeah definitely 😄

PR #10685, opened 2021-01-25
Proof of concept: two ways of documenting using OpenAPI

eduardoj:
This is a proof of concept of two ways of documenting the API using OpenAPI. Please, don't merge this pull request.

Two options are provided, with two examples:
- Group Attributes: define repeated endpoints, and reference inside them, in their descriptions, to the other one.
- Group Attributes2: define only an endpoint, and add a note in the description that there is another endpoint that can also be used.


eduardoj:
This proof of concept is no longer needed. Closing.

PR #10698, opened 2021-01-26
Mark properties of the /about route as required

dcermak:
The title, description and revision elements are always present according to https://build.opensuse.org/schema/about.xsd, so I've marked them as such.


dmarcoux (approved):
Thank you @dcermak!

PR #10706, opened 2021-01-28
Disable rspec nested groups globally

vpereira:
We are disabling it in almost all specs. There is no gain in having it enabled. To nest context in rspec is fine. 

-----------------


dmarcoux (approved):

danidoni (approved):

src/api/.rubocop_todo.yml
```diff
@@ -669,7 +669,8 @@ RSpec/DescribeClass:
 # Configuration parameters: SkipBlocks, EnforcedStyle.
 # SupportedStyles: described_class, explicit
 RSpec/DescribedClass:
-  Enabled: false
+  Max: 3
```
dmarcoux, line 672: I think you picked the wrong cop, so `RSpec/DescribedClass` instead of `RSpec/NestedGroups`. This change should also be done in the RuboCop main configuration file `src/api/.rubocop.yml`.

vpereira: yes, regarding doing it on the main file, I agree. I just followed the copy above/bellow which was done in the "TODO" file too. 

```
RSpec/MultipleMemoizedHelpers:
  Max: 27
```

dmarcoux: This was generated by RuboCop. It does that when some cops allow setting a maximum value. The goal is then to reduce that maximum value to the standard value. If we already know which maximum value we want (and it's not the default one), we can set it in the main configuration file.

PR #10707, opened 2021-01-28
Remove authentication requirement from POST /person/register documentation endpoint

danidoni:
It's not needed, following the source code:
https://github.com/openSUSE/open-build-service/blob/436ce907fd7d6e2abc201410922d1aa87444883f/src/api/app/controllers/person_controller.rb#L11

-----------------

### DIFF
--- a/src/api/public/apidocs-new/paths/person_register.yaml
+++ b/src/api/public/apidocs-new/paths/person_register.yaml
@@ -1,7 +1,5 @@
 post:
   summary: Registers a new person
-  security:
-    - basic_authentication: []
   requestBody:
     description: Data for the person to register.
     required: true


dmarcoux:
The failure in the `checkout_code` check is already addressed on _master_. So rebasing the branch will solve the issue.

danidoni:
> The failure in the `checkout_code` check is already addressed on _master_. So rebasing the branch will solve the issue.

Oh, nice! Rebasing!

krauselukas (approved):

krauselukas (changes requested):
Ah sorry @danidoni please just mention its about the endpoint documentation, not the endpoint it self in the commit message. :D I could imagine people getting nervous by just reading the commit message

danidoni:
> Ah sorry @danidoni please just mention its about the endpoint documentation, not the endpoint it self in the commit message. :D I could imagine people getting nervous by just reading the commit message

Good point! Updated!

krauselukas:
@danidoni would be awesome if you could change the commit message itself, its just a little bit missleading

danidoni:
> @danidoni would be awesome if you could change the commit message itself, its just a little bit missleading

Done!

krauselukas (approved):
Gracias

dmarcoux (approved):

PR #10710, opened 2021-01-28
Adjust EventSubscription::FindForEvent to handle group members

dmarcoux:
Depending on the subscriptions of the user

### DIFF
--- a/src/api/spec/services/event_receivers_fetcher_spec.rb
+++ b/src/api/spec/services/event_receivers_fetcher_spec.rb
@@ -1,0 +1,53 @@
+require 'rails_helper'
+
+RSpec.describe EventReceiversFetcher, type: :service do
+  describe '#call' do
+    # Create package with an inherited maintainer
+    let!(:inherited_maintainer) { create(:confirmed_user, :with_home) }
+    let!(:project) { inherited_maintainer.home_project }
+    let!(:package) { create(:package, project: project) }
+
+    # Create event receivers
+    let!(:receiver_role) { :target_maintainer }
+    let!(:user) { create(:confirmed_user) }
+    let!(:group) { create(:group) }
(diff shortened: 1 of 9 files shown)


dmarcoux:
[Filter by group](https://trello.com/c/d9PDB2mD/1604-filter-by-group)

src/api/app/services/event_receivers_gatherer.rb
```diff
@@ -0,0 +1,33 @@
+# Gather event receivers for a receiver role
+#   An event receiver is a user/group wanting to be notified about an event
+#   A receiver role is defined on the event itself. It could be a maintainer, bugowner, etc...
+class EventReceiversGatherer
```
dmarcoux, line 4: Let's use `Fetcher` instead of `Gatherer`. There's already another service with that suffix. It will be consistent.

danidoni: Fixed

src/api/spec/services/event_receivers_fetcher_spec.rb
```diff
@@ -22,7 +22,7 @@
   let!(:bs_request) { create(:bs_request_with_submit_action, state: :new, target_package: package) }
 
   describe '#call' do
-    subject { described_class.new(Event::RequestCreate.first, receiver_role).call }
+    subject { described_class.new(receiver_role).call }
```
dmarcoux, line 25: This change needs to be undone. An event is needed to initialize an instance of _EventReceiversFetcher_.

danidoni: Yes, I found it now :sweat_smile:

saraycp:
We finally decided to go for another approach. This PR can be closed but can be used as a reference.

PR #10713, opened 2021-01-29
Add notifications filter for incoming and outgoing requests

saraycp:
Applying the same logic existent in the Tasks section, we want to filter notifications for incoming and outgoing requests.
Unlike the Tasks, the incoming/outgoing notifications won't be restricted to requests in states new or review only.

The user who creates a request is not notified about that, of course. So, that user will never see notifications for their newly created requests. However, if someone else changes the state of the request (which is an outgoing request), the originator will be notified about that change. For example when another user declines that request.

Screenshot_2021-02-02 Open Build Service(2)

**Edit:**

`review-app` has a good bunch of notifications to play with.

I also added a bit of literature that can help to understand [What Is Notified And What Not](https://github.com/openSUSE/open-build-service/wiki/Notifications#what-is-notified-and-what-not)

-----------------


src/api/app/models/bs_request.rb
```diff
@@ -76,7 +76,7 @@ class BsRequest < ApplicationRecord
 
   has_many :bs_request_actions, dependent: :destroy
   has_many :reviews, dependent: :delete_all
-  has_many :comments, as: :commentable, dependent: :delete_all
+  has_many :comments, as: :commentable, dependent: :destroy
```
danidoni, line 79: Is this change related to the PR?

saraycp: Well, it's somehow related and we discovered this when testing different use cases in notifications. But I could maybe open a new PR just for that and extend the description. 

I tried to perform all the possible changes that can happen to a request (create one, add a review on it, change to different states, delete it) in order to check what kind of notification was created on each case.

When I did delete the request (something that only Admins via API can do) I realized that existing notifications for comments on that request were not deleted so we got an error when we tried to access the notifications page.

saraycp: New PR open for that: #10735

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Webui::Users::NotificationsController < Webui::WebuiController
   MAX_PER_PAGE = 300
-  VALID_NOTIFICATION_TYPES = ['read', 'reviews', 'comments', 'requests', 'unread'].freeze
+  VALID_NOTIFICATION_TYPES = ['read', 'reviews', 'comments', 'requests', 'unread', 'incoming_requests', 'outgoing_requests'].freeze
```
vpereira, line 3: not possible to transform it in symbols?

src/api/.rubocop_todo.yml
```diff
@@ -815,6 +814,7 @@ RSpec/LetSetup:
     - 'spec/controllers/webui/request_controller_spec.rb'
     - 'spec/controllers/webui/search_controller_spec.rb'
     - 'spec/controllers/webui/staging/workflows_controller_spec.rb'
+    - 'spec/controllers/webui/users/notifications_controller_spec.rb'
```
vpereira, line 817: we should disable this cop. It's normal in our setup to have to create objects but don't call it directly in the spec.

src/api/app/controllers/webui/users/notifications_controller.rb
saraycp: In case it's possible I wouldn't do that in this PR, maybe in another one for refactoring.

src/api/.rubocop_todo.yml
saraycp: I agree. In this case, I don't think it worth it to try to find an alternative for all the tests affected. At least not in the near future.

saraycp: Done @vpereira

src/api/spec/controllers/webui/users/notifications_controller_spec.rb
```diff
@@ -104,6 +104,80 @@
...
+
+      let(:maintained_request) do
+        create(:bs_request_with_submit_action,
+               target_package: target_package,
+               source_package: source_package,
+               creator: admin_user)
+      end
+
+      let!(:state_change_to_declined_notification) { create(:web_notification, :request_state_change, notifiable: declined_bs_request, subscriber: user) }
+      let(:request_created_notification) { create(:web_notification, :request_created, notifiable: maintained_request, subscriber: user) }
```
krauselukas, line 167: @saraycp you don't test for the `request_created_notification`, I guess you wanted to do the same as with the incoming requests context, to test that it is not contained when you filter. If not I would simply drop it

saraycp: In the example, I use `eq` instead of `include`, that ensures that `request_created_notification` is not in the result.
However, I'm going to split it into two examples so it's more clear.

saraycp: done, @krauselukas

krauselukas: :+1:

krauselukas (approved):
LGTM

PR #5610, opened 2018-08-09
Configure Rails generator to use integer as primary and foreign keys

bgeuken:
All our existing tables use integers as primary key. Using the "new"
default in Rails (bigint) would cause breakage in MySQL.

PR #3809, opened 2017-09-12
Convert database columns with type bigint => int.

evanrolfe:
https://trello.com/c/My00k8q2/532-3-p6-evaluate-change-of-table-id-int-type-made-by-rails-51

Since [rails PR 26266](https://github.com/rails/rails/pull/26266) any new tables created will have the primary key type set to bigint by default. However that will cause problems because we will not be able to create foreign keys between bigint and int columns. If we want to use bigint then we will need to convert all primary and foreign keys from int to bigint and this will take a considerable amount of work to get these migrations to run without causing too much interuption. So for the meantime I propose that we stick with int for the meantime and then evaluate if switching to bigint is worth the cost.

Further reading:
http://ridingtheclutch.com/post/160099545985/rails-51-new-default-bigint-primary-key-sort-of

PR #10715, opened 2021-01-29
Change Notifications#id to unsigned int

vpereira:
To avoid the integers wrapping soon, change it from int(11) to unsigned int(10)

I added the DO NOT MIGRATE since it should migrated just in our maintenance window. 

-----------------

If this PR requires any particular action or consideration before deployment,
please check the reasons or add your own to the list:

* [X] Requires a database migration that can cause downtime. Postpone deployment until maintenance window[1].
* [ ] Contains a data migration that can cause temporary inconsistency, so should be run at a specific point of time.
* [ ] Changes some configuration files (e.g. options.yml), so the changes have to be applied manually in the reference server.
* [ ] A new Feature Toggle[2] should be enabled in the reference server.
* [ ] Proper documentation or announcement has to be published upfront since the introduced changes can confuse the users.

[1] https://github.com/openSUSE/open-build-service/wiki/Deployment-of-build.opensuse.org#when-there-are-migrations
[2] https://github.com/openSUSE/open-build-service/wiki/Feature-Toggles-%28Flipper%29#you-want-real-people-to-test-your-feature

PR #10717, opened 2021-01-29
Adapt rails generator to create primary keys as unsigned int by default

vpereira:
Rails since version 5.1 has all primary keys being created as `Bigint` by default.  We changed it [back in the days](https://github.com/openSUSE/open-build-service/pull/3809), rolling it back to `integer` and forcing it in our [generators](https://github.com/openSUSE/open-build-service/pull/5610). We should however be doing it using [unsigned int instead of int](https://dev.mysql.com/doc/refman/8.0/en/integer-types.html).

How to test it:

- Pull the branch
- `rails g model foo name:string`
- `rails db:migrate`
- check your `schema.db`

```
# schema.rb

+  create_table "foos", id: :integer, unsigned: true, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4", force: :cascade do |t|
+    t.string "name"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+  end
+
```

-----------------

* [x] Requires a database migration that can cause downtime. Postpone deployment until maintenance window[1].

### DIFF
--- a/src/api/config/application.rb
+++ b/src/api/config/application.rb
@@ -106,7 +106,7 @@ module OBSApi
     config.generators do |g|
       g.factory_bot(false)
       g.test_framework :rspec
-      g.orm :active_record, primary_key_type: :integer
+      g.orm :active_record, primary_key_type: :unsigned_integer
     end
   end
 end


vpereira:
[Do research about using another data type for notification id](https://trello.com/c/t1vrsI1o/1607-do-research-about-using-another-data-type-for-notification-id)

coolo:
going bigint is still the correct thing to do. The worry back then was that you can't create foreign keys towards int if `integer` is bigint - which is true. But the primary key id can still be bigint and foreign keys can be `t.bigint`

You actually have the same problem with what you're proposing. If the primary key default is unsigned int, then every new foreign key needs to specify this and plain `integer` wouldn't do any longer.

vpereira:
> going bigint is still the correct thing to do. The worry back then was that you can't create foreign keys towards int if `integer` is bigint - which is true. But the primary key id can still be bigint and foreign keys can be `t.bigint`
> 
> You actually have the same problem with what you're proposing. If the primary key default is unsigned int, then every new foreign key needs to specify this and plain `integer` wouldn't do any longer.

Hi @coolo 

Hm interesting. After some initial tests, I couldn't spot any issues (I'm able to get notifications and display it in the UI) in the branch https://github.com/openSUSE/open-build-service/pull/10715. I think after ruby 2.4, where `BigNum` and `FixNum` got merged into `Integer` maybe even the old issue was gone too?

coolo:
How ruby names the types doesn't matter to mariadb.

coolo:
What's most important is the foreign key and the primary key it refers to have the same range. So if something fails for int vs bigint, it will fail also with int vs unsigned int. If something works for int vs unsigned int, it will also work for int vs bigint.

Now the problem is, I don't *really* understand what the problem with bigint back then was. New tables using bigint can't be the problem (which was reverted), I assume it's rather adding new foreign keys to existing tables. But even in `t.references` you use `type: :integer` or `type: :bigint` (any idea how to give unsigned int there?). Possibly that option was given in later rails versions?

vpereira:
> How ruby names the types doesn't matter to mariadb.

Yes sure, 

> How ruby names the types doesn't matter to mariadb.

I thought you are talking about the rails association, sure, in the DB that's no way to overcomeit

vpereira:
> Possibly that option was given in later rails versions?

Yes. its new. Back in the days, there was no way to do it.

vpereira:
> Yes. its new. Back in the days, there was no way to do it. You will probably to pass `:unsigned_integer`

vpereira:
@coolo https://github.com/rails/rails/blob/main/activerecord/lib/active_record/type/unsigned_integer.rb

vpereira:
> Now the problem is, I don't _really_ understand what the problem with bigint back then was.

I think the problem happens when you have foreign key constraints.

vpereira:
I will close this PR, let's see what the future holds for us :pray:

PR #10722, opened 2021-02-01
Mockup for the Notification's group filters

danidoni:
Screenshot from 2021-02-01 15-57-31

## TODO
- [ ] Fix code climate issues

-----------------

### DIFF
--- a/src/api/app/controllers/webui/users/notifications_controller.rb
+++ b/src/api/app/controllers/webui/users/notifications_controller.rb
@@ -61,23 +61,6 @@ class Webui::Users::NotificationsController < Webui::WebuiController
     notifications.page(params[:page]).per([total, MAX_PER_PAGE].min)
   end
 
-  # Returns a hash where the key is the name of the project and the value is the amount of notifications
-  # associated to that project. The hash is sorted by amount and then name.
-  def projects_for_filter
-    Project.joins(:notifications)
-           .where(notifications: { subscriber: User.session, delivered: false, web: true })
-           .order('name desc').group(:name).count # this query returns a sorted-by-name hash like { "home:b" => 1, "home:a" => 3  }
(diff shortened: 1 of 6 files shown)


danidoni:
[Filter by group](https://trello.com/c/d9PDB2mD/1604-filter-by-group)

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -91,9 +82,9 @@ def paginated_notifications
   end
 
   def notifications_filter
-    NotificationsFilterPresenter.new(projects_for_filter,
+    NotificationsFilterPresenter.new(ProjectFilterChoicesPresenter.new(params[:project]),
+                                     GroupFilterChoicesPresenter.new,
```
danidoni, line 86: What do you think of this kind of refactor? The thing is, with the group filtering now there are 5 parameters. Rubocop complains about that. I've been thinking in disabling that cop here but first I want to try this approach.

krauselukas:
@danidoni the mockup it self looks good to me, I would have only one thing in mind which might improve the handling a bit in case you are member of many groups. Maybe a way to collapse the filter group area. Otherwise you might need to scroll a bit to reach the project filters.

src/api/app/controllers/webui/users/notifications_controller.rb
vpereira: Better if we could switch from positional argument to keywords. In another hand if you are depending exclusively on params and coupling the classes together isn't a problem, then you could pass params to the `NotificationsFilterPresenter` and in the initializer you create the dependency with the `ProjectFilterChoicesPresenter` and `GroupFilterChoicesPresenter`

saraycp:
This was just a Mockup. We can close the PR.

PR #10730, opened 2021-02-02
Merge old migrations

vpereira:
Even though the default way to deploy our application is via `rake db:schema:load` to have a broken `rake db:migrate` doesn't help. In this PR I did the following:

- Merged old migrations and deleted them
- Checked if all sql `INSERT` calls are covered by the `seeds.rb`
- Adapted old data migrations that were looking for old migration files
- Created a circleci job to check if any new introduced chages are breaking the migrations

I kept the last migration file there as placeholder, to make sure that we don't break production. After we create a new migration, we could safely delete this placeholder migration

Feel free to `rake db:migrate` your db as you wish :)

For a follow-up PR: 

- Clean the data migrations too

-----------------


src/api/spec/db/data/remove_duplicate_repositories_spec.rb
```diff
@@ -1,6 +1,5 @@
 require 'rails_helper'
 require Rails.root.join('db/data/20170306084550_remove_duplicate_repositories.rb')
-require Rails.root.join('db/migrate/20170306084558_change_repositories_remote_project_name_to_not_null.rb')
```
dmarcoux, line 3: I think this is not the only thing which can be changed/removed in this file. See the comment on the lines 5 to 7 in this file. There's also some funky stuff happening in the _before_ and _after_ blocks below.

.circleci/config.yml
```diff
@@ -284,6 +304,9 @@ workflows:
   test_all:
     jobs:
       - checkout_code
+      - migrations_test:
```
dmarcoux, line 307: I know it's not something we already do, but it could be nice to run this only when at least one file is changed in the directory `src/api/db`.

src/api/spec/db/data/remove_duplicate_repositories_spec.rb
vpereira: yes, well it's for spec and the spec is passing, so i guess, we are good?  The question is if we really need those 4 years old data migrations?

dmarcoux: It's good if the specs are already passing. If we keep the file around, I would still prefer to remove some of that weird code. It's going to confuse anybody looking into this.

It's true that we might not need those 4 years old data migrations anymore.

vpereira: Yes, I would like to dig in the left over in another PR, otherwise the changelog will be even bigger. I think this task can be done in the "follow-up PR" as I described in the PR description.

danidoni:
The `rake data:migrate` is still broken. Do you plan to work on this too?

vpereira:
> The `rake data:migrate` is still broken. Do you plan to work on this too?

No, you just have to seed the DB, otherwise there is no data to migrate

If you look on the circleci config, I added a job to test the migrations, and we have the following there:

```rake db:drop && rake db:create && rake db:migrate && rake db:seed && rake data:migrate```

dmarcoux (approved):
Looking good, I gave it a try in my development environment and it works. Production will also be fine since we keep the current migration. Good job!

Please wait for another review before merging since this is quite a big change.

danidoni (approved):

PR #10734, opened 2021-02-03
[api] fix diffing of accepted maintenance_incident requests

adrianschroeter:
It just errored out because of missing target package name before

-----------------


adrianschroeter:
no clue about the webui test cases in rspec ... can ne1 give me a hint what is wrong?

danidoni:
The specs are complaining about the flash message not visible, if you check the [png artifacts](https://app.circleci.com/pipelines/github/openSUSE/open-build-service/2632/workflows/ed1e2f5c-1c6b-4990-99a9-65d291788751/jobs/56750/artifacts) you'll see that the screen is scrolled down, so the flash is not visible.

adrianschroeter:
I fail to see what this has to do with the diffing fix. If there is no change, the request creation should not have been possible already.

adrianschroeter:
found a crash in the logs ...

adrianschroeter:
```
[2f3c7205-480c-46cb-aef9-9fa7c3206a3b] [198:42.56] Completed 500 Internal Server Error in 327ms (ActiveRecord: 83.1ms | Backend: 
14.6ms | Allocations: 134363)
[2f3c7205-480c-46cb-aef9-9fa7c3206a3b] [198:42.56]   
[2f3c7205-480c-46cb-aef9-9fa7c3206a3b] VCR::Errors::UnhandledHTTPRequestError (

================================================================================
An HTTP request has been made that VCR does not know how to handle:
  GET http://backend:5352/source/home:tom:branches:ProjectWithRepo:Update/ProjectWithRepo_package

```

does this mean that the VCR needs to be re-done? Or is it something else?

could someone do this for me please?

krauselukas:
@adrianschroeter I can have a look to see whats going on here

krauselukas:
@adrianschroeter What exactly errored out before? The diff in an accepted maintenance incident works for me on master. What do I need to do to see what was wrong before?

adrianschroeter:
On Mittwoch, 10. Februar 2021, 14:06:38 CET Lukas Krause wrote:
> @adrianschroeter What exactly errored out before? The diff in an accepted maintenance incident works for me on master. What do I need to do to see what was wrong before?

learned meanwhile it only hits accepted requests which became a different state
later one. eg. declined. We have examples in internal system.

That fix is anyway solving this issue as acceptinfo should be used
when it exists. Independend of the request state.

-- 

src/api/app/models/bs_request_action/differ/for_source.rb
```diff
@@ -26,11 +26,11 @@ def diff(source_package_name)
...
           )
           source_diff(query_builder.project_name, query_builder.package_name, query.merge(query_builder.build))
-        elsif accepted?
-          query_builder = QueryBuilderForAccepted.new(
+        elsif bs_request_action.bs_request_action_accept_info.present?
+          query_builder = QueryBuilderForAcceptinfo.new(
             bs_request_action_accept_info: bs_request_action.bs_request_action_accept_info
           )
-          source_diff(target_project_name, target_package_name, query.merge(query_builder.build))
+          source_diff(bs_request_action.target_project, target_package_name, query.merge(query_builder.build))
```
hennevogel, line 33: Why would you not use `target_project_name` in this case?

src/api/app/models/bs_request_action/differ/for_source.rb
```diff
@@ -26,11 +26,11 @@ def diff(source_package_name)
             source_package_name: source_package_name
           )
           source_diff(query_builder.project_name, query_builder.package_name, query.merge(query_builder.build))
-        elsif accepted?
-          query_builder = QueryBuilderForAccepted.new(
+        elsif bs_request_action.bs_request_action_accept_info.present?
```
hennevogel, line 29: this is the same as the `accepted?` method below...

hennevogel:
@adrianschroeter doesn't work without removing the maintenance suffix, but I'm not sure what to remove it from...

hennevogel:
It also looks to me like the usual regex we use for this does not cover `Project.name` with a dot? 

```
Expected response to be a <2XX: success>, but was a <404: Not Found>

Response body: <status code="diff">

  <summary>The diff call for BaseDistro2.0:LinkedUpdateProject/kdelibs.BaseDistro2.0_LinkedUpdateProject failed: package 'kdelibs.BaseDistro2.0_LinkedUpdateProject' does not exist</summary>

</status>
```

PR #10731, opened 2021-02-02
[api][backend] add linkedbuild=alldirect functionality

adrianschroeter:
To allow rebuilds from directly linked projects only, excluding package
sources from indirect linked projects.

This is for example required to rebuild ServicePack layers only.

_Replace this line with a description of your changes. Please follow the suggestions in the comment below._

-----------------

If this PR requires any particular action or consideration before deployment,
please check the reasons or add your own to the list:

* [ ] Requires a database migration that can cause downtime. Postpone deployment until maintenance window[1].
* [ ] Contains a data migration that can cause temporary inconsistency, so should be run at a specific point of time.
* [ ] Changes some configuration files (e.g. options.yml), so the changes have to be applied manually in the reference server.
* [ ] A new Feature Toggle[2] should be enabled in the reference server.
* [ ] Proper documentation or announcement has to be published upfront since the introduced changes can confuse the users.

[1] https://github.com/openSUSE/open-build-service/wiki/Deployment-of-build.opensuse.org#when-there-are-migrations
[2] https://github.com/openSUSE/open-build-service/wiki/Feature-Toggles-%28Flipper%29#you-want-real-people-to-test-your-feature

PR #10740, opened 2021-02-05
Consolidate database migrations before 2017

vpereira:
Even though the default way to deploy our application is via rake db:schema:load to have a broken rake db:migrate doesn't help. In this PR I did the following:

Merged old migrations (older than 2017) and deleted them
Checked if all sql INSERT calls are covered by the seeds.rb
Adapted old data migrations that were looking for old migration files
Created a circleci job to check if any new introduced chages are breaking the migrations
Clean old migration versions from DB if there is no migration file for that

Feel free to rake db:migrate your db as you wish :)

For a follow-up PR: Clean the data migrations too

-----------------


src/api/app/models/repository.rb
```diff
@@ -281,7 +281,7 @@ def copy_to(new_project)
 #
 #  id                  :integer          not null, primary key
 #  block               :string(5)
-#  linkedbuild         :string(8)
+#  linkedbuild         :string(9)
```
vpereira, line 284: it should be introduced by https://github.com/openSUSE/open-build-service/pull/10731 but just if the migration was tested before.

src/api/db/migrate/20170306084558_change_repositories_remote_project_name_to_not_null.rb
```diff
@@ -1,10 +1,5 @@
 class ChangeRepositoriesRemoteProjectNameToNotNull < ActiveRecord::Migration[5.0]
   def up
-    # We need to run src/api/db/data/20170306084550_remove_duplicate_repositories.rb first,
```
vpereira, line 3: This statement doesn't hold true if you are setting up the database from scratch

src/api/db/migrate/20180516074538_explicitly_set_charset.rb
```diff
@@ -8,7 +8,7 @@ def up
       'group_maintainers', 'history_elements', 'kiwi_descriptions', 'notifications', 'product_media',
       'bs_request_counter'
     ].each do |table|
-      execute("ALTER TABLE #{table} CONVERT TO CHARACTER SET utf8")
+      execute("ALTER TABLE #{table} CONVERT TO CHARACTER SET utf8") if ActiveRecord::Base.connection.table_exists?(table)
```
vpereira, line 11: the table `data_migrations` just exists after you run `rake data:migrate`

src/api/db/migrate/20210205113812_squasher_clean.rb
```diff
@@ -0,0 +1,12 @@
+class SquasherClean < ActiveRecord::Migration[6.0]
+  class SchemaMigration < ApplicationRecord
+  end
+
+  def up
+    migrations = Dir.glob(File.join(File.dirname(__FILE__), '*.rb'))
+    versions = migrations.map { |file| File.basename(file)[/\A\d+/] }
+    SchemaMigration.where.not(version: versions).delete_all
```
vpereira, line 8: we clean old migrations from the schema that don't exist in the file system anymore.

dmarcoux:
Running `bundle exec rake db:drop db:create db:migrate db:seed data:migrate` works. Why did you go with migrations before 2017? Why 2017?

dmarcoux (approved):

vpereira:
[squash old migrations together](https://trello.com/c/sddCsQfK/1877-squash-old-migrations-together)

PR #10741, opened 2021-02-05
Remove notifications for missing notifiables

saraycp:
Add a data migration to remove dangling notifications that should have been removed when their associated comments where removed.

This is a follow-up of PR #10735.

**NOTE:** This PR contains a data migration but it won't cause any inconsistency or downtime.

-----------------

### DIFF
--- a/src/api/spec/db/data/remove_notifications_for_missing_notifiables_spec.rb
+++ b/src/api/spec/db/data/remove_notifications_for_missing_notifiables_spec.rb
@@ -1,0 +1,21 @@
+require 'rails_helper'
+require Rails.root.join('db/data/20210205131609_remove_notifications_for_missing_notifiables.rb')
+RSpec.describe RemoveNotificationsForMissingNotifiables, type: :migration do
+  describe 'up' do
+    let!(:notification1) { create(:notification, :comment_for_request) }
+    let!(:notification2) { create(:notification, :comment_for_request) }
+
+    before do
+      # Simulate the wrong behaviour we had before. When a BsRequest was removed,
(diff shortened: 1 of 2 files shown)


src/api/spec/db/data/remove_notifications_for_missing_notifiables_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+require Rails.root.join('db/data/20210205131609_remove_notifications_for_missing_notifiables.rb')
+RSpec.describe RemoveNotificationsForMissingNotifiables, type: :migration do
+  describe 'up' do
+    let!(:notification1) { create(:notification, :comment_for_request) }
+    let!(:notification2) { create(:notification, :comment_for_request) }
+
+    before do
+      # Simulate the wrong behaviour we had before. When a BsRequest was removed,
+      # all the comments were removed as well, but not their associated notifications.
+      # That happended because of a wrong association, dependent option: `delete_all` instead of `destroy`.
```
dmarcoux, line 11: Really not a blocker, just a small typo :)

```suggestion
      # That happened because of a wrong association, dependent option: `delete_all` instead of `destroy`.
```

vpereira (approved):

vpereira (approved):

src/api/spec/db/data/remove_notifications_for_missing_notifiables_spec.rb
saraycp: Ohh, thanks! Corrected!

dmarcoux (approved):

PR #10747, opened 2021-02-08
Reintroduce CleanupNotifications job

eduardoj:
Remove notifications older than a year with a daily job.

A `notifications_lifetime` config option can be set, to specify the number of days a notification persists before being removed.

-----------------

* [x] Changes some configuration files (e.g. options.yml), so the changes have to be applied manually in the reference server.

### DIFF
--- a/src/api/app/jobs/cleanup_notifications_job.rb
+++ b/src/api/app/jobs/cleanup_notifications_job.rb
@@ -1,0 +1,5 @@
+class CleanupNotificationsJob < ApplicationJob
+  def perform
+    NotificationsFinder.new.stale.delete_all
+  end
+end
--- a/src/api/app/queries/notifications_finder.rb
+++ b/src/api/app/queries/notifications_finder.rb
@@ -48,4 +48,14 @@ class NotificationsFinder
 
     self.class.new.for_subscribed_user.find_by(id: notification_id)
   end
+
+  def stale
+    @relation.where('created_at < ?', notifications_lifetime.days.ago)
+  end
+
+  private
+
+  def notifications_lifetime
+    CONFIG['notifications_lifetime'] ||= 365
+  end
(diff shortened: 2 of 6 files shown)


krauselukas (commented):
@eduardoj @saraycp LGTM, just asking myself if the job should already end up in the `clock.rb` file and if when and how often should we trigger it

Guess there is no need to run it often...

eduardoj:
> @eduardoj @saraycp LGTM, just asking myself if the job should already end up in the `clock.rb` file and if when and how often should we trigger it
> 
> Guess there is no need to run it often...

Nice catch! What about performing the job daily?

krauselukas:
> > @eduardoj @saraycp LGTM, just asking myself if the job should already end up in the `clock.rb` file and if when and how often should we trigger it
> > Guess there is no need to run it often...
> 
> Nice catch! What about performing the job daily?

No strong opinion on that, I guess once a week would also be sufficient, but the operation is not expensive, so every day seems to be okay as well :)

krauselukas (approved):

src/api/app/models/notification.rb
```diff
@@ -17,6 +17,10 @@ def event
     @event ||= event_type.constantize.new(event_payload)
   end
 
+  def self.cleanup
+    NotificationsFinder.new.stale.delete_all
```
dmarcoux, line 21: It's not a blocker, but I would put this directly in the cleanup job. I think it's unnecessary to have a class method for this. It would remind easy to test in the specs.

eduardoj: Yes, much better. I removed the `self.cleanup` method, replaced the previous test with a job test, and also added the `Job` suffix. Please have another look.

dmarcoux (approved):

PR #10753, opened 2021-02-09
Unconditionally show architecture and repository

### LINKED ISSUE #10738: build log screen does not show repo/architecture
# Issue Description
- pick a project with a lot of packages with quite some failed ones in different architectures and repos
- middle click on the build log of several failed packages in the monitor to open the logs in tabs
- the tabs do not show which repo and architecture the build log is for. So when debugging a larger project it's harder to see what  the log is about, eg

buildlog

Expected Result
============

Instead of showing only the project and package name in the window headline it should also show the repo and architecture of the buildlog in question. Actually package and project name are kind of redundant as they are already in the breadcrumbs also.

seto49:
Previously architecture and repository were only shown to users in the
build log while the job was running. This commit changes this behavior,
so that they are displayed independently of the job status.

Fixes #10738

-----------------

### DIFF
--- a/src/api/app/views/webui/package/_job_status.html.haml
+++ b/src/api/app/views/webui/package/_job_status.html.haml
@@ -1,9 +1,4 @@
 %p
-  %strong Repository:
-  = repository
-  %strong Architecture:
-  = architecture
-%p
   %strong Worker:
   = worker_id
   %strong Buildtime:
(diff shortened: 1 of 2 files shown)


danidoni:
Hi, thanks for the submission. It would be awesome if you could add some screenshots to document how it was before and how is it now.

krauselukas:
I had a look on the changes, I quickly did the screenshots:

**Before**
Screenshot_2021-02-09 Open Build Service(2)
**After**
Screenshot_2021-02-09 Open Build Service(1)

krauselukas (approved):
@seto49 LGTM, thanks a lot for your work :)

seto49:
Thanks everyone! I'll make sure to add screenshots in the next one.

PR #10756, opened 2021-02-09
Track some metrics about notifications

saraycp:
Every newly created notification is tracked _in situ_.
Some other metrics like the amount of different kind of notifications are tracked every 5 minutes.

MeasurementsJob has been slightly refactored.

-----------------

### DIFF
--- a/src/api/app/jobs/measurements_job.rb
+++ b/src/api/app/jobs/measurements_job.rb
@@ -5,36 +5,49 @@ class MeasurementsJob < ApplicationJob
     return unless CONFIG['amqp_options']
 
     RabbitmqBus.send_to_bus('metrics', "group count=#{Group.count}")
-    RabbitmqBus.send_to_bus('metrics', "user #{user_fields}")
+    RabbitmqBus.send_to_bus('metrics', "user #{measurements_to_fields(users_measurements)}")
+    RabbitmqBus.send_to_bus('metrics', "notification #{measurements_to_fields(notifications_measurements)}")
   end
 
   private
 
-  def user_fields
-    user_measures.map { |k, v| "#{k}=#{v}" }.join(',')
+  def measurements_to_fields(measurements)
+    measurements.map { |k, v| "#{k}=#{v}" }.join(',')
   end
 
-  def user_measures
+  def users_measurements
     {
       in_beta: User.in_beta.count,
       in_rollout: User.in_rollout.count,
       count: User.count
     }
-      .merge(role_fields)
-      .merge(state_fields)
+      .merge(roles_measurements)
+      .merge(states_measurements)
   end
 
-  def role_fields
+  def roles_measurements
     Role.global.pluck(:title).each_with_object({}) do |role_title, fields|
       fields[role_title.downcase.to_sym] = Role.find_by(title: role_title).users.count
       fields
     end
   end
 
-  def state_fields
+  def states_measurements
     User::STATES.each_with_object({}) do |state, fields|
       fields[state.to_sym] = User.where(state: state).count
       fields
     end
+  end
+
+  def notifications_measurements
+    {
+      count: Notification.count,
+      for_web: Notification.for_web.count,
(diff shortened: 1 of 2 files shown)


src/api/app/models/notification.rb
```diff
@@ -40,6 +42,13 @@ def unread?
   def unread_date
     last_seen_at || created_at
   end
+
+  private
+
+  def track_notification_creation
+    RabbitmqBus.send_to_bus('metrics',
+                            "notification.create,notifiable_type=#{notifiable_type},notifiable_id=#{notifiable_id},web=#{web},rss=#{rss} value=1")
```
hennevogel, line 50: `notifiable_id` is not really interesting and has a very high cardinality. This is to be avoided as it will lead to a slow index in the TSDB.

hennevogel (changes requested):

src/api/app/models/notification.rb
saraycp: You are right! I planned to use it to detect requests that had a lot of activity/notifications. But it's not really so important.

I've removed that tag. Please have another look.

src/api/app/jobs/measurements_job.rb
```diff
@@ -5,36 +5,49 @@ def perform
...
       fields
     end
   end
+
+  def notifications_measurements
+    {
+      count: Notification.count,
+      for_web: Notification.for_web.count,
+      for_rss: Notification.for_rss.count,
+      read: Notification.where(delivered: true).count,
```
dmarcoux, line 47: To reuse the `NotificationsFinder` query object:

```suggestion
      read: NotificationsFinder.new.read.count,
```

src/api/app/jobs/measurements_job.rb
```diff
@@ -5,36 +5,49 @@ def perform
...
+      unread: Notification.where(delivered: false).count,
```
dmarcoux, line 48: Same here:

```suggestion
      unread: NotificationsFinder.new.unread.count,
```

src/api/app/jobs/measurements_job.rb
```diff
@@ -5,36 +5,49 @@ def perform
...
+      about_comments: Notification.where(notifiable_type: 'Comment').count,
```
dmarcoux, line 49: This could also be possible with `NotificationsFinder.new.for_notifiable_type('comments').count`. It's only returning unread notifications though.

src/api/app/jobs/measurements_job.rb
```diff
@@ -5,36 +5,49 @@ def perform
...
+      about_requests: Notification.where(notifiable_type: 'BsRequest').count
```
dmarcoux, line 50: This could also be possible with `NotificationsFinder.new.for_notifiable_type('requests').count`. It's only returning unread notifications though.

src/api/app/jobs/measurements_job.rb
saraycp: changed

src/api/app/jobs/measurements_job.rb
saraycp: In this case, I wouldn't like to return only `unread` notifications. And I don't think it's worth to create a new finder for that.

src/api/app/jobs/measurements_job.rb
saraycp: In this case, I wouldn't like to return only `unread` notifications. And I don't think it's worth to create a new finder for that.

src/api/app/jobs/measurements_job.rb
saraycp: changed

vpereira (approved):

src/api/app/models/notification.rb
saraycp: @hennevogel this is ready for another review

hennevogel (approved):

PR #10758, opened 2021-02-09
Show correct message when hovering build info elements

eduardoj:
After introducing another element with CSS class 'toggle-build-info' in a previous commit, clicking in one of this elements should change the title for all elements which have this class.

Prevent having one build info element with title 'Click to keep it open' and another element with title 'Click to close it' at the same time.

This is a follow-up of PR #10748.

**Before:**

Peek 2021-02-09 11-58-Build-Info-error

**After:**

Peek 2021-02-09 11-55-Build-Info-fixed

-----------------

### DIFF
--- a/src/api/app/assets/javascripts/webui/buildresult.js
+++ b/src/api/app/assets/javascripts/webui/buildresult.js
@@ -76,7 +76,7 @@ function toggleBuildInfo() { // jshint ignore:line
     var infoContainer = $(this).parents('.toggle-build-info-parent').next();
     $(infoContainer).toggleClass('collapsed');
     $(infoContainer).removeClass('hover');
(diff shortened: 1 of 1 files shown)


eduardoj:
This fix can be tested in the [following link](https://obs-reviewlab.opensuse.org/eduardoj-fix_build_info_help_hover_message/package/show/home:Admin/package_1612881742_1) from the review-app, after clicking in "Hide excluded/disabled results".

dmarcoux (approved):
Nice! I totally forgot about that.

eduardoj:
> Nice! I totally forgot about that.

I realized about this after I reviewed the previous PR. :man_shrugging:

Issue #10695, opened 2021-01-26
Address issues of the POST /distributions API endpoint


dmarcoux:
<strike>The _POST /distributions API endpoint_ doesn't work.</strike> _It was fixed in #10762._

The _POST /distributions API endpoint_ is not documented in the old and new API documentations.

The _PUT /distributions_ API endpoint is already handling creation of distributions. This is confusing as _PUT_ endpoints are normally used to update data.

We need to refactor both endpoints to have the _POST_ only create distributions and the _PUT_ only update distributions. This would need to be done in a new API version.

dmarcoux:
For now, let's not tackle the _PUT_ endpoint.

PR #10762, opened 2021-02-10
Fix distributions POST endpoint

jcronenberg:
This PR is related to #10695
It fixes the POST endpoint of /distributions
It also introduces a spec to test the endpoint

To verify this feature:
1. Start dev environment
2. Look at current distributions `curl --user Admin:opensuse http://api.localhost:3000/distributions`
3. Make use of the POST /distributions endpoint for example like this:
Write these contents into a file e.g. `post.yml`:
```
<distributions>
  <distribution vendor="opensuse" version="Tumbleweed">
    <name>openSUSE Tumbleweed</name>
    <project>openSUSE:Factory</project>
    <reponame>openSUSE_Tumbleweed</reponame>
    <repository>snapshot</repository>
    <link>http://www.opensuse.org/</link>
    <icon width="8" height="8" url="https://static.opensuse.org/distributions/logos/opensuse.png"/>
    <icon width="16" height="16" url="https://static.opensuse.org/distributions/logos/opensuse.png"/>
    <architecture>i586</architecture>
    <architecture>x86_64</architecture>
  </distribution>
</distributions>
```
After that:
`curl --user Admin:opensuse -X POST --data-binary @post.yml http://api.localhost:3000/distributions`

4. Look at the distributions again as described in step 2 and confirm a new one has been added.

-----------------

### DIFF
--- a/src/api/app/controllers/distributions_controller.rb
+++ b/src/api/app/controllers/distributions_controller.rb
@@ -4,6 +4,7 @@ class DistributionsController < ApplicationController
 
   validate_action index: { method: :get, response: :distributions }
   validate_action upload: { method: :put, request: :distributions, response: :status }
+  validate_action create: { method: :post, request: :distributions }
 
   # GET /distributions
   # GET /distributions.xml
@@ -55,17 +56,9 @@ class DistributionsController < ApplicationController
   # POST /distributions
   # POST /distributions.xml
   def create
-    @distribution = Distribution.new(params[:distribution])
+    Distribution.parse(Xmlhash.parse(request.body.read), delete_current: false)
 
-    respond_to do |format|
-      if @distribution.save
-        format.xml  { render xml: @distribution, status: :created, location: @distribution }
-        format.json { render json: @distribution, status: :created, location: @distribution }
-      else
-        format.xml  { render xml: @distribution.errors, status: :unprocessable_entity }
-        format.json { render json: @distribution.errors, status: :unprocessable_entity }
-      end
-    end
+    render_ok
   end
 
   # DELETE /distributions/opensuse-11.4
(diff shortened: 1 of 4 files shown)


src/api/app/controllers/distributions_controller.rb
```diff
@@ -55,17 +56,9 @@ def upload
   # POST /distributions
   # POST /distributions.xml
   def create
-    @distribution = Distribution.new(params[:distribution])
+    Distribution.parse(Xmlhash.parse(request.body.read), delete_current: false)
```
vpereira, line 59: I don't think the method parse should call `create`. Maybe instead return 'Distribution#new' and then you try to save in this controller and keep the `respond_to` block the same?

vpereira (changes requested):

src/api/app/controllers/distributions_controller.rb
jcronenberg: So this was actually kind of the first thing I proposed. @dmarcoux suggested this approach as the point of this PR isn't to refactor the code, but instead to reuse what is already present. The whole controller needs to be refactored anyway in the future and this was meant to just to get the POST endpoint working while modifying the existing code for other methods as little as possible.

vpereira: > but instead to reuse what is already present

Therefore IMO, if you have the `Distribution.parse` returning `Distribution.new()`, you don't have to change anything on the controller.

jcronenberg: In the controller, yes. But it would require more changes to the model to do it this way without changing the behaviour of PUT.

eduardoj (approved):

danidoni (approved):

PR #10770, opened 2021-02-11
Mark all feature specs under dist directory as type: :feature

danidoni:
Doing so, the module under dist/t/spec/support/wait_for_ajax gets included
automatically on those feature specs

-----------------

### DIFF
--- a/dist/t/spec/features/0010_authentication_spec.rb
+++ b/dist/t/spec/features/0010_authentication_spec.rb
@@ -1,6 +1,6 @@
 require "spec_helper"
 
-RSpec.describe "Authentication" do
+RSpec.describe "Authentication", type: :feature do
   after(:example) do
     logout
   end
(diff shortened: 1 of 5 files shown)


krauselukas (approved):

dist/t/spec/features/0010_authentication_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "spec_helper"
 
-RSpec.describe "Authentication" do
+RSpec.describe "Authentication", type: :feature do
```
hennevogel, line 3: If you mark all anyway, can you just remove the condition to the include?

https://github.com/openSUSE/open-build-service/blob/master/dist/t/spec/support/wait_for_ajax.rb#L17

hennevogel (changes requested):

dist/t/spec/features/0010_authentication_spec.rb
danidoni: Done!

hennevogel (approved):

PR #10785, opened 2021-02-17
Update all of rails to version 5.2.4.5

hennevogel:
`bundle update --patch --strict rails`


eduardoj (approved):
LGTM.

- Having the _Ignore "HTTP Verb Confusion" as we did in master_ commit applied, it will be convinient to backport 24f99f16e86a410a4321b4175f59c8ff52c34507 as well.
- The last commit ("Refactor validator test") should be applied to `master` branch.

hennevogel:
> The last commit ("Refactor validator test") should be applied to master branch.

This commit is a cherry-pick from master.

eduardoj:
> > The last commit ("Refactor validator test") should be applied to master branch.
> 
> This commit is a cherry-pick from master.

Ops, true: ce0b4aac363f33d4a648c7fdb0fcecbc252fbefd

src/api/lib/opensuse/validator.rb
```diff
@@ -84,7 +84,7 @@ def validate(opt, content)
         if File.exist?(schema_base_filename + '.rng')
           schema = Nokogiri::XML::RelaxNG(File.open(schema_base_filename + '.rng'))
         elsif File.exist?(schema_base_filename + '.xsd')
-          schema = Nokogiri::XML::Schema(File.open(schema_base_filename + '.xsd'))
+          schema = Nokogiri::XML::Schema(File.open(schema_base_filename + '.xsd'), Nokogiri::XML::ParseOptions.new.nononet)
```
krauselukas, line 87: This change is not needed, our schemas do not include external entities, so network access for parsing is required. We are not doing it on master https://github.com/openSUSE/open-build-service/blob/50ef5823d66965dee95918f54e3d58fe70590bb4/src/api/app/lib/suse/validator.rb#L86 where we run the same version of nokogiri. It's not causing issues if we enable it, but I would stick with the defaults unless its needed

krauselukas (approved):

danidoni (approved):

PR #10786, opened 2021-02-17
Use a common http user agent

hennevogel:
We run into our own robot/spider protection otherwise.


hennevogel:
As you can see on https://openqa.opensuse.org/tests/1633432#step/rspec_webui_tests/13

danidoni (approved):

PR #10794, opened 2021-02-20
Set build flags in branches from original projects

### LINKED ISSUE #10793: Build flags in branches are inconsistent with the original project
# Issue Description
When branching a package the build flags are being ignored and package is being built for any architecture that is available in the original project.

Expected Result
============
Branched project is using the same set of build flags as in the original project.

How to Reproduce
===============
1. Go to the package view of the original project
2. Click on 'Branch package' and confirm branching
3. Go to the meta view of a branched project
4. Confirm that there are no build flags in meta

Further Information
===============
* OBS version 2.10.7

ar-lex:
Build a branched package only for repositories and architectures which are explicitly enabled in original
project avoiding unnecessary wrong builds, failures and unresolvable packages.

Fixes #10793

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1090,17 +1090,17 @@ class Project < ApplicationRecord
     # Copy the flags from the other project, adjusting them appropriately
     # for this one being a branch of it:
     #
-    # - enable building
+    # - always use the same 'build' flags
     # - disable 'publish' to save space and bandwidth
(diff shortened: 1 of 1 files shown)


adrianschroeter:
this be incompatible, inconsistent and break a lot, esp. maintenance. sorry, flags are the wrong way for this.

adrianschroeter:
but this code would also not do what the issue speaks about if I see that right.

PR #10798, opened 2021-02-23
Fix clear filters link on profiles

### LINKED ISSUE #10604: Wrong Link "Clear current search (...)" on the User's Involved Packages/Projects
# Issue Description
As a user taking part in the beta program, I want to clear the current search filters on the profile of another user. Right now, it does work, but clicking on the link `Clear current search (...)` redirects me to my own user profile. 

Expected Result
============
Clicking on the link `Clear current search (...)` should not redirect me to another user profile. It should simply do what it says, clear the current search.

How to Reproduce
===============
1. Login as a user taking part in the beta program
2. Visit the user profile of another user. Example: https://build.opensuse.org/users/dmarcoux
3. Choose a filter in the "Projects / Packages" or "Roles" filters and click on the `Apply Filter` button
4. See the link `Clear current search (...)`, then click on it.
5. You are redirected to your own user profile.

Further Information
===============
Example with a random user: 
wrong-link-to-clear-search-query

At the time of writing this issue, this link is created in the helper [display_applied_filters](https://github.com/openSUSE/open-build-service/blob/bdd155be180222bf6b45375ab1b7e49dc7c55b47/src/api/app/helpers/webui/user_helper.rb#L2-L11) from `Webui::UserHelper`.

jcronenberg:
Fix the "Clear current search ..." link when looking at another user's profile.
Fixes #10604

To verify this feature:

1. Start dev environment and make sure to start sphinx with `rake ts:start`
2. Log in as Admin and make sure 'Public beta program' is activated
3. Create another user
4. Go to that user's profile
5. Apply a filter
6. Click the "Clear current search..." link and observe that it now links the correct profile

-----------------

### DIFF
--- a/src/api/app/helpers/webui/user_helper.rb
+++ b/src/api/app/helpers/webui/user_helper.rb
@@ -1,12 +1,13 @@
 module Webui::UserHelper
-  def display_applied_filters(filters)
+  def display_applied_filters(filters, user)
     tag.span(class: 'd-block ml-4 mb-3') do
-      link_to(User.session!) do
-        tag.i(class: 'fas fa-trash-alt') do
-          tag.b(class: 'pl-2') do
+      link_to(user) do
+        safe_join(
+          [
+            tag.i(class: 'fas fa-trash-alt pr-2'),
             "Clear current search #{display_filters(filters.keys).to_sentence}"
-          end
-        end
+          ]
+        )
       end
     end
   end
--- a/src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
+++ b/src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
@@ -51,7 +51,7 @@
             Apply Filter
 
     - if display_filters?(filters.keys)
-      = display_applied_filters(filters)
+      = display_applied_filters(filters, displayed_user)
   .card-body.pt-0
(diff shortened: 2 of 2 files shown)


dmarcoux:
Beside the broken link, its style is also wrong.

weird-styling-link

It currently looks like this, the text shouldn't be bold and after having a look at the HTML, the text should be on the same level as the `<i>` tag, not inside. So with those changes, it would look like this:

link

src/api/app/helpers/webui/user_helper.rb
```diff
@@ -1,12 +1,15 @@
...
-      link_to(User.session!) do
-        tag.i(class: 'fas fa-trash-alt') do
-          tag.b(class: 'pl-2') do
-            "Clear current search #{display_filters(filters.keys).to_sentence}"
-          end
-        end
+      link_to(user) do
+        safe_join(
+          [
+            tag.i(class: 'fas fa-trash-alt'),
```
dmarcoux, line 7: We can remove the `span` and instead add the padding (`pr-2`) to the `i`. The text can be directly outputted.

```suggestion
            tag.i(class: 'fas fa-trash-alt pr-2'),
            "Clear current search #{display_filters(filters.keys).to_sentence}"
```

dmarcoux (approved):

PR #10820, opened 2021-03-01
Add the build results badge for the package

### LINKED ISSUE #2908: Badge for github integration
This is a feature request for a badge which shows the build-status on a github project. Would really be great!

hellcp:
This adds integration badges for packages

To verify this feature visit `/package/badge/:project/:package.svg` (with optional `percent` and `ratio` `type` parameters)

/package/badge/:project/:package.svg
gnome-mahjongg
/package/badge/:project/:package.svg?type=percent
type=percent
/package/badge/:project/:package.svg?type=ratio
type=ratio

Merging this requires a new package to be present on the host system: dejavu-fonts. I temporarily added it to Dockerfiles, since otherwise this can't really be built ;)

Fixes #2908


dist/obs-bundled-gems.spec
```diff
@@ -42,6 +42,7 @@ BuildRequires:  make
 BuildRequires:  mysql-devel
 BuildRequires:  nodejs
 BuildRequires:  python-devel
+BuildRequires:  ImageMagick-devel
```
hennevogel, line 45: I'm not so sure we want to introduce this dependency. ImageMagick has not the greatest [track record regarding security](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=imagemagick). Especially for releases.

hennevogel: https://caniuse.com/svg looks pretty good right? Do we need png?

hellcp: Sure, that still leaves calculating width of text for rendering svg though

hennevogel: > calculating width of text for rendering svg though

Why would we need rmagick for that?

hellcp: Calculating width of text otherwise is pretty complex, because rendering text in the first place is pretty complex. We could use per-character font widths based on ttf, but then this means compound glyphs aren't measured properly (technically in case of latin alphabet that works out, since we will in most cases have more space than needed than less space)

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -636,6 +636,31 @@ def binary_download
     end
   end
 
+  def badge
+    @badge = []
+    info = { text: 'build service', colour: 'gray' }
```
Conan-Kudo, line 641: `colour` -> `color` please.

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -897,4 +922,24 @@ def set_job_status
       @buildtime = nil
     end
   end
+
+  def text_width(text)
+    file = TTFunk::File.open('/usr/share/fonts/truetype/DejaVuSans.ttf')
```
Conan-Kudo, line 927: Is there a way we can expose this path to be configurable with `Makefile.include`?

In Fedora, this is stored as `/usr/share/fonts/dejavu-sans-fonts/DejaVuSans.ttf`.

Conan-Kudo (changes requested):

marxin:
Any chance this can be rebased and eventually merged?

hellcp:
Closed in favour of  #13169 (nice)

PR #10828, opened 2021-03-04
removed-gendered-terms

gonyere:
Replacing gendered terms 'he', 'his', 'guy' and 'he/she' with 'they', 'their' and 'person'. 

-----------------

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -151,7 +151,7 @@ XmlResult: status
 
 Parameters:
   cmd: change_password
-  cmd: lock             # will lock the user and his home projects
+  cmd: lock             # will lock the user and their home projects
   cmd: delete           # will mark the user as deleted and remove her home projects
 
 GET /person/<userid>/token
@@ -1683,7 +1683,7 @@ GET /search/owner
 Parameters:
 
   binary: specify the binary package to search for
-  user: specify a user login name to list his packages
+  user: specify a user login name to list their packages
   group: specify a group title name to list their packages
 
   devel: true/false: include devel package definitions?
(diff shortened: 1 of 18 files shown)


dmarcoux (approved):
Thank you for the PR

marcus-h:
On 2021-03-03 17:10:44 -0800, Emily Gonyer wrote:
> Replacing gendered terms &#39;he&#39;, &#39;his&#39;, &#39;guy&#39; and &#39;he/she&#39; with &#39;they&#39;, &#39;their&#39; and &#39;person&#39;.
>
Maybe a completely stupid question from non-native speaker: why is a
former singular pronoun switched to a plural one? For instance,

> -    # we don't need to show the requester if he is the same as the creator
> +    # we don't need to show the requester if they are the same as the creator

I would have expected something like "we don't need to show the requester
if it is the same as the creator". Mixing singular and plural sounds a
bit "strange" (again, I'm not a native speaker:) ).

dmarcoux:
@marcus-h It's not a stupid question :slightly_smiling_face:

This Wikipedia article explains well the usage of `they` in singular: https://en.wikipedia.org/wiki/Singular_they

marcus-h:
On 2021-03-04 13:04:10 +0000, Dany Marcoux wrote:
> @marcus-h It's not a stupid question :slightly_smiling_face:
>
> This Wikipedia article explains well the usage of `they` in singular: https://en.wikipedia.org/wiki/Singular_they
>
Great article, thanks a lot! A stupid reader, like me, is probably
confused when seeing the singular they for the first time but after
reading the Wikipedia article they will get it;) (I hope, I used it
correctly:) )

dmarcoux:
There is no need to call yourself stupid, I also found it strange the first time! :slightly_smiling_face: English is also not my native language. I'm glad to know that the article helped.

PR #10835, opened 2021-03-05
Replace `require_login` with Pundit in Webui::Staging::WorkflowsController

intrip:
This is a PR of a series which replaces require_login with Pundit.

Tackles `Webui::Staging::WorkflowsController`

Ref #10083

-----------------

### DIFF
--- a/src/api/app/controllers/webui/staging/workflows_controller.rb
+++ b/src/api/app/controllers/webui/staging/workflows_controller.rb
@@ -1,19 +1,21 @@
 class Webui::Staging::WorkflowsController < Webui::WebuiController
   VALID_STATES_WITH_REQUESTS = [:acceptable, :accepting, :review, :testing, :building, :failed, :unacceptable].freeze
 
-  before_action :require_login, except: [:show]
+  # TODO: Remove this when we'll refactor kerberos_auth
+  before_action :kerberos_auth, except: [:show]
   before_action :set_project, only: [:new, :create]
   before_action :set_workflow_project, except: [:new, :create]
   before_action :set_staging_workflow, except: [:new, :create]
-  after_action :verify_authorized, except: [:show, :new]
+  after_action :verify_authorized, except: [:show]
 
   def new
     if @project.staging
+      authorize @project.staging
       redirect_to staging_workflow_path(@project)
       return
     end
 
-    @staging_workflow = @project.build_staging
+    @staging_workflow = authorize @project.build_staging
   end
 
   def create
--- a/src/api/app/policies/staging/workflow_policy.rb
+++ b/src/api/app/policies/staging/workflow_policy.rb
@@ -1,6 +1,10 @@
 class Staging::WorkflowPolicy < ApplicationPolicy
-  def initialize(user, record)
-    super(user, record, user_optional: true)
+  def initialize(user, record, opts = {})
+    super(user, record, opts.reverse_merge(ensure_logged_in: true))
+  end
+
+  def new?
+    create?
   end
 
   def create?
--- a/src/api/app/views/webui/project/_tabs.html.haml
+++ b/src/api/app/views/webui/project/_tabs.html.haml
@@ -24,7 +24,7 @@
         = tab_link('Staging', staging_workflow_path(project.staging_workflow.project), false, 'scrollable-tab-link')
       - elsif project.staging&.persisted?
         = tab_link('Staging', staging_workflow_path(project), active, 'scrollable-tab-link')
-      - elsif policy(Staging::Workflow.new(project: project)).create?
+      - elsif Staging::WorkflowPolicy.new(pundit_user, Staging::Workflow.new(project: project), ensure_logged_in: false).create?
         = tab_link('Staging', new_staging_workflow_path(project: project), active, 'scrollable-tab-link')
 - else
   .bg-light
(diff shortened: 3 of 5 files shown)


src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -1,7 +1,9 @@
 class Webui::Staging::WorkflowsController < Webui::WebuiController
   VALID_STATES_WITH_REQUESTS = [:acceptable, :accepting, :review, :testing, :building, :failed, :unacceptable].freeze
 
-  before_action :require_login, except: [:show]
+  before_action :check_authorizable, except: [:show]
```
intrip, line 4: In order to `authorize` we need a [user, record] pair, this is a problem when we need to deal with a not found record: if we return a `404` whilst fetching the record ( before `authorize` ) we leak the existence of the records to anonymous users (which I don't think it's the desired behaviour).

This implementation does 2 checks:
1. checks for the user presence by relying in the `initializer` `ensure_logged_in` check
2. runs the `authorize` as usual afterwards

This approach Is also a very easy plug in replacement of `require_login` with the Policy.

*If leaking resource existence to anonymous users is not an issue then we can keep the single authorize call and the `raise ActiveRecord::RecordNotFound` logic.*

intrip:
@dmarcoux could you take a peek? 🙏

src/api/app/controllers/webui/staging/workflows_controller.rb
dmarcoux: > _If leaking resource existence to anonymous users is not an issue then we can keep the single authorize call and the `raise ActiveRecord::RecordNotFound` logic._

It's not an issue. Let's keep the single `authorize` call and the `raise ActiveRecord::RecordNotFound` logic.

dmarcoux:
As always, thank you @intrip for submitting this PR. I just replied to your comment. I'll have a look again once the PR is updated.

intrip:
@dmarcoux than you for the quick response 😄 .

I've revisited the code as suggested, please check again.

src/api/app/policies/staging/workflow_policy.rb
```diff
@@ -1,6 +1,10 @@
 class Staging::WorkflowPolicy < ApplicationPolicy
-  def initialize(user, record)
-    super(user, record, user_optional: true)
+  def initialize(user, record, opts = {})
+    super(user, record, opts.reverse_merge(ensure_logged_in: true))
+  end
+
+  def new?
+    true
```
dmarcoux, line 7: If the user isn't authorized to create a staging workflow, then we shouldn't authorize them to instantiate a new staging workflow. This line would then be:

```suggestion
    create?
```

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -1,13 +1,16 @@
...
+  # TODO: Remove this when we'll refactor kerberos_auth
+  before_action :kerberos_auth, except: [:show]
   before_action :set_project, only: [:new, :create]
   before_action :set_workflow_project, except: [:new, :create]
   before_action :set_staging_workflow, except: [:new, :create]
-  after_action :verify_authorized, except: [:show, :new]
+  after_action :verify_authorized, except: [:show]
 
   def new
+    authorize ::Staging::Workflow, :new?
```
dmarcoux, line 12: If we have this line right under `@staging_workflow = @project.build_staging` (on line 19 right now), we could check if the user can instantiate the new staging workflow.

intrip: Thank you for the suggestion!

src/api/app/controllers/webui/staging/workflows_controller.rb
```diff
@@ -1,7 +1,8 @@
 class Webui::Staging::WorkflowsController < Webui::WebuiController
   VALID_STATES_WITH_REQUESTS = [:acceptable, :accepting, :review, :testing, :building, :failed, :unacceptable].freeze
 
-  before_action :require_login, except: [:show]
+  # TODO: Remove this when we'll refactor kerberos_auth
+  before_action :kerberos_auth, except: [:show]
   before_action :set_project, only: [:new, :create]
   before_action :set_workflow_project, except: [:new, :create]
   before_action :set_staging_workflow, except: [:new, :create]
```
dmarcoux, line 8: The line just below this one should be changed to verify that `authorize` was called in `new`. I think you had this change before. It's somehow gone.

dmarcoux:
I've got one more comment, but it's already looking great.

src/api/app/controllers/webui/staging/workflows_controller.rb
intrip: I've added it back because [when we redirect](https://github.com/openSUSE/open-build-service/blob/bab35af63a6d1418424d37d3292662315a324068/src/api/app/controllers/webui/staging/workflows_controller.rb#L13) we do not call `authorize` hence the `verify_authorized` will error.
I suppose calling `authorize` is a duty of the redirected routed at that point, but if you prefer we can add back the `:new` and call `authorize` also on the redirected resource: `authorize @project.staging`.
Please note that we also have an automated test now so the `verify_authorize` is mostly an extra.

Apologies: I should've mentioned this explicitly to save 1 round trip, next time I'll be more explicit.

dmarcoux: Oh, I didn't think about that. It's not obvious at all at a first glance. I think it's better to be explicit here, even if it means calling `authorize` before the `redirect_to` call, so it would be `if authorize @project.staging` as you suggest.

No worries for not mentioning this, it's always a pleasure to review your PRs.

dmarcoux: Having the `authorize` inside the `if` is completely fine.

dmarcoux (approved):
As always, great work! Keep them coming :smile:

intrip:
> As always, great work! Keep them coming 😄

Thank you! Have a nice day!

dmarcoux:
@intrip You too!

PR #10842, opened 2021-03-09
Improve how build status details are displayed

### LINKED ISSUE #10837: unresolvable mouse over shows help instead of actual error
Looking at https://build.opensuse.org/package/show/openSUSE:Factory:Staging:N/wicked it's unresolvable. Moving the mouse over the unresolvable text shows the help instead of the the actual error message from the obs backend though. Also clicking the text doesn't change that.

unresolvable

MozillaFirefox-78.7.0-lp152.2.43.1.x86_64

dmarcoux:
Add a bit of spacing to separate different information and use colors to highlight important information about the build state.

Fixes #10837

Before:
before

Now:
after

### DIFF
--- a/src/api/app/views/webui/package/responsive_ux/_buildstatus.html.haml
+++ b/src/api/app/views/webui/package/responsive_ux/_buildstatus.html.haml
@@ -53,8 +53,8 @@
                     %i.fas.fa-clock.text-warning
                     This result is outdated
                 - if result.details
-                  %div
-                    %strong #{result.code}:
+                  .pt-1.mt-3
+                    %strong{ class: "build-state-#{result.code}" } #{result.code}:
                     %span= result.details
 
         - previous_repo = result.repository


dmarcoux:
@lnussel Please have a look at the screenshots in the PR description. Could you let me know if this is clearer to you?

dmarcoux:
> Looks better but isn't it a bit weird that the arrow points at the question mark? 

Do you mean this?
icon

If you do, the arrow points to the question mark icon since the icon is the only element consistently present throughout the build results section in the project and package pages. The build status text is sometimes linking to the build job logs or the project monitor page. So we have to link the popover (so the grey box which appears when clicking/hovering on the question mark icon or build state when possible) to an element which is always present.

> Maybe for the purpose here an :information_source:  _after_ the text would be better suited.

As for the position of the icon, we want to put the context-sensitive help with the question mark icon after the text. It's even in our [pattern library](https://obs-patterns.netlify.app/molecules/context-sensitive-help.html). To be 100% honest though, we're not always consistent with this. It's sometimes before the text. We tend to use the icon i-icon for information which is always displayed (so not context-sensitive help), like this taken from the project page overview:

i-icon-example

Anyway, I'll look at that text and icon position in another PR to not increase the scope.

> What does "build jobs exist" mean?

This is the description of the repository status. The message kind of sucks to be honest, I can understand why it's confusing. I just submitted another PR (#10843) to improve this.

I think there is simply too much information in that popover. For now, I want to be conservative in the changes here since this is used by a lot of people on OBS.

danidoni (approved):

PR #10848, opened 2021-03-10
Improve staging workflow delete form

dmarcoux:
Just something I've noticed while reviewing #10835.

- Use Bootstrap checkboxes (as defined in our patterns library)
- The label of checkboxes are now clickable (this is possible now that
  each checkbox has a unique ID)


saraycp (approved):
LGTM

Issue #10170, opened 2020-09-18
Styles are broken when running feature specs with the capybara mobile driver


danidoni:
# Styles are broken when running feature specs with capybara
Our CircleCI runs some of the beta feature specs under mobile emulation. To do so, it sets the `CAPYBARA_DRIVER=mobile` environment variable before running the feature specs.

While troubleshooting [some errors in our specs](https://app.circleci.com/pipelines/github/saraycp/open-build-service/688/workflows/52fc22bf-b364-4c2a-919b-8f37866c96cb/jobs/4065/parallel-runs/1) we found out that capybara was unable to click on some elements from a page because they where below an oversized Open Build Service logo:

![capybara-screenshot](https://user-images.githubusercontent.com/2650/93601702-e093ac00-f9c1-11ea-84fe-84a5b9c0537f.png)

The logo size is clearly wrong. It is set via CSS rules, and by looking at the rest of the elements in that page, looks like some other CSS rules are broken too. 

The broken styles are interfering with the proper execution of the feature specs. On mobile environments the oversized logo covers the half of the screen, hiding elements to the Capybara runner.

Note that Using the `:desktop` capybara driver also breaks the styles. Note the oversized logo and the search box in the capture below.

![Screenshot from 2020-09-18 14-25-16](https://user-images.githubusercontent.com/2650/93601950-38321780-f9c2-11ea-9c3e-80749dfa7f97.png)

Expected Result
============
The capybara driver should not break the styles of the application.

How to Reproduce
===============
1. Put a `docker-compose.selenium.yml` in the root of the project. You can find the contents for the file below.
2. Use this command line to open a session inside the dockerized project: `docker-compose -f docker-compose.yml -f docker-compose.selenium.yml run --service-ports --use-aliases --rm frontend bash -l`
3. Open a Vnc viewer (I used TigerVNC Viewer) pointing to localhost.
3. Run `export CAPYBARA_DRIVER=mobile` inside the dockerized session.
3. Run `rspec spec/features/beta/webui/projects/status_spec.rb` inside the dockerized session.
4. The application under test shows broken styles.

Further Information
===============

Use this `docker-compose.selenium.yml` to reproduce the issue.

```docker
version: "2.1"
services:
  frontend:
    depends_on:
      - selenium
    environment:
      RSPEC_HOST: frontend
  selenium:
    image: selenium/standalone-chrome-debug
    environment:
      VNC_NO_PASSWORD: 1
    ports:
      - 5900:5900
[...]

PR #10851, opened 2021-03-10
Set the responsive_ux layout as the default layout

dmarcoux:
The responsive_ux layout was renamed to webui and the "old" webui layout has thus been overwritten, it's now gone.

*Note: Views from the "old" webui layout will be removed in following PRs. I try to keep PRs small for easy reviews.*


dmarcoux:
I'm hitting the issue #10170... I'll look into why it's happening.

dmarcoux:
[Remove responsive_ux feature flag after the rollout](https://trello.com/c/KlHFqSGS/1626-remove-responsiveux-feature-flag-after-the-rollout)

dmarcoux:
We're going to tackle this more towards the end since it's just too big to start with. Too many specs are failing.

dmarcoux:
After looking more into this, not having the layout is going to be a blocker so we need to handle this now.

dmarcoux:
Sorry for the somewhat big PR... it's difficult to make this smaller since it's a all or nothing approach when switching the layout.

eduardoj (approved):

PR #10855, opened 2021-03-11
Move project edit action close to the target

danidoni:
Following the convention used for editing the details of a kiwi image, we're moving the "Edit Project" button close to the edited details.

**Before**
image

**After**
image

-----------------

### DIFF
--- a/src/api/app/views/webui/project/_basic_info.html.haml
+++ b/src/api/app/views/webui/project/_basic_info.html.haml
@@ -11,3 +11,9 @@
       %i No description set
     - else
       = render partial: 'webui/shared/collapsible_text', locals: { text: project.description }
+
+  - if policy(project).update?
+    .pt-4
+      = link_to(edit_project_path(project), class: 'nav-link', remote: true, title: 'Edit') do
+        %i.fas.fa-edit.text-secondary
+        %span.nav-item-name Edit
--- a/src/api/spec/features/webui/projects_spec.rb
+++ b/src/api/spec/features/webui/projects_spec.rb
@@ -15,20 +15,44 @@ RSpec.describe 'Projects', type: :feature, js: true do
     expect(page).to have_css('h3', text: project.title)
   end
 
-  it 'changing project title and description' do
-    login user
-    visit project_show_path(project: project)
+  describe 'changing project title and description' do
+    context 'when accepting the changes' do
+      it 'updates the project title, description and url' do
+        login user
+        visit project_show_path(project: project)
 
-    desktop? ? click_link('Edit Project') : click_menu_link('Actions', 'Edit Project')
-    expect(page).to have_text("Edit Project #{project}")
+        click_on('Edit')
+        expect(page).to have_text("Edit Project #{project}")
 
-    fill_in 'project_title', with: 'My Title hopefully got changed'
-    fill_in 'project_description', with: 'New description. Not kidding.. Brand new!'
-    click_button 'Update'
+        fill_in 'project_title', with: 'My Title hopefully got changed'
+        fill_in 'project_description', with: 'New description. No kidding.. Brand new!'
+        fill_in 'project_url', with: 'https://test.url'
+        click_button 'Update'
+        wait_for_ajax
 
-    visit project_show_path(project: project)
-    expect(find(:id, 'project-title')).to have_text('My Title hopefully got changed')
-    expect(find(:id, 'description-text')).to have_text('New description. Not kidding.. Brand new!')
+        expect(find(:id, 'project-title')).to have_text('My Title hopefully got changed')
+        expect(find(:id, 'description-text')).to have_text('New description. No kidding.. Brand new!')
+        expect(page).to have_text('https://test.url')
+      end
+    end
+
+    context 'when cancelling the changes' do
+      it "renders back the original project's details" do
+        login user
+        visit project_show_path(project: project)
+
+        click_on('Edit')
+        expect(page).to have_text("Edit Project #{project}")
+
+        fill_in 'project_title', with: 'My Title hopefully got changed'
+        fill_in 'project_description', with: 'New description. No kidding.. Brand new!'
+        click_link 'Cancel'
+        wait_for_ajax
+
+        expect(page).to have_text(project.title)
+        expect(page).to have_text(project.description)
+      end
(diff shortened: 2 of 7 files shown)


dmarcoux:
We need to split this up per field: Title, URL and Description

dmarcoux:
@danidoni and I just talked again about this and since the fields aren't always visible, we'll have to go with the generic "Edit Project" link.

src/api/app/views/webui/project/_basic_info.html.haml
```diff
@@ -11,3 +11,9 @@
       %i No description set
     - else
       = render partial: 'webui/shared/collapsible_text', locals: { text: project.description }
+
+  - if User.possibly_nobody.can_modify?(project) && feature_enabled?(:responsive_ux)
```
dmarcoux, line 15: We have a [Pundit policy for projects](https://github.com/openSUSE/open-build-service/blob/6687073ae4cdd8e1948e501fbc0843e6ac95cf52/src/api/app/policies/project_policy.rb), so we should use it:

```suggestion
  - if policy(project).update? && feature_enabled?(:responsive_ux)
```

danidoni: Ah, I always forget about Pundit... Refactoring!

danidoni:
I think we'll have some overlapping between this PR and the ones dealing with the `responsive_ux` feature flag, right @saraycp, @dmarcoux, @eduardoj ?

eduardoj:
@danidoni:

> I think we'll have some overlapping between this PR and the ones dealing with the `responsive_ux` feature flag, right @saraycp, @dmarcoux, @eduardoj ?

At least, not yet. The work to be done with the `responsive_ux` feature flag has been divided, into layout, projects, packages, ... And the part of "projects" is the only one that I see that could be affected. I work in the "projects" part. I will take care of the possible conflicts.

saraycp:
> I think we'll have some overlapping between this PR and the ones dealing with the `responsive_ux` feature flag, right @saraycp, @dmarcoux, @eduardoj ?

It does, but shouldn't be a big deal to handle the conflicts.

Regarding the action, I've always thought that `Edit Project` can make people think that the action implies much more than just changing title/description/URL. Maybe it's time to consider other texts. What about `Edit Details` or just `Edit`?

src/api/app/views/webui/project/_basic_info.html.haml
```diff
@@ -11,3 +11,9 @@
       %i No description set
     - else
       = render partial: 'webui/shared/collapsible_text', locals: { text: project.description }
+
+  - if policy(project).update? && feature_enabled?(:responsive_ux)
```
dmarcoux, line 15: It's not needed to check for the `responsive_ux` feature flag since we're removing this kind of code right now.

```suggestion
  - if policy(project).update?
```

danidoni: Okay

src/api/app/views/webui/project/_basic_info.html.haml
```diff
@@ -11,3 +11,9 @@
       %i No description set
     - else
       = render partial: 'webui/shared/collapsible_text', locals: { text: project.description }
+
+  - if policy(project).update?
+    .pt-4
+      = link_to(edit_project_path(project), class: 'nav-link', remote: true, title: 'Edit Project') do
```
dmarcoux, line 17: Sorry, I didn't see that in my last review...

```suggestion
      = link_to(edit_project_path(project), class: 'nav-link', remote: true, title: 'Edit') do
```

src/api/spec/features/webui/projects_spec.rb
```diff
@@ -15,20 +15,42 @@
...
-  it 'changing project title and description' do
-    login user
-    visit project_show_path(project: project)
+  describe 'changing project title and description' do
+    context 'when accepting the changes' do
+      it 'updates the project title, description and url' do
+        login user
+        visit project_show_path(project: project)
+
+        click_on('Edit Project')
```
dmarcoux, line 24: This works because the title of the link is still `Edit Project`. Since it's going to be changed, this needs to also change. 

```suggestion
        click_on('Edit')
```

src/api/spec/features/webui/projects_spec.rb
```diff
@@ -15,20 +15,42 @@
...
+
+        expect(find(:id, 'project-title')).to have_text('My Title hopefully got changed')
+        expect(find(:id, 'description-text')).to have_text('New description. Not kidding.. Brand new!')
+      end
+    end
 
-    click_on('Edit Project')
-    expect(page).to have_text("Edit Project #{project}")
+    context 'when cancelling the changes' do
+      it "renders back the original package's details" do
```
dmarcoux, line 38:
```suggestion
      it "renders back the original project's details" do
```

src/api/spec/features/webui/projects_spec.rb
```diff
@@ -15,20 +15,42 @@
...
+        login user
+        visit project_show_path(project: project)
 
-    fill_in 'project_title', with: 'My Title hopefully got changed'
-    fill_in 'project_description', with: 'New description. Not kidding.. Brand new!'
-    click_button 'Update'
+        click_on('Edit Project')
```
dmarcoux, line 42: Same as above.

```suggestion
        click_on('Edit')
```

src/api/app/views/webui/project/_basic_info.html.haml
danidoni: My fault

src/api/spec/features/webui/projects_spec.rb
```diff
@@ -15,20 +15,42 @@
     expect(page).to have_css('h3', text: project.title)
   end
 
-  it 'changing project title and description' do
-    login user
-    visit project_show_path(project: project)
+  describe 'changing project title and description' do
+    context 'when accepting the changes' do
+      it 'updates the project title, description and url' do
```
dmarcoux, line 20: Is it really updating the project's URL? It doesn't seem to be the case.

danidoni: Fixed

saraycp (approved):

PR #10857, opened 2021-03-11
Make involved packages/projects search case-insensitive

### LINKED ISSUE #10833: User Profile Redesign - Search
# Issue Description
The search of the involved projects and packages of the user profile page is case sensitive. This isn't user-friendly since users need to remember the exact case/capitalization of packages/projects in order to find them.

Expected Result
============
The search of the involved projects and packages of the user profile page is case **insensitive**.

How to Reproduce
===============
1. Log in with your user and join the beta program (if you haven't already) 
2. Go to https://build.opensuse.org/users/dmarcoux and search for `BAT`
3. See the empty list, although `dmarcoux` is involved in the packages `X11:Utilities / cbatticon` and `utilities / bat`. Those 2 packages should be listed.

Further Information
===============
* After reading #10323, it's pretty clear that a case-sensitive search is confusing users.

krauselukas:
A case-sensitivity in the search create overhead since the
user needs to remember the exact case/capitalization of
packages/projects in order to find them.

Fixes #10833

**NOTE**: I considered directly using `Arel` to do the query since this way it is possible to get the default behavior of the `LIKE` query which is case-insensitve. Although the collation of our database is set to `utf8mb4_unicode_ci` (so case-insensitve) the `:name` column of project/package is `utf8_bin` (so case-sensitive) on the DB level. I would not change it for now, since this might introduce side-effects in other parts of OBS. I don't expect any performance issues with the provided solution, so I would avoid to use the sledgehammer to crack a nut;

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -986,7 +986,7 @@ class User < ApplicationRecord
     related_items = project_or_package_class.related_to_user(id).where(relationships: { role_id: roles_to_filter }).or(
       project_or_package_class.related_to_group(group_ids).where(relationships: { role_id: roles_to_filter })
     )
-    related_items = related_items.where('name LIKE ?', "%#{filters[:search_text]}%") if filters[:search_text].present?
+    related_items = related_items.where('LOWER(name) LIKE ?', "%#{filters[:search_text].downcase}%") if filters[:search_text].present?
     related_items
   end
 end


krauselukas:
[Issue #10833 @lukaskrause3 ](https://trello.com/c/uEvbwbxb/1629-issue-10833-lukaskrause3)

danidoni (approved):

krauselukas:
Tests in CircleCI are passing, GitHub status didn't got updated

PR #10865, opened 2021-03-12
Merge responsive scrollable tabs

saraycp:
Remove the old version of the new scrollable tabs and move the corresponding CSS.

-----------------


src/api/app/assets/stylesheets/webui/application.scss
```diff
@@ -26,6 +26,7 @@
 @import 'datatables';
 @import 'card-component';
 @import 'breadcrumbs-component';
+@import 'tabs';
```
dmarcoux, line 29: What do you think of merging `tabs` and `tabs-component`? I don't think that we need to have two files for this since they are both for the same component.

dmarcoux:
In total, there are 4 files/views relying on the scrollable tabs. The configuration and project tabs are handled in this PR. The package tabs are handled in #10861 and the [user tabs](https://github.com/openSUSE/open-build-service/blob/758c462e93f58c6e501117557e6178cafb6e5e08/src/api/app/views/webui/user/responsive_ux/_tabs_profile.html.haml) also need to be handled. I don't mind if it's in another PR, but I think it would be nice to do it in this one since we're already doing it for other pages.

saraycp:
> the [user tabs](https://github.com/openSUSE/open-build-service/blob/758c462e93f58c6e501117557e6178cafb6e5e08/src/api/app/views/webui/user/responsive_ux/_tabs_profile.html.haml) also need to be handled

Sorry, I didn't explain that I planned to tackle the user profile tabs in another PR regarding the user. However, I finally decided to extract the changes and I've included them here.

@dmarcoux please have another look. This PR migrates the tabs code of the user profile, configuration and project pages.
As you mentioned, the packages page tabs were tackled in a previous PR.

src/api/app/assets/stylesheets/webui/application.scss
saraycp: Done!

eduardoj (approved):

src/api/app/views/webui/user/_involvement.html.haml
```diff
@@ -1,31 +1,6 @@
...
-        %li.nav-item
-          = link_to('#owned', id: 'owned-tab', class: 'nav-link text-nowrap',
-                    data: { toggle: 'tab' }, role: 'tab', aria: { controls: 'owned', selected: false }) do
-            Owned Projects/Packages
-            %span.badge.badge-primary
-              = owned.size
-        %li.nav-item.dropdown
-          %a.nav-link.dropdown-toggle{ href: '#', 'data-toggle': 'dropdown', role: 'button', 'aria-expanded': 'false', 'aria-haspopup': 'true' }
-          .dropdown-menu.dropdown-menu-right
+  = render partial: 'webui/user/tabs_profile', locals: { involved_packages: involved_packages,
```
dmarcoux, line 2: As we discussed:

```suggestion
  = render partial: 'tabs_profile', locals: { involved_packages: involved_packages,
```

saraycp: In this exact case, we need to specify the path because, by convention, it infers the plural: ~webui/users~.

```
 Showing /obs/src/api/app/views/webui/user/_involvement.html.haml where line #2 raised:

 Missing partial webui/users/_tabs_profile,
```

dmarcoux: Oh, alright then!

PR #10874, opened 2021-03-15
Force html processing for all formats except html, js and json

### LINKED ISSUE #10510: Edit package description: Cancel button not working
Using the OBS beta program, I tried to edit a package description, but then decided not to do so.
I used 'Cancel' to cancel the activity
Expected result: Go back to package overview
Actual result: Nothing happens (stays in edit mode)

danidoni:
The default html format makes the Cancel button to be processed wrongly as html
instead of js.

This change allows the Cancel button to work as ajax, as well as the rendering
of package pages for packages with dots in its name like
"yaml-cpp.openSUSE_Leap_15.2_Update".

Fixes #10510

-----------------

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -37,6 +37,8 @@ class Webui::PackageController < Webui::WebuiController
 
   after_action :verify_authorized, only: [:new, :create, :remove_file, :remove, :abort_build, :trigger_rebuild, :wipe_binaries, :save_meta, :save, :abort_build]
 
+  before_action :force_html_mime_type
+
   def index
     render json: PackageDatatable.new(params, view_context: view_context, project: @project)
   end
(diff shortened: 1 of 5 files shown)


dmarcoux:
I would be interested to know why is this needed for a package, but not for a project. Why?

danidoni:
> I would be interested to know why is this needed for a package, but not for a project. Why?

Because we're forcing a specific format [for packages](https://github.com/openSUSE/open-build-service/blob/master/src/api/config/routes/webui_routes.rb#L61) but not [for projects](https://github.com/openSUSE/open-build-service/blob/master/src/api/config/routes/webui_routes.rb#L188).

So a project can choose whatever format it needs for the response, but a project is forced to be treated as an `:html` no matter what.

danidoni:
I've opted for closing this PR in favor of #10879, because the new changes imply moving the Edit link close to the edited fields.

PR #10880, opened 2021-03-15
Introduce mobile tests for all feature tests 

eduardoj:
Before, only tests under `features/beta/` were tested also with the mobile viewport. After this pull request, all feature tests will be tested with two viewports: desktop and mobile.

Only tests that were failing with mobile were moved out from beta.

Some remaining tests inside `features/beta/` were not moved outside, as they were not failing. They will be migrated in a following pull request, to not make this one so difficult to review.

-----------------


.circleci/config.yml
```diff
@@ -265,23 +265,16 @@ jobs:
...
             if test "$CIRCLE_NODE_INDEX" = 0; then
               message="Running feature tests"
-              circleci tests glob 'spec/features/**/*_spec.rb' | grep -v 'spec/features/beta/' > $pickfile
             else
-              message="Running beta feature tests"
-              circleci tests glob 'spec/features/beta/**/*_spec.rb' > $pickfile
-            fi
-
-            if test "$CIRCLE_NODE_INDEX" = $((CIRCLE_NODE_TOTAL-1)); then
               message="Running beta feature tests for mobile"
```
dmarcoux, line 272:
```suggestion
              message="Running feature tests for mobile"
```

.circleci/config.yml
```diff
@@ -265,23 +265,16 @@ jobs:
           name: Run rspec feature tests
           command: |
             cd src/api
-            pickfile=log/pick.$CIRCLE_NODE_INDEX.list
 
             if test "$CIRCLE_NODE_INDEX" = 0; then
               message="Running feature tests"
```
dmarcoux, line 270:
```suggestion
              message="Running feature tests for desktop"
```

.circleci/config.yml
eduardoj: Ops, applied.

.circleci/config.yml
eduardoj: Applied, thanks.

dmarcoux (approved):

PR #10895, opened 2021-03-16
Remove non-responsive_ux code from helpers

dmarcoux:
repository_info is still in use since it's for the responsive UI.

responsive_namespace isn't anymore in use, so it can be safely removed.

### DIFF
--- a/src/api/app/helpers/webui/webui_helper.rb
+++ b/src/api/app/helpers/webui/webui_helper.rb
@@ -109,12 +109,9 @@ module Webui::WebuiHelper
 
     repo_state_class = repository_state_class(outdated, status)
 
-    data_options = {}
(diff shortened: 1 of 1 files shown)


src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -380,7 +372,7 @@ def link_content(text, css_class, icon)
     if css_class && css_class.include?('nav-link')
       capture do
         concat(tag.i('', class: "fas #{icon}"))
-        concat(tag.div(text, class: feature_enabled?(:responsive_ux) ? '' : 'small'))
+        concat(tag.div(text))
```
dmarcoux, line 375: Note for reviewers: This helper `link_content` is used for the `Sign Up` and `Log In` links. How they are displayed shouldn't have changed at all.

saraycp (approved):

PR #10896, opened 2021-03-16
Enable feature flags for beta users in rake task dev:bootstrap

dmarcoux:
It's easier to join the beta and test changes related to feature flags while also being able to test changes for non-beta users, so without any feature flag enabled.


danidoni (approved):

saraycp (approved):
I totally agree. It's the best way to test all the scenarios.

eduardoj (approved):

PR #10897, opened 2021-03-16
Move the rest of beta tests out from beta

eduardoj:
Migrate all tests from `spec/features/beta/` that already are tested in `spec/features/`.

The remaining tests left in `spec/features/beta/` are not moved because they are testing the current beta features: "notifications_redesign" and "user_profile_redesign".

-----------------


dmarcoux (approved):

eduardoj:
@dmarcoux, I revisited the tests, and the calls to `skip_on_mobile` in `users/admin_configuration_spec.rb` were not needed, so I removed them. I also added back the `users/profile_spec.rb` in beta. Please review it again.

dmarcoux:
@eduardoj What is the reason for putting `users/profile_spec.rb` back in the beta folder? Looking at the spec, it's not testing anything related to a feature flag. It's only testing the elements to join/leave the beta program.

eduardoj:
@dmarcoux:

> @eduardoj What is the reason for putting `users/profile_spec.rb` back in the beta folder? Looking at the spec, it's not testing anything related to a feature flag. It's only testing the elements to join/leave the beta program.

I added it because it tests elements inside the user profile page, so it is related to the "user_profile_redesign" feature flag. Currently we are not testing the "user_profile_redesign" flag in beta, but I wanted to prevent the back and forth of removing the file now and adding it again in a following pull request that adds the user_profile_redesign" flag to beta.

dmarcoux:
I'm kind of divided on this. You are right, we need to test this with and without the feature flag `user_profile_redesign`. I would prefer to not copy the file if possible.

dmarcoux (approved):

PR #10901, opened 2021-03-17
Run all feature tests in CircleCI

eduardoj:
Before, only tests under one subdirectory level were tested.

Fix the file pattern used to run the tests, and include all tests, independent of the subdirectory level.

-----------------

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -274,7 +274,7 @@ jobs:
             fi
 
             echo $message
-            bundle exec rspec --format progress --format RspecJunitFormatter -o /home/frontend/feature/feature.xml spec/features/**/*_spec.rb
+            bundle exec rspec --format progress --format RspecJunitFormatter -o /home/frontend/feature/feature.xml spec/features/
 
             mkdir coverage_results
             cp -R coverage/.resultset.json coverage_results/resultset-feature-${CIRCLE_NODE_INDEX}.json


dmarcoux (approved):
Oh wow, nice catch. Before this change, we were running 159 examples. Now, it's 201 examples. 42 examples were simply ignored... ouch!

coolo:
and these 42 examples didn't cover a single line more - according to codecov. Something doesn't fit :)

dmarcoux:
It's all fine according to coveralls. You can compare https://coveralls.io/builds/37969495 (when #10880 was merged and the coverage decreased by 0.5%)  and https://coveralls.io/builds/38011866 (when we merged this PR and the coverage increased by 0.5%). I don't know for others, but I do sometimes overlook the coverage check. I need to pay more attention to this.

coolo:
0.5% is basically the measuring error of the coverage. But enabling these 42 examples didn't move the coverage outside that.

These are the files missed so far:
```
spec/features/beta/webui/notifications_spec.rb
spec/features/beta/webui/users/notifications_controller_spec.rb
spec/features/beta/webui/users/profile_spec.rb
spec/features/beta/webui/users/user_contributions_spec.rb
spec/features/webui/kiwi/images_spec.rb
spec/features/webui/projects/status_spec.rb
spec/features/webui/requests/submissions_spec.rb
spec/features/webui/users/admin_configuration_spec.rb
spec/features/webui/users/profile_spec.rb
spec/features/webui/users/user_admin_edit_spec.rb
spec/features/webui/users/user_contributions_spec.rb
spec/features/webui/users/user_home_page_spec.rb
```

You would expect the user feature tests to show somewhere in the controller that isn't covered without.

dmarcoux:
> 0.5% is basically the measuring error of the coverage. But enabling these 42 examples didn't move the coverage outside that.

*Basically* the measuring error? What do you mean?

> These are the files missed so far:
> (...)

How did you get to that list?
 
> You would expect the user feature tests to show somewhere in the controller that isn't covered without.

I'm sorry, I don't understand the sentence. *To show somewhere in the controller that isn't covered without....* ???

coolo:
Ah, my mistake. The normal up and down in coverage is 0.05%

PR #10903, opened 2021-03-17
Update data_migrate from 6.5.0 - 6.7.0

vpereira:
This PR is a follow-up from #10867. But we reduced the scope of the bundler solver to `bundle update --minor --strict $GEMNAME`, updating just what was strictly necessary.


dmarcoux (approved):

dmarcoux:
I just noticed in the CHANGELOG (this was already in `6.1.0`, so we're already far past that version):

>Rails 5.0 is no longer maintained. The gem will still work but it is not being actively tested.

Issue #10324, opened 2020-10-20
WebUI#UserController needs some love


vpereira:
There are some methods, specially regarding with involved project/packages filtering that should refactored either to Service Objects or Presenters. Those methods need to be tested once they're extracted.

dmarcoux:
![](https://github.trello.services/images/mini-trello-icon.png) [Address issues for the involved projects/packages feature](https://trello.com/c/BPrPcOcO/2061-address-issues-for-the-involved-projects-packages-feature)

vpereira:
look into the pagination and caching opportunities

related with that https://github.com/openSUSE/open-build-service/issues/10332

dmarcoux:
Look into the private methods of the controller and see how we can refactor this to not have everything in the controller, especially the code for filters. It should be easier to test.

The pagination was added in #10334.

krauselukas:
Closed by https://github.com/openSUSE/open-build-service/pull/10911

PR #10911, opened 2021-03-17
Move involved packages/projects logic to an service object 

krauselukas:


krauselukas:
[Issue #10324 UserController needs some love](https://trello.com/c/RgwMERw6/1630-issue-10324-usercontroller-needs-some-love)

src/api/app/presenters/involved_items_presenter.rb
```diff
@@ -0,0 +1,131 @@
...
+  private
+
+  def prepare_filters(filters:)
+    filters['search_text'] = filters['search_text']&.strip
+    filters
+  end
+
+  def filter_by_role?
+    return true if roles_to_filter.present?
+
```
vpereira, line 59: why are you letting so many empty lines between the last expression being evaluated in the methods (aka the return value)?

krauselukas: @vpereira that's what Rubocop wanted :) its because of this https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/EmptyLineAfterGuardClause

dmarcoux: I'm fine with having an empty line or not.

krauselukas:
As communicated with @vpereira, I'm going to move the stuff from the presenter to a service class

krauselukas:
The method codeclimate is complaining about will change most likely when the code gets adapted to use the changes that are getting introduced with https://github.com/openSUSE/open-build-service/pull/10918 lets ignore them for now

danidoni (approved):

src/api/spec/services/user_service/involved_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  before do
+    User.session = user
+  end
+
+  describe '#involved_packages_and_projects' do
+    context 'when there are no filters set' do
+      let(:filters) { { 'search_text' => '' } }
+
+      subject { involved_service.involved_pkg_and_prj_paginated }
```
eduardoj, line 40: This `subject` line, as it is the same in all the `context` clauses and all the only `describe` section, could be moved outside, just after the `before` clause in line 32. And the rest `subject` lines, could be removed.

krauselukas: done

krauselukas:
There is one little issue with the combination of the owner role filter and other roles, I'm going to fix it now and also add a test case for it. Thanks @eduardoj for pointing to this

krauselukas:
> There is one little issue with the combination of the owner role filter and other roles, I'm going to fix it now and also add a test case for it. Thanks @eduardoj for pointing to this

Fixed

danidoni:
LGTM

PR #10916, opened 2021-03-18
Move the user profile edit action close to the user details

danidoni:
**Before**
image

**After**
image

-----------------

### DIFF
--- a/src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
+++ b/src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
@@ -1,4 +1,14 @@
-.basic-info{ class: feature_enabled?(:user_profile_redesign) ? 'in-place-editing' : '' }
+.editing-form.d-none
+  = render partial: 'webui/user/user_profile_redesign/edit_account_form',
+           locals: { displayed_user: user, role_titles: role_titles, account_edit_link: account_edit_link }
+
+.basic-info
+  - if configuration.accounts_editable?(user)
+    .d-flex.flex-row-reverse
+      = link_to('javascript:void(0);', id: 'toggle-in-place-editing', class: 'nav-link', remote: true, title: 'Edit Your Account') do
+        %i.fas.mr-2.fa-user-edit
+        %span.nav-item-name Edit
+
   .row.mb-3.mb-md-0
     .col-4.col-sm-2.col-md-12.text-center
       = image_tag_for(user, size: 200)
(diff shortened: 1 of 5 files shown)


src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
```diff
@@ -18,3 +28,15 @@
...
+
+:javascript
+  $('#toggle-in-place-editing').on('click', function () {
+    $('.in-place-editing .basic-info').toggleClass('d-none');
+    $('.in-place-editing .editing-form').toggleClass('d-none');
+  });
+  $('#cancel-in-place-editing').on('click', function () {
+    $('.in-place-editing .basic-info').toggleClass('d-none');
+    $('.in-place-editing .editing-form').toggleClass('d-none');
+    return false;
```
dmarcoux, line 41:
```suggestion
```

dmarcoux:
[Move "edit" actions like Edit Project closer to the area affected](https://trello.com/c/zDnCOw4M/1624-move-edit-actions-like-edit-project-closer-to-the-area-affected)

src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
danidoni: This is needed to prevent the wrapping form to be sent and update the changes.

dmarcoux: Oh right, because `Cancel` is a button.

dmarcoux (approved):

saraycp (changes requested):
A solid button with an icon inside is something that we don't use in our interface. Moreover, there isn't an equivalent alternative in out patter library:
https://obs-patterns.netlify.app/atoms/buttons.html

I'd go for something similar to what you introduced in projects and packages.
What about this?
Screenshot_2021-03-18 Profile of scp(3)

danidoni:
> A solid button with an icon inside is something that we don't use in our interface. Moreover, there isn't an equivalent alternative in our patter library: https://obs-patterns.netlify.app/atoms/buttons.html
> 
> I'd go for something similar to what you introduced in projects and packages.
> What about this?
> Screenshot_2021-03-18 Profile of scp(3)

Talked offline and decided make the edit button more consistent to existing edit buttons

src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
```diff
@@ -1,4 +1,14 @@
-.basic-info{ class: feature_enabled?(:user_profile_redesign) ? 'in-place-editing' : '' }
+.editing-form.d-none
+  = render partial: 'webui/user/user_profile_redesign/edit_account_form',
+           locals: { displayed_user: user, role_titles: role_titles, account_edit_link: account_edit_link }
+
+.basic-info
+  - if configuration.accounts_editable?(user)
+    .d-flex.flex-row-reverse
+      = link_to('javascript:void(0);', id: 'toggle-in-place-editing', class: 'nav-link', remote: true, title: 'Edit Your Account') do
+        %i.fas.mr-2.fa-user-edit
```
saraycp, line 9:
```suggestion
        %i.fas.fa-user-edit
```

saraycp: Then, it would look like this
Screenshot_2021-03-18 Profile of Admin
:

saraycp (approved):

PR #10924, opened 2021-03-18
Fix airbrake ignore rules

hennevogel:
A loop about enumerable returns the enumerable which is always something,
so true.


vpereira (approved):

dmarcoux:
There's a typo in the commit message: airb**r**ake

PR #10926, opened 2021-03-18
Adjust margin and padding on edit links

saraycp:
The links are now left-aligned with the rest of the elements, and there is not excessive space at the bottom nor between the icon and the text.

Follow-up of: #10905 

**Result:**

Screenshot_2021-03-18 paquete

Screenshot_2021-03-18 Profile of Admin

-----------------

### DIFF
--- a/src/api/app/views/webui/project/_basic_info.html.haml
+++ b/src/api/app/views/webui/project/_basic_info.html.haml
@@ -14,6 +14,6 @@
 
   - if policy(project).update?
     .pt-4
-      = link_to(edit_project_path(project), class: 'nav-link', remote: true, title: 'Edit') do
+      = link_to(edit_project_path(project), class: 'nav-link pl-0', remote: true, title: 'Edit') do
         %i.fas.fa-edit.text-secondary
         %span.nav-item-name Edit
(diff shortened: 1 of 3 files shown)


src/api/app/views/webui/project/_basic_info.html.haml
```diff
@@ -14,6 +14,6 @@
 
   - if policy(project).update?
     .pt-4
```
dmarcoux, line 16: Same for package which doesn't have any top padding.

```suggestion
    .pt-2
```

saraycp: I don't think it's a good idea because, when the description is collapsed, the `Edit` and the `Show more` links will be displayed more or less overlapped (occupying the same area).

Show_home_scp_paquete_Open_Build_Service(1)

So I will keep it with `pt-4` for both project and packages.

dmarcoux: Good point :slightly_smiling_face:

At the time of my comment, the Edit link for package didn't have any padding. it's fixed now :+1:

dmarcoux (approved):

eduardoj (approved):

PR #10928, opened 2021-03-18
Fix the permission checks to show the edit user profile form

danidoni:
I forgot the permission check to show the user profile form :facepalm:

Thanks to @dmarcoux for reporting it.

-----------------

### DIFF
--- a/src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
+++ b/src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
@@ -3,7 +3,7 @@
            locals: { displayed_user: user, role_titles: role_titles, account_edit_link: account_edit_link }
 
 .basic-info
-  - if configuration.accounts_editable?(user)
+  - if configuration.accounts_editable?(user) && policy(user).update?
     .d-flex.flex-row-reverse
       = link_to('javascript:void(0);', id: 'toggle-in-place-editing', class: 'nav-link', remote: true, title: 'Edit Your Account') do
         %i.fas.fa-user-edit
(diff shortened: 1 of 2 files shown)


danidoni:
[Move "edit" actions like Edit Project closer to the area affected](https://trello.com/c/zDnCOw4M/1624-move-edit-actions-like-edit-project-closer-to-the-area-affected)

src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
```diff
@@ -3,7 +3,7 @@
            locals: { displayed_user: user, role_titles: role_titles, account_edit_link: account_edit_link }
 
 .basic-info
-  - if configuration.accounts_editable?(user)
+  - if configuration.accounts_editable?(user) && policy(user).update?
```
eduardoj, line 6: Shouldn't also this check also be added [here](https://github.com/openSUSE/open-build-service/blob/dc04932fa9e9f515c74e5b22d2aecf04a1399264/src/api/app/views/webui/user/_index_actions.html.haml#L2), then?

saraycp (approved):

src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
saraycp: True! I've just checked that, without the feature flag enabled, I can edit my own profile but another user's one too.

saraycp (changes requested):

src/api/app/views/webui/user/user_profile_redesign/_basic_info.html.haml
saraycp: @danidoni can you please add the same condition to [here](https://github.com/openSUSE/open-build-service/blob/dc04932fa9e9f515c74e5b22d2aecf04a1399264/src/api/app/views/webui/user/_index_actions.html.haml#L2) ?

danidoni: Yes

saraycp (approved):

PR #10933, opened 2021-03-19
Make clear that the user is on their own profile page

saraycp:
Sometimes it's not clear if the user is on their own profile page or in someone else's one. This little change can give a clue.

-----------------

### DIFF
--- a/src/api/app/views/webui/users/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui/users/_breadcrumb_items.html.haml
@@ -10,7 +10,10 @@
       = @pagetitle.capitalize
   - when 'show'
     %li.breadcrumb-item.active{ 'aria-current' => 'page' }
-      Profile of #{@displayed_user}
+      - if feature_enabled?('user_profile_redesign') && @is_displayed_user
+        Your Profile
+      - else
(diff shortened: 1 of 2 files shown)


src/api/app/views/webui/users/_breadcrumb_items.html.haml
```diff
@@ -10,7 +10,10 @@
       = @pagetitle.capitalize
   - when 'show'
     %li.breadcrumb-item.active{ 'aria-current' => 'page' }
-      Profile of #{@displayed_user}
+      - if feature_enabled?('user_profile_redesign') && @is_displayed_user
```
krauselukas, line 13: @saraycp IMO this doesn't need to be part of the beta program

saraycp: Well, maybe. Anyway, the `user_profile_redesign` is going to be rolled out in the upcoming days, I think it’s not worth changing it right now, in a few days will be out.

krauselukas: Okay fine

krauselukas (approved):

PR #10918, opened 2021-03-18
Sort involved projects/packages alphabetically

dmarcoux:
Part of #10831

Issue #10831, opened 2021-03-04
User Profile Redesign - Feedback


dmarcoux:
I completely forgot to create an issue regarding this feedback we got from @krop on the mailing list.

> Quite bad on desktop. Instead of a compact list, now I see:
> 
> - the package names and description instead of the repository
> - packages from long dead versions for which I was bug owner (some openSUSE:
> 12.x ones)
> 
> That's a lot of scrolling and lot of useless information.

dmarcoux:
![](https://github.trello.services/images/mini-trello-icon.png) [Address issues for the involved projects/packages feature](https://trello.com/c/BPrPcOcO/2061-address-issues-for-the-involved-projects-packages-feature)

dmarcoux:
Regarding `the package names and description instead of the repository`, as we can see in both GIFs above, the repositories are never shown for the packages/projects. So adding repositories is definitely possible, but what would be the benefit?

As for `packages from long dead versions for which I was bug owner (some openSUSE: 12.x ones)`, it's pretty unclear right now how the packages/projects are sorted. In fact, they aren't sorted at all. This can result in "dead" packages/projects being displayed at the top of the list. We should offer various sorting options for packages/projects. How about sorting alphabetically and by most recent changes in packages/projects (We need to check if we can achieve this).

About `That's a lot of scrolling and lot of useless information.`, it's true that there is more scrolling right now. We added role(s) of the involved user and the package/project description (when available). The description is for sure taking way too much space, especially since long descriptions are simply displayed as such.

eduardoj:
In the same direction, @coogor also provided useful feedback in this issue: https://github.com/openSUSE/open-build-service/issues/10323

krop:
> As for `packages from long dead versions for which I was bug owner (some openSUSE: 12.x ones)`, it's pretty unclear right now how the packages/projects are sorted. In fact, they aren't sorted at all. This can result in "dead" packages/projects being displayed at the top of the list. We should offer various sorting options for packages/projects, like sorting alphabetically and by most recent changes in packages/projects (We need to check if we can achieve this).
> 

Screenshot attached. For whatever reason, I used to be bugowner for a couple 12.3 packages which are now listed. They're not visible with beta off

![Screenshot_20210304_180302](https://user-images.githubusercontent.com/1181826/110000929-bfbed900-7d0b-11eb-80a0-893b3df50fcf.png)

hennevogel:
[...]

PR #10937, opened 2021-03-19
Sort involved projects/packages alphabetically

eduardoj:

Superseds: #10918

Part of #10831

-----------------

### DIFF
--- a/src/api/app/services/user_service/involved.rb
+++ b/src/api/app/services/user_service/involved.rb
@@ -15,6 +15,16 @@ module UserService
       Kaminari.paginate_array(involved_packages_and_projects).page(@page).per(MAX_ITEMS_PER_PAGE)
     end
 
+    def sort_involved_items(involved_items)
+      involved_items.sort_by do |involved_item|
+        if involved_item.is_a?(Package)
+          "#{involved_item.project} / #{involved_item}".downcase
+        else
+          involved_item.name.downcase
+        end
+      end
+    end
+
     def involved_packages_and_projects
       involved_items = []
       roles = roles_to_filter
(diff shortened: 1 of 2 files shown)


src/api/app/services/user_service/involved.rb
```diff
@@ -28,7 +38,7 @@ def involved_packages_and_projects
         involved_items.concat(pkg_or_prj_unfiltered(user: @user, klass: Package)) if consider_involved_packages?
         involved_items.concat(pkg_or_prj_unfiltered(user: @user, klass: Project)) if consider_involved_projects?
       end
-      involved_items.uniq
+      sort_involved_items(involved_items.uniq)
```
krauselukas, line 41: @eduardoj you use `uniq` already inside the method

eduardoj: True. Let's remove it from the method, then.

eduardoj: Fixed. :+1:

src/api/app/services/user_service/involved.rb
```diff
@@ -15,6 +15,16 @@ def involved_pkg_and_prj_paginated
       Kaminari.paginate_array(involved_packages_and_projects).page(@page).per(MAX_ITEMS_PER_PAGE)
```
krauselukas, line 15: @eduardoj one suggestion, maybe call the `sort_involved_items` on the paginated elements, since sorting is memory hungry

krauselukas: ahh sorry doesnt work, forget about it....

krauselukas (approved):

PR #10949, opened 2021-03-29
fix spider test again

bugfinder:


dmarcoux:
I'm a bit lost here... The spider specs are passing on master. Why is this change needed?

bugfinder:
all these fixes along these lines are spurious fails, packages and projects already deleted and still found in the spider crawl probably because of some caching. It also depends on timing and number of threads/cpus you have during the test. this one reproducibly failed today in IBS SLE-15-SP3 builds

dmarcoux:
What is the error message? It seems like the package is not branched correctly.

hennevogel:
closing, we are way past SLE-15-SP3 and it still works on master...

PR #10958, opened 2021-03-30
Handle dots in user logins for group members page/route

### LINKED ISSUE #10955: Error when try remove group member with dot
# Issue Description
Error received when try to remove user with dot from group members. 

Request `DELETE /groups/GROUPNAME/user/u.name` return 404 Not Found.

Expected Result
============
Remove user with dot from group members.

How to Reproduce
===============
1. Go to 'https://os-build.devel.pro/group/show/GROUPNAME'
2. Add Group member with dot `u.name`
3. Click on 'Remove member from group' `(X)`

Further Information
===============
* OBS Server 2.10.10

dmarcoux:
Fixes #10955

Before, the `Delete group member` modal didn't appear for a user which has dots in their login.

The routes /groups/<group>/users/<user> were also not handling dots correctly, treating them as separators. Having a constraint for the user login solves this. This is documented in Rails, see https://guides.rubyonrails.org/routing.html#dynamic-segments

### DIFF
--- a/src/api/app/views/webui/groups/_group_members.html.haml
+++ b/src/api/app/views/webui/groups/_group_members.html.haml
@@ -10,7 +10,8 @@
         .card.m-1.p-2.group-user
           - if write_access
             = render(partial: 'webui/groups/group_members_delete_dialog', locals: { user: user, group: group })
-            = link_to('#', data: { toggle: 'modal', target: "#delete-group-members-modal-#{user}" }, title: 'Remove member from group') do
+            = link_to('#', data: { toggle: 'modal', target: "#delete-group-members-modal-#{dom_id(user)}" },
+                      title: 'Remove member from group') do
               .float-right
                 %i.fas.fa-sm.fa-times-circle.text-danger
           .row.no-gutters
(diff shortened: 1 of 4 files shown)


dmarcoux:
This raises (again) questions regarding characters we should allow in user logins. I fear that this issue is not the only one of its kind...

src/api/app/views/webui/groups/_group_members.html.haml
```diff
@@ -10,7 +10,8 @@
         .card.m-1.p-2.group-user
           - if write_access
             = render(partial: 'webui/groups/group_members_delete_dialog', locals: { user: user, group: group })
-            = link_to('#', data: { toggle: 'modal', target: "#delete-group-members-modal-#{user}" }, title: 'Remove member from group') do
+            = link_to('#', data: { toggle: 'modal', target: "#delete-group-members-modal-#{user.to_s.gsub('.', '\\\.')}" },
```
hennevogel, line 13: There is [`dom_id`](https://api.rubyonrails.org/classes/ActionView/RecordIdentifier.html#method-i-dom_id) for that purpose. We should not use the login for referral.

hennevogel (changes requested):

src/api/app/views/webui/groups/_group_members.html.haml
dmarcoux: True, I didn't think about that.

dmarcoux:

danidoni (approved):

hennevogel (approved):

PR #10960, opened 2021-03-31
Add file /public/apidocs-new.html to fix requests to /apidocs-new

### LINKED ISSUE #10631: 404 when accessing the API Swagger docs without a trailing slash
# Issue Description

Accessing https://build.opensuse.org/apidocs-new/ is showing the API Swagger docs, while https://build.opensuse.org/apidocs-new is a 404.

Expected Result
============
https://build.opensuse.org/apidocs-new shows the API Swagger docs.

How to Reproduce
===============
1. Go to https://build.opensuse.org/apidocs-new
2. See 404

Further Information
===============
We'll have to change the Apache configuration for the `apidocs-new` directory. Documentation for the [DirectorySlash directive](https://httpd.apache.org/docs/2.4/mod/mod_dir.html#directoryslash) in Apache. It's enabled by default, but it might be disabled in our Apache configuration.

vpereira:
Requests to /public/apidocs-new should be redirected to
/public/apidocs-new/. This fixes just in production, where we have the Apache webserver serving static content. It doesn't work in development mode.

Fixes #10631

-----------------

### DIFF
--- a/src/api/public/apidocs-new.html
+++ b/src/api/public/apidocs-new.html
@@ -1,0 +1,9 @@
+<!-- apidocs-new.html -->
+<html>
+        <head>
+        <meta http-equiv="refresh" content="0;url=/apidocs-new/" />
+        </head>
+        <body>
+        </body>
+</html>
+


vpereira:
regarding the review app: it Works but because the path on review-app is different, it redirects to `/apidocs-new` on the review-app which doesnt exist there.

dmarcoux:
It's already something if this works in production. It's a shame that it doesn't work in development... but I don't think it's worth investing more time to fix this for development.

dmarcoux (approved):

PR #10962, opened 2021-03-31
Requests improvements

eduardoj:
Follow-up improvements to #10961.

-----------------

### DIFF
--- a/src/api/app/controllers/webui/project_controller.rb
+++ b/src/api/app/controllers/webui/project_controller.rb
@@ -164,7 +164,6 @@ class Webui::ProjectController < Webui::WebuiController
   end
 
   def requests
-    @requests = @project.open_requests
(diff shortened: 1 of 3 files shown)


src/api/app/controllers/webui/project_controller.rb
```diff
@@ -164,7 +164,6 @@ def destroy
   end
 
   def requests
-    @requests = @project.open_requests
```
dmarcoux, line 167: It's weird that this instance variable isn't used, but looking at the views, it's really not needed.

dmarcoux (approved):

src/api/app/controllers/webui/project_controller.rb
eduardoj: Yes. It's probably some remaining code that was not deleted after changing the view to work with AJAX requests.

PR #10969, opened 2021-04-06
Document 'published/' get method

saraycp:
This `published/` endpoint ends up in the `SourceController#index` not in `PublishedController#index` as can be supposed.

It returns a list of the (published) projects.

~**NOTE:** The `published/`endpoint is going to list projects, repositories, architectures or binaries depending on the passed parameters. I can't find an appropriate Tag name for this group: `Published Elements`? `Published Sources` ? ... That's why I left the same name used in the old documentation: `Published Binary Package Tree`.~

**Edit:**
The group or tag name is finally going to be `Published Binaries`.

-----------------

### DIFF
--- a/src/api/public/apidocs-new/OBS-v2.10.50.yaml
+++ b/src/api/public/apidocs-new/OBS-v2.10.50.yaml
@@ -8,6 +8,7 @@ tags:
   - name: Groups
   - name: Issue Trackers
   - name: Person
+  - name: Published Binaries
   - name: Attribute Namespaces
   - name: Attributes
   - name: Request
@@ -91,6 +92,9 @@ paths:
     $ref: 'paths/person_login_token.yaml'
   /person/{login}/token/{id}:
     $ref: 'paths/person_login_token_id.yaml'
+
+  /published:
+    $ref: 'paths/published.yaml'
 
   /request:
     $ref: 'paths/request.yaml'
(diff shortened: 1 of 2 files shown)


src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -8,6 +8,7 @@ tags:
   - name: Groups
   - name: Issue Trackers
   - name: Person
+  - name: Published Binaries
```
vpereira, line 11: Binaries are not technically always correct. Historically we use it, however today a better term exists: Artifacts or Job Artifacs. Artifacts refer to the output of your build process that you would deploy/install. This could be a repo file, tar or binary. You then deploy these artifacts onto your system. Example of non binary artifact: `osc api '/published/zypp:SLE-15-Branch/SLE_15_Update/zypp:SLE-15-Branch.repo'`

saraycp: I'm with you, `binaries` is not the best name for that heterogeneous output. However, I believe that by using `Artifacts` in the new API documentation, the users can struggle when trying to find the endpoint or understanding what we are referring to. Moreover, they won't be able to match `Artifacts` with any element in the UI, where we still use `Binaries`.

I really like the idea, so I propose to track it for future refactoring or another version of the API.

krauselukas (approved):

danidoni (approved):

PR #10975, opened 2021-04-06
Published Binaries API documentation

saraycp:
Different endpoints to get to the binaries list (projects -> repositories -> architectures -> binaries).

-----------------

### DIFF
--- a/src/api/public/apidocs-new/paths/published_project_name_repository_name.yaml
+++ b/src/api/public/apidocs-new/paths/published_project_name_repository_name.yaml
@@ -1,0 +1,36 @@
+get:
+  summary: List the content of the directory tree where the binaries are published at the level project/repository.
+  description: >
+    Get a list of architectures' directories and other files (.repo, .ymp, etc.) present at the level
+    project/repository of the directory tree where the published binaries are stored.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - $ref: '../components/parameters/repository_name.yaml'
+  responses:
+    '200':
+      description: OK. The request has succeeded.
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/directory.yaml'
+          example:
+            entry:
+              - name: 'OBS:Server:Unstable.repo'
+              - name: 'OBS_Server.ymp'
+              - name: 'aarch64'
+              - name: 'x86_64'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      description: Not Found.
+      content:
+        application/xml; charset=utf-8:
+          schema:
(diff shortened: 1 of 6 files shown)


src/api/public/apidocs-new/paths/published_project_name_repository_name.yaml
```diff
@@ -0,0 +1,35 @@
+get:
+  summary: List the architectures of the pair project/repository with published binaries.
+  description: >
+    Get a list of the architectures of the given pair project and repository that already have published binaries.
+  security:
+    - basic_authentication: []
+  parameters:
```
krauselukas, line 7: @saraycp there is also the `view=status` parameter for this endpoint

src/api/public/apidocs-new/paths/published_project_name_repository_name.yaml
```diff
@@ -0,0 +1,35 @@
+get:
+  summary: List the architectures of the pair project/repository with published binaries.
+  description: >
+    Get a list of the architectures of the given pair project and repository that already have published binaries.
```
krauselukas, line 4: @saraycp I think that not a 100% correct, to me it looks more like the directory tree of the download repository/place where the binaries get published. Here an example https://download.opensuse.org/repositories/OBS:/Server:/Unstable/SLE_12_SP5/

krauselukas: Its basically what you described here, but one level up https://github.com/openSUSE/open-build-service/pull/10975/files#diff-941da4259ba92323e54fa18add5045aeed9001f3e26c9e7b4040e9c335f25331R103

src/api/public/apidocs-new/paths/published_project_name_repository_name.yaml
saraycp: I'm going to add that endpoint in another PR

krauselukas: ahh right, returns a different schema, fine then

src/api/public/apidocs-new/paths/published_project_name_repository_name.yaml
saraycp: You are completely right. I'm going to give a better description. One more time we are in a case where the name is not completely appropriate. 

The output of `/published/{project_name}/{repository_name}` can be something like:

```
<directory>
  <entry name="repocache"/>
  <entry name="repodata"/>
  <entry name="src"/>
  <entry name="x86_64"/>
</directory>
```
Then we can "inspect" one of them with the following endpoint:

`/published/{project_name}/{repository_name}/{architecture_name}` where `{architecture_name}` can be replaced by  `src` or `repocache` that are not "architectures".

krauselukas (approved):

PR #10983, opened 2021-04-07
Published binary, status and ymp endpoints

saraycp:
Document the rest of the `published/*` endpoints to:

- get the status of the building process
- download the binary
- generate a ymp pattern

Add a missing 403 error response that can happen on every `published/*` endpoint when the binary download is blocked.

 

-----------------

### DIFF
--- a/src/api/public/apidocs-new/paths/published_project_name_repository_name_architecture_name_binary_filename.yaml
+++ b/src/api/public/apidocs-new/paths/published_project_name_repository_name_architecture_name_binary_filename.yaml
@@ -1,0 +1,48 @@
+get:
+  summary: Return the binary file itself.
+  description: >
+    Allow to download the binary file that was published and stored under the directory
+    given by project/repository/architecture/.
+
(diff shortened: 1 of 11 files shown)


src/api/public/apidocs-new/paths/published_project_name_repository_name_architecture_name_binary_filename.yaml
```diff
@@ -0,0 +1,50 @@
...
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - $ref: '../components/parameters/repository_name.yaml'
+    - $ref: '../components/parameters/architecture_name.yaml'
+    - $ref: '../components/parameters/binary_filename.yaml'
+  responses:
+    '200':
+      description: OK
+      content:
+        application/*: # Media Types like application/x-rpm, application/octet-stream
```
danidoni, line 21: Uh, nice trick! :+1:

vpereira (approved):

PR #10986, opened 2021-04-08
Document create submit request workflow with OpenAPI

eduardoj:
Provide an example of using request API endpoints by documenting the create submit request workflow.

-----------------

### DIFF
--- a/src/api/public/apidocs-new/OBS-v2.10.50.yaml
+++ b/src/api/public/apidocs-new/OBS-v2.10.50.yaml
@@ -12,6 +12,37 @@ tags:
   - name: Attribute Namespaces
   - name: Attributes
   - name: Request
+    description: |
+      Please, before using request endpoints, make sure you have read and understood the "Request and Review System" section of
+      the [user guide](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.request_and_review_system.html).
+
+      In the next sections you will see how calls to request endpoints could be combined to fullfill different workflows.
+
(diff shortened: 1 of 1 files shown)


src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -12,6 +12,37 @@ tags:
...
+    description: |
+      Please, before using request endpoints, make sure you have read and understood the "Request and Review System" section of
+      the [user guide](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.request_and_review_system.html).
+
+      In the next sections you will see how calls to request endpoints could be combined to fullfill different workflows.
+
+      ## Workflow 1: Create a submit request from package A to package B
+
+      The steps described in this workflow could be easily replicated with this command:
+      `osc createrequest -a submit home:Admin:Origin A home:Admin:Destiny B -m "Submit A to B"`
```
danidoni, line 24:
```suggestion
      `osc createrequest -a submit home:Admin:Origin A home:Admin:Destination B -m "Submit A to B"`
```

danidoni (approved):

src/api/public/apidocs-new/OBS-v2.10.50.yaml
danidoni: :thought_balloon:

eduardoj: Ops. Changed. Thanks!

PR #10990, opened 2021-04-12
Extract backend request proxying to a concern

hennevogel:
Isolate this code, rename some functions and add some documentation to
make it more obvious.

### DIFF
--- a/src/api/app/controllers/application_controller.rb
+++ b/src/api/app/controllers/application_controller.rb
@@ -12,6 +12,7 @@ class ApplicationController < ActionController::Base
   include FlipperFeature
 
   include RescueHandler
+  include BackendProxy
 
   # session :disabled => true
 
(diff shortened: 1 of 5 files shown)


hennevogel:
Still thinking about the right way to test this...

src/api/app/lib/backend/proxy_request.rb
```diff
@@ -0,0 +1,136 @@
...
+      if @request.form_data?
+        # it's uncommon, but possible that we have both
+        query_string += '&' if query_string.present?
+        query_string += @request.raw_post
+      end
+      query_string = '?' + query_string if query_string.present?
+      path + query_string
+    end
+
+    def use_web_server_forward
```
vpereira, line 36: do we have to support different servers here?
And if yes, cannot we extract it to a lib/gem?

src/api/app/lib/backend/proxy_request.rb
```diff
@@ -0,0 +1,136 @@
...
+        return true
+      end
+
+      false
+    end
+
+    # FIXME: This shares a lot of logic with Backend::File, should be merged.
+    def proxy_from_backend
+      logger.debug "[backend] VOLLEY: #{@path}"
+      Backend::Test.start
```
vpereira, line 74: `Backend::Test.start` as well in production?

vpereira: this method is doing too many things.. couldn't we break it in smaller methods?

src/api/app/lib/backend/proxy_request.rb
```diff
@@ -0,0 +1,136 @@
...
+      when :get, :head
+        proxy_from_backend unless use_web_server_forward
+      when :post, :put, :delete
+        proxy_to_backend
+      end
+    end
+
+    private
+
+    def path_from_request
```
vpereira, line 24: could we have it tested? There isn't already gems to do this? I think even the standard lib can do something like that, right? https://docs.ruby-lang.org/en/2.7.0/URI/Generic.html#method-c-build

src/api/app/lib/backend/proxy_request.rb
```diff
@@ -0,0 +1,136 @@
...
+          volleyfile.write(segment)
+        end
+      end
+      opts[:length] = volleyfile.length
+      opts[:disposition] = 'inline' if ['text/plain', 'text/xml'].include?(opts[:type])
+      # streaming makes it very hard for test cases to verify output
+      opts[:stream] = false if Rails.env.test?
+      send_file(volleyfile.path, opts)
+      # close the file so it's not staying in the file system
+      volleyfile.close
```
vpereira, line 99: it will just be deleted, if the `unlink` parameter is set to true https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html#method-i-close otherwise the method `cache_request_body_to_file` wouldn't work at all.

src/api/app/lib/backend/proxy_request.rb
hennevogel: I try to limit the refactoring to clarification here.

src/api/app/lib/backend/proxy_request.rb
hennevogel: > do we have to support different servers here?

Yes.

> And if yes, cannot we extract it to a lib/gem?

Not really, it depends on our configuration.

src/api/app/lib/backend/proxy_request.rb
hennevogel: > Backend::Test.start as well in production?

First line i `Backend::Test.start` is `return unless Rails.env.test?`

src/api/app/lib/backend/proxy_request.rb
hennevogel: All of this needs to be tested. Just not sure how yet as it depends on `ActionDispatch::Request` but not really on a controller test...

src/api/app/lib/backend/proxy_request.rb
hennevogel: I try to limit the refactoring to clarification here.

vpereira: but the comment is misleading (aka wrong)

src/api/app/lib/backend/proxy_request.rb
vpereira: I would prefer to have the `if` inside of the method here to make it more explicit?

src/api/app/lib/backend/proxy_request.rb
vpereira: so could you at least extract to some lib? This code is irrelevant here.

hennevogel: How is it irrelevant? It's part of proxying. Either we proxy via file or we proxy via the plugins.

src/api/app/lib/backend/proxy_request.rb
hennevogel: I really try to limit the refactoring to clarification here. Otherwise this will be an endless rabbit hole. As you can see from the spec, this already leaked far into the app.

src/api/app/lib/backend/proxy_request.rb
hennevogel: Lots of things in here are wrong and misleading. I don't want to deal with those things now. In this PR I want to isolate the code.

hennevogel:
Alright so this is still controller code now. I will test this with a controller spec...

hennevogel:
I tried for a while now but I have no real idea how to test this. There are several specs that test this implicitly (like `SourceProjectConfigController#show`). 

I tried hooking into `Tempfile` as this is basically what `pass_to_backend` does, storing `request.body` or backend response to tmp files but `Tempfile` is a very busy class. I could make this a unit test of `BackendProxy` and stub everything but I'm not sure this would have any value...

Any other ideas?

PR #10991, opened 2021-04-12
Add build jobstatus and reason to swagger api docs

krauselukas:


danidoni (approved):

krauselukas:
[Build](https://trello.com/c/gsG0OEvj/1635-build)

src/api/public/apidocs-new/components/schemas/reason.yaml
```diff
@@ -0,0 +1,13 @@
...
+    type: string
+    example: 'source change'
+  time:
+    type: string
+    example: '1617982407'
+  oldsource:
+    type: string
+    example: 'da9ebc6a12b6da45f9ead3fc2f924889'
+xml:
+  name: reason
```
eduardoj, line 13: Nitpick: regarding the name of the file `reason.yaml`, I find it too generic. Even it describes an XML that contains a `<reason>` element, listing it together with all the other schemas makes it not so obvious that it refers to a build reason. Maybe is it better to change the name of the file from `reason.yaml` to `buildreason.yml`, as states in the original apidocs?

src/api/public/apidocs-new/paths/build_project_repository_arch_package_jobstatus.yaml
```diff
@@ -0,0 +1,25 @@
+get:
+  summary: Show the build status of a currently running build job.
+  description: >
+    Show the build status of a currently running build job. Shows an empty
+    result if no build job is running.
```
eduardoj, line 5: A paragraph referring to the schema used for output XML validation (`jobstatus.rng`) should be added. This is already documented in the [apidocs](https://build.opensuse.org/apidocs/index.html#113).

src/api/public/apidocs-new/components/schemas/reason.yaml
saraycp: I had the same feeling, yes. But I would have suggested `build_reason.yaml`. However, as we are still using terms like `jobstatus` I'm fine to continue using non-separated words like `buildreason`.

krauselukas: Done @eduardoj

src/api/public/apidocs-new/paths/build_project_repository_arch_package_jobstatus.yaml
krauselukas: Done @eduardoj

eduardoj (approved):

saraycp (approved):

PR #10992, opened 2021-04-12
[api] Add revision enforcement

adrianschroeter:


src/api/app/models/bs_request_action.rb
```diff
@@ -786,6 +786,18 @@ def check_for_expand_errors!(add_revision)
       query[:expand] = 1 unless updatelink
       query[:rev] = source_rev if source_rev
       dir = Xmlhash.parse(Backend::Api::Sources::Package.files(source_project, source_package, query))
+      unless add_revision
+        # Enforce revisions?
+        tprj = Project.get_by_name(target_project)
+        add_revision = tprj.instance_of?(Project && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
```
dmarcoux, line 792: This is why RuboCop is barking:

```suggestion
        add_revision = tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
```

src/api/app/models/bs_request_action.rb
```diff
@@ -786,6 +786,18 @@ def check_for_expand_errors!(add_revision)
       query[:expand] = 1 unless updatelink
       query[:rev] = source_rev if source_rev
       dir = Xmlhash.parse(Backend::Api::Sources::Package.files(source_project, source_package, query))
+      unless add_revision
+        # Enforce revisions?
+        tprj = Project.get_by_name(target_project)
+        add_revision = tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
+        if add_revision && source_rev
+          # Enforce expanded revision of given revision by user
+          if dir['entry'].is_a?(Array) && dir['entry'].map { |e| e['name'] }.include?('_link') ||
```
coolo, line 795: you don't need to have this if. You can just loop over dir.elements('entry') - it will give you an array in both cases.

adrianschroeter: A loop is slower than a map or hash lookup.

but I do not care in the end, feel free to change it.

src/api/app/models/bs_request_action.rb
adrianschroeter: fixed that a while ago. And another rubocop complain now.

src/api/app/models/bs_request_action.rb
coolo: how many entries to you expect there to be? a thousand or rather a million? I can run a benchmark

PR #11001, opened 2021-04-13
Remove duplication of what pass_to_backend does

hennevogel:
Handing proxying to the webserver or sending the file is
exactly what pass_to_backend does. No need to re-do this in
controller code.

Preparation for #10990

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -395,8 +395,8 @@ class SourceController < ApplicationController
   def lastevents
     path = get_request_path
 
-    # map to a GET, so we can X-forward it
-    volley_backend_path(path) unless forward_from_backend(path)
+    # force to a GET, so we can X-forward it
+    pass_to_backend(path, force_get: true)
   end
 
   # POST /source?cmd=createmaintenanceincident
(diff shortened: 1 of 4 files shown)


src/api/app/controllers/source_project_config_controller.rb
```diff
@@ -7,15 +7,7 @@ def show
...
-    return if forward_from_backend(config.full_path(sliced_params))
-
-    content = config.content(sliced_params)
-
-    unless content
-      render_404(config.errors.full_messages.to_sentence)
-      return
-    end
-    send_config(content, config.response[:type])
+    pass_to_backend(config.full_path(sliced_params))
```
hennevogel, line 10: There is one difference here. `content` was solely in memory, `pass_to_backend` stores it in a tmpfile. I don't think this makes any difference to this route.

src/api/app/controllers/source_controller.rb
```diff
@@ -395,8 +395,8 @@ def lastevents_public
   def lastevents
     path = get_request_path
```
krauselukas, line 396: @hennevogel you could even get rid of this line, because the `pass_to_backend` method defaults to `get_request_path` if the path you pass is nil. But not a blocker

krauselukas (approved):

hennevogel:
it's now included in #10990 please review there...

PR #11003, opened 2021-04-14

dmarcoux:
With a workflow in CircleCI to detect potential issues with the next Rails version.


.circleci/config.yml
```diff
@@ -42,6 +47,10 @@ aliases:
     name: Create database
     command: cd src/api; bundle exec rake db:setup RAILS_ENV=test
 
+  - &create_test_db-next_rails
+    name: Create database based on gems defined in src/api/Gemfile.next
```
hennevogel, line 51: Create database based on gems? huh?

src/api/Gemfile.next.lock
```diff
@@ -0,0 +1,559 @@
+GEM
```
hennevogel, line 1: Hmmm, how are we keeping this in sync and up to date?

hennevogel:
awesome

.circleci/config.yml
dmarcoux: To be honest... I don't know how to phrase this. I create the database and the next Rails version might changed how this is done. Let's say the rake tasks changed in the next Rails version, the command would have to be changed. This would probably not happen, but I still want to run with `next bundle exec rake (...)`.

src/api/Gemfile.next.lock
dmarcoux: I have to figure this out. It should be possible to run Depfu on Gemfile.next. This has to be set after the PR is merged.

.circleci/config.yml
hennevogel: `Create database based with next rails version`

.circleci/config.yml
```diff
@@ -292,6 +301,28 @@ jobs:
...
+      - <<: *frontend_base
+      - <<: *mariadb
+    steps:
+      - attach_workspace:
+         at: .
+      - run: *install_dependencies-next_rails
+      - run: *init_git_submodule
+      - run: *wait_for_database
+      - run: *create_test_db-next_rails
+      # TODO: Run all specs/tests with the next Rails version
```
hennevogel, line 315: Not sure you should run all. This is costly, time wise.

dmarcoux: Yes, it's true. On the other end, how are we sure that we have no errors due to the next Rails version? I was thinking about running this only on branches which have a `next_rails:` prefix like `next_rails:fix-deprecation-warnings-123` or `next_rails:something-something`.

hennevogel: Hm that sounds like a very good idea in combination with depfu.

dmarcoux: Filtering PRs based on their branch name isn't possible with CircleCI. I've looked at the environment variables like `CIRCLE_BRANCH` (it's always containing `pull/some_number` for PRs instead of the branch name given by the PR author). Filtering, with the `filters` key in the CircleCI configuration is only working for branches in the repository itself, not for PR since, just like `CIRCLE_BRANCH`, it's filtering PRs based on the value of `CIRCLE_BRANCH`. I think I'll give it a try with GItHub actions since it's possible to filter branches based on their branch name.

dmarcoux:
I'm closing this PR since I want to test the GitHub Actions in a PR and I need the branch name to match [the prefix defined in the GitHub Actions workflow config](https://github.com/openSUSE/open-build-service/pull/11003/files#diff-4f3874ad190ad71c4cf01d3719661e28f50a7656be5fbbe7b3d1459937ae1355R8). It's simpler to create a new PR from a branch matching that prefix.

PR #11008, opened 2021-04-19

dmarcoux:
With a workflow in GitHub Actions to detect potential issues with the next Rails version. It's not possible to use CircleCI since we can't filter PRs based on their branch name.

The module RailsVersion is to help with migrating code to a newer Rails version.


dmarcoux:
Closing yet another PR for next_rails. I need to have an upstream branch, so not in a fork, to be able to run the GitHub Action workflow.

PR #11015, opened 2021-04-20

dmarcoux:
With a job in CircleCI to detect potential issues with the next Rails version.

The module RailsVersion is to help with migrating code to a newer Rails version.


vpereira:
Looks like you forgot to install the package `cyrus-sasl-devel` on your GH action container

dmarcoux:
I'm hitting a lot of issues with the GitHub Actions setup.... running this in a container isn't working at all. So now I tried to run this on the host OS instead (Ubuntu 20.04) and then other issues happen. I'll try again to make it work on CircleCI and hopefully this will be enough this time.

dmarcoux:
It is expected that the new CircleCI job `next_rails` is failing since it means that we have to adapt the code to work on the Rails 6.1.x. However, the other CircleCI jobs should keep passing. This means that any code change with `RailsVersion.is_6_1?` are not causing issues in the current Rails version 6.0.x.

I will document this in our developer documentation in the wiki.

hennevogel:
What I don't like too much is that this job will take another circle-ci slot for probably close to 40 minutes...

dmarcoux:
I tried to make it work in GitHub Actions to work around that, but it's just not cutting it since I hit a lot of issues with the containers and without containers... it didn't work either due to some installation issues for gems.

hennevogel:
> I tried to make it work in GitHub Actions 

Sure I saw that, not criticizing your work. But we have to weigh slots vs. this 'feature'.

dmarcoux:
> Sure I saw that, not criticizing your work. But we have to weigh slots vs. this 'feature'.

I think having this is worth it. We have to be mindful of when we run this CircleCI, so perhaps not pushing such branches in our peak time.

src/api/config/initializers/airbrake.rb
```diff
@@ -50,8 +50,12 @@
   # A list of parameters that should be filtered out of what is sent to
   # Airbrake. By default, all "password" attributes will have their contents
   # replaced.
-  # https://github.com/airbrake/airbrake-ruby#blacklist_keys
-  c.blacklist_keys = [/password/i, /authorization/i]
+  # https://github.com/airbrake/airbrake-ruby#blocklist_keys
+  if RailsVersion.is_6_1?
```
hennevogel, line 54: This is supposed to be an example or is this really different? :smile:

hennevogel (approved):
We talked a bit offline and decided to just try and see how much this get's in the way of "normal" CI slots.

src/api/config/initializers/airbrake.rb
dmarcoux: It's really different and it serves as an example at the same time. It's a win-win situation!

PR #11030, opened 2021-04-22
Return true in release token model

krauselukas:
Currently the call method in the Token::Release model
returns the ReleaseTarget objects. Since we don't do
anything with the returned objects we can simply return
a boolean value instead.


vpereira:
hey @krauselukas make sure that all `call` methods should return the same thing. Normally it returns the object itself..

krauselukas:
lets deal with this in an appropriate way in another PR :)

PR #11034, opened 2021-04-23
Add uniqueness validation for token string

krauselukas:
On the database level we already ensure the uniqueness
of the token string through an uniq index. In order to
properly test this in the model specs and to have better
exception handling, the validation should also be present
in the token model.

### DIFF
--- a/src/api/app/models/token.rb
+++ b/src/api/app/models/token.rb
@@ -7,6 +7,7 @@ class Token < ApplicationRecord
   has_secure_token :string
 
   validates :user, presence: true
+  validates :string, uniqueness: { case_sensitive: false }
 
   include Token::Errors
 


dmarcoux:
Just to give a bit of background on why `case_sensitive: false` is needed.

The tokens are case-insensitive since the collation of the `string` column is `utf8_unicode_ci`:

https://github.com/openSUSE/open-build-service/blob/ac53b78d83ecde3a93d5c8e7486db3d2b75c8e5b/src/api/db/schema.rb#L1002

So introducing this change makes sense.

src/api/app/models/token.rb
```diff
@@ -7,6 +7,7 @@ class Token < ApplicationRecord
   has_secure_token :string
 
   validates :user, presence: true
+  validates_uniqueness_of :string, case_sensitive: false
```
dmarcoux, line 10: To pass the RuboCop checks:

```suggestion
  validates :string, uniqueness: { case_sensitive: false }
```

I guess this is a tiny bit better since we can group up column validations on a single line. I don't have a strong opinion on this one.

krauselukas: Done, thanks, somehow overlooked that rubocop complains....

PR #11039, opened 2021-04-26
Add specs for token base model

krauselukas:


vpereira (approved):

krauselukas:
[Fully spec out Token models](https://trello.com/c/qnW4vIKS/1649-fully-spec-out-token-models)

PR #11040, opened 2021-04-26
Add specs for release token model

krauselukas:


krauselukas:
[Fully spec out Token models](https://trello.com/c/qnW4vIKS/1649-fully-spec-out-token-models)

vpereira (approved):

PR #11047, opened 2021-04-26
Fix PackageControllerService

### LINKED ISSUE #11035: rebuild via token and project links are not working anymore
I got reports that using the rebuild token broke for builds where the packages are derived via project links are not working anymore since some days.

Please verify in test cases that backend gets a rebuild trigger in such situations after last weeks refactoring

(no more regression label?)

hennevogel:
We were handing down empty repository/arch parameters to the backend which
deal with that with an error.

This also hands out rebuild error message to users.

Fixes #11035


hennevogel:
This is fixed in the recfactoring that just landed in master

PR #11049, opened 2021-04-27
More debug logging

hennevogel:
While working on some feature relating to `Distribution` those helped me to understand what is going on with the validator and fetching files from the interconnect.


vpereira:
there isn't a better way to control logging, enabling it just when it's necessary? Having a way to toggle it or just having it enable in development?

hennevogel:
> there isn't a better way to control logging, enabling it just when it's necessary?

Not sure what you mean. What better way then log levels is there?

vpereira (approved):

PR #11050, opened 2021-04-27
Add a remote boolean for Distribution

hennevogel:
In preparation to move remote Distributions from memcache to a database cache so we can use ActiveRecord for them.

We should not merge this together with the rest of the refactoring (coming later).

### DIFF
--- a/src/api/db/data/20210520160000_set_default_for_distro_remote.rb
+++ b/src/api/db/data/20210520160000_set_default_for_distro_remote.rb
@@ -1,0 +1,14 @@
+class SetDefaultForDistroRemote < ActiveRecord::Migration[6.0]
+  def up
+    Distribution.unscoped.in_batches do |relation|
+      # rubocop:disable Rails/SkipsModelValidations
+      relation.update_all(remote: false)
+      # rubocop:enable Rails/SkipsModelValidations
+      sleep(0.01)
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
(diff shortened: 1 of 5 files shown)


saraycp (approved):

src/api/db/data/20210520160000_set_default_for_distro_remote.rb
```diff
@@ -0,0 +1,14 @@
+class SetDefaultForDistroRemote < ActiveRecord::Migration[6.0]
```
krauselukas, line 1: @hennevogel why is this necessary? you set the default in the db migration through `change_column_default` already and the field is new

hennevogel: https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value

src/api/db/data/20210520160000_set_default_for_distro_remote.rb
```diff
@@ -0,0 +1,14 @@
...
+  def up
```
krauselukas, line 2: @hennevogel why is this necessary? you set the default in the db migration through change_column_default already and the column is new

krauselukas (changes requested):

src/api/db/data/20210520160000_set_default_for_distro_remote.rb
hennevogel: https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value

PR #11052, opened 2021-04-27
Rename TriggerController#create to #trigger

hennevogel:
Use better naming. We don't create tokens here, we trigger them.
This helps with pundit messages etc.


vpereira:
I prefer to keep it as create, as we used to do: http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

PR #11053, opened 2021-04-27
Bugfix 11035

### LINKED ISSUE #11035: rebuild via token and project links are not working anymore

hennevogel:
Fixes #11035

### DIFF
--- a/src/api/app/models/token/rebuild.rb
+++ b/src/api/app/models/token/rebuild.rb
@@ -5,7 +5,7 @@ class Token::Rebuild < Token
 
   def call(options)
     # FIXME: Use the Package#rebuild? instead of calling the Backend directly
-    Backend::Api::Sources::Package.rebuild(package_from_association_or_params.project.name,
+    Backend::Api::Sources::Package.rebuild(project_from_association_or_params.name,
                                            package_from_association_or_params.name,
                                            options)
   end
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -38,8 +38,13 @@ class TriggerController < ApplicationController
   def set_package
     # We need to store in memory the package in order to do authorization
     if @token.package
+      @token.project_from_association_or_params = @token.package.project
       @token.package_from_association_or_params = @token.package
     elsif params[:project] && params[:package]
+      # If params[:project] is a Project that has a project link, then Package.get_by_project_and_name
+      # might get a Package from another Project if the @token.package_find_options allows following
+      # project links. The Token policy needs to make sure to authorize the right object then.
+      @token.project_from_association_or_params = Project.get_by_name(params[:project])
       @token.package_from_association_or_params = Package.get_by_project_and_name(params[:project],
                                                                                   params[:package],
                                                                                   @token.package_find_options)
--- a/src/api/app/models/token.rb
+++ b/src/api/app/models/token.rb
@@ -2,7 +2,7 @@ class Token < ApplicationRecord
   belongs_to :user
   belongs_to :package, inverse_of: :tokens
 
-  attr_accessor :package_from_association_or_params
+  attr_accessor :package_from_association_or_params, :project_from_association_or_params
 
   has_secure_token :string
 
--- a/src/api/app/policies/token/rebuild_policy.rb
+++ b/src/api/app/policies/token/rebuild_policy.rb
@@ -7,7 +7,12 @@ class Token
     def create?
       return false unless record.user.is_active?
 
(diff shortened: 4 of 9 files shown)


src/api/app/models/token/rebuild.rb
```diff
@@ -5,7 +5,7 @@ def self.token_name
 
   def call(options)
     # FIXME: Use the Package#rebuild? instead of calling the Backend directly
-    Backend::Api::Sources::Package.rebuild(package_from_association_or_params.project.name,
+    Backend::Api::Sources::Package.rebuild(project_from_association_or_params.name,
```
adrianschroeter, line 8: most likely crashing in remote project case?

hennevogel: It will never get here as `package_from_association_or_params` will be nil in the controller. You can't instantiate remote packages. It would also not make much sense to trigger a rebuild *for* a project that comes from remote right?

hennevogel:
```ruby
[6] pry(main)> Project.find_by(name: 'openSUSE.org').remoteurl
  Project Load (1.3ms)  SELECT `projects`.* FROM `projects` WHERE `projects`.`id` != 0 AND `projects`.`name` = 'openSUSE.org' LIMIT 1
=> "https://api.opensuse.org/public"
[7] pry(main)> Package.get_by_project_and_name('openSUSE.org:home:hennevogel', 'ponysay')
Trying to find local project openSUSE.org:home, remote_project hennevogel
  Project Load (1.1ms)  SELECT `projects`.* FROM `projects` WHERE `projects`.`id` != 0 AND `projects`.`name` = 'openSUSE.org:home' LIMIT 1
Trying to find local project openSUSE.org, remote_project home:hennevogel
  Project Load (0.8ms)  SELECT `projects`.* FROM `projects` WHERE `projects`.`id` != 0 AND `projects`.`name` = 'openSUSE.org' LIMIT 1
Found local project openSUSE.org for home:hennevogel with remoteurl https://api.opensuse.org/public
=> nil
```

vpereira (approved):

adrianschroeter:
On Dienstag, 27. April 2021, 16:31:04 CEST Henne Vogelsang wrote:
> 
> 
> 
> > @@ -5,7 +5,7 @@ def self.token_name
>  
>    def call(options)
>      # FIXME: Use the Package#rebuild? instead of calling the Backend directly
> -    Backend::Api::Sources::Package.rebuild(package_from_association_or_params.project.name,
> +    Backend::Api::Sources::Package.rebuild(project_from_association_or_params.name,
> 
> It will never get here as `package_from_association_or_params` will be nil in the controller. You can't instantiate remote packages. It would also not make much sense to trigger a rebuild *for* a project that comes from remote right?

the package source may come from remote, but the build is local.

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> the package source may come from remote, but the build is local.

So a project with a project-link to a project that has a project-link to remote?

We did not support this before.  [Have been using](https://github.com/openSUSE/open-build-service/blob/87ff5c460f00f0bc8aecd62f644c7a86f6ec19fc/src/api/app/controllers/trigger_controller.rb#L101) `Package.get_by_project_and_name` on the project and package parameters. It does not support remote projects. See https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/package.rb#L173

adrianschroeter:
On Dienstag, 27. April 2021, 17:53:48 CEST Henne Vogelsang wrote:
> > the package source may come from remote, but the build is local.
> 
> So a project with a project-link to a project that has a project-link to remote?

can also be a direct link, but also this scenario.

> We did not support this before.

we always support it in the api (or have an explicit exception when not).

> [Have been using](https://github.com/openSUSE/open-build-service/blob/87ff5c460f00f0bc8aecd62f644c7a86f6ec19fc/src/api/app/controllers/trigger_controller.rb#L101) `Package.get_by_project_and_name` on the project and package parameters. It does not support remote projects. See https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/package.rb#L173

this is exactly the situation when you have to check for the remote case. nil means that it
may exist remote. it would be an exception when it does not exist.

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> we always support it in the api (or have an explicit exception when not).

Not in this endpoint. At least not for a long time. Let me see what I can do...

PR #11056, opened 2021-04-28
Validate token trigger workflow

danidoni:
including
- definition of the new Token:Workflow,
- validation of the incoming webhook (check if it's a PR/MR with the right actions),
- extraction of some of the data we need from the payload (we still need to add those needed to report back)

### DIFF
--- a/src/api/app/policies/token/workflow_policy.rb
+++ b/src/api/app/policies/token/workflow_policy.rb
@@ -1,0 +1,11 @@
+class Token
+  class WorkflowPolicy < ApplicationPolicy
+    def initialize(_user, record)
+      super(record.user, record)
+    end
+
+    def create?
+      return false unless record.user.is_active?
+    end
+  end
+end
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -9,12 +9,13 @@ class TriggerController < ApplicationController
   skip_before_action :validate_params
   after_action :verify_authorized
 
-  before_action :validate_gitlab_event
+  before_action :validate_gitlab_event, if: :gitlab_webhook?
   before_action :set_token
   before_action :set_package
 
   include Trigger::Errors
 
+  # TODO: split into different controllers, there is some behaviour that only applies to one specific kind of token.
   def create
     authorize @token
     @token.user.run_as do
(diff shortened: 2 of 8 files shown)


eduardoj:
[Create Token::Workflow](https://trello.com/c/trXgK4jU/1644-create-tokenworkflow)

dmarcoux:
Code which is only for `Token::Workflow` should go in that class or in an object which `Token::Workflow` calls. Right now, that code is in `TriggerController` or called by `TriggerController` (like `TriggerControllerService::GithubEventExtractor`).

It would be cleaner to pass the webhook payload to `Token::Workflow` and deal with it in that class.

src/api/app/services/trigger_controller_service/github_event_extractor.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def valid?
+      @http_request.env['HTTP_X_GITHUB_EVENT'] &&
+        @http_request.env['HTTP_X_GITHUB_EVENT'] == 'pull_request' &&
+        @payload['action'].in?(ALLOWED_GITHUB_ACTIONS)
+    end
+
+    # TODO: We also have to deal with wwwurlencoded parameters
+    # TODO: What happens when some of the keys are missing?
+    def add_payload_attributes_to_token(token)
```
dmarcoux, line 18: The payload attributes should be stored in the EventSubscriptions, not in the token. The token might have multiple EventSubscriptions. So if we pass the webhook payload to the `Token::Workflow` call method in the `create` action of the `TriggerController`, then we don't need to store that payload in the token. In the end, we'll do something like this to create the EventSubscriptions (see the payload there with `options[:something]`):

```ruby
    EventSubscription.create!(eventtype: 'Event::BuildSuccess',
                              receiver_role: 'watcher',
                              user: user,
                              channel: 'scm',
                              enabled: true,
                              token: self,
                              payload: {
                                scm: :github,
                                repository_owner: options[:repository_owner],
                                repository_name: options[:repository_name],
                                commit_sha: options[:commit_sha]
                              })
```

src/api/app/policies/token/workflow_policy.rb
```diff
@@ -0,0 +1,13 @@
+class Token
+  class WorkflowPolicy < ApplicationPolicy
+    def initialize(_user, record)
+      super(record.user, record)
+    end
+
+    def create?
+      return false unless record.user.is_active?
+
+      PackagePolicy.new(record.user, record.package_from_association_or_params).update?
```
hennevogel, line 10: Which package is that?

eduardoj: Line removed.

danidoni:
Closing this in favor of #11060

PR #10388, opened 2020-11-04
[api] support per-package release trigger via token

bugfinder:
Allow to specify the source package to be processed on release action.
Supporting also single multibuild container releases

patch from adrian, applied in IBS since Sep 2020

PR #11058, opened 2021-04-29
[api] fix remote and multibuild cases in trigger handling

adrianschroeter:
Not yet tested, but better crashing then corrupting builds

rebuild/release used wrong package in multibuild case and crashed
in remote case before.

-----------------

### DIFF
--- a/src/api/app/models/token/release.rb
+++ b/src/api/app/models/token/release.rb
@@ -7,21 +7,25 @@ class Token::Release < Token
   end
 
   def call(_options)
+    # release from remote is not supported
+    raise ActiveRecord::RecordNotFound if package_from_association_or_params.nil?
     manual_release_targets = package_from_association_or_params.project.release_targets.where(trigger: 'manual')
     raise NoReleaseTargetFound, "#{package_from_association_or_params.project} has no release targets that are triggered manually" unless manual_release_targets.any?
 
     manual_release_targets.each do |release_target|
+      opts = { filter_source_repository: release_target.repository,
+                        manual: true,
+                        comment: 'Releasing via trigger event' }
+      opts[:multibuild_container] = package_name.gsub(/.*:/, '') if package_name.include?(':')
       release_package(package_from_association_or_params,
                       release_target.target_repository,
                       package_from_association_or_params.release_target_name,
-                      { filter_source_repository: release_target.repository,
-                        manual: true,
-                        comment: 'Releasing via trigger event' })
+                      opts)
     end
   end
 
   def package_find_options
-    { use_source: true, follow_project_links: false, follow_multibuild: false }
+    { use_source: true, follow_project_links: false, follow_multibuild: true }
   end
 end
 
--- a/src/api/app/models/token/service.rb
+++ b/src/api/app/models/token/service.rb
@@ -4,6 +4,8 @@ class Token::Service < Token
   end
 
   def call(_options)
+    # we can not work on remote sources
+    raise ActiveRecord::RecordNotFound if package_from_association_or_params.nil?
     Backend::Api::Sources::Package.trigger_services(package_from_association_or_params.project.name,
                                                     package_from_association_or_params.name,
                                                     user.login)
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -40,6 +40,7 @@ class TriggerController < ApplicationController
     if @token.package
       @token.project_from_association_or_params = @token.package.project
       @token.package_from_association_or_params = @token.package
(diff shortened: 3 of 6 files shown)


hennevogel:
I'm already 99% done with this :smile:

adrianschroeter:
On Donnerstag, 29. April 2021, 17:52:03 CEST Henne Vogelsang wrote:
> I'm already 99% done with this :smile: 

with what?

you are aware that the current code is dangerous?

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> with what?

With the usecase you are fixing here. I wrote so in #11053

> you are aware that the current code is dangerous?

Dangerous how? I'm aware it will crash once `package_from_association_or_params` or `:project_from_association_or_params` is something else than a `Package` or `Project` object.

src/api/app/models/token/release.rb
```diff
@@ -7,16 +7,20 @@ def self.token_name
...
+    raise ActiveRecord::RecordNotFound if @token.package_from_association_or_params.nil?
     manual_release_targets = package_from_association_or_params.project.release_targets.where(trigger: 'manual')
     raise NoReleaseTargetFound, "#{package_from_association_or_params.project} has no release targets that are triggered manually" unless manual_release_targets.any?
 
     manual_release_targets.each do |release_target|
-      release_package(package_from_association_or_params,
+      opts = { filter_source_repository: release_target.repository,
+                        manual: true,
+                        comment: 'Releasing via trigger event' }
+      opts[:multibuild_container] = package_name.gsub(/.*:/, '') if package_name.include?(':')
```
hennevogel, line 19: So far we did not support multibuild for release tokens. We used `follow_multibuild: false` for release tokens. See [here](https://github.com/openSUSE/open-build-service/blob/87ff5c460f00f0bc8aecd62f644c7a86f6ec19fc/src/api/app/controllers/trigger_controller.rb#L92-L101).  Which means we did not get past instantiating the package.

```ruby
Package.get_by_project_and_name('OBS:Server:Unstable', 'obs-server:obs-bundled-gems', { follow_multibuild: false })
Trying to find local project OBS:Server, remote_project Unstable
Trying to find local project OBS, remote_project Server:Unstable
Package::Errors::UnknownObjectError: Package not found: OBS:Server:Unstable/obs-server:obs-bundled-gems
from /srv/www/obs/api/app/models/package.rb:193:in `get_by_project_and_name'
```

You want to change this?

src/api/app/models/token/release.rb
```diff
@@ -7,16 +7,20 @@ def self.token_name
   end
 
   def call(_options)
+    # release from remote is not supported
+    raise ActiveRecord::RecordNotFound if @token.package_from_association_or_params.nil?
```
hennevogel, line 11: This class will never be instatiated if this is nil. It'll crash in the Token::ReleasePolicy. See [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/policies/token/release_policy.rb#L10).

src/api/app/models/token/service.rb
```diff
@@ -4,6 +4,8 @@ def self.token_name
   end
 
   def call(_options)
+    # we can not work on remote sources
+    raise ActiveRecord::RecordNotFound if @token.package_from_association_or_params.nil?
```
hennevogel, line 8: You also will never get here because the policy for this Token will not let you. See [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/policies/token/service_policy.rb#L10).

hennevogel (changes requested):

src/api/app/models/token/release.rb
adrianschroeter: it is support in our internal instance where it is crutial. There is a pull request for this in github since
a few month as well:

  https://github.com/openSUSE/open-build-service/pull/10388

without this all media would get released always no matter which multibuild container was specified.

yes, indeed that parameter needs to change as well.

src/api/app/models/token/service.rb
adrianschroeter: this is only for the source service where it is okay.

release and rebuild have different policies.

(it would been better not to ignore the package_find_options now to be able to have a single policy and to have execution, lookup and permission in a single file. seperating the package lookup from permission handling adds security risks IMHO since it needs to be kept in sync).

src/api/app/models/token/release.rb
adrianschroeter: we would not run into this, the release operation needs to have a package object (unlike rebuild).

but it needs to operation on the multibuild container only.

adrianschroeter:
should be solved different now

PR #11059, opened 2021-04-30
Download and parse workflows from SCM

krauselukas:
This PR introduces a basic solution to download the workflow yaml file from the SCM service (GitHub/GitLab) and to create workflows from it. Each workflow and the containing steps are separate objects that encapsulate the attributes and logic to execute it. By using models for it, we can make use of the built in validation capabilities. 

The code introduced here will be called from the `Token::Workflow#call` introduced in https://github.com/openSUSE/open-build-service/pull/11056
Inside the call we would do something like that:

```
yaml_file = Workflows::YAMLDownloadService.new(url: 'https://raw.githubusercontent.com/krauselukas/test_github_deployments/main/.obs/workflow.yaml').call

workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, pr_number: "1234").call

step = workflows.first.steps.first

step.call if step.valid?
```

**TODO:**
  * Write specs

### DIFF
--- a/src/api/Gemfile
+++ b/src/api/Gemfile
@@ -83,6 +83,8 @@ gem 'strong_migrations'
 # for some rails console fancy
 gem 'pry-rails'
 gem 'awesome_print'
+# flexible and safe downloading of remote files
+gem 'down'
 
 group :development, :production do
   # to have the delayed job daemon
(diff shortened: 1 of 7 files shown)


krauselukas:
[Create Token::Workflow](https://trello.com/c/trXgK4jU/1644-create-tokenworkflow)

krauselukas:
going to combine the changes of https://github.com/openSUSE/open-build-service/pull/11056 in an upstream branch

PR #11060, opened 2021-04-30
Workflow token

danidoni:
including:

* definition of the new Token:Workflow
* validation of the incoming webhook (check if it's a PR/MR with the right actions),
* extraction of some of the data we need from the payload


src/api/app/models/token/workflow.rb
```diff
@@ -0,0 +1,71 @@
...
+    # scm_extractor_payload = extractor.extract # returns { scm: 'github', repo_url: 'http://...' }
+
+    # yaml_file = Workflows::YAMLDownloadService.new(scm_extractor_payload).call
+
+    # Read configuration file
+    #   if the ref is not included in the config file's branch whitelist we do nothing.
+
+    # Decide what to do by action:
+    # FIXME: Who does this branching? Should we just overwrite the _service file with the data coming
+    #        from the workflow config file and trigger a Service token?
```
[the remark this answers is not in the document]
dmarcoux: Where is the documentation for that `_branch_request` file?

src/api/app/controllers/trigger_controller.rb
```diff
@@ -5,36 +5,42 @@ class TriggerController < ApplicationController
...
 
   include Trigger::Errors
 
+  # TODO: split into different controllers, there is some behaviour that only applies to one specific kind of token.
   def create
     authorize @token
     @token.user.run_as do
-      @token.call(params.slice(:repository, :arch).permit!)
+      # TODO: This is what we should have in place for the Token::Workflow
+      @token.call(params.slice(:repository, :arch).permit!.merge(request: request))
```
dmarcoux, line 23: This change was introduced since I expected that we would need the request object in the `Token::Workflow` model. I prefer this solution over having another controller only for `Token::Workflow`.

src/api/app/services/workflows/yaml_download_service.rb
```diff
@@ -0,0 +1,33 @@
+module Workflows
+  class YAMLDownloadService
```
dmarcoux, line 2: Just a tiny naming improvement:

```suggestion
  class YAMLDownloader
```

src/api/app/controllers/trigger_controller.rb
danidoni: We opted for a new controller because it was doing too much: half of the controller was dealing with workflow tokens only.

src/api/app/controllers/trigger_workflow_controller.rb
```diff
@@ -0,0 +1,40 @@
+class TriggerWorkflowController < TriggerController
+  before_action :set_scm_event
+  before_action :validate_scm_event
+
+  # TODO: split into different controllers, there is some behaviour that only applies to one specific kind of token.
+  def create
+    authorize @token
+    @token.user.run_as do
+      @token.call({ scm: scm, event: event, payload: payload })
```
saraycp, line 9: We don't need the `{ }` here.

src/api/app/controllers/trigger_controller.rb
dmarcoux: Yes, but my point is to pass the request object to the `Token::Workflow` model in the `call` method and everything unique to that model will be handled in the model itself.

saraycp: There are two different things:

1) What arguments should we pass to the `Token::Workflow`? 
Even if we don't split the code into two controllers, we didn't like the idea of passing the whole request to the model. We prefer to pass what the model needs instead. If we don't split the controller, it will be something like:
```@token.call(params.slice(:repository, :arch).permit!.merge({ scm: scm, event: event, payload: payload }))```

2) Having one or two controllers?
We need to do some checks that correspond only to workflow tokens. That's why we considered that checking different conditions was no so cool and having its own controller is :)

And after our decision in 1) we also need to get `scm`, `event` and `payload`, that are only necessary for workflow tokens.

PD: the `.merge(request: request))` was added during the rebase probably, we should remove it if we agree to go on with our approach.

File: src/api/app/models/workflow/step/branch_package_step.rb [hunk 5]
```diff
@@ -0,0 +1,77 @@
...
+        false
+      end
+
+      def check_source_access
+        return if remote_source?
+
+        options = { use_source: false, follow_project_links: true, follow_multibuild: true }
+        src_package = Package.get_by_project_and_name(source_project, source_package, options)
+
+        raise Pundit::NotAuthorizedError unless PackagePolicy.new(User.session, src_package).create_branch?
```
hennevogel, line 52:
```suggestion
        authorize src_package, :create_branch?
```

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+
+      def branch
```
hennevogel, line 55: *nitpick* We should DRY this right?

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -0,0 +1,77 @@
...
+        source_project + ":PR-#{@pr_number}"
+      end
+
+      def remote_source?
+        return true if Project.find_remote_project(source_project)
+
+        false
+      end
+
+      def check_source_access
```
hennevogel, line 46: All of this could also be a BranchPackageStep policy no?

src/api/app/services/scm_status_reporter.rb
```diff
@@ -0,0 +1,49 @@
+class SCMStatusReporter
+  attr_accessor :payload, :scm_token, :state
+
+  def initialize(payload, scm_token, event = nil)
```
hennevogel, line 4: *nitpick* `event=nil` to convey that this is the initial reporting is kind of hard to understand. Also makes setting `@state` below hard to understand.

src/api/app/services/trigger_controller_service/scm_extractor.rb
```diff
@@ -0,0 +1,56 @@
+module TriggerControllerService
+  # NOTE: this class is coupled to GitHub pull requests events and GitLab merge requests events.
+  class ScmExtractor
+    ALLOWED_GITHUB_ACTIONS = ['opened', 'synchronize', 'closed'].freeze
+    ALLOWED_GITLAB_ACTIONS = ['open', 'update', 'close', 'reopen', 'merge'].freeze
+
+    def initialize(scm, event, payload)
+      # TODO: What should we do when the user sends a wwwurlencoded payload? Raise an exception?
```
hennevogel, line 8: Why would you not allow to fake this? :)

src/api/app/services/trigger_controller_service/scm_extractor.rb
```diff
@@ -0,0 +1,56 @@
...
+      @payload = payload
+      @scm = scm
+      @event = event
+    end
+
+    def accepted_event_and_action?
```
hennevogel, line 14: Allowed or accepted. Please decide.

src/api/app/models/token/workflow.rb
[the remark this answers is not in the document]
hennevogel: @adrianschroeter yeah we are going to use this. We should talk about the format though, I think it would be more consistent if this would be (more specific) XML and not a json dump.

src/api/app/models/workflow/step/branch_package_step.rb
dmarcoux: @krauselukas and I talked about this last week and `authorize` isn't available outside controllers. I just found out by looking at the Pundit code that `Pundit.authorize` is exactly the same, but also available outside controllers. With this in mind, I would suggest doing this:

```suggestion
        Pundit.authorize src_package, :create_branch?
```

src/api/app/services/trigger_controller_service/scm_extractor.rb
saraycp: What do you mean?

A bit of context: when defining webhooks in GitHub, the user can choose between `application/x-www-form-urlencoded` or `application/json` content types, whereas GitLab only allows `application/json`.

In our user documentation we can encourage the user to always choose `application/json` ... or should we simply accept and handle both?

adrianschroeter:
On Montag, 3. Mai 2021, 12:47:45 CEST Henne Vogelsang wrote:
> 
> 
> 
> > +                                enabled: true,
> +                                token: self,
> +                                payload: payload)
> +    end
> +
> +    SCMStatusReporter.new(extractor.extract, scm_token).call
> +    # scm_extractor_payload = extractor.extract # returns { scm: 'github', repo_url: 'http://...' }
> +
> +    # yaml_file = Workflows::YAMLDownloadService.new(scm_extractor_payload).call
> +
> +    # Read configuration file
> +    #   if the ref is not included in the config file's branch whitelist we do nothing.
> +
> +    # Decide what to do by action:
> +    # FIXME: Who does this branching? Should we just overwrite the _service file with the data coming
> +    #        from the workflow config file and trigger a Service token?
> 
> @adrianschroeter yeah we are going to use this. We should talk about the format though, I think it would be more consistent if this would be (more specific) XML and not a json dump.

The dump is what from the remote server comes on purpose.

That way no OBS code needs to changed when remote server is changing or adding 
new functionality. It can get directly adapted in the service codes which deals
with it.

This also helps with remote incompatible changes, that way the OBS code can
stay compatible as it has to. And everybody can extent functionality without
the need to be an OBS admin.

-- 

Adrian Schroeter ***@***.***>

src/api/app/services/trigger_controller_service/scm_extractor.rb
saraycp: Renamed to `allowed_*" to be consistent with the constant names that the methods use: `ALLOWED_GITHUB_ACTIONS` and `ALLOWED_GITLAB_ACTIONS`.

src/api/app/controllers/trigger_workflow_controller.rb
saraycp: done

src/api/app/controllers/trigger_controller.rb
saraycp: I already removed `.merge(request: request))` part from the `TriggerController`. We pass `scm`, `event` and `payload` from `TriggerWorkflowController` instead.

src/api/app/services/workflows/yaml_download_service.rb
danidoni: Done

danidoni:
> And everybody can extent functionality without the need to be an OBS admin.

By OBS admin you mean an OBS developer?

adrianschroeter:
On Dienstag, 4. Mai 2021, 09:52:21 CEST Dani Donisa wrote:
> > And everybody can extent functionality without the need to be an OBS admin.
> 
> By OBS admin you mean an OBS developer?

both actually ;)

-- 

Adrian Schroeter ***@***.***>

src/api/app/controllers/trigger_controller.rb
vpereira: @dmarcoux the only thing that won't work is to have workflow tokens reaching the same action as the other tokens. Workflow tokens don't need to run the filter `set_package`. So even if you want to have both in one controller, they won't have the same entrypoint. So i guess, its better a new controller with one action `create` than a custom action name on the `TriggerController`

src/api/app/models/workflow/step/branch_package_step.rb
vpereira: `Pundit.authorize` is just `authorize` but with `include Pundit` being called before, no? In another hand is that a good thing to leak Pundit authorization to models?

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      event.update(mails_sent: true) if subscribers.empty?
+
+      begin
+        event.subscriptions(:scm).each do |subscription|
+          SCMStatusReporter.new(subscription.payload, subscription.token.scm_token, event).call
+        end
+      rescue StandardError => e
+        Airbrake.notify(e, event_id: event.id)
+      ensure
+        event.update(mails_sent: true)
```
hennevogel, line 16: How do you know that the emails are sent? :-)

src/api/app/models/workflow/step/branch_package_step.rb
dmarcoux: I don't see how we can do authorization in the controller for this since we need to parse the YAML file, then check if the token user is authorized or not to do whatever is defined in the YAML file.

src/api/app/controllers/trigger_controller.rb
```diff
@@ -5,36 +5,42 @@ class TriggerController < ApplicationController
   skip_before_action :extract_user
   # Authentication happens with tokens, so no login is required
   skip_before_action :require_login
+  # GitLab sends data as parameters which are not strings, like the project hash.
+  skip_before_action :validate_params
```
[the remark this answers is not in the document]
saraycp: After more detailed tests, it is clear that the `validate_params` can fail in both GitHub and GitLab webhooks. Each of them can send a non-string value as parameters when the content type is `application/json`.

For example, GitHub sends an integer value for "number" key (PR number) and GitLab sends a hash representing the project, the user, etc.

This is my proposal now: https://github.com/openSUSE/open-build-service/pull/11077/commits/0af6ac95e5165bfe42f7876acfd6281a0191f751 (commit in PR #11077)

src/api/app/services/scm_status_reporter.rb
```diff
@@ -0,0 +1,49 @@
...
+    @payload = payload
+    @scm_token = scm_token
+
+    @state = event.nil? ? 'pending' : scm_final_state(event)
+  end
+
+  def call
```
hennevogel, line 11: I still don't understand where you filter for the project/package. You only want to report for  a *specific* package in a *specific* project right?

dmarcoux: This is handled in the job.

dmarcoux:
I've rebased the branch on master after we started to merge split parts of this PR and merged them into `master`. I also fixed conflicts regarding some recent changes in the `trigger_controller.rb`.

src/api/app/controllers/trigger_workflow_controller.rb
```diff
@@ -0,0 +1,41 @@
+class TriggerWorkflowController < TriggerController
+  skip_before_action :set_package
+  before_action :set_scm_event
+  before_action :validate_scm_event
+
+  # TODO: split into different controllers, there is some behaviour that only applies to one specific kind of token.
+  def create
```
hennevogel, line 7: Feature flag?

src/api/app/models/event/base.rb
```diff
@@ -105,9 +107,10 @@ def receiver_roles
     end
 
     def initialize(attribs)
-      attributes = attribs.dup
+      attributes = attribs.dup.with_indifferent_access
       super()
       self.created_at = attribs[:time] if attributes[:time]
+      self.package = Package.get_by_project_and_name(attributes[:project], attributes[:package]) if attributes[:project] && attributes[:package]
```
hennevogel, line 113: This will make the whole Event creation barf if this goes wrong (there are several exceptions raised inside this method). Also many of the events (branch_command or build*) could be about packages that come over project links and/or from remote so there is no object to get.

src/api/app/services/trigger_controller_service/scm_extractor.rb
hennevogel: Ah sorry I was not aware of this. I think making json mandatory  in this case it a good compromise. Doesn't code anything...

vpereira:
everything is on master

PR #11070, opened 2021-05-03
Reduce complexity of ScmStatusReporter#initialize

saraycp:
Attending to Code Climate complains.

-----------------

### DIFF
--- a/src/api/app/services/scm_status_reporter.rb
+++ b/src/api/app/services/scm_status_reporter.rb
@@ -5,21 +5,7 @@ class SCMStatusReporter
     @payload = payload
     @scm_token = scm_token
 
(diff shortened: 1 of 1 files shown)


src/api/app/services/scm_status_reporter.rb
```diff
@@ -8,18 +8,7 @@ def initialize(payload, scm_token, event = nil)
     # Depending on the SCM, it's different
```
dmarcoux, line 8: This comment and the 2 following comment lines should also be moved to the new method.

saraycp: done

vpereira (approved):

PR #11071, opened 2021-05-03
Little fixes

saraycp:
Add little improvements

- Drop unnecessary curly brackets from method parameters.
- Don't assign to variable values that are going to be used only once.
- Add .with_indifferent_access to SCMExtractor resulting payload.
- Adjust comments.
- Rename 'accepted*_event_and_action' to 'allowed*_event_and_action'
  to be consistent with the constant names (ALLOWED_GITHUB_ACTIONS,
  ALLOWED_GITLAB_ACTIONS).

-----------------


saraycp:
Wrong target branch! :-1:

PR #11074, opened 2021-05-03
[api] fix hostsystem repository handling

adrianschroeter:
backend supports multiple entries. They are just a different kind of
path elements. So handling them via a flag in path elements instead
of heaving a single fixed entry per repository (actually per project
due to bug in former implementation).

-----------------

### DIFF
--- a/src/api/app/models/project/update_from_xml_command.rb
+++ b/src/api/app/models/project/update_from_xml_command.rb
@@ -183,7 +183,7 @@ class Project
 
       update_repository_flags(current_repo, xml_hash)
       update_release_targets(current_repo, xml_hash)
-      update_hostsystem(current_repo, xml_hash)
+      current_repo.save! if current_repo.changed?
       update_repository_architectures(current_repo, xml_hash)
       update_download_repositories(current_repo, xml_hash)
 
@@ -195,9 +195,20 @@ class Project
     def update_path_elements(current_repo, xml_hash)
       # destroy all current pathelements
       current_repo.path_elements.destroy_all
-      return unless xml_hash['path']
+      return unless xml_hash['path'] || xml_hash['hostsystem']
 
-      # recreate pathelements from xml
+      # recreate hostsystem elements from xml
+      position = 1
+      xml_hash.elements('hostsystem') do |hostsystem|
+        host_repo = Repository.find_by_project_and_name(hostsystem['project'], hostsystem['repository'])
+        raise SaveError, 'Using same repository as hostsystem element is not allowed' if hostsystem['project'] == project.name && hostsystem['repository'] == xml_hash['name']
+        raise SaveError, "Unknown hostsystem repository '#{hostsystem['project']}/#{hostsystem['repository']}'" unless host_repo
+
+        current_repo.path_elements.new(link: host_repo, position: position, kind: :hostsystem)
+        position += 1
+      end
+
+      # recreate path elements from xml
       position = 1
       xml_hash.elements('path') do |path|
         link_repo = Repository.find_by_project_and_name(path['project'], path['repository'])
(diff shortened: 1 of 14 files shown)


src/api/db/migrate/20210603113812_fix_hostsystem_definition.rb
```diff
@@ -0,0 +1,17 @@
+class FixHostsystemDefinition < ActiveRecord::Migration[6.0]
+  class SchemaMigration < ApplicationRecord
+  end
+
+  def up
+    # no real production data inside yet, so we skip any conversation.
+    safety_assured { remove_column :repositories, :hostsystem_id }
```
dmarcoux, line 7: It doesn't matter if the column has production data or not, removing a column is **not** safe. Please follow the recommendations from [strong_migrations](https://github.com/ankane/strong_migrations#removing-a-column) on how to safely drop a column. Without this, we're going to be down once this migration runs.

dmarcoux (changes requested):

src/api/db/migrate/20210603113812_fix_hostsystem_definition.rb
adrianschroeter: that way?

dmarcoux: 1. Change the code to ignore the column you want to drop.
2. Deploy
3. Create a migration to remove the column
4. Deploy and run the migration.

This cannot happen all at the same time. It has to be done in two steps, so it will have to be in 2 PRs. It's a bit more work, but then we're not down because we remove a column.

adrianschroeter: as I said, this column couldn't have used before, if it was the result would be broken anyway.

But we can move just the removal to another later migration if wanted.

dmarcoux: It doesn't matter if this column is used or not, ActiveRecord is caching the table columns. This is why it's crashing if you simply remove a column like you intended to.

Please remove the commented-out lines from the migration because this will be confusing anybody looking at the code.

src/api/db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_02_05_113812) do
+ActiveRecord::Schema.define(version: 2021_06_03_113812) do
```
dmarcoux, line 13: This version number is off. How did you create the migration included in this PR? You edited its filename by hand, right? Because this timestamp is in the future, so it cannot be possible that you generated it with `bundle exec rails g migration MyMigration123`.

src/api/db/migrate/20210603113812_fix_hostsystem_definition.rb
dmarcoux: The database schema is also wrong now since the column `hostsystem_id` is still being removed.

adrianschroeter:
On Montag, 3. Mai 2021, 17:03:11 CEST Dany Marcoux wrote:
> @dmarcoux commented on this pull request.
> 
> 
> 
> > @@ -0,0 +1,17 @@
> +class FixHostsystemDefinition < ActiveRecord::Migration[6.0]
> +  class SchemaMigration < ApplicationRecord
> +  end
> +
> +  def up
> +    # no real production data inside yet, so we skip any conversation.
> +    safety_assured { remove_column :repositories, :hostsystem_id }
> 
> The database schema is also wrong now since the column `hostsystem_id` is still being removed.

that is not important, as it should be created anyway and the content can not 
get used for anything so far.

-- 

Adrian Schroeter ***@***.***>

src/api/app/models/project/update_from_xml_command.rb
```diff
@@ -195,9 +195,22 @@ def update_repository_without_path_element(xml_hash)
...
       # destroy all current pathelements
       current_repo.path_elements.destroy_all
-      return unless xml_hash['path']
+      return unless xml_hash['path'] || xml_hash['hostsystem']
+
+      # recreate hostsystem elements from xml
+      position = 1
+      xml_hash.elements('hostsystem') do |hostsystem|
+        host_repo = Repository.find_by_project_and_name(hostsystem['project'], hostsystem['repository'])
+        if xml_hash['hostsystem']['project'] == project.name && xml_hash['hostsystem']['repository'] == xml_hash['name']
```
coolo, line 204: careful. You use `hostsystem` as `xml_hash['hostsystem']` and you should continue using that variable to access 'repository' and 'project'
Because  if 'hostsystem' can appear multiple times (as is the topic of your PR), then there is no single xml_hash['hostsystem']['project']

The fact that you didn't spot this yourself, shows a certain lack of test coverage

adrianschroeter: yeah .. that is broken indeed ... (current code is even more broken though as it removes hostsystem even in other repositories)

adrianschroeter: will add a simple test and a fix tomorrow morning, need to finish something else atm

adrianschroeter:
added a simple store and render test case now and fixed some more rubocop complains.

I will not touch it anymore except on request to avoid conflicts. But it would be good to get it deployed today, since it blocks the work on cross building projects atm.

hennevogel:
Pushed re-worked migrations, they are safe to run now. 

There is a spec for `Repository.hostsystem` though. Should this be removed?

adrianschroeter:
at least removed in that form, (we could re-introduce it in a form where it tests multiple hostsystem definitions in multiple repositories)

adrianschroeter:
did so now, thanks for updating the requests!

adrianschroeter:
assuming hennes push as review

PR #11073, opened 2021-05-03
Handle all kinds of OBS Links in the token controller

hennevogel:
- Remote projects
- Project with a project-link to a Project
- Project with a project-link to a remote Project
- Multibuild flavors

PR #11075, opened 2021-05-03
Follow #11073

hennevogel:


hennevogel:
we are going the other way around now...

PR #11077, opened 2021-05-03
Add some fixes and improvements

saraycp:
- Stop passing `request` to the `Token::Workflow#call`
- Skip `validate_params` for both GitHub and GitLab 

-----------------

### DIFF
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -19,8 +19,7 @@ class TriggerController < ApplicationController
   def create
     authorize @token
     @token.user.run_as do
-      # TODO: This is what we should have in place for the Token::Workflow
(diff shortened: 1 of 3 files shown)


eduardoj (approved):

src/api/app/controllers/trigger_controller.rb
```diff
@@ -20,7 +21,7 @@ def create
     authorize @token
     @token.user.run_as do
       # TODO: This is what we should have in place for the Token::Workflow
```
dmarcoux, line 23: This comment can also be removed.

saraycp: updated :+1:

danidoni (approved):

krauselukas (approved):

PR #11078, opened 2021-05-04
Return a "Bad request" error when the HTTP request headers are not valid ...

eduardoj:
... , for the trigger/workflow endpoint.

The response returned previously was a "Forbidden" error, that in this case makes little sense.

Also add a controller test for this case.

-----------------


eduardoj:
Adding this test will be handled in #11103. Closing.

PR #11080, opened 2021-05-04
Rely on CreateJob class for ReportToScmJob

dmarcoux:


vpereira (approved):

danidoni (approved):

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -1,20 +1,19 @@
...
-      begin
-        event.subscriptions(:scm).each do |subscription|
-          SCMStatusReporter.new(subscription.payload, subscription.token.scm_token, event).call
-        end
-      rescue StandardError => e
-        Airbrake.notify(e, event_id: event.id)
-      ensure
-        event.update(mails_sent: true)
-      end
+    EventSubscription.joins('INNER JOIN events ON event_subscriptions.eventtype = events.eventtype')
```
vpereira, line 9: we have to filter here per project and/or package name

vpereira (changes requested):

src/api/app/jobs/report_to_scm_job.rb
hennevogel: Maybe we should have `package` & `project` columns in `Event` & `EventSubscription`?

hennevogel: We came across this "need" a couple of time already.

dmarcoux: This is now handled with the changes we pushed today.

src/api/app/models/event/base.rb
```diff
@@ -105,9 +107,10 @@ def receiver_roles
     end
 
     def initialize(attribs)
-      attributes = attribs.dup
+      attributes = attribs.dup.with_indifferent_access
```
dmarcoux, line 110: This is the change which is causing one spec to fail.

dmarcoux: @vpereira fixed the spec!

PR #11081, opened 2021-05-04
Use the SCM's branch instead of the default one

danidoni:
We use the SCM's branch instead of the repository's default one


danidoni:
This PR has some security implications, as we would run the `.obs/workflows.yml` file from someone else's branch :S

PR #11089, opened 2021-05-05
Stop validating parameters in TriggerController

hennevogel:
All SCMs send data in there that is parsed to hashes for instance.

We removed this in the last refactoring...


krauselukas (approved):

eduardoj (approved):

hennevogel:
@eduardoj @krauselukas there is now a second commit....

saraycp (approved):

PR #11091, opened 2021-05-05
Use getmultibuild backend view for flavors

hennevogel:
The backend knows the definitive answer to available multibuild flavors on this end-point that takes all kinds of corner cases (links, broken links, services etc.) into account. Let's make use of it.

### DIFF
--- a/src/api/spec/models/package_spec.rb
+++ b/src/api/spec/models/package_spec.rb
@@ -834,4 +834,25 @@ RSpec.describe Package, vcr: true do
       it { expect { subject }.to raise_error(Package::CycleError) }
     end
   end
+
+  describe '.multibuild_flavor?', vcr: false do
+    let(:multibuild_package) { create(:package, name: 'multibuild_package', project: home_project) }
+
+    context 'with multibuild' do
+      before do
+        allow(Backend::Api::Sources::Package).to receive(:multibuild_flavors).and_return('<directory><entry name="flavor_a"/><entry name="flavor_c"/></directory>')
+      end
+
+      it { expect(multibuild_package).to be_multibuild_flavor('flavor_a') }
+      it { expect(multibuild_package).to be_multibuild_flavor('flavor_c') }
+    end
+
+    context 'without multibuild' do
+      before do
+        allow(Backend::Api::Sources::Package).to receive(:multibuild_flavors).and_return('<directory />')
+      end
+
+      it { expect(multibuild_package).not_to be_multibuild }
+    end
+  end
 end
--- a/src/api/spec/models/package_spec.rb
+++ b/src/api/spec/models/package_spec.rb
@@ -835,7 +835,7 @@ RSpec.describe Package, vcr: true do
     end
   end
 
-  describe '.multibuild_flavor?', vcr: false do
+  describe '.last_build_multibuild_flavor?', vcr: false do
     let(:multibuild_package) { create(:package, name: 'multibuild_package', project: home_project) }
 
     context 'with multibuild' do
@@ -843,8 +843,8 @@ RSpec.describe Package, vcr: true do
         allow(Backend::Api::Sources::Package).to receive(:multibuild_flavors).and_return('<directory><entry name="flavor_a"/><entry name="flavor_c"/></directory>')
       end
 
-      it { expect(multibuild_package).to be_multibuild_flavor('flavor_a') }
-      it { expect(multibuild_package).to be_multibuild_flavor('flavor_c') }
+      it { expect(multibuild_package).to be_last_build_multibuild_flavor('flavor_a') }
+      it { expect(multibuild_package).to be_last_build_multibuild_flavor('flavor_c') }
     end
 
     context 'without multibuild' do
@@ -852,7 +852,7 @@ RSpec.describe Package, vcr: true do
         allow(Backend::Api::Sources::Package).to receive(:multibuild_flavors).and_return('<directory />')
       end
 
-      it { expect(multibuild_package).not_to be_multibuild }
+      it { expect(multibuild_package).not_to be_last_build_was_multibuild }
     end
   end
 end
--- a/src/api/app/lib/backend/api/sources/package.rb
+++ b/src/api/app/lib/backend/api/sources/package.rb
@@ -137,6 +137,11 @@ module Backend
                                                        params: options, accepted: [:repository, :arch])
         end
 
+        # Returns the multibuild flavors for this package
+        def self.multibuild_flavors(project_name, package_name)
+          http_get(['/source/:project/:package', project_name, package_name], params: { view: :getmultibuild })
+        end
+
         # Returns the content of the source file
         # @return [String]
         def self.file(project_name, package_name, file_name)
--- a/src/api/app/models/concerns/multibuild_package.rb
+++ b/src/api/app/models/concerns/multibuild_package.rb
@@ -15,7 +15,10 @@ module MultibuildPackage
   end
 
   def multibuild?
-    file_exists?('_multibuild')
+    @multibuild_flavors ||= Xmlhash.parse(Backend::Api::Sources::Package.multibuild_flavors(project, name)).elements('entry').collect { |x| x['name'] }
+    return false if @multibuild_flavors.blank?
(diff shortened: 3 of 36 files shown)


src/api/spec/models/package_spec.rb
```diff
@@ -834,4 +834,11 @@
       it { expect { subject }.to raise_error(Package::CycleError) }
     end
   end
+
+  describe '.multibuild_flavor?' do
+    let(:multibuild_package) { create(:multibuild_package, name: 'multibuild_package', project: home_project, flavors: ['flavor_a', 'flavor_b']) }
+
+    it { expect(multibuild_package).to be_multibuild_flavor('flavor_a') }
```
eduardoj, line 841:
```suggestion
    let(:multibuild_package) { create(:multibuild_package, name: 'multibuild_package', project: home_project, flavors: ['flavor_a', 'flavor_b'], package: ['flavor_c']) }

    it { expect(multibuild_package).to be_multibuild_flavor('flavor_a') }
    it { expect(multibuild_package).to be_multibuild_flavor('flavor_c') }
```

The code is introducing parsing of `package` elements. Shouldn't we also add this to the tests, as suggested? In that case we should also modify the multibuild_package factory.

hennevogel:
> you notice that you work sometimes on expanded and sometime on none expanded sources?

Can you be a tad more specific what you are talking about...

hennevogel:
I guess you are talking about if the if the `_mulitbuild` file comes from the package link or not. Why would that matter?

coolo:
For what it's worth: I'm with Henne here, the frontend better understands the concept of multibuild and offers proper handling of it. If the runtime becomes a problem (which I doubt for triggers), then we can still look into making it a proper database field.

adrianschroeter:
A database field of _multibuild flavors need to get updated when linked package sources are changing and when server side services are running. Do you really want to add this complexity and overhead?

coolo:
Yes, I want to have the webui have full access to all informations - quickly. For a while I thought redis could be our saviour, but it wasn't really picked up from the frontend side. In that model, the backend itself would store the multibuild information - not all databases are SQL.

adrianschroeter:
On Dienstag, 11. Mai 2021, 12:40:02 CEST Henne Vogelsang wrote:
> > removed unnecessary code is for sure more readable...
> 
> Lack of clarity and digestible information about "special" cases and behavior is one of the things we suffer most from. You see this from the couple of refactorings in the Token area for instance. At some point we have to start clarifying. This point for this area is *now* as we are extending this with new features. 

These are not special cases, but just the normal features which you can combine.

The api has no reason to digest this information for any functionality where the api
is responsible.

If you can not oversee and handle them you should not try to and breaking existing functionality...

> > Atm links to _multibuild are broken, since you check unexpanded code for a file but operate on expanded.
> 
> Thanks for the hint, pushed changes that always expand for this check. Any other "special" case we are missing?

plenty, handle plenty _link instances, multi _multibuild files, service generated ones and the combination
of all these. And keep in mind that the source state must not fit the build state since there is a time 
delay, in cause you use this for any build result related code.

-- 

Adrian Schroeter ***@***.***>

mlschroe:
So what *is* it that the API wants to get by parsing the multibuild file? What *is* your need?

The problem really is that the data can change at any point in time. The multibuild file can come from a _link from any other project, even from a remote OBS instance. Even the backend isn't really sure about the current state. Even more, a broken link does not mean that the multibuild information is gone, it means that it is frozen to the old state. This is because we can't throw away old build results just because the source is broken or a remote server is unavailable.

Thus we fake it a bit in the backend: the source server stores the multibuild result when the scheduler asks it for the parsed sources. And it uses that stored data instead of parsing the multibuild file for requests.

If you want to have that information in the API (I don't really know for what reason), we need to add a new notification type that has the updated multibuild data as payload. Parsing the multibuild file by yourself is not going to work for all cases, I'm afraid.

hennevogel:
> So what is it that the API wants to get by parsing the multibuild file? What is your need?

To know if a string we get from the user is a valid multibuild flavor. For instance:

- Will the backend accept a rebuild call for this string or will it throw an exception?
- Should we display some page for this multibuild flavor or not?

> The problem really is that the data can change at any point in time. 

And that is fine, we are talking about some specific point in time. We are not talking about duplicating this data in the frontend. Just querying it.

> the source server stores the multibuild result when the scheduler asks it for the parsed sources. And it uses that stored data instead of parsing the multibuild file for requests.

How about we expose this stored data via some backend API we can query?

mlschroe:
There's already a query method:
```
/source/$project/$package?view=getmultibuild
```

mlschroe:
(Marco added that in commit ff7b9d901d2018d01202853ad67ee65383fe3d55)

mlschroe:
(I think he wanted to use that in osc, but then decided to do it in a different way. I'll need to ask him.)

hennevogel:
Okay something like that looks good :+1: 

As far as I understand this is "only" taking the cache into account right? So this would only return flavors after the first build (scheduler asks it for the parsed source) has been started right?

hennevogel:
And would it be possible to make it fill the cache if it's not fresh yet?

adrianschroeter:
there is no cache, it is the source view.

But on other contexts you would need the version of older sources use in last build.

And of course updates needs to happen when any indirect used source got updated as mentioned bevore.

hennevogel:
> But on other contexts you would need the version of older sources use in last build.

Not sure I understand what you are trying to say with this sentence. Revisions?

> And of course updates needs to happen when any indirect used source got updated as mentioned bevore.

Updates of what exactly?

hennevogel:
> there is no cache, it is the source view.

Hmmm, are you sure? https://github.com/openSUSE/open-build-service/blob/master/src/backend/BSSrcServer/Multibuild.pm#L44-L56

adrianschroeter:
On Montag, 17. Mai 2021, 17:31:50 CEST Henne Vogelsang wrote:
> > there is no cache, it is the source view.
> 
> Hmmm, are you sure? https://github.com/openSUSE/open-build-service/blob/master/src/backend/BSSrcServer/Multibuild.pm#L44-L56 

that cache is irrelevant for the flavors of the last build. The source revision of the last
build need to be used to determine the flavors of that one.

And as said before, when you look at the sources and store the result in a db
you need to update whenever any source is changing in the link chain. 

But you can not use that state when doing operations inside of the build/ tree.

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> that cache is irrelevant for the flavors of the last build. The source revision of the last build need to be used to determine the flavors of that one.

We already established this. And this is exactly what this cache is doing as far as I understand Michael.

> And as said before, when you look at the sources and store the result in a db you need to update whenever any source is changing in the link chain.

We also already established that no one is going to story anything in some database.

hennevogel:
@mlschroe questions still stands: Would it be possible to fill the cache, if empty, when reading it through this API?

hennevogel:
Made use of the backend route now. I'm still a bit worried that we might want to know about flavors before `the scheduler asks it for the parsed sources`.

adrianschroeter:
On Donnerstag, 20. Mai 2021, 14:17:45 CEST Henne Vogelsang wrote:
> Made use of the backend route now. I'm still a bit worried that we might want to know about flavors before `the scheduler asks it for the parsed sources`. 

again, you anyway need to use it in context of the operation you plan to do it.

When it is about the current source definition it something else when you want to operate on build results.

And even when the source state is broken, you still need to allow certain operations like
wipe binaries or using it for some views for example.

Yes, currently not possible via tokens, but when you create a model, I suppose you want
to use it also elsewhere in future.

-- 

Adrian Schroeter ***@***.***>

adrianschroeter:
You should rename "self.multibuild_flavors" method to something like "last_build_state_multibuild_flavors" at least to avoid a wrong usage at least. It is not the current source state and neither the last finished build.

So you break at least the situation where the last build result should get released and the flavors have changed meanwhile with this code atm. Yes, special case, but a regression.

hennevogel:
@mlschroe ping :wave:

hennevogel:
I renamed the method now...

PR #11092, opened 2021-05-05
Create _branch_request file

krauselukas:


src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -96,12 +99,40 @@ def gitlab_merge_request?
...
+
+        package.save_file({ file: branch_request_file, filename: '_branch_request' })
+      end
+
+      def branch_request_file_github
+        {
+          action: @scm_extractor_payload[:action],
+          pull_request: {
+            head: {
+              repo: { full_name: @scm_extractor_payload[:full_name] },
```
eduardoj, line 124:
```suggestion
              repo: { full_name: @scm_extractor_payload[:repository_full_name] },
```
Nitpick: `@scm_extractor_payload` has a couple of keys more. Here it is clear what `:full_name` refers to, but I find it more obvious to change that key to `:repository_full_name`.

eduardoj (approved):

src/api/app/models/workflow/step/branch_package_step.rb
krauselukas: Yes you are right, will change it

krauselukas: @eduardoj done

PR #11100, opened 2021-05-06
Create Workflows::YAMLToWorkflowsService

dmarcoux:

**TODO**:
- [x] Specs

### DIFF
--- a/src/api/app/services/workflows/yaml_to_workflows_service.rb
+++ b/src/api/app/services/workflows/yaml_to_workflows_service.rb
@@ -1,0 +1,24 @@
+module Workflows
+  class YAMLToWorkflowsService
+    def initialize(yaml_file:, scm_extractor_payload:)
+      @yaml_file = yaml_file
+      @scm_extractor_payload = scm_extractor_payload
+    end
+
+    def call
+      create_workflows
+    end
+
+    private
+
+    def create_workflows
+      parsed_workflows_yaml = YAML.safe_load(File.read(@yaml_file))
+      workflows = []
+
+      parsed_workflows_yaml.each do |_workflow_name, workflow|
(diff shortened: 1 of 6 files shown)


src/api/app/services/workflows/yaml_to_workflows_service.rb
```diff
@@ -0,0 +1,24 @@
+module Workflows
+  class YAMLToWorkflowsService
```
dmarcoux, line 2: It would be nice if we can have a better name for this. Two times workflow and having service in the name is redundant.

eduardoj, line 2:
```suggestion
  class ParseYAMLFileService
```
Only as suggestion.

src/api/app/services/workflows/yaml_to_workflows_service.rb
```diff
@@ -0,0 +1,24 @@
...
+      @scm_extractor_payload = scm_extractor_payload
+    end
+
+    def call
+      create_workflows
+    end
+
+    private
+
+    def create_workflows
```
dmarcoux, line 14: This isn't creating workflows. It is initializing workflows since it's doing `Workflow.new(...)`.

dmarcoux:
Just pushed to fix a conflict in Zeitwerk initializer.

krauselukas:
@dmarcoux the specs/service depends on the `Workflow` model, so we either wait for the model to be on master or put it in to this PR as well. 
@dmarcoux @eduardoj regarding the renaming, I agree, but if I rename it now here, we have to rename it every where, since we have a lot of PR's and branches now, I could imagine that this will be forgotten somewhere. How about we note this down and as soon as everything is in master, we do a global renaming?

dmarcoux:
@krauselukas I'll create a PR for the Workflow model.

vpereira (approved):

PR #11099, opened 2021-05-06
Create Workflows::YAMLDownloader service

dmarcoux:
Taken from #11060 in order to merge changes which can already go on `master`.

**TODO**:
- [x] Specs

PR #11101, opened 2021-05-06
Revert "Retrieve the workflows conf file from target branch"

danidoni:
Given we're not gonna implement the branch filters now, this PR reverts back #11083 to prevent anyone from using any other workflow configuration file but the one present in the reference branch (main, in our case).

### DIFF
--- a/src/api/app/services/workflows/yaml_downloader.rb
+++ b/src/api/app/services/workflows/yaml_downloader.rb
@@ -24,9 +24,9 @@ module Workflows
     def download_url
       case @scm_payload[:scm]
       when 'github'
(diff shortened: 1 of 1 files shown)


eduardoj (approved):

saraycp (changes requested):
I think we need to bring back the `default_branch` field to the payloads (SCMExtractor). We only have `source_branch` and `target_branch` now.

https://github.com/danidoni/open-build-service/blob/1c393f6aa4c17019f9e823c141eaa83ad44d3ed8/src/api/app/services/trigger_controller_service/scm_extractor.rb#L44

dmarcoux:
I suggest closing this PR and taking the discussion and changes to #11099.

danidoni:
Okay, I'll write a new PR against master when we have the pieces in place there

PR #11107, opened 2021-05-06
Create spec for ReportToScmJob

vpereira:

-----------------

### DIFF
--- a/src/api/app/jobs/report_to_scm_job.rb
+++ b/src/api/app/jobs/report_to_scm_job.rb
@@ -1,19 +1,22 @@
-class ReportToSCMJob < ApplicationJob
-  queue_as :scm # TODO: Is this fine?
+class ReportToScmJob < CreateJob
+  # We don't properly capitalize SCM in the class name since CreateJob is doing `CLASS_NAME.to_s.camelize.safe_constantize`
+  queue_as :scm
 
-  def perform
-    Event::Base.where(eventtype: ['Event::BuildFail', 'Event::BuildSuccess'], mails_sent: false).order(created_at: :asc).limit(1000).each do |event|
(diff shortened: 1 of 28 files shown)


src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -8,7 +8,7 @@ def perform(event_id)
 
     EventSubscription.joins('INNER JOIN events ON event_subscriptions.eventtype = events.eventtype AND event_subscriptions.package_id = events.package_id')
                      .where(events: { eventtype: ['Event::BuildFail', 'Event::BuildSuccess'], id: event_id }, channel: :scm)
-                     .where(Event.arel_table[:undone_jobs].gt(0))
+                     .where(Event::Base.arel_table[:undone_jobs].gt(0))
```
dmarcoux, line 11: We always forget the `::Base`, but somehow it worked before because we tested that exact line in the Rails console. Anyway, I'm fine with that change.

vpereira:
```
[1] pry(main)> Event.respond_to?(:arel_table)
=> false
[2] pry(main)> Event::Base.respond_to?(:arel_table)
=> true
[3] pry(main)> 
```
but yes, I remember that we tested it together.. who knows

vpereira:
I don't understand this pr @dmarcoux. It should just be adding specs to the `ReportToScmJob` and now we have 30 commits, tests are broken everywhere. Could you explain me what you did here?

dmarcoux:
I pushed the workflow_token branch and it affects this PR since the workflow_token branch is the target branch. Rebasing this PR on workflow_token should solve the weird history.

PR #11108, opened 2021-05-06
Create Token::Workflow model

dmarcoux:

I added every person who did changes to the policy as commit co-authors.


dmarcoux:
While I don't believe specs are necessary for the model since it's using classes which are tested, we do need to address the various TODO comments.

saraycp (approved):

saraycp:
This model will need some love soonish. But for now is ok.

PR #11110, opened 2021-05-06
Create Workflow::Step::BranchPackageStep model

dmarcoux:

**TODO**:
- [x] Rebase on `master` once #11109 is merged.

### DIFF
--- a/src/api/app/models/workflow/step/branch_package_step.rb
+++ b/src/api/app/models/workflow/step/branch_package_step.rb
@@ -2,11 +2,131 @@ class Workflow
   module Step
     class BranchPackageStep
       include ActiveModel::Model
-      # TODO: Placeholder to be able to test the Workflow model. The code for this class will be added in another PR.
-      def initialize(step_instructions:, scm_extractor_payload:); end
+
+      attr_reader :errors
+
+      validates :source_project, :source_package, presence: true
+
+      def initialize(step_instructions:, scm_extractor_payload:)
+        @step_instructions = step_instructions.with_indifferent_access
+        @scm_extractor_payload = scm_extractor_payload.with_indifferent_access
+        @errors = []
+      end
 
       def allowed_event_and_action?
-        true
+        new_pull_request?
+      end
+
+      def call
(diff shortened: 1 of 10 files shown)


src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -0,0 +1,138 @@
...
+
+      def source_project
+        @step_instructions['source_project']
+      end
+
+      def source_package
+        @step_instructions['source_package']
+      end
+
+      def destination_package
```
hennevogel, line 47: Can we please use the same names? source & target. This is confusing.

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -0,0 +1,138 @@
...
+        return if remote_source?
+
+        options = { use_source: false, follow_project_links: true, follow_multibuild: true }
+        src_package = Package.get_by_project_and_name(source_project, source_package, options)
+
+        raise Pundit::NotAuthorizedError unless PackagePolicy.new(User.session, src_package).create_branch?
+      end
+
+      def branch
+        check_source_access
```
hennevogel, line 73: Why only source? What about the destination? Or do you want to rely on the checks inside `BranchPackage`?

hennevogel (commented):
All those `if gitlab || github` sound to me as if you want to abstract this away into some model...

src/api/app/models/workflow/step/branch_package_step.rb
krauselukas: Yes the `BranchPackage` model checks for the destination, so the idea was to rely on them, since that is the same way its done in the controller endpoints for package branching.

saraycp:
This PR is updated after rebasing on master and replace `destination` by `target`.

saraycp:
@danidoni I think you re-introduced old code that is no longer needed. I guess you didn't add your spec on top of the updated code but force-pushed the spec on top of some old code.

danidoni (approved):

PR #11112, opened 2021-05-06
Recreate the foreign key after recreating the index

hennevogel:
We ran into some mariadb problem when not doing this.

```
SHOW ENGINE INNODB STATUS;
...
------------------------
LATEST FOREIGN KEY ERROR
------------------------
...
DELETE FROM `repositories` WHERE `repositories`.`id` = 3229851
Foreign key constraint fails for table `obs_production`.`repositories`:
...
CONSTRAINT `path_elements_ibfk_1` FOREIGN KEY (`parent_id`) REFERENCES `repositories` (`id`)
...
But the referencing table `obs_production`.`path_elements`
or its .ibd file or the required index does not currently exist!
```

### DIFF
--- a/src/api/db/migrate/20210505160725_add_kind_to_path_element_uniq_indexes.rb
+++ b/src/api/db/migrate/20210505160725_add_kind_to_path_element_uniq_indexes.rb
@@ -1,17 +1,15 @@
 class AddKindToPathElementUniqIndexes < ActiveRecord::Migration[6.0]
   def up
-    safety_assured { execute 'SET FOREIGN_KEY_CHECKS=0;' }
+    remove_foreign_key :path_elements, name: 'path_elements_ibfk_1'
     remove_index :path_elements, name: 'parent_repository_index'
     add_index :path_elements, ['parent_id', 'repository_id', 'kind'], name: 'parent_repository_index', unique: true
-  ensure
-    safety_assured { execute 'SET FOREIGN_KEY_CHECKS=1;' }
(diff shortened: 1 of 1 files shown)


adrianschroeter:
it is good to remove them when possible, but it should have actually not caused a broken db. so it might still happen...

hennevogel:
What is causing the broken DB is recreating the index while not checking foreign keys. At least that is what we can reproduce right?

adrianschroeter:
Not sure about that, it would not explain the breakages in other tables we had.

Btw, the real problem behind was the reported in-existing index (which was reported as existing) in the innodb engine. This error above was just the consequence of it.

Yes, I am all for merging this, but I am not sure that it really solves the problem. Even it might hide it atm.

andrii-suse:
It looks that this bug report suggest that corruption may happen when doing restarting  FK modifications with FOREIGN_KEY_CHECKS=0  https://jira.mariadb.org/browse/MDEV-8845 . In my understanding to prevent corruption - they recommend to make sure that the table is in consistent state after adding foreign key.

eduardoj:
I did some testing with a local development environment:
- I could reproduce the foreign key error with this steps:
  - Check out master with a commit previous to https://github.com/openSUSE/open-build-service/pull/11074/commits/6d6c58d020c14b6617e999b500e5963bb20c31b5
  - Create some repositories from the UI
  - Check out master
  - `rake db:migrate`
  - delete a repository results in an error
- In this scenario, where deleting repositories throw an error, applying the changes of this PR fixed it:
  - check out the branch of this pull request
  - `rake db:migrate:redo`
  - delete a repository works without any error

eduardoj (approved):

PR #11113, opened 2021-05-06
Disable Rails/BulkChangeTable

hennevogel:
This encourages change_table that isn't inspectable by strong_migrations
and threfore less safe. I rather have one too many SQL command executed than
one too many unsafe migration.

### DIFF
--- a/src/api/.rubocop.yml
+++ b/src/api/.rubocop.yml
@@ -278,26 +278,10 @@ Rails/Output:
 Rails/UniqueValidationWithoutIndex:
   Enabled: false
 
-# We decided not to modify old migrations
+# This encourages change_table that isn't inspectable by strong_migrations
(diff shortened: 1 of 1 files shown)


src/api/.rubocop.yml
```diff
@@ -278,26 +278,10 @@ Rails/Output:
 Rails/UniqueValidationWithoutIndex:
   Enabled: false
 
-# We decided not to modify old migrations
+# This encourages change_table that isn't inspectable by strong_migrations
+# and threfor less safe
```
dmarcoux, line 282:
```suggestion
# and therefore, less safe
```

danidoni (approved):

hennevogel:
codeclimate has problems currently. merging.

PR #11119, opened 2021-05-10
[api] fix crash in token trigger

adrianschroeter:
Happened for rebuild of multibuild flavor where source is from linked
project.

-----------------

### DIFF
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -91,9 +91,9 @@ class TriggerController < ApplicationController
     # @package is a String if @project has a project link, no need to do anything then.
     return unless @package.is_a?(Package)
 
-    # We use the package parameter if it is a valid multibuild flavor of the package
-    # See https://github.com/openSUSE/open-build-service/wiki/Links#mulitbuild-packages
-    @package = params[:package] if @package.multibuild_flavor?(params[:package])
+    # We use the package parameter if it is a multibuild flavor of a package
+    # The package object may not exist (different project, maybe even remote)
+    @package = params[:package] if params[:package].present? && params[:package].include?(':')
   end
 
   def package_from_project_link?


src/api/app/controllers/trigger_controller.rb
```diff
@@ -91,9 +91,9 @@ def set_multibuild_flavor
     # @package is a String if @project has a project link, no need to do anything then.
     return unless @package.is_a?(Package)
 
-    # We use the package parameter if it is a valid multibuild flavor of the package
-    # See https://github.com/openSUSE/open-build-service/wiki/Links#mulitbuild-packages
-    @package = params[:package] if @package.multibuild_flavor?(params[:package])
+    # We use the package parameter if it is a multibuild flavor of a package
+    # The package object may not exist (different project, maybe even remote)
+    @package = params[:package] if params[:package].include?(':')
```
hennevogel, line 96: At this point it must be, see line 92...

adrianschroeter: it can't be if it is remote (and see line 60 which is broken anyway, since you do not know if the packages froms from remote, even when there is a project link to remote)

but the code seems also not to handle local instances correctly, but this way it works in general and does not need plenty if special cases.

adrianschroeter: and you can not call multibuild_flavor? on a string neither.

hennevogel:
This is just #11091 missing

https://build.opensuse.org/package/view_file/openSUSE:Factory/casacore/_multibuild?expand=1

adrianschroeter:
no, that code is broken anyway, since it tries to operate with multibuild_flavor? also on strings

src/api/app/controllers/trigger_controller.rb
hennevogel: How can @package be a string after line 92?

hennevogel:
> no, that code is broken anyway, since it tries to operate with multibuild_flavor? also on strings

No matter how often you repeat this, it won't be true.

Line 92: `return unless @package.is_a?(Package)`

I guess the exception confused you? It was calling `multibuild_flavor?` on nil because in the case of *just* `<package>` elements as "flavors" (like in the case of casacore) `multibuild_flavors` returned `[nil]`. This is solved in #11091

src/api/app/controllers/trigger_controller.rb
adrianschroeter: @package ||= params[:package]
  is setting it to a string, no?

hennevogel: Only if `Package.get_by_project_and_name` above did not set it to a package object (there is a package in the DB in @project or in a project-link of @project) or bail out with an exception (in case there is no local package and we don't follow project links for this token).

hennevogel: And not matter if some line above sets it to a string. If `return unless...` on line 92 will return unless it's a `Package` object...

adrianschroeter:
A way better way to implement this would be to hand over the token model the lookup and get

a) The source object which you need to verify for read permission
b) The source/build object which you need to check for maintainer rights
c) And set the build object which you need to work on.

all of them can be different and it depends on the actual operation. But this seperation is not visible anymore in the controller where the logic is implemented atm.

adrianschroeter:
On Dienstag, 11. Mai 2021, 11:36:30 CEST Henne Vogelsang wrote:
> > okay, but the multibuild flavor are irrelevant for permission checking.
> 
> Sure, that is why we don't use them for permission checking...
> 
> > All what counts are the objects which needs to be requested for reading (with or without source access)
> and the ones you operate on.
> 
> Exactly. That is why we find the objects to operate on in `set_token`, `set_project` and `set_package`. And then find the objects to authorize in `set_object_to_authorize`. Lastly we take into account the multibuild names with `set_multibuild_flavor`. 

and again, there is no need to bind this to an object. Just hand it to the backend and be good.

atm this caused regressions because the parsing is not covering all situations as mentioned before.

And for permission checking (the only thing what the api is responsible for) needs to happen anyway
independ of the flavor name.

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> and again, there is no need to bind this to an object. Just hand it to the backend and be good.

We make things explicit instead of implicit so people working on the API can actually understand what is going on. It doesn't cost us much but brings great benefits. 

> atm this caused regressions because the parsing is not covering all situations as mentioned before.

Which is already solved in PR #11091. We are going in circles...

adrianschroeter:
this got solved meanwhile

PR #11124, opened 2021-05-11
Create job to report build status to SCM via SCMStatusReporter

vpereira:
Extracted from #11060

-----------------

### DIFF
--- a/src/api/app/jobs/report_to_scm_job.rb
+++ b/src/api/app/jobs/report_to_scm_job.rb
@@ -1,0 +1,27 @@
+# We don't properly capitalize SCM in the class name since CreateJob is doing `CLASS_NAME.to_s.camelize.safe_constantize`
+class ReportToScmJob < CreateJob
+  ALLOWED_EVENTS = ['Event::BuildFail', 'Event::BuildSuccess'].freeze
+
+  queue_as :scm
+
+  def perform(event_id)
+    event = Event::Base.find(event_id)
+    return false unless event.undone_jobs.positive?
+
+    event_type = event.eventtype
+    return false unless ALLOWED_EVENTS.include?(event_type)
+
+    event_package = Package.find_by_project_and_name(event.payload['project'], event.payload['package'])
+    return false if event_package.blank?
+
+    EventSubscriptionsFinder.new
+                            .for_scm_channel_with_token(event_type: event_type, event_package: event_package)
+                            .each do |event_subscription|
+      SCMStatusReporter.new(event_subscription.payload,
+                            event_subscription.token.scm_token,
+                            event_subscription.eventtype).call
+    end
+
+    true
+  end
+end
--- a/src/api/app/models/event/base.rb
+++ b/src/api/app/models/event/base.rb
@@ -106,7 +106,7 @@ module Event
     end
 
     def initialize(attribs)
-      attributes = attribs.dup
+      attributes = attribs.dup.with_indifferent_access
       super()
       self.created_at = attribs[:time] if attributes[:time]
       attributes.delete :eventtype
@@ -168,6 +168,13 @@ module Event
         self.undone_jobs += 1
       end
       save if self.undone_jobs.positive?
+    end
+
+    def mark_job_done!
+      return unless undone_jobs.positive?
+
+      self.undone_jobs -= 1
+      save!
     end
 
     # to be overwritten in subclasses
--- a/src/api/spec/jobs/report_to_scm_job_spec.rb
+++ b/src/api/spec/jobs/report_to_scm_job_spec.rb
@@ -1,0 +1,102 @@
+require 'rails_helper'
+
+RSpec.describe ReportToScmJob, vcr: false do
+  let(:user) { create(:confirmed_user, login: 'foolano') }
+  let(:token) { Token::Workflow.create(user: user, scm_token: 'fake_token') }
+  let(:project) { create(:project, name: 'project_1', maintainer: user) }
+  let(:package) { create(:package, name: 'package_1', project: project) }
+  let(:repository) { create(:repository, name: 'repository_1', project: project) }
+  let(:event) { Event::BuildSuccess.create({ project: project.name, package: package.name, repository: repository.name, reason: 'foo' }) }
+  let(:event_subscription) do
+    EventSubscription.create(token: token,
+                             user: user,
+                             package: package,
+                             receiver_role: 'reader',
+                             payload: { scm: 'github' },
+                             eventtype: 'Event::BuildSuccess',
+                             channel: :scm)
+  end
+
+  shared_examples 'not reporting to the SCM' do
+    it 'job return value is false' do
+      expect(subject).to be_falsey
+    end
+
+    it 'does not call the scm reporter' do
+      expect_any_instance_of(SCMStatusReporter).not_to receive(:call) # rubocop:disable RSpec/AnyInstance
+      subject
+    end
+  end
+
+  describe '#perform' do
+    subject { described_class.perform_now(event.id) }
+
+    context 'happy path' do
+      before do
+        event
+        event_subscription
+      end
+
+      it 'job return value is true' do
+        allow_any_instance_of(Octokit::Client).to receive(:create_status) # rubocop:disable RSpec/AnyInstance
+        expect(subject).to be_truthy
+      end
+
+      it 'does call the scm reporter' do
+        allow_any_instance_of(Octokit::Client).to receive(:create_status) # rubocop:disable RSpec/AnyInstance
+        expect_any_instance_of(SCMStatusReporter).to receive(:call).once # rubocop:disable RSpec/AnyInstance
+        subject
+      end
+    end
+
+    context 'when using a non-allowed event' do
+      let(:event) do
+        Event::Commit.create(project: project.name, package: package.name)
+      end
+
+      before do
+        event
+        event_subscription
+      end
+
+      it_behaves_like 'not reporting to the SCM'
+    end
+
+    context 'when the event is for some other project than the subscribed one' do
+      let(:event) { Event::BuildSuccess.create(project: 'some:other:project', package: package.name, repository: repository.name, reason: 'foo') }
+
+      before do
+        event
+        event_subscription
+      end
+
+      it_behaves_like 'not reporting to the SCM'
+    end
+
+    context 'when the event is for some other package than the subscribed one' do
+      let(:event) { Event::BuildSuccess.create(project: project.name, package: 'some_other_package', repository: repository.name, reason: 'foo') }
+
+      before do
+        event
+        event_subscription
+      end
+
+      it_behaves_like 'not reporting to the SCM'
+    end
+
+    context 'when the reporting raises an error' do
+      let(:event) { Event::BuildSuccess.create(project: project.name, package: package.name, repository: repository.name, reason: 'foo') }
+
+      before do
+        allow_any_instance_of(Octokit::Client).to receive(:create_status).and_raise(StandardError, '42') # rubocop:disable RSpec/AnyInstance
+        event
+        event_subscription
+      end
+
+      it 'does not call the scm reporter' do
+        expect_any_instance_of(SCMStatusReporter).to receive(:call).once # rubocop:disable RSpec/AnyInstance
+        subject
+      end
+    end
+  end
+end
--- a/src/api/app/jobs/create_job.rb
+++ b/src/api/app/jobs/create_job.rb
@@ -9,8 +9,7 @@ class CreateJob < ApplicationJob
 
     # in test suite the undone_jobs are 0 as the delayed jobs are not delayed
     event.with_lock do
-      event.undone_jobs -= 1
-      event.save!
+      event.mark_job_done!
     end
   end
 
--- a/src/api/app/models/event/build_fail.rb
+++ b/src/api/app/models/event/build_fail.rb
@@ -6,6 +6,8 @@ module Event
     self.description = 'Package has failed to build'
     receiver_roles :maintainer, :bugowner, :reader, :watcher
 
+    create_jobs :report_to_scm_job
+
     def subject
       "Build failure of #{payload['project']}/#{payload['package']} in #{payload['repository']}/#{payload['arch']}"
     end
--- a/src/api/app/models/event/build_success.rb
+++ b/src/api/app/models/event/build_success.rb
@@ -3,6 +3,8 @@ module Event
     self.message_bus_routing_key = 'package.build_success'
     self.description = 'Package has succeeded building'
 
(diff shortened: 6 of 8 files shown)


src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -0,0 +1,25 @@
+class ReportToScmJob < CreateJob
+  # We don't properly capitalize SCM in the class name since CreateJob is doing `CLASS_NAME.to_s.camelize.safe_constantize`
+  queue_as :scm
+
+  def perform(event_id)
+    event = Event::Base.find(event_id)
+    return true unless event
+
+    EventSubscription.joins('INNER JOIN events ON event_subscriptions.eventtype = events.eventtype AND event_subscriptions.package_id = events.package_id')
```
vpereira, line 9: We should extract it to a query object

src/api/app/models/event/base.rb
```diff
@@ -105,9 +107,10 @@ def receiver_roles
     end
 
     def initialize(attribs)
-      attributes = attribs.dup
+      attributes = attribs.dup.with_indifferent_access
       super()
       self.created_at = attribs[:time] if attributes[:time]
+      self.package = Package.get_by_project_and_name(attributes[:project], attributes[:package]) if attributes[:project] && attributes[:package]
```
vpereira, line 113: We should work this robuster, more bullet proof, in case project and/or package are remote. See https://github.com/openSUSE/open-build-service/wiki/Links

src/api/spec/models/update_notification_events_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 RSpec.describe UpdateNotificationEvents, type: :model do
   let!(:admin) { create(:admin_user, login: 'Admin') }
+  let!(:project_with_package) { create(:project_with_package, name: 'project_1', package_name: 'multibuild') }
```
vpereira, line 5: after creating a relationship with a package on `Event::Base`, passing nonexistent package and project could be a problem

src/api/app/models/event/base.rb
danidoni: Fixed

src/api/app/jobs/report_to_scm_job.rb
danidoni: Fixed

src/api/spec/models/update_notification_events_spec.rb
danidoni: We are gonna remove the relationship between `Event` and `Package` so this won't be a problem anymore.

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -0,0 +1,33 @@
+class ReportToScmJob < CreateJob
+  ALLOWED_EVENTS = ['Event::BuildFail', 'Event::BuildSuccess'].freeze
+
+  # We don't properly capitalize SCM in the class name since CreateJob is doing `CLASS_NAME.to_s.camelize.safe_constantize`
+  queue_as :scm
+
+  def perform(event_id)
+    event = Event::Base.find(event_id)
+    return true unless event
```
krauselukas, line 9: hmmm, should it return `false` here?

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    return false unless event.undone_jobs.positive?
+
+    event_type = event.eventtype
+    return false unless ALLOWED_EVENTS.include?(event_type)
+
+    # TODO: search for the package properly
```
krauselukas, line 16: The comment is a left over

danidoni: I'll remove it

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    EventSubscriptionsFinder.new
+                            .for_scm_channel_with_token(event_type: event_type, event_package: event_package)
+                            .each do |event_subscription|
+      SCMStatusReporter.new(event_subscription.payload,
+                            event_subscription.token.scm_token,
+                            event_subscription.eventtype).call
+    rescue StandardError => e
+      Airbrake.notify(e, event_id: event.id)
+      event.mark_job_done!
```
krauselukas, line 27: @danidoni you renamed the method in the `Event::Base`, you need to adapt it here

krauselukas (changes requested):

src/api/app/jobs/report_to_scm_job.rb
krauselukas: also to me just the method name `done!` is a little bit confusing, what is done? the job, the event?

danidoni: Yes, you are right. I've left the long name in both places

src/api/app/models/event/base.rb
```diff
@@ -7,6 +7,8 @@ class Base < ApplicationRecord
 
     after_create :create_project_log_entry_job, if: -> { (PROJECT_CLASSES | PACKAGE_CLASSES).include?(self.class.name) }
 
+    belongs_to :package
```
dmarcoux, line 10: This association should be removed if we aren't using it anymore.

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -0,0 +1,32 @@
+class ReportToScmJob < CreateJob
+  ALLOWED_EVENTS = ['Event::BuildFail', 'Event::BuildSuccess'].freeze
+
+  # We don't properly capitalize SCM in the class name since CreateJob is doing `CLASS_NAME.to_s.camelize.safe_constantize`
+  queue_as :scm
+
+  def perform(event_id)
+    event = Event::Base.find(event_id)
+    return false unless event
```
hennevogel, line 9: It will never come this far. 

```ruby
Event::Base.find(12345)
ActiveRecord::RecordNotFound: Couldn't find Event::Base with 'id'=12345
```

src/api/app/jobs/report_to_scm_job.rb
hennevogel, line 27: Why mark it was done in case of error?

hennevogel (changes requested):

src/api/app/models/event/base.rb
```diff
@@ -169,6 +169,13 @@ def perform_create_jobs
       save if self.undone_jobs.positive?
     end
 
+    def mark_job_done!
```
dmarcoux, line 172: This method is really similar to this code in `CreateJob`:

https://github.com/openSUSE/open-build-service/blob/1379e7286ef419c56bc4f6f71a95bbdda79d2400/src/api/app/jobs/create_job.rb#L10-L14

I think this could be aligned.

src/api/app/jobs/report_to_scm_job.rb
danidoni: True, fixed

src/api/app/jobs/report_to_scm_job.rb
dmarcoux: If somehow the job crashes due to an error, the `event.undone_jobs` field will never be changed in the `after_perform` callback from the parent class `CreateJob` (see code snippet below). A `ReportToScmJob` job would then run again for the event and keep on crashing.

https://github.com/openSUSE/open-build-service/blob/1379e7286ef419c56bc4f6f71a95bbdda79d2400/src/api/app/jobs/create_job.rb#L6-L15

hennevogel: Yes, and maybe it would run again an not crash?

src/api/app/models/event/base.rb
danidoni: [Done](https://github.com/openSUSE/open-build-service/pull/11124/files#diff-f2926e4c685d321e3d5c4d684167eab3511f970c7aa0b3aced8406eb6123de66R12)

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -0,0 +1,27 @@
+class ReportToScmJob < CreateJob
+  ALLOWED_EVENTS = ['Event::BuildFail', 'Event::BuildSuccess'].freeze
+
+  # We don't properly capitalize SCM in the class name since CreateJob is doing `CLASS_NAME.to_s.camelize.safe_constantize`
```
dmarcoux, line 4: This comment should be at the top of this file.

src/api/spec/jobs/report_to_scm_job_spec.rb
```diff
@@ -0,0 +1,102 @@
+require 'rails_helper'
+
+RSpec.describe ReportToScmJob, vcr: false do
+  let(:user) { create(:confirmed_user, login: 'foolano') }
+  let(:token) { Token::Workflow.create(user: user, scm_token: 'fake_token') }
+  let(:project) { create(:project, name: 'project_1', maintainer: user) }
+  let(:package) { create(:package, name: 'package_1', project: project) }
+  let(:repository) { create(:repository, name: 'repository_1', project: project) }
+  let(:event) { Event::BuildSuccess.create({ project: project.name, package: package.name, repository: repository.name, reason: 'foo' }) }
```
danidoni, line 9: This event creation triggers a `ReportToScmJob` run because [this after_create](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/event/base.rb#L166) and [our current](https://github.com/openSUSE/open-build-service/blob/master/src/api/config/environments/test.rb#L63) active job configuration.

src/api/spec/jobs/report_to_scm_job_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+                             channel: :scm)
+  end
+
+  shared_examples 'not reporting to the SCM' do
+    it 'job return value is false' do
+      expect(subject).to be_falsey
+    end
+
+    it 'does not call the scm reporter' do
+      expect_any_instance_of(SCMStatusReporter).not_to receive(:call) # rubocop:disable RSpec/AnyInstance
```
danidoni, line 26: This expectation can be refactored later in another PR using a custom status reporter passed as parameter to the `ReportToScmJob`.

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -0,0 +1,27 @@
+# We don't properly capitalize SCM in the class name since CreateJob is doing `CLASS_NAME.to_s.camelize.safe_constantize`
+class ReportToScmJob < CreateJob
+  ALLOWED_EVENTS = ['Event::BuildFail', 'Event::BuildSuccess'].freeze
+
+  queue_as :scm
+
+  def perform(event_id)
+    event = Event::Base.find(event_id)
```
dmarcoux, line 8: This previous comment from @hennevogel still stands:

> It will never come this far. 
> 
> ```ruby
> Event::Base.find(12345)
> ActiveRecord::RecordNotFound: Couldn't find Event::Base with 'id'=12345
> ```

I suggest changing the line to this:
```suggestion
    event = Event::Base.find_by(id: event_id)
```

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -0,0 +1,27 @@
...
+    return false unless event.undone_jobs.positive?
```
dmarcoux, line 9: In combination with the comment above changing the finder method.

```suggestion
    return false unless event && event.undone_jobs.positive?
```

src/api/app/jobs/report_to_scm_job.rb
danidoni: True

danidoni: Hm, I've been checking this and we'll be hitting [this after_perform](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/jobs/create_job.rb#L8) if we pass an id from a non-existent event. Thinking about alternatives, it's better to let it raise an error and detect this error in Errbit, so we can fix it outside this class

src/api/app/jobs/report_to_scm_job.rb
danidoni: Following previous comment logic, this change is not needed

src/api/app/jobs/report_to_scm_job.rb
dmarcoux: Yes, since there isn't a way around it without affecting all jobs inheriting from `CreateJob`, we'll leave it as it is. This is also an edge case. The event should always be in the database.

src/api/app/models/event_subscription.rb
```diff
@@ -26,6 +26,7 @@ class EventSubscription < ApplicationRecord
   belongs_to :user, inverse_of: :event_subscriptions
   belongs_to :group, inverse_of: :event_subscriptions
   belongs_to :token, inverse_of: :event_subscriptions
+  belongs_to :package
```
krauselukas, line 29: @danidoni this line is already in master, I would drop the commit that introduces this from your PR and rebase once more against master

krauselukas:
@danidoni I would clean up a bit the commit history, the third commit in this PR (Extend Event classes to track a package) does not really do what it describes anymore. Besided that it looks good to me

src/api/spec/jobs/report_to_scm_job_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe ReportToScmJob, vcr: false do
```
saraycp, line 3: Is `vcr` set to `false` on purpose? Could it just be removed?

src/api/app/models/event_subscription.rb
saraycp: exactly, I had to add it for the specs to work.

danidoni: Okay

danidoni:
> @danidoni I would clean up a bit the commit history, the third commit in this PR (Extend Event classes to track a package) does not really do what it describes anymore. Besided that it looks good to me

I will squash them all into just one commit.

src/api/spec/jobs/report_to_scm_job_spec.rb
danidoni: Enabling this just records calls to backend that are not related to the spec. So I think we don't need to enable this.

saraycp: so no need to add the flag, right?

krauselukas (approved):

src/api/spec/jobs/report_to_scm_job_spec.rb
danidoni: I'm working on an improved spec, I will take care of that there

PR #11130, opened 2021-05-12
Create the data_migrations table unless it exists

hennevogel:
This table is usually created once you run some data_migrate rake task. Let's be more explicit about this.


dmarcoux (approved):
This is matching what is defined upstream :+1:

https://github.com/ilyakatz/data-migrate/blob/6890921c08a702fb4b0989e402ac3a9144a0996f/lib/data_migrate/data_migrator.rb#L77-L81

PR #11133, opened 2021-05-14
Apply the background color all the way to the right in request diffs

### LINKED ISSUE #8172: webui: OBS request diff background color not always full width
# How to reproduce
https://build.opensuse.org/request/show/722568
The textbox for sev-tool.spec should show a horizontal scrollbar (if not, resize your browser window to a smaller size get the effect).
On this page, use the browser's zoom feature (Ctrl+KP_Plus) twice or more.
Scroll horizontally to the right.

# Observed behavior
The green background for the sev-tool.spec textbox now no longer extends to the right edge of the text:

col

# Expected Result
Apply the green background all the way to the right if possible even if zoomed.

eduardoj:
Fixes #8172.

-----------------

### DIFF
--- a/src/api/app/assets/stylesheets/webui/coderay.scss
+++ b/src/api/app/assets/stylesheets/webui/coderay.scss
@@ -2,13 +2,17 @@
   background: $card-bg;
   color: $body-color;
 }
-.CodeRay pre { margin: 0 }
+.CodeRay pre {
+  margin: 0;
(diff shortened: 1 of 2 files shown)


dmarcoux (approved):
I tested this with the following request: https://obs-reviewlab.opensuse.org/eduardoj-fix_8172/request/show/13

It works! :+1:

PR #11164, opened 2021-05-20
Replace `require_login` with Pundit in Webui::Staging::ProjectsController

intrip:
This is a PR of a series which replaces require_login with Pundit.

Tackles `Webui::Staging::ProjectsController`

Ref #10083

-----------------


intrip:
@dmarcoux could you take a peek? 🙇

hennevogel (approved):

hennevogel:
Thanks @intrip :bouquet:

PR #11177, opened 2021-05-23
[backend] Add buildhook feature

### LINKED ISSUE #11027: Feature for call custom code before and after build script
Hi folks! I think this should be useful:

**Problem**
Some times need execute custom code before or after executing build script.

Cases in my practice:
1. Uploading results of static code analyze to the analyzer server
2. Calculating special checksum of sources and binaries at build time using custom program

**Solution**

1. Add new API endpoint in bs_repserver for uploading buildhook-scripts into workers.
   API endpoint should be similar to `getworkercode` and `getbuildcode` 
2. Add step for downloading buildhook scripts in bs_worker
3. Add buildhook call before and after build script execution
4. Add buildhook configuration. Can be similar to publisherhook configuration. For example - content of BSConfig.pm:
   ```
   our $buildhookpath = "/some/path/buildhook";
   our $afterbuildhook = { ".*" => "test.sh" };
   our $beforebuildhook = { ".*" => "test.sh" };
   ```
   Content of `/some/path/buildhook/test.sh` can be bash script or other

kvelichko-omp:
This PR should close #11027 

For test in local environment:
1. Create path buildhook:
   ```
   mkdir buildhook
   ```
2. Create simple buildhook script `test.sh`:
   ```bash
   #!/bin/bash
   
   BUILDROOT=$1
   PROJID=$2
   PACKID=$3
   REPOID=$4
   ARCH=$5
   MODE=$6
   
   exec > >(tee -a "$BUILDROOT/.build.log") 2>&1
   
   if [ "$MODE" = "before" ]; then
     echo "Before hook executing" > $BUILDROOT/.hook.log
   else
     echo "Before hook result:"
     echo "-------------------"
     cat $BUILDROOT/.hook.log
     echo "-------------------"
     echo "After hook executing"
   fi
   ```
3. Copy file `src/backend/BSConfig.pm.template` to `src/backend/BSConfig.pm` and add next lines:
   ```
   our $buildhookpath = "/obs/buildhook";
   our $beforebuildhook = {".*" => "test.sh"};
   our $afterbuildhook = {".*" => "test.sh"};
   ```
4. Start OBS
   ```
   docker-compose up
   ```
5. Add package and build it
6. See build log

** After **
Screenshot from 2021-05-23 16-07-33

### DIFF
--- a/src/backend/bs_repserver
+++ b/src/backend/bs_repserver
@@ -3479,6 +3479,15 @@ sub getworkercode {
   return getcode($cgi, 'worker');
 }
 
+sub getbuildhook {
+  my ($cgi) = @_;
+  if ($BSConfig::buildhookpath) {
+    return getcode($cgi, $BSConfig::buildhookpath);
+  }else{
+    return undef
+  }
+}
+
 sub postrepo {
   my ($cgi, $projid, $repoid, $arch) = @_;
 
@@ -4366,6 +4375,7 @@ my $dispatches = [
   '!worker /worker $arch $port $state: workerid? working:bool? memory:num? disk:num? buildarch:arch* tellnojob:bool?' => \&workerstate,
   '!worker /getbuildcode' => \&getbuildcode,
   '!worker /getworkercode' => \&getworkercode,
+  '!worker /getbuildhook' => \&getbuildhook,
   '!worker POST:/putjob $arch $job $jobid $code:? now:num? kiwitree:bool? workerid?' => \&putjob,
   '!worker POST:/workerdispatched $arch $job $jobid hostarch:arch port workerid?' => \&workerdispatched,
   '!worker /getbinaries $project $repository $arch binaries: nometa:bool? metaonly:bool? workerid? now:num? module*' => \&getbinaries,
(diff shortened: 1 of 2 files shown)


adrianschroeter:
The hook would not be available for local or remote builds then. So anything, what is important for the build result would not be possible in such scripts.

So I wonder what is the use-case for this at all?

(And the mentioned API interface would actually a root security hole).

kvelichko-omp:
> The hook would not be available for local or remote builds then. So anything, what is important for the build result would not be possible in such scripts.

Yes it is

>So I wonder what is the use-case for this at all?

As I wrote in #10853:

1. Uploading results of static code analyze to the analyzer server
2. Calculating special checksum of sources and binaries at build time using custom program

Additionally, using this hook, I can try to solve the problem with umount  https://github.com/openSUSE/obs-build/pull/630

>(And the mentioned API interface in the issue would actually be a root security hole).

Hook configuration available only for system administrator.

Yes, a bad hook can disable a worker. But this is compensated by the fact that only qualified employees have the ability to set up hooks.

PR #11181, opened 2021-05-25
Remove package id from event table

krauselukas:
After changing the from the attempt of building an association
between an event and a package, to fetching the package
using the passed payload in the `ReportToScmJob`
in #11124
the package_id is not longer needed.

### DIFF
--- a/src/api/app/models/event/base.rb
+++ b/src/api/app/models/event/base.rb
@@ -346,12 +346,10 @@ end
 #  undone_jobs :integer          default(0)
 #  created_at  :datetime         indexed
 #  updated_at  :datetime
-#  package_id  :integer          indexed
 #
 # Indexes
 #
 #  index_events_on_created_at  (created_at)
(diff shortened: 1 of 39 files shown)


krauselukas:
[Remove the package field from Event](https://trello.com/c/gwg8uF61/1665-remove-the-package-field-from-event)

hennevogel (approved):

dmarcoux (changes requested):
Ignoring the column and dropping it in the same PR is going to cause downtime. This has to be done separately:

Steps:
1. Merge first PR to ignore the column
2. Deploy changes from first PR
3. Merge second PR to drop the column
4. Deploy changes from second PR
5. Merge third PR to remove "ignore code" from first PR
6. Deploy changes from third PR

dmarcoux (approved):
Forget my previous review. I was wrong since we deploy changes by installing an RPM package, we're good. This is done in two steps, installing the package, then running the migration.

PR #11184, opened 2021-05-26
[api] fix releaseing multibuild container

adrianschroeter:

-----------------

### DIFF
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -22,7 +22,7 @@ class TriggerController < ApplicationController
   def create
     authorize @token
     @token.user.run_as do
-      opts = {project: @project, package: @package, repository: params[:repository], arch: params[:arch]}
+      opts = { project: @project, package: @package, repository: params[:repository], arch: params[:arch] }
       opts[:multibuild_flavor] = @multibuild_container if @multibuild_container.present?
       @token.call(opts)
       render_ok
@@ -87,7 +87,7 @@ class TriggerController < ApplicationController
 
   def set_multibuild_flavor
     # Do NOT use @package.multibuild_flavor? here because the flavor need to be checked for the right source revision
-    @multibuild_container = params[:package].gsub(/.*:/, '') if params[:package].present? and params[:package].include?(':')
+    @multibuild_container = params[:package].gsub(/.*:/, '') if params[:package].present? && params[:package].include?(':')
   end
 
   def package_from_project_link?
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -22,7 +22,9 @@ class TriggerController < ApplicationController
   def create
     authorize @token
     @token.user.run_as do
-      @token.call(project: @project, package: @package, repository: params[:repository], arch: params[:arch])
+      opts = {project: @project, package: @package, repository: params[:repository], arch: params[:arch]}
+      opts[:multibuild_flavor] = @multibuild_container if @multibuild_container.present?
+      @token.call(opts)
       render_ok
     end
   end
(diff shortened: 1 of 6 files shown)


adrianschroeter:
P2 because appliance qa/release pipeline broke

hennevogel:
This is a feature, it can't be P2. We only ever followed multibuild for rebuild tokens, never for the release token. Please look at the history of `TriggerController`.

- [Apr 26, 2021](https://github.com/openSUSE/open-build-service/blob/a1eb7a46b270271592ac4ee53127184b840ae92c/src/api/app/models/token/release.rb#L24)
- [Oct 13, 2020](https://github.com/openSUSE/open-build-service/blob/87ff5c460f00f0bc8aecd62f644c7a86f6ec19fc/src/api/app/controllers/trigger_controller.rb#L92-L101)
- [Jun 7, 2019](https://github.com/openSUSE/open-build-service/blob/7ee7e72b099a12b8fa970817767ac06fda319339/src/api/app/controllers/trigger_controller.rb#L105-L114)

hennevogel:
Maybe you can tell us exactly what is broken and where?

adrianschroeter:
this now an almost 1 year old topic, most of the time with open pull request fixing it but it was always ignored. It happens internal exactly where I told you when you said that it is not used last week.

When stuff gets refactored it need include the open critical fixes. Esp. when it is used in production and the code contained grave bugs which leads to unwanted data 

however, I had to do another attempt with larger changes now, since the logic of the refactored code made little sense (please read the updated commit message again).

src/api/app/controllers/trigger_controller.rb
```diff
@@ -22,7 +22,9 @@ class TriggerController < ApplicationController
   def create
     authorize @token
     @token.user.run_as do
-      @token.call(project: @project, package: @package, repository: params[:repository], arch: params[:arch])
+      opts = {project: @project, package: @package, repository: params[:repository], arch: params[:arch]}
```
danidoni, line 25:
```suggestion
      opts = { project: @project, package: @package, repository: params[:repository], arch: params[:arch] }
```

src/api/app/controllers/trigger_controller.rb
```diff
@@ -80,34 +82,16 @@ def set_package
...
-    # Only if we consider multibuild or project links we might need to do more complicated things...
-    return unless @token.follow_links?
-    # @package is a String if @project has a project link, no need to do anything then.
-    return unless @package.is_a?(Package)
-
-    # We use the package parameter if it is a valid multibuild flavor of the package
-    # See https://github.com/openSUSE/open-build-service/wiki/Links#mulitbuild-packages
-    @package = params[:package] if @package.multibuild_flavor?(params[:package])
+    # Do NOT use @package.multibuild_flavor? here because the flavor need to be checked for the right source revision
+    @multibuild_container = params[:package].gsub(/.*:/, '') if params[:package].present? and params[:package].include?(':')
```
danidoni, line 90:
```suggestion
    @multibuild_container = params[:package].gsub(/.*:/, '') if params[:package].present? && params[:package].include?(':')
```

danidoni (changes requested):
Please accept the proposed changes to comply with [Rubocop](https://rubocop.org/) guidelines

adrianschroeter:
On Donnerstag, 27. Mai 2021, 18:13:11 CEST Henne Vogelsang wrote:
> As you can see from the spec I've adopted for you, the logic is the same. You **just** moved the `if @package.is_a?(Package)` into `package_from_project_link?`.
> 
> But you again made a lot of the decisions regarding the various type of OBS links implicit. Everybody has to **know** that a remote package is always included via project link. Everybody has to **remember** that `:` in the package parameter means multibuild. 

> Everybody has to **know** `set_multibuild_flavor` will only be executed when the token follows multibuilds/project-links. 

that was the case before and it is wrong and yes complete braindead. It is not the case anymore with my commit.

also you do not need to understand that when you set the params[:package] to a string that it means multibuild flavor,
which complicated the logic a lot for no good reason.

> Everybody has to **know** that packages from remote project links are a special case. And you also removed any trace to the developer documentation where people can learn about those things.

because the code was also removed since not needed and just complicate things and caused bugs.

> So you successfully implemented a **new** feature (follow multibuild for `Token::Release`) and that is cool. But at the same time you made things next to incomprehensible again for future developers...

it is a long standing functionality which was never propper working and broken again in several ways over the last year. Even though
there was always a pull request with fix open.


-- 

Adrian Schroeter ***@***.***>

hennevogel (approved):
I object to this but don't want to sink more time into it. See my comment.

adrianschroeter:
On Donnerstag, 27. Mai 2021, 18:13:11 CEST Henne Vogelsang wrote:
> As you can see from the spec I've adopted for you, the logic is the same. You **just** moved the `if @package.is_a?(Package)` into `package_from_project_link?`.
> 
> But you again made a lot of the decisions regarding the various type of OBS links implicit. Everybody has to **know** that a remote package is always included via project link. Everybody has to **remember** that `:` in the package parameter means multibuild. Everybody has to **know** `set_multibuild_flavor` will only be executed when the token follows multibuilds/project-links. Everybody has to **know** that packages from remote project links are a special case. And you also removed any trace to the developer documentation where people can learn about those things.

feel free to add the documentaion again, but you basically need to add it
everywhere where an object gets looked up and you do some changes (source
or binary operations).

The trigger code is not so much special here.

What could still be done better is that the objects which need to get 
looked up for read access and the ones for write acccess could
handled explicit.

Atm. this is the big thing you need to know when you look at this
code. And that works only atm, because the right lookup of params 
in the token models are luckily happen for the read checks.

But that might not be the case anymore on certain future operations,
where you need to have all the model code in mind when touching the controller.

> So you successfully implemented a **new** feature (follow multibuild for `Token::Release`) and that is cool. But at the same time you made things next to incomprehensible again for future developers...

just for the record, this goes back to

 https://github.com/openSUSE/open-build-service/pull/10388

and we discussed in multiple pull requests and escalations always the same topics.

I had to redo this code multiple times and we had multiple times a breakage
here. So it is not just you who spend a lot of time here.

> But I give up, you seem to want to have it the hard way. The next poor soul extending or refactoring this will run into the giant traps laid out by mulitbuild/project links again. So be it.

I doubt that this poor soul would have understood that strings are a
signal for remote link or multibuild flavor and that the two cases
may need to be handled different in the models.

and there is just no reason to handle project links when setting
the multibuild flavours. These concepts are complete independend.

-- 

Adrian Schroeter ***@***.***>

PR #11188, opened 2021-05-28
[api] fix missing token-rebuild-multibuild case

adrianschroeter:
forgot to add this chunk in 45bb4dd7b8b6c515b161b8cecbf03bac4f8de40e

We need to transfer multibuild container name for rebuild operations

-----------------


hennevogel:
You need help with rspec?

danidoni:
I'll try to address the rspec failures and make this green

danidoni:
Okay, specs fixed. I'll squash the commits

vpereira (approved):

hennevogel (changes requested):
Hm it looks to me as if you just recorded new requests with VCR without getting rid of the old ones.

danidoni:
> Hm it looks to me as if you just recorded new requests with VCR without getting rid of the old ones.

Fixed

vpereira:
@hennevogel your request was addressed, therefore i will merge it.

PR #11193, opened 2021-05-31
List scm_token field in the person/token/index XML view

dmarcoux:


src/api/app/views/person/token/index.xml.builder
```diff
@@ -1,6 +1,12 @@
 xml.directory(count: @list.length) do |dir|
   @list.each do |token|
     p = { id: token.id, string: token.string, kind: token.token_name }
+
+    # TODO: Put this in the hash above once the trigger_workflow feature is completely rolled out
+    if Flipper.enabled?(:trigger_workflow)
+      p[:scm_token] = token.scm_token
```
hennevogel, line 7: Why?

hennevogel (changes requested):

src/api/app/views/person/token/index.xml.builder
dmarcoux: For convenience.

dmarcoux:
We discussed about this and decided to go against this since SCM tokens are displayed once and never again. We should then not displayed them in our API.

PR #11195, opened 2021-05-31
Remove package_id from ignored_columns in event base model

krauselukas:
Ignoring the package_id columns was only needed prior
to a db migration which removed the corresponding column.
After the migration is successfully applied this line can
be removed again.

### DIFF
--- a/src/api/app/models/event/base.rb
+++ b/src/api/app/models/event/base.rb
@@ -4,7 +4,6 @@ module Event
   class Base < ApplicationRecord
     self.inheritance_column = 'eventtype'
     self.table_name = 'events'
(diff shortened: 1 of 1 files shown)


krauselukas:
[Remove "ignored_columns" line from Event::Base once migration is applied to OBS and IBS](https://trello.com/c/3D8wnP82/1668-remove-ignoredcolumns-line-from-eventbase-once-migration-is-applied-to-obs-and-ibs)

vpereira:
lifting the "DO NOT MERGE" tag as it was already deployed on ibs

vpereira (approved):

PR #11206, opened 2021-06-03
add research notice to readme

yochannah:
... so participants are aware this repo is participating in a study adn metrics will be gathered :)

let me know if there are any issues or questions.

-----------------

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,3 +54,12 @@ After finishing the installation of your base system, follow these steps:
 
 If you have a more complex setup (e.g. a distributed backend), we recommend to read the Administration
 chapter in our [user manual](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.admin.html).
+
+### Research notice
+
+Please note that this repository is participating in a study into sustainability of open source projects. Data will be gathered about this repository for approximately the next 12 months, starting from June 2021.
+
+Data collected will include number of contributors, number of PRs, time taken to close/merge these PRs, and issues closed.
+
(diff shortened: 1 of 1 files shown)


dmarcoux:
Sorry, but I don't see the point in having this in the README. I'll wait on a second opinion before taking action on this.

@openSUSE/obs-frontend: What do you think? Do you agree with me?

vpereira:
Looking the description on https://sustainable-open-science-and-software.github.io/, I don't see how could we be in the scope of this research. Could you please provide more information @yochannah ?

hennevogel:
hey @yochannah I don't see the point in adding this. We are a public open source project with [public data on github](https://docs.github.com/en/github/site-policy/github-privacy-statement#public-repositories) that you can request at will.

yochannah:
Hey folks - the scope of this research project covers any open source project - whilst many are science or research projects it's fine for non-research projects to participate too. 

The reason for the readme notification is to try and avoid being a "spooky" researcher who is gathering info without making people aware (github permits open use of github data, but it is still politer to make people aware). 

Would there be a better way to help people know about this than the readme?

dmarcoux:
I would list the open source projects on the research project website. It's much less time consuming for both parties, the open source contributors and you. There's no need to review a PR for contributors and you can directly list projects on the website without having to submit a PR on all projects involved in your research. I find this important because a lot of open source projects are quite overloaded with work... not every project has a team of paid software engineers like the Open Build Service.

PR #11212, opened 2021-06-04
Pass X-Request-ID with each request to the backend

vpereira:
Pass X-Request-ID with each request to the backend
    
Once we also log this in the backend log, this will
make it easier to find all requests that were triggered
by a certain user request.

patch from @darix

### DIFF
--- a/src/api/app/controllers/application_controller.rb
+++ b/src/api/app/controllers/application_controller.rb
@@ -12,6 +12,7 @@ class ApplicationController < ActionController::Base
   include FlipperFeature
 
   include RescueHandler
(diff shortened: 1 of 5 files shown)


darix:
You committed the original patch file too.

vpereira:
> You committed the original patch file too.

yup, because we had to change some bits to make it work, and I wanted to keep the original around. As soon as I remove the "WIP", I will drop it and change the commit message :+1:

hennevogel (approved):

mlschroe:
\o/ nice!

PR #11215, opened 2021-06-07
Handle PR/MR updates in the SCM integration

saraycp:
Every time a Pull Request is updated, for example, the user adds a new commit to the existing PR, GitHub is going to send the event `pull_request` with `synchronize` action. We start handling those events.

As the package is probably already branched, we just need to ensure that:
- the source code gets updated and the build is retriggered, as soon as the `_branch_request` file gets updated;
- ~we cancel the jobs related to the previous update (or the creation of the PR);~
- we updated the payload on the corresponding `EventSubscriptions` since as soon as the build finish, they have to report to the most recent `commit SHA`.


src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -11,18 +11,24 @@ def initialize(step_instructions:, scm_extractor_payload:)
...
       end
 
       def call
-        return unless valid? && new_pull_request?
+        return unless valid? && allowed_event_and_action?
 
-        branched_package = branch
-        return unless branched_package
+        if updated_pull_request? && target_package.exists?
+          # Having _service in the branched package, it will be updated and will rebuild automatically
```
hennevogel, line 21: Hm, who will update it?

dmarcoux:
I think some of the commits from this PR could be split out in smaller PRs. It would be easier to review and merge the changes.

vpereira:
> I think some of the commits from this PR could be split out in smaller PRs. It would be easier to review and merge the changes.

there are 2 commits. Do you want to break them, each one in one PR?

saraycp:
> > I think some of the commits from this PR could be split out in smaller PRs. It would be easier to review and merge the changes.
> 
> there are 2 commits. Do you want to break them, each one in one PR?

@vpereira I squashed the commits, rebased on master, and resolved the conflicts.
I also adapted some code to avoid Code Climate complaints and replaced `User.session` with `@token.user` to make up the `target_project_name`.

src/api/app/services/trigger_controller_service/scm_extractor.rb
```diff
@@ -11,6 +11,8 @@ def initialize(scm, event, payload)
       @event = event
     end
 
+    # TODO: this check seems redundant, but it prevents extracting the payload if the event and action
+    # are not the admited ones. See BranchPackageStep.
```
danidoni, line 15:
```suggestion
    # are not the allowed ones. See BranchPackageStep.
```

danidoni: :thought_balloon:

dmarcoux:
Regarding the conflict in `src/api/app/models/workflow/step/branch_package_step.rb`, this is due to the changes from #11225. 

The methods `source_project`, `source_package`, `target_project` and `target_package` are now public since we depend on them to be able to report back to the SCM.

danidoni (approved):

src/api/app/models/token/workflow.rb
```diff
@@ -29,10 +29,10 @@ def run_step_and_report(step, scm_extractor_payload, scm_token)
 
     set_subscription(package_from_step, scm_extractor_payload)
 
-    Project.get_by_name(step.target_project).repositories.each do |repository|
+    Project.get_by_name(step.target_project_name).repositories.each do |repository|
       # TODO: Fix n+1 queries
       repository.architectures.each do |architecture|
-        SCMStatusReporter.new({ project: step.target_project, package: step.target_package, repository: repository.name, arch: architecture.name },
+        SCMStatusReporter.new({ project: step.target_project_name, package: step.target_package, repository: repository.name, arch: architecture.name },
```
dmarcoux, line 35: One renaming is missing:

```suggestion
        SCMStatusReporter.new({ project: step.target_project_name, package: step.target_package_name, repository: repository.name, arch: architecture.name },
```

dmarcoux (changes requested):
There is also a typo in the second commit message:

Adjust Bra**n**chPackageStep model specs

src/api/app/models/token/workflow.rb
saraycp: done!

dmarcoux (approved):

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -14,43 +14,55 @@ def initialize(step_instructions:, scm_extractor_payload:, token:)
...
       end
 
       private
 
+      def target_package
+        Package.find_by_project_and_name(target_project_name, target_package_name)
+      end
+
+      def create_or_update_branched_package
+        return target_package if updated_pull_request? && target_package.present?
```
hennevogel, line 55: No one is touching the branched package. That means it won't rebuild right?

hennevogel (changes requested):

src/api/app/models/workflow/step/branch_package_step.rb
hennevogel: Also the `branch_request_file` contains the `sha` that won't be updated...

saraycp: I've adapted the BranchPackageStep code. When the PR/MR is updated, the `_branch_request` file is updated too, so it contains the new commit SHA. This way the source code gets updated and the build is triggered again.

I also adapted the specs accordingly.

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -95,7 +105,12 @@ def new_pull_request?
...
-      def add_branch_request_file(package:)
+      def updated_pull_request?
+        (github_pull_request? && @scm_extractor_payload[:action] == 'synchronize') ||
+          (gitlab_merge_request? && @scm_extractor_payload[:action] == 'update')
+      end
+
+      def add_or_update_branch_request_file(package:)
         branch_request_file = case @scm_extractor_payload[:scm]
                               when 'github'
                                 branch_request_file_github
```
danidoni, line 116: What about a better name like `branch_request_file_options_for_github`?

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -14,43 +14,53 @@ def initialize(step_instructions:, scm_extractor_payload:, token:)
...
 
+      def target_package
+        Package.find_by_project_and_name(target_project_name, target_package_name)
+      end
+
+      def find_or_create_branched_package
+        return target_package if updated_pull_request? && target_package.present?
+
+        branched_package = branch
+        return branched_package if branched_package
```
danidoni, line 59:
```suggestion
        branch # returns a branched package or nil, so the if is not really needed
```

danidoni: returns a branched package or nil, so the if is not really needed

saraycp: done

src/api/app/models/workflow/step/branch_package_step.rb
saraycp: I've used `branch_request_content_for_github` instead, since what the method returns is going to be the content of the file.

PR #11218, opened 2021-06-08
Use home project as namespace for branched package

vpereira:
As requested, namespace `target_project` under `home:$USER` 

At least for testing, to create the home project was necessary.

To test, I'm setting the `User.session!` to `user`, however this information should come from `token.user`

 
TODO:
- [x] check dependency on home project
- [ ] pass `token.user` to the `BranchPackageStep.call` method
https://trello.com/c/F014licj/1688-place-the-branched-package-under-home

### DIFF
--- a/src/api/app/models/workflow/step/branch_package_step.rb
+++ b/src/api/app/models/workflow/step/branch_package_step.rb
@@ -42,7 +42,8 @@ class Workflow
       end
 
       def target_project
-        source_project + ":PR-#{@scm_extractor_payload[:pr_number]}"
+        # TODO: should be token.user.login
+        "home:#{User.session!.login}:#{source_project}:PR-#{@scm_extractor_payload[:pr_number]}"
       end
 
       def remote_source?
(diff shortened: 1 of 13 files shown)


vpereira:
[Place the branched package under home:](https://trello.com/c/F014licj/1688-place-the-branched-package-under-home)

saraycp:
> To test, I'm setting the `User.session!` to `user`, however this information should come from `token.user`

After #11221  gets merged, the `token` will be passed to the `BranchPackageStep` model, so you can use `@token.user` instead of the session.

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -42,7 +42,9 @@ def target_package
       end
 
       def target_project
-        source_project + ":PR-#{@scm_extractor_payload[:pr_number]}"
+        # TODO
+        # it must be token.user.login
```
hennevogel, line 46: *nitpick* 
```suggestion
        # TODO: should be token.user.login
```

hennevogel (approved):

saraycp (approved):
LGTM. Let's try to merge and rebase PR #11221 first.

PR #11220, opened 2021-06-08
[backend] forward and log X-Request-ID headers

mlschroe:
This allows us to track the origin of requests more easily.
The api was changed to set this header in commit
863e09b8d2e2e88ddd953e201ce84230b84ea44a.

-----------------


hennevogel:
awesome

src/backend/bs_repserver
```diff
@@ -2129,6 +2129,7 @@ sub notify_jobresult {
   $buildtype = $1 if $info->{'file'} =~ /\.(spec|dsc|kiwi|livebuild)$/;
   $buildtype ||= Build::recipe2buildtype($info->{'file'}) || 'unknown';
   $buildtype = $info->{'imagetype'} && ($info->{'imagetype'}->[0] || '') eq 'product' ? 'kiwi-product' : 'kiwi-image' if $buildtype eq 'kiwi';
+  $ninfo{'buildtype'} = $buildtype;
```
hennevogel, line 2132: double awesome!

PR #11087, opened 2021-05-05
Single Distro API

hennevogel:
Makes it possible to CRUD a single distribution.

Fixes #10695

PR #11223, opened 2021-06-08
Remove distro hashing

hennevogel:
We render/display `Distribution`s from a hash because the `Distribution`s from the interconnect are from an XML.
This PR changes `Distribution`s to be fully based on records. Including the in-memory cache for `Distribution`s from remote.

### DIFF
--- a/src/api/app/controllers/distributions_controller.rb
+++ b/src/api/app/controllers/distributions_controller.rb
@@ -46,6 +46,8 @@ class DistributionsController < ApplicationController
   # PATCH/PUT /distributions/1234
   def update
     distribution = Distribution.find(params[:id])
+    # We don't allow updating remote distributions
+    distribution.readonly! if distribution.remote
 
     if distribution.update_from_xmlhash(@body_xml)
       render_ok
@@ -58,6 +60,8 @@ class DistributionsController < ApplicationController
   # DELETE /distributions/1234
   def destroy
     distribution = Distribution.find(params[:id])
+    # We don't allow deleting remote distributions
+    distribution.readonly! if distribution.remote
     distribution.destroy
 
     render_ok
--- a/src/api/app/controllers/distributions_controller.rb
+++ b/src/api/app/controllers/distributions_controller.rb
@@ -13,7 +13,7 @@ class DistributionsController < ApplicationController
 
   # GET /distributions
   def index
-    @distributions = Distribution.all_as_hash
+    @distributions = Distribution.local
 
     respond_to do |format|
       format.xml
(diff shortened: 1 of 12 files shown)


hennevogel:
Removed DNM label, #11087 was merged.

danidoni (dismissed):

src/api/app/controllers/distributions_controller.rb
```diff
@@ -23,7 +23,7 @@ def index
 
   # GET /distributions/1234
   def show
-    @distribution = Distribution.find(params[:id]).to_hash
+    @distribution = Distribution.find(params[:id])
```
danidoni, line 26: If you move this into a `before_action` you can reuse it below and fix the code climate issues about `destroy` action too complex

hennevogel:
:monocle_face: the codeclimate issue is about `RepositoriesController`

src/api/app/controllers/distributions_controller.rb
hennevogel: :monocle_face: the codeclimate issue is about `RepositoriesController`

danidoni: So true! Forget it, then.

src/api/config/clock.rb
```diff
@@ -48,6 +48,10 @@ module Clockwork
     BsRequest.delayed_auto_accept
   end
 
+  every(1.hour, 'refresh remote distros') do
+    FetchRemoteDistributionsJob.perform_later
```
vpereira, line 52: why this is necessary after this PR and it wasn't needed before it?

hennevogel: Before we stored the `distribution.xml` from the  interconnect projects in memcache. See

https://github.com/openSUSE/open-build-service/pull/11223/files#diff-66b82a7133958122ff9b8c0436aba2f4a3eef044f869b90c61209221afc288f4L54-L57

hennevogel: then converted XML to XMLHash in the `DistributionController`. Used that XMLHash in the XML views.

Which also means we where processing the Hash in `RepositoryController` when displaying them in the UI yadda yadda yadda. Hash overflow....

vpereira: > Before we stored the `distribution.xml` from the interconnect projects in memcache

Isn't this approach better? Why should we keep updating it every hour regardless of traffic or not?

hennevogel: > Isn't this approach better? Why should we keep updating it every hour regardless of traffic or not?

Better how?

vpereira (approved):

PR #11225, opened 2021-06-09
Report build results of multiple repositories/architectures

dmarcoux:


dmarcoux:
[Report build results of multiple repositories/architectures](https://trello.com/c/lCtUZqWY/1674-report-build-results-of-multiple-repositories-architectures)

vpereira (approved):

src/api/app/models/token/workflow.rb
```diff
@@ -28,7 +28,14 @@ def run_step_and_report(step, scm_extractor_payload, scm_token)
     raise "We couldn't branch your package" unless package_from_step
 
     set_subscription(package_from_step, scm_extractor_payload)
-    SCMStatusReporter.new(scm_extractor_payload, scm_token).call
+
+    Project.get_by_name(step.source_project).repositories.each do |repository|
```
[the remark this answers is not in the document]
dmarcoux: In this case, it's going to be the same since `step.target_project` is a subproject of `step.source_project`.

src/api/app/services/scm_status_reporter.rb
```diff
@@ -12,25 +13,30 @@ def call
     if github?
       github_client = Octokit::Client.new(access_token: @scm_token)
       # https://docs.github.com/en/rest/reference/repos#create-a-commit-status
-      github_client.create_status("#{@payload[:target_repository_full_name]}",
-                                  @payload[:commit_sha],
+      github_client.create_status("#{@event_subscription_payload[:target_repository_full_name]}",
```
danidoni, line 16: Are those `#{}` really needed?

dmarcoux: You're right, I don't think that they are needed.

PR #11228, opened 2021-06-10
Report build results of multibuild flavors to SCM

dmarcoux:
We don't know the multibuild flavors before doing the initial report, so they will only be reported back to the SCM once the builds are done.

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -32,6 +32,7 @@ class Token::Workflow < Token
     Project.get_by_name(step.target_project).repositories.each do |repository|
       # TODO: Fix n+1 queries
       repository.architectures.each do |architecture|
+        # We cannot report multibuild flavors here... so they will be missing from the initial report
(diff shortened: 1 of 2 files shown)


dmarcoux:
[Report build results of multiple repositories/architectures](https://trello.com/c/lCtUZqWY/1674-report-build-results-of-multiple-repositories-architectures)

src/api/app/models/token/workflow.rb
```diff
@@ -32,6 +32,7 @@ def run_step_and_report(step, scm_extractor_payload, scm_token)
     Project.get_by_name(step.target_project).repositories.each do |repository|
       # TODO: Fix n+1 queries
       repository.architectures.each do |architecture|
+        # We cannot report multibuild flavors here... so they will be missing from the initial report
```
dmarcoux, line 35: This might be possible if #11091 is merged. Let's see...

hennevogel (approved):

PR #11229, opened 2021-06-10
Fix filter text for involved projects/packages

### LINKED ISSUE #10811: Text for Empty List of a User's Involved Projects/Packages Is Sometimes Wrong
# Issue Description
On the profile page of a user, the text for an empty list of involved projects/packages is sometimes wrong. For example, if a user is only involved in projects and we filter by _Packages_, a text is displayed to let the current user that there are no involved projects/packages. This is wrong since the user is involved in some projects.

Expected Result
============
We should take into consideration whether the user is involved in projects/packages when filtering by _Packages_ or _Projects_ and change the text if needed.

How to Reproduce
===============

It's quite easy to reproduce in our development environment. Create some test data by running `bundle exec rake dev:development_testdata:create` in the frontend container.

1. Go to http://localhost:3000/users/Iggy and see that Iggy is involved in the project `home:Iggy`
2. Filter by _Packages_
3. See the wrong text `

Further Information
===============

Iggy's profile and the list of involved projects/packages:
iggy

Filter by _Packages_ of Iggy's involved projects/packages (with the wrong text):
wrong-text

rubhanazeem:
Filter message was hardcoded and was showing the wrong text.

Fixes #10811

### DIFF
--- a/src/api/app/helpers/webui/user_helper.rb
+++ b/src/api/app/helpers/webui/user_helper.rb
@@ -86,7 +86,22 @@ module Webui::UserHelper
     end
   end
 
+  def filter_message(params)
+    roles = params.select { |param| param.include?('role') }.keys
+    result = "This user is not involved in any #{project_package_message}"
+    result += " for the selected #{'role'.pluralize(roles.count)}" if roles.present?
+    "#{result}."
+  end
+
   private
+
+  def project_package_message
+    arr = []
+    arr << 'project' if params['involved_projects']
+    arr << 'package' if params['involved_packages']
+
+    arr.blank? ? 'project or package' : arr.join(' or ')
+  end
 
   def display_filters(filters)
     filters.collect do |filter|
--- a/src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
+++ b/src/api/app/views/webui/user/user_profile_redesign/_involvement.html.haml
@@ -60,7 +60,7 @@
       = display_applied_filters(filters, displayed_user)
   .card-body.pt-0
     - if involved_items.blank?
-      This user is not involved in any project or package.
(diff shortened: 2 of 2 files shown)


dmarcoux:
I'm pretty sure I created an issue for that, right? If it's the case, please link it in the commit description so we can close the issue when merging this PR.

dmarcoux:
The commit message `Updated filter message.` is also quite generic, could you be a bit more specific?

hennevogel:
We even have some [suggestions for commit messages](https://github.com/openSUSE/open-build-service/blob/master/CONTRIBUTING.md#proper-commit-messages)

src/api/app/helpers/webui/user_helper.rb
```diff
@@ -86,6 +86,17 @@ def activity_date_commits(projects)
     end
   end
 
+  def filter_message(params)
```
dmarcoux, line 89: This code is only considering the `Projects / Packages` filter, although we also need to consider the `Role` filter. See the screenshot below, the message is also wrong because it should be `This user is not involved in any project or package for the selected role`. Multiple roles can be selected, so to pluralize or not the word `role`, you can use [pluralize](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-pluralize).

role-filter

dmarcoux:
It's me again :wave:... Regarding commit messages, if you use the present tense, they will be shorter and you won't miss any detail. It's a win-win situation, you type less and it's conveying the same information. You can also drop the dot at the end of the commit message, it's not needed.

So with all this, `Fixed filter text for involved projects/packages.` becomes `Fix filter text for involved projects/packages`.

dmarcoux (approved):

PR #11230, opened 2021-06-10
Link to the Commit Style wiki page in the 'Proper Commit Messages' section

dmarcoux:
Having a single source of truth is better than having duplicated information.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,18 +59,7 @@ of code to be reviewed.
 
 ## Proper Commit Messages
 
-We are keen on proper commit messages because they will help us to maintain
-this piece of code in the future. So for the benefit of all the people who will
-look at this commit now and in the future, follow this style:
-
-- The title of your commit should summarizes **what** has been done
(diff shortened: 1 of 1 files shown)


dmarcoux:
We either link to the Commit Style wiki page or pull its content in the CONTRIBUTING.md. I prefer the approach taken in this PR.

What do you think?

danidoni (approved):

hennevogel:
I don't like throwing heaps and heaps of rules at people. It's a big contribution hurdle. Remember how you felt with your first open source contribution. I was shitting my pants that I would do something wrong and that people would be mean or that something would break. If you would have shown me a 20 item list to follow I would just not do it.

dmarcoux:
@hennevogel True, a big list of things to do can be scary for no contributors. So you prefer to keep it as it is?

dmarcoux:
@danidoni is for the change, @hennevogel is against it. I didn't get other feedback, so I'll close this PR since it's not unanimous.

PR #11237, opened 2021-06-11
Hardcode 'opened' on _branch_request file

saraycp:
GitHub actions like `synchronize` are not handled by `check_for_branch_request` method in [obs-service-tar_scm](https://github.com/openSUSE/obs-service-tar_scm/blob/2319f50e741e058ad599a6890ac5c710112d5e48/TarSCM/tasks.py#L145). Only `opened` action.

For now, we are going to make the `_branch_request` file contain `'opened'` action to ensure the source files are updated correctly in the OBS package.

There is nothing to change on the GitLab part in this regard, as  `check_for_branch_request` checks ` 'object_kind'` which is always `'merge_request'`, it doesn't change on different actions (there is a `object_attributes > action` key in GiLab's payload  for that)

### DIFF
--- a/src/api/app/models/workflow/step/branch_package_step.rb
+++ b/src/api/app/models/workflow/step/branch_package_step.rb
@@ -122,7 +122,10 @@ class Workflow
 
       def branch_request_content_github
         {
-          action: @scm_extractor_payload[:action],
+          # TODO: change to @scm_extractor_payload[:action]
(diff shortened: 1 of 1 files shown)


danidoni (approved):

dmarcoux:
I created an issue in obs-service-tar_scm: https://github.com/openSUSE/obs-service-tar_scm/issues/415

dmarcoux (approved):

PR #11245, opened 2021-06-15
Update Rails version in Gemfile.next.lock

dmarcoux:
I have overwritten `Gemfile.next.lock` with `Gemfile.lock`, then updated Rails with `next bundle update rails`


dmarcoux:
It's expected that the `next_rails` fails when trying to both the Rails application. This is exactly what needs to be fixed with `next_rails`. Every other CI check should be green.

vpereira (approved):

dmarcoux:

PR #11250, opened 2021-06-16
Feature: Distribution variants

hennevogel:
If two `Distribution` from the same vendor have the same repository name, they are considered "variants". You can only enable one variant in the UI.
    
This is useful for stacked distributions where one includes the other. Like Arch community which includes all packages from Arch extra.
    
Additionally:
    
This moves the distribution related UI functionality into it's own controller: Webui::DistributionsController

Peek 2021-06-24 15-15


hennevogel:
Do not merge, it's based on #11223 which is not merged yet...

hennevogel:
#11223 is merged. Removing DNM label.

src/api/app/assets/javascripts/webui/distributions.js
```diff
@@ -0,0 +1,8 @@
+function toggleDistribution() { // jshint ignore:line
+  $('.distribution-input').on('change', function (){
+    var distributionId = $( this ).attr('data-distribution');
+    $('#distribution-' + distributionId + '-checkbox').prop( "disabled", true );
```
vpereira, line 4: can we use single quotes here too?

src/api/app/controllers/webui/distributions_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  after_action :verify_authorized
+
+  # GET /projects/:project/distributions/new
+  def new
+    authorize @project, :update?
+
+    @distributions = Distribution.all.group_by(&:vendor)
+    return if @distributions.present?
+
+    if User.admin_session?
```
vpereira, line 13: couldn't you instead of write custom code, use the `before_action :require_admin` ?

src/api/app/models/concerns/project_distribution.rb
```diff
@@ -0,0 +1,21 @@
+module ProjectDistribution
+  extend ActiveSupport::Concern
+
+  # Check if the project has a path_element matching project and repository
+  def has_distribution(project_name, repository)
+    has_local_distribution(project_name, repository) || has_remote_distribution(project_name, repository)
+  end
+
+  def has_remote_distribution(project_name, repository)
```
vpereira, line 9: I know the original code doesn't document those two methods. It would however be good to have it documented, either as a code (spec) or inline comment like the method `has_distribution`

vpereira (changes requested):

src/api/app/controllers/webui/distributions_controller.rb
hennevogel: I'm not writing custom code, I'm copy and pasting it 😄  `require_admin` wouldn't redirect to the interconnect.

src/api/app/models/concerns/project_distribution.rb
hennevogel: It's documented on line 4 no?

src/api/app/controllers/webui/distributions_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  private
+
+  def create_repository_from_distribution
+    repository = Repository.new_from_distribution(@distribution)
+    repository.project = @project
+
+    if repository.save
+      @project.store(comment: "Added #{repository.name} repository")
+    else
+      flash.now[:error] = "Can not add repository: #{repository.errors.full_messages.to_sentence}"
```
dmarcoux, line 42:
```suggestion
      flash.now[:error] = "Can't add repository: #{repository.errors.full_messages.to_sentence}"
```

src/api/app/controllers/webui/distributions_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    @project.repositories.delete(@repository)
+    if @project.valid?
+      @project.store(comment: "Removed #{@repository.name} repository")
+    else
+      flash.now[:error] = "Failed to remove repository: #{@project.errors.full_messages.to_sentence}"
+    end
+  end
+
+  def set_distribution
+    @distribution = Distribution.find(params[:distribution])
```
dmarcoux, line 56: I don't see the point of having a method which is then a `before` callback if it's only used in one action. It can be done in the action directly and simplify the code.

src/api/app/models/distribution.rb
```diff
@@ -38,6 +38,10 @@ def update_from_xmlhash(xmlhash)
 
     self
   end
+
+  def variants
+    Distribution.where(vendor: vendor, reponame: reponame).where.not(id: id)
```
dmarcoux, line 43: This yells query object to me. It's for sure not an instance method.

src/api/app/views/webui/repositories/_breadcrumb_items.html.haml
```diff
@@ -12,6 +12,5 @@
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
     - if defined?(@repository) && current_page?(project_repository_state_path)
       Repository State
-    - elsif current_page?(repositories_distributions_path)
-      Add from a Distribution
+
```
dmarcoux, line 15: This extra line isn't needed.

```suggestion
```

src/api/app/views/webui/distributions/new.html.haml
```diff
@@ -0,0 +1,39 @@
...
+    %h3 Add Repositories to #{@project}
+    .row
+      - @distributions.each do |vendor, distributions|
+        .col-md-12
+          .pb-4
+            %h6.pb-2
+              - if resolve_asset_path("distributions/#{vendor.downcase}.svg")
+                = image_tag("distributions/#{vendor.downcase}.svg")
+              #{vendor} distributions
+              %span.badge.badge-primary
```
dmarcoux, line 16: I don't know about this... Is it really needed?

src/api/app/models/distribution.rb
hennevogel: Hm but it depends on the instance. Only instances have variants right?

hennevogel: Anyway it's not used at all. I can remove it if you don't like it.

src/api/app/views/webui/distributions/new.html.haml
hennevogel: Needed? No

hennevogel:
Some context for this one: https://survey.opensuse.org/index.php/517118?lang=en

src/api/app/models/distribution.rb
dmarcoux: Let's remove it :)

src/api/app/views/webui/distributions/new.html.haml
dmarcoux: I should have phrased my previous comment more like `I don't think knowing the total of distributions is really useful`

src/api/app/controllers/webui/distributions_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  private
+
+  def create_repository_from_distribution
+    repository = Repository.new_from_distribution(@distribution)
+    repository.project = @project
+
+    if repository.save
+      @project.store(comment: "Added #{repository.name} repository")
+    else
+      flash.now[:error] = "Can't add repository: #{repository.errors.full_messages.to_sentence}"
```
dmarcoux, line 42: When trying to add a variant when there is already another enabled (like enabling `Arch Extra`, then trying to enable `Arch Community`), this results in an error message like:

```
Can't add repository: Name Arch is already used by a repository of this project 
```

This is somewhat cryptic. Could we improve it somehow?

hennevogel: how is this cryptic?

dmarcoux: I think an error message like `Only one variant of repository_name can be enabled.` would be much better.

hennevogel: This is not limited to variants. You can't have two repositories with the same name no matter what their relationship is.

hennevogel: Hm maybe I could catch exactly this validation error and change the message in the distribution controller...

hennevogel: Might also be related to #11317

src/api/app/controllers/webui/distributions_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+                  alert: 'There are no distributions configured. Maybe you want to connect to one of the public OBS instances?')
+    else
+      redirect_to(project_repositories_path(project: @project))
+    end
+  end
+
+  # PATCH /projects/:project_name/distributions/:id/toggle
+  def toggle
+    authorize @project, :update?
+    @distribution = Distribution.find(params[:distribution])
```
vpereira, line 23: are we super confident that this will never fail? Should we use `find!` or check if `@distribution.present?` ?

src/api/app/controllers/webui/distributions_controller.rb
vpereira: will you investigate if it is related with #11317 and try to address it here?

src/api/app/models/repository.rb
```diff
@@ -62,6 +62,13 @@ def self.find_by_project_and_name(project, repo)
     nil
   end
 
+  def self.find_by_project_and_name!(project, repo)
+    result = find_by_project_and_name(project, repo)
+    return ActiveRecord::RecordNotFound unless result
```
vpereira, line 67: :pick: `return ActiveRecord::RecordNotFound if result.blank?` to be explicit

src/api/app/views/webui/distributions/new.html.haml
```diff
@@ -0,0 +1,37 @@
...
+                                                                  content: 'Variants are distributions of the same general kind that only
+                                                                            differ in the set of packages they include. You should consult
+                                                                            the distributions documentation to learn more about the variants.
+                                                                            It is only possible to add one distribution variant as repository.'
+                                                                } }
+                      \:
+                    - variants.each do |variant|
+                      = form_tag(toggle_project_distributions_path, remote: true, id: "distribution-#{variant.id}-form") do
+                        - checked = @project.has_distribution(variant.project, variant.repository) ? 'checked' : nil
+                        = render partial: 'variant', object: variant, locals: { checked: checked }
```
vpereira, line 34: if you could have the whole section from line [17](https://github.com/openSUSE/open-build-service/pull/11250/commits/da6e9a6aaeff23c9dbd741354beb719e18606b08#diff-685a2370d164f72f6d8bb4fa79af1561e894b348f1adaa120b1387a527586a78R17) upto here in a partial, then you didn't have to render a partial inside of a each loop, which normally impacts badly the performance..

src/api/app/views/webui/repositories/_project_repositories.html.haml
```diff
@@ -10,10 +10,11 @@
 
     %ul.list-inline.mb-0
       - if user_can_modify
-        %li.list-inline-item
-          = link_to(repositories_distributions_path(project: project), class: 'nav-link') do
-            %i.fas.fa-plus-circle.text-primary
-            Add from a Distribution
+        - if Distribution.any? || User.admin_session?
```
vpereira, line 13: is admin power here really mandatory?

src/api/spec/models/repository_spec.rb
```diff
@@ -135,4 +135,29 @@
       it { expect(subject.download_repositories.pluck(:arch, :url)).to contain_exactly([dod_repository.arch, dod_repository.url]) }
     end
   end
+
+  describe '#new_from_distribution' do
+    let(:target_project) { create(:project) }
+    let(:project) { create(:project_with_repository) }
+    let(:distribution) { create(:distribution, project: project, repository: project.repositories.first.name) }
+    let(:subject_repository) do
+      repository = Repository.new_from_distribution(distribution)
```
vpereira, line 144: is it possible to do it with Factories?

vpereira (changes requested):

src/api/app/controllers/webui/distributions_controller.rb
hennevogel: `find` always raises. 

```ruby
Distribution.find(1312)
ActiveRecord::RecordNotFound: Couldn't find Distribution with 'id'=1312
```

src/api/app/views/webui/distributions/new.html.haml
hennevogel: I would but I can't group a collection AFAIK. The performance impact is also bearable, the number of distributions is rather static.

src/api/app/views/webui/repositories/_project_repositories.html.haml
hennevogel: Do create an interconnect? Sure.

src/api/spec/models/repository_spec.rb
hennevogel: Hm sure but what would we test then? `new_from_distribution` is the subject.

src/api/app/controllers/webui/distributions_controller.rb
hennevogel: No I will address this later.

vpereira (approved):

PR #11251, opened 2021-06-17
Recover from Sphinx connection errors on search

danidoni:
When the connection to the search service drops, shows an error message instead of crashing and logs the exception as well.

### DIFF
--- a/src/api/app/controllers/webui/search_controller.rb
+++ b/src/api/app/controllers/webui/search_controller.rb
@@ -163,5 +163,9 @@ class Webui::SearchController < Webui::WebuiController
                                 issue_tracker_name: @search_tracker)
     @results = search.search(page: params[:page], per_page: @per_page)
     flash[:notice] = 'Your search did not return any results.' if @results.empty?
+  rescue ThinkingSphinx::SphinxError => e
+    flash[:error] = "There has been an error performing the search. We're working to fix it. Please, try again later."
+    Airbrake.notify(e)
+    @results = []
   end
 end


danidoni:
[Make Sphinx integration more robust](https://trello.com/c/eh2z79U3/1691-make-sphinx-integration-more-robust)

src/api/app/controllers/webui/search_controller.rb
```diff
@@ -163,5 +163,9 @@ def search
                                 issue_tracker_name: @search_tracker)
     @results = search.search(page: params[:page], per_page: @per_page)
     flash[:notice] = 'Your search did not return any results.' if @results.empty?
+  rescue ThinkingSphinx::ConnectionError => e
+    flash[:error] = 'Error connecting to the search service.'
```
hennevogel, line 167: Not very helpful for you customers. What can they do about it? :)

And what about all the other exceptions that could occur?

hennevogel (changes requested):

src/api/app/controllers/webui/search_controller.rb
hennevogel: > I think our customers can do nothing except wait until we fix whatever is the reason for the connection error. It's on our side to detect sphinx downtimes via monitoring and bring it back.

That's what you should tell them then.

hennevogel:
@danidoni you should squash commits

dmarcoux (approved):

hennevogel (approved):

PR #11260, opened 2021-06-18
Improve performance of Request#show with multiple actions 

rubhanazeem:
Request#show was loading the complete diff history in a single HTTP
request which was time consuming and even crashing the browser due to
large html content. Now the content can be loaded based on separate actions.

**Update:** The last version of this code will load the whole page first and, after that, will perform an ajax call to get the content of the first BsRequest Action tab (i.e. "Submit chromium"). Then, every time we click on a new tab, a new ajax call is performed to retrieve its content, unless it was already retrieved.

video


src/api/app/assets/javascripts/webui/packages.js
```diff
@@ -14,13 +14,13 @@ $(function ($) {
     }
   });
 
-  $('.expand-diffs').on('click', function () {
+  $('body').on('click', '.expand-diffs', function () {
```
dmarcoux, line 17: What makes this change necessary?

rubhanazeem: It was because we re-render the partial in ajax and `$('.expand-diffs').on('click', function () {` was not enough to find the element in DOM.

dmarcoux: Could you add this to the commit description? This isn't obvious.

rubhanazeem: Sure, I'll add it to the commit description.

src/api/app/assets/javascripts/webui/request.js
```diff
@@ -107,3 +107,23 @@ function requestAddAutocomplete(autocompleteElement) { // jshint ignore:line
     $('.hideable input:visible').removeAttr('disabled');
   });
 }
+
+$(document).ready(function(){
+  $('a[data-toggle="tab"]').on('shown.bs.tab', function () {
```
dmarcoux, line 112: With this code, we're loading diffs whenever a tab is clicked. This is great.

There's one issue though. Let's say that we have a request with 2 tabs: _Tab 1_ and _Tab 2_. So when the page is loaded, _Tab 1_ will also be loaded. I then click on _Tab 2_ and its diffs are asynchronously loaded. I go back to _Tab 1_ and its diffs are loaded **again**. So whenever I click on a tab for which its diffs were previously loaded, they are loaded again since this code is sending another AJAX request. This should be avoided.

dmarcoux (changes requested):

src/api/app/assets/javascripts/webui/request.js
saraycp: This suggestion is already tackled. We have added new commits to make it easier to understand the steps we followed. We'll like to squash them before it gets merged.

Can you please have a look, @dmarcoux ?

saraycp:
This PR is ready for review.

We have decided to ignore CodeClimate complaints since it's related to some old code that we have just moved to a new private method, `actions_details`. We don't plan to refactor this code now.

DO_NOT_MERGE, we'd like to squash the commits before it gets merged.

src/api/app/assets/javascripts/webui/request.js
rubhanazeem: Done. It's updated. We don't send a new AJAX now.

rubhanazeem:
The commits are already squashed. The PR is ready for review. Removing DO NOT MERGE.

src/api/app/assets/javascripts/webui/request.js
```diff
@@ -110,7 +110,7 @@ function requestAddAutocomplete(autocompleteElement) { // jshint ignore:line
 
 $(document).ready(function(){
   var element = $('.bs-request-actions li:first-child a:first-child');
-  if (element['length'] !== 0){
+  if (element.length !== 0){
```
dmarcoux, line 113: This change belongs to the previous commit ` Improve performance of Request#show with multiple actions`.

rubhanazeem: Done. It's moved to the correct commit.

dmarcoux (approved):
Great work!

PR #11261, opened 2021-06-19
Handle the project and package indexing offline

danidoni:
Propagating the changes to projects and packages using jobs allows us to recover better from integration errors with Sphinx.

If something happened with the search service, the content will propagate to Sphinx eventually, once we have fixed whatever it is
happening.

    https://github.com/collectiveidea/delayed_job#gory-details

NOTE: The original in-line implementation had a "changes detection" to avoid indexing the models if there were no changes to name, title or description. We lose the changes detection mechanism when deferring the indexing into jobs, but its worth enough.


danidoni:
[Make Sphinx integration more robust](https://trello.com/c/eh2z79U3/1691-make-sphinx-integration-more-robust)

src/api/app/jobs/populate_to_sphinx_job.rb
```diff
@@ -0,0 +1,8 @@
+class PopulateToSphinxJob < ApplicationJob
+  def perform(id:, model_name:)
+    object = model_name.to_s.camelize.safe_constantize.find(id)
```
hennevogel, line 3: Maybe...

```suggestion
    object = model_name == 'Project' ? Project.find(id) : Package.find(id)
```

hennevogel: I guess we should also guard about the record being gone...

danidoni: Done

src/api/app/jobs/populate_to_sphinx_job.rb
```diff
@@ -0,0 +1,20 @@
+class PopulateToSphinxJob < ApplicationJob
+  ALLOWED_MODELS = {
+    package: Package,
+    project: Project
+  }.freeze
+
+  queue_as :quick
+
+  def perform(id:, model_name:)
+    model_class = ALLOWED_MODELS[model_name]
```
vpereira, line 10: `ALLOWED_MODELS.fetch(model_name)` is better because the exception handling is done automatically when a lookup fails on fetch since it raises a KeyError

vpereira (changes requested):

src/api/app/jobs/populate_to_sphinx_job.rb
danidoni: Sure!

src/api/app/models/package.rb
```diff
@@ -1408,6 +1409,13 @@ def update_activity
   def fetch_rev_from_history_cache(rev)
     Rails.cache.read(['history', self, rev]) || Rails.cache.read(['history_md5', self, rev])
   end
+
+  def populate_to_sphinx
+    if title_previously_changed? ||
```
hennevogel, line 1414: You can create a project without title and description. 

Screenshot from 2021-06-22 17-13-40

hennevogel (changes requested):

src/api/app/models/package.rb
danidoni: So true. I've fixed both Project and Package, because we can create packages without title or description as well...

hennevogel (approved):

PR #11263, opened 2021-06-21
Add revision enforcement

coolo:
Continuing on https://github.com/openSUSE/open-build-service/pull/10992

This is a security flaw in our development flow so we want to have this fixed asap - but the original PR didn't move in the last 2 months.

### DIFF
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -792,17 +792,13 @@ class BsRequestAction < ApplicationRecord
         add_revision = tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
         if add_revision && source_rev
           # Enforce expanded revision of given revision by user
-          if dir['entry'].is_a?(Array) && dir['entry'].map { |e| e['name'] }.include?('_link') ||
-             dir['entry'].is_a?(Hash) && dir['entry']['name'] == '_link'
-            raise ExpandError, 'Unexpanded links are forbidden by OBS:EnforceRevisionsInRequests'
-          end
+          raise ExpandError, 'Unexpanded links are forbidden by OBS:EnforceRevisionsInRequests' if dir.elements('entry').any? { |e| e['name'] == '_link' }
         end
       end
       if add_revision && !source_rev
-        if action_type == :maintenance_release && dir['entry']
+        if action_type == :maintenance_release
           # patchinfos in release requests get not frozen to allow to modify meta data
-          return if dir['entry'].is_a?(Array) && dir['entry'].map { |e| e['name'] }.include?('_patchinfo')
-          return if dir['entry'].is_a?(Hash) && dir['entry']['name'] == '_patchinfo'
+          return if dir.elements('entry').any? { |e| e['name'] == '_patchinfo' }
         end
         self.source_rev = dir['srcmd5']
       end
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -786,16 +786,16 @@ class BsRequestAction < ApplicationRecord
       query[:expand] = 1 unless updatelink
       query[:rev] = source_rev if source_rev
       dir = Xmlhash.parse(Backend::Api::Sources::Package.files(source_project, source_package, query))
-      unless add_revision
-        # Enforce revisions?
-        tprj = Project.get_by_name(target_project)
-        if tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
-          raise ExpandError, 'updatelink option is forbidden for OBS:EnforceRevisionsInRequests' if updatelink
 
-          # fix the revision to the expanded sources at the time of submission
-          self.source_rev = dir['srcmd5']
-        end
+      # Enforce revisions?
+      tprj = Project.get_by_name(target_project)
+      if tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
+        raise ExpandError, 'updatelink option is forbidden for requests against projects with the attribute OBS:EnforceRevisionsInRequests' if updatelink
+
+        # fix the revision to the expanded sources at the time of submission
+        self.source_rev = dir['srcmd5']
       end
+
       if add_revision && !source_rev
         if action_type == :maintenance_release
           # patchinfos in release requests get not frozen to allow to modify meta data
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -789,8 +789,9 @@ class BsRequestAction < ApplicationRecord
       unless add_revision
         # Enforce revisions?
         tprj = Project.get_by_name(target_project)
-        add_revision = tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
-        if add_revision
+        if tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
+          raise ExpandError, 'updatelink option is forbidden for OBS:EnforceRevisionsInRequests' if updatelink
+
           # fix the revision to the expanded sources at the time of submission
           self.source_rev = dir['srcmd5']
         end
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -791,7 +791,7 @@ class BsRequestAction < ApplicationRecord
         tprj = Project.get_by_name(target_project)
         add_revision = tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
         if add_revision
-          # fix the revision to the expanded sources
+          # fix the revision to the expanded sources at the time of submission
           self.source_rev = dir['srcmd5']
         end
       end
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -790,9 +790,9 @@ class BsRequestAction < ApplicationRecord
         # Enforce revisions?
         tprj = Project.get_by_name(target_project)
         add_revision = tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
-        if add_revision && source_rev
-          # Enforce expanded revision of given revision by user
-          raise ExpandError, 'Unexpanded links are forbidden by OBS:EnforceRevisionsInRequests' if dir.elements('entry').any? { |e| e['name'] == '_link' }
+        if add_revision
+          # fix the revision to the expanded sources
+          self.source_rev = dir['srcmd5']
         end
       end
       if add_revision && !source_rev
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -786,6 +786,18 @@ class BsRequestAction < ApplicationRecord
       query[:expand] = 1 unless updatelink
       query[:rev] = source_rev if source_rev
       dir = Xmlhash.parse(Backend::Api::Sources::Package.files(source_project, source_package, query))
+      unless add_revision
+        # Enforce revisions?
+        tprj = Project.get_by_name(target_project)
+        add_revision = tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
+        if add_revision && source_rev
+          # Enforce expanded revision of given revision by user
+          if dir['entry'].is_a?(Array) && dir['entry'].map { |e| e['name'] }.include?('_link') ||
+             dir['entry'].is_a?(Hash) && dir['entry']['name'] == '_link'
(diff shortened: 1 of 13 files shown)


src/api/app/models/bs_request_action.rb
```diff
@@ -786,11 +786,19 @@ def check_for_expand_errors!(add_revision)
       query[:expand] = 1 unless updatelink
       query[:rev] = source_rev if source_rev
       dir = Xmlhash.parse(Backend::Api::Sources::Package.files(source_project, source_package, query))
+      unless add_revision
+        # Enforce revisions?
+        tprj = Project.get_by_name(target_project)
+        add_revision = tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
+        if add_revision
+          # fix the revision to the expanded sources
```
hennevogel, line 794: *Nitpick*:
```suggestion
          # fix the revision to the expanded sources at the time of submission
```

src/api/test/functional/request_controller_test.rb
```diff
@@ -202,6 +202,40 @@ def test_submit_request_of_new_package_with_devel_package
     assert_response :success
   end
 
+  def test_submit_with_and_without_revision
```
hennevogel, line 205: One test should be enough right? 😄

hennevogel (approved):

src/api/test/functional/request_controller_test.rb
coolo: Yes, we can remove this one IMO

adrianschroeter:
On Montag, 21. Juni 2021, 11:58:17 CEST Stephan Kulow wrote:
> @coolo commented on this pull request.
> 
> 
> 
> > @@ -202,6 +202,40 @@ def test_submit_request_of_new_package_with_devel_package
>      assert_response :success
>    end
>  
> +  def test_submit_with_and_without_revision
> 
> Yes, we can remove this one IMO

please don't, I like to extend it to check also to verify that wrong submitted 
revisions get catched.

And that won't work with VCR cassettes.

-- 

Adrian Schroeter ***@***.***>

coolo:
why not? I added exactly that actually

adrianschroeter:
On Montag, 21. Juni 2021, 12:04:17 CEST Stephan Kulow wrote:
> why not? I added exactly that actually

I do not see that you constructed a linked package resource where you specified 
the unexpanded md5sum in a submit action on request creation. Did you?

-- 

Adrian Schroeter ***@***.***>

coolo:
I submit revision 2, which points to a _link pointing to a 3rd package - and expect an expanded srcmd5 written in the request

adrianschroeter:
On Montag, 21. Juni 2021, 14:14:15 CEST Stephan Kulow wrote:
> I submit revision 2, which points to a _link pointing to a 3rd package - and expect an expanded srcmd5 written in the request

But rev=2 is not part of the action xml, right?

In that case an error must occur, since you reference explicit to a not frozen revision.

The user must know that he can not reference _link file changes basically, since he
can not submit such changes that way anymore.

-- 

Adrian Schroeter ***@***.***>

coolo:
rev=2 is part of the original xml, but not of the one "saved". I guess I don't understand the story. Basically the tpackage being a _link?

adrianschroeter:
It is about the source package, we need to verify following situations with the attribute set:

1) request action xml has no revision specified

2) request action xml has a revsion specified (no matter if md5sum or number).

2a) Source revision expands to a source without any _link file. We can be sure the source is frozen.
2b) Source revsion expands to a source containing a _link file, means we can not be sure that it is frozen (we do not want to evaluate the _link file content here). So we need to error out, since the user explicit used this revision. (Maybe in good face by suggesting a _link file change, but this is too complex to handle here).

Therefore we need a source server doing the source merging for a proper test case.

coolo:
I mean how can a package expand to a _link? if you can point me to one such case on OBS, I can extend the test.

adrianschroeter:
On Montag, 21. Juni 2021, 14:52:08 CEST Stephan Kulow wrote:
> I mean how can a package expand to a _link?

just pick any not expanded revision and use it. You need to have a package with
a _link as source of course for that.

In that case the request creation should fail with the right error.

Doing the same with the xsrcmd5 sum needs to work.

-- 

Adrian Schroeter ***@***.***>

coolo:
If you have a devel project, it will contain a link to Factory. If I submit that with revision=HEAD, it will be expanded and the expanded srcmd5 is written into the database. And these expanded sources are what we'll review

hennevogel:
@adrianschroeter so can we merge this and you write this test case afterward? You won't change the code right?

adrianschroeter:
On Montag, 21. Juni 2021, 17:37:25 CEST Henne Vogelsang wrote:
> @adrianschroeter so can we merge this and you write this test case afterward? You won't change the code right?

yes, I just asked to keep the mini test.


-- 

Adrian Schroeter ***@***.***>

marcus-h:
On 2021-06-21 02:38:19 -0700, Stephan Kulow wrote:
> This is a security flaw in our development flow so we want to have this fixed asap - but the original PR didn&#39;t move in the last 2 months.
> You can view, comment on, or merge this pull request online at:
>
> diff --git a/src/api/app/models/bs_request_action.rb b/src/api/app/models/bs_request_action.rb
> index efeddae94d..5d24d5db12 100644
> --- a/src/api/app/models/bs_request_action.rb
> +++ b/src/api/app/models/bs_request_action.rb
> @@ -786,11 +786,19 @@ class BsRequestAction < ApplicationRecord
>        query[:expand] = 1 unless updatelink
>        query[:rev] = source_rev if source_rev
>        dir = Xmlhash.parse(Backend::Api::Sources::Package.files(source_project, source_package, query))
> +      unless add_revision
> +        # Enforce revisions?
> +        tprj = Project.get_by_name(target_project)
> +        add_revision = tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
> +        if add_revision
> +          # fix the revision to the expanded sources at the time of submission
> +          self.source_rev = dir['srcmd5']

In case of an <options><updatelink>true</updatelink></options> request,
self.source_rev could still point to an unexpanded link. Since our
tooling only displays the expanded sources (for instance, when using
the /request/1234?cmd=diff route), this could be used (untested) to
"trick" a reviewer to accept accept a review. Once the request is
eventually accepted, the target package contains a _link file. At
this point, the _link could expand to "malicious"/unreviewed sources.

coolo:
true, will add an exception for this case

coolo:
readded the minitest (everyone her hobby :) and added code and test protecting against updatelink

marcus-h:
On 2021-06-21 10:25:27 -0700, Stephan Kulow wrote:
> readded the minitest (everyone her hobby :) and added code and test protecting against updatelink
>
Maybe we should execute the "enforce revisions?" codepath unconditionally?
(That is, do not guard it with "unless add_revision".). Otherwise, one
could create a request via POST /request?cmd=create&addrevision=1, which
has the updatelink option set to true. Such a request can be created but
can never be accepted (if I get the code right). Hence, this is just for
sanity and not security reasons:)

coolo:
agreed, changed that now

src/api/app/models/bs_request_action.rb
```diff
@@ -786,11 +786,20 @@ def check_for_expand_errors!(add_revision)
       query[:expand] = 1 unless updatelink
       query[:rev] = source_rev if source_rev
       dir = Xmlhash.parse(Backend::Api::Sources::Package.files(source_project, source_package, query))
+
+      # Enforce revisions?
+      tprj = Project.get_by_name(target_project)
+      if tprj.instance_of?(Project) && tprj.find_attribute('OBS', 'EnforceRevisionsInRequests').present?
+        raise ExpandError, 'updatelink option is forbidden for OBS:EnforceRevisionsInRequests' if updatelink
```
hennevogel, line 793:
```suggestion
        raise ExpandError, 'updatelink option is forbidden for requests against projects with the attribute OBS:EnforceRevisionsInRequests' if updatelink
```

src/api/db/data/20210409135426_add_enforce_revision.rb
```diff
@@ -0,0 +1,20 @@
...
+    role = Role.find_by_title('maintainer')
+    AttribTypeModifiableBy.find_or_create_by(role_id: role.id, attrib_type_id: at.id)
+
+    update_all_attrib_type_descriptions
+  end
+
+  def self.down
+    AttribType.find_by_namespace_and_name('OBS', 'EnforceRevisionsInRequests').delete
+  end
+end
```
hennevogel, line 20: I'm missing changes to `db/data_schema.rb`, should happen once you run the data migration.

src/api/spec/models/bs_request_action_spec.rb
```diff
@@ -309,4 +309,65 @@
...
+          expect(bs_request_action.source_rev.length).to eq(32)
+        end
+
+        context 'with updatelink' do
+          let(:bs_request) { create(:bs_request_with_submit_action, creator: user, target_package: target_package, source_package: source_package, updatelink: true) }
+
+          it 'throws exception' do
+            expect { bs_request.sanitize! }.to raise_error do |exception|
+              expect(exception.message.to_s).to eq('updatelink option is forbidden for OBS:EnforceRevisionsInRequests')
+            end
```
hennevogel, line 360:
```suggestion
             expect { bs_request.sanitize! }.to raise_error(APIError::ExpandError)
```

hennevogel: No need to test for the message text, just makes this harder to change.

hennevogel (changes requested):

src/api/db/data/20210409135426_add_enforce_revision.rb
coolo: hmm, this is a data migration

marcus-h:
Maybe we should also forbid source_rev == 'upload' in the enforce
revisions codepath (the upload rev can "survive" expand=1).

coolo:
That will never be returned as srcmd5 - and we rewrite the source_rev with the expanded srcmd5. So I don't see an issue here.
It's rather an issue for the non-attributed projects, which brings us to the question: why shouldn't this be the default?

marcus-h:
On 2021-06-22 22:00:06 -0700, Stephan Kulow wrote:
> That will never be returned as srcmd5 - and we rewrite the source_rev with the expanded srcmd5. So I don't see an issue here.

No, the upload rev survives the expand:

***@***.***:~> curl -u Marcus_H 'https://api.opensuse.org/source/home:Marcus_H:lnk2/b3?rev=upload&expand=1'
Enter host password for user 'Marcus_H':
<directory name="b3" rev="upload" srcmd5="upload">
  <entry name="foo" md5="14758f1afd44c09b7992073ccf00b43d" size="7" mtime="1624391770"/>
</directory>
***@***.***:~>

> It's rather an issue for the non-attributed projects, which brings us to the question: why shouldn't this be the default?
>
No idea/no opinion:)

coolo:
```
created request id 901414
coolo@nerissa#build-service>osc api /request/901414
<request id="901414" creator="coolo">
  <action type="submit">
    <source project="home:Marcus_H:lnk2" package="b3" rev="upload"/>
    <target project="home:coolo" package="b3"/>
  </action>
  <state name="review" who="coolo" when="2021-06-23T07:05:22">
    <comment/>
  </state>
  <review state="new" when="2021-06-23T07:05:22" by_package="b3" by_project="home:Marcus_H:lnk2"/>
  <description>trying</description>
</request>
```

This sounds like an even more severe bug - I guess we treat that one in an extra PR.

src/api/db/data/20210409135426_add_enforce_revision.rb
hennevogel: Yeah and this is `data_schema.rb`. So if you setup a fresh database for instance, the data migrations are not running afterward.

hennevogel (approved):

marcus-h:
On 2021-06-22 22:00:06 -0700, Stephan Kulow wrote:
> It's rather an issue for the non-attributed projects, which brings us to the question: why shouldn't this be the default?
>
Hmm thinking about it again: is the OBS:EnforceRevisionsInRequests really
the right mean/tool to "fix" the underlying problem? For instance, when
using this attribute, we cannot use OBS's automerge feature anymore
and, consequently, we could "lose" a commit.

Example:
- let's assume prjA/pkg and prjB/pkg are "simple" (no fixed rev etc.)
  branches of prjC/pkg
- prjA/pkg: add a new patch to the spec file
- create SR prjA/pkg -> prjC/pkg (req1234)
- prjB/pkg: mark a file in the spec %files section list as %config
- create SR prjB/pkg -> prjC/pkg (req(5678)
- review req1234 and accept a review (not the request)
- review req5678 and accept a review (not the request)
- all bot related reviews are done for req1234 -> accept the request
- all bot related reviews are done for req5678 -> accept the request
  => the change from req1234 is lost (due to the "disabled" automerge)

Of course, one could also construct an example where a "hidden"
automerge can lead to issues...

Maybe we should track against which revision of a/the target package
a review was made and reopen the review if the revision changed
during the "whole" review process?

coolo:
I'm all for it - but that's a whole different story, the source revision has no real mapping in the API database. For status API we bind them to repo revisions, but request reviews are an old construct

PR #11268, opened 2021-06-22
Recover from sphinx errors when creating an attribute

danidoni:
Follows-up #11261


danidoni:
Closed in favor of #11286

PR #11275, opened 2021-06-23
Do not allow requests where the rev equals to 'upload'

### LINKED ISSUE #11273: Requests with rev=upload should be forbidden
rev=upload is basically a fake revision that can be altered before commit, requests with that "revision" don't make any sense.

@marcus-h pointed this out in https://github.com/openSUSE/open-build-service/pull/11263#issuecomment-866577739, but this is a problem independent of that PR

coolo:
The issue was discovered while adding source revision enforcement.

Fixes #11273, but I'm not really sure where to test this. Suggestions?

### DIFF
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -84,6 +84,7 @@ class BsRequestAction < ApplicationRecord
     errors.add(:source_package, 'is invalid package name') if source_package && !Package.valid_name?(source_package)
     errors.add(:target_project, 'is invalid project name') if target_project && !Project.valid_name?(target_project)
     errors.add(:source_project, 'is invalid project name') if source_project && !Project.valid_name?(source_project)
+    errors.add(:source_rev, 'should not be upload') if source_rev == 'upload'
 
     # TODO: to be continued
   end


src/api/app/models/bs_request_action.rb
```diff
@@ -84,6 +84,7 @@ def check_sanity
     errors.add(:source_package, 'is invalid package name') if source_package && !Package.valid_name?(source_package)
     errors.add(:target_project, 'is invalid project name') if target_project && !Project.valid_name?(target_project)
     errors.add(:source_project, 'is invalid project name') if source_project && !Project.valid_name?(source_project)
+    errors.add(:source_rev) if source_rev && source_review == 'upload'
```
hennevogel, line 87: no message?

hennevogel (changes requested):

hennevogel:
> but I'm not really sure where to test this. Suggestions?

Don't test this, there is barely any, let alone custom, logic involved.

marcus-h:
On 2021-06-23 04:27:48 -0700, Stephan Kulow wrote:
> diff --git a/src/api/app/models/bs_request_action.rb b/src/api/app/models/bs_request_action.rb
> index 77dbd5e9e78..9896f5ddbb1 100644
> --- a/src/api/app/models/bs_request_action.rb
> +++ b/src/api/app/models/bs_request_action.rb
> @@ -84,6 +84,7 @@ def check_sanity
>      errors.add(:source_package, 'is invalid package name') if source_package && !Package.valid_name?(source_package)
>      errors.add(:target_project, 'is invalid project name') if target_project && !Project.valid_name?(target_project)
>      errors.add(:source_project, 'is invalid project name') if source_project && !Project.valid_name?(source_project)
> +    errors.add(:source_rev) if source_rev && source_review == 'upload'

s/source_review/source_rev/ :)

(can't we directly write "... if source_rev == 'upload'"?)

coolo:
Added message, removed the extra check for nil - not adding a test

hennevogel (approved):

PR #11281, opened 2021-06-23
Index in_beta and in_rollout fields in users table

eduardoj:
These fields were not indexed before.

Fix Rails/AddColumnIndex RuboCop offenses.

Follow-up to #11270.

-----------------

### DIFF
--- a/src/api/db/migrate/20190704072437_add_column_in_rollout_to_users.rb
+++ b/src/api/db/migrate/20190704072437_add_column_in_rollout_to_users.rb
@@ -1,5 +1,5 @@
 class AddColumnInRolloutToUsers < ActiveRecord::Migration[5.2]
   def change
-    add_column :users, :in_rollout, :boolean, default: true, index: true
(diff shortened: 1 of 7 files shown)


src/api/db/migrate/20190704072437_add_column_in_rollout_to_users.rb
```diff
@@ -1,5 +1,5 @@
 class AddColumnInRolloutToUsers < ActiveRecord::Migration[5.2]
   def change
-    add_column :users, :in_rollout, :boolean, default: true, index: true
```
dmarcoux, line 3: So this was ignored, but where is the migration to add this index? I don't see it in this PR.

eduardoj: I forgot to add the file with the migration. Added now. Thanks!

dmarcoux (approved):

PR #11282, opened 2021-06-24
Do not rely on named scope chain for BsRequest::FindFor::Query

dmarcoux:
Since Rails 6.1, named scope chains will no longer leak the scope to class level querying methods.  As a result of this, the
"BsRequest::FindFor::Query#all" starts to perform some of its queries again on the unscoped relation instead of keeping them chained. This resulted in unexpected/wrong results.

See https://guides.rubyonrails.org/6_1_release_notes.html#active-record-notable-changes
Related to #9748

### DIFF
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -58,9 +58,8 @@ class BsRequest < ApplicationRecord
   scope :by_package_reviews, ->(package_ids) { where(reviews: { package: package_ids }) }
   scope :by_group_reviews, ->(group_ids) { where(reviews: { group: group_ids }) }
 
-  scope :for_user, ->(params) { BsRequest::FindFor::User.new(params).all }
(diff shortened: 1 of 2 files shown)


src/api/app/models/bs_request.rb
```diff
@@ -58,9 +58,8 @@ class BsRequest < ApplicationRecord
   scope :by_package_reviews, ->(package_ids) { where(reviews: { package: package_ids }) }
   scope :by_group_reviews, ->(group_ids) { where(reviews: { group: group_ids }) }
 
-  scope :for_user, ->(params) { BsRequest::FindFor::User.new(params).all }
```
dmarcoux, line 61: Those removed scopes weren't used, beside in `BsRequest::FindFor::Query`, so it's safe to remove them.

dmarcoux:
It's expected that the `next_rails` CI check isn't passing. There is still one failing spec to fix. It will happen in another PR.

hennevogel (approved):

PR #11284, opened 2021-06-24
Do not rely on named scope chains in Rails 6.1

### LINKED ISSUE #9748: Deprecation Warning on complex scoped relations

krauselukas:
Fixes #9748


dmarcoux:

vpereira (approved):

PR #11285, opened 2021-06-24
Split RSpec and Minitest in two next_rails CI jobs

dmarcoux:
This is needed because they need a different setup.


dmarcoux:

vpereira (approved):

PR #11287, opened 2021-06-25
Do no expand the upload rev

marcus-h:
When trying to expand the upload rev, a bogus tree files is written
(if the upload rev contains a _link file). For instance, a tree file
can look like this:

c157a79031e1c40f85931829bc5fc552  foo
f9079d5de256dd593c7eabdb29699e53  /LINK
upload  /LOCAL

Since the upload rev has no "real" srcmd5, the format of the tree
file is corrupted. Such a corrupted tree file can be used to generate
a <directory/> xml that has an <entry/> whose "name" attribute is the
empty string and whose "md5" attribute contains 32 "garbage" bytes
(that is, an illegal md5). For instance,

marcus@linux:~> curl http://localhost:5352/source/home:mallory/lnk2?expand=1
<directory name="lnk2" rev="4fc0e0e48934e51692aeec749600d854" vrev="10" srcmd5="4fc0e0e48934e51692aeec749600d854">
  <linkinfo project="home:mallory" package="lnk2" rev="2568453a6f3542d64649fdd50186cb28" srcmd5="2568453a6f3542d64649fdd50186cb28" lsrcmd5="6497434b763387d4daba84e582636df7"/>
  <entry name="" md5="upload  /LOCAL                  " error="No such file or directory"/>
  <entry name="foo" md5="c157a79031e1c40f85931829bc5fc552" size="4" mtime="1624438950"/>
</directory>
marcus@linux:~>

Such a <directory/> may confuse "broken" clients (since the error
attribute is present, clients have the chance to refuse such a
<directory/>).

(Note: in the curl call we do not expand the upload rev; the upload
rev is only expanded once to generate the broken tree file; the
broken tree file is then used to generate the broken <entry/> from
above)

In order to avoid this, simply do not expand the upload rev. Note
that if a link's rev or a passed in linkrev points to the upload rev,
the expansion fails because the revision returned by $getrev->(...)
is "strange" (see the code in handlelinks). Hence, it is sufficient
to check at the beginning of handlelinks if the passed revision is
the upload rev.

-----------------

### DIFF
--- a/src/backend/BSSrcServer/Link.pm
+++ b/src/backend/BSSrcServer/Link.pm
@@ -535,6 +535,7 @@ sub applylink {
 sub handlelinks {
   my ($rev, $files, $li) = @_;
 
+  return "bad srcmd5: $rev->{'srcmd5'}" unless $rev->{'srcmd5'} =~ /^[0-9a-f]{32}$/;
   my @linkinfo;
   my %seen;
   my $projid = $rev->{'project'};


mlschroe:
I'm not super happy about this, but noone is using the upload rev anyway...

mlschroe (approved):

mlschroe:
We should probably be more strict here, i.e. reject anything that's not a valid md5.

marcus-h:
On 2021-06-25 12:26:26 +0000, Michael Schroeder wrote:
> I'm not super happy about this, but noone is using the upload rev anyway...
>
Hmm ok. What about the following?

diff --git a/src/backend/BSSrcServer/Link.pm b/src/backend/BSSrcServer/Link.pm
index a1471bbbe5..eeb3cd0e61 100644
--- a/src/backend/BSSrcServer/Link.pm
+++ b/src/backend/BSSrcServer/Link.pm
@@ -541,6 +541,7 @@ sub handlelinks {
   my $packid = $rev->{'package'};
   my $linkrev = $rev->{'linkrev'};
   my $ignoreserviceerrors = $rev->{'ignoreserviceerrors'};
+  $rev->{'srcmd5'} = BSSrcrep::addmeta($projid, $packid, $files) if $rev->{'srcmd5'} eq 'upload';
   push @linkinfo, {'project' => $projid, 'package' => $packid, 'srcmd5' => $rev->{'srcmd5'}, 'rev' => $rev->{'rev'}};
   delete $rev->{'srcmd5'};
   delete $rev->{'linkrev'};

mlschroe:
I actually prefer your first solution that returns an error. As said, nobody uses the upload rev.

marcus-h:
On 2021-06-25 06:30:17 -0700, Michael Schroeder wrote:
> I actually prefer your first solution that returns an error. As said, nobody uses the upload rev.
>
Ok. I just made the check more strict (as you suggested) and did
a forced push.

PR #11291, opened 2021-06-25
Simplify package feature spec

hennevogel:
No need to visit tons of pages all the time. Or test
different kinds of validations in a feature spec.
Also refreshed the casettes.


danidoni (approved):

src/api/spec/features/webui/packages_spec.rb
```diff
@@ -374,27 +361,21 @@
       end
     end
 
-    describe 'in a project not owned by the user, eg. global namespace' do
+    describe 'in a project not owned by the user' do
       let(:admin_user) { create(:admin_user, :with_home) }
       let(:global_project) { create(:project, name: 'global_project') }
 
-      it 'as non-admin user' do
+      it 'forbidden as user' do
```
dmarcoux, line 368: This should be a spec for policies since it's only testing authorization. Anyway, it's already a bit better now, it's at least doing less.

hennevogel: true

PR #11296, opened 2021-06-28
Taking diff generation out of the HTTP request cycle

rubhanazeem:
Add diffs generation in a delayed job and use the cache option to fetch diffs from the backend.


src/api/app/models/bs_request_action.rb
```diff
@@ -245,11 +246,12 @@ def notify_params(ret = {})
...
 
   def contains_change?
-    sourcediff(nodiff: 1).present?
-  rescue BsRequestAction::Errors::DiffError
-    # if the diff can'be created we can't say
-    # but let's assume the reason for the problem lies in the change
-    true
+    spkg = Package.find_by_project_and_name(source_project, source_package)
+    tpkg = Package.find_by_project_and_name(target_project, target_package)
+
```
danidoni, line 251: `spkg` and `tpkg` variable names can be improved :thought_balloon:

src/api/app/views/webui/request/_sourcediff_tab.html.haml
```diff
@@ -7,43 +7,50 @@
...
-          .btn-group.float-right
-            %button.btn.btn-outline-secondary.expand-diffs{ data: { package: action[:spkg] } }
-              Expand all
-            %button.btn.btn-outline-secondary.collapse-diffs{ data: { package: action[:spkg] } }
-              Collapse all
+    - if refresh
+      .col-lg-12
+        .clearfix.mb-2.text-center
+          .btn.btn-outline-primary.cache-refresh{ title: 'Refresh results', onclick: "reloadRequestAction(#{index})" }
+            Crunching the latest data. Refresh again in few seconds
```
danidoni, line 14:
```suggestion
            Crunching the latest data. Refresh again in a few seconds
```

src/api/app/views/webui/request/_sourcediff_tab.html.haml
```diff
@@ -7,43 +7,50 @@
...
+                                                                                          package: action[:spkg] }
+            - else
+              .mb-2
+                %p.lead
+                  No source changes.
+        .col-lg-4
+          = render partial: 'webui/shared/buildresult_box',
+                  locals: { project: action[:sprj], package: action[:spkg], bs_request: bs_request, index: index }
+          - if sourcediff['issues'].present?
+            = render partial: 'issues_table', locals: { issues: sourcediff['issues'] }
```
danidoni, line 56: You can move a lot of this logic into the `RequestActionController` and maybe view helpers / decorators so the view is as simple as possible

src/api/app/models/bs_request_action.rb
```diff
@@ -37,6 +37,7 @@ class BsRequestAction < ApplicationRecord
   }
 
   before_validation :set_target_associations
+  after_save :cache_diffs
```
hennevogel, line 40: This would mean for every update of every attribute we request the diff. This might not be such a good idea. Generating a diff is an expensive operation, but so can be requesting it.

src/api/app/views/webui/request/_sourcediff_tab.html.haml
hennevogel: Please keep in mind that in the not so far future we are going for a re-design of this page which will probably include a lot of refactoring. So doing large refactorings now might be not the right time.

hennevogel (changes requested):

src/api/app/views/webui/request/_sourcediff_tab.html.haml
hennevogel: Please keep in mind that in the not so far future we are going for a re-design of this page which will probably include a lot of refactoring. So doing large refactorings now might be not the right time.

src/api/app/jobs/bs_request_action_webui_infos_job.rb
```diff
@@ -48,8 +49,18 @@ def for_a_superseded_request(superseded_request, bs_request_action, source_packa
       BsRequestAction::Differ::ForSource.new(
         bs_request_action: bs_request_action,
         source_package_names: source_package_names,
-        options: { superseded_bs_request_action: superseded_bs_request_action }
+        options: query.merge({ superseded_bs_request_action: superseded_bs_request_action })
       ).perform
     end
   end
+
+  def query
```
hennevogel, line 57: Shouldn't `skip_access_check` also be in here? Or alternatively, shouldn't all of this be on line 15?


And `query` is not such a nice name. It's not a query right?

src/api/app/jobs/bs_request_action_webui_infos_job.rb
```diff
@@ -48,8 +49,18 @@ def for_a_superseded_request(superseded_request, bs_request_action, source_packa
...
+    {
+      filelimit: nil,
+      tarlimit: nil,
```
hennevogel, line 60: Is `nil` the right option here?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -139,10 +139,13 @@ def show
...
     @diff_limit = params[:full_diff] ? 0 : nil
     @index = params[:index].to_i
-    @actions = @bs_request.webui_actions(filelimit: @diff_limit, tarlimit: @diff_limit, diff_to_superseded: @diff_to_superseded, diffs: true, action_id: params['id'].to_i)
+    @actions = @bs_request.webui_actions(filelimit: @diff_limit, tarlimit: @diff_limit, diff_to_superseded: @diff_to_superseded, diffs: true,
+                                         action_id: params['id'].to_i, cacheonly: 1)
     @action = @actions.find { |action| action[:id] == params['id'].to_i }
     @active = @action[:name]
     @not_full_diff = BsRequest.truncated_diffs?(@actions)
+    errors = @action[:sourcediff]&.map { |a| a[:error] }&.compact
+    @refresh = errors.present? && errors.any? { |e| e.include?('cache') }
```
hennevogel, line 148: This sounds a bit brittle to me. If there ever will be any other error message that contains this string...

hennevogel (changes requested):

src/api/app/views/webui/request/_sourcediff_tab.html.haml
danidoni: Hm, good point. Light refactoring is fine for me, just removing "the noise": those `files`, `diff_params` variables and some of those `if`'s because logic in haml is not fun...

src/api/app/models/bs_request_action.rb
rubhanazeem: Done. I've updated the names. I wasn't able to use the complete names because those variables are names of attributes as well.

src/api/app/views/webui/request/_sourcediff_tab.html.haml
rubhanazeem: Done.

src/api/app/views/webui/request/_sourcediff_tab.html.haml
rubhanazeem: We'll come back to this later when we start to refactor this part.

src/api/app/models/bs_request_action.rb
rubhanazeem: It's updated. I've moved it to `after_create`

src/api/app/jobs/bs_request_action_webui_infos_job.rb
rubhanazeem: I've removed this method. It's not needed anymore.

src/api/app/jobs/bs_request_action_webui_infos_job.rb
rubhanazeem: All the nil values are removed.

src/api/app/controllers/webui/request_controller.rb
rubhanazeem: I'm now using the whole error message to verify the correct error.

vpereira:
I removed the review-app label, because we add the branch name in the url path, and therefore any javascript query to an address starting with "/" goes to the wrong place. We should update the review-app to rewrite the requests adding this new path to the requests.

rubhanazeem:
> I removed the review-app label, because we add the branch name in the url path, and therefore any javascript query to an address starting with "/" goes to the wrong place. We should update the review-app to rewrite the requests adding this new path to the requests.

It was fixed in #11307

PR #11298, opened 2021-06-28
Report results for Broken and Unresolved builds

danidoni:
- [ ] Implement frontend part
- [ ] Implement backend part

### DIFF
--- a/src/api/app/jobs/report_to_scm_job.rb
+++ b/src/api/app/jobs/report_to_scm_job.rb
@@ -1,6 +1,11 @@
 # We don't properly capitalize SCM in the class name since CreateJob is doing `CLASS_NAME.to_s.camelize.safe_constantize`
 class ReportToScmJob < CreateJob
-  ALLOWED_EVENTS = ['Event::BuildFail', 'Event::BuildSuccess'].freeze
+  ALLOWED_EVENTS = [
+    'Event::BuildBroken',
+    'Event::BuildFail',
+    'Event::BuildSuccess',
+    'Event::BuildUnresolvable'
+  ].freeze
 
   queue_as :scm
 
--- a/src/api/app/models/event/build_broken.rb
+++ b/src/api/app/models/event/build_broken.rb
@@ -1,0 +1,12 @@
+module Event
+  class BuildBroken < Build
+    self.message_bus_routing_key = 'package.build_broken'
+    self.description = 'Package build is broken'
+
+    create_jobs :report_to_scm_job
+
+    def state
+      'broken'
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


danidoni:
[Report back to SCM build results 'broken' and 'unresolvable'](https://trello.com/c/g5Ud3aZw/1693-report-back-to-scm-build-results-broken-and-unresolvable)

dmarcoux:
[Report back to SCM when build results aren't `failed` and `succeeded`](https://trello.com/c/0N3jHq5M/2257-report-back-to-scm-when-build-results-arent-failed-and-succeeded)

src/api/app/jobs/report_to_scm_job.rb
```diff
@@ -1,6 +1,11 @@
 # We don't properly capitalize SCM in the class name since CreateJob is doing `CLASS_NAME.to_s.camelize.safe_constantize`
 class ReportToScmJob < CreateJob
-  ALLOWED_EVENTS = ['Event::BuildFail', 'Event::BuildSuccess'].freeze
+  ALLOWED_EVENTS = [
+    'Event::BuildBroken',
+    'Event::BuildFail',
+    'Event::BuildSuccess',
```
coolo, line 6: Note that a build can switch from success to fail and back to unchanged. So if you reported fail, you need to look at BuildUnchanged too

dmarcoux: So `Event::BuildUnchanged` is also to be reported as `success`... this is weird, but whatever, this is how it is.

https://github.com/openSUSE/open-build-service/blob/af9297d8d8fb79e14951f657a792474c53179e67/src/api/app/models/event/build_unchanged.rb#L4

coolo: Don't shoot the messenger :)

wineee:
It's a useful feature, is it still under development?

coolo:
If I may suggest an approach that might be a little easier to handle. There is a `RepoBuildFinished` event. Once that is in, send broken builds that weren't reported before (i.e. store it somewhere once someone subscribes the event).

Unresolvable state is unfortunately very temporary:
If you have package A, package B depending on A's last built version and then package C requiring B: C is unresolvable after A has built a new version, but once B has built (and refreshed the required version), C is fine again. People asking for unresolvable emails are surely not asking for these temporary states, but about actionable states.

wineee:
> If I may suggest an approach that might be a little easier to handle. There is a `RepoBuildFinished` event. Once that is in, send broken builds that weren't reported before (i.e. store it somewhere once someone subscribes the event).

Does the `RepoBuildFinished` event mean the build broken？ Doesn't build failure/success trigger this event?
（here are no :reason  payload_keys in  `RepoBuildFinished` ）

PR #11309, opened 2021-06-30
[backend] add a field in product spec for default repository name

adrianschroeter:
bsc#1187425

Allows zypper to guess the repository name a bit more reliable.

-----------------

### DIFF
--- a/src/backend/BSProductXML.pm
+++ b/src/backend/BSProductXML.pm
@@ -214,6 +214,7 @@ our $product = [
               'defaultlang',
               'datadir',
               'descriptiondir',
+	      'default_obs_repository_name', # needs to be a sync with the name in /distributions
               [ 'releasepackage' => 'name', 'flag', 'version', 'release' ],
               'distribution',
               [[ 'obsoletepackage' => '_content' ]],


mlandres:
@adrianschroeter; How does the element look like?


In bug#1153687#c9 you suggested to add definition of both,  hostname and platform:

  So I would recommend something like
   <default_repository_name>openSUSE_Tumbleweed</default_repository_name>
  Also a
   <default_obs_url>obs://build.opensuse.org</default_obs_url>
  may make sense, since in OBS the url syntax is define as 
  obs://$hostname/$project/...


What zypper needs to guess is the _final_URL_ for the $project repository.


So if in a obs://[$hostname/]$project passed to zypper the hostname is missing or "build.opensuse.org", we replace it by "https://download.opensuse.org/repositories/"

For $hostname we do not know, the user had to configure the obs_baseUrl in the zypper.conf because we can't know where it is.


If the .product file defines hints for zypper, the easiest for us would be: 

   <>https://download.opensuse.org/repositories/%PROJECT%/openSUSE_Leap_15.3<\>

so we can just replace the %PROJECT%.

PR #11313, opened 2021-06-30
Validate presence of scm_token for Token::Workflow

dmarcoux:
We always need an SCM token for a Token::Workflow record, so we should check that it's present. We don't want to valid its format or anything else for now.


dmarcoux:
[Validate presence of scm_token in Token::Workflow model](https://trello.com/c/LdSZ3UQr/1699-validate-presence-of-scmtoken-in-tokenworkflow-model)

krauselukas (approved):

PR #11314, opened 2021-06-30
Worker metrics

hennevogel:
There is lots of interesting information about the workers, job pool and schedulers in here.
    
Also simplifies the caching strategy a bit. It didn't make much sense before.
~~And improves `Rabbitmq` performance.~~

### DIFF
--- a/src/api/app/models/worker_status.rb
+++ b/src/api/app/models/worker_status.rb
@@ -3,8 +3,8 @@ class WorkerStatus
 
   class << self
     def hidden
-      mydata = Rails.cache.read('workerstatus')
-      ws = Nokogiri::XML(mydata || Backend::Api::BuildResults::Worker.status).root
+      mydata = Rails.cache.fetch('workerstatus') { Backend::Api::BuildResults::Worker.status }
+      ws = Nokogiri::XML(mydata).root
       # remove information about projects which are not visible to the current user
       hidden_projects(ws)
     end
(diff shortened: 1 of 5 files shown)


src/api/lib/rabbitmq_bus.rb
```diff
@@ -2,7 +2,7 @@ class RabbitmqBus
   cattr_accessor :connection, :exchange, :failed
 
   def self.send_to_bus(channel, data)
-    channel = "#{Configuration.amqp_namespace}.#{channel}"
+    channel = "#{CONFIG['amqp_namespace']}.#{channel}"
```
vpereira, line 5: why is that better?

src/api/app/models/worker_status.rb
```diff
@@ -37,29 +37,73 @@ def project_names(prjs)
...
       @squeues.each_pair { |key, value| StatusHistory.create(time: @mytime, key: key, value: value) }
     end
 
-    backend_status
+    send_to_bus
   end
 
   private
 
+  def send_to_bus
```
vpereira, line 61: could break this method in a service?

src/api/.rubocop_todo.yml
```diff
@@ -190,6 +190,7 @@ Metrics/ClassLength:
     - 'app/models/staging/workflow.rb'
     - 'app/models/user.rb'
     - 'app/models/user_ldap_strategy.rb'
+    - 'app/models/worker_status.rb'
```
vpereira, line 193: why?

src/api/lib/rabbitmq_bus.rb
vpereira, line 5: That's because there are 3 SQL queries are the result of calling `first` 3 times, inside method_missing. However the method `amqp_namespace` exists. We should fix that. 

The method:

```
  def amqp_namespace
    CONFIG['amqp_namespace'] || 'opensuse.obs'
  end
```
shouldn't generate any SQL query other than calling `CONFIG['amqp_namespace']`

src/api/app/models/worker_status.rb
hennevogel: After sleeping on it it might be more clever to do this in a job with another frequency. Every 30 seconds might be a bit too much.

src/api/lib/rabbitmq_bus.rb
hennevogel: @eduardoj suggested to go for [`config_for`](https://guides.rubyonrails.org/configuring.html#custom-configuration) instead of all of this custom code. I dropped the commit.

eduardoj (approved):

hennevogel:
codeclimate want's me to break up `send_job_metrics` but I don't see why I should. Can we mark this as wontfix?

eduardoj:
> codeclimate want's me to break up `send_job_metrics` but I don't see why I should. Can we mark this as wontfix?

Fine for me.

PR #11318, opened 2021-07-01
Implement architectures and repositories filters for OBS workflows

dmarcoux:
We are going to report back the build results to the SCM according to the architectures and repositories defined in the filters of OBS workflows.

While we're happy with the code as it is now, there are still some potential improvements like validating values provided to the filters. As noted in comments, we could also use a query object for some methods. We could gather all errors together before raising if any error is present. All of this is nice to have, but not a blocker to delivering this feature. We're going to tackle as much as possible in the remaining days of the deliverable.

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -14,42 +14,11 @@ class Token::Workflow < Token
     workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_extractor_payload: scm_extractor_payload, token: self).call
     workflows.each do |workflow|
       workflow.steps.each do |step|
-        run_step_and_report(step, scm_extractor_payload, scm_token)
+        step.call({ workflow_filters: workflow.filters })
       end
     end
   rescue Octokit::Unauthorized, Gitlab::Error::Unauthorized => e
     raise Token::Errors::SCMTokenInvalid, e.message
-  end
-
-  private
-
-  def run_step_and_report(step, scm_extractor_payload, scm_token)
-    package_from_step = step.call
-
-    set_subscription(package_from_step, scm_extractor_payload)
-
-    Project.get_by_name(step.target_project_name).repositories.each do |repository|
-      # TODO: Fix n+1 queries
-      repository.architectures.each do |architecture|
-        # We cannot report multibuild flavors here... so they will be missing from the initial report
-        SCMStatusReporter.new({ project: step.target_project_name, package: step.target_package_name, repository: repository.name, arch: architecture.name },
-                              scm_extractor_payload, scm_token).call
-      end
-    end
-  end
-
-  def set_subscription(package_from_step, scm_extractor_payload)
-    ['Event::BuildFail', 'Event::BuildSuccess'].each do |build_event|
-      # TODO: Deal with old EventSubscription (this can happen when someone pushes a new commit to a PR/branch, then we only want to report to the latest commit)
-      EventSubscription.create!(eventtype: build_event,
-                                receiver_role: 'reader', # We pass a valid value, but we don't need this.
-                                user: user,
-                                channel: 'scm',
-                                enabled: true,
-                                token: self,
-                                package: package_from_step,
-                                payload: scm_extractor_payload)
-    end
   end
 end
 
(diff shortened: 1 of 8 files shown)


krauselukas:
[Limit reporting of build results via filters for architectures and repositories](https://trello.com/c/HXYMXh74/1696-limit-reporting-of-build-results-via-filters-for-architectures-and-repositories)

src/api/app/models/token/workflow.rb
```diff
@@ -15,14 +15,34 @@ def call(options)
...
-        run_step_and_report(step, scm_extractor_payload, scm_token)
+        run_step_and_report(step, scm_extractor_payload, scm_token, workflow.filters)
       end
     end
   end
 
   private
 
-  def run_step_and_report(step, scm_extractor_payload, scm_token)
+  def workflow_repositories(target_project_name, filters)
```
hennevogel, line 25: Isn't this more like a method of `BranchPackageStep`?

dmarcoux: Even before the changes from this PR, `run_step_and_report` was already doing too much. I agree with you that the code should be moved to the step.

dmarcoux:
Regarding the CodeClimate issue, this is going to be addressed in another PR. Using a query object should reduce the number of methods in the `BranchPackageStep` model.

src/api/app/models/workflow.rb
```diff
@@ -24,6 +38,26 @@ def steps
...
   end
 
+  def filters
+    filters = {}
+
+    return filters if supported_filters.blank?
+
+    SUPPORTED_FILTERS.each do |filter|
+      SUPPORTED_FILTER_TYPES.each do |filter_type|
+        # The filter type might not be present... so in that case, we go the next
```
danidoni, line 48:
```suggestion
        # The filter type might not be present... so in that case, we go to the next
```
:thought_balloon:

danidoni (approved):

src/api/app/models/workflow.rb
dmarcoux: Thanks, this is addressed.

src/api/app/models/workflow.rb
```diff
@@ -1,12 +1,26 @@
...
     @scm_extractor_payload = scm_extractor_payload
     @token = token
+
     raise Token::Errors::InvalidWorkflowStepDefinition, "Invalid workflow step definition: #{errors.to_sentence}" unless
       valid?
+
+    # Filters aren't mandatory in a workflow
+    return unless @workflow.key?(:filters)
+
+    raise Token::Errors::UnsupportedWorkflowFilters, "Unsupported filters: #{@unsupported_filters.keys.to_sentence}" if unsupported_filters?
```
hennevogel, line 18: You can use validations outside or active record too. See https://api.rubyonrails.org/classes/ActiveModel/Validations.html

src/api/app/models/workflow.rb
```diff
@@ -24,6 +38,26 @@ def steps
...
+    return filters if supported_filters.blank?
+
+    SUPPORTED_FILTERS.each do |filter|
+      SUPPORTED_FILTER_TYPES.each do |filter_type|
+        # The filter type might not be present... so in that case, we go to the next
+        next unless (filter_values = supported_filters.dig(filter, filter_type))
+
+        filters[filter] = { "#{filter_type}": filter_values }
+
+        # As soon as a supported filter type is present, we get out of the loop since the following filter types have a lower precedence
```
hennevogel, line 53: Are they? Am I not free in the order I write them into the `workflow.yml`?

hennevogel: Ah this is the order of `SUPPORTED_FILTER_TYPES`. I find this a bit too implicit.

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -160,6 +172,30 @@ def create_or_update_subscriptions(branched_package)
           subscription.update!(payload: @scm_extractor_payload)
         end
       end
+
+      # TODO: This could be in a query object
+      def workflow_repositories(target_project_name, filters)
+        repositories = Project.get_by_name(target_project_name).repositories
```
hennevogel, line 178: https://github.com/openSUSE/open-build-service/wiki/Links#instantiation-returns-string

src/api/app/models/workflow.rb
dmarcoux: Yes, this is what I had in mind when writing this in the PR description: `We could gather all errors together before raising if any error is present.`

PR #11320, opened 2021-07-01
Sanitize simple_format

hennevogel:
simple_format marks output as html safe. All of this data includes user input.
Through XML parsing we even store HTML in the database.

```ruby
Xmlhash.parse('<project name="hans"><description>&lt;h2&gt;Hello World&lt;/h2&gt;</description></project>')
=> {"name"=>"hans", "description"=>"<h2>Hello World</h2>"}
```

So better sanitize the simple_format output.

### DIFF
--- a/src/api/app/views/layouts/webui/_flash.html.haml
+++ b/src/api/app/views/layouts/webui/_flash.html.haml
@@ -19,7 +19,7 @@
           = link_to('more info', 'javascript:void(0)', class: 'btn-more alert-link')
           .more_info.d-none
             .more-info-content
-              = simple_format(@more_info)
+              = sanitize(simple_format(@more_info), tags: %w[br p])
         -# haml-lint:enable InstanceVariables
         %button.close{ type: 'button', 'data-dismiss': 'alert', 'aria-label': 'Close' }
           %span{ 'aria-hidden': 'true' }
--- a/src/api/app/views/webui/patchinfo/show.html.haml
+++ b/src/api/app/views/webui/patchinfo/show.html.haml
@@ -30,7 +30,7 @@
       .card.mb-3
         %h5.card-header Message
         .card-body
-          = simple_format @patchinfo.message
+          = sanitize(simple_format(@patchinfo.message), tags: %w[br p])
 
   .col-md-6
     .card.mb-3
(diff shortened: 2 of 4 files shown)


danidoni (approved):

eduardoj (approved):

dmarcoux (changes requested):
Does it really change something? Because according to the documentation of [simple_format](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format), this helper is already sanitizing the input. You need to explicitly set the `sanitize` option to `false` to not sanitize the input.

hennevogel:
> Does it really change something?

ARGL I forgot to append the allowed tags. Sorry, doing too many things at once...

coolo:
Now that asks for a DRY approach

dmarcoux:
So this change is only to sanitize `br` and `p` HTML tags, why? Is this really a problem if those 2 tags are not sanitized?

coolo:
the tags are the _allowed_ tags - they are not stripped. Basically Henne is after `<h2>` and co. But why `<p>` not `<div>`?

dmarcoux:
> the tags are the _allowed_ tags - they are not stripped.

:facepalm: Yes, I was checking that right now... hehehe

hennevogel:
> But why <p> not <div>?

Because `<p>` is the tag for paragraphs in `simple_format`

hennevogel:
> Now that asks for a DRY approach

I would rather have this explicit so people think about the consequences of this.

eduardoj:
> Now that asks for a DRY approach

`haml_lint` is reporting four issues. All of them are this error:
```
[W] RuboCop: Style/PercentLiteralDelimiters: `%w`-literals should be delimited by `[` and `]`.
```

No DRY approach asked, yet, IMHO.

dmarcoux:
> > Now that asks for a DRY approach
> 
> I would rather have this explicit so people think about the consequences of this.

It's easy to forget to have a tag. I prefer having an helper for this.

dmarcoux (approved):
I prefer to have an helper for this, but I won't fight for it.

eduardoj (approved):

src/api/app/views/layouts/webui/_flash.html.haml
```diff
@@ -19,7 +19,7 @@
           = link_to('more info', 'javascript:void(0)', class: 'btn-more alert-link')
           .more_info.d-none
             .more-info-content
-              = simple_format(@more_info)
+              = sanitize(simple_format(@more_info), tags: %w[br p])
```
vpereira, line 22: now it makes sense :+1:

PR #11321, opened 2021-07-01
raise if yaml file doesn't exist

### LINKED ISSUE #11240: Setting an GitHub -> OBS integration with no workflows file in the project root fails
# Issue Description
Setting an GitHub -> OBS integration with no `.obs/workflows.yml` file in the project root fails. It raises an `no implicit conversion of Array into String` error in our [errbit](https://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/60b5459484bf4607a94657f7) instance and GitHub's Continuous Integration status reports that `Continuous integration has not been set up`.

Expected Result
============
OBS raising a meaningful error when there is no yml file in the main branch.

How to Reproduce
===============
1. Set up the GitHub integration as explained in [here](https://openbuildservice.org/2021/05/31/scm-integration/) but do not create any `.obs/workflows.yml` file in the repository.
2. Create a PR on the repo.
3. See the `Continuous integration has not been set up` message.
4. See an error in [Errbit](https://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/60b5459484bf4607a94657f7) about `no implicit conversion of Array into String`.

vpereira:
Fixes #11240

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -10,7 +10,10 @@ class Token::Workflow < Token
     return unless extractor.allowed_event_and_action?
 
     scm_extractor_payload = extractor.call
-    yaml_file = Workflows::YAMLDownloader.new(scm_extractor_payload).call
+    yaml_downloader = Workflows::YAMLDownloader.new(scm_extractor_payload)
+    yaml_file = yaml_downloader.call
+    raise Token::Errors::NonExistentWorkflowsFile, yaml_downloader.errors.join('\n') if yaml_file.blank?
+
     workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_extractor_payload: scm_extractor_payload, token: self).call
 
     workflows.each do |workflow|
--- a/src/api/app/models/token/errors.rb
+++ b/src/api/app/models/token/errors.rb
@@ -4,4 +4,8 @@ module Token::Errors
   class NoReleaseTargetFound < APIError
     setup 404
   end
+
+  class NonExistentWorkflowsFile < APIError
+    setup 404
+  end
 end
--- a/src/api/app/services/workflows/yaml_downloader.rb
+++ b/src/api/app/services/workflows/yaml_downloader.rb
@@ -19,6 +19,7 @@ module Workflows
       Down.download(download_url, max_size: MAX_FILE_SIZE)
     rescue Down::Error => e
       @errors << e.message
+      nil
     end
 
     def download_url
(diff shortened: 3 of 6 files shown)


src/api/app/models/token/workflow.rb
```diff
@@ -11,6 +11,8 @@ def call(options)
 
     scm_extractor_payload = extractor.call
     yaml_file = Workflows::YAMLDownloader.new(scm_extractor_payload).call
+    raise 'workflows.yml is missing' if yaml_file.blank?
```
dmarcoux, line 14: When this happens, do we get the error in the webhook response on the SCM? If not, this won't help out users to know what is wrong.

vpereira:

src/api/app/models/token/workflow.rb
vpereira: yes https://github.com/openSUSE/open-build-service/pull/11321/files#diff-184bebd6d68042ab1a0659db7e98ad054cf2384141156b5c0d2f6ef9eed6931aR43

src/api/app/models/token/workflow.rb
```diff
@@ -10,7 +10,10 @@ def call(options)
     return unless extractor.allowed_event_and_action?
 
     scm_extractor_payload = extractor.call
-    yaml_file = Workflows::YAMLDownloader.new(scm_extractor_payload).call
+    yaml_downloader = Workflows::YAMLDownloader.new(scm_extractor_payload)
+    yaml_file = yaml_downloader.call
+    raise Token::Errors::NonExistentWorkflowsFile, yaml_downloader.errors.join('\n') if yaml_file.blank?
```
krauselukas, line 15: @vpereira the exception you've created wasnt raised in my tests, since `yaml_file` is not blank when the `workflow.yaml` doesnt exist, `yaml_file` contains `["404 Not Found"]`

krauselukas (changes requested):

src/api/app/models/token/workflow.rb
hennevogel: It would also be nice if you would include a message so customers know what they can do.

vpereira: @hennevogel we are using the error message coming from `Down`.. Do you want a static message, like 'workflows.yml couldn't be found' ?

vpereira: @krauselukas please could you look again?

vpereira: @hennevogel please take a look again

krauselukas: @vpereira okay, works now

dmarcoux: What would be an example of an error message coming from `Down`? It might already be good enough.

krauselukas (approved):
Fine for me :)

src/api/app/models/token/workflow.rb
krauselukas: @dmarcoux so in my tests the error message from down was the simple "404 Not Found"
with the exception class from victor it looks like this ```<status code="non_existent_workflows_file">
  <summary>404 Not Found</summary>
</status>```

PR #11322, opened 2021-07-01
Do not run the workflow steps if some are unsupported

danidoni:
This change aborts the whole step running when there is an unsupported step.

This is how it will respond if there are unsupported specs:

image

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -12,21 +12,18 @@ class Token::Workflow < Token
     scm_extractor_payload = extractor.call
     yaml_file = Workflows::YAMLDownloader.new(scm_extractor_payload, token: self).call
     workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_extractor_payload: scm_extractor_payload, token: self).call
-
     workflows.each do |workflow|
       workflow.steps.each do |step|
         run_step_and_report(step, scm_extractor_payload, scm_token)
-      rescue Octokit::Unauthorized, Gitlab::Error::Unauthorized => e
-        raise Token::Errors::SCMTokenInvalid, e.message
       end
     end
+  rescue Octokit::Unauthorized, Gitlab::Error::Unauthorized => e
+    raise Token::Errors::SCMTokenInvalid, e.message
   end
 
   private
 
   def run_step_and_report(step, scm_extractor_payload, scm_token)
-    raise 'Invalid workflow step definition' unless step.valid?
-
     package_from_step = step.call
 
     raise "We couldn't branch your package" unless package_from_step
--- a/src/api/app/models/workflow.rb
+++ b/src/api/app/models/workflow.rb
@@ -5,20 +5,54 @@ class Workflow
     @workflow = workflow
     @scm_extractor_payload = scm_extractor_payload
     @token = token
+    raise Token::Errors::InvalidWorkflowStepDefinition, "Invalid workflow step definition: #{errors.to_sentence}" unless
+      valid?
+  end
+
+  def valid?
+    unsupported_steps.none? && !invalid_steps?
   end
 
   def steps
-    steps = []
-    return steps if @workflow['steps'].blank?
-
-    @workflow['steps'].each do |step|
-      step.each do |step_name, step_instructions|
-        next if SUPPORTED_STEPS[step_name].blank?
-
-        new_step = SUPPORTED_STEPS[step_name].new(step_instructions: step_instructions, scm_extractor_payload: @scm_extractor_payload, token: @token)
-        steps << new_step if new_step.allowed_event_and_action?
-      end
+    @steps ||= workflow_steps.each_with_object([]) do |step_definition, acc|
+      step_definition
+        .select { |step_name, _| SUPPORTED_STEPS.key?(step_name) }
+        .map { |step_name, step_instructions| initialize_step(step_name, step_instructions) }
+        .select(&:allowed_event_and_action?)
+        .select { |new_step| acc << new_step }
+      acc
     end
-    steps
+  end
+
+  def errors
+    unsupported_steps.each_with_object([]) do |step_definition, acc|
+      step_definition.each do |step_name, _|
+        acc << "'#{step_name}' is not a supported step"
+      end
+      acc
+    end
+  end
+
+  private
+
+  def invalid_steps?
+    steps.reject(&:valid?).any?
+  end
+
+  def workflow_steps
+    @workflow.fetch('steps', {})
+  end
+
+  def unsupported_steps
+    @unsupported_steps ||= workflow_steps.each_with_object([]) do |step_definition, acc|
+      rejected_steps = step_definition.reject { |step_name, _| SUPPORTED_STEPS.key?(step_name) }
+      rejected_steps.empty? ? acc : acc << rejected_steps
+    end
+  end
+
+  def initialize_step(step_name, step_instructions)
+    SUPPORTED_STEPS[step_name].new(step_instructions: step_instructions,
+                                   scm_extractor_payload: @scm_extractor_payload,
+                                   token: @token)
   end
 end
--- a/src/api/spec/models/token/workflow_spec.rb
+++ b/src/api/spec/models/token/workflow_spec.rb
@@ -121,7 +121,7 @@ RSpec.describe Token::Workflow, vcr: true do
       end
 
       it 'raises an "Invalid workflow step definition" error' do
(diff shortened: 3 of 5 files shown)


danidoni:
[Workflows with non-valid steps shouldn't run at all](https://trello.com/c/mY7sE22C/1697-workflows-with-non-valid-steps-shouldnt-run-at-all)

vpereira:

src/api/app/services/workflows/step_iterator.rb
```diff
@@ -0,0 +1,26 @@
+module Workflows
```
dmarcoux, line 1: I don't see the point in having this class. It's increasing complexity.

danidoni: No, it's not, in fact it's helping me reducing the complexity of `Token::Workflow#call` method. CodeClimate whined about it before.

dmarcoux: It's a whole new file to only check if all workflows are valid by running `raise (...) workflows.reject(&:valid?).any?`. Everything else is boilerplate code.

danidoni: Okay, let me see what I can do

src/api/app/models/token/workflow.rb
```diff
@@ -10,23 +10,35 @@ def call(options)
     return unless extractor.allowed_event_and_action?
 
     scm_extractor_payload = extractor.call
-    yaml_file = Workflows::YAMLDownloader.new(scm_extractor_payload).call
+    yaml_downloader = Workflows::YAMLDownloader.new(scm_extractor_payload)
```
dmarcoux, line 13: This is undoing changes from #11338. Please revert your changes.

src/api/app/models/token/workflow.rb
```diff
@@ -10,23 +10,35 @@ def call(options)
...
     end
+  rescue Octokit::Unauthorized, Gitlab::Error::Unauthorized => e
+    raise Token::Errors::SCMTokenInvalid, e.message
   end
 
   private
 
-  def run_step_and_report(step, scm_extractor_payload, scm_token)
-    raise 'Invalid workflow step definition' unless step.valid?
+  def invalid_workflow_error!
```
dmarcoux, line 33: Why not use the content of this method directly? Having a method for a one liner doesn't improve the code.

Same thing for `invalid_workflows?`.

src/api/app/models/token/workflow.rb
```diff
@@ -10,23 +10,35 @@ def call(options)
...
+    invalid_workflow_error! if invalid_workflows?(workflows)
 
     workflows.each do |workflow|
       workflow.steps.each do |step|
         run_step_and_report(step, scm_extractor_payload, scm_token)
-      rescue Octokit::Unauthorized, Gitlab::Error::Unauthorized => e
-        raise Token::Errors::SCMTokenInvalid, e.message
       end
     end
+  rescue Octokit::Unauthorized, Gitlab::Error::Unauthorized => e
```
dmarcoux, line 27: What makes this change necessary?

src/api/app/models/token/workflow.rb
danidoni: Ouch, I messed up the rebase.

src/api/app/models/token/workflow.rb
danidoni: We're checking for the errors on each iteration. It's easier to reason about the loops without the rescue inside, and it does the same.

src/api/app/models/workflow.rb
```diff
@@ -7,18 +7,57 @@ def initialize(workflow:, scm_extractor_payload:, token:)
...
+  def unsupported_steps?
+    unsupported_steps.any?
+  end
+
+  def invalid_steps?
+    extract_workflow_steps.reject(&:valid?).any?
+  end
+
+  def unsupported_steps
+    @workflow['steps'].each_with_object([]) do |step_definition, acc|
```
dmarcoux, line 41: We should memoize the result of this method since it's used in multiple methods throughout this file.

Something like `@unsupported_steps ||= (the whole method code)`

src/api/app/models/workflow.rb
```diff
@@ -7,18 +7,57 @@ def initialize(workflow:, scm_extractor_payload:, token:)
...
+      rejected_steps = step_definition.reject { |step_name, _| SUPPORTED_STEPS.key?(step_name) }
+      rejected_steps.empty? ? acc : acc << rejected_steps
+    end
+  end
+
+  def extract_workflow_steps
+    @workflow['steps'].each_with_object([]) do |step_definition, acc|
```
dmarcoux, line 48: Same as above. We should memoize the result of this method since it's used in multiple methods throughout this file.

Something like `@steps ||= (the whole method code)`

src/api/app/models/token/workflow.rb
```diff
@@ -12,21 +12,21 @@ def call(options)
     scm_extractor_payload = extractor.call
     yaml_file = Workflows::YAMLDownloader.new(scm_extractor_payload).call
     workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_extractor_payload: scm_extractor_payload, token: self).call
+    raise Token::Errors::InvalidWorkflowStepDefinition, 'Invalid workflow step definition' if
```
dmarcoux, line 15: This error should be raised in `Workflow`. I would raise this error when workflows are initialized, so in the `initialize` method of the `Workflow` class. In the error message, we should also use the `errors` method from `Workflow` to help users identify what needs to be done. `Invalid workflow step definition` doesn't help much.

src/api/app/models/workflow.rb
danidoni: Yeah, this is something I know I have to improve, but I haven't, yet.

src/api/app/models/workflow.rb
```diff
@@ -7,18 +7,57 @@ def initialize(workflow:, scm_extractor_payload:, token:)
...
+      step_definition
+        .select { |step_name, _| SUPPORTED_STEPS.key?(step_name) }
+        .map { |step_name, step_instructions| build_step(step_name, step_instructions) }
```
dmarcoux, line 51: At first, I thought this was somehow related to some build results due to `build_step`... although we're totally not in that context in this class. I think the method isn't needed, we could do `SUPPORTED_STEPS[step_name].new(...)` here directly.

src/api/app/models/workflow.rb
danidoni: I've made some improvements in a separate commit

src/api/app/models/workflow.rb
```diff
@@ -5,20 +5,63 @@ def initialize(workflow:, scm_extractor_payload:, token:)
     @workflow = workflow
     @scm_extractor_payload = scm_extractor_payload
     @token = token
+    @unsupported_steps = extract_unsupported_steps
```
dmarcoux, line 8: The instance variable and the method can have the same name:

```suggestion
    @unsupported_steps = unsupported_steps
```

src/api/spec/models/workflow_spec.rb
```diff
@@ -37,4 +51,59 @@
...
+
+    context 'with several unsupported steps' do
+      let(:yaml) do
+        { 'steps' => [{ 'unsupported_step' => {} },
+                      { 'unsupported_step' => {} },
+                      { 'branch_package' => {} }] }
+      end
+
+      it 'has several errors' do
+        expect { subject }.to raise_error("Invalid workflow step definition: 'unsupported_step' is not a supported step.\n'unsupported_step' is not a supported step.")
```
dmarcoux, line 105: I believe this might solve your failing spec.

```suggestion
        expect { subject }.to raise_error(Token::Errors::InvalidWorkflowStepDefinition, "Invalid workflow step definition: 'unsupported_step' is not a supported step.\n'unsupported_step' is not a supported step.")
```

src/api/spec/models/token/workflow_spec.rb
```diff
@@ -121,7 +121,7 @@
       end
 
       it 'raises an "Invalid workflow step definition" error' do
-        expect { subject }.to raise_error('Invalid workflow step definition')
+        expect { subject }.to raise_error('Invalid workflow step definition: ')
```
dmarcoux, line 124: This is kind of weird due to the extra colon, but it happens only since the steps are empty. I would say it's fine.

dmarcoux (approved):

src/api/spec/models/token/workflow_spec.rb
danidoni: We can improve later :+1:

PR #11325, opened 2021-07-02
SCM integration - Support for private instances (Github and Gitlab)

### LINKED ISSUE #11198: SCM integration doesn't support gitlab private instances
Based on this code https://github.com/openSUSE/open-build-service/blob/master/src/api/app/services/workflows/yaml_downloader.rb#L28, we just support gitlab.com, and not i.e gitlab.suse.de

rubhanazeem:
Remove hardcoded URLs and pass the hostname in the payload to generate URLs.

Fixes #11198

### DIFF
--- a/src/api/app/services/scm_status_reporter.rb
+++ b/src/api/app/services/scm_status_reporter.rb
@@ -11,14 +11,14 @@ class SCMStatusReporter
 
   def call
     if github?
-      github_client = Octokit::Client.new(access_token: @scm_token)
+      github_client = Octokit::Client.new(access_token: @scm_token, api_endpoint: @event_subscription_payload[:api_endpoint])
       # https://docs.github.com/en/rest/reference/repos#create-a-commit-status
       github_client.create_status(@event_subscription_payload[:target_repository_full_name],
                                   @event_subscription_payload[:commit_sha],
                                   @state,
                                   status_options)
     else
-      gitlab_client = Gitlab.client(endpoint: 'https://gitlab.com/api/v4',
+      gitlab_client = Gitlab.client(endpoint: "#{@event_subscription_payload[:api_endpoint]}/api/v4",
                                     private_token: @scm_token)
       # https://docs.gitlab.com/ce/api/commits.html#post-the-build-status-to-a-commit
       gitlab_client.update_commit_status(@event_subscription_payload[:project_id],
--- a/src/api/app/services/trigger_controller_service/scm_extractor.rb
+++ b/src/api/app/services/trigger_controller_service/scm_extractor.rb
@@ -38,6 +38,12 @@ module TriggerControllerService
     end
 
     def github_extractor_payload
+      host = URI.parse(@payload.dig('sender', 'url')).host
+      api_endpoint = if host.start_with?('api.github.com')
+                       "https://#{host}"
+                     else
+                       "https://#{host}/api/v3/"
+                     end
       {
         scm: 'github',
         commit_sha: @payload.dig('pull_request', 'head', 'sha'),
@@ -47,7 +53,8 @@ module TriggerControllerService
         action: @payload['action'], # TODO: Names may differ, maybe we need to find our own naming (defer to service?)
         source_repository_full_name: @payload.dig('pull_request', 'head', 'repo', 'full_name'),
         target_repository_full_name: @payload.dig('pull_request', 'base', 'repo', 'full_name'),
-        event: @event
+        event: @event,
+        api_endpoint: api_endpoint
       }.with_indifferent_access
     end
 
@@ -66,8 +73,7 @@ module TriggerControllerService
         project_id: @payload.dig('project', 'id'),
         path_with_namespace: @payload.dig('project', 'path_with_namespace'),
         event: @event,
-        scheme: uri.scheme,
-        host: uri.host
+        api_endpoint: "#{uri.scheme}://#{uri.host}"
       }.with_indifferent_access
     end
   end
--- a/src/api/app/services/trigger_controller_service/scm_extractor.rb
+++ b/src/api/app/services/trigger_controller_service/scm_extractor.rb
@@ -52,10 +52,12 @@ module TriggerControllerService
     end
 
     def gitlab_extractor_payload
+      http_url = @payload.dig('project', 'http_url')
+      uri = URI.parse(http_url)
       {
         scm: 'gitlab',
         object_kind: @payload['object_kind'],
-        http_url: @payload.dig('project', 'http_url'),
+        http_url: http_url,
         commit_sha: @payload.dig('object_attributes', 'last_commit', 'id'),
         pr_number: @payload.dig('object_attributes', 'iid'),
         source_branch: @payload.dig('object_attributes', 'source_branch'),
@@ -63,7 +65,9 @@ module TriggerControllerService
         action: @payload.dig('object_attributes', 'action'), # TODO: Names may differ, maybe we need to find our own naming (defer to service?)
         project_id: @payload.dig('project', 'id'),
         path_with_namespace: @payload.dig('project', 'path_with_namespace'),
-        event: @event
+        event: @event,
+        scheme: uri.scheme,
+        host: uri.host
       }.with_indifferent_access
     end
   end
--- a/src/api/app/services/workflows/yaml_downloader.rb
+++ b/src/api/app/services/workflows/yaml_downloader.rb
@@ -23,7 +23,7 @@ module Workflows
       when 'github'
         "https://raw.githubusercontent.com/#{@scm_payload[:target_repository_full_name]}/#{@scm_payload[:target_branch]}/.obs/workflows.yml"
       when 'gitlab'
-        "https://gitlab.com/#{@scm_payload[:path_with_namespace]}/-/raw/#{@scm_payload[:target_branch]}/.obs/workflows.yml"
+        "#{@scm_payload[:scheme]}://#{@scm_payload[:host]}/#{@scm_payload[:path_with_namespace]}/-/raw/#{@scm_payload[:target_branch]}/.obs/workflows.yml"
       end
     end
   end
--- a/src/api/app/services/workflows/yaml_downloader.rb
+++ b/src/api/app/services/workflows/yaml_downloader.rb
@@ -2,8 +2,9 @@ module Workflows
   class YAMLDownloader
     MAX_FILE_SIZE = 1024 * 1024 # 1MB
 
-    def initialize(scm_payload)
+    def initialize(scm_payload, token:)
       @scm_payload = scm_payload
+      @token = token
     end
 
     def call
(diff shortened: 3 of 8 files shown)


src/api/app/controllers/trigger_workflow_controller.rb
```diff
@@ -34,6 +34,10 @@ def event
     @github_event || @gitlab_event
   end
 
+  def github_enterprise_host
+    request.headers['X-GitHub-Enterprise-Host']
```
dmarcoux, line 38: There is no way to get the URL from the webhook payload? Because with this implementation, it's creeping up in many different classes...

rubhanazeem: The payloads are very similar in both cases (enterprise/public). Only headers contain this information.

dmarcoux: I had a look at GitHub's documentation on webhook events and payloads and just as an example for the `push` event, there are multiple fields with the URL in the payload: https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push.

Same story for the `pull_request` event. It seems to be the case for all events.

krauselukas (commented):
Hey @rubhanazeem seems like you still using the hardcoded version of the gitlab url for the Status Reporter https://github.com/openSUSE/open-build-service/blob/93f6dcc696774ec8e344347c9af8fa577278894e/src/api/app/services/scm_status_reporter.rb#L25

src/api/app/controllers/trigger_workflow_controller.rb
rubhanazeem: I see. Okay, I'm updating it.

rubhanazeem: @dmarcoux how can we differentiate the requests from enterprise and public instances? It's important to differentiate the requests because later on, we use `octokit` and octokit requires api_endpoint for enterprise API calls.
https://github.com/octokit/octokit.rb#interacting-with-the-githubcom-apis-in-github-enterprise

We need a way to differentiate these requests, this is what we do by checking the header `X-GitHub-Enterprise-Host`. We can also do that by having a regex for URLs coming from payload but that would not be a solid solution. What do you think?

rubhanazeem:
> Hey @rubhanazeem seems like you still using the hardcoded version of the gitlab url for the Status Reporter
> 
> https://github.com/openSUSE/open-build-service/blob/93f6dcc696774ec8e344347c9af8fa577278894e/src/api/app/services/scm_status_reporter.rb#L25

@krauselukas thanks for pointing that out. It's updated.

src/api/app/controllers/trigger_workflow_controller.rb
hennevogel: > the requests because later on, we use octokit and octokit requires api_endpoint for enterprise API calls.

Why do we need to?

rubhanazeem: @dmarcoux its updated. I’ve removed the additional parameter.

rubhanazeem: @hennevogel I know atm we only support public repos but it was easy to handle endpoints for private instances in a single go. At this point we just have the ability to separate public and private repos. In future if we decide to add support for private instances then we already have the implementation here.

src/api/app/services/trigger_controller_service/scm_extractor.rb
```diff
@@ -47,23 +53,28 @@ def github_extractor_payload
...
         commit_sha: @payload.dig('object_attributes', 'last_commit', 'id'),
         pr_number: @payload.dig('object_attributes', 'iid'),
         source_branch: @payload.dig('object_attributes', 'source_branch'),
         target_branch: @payload.dig('object_attributes', 'target_branch'),
         action: @payload.dig('object_attributes', 'action'), # TODO: Names may differ, maybe we need to find our own naming (defer to service?)
         project_id: @payload.dig('project', 'id'),
         path_with_namespace: @payload.dig('project', 'path_with_namespace'),
-        event: @event
+        event: @event,
+        scheme: uri.scheme,
```
dmarcoux, line 76: How about we do the same as in `github_extractor_payload` by having a key `api_endpoint`?

rubhanazeem: Done.

src/api/app/services/scm_status_reporter.rb
```diff
@@ -11,14 +11,14 @@ def initialize(event_payload, event_subscription_payload, scm_token, event_type
...
-      github_client = Octokit::Client.new(access_token: @scm_token)
+      github_client = Octokit::Client.new(access_token: @scm_token, api_endpoint: @event_subscription_payload[:api_endpoint])
       # https://docs.github.com/en/rest/reference/repos#create-a-commit-status
       github_client.create_status(@event_subscription_payload[:target_repository_full_name],
                                   @event_subscription_payload[:commit_sha],
                                   @state,
                                   status_options)
     else
-      gitlab_client = Gitlab.client(endpoint: 'https://gitlab.com/api/v4',
+      gitlab_client = Gitlab.client(endpoint: "#{@event_subscription_payload[:scheme]}://#{@event_subscription_payload[:host]}/api/v4",
```
dmarcoux, line 21:
```suggestion
      gitlab_client = Gitlab.client(endpoint: "#{@event_subscription_payload[:api_endpoint]}/api/v4",
```

src/api/app/services/workflows/yaml_downloader.rb
```diff
@@ -25,9 +26,10 @@ def download_yaml_file
     def download_url
       case @scm_payload[:scm]
       when 'github'
-        "https://raw.githubusercontent.com/#{@scm_payload[:target_repository_full_name]}/#{@scm_payload[:target_branch]}/.obs/workflows.yml"
+        client = Octokit::Client.new(access_token: @token.scm_token, api_endpoint: @scm_payload[:api_endpoint])
+        client.content("#{@scm_payload[:target_repository_full_name]}", path: '/.obs/workflows.yml')[:download_url]
       when 'gitlab'
-        "https://gitlab.com/#{@scm_payload[:path_with_namespace]}/-/raw/#{@scm_payload[:target_branch]}/.obs/workflows.yml"
+        "#{@scm_payload[:scheme]}://#{@scm_payload[:host]}/#{@scm_payload[:path_with_namespace]}/-/raw/#{@scm_payload[:target_branch]}/.obs/workflows.yml"
```
dmarcoux, line 32: Same as my previous comment:

```suggestion
        "#{@scm_payload[:api_endpoint]}/#{@scm_payload[:path_with_namespace]}/-/raw/#{@scm_payload[:target_branch]}/.obs/workflows.yml"
```

src/api/spec/services/workflows/yaml_downloader_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
-RSpec.describe Workflows::YAMLDownloader, type: :service do
-  let(:yaml_downloader) { described_class.new(payload) }
+RSpec.describe Workflows::YAMLDownloader, type: :service, vcr: true do
+  let(:yaml_downloader) { described_class.new(payload, token: FactoryBot.build(:workflow_token)) }
```
dmarcoux, line 4: `FactoryBot` shouldn't be needed here.

```suggestion
  let(:yaml_downloader) { described_class.new(payload, token: build(:workflow_token)) }
```

rubhanazeem: Done.

src/api/app/services/workflows/yaml_downloader.rb
rubhanazeem: done

src/api/app/services/scm_status_reporter.rb
rubhanazeem: done

src/api/spec/services/workflows/yaml_downloader_spec.rb
```diff
@@ -11,17 +11,32 @@
...
     context 'gitlab' do
-      let(:payload) { { scm: 'gitlab', target_branch: 'master', path_with_namespace: 'openSUSE/obs-server' } }
+      let(:payload) { { scm: 'gitlab', target_branch: 'master', path_with_namespace: 'openSUSE/obs-server', scheme: 'https', host: 'gitlab.com' } }
       let(:url) { "https://gitlab.com/#{payload[:path_with_namespace]}/-/raw/#{payload[:target_branch]}/.obs/workflows.yml" }
 
       it { expect(Down).to have_received(:download).with(url, max_size: max_size) }
     end
   end
+
+  def github_payload
```
dmarcoux, line 28: Does this really need to be a method? `let(:github_payload)` should also be fine here.

src/api/spec/services/workflows/yaml_downloader_spec.rb
```diff
@@ -11,17 +11,32 @@
...
-      let(:payload) { { scm: 'github', target_branch: 'master', target_repository_full_name: 'openSUSE/obs-server' } }
+      let(:payload) { github_payload }
       let(:url) { "https://raw.githubusercontent.com/#{payload[:target_repository_full_name]}/#{payload[:target_branch]}/.obs/workflows.yml" }
 
       it { expect(Down).to have_received(:download).with(url, max_size: max_size) }
     end
 
     context 'gitlab' do
-      let(:payload) { { scm: 'gitlab', target_branch: 'master', path_with_namespace: 'openSUSE/obs-server' } }
+      let(:payload) { { scm: 'gitlab', target_branch: 'master', path_with_namespace: 'openSUSE/obs-server', scheme: 'https', host: 'gitlab.com' } }
```
dmarcoux, line 21: This should be changed too now that we have `api_endpoint`, no?

src/api/spec/services/workflows/yaml_downloader_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
-RSpec.describe Workflows::YAMLDownloader, type: :service do
-  let(:yaml_downloader) { described_class.new(payload) }
+RSpec.describe Workflows::YAMLDownloader, type: :service, vcr: true do
```
dmarcoux, line 3: VCRs are needed for what exactly? Nothing about this should depend on the backend.

rubhanazeem: It's needed because there is octokit used.

src/api/spec/services/workflows/yaml_downloader_spec.rb
rubhanazeem: done.

src/api/spec/services/workflows/yaml_downloader_spec.rb
rubhanazeem: Sure, it's updated.

src/api/spec/services/workflows/yaml_downloader_spec.rb
dmarcoux: Octokit has nothing to do with the Perl backend, it's interacting with GitHub. I'm pretty sure VCRs can be removed for specs of `Workflows::YAMLDownloader` and the specs would still pass. I'm testing this out locally and somehow I'm getting this error for the GitHub spec:

```     
Octokit::Unauthorized:
       GET https://api.github.com/repos/rubhanazeem/hello_world/contents//.obs/workflows.yml: 401 - Bad credentials
```

rubhanazeem: Exactly! VCR can be used for any external service.

rubhanazeem: Because incase of GitLab there is no API call to GitLab.

dmarcoux: > Exactly! VCR can be used for any external service.

Yeah, I was having a brain fart... hahaha

> Why is it needed for GitHub, but not for GItLab?

I checked again the code and it's all good.

dmarcoux:
@rubhanazeem I fixed the conflicts for you since it came from my changes merged on master.

rubhanazeem:
> @rubhanazeem I fixed the conflicts for you since it came from my changes merged on master.

@dmarcoux I had to rebase it again because some of my charges were dropped.

dmarcoux:
> @dmarcoux I had to rebase it again because some of my charges were dropped.

:facepalm: Sorry for that! I'll leave it to the PR author next time.

There are some failing specs: https://app.circleci.com/pipelines/github/openSUSE/open-build-service/3921/workflows/1da1373e-9ffc-4e7f-bff4-1ef1c5047d8d/jobs/64789

rubhanazeem:
@dmarcoux Yeah, I saw. They are coming from `Avoid leaking Workflows::YAMLDownloader errors outside the class`.

dmarcoux (approved):

PR #11331, opened 2021-07-05
Improve error message in case workflows.yml is not found

vpereira:
In case we want to give a hint that the problem happens when we are downloading the `workflows.yml`

follow-up of #11321

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -12,7 +12,10 @@ class Token::Workflow < Token
     scm_extractor_payload = extractor.call
     yaml_downloader = Workflows::YAMLDownloader.new(scm_extractor_payload)
     yaml_file = yaml_downloader.call
-    raise Token::Errors::NonExistentWorkflowsFile, yaml_downloader.errors.join('\n') if yaml_file.blank?
+    if yaml_file.blank?
+      raise Token::Errors::NonExistentWorkflowsFile,
+            ".obs/workflows.yml could not be downloaded on the SCM branch #{scm_extractor_payload[:target_branch]}: #{yaml_downloader.errors.join('\n')}"
+    end
 
     workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_extractor_payload: scm_extractor_payload, token: self).call
 


dmarcoux:
You mean follow-up to #11321, right?

src/api/app/models/token/workflow.rb
```diff
@@ -12,7 +12,7 @@ def call(options)
     scm_extractor_payload = extractor.call
     yaml_downloader = Workflows::YAMLDownloader.new(scm_extractor_payload)
     yaml_file = yaml_downloader.call
-    raise Token::Errors::NonExistentWorkflowsFile, yaml_downloader.errors.join('\n') if yaml_file.blank?
+    raise Token::Errors::NonExistentWorkflowsFile, "issue with workflows.yml: #{yaml_downloader.errors.join('\n')}" if yaml_file.blank?
```
dmarcoux, line 15: We should let users know that we're looking for their `.obs/workflows.yml` file on their SCM's default branch.

vpereira: ideas?

vpereira:

src/api/app/models/token/workflow.rb
```diff
@@ -12,7 +12,7 @@ def call(options)
     scm_extractor_payload = extractor.call
     yaml_downloader = Workflows::YAMLDownloader.new(scm_extractor_payload)
     yaml_file = yaml_downloader.call
-    raise Token::Errors::NonExistentWorkflowsFile, yaml_downloader.errors.join('\n') if yaml_file.blank?
+    raise Token::Errors::NonExistentWorkflowsFile, "issues with workflows.yml: #{yaml_downloader.errors.join('\n')}" if yaml_file.blank?
```
dmarcoux, line 15: Maybe something like this? The specs should also be updated.

```suggestion
    raise Token::Errors::NonExistentWorkflowsFile, '".obs/workflows.yml" could not be downloaded on the branch "TARGET_BRANCH": #{yaml_downloader.errors.join('\n')}" if yaml_file.blank?
```

dmarcoux (approved):

PR #11337, opened 2021-07-05
Improve setup appliance

M0ses:
* change system-user-* to new mechanism
* Redirect error message in setup-appliance.sh db:migrate as it`s only used for testing and may fail on first startup
* Do not run setup-appliance.sh parallel
*

### DIFF
--- a/dist/setup-appliance.sh
+++ b/dist/setup-appliance.sh
@@ -691,6 +691,7 @@ ENABLE_OPTIONAL_SERVICES=0
 ENABLE_FORCEPROJECTKEYS=1
 MYSQL_USER=root
 MYSQL_PASS=opensuse
+PID_FILE=/run/setup-appliance.pid
 
 # package or appliance defaults
 if [ -e /etc/sysconfig/obs-server ]; then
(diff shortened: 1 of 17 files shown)


dist/setup-appliance.sh
```diff
@@ -709,9 +710,23 @@ fi
 
 
 if [[ ! $BOOTSTRAP_TEST_MODE == 1 && $0 != "-bash" ]];then
-
+  logline "Starting "`basename $0`" at "`date`
   NON_INTERACTIVE=0
 
+  if [ -f $PID_FILE ];then
+    APID=`cat $PID_FILE`
+    if [ ! -f /proc/$APID/status ];then
```
hennevogel, line 718: If you switch the conditions around it's way easier to understand IMHO

M0ses: I didn't get the point - please explain!

hennevogel:
```bash
if true; then
...
else
...
fi
```

vs.

```bash
if !true; then
...
else
...
fi
```

M0ses: Feel free to send an PR - and test it - if you really unhappy with the current solution.
Testing these changes is a lot of effort and I'm not willing to do this effort only for "coding style"

eduardoj: I'm with Henne here. I will create a PR later, with his suggestion, so we can also merge this PR.

eduardoj (approved):

PR #11343, opened 2021-07-06
Handle the Psych::SyntaxError exception in case workflows.yml isnt a …

### LINKED ISSUE #11279: [production][trigger_workflow#create] (<unknown>): mapping values are not allowed in this context at line 4 column 38
trigger_workflow#create threw this 1 times
<pre>
(&lt;unknown&gt;): mapping values are not allowed in this context at line 4 column 38
</pre>
[https://build.opensuse.org/trigger/workflow?id=5032](https://build.opensuse.org/trigger/workflow?id=5032)"

[See this exception on Errbit](http://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/60cb66d384bf4607a859d062)

vpereira:
Fixes #11279

### DIFF
--- a/src/api/app/models/token/errors.rb
+++ b/src/api/app/models/token/errors.rb
@@ -12,4 +12,8 @@ module Token::Errors
   class SCMTokenInvalid < APIError
     setup 401
   end
+
+  class WorkflowsYamlNotParsable < APIError
+    setup 400
+  end
 end
--- a/src/api/app/services/workflows/yaml_to_workflows_service.rb
+++ b/src/api/app/services/workflows/yaml_to_workflows_service.rb
@@ -13,7 +13,12 @@ module Workflows
     private
 
     def create_workflows
-      parsed_workflows_yaml = YAML.safe_load(File.read(@yaml_file))
+      begin
+        parsed_workflows_yaml = YAML.safe_load(File.read(@yaml_file))
+      rescue Psych::SyntaxError => e
+        raise Token::Errors::WorkflowsYamlNotParsable, "Unable to parse .obs/workflows.yml: #{e.message}"
+      end
+
       workflows = []
 
       parsed_workflows_yaml.each do |_workflow_name, workflow|
--- a/src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
+++ b/src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
@@ -3,6 +3,7 @@ require 'rails_helper'
 RSpec.describe Workflows::YAMLToWorkflowsService, type: :service do
   include_context 'a scm payload hash'
   let(:workflows_yml_file) { File.expand_path(Rails.root.join('spec/support/files/workflows.yml')) }
+  let(:unparsable_workflows_yml_file) { File.expand_path(Rails.root.join('spec/support/files/unparsable_workflows.yml')) }
   let(:token) { create(:workflow_token) }
 
   describe '#call' do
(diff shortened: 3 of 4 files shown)


src/api/app/services/workflows/yaml_to_workflows_service.rb
```diff
@@ -13,7 +13,12 @@ def call
     private
 
     def create_workflows
-      parsed_workflows_yaml = YAML.safe_load(File.read(@yaml_file))
+      begin
+        parsed_workflows_yaml = YAML.safe_load(File.read(@yaml_file))
+      rescue Psych::SyntaxError
```
dmarcoux, line 18: Could it be useful to put the message from `Psych::SyntaxError`? I'm wondering because I don't know what exactly the message could be.

So the code would then be:
```ruby
      rescue Psych::SyntaxError => e
        raise WorkflowsYamlNotParsable, "Unable to parse .obs/workflows.yml: #{e.message}"
      end
```

vpereira: the message is what we see in the errbit error.. maybe useful for the user, not sure. Without having the yaml file is definitely not useful.

src/api/app/models/token/errors.rb
```diff
@@ -12,4 +12,8 @@ class NonExistentWorkflowsFile < APIError
   class SCMTokenInvalid < APIError
     setup 401
   end
+
+  class WorkflowsYamlNotParsable < APIError
```
dmarcoux, line 16: This error isn't really related to the `Token` models. Maybe `src/api/app/models/workflow/errors.rb` would be a better location. That file doesn't exist yet, so it needs to be created.

src/api/app/services/workflows/yaml_to_workflows_service.rb
```diff
@@ -13,7 +13,12 @@ def call
     private
 
     def create_workflows
-      parsed_workflows_yaml = YAML.safe_load(File.read(@yaml_file))
+      begin
+        parsed_workflows_yaml = YAML.safe_load(File.read(@yaml_file))
+      rescue Psych::SyntaxError => e
+        raise WorkflowsYamlNotParsable, "Unable to parse .obs/workflows.yml: #{e.message}"
```
dmarcoux, line 19: This line is wrong:

```
NameError: uninitialized constant Workflows::YAMLToWorkflowsService::WorkflowsYamlNotParsable
```

We should have a spec to cover that this line correctly raises.

dmarcoux (changes requested):

src/api/app/services/workflows/yaml_to_workflows_service.rb
dmarcoux: They know that it's the `.obs/workflows.yml` file from their SCM repository, so it can help them.

src/api/app/models/token/errors.rb
vpereira: yes, later maybe

src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
```diff
@@ -19,5 +20,12 @@
         expect(service.call.first).to be_a(Workflow)
       end
     end
+
+    context 'with a invalid workflows.yml' do
+      it 'initializes a workflow object' do
```
dmarcoux, line 25:
```suggestion
      it 'raises a user-friendly error' do
```

dmarcoux (approved):

vpereira:

PR #11347, opened 2021-07-07
Add real_time_tidy to ThinkingSphinx dev configuration

saraycp:
This will clean up real-time orphan records after reindexing. As the first step, to ensure this is not causing any error, we are going to enable it only in development environment.


vpereira (approved):

eduardoj (approved):

saraycp:
Both `ci/circleci: backend_test` and `ci/circleci: checkout_code` passed in CircleCI but the result didn't arrive to the PR.
So I'm going to merge.

PR #11348, opened 2021-07-07
Add data migration to handle old event_subscriptions

### LINKED ISSUE #11346: [production][] no implicit conversion of nil into String
threw this 56 times
<pre>
no implicit conversion of nil into String
</pre>

[See this exception on Errbit](http://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/60e560fa84bf4607e2ab93cd)

rubhanazeem:
We need to migrate SCM event_subscriptions according to the new payload structure.

Co-authored-by: @saraycp 

Fixes #11346

### DIFF
--- a/src/api/db/data/20210707102506_add_api_endpoints_to_payload.rb
+++ b/src/api/db/data/20210707102506_add_api_endpoints_to_payload.rb
@@ -1,0 +1,21 @@
+class AddApiEndpointsToPayload < ActiveRecord::Migration[6.1]
+  def up
+    EventSubscription.where(channel: 'scm').each do |event_sub|
+      next if event_sub.payload['api_endpoint'].present?
+
+      if event_sub.payload['scm'] == 'github'
+        event_sub.payload['api_endpoint'] = 'https://api.github.com'
+      else
+        http_url = event_sub.payload['http_url']
+        uri = URI.parse(http_url)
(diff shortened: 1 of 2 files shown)


src/api/db/data/20210707102506_add_api_endpoints_to_payload.rb
```diff
@@ -0,0 +1,21 @@
+class AddApiEndpointsToPayload < ActiveRecord::Migration[6.1]
+  def up
+    EventSubscription.where(channel: 'scm').each do |event_sub|
+      next if event_sub.payload['api_endpoint'].present?
+
+      if event_sub.payload['scm'] == 'github'
+        event_sub.payload['api_endpoint'] = 'https://api.github.com'
```
hennevogel, line 7: Will this be correct if this migration is run *after* #11325?

hennevogel (changes requested):

src/api/db/data/20210707102506_add_api_endpoints_to_payload.rb
rubhanazeem: Yes, because we only update this if `event_sub.payload['api_endpoint']` is not available. After #11325 it always is available.

hennevogel (approved):

Issue #11127, opened 2021-05-12
Branching a package copies the description and the title but not the url


danidoni:
# Issue Description
Branching a package copies the title and the description, but not the url.

As can be seen in [this branch of ding](https://build.opensuse.org/package/show/home:danidoni:branches:X11:Utilities/ding), it lacks the url the [linked package](https://build.opensuse.org/package/show/X11:Utilities/ding) has.

Expected Result
============
The branching should copy the title, the description and the url.

How to Reproduce
===============
1. Go to [ding](https://build.opensuse.org/package/show/X11:Utilities/ding) package.
2. Branch it off clicking on the left action menu.
3. See no url in the branched package.

vishaldeepak:
Id like to give this a shot

dmarcoux:
@vishaldeepak Sure, let us know if something isn't clear.

saraycp:
This was fixed in PR [#11352](https://github.com/openSUSE/open-build-service/pull/11352)

PR #11352, opened 2021-07-07
When branching a package also copy its url

vishaldeepak:
Hi,
In branch_package.rb when creating a new package also copy the url. This addresses the issue #11127 .

To see the changes 
- Goto any package example, http://localhost:3000/package/show/home:Admin/FirstPackage
- Click on Branch Package
- Hit branch
- The newly created branch should also have the URL

Added Rspec Test Case as necessary.

Before
Screenshot 2021-07-08 at 12 38 43 AM

After:
Screenshot 2021-07-08 at 12 41 46 AM

### DIFF
--- a/src/api/app/models/branch_package.rb
+++ b/src/api/app/models/branch_package.rb
@@ -183,7 +183,7 @@ class BranchPackage
         raise DoubleBranchPackageError.new(tprj.name, tpkg.name), "branch target package already exists: #{tprj.name}/#{tpkg.name}" unless params[:force]
       else
         if pac.is_a?(Package)
-          tpkg = tprj.packages.new(name: pack_name, title: pac.title, description: pac.description)
+          tpkg = tprj.packages.new(name: pack_name, title: pac.title, description: pac.description, url: pac.url)
           tpkg.bcntsynctag = pac.bcntsynctag
         else
           tpkg = tprj.packages.new(name: pack_name)
@@ -273,12 +273,13 @@ class BranchPackage
 
       title = "Branch project for package #{params[:package]}"
       description = "This project was created for package #{params[:package]} via attribute #{@attribute}"
+      base_project_url = Project.find_by_name(params[:project]).try(:url)
       if params[:request]
         title = "Branch project based on request #{params[:request]}"
         description = "This project was created as a clone of request #{params[:request]}"
       end
       @add_repositories = true # new projects shall get repositories
-      tprj = Project.new(name: @target_project, title: title, description: description)
+      tprj = Project.new(name: @target_project, title: title, description: description, url: base_project_url)
       tprj.relationships.new(user: User.session!, role: Role.find_by_title!('maintainer'))
       tprj.flags.new(flag: 'build', status: 'disable') if @extend_names
       tprj.flags.new(flag: 'access', status: 'disable') if @noaccess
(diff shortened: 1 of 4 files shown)


src/api/spec/models/branch_package_spec.rb
```diff
@@ -97,5 +97,14 @@
         end
       end
     end
+
+    context 'Newly created Package' do
+      it 'retains url' do
+        leap_project
+        branch_apache_package.branch
+        project = Package.where(name: 'apache2').where.not(id: apache.id).first
```
coolo, line 105: that `project` variable is actually a package. But that opens a side discussion: why not copy the url in the branch project also? It's just one bug report away

src/api/app/models/branch_package.rb
```diff
@@ -273,12 +273,13 @@ def create_branch_project
 
       title = "Branch project for package #{params[:package]}"
       description = "This project was created for package #{params[:package]} via attribute #{@attribute}"
+      base_project_url = Project.find_by_name(params[:project]).try(:url)
```
vishaldeepak, line 276: is there a chance that params[:project] will not be received, also if there is a `params[:request]` should there be any changes?

src/api/spec/models/branch_package_spec.rb
vishaldeepak: made changes as reqiured @coolo. One thing worth looking at - https://github.com/openSUSE/open-build-service/pull/11352/files#r666173752

src/api/app/models/branch_package.rb
danidoni: Looking into the code base, right now `params[:project]` is always being passed but there is nothing in the code that prevents anyone from not passing it.

I don't know what to say about your second question... Maybe someone else can shed a bit of light?

vishaldeepak: I guess then its a misunderstanding from me, I think than its ready for a merge, unless there is any other comments

dmarcoux: The presence of `params[:request]` shouldn't affect the URL, so the changes are fine.

dmarcoux (changes requested):
Thank you @vishaldeepak for submitting this PR. Could you rebase the PR on `master` to remove the unneeded merge commit `Merge pull request #11331 from vpereira/fix_11240`?

Everything else looks good.

vishaldeepak:
Rebased with master. Let me know if there is some other priority issue that I could take up.

src/api/spec/models/branch_package_spec.rb
```diff
@@ -97,5 +97,19 @@
         end
       end
     end
+
+    context 'retains url' do
+      it 'for a newly created package' do
+        branch_apache_package.branch
+        package = Package.where(name: 'apache2').where.not(id: apache.id).first
+        expect(package.url).to eq(apache.url)
```
dmarcoux, line 105: Somehow this spec is now failing? Were some of the changes lost in the rebase?

vishaldeepak: apologies, but it should be resolved now.

dmarcoux: No need to apologize, mistakes happen :slightly_smiling_face:

dmarcoux (approved):
Nice work @vishaldeepak!

PR #11354, opened 2021-07-08
Refactor branch package step

danidoni:
This PR moves some methods into `BranchPackageStep`


dmarcoux:
This is doing a lot of things which are already done in #11318. I would suggest closing this PR since it's going to clash with #11318 once it's merged.

PR #11358, opened 2021-07-08
Setup filters in .obs/workflows.yml

dmarcoux:
To use the feature implemented in #11318.

The architectures and repositories are the same as the ones from the CI check `OBS Package Build`.

Once we merge this, following PRs will only report what we defined in the filters.

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -3,3 +3,12 @@ workflow:
     - branch_package:
         source_project: OBS:Server:Unstable
         source_package: obs-server
+  filters:
+    architectures:
+      only:
+        - x86_64
+    repositories:
+      only:
+        - SLE_12_SP5
+        - SLE_15_SP1
+        - SLE_15_SP2


dmarcoux:
[Limit reporting of build results via filters for architectures and repositories](https://trello.com/c/HXYMXh74/1696-limit-reporting-of-build-results-via-filters-for-architectures-and-repositories)

.obs/workflows.yml
```diff
@@ -3,3 +3,12 @@ workflow:
...
         source_package: obs-server
+  filters:
+    architectures:
+      only:
+        - x86_64
+    repositories:
+      only:
+        - SLE_12_SP5
+        - SLE_15_SP1
+        - SLE_15_SP2
```
krauselukas, line 14: @dmarcoux I know that these are the repositories we used to report the status for, but the latest stable release for SLE 15 is SP3, maybe we should include it to the list, what do you think?

dmarcoux: For now, I would keep it to what we had. SLE15 SP3 is currently unresolvable.

krauselukas: okay, not resolvable at the moment

krauselukas (approved):

eduardoj (approved):

PR #11361, opened 2021-07-08
Fix field set format

hennevogel:
Field sets are, like tag sets, separated with a comma and no spaces.


eduardoj (approved):

hennevogel:
This seems to have stop our telegraf measurements.

PR #11365, opened 2021-07-09
Adopt OBS repositories we CI

hennevogel:
Factory for the future, SLES for production, Leap for the appliance.

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -9,6 +9,7 @@ workflow:
         - x86_64
     repositories:
       only:
-        - SLE_12_SP5
+        - openSUSE_Factory
+        - openSUSE_15.1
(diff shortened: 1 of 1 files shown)


eduardoj (approved):

coolo:
Shouldn't openSUSE_Factory build for that to be a requirement? That's btw the problem with your approach to take the workfow from master - it's easy to break it and have passing CI

hennevogel:
> Shouldn't openSUSE_Factory build for that to be a requirement? 

What?

hennevogel:
>  That's btw the problem with your approach to take the workfow from master - it's easy to break it and have passing CI

Yeah currently we do this to be on the safe side. Taking this from forks would require more thought.

PR #11378, opened 2021-07-12
Send scheduler metrics first

eduardoj:
"jobs building" metrics could be a thousand metrics, while the rest sent by the WorkerMeasurementsJob could hardly reach a hundred metrics.

By switching the order the metrics are sent leaving the "jobs building" metrics at the end, the scheduler metrics are guaranteed to be parsed.

Monkey Patched in production, until scheduler metrics can be seen in Grafana.

-----------------

### DIFF
--- a/src/api/app/jobs/worker_measurements_job.rb
+++ b/src/api/app/jobs/worker_measurements_job.rb
@@ -10,8 +10,8 @@ class WorkerMeasurementsJob < ApplicationJob
     @architecture_names = Architecture.available.pluck(:name)
 
     send_worker_metrics
-    send_job_metrics
     send_scheduler_metrics
+    send_job_metrics
   end
 
   private


dmarcoux:
This has been monkey-patched in production for 2 days, what is the reason for the `DO NOT MERGE` label?

vpereira:
> This has been monkey-patched in production for 2 days, what is the reason for the `DO NOT MERGE` label?

I don't see why we have to test it in production if we have a AHM development environment: https://github.com/openSUSE/open-build-service/wiki/Site-Reliability#development-environment

eduardoj:
> This has been monkey-patched in production for 2 days, what is the reason for the `DO NOT MERGE` label?

This patch was a way of testing that the problem was not related to sending scheduler metrics.

Even now, that scheduler metrics are correctly shown in Grafana, this patch doesn't provide a full fix for the problem. "jobs building" metrics are not all shown. There seems to be a problem sending so many metrics at once.

eduardoj:
> > This has been monkey-patched in production for 2 days, what is the reason for the `DO NOT MERGE` label?
> 
> I don't see why we have to test it in production if we have a AHM development environment: https://github.com/openSUSE/open-build-service/wiki/Site-Reliability#development-environment

Using the AHM development environment would have required setting up fake "job building" results - not a straightforward task. And even after that, there was no guarantee that the error could have been reproduced.

vpereira:
> Using the AHM development environment would have required setting up fake "job building" results - not a straightforward task. And even after that, there was no guarantee that the problem could have been reproduced.

the AHM development environment is just a development environment + Grafana + Influxdb. If you build a package it will be built as usual.. Or you enter in a rails console (while running the development environment) and schedule some jobs manually, or even you setup an appliance with Grafana and Influxdb and run your tests there.. IMO production is definitely not the place to do these tests

eduardoj:
After #11392 got merged and deployed, this monkey patch is no longer needed.

PR #11385, opened 2021-07-15
Stop sending "building jobs" measurements

eduardoj:
Sending "building jobs" rise up two problems:
1) sending them before scheduler measurements, Grafana didn't get data about scheduler measurements,
2) sending them after scheduler measurements, Grafana didn't get all the data regarding "building jobs" measurements

In both cases, graphics shown in Grafana about "building jobs" and schedulers are currently not accurate.

Therefore, stop sending "building jobs" measurements till the root cause of the problem is figured out. Tests should be done in a test environment, not in production.

Overrides #11378.

-----------------

### DIFF
--- a/src/api/app/jobs/worker_measurements_job.rb
+++ b/src/api/app/jobs/worker_measurements_job.rb
@@ -37,13 +37,6 @@ class WorkerMeasurementsJob < ApplicationJob
       RabbitmqBus.send_to_bus('metrics', "jobs,arch=#{architecture_name},state=waiting count=#{waiting}")
       RabbitmqBus.send_to_bus('metrics', "jobs,arch=#{architecture_name},state=blocked count=#{blocked}")
     end
-
-    @workerstatus.search('//building').each do |job|
-      hostarch = job.attributes['hostarch'].value
-      arch = job.attributes['arch'].value
-      progression = Time.now.to_i - job.attributes['starttime'].value.to_i
-      RabbitmqBus.send_to_bus('metrics', "jobs,hostarch=#{hostarch},arch=#{arch},state=building progression=#{progression},count=1")
-    end
   end
 
   def send_scheduler_metrics


hennevogel:
I have two questions: 

- What does it cost to keep sending those?
- What does it cost to stop sending them?

vpereira:
Hey @eduardoj, and if you move just the `send_job_metrics` for its own job, running totally independent from the others? Actually IMO every single method in this job should be a job itself... what do you think?

eduardoj:
> I have two questions:
> 
>     * What does it cost to keep sending those?
One of these three options:
- The cost of reverting the monkey patch, in case we want to follow master.
- The cost of applying the monkey patch in every deployment, in case we want to keep it.
- The cost of approving, merging and deploying the PR with the monkey patch, in case we want to apply the PR with the monkey patch.

>     * What does it cost to stop sending them?
The cost of approving, merging, and deploying this PR.

hennevogel:
> The cost of approving, merging, and deploying this PR.

And future costs?

eduardoj:
> Hey @eduardoj, and if you move just the `send_job_metrics` for its own job, running totally independent from the others? Actually IMO every single method in this job should be a job itself... what do you think?

All metrics are sent without issues to RabbitMQ from the current job. No problems or issues there.

The problem is that Grafana does not recognize the same amount of metrics that are received from RabbitMQ. Either the sent data is not correct, or there is an issue related to Telegraf, InfluxDB and/or Grafana with that metrics. Splitting the sending of the metrics will not help, IMHO. But we could also test this option in a test environment.

eduardoj:
> > The cost of approving, merging, and deploying this PR.
> 
> And future costs?

We will have wrong graphics in Grafana, for the time this PR is not merged. Either we should be aware of this period of wrong data, or we should delete this data.

hennevogel:
> We will have wrong graphics in Grafana, for the time this PR is not merged. Either we should be aware of this period of wrong data, or we should delete this data.

And what will it cost replicating the bug outside of production?

eduardoj:
> And what will it cost replicating the bug outside of production?

I hope that it will be less than having wrong data/graphics in Grafana until the time we finally find a fix. Only guessing.

hennevogel:
> wrong data/graphics in Grafana

And having this data wrong in Grafana costs what exactly? 😃

eduardoj:
> And having this data wrong in Grafana costs what exactly? :smiley:

Not having wrong data in Grafana will prevent us from:
- giving explanations about wrong data in the Grafana panels (writing side warning notes...),
- removing wrong data after the issue is fixed, that would be more data than if this PR is approved, merged and deployed today.

If these costs are considered relevant or not, or we could live with it, that's another issue.

eduardoj:
#11392 offers a better solution.

PR #11387, opened 2021-07-15
[dist] remove Wants= from most of our systemd service files

M0ses:
"Wants=" will start the configured service, if disabled or not. This is
not we want with our obscheduler or OBS API services.

In a distributed setup like build.opensuse.org it is not common that the
database runs on the same host like the API, therefor starting mariadb
or running obsapisetup is not recommended.


krauselukas (approved):
@M0ses Thanks!

PR #11392, opened 2021-07-16
Correct queue measurements

hennevogel:
- Drop measurement of individual jobs, progression doesn't provide much value and we already track performance with build events
- Use single field measurement, they are easier to plot
- Do not send 0 value measurements
- Use value as field name, it's default in grafana


eduardoj (approved):

hennevogel:
Hm, now coolos face again?

PR #11394, opened 2021-07-17
Search controller refactoring

vpereira:
Extract code from `SearchController#find_attribute` to methods

There are more refactoring to come, for now, just some cleaning tasks.

-----------------

### DIFF
--- a/src/api/spec/controllers/search_controller_spec.rb
+++ b/src/api/spec/controllers/search_controller_spec.rb
@@ -1,0 +1,50 @@
+require 'rails_helper'
+require 'webmock/rspec'
+
+RSpec.describe SearchController, vcr: true do
+  render_views
+
+  let!(:user) { create(:confirmed_user, :with_home, login: 'foo') }
+
+  before do
+    login user
+  end
+
+  shared_examples 'find project' do
+    it { expect(response).to have_http_status(:success) }
+    it { expect(Xmlhash.parse(response.body)['project']['name']).to eq(user.home_project.name) }
+  end
+
+  shared_examples 'find package' do
+    it { expect(response).to have_http_status(:success) }
+    it { expect(Xmlhash.parse(response.body)['package']['project']).to eq(user.home_project.name) }
+  end
+
+  describe 'search for projects' do
+    subject! { get :project, params: { match: "@name='#{user.home_project.name}'" } }
+
+    it_behaves_like 'find project'
+  end
+
+  describe 'search for project ids' do
+    subject! { get :project_id, params: { match: "@name='#{user.home_project.name}'" } }
+
+    it_behaves_like 'find project'
+  end
+
+  describe 'search for packages' do
+    let!(:package) { create(:package, name: 'apacheX', project: user.home_project) }
+
+    subject! { get :package, params: { match: "@name='apacheX'" } }
+
+    it_behaves_like 'find package'
+  end
+
+  describe 'search for package ids' do
+    let!(:package) { create(:package, name: 'apacheX', project: user.home_project) }
+
+    subject! { get :package_id, params: { match: "@name='apacheX'" } }
+
+    it_behaves_like 'find package'
+  end
+end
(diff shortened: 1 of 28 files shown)


hennevogel:
Will you move over the rest of the tests from `test/functional/search_controller_test.rb` too?

src/api/spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,107 @@
+require 'rails_helper'
+require 'webmock/rspec'
+
+RSpec.describe SearchController, vcr: true do
+  render_views
+
+  shared_examples 'block anonymous request' do
+    it { expect(response).to have_http_status(:redirect) }
```
hennevogel, line 8: I don't think this makes much sense. `require_login` is a "default" `before_filter` from `ApplicationController`.

vpereira:
> Will you move over the rest of the tests from `test/functional/search_controller_test.rb` too?

Yes, we should at least cover all the scenarios covered by minitest, yes :+1:

src/api/spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  shared_examples 'find project' do
+    it { expect(response).to have_http_status(:success) }
+    it { expect(Xmlhash.parse(response.body)).to have_key('project') }
+    it { expect(Xmlhash.parse(response.body)['project']).to have_key('name') }
+    it { expect(Xmlhash.parse(response.body)['project']['name']).to eq(user.home_project.name) }
```
hennevogel, line 15: Aren't those 3 the same test? If you don't have a key project you will for sure not have `['project']['name']` with some expected content.

src/api/spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  shared_examples 'find package' do
+    it { expect(response).to have_http_status(:success) }
+    it { expect(Xmlhash.parse(response.body)).to have_key('package') }
+    it { expect(Xmlhash.parse(response.body)['package']).to have_key('name') }
+    it { expect(Xmlhash.parse(response.body)['package']['name']).to eq('apacheX') }
```
hennevogel, line 22: Same here...

hennevogel (changes requested):

src/api/spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+      it_behaves_like 'block anonymous request'
+    end
+
+    context 'authenticated' do
+      let!(:user) { create(:confirmed_user, :with_home, login: 'foo') }
+
+      before do
+        login user
+      end
```
hennevogel, line 38: You can move this into the shared example.

vpereira: done

src/api/spec/controllers/search_controller_spec.rb
vpereira: done

src/api/spec/controllers/search_controller_spec.rb
vpereira: your suggestion is to keep just the last one?

hennevogel: jupp

src/api/spec/controllers/search_controller_spec.rb
vpereira: done

src/api/spec/controllers/search_controller_spec.rb
vpereira: done

src/api/spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'rails_helper'
+require 'webmock/rspec'
+
+RSpec.describe SearchController, vcr: true do
+  render_views
+
+  let!(:user) { create(:confirmed_user, :with_home, login: 'foo') }
```
hennevogel, line 7: nitpick: `let` should be enough...

hennevogel (approved):

PR #11397, opened 2021-07-19
Search controller refactoring

vpereira:
Refactor method `SearchController#filter_items` to private and change how to handle `limit` and `offset`. Add spec for this case

-----------------

### DIFF
--- a/src/api/spec/controllers/search_controller_spec.rb
+++ b/src/api/spec/controllers/search_controller_spec.rb
@@ -4,7 +4,7 @@ require 'webmock/rspec'
 RSpec.describe SearchController, vcr: true do
   render_views
 
-  let!(:user) { create(:confirmed_user, :with_home, login: 'foo') }
+  let(:user) { create(:confirmed_user, :with_home, login: 'foo') }
   let(:project) { user.home_project }
 
   before do
@@ -39,19 +39,13 @@ RSpec.describe SearchController, vcr: true do
     subject! { get :package, params: { match: "@name='apacheX'" } }
 
     it_behaves_like 'find package'
-
-    context 'With offset' do
-      subject! { get :package, params: { offset: 'XXX', match: "@name='apacheX'" } }
-
-      it_behaves_like 'find package'
-    end
   end
 
   describe 'search for packages with offset = 1' do
     let!(:package) { create(:package, name: 'apacheX', project: user.home_project) }
     let!(:project) { create(:project_with_package, name: 'Foo', maintainer: user, package_name: 'apacheX') }
 
-    subject! { get :package, params: { offset: '1', match: "@name='apacheX'" } }
+    subject! { get :package, params: { limit: '1', offset: '1', match: "@name='apacheX'" } }
 
     it_behaves_like 'find package'
   end
--- a/src/api/spec/controllers/search_controller_spec.rb
+++ b/src/api/spec/controllers/search_controller_spec.rb
@@ -5,6 +5,7 @@ RSpec.describe SearchController, vcr: true do
   render_views
 
   let!(:user) { create(:confirmed_user, :with_home, login: 'foo') }
+  let(:project) { user.home_project }
 
   before do
     login user
@@ -12,12 +13,12 @@ RSpec.describe SearchController, vcr: true do
 
   shared_examples 'find project' do
     it { expect(response).to have_http_status(:success) }
-    it { expect(Xmlhash.parse(response.body)['project']['name']).to eq(user.home_project.name) }
+    it { expect(Xmlhash.parse(response.body)['project']['name']).to eq(project.name) }
   end
 
   shared_examples 'find package' do
     it { expect(response).to have_http_status(:success) }
-    it { expect(Xmlhash.parse(response.body)['package']['project']).to eq(user.home_project.name) }
+    it { expect(Xmlhash.parse(response.body)['package']['project']).to eq(project.name) }
   end
 
   describe 'search for projects' do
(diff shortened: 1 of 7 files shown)


vpereira:
## no offset, no limit

```
$ dosc api '/search/package?match=%40name+%3D+%27Foo%27+or+title+%3D+%27Foo%27+or+description+%3D+%27Foo%27'
WARNING: SSL certificate checks disabled. Connection is insecure!

<collection matches="3">
<package name="Foo" project="home:Admin">
  <title/>
  <description/>
</package>
<package name="Foo" project="Bar">
  <title/>
  <description/>
</package>
<package name="Foo" project="Test">
  <title/>
  <description/>
</package>
</collection>
```
## offset 1 = skip first

```
$ dosc api '/search/package?match=%40name+%3D+%27Foo%27+or+title+%3D+%27Foo%27+or+description+%3D+%27Foo%27&offset=1'
WARNING: SSL certificate checks disabled. Connection is insecure!

<collection matches="3">
<package name="Foo" project="Bar">
  <title/>
  <description/>
</package>
<package name="Foo" project="Test">
  <title/>
  <description/>
</package>
</collection>
```
## offset = 1, limit = 1 (just show the second record)

```
dosc api '/search/package?match=%40name+%3D+%27Foo%27+or+title+%3D+%27Foo%27+or+description+%3D+%27Foo%27&offset=1&limit=1'
WARNING: SSL certificate checks disabled. Connection is insecure!

<collection matches="3">
<package name="Foo" project="Bar">
  <title/>
  <description/>
</package>
</collection>
```

Note that the `collection matches` are always wrong if there is some filtering. That's because we compute it before applying the filter https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/search_controller.rb#L200 is that a bug or feature @adrianschroeter ?

src/api/spec/controllers/search_controller_spec.rb
```diff
@@ -5,19 +5,20 @@
   render_views
 
   let!(:user) { create(:confirmed_user, :with_home, login: 'foo') }
```
hennevogel, line 7: Now you really don't need to use `let!` anymore..

src/api/spec/controllers/search_controller_spec.rb
```diff
@@ -38,6 +39,21 @@
     subject! { get :package, params: { match: "@name='apacheX'" } }
 
     it_behaves_like 'find package'
+
+    context 'With offset' do
+      subject! { get :package, params: { offset: 'XXX', match: "@name='apacheX'" } }
+
+      it_behaves_like 'find package'
```
hennevogel, line 46: I'm not sure what you are testing here? That `offset: 'XXX'` has no effect?

src/api/spec/controllers/search_controller_spec.rb
```diff
@@ -38,6 +39,21 @@
...
+    end
+  end
+
+  describe 'search for packages with offset = 1' do
+    let!(:package) { create(:package, name: 'apacheX', project: user.home_project) }
+    let!(:project) { create(:project_with_package, name: 'Foo', maintainer: user, package_name: 'apacheX') }
+
+    subject! { get :package, params: { limit: '1', offset: '1', match: "@name='apacheX'" } }
+
+    it_behaves_like 'find package'
```
hennevogel, line 56: What about the other cases? Like no `limit` or no `offset`?

hennevogel (changes requested):

src/api/spec/controllers/search_controller_spec.rb
vpereira: yes, well i added this spec before changing the code.. and before the original code was doing some manual convertion: https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/search_controller.rb#L125, i can drop it now for sure.

src/api/spec/controllers/search_controller_spec.rb
vpereira: done

src/api/spec/controllers/search_controller_spec.rb
vpereira: it's being tested here https://github.com/openSUSE/open-build-service/blob/4dd82b05aad4342e1538d438ea56d3cd74b03328/src/api/spec/controllers/search_controller_spec.rb#L41

src/api/spec/controllers/search_controller_spec.rb
hennevogel: Yeah a simple `.to_i` should suffice...

src/api/spec/controllers/search_controller_spec.rb
hennevogel: Hm maybe I don't understand those parameters correctly. `limit=2` would mean out of 5 results I get 2. And `offset=4` would mean I get all results from result 5 onward right? Those cases I mean...

vpereira: you can see https://github.com/openSUSE/open-build-service/pull/11397#issuecomment-883320571 where i explain it (or at least try to) with examples... `offset=1` means we do like `results[1..]`. With `limit` you limit the number of results. So `offset=1` and `limit=4`, would mean something like `results[1..].take(4)` 

i.e:

```
2.6.5 :008 > offset=1
 => 1 
2.6.5 :009 > limit=4
 => 4 
2.6.5 :010 > 0.upto(10).to_a[offset..].take(limit)
 => [1, 2, 3, 4] 
2.6.5 :011 > 
```

src/api/spec/controllers/search_controller_spec.rb
vpereira: done

src/api/spec/controllers/search_controller_spec.rb
hennevogel: Okay, so where is this tested?

hennevogel: We talked offline, this is in minitest and will be moved later...

hennevogel (approved):

PR #11399, opened 2021-07-19
Introduce tokens in web UI

eduardoj:
Manage user tokens from the web user interface.

This new option can  be accessed from "User Profile", "Manage Your Tokens".

This option will be shown only if the ~`user_profile_redesign`~ `trigger_workflow` feature flag is enabled.

This pull request only implements listing and deleting tokens.

Policies for using tokens from the web-ui were created under `app/policies/webui/tokens/`, to not colide with the existing token policies, under `app/policies/token/`.

-----------------
Switching to Draft. Things to be done before switching to "Ready for review":
* [x] Add pagination.
* [x] Add "Create token".
* [x] Don't show the token string field in the list of tokens.
-----------------

### DIFF
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -5,7 +5,7 @@ class Webui::Users::TokensController < Webui::WebuiController
   after_action :verify_policy_scoped, only: :index
 
   def index
-    @tokens = policy_scope([:webui, Token])
+    @tokens = policy_scope([:webui, Token]).page(params[:page])
   end
 
   def destroy
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -1,0 +1,26 @@
+class Webui::Users::TokensController < Webui::WebuiController
+  before_action :set_token, only: [:destroy]
+
+  after_action :verify_authorized, except: :index
+  after_action :verify_policy_scoped, only: :index
+
+  def index
+    @tokens = policy_scope([:webui, Token])
+  end
+
+  def destroy
+    authorize [:webui, @token]
+    @token.destroy
+    flash[:success] = 'Token was successfully deleted.'
+    redirect_to tokens_url
+  end
+
+  private
+
+  def set_token
+    @token = Token.find(params[:id])
+  rescue ActiveRecord::RecordNotFound => e
+    flash[:error] = e.message
+    redirect_to tokens_url
+  end
+end
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -15,7 +15,7 @@ class Webui::Users::TokensController < Webui::WebuiController
   end
 
   def create
-    @token = Token.token_type(@params[:operation]).new(@params.merge(user: User.session, package: @package))
+    @token = Token.token_type(@params[:type]).new(@params.except(:type).merge(user: User.session, package: @package))
 
     authorize [:webui, @token]
 
@@ -49,31 +49,32 @@ class Webui::Users::TokensController < Webui::WebuiController
   end
 
   def set_params
-    @params = params.require(:token).except(:string_readonly).permit(:operation, :project_name, :package_name, :scm_token).tap do |token_parameters|
-      token_parameters.require(:operation)
+    @params = params.except(:project_name, :package_name).require(:token).except(:string_readonly).permit(:type, :scm_token).tap do |token_parameters|
+      token_parameters.require(:type)
     end
+    @extra_params = params.slice(:project_name, :package_name).permit!
   end
 
   def set_package
-    return if @params[:project_name].blank? && @params[:package_name].blank?
+    return if @extra_params[:project_name].blank? && @extra_params[:package_name].blank?
 
     # Prevent setting a package for a workflow token
-    return if @params[:operation] == 'workflow'
+    return if @params[:type] == 'workflow'
 
     # Check if only project_name or only package_name are present
-    if @params[:project_name].present? ^ @params[:package_name].present?
+    if @extra_params[:project_name].present? ^ @extra_params[:package_name].present?
       flash.now[:error] = 'When providing an optional package, both Project name and Package name must be provided.'
       render partial: 'create' and return
     end
 
     # If both project_name and package_name are present, check if this is an existing package
     begin
-      @package = Package.get_by_project_and_name(@params[:project_name], @params[:package_name])
+      @package = Package.get_by_project_and_name(@extra_params[:project_name], @extra_params[:package_name])
     rescue Project::UnknownObjectError
-      flash.now[:error] = "When providing an optional package, the package must exist. Project '#{@params[:project_name]}' does not exist."
+      flash.now[:error] = "When providing an optional package, the package must exist. Project '#{@extra_params[:project_name]}' does not exist."
       render partial: 'create'
     rescue Package::UnknownObjectError
-      flash.now[:error] = "When providing an optional package, the package must exist. Package '#{@params[:project_name]}/#{@params[:package_name]}' does not exist."
+      flash.now[:error] = "When providing an optional package, the package must exist. Package '#{@extra_params[:project_name]}/#{@extra_params[:package_name]}' does not exist."
       render partial: 'create'
     end
   end
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -1,6 +1,6 @@
 class Webui::Users::TokensController < Webui::WebuiController
   before_action :set_token, only: [:destroy]
-  before_action :set_params, :set_package, :set_scm_token, only: [:create]
+  before_action :set_params, :set_package, only: [:create]
 
   after_action :verify_authorized, except: :index
   after_action :verify_policy_scoped, only: :index
@@ -15,18 +15,19 @@ class Webui::Users::TokensController < Webui::WebuiController
   end
 
   def create
-    @token = Token.token_type(@params[:operation]).new(@params.merge(user: User.session))
+    @token = Token.token_type(@params[:operation]).new(@params.merge(user: User.session, package: @package))
 
     authorize [:webui, @token]
 
-    @token.save
-
     respond_to do |format|
       format.js do
-        render partial: 'create', locals: {
-          string: @token.string,
-          flash: { success: "Token successfully created! Make sure you save it - you won't be able to access it again." }
-        }
+        if @token.save
+          flash.now[:success] = "Token successfully created! Make sure you save it - you won't be able to access it again."
+          render partial: 'create', locals: { string: @token.string }
+        else
+          flash.now[:error] = "Failed to create token: #{@token.errors.full_messages.to_sentence}."
+          render partial: 'create'
+        end
       end
     end
   end
@@ -48,21 +49,32 @@ class Webui::Users::TokensController < Webui::WebuiController
   end
 
   def set_params
-    @params = params.require(:token).except(:string_readonly).permit(:operation, :scm_token).tap do |token_parameters|
+    @params = params.require(:token).except(:string_readonly).permit(:operation, :project_name, :package_name, :scm_token).tap do |token_parameters|
       token_parameters.require(:operation)
     end
-    @package_params = params.slice(:project_name, :package_name).permit!
   end
 
   def set_package
-    return unless @package_params[:project_name].present? && @package_params[:package_name].present?
+    return if @params[:project_name].blank? && @params[:package_name].blank?
 
-    @package = Package.get_by_project_and_name(@package_params[:project_name], @package_params[:package_name])
-  end
+    # Prevent setting a package for a workflow token
+    return if @params[:operation] == 'workflow'
 
-  def set_scm_token
-    return unless @params[:operation] == 'workflow'
+    # Check if only project_name or only package_name are present
+    if @params[:project_name].present? ^ @params[:package_name].present?
+      flash.now[:error] = 'When providing an optional package, both Project name and Package name must be provided.'
+      render partial: 'create' and return
+    end
 
-    @scm_token = @params[:scm_token]
+    # If both project_name and package_name are present, check if this is an existing package
+    begin
+      @package = Package.get_by_project_and_name(@params[:project_name], @params[:package_name])
+    rescue Project::UnknownObjectError
+      flash.now[:error] = "When providing an optional package, the package must exist. Project '#{@params[:project_name]}' does not exist."
+      render partial: 'create'
+    rescue Package::UnknownObjectError
+      flash.now[:error] = "When providing an optional package, the package must exist. Package '#{@params[:project_name]}/#{@params[:package_name]}' does not exist."
+      render partial: 'create'
+    end
   end
 end
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -1,11 +1,34 @@
 class Webui::Users::TokensController < Webui::WebuiController
   before_action :set_token, only: [:destroy]
+  before_action :set_params, :set_package, :set_scm_token, only: [:create]
 
   after_action :verify_authorized, except: :index
   after_action :verify_policy_scoped, only: :index
 
   def index
     @tokens = policy_scope([:webui, Token]).page(params[:page])
+  end
+
+  def new
+    @token = Token.new
+    authorize [:webui, @token]
+  end
+
+  def create
+    @token = Token.token_type(@params[:operation]).new(@params.merge(user: User.session))
+
+    authorize [:webui, @token]
+
+    @token.save
+
+    respond_to do |format|
+      format.js do
+        render partial: 'create', locals: {
+          string: @token.string,
+          flash: { success: "Token successfully created! Make sure you save it - you won't be able to access it again." }
+        }
+      end
+    end
   end
 
   def destroy
@@ -22,5 +45,21 @@ class Webui::Users::TokensController < Webui::WebuiController
   rescue ActiveRecord::RecordNotFound => e
     flash[:error] = e.message
     redirect_to tokens_url
+  end
+
+  def set_params
+    @params = params.require(:token).except(:string_readonly).permit(:operation, :project_name, :package_name, :scm_token).tap do |token_parameters|
+      token_parameters.require(:operation)
+    end
+  end
+
+  def set_package
+    @package = Package.get_by_project_and_name(@params[:project_name], @params[:package_name]) if @params[:project_name].present? && @params[:package_name].present?
+  end
+
+  def set_scm_token
+    return unless @params[:operation] == 'workflow'
+
+    @scm_token = @params[:scm_token]
   end
 end
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -48,13 +48,16 @@ class Webui::Users::TokensController < Webui::WebuiController
   end
 
   def set_params
-    @params = params.require(:token).except(:string_readonly).permit(:operation, :project_name, :package_name, :scm_token).tap do |token_parameters|
+    @params = params.require(:token).except(:string_readonly).permit(:operation, :scm_token).tap do |token_parameters|
       token_parameters.require(:operation)
     end
+    @package_params = params.slice(:project_name, :package_name).permit!
   end
 
   def set_package
-    @package = Package.get_by_project_and_name(@params[:project_name], @params[:package_name]) if @params[:project_name].present? && @params[:package_name].present?
+    return unless @package_params[:project_name].present? && @package_params[:package_name].present?
+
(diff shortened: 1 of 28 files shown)


src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -0,0 +1,24 @@
+class Webui::Users::TokensController < Webui::WebuiController
+  before_action :set_token, only: [:destroy]
+
+  after_action :verify_authorized, except: :index
+  after_action :verify_policy_scoped, only: :index
+
+  def index
+    @tokens = policy_scope([:webui, Token])
```
krauselukas, line 8:
```suggestion
    @tokens = policy_scope([:webui, Token]).page(params[:page])
```

krauselukas: Could be a oneliner

eduardoj: Done.

src/api/app/policies/webui/token/release_policy.rb
```diff
@@ -0,0 +1 @@
+class Webui::Token::ReleasePolicy < Webui::TokenPolicy; end
```
hennevogel, line 1: Why namespace the policy?

eduardoj: From the description of the PR: "Policies for using tokens from the web-ui where created under `app/policies/webui/tokens/`, to not colide with the existing token policies, under `app/policies/webui/tokens/`."

vpereira:
Hey, I would like to raise some concerns regarding the logic being implemented. I'm not sure if they are already planned, but I would like to document those in my comment. 

Screenshot_2021-07-26_16-30-35

The column ID is unnecessarily long. Maybe we should introduce a name for the token. Looking my Gitlab/Github tokens, I like to name them, to know where I'm using them. It makes easier  to manage them, specially for long term tokens.

Screenshot_2021-07-26_16-31-29

Today with `osc token`, I'm able to create a token for a project globally, without assign it to a specific package. I.e `osc token --create home:vpereirabr` works without a problem. However the WebUI is behaving differently here:

Screenshot_2021-07-26_16-31-29

We are still missing the "edit" button, where we would be able for example for `workflow tokens` to update the SCM token without displaying it.. will that be implemented in this deliverable as spec'ed here https://trello.com/c/c1kcRa5n/1722-crud-ui-for-tokens?

eduardoj:
> The column ID is unnecessarily long.

This should be fixed since last week.

> Maybe we should introduce a name for the token. Looking my Gitlab/Github tokens, I like to name them, to know where I'm using them. It makes easier to manage them, specially for long term tokens.

It is also convinient to have a `created_at` field, to help to remember the tokens. We don't have the name field, neither the `created_at` field. In case of introducing those fields, I would prefer that to be done in another pull request.

> Today with `osc token`, I'm able to create a token for a project globally, without assign it to a specific package. I.e `osc token --create home:vpereirabr` works without a problem. However the WebUI is behaving differently here:

The error message you receive, explains it: "when providing an optional package, both project name and package name must be provided.". You only provided a project name. The same error happens if only a package name (without a project name) is provided.

> We are still missing the "edit" button, where we would be able for example for `workflow tokens` to update the SCM token without displaying it.. will that be implemented in this deliverable as spec'ed here https://trello.com/c/c1kcRa5n/1722-crud-ui-for-tokens?

I still miss it, too. I hope that it will be implemented in this deliverable. The edit button is not part of this pull request, so it will be easier to review. As soon as we have the list, create and delete functionality finished, I'm more than happy. Later on, we could add the "edit" button.

vpereira:
> The error message you receive, explains it: "when providing an optional package, both project name and package name must be provided.". You only provided a project name. The same error happens if only a package name (without a project name) is provided.

Ok sounds like if I run `osc token --create $PROJECT` the token get created but not assigned to any project, right? So you cannot have a token valid o all packages under a specific project?

vpereira:
> > The column ID is unnecessarily long.
> 
> This should be fixed since last week.

So maybe the review-app didn't redeploy it? https://obs-reviewlab.opensuse.org/eduardoj-webui_for_user_tokens/my/tokens

eduardoj:
> Ok sounds like if I run `osc token --create $PROJECT` the token get created but not assigned to any project, right?

Right.

> So you cannot have a token valid to all packages under a specific project?

Exactly, you cannot have a token valid to all packages under a specific project. These are the fields returned in the current table for tokens:
```
> Token
=> Token(id: integer, string: string, user_id: integer, package_id: integer, type: string, scm_token: string)
>
```
In case we want to provide that funcionality, we should add, at least, a `project_id` field to that database table.

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+    authorize [:webui, @token]
+    @token.destroy
+    flash[:success] = 'Token was successfully deleted.'
+    redirect_to tokens_url
+  end
+
+  private
+
+  def set_token
+    @token = Token.find(params[:id])
```
dmarcoux, line 48: I would move this code in `destroy` since it's only needed for this action. This would make the controller a tiny bit simpler with one less `before_action`.

src/api/app/views/webui/users/tokens/index.html.haml
```diff
@@ -0,0 +1,60 @@
+- @pagetitle = 'Tokens'
+
+.card
+  .card-body
+    %h3= @pagetitle
+
+    .text-right
+      = link_to new_token_path do
+        %i.fas.fa-plus-circle.text-primary
+        Create Token
```
dmarcoux, line 10: Comparing with other UIs, we have `Create (...)` link at the bottom left of lists. Examples:
- Create Package: https://obs-reviewlab.opensuse.org/eduardoj-webui_for_user_tokens/project/show/home:Admin
- Add Member: https://obs-reviewlab.opensuse.org/eduardoj-webui_for_user_tokens/groups/group_1

src/api/app/controllers/webui/users/tokens_controller.rb
eduardoj: > I would move this code in `destroy` since it's only needed for this action. This would make the controller a tiny bit simpler with one less `before_action`.

In case we implement the update and trigger actions, it would be nice to have the `before_action` back.

src/api/app/views/webui/users/tokens/index.html.haml
```diff
@@ -0,0 +1,60 @@
...
+              %tr
+                %td.col-1= token.id
+                %td.col-2= token.class.token_name.capitalize
+                %td.col-auto= link_to("#{truncate(token.package.project.name, length: 80)}/#{truncate(token.package.name, length: 80)}",
+                  package_show_path(project: token.package.project, package: token.package)) if token.package
+                %td.col-1
+                  = link_to '#', title: 'Delete Token', class: 'text-nowrap',
+                    data: { toggle: 'modal', target: '#delete-token-modal', token_id: token.id, action: token_path(token) } do
+                    %i.fas.fa-times-circle.text-danger
+                    Delete
```
dmarcoux, line 34: In tables like in [`Source Files` of a package](https://obs-reviewlab.opensuse.org/eduardoj-webui_for_user_tokens/package/show/home:Admin/ruby), we don't have text for `delete` icon.

```suggestion
```

src/api/app/views/webui/users/tokens/index.html.haml
eduardoj: Ok. I will move it to the bottom left of the list.

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def set_params
+    @params = params.require(:token).except(:string_readonly).permit(:operation, :scm_token).tap do |token_parameters|
+      token_parameters.require(:operation)
+    end
+    @package_params = params.slice(:project_name, :package_name).permit!
+  end
+
+  def validate_package
+    return unless @package_params[:project_name].present? ^ @package_params[:package_name].present?
```
dmarcoux, line 59: Is this a typo or you really want to use `^` (binary XOR operator) here?

eduardoj: I wanted to use the `^`  (binary XOR operator) here. For fetching a package we need a project name and a package name. Only a project name or only a package name is not enough to retrieve a package. See also [this comment](#issuecomment-886793814).

src/api/app/views/webui/users/tokens/index.html.haml
eduardoj: Applied.

src/api/app/views/webui/users/tokens/index.html.haml
eduardoj: Done.

src/api/app/controllers/webui/users/tokens_controller.rb
dmarcoux: If it's going to be used in some other upcoming actions, then fine :+1:.

src/api/app/controllers/webui/users/tokens_controller.rb
dmarcoux: So having the more commonly used `&&` would also be fine here. Then this would become:

```suggestion
    return unless @package_params[:project_name].present? && @package_params[:package_name].present?
```

src/api/app/policies/webui/token_policy.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    def resolve
+      scope.where(user: user).where.not(type: 'Token::Rss').includes([:package])
+    end
+  end
+
+  def create?
+    # TODO: when user_profile_redesign is rolled out, remove the Flipper check
+    Flipper.enabled?(:user_profile_redesign, user)
```
dmarcoux, line 16: We should also check if the token to be created is assigned to the user, no?  

```suggestion
    record.user == user && Flipper.enabled?(:user_profile_redesign, user)
```

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      @package = Package.get_by_project_and_name(@package_params[:project_name], @package_params[:package_name])
+    rescue Package::UnknownObjectError
+      render partial: 'create', locals: {
+        flash: { error: 'Failed to create token: when providing an optional package, the package must exist. '\
+                        "Package '#{@package_params[:project_name]}/#{@package_params[:package_name]}' does not exist." }
+      }
+    end
+  end
+
+  def set_scm_token
```
dmarcoux, line 79: This could go in the `initialize` method of `Token::Workflow` since it's only for that model.

dmarcoux:
@eduardoj @krauselukas As discussed in our programming session, I pushed 2 refactoring commits to the PR.

src/api/app/policies/webui/token/release_policy.rb
hennevogel: I understand namespaces, but why do this? You expect them to behave differently then the API? And if so how?

src/api/app/policies/webui/token_policy.rb
eduardoj: Fixed.

hennevogel:
Would it be possible to hide the project/package fields for the workflow token?

eduardoj:
> Would it be possible to hide the project/package fields for the workflow token?

Only for the workflow token... sure. On it.

src/api/app/views/webui/users/tokens/new.html.haml
```diff
@@ -0,0 +1,84 @@
...
+                  .custom-control.custom-radio.custom-control-inline
+                    = radio.radio_button(class: 'custom-control-input')
+                    = radio.label(class: 'custom-control-label')
+
+          .form-row
+            .col
+              = render partial: 'webui/shared/autocomplete', locals: { html_id: 'token[project_name]', label: 'Project name', required: false,
+                                                                       data: { source: autocomplete_projects_path(local: true) } }
+            .col
+              = render partial: 'webui/shared/autocomplete', locals: { html_id: 'token[package_name]', label: 'Package name', required: false,
```
hennevogel, line 35: This field should be disabled until the project field is filled.

eduardoj:
> > > The column ID is unnecessarily long.
> > 
> > This should be fixed since last week.
> 
> Maybe we didn't test it extensively on Google Chrome?

@vpereira: The `col-*` classes have been removed from the `td` elements. I found a reproducer with Google Chromium, and it is fixed now. Could you confirm that is working for you?

src/api/app/views/webui/users/tokens/new.html.haml
```diff
@@ -0,0 +1,84 @@
+- operation_options = Token::OPERATIONS.map { |a| "Token::#{a}".constantize.token_name }
```
hennevogel, line 1: workflow only if you are in beta.

eduardoj:
> > Would it be possible to hide the project/package fields for the workflow token?
> 
> Only for the workflow token... sure. On it.

@hennevogel: Fixed. Kudos to @krauselukas.

src/api/app/views/webui/users/tokens/new.html.haml
eduardoj: I changed the flag checked from `user_profile_redesign` to `trigger_workflow`. I also updated the PR description. @hennevogel please review again.

src/api/app/views/webui/users/tokens/new.html.haml
eduardoj: Done.

vpereira:
Hi @eduardoj 

I think we have some issues with the mobile resolution:

Screenshot_2021-07-28_14-09-53

The search boxes beside each other makes hard to work with autocomplete, the boxes are quite small too

Screenshot_2021-07-28_14-10-11

I think we do style the buttons differently on mobile no?

Screenshot_2021-07-28_14-10-36

The button "delete" is not usable in this resolution

eduardoj:
> I think we have some issues with the mobile resolution:

Out of curiosity, what viewport resolution are you using? Less than 320 pixels width?

> The search boxes beside each other makes hard to work with autocomplete, the boxes are quite small too

I will modify the form to wrap the input boxes in small width vieports.

> I think we do style the buttons differently on mobile no?

I suppose you're referring to their placement, not to show them without space between them. I'll fix that. Please correct me if you wanted to say something different.

> The button "delete" is not usable in this resolution

I will add padding to them.

vpereira:
> Out of curiosity, what viewport resolution are you using? Less than 320 pixels width?

I'm using Iphone 5S

eduardoj:
> > The search boxes beside each other makes hard to work with autocomplete, the boxes are quite small too
> 
> I will modify the form to wrap the input boxes in small width vieports.

Done.

> > I think we do style the buttons differently on mobile no?
> 
> I suppose you're referring to their placement, not to show them without space between them. I'll fix that. Please correct me if you wanted to say something different.

The placement and distance between buttons has been adjusted. See this point and the previous one in this screenshot:

Screenshot from 2021-07-29 08-40-36

> > The button "delete" is not usable in this resolution
> 
> I will add padding to them.

I added a commit to add button classes to the icon. So it will be easier to click, also in mobile devices.

I also added a class to wrap long package names:

Screenshot from 2021-07-29 08-37-46

vpereira (approved):
approved. Maybe would be nice to squash some commits?

src/api/app/models/token.rb
```diff
@@ -41,6 +47,30 @@ def package_find_options
   def follow_links?
     package_find_options[:follow_multibuild] || package_find_options[:follow_project_links]
   end
+
+  protected
+
+  def validate_package
```
hennevogel, line 53: This part is not as simple as that. Each token type has it's own way of finding a package with various options and quirks. Including which names it accepts etc.

src/api/app/views/webui/users/tokens/_actions.html.haml
```diff
@@ -0,0 +1 @@
+= link_to('Back to List of Tokens', tokens_path, title: 'Back to List of Tokens', class: 'btn btn-outline-secondary px-4 mt-3 mt-sm-0')
```
hennevogel, line 1: 💭  Browsers have back buttons, is this really needed?

eduardoj:
> Maybe would be nice to squash some commits?

I'm up for it. What of them, exactly, if I may ask?

src/api/spec/controllers/webui/users/tokens_controller_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+        create(:service_token, user: user)
+        create(:workflow_token, user: user)
+        create(:rss_token, user: user)
+        create(:release_token, user: other_user)
+
+        login user
+        get :index
+      end
+
+      it { expect(response).to render_template('webui/users/tokens/index') }
```
hennevogel, line 30: What would stop this from functioning? :)

src/api/spec/controllers/webui/users/tokens_controller_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+  let(:user) { create(:confirmed_user, login: 'foo') }
+  let(:other_user) { create(:confirmed_user, login: 'bar') }
+
+  before do
+    Flipper[:trigger_workflow].enable
+  end
+
+  describe 'GET #index' do
+    context 'user is not logged in' do
+      it_behaves_like 'require logged in user' do
```
hennevogel, line 13: `require_login` is already tested in the `WebuiController` spec. No need to do it again here for every action.

vpereira:
> I'm up for it. What of them, exactly, if I may ask?

I don't know, If all the introduced changes just work together, 1 commit: "Introduce tokens in web UI", would be enough. The little single commits are good to have in the PR, but they shouldn't land in master as separated unities. Like you added first the index, then you added the button for create a new token, then you added a button to delete.. those should be just one commit, IMO.

src/api/spec/controllers/webui/users/tokens_controller_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+      before do
+        login user
+      end
+
+      subject { post :create, xhr: true, params: form_parameters }
+
+      context 'no project and package parameters' do
+        let(:form_parameters) { { token: { operation: 'release' } } }
+
+        include_examples 'check for flashing a success'
```
hennevogel, line 97: I would feel more comfortable if you would test that the things you want to change, change. Instead of testing this only from the "outside" via the flash.

hennevogel (changes requested):

src/api/app/models/token.rb
```diff
@@ -3,15 +3,21 @@ class Token < ApplicationRecord
...
+  attr_accessor :object_to_authorize, :operation, :project_name, :package_name
 
   has_secure_token :string
 
   validates :user, presence: true
   validates :string, uniqueness: { case_sensitive: false }
+  validates :scm_token, absence: true, if: -> { type != 'Token::Workflow' }
+  validate :validate_package
+  # TODO: validate_package: Is this for all tokens except Token::Workflow?
+  # TODO: validate_package: Check if this is affecting the API or maybe even osc
```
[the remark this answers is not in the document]
eduardoj: Let's do it now.

src/api/spec/controllers/webui/users/tokens_controller_spec.rb
eduardoj: Lines `it { expect(response).to render_template('...') }` removed. As this was the only thing that was tested for tne `new` action, that test was also removed.

src/api/spec/controllers/webui/users/tokens_controller_spec.rb
eduardoj: "require logged in user" tests removed.

src/api/spec/controllers/webui/users/tokens_controller_spec.rb
eduardoj: I added tests for checking the number of tokens. Please have a look again.

src/api/app/views/webui/users/tokens/_actions.html.haml
eduardoj: In this case I would say yes. This 'Back to List of Tokens' link will be shown after a token has been created. If you press the back button of your browser, you will not request a page in the server, but you will still see the previous list of tokens, without the newly created one. It doesn't refresh automatically.

Do you have some other solution in mind?

src/api/app/models/token.rb
eduardoj: I moved back the validation of the package parameters (project_name and package_name) and its search to the controller.

src/api/app/models/token.rb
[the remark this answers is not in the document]
eduardoj: After removing `validate_package` from the Token model, both TODOs can be removed.

src/api/app/models/token.rb
```diff
@@ -3,15 +3,18 @@ class Token < ApplicationRecord
   has_many :event_subscriptions, dependent: :destroy
   belongs_to :package, inverse_of: :tokens
 
-  attr_accessor :object_to_authorize
+  attr_accessor :object_to_authorize, :operation, :project_name, :package_name
```
hennevogel, line 6: No need for those anymore right?

vpereira (approved):

hennevogel (changes requested):

src/api/app/models/token.rb
eduardoj: Nop. Removed.

hennevogel (approved):

PR #11401, opened 2021-07-20
Require rack version we have in the bundle

hennevogel:
After merge we need to link to [devel:languages:ruby:extensions/rubygem-rack](https://build.opensuse.org/package/show/devel:languages:ruby:extensions/rubygem-rack) in 2.10:Staging.


M0ses (approved):
LGTM

eduardoj (approved):

PR #11407, opened 2021-07-21
Enable rendered_view_with_filenames annotation in development

vpereira:
Enable annotation in development mode to make easier to identify which file we have to change when updating views.

More info: https://www.bigbinary.com/blog/rails-6-1-adds-annotate_rendered_view_with_filenames-to-annotate-html-output

-----------------

### DIFF
--- a/src/api/config/environments/development.rb
+++ b/src/api/config/environments/development.rb
@@ -72,7 +72,7 @@ OBSApi::Application.configure do
   # config.i18n.raise_on_missing_translations = true
 
   # Annotate rendered view with file names.
-  # config.action_view.annotate_rendered_view_with_filenames = true
+  config.action_view.annotate_rendered_view_with_filenames = true
 
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.


krauselukas (commented):
Nice feature, but somehow I doesn't work for me, I tested both in Chrome and Firefox, no output in the inspector. @vpereira any idea?

krauselukas:
> Nice feature, but somehow I doesn't work for me, I tested both in Chrome and Firefox, no output in the inspector. @vpereira any idea?

Could it be that this is only supported for `.erb` at the moment? https://github.com/rails/rails/pull/38848 & https://blog.saeloun.com/2020/05/11/rails-support-annotates-html-output-with-template-file-names.html

vpereira:
> Could it be that this is only supported for `.erb` at the moment? [rails/rails#38848](https://github.com/rails/rails/pull/38848) & https://blog.saeloun.com/2020/05/11/rails-support-annotates-html-output-with-template-file-names.html

Yes https://github.com/haml/haml-rails/issues/172 I would however update the configuration now, since it doesnt have any side effect and we get it automatically enabled as soon as haml support it

dmarcoux (approved):
Let's enable this and as @vpereira said, whenever this is supported in Haml, then we'll get it.

PR #11413, opened 2021-07-22
Implement step to link a package to a project

rubhanazeem:
The SCM integration allows triggering a new step, LinkPackageStep, when some event happens in the SCM side (new/updated PR/MR).

With this step, the user can create a package linked to another existing source package. This functionality is based on `osc linkpac`.

We plan to ignore the `codeclimate` check complaint. 

~DO NOT MERGE** until PR #11433 is merged.~

### DIFF
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/ClassLength
 class Workflow::Step
   include ActiveModel::Model
 
@@ -47,9 +48,7 @@ class Workflow::Step
   end
 
   def remote_source?
-    return true if Project.find_remote_project(source_project_name)
-
-    false
+    Project.find_remote_project(source_project_name).present?
   end
 
   def add_or_update_branch_request_file(package:)
@@ -84,6 +83,7 @@ class Workflow::Step
       object_attributes: { source: { default_branch: scm_extractor_payload[:commit_sha] } } }.to_json
   end
 
+  # FIXME: remove this and use create_subscriptions and update_subscriptions as soon as BranchPackageStep is refactored
   def create_or_update_subscriptions(package, workflow_filters)
     ['Event::BuildFail', 'Event::BuildSuccess'].each do |build_event|
       subscription = EventSubscription.find_or_create_by!(eventtype: build_event,
@@ -93,6 +93,29 @@ class Workflow::Step
                                                           enabled: true,
                                                           token: @token,
                                                           package: package)
+      subscription.update!(payload: scm_extractor_payload.merge({ workflow_filters: workflow_filters }))
+    end
+  end
+
+  def create_subscriptions(package, workflow_filters)
+    ['Event::BuildFail', 'Event::BuildSuccess'].each do |build_event|
+      EventSubscription.create!(eventtype: build_event,
+                                receiver_role: 'reader', # We pass a valid value, but we don't need this.
+                                user: @token.user,
+                                channel: 'scm',
+                                enabled: true,
+                                token: @token,
+                                package: package,
+                                payload: scm_extractor_payload.merge({ workflow_filters: workflow_filters }))
+    end
+  end
+
+  def update_subscriptions(package, workflow_filters)
+    ['Event::BuildFail', 'Event::BuildSuccess'].each do |build_event|
+      subscription = EventSubscription.find_by(eventtype: build_event,
+                                               channel: 'scm',
+                                               token: @token,
+                                               package: package)
       subscription.update!(payload: scm_extractor_payload.merge({ workflow_filters: workflow_filters }))
     end
   end
@@ -121,3 +144,4 @@ class Workflow::Step
     architectures
   end
 end
+# rubocop:enable Metrics/ClassLength
--- a/src/api/app/models/workflow/step/link_package_step.rb
+++ b/src/api/app/models/workflow/step/link_package_step.rb
@@ -1,0 +1,96 @@
+class Workflow::Step::LinkPackageStep < ::Workflow::Step
+  validates :source_package_name, presence: true
+
+  def call(options = {})
+    return unless valid?
+
+    workflow_filters = options.fetch(:workflow_filters, {})
+
+    # Updated PR
+    if validator.updated_pull_request? && target_package.present?
+      update_subscriptions(target_package, workflow_filters)
+    else # New PR
+      create_target_package
+      create_subscriptions(target_package, workflow_filters)
+    end
+
+    add_or_update_branch_request_file(package: target_package)
+    report_to_scm(workflow_filters)
+    target_package
+  end
+
+  private
+
+  def target_project
+    Project.find_by(name: target_project_name)
+  end
+
+  def create_target_package
+    create_project_and_package
+    create_special_package
+    create_link
+  end
+
+  def create_project_and_package
+    check_source_access
+
+    raise PackageAlreadyExists, "Can not link package. The package #{target_package_name} already exists." if target_package.present?
+
+    if target_project.nil?
+      project = Project.create!(name: target_project_name)
+      project.commit_user = User.session
+      project.relationships.create!(user: User.session, role: Role.find_by_title('maintainer'))
+      project.store
+    end
+
+    target_project.packages.create(name: target_package_name)
+  end
+
+  # Will raise an exception if the source package is not accesible
+  def check_source_access
+    return if remote_source?
+
+    Package.get_by_project_and_name(source_project_name, source_package_name)
+  end
+
+  # NOTE: the next lines are a temporary fix the allow the service to run in a linked package. A project service is needed.
+  def create_special_package
+    return if Package.find_by_project_and_name(target_project, '_project')
+
+    special_package = target_project.packages.create(name: '_project')
+    special_package.save_file({ file: special_package_file_content, filename: '_service' })
+  end
+
+  def special_package_file_content
+    <<~XML
+      <services>
+        <service name="format_spec_file" mode="localonly"/>
+      </services>
+    XML
+  end
+
+  def create_link
+    Backend::Api::Sources::Package.write_link(target_project_name,
+                                              target_package_name,
+                                              @token.user,
+                                              link_xml(project: source_project_name, package: source_package_name))
+
+    target_package
+  end
+
+  def link_xml(opts = {})
+    # "<link package=\"foo\" project=\"bar\" />"
+    Nokogiri::XML::Builder.new { |x| x.link(opts) }.doc.root.to_s
+  end
+
+  def report_to_scm(workflow_filters)
+    workflow_repositories(target_project_name, workflow_filters).each do |repository|
+      # TODO: Fix n+1 queries
+      workflow_architectures(repository, workflow_filters).each do |architecture|
+        # We cannot report multibuild flavors here... so they will be missing from the initial report
+        SCMStatusReporter.new({ project: target_project_name, package: target_package_name, repository: repository.name, arch: architecture.name },
+                              scm_extractor_payload, @token.scm_token).call
+      end
+    end
+  end
+end
--- a/src/api/spec/models/workflow/step/link_package_step_spec.rb
+++ b/src/api/spec/models/workflow/step/link_package_step_spec.rb
@@ -1,0 +1,310 @@
+require 'rails_helper'
+
+RSpec.describe Workflow::Step::LinkPackageStep, vcr: true do
+  let!(:user) { create(:confirmed_user, :with_home, login: 'Iggy') }
+  let(:token) { create(:workflow_token, user: user) }
+  let(:target_project_name) { "home:#{user.login}:#{project.name}:PR-1" }
+
+  subject do
+    described_class.new(step_instructions: step_instructions,
+                        scm_extractor_payload: scm_extractor_payload,
+                        token: token)
+  end
+
+  RSpec.shared_context 'source_project not provided' do
+    let(:step_instructions) { { source_package: package.name } }
+
+    it { expect { subject.call }.not_to(change(Package, :count)) }
+  end
+
+  RSpec.shared_context 'source_package not provided' do
+    let(:step_instructions) { { source_project: package.project.name } }
+
+    it { expect { subject.call }.not_to(change(Package, :count)) }
+  end
+
+  RSpec.shared_context 'failed when source_package does not exist' do
+    let(:step_instructions) do
+      {
+        source_project: project.name,
+        source_package: 'this_package_does_not_exist'
+      }
+    end
+
+    it { expect { subject.call }.to raise_error(Package::Errors::UnknownObjectError) }
+  end
+
+  RSpec.shared_context 'project and package does not exist' do
+    let(:step_instructions) do
+      {
+        source_project: 'invalid project',
+        source_package: 'invalid package'
+      }
+    end
+
+    it { expect { subject.call }.to raise_error(Project::Errors::UnknownObjectError) }
+  end
+
+  RSpec.shared_context 'target package already exists' do
+    # Emulates that the target project and package already existed
+    let!(:linked_project) { create(:project, name: "home:#{user.login}:#{package.project.name}:PR-1", maintainer: user) }
+    let!(:linked_package) { create(:package_with_file, name: package.name, project: linked_project) }
+
+    let(:step_instructions) do
+      {
+        source_project: project.name,
+        source_package: package.name
+      }
+    end
+
+    it { expect { subject.call }.to raise_error(PackageAlreadyExists) }
+  end
+
+  RSpec.shared_context 'failed without link permissions' do
+    let(:step_instructions) do
+      {
+        source_project: project.name,
+        source_package: package.name
+      }
+    end
+
+    before do
+      # rubocop:disable RSpec/AnyInstance
+      allow_any_instance_of(Package).to receive(:disabled_for?).with('sourceaccess', nil, nil).and_return(true)
+      # rubocop:enable RSpec/AnyInstance
+    end
+
+    it { expect { subject.call }.to raise_error(Package::Errors::ReadSourceAccessError) }
+  end
+
+  RSpec.shared_context 'successful new PR or MR event' do
+    let(:step_instructions) do
+      {
+        source_project: package.project.name,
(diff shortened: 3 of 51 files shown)


src/api/app/controllers/source_controller.rb
```diff
@@ -358,7 +358,6 @@ def update_file
     end
 
     Package.verify_file!(@pack, params[:filename], request.raw_post)
-
```
vpereira, line 361: are those changes necessary?

src/api/app/controllers/source_project_package_meta_controller.rb
```diff
@@ -21,7 +21,6 @@ class SourceProjectPackageMetaController < SourceController
   # GET /source/:project/:package/_meta
   def show
     pack = Package.get_by_project_and_name(@project_name, @package_name, use_source: false)
-
```
vpereira, line 24: is it necessary?

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,216 @@
...
+
+        # TODO: The source code of linked package can be modified/updated?
+        # If so, a real file is created in the linked package because the source shouldn't be affected by the changes?
+        add_or_update_branch_request_file(package: linked_package)
+
+        workflow_filters = options.fetch(:workflow_filters, {})
+        create_or_update_subscriptions(linked_package, workflow_filters)
+
+        # TODO: remove this line
+        target_project.repositories << Repository.find(2)
```
vpereira, line 39: :eyes:

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,216 @@
+class LinkPackage; end
+module LinkPackage::Errors
```
vpereira, line 2: this should be moved to its own `errors.rb` file

src/api/.rubocop_todo.yml
```diff
@@ -655,6 +655,7 @@ RSpec/AnyInstance:
     - 'spec/models/issue_tracker_spec.rb'
     - 'spec/models/package_spec.rb'
     - 'spec/models/project_spec.rb'
+    - 'spec/models/workflow/step/link_package_step_spec.rb'
```
dmarcoux, line 658: It's preferable to disable this cop only on the line where this applies.

src/api/spec/models/workflow/step/link_package_step_spec.rb
```diff
@@ -0,0 +1,305 @@
...
+  RSpec.shared_context 'failed without link permissions' do
+    let(:step_instructions) do
+      {
+        source_project: project.name,
+        source_package: package.name
+      }
+    end
+
+    before do
+      allow_any_instance_of(Package).to receive(:disabled_for?).with('sourceaccess', nil, nil).and_return(true)
```
dmarcoux, line 71: It's not possible to mock this directly without `any_instance`?

saraycp: That's what we did at the beginning, however, we didn't manage to make the test pass. After hours of trying different alternatives, we decided to go back to `allow_any_instance_of`and not to block this PR anymore.

src/api/.rubocop_todo.yml
saraycp: Done

src/api/app/models/workflow/step.rb
```diff
@@ -25,4 +25,99 @@ def source_project_name
   def target_project_name
     "home:#{@token.user.login}:#{source_project_name}:PR-#{scm_extractor_payload[:pr_number]}"
   end
+
+  def source_package_name
```
dmarcoux, line 29: I don't think all steps really need this. So far, it's only 2 out of 3 steps: `link_package` and `branch_package`. I would prefer to keep the duplicated code in 2 steps and extract this if we happen to need the same code in a third step. This is a good example of the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,102 @@
...
+    Backend::Api::Sources::Package.write_link(target_project_name,
+                                              target_package_name,
+                                              @token.user,
+                                              link_xml(project: source_project_name, package: source_package_name))
+
+    target_package
+  end
+end
+
+# TODO: triger service is not working as expected for linked packages.
```
dmarcoux, line 101: This should be included in our upcoming blog post about this deliverable.

src/api/spec/models/workflow/step/link_package_step_spec.rb
```diff
@@ -0,0 +1,310 @@
...
+    it { expect { subject.call }.to(change(Package, :count).by(0)) }
+    it { expect { subject.call.source_file('_branch_request') }.not_to raise_error }
+    it { expect { subject.call.source_file('_link') }.not_to raise_error }
+    it { expect(subject.call.source_file('_link')).to eq('<link project="foo_project" package="bar_package"/>') }
+
+    it 'updates _branch_request file including new commit sha' do
+      expect(subject.call.source_file('_branch_request')).to include('456')
+    end
+
+    it { expect { subject.call }.to(change(EventSubscription.where(eventtype: 'Event::BuildFail'), :count).by(0)) }
```
dmarcoux, line 139: This is easier to understand. This comment applies to all similar examples.

```suggestion
    it { expect { subject.call }.not_to(change(EventSubscription.where(eventtype: 'Event::BuildFail'), :count)) }
```

src/api/spec/models/workflow/step/link_package_step_spec.rb
```diff
@@ -0,0 +1,310 @@
...
+  RSpec.shared_context 'successful new PR or MR event' do
+    let(:step_instructions) do
+      {
+        source_project: package.project.name,
+        source_package: package.name
+      }
+    end
+
+    it { expect { subject.call }.to(change(Project, :count).by(1)) }
+    it { expect { subject.call }.to(change(Package, :count).by(2)) }
```
dmarcoux, line 89: Is it really creating 2 packages? I think it should be 1 package.

dmarcoux (dismissed):

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    # NOTE: the next lines are a temporary fix the allow the service to run in a linked package. A project service is needed.
+    return if Package.find_by_project_and_name(target_project, '_project')
+
+    special_package = target_project.packages.create(name: '_project')
+    special_package.save_file({ file: special_package_file_content, filename: '_service' })
+  end
+
+  def find_or_create_linked_package
+    return target_package if validator.updated_pull_request? && target_package.present?
```
hennevogel, line 66: Anyone can make the condition true and give you any `target_package` they like in a request.

I think you should not mix creating and updating like this, KISS.

src/api/spec/models/workflow/step/link_package_step_spec.rb
rubhanazeem: It's creating two packages. The other one is [this](https://github.com/openSUSE/open-build-service/pull/11413/files#diff-b03f5a9fe996b716789616806c138c935118ab81fae1527c9bdcbd03739eb13bR61). This special package is needed to trigger the service.

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    create_project_and_package
+    link
+  end
+
+  def check_source_access
+    return if remote_source?
+
+    options = { use_source: false, follow_project_links: true, follow_multibuild: true }
```
hennevogel, line 75: `follow_multibuild: true` makes little sense for this I would say. As every flavor has the same object and you are only interested in the object right?

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    begin
+      src_package = Package.get_by_project_and_name(source_project_name, source_package_name, options)
```
hennevogel, line 78: If you are following project links then...

https://github.com/openSUSE/open-build-service/wiki/Links#instantiating-returns-nil

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,102 @@
...
+    rescue Package::UnknownObjectError
+      raise Workflow::Step::Errors::CanNotLinkPackageNotFound, "Package #{source_project_name}/#{source_package_name} not found, it could not be linked."
```
hennevogel, line 80: Not sure you need to catch and re-raise this just to add `...it could not be linked.`

hennevogel (changes requested):

src/api/app/models/workflow/step/errors.rb
```diff
@@ -0,0 +1,17 @@
+module Workflow::Step::Errors
+  extend ActiveSupport::Concern
+
+  class CanNotLinkPackageNotFound < APIError
```
vpereira, line 4: shouldn't this class name end with `Error` ?

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,102 @@
...
+  def call(options = {})
+    return unless valid?
+
+    linked_package = find_or_create_linked_package
+    add_or_update_branch_request_file(package: linked_package)
+    workflow_filters = options.fetch(:workflow_filters, {})
+    create_or_update_subscriptions(linked_package, workflow_filters)
+
+    workflow_repositories(target_project_name, workflow_filters).each do |repository|
+      # TODO: Fix n+1 queries
```
vpereira, line 13: are we going to fix the n+1 queries on this PR or later?

src/api/app/models/workflow/step/errors.rb
saraycp: Having these two files as an example, it seems that's not the tendency:

https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/workflow/errors.rb
https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/branch_package/errors.rb

src/api/app/models/workflow/step/link_package_step.rb
saraycp: We should fix that for both `link_package` and `branch_package` steps. But we don't plan to do it now.

src/api/app/models/workflow/step/link_package_step.rb
rubhanazeem: We have to remove this comment. The problem with the service is already fixed by introducing the special package called `_project`. [Here](https://github.com/openSUSE/open-build-service/pull/11413/files#diff-b03f5a9fe996b716789616806c138c935118ab81fae1527c9bdcbd03739eb13bR62-R66) you can check it out.

src/api/spec/models/workflow/step/link_package_step_spec.rb
saraycp: [Here](https://github.com/openSUSE/open-build-service/pull/11413/files#diff-b03f5a9fe996b716789616806c138c935118ab81fae1527c9bdcbd03739eb13bR62-R66) is the "patch" to solve the problem with the service.

src/api/spec/models/workflow/step/link_package_step_spec.rb
rubhanazeem: Done

src/api/app/models/workflow/step/link_package_step.rb
rubhanazeem: Alright, we have removed this option and decided to go with the default value for `follow_multibuild` i.e. `false`.

src/api/app/models/workflow/step/link_package_step.rb
rubhanazeem: We are now handling the case where we get nil because the project is a remote one. We are now raising an exception in this scenario.

src/api/app/models/workflow/step/link_package_step.rb
rubhanazeem: We have removed the custom exception and let the original one raise.

src/api/app/models/workflow/step.rb
hennevogel: `Workflow::Step` might not be the right place, I agree. But for *sure* we should not have this complicated code twice.

saraycp: @dmarcoux , we can understand your point, but keep in mind it's not only the `source_package_name` method, there are more than 10 methods that are identical in both `LinkPackageStep` and `BranchPackageStep`.
Take also into account that, at the time we implemented this, we had 2 out of 2 steps containing the same code.

We think all of this will change as we are introducing new steps. We prefer to avoid duplication now and plan better refactoring for the near future. Anyway, we are going to move the extraction to an independent PR (which is in [this commit](https://github.com/openSUSE/open-build-service/pull/11413/commits/acadd11f6be3dcd1bbaf7761dd108be33a1fda08)).

saraycp: This code has been moved to a new PR: #11433

src/api/app/models/workflow/step/errors.rb
dmarcoux: Having `Errors` in the module is already enough. It results in `Workflow::Step::Errors::CanNotLinkPackageNotFound`.

hennevogel: it isn't used anymore right?

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,107 @@
...
+    link
+  end
+
+  def check_source_access
+    return if remote_source?
+
+    options = { use_source: false, follow_project_links: true }
+    src_package = Package.get_by_project_and_name(source_project_name, source_package_name, options)
+
+    raise Package::Errors::UnknownObjectError if src_package.nil?
```
hennevogel, line 78: Hm why? If it's nil we know it's "below" a valid remote project. The assumption then is you have access.

hennevogel (changes requested):

src/api/app/models/workflow/step/errors.rb
rubhanazeem: Yes, we have removed this file.

src/api/app/models/workflow/step/errors.rb
```diff
@@ -0,0 +1,17 @@
...
+    setup 404
+  end
+
+  class DuplicatePackageError < APIError
```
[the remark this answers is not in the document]
rubhanazeem: Done

src/api/app/models/workflow/step/link_package_step.rb
rubhanazeem: Done.

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,99 @@
...
+    link
+  end
+
+  def check_source_access
+    return if remote_source?
+
+    options = { use_source: false, follow_project_links: true }
+    src_package = Package.get_by_project_and_name(source_project_name, source_package_name, options)
+
+    Pundit.authorize(@token.user, src_package, :link?) if src_package.present?
```
hennevogel, line 72: This is the same as `use_source: true`

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,99 @@
...
+  def create_linked_package
+    create_project_and_package
+    create_special_package
+    link
+  end
+
+  def check_source_access
+    return if remote_source?
+
+    options = { use_source: false, follow_project_links: true }
```
hennevogel, line 69: `follow_project_links` is on by default.

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    Pundit.authorize(@token.user, src_package, :link?) if src_package.present?
+  end
+
+  def link_xml(opts = {})
+    # "<link package=\"foo\" project=\"bar\" />"
+    Nokogiri::XML::Builder.new { |x| x.link(opts) }.doc.root.to_s
+  end
+
+  def link
```
hennevogel, line 80: 💭 this sounds like `create_link` to me.

saraycp: done

src/api/app/models/workflow/step/link_package_step.rb
saraycp: true! done

src/api/app/models/workflow/step/link_package_step.rb
saraycp: We removed that line and now use use `use_source: true` which is the default value, so we expect to get a `Package::Errors::ReadSourceAccessError` exception when the user can't access it.

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  # NOTE: the next lines are a temporary fix the allow the service to run in a linked package. A project service is needed.
+  def create_special_package
+    return if Package.find_by_project_and_name(target_project, '_project')
+
+    special_package = target_project.packages.create(name: '_project')
+    special_package.save_file({ file: special_package_file_content, filename: '_service' })
+  end
+
+  def create_linked_package
```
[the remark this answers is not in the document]
saraycp: We have renamed the method

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -0,0 +1,99 @@
...
+  validates :source_package_name, presence: true
+
+  def call(options = {})
+    return unless valid?
+
+    workflow_filters = options.fetch(:workflow_filters, {})
+
+    # Updated PR
+    if validator.updated_pull_request? && target_package.present?
+      linked_package = target_package
```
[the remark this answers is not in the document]
saraycp: We have changed the `call` method, we no longer use the variable `linked_package` but the method `target_package`.

hennevogel (approved):

PR #11419, opened 2021-07-26
Move from Leap 15.1 to Leap 15.3

vpereira:
Move our development environment from Leap 15.1 to Leap 15.3

To test it:

checkout the branch and run `docker-compose down && rake docker:build && docker-compose up`

-----------------

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,17 +1,17 @@
 version: "2.1"
 services:
   db:
-    image: registry.opensuse.org/obs/server/unstable/container/leap151/containers/openbuildservice/mariadb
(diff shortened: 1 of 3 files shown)


vpereira:
I removed review-app because it will run it on leap 15.1.. i however did a PR for the `obs-review-app` repository too https://github.com/openSUSE/obs-review-apps/blob/master/files/docker-compose.yml

dmarcoux (approved):

hennevogel:
You are going to have to tell people to update and how now.

dmarcoux:
As @vpereira wrote in the PR description, doing `docker-compose down && rake docker:build && docker-compose up` should be enough to update the development environment.

hennevogel:
I think we should also backport this to 2.10

PR #11425, opened 2021-07-27
Track RuboCop offenses in CodeClimate

dmarcoux:
To have a better understanding of the project's maintainability with CodeClimate's dashboard.


dmarcoux:
[Align RuboCop and CodeClimate rules/parameters](https://trello.com/c/LGB9wpQz/1718-align-rubocop-and-codeclimate-rules-parameters)

hennevogel:
Happens when you run this in the top level directory

`rubocop.ruby2.7 -c src/api/.rubocop.yml`

hennevogel:
All the settings in the file are also relative to `src/api`

dmarcoux:
I'm not sure if we can somehow replicate our weird RuboCop setup in CodeClimate.

dmarcoux:
After a break... I'm thinking this approach is not the right one. It's going to be a pain to maintain, always having to update the RuboCop version as defined in the `channel` field in `.codeclimate.yml`. It's simpler to do the opposite and rely on CodeClimate for checks which are the same for RuboCop and CodeClimate.

PR #11426, opened 2021-07-27
Display copypastable checkout instructions

### LINKED ISSUE #63: feature request: provide osc download url
It would be great if OBS would show a url on the overview page (e.g. under "my actions") that I can just copy and paste so that I can create a local clone on my computer, i.e. "create a local clone using 'osc ...'". something similar to what github has. Since I don't use the build server that often I always have to look up how to do this.

bmwiedemann:
Without this patch, thousands of OBS users had to spend multiple seconds for many packages to type or copy the right parts of the URL into a shell window.

We want do it like github and gitlab and make it easy for them.

Fixes #63

-----------------

### DIFF
--- a/src/api/app/views/webui/package/side_links/_checkout_package.html.haml
+++ b/src/api/app/views/webui/package/side_links/_checkout_package.html.haml
@@ -1,0 +1,9 @@
+%li
+  = link_to('#', aria: { controls: 'collapse-checkout', expanded: false },
+    'data-toggle': 'collapse', 'data-target': '#collapse-checkout-package', role: 'button') do
+    %i.fas.fa-download
+    Checkout Package
+  .collapse.collapse-checkout.border#collapse-checkout-package
+    %code
+      - api_url = ::Configuration.api_url ? " -A #{::Configuration.api_url}" : ''
+      osc#{api_url} checkout #{project}/#{package} && cd $_
--- a/src/api/app/views/webui/package/_side_links.html.haml
+++ b/src/api/app/views/webui/package/_side_links.html.haml
@@ -23,3 +23,5 @@
                                                                                   linkinfo: linkinfo, revision: revision }
 
   = render partial: 'extra_actions', locals: { project: project, package: package }
+
+  = render partial: 'webui/package/side_links/checkout_package', locals: { project: project, package: package }


src/api/app/views/webui/package/_files_view.html.haml
```diff
@@ -23,6 +23,7 @@
       = link_to(new_project_package_files_path(project, package)) do
         %i.fas.fa-plus-circle.text-primary
         Add File
+    = sprintf("checkout this package with osc -A %s checkout %s/%s && cd $_", "FIXME: https://api.opensuse.org", project, package)
```
darix, line 26: = "Checkout this package with osc -A #{api_url} checkout #{project}/#{package} && cd $_"

src/api/app/views/webui/package/_files_view.html.haml
```diff
@@ -23,6 +23,7 @@
       = link_to(new_project_package_files_path(project, package)) do
         %i.fas.fa-plus-circle.text-primary
         Add File
+    = "Checkout this package with osc -A {FIXME:api_url} checkout #{project}/#{package} && cd $_"
```
hennevogel, line 26: There is `Configuration.api_url` but that one is not necessarily set. You could display this only if it's set.

bmwiedemann: OK. did that.

bmwiedemann:
https://obs-reviewlab.opensuse.org/bmwiedemann-fix63/package/show/home:Admin/hello_world would work, but does not look nice. E.g. we could have a "Checkout Package" with an icon next to the "Add File" and that could expand or popup.

dmarcoux:
Thank you @bmwiedemann for submitting a PR for this :slightly_smiling_face:

The `Checkout Package` link doesn't really belong in the context of `Source Files`. I suggest moving the `Checkout Package` link with the other links at the top. See my screenshot:

move-checkout-package-link

dmarcoux:
I took a closer look to this and decided to put `Checkout Package` in the side links. I pushed the changes and here's a preview of how it looks now:

preview

dmarcoux:

hennevogel (approved):

src/api/app/views/webui/package/side_links/_checkout_package.html.haml
```diff
@@ -0,0 +1,8 @@
+%li
+  = link_to('#', aria: { controls: 'collapse-checkout', expanded: false }, 'data-toggle': 'collapse', href: '.collapse-checkout', role: 'button') do
```
hennevogel, line 2:
```suggestion
  = link_to('#', aria: { controls: 'collapse-checkout', expanded: false }, 'data-toggle': 'collapse', role: 'button') do
```

hennevogel: Or button with data-target would probably the better choice. 

https://getbootstrap.com/docs/4.6/components/collapse/

Will also make the spider test like it.

bmwiedemann: Just that diff did not work (was not expanding anymore), but thanks to your link to the docs I found something that works. Can it still be improved?

src/api/app/views/webui/package/side_links/_checkout_package.html.haml
```diff
@@ -0,0 +1,9 @@
+%li
+  = link_to('#', aria: { controls: 'collapse-checkout', expanded: false },
+    'data-toggle': 'collapse', 'data-target': '#collapse-checkout1', role: 'button') do
+    %i.fas.fa-download
+    Checkout Package
+  .collapse.collapse-checkout.border#collapse-checkout1
```
dmarcoux, line 6: A slightly improved name:
```suggestion
  .collapse.collapse-checkout.border#collapse-checkout-package
```

src/api/app/views/webui/package/side_links/_checkout_package.html.haml
```diff
@@ -0,0 +1,9 @@
+%li
+  = link_to('#', aria: { controls: 'collapse-checkout', expanded: false },
+    'data-toggle': 'collapse', 'data-target': '#collapse-checkout1', role: 'button') do
```
dmarcoux, line 3: A slightly improved name:
```suggestion
    'data-toggle': 'collapse', 'data-target': '#collapse-checkout-package', role: 'button') do
```

dmarcoux (approved):

hennevogel:
@bmwiedemann 💐

PR #11427, opened 2021-07-27
Refactor SearchController extracting query to query object

vpereira:
In this PR I did the following changes

- Extract complex query to a class
- Add specs for that new class
- Adapt SearchController to use it
- Update rubocop todo file

-----------------


hennevogel (changes requested):

src/api/.rubocop_todo.yml
```diff
@@ -293,6 +293,7 @@ Metrics/CyclomaticComplexity:
     - 'app/models/project_status/calculator.rb'
     - 'app/models/relationship.rb'
     - 'app/models/repository.rb'
+    - 'app/models/search_finder.rb'
```
dmarcoux, line 296: It's better now, but there's still work to reduce the complexity. I'm fine with that :+1:

src/api/app/models/search_finder.rb
```diff
@@ -0,0 +1,89 @@
+class SearchFinder
```
dmarcoux, line 1: This class is under `app/models`, but it's kind of a mix between a model with its validations and a query object. I'm not sure about this... but I don't really have a proposal right now.

vpereira: yes. I started it as query object, but @hennevogel asked me to move to the `app/models` as we do with `OwnerSearch` for example.. I think that's ok now (code will be definitely extracted from there later).

src/api/app/models/search_finder.rb
```diff
@@ -0,0 +1,89 @@
...
+  validates :what, inclusion: [:package, :project, :repository, :request, :person, :channel, :channel_binary, :released_binary, :issue]
+
+  def initialize(what:, search_items: [], render_all: false, params: {})
+    @what = what
+    @render_all = render_all
+    @search_items = search_items
+    @params = params
+    @included_classes = []
+    @preloaded_classes = []
+    @relation = nil
```
dmarcoux, line 15: This line isn't needed because an uninitialized instance variable is already going to evaluate to `nil`.

```suggestion
```

dmarcoux (changes requested):
Just a small change, then it's all good :+1:

dmarcoux (approved):

dmarcoux:
@hennevogel Please have another look

hennevogel (approved):

PR #11431, opened 2021-07-28
Have a maximum of 2 expectations per example in specs

dmarcoux:
1 expectation per example is way too strict.


saraycp (approved):

eduardoj (approved):
It is for me an improvement. I would also be fine with a higher number, or with disabling the cop.

dmarcoux:
> I would also be fine with a higher number, or with disabling the cop.

We can always readjust this number. As for disabling the cop, I do believe that it is beneficial. Most examples don't need multiple expectations, so it's fine to disable the cop for extreme cases like some feature specs.

PR #11433, opened 2021-07-28
Extract common code from BranchPackageStep

saraycp:
There will be some common code in both Workflow::Step::BranchPackageStep and the model Workflow::Step::LinkPackageStep that is to come. We move that code to Workflow::Step for now.

See this discussion to get some context: https://github.com/openSUSE/open-build-service/pull/11413#pullrequestreview-716833890

### DIFF
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -25,4 +25,99 @@ class Workflow::Step
   def target_project_name
     "home:#{@token.user.login}:#{source_project_name}:PR-#{scm_extractor_payload[:pr_number]}"
   end
+
+  def source_package_name
+    step_instructions['source_package']
+  end
+
+  def target_package_name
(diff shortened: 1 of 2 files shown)


dmarcoux:
Even if I think this is wrong as discussed in the linked PR, I won't block this PR... Duplication is much better if we know that the abstraction is wrong. None of that code is needed for the `configure_repositories` step.

dmarcoux:
[Find a good abstraction for common code of Workflow::Step::BranchPackageStep and Workflow::Step::LinkPackageStep](https://trello.com/c/SfgC8hsE/2304-find-a-good-abstraction-for-common-code-of-workflowstepbranchpackagestep-and-workflowsteplinkpackagestep)

dmarcoux:
We agreed on creating a card in our backlog to find a better abstraction.

dmarcoux (approved):

PR #11446, opened 2021-08-02
Update the SCM token of workflow tokens in the UI

vpereira:
As a first step, only the SCM token of workflow tokens can be updated. Updating other token types isn't possible right now, so the `Edit` icon isn't displayed for non-workflow tokens. Regenerating the string of a token is going to happen later.

Also made some changes to be consistent throughout the Create/Edit forms and with other forms like having breadcrumbs for the `Create Token` and `Edit Token` pages.

### DIFF
--- a/src/api/app/views/webui/users/tokens/edit.html.haml
+++ b/src/api/app/views/webui/users/tokens/edit.html.haml
@@ -1,0 +1,25 @@
+.card
+  .card-body
+    %h3.mb-3 Edit Token
+
+    .row
+      .col-12.col-md-10.col-lg-8
+        -# "as: :token" to have a consistent name across all various token classes
+        = form_for(@token, as: :token, url: token_path, method: :put, local: true) do |f|
+          .form-row#edit-token
+            .col-12.col-md-10.col-lg-9
+              .form-group
+                = f.label(:id, 'Id:')
+                = @token.id
+              .form-group
+                = f.label(:type, 'Operation:')
+                = @token.class.token_name.capitalize
+              .form-group
+                = f.label(:scm_token) do
+                  SCM token:
+                  %abbr.text-danger{ title: 'required' } *
+                .input-group
+                  = f.password_field(:scm_token, size: 80, class: 'form-control', placeholder: 'Please enter your new SCM token', required: true)
+          .actions
+            = link_to('Cancel', tokens_path, title: 'Cancel', class: 'btn btn-outline-secondary px-4 mr-3 mt-3 mt-sm-0')
+            = f.submit('Update', class: 'btn btn-primary px-4 mt-3 mt-sm-0')
--- a/src/api/app/views/webui/users/tokens/index.html.haml
+++ b/src/api/app/views/webui/users/tokens/index.html.haml
@@ -22,8 +22,11 @@
                 %td= token.class.token_name.capitalize
                 %td.text-word-break-all= link_to("#{truncate(token.package.project.name, length: 80)}/#{truncate(token.package.name, length: 80)}",
                   package_show_path(project: token.package.project, package: token.package)) if token.package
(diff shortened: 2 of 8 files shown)


vpereira:
[CRUD - UI for tokens](https://trello.com/c/c1kcRa5n/1722-crud-ui-for-tokens)

src/api/app/views/webui/users/tokens/index.html.haml
```diff
@@ -23,7 +23,10 @@
                 %td.text-word-break-all= link_to("#{truncate(token.package.project.name, length: 80)}/#{truncate(token.package.name, length: 80)}",
                   package_show_path(project: token.package.project, package: token.package)) if token.package
                 %td.text-center
-                  = link_to '#', title: 'Delete Token', class: 'btn btn-sm btn-outline-danger',
+                  - if policy([:webui, token]).edit?
+                    = link_to(edit_token_path(token), title: 'Edit Token') do
+                      %i.fas.fa-edit
+                  = link_to '#', title: 'Delete Token', class: 'btn btn-sm',
```
saraycp, line 29: :right_anger_bubble:  we could align the icons. But it's not necessarily required for this PR.

Screenshot_2021-08-02 Open Build Service(2)

saraycp (changes requested):

src/api/app/views/webui/users/tokens/index.html.haml
dmarcoux: True, this is ugly! I'll look into this.

src/api/app/views/webui/users/tokens/edit.html.haml
```diff
@@ -0,0 +1,25 @@
...
+                = @token.id
+              .form-group
+                = f.label(:type, 'Operation:')
+                = @token.class.token_name.capitalize
+              .form-group
+                = f.label(:scm_token) do
+                  SCM token:
+                  %abbr.text-danger{ title: 'required' } *
+                .input-group
+                  = f.password_field(:scm_token, size: 80, class: 'form-control', placeholder: 'Please enter your new SCM token', required: true)
```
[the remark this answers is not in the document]
dmarcoux: True, I won't use a password field.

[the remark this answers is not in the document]
dmarcoux: With a password field, we cannot potentially leak the SCM token even when a workflow token already has a SCM token. It will simply display the placeholder in this case. With a normal input field, the SCM token is always displayed. I prefer to never show the SCM token since this is how SCMs handle their tokens.

I'll change the "Create Token" form to use a password field.

src/api/app/views/webui/users/tokens/index.html.haml
dmarcoux: Done, please have another look

src/api/app/views/webui/users/tokens/edit.html.haml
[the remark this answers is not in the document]
dmarcoux: > I'll change the "Create Token" form to use a password field.

Done, please have another look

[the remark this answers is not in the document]
saraycp: Fine! :+1:

src/api/app/views/webui/users/tokens/index.html.haml
saraycp: Better now.

saraycp (approved):

src/api/app/views/webui/users/tokens/index.html.haml
dmarcoux: It could still be improved, like having a smaller column, but I would say this is out of scope for this PR.

PR #11447, opened 2021-08-02
Merge `webui` namespaced Token policies with previous Token policies

eduardoj:
Reduce the technical debt introduced with #11399, by merging previous Token policies with `webui` namespaced Token policies.

Introduce a `trigger` policy, used when a token is triggered, instead of using a `create` policy in trigger controllers.

The selection of the user to check for token policies, relies now in the controller:
- for trigger controllers, the user attached to a token (see `pundit_user` method in `app/controllers/trigger_controller.rb`).
- for webui controllers, the user of the session.

-----------------


dmarcoux:
In case you haven't rebased on master already, you should do it to include the latest changes for the tokens UI. This newly introduced line needs to also be changed:

https://github.com/openSUSE/open-build-service/blob/aaacec02d7de2868720e4aa46054247747482434/src/api/app/views/webui/users/tokens/index.html.haml#L26

eduardoj:
> In case you haven't rebased on master already, you should do it to include the latest changes for the tokens UI. This newly introduced line needs to also be changed:
> 
> https://github.com/openSUSE/open-build-service/blob/aaacec02d7de2868720e4aa46054247747482434/src/api/app/views/webui/users/tokens/index.html.haml#L26

I rebased, and modified some other places, but I missed that one. Fixed. Thanks!

hennevogel (approved):

PR #11453, opened 2021-08-03
Measure finer grained namespaces

hennevogel:
If a build happens in a project that is maintained, record the full project name as namespace tag. Allows us to have finer grained measurements for important projects.

Additionally start measuring the new buildtype payload.

### DIFF
--- a/src/api/app/models/event/build.rb
+++ b/src/api/app/models/event/build.rb
@@ -19,11 +19,12 @@ module Event
 
     def metric_tags
       {
-        namespace: payload['project'].split(':').first,
+        namespace: ::Project.find_by_name(payload['project']).maintained_namespace,
         worker: payload['workerid'],
         arch: payload['arch'],
         reason: reason,
-        state: state
+        state: state,
+        buildtype: payload['buildtype']
       }
     end
 
--- a/src/api/app/models/concerns/project_maintenance.rb
+++ b/src/api/app/models/concerns/project_maintenance.rb
@@ -73,4 +73,15 @@ module ProjectMaintenance
            .joins(repositories: :release_targets)
            .where('release_targets.trigger = "maintenance"').includes(target_repositories: :project)
   end
+
+  def maintained_namespace
+    default = name.split(':').first
+
+    maintenance_project = Project.get_maintenance_project
+    return default unless maintenance_project
+
+    return maintenance_project.name if name.start_with?(maintenance_project.name)
+
+    maintenance_project.maintained_project_names.sort_by(&:length).reverse.find { |maintained_project_name| maintained_project_name.in?(name) } || default
+  end
 end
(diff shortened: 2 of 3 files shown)


hennevogel:
Do not merge, depends on #11456

hennevogel:
removed DNM label, #11456 is merged...

src/api/app/models/event/build.rb
```diff
@@ -54,5 +63,21 @@ def my_message_id
...
+      return maintenance_project.name if payload['project'].start_with?(maintenance_project.name)
+
+      # maintained projects and their sub-projects
+      return unless maintenance_project.maintained_project_names.any? { |name| payload['project'].start_with?(name) }
+
+      maintenance_project.maintained_project_names.sort_by(&:length).reverse_each do |name|
+        return name if name.in?(payload['project'])
+      end
+    end
+    # rubocop:enable Metrics/CyclomaticComplexity
```
danidoni, line 81:
```suggestion
    def maintained_project_name
      maintenance_project = Project.get_maintenance_project
      return unless maintenance_project

      return maintenance_project.name if payload['project'].start_with?(maintenance_project.name)

      # maintained projects and their sub-projects
      return unless maintenance_project.maintained_project_names.any? { |name| payload['project'].start_with?(name) }

      maintenance_project.maintained_project_names.sort_by(&:length).reverse.find { |name| name.in?(payload['project']) }
    end
```

You can use this to make it simpler

src/api/app/models/event/build.rb
```diff
@@ -54,5 +63,21 @@ def my_message_id
...
+
+    # rubocop:disable Metrics/CyclomaticComplexity
+    def maintained_project_name
+      maintenance_project = Project.get_maintenance_project
+      return unless maintenance_project
+
+      return maintenance_project.name if payload['project'].start_with?(maintenance_project.name)
+
+      # maintained projects and their sub-projects
+      return unless maintenance_project.maintained_project_names.any? { |name| payload['project'].start_with?(name) }
```
danidoni, line 75: This could be moved into `Project` also.

src/api/app/models/event/build.rb
hennevogel: hm, where is the difference exactly?

danidoni: Combining conditions and loops makes thing harder to understand

src/api/app/models/event/build.rb
hennevogel: true

src/api/app/models/event/build.rb
danidoni: So this: `maintenance_project.maintained_project_names.sort_by(&:length).reverse.find { |name| name.in?(payload['project']) }`

src/api/app/models/event/build.rb
hennevogel: it is now.

PR #11456, opened 2021-08-03
Introduces Project.get_maintenance_project!

hennevogel:
Sometimes we want to call this without an exception if there
is no project with that attribute.

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -258,12 +258,17 @@ class Project < ApplicationRecord
     end
 
     def get_maintenance_project(at = nil)
-      # hardcoded default. frontends can lookup themselfs a different target via attribute search
       at ||= AttribType.find_by_namespace_and_name!('OBS', 'MaintenanceProject')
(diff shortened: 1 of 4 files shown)


src/api/app/models/project.rb
```diff
@@ -258,12 +258,19 @@ def get_by_name(name, opts = {})
...
 
     def get_maintenance_project(at = nil)
-      # hardcoded default. frontends can lookup themselfs a different target via attribute search
       at ||= AttribType.find_by_namespace_and_name!('OBS', 'MaintenanceProject')
       maintenance_project = Project.find_by_attribute_type(at).first
-      unless maintenance_project && check_access?(maintenance_project)
-        raise Project::Errors::UnknownObjectError, 'There is no project flagged as maintenance project on server and no target in request defined.'
-      end
+
+      return nil unless maintenance_project
```
danidoni, line 264:
```suggestion
      return unless maintenance_project
```
:thought_balloon:

danidoni (approved):

PR #9559, opened 2020-05-15
Introduce beta feature tests for mobile

eduardoj:
We register a new driver called `:mobile`, where we emulate a mobile viewport. We also renamed the default driver to `:desktop`.

To run the tests for mobile, set the environment variable. For example:

```
CAPYBARA_DRIVER=mobile rspec spec/features/beta/webui/example_spec.rb
```

Without specifying the environment variable, tests should run within the desktop driver, as always.

In CircleCI we run the feature beta tests for desktop and mobile in the last container for the `rspec` CircleCI job.

PR #11468, opened 2021-08-05
Unify containers images used by CI and development environment 

eduardoj:
Taking advantage of the binary compatibility of both distributions, SLE 15 SP3 and openSUSE Leap 15.3, and after creating a new source of docker images based in SLE 15 SP3 (see https://build.opensuse.org/project/show/OBS:Server:Unstable:Containers ), we use those docker images for both, our CI setup and our development environment.

Upgrading the version of the distribution also upgraded the Chromium version. Two issues rised:
1. All feature tests failed: this was fixed by adding the `--disable-gpu` flag to the options of Chromium.
1. A false positive in feature package tests was fixed, making the tests for adding a user and adding a group fail. The bad
behavior was, that clicking the links 'Add User' or 'Add Group' was possible, even if the link was outside the visual area. ~Skipping~ Scrolling down to these links in these failing tests in mobile, make the tests pass again. See also [this comment](https://github.com/openSUSE/open-build-service/pull/9559#pullrequestreview-413770036), where it was considered, to scroll down until an element that should be clicked appeared into the visual area.

-----------------

### DIFF
--- a/src/api/spec/support/shared_examples/features/beta/user_tab.rb
+++ b/src/api/spec/support/shared_examples/features/beta/user_tab.rb
@@ -41,7 +41,9 @@ RSpec.shared_examples 'user tab' do
     end
 
     scenario 'Add non existent user' do
-      click_link('Add User')
+      add_user_link = page.find('a', text: 'Add User')
+      page.scroll_to(add_user_link)
+      add_user_link.click
       sleep 1 # FIXME: Needed to avoid a flickering test because the animation of the modal is sometimes faster than capybara
 
       within('#add-user-role-modal') do
@@ -53,7 +55,9 @@ RSpec.shared_examples 'user tab' do
     end
 
     scenario 'Add an existing user' do
-      click_link('Add User')
+      add_user_link = page.find('a', text: 'Add User')
+      page.scroll_to(add_user_link)
+      add_user_link.click
       sleep 1 # FIXME: Needed to avoid a flickering test because the animation of the modal is sometimes faster than capybara
 
       within('#add-user-role-modal') do
@@ -69,7 +73,9 @@ RSpec.shared_examples 'user tab' do
       end
 
       # Adding a user twice...
-      click_link('Add User')
+      add_user_link = page.find('a', text: 'Add User')
+      page.scroll_to(add_user_link)
+      add_user_link.click
       sleep 1 # FIXME: Needed to avoid a flickering test because the animation of the modal is sometimes faster than capybara
 
       within('#add-user-role-modal') do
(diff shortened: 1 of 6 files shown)


src/api/spec/support/shared_examples/features/beta/user_tab.rb
```diff
@@ -41,6 +41,8 @@
     end
 
     scenario 'Add non existent user' do
+      skip_on_mobile
```
danidoni, line 44: Can't we avoid skipping the test somehow?

eduardoj: As the description says, in [this comment](https://github.com/openSUSE/open-build-service/pull/9559#pullrequestreview-413770036) it was considered, to scroll down until an element that should be clicked appeared into the visual area.

I'm open to other alternatives, too.

hennevogel:
There is also [obs-review-apps](https://github.com/openSUSE/obs-review-apps/blob/master/files/docker-compose.yml)

eduardoj:
> There is also [obs-review-apps](https://github.com/openSUSE/obs-review-apps/blob/master/files/docker-compose.yml)

There you have it: https://github.com/openSUSE/obs-review-apps/pull/14

src/api/spec/support/shared_examples/features/beta/user_tab.rb
hennevogel: doesn't has to happen here and now right?

hennevogel (approved):

src/api/spec/support/shared_examples/features/beta/user_tab.rb
dmarcoux: I found a solution for this, scrolling to the element will work. See the commit I pushed.

dmarcoux (changes requested):
Let's not skip the specs for mobile. I found a fix for this, the latest commit I pushed to the PR.

eduardoj:
@dmarcoux:

> Let's not skip the specs for mobile. I found a fix for this, the latest commit I pushed to the PR.

Thanks! Applied. I modified the commit description accordingly. Please, have a look again.

src/api/spec/support/shared_examples/features/beta/user_tab.rb
```diff
@@ -148,6 +154,8 @@
     end
 
     scenario 'Add non existent group' do
+      skip_on_mobile
```
dmarcoux, line 157: We can do the same here as in the examples with `Add User`, just switching `Add User` to `Add Group` and it should work.

eduardoj: On it.

eduardoj: Done. I fixed the commit description again. Please take a look.

dmarcoux: This part of the commit message isn't true anymore: `Or skip the test in mobile.` It should be removed.

dmarcoux: Done!

dmarcoux (approved):

PR #11471, opened 2021-08-05
Consistently use User.get_default_admin

hennevogel:


src/api/spec/jobs/consistency_check_job_spec.rb
```diff
@@ -28,6 +28,7 @@
     let!(:project) { create(:project, name: 'super_project') }
     let(:consistency_checkjob) { described_class.new }
     let(:error_message) { "Project meta is different in backend for super_project\n{:foo=>\"bar\"}" }
+    let!(:admin_user) { create(:admin_user, login: 'Admin') }
```
krauselukas, line 31: @hennevogel why do you had to add this now? The logic of the code is stayed the same, just curious

src/api/spec/support/database_cleaner.rb
```diff
@@ -18,7 +18,7 @@
     # For feature test we use deletion instead of transactions because the
     # test suite and the capybara driver do not use the same server thread.
     DatabaseCleaner.strategy = if example.metadata[:type] == :feature || example.metadata[:type] == :migration || example.metadata[:thinking_sphinx] == true
-                                 # Omit truncating what we have set up in db/seeds.rb except users and roles_user
```
dmarcoux, line 21: This comment wasn't true or it changed? I don't see how this changed.

src/api/spec/jobs/consistency_check_job_spec.rb
hennevogel: We truncate the users table between test runs, including the Admin user.

src/api/spec/support/database_cleaner.rb
hennevogel: Yeah there are way more things we don't truncate. And we *do* truncate `users`.

dmarcoux (approved):

PR #11461, opened 2021-08-04
Automatically apply labels to PRs based on the changed files

dmarcoux:
This is done with the help of the GitHub action [pull request labeler](https://github.com/actions/labeler).

We sometimes forget to apply labels. Having labels on PRs is useful to categorize them and knowing what they changed.

PR #11472, opened 2021-08-05
Add GitHub workflow to apply labels to PRs

dmarcoux:
I totally forgot this part in #11461. That PR configured the pull request labeler, but it didn't run since there was no GitHub workflow... Oopsie!

### DIFF
--- a/.github/workflows/pull_request_labeler.yml
+++ b/.github/workflows/pull_request_labeler.yml
@@ -1,0 +1,11 @@
+name: "Pull Request Labeler"
+on:
+- pull_request_target
+
+jobs:
+  pull_request_labeler:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@v3
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"


.github/workflows/pull_request_labeler.yml
```diff
@@ -0,0 +1,11 @@
...
+on:
+- pull_request_target
+
+jobs:
+  pull_request_labeler:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@v3
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
```
hennevogel, line 11: Hm I guess this wont work as secrets are not shared between forks...

dmarcoux: I tested this and it worked, but I didn't test with PRs coming from a fork. You're right :disappointed:, see the official docs for [GITHUB_TOKEN](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token).

dmarcoux: Wait, this should work!

If this workflow ran on `pull_request` instead of `pull_request_target`, then it wouldn't work. With `pull_request_target` though, it works since it's executing this workflow in the context of the PR target, so `openSUSE/open-build-service`. See the following quote from this [GitHub blog post](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/):

_Emphasis is mine..._
> (...) we’ve added a new `pull_request_target` event, which behaves in an almost identical way to the `pull_request` event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the **workflow is (...) given access to a read/write token as well as secrets enabling the maintainer to (...) label a pull request.** (...)

hennevogel (approved):

PR #11474, opened 2021-08-05
Apply "Test Suite / CI" label to PRs changing our OBS workflows

dmarcoux:


dmarcoux:
Somehow the `pull request labeler` action is failing, but without an error message... it's not really useful. Could it be failing because we're changing its configuration for the label which would be applied on this PR?

dmarcoux:
> Somehow the `pull request labeler` action is failing, but without an error message... it's not really useful. Could it be failing because we're changing its configuration for the label which would be applied on this PR?

I've checked the [GitHub status page](https://www.githubstatus.com/) and they are currently having issues with GitHub actions.

hennevogel:
restarting helps...

hennevogel (approved):

PR #11482, opened 2021-08-09
Use bundler_ext instead of bundler if Gemfile.in file present

OlegGirko:
This gem disables Bundler's enforcement of exact gem versions in favour of basic constraint checks written in `Gemfile.in` file.

It's not included in `Gemfile`, so it should be installed explicitly by your system's package manager (as a dependency of OBS API package, for example) in order to allow system manager to take over responsibility of gem management from Bundler.

Also, it's packager's responsibility to place appropriate `Gemfile.in` file to enable this feature.

This is needed for installations where Ruby gems are managed by system package manager (like RPM or APT),
so bundler should not interfere with that and refuse to start application if gems with versions slightly different than ones mentioned in `Gemfile.lock` are installed in the system.

### DIFF
--- a/src/api/config/application.rb
+++ b/src/api/config/application.rb
@@ -10,8 +10,25 @@ require 'action_view/railtie'
 require 'sprockets/railtie'
 require 'rails/test_unit/railtie'
 
-# Assets should be precompiled for production (so we don't need the gems loaded then)
-Bundler.require(*Rails.groups(assets: ['development', 'test']))
+# The bundler_ext rubygem disables enforcement of gem versions in
+# `Gemfile.lock` in favour of the basic constraints defined in the file
+# `Gemfile.in`. The bundler_ext rubygem is not part of the OBS bundle, so you
+# need to explicitly install it, however you install gems on your system. Eg.
+# with gem or your system package manager. Then create the appropriate file
+# with `cp Gemfile Gemfile.in`. For more information see the bundler_ext
+# documentation.
+
+# WARNING: You will be on your own with problems if you use bundler_ext, we
+# only ensure our app works with the exact gems specified in our Gemfile.lock
+gemfile_in = File.expand_path('../Gemfile.in', __dir__)
+if File.exist?(gemfile_in)
+  require 'bundler_ext'
+  BundlerExt.system_require(gemfile_in, *Rails.groups(assets: ['development', 'test']))
+else
+  # Assets should be precompiled for production (so we don't need the gems loaded then)
+  Bundler.require(*Rails.groups(assets: ['development', 'test']))
+end
+
 require_relative '../lib/rabbitmq_bus'
 
 module OBSApi
--- a/src/api/config/boot.rb
+++ b/src/api/config/boot.rb
@@ -1,3 +1,5 @@
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
+unless File.exist?(File.expand_path('../Gemfile.in', __dir__))
+  ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
 
-require 'bundler/setup' # Set up gems listed in the Gemfile.
+  require 'bundler/setup' # Set up gems listed in the Gemfile.
+end


src/api/config/boot.rb
```diff
@@ -1,3 +1,5 @@
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
+if !File.exist?(File.expand_path('../Gemfile.in', __dir__))
```
hennevogel, line 1:
```suggestion
unless File.exist?(File.expand_path('../Gemfile.in', __dir__))
```

src/api/config/application.rb
```diff
@@ -10,8 +10,14 @@
 require 'sprockets/railtie'
 require 'rails/test_unit/railtie'
 
-# Assets should be precompiled for production (so we don't need the gems loaded then)
-Bundler.require(*Rails.groups(assets: ['development', 'test']))
+gemfile_in = File.expand_path('../Gemfile.in', __dir__)
+if File.exist?(gemfile_in)
+  require 'bundler_ext'
```
hennevogel, line 15:
```suggestion
  # Make sure you have this gem installed, it's not part of our bundle.
  require 'bundler_ext'
```

hennevogel (changes requested):

hennevogel:
I think I would also fancy a more detailed description in the commit message. Something like

```
To set this up install the gem (`gem install bundler_ext`) and create a copy of the Gemfile (`cp Gemfile Gemfile.in`).
```

src/api/config/boot.rb
OlegGirko: "Unless" if more difficult to parse then "if not" for non-English speakers.
Doing this anyway, just because your CI doesn't requires it.

src/api/config/application.rb
OlegGirko: Adding even more detailed comment.

OlegGirko:
> I think I would also fancy a more detailed description in the commit message. Something like
> 
> ```
> To set this up install the rubygem (`gem install bundler_ext`) and create a copy of the Gemfile (`cp Gemfile Gemfile.in`).
> ```

Added more details to commit message.

OlegGirko:
Also, fixed trailing whitespace in the comment with details about `bundler_ext`.

src/api/config/application.rb
```diff
@@ -10,8 +10,23 @@
...
+if File.exist?(gemfile_in)
+  # This gem disables Bundler's enforcement of exact gem versions
+  # in favour of basic constraint checks written in "Gemfile.in" file.
+  # It's not included in "Gemfile", so it should be installed explicitly
+  # by your system's package manager
+  # (as a dependency of OBS API package, for example)
+  # in order to allow system manager to take over responsibility of
+  # gem management from Bundler.
+  # Also, it's packager's responsibility to place appropriate
+  # "Gemfile.in" file to enable this feature.
```
hennevogel, line 23:
```suggestion
  # The bundler_ext rubygem disables bundler's enforcement of exact gem versions in `Gemfile.lock`
  # in favour of the basic constraints defined in the file `Gemfile.in`.
  # The bundler_ext rubygem is not part OBS, so you need to explicitly install it, however you
  # install gems on your system. Eg. with gem or your system package manager.
  # Then create the appropriate file with `cp Gemfile Gemfile.in`
  # WARNING: You will be on your own with problems if you do this, we only ensure our app works with the gems
  #                     specified in Gemfile.lock 
```

hennevogel (changes requested):

src/api/config/application.rb
OlegGirko: Actually, "create the appropriate file with `cp Gemfile Gemfile.in`" is not exactly correct: `Gemfile.in` has to be patched after that. You can see how to do it in my `.spec` file.

hennevogel: You mean the name changes? From activerecord to active_record?

OlegGirko: Yes. This is what I do in `.spec` file:
```
sed -i \
    -e "s/'activerecord'/'active_record'/" \
    -e "s/'actionmailer'/'action_mailer'/" \
    -e "s/'yajl-ruby'/'yajl'/" \
    -e "s/'sprite-factory'/'sprite_factory'/" \
    -e "s/'minitest-fail-fast'/'minitest-fail_fast'/" \
    %{buildroot}%{_datadir}/obs-api/Gemfile.in
```

hennevogel: pushed in a slightly different way.

hennevogel (approved):

PR #11496, opened 2021-08-13
Remove deleted project and package records from sphinx indexes

krauselukas:
At the moment we only populate new records to sphinx. But we
don't remove those indexes after those records get deleted from
the database.
Usually the Thinking Sphinx gem is taking care of this automatically,
but since we differ from the regular setup by writing to the sphinx
index in a delayed job, we also have to consider the deletion.

There are still other issues, when updating/deleting associated attributes on the package/project. But this needs still more investigation, and will be part of another PR.

### DIFF
--- a/src/api/app/jobs/delete_from_sphinx_job.rb
+++ b/src/api/app/jobs/delete_from_sphinx_job.rb
@@ -1,0 +1,21 @@
+class DeleteFromSphinxJob < ApplicationJob
+  queue_as :quick
+
+  def perform(id, klass)
+    delete_from_sphinx(id: id, klass: klass)
+  end
+
+  private
+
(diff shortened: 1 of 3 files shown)


krauselukas:
This is basically extracting the delete callback logic from thinking sphinx to the delayed job https://github.com/pat/thinking-sphinx/blob/922e7613fb04d520a587bf7c6de26bcdc98b4019/lib/thinking_sphinx/active_record/callbacks/delete_callbacks.rb#L18

hennevogel:
checkout code obviously ran...

Screenshot from 2021-08-16 11-05-26

https://app.circleci.com/pipelines/github/krauselukas/open-build-service/162/workflows/bd835216-a509-489a-be64-3bc8073aef66

PR #11511, opened 2021-08-18
Add selinux :z and :Z labels to the docker compose volumes

dcermak:
The current docker compose configs do not work on systems with selinux enabled, as the labels in the container will be wrong and all kinds of things will fail. Thus we must add :z and :Z labels to all volumes.


docker-compose.ha.yml
```diff
@@ -3,7 +3,7 @@ services:
   proxy:
     image: haproxy:1.7
     volumes:
-      - ./contrib/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro
+      - ./contrib/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:Z
     ports:
```
coolo, line 7: According to https://docs.docker.com/storage/bind-mounts/#differences-between--v-and---mount-behavior you can have both `ro` and `Z` in a comma seperated list, not sure why only some files are `ro` though

dcermak: Oh, I was not aware of the option to provide a comma separated list. Let me give that a try.

dcermak:
I have given this a spin locally and it appears to work just fine on a system with selinux in enforcing mode.

hennevogel (approved):

PR #11516, opened 2021-08-19
Fix measuring request duration

hennevogel:
The `state` getter if overloaded and transforms the string to a symbol
(why???), we cover for that now.

Also simplifying the method a bit.

### DIFF
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -15,7 +15,7 @@ class BsRequest < ApplicationRecord
     'bs_request_actions.type'
   ].freeze
 
-  FINAL_REQUEST_STATES = ['accepted', 'declined', 'superseded', 'revoked'].freeze
+  FINAL_REQUEST_STATES = [:accepted, :declined, :superseded, :revoked].freeze
 
   VALID_REQUEST_STATES = [:new, :deleted, :declined, :accepted, :review, :revoked, :superseded].freeze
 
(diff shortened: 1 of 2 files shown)


src/api/app/models/bs_request.rb
```diff
@@ -813,13 +813,14 @@ def setincident(incident)
...
+    return unless state_changed?
     # new->review && review->new are not worth an event - it's just spam
-    return if state.to_s.in?(intermediate_state) && state_was.to_s.in?(intermediate_state)
+    return unless conclusive?
 
     options = event_parameters
-    options[:duration] = (updated_at - created_at).to_i if conclusive?
+
+    # measure duration unless superseding a final state, like revoked -> superseded
+    options[:duration] = (updated_at - created_at).to_i if FINAL_REQUEST_STATES.exclude?(state_was.to_sym) && FINAL_REQUEST_STATES.exclude?(state)
```
krauselukas, line 823: @hennevogel the previous behaviour was `FINAL_REQUEST_STATES.exclude?(state_was.to_sym) && FINAL_REQUEST_STATES.include?(state)` is this change intended

krauselukas: so a state change from `:new -> :accepted` would not record the duration

hennevogel: Of course not. copy pasta :) Should be `exclude(was) && include(is)`

- `FINAL_REQUEST_STATES.exclude?(:new) => true`
- `FINAL_REQUEST_STATES.include?(:accepted) => true`

or 

- `FINAL_REQUEST_STATES.exclude?(:accepted) => false`
- `FINAL_REQUEST_STATES.include?(:superseded) => true`

Good catch!

krauselukas (approved):

PR #11527, opened 2021-08-24
Recover from errors when commenting on a BsRequest

### LINKED ISSUE #10630: WebUI error when submitting a comment to a SR
# Issue Description
After filling out the comment text field on a SR I get the error message "Failed to create comment" and the homepage rendered in the "comments-element".

Expected Result
============
My comment is posted successfully as it is previewed.

How to Reproduce
===============
1. Go to https://build.opensuse.org/request/show/855492
2. Fill out comment text field (not "my decision")
3. Click on "Add comment"
4. See error (screenshot below)

Further Information
===============
* Happened this morning
* I participate in the public beta

Screenshot from 2021-01-14 11-44-07

danidoni:
Recover from errors when commenting on a request. 

We were recovering from errors by redirecting the flow to the main page, as we do with non-AJAX flows. Now we just show the error details on the comment box, if there are any.

Here you can see how the commenting box behaves when there is an error:

Fixes #10630

### DIFF
--- a/src/api/app/controllers/webui/comments_controller.rb
+++ b/src/api/app/controllers/webui/comments_controller.rb
@@ -1,8 +1,13 @@
 class Webui::CommentsController < Webui::WebuiController
   before_action :require_login
-  before_action :find_commentable, only: :create
+  before_action :set_commented, only: :create
 
   def create
+    if @commented.nil?
+      flash.now[:error] = "Failed to create comment: This #{@commentable_type.name.downcase} does not exist anymore."
+      render partial: 'layouts/webui/flash' and return
+    end
+
     comment = @commented.comments.new(permitted_params)
     User.session!.comments << comment
     @commentable = comment.commentable
@@ -106,13 +111,12 @@ class Webui::CommentsController < Webui::WebuiController
     params.require(:comment).permit(:body, :parent_id)
   end
 
-  def find_commentable
-    commentable = [Project, Package, BsRequest].find { |klass| klass.name == params[:commentable_type] }
+  def set_commented
+    @commentable_type = [Project, Package, BsRequest].find { |klass| klass.name == params[:commentable_type] }
+    @commented = @commentable_type&.find_by(id: params[:commentable_id])
+    return if @commentable_type.present?
 
-    if commentable.nil?
-      redirect_to(root_path, error: 'Failed to create comment')
-    else
-      @commented = commentable.find(params[:commentable_id])
-    end
+    flash[:error] = "Invalid commentable #{params[:commentable_type]} supplied."
+    render partial: 'layouts/webui/flash'
   end
 end
--- a/src/api/spec/controllers/webui/comments_controller_spec.rb
+++ b/src/api/spec/controllers/webui/comments_controller_spec.rb
@@ -13,23 +13,24 @@ RSpec.describe Webui::CommentsController, type: :controller do
     let(:bs_request) { create(:set_bugowner_request) }
 
     context 'with invalid commentable_type' do
+      subject! { post :create, params: comment_params }
+
       let(:comment_params) do
         { comment: { body: 'This is AWESOME!' }, commentable_type: 'FOOBAR', commentable_id: 31_337 }
       end
 
-      subject { post :create, params: comment_params }
-
-      it { expect(subject.request.flash[:success]).to be_nil }
-      it { expect(subject.request.flash[:error]).not_to(be_nil) }
-      it { expect(subject).to redirect_to(root_path) }
+      it 'fails to create a comment' do
+        expect(flash[:error]).to eq('Invalid commentable FOOBAR supplied.')
+      end
     end
 
     context 'with a valid comment' do
+      subject! { post :create, params: comment_params }
+
       RSpec.shared_examples 'saving a comment' do
-        before do
-          params = { comment: { body: "This #{commentable.model_name.singular} is AWESOME!" },
-                     commentable_type: commentable.class, commentable_id: commentable.id }
-          post :create, params: params
+        let(:comment_params) do
+          { comment: { body: "This #{commentable.model_name.singular} is AWESOME!" },
+            commentable_type: commentable.class, commentable_id: commentable.id }
         end
 
         it { expect(flash[:success]).to eq('Comment created successfully.') }
@@ -55,20 +56,61 @@ RSpec.describe Webui::CommentsController, type: :controller do
       end
     end
 
-    context 'saving a comment without body' do
-      before do
-        params = { comment: { body: '' }, commentable_type: package.class, commentable_id: package.id }
-        post :create, params: params
+    context 'with a commentable that does not exist' do
+      subject! { post :create, params: comment_params }
+
+      context 'of a project' do
+        let(:commentable) { project }
+        let(:comment_params) do
+          { comment: { body: 'This project is AWESOME!' },
+            commentable_type: commentable.class, commentable_id: -commentable.id }
+        end
+
+        it { expect(flash[:error]).to eq('Failed to create comment: This project does not exist anymore.') }
+        it { expect(package.comments.count).to eq(0) }
       end
+
+      context 'of a package' do
+        let(:commentable) { package }
+        let(:comment_params) do
+          { comment: { body: 'This package is AWESOME!' },
+            commentable_type: commentable.class, commentable_id: -commentable.id }
+        end
+
+        it { expect(flash[:error]).to eq('Failed to create comment: This package does not exist anymore.') }
+        it { expect(package.comments.count).to eq(0) }
+      end
+
+      context 'of a bs_request' do
+        let(:commentable) { bs_request }
+        let(:comment_params) do
+          { comment: { body: 'This bs_request is AWESOME!' },
+            commentable_type: commentable.class, commentable_id: -commentable.id }
+        end
+
+        it { expect(flash[:error]).to eq('Failed to create comment: This bsrequest does not exist anymore.') }
+        it { expect(package.comments.count).to eq(0) }
+      end
+    end
+
+    context 'saving a comment without body' do
+      let(:comment_params) do
+        { comment: { body: '' }, commentable_type: package.class, commentable_id: package.id }
+      end
+
+      subject! { post :create, params: comment_params }
 
       it { expect(flash[:error]).to eq("Failed to create comment: Body can't be blank.") }
       it { expect(package.comments.count).to eq(0) }
     end
 
     context "does not allow to overwrite the comment's user" do
+      let(:comment_params) { { comment: { body: 'This project is AWESOME!', user_id: user }, commentable_type: project.class, commentable_id: project.id } }
+
+      subject { post :create, params: comment_params }
+
       it 'raises an error' do
-        params = { comment: { body: 'This project is AWESOME!', user_id: user }, commentable_type: project.class, commentable_id: project.id }
-        expect { post :create, params: params }.to raise_error(ActionController::UnpermittedParameters)
+        expect { subject }.to raise_error(ActionController::UnpermittedParameters)
       end
     end
   end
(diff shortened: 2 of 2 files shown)


src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -3,18 +3,23 @@ class Webui::CommentsController < Webui::WebuiController
   before_action :find_commentable, only: :create
 
   def create
-    comment = @commented.comments.new(permitted_params)
-    User.session!.comments << comment
-    @commentable = comment.commentable
+    if @commented.nil?
+      flash.now[:error] = 'Failed to create comment: The object the comment refers to does not exist anymore.'
```
hennevogel, line 7: `object` is not something your customers will understand.

danidoni: Yeah, I'm still working on that messages. I don't like it either

dmarcoux: As discussed with @danidoni, we can tell for which kind of object the comment was for. This is passed in the params (`params[:commentable_type]`) when the form is submitted.

src/api/spec/controllers/webui/comments_controller_spec.rb
```diff
@@ -55,6 +53,47 @@
       end
     end
 
+    context 'with a commentable that does not exist' do
+      context 'of a project' do
+        let(:commentable) { project }
+
+        before do
+          params = { comment: { body: "This #{commentable.model_name.singular} is AWESOME!" },
```
dmarcoux, line 61: I don't think `singular` is needed here. I suggest simply typing `project` instead of having some code which is pretty much like the implementation.

```suggestion
          params = { comment: { body: "This #{commentable.model_name} is AWESOME!" },
```

src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -1,20 +1,26 @@
...
+  before_action :set_commentable_type, only: :create
   before_action :find_commentable, only: :create
 
   def create
-    comment = @commented.comments.new(permitted_params)
-    User.session!.comments << comment
-    @commentable = comment.commentable
+    if @commented.nil?
+      flash.now[:error] = "Failed to create comment: This #{@commentable_type.name.downcase} does not exist anymore."
+      render partial: 'layouts/webui/flash'
```
dmarcoux, line 9: This action is called from JS, no? So I don't think it's needed to render this.

src/api/app/views/webui/shared/_error.html.haml
```diff
@@ -0,0 +1 @@
+= flash[:error]
```
dmarcoux, line 1: Why is this shared partial needed? We already have [app/views/layouts/webui/_flash.html.haml](https://github.com/openSUSE/open-build-service/blob/775d283116ba0eec1e3f0647f42db75bc881a824/src/api/app/views/layouts/webui/_flash.html.haml) which is handling flash messages.

src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -67,13 +73,13 @@ def permitted_params
...
-    if commentable.nil?
-      redirect_to(root_path, error: 'Failed to create comment')
-    else
-      @commented = commentable.find(params[:commentable_id])
-    end
+    raise ArgumentError, "Invalid commentable #{params[:commentable_type]} supplied." unless @commentable_type
+  end
+
+  def find_commentable
+    @commented = @commentable_type.find_by(id: params[:commentable_id]) unless @commentable_type.nil?
```
dmarcoux, line 83: This method is called `find_commentable`, but it sets the instance variable `@commented`. It's a bit confusing. We could also merge `set_commentable_type` and `find_commentable`. Having two callbacks with such a minimal amount of code isn't needed. One would be enough.

src/api/spec/controllers/webui/comments_controller_spec.rb
danidoni: True, I unpacked a shared_context and forgot about that line...

src/api/app/controllers/webui/comments_controller.rb
danidoni: Unfortunately, it is needed, otherwise the request fails with no feedback. This way at least the user knows that the operation did not succeed.

src/api/app/views/webui/shared/_error.html.haml
danidoni: True, removing

src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -67,13 +73,13 @@ def permitted_params
...
-    commentable = [Project, Package, BsRequest].find { |klass| klass.name == params[:commentable_type] }
+  def set_commentable_type
+    @commentable_type = [Project, Package, BsRequest].find { |klass| klass.name == params[:commentable_type] }
 
-    if commentable.nil?
-      redirect_to(root_path, error: 'Failed to create comment')
-    else
-      @commented = commentable.find(params[:commentable_id])
-    end
+    raise ArgumentError, "Invalid commentable #{params[:commentable_type]} supplied." unless @commentable_type
```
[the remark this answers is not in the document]
danidoni: Okay, I'm fine with using the error flash, but I want to stop the flow in that `before_action`, let me try redirecting from the `before_action` filter as it stops the flow.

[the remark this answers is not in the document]
danidoni: Okay, so filling the `flash[:error]` and rendering a partial is enough in the `before_filter` to stop the flow in this case.

src/api/app/controllers/webui/comments_controller.rb
danidoni: Right, done.

src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -67,13 +72,12 @@ def permitted_params
...
+  def set_commented
+    @commentable_type = [Project, Package, BsRequest].find { |klass| klass.name == params[:commentable_type] }
+    return if @commentable_type
 
-    if commentable.nil?
-      redirect_to(root_path, error: 'Failed to create comment')
-    else
-      @commented = commentable.find(params[:commentable_id])
-    end
+    @commented = @commentable_type.find_by(id: params[:commentable_id])
```
dmarcoux, line 79: Any reason for not doing  `@commentable_type.find(params[:commentable_id])`?

src/api/spec/controllers/webui/comments_controller_spec.rb
```diff
@@ -19,9 +19,7 @@
 
       subject { post :create, params: comment_params }
 
-      it { expect(subject.request.flash[:success]).to be_nil }
-      it { expect(subject.request.flash[:error]).not_to(be_nil) }
-      it { expect(subject).to redirect_to(root_path) }
+      it { expect { subject }.to raise_error(ArgumentError) }
```
dmarcoux, line 22: This spec should be updated, we're back to rendering a flash.

src/api/spec/controllers/webui/comments_controller_spec.rb
```diff
@@ -139,9 +178,7 @@
...
```
dmarcoux, line 181: This spec should be updated, we're back to rendering a flash.

src/api/app/controllers/webui/comments_controller.rb
danidoni: No, :thinking: I'll change it

danidoni: Wait, that code is wrong... I need to rewrite it...

src/api/app/views/webui/shared/_error.html.haml
eduardoj: This file was introduced here and removed in the third commit "Remove useless partial". That commit should be sqashed with the first commit.

src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -1,29 +1,35 @@
...
-             end
-
-    if Flipper.enabled?(:request_show_redesign, User.session) && comment.commentable_type == 'BsRequest'
-      render(partial: 'webui/comment/beta/comments_thread',
-             locals: { comment: comment.root, commentable: @commentable, level: 1 },
-             status: status)
+    if @commented.nil?
+      flash.now[:error] = "Failed to create comment: This #{@commentable_type.name.downcase} does not exist anymore."
+      render partial: 'layouts/webui/flash'
     else
```
eduardoj, line 9: Introducing this `if - else` statement makes codeclimate warn with "Method create has a Cognitive Complexity of 9 (exceeds 8 allowed).". Consider using [`and return`](https://guides.rubyonrails.org/layouts_and_rendering.html#avoiding-double-render-errors) after `render partial`, so the `else` clause can be removed.

danidoni: Done

src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -1,8 +1,13 @@
 class Webui::CommentsController < Webui::WebuiController
   before_action :require_login
-  before_action :find_commentable, only: :create
+  before_action :set_commented, only: :create
 
   def create
+    if @commented.nil?
+      flash.now[:error] = "Failed to create comment: This #{@commentable_type.name.downcase} does not exist anymore."
+      return render partial: 'layouts/webui/flash'
```
eduardoj, line 8: I prefer it this way:
```suggestion
      render partial: 'layouts/webui/flash' and return
```
This solution is described in the Rails guides, we don't need the method `create` to return what `render` returns, and I don't see the benefit to shorten the code for readibility. See also [this explanation](https://codereview.stackexchange.com/questions/996/return-render-in-ruby-on-rails/997#997). What do you think?

danidoni: Oh sorry, I completely overlooked the link. I'll change it now

danidoni: You are completely right, I didn't like it very much. Indeed it seemed to return something when in fact it didn't

src/api/spec/controllers/webui/comments_controller_spec.rb
```diff
@@ -55,6 +54,47 @@
...
     end
 
+    context 'with a commentable that does not exist' do
+      context 'of a project' do
+        let(:commentable) { project }
+
+        before do
+          params = { comment: { body: 'This project is AWESOME!' },
+                     commentable_type: commentable.class, commentable_id: -commentable.id }
+          post :create, params: params
```
eduardoj, line 64: We are testing the action `POST #create` in all these tests. I would have preferred to replace all the lines:
```
post :create, params: params
```
with
```
subject { post :create, params: comment_params }
```
, as it was used previously.

Later I would have moved out all the `subject {...}` calls to only one line, after the `describe 'POST #create' do` block definition. And then adapt the tests, to use `subject`.

The reasoning is that in the `describe 'POST #create' do` block all the tests share the same subject. They always make a call to `post :create`.

This could be done later. I will not stop this PR from being merged only for this reason. What do you think?

danidoni: I added a comment with a proposal of spec refactoring. Let me know what do you think

eduardoj: It seems that yes, the first step of the three has been addressed. I modified my previous comment to point out what are the three steps I'm talking about.

danidoni: Well, I could not do it **exactly** as you propose, because the spec uses two different subjects: the immediate one `subject!` and a delayed one `subject`.

That last one is needed for the use case where we're expecting an error raise (see [here](https://github.com/openSUSE/open-build-service/pull/11527/files#diff-bee2469df30c00bd819919cc1cedb771cd5d830c614b8e1e456adb7717475bacR113), I need the subject to trigger inside the `expect { ... }` block to catch the error, otherwise, the spec fails).

So I cannot move them all to just one `subject` block...

Do you agree?

eduardoj: True. The test of checking for errors is the only one that requires `subject` (without `!`). I'm fine with the given solution, then.

eduardoj (approved):

PR #11536, opened 2021-08-26
[api] support request history search by time

adrianschroeter:
Support xpath search for

 review/@when
 review/history/@when
 history/@when

attributes (OBS-149)

-----------------

### DIFF
--- a/src/api/lib/xpath_engine.rb
+++ b/src/api/lib/xpath_engine.rb
@@ -1,4 +1,5 @@
 # rubocop:disable Layout/LineLength
+# rubocop:disable Metrics/MethodLength
 class XpathEngine
   require 'rexml/parsers/xpathparser'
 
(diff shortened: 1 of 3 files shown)


src/api/lib/xpath_engine.rb
```diff
@@ -220,7 +220,10 @@ def initialize
         'review/@by_group' => { cpart: 'r.by_group', joins: 'LEFT JOIN reviews r ON r.bs_request_id = bs_requests.id' },
         'review/@by_project' => { cpart: 'r.by_project', joins: 'LEFT JOIN reviews r ON r.bs_request_id = bs_requests.id' },
         'review/@by_package' => { cpart: 'r.by_package', joins: 'LEFT JOIN reviews r ON r.bs_request_id = bs_requests.id' },
+	'review/@when' => { cpart: 'bs_requests.updated_at', joins: 'LEFT JOIN reviews r ON r.bs_request_id = bs_requests.id' },
```
dmarcoux, line 223: The indentation for the new lines should be the same as the other lines.

eduardoj (approved):

PR #11550, opened 2021-08-30
Update bundler version to a recent one

hennevogel:
CI will never pass as the version of bundler is hardcoded in the CI containers. Changing it would make this PRs CI pass and everyone elses fail. 

You need to trust me and let me commit without CI :)


dmarcoux (approved):

hennevogel:
here goes nothing...

PR #11555, opened 2021-08-31
Create view component for download repository link

dmarcoux:
This replaces the presenter Repositories::StatePresenter. Presenters and view components are really similar, but the template code (HAML) is testable in view components and not in presenters. The view component is also tightly coupled to its template code unlike the presenter which is using a shared partial. In this case, I decided against testing the view component since it's pretty simple and it has almost no logic. Still, I believe it's beneficial to have a view component since we reuse it.

Having design patterns which are too similar is confusing. This is why retiring presenters in favor of view components is beneficial. This is the first PR where we tackle this.

The 2 pages were this view component is used:
_Binaries of a package_
example

_State of a repository for a project_
example-2


danidoni (approved):

dmarcoux:
I realized that I forgot to add a preview for the view component. I added it and it's available at http://obs-reviewlab.opensuse.org/dmarcoux-download_repository_link_component/rails/view_components/download_repository_link_component/preview

krauselukas (approved):
Great

krauselukas:
coverage/coveralls seems stuck, will merge

PR #11567, opened 2021-09-01
[api] avoid delivering notifications to unconfirmed users

adrianschroeter:
eg. do not send out mails anymore when a user got locked

### DIFF
--- a/src/api/app/models/event_subscription/find_for_event.rb
+++ b/src/api/app/models/event_subscription/find_for_event.rb
@@ -54,7 +54,7 @@ class EventSubscription
       receivers.each do |receiver|
         case receiver
         when User
-          new_receivers << receiver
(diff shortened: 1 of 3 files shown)


src/api/app/models/event_subscription/find_for_event.rb
```diff
@@ -54,7 +54,7 @@ def filter_and_convert_groups_without_emails_to_users(receivers)
       receivers.each do |receiver|
         case receiver
         when User
-          new_receivers << receiver
+          new_receivers << receiver if receiver.confirmed?
```
vpereira, line 57: @adrianschroeter we don't have a method `confirmed?`. Instead we have a `User#is_active?`

adrianschroeter: ups, fixed

PR #11569, opened 2021-09-02
Use workflow configuration file from the base branch for GitHub

eduardoj:
Until this change, we always use the workflow configuration file from the default branch of a repository.

Add a missing "ref" attribute to the Octokit call, with the branch that must be used, to be able to retrieve the correct workflow configuration file.


dmarcoux (approved):
The commit description is a bit too verbose. Anyway, :+1:

PR #11570, opened 2021-09-02
Create ScmWebhook model and reorganize validations for webhook events

dmarcoux:
`ScmWebhook` contains the payload extracted from a SCM webhook and provides helper methods to know which webhook event we're dealing with. Where we previously had a payload extracted from a SCM webhook, we now pass around an instance of `ScmWebhook` which still has the payload.

Before, validation of webhook events was happening in two places: `WorkflowEventAndActionValidator` and `TriggerControllerService::ScmExtractor`. This was confusing since we sometimes need to check what kind of webhook event we're dealing with. This forced us to do questionable things like using an instance of `WorkflowEventAndActionValidator` to check, for example, if the webhook is for an updated GitHub pull request. Now, there's a single source for such questions, it's the `ScmWebhook` model. The webhook events are also validated in a single class, `ScmWebhookEventValidator`.


dmarcoux:
[Support closing/merging/reopening actions for PRs/MRs](https://trello.com/c/nlIsH1Wp/1737-support-closing-merging-reopening-actions-for-prs-mrs)

eduardoj (approved):

src/api/app/models/scm_webhook.rb
```diff
@@ -0,0 +1,34 @@
+# Contains the payload extracted from a SCM webhook and provides helper methods to know which webhook event we're dealing with
+class ScmWebhook
```
vpereira, line 2: couldn't we call it `SCMWebhook` ?

src/api/app/validators/scm_webhook_event_validator.rb
```diff
@@ -0,0 +1,46 @@
+class ScmWebhookEventValidator < ActiveModel::Validator
```
vpereira, line 1: `SCMWebhookEventValidator`

src/api/app/models/scm_webhook.rb
dmarcoux: I decided to go with _Scm_ in lowercase to be consistent with other classes also using this case, like `ReportToScmJob`, `TriggerControllerService::ScmExtractor`, `BadScmHeaders` and `BadScmPayload`.

src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
```diff
@@ -9,21 +9,21 @@
   describe '#call' do
     context 'with webhook payload from gitlab' do
       it 'initializes a workflow object' do
-        service = Workflows::YAMLToWorkflowsService.new(yaml_file: workflows_yml_file, scm_extractor_payload: gitlab_extractor_payload, token: token)
+        service = Workflows::YAMLToWorkflowsService.new(yaml_file: workflows_yml_file, scm_webhook: ScmWebhook.new(payload: gitlab_extractor_payload), token: token)
```
vpereira, line 12: couldn't it be a `let` too?

src/api/app/validators/scm_webhook_event_validator.rb
dmarcoux: Same comment as the other regarding the case of _SCM_. I won't change this.

src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
dmarcoux: I'll use a subject.

src/api/spec/services/trigger_controller_service/scm_extractor_spec.rb
```diff
@@ -95,8 +53,10 @@
         }
       end
 
-      it 'returns a hash with the extracted data from the github payload' do
-        expect(subject.call).to include(expected_hash)
+      it 'returns an instance of ScmWebhook with the extracted data from the GitHub payload' do
+        scm_webhook = subject.call
```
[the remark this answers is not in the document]
dmarcoux: I just double-checked, because I wasn't sure anymore, [`subject` is memoized within an example](https://relishapp.com/rspec/rspec-core/v/3-10/docs/subject/explicit-subject#the-%60subject%60-is-memoized-within-an-example-but-not-across-examples), so I can simply use the subject multiple times in an example.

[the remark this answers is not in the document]
dmarcoux: Please have another look.

src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
dmarcoux: Please have another look.

rubhanazeem (approved):

PR #11573, opened 2021-09-02
Fail if the Workflow token does not have a payload

### LINKED ISSUE #11535: [production][trigger#create] undefined method `with_indifferent_access' for nil:NilClass
trigger#create threw this 1 times
<pre>
undefined method `with_indifferent_access&#39; for nil:NilClass
</pre>
[https://api.opensuse.org/trigger/runservice](https://api.opensuse.org/trigger/runservice)"

[See this exception on Errbit](http://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/6127564d84bf4607ebfd7c00)

danidoni:
It's possible to trigger a workflow token via the API. But doing so makes no sense as we cannot provide a payload (the token gets the required data from the payload. If there is no payload, we can't get the required data).

This PR ensures the API gets a proper error when trying to trigger a Workflow Token without a payload.

Fixes #11535

### DIFF
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -29,6 +29,8 @@ class TriggerController < ApplicationController
       @token.call(opts)
       render_ok
     end
+  rescue ArgumentError => e
+    render_error status: 400, message: e
   end
 
   private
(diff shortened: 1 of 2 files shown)


danidoni:
[Fix error in trigger_controller.rb](https://trello.com/c/xInqte2G/1741-fix-error-in-triggercontrollerrb)

rubhanazeem (approved):
LGTM!

eduardoj (approved):

PR #11580, opened 2021-09-03
Generalize workflow step validations

danidoni:
- [x] Move the validations to the Workflow::Step
- [ ] Write tests for the Workflow::Step covering the generic validation
- [ ] Implement the validations for the rest of the Workflow::Step inherited classes


danidoni:
[Move Workflow::Step::ConfigureRepositories#validate_step_instructions to Workflow::Step](https://trello.com/c/WO0i30Nj/1740-move-workflowstepconfigurerepositoriesvalidatestepinstructions-to-workflowstep)

src/api/app/models/workflow/step/configure_repositories.rb
```diff
@@ -1,8 +1,6 @@
 class Workflow::Step::ConfigureRepositories < Workflow::Step
-  REQUIRED_KEYS = [:source_project, :repositories].freeze
```
dmarcoux, line 2: I don't see why we can't use a constant still. Every step needs to have this constant and if it doesn't, the validation is going to break. Nothing will work without this, so it's not possible to forget this constant.

dmarcoux (approved):

PR #11581, opened 2021-09-03
Fix IssueTracker validation errors

hennevogel:
IssueTracker.create will always return an IssueTracker object,
even if it's not valid.


roolebo:
@hennevogel thanks for filing it!

eduardoj (approved):

Issue #11408, opened 2021-07-21
[production][] POST https://api.github.com/repos/openSUSE/open-build-service/statuses/b9ab04955829fef2a47a59b010...


krauselukas:
threw this 247 times
<pre>
POST https://api.github.com/repos/openSUSE/open-build-service/statuses/b9ab04955829fef2a47a59b010e0d4561aba5616: 404 - Not Found // See: https://docs.github.com/rest/reference/repos#create-a-commit-status
</pre>

[See this exception on Errbit](http://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/60f015a584bf4607dfab953e)

krauselukas:
When a user uses a github token with insufficient permissions, the `Octokit` gem throws an `Octokit::NotFound` exception which is not handled in the `SCMStatusReporter` https://github.com/openSUSE/open-build-service/blob/48b331bf94f1ca32559a877f752a6f8eb63cf8ef/src/api/app/services/scm_status_reporter.rb#L16

Since this will happen more frequently with a growing number of users, we should handle this properly.

dmarcoux:
![](https://github.trello.services/images/mini-trello-icon.png) [Handle Octokit::NotFound exception in SCMStatusReporter](https://trello.com/c/xZJHK2Ek/2300-handle-octokitnotfound-exception-in-scmstatusreporter)

dmarcoux:
![](https://github.trello.services/images/mini-trello-icon.png) [Handle Gitlab/Octokit Errors in SCMStatusReporter](https://trello.com/c/NgyLPdA8/2303-handle-gitlab-octokit-errors-in-scmstatusreporter)

dmarcoux:
Closed in #11592

PR #11592, opened 2021-09-08
Rescue Octokit and Gitlab exceptions and report to errbit

### LINKED ISSUE #11438: [production][] POST https://api.github.com/repos/vpereira/hello_world/statuses/5ca5adbea86f15e42a708deef2189f8b0...
threw this 162 times
<pre>
POST https://api.github.com/repos/vpereira/hello_world/statuses/5ca5adbea86f15e42a708deef2189f8b06551d68: 401 - Bad credentials // See: https://docs.github.com/rest
</pre>

[See this exception on Errbit](http://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/6102920084bf4607ebfd7302)

rubhanazeem:
Closes #11438, #11408

### DIFF
--- a/src/api/app/services/scm_exception_handler.rb
+++ b/src/api/app/services/scm_exception_handler.rb
@@ -1,0 +1,55 @@
+class SCMExceptionHandler
+  include ActiveSupport::Rescuable
+  attr_accessor :event_payload, :event_subscription_payload
+
+  rescue_from Octokit::AbuseDetected,
+              Octokit::AccountSuspended,
+              Octokit::BillingIssue,
+              Octokit::BranchNotProtected,
+              Octokit::Conflict,
+              Octokit::Forbidden,
+              Octokit::RepositoryUnavailable,
+              Octokit::NotFound,
+              Octokit::OneTimePasswordRequired,
+              Octokit::Unauthorized,
+              Octokit::UnavailableForLegalReasons,
+              Octokit::UnsupportedMediaType,
+              Octokit::CommitIsNotPartOfPullRequest,
+              Octokit::InstallationSuspended,
+              Octokit::SAMLProtected,
+              Octokit::TooLargeContent,
+              Octokit::TooManyLoginAttempts,
+              Octokit::UnverifiedEmail,
+              Octokit::InvalidRepository,
+              Octokit::PathDiffTooLarge,
+              Octokit::ServiceUnavailable,
+              Octokit::InternalServerError do |exception|
+    # FIXME: Inform users about the exceptions
+    log(exception)
+  end
+
+  rescue_from Gitlab::Error::Conflict,
+              Gitlab::Error::Forbidden,
+              Gitlab::Error::InternalServerError,
+              Gitlab::Error::MissingCredentials,
+              Gitlab::Error::NotFound,
+              Gitlab::Error::ServiceUnavailable,
+              Gitlab::Error::TooManyRequests,
+              Gitlab::Error::Unauthorized do |exception|
+    # FIXME: Inform users about the exceptions
+    log(exception)
+  end
+
+  def initialize(event_payload, event_subscription_payload, scm_token)
+    @event_payload = event_payload.deep_symbolize_keys
+    @event_subscription_payload = event_subscription_payload.deep_symbolize_keys
+    @scm_token = scm_token
+  end
+
+  private
+
+  def log(exception)
+    token = Token::Workflow.find_by(scm_token: @scm_token)
+    Rails.logger.error "#{exception.class}: #{exception.message}. TokenID: #{token.id}, User: #{token.user.login}, Event Subscription Payload: #{@event_subscription_payload}"
+  end
+end
--- a/src/api/app/services/scm_status_reporter.rb
+++ b/src/api/app/services/scm_status_reporter.rb
@@ -1,10 +1,8 @@
-class SCMStatusReporter
-  attr_accessor :event_payload, :event_subscription_payload, :scm_token, :state
+class SCMStatusReporter < SCMExceptionHandler
+  attr_accessor :state
 
   def initialize(event_payload, event_subscription_payload, scm_token, event_type = nil)
-    @event_payload = event_payload.deep_symbolize_keys
-    @event_subscription_payload = event_subscription_payload.deep_symbolize_keys
-    @scm_token = scm_token
+    super(event_payload, event_subscription_payload, scm_token)
 
     @state = event_type.nil? ? 'pending' : scm_final_state(event_type)
   end
@@ -26,6 +24,8 @@ class SCMStatusReporter
                                          @state,
                                          status_options)
     end
+  rescue Octokit::Error, Octokit::InvalidRepository, Gitlab::Error::Error => e
+    rescue_with_handler(e) || raise(e)
   end
 
   private
--- a/src/api/config/initializers/zeitwerk.rb
+++ b/src/api/config/initializers/zeitwerk.rb
@@ -11,6 +11,7 @@ Rails.autoloaders.each do |autoloader|
     'rss_channel' => 'RSSChannel',
     'url_generator' => 'URLGenerator',
     'scm_status_reporter' => 'SCMStatusReporter',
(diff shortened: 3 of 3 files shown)


rubhanazeem:
[Handle Gitlab/Octokit Errors in SCMStatusReporter](https://trello.com/c/NgyLPdA8/1745-handle-gitlab-octokit-errors-in-scmstatusreporter)

src/api/app/services/scm_status_reporter.rb
```diff
@@ -13,18 +13,26 @@ def call
...
-                                  @event_subscription_payload[:commit_sha],
-                                  @state,
-                                  status_options)
+      begin
+        github_client.create_status(@event_subscription_payload[:target_repository_full_name],
+                                    @event_subscription_payload[:commit_sha],
+                                    @state,
+                                    status_options)
+      rescue Octokit::NotFound, Octokit::InternalServerError => e
+        Airbrake.notify(e)
```
hennevogel, line 22: This makes very little sense as *exactly* this will happen automatically right?

rubhanazeem: Do you mean `errbit` will be informed automatically?

dmarcoux: Yes, whenever an exception is raised without being rescued, Errbit is notified.

rubhanazeem: Thanks for the correction. It's updated.

src/api/app/services/scm_status_reporter.rb
```diff
@@ -26,6 +78,9 @@ def call
                                          @state,
                                          status_options)
     end
+  rescue StandardError => e
```
rubhanazeem, line 81: Rubocop prefers to use `StandardError` for blind exceptions.

vpereira: wouldn't that be easier to do something like
```
ALL_AUTH_EXCEPTIONS = OCTOKIT_USER_ERRORS + GITLAB_USER_ERRORS
...

rescue *ALL_AUTH_EXCEPTIONS => e
  Airbreak.notify(e)
end
```

rubhanazeem: This will only notify about the auth exceptions. What we need is to catch every possible exception. If the exception is coming from the list defined above then inform the user about that, otherwise, send it to errbit.

vpereira: hm, but "catch" all exceptions `StandardError` but notify just the auth exceptions, isn't task of this service.

hennevogel: You don't have to catch the other exceptions to re-raise them. Just let me bubble up, if no one else catches them they end up with Airbrake.

src/api/app/services/scm_status_reporter.rb
```diff
@@ -26,6 +28,8 @@ def call
...
+    rescue_with_handler(e) || raise(e)
   end
```
hennevogel, line 33:
```suggestion
  rescue Octokit, Gitlab::Error => e
    rescue_with_handler(e)
  end
```

src/api/app/services/scm_status_reporter.rb
rubhanazeem: I don't have the list of exceptions available anymore. I've removed that list. I can still add that list and catch only those but there will be some duplication which I want to avoid. Have at this one https://github.com/openSUSE/open-build-service/pull/11592/files#diff-bc3a2624b259363910c0947fe29487b8d34e43ceccb8b0fe35efd463c84cd755R4

src/api/app/services/scm_status_reporter.rb
rubhanazeem: @hennevogel the rescue exceptions list has been updated but we still need to use this `raise(e)`. `rescue_with_handler` apparently ignores every exception that is not mentioned in the rescue statement.

src/api/app/services/scm_status_reporter.rb
```diff
@@ -1,7 +1,9 @@
-class SCMStatusReporter
+class SCMStatusReporter < SCMExceptionHandler
   attr_accessor :event_payload, :event_subscription_payload, :scm_token, :state
 
   def initialize(event_payload, event_subscription_payload, scm_token, event_type = nil)
+    super()
```
vpereira, line 5: do you need the parenthesis here?

rubhanazeem: Yes, otherwise it passes all parameters to the parent class. To avoid parameters I need paranthesis.

src/api/app/services/scm_exception_handler.rb
```diff
@@ -0,0 +1,38 @@
...
+    # FIXME: Inform users about the exceptions
+  end
+
+  rescue_from Gitlab::Error::Conflict,
+              Gitlab::Error::Forbidden,
+              Gitlab::Error::InternalServerError,
+              Gitlab::Error::MissingCredentials,
+              Gitlab::Error::NotFound,
+              Gitlab::Error::ServiceUnavailable,
+              Gitlab::Error::Unauthorized do |exception|
```
dmarcoux, line 35: I think we should add `Gitlab::Error::TooManyRequests` to this list if somehow the user gets rate-limited. They could always try later. It's quite unlikely, but who knows...

src/api/app/services/scm_exception_handler.rb
```diff
@@ -0,0 +1,38 @@
...
+              Octokit::InstallationSuspended,
+              Octokit::SAMLProtected,
+              Octokit::TooLargeContent,
+              Octokit::TooManyLoginAttempts,
+              Octokit::UnverifiedEmail,
+              Octokit::InvalidRepository,
+              Octokit::PathDiffTooLarge,
+              Octokit::ServiceUnavailable,
+              Octokit::InternalServerError do |exception|
+    # FIXME: Inform users about the exceptions
```
hennevogel, line 26: Can we at least log this so we have something to debug the problems of people?

hennevogel (changes requested):

src/api/app/services/scm_exception_handler.rb
vpereira: if we have the request id, couldn't we just grep the log file for that and get the stacktrace from there - or from errbit?

rubhanazeem: It won't add anything to errbit. I'm not sure about the request id.

rubhanazeem: I've added an error logger.

vpereira: my question was to @hennevogel. I wanted exactly to avoid the "Rails logger" calls everywhere. There should be better way to introspect code in 2021, other than custom stuff on error logger, right?

hennevogel: Logging is a temporary measure, just to be able to debug this for people for now. In the future we are going to expose this to the users. Because it is *them* who can do something about this. we can't.

dmarcoux (approved):

src/api/app/services/scm_exception_handler.rb
```diff
@@ -0,0 +1,55 @@
...
+  def initialize(event_payload, event_subscription_payload, scm_token)
+    @event_payload = event_payload.deep_symbolize_keys
+    @event_subscription_payload = event_subscription_payload.deep_symbolize_keys
+    @scm_token = scm_token
+  end
+
+  private
+
+  def log(exception)
+    token = Token::Workflow.find_by(scm_token: @scm_token)
```
dmarcoux, line 52: It could be possible that two tokens have the same `scm_token`, although it's really an edge case... It's fine to not handle this.

dmarcoux (approved):
:+1: for the changes with the `log` method.

src/api/app/services/scm_exception_handler.rb
dmarcoux: @hennevogel Please have another look.

PR #11595, opened 2021-09-08
Adapt LinkPageStep to handle PR merge and reopening events

vpereira:
Until now we just ignored both scenarios.. from now on we are going to handle them accordingly

TODO:
- [x] ~Adapt test suite to cover new introduced scenarios~ I didn't see necessity since the code coverage is big

### DIFF
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -111,6 +111,10 @@ class Workflow::Step
     end
   end
 
+  def delete_subscriptions(package)
+    EventSubscription.where(channel: 'scm', token: @token, package: package).delete_all
+  end
+
   def update_subscriptions(package, workflow_filters)
     ['Event::BuildFail', 'Event::BuildSuccess'].each do |build_event|
       subscription = EventSubscription.find_by(eventtype: build_event,
--- a/src/api/app/models/workflow/step/link_package_step.rb
+++ b/src/api/app/models/workflow/step/link_package_step.rb
@@ -5,25 +5,52 @@ class Workflow::Step::LinkPackageStep < ::Workflow::Step
   def call(options = {})
     return unless valid?
 
-    # FIXME: Support closed/merged/reopened PRs
-    return if scm_webhook.closed_merged_pull_request? || scm_webhook.reopened_pull_request?
-
     workflow_filters = options.fetch(:workflow_filters, {})
 
-    # Updated PR
-    if scm_webhook.updated_pull_request? && target_package.present?
-      update_subscriptions(target_package, workflow_filters)
-    else # New PR
+    case
+    when scm_webhook.new_pull_request?
+      create_link_package(workflow_filters)
+    when scm_webhook.updated_pull_request?
+      update_link_package(workflow_filters)
+    when scm_webhook.closed_merged_pull_request?
+      destroy_target_project
+    when scm_webhook.reopened_pull_request?
+      restore_target_project(workflow_filters)
+    end
+  end
+
+  private
+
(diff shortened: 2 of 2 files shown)


dmarcoux:
[Support closing/merging/reopening actions for PRs/MRs](https://trello.com/c/nlIsH1Wp/1737-support-closing-merging-reopening-actions-for-prs-mrs)

src/api/app/models/workflow/step.rb
```diff
@@ -111,6 +111,20 @@ def create_subscriptions(package, workflow_filters)
     end
   end
 
+  def destroy_subscriptions(package, _workflow_filters)
+    ['Event::BuildFail', 'Event::BuildSuccess'].each do |build_event|
+      subscription = EventSubscription.find_by(eventtype: build_event,
+                                               channel: 'scm',
+                                               token: @token,
+                                               package: package)
+      subscription.destroy
```
dmarcoux, line 120: There are no before/after destroy callbacks on the `EventSubscription` model, so we could delete them all in one SQL query with `EventSubscription.where(...).delete_all`. This would be more efficient.

vpereira: so, I just uncommented my code and adapted to use `delete_all` @dmarcoux

src/api/app/models/workflow/step.rb
```diff
@@ -111,6 +111,10 @@ def create_subscriptions(package, workflow_filters)
     end
   end
 
+  def destroy_subscriptions(package, _workflow_filters)
```
dmarcoux, line 114: I don't see why we would ever need workflow filters when deleting `EventSubscription` records.

```suggestion
  def destroy_subscriptions(package)
```

src/api/app/models/workflow/step.rb
```diff
@@ -111,6 +111,10 @@ def create_subscriptions(package, workflow_filters)
     end
   end
 
+  def destroy_subscriptions(package)
```
dmarcoux, line 114: We should still rename the method so it's clear that it's calling `delete_all` and not `destroy_all`.

```suggestion
  def delete_all_subscriptions(package)
```

src/api/app/models/workflow/step.rb
vpereira: done

src/api/app/models/workflow/step.rb
vpereira: I prefer without `_all_` thank you for the suggestion.

src/api/app/models/workflow/step/link_package_step.rb
```diff
@@ -5,25 +5,52 @@ class Workflow::Step::LinkPackageStep < ::Workflow::Step
...
+      destroy_target_project
+    when scm_webhook.reopened_pull_request?
+      restore_target_project(workflow_filters)
+    end
+  end
+
+  private
+
+  def update_link_package(workflow_filters)
+    if target_package.blank?
```
vpereira, line 25: maybe we should have a method named `create_or_update_subscriptions` where we create subscriptions if they don't exist otherwise we update them.. what do you think @dmarcoux

dmarcoux: This method already exists in `Workflow::Step`. We also have `create_subscriptions` and `update_subscriptions`. I'm not sure if we either go with the separate methods or the single one which combines both. Do you have a preference?

https://github.com/openSUSE/open-build-service/blob/2acf0549256926c6b96190bab7a9f054e44ad447/src/api/app/models/workflow/step.rb#L87-L99

dmarcoux:
Superseded by #11596

PR #11599, opened 2021-09-10
Replace exceptions by errors

saraycp:
Set errors instead of exceptions in WorkflowStepsValidator and WorkflowFiltersValidator, so the user can know all the non-valid attributes at once.

Also, improve error messages and adapt specs.


src/api/app/validators/workflow_steps_validator.rb
```diff
@@ -30,14 +29,20 @@ def no_steps?
 
   def errors
     acc = []
```
vpereira, line 31: what is `acc`? couldn't we use a better name here?

saraycp: definitely!

saraycp: done

saraycp:
@vpereira I've added an extra test in a new commit.

vpereira (approved):

saraycp:
> Errors `UnsupportedWorkflowFilterTypes` and `InvalidWorkflowStepDefinition` are no longer raised. We could remove their definitions, in `app/models/token/errors.rb` and in `app/models/workflow/errors.rb`.

Thanks! Updated @eduardoj

eduardoj (approved):

PR #11600, opened 2021-09-10
Consider multibuild suffix when retriggering package build through webui

### LINKED ISSUE #11598: webui: rebuild trigger on multibuild always triggers flavor ""
Ran into that trap a few times:

e.g.

https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:C/qemu:testsuite/standard/x86_64

The build shows 'failed', so I want to give it a retry - click 'trigger rebuild' at the bottom.

The webui takes the command, but instead of triggering a rebuild of the :testsuite flavor, OBS triggers a rebuild of the "" flavor.

using osc rebuildpac …:testsuite triggers the thing I expect to be triggered

krauselukas:
We need to pass the full package name (including the multibuild suffix) to the backend
when retriggering a package build through the webui.
Currently we pass the package name of the package object
to the backend, which will cause a rebuild of all
multibuild flavors.

Fixes #11598

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -504,9 +504,8 @@ class Webui::PackageController < Webui::WebuiController
   end
 
   def trigger_rebuild
-    rebuild_trigger = PackageControllerService::RebuildTrigger.new(package: @package, project: @project,
-                                                                   repository: params[:repository],
-                                                                   arch: params[:arch])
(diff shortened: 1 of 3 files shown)


krauselukas:
I branched a multibuild package in the review app in order to make it easier for you to test the changes https://obs-reviewlab.opensuse.org/krauselukas-fixissue_11598/package/show/home:Admin/apache2

vpereira (approved):

eduardoj (approved):

PR #11601, opened 2021-09-10
Refactor BranchPackageStep and LinkPackageStep

saraycp:
Refactor BranchPackageStep the same way we did with LinkPackageStep.

- Unify variable names
- Simplify methods
- Split creation and update code following the [KISS principle](https://www.interaction-design.org/literature/topics/keep-it-simple-stupid). Except for `create_or_update_subscriptions`.
- Move common code to Workflow::Step

Follow up of #11413

*NOTE:* I tried to make every commit as simple as possible, they can be squash before merging.

### DIFF
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -145,5 +145,16 @@ class Workflow::Step
 
     architectures
   end
+
+  def report_to_scm(workflow_filters)
+    workflow_repositories(target_project_name, workflow_filters).each do |repository|
+      # TODO: Fix n+1 queries
+      workflow_architectures(repository, workflow_filters).each do |architecture|
+        # We cannot report multibuild flavors here... so they will be missing from the initial report
+        SCMStatusReporter.new({ project: target_project_name, package: target_package_name, repository: repository.name, arch: architecture.name },
+                              scm_webhook.payload, @token.scm_token).call
+      end
+    end
+  end
 end
 # rubocop:enable Metrics/ClassLength
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -1,4 +1,3 @@
-# rubocop:disable Metrics/ClassLength
 class Workflow::Step
   include ActiveModel::Model
 
@@ -24,8 +23,6 @@ class Workflow::Step
     Package.find_by_project_and_name(target_project_name, target_package_name)
   end
 
-  # FIXME: Remove this and use create_subscriptions and update_subscriptions as soon as BranchPackageStep and Token::Workflow are refactored
-  #        This method is public since it has to be called from Token::Workflow. This shouldn't be needed.
   def create_or_update_subscriptions(package, workflow_filters)
     ['Event::BuildFail', 'Event::BuildSuccess'].each do |build_event|
       subscription = EventSubscription.find_or_create_by!(eventtype: build_event,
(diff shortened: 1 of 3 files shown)


src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -8,48 +8,24 @@ def call(options = {})
...
+    # Updated PR
+    if scm_webhook.updated_pull_request? && target_package.present?
+      update_subscriptions(target_package, workflow_filters)
+    else # New PR or updated PR but target_package is not set
+      create_target_package
+      create_subscriptions(target_package, workflow_filters)
     end
 
-    branched_package
+    add_or_update_branch_request_file(package: target_package)
```
vpereira, line 21: what I don't understand @saraycp  is why this `add_or_update_branch_request_file` is different than `create_or_update_subscriptions`? Shouldn't we then break it in two methods?

saraycp: If we had two methods, they'll be identical. I think I'm going to rename the method, that's better

saraycp: done!

src/api/app/models/workflow/step.rb
```diff
@@ -24,21 +24,6 @@ def target_package
     Package.find_by_project_and_name(target_project_name, target_package_name)
   end
 
-  # FIXME: Remove this and use create_subscriptions and update_subscriptions as soon as BranchPackageStep and Token::Workflow are refactored
-  #        This method is public since it has to be called from Token::Workflow. This shouldn't be needed.
-  def create_or_update_subscriptions(package, workflow_filters)
```
dmarcoux, line 29: This method is still used here:
https://github.com/openSUSE/open-build-service/blob/a334e384ecc45aed5c9a7234a26ec78490c34757/src/api/app/models/token/workflow.rb#L66

saraycp: I wonder if it's really a good idea to split `create_or_update_subscriptions` into two separate methods just to follow the KISS principle. This is forcing to add more conditions with repeated code.

This piece of code

```
if scm_webhook.updated_pull_request?
  if target_package.blank?
    create_target_package
    create_subscriptions(target_package, workflow_filters)
  else
    update_subscriptions(target_package, workflow_filters)
  end
elsif scm_webhook.new_pull_request?
  create_target_package
  create_subscriptions(target_package, workflow_filters)
end
 ```
 
 can be simplified to:

```
create_target_package if scm_webhook.new_pull_request? || (scm_webhook.updated_pull_request? && target_package.blank?)
create_or_update_subscriptions(target_package, workflow_filters)
 ```

@dmarcoux  Would you agree if I bring back `create_or_update_subscriptions` only?

dmarcoux: I'm fine with having `create_or_update_subscriptions`, as long as we don't keep `create_subscriptions` and `update_subscriptions` around.

saraycp: Ok. Code updated! The commits are different. Please have another look

dmarcoux:
It's not a blocker for this PR, but it's a lot of commits for only 3 files changed.

saraycp:
> It's not a blocker for this PR, but it's a lot of commits for only 3 files changed.

As I mentioned in the description, I added simple commits to simplify revision and will squash them before merging.

saraycp:
> > It's not a blocker for this PR, but it's a lot of commits for only 3 files changed.
> 
> As I mentioned in the description, I added simple commits to simplify revision and will squash them before merging.

@dmarcoux I have already squashed most of the commits

vpereira (approved):

PR #11610, opened 2021-09-13
Add missing "Tokens" link in breadcrumbs in "Trigger Token" page

eduardoj:
Before:

Screenshot from 2021-09-13 15-59-34

After:

Screenshot from 2021-09-13 15-59-53

### DIFF
--- a/src/api/app/views/webui/users/token_triggers/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui/users/token_triggers/_breadcrumb_items.html.haml
@@ -2,5 +2,7 @@
   = link_to('Your Profile', user_path(User.session!))
 
 - if action_name == 'show'
+  %li.breadcrumb-item
+    = link_to 'Tokens', tokens_path
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
     Trigger Token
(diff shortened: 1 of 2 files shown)


vpereira (approved):

src/api/app/views/webui/users/token_triggers/_breadcrumb_items.html.haml
```diff
@@ -2,5 +2,7 @@
   = link_to('Your Profile', user_path(User.session!))
 
 - if action_name == 'show'
+  %li.breadcrumb-item{ 'aria-current' => 'page' }
```
dmarcoux, line 5:
```suggestion
  %li.breadcrumb-item
```

This isn't the current page, so setting `aria-current` is wrong here.

I've noticed that this is also wrong in the breadcrumbs from `src/api/app/views/webui/users/tokens/_breadcrumb_items.html.haml`. This could also be fixed in this PR: 

https://github.com/openSUSE/open-build-service/blob/528a66a6ee5e55b497f561f092e837fbc05abc07/src/api/app/views/webui/users/tokens/_breadcrumb_items.html.haml#L9-L10

https://github.com/openSUSE/open-build-service/blob/528a66a6ee5e55b497f561f092e837fbc05abc07/src/api/app/views/webui/users/tokens/_breadcrumb_items.html.haml#L14-L15

dmarcoux (changes requested):

src/api/app/views/webui/users/token_triggers/_breadcrumb_items.html.haml
eduardoj: True. Fixed. Thanks!

dmarcoux (approved):

saraycp (approved):

PR #11617, opened 2021-09-15
Create SignUp component

vpereira:
### TODO:

- [x] adapt all views to use it
- [x] add specs
- [x] clean up unnecessary views and helpers


vpereira:
Accessing the route `/users/new` leads you to the new component

src/api/app/components/sign_up_component.rb
```diff
@@ -0,0 +1,18 @@
...
+class SignUpComponent < ApplicationComponent
+  attr_accessor :submit_btn_text
+
+  def initialize(submit_btn_text:)
+    super
+    @submit_btn_text = submit_btn_text || 'Sign up'
+  end
+
+  def proxy_auth_enabled?
+    CONFIG['proxy_auth_mode'] == :on
```
dmarcoux, line 12: This view component would be easier to test if it doesn't depend on `CONFIG`. I would prefer to pass the config in the `initialize`.

src/api/app/components/sign_up_component.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class SignUpComponent < ApplicationComponent
+  attr_accessor :submit_btn_text
+
+  def initialize(submit_btn_text:)
```
dmarcoux, line 6: _Nitpick_

Instead of doing `@submit_btn_text = submit_btn_text || 'Sign up'` on line 8, I suggest this:
```suggestion
  def initialize(submit_btn_text: 'Sign up')
```

Then the line 8 would be `@submit_btn_text = submit_btn_text`.

src/api/app/views/webui/main/index.html.haml
```diff
@@ -22,7 +22,7 @@
       .card.mb-3
         %h5.card-header New here? Sign up!
         .card-body
-          = render partial: 'webui/shared/sign_up', locals: { submit_btn_text: 'Sign Up' }
+          = render SignUpComponent.new(submit_btn_text: 'Sign up')
```
dmarcoux, line 25:
```suggestion
          = render SignUpComponent.new(submit_btn_text: 'Sign Up')
```

src/api/app/components/sign_up_component.rb
vpereira: That's why my first try, but sometimes we pass to the component params[:submit_btn_text] which can be nil, look https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/users_controller.rb#L45

src/api/app/components/sign_up_component.rb
vpereira: I'm passing it now, but IMO to stub some keys in a hash isn't really hard...

src/api/app/components/sign_up_component.rb
dmarcoux: > I would prefer we get rid of passing this via `params`

I agree with you. If you can get rid of this, it's great!

src/api/app/views/webui/main/index.html.haml
```diff
@@ -22,7 +22,7 @@
       .card.mb-3
         %h5.card-header New here? Sign up!
         .card-body
-          = render partial: 'webui/shared/sign_up', locals: { submit_btn_text: 'Sign Up' }
+          = render SignUpComponent.new(submit_btn_text: 'Sign Up')
```
dmarcoux, line 25: `Sign Up` is the default, so we can simply call `new`.

```suggestion
          = render SignUpComponent.new
```

dmarcoux:
Any plans to also replace https://github.com/openSUSE/open-build-service/blob/11a5904148576ffb4a699e2d811d97ab07fda88c/src/api/app/views/layouts/webui/_sign_up_form.html.haml with this new component?

vpereira:
> Any plans to also replace https://github.com/openSUSE/open-build-service/blob/11a5904148576ffb4a699e2d811d97ab07fda88c/src/api/app/views/layouts/webui/_sign_up_form.html.haml with this new component?

That's other component. I can tackle it in a single PR, since it may or may not be able to replace by this component.

dmarcoux (approved):
Nice work!

PR #11623, opened 2021-09-15
Rename source project to project for workflows

saraycp:
Replace `source_project` by `project` in ConfigureRepositoriesStep 

The 'source_project_name' shouldn't be required for ConfigureRepositoriesStep anymore. For this step, I create a new method `project_name` and add the corresponding validation.

As a consequence, the `target_project_name` method is affected so the definition is moved to each type of step: BranchPackageStep, LinkPackageStep and ConfigureRepositoriesStep

### DIFF
--- a/src/api/app/models/workflow/step/branch_package_step.rb
+++ b/src/api/app/models/workflow/step/branch_package_step.rb
@@ -1,5 +1,7 @@
 class Workflow::Step::BranchPackageStep < ::Workflow::Step
   REQUIRED_KEYS = [:source_project, :source_package].freeze
+
+  validates :source_project_name, presence: true
   validates :source_package_name, presence: true
 
   def call(options = {})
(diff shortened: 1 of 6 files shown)


danidoni (approved):

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -9,6 +11,14 @@ def call(options = {})
     branch_package(workflow_filters)
   end
 
+  def source_project_name
+    step_instructions[:source_project]
+  end
+
+  def target_project_name
+    "home:#{@token.user.login}:#{source_project_name}:PR-#{scm_webhook.payload[:pr_number]}"
```
dmarcoux, line 19: I don't really like that this is duplicated now, because this code is going to change in our current deliverable and the logic will be duplicated in 3 steps.

saraycp: Ok, I can move it back to Workflow::Step and overwrite it inside ConfigureRepositoriesStep.

saraycp: done

dmarcoux (approved):

PR #11624, opened 2021-09-16
[backend] Add experimental forcebinaryidmeta build flag

mlschroe:
This always uses the binary id instead of the sourcemd5 in
the meta file. Turning this on easily leads to endless rebuilds
if the project contains build cycles and there is no other
means like build compare to end builds.


coolo:
I also had to add nometa = 1 in getpreinstallimage - or it would generate invalid meta and cause endless rebuilds of the same package

PR #11625, opened 2021-09-16
Refactor the Workflow step calling responsibility

danidoni:
The workflow should be the responsible of calling its steps, this way it
has better control on the step execution.

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -13,67 +13,9 @@ class Token::Workflow < Token
 
     yaml_file = Workflows::YAMLDownloader.new(scm_webhook.payload, token: self).call
     workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: scm_webhook, token: self).call
-
-    case
-    when scm_webhook.closed_merged_pull_request?
-      destroy_target_projects(workflows)
-    when scm_webhook.reopened_pull_request?
-      restore_target_projects(workflows)
-    when scm_webhook.new_pull_request?, scm_webhook.updated_pull_request?
-      call_steps(workflows)
-    end
+    workflows.each(&:call)
   rescue Octokit::Unauthorized, Gitlab::Error::Unauthorized => e
     raise Token::Errors::SCMTokenInvalid, e.message
-  end
-
-  private
-
-  def destroy_target_projects(workflows)
-    workflows.each do |workflow|
-      # Do not process steps for which there's nothing to do
-      workflow_steps = workflow.steps.reject { |step| step.instance_of?(::Workflow::Step::ConfigureRepositories) }
-      target_packages = workflow_steps.map(&:target_package).uniq.compact
-      delete_subscriptions(target_packages)
-
-      target_project_names = workflow_steps.map(&:target_project_name).uniq.compact
-      destroy_all_target_projects(target_project_names)
-    end
-  end
-
-  # We want the callbacks to run after destroy, so we can't use delete_all
-  def destroy_all_target_projects(target_project_names)
-    Project.where(name: target_project_names).destroy_all
-  end
-
-  def delete_subscriptions(packages)
-    EventSubscription.where(channel: 'scm', token: self, package: packages).delete_all
-  end
-
-  def restore_target_projects(workflows)
-    token_user_login = user.login
-
-    workflows.each do |workflow|
-      # Do not process steps for which there's nothing to do
-      workflow_steps = workflow.steps.reject { |step| step.instance_of?(::Workflow::Step::ConfigureRepositories) }
-      target_project_names = workflow_steps.map(&:target_project_name).uniq.compact
-      target_project_names.each do |target_project_name|
-        Project.restore(target_project_name, user: token_user_login)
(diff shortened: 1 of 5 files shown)


src/api/app/models/token/workflow.rb
```diff
@@ -14,17 +15,21 @@ def call(options)
     yaml_file = Workflows::YAMLDownloader.new(scm_webhook.payload, token: self).call
     workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: scm_webhook, token: self).call
 
+    # FIXME: The scm_webhook has hidden objects.
```
dmarcoux, line 18: What do you mean exactly with `hidden objects`? Is it that `ScmWebhook` could be split in multiple classes?

danidoni: Sorry, the `Token::Workflow` object has hidden objects inside not the `scm_webhook`.

danidoni: I've updated the comment so it is clearer

src/api/app/models/workflow.rb
```diff
@@ -21,6 +21,19 @@ def initialize(attributes = {})
   validates_with WorkflowStepsValidator
   validates_with WorkflowFiltersValidator
 
+  def call(user)
+    case
```
danidoni, line 25: Here we will implement the filters defined in [our trello](https://trello.com/c/BLeVSxhC/1751-implement-branches-filter), under the complete control of the workflow.

danidoni:
[Implement branches filter](https://trello.com/c/BLeVSxhC/1751-implement-branches-filter)

src/api/spec/models/token/workflow_spec.rb
```diff
@@ -135,15 +135,18 @@
       let(:payload) { github_payload_closed }
       let(:workflows_yml_file) { File.expand_path(Rails.root.join('spec/support/files/workflows.yml')) }
       let(:downloader) { instance_double(Workflows::YAMLDownloader) }
+      let(:target_project) { create(:project, name: 'home:Iggy:test-project:PR-4') }
 
       before do
+        target_project
```
eduardoj, line 141:
```suggestion
      let!(:target_project) { create(:project, name: 'home:Iggy:test-project:PR-4') }

      before do
```
Using `let!` prevents from calling `target_project` in the `before` block.

eduardoj (commented):
As the code is moved from `app/models/token/workflow.rb` to `app/models/workflow.rb`, shouldn't be some tests also moved from `spec/models/token/workflow_spec.rb` to `spec/models/workflow_spec.rb`?

src/api/spec/models/token/workflow_spec.rb
danidoni: true

danidoni:
> As the code is moved from `app/models/token/workflow.rb` to `app/models/workflow.rb`, shouldn't be some tests also moved from `spec/models/token/workflow_spec.rb` to `spec/models/workflow_spec.rb`?

Hm, yes. I'll move some of the use cases in the `spec/models/token/workflow_spec.rb` to the `spec/models/workflow_spec.rb`

src/api/app/models/token/workflow.rb
```diff
@@ -13,68 +13,10 @@ def call(options)
...
-
-    case
-    when scm_webhook.closed_merged_pull_request?
-      destroy_target_projects(workflows)
-    when scm_webhook.reopened_pull_request?
-      restore_target_projects(workflows)
-    when scm_webhook.new_pull_request?, scm_webhook.updated_pull_request?
-      call_steps(workflows)
-    end
+    workflows.each { |workflow| workflow.call(user) }
```
dmarcoux, line 16: We don't need to pass the user to the workflow since the workflow already has an instance of the current token, so it's possible to do `token.user` in the workflow.

src/api/app/models/workflow.rb
```diff
@@ -68,4 +81,32 @@ def initialize_step(step_name, step_instructions)
...
+    target_packages = steps.map(&:target_package).uniq.compact
+    EventSubscription.where(channel: 'scm', token: self, package: target_packages).delete_all
+
+    target_project_names = processable_steps.map(&:target_project_name).uniq.compact
+    # We want the callbacks to run after destroy, so we can't use delete_all
+    Project.where(name: target_project_names).destroy_all
+  end
+
+  def restore_target_projects(user)
+    token_user_login = user.login
```
dmarcoux, line 97: This goes along with my previous comment in `Token::Workflow`, we should use the token user:

```suggestion
    token_user_login = token.user.login
```

src/api/app/models/token/workflow.rb
danidoni: True, fixed

danidoni:
> As the code is moved from `app/models/token/workflow.rb` to `app/models/workflow.rb`, shouldn't be some tests also moved from `spec/models/token/workflow_spec.rb` to `spec/models/workflow_spec.rb`?

Check it out again, @eduardoj !

eduardoj (approved):

dmarcoux (approved):

PR #11630, opened 2021-09-17
Check presence of name before checking multibuild suffix

### LINKED ISSUE #11629: [production][] undefined method `start_with?' for nil:NilClass
request#global_command threw this 44 times
<pre>
undefined method `start_with?&#39; for nil:NilClass
</pre>
[https://api.opensuse.org/request?cmd=create&amp;addrevision=1](https://api.opensuse.org/request?cmd=create&amp;addrevision=1)"

[See this exception on Errbit](http://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/602a91fe84bf4605432b23c8)

rubhanazeem:
Fixes #11629

### DIFF
--- a/src/api/app/models/concerns/multibuild_package.rb
+++ b/src/api/app/models/concerns/multibuild_package.rb
@@ -7,6 +7,8 @@ module MultibuildPackage
     end
 
     def striping_multibuild_suffix(name)
+      return name if name.blank?
+
       # exception for package names used to have a collon
       return name if name.start_with?('_patchinfo:', '_product:')
 


src/api/app/models/concerns/multibuild_package.rb
```diff
@@ -7,6 +7,8 @@ def valid_multibuild_name?(name)
     end
 
     def striping_multibuild_suffix(name)
+      return name if name.blank?
```
hennevogel, line 10: This method is supposed to give back a package name in *any* case. The bug must be in the calling code, with this change the problem would just manifest itself somewhere down the call path.

hennevogel (changes requested):

src/api/app/models/concerns/multibuild_package.rb
rubhanazeem: There are a couple of calls to this method coming from different files. I'll close this PR for now.

PR #11633, opened 2021-09-17
Implement rebuild_package step

saraycp:
See the [developer documentation](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration#rebuild-a-package) for the `rebuild_package` step.

### DIFF
--- a/src/api/app/controllers/concerns/triggerable.rb
+++ b/src/api/app/controllers/concerns/triggerable.rb
@@ -4,11 +4,11 @@ module Triggerable
   def set_project
     # By default we operate on the package association
     @project = @token.package.try(:project)
-    # If the token has no package, let's find one from the parameters
-    @project ||= Project.get_by_name(params[:project])
+    # If the token has no package, let's find one from the parameters or the step intructions
+    @project ||= Project.get_by_name(@project_name)
     # Remote projects are read-only, can't trigger something for them.
     # See https://github.com/openSUSE/open-build-service/wiki/Links#project-links
-    raise Project::Errors::UnknownObjectError, "Sorry, triggering tokens for remote project \"#{params[:project]}\" is not possible." unless @project.is_a?(Project)
+    raise Project::Errors::UnknownObjectError, "Sorry, triggering tokens for remote project \"#{@project_name}\" is not possible." unless @project.is_a?(Project)
   end
 
   def set_package
@@ -16,14 +16,14 @@ module Triggerable
     @package = @token.package
     # If the token has no package, let's find one from the parameters
     @package ||= Package.get_by_project_and_name(@project,
-                                                 params[:package],
+                                                 @package_name,
                                                  @token.package_find_options)
     return unless @project.links_to_remote?
 
     # The token has no package, we did not find a package in the database but the project has a link to remote.
     # See https://github.com/openSUSE/open-build-service/wiki/Links#project-links
     # In this case, we will try to trigger with the user input, no matter what it is
-    @package ||= params[:package]
+    @package ||= @package_name
     # TODO: This should not happen right? But who knows...
     raise ActiveRecord::RecordNotFound unless @package
   end
@@ -34,7 +34,7 @@ module Triggerable
 
   def set_multibuild_flavor
     # Do NOT use @package.multibuild_flavor? here because the flavor need to be checked for the right source revision
-    @multibuild_container = params[:package].gsub(/.*:/, '') if params[:package].present? && params[:package].include?(':')
+    @multibuild_container = @package_name.gsub(/.*:/, '') if @package_name.present? && @package_name.include?(':')
   end
 
   def package_from_project_link?
--- a/src/api/app/controllers/trigger_controller.rb
+++ b/src/api/app/controllers/trigger_controller.rb
@@ -16,6 +16,9 @@ class TriggerController < ApplicationController
 
   before_action :validate_gitlab_event, if: :gitlab_webhook?
   before_action :set_token
+  before_action :set_project_name
+  before_action :set_package_name
+  # From Triggerable
   before_action :set_project
   before_action :set_package
   before_action :set_object_to_authorize
@@ -63,5 +66,13 @@ class TriggerController < ApplicationController
 
   def pundit_user
     @token.user
+  end
+
+  def set_project_name
+    @project_name = params[:project]
+  end
+
+  def set_package_name
+    @package_name = params[:package]
   end
 end
--- a/src/api/app/models/workflow.rb
+++ b/src/api/app/models/workflow.rb
@@ -4,7 +4,8 @@ class Workflow
   SUPPORTED_STEPS = {
     branch_package: Workflow::Step::BranchPackageStep,
     link_package: Workflow::Step::LinkPackageStep,
-    configure_repositories: Workflow::Step::ConfigureRepositories
+    configure_repositories: Workflow::Step::ConfigureRepositories,
+    rebuild_package: Workflow::Step::RebuildPackage
   }.freeze
 
   SUPPORTED_FILTERS = [:architectures, :repositories].freeze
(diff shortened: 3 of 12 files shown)


src/api/app/controllers/concerns/triggerable.rb
```diff
@@ -4,26 +4,26 @@ module Triggerable
   def set_project
     # By default we operate on the package association
     @project = @token.package.try(:project)
-    # If the token has no package, let's find one from the parameters
-    @project ||= Project.get_by_name(params[:project])
+    # If the token has no package, let's find one from the parameters or the step intructions
+    @project ||= Project.get_by_name(@project_name)
```
danidoni, line 8: :thought_balloon:  You could just call the methods that get those values instead of relying in instance variables that will be nil if someone forgets to fill them in. Calling methods that do not exist will throw a nice error that the programmer would notice faster. You could even write some default ones that throw `NotImplementedError` to make even more clear that they should be overriden.

src/api/app/models/workflow/step/rebuild_package.rb
```diff
@@ -0,0 +1,44 @@
+class Workflow::Step::RebuildPackage < ::Workflow::Step
+  include Triggerable
```
hennevogel, line 2: I really feel uncomfortable with using a controller concern in a model. Could the be something else?

src/api/spec/controllers/concerns/triggerable_spec.rb
```diff
@@ -93,13 +103,27 @@ def stub_params(project_name:, package_name:)
...
       it 'authorizes the project if the package is from a project with a link' do
         stub_params(project_name: project_with_a_link.name, package_name: package.name)
         fake_controller_instance.set_project
         fake_controller_instance.set_package
         fake_controller_instance.set_object_to_authorize
         expect(fake_controller_instance.instance_variable_get(:@token).object_to_authorize).to eq(project_with_a_link)
       end
+    end
+
+    context 'project with project-link and a local package' do
```
hennevogel, line 119: Why do we test this now? What changed?

vpereira: in this area nothing changed. I'm assuming @saraycp wanted to test the `project_name` and `package_name` instance variables?

src/api/app/models/workflow/step/rebuild_package.rb
vpereira: maybe because steps shouldn't be models? Another hint is having to call `Pundit.authorize` on that..

vpereira: I would prefer to to let this possible renaming of `Triggerable` controller concern to a different PR. Would that be ok @hennevogel ?

hennevogel: sure thing, let's add a card to the backlog please

dmarcoux: > sure thing, let's add a card to the backlog please

Done

src/api/app/models/workflow.rb
```diff
@@ -84,7 +85,7 @@ def supported_filters
 
   def destroy_target_projects
     # Do not process steps for which there's nothing to do
-    processable_steps = steps.reject { |step| step.instance_of?(::Workflow::Step::ConfigureRepositories) }
+    processable_steps = steps.reject { |step| step.instance_of?(::Workflow::Step::ConfigureRepositories) || step.instance_of?(::Workflow::Step::RebuildPackage) }
```
vpereira, line 88: We should improve it later. Either having a field `skip_destroy_restore` or via class heritage... could it become a card in the backlog @dmarcoux ?

src/api/spec/controllers/concerns/triggerable_spec.rb
hennevogel: What's there to test? If instance variables don't work anymore we can go home anyway :)

vpereira: > What's there to test? If instance variables don't work anymore we can go home anyway :)

No, if using it changed some behavior in the controller, beside the other methods that were updated on Triggerable concern (`set_project` and `set_package`)

src/api/app/controllers/trigger_controller.rb
```diff
@@ -16,6 +16,9 @@ class TriggerController < ApplicationController
 
   before_action :validate_gitlab_event, if: :gitlab_webhook?
   before_action :set_token
+  before_action :set_project_name
```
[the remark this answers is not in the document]
vpereira: yes, almost like `ApplicationController` and `Webui::Requests::SubmitionsController`... anything concrete to be done in this PR?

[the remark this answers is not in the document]
dmarcoux: I created a card for this in our backlog.

src/api/spec/controllers/concerns/triggerable_spec.rb
hennevogel: The old tests should show this no?

src/api/app/models/workflow/step/rebuild_package.rb
```diff
@@ -0,0 +1,39 @@
...
+    # Call Triggerable method to set all the elements needed for rebuilding
+    set_project_name
+    set_package_name
+    set_project
+    set_package
+    set_object_to_authorize
+    set_multibuild_flavor
+
+    Pundit.authorize(@token.user, @token, :rebuild?)
+    # Perform rebuild
```
dmarcoux, line 21: It's quite obvious that `rebuild_package` rebuilds the package.

```suggestion
```

src/api/app/models/workflow/step/rebuild_package.rb
```diff
@@ -0,0 +1,39 @@
...
+    @project_name = step_instructions[:project]
+  end
+
+  def set_package_name
+    @package_name = step_instructions[:package]
+  end
+
+  private
+
+  def rebuild_package(_workflow_filters = {})
```
dmarcoux, line 35: We don't need workflow filters, so let's remove it.

```suggestion
  def rebuild_package
```

src/api/app/models/workflow/step/rebuild_package.rb
```diff
@@ -0,0 +1,39 @@
...
+    Backend::Api::Sources::Package.rebuild(project_name,
```
dmarcoux, line 36: The line break isn't needed.

```suggestion
    Backend::Api::Sources::Package.rebuild(project_name, package_name)
```

dmarcoux (changes requested):

src/api/spec/models/workflow/step/rebuild_package_step_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'rails_helper'
+
+RSpec.describe Workflow::Step::RebuildPackage, vcr: true do
+  let!(:user) { create(:confirmed_user, :with_home, login: 'Iggy') }
+  let(:token) { create(:workflow_token, user: user) }
```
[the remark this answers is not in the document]
vpereira: I think it's not true. `subject` will be lazily evaluated on the `context` and it will use the `token` defined in the `context`.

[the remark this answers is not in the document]
dmarcoux: With _`token` is global_, I meant that it's defined at the top of the `describe` block and it's unclear why, since only the first example is using this. If you want to know what exactly is `token`, you need to look at the top of the file instead of just looking at the context itself. Anyway, it's not a blocker, but it's not a good practice. I know that we do this in many specs, but it's not right.

dmarcoux (approved):

PR #11636, opened 2021-09-20
[backend] fix typo in get_buildinfo

adrianschroeter:


coolo:
Dup of #11634

adrianschroeter:
k, but I merge this one anyway now, since we need it and the frontend part is still in review, sorry.

PR #11637, opened 2021-09-20
Use SignUpComponent instead of custom sign up form

vpereira:
Following up the feedback https://github.com/openSUSE/open-build-service/pull/11617#issuecomment-920703797

Here we are using the `SignUpComponent` in every sign up form. However both sign up forms weren't looking the same. 

Before change:

Screenshot_2021-09-20_11-42-54

After change:

Screenshot_2021-09-20_11-47-27


dmarcoux:
The `Before` version is definitely easier to click on mobile. We could change that, it's only a bit of CSS with `w-100` on the button itself.

vpereira:
> The `Before` version is definitely easier to click on mobile. We could change that, it's only a bit of CSS with `w-100` on the button itself.

you mean change that in the component?

dmarcoux:
> We have to handle as well the link (maybe centering it).. maybe accommodate the following code in the component:
> 
> ```
> .login-separator.text-center
>   %hr.my-4
>   %span.bg-white.px-3.text-muted or
> .text-center
>   = log_in_link
> ```

This would be nice :+1:

danidoni:
LGTM!

vpereira:
@dmarcoux there was already some improvement, however it still look weird when the sign up form isn't modal

Screenshot_2021-09-28_15-33-13

Screenshot_2021-09-28_15-32-55

dmarcoux:
To have the same position for the `or` text in the `SignUpComponent`, we need to have this CSS: https://github.com/openSUSE/open-build-service/blob/f0f4d2c8cf7984dba0e60f78fe988545de9a78e9/src/api/app/assets/stylesheets/webui/layout.scss#L100-L106

With the current code, it only applies to the Sign Up modal. This is why the `SignUpComponent` renders differently.

danidoni (approved):

dmarcoux (changes requested):
See my previous [comment](https://github.com/openSUSE/open-build-service/pull/11637#issuecomment-930051596).

PR #11638, opened 2021-09-20
Fix non-functional spec

danidoni:
There was a spec with no `it` block, so it did nothing.


rubhanazeem (approved):

rubhanazeem:
It's using some `shared_contexts` and `it` block is available in shared context.

danidoni:
> It's using some `shared_contexts` and `it` block is available in shared context.

Ouch, you're right! I'm gonna close this.

PR #11640, opened 2021-09-20
Remove duplicated validation in branch and link package step

krauselukas:
The super class of the steps already takes care of validating
the presence of the source package and source project names through
the REQUIRED_KEYS constant.

### DIFF
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -39,12 +39,21 @@ class Workflow::Step
 
   def validate_step_instructions
     self.class::REQUIRED_KEYS.each do |required_key|
-      errors.add(:base, "The '#{required_key}' key is missing") unless step_instructions.key?(required_key)
+      unless step_instructions.key?(required_key)
+        errors.add(:base, "The '#{required_key}' key is missing")
+        next
+      end
+
+      errors.add(:base, "The '#{required_key}' key must provide a value") if step_instructions[required_key].blank?
     end
   end
 
   def source_package_name
     step_instructions[:source_package]
+  end
+
+  def source_project_name
+    step_instructions[:source_project]
   end
 
   def target_package_name
--- a/src/api/app/models/workflow/step/branch_package_step.rb
+++ b/src/api/app/models/workflow/step/branch_package_step.rb
@@ -1,18 +1,11 @@
 class Workflow::Step::BranchPackageStep < ::Workflow::Step
   REQUIRED_KEYS = [:source_project, :source_package].freeze
 
-  validates :source_project_name, presence: true
-  validates :source_package_name, presence: true
-
   def call(options = {})
     return unless valid?
 
     workflow_filters = options.fetch(:workflow_filters, {})
     branch_package(workflow_filters)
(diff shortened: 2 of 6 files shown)


src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -1,20 +1,13 @@
 class Workflow::Step::BranchPackageStep < ::Workflow::Step
   REQUIRED_KEYS = [:source_project, :source_package].freeze
 
-  validates :source_project_name, presence: true
-  validates :source_package_name, presence: true
```
dmarcoux, line 5: Those validations need to stay.

dmarcoux (changes requested):

dmarcoux (dismissed):

dmarcoux:
We could also remove this:

https://github.com/openSUSE/open-build-service/blob/b74e17fbc4e7ca68f1a1be317bb71f6112ec4b4e/src/api/app/models/workflow/step/configure_repositories.rb#L5

src/api/app/models/workflow/step.rb
```diff
@@ -37,16 +37,26 @@ def create_or_update_subscriptions(package, workflow_filters)
 
   protected
 
-  def validate_step_instructions
+  def validate_step_keys
```
dmarcoux, line 40: I would keep the validations in a single method to not loop twice over the required keys. It's not much overheard, but the code could be like this:

```ruby
  def validate_step_instructions
    self.class::REQUIRED_KEYS.each do |required_key|
      unless step_instructions.key?(required_key)
        errors.add(:base, "The '#{required_key}' key is missing")
        next
      end

      errors.add(:base, "The '#{required_key}' key must provide a value") if step_instructions[required_key].blank?
    end
  end
```

src/api/app/models/workflow/step.rb
```diff
@@ -37,16 +37,26 @@ def create_or_update_subscriptions(package, workflow_filters)
...
     self.class::REQUIRED_KEYS.each do |required_key|
       errors.add(:base, "The '#{required_key}' key is missing") unless step_instructions.key?(required_key)
     end
   end
 
+  def validate_step_values
+    self.class::REQUIRED_KEYS.each do |required_key|
+      errors.add(:base, "The '#{required_key}' key must provide a value") if step_instructions[required_key].blank?
```
dmarcoux, line 48: If `required_key` isn't present in `step_instructions`, `step_instructions[required_key].blank?` will cause the following exception:

```
NoMethodError (undefined method `blank?' for nil:NilClass)
```

src/api/app/models/workflow/step.rb
krauselukas: :+1: done

dmarcoux:
@krauselukas I pushed to fix the failing spec. The test suite should be passing now.

danidoni (approved):

dmarcoux:
The CI check `coverage/coveralls` seems to be stuck... I'm merging

krauselukas:
@dmarcoux merci

PR #11643, opened 2021-09-20
Implement event filter for workflows

dmarcoux:
See the [developer documentation](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration#event-filter) for details on the event filter.


dmarcoux:
[Implement event filter](https://trello.com/c/6jxTorRF/1752-implement-event-filter)

src/api/app/models/workflow.rb
```diff
@@ -82,6 +81,14 @@ def supported_filters
     @supported_filters ||= workflow_instructions[:filters]&.select { |key, _value| SUPPORTED_FILTERS.include?(key.to_sym) }
   end
 
+  def event_matches_event_filter?
+    return true unless supported_filters.key?(:event) # TODO: This will break if supported_filters is nil
```
danidoni, line 85: Make `supported_filters` return `{}` if it does not work.

src/api/spec/models/workflow_spec.rb
```diff
@@ -1,20 +1,18 @@
 require 'rails_helper'
 
-RSpec.describe Workflow, type: :model, vcr: true do
```
dmarcoux, line 3: There are no cassettes for the specs of this model. This can be safely removed since the backend isn't involved.

src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
```diff
@@ -1,7 +1,34 @@
 require 'rails_helper'
 
 RSpec.describe Workflows::YAMLToWorkflowsService, type: :service do
-  include_context 'a scm payload hash'
```
dmarcoux, line 4: We get rid of this shared context since it's only used in this file now with the changes in `src/api/spec/models/workflow_spec.rb`.

src/api/app/models/workflow.rb
```diff
@@ -47,27 +50,27 @@ def steps
...
 
-    SUPPORTED_FILTERS.each do |filter|
-      SUPPORTED_FILTER_TYPES.each do |filter_type|
-        # The filter type might not be present... so in that case, we go to the next
-        next unless (filter_values = supported_filters.dig(filter, filter_type))
+    @filters ||= SUPPORTED_FILTERS.each_with_object({}) do |filter, filters_hash|
+      if filter == :event
+        filters_hash[filter] = supported_filters[filter]
+      else
+        SUPPORTED_FILTER_VALUES.each do |filter_value|
```
hennevogel, line 59: I think the same (precedence of `only` over `ignore`) we do already in the steps. 

https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/workflow/step.rb#L108-L110

eduardoj: From line 12: `SUPPORTED_FILTER_VALUES = [:only, :ignore].freeze` should be enough.

I'm missing something. What are you suggesting?

dmarcoux: This loop and the link that @hennevogel shared implement the same logic. We don't need this logic duplication. Let's remove it in the `Workflow` model.

src/api/app/models/workflow.rb
```diff
@@ -47,27 +50,27 @@ def steps
   end
 
   def filters
-    filters = {}
-
-    return filters if supported_filters.blank?
+    return {} if filters_configuration.blank?
```
dmarcoux, line 53: I don't like this change from `supported_filters` to `filters_configuration`. We use the term `supported` everywhere in the workflow feature. I agree that the code in this method is now a bit confusing due to the `if/else`, but renaming `supported_filters` to `filters_configuration` doesn't improve the code.

hennevogel: Okay, I have no super string feelings about this.

rubhanazeem (approved):

PR #11645, opened 2021-09-21
Support target_project key for branch and link package step

krauselukas:
DO NOT MERGE: PR contains breaking changes, we need to communicate this before merging and deploying! But ready for reviews :)


dmarcoux:
[Adapt workflow steps for push events on branches](https://trello.com/c/KUFzNTDw/1749-adapt-workflow-steps-for-push-events-on-branches)

src/api/app/models/workflow/step.rb
```diff
@@ -15,7 +15,9 @@ def call(_options)
   end
 
   def target_project_name
-    "home:#{@token.user.login}:#{source_project_name}:PR-#{scm_webhook.payload[:pr_number]}"
+    return step_instructions[:target_project] if scm_webhook.push_event?
+
+    "#{step_instructions[:target_project]}:#{source_project_name}:PR-#{scm_webhook.payload[:pr_number]}"
```
hennevogel, line 20: While we are at it we wanted to get off `source_project_name` here right? So more like

```ruby
"#{step_instructions[:target_project]}:#{scm_webhook.payload[:target_repository_full_name].split('/').join(':')}:PR-#{scm_webhook.payload[:pr_number]}"
```

krauselukas: Alright, I overlooked that the naming scheme for pull request also changes, going to adapt it. Looking at the wiki now the naming looks a bit different `OBS:Server:Unstable:CI:PR-$MY_PR_NUMBER/obs-server`, but I like the way you described. So I would go for it

hennevogel: I changed the wiki in the meanwhile ;-)

krauselukas: @hennevogel Alright, so changed the code and test to use the new naming schema for PR's

krauselukas:
Put back into WIP state because I have to check if permissions on the project are handled correctly

src/api/app/models/workflow/step.rb
```diff
@@ -15,7 +15,15 @@ def call(_options)
   end
 
   def target_project_name
-    "home:#{@token.user.login}:#{source_project_name}:PR-#{scm_webhook.payload[:pr_number]}"
+    return step_instructions[:target_project] if scm_webhook.push_event?
+
+    pr_subproject_name = if scm_webhook.payload[:scm] == 'github'
```
vpereira, line 20: isn't too optimistic to assume that this whole chain of commands will never fail?

krauselukas: why? `target_repository_full_name` and `path_with_namespace` are always part of the payload, and there will always be one and only one `/`.

dmarcoux: > there will always be one and only one `/`.

True :+1:. Slashes are replaced by dashes (`-`) in repository names.

src/api/app/models/workflow/step.rb
```diff
@@ -15,7 +15,15 @@ def call(_options)
...
+                           scm_webhook.payload[:target_repository_full_name].split('/').first(2).join(':')
```
dmarcoux, line 21: There is no need to split if you join the string again. Replacing the slashes for colon is much more straightforward.

```suggestion
                           scm_webhook.payload[:target_repository_full_name].gsub('/', ':')
```

src/api/app/models/workflow/step.rb
```diff
@@ -15,7 +15,15 @@ def call(_options)
...
+                         else
+                           scm_webhook.payload[:path_with_namespace].split('/').first(2).join(':')
```
dmarcoux, line 23: Same as above.

src/api/app/models/workflow/step.rb
krauselukas: done

src/api/app/models/workflow/step.rb
krauselukas: done

rubhanazeem (approved):
@krauselukas the changes look good to me. :+1:

krauselukas:
Going to merge this now, and communicate the changes before the next deployment

PR #11647, opened 2021-09-21
[dist] obsstoragesetup: do not prompt before wipe

luizluca:
According to lvcreate man, --wipesignature, "there is a prompt for
each signature detected to confirm its wiping (unless --yes is used to
override confirmations.)". However, obsstoragesetup is executed as a
service and it will abort if an interactive prompt is requested.

The result is that workers are started without a proper storage and
they will start to use rootfs, that might not be prepared to handle
the amount of worker data.

Use --yes for lvcreate to avoid interactve prompts.

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

### DIFF
--- a/dist/obsstoragesetup
+++ b/dist/obsstoragesetup
@@ -263,7 +263,7 @@ case "$1" in
 						I=$(( $I + 1 ))
 				
 						lverr=$(mktemp)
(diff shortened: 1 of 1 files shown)


luizluca:
It should be backported to 2.10

Also, it would be nice to fail obsworker service if obsstoragesetup fails.

PR #11652, opened 2021-09-22
Refactor Webui::WebuiController#check_user

dmarcoux:
The logic of WebuiControllerService::UserChecker isn't leaking out of the service anymore. This is to reuse the service.

### DIFF
--- a/src/api/spec/services/webui_controller_service/user_checker_spec.rb
+++ b/src/api/spec/services/webui_controller_service/user_checker_spec.rb
@@ -2,67 +2,106 @@ require 'rails_helper'
 require 'ostruct'
 
 RSpec.describe ::WebuiControllerService::UserChecker do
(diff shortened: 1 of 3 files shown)


src/api/spec/services/webui_controller_service/user_checker_spec.rb
```diff
@@ -2,67 +2,106 @@
...
+
+        before do
+          allow(User).to receive(:find_nobody!)
+        end
+
+        it { is_expected.to be(true) }
+
+        it 'sets the current user to the anonymous user' do
+          subject
+          expect(User).to have_received(:find_nobody!)
```
dmarcoux, line 47: Testing with `User.session` doesn't work for the `nobody` user. So this is why I'm doing this.

danidoni (approved):

PR #11658, opened 2021-09-24
Support push event in branch and link package step

krauselukas:
See the developer documentation for the [branch_package](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration#branch-a-package) and [link_package](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration#link-a-package-to-a-project) steps.

### DIFF
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -65,9 +65,14 @@ class Workflow::Step
   end
 
   def target_package_name
-    return step_instructions[:target_package] if step_instructions[:target_package].present?
+    package_name = step_instructions[:target_package] || source_package_name
 
-    source_package_name
+    case
+    when scm_webhook.pull_request_event?
+      package_name
+    when scm_webhook.push_event?
+      "#{package_name}-#{scm_webhook.payload[:commit_sha]}"
+    end
   end
 
   private
--- a/src/api/app/services/trigger_controller_service/scm_extractor.rb
+++ b/src/api/app/services/trigger_controller_service/scm_extractor.rb
@@ -43,7 +43,8 @@ module TriggerControllerService
                          # We need this for Workflow::Step#branch_request_content_github
                          source_repository_full_name: @payload.dig(:repository, :full_name),
                          # We need this for SCMStatusReporter#call
-                         target_repository_full_name: @payload.dig(:repository, :full_name) })
+                         target_repository_full_name: @payload.dig(:repository, :full_name),
+                         ref: @payload[:ref] })
       end
       payload
     end
@@ -74,7 +75,8 @@ module TriggerControllerService
                          target_branch: @payload[:ref].sub('refs/heads/', ''),
                          # We need this for Workflows::YAMLDownloader#download_url
                          path_with_namespace: @payload.dig(:project, :path_with_namespace),
-                         project_id: @payload[:project_id] })
+                         project_id: @payload[:project_id],
+                         ref: @payload[:ref] })
       end
       payload
     end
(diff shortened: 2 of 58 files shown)


src/api/app/models/workflow/step.rb
```diff
@@ -65,6 +65,12 @@ def source_project_name
   end
 
   def target_package_name
+    if scm_webhook.push_event?
```
dmarcoux, line 68: We're now checking twice if `step_instructions[:target_package]` is present. We should do this once, then add the commit SHA or not depending on the webhook event. The code would be simpler. Something like this:

```ruby
def target_package_name
  package_name = step_instructions[:target_package] || source_package_name

  case
  when scm_webhook.pull_request_event?
    package_name
  when scm_webhook.push_event?
    "#{package_name}-#{scm_webhook.payload[:commit_sha]}"
  end
end
```

krauselukas: @dmarcoux Done :+1:

krauselukas:
@danidoni and me tested this manully and it worked, workflows running on the push event and status gets reported to the SCM

src/api/app/services/trigger_controller_service/scm_extractor.rb
```diff
@@ -43,7 +43,8 @@ def github_extractor_payload
                          # We need this for Workflow::Step#branch_request_content_github
                          source_repository_full_name: @payload.dig(:repository, :full_name),
                          # We need this for SCMStatusReporter#call
-                         target_repository_full_name: @payload.dig(:repository, :full_name) })
+                         target_repository_full_name: @payload.dig(:repository, :full_name),
+                         ref: @payload.dig(:ref) })
```
krauselukas, line 47: We had to add the ref field to the payload, since its used in the validator to check if its a push event, this got overlooked in a previous PR

krauselukas: Its needed here https://github.com/openSUSE/open-build-service/blob/934aa35442f7789ea1ce761f07301f49b36b68f6/src/api/app/validators/scm_webhook_event_validator.rb#L52

dmarcoux (approved):

eduardoj (commented):
RSpec tests fails for me locally with:

```
Failures:

  1) Workflow::Step::LinkPackageStep#call when the SCM is GitLab for a new MR event behaves like target package already exists 
     Failure/Error: let!(:linked_project) { create(:project, name: final_target_project_name, maintainer: user) }
     
     NameError:
       undefined local variable or method `final_target_project_name' for #<RSpec::ExampleGroups::WorkflowStepLinkPackageStep::Call::WhenTheSCMIsGitLab::ForANewMREvent::BehavesLikeTargetPack
ageAlreadyExists:0x00005646c18e20e8>
     Shared Example Group: "target package already exists" called from ./spec/models/workflow/step/link_package_step_spec.rb:367
     # ./spec/models/workflow/step/link_package_step_spec.rb:52:in `block (3 levels) in <top (required)>'

Finished in 28.46 seconds (files took 8.51 seconds to load)
90 examples, 1 failure

Failed examples:

rspec './spec/models/workflow/step/link_package_step_spec.rb[1:1:2:3:4:1]' # Workflow::Step::LinkPackageStep#call when the SCM is GitLab for a new MR event behaves like target package alread
y exists 

Randomized with seed 40952
```

eduardoj (approved):

eduardoj (approved):

bluca:
@krauselukas is it possible to filter on pushed tags, other than push to branches? Use case is to run the workflow when a tag is pushed. If not, I can file an RFE.

krauselukas:
hey @bluca at the moment we only take push events to branches into account, but we have something on the feature map to take tags into account. Would be great if you could file an RFE, this way we can track that there is an interest in supporting this!

bluca:
Ok, opened https://github.com/openSUSE/open-build-service/issues/11665

PR #11675, opened 2021-09-29
 Mark all read/unread for notifications

rubhanazeem:


rubhanazeem:
[Mark all as read](https://trello.com/c/0I3J2VkI/1757-mark-all-as-read)

src/api/app/components/notification_action_bar_component.rb
```diff
@@ -0,0 +1,29 @@
...
+    @type = type
+    @update_path = update_path + '&update_all=true'
+  end
+
+  def button_text(all: false)
+    if all
+      type == 'read' ? "Mark all as 'Unread'" : "Mark all as 'Read'"
+    else
+      type == 'read' ? "Mark as 'Unread'" : "Mark as 'Read'"
+    end
```
danidoni, line 18: I find this combination of conditions and ternaries difficult to read. Maybe rewriting it as guard clauses it's easier to read?

rubhanazeem: Thanks for the suggestion. It's updated :slightly_smiling_face:

krauselukas (commented):
Generally it looks good, I'm just thinking if it might be better to align the new button on the right side, I could imagine that people will click by accident on the wrong one when selecting notifications. A bit room might help, what do you think @rubhanazeem

rubhanazeem:
@krauselukas yeah good idea. It's updated.

src/api/app/components/notification_action_bar_component.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class NotificationActionBarComponent < ApplicationComponent
```
dmarcoux, line 3: `NotificationActionBar` is quite a generic name. We could name this component `MarkAsReadButton`. Splitting up this component in multiple components like `MarkAsUnreadButton` and `MarkAllAsReadButton` would be maybe too much abstraction. So having a single component is good enough, at least for now.

dmarcoux (changes requested):
Good work! Only minor changes are needed.

src/api/app/components/notification_action_bar_component.rb
rubhanazeem: `NotificationActionBar` has three things. `Select All`, `Mark as read`, and `Mark all as read`. Then the same actions for unread operation. Naming it `MarkAsReadButton` would be a little confusing because we are leaving out the unread operations. We only have one notifications page and one notification action bar. `NotificationActionBar` still seems a better choice. What do you think @dmarcoux ? Should we still rename it?

src/api/app/components/notification_action_bar_component.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def button_text(all: false)
+    text = type == 'read' ? 'Unread' : 'Read'
+    return "Mark all as '#{text}'" if all
+
+    "Mark as '#{text}'"
+  end
+
+  def disable_with_content(all: false)
```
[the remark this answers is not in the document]
rubhanazeem: Done.

src/api/app/components/notification_action_bar_component.rb
dmarcoux: Oh, true I totally overlooked that this component also has the `Select All` checkbox.

Renaming this component isn't needed then.

dmarcoux (approved):

krauselukas (approved):

src/api/app/components/notification_action_bar_component.html.haml
```diff
@@ -5,9 +5,8 @@
         .custom-control.custom-checkbox
           = check_box_tag('select-all-notifications', 1, false, class: 'custom-control-input')
           %label.custom-control-label.align-middle.mr-4#select-all-label{ for: 'select-all-notifications' } Select All
-          :ruby
-            button_text = type == 'read' ? "Mark as 'Unread'" : "Mark as 'Read'"
-            disable_with_content = button_text + tag.i(class: 'fas fa-spinner fa-spin ml-2')
           = button_tag(type: 'submit', class: 'btn btn-sm btn-outline-success px-3', id: 'done-button', disabled: 'disabled',
                        'data-disable-with': disable_with_content) do
             = button_text
+          = link_to(button_text(all: true), @update_path, method: :put, remote: true,
```
hennevogel, line 11: Some design feedback. It's never such a good idea to have two call-to-actions buttons on one page.

Also the difference is very subtle (basically one word) and for less than 300 Notifications the buttons do exactly the same.

Screenshot from 2021-10-05 12-12-34

rubhanazeem: @hennevogel do we need to update anything? Or do you have any suggestions that could make it better?

hennevogel: Some ideas:

- Use different colors for the buttons or don't use colors at all. The actions are equally "important".
- Display `Mark all as read` only if there are more Notifications than currently on the page
- Change `Mark as read` to `Mark selected as read`

rubhanazeem: @hennevogel it's updated. 

Screenshot_2021-10-05 Screenshot

PR #11682, opened 2021-09-30
Update bundler version 2.2.27

hennevogel:
The [package in O:S:U](https://build.opensuse.org/package/show/OBS:Server:Unstable/rubygem-bundler) is updated, the container is on it's way.


krauselukas (approved):

hennevogel:
CI will never finish as it's based on a container that needs the old bundler version: 🐔 vs. 🥚

PR #11687, opened 2021-10-01
Click and reveal long project and package combinations

danidoni:
## Before
image

## After
image

### DIFF
--- a/src/api/app/views/webui/users/notifications/_notifications_list.html.haml
+++ b/src/api/app/views/webui/users/notifications/_notifications_list.html.haml
@@ -29,7 +29,7 @@
                     = check_box_tag('notification_ids[]', notification.id, false,
                                     id: "notification_ids_#{notification.id}", class: 'custom-control-input')
                     = label_tag("notification_ids_#{notification.id}", '', class: 'custom-control-label')
-                .col
+                .col-10
                   .row
                     .col
                       - if notification.notifiable_type == 'BsRequest'
@@ -56,7 +56,7 @@
                     .col-auto.pr-0
                       = render partial: 'notification_avatars', locals: { avatar_objects: notification.avatar_objects }
                     .col-auto.pl-xs-2
-                      %span.text-word-break-all= action_description(notification)
+                      .smart-overflow= action_description(notification)
                   .row.d-none.d-md-block.pl-4
                     .col
                       %p.mt-3.mb-0= notification.excerpt
@@ -64,3 +64,5 @@
 
 - content_for :ready_function do
   handleNotificationCheckboxSelection();
+  applySmartOverflow();
+  window.onresize = applySmartOverflow;
--- a/src/api/app/assets/javascripts/webui/application.js
+++ b/src/api/app/assets/javascripts/webui/application.js
@@ -55,3 +55,4 @@
 //= require webui/notification.js
 //= require webui/navigation.js
 //= require webui/user_profile.js
+//= require webui/long_text.js
--- a/src/api/app/assets/javascripts/webui/long_text.js
+++ b/src/api/app/assets/javascripts/webui/long_text.js
@@ -1,0 +1,17 @@
+/* exported applySmartOverflow */
+function applySmartOverflow() {
+    $(".smart-overflow").each(function(_, el) {
+        $(el).find(".ellipsis-link").remove();
+
+        if (el.offsetWidth < el.scrollWidth) {
+            var link = document.createElement('a');
+            link.href = '#'; link.className = 'ellipsis-link';
+
+            link.addEventListener('click', function (e) {
+                e.preventDefault();
+                el.classList.remove('smart-overflow');
+            });
+            el.appendChild(link);
+        }
+    });
+}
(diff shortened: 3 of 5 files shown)


danidoni:
[Fix notification content with long project/package name combinations](https://trello.com/c/LqMYQLdh/1758-fix-notification-content-with-long-project-package-name-combinations)

krauselukas (commented):
Generally it looks good, the only issue I see that there is no indicator that this is scroll able, I checked on my phone and if I wouldn't know I would think the text is just cut. Maybe showing little arrows or similar?

danidoni:
> @danidoni Generally it looks good, the only issue I see that there is no indicator that this is scroll able, I checked on my phone and if I wouldn't know I would think the text is just cut. Maybe showing little arrows or similar?

Yeah, I thought about the same but I'm not sure now to do it...

danidoni:
> > @danidoni Generally it looks good, the only issue I see that there is no indicator that this is scroll able, I checked on my phone and if I wouldn't know I would think the text is just cut. Maybe showing little arrows or similar?
> 
> Yeah, I thought about the same but I'm not sure now to do it... Do you have any ideas on how to do it?

@krauselukas suggested to add a shadow over the hidden content to draw attention to the fact that the content can be scrollable horizontally. That change is now present in the PR.

dmarcoux (changes requested):
Using shadows to hint at scrollable content is somewhat weird. The shadows are still visible when the content is under the top navigation, see this screenshot:

artifact

A better solution would be to shorten the text with an ellipsis, then hint at clicking on the text to fully display it.

danidoni:
> Using shadows to hint at scrollable content is somewhat weird. The shadows are still visible when the content is under the top navigation, see this screenshot:
> 
> artifact
> 
> A better solution would be to shorten the text with an ellipsis, then hint at clicking on the text to fully display it.

That is unfortunately not a perfect solution though, because, for example, in the "Latest updates" section in the main page, the "click-to-reveal" would conflict with the link.

hennevogel:
> "click-to-reveal"

Have you tried accessing this page with your *actual* mobile?

danidoni:
> > "click-to-reveal"
> 
> Have you tried accessing this page with your _actual_ mobile?

Yes, I did.

danidoni:
I've switched the implementation to the "click-and-reveal". It actually makes sense here and it looks better.

danidoni:
In the end I've decided to use an ellipsis to signal that the project and package name is longer than the viewport with. No content reveal is implemented in this PR, though.

hennevogel:
> No content reveal is implemented in this PR, though.

This makes little sense right? As you don't know what the notification is about anymore...

danidoni:
> > No content reveal is implemented in this PR, though.
> 
> This makes little sense right? As you don't know what the notification is about anymore...

Okay, let me give it another try to the click and reveal approach...

danidoni:
I think I got it this time. Please review again @hennevogel @dmarcoux @krauselukas !

danidoni:
@dmarcoux detected that the code does not react to viewport changes, so when switching from landscape to portrait, this implementation does not work. I'll see what I can do.

danidoni:
> @dmarcoux detected that the code does not react to viewport changes, so when switching from landscape to portrait, this implementation does not work. I'll see what I can do.

Fixed, now.

src/api/app/views/layouts/webui/webui.html.haml
```diff
@@ -31,7 +31,7 @@
     = javascript_tag do
       = yield :head_javascript
       var _paq = _paq || [];
-      $(function() { #{yield :ready_function} });
+      $(function() { #{yield :ready_function} window.onresize = applySmartOverflow; });
```
krauselukas, line 34: @danidoni I would only add this to the views where it is required

danidoni: Fixed

krauselukas (approved):

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
```diff
@@ -29,7 +29,7 @@
                     = check_box_tag('notification_ids[]', notification.id, false,
                                     id: "notification_ids_#{notification.id}", class: 'custom-control-input')
                     = label_tag("notification_ids_#{notification.id}", '', class: 'custom-control-label')
-                .col
+                .col-10
```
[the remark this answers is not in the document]
danidoni: Before, the full-length text had the option to wrap onto the next line, reducing the width of the container, and allowing the columns to distribute the available space. Now, the text is on a single-line and it tries to fill as much space as needed giving no space for the rest of the columns. That `.col-10` ensures the first column get at least some space.

dmarcoux:
I gave this a try on my mobile and it's quite nice, although the surface to tap is a bit small with the ellipsis being only 3 dots. I wonder if we could make this even better by tapping on the project/package names to reveal the complete project/package names. What do you think? This would be much easier on mobile. It's not always possible, but here, it is since the project/package names aren't in a link.

dmarcoux (approved):

PR #11698, opened 2021-10-05
Fetch only the group IDs in NotificationsFinder

dmarcoux:
This is more efficient than fetching all group attributes to only take IDs.

### DIFF
--- a/src/api/app/queries/notifications_finder.rb
+++ b/src/api/app/queries/notifications_finder.rb
@@ -21,7 +21,7 @@ class NotificationsFinder
 
   def for_subscribed_user(user = User.session)
     @relation.where("(subscriber_type = 'User' AND subscriber_id = ?) OR (subscriber_type = 'Group' AND subscriber_id IN (?))",
-                    user, user.groups.map(&:id))
+                    user, user.groups.pluck(:id))
   end
 
   def for_incoming_requests(user = User.session)


hennevogel (approved):

src/api/app/queries/notifications_finder.rb
```diff
@@ -21,7 +21,7 @@ def without_notifiable
 
   def for_subscribed_user(user = User.session)
     @relation.where("(subscriber_type = 'User' AND subscriber_id = ?) OR (subscriber_type = 'Group' AND subscriber_id IN (?))",
-                    user, user.groups.map(&:id))
+                    user, user.groups.pluck(:id))
```
coolo, line 24: Note that `.select(:id)` is often even more efficient as you're skipping the roundtrip between rails and the database.

dmarcoux: The groups of the user are not loaded, there will be a query to the database no matter what. With `select(:id)`, we end up with:

```
  Group Load (0.6ms)  SELECT DISTINCT `groups`.`id` FROM `groups` INNER JOIN `groups_users` ON `groups`.`id` = `groups_users`.`group_id` WHERE `groups_users`.`user_id` = 1 ORDER BY `groups`.`title` ASC
=> [#<Group:0x000055e686b69b68 id: 1>, #<Group:0x000055e686aaeb10 id: 2>]
```

In comparison, with `pluck(:id)`, we end up with:
```
   (0.4ms)  SELECT DISTINCT `groups`.`id` FROM `groups` INNER JOIN `groups_users` ON `groups`.`id` = `groups_users`.`group_id` WHERE `groups_users`.`user_id` = 1 ORDER BY `groups`.`title` ASC
=> [1, 2]
```

`pluck` is better here, there is no need to run `map(&:id)` on the result, unlike `select(:id)`.

coolo: well, you put the result back into a relation and there it does make a difference.

coolo:
```
> Relationship.where("group_id in (?)", user.groups.pluck(:id)).first
   (0.7ms)  SELECT DISTINCT `groups`.`id` FROM `groups` INNER JOIN `groups_users` ON `groups`.`id` = `groups_users`.`group_id` WHERE `groups_users`.`user_id` = 138 ORDER BY `groups`.`title` ASC
  Relationship Load (0.7ms)  SELECT `relationships`.* FROM `relationships` WHERE (group_id in (101,78)) ORDER BY `relationships`.`id` ASC LIMIT 1
```

vs

```
> Relationship.where("group_id in (?)", user.groups.select(:id)).first
  Relationship Load (1.5ms)  SELECT `relationships`.* FROM `relationships` WHERE (group_id in (SELECT DISTINCT `groups`.`id` FROM `groups` INNER JOIN `groups_users` ON `groups`.`id` = `groups_users`.`group_id` WHERE `groups_users`.`user_id` = 138 ORDER BY `groups`.`title` ASC))"
```

Both are around 1.5ms in total (because it's just 2 groups anyway), but in between the 0.7ms is reading database output into ruby, digesting it, putting it back into database, reading the output again, ... the 0.7ms only measures the actual query time.

And my comment is not an objection to the change. But it's important IMO to keep this in mind as the number of items returned grows - because your above comment clearly compares apples with oranges.

dmarcoux: > because your above comment clearly compares apples with oranges.

Ah @coolo, as always so tactful... You bring good points, it's your wording or tone which is often the issue.

Anyway, next time, be clearer in your comments. Now that you wrote `well, you put the result back into a relation and there it does make a difference.`, I understand what you meant in the first place.

PR #11688, opened 2021-10-04
Fix project long name in comments header

danidoni:

PR #11702, opened 2021-10-07
Fix long project names in the comment section of project details page

danidoni:
In this case it does not implement the "click-to-reveal" from #11687 because we have the full name of the project at the top, so we are not loosing any information here.

## Before
Screenshot_20211007_093619

## After
Screenshot_20211007_093537


saraycp:
With the same reasoning we had [here](https://github.com/openSUSE/open-build-service/pull/11688#issuecomment-946806531), we can also close this PR in favour of #11753 11753

PR #11711, opened 2021-10-08
Simplify method name in FindForEvent model

eduardoj:


hennevogel:
I guess this is missing some changes?

PR #11724, opened 2021-10-13
Keep columns sort and shown entries per page in datatable state 

### LINKED ISSUE #7535: Saving the state of Datatables can mislead users
_Originally posted by @coolo in https://github.com/openSUSE/open-build-service/pull/7427#issuecomment-489443630_

It can lead to strange behaviour - I left the page this morning when I had > 100 requests and now that I'm down to 40, it says there is no data. Took me a moment to realize I'm on page 5 of 4.

screencapture-build-opensuse-org-user-tasks-2019-05-05-18_49_28

dmarcoux:
The selected page and the search string aren't saved in the datatable state. This was often confusing users.

Fixes #7535

### DIFF
--- a/src/api/app/assets/javascripts/webui/datatables.js
+++ b/src/api/app/assets/javascripts/webui/datatables.js
@@ -17,7 +17,14 @@ var DEFAULT_DT_PARAMS = {
   pageLength: 25,
   stateSave: true,
   pagingType: 'full',
-  stateDuration: 0 // forever
+  stateDuration: 0, // forever
+  // Save the state of the columns sort and the number of shown entries per page
+  stateSaveParams: function (_settings, data) {
+    // Do not keep the selected page in the datatable state
+    data.start = 0;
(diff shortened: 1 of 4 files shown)


dmarcoux:
> I guess the state saving was meant to keep the "show 10 or 20 packages" setting

It's true that keeping this setting saved forever would be quite useful, since I doubt users are often changing this. As for storing the pagination and the search string, I agree with you @coolo that this isn't useful. Sadly, it's not possible to keep the state of a certain setting forever while only keeping the state of other settings only for the current session.

I will look into keeping the state of the "show 10 or 20 packages"  setting forever, while not keeping the state of other settings. This seems to be the best solution.

dmarcoux:
@hennevogel: We could for sure do something similar, but I'm not going to do this now. What I described in my previous comment is a good enough solution for now.

dmarcoux:
Thank you everyone for the quick feedback :slightly_smiling_face:

PR #11727, opened 2021-10-13
Align the configuration of CodeClimate checks with those of RuboCop cops

dmarcoux:
Using the CodeClimate plugin for RuboCop is not possible due to our somewhat quirky RuboCop setup with our 2 configurations in the Git repository's root and `src/api/` directories.

This solution is good enough for now to allow us to track the code quality changes over time with the CodeClimate UI.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,4 +1,37 @@
 version: "2"
+# Align the configuration of CodeClimate checks with those of RuboCop cops
+# Documentation: https://docs.codeclimate.com/docs/maintainability#section-checks
+checks:
+  # Same as cop Metrics/ParameterLists from RuboCop
+  # https://docs.rubocop.org/rubocop/cops_metrics.html#metricsparameterlists
+  argument-count:
+    config:
+      threshold: 5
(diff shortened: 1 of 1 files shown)


dmarcoux:
> Did we check if the thresholds are aligned?

I was thinking about doing this in a follow-up PR, but there isn't a reason to split this up in two PRs. I'll push changes for this.

vpereira:
> Yes, I don't think it would make sense to break it in another PR, since the next person committing code could be affected by that

Screenshot_2021-10-14_10-22-30

We should adjust the thresholds - and not try to fix the code - To make this numbers drop to an acceptable number i.e 0 :)

dmarcoux:
Please have a look again, I updated the PR's title and description to apply what we discussed today in our standup.

dmarcoux:
There are still 80 issues reported by CodeClimate since the checks configuration is stricter now that it's aligned with the settings of RuboCop cops. There are also JS files which are considered issues when they have over 100 lines, I think refactoring this is definitely something we'll have to tackle one day. For now, it's good to track those issues in CodeClimate.

PR #11728, opened 2021-10-13
Turn NotificationFilterPresenter into a view component

### LINKED ISSUE #11557: Retire NotificationsFilterPresenter

>

krauselukas:
Presenters and view components are really similar, but the template code (Haml) is testable in view components and not in presenters. Having design patterns which are too similar is confusing. This is why we're retiring presenters in favor of view components.

Closes #11557


krauselukas:
[Retire presenters](https://trello.com/c/PEfGTPrJ/1769-retire-presenters)

dmarcoux:
I'll take over from @krauselukas since he's away for a while.

PR #11750, opened 2021-10-19
Remove deprecated `/statistics/rating` and `/statistics/highest_rated` API endpoints

eduardoj:
As these endpoints are no longer used, we can also remove the `ratings` database table, and the `Ratings` model.

These endpoints were already deprecated in OBS 2.10. See: https://github.com/openSUSE/open-build-service/blob/1fca4d98fc9352838983934bf56fb0d00a119c33/docs/ReleaseNotes-2.10#L140-L144


eduardoj:
> A few more lines need to be removed.

True. All removed. Thanks! Please have another look.

PR #11753, opened 2021-10-19
Simplify comments' section header

saraycp:
Replace "Comments for X" with simply "Comments" instead of finding a way to shorten the long names.

The comments are displayed in the context of a certain item (project, package) so there is no need to repeat the item's name, which is clear in other parts like the breadcrumbs or the title.

In comparison, we don't display something like `Build results for package-name` even if we have a `Build results` section in the different items pages.

Supersedes: #11688

**Before:**

Screenshot_2021-10-19 Lorem_ipsum_dolor_sit_amet consectetur_adipiscing_elit Donec_at_luctus_arcu sit_amet_lobortis_lacus

**After:**

Screenshot_2021-10-19 Lorem_ipsum_dolor_sit_amet consectetur_adipiscing_elit Donec_at_luctus_arcu sit_amet_lobortis_lacus(1)

### DIFF
--- a/src/api/app/views/webui/package/show.html.haml
+++ b/src/api/app/views/webui/package/show.html.haml
@@ -68,8 +68,7 @@
   .comments
     .card#comments-list
       %h5.card-header.text-word-break-all
-        Comments for
-        = @package.name
+        Comments
         %span.badge.badge-primary{ id: "comment-counter-package-#{@package.id}" }
           = @comments.length
       .card-body#comments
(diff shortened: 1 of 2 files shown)


saraycp:
We also have comments on the request page but there it's important to keep the request's number they are related to since we can have many comments tabs. They are comments for the current and the superseded requests. Example:

Screenshot_2021-10-20 Request 10 (new)

PR #11759, opened 2021-10-20
Minor improvements for staging projects

dmarcoux:
Some minor improvements for staging projects after looking into [src/api/app/views/staging/staging_projects/show.xml.builder](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/views/staging/staging_projects/show.xml.builder).


dmarcoux:
I had to change the code to fix a bug.

PR #11764, opened 2021-10-21
Enable docker layer caching

vpereira:
Based on https://circleci.com/docs/2.0/docker-layer-caching/ we could enable layer caching.

The documentation says that one could just profit from that changes, when build the containers locally, but as saw here https://app.circleci.com/pipelines/github/vpereira/open-build-service/775/workflows/e0f6685d-9755-498c-bb47-d7622c13e7cc/jobs/14209, it caches as well the downloaded layers.

extracted from #11758

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,6 +81,8 @@ jobs:
       - <<: *frontend_base
     steps:
       - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
       - run:
           name: Run scripts linters
           command: make -C dist scripts_linters


coolo:
Is every fork of the repo automatically on the performance plan then?

vpereira:
> With this change, we would then save a bit more than 1 minute in the total CI duration? Cool!

sometimes more. Downloading over and over again the same images means that the CI time is bound to the network time. Caching it locally would avoid much fluctuation on the pipeline. Some experiments could be done to decide in exactly which point we should cache the docker layers (after download/before checkout/after dependencies are installed, etc). Such experiments however could be done later.

PR #11781, opened 2021-10-25
WIP: New feature: check upgrade

pandom79:


pandom79:
Update:
First of all, please ignore the "code".....I'm only testing to understand!
I want update you about some points:

I am able to contact the backend to execute the script.
I also can generate the same files of the obs service,
they are present on the file system but I can't see them on the package source file list.
Probably, I'm wronging something or it always requires a process of the new revision generation which is not good for our goal.

I created a new web service which only takes the script parameters and executes it.
At this point, I'm wondering: "What is the sense to generate the files ?"
As someone said, this shouldn't be an obs service but a simple front-end functionality.
Now, clicking on "check upgrade" we could show a simple html interface to put the parameters and run the script contacting the web service just created.
Obviously, that will require a simple Webui::WebuiController which will handle the requests and will run the classic CRUD operations on a new table which will be related with Package table with a one-to-one relation.
This is much better IMO.

For both solution:
At the moment, we are only talking of a manual action but I think that an automatic "check upgrade" are required otherwise it will be a strongly incomplete functionality IMO.
The current ApplicationJob had the efficiency problem because it scanned all the packages and we have much packages...I am very agree with you!
That can be simply resolved from query by setting LIMIT and OFFSET. To handle them could we put in a YAML file which Ruby looks like easily to handle.
However, the second solution is much better about that because directly from query we can pull out only the packages which have a record in CheckUpgrade Table replacing read the "_service" file and to check if "check upgrade" service is defined. (more slow)

Personal opinion:
With the first solution, all our operation require access/write/read/delete on the filesystem....
The second solution is more fast!
The I/O operation on the file system have a major throughtput than to data base.

At this point, I stop the develop.
We have to decide what we do.
Regards

pandom79:
This is my idea about how it should work.
Obviously, you would want to change something or
something could be escaped me ....
Please note: after the changes, we can convert html to haml

PR #11789, opened 2021-10-27
Allow users to name their tokens

eduardoj:
Peek 2021-10-29 18-31_Add_name_to_tokens

### DIFF
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -1,6 +1,6 @@
 class Webui::Users::TokensController < Webui::WebuiController
   before_action :set_token, only: [:edit, :update, :destroy]
-  before_action :set_params, :set_package, only: [:create]
+  before_action :set_parameters, :set_package, only: [:create]
 
   after_action :verify_authorized, except: :index
   after_action :verify_policy_scoped, only: :index
@@ -27,17 +27,15 @@ class Webui::Users::TokensController < Webui::WebuiController
           flash.now[:success] = "Token string successfully regenerated! Make sure you save it - you won't be able to access it again."
           render partial: 'update', locals: { string: @token.string }
         else
-          flash.now[:error] = 'Failed to regenerate Token string'
+          flash.now[:error] = "Failed to regenerate Token string: #{@token.errors.full_messages.to_sentence}"
           render partial: 'update'
         end
       end
       format.html do
-        new_scm_token = params.require(:token).except(:string_readonly).permit(:scm_token)
-
-        if @token.update(new_scm_token)
+        if @token.update(update_parameters)
           flash[:success] = 'Token successfully updated'
         else
-          flash[:error] = 'Failed to update Token'
+          flash[:error] = "Failed to update Token: #{@token.errors.full_messages.to_sentence}"
         end
 
         redirect_to tokens_url
@@ -79,12 +77,16 @@ class Webui::Users::TokensController < Webui::WebuiController
     redirect_to tokens_url
   end
 
-  def set_params
-    @params = params.except(:project_name, :package_name).require(:token).except(:string_readonly).permit(:type, :scm_token).tap do |token_parameters|
+  def set_parameters
+    @params = params.except(:project_name, :package_name).require(:token).except(:string_readonly).permit(:type, :name, :scm_token).tap do |token_parameters|
       token_parameters.require(:type)
     end
     @params = @params.except(:scm_token) unless @params[:type] == 'workflow'
     @extra_params = params.slice(:project_name, :package_name).permit!
+  end
+
+  def update_parameters
+    params.require(:token).except(:string_readonly).permit(:name, :scm_token).reject! { |k, v| k == 'scm_token' && (@token.type != 'Token::Workflow' || v.empty?) }
   end
 
   def set_package
--- a/src/api/app/models/token.rb
+++ b/src/api/app/models/token.rb
@@ -7,6 +7,11 @@ class Token < ApplicationRecord
 
   has_secure_token :string
 
+  before_validation do
+    self.name ||= ''
+  end
+
+  validates :name, length: { maximum: 64 }
   validates :user, presence: true
   validates :string, uniqueness: { case_sensitive: false }
   validates :scm_token, absence: true, if: -> { type != 'Token::Workflow' }
@@ -51,6 +56,7 @@ end
 # Table name: tokens
 #
 #  id         :integer          not null, primary key
+#  name       :string(64)       default("")
 #  scm_token  :string(255)      indexed
 #  string     :string(255)      indexed
 #  type       :string(255)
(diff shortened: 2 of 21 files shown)


src/api/db/migrate/20211027141958_add_name_to_tokens.rb
```diff
@@ -0,0 +1,10 @@
+class AddNameToTokens < ActiveRecord::Migration[6.1]
+  def up
+    add_column :tokens, :name, :string, limit: 30
```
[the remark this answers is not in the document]
dmarcoux: Think about how we're going to display this name in the UI. Will 64 characters fit in the UI?

src/api/app/models/token.rb
```diff
@@ -7,6 +7,11 @@ class Token < ApplicationRecord
 
   has_secure_token :string
 
+  before_validation do
+    self.name = '' if name.nil?
```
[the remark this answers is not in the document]
dmarcoux: How about this? `name` will be set to an empty string only if its value is either `nil` or `false.

```suggestion
    self.name ||= ''
```

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -32,9 +32,7 @@ def update
         end
       end
       format.html do
-        new_scm_token = params.require(:token).except(:string_readonly).permit(:scm_token)
-
-        if @token.update(new_scm_token)
+        if @token.update(update_parameters)
           flash[:success] = 'Token successfully updated'
         else
           flash[:error] = 'Failed to update Token'
```
dmarcoux, line 38: If something goes wrong, users won't know what went wrong. We should include error messages in the `flash`.

```suggestion
          flash[:error] = "Failed to update Token: #{@token.errors.full_messages.to_sentence}"
```

src/api/app/models/token.rb
```diff
@@ -7,6 +7,11 @@ class Token < ApplicationRecord
...
+  end
+
+  validates :name, length: { maximum: 30 }
```
dmarcoux, line 14: This validation doesn't match the length of the column anymore.

```suggestion
  validates :name, length: { maximum: 64 }
```

src/api/app/models/token.rb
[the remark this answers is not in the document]
danidoni: This one is better, is short but explicit. :+1:

src/api/app/views/webui/users/tokens/edit.html.haml
```diff
@@ -23,6 +23,12 @@
               .form-group
                 = f.label(:type, 'Operation:')
                 = @token.class.token_name.capitalize
+              .form-group
+                = f.label(:name, 'Name:')
+                %span.d-none#original-token-name
```
dmarcoux, line 28: What is the reason for having this? The token name is displayed in the input below when the page is loaded, so I suggest removing this.

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -80,13 +78,17 @@ def set_token
...
-    @params = params.except(:project_name, :package_name).require(:token).except(:string_readonly).permit(:type, :scm_token).tap do |token_parameters|
+    @params = params.except(:project_name, :package_name).require(:token).except(:string_readonly).permit(:type, :name, :scm_token).tap do |token_parameters|
       token_parameters.require(:type)
     end
     @params = @params.except(:scm_token) unless @params[:type] == 'workflow'
     @extra_params = params.slice(:project_name, :package_name).permit!
   end
 
+  def update_parameters
+    params.require(:token).except(:string_readonly).permit(:name, :scm_token).reject! { |k, v| k == 'scm_token' && v.empty? }
```
dmarcoux, line 89: It should be possible to update `scm_token` only if the token's type is `workflow`.

src/api/db/migrate/20211027141958_add_name_to_tokens.rb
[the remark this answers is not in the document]
danidoni: We discussed this with @eduardoj offline. I agree that 30 removes issues with long names on the UI, but I think it's too short: e.g: `Open Build Service GitHub Token` is already 31 characters...
Maybe we can find a sweet spot between 30 and 64, what about 45?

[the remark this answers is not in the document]
dmarcoux: I'm not against longer names, but the UI will have to deal with them in some way. Truncating should be good enough.

src/api/app/models/token.rb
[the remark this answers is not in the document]
eduardoj: Suggestion `self.name ||= ''` applied.

src/api/app/controllers/webui/users/tokens_controller.rb
eduardoj: True. I also changed the error message in the case of regenerating the SCM string.

src/api/app/models/token.rb
eduardoj: Ops. Changed.

eduardoj:
I renamed the method `set_params` to `set_parameters`, and rebased on master.

src/api/app/controllers/webui/users/tokens_controller.rb
eduardoj: Yes. I added that check in the `reject!` block.

src/api/app/views/webui/users/tokens/edit.html.haml
eduardoj: That's used after clicking in "Regenerate Token". Instead of having this:

Screenshot from 2021-11-02 11-49-28

we would have something similar to this:

Screenshot from 2021-11-02 11-54-33

dmarcoux: Okay, this isn't obvious at all... but there's not much we can do for this now in this PR. Thanks for the explanation.

src/api/db/migrate/20211027141958_add_name_to_tokens.rb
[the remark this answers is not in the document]
vpereira: so I think we can adjust it later when working on the UI, right?

src/api/app/views/person/token/index.xml.builder
```diff
@@ -1,6 +1,6 @@
 xml.directory(count: @list.length) do |dir|
   @list.each do |token|
-    p = { id: token.id, string: token.string, kind: token.token_name }
+    p = { id: token.id, string: token.string, kind: token.token_name, name: token.name }
```
vpereira, line 3: :thought_balloon: `token.token_name` and `token.name` could be a source of confusion

vpereira (approved):

src/api/app/views/person/token/index.xml.builder
dmarcoux: I agree :+1: 

I created a card in our backlog for this: https://trello.com/c/0yujk7O5/2391-remove-tokentokenname-method

vpereira: @dmarcoux 

It would be much better to have fixed it here than try to refactor it in the future (spoiler: I'm in the future, and we won't refactor it :))

PR #11793, opened 2021-10-28
Validate project/package names in steps

rubhanazeem:


krauselukas (approved):
Looks pretty good to me

src/api/spec/validators/workflow_steps_validator_spec.rb
```diff
@@ -51,7 +51,8 @@
       it 'is not valid and has an error message' do
         subject.valid?
         expect(subject.errors.full_messages.to_sentence).to eq("The following workflow steps are unsupported: 'unsupported_step' and " \
-                                                               "The 'source_package' key is missing and The 'target_project' key is missing")
+                                                               "The 'source_package' key is missing, The 'target_project' key is missing, " \
+                                                               "and invalid source package ''")
```
dmarcoux, line 55: `invalid source package ''` should be in the error messages since the `source_package` key is missing. We should avoid this duplication in the error messages, it can confuse users.

rubhanazeem: @dmarcoux Okay, I'll add this in a follow-up PR.

PR #11794, opened 2021-10-28
Run validations upfront, then report any error back to the SCM

dmarcoux:
This will help users debug their workflows if something is wrong in their `.obs/workflows.yml`.

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -7,14 +7,16 @@ class Token::Workflow < Token
 
   def call(options)
     set_triggered_at
-    raise ArgumentError, 'A payload is required' if options[:payload].nil?
+    raise Token::Errors::MissingPayload, 'A payload is required' if options[:payload].nil?
 
-    scm_webhook = TriggerControllerService::ScmExtractor.new(options[:scm], options[:event], options[:payload]).call
-    return unless scm_webhook.valid?
+    @scm_webhook = TriggerControllerService::ScmExtractor.new(options[:scm], options[:event], options[:payload]).call
+    yaml_file = Workflows::YAMLDownloader.new(@scm_webhook.payload, token: self).call
+    @workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: @scm_webhook, token: self).call
 
-    yaml_file = Workflows::YAMLDownloader.new(scm_webhook.payload, token: self).call
-    workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: scm_webhook, token: self).call
-    workflows.each(&:call)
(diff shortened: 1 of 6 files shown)


src/api/app/models/token/workflow.rb
```diff
@@ -7,14 +7,16 @@ def self.token_name
 
   def call(options)
     set_triggered_at
-    raise ArgumentError, 'A payload is required' if options[:payload].nil?
+    raise Token::Errors::MissingPayload, 'A payload is required' if options[:payload].nil?
```
dmarcoux, line 10: Raising a custom error to ensure that specs are valid. It would be possible to have a spec pass when calling `call` with too many arguments, even though this fictional spec wouldn't actually test this line.

src/api/app/services/workflows/yaml_to_workflows_service.rb
```diff
@@ -21,7 +21,6 @@ def create_workflows
 
       parsed_workflows_yaml
         .map { |_workflow_name, workflow_instructions| Workflow.new(workflow_instructions: workflow_instructions, scm_webhook: @scm_webhook, token: @token) }
-        .select(&:valid?)
```
dmarcoux, line 24: We want to get all workflows back to gather potential validation errors of non-valid workflows.

src/api/spec/controllers/trigger_workflow_controller_spec.rb
```diff
@@ -1,38 +1,27 @@
 require 'rails_helper'
 
 RSpec.describe TriggerWorkflowController, type: :controller, beta: true do
-  let(:user) { create(:confirmed_user, :in_beta, login: 'foo') }
```
dmarcoux, line 4: Local variables are defined in the context where they are needed to clarify what is needed in a specific context. It's much easier to fix specs with this approach.

src/api/spec/models/token/workflow_spec.rb
```diff
@@ -1,165 +1,133 @@
 require 'rails_helper'
 
-RSpec.describe Token::Workflow, vcr: true do
```
dmarcoux, line 3: Mocking as much as possible to not test the classes which `Token::Workflow` calls. Those classes are already tested in their own spec file.

danidoni (approved):

src/api/spec/models/token/workflow_spec.rb
rubhanazeem: Can you check if we have cassettes for this spec file? I see you have removed the `vcr: true` and we can remove the cassettes I think.

krauselukas: I just checked, couldnt find any

krauselukas (approved):

krauselukas:
Hmmm, backend_test seems stuck, since no backend changes involved I will merge anyway

PR #11801, opened 2021-10-29
Record time and date of the last trigger of a token

krauselukas:
In the future we want to enable users to see when they used/triggered their tokens the last time. Therefore we need to add a new column to the tokens table and adapt existing api endpoints to show the new field.

### DIFF
--- a/src/api/spec/models/token/service_spec.rb
+++ b/src/api/spec/models/token/service_spec.rb
@@ -1,0 +1,30 @@
+require 'rails_helper'
+
+RSpec.describe Token::Service do
+  let(:user) { create(:user, login: 'foo') }
+  let(:token) { create(:service_token, user: user) }
+
+  subject { token.call(package: 'bar') }
+
+  describe '#call' do
+    context 'successful triggered' do
+      before do
+        allow(Backend::Api::Sources::Package).to receive(:trigger_services).and_return(true)
+      end
+
+      it 'records the current date and time in the triggered_at column' do
+        expect { subject }.to change(token, :triggered_at)
+      end
+    end
+
+    context 'triggered without success' do
+      before do
+        allow(Backend::Api::Sources::Package).to receive(:trigger_services).and_raise(Backend::Error, 'something went wrong')
+      end
+
+      it 'records the current date and time in the triggered_at column' do
+        expect { subject }.to raise_error(Backend::Error).and(change(token, :triggered_at))
+      end
+    end
+  end
+end
--- a/src/api/spec/cassettes/Token_Release/_call/when_a_manual_release_target_is_set/records_the_current_date_and_time_in_the_triggered_at_column.yml
+++ b/src/api/spec/cassettes/Token_Release/_call/when_a_manual_release_target_is_set/records_the_current_date_and_time_in_the_triggered_at_column.yml
@@ -1,0 +1,245 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: http://backend:5352/build/Bar?cmd=suspendproject&comment=Releasing%20via%20trigger%20event
+    body:
+      encoding: UTF-8
+      string: ''
+    headers:
+      Content-Type:
+      - application/octet-stream
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+      Accept:
+      - "*/*"
+      User-Agent:
+      - Ruby
+  response:
+    status:
+      code: 200
+      message: OK
+    headers:
+      Content-Type:
+      - text/xml
+      Cache-Control:
+      - no-cache
+      Connection:
+      - close
+      Content-Length:
+      - '21'
+    body:
+      encoding: UTF-8
+      string: '<status code="ok" />
+
+'
+  recorded_at: Tue, 02 Nov 2021 11:08:53 GMT
+- request:
+    method: post
+    uri: http://backend:5352/source/Bar/bar_package?cmd=copy&comment=Release%20from%20Bar:Staging%20/%20bar_package&expand=1&noservice=1&opackage=bar_package&oproject=Bar:Staging&user=foo&withacceptinfo=1&withvrev=1
+    body:
+      encoding: UTF-8
+      string: ''
+    headers:
+      Content-Type:
+      - application/octet-stream
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+      Accept:
+      - "*/*"
+      User-Agent:
+      - Ruby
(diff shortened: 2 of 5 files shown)


src/api/public/apidocs-new/components/schemas/person_tokens.yaml
```diff
@@ -21,6 +21,10 @@ properties:
           type: string
           xml:
             attribute: true
+        triggered_at:
+          type: string
```
dmarcoux, line 25: I think we're missing the `format: date-time` here. See the [OpenAPI specification for data types](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#data-types).

src/api/spec/models/token/service_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.describe Token::Service do
+  let(:user) { create(:user, login: 'foo') }
+  let(:token) { create(:service_token, user: user) }
+  let(:time_now) { Time.now.utc.to_s }
```
vpereira, line 6: You can drop the `Timecop.freeze` and etc, if you just define the time as something deterministic like:

```
let(:time_now) { Time.new('2021-11-02 12:41:33 +0100').utc.to_s }
```

src/api/db/migrate/20211028111158_add_triggered_at_to_tokens.rb
```diff
@@ -0,0 +1,5 @@
+class AddTriggeredAtToTokens < ActiveRecord::Migration[6.1]
+  def change
+    add_column :tokens, :triggered_at, :datetime
```
saraycp, line 3: I wonder if it's more accurate naming the column as `last_triggering_at` instead of `triggered_at` :thinking:

src/api/spec/models/token/service_spec.rb
krauselukas: @vpereira but how do I control which time is set in the token object if I dont use the freeze?

src/api/public/apidocs-new/components/schemas/person_tokens.yaml
krauselukas: I thought we are not yet dealing with the formats in openapi docs, but fine for me, added the format :)

src/api/spec/models/token/service_spec.rb
krauselukas: I mean I get that I can use a static date and time, but somehow I need to ensure that the same time is set in the `triggered_at` field after the token got triggered. Afraid of flickering tests if I dont use the freeze

src/api/app/models/token/workflow.rb
```diff
@@ -16,6 +16,8 @@ def call(options)
     workflows.each(&:call)
   rescue Octokit::Unauthorized, Gitlab::Error::Unauthorized => e
     raise Token::Errors::SCMTokenInvalid, e.message
+  ensure
+    set_triggered_at
```
saraycp, line 20: I would also suggest setting the triggering time right after the token is called, not after the whole `.call`method is performed.

For example, if the token needs to run many workflows with a big list of steps, the difference between the time when the user really triggered the token and the one we set in the database can be significant.

What do you think @krauselukas @dmarcoux @vpereira ?

src/api/db/migrate/20211028111158_add_triggered_at_to_tokens.rb
dmarcoux: I would go with `last_trigger_at`.

src/api/app/models/token/workflow.rb
saraycp: Edit: "... right after/before the token is called"

src/api/spec/models/token/service_spec.rb
dmarcoux: Without `Timecop`, you'll still have to mock the value returned by `Time.zone.now`. Using `Timecop` is fine.

src/api/app/models/token/workflow.rb
krauselukas: Alright, I will change the approach

src/api/db/migrate/20211028111158_add_triggered_at_to_tokens.rb
krauselukas: But doesn't `triggered_at` imply that it is the `last` trigger? Sure, I can rename it, but to me it aligns better with the terms `updated_at` and `created_at`

vpereira: I would go with `triggered_at`

src/api/spec/models/token/service_spec.rb
vpereira: You shouldn't care about the specific value but if the behavior is there. I.e is the value changing, is the value other than nil, etc.

vpereira: If you really want to test the specific time, you should go with https://api.rubyonrails.org/v5.2.6/classes/ActiveSupport/Testing/TimeHelpers.html

src/api/app/models/token/workflow.rb
krauselukas: This is addressed, setting the now the date and time right at the beginning of the token call

vpereira (commented):
Hey, I see that you are calling `set_triggered_at` in all `call` methods. All classes calling it are subclass from token. Would not be better to have on `Token#call` the call to `set_triggered_at` and the subclasses just call `super` ?

krauselukas:
> Hey, I see that you are calling `set_triggered_at` in all `call` methods. All classes calling it are subclass from token. Would not be better to have on `Token#call` the call to `set_triggered_at` and the subclasses just call `super` ?

Yeah I thought the same, but then we have the thing again with the RSS token, thats why I called it in the subclasses

src/api/db/migrate/20211028111158_add_triggered_at_to_tokens.rb
saraycp: For me, `updated_at` implies "the last modification". With the `triggered_at` I don't have the same feeling.
Maybe because `updated_at` is too integrated into my brain. Anyway, I'm ok with `triggered_at`.

src/api/spec/models/token/service_spec.rb
krauselukas: @vpereira please have another look, I now only check if the `triggered_at` field changed, not checking for a particular date

dmarcoux (approved):

PR #11804, opened 2021-11-02
Since 1/9/2021, github won't accept unauthenticated git

vpereira:
More details: https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "src/backend/build"]
-	path = src/backend/build
-	url = git://github.com/openSUSE/obs-build.git
+  path = src/backend/build
+  url = https://github.com/openSUSE/obs-build.git


krauselukas (commented):
I would just fix the indentation and maybe rename the commit description, you describe what happend not what your commit is doing

krauselukas:
is the codeclimate failure related to the changes, seem it cannot fetch the code anymore

vpereira:
> is the codeclimate failure related to the changes, seems it cannot fetch the code anymore

it is probably trying to fetch information from master and github is blocking it

vpereira:
> I would just fix the indentation and maybe rename the commit description, you describe what happend not what your commit is doing

updated!

eduardoj (approved):
Tabs have changed to two spaces. It should not be an issue.

PR #11806, opened 2021-11-03
Refactor out table of token index UI

krauselukas:
We want to move away from simply showing db data in a table.

WIP: Maybe some specs and fine tuning

Before:
Screenshot_2021-11-03 Open Build Service(1)

After:
Screenshot_2021-11-03 Open Build Service

Screen Shot 2021-11-04 at 16 07 24

### DIFF
--- a/src/api/app/components/token_card_component.html.haml
+++ b/src/api/app/components/token_card_component.html.haml
@@ -1,0 +1,27 @@
+.card.m-1.p-2
+  .card-body
+    - if @token.name.present?
+      %h5.card-title
+        = @token.name
+    - else
+      %h5.card-title.font-italic No name
+    %p.card-text
+      = operation
+    - if @token.package
+      %p.card-text.text-word-break-all
+        Package:
+        = token_package_link
+    - if @token.triggered_at.present?
+      %p.card-text
+        "Last trigger on #{@token.triggered_at}"
+    .row
+      .col
+        - if policy(@token).edit?
+          = link_to(edit_token_path(@token), title: 'Edit Token') do
+            %i.fas.fa-edit
+        - if policy(@token).show?
+          = link_to(token_trigger_path(@token), title: 'Trigger Token') do
+            %i.fas.fa-project-diagram
+        = link_to '#', title: 'Delete Token',
+          data: { toggle: 'modal', target: '#delete-token-modal', token_id: @token.id, action: token_path(@token) } do
+          %i.fas.fa-times-circle.text-danger
--- a/src/api/app/components/token_card_component.rb
+++ b/src/api/app/components/token_card_component.rb
@@ -1,0 +1,18 @@
+class TokenCardComponent < ApplicationComponent
+  with_collection_parameter :token
+
+  def initialize(token:)
+    super
+
+    @token = token
+  end
+
+  def operation
+    "Operation: #{@token.class.token_name.capitalize}"
+  end
+
+  def token_package_link
+    link_to("#{truncate(@token.package.project.name, length: 32)}/#{truncate(@token.package.name, length: 32)}",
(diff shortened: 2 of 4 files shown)


krauselukas:
[Redesign the Tokens UI](https://trello.com/c/RZWPohVw/1787-redesign-the-tokens-ui)

src/api/app/components/token_card_component.html.haml
```diff
@@ -0,0 +1,24 @@
+.card.m-1.p-2
+  .card-body
+    %h5.card-title
```
dmarcoux, line 3: If a token doesn't have a name, I would display _No name_ (in italic) instead of an empty `<h5>`.

src/api/app/views/webui/users/tokens/index.html.haml
```diff
@@ -3,36 +3,11 @@
...
-                      %i.fas.fa-edit
-                  - if policy(token).show?
-                    = link_to(token_trigger_path(token), title: 'Trigger Token') do
-                      %i.fas.fa-project-diagram
-                  = link_to '#', title: 'Delete Token',
-                    data: { toggle: 'modal', target: '#delete-token-modal', token_id: token.id, action: token_path(token) } do
-                    %i.fas.fa-times-circle.text-danger
-
+      .d-flex.flex-wrap
+        - @tokens.each do |token|
```
dmarcoux, line 9: This loop could be replaced by a one-liner:

```suggestion
        = render TokenCardComponent.with_collection(@tokens)
```

src/api/app/components/token_card_component.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    @token = token
+  end
+
+  def operation
+    "Operation: #{@token.class.token_name.capitalize}"
+  end
+
+  def token_package_link
+    link_to("#{truncate(@token.package.project.name, length: 80)}/#{truncate(@token.package.name, length: 80)}",
```
dmarcoux, line 13: So this link could be 161 characters long... this is too much. It's going to increase the card's width and the UI will look weird. It's also a waste of space. The package and project names should be truncated or elided.

krauselukas: True, its very long, I set the truncation to 64 in total, so the same length that we allow for the token names

src/api/app/views/webui/users/tokens/index.html.haml
krauselukas: Thanks, was actually searching already how to use collections :)

src/api/app/components/token_card_component.html.haml
krauselukas: Sounds good to me, adapted

rubhanazeem (approved):

src/api/spec/components/token_card_component_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  let(:user) { create(:user) }
+
+  before do
+    Flipper.enable(:trigger_workflow)
+    User.session = user
+    render_inline(described_class.new(token: token))
+  end
+
+  context 'token with a name' do
+    let(:token) { create(:rebuild_token, user: user, name: 'foo_token') }
```
dmarcoux, line 13: I don't think it's needed to store the token in the database to properly test this view component, so we can rely on `build_stubbed` instead of `create` for all tokens in this file.

Here's an [article](https://thoughtbot.com/blog/use-factory-girls-build-stubbed-for-a-faster-test) on this.

krauselukas: right, shouldn't be necessary

krauselukas: done

dmarcoux (approved):

PR #11811, opened 2021-11-04
Don't run project/package name validations if the keys are missing

rubhanazeem:
Validating a project/package should happen only if the key is present in the step instructions to avoid duplication in error messages.

### DIFF
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -150,11 +150,8 @@ class Workflow::Step
 
   # Only used in LinkPackageStep and BranchPackageStep.
   def validate_source_project_and_package_name
-    errors.add(:base, "invalid source project '#{source_project_name}'") unless Project.valid_name?(source_project_name)
-    errors.add(:base, "invalid source package '#{source_package_name}'") unless Package.valid_name?(source_package_name)
-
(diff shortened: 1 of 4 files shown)


src/api/app/models/workflow/step.rb
```diff
@@ -150,8 +150,8 @@ def report_to_scm(workflow_filters)
 
   # Only used in LinkPackageStep and BranchPackageStep.
   def validate_source_project_and_package_name
-    errors.add(:base, "invalid source project '#{source_project_name}'") unless Project.valid_name?(source_project_name)
-    errors.add(:base, "invalid source package '#{source_package_name}'") unless Package.valid_name?(source_package_name)
+    errors.add(:base, "invalid source project '#{source_project_name}'") if step_instructions[:source_project] && !Project.valid_name?(source_project_name)
+    errors.add(:base, "invalid source package '#{source_package_name}'") if step_instructions[:source_package] && !Package.valid_name?(source_package_name)
```
dmarcoux, line 154: A small nitpick for consistency... The code for `target_project` could be in the same format as this line.

rubhanazeem: Done.

saraycp (approved):
LGTM

PR #11812, opened 2021-11-04
Add a delay to wait for the AJAX call

eduardoj:
This should prevent a race condition to happen between clicking the checkbox and reloading the page.

In the past we tried [waiting with javascript](https://github.com/openSUSE/open-build-service/pull/10754). Then we opted for [not using javascript](https://github.com/openSUSE/open-build-service/pull/10782).


eduardoj:
I will try the `Capybara.default_max_wait_time = 6` approach in another PR.

PR #11814, opened 2021-11-05
Use ci docker image and the already installed gems

vpereira:
This PR includes the following changes:

The frontend-base image has chromium installed, influences the download
time from every single CI step using it.

When building the frontend-* images on OBS we bundle already our gems.
Now we are using them while in CI, avoiding downloading it again

### DIFF
--- a/.circleci/conditional_config.yml
+++ b/.circleci/conditional_config.yml
@@ -1,4 +1,3 @@
-version: 2.1
 
 parameters:
   run-backend-tests:
@@ -49,11 +48,15 @@ images:
     image: registry.opensuse.org/obs/server/unstable/containers/containers/openbuildservice/frontend-base:latest
     <<: *common_frontend_config
 
+  - &frontend_base_ci
+    image: registry.opensuse.org/obs/server/unstable/containers/containers/openbuildservice/frontend-base-ci:latest
+    <<: *common_frontend_config
+
 aliases:
   - &install_dependencies
     name: install dependencies
     command: |
-      cd ./src/api && bundle check --path=vendor/bundle || bundle install --path=vendor/bundle --jobs=4 --retry=3
+      cd ./src/api && bundle check  || bundle install --jobs=4 --retry=3
   - &wait_for_database
     name: Wait for DB
     command: mysqladmin ping -h db
@@ -106,7 +109,6 @@ aliases:
     save_cache:
       key: circlev2-{{ .Branch }}-{{ checksum "./src/api/Gemfile.lock" }}
       paths:
-        - vendor/bundle
         - src/api/tmp/rubocop_cache_root_dir
         - src/api/tmp/rubocop_cache_rails_dir
  
@@ -124,7 +126,7 @@ jobs:
 
   linters:
     docker:
-      - <<: *frontend_base
+      - <<: *frontend_base_ci
     steps:
       - attach_workspace:
          at: .
@@ -173,9 +175,7 @@ jobs:
     
   checkout_code:
     docker:
-      - <<: *frontend_base
-    steps:
-      - checkout
+      - <<: *frontend_base_ci
       - *restore_gem_cache
       - run: *install_dependencies
       - *save_gem_cache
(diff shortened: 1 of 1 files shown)


.circleci/conditional_config.yml
```diff
@@ -49,11 +49,19 @@ images:
     image: registry.opensuse.org/obs/server/unstable/containers/containers/openbuildservice/frontend-base:latest
     <<: *common_frontend_config
 
+  - &frontend_ci
+    image: registry.opensuse.org/obs/server/unstable/containers/containers/openbuildservice/frontend-base-ci:latest
```
vpereira, line 53: new image 600 MB smaller than `frontend-base`

.circleci/conditional_config.yml
```diff
@@ -173,7 +181,7 @@ jobs:
     
   checkout_code:
     docker:
-      - <<: *frontend_base
+      - <<: *frontend_ci
```
vpereira, line 184: we don't need chromium here

.circleci/conditional_config.yml
```diff
@@ -197,7 +205,7 @@ jobs:
   rspec:
     parallelism: 3
     docker:
-      - <<: *frontend_base
+      - <<: *frontend_ci
```
vpereira, line 208: we don't need chromium here

.circleci/conditional_config.yml
```diff
@@ -275,7 +283,7 @@ jobs:
 
   migrations_tests:
     docker:
-      - <<: *frontend_backend
+      - <<: *frontend_ci
```
vpereira, line 286: we don't need chromium here

.circleci/conditional_config.yml
```diff
@@ -325,8 +333,7 @@ jobs:
 
   coverage:
     docker:
-      - <<: *frontend_base
-
+      - <<: *frontend_ci
```
vpereira, line 336: we don't need chromium here

.circleci/conditional_config.yml
```diff
@@ -385,7 +392,7 @@ jobs:
 
   next_rails-rspec:
     docker:
-      - <<: *frontend_base
+      - <<: *frontend_ci
```
vpereira, line 395: we don't need chromium here

.circleci/conditional_config.yml
hennevogel: I must say I find the naming a bit confusing. We use `frontend_base` also in the CI. As many others of the images. CI isn't what makes this image different.

Also why is the `frontend_ci` image not based on the `frontend_base` image?

hennevogel (changes requested):

.circleci/conditional_config.yml
hennevogel: Ah sorry, the other way around.

.circleci/conditional_config.yml
eduardoj: We also don't need the backend here. :+1:

.circleci/conditional_config.yml
vpereira: now, `frontend_base` is based on `frontend_ci`

.circleci/conditional_config.yml
dmarcoux: We run the feature specs in this job, so Chromium is needed.

.circleci/conditional_config.yml
dmarcoux: The image is called `frontend-base-ci` on the registry, but it's called `frontend-ci` here. I suggest keeping the same name here.

.circleci/conditional_config.yml
vpereira: fixed.

.circleci/conditional_config.yml
vpereira: would be ok for you @hennevogel ?

hennevogel: What exactly? Using the same name? For sure.

PR #11817, opened 2021-11-08
Cache gems in the system directory

hennevogel:
In the development environment `vendor/` is a directory in the git checkout
that is mounted into the container. We can't use this as place to cache
the gems in the frontend-* images. Use the system directory instead.


hennevogel:
this wont work as cache restore only has user permissions.

PR #11803, opened 2021-11-02
Improve gem caching circleci and make linters job mandatory before rspec, minitest and feature jobs

vpereira:
In this PR, we improved the caching among jobs, dropping around 2 minutes in average while running `minitest`, `specs` and `feature` jobs.

i.e: https://app.circleci.com/pipelines/github/vpereira/open-build-service/817/workflows/a4476512-e2e7-4260-b59f-9b3747e357ee/jobs/14592

Beside that now, we just start `rspec`, `minitest` and `feature` jobs if `linter` runs successfully.

PR #11819, opened 2021-11-08
Revert gem caching on circleci

hennevogel:
Reverts parts of openSUSE/open-build-service#11803

We are already caching the gems in the container images we build ourselves. The frontend-base-ci container rebuilds every time we push to master. No need to cache gems again on circle.

### DIFF
--- a/.circleci/conditional_config.yml
+++ b/.circleci/conditional_config.yml
@@ -12,7 +12,7 @@ parameters:
   run-javascripts-linter:
     type: boolean
     default: false
-  
+
   run-apidocs-linter:
     type: boolean
     default: false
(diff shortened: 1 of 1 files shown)


hennevogel:
I have changed [OBS:Server:Unstable:Containers/frontend-base-ci](https://build.opensuse.org/package/rdiff/OBS:Server:Unstable:Containers/frontend-base-ci?linkrev=base&rev=7) already

hennevogel:
And now also adopted our webhook setup to trigger `frontend-base-ci` and not `frontend-base`.

hennevogel:
And removed the services from `frontend-base`

hennevogel:
## Before

Before


## After

After

PR #11820, opened 2021-11-08
Add missing `target_project` key in workflows

eduardoj:


.obs/workflows.yml
```diff
@@ -3,6 +3,7 @@ workflow:
     - branch_package:
         source_project: OBS:Server:Unstable
         source_package: obs-server
+        target_project: OBS:Server:Unstable
```
hennevogel, line 6: Currently we build here https://build.opensuse.org/project/subprojects/home:bs-team

hennevogel (changes requested):

.obs/workflows.yml
hennevogel: And we should keep doing this IMHO

eduardoj: Ops, sure. Changed.

hennevogel (approved):

PR #11829, opened 2021-11-10
Fix flickering smoke test

eduardoj:
Searching for mariadb and mysql services resulted sometimes in `mariadb`, and some other times in `mysql`.

`systemctl list-units` always return the services in alphabetical order. Rely on this, to stop searching on the first match from `mariadb` or `mysql`.

This will prevent this flickering errors to happen in [openQA tests](https://openqa.opensuse.org/group_overview/62).


M0ses (approved):
LGTM

PR #11831, opened 2021-11-10
Add workflow runs model and table

eduardoj:

### DIFF
--- a/src/api/app/models/workflow_run.rb
+++ b/src/api/app/models/workflow_run.rb
@@ -1,0 +1,31 @@
+class WorkflowRun < ApplicationRecord
+  validates :response_url, length: { maximum: 255 }
+  validates :request_headers, :request_payload, :status, presence: true
+
+  belongs_to :token, class_name: 'Token::Workflow'
+
+  enum status: {
+    running: 0,
+    success: 1,
+    fail: 2
+  }
+end
+
+# == Schema Information
+#
+# Table name: workflow_runs
+#
+#  id              :integer          not null, primary key
+#  request_headers :text(65535)      not null
+#  request_payload :text(65535)      not null
+#  response_body   :text(65535)
+#  response_url    :string(255)
+#  status          :integer          default("running"), not null
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  token_id        :integer          not null, indexed
+#
+# Indexes
+#
+#  index_workflow_runs_on_token_id  (token_id)
+#
--- a/src/api/db/migrate/20211110142710_create_workflow_runs.rb
+++ b/src/api/db/migrate/20211110142710_create_workflow_runs.rb
@@ -1,0 +1,15 @@
+class CreateWorkflowRuns < ActiveRecord::Migration[6.1]
+  def change
+    create_table :workflow_runs, id: :integer do |t|
+      t.text :request_headers, null: false
+      t.text :request_payload, null: false
+      t.integer :status, limit: 1, default: 0, null: false
+      t.text :response_body
+      t.string :response_url
+
+      t.timestamps
+
+      t.belongs_to :token, index: true, type: :integer, null: false
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


src/api/db/migrate/20211110142710_create_workflow_runs.rb
```diff
@@ -0,0 +1,15 @@
+class CreateWorkflowRuns < ActiveRecord::Migration[6.1]
+  def change
+    create_table :workflow_runs, id: :integer do |t|
+      t.text :request_headers
+      t.text :request_payload
+      t.string :status
```
dmarcoux, line 6: Having an [enum](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html) for `status` would be better to limit this to only a set of values. The column would then be a tinyint (so integer with `limit: 1` in the migration) instead of a string.

src/api/db/migrate/20211110142710_create_workflow_runs.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.text :response
```
dmarcoux, line 7: This would be only for the response body, right?

```suggestion
      t.text :response_body
```

src/api/db/migrate/20211110142710_create_workflow_runs.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.string :response_url
+
+      t.timestamps
+
+      t.belongs_to :token, index: true, type: :integer
```
dmarcoux, line 12: A workflow run should exist only if it's linked to a token, so `token_id` cannot be `null`:

```suggestion
      t.belongs_to :token, index: true, type: :integer, null: false
```

src/api/app/models/workflow_run.rb
```diff
@@ -0,0 +1,3 @@
+class WorkflowRun < ApplicationRecord
```
dmarcoux, line 1: This model should have validations.

dmarcoux (changes requested):

src/api/db/migrate/20211110142710_create_workflow_runs.rb
eduardoj: Sure. Changed.

src/api/db/migrate/20211110142710_create_workflow_runs.rb
eduardoj: Exactly. Changed.

src/api/db/migrate/20211110142710_create_workflow_runs.rb
eduardoj: Right. Change applied.

src/api/app/models/workflow_run.rb
eduardoj: Model validations added. The migration was also adapted. Please have a look again.

dmarcoux: I would add a few more validations:

- `status` has to be one of the valid values with [inclusion](https://guides.rubyonrails.org/active_record_validations.html#inclusion).
- Length of the various string/text columns

saraycp (approved):
LGTM

rubhanazeem:
The PR has been updated. We forgot to add the class name and the foreign key in models. Please have a look again.

src/api/app/models/workflow_run.rb
rubhanazeem: Regarding the `status` we don't need to add an extra validation. It's an enum and enum will only allow the values defined in the model. 
And we are not sure about the length of the string/text columns. Because it's a payload and it's huge. I'd leave it like that. What do you think?

dmarcoux: Oh right, there's a built-in validation with the enum, then it's all good :+1:.

I think validating the length of `response_url` should be enough.

eduardoj: Validation for the length of `response_url` added. Please have a look.

dmarcoux (approved):

Issue #11665, opened 2021-09-28
Github webhook: add support for tags in 'filter' configuration


bluca:
Recently the SCM workflow interface added support for filtering on push to branches. It would be nice, and in line with other CI systems, to be able to filter on tag push events, so that a job can be configured only to run when a new tag is pushed.

The practical use case is to trigger a source service run only when a release tag is pushed to a repository, rather than on every push to a branch.

dmarcoux:
Thank you @bluca for creating this issue. We're definitely going to take your feedback into account for the upcoming development in the SCM/CI integration.

dmarcoux:
![](https://github.trello.services/images/mini-trello-icon.png) [Support push events for tags](https://trello.com/c/C8Ocj5ua/2359-support-push-events-for-tags)

dmarcoux:
In the SCM/CI integration, OBS now supports push events for tags. It's possible to run workflows only if tags are pushed. As for running a workflow only when a tag's name is matching a certain string (with support for basic regular expression), this will be added later.

We've documented this in our latest blog post under the [Push Events for Tags](https://openbuildservice.org/2021/11/22/scm-workflow-runs/) section. Details about how this works is in the [user documentation](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html).

Let us know how this works for you @bluca.

bluca:
Thank you, that's great!

The blog post says `These kinds of events will affect the branch package and the link package steps only.` - is there any chance of getting this also enabled for `rebuild_package`? That's what it would be useful to me - a release tag is pushed, trigger the service to rebuild. So that only releases are built, and other pushes are ignored.

dmarcoux:
This isn't only for branch package and link package steps. All steps can be used. I can see how this part of the blog post can be confusing. I'm going to rephrase this.

For the branch package and link package steps, the package created by those steps will contain the name of the tag which triggered the webhook event.

As for the rebuild package step, nothing changed in its behavior. If you have a workflow running only when push events for tags happen (so with the filter `event: tag_push` as presented in the blog post), the package will be rebuilt whenever a tag is pushed.

bluca:
Thanks, tested and the trigger works! But: the rebuild_package is only doing a rebuild of the current sources, it does not trigger a service run?

dmarcoux:
True, it doesn't trigger a service run, so new sources aren't pulled. We need to handle this.

[...]

PR #11832, opened 2021-11-10
Support push events for tags

dmarcoux:
Part of #11665 (filters for tags aren't included in this PR)

GitHub has a single [push event for commits and tags](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push). GitLab has two separate push events for [commits](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#push-events) and [tags](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#tag-events).

For details on Git references, `git check-ref-format` is a great reference: https://git-scm.com/docs/git-check-ref-format

### DIFF
--- a/src/api/app/validators/scm_webhook_event_validator.rb
+++ b/src/api/app/validators/scm_webhook_event_validator.rb
@@ -1,6 +1,6 @@
 class ScmWebhookEventValidator < ActiveModel::Validator
   ALLOWED_GITHUB_EVENTS = ['pull_request', 'push'].freeze
-  ALLOWED_GITLAB_EVENTS = ['Merge Request Hook', 'Push Hook'].freeze
+  ALLOWED_GITLAB_EVENTS = ['Merge Request Hook', 'Push Hook', 'Tag Push Hook'].freeze
 
   ALLOWED_PULL_REQUEST_ACTIONS = ['closed', 'opened', 'reopened', 'synchronize'].freeze
   ALLOWED_MERGE_REQUEST_ACTIONS = ['close', 'merge', 'open', 'reopen', 'update'].freeze
@@ -41,7 +41,7 @@ class ScmWebhookEventValidator < ActiveModel::Validator
       return true if ALLOWED_MERGE_REQUEST_ACTIONS.include?(@record.payload[:action])
 
       @record.errors.add(:base, 'Merge request action not supported.')
-    when 'Push Hook'
+    when 'Push Hook', 'Tag Push Hook'
       valid_push_event?
     else
       true
@@ -49,8 +49,8 @@ class ScmWebhookEventValidator < ActiveModel::Validator
   end
 
   def valid_push_event?
-    return true if @record.payload.fetch(:ref, '').start_with?('refs/heads/')
+    return true if @record.payload.fetch(:ref, '').start_with?('refs/heads/', 'refs/tags/')
 
-    @record.errors.add(:base, 'Push event supported only for branches.')
+    @record.errors.add(:base, 'Push event supported only for branches/tags with a valid reference.')
   end
 end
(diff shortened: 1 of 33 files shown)


dmarcoux:
[Support push events for tags](https://trello.com/c/C8Ocj5ua/1795-support-push-events-for-tags)

src/api/app/validators/scm_webhook_event_validator.rb
```diff
@@ -49,8 +49,8 @@ def valid_gitlab_event?
   end
 
   def valid_push_event?
-    return true if @record.payload.fetch(:ref, '').start_with?('refs/heads/')
+    return true if @record.payload.fetch(:ref, '').start_with?('refs/heads/', 'refs/tags/')
 
-    @record.errors.add(:base, 'Push event supported only for branches.')
+    @record.errors.add(:base, 'Push event supported only for branches and tags.')
```
danidoni, line 54: :thought_balloon: Is there any other case for pushing, apart from branches or tags?

dmarcoux: References are only for branches or tags, but we never know if someone tinkers with the payload. This is to ensure that we're dealing with a proper payload for either a branch or a tag.

src/api/app/models/workflow.rb
```diff
@@ -74,10 +74,17 @@ def supported_filters
 
   def event_matches_event_filter?
     return true unless supported_filters.key?(:event)
-    return true if filters[:event] == 'push' && scm_webhook.push_event?
-    return true if filters[:event] == 'pull_request' && scm_webhook.pull_request_event?
 
-    false
+    case filters[:events]
```
dmarcoux, line 78: This code is difficult to test, since its method  `event_matches_event_filter?` is private. We wanted to turn this method into a custom validator, but we didn't do it. A service would be a better choice here. It could be easily tested and it would do just one thing, check if the event filter matches the webhook event. I would like to do this in a follow-up PR.

vpereira: > This code is difficult to test, since its method event_matches_event_filter? is private.

Not sure if I understood right, but following the https://github.com/openSUSE/open-build-service/pull/11876#discussion_r752310895, you could just `send` around it, no?

dmarcoux: In this case, I have a clear alternative to the current state. This will be turned into a service, so I don't need to test the private method with `send`.

rubhanazeem:
Do we have to add anything for `configure_repository` or `rebuild_package` step or is it already handled by the changes in `src/api/app/validators/scm_webhook_event_validator.rb`? I see there are changes in the other step files.

dmarcoux:
The behavior of `configure_repositories` and `rebuild_package` doesn't change with push events for tags, so the code doesn't change either.

danidoni (approved):

rubhanazeem (approved):

src/api/spec/models/workflow/step/branch_package_step_spec.rb
```diff
@@ -284,6 +284,57 @@
...
+
+        it 'does not report back to the SCM' do
+          allow(SCMStatusReporter).to receive(:new)
+          subject.call
+          expect(SCMStatusReporter).not_to have_received(:new)
+        end
+
+        it_behaves_like 'failed when source_package does not exist'
+        it_behaves_like 'failed without branch permissions'
+        it_behaves_like 'fails with insufficient write permission on target project'
```
rubhanazeem, line 336: We talk about not using shared context but I realize it's already implemented and refactoring could be time-consuming.

saraycp (approved):

PR #11834, opened 2021-11-10
Add missing parenthesis in if condition

eduardoj:
Fixes the syntax error introduced in #11829.


M0ses:
discussed with @eduardoj - LGTM

M0ses (approved):

PR #11836, opened 2021-11-11
[backend] bs_worker: implement needspubkeyforbuild

### LINKED ISSUE #9854: optionally provide gpg pubkey to spec files
In order to build custom installation medias in forked projects, one needs to make the gpg signkey of the project available to eg. installation-images and skelcd. For the main distro openSUSE-build-key does that. When branching installation-images it would be nice if it would pick up the key automatically too. So maybe something analogous to #needssslcertforbuild would be useful?

lnussel:
fixes #9854


lnussel:
I don't have any deployment, so this is untested :)

mlschroe (changes requested):
I don't think you should try to get the pubkey with `$server/getsslcert`

mlschroe:
The downside is of course that the pubkey used for publishing may not need to be the pubkey for building.
We have the same problem with the ssl certs, where we need to do config hacks so that maintenance updates get the final cert and not the openSUSE:Maintenance cert.

lnussel:
right. was meant to be getsignkey. c&p error :-/

lnussel:
@mlschroe ?

PR #11837, opened 2021-11-11
Turn Workflow#event_matches_event_filter into an ActiveModel::Validator

danidoni:


danidoni:
[Turn Workflow#event_matches_event_filter? into a ActiveModel::Validator](https://trello.com/c/4U9DYiNb/1799-turn-workfloweventmatcheseventfilter-into-a-activemodelvalidator)

src/api/app/validators/workflow_event_filter_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+    # FIXME: replace private method calls
+    if workflow.send(:supported_filters).key?(:event)
+      # The filters method is needed for the branch_matches_branches_filter that will eventually be a validator.
+      workflow.errors.add(:base, "Workflow filter not supported: #{workflow.filters[:event]}") unless
+        SUPPORTED_FILTER_EVENTS.include?(workflow.filters[:event])
+    else
+      workflow.errors.add(:base, 'Workflow filter not present')
+    end
+
+    false
```
dmarcoux, line 14: Returning a boolean isn't needed for a custom validator. This line can be removed.

src/api/app/validators/workflow_event_filter_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+  SUPPORTED_FILTER_EVENTS = ['push', 'pull_request'].freeze
+
+  def validate(workflow)
+    # FIXME: replace private method calls
+    if workflow.send(:supported_filters).key?(:event)
+      # The filters method is needed for the branch_matches_branches_filter that will eventually be a validator.
+      workflow.errors.add(:base, "Workflow filter not supported: #{workflow.filters[:event]}") unless
+        SUPPORTED_FILTER_EVENTS.include?(workflow.filters[:event])
+    else
+      workflow.errors.add(:base, 'Workflow filter not present')
```
dmarcoux, line 11: It's totally fine to not have any filter, so this should be removed.

src/api/app/validators/workflow_event_filter_validator.rb
```diff
@@ -0,0 +1,16 @@
+class WorkflowEventFilterValidator < ActiveModel::Validator
+  SUPPORTED_FILTER_EVENTS = ['push', 'pull_request'].freeze
+
+  def validate(workflow)
+    # FIXME: replace private method calls
+    if workflow.send(:supported_filters).key?(:event)
+      # The filters method is needed for the branch_matches_branches_filter that will eventually be a validator.
+      workflow.errors.add(:base, "Workflow filter not supported: #{workflow.filters[:event]}") unless
```
dmarcoux, line 8: Error messages could be clearer if we split this up per event. So for example, if a webhook is for a push event, but the workflow should only run for pull requests, the error message could be quite specific like:

`Workflow $WORKFLOW_NAME ignored since the event filter 'push' doesn't match the webhook event.`

src/api/spec/validators/workflow_event_filter_validator_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+    end
+
+    context 'there are no event filters' do
+      let(:filters) { {} }
+
+      subject { Workflow.new(workflow_instructions: yaml) }
+
+      it 'is not valid and has an error message' do
+        expect(subject).not_to(be_valid)
```
dmarcoux, line 33: This will also call all other validation code of the `Workflow` model. So even if the custom validator doesn't change, its specs could break due to changes in other validation code. I suggest using a fake model like we did in [this spec](https://github.com/openSUSE/open-build-service/blob/4ce0dfe110d023e90c3673fa9d673391490425e8/src/api/spec/validators/workflow_filters_validator_spec.rb#L4-L10) for another custom validator.

dmarcoux (changes requested):

dmarcoux:
As discussed with @danidoni, I realized that we cannot refactor this code to add validation errors. We should leave the code as it is. Here's why this refactoring shouldn't happen.

A token is called and its workflows are validated. Any validation error is returned to the user in order to help them debug their faulty workflows configuration. The validation errors are gathered and sent back to the user through the webhook response in the [TriggerWorkflowController](https://github.com/openSUSE/open-build-service/blob/4ce0dfe110d023e90c3673fa9d673391490425e8/src/api/app/controllers/trigger_workflow_controller.rb#L9-L15). If we add a validation error when the webhook event doesn't match the event filter, webhook responses will be a _400_ whenever this happens. This could confuse users if they think that somehow their workflows configuration is wrong, even though it's working as expected.

As an extra note, it would be great to somehow inform users when a workflow is skipped if the webhook event doesn't match the event filter. This is however much more work and we shouldn't do this now.

PR #11838, opened 2021-11-11
Show token id after creation

saraycp:
I was tempted to refactor the code to display a kind of a show view right after the token is created. Then the view wouldn't contain read-only text fields but text values.

However, I opted to follow the easy path and simply add the ID value in the current version of the code. The refactoring can be done in the near future.

Screenshot_2021-11-11 Open Build Service(3)


eduardoj:
Nitpick: I would put the new "Id: $VALUE" text over the horizontal line. It would make more obvious, that a new field has been added.

saraycp:
> Nitpick: I would put the new "Id: $VALUE" text over the horizontal line. It would make more obvious, that a new field has been added.

Attending to this comment, I have introduced some changes. It looks like this now:

Screenshot_2021-11-11 Open Build Service(3)

danidoni (approved):

eduardoj (approved):

PR #11840, opened 2021-11-11
WIP: Add WorkflowBranchesFiltersValidator to Workflow model

vpereira:
TODO:

- [ ] add specs

### DIFF
--- a/src/api/app/models/workflow.rb
+++ b/src/api/app/models/workflow.rb
@@ -19,12 +19,12 @@ class Workflow
 
   validates_with WorkflowStepsValidator
   validates_with WorkflowFiltersValidator
+  validates_with WorkflowBranchesFiltersValidator
 
   def call
     # TODO: This could be in a custom validator WorkflowEventFilterValidator
     return unless event_matches_event_filter?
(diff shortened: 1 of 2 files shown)


dmarcoux:
As discussed with @vpereira, I realized that we cannot refactor this code to add validation errors. We should leave the code as it is. Here's why this refactoring shouldn't happen.

A token is called and its workflows are validated. Any validation error is returned to the user in order to help them debug their faulty workflows configuration. The validation errors are gathered and sent back to the user through the webhook response in the [TriggerWorkflowController](https://github.com/openSUSE/open-build-service/blob/4ce0dfe110d023e90c3673fa9d673391490425e8/src/api/app/controllers/trigger_workflow_controller.rb#L9-L15). If we add a validation error when the target branch contained in the webhook payload doesn't match the branches filter, webhook responses will be a _400_ whenever this happens. This could confuse users if they think that somehow their workflows configuration is wrong, even though it's working as expected.

As an extra note, it would be great to somehow inform users when a workflow is skipped if the target branch contained in the webhook payload doesn't match the branches filter. This is however much more work and we shouldn't do this now.

PR #11841, opened 2021-11-11
Track workflow runs

rubhanazeem:
TODO:

1. [x] Initialize workflow run
2. [x] Track workflow run (success/fail)
3. [x] Add specs

### DIFF
--- a/src/api/app/controllers/trigger/errors.rb
+++ b/src/api/app/controllers/trigger/errors.rb
@@ -7,12 +7,6 @@ module Trigger::Errors
           'No valid token found'
   end
 
-  class BadScmHeaders < APIError
-    setup 'bad_request',
-          400,
-          'Could not find the required HTTP request headers X-GitHub-Event or X-Gitlab-Event'
-  end
-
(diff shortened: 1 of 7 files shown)


src/api/app/controllers/source_attribute_controller.rb
```diff
@@ -68,7 +68,7 @@ def update
 
       attrib.container = @attribute_container
 
-      raise APIError, message: attrib.errors.full_messages.join('\n'), status: 400 unless attrib.valid?
+      raise APIError, attrib.errors.full_messages.join('\n') unless attrib.valid?
```
rubhanazeem, line 71: This is because we added the initializer in the APIError. This way of raising an error was working but it's not the standard way of raising an exception according to the documentation.

dmarcoux: This change isn't needed anymore.

rubhanazeem: Done.

PR #11848, opened 2021-11-12
Use a short commit sha to report back

danidoni:
When we report back the results of some of the workflows, the commit sha
is too big to fit into the space reserved for the PR checks.

This change cuts the commit SHA to be 7 characters long. That way it
fits better in the screen.

**Before**
image

**After**
image


src/api/app/models/workflow/step.rb
```diff
@@ -64,14 +66,19 @@ def source_project_name
...
     package_name = step_instructions[:target_package] || source_package_name
 
     case
     when scm_webhook.pull_request_event?
       package_name
     when scm_webhook.push_event?
-      "#{package_name}-#{scm_webhook.payload[:commit_sha]}"
+      commit_sha = scm_webhook.payload[:commit_sha]
+      if short_commit_sha
+        "#{package_name}-#{commit_sha[0..SHORT_COMMIT_SHA_LENGTH]}"
```
dmarcoux, line 78: With the following approach, we can set `SHORT_COMMIT_SHA_LENGTH` to `7`:

```suggestion
        "#{package_name}-#{commit_sha.slice(0, SHORT_COMMIT_SHA_LENGTH)}"
```

src/api/app/models/workflow/step.rb
```diff
@@ -1,6 +1,8 @@
 class Workflow::Step
   include ActiveModel::Model
 
+  SHORT_COMMIT_SHA_LENGTH = 6
```
dmarcoux, line 4: The value should be `7`. See my other comment on how to make this work with `7`.

src/api/app/models/workflow/step.rb
```diff
@@ -64,14 +66,19 @@ def source_project_name
     step_instructions[:source_project]
   end
 
-  def target_package_name
+  def target_package_name(short_commit_sha: true)
```
dmarcoux, line 69: With the default being a short commit SHA, it could potentially break the `branch_package` and `link_package` steps if there are commits with the SHA in different SCM repositories. Two packages in a project could then have the same name, this would raise an exception.

danidoni: Yeah, I noticed this before. That's why I marked at draft again. I have to push more changes. :+1:

src/api/app/models/workflow/step.rb
danidoni: Good catch!

dmarcoux: This still needs to happen.

danidoni:
[Send back a shorter commit SHA in package names](https://trello.com/c/G2rlQuMI/1796-send-back-a-shorter-commit-sha-in-package-names)

src/api/app/models/workflow/step.rb
```diff
@@ -142,7 +149,7 @@ def report_to_scm(workflow_filters)
       # TODO: Fix n+1 queries
       workflow_architectures(repository, workflow_filters).each do |architecture|
         # We cannot report multibuild flavors here... so they will be missing from the initial report
-        SCMStatusReporter.new({ project: target_project_name, package: target_package_name, repository: repository.name, arch: architecture.name },
+        SCMStatusReporter.new({ project: target_project_name, package: target_package_name(short_commit_sha: true), repository: repository.name, arch: architecture.name },
```
dmarcoux, line 152: Setting `short_commit_sha` to `true` also needs to happen in `ReportToScmJob`. The package is in `@event.payload`: https://github.com/openSUSE/open-build-service/blob/0cb661478d70ecda31c6f9b6c052144c53e7b9bf/src/api/app/jobs/report_to_scm_job.rb#L11

dmarcoux (changes requested):

src/api/app/models/workflow/step.rb
saraycp: Exactly, I already talked about this with Dani. We are reporting with a short commit sha the first time we call `SCMStatusReporter` to set the "pending" state. The following reports, after a build is finished, will contain a long commit sha.

saraycp (changes requested):

src/api/app/models/workflow/step.rb
danidoni: Fixed, now

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -44,7 +44,7 @@ def create_branched_package
     begin
       BranchPackage.new({ project: source_project_name, package: source_package_name,
                           target_project: target_project_name,
-                          target_package: target_package_name }).branch
+                          target_package: target_package_name(short_commit_sha: true) }).branch
```
dmarcoux, line 47: This is wrong. It will create the package with name  containing the short commit SHA. It's only the report which needs to change, so the ReportoScmJob as I pointed out in my previous comment.

src/api/app/models/workflow/step.rb
```diff
@@ -14,6 +16,7 @@ def call(_options)
     raise AbstractMethodCalled
   end
 
+  # FIXME: Rename this function because its clashing with names below
```
dmarcoux, line 19: I don't agree with this comment. If we turn `workflow_repositories` into a query object, this wouldn't be an issue anymore.

src/api/app/models/workflow/step.rb
```diff
@@ -66,14 +68,19 @@ def source_project_name
     step_instructions[:source_project]
   end
 
-  def target_package_name
+  def target_package_name(short_commit_sha: false)
```
hennevogel, line 71: Hm, why is this an option?

danidoni: Because there are places where we need the target package name but we don't want to get the short commit sha appended but the full one.

hennevogel:
What's with all the cassettes?

danidoni:
> What's with all the cassettes?

I changed some of the requests because I added the long commit SHA in the specs, to test the cutting. Some of those specs are actually deletions because no longer apply and some of them are re-recording of the specs with the new long SHA.

src/api/spec/services/scm_status_reporter_spec.rb
```diff
@@ -69,5 +69,63 @@
...
     end
+
+    context 'when sending a report back to GitHub' do
+      let(:event_payload) do
+        { project: 'home:danidoni', package: 'hello_world',
+          repository: 'openSUSE_Tumbleweed', arch: 'x86_64' }
+      end
+      let(:event_subscription_payload) do
+        { scm: 'github', short_package_name: 'hello_world-1234567',
+          target_repository_full_name: 'danidoni/hello_world', commit_sha: '1234567' }
```
saraycp, line 80:
```suggestion
          target_repository_full_name: 'danidoni/hello_world', commit_sha: '123456789' }
```

It doesn't make any difference in the spec, but can be confusing to have a value like the short commit sha when the `commit_sha` is always stored in the long version.

src/api/spec/services/scm_status_reporter_spec.rb
```diff
@@ -69,5 +69,63 @@
...
+        }
+      end
+      let(:octokit_client) { instance_spy('Octokit::Client', create_status: true) }
+
+      subject { scm_status_reporter.call }
+
+      it 'sends a short commit sha' do
+        allow(Octokit::Client).to receive(:new).and_return(octokit_client)
+        subject
+        expect(octokit_client).to have_received(:create_status).with('danidoni/hello_world', '1234567', state, status_options)
```
saraycp, line 98:
```suggestion
        expect(octokit_client).to have_received(:create_status).with('danidoni/hello_world', '123456789', state, status_options)
```

src/api/spec/services/scm_status_reporter_spec.rb
```diff
@@ -69,5 +69,63 @@
...
+    end
+
+    context 'when sending a report back to GitLub' do
+      let(:event_payload) do
+        { project: 'home:danidoni', package: 'hello_world',
+          repository: 'openSUSE_Tumbleweed', arch: 'x86_64' }
+      end
+      let(:event_subscription_payload) do
+        { scm: 'gitlab', short_package_name: 'hello_world-1234567',
+          project_id: '26_212_710', commit_sha: '1234567' }
```
saraycp, line 109:
```suggestion
          project_id: '26_212_710', commit_sha: '123456789' }
```

src/api/spec/services/scm_status_reporter_spec.rb
```diff
@@ -69,5 +69,63 @@
...
+        }
+      end
+      let(:gitlab_instance) { instance_spy('Client', update_commit_status: true) }
+
+      subject { scm_status_reporter.call }
+
+      it 'sends a short commit sha' do
+        allow(Gitlab).to receive(:client).and_return(gitlab_instance)
+        subject
+        expect(gitlab_instance).to have_received(:update_commit_status).with('26_212_710', '1234567', state, status_options)
```
saraycp, line 127:
```suggestion
        expect(gitlab_instance).to have_received(:update_commit_status).with('26_212_710', '123456789', state, status_options)
```

saraycp (approved):

PR #11850, opened 2021-11-12
Drop response_url from workflow_runs

rubhanazeem:
Initially, we thought that we needed the field response_url to report
back to SCM.  We just realise we are not responding back to SCM.

### DIFF
--- a/src/api/app/models/workflow_run.rb
+++ b/src/api/app/models/workflow_run.rb
@@ -1,5 +1,4 @@
 class WorkflowRun < ApplicationRecord
-  validates :response_url, length: { maximum: 255 }
   validates :request_headers, :request_payload, :status, presence: true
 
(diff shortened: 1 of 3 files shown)


danidoni (dismissed):

krauselukas (changes requested):
@rubhanazeem I would suggest to first ignore the column as suggested here https://github.com/ankane/strong_migrations#removing-a-column

krauselukas (approved):
@rubhanazeem as we talked, the column was never used in the code, so there shouldnt be anything cached by active record. So its fine to migrate this without ignoring the column upfront

dmarcoux:
This column is to track the URL which we use to report back to the SCM. This isn't about reporting back to the SCM, it's about letting users know where exactly we reported back to. This PR should be reverted.

PR #11857, opened 2021-11-15
Handle Octokit::InvalidRepository on YAMLDownloader

vpereira:
Give users the right feedback in case the repository doesnt exist
anymore when trying to download the `workflows.yml`

The issue https://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/60bff83684bf4607a946599e has two root causes (to investigate them, just navigate through them, clicking on Older/Newer): 

a) When the repository doesn't exist while downloading the YAML file
b) When the repository was removed, but OBS still reporting to github, i.e PR is open, but the repository was deleted. 

This PR fixes just the first case

### DIFF
--- a/src/api/app/models/token/errors.rb
+++ b/src/api/app/models/token/errors.rb
@@ -5,6 +5,10 @@ module Token::Errors
     setup 404
   end
 
+  class NonExistentRepository < APIError
+    setup 404
+  end
+
   class NonExistentWorkflowsFile < APIError
     setup 404
   end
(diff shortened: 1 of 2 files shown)


vpereira:
[Catch error for invalid GitHub repositories in Workflows::YAMLDownloader](https://trello.com/c/Oc1PJo0j/1797-catch-error-for-invalid-github-repositories-in-workflowsyamldownloader)

krauselukas (approved):

PR #11868, opened 2021-11-16
Fix inconsistent target project name

### LINKED ISSUE #11854: Workflow steps target_project_name mismatch
The link_package step uses `:` between org and repo, the configure step uses `-`. 

```diff
home:bs-team:openSUSE:open-build-service:PR-11852
home:bs-team:openSUSE-open-build-service:PR-11852
```

Means you can't use those two steps together.

Screenshot from 2021-11-12 18-11-41

saraycp:
Fixes #11854

The `Workflow::Step::ConfigureRepositories#target_project_name` method was a redefinition of the `Workflow::Step#target_project_name`.

They should be identical except for the YAML key containing the project name. So there is no need to keep both. In fact, this is leading to inconsistencies. For example with the format of the name (using "-" or ":").

Now we only have one definition of the method. The only part we redefine is the project name. For this, we have the `target_project_base_name` method implemented on each step.

### DIFF
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -15,15 +15,17 @@ class Workflow::Step
   end
 
   def target_project_name
-    return step_instructions[:target_project] if scm_webhook.push_event?
+    return target_project_base_name if scm_webhook.push_event?
+
+    return nil unless scm_webhook.pull_request_event?
 
     pr_subproject_name = if scm_webhook.payload[:scm] == 'github'
-                           scm_webhook.payload[:target_repository_full_name].tr('/', ':')
+                           scm_webhook.payload[:target_repository_full_name]&.tr('/', ':')
                          else
-                           scm_webhook.payload[:path_with_namespace].tr('/', ':')
+                           scm_webhook.payload[:path_with_namespace]&.tr('/', ':')
                          end
 
-    "#{step_instructions[:target_project]}:#{pr_subproject_name}:PR-#{scm_webhook.payload[:pr_number]}"
+    "#{target_project_base_name}:#{pr_subproject_name}:PR-#{scm_webhook.payload[:pr_number]}"
   end
 
   def target_package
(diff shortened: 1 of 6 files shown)


src/api/app/models/workflow/step.rb
```diff
@@ -77,6 +77,10 @@ def target_package_name
 
   private
 
+  def target_project_base_name
+    raise NotImplementedError
```
dmarcoux, line 81: We should raise `AbstractMethodCalled` instead of `NotImplementedError`:

```suggestion
    raise AbstractMethodCalled
```

 `NotImplementedError` is for platform limitations like if the OS doesn't support a certain feature. This is documented [here](https://ruby-doc.org/core-2.5.0/NotImplementedError.html).

src/api/spec/models/workflow/step/configure_repositories_spec.rb
```diff
@@ -301,87 +298,7 @@
       end
 
       it 'raises an error due to an inexistent target project' do
-        expect { subject.call }.to raise_error(Project::Errors::UnknownObjectError, "Project not found: #{target_project_name}")
-      end
-    end
-  end
-
-  describe '#target_project_name' do
```
dmarcoux, line 309: With the removal of those specs, the method isn't tested anywhere. We should instead keep those specs, but move them to `src/api/spec/models/workflow/step_spec.rb` (this file doesn't exist yet)

dmarcoux (changes requested):

src/api/app/models/workflow/step.rb
saraycp: done

src/api/spec/models/workflow/step/configure_repositories_spec.rb
saraycp: done

saraycp:
@dmarcoux please have another look. See the new stuff [here](https://github.com/openSUSE/open-build-service/compare/b9f71901419c23f8a25eb0d6729e709775138f5c..29cc1342ea019133366cc248404660cfbfe63534)

dmarcoux (approved):

PR #11870, opened 2021-11-17
Create workflow runs index view

krauselukas:
We are tracking executed workflow runs and the current state of them
through the `WorkflowRun` model.
Now that this data is available for the users we need the corresponding
webui elements in order to access those infos.

TODO:

- [x] add pagination
- [x] add controller and policy specs 
- [x] add the breadcrumbs
- [ ] ~~maybe split the index view into components~~

Example screenshots (data is fake):

### DIFF
--- a/src/api/app/policies/workflow_run_policy.rb
+++ b/src/api/app/policies/workflow_run_policy.rb
@@ -1,0 +1,21 @@
+class WorkflowRunPolicy < ApplicationPolicy
+  class Scope
+    def initialize(user, scope, opts = {})
+      @user = user
+      @scope = scope
+      @opts = opts
+    end
+
+    def resolve
+      token = Token.find_by(id: opts[:token_id])
+      raise Pundit::NotAuthorizedError, 'you are not authorized to access those workflow runs' unless token.user == user
+      raise Pundit::NotAuthorizedError, 'the token is not of type workflow' unless token.type == 'Token::Workflow'
+
+      scope.where(token_id: token.id).order(created_at: :desc)
+    end
+
+    private
+
+    attr_reader :user, :scope, :opts
+  end
+end
--- a/src/api/app/views/webui/workflow_runs/index.html.haml
+++ b/src/api/app/views/webui/workflow_runs/index.html.haml
@@ -6,6 +6,8 @@
     - if @workflow_runs.blank?
       %p There are no workflow runs for this token yet
     - else
+      .text-center
+        %span.ml-3= page_entries_info @workflow_runs, entry_name: ''
       .accordion.pt-3#workflow-runs-accordion
         .card
         - @workflow_runs.each do |workflow_run|
@@ -55,3 +57,4 @@
                   %h3 Response Body
                   %pre.border.p-2#response-body
                     = workflow_run.response_body
+        = paginate @workflow_runs, views_prefix: 'webui'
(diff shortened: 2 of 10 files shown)


krauselukas:
[Recent Workflow Runs in the UI](https://trello.com/c/WnMfaBDl/1794-recent-workflow-runs-in-the-ui)

krauselukas:
Some screenshots of the current state, don't look to close on the data, its just whatever

src/api/app/views/webui/workflow_runs/index.html.haml
```diff
@@ -0,0 +1,54 @@
+- @pagetitle = 'Workflow Runs'
+
+.card
+  .card-body
+    %h3= @pagetitle
+    .accordion.pt-3#workflow-runs-accordion
+      - @workflow_runs.each do |workflow_run|
```
dmarcoux, line 7: We should at least display a message like _This token wasn't triggered yet._ if there are no workflow runs for a token. Otherwise, it looks like something is wrong/broken.

workflow-runs

src/api/app/policies/token_policy.rb
```diff
@@ -45,4 +45,8 @@ def webui_trigger?
   def show?
     webui_trigger?
   end
+
+  def workflow_runs_index?
```
dmarcoux, line 49: In [Webui::Users::TokensController](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/users/tokens_controller.rb), the tokens are already scoped to the logged-in user. I don't see a need to redo a policy check for this. The link can be displayed if the token is a `Token::Workflow` and if somehow a user tries to access the workflow runs page for a non-workflow token, `WorkflowRunPolicy` will prevent that.

src/api/app/policies/workflow_run_policy.rb
```diff
@@ -0,0 +1,21 @@
...
+      @scope = scope
+      @opts = opts
+    end
+
+    def resolve
+      token = Token.find_by(id: opts[:token_id])
+      raise Pundit::NotAuthorizedError, 'the token user not authorized to access the workflow runs' unless token.user == user
+      raise Pundit::NotAuthorizedError, 'the token is not of type workflow' unless token.type == 'Token::Workflow'
+
+      scope.where(token_id: token.id)
```
rubhanazeem, line 14: Do you want to add `order(created_at: :desc)` or `order(id: :desc)`, so we show the last workflow_run at the top?

src/api/config/routes/webui_routes.rb
```diff
@@ -392,4 +392,6 @@
       end
     end
   end
+
+  resources :workflow_runs, only: [:index], controller: 'webui/workflow_runs'
```
rubhanazeem, line 396: Would it be a good idea to have a nested route, something like `token_id/workflow_runs`?

src/api/app/policies/workflow_run_policy.rb
krauselukas: good point, applied!

src/api/config/routes/webui_routes.rb
krauselukas: yes, I moved it under the tokens resources :)

src/api/app/policies/token_policy.rb
krauselukas: You are right, no need to double check, I removed the policy and only check for the token type in order to show the link only for the workflow tokens

rubhanazeem (approved):

krauselukas (commented):
@vpereira as discussed, got rid of some duplicated code in the ~~controller~~ policy spec

PR #11873, opened 2021-11-18
Support push events for tags in ScmWebhook

dmarcoux:


src/api/spec/models/workflow/step/branch_package_step_spec.rb
```diff
@@ -253,15 +253,16 @@
         end
       end
 
-      context 'for a push event' do
+      context 'with a push event for a commit' do
```
dmarcoux, line 256: Cassette files were renamed due to changes like this one.

src/api/spec/models/workflow/step/branch_package_step_spec.rb
```diff
@@ -253,15 +253,16 @@
...
         let(:scm_webhook) do
           ScmWebhook.new(payload: {
                            scm: 'github',
                            event: 'push',
                            target_branch: 'main',
                            source_repository_full_name: 'reponame',
                            commit_sha: commit_sha,
-                           target_repository_full_name: 'openSUSE/open-build-service'
+                           target_repository_full_name: 'openSUSE/open-build-service',
+                           ref: 'refs/heads/branch_123'
```
dmarcoux, line 265: Adding the `ref` is needed only for GitHub since it's the only way to tell if a push event is for commits or tags. GitLab doesn't need this change since it has 2 separate push events for commits and tags.

saraycp (approved):

PR #11876, opened 2021-11-18
Support push events for tags in Workflow::Step#target_{project,package}_name

dmarcoux:

### DIFF
--- a/src/api/spec/models/workflow/step_spec.rb
+++ b/src/api/spec/models/workflow/step_spec.rb
@@ -1,17 +1,73 @@
 require 'rails_helper'
 
 RSpec.describe Workflow::Step do
-  let(:user) { create(:confirmed_user, :with_home, login: 'Iggy') }
-  let(:token) { create(:workflow_token, user: user) }
-  let(:step) do
-    Class.new(described_class) do
-      def target_project_base_name
-        'OBS:Server:Unstable'
+  describe '#target_package_name' do
+    subject { described_class.new(step_instructions: step_instructions, scm_webhook: scm_webhook).send(:target_package_name) }
+
+    context 'for a pull request_event when target_package is in the step instructions' do
+      let(:step_instructions) { { target_package: 'hello_world' } }
+      let(:scm_webhook) { ScmWebhook.new(payload: { scm: 'github', event: 'pull_request' }) }
+
+      it 'returns the value of target_package' do
+        expect(subject).to eq('hello_world')
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


src/api/spec/models/workflow/step_spec.rb
```diff
@@ -32,9 +88,7 @@ def target_project_base_name
     let(:scm_webhook) { ScmWebhook.new(payload: payload) }
 
     subject do
-      step.new(step_instructions: step_instructions,
-               scm_webhook: scm_webhook,
-               token: token).target_project_name
```
dmarcoux, line 37: Passing a token isn't necessary to make the specs pass, so I removed it.

src/api/spec/models/workflow/step_spec.rb
```diff
@@ -1,17 +1,73 @@
...
 
 RSpec.describe Workflow::Step do
-  let(:user) { create(:confirmed_user, :with_home, login: 'Iggy') }
-  let(:token) { create(:workflow_token, user: user) }
-  let(:step) do
-    Class.new(described_class) do
-      def target_project_base_name
-        'OBS:Server:Unstable'
+  describe '#target_package_name' do
+    subject { described_class.new(step_instructions: step_instructions, scm_webhook: scm_webhook).send(:target_package_name) }
```
dmarcoux, line 5: I don't see why this should be public or extracted to a new class only to be tested. I prefer to use `send` to test this protected method.

dmarcoux: Making a method public just to test it, I don't see how this makes the code better. Creating a new class from that method is too much abstraction. I prefer to keep the code as it is and test with `send`.

vpereira: > Making a method public just to test it, I don't see how this makes the code better.

That's IMO, the wrong assumption. The method being private and having to be directed tested is an indication that the design isn't right. Here we highlight the importance of using unit tests to expose design flaws such as this :+1:

dmarcoux: We don't call that method anywhere beside in other methods of the model itself, so making the method public is bonkers.

vpereira: AFAIK we don't define a model API based on "how many methods calls it". The method `target_package_name` could/should be very well a public method from the class `Step`, like we have everywhere else i.e: `User#home_project` and `User#home_project_name`.

danidoni (approved):

PR #11877, opened 2021-11-18
Support push events for tags in event filter of OBS workflows

dmarcoux:

Documentation for this:
- [Developer](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration#event-filter)
- [User](https://github.com/openSUSE/obs-docu/pull/205)

### DIFF
--- a/src/api/app/models/workflow.rb
+++ b/src/api/app/models/workflow.rb
@@ -29,7 +29,7 @@ class Workflow
       destroy_target_projects
     when scm_webhook.reopened_pull_request?
       restore_target_projects
-    when scm_webhook.new_pull_request?, scm_webhook.updated_pull_request?, scm_webhook.push_event?
+    when scm_webhook.new_pull_request?, scm_webhook.updated_pull_request?, scm_webhook.push_event?, scm_webhook.tag_push_event?
       steps.each do |step|
         step.call({ workflow_filters: filters })
       end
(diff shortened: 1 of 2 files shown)


vpereira (approved):

src/api/app/models/workflow.rb
```diff
@@ -29,7 +29,7 @@ def call
       destroy_target_projects
     when scm_webhook.reopened_pull_request?
       restore_target_projects
-    when scm_webhook.new_pull_request?, scm_webhook.updated_pull_request?, scm_webhook.push_event?
+    when scm_webhook.new_pull_request?, scm_webhook.updated_pull_request?, scm_webhook.push_event?, scm_webhook.tag_push_event?
```
dmarcoux, line 32: This change makes the specs pass. However, push events for tags still won't trigger a workflow since `ScmWebhookEventValidator` doesn't accept those events. This is the final step to enable push events for tags.

krauselukas (approved):

src/api/app/models/workflow.rb
```diff
@@ -74,10 +74,17 @@ def supported_filters
 
   def event_matches_event_filter?
     return true unless supported_filters.key?(:event)
-    return true if filters[:event] == 'push' && scm_webhook.push_event?
-    return true if filters[:event] == 'pull_request' && scm_webhook.pull_request_event?
 
-    false
+    case filters[:event]
```
dmarcoux, line 78: I noticed that there was a typo here since the last approval from @krauselukas.

I changed this line from `case filters[:event]` to `case filters[:events]`.

rubhanazeem (approved):

PR #11882, opened 2021-11-22
Reintroduce checking for `Token` string

eduardoj:
Token authentication can be performed by passing a token string through an 'Authorization' HTTP header. The format to pass this token follows this expression:

`Token TOKEN_STRING`

This changed in commit deea2dfcd9eea704eaed0532793160f7c7ba88b6, where the check for the 'Token' string was removed.

Reintroduce the check for the 'Token' string, as was initially defined, and is described in the API documentation.

### DIFF
--- a/src/api/app/services/trigger_controller_service/token_extractor.rb
+++ b/src/api/app/services/trigger_controller_service/token_extractor.rb
@@ -16,9 +16,15 @@ module TriggerControllerService
 
     private
 
+    def extract_token_string_from_authorization_header
+      http_auth = @http_request.env['HTTP_AUTHORIZATION']
+
+      http_auth.slice(6..-1) if http_auth.present? && http_auth[0..4] == 'Token'
+    end
+
     def extract_auth_token_from_headers
       auth_token = @http_request.env['HTTP_X_GITLAB_TOKEN'] ||
-                   @http_request.env['HTTP_AUTHORIZATION'].to_s.slice(6..-1)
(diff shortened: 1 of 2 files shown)


src/api/app/services/trigger_controller_service/token_extractor.rb
```diff
@@ -16,9 +16,15 @@ def call
 
     private
 
+    def extract_token_string_from_authorization_header
+      http_auth = @http_request.env['HTTP_AUTHORIZATION']
+
+      http_auth.slice(6..-1) if http_auth.present? && http_auth[0..4] == 'Token'
```
dmarcoux, line 22: Does it have to be strict by enforcing the use of `Token`?

eduardoj: It seems to me that it was a change that was not planned. The current documentation doesn't reflect that change, either: https://build.opensuse.org/apidocs/index#128

The current alternative accepts any combination of five first characters. If we go for an option of using an authentication header, we should follow a `Authorization: <type> <credentials>` convention, as described [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication). It could colide with some other authentication scheme which a type of five characters.

dmarcoux: Alright, let's enforce what we have in the API documentation.

dmarcoux (approved):

PR #11884, opened 2021-11-23
[api] fix project copy

adrianschroeter:
take care about
- optional repository flags
- project links


hennevogel:
By fix you mean extend with those features right?

adrianschroeter:
no fixing, as it is a misbehaviour not to copy these parts and leads to broken setup.

hennevogel (approved):

PR #11886, opened 2021-11-23
[backend] BSPublisher/Container.pm ... collect all errors, dump at end

bugfinder:


src/backend/BSPublisher/Container.pm
```diff
@@ -463,6 +463,7 @@ sub upload_to_notary {
   mkdir_p($uploaddir);
   unlink($pubkeyfile);
   writestr($pubkeyfile, undef, $pubkey);
+  my $failed_uploads;
   for my $uploadkey (sort keys %$notary_uploads) {
```
mlschroe, line 467: Why not %failed_uploads?

mlschroe (changes requested):

src/backend/BSPublisher/Container.pm
bugfinder: fine now?

mlschroe (approved):

PR #11887, opened 2021-11-24
Autofocus into the search input field

danidoni:
So we can just start searching for something when the page loads instead of having to click on it


rubhanazeem:
@danidoni can you please rerun the failing tests?

rubhanazeem (approved):

PR #11903, opened 2021-11-26
Add source project name Apidoc endpoint

danidoni:
This is the new ApiDoc endpoint that will eventually replace the current endpoint here https://build.opensuse.org/apidocs/index#44

### DIFF
--- a/src/api/public/apidocs-new/OBS-v2.10.50.yaml
+++ b/src/api/public/apidocs-new/OBS-v2.10.50.yaml
@@ -4,20 +4,20 @@ servers:
   - url: '/'
 
 tags:
-  - name: General Information
   - name: Announcements
-  - name: Attributes
   - name: Attribute Namespaces
+  - name: Attributes
   - name: Build
   - name: Configuration
   - name: Distributions
+  - name: General Information
   - name: Groups
   - name: Issue Trackers
   - name: Notifications
   - name: Person
   - name: Published Binaries
   - name: Requests
-  - name: Sources
+  - name: Sources - Projects
   - name: Trigger
   - name: Workers
 
@@ -176,6 +176,8 @@ paths:
     $ref: 'paths/source_branch.yaml'
   /source?cmd=orderkiwirepos:
     $ref: 'paths/source_orderkiwirepos.yaml'
+  /source/{project_name}?cmd=addchannels:
+    $ref: 'paths/source_project_name_cmd_addchannels.yaml'
 
   /trigger/{operation}:
     $ref: 'paths/trigger_operation.yaml'
--- a/src/api/public/apidocs-new/paths/source_project_name_cmd_addchannels.yaml
+++ b/src/api/public/apidocs-new/paths/source_project_name_cmd_addchannels.yaml
@@ -1,0 +1,33 @@
+post:
+  summary: Add channels and extend repository list.
+  description: |
+    Add channels for each one of the provided project packages.
+    If the mode parameter is provided, that mode is added to the package channel.
+    The mode parameter can be:
+      - 'skip_disabled'
+      - 'enable_all'
+      - 'add_disabled' being the default.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - in: path
+      name: mode
+      schema:
+        type: string
+        enum:
+          - skip_disabled
+          - enable_all
+          - add_disabled
+      required: false
+      description: The channel will be added to the package using this mode.
+      example: "add_disabled"
+  responses:
+    '200':
+      $ref: '../components/responses/succeeded.yaml'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      $ref: '../components/responses/unknown_project.yaml'
+  tags:
+    - Sources - Projects


danidoni:

src/api/public/apidocs-new/paths/source_project_name.yaml
```diff
@@ -0,0 +1,23 @@
+post:
+  summary: Add channels to a source
```
dmarcoux, line 2: This is only true if the command is `addchannels`. Otherwise, this endpoint does something completely different.

danidoni: Yeah, true

danidoni: I've changed the apidoc so it makes reference only to this command, I think it makes much sense now.

src/api/public/apidocs-new/paths/source_project_name_cmd_addchannels.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - $ref: '../components/parameters/project_name.yaml'
+  responses:
+    '200':
+      $ref: '../components/responses/succeeded.yaml'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      $ref: '../components/responses/unknown_project.yaml'
+  tags:
+    - Source
```
saraycp, line 16: IIRC we planned not to include all the source endpoints under the same tag, but to have many of them for clarity:

- Sources (for global commands)
- Sources - Projects
- Sources - Packages
- Sources - Files
- **Sources - Project Command (the current one)**
- Sources - Package Command

But @eduardoj could confirm.

Anyway, if we finally keep them all under the same tag, it should be plural: `Sources`.

eduardoj: Let's use four groups:
- Sources (for global commands)
- Sources - Projects
- Sources - Packages
- Sources - Files

This will mimic the sections from the current apidocs, and and a new section for global commands. I hope this makes things more clear.

saraycp:
```suggestion
    - Sources - Projects
```

danidoni: I'll move the Project Command under Sources - Projects. Is that ok for you @saraycp ?

src/api/public/apidocs-new/paths/source_project_name_cmd_addchannels.yaml
```diff
@@ -0,0 +1,16 @@
+post:
+  summary: Add channels to a source
```
eduardoj, line 2:
```suggestion
  summary: Add channels to a source.
```
Let's finish sentences with a dot.

src/api/public/apidocs-new/paths/source_project_name_cmd_addchannels.yaml
```diff
@@ -0,0 +1,16 @@
...
+  description: Add channels to a source
```
eduardoj, line 3:
```suggestion
```
If the description is the same as the summary, it doesn't add value. Better remove it.

src/api/public/apidocs-new/paths/source_project_name_cmd_addchannels.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - $ref: '../components/parameters/project_name.yaml'
+  responses:
+    '200':
+      $ref: '../components/responses/succeeded.yaml'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      $ref: '../components/responses/unknown_project.yaml'
+  tags:
+    - Sources - Projects
```
eduardoj, line 16: The tag `Sources - Projects` is missing in the list of tags in the `OBS-v2.10.50.yaml` file. We should only add a `Sources - Projects` entry in that list.

danidoni: I've added the tag to the `OBS-v2.10.50.yaml` file. Let me know if it's ok now.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -19,6 +19,7 @@ tags:
   - name: Request
   - name: Trigger
   - name: Workers
+  - name: Sources - Projects
```
eduardoj, line 22:
```suggestion
  - name: Request
  - name: Sources - Projects
  - name: Trigger
  - name: Workers
```
Just to maintain the list alphabetically sorted.

src/api/public/apidocs-new/paths/source_project_name_cmd_addchannels.yaml
eduardoj: Now it's added. Fine. Just only, insert the entry so the list keeps the alphabeticall order.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -4,20 +4,20 @@ servers:
...
   - name: Distributions
+  - name: General Information
   - name: Groups
   - name: Issue Trackers
   - name: Notifications
   - name: Person
   - name: Published Binaries
-  - name: Requests
-  - name: Sources
+  - name: Request
```
rubhanazeem, line 19: Why is it changed to `Request`? Is it a typo? :thinking:

saraycp (changes requested):

src/api/public/apidocs-new/OBS-v2.10.50.yaml
danidoni: It is, sorry :facepalm: I resolved a conflict and this slipped in

rubhanazeem (approved):

saraycp (approved):

PR #11912, opened 2021-11-30
Add source project name copy cmd apidoc endpoint

danidoni:
This PR documents the `POST /source/<project>?cmd=copy` endpoint
with the new OpenAPI specification.

The `copy` command copies the entire project from the source project
provided by `oproject` to the target project provided by `project`.

The params `makeolder` and `makeoriginolder` update the relevant vrev.
The param `withbinaries` ensures we copy also the binaries.

### DIFF
--- a/src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
+++ b/src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
@@ -1,0 +1,70 @@
+post:
+  summary: Copy the entire project.
+  description: |
+    The `copy` command copies the entire project from the source project provided by `oproject` to the target project provided by `project`.
+
+    The params `makeolder` and `makeoriginolder` update the relevant vrev.
+    The param `withbinaries` ensures we copy also the binaries.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - in: query
+      name: oproject
+      schema:
+        type: string
+      required: true
+      description: Origin project name
+      example: "home:Admin"
+    - in: query
+      name: opackage
+      schema:
+        type: string
+      required: true
+      description: Origin package name
+      example: "ctris"
+    - in: query
+      name: makeolder
+      schema:
+        type: string
+      description: Make target older, the source vrev is bumped by two numbers and target by one. This parameter is interpreted as `true` if present, `false` otherwise.
+      example: 1
+    - in: query
+      name: makeoriginolder
+      schema:
+        type: string
+      description: Make origin older, the source vrev is extended and target is guaranteed to be newer. This parameter is interpreted as `true` if present, `false` otherwise.
+      example: 1
+    - in: query
+      name: withbinaries
+      schema:
+        type: string
+      description: Copies also binaries on copy command. This parameter is interpreted as `true` if present, `false` otherwise.
+      example: 1
+    - in: query
+      name: nodelay
+      schema:
+        type: string
+      description: If the parameter is present, the copying is done right away. If not present, the copying is done as a deferred job.
+      example: 1
+  responses:
+    '200':
+      $ref: '../components/responses/succeeded.yaml'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '403':
+      description: |
+        Forbidden.
+
+        no permission to execute command 'copy'.
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+          example:
+            code: no_permission_to_change
+            summary: no permission to execute command 'copy' ...
+    '404':
+      $ref: '../components/responses/unknown_project.yaml'
+  tags:
+    - Sources - Projects
--- a/src/api/public/apidocs-new/OBS-v2.10.50.yaml
+++ b/src/api/public/apidocs-new/OBS-v2.10.50.yaml
@@ -184,6 +184,8 @@ paths:
 
   /source/{project_name}?cmd=addchannels:
     $ref: 'paths/source_project_name_cmd_addchannels.yaml'
+  /source/{project_name}?cmd=copy:
+    $ref: 'paths/source_project_name_cmd_copy.yaml'
   /source/{project_name}?cmd=createmaintenanceincident:
     $ref: 'paths/source_project_name_cmd_createmaintenanceincident.yaml'
   /source/{project_name}?cmd=createpatchinfo:


danidoni:

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
```diff
@@ -0,0 +1,52 @@
+post:
+  summary: Copy the entire project.
+  description: |
+    Copy the entire project.
```
saraycp, line 4: What about replacing this description with something like you explained in the [PR description](https://github.com/openSUSE/open-build-service/pull/11912#issue-791624647)?

Otherwise, I would remove the description if it's the same as the summary.

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
```diff
@@ -0,0 +1,52 @@
...
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - in: path
```
saraycp, line 9:
```suggestion
    - in: query
```

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
```diff
@@ -0,0 +1,52 @@
...
+      name: oproject
+      schema:
+        type: string
+      required: true
+      description: Origin project name
+      example: "home:Admin"
+    - in: path
```
saraycp, line 16:
```suggestion
    - in:query
```

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
```diff
@@ -0,0 +1,52 @@
...
+      name: opackage
+      schema:
+        type: string
+      required: true
+      description: Origin package name
+      example: "ctris"
+    - in: path
```
saraycp, line 23:
```suggestion
    - in: query
```

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
```diff
@@ -0,0 +1,52 @@
...
+      name: makeolder
+      schema:
+        type: boolean
+      required: false
+      description: Make target older, the source vrev is bumped by two numbers and target by one.
+      example: 1
+    - in: path
```
saraycp, line 30:
```suggestion
    - in: query
```

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
```diff
@@ -0,0 +1,52 @@
...
+      name: makeoriginolder
+      schema:
+        type: boolean
+      required: false
+      description: Make origin older, the source vrev is extended and target is guaranteed to be newer.
+      example: 1
+    - in: path
```
saraycp, line 37:
```suggestion
    - in: query
```

saraycp (changes requested):

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
```diff
@@ -0,0 +1,55 @@
...
+      schema:
+        type: string
+      required: true
+      description: Origin package name
+      example: "ctris"
+    - in: query
+      name: makeolder
+      schema:
+        type: boolean
+      required: false
```
eduardoj, line 30:
```suggestion
      required: false
```
All `required: false` lines could be removed. `required: false` is the default option. It ocurrs in this line and two more lines.

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
```diff
@@ -0,0 +1,52 @@
...
+      schema:
+        type: string
+      description: Make origin older, the source vrev is extended and target is guaranteed to be newer. This parameter is interpreted as `true` if present, `false` otherwise.
+      example: 1
+    - in: query
+      name: withbinaries
+      schema:
+        type: string
+      description: Copies also binaries on copy command. This parameter is interpreted as `true` if present, `false` otherwise.
+      example: 1
```
saraycp, line 43: According to [this line](https://github.com/openSUSE/open-build-service/blob/3743e2a80736cabfa2f714ba8e0c36ec48523ee4/src/api/app/controllers/source_controller.rb#L623) and [this line](https://github.com/openSUSE/open-build-service/blob/3743e2a80736cabfa2f714ba8e0c36ec48523ee4/src/api/app/controllers/source_controller.rb#L630) we only check for the existence of `makeolder`, `makeoriginolder` and `withbinaries` but we don't check their values.

Then having something like `makeolder=0` or `makeolder=bla` will be considered true. Only when the parameter is not passed won't apply it.

However, this is an inconsistency with the backend that is expecting a boolean value.
Testing with  `makeolder=bla`,  this is the error message we get: `copy failed: not boolean`.

We can declare this as a `string` for now and postpone some research with [this card](https://trello.com/c/i2q5t19Q/2446-research-about-parameters-acting-as-parameters).

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
```diff
@@ -0,0 +1,52 @@
...
+  responses:
+    '200':
+      $ref: '../components/responses/succeeded.yaml'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      $ref: '../components/responses/unknown_project.yaml'
```
saraycp, line 50: Sorry I didn't catch this before. There are some 403 that can happen and are not documented. The errors [CmdExecutionNoPermission](https://github.com/openSUSE/open-build-service/blob/3743e2a80736cabfa2f714ba8e0c36ec48523ee4/src/api/app/controllers/source_controller.rb#L620) and [ProjectCopyNoPermission](https://github.com/openSUSE/open-build-service/blob/3743e2a80736cabfa2f714ba8e0c36ec48523ee4/src/api/app/controllers/source_controller.rb#L630) raise 403 errors. You can find them [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/source/errors.rb).

saraycp (changes requested):

src/api/public/apidocs-new/paths/source_project_name_cmd_copy.yaml
danidoni: So true!

eduardoj (changes requested):
The parameter `nodelay` should also be documented.

saraycp (approved):

eduardoj (approved):

PR #11918, opened 2021-12-01
[backend] bs_publish allow a syntax for rsync via ssh

bugfinder:
using ssh://[ssh_params::][rsync_user@]hostname::module[/path] allow a way of specifying stageserver for rsync via ssh


bugfinder:
ingnore, we have another approach

PR #11932, opened 2021-12-03
[api] Set 200 OK response for undelete action on project

saraycp:
This way, the endpoint will render a `200 OK` response instead of a `204 No Content` response when the 'undelete' is correctly performed.

PR #11933, opened 2021-12-03
[api] Fix undelete error response

saraycp:
Follow-up #11932

Before, the output was an error in XML format embedded in another XML structure. Like this:

```
 <status code="not_found">
    <summary>
         &lt;status code="404"&gt;
          &lt;summary&gt;project 'Sandbox' already exists&lt;/summary&gt;
          &lt;details&gt;404 project 'Sandbox' already exists&lt;/details&gt;
          &lt;/status&gt;
    </summary>
</status>
```

Now we get one single XML containing the error, using `400` code instead of `404`. Like this:

```
<status code="not_removed_project">
  <summary>The project 'sand3' already exists, it can't be undeleted.</summary>
</status>
```

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -129,6 +129,8 @@ class Project < ApplicationRecord
     end
 
     def restore(project_name, backend_opts = {})
+      return nil if find_by_name(project_name)
+
       Backend::Api::Sources::Project.undelete(project_name, backend_opts)
 
       # read meta data from backend to restore database object
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -532,8 +532,12 @@ class SourceController < ApplicationController
   def project_command_undelete
     raise CmdExecutionNoPermission, "no permission to execute command 'undelete'" unless User.session!.can_create_project?(params[:project])
 
-    Project.restore(params[:project])
-    render_ok
+    if Project.restore(params[:project])
+      render_ok
+    else
+      render_error status: 400, errorcode: 'not_removed_project',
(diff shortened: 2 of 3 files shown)


src/api/app/models/project.rb
```diff
@@ -129,6 +129,8 @@ def deleted?(project_name)
     end
 
     def restore(project_name, backend_opts = {})
+      return nil if find_by_name(project_name)
```
krauselukas, line 132: @saraycp maybe its better to use the `get_by_name` method as stated in the comment above the `find_by_name` method https://github.com/openSUSE/open-build-service/blob/2d22e0c11b71c4bfbcd97701f9535c9c77039123/src/api/app/models/project.rb#L228 you could then rescue from the exception it throws when there is no project

saraycp: I want to "undelete" a project that was previously deleted. If I use `get_by_name`, an exception is raised because it doesn't find the object (obviously), so it won't perform the rest of the code of "undelete".

Anyway, I've realized the solution is not going to be straightforward. I'm going to close this PR and simply introduce this change: https://github.com/openSUSE/open-build-service/pull/11932

saraycp:
This solution doesn't cover all the possible scenarios and the backend is responding with a 404 not_found error, which makes little sense since the resource exists.

I think a 400 error can be a better option. In fact, the package "undelete" action expects a 400 error code according to [this test](https://github.com/openSUSE/open-build-service/blob/f584139e08d6f78f0fc9cfd6edfac07ec0037340/src/api/test/functional/source_controller_test.rb#L2357).

So I'll close this PR and will postpone the refactoring. See this [Trello card](https://trello.com/c/trOQnIX6).

PR #11934, opened 2021-12-03
As apache2 doesn't hand over the proxy env settings

worldcitizen:
As apache2 doesn't hand over the proxy env settings
These have been "optionally" been added to the obs.conf vhost configuration
An extra debug line has been added to remote_url.rb


worldcitizen:
As the code has been improved, will create a new pull request.

PR #11935, opened 2021-12-03
Add source project name with release command

danidoni:
This PR documents the POST /source/<project>?cmd=release endpoint with the new OpenAPI specification.

### DIFF
--- a/src/api/public/apidocs-new/paths/source_project_name_cmd_release.yaml
+++ b/src/api/public/apidocs-new/paths/source_project_name_cmd_release.yaml
@@ -1,0 +1,47 @@
+post:
+  summary: Release the project.
+  description: |
+    Release source and binaries for a repository of the project, if you have the permissions to do so.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - in: query
+      name: nodelay
+      schema:
+        type: string
+      description: If this parameter is present, do not delay the relase. If this parameter is not present, the release will be delayed to be done later.
+      example: 1
+    - in: query
+      name: target_project
+      schema:
+        type: string
+      description: Project containing the repository targeted by the release.
+      example: 'openSUSE:Factory'
+    - in: query
+      name: target_repository
+      schema:
+        type: string
+      description: Repository targeted by the release.
+      example: standard
+    - in: query
+      name: repository
+      schema:
+        type: string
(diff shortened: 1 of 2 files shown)


danidoni:

dmarcoux:
Ready to review after adding a description to the endpoint's documentation.

src/api/public/apidocs-new/paths/source_project_name_cmd_release.yaml
```diff
@@ -0,0 +1,23 @@
+post:
+  summary: Release the project
+  description: |
+    Release source and binaries of the project, if you have the permissions to do so.
+  security:
+    - basic_authentication: []
+  parameters:
```
eduardoj, line 7: These parameters should also be documented: `:targetproject, :targetreposiory, :repository, :setrelease`. See: https://github.com/openSUSE/open-build-service/blob/12bad1ae9bd785f9164f42c4bb8266d12fabd54e/src/api/app/controllers/source_controller.rb#L559

eduardoj (changes requested):

src/api/public/apidocs-new/paths/source_project_name_cmd_release.yaml
dmarcoux: *project* is already documented, we use *project_name* in the documentation. It's not an issue that we use *project_name* instead of *project* since it's a path parameter.

As for *user*, this should not be documented as even if someone passes a user in the query string, it's overwritten with the logged-in user:
https://github.com/openSUSE/open-build-service/blob/12bad1ae9bd785f9164f42c4bb8266d12fabd54e/src/api/app/controllers/source_controller.rb#L539-L541

dmarcoux: This should be addressed now. Please have another look.

src/api/public/apidocs-new/paths/source_project_name_cmd_release.yaml
```diff
@@ -0,0 +1,47 @@
+post:
+  summary: Release the project
```
eduardoj, line 2:
```suggestion
  summary: Release the project.
```
Nitpick: missing dot at the end of the sentence.

Also missing in the `description` fields of the query parameters.

eduardoj (approved):

PR #11950, opened 2021-12-07
Document project showlinked command with OpenAPI

eduardoj:


src/api/public/apidocs-new/paths/source_project_name_cmd_showlinked.yaml
```diff
@@ -0,0 +1,43 @@
+post:
+  summary: List projects linking to the current project.
+  description: |
+    Return a list of projects linking to the project passed as parameter. This command doesn't perform any action.
```
dmarcoux, line 4: I think this would read better. What do you think?

```suggestion
    Return a list of projects linking to the provided project. This command doesn't perform any action.
```

src/api/public/apidocs-new/paths/source_project_name_cmd_showlinked.yaml
```diff
@@ -0,0 +1,43 @@
+post:
+  summary: List projects linking to the current project.
```
dmarcoux, line 2:
```suggestion
  summary: List projects linking to the provided project.
```

src/api/public/apidocs-new/paths/source_project_name_cmd_showlinked.yaml
eduardoj: Yes, way better. Thanks! Applied.

src/api/public/apidocs-new/paths/source_project_name_cmd_showlinked.yaml
eduardoj: Also applied.

dmarcoux (approved):

PR #11951, opened 2021-12-07
Store changes on the backend in repositories_configuration step

krauselukas:
In order to properly set the repositories on a project through
the workflows, we need to tell the backend about the changes as well,
otherwise the packages won't start to build against those repos.

### DIFF
--- a/src/api/app/models/workflow/step/configure_repositories.rb
+++ b/src/api/app/models/workflow/step/configure_repositories.rb
@@ -24,6 +24,10 @@ class Workflow::Step::ConfigureRepositories < Workflow::Step
         repository.architectures << @supported_architectures.select { |architecture| architecture.name == architecture_name }
       end
     end
+
+    # We have to store the changes on the backend
+    target_project.store(comment: "Added the following repositories to the project: #{step_instructions[:repositories].pluck(:name).compact.to_sentence}",
+                         login: @token.user.login)
   end
 
   private


src/api/app/models/workflow/step/configure_repositories.rb
```diff
@@ -12,18 +12,25 @@ def call(_options = {})
...
       repository.repository_architectures.destroy_all
+      repository_names << repository_instructions[:name]
 
       repository_instructions[:architectures].uniq.each do |architecture_name|
         repository.architectures << @supported_architectures.select { |architecture| architecture.name == architecture_name }
       end
     end
+
+    # We have to store the changes on the backend
+    target_project.store(comment: "Added the following repositories to the project: #{repository_names.to_sentence}",
```
dmarcoux, line 32: Storing repository names in a variable isn't necessary. This can be done with:

```suggestion
    target_project.store(comment: "Added the following repositories to the project: #{step_instructions[:repositories].pluck(:name).compact.to_sentence}",
```

`compact` is to remove any potential `nil` values.

krauselukas: @dmarcoux done

dmarcoux (approved):

dmarcoux (approved):

eduardoj (approved):

PR #11959, opened 2021-12-09
Rename @notifications_total to @notifications_count

dmarcoux:
The total refers to the count of all notifications of a user. It's not the count of the notifications to be returned to the user. `@notifications_per_page` isn't a valid name here since it refers to how many notifications are returned per page. This changes depending on the `show_all` query parameter.

### DIFF
--- a/src/api/app/controllers/person/notifications_controller.rb
+++ b/src/api/app/controllers/person/notifications_controller.rb
@@ -11,7 +11,7 @@ module Person
     # GET /my/notifications
     def index
       @notifications = paginated_notifications
-      @notifications_total = @notifications.count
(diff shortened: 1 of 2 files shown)


dmarcoux:
[Rename variable @notifications_total in API code](https://trello.com/c/d1oI0MaO/1809-rename-variable-notificationstotal-in-api-code)

eduardoj:
@dmarcoux , I rebased on master, to get rid of the linter errors.

saraycp (approved):
I agree, `notifications_count` is a good name.

eduardoj (approved):

PR #11960, opened 2021-12-09
Rename "Show all" in notifications API and web UI

dmarcoux:
It never showed all notifications. The "Show all" link and parameter show a maximum number of notifications (300 right now).

### DIFF
--- a/src/api/public/apidocs-new/paths/my_notifications.yaml
+++ b/src/api/public/apidocs-new/paths/my_notifications.yaml
@@ -6,14 +6,17 @@ get:
   parameters:
     - in: path
       name: project
-      required: false
       schema:
         type: string
     - in: path
       name: page
-      required: false
       schema:
         type: int
+    - in: query
+      name: show_maximum
+      schema:
+        type: string
+      example: 1
   responses:
     '200':
       description: |
--- a/src/api/app/controllers/person/notifications_controller.rb
+++ b/src/api/app/controllers/person/notifications_controller.rb
@@ -16,7 +16,7 @@ module Person
 
     private
 
-    def show_all(notifications)
+    def show_maximum(notifications)
       total = notifications.size
       notifications.page(params[:page]).per([total, MAX_PER_PAGE].min)
     end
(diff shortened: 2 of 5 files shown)


dmarcoux:
[Rename "Show all" to "Show/Display up to X/Show more"](https://trello.com/c/TMaQQMJW/1808-rename-show-all-to-show-display-up-to-x-show-more)

src/api/public/apidocs-new/paths/my_notifications.yaml
```diff
@@ -6,14 +6,17 @@ get:
   parameters:
     - in: path
       name: project
-      required: false
```
dmarcoux, line 9: By default, parameters aren't required, so it doesn't have to be documented.

saraycp (approved):

src/api/public/apidocs-new/paths/my_notifications.yaml
```diff
@@ -6,14 +6,17 @@ get:
...
         type: string
     - in: path
       name: page
-      required: false
       schema:
         type: int
+    - in: path
+      name: show_maximum
+      schema:
+        type: int
```
eduardoj, line 18:
```suggestion
        type: string
    - in: path
      name: show_maximum
      schema:
        type: string
```
According to [the documentation](https://swagger.io/docs/specification/data-models/data-types/#numbers), OpenAPI has two numeric types, `number` and `integer`. I would suggest to use `string`, as the current code doesn't strictly require an integer.

src/api/public/apidocs-new/paths/my_notifications.yaml
```diff
@@ -6,14 +6,17 @@ get:
...
       name: project
-      required: false
       schema:
         type: string
     - in: path
       name: page
-      required: false
       schema:
         type: int
+    - in: path
```
saraycp, line 15:
```suggestion
    - in: query
```

I'm going to open another PR to change the other parameters too and add a few more that are missing...

saraycp (changes requested):

eduardoj (approved):

saraycp (approved):

PR #11962, opened 2021-12-09
Support allowing build-depending on hidden projects

### LINKED ISSUE #7407: Support OBS ACLs to hide team-specific projects from other teams yet allowing dependencies
# Issue Description
Background
As a maintainer, i need to support creation of different projects in OBS for multiple product teams and restricting the access to only for developers of the product.

Usecase
For **Team1** we need to hide all Team1 related projects (`team1:target, team1:development. team1:sdk`) from other teams, with only Team1 team should to see and use this projects.
These instructions are used in the project configuration to hide the projects from others, but the issue is that hidden project cannot be used as a dependencies in any other `team1:*` projects:
`<sourceaccess> <disable/> </sourceaccess>  <access> <disable/> </access>`

Expected Result
============
* The selected OBS projects should be visible only to their project teams
* Only explicitly selected people/groups see any information about the selected OBS projects
* Restricted projects should able to be used as dependencies by other projects without necessarily sharing their sources but only sharing the binaries
* Sources aren't leaked out to other teams
* Binaries aren't leaked out to other teams
* Published APT repositories for resulting packages are protected as well

How to Reproduce
===============
1. create a new project `project:test:team1`
2. make it depend on a Debian DoD, for example.
3. configure it with `<access> <disable/> </access>`
4. ensure you're the only user with access there
5. create a new project `project:test:team2`
6. make it depend on `project:test:team1`
7. configure it with `<access> <disable/> </access>`
8. ensure you're the only user with access there
9. add a package in `team1` and one in `team2` that depends on it (for instance pick a random library and bump up its version in `team1` and pick a depending package adding a versioned dependency on the library to ensure it looks for the one in `team1`)

Current Result
============

Currently, when adding the <access> attribute results in a Project::Forbidden error to me, which ideally shouldn't happen given that I am the maintainer for both these projects.

An old [bug report](https://github.com/openSUSE/open-build-service/issues/791) on OBS project bug tracker has a similar report stating that changing `<access>` levels requires extra privileges but nothing was properly concluded in that bug report

Note: The **Project::Forbidden** error happens if you try to edit meta file with `<access>` levels after creating a projects. You can workaround it by selecting "Hide the entire project" in UI while creating new project from web UI. Ideally, it should allow changing project attributes through project meta configuration.

Note2: Even with the above workaround, if we are able to disable `<access>`, it hides the projects for all other users who are not in list" which solves the above `meta attributes` problem. But this doesn't allow to use this project as dependencies in other.

If i try to use `project:test:team2` in `project:test:team1` for build dependencies i see the error **The current backend implementation is not using binaries from read access protected projects**

andrewshadura:
As #7407 describes, it is sometimes necessary to allow a user to use binary artifacts of a project without allowing them to see the project itself. This merge request implements a new meta config element `<allowbuilddep>`. Projects mentioned as `<allowbuilddep name="accessing-project"/>` entries in the meta config are allowed to fetch build dependencies from the project unconditionally.

This patchset was originally developed by a colleague of mine originally for 2.7; I have rebased it onto the latest Git master, updated the migration timestamp and tested it on a local test instance.

Fixes #7407.


eduardoj:
Linters checks are failing. A rebase on master should fix it.

andrewshadura:
Test `validate_link_xml_attribute returns an error if target project access is disabled` is failing because an error was turned into a warning, but I’m not sure how to fix it; there doesn’t seem to be a way to catch that.

eduardoj:
> Test `validate_link_xml_attribute returns an error if target project access is disabled` is failing because an error was turned into a warning, but I’m not sure how to fix it; there doesn’t seem to be a way to catch that.

The actual error message is not correct. The error message returned by the test will be improved after #11973 is merged. This is what it should have been reported:
```
--- expected
+++ actual
-{:error=>"Project links work only when both projects have same read access protection level: the_project -> home:Iggy"}
+{}
```

andrewshadura:
True; however, I’m not sure how to proceed anyway. I guess, it’d be best to actually change the test to verify the new relationship properly.

src/api/app/controllers/source_project_meta_controller.rb
```diff
@@ -23,6 +23,11 @@ def show
     else
       # access check
       prj = Project.get_by_name(params[:project])
+      # since users may now see different versions of the
+      # project XML depending on their access permissions
+      # to linked projects, and the cache is global, we must
+      # invalidate the cache before reconstructing the xml.
+      Rails.cache.delete("xml_project_#{prj.id}") if prj.id
```
coolo, line 30: if you delete the cache here you can just as well not cache it at all - so better fix the cache key

andrewshadura: Fixed

andrewshadura:
I have reworked the pull request a bit and added a proper test.

adrianschroeter:
This is increasing the complexity of permission checking a lot. Also any project which is allowed to build against such a allowbuilddep defining project would be allowed to export everything.

So I wonder what we would win here at all, why can't you succeed with setting the user roles in both projects?

Also it would break interconnect if used, so we would need to prepare updates for old OBS instances.

andrewshadura:
> This is increasing the complexity of permission checking a lot. Also any project which is allowed to build against such a allowbuilddep defining project would be allowed to export everything.
>
> So I wonder what we would win here at all, why can't you succeed with setting the user roles in both projects?

Unless I changed the logic by accident in my latest refactoring, the idea is to be able to share binaries only. Our customers wanted this to enable teams who work on not yet public or non-public projects to still be able to provide artifacts to other teams without exposing everything. A number of people, myself included, explored other options for quite some time and we weren’t able to fit them into the existing permissions model.
 
> Also it would break interconnect if used, so we would need to prepare updates for old OBS instances.

I understand the primary users of this feature would not be using interconnect much, except maybe to fetch build artifacts from build.opensuse.org, so I wouldn’t worry too much. However, I do have patches for older OBS versions as well, but they’re not as polished as this pull request.

andrewshadura:
And yeah, of course, anyone who’s got binaries would be able to re-share them, but that’s fine, as even if OBS tried to prevent that, they would be able to do so anyway by a variety of other methods. However, the primary aim is to prevent access to the source. Limit access to binaries is necessary to some reasonable degree only.

andrewshadura:
Is there anything else I can do to help this move forward? Any clarifications, code changes, something else?

andrewshadura:
@adrianschroeter, @hennevogel, what is necessary for this to move forward?

adrianschroeter:
As written before, we fear the increased complexity and I am relative sure that the checks are not complete. However, it would require a quite complex review to all possibilities to access data (esp in backend specials where sources and binaries are accessed over the project boundaries. Since we need a new kind of filtering as taking over src.rpm for example must not be possible, I see no code for that. But this is just one example, I am sure there is more). 

This quite large effort is unfortunately no doable for us atm due to other jobs. (Maybe later this year, sorry).

andrewshadura:
What if we make the effort surface smaller by documenting that any build artifacts will be shared, so using this feature with src.rpm and similar will expose sources anyway? After all, there are some Debian packages including sources (e.g. Go or Rust packages are usually sources rolled in a .deb, and there are also packages like `linux-source`), so the user of this feature should be aware of what they’re doing anyway.

Issue #11902, opened 2021-11-26
Button "Mark all as 'Read'" never displayed


saraycp:
This is a follow-up of PR #11675 where we tried to introduce the "Mark all as 'Read/Unread'" button.

When we have more than `MAX_PER_PAGE` notifications (currently 300), the "Mark all as 'Read/Unread'"  button should appear but this is not happening. As this image shows:

![Screenshot 2021-11-26 at 10-53-35 Open Build Service](https://user-images.githubusercontent.com/2581944/143562140-54df36c2-f1cd-4981-87d8-103ffe3ab6fa.png)

The reason is that we are using the wrong counter in the condition that checks if the button should be visible or not.

We should also try to improve the spec to detect this case.

PR #11963, opened 2021-12-09
Fix "Mark all as 'Read'" button

saraycp:
Fixes https://github.com/openSUSE/open-build-service/issues/11902

The button "Mark all as 'Read'" should be displayed when the total number of existing notifications exceeds the maximum per page, which is 300 right now.

The button wasn't appearing because we were comparing it with the wrong counter. We now use the total number of notifications (after applying filters) instead of the current number of notifications per page.

I've slightly refactored the code, renamed the spec file and added some new tests.


krauselukas (commented):
@saraycp looks good to me, only one thing I saw, there is already a notification feature spec `src/api/spec/features/beta/webui/notifications_spec.rb`. Since you renamed the the other feature spec would it make sense to unify them?

saraycp:
> @saraycp looks good to me, only one thing I saw, there is already a notification feature spec `src/api/spec/features/beta/webui/notifications_spec.rb`. Since you renamed the the other feature spec would it make sense to unify them?

@krauselukas the test you mention is not related to the [User's Notifications page](https://build.opensuse.org/my/notifications) but to the [Notifications Settings page](https://build.opensuse.org/my/subscriptions) (a.k.a Subscriptions page). I've added a new commit to rename that spec inside and outside the _beta_ namespace.

So this is the situation now:

- Spec for the **User's Notification** page:
  - spec/features/beta/webui/users/notifications_spec.rb (former .../notifications_controller_spec.rb) 
- Specs for the beta and non-beta versions of the **Subscriptions/Notifications Settings** page:
  - spec/features/beta/webui/notifications_settings_spec.rb (former .../notifications_spec.rb)
  - spec/features/webui/notifications_settings_spec.rb (former .../notifications_spec.rb)

krauselukas:
> > @saraycp looks good to me, only one thing I saw, there is already a notification feature spec `src/api/spec/features/beta/webui/notifications_spec.rb`. Since you renamed the the other feature spec would it make sense to unify them?
> 
> @krauselukas the test you mention is not related to the [User's Notifications page](https://build.opensuse.org/my/notifications) but to the [Notifications Settings page](https://build.opensuse.org/my/subscriptions) (a.k.a Subscriptions page). I've added a new commit to rename that spec inside and outside the _beta_ namespace.
> 
> So this is the situation now:
> 
>     * Spec for the **User's Notification** page:
>       
>       * spec/features/beta/webui/users/notifications_spec.rb (former .../notifications_controller_spec.rb)
> 
>     * Specs for the beta and non-beta versions of the **Subscriptions/Notifications Settings** page:
>       
>       * spec/features/beta/webui/notifications_settings_spec.rb (former .../notifications_spec.rb)
>       * spec/features/webui/notifications_settings_spec.rb (former .../notifications_spec.rb)

@saraycp if its about the subscription page, I would say we should call it `subscriptions_spec.rb`. I know not really the goal of your PR, but something like `notifications_settings` we dont really have, so I would name it like the view is named

krauselukas (approved):

saraycp:
> @saraycp if its about the subscription page, I would say we should call it `subscriptions_spec.rb`. I know not really the goal of your PR, but something like `notifications_settings` we dont really have, so I would name it like the view is named

Good idea. I've renamed the files to _subscriptions_spec.rb_.

I introduced some more changes, I realized the _"Mark selected as 'Read'"_ stopped working due to my changes. Now it's fixed. See what's new [here](https://github.com/openSUSE/open-build-service/compare/d25417fc8b77f8487c3597ccb4dfd25e6d414345..f2b5ace03eb609f7e2e3b91a434e55c454b73fe8).

krauselukas (approved):

PR #11970, opened 2021-12-13
Create notification update api endpoint

krauselukas:
Since there is only one attribute for notifications were it makes
sense to change it by hand, the update endpoint doesn't work
with xml payload or a query parameter, but simply toggles the
`delivered` attribute of a notification to mark it as read/unread.

### DIFF
--- a/src/api/public/apidocs-new/paths/my_notifications_id.yaml
+++ b/src/api/public/apidocs-new/paths/my_notifications_id.yaml
@@ -1,0 +1,30 @@
+put:
+  summary: Mark a notification as read/unread.
+  description: |
+    If a notification is unread, it will be marked as read. If a notification is read, it will be marked as unread.
+  security:
+    - basic_authentication: []
+  parameters:
+    - in: path
+      name: id
+      required: true
+      schema:
+        type: integer
+  responses:
+    '200':
+      $ref: '../components/responses/succeeded.yaml'
+    '403':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      description: Not Found
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+          examples:
+            Not Found:
+              value:
+                code: not_found
+                summary: Couldn't find Notification with 'id'=213453
+  tags:
+    - Notifications
--- a/src/api/spec/controllers/person/notifications_controller_spec.rb
+++ b/src/api/spec/controllers/person/notifications_controller_spec.rb
@@ -94,4 +94,45 @@ RSpec.describe Person::NotificationsController do
       it { expect(response).to have_http_status(:unauthorized) }
     end
   end
+
+  describe '#update' do
+    let!(:notification) { create(:web_notification, :comment_for_package, subscriber: user) }
+
+    before do
+      Flipper[:notifications_redesign].enable
+    end
+
+    context 'called by an unauthorized user' do
+      let(:other_user) { create(:confirmed_user, :in_beta) }
+
+      before do
+        login other_user
+        put :update, params: { format: :xml, id: notification.id }
+      end
+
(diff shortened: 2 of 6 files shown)


krauselukas:
[Notifications Write API](https://trello.com/c/NaLNZT9e/1812-notifications-write-api)

src/api/public/apidocs-new/paths/my_notifications_id.yaml
```diff
@@ -0,0 +1,31 @@
...
+    Toggle the boolean value of the `delivered` attribute of a notification, which
+    marks the notification as read/unread.
+  security:
+    - basic_authentication: []
+  parameters:
+    - in: path
+      name: id
+      required: true
+      schema:
+        type: int
```
dmarcoux, line 13: This should be `integer`, not `int`. See https://swagger.io/docs/specification/data-models/data-types/

```suggestion
        type: integer
```

src/api/public/apidocs-new/paths/my_notifications_id.yaml
```diff
@@ -0,0 +1,31 @@
...
+      description: Not Found
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
+          examples:
+            Not Found:
+              value:
+                code: not_found
+                summary: Notification with id 213453 does not exist
```
dmarcoux, line 29: With my suggested changes in the controller, this message will be slightly different:

```suggestion
                summary: Couldn't find Notification with 'id'=213453
```

src/api/spec/controllers/person/notifications_controller_spec.rb
```diff
@@ -94,4 +94,46 @@
...
+
+      it 'toggles the delivered attribute' do
+        notification.reload
+        expect(notification.reload.delivered).to eq(true)
+      end
+
+      it { expect(response).to have_http_status(:success) }
+    end
+
+    context 'notification doesnt exist' do
```
dmarcoux, line 130:
```suggestion
    context "notification doesn't exist" do
```

src/api/spec/controllers/person/notifications_controller_spec.rb
```diff
@@ -94,4 +94,46 @@
...
+    end
+
+    context 'called by an authorized user' do
+      before do
+        login user
+        put :update, params: { format: :xml, id: notification.id }
+      end
+
+      it 'toggles the delivered attribute' do
+        notification.reload
```
dmarcoux, line 123: Why is doing `reload` twice needed?

dmarcoux (changes requested):

src/api/spec/controllers/person/notifications_controller_spec.rb
krauselukas: I had some weird problems without the reload, nevermind it's not needed!

krauselukas:
@dmarcoux applied your suggestions, please have another look :)

src/api/public/apidocs-new/paths/my_notifications_id.yaml
```diff
@@ -0,0 +1,31 @@
+put:
+  summary: Toggle the read/unread state of a notification.
+  description: |
+    Toggle the boolean value of the `delivered` attribute of a notification, which
```
dmarcoux, line 4: This is referring to the internal representation of notifications, but it's not useful to users. I would suggest something like this summary and description:

Summary: Mark a notification as read/unread
Description: If a notification is unread, it will be marked as read. If a notification is read, it will be marked as unread.

krauselukas: @dmarcoux Agree, applied

dmarcoux (approved):

eduardoj (approved):

PR #11980, opened 2021-12-14
Rely only on Flipper to check if feature toggle is enabled

dmarcoux:
Removing this code is needed to be able to roll out the notifications_redesign feature toggle. Otherwise, users in the rollout group would still need to be part of the beta program to use this feature... which is nullifying the purpose of rolling out a feature.


krauselukas (approved):

dmarcoux:
Rebased on *master* to fix a conflict in `src/api/app/controllers/person/notifications_controller.rb`.

krauselukas (approved):

eduardoj (approved):

PR #11981, opened 2021-12-14
scm support: turn title hack into meta element

adrianschroeter:
and supporting submit requests with git based packages.

### DIFF
--- a/src/backend/bs_srcserver
+++ b/src/backend/bs_srcserver
@@ -342,9 +342,9 @@ sub triggerservicerun {
   my ($cgi, $projid, $packid) = @_;
   my $rev = getrev($projid, $packid);
   my $pack = BSRevision::readpack_local($projid, $packid, 1);
-  if ($pack && $pack->{'title'} =~ /^obs-scm:(.*)/) {
+  if ($pack && $pack->{'scmsync'}) {
     die("Cannot use the scm bridge with old style services\n") if $BSConfig::old_style_services;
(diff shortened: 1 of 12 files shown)


src/backend/bs_srcserver
```diff
@@ -342,9 +342,9 @@ sub triggerservicerun {
   my ($cgi, $projid, $packid) = @_;
   my $rev = getrev($projid, $packid);
   my $pack = BSRevision::readpack_local($projid, $packid, 1);
-  if ($pack && $pack->{'title'} =~ /^obs-scm:(.*)/) {
+  if ($pack && $pack->{'scm'}) {
```
eduardoj, line 345:
```suggestion
  if ($pack && $pack->{'scmsync'}) {
```
Shouldn't it be `scmsync`? And also in the rest of the occurrences in the same file (I count five occurrences more).

src/api/db/migrate/20211214111900_add_scm_elements.rb
```diff
@@ -0,0 +1,6 @@
+class AddScmElements < ActiveRecord::Migration[6.1]
```
dmarcoux, line 1: This would be more descriptive:

```suggestion
class AddScmsyncToPackagesAndProjects< ActiveRecord::Migration[6.1]
```

The file should also be renamed to follow this change.

src/backend/bs_srcserver
adrianschroeter: right, missed to change the second commit as well .. will take care

eduardoj (approved):

PR #11982, opened 2021-12-14
Gather information when there is no binary package

### LINKED ISSUE #11629: [production][] undefined method `start_with?' for nil:NilClass

danidoni:
Sometimes we get an error because we're missing data coming from the
backend notifications during a binary release.

This PR attempts to get some more details surrounding this error.

This is a first approach to fix #11629

### DIFF
--- a/src/api/app/models/binary_release.rb
+++ b/src/api/app/models/binary_release.rb
@@ -87,7 +87,11 @@ class BinaryRelease < ApplicationRecord
           hash[:binary_updateinfo] = binary['updateinfoid']
           hash[:binary_updateinfo_version] = binary['updateinfoversion']
         end
+
+        # Send the binary to Airbrake, to get a deeper understanding when there is no binary['package']
+        Airbrake.notify('Binary release with no package payload', binary: binary) unless binary['package']
         source_package = Package.striping_multibuild_suffix(binary['package'])
+
         rp = Package.find_by_project_and_name(binary['project'], source_package)
         if source_package.include?(':') && !source_package.start_with?('_product:')
           flavor_name = binary['package'].gsub(/^#{source_package}:/, '')


adrianschroeter:
the event job is still there for a retry, no need to add code into git for this... (the data must not be lost or we would corrupt our package tracking history)

danidoni:
> the event job is still there for a retry, no need to add code into git for this... (the data must not be lost or we would corrupt our package tracking history)

Unfortunately we don't have any pending job of `job_class: UpdateReleasedBinariesJob`, so there is no way of getting the needed information :shrug:

danidoni:
> the event job is still there for a retry, no need to add code into git for this... (the data must not be lost or we would corrupt our package tracking history)

Do you know why are we getting no package information from the binary in [here](https://github.com/openSUSE/open-build-service/pull/11982/files#diff-66b13f83703fa7445d86fa05460876542faf1729c2fcc8a63fa1f92e6ec189f3R90)

danidoni:
BTW @adrianschroeter, this PR doesn't change the existing flow, it just uploads a bit more of information on the errbit so I can fix [this error](https://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/602a91fe84bf4605432b23c8). So we would be in slightly better shape, but not really fixing it yet.

adrianschroeter:
On Mittwoch, 15. Dezember 2021, 11:55:50 CET Dani Donisa wrote:
> BTW @adrianschroeter, this PR doesn't change the existing flow, it just uploads a bit more of information on the errbit so I can fix [this error](https://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/602a91fe84bf4605432b23c8). So we would be in slightly better shape, but not really fixing it yet.

my point is that we need to debug it on CLI as usual and rescue the old data, since it
must not get lost for the history.

-- 

Adrian Schroeter ***@***.***>

src/api/app/models/binary_release.rb
```diff
@@ -87,7 +87,15 @@ def self.update_binary_releases_via_json(repository, json, time = Time.now)
...
         end
-        source_package = Package.striping_multibuild_suffix(binary['package'])
+
+        # This is a hack to attempt to get a deeper understanding when there is no binary['package']
+        source_package = if binary['package']
+                           Package.striping_multibuild_suffix(binary['package'])
+                         else
+                           Airbrake.notify('Binary release with no package payload', binary: binary)
+                           nil
+                         end
```
eduardoj, line 97:
```suggestion
        # Send the binary to Airbrake, to get a deeper understanding when there is no binary['package']
        Airbrake.notify('Binary release with no package payload', binary: binary) unless binary['package']
        source_package = Package.striping_multibuild_suffix(binary['package'])
```
Maybe this way it is clearer that there is no change involved in the package tracking history. This would also get rid of the Brakeman warning.

danidoni: Okay, let's try it!

danidoni:
I'm closing this in favor of #11989

PR #11985, opened 2021-12-15
[api][backend] compability mode for scmsync

adrianschroeter:
Hiding the tag element for remote OBS instances.

Currently only doing so per /public route, it would be better to do
so by user-agent. Unfortunatly the interconnect is not using an own
identifier so far.


adrianschroeter:
thanks!

PR #11989, opened 2021-12-15
[api] fix a crash in package tracking

adrianschroeter:
when the image/container contains binaries provided via DoD


danidoni:
@adrianschroeter this PR supersedes #11982?

adrianschroeter:
it is supposed to fix the origin problem (untested), but it is not really any debug help. However, I think it is anyway done much faster on CLI (also because the only way to fix the incomplete data caused by it).

So not directly a replacement for your pull request :)

danidoni (approved):

eduardoj (approved):

PR #11996, opened 2021-12-17
[api] fix crash on project runservice command

adrianschroeter:
api can not update backend_data since no package object exists for "_project" package

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -1056,7 +1056,7 @@ class SourceController < ApplicationController
     path += build_query_from_hash(params, [:cmd, :comment, :user])
     pass_to_backend(path)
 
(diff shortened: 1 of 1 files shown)


adrianschroeter:
note: this is hot deployed on api.opensuse.org

vpereira:
> note: this is hot deployed on api.opensuse.org

Please check https://github.com/openSUSE/open-build-service/wiki/Deployment-of-build.opensuse.org#monkey-patching-on-production

If we don't notify our deployment tools, there is no guarantee that it will be there after the next deployment.

adrianschroeter:
no clue where this tool should come from, I have to admit. So far we had the agreement to put it in the motd.

However, merging it again to avoid that it breaks again over christmas (it is my last day today). (the rspec failure is independ of this change)

PR #12002, opened 2021-12-20
Remove mention to Beta in notifications API docs

saraycp:
The `notifications_redesign` is no longer a beta feature, now it's rolled out for every single user.

### DIFF
--- a/src/api/public/apidocs-new/paths/my_notifications.yaml
+++ b/src/api/public/apidocs-new/paths/my_notifications.yaml
@@ -1,6 +1,9 @@
 get:
   summary: List the notifications of the authenticated user.
-  description:  "**(Beta/Unstable)** List user's notifications."
+  description:  |
+    **(Unstable)** List user's notifications.
+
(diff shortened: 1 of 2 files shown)


src/api/public/apidocs-new/paths/my_notifications.yaml
```diff
@@ -1,6 +1,6 @@
 get:
   summary: List the notifications of the authenticated user.
-  description:  "**(Beta/Unstable)** List user's notifications."
+  description:  "List user's notifications."
```
eduardoj, line 3: I would add that this endpoint is available in the Unstable version. It has not yet been introduced in any official OBS release.

saraycp: Ok, I'll replace **(Beta/Unstable)** with **(Unstable)** and will add a note in the description; otherwise, I think it's probably not clear what that means.

eduardoj (approved):

PR #12018, opened 2021-12-23
Retire NotificationPresenter in favor of view components

### LINKED ISSUE #11558: Retire NotificationPresenter

>

saraycp:
`NotificationComponent` renders a notification with the help of a few small view components, each having a single responsibility.

Fixes #11558

### DIFF
--- a/src/api/app/components/notification_component.html.haml
+++ b/src/api/app/components/notification_component.html.haml
@@ -1,0 +1,30 @@
+.list-group-item.px-0.px-md-1.py-3
+  .row
+    .col-auto.pr-0
+      .custom-control.custom-checkbox
+        = check_box_tag('notification_ids[]', @notification.id, false,
+                        id: "notification_ids_#{@notification.id}", class: 'custom-control-input')
+        = label_tag("notification_ids_#{@notification.id}", '', class: 'custom-control-label')
+    .col-10
+      .row
+        .col
+          - if @notification.notifiable_type == 'BsRequest'
+            = image_tag('icons/request-icon.svg', height: 18, title: 'Request notification')
+            = render NotificationNotifiableLinkComponent.new(@notification)
+            %small.text-nowrap #{time_ago_in_words(@notification.created_at)} ago
+            %span.badge.ml-1{ class: "badge badge-#{helpers.request_badge_color(@notification.notifiable.state)}" }
+              = @notification.notifiable.state
+          - else
+            %i.fas.fa-comments{ title: 'Comment notification' }
+            = render NotificationNotifiableLinkComponent.new(@notification)
+            %small.text-nowrap #{time_ago_in_words(@notification.created_at)} ago
+        .col-auto.actions.ml-auto.align-self-end.align-self-md-start
+          = render NotificationMarkButtonComponent.new(@notification, @selected_filter, params[:page], params[:show_more])
+      .row.mt-1.pl-sm-4
+        .col-auto.pr-0
+          = render NotificationAvatarsComponent.new(@notification)
+        .col-auto.pl-xs-2
+          = render NotificationActionDescriptionComponent.new(@notification)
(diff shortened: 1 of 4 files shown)


src/api/app/components/notification_avatars_component.html.haml
```diff
@@ -0,0 +1,16 @@
+%ul.list-inline.d-flex.flex-row-reverse.avatars.m-0
+  - if avatar_objects.size > 6
+    %li.list-inline-item
+      - hidden = (avatar_objects.size - 6)
```
danidoni, line 4: What about moving this logic into the component? That way you can document those `6` and present a simpler interface in the view. :thought_balloon:

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
```diff
@@ -21,46 +21,8 @@
...
-                  .row.mt-1.pl-sm-4
-                    .col-auto.pr-0
-                      = render partial: 'notification_avatars', locals: { avatar_objects: notification.avatar_objects }
-                    .col-auto.pl-xs-2
-                      .smart-overflow= action_description(notification)
-                  .row.d-none.d-md-block.pl-4
-                    .col
-                      %p.mt-3.mb-0= notification.excerpt
+          - notifications.each do |notification|
+            = render NotificationComponent.new(notification, selected_filter)
```
danidoni, line 25: I think you can use the collections helpers in view controllers[^1] to render collections.

[^1]: https://viewcomponent.org/guide/collections.html

src/api/app/components/notification_component.html.haml
```diff
@@ -0,0 +1,40 @@
...
+            %small.text-nowrap #{time_ago_in_words(@notification.created_at)} ago
+            %span.badge.ml-1{ class: "badge badge-#{helpers.request_badge_color(@notification.notifiable.state)}" }
+              = @notification.notifiable.state
+          - else
+            %i.fas.fa-comments{ title: 'Comment notification' }
+            = render NotificationNotifiableLinkComponent.new(@notification)
+            %small.text-nowrap #{time_ago_in_words(@notification.created_at)} ago
+        .col-auto.actions.ml-auto.align-self-end.align-self-md-start
+          :ruby
+            title, icon = @notification.unread? ? ['Mark as "Read"', 'fa-check'] : ['Mark as "Unread"', 'fa-undo']
```
danidoni, line 23: I think this can also be pushed up into the component so we only ask for the title and the icon in the view, and the component does the logic :thought_balloon:

src/api/app/components/notification_avatars_component.html.haml
saraycp: I added a `NotificationAvatarsComponent` for this. I also refactored the code a bit, moving logic from the view to the component and using a constant instead of `6`, among others.

src/api/app/views/webui/users/notifications/_notifications_list.html.haml
saraycp: True! I've changed the call to the component to use the collection instead.

src/api/app/components/notification_component.html.haml
saraycp: I added a `NotificationMarkButtonComponent` for this.

saraycp:
@danidoni I addressed your feedback, can you have a look? 
I've created a `NotificationComponent` that calls to other components for many elements of the notification.

danidoni:
> @danidoni I addressed your feedback, can you have a look? I've created a `NotificationComponent` that calls to other components for many elements of the notification.

LGTM!

hennevogel:
[Turn NotificationPresenter into a view component](https://trello.com/c/U8c9ghnP/2384-turn-notificationpresenter-into-a-view-component)

danidoni (approved):

PR #12029, opened 2021-12-27
Extract code from BranchPackage#create_branch_packages

vpereira:
Extract code to private helper methods

### DIFF
--- a/src/api/app/models/branch_package.rb
+++ b/src/api/app/models/branch_package.rb
@@ -90,6 +90,22 @@ class BranchPackage
 
   private
 
+  def branch_target_package(package_hash)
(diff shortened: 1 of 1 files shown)


src/api/app/models/branch_package.rb
```diff
@@ -90,6 +90,22 @@ def branch
 
   private
 
+  def branch_target_package(packet_hash)
+    return unless packet_hash.key?(:target_package)
```
danidoni, line 94:
```suggestion
  def branch_target_package(package_hash)
    return unless package_hash.key?(:target_package)
```

danidoni (approved):

PR #12045, opened 2022-01-04
Clean BinaryRelease and add Spec for BinaryRelease.update_binary_releases

vpereira:
In this PR i removed some unnecessary comments (which should be done as well in other models) and expanded the spec for the `BinaryRelease` model

### DIFF
--- a/src/api/spec/models/binary_release_spec.rb
+++ b/src/api/spec/models/binary_release_spec.rb
@@ -6,14 +6,54 @@ RSpec.describe BinaryRelease do
       'disturl' => '/foo/bar',
       'supportstatus' => 'foo',
       'binaryid' => '31337',
-      'buildtime' => '1640772016'
+      'buildtime' => '1640772016',
+      'binaryarch' => 'i386',
+      'name' => 'foo'
     }
   end
 
-  describe '.update_binary_releases_via_json' do
-    let(:user) { create(:confirmed_user, :with_home, login: 'foo') }
-    let!(:repository) { create(:repository, project: user.home_project) }
+  let(:user) { create(:confirmed_user, :with_home, login: 'foo') }
+  let!(:repository) { create(:repository, project: user.home_project) }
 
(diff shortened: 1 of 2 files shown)


src/api/spec/models/binary_release_spec.rb
```diff
@@ -6,10 +6,26 @@
...
       'supportstatus' => 'foo',
       'binaryid' => '31337',
-      'buildtime' => '1640772016'
+      'buildtime' => '1640772016',
+      'binaryarch' => 'i386',
+      'name' => 'foo'
     }
   end
 
+  describe '.update_binary_releases' do
```
krauselukas, line 15: @vpereira just thinking if it might make sense to create a `BinaryRelease` record upfront, to see if it really gets updated. If I understand correctly thats what this method is meant for, to check the backend for existing binaries and then to update both on the frontend and backend.

krauselukas: So what I mean is that now we see that it successfully created a new record since there is no matching one, but we don't cover that it correctly checks existing records and replaces them with an updated version if exists

src/api/spec/models/binary_release_spec.rb
```diff
@@ -6,14 +6,54 @@
...
+      it { expect { subject }.to change(BinaryRelease, :count).by(1) }
+    end
 
+    context 'an existing binary release should be updated' do
+      let!(:binary_release) { create(:binary_release, repository: repository) }
+      let(:repeated_binary_hash) do
+        {
+          'disturl' => binary_release.binary_disturl,
+          'supportstatus' => binary_release.binary_supportstatus,
+          'binaryid' => '31338', # change change a value to avoid both objects to be identical
```
krauselukas, line 37:
```suggestion
          'binaryid' => '31338', # change a value to avoid both objects to be identical
```

krauselukas (approved):

PR #12046, opened 2022-01-04
Refactor comments presenter to a view component

### LINKED ISSUE #11559: Retire CommentsControllerPresenters::CommentPresenter

>

krauselukas:
We've decided to use view components in favor of presenters.

Fixes #11559

DRAFT:

- [x] Add specs


danidoni:
[Retire presenters](https://trello.com/c/PEfGTPrJ/1769-retire-presenters)

src/api/spec/components/comment_component_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'rails_helper'
+
+RSpec.describe CommentComponent, type: :component do
+  let(:builder) { Builder::XmlMarkup.new }
+
+  before do
+    render_inline(described_class.new(comment: comment, obj_is_user: true, builder: builder))
+  end
+
+  shared_examples 'rendering the body, who, when and the id' do
```
dmarcoux, line 10: I won't block a PR for this, but shared examples are a pain to update. Would it be possible to avoid using shared examples?

danidoni: It's simple enough now and it's not actually giving problems. Once it starts to be a real pain, we can inline it, but not now.

dmarcoux (approved):

PR #12052, opened 2022-01-07
[api] fix owner search by container regression

adrianschroeter:
Invalid 400 when only a package container name is given.

mostly hidden because osc searches also for binary name with same name,
but there are cases where no binary rpm has the container name


src/api/app/controllers/search_controller.rb
```diff
@@ -71,13 +71,26 @@ def owner_group_or_user
     end
   end
 
-  def owner_package_or_project
-    return if params[:project].blank?
+  def owner_packages_or_projects
+    return if params[:project].blank? and params[:package].blank?
```
danidoni, line 75:
```suggestion
    return if params[:project].blank? && params[:package].blank?
```

Please, do not use the `and` as a boolean operator. The `and` and `&&` are not exactly the same, the first one is used for flow control not for boolean logic... You can find more info [here](https://avdi.codes/using-and-and-or-in-ruby/)

danidoni (changes requested):

src/api/test/functional/search_controller_test.rb
```diff
@@ -509,12 +509,14 @@ def test_find_owner
     get '/search/owner?user=fred&filter=INVALID'
     assert_response 404
     assert_xml_tag tag: 'status', attributes: { code: 'not_found' }
-    get '/search/owner?package=TestPack'
-    assert_response 400
```
vpereira, line 513: How is it a regression if there was test case for that?

adrianschroeter:
On Freitag, 7. Januar 2022, 16:23:06 CET Victor Pereira wrote:
> @vpereira commented on this pull request.
> 
> 
> 
> > @@ -509,12 +509,14 @@ def test_find_owner
>      get '/search/owner?user=fred&filter=INVALID'
>      assert_response 404
>      assert_xml_tag tag: 'status', attributes: { code: 'not_found' }
> -    get '/search/owner?package=TestPack'
> -    assert_response 400
> 
> How is it a regression if there was test case for that?

it was introduced afterwards (or together?) with the regression.

You can see in the osc code that it used to be different otherwise
the additional check woud be pointless. And I can remeber of the
bugreport that these packages were not found (used to be an escalation).

-- 

Adrian Schroeter ***@***.***>

saraycp:
I rebased on master and forced push to re-run `codeclimate`.

vpereira (approved):

danidoni (approved):

Issue #11872, opened 2021-11-17
Unable to accept a maintenance incident if release package has linked upstream package


osnwt:
# Issue Description

**Environment:** local OBS instance 2.10.11 (latest appliance)
**Base project:** has `<link project="openSUSE.org:openSUSE:Leap:15.2:Update"/>` in its meta configuration
**Development project:** contains a patched source for an existing package from upstream (say, `shadow`)

Submitting a maintenance incident is ok, but attempt to accept it fails with the following error:

`Failed to change state: Branch call with missingok parameter but branched source (test:base/shadow) exists.!`

# Possible cause

`bs_request_action_maintenance_incident.rb`:

      # it is fine to have new packages
      branch_params[:missingok] = 1 unless Package.exists_by_project_and_name(branch_params[:project], package_name, follow_project_links: true)
      ret = BranchPackage.new(branch_params).branch

`branch_package.rb`:

      if params[:missingok]
        if Package.exists_by_project_and_name(params[:project], params[:package], follow_project_links: true, allow_remote_packages: true)
          raise NotMissingError, "Branch call with missingok parameter but branched source (#{params[:project]}/#{params[:package]}) exists."
        end

The code in `bs_request_action_maintenance_incident.rb` sets `missingok = 1`, but it does not look for remote packages (which may exist in the linked upstream project) because `allow_remote_packages: false` by default. Then the code in `branch_package.rb` raises an error since it looks **also** for remote packages due to `allow_remote_packages: true` (not seen by the first statement which sets `missingok`).

I am not an OBS expert. Maybe my project is improperly configured. But this looks to me like a bug, isn't it?

PR #12060, opened 2022-01-10
Fix issue 11872 (unable to accept a maintenance request)

osnwt:
This patch fixes the following issue: https://github.com/openSUSE/open-build-service/issues/11872

### DIFF
--- a/src/api/app/models/bs_request_action_maintenance_incident.rb
+++ b/src/api/app/models/bs_request_action_maintenance_incident.rb
@@ -154,7 +154,7 @@ class BsRequestActionMaintenanceIncident < BsRequestAction
         end
       end
       # it is fine to have new packages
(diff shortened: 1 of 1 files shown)


danidoni:
Thank you for your contribution @osnwt! It would be awesome if you could provide some small test to verify the fix, though!

osnwt:
> Thank you for your contribution @osnwt! It would be awesome if you could provide some small test to verify the fix, though!

Unfortunately, I have zero Ruby knowledge to provide a good test which should, perhaps, be a pipeline creating few projects. Moreover, I expected to get some feedback since may miss some use cases which I did not meet.

I tried to ask @IRC first, but instead was sent to a paid service for consultations ("it requires too much time to explain"). So I've solved the issue for myself, and it works well for my local OBS instance. Repeating the above said, I also have no time to learn the system more than required to fix it while other contributors even were not going to help me to investigate the issue.

I explained the problem, pointed to the code and provided some fix. Now someone who knows the internals should verify it and decide if that's a bug or a feature.

PR #12038, opened 2021-12-30
Update rubocop rails to 2.13.0

danidoni:
This PR supersedes #12027

PR #12062, opened 2022-01-10
Enable belongs to required by default

krauselukas:
Since Rails 5.0 `belongs_to` automatically checks for the presence
of the related record. We opted out this behavior in previous Rails
upgrades. It's time to enable the default behavior (also since
rubocop-rails 2.13 introduced a cop that checks for redundant
validation to check for the presence of a record).

Related to https://github.com/openSUSE/open-build-service/pull/12038

After enabling `belongs_to_required_by_default` we have to
explicitly tell rails where the associated record is optional.
I set it to optional where ever it is not disallowed on DB level with the
`not_null` value or where no explicit validation for presence
was set in the past. This way we keep the previous behavior and
don't break any code.

### DIFF
--- a/src/api/app/models/attrib.rb
+++ b/src/api/app/models/attrib.rb
@@ -8,8 +8,8 @@ class Attrib < ApplicationRecord
   delegate :namespace, to: :attrib_type
 
   #### Associations macros (Belongs to, Has one, Has many)
-  belongs_to :package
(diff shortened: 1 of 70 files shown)


danidoni (approved):

dmarcoux:
Nice work @krauselukas

Issue #10093, opened 2020-08-28
why new "memoryperjob" make no compliant worker for ppc64 not ppc64le ?


michelmno:
why new "memoryperjob" make no compliant worker for ppc64 not ppc64le ?

a good example is kokkos package https://build.opensuse.org/package/show/science/kokkos
 * where ppc64 reports indefinitely "no compliant workers (constraints mismatch hint: hardware)"
 * while ppc64le build triggered

Expected Result
============
Expect to have ppc64 treated same way as ppc64le concerning _constraints analysing.

How to Reproduce
===============
see kokkos build results in science project (see above link)

There are other packages with same problem but more complexe _constraints file like MozillaThunderbird:
https://build.opensuse.org/package/show/mozilla:Factory/MozillaThunderbird

Note: memoryperjob introduced by https://github.com/openSUSE/open-build-service/pull/10008

adrianschroeter:
On Freitag, 28. August 2020, 12:24:33 CEST Michel Normand wrote:
> why new "memoryperjob" make no compliant worker for ppc64 not ppc64le ?
> 
> a good example is kokkos package https://build.opensuse.org/package/show/science/kokkos
>  * where ppc64 reports indefinitely "no compliant workers (constraints mismatch hint: hardware)"
>  * while ppc64le build triggered
> 
> Expected Result
> ============
> Expect to have ppc64 treated same way as ppc64le concerning _constraints analysing.

that is a wrong expectation, not all workers can build big and little endian.

So it may be a very valid result that we do not have a worker for a certain
architecture being able to handle the constraints.

-- 

Adrian Schroeter <adrian@suse.de>
Build Infrastructure Project Manager

SUSE Software Solutions Germany GmbH,  Maxfeldstr. 5, 90409 Nuernberg, Germany
(HRB 247165, AG München), Geschäftsführer: Felix Imendörffer

michelmno:
> that is a wrong expectation, not all workers can build big and little endian.

I did not knew that.
What are the obs-power* workers that are not able to build ppc64 ?
Is is specified by a specific variable/value in output of  osc api /worker/ppc64le:obs-power*  command ?

krauselukas:
@adrianschroeter This is expected behavior and we can close it?

michelmno:
If we say this is expected behavior then many packages would not build for ppc64, while building for ppc64le.

dmarcoux:
@adrianschroeter It's expected behavior or not?

[...]

Issue #1953, opened 2016-07-20
for _constraints: memory requirements vary with available # or cores


DimStar77:
The _constraints system is quite powerful, but every so often I lack a possibility to request:
- I need <n> MB of RAM multiplied with however many cores you offer me

The rationale is rather simple: having a package that does parallel building will use more memory, the more CPU cores it receives.

Practical example, openSUSE:Factory/libtorrent-rasterbar

It requires 3.5G when building with make -j4
It required 1.5GB when building with make -j2
It required ~ 800M when buildnig with make -j1

the .spec file builds using make %{?_smp_mflags} - so the number of workers is injected by the worker / build script.

In this case if would be great to have a way to specify: I need 800M / core
The build script could then dynamically lower the number of parallel build jobs, based on the available RAM

AdamMajer:
Inability to specify memory per core is causing build failures with g++, especially in packages that have heavy utilization of templates. And simply adding higher memory requirements tends to drive these workloads onto machines with more cores which doesn't solve the problem.

rguenth:
Duplicate of #4433

adrianschroeter:
as said in the other report, I don't want to make the constraint system over complicate, since it is already a bottleneck and the next level of problem would be "I just have one compile job which needs 6GB for this generated gsoap file , the others just need 500MB and I need to configure that...".

The only sane way is to decide this per source in the spec file. OBS can not know anything about the need...

rguenth:
On Wed, 7 Mar 2018, Adrian Schröter wrote:

> as said in the other report, I don't want to make the constraint system over complicate, since it is already a bottleneck and the next level of problem would be "I just have one compile job which needs 6GB for this generated gsoap file , the others just need 500MB and I need to configure that...".
> 
> The only sane way is to decide this per source in the spec file. OBS can not know anything about the need...

Which is why we're trying to tell it ...

Why not remove _constraints completely?

marxin:
I would like to refresh discussion about this.
There are currently packages like MozillaFirefox, chromium, libreoffice that define it's own override of _threads. E.g. on firefox one can see:

```
[  343s] + _threads=16
[  343s] + _core_memory=1900
[  343s] + echo 'Available memory:'
[  343s] Available memory:
[  343s] + cat /proc/meminfo
[  343s] MemTotal:       11996308 kB
...
[  343s] + echo 'System jobs: 16'
[  343s] System jobs: 16
[  343s] + test 16 -gt 1
[...]

PR #12064, opened 2022-01-11
Use jobs in _constraints if available instead of worker jobs

### LINKED ISSUE #10167: memoryperjob constraint leaves package in unbuildable state
# Issue Description
build.opensuse.org:/science/polymake is using the `memoryperjob` constraint (which seems to be slated to supersede the previous `%limit_build` macro).

Observed Result
============
```
» cd polymake; osc bl openSUSE_Tumbleweed x86_64
package build was not possible:

no compliant workers (constraints mismatch hint: hardware:memoryperjob)

Please adapt your constraints.
```

Expected Result
============
Build the package, and build it in any case.

The default job strength for obsworkers on opensuse.org is -j4, and 3400x4=13600 is _still less_ than the memory asked by even larger packages such as MozillaFirefox.spec (16G).

Even if there was no worker was available with 13G of memory, the scheduler could reduce the jobcount accordingly until the constraint can be met—which makes this quite a powerful attribute in principle compared to the rather static `memory`/`physicalmemory` attr that firefox and others are using.

antlarr:
When multiplying the memoryperjob by the jobs to check if a worker
has enough memory, use the "jobs" value set in the _constraint
file if available instead of the worker jobs (core count) (which
is still used if "jobs" is not specified in order to keep the
current behaviour).

Multiplying memoryperjob by the worker's jobs (which I assume to be
the number of cpu cores) is basically wrong because usually the
number of concurrent jobs are set by dividing the total memory by
the memoryperjob (for example, with the %memory_limit macro) so,
if the check is done multiplying memoryperjob by number of cpu cores
then it can happen as in #10167 or #10093 where workers matching
the constraints can't be found because obs assumes the worker's
number of jobs will be used when the package could limit that
with %limit_build.

At least this allows the constraints to be reduced by setting
both memoryperjob and jobs together in the _constraints file
and would help fix #10167 and #10093.

### DIFF
--- a/src/backend/BSDispatcher/Constraints.pm
+++ b/src/backend/BSDispatcher/Constraints.pm
@@ -111,7 +111,7 @@ sub oracle {
     my $memory = ($worker->{'hardware'}->{'memory'} || 0);
     my $swap = ($worker->{'hardware'}->{'swap'} || 0);
     return 0 if $constraints->{'hardware'}->{'memory'} && getmbsize($constraints->{'hardware'}->{'memory'}) > ( $memory + $swap );
-    return 0 if $constraints->{'hardware'}->{'memoryperjob'} && getmbsize($constraints->{'hardware'}->{'memoryperjob'}) * ($worker->{'hardware'}->{'jobs'} || 1) > ( $memory + $swap );
+    return 0 if $constraints->{'hardware'}->{'memoryperjob'} && getmbsize($constraints->{'hardware'}->{'memoryperjob'}) * ($constraints->{'hardware'}->{'jobs'} || $worker->{'hardware'}->{'jobs'} || 1) > ( $memory + $swap );
     return 0 if $constraints->{'hardware'}->{'physicalmemory'} && getmbsize($constraints->{'hardware'}->{'physicalmemory'}) > $memory;
     if ($constraints->{'hardware'}->{'cpu'}) {
       my %workerflags = map {$_ => 1} @{$worker->{'hardware'}->{'cpu'}->{'flag'} || []};


adrianschroeter:
it would overwrite them indeed, but it would would create not large enough build environments now. Because the job would be started with more jobs and therefore not enough memory.

We would need to introduce another way here to make the number of jobs dynamic. But this must happen most likely independend of the constraints mechanism.

luc14n0:
What if we, at least for now, would cover two possible scenarios, 1) `memoryperjob` and `jobs` are set in the `_contraints` file and 2) only `memoryperjob` is set?

```
# If <memoryperjob> and <jobs> are set in _contraint file: <memoryperjob> * <jobs> < ( memory + swap )
return 0 if $constraints->{'hardware'}->{'memoryperjob'} && $constraints->{'hardware'}->{'jobs'} && getmbsize($constraints->{'hardware'}->{'memoryperjob'}) * $constraints->{'hardware'}->{'jobs'} < ( $memory + $swap );

# If only <memoryperjob> is set in _contraint file, use safest option for number of jobs (1):
# <memoryperjob> < ( memory + swap )
return 0 if $constraints->{'hardware'}->{'memoryperjob'} && getmbsize($constraints->{'hardware'}->{'memoryperjob'}) < ( $memory + $swap );
```
I haven't tested it though, but the logic seems correct. What do you think?

adrianschroeter:
nope it still has the same problem that the job will get started with more jobs and will need therefore more memory.

StefanBruens:
Currently, people are stepping away from `memoryperjob`, as it makes jobs unresolvable (mostly on PPC and aarch64), and badly influences scheduling on x86_64 (reduces number of eligible workers).

Using `%limit_build` needlessly only uses some of the available CPUs, which was the reason `memoryperjob` was introduced in the first place.

The need for a better scheduling is nothing new, it was already mentioned in the original memoryperjob issue here:
https://github.com/openSUSE/open-build-service/issues/1953#issuecomment-678975299

PR #12071, opened 2022-01-12
[api] fix access to build area for git projects

adrianschroeter:
Package objects do not exist in such projects
(same behavior as with remote projects)

should fix webui behaviour as well

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -162,6 +162,8 @@ class Package < ApplicationRecord
     prj = internal_get_project(project)
     return unless prj # remote prjs
 
+    return nil if prj.scmsync.present?
+
     if pkg.nil? && opts[:follow_project_links]
       pkg = prj.find_package(package, opts[:check_update_project])
     elsif pkg.nil?


src/api/app/models/user.rb
```diff
@@ -467,8 +467,10 @@ def can_modify_project?(project, ignore_lock = nil)
 
   # FIXME: This should be a policy
   # package is instance of Package
+  # rubocop:disable Metrics/PerceivedComplexity
   def can_modify_package?(package, ignore_lock = nil)
     return false if package.nil? # happens with remote packages easily
+    return false if package.project.scmsync.present? # packages are maintained in scm
```
hennevogel, line 473: Hm, you just want to avoid storing files to the backend right? This seems to be a bit of overkill, there are lots of database-only things that could/should be writeable.

How about #12277?

hennevogel (changes requested):

src/api/app/models/user.rb
hennevogel: Are you sure. I can happily add packages and files to a project with `scmsync`

hennevogel: Existing objects also do not get removed when adding `scmsync`.

adrianschroeter:
I removed the second (invalid) commit

src/api/app/models/user.rb
adrianschroeter: no be part of pull request anymore

PR #12076, opened 2022-01-13
Check pull/merge request actions before initializing dependencies

### LINKED ISSUE #11853: ScmWebhookEventValidator too strict
There are tons of actions that can happen on a pull/merge requests. Like applying a label, editing the description and whatnot. Every of those actions will fire a webhook. 

Currently we report an error back to the webhook because we don't "support" those events. We should just return and do nothing. Otherwise you end up with tons of errors on the SCM side and the *real* errors are hard to find.

rubhanazeem:
Fixes #11853

### DIFF
--- a/src/api/app/controllers/trigger_workflow_controller.rb
+++ b/src/api/app/controllers/trigger_workflow_controller.rb
@@ -1,14 +1,15 @@
 class TriggerWorkflowController < TriggerController
   # We don't need to validate that the body of the request is XML. We receive JSON
   skip_before_action :validate_xml_request, :set_project_name, :set_package_name, :set_project, :set_package, :set_object_to_authorize, :set_multibuild_flavor
-  before_action :create_workflow_run
   before_action :set_scm_event
+  before_action :abort_trigger_if_ignored_pull_request_action
+  before_action :create_workflow_run
   before_action :validate_scm_event
 
   def create
     authorize @token, :trigger?
     @token.user.run_as do
-      validation_errors = @token.call(scm: scm, event: event, payload: payload, workflow_run: @workflow_run)
+      validation_errors = @token.call(workflow_run: @workflow_run, scm_webhook: @scm_webhook)
 
       if validation_errors.none?
         @workflow_run.update(status: 'success', response_body: render_ok)
@@ -59,5 +60,11 @@ class TriggerWorkflowController < TriggerController
   def create_workflow_run
     request_headers = request.headers.to_h.keys.map { |k| "#{k}: #{request.headers[k]}" if k.match?(/^HTTP_/) }.compact.join("\n")
     @workflow_run = @token.workflow_runs.create(request_headers: request_headers, request_payload: request.body.read)
+  end
+
+  def abort_trigger_if_ignored_pull_request_action
+    @scm_webhook = TriggerControllerService::ScmExtractor.new(scm, event, payload).call
+
+    render_ok if @scm_webhook && @scm_webhook.ignored_pull_request_action?
   end
 end
--- a/src/api/app/models/scm_webhook.rb
+++ b/src/api/app/models/scm_webhook.rb
@@ -6,6 +6,11 @@ class ScmWebhook
 
   validates_with ScmWebhookEventValidator
 
+  IGNORED_PULL_REQUEST_ACTIONS = ['assigned', 'auto_merge_disabled', 'auto_merge_enabled', 'converted_to_draft',
+                                  'edited', 'labeled', 'locked', 'ready_for_review', 'review_request_removed',
+                                  'review_requested', 'unassigned', 'unlabeled', 'unlocked'].freeze
+  IGNORED_MERGE_REQUEST_ACTIONS = ['approved', 'unapproved'].freeze
+
   def initialize(attributes = {})
     super
     # To safely navigate the hash and compare keys
@@ -44,6 +49,10 @@ class ScmWebhook
     github_pull_request? || gitlab_merge_request?
   end
 
+  def ignored_pull_request_action?
+    ignored_github_pull_request_action? || ignored_gitlab_merge_request_action?
+  end
+
   private
 
   def github_push_event?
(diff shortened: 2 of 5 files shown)


src/api/app/models/scm_webhook.rb
```diff
@@ -44,6 +44,14 @@ def pull_request_event?
     github_pull_request? || gitlab_merge_request?
   end
 
+  def allowed_pull_request_or_merge_request_event?
+    if github_pull_request?
+      ScmWebhookEventValidator::ALLOWED_PULL_REQUEST_ACTIONS.include?(@payload[:action])
+    else
+      ScmWebhookEventValidator::ALLOWED_MERGE_REQUEST_ACTIONS.include?(@payload[:action])
```
danidoni, line 51: I think we also need to filter out other events like editing a PR description, that now are reaching our code but don't have any workflow defined.

src/api/app/validators/scm_webhook_event_validator.rb
```diff
@@ -4,6 +4,10 @@ class ScmWebhookEventValidator < ActiveModel::Validator
 
   ALLOWED_PULL_REQUEST_ACTIONS = ['closed', 'opened', 'reopened', 'synchronize'].freeze
   ALLOWED_MERGE_REQUEST_ACTIONS = ['close', 'merge', 'open', 'reopen', 'update'].freeze
+  NOT_ALLOWED_PULL_REQUEST_ACTIONS = ['assigned', 'auto_merge_disabled', 'auto_merge_enabled', 'converted_to_draft',
+                                      'edited', 'labeled', 'locked', 'ready_for_review', 'review_request_removed',
+                                      'review_requested', 'unassigned', 'unlabeled', 'unlocked'].freeze
+  NOT_ALLOWED_MERGE_REQUEST_ACTIONS = ['approved', 'unapproved'].freeze
```
saraycp, line 10: Do we really need to specify the list of not allowed ones? Is it not enough to ask if the incoming event IS NOT in the ALLOWED_*_ACTIONS?

src/api/app/models/scm_webhook.rb
```diff
@@ -44,6 +44,18 @@ def pull_request_event?
...
+  end
+
+  def not_allowed_pull_request_event?
+    github_pull_request? && ScmWebhookEventValidator::NOT_ALLOWED_PULL_REQUEST_ACTIONS.include?(@payload[:action])
+  end
+
+  def not_allowed_merge_request_event?
+    gitlab_merge_request? && ScmWebhookEventValidator::NOT_ALLOWED_MERGE_REQUEST_ACTIONS.include?(@payload[:action])
+  end
+
```
saraycp, line 58: I wonder if these methods can be converted to positive, using ALLOWED_*_ACTIONS instead of NOT_ALLOWED_*_ACTIONS :thinking:

saraycp (changes requested):

src/api/app/models/scm_webhook.rb
rubhanazeem: It's updated.

src/api/app/controllers/trigger_workflow_controller.rb
```diff
@@ -60,4 +61,11 @@ def create_workflow_run
     request_headers = request.headers.to_h.keys.map { |k| "#{k}: #{request.headers[k]}" if k.match?(/^HTTP_/) }.compact.join("\n")
     @workflow_run = @token.workflow_runs.create(request_headers: request_headers, request_payload: request.body.read)
   end
+
+  def check_supported_scm_event
+    @scm_webhook = TriggerControllerService::ScmExtractor.new(scm, event, payload).call
+
+    # We don't want to return error to SCM for the events that are not supported.
+    render_ok if @scm_webhook.present? && !@scm_webhook.allowed_pull_request_or_merge_request_event?
```
saraycp, line 69: After discussing offline if we can simply use `!@scm_webhook.valid?` here, we concluded that we can't because the validator not only validates events for pull request, but also validates push events (see code [here](https://github.com/openSUSE/open-build-service/blob/e6ce39895e7f8e9a51d0134868206826fea857ef/src/api/app/validators/scm_webhook_event_validator.rb#L51)). And we don't plan to ignore the errors related to `push` for now.

dmarcoux:
As discussed with @rubhanazeem, I pushed a commit with my proposal for some minor improvements. Feel free to squash or drop the commit :slightly_smiling_face:

src/api/app/models/token/workflow.rb
```diff
@@ -9,10 +9,10 @@ def self.token_name
 
   def call(options)
     set_triggered_at
+    @scm_webhook = options[:scm_webhook]
 
-    raise Token::Errors::MissingPayload, 'A payload is required' if options[:payload].nil?
+    raise Token::Errors::MissingPayload, 'A payload is required' if @scm_webhook.nil?
```
dmarcoux, line 14: I don't think this is right, since `@scm_webhook` contains the payload. There's also a similar validation happening in the controller. We can clean this up later. For now I suggest going with:

```suggestion
    raise Token::Errors::MissingPayload, 'A payload is required' if @scm_webhook.payload.nil?
```

rubhanazeem: Right, it's updated.

src/api/app/models/token/workflow.rb
```diff
@@ -9,10 +9,10 @@ def self.token_name
 
   def call(options)
     set_triggered_at
+    @scm_webhook = options[:scm_webhook]
 
-    raise Token::Errors::MissingPayload, 'A payload is required' if options[:payload].nil?
+    raise Token::Errors::MissingPayload, 'A payload is required' if @scm_webhook&.payload.nil?
```
dmarcoux, line 14: `payload` is always going to be there since it's defined as a `attr_accessor`: https://github.com/openSUSE/open-build-service/blob/550f59cdb733d0634458e5e229e5b3f0afa20a5c/src/api/app/models/scm_webhook.rb#L5

So with this in mind, we don't need the safe navigation:
```suggestion
    raise Token::Errors::MissingPayload, 'A payload is required' if @scm_webhook.payload.nil?
```

rubhanazeem: Some specs start to fail if we remove safe navigation.

dmarcoux: I pushed a fix for that.

rubhanazeem: Perfect. I've squashed the new commit.

dmarcoux (approved):

saraycp (approved):

PR #12077, opened 2022-01-13
Show meaningful data for each workflow run

danidoni:
This PR changes the Workflow Run index page to show a bit more of detail on each displayed item to quickly identify each run.

## Before
image

## After
image

Missing:
- [x] Refactor the view to use View Components
- [x] Write specs
- [x] Improve the code a bit more

PR #12098, opened 2022-01-18
Hide unsupported actions in WorkflowRun's index page

danidoni:
We do not want to show anything if the action is not supported.

This fixes #12077 showing "unsupported" tags in the view.

## Before
image

## After
image

### DIFF
--- a/src/api/app/components/workflow_run_row_component.rb
+++ b/src/api/app/components/workflow_run_row_component.rb
@@ -11,8 +11,6 @@ class WorkflowRunRowComponent < ApplicationComponent
   def hook_action
     return payload['action'] if
       hook_event == 'pull_request' && ScmWebhookEventValidator::ALLOWED_PULL_REQUEST_ACTIONS.include?(payload['action'])
-
-    'Unsupported'
   end
 
   def hook_event
--- a/src/api/spec/components/workflow_run_row_component_spec.rb
+++ b/src/api/spec/components/workflow_run_row_component_spec.rb
@@ -121,7 +121,7 @@ RSpec.describe WorkflowRunRowComponent, type: :component do
       end
 
       it 'does not show the action anywhere' do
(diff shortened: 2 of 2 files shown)


danidoni:
[Improvements for WorkflowRuns Index UI](https://trello.com/c/3axxhgrd/1820-improvements-for-workflowruns-index-ui)

src/api/app/components/workflow_run_row_component.rb
```diff
@@ -11,8 +11,6 @@ def initialize(workflow_run:)
   def hook_action
```
dmarcoux, line 11: `payload['action']` will be `nil` if the webhook event is not for a pull request. In this case, the Bootstrap badge shouldn't be displayed.

danidoni: That's right, but it would show up for unsupported pull request actions, and we don't want that.

danidoni: We can always remove the condition when we properly skip running workflow runs for unsupported actions.

dmarcoux: > We can always remove the condition when we properly skip running workflow runs for unsupported actions.

Alright, I am fine with this.

src/api/spec/components/workflow_run_row_component_spec.rb
```diff
@@ -135,6 +135,7 @@
     let(:request_payload) do
       <<~END_OF_REQUEST
         {
+          "action": "unsupported",
```
dmarcoux, line 138: Push event payloads don't have `action`, so the newly added spec on line 153 should pass without this line.

```suggestion
```

danidoni: That's true, removing

dmarcoux (approved):

PR #12104, opened 2022-01-18
Simplify token creation code

rubhanazeem:
Right now, we don't have `token#show` and the only way to view a token is using `tokens#edit`. And we reuse the view `tokens/new` to show the token after creating a token, it is full of hidden fields and it's confusing.

This PR introduces `tokens#show` to simplify the token creation process and to have a new page to view tokens.
The token creation code is much simpler now and it's following Rails convention.

### DIFF
--- a/src/api/app/components/copy_to_clipboard_input_component.html.haml
+++ b/src/api/app/components/copy_to_clipboard_input_component.html.haml
@@ -1,5 +1,4 @@
--# TODO: Relying on a form builder is a bit weird. We could perhaps directly use `text_field_tag` instead and pass an input value to the component.
-= @form_builder.text_field(:string_readonly, id: 'copy-to-clipboard-readonly', class: 'form-control', readonly: true)
+= text_field_tag(nil, @token_string, readonly: true, id: 'copy-to-clipboard-readonly', class: 'form-control')
 .input-group-append#copy-to-clipboard
   %span.input-group-text
     %i.fas.fa-clipboard
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -1,5 +1,5 @@
 class Webui::Users::TokensController < Webui::WebuiController
-  before_action :set_token, only: [:edit, :update, :destroy]
+  before_action :set_token, only: [:edit, :update, :destroy, :show]
   before_action :set_parameters, :set_package, only: [:create]
 
   after_action :verify_authorized, except: :index
@@ -49,13 +49,14 @@ class Webui::Users::TokensController < Webui::WebuiController
     authorize @token
 
     respond_to do |format|
-      format.js do
+      format.html do
         if @token.save
-          flash.now[:success] = "Token successfully created! Make sure you save it - you won't be able to access it again."
-          render partial: 'create', locals: { string: @token.string }
+          flash[:success] = "Token successfully created! Make sure you save it - you won't be able to access it again."
+          session[:show_token] = 'true'
+          redirect_to token_path(@token)
         else
-          flash.now[:error] = "Failed to create token: #{@token.errors.full_messages.to_sentence}."
-          render partial: 'create'
+          flash[:error] = "Failed to create token: #{@token.errors.full_messages.to_sentence}."
+          render :new
         end
       end
     end
@@ -66,6 +67,10 @@ class Webui::Users::TokensController < Webui::WebuiController
     @token.destroy
     flash[:success] = 'Token was successfully deleted.'
     redirect_to tokens_url
+  end
+
+  def show
+    authorize @token
   end
 
   private
@@ -97,20 +102,20 @@ class Webui::Users::TokensController < Webui::WebuiController
 
     # Check if only project_name or only package_name are present
     if @extra_params[:project_name].present? ^ @extra_params[:package_name].present?
-      flash.now[:error] = 'When providing an optional package, both Project name and Package name must be provided.'
-      render partial: 'create' and return
+      flash[:error] = 'When providing an optional package, both Project name and Package name must be provided.'
+      render :new and return
     end
 
     # If both project_name and package_name are present, check if this is an existing package
     begin
       @package = Package.get_by_project_and_name(@extra_params[:project_name], @extra_params[:package_name])
     rescue Project::UnknownObjectError
-      flash.now[:error] = "When providing an optional package, the package must exist. Project '#{elide(@extra_params[:project_name])}' does not exist."
-      render partial: 'create'
+      flash[:error] = "When providing an optional package, the package must exist. Project '#{elide(@extra_params[:project_name])}' does not exist."
+      render :new
     rescue Package::UnknownObjectError
-      flash.now[:error] = 'When providing an optional package, the package must exist. ' \
-                          "Package '#{elide(@extra_params[:project_name])}/#{elide(@extra_params[:package_name])}' does not exist."
-      render partial: 'create'
+      flash[:error] = 'When providing an optional package, the package must exist. ' \
+                      "Package '#{elide(@extra_params[:project_name])}/#{elide(@extra_params[:package_name])}' does not exist."
+      render :new
     end
   end
 end
--- a/src/api/app/views/webui/users/tokens/_breadcrumb_items.html.haml
+++ b/src/api/app/views/webui/users/tokens/_breadcrumb_items.html.haml
@@ -5,7 +5,7 @@
 - when 'index'
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
     Tokens
-- when 'new'
+- when 'new', 'create'
   %li.breadcrumb-item
     = link_to 'Tokens', tokens_path
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
@@ -15,3 +15,8 @@
     = link_to 'Tokens', tokens_path
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
     Edit Token
+- when 'show'
+  %li.breadcrumb-item
+    = link_to 'Tokens', tokens_path
+  %li.breadcrumb-item.active{ 'aria-current' => 'page' }
+    Show Token
--- a/src/api/app/views/webui/users/tokens/new.html.haml
+++ b/src/api/app/views/webui/users/tokens/new.html.haml
@@ -7,24 +7,7 @@
 
     .row
       .col-12.col-md-10.col-lg-8
-        = form_with(model: @token, method: :post, local: false) do |f|
-          .visible-when-created.d-none
-            %p.font-weight-bold Your new OBS Personal Access Token
-
-            .form-row#token-id
-              .col-sm
-                .form-group.mb-0
-                  = f.label(:id, 'Id:')
-                  %span= @token.id
-
-            .form-row#created-token
-              .col-12.col-md-10.col-lg-9
-                %fieldset.form-group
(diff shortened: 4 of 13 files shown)


src/api/app/components/copy_to_clipboard_input_component.html.haml
```diff
@@ -1,5 +1,5 @@
 -# TODO: Relying on a form builder is a bit weird. We could perhaps directly use `text_field_tag` instead and pass an input value to the component.
-= @form_builder.text_field(:string_readonly, id: 'copy-to-clipboard-readonly', class: 'form-control', readonly: true)
+%input.form-control{value: @token_string, disabled: true}
```
dmarcoux, line 2: I suggest using the [text_field_tag](https://api.rubyonrails.org/classes/ActionView/Helpers/FormTagHelper.html#method-i-text_field_tag) helper instead. The id `copy-to-clipboard-readonly` should still be used for the JavaScript to keep working.

src/api/app/views/webui/users/tokens/new.html.haml
```diff
@@ -7,24 +7,7 @@
...
-
-            .form-row#created-token
-              .col-12.col-md-10.col-lg-9
-                %fieldset.form-group
-                  = f.label(:string_readonly, 'Secret:', class: 'col-form-label')
-                  .input-group
-                    = render CopyToClipboardInputComponent.new(form_builder: f)
-                %hr
-
+        = form_with(model: @token ||= Token.new, method: :post, local: true) do |f|
```
dmarcoux, line 10: [local: true](https://api.rubyonrails.org/v6.1.4/classes/ActionView/Helpers/FormHelper.html#method-i-form_with) is the default value, so it can be removed.

rubhanazeem: The default value is `local: false`. It sets `remote: true`.

src/api/app/views/webui/users/tokens/_breadcrumb_items.html.haml
```diff
@@ -5,7 +5,7 @@
 - when 'index'
   %li.breadcrumb-item.active{ 'aria-current' => 'page' }
     Tokens
-- when 'new'
+- when 'new', 'create'
```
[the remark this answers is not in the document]
rubhanazeem: I've deleted the partial `_create.js.erb`. But we need `create` here with `new` for error cases. If a record is not saved it stays in create action. That's true it doesn't have a template but it renders new.

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -49,13 +49,14 @@ def create
     authorize @token
 
     respond_to do |format|
-      format.js do
+      format.html do
         if @token.save
-          flash.now[:success] = "Token successfully created! Make sure you save it - you won't be able to access it again."
-          render partial: 'create', locals: { string: @token.string }
+          flash[:success] = "Token successfully created! Make sure you save it - you won't be able to access it again."
+          flash[:show_token] = 'true'
```
dmarcoux, line 55: This flash is displayed to users when they create a token (see the `true` in the screenshot below):

bug

Using an instance variable instead of storing something in a flash message would be a better option here. This way, you can still check if this instance variable is set, then display the token string

src/api/app/views/webui/users/tokens/new.html.haml
dmarcoux: Oh, I double-checked and you are right. Then it's all good :+1:

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,64 @@
+- @pagetitle = 'Show Token'
+
+.card
+  .card-body
+    %h3.mb-3= @pagetitle
+
+    .row
+      .col-12.col-md-10.col-lg-8
+        %p.font-weight-bold Your new OBS Personal Access Token
```
dmarcoux, line 9: This text should be rendered only right after the creation of the token, so when we show the token string with `session[:show_token]`. Afterwards, the token isn't new anymore.

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,64 @@
...
+            .col-sm
+              .form-group.ui-front
+                %label.font-weight-bold
+                  Package name:
+                %span
+                  = @token.package.name
+
+        .actions
+          = render partial: 'actions'
+          = link_to('Edit', edit_token_path(@token), title: 'Edit', class: 'btn btn-outline-primary px-4 mr-3 mt-3 mt-sm-0')
```
dmarcoux, line 64: Edit links/buttons should have the `fas.fa-edit` icon as described in the [icons](https://obs-patterns.netlify.app/atoms/icons.html) page of our pattern library.

We tend to have _Edit_ links next to the name/title of object. In this case, it would be right next to the token name (`<h3>` at the top of this view).

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,64 @@
+- @pagetitle = 'Show Token'
+
+.card
+  .card-body
+    %h3.mb-3= @pagetitle
```
dmarcoux, line 5: If the token's name is defined, I suggest displaying here. If not, I suggest displaying _No name_ as we do on the tokens index page. We should change the `@pagetitle` to reflect this in the view and in the browser's window title.

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,64 @@
...
+          .form-row
+            .col-sm
+              .form-group.ui-front
+                %label.font-weight-bold
+                  Package name:
+                %span
+                  = @token.package.name
+
+        .actions
+          = render partial: 'actions'
```
dmarcoux, line 63: I don't see a need for this partial, since it only contains a single link and it's never reused. Let's put the link directly here instead. It could also be a simple link, not a button. It would look nicer by not being so "blocky".

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,64 @@
...
+                  = render CopyToClipboardInputComponent.new(token_string: @token.string)
+              %hr
+
+        .form-row
+          .col
+            .form-group
+              %label.col-form-label.mr-2.font-weight-bold
+                Type:
+              %span
+                = @token.type
```
dmarcoux, line 35: We should be consistent with the other views when displaying the token type. On the `new` view, we do it this way:

```suggestion
                = @token.type.constantize.token_name
```

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,64 @@
...
+
+        .form-row
+          .col-sm
+            .form-group.ui-front
+              %label.font-weight-bold
+                Name:
+              %span
+                = @token.name
```
dmarcoux, line 43: Just like on the tokens index page, we should display _No name_ if the token doesn't have a name.

dmarcoux (changes requested):

src/api/app/views/webui/users/tokens/show.html.haml
rubhanazeem: This partial has been used in other views as well. I've updated the button style but haven't removed the partial itself.

src/api/app/views/webui/users/tokens/show.html.haml
rubhanazeem: I've removed it since the name has been assigned to `@pagetitle`.

src/api/app/views/webui/users/tokens/show.html.haml
dmarcoux: You're right, it's used in this view and `src/api/app/views/webui/users/tokens/_update.js.erb`.

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,60 @@
+- @pagetitle = @token.name.presence || 'No name'
```
dmarcoux, line 1: If multiple tabs are open, this small change helps in telling what the tab is about:

```suggestion
- @pagetitle = "Token - #{@token.name.presence || 'No name'}"
```

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,60 @@
...
+
+.card
+  .card-body
+    %h3.mb-3
+      = @pagetitle
+      - if policy(@token).edit?
```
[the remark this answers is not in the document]
rubhanazeem: I can put it back but without a button, at the bottom of the page, it's not going to look good. Should we put it back in button?

hennevogel:
In general I think the actions (at least trigger and workflow runs) should go to the show view.

dmarcoux:
> In general I think the actions (at least trigger and workflow runs) should go to the show view.

With the `trigger` view, I agree that it can be merged in the `show` view. It's also noted in the card. As for the workflow runs, we can have a link to that page on both the tokens index and show views.

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,68 @@
+- @pagetitle = "Token - #{@token.name.presence || 'No name'}"
```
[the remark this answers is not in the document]
rubhanazeem: They are back.

danidoni (approved):

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,68 @@
...
+                = @token.token_name
+
+        - if @token.package
+          .form-row
+            .col-sm
+              .form-group.ui-front
+                %label.font-weight-bold
+                  Project name:
+                %span
+                  = @token.package.project.name
```
dmarcoux, line 46: This could be a link to the project.

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,68 @@
...
+
+          .form-row
+            .col-sm
+              .form-group.ui-front
+                %label.font-weight-bold
+                  Package name:
+                %span
+                  = @token.package.name
```
dmarcoux, line 54: This could be a link to the package.

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,68 @@
...
+                Type:
+              %span
+                = @token.token_name
+
+        - if @token.package
+          .form-row
+            .col-sm
+              .form-group.ui-front
+                %label.font-weight-bold
+                  Project name:
```
dmarcoux, line 44: This isn't a name anymore, but a link, so we can remove `name` from the label:

```suggestion
                  Project:
```

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -0,0 +1,68 @@
...
+                %span
+                  = link_to(@token.package.project.name, project_show_path(@token.package.project))
+
+          .form-row
+            .col-sm
+              .form-group.ui-front
+                %label.font-weight-bold
+                  Package name:
```
dmarcoux, line 52: Same as my previous comment, this isn't a name anymore, but a link:

```suggestion
                  Package:
```

dmarcoux (approved):

Issue #12031, opened 2021-12-28
Issue abbreviation system mismatching


nopeinomicon:
# Issue Description
(Note: this issue is exhibited on the openSUSE instance of OBS, but due to its nature I assume it is universal for the program in general)
In any areas where issue abbreviations from the changes file are gathered and listed, there is a problem in the case of issue tags with multiple '#' characters where the issue will be read twice assuming the characters required for both issue templates are present. It's a tad difficult to explain in the abstract so I'll provide an illustration of the example I linked below

In the example the string `gh#cli/cli#3344` is interpreted by the server as two different issues. It first catches the entirety of the string as a GitHub issue `gh#cli/cli#3344` (as it should) but it *also* reads an OpenOffice issue (incorrectly), just grabbing the last part of the string that matches the OpenOffice issue pattern `i#3344`.

I have not tested this issue with any other edge cases like this that would trigger two issue templates like this, but they are completely possible. For example referencing an issue from https://github.com/NVIDIA/TensorRT would end up causing the server to detect both that issue and a CPAN Public Bug Tracker issue.

Expected Result
============
The issue as typed in the changes file is displayed, with no extraneous duplicates or mismatches

How to Reproduce
===============
1. Reference in the changes file (or other place where issues are tracked) a given GitHub issue (or other issue type with two '#'s) where the ending characters match another issue tracker template 
2. Go to some place where OBS will process and display the issues referenced in the file (in my case I observed this behavior on the package submission request page)

Further Information
===============
* A link to an OBS instance showing the issue: https://build.opensuse.org/request/show/942879

PR #12109, opened 2022-01-19
Avoid mismatching with different issue trackers

saraycp:
Having a `x.changes` file that contains a text like `- Fixes gh#cli/cli#1 issue.`, the only issue tracker that should match is the `gh` one, but not the `i` issue tracker.

The code is wrongly assuming that `i#1` coming from _gh#cli/cl**i#1**_ is an issue referenced in the `x.changes` file, but it shouldn't.
 
By setting word boundaries on the regular expression, we want to prevent the mismatches.

- `gh#cli/cli#1`
- `bcn#123445/boo#98487`

Closes https://github.com/openSUSE/open-build-service/issues/12031


krauselukas:
@adrianschroeter @bugfinder I dont know how this is setup for the backend in production, I think you need to adjust the config/xml file that contains the issue_tracker regular expression for the backend in order to make this work https://github.com/openSUSE/open-build-service/blob/0723856ff7913d81bce7d6075451e0b8c8a6c5e5/src/backend/bs_srcserver#L3204

dmarcoux:
It seems like the regex should always start and finish with `\b`. We should enforce this with a validation on the `IssueTracker` model.

hennevogel:
And why do we do this in the regex column instead where we use that regex?

krauselukas:
@hennevogel I can't only adapt the code to use the word boundary since the frontend is sending the issue tracker informations to the backend in a delay job https://github.com/openSUSE/open-build-service/blob/28b5bdf5f55269c2cc2db9631c35bf0169010b78/src/api/app/jobs/issue_tracker_write_to_backend_job.rb#L3 since the backend is creating the issues we have to send the correct regular expressions. Thats why we have to update the database... also just figured this out
@dmarcoux I added a validation to the issue tracker model to check for the word boundary in the future, please have a look :)

krauselukas:
> @adrianschroeter @bugfinder I dont know how this is setup for the backend in production, I think you need to adjust the config/xml file on our production instances that contains the issue_tracker regular expression for the backend in order to make this work
> 
> https://github.com/openSUSE/open-build-service/blob/0723856ff7913d81bce7d6075451e0b8c8a6c5e5/src/backend/bs_srcserver#L3204

Overlooked that there is a delayed job to update the backend side, so no action required

krauselukas:
I see have to adapt a couple of specs and minitest

adrianschroeter:
it does not make much sense to enforce that the config must contain \b in start and end. If we want this, we should adapt the parser instead.

adrianschroeter:
Not sure if we want to change this behaviour though, because it would also change the results of past changelog entries. Why should we support the writing of combining issues using slashes now?

krauselukas:
> Not sure if we want to change this behaviour though, because it would also change the results of past changelog entries. Why should we support the writing of combining issues using slashes now?

Do you mean by combining issues something like this `foo#123/foo#456` ? This would currently result in two issues, so its already supported, the word boundary doesn't change this if I understand you correctly.

@adrianschroeter Since the parsing of the text is happening on the backend, can I pass this on to you to fix the issue?

krauselukas:
@adrianschroeter just to be clear, what I want to prevent is that something like `gh#cli/cli#1` results in the creation of `gh#cli/cli#1`
and `i#1`

krauselukas:
Closing this since it should be fixed on the parser on the backend

PR #12113, opened 2022-01-19
Refactor MultibuildPackage concern

vpereira:
* Move code from accessing the `_multibuild` file from the concern to the backend lib
* Caching the file content for subsequent queries
* Add specs

### DIFF
--- a/src/api/app/lib/backend/api/sources/package.rb
+++ b/src/api/app/lib/backend/api/sources/package.rb
@@ -51,6 +51,14 @@ module Backend
           nil
         end
 
+        # Returns the content of the _multibuild file (if present)
+        # @return [String]
+        def self.multibuild(project_name, package_name)
+          http_get(['/source/:project/:package/_multibuild', project_name, package_name])
+        rescue Backend::NotFoundError
+          nil
+        end
+
         # It triggers all the services of a package
         # @return [String]
         def self.trigger_services(project_name, package_name, user_login)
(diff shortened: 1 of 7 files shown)


src/api/app/lib/backend/api/sources/package.rb
```diff
@@ -51,6 +51,18 @@ def self.service(project_name, package_name)
           nil
         end
 
+        # Returns the content of the _multibuild file (if present)
+        # @return [String]
+        def self.multibuild(project_name, package_name)
+          Rails.cache.fetch("multibuild_#{project_name}_#{package_name}") do
```
hennevogel, line 57: A cache is nice but a cache also need cache invalidation.

In this case the file is controlled by the backend so the backend would need to invalidate the cache. I wouldn't have an idea how...

vpereira: `"multibuild_#{cache_key_with_version}"` would be enough no? The content of this file won't touch without the package container (package model) being changed, right?

vpereira: I don't think the `_multibuild` files changes so often to justify a query to backend everytime we call 
` Package.get_by_project_and_name(target_project_name, target_package_name, follow_multibuild: true)`

hennevogel: > The content of this file won't change without the package container (package model) being changed, right?

Not right. The backend will change this file at will.

PR #12115, opened 2022-01-20
[api] avoid locking database for invalidating cache

adrianschroeter:
- always writing and lock the db to invalidate caches, even when they not get created is not
  the right approach ... Only dropping the memcache entries if existing now without blocking db
  (invalidating caches must not be much more expensive then generating the cached data)

- fixed review handling (reviews can combine the parameters, elsif is wrong therefore)


src/api/app/models/bs_request.rb
```diff
@@ -1199,14 +1199,16 @@ def _assignreview_update_reviews(reviewer, opts, new_review = nil)
     review_comment
   end
 
-  # TODO: Once the feature flag notifications_redesign is removed, remove this method if we decide to delete the "Tasks" page.
```
dmarcoux, line 1202: This comment needs to stay.

dmarcoux (dismissed):

adrianschroeter:
On Donnerstag, 20. Januar 2022, 10:42:47 CET Dany Marcoux wrote:
> @dmarcoux requested changes on this pull request.
> 
> 
> 
> > @@ -1199,14 +1199,16 @@ def _assignreview_update_reviews(reviewer, opts, new_review = nil)
>      review_comment
>    end
>  
> -  # TODO: Once the feature flag notifications_redesign is removed, remove this method if we decide to delete the "Tasks" page.
> 
> This comment needs to stay.

why? the tasks page for sure is not replaceable with notifications, it is a different thing...

-- 

Adrian Schroeter ***@***.***>

dmarcoux:
> why? the tasks page for sure is not replaceable with notifications, it is a different thing...

I am aware that they are different things. As the comment says, we would delete this code if we consider the notifications to be a viable replacement to the tasks page. Since the notifications feature isn't done, we cannot tell whether this will be the case or not. So please, leave this comment there. The comment will be removed whenever the notifications feature toggle is deleted, even if the tasks page stays.

adrianschroeter:
On Donnerstag, 20. Januar 2022, 14:00:56 CET Dany Marcoux wrote:
> > why? the tasks page for sure is not replaceable with notifications, it is a different thing...
> 
> I am aware that they are different things. As the comment says, we would delete this code if we consider the notifications to be a viable replacement to the tasks page. Since the notifications feature isn't done, we cannot tell whether this will be the case or not. So please, leave this comment there. The comment will be removed whenever the notifications feature toggle is deleted, even if the tasks page stays.

hello? tasks are a vital part of our workflows, not only in web interface, also in other parts.
A time based notification system, which even can not be adapted for the past can not be a replacement
by design.

I can add the line again, if this comment is your only concern here, but you are on a fundamental
wrong track regarding the workflows.

-- 

Adrian Schroeter ***@***.***>

src/api/app/models/bs_request.rb
```diff
@@ -1199,14 +1199,16 @@ def _assignreview_update_reviews(reviewer, opts, new_review = nil)
...
     target_package_ids = bs_request_actions.with_target_package.pluck(:target_package_id)
     target_project_ids = bs_request_actions.with_target_project.pluck(:target_project_id)
 
-    user_ids = Relationship.where(package_id: target_package_ids).or(
+    group_users = Relationship.where(package_id: target_package_ids).or(
       Relationship.where(project_id: target_project_ids)
-    ).groups.joins(:groups_users).pluck('groups_users.user_id')
+    ).groups.joins(:groups_users)
+
+    user_ids = group_users.pluck('groups_users.user_id')
```
vpereira, line 1211: Hm, hey @adrianschroeter I don't see what you are trying to achieve, but just separating the commands, won't make them to run separated.  `where`, `select`, `order`, `group`, and other query methods return an instance of `ActiveRecord::Relation`, which represents the results of a SQL query. It doesn’t run the query until you try and access one of the records returned by it. So the query will just just be evaluated when you call `.pluck`

adrianschroeter:
On Freitag, 21. Januar 2022, 11:20:28 CET Victor Pereira wrote:
> @vpereira commented on this pull request.
> 
> 
> 
> >        Relationship.where(project_id: target_project_ids)
> -    ).groups.joins(:groups_users).pluck('groups_users.user_id')
> +    ).groups.joins(:groups_users)
> +
> +    user_ids = group_users.pluck('groups_users.user_id')
> 
> Hm, hey @adrianschroeter I don't see what you are trying to achieve, but just separating the commands, won't make them to run separated.  `where`, `select`, `order`, `group`, and other query methods return an instance of `ActiveRecord::Relation`, which represents the results of a SQL query. It doesn’t run the query until you try and access one of the records returned by it. So the query will just just be evaluated when you call `.pluck` 

group_users is used a second time below and I do not wanted to have the long lines twice.

-- 

Adrian Schroeter ***@***.***>

vpereira:
> group_users is used a second time below and I do not wanted to have the long lines twice.

so could you then extract it to a method instead?

hennevogel:
> always writing and lock the db to invalidate caches, even when they not get created is not the right approach ... Only dropping the memcache entries if existing now without blocking db (invalidating caches must not be much more expensive then generating the cached data)

Using `.updated_at` for cache versioning is standard practice (see [cache_key_with_version](https://api.rubyonrails.org/classes/ActiveRecord/Integration.html#method-i-cache_key_with_version)). In fact there is [fragment caching](https://guides.rubyonrails.org/caching_with_rails.html#fragment-caching) going on in the tasks page that makes use of this (which would break with this change). 

https://github.com/openSUSE/open-build-service/blob/a92564b79bd69989927ef83aa4e01bc6f950f976/src/api/app/views/webui/users/tasks/index.html.haml#L6

It's also not expensive to write a couple of `User` or `Group` rows in the db, why do you think so? However what I do consider expensive (to maintain) is re-implementing this cache invalidation in non-standard Rails practice manually for no apparent reason.

> fixed review handling (reviews can combine the parameters, elsif is wrong therefore)

Not quite sure what you think you fix but there is *either* a `by_user` *or* a `by_group` association. Never both. 

https://github.com/openSUSE/open-build-service/blob/a92564b79bd69989927ef83aa4e01bc6f950f976/src/api/app/models/review.rb#L331-L334

From production:

```
[1] pry(main)> Review.where.not(by_user: nil).where.not(by_group: nil).count
=> 0
```

hennevogel:
Okay I remember why this was a problem: monstrous requests. A `BsRequest` with hundreds of `BsRequestActions` will run this `after_commit` callback and *that* will probably block the `BsRequest` table for as long as it takes to gather the affected people by all `BsRequest` actions.

I guess the real solution to avoid this is to avoid monstrous request *AND* move all of this out of the callback into a job.

PR #12120, opened 2022-01-21
Collect artifacts created on every run worflow step

saraycp:
On every webhook OBS receives from an SCM, OBS follows the instructions given by `workflows.yml` performing the steps on it.

Those steps can handle or create different bunches of artifacts. For example, _repositories_ and _architectures_ if the step is `configure_repositories` or _projects_ and _packages_ in steps like `link_package` etc.

After each workflow run, we want to display all the artifacts involved so the user can better understand what was going on.

In this PR, we are collecting the main information about the artifacts we get on every step and storing them in DB with JSON format.

The JSON for a _link_package_ step can look like this:

```
{
  "source_project": "<source project name>",
  "source_package": "<source package name>",
  "target_project": "<target project name>",
  "target_package": "<target package name>"
}
```

With the information in JSON, we can later display a text and even a link to the artifact from the WorkflowRun show view. 

The view should look like this:

image (1)

### DIFF
--- a/src/api/app/models/workflow.rb
+++ b/src/api/app/models/workflow.rb
@@ -10,7 +10,7 @@ class Workflow
 
   SUPPORTED_FILTERS = [:architectures, :branches, :event, :repositories].freeze
 
-  attr_accessor :workflow_instructions, :scm_webhook, :token
+  attr_accessor :workflow_instructions, :scm_webhook, :token, :workflow_run_id
 
   def initialize(attributes = {})
     super
--- a/src/api/app/models/workflow.rb
+++ b/src/api/app/models/workflow.rb
@@ -31,9 +31,14 @@ class Workflow
       restore_target_projects
     when scm_webhook.new_pull_request?, scm_webhook.updated_pull_request?, scm_webhook.push_event?, scm_webhook.tag_push_event?
       steps.each do |step|
-        step.call({ workflow_filters: filters })
+        call_step_and_collect_artifacts(step)
       end
     end
+  end
+
+  # ArtifactsCollector can only be called if the step.call doesn't return nil because of a validation error
+  def call_step_and_collect_artifacts(step)
+    step.call({ workflow_filters: filters }) && Workflows::ArtifactsCollector.new(step: step, workflow_run_id: workflow_run_id).call
   end
 
   def steps
--- a/src/api/app/models/workflow/step.rb
+++ b/src/api/app/models/workflow/step.rb
@@ -50,6 +50,24 @@ class Workflow::Step
     end
   end
 
+  def target_package_name(short_commit_sha: false)
+    package_name = step_instructions[:target_package] || source_package_name
+
+    case
+    when scm_webhook.pull_request_event?
+      package_name
+    when scm_webhook.push_event?
+      commit_sha = scm_webhook.payload[:commit_sha]
+      if short_commit_sha
+        "#{package_name}-#{commit_sha.slice(0, SHORT_COMMIT_SHA_LENGTH)}"
+      else
+        "#{package_name}-#{commit_sha}"
+      end
+    when scm_webhook.tag_push_event?
+      "#{package_name}-#{scm_webhook.payload[:tag_name]}"
+    end
+  end
+
   protected
 
   def validate_step_instructions
@@ -73,24 +91,6 @@ class Workflow::Step
 
   def target_package_names
     [target_package_name(short_commit_sha: true)] + multibuild_flavors
-  end
-
-  def target_package_name(short_commit_sha: false)
-    package_name = step_instructions[:target_package] || source_package_name
-
-    case
-    when scm_webhook.pull_request_event?
-      package_name
-    when scm_webhook.push_event?
-      commit_sha = scm_webhook.payload[:commit_sha]
-      if short_commit_sha
-        "#{package_name}-#{commit_sha.slice(0, SHORT_COMMIT_SHA_LENGTH)}"
-      else
-        "#{package_name}-#{commit_sha}"
-      end
-    when scm_webhook.tag_push_event?
-      "#{package_name}-#{scm_webhook.payload[:tag_name]}"
-    end
   end
 
   private
--- a/src/api/app/models/workflow_artifacts_per_step.rb
+++ b/src/api/app/models/workflow_artifacts_per_step.rb
@@ -1,0 +1,26 @@
+# One instance of this model represents a bunch of artifacts (related to one step).
+# We chose WorkflowArtifactsPerStep name over names like WorkflowArtifact or WorkflowArtifacts
+# because it fits better semantically and follows Rails convention of being singular.
+class WorkflowArtifactsPerStep < ApplicationRecord
+  belongs_to :workflow_run, optional: false
+
+  serialize :artifacts, JSON
+
+  validates :step, :artifacts, presence: true
+end
+
+# == Schema Information
+#
+# Table name: workflow_artifacts_per_steps
+#
+#  id              :integer          not null, primary key
+#  artifacts       :text(65535)
+#  step            :string(255)
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  workflow_run_id :integer          not null, indexed
+#
+# Indexes
+#
+#  index_workflow_artifacts_per_steps_on_workflow_run_id  (workflow_run_id)
+#
--- a/src/api/app/models/workflow_run.rb
+++ b/src/api/app/models/workflow_run.rb
@@ -3,6 +3,8 @@ class WorkflowRun < ApplicationRecord
   validates :request_headers, :status, presence: true
 
   belongs_to :token, class_name: 'Token::Workflow', optional: true
+  has_many :artifacts, class_name: 'WorkflowArtifactsPerStep', dependent: :destroy
+
   paginates_per 20
 
   enum status: {
--- a/src/api/app/services/workflows/artifacts_collector.rb
+++ b/src/api/app/services/workflows/artifacts_collector.rb
@@ -1,0 +1,24 @@
+module Workflows
+  class ArtifactsCollector
+    def initialize(step:, workflow_run_id:)
+      @step = step
+      @workflow_run_id = workflow_run_id
+    end
+
+    def call
+      artifacts = case @step.class.name
+                  when 'Workflow::Step::BranchPackageStep', 'Workflow::Step::LinkPackageStep'
+                    {
+                      source_project: @step.step_instructions[:source_project],
+                      source_package: @step.step_instructions[:source_package],
+                      target_project: @step.target_project_name,
+                      target_package: @step.target_package_name
+                    }
+                  when 'Workflow::Step::RebuildPackage', 'Workflow::Step::ConfigureRepositories'
+                    @step.step_instructions
+                  end
+
+      WorkflowArtifactsPerStep.find_or_create_by(workflow_run_id: @workflow_run_id, step: @step.class.name, artifacts: artifacts.to_json)
+    end
+  end
+end
--- a/src/api/db/schema.rb
+++ b/src/api/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_12_14_111900) do
+ActiveRecord::Schema.define(version: 2022_01_26_155601) do
 
   create_table "architectures", id: :integer, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", options: "ENGINE=InnoDB ROW_FORMAT=DYNAMIC", force: :cascade do |t|
     t.string "name", null: false, collation: "utf8_general_ci"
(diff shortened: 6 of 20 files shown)


src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -64,4 +66,15 @@ def create_branched_package
...
+
+  def write_report
+    return unless workflow_run = WorkflowRun.find(workflow_run_id)
+
+    # step_key = :branch_package
+    step_key = Workflow::SUPPORTED_STEPS.key(self.class).to_s
+    report = workflow_run.report ? JSON.parse(workflow_run.report) : {}
+    report[step_key] ||= []
+    report[step_key] << { 'source_project' => source_project_name, 'source_package' => source_package_name, 'target_project' => target_project_name, 'target_package' => target_package_name }
+    workflow_run.update(report: report.to_json)
```
hennevogel, line 78: If you store json, why don't you [serialize](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html) it?

src/api/app/models/workflow/step/configure_repositories.rb
```diff
@@ -64,4 +68,15 @@ def validate_project_name
 
     errors.add(:base, "invalid project '#{target_project_base_name}'")
   end
+
+  def write_report
+    return unless workflow_run = WorkflowRun.find(workflow_run_id)
```
hennevogel, line 73: Looks to me like you want to dry this up in `Workflow::Step`. The only difference between the step types is the json content right?

Would also greatly simplify the specs you need to touch etc.

src/api/app/models/workflow/step/branch_package_step.rb
danidoni, line 78: I would change `report` to `artifacts`

hennevogel (changes requested):

src/api/app/models/workflow/step/configure_repositories.rb
saraycp: I did some refactoring already. I moved the `write_report` to the Step model and in every specific step, I simply define the method `artifacts` that returns only the hash with the information.

src/api/app/models/workflow/step/branch_package_step.rb
saraycp: Sorry, I stashed that change by mistake.

src/api/db/schema.rb
```diff
@@ -1066,6 +1066,7 @@
     t.datetime "updated_at", precision: 6, null: false
     t.integer "token_id", null: false
     t.string "response_url"
+    t.text "report"
```
danidoni, line 1069: This would be "artifacts", right?

src/api/app/models/workflow/step/branch_package_step.rb
saraycp: Included in my last push now: https://github.com/openSUSE/open-build-service/compare/b73b7db0ad4b62124223778927f89056fced568e..428c6fb5e0ab662dac76b84dfe6d5893da46d0d8

saraycp: Yes, why not. Let me finish everything using "report" and then I'll replace all the code.

src/api/db/schema.rb
saraycp: could be, yes

src/api/app/models/workflow/step.rb
```diff
@@ -93,8 +93,25 @@ def target_package_name(short_commit_sha: false)
     end
   end
 
+  def write_report
+    return unless (workflow_run = WorkflowRun.find(workflow_run_id))
```
hennevogel, line 97:
```suggestion
    workflow_run = WorkflowRun.find(workflow_run_id)
    return unless workflow_run
```

hennevogel: Don't make me think :)

src/api/spec/models/workflow/step/rebuild_package_step_spec.rb
```diff
@@ -21,10 +22,19 @@
                    })
   end
 
+  let(:written_report) do
```
hennevogel, line 25: No need for this for every `::Step` variant right?

hennevogel (changes requested):

src/api/app/models/workflow/step.rb
saraycp: changed

saraycp:
@danidoni @hennevogel, I have renamed `report` to `artifacts`.
I've also removed the commits related to each specific `::Step`, I'll try to test the artifacts only in `spec/models/workflow/step_spec.rb`

src/api/app/models/workflow_run.rb
```diff
@@ -2,6 +2,8 @@ class WorkflowRun < ApplicationRecord
   validates :response_url, length: { maximum: 255 }
   validates :request_headers, :status, presence: true
 
+  serialize :report, JSON
```
dmarcoux, line 5: This should be `artifacts`, right?

src/api/app/services/workflows/artifacts_collector.rb
```diff
@@ -0,0 +1,49 @@
...
+      artifacts = case step.type # example: branch_package, configure_repositories, etc... or maybe use the class names like Workflow::Step::ConfigureRepositories, etc...
+                  when 'branch_package'
+                    { step.source_project, step.source_package, step.target_project, step.target_package }
+                  when 'link_package'
+                    { step.source_project, step.source_package, step.target_project, step.target_package }
+                  when 'configure_repositories'
+                    { step.project, step.repositories, step.architectures }
+                  when 'rebuild_package'
+                    { step.project, step.package }
+                  end
```
hennevogel, line 19: How do you know what is what? By position in the json?

dmarcoux: For the UI, there will be a view component which knows how to handle every kind of step. With a similar case statement, it will display the artifacts correctly by knowing which step has which keys.

src/api/app/models/workflow_artifacts_per_step.rb
```diff
@@ -0,0 +1,21 @@
+class WorkflowArtifactsPerStep < ApplicationRecord
+  belongs_to :workflow_run, required: true
+
+  serialize :artifacts, JSON
+end
+
+# == Schema Information
+#
+# Table name: workflow_artifacts_per_steps
```
hennevogel, line 9: Not quite sure why you would use that implementation detail (that you show them by step) into the table name.

src/api/app/services/workflows/artifacts_collector.rb
hennevogel: I see, this is just pseudo code...

src/api/app/models/workflow_artifacts_per_step.rb
saraycp: The naming is a problem here.

The model, let's call it X, represents a bunch of artifacts related to one step in a JSON format  (i.e. target_package, repositories etc.)

**WorkflowArtifacts**

The first idea was to call the method `WorkflowArtifacts` plural, because is the name that better represents what I described before. Choosing this method name implies not following rails standards and changing the **inflector**.

**WorkflowArtifact**

An alternative is just to call it `WorkflowArtifact` singular. But is not so good semantically. Because one instance of that model doesn't represent one artifact but many (related to one step).

Moreover, having something like `WorkflowArtifact#artifacts` sounds weird.

**WorkflowArtifactsPerStep**

That's why I chose `WorkflowArtifactsPerStep` which semantically is perfect, follows the standard of being singular, and `WorkflowArtifactsPerStep#artifacts` sounds better.

Please let me know if any of you have a better name for the model or for the column `X#artifacts`.

hennevogel: I see, thanks for the explanation.

src/api/app/models/workflow.rb
```diff
@@ -31,11 +31,17 @@ def call
       restore_target_projects
     when scm_webhook.new_pull_request?, scm_webhook.updated_pull_request?, scm_webhook.push_event?, scm_webhook.tag_push_event?
       steps.each do |step|
-        step.call({ workflow_filters: filters })
+        # TODO: improve this call.
```
dmarcoux, line 34: What should be improved exactly? Will it happen in this PR?

src/api/app/services/workflows/artifacts_collector.rb
```diff
@@ -0,0 +1,40 @@
...
+                      target_project: @step.target_project_name,
+                      target_package: @step.target_package_name
+                    }
+                  when 'Workflow::Step::RebuildPackage'
+                    {
+                      project: @step.step_instructions[:project],
+                      package: @step.step_instructions[:package]
+                    }
+                  when 'Workflow::Step::ConfigureRepositories'
+                    # We only need the repositories' names and their archtectures.
```
dmarcoux, line 24:
```suggestion
                    # We only need the repositories' names and their architectures.
```

src/api/app/services/workflows/artifacts_collector.rb
```diff
@@ -0,0 +1,40 @@
...
+                    repositories_and_architectures = @step.step_instructions[:repositories].map do |repository|
+                      repository.delete(:target_project)
```
dmarcoux, line 26: With the changes from #12128, this needs to be updated:

```suggestion
                      repository.delete(:paths)
```

src/api/app/models/workflow.rb
saraycp: Sorry, I forgot to remove the comment (on it right now). It was added before I moved the code to the method `call_step_and_collect_artifacts` because of high cyclomatic complexity.

src/api/app/services/workflows/artifacts_collector.rb
```diff
@@ -0,0 +1,40 @@
...
+      artifacts = case @step.class.name
+                  when 'Workflow::Step::BranchPackageStep', 'Workflow::Step::LinkPackageStep'
+                    {
+                      source_project: @step.step_instructions[:source_project],
+                      source_package: @step.step_instructions[:source_package],
+                      target_project: @step.target_project_name,
+                      target_package: @step.target_package_name
+                    }
+                  when 'Workflow::Step::RebuildPackage'
+                    {
```
dmarcoux, line 19: This could also be written as `@step.step_instructions.slice(:project, :package)`.

src/api/app/services/workflows/artifacts_collector.rb
saraycp: thanks! done

src/api/app/services/workflows/artifacts_collector.rb
saraycp: good catch!

src/api/app/services/workflows/artifacts_collector.rb
saraycp: corrected

src/api/app/models/workflow.rb
saraycp: comment removed

saraycp:
@dmarcoux thanks for your feedback. I did

- rebase on master
- adapt code and specs to deal with the new `paths:` for repositories,
- change/removed/added some comments to the code

src/api/app/services/workflows/artifacts_collector.rb
```diff
@@ -0,0 +1,35 @@
...
+                      source_package: @step.step_instructions[:source_package],
+                      target_project: @step.target_project_name,
+                      target_package: @step.target_package_name
+                    }
+                  when 'Workflow::Step::RebuildPackage'
+                    @step.step_instructions
+                  when 'Workflow::Step::ConfigureRepositories'
+                    # We only need the repositories' names and their architectures.
+                    repositories_and_architectures = @step.step_instructions[:repositories].map do |repository|
+                      repository.delete(:paths)
```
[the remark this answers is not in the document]
saraycp: I'm not so sure the paths are so important :thinking:  but it doesn't harm to display them as we display the architectures.

I'm working on the UI right now. I'll tackle the paths as an improvement as soon as the UI is ready for review.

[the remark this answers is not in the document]
dmarcoux: When building a package, its dependencies are fetched from the package's project and its repositories. If a repository doesn't have a repository path, it's not providing anything. So yes, repository paths are important!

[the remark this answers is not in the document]
saraycp: I meant that I wasn't sure that they are important to be in the artifacts, as we are going to provide a link to the repositories tab of the project, where the user can see all the details about them.
But yes, as I said, it's not a problem for me to add them to the artifacts as well.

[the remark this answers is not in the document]
saraycp: @dmarcoux The paths are already included in the artifacts. Please have a look to the last update of this PR.

rubhanazeem (approved):

hennevogel (approved):

PR #12123, opened 2022-01-24
Implement filters for Workflow Runs

danidoni:
This PR implements several filters to help you narrow down your search
for a specific workflow run.

It filters by workflow run status (success, running, failure) and by
source event (pull request, push, both GitHub and GitLab).

image

*TODO*

- [x] Finish specs for the new WorkflowRunFilterComponent
- [x] Write specs for WorkflowRunFilterLinkComponent
- [x] Write spec for WorkflowRunsFinder

### DIFF
--- a/src/api/app/components/workflow_run_filter_component.html.haml
+++ b/src/api/app/components/workflow_run_filter_component.html.haml
@@ -1,0 +1,21 @@
+.row.list-group-flush
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'All', amount: @count['all'],
+                                              filter_item: {}, selected_filter: @selected_filter)
+.row.list-group-flush.mt-5
+  %h5.ml-3 Status
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Succeeded', amount: @count['success'],
+                                              filter_item: { status: 'success' }, selected_filter: @selected_filter)
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Running', amount: @count['running'],
+                                              filter_item: { status: 'running' }, selected_filter: @selected_filter)
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Failed', amount: @count['fail'],
+                                              filter_item: { status: 'fail' }, selected_filter: @selected_filter)
+.row.list-group-flush.mt-5
+  %h5.ml-3 Event type
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Pull Requests', amount: @count['pull_request'],
+                                              filter_item: { event_type: 'pull_request' }, selected_filter: @selected_filter)
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Push', amount: @count['push'],
+                                              filter_item: { event_type: 'push' }, selected_filter: @selected_filter)
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Merge Request Hook', amount: @count['Merge Request Hook'],
+                                              filter_item: { event_type: 'Merge Request Hook' }, selected_filter: @selected_filter)
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Push Hook', amount: @count['Push Hook'],
+                                              filter_item: { event_type: 'Push Hook' }, selected_filter: @selected_filter)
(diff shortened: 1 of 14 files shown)


danidoni:
[Improvements for WorkflowRuns Index UI](https://trello.com/c/3axxhgrd/1820-improvements-for-workflowruns-index-ui)

src/api/app/components/workflow_run_filter_link_component.rb
```diff
@@ -0,0 +1,38 @@
...
+      @filter_item[:status] == @selected_filter[:status]
+    elsif @selected_filter[:event_type].present?
+      @filter_item[:event_type] == @selected_filter[:event_type]
+    elsif @selected_filter.empty?
+      @filter_item.empty?
+    end
+  end
+
+  # This method won't be needed in Ruby 2.6+ by using the option `exception: true` for Integer(...), it's then a one-liner
+  def ensure_integer_amount(amount)
```
krauselukas, line 33: @danidoni is this really necessary? on all items you did a `count` in the finder class, dont see how this should be a string :)

danidoni: Hmm, let me check

danidoni: True, removed

src/api/app/components/workflow_run_filter_component.html.haml
```diff
@@ -0,0 +1,21 @@
...
+  %h5.ml-3 Status
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Succeeded', amount: @count['success'],
+                                              filter_item: { status: 'success' }, selected_filter: @selected_filter)
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Running', amount: @count['running'],
+                                              filter_item: { status: 'running' }, selected_filter: @selected_filter)
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Failed', amount: @count['fail'],
+                                              filter_item: { status: 'fail' }, selected_filter: @selected_filter)
+.row.list-group-flush.mt-5
+  %h5.ml-3 Event type
+  = render WorkflowRunFilterLinkComponent.new(token: @token, text: 'Pull Requests', amount: @count['pull_request'],
```
dmarcoux, line 14: I don't believe we should split up pull requests and merge requests. Conceptually, they are the same, so I would put them under the same filter. Same for `Push` and `Push Hook`.

On top of this, having a filter for push events related to tags could be useful. What do you think?

dmarcoux:
[Filter workflow runs by state, event, action](https://trello.com/c/JLlyTsPD/2483-filter-workflow-runs-by-state-event-action)

src/api/app/policies/workflow_run_policy.rb
```diff
@@ -1,9 +1,11 @@
 class WorkflowRunPolicy < ApplicationPolicy
-  class Scope
+  class Scope < Scope
     def initialize(user, scope, opts = {})
       @user = user
       @scope = scope
       @opts = opts
+
+      #super(user, scope)
```
hennevogel, line 8: Hm, what's this about?

krauselukas: Commenting this line was not on purpose. Why changing this policy now, maybe @danidoni can explain why this happend?

danidoni: I don't remember changing this line, so probably was an accident

src/api/app/components/workflow_run_filter_component.html.haml
krauselukas: As discussed, this is going to happen in an follow up, we created a card in the backlog for that

danidoni:
LGTM!

src/api/app/policies/workflow_run_policy.rb
```diff
@@ -1,9 +1,11 @@
 class WorkflowRunPolicy < ApplicationPolicy
-  class Scope
+  class Scope < Scope
     def initialize(user, scope, opts = {})
       @user = user
       @scope = scope
       @opts = opts
+
+      super(user, scope)
```
hennevogel, line 8: I don't like this uncommented change to this policy inside a large commit. Please separate this commit and explain it in the commit message.

hennevogel (changes requested):

src/api/app/policies/workflow_run_policy.rb
krauselukas: @hennevogel I don't see a reason to touch the policy at all in this PR, I removed those changes.

saraycp:
I suggest a correction for the commit description "*Move hook_event logic from workflow_run_row view component to model*"  `s/the model it self then the view component./the model itself than the view component.`

PR #12131, opened 2022-01-27
Add step `set_flags`

rubhanazeem:
Added a step to set flags. With this new step, it's possible to set flags for a project or package. Limiting the flag to a certain repository or architecture is possible as well.

### DIFF
--- a/src/api/app/models/workflow/step/set_flags.rb
+++ b/src/api/app/models/workflow/step/set_flags.rb
@@ -1,0 +1,83 @@
+class Workflow::Step::SetFlags < ::Workflow::Step
+  REQUIRED_KEYS = [:flags].freeze
+  REQUIRED_FLAG_KEYS = [:type, :status, :project].freeze
+  OPTIONAL_FLAG_KEYS = [:package, :repository, :architecture].freeze
+
+  validate :validate_flags
+
+  def call(_options = {})
+    return unless valid?
+
+    set_flags
+  end
+
+  def flags
+    step_instructions[:flags]
+  end
+
+  private
+
+  def set_flags
+    ActiveRecord::Base.transaction do
+      flags.each do |flag|
+        main_object = project_or_package(flag)
+        check_access(main_object)
+        architecture_id = Architecture.find_by_name(flag[:architecture]).id if flag[:architecture]
+        existing_flag = main_object.flags.exists?(flag: flag[:type], repo: flag[:repository], architecture_id: architecture_id)
+        next if existing_flag
+
+        main_object.add_flag(flag[:type], flag[:status], flag[:repository], flag[:architecture])
+        main_object.save!
+      end
+    end
+  end
+
+  def project_or_package(flag)
+    project = Project.find_by!(name: target_project_name(project_name: flag[:project]))
+    package = project.packages.find_by(name: target_package_name(package_name: flag[:package])) if flag[:package].present?
+    package.presence || project
+  end
+
+  def check_access(object)
+    raise Pundit::NotAuthorizedError unless Pundit.policy(token.user, object).update?
+  end
+
+  def validate_flags
+    return if flags.all? { |flag| (REQUIRED_FLAG_KEYS - flag.keys).empty? }
+
+    required_flag_keys_sentence ||= REQUIRED_FLAG_KEYS.map { |key| "'#{key}'" }.to_sentence
+    errors.add(:base, "set_flags step: All flags must have the #{required_flag_keys_sentence} keys")
+  end
+
+  # TODO: Totally duplicated from Workflow::Step. Remove the duplication by using a service instead for all steps depending on this method.
+  def target_project_name(project_name:)
+    return project_name if scm_webhook.push_event? || scm_webhook.tag_push_event?
+
+    return nil unless scm_webhook.pull_request_event?
+
+    pr_subproject_name = if scm_webhook.payload[:scm] == 'github'
+                           scm_webhook.payload[:target_repository_full_name]&.tr('/', ':')
+                         else
+                           scm_webhook.payload[:path_with_namespace]&.tr('/', ':')
+                         end
+
+    "#{project_name}:#{pr_subproject_name}:PR-#{scm_webhook.payload[:pr_number]}"
+  end
+
+  # TODO: Totally duplicated from Workflow::Step. Remove the duplication by using a service instead for all steps depending on this method.
+  def target_package_name(package_name:, short_commit_sha: false)
+    case
+    when scm_webhook.pull_request_event?
+      package_name
+    when scm_webhook.push_event?
+      commit_sha = scm_webhook.payload[:commit_sha]
+      if short_commit_sha
+        "#{package_name}-#{commit_sha.slice(0, SHORT_COMMIT_SHA_LENGTH)}"
+      else
+        "#{package_name}-#{commit_sha}"
+      end
+    when scm_webhook.tag_push_event?
+      "#{package_name}-#{scm_webhook.payload[:tag_name]}"
+    end
+  end
+end
--- a/src/api/spec/models/workflow/step/set_flags_step_spec.rb
+++ b/src/api/spec/models/workflow/step/set_flags_step_spec.rb
@@ -1,0 +1,231 @@
+require 'rails_helper'
+
+RSpec.describe Workflow::Step::SetFlags do
+  let(:user) { create(:confirmed_user, :with_home, login: 'Iggy') }
+  let(:token) { create(:workflow_token, user: user) }
+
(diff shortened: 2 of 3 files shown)


src/api/app/models/workflow/step/set_flags.rb
```diff
@@ -0,0 +1,66 @@
...
+    flags.each do |flag|
+      project = Project.find_by(name: flag[:project])
+      package = if project.present? && flag[:package].present?
+                  project.packages.find_by(name: flag[:package])
+                else
+                  nil
+                end
+
+      main_object = package.present? ? package : project
+      Pundit.authorize(@main_object, :update?)
```
vpereira, line 31: I think here it shouldn't be `@main_object` but `main_object`, right?

rubhanazeem: Oh right.

src/api/app/models/workflow/step/set_flags.rb
```diff
@@ -0,0 +1,55 @@
...
+  private
+
+  def set_flags
+    ActiveRecord::Base.transaction do
+      flags.each do |flag|
+        project = Project.find_by(name: flag[:project])
+        package = project.packages.find_by(name: flag[:package]) if project.present? && flag[:package].present?
+        main_object = package.presence || project
+        check_access(main_object)
+        main_object.add_flag(flag[:type], flag[:status], flag[:repository], flag[:architecture]).save
```
hennevogel, line 28: What about error reporting in case of `.save` going wrong?

src/api/app/models/workflow/step/set_flags.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def check_access(object)
+    raise Pundit::NotAuthorizedError unless Pundit.policy(token.user, object).update?
+  end
+
+  # rubocop:disable Metrics/CyclomaticComplexity
+  def validate_project_and_package
+    flags.each do |flag|
+      project = Project.find_by(name: flag[:project])
+      package_error = !(project.nil? || flag[:package].nil? || (flag[:package].present? && project.present? && project.packages.find_by(name: flag[:package]).present?))
```
hennevogel, line 41: This is basically not understandable. How about not lumping those many checks together?

DavidKang: Agree with @hennevogel. IMHO, you could create a method, something like:

```ruby
def package_error?(project, flag)
  return false if project.nil? || flag[:package].nil?
  return false if flag[:package].present? && project.present? && project.packages.find_by(name: flag[:package]).present?
  return true
end
```

This should also reduce the `Metrics/CyclomaticComplexity`

dmarcoux:
About the PR title and the commit message, it's redundant to write `add new XYZ`. Writing `add XYZ` is already enough.

src/api/app/models/workflow/step/set_flags.rb
```diff
@@ -0,0 +1,95 @@
...
+        main_object.save!
+      end
+    end
+  end
+
+  def check_access(object)
+    raise Pundit::NotAuthorizedError unless Pundit.policy(token.user, object).update?
+  end
+
+  def validate_project_and_package
```
dmarcoux, line 43: @rubhanazeem and I talked about this validation and it won't be possible to validate the existence of projects/packages. `set_flags` depends on records (projects/packages/repositories) created by previous steps in a workflow, then applies flags to those records. Since we validate **all** steps before running them, we cannot ensure the existence of those records, so this validation cannot reliably work. Validating step one by one is not possible without having to potentially run only parts of a workflow if certain steps are non-valid. This can lead to major issues and this is why we started to validate **all** steps before running them.

src/api/app/models/workflow/step/set_flags.rb
```diff
@@ -0,0 +1,82 @@
...
+    step_instructions[:flags]
+  end
+
+  private
+
+  def set_flags
+    ActiveRecord::Base.transaction do
+      flags.each do |flag|
+        project = Project.find_by(name: target_project_name(project_name: flag[:project]))
+        raise ActiveRecord::RecordNotFound, "Couldn't find Project #{flag[:project]}" if project.nil?
```
vpereira, line 24: if you are doing a `find_by` and raising `ActiveRecord::RecordNotFound`, couldn't you use `find_by!` ?

src/api/app/models/workflow/step/set_flags.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def validate_flags
+    return if flags.all? { |flag| (REQUIRED_FLAG_KEYS - flag.keys).empty? }
+
+    required_flag_keys_sentence ||= REQUIRED_FLAG_KEYS.map { |key| "'#{key}'" }.to_sentence
+    errors.add(:base, "set_flags step: All flags must have the #{required_flag_keys_sentence} keys")
+  end
+
+  # TODO: Totally duplicated from Workflow::Step. Remove the duplication by using a service instead for all steps depending on this method.
```
dmarcoux, line 51: We're tracking this refactoring in our [backlog](https://trello.com/c/WrOjCG1m/2490-remove-duplicated-methods-targetprojectname-and-targetpackagename).

src/api/app/models/workflow/step/set_flags.rb
```diff
@@ -0,0 +1,82 @@
...
+    pr_subproject_name = if scm_webhook.payload[:scm] == 'github'
+                           scm_webhook.payload[:target_repository_full_name]&.tr('/', ':')
+                         else
+                           scm_webhook.payload[:path_with_namespace]&.tr('/', ':')
+                         end
+
+    "#{project_name}:#{pr_subproject_name}:PR-#{scm_webhook.payload[:pr_number]}"
+  end
+
+  # TODO: Totally duplicated from Workflow::Step. Remove the duplication by using a service instead for all steps depending on this method.
```
dmarcoux, line 66: Same as [above](https://github.com/openSUSE/open-build-service/pull/12131/files#r797497094).

src/api/app/models/workflow/step/set_flags.rb
rubhanazeem: Good point. It's updated.

src/api/spec/models/workflow/step/set_flags_step_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+                project: 'home:Iggy',
+                package: 'ctris'
+              }
+            ]
+        }
+      end
+
+      before do
+        # rubocop:disable RSpec/AnyInstance
+        allow_any_instance_of(Package).to receive(:commit_user).and_return(user)
```
vpereira, line 128: couldn't you just do it for `target_package`, i.e:

```allow(target_package).to receive(:commit_user).and_return(user)``` ?

File: src/api/spec/models/workflow/step/set_flags_step_spec.rb [hunk 9]
```diff
@@ -0,0 +1,233 @@
...
+                project: 'home:Iggy',
+                repository: 'openSUSE_Tumbleweed',
+                architecture: 'x86_64'
+              }
+            ]
+        }
+      end
+
+      before do
+        target_project.add_flag('build', 'enable', 'openSUSE_Tumbleweed', 'x86_64')
```
vpereira, line 167: It could become a factory. If you look on `spec/factories/project.rb` we have factories for `forbidden_project` and `locked_project`.. we could for sure have one `build_enabled_project`

src/api/spec/models/workflow/step/set_flags_step_spec.rb
rubhanazeem: I know and I've tried using `allow` it doesn't solve the problem. `commit_user` always stays empty even when you add `commit_user` in `create`.

src/api/spec/models/workflow/step/set_flags_step_spec.rb
rubhanazeem: That's right we can have a new factory but we don't need one in this case because I have a single instance of `target_project`. I need the flag in only one example.

src/api/spec/models/workflow/step/set_flags_step_spec.rb
vpereira: you should login or set the `User.session = user` before creating the package. It is the same issue if you try to push something to backend in the rails console and `User.session == nil`

rubhanazeem: Oh wow. That fixed the problem. Thanks.

dmarcoux (approved):

krauselukas (approved):
Nice!

PR #12154, opened 2022-02-01
Create an initial generic status

vpereira:
To give a feedback to the users as soon as possible, we are setting initially a generic status to the PR. Package name, multibuild flavors and other details are going to be reported by OBS later with the delayed job `ReportToSCMJob`.

Screenshot when we first create the PR:

Screenshot_2022-02-02_17-27-10

Screenshot when the build process reported back:

Screenshot_2022-02-02_17-32-34

Another thing being changed in this PR is the way that we report back to SCM. Until now we were adding to specific steps, the ability to report back. Now we are trying to come up with a more holistic approach to the problem.

From this card, we have some TODO that should become cards in our backlog:

- [x] For the OBS integration workflow initialization, we would link it to the `workflow_runs` associated with a token (we still have to figure out how to make it public) (card created in the backlog for that)
- [x] Have an event `Event::BuildStarted` coming from backend. Then we could provide the user with more feedback other than "pending" and "done" (card created in the backlog for that)
- ~~[ ] Handle event filtering in the workflow level, not in the steps level. The filters aren't just for reporting~~.

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -17,7 +17,12 @@ class Token::Workflow < Token
     yaml_file = Workflows::YAMLDownloader.new(@scm_webhook.payload, token: self).call
     @workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: @scm_webhook, token: self, workflow_run_id: options[:workflow_run].id).call
 
-    @workflows.each(&:call) if validation_errors.none?
+    return validation_errors unless validation_errors.none?
+
+    # This is just an initial generic report to give a feedback asap. Initial status pending
+    ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token).call
+    @workflows.each(&:call)
+    ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, 'success').call
 
     # Always returning validation errors to report them back to the SCM in order to help users debug their workflows
     validation_errors
--- a/src/api/app/models/workflow/step/branch_package_step.rb
+++ b/src/api/app/models/workflow/step/branch_package_step.rb
@@ -21,10 +21,7 @@ class Workflow::Step::BranchPackageStep < ::Workflow::Step
     add_branch_request_file(package: target_package)
 
     # SCMs don't support statuses for tags, so we don't need to report back in this case
-    unless scm_webhook.tag_push_event?
-      create_or_update_subscriptions(target_package, workflow_filters)
-      report_to_scm(workflow_filters)
-    end
+    create_or_update_subscriptions(target_package, workflow_filters) unless scm_webhook.tag_push_event?
 
     target_package
   end
--- a/src/api/app/services/scm_initial_status_reporter.rb
+++ b/src/api/app/services/scm_initial_status_reporter.rb
@@ -1,0 +1,14 @@
+class ScmInitialStatusReporter < SCMStatusReporter
+  attr_accessor :state
+
+  def initialize(event_payload, event_subscription_payload, scm_token, event_type = nil)
+    super(event_payload, event_subscription_payload, scm_token)
+    @state = event_type.nil? ? 'pending' : 'success'
+  end
+
+  private
+
+  def status_options
+    { context: 'OBS SCM/CI Workflow Integration started' }
+  end
+end
--- a/src/api/app/models/workflow/step/link_package_step.rb
+++ b/src/api/app/models/workflow/step/link_package_step.rb
@@ -18,10 +18,7 @@ class Workflow::Step::LinkPackageStep < ::Workflow::Step
     add_branch_request_file(package: target_package)
 
     # SCMs don't support statuses for tags, so we don't need to report back in this case
-    unless scm_webhook.tag_push_event?
-      create_or_update_subscriptions(target_package, workflow_filters)
-      report_to_scm(workflow_filters)
-    end
+    create_or_update_subscriptions(target_package, workflow_filters) unless scm_webhook.tag_push_event?
 
     target_package
   end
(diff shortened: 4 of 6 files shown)


src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -21,10 +21,7 @@ def branch_package(workflow_filters = {})
     add_branch_request_file(package: target_package)
 
     # SCMs don't support statuses for tags, so we don't need to report back in this case
-    unless scm_webhook.tag_push_event?
-      create_or_update_subscriptions(target_package, workflow_filters)
-      report_to_scm(workflow_filters)
-    end
+    create_or_update_subscriptions(target_package, workflow_filters) unless scm_webhook.tag_push_event?
```
vpereira, line 24: We should evaluate the workflow_filters, in the workflow level, not inside the steps. As it is now, when I create a new PR via github UI (git generates a push event), or push to github in a new branch, even if i have an event filter as described here https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration#event-filter, filtering just for pull requests, I still run all the workflow steps.

dmarcoux: We're passing the workflow filters to this method in order to have them in the event subscription payload. This is needed for the `ReportToScmJob`:

https://github.com/openSUSE/open-build-service/blob/944d7ec129ffe0aa2fb1fd05ced5d035d88c9d0a/src/api/app/jobs/report_to_scm_job.rb#L34-L40

src/api/app/services/scm_initial_status_reporter.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def initialize(event_payload, event_subscription_payload, scm_token, event_type = nil)
+    super(event_payload, event_subscription_payload, scm_token)
+    @state = event_type.nil? ? 'pending' : 'success'
+  end
+
+  private
+
+  def status_options
+    { context: 'OBS integration workflow initialized' }
```
hennevogel, line 12:
```suggestion
    { context: 'OBS SCM/CI Workflow Integration started' }
```

src/api/app/models/token/workflow.rb
```diff
@@ -17,7 +17,12 @@ def call(options)
     yaml_file = Workflows::YAMLDownloader.new(@scm_webhook.payload, token: self).call
     @workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: @scm_webhook, token: self, workflow_run_id: options[:workflow_run].id).call
 
-    @workflows.each(&:call) if validation_errors.none?
+    return validation_errors unless validation_errors.none?
+
+    # This is just an initial generic report to give a feedback asap. Initial status pending
+    ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token).call
+    @workflows.each(&:call)
+    ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, 'success').call
```
hennevogel, line 25: So what happens if all the workflows fails? The initial status will be forever in yellow right?

Can we handle this?

src/api/app/models/workflow/step/branch_package_step.rb
dmarcoux: As for running all steps when we pass an event filter, this sounds definitely wrong what we're doing. Steps shouldn't be called if the webhook event doesn't match the event filter.

dmarcoux:
Regarding `Handle event filtering in the workflow level, not in the steps level. The filters aren't just for reporting.`, this is already implemented in the `Workflow` model:

https://github.com/openSUSE/open-build-service/blob/3784e47e4b37586f2f875df60963353595a37437/src/api/app/models/workflow.rb#L23-L24

https://github.com/openSUSE/open-build-service/blob/3784e47e4b37586f2f875df60963353595a37437/src/api/app/models/workflow.rb#L80-L93

src/api/app/models/workflow/step/branch_package_step.rb
dmarcoux: I had a look and it should be working fine. See this [comment](https://github.com/openSUSE/open-build-service/pull/12154#issuecomment-1028856756). Is it not working somehow?

src/api/app/models/token/workflow.rb
vpereira: Yes, that was my wish too.. that for example, if any exception happens while running the workflows, it could be reflected as well in that message. I still have to figure out how to make it.. i was thinking in something like:

```
# body of the method call
workflow_status = 'success'
ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token).call
@workflows.each(&:call)
rescue StandardError
 workflow_state = 'failure'
ensure 
  ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, workflow_state).call
  validation_errors 
end
```

src/api/app/services/scm_initial_status_reporter.rb
vpereira: done :+1:

vpereira:
> Regarding `Handle event filtering in the workflow level, not in the steps level. The filters aren't just for reporting.`, this is already implemented in the `Workflow` model:

Yes, I see we have this in the `models/workflow.rb` and not on the `models/token/workflow.rb`, which will run the code:

```
    # models/token/workflow.rb 
    # This is just an initial generic report to give a feedback asap. Initial status pending
    ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token).call
    @workflows.each(&:call)
    ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, 'success').call
```
regardless of the filter, right? Should I pass the filter around and test for it?

src/api/app/models/token/workflow.rb
hennevogel: Okay, doesn't have to be here. We can push this out if you want to.

vpereira: yes, I spent sometime golfing around with this code and I'm still not happy with the solution using `ensure`

hennevogel (approved):

src/api/app/models/token/workflow.rb
hennevogel: please do not forget to create a card in the backlog

vpereira: https://trello.com/c/HDjqdG9t/2501-make-the-initial-status-report-more-robust

PR #12157, opened 2022-02-01
Workflow artifacts UI

saraycp:
Implement view component to display the artifacts per step.

As a first approach, the view looks like this:

Screenshot 2022-02-03 at 10-50-42 Open Build Service

This is a follow-up of https://github.com/openSUSE/open-build-service/pull/12120.

**TODO:**

- [X] finish implementing configure_respositories part of the component and preview
- [X] component specs
- [X] configure_repositories view component spec

### DIFF
--- a/src/api/app/components/workflow_artifacts_per_step_component.rb
+++ b/src/api/app/components/workflow_artifacts_per_step_component.rb
@@ -1,0 +1,89 @@
+class WorkflowArtifactsPerStepComponent < ApplicationComponent
+  with_collection_parameter :artifacts_per_step
+
+  attr_reader :artifacts_per_step, :step, :artifacts
+
+  def initialize(artifacts_per_step:)
+    super
+    @artifacts_per_step = artifacts_per_step
+    @step = @artifacts_per_step.step
+    @artifacts = @artifacts_per_step.artifacts
+  end
+
+  def call
+    parsed_artifacts = JSON.parse(artifacts).deep_symbolize_keys
+
+    case step
+    when 'Workflow::Step::BranchPackageStep'
+      artifacts_for_branch_or_link_package(parsed_artifacts, 'Branched package from ')
+    when 'Workflow::Step::LinkPackageStep'
+      artifacts_for_branch_or_link_package(parsed_artifacts, 'Linked package from ')
+    when 'Workflow::Step::RebuildPackage'
+      artifacts_for_rebuild_package(parsed_artifacts)
+    when 'Workflow::Step::ConfigureRepositories'
+      artifacts_for_configure_repositories(parsed_artifacts)
+    end
+  rescue JSON::ParserError, ActionController::UrlGenerationError => e
+    Airbrake.notify(e, artifacts_per_step_id: artifacts_per_step.id)
+    tag.li { concat("Could not display artifacts for #{step.split('::').last.titleize}") }
+  end
+
+  private
+
+  def artifacts_for_branch_or_link_package(parsed_artifacts, label)
+    source_path = helpers.package_show_path(project: parsed_artifacts[:source_project], package: parsed_artifacts[:source_package])
+    target_path = helpers.package_show_path(project: parsed_artifacts[:target_project], package: parsed_artifacts[:target_package])
+
+    tag.li do
+      concat(label)
+      concat(link_to("#{parsed_artifacts[:source_project]}/#{parsed_artifacts[:source_package]}", source_path))
+      concat(' to ')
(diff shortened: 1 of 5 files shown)


src/api/app/components/workflow_artifacts_per_step_component.html.haml
```diff
@@ -0,0 +1 @@
+= artifacts_sentence(self)
```
dmarcoux, line 1: Since this template is only calling a single method of the view component, I suggest renaming the `artifacts_sentence` method to `call` and deleting this Haml template. It would achieve the same. Here's a reference about this from [view_component's documentation](https://viewcomponent.org/guide/templates.html#inline).

src/api/app/components/workflow_artifacts_per_step_component.rb
```diff
@@ -0,0 +1,90 @@
...
+        concat(tag.li { repository_sentence(repository) })
+      end
+    end
+  end
+
+  def repository_sentence(repository)
+    concat(tag.span("#{repository[:name]}", class: 'font-italic'))
+    concat(" for architectures ")
+    concat(tag.span("#{repository[:architectures].to_sentence}", class: 'font-italic'))
+    concat(" for the following paths: ")
```
dmarcoux, line 82: `following` isn't needed here as it doesn't provide additional information. I suggest removing it:

```suggestion
    concat(" for the paths: ")
```

src/api/app/components/workflow_artifacts_per_step_component.rb
```diff
@@ -0,0 +1,90 @@
...
+      artifacts_for_branch_or_link_package(context, parsed_artifacts, 'Branched package from ')
+    when 'Workflow::Step::LinkPackageStep'
+      artifacts_for_branch_or_link_package(context, parsed_artifacts, 'Linked package from ')
+    when 'Workflow::Step::RebuildPackage'
+      artifacts_for_rebuild_package(context, parsed_artifacts)
+    when 'Workflow::Step::ConfigureRepositories'
+      artifacts_for_configure_repositories(context, parsed_artifacts)
+    end
+  rescue JSON::ParserError
+    nil
```
dmarcoux, line 27: We should display at least an error, otherwise users won't know what happened.

saraycp: done

src/api/app/components/workflow_artifacts_per_step_component.rb
saraycp: done

src/api/app/components/workflow_artifacts_per_step_component.html.haml
saraycp: Good advice, thanks.
I also replaced the `context` that we previously passed from the view by `helpers`.

saraycp:
Specs broke after the changes. On it.

saraycp:
> Specs broke after the changes. On it.

fixed

danidoni (approved):

src/api/app/components/workflow_artifacts_per_step_component.rb
```diff
@@ -0,0 +1,89 @@
...
+    package_path = helpers.package_show_path(project: parsed_artifacts[:project], package: parsed_artifacts[:package])
+
+    tag.li do
+      concat('Rebuilt package ')
+      concat(link_to("#{parsed_artifacts[:project]}/#{parsed_artifacts[:package]}", package_path))
+      concat('.')
+    end
+  end
+
+  def artifacts_for_configure_repositories(parsed_artifacts)
```
dmarcoux, line 56: The links for a `configure_repositories` step do not point to the right project and repositories pages.

See this screenshot of a workflow run for a pull request event:
artifacts-ui

The project should be `home:bs-team:openSUSE:open-build-service:PR-12131` instead of `home:bs-team`. `Workflow::ArtifactsCollector` has to be changed for `configure_repositories` to rely on `@step.target_project_name`, so the same as the artifacts for `link_package` and `branch_package` steps:
https://github.com/openSUSE/open-build-service/blob/ca9aaad9b5cab038129f4ade4d5a15efb0df7ab6/src/api/app/services/workflows/artifacts_collector.rb#L17-L19

This can be changed in a follow-up PR as it doesn't change the code here.

dmarcoux: I've created a card in our backlog for this: https://trello.com/c/SWtE1NiH/2500-use-targetprojectname-for-the-artifacts-of-configurerepositories

saraycp: Lucky that you caught this. Fixed in https://github.com/openSUSE/open-build-service/pull/12166

PR #12159, opened 2022-02-03
[api] fix permission handling of projects with remote refernces

adrianschroeter:
Allowing modifing non project meta config again, like the build configuration


hennevogel (changes requested):
I find this a bit confusing. Currently this new rules only applies to `prjconf` and only in the API. But the `ProjectPolicy` is used everywhere throughout the app.

If there are different rules for project configuration, can we clarify them first (for instance what about the other stuff like _attribute or pubkey?) before we apply them? And then apply them everywhere.

adrianschroeter:
On Donnerstag, 3. Februar 2022, 14:05:40 CET Henne Vogelsang wrote:
> 
> I find this a bit confusing. Currently this new rules only applies to `prjconf` and only in the API. But the `ProjectPolicy` is used everywhere throughout the app.
> 
> If there are different rules for project configuration, can we clarify them first (for instance what about the other stuff like _attribute or pubkey?) before we apply them?

_attribute is a complete own story since it is per attribute.

the exception are the DoD and OBS remote definitions in meta, only these needs extra admin-only protections.

Everything else should still be editable by the maintainers.

This seems to be a regression by the introduction of the policy model (but I haven't verified that)

-- 

Adrian Schroeter ***@***.***>

PR #12167, opened 2022-02-04
[api] take care of devel package definition in owner search

adrianschroeter:
Lookup the owner definition in owner search, except this got
disabled in root project or via parameter.

### DIFF
--- a/src/api/app/controllers/search_controller.rb
+++ b/src/api/app/controllers/search_controller.rb
@@ -82,10 +82,17 @@ class SearchController < ApplicationController
         return projects unless params[:package]
 
         pkgs = []
+        search = OwnerSearch::Container.new(params)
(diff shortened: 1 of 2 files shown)


adrianschroeter:
the poor-mans version of that is applied on api.opensuse.org

adrianschroeter:
it should be done in "app/models/owner_search/container.rb" design wise, but the current model design does not allow to access the root project. so fixing it this way for now.

danidoni:
@adrianschroeter can you please fix the issues [raised by Rubocop](https://app.circleci.com/pipelines/github/openSUSE/open-build-service/5708/workflows/688fd9b3-4872-4bb5-af2e-ad635d96929b/jobs/77524)?

danidoni (approved):

PR #12177, opened 2022-02-08
allow kiwi versions not strictly x.y.z scheme

adrianschroeter:
Kiwi just allows numbers and dots. This fixes errors when working on our
official templates for Leap, using just "15.3" as version for example.


adrianschroeter:
should get also backported to 2.10 as we distribute these templates also for remote instances

PR #12185, opened 2022-02-10
Implement the new watchlist feature

danidoni:
This new watchlist feature allows the user to save Projects, Packages and BsRequests for later access.
We are introducing a polymorphic association to do so.

The new Watchlist implementation is going to be behind the `new_watchlist` feature flag.

The UI is going to remain as similar as possible to the existing one, for now:

Screenshot 2022-02-18 at 17-41-52 Request 2 (review)

For similarity with the old version of the Watchlist, the empty state will display the header of each list but empty. Compare the two versions:

**Old version**

Screenshot 2022-02-16 at 09-58-11 Open Build Service

**New version**

Screenshot 2022-02-16 at 09-59-05 Open Build Service

Link to the [review app](https://obs-reviewlab.opensuse.org/danidoni-new-watchlist-implementation)


danidoni:
[New Watchlist to track projects, packages and requests (L)](https://trello.com/c/DkeB6Rf9/1834-new-watchlist-to-track-projects-packages-and-requests-l)

src/api/app/components/watchlist_component.rb
```diff
@@ -0,0 +1,63 @@
+class WatchlistComponent < ApplicationComponent
+  def initialize(user:)
+    super
+
+    @user = user
+  end
+
+  def watched_item?
```
danidoni, line 8: Shouldn't be this plural?

src/api/app/models/project.rb
```diff
@@ -55,6 +55,8 @@ def autocomplete(search)
   has_many :linked_repositories, through: :path_elements, source: :link, foreign_key: :repository_id
   has_many :repository_architectures, -> { order('position') }, through: :repositories
 
+  has_many :watched_items, as: :watchable, dependent: :destroy
+  # FIXME: We will remove the following association when the one above goes out of beta
```
dmarcoux, line 59: To easily find this comment when we're removing the notifications_redesign feature toggle, I would include the feature toggle's name in the comment.

```suggestion
  # FIXME: We will remove the following association when the notifications_redesign goes out of beta
```

src/api/app/models/watched_item.rb
```diff
@@ -2,7 +2,8 @@ class WatchedItem < ApplicationRecord
   belongs_to :watchable, polymorphic: true
   belongs_to :user
 
-  validates :user_id, uniqueness: { scope: [:watchable_type, :watchable_id] }
+  validates :watchable_type, :watchable_id, :presence => true
```
saraycp, line 5: According to [this cop](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsredundantpresencevalidationonbelongsto), we don't need to validate the presence

src/api/app/components/watchlist_component.rb
saraycp: This method is asking if the item on the current page (`current_page_item`) is already on the watchlist.

danidoni: Then maybe It would be `item_watched?` instead?

src/api/app/models/project.rb
danidoni: But the feature toggle should be `new_watchlist` right?

saraycp: I added a comment mentioning the `new_watchlist` feature flag.

src/api/app/components/watchlist_component.rb
saraycp: I renamed the methods related to the item we have on current page. All of them are `current_page_item*`. So this one is renamed to  `current_page_item_in_watchlist?`.

src/api/app/models/project.rb
dmarcoux: > But the feature toggle should be `new_watchlist` right?

True! At the time of writing my previous comment, I didn't know we would have a new feature flag :slightly_smiling_face:.

src/api/app/components/watchlist_component.html.haml
```diff
@@ -0,0 +1,32 @@
+
```
dmarcoux, line 1: This extra line can be removed.

```suggestion
```

src/api/app/components/watchlist_component.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    @user = user
+  end
+
+  def current_page_item_in_watchlist?
+    @user.watched_items.includes(:watchable).where(watchable: current_page_item).any?
+  end
+
+  def current_page_item_watchable?
+    params[:package] || params[:project] || params[:number]
```
dmarcoux, line 13: Where is `params` coming from in the context of a view component?

src/api/app/components/watchlist_component.html.haml
saraycp: You're better than the linters :smile: thanks!

src/api/app/components/watchlist_component.rb
saraycp: `params` is simply accessible from the view component. But now that you mention this, maybe it's better to pass what we need to the view component initializer and don't rely on `params` inside the view component... Let me see if that is doable.

What do you think?

src/api/app/components/watchlist_component.html.haml
dmarcoux: I wonder why it's not caught by linters to be honest... loll

src/api/app/components/watchlist_component.rb
dmarcoux: > it's better to pass what we need to the view component initializer

Yes, I agree. We should avoid global state in view components, it makes them difficult to test. I'll add a note about this in our [developer page for view components](https://github.com/openSUSE/open-build-service/wiki/View-Components).

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+    context 'when two different users try to watch the same project' do
+      let(:watched_item) { create(:watched_item, :for_projects) }
+
+      before { create(:watched_item, :for_projects) }
+
+      it 'is valid' do
+        expect(watched_item).to be_valid
+      end
+    end
+  end
```
saraycp, line 61: I think specifying the same **project** in both factory creations is missing here :thinking:

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+
+    context 'when the user tries to add different projects to the watchlist' do
+      let(:watched_item) { build(:watched_item, :for_projects) }
+
+      before { create(:watched_item, :for_projects) }
+
+      it 'is valid' do
+        expect(watched_item).to be_valid
+      end
+    end
```
saraycp, line 50: I think specifying the same **user** in both factory creations is missing here :thinking:

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+
+    context 'when the user tries to add different packages to the watchlist' do
+      let(:watched_item) { build(:watched_item, :for_packages) }
+
+      before { create(:watched_item, :for_packages) }
+
+      it 'is valid' do
+        expect(watched_item).to be_valid
+      end
+    end
```
saraycp, line 109: Same. I think specifying the same **user** in both factory creations is missing here :thinking:

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+    context 'when two different users try to watch the same package' do
+      let(:watched_item) { create(:watched_item, :for_packages) }
+
+      before { create(:watched_item, :for_packages) }
+
+      it 'is valid' do
+        expect(watched_item).to be_valid
+      end
+    end
+  end
```
saraycp, line 120: Same. I think specifying the same **package** in both factory creations is missing here :thinking:

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      let(:watched_item) { build(:watched_item, :for_bs_requests, user: nil) }
+
+      it 'is not valid' do
+        expect(watched_item).not_to be_valid
+        expect(watched_item.errors.full_messages).to eql(["User must exist"])
+      end
+    end
+
+    context 'when the user tries to add the same request twice to the watchlist' do
+      let(:bs_request) { create(:bs_request_with_submit_action) }
```
saraycp, line 148: there is a `bs_request` created in the parent context.

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+    context 'when two different users try to watch the same package' do
+      let(:watched_item) { create(:watched_item, :for_bs_requests) }
+
+      before { create(:watched_item, :for_bs_requests) }
+
+      it 'is valid' do
+        expect(watched_item).to be_valid
+      end
+    end
+  end
```
saraycp, line 178: Same. I think specifying the same **bs_request** in both factory creations is missing here :thinking:

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      let(:watched_item) { build(:watched_item, :for_projects, user: nil) }
+
+      it 'is not valid' do
+        expect(watched_item).not_to be_valid
+        expect(watched_item.errors.full_messages).to eql(["User must exist"])
+      end
+    end
+
+    context 'when the user tries to add the same project twice to the watchlist' do
+      let(:project) { create(:project) }
```
saraycp, line 30: There is a project already created in the parent context. This can be removed.

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      let(:user) { create(:user) }
```
saraycp, line 31: There is a user created at the beginning of the spec. This can be removed.

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+
+      it 'is not valid' do
+        expect(watched_item).not_to be_valid
+        expect(watched_item.errors.full_messages).to eql(["User must exist"])
+      end
+    end
+
+    context 'when the user tries to add the same package twice to the watchlist' do
+      let(:package) { create(:package) }
+      let(:user) { create(:user) }
```
saraycp, line 90: There is a user already created at the beginning of the spec. This can be removed.

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+
+      it 'is valid' do
+        expect(watched_item).to be_valid
+      end
+    end
+  end
+
+  context 'when watching a request' do
+    let(:bs_request) { create(:bs_request_with_submit_action) }
+    let(:user) { create(:user) }
```
saraycp, line 124: There is a user already created at the beginning of the spec. This can be removed.

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      let(:watched_item) { build(:watched_item, :for_packages, user: nil) }
+
+      it 'is not valid' do
+        expect(watched_item).not_to be_valid
+        expect(watched_item.errors.full_messages).to eql(["User must exist"])
+      end
+    end
+
+    context 'when the user tries to add the same package twice to the watchlist' do
+      let(:package) { create(:package) }
```
saraycp, line 89: There is a package already created in the parent context. This can be removed.

src/api/app/components/watchlist_component.rb
danidoni: It's not available from the view component specs, so we need to change this...

src/api/spec/models/watched_item_spec.rb
danidoni: Yes it is :+1:

src/api/app/components/watchlist_component.rb
saraycp: @dmarcoux could you have another look? I added a new commit refactoring this part of the code.

src/api/app/components/watchlist_component.html.haml
```diff
@@ -0,0 +1,38 @@
...
+        %i.fas.fa-times
+
+    - if @object_to_be_watched
+      .nav-item.pb-2.mb-4.border-bottom.border-gray-500
+        .h6.text-word-break-all.text-light= title
+        = link_to(toggle_watchable_path, method: :put, class: 'nav-link') do
+          - if object_to_be_watched_in_watchlist?
+            %p.mb-0.text-light
+              %i.fas.fa-times-circle
+              %span= remove_text
```
danidoni, line 14:
```suggestion
              %span= remove_from_watchlist_text
```

src/api/app/components/watchlist_component.html.haml
```diff
@@ -0,0 +1,38 @@
...
+          - else
+            %p.mb-0.text-light
+              %i.fas.fa-plus-circle
+              %span= watch_text
```
danidoni, line 18:
```suggestion
              %span= add_to_watchlist_text
```

src/api/app/components/watchlist_component.rb
```diff
@@ -0,0 +1,85 @@
+class WatchlistComponent < ApplicationComponent
+  def initialize(user:, project_name: nil, package_name: nil, bs_request_number: nil)
+    super
+
+    @user = user
+    @project_name = project_name
+    @package_name = package_name
+    @bs_request_number = bs_request_number
```
danidoni, line 8: We don't really use those instances in the template, right? I would change it to local variables then.

src/api/db/schema.rb
```diff
@@ -1051,6 +1051,17 @@
     t.index ["login"], name: "users_login_index", unique: true, length: 255
   end
 
+  create_table "watched_items", id: :integer, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.string "watchable_type", null: false
+    t.bigint "watchable_id", null: false
```
dmarcoux, line 1056: This should be an integer, not a bigint.

src/api/db/schema.rb
```diff
@@ -1051,6 +1051,17 @@
...
+    t.bigint "user_id"
```
dmarcoux, line 1057: This should be an integer, not a bigint.

src/api/app/components/watchlist_component.rb
saraycp: We finally agreed on passing the parameters like this `WatchlistComponent.new(user: User.session!, project_name: params[:project], package_name: params[:package], bs_request_number: params[:number])` so we don't use `params` inside the view component.

src/api/app/components/watchlist_component.html.haml
saraycp: done

src/api/app/components/watchlist_component.html.haml
saraycp: done

src/api/app/assets/stylesheets/webui/new_watchlist/watchlist.scss
```diff
@@ -0,0 +1,4 @@
+.color-inverted-svg {
```
dmarcoux, line 1: :nitpick:

Is this really only for SVGs? This works for any image. I suggest renaming the class to:

```suggestion
.invert-color {
```

src/api/app/components/watched_items_list_component.rb
```diff
@@ -0,0 +1,41 @@
+class WatchedItemsListComponent < ApplicationComponent
+  LIST_TITLE = {
+    'Package' => 'Packages you are watching',
+    'Project' => 'Projects you are watching',
+    'BsRequest' => 'Requests you are watching'
+  }.freeze
+
+  EMPTY_LIST_TEXTS = {
+    'Package' => 'There are no packages in the list yet.',
```
dmarcoux, line 9:
```suggestion
    'Package' => 'There are no packages in the watchlist yet.',
```

src/api/app/components/watched_items_list_component.rb
```diff
@@ -0,0 +1,41 @@
...
+    'Project' => 'There are no projects in the list yet.',
```
dmarcoux, line 10:
```suggestion
    'Project' => 'There are no projects in the watchlist yet.',
```

src/api/app/components/watched_items_list_component.rb
```diff
@@ -0,0 +1,41 @@
...
+    'BsRequest' => 'There are no requests in the list yet.'
```
dmarcoux, line 11:
```suggestion
    'BsRequest' => 'There are no requests in the watchlist yet.'
```

dmarcoux:
There is a small typo twice in the first commit (once in the message and once in the description): 

Wat**c**hedItem (**c** is missing)

dmarcoux:
Before approving, is this a conscious decision to have a clear cut between the watched projects with and without the `new_watchlist` feature toggle enabled?

With the current implementation, a user watches a project without `new_watchlist` and it's not watched after joining the beta (thus enabling `new_watchlist`). Same if a user leaves the beta, all watched projects aren't taken over. I don't think it's a big issue. It doesn't have to be implemented.

src/api/app/components/package_icon_and_link_component.html.haml
```diff
@@ -0,0 +1,3 @@
+= link_to package_show_path(@package.project, @package), class: 'text-word-break-all' do
```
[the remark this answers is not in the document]
hennevogel: Screenshot from 2022-02-17 14-43-17

hennevogel (changes requested):

danidoni:
> 

Yes, it was a conscious decision.

src/api/app/components/package_icon_and_link_component.html.haml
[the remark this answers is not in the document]
dmarcoux: If there isn't any logic in this view component to be tested, then I agree with you @hennevogel. It is the case for those *IconAndLinkComponent.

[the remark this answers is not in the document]
danidoni: We created this to remove a big `if/else` in the parent component code. Let me see if we can remove those 3 components and not revert to that `if/else`...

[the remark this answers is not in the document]
hennevogel: please don't ignore this part: "I think don't think this is needed in the UI"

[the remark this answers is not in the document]
saraycp: We are mixing two things here:

a) The header/title (the screenshot attached by @hennevogel before).
b) Each item of the list, which is currently generated by `*IconAndLinkComponent`.

Please, look at the red texts and arrows in this screenshot:

watchlist

Having that clear...
I agree we can remove a) if we agree that it's enough to have the package name in the breadcrumb and URL. My screenshot is an example of a page where the package name only appears in those places.

Regarding b), we moved the code to view components for two main reasons: removing a big `if/else`, as Dani said, and also because we thought they can be reused in other parts of the application (the breadcrumbs for example, at least the project part for now).

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  def toggle
+    watched_item = User.session!.watched_items.find_by(watchable: @item)
+
+    if watched_item
+      watched_item.destroy
+    else
+      User.session!.watched_items.create(watchable: @item)
+    end
+
+    redirect_back(fallback_location: root_path)
```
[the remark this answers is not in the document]
krauselukas: besides what already raised, nothing to add :+1: great stuff @danidoni @saraycp

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+            elsif params[:number]
+              BsRequest.find_by(number: params[:number])
+            end
+  end
+
+  def check_user_belongs_feature_flag
+    raise NotFoundError unless Flipper.enabled?(:new_watchlist, User.session)
+  end
+
+  def flash_subject_per_watchable(item)
```
krauselukas, line 41: This could go after applying the suggested changes

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def toggle
+    watched_item = User.session!.watched_items.find_by(watchable: @item)
+
+    if watched_item
+      watched_item.destroy
+      flash[:success] = "Removed #{flash_subject_per_watchable(@item)} from the watchlist"
+    else
+      User.session!.watched_items.create(watchable: @item)
+      flash[:success] = "Added #{flash_subject_per_watchable(@item)} to the watchlist"
```
krauselukas, line 19: Same here

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    Project => 'project',
+    BsRequest => 'request'
+  }.freeze
+
+  def toggle
+    watched_item = User.session!.watched_items.find_by(watchable: @item)
+
+    if watched_item
+      watched_item.destroy
+      flash[:success] = "Removed #{flash_subject_per_watchable(@item)} from the watchlist"
```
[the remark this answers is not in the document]
danidoni: True, fixing!

src/api/app/views/layouts/webui/webui.html.haml
```diff
@@ -76,4 +76,10 @@
 
       -# Collapsible menu shared between top and bottom navigation
       - if User.session
-        = render partial: 'layouts/webui/watchlist'
+        - if feature_enabled?(:new_watchlist)
+          = render WatchlistComponent.new(user: User.session!,
+          project_name: params[:project],
+          package_name: params[:package],
+          bs_request_number: params[:number])
```
hennevogel, line 83: Sounds to me you want to use `@project`, `@package` etc. instead of passing around params and instantiating the objects again in your component.

src/api/config/routes/webui_routes.rb
```diff
@@ -340,6 +340,10 @@
         resources :workflow_runs, only: [:index, :show], controller: 'webui/workflow_runs'
       end
       resources :token_triggers, only: [:show, :update], controller: 'webui/users/token_triggers'
+
+      put 'watched_items/package/:project/:package/toggle' => :toggle, controller: 'webui/watched_items', as: :toggle_package_watchable
```
hennevogel, line 344: Why aren't you using resourceful routes?

hennevogel (changes requested):

src/api/config/routes/webui_routes.rb
eduardoj: Here I don't talk about using resourceful routes, but using RESTful endpoints.

Regarding the `/toggle` part, [this](https://stackoverflow.com/questions/3266292/best-way-to-implement-a-restful-toggle-action/5030057#5030057) is an interesting answer from Stack Overflow, about how much a toggle endpoint is RESTful.

Using a toggle method in our endpoints is something we have introduced not so long ago. See #11250 and #11970.

It would be nice to make those endpoints and this last one RESTful.

src/api/app/views/layouts/webui/webui.html.haml
saraycp: Good idea, we adapted the code for this.

src/api/config/routes/webui_routes.rb
saraycp: We could partially convert this to resourceful routes. Let us know what you think.

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  before_action :set_item
+
+  FLASH_PER_WATCHABLE_TYPE = {
+    Package => 'package',
+    Project => 'project',
+    BsRequest => 'request'
+  }.freeze
+
+  def toggle
+    watched_item = User.session!.watched_items.find_by(watchable: @item)
```
hennevogel, line 12: If you require people to be logged in, use `before_action :require_login`

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    redirect_back(fallback_location: root_path)
+  end
+
+  private
+
+  def set_item
+    @item = if params[:package]
+              Package.find_by_project_and_name(params[:project], params[:package])
+            elsif params[:project]
+              Project.find_by(name: params[:project])
```
hennevogel, line 31: Did you consider [links](https://github.com/openSUSE/open-build-service/wiki/Links)?

src/api/config/routes/webui_routes.rb
```diff
@@ -340,6 +340,12 @@
         resources :workflow_runs, only: [:index, :show], controller: 'webui/workflow_runs'
       end
       resources :token_triggers, only: [:show, :update], controller: 'webui/users/token_triggers'
+
+      resource :watched_items, controller: 'webui/watched_items', only: [:toggle], constraints: cons do
+        put '/package/:project/:package/toggle' => :toggle, as: :toggle_package
```
hennevogel, line 345: This isn't [resourceful routing](https://guides.rubyonrails.org/routing.html#resource-routing-the-rails-default)...

src/api/spec/controllers/webui/watched_items_controller_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+        it 'redirects back' do
+          expect(response).to redirect_to(root_path)
+        end
+
+        it 'adds the watched item to the watchlist' do
+          expect(user.watched_items.map(&:watchable)).to include(package)
+        end
+
+        it 'shows the flash with the success message' do
```
hennevogel, line 27: Why test the flash? You expect `ActionDispatch::Flash` to break? :)

src/api/spec/controllers/webui/watched_items_controller_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      context 'when the watched item is not already watched' do
+        let(:user) { create(:confirmed_user) }
+        let(:package) { create(:package) }
+
+        before do
+          login user
+          put :toggle, params: { package: package, project: package.project }
+        end
+
+        it 'redirects back' do
```
hennevogel, line 19: Why test the redirect? You expect `ActionController::Redirecting ` to break? :)

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+
+  context 'when watching a project' do
+    let(:project) { create(:project) }
+    let(:watched_item) { create(:watched_item, :for_projects) }
+
+    it 'is valid with valid attributes' do
+      expect(watched_item).to be_valid
+    end
+
+    it 'is not valid when missing a project' do
```
hennevogel, line 14: Why test this? You expect `ActiveRecord::Associations` to break?

hennevogel (changes requested):

src/api/app/controllers/webui/watched_items_controller.rb
danidoni: No, we didn't. We'll open a follow-up PR to address this.

src/api/spec/controllers/webui/watched_items_controller_spec.rb
danidoni: Okay, we'll remove this kind of tests, they don't add much value...

src/api/app/controllers/webui/watched_items_controller.rb
dmarcoux: Using Pundit would be better. If you decide to go with `before_action :require_login`, please add this controller to the list of controllers to be refactored in #10083.

src/api/spec/models/watched_item_spec.rb
```diff
@@ -0,0 +1,176 @@
+require 'rails_helper'
+
+RSpec.describe WatchedItem do
```
dmarcoux, line 3: With [shoulda-matchers](https://github.com/thoughtbot/shoulda-matchers#usage) (we have this gem in our Gemfile), we can replace a lot of these specs with a single spec using the matcher `validate_uniqueness_of`.

src/api/spec/controllers/webui/watched_items_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        before { login user }
+
+        subject { put :toggle, params: { package: package, project: package.project } }
+
+        it 'raises an exception' do
+          expect { subject }.to raise_error(NotFoundError)
+        end
+      end
+
+      context 'when the watched item is already watched' do
```
hennevogel, line 59: This is the same test as above. Both test `User.session!`

hennevogel (changes requested):

src/api/app/controllers/webui/watched_items_controller.rb
saraycp: Thanks for the warning. Added to the list in #10083.

src/api/spec/controllers/webui/watched_items_controller_spec.rb
saraycp: done

src/api/spec/models/watched_item_spec.rb
saraycp: done

src/api/spec/controllers/webui/watched_items_controller_spec.rb
saraycp: done

src/api/app/controllers/webui/watched_items_controller.rb
hennevogel: I'm a bit worried here as this will have influence on notifications one day-

src/api/spec/controllers/webui/watched_items_controller_spec.rb
saraycp: done

src/api/app/controllers/webui/watched_items_controller.rb
saraycp: We added a card in our backlog for this: https://trello.com/c/9wC6V7qQ/2520-handle-links-in-watchlist

src/api/config/routes/webui_routes.rb
saraycp: We have a clear idea of how to do this but don't want to block this PR. We'll add the changes in a follow-up PR.

hennevogel (approved):

PR #12203, opened 2022-02-14
Use dom_id for mark_as_read_or_unread_button

dmarcoux:
This is what ActionView uses to associate records with DOM elements. Let's use it too!

https://api.rubyonrails.org/classes/ActionView/RecordIdentifier.html#method-i-dom_id

The generated ID is slightly different, `dom_id` uses underscores instead of dashes. So before, it was `update-notification-123`, now it's `update_notification_123`. It's not an issue, as we don't refer to this ID in any code.

Nothing changes visually, but here's the helper in use with the new ID:
preview


dmarcoux:
I was going to introduce this in #12018, but this change isn't related to the view components created in that PR. Instead, I decided to submit this small PR.

eduardoj (approved):

PR #12210, opened 2022-02-15
Create RuboCop cop to avoid global state in view components

dmarcoux:
To enforce what we've defined in our developer wiki: https://github.com/openSUSE/open-build-service/wiki/View-Components#avoid-global-state

Two nice articles to understand how custom cops for RuboCop work:
- https://thoughtbot.com/blog/rubocop-custom-cops-for-custom-needs
- https://evilmartians.com/chronicles/custom-cops-for-rubocop-an-emergency-service-for-your-codebase

### DIFF
--- a/src/api/spec/lib/rubocop/cop/view_component/avoid_global_state_spec.rb
+++ b/src/api/spec/lib/rubocop/cop/view_component/avoid_global_state_spec.rb
@@ -1,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe RuboCop::Cop::ViewComponent::AvoidGlobalState, :config do
+  context 'when a view component uses params' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        class MyComponent < ApplicationComponent
+          def initialize
+            @abc = params[:abc]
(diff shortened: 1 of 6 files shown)


danidoni (approved):

src/api/spec/lib/rubocop/cop/view_component/avoid_global_state_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+require './lib/rubocop/cop/view_component/avoid_global_state' # TODO: Why is Zeitwerk not auto-loading this?
```
dmarcoux, line 2: I'll address this by moving the custom cop under `src/api/app/lib/` instead of `src/api/lib/`.

dmarcoux:
I've fixed the auto-loading of the custom cop by moving its file under `src/api/app/lib` and changing the Zeitwerk inflector. @danidoni, please have another look.

krauselukas (approved):
Not an expert with custom rubocops, but what I see looks good to me :)

saraycp (approved):
Works fine locally! :+1:

PR #12226, opened 2022-02-17
Instrument Workflow and Workflow::Step

dmarcoux:
With this, I want to know how many workflows are instantiated, then executed. This will differ a lot if users rely on filters or not. It will be interesting to know. It's basic, but it's a first step.

I also want to know which steps are the most popular in workflows. It allows us to see how the adoption is going for newly introduced steps.


src/api/app/models/workflow.rb
```diff
@@ -132,4 +136,12 @@ def restore_target_projects
       processable_steps.first.create_or_update_subscriptions(target_package, filters)
     end
   end
+
+  def track_instantiation
```
hennevogel, line 140: Wouldn't that be

Screenshot from 2022-02-17 18-06-54


https://obs-measure.opensuse.org/d/GRaKLnm7k/scm-ci-integration?orgId=1&viewPanel=2

hennevogel (changes requested):

src/api/app/instrumentations/workflow_instrumentation.rb
```diff
@@ -0,0 +1,21 @@
+module WorkflowInstrumentation
```
dmarcoux, line 1: What do you think of this approach to extract all instrumentation code to a separate file? Each model has its own instrumentation. The only change needed to use this instrumentation, beside including this module in the `Workflow` class, is to add `run_callbacks` to the methods for which we defined callbacks.

src/api/app/instrumentations/workflow_step_instrumentation.rb
```diff
@@ -0,0 +1,16 @@
...
+  included do
+    define_model_callbacks :initialize
+
+    after_initialize :track_instantiation
+  end
+
+  private
+
+  def track_instantiation
+    RabbitmqBus.send_to_bus('metrics', "workflow_step.#{self.class.name.demodulize.underscore}.instantiation value=1")
```
dmarcoux, line 14: With this data, I would like to have a panel showing how much each step is instantiated. I wonder if this panel would always need to be updated whenever we introduce a new workflow step. I would like to avoid this. If I change this line to the following, I could group by value, right?

```suggestion
    RabbitmqBus.send_to_bus('metrics', "workflow_step.instantiation value=#{self.class.name.demodulize.underscore}")
```

src/api/spec/models/workflow/step_spec.rb
```diff
@@ -62,6 +62,10 @@
   describe '#target_project_name' do
     let(:step) do
       Class.new(described_class) do
+        def self.name
```
dmarcoux, line 65: This is needed to make tests pass now that we have instrumentation for `Workflow::Step`. Without this, `self.class.name` in the instrumentation code returns `nil` since we're defining an anonymous class here. It's only an issue in this spec.

src/api/app/instrumentations/workflow_step_instrumentation.rb
```diff
@@ -0,0 +1,16 @@
...
+  included do
+    define_model_callbacks :initialize
+
+    after_initialize :track_instantiation
+  end
+
+  private
+
+  def track_instantiation
+    RabbitmqBus.send_to_bus('metrics', "workflow_step,step=#{self.class.name.demodulize.underscore},action=instantiation count=1")
```
dmarcoux, line 14: With the workflow step class name being a tag, I will be able to group by `tag(step)`. I will create a Grafana panel for the instantiation of workflow steps without having to update this panel whenever a new step is introduced.

hennevogel (changes requested):

src/api/app/models/workflow.rb
```diff
@@ -1,5 +1,8 @@
+# See TODOs in this class to make this class smaller
+# rubocop:disable Metrics/ClassLength
```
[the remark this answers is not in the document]
dmarcoux: I agree, this is better :+1:

.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
```
dmarcoux, line 1: RuboCop TODOs were regenerated since refactoring the Workflow model to make the class smaller is out of scope for this PR.

hennevogel (approved):

PR #12235, opened 2022-02-21
Set password to opensuse for users generated by dev rake task

dmarcoux:
Just like we do for the Admin user. This is convenient when testing things in the development environment.

### DIFF
--- a/src/api/spec/README.md
+++ b/src/api/spec/README.md
@@ -100,7 +100,7 @@ You can use an inherited factory to add or override attributes.
     email { Faker::Internet.email }
     realname { Faker::Name.name }
     sequence(:login) { |n| "user_#{n}" }
-    password 'buildservice'
+    password 'opensuse'
 
     factory :confirmed_user do
       state 2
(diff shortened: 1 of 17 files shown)


src/api/lib/tasks/dev.rake
```diff
@@ -398,6 +398,9 @@ namespace :dev do
       # Trigger package builds for home:admin
       home_admin.store
 
+      # The password of every user will be 'opensuse', so we can easily log in as one of them if needed
+      User.all_without_nobody.each { |user| user.update(password: 'opensuse', password_confirmation: 'opensuse') }
+
```
eduardoj, line 403: This is controlled in the [user factory](https://github.com/openSUSE/open-build-service/blob/ee129e3e08909ace74beaeed31e8d144840ffccb/src/api/spec/factories/users.rb#L6). Maybe you want to change that line. Is there any other use case for modifying the passwords with the dev rake task?

dmarcoux: You have a good point, this would be simpler.

> Is there any other use case for modifying the passwords with the dev rake task?

No, I didn't think about changing the factory directly.

dmarcoux:
I would need to spend a bit more time on this since a lot of specs are breaking when changing the user factories. I prefer to focus on other things, I'm closing this PR.

PR #12236, opened 2022-02-21
resourceful routes for watched items

saraycp:
Right now we use custom routes for creating/deleting watched items. Following Rails conventions for routing, we should update the custom routes for watched items toggle to resourceful routes.

We can't use `member` for some of these new routes because we have `param: :name` for both projects and package nested routes and it's not possible to distinguish if the name is for `project` or `package`. 

The [last paragraph of this article](https://guides.rubyonrails.org/v6.1/routing.html#adding-member-routes) describes that we can ommit `member` to obtain the same result but using the param names we need.

### DIFF
--- a/src/api/app/controllers/webui/watched_items_controller.rb
+++ b/src/api/app/controllers/webui/watched_items_controller.rb
@@ -9,7 +9,7 @@ class Webui::WatchedItemsController < Webui::WebuiController
     BsRequest => 'request'
   }.freeze
 
-  def toggle
+  def toggle_watched_item
     watched_item = User.session!.watched_items.find_by(watchable: @item)
 
     if watched_item
@@ -26,10 +26,10 @@ class Webui::WatchedItemsController < Webui::WebuiController
   private
 
   def set_item
-    @item = if params[:package]
-              Package.find_by_project_and_name(params[:project], params[:package])
-            elsif params[:project]
-              Project.find_by(name: params[:project])
+    @item = if params[:project_name] && params[:package_name]
+              Package.find_by_project_and_name(params[:project_name], params[:package_name])
+            elsif params[:project_name]
+              Project.find_by(name: params[:project_name])
             elsif params[:number]
               BsRequest.find_by(number: params[:number])
             end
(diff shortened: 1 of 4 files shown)


danidoni (approved):

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -26,12 +26,12 @@ def toggle
...
-              Package.find_by_project_and_name(params[:project], params[:package])
-            elsif params[:project]
-              Project.find_by(name: params[:project])
-            elsif params[:number]
-              BsRequest.find_by(number: params[:number])
+    @item = if params[:project_name] && params[:package_name]
+              Package.find_by_project_and_name(params[:project_name], params[:package_name])
+            elsif params[:project_name]
+              Project.find_by(name: params[:project_name])
+            elsif params[:request_number]
```
dmarcoux, line 33: Watching a request doesn't work as this should be `params[:number]`:

```suggestion
            elsif params[:number]
```

This should have been caught by a spec. Nothing is checking if this controller works for requests.

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -26,12 +26,12 @@ def toggle
...
+              BsRequest.find_by(number: params[:request_number])
```
dmarcoux, line 34: Same here:

```suggestion
              BsRequest.find_by(number: params[:number])
```

dmarcoux (changes requested):

src/api/app/controllers/webui/watched_items_controller.rb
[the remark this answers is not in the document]
saraycp: Using `member` in this case, the route is  `requests/:number/toggle_watched_item` not `requests/:request_number/toggle_watched_item`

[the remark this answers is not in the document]
rubhanazeem: Thanks for pointing this out. It's updated. :slightly_smiling_face:

src/api/app/controllers/webui/watched_items_controller.rb
rubhanazeem: It's updated and I've added some new specs for this. :slightly_smiling_face:

saraycp (commented):
LGTM now, but I can't approve :)

dmarcoux (approved):

PR #12243, opened 2022-02-22
Improve queries in WatchlistComponent

dmarcoux:
Two changes which are explained in two separate commits. Please have a look at the commit message and their description.


dmarcoux:
We should avoid doing queries in view components, but refactoring this is out of scope for this PR.

danidoni (approved):

PR #12248, opened 2022-02-23
Avoid database queries in view components

dmarcoux:
This will help us in having better view components which only handle view code, not fetching data from the database. It also enforces what we wrote in our [developer documentation on view components](https://github.com/openSUSE/open-build-service/wiki/View-Components#avoid-database-queries).

### DIFF
--- a/src/api/config/initializers/db_query_matchers.rb
+++ b/src/api/config/initializers/db_query_matchers.rb
@@ -1,0 +1,15 @@
+DBQueryMatchers.configure do |config|
+  config.ignores = [
+    # Executed multiple times when fetching something in the configuration
+    /^SELECT `configurations`.* FROM `configurations`/,
+    # Executed by the schema cache
+    /^SHOW FULL FIELDS FROM/,
+    # TODO: Explain this
+    /^SELECT column_name\nFROM information_schema.statistics/,
+    # Executed when there is a transaction
+    /^SAVEPOINT active_record/,
+    # Executed when there is a transaction
+    /^RELEASE SAVEPOINT active_record/
+  ]
+  config.ignore_cached = true
+end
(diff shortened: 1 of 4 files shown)


danidoni:
Cool!

danidoni:
Will this run against every single view component spec?

dmarcoux:
> Will this run against every single view component spec?

Yes, since we're overwriting `render_inline` in _src/api/spec/support/view_component.rb_ to check for the number of executed database queries.

dmarcoux:
Rebased on master to follow the Ruby update `3.1`

src/api/config/initializers/db_query_matchers.rb
```diff
@@ -0,0 +1,15 @@
+DBQueryMatchers.configure do |config|
+  config.ignores = [
+    # Executed multiple times when fetching something in the configuration
```
coolo, line 3: BTW: can we stop that? I have a branch somewhere to cache this as it's just very annoying (in the debug logs)

dmarcoux: I don't know why we're storing configuration in the database. Rails can be configured with custom configuration as explained in [Configuring Rails Applications](https://guides.rubyonrails.org/configuring.html#custom-configuration). This is what I would do to refactor this.

coolo: Because there is a UI to change it

dmarcoux:
There is now a way to do this with Rails, it was merged recently: https://github.com/rails/rails/pull/50281

coolo:
> There is now a way to do this with Rails, it was merged recently: [rails/rails#50281](https://github.com/rails/rails/pull/50281)
giphy

PR #12253, opened 2022-02-24
Add new watchable items in the subscription page

danidoni:
This PR adds the functionality to subscribe to events triggered for watched packages and request, recently added under the beta program.

*HOW TO TEST:*

1. Navigate to the subscription page http://localhost:3000/my/subscriptions
2. Check some of the new added event subscription like `Watching the request` under the `Request state was changed` section
3. Create a request and add it to your watchlist (make sure your user is part of the beta program)
4. Change the state of the request
5. Enter the rails console and run `SendEventEmailsJob.perform_now` to trigger the creation of the notifications
6. Check in the webui if the notification appears

*TODO:*

- [x] Fix minitests
- [x] Add specs

### DIFF
--- a/src/api/app/models/event/base.rb
+++ b/src/api/app/models/event/base.rb
@@ -251,7 +251,23 @@ module Event
       project = ::Project.find_by_name(payload['project'])
       return [] if project.blank?
 
-      project.watched_projects.map(&:user)
+      project.watched_projects.map(&:user).concat(project.watched_items.map(&:user)).uniq
+    end
+
+    def package_watchers
+      package = Package.get_by_project_and_name(payload['project'], payload['package'], { follow_multibuild: true, follow_project_links: false, use_source: false })
+      return [] if package.blank?
+
+      package.watched_items.map(&:user)
+    rescue Package::Errors::UnknownObjectError, Project::Errors::UnknownObjectError
+      []
+    end
+
+    def request_watchers
+      bs_request = BsRequest.find_by(number: payload['number'])
+      return [] if bs_request.blank?
+
+      bs_request.watched_items.map(&:user)
     end
 
     def _roles(role, project, package = nil)
--- a/src/api/app/models/event/request.rb
+++ b/src/api/app/models/event/request.rb
@@ -77,15 +77,48 @@ module Event
       action_maintainers('sourceproject', 'sourcepackage')
     end
 
-    def target_watchers
-      find_watchers('targetproject')
+    def source_watchers
+      source_or_target_project_watchers(project_type: 'sourceproject')
     end
 
-    def source_watchers
-      find_watchers('sourceproject')
+    def target_watchers
+      source_or_target_project_watchers(project_type: 'targetproject')
+    end
+
+    def source_package_watchers
+      source_or_target_package_watchers(project_type: 'sourceproject', package_type: 'sourcepackage')
+    end
+
+    def target_package_watchers
+      source_or_target_package_watchers(project_type: 'targetproject', package_type: 'targetpackage')
     end
 
     private
+
+    def source_or_target_project_watchers(project_type:)
+      # TODO: remove old `find_watchers` implementation after rolling out `new_watchlist` from beta program
+      watchers = find_watchers(project_type) +
+                 payload['actions'].map { |action| action[project_type] }
+                                   .map { |project_name| Project.find_by_name(project_name) }
+                                   .compact.map(&:watched_items)
+                                   .flatten.map(&:user)
+      watchers.uniq
+    end
+
+    def source_or_target_package_watchers(project_type:, package_type:)
+      payload['actions'].map { |action| [action[project_type], action[package_type]] }
+                        .map do |project_name, package_name|
+        next if project_name.blank? || package_name.blank?
+
+        Package.get_by_project_and_name(project_name,
+                                        package_name,
+                                        { follow_multibuild: true, follow_project_links: false, use_source: false })
+      rescue Package::Errors::UnknownObjectError, Project::Errors::UnknownObjectError
+        nil
+      end
+                        .compact.map(&:watched_items)
+                        .flatten.map(&:user)
+    end
 
     def action_maintainers(prjname, pkgname)
       payload['actions'].map do |action|
--- a/src/api/app/models/event/build_fail.rb
+++ b/src/api/app/models/event/build_fail.rb
@@ -4,7 +4,7 @@ module Event
 
     self.message_bus_routing_key = 'package.build_fail'
     self.description = 'Package has failed to build'
-    receiver_roles :maintainer, :bugowner, :reader, :watcher
+    receiver_roles :maintainer, :bugowner, :reader, :watcher, :package_watcher, :request_watcher
 
     create_jobs :report_to_scm_job
 
--- a/src/api/app/models/event/comment_for_package.rb
+++ b/src/api/app/models/event/comment_for_package.rb
@@ -3,7 +3,7 @@ module Event
     include CommentEvent
     self.message_bus_routing_key = 'package.comment'
     self.description = 'New comment for package created'
-    receiver_roles :maintainer, :bugowner, :watcher
+    receiver_roles :maintainer, :bugowner, :watcher, :package_watcher
     payload_keys :project, :package, :sender
 
     def subject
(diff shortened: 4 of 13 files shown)


krauselukas:
[Get notifications about packages and requests I'm watching (L)](https://trello.com/c/XmvYa9qx/1840-get-notifications-about-packages-and-requests-im-watching-l)

src/api/app/models/event/base.rb
```diff
@@ -254,6 +254,20 @@ def watchers
       project.watched_projects.map(&:user)
     end
 
+    def package_watchers
+      package = Package.find_by_name(payload['package'])
```
hennevogel, line 258: you will not find a package that way...

danidoni: True

src/api/app/models/event/base.rb
```diff
@@ -254,6 +254,20 @@ def watchers
       project.watched_projects.map(&:user)
     end
 
+    def package_watchers
+      package = Package.find_by_project_and_name(payload['project'], payload['package'])
```
hennevogel, line 258: Links....

src/api/app/models/event/request.rb
```diff
@@ -77,12 +77,36 @@ def source_maintainers
...
+      find_watchers('targetproject') +
+        payload['actions'].map { |action| action['targetproject'] }
+                          .map { |target_project_name| Project.find_by_name(target_project_name) }
+                          .compact.map(&:watched_items)
+                          .flatten.map(&:user)
+    end
+
+    def source_package_watchers
+      payload['actions'].map { |action| action['sourcepackage'] }
+                        .map { |source_package_name| Package.find_by_name(source_package_name) }
```
dmarcoux, line 100: This won't find a specific package with the provided package name. It needs to also be scoped to the project.

src/api/app/models/event/request.rb
```diff
@@ -77,12 +77,36 @@ def source_maintainers
...
+                        .compact.map(&:watched_items)
+                        .flatten.map(&:user)
+    end
+
+    def target_package_watchers
+      payload['actions'].map { |action| action['targetpackage'] }
+                        .map { |target_package_name| Package.find_by_name(target_package_name) }
```
dmarcoux, line 107: This won't find a specific package with the provided package name. It needs to also be scoped to the project.

saraycp:
Even if the user is out of beta, they see the new types of subscriptions on the subscriptions page. For example "Watching the package". Is it like that on purpose?

rubhanazeem:
> Even if the user is out of beta, they see the new types of subscriptions on the subscriptions page. For example "Watching the package". Is it like that on purpose?

I'm not able to get `Watching the package` if the user is out of beta. If I join beta I see this new subscription.

krauselukas:
> Even if the user is out of beta, they see the new types of subscriptions on the subscriptions page. For example "Watching the package". Is it like that on purpose?

@saraycp maybe because you are the Admin user? We always show it to admins, since he is able to set the "global" subscriptions for events

saraycp:
> Even if the user is out of beta, they see the new types of subscriptions on the subscriptions page. For example "Watching the package". Is it like that on purpose?

Sorry, the reason is that I was logged in with Admin user, and admins can always see all kinds of subscriptions, according to [these lines](https://github.com/openSUSE/open-build-service/pull/12253/commits/78becba114b585cbb844b2f33e93d8d05e346c20#diff-78824bf6bc903270b49ec6197d9129541c98607ebc17c79c8d3d4b1c72db84cbR59-R65).

Thanks, @rubhanazeem for pointing me to this.

src/api/app/models/event/base.rb
```diff
@@ -254,6 +254,22 @@ def watchers
       project.watched_projects.map(&:user)
     end
 
+    def package_watchers
+      package = Package.get_by_project_and_name(payload['project'], payload['package'], { follow_multibuild: true, follow_project_links: false, use_source: false })
+      return [] if package.blank?
```
hennevogel, line 259: What about the possible exceptions?

src/api/app/models/event/base.rb
```diff
@@ -254,6 +254,22 @@ def watchers
...
+
+      package.watched_items.map(&:user)
+    rescue Package::Errors::UnknownObjectError
```
eduardoj, line 262:
```suggestion
    rescue Package::Errors::UnknownObjectError, Project::Errors::UnknownObjectError
```

src/api/app/models/event/request.rb
```diff
@@ -77,16 +77,48 @@ def source_maintainers
...
 
     private
 
+    def source_or_target_project_watchers(project_type:)
+      # TODO: remove old `find_watchers` implementation after rolling out `new_watchlist` from beta program
+      find_watchers(project_type) +
+        payload['actions'].map { |action| action[project_type] }
+                          .map { |project_name| Project.find_by_name(project_name) }
+                          .compact.map(&:watched_items)
+                          .flatten.map(&:user)
```
eduardoj, line 104: Shouldn't we add here a `rescue Project::Errors::UnknownObjectError`?

src/api/app/models/event/request.rb
```diff
@@ -77,16 +77,48 @@ def source_maintainers
...
+
+        Package.get_by_project_and_name(project_name,
+                                        package_name,
+                                        { follow_multibuild: true, follow_project_links: false, use_source: false })
+      end
+                        .compact.map(&:watched_items)
+                        .flatten.map(&:user)
+    rescue Package::Errors::UnknownObjectError, Project::Errors::UnknownObjectError
+      []
+    end
```
eduardoj, line 120: Shouldn't the `rescue` clause be moved just after the call to `get_by_project_and_name`? I'm thinking in a case (not yet confirmed) where several actions could contain different project/package combinations. The code, as it is now, will return `[]` just after the first ocurrence of project/package not found.

src/api/app/models/event/request.rb
danidoni: Not this one, the `Project.find_by_name` will return `nil` instead of :boom:, so we don't need to catch anything. Then `nil` will be discarded by the `compact` call.

src/api/app/models/event/request.rb
danidoni: Fixed, good catch!

src/api/app/models/event/request.rb
eduardoj: Ops, true. I got confused with `Project.get_by_name`.

By the way, `Project.find_by_name` [will be obsoleted](https://github.com/openSUSE/open-build-service/blob/4842f9f1c85a1b5b9050d2f0025fd4b215846e10/src/api/app/models/project.rb#L259-L261). We can live with it. Having around 40 ocurrences of Project.find_by_name in the code it is a signal that we still should use this method, and that at least one PR only refactoring these calls would be needed.

src/api/app/models/event/base.rb
```diff
@@ -254,6 +254,22 @@ def watchers
       project.watched_projects.map(&:user)
```
saraycp, line 254: This needs to be adapted to get `watched_items` instead of `watched_projects` when the feature flag is enabled.

saraycp (changes requested):

src/api/app/models/event/base.rb
krauselukas: @saraycp its fixed, good catch!

krauselukas: I will add a spec to validate its working

krauselukas: @saraycp okay added the test case :)

src/api/app/models/event/base.rb
```diff
@@ -251,7 +251,23 @@ def watchers
       project = ::Project.find_by_name(payload['project'])
       return [] if project.blank?
 
-      project.watched_projects.map(&:user)
+      project.watched_projects.map(&:user).concat(project.watched_items.map(&:user)).uniq
```
saraycp, line 254: The combination of the two lists is what we need? :thinking: 

Is not better to use `project.watched_items` when the feature flag is enabled and `project.watched_projects` when it is not enabled?

krauselukas: the problem is user might miss out notifications when they join the beta program. Since we haven't migrated the watched projects from the non beta watchlist over to the new one its safer to simply use them both.

krauselukas: we also didn't remove + add event subscriptions when a user switches from beta to non beta and the other way around, I would be consistent with this decision.

saraycp (approved):

eduardoj (approved):

PR #12245, opened 2022-02-22
Show link to add a package to the watchlist for multibuild packages

eduardoj:
With the new watchlist implementation, an error was thrown when browsing the live build log for a multibuild package. The guilty was `@package` returning a string instead of an object.

This PR includes the following changes:
- In case of a multibuild package: `@package` used to be a string, now `@package` contains an object.
- `@package_name` is introduced and contains the package string.
- The retrieval of build logs of remote projects was not supported. This code was removed. This means that `@package` is no longer nil.
- Thanks to the previous changes, checking for the permissions is more straight forward, as `@package` is no longer nil.

These changes imply some improvements in the UI for the live build log page for a multibuild package.

**Before:**
- The tabs where not shown
- The breadcrums contained a link to a package with the multibuild flavor. This always redirected to the base package (the package without the flavor).

Screenshot from 2022-02-25 09-41-49

**Now:**
- The tabs are shown.
- The breadcrums contains a link the the base package.

Screenshot from 2022-02-25 09-42-10

Fixes #12242 .

Co-authored-by: Saray Cabrera Padrón <scabrerapadron@suse.de>

**TODO:**
- [x] Add tests.

PR #12255, opened 2022-02-25
Avoid watchlist to crash with special cases

saraycp:
Right now, the new watchlist implementation fails when `@project` or `@package` don't contain an object but a string or nil. 
This happens when we have links and multibuild packages.

With the introduced code, we make sure we always get an object when this is possible. Otherwise, we assume they are nil, so we don't offer an item to add to the watchlist. We decided to add that logic inside the `WatchlistComponent` because it would be too much ruby code for the layout view.

In the case of inherited packages belonging to project links, we decided not to provide the link to add the package to the watchlist. It's not straightforward to keep track of the linked project.

After this is merged, we can enable back the `new_watchlist` feature flag in production. First for the `staff` group and then for the `beta` group.

### DIFF
--- a/src/api/app/components/watchlist_component.rb
+++ b/src/api/app/components/watchlist_component.rb
@@ -11,15 +11,36 @@ class WatchlistComponent < ApplicationComponent
     'BsRequest' => 'Watch this request'
   }.freeze
 
-  def initialize(user:, project: nil, package: nil, bs_request: nil)
+  def initialize(user:, bs_request: nil, package: nil, project: nil)
     super
 
     @user = user
-    # NOTE: the order of the array is important, when project and package are both present we ensure it takes package.
-    @object_to_be_watched = [bs_request, package, project].compact.first
+    @object_to_be_watched = object_to_be_watched(bs_request, package, project)
   end
 
   private
+
+  def object_to_be_watched(bs_request, package, project)
+    return bs_request if bs_request
+
+    # this is a remote project, don't offer watching.
+    return unless project.is_a?(Project)
+
+    # there is no package, offer watching the project
+    return project unless package
+
+    # maybe package is a multibuild flavor? Try do look up the object of the flavor.
+    # rubocop:disable Style/IfUnlessModifier
+    if package.is_a?(String)
+      package = Package.get_by_project_and_name(project, package, { follow_multibuild: true })
+    end
+    # rubocop:enable Style/IfUnlessModifier
+
+    # the package is coming via a project link, don't offer watching it.
+    return if package.project != project
+
+    package
+  end
 
   def object_to_be_watched_in_watchlist?
     @user.watched_items.exists?(watchable: @object_to_be_watched)
--- a/src/api/app/views/layouts/webui/webui.html.haml
+++ b/src/api/app/views/layouts/webui/webui.html.haml
@@ -78,8 +78,9 @@
       - if User.session
         - if feature_enabled?(:new_watchlist)
           = render WatchlistComponent.new(user: User.session!,
-          project: @project,
-          package: @package,
-          bs_request: @bs_request)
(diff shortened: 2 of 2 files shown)


src/api/app/views/layouts/webui/webui.html.haml
```diff
@@ -77,9 +77,22 @@
...
       - if User.session
         - if feature_enabled?(:new_watchlist)
+          :ruby
+            # NOTE: temporary fix to avoid watchlist crash with special packages or projects (remote, links, multibuild)
+            project = @project if @project.is_a?(Project)
+            package = @package if project && @package.is_a?(Package)
+
+            # NOTE: For a while, we'll track in the log those special cases for future improvements.
+            if (@project && !project) || (@package && !package)
+              Rails.logger.debug do
```
hennevogel, line 87: We don't do debug logs in production

saraycp: This was removed with the last changes.

dmarcoux:
I'm against this change, because as [I wrote in a previous PR which was quite similar](https://github.com/openSUSE/open-build-service/pull/12245#discussion_r812658017):

> If we provide a package which isn't really a package, but a string to this component, we should change the code where this component is instantiated.
>
>A view component should deal with view code, not fetching data from the database. (...)

What is preventing us from repeating the same approach as #12245? This would imply refactoring the controllers which are returning strings instead of a package/project/bs_request.

dmarcoux (dismissed):
I'm against this change and I refer to a comment [I wrote in a previous PR which was quite similar](https://github.com/openSUSE/open-build-service/pull/12245#discussion_r812658017):

> If we provide a package which isn't really a package, but a string to this component, we should change the code where this component is instantiated.
>
>A view component should deal with view code, not fetching data from the database. (...)

What is preventing us from repeating the same approach as #12245? This would imply refactoring the controllers which are returning strings instead of a package/project/bs_request. I agree that the layout view shouldn't handle this logic, this is why it belongs in the controllers, not in view code.

src/api/app/components/watchlist_component.rb
```diff
@@ -11,16 +11,36 @@ class WatchlistComponent < ApplicationComponent
     'BsRequest' => 'Watch this request'
   }.freeze
 
-  def initialize(user:, project: nil, package: nil, bs_request: nil)
+  def initialize(user:, bs_request: nil, package: nil, project: nil)
```
dmarcoux, line 14: What is the reason behind this change?

eduardoj: To arrange the parameters in the same order they will be called inside the initialize, a couple of lines below:
```
@object_to_be_watched = object_to_be_watched(bs_request, package, project)
```

dmarcoux: Humm... alright.

dmarcoux:
It doesn't feel good to patch the view component, it's definitely adding technical debt instead of solving the core issue in the controllers. How many controllers are causing issues?

eduardoj:
> It doesn't feel good to patch the view component, it's definitely adding technical debt instead of solving the core issue in the controllers. How many controllers are causing issues?

That's the question. If we patch the view component without adding any alert or log, as it is currently proposed by the PR, I feel we lose the chance of knowing what to fix.

If we don't do debug logs in production, could we just change that line to this one?
```
Rails.logger.info do
```
Would this be fine for you, @hennevogel?

hennevogel:
> It doesn't feel good to patch the view component, it's definitely adding technical debt instead of solving the core issue in the controllers. How many controllers are causing issues?

There is not core issue in controllers. There is an architecture (Links) that allows this to happen. You will not solve this problem in *any* code on top.

*Please* work with this architecture in the *new* component that you add instead of moving this deeper into the stack.

PR #12256, opened 2022-02-25
Be more specific with the naming of the `watcher` receiver role

krauselukas:
Since we are going to support to receive notifications
for packages and requests followed in the watchlist, we need
to add new receiver roles to the event subscriptions. In order
to add the new watcher roles, we need to rename the existing
watcher roles for projects.

Extracted from PR #12253

### DIFF
--- a/src/api/app/models/event/base.rb
+++ b/src/api/app/models/event/base.rb
@@ -247,7 +247,7 @@ module Event
       ret
     end
 
-    def watchers
+    def project_watchers
       project = ::Project.find_by_name(payload['project'])
       return [] if project.blank?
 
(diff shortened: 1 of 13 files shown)


krauselukas:
[Get notifications about packages and requests I'm watching (L)](https://trello.com/c/XmvYa9qx/1840-get-notifications-about-packages-and-requests-im-watching-l)

hennevogel (changes requested):
What are you going to do about existing `EventSubscription`?

From production:
```ruby
[4] pry(main)> EventSubscription.where(receiver_role: 'watcher').count
=> 912
```

krauselukas:
@hennevogel right, there needs to be a data migration! We also observed another thing, which makes it a bit tricky to keep the changes from this PR seperated from https://github.com/openSUSE/open-build-service/pull/12253 will close this PR now and adapt the other one

PR #12260, opened 2022-02-28
Add type parameter to mark read/unread button

### LINKED ISSUE #12231: Mark as "unread" on the object page does not work
# Issue Description

The "Mark as unread" button on the object page does not work.

Request-17-Submit-package_1644340548_4-Open-Build-Service

Expected Result
============
If I click on "Mark as unread" from the request/package/project page, the object should appear on the Unread list again.  

How to Reproduce
===============

1. Go to the Notifications page
2. Mark one request notification as Read
3. Go to the request page
4. Click on the Mark as unread button you find in the top-right corner (marked in the screenshot)
5. You'll see it doesn't move the notification back to "Unread"

rubhanazeem:
On `_notification_toolbar` the `Mark as unread` button was not working because of the missing parameter `type`. 

Closes #12231


src/api/spec/helpers/webui/notification_helper_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe Webui::NotificationHelper, type: :helper do
```
dmarcoux, line 3: Those specs aren't testing much. I wonder if testing this helper is worth it. There isn't much logic beside the two ternary operators. The rest of the code relies on Rails methods which are covered upstream.

Testing the helper to check the type, title and icon could be done. What do you think?

src/api/spec/helpers/webui/notification_helper_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  describe '#mark_as_read_or_unread_button' do
+    context 'for unread notification' do
+      let(:notification) { create(:web_notification, delivered: false) }
+
+      it 'contains type as an additional parameter' do
```
dmarcoux, line 8: This is exactly the same description as in the spec for a read notification, although the link isn't the same. This doesn't explain what is different in those 2 specs.

src/api/spec/helpers/webui/notification_helper_spec.rb
rubhanazeem: Good idea. I've added additional specs for title and icon :slightly_smiling_face:

src/api/spec/helpers/webui/notification_helper_spec.rb
rubhanazeem: Both specs test the same thing. One for `read` notification and one for `unread`. But I've removed the `it` block. It's updated.

dmarcoux (approved):

PR #12262, opened 2022-02-28
Fix reencoding when no fail log present

danidoni:


src/api/spec/mailers/event_mailer_spec.rb
```diff
@@ -155,8 +155,18 @@
...
+      context 'and there is a payload' do
+        it 'renders the body' do
+          expect(mail.body.encoded).to have_text('Last lines of build log:')
+        end
+      end
+
+      context 'but there is no payload' do
+        let(:faillog) { nil }
+
+        it 'renders the body' do
```
dmarcoux, line 167: This is true? I think it should be something like:

```suggestion
        it 'renders the body, but does not have a build log' do
```

dmarcoux:
@danidoni A flickering spec failed, you'll have to restart the CI yourself as I cannot do it since it's your CircleCI fork.

vpereira (approved):

PR #12269, opened 2022-03-01
Add events for creating and deleting relationships

eduardoj:

### DIFF
--- a/src/api/app/models/relationship.rb
+++ b/src/api/app/models/relationship.rb
@@ -60,6 +60,8 @@ class Relationship < ApplicationRecord
     bugowners.joins(:user).merge(User.with_email)
   }
 
+  after_create :create_event_after_create
+
   # we only care for project<->user relationships, but the cache is not *that* expensive
   # to recalculate
   after_create :discard_cache
@@ -125,6 +127,12 @@ class Relationship < ApplicationRecord
     with_groups_and_roles_query.pluck('groups.title', 'roles.title')
   end
 
+  def create_event_before_delete
+    return unless User.session
+
+    Event::RelationshipDelete.create(event_parameters)
+  end
+
   private
 
   class << self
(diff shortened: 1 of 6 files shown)


src/api/app/models/relationship.rb
```diff
@@ -158,6 +166,20 @@ def check_global_role
   def allowed_user
     raise NotFoundError, "Couldn't find user #{user.login}" if user && user.is_nobody?
   end
+
+  def event_create
+    return if User.session == user
```
hennevogel, line 171: Not sure we should skip this, this could rather be limited by `receiver_roles` right?

src/api/app/models/relationship.rb
```diff
@@ -125,6 +127,12 @@ def self.with_groups_and_roles
     with_groups_and_roles_query.pluck('groups.title', 'roles.title')
   end
 
+  def event_delete
+    return if User.session == user
```
dmarcoux, line 131: Could we rely on Pundit here instead? This way, authorization code is only in policies. It's obviously not the case right now in OBS, but it's the end goal.

src/api/app/models/relationship.rb
dmarcoux, line 171: Same comment as above to use Pundit.

src/api/app/models/relationship.rb
dmarcoux: Oh alright!

src/api/app/models/relationship.rb
eduardoj: True. `receiver_roles` could be set to define who gets notified. But it is not need for now. We can analyze `receiver_roles` better when we implement the notifications.

src/api/app/models/relationship.rb
```diff
@@ -60,6 +60,8 @@ class Relationship < ApplicationRecord
     bugowners.joins(:user).merge(User.with_email)
   }
 
+  after_create :event_create
```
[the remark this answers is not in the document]
eduardoj: I changed `event_create` to `create_event_after_create`, and `event_delete` to `create_event_before_delete`. Please take a look again.

PR #12277, opened 2022-03-04
Can not store files in SCM bridged packages

hennevogel:
They are maintained on the SCM...

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -1164,6 +1164,8 @@ class Package < ApplicationRecord
   end
 
   def delete_file(name, opt = {})
+    raise ScmsyncReadOnly if scmsync.present?
+
     delete_opt = {}
     delete_opt[:keeplink] = 1 if opt[:expand]
     delete_opt[:user] = User.session!.login
@@ -1239,6 +1241,8 @@ class Package < ApplicationRecord
   end
 
   def save_file(opt = {})
+    raise ScmsyncReadOnly if scmsync.present?
+
     content = '' # touch an empty file first
     content = opt[:file] if opt[:file]
 
--- a/src/api/app/models/service.rb
+++ b/src/api/app/models/service.rb
@@ -1,6 +1,7 @@
 # Class to read and write the "_services" file on the Backend
 class Service
   include ActiveModel::Model
+  include Package::Errors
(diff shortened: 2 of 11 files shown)


hennevogel:
@adrianschroeter should something happen with existing files if you "convert" a package to scmsync?

hennevogel:
> should something happen with existing files if you "convert" a package to scmsync?

I see, the service is just removing them...

adrianschroeter:
it would be good if the error message contains the URL to the scm, that would allow the user to work on the sources directly

hennevogel:
> it would be good if the error message contains the URL to the scm, that would allow the user to work on the sources directly

It now always present and file editing isn't offered anymore.

Screenshot from 2022-03-04 16-48-40

src/api/app/models/package.rb
```diff
@@ -1161,6 +1161,8 @@ def patchinfo
   end
 
   def delete_file(name, opt = {})
+    raise ScmsyncReadOnly if scmsync.present?
```
adrianschroeter, line 1164: I would still handover the scm url to the user here, so osc users have a chance to see where they can proceed.

src/api/app/models/package.rb
```diff
@@ -1236,6 +1238,8 @@ def self.verify_file!(pkg, name, content)
   end
 
   def save_file(opt = {})
+    raise ScmsyncReadOnly if scmsync.present?
```
adrianschroeter, line 1241: same here

src/api/app/models/service.rb
```diff
@@ -82,6 +83,8 @@ def add_service(name, parameters = [], mode = nil)
   end
 
   def save
+    raise ScmsyncReadOnly if package.scmsync.present?
```
adrianschroeter, line 86: and here

adrianschroeter (approved):
but also fine for me that way.

hennevogel:
hmmm

src/api/app/models/package.rb
krauselukas: @adrianschroeter Those methods are not used for the API. I added it to `source_controller.rb` and also provide the SCM URL.

hennevogel:

PR #12292, opened 2022-03-09
Add rails task to create workflow token and runs

saraycp:
With this task, we can easily fill in our development database with a workflow token and some workflow runs.
Data that helps us to try out things during development and in review-app.

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -249,6 +249,42 @@ namespace :dev do
     end
   end
 
+  # Run this task with: rails dev:workflow_runs:data"
+  namespace :workflow_runs do
+    desc 'Creates a workflow token, workflow runs, artifacts and some of their dependencies'
+    task data: :environment do
+      unless Rails.env.development?
+        puts "You are running this rake task in #{Rails.env} environment."
+        puts 'Please only run this task with RAILS_ENV=development'
+        puts 'otherwise it will destroy your database data.'
+        return
+      end
+
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+
+      puts 'Creating workflow token and workflow runs...'
+
+      admin = User.get_default_admin
+      User.session = admin
+      project = find_or_create_project(admin.home_project_name, admin)
+
+      workflow_token = Token::Workflow.find_by(name: 'Testing token') || create(:workflow_token, user: admin, name: 'Testing token')
+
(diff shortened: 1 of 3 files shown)


danidoni (approved):

src/api/lib/tasks/dev.rake
```diff
@@ -249,6 +249,65 @@ namespace :dev do
...
+      source_project = find_or_create_project(admin.home_project_name, admin)
+      target_project = find_or_create_project('home:Admin:CI:iggy:hello_world:PR-1', admin)
+
+      source_package = source_project.packages.find_by(name: 'hello_world') || create(:package_with_file, name: 'hello_world', project: source_project, file_name: 'README.md', file_content: '# Lorem ipsum')
+      target_package = target_project.packages.find_by(name: 'hello_world') || create(:package_with_file, name: 'hello_world', project: target_project, file_name: 'README.md', file_content: '# Lorem ipsum')
+
+      token = Token::Workflow.find_by(name: 'Testing token') || create(:workflow_token, user: admin, name: 'Testing token')
+
+      ['running', 'success', 'fail'].each do |status|
+        workflow_run = create(:workflow_run, token: token, status: status)
```
dmarcoux, line 280: Would it be possible to also have data in `response_body` and `response_url`?

src/api/lib/tasks/dev.rake
```diff
@@ -249,6 +249,65 @@ namespace :dev do
...
+    task data: :environment do
+      unless Rails.env.development?
+        puts "You are running this rake task in #{Rails.env} environment."
+        puts 'Please only run this task with RAILS_ENV=development'
+        puts 'otherwise it will destroy your database data.'
+        return
+      end
+
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
```
hennevogel, line 264: soo why isn't all of this a factory?

src/api/lib/tasks/dev.rake
saraycp: As `response_body` and `response_url` change depending on different statuses, I decided to create factories for workflow runs with different status and refactor the task code a bit.

src/api/lib/tasks/dev.rake
saraycp: I added a new commit on top to create the objects to which the artifacts link to.

dmarcoux (approved):

PR #12295, opened 2022-03-10
Don't retry requesting the last chunk on live build log

eduardoj:
It is unnecessary to repeat the request of a log chunk after reaching a final state. This is a follow up to fixing the live build log not to stop in a not final state.

Follow up to #12294.

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -485,8 +485,6 @@ class Webui::PackageController < Webui::WebuiController
         @finished = true
       end
     end
-
-    logger.debug 'finished ' + @finished.to_s
   end
 
   def abort_build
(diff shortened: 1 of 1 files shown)


dmarcoux:
Should this be marked as ready for review?

eduardoj:
> Should this be marked as ready for review?

This PR could be marked as ready for review after #12994 is merged. This PR is based in the branch of #12294. Their two first commits are the same of #12294.

I wanted to separate the changes in two PRs, to explain better the changes introduced. And also to point out that this last one is an improvement to the code after the first PR is merged.

> I don't understand this PR... especially after having a look at #12294. So the changes in #12294 are not needed? Calling again `get_log_chunk` isn't needed?

IMHO, after a build has reached a final state (this is guaranteed after #12294 is merged), calling again `get_log_chunk` makes little sense.

dmarcoux:
I'm confused because this PR is removing the `if` statement which is changed in #12294.

dmarcoux (approved):

PR #12296, opened 2022-03-10
[backend] add call to imadata

aplanas:
The `imadata` script should land in [system:packagemanager](https://build.opensuse.org/project/show/system:packagemanager) if this moves forward: https://build.opensuse.org/request/show/960798

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -734,7 +734,19 @@ sub createrepo_rpmmd {
     }
     unlink("$extrep/repodata/modules.yaml");
   }
-  
+
+  # add imadata data
+  # IMA metadata could be included as a feature in createrepo_c (check
+  # issue #313), once included it will deprecate this section, but can
+  # cause a problem if the XML name is different.
+  unlink("$extrep/repodata/imadata.xml");
+  my $imadata_bin = '/usr/bin/imadata';
+  if (-e "$imadata_bin") {
+      print "    adding imadata.xml to repodata\n";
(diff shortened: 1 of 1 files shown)


adrianschroeter:
Is there any bigger planing regarding IMA? 
Just adding this here won't help much and we would need to stay forever compatible if we do so now....
Is there any write up what is planed, including feeds for products, maintenance and friend channels?

adrianschroeter:
Beside of the publishing topic, we discussed the performance impact of this. We would prefer to avoid to open all rpms another time.

@M0ses can you check if we could extend createrepo_c to generate this file?

We could use it as well for product building then. Please refer to this upstream discussion:

https://github.com/rpm-software-management/createrepo_c/issues/313

apart from this we would need an option flag to enable this feature, because we do not want to change all existing repositories. But we can take care once we have the tooling.

aplanas:
> We would prefer to avoid to open all rpms another time.

The imadata.py code (https://github.com/aplanas/imadata/blob/main/imadata.py#L23) is reading the header, without checking the signatures. Locally is fast, and is using `multiprocessing` to load balance the reading.

But I agree that is better a full upstream support from createrepo_c, as was what I intended last week.

aplanas:
As an aside, one interesting area is how to format the list to make better gzip compression.

The sha256 hashes are almost random data, so I am wondering if instead of using XML, we can use some plain CSV, and order the lines by the hash column. This can help the compression ratio. The problem is that is not XML.

M0ses:
superseded by fddfa1ce6d505b405e6a87eb3d3f436c5a79e332

PR #12299, opened 2022-03-11
Introduce browser notifications for live build logs

### LINKED ISSUE #7920: Add browser notification in the build log page
After opening a build log page I'm interested on, I want to be notified after the build finishes with a browser notification.

See https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API for more information.

eduardoj:
Now it is possible to receive browser notifications from the live build log.

Fixes #7920

Example of a notification of a succeeded build:

Screenshot from 2022-03-11 14-23-53_browser_notification_succeeded

Example of a notification of a failed build:

Screenshot from 2022-03-11 14-25-11_browser_notification_failed

A button "Browser Notifications" was added to ask for permission to receive browser notifications:
- It will appear enabled when notifications are not yet allowed or denied.
- It will appear disabled when the permission is already granted or revoked. In this case, the permission could be deleted through the browser configuration.

Peek 2022-03-11 14-14_Browser_notifications_on_off

To test this PR:
- Go to a live log page of a package in the [review app](https://obs-reviewlab.opensuse.org/eduardoj-live_build_log_browser_notifications/package/live_build_log/home:Admin/hello_world/openSUSE_Tumbleweed/x86_64).
- Enable your browser notifications:
  - Click in the button "Browser Notifications".
  - Answer with "Allow".
- Click on "Trigger Rebuild"
- Return to the live build log page.
- Wait until the build finishes. Then, a browser notification should appear in your environment.

### DIFF
--- a/src/api/app/assets/javascripts/webui/live_build_log.js
+++ b/src/api/app/assets/javascripts/webui/live_build_log.js
@@ -2,6 +2,7 @@ var LiveLog = function(wrapperId, startButton, stopButton, status, finished, inf
   this.wrapper = $(wrapperId);
   this.startButton = $(startButton);
   this.stopButton = $(stopButton);
+  this.notificationsButton = $('.link_notifications');
   this.status = $(status);
   this.faviconFinished = faviconFinished;
   this.faviconRunning = faviconRunning;
@@ -20,6 +21,13 @@ $.extend(LiveLog.prototype, {
   initialize: function() {
     this.startButton.click($.proxy(this.start, this));
     this.stopButton.click($.proxy(this.stop, this));
+    if (Notification.permission === 'default') {
+      this.notificationsButton.click($.proxy(this.askNotificationPermission, this));
+    }
+    else {
+      this.notificationsButton.html(this.notificationsButton.html().replace('Request Browser Notifications', 'Browser Notifications ' + Notification.permission));
+      this.notificationsButton.addClass('disabled');
+    }
     this.start();
     this.initial = false;
     return this;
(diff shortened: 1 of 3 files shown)


src/api/app/assets/javascripts/webui/live_build_log.js
```diff
@@ -20,6 +21,12 @@ $.extend(LiveLog.prototype, {
   initialize: function() {
     this.startButton.click($.proxy(this.start, this));
     this.stopButton.click($.proxy(this.stop, this));
+    if (Notification.permission === 'granted' || Notification.permission === 'denied') { // jshint ignore:line
```
vpereira, line 24: you could use `jQuery.inArray()` for that like we do in ruby, no?

```
if(jQuery.inArray(Notification.permission, ['granted', 'denied']) === -1) 
{
  // not found 
} else {
  // found
}
```

src/api/app/views/webui/package/_live_build_log_controls.html.haml
```diff
@@ -30,3 +30,7 @@
               class: 'link_abort_build live-link-action btn-danger d-none') do
       %i.far.fa-times-circle
       Abort Build
+
+  = link_to('#', class: 'link_notifications live-link-action btn-secondary ml-auto', title: 'Request permission for browser notifications') do
+    %i.far.fa-bell
+    Browser Notifications
```
danidoni, line 36: I would rephrase this as an action, like 'Request Browser Notifications', then once disabled I would add a help title like 'Browser Notifications Granted' or something like this so the user knows why it's greyed out

src/api/app/assets/javascripts/webui/live_build_log.js
eduardoj: You made me revisit the documentation. Given that `Notification.permission` [can only have three possible values](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API#checking_current_permission_status), I will change the condition to only check for `default`.

src/api/app/views/webui/package/_live_build_log_controls.html.haml
danidoni: Okay, but I would change the title to something else so it's clear that you cannot click on that button because you already did it.

src/api/app/assets/javascripts/webui/live_build_log.js
eduardoj: I changed the condition. Please take a look again.

src/api/app/views/webui/package/_live_build_log_controls.html.haml
eduardoj: I renamed the button to 'Request Browser Notifications'. Please have a look.

eduardoj: > Okay, but I would change the title to something else so it's clear that you cannot click on that button because you already did it.

Now it will show either "Browser Notifications Granted" or "Browser Notifications Denied". I will extract the duplicated code into a function later. Please have a look.

danidoni (approved):

dmarcoux (approved):

src/api/app/assets/javascripts/webui/live_build_log.js
```diff
@@ -95,29 +103,58 @@ $.extend(LiveLog.prototype, {
...
+      Notification.requestPermission().then(function (permission) { // jshint ignore:line
+        if (permission === 'granted' || permission === 'denied') {
+          liveLog.notificationsButton.html(liveLog.notificationsButton.html().replace('Request Browser Notifications', 'Browser Notifications ' + Notification.permission)); // jshint ignore:line
+          liveLog.notificationsButton.addClass('disabled'); // jshint ignore:line
+        }
+      });
+    }
+  },
+
+  notificationSupport: function () {
```
dmarcoux, line 152: Whenever we start having more browser notifications, this should be extracted to another file. For now, it's good enough.

PR #4842, opened 2018-04-11
[ci] Run Rubocop with Ruby 2.5.0

dmarcoux:

PR #12302, opened 2022-03-11
Upgrade OBS to use Ruby 3.1

danidoni:
Due to the EOL of Ruby 2.5 we need to make OBS compatible with recent Ruby versions, otherwise we are gonna fall behind gem updates. Some of those gem updates drop compatibility with Ruby 2.5.

See #12211, or #12237

You can test this change locally following this guide:

https://github.com/openSUSE/open-build-service/wiki/Ruby-Update#development-environment


src/api/app/helpers/maintenance_helper.rb
```diff
@@ -140,7 +140,7 @@ def release_package_create_main_package(request, source_package, target_package_
     upload_params[:cmd] = 'commitfilelist'
     upload_params[:noservice] = '1'
     upload_params[:requestid] = request.number if request
-    upload_path = "/source/#{URI.escape(target_project.name)}/#{URI.escape(base_package_name)}"
+    upload_path = "/source/#{CGI.escape(target_project.name)}/#{CGI.escape(base_package_name)}"
```
coolo, line 143: be careful with CGI.escape on path components.

```
URI.escape('devel:libraries:c_c++')
=> "devel:libraries:c++"

curl 'src-opensuse:5352/source/devel:libraries:c_c++' => 
<directory>
....

CGI.escape('devel:libraries:c_c++')
=> "devel%3Alibraries%3Ac%2B%2B"

curl 'src-opensuse:5352/build/devel%3Alibraries%3Ac%2B%2B'
<status code="404">
  <summary>project 'devel:libraries:c++' does not exist</summary>
  <details>404 project 'devel:libraries:c++' does not exist</details>

```

src/api/app/helpers/webui/webui_helper.rb
```diff
@@ -13,10 +13,8 @@ def bugzilla_url(email_list = '', desc = '')
       cc = ('&cc=' + email_list[1..-1].join('&cc=')) if email_list
     end
 
-    URI.escape(
-      "#{@configuration['bugzilla_url']}/enter_bug.cgi?classification=7340&product=openSUSE.org" \
-      "&component=3rd party software&assigned_to=#{assignee}#{cc}&short_desc=#{desc}"
-    )
+    "#{@configuration['bugzilla_url']}/enter_bug.cgi?classification=7340&product=openSUSE.org" \
+      "&component=3rd%20party%20software&assigned_to=#{CGI.escape(assignee.to_s)}#{CGI.escape(cc.to_s)}&short_desc=#{CGI.escape(desc.to_s)}"
```
coolo, line 17: Consider 

```
{assigned_to: assignee, short_desc: desc, component: '3rd party software'}.to_query
=> "assigned_to=assignee&component=3rd+party+software&short_desc=desc"
```

A little less involved

src/api/app/helpers/maintenance_helper.rb
mlschroe: This seems to be a copy'n'paste problem on your side:
```
CGI.escape('devel:libraries:c_c++')
=> "devel%3Alibraries%3Ac_c%2B%2B"
curl 'src-opensuse:5352/build/devel%3Alibraries%3Ac_c%2B%2B'
<directory>
...
```

coolo: Right, it's c_c++ :)

coolo: Still be careful :)

mlschroe: The problem case is spaces in path components:
```
CGI.escape('foo bar')
=> "foo+bar"
```
That won't fly as path component.

src/api/app/models/package_file.rb
```diff
@@ -6,6 +6,6 @@ class PackageFile < Backend::File
 
   # calculates the real url on the backend to search the file
   def full_path(query = {})
-    URI.encode("/source/#{project_name}/#{package_name}/#{name}") + "?#{query.to_query}"
+    CGI.unescape("/source/#{project_name}/#{package_name}/#{name}") + "?#{query.to_query}"
```
mlschroe, line 9: Unescape? Don't you mean escape?

danidoni: Yes, we know that's not right. We're fixing it [somewhere else][1] to reduce this PR. 

[1]: #12304

src/api/app/helpers/webui/webui_helper.rb
hennevogel: all of this is done in the other PRs now...

dmarcoux:
Should we update [.ruby_version](https://github.com/openSUSE/open-build-service/blob/02d0a500f9b024e6970a629713c83ea1506f385a/.ruby-version)?

danidoni:
> Should we update [.ruby_version](https://github.com/openSUSE/open-build-service/blob/02d0a500f9b024e6970a629713c83ea1506f385a/.ruby-version)?

Yes, but... Do we use that? In dev we have the ruby version set by the docker containers and in production we have the ruby version set by the OS...

rubhanazeem:
> > Should we update [.ruby_version](https://github.com/openSUSE/open-build-service/blob/02d0a500f9b024e6970a629713c83ea1506f385a/.ruby-version)?
> 
> Yes, but... Do we use that? In dev we have the ruby version set by the docker containers and in production we have the ruby version set by the OS...

If it's not used, should we remove it?

vpereira:
> Yes, but... Do we use that? In dev we have the ruby version set by the docker containers and in production we have the ruby version set by the OS...

You use it if you have different ruby versions in your system. In production I don't think it's necessary. In the appliance, it is probaly necessary. I would keep it and update it to 3.1

danidoni:
> > Yes, but... Do we use that? In dev we have the ruby version set by the docker containers and in production we have the ruby version set by the OS...
> 
> You use it if you have different ruby versions in your system. In production I don't think it's necessary. In the appliance, it is probaly necessary. I would keep it and update it to 3.1

Okay, udpated.

coolo:
It seems only for old Travis CI: https://github.com/openSUSE/open-build-service/pull/4842/commits/df2bc7b938d4176222cbafacf93ee7c1d9623396

hennevogel:
> It seems only for old Travis CI

Nope, it's also use by various other tools. Like rubocop...

hennevogel (approved):

PR #12304, opened 2022-03-11
URI.encode is removed, use Addressable::URI.encode

danidoni:
See https://github.com/ruby/uri/commit/61c6a47ebf1f2726b60a2bbd70964d64e14b1f98

`URI.escape` is deprecated in Ruby 3.1. This PR helps to reduce the size of the `ruby31` branch, the upcoming changeset to make OBS compatible with Ruby 3.1.

See #12302

PR #12305, opened 2022-03-11
Replace URI.escape with Addressable::URI.escape 

danidoni:
In order to prepare for the update of Ruby 3.1, we're extracting as many changes as possible and pushing them to master. 

`URI.escape` will be deprecated in Ruby 3.1 (see #12304)

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -428,8 +428,9 @@ class Package < ApplicationRecord
   end
 
   def self.source_path(project, package, file = nil, opts = {})
-    path = "/source/#{URI.escape(project)}/#{URI.escape(package)}"
-    path += "/#{URI.escape(file)}" if file.present?
+    path = "/source/#{project}/#{package}"
+    path = Addressable::URI.escape(path)
+    path += "/#{CGI.escape(file)}" if file.present?
     path += '?' + opts.to_query if opts.present?
     path
   end
@@ -1324,7 +1325,7 @@ class Package < ApplicationRecord
       permitted_params = params.permit(:repository, :arch, :package, :code, :wipe)
 
       # do not use project.name because we missuse the package source container for build container operations
-      Backend::Connection.post("/build/#{URI.escape(build_project)}?cmd=#{command}&#{permitted_params.to_h.to_query}")
+      Backend::Connection.post(Addressable::URI.escape("/build/#{build_project}?cmd=#{command}&#{permitted_params.to_h.to_query}"))
     rescue Backend::Error, Timeout::Error, Project::WritePermissionError => e
       errors.add(:base, e.message)
       return false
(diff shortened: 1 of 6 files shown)


dmarcoux:
Is there something missing to mark this as ready for review? Otherwise, I would approve and merge this PR.

darix:
CGI.escape for path components feels wrong. did you test that with paths that contain " " e.g.?

coolo:
Did some testing: https://github.com/openSUSE/open-build-service/issues/12308 - the problems go deep :)

danidoni:
> CGI.escape for path components feels wrong. did you test that with paths that contain " " e.g.?
> 
> Doesnt rails pull "addressable" already and you could just use that?
> 
> https://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape

Yes, indeed. See https://github.com/openSUSE/open-build-service/pull/12304

danidoni:
[Ruby 3.1 Update](https://trello.com/c/l1cHI0L9/1847-ruby-31-update)

hennevogel (approved):

rubhanazeem (approved):

src/api/app/models/package.rb
```diff
@@ -428,8 +428,9 @@ def sources_changed(opts = {})
   end
 
   def self.source_path(project, package, file = nil, opts = {})
-    path = "/source/#{URI.escape(project)}/#{URI.escape(package)}"
-    path += "/#{URI.escape(file)}" if file.present?
+    path = "/source/#{project}/#{package}"
+    path = Addressable::URI.escape(path)
+    path += "/#{CGI.escape(file)}" if file.present?
```
dmarcoux, line 433: In other changes from this PR, `URI.escape` was changed to `Addressable::URI.escape`. Is there a reason here for not doing the same?

hennevogel: Anchors are a valid part of URIs.

```ruby
[6] pry(#<SourceController>)> Addressable::URI.escape('bnc#620675.diff')
=> "bnc#620675.diff"
[6] pry(#<SourceController>)> CGI.escape('bnc#620675.diff')
=> "bnc%23620675.diff"
```

hennevogel: `file` is from `params` which is unescaped by RoR. We need to escape it again to send to the backend.

dmarcoux: Sure about anchors. I just wonder why it's not like this:

```suggestion
    path += "/#{Adresssable::URI.escape(file)}" if file.present?
```

dmarcoux: Alright, thanks for the output in your previous comment. I didn't check before how those behave differently.

danidoni: Yes, we found out that in this case `Addressable::URI.escape` was not escaping some characters that we need to be escaped... For example a `#` in the middle of a package name.

PR #12307, opened 2022-03-11
Adapt rails task to add special projects and packages

saraycp:
Project with link, multibuild package and packages with local and remote links.

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -304,7 +304,9 @@ namespace :dev do
       admin = User.get_default_admin
       User.session = admin
 
-      interconnect = create(:project, name: 'openSUSE.org', remoteurl: 'https://api.opensuse.org/public')
+      interconnect = create(:remote_project, name: 'openSUSE.org', remoteurl: 'https://api.opensuse.org/public')
+      # The interconnect doesn't work unless we set the distributions
+      FetchRemoteDistributionsJob.perform_now
       tw_repository = create(:repository, name: 'snapshot', project: interconnect, remote_project_name: 'openSUSE:Factory')
 
       # the home:admin is not created because the Admin user is created in seeds.rb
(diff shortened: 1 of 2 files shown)


src/api/lib/tasks/dev.rake
```diff
@@ -276,7 +303,9 @@ namespace :dev do
       admin = User.get_default_admin
       User.session = admin
 
-      interconnect = create(:project, name: 'openSUSE.org', remoteurl: 'https://api.opensuse.org/public')
+      interconnect = create(:remote_project, name: 'openSUSE.org', remoteurl: 'https://api.opensuse.org/public')
+      # The interconnect doesn't work unless we set the distributuions
```
danidoni, line 307:
```suggestion
      # The interconnect doesn't work unless we set the distributions
```

src/api/lib/tasks/dev.rake
```diff
@@ -276,7 +276,9 @@ namespace :dev do
       admin = User.get_default_admin
       User.session = admin
 
-      interconnect = create(:project, name: 'openSUSE.org', remoteurl: 'https://api.opensuse.org/public')
+      interconnect = create(:remote_project, name: 'openSUSE.org', remoteurl: 'https://api.opensuse.org/public')
+      # The interconnect doesn't work unless we set the distributions
+      FetchRemoteDistributionsJob.perform_now
```
saraycp, line 281: I'm trying to create the distribution(s) using factories instead of calling the job. Stil trying to make it build when the distribution is there.

saraycp: In the end, we decided to leave the call to the job.

danidoni (approved):

src/api/lib/tasks/dev.rake
```diff
@@ -431,6 +433,12 @@ namespace :dev do
       Configuration.download_url = 'https://download.opensuse.org'
       Configuration.save
 
+      # Other special projects and packages
+      create(:project, name: 'linked_project', link_to: home_admin)
+      create(:multibuild_package, project: home_admin, name: 'multibuild_package')
+      create(:package_with_link, project: home_admin, name: 'linked_package')
+      create(:package_with_remote_link, project: home_admin, name: 'remotely_linked_package', remote_project_name: 'openSUSE.org:openSUSE:Factory', remote_package_name: 'aaa_base')
```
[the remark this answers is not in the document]
saraycp: I didn't manage to fix it, the error is ` Files could not be expanded: openSUSE.org:openSUSE:Factory/aaa_base: bad uri: /source//aaa_base?expand=1 `

But I considered that, at least, we have the package there with the correct `_link` file pointing to the remote package, even if it's not fetching correctly the files. But I can also remove it if it makes no sense.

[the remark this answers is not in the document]
krauselukas: Not sure what going on, for me its the same, and I cant use the interconnect manually after running this rake task, I throws an error when trying to branch a package through it

dmarcoux:
Rebasing the branch for the Ruby update `3.1`.

PR #12311, opened 2022-03-15
Ensure faker-generated data is valid

danidoni:
Faker sometimes generates names of cats that have spaces, but Group
does not allow spaces in the title.

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -193,7 +193,7 @@ namespace :dev do
 
       # Users
       admin = User.where(login: 'Admin').first || create(:admin_user, login: 'Admin')
-      group = create(:groups_user, user: admin, group: create(:group, title: Faker::Creature::Cat.name)).group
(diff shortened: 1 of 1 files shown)


src/api/lib/tasks/dev.rake
```diff
@@ -193,7 +193,7 @@ namespace :dev do
 
       # Users
       admin = User.where(login: 'Admin').first || create(:admin_user, login: 'Admin')
-      group = create(:groups_user, user: admin, group: create(:group, title: Faker::Creature::Cat.name)).group
+      group = create(:groups_user, user: admin, group: create(:group, title: Faker::Creature::Cat.name.tr(' ', '-'))).group
```
hennevogel, line 196: This will work until someone invents more sophisticated cat names in `Faker`.

There are other `Faker` modules for words like [lorem](https://github.com/faker-ruby/faker/blob/master/doc/default/lorem.md) or better use a sequence to avoid duplicated titles...

hennevogel (approved):

PR #12314, opened 2022-03-16
Do not check for checks and missing_checks for empty status_reports

rubhanazeem:
Most of the times status reports are nil for bs_requests and `GET '/status_reports/requests/<id>/reports'` results in an error 500. This PR avoids the 500 error and returns 404 instead.

```
Couldn't find Status::Report with [WHERE `status_reports`.`checkable_id` = ? AND `status_reports`.`checkable_type` = ?]
```

### DIFF
--- a/src/api/app/controllers/status/reports_controller.rb
+++ b/src/api/app/controllers/status/reports_controller.rb
@@ -17,7 +17,7 @@ class Status::ReportsController < ApplicationController
                        @checkable.status_reports.find_by!(uuid: params[:report_uuid])
                      else
                        # request reports don't have uuid
-                       @checkable.status_reports.first
+                       @checkable.status_reports.first!
                      end
   end
 end


coolo:
IMO this should return a 404 instead

coolo:
That's what status_reports for repos do:

```coolo@c551#>osc api /status_reports/published/openSUSE:Factory/standard/reports/NOT
Couldn't find Status::Report with [WHERE `status_reports`.`checkable_id` = ? AND `status_reports`.`checkable_type` = ? AND `status_reports`.`uuid` = ?]
```

rubhanazeem:
> IMO this should return a 404 instead

Thanks for checking it out. I'll change it to 404

src/api/app/controllers/status/reports_controller.rb
```diff
@@ -6,6 +6,8 @@ class Status::ReportsController < ApplicationController
   # GET /status_reports/published/:project_name/:repository_name/reports/:uuid
   # GET /status_reports/build/:project_name/:repository_name/:arch/reports/:uuid
   def show
+    return raise ActiveRecord::RecordNotFound, 'Empty status report' if @status_report.nil?
```
dmarcoux, line 9: In `set_status_report`, `@checkable.status_reports.find_by!(uuid: params[:report_uuid])` will already raise `ActiveRecord::RecordNotFound` if a status report isn't found. We can do the same for the other line (`@checkable.status_reports.first`) by using `first!` instead of `first`.

dmarcoux (changes requested):

PR #12320, opened 2022-03-16
ruby does not like inline editing, back to sed

hennevogel:
warning: Can't do inplace edit for stdio

https://build.opensuse.org/package/live_build_log/OBS:Server:Unstable/obs-server:obs-bundled-gems/15.3/x86_64


hennevogel:
You can verify on obs-test now

`grep '#!.*bin/ruby' /usr/lib64/obs-api/ruby/*/gems/*/bin/* /usr/lib64/obs-api/ruby/*/bin/*`

PR #12324, opened 2022-03-17
Update bundler binstub

hennevogel:
It has lot's of new features now...


dmarcoux:
What are the new features? I'm curious :smile:

dmarcoux:
The generated code is dealing with lockfiles, requirements and checking if the bundler version is matching the one from Gemfile. All stuff which bundler already does. Anyway, a binstub is just a wrapper around an executable and with the changes in this PR, I don't see what is new. They're practical for commands often used, so :+1:

hennevogel:
If `bundler` or `bundle` are in your path. If not...

PR #12325, opened 2022-03-17
Use the application binstubs

hennevogel:
Instead of relying on the commands being in the path

### DIFF
--- a/src/api/Procfile
+++ b/src/api/Procfile
@@ -1,4 +1,4 @@
-web: bundle exec rails server -b 0.0.0.0
+web: bin/rails server -b 0.0.0.0
 delayed: bundle exec script/delayed_job.api.rb run
 clock: bin/clockworkd --log-dir=log -l -c config/clock.rb run
-search: bundle exec rails sphinx:start_for_development
+search: bin/rails sphinx:start_for_development
--- a/dist/systemd/obs-clockwork.service
+++ b/dist/systemd/obs-clockwork.service
@@ -8,8 +8,8 @@ Environment = "RAILS_ENV=production"
 User = @@APACHE_USER@@
 Group = @@APACHE_GROUP@@
 WorkingDirectory = @@OBS_API_PREFIX@@
-ExecStart = bin/clockworkd --log-dir=log -l -c config/clock.rb start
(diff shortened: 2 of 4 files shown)


src/api/Procfile
```diff
@@ -1,4 +1,4 @@
 web: bundle exec rails server -b 0.0.0.0
 delayed: bundle exec script/delayed_job.api.rb run
 clock: bundle exec clockworkd --log-dir=log -l -c config/clock.rb run
-search: bundle exec rails sphinx:start_for_development
+search: bundle exec bin/rails sphinx:start_for_development
```
dmarcoux, line 4: This could be:

```suggestion
search: bin/rails sphinx:start_for_development
```

src/api/Procfile
```diff
@@ -1,4 +1,4 @@
 web: bundle exec rails server -b 0.0.0.0
```
dmarcoux, line 1: This line should also be changed.

dmarcoux:
As discussed with @hennevogel yesterday, executing `bin/rails` generates the following output since the binstub was generated by bundler. It seems like we should generate the binstub with Rails, not bundler.
```
Beginning in Rails 4, Rails ships with a `rails` binstub at ./bin/rails that
should be used instead of the Bundler-generated `rails` binstub.

If you are seeing this message, your binstub at ./bin/rails was generated by
Bundler instead of Rails.

You might need to regenerate your `rails` binstub locally and add it to source
control:

 rails app:update:bin           # Bear in mind this generates other binstubs
                                # too that you may or may not want (like yarn)

If you already have Rails binstubs in source control, you might be
inadvertently overwriting them during deployment by using bundle install
with the --binstubs option.

If your application was created prior to Rails 4, here's how to upgrade:

  bundle config --delete bin    # Turn off Bundler's stub generator
  rails app:update:bin          # Use the new Rails executables
  git add bin                   # Add bin/ to source control

You may need to remove bin/ from your .gitignore as well.

When you install a gem whose executable you want to use in your app,
generate it and add it to source control:

  bundle binstubs some-gem-name
  git add bin/new-executable
```

dmarcoux (approved):

PR #12330, opened 2022-03-18
Approver cannot be the request creator

danidoni:
In order to enforce two different reviewers, the approver cannot be the same person than the creator of a request, if the proper attribute is enabled at target project or target package.

## How to test this
1. In the review-app, go to your home project
2. Add an `CreatorCannotAcceptOwnRequests` attribute there.
3. Go to the project's hello-world package
4. Branch it and change the sources so you can submit a request back
5. Go to the request and accept it.

### Expected result
As you have the `CreatorCannotAcceptOwnRequests` attribute set and you are the same approver user as the request's creator user, an error will popup preventing you to approve the request.

If you go now and remove the attribute from the project, you'll be allowed to approve your own request.

You could also approve it with or without the attribute set if you log in with another user with proper permissions in the project (as it's not the same person as the request creator).

## TODO
- [x] Implementation of the validation
- [x] Specs
- [x] Data migration to create the new Attribute
- [x] Testing in API

### DIFF
--- a/src/api/app/models/bs_request_permission_check.rb
+++ b/src/api/app/models/bs_request_permission_check.rb
@@ -124,6 +124,13 @@ class BsRequestPermissionCheck
 
       check_newstate_action!(action, opts)
 
+      # TODO: Get the relevant project attribute, from the target project or target package. Retrieve the accepter and check if it's the same person than the creator. And fail if true
+      target_package = Package.get_by_project_and_name(action.target_project, action.target_package) if Package.exists_by_project_and_name(action.target_project, action.target_package)
+      target_project = Project.find_by_name(action.target_project) if action.target_project
+      cannot_accept_request = target_package&.find_attribute('OBS', 'CreatorCannotAcceptOwnRequests').present?
+      cannot_accept_request ||= target_project&.find_attribute('OBS', 'CreatorCannotAcceptOwnRequests').present?
+      raise BsRequest::Errors::CreatorCannotAcceptOwnRequests if cannot_accept_request && User.session!.login == req.creator
+
       # abort immediatly if we want to write and can't
       next unless accept_check && !@write_permission_in_this_action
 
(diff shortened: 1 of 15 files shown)


src/api/app/models/bs_request_permission_check.rb
```diff
@@ -124,6 +124,13 @@ def cmd_changestate_permissions(opts)
 
       check_newstate_action!(action, opts)
 
+      # TODO: Get the relevant project attribute, from the target project or target package. Retrieve the accepter and check if it's the same person than the creator. And fail if true
+      target_package = Package.get_by_project_and_name(action.target_project, action.target_package)
```
coolo, line 128: note that target packages often don't exist. this code looks crashing in such a case

danidoni: Yeah, I noticed... Incoming push...

src/api/spec/models/bs_request_spec.rb
```diff
@@ -220,6 +220,128 @@
...
+          end
+
+          it 'accepts the request' do
+            request.change_state(newstate: 'accepted', force: true)
+            expect(request.state).to be(:accepted)
+          end
+        end
+      end
+
+      context 'and the target project does not have an attribute to disallow acceptance by the creator' do
```
hennevogel, line 272: This is already tested elsewhere I bet...

src/api/spec/models/bs_request_spec.rb
```diff
@@ -220,6 +220,128 @@
...
+
+          it 'accepts the request' do
+            request.change_state(newstate: 'accepted', force: true)
+
+            expect(request.state).to be(:accepted)
+          end
+        end
+      end
+
+      context 'and the target package does not have an attribute to disallow acceptance by the creator' do
```
hennevogel, line 327: Same here. CI time is developer time 💰

hennevogel (approved):

src/api/spec/models/bs_request_spec.rb
danidoni: Hm, okay, I'll remove those specs

PR #12331, opened 2022-03-18
Rename methods to create/delete relationship events

saraycp:
The method names `create_event_after_create` and `create_event_before_delete` sound a bit weird to me. Especially in `after_create :create_event_after_create`.

I propose these other names, let me know what you think.

### DIFF
--- a/src/api/app/models/relationship.rb
+++ b/src/api/app/models/relationship.rb
@@ -60,7 +60,7 @@ class Relationship < ApplicationRecord
     bugowners.joins(:user).merge(User.with_email)
   }
 
-  after_create :create_event_after_create
+  after_create :create_relationship_create_event
 
   # we only care for project<->user relationships, but the cache is not *that* expensive
   # to recalculate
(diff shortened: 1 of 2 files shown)


rubhanazeem (approved):
I agree, the proposed names look better.

src/api/app/models/relationship.rb
```diff
@@ -127,7 +127,7 @@ def self.with_groups_and_roles
     with_groups_and_roles_query.pluck('groups.title', 'roles.title')
   end
 
-  def create_event_before_delete
+  def create_relationship_deletion_event
```
vpereira, line 130: IMO `create_relationship_delete_event` is better

src/api/app/models/relationship.rb
```diff
@@ -167,7 +167,7 @@ def allowed_user
     raise NotFoundError, "Couldn't find user #{user.login}" if user && user.is_nobody?
   end
 
-  def create_event_after_create
+  def create_relationship_creation_event
```
vpereira, line 170: IMO `create_relationship_create_event` is better

saraycp: `after_create: create_relationship_create_event` VS `after_create: create_relationship_creation_event` :thinking:

src/api/app/models/relationship.rb
saraycp: done

PR #12335, opened 2022-03-22
Do not display rolled-out beta features on Manage Beta Features page

### LINKED ISSUE #12322: Do not display rolled-out beta features on Manage Beta Features page
# Issue Description
In the [Manage Beta Features](https://build.opensuse.org/my/beta_features) page, rolled-out beta features are displayed, even though they cannot be disabled/enabled anymore. This might lead users into thinking something is broken, although it's all fine.

Expected Result
============

Rolled-out beta features should not be displayed on the _Manage Beta Features_.

How to Reproduce
===============
1. Go to [Manage Beta Features](https://build.opensuse.org/my/beta_features)
2. Disable a rolled-out beta feature
3. See how it stays active, even though the switch is now disabled

dmarcoux:
Closes #12322

### DIFF
--- a/src/api/app/controllers/webui/users/beta_features_controller.rb
+++ b/src/api/app/controllers/webui/users/beta_features_controller.rb
@@ -4,6 +4,8 @@ class Webui::Users::BetaFeaturesController < Webui::WebuiController
   def index
     @disabled_beta_features = policy_scope(DisabledBetaFeature).pluck(:name)
     @user = User.session!
+    @beta_features = ENABLED_FEATURE_TOGGLES.reject { |feature_toggle| beta_features_to_reject.include?(feature_toggle[:name].to_s) }
+                                            .sort_by { |feature_toggle| feature_toggle[:name] }
   end
 
   def update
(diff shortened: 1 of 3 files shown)


dmarcoux:
To review this, add the rollout group to a feature in the Flipper UI, then see how it's not displayed on the _Manage Beta Features_ page.

src/api/app/controllers/webui/users/beta_features_controller.rb
```diff
@@ -4,6 +4,10 @@ class Webui::Users::BetaFeaturesController < Webui::WebuiController
   def index
     @disabled_beta_features = policy_scope(DisabledBetaFeature).pluck(:name)
     @user = User.session!
+
+    # Rolled-out beta feature cannot be switched, so we do not display them to users
+    rolled_out_beta_features = Flipper::Adapters::ActiveRecord::Gate.where(key: 'groups', value: 'rollout').pluck(:feature_key)
```
saraycp, line 9: I know in production we usually go through the "rollout way" to enable the feature to everyone... But we could also take into account features that are "Fully Enabled", in case someone goes that way (something that happens frequently in the development environment).

So, apart from rejecting `Flipper::Adapters::ActiveRecord::Gate.where(key: 'groups', value: 'rollout')` you can also reject `Flipper::Adapters::ActiveRecord::Gate.where(key: 'boolean', value: 'true')`.

What do you think?

dmarcoux: Good point, I'll do that.

dmarcoux: It's there now.

saraycp (approved):

dmarcoux:
Code Climate passed, even though it says it couldn't analyze this pull request.

PR #12345, opened 2022-03-24
Fix error trying to find the target package to be deleted

saraycp:
Fix issue #12122

When a PR is closed/merged, we want to remove the temporary project created by the workflow as well as their event subscriptions. These two removals only make sense for steps like `BranchPackageStep` and `LinkPackageStep`. So we prevent processing other steps like `ConfigureRepositories` or `RebuildPackage` (which don't create any temporary project or event subscriptions).

This PR also fixes the EventSuscription removal, since we were passing an incorrect token, so we always got an empty list.

TODO:

- [X] Specs

### DIFF
--- a/src/api/spec/models/workflow_spec.rb
+++ b/src/api/spec/models/workflow_spec.rb
@@ -10,7 +10,10 @@ RSpec.describe Workflow, type: :model, vcr: true do
   end
 
   describe '#call' do
-    let(:yaml) { { 'steps' => [{ 'branch_package' => { 'source_project' => 'test-project', 'source_package' => 'test-package', 'target_project' => 'test-target-project' } }] } }
+    let(:yaml) do
+      { 'steps' => [{ 'branch_package' => { 'source_project' => 'test-project', 'source_package' => 'test-package', 'target_project' => 'test-target-project',
+                                            'target_package' => 'test-target-package' } }] }
+    end
 
     context 'PR was reopened' do
       let(:extractor_payload) do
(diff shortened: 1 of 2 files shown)


danidoni:
[Undefined method start_with? in trigger_workflow#create (M)](https://trello.com/c/u19srond/1861-undefined-method-startwith-in-triggerworkflowcreate-m)

dmarcoux:
I don't know why CodeClimate didn't work here (it's a 500), but for such a small change, I can live without the CodeClimate analysis.

`ci/circleci: checkout_code` obviously passed, otherwise we wouldn't have gotten the other CI checks.

dmarcoux (approved):

src/api/spec/models/workflow_spec.rb
```diff
@@ -43,10 +43,30 @@
...
 
-      before { login user }
-
-      it 'removes the target project' do
-        expect { subject.call }.to change(Project, :count).from(2).to(1)
+      context 'we are dealing with a branch package step' do
+        before { login user }
+
+        it 'removes the target project' do
+          expect { subject.call }.to change(Project, :count).from(2).to(1)
```
dmarcoux, line 50: Would it be possible to also check that `EventSubscription` instances are deleted?

src/api/spec/models/workflow_spec.rb
```diff
@@ -42,11 +45,37 @@
...
+            }
+          ] }
+        end
 
-      before { login user }
+        before { login user }
 
-      it 'removes the target project' do
-        expect { subject.call }.to change(Project, :count).from(2).to(1)
+        it 'does not remove the target project' do
```
dmarcoux, line 76: Could we also check that no event subscriptions are deleted?

danidoni: I was unsure if to add this test or not... I'll add it

danidoni: Added

PR #12351, opened 2022-03-24
Fixing access checks for build log view

adrianschroeter:
Was crashing when package came from remote project.

Also had different behaviour on source_access check, using now same
permission check as with api route.


src/api/app/controllers/webui/package_controller.rb
```diff
@@ -699,25 +699,18 @@ def check_build_log_access
 
     @package_name = params[:package]
     begin
-      @package = Package.get_by_project_and_name(@project, @package_name, use_source: false,
+      @package = Package.get_by_project_and_name(@project, @package_name, use_source: true,
```
dmarcoux, line 702: With this change from `use_source: false` to `use_source: true`, then rescuing from `Package::ReadAccessError` and `Project::ReadAccessError`, `check_build_log_access` is now like `set_project` and `require_package`. 

See:
https://github.com/openSUSE/open-build-service/blob/eeaaddc878c6c575030c6fff53596762465c621b/src/api/app/controllers/webui/webui_controller.rb#L45-L49

https://github.com/openSUSE/open-build-service/blob/eeaaddc878c6c575030c6fff53596762465c621b/src/api/app/controllers/webui/webui_controller.rb#L129-L150

dmarcoux: I pushed changes to follow this.

dmarcoux:
@adrianschroeter What do you think of this?

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -12,12 +12,12 @@ class Webui::PackageController < Webui::WebuiController
...
 
   before_action :set_project, only: [:show, :edit, :update, :index, :users, :dependency, :binary, :binaries, :requests, :statistics, :revisions,
-                                     :new, :branch_diff_info, :rdiff, :create, :save, :remove,
+                                     :new, :branch_diff_info, :rdiff, :create, :save, :remove, :live_build_log, :update_build_log,
                                      :remove_file, :save_person, :save_group, :remove_role, :view_file, :abort_build, :trigger_rebuild,
                                      :trigger_services, :wipe_binaries, :buildresult, :rpmlint_result, :rpmlint_log, :meta, :save_meta, :files]
 
   before_action :require_package, only: [:edit, :update, :show, :dependency, :binary, :binaries, :requests, :statistics, :revisions,
-                                         :branch_diff_info, :rdiff, :save, :save_meta, :remove,
+                                         :branch_diff_info, :rdiff, :save, :save_meta, :remove, :live_build_log, :update_build_log,
```
adrianschroeter, line 20: requiring a package for build log would bring back the cash for remote packages again, no?

dmarcoux: `require_package` fetches the package with the same options for `Package.get_by_project_and_name` as what you pushed before. Previously, you switched the `use_source` option from `false` to `true`, but this option has a default value of `true`, see: https://github.com/openSUSE/open-build-service/blob/eeaaddc878c6c575030c6fff53596762465c621b/src/api/app/models/package.rb#L155-L157

The other options (`follow_project_links: true, follow_multibuild: true`) are the same in `require_package`. As for rescuing from `Package::ReadAccessError` and `Project::ReadAccessError`, those are already caught in `require_package` since both errors have the ancestor `APIError`:

https://github.com/openSUSE/open-build-service/blob/eeaaddc878c6c575030c6fff53596762465c621b/src/api/app/controllers/webui/webui_controller.rb#L139

```bash
# inside Rails console
Package::ReadAccessError.ancestors.include?(APIError)
=> true
Project::ReadAccessError.ancestors.include?(APIError)
=> true
```

I will test this further, but I think it's going to work.

adrianschroeter: well, yes, default value is true, but setting it to false is wrong (you may of course also remove the extra parameter)

adrianschroeter:
I had to redo this submission, since current code change a lot, but also blocked people in production in various situations.

This also includes fixes for security regressions, so this needs to be merged asap.

In general the spider test showed most likely some of the crashes, I removed therefore therefore the check to skip many UI taps entirely when spider test is running. Most likely caused by remote project situation?

adrianschroeter:
@hvogel: you pushed a very old version again here? I suppose that was not intentional?

hennevogel:
> you pushed a very old version again here? 

Nope, I just fixed the haml linter complaints.

adrianschroeter:
hm, github showed me just a single file in the files changed tab, but it seems to work again.
btw this is patch 145 in internal package and test suite succeeds there.

hennevogel:
@adrianschroeter can you maybe setup a project/package that shows what you are trying to fix with this? Like in [OBS:test](https://build-test.opensuse.org/project/show/OBS:test) on build-test or something?

adrianschroeter:
On Dienstag, 19. September 2023, 16:08:03 CEST Henne Vogelsang wrote:
> @adrianschroeter can you maybe setup a project/package that shows what you are trying to fix with this? Like in [OBS:test](https://build-test.opensuse.org/project/show/OBS:test) on build-test or something?

You could create a test browsing proejcts and accessing build logs
using the fixtures. Just access it via RemoteInstance project.

I can easily add such test case for the minitest suite, if you want.

But any browsing of eg. RemoteInstance:BaseDistro* project and access of
a build log leaded to crashes. 

In case of scmsync projects no remote project is needed. It got either
blocked or crashed in case of (local or remote) project link as well.

-- 

Adrian Schroeter ***@***.***>

hennevogel:
And even if you build the URL yourself somewhere like 

https://build-test.opensuse.org/package/live_build_log/openSUSE.org:openSUSE:Factory/aaa_base/standard/x86_64

it will not crash...

adrianschroeter:
On Dienstag, 19. September 2023, 16:59:32 CEST Henne Vogelsang wrote:
> `RemoteInstance` is a is just a project with an `remoteurl`. Like [openSUSE.org](https://build-test.opensuse.org/project/show/openSUSE.org) on build-test. How do you event get to a package in the UI there. You can't, it just throws a 404.
> 
> So this can't be it...

when you have a local build reusing sources from remote you need to be able to check your
local log files.

This regression is quite a blocker not only for all our internal instances, but
also for external partner instances.

The scmsync part is analog. Escept that it is also broken on a local instance now.

-- 

Adrian Schroeter ***@***.***>

adrianschroeter:
One situation is setup here: https://build-test.opensuse.org/package/live_build_log/home:adrianSuSE:Tools/cc-tool/15.3/x86_64

but there are of course also other ways to trigger this.

hennevogel:
> when you have a local build reusing sources from remote you need to be able to check your
local log files.

Alright so you mean `linkedbuild` of projects with a remote project link. But you know that nearly *every* package action about this is broken right? Starting from `Webui::PackageController#show`...

hennevogel:
are there other setups you know where this can happen? Asking for a friend who wants to write a spec... 😉

adrianschroeter:
On Mittwoch, 20. September 2023, 14:07:24 CEST Henne Vogelsang wrote:
> are there other setups you know where this can happen? Asking for a friend who wants to write a spec... 😉 

eg. a single project with scmsync (currently just blocked, which was the escalation reason)

or a local project point to another local project using scmsync (will crash in the same way).

apart from that the missing permission check is worth a test case.
-- 

Adrian Schroeter ***@***.***>

adrianschroeter:
rebased again and add the crash fix of the dangerous part in an extra commit now.

adrianschroeter:
Rebased it again to solve conflicts.

This needs to solved this week as it blocks Leap16 development now!

hennevogel:
done slightly differently in #16327

adrianschroeter:
the broken permission handling is not solved in #16327

src/api/app/controllers/build_controller.rb
```diff
@@ -136,7 +136,7 @@ def builddepinfo
 
   def logfile
     # for permission check
-    pkg = Package.get_by_project_and_name(params[:project], params[:package], follow_multibuild: true)
+    pkg = Package.get_by_project_and_name(params[:project], params[:package], follow_multibuild: true, use_source: true)
```
hennevogel, line 139: `use_source: true` is the default...

src/api/app/controllers/webui/packages/build_log_controller.rb
```diff
@@ -105,13 +105,17 @@ def check_build_log_access
         @package_name = params[:package]
 
         # No need to check for the package, they only exist on the backend in this case
+        # FIXME: scmsync is not the criteria, should also be done for remote packages
         if @project.scmsync
-          @can_modify = User.possibly_nobody.can_modify?(@project)
```
hennevogel, line 109: `@can_modify` is used in the views to decide if we offer rebuild/abort build buttons.

src/api/app/controllers/webui/packages/build_log_controller.rb
```diff
@@ -105,13 +105,17 @@ def check_build_log_access
...
-          return
+          return if User.possibly_nobody.can_source_access?(@project)
```
hennevogel, line 110: `source_access` permission is bound to the reader role. I don't think you want people to have the reader role to read build logs right?

What is it you are worried about?

Projects that are access protected? If you don't have access you are not going to reach this point as you'll be redirected in line 101...

Projects that are sourceaccess protected? Why would you not be allowed to read the build log for them?

src/api/app/controllers/webui/packages/build_log_controller.rb
```diff
@@ -105,13 +105,17 @@ def check_build_log_access
...
+
+          redirect_to package_show_path(project: @project.name, package: @package_name),
+                      error: 'Could not access build log'
+          return false
         end
 
         begin
-          @package = Package.get_by_project_and_name(@project, @package_name, use_source: false,
+          @package = Package.get_by_project_and_name(@project, @package_name, use_source: true,
```
hennevogel, line 118: Same here, `use_source: true` is the default.

src/api/app/controllers/build_controller.rb
hennevogel, line 139: `use_source: true` is the default...

hennevogel (changes requested):

adrianschroeter:
On Donnerstag, 20. Juni 2024, 15:47:50 CEST Henne Vogelsang wrote:
> 
> 
> 
> >          if @project.scmsync
> -          @can_modify = ***@***.***)
> 
> ***@***.***_modify` is used in the views to decide if we offer rebuild/abort build buttons.

okay, we can have it in addition.

> >          if @project.scmsync
> -          @can_modify = ***@***.***)
> -          return
> +          return if ***@***.***)
> 
> `source_access` permission is bound to the reader role. I don't think you want people to have the reader role to read build logs right?
>
> What is it you are worried about?

that source access is forbidden and you can still access the build log.

> Projects that are access protected? If you don't have access you are not going to reach this point as you'll be redirected in line 101...

there is also sourceaccess beside acccess.

> Projects that are sourceaccess protected? Why would you not be allowed to read the build log for them?


A user might have access rights, but no sourceaccess rights.

> >          end
>  
>          begin
> -          @package = ***@***.***, @package_name, use_source: false,
> +          @package = ***@***.***, @package_name, use_source: true,
> 
> Same here, `use_source: true` is the default.
> 
> > @@ -136,7 +136,7 @@ def builddepinfo
>  
>    def logfile
>      # for permission check
> -    pkg = Package.get_by_project_and_name(params[:project], params[:package], follow_multibuild: true)
> +    pkg = Package.get_by_project_and_name(params[:project], params[:package], follow_multibuild: true, use_source: true)
> 
> `use_source: true` is the default...
> 
> 


-- 

Adrian Schroeter ***@***.***>

adrianschroeter:
I also tried to fix the remote project situation as this would be a regression to our deployed patched/fixed internal instances otherwise.

adrianschroeter:
| Projects that are sourceaccess protected? Why would you not be allowed to read the build log for them?

to be explicit, this is the policy as the build log might expose the source part which is critical (as security fixes).

And it is also implemented that way in most places, just not here in the webui.

PR #12354, opened 2022-03-24
Clean up running workflows with closed PRs

saraycp:
After many occurrences of the error that we are trying to fix with PR #12345, some workflow runs stayed in 'running' status forever and the temporary projects created by the workflow were not removed as expected even if the corresponding pull request or merge request was closed/merged.

The workflow list looks like this in those cases (the workflow run status is `running` and the PR status is `closed`):
Screenshot 2022-03-25 at 18-10-50 Open Build Service

Apart from fixing the errors or catching them correctly, we need to ensure we clean up those projects. This rake task is introduced for that purpose.

~**NOTE:** PR #12345 needs to be merged first before we run this task in production, so we make sure we remove all the cases that happened before the fix.~

**NOTE 2:** Now that some methods are moved from the component to the model, [WorkflowRunHeaderComponent](https://github.com/openSUSE/open-build-service/commit/3e0d02314ff1377e97f4a7cdfa107b07f7e89dcd#diff-61beba557cf7e6990b08e998aa7db01d051be1933b9f5e126a0553c44939243cR1) probably doesn't need to inherit from WorkflowRunRowComponent but simply call the methods of the model. Something to be done in another PR, [this is the card](https://trello.com/c/z6d7XE2P/2566-remove-inheritance-between-components).

**NOTE 3:** Once this PR is merged, we'll need to manually run this rake task in production.

### DIFF
--- a/src/api/lib/tasks/workflows.rake
+++ b/src/api/lib/tasks/workflows.rake
@@ -1,0 +1,33 @@
+namespace :workflows do
+  desc 'Remove projects that were not closed as expected and set workflow run status to running'
+  task cleanup_non_closed_projects: :environment do
+    workflow_runs = WorkflowRun.where(status: 'running')
+                               .select do |workflow_run|
+                                 workflow_run.hook_event.in?(['pull_request', 'Merge Request Hook']) &&
+                                   workflow_run.hook_action.in?(['closed', 'close', 'merge'])
+                               end
+
+    puts "There are #{workflow_runs.count} workflow runs affected"
+
+    workflow_runs.each do |workflow_run|
+      projects = Project.where('name LIKE ?', "%#{target_project_name_postfix(workflow_run)}")
+
+      # If there is more than one project, we don't know which of them is the one related to the current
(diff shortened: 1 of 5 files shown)


coolo:
I assumed PR 12345 was just an example number :)

saraycp:
> I assumed PR 12345 was just an example number :)

Just a nice coincidence :D

src/api/lib/tasks/workflows.rake
```diff
@@ -0,0 +1,33 @@
+namespace :workflows do
```
rubhanazeem, line 1: The purpose of this rake task is to only clean the data, right? After this PR is merged do we still need this task or the workflow_runs will change their state as expected? :slightly_smiling_face:

saraycp: Once this PR is merged, we'll need to manually run this rake task in production.
After that, all the workflow_runs affected by the same failure will change their status.

The task can be useful in case new similar failures happen again in the future.

rubhanazeem (approved):

PR #12358, opened 2022-03-25
[backend] add "--filelists_ext" parameter for createrepo

aplanas:
If "filelists_ext" option is enabled for the rpmmd repository, it will
call createrepo_c with the --filelists_ext parameter.

This parameter will create a new optional variant of filelists.xml,
named filelists_ext.xml.  This will contain, besides the list of files,
the hash (usually SHA256) registered in the RPM header for this file
(only for file types, not for ghost or dir types).

Additionally, a new field "checksum" will register the hash algorithm
used by librpm for those file entries.

Signed-off-by: Alberto Planas <aplanas@suse.com>

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -617,6 +617,7 @@ sub createrepo_rpmmd {
   push @createrepoargs, '--content', 'debug' if $data->{'dbgsplit'};
   push @createrepoargs, '--content', 'update' if ($repoinfo->{'projectkind'} || '') eq 'maintenance_incident';
   push @createrepoargs, '--content', 'update' if ($repoinfo->{'projectkind'} || '') eq 'maintenance_release';
+  push @createrepoargs, '--filelists_ext' if $options{'filelists_ext'};
   # createrepo_c 0.20.0 defaults to keep always additional metadata. We must change
   # to old behaviour or old data would stay forever when it got removed actually
   push @createrepoargs, '--discard-additional-metadata';


aplanas:
This is an alternative for https://github.com/openSUSE/open-build-service/pull/12296, using the new `--hash` parameter drafted in `createrepo_c` (https://github.com/rpm-software-management/createrepo_c/pull/316)

I am not sure where this `$options->{'hash'}` can be set, as I copy it from `$options{'zchunk'}` and I do not see set it with a default value in the code.

adrianschroeter:
can you please prefix your commit with   [backend]   ?

adrianschroeter:
the minitest failures seem to be cause by failing createrepo runs, will need to check why later today.

adrianschroeter:
> the minitest failures seem to be cause by failing createrepo runs, will need to check why later today.

The failures seem to be cause by the broken createrepo_c, independend of this cange.

 @M0ses is that already on your radar?

aplanas:
> can you please prefix your commit with [backend] ?

Sure. Done.

adrianschroeter:
and please write it using  defined():

if defined($options->{'filelists_ext'});

it seems to be the cause for the failures.

aplanas:
> and please write it using defined():

Done!

aplanas:
Oh sorry, I see now my bug : ((

adrianschroeter:
On Dienstag, 5. April 2022, 08:01:33 CEST Alberto Planas wrote:
> Oh sorry, I see now my bug : ((

credits are going to @M0ses for spotting it btw :)

-- 

Adrian Schroeter ***@***.***>

M0ses (approved):
LGTM

PR #12362, opened 2022-03-29
Add trigger service step

krauselukas:
This PR implement a step that enables users to trigger services through their workflows.

*TODO*

- [x] Write and adapt specs
- [x] Tackle Rubocop offenses 
- [x] Testing with a real setup

### DIFF
--- a/src/api/app/controllers/concerns/triggerable.rb
+++ b/src/api/app/controllers/concerns/triggerable.rb
@@ -11,8 +11,7 @@ module Triggerable
     raise Project::Errors::UnknownObjectError, "Sorry, triggering tokens for remote project \"#{@project_name}\" is not possible." unless @project.is_a?(Project)
   end
 
-  # rubocop:disable Naming/AccessorMethodName
-  def set_package(package_find_options: {})
+  def set_package(package_find_options: {}) # rubocop:disable Naming/AccessorMethodName
     package_find_options = @token.package_find_options if package_find_options.blank?
     # By default we operate on the package association
     @package = @token.package
--- a/src/api/app/controllers/concerns/triggerable.rb
+++ b/src/api/app/controllers/concerns/triggerable.rb
@@ -11,13 +11,15 @@ module Triggerable
     raise Project::Errors::UnknownObjectError, "Sorry, triggering tokens for remote project \"#{@project_name}\" is not possible." unless @project.is_a?(Project)
   end
 
-  def set_package
+  # rubocop:disable Naming/AccessorMethodName
+  def set_package(package_find_options: {})
+    package_find_options = @token.package_find_options if package_find_options.blank?
     # By default we operate on the package association
     @package = @token.package
     # If the token has no package, let's find one from the parameters
     @package ||= Package.get_by_project_and_name(@project,
                                                  @package_name,
-                                                 @token.package_find_options)
+                                                 package_find_options)
     return unless @project.links_to_remote?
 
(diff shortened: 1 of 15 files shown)


src/api/app/models/token/workflow.rb
```diff
@@ -31,11 +31,6 @@ def call(options)
     raise Token::Errors::SCMTokenInvalid, e.message
   end
 
-  # Only used by rebuild steps
-  def package_find_options
-    { use_source: false, follow_project_links: true, follow_multibuild: true }
```
hennevogel, line 36: You noticed that they differ per token type right?

krauselukas: Yes, thats why I moved it from the workflow token model to the individual steps, in order to pass the correct options for the Triggerable concern

src/api/app/controllers/concerns/triggerable.rb
```diff
@@ -27,6 +28,7 @@ def set_package
     # TODO: This should not happen right? But who knows...
     raise ActiveRecord::RecordNotFound unless @package
   end
+  # rubocop:enable Naming/AccessorMethodName
```
dmarcoux, line 31: This comment could be right under `def set_package(package_find_options:)`.

krauselukas: Done

saraycp:
I've done a few manual tests locally and worked fine. I'd like to test more, but for now, I've detected a few typos in commit descriptions:

First commit:
- replace "_Since we going_" with "_Since we are going_"
- replace  "_we going to have more then_" with "_we are going to have more than_"

Second commit:
- replace "_inorder_" with "_in order_"

I also wonder if it makes sense to rescue from possible exceptions raised by the [backend call](https://github.com/openSUSE/open-build-service/pull/12362/files#diff-0640a1ff102c6e55ae84ad3a8b8f55ebd971c99a11a338c90a785263f24a6ebbR18).
In my case I got one for not having the service correctly set up:

```
Backend::NotFoundError: <status code="404">
  <summary>no source service defined!</summary>
  <details>404 no source service defined!</details>
</status>
```
 I know we are not doing so for the [RebuildPackage step](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/workflow/step/rebuild_package.rb#L36), but we rescue from other exceptions in steps like [BranchPackageStep](https://github.com/openSUSE/open-build-service/blob/96148b1a43e9c8d854d37695f04deda03cb04919/src/api/app/models/workflow/step/branch_package_step.rb#L51).

src/api/app/controllers/concerns/triggerable.rb
```diff
@@ -11,13 +11,16 @@ def set_project
     raise Project::Errors::UnknownObjectError, "Sorry, triggering tokens for remote project \"#{@project_name}\" is not possible." unless @project.is_a?(Project)
   end
 
-  def set_package
+  # rubocop:disable Naming/AccessorMethodName
+  def set_package(package_find_options: {})
+    # rubocop:enable Naming/AccessorMethodName
```
danidoni, line 16:
```suggestion
  def set_package(package_find_options: {}) # rubocop:disable Naming/AccessorMethodName
```

I think this works, too...

krauselukas: done @danidoni

danidoni (approved):

saraycp (approved):

PR #12364, opened 2022-03-29
Get consitent repository name and URL for workflow runs

saraycp:
The workflow run details contain a link to the SCM's repository. Compare these two screenshots of workflow runs with real incoming payloads:

Screenshot from 2022-03-28 19-06-32

Screenshot from 2022-03-28 19-06-49

For GitHub-related workflow runs, we are displaying repository names like `foo/bar`.
For GitLab-related workflow runs, we were displaying repository names like `bar`. For consistency, I have decided to get the information from another payload field, so we can also have `foo/bar`.

Apart from the names, if we compare the links (at the bottom of the screenshots), we see that GitLab-related workflow runs used to point to a wrong URL because we were getting the information from a payload field that could contain `git@gitlab.com:foo/bar.git` instead of an HTTP(s) URL. So I'm now using another payload field to get the correct URL.

~**NOTE:** I'll need to adapt this PR after PR #12354 is merged as part of this code is refactored there.~

### DIFF
--- a/src/api/app/models/workflow_run.rb
+++ b/src/api/app/models/workflow_run.rb
@@ -48,13 +48,13 @@ class WorkflowRun < ApplicationRecord
   end
 
   def repository_name
-    payload.dig('repository', 'full_name') || # For GitHub
-      payload.dig('repository', 'name') # For GitLab
+    payload.dig('repository', 'full_name') || # For GitHub on pull_request and push events
+      payload.dig('project', 'path_with_namespace') # For GitLab on merge request and push events
   end
 
   def repository_url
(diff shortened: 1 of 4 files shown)


dmarcoux:
The changes match the documentation on webhook events from GitLab: https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html

dmarcoux (approved):

dmarcoux:
What changed?

saraycp:
> What changed?

I simply rebase on master. There were no changes in the PR.

PR #12365, opened 2022-03-29
Create migration to create a new workflow payload column

danidoni:
This PR creates a new column with the proper type to store JSON documents up to 4GB.

To do this [without blocking the table](https://github.com/ankane/strong_migrations#changing-the-type-of-a-column) we have to:
1. Create the new column
2. Write to both columns
3. Backfill data from the old column to the new column

TODO:
- [x] Write a test to check the data migration (the backfilling)

And deploy this before the next changes:

Done in another #12375:
1. Move reads from the old column to the new column
2. Stop writing to the old column

Done in the last PR:
1. Drop the old column

### DIFF
--- a/src/api/app/controllers/trigger_workflow_controller.rb
+++ b/src/api/app/controllers/trigger_workflow_controller.rb
@@ -71,7 +71,10 @@ class TriggerWorkflowController < TriggerController
     raise Trigger::Errors::InvalidToken, 'Wrong token type. Please use workflow tokens only.' unless @token.is_a?(Token::Workflow)
 
     request_headers = request.headers.to_h.keys.map { |k| "#{k}: #{request.headers[k]}" if k.match?(/^HTTP_/) }.compact.join("\n")
-    @workflow_run = @token.workflow_runs.create(request_headers: request_headers, request_payload: request.body.read)
+    # We need to write the payload into `request_json_payload` too, while the `request_payload` is still around
+    @workflow_run = @token.workflow_runs.create(request_headers: request_headers,
+                                                request_payload: request.body.read,
+                                                request_json_payload: request.body.read)
   end
 
   def abort_trigger_if_ignored_pull_request_action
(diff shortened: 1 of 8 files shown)


saraycp:
> Done in the last PR:
> 
>     1. Drop the old column

2. Rename `new_request_payload` to `request_payload`.


I think there is a missing step in the description, after dropping the old column.

danidoni:
> > Done in the last PR:
> > ```
> > 1. Drop the old column
> > ```
> 
>     2. Rename `new_request_payload` to `request_payload`.
> 
> 
> I think there is a missing step in the description, after dropping the old column.

Mmm not really, my plan was not to rename back the new column to the old name, but to keep the new name. I've improved the column name, the previous name was bad.

danidoni:
[Data too long for column request_payload (M)](https://trello.com/c/oaqQLSbO/1862-data-too-long-for-column-requestpayload-m)

src/api/app/controllers/trigger_workflow_controller.rb
```diff
@@ -71,7 +71,10 @@ def create_workflow_run
     raise Trigger::Errors::InvalidToken, 'Wrong token type. Please use workflow tokens only.' unless @token.is_a?(Token::Workflow)
 
     request_headers = request.headers.to_h.keys.map { |k| "#{k}: #{request.headers[k]}" if k.match?(/^HTTP_/) }.compact.join("\n")
-    @workflow_run = @token.workflow_runs.create(request_headers: request_headers, request_payload: request.body.read)
+    # We need to write the payload into `request_json_payload` too, while the `request_payload` is still around
+    @workflow_run = @token.workflow_runs.create(request_headers: request_headers,
+                                                request_payload: request.body.read,
+                                                request_json_payload: request.body.read)
```
dmarcoux, line 77: When this PR is deployed, but before the migration adding `request_json_payload` is executed, there will be errors happening here. To prevent those errors, you'll have to split this PR. The first PR will have the migration. The second PR will have this code and the data migration to backfill data.

danidoni:
#12376 supersedes this PR

PR #12375, opened 2022-03-30
Stop writing to the old workflow runs request payload column

danidoni:
This PR is a follow-up of #12365.

**NOTE**: That PR has to be already deployed before this one hits production.

This PR shifts the reads from the old `request_payload` to `request_json_payload` so we can stop using that column.

TODO:
- [ ] Deploy #12365


danidoni:
#12376 supersedes this PR

PR #12398, opened 2022-04-04
Make use of the trigger_services workflow step

hennevogel:
Keep pouring


hennevogel:
coveralls hangs, see codecov report.

dmarcoux:
From OBS:Server:Unstable/obs-server:

trigger

PR #12403, opened 2022-04-04
Use 2.10 containers in CI

hennevogel:
After OSU switched to Ruby 3.1 we need to use 2.10 containers.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,21 +7,25 @@ images:
       NOKOGIRI_USE_SYSTEM_LIBRARIES: 1
 
   - &mariadb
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -196,13 +200,11 @@ jobs:
...
     docker:
-      - <<: *frontend_backend
+      - <<: *frontend_base
       - <<: *mariadb
     steps:
       - attach_workspace:
-         at: .
-      - run: *install_dependencies
-      - run: *init_git_submodule
+         at: .frontend_ba
```
saraycp, line 207: Is this correct?

hennevogel: it was not. removed

dmarcoux (approved):

PR #12407, opened 2022-04-05
Reset User.session

hennevogel:
Assigning User.session creates a thread variable. Thread
variables are not reset in delayed jobs. So whoever ran
the first job also ran all the jobs after it...

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -1122,22 +1122,22 @@ class Project < ApplicationRecord
 
   # called either directly or from delayed job
   def do_project_release(params)
-    User.session ||= User.find_by!(login: params[:user])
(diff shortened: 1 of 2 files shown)


hennevogel:
depends on #12403

danidoni (approved):

src/api/app/models/project.rb
```diff
@@ -1122,22 +1122,23 @@ def all_sources_changed
 
   # called either directly or from delayed job
   def do_project_release(params)
-    User.session ||= User.find_by!(login: params[:user])
-
-    # uniq timestring for all targets
-    time_now = Time.now.utc
+    User.find_by!(login: params[:user]).run_as do
```
dmarcoux, line 1126: To fix the issue raised by the RuboCop:

```suggestion
```

PR #12411, opened 2022-04-06
Introduce tests for backend workers code

eduardoj:
Add the following tests for the `getsources` subroutine:
- Return result.
- Add 'materials' key to $buildinfo variable (used in the SLSA provenance generation).

~The Test::Mock::BSRPC module was modified to handle requests which return a cpio file.~


src/backend/t/1000-bs_worker.t
```diff
@@ -0,0 +1,51 @@
...
+@::ARGV = ('--testcase');
+require_ok('./bs_worker');
+
+$BSConfig::bsdir = "$FindBin::Bin/data/1000";
+
+my $buildinfo = {
+  srcserver  => $BSConfig::srcserver,
+  project    => 'project1',
+  package    => 'package1',
+  srcmd5     => '1234123412341234',
```
mlschroe, line 25: The srcmd5 should be the same as the verifymd5 here.

src/backend/t/1000-bs_worker.t
```diff
@@ -0,0 +1,51 @@
...
+use lib "$FindBin::Bin/lib/";
+
+use Test::Mock::BSConfig;
+use Test::Mock::BSRPC;
+
+
+@::ARGV = ('--testcase');
+require_ok('./bs_worker');
+
+$BSConfig::bsdir = "$FindBin::Bin/data/1000";
```
mlschroe, line 19: Mybe add
```
my $tmpdir = " $FindBin::Bin/tmp/1000";
BSUtil::rm_rf($tmpdir);
```

src/backend/t/1000-bs_worker.t
```diff
@@ -0,0 +1,51 @@
...
+  "srcserver/getsources?project=project1&package=package1&srcmd5=$buildinfo->{srcmd5}"
+    => 'srcserver/getsources',
+};
+
+my (@got, @expected);
+
+
+# getsources
+
+@got = getsources($buildinfo, $BSConfig::bsdir.'/bs_worker');
```
mlschroe, line 39: Use $tmpdir instead of $BSConfig::bsdir/bs_worker?

src/backend/t/data/1000/bs_worker/hello_world.spec
```diff
@@ -0,0 +1,27 @@
+Name:       hello_world
```
mlschroe, line 1: Please remove the data/1000/bs_worker directory if you use a tmpdir

src/backend/t/lib/Test/Mock/BSRPC.pm
```diff
@@ -57,7 +57,12 @@ BEGIN {
 
       my $receiver = $param->{'receiver'};
       if ($receiver) {
-	$ret = $receiver->(BSHTTP::str2req($ret), $param, $xmlargs || $param->{'receiverarg'}) if $receiver;
+        if ($receiver == \&BSHTTP::file_receiver) {
+          $ret = $receiver->(BSHTTP::str2req($ret), $param, $xmlargs || $param->{'receiverarg'});
+        }
+        elsif ($receiver == \&BSHTTP::cpio_receiver) {
+          $ret = $receiver->({ '__data' => $ret }, $param, $xmlargs || $param->{'receiverarg'});
+        }
```
mlschroe, line 65: I don't get why this is needed. BSHTTP::str2req does exactly what you do here, so why doesn't the original code work?

src/backend/t/1000-bs_worker.t
eduardoj: I modified `$BSConfig::bsdir` to be `$FindBin::Bin`. Otherwise we have problems with extracting the files in other subdirectories.

Now a `$tmpdir` is used, created at the beginning of the tests, and removed afterwards.

Keeping a subdirectory with git wouldn't work. It will be needed an empty `.keep` file to store it in git, and it would be recognized as a spurious file.

src/backend/t/1000-bs_worker.t
eduardoj: Done.

src/backend/t/lib/Test/Mock/BSRPC.pm
eduardoj: True. :man_facepalming: I reverted this change.

eduardoj:
CircleCI had problems today. Anyway, backend tests are run thanks to the workflow configuration for testing the builds for 15.3. :tada: An excerpt of the [build log](https://build.opensuse.org/package/live_build_log/home:bs-team:openSUSE:open-build-service:PR-12411/obs-server/15.3/x86_64):

```
[...]
[  419s] ok 2 - Checking download url
[  419s] ok
[  420s] t/1000-bs_worker.t .......................... 
[  420s] 1..3
[  420s] ok 1 - require './bs_worker';
[  420s] ok 2 - getsources - Return value
[  420s] ok 3 - getsources - Add 'materials' key to $buildinfo
[  420s] ok
[  420s] All tests successful.
[...]
```

mlschroe (approved):

PR #12414, opened 2022-04-06
Instrument more exceptions

hennevogel:
So far we have instrumentation for exceptions that happen in ActionController.
There are many more exceptions occuring outside of this component as you can
see on errbit. Let's instrument those too.

### DIFF
--- a/src/api/config/initializers/airbrake.rb
+++ b/src/api/config/initializers/airbrake.rb
@@ -107,6 +107,9 @@ def ignore_by_class?(notice)
 end
 
 def ignore_exception?(notice)
+  notice[:errors].pluck(:type).each do |type|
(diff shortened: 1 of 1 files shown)


src/api/config/initializers/airbrake.rb
```diff
@@ -107,6 +107,9 @@ def ignore_by_class?(notice)
 end
 
 def ignore_exception?(notice)
+  notice[:errors].pluck(:type).each do |type|
```
dmarcoux, line 110: It's a bit weird to have this code here. I suggest putting it in the block of `Airbrake.add_filter`, right above `notice.ignore! if ignore_exception?(notice)`.

vpereira (approved):

PR #12417, opened 2022-04-07
Add Workflow Runs filter by pull/merge request action

rubhanazeem:
Currently, it is not possible to filter workflow runs by action type i.e., we can not filter workflow runs for `pull_requests/merge requests` by actions like `opened`, `closed`, etc. This PR allows us to filter `pull_requests/merge requests` by actions.

filter option

### DIFF
--- a/src/api/app/components/workflow_run_request_action_filter_component.html.haml
+++ b/src/api/app/components/workflow_run_request_action_filter_component.html.haml
@@ -1,0 +1,12 @@
+= form_with(url: token_workflow_runs_path(@token_id), method: :get, local: true, class: 'form-inline') do
+  = hidden_field_tag('generic_event_type', @generic_event_type)
+  = label_tag('request_action', 'Action', class: 'sr-only')
+  .input-group.mr-sm-2
+    .input-group-prepend
+      .input-group-text
+        Action
+    = select_tag('request_action',
+                  options_for_select(@filter_options, @request_action),
+                  class: 'form-control custom-select')
+  %button.btn.btn-primary#filter-button{ type: :submit }
+    Apply
--- a/src/api/app/queries/workflow_runs_finder.rb
+++ b/src/api/app/queries/workflow_runs_finder.rb
@@ -22,12 +22,17 @@ class WorkflowRunsFinder
     end
   end
 
-  def with_generic_event_type(generic_event_type)
+  def with_generic_event_type(generic_event_type, request_action = nil)
     query = find_real_event_types(generic_event_type).map do |real_event_type|
       "request_headers LIKE '%#{real_event_type}%'"
     end.join(' OR ')
 
-    @relation.where(query)
+    workflow_runs = @relation.where(query)
+    if request_action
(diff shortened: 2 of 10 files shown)


src/api/app/views/webui/workflow_runs/index.html.haml
```diff
@@ -16,9 +16,12 @@
           %p There are no workflow runs for this token yet
       - else
         .card-body
+          - if @selected_filter[:event_type] == 'pull_request'
+            = render WorkflowRunRequestActionFilterComponent.new(token_id: @token.id)
           .text-center
             %span.ml-3= page_entries_info(@workflow_runs)
           .accordion.pt-3#workflow-runs-accordion
+            =# render(partial: 'workflow_runs_accordion', locals: { workflow_runs: @workflow_runs })
```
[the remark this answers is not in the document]
rubhanazeem: Ah it's just WIP changes

src/api/app/queries/workflow_runs_finder.rb
```diff
@@ -13,7 +13,7 @@ def group_by_event_type
     end
   end
 
-  def with_event_type(event_type, action=nil)
+  def with_event_type(event_type, action = nil)
```
danidoni, line 16:
```suggestion
  def with_event_type(event_type, request_action = nil)
```
To be consistent with the rest of the changes

src/api/app/components/workflow_run_request_action_filter_component.html.haml
```diff
@@ -0,0 +1,12 @@
+= form_with(url: token_workflow_runs_path(@token_id), method: :get, local: true, class: 'form-inline') do
+  = label_tag('request_action', 'Action', class: 'sr-only')
+  .input-group.mr-sm-2
+    .input-group-prepend
+      .input-group-text
+        Action
+    = select_tag('request_action',
```
krauselukas, line 7: @rubhanazeem its a small thing, but the select box doesn't have the round edges on the right side 
image
at other places we have it 
image

krauselukas: would be cool to submit the form on select, since you can only choose one filter there is no need to first select and then submit through the `Apply` button. But that could be a follow up as well

krauselukas:
Another thing I just realized, if you filter by a action where no workflow runs exist you cannot select another filter anymore 
image

maybe it makes sense to show the select box here as well

src/api/app/components/workflow_run_request_action_filter_component.html.haml
rubhanazeem: > would be cool to submit the form on select, since you can only choose one filter there is no need to first select and then submit through the `Apply` button. But that could be a follow up as well

I know we can have the form submitted on change but that's not going to be a clean solution because we also have pagination on the same page. Pagination doesn't update automatically with js request. That's why I moved to this solution.

krauselukas (approved):

PR #12418, opened 2022-04-07
Rework token name

vpereira:
Instead of have this class method manually defined in each `Token` class, have it in the parent class in a generic way.

Beside that we standardized the way to access it in the views

From here we have some options:

a) Rename the method to something more useful
b) Move the method `token_name` to the component or transform it in a helper

How to test:

Try to create a new token or edit an existing one

### DIFF
--- a/src/api/app/models/token.rb
+++ b/src/api/app/models/token.rb
@@ -23,6 +23,10 @@ class Token < ApplicationRecord
     self.class.token_name
   end
 
+  def self.token_name
+    name.demodulize.downcase
+  end
+
   def self.token_type(action)
     case action
     when 'rebuild'
--- a/src/api/app/models/token.rb
+++ b/src/api/app/models/token.rb
@@ -20,11 +20,11 @@ class Token < ApplicationRecord
   OPERATIONS = ['Rebuild', 'Release', 'Service', 'Workflow'].freeze
 
   def token_name
-    self.class.token_name
+    self.class.token_name.downcase
   end
 
   def self.token_name
-    name.demodulize.downcase
+    name.demodulize
   end
 
   def self.token_type(action)
--- a/src/api/app/views/webui/users/token_triggers/show.html.haml
+++ b/src/api/app/views/webui/users/token_triggers/show.html.haml
@@ -13,7 +13,7 @@
             = @token.id
           .form-group
             = f.label(:type, 'Operation:')
-            = @token.class.token_name.capitalize
+            = @token.token_name.capitalize
           - token_package = @token.package
           - if token_package.present?
             = fields_for(:package, token_package) do |token_fields|
--- a/src/api/app/views/webui/users/tokens/edit.html.haml
+++ b/src/api/app/views/webui/users/tokens/edit.html.haml
@@ -22,7 +22,7 @@
                 = @token.id
               .form-group
                 = f.label(:type, 'Operation:')
-                = @token.class.token_name.capitalize
+                = @token.token_name.capitalize
               .form-group
                 = f.label(:name, 'Name:')
                 %span.d-none#original-token-name
--- a/src/api/app/components/token_card_component.rb
+++ b/src/api/app/components/token_card_component.rb
@@ -8,7 +8,7 @@ class TokenCardComponent < ApplicationComponent
   end
 
   def operation
-    "Operation: #{@token.class.token_name.capitalize}"
+    "Operation: #{@token.token_name.capitalize}"
   end
 
   def token_package_link
--- a/src/api/app/models/token/rebuild.rb
+++ b/src/api/app/models/token/rebuild.rb
@@ -1,8 +1,4 @@
 class Token::Rebuild < Token
-  def self.token_name
-    'rebuild'
-  end
-
   def call(options)
     set_triggered_at
     package_name = options[:package].to_param
--- a/src/api/app/models/token/release.rb
+++ b/src/api/app/models/token/release.rb
@@ -1,10 +1,6 @@
 class Token::Release < Token
   include MaintenanceHelper
 
-  def self.token_name
-    'release'
-  end
-
   def call(options)
     set_triggered_at
     return unless options[:package]
--- a/src/api/app/models/token/rss.rb
+++ b/src/api/app/models/token/rss.rb
@@ -1,7 +1,4 @@
 class Token::Rss < Token
-  def self.token_name
-    'rss'
-  end
 end
 
 # == Schema Information
--- a/src/api/app/models/token/service.rb
+++ b/src/api/app/models/token/service.rb
@@ -1,8 +1,4 @@
 class Token::Service < Token
-  def self.token_name
-    'runservice'
-  end
-
   def call(options)
     set_triggered_at
     Backend::Api::Sources::Package.trigger_services(options[:project].to_param,
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -3,10 +3,6 @@ class Token::Workflow < Token
 
   validates :scm_token, presence: true
 
-  def self.token_name
-    'workflow'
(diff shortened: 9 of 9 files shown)


src/api/app/models/token.rb
```diff
@@ -20,7 +20,11 @@ class Token < ApplicationRecord
   OPERATIONS = ['Rebuild', 'Release', 'Service', 'Workflow'].freeze
 
   def token_name
-    self.class.token_name
+    self.class.token_name.downcase
+  end
+
+  def self.token_name
```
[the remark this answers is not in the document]
vpereira: as agreed offline, that's exactly what he was proposing. We are implementing the `#to_s` to return the token type without the prefix `Token::`

src/api/app/views/person/token/index.xml.builder
```diff
@@ -1,6 +1,6 @@
 xml.directory(count: @list.length) do |dir|
   @list.each do |token|
-    p = { id: token.id, string: token.string, kind: token.token_name, name: token.name, triggered_at: token.triggered_at }
+    p = { id: token.id, string: token.string, kind: token, name: token.name, triggered_at: token.triggered_at }
```
vpereira, line 3: TBH, I don't like that we take a field and transform it to `to_s` if it this field isn't really either the name or the token type.

src/api/app/views/webui/users/token_triggers/show.html.haml
```diff
@@ -13,7 +13,7 @@
             = @token.id
           .form-group
             = f.label(:type, 'Operation:')
-            = @token.class.token_name.capitalize
+            = @token.capitalize
```
rubhanazeem, line 16: Is it correct? Shouldn't it be `@token.name.capitalize` ?

src/api/app/views/webui/users/tokens/edit.html.haml
```diff
@@ -22,7 +22,7 @@
                 = @token.id
               .form-group
                 = f.label(:type, 'Operation:')
-                = @token.class.token_name.capitalize
+                = @token.capitalize
```
rubhanazeem, line 25: Here as well

src/api/app/views/webui/users/token_triggers/show.html.haml
vpereira: yes, it is correct. We are displaying the "Operation" there, so it will be "Service", "Workflow", "Release or "Rebuild" and not the custom name given by the user.

src/api/app/views/webui/users/tokens/edit.html.haml
vpereira: the same thing: We are talking about the operation.. it is correct.

src/api/app/views/webui/users/token_triggers/show.html.haml
saraycp: I understand it's not the name given by the user but the "type of token". But, shouldn't the code be like:

```suggestion
            = @token.to_s.capitalize
```

src/api/app/views/webui/users/tokens/edit.html.haml
saraycp: Same here:

```suggestion
                = @token.to_s.capitalize
```

src/api/app/views/webui/users/token_triggers/show.html.haml
vpereira: if you do something like: `puts Token.first` ruby knows it should try to call `to_s` on that object. We normally ask for changes asking to  avoid the explicit "to_s", no?

src/api/app/views/person/token/index.xml.builder
danidoni: I don't like using this also. `Token.token_name` is wrong, but using `token.to_s` is wrong also. I would go by naming to `token_name` as `token_kind` to make it explicit what are we talking about.

vpereira: Note that before we had `Token.token_name` AND `Token#token_name` they were consistent.  The user defined name was introduced later unfortunately as "name". So now we have `Token#name`, and `Token.name` meaning totally different things..

We have already `Token#type`, `Token.token_type`, `Token.name`, `Token#to_s`.. i think `Token#kind` is one more variation for confusion... maybe `Token#operation`? `Token#service` could work too, but since we have already `Token::Service´ that would be confusing too.. 

I personally don't have anything against `Token#token_name` though :)

src/api/app/views/webui/users/token_triggers/show.html.haml
saraycp: I checked it and failed, but not in the context of a view.


```
[11] pry(main)> token.to_s.capitalize
=> "Workflow"
[12] pry(main)> token.capitalize
NoMethodError: undefined method `capitalize' for #<Token::Workflow id: 1, string: "awfaoh4m99jlm3q0vehya6eth1twj23l", user_id: 1, package_id: nil, type: "Token::Workflow", scm_token: "nfx7ldgle29ra23td1v57axxwf56n30f", name: "Testing token", triggered_at: nil>
from /usr/lib64/ruby/gems/3.1.0/gems/activemodel-6.1.4.6/lib/active_model/attribute_methods.rb:469:in `method_missing'
```

I'll check it there too.

src/api/app/views/person/token/index.xml.builder
saraycp: What we call `token_name` is, in fact, the token type but in a humanized way. What about `humanized_token_type`?

src/api/app/views/webui/users/token_triggers/show.html.haml
saraycp: It also crashes when I go to the token edit view:

saraycp: > if you do something like: `puts Token.first` ruby knows it should try to call `to_s` on that object. We normally ask for changes asking to avoid the explicit "to_s", no?

I think what you mean here is what we do when we interpolate. For example, `"#{token}"` is equivalent to `"#{token.to_s}"
`, so we avoid specifying `.to_s` as it is redundant.

hennevogel:
How about going at it from the other side?

`rename_column(:token, :name, :description)` 

And then forget about all of this..

hennevogel:
Judging from `Token.where.not(name: '').map(&:name)` on production, this is also how people use it. They describe what the token is about...

danidoni:
> How about going at it from the other side?
> 
> `rename_column(:token, :name, :description)`
> 
> And then forget about all of this..

This!!

vpereira:
I would still submit (this commit)[https://github.com/openSUSE/open-build-service/pull/12418/commits/b081a925f10852e481ddc15abe95091faf2648e6] in this PR, where we remove the manual method definition in each subclass and instead have it in the parent class.

src/api/app/views/person/token/index.xml.builder
vpereira: Hm, I think we don't want to express the token type but the service being triggered. `service`, `trigger_service`  or `service_type` would be IMO better, but then we have already a `Token::Service` which somehow could lead to confusion too

src/api/app/models/token.rb
```diff
@@ -20,7 +20,11 @@ class Token < ApplicationRecord
   OPERATIONS = ['Rebuild', 'Release', 'Service', 'Workflow'].freeze
 
   def token_name
-    self.class.token_name
+    self.class.token_name.downcase
```
hennevogel, line 23: Nitpick: it's already downcased right?

hennevogel (approved):

src/api/app/models/token.rb
vpereira: yup, fixed it :+1:

PR #12421, opened 2022-04-08
Add new filter for tag push events

saraycp:
Workflow runs can be filtered now by tag push events (from GitHub and GitLab).

Keep in mind that, for GitHub, push events for tags are in the same event type as push events for commits (it's simply "push"). For GitLab, it's a separate event ("Tag Push Hook").

### DIFF
--- a/src/api/app/models/workflow_run.rb
+++ b/src/api/app/models/workflow_run.rb
@@ -67,6 +67,17 @@ class WorkflowRun < ApplicationRecord
     payload.dig(*mapped_source_url) if mapped_source_url
   end
 
+  def generic_event_type
+    # We only have filters for push, tag_push, and pull_request
+    if hook_event == 'Push Hook' || payload.fetch('ref', '').match('refs/heads')
+      'push'
+    elsif hook_event == 'Tag Push Hook' || payload.fetch('ref', '').match('refs/tag')
+      'tag_push'
+    elsif hook_event.in?(['pull_request', 'Merge Request Hook'])
+      'pull_request'
+    end
+  end
+
   private
 
   def parsed_request_headers
--- a/src/api/app/queries/workflow_runs_finder.rb
+++ b/src/api/app/queries/workflow_runs_finder.rb
@@ -1,10 +1,8 @@
 class WorkflowRunsFinder
   EVENT_TYPE_MAPPING = {
-    'pull_request' => 'pull_request',
-    'Merge Request Hook' => 'pull_request',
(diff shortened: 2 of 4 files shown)


src/api/app/models/workflow_run.rb
```diff
@@ -67,6 +67,17 @@ def event_source_url
     payload.dig(*mapped_source_url) if mapped_source_url
   end
 
+  def generic_event_type
+    # We only have filters for push, tag_push, and pull_request
+    if hook_event == 'Push Hook' || payload.fetch('ref', '').match('refs/heads')
```
hennevogel, line 72: Is that `github  || gitlab`?

hennevogel: Or why are we doing this?

rubhanazeem: `Push Hook` represents `gitlab` and `refs/heads` represents `github` for push events. We need this because in the view we don't differentiate b/w github and gitlab events. That's why we made a generic event to represent `push`

hennevogel: Okay RTFMd...

src/api/app/queries/workflow_runs_finder.rb
```diff
@@ -17,18 +15,24 @@ def all
...
-      "request_headers LIKE '%#{real_event_type}%'"
-    end.join(' OR ')
+    query = case generic_event_type
+            when 'tag_push'
+              "request_headers LIKE '%: Tag Push Hook%' OR JSON_EXTRACT(request_payload, '$.ref') LIKE '%refs/tags/%'"
+            when 'push'
+              "request_headers LIKE '%: Push Hook%' OR JSON_EXTRACT(request_payload, '$.ref') LIKE '%refs/heads/%'"
+            else
+              EVENT_TYPE_MAPPING[generic_event_type].map do |event_type|
+                "request_headers LIKE '%#{event_type}%'"
```
hennevogel, line 30: Guess you could be as specific as above with the `:`

hennevogel (changes requested):

src/api/app/models/workflow_run.rb
rubhanazeem: > Okay and for github there is only `push` no matter if it's a push ref or push tag?

Exactly, for Github it's `push` for both `tag` and `push` events. The only difference is in `ref`

hennevogel (approved):

PR #12427, opened 2022-04-11
Use patched yajl ruby dependency

danidoni:
There is a security vulnerability on yajl-ruby, but the upstream patch does not work on ruby 2.5. There is a [patched rubygem package](https://build.opensuse.org/package/binaries/OBS:Server:2.10:Staging/rubygem-yajl-ruby), that we use it for 2.10.

### DIFF
--- a/dist/obs-bundled-gems.spec
+++ b/dist/obs-bundled-gems.spec
@@ -66,7 +66,7 @@ Requires:       perl(GD)
 Requires:       rubygem(ruby:2.5.0:bundler)
 Requires:       rubygem(ruby:2.5.0:rake:%{rake_version})
 Requires:       rubygem(ruby:2.5.0:rack:%{rack_version})
-BuildRequires:  rubygem-yajl-ruby
+BuildRequires:  ruby2.5-rubygem-yajl-ruby
 
 %description -n obs-api-deps
 To simplify splitting the test suite packages off the main package,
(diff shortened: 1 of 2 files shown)


dist/obs-bundled-gems.spec
```diff
@@ -98,6 +99,10 @@ cp %{S:0} %{S:1} .
 # copy gem files into cache
 mkdir -p vendor/cache
 cp %{_sourcedir}/vendor/cache/*.gem vendor/cache
+# Use packaged yajl ruby gem instead of the one from rubygems.
+# This one is patched and runs on ruby 2.5
+# See https://github.com/brianmario/yajl-ruby/security/advisories/GHSA-jj47-x69x-mxrm
+cp %{_libdir}/ruby/gems/2.5.0/cache/yajl-ruby-1.4.2.gemr}/vendor/cache/*.gem vendor/cache
```
krauselukas, line 105: @danidoni the `.gemr}` looks like a typo to me

hennevogel, line 105: `gemr`

Are you test building this anywhere?

hennevogel (changes requested):

dist/obs-bundled-gems.spec
danidoni: Definitely

hennevogel:
Handled differently in #12436 now

PR #12434, opened 2022-04-12
Rename token name description

vpereira:
To eventually reduce confusion, between `Token.name`, `Token#name´, `Token#token_name` and `Token.token_name`, we are renaming `Token#name` to `Token#description`


vpereira:
[Remove Token.token_name method (S)](https://trello.com/c/0yujk7O5/1874-remove-tokentokenname-method-s)

src/api/app/views/person/token/index.xml.builder
```diff
@@ -1,6 +1,6 @@
 xml.directory(count: @list.length) do |dir|
   @list.each do |token|
-    p = { id: token.id, string: token.string, kind: token.token_name, name: token.name, triggered_at: token.triggered_at }
+    p = { id: token.id, string: token.string, kind: token.token_name, name: token.description, triggered_at: token.triggered_at }
```
hennevogel, line 3: Hm I think we can rename this too. There is no `name` support in `osc`, and we introduced `name` just 6 months ago (b3f0122ca0866a866de282b73d36f9b1f6a11ebb)...

hennevogel (changes requested):

src/api/app/models/token.rb
```diff
@@ -8,10 +8,10 @@ class Token < ApplicationRecord
   has_secure_token :string
 
   before_validation do
-    self.name ||= ''
+    self.description ||= ''
```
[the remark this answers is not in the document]
vpereira: I was asking that myself, but didn't want to change the running system. Was that an workaround?

src/api/app/views/person/token/index.xml.builder
vpereira: ah just 6 months ago? Yes for sure in such short period time, the osc team didn't pick up this change. I will change it.

src/api/app/models/token.rb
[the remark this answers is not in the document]
hennevogel: Who knows. Doesn't hurt...

src/api/app/views/person/token/index.xml.builder
vpereira: fixed

hennevogel (approved):

PR #12440, opened 2022-04-13
Provide debug symbol packages for Debian based builds

MatthiasB-EB:
The Open Build Service can be used to create binary packages for various Linux distributions. In case of Debian based Distributions (e.g. Ubuntu) the automatically generated repositories, miss additional packages with debug symbols (provided as *.ddeb files). While the *.ddeb files are available, they are not referenced by the repositories meta data/Package files.

The provided contribution generates additional repositories on the web server, referencing the *.ddeb files. The corresponding repository files are placed in a debian-debug sub-directory. The implementation is more or less a simple copy of the code, already used to provide the additional debian-installer repo. Adding the debian-debug repository to an actual .deb based Linux Distribution makes it possible to simply install debug symbol packages via apt.

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -975,6 +975,20 @@ sub createrepo_debian {
   } else {
     rmdir($udebs);
   }
+
+  my $ddebs = "$extrep/debian-debug";
+  mkdir_p($ddebs) unless -d $ddebs;
+  if (qsystem('chdir', $ddebs, 'stdout', 'Packages.new', 'dpkg-scanpackages', '-t', 'ddeb', '-m', '..', '/dev/null')) {
+    die("    dpkg-scanpackages for ddebs failed: $?\n");
+  }
+  compress_and_rename("$ddebs/Packages.new", "$ddebs/Packages");
+
+  if ( -e "$ddebs/Packages") {
+    createrelease_debian($ddebs, $projid, $repoid, $data, $options);
+  } else {
+    rmdir($ddebs);
+  }
+
 }
 
 sub createrelease_debian {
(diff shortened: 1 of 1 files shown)


src/backend/bs_publish
```diff
@@ -975,6 +975,20 @@ sub createrepo_debian {
   } else {
     rmdir($udebs);
   }
+
+  my $ddebs = "$extrep/debian-debug";
+  mkdir_p($ddebs) unless -d $ddebs;
+  if (qsystem('chdir', $ddebs, 'stdout', 'Packages.new', 'dpkg-scanpackages', '-t', 'ddeb', '-m', '..', '/dev/null')) {
```
bluca, line 981: The debug packages automatically generated by debhelper are actually called foo-dbgsym.deb - see list on http://debug.mirrors.debian.org/debian-debug/pool/main/

MatthiasB-EB: To my understanding this was only the case for the manually created packages. With a newer compat level, those should be generated automatically with a ddeb ending. From my understanding the general idea of the ddeb files was, to be able to handle the ddeb files separately in case you want to mirror a repository. 
https://wiki.debian.org/AutomaticDebugPackages

bluca: That's not the case, if you look at the repository i linked you'll see it

MatthiasB-EB: Ok, looks like it's at the moment only the case for Ubuntu https://wiki.ubuntu.com/Debug%20Symbol%20Packages
For Ubuntu the main repository at http://archive.ubuntu.com/ubuntu/ does not contain any dbgsym/ddeb files, those are located here http://ddebs.ubuntu.com/
It looks like, that Ubuntu seems to differ in the implemented behavior of dh_strip

bluca: It's not just for the moment, it's never going to be any different. Match for -dbgsym instead and it will work for both.

MatthiasB-EB: Sure, but this wouldn't be possible with the current implementation of dpkg-scanpackages, since -t defines the used file extension https://git.dpkg.org/cgit/dpkg/dpkg.git/tree/scripts/dpkg-scanpackages.pl#n249 (at least this is my understanding)

bluca: Yes, it needs to be handled manually

PR #12441, opened 2022-04-13
Downgrade rake to 13.0.1

hennevogel:
Also bring back the test for the rake version and make sure the correct alternative is set on the appliance

### DIFF
--- a/dist/obs-bundled-gems.spec
+++ b/dist/obs-bundled-gems.spec
@@ -15,6 +15,8 @@
 # Please submit bugfixes or comments via http://bugs.opensuse.org/
 #
 
+%define rack_version 2.2.3
+%define rake_version 13.0.1
 
 Name:           obs-bundled-gems
 Version:        2.10~pre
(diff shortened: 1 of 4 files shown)


dist/obs-bundled-gems.spec
```diff
@@ -105,7 +105,12 @@ bundle config build.nokogiri --use-system-libraries
 
 bundle --local --path %{buildroot}%_libdir/obs-api/
 
+# Make sure rake and rack in Gemfile.lock match the versions from the
+# rubygem-rack and ruby2.7 packages
+# otherwise Passenger wont start the app
+
 test -f %{buildroot}%_libdir/obs-api/ruby/2.7.0/gems/rack-%{rack_version}/rack.gemspec
+test -f %{buildroot}%_libdir/obs-api/ruby/2.7.0/gems/rack-%{rake_version}/rack.gemspec
```
M0ses, line 113: Is this line correct? rack != rake. it looks like you copied the line and changed only rack_version -> rake_version but not changed the gem name.

hennevogel: For sure not! Thanks

hennevogel:
Test building in https://build.opensuse.org/package/show/home:hennevogel:branches:OBS:Server:2.10:Staging/obs-server

hennevogel:
@M0ses can you have a look again?

rubhanazeem (approved):

PR #12443, opened 2022-04-13
Move setup/test tasks from openQA test suite here

hennevogel:
It's easier to keep things up to date in one place...


hennevogel:
Will be used by https://github.com/os-autoinst/os-autoinst-distri-obs/pull/28

vpereira (approved):

PR #12444, opened 2022-04-13
Move setup/test tasks from openQA test suite here

hennevogel:
It's easier to keep things up to date in one place...

### DIFF
--- a/dist/t/Makefile
+++ b/dist/t/Makefile
@@ -1,0 +1,25 @@
+.ONESHELL:
+
+all:
+	@echo "Targets: install test dbclean installclean"
+
+install:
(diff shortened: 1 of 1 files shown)


hennevogel:
Will be used by https://github.com/os-autoinst/os-autoinst-distri-obs/pull/28

dist/t/Makefile
```diff
@@ -0,0 +1,25 @@
+.ONESHELL:
```
dmarcoux, line 1: This is new or was it taken from another SCM repository?

dmarcoux:
Isn't this PR pretty much just like #12443?

dist/t/Makefile
hennevogel: This whole Makefile is new

dmarcoux:
Oh, I didn't notice this one is on `master`... :facepalm:

PR #12445, opened 2022-04-14
Track errors from SCMStatusReporter in WorkflowRuns

krauselukas:
In the past we only wrote the exception raised in the scm status reporter to the logfiles, which are not accessible for the users. In order to help the users to debug, we want to show the errors in the workflow runs UI. For now simply by writing it to the response body. 

*How to test:*

Easiest way to test this is by simply setting up a workflow with a `branch_package` step and to mock the code by raising one of the `Octokit` exceptions if the `SCMStatusReporter#call` or somehow let the reporter fail by for example cutting network connection... 

*TODO:*

- [x] Add testcase to `scm_status_reporter_spec.rb` to check that workflow_run tracks error
- [x] Make sure migration can be run without downtime, adjust when necessary @krauselukas 
~- [ ] "Translate" exception message from Octokit to something meaningful~ (follow up PR)

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -10,12 +10,11 @@ class Token::Workflow < Token
     set_triggered_at
     @scm_webhook = options[:scm_webhook]
     workflow_run = options[:workflow_run]
-
     raise Token::Errors::MissingPayload, 'A payload is required' if @scm_webhook.payload.blank?
 
     workflow_run.update(response_url: @scm_webhook.payload[:api_endpoint])
     yaml_file = Workflows::YAMLDownloader.new(@scm_webhook.payload, token: self).call
-    @workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: @scm_webhook, token: self, workflow_run_id: workflow_run.id).call
+    @workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: @scm_webhook, token: self, workflow_run: workflow_run).call
 
     return validation_errors unless validation_errors.none?
 
--- a/src/api/app/services/scm_exception_handler.rb
+++ b/src/api/app/services/scm_exception_handler.rb
@@ -26,6 +26,7 @@ class SCMExceptionHandler
               Octokit::InternalServerError do |exception|
     # FIXME: Inform users about the exceptions
     log(exception)
+    log_to_workflow_run(exception, 'GitHub') if @workflow_run.present?
   end
 
   rescue_from Gitlab::Error::Conflict,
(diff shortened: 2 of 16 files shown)


krauselukas:
[Track errors from SCMStatusReporter in WorkflowRuns (M)](https://trello.com/c/V6E2JPig/1819-track-errors-from-scmstatusreporter-in-workflowruns-m)

rubhanazeem:
@krauselukas some related specs are failing

krauselukas:
> @krauselukas some related specs are failing

Fixed it :)

src/api/app/models/token/workflow.rb
```diff
@@ -20,9 +19,9 @@ def call(options)
     return validation_errors unless validation_errors.none?
 
     # This is just an initial generic report to give a feedback asap. Initial status pending
-    ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, workflow_run).call
+    ScmInitialStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, workflow_run.reload).call
```
hennevogel, line 22: Why reload?

src/api/app/services/scm_exception_handler.rb
```diff
@@ -36,20 +35,19 @@ class SCMExceptionHandler
...
+    @workflow_run = workflow_run
   end
 
   private
 
-  def log(exception)
-    token = Token::Workflow.find_by(scm_token: @scm_token)
-    Rails.logger.error "#{exception.class}: #{exception.message}. TokenID: #{token.id}, User: #{token.user.login}, Event Subscription Payload: #{@event_subscription_payload}"
+  def log_to_workflow_run(exception)
+    @workflow_run.update_to_fail("Failed to report back to SCM: #{exception.message}")
```
hennevogel, line 51: You know which SCM right?

src/api/app/services/scm_initial_status_reporter.rb
```diff
@@ -2,7 +2,7 @@ class ScmInitialStatusReporter < SCMStatusReporter
   attr_accessor :state
 
   def initialize(event_payload, event_subscription_payload, scm_token, workflow_run, event_type = nil)
-    super(event_payload, event_subscription_payload, scm_token)
+    super(event_payload, event_subscription_payload, scm_token, nil, workflow_run.id)
```
hennevogel, line 5: Hm why pass down `nil` if it's never not `nil`?

hennevogel (changes requested):

src/api/app/services/scm_initial_status_reporter.rb
rubhanazeem, line 5: Shouldn't it be?
```
super(event_payload, event_subscription_payload, scm_token, event_type, workflow_run.id)
```

krauselukas: true

src/api/app/models/token/workflow.rb
krauselukas: used it at some point, no reason anymore, I removed it.

src/api/app/services/scm_exception_handler.rb
krauselukas: jupp :)

src/api/app/services/scm_initial_status_reporter.rb
krauselukas: :+1:

krauselukas:
@hennevogel @rubhanazeem please have another look :)

rubhanazeem (approved):

PR #12464, opened 2022-04-21
Implement New Watchlist API

rubhanazeem:
## Old endpoint:
`osc -A dosc api -X PUT 'person/Admin' -T person.xml`
- It's backward compatible.

### Old watchlist version:
  ```
  <person>
    <login>Admin</login>
    <email>root@localhost</email>
    <realname>OBS Instance Superuser</realname>
    <state>confirmed</state>
    <globalrole>Admin</globalrole>
    <ignore_auth_services>false</ignore_auth_services>
    <watchlist>
      <project name="MaintenanceProject"/>
      <project name="home:Admin"/>
    </watchlist>
  </person>
  ```  
  ### New watchlist version:
```
<person>
  <login>Admin</login>
  <email>root@localhost</email>
  <realname>OBS Instance Superuser</realname>
  <state>confirmed</state>
  <globalrole>Admin</globalrole>
  <ignore_auth_services>false</ignore_auth_services>
  <watchlist>
    <project name="MaintenanceProject"/>
    <project name="home:Admin"/>
    <package name="ad" project="home:Admin"/>
    <request number="7"/>
  </watchlist>
</person>
```

### GET watchlist:
`osc -A dosc api 'person/Admin/watchlist'`

### PUT watchlist::
`osc -A dosc api -X PUT 'person/Admin/watchlist' -T watchlist_only.xml`

#### watchlist_only.xml
```
<watchlist>
  <project name="home:Admin"/>
  <project name="MaintenanceProject"/>
  <package name="ad" project="home:Admin"/>
  <request number="7"/>
</watchlist>
```

### DIFF
--- a/src/api/app/controllers/person_controller.rb
+++ b/src/api/app/controllers/person_controller.rb
@@ -10,7 +10,8 @@ class PersonController < ApplicationController
   skip_before_action :extract_user, only: [:command, :register]
   skip_before_action :require_login, only: [:command, :register]
 
-  before_action :set_user, only: [:post_userinfo, :change_my_password]
+  before_action :set_user, only: [:post_userinfo, :change_my_password, :get_watchlist, :put_watchlist]
+  before_action :check_user_belongs_feature_flag, only: [:get_watchlist, :put_watchlist]
 
   def show
     @list = if params[:prefix]
@@ -134,6 +135,34 @@ class PersonController < ApplicationController
     render_ok
   end
 
+  def get_watchlist
+    if @user
+      authorize @user, :check_watchlist?
+
+      render xml: @user.render_axml(render_watchlist_only: true)
+    else
+      @errorcode = 404
+      @summary = 'Requested non-existing user'
+      render_error(status: @errorcode)
+    end
+  end
+
+  def put_watchlist
+    if @user
+      authorize @user, :check_watchlist?
+    else
+      @errorcode = 404
+      @summary = 'Requested non-existing user'
+      render_error(status: @errorcode) && return
+    end
+
+    xml = Xmlhash.parse(request.raw_post)
+    ActiveRecord::Base.transaction do
+      update_new_watchlist(@user, xml)
+    end
+    render_ok
+  end
+
   class NoPermissionToGroupList < APIError
     setup 401, 'No user logged in, permission to grouplist denied'
   end
@@ -193,16 +222,21 @@ class PersonController < ApplicationController
   end
 
   def update_watchlist(user, xml)
-    new_watchlist = []
-    xml.get('watchlist').elements('project') do |e|
-      new_watchlist << e['name']
-    end
+    # Keeping the backwards compatibility
+    if Flipper.enabled?(:new_watchlist, User.session)
+      update_new_watchlist(user, xml)
+    else
+      new_watchlist = []
+      xml.get('watchlist').elements('project') do |e|
+        new_watchlist << e['name']
+      end
 
-    new_watchlist.map! do |name|
-      WatchedProject.find_or_create_by(project: Project.find_by_name!(name), user: user)
+      new_watchlist.map! do |name|
+        WatchedProject.find_or_create_by(project: Project.find_by_name!(name), user: user)
+      end
+      user.watched_projects.replace(new_watchlist)
+      Rails.cache.delete(['watched_project_names', user])
     end
-    user.watched_projects.replace(new_watchlist)
-    Rails.cache.delete(['watched_project_names', user])
   end
   private :update_watchlist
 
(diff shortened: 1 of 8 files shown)


src/api/app/controllers/person_controller.rb
```diff
@@ -134,6 +135,34 @@ def put_userinfo
     render_ok
   end
 
+  def get_watchlist
+    if @user == User.session
+      render xml: User.session.render_axml(render_watchlist_only: true)
+    else
+      render_error(status: 403, errorcode: 'anonymous_user',
+                   message: 'Anonymous user is not allowed here')
```
hennevogel, line 143: If you are checking for nobody there is `.is_nobody?` or `.session!`. If not this error message is pretty misleading...

src/api/app/helpers/watchlist_helper.rb
```diff
@@ -0,0 +1,15 @@
+module WatchlistHelper
+  def self.render(my_model, xml)
```
hennevogel, line 2: Hm why is this a helper and not a view?

src/api/app/controllers/person_controller.rb
```diff
@@ -134,6 +135,34 @@ def put_userinfo
...
+    end
+  end
+
+  def put_watchlist
+    if @user
+      unless @user.login == User.session!.login || User.admin_session?
```
hennevogel, line 149: This is a policy

hennevogel (dismissed):

src/api/app/helpers/watchlist_helper.rb
rubhanazeem: Because this code is used twice and we don't have a clear way to create partials using `XML builder`. That's why I moved it to a helper to keep it DRY and we can not simply render this XML in another root element, XML builder doesn't allow that.

src/api/app/models/user.rb
```diff
@@ -871,6 +871,18 @@ def run_as
     end
   end
 
+  def watchlist_projects
```
dmarcoux, line 874: This method and the others below could be in a query object. This way, we wouldn't increase the size of the `User` model which is already quite large with almost 900+ lines of code.

src/api/app/helpers/watchlist_helper.rb
dmarcoux: It should be possible to render a partial in a view relying on XML builder. This is much easier to understand than having a helper which is acting as a partial.

rubhanazeem: Thank you. It's updated

src/api/app/controllers/person_controller.rb
```diff
@@ -193,16 +222,21 @@ def internal_register
   end
 
   def update_watchlist(user, xml)
-    new_watchlist = []
-    xml.get('watchlist').elements('project') do |e|
-      new_watchlist << e['name']
-    end
+    # Keeping the backwards compatibility
+    if Flipper.enabled?(:new_watchlist, user)
```
saraycp, line 226: In general, everything I tested manually worked fine when the user performing the action is the same as the user to be modified (the one we are editing in the XML).

However, when an admin user is trying to see or update another user, this is not working as expected.

The reason is that in ` check_user_belongs_feature_flag` we are checking `Flipper.enabled?(:new_watchlist, User.session)` so the user performing the action (_Admin_), however, in the line I'm commenting, we are checking `Flipper.enabled?(:new_watchlist, user)` where `user` is not the one in the session but the one given by the `login` tag in the XML (_scp_).

saraycp (changes requested):
In general, everything I tested manually worked fine, except when there is an admin user trying to see or update another user's watchlist. Please read more details in my comments.

src/api/app/controllers/person_controller.rb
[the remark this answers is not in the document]
rubhanazeem: Thanks for pointing it out. It's updated

src/api/spec/controllers/person_controller_spec.rb
```diff
@@ -174,4 +195,59 @@
...
+            <project name="test-proj"/>
+            <package name="test-pkg" project="test-proj"/>
+          </watchlist>
+        XML_DATA
+      end
+
+      before do
+        project = create(:project, name: 'test-proj')
+        package = create(:package, project: project, name: 'test-pkg')
+        Flipper.enable_actor(:new_watchlist, user)
```
saraycp, line 213:
```suggestion
      let(:project) { create(:project, name: 'test-proj') }
      let(:package) { create(:package, project: project, name: 'test-pkg') }

      before do
        Flipper.enable_actor(:new_watchlist, user)
```

danidoni (approved):

src/api/spec/controllers/person_controller_spec.rb
rubhanazeem: We don't need this project and package outside `before` block :slightly_smiling_face:

saraycp: alright :+1:

PR #12470, opened 2022-04-21
Receive notifications for relationships

danidoni:
Receive notifications when a relationship is created or deleted. It is available for any kind of role if the user subscribed to the relationship events.

## This is how it looks like:

Screenshot 2022-04-29 at 12-25-36 Open Build Service

## How to test this PR
- Given you have two users ready, `Admin` and `Iggy` for example, with home projects.
- Iggy should enable the notifications for the "Relationship  was created" and "Relationship was deleted" event types.
- Admin should go to its home project and add Iggy as maintainer, for example.
- Iggy should get a notification saying it's been added as maintainer into `home:Admin` project.

### DIFF
--- a/src/api/app/components/notification_action_description_component.rb
+++ b/src/api/app/components/notification_action_description_component.rb
@@ -3,6 +3,13 @@ class NotificationActionDescriptionComponent < ApplicationComponent
     super
 
     @notification = notification
+    @role = @notification.event_payload['role']
+    @user = @notification.event_payload['who']
+    @target_object = if @notification.event_payload['package']
+                       "#{@notification.event_payload['project']} / #{@notification.event_payload['package']}"
+                     else
+                       @notification.event_payload['project']
+                     end
   end
 
   def call
@@ -15,6 +22,10 @@ class NotificationActionDescriptionComponent < ApplicationComponent
       when 'Event::CommentForPackage'
         commentable = @notification.notifiable.commentable
         "#{commentable.project.name} / #{commentable.name}"
+      when 'Event::RelationshipCreate'
+        "#{@user} made you #{@role} of #{@target_object}"
+      when 'Event::RelationshipDelete'
+        "#{@user} removed you as #{@role} of #{@target_object}"
       end
     end
   end
--- a/src/api/app/components/notification_avatars_component.rb
+++ b/src/api/app/components/notification_avatars_component.rb
@@ -10,8 +10,11 @@ class NotificationAvatarsComponent < ApplicationComponent
   private
 
   def avatar_objects
-    @avatar_objects ||= if @notification.notifiable_type == 'Comment'
+    @avatar_objects ||= case @notification.notifiable_type
+                        when 'Comment'
                           commenters
+                        when 'Project', 'Package'
+                          [User.find_by(login: @notification.event_payload['who'])]
                         else
                           reviews = @notification.notifiable.reviews
                           reviews.select(&:new?).map(&:reviewed_by) + User.where(login: @notification.notifiable.creator)
--- a/src/api/app/models/event/relationship.rb
+++ b/src/api/app/models/event/relationship.rb
@@ -1,7 +1,29 @@
 module Event
   class Relationship < Base
     self.abstract_class = true
-    payload_keys :description, :who, :user, :group, :project, :package, :role
+    payload_keys :description, :who, :user, :group, :project, :package, :role, :notifiable_id
     shortenable_key :description
+
+    def parameters_for_notification
+      super.merge({ notifiable_type: notifiable_type, notifiable_id: notifiable_id })
+    end
+
+    def any_roles
+      [User.find_by(login: payload['user']) || Group.find_by(title: payload['group'])]
+    end
+
+    def notifiable_type
+      return 'Package' if payload['package']
+
+      'Project'
+    end
+
+    def notifiable_id
+      # FIXME: Inherited package coming from a project link via attribute link resolves to the upstream package. This is confusing at least. We need to think about this behaviour later.
+      return Package.get_by_project_and_name(payload['project'], payload['package']).id if payload['package']
+      return Project.get_by_name(payload['project']).id if Project.exists_by_name(payload['project'])
+
+      nil
+    end
   end
 end
--- a/src/api/app/models/event/relationship_create.rb
+++ b/src/api/app/models/event/relationship_create.rb
@@ -2,6 +2,8 @@ module Event
   class RelationshipCreate < Relationship
     self.message_bus_routing_key = 'relationship.create'
     self.description = 'Relationship was created'
+
+    receiver_roles :any_role
   end
 end
 
--- a/src/api/app/models/event/relationship_delete.rb
+++ b/src/api/app/models/event/relationship_delete.rb
@@ -2,6 +2,8 @@ module Event
   class RelationshipDelete < Relationship
     self.message_bus_routing_key = 'relationship.delete'
     self.description = 'Relationship was deleted'
+
+    receiver_roles :any_role
   end
 end
 
--- a/src/api/app/models/event_subscription/find_for_event.rb
+++ b/src/api/app/models/event_subscription/find_for_event.rb
@@ -9,10 +9,10 @@ class EventSubscription
     def subscriptions(channel = :instant_email)
       receivers_and_subscriptions = {}
 
-      event.class.receiver_roles.flat_map do |receiver_role|
+      event.class.receiver_roles.each do |receiver_role|
         # Find the users/groups who are receivers for this event
         receivers = event.send("#{receiver_role}s")
-        receivers = filter_receivers(receivers, channel)
+        receivers = expand_receivers(receivers, channel)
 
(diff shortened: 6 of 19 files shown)


danidoni:
[Add the new subscriptions for Relationship to the Notifications system (L)](https://trello.com/c/Rtxe19ar/1875-add-the-new-subscriptions-for-relationship-to-the-notifications-system-l)

coolo:
if a maintainer relationship is deleted, the removed maintainer gets the notification, right? I guess that's the main use case.

danidoni:
[Refactor notification components](https://trello.com/c/Qq7Yq9nq/2571-refactor-notification-components)

danidoni:
All those `case` in the code are clear hints that those components are candidates to be refactored soon. It would be better to write the code to notify about a new event type in a single place, not scattered around in different components. See the attached Trello card for details

dmarcoux:
@danidoni You'll have to restart the CI, I think the failing spec isn't related to the changes.

src/api/app/models/event/relationship_create.rb
```diff
@@ -2,6 +2,8 @@ module Event
   class RelationshipCreate < Relationship
     self.message_bus_routing_key = 'relationship.create'
     self.description = 'Relationship was created'
+
+    receiver_roles :maintainer, :bugowner, :reviewer, :downloader, :reader
```
danidoni, line 6: We realized that this receiver_roles makes no sense at all. We're gonna remove it, but we need to change then the way we look for the subscribers in the `EventSubscription::FindForEvent#subscriptions`

src/api/app/components/notification_action_description_component.rb
```diff
@@ -15,9 +16,20 @@ def call
...
+      when 'Event::RelationshipCreate'
+        role = @notification.event_payload['role']
+        user = @notification.event_payload['who']
+        object = @notification.event_payload['package'] || @notification.event_payload['project']
+        "#{user} made you #{role} of #{object}"
+      when 'Event::RelationshipDelete'
+        role = @notification.event_payload['role']
+        user = @notification.event_payload['who']
+        object = @notification.event_payload['package'] || @notification.event_payload['project']
+        "#{user} removed you as #{role} of #{object}"
```
coolo, line 28: Is the project name mentioned elsewhere? Most packages have multiple instances and context matters.

danidoni: Mm right, I'll fix that

danidoni: Fixed

src/api/app/models/event_subscription/find_for_event.rb
```diff
@@ -7,9 +7,55 @@ def initialize(event)
     end
 
     def subscriptions(channel = :instant_email)
+      if event.instance_of?(Event::RelationshipCreate) || event.instance_of?(Event::RelationshipDelete)
```
danidoni, line 10: I think this is not really needed. I'm gonna remove it

krauselukas: so you would always use the `target_user_strategy` method right? If I see it correctly `receiver_roles_strategy` would not work with the `any_roles` role right?

krauselukas: because the `any_role` is not implemented in `Event::Base`

danidoni: I should get rid of `target_user_strategy` and use only `receiver_roles_strategy`. We have `any_role` implemented in `Event::Relationship`

src/api/app/models/event/relationship.rb
```diff
@@ -1,7 +1,30 @@
 module Event
   class Relationship < Base
     self.abstract_class = true
-    payload_keys :description, :who, :user, :group, :project, :package, :role
+    payload_keys :description, :when, :who, :user, :group, :project, :package, :role, :notifiable_id
```
hennevogel, line 4: Hm, `when` is always the same as `created_at` right?

src/api/app/models/event/relationship.rb
```diff
@@ -1,7 +1,30 @@
...
     shortenable_key :description
+
+    def parameters_for_notification
+      super.merge({ notifiable_type: notifiable_type, notifiable_id: notifiable_id })
```
hennevogel, line 8: I think AR is clever enough to figure all the out for itself if you just assign the object as `notifiable`...

src/api/app/models/event/relationship_delete.rb
```diff
@@ -2,6 +2,8 @@ module Event
   class RelationshipDelete < Relationship
     self.message_bus_routing_key = 'relationship.delete'
     self.description = 'Relationship was deleted'
+
+    receiver_roles :any_role
```
hennevogel, line 6: 👍🏿

hennevogel (changes requested):
I don't think anyone needs the distinction between Added/Removed. At least in the subscription area. But that could be done some when else..

src/api/app/models/event/relationship.rb
danidoni: No, it's not working because this is a hash and then I need to branch by `notifiable_type`...

hennevogel: ok

src/api/app/models/event/relationship.rb
danidoni: Yes, basically yes. Let me change this.

danidoni: This param equals to `created_at` now

hennevogel: What I mean: Why `Relationship` or `Time.zone.now` at all? The `Event.created_at` or the `Notification.created_at` is enough...

saraycp (changes requested):
There is a corner case where this implementation fails.
Steps:
- The user goes to the subscriptions page
- Choose the groups they want to receive the notifications from.
- Select the channel for the group:
  - if they select both email and web, the notification is created.
  - if they select only web channel, the notification is **not** created.

Screenshot 2022-05-10 at 12-42-37 Open Build Service

saraycp (dismissed):
The email part of the notification system (for Relationship events) is not implemented. I'm working on it...

src/api/app/models/event/relationship.rb
danidoni: I think it's already using the `created_at` [here[(https://github.com/openSUSE/open-build-service/pull/12470/files#diff-1e95a883031a5d59e3bd3fcda74ecb74e3739d3cba2b164ae15fac0a63bc6970R13), right? So this `when` parameter is useless...

hennevogel: yeah

src/api/app/components/notification_avatars_component.rb
```diff
@@ -10,8 +10,11 @@ def initialize(notification)
   private
 
   def avatar_objects
-    @avatar_objects ||= if @notification.notifiable_type == 'Comment'
+    @avatar_objects ||= case @notification.notifiable_type
+                        when 'Comment'
                           commenters
+                        when 'Project', 'Package'
+                          User.where(login: @notification.event_payload['who'])
```
hennevogel, line 17: You expect more than one `User`?

src/api/app/models/event/relationship.rb
hennevogel, line 4: You are not using `when`...

src/api/app/models/event/relationship.rb
```diff
@@ -1,7 +1,30 @@
...
+      'Project'
+    end
+
+    def notifiable_id
+      notifiable = if payload['package']
+                     Package.get_by_project_and_name(payload['project'], payload['package'])
+                   else
+                     Project.get_by_name(payload['project'])
+                   end
+      notifiable&.id
```
hennevogel, line 27: Did you think about all the types of [links](https://github.com/openSUSE/open-build-service/wiki/Links) here?

src/api/app/models/relationship.rb
```diff
@@ -177,7 +177,8 @@ def event_parameters
     parameters = { who: User.session.login,
                    user: user&.login,
                    group: group&.title,
-                   role: role.title }
+                   role: role.title,
+                   notifiable_id: id.to_i }
```
hennevogel, line 181: You expect this to me something else than an int?

hennevogel (changes requested):

src/api/app/components/notification_avatars_component.rb
danidoni: Nope, fixing

src/api/app/models/event/relationship.rb
danidoni: True

src/api/app/models/event/relationship.rb
danidoni: Not really, no

src/api/app/models/relationship.rb
danidoni: No

hennevogel (approved):

PR #12477, opened 2022-04-22
Set case-insensitive order for watchlist elements

saraycp:
The sorting was a bit weird when combining uppercase and lowercase names. This is the explanation:

Performing a simple order query like `Project.all.order('name')` will sort the results on a case-sensitive manner (displaying the uppercase ones first).

order

Using `LOWER` like `Project.all.order('LOWER(name)')` we can sort on a case-insensitive manner. It groups the same letters together but the results mix uppercase and lowercase on a messy way.

order_I

The combination of case-insensitive and case-sensitive sorting like `Project.all.order('lower(name), name')` gets exactly what we want.

order_II

Follw up https://github.com/openSUSE/open-build-service/pull/12460

### DIFF
--- a/src/api/app/components/watchlist_component.rb
+++ b/src/api/app/components/watchlist_component.rb
@@ -61,12 +61,15 @@ class WatchlistComponent < ApplicationComponent
     end
   end
 
+  # Sort projects and packages names in a case-insensitive manner first
+
   def projects
-    @projects ||= Project.joins(:watched_items).where(watched_items: { user: @user }).order(:name)
+    @projects ||= Project.joins(:watched_items).where(watched_items: { user: @user }).order('LOWER(name), name')
   end
 
   def packages
-    @packages ||= Package.includes(:project).joins(:watched_items).where(watched_items: { user: @user }).order('projects.name, packages.name')
+    @packages ||= Package.joins(:project).joins(:watched_items).where(watched_items: { user: @user })
+                         .order('LOWER(projects.name), projects.name, LOWER(packages.name), packages.name')
   end
 
   def bs_requests


src/api/app/components/watchlist_component.rb
```diff
@@ -61,12 +61,15 @@ def toggle_watchable_path
...
+  # Sort projects and packages names in a case-insensitive manner first
+
   def projects
-    @projects ||= Project.joins(:watched_items).where(watched_items: { user: @user }).order(:name)
+    @projects ||= Project.joins(:watched_items).where(watched_items: { user: @user }).order('LOWER(name), name')
   end
 
   def packages
-    @packages ||= Package.includes(:project).joins(:watched_items).where(watched_items: { user: @user }).order('projects.name, packages.name')
+    @packages ||= Package.joins(:project).joins(:watched_items).where(watched_items: { user: @user })
```
krauselukas, line 71: @saraycp what is the reason to switch from `includes` to `joins`? I think we work with the associated project of the gathered packages here https://github.com/openSUSE/open-build-service/blob/80edd61e9b411c34a052e56702e9e88a18dc9a4f/src/api/app/components/watched_items_list_component.html.haml#L5
So I think using the eager loading of includes is just fine to avoid additional queries

krauselukas: @saraycp ahh okay, sorry for the noise, it doesnt work with `includes` in this case...

krauselukas (approved):

src/api/app/components/watchlist_component.rb
saraycp: With `includes` the query failed when trying to order by project name. I mean, the problem was on this part `.order('LOWER(projects.name), projects.name')`. I tried both plural and singular...

If you have any better idea, I would appreciate it.

PR #12485, opened 2022-04-25
Make the link to a request a bit more descriptive in the watchlist

saraycp:
In the watchlist, a link to a request used to look like `Request #123`. Now we display a bit more information, the action type. So we have something like `Add Role Request #123`.

This is the result
descriptive_request_link

This is just the first step, we are going to display even more information, which is hidden at first sight. The final result will look similar to this mockup:

mockup_watchlist_request


src/api/app/helpers/webui/request_helper.rb
```diff
@@ -167,4 +167,18 @@ def list_maintainers(maintainers)
       user_with_realname_and_icon(maintainer, short: true)
     end.to_sentence.html_safe
   end
+
+  def descriptive_request_link(bs_request)
```
dmarcoux, line 171: You wrote in the commit message that this is going to be used in other areas of the code, but I don't see it here. Is this going to happen in another PR?

saraycp: This is not the method I was talking about.
I said the method [request_type_of_action](https://github.com/openSUSE/open-build-service/pull/12485/commits/9eba086259fb7da11c52765e6fcdcf46247c12af#diff-02694479bd751d8fea3a83a7d73e429e721178934e680726987b3856e15dd591R172) is going to be used from another part of the code,  so I move the helper from the component to a helper. This way, I can use it from the component and from the `descriptive_request_link` helper as well.

dmarcoux: Oh okay, I understand now :+1:

dmarcoux (approved):

src/api/app/components/watched_items_list_component.html.haml
```diff
@@ -11,8 +11,6 @@
           %i.fas.fa-cubes.mr-1
           #{item}
       - when 'BsRequest'
-        = link_to(request_show_path(number: item.number), class: 'text-word-break-all') do
-          = image_tag('icons/request-icon.svg', height: 18, class: 'mr-1 color-inverted')
-          Request ##{item.number}
+        = helpers.descriptive_request_link(item)
```
hennevogel, line 14: What are you gaining by moving those three lines of haml into a helper? I know what `link_to` and `image_tag` etc. do. I have to look up what this helper does....

hennevogel: I would also prefer: icon number type. Like 

request-list

hennevogel (changes requested):

src/api/app/components/watched_items_list_component.html.haml
saraycp: I no longer rely on a helper and the list displays icon, number and type as suggested.

PR #12488, opened 2022-04-26
Add instrumentation to the endpoint NotificationsController#update.

vpereira:
Because of the bulk update approach used, it bypasses all model callbacks, therefore the instrumentation must be done manually, loosing some information for the sake of speed.


dmarcoux:
What is the information which is lost?

vpereira:
> What is the information which is lost?

the kind of notifiable.

vpereira:
[Instrument Notification Feature (M)](https://trello.com/c/T4SRjRHW/1876-instrument-notification-feature-m)

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -22,10 +22,14 @@ def update
                       fetch_notifications.where(id: params[:notification_ids])
                     end
     # rubocop:disable Rails/SkipsModelValidations
-    unless notifications.update_all('delivered = !delivered')
+    updated_notifications = notifications.update_all('delivered = !delivered')
+    # rubocop:enable Rails/SkipsModelValidations
+
+    if updated_notifications.blank?
```
dmarcoux, line 28: This code is always going to be **false**, since `update_all` returns the number of records as an integer, not a string. So instead of using `blank?`, we should use `zero?`. We can also avoid checking a second time for the number of updated notifications. The code to send the metrics doesn't have to be in a method, it's just called here and it's quite short.

```ruby
    if updated_notifications.zero?
      # flash message code
    else
      # send metrics to RabbitMQ
    end
```

dmarcoux (changes requested):

src/api/app/controllers/webui/users/notifications_controller.rb
hennevogel: How will this happen? `update_all` returns always number I think.

```
[2] pry(main)> User.where(login: 'peter').count
=> 0
[3] pry(main)> User.where(login: 'peter').update_all(biography: 'franz')
=> 0
```

```
[4] pry(main)> User.where(login: 'hennevogel').count
=> 1
[6] pry(main)> User.where(login: 'hennevogel').update_all(biography: 'franz')
=> 1
```

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -22,10 +22,14 @@ def update
...
-    unless notifications.update_all('delivered = !delivered')
+    updated_notifications = notifications.update_all('delivered = !delivered')
+    # rubocop:enable Rails/SkipsModelValidations
+
+    if updated_notifications.zero?
       flash.now[:error] = "Couldn't mark the notifications as #{notifications.first.unread? ? 'read' : 'unread'}"
     end
-    # rubocop:enable Rails/SkipsModelValidations
+
+    send_notifications_information_rabbitmq(updated_notifications.size) if updated_notifications.zero?
```
hennevogel:
```
[10] pry(main)> 1.size
=> 8
[11] pry(main)> 6.size
=> 8
[12] pry(main)> 9382721.size
=> 8
```

hennevogel (changes requested):

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -22,10 +22,14 @@ def update
...
                     end
     # rubocop:disable Rails/SkipsModelValidations
-    unless notifications.update_all('delivered = !delivered')
+    updated_notifications = notifications.update_all('delivered = !delivered')
+    # rubocop:enable Rails/SkipsModelValidations
+
+    if updated_notifications.zero?
       flash.now[:error] = "Couldn't mark the notifications as #{notifications.first.unread? ? 'read' : 'unread'}"
+    else
+      send_notifications_information_rabbitmq(updated_notifications.size)
```
dmarcoux, line 31: `updated_notifications` is already an integer, `.size` isn't needed.

```suggestion
      send_notifications_information_rabbitmq(updated_notifications)
```

src/api/app/controllers/webui/users/notifications_controller.rb
vpereira: I think I had `empty?` and it was suggested to `blank?` when the right approach would be `zero?`

src/api/app/controllers/webui/users/notifications_controller.rb
vpereira: ah no right, I was expecting an array of elements changed, not a number.

src/api/app/controllers/webui/users/notifications_controller.rb
vpereira: yup, now I see, I was expecting the the array of elements changed, not a number. :+1:

hennevogel (approved):

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -91,4 +95,9 @@ def paginated_notifications
   def show_read_all_button?
     fetch_notifications.count > Notification::MAX_PER_PAGE
   end
+
+  def send_notifications_information_rabbitmq(number_of_records)
+    RabbitmqBus.send_to_bus('metrics',
+                            "notification.delivered,web=true,rss=false value=#{number_of_records}")
```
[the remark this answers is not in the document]
dmarcoux: What is the definition of delivered in this context? It's not tied to the `delivered` columns in the `notification` model?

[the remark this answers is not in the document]
hennevogel: > What is the definition of delivered in this context?

Whatever you want to measure. Maybe this is rather `notification.state_change`? This is what we are interested in AFAIU.

PR #12490, opened 2022-04-26
Add notification filters for Relationship records

dmarcoux:
This PR depends on #12470.

image

### DIFF
--- a/src/api/app/queries/notifications_finder.rb
+++ b/src/api/app/queries/notifications_finder.rb
@@ -27,6 +27,16 @@ class NotificationsFinder
     @relation.where(notifiable: User.session.outgoing_requests(all_states: true), delivered: false)
   end
 
+  def for_relationships_created
+    @relation.where(event_type: 'Event::RelationshipCreate', delivered: false)
+  end
+
+  def for_relationships_deleted
+    @relation.where(event_type: 'Event::RelationshipDelete', delivered: false)
+  end
+
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # We need to refactor this method, the `case` statement is way too big
   def for_notifiable_type(type = 'unread')
     notifications = self.class.new(with_notifiable)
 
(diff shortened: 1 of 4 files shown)


src/api/app/queries/notifications_finder.rb
```diff
@@ -27,6 +27,14 @@ def for_outgoing_requests
...
 
+  def for_relationships_created
+    @relation.where(notifiable_type: 'Relationship', event_type: 'Event::RelationshipCreate', delivered: false)
+  end
+
+  def for_relationships_deleted
+    @relation.where(notifiable_type: 'Relationship', event_type: 'Event::RelationshipDelete', delivered: false)
+  end
+
   def for_notifiable_type(type = 'unread')
```
dmarcoux, line 38: I really don't like the huge `case` statement in this method... I'm looking into refactoring this.

dmarcoux:
[Add the new subscriptions for Relationship to the Notifications system (L)](https://trello.com/c/Rtxe19ar/1875-add-the-new-subscriptions-for-relationship-to-the-notifications-system-l)

src/api/app/queries/notifications_finder.rb
dmarcoux: After looking into this, I tried the following:
1. Splitting up this query object in multiple query objects which are sometimes dependent on one another (like `unread` is used  `comments` and `requests`. This makes the code more complicated with the increased abstraction.
2. Use a hash with the type (`comments`, `read`, etc...) mapping to the method which needs to be called, something like this:
```ruby
  def for_notifiable_type(type = 'unread')
    notifications = self.class.new(with_notifiable)

    finders = {
      read: notifications.method(:read),
      unread: notifications.method(:unread),
      comments: notifications.method(:for_comments),
      requests: notifications.method(:for_requests),
      incoming_requests: notifications.method(:for_incoming_requests),
      outgoing_requests: notifications.method(:for_outgoing_requests)
    }

    # Fetch notifications for the type and if it's somehow not found, default back to unread notifications
    finders.fetch(type.to_sym, finders[:unread]).call
  end
```

I find both potential solutions to not improve the code. Someone has another idea?

vpereira: This is the pythonic solution for lack of support for switch in python. tbh, i think the `case/when` is easier to read. The "huge" is just indentation, since the number of conditions are the same. I would be interested to know how codeclimate will calculate it.

saraycp (approved):

PR #12492, opened 2022-04-26
Add collapsible tooltips to display more info about requests in watchlist

saraycp:
The information about source and target projects and packages together with the state of the request, is displayed on a tooltip which is initially collapsed.

watchlist_request_general_view

On mobile:

watchlist_request_mobile

Multiple actions request:

multiple_actions_request_in_watchlist

### DIFF
--- a/src/api/app/components/bs_request_action_source_and_target_component.rb
+++ b/src/api/app/components/bs_request_action_source_and_target_component.rb
@@ -1,0 +1,37 @@
+class BsRequestActionSourceAndTargetComponent < ApplicationComponent
+  attr_reader :bs_request_action, :number_of_bs_request_actions
+
+  def initialize(bs_request)
+    super
+
+    @bs_request = bs_request
+    @bs_request_action = @bs_request.bs_request_actions.first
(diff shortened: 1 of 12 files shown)


src/api/app/components/bs_request_action_source_and_target_component.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    [bs_request_action.target_project, bs_request_action.target_package].compact.join(' / ')
+  end
+
+  def bs_request_action
+    @bs_request_action ||= @bs_request.bs_request_actions.first
+  end
+
+  def number_of_bs_request_actions
+    @number_of_bs_request_actions ||= @bs_request.bs_request_actions.size
```
danidoni, line 39: I think, this can be set once in the constructor and then the method and the memoization is not needed if you use an `attr_reader :number_of_bs_request_actions`

saraycp: done

danidoni (approved):

rubhanazeem (approved):

PR #12494, opened 2022-04-27
Add instrumentation to EventSubscription#enabled

vpereira:
Instrument the EventSubscription model. Send metrics to RabbitMQ always when an event subscription gets enabled/disabled

### DIFF
--- a/src/api/app/models/event_subscription.rb
+++ b/src/api/app/models/event_subscription.rb
@@ -57,6 +57,8 @@ class EventSubscription < ApplicationRecord
     end
   }
 
+  after_save :track_event_subscription_enabled, if: :saved_change_to_enabled?
+
   def self.receiver_roles_to_display(user)
     roles = RECEIVER_ROLE_TEXTS.keys
     # We have to show the new_watchlist subscription options to the admin in order to allow the global
@@ -99,6 +101,11 @@ class EventSubscription < ApplicationRecord
 
   def self.without_disabled_or_internal_channels
     channels.keys.reject { |channel| channel == 'disabled' || channel.in?(INTERNAL_ONLY_CHANNELS) }
+  end
+
+  def track_event_subscription_enabled
+    RabbitmqBus.send_to_bus('metrics',
+                            "event_subscription.enabled,event_type=#{eventtype},channel=#{channel} value=1")
   end
 end
 
--- a/src/api/spec/models/event_subscription_spec.rb
+++ b/src/api/spec/models/event_subscription_spec.rb
@@ -1,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe EventSubscription, type: :model do
+  describe 'Instrumentation' do
+    let(:user) { create(:confirmed_user, :with_home, login: 'cameron') }
+    let(:token) { create(:workflow_token, user: user) }
+
+    let!(:target_project) { create(:project, name: 'test-target-project:openSUSE:open-build-service:PR-4', maintainer: user) }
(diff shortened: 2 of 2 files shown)


vpereira:
[Instrument Notification Feature (M)](https://trello.com/c/T4SRjRHW/1876-instrument-notification-feature-m)

dmarcoux:
Could this code be in a separate file under https://github.com/openSUSE/open-build-service/tree/master/src/api/app/instrumentations? This way whenever it's possible, all instrumentation code is under the same directory.

vpereira:
> Could this code be in a separate file under https://github.com/openSUSE/open-build-service/tree/master/src/api/app/instrumentations? This way whenever it's possible, all instrumentation code is under the same directory. With concerns, this code can be included in the model.

Later maybe...

src/api/app/models/event_subscription.rb
```diff
@@ -100,6 +102,11 @@ def parameters_for_notification
   def self.without_disabled_or_internal_channels
     channels.keys.reject { |channel| channel == 'disabled' || channel.in?(INTERNAL_ONLY_CHANNELS) }
   end
+
+  def track_event_subscription_enabled
+    RabbitmqBus.send_to_bus('metrics',
+                            "event_subscription.enabled,event_type=#{eventtype},channel=#{channel},value=1")
```
dmarcoux, line 108: `value` will be processed as a tag if the last comma isn't replaced by a space.

```suggestion
                            "event_subscription.enabled,event_type=#{eventtype},channel=#{channel} value=1")
```

dmarcoux:
> > Could this code be in a separate file under https://github.com/openSUSE/open-build-service/tree/master/src/api/app/instrumentations? This way whenever it's possible, all instrumentation code is under the same directory. With concerns, this code can be included in the model.
> 
> Later maybe...

Could you please do it now? This is what we defined in our wiki for [Application Health Monitoring](https://github.com/openSUSE/open-build-service/wiki/Application-Health-Monitoring#best-practices).

vpereira:
> Could you please do it now? This is what we defined in our wiki for [Application Health Monitoring](https://github.com/openSUSE/open-build-service/wiki/Application-Health-Monitoring#best-practices).

I don't think that this approach is good. Specially when involving callbacks.  If you declare the same method twice in a ruby class just the last one is executed. So if I have a callback `after_save` in a module, and include in a mixing as you are doing [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/scm_webhook.rb#L4) if somebody later in the same model, add a `after_save`, the first one will be overwritten.  Better would be to keep all callbacks together, or have a rubocop plugin to warn us when it we have duplicated calls to the same callback.

dmarcoux:
Just to explain the discussion @vpereira and I had privately.

I didn't know that the changes I added to the wiki for [Application Health Monitoring](https://github.com/openSUSE/open-build-service/wiki/Application-Health-Monitoring#best-practices) weren't discussed/approved by the team. I assumed this was done as the card tracking this work was moved to our `Done` lane. Sorry for that.

I agree with @vpereira that it's potentially quite problematic if we overwrite a callback when we include a concern defined under https://github.com/openSUSE/open-build-service/tree/master/src/api/app/instrumentations. This wasn't an issue so far, but it's risky to not consider this. It will definitely hit us one day and I prefer to prevent this now. It leaves us with having the instrumentation code in the models themselves, even though it's not optimal. For now, I don't see another solution.

dmarcoux (approved):

PR #12495, opened 2022-04-27
[api] support submit request of _project

adrianschroeter:
For example for submiting build config (osc meta prjconf) changes.

Also fixing copy operation on _project package container

OBS-190

### DIFF
--- a/src/api/app/controllers/source_package_command_controller.rb
+++ b/src/api/app/controllers/source_package_command_controller.rb
@@ -284,17 +284,21 @@ class SourcePackageCommandController < SourceController
     end
 
     # create target package, if it does not exist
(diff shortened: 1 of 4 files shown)


coolo:
This still a doomed concept. Noone uses that as the flaws outweigh the benefits over email :(

hennevogel:
> This still a doomed concept

@coolo can you elaborate a bit?

coolo:
The only ever request tried was https://build.suse.de/request/show/277159 and it failed drastically. And SP5 got meanwhile at least 8 changes applied - all by means of chatops or emailsops.

Pretending _project is a package when it's not breaks just too many tools/concepts - including OBS's own webui.

adrianschroeter:
replaced with git projects now

PR #12499, opened 2022-04-28
[backend] avoid warnings from createrepo

adrianschroeter:
Use --discard-additional-metadata only together with --update, it has no
effect otherwise


M0ses (approved):
LGTM

PR #12506, opened 2022-04-29
[api] support to limit release source project

adrianschroeter:
Introducing the new OBS:LimitReleaseSourceProject attribute which allows to list
accepted source project.

It has only an effect on projects atm.


src/api/app/models/bs_request_action_maintenance_release.rb
```diff
@@ -162,6 +162,16 @@ def sanity_check!
         repo.check_valid_release_target!(rt.target_repository)
       end
     end
+
+    tprj = Project.get_by_name(target_project)
```
hennevogel, line 166: This will throw an exception if the `target_project` does not exist. There is also an association you can use called `target_project_object`.

hennevogel (changes requested):

PR #12521, opened 2022-05-02
[dist] use psmisc (pgrep/pkill) in obsscheduler script

bugfinder:
and add the proper dependency in obs-server package
(before we were silently relying on sysvinit-tools present)

### DIFF
--- a/dist/obs-server.spec
+++ b/dist/obs-server.spec
@@ -225,6 +225,8 @@ Requires:       user(obsservicerun)
 Requires:       zstd
 # needed for optional sqlite databases, which are default for new installations
 Requires:     perl-DBD-SQLite
+# psmisc for obsscheduler script
+Requires:       psmisc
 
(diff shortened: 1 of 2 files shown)


dist/obs-server.spec
```diff
@@ -225,6 +225,8 @@ Requires:       user(obsservicerun)
 Requires:       zstd
 # needed for optional sqlite databases, which are default for new installations
 Requires:     perl-DBD-SQLite
+# psmisc for obsscheduler script
+Requires:       psmisc
```
eduardoj, line 229:
```suggestion
```
Is `psmisc` still needed after changing `checkproc` to `pgrep` and `killall` to `pkill`?

eduardoj: I mean, `pgrep` and `pkill` are provided by `procps`, which is already required.

mlschroe:
Obsolete by now.

PR #12528, opened 2022-05-04
[backend] create new signing keys as rsa@4096

adrianschroeter:


adrianschroeter:
Just created for reference, 4096 might slow us down to much. But let's not forget it ...

dirkmueller:
GnuPG v2.2.22 (August 2020) upgraded to RSA3072, which I think we should resemble as well. 

see https://wiki.gnupg.org/LargeKeys

msmeissn:
i would say we can start to proceed with rsa3072 as default now. (and evalute parametric setting, including also elliptic curves)

mlschroe:
Wait wait, with this you are killing some UEFI versions. We had to go back to 2048 because of this in the past.

dirkmueller:
windows compatibility certification requires RSA2048 or better, see https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/windows-secure-boot-key-creation-and-management-guidance#21-metrics-used

adrianschroeter:
will get done via https://github.com/openSUSE/open-build-service/pull/13807

PR #12538, opened 2022-05-05
[backend] add bs_admin --shutdown-all-schedulers option

mlschroe:
We also accept an optional wait timeout. If the timeout is
reached, bs_admin exits with status 2.


eduardoj (approved):

M0ses (approved):
LGTM

PR #12539, opened 2022-05-05
Remove watchlist items with confirmation

saraycp:
Ask for confirmation every time we try to remove an item from the wachtlist.

In other to display the dialog on an outer position to cover the whole page content, we need to pass the details about the item by data attributes.

Follow up on https://github.com/openSUSE/open-build-service/pull/12537

Screenshot 2022-05-05 at 20-10-39 Open Build Service


src/api/app/components/confirmation_dialog_component.rb
```diff
@@ -0,0 +1,20 @@
...
+  attr_accessor :modal_id, :action, :modal_title, :confirmation_text, :method
+
+  def initialize(modal_id:, options: {})
+    super
+
+    @modal_id = modal_id
+    @action = options[:action] || '#'
+    @modal_title = options[:modal_title] || 'Do you really want to remove this item?'
+    @confirmation_text = options[:confirmation_text] || 'Please confirm you want to remove this item.'
+    @method = options[:method] || :delete
```
krauselukas, line 18: @saraycp I would be careful with defaulting to a `delete` action, once forgotten to set we might have a problem. IMO its not worth defaulting to anything here.

krauselukas (changes requested):

src/api/app/components/confirmation_dialog_component.rb
saraycp: Alright, I'm going to make the `method` parameter compulsory, so we are forced to consciously decide which method we want to use when we use the component.

src/api/app/components/watched_items_list_component.html.haml
```diff
@@ -3,29 +3,42 @@
...
             %i.fas.fa-cubes.mr-1
-            #{item}
-          = link_to(project_toggle_watched_item_path(project_name: item.name), method: :put, class: 'text-light ml-auto') do
-            %i.fa.fa-times-circle
+            = name
+          = link_to('#', class: 'text-light ml-auto',
+                    data: { toggle: 'modal',
+                            target: '#delete-item-from-watchlist-modal',
+                            confirmation_text: "Please confirm you want to remove '#{name}' from the watchlist.",
+                            action: project_toggle_watched_item_path(project_name: name) }) do
```
krauselukas, line 27: @saraycp maybe you can also set the modal title here, would be a bit nicer to mention the type as well. So instead of `'Do you really want to remove this item?'` maybe something like `'Do you really want to remove this #{type}?'

danidoni (approved):

src/api/app/components/watched_items_list_component.html.haml
saraycp: Good suggestion. Done

src/api/app/components/confirmation_dialog_component.rb
saraycp: done

saraycp:
> @saray maybe a tiny spec for the new component? :)

I discarded adding a component preview since the dialog is hidden by default (only an `on click` event triggers the JavaScript code to make it visible). So it makes no sense to have a view component to display a hidden element.

But we can have a spec, sure!! Please have a look to the last commit.

PR #12562, opened 2022-05-11
Do not swallow exceptions when developing and testing

danidoni:
On development and testing environments some exceptions never bubble up to the surface making debugging difficult.
This change retriggers the errors when working in development and testing environments.


src/api/app/services/notification_service/notifier.rb
```diff
@@ -30,6 +30,7 @@ def call
       end
     rescue StandardError => e
       Airbrake.notify(e, event_id: @event.id)
+      raise e unless Rails.env.production?
```
hennevogel, line 33: This all makes little sense. Why catch and report this at all?

hennevogel (changes requested):

src/api/app/services/notification_service/notifier.rb
danidoni: Do you think we can remove this error handler? I do not remember why we put this there, this change was introduced 2 years ago... This is used by Clockwork, that has its own error handler, and it is also used in a data migration, which will report errors on the stderr...

hennevogel (approved):

PR #12564, opened 2022-05-11
[api] fix do not auto decline request with an embargo date

adrianschroeter:


eduardoj:
I rebased on master. This change needs to be tested in [spec/models/bs_request_spec.rb](https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/models/bs_request_spec.rb).

hennevogel (approved):

PR #12565, opened 2022-05-11
Use Ajax for watchlist functionality

rubhanazeem:
If someone adds or removes an item from the watchlist, the whole page refreshes and it closes the watchlist bar.
This PR is introducing ajax for watchlist add/remove features to avoid page reloads.


src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -2,6 +2,7 @@ class Webui::WatchedItemsController < Webui::WebuiController
   before_action :require_login
   before_action :check_user_belongs_feature_flag
   before_action :set_item
+  before_action :set_data
```
hennevogel, line 5:
```suggestion
  before_action :set_watchable
```

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -20,7 +21,10 @@ def toggle_watched_item
       flash[:success] = "Added #{FLASH_PER_WATCHABLE_TYPE[@item.class]} to the watchlist"
     end
 
-    redirect_back(fallback_location: root_path)
+    respond_to do |format|
+      format.html
```
hennevogel, line 25: Why keep that around?

hennevogel (changes requested):

src/api/app/controllers/webui/watched_items_controller.rb
rubhanazeem: We don't need it. All the requests to this route should be js requests now. Thanks for pointing that out.

src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -35,6 +38,12 @@ def set_item
             end
   end
 
+  def set_watchable
+    @package = Package.find_by_project_and_name(params[:project_name], params[:package_name])
+    @project = Project.find_by(name: params[:project_name])
```
hennevogel, line 43:
```suggestion
    @project = Project.get_by_name(params[:project_name])
```

hennevogel: think about [links](https://github.com/openSUSE/open-build-service/wiki/Links)....

src/api/app/components/confirmation_dialog_component.html.haml
```diff
@@ -6,7 +6,7 @@
       .modal-body
         %p.confirmation-text= confirmation_text
 
-        = form_tag(action, method: method) do
+        = form_tag(action, method: method, remote: true) do
```
saraycp, line 9: The parameters `action` and `method` are passed when we call the component, in other to make this component reusable from other dialogs of the application. I would do the same with `remote` because not all the dialogs are going to use Ajax.

Anyway, this change is not compulsory in this PR. We can change it at the time we are going to re-use the component. We have [a card](https://trello.com/c/gNicWZMY/2578-replace-partials-by-confirmationdialogcomponent) for that in our Free Style backlog.

src/api/app/controllers/webui/watched_items_controller.rb
saraycp: One of the cases I tested by chance: we have a project named [link_project](https://obs-reviewlab.opensuse.org/rubhanazeem-watchlist-ajax/project/show/linked_project) linked to another one, as its meta contains `<link project="home:Admin"/>`. So it has a bunch of inherited packages. I go to the overview of [one of the inherited packages](https://obs-reviewlab.opensuse.org/rubhanazeem-watchlist-ajax/package/show/linked_project/ad) and try to add it to the Watchlist, however, we are not displaying the "Watch this package" link there. In this corner case, we are safe.

rubhanazeem: @hennevogel it's updated. We did some more tests and found that we do not allow the user to watch the linked package. This [check](https://github.com/openSUSE/open-build-service/blob/a03fb07c37cf3f040d7447054f76a866a7230f9f/src/api/app/components/watchlist_component.rb#L35) does that work.

hennevogel: Okay but you still will have to use the other methods...

hennevogel (approved):

PR #12570, opened 2022-05-12
[api] protect us against broken requests due to multiple entries

adrianschroeter:
The active record validation is not protecting us here on creation
time, so introducing uniq_keys for validation.

Also we can conflict correctly between different action types this way.


hennevogel (changes requested):
Please also remove the `validates :type, uniqueness` on BsRequestAction as it'll not happen anymore with this...

src/api/app/models/bs_request.rb
```diff
@@ -1151,12 +1152,16 @@ def change_priorities?(new_priority)
...
   def check_bs_request_actions!(opts = {})
+    uniq_keys = []
     bs_request_actions.each do |action|
+      uniq_keys << action.uniq_key
       action.check_action_permission!(opts[:skip_source])
       action.check_for_expand_errors!(!@addrevision.nil?)
       raisepriority(action.minimum_priority)
     end
 
+    raise ConflictingActions, 'Conflicting Actions' if uniq_keys.length > uniq_keys.uniq.length
```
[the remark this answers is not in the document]
adrianschroeter: yes, done

src/api/app/models/bs_request_action.rb
```diff
@@ -824,6 +818,10 @@ def set_sourceupdate_default(user)
     update(sourceupdate: 'cleanup') if target_project && user.branch_project_name(target_project) == source_project
   end
 
+  def uniq_key
+    ' '
```
krauselukas, line 822:
```suggestion
    raise AbstractMethodCalled
```

hennevogel: I don't think this is a good idea. This runs in a `before_validation`, so basically on instantiation. 

Pushing this out to individual `BsRequestAction` validations would be a ton of work I would like to avoid here.

krauselukas: alright

krauselukas (approved):

PR #12476, opened 2022-04-22
Add instrumentation to track changes on Notification#delivered

vpereira:
If a Notification is delivered, we want to track it within our SRE stack

PR #12582, opened 2022-05-13
Remove duplicated metrics for notifications

saraycp:
Since PR #12488 was created we no longer need the code introduced in PR #12476.

We are sending similar metrics from the notifications controller because the bulk update does not trigger callbacks. So we can get rid of the code in the model.


danidoni (approved):

saraycp:
I close this PR in favour of: https://github.com/openSUSE/open-build-service/pull/12586

PR #12586, opened 2022-05-16
Fix metrics for notifications

saraycp:
We want to track the notifications that were delivered (marked as read), not those that changed the 'delivered' field to any value. That's why the instrumentation needed to be adapted.

We now have to perform two bulk updates to be able to distinguish the notifications that are marked as read and those marked as unread.

Please read the commits descriptions for more details.

### DIFF
--- a/src/api/app/controllers/webui/users/notifications_controller.rb
+++ b/src/api/app/controllers/webui/users/notifications_controller.rb
@@ -21,13 +21,14 @@ class Webui::Users::NotificationsController < Webui::WebuiController
                       fetch_notifications.where(id: params[:notification_ids])
                     end
     # rubocop:disable Rails/SkipsModelValidations
-    updated_notifications = notifications.update_all('delivered = !delivered')
+    read_count = notifications.where(delivered: false).update_all('delivered = !delivered')
+    unread_count = notifications.where(delivered: true).update_all('delivered = !delivered')
     # rubocop:enable Rails/SkipsModelValidations
 
-    if updated_notifications.zero?
+    if read_count.zero? && unread_count.zero?
       flash.now[:error] = "Couldn't update the notifications"
     else
-      send_notifications_information_rabbitmq(updated_notifications)
+      send_notifications_information_rabbitmq(read_count, unread_count)
     end
 
     respond_to do |format|
@@ -91,8 +92,8 @@ class Webui::Users::NotificationsController < Webui::WebuiController
     fetch_notifications.count > Notification::MAX_PER_PAGE
   end
 
-  def send_notifications_information_rabbitmq(number_of_records)
-    RabbitmqBus.send_to_bus('metrics',
-                            "notification.delivered,web=true,rss=false value=#{number_of_records}")
+  def send_notifications_information_rabbitmq(read_count, unread_count)
+    RabbitmqBus.send_to_bus('metrics', "notification,action=read value=#{read_count}") if read_count.positive?
+    RabbitmqBus.send_to_bus('metrics', "notification,action=unread value=#{unread_count}") if unread_count.positive?
   end
 end
--- a/src/api/app/models/notification.rb
+++ b/src/api/app/models/notification.rb
@@ -54,9 +54,11 @@ class Notification < ApplicationRecord
                             "notification.create,notifiable_type=#{notifiable_type},web=#{web},rss=#{rss} value=1")
   end
 
+  # This is only called when the request come from the API. The UI performs 'update_all' that does not trigger callbacks.
+  # This metrics complements the same metrics tracked in Webui::Users::NotificationsController#send_notifications_information_rabbitmq
   def track_notification_delivered
     RabbitmqBus.send_to_bus('metrics',
-                            "notification.delivered,notifiable_type=#{notifiable_type},web=#{web},rss=#{rss} value=1")
+                            "notification,action=#{delivered ? 'read' : 'unread'} value=1")
   end
 end
 
--- a/src/api/spec/models/notification_spec.rb
+++ b/src/api/spec/models/notification_spec.rb
@@ -68,31 +68,31 @@ RSpec.describe Notification do
 
   describe 'Instrumentation' do
     let!(:test_user) { create(:confirmed_user, login: 'foo') }
-    let!(:rss_notification) { create(:rss_notification, subscriber: test_user) }
+    let!(:web_notification) { create(:web_notification, subscriber: test_user) }
 
     before do
-      allow(RabbitmqBus).to receive(:send_to_bus).with('metrics', 'notification.delivered,notifiable_type=,web=false,rss=true value=1')
+      allow(RabbitmqBus).to receive(:send_to_bus).with('metrics', 'notification,action=read value=1')
     end
 
     context 'if delivered change, we should track it' do
       before do
-        rss_notification.delivered = true
+        web_notification.delivered = true
       end
 
       it do
-        rss_notification.save
-        expect(RabbitmqBus).to have_received(:send_to_bus).with('metrics', 'notification.delivered,notifiable_type=,web=false,rss=true value=1')
+        web_notification.save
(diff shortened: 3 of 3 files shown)


src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -21,13 +21,14 @@ def update
                       fetch_notifications.where(id: params[:notification_ids])
                     end
     # rubocop:disable Rails/SkipsModelValidations
-    updated_notifications = notifications.update_all('delivered = !delivered')
+    number_of_marked_as_read = notifications.where(delivered: false).update_all('delivered = !delivered')
```
danidoni, line 24: What do you think of `read_count`? Is smaller, but do you think it conveys the same meaning?

saraycp: I think it can work

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -91,8 +92,8 @@ def show_read_all_button?
     fetch_notifications.count > Notification::MAX_PER_PAGE
   end
 
-  def send_notifications_information_rabbitmq(number_of_records)
+  def send_notifications_information_rabbitmq(number_of_marked_as_read)
     RabbitmqBus.send_to_bus('metrics',
-                            "notification.delivered,web=true,rss=false value=#{number_of_records}")
+                            "notification.marked_as_read,web=false,rss=true number_of_marked_as_read=#{number_of_marked_as_read}")
```
hennevogel, line 97:
```suggestion
                            "notification.marked_as_read,web=false,rss=true value=#{number_of_marked_as_read}")
```

src/api/app/models/notification.rb
```diff
@@ -54,9 +54,11 @@ def track_notification_creation
                             "notification.create,notifiable_type=#{notifiable_type},web=#{web},rss=#{rss} value=1")
   end
 
+  # This is only called when the request come from the API. The UI performs 'update_all' that does not trigger callbacks.
+  # This metrics complements the same metrics tracked in Webui::Users::NotificationsController#send_notifications_information_rabbitmq
   def track_notification_delivered
     RabbitmqBus.send_to_bus('metrics',
-                            "notification.delivered,notifiable_type=#{notifiable_type},web=#{web},rss=#{rss} value=1")
+                            "notification.marked_as_read,web=#{web},rss=#{rss} number_of_marked_as_read=1")
```
hennevogel, line 61:
```suggestion
                            "notification.marked_as_read,web=#{web},rss=#{rss} value=1")
```

hennevogel (changes requested):

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -91,8 +92,8 @@ def show_read_all_button?
     fetch_notifications.count > Notification::MAX_PER_PAGE
   end
 
-  def send_notifications_information_rabbitmq(number_of_records)
+  def send_notifications_information_rabbitmq(number_of_marked_as_read)
     RabbitmqBus.send_to_bus('metrics',
-                            "notification.delivered,web=true,rss=false value=#{number_of_records}")
+                            "notification.marked_as_read,web=false,rss=true value=#{number_of_marked_as_read}")
```
hennevogel, line 97: I would suggest we switch to a `action` tag instead of hard coding this into the series name. Makes it easier to mix this into one query. There is also *no* way to mark an RSS notification as read so we don't need to track this at all. So like:

```suggestion
                            "notification,action=read value=#{number_of_marked_as_read}")
```

src/api/app/models/notification.rb
```diff
@@ -54,9 +54,11 @@ def track_notification_creation
                             "notification.create,notifiable_type=#{notifiable_type},web=#{web},rss=#{rss} value=1")
   end
 
+  # This is only called when the request come from the API. The UI performs 'update_all' that does not trigger callbacks.
+  # This metrics complements the same metrics tracked in Webui::Users::NotificationsController#send_notifications_information_rabbitmq
   def track_notification_delivered
     RabbitmqBus.send_to_bus('metrics',
-                            "notification.delivered,notifiable_type=#{notifiable_type},web=#{web},rss=#{rss} value=1")
+                            "notification.marked_as_read,web=#{web},rss=#{rss} value=1")
```
hennevogel, line 61:
```suggestion
                            "notification,action=read value=1")
```

src/api/spec/models/notification_spec.rb
```diff
@@ -71,7 +71,7 @@
     let!(:rss_notification) { create(:rss_notification, subscriber: test_user) }
 
     before do
-      allow(RabbitmqBus).to receive(:send_to_bus).with('metrics', 'notification.delivered,notifiable_type=,web=false,rss=true value=1')
+      allow(RabbitmqBus).to receive(:send_to_bus).with('metrics', 'notification.marked_as_read,web=false,rss=true number_of_marked_as_read=1')
```
hennevogel, line 74:
```suggestion
      allow(RabbitmqBus).to receive(:send_to_bus).with('metrics', 'notification,action=read value=1')
```

src/api/spec/models/notification_spec.rb
```diff
@@ -92,7 +92,7 @@
 
       it do
         rss_notification.save
-        expect(RabbitmqBus).not_to have_received(:send_to_bus).with('metrics', 'notification.delivered,notifiable_type=,web=false,rss=true value=1')
+        expect(RabbitmqBus).not_to have_received(:send_to_bus).with('metrics', 'notification.marked_as_read,web=false,rss=true number_of_marked_as_read=1')
```
hennevogel, line 95:
```suggestion
        expect(RabbitmqBus).not_to have_received(:send_to_bus).with('metrics', 'notification,action=read value=1')
```

src/api/app/controllers/webui/users/notifications_controller.rb
hennevogel: Same way you could track the other number (number_of_marked_as_unread)...

hennevogel (changes requested):

hennevogel:
Sorry for the back and forth...

src/api/app/controllers/webui/users/notifications_controller.rb
```diff
@@ -91,8 +92,8 @@ def show_read_all_button?
     fetch_notifications.count > Notification::MAX_PER_PAGE
   end
 
-  def send_notifications_information_rabbitmq(number_of_records)
-    RabbitmqBus.send_to_bus('metrics',
-                            "notification.delivered,web=true,rss=false value=#{number_of_records}")
+  def send_notifications_information_rabbitmq(read_count, unread_count)
+    RabbitmqBus.send_to_bus('metrics', "notification,action=read,web=true,rss=false value=#{read_count}") if read_count.positive?
+    RabbitmqBus.send_to_bus('metrics', "notification,action=unread,web=true,rss=false value=#{unread_count}") if unread_count.positive?
```
hennevogel, line 97: There is no need for the `web=true,rss=false` tags anymore after #12594

src/api/app/models/notification.rb
```diff
@@ -54,9 +54,11 @@ def track_notification_creation
                             "notification.create,notifiable_type=#{notifiable_type},web=#{web},rss=#{rss} value=1")
   end
 
+  # This is only called when the request come from the API. The UI performs 'update_all' that does not trigger callbacks.
+  # This metrics complements the same metrics tracked in Webui::Users::NotificationsController#send_notifications_information_rabbitmq
   def track_notification_delivered
     RabbitmqBus.send_to_bus('metrics',
-                            "notification.delivered,notifiable_type=#{notifiable_type},web=#{web},rss=#{rss} value=1")
+                            "notification,action=#{delivered ? 'read' : 'unread'},web=#{web},rss=#{rss} value=1")
```
hennevogel, line 61: Same here

hennevogel (changes requested):

vpereira (approved):

hennevogel (approved):

PR #12591, opened 2022-05-17
Introduce tags for notification counts

hennevogel:
We only used fields so far which makes it very hard/slow to query.
Tags are also way easier to mix/match in queries.


hennevogel:
This will need changes in the dashboards, I'll do them once this is merged.

danidoni (approved):

PR #12597, opened 2022-05-18
[api] support request approve in new state

adrianschroeter:
This avoids necessary checking and possible races when doing the final
approve of a request. Accepting an approve and triggering the request
accept in case all reviews are done already directly

### DIFF
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -21,6 +21,8 @@ class BsRequest < ApplicationRecord
 
   OBSOLETE_STATES = [:declined, :superseded, :revoked].freeze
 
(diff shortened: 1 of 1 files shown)


src/api/app/models/bs_request.rb
```diff
@@ -636,7 +636,7 @@ def assignreview(opts = {})
   end
 
   def approval_handling(new_approver, opts)
-    raise InvalidStateError, 'request is not in review state' unless state == :review
+    raise InvalidStateError, 'request is not in review state' unless state == :review || state == :new
```
vpereira, line 639: `raise InvalidStateError, 'request is not in review state' unless [:review, :new].include?(state)`

PR #12598, opened 2022-05-18
Extend event id column capacity

saraycp:
The events id column overflowed, so we had to change the column type from `integer` to `bigint`.

We did this on the fly in production, since the overflow was causing problems. Now we add the corresponding migration to the code.


src/api/db/schema.rb
```diff
@@ -425,8 +425,8 @@
     t.index ["workflow_run_id"], name: "index_event_subscriptions_on_workflow_run_id"
   end
 
-  create_table "events", id: :integer, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", options: "ENGINE=InnoDB ROW_FORMAT=DYNAMIC", force: :cascade do |t|
-    t.string "eventtype", null: false, collation: "utf8_unicode_ci"
+  create_table "events", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", options: "ENGINE=InnoDB ROW_FORMAT=DYNAMIC", force: :cascade do |t|
+    t.string "eventtype", null: false, collation: "utf8mb3_unicode_ci"
```
saraycp, line 429: The id parameter disappeared from this line. Is this normal?

I checked in `dbconsole` after running the migration and  the `bigint` is there:

```
MariaDB [api_development]> describe events;                                                                                                    
+-------------+--------------+------+-----+---------+----------------+
| Field       | Type         | Null | Key | Default | Extra          |
+-------------+--------------+------+-----+---------+----------------+
| id          | bigint(20)   | NO   | PRI | NULL    | auto_increment |
| eventtype   | varchar(255) | NO   | MUL | NULL    |                |
| payload     | text         | YES  |     | NULL    |                |
| created_at  | datetime     | YES  | MUL | NULL    |                |
| updated_at  | datetime(6)  | YES  |     | NULL    |                |
| undone_jobs | int(11)      | YES  |     | 0       |                |
| mails_sent  | tinyint(1)   | YES  | MUL | 0       |                |
+-------------+--------------+------+-----+---------+----------------+

```

src/api/db/migrate/20220518094002_increase_event_id_column_capacity.rb
```diff
@@ -0,0 +1,9 @@
+class IncreaseEventIdColumnCapacity < ActiveRecord::Migration[6.1]
+  def up
+    safety_assured { change_column :events, :id, :bigint, auto_increment: true }
+  end
+
+  def down
+    safety_assured { change_column :events, :id, :integer, auto_increment: true }
```
hennevogel, line 7: This is irreversible IMHO.

src/api/app/models/issue_tracker.rb
```diff
@@ -349,6 +349,7 @@ def update_package_meta
 #  label          :text(65535)      not null
 #  name           :string(255)      not null
 #  password       :string(255)
+#  publish_issues :boolean          default(TRUE)
```
hennevogel, line 352: Not related but from #11351

hennevogel (changes requested):

src/api/db/schema.rb
hennevogel: > The id parameter disappeared from this line. Is this normal?

What happens if you create a new table?

src/api/db/migrate/20220518094002_increase_event_id_column_capacity.rb
saraycp: I wonder if it's better to raise an exception or simply do nothing.

```
  def down
    raise ActiveRecord::IrreversibleMigration
  end
```
VS 

```
  def down; end
```

danidoni: I would raise the exception

hennevogel: https://guides.rubyonrails.org/active_record_migrations.html#using-reversible

src/api/db/schema.rb
```diff
@@ -10,10 +10,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_04_13_223301) do
+ActiveRecord::Schema.define(version: 2022_05_18_094002) do
 
   create_table "architectures", id: :integer, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", options: "ENGINE=InnoDB ROW_FORMAT=DYNAMIC", force: :cascade do |t|
-    t.string "name", null: false, collation: "utf8_general_ci"
+    t.string "name", null: false, collation: "utf8mb3_general_ci"
```
hennevogel, line 16: Please do not include those changes here. They might stem from a mariadb update in the dev-env.

hennevogel (changes requested):

saraycp:
@hennevogel @danidoni I've pushed some changes, please have another look:

- raise the exception in the migration's `down` method.
- no changes on encoding are introduced
- the only change in the schema.rb, is removing the `id` parameter for `events` table.

hennevogel (approved):

saraycp:
I pushed again just to remove the annotation regarding `publish_issues`, nothing to do with the current migration.

PR #12606, opened 2022-05-19
Catch faraday connection exception in SCM reporter

krauselukas:
Octokit raises this exception when the connection
to github timesout or fails.

### DIFF
--- a/src/api/app/services/scm_status_reporter.rb
+++ b/src/api/app/services/scm_status_reporter.rb
@@ -36,6 +36,8 @@ class SCMStatusReporter < SCMExceptionHandler
     @workflow_run.update_to_fail("Failed to report back to GitHub: #{e.message}") if @workflow_run.present?
   rescue Octokit::Error, Gitlab::Error::Error => e
     rescue_with_handler(e) || raise(e)
+  rescue Faraday::ConnectionFailed => e
+    @workflow_run.update_to_fail("Failed to report back to GitHub: #{e.message}")
   end
 
   private


krauselukas:
[Catch Faraday (connection) exceptions in SCMExceptionHandler (M)](https://trello.com/c/wF4zkRY9/1894-catch-faraday-connection-exceptions-in-scmexceptionhandler-m)

src/api/app/services/scm_status_reporter.rb
```diff
@@ -36,6 +36,8 @@ def call
     @workflow_run.update_to_fail("Failed to report back to GitHub: #{e.message}") if @workflow_run.present?
   rescue Octokit::Error, Gitlab::Error::Error => e
     rescue_with_handler(e) || raise(e)
+  rescue Faraday::ConnectionFailed
+    @workflow_run.update_to_fail('Failed to report back to GitHub: Could not connect to the GitHub API')
```
rubhanazeem, line 40:
```
rescue Faraday::ConnectionFailed => e
    @workflow_run.update_to_fail("Failed to report back to GitHub: #{e.message}")
```
Have you looked into the exception message? Is it something meaningful?

krauselukas: @rubhanazeem yes, looking at what we received in errbit its meaningful `Failed to open TCP connection to api.github.com:443 (Connection timed out - user specified timeout)`
I show the message that comes from faraday now

rubhanazeem (approved):

PR #12610, opened 2022-05-19
Merge request response tab into incoming webhook

danidoni:
The content of the request and response tabs belongs to the workflow and
it's better shown together.

__This is how it looks like:__
image

### DIFF
--- a/src/api/app/components/workflow_run_detail_component.html.haml
+++ b/src/api/app/components/workflow_run_detail_component.html.haml
@@ -1,33 +1,27 @@
 %ul.nav.nav-tabs#workflow-run-tabs{ role: 'tablist' }
   %li.nav-item{ role: 'presentation' }
-    %a.nav-link.active{ id: "request-tab#{id}", data: { toggle: 'tab' }, href: "#request-tab-content#{id}",
-                        role: 'tab', aria: { controls: "request-tab-content#{id}", selected: 'true' } }
-      Request
-  %li.nav-item{ role: 'presentation' }
-    %a.nav-link{ id: "response-tab#{id}", data: { toggle: 'tab' }, href: "#response-tab-content#{id}",
-                  role: 'tab', aria: { controls: "response-tab-content#{id}", selected: 'false' } }
-      Response
+    %a.nav-link.active{ id: "incoming-webhook-tab#{id}", data: { toggle: 'tab' }, href: "#incoming-webhook-tab-content#{id}",
+                        role: 'tab', aria: { controls: "incoming-webhook-tab-content#{id}", selected: 'true' } }
+      Webhook from #{scm_vendor}
   %li.nav-item{ role: 'presentation' }
     %a.nav-link{ id: "artifacts#{id}", data: { toggle: 'tab' }, href: "#artifacts-tab-content#{id}",
                  role: 'tab', aria: { controls: "artifacts-tab-content#{id}", selected: 'false' } }
       Artifacts
 .tab-content.p-3#workflow-run-tabs-content
-  .tab-pane.fade.show.active{ id: "request-tab-content#{id}", role: 'tabpanel',
-                              aria: { labelledby: "request-tab#{id}" } }
-    %h5 Request Headers
-    %pre.border.p-2#request-headers
-      = request_headers
-    %h5 Request Payload
-    %pre.border.p-2#request-payload
-      = pretty_request_payload
-  .tab-pane.fade{ id: "response-tab-content#{id}", role: 'tabpanel',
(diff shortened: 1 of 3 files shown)


src/api/app/components/workflow_run_detail_component.html.haml
```diff
@@ -1,33 +1,27 @@
...
-    %a.nav-link.active{ id: "request-tab#{id}", data: { toggle: 'tab' }, href: "#request-tab-content#{id}",
-                        role: 'tab', aria: { controls: "request-tab-content#{id}", selected: 'true' } }
-      Request
-  %li.nav-item{ role: 'presentation' }
-    %a.nav-link{ id: "response-tab#{id}", data: { toggle: 'tab' }, href: "#response-tab-content#{id}",
-                  role: 'tab', aria: { controls: "response-tab-content#{id}", selected: 'false' } }
-      Response
+    %a.nav-link.active{ id: "incoming-webhook-tab#{id}", data: { toggle: 'tab' }, href: "#incoming-webhook-tab-content#{id}",
+                        role: 'tab', aria: { controls: "incoming-webhook-tab-content#{id}", selected: 'true' } }
+      Incoming Webhook
```
hennevogel, line 5: Do we know the type of SCM at this point?

danidoni: Yes, we do

hennevogel: Then I would suggest `Webhook from $SCM`

danidoni:
[Separate webhook from SCMStatusReporter responses (L)](https://trello.com/c/lc0mYXmy/1891-separate-webhook-from-scmstatusreporter-responses-l)

saraycp (commented):
In general, it looks good to me. But there is something that worries me: the payload is usually **too** long, so I'm afraid "Webhook Response Body" can go unnoticed.

Possible solutions that come to my mind:
- change the order
- make the payload collapsible

src/api/app/components/workflow_run_detail_component.html.haml
```diff
@@ -1,33 +1,27 @@
...
       Artifacts
 .tab-content.p-3#workflow-run-tabs-content
-  .tab-pane.fade.show.active{ id: "request-tab-content#{id}", role: 'tabpanel',
-                              aria: { labelledby: "request-tab#{id}" } }
-    %h5 Request Headers
+  .tab-pane.fade.show.active{ id: "incoming-webhook-tab-content#{id}", role: 'tabpanel',
+                              aria: { labelledby: "incoming-webhook-tab#{id}" } }
+    %h5 Webhook URL
+    %pre.border.p-2#response-url
+      = response_url # Change this to incoming webhook url
```
rubhanazeem, line 15: Do you want to remove `Change this to incoming webhook url`?

src/api/app/components/workflow_run_detail_component.html.haml
```diff
@@ -1,33 +1,27 @@
...
+    %h5 Webhook Request Headers
     %pre.border.p-2#request-headers
-      = request_headers
-    %h5 Request Payload
+      = request_headers # Change this to incoming_webhook_headers
```
rubhanazeem, line 18: Same as above

src/api/app/components/workflow_run_detail_component.html.haml
```diff
@@ -1,33 +1,27 @@
...
-                  aria: { labelledby: "response-tab#{id}" } }
-    %h5 Response URL
-    %pre.border.p-2#response-url
-      = response_url
-    %h5 Response Body
+      = pretty_request_payload # Change this to incoming webhook payload
+    %h5 Webhook Response Body
     %pre.border.p-2#response-body
-      = response_body
+      = response_body # change this to incoming webhook body
```
rubhanazeem, line 24: And here

danidoni:
> In general, it looks good to me. But there is something that worries me: the payload is usually **too** long, so I'm afraid "Webhook Response Body" can go unnoticed.
> 
> Possible solutions that come to my mind:
> 
>     * change the order
> 
>     * make the payload collapsible

Hmm, when the workflow run fails you probably wanna see the response first, to see what went wrong, and then check what did you send to make it fail, so yeah, I'll reorder them.

src/api/app/components/workflow_run_detail_component.html.haml
danidoni: True!

saraycp (approved):

rubhanazeem (approved):

PR #12612, opened 2022-05-19
Add error messages for SCM exception handler

rubhanazeem:
We receive some exceptions from GitHub and GitLab. The error messages in these exceptions are very general. This PR proposes new error messages to help users in debugging process.

### DIFF
--- a/src/api/app/services/scm_exception_handler.rb
+++ b/src/api/app/services/scm_exception_handler.rb
@@ -51,6 +51,6 @@ class SCMExceptionHandler
   private
 
   def log_to_workflow_run(exception, scm)
-    @workflow_run.update_to_fail("Failed to report back to #{scm}: #{exception.message}")
+    @workflow_run.update_to_fail("Failed to report back to #{scm}: #{ScmExceptionMessage.for(exception: exception, scm: scm)}")
   end
 end
--- a/src/api/app/services/scm_exception_message.rb
+++ b/src/api/app/services/scm_exception_message.rb
@@ -1,0 +1,83 @@
+class ScmExceptionMessage
+  GITHUB_EXCEPTIONS = {
+    Octokit::AbuseDetected =>
+      'You have triggered an abuse detection mechanism and have been temporarily blocked from content creation. Please retry your request again later.',
+    Octokit::AccountSuspended =>
+      'Sorry. Your account is suspended.',
+    Octokit::BillingIssue =>
+      'The repository has been disabled due to a billing issue with the owner account.',
+    Octokit::BranchNotProtected =>
+      'This branch is not protected.',
+    Octokit::Forbidden =>
+      'Request is forbidden.',
+    Octokit::RepositoryUnavailable =>
+      'Repository access is blocked.',
+    Octokit::NotFound =>
+      'Content not found.',
+    Octokit::OneTimePasswordRequired =>
+      'Must specify two-factor authentication OTP code.',
+    Octokit::Unauthorized =>
+      'Unauthorized request. Please check your credentials again.',
+    Octokit::UnavailableForLegalReasons =>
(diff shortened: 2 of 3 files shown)


src/api/app/helpers/webui/scm_exception_messages_helper.rb
```diff
@@ -0,0 +1,90 @@
...
+  def exception_message(exception:, scm:)
+    case scm
+    when 'GitHub'
+      github_exception_message(exception: exception)
+    when 'GitLab'
+      gitlab_exception_message(exception: exception)
+    end
+  end
+
+  # rubocop:disable Metrics/CyclomaticComplexity
```
danidoni, line 11: Maybe you could split this class into three: `Webui::ScmExceptionMessageHelper`, one for Github errors and another for GitLab errors. Then every vendor has their errors in one class and this class is smaller

src/api/app/services/scm_exception_handler.rb
```diff
@@ -51,6 +52,6 @@ def initialize(event_payload, event_subscription_payload, scm_token, workflow_ru
   private
 
   def log_to_workflow_run(exception, scm)
-    @workflow_run.update_to_fail("Failed to report back to #{scm}: #{exception.message}")
+    @workflow_run.update_to_fail("Failed to report back to #{scm}: #{exception_message(exception: exception, scm: scm)}")
```
danidoni, line 55: Instead of including the Module, you can rewrite it as a class and ask for something like `ScmExceptionMessages.for(scm).explanation(exception)`.

src/api/app/helpers/webui/scm_exception_messages_helper.rb
rubhanazeem: This is because of the long method. Splitting this into separate classes will still produce this warning because the methods' size will remain the same.

src/api/app/services/scm_exception_handler.rb
rubhanazeem: that sounds good. I'll update it.

danidoni:
One idea to fix the [CodeClimate](https://codeclimate.com/github/openSUSE/open-build-service/pull/12612) issue is to transform the `case` into a hash with exception class names as keys and exception messages as values

src/api/app/services/scm_exception_message.rb
```diff
@@ -0,0 +1,83 @@
+class ScmExceptionMessage
+  GITHUB_EXCEPTIONS = {
+    Octokit::AbuseDetected =>
+      'You have triggered an abuse detection mechanism and have been temporarily blocked from content creation. Please retry your request again later.',
+    Octokit::AccountSuspended =>
+      'Sorry. Your account is suspended.',
+    Octokit::BillingIssue =>
+      'The repository has been disabled due to a billing issue with the owner account.',
+    Octokit::BranchNotProtected =>
+      'This branch not protected.',
```
danidoni, line 10:
```suggestion
      'This branch is not protected.',
```

src/api/app/services/scm_exception_message.rb
```diff
@@ -0,0 +1,83 @@
...
+    Octokit::Forbidden =>
+      'Request forbidden.',
```
danidoni, line 12:
```suggestion
      'Request is forbidden.',
```

src/api/app/services/scm_exception_message.rb
```diff
@@ -0,0 +1,83 @@
...
+    Octokit::RepositoryUnavailable =>
+      'Repository access is blocked.',
+    Octokit::NotFound =>
+      'Content does not found.',
```
danidoni, line 16:
```suggestion
      'Content not found.',
```

danidoni (approved):

PR #12620, opened 2022-05-23
Update table schema

danidoni:
Make the charset explicit. utf8 is an alias for utf8mb3.

See https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8mb3.html and https://mariadb.com/kb/en/unicode/


danidoni:
> Is this also true for mariadb? :)

Yes, it is:

> No supplementary characters are stored. Until [MariaDB 10.5](https://mariadb.com/kb/en/what-is-mariadb-105/), this was an alias for utf8. From [MariaDB 10.6](https://mariadb.com/kb/en/what-is-mariadb-106/), utf8 is by default an alias for utf8mb3.

See https://mariadb.com/kb/en/unicode/

hennevogel (approved):

PR #12647, opened 2022-05-30
Adopt rack version to last update

hennevogel:


hennevogel:
skipping CI, no code changes, to speed up release.

PR #12652, opened 2022-05-31
[api] updateinfo counters must depend on maintenance project

adrianschroeter:
otherwise they are global and anyone can influence the counters in any maintenance
project.

This broke 2015 (1c9c7a0d207 ?)


src/api/app/models/updateinfo_counter.rb
```diff
@@ -1,10 +1,17 @@
...
     month = time.month if template =~ /%M/
     day   = time.day   if template =~ /%D/
 
-    UpdateinfoCounter.find_or_create_by(year: year, month: month, day: day)
+    # fix broken database entries without a defined maintenance project
+    # it used to be broken from 2015-2022
+    if (uc = UpdateinfoCounter.find_by(maintenance_db_project_id: nil, year: year, month: month, day: day))
+      uc.maintenance_db_project_id = maintenance_project.id
+      uc.save
+    end
```
hennevogel, line 12: This seems dangerous to me judging from production data. For instance on OBS we only have counters by year. How do you know which maintenance project it belongs to? There are 574 of them, 3 if you omit `home:`.

```
#<UpdateinfoCounter:0x00007f26914785b0 id: 30, maintenance_db_project_id: nil, day: nil, month: nil, year: 2022, counter: 10028>
```

And also `time` here is [always `Time.zone.now`](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/maintenance_incident.rb#L75) so you will never touch any `UpdateinfoCounter` other than the current one.

I don't think this data is salvageable without human interaction...

hennevogel: @adrianschroeter how about we do this manually on IBS/OBS and call it a day?

adrianschroeter:
On Donnerstag, 14. Juli 2022, 14:18:22 CEST Henne Vogelsang wrote:
> 
> 
> 
> > +    # fix broken database entries without a defined maintenance project
> +    # it used to be broken from 2015-2022
> +    if (uc = UpdateinfoCounter.find_by(maintenance_db_project_id: nil, year: year, month: month, day: day))
> +      uc.maintenance_db_project_id = maintenance_project.id
> +      uc.save
> +    end
> 
> @adrianschroeter how about we do this manually on IBS/OBS and call it a day?

there are also instances out there which using the maintenance workflow, I would not ignore them

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> there are also instances out there which using the maintenance workflow, I would not ignore them

Then the only thing we can do is some interactive rake task that asks and mention this in the Changelog...

PR #12678, opened 2022-06-07
Add configuration option to prevent adding a maintainer when creating a project

eduardoj:
Currently, calling the `PUT /source/:project/_meta` API endpoint on a non-existent project adds the user who calls the endpoint as a maintainer of the project. In the future, we want to modify the behavior of this endpoint, by not adding the user as a maintainer. This means that we introduce a backward incompatible change in the API.

To introduce this change smoothly, we add a configuration option that will allow OBS instances administrators to decide when they want to adopt this new behavior.

### DIFF
--- a/src/api/app/controllers/source_project_meta_controller.rb
+++ b/src/api/app/controllers/source_project_meta_controller.rb
@@ -70,10 +70,12 @@ class SourceProjectMetaController < SourceController
       else
         project = Project.new(name: @project_name)
         project.update_from_xml!(@request_data)
-        # FIXME3.0: don't modify send data
-        maintainer_role = Role.find_by_title!('maintainer')
-        unless project.relationships.any? { |relationship| relationship.user == User.session! && relationship.role == maintainer_role }
-          project.relationships.find_or_initialize_by(user: User.session!, role: maintainer_role)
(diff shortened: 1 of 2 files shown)


adrianschroeter (approved):

src/api/app/controllers/source_project_meta_controller.rb
```diff
@@ -70,10 +70,12 @@ def update
       else
         project = Project.new(name: @project_name)
         project.update_from_xml!(@request_data)
-        # FIXME3.0: don't modify send data
-        maintainer_role = Role.find_by_title!('maintainer')
-        unless project.relationships.any? { |relationship| relationship.user == User.session! && relationship.role == maintainer_role }
-          project.relationships.find_or_initialize_by(user: User.session!, role: maintainer_role)
+        unless CONFIG['prevent_adding_maintainer_in_project_creation_with_api']
```
hennevogel, line 73: People tend to expect that those options are on/off switches. So they will change your example to `= true` and wonder why it isn't working. `Configuration` has some support for this, check it out.

hennevogel (changes requested):

hennevogel (approved):

PR #12680, opened 2022-06-08
Share workflow tokens

krauselukas:
After the attempt to share the workflow tokens with other users through the `Relationship` model I decided to not follow this path. Its more complicated than creating a new join table to do so. IMO the best would be to create a new table for workflow tokens in order to decouple it from the other tokens types.  

This is what the UI looks like:

token_users_links

Screenshot 2022-06-10 at 11-07-30 Open Build Service

TODO:

- [x] Adapting the pundit policies
- [x] User interface to add users to the workflow tokens
- [x] Specs
- [x] Show who is `Token.user` in the UI
- [x] Finish the group part
- [x] Apply policies to the new controller to prevent users that are not associated with the token to manage users.
- [x] Fix authentication/policy to allow users that are not `Token.user` to edit, show,  etc the token.
- TO DO LATER:
- [ ] Rename associations and adapt the code affected to use the new name
- [ ] Stop using the two scopes in the policy `scope.owned_tokens(user) + scope.shared_tokens(user)`
- [ ] Create data migration to transfer existing workflow tokens to the join table
- [ ] Use Ajax when adding/removing users from the token's users page. Can be done in another PR.


src/api/app/models/token.rb
```diff
@@ -17,6 +17,9 @@ class Token < ApplicationRecord
 
   include Token::Errors
 
+  scope :owned_tokens, ->(user) { where(user: user).where.not(type: ['Token::Rss']).includes(package: :project) }
+  scope :shared_tokens, ->(user) { user.workflow_tokens.where.not(type: ['Token::Rss']).includes(package: :project) }
```
saraycp, line 21: `user.workflow_tokens` only returns instances of `Token::Workflow`, so it makes no sense to add `where.not(type: ['Token::Rss'])`

src/api/app/models/workflow_token_user.rb
```diff
@@ -0,0 +1,4 @@
+class WorkflowTokenUser < ApplicationRecord
+  belongs_to :user
+  belongs_to :token, class_name: 'Token::Workflow'
+end
```
saraycp, line 4: In general, I miss the relationship with OBS groups as well.

src/api/app/models/user.rb
```diff
@@ -24,6 +24,10 @@ class User < ApplicationRecord
   has_many :comments, dependent: :destroy, inverse_of: :user
   has_many :status_messages
   has_many :tokens, class_name: 'Token', dependent: :destroy, inverse_of: :user
+  
+  has_many :workflow_token_users, dependent: :destroy
+  has_many :workflow_tokens, class_name: 'Token::Workflow', through: :workflow_token_users, source: :token
```
hennevogel, line 29: Why "has many through"? What are the plans with the in-between model?

Sounds to me like [has and belongs to many](https://guides.rubyonrails.org/association_basics.html#the-has-and-belongs-to-many-association) should be enough, no?

src/api/app/models/workflow_token_user.rb
danidoni: We'll tackle that later

src/api/app/models/user.rb
```diff
@@ -24,6 +24,14 @@ class User < ApplicationRecord
   has_many :comments, dependent: :destroy, inverse_of: :user
   has_many :status_messages
   has_many :tokens, class_name: 'Token', dependent: :destroy, inverse_of: :user
+
+  has_and_belongs_to_many :shared_workflow_tokens,
```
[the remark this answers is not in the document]
danidoni: I wanted to convey the meaning that those may not be **your** tokens, but someone else lent them to you.

src/api/app/models/workflow_token_user.rb
```diff
@@ -0,0 +1,17 @@
+class WorkflowTokenUser < ApplicationRecord
```
saraycp, line 1: Reminder to remove this model

danidoni: Done

src/api/app/models/user.rb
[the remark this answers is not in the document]
hennevogel: I think we agreed in the planing that there is no such distinction right? :)

src/api/app/controllers/webui/users/tokens/users_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    if @token.shared_among.include?(user)
+      redirect_back(fallback_location: root_path, notice: "User #{user.login} is already associated to the token")
+    elsif @token.shared_among << user
+      redirect_back(fallback_location: root_path, success: "User #{user.login} has been added to the token successfully")
+    end
+  end
+
+  def destroy
+    user = User.find(params[:id])
+    if @token.shared_among.destroy(user)
```
hennevogel, line 21: Uh, be careful. https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy

hennevogel (changes requested):

src/api/app/controllers/webui/users/tokens/users_controller.rb
```diff
@@ -0,0 +1,33 @@
+class Webui::Users::Tokens::UsersController < Webui::WebuiController
+  before_action :set_token
+
+  def index
+    @users = @token.shared_among
+    # TODO: @groups =
+  end
+
+  def create
```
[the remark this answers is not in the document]
danidoni: So, true, fixing now

src/api/app/controllers/webui/users/tokens/users_controller.rb
saraycp: This is not risky, it's just removing the entry in the join table. As you can see here:

```
[7] pry(main)> token.shared_among.destroy(user) 
  TRANSACTION (0.6ms)  BEGIN
  HABTM_SharedAmong Load (0.8ms)  SELECT `workflow_token_users`.* FROM `workflow_token_users` 
WHERE `workflow_token_users`.`token_id` = 1 AND `workflow_token_users`.`user_id` = 1
  Token::Workflow::HABTM_SharedAmong Destroy (0.3ms)  DELETE FROM `workflow_token_users` WHERE
 `workflow_token_users`.`token_id` = 1 AND `workflow_token_users`.`user_id` = 1
  TRANSACTION (1.5ms)  COMMIT
```

```
[14] pry(main)> token.shared_among.delete(user)                                               
  TRANSACTION (0.3ms)  BEGIN                                                                  
  Token::Workflow::HABTM_SharedAmong Destroy (0.3ms)  DELETE FROM `workflow_token_users` WHERE
 `workflow_token_users`.`token_id` = 1 AND `workflow_token_users`.`user_id` = 1               
  TRANSACTION (1.4ms)  COMMIT
  ```
  
  And as it's explained here:
  https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html

> What gets deleted?

> There is a potential pitfall here: [has_and_belongs_to_many](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_and_belongs_to_many) and [has_many](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many) :through associations have records in join tables, as well as the associated records. So when we call one of these deletion methods, what exactly should be deleted?

> The answer is that it is assumed that deletion on an association is about removing the link between the owner and the associated object(s), rather than necessarily the associated objects themselves. So with [has_and_belongs_to_many](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_and_belongs_to_many) and [has_many](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many) :through, the join records will be deleted, but the associated records won't.

> This makes sense if you think about it: if you were to call post.tags.delete(Tag.find_by(name: 'food')) you would want the 'food' tag to be unlinked from the post, rather than for the tag itself to be removed from the database.

Comment (saraycp) [reply src/api/app/controllers/webui/users/tokens/users_controller.rb]: I'm simply going to replace `destroy` by `delete` as we don't need to fire the callbacks.

File: src/api/app/controllers/webui/users/tokens/groups_controller.rb [hunk 8]
```diff
@@ -0,0 +1,23 @@
+class Webui::Users::Tokens::GroupsController < Webui::WebuiController
+  def create
```
hennevogel, line 2: Authorization?

src/api/app/controllers/webui/users/tokens/groups_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+    token = Token::Workflow.find(params[:token_id])
+    group = Group.find_by(title: params[:groupid])
+
+    if token.groups_shared_among.include?(group)
+      redirect_back(fallback_location: root_path, notice: "Group #{group.title} is already associated to the token")
+    elsif token.groups_shared_among << group
+      redirect_back(fallback_location: root_path, success: "Group #{group.title} has been added to the token successfully")
+    end
```
hennevogel, line 10: Not sure why you would error on this? Just don't do anything, what the user requested is already done...

src/api/app/controllers/webui/users/tokens/users_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    user = User.find_by(login: params[:userid])
+
+    if @token.users_shared_among.include?(user)
+      redirect_back(fallback_location: root_path, notice: "User #{user.login} is already associated to the token")
+    elsif @token.users_shared_among << user
+      redirect_back(fallback_location: root_path, success: "User #{user.login} has been added to the token successfully")
+    end
```
hennevogel, line 16: Same here basically. `if @token.users_shared_among.include?(user)` you are done. No need to error.

hennevogel (changes requested):

src/api/app/controllers/webui/users/tokens/groups_controller.rb
saraycp: This is not an error but a 'notice', just an informative blue flash message.

src/api/app/controllers/webui/users/tokens/users_controller.rb
saraycp: This is not an error but a 'notice', just an informative blue flash message.

src/api/app/controllers/webui/users/tokens/groups_controller.rb
saraycp: Do you still prefer to do nothing?

src/api/app/controllers/webui/users/tokens/users_controller.rb
saraycp: Do you still prefer to do nothing?

src/api/app/policies/token_policy.rb
```diff
@@ -31,7 +33,10 @@ def update?
 
   def create?
     # TODO: when trigger_workflow is rolled out, remove the Flipper check
-    record.user == user && record.type != 'Token::Rss' && Flipper.enabled?(:trigger_workflow, user)
+    return false unless Flipper.enabled?(:trigger_workflow, user)
+    return false unless record.type != 'Token::Rss'
+
+    record.user == user || record.users_shared_among.include?(user) || record.groups_shared_among.map(&:users).flatten.include?(user)
```
saraycp, line 39: We need to fix this, we use this policy for any type of token. And for tokens different from workflow, `users_shared_among` and `groups_shared_among` are not defined.

saraycp (changes requested):

src/api/app/controllers/webui/users/tokens/groups_controller.rb
hennevogel: Indeed. Less, more understandable code.

src/api/app/controllers/webui/users/tokens/users_controller.rb
hennevogel: See above.

src/api/app/controllers/webui/users/tokens/groups_controller.rb
danidoni: Done

src/api/app/controllers/webui/users/tokens/groups_controller.rb
```diff
@@ -1,23 +1,29 @@
 class Webui::Users::Tokens::GroupsController < Webui::WebuiController
+  before_action :set_token
+
   def create
-    token = Token::Workflow.find(params[:token_id])
+    authorize @token
     group = Group.find_by(title: params[:groupid])
```
hennevogel, line 6: This might be nil and sounds like a `before_action`

hennevogel (changes requested):

src/api/app/policies/token_policy.rb
```diff
@@ -31,7 +33,10 @@ def update?
 
   def create?
     # TODO: when trigger_workflow is rolled out, remove the Flipper check
-    record.user == user && record.type != 'Token::Rss' && Flipper.enabled?(:trigger_workflow, user)
+    return false unless Flipper.enabled?(:trigger_workflow, user)
+    return false unless record.type != 'Token::Rss'
```
hennevogel, line 37: 👍🏿

src/api/spec/controllers/webui/users/tokens/groups_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      end
+
+      subject! do
+        post :create, params: { groupid: other_group, token_id: workflow_token }
+      end
+
+      it { is_expected.to redirect_to(token_users_path(workflow_token)) }
+      it { expect(flash[:success]).to be_present }
+      it { expect(workflow_token.groups_shared_among).to match_array([token_group, other_group]) }
+    end
```
hennevogel, line 40: You have no logic for this in the controller anymore. Remove the test, CI time is developer time :)

src/api/spec/controllers/webui/users/tokens/groups_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      subject! do
+        post :create, params: { groupid: other_group, token_id: workflow_token }
+      end
+
+      it { is_expected.to redirect_to(root_path) }
+      it { expect(flash[:success]).to be_nil }
+      it { expect(workflow_token.groups_shared_among).to match_array([token_group]) }
+    end
+  end
```
hennevogel, line 55: You are testing authorization in a controller test. You already do that in the policy spec. CI time is developer time.

src/api/spec/controllers/webui/users/tokens/groups_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    context 'someone else tries to remove an ownership' do
+      let(:third_user) { create(:confirmed_user) }
+
+      before { login third_user }
+
+      subject! do
+        post :destroy, params: { id: token_group.id, token_id: workflow_token }
+      end
+
+      it { is_expected.to redirect_to(root_path) }
```
hennevogel, line 81: You are testing authorization in a controller test. You already do that in the policy spec. CI time is developer time.

src/api/spec/controllers/webui/users/tokens/users_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    context 'someone else tries to own the token' do
+      let(:third_user) { create(:confirmed_user) }
+
+      before { login third_user }
+
+      subject! do
+        post :create, params: { userid: third_user, token_id: workflow_token }
+      end
+
+      it { is_expected.to redirect_to(root_path) }
```
hennevogel, line 50: You are testing authorization in a controller test. You already do that in the policy spec. CI time is developer time.

src/api/spec/controllers/webui/users/tokens/users_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    context 'someone else tries to remove an ownership' do
+      let(:third_user) { create(:confirmed_user) }
+
+      before { login third_user }
+
+      subject! do
+        post :destroy, params: { id: token_user.id, token_id: workflow_token }
+      end
+
+      it { is_expected.to redirect_to(root_path) }
```
hennevogel, line 80: You are testing authorization in a controller test. You already do that in the policy spec. CI time is developer time.

src/api/spec/models/token/workflow_spec.rb
```diff
@@ -1,6 +1,42 @@
...
+  describe '#workflow_token_users' do
+    let(:token_user) { create(:confirmed_user) }
+
+    context 'asking for workflow tokens' do
+      let(:workflow_token) { create(:workflow_token, user: token_user) }
+
+      before { token_user.shared_workflow_tokens << workflow_token }
+
+      it 'returns the users whom I shared the token' do
+        expect(workflow_token.users_shared_among).to include(token_user)
```
hennevogel, line 13: What are you testing here? AR associations? I'm pretty sure they do that in the AR spec :) CI time is developer time.

src/api/spec/models/token/workflow_spec.rb
```diff
@@ -1,6 +1,42 @@
...
+  describe '#workflow_token_groups' do
+    let(:token_group) { create(:group) }
+
+    context 'asking for workflow tokens' do
+      let(:workflow_token) { create(:workflow_token) }
+
+      before { token_group.shared_workflow_tokens << workflow_token }
+
+      it 'returns the groups whom I shared the token' do
+        expect(workflow_token.groups_shared_among).to include(token_group)
```
hennevogel, line 35: What are you testing here? AR associations? I'm pretty sure they do that in the AR spec :) CI time is developer time.

hennevogel (changes requested):

src/api/config/routes/webui_routes.rb
```diff
@@ -349,6 +349,8 @@
 
       resources :tokens, controller: 'webui/users/tokens' do
         resources :workflow_runs, only: [:index, :show], controller: 'webui/workflow_runs'
+        resources :users, only: [:index, :create, :destroy], controller: 'webui/users/tokens/users'
+        resources :groups, only: [:create, :destroy], controller: 'webui/users/tokens/groups'
```
hennevogel, line 353: you need `constraints: cons` somewhere for those routes for users with a `.` in their username.

hennevogel (changes requested):

src/api/config/routes/webui_routes.rb
saraycp: done

eduardoj (approved):

PR #12687, opened 2022-06-10
Set wrap-text in confirmation modal texts

saraycp:
Modify ConfirmationDialogComponent to avoid texts going out of the modal like this:

Screenshot 2022-06-10 at 12-02-19 Open Build Service

and to have something like this:

Screenshot 2022-06-10 at 12-05-55 Open Build Service


rubhanazeem (approved):

saraycp:
`ci/circleci: checkout_code` check passed but wasn't reported to GitHub, so I can merge it.

PR #12689, opened 2022-06-10
Do not link to expanded sources with rev=

### LINKED ISSUE #12674: package/show sometimes shows wrong rev
# Issue Description
package/show sometimes shows wrong (latest) revision
I noticed that https://build.opensuse.org/package/show/X11:XOrg/intel-vaapi-driver?rev=5
as linked from https://build.opensuse.org/package/revisions/X11:XOrg/intel-vaapi-driver?page=2
shows the wrong rev. However, rev=1 works, so it is not always broken.

Expected Result
============
should show the old revision specified by the rev= parameter or print an error message why that is not possible.

How to Reproduce
===============
1. Go to https://build.opensuse.org/package/show/X11:XOrg/intel-vaapi-driver?rev=5

bmwiedemann:
Do not link to expanded sources with `rev=`

With default expand=1 it nearly always shows the latest version
from the `_link`, but if a user clicks "Browse Source"
to see an old revision, the user wants to see the old revision.

Fixes #12674

Needs testing.

### DIFF
--- a/src/api/app/views/webui/package/_commit_item.html.haml
+++ b/src/api/app/views/webui/package/_commit_item.html.haml
@@ -30,6 +30,6 @@
           %i.fas.fa-copy.text-gray-500
           Files Changed
     %li.nav-item
-      = link_to(package_show_path(@project, @package, rev: commit['rev']), class: 'nav-link') do
+      = link_to(package_show_path(@project, @package, rev: commit['rev'], expand: 0), class: 'nav-link') do
         %i.fas.fa-folder-open.text-warning
         Browse Source


hennevogel:
This will have other effects as well. Like you will only see the `_link` file in linked packages etc. Not sure this is what people expect. @adrianschroeter?

adrianschroeter:
A revision might point to unexpanded or expanded sources (xsrcmd5). I would not not enforce expanding/merging, since it might not what one want.

dmarcoux:
After reading through the discussion here, it seems like we don't want to have the changes included in this PR. I'm closing the PR. Feel free to re-open if I was mistaken.

PR #12700, opened 2022-06-16
Toggle visibility of the secret token

danidoni:
In order to effectively share a token, the user that gets the shared
token needs to know the token's secret. As of now, no one but the token
creator knows the token's secret.

This change fixes that, allowing a user with the proper permissions to
show and hide the secret from the token's page.

This is how it looks like:
Screenshot from 2022-06-16 16-42-26
Screenshot from 2022-06-16 16-42-03

### DIFF
--- a/src/api/app/views/webui/users/tokens/hide.js.erb
+++ b/src/api/app/views/webui/users/tokens/hide.js.erb
@@ -1,3 +1,3 @@
 $('#token-reveal').show();
-$('#token-hide').hide();
+$('#token-hide').addClass('d-none');
 $('#token-secret').empty();
--- a/src/api/app/views/webui/users/tokens/hide.js.erb
+++ b/src/api/app/views/webui/users/tokens/hide.js.erb
@@ -1,0 +1,3 @@
+$('#token-reveal').show();
+$('#token-hide').hide();
+$('#token-secret').empty();
--- a/src/api/app/views/webui/users/tokens/hide.js.erb
+++ b/src/api/app/views/webui/users/tokens/hide.js.erb
@@ -1,3 +1,0 @@
-$('#token-reveal').show();
-$('#token-hide').addClass('d-none');
-$('#token-secret').empty();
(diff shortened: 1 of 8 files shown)


src/api/app/views/webui/users/tokens/hide.js.erb
```diff
@@ -0,0 +1,3 @@
+$('#token-reveal').show();
+$('#token-hide').addClass('d-none');
+$('#token-secret').empty();
```
rubhanazeem, line 3: This could have been done by plain js. Why add ajax?

src/api/app/views/webui/users/tokens/show.js.erb
```diff
@@ -0,0 +1,3 @@
+$('#token-secret').html("<%= escape_javascript(render(CopyToClipboardInputComponent.new(input_text: @token.string))) %>");
+$('#token-reveal').hide();
+$('#token-hide').removeClass('d-none');
```
rubhanazeem, line 3: Same as above.

rubhanazeem (dismissed):

src/api/app/views/webui/users/tokens/show.js.erb
```diff
@@ -0,0 +1,3 @@
+$('#token-secret').html("<%= escape_javascript(@token.string) %>");
```
danidoni, line 1: I removed the `CopyToClipboardInputComponent` and went by a plain text because the component does not support several instances in the same page.

src/api/app/views/webui/users/tokens/hide.js.erb
danidoni: I was thinking that using javascript increased security of the thing because you had to authenticate on clicking the reveal button, but now I think that if we go for the simple plan js, you need to be authenticated as well...
Let me try

danidoni: Fixed. Please, review again!

src/api/app/views/webui/users/tokens/show.js.erb
danidoni: This is also fixed

danidoni:
In the end this was not a good idea, so closing

PR #12706, opened 2022-06-17
Move JavaScript part into the ConfirmationDialogComponent

saraycp:
Instead of implementing similar pieces of JS code from different views, a generic piece of JS is moved to the component.

TODO:

- [ ] Fix issue in watchlist. When I remove one item, the link "Watch this item/Remove this item" disappears and it shouldn't.
- [ ] Check if having the generic JS affects other existing confirmation dialogs.
- [ ] Rename component from ConfirmationDialogComponent to DeleteConfirmationDialogComponent?


saraycp:
I close this PR, I'm opening little PRs instead.

PR #12708, opened 2022-06-21
Add new column executor to tokens table

danidoni:
This is the first of several PRs to rename the existing `Token.user` relationship to `Token.executor` in a safe manner.

In this PR, we:
1. Create the new column
2. Move writes to the new column
3. Backfill data

Then, in PRs following this one, we'll move reads to the new column (#12714), and finally, delete the old `user` column.

### DIFF
--- a/src/api/app/controllers/person/token_controller.rb
+++ b/src/api/app/controllers/person/token_controller.rb
@@ -20,7 +20,7 @@ module Person
 
       pkg = (Package.get_by_project_and_name(params[:project], params[:package]) if params[:project] || params[:package])
 
-      @token = Token.token_type(params[:operation]).create(description: params[:description], user: @user, package: pkg, scm_token: params[:scm_token])
+      @token = Token.token_type(params[:operation]).create(description: params[:description], user: @user, executor_id: @user&.id, package: pkg, scm_token: params[:scm_token])
       return if @token.valid?
 
       render_error status: 400,
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -44,7 +44,9 @@ class Webui::Users::TokensController < Webui::WebuiController
   end
 
   def create
-    @token = Token.token_type(@params[:type]).new(@params.except(:type).merge(user: User.session, package: @package))
+    @token = Token.token_type(@params[:type]).new(@params.except(:type).merge(user: User.session,
+                                                                              executor_id: User.session&.id,
+                                                                              package: @package))
 
     authorize @token
 
--- a/src/api/app/models/token.rb
+++ b/src/api/app/models/token.rb
@@ -77,15 +77,17 @@ end
 #  string       :string(255)      indexed
 #  triggered_at :datetime
 #  type         :string(255)
+#  executor_id  :integer          default(0), not null, indexed
 #  package_id   :integer          indexed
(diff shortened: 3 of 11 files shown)


danidoni:
[Rename Token.user to Token.{owner,executor,run_as,whatever}](https://trello.com/c/0YzsOn6n/1908-rename-tokenuser-to-tokenownerexecutorrunaswhatever)

src/api/app/controllers/person/token_controller.rb
```diff
@@ -20,7 +20,7 @@ def create
 
       pkg = (Package.get_by_project_and_name(params[:project], params[:package]) if params[:project] || params[:package])
 
-      @token = Token.token_type(params[:operation]).create(description: params[:description], user: @user, package: pkg, scm_token: params[:scm_token])
+      @token = Token.token_type(params[:operation]).create(description: params[:description], user: @user, executor_id: @user&.id, package: pkg, scm_token: params[:scm_token])
```
hennevogel, line 23:
```suggestion
      @token = Token.token_type(params[:operation]).create(description: params[:description], user: @user, executor: @user, package: pkg, scm_token: params[:scm_token])
```

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -44,7 +44,9 @@ def update
   end
 
   def create
-    @token = Token.token_type(@params[:type]).new(@params.except(:type).merge(user: User.session, package: @package))
+    @token = Token.token_type(@params[:type]).new(@params.except(:type).merge(user: User.session,
+                                                                              executor_id: User.session&.id,
```
hennevogel, line 48:
```suggestion
                                                                              executor: User.session,
```

src/api/app/models/token.rb
```diff
@@ -77,15 +77,17 @@ def owned_by?(some_user)
 #  string       :string(255)      indexed
 #  triggered_at :datetime
 #  type         :string(255)
+#  executor_id  :integer          default(0), not null, indexed
```
hennevogel, line 80: Why don't you setup the association?

hennevogel (changes requested):

src/api/app/models/token.rb
danidoni: If I setup the association now, then ActiveRecord is aware of it, and then every single spec detects the missing `executor` data (is a not null field, then it's required) and complains about it, then I have to adapt tons of specs (factories, controllers, etc) just to make ActiveRecord happy, because in the end I will replace `user` by `executor` in the specs, so those changes that I would need to do are not really necessary. Using just the _id field goes under the radar of ActiveRecord, allows me to populate the field and perform the backfilling. Nothing less, nothing more.

danidoni: That is why you see the weird magic above

hennevogel (changes requested):

src/api/lib/tasks/databases.rake
```diff
@@ -6,4 +6,16 @@ namespace :db do
     puts 'warning for migrating your data run data:migrate'
     puts ''
   end
+
+  namespace :backfill do
```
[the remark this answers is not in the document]
danidoni: Yes, It is.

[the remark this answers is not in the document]
hennevogel: Why is it a rake task then? 😄

PR #12714, opened 2022-06-22
Move writes from tokens user to tokens executor

danidoni:
This is the second of several PRs to rename the existing Token.user relationship to Token.executor in a safe manner.

This is a follow-up to  #12708

In this PR, we:

1. Move reads from the new column
2. Disable the old column so ActiveRecord forgets it

In the last PR following this one, we'll delete the old user column.

### DIFF
--- a/src/api/app/components/token_card_component.html.haml
+++ b/src/api/app/components/token_card_component.html.haml
@@ -15,7 +15,7 @@
         Package:
         = token_package_link
     %p.card-text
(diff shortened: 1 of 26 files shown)


danidoni:
[Rename Token.user to Token.{owner,executor,run_as,whatever}](https://trello.com/c/0YzsOn6n/1908-rename-tokenuser-to-tokenownerexecutorrunaswhatever)

PR #12716, opened 2022-06-22
Activate rails 6.1 default configs

rubhanazeem:
This PR activates defaults for Rails 6.1 before changing `config.load_defaults` to 6.1 and remove new_framework_defaults from Rails 6.0 update.


hennevogel:
That's the equivalent of setting `config.load_defaults(6.1)` in `config/application.rb` and remove both files right? :)

danidoni (approved):

PR #12719, opened 2022-06-22
Rename token's user column to executor

danidoni:
This change needs to be done during the maintenance window because it is
not a safe operation.

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -42,8 +42,8 @@ class Token::Workflow < Token
   end
 
   def owned_by?(some_user)
-    # TODO: remove the first condition if we migrate the Token.user to Token.users
-    user == some_user || users.include?(some_user) || groups.map(&:users).flatten.include?(some_user)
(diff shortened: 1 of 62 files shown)


src/api/app/models/token/workflow.rb
```diff
@@ -42,8 +42,8 @@ def call(options)
   end
 
   def owned_by?(some_user)
-    # TODO: remove the first condition if we migrate the Token.user to Token.users
-    user == some_user || users.include?(some_user) || groups.map(&:users).flatten.include?(some_user)
+    # TODO: remove the first condition if we migrate the Token.user to Token.users_shared_among
```
dmarcoux, line 45: I believe this comment should be updated:

```suggestion
    # TODO: remove the first condition if we migrate the Token.executor to Token.users_shared_among
```

dmarcoux, line 45: `users_shared_among` doesn't exist anymore, it was renamed to `users`. The first part of the comment could be clearer and `Token.user` is now `Token.executor`:

```suggestion
    # TODO: remove the first condition if we data-migrate the Token.executor to Token.users
```

PR #12720, opened 2022-06-22
Move JavaScript part into the ConfirmationDialogComponent

saraycp:
Instead of calling the same JS function from different views, we call the JS from the component. It iterates over all the modals in the DOM whose id starts with `#delete`, takes the expected data attributes, if any, and sets those values in the modal content. 

**How to test?**

As this code iterates over all deletion modals in the DOM, I would test
- **deleting elements that already call the component**
  - add a project to the watchlist
  - remove the project from the watchlist using the deletion icon next to the project name
  - this should open a modal containing the project name
  - click the "Remove" button, and the project should disappear from the list.
- **deleting elements that don't call the component yet**
  - go to the same project page
  - add a few comments
  - delete a comment
  - this should open a modal, even if we are not using the component for the deletion of the comments yet
  - the comment should be removed

### DIFF
--- a/src/api/app/assets/javascripts/webui/delete_confirmation_dialog.js
+++ b/src/api/app/assets/javascripts/webui/delete_confirmation_dialog.js
@@ -1,4 +1,18 @@
-function setValuesOnDeleteConfirmationDialog(modalId) { // jshint ignore:line
+function collectDeleteConfirmationModalsAndSetValues() { // jshint ignore:line
+  $.each(modalIds(), function( _index, modalId ) {
+    setValuesOnDeleteConfirmationDialog(modalId);
+  });
+}
+
+function modalIds() {
+  var targets = $('a[data-toggle="modal"][data-target^="#delete"]').toArray().map(
+    function(e) { return $(e).data('target');
+  });
+
+  return $.unique(targets);
+}
+
+function setValuesOnDeleteConfirmationDialog(modalId) {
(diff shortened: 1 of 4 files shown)


src/api/app/assets/javascripts/webui/delete_confirmation_dialog.js
```diff
@@ -1,4 +1,18 @@
-function setValuesOnDeleteConfirmationDialog(modalId) { // jshint ignore:line
+function collectDeleteConfirmationModalsAndSetValues() { // jshint ignore:line
+  $.each(modalIds(), function( _index, modalId ) {
+    setValuesOnDeleteConfirmationDialog(modalId);
+  });
+}
+
+function modalIds() {
+  var targets = $('a[data-toggle="modal"][data-target^="#delete"]').toArray().map(
```
danidoni, line 8: Is this `toArray` really needed?

saraycp: Yes, if I don't use `toArray` here, I'll get an Object as result, instead of an array of ids.

This might be handled differently, but for me it was simpler to use `Array.map()` as explained [here](https://www.geeksforgeeks.org/javascript-array-map-method/).

danidoni: Okay

danidoni (approved):

PR #12736, opened 2022-06-27
Introduce validation for EmbargoDate attribute

### LINKED ISSUE #12649: Missing validation (timezone) for EmbargoDate attribute
# Issue Description
The `OBS:EmbargoDate` attribute is used for coordinating/blocking the release of an update until the issue is still under embargo. A CRD can also have a specific time set, not just the date.

The attribute is mainly used in the maintenance workflow of the internal build service instance but it can be used also on the public OBS instance.

Setting the attribute to '2022-03-08 10:00 PT' lead to a date of '2022-03-08 10:00 UTC' but if converted to UTC it should have been '2022-03-08 19:00 UTC'

Expected Result
============
The value entered is validated and rejected if what entered is not in the expected format. Not a bug but more a wish, could it be possible to have the date/time expressed in ISO 8601?

How to Reproduce
===============
1. Set the attribute to a date where is also specified the timezone (not UTC) like for example `2022-03-08 10:00 PT`
2. IBS accepts the value set (no errors)
3. The value is not correctly converted to UTC and the specified timezone is ignored

Further Information
===============
* [internal discussion](https://suse.slack.com/archives/C02BXKBMXNV/p1653912743336149) before filing this issue
* [commit introducing the attribute](https://github.com/openSUSE/open-build-service/commit/125c7b85ce9ae5671203c67a5f9bbba82aaa01eb)
* [embargo_handler.rb](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/project/embargo_handler.rb)

eduardoj:
This PR introduces validation for the EmbargoDate attribute.

This means it will be no longer possible to introduce an invalid EmbargoDate attribute.

Anyway, the step of checking for an embargo date when releasing packages still handles the case of invalid embargo dates, which were stored previously in the database. This PR doesn't contain any data migration to modify embargo dates according to the new validation rules.

As a consequence of introducing this validation in the `Attrib` model, some specs were modified to make use of invalid `Attrib` objects. Regarding maintenance minitests, as the case of invalid EmbargoDate attributes is already tested in specs, the minitests were adapted to handle the case of valid EmbargoDate attributes.

To test this PR:
- open the review-app
- go to a project
- click on the "Attributes" tab
- click on the "Add Attribute" link
- select "OBS:EmbargoDate" and click in the "Add" button
- fill the "Value:" text box with an embargo date (valid or invalid)
- click on "Save" and expect to see a success message, or an error message

Fixes #12649.

### DIFF
--- a/src/api/app/models/attrib.rb
+++ b/src/api/app/models/attrib.rb
@@ -27,6 +27,7 @@ class Attrib < ApplicationRecord
   validates :project, presence: true, if: proc { |attrib| attrib.package_id.nil? }
 
   validate :validate_value_count,
+           :validate_embargo_date_value,
            :validate_issues,
            :validate_allowed_values_for_attrib_type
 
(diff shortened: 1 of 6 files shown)


deneb-alpha:
hello @eduardoj,
I had a quick look at the code and, if I'm not "lost in translation", your implementation is considering as valid only a date in the format `YYYY-MM-DD` while, as far as I know, a CRD (coordinated release date) for an embargo can also be in the form of date and time, not just date.

Given that this change is more relevant for the security team, I would really suggest to check with them if this restriction is fine for them or if they need to also have a validation covering the specified time.

msmeissn:
yes, the embargodate can OPTIONALLY contain also a time, if possible including timezone specifier.

e.g.

2022-06-27 12:00UTC
2022-06-27 12:00CET

eduardoj:
I rebased over the PR that needs to be merged first, and applied your suggestions. The functionality can already be tested with the review-app.

However, some code must be refactored, and we should wait until the other PR gets merged.

This is how it looks like:

Peek 2022-07-01 15-28_EmbargoDate

src/api/app/models/attrib.rb
```diff
@@ -132,6 +164,15 @@ def validate_value_count
     errors.add(:values, "has #{values.length} values, but only #{value_count} are allowed") if value_count && value_count != values.length
   end
 
+  def validate_embargo_date_value
+    return unless attrib_type && name == 'EmbargoDate'
+
+    value = values[0]&.value
+    return if value.nil? || value == ''
```
krauselukas, line 171:
```suggestion
    return if value.blank?
```

eduardoj: Applied. Thanks!

krauselukas (approved):

PR #12740, opened 2022-06-29
Parse timezone (and abbreviations) for embargo date

eduardoj:
Before this change, the timezone of an embargo date was ignored.

```
> Time.parse('2022-10-10 10:10:10 CEST').utc
=> 2022-10-10 10:10:10 UTC
```

After this change, the timezone of an embargo date is no longer ignored.

```
> Time.zone.parse('2022-10-10 10:10:10 CEST').utc
=> 2022-10-10 08:10:10 UTC
```

Also throw an error in case that the timezone is not recognized. Otherwise it would take the 'UTC' timezone without erroring.

Related to #12649.

### DIFF
--- a/src/api/app/models/project/embargo_handler.rb
+++ b/src/api/app/models/project/embargo_handler.rb
@@ -16,18 +16,46 @@ class Project
     def embargo_date(embargo_attrib)
       return if embargo_attrib.nil?
 
-      begin
-        embargo = Time.parse(embargo_attrib&.value.to_s).utc
-        # no time specified, allow it next day
-        embargo = embargo.tomorrow if /^\d{4}-\d\d?-\d\d?$/.match?(embargo_attrib.value)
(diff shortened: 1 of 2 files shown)


src/api/app/models/project/embargo_handler.rb
```diff
@@ -16,18 +16,46 @@ def embargo_date_attribute
...
       raise BsRequest::Errors::UnderEmbargo, "The project #{@project.name} is under embargo until #{embargo_attrib}" if embargo > Time.now.utc
     end
 
     def call
       embargo_date(embargo_date_attribute)
     end
+
+    private
+
+    def check_timezone_identifier(value)
```
hennevogel, line 37: This needs specs I would say

hennevogel (changes requested):

hennevogel (approved):

PR #12755, opened 2022-07-05
Use component to delete status messages

saraycp:

To test this, you have to be part of the Staff group. Then go to the home page of OBS to create and delete Status Messages.


eduardoj (approved):

eduardoj (approved):

saraycp:
I've just pushed some changes to not render the component n-times (one per status message). We render it only once and if the user has right to delete it.

PR #12761, opened 2022-07-05
Use component to delete group members

saraycp:

To test this

- log in as Admin
- go to Configuration > Manage Groups
- click on the name of one of the groups, you'll see its members
- remove one member (you can add some first)
- check if  he name of the member is in the confirmation dialog
- click on "Remove" and check if the member disappears from the group


src/api/app/views/webui/groups/_group_members.html.haml
```diff
@@ -9,9 +9,16 @@
       - group.users.each do |user|
         .card.m-1.p-2.group-user
           - if write_access
-            = render(partial: 'webui/groups/group_members_delete_dialog', locals: { user: user, group: group })
-            = link_to('#', data: { toggle: 'modal', target: "#delete-group-members-modal-#{dom_id(user)}" },
-                      title: 'Remove member from group') do
+            = render DeleteConfirmationDialogComponent.new(modal_id: 'delete-group-members-modal',
+                                                           method: :delete,
+                                                           options: { modal_title: 'Do you want to remove this group member?' })
```
eduardoj, line 14: Rendering the component could happen only once if it is moved outside the `group.users.each do |user|` loop.

saraycp: you are right, I'll change it

saraycp: Now I render the component after the loop.

saraycp:
I've just updated this PR. Now I render the component only if the user has write access.

PR #12766, opened 2022-07-06
Describe what the "Share Token" page is for

eduardoj:
The page "Share Token" missed user documentation. This PR adds documentation explaining what is to "Share a token", and how can it be used.

Replace the literal "Users" with "Share". The literal "Users" is not so accurate since the introduction of sharing a token with groups.

### DIFF
--- a/src/api/app/views/webui/users/tokens/users/index.html.haml
+++ b/src/api/app/views/webui/users/tokens/users/index.html.haml
@@ -1,14 +1,26 @@
 :ruby
   @pagetitle = if @token.description.blank?
-                 'Users of the token'
+                 'Share token'
                else
-                 "Users of the '#{@token.description}' token"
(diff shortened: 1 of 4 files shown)


src/api/app/views/webui/users/tokens/users/index.html.haml
```diff
@@ -1,14 +1,26 @@
...
-                 'Users of the token'
+                 'Share token'
                else
-                 "Users of the '#{@token.description}' token"
+                 "Share '#{@token.description}' token"
                end
 
 .card.mb-3
   .card-body
+    %h3 Share Your Token with Others
```
danidoni, line 10: I think this title is too long, what about `Token Sharing`?

eduardoj: Fine. Applied.

PR #12768, opened 2022-07-06
Test rack/rake gem versions

hennevogel:
The system and bundle versions of rake and rack need to match up.
    
The rake and rack gems are dependencies of passenger that it will load, from the
system, during booting itself. If our bundle versions deviate, then booting our app
will try to load those gems again, which will fail. Which makes for an unbootable
app in production. Which will make people 😠 

See #12756 for a failing example

CI now fails like this

Screenshot from 2022-07-06 16-50-38

Build now fails like this

Screenshot from 2022-07-06 16-59-20

### DIFF
--- a/.circleci/conditional_config.yml
+++ b/.circleci/conditional_config.yml
@@ -203,7 +203,11 @@ jobs:
       - run: *install_circle_cli
       - run:
           name: Setup application
-          command: cd src/api; bundle exec rake dev:prepare assets:precompile RAILS_ENV=test FORCE_EXAMPLE_FILES=1
+          command: |
+            make -C src/api test_rake
+            make -C src/api test_rack
+            cd src/api
+            bundle exec rake dev:prepare assets:precompile RAILS_ENV=test FORCE_EXAMPLE_FILES=1
       - persist_to_workspace:
           root: .
           paths:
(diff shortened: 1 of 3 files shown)


.circleci/conditional_config.yml
```diff
@@ -203,7 +203,11 @@ jobs:
       - run: *install_circle_cli
       - run:
           name: Setup application
-          command: cd src/api; bundle exec rake dev:prepare assets:precompile RAILS_ENV=test FORCE_EXAMPLE_FILES=1
+          command: |
+            make -C src/api test_rake
+            make -C src/api test_rack
```
eduardoj, line 208: Could these two `make` checks already be run between `- run: *install_dependencies` and `- *save_bundler_cache`?

eduardoj: Because we don't want to spend any time of our CI saving the bundler cache if we will throw an error anyways.

hennevogel: It's not wasted as we are going to use the cache. The usual workflow will be

1. depfu updates rack (saves the cache for this bundle)
2. CI fails because no one updated the [rubygem-rack](https://build.opensuse.org/package/show/OBS:Server:Unstable/rubygem-rack) package
3. Someone updates rubygem-rack
4. CI won't fail anymore (and use the cache)

eduardoj: Ok, let's try it.

PR #12771, opened 2022-07-07
[api] support staging without write access in GA projects

adrianschroeter:
SLSA requires seperation of release managers and code stream managers.
Therefore release managers just approve reviews and the code stream
managers using the request approve mechanism.

As consequence the release managers have no write access anymore
in the GA projects:
 * request gets not accepted directly via staging api, just indirect
   via approve mechanism
 * modifications outside of the requests (like fiddeling with local links)
   are not allowed anymore

jsc#OBS-200

### DIFF
--- a/src/api/app/controllers/concerns/rescue_authorization_handler.rb
+++ b/src/api/app/controllers/concerns/rescue_authorization_handler.rb
@@ -52,7 +52,7 @@ module RescueAuthorizationHandler
       when :anonymous_user
         'Please login to access the resource'
       when :request_state_change
-        "Request #{exception.record.number} is not acceptable"
+        "Request #{exception.record.number} would not be acceptable by you"
       else
         "Sorry, you are not authorized to #{action_for_exception(exception)} this #{ActiveSupport::Inflector.underscore(exception.record.class.to_s).humanize(capitalize: false)}."
       end
--- a/src/api/app/controllers/concerns/rescue_authorization_handler.rb
+++ b/src/api/app/controllers/concerns/rescue_authorization_handler.rb
@@ -51,6 +51,8 @@ module RescueAuthorizationHandler
       case exception.reason
       when :anonymous_user
         'Please login to access the resource'
+      when :request_state_change
+        "Request #{exception.record.number} is not acceptable"
       else
         "Sorry, you are not authorized to #{action_for_exception(exception)} this #{ActiveSupport::Inflector.underscore(exception.record.class.to_s).humanize(capitalize: false)}."
       end
--- a/src/api/app/controllers/staging/staging_projects_controller.rb
+++ b/src/api/app/controllers/staging/staging_projects_controller.rb
@@ -48,7 +48,6 @@ class Staging::StagingProjectsController < Staging::StagingController
 
   def accept
     authorize @staging_project, :accept?
-    authorize @project, :update?
 
     # check general state
     raise StagingProjectNotAcceptable, 'Staging project is not in state acceptable.' unless can_accept?
--- a/src/api/app/controllers/staging/staging_projects_controller.rb
+++ b/src/api/app/controllers/staging/staging_projects_controller.rb
@@ -49,19 +49,31 @@ class Staging::StagingProjectsController < Staging::StagingController
   def accept
     authorize @staging_project, :accept?
 
-    # check general state
-    raise StagingProjectNotAcceptable, 'Staging project is not in state acceptable.' unless can_accept?
-
-    StagingProjectAcceptJob.perform_later(project_id: @staging_project.id, user_login: User.session!.login)
-    render_ok
+    if acceptable?(force: params[:force].present?)
+      StagingProjectAcceptJob.perform_later(project_id: @staging_project.id, user_login: User.session!.login)
+      render_ok
+    else
+      render_error(
+        status: 400,
+        errorcode: 'staging_project_not_in_acceptable_state',
+        message: "Staging Project is not acceptable: #{@acceptable_error}"
+      )
+    end
   end
 
   private
 
-  def can_accept?
-    return true if @staging_project.overall_state == :acceptable
+  def acceptable?(force: false)
+    @acceptable_error = 'has reviews open' unless @staging_project.missing_reviews.empty?
 
-    params[:force].present? && @staging_project.force_acceptable?
+    if force
+      unless @staging_project.overall_state.in?(StagingProject::FORCEABLE_STATES)
+        @acceptable_error = "is not in state #{StagingProject::FORCEABLE_STATES.to_sentence(last_word_connector: ' or ')}"
+      end
+    else
+      @acceptable_error = "#{@staging_project.overall_state} is not an acceptable state" unless @staging_project.overall_state == :acceptable
+    end
+    @acceptable_error.blank?
   end
 
   def set_options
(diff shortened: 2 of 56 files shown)


adrianschroeter:
this is just currently an untested draft, needs further investigation and testing. And esp. review of permission handling.

eduardoj:
@adrianschroeter, only a small note about labelling pull requests: we usually create issues and label them with a priority label (P2 for example). And after that we create pull requests which could solve the issue. This helps us to separate the problem (and assign to it a prioriy) from the solution.

src/api/app/models/project.rb
```diff
@@ -1122,6 +1122,12 @@ def open_requests_with_by_project_review
                                                        .joins(:reviews), name).call
   end
 
+  def open_unapproved_requests_with_by_project_review
+    # Includes also by_package reviews for packages contained in this project
+    OpenRequestsWithByProjectReviewFinder.new(BsRequest.where(state: [:new, :review]).where.not(approver: nil)
```
coolo, line 1127: And so this call will find requests with by_project review and approver - but you're looking for unapproved requests (according to the name of the function) and those do *not* have an approver. 

What you really want to look at is if all staged requests are approved - those are listed in `@staging_project.staged_requests`.

src/api/app/controllers/staging/staging_projects_controller.rb
```diff
@@ -48,7 +48,7 @@ def copy
 
   def accept
     authorize @staging_project, :accept?
-    authorize @project, :update?
+    authorize @project, :update? if @staging_project.open_unapproved_requests_with_by_project_review
```
coolo, line 51: That should be part of staging project accept policy instead. Skipping authorize calls like this makes them basically untestable as policies - and the whole claim behind not testing controller permissions is that policies in itself are tested (IMO)

coolo (changes requested):

src/api/app/models/project.rb
hennevogel: Yeah I have some commits not pushed that change/test this. I got sidetracked by some nasty bug in `BsRequest.auto_accept`...

src/api/app/models/staging/workflow.rb
```diff
@@ -115,7 +115,7 @@ def update_managers_group
     # update reviewer group in backlog requests
     target_of_bs_requests.stageable(old_managers_group.title).each do |bs_request|
       bs_request.addreview(by_group: new_managers_group.title, comment: 'Staging manager group changed')
-      bs_request.change_review_state(:accepted, by_group: old_managers_group.title, comment: 'Staging manager group changed')
+      bs_request.change_review_state(:accepted, by_group: old_managers_group.title, comment: 'Staging manager group changed') unless bs_request.approver
```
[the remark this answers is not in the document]
coolo: not in update_managers_group though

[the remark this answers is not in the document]
hennevogel: indeed

src/api/app/models/bs_request.rb
```diff
@@ -506,6 +506,15 @@ def permission_check_change_state!(opts)
     check_bs_request_actions!(skip_source: true)
   end
 
+  def permission_check_change_state(opts)
+    begin
+      permission_check_change_state!(opts)
+    rescue PostRequestNoPermission
```
hennevogel, line 512: There are possibly more:

- BsRequest::Errors::CreatorCannotAcceptOwnRequests
- PostRequestMissingParameter

Not sure if we should catch them, sounds confusing to me...

hennevogel: or we catch in the policy directly...

src/api/app/policies/project_policy.rb
```diff
@@ -49,7 +49,7 @@ def accept?
     record.staged_requests.each do |request|
       # we pretend the user asked for force, we only want to check permissions
       # not if it makes sense
-      request.permission_check_change_state!(newstate: 'accepted', force: true)
+      return false unless request.permission_check_change_state(newstate: 'accepted', force: true)
```
coolo, line 52: note that force accept is no longer an option if you're not target maintainer

hennevogel: As far as I can see this is just about https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/bs_request_permission_check.rb#L354-L355

coolo: Jein. The policy uses force and skips the review check by that - but we later check that the project is in acceptable and so staging projects in review get blocked on a different level. The task of the policy is to check if it's technically possible, not if the staging project is "acceptable" or not.

But we also checked that the accepting user is target maintainer and so we *could* skip review checks. If this is no longer guaranteed, we need to check that the request is actually acceptable in full glory. I.e. if accepting the last review *really* will accept the request by approval.

hennevogel: > The task of the policy is to check if it's technically possible, not if the staging project is "acceptable" or not.

The policy is about authorization, checking permissions. If someone is allowed to do something. But as you know things are tangled deeply...

>  if accepting the last review really will accept the request by approval

That is what we do no?

hennevogel:
And now depends on #12845 12845

hennevogel:
@coolo can you have another look please?

src/api/app/models/project.rb
hennevogel: we talked offline

this is fine...

src/api/app/models/bs_request.rb
hennevogel: I decided against catching those "validations" in the permission check.

src/api/app/controllers/concerns/rescue_authorization_handler.rb
```diff
@@ -43,6 +43,8 @@ def authorization_message(exception)
       case exception.reason
       when :anonymous_user
         'Please login to access the resource'
+      when :request_state_change
+        "Request #{exception.record.number} would not be acceptable by you"
```
dmarcoux, line 47: A request state change means someone is potentially declining a request, not only accepting it. This message is confusing. How about this?

```suggestion
        "You cannot accept nor decline the request #{exception.record.number}."
```

coolo (approved):

src/api/app/controllers/concerns/rescue_authorization_handler.rb
hennevogel: This only ever happens in accepting staging projects

hennevogel: I can rename the `.reason` to be more staging related if you want to

hennevogel:
Let's go then @adrianschroeter @coolo right?

adrianschroeter:
sure, it works in production since quite some time.

coolo:
This broke staging accept as the release tools always accept with force after checking on client side that everything is in acceptable. As acceptable is not in `FORCEABLE_STATES` this fails now. 

The reason we have to accept with force is that if you accept two stagings at the same time, the 2nd one will flip to 'building' due to the scheduler thinking "oh, something in Factory changed, let me check if all the projects linking to it need to change".

DimStar77:
[That's already monkey patched on OBS
](https://github.com/openSUSE/open-build-service/pull/13004)

PR #12792, opened 2022-07-11
Run rake/rack tests only if the bundle changes

hennevogel:
If the PR does not change the bundle, there is no need to check
for a rake/rack mismatch in CI.


hennevogel:
In the last run the Spider test succeeded. See 

https://app.circleci.com/pipelines/github/openSUSE/open-build-service?branch=pull%2F12792

dmarcoux (approved):

PR #12803, opened 2022-07-13
Introduces commit_opts[:login] for delete_on_backend

hennevogel:
And uses this from BsRequestActionDelete.

Before we have relied on User.run_as in BsRequest.auto_accept
to set User.session (which is a Thread/fiber-local variable).
But in an `after_commit` callback this is not working reliably.
Maybe a race or something with nested transcations or whatnot.
Couldn't figure it out...


eduardoj:
Typo in the commit message: `transcations`.

hennevogel:
> transcations

🏳️‍🌈 🐱

PR #12808, opened 2022-07-14
Allow users to manage their own home project and its packages

### LINKED ISSUE #10067: Project owner can not reacquire their project from webui
# Issue Description
After unchecking all roles of a project user from webui, the user will disappear normally after the page refresh. Same disappear for project owner. It seems you can not reacquire the project from webui.

Expected Result
============
Keep the explicit owner of a project (maybe just home project?) even if there are no role be checked in the project configuration

(Even if we can use `osc maintainer -a` to add ourselves to our project from cli)

How to Reproduce
===============
1. Go to self home project page
2. Click on 'Users'
3. Uncheck all roles for yourself
4. Refresh page
5. See your own user no longer exists

dmarcoux:
This is needed since users sometimes remove themselves from the maintainers of their own home project. It should have never been possible in the first place for users to do this, but here we are. It was quite confusing to users since they could still see the `Create Package` or `Add User` actions, but they were never authorized to do those actions. They should have been able to do them though. This is now addressed.

For now, we can help users by applying this fix. This code from the custom authorization system is called in Pundit, so both authorization systems (custom and Pundit) are covered.

Another important change here is that nothing can prevent a user from managing their home project or a package in their home project. As an example, it was previously possible to prevent users from accessing the source of their home project by setting `sourceaccess` to `disable` in the local permissions.

Fixes #10067

--- For reviewers

_First scenario: Add yourself back_

1. Log in as a non-admin user (username: _Iggy_ / password: _buildservice_)
2. Go to your home project's users page
3. Remove yourself from the list of users
4. Add yourself back (it works with the changes from this PR, but not without them)

_Second scenario: Create packages/subprojects in your home project, while you aren't listed as a user of your home project_

1. Log in as a non-admin user (username: _Iggy_ / password: _buildservice_)
2. Go to your home project's users page
3. Remove yourself from the list of users
4. Create a package and a subproject in your home project (it works with the changes from this PR, but not without them)

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -590,6 +590,7 @@ class User < ApplicationRecord
   # if context is a package, check permissions in package, then if needed continue with project check
   # if context is a project, check it, then if needed go down through all namespaces until hitting the root
   # return false if none of the checks succeed
+  # rubocop:disable Metrics/PerceivedComplexity
   def has_local_permission?(perm_string, object)
     roles = Role.ids_with_permission(perm_string)
     return false unless roles
@@ -600,8 +601,17 @@ class User < ApplicationRecord
       logger.debug "running local permission check: user #{login}, package #{object.name}, permission '#{perm_string}'"
       # check permission for given package
       parent = object.project
+
+      # Users have permissions to manage packages in their own home project
+      # This is needed since users sometimes remove themselves from the maintainers of their own home project
+      return true if parent.name == home_project_name
     when Project
       logger.debug "running local permission check: user #{login}, project #{object.name}, permission '#{perm_string}'"
+
+      # Users have permissions to manage their own home project
+      # This is needed since users sometimes remove themselves from the maintainers of their own home project
+      return true if object.name == home_project_name
+
       # check permission for given project
       parent = object.parent
     when nil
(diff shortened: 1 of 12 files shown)


src/api/app/models/user.rb
```diff
@@ -590,6 +590,7 @@ def has_local_role?(role, object)
   # if context is a package, check permissions in package, then if needed continue with project check
   # if context is a project, check it, then if needed go down through all namespaces until hitting the root
   # return false if none of the checks succeed
+  # rubocop:disable Metrics/PerceivedComplexity
```
dmarcoux, line 593: Refactoring this method is clearly out of scope for this fix.

danidoni (approved):

eduardoj (changes requested):
Failing tests in `spec/controllers/webui/package_controller_spec.rb` should be modified.

src/api/spec/controllers/webui/package_controller_spec.rb
```diff
@@ -264,7 +264,8 @@ def remove_file_post
   end
 
   describe 'GET #revisions' do
-    let(:package) { create(:package_with_revisions, name: 'package_with_one_revision', revision_count: 1, project: source_project) }
```
dmarcoux, line 267: The specs were previously relying on a package in `source_project`, which the home project of `user`. Preventing a user from accessing the source of their own home project isn't possible anymore.

eduardoj (approved):
> Failing tests in `spec/controllers/webui/package_controller_spec.rb` should be modified.

The tests have been adapted. Looks good to me.

PR #12820, opened 2022-07-18
Allow /public access to files with spaces

coolo:
controller has it's own usage of CGI.escape not ported. So use
the model method available to build the path.


coolo:
There are two some `CGI.escape` left that worry me as they are related to path (and then quite a few more in specs and scripts were due to selection of fixtures it doesn't matter in reality)

```
app/helpers/maintenance_helper.rb:    cp_path = "/source/#{CGI.escape(target_project.name)}/#{CGI.escape(target_package_name)}"
app/models/project.rb:      path += "/#{CGI.escape(file)}" if file.present?
```

PR #12824, opened 2022-07-18
[api] sanities permission check

adrianschroeter:
only because you can not write to project (may exist and be locked) it does not mean
you can create it when you have write access on the upper level.


dmarcoux (dismissed):
We should change `can_create_project?` to include this new check. This would be way more robust since `can_create_project?` isn't only used in this line. Otherwise, authorization doesn't have the same rules everywhere.

dmarcoux:
So this is now a P2, I will simply fix the linters and be done with it. It's clearly a shortcut, but I have other things to do.

eduardoj (approved):

PR #12825, opened 2022-07-18
Downgrade nokogiri from 1.13.7 back to 1.13.6

danidoni:
This PR removes the need to upgrade nokogiri in #12801. That nokogiri upgrade requires ruby version >= 2.6.0 on [the 2.10 build](https://build.opensuse.org/package/show/OBS:Server:2.10:Staging/obs-server) and we don't want to upgrade ruby right now.

### DIFF
--- a/src/api/Gemfile.lock
+++ b/src/api/Gemfile.lock
@@ -247,7 +247,7 @@ GEM
     mousetrap-rails (1.4.6)
     mysql2 (0.5.2)
     nio4r (2.5.8)
-    nokogiri (1.13.7)
+    nokogiri (1.13.6)
       mini_portile2 (~> 2.8.0)
       racc (~> 1.4)
     parallel (1.17.0)


hennevogel (changes requested):
explain yourself stanger!

danidoni:
> explain yourself stanger!

Yeah, sorry.

danidoni:
> The version numbers are wrong. _17_ should be _7_ and _16_ should be _6_.
> 
> Nokogiri _1.13.6_ still requires Ruby >= 2.6 as we can see here: https://rubygems.org/gems/nokogiri/versions/1.13.6. You would have to downgrade to _1.12.5_ to support Ruby >= 2.5 (see https://rubygems.org/gems/nokogiri/versions/1.12.5). We cannot do this though as we would have security vulnerabilities.

Thanks, you are right about the numbering, my bad.

About the ruby version I don't understand what's happening here: before the change that pulled in nokogiri v1.13.7 (#12801) we had nokogiri v1.13.6, and it was working fine with ruby version 2.5.9p229 :thinking:

dmarcoux:
It's possible that nokogiri _1.13.6_ worked with Ruby _2.5_, but the nokogiri developers aren't supporting this Ruby version anymore since nokogiri _1.12.5_. We can also stick to _1.13.6_, but nokogiri is somewhat often hit with security vulnerabilities.... so I guess it's a short/mid-term solution.

hennevogel (approved):

PR #12828, opened 2022-07-19
Some more URL escape fun

### LINKED ISSUE #12826: CGI.escape vs. Addressable::URI.escape
There are two more `CGI.escape` left that worry me as they are related to path (and then quite a few more in specs and scripts were due to selection of fixtures it doesn't matter in reality)

```
app/helpers/maintenance_helper.rb:    cp_path = "/source/#{CGI.escape(target_project.name)}/#{CGI.escape(target_package_name)}"
app/models/project.rb:      path += "/#{CGI.escape(file)}" if file.present?
```

_Originally posted by @coolo in https://github.com/openSUSE/open-build-service/issues/12820#issuecomment-1186935309_

hennevogel:
Fixes #12826

### DIFF
--- a/src/api/app/helpers/maintenance_helper.rb
+++ b/src/api/app/helpers/maintenance_helper.rb
@@ -169,7 +169,7 @@ module MaintenanceHelper
         cp_params[:freezelink] = 1
       end
     end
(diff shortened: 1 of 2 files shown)


dmarcoux (approved):

hennevogel:

PR #12831, opened 2022-07-19
Stop relying on bundle_gems service

hennevogel:
Use vendor/cache from git instead


coolo:
BTW: Consider tracking them as LFS objects so the gems themselves aren't part of the git history you checkout

hennevogel:

PR #12836, opened 2022-07-20
Use component to delete users/groups

saraycp:
On the project or package users page, we can delete users or groups with different roles.

**How to test?**

- Go to project or package page.
- Add users and groups with any combination of roles.
- Click to remove one user. The deletion should work.
- Click to remove one group. The deletion should work.


eduardoj:
I rebased on master to get rid of the conflicts.

eduardoj (approved):

dmarcoux (approved):

PR #12842, opened 2022-07-21
Simplify require_request method in Webui::RequestController

dmarcoux:
`require_request` is used in a before_filter callback for `changerequest`, `show` and `request_action` actions. The request number has to be passed in the [show](https://github.com/openSUSE/open-build-service/blob/9dbda912de8d0a62aa135f11a7255b00bbb66ca0/src/api/config/routes/webui_routes.rb#L291) and [request_action](https://github.com/openSUSE/open-build-service/blob/9dbda912de8d0a62aa135f11a7255b00bbb66ca0/src/api/config/routes/webui_routes.rb#L297) routes, so it's impossible to not have `number` in the `params`. As for `changerequest`, the request number is also passed in the `params` via a [hidden field](https://github.com/openSUSE/open-build-service/blob/349e41e8a177cd9094d4af8715ebe554f384b28f/src/api/app/views/webui/request/_decision_tab.html.haml#L2) when the form is submitted. So unless a user changes the HTML, it's always going to be there. And if this happens, there's nothing we can do.

This explains why relying on required_parameters and all this custom code is not needed. We can instead do it the Rails way and rely on `find_by!`, which raises `ActiveRecord::RecordNotFound` if the request isn't found, then the controller responds with 404.

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -288,12 +288,7 @@ class Webui::RequestController < Webui::WebuiController
   end
 
   def require_request
-    required_parameters :number
-    @bs_request = BsRequest.find_by_number(params[:number])
-    return if @bs_request
-
-    flash[:error] = "Can't find request #{params[:number]}"
-    redirect_back(fallback_location: root_url) && return
+    @bs_request = BsRequest.find_by!(number: params[:number])
   end
 
   def target_package_maintainers
--- a/src/api/spec/controllers/webui/request_controller_spec.rb
+++ b/src/api/spec/controllers/webui/request_controller_spec.rb
@@ -48,15 +48,6 @@ RSpec.describe Webui::RequestController, vcr: true do
       end
     end
 
-    context 'if request does not exist' do
-      before do
-        get :show, params: { number: '200000' }
-      end
-
-      it { expect(flash[:error]).to eq("Can't find request 200000") }
-      it { expect(response).to redirect_to(root_url) }
-    end
-
     context 'when there are package maintainers' do
       # The hint will only be shown, when the target package has at least one
       # maintainer. So we'll gonna add a maintainer to the target package.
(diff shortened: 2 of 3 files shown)


dmarcoux:
Hum.. I just discovered that we overwrite the dynamic finder `find_by_number!` with [our own version](https://github.com/openSUSE/open-build-service/blob/349e41e8a177cd9094d4af8715ebe554f384b28f/src/api/app/models/bs_request.rb#L110).

dmarcoux:
Once  #12844 is merged, we can rebase this PR on master and mark it as ready for review.

dmarcoux:
[Simplify `set_request` in Webui::RequestController](https://trello.com/c/J971La7q/1936-simplify-setrequest-in-webuirequestcontroller)

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -288,12 +288,7 @@ def set_superseded_request
   end
 
   def require_request
-    required_parameters :number
-    @bs_request = BsRequest.find_by_number(params[:number])
-    return if @bs_request
-
-    flash[:error] = "Can't find request #{params[:number]}"
-    redirect_back(fallback_location: root_url) && return
+    @bs_request = BsRequest.find_by_number!(params[:number])
```
hennevogel, line 291:
```suggestion
    @bs_request = BsRequest.find_by!(number: params[:number])
```

src/api/spec/controllers/webui/request_controller_spec.rb
```diff
@@ -49,12 +49,7 @@
     end
 
     context 'if request does not exist' do
-      before do
-        get :show, params: { number: '200000' }
-      end
-
-      it { expect(flash[:error]).to eq("Can't find request 200000") }
-      it { expect(response).to redirect_to(root_url) }
+      it { expect { get :show, params: { number: '200000' } }.to raise_error(ActiveRecord::RecordNotFound) }
```
hennevogel, line 52: Now that you removed all that custom code, time to remove the superfluous spec. I'm pretty sure `find_by!` is tested in the `ActiveRecord` test suite :)

hennevogel (changes requested):

src/api/app/controllers/webui/request_controller.rb
dmarcoux: Sure, why not. I don't have a preference here.

src/api/spec/controllers/webui/request_controller_spec.rb
dmarcoux: Yep, that's right!

hennevogel (approved):

PR #12844, opened 2022-07-21
Do not overwrite dynamic finder find_by_number! for BsRequest

dmarcoux:
It's confusing any developer who's used to the built-in dynamic finders for ActiveRecord models. If we want a custom exception message for BsRequest#number, we can still do it without overwriting `find_by_number!`.

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -57,8 +57,8 @@ class Webui::RequestController < Webui::WebuiController
       request.addreview(opts)
     rescue BsRequestPermissionCheck::AddReviewNotPermitted
       flash[:error] = "Not permitted to add a review to '#{params[:number]}'"
-    rescue ActiveRecord::RecordInvalid, APIError => e
-      flash[:error] = "Unable add review to '#{params[:number]}': #{e.message}"
+    rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotFound, APIError => e
+      flash[:error] = "Unable to add review to request with id '#{params[:number]}': #{e.message}"
     end
     redirect_to controller: :request, action: 'show', number: params[:number]
   end
--- a/src/api/app/controllers/webui/requests/submissions_controller.rb
+++ b/src/api/app/controllers/webui/requests/submissions_controller.rb
@@ -54,6 +54,8 @@ module Webui
                                  superseded_by: @bs_request.number)
         rescue APIError => e
           supersede_errors << e.message.to_s
+        rescue ActiveRecord::RecordNotFound
+          supersede_errors << "Couldn't find request with id '#{request_number}'"
         end
 
         return if supersede_errors.empty?
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -107,18 +107,6 @@ class BsRequest < ApplicationRecord
     end
   end
 
-  def self.find_by_number!(number)
-    # overload for propper error reporting
-    r = BsRequest.find_by_number(number)
-    unless r
-      # the external visible request id is stored in number row.
-      # the database id must not be exposed to the outside
-      raise NotFoundError, "Couldn't find request with id '#{number}'"
-    end
-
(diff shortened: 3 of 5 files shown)


src/api/app/controllers/webui/requests/submissions_controller.rb
```diff
@@ -54,6 +54,8 @@ def supersede_requests
                                  superseded_by: @bs_request.number)
         rescue APIError => e
           supersede_errors << e.message.to_s
+        rescue ActiveRecord::RecordNotFound
+          supersede_errors << "Couldn't find request with id '#{request_number}'"
```
dmarcoux, line 58: To have a nicer message for users whenever something goes wrong as they submit a BsRequest. This keeps the same behavior as before.

dmarcoux:
I have reviewed usage of `BsRequest.find_by_number!` to ensure that the `ActiveRecord::RecordNotFound` is handled correctly. In the web UI, we now rescue the exception to have a better UX for users. As for the API, this exception is already handled by [RescueHandler](https://github.com/openSUSE/open-build-service/blob/349e41e8a177cd9094d4af8715ebe554f384b28f/src/api/app/controllers/concerns/rescue_handler.rb#L46-L48) in the `ApplicationController`.

dmarcoux:
[Simplify `set_request` in Webui::RequestController](https://trello.com/c/J971La7q/1936-simplify-setrequest-in-webuirequestcontroller)

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -57,8 +57,8 @@ def add_reviewer
       request.addreview(opts)
     rescue BsRequestPermissionCheck::AddReviewNotPermitted
       flash[:error] = "Not permitted to add a review to '#{params[:number]}'"
-    rescue ActiveRecord::RecordInvalid, APIError => e
-      flash[:error] = "Unable add review to '#{params[:number]}': #{e.message}"
+    rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotFound, APIError => e
+      flash[:error] = "Unable to add review to request with id '#{params[:number]}': #{e.message}"
```
eduardoj, line 61: With this change, the `rescue` sections handle exceptions from both lines, the `BsRequest.find_by_number!(params[:number])` and the `request.addreview(opts)` one.

I would split the code more:
- moving the `request = BsRequest.find_by_number!(params[:number])` line to the beginning of the method, and rescue there for the `ActiveRecord::RecordNotFound` exception,
- then fill the `opts` hash
- and lastly create a begin block for the `request.addreview(opts)` call, with the rescue sections we had before.

The reason is that we make clearer what exceptions to expect from running each chunk of code.

Like this:
```ruby
  def add_reviewer
    begin
      request = BsRequest.find_by_number!(params[:number])
    rescue ActiveRecord::RecordNotFound => e
      flash[:error] = "Unable to add review to request with id '#{params[:number]}': #{e.message}"
      return
    end
 
    opts = {}
    case params[:review_type]
    when 'review-user'
      opts[:by_user] = params[:review_user]
    when 'review-group'
      opts[:by_group] = params[:review_group]
    when 'review-project'
      opts[:by_project] = params[:review_project]
    when 'review-package'
      opts[:by_project] = params[:review_project]
      opts[:by_package] = params[:review_package]
    end
    opts[:comment] = params[:review_comment] if params[:review_comment]

    begin
      request.addreview(opts)
    rescue BsRequestPermissionCheck::AddReviewNotPermitted
      flash[:error] = "Not permitted to add a review to '#{params[:number]}'"
    rescue ActiveRecord::RecordInvalid, APIError => e
      flash[:error] = "Unable to add review to request with id '#{params[:number]}': #{e.message}"
    end
  end
```

dmarcoux: I'm not against this suggestion, but I find it out of scope for this PR. I prefer to not do this here.

eduardoj: Ok.

danidoni (approved):

PR #12845, opened 2022-07-21
Move the accept logic into StagingProject

hennevogel:
Skinny controller/job, fat model. This is logic of the StagingProject not the job.


src/api/spec/models/project/staging_project_spec.rb
```diff
@@ -273,5 +273,118 @@
         end
       end
     end
+
+    describe '.accept' do
```
hennevogel, line 277: This is basically the `staging_project_accept_job_spec.rb` moved here, only some minor changes...

danidoni (approved):

hennevogel:

PR #12860, opened 2022-07-25
Introduce Changes tab for requests show redesign

eduardoj:
Screenshot from 2022-07-25 17-52-00

### DIFF
--- a/src/api/app/assets/javascripts/webui/request.js
+++ b/src/api/app/assets/javascripts/webui/request.js
@@ -112,6 +112,7 @@ function requestAddAutocomplete(autocompleteElement) { // jshint ignore:line
 }
 
 $(document).ready(function(){
+  // TODO: Remove the enclosing code when the request_show_redesign feature is finished - START
   var element = $('.bs-request-actions li:first-child a:first-child');
   if (element.length !== 0){
     loadDiffs($(element));
@@ -133,8 +134,13 @@ $(document).ready(function(){
       });
     }
   });
+  // TODO: Remove the enclosing code when the request_show_redesign feature is finished - END
+  $('#changes-item').on('shown.bs.tab', function () {
+    loadChanges();
+  });
 });
 
+// TODO: Remove the following method when the request_show_redesign feature is finished
 function reloadRequestAction(index){ // jshint ignore:line
   var element = $('.request-tab[data-index=' + index + ']');
   $('.tab-pane.sourcediff.active').html('');
@@ -143,6 +149,7 @@ function reloadRequestAction(index){ // jshint ignore:line
   }
 }
 
+// TODO: Remove the following method when the request_show_redesign feature is finished
 function loadDiffs(element){
   $('.loading-diff').removeClass('invisible');
   var index = element.data('index');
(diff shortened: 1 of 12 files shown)


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -177,6 +177,26 @@ def request_action
     end
   end
 
+  def request_action_changes
```
dmarcoux, line 180: With the use of this new action, the current action `request_action` could be deleted once the `request_show_redesign` feature is finished. Could you add a comment above `def request_action` for this?

eduardoj: Yes. Done.

src/api/app/assets/javascripts/webui/request.js
```diff
@@ -162,3 +165,17 @@ function loadDiffs(element){
     }
   });
 }
+
+function loadChanges(changesItem) {
```
dmarcoux, line 169: In this file, there is some JS code (like `loadDiffs` and probably more) which will be deprecated once the request show redesign is rolled out. Could we mark the code with TODOs? This will make the transition smoother.

eduardoj: I added a couple of comments. Please have a look.

dmarcoux (changes requested):
It's somehow broken when the diffs are loading. The URL of the AJAX call is broken, having `undefined` instead of the request number and request action ID.

error

src/api/app/components/sourcediff_component.html.haml
```diff
@@ -0,0 +1,37 @@
+.row
+  .col
+    - if @refresh
+      .clearfix.mb-2.text-center
+        .btn.btn-outline-primary.cache-refresh{ title: 'Refresh results', onclick: 'loadChanges();' }
```
dmarcoux, line 5: The `onclick` in this line is causing the issue I mentioned in my previous review.

eduardoj: Yes. I changed `loadChanges` to not receive any parameter. This should fix it.

dmarcoux (approved):

src/api/app/components/sourcediff_component.rb
```diff
@@ -0,0 +1,22 @@
+class SourcediffComponent < ApplicationComponent
+  attr_accessor :bs_request, :action, :index, :refresh
+
+  delegate :diff_label, to: :helpers
+  delegate :diff_data, to: :helpers
```
danidoni, line 5: Huh, didn't know about this! I just `include WhateverHelperINeed`

danidoni (approved):

src/api/app/components/sourcediff_component.rb
dmarcoux: Including isn't needed, helpers are always available behind `helpers.my_helper_123`. This is documented in our [developer wiki](https://github.com/openSUSE/open-build-service/wiki/View-Components#how-to-use-custom-helpers-in-view-components).

PR #12861, opened 2022-07-26
Display request creator and creation date in request show redesign

dmarcoux:
The reviewers and other details like _Superseded by 123_ will be displayed in the _Overview_ tab.

Preview of the changes:
preview

### DIFF
--- a/src/api/app/views/webui/request/beta_show.html.haml
+++ b/src/api/app/views/webui/request/beta_show.html.haml
@@ -4,11 +4,16 @@
 
 .card
   .card-body.p-0
-    .card-title.px-4.pt-4
-      %h3
-        = @pagetitle
-        %span.badge.ml-1{ class: "badge-#{request_badge_color(@bs_request.state)}" }
-          = @bs_request.state
+    .card-title.px-4.pt-4.mb-0
+      .d-md-flex
+        %h3.mr-md-4
+          = @pagetitle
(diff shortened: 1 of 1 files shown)


dmarcoux:
[Propose/Implement an alternative for the content of small box in overview (M)](https://trello.com/c/jIoq40FI/1935-propose-implement-an-alternative-for-the-content-of-small-box-in-overview-m)

hennevogel:
Design wise, why is this information more important than all the others you will put on overview?

dmarcoux:
> Design wise, why is this information more important than all the others you will put on overview?

There's only a limited amount of information I can fit, unless I have a row for the other details and even then, I don't see how it would fit without overloading the page top section.

rubhanazeem:
> Design wise, why is this information more important than all the others you will put on overview?

We have all the extra information in the overview tab. I think this information can't fit on the top with title
overview

danidoni (approved):

rubhanazeem (approved):

hennevogel:
> without overloading the page top section

That is my question: Why do you put this there and not on the overview page? What's the significance?

dmarcoux:
> Why do you put this there and not on the overview page? What's the significance? Why does this particular piece of information be available everywhere?

I took this decision as I find the request creator to be important information to anyone looking at the request. It also fits nicely there. The request creation date is not as important, but it gives a bit more context on when the request was created.

PR #12864, opened 2022-07-26
Clarify how exceptions are caught in request#add_review

eduardoj:
Also move searching for the request to the beginning of the method, to fail fast and first in case that the request is not found.

See this thread of comments: https://github.com/openSUSE/open-build-service/pull/12844#discussion_r926929787

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -40,27 +40,17 @@ class Webui::RequestController < Webui::WebuiController
   end
 
   def add_reviewer
-    begin
-      opts = {}
-      case params[:review_type]
-      when 'review-user'
-        opts[:by_user] = params[:review_user]
-      when 'review-group'
-        opts[:by_group] = params[:review_group]
-      when 'review-project'
-        opts[:by_project] = params[:review_project]
-      when 'review-package'
-        opts[:by_project] = params[:review_project]
-        opts[:by_package] = params[:review_package]
+    request = BsRequest.find_by_number(params[:number])
+    if request.nil?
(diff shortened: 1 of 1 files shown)


dmarcoux:
Could we rephrase the commit message to `Clarify how exceptions are caught in request#add_review`?

eduardoj:
> Could we rephrase the commit message to `Clarify how exceptions are caught in request#add_review`?

Sure. Done.

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -41,25 +41,17 @@ def create
...
-        opts[:by_project] = params[:review_project]
-        opts[:by_package] = params[:review_package]
-      end
-      opts[:comment] = params[:review_comment] if params[:review_comment]
-
       request = BsRequest.find_by_number!(params[:number])
-      request.addreview(opts)
+    rescue ActiveRecord::RecordNotFound => e
+      flash[:error] = "Unable to add review to request with id '#{params[:number]}': #{e.message}"
+      return
```
dmarcoux, line 47: If somehow `ActiveRecord::RecordNotFound` is raised and caught in this `rescue` block, this `return` will cause the UI to simply not respond. This is quite confusing, see here:

nothing-happening

The solution is either to duplicate the `redirect_to` call on line 57 below or to reorganize the code.

eduardoj: Ops, true. On it.

eduardoj: Given that we are rescuing only `ActiveRecord::RecordNotFound` exceptions, I changed the code from a `find_by_number!` to a `find_by_number`, and later checked with `.nil?`.

dmarcoux (approved):

PR #12871, opened 2022-07-28
Compress CI artifacts

hennevogel:
Use less of our CI credits on storage by compressing logs and
capybara dumps. Also makes downloading things faster...

### DIFF
--- a/.circleci/conditional_config.yml
+++ b/.circleci/conditional_config.yml
@@ -96,28 +96,21 @@ aliases:
     name: Setup application
     command: cd src/api; bundle exec rake dev:bootstrap[old_test_suite] RAILS_ENV=test FORCE_EXAMPLE_FILES=1
 
-  - &store_minitest_artifacts
-    store_artifacts:
-      path: ./src/api/log/
-      destination: minitest
+  - &compress_log_artifacts
+    name: Compress Log Artifacts
+    command: |
+      cd src/api/log
+      for i in *.log; do xz -T 0 -v $i; done
 
-  - &store_minitest_test_results
-    store_test_results:
-      path: /home/frontend/minitest
-
-  - &store_rspec_artifacts
-    store_artifacts:
-      path: ./src/api/log
-      destination: rspec
-
-  - &store_rspec_test_results
-    store_test_results:
-      path: /home/frontend/rspec
-
-  - &store_capybara_artifacts
-    store_artifacts:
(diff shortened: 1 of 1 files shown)


hennevogel:
Screenshot from 2022-07-28 11-54-50

https://app.circleci.com/pipelines/github/openSUSE/open-build-service/7470/workflows/0d96b07c-ac0a-47e7-8c12-e9751c71480a/jobs/90696/artifacts

.circleci/conditional_config.yml
```diff
@@ -251,8 +244,12 @@ jobs:
           root: .
           paths:
             - src/api/coverage_results
-      - <<: *store_rspec_artifacts
-      - <<: *store_rspec_test_results
+      - store_test_results:
+          path: /home/frontend/rspec
+      - run: *compress_log_artifacts
+      - store_artifacts:
```
dmarcoux, line 250: This step is always happening after `compress_log_artifacts`. Could we also extract this like you did with `compress_log_artifacts` and re-use this code wherever needed?

.circleci/conditional_config.yml
```diff
@@ -397,12 +402,16 @@ jobs:
...
-          destination: feature
-      - <<: *store_capybara_artifacts
       - store_test_results:
           path: /home/frontend/feature
+      - run: *compress_log_artifacts
+      - store_artifacts:
+          path: ./src/api/log
+          destination: log
+      - run: *compress_capybara_artifacts
+      - store_artifacts:
```
dmarcoux, line 412: This step is always happening after `compress_capybara_artifacts`. Could we also extract this like you did with `compress_capybara_artifacts` and re-use this code wherever needed?

.circleci/conditional_config.yml
hennevogel: Yeah we could but I' not so sure about the `- <<:` syntax...

dmarcoux: Alright, it can also stay as it is now.

dmarcoux (approved):

.circleci/conditional_config.yml
hennevogel: The problem is that the anchor then doesn't have to reference some simple mapping but an array which needs to get merged with another array. I'm not getting any wiser from the yaml docu...

hennevogel:

PR #12878, opened 2022-07-29
Generalize rate-limiting and other authentication schemes

eduardoj:
Specifics on other authentication schemes and policies on abusing of rate limiting should not be detailed in the general API documentation. This information is dependant on each instance.

Instead, provide a link to the site main page or refer to the OBS admin.

### DIFF
--- a/src/api/public/apidocs-new/OBS-v2.10.50.yaml
+++ b/src/api/public/apidocs-new/OBS-v2.10.50.yaml
@@ -62,7 +62,8 @@ info:
     There is no API versioning, since there is no need for it right now.
 
     Only rudimentary rate limiting is implemented, so please be gentle when using the API concurrently, especially with potentially expensive operations.
-    In case of abuse, we will limit/remove your access.
+
+    Your OBS Admin might have implemented other authentication schemes or stricter rate limiting in front of the application. Check the [site main page](/) for details.
 
     For command-line users, we recommend using [osc](https://github.com/openSUSE/osc) with its _api_ command to interact with the API.
     It's as simple as this example: `osc api /about` (_about_ is one of the endpoints documented below)


src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -32,8 +32,7 @@ info:
 
     There is no API versioning, since there is no need for it right now.
 
-    Only rudimentary rate limiting is implemented, so please be gentle when using the API concurrently, especially with potentially expensive operations.
-    In case of abuse, we will limit/remove your access.
```
hennevogel, line 36: Hm I would not remove this. Just add the sentence below.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -32,8 +32,7 @@ info:
...
+    Other authentication schemes or rate limiting could exist in front of the application. In such case, please refer to the [site main page](/) or talk to your OBS admin.
```
hennevogel, line 35:
```suggestion
    Your OBS Admin might have implemented other authentication schemes or stricter rate limiting in front of the application. Check the the [site main page](/) for details.
```

hennevogel (changes requested):

src/api/public/apidocs-new/OBS-v2.10.50.yaml
eduardoj: This text is documentation that will ship with each instance of the code. As the commit says "Specifics of rate limiting or other authentication schemes should not be detailed in the general API documentation.".

eduardoj: This line "Only rudimentary rate limiting is implemented, so please be gentle when using the API concurrently, especially with potentially expensive operations. In case of abuse, we will limit/remove your access." should be moved to some place dependant on the OBS instance. For example, in the `Description` field which can be modified under `/configuration`:

Screenshot from 2022-07-29 13-12-41

src/api/public/apidocs-new/OBS-v2.10.50.yaml
eduardoj: Applied.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
danidoni: I think it's right to remove that line. There is nothing at application level that matches that statement, so technically that is not true...

danidoni (approved):

PR #12892, opened 2022-08-03
Limit the classes we eager load in development

hennevogel:
To the classes we use in associations


dmarcoux (approved):
The application is now properly starting up.

PR #12858, opened 2022-07-25
Merge comments and history elements in a single timeline

danidoni:
Instead of showing comments and history elements a part we merge them
into one single timeline to have a better overview of the events

This is how it would look like:
timeline_001

Things missing:
- [ ] Outline some specs for the components, so we have a rough idea of the required testing that we'll implement once we are commited with this design.
- [x] Fix the CSS artifacts (we're gonna do it [in this card](https://trello.com/c/YoB5dSQI/1939-fix-history-timeline-css-issue-m)
- [ ] Fix the comments issues (we're gonna do it [in this card](https://trello.com/c/GTxp0rnP/1940-handle-the-new-comments-and-the-replies-correctly-m)

PR #12893, opened 2022-08-03
Do not display the request creator name, only their login

dmarcoux:
To align with how we display users on the request show redesign in the changes from #12858

Before:
before

Now:
now


dmarcoux:
After talking with @danidoni, we prefer to keep the longer version with the user's real name and login. This will be changed in #12858.

PR #12896, opened 2022-08-03
Load the app in production

hennevogel:
Make sure we can boot the app in production.

### DIFF
--- a/.circleci/conditional_config.yml
+++ b/.circleci/conditional_config.yml
@@ -143,6 +143,25 @@ jobs:
       - run: echo "run-dist-linter is << pipeline.parameters.run-dist-linter >>"
       - run: echo "run-haml-linter is << pipeline.parameters.run-haml-linter >>"
 
+  checkout_code:
+    docker:
+      - <<: *frontend_base
+    steps:
+      - checkout
+      - *restore_bundler_cache
+      - *restore_rubocop_cache
+      - run: *install_dependencies
+      - *save_bundler_cache
+      - run: *install_circle_cli
+      - run:
+          name: Setup application
+          command: |
+            cd src/api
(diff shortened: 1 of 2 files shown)


.circleci/conditional_config.yml
```diff
@@ -155,6 +156,9 @@ jobs:
             cd src/api;
             bundle exec rake dev:lint:rubocop:all
       - *save_rubocop_cache
+      - run:
+          name: Test loading the App in production
+          command: RAILS_ENV=production bundle exec rake db:setup zeitwerk:check
```
dmarcoux, line 161: What is the point of doing `db:setup`? `zeitwerk:check` doesn't need the database to ensure classes are loaded properly.

hennevogel: it does. the check some after the check for the DB...

eduardoj (approved):

.circleci/conditional_config.yml
```diff
@@ -336,7 +337,18 @@ jobs:
           name: Run spider
           command: |
             cd src/api
-            bundle exec rake test:spider
+            case $CIRCLE_NODE_INDEX in
+              0)
+                bundle exec rails test test/functional/webui/spider_test.rb -n test_spider_anonymously
+                echo "Test loading the application in production"
```
dmarcoux, line 343: As I already told you @hennevogel, I find this weird to test this in a job (`spider` in this case) which has nothing to do with testing the loading of the application. At the same time, having an extra job only for this is a bit overkill.

dmarcoux (approved):

.circleci/conditional_config.yml
hennevogel: indeed

PR #12879, opened 2022-07-29
Fix missing HistoryElements

danidoni:
In development some `HistoryElements` subtypes do not show when querying the `BsRequest#request_history_elements`.

In development they do not appear because we do not eager load classes, while in production it works flawlessly.

PR #12898, opened 2022-08-03
Fix eager loading token STI classes

danidoni:
While loading the Token STI classes present under
app/models/token/**.rb,

We need to load the sti subclasses under token/**.rb to avoid the
problems specified in #12879. Right, but `app/models/token/errors.rb` is
not an sti subclass, its a module. Its eager loading fails with

    /obs/src/api/app/models/token.rb:18:in `<class:Token>': uninitialized constant Token::Errors (NameError)

    include Token::Errors
                 ^^^^^^^^

Because we didn't load `Token` yet...

See https://github.com/openSUSE/open-build-service/pull/12879


hennevogel:
Guess you missed #12892

danidoni:
Wow, yes, totally missed it

PR #12903, opened 2022-08-04
Run RuboCop cops from ViewComponent namespace only for view components

dmarcoux:
This approach is more efficient since only view components are linted, not all files. It's also much more robust, since it relies on the path, not on class names ending with `Component`.

[rubocop-rspec](https://github.com/rubocop/rubocop-rspec/blob/5416f729d4cd0c676334c356a67240dbd2aafd32/config/default.yml#L2-L7) and [rubocop-graphql](https://github.com/DmitryTsepelev/rubocop-graphql/blob/c9303c6b6c136df8f1e4c1ce4b10006c0f43b9a6/config/default.yml#L1-L4) use the same technique.

For reviewers: If you add `params[:something]` to any view component and run `bundle exec rubocop`, it will raise an offense. It doesn't though for other files. This proves the `Include` configuration works.

### DIFF
--- a/src/api/spec/lib/rubocop/cop/view_component/avoid_global_state_spec.rb
+++ b/src/api/spec/lib/rubocop/cop/view_component/avoid_global_state_spec.rb
@@ -2,9 +2,13 @@ require 'rails_helper'
 require './lib/rubocop/cop/view_component/avoid_global_state'
 
 RSpec.describe RuboCop::Cop::ViewComponent::AvoidGlobalState, :config do
+  let(:config) do
+    RuboCop::Config.new({ 'ViewComponent' => { 'Include' => ['app/components/**/*.rb'] } }, '/some/.rubocop.yml')
+  end
+
   context 'when a view component uses params' do
     it 'registers an offense' do
-      expect_offense(<<~RUBY)
+      expect_offense(<<~RUBY, 'app/components/my_component.rb')
         class MyComponent < ApplicationComponent
           def initialize
             @abc = params[:abc]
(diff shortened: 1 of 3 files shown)


hennevogel (approved):

src/api/spec/lib/rubocop/cop/view_component/avoid_global_state_spec.rb
```diff
@@ -26,16 +26,4 @@ def initialize
       RUBY
     end
   end
-
-  context 'when a class which is not a view component uses params' do
```
dmarcoux, line 30: This example needs to be removed since it doesn't apply anymore. It's not code restricting on which file the cop runs, but the new RuboCop configuration with `Include` defined in this PR.

eduardoj (approved):

dmarcoux:
I have found something which I went to investigate in the specs. It's apparently possible to mock file paths and names.

src/api/spec/lib/rubocop/cop/view_component/avoid_global_state_spec.rb
```diff
@@ -2,9 +2,13 @@
 require './lib/rubocop/cop/view_component/avoid_global_state'
 
 RSpec.describe RuboCop::Cop::ViewComponent::AvoidGlobalState, :config do
+  let(:config) do
+    RuboCop::Config.new({ 'ViewComponent' => { 'Include' => ['app/components/**/*.rb'] } }, '/some/.rubocop.yml')
```
dmarcoux, line 6: We are now mocking the RuboCop config used in this spec file. It proves that the changes are working.

PR #12908, opened 2022-08-05
Use same request show action for beta and non-beta users

eduardoj:


src/api/app/views/webui/request/show.html.haml
```diff
@@ -1,69 +1,116 @@
...
+                                                       number: @bs_request.number,
+                                                       diff_to_superseded: @diff_to_superseded,
+                                                       superseding: @bs_request.superseding }
+  .card.mb-3
+    .card-header.d-flex.justify-content-between
+      %h5
+        Overview
+    .card-body
+      .row
+        .col-md-8#overview
```
eduardoj, line 68: The HAML linter fails now because there are two elements with the `#overview` identifier.

This should not be a problem since one element is in the `if` clause, and the other element is in the `else` clause. :thinking:

eduardoj:
> I don't think we want to use the same view file for both. This is going to be a nightmare to untangle. Why not render something different in the controller depending on `feature_enabled?(:request_show_redesign)`?

Fine for me. With that approach the error of the linter with the #overview identifier will not appear, also.

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -95,9 +93,18 @@ def modify_review
...
 
+    @is_author = @bs_request.creator == User.possibly_nobody.login
     @is_target_maintainer = @bs_request.is_target_maintainer?(User.session)
+
+    reviews = @bs_request.reviews.where(state: 'new')
+    @my_open_reviews = reviews.select { |review| review.matches_user?(User.session) }
+    @can_add_reviews = @bs_request.state.in?([:new, :review]) && (@is_author || @is_target_maintainer || @my_open_reviews.present?)
+
+    # TODO: Remove till the end of the method once request_show_redesign is rolled out
+    return if Flipper.enabled?(:request_show_redesign, User.session)
```
dmarcoux, line 106: I suggest using a good old `if/else`. Under the `if` statement, we do everything we need for the beta version and in the `else` statement, everything which is for the current version. This will create duplication, but it's only temporary. Once we are done with the redesign, the migration will be super simple, we simply remove the `else` part of the code and the duplication is gone.

eduardoj: I applied this suggestion.

I used `Flipper.enabled?(:request_show_redesign, User.session)` instead of `feature_enabled?(:request_show_redesign)` because the last one renders a 404 in case that the feature is not enabled.

@hennevogel, @dmarcoux please take a look again.

danidoni (approved):

dmarcoux (approved):

PR #12911, opened 2022-08-08
Remove unneeded class for heading in request show redesign

dmarcoux:
Nothing is rendered to the right of this heading, so we can safely remove the CSS class.


saraycp (approved):

dmarcoux:
The `linters` check keeps taking a looooong time to finish, like 5 times more than usual (it was stuck even for more than 1 hour before I noticed). I pushed again to see if this solves the issue...

hennevogel (approved):

PR #12919, opened 2022-08-09
Put Iggy in the staff group

hennevogel:
We sometimes limit features to the staff group, we should
have at least one staff user in the development data.


dmarcoux (approved):
Good idea!

eduardoj (approved):

PR #12924, opened 2022-08-09
Make feature configuration an admin thing

hennevogel:
We shouldn't decide in code what is enabled/disabled.
People can do that themselves on /flipper

### DIFF
--- a/src/api/config/initializers/flipper.rb
+++ b/src/api/config/initializers/flipper.rb
@@ -1,8 +1,7 @@
-ENABLED_FEATURE_TOGGLES = [
+FEATURE_TOGGLES = [
   { name: :trigger_workflow, description: 'Better SCM and CI integration with OBS workflows' },
-  { name: :new_watchlist, description: 'New implementation of watchlist including projects, packages and requests' }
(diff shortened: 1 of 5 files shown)


src/api/config/initializers/flipper.rb
```diff
@@ -15,21 +14,17 @@
   end
 
   Flipper.register(:beta) do |user|
-    user.respond_to?(:in_beta?) && user.in_beta?
+    # The user has to be in beta for this group to be active...
+    user.respond_to?(:in_beta?) && user.in_beta? &&
+      # ...and if the user didn't disable the feature, it will be active
+      user.respond_to?(:disabled_beta_features) && !user.disabled_beta_features.exists?(name: feature_toggle_name)
```
dmarcoux, line 20: `feature_toggle_name` doesn't exist in this context. On top of this, the implementation in this PR doesn't work since it's a all-or-nothing approach like we had before for the beta program. We need to have a Flipper group for every feature toggle part of the beta program. Without this, there is no granularity and users of the beta program cannot enable/disable a specific beta feature.

hennevogel:
obviously needs more thought...

PR #12926, opened 2022-08-10
Mention limitations of the request show redesign

dmarcoux:
Preview:

Message at the top of the page + Build Results:
preview-1

Mentioned Issues:
preview-2

### DIFF
--- a/src/api/app/views/webui/request/beta_show.html.haml
+++ b/src/api/app/views/webui/request/beta_show.html.haml
@@ -6,6 +6,21 @@
   = render partial: 'actions'
   = render partial: 'add_reviewer_dialog'
 
+.alert.alert-info{ role: 'alert' }
+  As part of the
+  = succeed ',' do
+    = link_to('beta program', my_beta_features_path)
+  we are currently working on a redesign of this page to help you in your daily work.
+  Certain features are missing like multi-actions requests, but they will be implemented in the future.
+  Give us feedback on this redesign at GitHub by
+  = link_to('opening an issue', 'https://github.com/openSUSE/open-build-service/issues/new/choose')
+  or talk to us on IRC in the
+  %i #opensuse-buildservice
+  on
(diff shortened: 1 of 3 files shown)


src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -6,6 +6,21 @@
   = render partial: 'actions'
   = render partial: 'add_reviewer_dialog'
 
+.alert.alert-info{ role: 'alert' }
+  As part of the
+  = succeed ',' do
```
dmarcoux, line 11: This is needed to avoid `beta program ,`, so now we have `beta program,` as it should be.

dmarcoux:
[Make the beta view the default for people in beta (M)](https://trello.com/c/tuF8tYhx/1943-make-the-beta-view-the-default-for-people-in-beta-m)

src/api/app/views/webui/request/beta_show_tabs/_build_results.html.haml
```diff
@@ -1 +1,7 @@
-Build results
+%p
+  The build results aren't here yet. For now, you can see them on the request's
+  = link_to('source package', package_show_path(project, package))
+  page. We're working on having them here as soon as possible.
```
hennevogel, line 4: This is not the request's page. It's the source package.

I would also not make promises like `as soon as possible` :)

hennevogel (changes requested):

src/api/app/views/webui/request/beta_show_tabs/_build_results.html.haml
dmarcoux: > This is not the request's page. It's the source package.

The link is on `source package`, so this is already the case.

> I would also not make promises like as soon as possible :)

Good point, I pushed some changes.

hennevogel (approved):

PR #12948, opened 2022-08-15
Fix height for comment anchors in request history

rubhanazeem:


rubhanazeem:
https://trello.com/c/m3aezM5E/2681-comment-anchors-dont-account-for-the-height-of-the-top-navbar

rubhanazeem:
[Comment anchors don't account for the height of the top navbar](https://trello.com/c/m3aezM5E/2681-comment-anchors-dont-account-for-the-height-of-the-top-navbar)

src/api/app/assets/javascripts/webui/request.js
```diff
@@ -138,6 +138,10 @@ $(document).ready(function(){
   $('#changes-item').on('shown.bs.tab', function () {
     loadChanges();
   });
+
+  var shiftWindow = function() { scrollBy(0, -57) };
```
dmarcoux, line 142: We definitely a comment here for this, especially for the `-57`. What is this value exactly? The height of the navbar?

rubhanazeem: Yes, the height is 57

hennevogel: Hm, isn't this broken for every anchor on every page?

dmarcoux: Yes and with this approach, it solves the issue, but only partially. For example, if you click on an anchor, then scroll a bit to look at something else, then click on the anchor again, it's broken again:

preview

rubhanazeem: I'm not sure if this is broken for every anchor. Do you have any examples where it is broken outside this page?

dmarcoux: Yes, in the comments on the project#show page. Same for the package#show page.

dmarcoux:
As we discussed, we're closing this. We'll look for another solution.

PR #12954, opened 2022-08-16
Instrument DisabledBetaFeature model

dmarcoux:
With those measurements, we will know which beta features are disabled and how this evolves over the course of time whenever we change those beta features. We will also see spikes in the enabling/disabling of beta features.

### DIFF
--- a/src/api/app/jobs/measurements_job.rb
+++ b/src/api/app/jobs/measurements_job.rb
@@ -8,6 +8,7 @@ class MeasurementsJob < ApplicationJob
     RabbitmqBus.send_to_bus('metrics', "user #{measurements_to_fields(users_measurements)}")
     notifications_measurements
     subscription_measurements
+    beta_features_measurements
   end
 
   private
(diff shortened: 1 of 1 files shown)


src/api/app/jobs/measurements_job.rb
```diff
@@ -76,4 +77,10 @@ def notifications_measurements
       RabbitmqBus.send_to_bus('metrics', "notification_count,notifiable=#{type.first} value=#{type.second}")
     end
   end
+
+  def beta_features_measurements
+    ENABLED_FEATURE_TOGGLES.pluck(:name).each do |feature_name|
+      RabbitmqBus.send_to_bus('metrics', "beta_feature_count,feature=#{feature_name},action=disabled value=#{DisabledBetaFeature.where(name: feature_name).count}")
```
hennevogel, line 83: Hm, why both? We can aggregate this in grafana no?

dmarcoux: True, I didn't consider this.

dmarcoux: I removed the callbacks now.

eduardoj, line 83:
```suggestion
      RabbitmqBus.send_to_bus('metrics', "beta_feature_count,feature=#{feature_name},status=disabled value=#{DisabledBetaFeature.where(name: feature_name).count}")
```

`action` doesn't apply exactly to a count of a beta feature records which are disabled. `status` perhaps?

PR #12958, opened 2022-08-17
Document `/search/project` endpoint with OpenAPI

eduardoj:


dmarcoux:
I tried with the match parameter `@name='home:Admin'`, so it should match the `home:Admin` project, but it returns an empty collection, even if it says `matches="1"`:
```
<collection matches="1">
</collection>
```

Is this expected?

dmarcoux:
This PR should also include the deprecated POST endpoint for the same route, so similar to what has been done in #12963.

dmarcoux:
> This PR should also include the deprecated POST endpoint for the same route, so similar to what has been done in #12963.

We can handle this in a follow-up PR.

dmarcoux:
> I tried with the match parameter `@name='home:Admin'`, so it should match the `home:Admin` project, but it returns an empty collection, even if it says `matches="1"`:
> 
> ```
> <collection matches="1">
> </collection>
> ```
> 
> Is this expected?

After talking with @eduardoj, we are going to track this issue in our backlog. It is not related to the API documentation.

eduardoj:
> This PR should also include the deprecated POST endpoint for the same route, so similar to what has been done in #12963.

Yes, done. Please have another look.

dmarcoux (approved):

dmarcoux:
> I tried to reproduce this issue but I couldn't. Could you confirm this still happens to you?

It's fine after rebuilding my development environment.

PR #12963, opened 2022-08-18
Document deprecated POST search endpoints

dmarcoux:
The details of the POST endpoints aren't listed to avoid having outdated information if we update their GET counterparts.


dmarcoux:
[Deprecate the POST action from /search/* endpoints, in the API router (M)](https://trello.com/c/fKDHHEc7/1958-deprecate-the-post-action-from-search-endpoints-in-the-api-router-m)

eduardoj (approved):

PR #12972, opened 2022-08-21
[backend] add new image format: mkosi

bluca:
Modern image builder from the systemd project:
https://github.com/systemd/mkosi

Required to ensure the source server finds the recipe, and to allow
the full implementation in obs-build.

Required for: https://github.com/openSUSE/obs-build/pull/861

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -2134,7 +2134,7 @@ sub publish {
 	} elsif ($bin =~ /ONIE\.bin(?:\.sha256(?:\.asc)?)?$/) {
 	  $p = "onie/$bin";
 	  $kiwimedium = "$arch/$1" if $bin =~ /(.+)ONIE\.bin$/;
-	} elsif ($bin =~ /(.*)\.raw(?:\.install)?(?:\.(?:gz|bz2|xz))?(?:\.sha256(?:\.asc)?)?$/) {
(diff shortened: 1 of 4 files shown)


src/backend/BSSched/Checker.pm
```diff
@@ -702,7 +702,7 @@ sub expandandsort {
       $buildtype = 'modulemd';
     } elsif ($info && $info->{'file'}) {
       # directly implement most common types
-      if ($info->{'file'} =~ /\.(spec|dsc|kiwi|livebuild)$/) {
+      if ($info->{'file'} =~ /\.(spec|dsc|kiwi|livebuild)$/ || $info->{'file'} =~ m/^(mkosi)\./) {
```
mlschroe, line 705: I think you should drop that change, as this is not that speed critical. There will probably not be tousands of mkosi recipes in the project, so we can just call the always correct Build::recipe2buildtype()

bluca: @mlschroe thanks for your review, tested and removed.

bluca:
@mlschroe gentle ping - added a commit to allow the signing process to be used as well

adrianschroeter (approved):

bluca:
rebased to remove the merge commit, no changes

PR #12974, opened 2022-08-22
Validate limit in OwnerSearch::Base

dmarcoux:
The other search endpoints do not rely on the `limit`parameter, at least not in the same way. They use it for the pagination, so validating the limit only in `OwnerSearch::Base` is fine.


dmarcoux:
[In /search/* endpoints, validate the "limit" query parameter (S)](https://trello.com/c/IpfNMt70/1959-in-search-endpoints-validate-the-limit-query-parameter-s)

rubhanazeem (approved):

eduardoj (approved):

PR #12975, opened 2022-08-22
Address haml-lint offenses for InstanceVariables

dmarcoux:
Those offenses are now reported after updating haml-lint in https://github.com/openSUSE/open-build-service/pull/12973. They weren't caught in that PR, since we run haml-lint only when Haml templates are updated.

This is what haml-lint reported for all the partials updated in this PR:

> InstanceVariables: Avoid using instance variables in not_breadcrumb_items views


dmarcoux:
This is becoming way too large to easily review. I will change my approach, so I'm closing this PR.

dmarcoux:
The follow-up is here: https://github.com/openSUSE/open-build-service/pull/12977

PR #12979, opened 2022-08-22
Provide a general description about the xpath predicate for search endpoints

krauselukas:
Most of the search endpoints use xpath expression. Since all of them
require a xpath predicate passed through the match query parameter
we can mention this in the search group endpoint description with
some little examples.

image


src/api/public/apidocs-new/paths/search_project_id.yaml
```diff
@@ -28,10 +28,6 @@ get:
             - `group/@role`: name of the role which a group could be related with this project.
           - More complex predicates are: `maintenance/maintains/@project`, `repository/@name`,
             `repository/path/@project`, `repository/releasetarget/@trigger`.
-
-        These functions can also be used with predicates: `boolean`, `contains`, `not`, `starts_with`, `ends_with`.
```
dmarcoux, line 32: There is the same documentation in `GET /search/project`. To be consistent, we should also remove it.

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -18,6 +18,30 @@ tags:
...
+      and which extra query parameters are supported or might be required in order to run the search.
+
+
+      These functions can be used with the predicates: `boolean`, `contains`, `not`, `starts_with`, `ends_with`.
+
+      _Example_:
+
+      For a list of all packages with the name "obs-server" that are part of projects where the name starts with "OBS:".
+
+      `/search/package?match=@name='obs-server'+and+starts_with(@project,'OBS:')`
```
eduardoj, line 34:
```suggestion
      For a list of all packages that are part of projects where the name starts with "OBS:".

      `/search/package?match=starts_with(@project,'OBS:')`
```
I suggest to show in this example only the use of a function, without the `and` operator. It's more clear to introduce concepts once at a time.

src/api/public/apidocs-new/paths/search_project_id.yaml
```diff
@@ -28,10 +28,6 @@ get:
...
-
-        More than one predicate can be combined with the following operators: `eq`, `and`, `or`, `neq`, `gt`, `lt`, `gteq`, `lteq`.
```
eduardoj, line 34: With the recent merges of the last pull requests, there are more text like this one that could be removed, from other endpoints.

krauselukas: @eduardoj done :)

src/api/public/apidocs-new/OBS-v2.10.50.yaml
krauselukas: makes sense, done

src/api/public/apidocs-new/paths/search_project_id.yaml
krauselukas: I removed the other mentions of this as well :+1:

eduardoj (approved):

dmarcoux (approved):

PR #12999, opened 2022-08-25
Call `Backend::Test.start` only when necessary

eduardoj:
The `Backend::Test.start` call is only needed when running the minitests. This PR removes the calls to start the test backend from the codebase, not from the tests.

Once the backend test is up, it keeps running until minitests are finished.

Add a call to start the test backend in the first minitest from the group of tests run in the `test:api:group1` group. This makes it possible to run this whole group of tests with a test backend.

### DIFF
--- a/src/api/app/controllers/application_controller.rb
+++ b/src/api/app/controllers/application_controller.rb
@@ -114,7 +114,6 @@ class ApplicationController < ActionController::Base
 
   def volley_backend_path(path)
     logger.debug "[backend] VOLLEY: #{path}"
-    Backend::Test.start
     backend_http = Net::HTTP.new(CONFIG['source_host'], CONFIG['source_port'])
     backend_http.read_timeout = 1000
 
--- a/src/api/app/controllers/search_controller.rb
+++ b/src/api/app/controllers/search_controller.rb
@@ -103,8 +103,6 @@ class SearchController < ApplicationController
   end
 
   def owner
-    Backend::Test.start if Rails.env.test?
-
     if params[:binary].present?
       owners = OwnerSearch::Assignee.new(params).for(params[:binary])
     elsif (obj = owner_group_or_user)
--- a/src/api/app/controllers/webui/search_controller.rb
+++ b/src/api/app/controllers/webui/search_controller.rb
@@ -8,8 +8,6 @@ class Webui::SearchController < Webui::WebuiController
   end
 
   def owner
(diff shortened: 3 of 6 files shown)


hennevogel:
Try to run the minitest suite with a different seed (different order or tests).

You can also try to run single tests as only `Webui::IntegrationTests` start this in `setup` (means only the Spider test).

hennevogel (changes requested):
Try to run the minitest suite with a different seed (different order or tests).

You can also try to run single tests as only Webui::IntegrationTests start this in setup (means only the Spider test).

eduardoj:
> I fear running the minitest suite with a different seed (different order or tests) will lead to a different result.

It seems that different seeds don't make a difference in the order the tests are run. The tests run by the CI are always run in alphabetical order: alphabetically for the name of the file, and inside each file, alphabetically for the name of the test. Even they have different seeds. This can be checked within the minitest runs in other pull requests.

This for the task `test:api:group1`:
Screenshot from 2022-08-26 15-18-16

This for the task `test:api:group2`:
Screenshot from 2022-08-26 15-19-34

> You can also try to run single tests as only Webui::IntegrationTests start this in setup (means only the Spider test).

That would mean that a call to `Backend::Test.start` should be added to that particular test failing. That has been the case for a lot of tests. We shouldn't pollute the codebase which is going to be used in production with calls to the test suite.

hennevogel:
> It seems that different seeds don't make a difference in the order the tests are run. 

On CI not true because we use a file list. Just running with `rake test` will be different...

> That would mean that a call to Backend::Test.start should be added to that particular test failing. That has been the case for a lot of tests.

Sure, I know what the solution is. It's just that you have to untangle things as `Backend::Test.start` has state. Once it's booted it's booted, the next call to it does nothing. I personally also don't really run those tests manually, but others do. Not sure how, probably just a single test file.

> We shouldn't pollute the codebase which is going to be used in production with calls to wake up a service only used in test mode.

We all agree on that 😺

hennevogel (approved):
We decided to go ahead as it is and fix whatever might fail in other conditions later...

hennevogel:
codecov hangs. [coverage is fine](https://app.circleci.com/pipelines/github/eduardoj/open-build-service/2285/workflows/0e002e12-99eb-4612-a803-1cfe0f153e47/jobs/14954). merging!

PR #13000, opened 2022-08-25
Update documentation of search owner endpoint with OpenAPI

saraycp:


src/api/public/apidocs-new/paths/search_owner.yaml
```diff
@@ -1,32 +1,40 @@
...
-  summary: Search for the default responsible person or group (the owner).
+  summary: Search for the users or groups who are the owners of a project or package.
   description: |
-    Search for the default responsible person or group (the owner) by a user, group
-    or binary name.
+    The owners are users/groups with the maintainer or bugowner roles in the given project/package.
+
+    The owner search will get results only if one of the ancestors or related projects of the searched
+    package/project was previously set as the root of the search.
+    In OBS the default root of the search is the project openSUSE.
```
hennevogel, line 8: OBS is not build.opensuse.org, we should not mention specific instances here I think...

saraycp: I rephrased the description and also added some examples. Please have another look.

eduardoj (approved):

src/api/public/apidocs-new/paths/search_owner.yaml
```diff
@@ -1,32 +1,39 @@
 get:
-  summary: Search for the default responsible person or group (the owner).
+  summary: Search for the users or groups who are the owners of a project or package.
   description: |
-    Search for the default responsible person or group (the owner) by a user, group
-    or binary name.
+    The owners are users/groups with the maintainer or bugowner roles in the given project/package.
+
+    The owner search will get results only if you specify a project or if one of the ancestors or
+    related projects of the searched package/project was previously set as the root of the search.
```
hennevogel, line 7:
```suggestion
   By default the owner search will get results from projects (and their sub-projects) that have the OBS:OwnerRootProject attribute.
```

hennevogel: Then add to the rest to the `project` parameter below

```
By default the owner search will get results from projects (and their sub-projects) that have the OBS:OwnerRootProject attribute. Use this parameter to overwrite the default and search in this project (and it's sub-projects).
```

src/api/public/apidocs-new/paths/search_owner.yaml
```diff
@@ -1,32 +1,39 @@
...
+        type: string
+      description: The name of the package whose owners you want to search.
     - in: query
       name: binary
       schema:
         type: string
       description: |
-        The name of a build binary. Either the `binary`, `user` or `group`
-        parameter must be provided.
+        The name of a package's build binary.
```
hennevogel, line 24:
```suggestion
        The name of a binary package whose owners you want to search.
```

src/api/public/apidocs-new/paths/search_owner.yaml
```diff
@@ -1,32 +1,39 @@
...
     - in: query
       name: user
       schema:
         type: string
       description: |
-        The login of a user. Either the `binary`, `user` or `group`
-        parameter must be provided.
+        The login of a user.
```
hennevogel, line 30: This does make much sense? What is this searching?

src/api/public/apidocs-new/paths/search_owner.yaml
```diff
@@ -73,16 +80,70 @@ get:
...
+          examples:
+            only package:
+              summary: Search by package name
+              value:
+                owner:
+                  rootproject: ''
+                  project: games
+                  package: ctris
+                  person:
+                    name: Iggy
```
hennevogel, line 92: 💭  I like examples that work with build.o.o more :)

hennevogel (changes requested):

src/api/public/apidocs-new/paths/search_owner.yaml
hennevogel, line 30:
```suggestion
        Search what this user is owner of.
```

src/api/public/apidocs-new/paths/search_owner.yaml
```diff
@@ -1,32 +1,39 @@
...
-        parameter must be provided.
+        The login of a user.
     - in: query
       name: group
       schema:
         type: string
       description: |
-        The title of a group. Either the `binary`, `user` or `group`
-        parameter must be provided.
+        The title of a group.
```
hennevogel, line 36:
```suggestion
        The title of a group. Search what this group is owner of.
```

hennevogel (changes requested):

src/api/public/apidocs-new/paths/search_owner.yaml
saraycp: I finally used real project/packages names but not user names. I hope you don't mind. My reasons:

- This documentation is not only for build.o.o but for any other instances.
- I prefer people consciously try out the right data. So people don't try out without thinking.
- The real data we use for the examples can be removed at some point from build.o.o, which will force us to update the documentation. 
- For consistency, the rest of the endpoints use fake data.
- I'm not giving values (examples) to the different parameters on purpose. Because there isn't a good way to provide the correct combination of parameters that makes the search work if users simply click on Execute. Instead, I'm adding examples of 200 responses and, based on those examples, people will need to fill in the parameters fields to try out anyway.

If we finally decide that using real data from build.o.o is the right thing, we can adapt the whole documentation at once in a future iteration.

src/api/public/apidocs-new/paths/search_owner.yaml
saraycp: done

src/api/public/apidocs-new/paths/search_owner.yaml
saraycp: done

src/api/public/apidocs-new/paths/search_owner.yaml
saraycp: done

src/api/public/apidocs-new/paths/search_owner.yaml
```diff
@@ -1,32 +1,39 @@
...
   description: |
-    Search for the default responsible person or group (the owner) by a user, group
-    or binary name.
+    The owners are users/groups with the maintainer or bugowner roles in the given project/package.
+
+    By default the owner search will get results from projects (and their sub-projects) that have the
+    OBS:OwnerRootProject attribute.
+
+    To do the search, you can provide the package, the user, the group or a package's binary name
+    in combination with the project.
```
hennevogel, line 10: This sounds as if you *need* the project.

hennevogel, line 10:
```suggestion
    You can search for the package name, a user login, a group title or a binary package name.
    If you don't want to search the default projects, you can provide another project to search in.
```

saraycp:
I've rebased on master

hennevogel (changes requested):

hennevogel (approved):

PR #13004, opened 2022-08-29
Enable to 'force' acceptable staging projects

hennevogel:
The release-tools do this to avoid some race that might occur.

https://github.com/openSUSE/openSUSE-release-tools/commit/a354d6c88206a708014d68ef46f02e238a36d4e3

### DIFF
--- a/src/api/app/models/concerns/staging_project.rb
+++ b/src/api/app/models/concerns/staging_project.rb
@@ -15,7 +15,7 @@ module StagingProject
   end
 
   HISTORY_EVENT_TYPES = [:staging_project_created, :staged_request, :unstaged_request].freeze
-  FORCEABLE_STATES = [:building, :failed, :testing].freeze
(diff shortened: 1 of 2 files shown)


src/api/app/models/concerns/staging_project.rb
```diff
@@ -15,7 +15,7 @@ module StagingProject
   end
 
   HISTORY_EVENT_TYPES = [:staging_project_created, :staged_request, :unstaged_request].freeze
-  FORCEABLE_STATES = [:building, :failed, :testing].freeze
+  FORCEABLE_STATES = [:building, :failed, :testing, :acceptable].freeze
```
dmarcoux, line 18: I wonder why this isn't simply for all states?

hennevogel: Because there are states you don't want to allow to force. Like `:empty`, `:review` or intermediate states like `:accepting`

dmarcoux (approved):

PR #13020, opened 2022-08-31
Display review's project/package name as string

saraycp:
Sometimes the package associated with the review is deleted, i.e. after accepting a deletion request. So we stop retrieving those packages and simply display their names as a string.

We do something similar with the deleted projects. Before we didn't display the associated review at all.

As we can't retrieve the users and their avatars, we display the project or package icon as an avatar in the REVIEW section.

**When the project was deleted:**

Screenshot 2022-09-01 at 12-15-06 Open Build Service

**When the package was deleted:**

Screenshot 2022-09-01 at 12-14-53 Open Build Service

### DIFF
--- a/src/api/app/components/bs_request_overview_avatars_component.rb
+++ b/src/api/app/components/bs_request_overview_avatars_component.rb
@@ -13,12 +13,24 @@ class BsRequestOverviewAvatarsComponent < ApplicationComponent
     @avatar_objects ||= if @review.for_user?
                           [@review.user]
                         elsif @review.for_group?
-                          [@review.group.users, @review.group].flatten
+                          group_avatar_objects
                         elsif @review.for_package?
-                          @review.package.project.users
+                          package_avatar_objects
                         elsif @review.for_project?
-                          @review.project.users
+                          project_avatar_objects
                         end
+  end
+
+  def group_avatar_objects
+    [@review.group.users, @review.group].flatten
+  end
+
+  def package_avatar_objects
+    [@review.package&.project&.users].flatten.compact
+  end
+
+  def project_avatar_objects
+    [@review.project&.users].flatten.compact
   end
 
   def avatars_to_display
--- a/src/api/app/views/webui/request/beta_show_tabs/_review_summary.html.haml
+++ b/src/api/app/views/webui/request/beta_show_tabs/_review_summary.html.haml
@@ -1,28 +1,23 @@
--# FIXME: We don't render anything when the project got deleted because this nullifies the association. It should point to
-  `Project.deleted_instance` or maybe the project got created with the same name again in the meanwhile?
-- unless (review.for_project? || review.for_package?) && review.project.nil?
-  .d-flex.flex-row.pt-1
-    .pl-3
-      - if review.accepted?
-        %i.fas.fa-sm.fa-check.text-primary.align-middle.pr-1
-      - if review.declined?
-        %i.fas.fa-times.text-danger.align-middle.pr-2
-      - if review.new?
-        %i.fas.fa-2xs.fa-circle.text-warning.align-middle.pr-2
+.d-flex.flex-row.pt-1
+  .pl-3
+    - if review.accepted?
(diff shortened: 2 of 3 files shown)


src/api/app/views/webui/request/beta_show_tabs/_review_summary.html.haml
```diff
@@ -1,28 +1,27 @@
...
-      - if review.for_user?
-        = user_with_realname_and_icon(review.by_user, short: true, no_icon: true)
-      - elsif review.for_group?
-        = link_to(review.by_group, group_path(Group.find_by(title: review.by_group)))
-      - elsif review.for_package?
-        = link_to("#{review.by_project} / #{review.by_package}", package_users_path(project: review.by_project,
-          package: review.by_package))
-      - elsif review.for_project?
-        = link_to(review.by_project, project_users_path(project: review.by_project))
+  = render BsRequestOverviewAvatarsComponent.new(review) unless deleted_project || deleted_package
```
dmarcoux, line 14: The logic which is checking if this view component must be rendered or not should be inside the view component itself. This is possible with [render?](https://github.com/openSUSE/open-build-service/wiki/View-Components#conditional-rendering). Otherwise, this logic needs to be duplicated whenever we reuse this component.

dmarcoux (changes requested):

src/api/app/views/webui/request/beta_show_tabs/_review_summary.html.haml
saraycp: Thanks for the reminder! Have another look, please.

saraycp:
I pushed an improvement suggested by @dmarcoux. When the review is associated to a project or package which is deleted, I don't hide the avatars but I display the project/package icon as an avatar.

**When the project was deleted:**

Screenshot 2022-09-01 at 12-15-06 Open Build Service

**When the package was deleted:**

Screenshot 2022-09-01 at 12-14-53 Open Build Service

dmarcoux (approved):

src/api/app/components/bs_request_overview_avatars_component.rb
```diff
@@ -13,14 +13,26 @@ def avatar_objects
     @avatar_objects ||= if @review.for_user?
                           [@review.user]
                         elsif @review.for_group?
-                          [@review.group.users, @review.group].flatten
+                          group_avatar_objects
                         elsif @review.for_package?
-                          @review.package.project.users
+                          package_avatar_objects
```
saraycp, line 18: I had to extract part of the code to methods to avoid high cyclomatic complexity.

src/api/app/components/bs_request_overview_avatars_component.rb
```diff
@@ -13,14 +13,26 @@ def avatar_objects
...
+                          project_avatar_objects
                         end
   end
 
+  def group_avatar_objects
+    [@review.group.users, @review.group].flatten
+  end
+
+  def package_avatar_objects
+    [@review&.package&.project&.users].flatten.compact
```
dmarcoux, line 29: The first `&` isn't needed:

```suggestion
    [@review.package&.project&.users].flatten.compact
```

src/api/app/components/bs_request_overview_avatars_component.rb
```diff
@@ -13,14 +13,26 @@ def avatar_objects
...
+  end
+
+  def project_avatar_objects
+    [@review&.project&.users].flatten.compact
```
dmarcoux, line 33: The first `&` isn't needed, since the `project` association is always defined:

```suggestion
    [@review.project&.users].flatten.compact
```

src/api/app/components/bs_request_overview_avatars_component.rb
saraycp: corrected

src/api/app/components/bs_request_overview_avatars_component.rb
saraycp: corrected

dmarcoux (approved):

PR #13021, opened 2022-08-31
Preparation for Rails 7 - Refactor wrap_parameters.rb initializer

dmarcoux:
We bring back the changes which were reverted in #13019 and this time, we have a custom Rack middleware to prefer XML over HTML whenever the HTTP header HTTP_ACCEPT isn't set in a request.

### DIFF
--- a/src/api/config/initializers/wrap_parameters.rb
+++ b/src/api/config/initializers/wrap_parameters.rb
@@ -2,10 +2,8 @@
 
 # This file contains settings for ActionController::ParamsWrapper which
 # is enabled by default.
-require 'active_support/core_ext/hash/conversions'
 require 'action_dispatch/http/request'
 require 'active_support/core_ext/hash/indifferent_access'
-require 'api_error'
 
 # Disable all default rails parameter parsing
 # Context: https://github.com/rails/rails/blob/39413de44c0e2c0dd2d964be5985b03d8f968a7b/guides/source/3_1_release_notes.md#configinitializerswrap_parametersrb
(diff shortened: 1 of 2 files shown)


src/api/config/initializers/wrap_parameters.rb
```diff
@@ -19,54 +17,44 @@
...
   end
+end
+
+new_parsers = original_parsers.merge({ Mime[:json].symbol => json_parser, Mime[:xml].symbol => xml_parser })
 
+ActionDispatch::Request.parameter_parsers = new_parsers
+
+# Custom Rack middleware to prefer XML over HTML whenever the HTTP header HTTP_ACCEPT isn't set in a request.
+# Without this, clients like `osc` attempt to fetch HTML and they end up getting 404s as OBS has been defaulting to XML
+# since the commit 8145e40a6266f40ce771052fd11e372efc85e116 was introduced many, many years ago. Changing all clients
```
hennevogel, line 44: Uh that was even before the UI & API merge 😲

src/api/config/initializers/wrap_parameters.rb
```diff
@@ -19,54 +17,44 @@
...
+# is a lot of work, so we stick to this default.
+#
+# Details on HTTP_ACCEPT:
+# https://www.rfc-editor.org/rfc/rfc9110.html#name-accept
+class PreferXmlOverHtml
```
hennevogel, line 49: I would prefer to move this out of this file into it's own initializer or rename this one....

hennevogel (changes requested):

src/api/config/initializers/wrap_parameters.rb
dmarcoux: I will move this custom Rack middleware to its own initializer.

dmarcoux: Done, please have another look.

hennevogel (approved):

PR #13022, opened 2022-08-31
Distinguish comments from history elements in the request conversation

dmarcoux:
With bubbles around comments, we distinguish them from history elements in the request conversation. On top of this, we display different icons for some of the history elements, making them stand out even more. For example, when a review was accepted, we show the same checkmark icon as in the _Reviews_ section.

Preview:

Screenshot 2022-09-05 at 17-07-33 Open Build Service

example

### DIFF
--- a/src/api/app/assets/javascripts/webui/comment.js
+++ b/src/api/app/assets/javascripts/webui/comment.js
@@ -45,7 +45,13 @@ $(document).ready(function(){
 
   // This is being used to render a new root comment by the beta request show view
   $('.comment_new').on('ajax:complete', '.post-comment-form', function(_, data) {
-    $(this).closest('.comment_new').prev().append(data.responseText);
+    $(this).closest('.comment_new').prev().append(
+      '<div class="timeline-item">' +
+        '<div class="comments-thread">' +
+          data.responseText +
+        '</div>' +
+      '</div>'
+    );
     $(this).trigger("reset");
   });
 
@@ -88,14 +94,14 @@ $(document).ready(function(){
     });
   });
 
-  $('body').on('click', 'button[id*="edit_button_of_"]', function (e) {
-    var closest = $(e.target).parent().parent().find('button[id*="reply_button_of_"]');
+  $('body').on('click', '[id*="edit_button_of_"]', function (e) {
+    var closest = $(e.target).parent().parent().find('[id*="reply_button_of_"]');
     if (!closest.hasClass('collapsed'))
       closest.trigger('click');
   });
 
-  $('body').on('click', 'button[id*="reply_button_of_"]', function (e) {
-    var closest = $(e.target).parent().parent().find('button[id*="edit_button_of_"]');
+  $('body').on('click', '[id*="reply_button_of_"]', function (e) {
+    var closest = $(e.target).parent().parent().find('[id*="edit_button_of_"]');
     if (!closest.hasClass('collapsed'))
       closest.trigger('click');
   });
--- a/src/api/app/assets/stylesheets/webui/comments.scss
+++ b/src/api/app/assets/stylesheets/webui/comments.scss
@@ -21,6 +21,57 @@
   }
 }
 
+.comment-bubble {
+  $comment-bubble-outline: $gray-400;
+  @extend .rounded;
+  @extend .p-3;
+  @extend .mb-2;
+  border: 1px solid $comment-bubble-outline;
+  position: relative;
+  min-width: 20%;
+
+  // Arrow on the top left of the comment bubble
+  $comment-bubble-arrow-top-position: 0;
+  $comment-bubble-arrow-left-position: 14px;
+  $comment-bubble-arrow-size: 16px;
+  // This creates two triangles.
+  // `::before` creates a triangle which matches the color of the comment bubble's outline.
+  //
+  // `::after` creates a triangle which matches the color of Bootstrap's card background color.
+  // It is displayed on top of the one created by `::before` with a small offset.
+  //
+  // Together, they produce a single comment bubble arrow... Magic!
+  &::before, &::after {
+    content: '';
+    position: absolute;
+    top: $comment-bubble-arrow-top-position;
+    left: $comment-bubble-arrow-left-position;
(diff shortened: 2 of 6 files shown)


src/api/app/assets/stylesheets/webui/comments.scss
```diff
@@ -21,6 +21,49 @@
...
+  $comment-bubble-arrow-top-position: 0;
+  $comment-bubble-arrow-left-position: 16px;
+  $comment-bubble-arrow-size: 16px;
+  // This creates two triangles.
+  // `::before` creates a triangle which matches the color of the comment bubble's outline.
+  //
+  // `::after` creates a triangle which matches the color of Bootstrap's card background color.
+  // It is displayed on top of the one created by `::before` with a small offset.
+  //
+  // Together, they produce a single comment bubble arrow... Magic!
```
dmarcoux, line 44: I've got inspired by some of the CSS [here](https://sharkcoder.com/visual/shapes#section4) and [there](https://sharkcoder.com/visual/shapes#section12). I use Sass variables in order to describe what those values are, hopefully making this easier to understand.

src/api/app/components/bs_request_history_element_component.rb
```diff
@@ -10,4 +10,28 @@ def initialize(element:)
...
+      tag.i(nil, class: 'fas fa-times text-danger')
+    when 'RequestReviewAdded'
+      tag.i(nil, class: 'fas fa-2xs fa-circle text-warning')
+    else
+      tag.i(nil, class: 'fas fa-code-commit text-dark')
+    end
+  end
+
+  # While all history elements possibly have a comment, not all of them are from an actual human...
+  def element_with_comment_from_human?
```
dmarcoux, line 30: For an example of a history element with a comment which isn't from an actual human, see this:

https://github.com/openSUSE/open-build-service/blob/5d51c3548328a27e64991ccba6f3ba1873a3e218/src/api/app/models/bs_request.rb#L454-L455

danidoni (approved):

hennevogel (changes requested):
I still find the different sizes of the `.comment-bubble` make the page look too busy

dmarcoux:
> I still find the different sizes of the `.comment-bubble` make the page look too busy

Smaller comments are replies to a previous comment, so this is why they are smaller. I can make them all the same size, but we would lose this context. I pushed changes to make all comments the same size. Let me know what you think.

danidoni:
> > I still find the different sizes of the `.comment-bubble` make the page look too busy
> 
> Smaller comments are replies to a previous comment, so this is why they are smaller. I can make them all the same size, but we would lose this context. I pushed changes to make all comments the same size. I find this to be confusing, at least when replying to a comment since we render replies directly under the comment for which we're replying to, but without indication why it's like this.
> 
> Let me know what you think.

I prefer the way before this last change, very comments wow :dog:

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -8,8 +8,39 @@
...
+              = link_to('#', data: { toggle: 'modal', target: "#delete-comment-modal-#{comment.id}" },
+                        class: 'dropdown-item delete_link', title: 'Delete comment') do
+                Delete
+
+      = helpers.render_as_markdown(comment)
+
+      - if User.session
+        = link_to("#reply_form_of_#{comment.id}", id: "reply_button_of_#{comment.id}", 'data-toggle': 'collapse',
+                  class: 'd-block mt-2 font-italic small collapsed') do
+          Reply
```
hennevogel, line 31: This could use some block structure. It's just a inline text element in a div. That is also why you have to apply margin-top to it.

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -8,8 +8,39 @@
...
+        .collapse{ id: "reply_form_of_#{comment.id}" }
```
hennevogel, line 32: 💭  UX wise I would expect whatever I add here to be part of the box that you render the form is in...

hennevogel (changes requested):

src/api/app/components/bs_request_comment_component.html.haml
dmarcoux: True, I'll render this form outside of the comment bubble.

src/api/app/components/bs_request_comment_component.html.haml
dmarcoux: This is addressed.

src/api/app/components/bs_request_comment_component.html.haml
dmarcoux: This is addressed.

danidoni (approved):

src/api/app/assets/javascripts/webui/comment.js
```diff
@@ -45,7 +45,13 @@ $(document).ready(function(){
 
   // This is being used to render a new root comment by the beta request show view
   $('.comment_new').on('ajax:complete', '.post-comment-form', function(_, data) {
-    $(this).closest('.comment_new').prev().append(data.responseText);
+    $(this).closest('.comment_new').prev().append(
+      '<div class="timeline-item">' +
```
dmarcoux, line 49: Without this dirty hack, new comments wouldn't render properly since they're missing those two `div`s. To properly address this issue, we need to redo how we create comments in `Webui::CommentsController`. New root comments are handled the same way as replies.

krauselukas:
move_reply_link

@dmarcoux the font size of the reply link is a bit small IMO. How about moving it to the dropdown with the other actions like `Delete`?

dmarcoux:
@krauselukas: While the `Reply` links are somewhat important, they aren't as important as the history elements and comments' headlines (like `user_x wrote 2 days ago`). This is why I slightly decreased the font size of the `Reply` links. This is important for the information hierarchy.

hennevogel (approved):

dmarcoux:
@krauselukas and I talked about his previous comment and we're going to keep the `Reply` links as they are.

PR #13023, opened 2022-08-31
Show Mentioned Issues if there are submit actions

saraycp:
I none of the request's actions is a submit, there won't be any mentioned issues. So we hide the tab.


src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -55,9 +55,10 @@
         %li.nav-item.scrollable-tab-link#changes-item{ 'data-request-number': @bs_request.number, 'data-request-action-id': @actions.first[:id] }
           = link_to('Changes', '#changes', class: 'nav-link text-nowrap', 'aria-controls': 'changes', 'aria-selected': 'false',
                     'data-toggle': 'tab', role: 'tab')
-      %li.nav-item.scrollable-tab-link
-        = link_to('Mentioned Issues', '#mentioned-issues', class: 'nav-link text-nowrap', 'aria-controls': 'mentioned-issues',
-                  'aria-selected': 'false', 'data-toggle': 'tab', role: 'tab')
+      - if @actions.any? { |action| action[:type] == :submit }
```
dmarcoux, line 58: In the "old" request show page, we would display the _Mentioned Issues_ section if they are present in an request action's source diffs. This is where it happens:

https://github.com/openSUSE/open-build-service/blob/66880c8c9b5176bbf606b0c13ba660283f889850/src/api/app/components/sourcediff_tab_component.html.haml#L51

Is it only _submit_ actions which can possibly have _mentioned issues_? I know this is what the [Trello card](https://trello.com/c/k7STvZL6/1966-hide-mentioned-issues-tab-on-bsrequests-without-a-bsrequestactionsubmit-s) says, but I'm unsure if it's right or wrong. We would potentially need to adapt this once more in the future if other request actions also have _mentioned issues_.

coolo: Maintenance incidents also need them for sure. e.g. https://build.opensuse.org/request/show/992584

hennevogel: We are in general currently limit ourselves to submit actions...

src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -55,9 +55,10 @@
         %li.nav-item.scrollable-tab-link#changes-item{ 'data-request-number': @bs_request.number, 'data-request-action-id': @actions.first[:id] }
           = link_to('Changes', '#changes', class: 'nav-link text-nowrap', 'aria-controls': 'changes', 'aria-selected': 'false',
                     'data-toggle': 'tab', role: 'tab')
-      %li.nav-item.scrollable-tab-link
-        = link_to('Mentioned Issues', '#mentioned-issues', class: 'nav-link text-nowrap', 'aria-controls': 'mentioned-issues',
-                  'aria-selected': 'false', 'data-toggle': 'tab', role: 'tab')
+      - if @action[:type].in?(actions_for_diff)
```
dmarcoux, line 58: This check should also be done below where we render the tab's content.

src/api/app/views/webui/request/beta_show.html.haml
saraycp: Thanks for the clarification, @coolo 

Contrary to what the card says, we are not going to limit this to submit actions only.
As we did with the Changes tab, we are going to display the tab if there are diffs involved. So with actions :submit, :delete, :maintenance_incident, :maintenance_release.

The content of the Mentioned Issues tab will be handled later.

src/api/app/views/webui/request/beta_show.html.haml
saraycp: True! done

dmarcoux (approved):

PR #13027, opened 2022-09-01
Consider maintainer role when showing reviewers for review by_package

krauselukas:
Right now we show all users that have a role assigned on a package,
no matter if the role empowers or requires the user to leave a review
or not. We only want to show user that have a reviewer or maintainer
role on the package. Also if the package has a maintainer or reviewer
assigned we first want to show those and only default to the project
one's if they are missing.

## How to test
1. Create `Project A` and inside it create `Package A`.
2. Add `Iggy` as `reviewer`.
3. Add `user_1` as `maintainer`.
4. Create `Package B` inside `Project B`.
5. Add `user_2` as `maintainer` of `Package B`.
6. Add `user_3` as `reviewer` of `Package B`.
7. Branch `Package A`, change something and `Submit Request`.
8. In the request, `Add Review` for a package `Package B`.
9. In the request,  the `Review` section should list: `Iggy` and `user_2`.

### DIFF
--- a/src/api/app/components/bs_request_overview_avatars_component.rb
+++ b/src/api/app/components/bs_request_overview_avatars_component.rb
@@ -26,7 +26,13 @@ class BsRequestOverviewAvatarsComponent < ApplicationComponent
   end
 
   def package_avatar_objects
-    [@review.package&.project&.users].flatten.compact
+    relationships = Relationship.includes(:user, :group)
+    # if the package explicitly has maintainer or reviewer assigned return them
+    reviewers = relationships.for_package(@review.package).user_with_maintainer_role
+    # otherwise, check the related project for responsible user
+    reviewers = relationships.for_project(@review.package.project).user_with_maintainer_role if
+      reviewers.empty?
+    reviewers
   end
 
   def project_avatar_objects
--- a/src/api/app/models/relationship.rb
+++ b/src/api/app/models/relationship.rb
@@ -50,10 +50,19 @@ class Relationship < ApplicationRecord
     joins(:role, :group).order('roles.title, groups.title')
   }
   scope :maintainers, lambda {
-    where(role_id: Role.hashed['maintainer'])
+    where(role: Role.hashed['maintainer'])
   }
+  # FIXME: This probably can be refactored to avoid instantiation inside the `map`
+  scope :user_with_maintainer_role, lambda {
+    where(role: Role.hashed['maintainer'])
+      .map { |relation| relation.user || relation.group.users }
+      .flatten.uniq
+  }
+  scope :for_package, ->(package) { where(package: package) }
+  scope :for_project, ->(project) { where(project: project) }
+
   scope :bugowners, lambda {
-    where(role_id: Role.hashed['bugowner'])
+    where(role: Role.hashed['bugowner'])
   }
 
   scope :bugowners_with_email, lambda {
--- a/src/api/spec/components/bs_request_overview_avatars_component_spec.rb
+++ b/src/api/spec/components/bs_request_overview_avatars_component_spec.rb
@@ -1,0 +1,51 @@
+require 'rails_helper'
+
+RSpec.describe BsRequestOverviewAvatarsComponent, type: :component do
+  describe '#package_avatar_objects' do
+    let(:user) { create(:user, login: 'King') }
+    let(:review) { create(:review, by_project: package.project.name, by_package: package.name) }
+    let(:project) { create(:project_with_package, name: 'Apache', package_name: 'apache2') }
+    let(:package) { project.packages.first }
+    let(:other_user) { create(:user, login: 'bob', realname: 'Bob') }
+
+    context 'when we have a review by_package' do
+      context 'and we have a maintainer assigned to the package' do
+        let(:maintainer) { Role.hashed['maintainer'] }
+
+        before do
+          package.relationships.create(user: other_user, role: maintainer)
+          render_inline(described_class.new(review))
+        end
+
+        it 'renders the maintainers of the package' do
+          expect(rendered_content).to have_selector('img[title="Bob"]', count: 1)
+        end
+      end
+
+      context 'but we do not have a maintainer assigned to the package' do
+        context 'but we have a maintainer on the packages project' do
+          let(:maintainer) { Role.hashed['maintainer'] }
+
+          before do
+            project.relationships.create(user: other_user, role: maintainer)
+            render_inline(described_class.new(review))
+          end
+
+          it 'renders the maintainers of the project' do
+            expect(rendered_content).to have_selector('img[title="Bob"]', count: 1)
+          end
+        end
+
+        context 'and we do not have it anywhere else' do
+          before do
+            render_inline(described_class.new(review))
+          end
+
+          it 'do not render a maintainer' do
+            expect(rendered_content).to have_selector('img[title="Bob"]', count: 0)
+          end
+        end
+      end
+    end
+  end
+end


src/api/app/components/bs_request_overview_avatars_component.rb
```diff
@@ -28,4 +28,13 @@ def avatars_to_display
   def number_of_hidden_avatars
     [0, avatar_objects.size - MAXIMUM_DISPLAYED_AVATARS].max
   end
+
+  def reviewers_for_by_package
```
hennevogel, line 32: This sounds like another Package/Project scope to me :)

src/api/app/mixins/has_relationships.rb
```diff
@@ -81,9 +81,9 @@ def maintainers
     (direct_users + users_in_groups).uniq
   end
 
-  def reviewers
-    direct_users = relationships.with_users_and_roles_query.reviewers.pluck('users.login').map! { |user| User.find_by_login!(user) }
-    users_in_groups = relationships.with_groups_and_roles_query.reviewers.pluck('groups.title')
+  def maintainers_and_reviewers
```
hennevogel, line 84: I think you can do all of this before instantiation with a scope and [.distinct](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-distinct). Not that we'll have many instantiations to throw away but it might be more elegant and outside this weird mixin...

src/api/app/components/bs_request_overview_avatars_component.rb
```diff
@@ -40,4 +36,15 @@ def avatars_to_display
   def number_of_hidden_avatars
     [0, avatar_objects.size - MAXIMUM_DISPLAYED_AVATARS].max
   end
+
+  def package_avatar_objects
+    relationships = @review.package&.relationships.for_maintainer_and_reviewer_roles.includes(:user, :group)
+    reviewers = relationships.map { |relation | relation.user_id.present? ? relation.user : relation.group.users }.flatten.uniq!
```
danidoni, line 42: This should be pushed to the `Review` model, we should tell `Review` to give us it's reviewers instead of asking too many details from it to do it by ourselves.

src/api/app/components/bs_request_overview_avatars_component.rb
```diff
@@ -40,4 +36,15 @@ def avatars_to_display
...
+    # if the package explicitly has maintainer or reviewer assigned return them, otherwise
+    # check the related project for responsible user
+    return reviewers if reviewers.present?
+
+    relationships = @review.package&.project&.relationships.for_maintainer_and_reviewer_roles.includes(:user, :group)
+    relationships.map { |relation | relation.user_id.present? ? relation.user : relation.group.users }.flatten.uniq!
```
danidoni, line 48: I belive this can be pushed to `Review` too, as noted above.

src/api/app/components/bs_request_overview_avatars_component.rb
danidoni: Also, I'm not sure we want to use`uniq!` (see [the api doc](https://rubydoc.info/stdlib/core/3.0.0/Array:uniq!)) as it returns `nil` if no duplicates are found instead returning the original array...

src/api/app/components/bs_request_overview_avatars_component.rb
```diff
@@ -26,7 +26,12 @@ def group_avatar_objects
   end
 
   def package_avatar_objects
-    [@review.package&.project&.users].flatten.compact
+    reviewers = @review.maintainers_and_reviewers_for_package
+    # if the package explicitly has maintainer or reviewer assigned return them, otherwise
+    # check the related project for responsible user
+    return reviewers if reviewers.present?
+
+    @review.maintainers_and_reviewers_for_project
```
rubhanazeem, line 34: Is it possible there are no reviewers or maintainers present for a project?

danidoni: It is possible, yes

src/api/app/components/bs_request_overview_avatars_component.rb
```diff
@@ -25,6 +25,14 @@ def group_avatar_objects
     [@review.group.users, @review.group].flatten
   end
 
+  def package_avatar_objects
+    # if the package explicitly has maintainer or reviewer assigned return them
+    reviewers = MaintainersAndReviewersFinder.new(@review).for_package
+    # otherwise, check the related project for responsible user
+    reviewers = MaintainersAndReviewersFinder.new(@review).for_project if reviewers.present?
```
rubhanazeem, line 32: shouldn't we check it `unless reviewers.present?`

src/api/app/components/bs_request_overview_avatars_component.rb
rubhanazeem: In that case, we can return users who are not reviewers or maintainers.

src/api/app/components/bs_request_overview_avatars_component.rb
danidoni: Ah, that's true. Fixing

src/api/app/components/bs_request_overview_avatars_component.rb
krauselukas: @rubhanazeem I would not return anything else than reviewers or maintainers. If both package and project don't have any reviewer or maintainers assigned there is no one who could review. To me thats a sign that the package or project is not maintained anymore. So just showing the reference to the package is fine IMO

krauselukas:
Besides this https://github.com/openSUSE/open-build-service/pull/13027#discussion_r965713889 it looks good to me @danidoni

src/api/app/models/relationship.rb
```diff
@@ -50,10 +50,19 @@ class Relationship < ApplicationRecord
     joins(:role, :group).order('roles.title, groups.title')
   }
   scope :maintainers, lambda {
-    where(role_id: Role.hashed['maintainer'])
+    where(role: Role.hashed['maintainer'])
   }
+  # FIXME: This probably can be refactored to avoid instantiation inside the `map`
+  scope :for_maintainer_and_reviewer_roles, lambda {
+    where(role: Role.hashed['maintainer'])
```
krauselukas, line 57: @danidoni this doesn't consider the `reviewer` role anymore

hennevogel: But you for sure should change the name...

krauselukas: But only for the reviewers of the target package of the request right? So if I add a review `by_package` for a package that is not the target of the request, the reviewer won't appear. I don't know if that is a use case

krauselukas: Not sure, but but lets say the request targets package "apache2" and I add a review `by_package` for package `apache-rex` it will only show the reviewers for package `apache2`. If thats a real life scenario I would say we need to include the `reviewer` role in the scope.

danidoni: This is how it works now:
- It shows **only** reviewers from request target's package + project not maintainers
- It shows **only maintainers** from explicit reviews (by package or by project) not reviewers

Is this how it should work?

danidoni: This is how it works now:
- It shows **only** reviewers from request target's package + project not maintainers
- It shows **only maintainers** from explicit reviews (by package or by project) not reviewers

Is this how it should work?

krauselukas: As discussed offline, for now we just consider the maintainer role for the review `by_package`

src/api/app/models/relationship.rb
```diff
@@ -50,10 +50,19 @@ class Relationship < ApplicationRecord
     joins(:role, :group).order('roles.title, groups.title')
   }
   scope :maintainers, lambda {
-    where(role_id: Role.hashed['maintainer'])
+    where(role: Role.hashed['maintainer'])
   }
+  # FIXME: This probably can be refactored to avoid instantiation inside the `map`
+  scope :users_and_group_users_for_maintainer_role, lambda {
```
hennevogel, line 56:
```suggestion
  scope :for_maintainer_role, lambda {
```

krauselukas: done

hennevogel (approved):

krauselukas:
codecov hangs but is green, merging...

src/api/spec/components/bs_request_overview_avatars_component_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    let(:project) { create(:project_with_package, name: 'Apache', package_name: 'apache2') }
+    let(:package) { project.packages.first }
+    let(:other_user) { create(:user, login: 'bob', realname: 'Bob') }
+
+    context 'when we have a review by_package' do
+      context 'and we have a maintainer assigned to the package' do
+        let(:maintainer) { Role.hashed['maintainer'] }
+
+        before do
+          package.relationships.create(user: other_user, role: maintainer)
```
eduardoj, line 16: Nitpick: we could use the [relationship factory](https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/factories/relationship.rb)...

src/api/spec/components/bs_request_overview_avatars_component_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+          expect(rendered_content).to have_selector('img[title="Bob"]', count: 1)
+        end
+      end
+
+      context 'but we do not have a maintainer assigned to the package' do
+        context 'but we have a maintainer on the packages project' do
+          let(:maintainer) { Role.hashed['maintainer'] }
+
+          before do
+            project.relationships.create(user: other_user, role: maintainer)
```
eduardoj, line 30: Nitpick: also here, we could use the [relationship factory](https://github.com/openSUSE/open-build-service/blob/master/src/api/spec/factories/relationship.rb)...

src/api/spec/components/bs_request_overview_avatars_component_spec.rb
danidoni: Hm, yeah

PR #13029, opened 2022-09-01
Handle multiple BsRequestActionSubmit actions

rubhanazeem:
In the beta request show page, we introduced a dropdown to select among different request actions.

How to test?

- Create a request with multiple actions:
`osc --apiurl=http://localhost:3000/ creq -a submit home:Admin hello_world home:Iggy -a submit home:Admin ruby home:Iggy`
or use this
`osc --apiurl=http://localhost:3000 creq $(osc --apiurl=http://localhost:3000 list home:Admin | xargs -I % echo "-a submit home:Admin % home:Iggy" | tr "\n" " ")`

- Open the last request show page in the web UI (with the `request_show_redesign` beta feature enabled)
- You have to see a dropdown on the left. Change between actions.
- You will see some changes
  - The text under the title of the request (the one describing the action)
  - The Changes tab content changes

### DIFF
--- a/src/api/app/assets/javascripts/webui/request.js
+++ b/src/api/app/assets/javascripts/webui/request.js
@@ -135,6 +135,33 @@ $(document).ready(function(){
     }
   });
   // TODO: Remove the enclosing code when the request_show_redesign feature is finished - END
+
+  // NOTE: Add value of selected tab in dropdown urls
+  $('#request-tabs .nav-item .nav-link').on('click', function() {
+    var tabName = $(this).attr('aria-controls');
+    $.each($('.dropdown-menu .dropdown-item'), function(){
+      var href = $(this).attr('href');
+      if(href){
+        var regex = /tab_name=([aA-zZ-]+)/;
+        var matches = href.match(regex);
+        if(matches !== null){
+          href = href.replace(matches[0], 'tab_name=' + tabName);
+        } else if(href.match(/\?/g)){
+          href = href + '&tab_name=' + tabName;
+        } else {
+          href = href + '?tab_name=' + tabName;
+        }
+        $(this).attr('href', href);
+      }
+    });
+  });
+
+  var selectedTab = $('.selected-tab').data('selected');
+  if(document.getElementById(selectedTab + '-nav-item')){
+    $('#'+selectedTab + '-nav-item a').click();
+  } else{
+    $('#conversation-nav-item a').click();
+  }
 });
 
 // TODO: Remove the following method when the request_show_redesign feature is finished
--- a/src/api/app/assets/javascripts/webui/request.js
+++ b/src/api/app/assets/javascripts/webui/request.js
@@ -135,9 +135,6 @@ $(document).ready(function(){
     }
   });
   // TODO: Remove the enclosing code when the request_show_redesign feature is finished - END
-  $('#changes-item').on('shown.bs.tab', function () {
-    loadChanges();
-  });
 });
 
 // TODO: Remove the following method when the request_show_redesign feature is finished
@@ -165,20 +162,6 @@ function loadDiffs(element){
   $.ajax({
     url: url,
     success: function(){
-      $('.loading-diff').addClass('invisible');
-    }
-  });
-}
-
-function loadChanges() {
-  $('.loading-diff').removeClass('invisible');
-  var element = $('#changes-item');
-  // Always retrieve the changes of the first request action, by now
-  // TODO: request-action-id should be retrieved from the select component, when it is introduced
-  var url = '/request/' + element.data('request-number') + '/request_action/' + element.data('request-action-id') + '/changes';
-  $.ajax({
-    url: url,
-    success: function() {
       $('.loading-diff').addClass('invisible');
     }
   });
--- a/src/api/app/components/request_decision_component.rb
+++ b/src/api/app/components/request_decision_component.rb
@@ -1,11 +1,10 @@
 class RequestDecisionComponent < ApplicationComponent
-  def initialize(bs_request:, actions:, is_target_maintainer:, is_author:)
+  def initialize(bs_request:, action:, is_target_maintainer:, is_author:)
     super
 
     @bs_request = bs_request
-    @actions = actions
     @is_target_maintainer = is_target_maintainer
-    @action = @actions.first
+    @action = action
     @is_author = is_author
   end
 
@@ -14,7 +13,7 @@ class RequestDecisionComponent < ApplicationComponent
   end
 
   def single_action_request
-    @actions.count == 1
+    @single_action_request ||= @bs_request.bs_request_actions.count == 1
   end
 
   def confirmation
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -103,7 +103,7 @@ class Webui::RequestController < Webui::WebuiController
       @diff_to_superseded_id = params[:diff_to_superseded]
 
       # Handling request actions
-      action_id = params[:action_id] || @bs_request.bs_request_actions.first.id
+      action_id = params[:request_action_id] || @bs_request.bs_request_actions.first.id
       @action = @bs_request.webui_actions(filelimit: @diff_limit, tarlimit: @diff_limit, diff_to_superseded: @diff_to_superseded,
                                           diffs: true, action_id: action_id.to_i, cacheonly: 1).first
       @active_action = @bs_request.bs_request_actions.find(action_id)
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -114,6 +114,7 @@ class Webui::RequestController < Webui::WebuiController
       @open_reviews_for_staging_projects = @bs_request.reviews.opened.for_staging_projects
       @not_full_diff = BsRequest.truncated_diffs?([@action])
       @refresh = @action[:diff_not_cached]
+      @active_tab = params[:tab_name] || 'conversation'
 
       if @refresh
         bs_request_action = BsRequestAction.find(@action[:id])
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -1,11 +1,11 @@
 class Webui::RequestController < Webui::WebuiController
   helper 'webui/package'
 
-  before_action :require_login, except: [:show, :sourcediff, :diff, :request_action, :request_action_changes]
+  before_action :require_login, except: [:show, :sourcediff, :diff, :request_action]
   # requests do not really add much value for our page rank :)
   before_action :lockout_spiders
-  before_action :require_request, only: [:changerequest, :show, :request_action, :request_action_changes]
-  before_action :set_superseded_request, only: [:show, :request_action, :request_action_changes]
+  before_action :require_request, only: [:changerequest, :show, :request_action]
+  before_action :set_superseded_request, only: [:show, :request_action]
   before_action :check_ajax, only: :sourcediff
 
   after_action :verify_authorized, only: [:create]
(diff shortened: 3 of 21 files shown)


rubhanazeem:
[Handle multiple BsRequestActionSubmit actions (L)](https://trello.com/c/C4vf49ZI/1964-handle-multiple-bsrequestactionsubmit-actions-l)

src/api/app/assets/javascripts/webui/request.js
```diff
@@ -135,9 +135,6 @@ $(document).ready(function(){
     }
   });
   // TODO: Remove the enclosing code when the request_show_redesign feature is finished - END
-  $('#changes-item').on('shown.bs.tab', function () {
```
danidoni, line 138: The `request_show_redesign` feature is finished already? That's why we remove this code?

saraycp: No, but we don't need this code anymore after we introduced our changes. The reason is that we no longer call Ajax but we load the page and display the content of one action at a time.

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -101,13 +101,25 @@ def show
 
       @diff_limit = params[:full_diff] ? 0 : nil
       @diff_to_superseded_id = params[:diff_to_superseded]
-      @actions = @bs_request.webui_actions(filelimit: @diff_limit, tarlimit: @diff_limit, diff_to_superseded: @diff_to_superseded, diffs: false)
-      @action = @actions.first
+
+      # Handling request actions
+      action = @bs_request.bs_request_actions.first
```
danidoni, line 106: :thought_balloon: You can merge those two lines as `action` is not used anywhere else, right?

dmarcoux:
We should have at least one request with multiple actions in the development data here: https://github.com/openSUSE/open-build-service/blob/a18d10329515b844b9d0948c76ae10d6b24dbb5d/src/api/lib/tasks/dev.rake#L248

This way, it's easier for everyone to review. Otherwise, everyone needs to create a request with multiple actions locally over and over again.

hennevogel:
> We should have at least one request with multiple actions in the development data here

I think you should also have one with mixed action types.

src/api/app/models/bs_request_action_change_devel.rb
```diff
@@ -30,6 +30,10 @@ def uniq_key
     "changedevel/#{target_project}/#{target_package}"
   end
 
+  def name
```
hennevogel, line 33: In the method above you see what makes this action uniq. It's also how we should identify this action in the dropdown. Same for all the other action names.

src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -42,17 +42,32 @@
...
+      .col-md-10.col-sm-12
+        %p.px-4= request_action_header(@action, @bs_request.creator)
+      - if @bs_request.bs_request_actions.count > 1
+        .col-md-2.col-sm-10
+          .dropdown.d-flex.justify-content-end.mr-4
+            %button.btn.btn-secondary.dropdown-toggle{ 'aria-expanded' => 'false', 'data-toggle' => 'dropdown', :type => 'button' }
+              = @action[:name]
+            .dropdown-menu
+              - @bs_request.bs_request_actions.each do |action|
+                = link_to(action.name,
```
hennevogel, line 55: Please consider tab anchors in this link.

hennevogel (changes requested):

hennevogel:
I find it confusing that the "overview" doesn't mention that there are more actions.

Screenshot from 2022-09-07 13-06-37

src/api/app/models/bs_request_action.rb
```diff
@@ -822,6 +822,11 @@ def uniq_key
     ' '
   end
 
+  # This is duplication of part of the code in the BsRequest#action_details
```
dmarcoux, line 825: I don't understand this comment. What do you mean?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -101,13 +101,25 @@ def show
...
+      action_id = params[:action_id] || action.id
```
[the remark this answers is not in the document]
rubhanazeem: This is already optional. It's part of the query string. Can you please explain a bit more?

src/api/app/models/bs_request_action_change_devel.rb
rubhanazeem: Okay, we are now using the `uniq_key` but in a modified version that humanizes the keys

src/api/app/models/bs_request_action.rb
saraycp: We droped this comment, it makes no sense anymore.

saraycp:
After trying out different approaches, I propose this change in the UI.

**Previous proposal:**
Screenshot 2022-09-07 at 20-04-50 Open Build Service

**New proposal:**
Screenshot 2022-09-07 at 20-01-59 Open Build Service

What do you think? I still need to adapt some alignment and make it work for different viewports. 

I opened a new commit on top which should be squash as soon as the proposal is approved or discarded.

src/api/app/components/request_decision_component.rb
```diff
@@ -14,7 +13,7 @@ def render?
   end
 
   def single_action_request
-    @actions.count == 1
+    @bs_request.bs_request_actions.count == 1
```
dmarcoux, line 16: This should be memoized, otherwise you'll fetch the actions over and over again when this method is called.

```suggestion
    @single_action_request ||= @bs_request.bs_request_actions.count == 1
```

src/api/app/views/webui/request/beta_show.html.haml
dmarcoux: Yes, otherwise when we select an action, the selected tab is lost.

select-action

dmarcoux:
When refreshing the page, the selected tab is lost since the anchor (like `#changes`) isn't in the URL. This shouldn't be the case.

page-refresh

src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -26,33 +26,50 @@
...
+        Created by
+        = user_with_realname_and_icon(@bs_request.creator)
+        = fuzzy_time(@bs_request.created_at)
+        - if @bs_request.superseding.present?
+          = '. Supersedes'
+          - @bs_request.superseding.each do |supersed|
+            = link_to "##{supersed['number']}", number: supersed['number']
+
+      - if @bs_request.bs_request_actions.count > 1
+        %p This request contains multiple actions. Select one:
```
dmarcoux, line 47: The text `Select one:` is superfluous here, since the dropdown's text has the same call to action with `Select action`. We can remove `Select one:`.

src/api/app/views/webui/request/beta_show.html.haml
rubhanazeem: I've addressed it. It wasn't possible by just using anchors because there is JS involved that helps in activating the tabs

src/api/app/components/request_decision_component.rb
rubhanazeem: It's updated

src/api/app/views/webui/request/beta_show.html.haml
rubhanazeem: I've changed the text. Now it shows the same value as it is available in the dropdown menu

src/api/app/controllers/webui/request_controller.rb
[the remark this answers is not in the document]
hennevogel: https://guides.rubyonrails.org/routing.html#dynamic-segments

src/api/app/views/webui/request/beta_show.html.haml
saraycp: @rubhanazeem I think the idea is to remote "Select one:" from the text before the dropdown. Not from the dropdown.

saraycp:
> You now have some of the information (`type`, `source`) twice
> 
> some other information (`target`) that would help me to decide which action I want to view I only get _after_ I selected it.

@rubhanazeem I think the idea was to remove "Select one:" from the text before the dropdown. Not from the dropdown. See https://github.com/openSUSE/open-build-service/pull/13029#discussion_r965765079

src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -26,56 +26,78 @@
...
+          = '. Supersedes'
+          - @bs_request.superseding.each do |supersed|
+            = link_to "##{supersed['number']}", number: supersed['number']
+
+      - if @bs_request.bs_request_actions.count > 1
+        %p This request contains multiple actions
+
+        .d-flex
+          .dropdown.ml-2
+            %button.btn.btn-secondary.dropdown-toggle{ 'aria-expanded' => 'false', 'data-toggle' => 'dropdown', :type => 'button', 'data-boundary' => 'viewport' }
```
hennevogel, line 51:
```suggestion
            %button.btn.dropdown-toggle{ 'aria-expanded' => 'false', 'data-toggle' => 'dropdown', :type => 'button', 'data-boundary' => 'viewport' }
```

hennevogel: I find it weird that it's using the secondary color...

rubhanazeem: I've removed this class. It was used to clearly distinguish the text from the dropdown

src/api/app/models/bs_request_action.rb
```diff
@@ -822,6 +822,25 @@ def uniq_key
     ' '
   end
 
+  def name
+    uniq_key
+  end
+
+  def details
```
hennevogel, line 829: All of this sounds like view code to me...

src/api/app/controllers/webui/request_controller.rb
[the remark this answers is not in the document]
saraycp: I added a commit to handle this. I don't use Dynamic Segments but [Bound Parameters](https://guides.rubyonrails.org/routing.html#bound-parameters).

Changing the route like this ensures breaking nothing for the old UI and working for the new UI in these cases:

- when no action is passed: `request/show/:number` and `request/show/:number/request_action`
- when the action is passed: `request/show/:number/request_action/:request_action_id`

Is this what you had in mind?

PS: I need to adapt the code that keeps the same tab open, it broke after I applied my changes. On it.

src/api/app/models/bs_request_action.rb
```diff
@@ -822,6 +822,23 @@ def uniq_key
...
```
hennevogel, line 829: This should be named `commit_details`

src/api/app/models/bs_request_action.rb
```diff
@@ -822,6 +822,23 @@ def uniq_key
...
+    package = Package.find_by_project_and_name(source_project, source_package)
+
+    return { comment: 'Commit information is missing' } if package.nil? || source_rev.nil?
+
+    hash = package.commit(source_rev) || package.commit
+
+    { comment: hash['comment']&.truncate(50) || name,
+    time: "#{ApplicationController.helpers.time_ago_in_words(hash['time'].to_i)} ago",
+    user: hash['user'],
+    rev: hash['rev'] }
```
hennevogel, line 839: I still don't like the mangling of the hash here. This is view code. I think this method should return `nil`, `package.commit(source_rev)` or `package.commit`. And all the string mangling you can do in the view/component.

saraycp: done

src/api/app/models/bs_request_action.rb
saraycp: done

src/api/app/models/bs_request_action.rb
saraycp: Part of this code is now moved to the view

krauselukas (changes requested):
The file extension of `src/api/app/views/webui/request/_action_text.haml` should be `.html.haml`

saraycp:
I already pushed some changes: fixing linter complaints and changing the partial extension.

src/api/app/views/webui/request/_action_text.html.haml
```diff
@@ -0,0 +1,10 @@
+- if (details = action.commit_details)
+  %p.mt-0.mb-0.font-weight-normal
+    = details['comment']&.truncate(50) || action.name
+    %span.float-right.ml-5
+      = details['rev']
```
krauselukas, line 5: Screenshot 2022-09-12 at 10-55-44 Open Build Service

To me its not clear on the first glance what the revision number means. I just saw the number 2 and started thinking what it possibly means. But could be a follow up to somehow better indicate what this number is about

saraycp: It didn't take too long, so we decided to improve this a bit. This is how it looks like now:

Screenshot 2022-09-12 at 11-23-20 Open Build Service

src/api/app/views/webui/request/beta_show.html.haml
krauselukas: Anchors are working, so I would say this is addressed

src/api/app/models/bs_request_action.rb
```diff
@@ -822,6 +822,25 @@ def uniq_key
...
+  end
+
+  def details
+    package = Package.find_by_project_and_name(source_project, source_package)
+    if package
+      hash = package.commit(source_rev)
+      { comment: hash['comment']&.truncate(50) || name,
+        time: "#{ApplicationController.helpers.time_ago_in_words(hash['time'].to_i)} ago",
+        user: hash['user'],
+        srcmd5: hash['srcmd5'].first(7) }
```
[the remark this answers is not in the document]
krauselukas: This got addressed :)

src/api/app/models/bs_request_action.rb
krauselukas: :+1:

src/api/app/controllers/webui/request_controller.rb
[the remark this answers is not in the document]
saraycp: FYI: tab issue was fixed as well.

src/api/app/views/webui/request/_action_text.html.haml
krauselukas: Cool, much better

src/api/app/controllers/webui/request_controller.rb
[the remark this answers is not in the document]
krauselukas: looks good to me

src/api/app/models/bs_request_action_add_role.rb
```diff
@@ -51,6 +51,10 @@ def render_xml_attributes(node)
     node.group(name: group_name, role: role)
   end
 
+  def name
```
krauselukas, line 54: I didn't managed to create a multi action request for `add_role`. Is this possible? Maybe with some of the other action types its not possible as well and the code is not needed? I'm also happy to check this in an follow up

krauselukas (approved):

krauselukas:
Codecov hangs but is green and Henne's review got addressed, I will merge

PR #13033, opened 2022-09-02
Replace instance variable by passed local variable

saraycp:
Change missing from previous commit [d1210ac](https://github.com/openSUSE/open-build-service/commit/d1210aca842c01795e262602de6878d137c5e387) where the goal was stopping using `@linkinfo` inside the partial to use `linkinfo` instead.

### DIFF
--- a/src/api/app/views/webui/package/_revision_diff_detail_buttons.html.haml
+++ b/src/api/app/views/webui/package/_revision_diff_detail_buttons.html.haml
@@ -7,9 +7,7 @@
   href: "#revision_details_#{index + 1}",
   title: 'Go down to the next diff' }
     %i.fas.fa-chevron-down
-  -# haml-lint:disable InstanceVariables
-  - if file_view_path && has_content && !@linkinfo && viewable_file?(filename)
-    -# haml-lint:enable InstanceVariables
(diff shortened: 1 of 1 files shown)


src/api/app/views/webui/package/_revision_diff_detail_buttons.html.haml
```diff
@@ -8,7 +8,7 @@
   title: 'Go down to the next diff' }
     %i.fas.fa-chevron-down
   -# haml-lint:disable InstanceVariables
```
dmarcoux, line 10: This comment and the other one on line 12 should be removed.

src/api/app/controllers/webui/request_controller copy.rb
```diff
@@ -0,0 +1,436 @@
+class Webui::RequestController < Webui::WebuiController
```
dmarcoux, line 1: This isn't supposed to be in this PR, right?

saraycp: No! Sorry about that.
I pushed the correct changes already

dmarcoux (approved):

PR #13036, opened 2022-09-02
Add Dario to .gitmessage

rubhanazeem:


hennevogel (approved):

.gitmessage
```diff
@@ -29,3 +29,4 @@
 # Co-authored-by: Marco Strigl <mstrigl@suse.com>
 # Co-authored-by: Michael Schroeder <mls@suse.de>
 # Co-authored-by: Frank Schreiner <schreiner@suse.de>
+# Co-authored-by: Dario Leidi <dleidi@suse.de>
```
hennevogel, line 32: This doesn't match the address Dario is using in commits....

hennevogel (changes requested):

.gitmessage
rubhanazeem: Thanks for checking it out. It's updated

hennevogel (approved):

PR #13038, opened 2022-09-02
Stop handling linkinfo in sourcediff components

saraycp:
The partial `_revision_diff_detail` is called from the package's controller action `rdiff` (alias `package_rdiff`) or from the request view.

In the first situation, we are setting @linkinfo on the [package controller](https://github.com/openSUSE/open-build-service/blob/334ae03e608468a85b152ce4310571cfc3bb7630/src/api/app/controllers/webui/package_controller.rb#L256) and then it is passed to the partial.

In the second situation, we never set @linkinfo on the [request controller](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/request_controller.rb).
So there is no need to pass @linkinfo to the component in order to always end up passing nil to the partial.

The changes were introduced in [this commit](https://github.com/openSUSE/open-build-service/commit/d1210aca842c01795e262602de6878d137c5e387) and this [follow-up PR](https://github.com/openSUSE/open-build-service/pull/13033).

### DIFF
--- a/src/api/app/components/sourcediff_component.html.haml
+++ b/src/api/app/components/sourcediff_component.html.haml
@@ -31,7 +31,7 @@
                                                                                   index: file_index,
                                                                                   last: sourcediff['filenames'].last == filename,
                                                                                   package: @action[:spkg],
-                                                                                  linkinfo: linkinfo }
+                                                                                  linkinfo: nil }
           - else
             .mb-2
               %p.lead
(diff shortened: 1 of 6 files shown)


saraycp:
@dmarcoux do you agree or am I misunderstanding something?

dmarcoux:
> @dmarcoux do you agree or am I misunderstanding something?

Yes, I agree

saraycp:
> > @dmarcoux do you agree or am I misunderstanding something?
> 
> Yes, I agree

Alright! Then this is ready for review now.

dmarcoux (approved):

PR #13034, opened 2022-09-02
WIP: Add methods to fetch incoming and outgoing requests

rubhanazeem:
co-authored-by: Dario Leidi <dleidi@suse.de>

closes #4602

PR #13042, opened 2022-09-05
Additional summary info for incoming/outgoing project requests

ncounter:
closes https://github.com/openSUSE/open-build-service/issues/4602

Show insights for `incoming` and `outgoing` project requests.
Arrows and numbers also react with a `title` text explaining which one of the two types are.

Note: the total number could still diff from the `incoming+outgoing` because there are other types of requests included in the total count ([see here](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/project.rb#L1212))

Before:
before-split

After:
after-split


dmarcoux:
This is a duplicate of #13034, right?

ncounter:
> This is a duplicate of #13034, right?

Yes it is, @rubhanazeem initially opened it to share some code with me.

src/api/app/models/project.rb
```diff
@@ -1240,6 +1240,16 @@ def open_requests
     { reviews: reviews, targets: targets, incidents: incidents, maintenance_release: maintenance_release }
   end
 
+  def open_incoming_requests
+    ids = open_requests.values.sum
+    BsRequest.includes(:bs_request_actions).where(id: ids).where(bs_request_actions: { target_project: name })
```
dmarcoux, line 1245: [open_requests](https://github.com/openSUSE/open-build-service/blob/260ba3593885feecb54c8e9b55cd83002982bf86/src/api/app/models/project.rb#L1212) doesn't return request IDs, but request numbers. This line should be changed. Same for the other in `open_outgoing_requests`.

dmarcoux (changes requested):

src/api/app/models/project.rb
ncounter: Good catch, thank you. Fixed.

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -526,6 +526,8 @@ def load_project_info
 
     reqs = @project.open_requests
     @requests = (reqs[:reviews] + reqs[:targets] + reqs[:incidents] + reqs[:maintenance_release]).sort!.uniq
+    @incoming_requests = RequestsFinder.new.incoming(reqs.values.sum, @project.name)
+    @outgoing_requests = RequestsFinder.new.outgoing(reqs.values.sum, @project.name)
```
hennevogel, line 530: You instantiate possibly very man requests and load them into memory in a instance variable....

src/api/app/views/webui/project/side_links/_requests.html.haml
```diff
@@ -10,3 +10,14 @@
   %i.fa.fa-info-circle.text-info
   = link_to(path) do
     = pluralize(requests.size, 'open request')
+
+  %small
+    (
+    %span{ title: 'incoming' }
+      %i.fa.fa-arrow-down
+      #{incoming_requests.size}
```
hennevogel, line 18: ...to call `size` on the array in the view.

src/api/app/views/webui/project/show.html.haml
```diff
@@ -26,7 +26,9 @@
                                                   requests: @requests,
                                                   package: @package,
                                                   linking_projects: @linking_projects,
-                                                  project_maintenance_project: @project_maintenance_project }
+                                                  project_maintenance_project: @project_maintenance_project,
+                                                  incoming_requests: @incoming_requests,
+                                                  outgoing_requests: @outgoing_requests }
```
hennevogel, line 31: ...to pass them around the stack...

hennevogel (changes requested):

hennevogel:
How about you pass around the number instead? 😃

ncounter:
> How about you pass around the number instead? smiley

Very good point, thanks for the heads up. I fixed it in https://github.com/openSUSE/open-build-service/pull/13042/commits/381f8b930e6b235f5ca5ffbfe6c4a3fc5851f4ee

src/api/app/controllers/webui/project_controller.rb
```diff
@@ -526,8 +526,8 @@ def load_project_info
 
     reqs = @project.open_requests
     @requests = (reqs[:reviews] + reqs[:targets] + reqs[:incidents] + reqs[:maintenance_release]).sort!.uniq
-    @incoming_requests = RequestsFinder.new.incoming(reqs.values.sum, @project.name)
-    @outgoing_requests = RequestsFinder.new.outgoing(reqs.values.sum, @project.name)
+    @incoming_requests_size = RequestsFinder.new.incoming(reqs.values.sum, @project.name).size
+    @outgoing_requests_size = RequestsFinder.new.outgoing(reqs.values.sum, @project.name).size
```
hennevogel, line 530: Now you load a lot of requests into memory into an instance variable array to count them. But your database supports counting with [`.count`](https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-count). 

https://medium.com/ruby-daily/comparing-count-length-and-size-in-ruby-on-rails-e061840bc19e

hennevogel (changes requested):

src/api/app/controllers/webui/project_controller.rb
ncounter: I will narrow down the scope of the query to count the db records only and return the number at that level. Yeah it will be much more efficient. The only counter side is that we leave out the option to reuse the query for wider uses (for instance to fetch and render only incoming requests), but I guess we can still write a specific query for that afterwards if and when it will be needed.

rubhanazeem (approved):

src/api/app/queries/requests_finder.rb
```diff
@@ -0,0 +1,13 @@
+class RequestsFinder
```
hennevogel, line 1: Is there a reason you didn't add this to `OpenRequestsFinder` but created your own query object?

ncounter: @hennevogel I'm addressing it with https://github.com/openSUSE/open-build-service/pull/13042/commits/651f9b9215f172d372a267c4d8f3c831dd13626e , what do you think?

src/api/app/queries/open_requests_finder.rb
```diff
@@ -12,4 +12,16 @@ def call
   def requests_finder
     @relation
   end
+
+  def requests_with_actions_by_numbers(request_numbers)
```
hennevogel, line 16:
```suggestion
  def requests_with_actions(request_numbers)
```

src/api/app/queries/open_requests_finder.rb
```diff
@@ -12,4 +12,16 @@ def call
...
+    @relation.includes(:bs_request_actions).where(number: request_numbers)
+  end
+
+  def count_incoming(request_numbers)
+    requests_with_actions_by_numbers(request_numbers).where(bs_request_actions: { target_project: @project_name }).count
```
hennevogel, line 21:
```suggestion
    requests_with_actions(request_numbers).where(bs_request_actions: { target_project: @project_name }).count
```

src/api/app/queries/open_requests_finder.rb
```diff
@@ -12,4 +12,16 @@ def call
...
+  end
+
+  def count_outgoing(request_numbers)
+    requests_with_actions_by_numbers(request_numbers).where(bs_request_actions: { source_project: @project_name }).count
```
hennevogel, line 25:
```suggestion
    requests_with_actions(request_numbers).where(bs_request_actions: { source_project: @project_name }).count
```

hennevogel (changes requested):

hennevogel (approved):

PR #13051, opened 2022-09-07
Simplify timeline and its items

dmarcoux:


src/api/app/assets/stylesheets/webui/notifications.scss
```diff
@@ -1,10 +1,11 @@
 $notifications-filter-box-height: 3.5rem;
+$avatars-counter-size: 1.4375rem;
```
dmarcoux, line 2: We want to reuse this in the timeline CSS instead of having what appears to be a magic number.

src/api/app/assets/stylesheets/webui/timeline.scss
```diff
@@ -1,15 +1,10 @@
 // This stylesheet defines styles for use in the comments and request history merged timeline
-.timeline-item {
-    position: relative;
-}
+.timeline {
+    $timeline-border-width: 0.125rem;
+    border-left: $timeline-border-width solid $gray-400;
 
-.timeline-item:before {
```
dmarcoux, line 6: What this code did is render a 2px-wide gray line next to each timeline item. Instead, we can apply a left border to the timeline as a whole. This is much easier to understand.

src/api/app/views/webui/comment/beta/_comments_thread.html.haml
```diff
@@ -1,3 +1 @@
-.timeline-item
```
dmarcoux, line 1: This would render an extra `timeline-item` div which made the HTML a bit harder to follow since there were timeline items rendered in other timeline items.

We have to keep using a partial here since this is called only in the `webui/comments_controller` with AJAX. See:
- https://github.com/openSUSE/open-build-service/blob/5e8fef2198cd0c1665d04bab6e9de3e8f401740b/src/api/app/controllers/webui/comments_controller.rb#L19
- https://github.com/openSUSE/open-build-service/blob/5e8fef2198cd0c1665d04bab6e9de3e8f401740b/src/api/app/controllers/webui/comments_controller.rb#L58
- https://github.com/openSUSE/open-build-service/blob/5e8fef2198cd0c1665d04bab6e9de3e8f401740b/src/api/app/controllers/webui/comments_controller.rb#L84

The AJAX calls are handled here:
- https://github.com/openSUSE/open-build-service/blob/5e8fef2198cd0c1665d04bab6e9de3e8f401740b/src/api/app/assets/javascripts/webui/comment.js#L43
- https://github.com/openSUSE/open-build-service/blob/5e8fef2198cd0c1665d04bab6e9de3e8f401740b/src/api/app/assets/javascripts/webui/comment.js#L61
- https://github.com/openSUSE/open-build-service/blob/5e8fef2198cd0c1665d04bab6e9de3e8f401740b/src/api/app/assets/javascripts/webui/comment.js#L81

danidoni (approved):

rubhanazeem (approved):

PR #13053, opened 2022-09-07
Set scroll offset for all the pages

ncounter:
This PR adds two CSS rules:
- scroll offset to let anchor scroll the page but do not scroll behind the fixed top nav bar
- let the scroll happen smoothly and visible, instead of jumping to the offset with no animation

scroll-to-anchor

Fixes: https://trello.com/c/m3aezM5E/1953-comment-anchors-dont-account-for-the-height-of-the-top-navbar-m

### DIFF
--- a/src/api/spec/support/shared_examples/features/bootstrap_user_tab.rb
+++ b/src/api/spec/support/shared_examples/features/bootstrap_user_tab.rb
@@ -41,9 +41,8 @@ RSpec.shared_examples 'bootstrap user tab' do
     end
 
     it 'Add non existent user' do
-      add_user_link = page.find('a', text: 'Add User')
-      page.scroll_to(add_user_link)
-      add_user_link.click
+      # FIXME: on mobile, when `screen-offset-top` CSS rule is in place, scrolling and clicking at capybara level does not work
+      page.evaluate_script('$("#add-user").click()')
       sleep 1 # FIXME: Needed to avoid a flickering test because the animation of the modal is sometimes faster than capybara
 
       within('#add-user-role-modal') do
@@ -55,9 +54,8 @@ RSpec.shared_examples 'bootstrap user tab' do
     end
 
     it 'Add an existing user' do
-      add_user_link = page.find('a', text: 'Add User')
-      page.scroll_to(add_user_link)
-      add_user_link.click
+      # FIXME: on mobile, when `screen-offset-top` CSS rule is in place, scrolling and clicking at capybara level does not work
+      page.evaluate_script('$("#add-user").click()')
       sleep 1 # FIXME: Needed to avoid a flickering test because the animation of the modal is sometimes faster than capybara
 
       within('#add-user-role-modal') do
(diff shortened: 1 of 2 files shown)


hennevogel:
This will not work for external links right?

ncounter:
> This will not work for external links right?

Exactly, the condition is looking up only for anchors, that under the hood it means it considers only `<a href="#....">` by the condition `if (href.startsWith('#')) {`

dmarcoux:
We should fix this for the whole application, since this isn't an issue only in the request pages.

ncounter:
> We should fix this for the whole application, since this isn't an issue only in the request pages.

I agree, I just realized I implemented it in `request.js` but it is already _abstract enough_ to be moved and used through all the pages.

dmarcoux:
A much simpler solution would be to use [scroll-padding-top](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding-top). Did you give this a try? You would have to set this on `html` here:

https://github.com/openSUSE/open-build-service/blob/8266deba684eb8b6f23c408c0ce7e7e6bc77be33/src/api/app/assets/stylesheets/webui/application.scss#L75-L77

The height of the fixed navigation bar is defined here:
https://github.com/openSUSE/open-build-service/blob/8266deba684eb8b6f23c408c0ce7e7e6bc77be33/src/api/app/assets/stylesheets/webui/bootstrap_variables/sizes.scss#L2

ncounter:
@hennevogel , @dmarcoux I reverted the `js` implementation as I tried out the CSS rule and it works just fine.


For completeness I leave below useful links about it:
- https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding-top
- https://caniuse.com/?search=scroll-padding-top

dmarcoux (approved):
:shipit:

dmarcoux:
[Comment anchors don't account for the height of the top navbar (M)](https://trello.com/c/m3aezM5E/1953-comment-anchors-dont-account-for-the-height-of-the-top-navbar-m)

rubhanazeem (approved):

rubhanazeem:
Some feature tests are failing. Can you please have a look? I restarted the failing tests but they failed again

src/api/spec/support/shared_examples/features/bootstrap_user_tab.rb
```diff
@@ -41,9 +41,8 @@
     end
 
     it 'Add non existent user' do
-      add_user_link = page.find('a', text: 'Add User')
-      page.scroll_to(add_user_link)
-      add_user_link.click
+      # HACK: on mobile, when `screen-offset-top` CSS rule is in place, scrolling and clicking at capybara level does not work
```
krauselukas, line 44: @ncounter tiny thing, but I would suggest to use the term `FIXME:` instead or in addition to `HACK:`. This way its easier to grep through the code in the future to possibly find things to fix :D

krauselukas (approved):

danidoni (approved):

PR #13071, opened 2022-09-14
Fix link of adding/removing a package in watchlist

eduardoj:
The `@project` variable was not assigned in AJAX requests - it was only assigned in the webui controller. This made not possible to find the package.

Making sure that the `@project` variable is also set for a package in the WatchedItems controller fixes the issue.

How to test the fix, for reviewers:
- Go to a page of a package.
- Click in Watchlist.
- Add/Remove the package from the Watchlist.
- See that the link has changed to Remove/add package from the watchlist.

### DIFF
--- a/src/api/app/controllers/webui/watched_items_controller.rb
+++ b/src/api/app/controllers/webui/watched_items_controller.rb
@@ -28,13 +28,14 @@ class Webui::WatchedItemsController < Webui::WebuiController
   private
 
   def set_watchable
-    @watchable = if params[:project_name] && params[:package_name]
-                   @package = Package.get_by_project_and_name(params[:project_name], params[:package_name])
-                 elsif params[:project_name]
-                   @project = Project.get_by_name(params[:project_name])
-                 elsif params[:number]
-                   @bs_request = BsRequest.find_by(number: params[:number])
-                 end
+    if params[:project_name]
+      @project = Project.get_by_name(params[:project_name])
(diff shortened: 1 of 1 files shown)


src/api/app/controllers/webui/watched_items_controller.rb
```diff
@@ -28,10 +28,9 @@ def toggle_watched_item
   private
 
   def set_watchable
-    @watchable = if params[:project_name] && params[:package_name]
-                   @package = Package.get_by_project_and_name(params[:project_name], params[:package_name])
-                 elsif params[:project_name]
+    @watchable = if params[:project_name]
                    @project = Project.get_by_name(params[:project_name])
+                   params[:package_name] ? @package = Package.get_by_project_and_name(params[:project_name], params[:package_name]) : @project
```
danidoni, line 33: I think this method is a bit convoluted...
There's an `if ... elseif ... end` and a nested ternary, that is a hidden `if ... else`... There's too much going on here...
What about splitting setting `@package` and `@project`, then setting `@watchable`?.

eduardoj: Do you mean like this?

```suggestion
                   @package = Package.get_by_project_and_name(params[:project_name], params[:package_name]) if params[:package_name]
                   @package || @project
```

eduardoj: I changed the code to what we agreed offline, @danidoni. Please take a look.

danidoni (approved):

eduardoj:
codecov hangs. Merging.

PR #13077, opened 2022-09-15
Reset values for delete confirmation modal on toggle watched item

rubhanazeem:
We render `DeleteConfirmationDialogComponent` in the layout with `WatchlistComponent`. When an item is added or removed from watchlist the `WatchlistComponent` is rendered again without taking into consideration the `DeleteConfirmationDialogComponent`. This change reset the values used in the delete confirmation modal


rubhanazeem:
[Fix Ajax issue in Watchlist: the individual delete action doesn't work right after adding an element (M)](https://trello.com/c/rdplRqYZ/1976-fix-ajax-issue-in-watchlist-the-individual-delete-action-doesnt-work-right-after-adding-an-element-m)

eduardoj (approved):

danidoni (approved):

PR #13079, opened 2022-09-15
Split package and project name by the slice delimiter only

ncounter:
When the package or project name is very long and the watchlist bar width is not wide enough, the name breaks and go to the next line at any character as needed. With this PR the name breaks into slices and go to the next line with the entire slice at a meaningful character (for packages the delimiter is `/`, for projects it is `:`)

After:
after
after-html

Fixes: https://trello.com/c/m9bg1tV7


ncounter:
[Line break long project/package links differently (S)](https://trello.com/c/m9bg1tV7/1978-line-break-long-project-package-links-differently-s)

danidoni (approved):

PR #13083, opened 2022-09-15
Remove architectures and repositories filters

saraycp:
After months of tracking the usage of the SCM/CI integration elements, we realized the `architectures` and `repositories` filters have never been used. So we conclude those filters are not helpful for our customers and we decide to get rid of them.

Please read the commit descriptions for more details.

**How to test?**

However you test the workflow, I'd recommend you to test these two scenarios:

1. Adding any of the removed filters to the workflows.yaml. 
The filters part of the YAML could look like this:
```
filters:
    event: pull_request
    repositories:
      only:
        - openSUSE_Tumbleweed
```
As they are unsupported filters now, you need to get an error message like this in the workflow run:
```
<status code="unknown">
  <summary>Filters repositories are unsupported and Documentation for filters: ...
</status>
```
2. Remove `architectures` and `repositories` filters from your workflows.yml. Trigger a workflow. Make sure the workflow runs correctly until the end, it doesn't show any error in the workflow run and it reports back to the SCM correctly.


krauselukas (approved):

saraycp:
Rebased on master

krauselukas:
Codecov fails to report but is green, merging...

PR #13084, opened 2022-09-15
Show confirmation dialog when deleting from the upper link from the watchlist

eduardoj:
The user is already asked for confirmation when deleting an element of the watchlist, clicking on the delete icons. Asking for confirmation was missing, when deleting from the upper link of the watchlist.


krauselukas (approved):

krauselukas:
Codecov fails to report but is green, merging...

PR #13088, opened 2022-09-16
[api] fix access to backend via SSL

adrianschroeter:
* enable SSL when https scheme is specified
 * Allow to disable cert verification in options.yml

belongs to OBS-210

### DIFF
--- a/src/api/app/lib/backend/connection.rb
+++ b/src/api/app/lib/backend/connection.rb
@@ -9,13 +9,26 @@ module Backend
       CONFIG['source_port']
     end
 
+    cattr_accessor :use_ssl, instance_accessor: false do
+      CONFIG['source_protocol'] == 'https'
+    end
+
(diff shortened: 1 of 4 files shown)


src/api/app/lib/backend/connection.rb
```diff
@@ -9,13 +9,18 @@ class Connection
       CONFIG['source_port']
     end
 
+    def initialize
+      @use_ssl = (CONFIG['source_protocol'] == 'https')
+      @verify_mode = (CONFIG['source_protocol_ssl_verification'] == false ? OpenSSL::SSL::VERIFY_NONE : OpenSSL::SSL::VERIFY_PEER)
```
hennevogel, line 14: See above how it's done in this class. Let's stick to one way of doing this...

hennevogel (changes requested):

hennevogel (approved):

PR #13089, opened 2022-09-16
[backend] bs_worker: fix SSL enabling

mlschroe:


src/backend/BSConSign.pm
```diff
@@ -55,21 +55,12 @@ sub createsig {
...
-  my $fn = 'rpmsig-req.bin';
-  my $t = $sd->{'signtime'} || time();
-  my $literal_data_pkt = pack('CCa*N', 0x62, length($fn), $fn, $t).$payload;
-  $literal_data_pkt = BSPGP::pkencodepacket(11, $literal_data_pkt);
-  # create compressed packet
-  my $packets = "$onepass_sig_pkt$literal_data_pkt$sig";
-  my $compressed_pkt;
-  IO::Compress::RawDeflate::rawdeflate(\$packets, \$compressed_pkt);
-  return pack('CC', 0xa3, 1).$compressed_pkt;
+  my $packets = BSPGP::onepass_signed_message($payload, $sig, 'rpmsig-req.bin');
```
eduardoj, line 58:
```suggestion
  my $packets = BSPGP::onepass_signed_message($payload, $sig, 'rpmsig-req.bin', $timestamp);
```
`onepass_signed_message` expects a `$t` parameter:
```perl
sub onepass_signed_message {
  my ($data, $sig, $fn, $t) = @_;
```
Should we also pass here `$timestamp`?

src/backend/BSConSign.pm
```diff
@@ -55,21 +55,12 @@ sub createsig {
...
+  # compress packets like gpg does
+  my $compressed_pkts;
+  IO::Compress::RawDeflate::rawdeflate(\$packets, \$compressed_pkts);
+  $packets = pack('CC', 0xa3, 1).$compressed_pkts;
+  return $packets;
```
eduardoj, line 63:
```suggestion
  return pack('CC', 0xa3, 1).$compressed_pkts;
```
Do we need this assignment to `$packets`?

src/backend/BSConSign.pm
mlschroe: We want/must to reuse the timestamp from the signature.

PR #13102, opened 2022-09-19
Track current object in the watchlist

rubhanazeem:
Before this update, we were trying to generalize watch items. This is not correct because the current object could be different from a watched item. That means on a request page current object is the request. Whereas, the watched items could be a package, project, or request.


rubhanazeem:
[Unnecessary update of the upper link in watchlist](https://trello.com/c/6mDhnFWy/1981-unnecessary-update-of-the-upper-link-in-watchlist)

danidoni (approved):

PR #13108, opened 2022-09-20
Remove trigger_workflow from initializer

saraycp:
That makes the feature not to be listed on the beta features configuration page.

TODO in the reference instance as soon as this is deployed:

- [x] Move all the users to the rollout group.
- [ ] Add group `rollout` to the `trigger_workflow` feature in Flipper.
- [ ] Remove group `trigger_workflow` in the `trigger_workflow` feature in Flipper.

### DIFF
--- a/src/api/config/initializers/flipper.rb
+++ b/src/api/config/initializers/flipper.rb
@@ -1,5 +1,4 @@
 ENABLED_FEATURE_TOGGLES = [
-  { name: :trigger_workflow, description: 'Better SCM and CI integration with OBS workflows' },
   { name: :new_watchlist, description: 'New implementation of watchlist including projects, packages and requests' },
   { name: :request_show_redesign, description: 'Redesign of the request pages to improve the collaboration workflow' }
 ].freeze


saraycp:
Do not merge until we are prepared to announce the roll out of the feature.

rubhanazeem (approved):

dmarcoux:
> That makes the feature not to be listed on the beta features configuration page.

This isn't true, see the controller code here:
https://github.com/openSUSE/open-build-service/blob/6be8b03603e0ef361314e1513102d9504d47286a/src/api/app/controllers/webui/users/beta_features_controller.rb#L40-L48

Such changes aren't needed until a feature toggle is fully removed (so there are no remaining Flipper checks for it).

This doesn't affect production, but our development environment and review apps do not have this feature toggle anymore, although it's needed to test things like the Tokens UI for example.

We don't have to revert this PR since we're going to remove completely the Flipper checks for  `trigger_workflow` in the current deliverable. I also updated the [developer documentation](https://github.com/openSUSE/open-build-service/wiki/Feature-Toggles/_compare/2e938f63427a03bd156d300e53d0743e2aed4ea0...303abe32221ee91f8ef020249a53def9849d2ae5). As for removing the feature toggle from the Flipper initiailizer, this is already documented [here](https://github.com/openSUSE/open-build-service/wiki/Feature-Toggles#delete-a-feature-toggle).

PR #13110, opened 2022-09-20
Add old watched projects to new watched items

eduardoj:
The old watchlist implementation only kept projects. The new watchlist implementation uses other table (`watched_items`) to keep watched items (projects, packages and requests).

Add a data migration to import into the new implementation, the projects which were watched with the old implementation.

Import the projects only if the user doesn't watch any item in the new implementation.

To apply this data migration: `bin/rake data:migrate`

For reviewers:
- Change watched projects accordingly, with and without the beta.
- Run (or rerun) the data migration in the rails console with: `VERSION=2022092011532 bin/rake data:migrate:redo`

### DIFF
--- a/src/api/db/data/20220920115342_copy_old_watched_projects.rb
+++ b/src/api/db/data/20220920115342_copy_old_watched_projects.rb
@@ -1,0 +1,9 @@
+class CopyOldWatchedProjects < ActiveRecord::Migration[6.1]
+  def up
+    WatchedProject.all.each do |old_watched_project|
+      next if old_watched_project.user.watched_items.count.positive?
+
+      WatchedItem.find_or_create_by(watchable: old_watched_project.project, user: old_watched_project.user)
+    end
+  end
+end
--- a/src/api/db/data_schema.rb
+++ b/src/api/db/data_schema.rb
@@ -1,1 +1,1 @@
-DataMigrate::Data.define(version: 20220318123314)
+DataMigrate::Data.define(version: 20220920115342)


src/api/db/data/20220920115342_copy_old_watched_projects.rb
```diff
@@ -0,0 +1,7 @@
+class CopyOldWatchedProjects < ActiveRecord::Migration[6.1]
+  def up
+    WatchedProject.all.each do |old_watched_project|
+      WatchedItem.find_or_create_by(watchable: old_watched_project.project, user: old_watched_project.user)
```
hennevogel, line 4:
```suggestion
    WatchedProject.all.each do |old_watched_project|
      next if old_watched_project.user.watched_items.count.positive?
      
      WatchedItem.find_or_create_by(watchable: old_watched_project.project, user: old_watched_project.user)
```

hennevogel: I wouldn't mess with peoples *new* watchlist

hennevogel (changes requested):

src/api/db/data/20220920115342_copy_old_watched_projects.rb
eduardoj: I see. Suggestion applied. I also added the following sentence to the commit description: _"Import the projects only if the user doesn't watch any item in the new implementation."_ .

Please take another look.

hennevogel (approved):

src/api/db/data/20220920115342_copy_old_watched_projects.rb
saraycp: @hennevogel @eduardoj  :thinking: I'm not convinced of this decision...

We know some users made the effort to add all the projects they had in the old watchlist to the new one.

But some other users could have tried out the new watchlist by adding random projects/packages/requests and then they disabled the feature flag to continue using the possibly long list of projects they had in the old watchlist. 
In that case, I wouldn't like them to lose the list they had.

hennevogel: > But some other users could have tried out the new watchlist by adding random projects/packages/requests and then they disabled the feature flag to continue using the possibly long list of projects they had in the old watchlist.
In that case, I wouldn't like them to lose the list they had.

Then you need to make this more complicated and add a feature to do the thing in between...

eduardoj:
#13126 offers a better approach. Closing.

PR #13144, opened 2022-09-28
Move the backend proxy to a concern

hennevogel:
and document it a bit...

No code changes, just moving around and adding some comments.


hennevogel:
At some point we should start including this in the controllers that use `pass_to_backend` instead of including it generally...

krauselukas (approved):

PR #13149, opened 2022-09-29
Include `cacheonly` and `nodiff` parameters in package diff

eduardoj:
Calling the API endpoint `/source/:project_name/:package_name?cmd=diff` with `cacheonly` and/or `nodiff` parameters didn't make a difference when retrieving the results from the backend. Both parameters should have been included in the list of parameters accepted by the endpoint above mentioned as they were introduced. Please have a look at the descriptions of the commits for details.


danidoni (approved):

eduardoj:
Codecov fails to report, merging...

PR #13151, opened 2022-09-29
OpenAPI specifications for get source

rubhanazeem:
Document OpenAPI specifications for `/source` to get the list of projects or the list of deleted projects

### DIFF
--- a/src/api/public/apidocs-new/OBS-v2.10.50.yaml
+++ b/src/api/public/apidocs-new/OBS-v2.10.50.yaml
@@ -255,7 +255,8 @@ paths:
     $ref: 'paths/source_orderkiwirepos.yaml'
 
   # Sources - Projects
-
+  /source:
+    $ref: 'paths/source.yaml'
   /source/{project_name}?cmd=addchannels:
     $ref: 'paths/source_project_name_cmd_addchannels.yaml'
   /source/{project_name}?cmd=copy:
--- a/src/api/public/apidocs-new/paths/source.yaml
+++ b/src/api/public/apidocs-new/paths/source.yaml
@@ -1,0 +1,53 @@
+get:
+  summary: List all the projects
+  description: |
+    Get a list of projects
+  parameters:
+    - name: deleted
+      in: query
+      schema:
+        type: integer
+      description: Set to 1 to include deleted projects in the listing. This is available for Admins only.
+      example: 0
+      required: false
+
+  responses:
+    '200':
+      description: OK. The request has succeeded.
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/source/projects.yaml'
+          examples:
+            existing_projects:
+              entry:
+                name: "home:Admin"
+            deleted_projects:
+              entry:
+                name: "home:Admin"
+
+    '400':
+      description: Wrong value for deleted parameter.
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/api_response.yaml'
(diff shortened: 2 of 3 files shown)


rubhanazeem:
[Sources - Projects (L)](https://trello.com/c/MSXY9t4w/1986-sources-projects-l)

src/api/public/apidocs-new/OBS-v2.10.50.yaml
```diff
@@ -255,7 +255,8 @@ paths:
     $ref: 'paths/source_orderkiwirepos.yaml'
 
   # Sources - Projects
-
+  /source:
+    $ref: 'paths/source_projects.yaml'
```
eduardoj, line 259:
```suggestion
    $ref: 'paths/source.yaml'
```
Let's keep the convention that every path file has the name of its path and arguments.

src/api/public/apidocs-new/paths/source_projects.yaml
```diff
@@ -0,0 +1,44 @@
...
+            type: object
+            properties:
+              code:
+                type: string
+                xml:
+                  attribute: true
+              summary:
+                type: string
+            xml:
+              name: status
```
eduardoj, line 35:
```suggestion
            $ref: '../schemas/api_response.yaml'
```
It should be possible to use the `api_response.yaml` schema here.

rubhanazeem: `api_response.yaml` has more information than we need here. We just need code and summary

eduardoj: The schema defines all possible fields and how they are arranged. But it doesn't mean that we have to use all the fields, in all the examples.

If in this case the example uses ony `code` and `summary`, there is no need to define another schema inline. When we don't assign values to the other fields, those fields are simply not shown.

rubhanazeem: I see. Okay, I've updated it

src/api/public/apidocs-new/paths/source.yaml
```diff
@@ -0,0 +1,43 @@
+get:
+  summary: List all the projects
+  description: |
+    Get a list of projects or a list of deleted projects.
```
hennevogel, line 4:
```suggestion
    Get a list of projects
```

src/api/public/apidocs-new/paths/source.yaml
```diff
@@ -0,0 +1,43 @@
...
+  parameters:
+    - name: deleted
+      in: query
+      schema:
+        type: integer
+      description: 0 for existing projects and 1 for deleted projects.
```
hennevogel, line 10:
```suggestion
      description: Set to 1 to include deleted projects in the listing. This is available for Admins only.
```

src/api/public/apidocs-new/paths/source.yaml
```diff
@@ -0,0 +1,43 @@
...
+      example: 0
+      required: false
+
+  responses:
```
hennevogel, line 14: There is one more

```
osc api /source/\?deleted=0 
Server returned an error: HTTP Error 403: Forbidden
only admins can see deleted projects
```

hennevogel (changes requested):

src/api/public/apidocs-new/paths/source.yaml
rubhanazeem: I've documented it

hennevogel (approved):

PR #13156, opened 2022-09-29
Move Package.download_url_for_binary back to PackageController

### LINKED ISSUE #13147: Bogus error from https://build.opensuse.org/package/binaries/openSUSE:Factory:ARM/aaa_base/standard
https://build.opensuse.org/package/binaries/openSUSE:Factory:ARM/aaa_base/standard
results in 
404 repository 'openSUSE:Factory/standard' has no architecture 'aarch64'

hennevogel:
We need to account for packages that come through project links. The only reasonable place you can know about this is the controller. An instantiated `Package` doesn't know which `Project` it was instantiated with in `Package.get_by_project_and_name`...

Fixes #13147

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -23,7 +23,7 @@ class Webui::PackageController < Webui::WebuiController
 
   before_action :validate_xml, only: [:save_meta]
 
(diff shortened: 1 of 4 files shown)


hennevogel:
Depends on #13178

danidoni (approved):

PR #13166, opened 2022-10-04
OpenAPI specifications to get and update src project meta

rubhanazeem:


rubhanazeem:
[Sources - Projects (L)](https://trello.com/c/MSXY9t4w/1986-sources-projects-l)

src/api/public/apidocs-new/paths/source_project_name_meta.yaml
```diff
@@ -33,3 +33,50 @@ get:
...
+  summary: Write project meta file
+  description: |
+    Write project's meta file. Create the project if it doesn't exist.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - in: query
+      name: comment
+      required: false
```
eduardoj, line 47:
```suggestion
```
It the parameter is not required, there is no need to set `required` to `false` and this line could be removed.

src/api/public/apidocs-new/paths/source_project_name_meta.yaml
```diff
@@ -33,3 +33,50 @@ get:
...
+              - aarch64
+  responses:
+    '200':
+      $ref: '../components/responses/succeeded.yaml'
+    '400':
+      $ref: '../components/responses/bad_request_error.yaml'
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      $ref: '../components/responses/unknown_project.yaml'
```
eduardoj, line 80:
```suggestion
```
I couldn't reproduce this response.

If the path parameter `project_name` is different to the `name` element passed in the `<project>` element of the body xml, I receive a 400 error:
```xml
<status code="project_name_mismatch">
  <summary>project name in xml data does not match resource path component</summary>
</status>
```

If they are both the same and the project doesn't exist, the project gets created. It's already in the description.

src/api/public/apidocs-new/paths/source_project_name_meta.yaml
```diff
@@ -33,3 +33,50 @@ get:
...
+            role: maintainer
+          repository:
+            name: openSUSE_Tumbleweed
+            path:
+              project: openSUSE.org:openSUSE:Factory
+              repository: snapshot
+            arch:
+              - x86_64
+              - aarch64
+  responses:
```
saraycp, line 72: I miss a few 403 responses here. I don't think we need to list all of them but at least one...

when the project is locked (see [this line](https://github.com/openSUSE/open-build-service/blob/7e5cf6630d2e5202ca5dd4796c66010560b8b234/src/api/app/controllers/source_project_meta_controller.rb#L46)), the response is
```
<status code="change_project_no_permission">
  <summary>The project home:Admin is locked</summary>
</status>
```

when the user doesn't have permission to modify the project (see [this line](https://github.com/openSUSE/open-build-service/blob/7e5cf6630d2e5202ca5dd4796c66010560b8b234/src/api/app/controllers/source_project_meta_controller.rb#L49)), the response is:
```
<status code="change_project_no_permission">
  <summary>no permission to change project</summary>
</status>
```

when the user is not allowed to create a new project (see [this line](https://github.com/openSUSE/open-build-service/blob/7e5cf6630d2e5202ca5dd4796c66010560b8b234/src/api/app/controllers/source_project_meta_controller.rb#L55)), the response is:
```
<status code="create_project_no_permission">
  <summary>no permission to create project home:yyy</summary>
</status>
```

There are more responses like these ones as a result of calling the methods `ensure_access_to_edit_remote_project`, `ensure_xml_attributes_are_valid`, `remove_repositories!`
starting in [line 60](https://github.com/openSUSE/open-build-service/blob/7e5cf6630d2e5202ca5dd4796c66010560b8b234/src/api/app/controllers/source_project_meta_controller.rb#L60).

src/api/public/apidocs-new/paths/source_project_name_project_file.yaml
```diff
@@ -0,0 +1,60 @@
...
+            not boolean:
+              value:
+                code: 400
+                summary: "not boolean"
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      $ref: '../components/responses/unknown_project.yaml'
+  tags:
+    - Sources - Projects
```
saraycp, line 60: I know this endpoint was listed in the card about `Sources - Projects` but I think it should belong to `Sources - Packages` as `/source/{project_name}/_project/{file_name}` is, in fact, the route to a file inside a package named `_project`

src/api/public/apidocs-new/paths/source_project_name_project_file.yaml
```diff
@@ -0,0 +1,60 @@
...
+        type: integer
+      description: Set to 1 to include _meta files
+      example: 1
+    - in: query
+      name: rev
+      required: false
+      schema:
+        type: integer
+      description: Revision number
+  responses:
```
saraycp, line 23: I also miss an example of 403 error here. At least, there is one raised by the exception `Package::ReadSourceAccessError` when calling `validate_read_access_of_deleted_package` in [this line](https://github.com/openSUSE/open-build-service/blob/7e5cf6630d2e5202ca5dd4796c66010560b8b234/src/api/app/controllers/source_controller.rb#L295).

src/api/public/apidocs-new/paths/source_project_name_project_file.yaml
saraycp: Let's postpone the conversation about this. It's just a matter of which list we display the endpoint in. Something we can decide later when all the `source` endpoints are implemented.

src/api/public/apidocs-new/paths/source_project_name_meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+            repository:
+              name: openSUSE_Tumbleweed
+              path:
+                project: openSUSE.org:openSUSE:Factory
+                repository: snapshot
+              arch:
+                - x86_64
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
```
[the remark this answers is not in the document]
rubhanazeem: I've documented this response

src/api/public/apidocs-new/paths/source_project_name_meta.yaml
rubhanazeem: Done. I've documented this response

src/api/public/apidocs-new/components/responses/change_project_no_permission.yaml
```diff
@@ -0,0 +1,15 @@
+description: |
+  Change project no permission.
```
saraycp, line 2:
```suggestion
  The user does not have permission to change the project.
```

src/api/public/apidocs-new/paths/source_project_name_project_file.yaml
```diff
@@ -0,0 +1,62 @@
...
+            $ref: '../components/schemas/api_response.yaml'
+          examples:
+            not boolean:
+              value:
+                code: 400
+                summary: "not boolean"
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '403':
+      $ref: '../components/responses/change_project_no_permission.yaml'
```
eduardoj, line 58:
```suggestion
```
This is the GET request. I guess these change project permissions errors are only returned with the PUT endpoint of the next commit.

saraycp: This was discussed offline. The GET request can end up raising [this exception](https://github.com/openSUSE/open-build-service/blob/71355f1d6161139b445a28db1fb799e5b1e88880/src/api/app/models/package/errors.rb#L30). As they aren't exactly the same, it's better not to extract them to the same file.

Each request declares its own response.

eduardoj (approved):

PR #13169, opened 2022-10-04
Add the build results badge for the package

### LINKED ISSUE #2908: Badge for github integration

hellcp-work:
This adds integration badges for packages

To verify this feature visit `/package/badge/:project/:package.svg` (with optional `percent` and `ratio` type `parameters`)

/package/badge/:project/:package.svg

/package/badge/:project/:package.svg?type=percent

/package/badge/:project/:package.svg?type=ratio

I changed it so that instead of requiring the package, the font is bundled under `app/assets/fonts`.

Fixes #2908 and supersedes #10820

### DIFF
--- a/src/api/config/routes/webui_routes.rb
+++ b/src/api/config/routes/webui_routes.rb
@@ -271,6 +271,7 @@ OBSApi::Application.routes.draw do
         resources :submissions, controller: 'webui/requests/submissions', only: [:new, :create], constraints: cons
         resource :files, controller: 'webui/packages/files', only: [:new, :create], constraints: cons
         put 'toggle_watched_item', controller: 'webui/watched_items', constraints: cons
+        resource :badge, controller: 'webui/packages/badge', only: [:show], constraints: ->(request) { request.format == :svg }
       end
 
       resources :role_additions, controller: 'webui/requests/role_additions', only: [:new, :create], constraints: cons
--- a/src/api/spec/features/webui/badge_spec.rb
+++ b/src/api/spec/features/webui/badge_spec.rb
@@ -1,0 +1,85 @@
+require 'browser_helper'
+
+RSpec.describe 'Badge', type: :feature do
+  let(:user) { create(:confirmed_user, :with_home, login: 'tom') }
+  let(:source_project) { user.home_project }
+  let(:source_package) { create(:package, name: 'my_package', project: source_project) }
+
+  context 'without build results' do
+    before do
+      path = "#{CONFIG['source_url']}/build/#{user.home_project}/_result?locallink=1&multibuild=1&lastbuild=1" \
+             "&package=#{source_package}&view=status"
+      stub_request(:get, path).and_return(body:
+        %(<resultlist state="eb0459ee3b000176bb3944a67b7c44fa">
+          </resultlist>))
+    end
+
+    it 'displays the correct unknown state despite percent type' do
+      visit project_package_badge_path(source_project.name, source_package.name, format: 'svg', type: 'percent')
+      xml = Capybara.string(page.body)
+      expect(xml).to have_text('unknown')
+    end
+
+    it 'displays the correct unknown state' do
+      visit project_package_badge_path(source_project.name, source_package.name, format: 'svg')
+      xml = Capybara.string(page.body)
+      expect(xml).to have_text('unknown')
+    end
+  end
+
+  context 'with failing and succeeding build results' do
+    before do
+      path = "#{CONFIG['source_url']}/build/#{user.home_project}/_result?locallink=1&multibuild=1&lastbuild=1" \
+             "&package=#{source_package}&view=status"
+      stub_request(:get, path).and_return(body:
+        %(<resultlist state="eb0459ee3b000176bb3944a67b7c44fa">
+            <result project="home:tom" repository="openSUSE_Leap_42.1" arch="x86_64" code="succeeded" state="building">
+              <status package="my_package" code="succeeded" />
+            </result>
+            <result project="home:tom" repository="images" arch="x86_64" code="failed" state="building">
+              <status package="my_package" code="failed" />
+            </result>
+          </resultlist>))
+    end
+
+    it 'displays the correct percentage' do
+      visit project_package_badge_path(source_project.name, source_package.name, format: 'svg', type: 'percent')
+      xml = Capybara.string(page.body)
+      expect(xml).to have_text('50%')
+    end
+
+    it 'displays the correct failed state' do
+      visit project_package_badge_path(source_project.name, source_package.name, format: 'svg')
+      xml = Capybara.string(page.body)
+      expect(xml).to have_text('failed')
+    end
+  end
+
+  context 'with succeeding build results' do
+    before do
+      path = "#{CONFIG['source_url']}/build/#{user.home_project}/_result?locallink=1&multibuild=1&lastbuild=1" \
+             "&package=#{source_package}&view=status"
+      stub_request(:get, path).and_return(body:
+        %(<resultlist state="eb0459ee3b000176bb3944a67b7c44fa">
+            <result project="home:tom" repository="openSUSE_Leap_42.1" arch="x86_64" code="succeeded" state="building">
+              <status package="my_package" code="succeeded" />
+            </result>
+            <result project="home:tom" repository="images" arch="x86_64" code="succeeded" state="building">
+              <status package="my_package" code="succeeded" />
+            </result>
+          </resultlist>))
+    end
+
+    it 'displays the correct percentage' do
+      visit project_package_badge_path(source_project.name, source_package.name, format: 'svg', type: 'percent')
+      xml = Capybara.string(page.body)
+      expect(xml).to have_text('100%')
+    end
+
+    it 'displays the suceeded state' do
+      visit project_package_badge_path(source_project.name, source_package.name, format: 'svg')
+      xml = Capybara.string(page.body)
+      expect(xml).to have_text('succeeded')
+    end
+  end
+end
--- a/src/api/app/assets/images/badge-failed.svg
+++ b/src/api/app/assets/images/badge-failed.svg
@@ -1,0 +1,30 @@
+<svg aria-label="build service: failed" height="20" role="img" version="1.1" viewBox="0 0 139 20" width="139" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<title>build service: failed</title>
+<linearGradient id="s" x2="0" y2="100%">
+<stop offset="0" stop-color="#bbb" stop-opacity=".1"/>
+<stop offset="1" stop-opacity=".1"/>
+</linearGradient>
+<clipPath id="r">
+<rect fill="#fff" height="20" rx="3" width="139"/>
+</clipPath>
+<g clip-path="url(#r)">
+<rect fill="#555" height="20" width="98" x="0"/>
+<rect fill="#e05d44" height="20" width="41" x="98"/>
+<rect fill="url(#s)" height="20" width="139"/>
+</g>
+<g font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-anchor="middle" text-rendering="geometricPrecision">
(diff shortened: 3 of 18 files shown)


danidoni:
> Looking at this, it looks like a flickering test, is this a known issue?
> 
> ```
>   1) Subscriptions update group notifications settings disable a group's notifications for email and web channels
>      Failure/Error: expect(find_field("groups_#{group}_email", visible: false)).not_to be_checked
>        expected `#<Capybara::Node::Element tag="input" path="/HTML/BODY[1]/DIV[1]/DIV[3]/DIV[1]/DIV[3]/DIV[2]/DIV[1]/DIV[2]/DIV[1]/DIV[1]/DIV[1]/DIV[1]/DIV[1]/INPUT[1]">.checked?` to be falsey, got true
>      # ./spec/features/webui/subscriptions_spec.rb:58:in `block (3 levels) in <top (required)>'
> ```

Retriggered the CircleCI failure and it's gone. So yeah, a flickering test...

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,139 @@
...
+    #        [:gray, :green, :yellow, :orange, :blue, :red]
+    info = { text: 'build service', color: color(:gray), font_size: 11, icon_size: 14 }
+    icon = Rails.application.assets['obs-logo_no-text.svg']
+    info[:icon] = "data:#{icon.content_type};base64,#{Base64.strict_encode64(icon.source)}"
+    info[:text_width] = (text_width('build service') * info[:font_size]).ceil
+    @badge << info
+    project = Project.find_by_name(params[:project])
+    package = Package.find_by_project_and_name(params[:project], params[:package])
+    # Return if the following part will fail
+    return badge_processing unless package.buildresult(project).results[package.name]
```
danidoni, line 28: Looks like you are trying to return something coming from `badge_processing` but you actually aren't, right? I find it confusing. I suggest to split this into the condition and the badge processing.

In fact, the `badge_processing` happens in both cases of the condition, so I would do the badge processing before, and branching later

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,139 @@
...
+  end
+
+  private
+
+  def color(value)
+    case value
+    when :green then '#97ca00'
+    when :yellow then '#dfb317'
+    when :orange then '#fe7d37'
+    when :blue then '#007ec6'
```
danidoni, line 47: This can be replaced by constants:
```
GREEN = '#97ca00'
YELLOW = '#dfb317'
```

Also, this looks like view code, not controller code...

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,139 @@
+class Webui::BadgeController < Webui::WebuiController
+  before_action :set_project, only: [:show]
+  before_action :require_package, only: [:show]
+  before_action :require_login, except: [:show]
+
+  def show
+    # Badge height
+    @height = 20
```
danidoni, line 8: This looks like could be a constant

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,139 @@
...
+    when :red then '#e05d44'
+    else '#555'
+    end
+  end
+
+  # This is always run before the badge is shown, to calculate all the metrics
+  # and set the correct title
+  def badge_processing
+    @title = @badge.pluck(:text).join(': ')
+    @width = calculate_positions
```
danidoni, line 57: `calculate_positions` is not used anywhere else so I would just merge `badge_processing` and `calculate_positions`.

danidoni (changes requested):

src/api/spec/controllers/webui/badge_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'webmock/rspec'
+require 'rails_helper'
+RSpec.describe Webui::BadgeController, vcr: true do
+  let(:user) { create(:confirmed_user, :with_home, login: 'tom') }
+  let(:source_project) { user.home_project }
+  let(:source_package) { create(:package, name: 'my_package', project: source_project) }
+
+  describe 'GET #show' do
+    context 'require_package before_action' do
```
danidoni, line 9: What are you trying to model in this context?

src/api/app/views/webui/badge/show.svg.haml
```diff
@@ -0,0 +1,39 @@
...
+  %g{ 'clip-path': 'url(#r)' }
+    - @badge.each do |item|
+      :ruby
+        color = case item[:color]
+                when :green then '#97ca00'
+                when :yellow then '#dfb317'
+                when :orange then '#fe7d37'
+                when :blue then '#007ec6'
+                when :red then '#e05d44'
+                else '#555'
```
danidoni, line 24: If you extract the `Badge` object, you can push this there.

src/api/app/views/webui/badge/show.svg.haml
```diff
@@ -0,0 +1,39 @@
+%svg{ xmlns: 'http://www.w3.org/2000/svg',
+  width: @width,
+  height: Webui::BadgeController::BADGE_HEIGHT,
```
danidoni, line 3: This should be also pushed to the `Badge` object

danidoni (dismissed):

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,149 @@
+class Webui::BadgeController < Webui::WebuiController
+  before_action :set_project, only: [:show]
+  before_action :require_package, only: [:show]
+  before_action :require_login, except: [:show]
+
+  class Badge
```
danidoni, line 6: Can you move the class to `app/models/` directory, please? The convention is to have the objects there.

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Webui::BadgeController < Webui::WebuiController
+  before_action :set_project, only: [:show]
+  before_action :require_package, only: [:show]
+  before_action :require_login, except: [:show]
+
+  def show
+    @badge = Badge.new
+    @badge.add_segment({ text: 'build service', icon_asset: Rails.application.assets['obs-logo_no-text.svg'] })
+
+    # Return if the following part will fail
```
danidoni, line 10:
```suggestion
```
This comment is not needed anymore

src/api/config/routes/webui_routes.rb
```diff
@@ -389,6 +389,9 @@
     ### /apidocs
     get 'apidocs', to: redirect('/apidocs/index')
     get 'apidocs/(index)' => 'webui/apidocs#index', as: 'apidocs_index'
+    controller 'webui/badge' do
+      get 'badge/:project/:package.svg' => :show, constraints: cons, defaults: { format: 'svg' }
+    end
```
eduardoj, line 394:
```suggestion

    resource :badge, controller: 'webui/badge', only: [], constraints: cons do
      get ':project/:package.:format' => :show, constraints: ->(request) { request.format == :svg }
    end
```
I would change from using a `controller` block to a `resource` block, like suggested. We use `resource` or `resources` whenever is possible. What do you think?

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Webui::BadgeController < Webui::WebuiController
+  before_action :set_project, only: [:show]
+  before_action :require_package, only: [:show]
```
eduardoj, line 3:
```suggestion
  before_action :set_project
  before_action :require_package
```
There is no need to define `only: [:show]` when there is only one action in the controller.

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  before_action :require_login, except: [:show]
```
eduardoj, line 4:
```suggestion
```
By default `require_login` is not needed. We can drop this line.

src/api/config/routes/webui_routes.rb
```diff
@@ -389,8 +389,9 @@
     ### /apidocs
     get 'apidocs', to: redirect('/apidocs/index')
     get 'apidocs/(index)' => 'webui/apidocs#index', as: 'apidocs_index'
-    controller 'webui/badge' do
-      get 'badge/:project/:package.svg' => :show, constraints: cons, defaults: { format: 'svg' }
+
+    resource :badge, controller: 'webui/badge', only: [], constraints: cons do
+      get ':project/:package.:format' => :show, constraints: ->(request) { request.format == :svg }
```
danidoni, line 394: Nice!

src/api/app/models/badge.rb
```diff
@@ -0,0 +1,136 @@
...
+  def initialize
+    @segments = []
+    @width = 0
+    @title = 'Badge'
+    @height = BADGE_HEIGHT
+  end
+
+  def add_segment(segment = {})
+    # The possible fields inside of a badge hash are:
+    # icon: it's provided as Rails asset, to make it easier to get content type and hash
```
hennevogel, line 17: Why is this configurable if it ever will be just this one icon? :)

src/api/app/models/badge.rb
```diff
@@ -0,0 +1,136 @@
...
+    # text: the text that will be shown in that segment
+    # text_width: text_width function is commonly used here, but we can provide other
```
hennevogel, line 19: Same here, we don't do that so why do we have code to do that?

hennevogel (commented):
In general there is too much code for my taste for this *single* need. Supporting different "segments" and calculating everything based on this etc. 

KISS and worry bit less about the future, more about the now.

src/api/spec/models/badge_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'rails_helper'
+require 'webmock/rspec'
+
+RSpec.describe Badge, vcr: true do
```
eduardoj, line 4:
```suggestion

RSpec.describe Badge, type: :model do
```
No need to require `webmock/rspec` or use `vcr`. It's a good practice to set the type of the spec, in this case to `:model`.

src/api/spec/controllers/webui/badge_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'webmock/rspec'
+require 'rails_helper'
+RSpec.describe Webui::BadgeController, vcr: true do
```
eduardoj, line 3:
```suggestion

RSpec.describe Webui::BadgeController, type: :controller do
```
No need to require `webmock/rspec` or use `vcr`. It's a good practice to set the type of the spec, in this case to :controller.

eduardoj: Sorry, the comment was right, the suggestion was wrong. It should have removed the `require 'webmock/rspec'`. You fixed it. Thanks! :+1:

eduardoj (approved):
`status_segment` could be refactored later, if needed. Nice work!

src/api/app/models/badge.rb
```diff
@@ -0,0 +1,134 @@
+class Badge
```
hennevogel, line 1: I'm still not convinced that we need to maintain all this code for a relatively fixed amount of badges. There are only 12 possible build results for a package. There are only 2 possible build result states per project. Sounds like completely static badges for packages to me. And for projects it sounds like a badge template with some text replacement.

hennevogel (changes requested):

danidoni:
Some good commit-squashing is needed here...

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+  def show
+    results = @package.buildresult(@project, false, true).results[@package.name]
+    render inline: create_badge(results)
+  end
+
+  private
+
+  def create_badge(results)
+    filename = 'badge-unknown.svg'
+    return Rails.application.assets[filename].source if results.empty?
```
eduardoj, line 14:
```suggestion
    return Rails.application.assets[filename].source if results.blank?
```
Where there aren't any build results, `results` is `nil`.

src/api/spec/features/webui/badge_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  context 'without build results' do
+    before do
+      path = "#{CONFIG['source_url']}/build/#{user.home_project}/_result?locallink=1&multibuild=1&lastbuild=1" \
+             "&package=#{source_package}&view=status"
+      stub_request(:get, path).and_return(body:
+        %(<resultlist state="eb0459ee3b000176bb3944a67b7c44fa">
+          </resultlist>))
+    end
+
+    it 'displays the correct percentage' do
```
eduardoj, line 17:
```suggestion
    it 'displays the correct unknown state' do
```

src/api/spec/features/webui/badge_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      visit show_badge_path(source_project, source_package, format: 'svg', type: 'percent')
+      xml = Capybara.string(page.body)
+      expect(xml).to have_text('unknown')
+    end
+
+    it 'displays the correct failed state' do
```
eduardoj, line 23:
```suggestion
    it 'displays the correct unknown state' do
```

src/api/spec/features/webui/badge_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+          </resultlist>))
+    end
+
+    it 'displays the correct percentage' do
+      visit show_badge_path(source_project, source_package, format: 'svg', type: 'percent')
+      xml = Capybara.string(page.body)
+      expect(xml).to have_text('100%')
+    end
+
+    it 'displays the correct failed state' do
```
eduardoj, line 79:
```suggestion
    it 'displays the succeeded state' do
```

src/api/app/controllers/webui/badge_controller.rb
eduardoj: Without any build results, leaving the code with `empty?`, it errors with:
```
Completed 500 Internal Server Error in 167ms (ActiveRecord: 9.7ms | Backend: 7.5ms | Allocations: 110903)
NoMethodError (undefined method `empty?' for nil:NilClass

    return Rails.application.assets[filename].source if results.empty?
                                                               ^^^^^^^):

app/controllers/webui/badge_controller.rb:14:in `create_badge'

```

eduardoj (dismissed):

hennevogel:
Now what I'm missing is maybe beyond this PR..

- how are people supposed to know about this route?
- how are people supposed to know what they need to paste into their README.md?
- how can I limit this to a specific repository or arch?

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+    filename = 'badge-succeeded.svg' if results.all? { |r| r.code == 'succeeded' }
+    filename = 'badge-percent.svg' if params[:type] == 'percent'
+    prepare_badge_file(filename, results)
+  end
+
+  def prepare_badge_file(filename, results)
+    file = Rails.application.assets[filename].source
+    succeeded = results.select { |r| r.code == 'succeeded' }.try(:length).to_i
+    text = "#{100 * succeeded / results.length}%"
+    file.gsub!('%PERCENTAGE%', text)
```
dmarcoux, line 26: Combining the `create_badge` and `prepare_badge_file` methods would simplify the code since some code is only needed if a specific badge is needed, like this line.

dmarcoux (changes requested):

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,31 @@
+class Webui::BadgeController < Webui::WebuiController
+  before_action :set_project
+  before_action :require_package
+
+  def show
+    results = @package.buildresult(@project, false, true).results[@package.name]
```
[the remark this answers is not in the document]
dmarcoux: We also fetch the build results whenever the package#show page is rendered and there, we don't even cache them. That page has a much bigger impact than what I expect from this build results badge feature, so I guess we're good without cache here. It's not ideal, but we can live with it.

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,39 @@
+class Webui::BadgeController < Webui::WebuiController
+  before_action :set_project
+  before_action :require_package
+
+  def show
+    results = buildresults
+    results = results.select { |r| r.architecture == params[:arch] } if params[:arch]
```
hennevogel, line 7:
```suggestion
    results = results.select { |r| r.architecture == params[:arch] } if params[:architecture]
```

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    results = results.select { |r| r.repository == params[:repo] } if params[:repo]
```
hennevogel, line 8:
```suggestion
    results = results.select { |r| r.repository == params[:repo] } if params[:repository]
```

src/api/app/controllers/webui/badge_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    send_data(create_badge(results), type: 'image/svg+xml', disposition: 'inline')
+  end
+
+  private
+
+  def buildresults
+    Rails.cache.fetch(['buildresults', @project.name, @package.name], expires_in: 10.minutes) do
```
hennevogel, line 15: I'm not sure a time based cache invalidation is worth it. We allow fetching build results anonymously from `Webui::PackageController#show`, `Webui::RequestController#show` and `Webui::ProjectController#monitor` and possibly more places. An event based  cache invalidation would be nice, but you don't have to do this now.

hennevogel (changes requested):

hennevogel (approved):

PR #13170, opened 2022-10-04
OpenAPI specifications to list patterns of a project

rubhanazeem:
API specifications for `GET 'source/:project/_pattern'`

### DIFF
--- a/src/api/public/apidocs-new/components/schemas/source/project_pattern.yaml
+++ b/src/api/public/apidocs-new/components/schemas/source/project_pattern.yaml
@@ -1,0 +1,41 @@
+type: object
+properties:
+  name:
+    type: string
+    xml:
+      attribute: true
+  rev:
+    type: string
+    xml:
+      attribute: true
+  vrev:
+    type: string
+    xml:
+      attribute: true
+  srcmd5:
+    type: string
+    xml:
+      attribute: true
+  entry:
+    type: array
+    items:
+      type: object
+      properties:
+        name:
+          type: string
+          xml:
+            attribute: true
+        md5:
+          type: string
+          xml:
+            attribute: true
+        size:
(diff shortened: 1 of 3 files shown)


src/api/public/apidocs-new/components/schemas/source/project_pattern.yaml
```diff
@@ -0,0 +1,39 @@
...
+    type: string
+    xml:
+      attribute: true
+  srcmd5:
+    type: string
+    xml:
+      attribute: true
+  entry:
+    type: object
+    properties:
```
saraycp, line 21:
```suggestion
  entry:
    type: array
    items:
      type: object
      properties:   
```

There can be more than one entry in the directory. This is an example of `_pattern` with more than one file:  https://build.opensuse.org/package/show/Cloud:OpenStack:Victoria/_pattern

src/api/public/apidocs-new/paths/source_project_name_pattern.yaml
```diff
@@ -0,0 +1,31 @@
+get:
+  summary: Get list of patterns
+  description: |
+    Get list of all patterns for the project
```
saraycp, line 4:
```suggestion
    Get a list of all patterns for the project
```

saraycp (changes requested):

danidoni (dismissed):

src/api/public/apidocs-new/components/schemas/source/project_pattern.yaml
rubhanazeem: Thanks for looking into it. It's updated

danidoni (approved):

saraycp (approved):

PR #13178, opened 2022-10-05
Intropduces Repository.to_param

hennevogel:
We almost exclusively use the repository name to refer to it. Let's make this the default for parameters instead of sprinkling the code base with .name calls


hennevogel:
codecov hangs. [coverage is fine](https://app.circleci.com/pipelines/github/openSUSE/open-build-service/8276/workflows/24696de8-06f9-4553-b7bc-cec0412c9a54/jobs/97494). merging!

PR #13181, opened 2022-10-05
Add Jacob to .gitmessage and sort the list alphabetically

saraycp:


hennevogel:
codecov hangs. coverage is fine. merging!

PR #13183, opened 2022-10-05
[backend] create output dir before reconstruct_container

M0ses:
Needs also to be backported to 2.10 branch


M0ses:
The problem only occurs if no $BSConfig::container_registries are configured.

mlschroe (approved):

PR #13189, opened 2022-10-06
Make brakeman use ruby 3.1.1

hellcp-work:
This prevents us from using some Ruby 3 syntax, and this is a simple way to fix that

(Yes, I did misspell brakeman in the branch name)


danidoni (approved):

hennevogel:
But we are using 3.1.2 in production :)

PR #13202, opened 2022-10-10
[api] add @on_medium xpath search for released binaries

adrianschroeter:
Allows to search for content of given media, like containers


eduardoj:
If this PR still makes sense, it would require to update the API documentation at least in this points:
- https://github.com/openSUSE/open-build-service/blob/93a444f2fd385f8069703202435fc40ed3e4e1a0/src/api/public/apidocs/paths/search_released_binary.yaml#L24
- https://github.com/openSUSE/open-build-service/blob/93a444f2fd385f8069703202435fc40ed3e4e1a0/src/api/public/apidocs/paths/search_released_binary_id.yaml#L23

eduardoj (changes requested):
See comment above.

PR #13213, opened 2022-10-13
Display relevant information about the actions in the action selector

hellcp-work:
This PR makes it so that the following information is always visible in the action selector dropdown:
* action type (through `action.name`)
* source_project/source_package (where applicable, only project or both)
* target_project/target_package (plus other targets if included in `action.name`)
* commit information (if available)

# To verify
1. Create a request with multiple actions
2. Click on `Select Action ▾` button and see a dropdown
3. Dropdown should include all the information specified above wherever applicable

image

### DIFF
--- a/src/api/app/views/webui/request/_action_text.html.haml
+++ b/src/api/app/views/webui/request/_action_text.html.haml
@@ -1,10 +1,28 @@
-- if (details = action.commit_details)
+- details = action.commit_details
+- if details
+  - if details.key?('comment')
+    %p.mt-0.mb-0.font-weight-normal
+      = details['comment'].truncate(50)
+    %p.mb-0.mt-0.text-muted.font-weight-light.small
+      = action.name
+  - else
+    %p.mt-0.mb-0.font-weight-normal
+      = action.name
+  %span.float-right.ml-5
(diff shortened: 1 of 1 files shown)


src/api/app/views/webui/request/_actions_details.html.haml
```diff
@@ -11,15 +11,19 @@
...
+            - if action_item.type == 'submit'
+              = link_to((render partial: 'action_text', locals: { action: action_item }),
+                        request_show_path(number: bs_request.number,
+                                          request_action_id: action_item.id,
+                                          full_diff: diff_limit,
+                                          diff_to_superseded: diff_to_superseded_id),
+                        class: "dropdown-item #{action_item.id == active_action.id ? 'disabled' : ''}")
+            - else
+              %span.dropdown-item.disabled
+                = render partial: 'action_text', locals: { action: action_item }
```
hennevogel, line 26: We purposefully skip everything beside `BsRequestActionSubmit` as we have no plan yet how to display them, no need to bring them in for this single thing.

hennevogel (changes requested):

hennevogel (approved):

PR #13217, opened 2022-10-14
Show comments and activity from superseded requests

danidoni:
This PR adds an expander/collapser link to toggle visibility of superseded comments and request history.

**Before**
image

**After**
_Collapsed_

image

_Expanded_

image

### DIFF
--- a/src/api/app/assets/stylesheets/webui/timeline.scss
+++ b/src/api/app/assets/stylesheets/webui/timeline.scss
@@ -1,11 +1,12 @@
 // This stylesheet defines styles for use in the comments and request history merged timeline
 .timeline {
     $timeline-border-width: 0.125rem;
+    $timeline-offset: -(($avatars-counter-size / 2) + $timeline-border-width);
     border-left: $timeline-border-width solid $gray-400;
 
     .timeline-item {
         position: relative;
-        left: -(($avatars-counter-size / 2) + $timeline-border-width); // The avatars of users involved in timeline items will be displayed on the timeline border
+        left: $timeline-offset; // The avatars of users involved in timeline items will be displayed on the timeline border
 
         $timeline-item-avatar-margin: 0.5rem;
         img.avatars-counter {
@@ -14,6 +15,32 @@
 
         .timeline-item-comment {
             margin-left: $avatars-counter-size + $timeline-border-width + $timeline-item-avatar-margin;
+        }
+
+        i.timeline-break, i.timeline-offset {
+            @extend .py-2;
+            @extend .mr-2;
+            @extend .fas;
+            @extend .fa-ellipsis-vertical;
+            margin-left: 0.66rem;
+        }
+        i.timeline-break {
+            @extend .bg-white;
+            @extend .text-dark;
+        }
+        a:hover i.timeline-break {
+            filter: none;
+        }
+        i.timeline-offset {
+            color: transparent;
+        }
+    }
+
+    .collapse, .collapsing {
+        position: relative;
+        left: $timeline-offset;
+        .timeline-item {
+            left: 0;
         }
     }
 }
--- a/src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
+++ b/src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
@@ -35,7 +35,40 @@
   .col-md-8.order-md-1.order-sm-2
     %h4.list-group.mb-4 Comments & Request History
     .timeline{ data: { comment_counter: local_assigns[:comment_counter_id] } }
+      .timeline-item
+        .d-inline-flex
+          - creator = User.find_by_login(bs_request.creator) || User.nobody
+          = image_tag_for(creator, size: 35, custom_class: 'rounded-circle bg-light border border-gray-400 avatars-counter')
+          %p
+            %i.fas.fa-code-commit.text-dark
+            = link_to(realname_with_login(creator), user_path(creator))
+            created this request
+            = link_to('#request-creation', title: l(bs_request.created_at.utc), name: 'request-creation') do
+              #{time_ago_in_words(bs_request.created_at)} ago
+            - if bs_request.superseding.any?
+              superseding
+              - bs_request.superseding.each do |superseded_request|
+                = link_to("request ##{superseded_request.number}", request_show_path(superseded_request.number, anchor: 'overview'))
+
+      - bs_request.superseding.each do |superseding_request|
+        .timeline-item.pb-4
+          = link_to('#collapse-superseding', class: 'd-flex align-items-center',
+              data: { toggle: 'collapse' },
+              aria: { expanded: false, controls: 'collapse-superseding' }) do
+            %i.timeline-break.expander>
+            %i.timeline-offset.collapser>
+
+            %i.fas.fa-fw.fa-chevron-right.expander.mr-1{ title: 'Show history' }>
+            %i.fas.fa-fw.fa-chevron-down.collapser.mr-1{ title: 'Hide history' }>
+            %span.expander
+              Expand history from superseded request ##{superseding_request.number}
+            %span.collapser
+              Collapse history from superseded request ##{superseding_request.number}
+        .collapse.mb-4#collapse-superseding
+          = render BsRequestActivityTimelineComponent.new(bs_request: superseding_request)
+
       = render BsRequestActivityTimelineComponent.new(bs_request: bs_request)
+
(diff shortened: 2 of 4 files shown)


danidoni:
> Would something like this make it more clear that it's a continuation of the same thing? When unfolded it would be a continuous line.

I gave it a spin: I like the idea of showing the history elements in a timeline fashion, but then it's a bit confusing that the first thing you read does not come from the request that you are processing... I believe we need to think a bit more on that.

hennevogel:
> but then it's a bit confusing that the first thing you read does not come from the request that you are processing... I believe we need to think a bit more on that.

But this is exactly the point of superseding something. The history of the superseded request **is** part of the request you are processing.

danidoni:
> > but then it's a bit confusing that the first thing you read does not come from the request that you are processing... I believe we need to think a bit more on that.
> 
> But this is exactly the point of superseding something. The history of the superseded request **is** part of the request you are processing.

Yes, what I find confusing is not that I see the history from the superseded request, but seeing it first, instead of the info of the request I'm visiting

danidoni:

image

I need to fix the alignment of those elements but I think it's better this way, so the first line puts you in context and then you can expand the history if you want.

What do you think?

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
```diff
@@ -29,7 +29,37 @@
...
+            = link_to(realname_with_login(creator), user_path(creator))
+            created this request
+            = link_to('#request-creation', title: l(bs_request.created_at.utc), name: 'request-creation') do
+              #{time_ago_in_words(bs_request.created_at)} ago
+            - if bs_request.superseding.any?
+              superseding
+              - bs_request.superseding.each do |superseded_request|
+                = link_to("request ##{superseded_request.number}", request_show_path(superseded_request.number, anchor: 'overview'))
+
+    - bs_request.superseding.each do |superseding_request|
```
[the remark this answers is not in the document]
danidoni: Done, please, check it out

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
```diff
@@ -29,7 +29,36 @@
...
+                = link_to("request ##{superseded_request.number}", request_show_path(superseded_request.number, anchor: 'overview'))
+
+      - bs_request.superseding.each do |superseding_request|
+        .timeline-item
+          %i.fas.fa-ellipsis-vertical.bg-white.pt-2.mb-4.pb-2{ style: "margin-left: 0.66rem"}
+          = link_to('#collapse-superseding',
+              data: { toggle: 'collapse' },
+              aria: { expanded: false, controls: 'collapse-superseding' }) do
+            %i.fas.fa-chevron-right.expander{ title: 'Show collapse' }
+            %i.fas.fa-chevron-down.collapser{ title: 'Hide collapse' }
```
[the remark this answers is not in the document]
danidoni: Added

src/api/app/assets/stylesheets/webui/timeline.scss
```diff
@@ -15,5 +15,24 @@
...
+            @extend .fas;
+            @extend .fa-ellipsis-vertical;
+            margin-left: 0.66rem;
+        }
+        i.timeline-break {
+            @extend .bg-white;
+            @extend .text-dark;
+        }
+        a:hover i.timeline-break {
+          filter: none!important;
```
dmarcoux, line 32: A space is missing:

```suggestion
          filter: none !important;
```

We should also explain in a comment why `!important` is needed here and above in the file _collapse-component.scss_.

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
```diff
@@ -33,7 +33,38 @@
...
+
+      - bs_request.superseding.each do |superseding_request|
+        .timeline-item
+          = link_to('#collapse-superseding',
+              data: { toggle: 'collapse' },
+              aria: { expanded: false, controls: 'collapse-superseding' }) do
+            %i.timeline-break.expander>
+            %i.timeline-offset.collapser>
+
+            %i.fas.fa-fw.fa-chevron-right.expander{ title: 'Show collapse' }>
```
dmarcoux, line 59: The title is not really needed here since we have the text right next to this icon. Same for the icon on the next line. If you prefer to keep the titles, I suggest changing _collapse_ to _history_:

```suggestion
            %i.fas.fa-fw.fa-chevron-right.expander{ title: 'Show history' }>
```

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
```diff
@@ -33,7 +33,38 @@
...
+            %i.fas.fa-fw.fa-chevron-down.collapser{ title: 'Hide collapse' }>
+            Expand history from superseded request
```
dmarcoux, line 61: It's a bit confusing that this text still is `Expand history (...)` when the history was already expanded like in this example:

confusing

The text should reflect the state of the collapse element.

dmarcoux (changes requested):

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
danidoni: Hm, good one

src/api/app/assets/stylesheets/webui/timeline.scss
danidoni: It was not needed. I removed it.

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
```diff
@@ -33,7 +33,42 @@
...
+          = link_to('#collapse-superseding',
+              data: { toggle: 'collapse' },
+              aria: { expanded: false, controls: 'collapse-superseding' }) do
+            %i.timeline-break.expander>
+            %i.timeline-offset.collapser>
+
+            %i.fas.fa-fw.fa-chevron-right.expander{ title: 'Show history' }>
+            %i.fas.fa-fw.fa-chevron-down.collapser{ title: 'Hide history' }>
+            %span.expander
+              Expand history from superseded request
```
dmarcoux, line 62: To be consistent with the rest of the page, we should add `#` in front of the request number. Same for the `Collapse (...)` part.

dmarcoux:
Somehow, the avatars are cut in half when the history of the superseded request is expanding/collapsing. This isn't a blocker, but it would be nice to fix this small issue. Look at my cursor in this GIF to see the issue:

glitch

src/api/app/assets/stylesheets/webui/timeline.scss
```diff
@@ -15,5 +15,24 @@
...
+            @extend .fas;
+            @extend .fa-ellipsis-vertical;
+            margin-left: 0.66rem;
+        }
+        i.timeline-break {
+            @extend .bg-white;
+            @extend .text-dark;
+        }
+        a:hover i.timeline-break {
+          filter: none;
```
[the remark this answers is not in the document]
danidoni: Fixed (I misunderstod this as "I already fixed this wrong indentation")

dmarcoux:
To understand where @hellcp-work came up with `-0.84375rem`, I looked up where this could come from and it's exactly this:

https://github.com/openSUSE/open-build-service/blob/a37868ae778e212287d1e5ae6514857699396360/src/api/app/assets/stylesheets/webui/timeline.scss#L8

So if this suggestion fixes the issue, let's use a Sass variable to store this calculation and re-use it in the suggestion and the code I just shared. This way, we do not rely on what looks like a "magic" number.

src/api/app/assets/stylesheets/webui/timeline.scss
```diff
@@ -15,5 +16,31 @@
...
+            @extend .text-dark;
+        }
+        a:hover i.timeline-break {
+            filter: none;
+        }
+        i.timeline-offset {
+            color: transparent;
+        }
+        .collapse, .collapsing {
+  		position: relative;
```
dmarcoux, line 39: This indentation should be fixed.

dmarcoux:
The latest commit fixed the avatars when expanding/collapsing the history :+1:.

Sadly, I found another issue on mobile, see the expand/collapse text on mobile:
mobile

danidoni:
> Let me have a go at fixing that too real quick

Awesome!

danidoni:
Squashing again

PR #13225, opened 2022-10-17
Remove use of CSS class details-with-coderay

eduardoj:
The definition of the CSS class was removed in commit: 433b7d74a5311269281e524f33d10f52beb5b712

The use of this class should have been also removed in that commit.


krauselukas (approved):

krauselukas:
codecov hangs... merging

PR #13226, opened 2022-10-17
Move review request action button closer to context area

ncounter:
Moving the `Add a request` button within the conversation page, closer to the side link area related to the review section.

Note: the `_actions.html.haml` has been marked `TODO - to drop` because after the roll out of the beta feature redesign, it will be no longer needed.

## After
feature-after

Fixes https://trello.com/c/p313TsCI

### DIFF
--- a/src/api/app/views/webui/request/_actions.html.haml
+++ b/src/api/app/views/webui/request/_actions.html.haml
@@ -1,3 +1,4 @@
+-# TODO: drop this partial when `request_show_redesign` feature is rolled out
 - content_for :actions do
   %li.nav-item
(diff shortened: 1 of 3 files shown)


ncounter:
[Move the Add a Review action below the Reviewers list (S)](https://trello.com/c/p313TsCI/2013-move-the-add-a-review-action-below-the-reviewers-list-s)

danidoni:
> When a request doesn't have a review yet, the changes from this PR prevent someone from adding a review, even if they have the right permissions. This is wrong and it should be fixed.
> 
> Even though this has already been deployed, nobody noticed this issue since a request always has reviews automatically added whenever it's created. In some edge cases like when a target project/package of a request doesn't have a maintainer, this issue will be problematic.

I just found it out now, when creating a request from a branched package locally

PR #13227, opened 2022-10-17
Refactoring "review" side section

ncounter:
Refactoring:
- do not use a link implementation for an action button, use a proper button instead
- do not group reviews by type (accepted|open|declined), just list them by timeline
 
Rewording:
- `add review` --> `add reviewer`

Additional changes:
- show _who was added_ as a reviewer in the history

## After
image

### DIFF
--- a/src/api/app/components/add_review_dropdown_component.html.haml
+++ b/src/api/app/components/add_review_dropdown_component.html.haml
@@ -7,10 +7,10 @@
   .dropdown#add-review-dropdown-component
     .btn.btn-success.dropdown-toggle{ type: 'button', data: { toggle: 'dropdown' }, aria: { expanded: 'false' } }
       Review
-    .dropdown-menu.dropdown-menu-right
+    .dropdown-menu.dropdown-menu-left
       %h5.dropdown-header Give a review for...
       - @my_open_reviews.each do |review|
-        = button_tag(type: 'button', class: 'dropdown-item m-2', data: { toggle: 'collapse', target: '#review-form-collapse',
+        = button_tag(type: 'button', class: 'dropdown-item', data: { toggle: 'collapse', target: '#review-form-collapse',
                     review: review.id }, aria: { expanded: 'false', controls: 'review-form-collapse' }) do
           = reviewer_icon_and_text(review: review)
 
--- a/src/api/app/models/history_element/request_review_added.rb
+++ b/src/api/app/models/history_element/request_review_added.rb
@@ -1,12 +1,28 @@
 module HistoryElement
   class RequestReviewAdded < HistoryElement::Request
-    # self.description_extension is review id
     def description
       'Request got a new review request'
     end
 
     def user_action
-      'added a reviewer'
+      "#{user_action_prefix} #{action_target} #{user_action_suffix}"
+    end
+
+    def user_action_prefix
+      'added'
+    end
+
+    def action_target
+      review.reviewed_by
+    end
+
+    def user_action_suffix
+      'as a reviewer'
+    end
+
+    # self.description_extension is review id
+    def review
+      ::Review.find(description_extension)
     end
   end
 end
--- a/src/api/app/models/history_element/request_review_added.rb
+++ b/src/api/app/models/history_element/request_review_added.rb
@@ -6,7 +6,7 @@ module HistoryElement
     end
 
     def user_action
-      'added review'
+      'added a reviewer'
     end
   end
 end
--- a/src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
+++ b/src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
@@ -1,18 +1,17 @@
 .row
   .col-md-12
-    .mb-4#description-text
-      - if bs_request.description.present?
-        = render partial: 'webui/shared/collapsible_text', locals: { text: bs_request.description }
-      - else
-        %i No description set
+    .d-flex.justify-content-between.mb-4
+      #description-text
+        - if bs_request.description.present?
+          = render partial: 'webui/shared/collapsible_text', locals: { text: bs_request.description }
+        - else
+          %i No description set
 .row
   = render partial: 'webui/request/beta_show_tabs/conversation_aside',
            locals: { bs_request: bs_request,
              can_add_reviews: can_add_reviews,
              my_open_reviews: my_open_reviews,
-             accepted_reviews: accepted_reviews,
-             declined_reviews: declined_reviews,
-             open_reviews: open_reviews,
+             request_reviews: request_reviews,
              open_reviews_for_staging_projects: open_reviews_for_staging_projects,
              package_maintainers: package_maintainers,
              project_maintainers: project_maintainers,
--- a/src/api/app/views/webui/request/beta_show_tabs/_conversation_aside.html.haml
+++ b/src/api/app/views/webui/request/beta_show_tabs/_conversation_aside.html.haml
@@ -1,23 +1,18 @@
 .col-md-4.order-md-2.order-sm-1.mb-4#side-links
   -# REVIEWERS
-  - if accepted_reviews.present? || declined_reviews.present? || open_reviews.present?
-    .d-flex.justify-content-between
-      %h6
-        Reviews
-      = render AddReviewDropdownComponent.new(bs_request: bs_request, user: User.session,
+  - if request_reviews.present?
+    .mt-4
+      .text-right
+        = render AddReviewDropdownComponent.new(bs_request: bs_request, user: User.session,
                                               can_add_reviews: can_add_reviews,
                                               my_open_reviews: my_open_reviews)
-    = render AddReviewCollapsibleComponent.new
-    .mb-4
-      = render partial: 'webui/request/beta_show_tabs/review_summary', collection: accepted_reviews, as: :review
-      = render partial: 'webui/request/beta_show_tabs/review_summary', collection: declined_reviews, as: :review
-      = render partial: 'webui/request/beta_show_tabs/review_summary', collection: open_reviews, as: :review
-
-  - if can_add_reviews
-    = link_to('#', data: { toggle: 'modal', target: '#add-reviewer-modal' }, title: 'Add a Review', class: 'ml-3') do
-      %i.fas.fa-plus-circle.fa-lg.mr-1
-      %span.nav-item-name Add a Review
-    = render partial: 'add_reviewer_dialog'
+      .mb-2
+        %h6
+          Reviewers
+      = render AddReviewCollapsibleComponent.new
+      .mb-4
+        = render partial: 'webui/request/beta_show_tabs/review_summary', collection: request_reviews, as: :review
+  = render partial: 'webui/request/beta_show_tabs/ask_for_review', locals: { can_add_reviews: can_add_reviews }
 
   - open_reviews_for_staging_projects.each do |review|
     .pl-3
--- a/src/api/app/components/bs_request_history_element_component.html.haml
+++ b/src/api/app/components/bs_request_history_element_component.html.haml
@@ -6,6 +6,11 @@
     - if element.instance_of?(HistoryElement::RequestSuperseded)
       superseded this request with
       = link_to("request ##{element.description_extension}", request_show_path(element.description_extension))
+    - elsif element.instance_of?(HistoryElement::RequestReviewAdded)
+      = element.user_action_prefix
+      %strong
+        = element.action_target
+      = element.user_action_suffix
     - else
       = element.user_action
     = link_to("#status-history-#{element.id}", title: l(element.created_at.utc), name: "status-history-#{element.id}") do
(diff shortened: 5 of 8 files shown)


src/api/app/models/history_element/review_accepted.rb
```diff
@@ -5,7 +5,7 @@ def description
     end
 
     def user_action
-      'accepted review'
+      'accepted the request'
```
hennevogel, line 8: In "normal" circumstances (not in all) accepting the request might happen if you accept the last review. But not always.

src/api/app/components/add_review_dropdown_component.html.haml
```diff
@@ -2,15 +2,16 @@
   .btn.btn-success.toggle-review-form{ type: 'button', data: { toggle: 'collapse', target: '#review-form-collapse',
                                        review: @my_open_reviews.first.id, reviewer_icon: reviewer_icon_and_text(review: @my_open_reviews.first) },
                                        aria: { expanded: 'false', controls: 'review-form-collapse' } }
-    Review
+    Review as
+    = @my_open_reviews.first.reviewer
```
hennevogel, line 6: I don't like this UX wise. This gives the role more emphasis than it has, the important part is giving the review. That a users has 2 reviews open for different roles is a corner case anyway...

hennevogel (changes requested):

src/api/app/models/history_element/review_accepted.rb
ncounter: I am not sure I get it, maybe there are corner cases I am not aware of.
My main puzzling part starts by having "the result of a review" as a type of itself. A review should be _asked_ or _given_, and the result of that action is _accepting|declining_ a request. Whenever I submit a review I accept or decline the request, not the review. Am I missing something?

Maybe, as an agreement, we could reword this into `submitted a review`?

src/api/app/components/add_review_dropdown_component.html.haml
hennevogel: For the sake of recognition I would have the button say the same thing in all circumstances that that is "Review"

src/api/app/models/history_element/review_accepted.rb
hennevogel: >  A review should be asked or given, and the result of that action is accepting|declining a request.

Exactly here lies your misunderstanding.  A review *is* asked for and the result of that action is accepting|declining **the review**.

src/api/app/components/add_review_dropdown_component.html.haml
ncounter: done

src/api/app/models/history_element/review_accepted.rb
ncounter: I reverted the change.

It's getting out of scope as it is more about the workflow and the logic behind it, that is probably a bit tricky. I will track this down in a separate issue/card.

Thank you so far

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
```diff
@@ -7,22 +7,20 @@
...
-    - if accepted_reviews.present? || declined_reviews.present? || open_reviews.present?
+    - if request_reviews.present?
       .d-flex.justify-content-between
         %h4
-          Reviews
-        = render AddReviewDropdownComponent.new(bs_request: bs_request, user: User.session,
-                                                can_add_reviews: can_add_reviews,
-                                                my_open_reviews: my_open_reviews)
+          Reviewers
+        %button.btn.btn-link{ data: { toggle: 'modal', target: '#add-reviewer-modal' }, title: 'Ask for a Review' }
```
dmarcoux, line 14: The alignment of this button with the heading text is off:

alignment-is-off

dmarcoux (changes requested):

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
ncounter: As it was before unfortunately
offset

I'll try to get it adjusted anyway

ncounter: It's a workaround more than a fix --> https://github.com/openSUSE/open-build-service/pull/13227/commits/86f0dea7b4f8a60753ea769fe60fd9ed6983d9bf

danidoni: Why it's a workaround and not a fix? :thinking:

ncounter: Because we are adding a specific margin-top value (`mt-2`) to the `<h4>` to vertically align it to the button, but if something in that context will change (button top padding or margin, etc) the alignment will need another round of adjustment.
But it can also be just a matter of definition, it actually fixes it :slightly_smiling_face:

danidoni: Okay, I understand it now. Thanks!

danidoni (approved):

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
```diff
@@ -7,24 +7,19 @@
...
+    - if request_reviews.present?
       .d-flex.justify-content-between
-        %h4
-          Reviews
-        = render AddReviewDropdownComponent.new(bs_request: bs_request, user: User.session,
-                                                can_add_reviews: can_add_reviews,
-                                                my_open_reviews: my_open_reviews)
+        %h4.mt-2
+          Reviewers
+        = render partial: 'webui/request/beta_show_tabs/ask_for_review', locals: { can_add_reviews: can_add_reviews }
```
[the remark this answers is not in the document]
ncounter: Why do you feel it is _counter intuitive_?

Let me drive you through the rationale I went into: we have **two** action buttons here, one to add a new reviewer, one to submit a review. From the button action type perspective, the submit one is stronger, and it has to go at the bottom. Therefore the add one becomes an additional action of the box, and it is very related to the title of it: in the end it adds a reviewer. It can go to the right top because it belongs more to some potential _toolbar_ with properties and actions within that context, and toolbars always (almost) go at the top.

Anyway your early feedback is important, I'll keep thinking about it. Thank you.

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
ncounter: As discussed offline, `m-0` to the `h4` and `.align-items-center` to the container does the fix as well, deleting the `h4` bottom margin default, but moving it up to the header container level.

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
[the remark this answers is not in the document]
ncounter: After thinking it twice and trying to _merge_ all our points of view, I came up with the following:
- the **Review** button is related to the reviewer list but it is moreover related to the Request itself, so that action button belongs more to the whole page (thinking about GitHub PRs, to name one). Therefore I will move the **Review** button outside of the reviewer side box and place it on top of the conversation tab content.
- because of the change as described above, the **ask for a review** button can gain back its original place at the bottom of the reviewer list, because now it is the only button related to that box and on top of that it can gain back a default action button style
refactoring-alt

Thanks for the discussion we came up with, I think, and even better solution of all the other options.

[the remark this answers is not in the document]
ncounter: Resolving this conversation as we agreed after offline discussion.

ncounter:
@hennevogel I addressed the changes you requested, could you give a second round of a review? Thank you

src/api/app/views/webui/request/beta_show_tabs/_ask_for_review.haml
```diff
@@ -0,0 +1,5 @@
+- if can_add_reviews
+  %button.btn.btn-outline-secondary{ data: { toggle: 'modal', target: '#add-reviewer-modal' }, title: 'Ask for a Review' }
+    %i.fas.fa-plus-circle.mr-1.text-outline-secondary
+    %span.nav-item-name Ask for a Review
```
hennevogel, line 4:
```suggestion
    %span.nav-item-name Add Reviewer
```

src/api/app/views/webui/request/beta_show_tabs/_ask_for_review.haml
```diff
@@ -0,0 +1,5 @@
+- if can_add_reviews
+  %button.btn.btn-outline-secondary{ data: { toggle: 'modal', target: '#add-reviewer-modal' }, title: 'Ask for a Review' }
```
hennevogel, line 2:
```suggestion
  %button.btn.btn-outline-secondary{ data: { toggle: 'modal', target: '#add-reviewer-modal' }, title: 'Add Reviewer' }
```

src/api/app/models/history_element/request_review_added.rb
```diff
@@ -6,7 +6,7 @@ def description
     end
 
     def user_action
-      'added review'
+      'asked for a review'
```
hennevogel, line 9:
```suggestion
      'added a reviewer'
```

hennevogel (changes requested):

src/api/app/models/history_element/request_review_added.rb
ncounter: Done

src/api/app/views/webui/request/beta_show_tabs/_ask_for_review.haml
ncounter: Done

src/api/app/views/webui/request/beta_show_tabs/_ask_for_review.haml
ncounter: Done

ncounter:
> The review button looks like it has some margin at the bottom?
> 
> image

Mmmmm, this is new, and also not happening on my side on multiple browsers (FireFox, Chrome, etc...). Could it be you have some CSS cache or hitting some corner case?

hennevogel:
@ncounter this is just viewing https://obs-reviewlab.opensuse.org/ncounter-ncouter-move-review-request-button-refactoring/request/show/9

ncounter:
> @ncounter this is just viewing https://obs-reviewlab.opensuse.org/ncounter-ncouter-move-review-request-button-refactoring/request/show/9

This is fixed now: that issue happens when the Review is for one single option (not a dropdown button). The description of the review had a margin bottom rule that was targeting actually its container so I moved outside and not everything is aligned and with the right height.

src/api/app/models/history_element/request_review_added.rb
```diff
@@ -1,12 +1,16 @@
 module HistoryElement
   class RequestReviewAdded < HistoryElement::Request
-    # self.description_extension is review id
     def description
       'Request got a new review request'
     end
 
     def user_action
-      'added review'
+      "added <strong>#{review.reviewed_by}</strong> as a reviewer".html_safe
```
dmarcoux, line 8: It's totally out of scope for this PR, but this will have to be refactored one day. This code doesn't belong in a model, it's view code. Could you please create a card in the request workflow redesign backlog for this? It should be a refactoring of all HistoryElement models and the views where they are used.

ncounter: @dmarcoux : for the time being, until they get refactored them all, what do you think about the following implementation to get rid of the `html_safe` call and still have the additional view output? https://github.com/openSUSE/open-build-service/pull/13227/commits/9d388407cb0e11fe5e9d3d87892f69350e8908b9 (CC-ing @hennevogel)

dmarcoux: It's good enough for now. This will be refactored in the future anyway.

src/api/app/views/webui/request/beta_show_tabs/_conversation_aside.html.haml
```diff
@@ -1,30 +1,25 @@
...
-      %i.fas.fa-plus-circle.fa-lg.mr-1
-      %span.nav-item-name Add a Review
-    = render partial: 'add_reviewer_dialog'
+      .mb-2
+        %h6
+          Reviewers
+      = render AddReviewCollapsibleComponent.new
+      .mb-4
+        = render partial: 'webui/request/beta_show_tabs/review_summary', collection: request_reviews, as: :review
+      = render partial: 'webui/request/beta_show_tabs/ask_for_review', locals: { can_add_reviews: can_add_reviews }
```
dmarcoux, line 15: Displaying the `Add reviewer` button shouldn't depend on the presence of reviews for the request. This partial should be outside the `if request_reviews.present?` statement.

dmarcoux (changes requested):

src/api/app/views/webui/request/beta_show_tabs/_conversation_aside.html.haml
ncounter: You're right, I think I broke indentation during the refactoring. Fixing it

dmarcoux (approved):

PR #13232, opened 2022-10-18
[api] fix incomplete releases via token

adrianschroeter:
It failed with an absurd error in the middle of a release when another
releasetarget lacked the permissions.

The `release_package` method of the `MaintenanceHelper` mixin checks
for permission on the given release target/project. But to avoid
incomplete releases, we need to check all release targets of the
collection upfront, before we start to release them one by one.

### DIFF
--- a/src/api/app/models/token/release.rb
+++ b/src/api/app/models/token/release.rb
@@ -10,6 +10,7 @@ class Token::Release < Token
     manual_release_targets = package_to_release.project.release_targets.where(trigger: 'manual')
     raise NoReleaseTargetFound, "#{package_to_release.project} has no release targets that are triggered manually" unless manual_release_targets.any?
 
+    # releasing ...
     manual_release_targets.each do |release_target|
       opts = { filter_source_repository: release_target.repository,
                manual: true,
--- a/src/api/app/policies/token/release_policy.rb
+++ b/src/api/app/policies/token/release_policy.rb
@@ -1,7 +1,19 @@
 class Token::ReleasePolicy < TokenPolicy
   def trigger?
     return false unless user.is_active?
+    return false unless sufficient_permission_on_all_release_targets?
 
     PackagePolicy.new(user, record.object_to_authorize).update?
+  end
+
+  private
+
+  # we need to check all release targets upfront to avoid incomplete releases
+  def sufficient_permission_on_all_release_targets?
+    record.object_to_authorize.project.release_targets.where(trigger: 'manual').each do |release_target|
+      unless ProjectPolicy.new(user, release_target.target_repository.project).update?
+        raise Pundit::NotAuthorizedError, query: :trigger?, record: release_target.target_repository.project, reason: :unsufficient_permission_on_release_target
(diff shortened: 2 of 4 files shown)


adrianschroeter:
(it is not enough to check permissions on token creation time, as permissions may change afterwards)

src/api/app/models/token/release.rb
```diff
@@ -10,6 +10,14 @@ def call(options)
     manual_release_targets = package_to_release.project.release_targets.where(trigger: 'manual')
     raise NoReleaseTargetFound, "#{package_to_release.project} has no release targets that are triggered manually" unless manual_release_targets.any?
 
+    # first a permission check to avoid incomplete releases
+    manual_release_targets.each do |release_target|
+      unless User.session!.can_modify?(release_target.target_repository.project)
+        raise CmdExecutionNoPermission, "no permission to write in project #{release_target.target_repository.project.name}"
```
krauselukas, line 16: `CmdExecutionNoPermission` throws an uninitialized constant error, will have a look

krauselukas (changes requested):

src/api/app/models/token/release.rb
krauselukas: okay this is addressed

krauselukas:
let me move this to a pundit policy

src/api/app/policies/token/release_policy.rb
```diff
@@ -1,7 +1,17 @@
...
 
     PackagePolicy.new(user, record.object_to_authorize).update?
   end
+
+  private
+
+  # we need to check all release targets upfront to avoid incomplete releases
+  def sufficient_permission_on_all_release_targets?
+    record.object_to_authorize.project.release_targets.where(trigger: 'manual').each do |release_target|
+      return false unless user.can_modify?(release_target.target_repository.project)
```
dmarcoux, line 14: For this authorization check, we should also use Pundit with the `ProjectPolicy#update?`: https://github.com/openSUSE/open-build-service/blob/a37868ae778e212287d1e5ae6514857699396360/src/api/app/policies/project_policy.rb#L12

krauselukas: @dmarcoux good point, done

dmarcoux (approved):

src/api/app/policies/token/release_policy.rb
```diff
@@ -1,7 +1,17 @@
...
 
     PackagePolicy.new(user, record.object_to_authorize).update?
   end
+
+  private
+
+  # we need to check all release targets upfront to avoid incomplete releases
+  def sufficient_permission_on_all_release_targets?
+    record.object_to_authorize.project.release_targets.where(trigger: 'manual').each do |release_target|
+      return false unless ProjectPolicy.new(user, release_target.target_repository.project).update?
```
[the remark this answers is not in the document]
krauselukas: @hennevogel please have a look, created a custom message

hennevogel (approved):

PR #13236, opened 2022-10-18
Isolate and slightly refactor build results' code

saraycp:
Refactor the bare minimum to make Build Results work in the new request redesigned page, in the new Build Results tab. 

Main changes:

- Extract the code of the build results to their own files, to make it independent of the rpm lint code. 
- Remove previously existing _Build Results_ and _RPM Lint_ tabs.
- Adapt JavaScript accordingly.
- Stop using `index` to identify the different actions in one request.
 
 This is how it looks so far: 
 
Screenshot 2022-10-19 at 11-03-56 Open Build Service

`codeclimate` complains about the long `RequestController#show` action, we are ignoring it during the implementation of `request_show_redesign`.

**How to test?**

I'd compare each case with and without the feature flag enabled.

Check different cases:
- When the request contains multiple actions (make sure the source project has working repositories set). Switch between actions. Example [here](https://obs-reviewlab.opensuse.org/saraycp-redesign_build_results_i/request/show/9/request_action/9#tab-pane-build-results).
- When the source project has no repositories set. Example [here](https://obs-reviewlab.opensuse.org/saraycp-redesign_build_results_i/request/show/7#tab-pane-build-results).
- When there are no build results. Example [here](https://obs-reviewlab.opensuse.org/saraycp-redesign_build_results_i/request/show/2#tab-pane-build-results).
- When all the results are excluded and hidden. Example [here](https://obs-reviewlab.opensuse.org/saraycp-redesign_build_results_i/request/show/8#tab-pane-build-results).

### DIFF
--- a/src/api/app/assets/javascripts/webui/application.js
+++ b/src/api/app/assets/javascripts/webui/application.js
@@ -57,5 +57,6 @@
 //= require webui/long_text.js
 //= require webui/new_watchlist/collapsible_tooltip.js
 //= require webui/request_show_redesign/add_review.js
+//= require webui/request_show_redesign/build_results.js
 //= require webui/delete_confirmation_dialog.js
 //= require webui/nav_tabs.js
--- a/src/api/app/assets/javascripts/webui/buildresult.js
+++ b/src/api/app/assets/javascripts/webui/buildresult.js
@@ -1,3 +1,7 @@
+// TODO: replace with the content of
+// app/assets/javascripts/webui/request_show_redesign/build_results.js
+// after the rollout of 'request_show_redesign'.
+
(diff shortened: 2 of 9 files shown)


danidoni (approved):
I believe we can move all the build_results partials into a component. But we can do it elsewhere

saraycp:
> I believe we can move all the build_results partials into a component. But we can do it elsewhere

All the code related to Build Results needs refactoring. Moving part of it to components is an option, yes.

dmarcoux (approved):
As @saraycp wrote, some of the code will have to be refactored, but it's already a good start :+1:

PR #13240, opened 2022-10-19
Refactor EventMailer

hennevogel:
- Make it more ActionMailer to it's getting clearer what happens here
- Move default headers from Event::Base into the mailer


src/api/app/mailers/event_mailer.rb
```diff
@@ -1,55 +1,44 @@
...
-    subscribers = subscribers.to_a
-    return if subscribers.empty?
+    # FIXME: Use action_mailer.default_url_options instead
+    # https://guides.rubyonrails.org/action_mailer_basics.html#generating-urls-in-action-mailer-views
+    @host = ::Configuration.obs_url
 
-    recipients = subscribers.map(&:display_name).compact_blank
-    return if recipients.empty?
+    sender = event.originator
+    sender ||= DefaultSender.new(::Configuration.admin_email, 'OBS Notification')
```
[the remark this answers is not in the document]
hennevogel: If we would have something like that, sure. But we don't...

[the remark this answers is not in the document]
hennevogel: There is `Configuration.name` but it doesn't really work, not sure it ever did.

src/api/app/jobs/send_event_emails_job.rb
```diff
@@ -23,7 +23,7 @@ def perform
   def send_email(subscribers, event)
     return if subscribers.empty?
 
-    EventMailer.event(subscribers, event).deliver_now
+    EventMailer.with(subscribers: subscribers, event: event).notification_email.deliver_now
```
danidoni, line 26: According to the `EventMailer` [below](https://github.com/openSUSE/open-build-service/pull/13240/files#diff-97983eb9316b1f8a2883c0380827551d4e8a2c27edb6978860107025c9572bc0R13), if `@recipients` or `@event` is blank, `notification_email` will return `nil`. Looks like this will crash with undefined method `.deliver_now` for NilClass or something like this...

hennevogel: Yeah but it's never nil....

danidoni: But then, if it's never nil, why do bother checking for `@recipients` or `@event`?

hennevogel: Dunno actually 😸

danidoni (approved):

PR #13247, opened 2022-10-20
Fix issue retrieving building_package

saraycp:
Sometimes the building/source package is not available from the request page because the package was removed.

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -116,8 +116,7 @@ class Webui::RequestController < Webui::WebuiController
       @refresh = @action[:diff_not_cached]
 
       # Handling build results
-      @building_package = building_package
-      @building_project = @bs_request.staged_request? ? @bs_request.staging_project : @building_package.project
+      @staging_project = @bs_request.staging_project.name unless @bs_request.staging_project_id.nil?
 
       if @refresh
         bs_request_action = BsRequestAction.find(@action[:id])
@@ -434,13 +433,5 @@ class Webui::RequestController < Webui::WebuiController
       show_project_maintainer_hint: @show_project_maintainer_hint,
       actions: @actions
     }
-  end
-
-  def building_package
-    @building_package ||= Package.get_by_project_and_name(@active_action.source_project,
-                                                          @active_action.source_package,
-                                                          use_source: false,
-                                                          follow_multibuild: true,
-                                                          follow_project_links: true)
   end
 end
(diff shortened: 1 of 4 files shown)


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -436,11 +436,10 @@ def render_request_update
     }
   end
 
+  # TODO: Improve this code. Call require_package instead?
```
hennevogel, line 439: Does this make this a draft PR? :)

saraycp: No, the comment is there to improve it later. It's no longer a draft PR.

hennevogel: Have you created a card in the backlog? :)

saraycp: I'm already working on the improvement. But I prefer to have the quick fix merged and not monkey patched ASAP.

saraycp:
As this PR is still open, I pushed the improvements as well. Please have a look @hennevogel and @krauselukas

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -116,8 +116,7 @@ def show
       @refresh = @action[:diff_not_cached]
 
       # Handling build results
-      @building_package = building_package
-      @building_project = @bs_request.staged_request? ? @bs_request.staging_project : @building_package.project
+      @staging_project = @bs_request.staging_project.name unless @bs_request.staging_project_id.nil?
```
danidoni, line 119: Why this `@building_project` -> `@staging_project` renaming? This can also happen outside of staging projects, right?

saraycp: Sorry for the confusion, `@staging_project` is not replacing `@building_project`.

I stopped passing `@building_package` and `@building_project` objects to the partial and pass only their names. But the project name can be either the project given by the action or a staging project. That's why I'm retrieving the staging project name in that line and afterwards I pass one or the other:

```
= render partial: 'webui/request/beta_show_tabs/build_results', locals: { project: @staging_project || @action[:sprj],
```

Once inside the partial, how do we know if the passed project name corresponds to the staging project or the project in the action? Before, I passed  `@bs_request` to the partial and checked inside. Now I decided not to pass the `@bs_request` but only a boolean `is_staged_request: @staging_project.present?`.

I hope it's clearer now, please ask me if not.

danidoni: Okay, clear

krauselukas (approved):

krauselukas:
@saraycp I'm fine with ignoring codeclimate for now, since it was an issue before

PR #13260, opened 2022-10-24
[api] Add information about autoaccept in notifications

adrianschroeter:
The mails hide the fact that auto-accept requests will get accept when no reaction happens.


adrianschroeter:
I did not test that, I have to admit ...

PR #13263, opened 2022-10-24
Fix actions' hrefs when the url contains a line diff

eduardoj:
There is no need to include the `action_id` in the fraction identifier (anchor) of the URL. In case of browsing an action, that same action is already included in the URL. Therefore, I reverted the three commits related to this change.

In case we browse a diff line, change `href` attributes of other actions' links to point to the Changes tab.

### DIFF
--- a/src/api/app/components/sourcediff_component.html.haml
+++ b/src/api/app/components/sourcediff_component.html.haml
@@ -31,8 +31,7 @@
                                                                                   index: file_index,
                                                                                   last: sourcediff['filenames'].last == filename,
                                                                                   package: @action[:spkg],
-                                                                                  linkinfo: nil,
-                                                                                  action_id: @action[:id] }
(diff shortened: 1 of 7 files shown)


eduardoj:
> This is not entirely true, you can load up https://obs-reviewlab.opensuse.org/eduardoj-request_show_redesignfix_sharable_urls/request/show/9#diff_0_n2 (which is a link that doesn't explicitly state which action it is), but the first action will presumably always be the same, so it shouldn't matter.

Yes, you are right: in the case of loading a multi-action request, not specifying the action id in the URL results in browsing the first action id.

My point is that, given the different urls which show changes in a request with or without multiple actions, it is always defined which action will be shown, already in the URL. So no need of replicate it in the fraction identifier (anchor).

eduardoj:
codecov reporting hangs, merging...

PR #13277, opened 2022-10-26
Remove reference to the user.rng schema in the GET watchlist endpoint documentation

eduardoj:
No validation schema is used in this endpoint.


danidoni (approved):

eduardoj:
codecov reporting hangs, merging...

Issue #13290, opened 2022-10-28
Incorrect scmsync tag set on branch_package step for SCM/CI integration


krauselukas:
# Issue Description

A [PR](https://github.com/SUSE/BCI-dockerfile-generator/pull/86) against the 'tumbleweed' branch with the following `workflow.yml` resulted in setting an incorrect `<scmsync>` tag in the meta of the branched package.

*workflow.yml:*
```
---
workflow:
  steps:
    - branch_package:
        source_project: home:dancermak:BCI:Tumbleweed
        source_package: 389-ds-container
        target_project: home:dancermak:BCI:Tumbleweed:Staging
```

Resulting package meta on [branched package](https://build.opensuse.org/package/meta/home:dancermak:BCI:Tumbleweed:Staging:SUSE:BCI-dockerfile-generator:PR-86/389-ds-container):
```
<package name="389-ds-container" project="home:dancermak:BCI:Tumbleweed:Staging:SUSE:BCI-dockerfile-generator:PR-86">
  <title/>
  <description/>
  <scmsync>https://github.com/SUSE/BCI-dockerfile-generator.git?subdir=389-ds-container#tumbleweed?subdir=389-ds-container#cc3a337b4682ae040016a4a11e59cc30d581a475</scmsync>
</package>
```

This is the incorrect syntax to point to the new commit using the scm bridge.
See https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_bridge.html

Expected Result
============

The package meta of of the  [branched package](https://build.opensuse.org/package/meta/home:dancermak:BCI:Tumbleweed:Staging:SUSE:BCI-dockerfile-generator:PR-86/389-ds-container) should look like this:

```
<package name="389-ds-container" project="home:dancermak:BCI:Tumbleweed:Staging:SUSE:BCI-dockerfile-generator:PR-86">
  <title/>
  <description/>
  <scmsync>https://github.com/SUSE/BCI-dockerfile-generator.git?subdir=389-ds-container#cc3a337b4682ae040016a4a11e59cc30d581a475</scmsync>
</package>
```

How to Reproduce
===============
1. Create a package with the name `389-ds-container`
2. Add the `<scmsync>` of the failing example to the package meta and see it failing by not fetching the sources...
3. Modify the `<scmsync>` tag again, now exchanging it with the correct example. See it pulling the correct sources... (please be aware that the commit sha can change whenever someone pushed to the PR, so make sure you use the latest commit sha)

[...]

PR #13298, opened 2022-11-01
Fix scmsync tag creation for packages with query/fragment in the url

dcermak:
The SCM workflow creates bogus scmsync URLs when the branched package contains any query or fragment in its url, because the fragment is never removed, it is just appended. To fix this, we use the addressable module to override the fragment via the URI parser instead of doing this ourselves.

This fixes https://github.com/openSUSE/open-build-service/issues/13290

### DIFF
--- a/src/api/app/models/concerns/scm_sync_enabled_step.rb
+++ b/src/api/app/models/concerns/scm_sync_enabled_step.rb
@@ -2,12 +2,19 @@ module ScmSyncEnabledStep
   extend ActiveSupport::Concern
 
   def set_scmsync_on_target_package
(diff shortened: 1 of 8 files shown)


src/api/app/models/concerns/scm_sync_enabled_step.rb
```diff
@@ -1,13 +1,22 @@
+require 'addressable'
```
dmarcoux, line 1: This line isn't needed, it should be removed.

dcermak: Done

dmarcoux (approved):

dmarcoux:
Thank you @dcermak for submitting this PR!

PR #13303, opened 2022-11-02
Use request.approver during unstaging

hennevogel:
During unstaging we temporarily re-open declined request to be able to add a review for the staging manager group. We do this so the request will be in the staging workflow backlog automatically if it's reopened.

We need to run the re-open as approver, if set. Because the logged in user might have no permissions to decline the request because they lack write permissions on the target project.

See `BsRequestPermissionCheck.extra_permissions_check_changestate` which checks for this.

Fixes OBS-230


coolo:
There is very likely also no approver at this stage.

hennevogel:
> There is very likely also no approver at this stage.

Indeed...

PR #13306, opened 2022-11-02
Cover request controller beta show with specs

danidoni:
The beta side of the `Webui::RequestController#show` is not covered by tests. This PR tries to address that in order to have some safety net before attempting a refactor of the code present in that method.


src/api/spec/controllers/webui/request_controller/show_spec.rb
```diff
@@ -0,0 +1,438 @@
...
+
+  describe 'GET #show' do
+    context 'when the user is part of the beta program' do
+      render_views
+
+      context 'when the user has only one submit action' do
+        let!(:comment) { create(:comment, commentable: bs_request) }
+
+        before do
+          Flipper.enable(:request_show_redesign, receiver)
```
dmarcoux, line 26: Repeating this in every context isn't needed if you put this in a `before` block of the parent context `'when the user is part of the beta program'`.

src/api/spec/controllers/webui/request_controller/show_spec.rb
```diff
@@ -0,0 +1,438 @@
...
+
+        it 'shows the request comment' do
+          expect(response.body).to have_selector('.timeline-item', text: comment.body)
+        end
+
+        it { expect(response.body).to have_text("The RPM Lint results aren't here yet") }
+        it { expect(response.body).to have_text("The mentioned issues aren't here yet") }
+      end
+
+      # TODO: prepare one use case per action and check the dropdown shows the corresponding action
```
dmarcoux, line 49: Is this a TODO for this PR?

src/api/spec/controllers/webui/request_controller/show_spec.rb
danidoni: It's true. Changing...

src/api/spec/controllers/webui/request_controller/show_spec.rb
```diff
@@ -0,0 +1,415 @@
...
+      context 'when the user has only one submit action' do
+        let!(:comment) { create(:comment, commentable: bs_request) }
+
+        before do
+          login receiver
+          get :show, params: { number: bs_request.number }
+        end
+
+        it 'shows the new redesign page' do
+          expect(controller).to render_template('webui/request/beta_show')
```
hennevogel, line 35: What do you think will break? `if` or `Flipper`? I don't get the point of this example.

src/api/spec/controllers/webui/request_controller/show_spec.rb
```diff
@@ -0,0 +1,415 @@
...
+
+        it 'shows the request comment' do
+          expect(response.body).to have_selector('.timeline-item', text: comment.body)
+        end
+
+        it { expect(response.body).to have_text("The RPM Lint results aren't here yet") }
+        it { expect(response.body).to have_text("The mentioned issues aren't here yet") }
+      end
+
+      context 'when the request has multiple submit actions' do
```
hennevogel, line 52: Same here, which logic are you testing? `if submit_actions.count > 1`? I'm sure that one `if` breaks we have other problems than not rendered views ;)

hennevogel (changes requested):

src/api/spec/controllers/webui/request_controller/show_spec.rb
danidoni: This was a leftover from the refactoring. I'll remove it.

src/api/spec/controllers/webui/request_controller/show_spec.rb
```diff
@@ -0,0 +1,415 @@
...
+        end
+
+        it 'shows the new redesign page' do
+          expect(controller).to render_template('webui/request/beta_show')
+          expect(response.body).to have_text("Request #{bs_request.number}")
+        end
+
+        it 'shows the history elements for the request' do
+          expect(response.body).to have_text("Created by\n #{submitter.name}")
+          expect(response.body).to have_text("(#{bs_request.creator})\ncreated this request")
```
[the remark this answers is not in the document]
danidoni: Yes, but this is basically an integration test to prove that the refactoring I'm doing starting from #13347 does not break anything. 

This is indeed a view test in a controller test, that's intended. I agree that most of the testing code in this spec belongs to model specs or view component specs, but what I need to check now, is that the data I send from the controller flows downstream to all the partials and view components.

Once the refactoring is finished I can move those expectations where they belong.

src/api/spec/controllers/webui/request_controller/show_spec.rb
danidoni: See above, same logic I'm following here

danidoni: This is a canary test for the refactoring.

src/api/spec/controllers/webui/request_controller/show_spec.rb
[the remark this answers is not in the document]
danidoni: I've removed those test cases from the controller spec

src/api/spec/controllers/webui/request_controller/show_spec.rb
danidoni: The same here, I've removed those test cases from the controller specs

src/api/spec/controllers/webui/request_controller/show_spec.rb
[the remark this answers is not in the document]
danidoni: We'll add them to the view component tests

src/api/spec/controllers/webui/request_controller/show_spec.rb
```diff
@@ -0,0 +1,360 @@
...
+
+      context 'when the request has an add maintainer request action'
+      context 'when the request has an add bugowner request action'
+      context 'when the request has a submit request action'
+      context 'when the request has a delete request action'
+      context 'when the request has a maintenance incident request action'
+      context 'when the request has a maintenance release request action'
+      context 'when the request has a set bugowner request action'
+
+      context 'when having an accepted review' do
```
hennevogel, line 48: I wonder what the significance of this context is? Why do you test this?

src/api/spec/controllers/webui/request_controller/show_spec.rb
```diff
@@ -0,0 +1,360 @@
...
+  describe 'GET #show' do
+    context 'when the user is part of the beta program' do
+      render_views
+
+      before do
+        Flipper.enable(:request_show_redesign, receiver)
+      end
+
+      context 'when the user has only one submit action' do
+        let!(:comment) { create(:comment, commentable: bs_request) }
```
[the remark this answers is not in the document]
danidoni: It is, yes. Removing

src/api/spec/controllers/webui/request_controller/show_spec.rb
```diff
@@ -0,0 +1,360 @@
...
+
+        before do
+          login receiver
+          get :show, params: { number: bs_request.number }
+        end
+
+        it 'shows the new redesign page' do
+          expect(response.body).to have_text("Request #{bs_request.number}")
+          expect(response.body).to have_text('beta program')
```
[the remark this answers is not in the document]
danidoni: I'm gonna remove those template checking from all the spec, because that is testing a specific implementation and doing that makes our tests brittle. We should avoid doing that. But I need to check that those templates are used indirectly... I marked the PR as draft, because I'm changing stuff and it's not really "for review" right now...

src/api/spec/controllers/webui/request_controller/show_spec.rb
danidoni: I test this, because when there are accepted reviews, we render the accordion at the bottom. I want to test that the review accordion is working.

dmarcoux: The [AccordionReviewsComponent](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/components/accordion_reviews_component.rb) could be tested directly without having to test it here in controller specs. This will involve less setup in the specs and should be a tiny bit faster.

danidoni: Yes, and no. Exhaustive testing should go into the AccordionReviewsComponent, yes. But we need to test the integration of the component with the partials also (if we need  a param in the component and somehow the caller don't reflect that in the partial the component specs won't catch that kind of error). Testing the "happy path" is enough for this.

danidoni:
I'll close this PR. This was not the way to go with testing the features. We have #13926 for that.

PR #13310, opened 2022-11-03
Extract methods used by many tasks to a file

saraycp:
This is the first in a series of PRs refactoring the dev tasks. The plan is to move some tasks from `dev.rake` to different individual files. They will be placed in the `lib/tasks/dev` directory.

In this first PR, I extract methods that are going to be used by different tasks to a file we can require later.

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -2,11 +2,12 @@
 
 require 'fileutils'
 require 'yaml'
+require 'tasks/dev/rake_support'
 
 namespace :dev do
   task :prepare do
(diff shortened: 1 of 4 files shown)


src/api/lib/tasks/dev/helper_methods.rb
```diff
@@ -0,0 +1,23 @@
+def create_and_assign_project(project_name, user)
```
dmarcoux, line 1: We should put those methods inside a module, otherwise they are added to `Object` as private methods.

dmarcoux (changes requested):

dmarcoux (approved):

PR #13311, opened 2022-11-03
Use importmaps

dmarcoux:
Everything is detailed in the commit messages.


dmarcoux:
I need to put more thoughts into this.

PR #13319, opened 2022-11-03
Add debug mode to EventSubscription::FindForEvent#subscription

krauselukas:
We recently faced an issue where a user received notifications
for events without subscription. In order to identify the source/cause,
debugging output will help to follow the connections between Event,
Subscription and Receiver.

*Example output:*

```
Looking at Admin for 'commenter' and channel 'instant_email'
Skipped receiver Admin, since it is the originator of the event...
People to receive something: []

Looking at Admin for 'source_maintainer' and channel 'instant_email'
Skipped receiver Admin, since it is the originator of the event...
People to receive something: []

Looking at foo, bar for 'target_maintainer' and channel 'instant_email'
Expanding group bar...
Looking at foo, foo for 'target_maintainer' and channel 'instant_email'
Skipped receiver foo, since it is already in the list...
People to receive something: ["foo"]

Looking at Admin for 'creator' and channel 'instant_email'
Skipped receiver Admin, since it is the originator of the event...
People to receive something: ["foo"]

Looking at Admin, bar for 'reviewer' and channel 'instant_email'
Expanding group bar...
Looking at Admin, foo for 'reviewer' and channel 'instant_email'
Skipped receiver Admin, since it is the originator of the event...
Skipped receiver foo, since it is already in the list...
People to receive something: ["foo"]

```

### DIFF
--- a/src/api/app/models/event_subscription/find_for_event.rb
+++ b/src/api/app/models/event_subscription/find_for_event.rb
@@ -2,19 +2,23 @@ class EventSubscription
   class FindForEvent
     attr_reader :event
 
-    def initialize(event)
+    def initialize(event, debug: false)
       @event = event
+      @debug = debug
     end
 
+    # rubocop: disable Rails/Output
     def subscriptions(channel = :instant_email)
       receivers_and_subscriptions = {}
 
       event.class.receiver_roles.each do |receiver_role|
         # Find the users/groups who are receivers for this event
-        receivers = event.send("#{receiver_role}s")
-        next if receivers.blank?
+        receivers_before_expand = event.send("#{receiver_role}s")
+        next if receivers_before_expand.blank?
 
-        receivers = expand_receivers(receivers, channel)
+        puts "Looking at #{receivers_before_expand.map(&:to_s).join(', ')} for '#{receiver_role}' and channel '#{channel}'" if @debug
+        receivers = expand_receivers(receivers_before_expand, channel)
+        puts "Looking at #{receivers.map(&:to_s).join(', ')} for '#{receiver_role}' and channel '#{channel}'" if @debug && (receivers_before_expand - receivers).any?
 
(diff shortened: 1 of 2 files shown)


src/api/app/models/event_subscription/find_for_event.rb
```diff
@@ -6,34 +6,89 @@ def initialize(event)
       @event = event
     end
 
-    def subscriptions(channel = :instant_email)
+    def subscriptions(channel = :instant_email, debug: false)
+      if debug
+        # The logger output spoils the debug output. Therefore its silenced when running in debug mode.
```
hennevogel, line 11: This is only true in development mode in which you probably will never run this :)

src/api/app/models/event_subscription/find_for_event.rb
```diff
@@ -6,34 +6,89 @@ def initialize(event)
...
+      else
+        receive_subscriptions(channel)
+      end
+    end
+
+    private
+
+    # rubocop: disable Rails/Output
+    def receive_subscriptions(channel, debug: false)
+      @debug = debug
```
hennevogel, line 24: There is an initialize method to do this, like we do for event there.

src/api/app/models/event_subscription/find_for_event.rb
krauselukas: I removed the part where I silence the active record logger

krauselukas:
@hennevogel simplified the output, don't print the whole objects anymore.

krauselukas:
I would start with this and add more output in case its not sufficient to identify the issues we face

krauselukas:
[Add debug flag/mode to `EventSubscription::FindForEvent`](https://trello.com/c/MxIYK4Ze/2021-add-debug-flag-mode-to-eventsubscriptionfindforevent)

dmarcoux (approved):

PR #13339, opened 2022-11-07
[backend] support delaying of lookat elements of type high

mlschroe:
We only delayed med type elements before. Some projects generate a lot of high events over time, which led to the med event delay mechanism no longer being effective.


src/backend/BSSched/Lookat.pm
```diff
@@ -222,8 +223,28 @@ sub lookatprp {
...
 }
 
+sub setdelayed {
+  my ($gctx, $prp, $checktime) = @_;
+  if (!$checktime) {
+    delete $gctx->{'nextmed'}->{$prp};
+    delete $gctx->{'nexthigh'}->{$prp};
+    return;
+  }
+  my $medfactor = 1 + @{$gctx->{'lookat_med'}} / 100;
```
coolo, line 238: A comment explaining these magic numbers would be awesome

PR #13341, opened 2022-11-07
Add markdown preview to status message

saraycp:
The status messages accept markdown. We find it helpful to see a preview before we save it.

write_preview_status

**How to test in review-app?**

- Log in as Admin.
- Go to the OBS home page.
- Click on "Create Status Message"
- Add some text applying markdown.
- Click on the _Preview_ tab and check if the markdown is applied correctly.
- Save the status message and check the message is displaying, with the correct markdown applied, in the News section.


dmarcoux:
@saraycp You'll have to restart the `rspec` check in your CircleCI fork. I cannot do it. Using your own CircleCI fork shouldn't be needed since we never run out of credits on our CircleCI account.

PR #13343, opened 2022-11-07
Enable Style/StringConcatenation cop using unsafe autocorrection

krauselukas:
DRAFT: There are more places where manual adaptions are required.


krauselukas:
Talked about this with @dmarcoux going to split this up into smaller chunks

PR #13346, opened 2022-11-08
Enable `Style/StringConcatenation` cop in rubocop main config

krauselukas:
Exclude files that offend the Style/StringConcatenation` cop in the main conf in order to tackle them one by one, since we can set the exclude-limit only globally.

I would like to tackle this cop in a series of PR's, instead of a big chunk. See https://github.com/openSUSE/open-build-service/pull/13343 which is hard to review. The problem is I can only set the exclude limit for all cops. Since I don't want to enable all other currently disabled cops, I would exclude the affected files for this cop in the main conf and then remove them one by one.


krauselukas:
[Reduce the amount of TODOs for rubocop](https://trello.com/c/iwkSTxjg/2022-reduce-the-amount-of-todos-for-rubocop)

src/api/.rubocop.yml
```diff
@@ -224,6 +224,154 @@ Style/StringLiterals:
     # due to changes from single to double quotes... we have better things to do!
     - 'config/**/*'
 
+# Exclude files for this cop here since we want to tackle them one by one. Our exclusion limit is to low, so
```
dmarcoux, line 227:
```suggestion
# Exclude files for this cop here since we want to tackle them one by one. Our exclusion limit is too low, so
```

krauselukas: thanks, done :)

dmarcoux (approved):

PR #13347, opened 2022-11-08
Use local variables whenever possible

danidoni:
This is a follow up of https://github.com/openSUSE/open-build-service/pull/13306

A variable is a bit of complexity added. Instance variables acts as
global variables, and global variables add global complexity.

Using local variables contains that complexity inside the scope of the
local variable, that is, the function being called.

See https://softwareengineering.stackexchange.com/a/318483

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -13,32 +13,32 @@ class Webui::RequestController < Webui::WebuiController
   def show
     # TODO: Remove this `if` condition, and the `else` clause once request_show_redesign is rolled out
     if Flipper.enabled?(:request_show_redesign, User.session)
-      @is_author = @bs_request.creator == User.possibly_nobody.login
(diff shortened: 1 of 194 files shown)


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -49,7 +49,27 @@ def show
       end
 
       # TODO: Remove this `render` line once request_show_redesign is rolled out
-      render :beta_show
+      respond_to do |format|
+        format.html do
+          render 'beta_show', locals: {
```
hennevogel, line 54: Do you see anyone else using render+locals in a controller in this app? I'm not quite sure this adds clarity...

dmarcoux: It goes against the Rails convention. Top-level views always use instance variables, then they pass those as local variables to partials.

danidoni:
I'm gonna try some other approatch

PR #13348, opened 2022-11-08
Avoid failing packtrack jobs

adrianschroeter:
Due to large amount of content in single registry release repositories.

### DIFF
--- a/src/api/app/jobs/update_released_binaries_job.rb
+++ b/src/api/app/jobs/update_released_binaries_job.rb
@@ -9,4 +9,9 @@ class UpdateReleasedBinariesJob < CreateJob
 
     BinaryRelease.update_binary_releases(repo, pl['payload'], event.created_at)
   end
+
+  def max_run_time
+    # twice the time of the default (8 hours now)
+    28800
+  end
 end


adrianschroeter:
just the timeout increase for now, the parallel processing needs some changes in the transaction grouping

src/api/config/initializers/delayed_job_config.rb
```diff
@@ -0,0 +1,4 @@
+# esp large release tracking jobs can take a long time
+# twice the time of the default (8 hours now)
+Delayed::Worker.max_run_time = 28800
```
dmarcoux, line 3: This will affect all jobs, which we should avoid. If something goes wrong in a job and somehow the worker is stuck, it will taking even longer before it's finally stopped. Instead, we need to change the timeout of the jobs which are needing more time.

If we use the approach from this file, we should lower the timeout of all jobs which don't require that much time. See the [delayed_job README](https://github.com/collectiveidea/delayed_job#custom-jobs) on how to do this with `max_run_time`. Another less invasive approach would be to apply what is proposed [here](https://stackoverflow.com/questions/26179508/timeout-in-a-delayed-job). This would only be in the jobs which really need more time, the remaining jobs are unchanged.

dmarcoux (changes requested):

src/api/app/jobs/update_released_binaries_job.rb
```diff
@@ -9,4 +9,9 @@ def perform(event_id)
 
     BinaryRelease.update_binary_releases(repo, pl['payload'], event.created_at)
   end
+
+  def max_run_time
+    # twice the time of the default (8 hours now)
+    28800
```
dmarcoux, line 15: This won't work by itself as I have written in my previous comment. The delayed_job README explains why:

> NOTE: this can ONLY be used to set a max_run_time that is lower than Delayed::Worker.max_run_time. Otherwise the lock on the job would expire and another worker would start the working on the in progress job.

If we use this approach, we need to increase the timeout of all jobs while lowering the timeout of all jobs which don't require that much time. A better approach is the second option I proposed in my previous comment:

> (...) apply what is proposed [here](https://stackoverflow.com/questions/26179508/timeout-in-a-delayed-job). This would only be in the jobs which really need more time, the remaining jobs are unchanged.

adrianschroeter:
okay, I do not have time for it atm and I configured the affected instance manually for now

PR #13350, opened 2022-11-08
Push some request controller code to bs request model

danidoni:
This is a follow-up of https://github.com/openSUSE/open-build-service/pull/13347

### DIFF
--- a/src/api/app/models/bs_request.rb
+++ b/src/api/app/models/bs_request.rb
@@ -1001,6 +1001,14 @@ class BsRequest < ApplicationRecord
     !staging_project_id.nil?
   end
 
+  def is_author?(user)
+    creator == user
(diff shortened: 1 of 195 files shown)


src/api/app/models/bs_request.rb
```diff
@@ -1001,6 +1001,14 @@ def staged_request?
     !staging_project_id.nil?
   end
 
+  def is_author?(user)
+    creator == user
+  end
+
+  def can_add_reviews?(is_author, is_target_maintainer, my_open_reviews)
```
dmarcoux, line 1008: This sounds more like a Pundit policy.

danidoni:
I'll close this as I believe we can address the complexity easier once we roll out the feature

PR #13351, opened 2022-11-08
Move code closer to where it will be used

danidoni:
This is a follow-up of https://github.com/openSUSE/open-build-service/pull/13350

We don't really need to calculate those variables in the controller just to push them down over several partials.

Bringing that variable creation closer to where it will be used will reduce the likelihood of interfering with future changes.

As a side note, the creation of those variables in the _conversation partial hints at the creation of a View Component to hold the
conversation view.


danidoni:
Like in #13350 I'll close this as I believe we can address the complexity easier once we roll out the feature

PR #13352, opened 2022-11-08
Simplify the request decision component

danidoni:
This is a follow-up of https://github.com/openSUSE/open-build-service/pull/13351


danidoni:
Like in #13350 I'll close this as I believe we can address the complexity easier once we roll out the feature

PR #13353, opened 2022-11-08
Make the request controller show action simpler

danidoni:
This is a follow-up of https://github.com/openSUSE/open-build-service/pull/13352


danidoni:
I'm not sure about this refactoring anymore...

PR #13360, opened 2022-11-09
Improve new build results view

saraycp:
Propose another design for the new build results tab including
- bigger and soberer help icon,
- many results per row in large viewports and minimized content in small ones,
- more descriptive tooltip,

among many other improvements.

build_results_redesign

TODO:

- [x] Component specs
- [x] Component preview

**How to test?**

If you already ran the `development_testdata` rake task you'll have a few requests already, you can see them in the Tasks menu.
Add some repositories to the project related to the request you choose. 
Then go to see the build results. For example, request for `home:Admin/package_b`.

In review-app:
https://obs-reviewlab.opensuse.org/saraycp-improve_new_build_results/request/show/9/request_action/9#tab-pane-build-results

### DIFF
--- a/src/api/app/assets/stylesheets/request_show_redesign/build_results.scss
+++ b/src/api/app/assets/stylesheets/request_show_redesign/build_results.scss
@@ -1,29 +1,9 @@
 .build-result {
-  // One element per row
-  // The div width is 100% minus the sum of its margins: ml-2 + mr-2 = 0.5rem + 0.5rem = 1rem
-  width: calc(100% - 1rem);
+  width: 220px;
   border-left: thick solid $gray-300;
   cursor: help;
 
   &:hover {
     background-color: $gray-100;
-  }
-
-  @include media-breakpoint-up(sm) {
-    // Up to 2 elements in a row
-    // The div width is 50% minus the sum of its margins: ml-2 + mr-5 = 0.5rem + 3rem = 3.5rem
-    width: calc(50% - 3.5rem);
-  }
-
-  @include media-breakpoint-up(lg) {
-    // Up to 4 elements in a row
-    // The div width is 25% minus the sum of its margins: ml-2 + mr-5 = 0.5rem + 3rem = 3.5rem
-    width: calc(25% - 3.5rem);
-  }
-
-  @include media-breakpoint-up(xxl) {
-    // Up to 5 elements in a row
-    // The div width is 20% minus the sum of its margins: ml-2 + mr-5 = 0.5rem + 3rem = 3.5rem
-    width: calc(20% - 3.5rem);
   }
 }
--- a/src/api/app/assets/stylesheets/request_show_redesign/build_results.scss
+++ b/src/api/app/assets/stylesheets/request_show_redesign/build_results.scss
@@ -1,0 +1,29 @@
+.build-result {
+  // One element per row
+  // The div width is 100% minus the sum of its margins: ml-2 + mr-2 = 0.5rem + 0.5rem = 1rem
+  width: calc(100% - 1rem);
+  border-left: thick solid $gray-300;
+  cursor: help;
+
+  &:hover {
+    background-color: $gray-100;
+  }
+
+  @include media-breakpoint-up(sm) {
+    // Up to 2 elements in a row
+    // The div width is 50% minus the sum of its margins: ml-2 + mr-5 = 0.5rem + 3rem = 3.5rem
+    width: calc(50% - 3.5rem);
+  }
+
+  @include media-breakpoint-up(lg) {
+    // Up to 4 elements in a row
+    // The div width is 25% minus the sum of its margins: ml-2 + mr-5 = 0.5rem + 3rem = 3.5rem
+    width: calc(25% - 3.5rem);
+  }
+
+  @include media-breakpoint-up(xxl) {
+    // Up to 5 elements in a row
+    // The div width is 20% minus the sum of its margins: ml-2 + mr-5 = 0.5rem + 3rem = 3.5rem
+    width: calc(20% - 3.5rem);
+  }
+}
--- a/src/api/app/components/build_result_for_architecture_component.rb
+++ b/src/api/app/components/build_result_for_architecture_component.rb
@@ -1,0 +1,68 @@
+class BuildResultForArchitectureComponent < ApplicationComponent
+  attr_accessor :result, :project, :package
+
+  def initialize(result, project, package)
+    super
+
+    @result = result
+    @project = project
+    @package = package
+  end
+
+  private
+
+  def result_code
+    result.code
+  end
+
+  def help
+    help = "<p><strong>Package build ( #{build_status_icon} #{result_code} ):</strong> #{build_status_help}</p>"
+    help += "<p><u>Details</u>: #{build_status_details}</p>" if build_status_details.present?
+    help += "<p><strong>Repository status ( #{repository_status_icon} #{result.state} ): </strong>#{repository_status_help}</p>"
+    help
+  end
+
+  def build_status_help
+    Buildresult.status_description(result_code)
+  end
+
+  def build_status_details
+    result.details
+  end
+
+  def repository_status_help
+    result.is_repository_in_db ? helpers.repository_info(result.state) : 'This result is outdated'
+  end
+
+  def build_status_icon_with_link
+    return build_status_icon if result_code.in?(['-', 'unresolvable', 'blocked', 'excluded', 'scheduled'])
+
+    link_to(package_live_build_log_path(project: project.to_s, package: package,
+                                        repository: result.repository, arch: result.architecture), rel: 'nofollow') do
+      capture { build_status_icon }
+    end
+  end
+
+  def build_status_icon
+    tag.i(class: "fa #{helpers.build_status_icon(result_code)} text-gray-500")
+  end
+
+  def repository_status_icon
+    helpers.repository_status_icon(status: result.state)
+  end
+
+  def status_border_color(status)
+    build_result = Buildresult.new(status)
+    return 'border-warning' if build_result.in_progress_status?
+    return 'border-success' if build_result.successful_final_status?
+    return 'border-danger' if build_result.unsuccessful_final_status?
+    return 'border-gray-300' if build_result.refused_status?
+  end
+
+  def build_status
+    return result_code if result_code.in?(['-', 'unresolvable', 'blocked', 'excluded', 'scheduled'])
+
+    link_to(result_code, package_live_build_log_path(project: project.to_s, package: package,
+                                                     repository: result.repository, arch: result.architecture), rel: 'nofollow')
+  end
+end
--- a/src/api/app/models/buildresult.rb
+++ b/src/api/app/models/buildresult.rb
@@ -1,4 +1,6 @@
 class Buildresult
+  attr_accessor :status
+
   AVAIL_STATUS_VALUES = {
     succeeded: 0,
     failed: 1,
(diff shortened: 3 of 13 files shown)


saraycp:
The help icon is now bigger to make it touchable. Compare these two:

Screenshot 2022-11-10 at 10-37-55 Open Build Service

src/api/app/views/webui/request/beta_show_tabs/_build_status.html.haml
```diff
@@ -17,36 +17,37 @@
...
-                  = result_help
-                %i.fa.fa-question-circle.d-lg-none.text-info{ data: { placement: 'top', toggle: 'popover', content: "#{result_help}" } }
-              .col-3
-                - if result.code.in?(['-', 'unresolvable', 'blocked', 'excluded', 'scheduled'])
-                  %i{ class: "fa #{build_status_icon(result.code)}" }
-                - else
-                  = link_to(package_live_build_log_path(project: project.to_s, package: package,
-                      repository: result.repository, arch: result.architecture), rel: 'nofollow') do
+            :ruby
+              result_help = repository_info(result.state) if result.is_repository_in_db
```
danidoni, line 23: I belive we should move this into a View Component, the logic would be pushed into the component class methods/instance variables and the view would be easier. And you get easier testing by the same price :moneybag:

dmarcoux:
Comparing the two designs you shared in the PR description, the difference now is that the accordion is gone. It's still somewhat boxy. Do we really need those boxes around the build results for every architecture? I don't think so. Displaying a separator (as a right/bottom border) between build results should be enough. It's less heavy on the eyes. The elements also don't have to fill up the available space (so turning the class `flex-fill` into `flex` would be a good idea).

Before:
boxy
boxy-mobile

My proposal:
less-boxy
less-boxy-mobile

saraycp:
> Displaying a separator

I'll give it a try. Thanks!

saraycp:
> is also position independent, the border being always on the left

I like the idea of having it only on the left.

Let's see what I can do with the colors... we have 15 different status but we don't have a different color for each of them.

saraycp:
> Let's see what I can do with the colors... we have 15 different status but we don't have a different color for each of them.

@hellcp-work, this is what I meant before. Here you have all the possible statuses:

Screenshot 2022-11-10 at 12-43-25 Open Build Service

That _failed_, _unresolvable_ and _broken_ are all red is okay because they are considered final statuses, among others.
I have to find the right colors for the rest.

src/api/app/views/webui/request/beta_show_tabs/_build_status.html.haml
saraycp: I would say the whole build results element is going to be moved to a component...
I'd like to do that as soon as I finish with the UI improvements.

saraycp: Maybe in another PR.

saraycp:
> You could go for just four or five colours, indicating: success, failure, excluded and in progress/waiting (not sure if waiting is important enough to signify by itself)

That's more or less what I did. Please check the commit "_Set status color on a left border_", @hellcp-work

I split the statues in three groups: `in_progress_status` (yellow), `successful_final_status` (green), and `unsuccessful_final_status` (red).

This is what they look like so far:

Screenshot 2022-11-10 at 13-59-12 Open Build Service

Now I'm going to re-think the style in the status text. I could let them be gray and leave the color to the border only.

src/api/app/models/buildresult.rb
```diff
@@ -62,6 +62,18 @@ def self.final_status?(status)
     status.in?(['succeeded', 'failed', 'unresolvable', 'broken', 'disabled', 'excluded'])
   end
 
+  def self.successful_final_status?(status)
+    status.in?(['succeeded'])
+  end
+
+  def self.unsuccessful_final_status?(status)
+    status.in?(['failed', 'unresolvable', 'broken', 'disabled', 'excluded'])
```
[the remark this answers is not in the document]
saraycp: Those two are now light gray, meaning they are not building.
Please check commit "_Set gray border to disabled and excluded results_".

Screenshot 2022-11-10 at 15-14-12 Open Build Service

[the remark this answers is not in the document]
saraycp: This is what it looks like next to the orders, at a glance it's clear they are disabled/excluded.

Screenshot 2022-11-10 at 15-28-46 Open Build Service

saraycp:
I've pushed commit "_WIP: Remove color from status text_" where I remove the color from the status text and unify the icon and the border color.

Screenshot 2022-11-10 at 16-13-32 Open Build Service

src/api/app/components/build_status_for_architecture_component.html.haml
```diff
@@ -0,0 +1,31 @@
...
+            repository: result.repository, arch: result.architecture), rel: 'nofollow') do
+          %i{ class: "fa #{helpers.build_status_icon(result_code)} text-gray-500" }
+      %span
+        -# TODO: rework arch_repo_table_cell if we finally remove the colors from the status text.
+        = result_code
+        =# helpers.arch_repo_table_cell(result.repository, result.architecture, package, 'code' => result_code, 'details' => result.details)
+    .repository-status
+      = helpers.repository_status_icon(status: result.state)
+      %span
+        = result.state
```
ncounter, line 31: What about dropping this text and moving it as an icon tooltip text only?
The status is already reported in the help box and maybe it is not so relevant to display at the first glance. The icon giving an initial clue about its state could be enough.
By that it would get less verbose and every "box" can be shrunk a bit.

saraycp: In such case, I'll move the icon in front of the architecture:

Screenshot 2022-11-11 at 16-43-37 Open Build Service

Otherwise it will look like this:
Screenshot 2022-11-11 at 16-39-49 Open Build Service

saraycp: ... what makes me think that we don't need two lines for that...

saraycp:
This is what this looks like so far:

Small viewports:

Request-9-Submit-package_a-Open-Build-Service(3)

Large viewports while hovering:

hovering


Thanks @ncounter for your great ideas.

src/api/app/assets/stylesheets/request_show_redesign/build_results.scss
```diff
@@ -0,0 +1,21 @@
...
+
+    @include media-breakpoint-up(sm) {
+      width: calc(50% - 3.5rem);
+    }
+
+    @include media-breakpoint-up(lg) {
+      width: calc(25% - 3.5rem);
+    }
+
+    border-left: thick solid $gray-300;
```
ncounter, line 14: I guess this got duplicated on the way from line 4 above.

saraycp:
> I would argue that failed status is more important than succeeded for the purpose of review, so it should probably be the first thing the reviewer sees.

@hellcp-work My intention was to display the same colors together. The order given by the constant allows that more or less.
But I see your point, what about deciding the right order and implementing it in another PR?

src/api/app/assets/stylesheets/request_show_redesign/build_results.scss
```diff
@@ -0,0 +1,23 @@
...
+    width: calc(100% - 1rem);
+    border-left: thick solid $gray-300;
+    cursor: help;
+
+    &:hover {
+      background-color: $gray-100;
+    }
+
+    @include media-breakpoint-up(sm) {
+      width: calc(50% - 3.5rem);
```
dmarcoux, line 12: Where is `3.5.rem` coming from?

saraycp: The addition of the left and right margins applied to the result box.

dmarcoux: A comment or even better, some Sass variables would make this obvious. Otherwise, it looks like a magic number and in a few months from now, we don't know anymore what this number refers to.

saraycp: I added some comments

saraycp:
This is ready for review now. I squashed many of the commits, updated the PR description, and attached a gift showing the final version.

src/api/app/components/build_result_for_architecture_component.rb
```diff
@@ -0,0 +1,67 @@
...
+  def build_status_icon
+    tag.i(class: "fa #{helpers.build_status_icon(result_code)} text-gray-500")
+  end
+
+  def repository_status_icon
+    helpers.repository_status_icon(status: result.state)
+  end
+
+  def status_border_color(status)
+    return 'border-warning' if Buildresult.in_progress_status?(status)
```
danidoni, line 55: I believe you can initialize BuildResult with the status and then query that buildresult about it's several states, instead of feeding the status every time. You know:
```
buildresult = BuildResult.new(status)
buildresult.in_progress?
buildresult.successful?
etc
```
Otherwise, there is no reason to use a class, just a `BuildResult` module with methods inside.

saraycp: I've just pushed the changes

saraycp: I added a card for that: https://trello.com/c/6cqcj2DI/54-replace-class-methods-by-instance-methods-in-model-buildresult

hennevogel:
@saraycp build results don't work at all anymore now. Neither in `Package#show` nor in `BsRequest#show`

https://obs-reviewlab.opensuse.org/saraycp-improve_new_build_results/package/show/home:Admin/ctris

saraycp:
> @saraycp build results don't work at all anymore now. Neither in `Package#show` nor in `BsRequest#show`
> 
> https://obs-reviewlab.opensuse.org/saraycp-improve_new_build_results/package/show/home:Admin/ctris

The project `home:Admin` didn't have non-broken(*) repositories set. I set a few now, so you can see results in:

https://obs-reviewlab.opensuse.org/saraycp-improve_new_build_results/package/show/home:Admin/ctris
https://obs-reviewlab.opensuse.org/saraycp-improve_new_build_results/request/show/9#tab-pane-build-results


(*) I created a card in Free Style with the details:
https://trello.com/c/buZQe98N/2834-make-a-repository-work-when-set-in-dev-data

src/api/app/components/build_result_for_architecture_component.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  private
+
+  def result_code
+    result.code
+  end
+
+  def help
+    help = "<p><strong>Package build ( #{build_status_icon} #{result_code} ):</strong> #{build_status_help}</p>"
+    help += "<p><u>Details</u>: #{build_status_details}</p>" if build_status_details.present?
```
ncounter, line 20: What if we keep the `build_status_details` closer to the `build_status_help` part as they are related one each other? Additionally I'd uniform the `strong` style for the label instead of the `underline` there. Something like the following
image

saraycp: We need to find a better place and way to display the details, which can be long text.
We could even split the static description and the details which are specific.

I create a card for that: https://trello.com/c/jJgasnGs/57-separate-static-description-of-the-status-and-the-details

ncounter: Thanks for taking care of that. We will tackle this in detail later, no reason to be picky and stop the rest here. :+1:

saraycp:
> Single architecture repos look a bit weird


@hennevogel, @ncounter and I worked to improve the view when there are only a few architectures per repository. We created a new commit on top of the rest.
Please have another look. This is the result:


float-left-results

src/api/spec/components/build_result_for_architecture_component_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'rails_helper'
+
+RSpec.describe BuildResultForArchitectureComponent, type: :component do
+  # let(:user) { create(:confirmed_user, :with_home, login: 'tome') }
```
dmarcoux, line 4: Should we delete the comments here and below?

dmarcoux (approved):
Awesome work!

PR #13361, opened 2022-11-10
Replace `arch_repo_table_cell` helper method with a view component

krauselukas:
This is what the component is rendering:


krauselukas:
[Stop using `arch_repo_table_cell`, rely on a view component (M)](https://trello.com/c/r0kxU7ij/2041-stop-using-archrepotablecell-rely-on-a-view-component-m)

src/api/app/helpers/webui/buildresult_helper.rb
```diff
@@ -16,33 +16,6 @@ module Webui::BuildresultHelper
...
-      css_class = "build-state-#{code}"
-      # special case for scheduled jobs with constraints limiting the workers a lot
-      css_class = 'text-warning' if code == 'scheduled' && link_title.present?
-    end
-
-    capture do
-      if code.in?(['-', 'unresolvable', 'blocked', 'excluded', 'scheduled'])
-        concat(tag.span(code, id: status_id, class: "#{css_class} toggle-build-info", title: 'Click to keep it open'))
-      else
-        concat(link_to(code.gsub(/\s/, '&nbsp;'),
```
krauselukas, line 39: `code.gsub(/\s/, '&nbsp;'` this part is not needed, guess its a left over from when it was used for projects as well where we have `code` like `Succeeded: 1`, but I couldn't find any case for packages

krauselukas: so I don't do this in the view component

src/api/app/components/buildresult_status_link_component.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def css_class
+    # special case for scheduled jobs with constraints limiting the workers a lot
+    return 'text-warning' if @build_status == 'scheduled' && @build_details.present?
+
+    "build-state-#{@build_status}" if @build_status.present?
+  end
+
+  def span_id
+    "id-#{@package_name}_#{@repository}_#{@architecture}"
```
krauselukas, line 25: I didn't transferred the usage of `valid_xml_id` from the helper, I think this is not required since the escaped character are not allowed anyway in the package name etc.

rubhanazeem (approved):

src/api/app/components/buildresult_status_link_component.rb
```diff
@@ -0,0 +1,27 @@
+class BuildresultStatusLinkComponent < ApplicationComponent
+  def initialize(repository:, architecture:, project:, package_name:, build_status:, build_details:)
+    super
+
+    @repository = repository
```
dmarcoux, line 5: Are `repository`, architecture`, `project` and `package_name` all names? If so, they should all end with `_name` to be consistent. It would be clearer what is expected to be passed into this view component.

krauselukas: right those are not objects, I changed the naming @dmarcoux

src/api/app/views/webui/package/_buildstatus.html.haml
```diff
@@ -40,7 +40,9 @@
                 = result.architecture
             .build-state.toggle-build-info-parent
               %i.fa.fa-question-circle.text-info.px-2.pl-lg-1.toggle-build-info{ title: 'Click to keep it open' }
-              = arch_repo_table_cell(result.repository, result.architecture, package, 'code' => result.code, 'details' => result.details)
+              = render(BuildresultStatusLinkComponent.new(repository_name: result.repository, architecture_name: result.architecture,
+                                                          project_name: project.name, package_name: package,
+                                                          build_status: result.code, build_details: result.details))
```
saraycp, line 45: Inside the component, `build_details` is only used to check if the details are present. [See line 19 of buildresult_status_link_component.rb](https://github.com/openSUSE/open-build-service/pull/13361/files#diff-8c4bdda2952f00c540f46af6859769c095165f9f696de5bfdb65a25d9024cc34R19).

What about passing a boolean instead?

Something like

```
= render(BuildresultStatusLinkComponent.new(repository: result.repository, architecture: result.architecture,
                                                              project: project, package_name: package,
                                                              build_status: result.code, contain_build_details: result.details.present?))
```

dmarcoux: I would be against this change, since this logic will have to be duplicated whenever we call this component. If the logic somehow changes, it needs to be adapted in multiple places instead of only one.

krauselukas:
pushed again, just fixed the spelling in the spec

src/api/app/views/webui/package/_buildstatus.html.haml
saraycp: Alright

saraycp (approved):

PR #13364, opened 2022-11-11
Fill RPM lint tab

eduardoj:
Peek 2022-11-15 19-44_request_workflow_redesign_fill_rpm_lint_tab_II

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -579,8 +579,12 @@ class Webui::PackageController < Webui::WebuiController
     if @repo_list.empty?
       render partial: 'no_repositories', locals: { project: @project }
     else
-      render partial: 'rpmlint_result', locals: { index: params[:index], project: @project, package: @package,
-                                                  repository_list: @repo_list, repo_arch_hash: @repo_arch_hash }
+      # TODO: this is part of the temporary changes done for 'request_show_redesign'.
+      request_show_redesign_partial = 'webui/request/beta_show_tabs/rpm_lint_result' if params.fetch(:inRequestShowRedesign, false)
+
+      render partial: (request_show_redesign_partial || 'rpmlint_result'), locals: { index: params[:index], project: @project, package: @package,
+                                                                                     repository_list: @repo_list, repo_arch_hash: @repo_arch_hash,
+                                                                                     is_staged_request: params[:is_staged_request] }
     end
   end
 
--- a/src/api/app/views/webui/request/beta_show_tabs/_rpm_lint.html.haml
+++ b/src/api/app/views/webui/request/beta_show_tabs/_rpm_lint.html.haml
@@ -6,11 +6,11 @@
     ajax_data['is_staged_request'] = h(is_staged_request) if is_staged_request
 
   .rpm-lint-content{ data: ajax_data }
-    .pt-1.pb-3.bg-white.d-flex.align-items-start.justify-content-between.sticky-top
+    .pt-1.bg-white.d-flex.flex-wrap-reverse.align-items-baseline.sticky-top
       .result
-      .btn.btn-outline-primary.rpm-lint-refresh{ onclick: 'updateRpmLintResultBeta()',
-                                                 accesskey: 'r', title: 'Refresh Rpm Lint Results',
-                                                 data: { rpm_lint_results_url: rpmlint_result_path } }
+      .btn.btn-outline-primary.rpm-lint-refresh.ml-auto.mb-3{ onclick: 'updateRpmLintResultBeta()',
+                                                              accesskey: 'r', title: 'Refresh Rpm Lint Results',
+                                                              data: { rpm_lint_results_url: rpmlint_result_path } }
         Refresh
         %i.fas.fa-sync-alt{ id: 'rpm-lint-reload' }
     -# For a request staged in a staging project, we display the Build Results / RPM Lint from the staging project instead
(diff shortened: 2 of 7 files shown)


eduardoj:
> Would it be possible to remove the second scroll inside of the tab? It doesn't really make sense in this context

On it. :+1:

dmarcoux:
I guess testing this in the review isn't working since it was removed from the PR, so what is a good way to test this locally?

dmarcoux:
Would it be possible to put the two dropdowns and the `Refresh` button on the same line?

refresh-button

eduardoj:
> Would it be possible to put the two dropdowns and the `Refresh` button on the same line? All the controls would be in the same location.

Sure. On it since yesterday.

eduardoj:
@hellcp-work, @dmarcoux I've addressed your suggestions. Please take a look again.

src/api/app/views/webui/package/_rpmlint_result.html.haml
```diff
@@ -8,7 +8,9 @@
       - repo_arch_hash.each do |repository, architectures|
         = select_tag("rpmlint_arch_select_#{index}_#{repository}", options_for_select(architectures.reverse),
                      class: "rpmlint_arch_select_#{index} custom-select", onchange: "updateRpmlintDisplay('#{index}')")
-%pre.rpmlint-result.text-pre-wrap{ id: "rpmlint_display_#{index}" }
+-# haml-lint:disable InlineStyles
+%pre.rpmlint-result.text-pre-wrap{ id: "rpmlint_display_#{index}", style: 'max-height: 35rem;' }
```
dmarcoux, line 12: This inline style isn't needed if we have a more precise CSS rule for `.rpmlint-result` depending on the view where it's used. In the request show redesign, we could instead rely on `#rpmlint-log .rpmlint-result`.

eduardoj: I thought we would get rid of this partial soonish. But it is not the case.

I will adjust the CSS rules.

eduardoj: SCSS adjusted. Please have a look.

src/api/app/views/webui/request/beta_show_tabs/_rpm_lint_result.html.haml
```diff
@@ -1,6 +1,11 @@
...
-  The RPM Lint results aren't here yet. For now, you can see them on the request's
-  source package page. We're working on having them here in the future.
+#rpmlint-log{ data: { project: project.name, package: package.name } }
+  - unless repository_list.empty?
+    .form-inline
+      = select_tag('rpmlint_repo_select', options_for_select(repository_list.sort),
+                   class: 'custom-select', onchange: 'updateRpmLintArchitectures()')
+      - repo_arch_hash.each do |repository, architectures|
+        = select_tag("rpmlint_arch_select_#{repository}", options_for_select(architectures.reverse),
+                     class: 'rpmlint_arch_select custom-select', onchange: 'updateRpmlintDisplay()')
```
saraycp, line 8: If I understand correctly, you moved JS functions related to RPM Lint to an independent file and renamed the functions... so this should be changed:

```suggestion
                     class: 'rpmlint_arch_select custom-select', onchange: 'updateRpmLintLog()')
```

saraycp (changes requested):

src/api/app/views/webui/request/beta_show_tabs/_rpm_lint_result.html.haml
eduardoj: Very true. Suggestion applied.Thanks!

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -579,8 +579,12 @@ def rpmlint_result
     if @repo_list.empty?
       render partial: 'no_repositories', locals: { project: @project }
     else
-      render partial: 'rpmlint_result', locals: { index: params[:index], project: @project, package: @package,
-                                                  repository_list: @repo_list, repo_arch_hash: @repo_arch_hash }
+      # TODO: this is part of the temporary changes done for 'request_show_redesign'.
```
danidoni, line 582: Do we need to refactor this later? If so, create a card, please.

src/api/app/views/webui/request/beta_show_tabs/_rpm_lint.html.haml
```diff
@@ -0,0 +1,29 @@
+- if package
+  :ruby
+    ajax_data = {}
+    ajax_data['project'] = h(project) if project
+    ajax_data['package'] = h(package) if package
+    ajax_data['is_staged_request'] = h(is_staged_request) if is_staged_request
```
danidoni, line 6: This looks like could be a View Component

src/api/app/controllers/webui/package_controller.rb
danidoni: Talked with @eduardoj offline about this. We're gonna sweep all the TODO's regarding the 'requet_show_redesign' at once, later, in one single card. No need to create a card to track this down.

eduardoj:
> We could adjust the buttons to fit better the small viewports:

@saraycp, I modified the margins and some other properties of the buttons. Please take a look again.

Screenshot from 2022-11-17 19-57-23

saraycp (approved):

src/api/app/views/webui/request/beta_show_tabs/_rpm_lint.html.haml
eduardoj: I don't think it is strictly necessary to do these changes in a View Component in this pull request. But it could be done in a follow-up pull request.

danidoni: Fine for me

PR #13365, opened 2022-11-11
Remove index on updated_at column in packages and projects tables

adrianschroeter:
This index is super harmful as it allows only one package operation at the same time and generating this index can take a long time. This leads to timeout errors on package operations easily.

I am not aware where this index would be useful atm.

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -1440,7 +1440,7 @@ end
 #  title           :string(255)
 #  url             :string(255)
 #  created_at      :datetime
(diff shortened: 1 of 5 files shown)


adrianschroeter:
DO_NOT_MERGE atm, to check if we run into any regression because of it in the next days.

But atm it only solved the timeout errors when checkin sources.

eduardoj:
Indexes for fields `updated_at` in projects and packages were introduced in 24eb95805205d0918b2ff356de05ce03751ba961.

eduardoj (approved):

PR #13366, opened 2022-11-11
Rewrite the ternary so it's easier on the eyes

danidoni:
Ternaries are harder to read than regular conditionals (if's). This one was extra hard because it was embedded in a string interpolation.

I found useful to test a private method this time


src/api/spec/models/bs_request_action_spec.rb
```diff
@@ -360,4 +360,38 @@
...
+    end
+    let(:bs_request) { create(:bs_request_with_submit_action, creator: user, target_package: target_package, source_package: source_package) }
+    let(:bs_request_action) { bs_request.bs_request_actions.first }
+
+    context 'when there is a problem with the backend' do
+      context 'and we have a revision' do
+        before { bs_request_action.source_rev = '1234' }
+
+        it 'returns an error message' do
+          mocked_backend = class_double('Backend::Api::Sources::Package')
```
dmarcoux, line 380: Mocking the backend call should be extracted in a `before` block under the `'when there is a problem with the backend'` context. This way, it doesn't have to be mocked in two examples.

eduardoj (approved):

PR #13379, opened 2022-11-15
Introduces obs-api-analytics sub-package

hennevogel:
To generate analytics with goaccess for today so far, last 7 days and every month.

### DIFF
--- a/dist/goaccess/obs-goaccess-last-month.sh
+++ b/dist/goaccess/obs-goaccess-last-month.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/bash
+# Generates goaccess HTML data from last months rotated apache access logs
+# Runs from cron.d on every first of the month
+
+NOW_MONTH=$(date +%Y-%m)
+PREVIOUS_MONTH=$(date -d "$nowmonth-15 last month" '+%Y%m')
+
+mkdir -p /srv/www/obs/api/public/analytics/$PREVIOUS_MONTH/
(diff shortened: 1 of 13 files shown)


dist/goaccess/obs-goaccess-last-month.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/bash
+# Generates goaccess HTML data from last months rotated apache access logs
+# Runs from cron.d on every first of the month
+
+NOW_MONTH=$(date +%Y-%m)
+PREVIOUS_MONTH=$(date -d "$nowmonth-15 last month" '+%Y%m')
```
eduardoj, line 6:
```suggestion
CURRENT_MONTH=$(date +%Y-%m)
PREVIOUS_MONTH=$(date -d "$CURRENT_MONTH-15 last month" '+%Y%m')
```
There is some inconsistency here...

dist/goaccess/obs-goaccess-last-month.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/bash
+# Generates goaccess HTML data from last months rotated apache access logs
+# Runs from cron.d on every first of the month
```
eduardoj, line 3: This comment refers to when this script is run. Shouldn't it be placed into the `dist/goaccess/obs-goaccess-last-month.crontab` file?

Also, in that `.contrab` file, this line `0 0 5 * *` is not exactly the same as what the comment says: ` # Runs from cron.d on every first of the month`...

PR #13382, opened 2022-11-15
Enhance the BsRequestAction factory

danidoni:
Provides a BsRequest submit action with a file diff.

## How To Test It

- Run the `dev:development_testdata:create` rake task on a local OBS instance.
- Once it finishes, boot up the local instance [as usual](https://github.com/openSUSE/open-build-service/wiki/Development-Environment-Tips-&-Tricks#control-services).
- http://localhost:3000/package/show/home:Admin/package_with_diff is the created package with the file diff.
- Navigate to [its open requests](http://localhost:3000/package/requests/home:Admin/package_with_diff) 
image
- And check the Changes tab of that request. There you'll find the file diff
image

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -109,6 +109,12 @@ namespace :dev do
         source_package: ruby_iggy
       )
 
+      create(:bs_request_action_submit_with_diff,
+             creator: iggy,
+             source_project_name: home_admin.name,
+             source_package_name: 'package_with_diff',
+             target_package_name: 'package_with_diff')
+
       test_package = create(:package, name: 'hello_world', project: home_admin)
       backend_url = "/source/#{CGI.escape(home_admin.name)}/#{CGI.escape(test_package.name)}"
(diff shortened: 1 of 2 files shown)


dmarcoux (approved):

src/api/lib/tasks/dev.rake
```diff
@@ -109,6 +109,12 @@ namespace :dev do
         source_package: ruby_iggy
       )
 
+      create(:bs_request_action_submit_with_diff,
```
saraycp, line 112: Instead of extending this method, we could call tasks declared in independent files.
Like we do [here](https://github.com/openSUSE/open-build-service/blob/b4783e908723bc26796767d6972f14f203fb2563/src/api/lib/tasks/dev.rake#L231) to create request with multiple actions.

We can rename the file [lib/tasks/dev/multiple_actions_request.rake](https://github.com/openSUSE/open-build-service/blob/master/src/api/lib/tasks/dev/multiple_actions_request.rake) to `lib/tasks/dev/requests.rake` and move all that code there.

PR #13383, opened 2022-11-15
Show maintainers in the beta request conversation

ncounter:
Fix https://trello.com/c/Gg6UnB4W

Showing maintainers in the Request conversation view.
List visibility is limited to a predefined size of 5, if they are more there is a button to open a popup with the whole list.

## TODO:
- [x] show package maintainers list in the beta redesigned request feature
- [x] show project maintainer list in the beta redesigned request feature
- [x] show the hint-alert when user is Project maintainer but not Package maintainer
- [x] user icon with circle border, something like `BsRequestOverviewAvatarsComponent`

## PR Review
Commit by commit

## Review by playing with it
1. If you already ran the `development_testdata` rake task you'll have a few requests already, you can see them in the Tasks menu.
2. Choose a request from the list.
3. Add multiple users as maintainers for both the target package and the target project of the request.
4. Go to see the conversation tab of the request.
5. On the right side you'll see the list of Package Maintainers and Project Maintainers.
6. Take your user out of the package maintainer list in case you are in it.
7. Take your user in the project maintainer list in case you are not yet in it.
8. Go back again to see the conversation tab of the request.
9. The hint-alert message should appear at the bottom of the right column.

## Screenshots preview
image


ncounter:
[Display maintainers of the target project/package (M)](https://trello.com/c/Gg6UnB4W/2039-display-maintainers-of-the-target-project-package-m)

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -15,6 +15,7 @@ def show
     if Flipper.enabled?(:request_show_redesign, User.session)
       @is_author = @bs_request.creator == User.possibly_nobody.login
       @is_target_maintainer = @bs_request.is_target_maintainer?(User.session)
+      @can_handle_request = @bs_request.state.in?([:new, :review, :declined]) && (@is_target_maintainer || @is_author)
```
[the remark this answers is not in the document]
ncounter: Thanks for the pointer. I made the change in 95ef8b748df15835952749f37779cfb4fb881bc6

src/api/app/policies/bs_request_policy.rb
```diff
@@ -7,4 +7,8 @@ def create?
 
     false
   end
+
+  def can_handle_request?(is_target_maintainer, is_author)
+    return true if record.state.in?([:new, :review, :declined]) && (is_target_maintainer || is_author)
```
dmarcoux, line 12: This code can be simplified:

```suggestion
    record.state.in?([:new, :review, :declined]) && (is_target_maintainer || is_author)
```

ncounter: copy/paste is a bad attitude I need to stop at some point :sweat_smile:

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -34,6 +35,13 @@ def show
       @open_reviews_for_staging_projects = @bs_request.reviews.opened.for_staging_projects
       @refresh = @action[:diff_not_cached]
 
+      # retrieve a list of all package maintainers that are assigned to at least one target package
+      @package_maintainers = target_package_maintainers
```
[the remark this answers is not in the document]
ncounter: I agree it is a relevant information. The very next question is: do we want to list the project maintainer in the same list of the package maintainers or do we want to differentiate them a little bit? And also: is the alert hint message still relevant or? it talks to the project maintainer for a different but overlapping reason.....

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -15,6 +15,7 @@ def show
     if Flipper.enabled?(:request_show_redesign, User.session)
       @is_author = @bs_request.creator == User.possibly_nobody.login
       @is_target_maintainer = @bs_request.is_target_maintainer?(User.session)
+      @can_handle_request = BsRequestPolicy.new(User, @bs_request).can_handle_request?(@is_target_maintainer, @is_author)
```
dmarcoux, line 18: As we discussed together @ncounter and I, this authorization check could be inlined in the view (with the usual `policy(my_record).my_method?`) instead of being passed around through multiple view levels with the `@can_handle_request` variable.

src/api/app/controllers/webui/request_controller.rb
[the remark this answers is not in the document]
dmarcoux: We should show the package and project maintainers, but clearly distinguish them from one another. Package maintainers should be listed before project maintainers, since project maintainers are the last resort maintainers, if somehow package maintainers aren't active or available.

src/api/app/policies/bs_request_policy.rb
```diff
@@ -7,4 +7,8 @@ def create?
 
     false
   end
+
+  def can_handle_request?(is_target_maintainer, is_author)
```
dmarcoux, line 11: To follow Pundit conventions, we should rename this method to `handle_request?` and encapsulate the logic from `is_target_maintainer` and `is_author` into this method. This way, we don't have to duplicate the logic whenever calling this Pundit policy.

src/api/app/views/webui/request/beta_show_tabs/_conversation_aside.html.haml
```diff
@@ -0,0 +1,53 @@
...
+        - else
+          - package_maintainers.take(size_limit).pluck(:login).each do |package_maintainer|
+            %div
+              = user_with_realname_and_icon(package_maintainer, short: true)
+          %p.mt-3
+            = link_to('Show all maintainers', '#', class: 'btn btn-outline-secondary show_dialog',
+                      data: { toggle: 'modal', target: '#show-maintainers-modal' })
+            = render partial: 'webui/request/show_package_maintainers', locals: { package_maintainers: package_maintainers }
+
+      - if can_handle_request && show_project_maintainer_hint
```
dmarcoux, line 49: In a previous comment, I wrote about relying on the `policy(...)` method from Pundit. This is where we would use it as such:

```suggestion
      - if policy(bs_request).handle_request? && show_project_maintainer_hint
```

src/api/app/policies/bs_request_policy.rb
dmarcoux, line 11: To follow Pundit conventions, we should rename this method to `handle_request?` and encapsulate the logic from `is_target_maintainer` and `is_author` into this method. This way, we don't have to duplicate the logic whenever calling this Pundit policy.

The method signature would become:

```suggestion
  def handle_request?
```

src/api/app/controllers/webui/request_controller.rb
ncounter: Thanks a lot for all the tips and the whole overview about policies and Pundit.
I addressed them all now.

src/api/app/policies/bs_request_policy.rb
```diff
@@ -7,4 +7,10 @@ def create?
 
     false
   end
+
+  def handle_request?
+    is_target_maintainer = record.is_target_maintainer?(User.session)
```
dmarcoux, line 12: In the context of this policy, the user to be authorized is available through `user`. So with this in mind, this line becomes:

```suggestion
    is_target_maintainer = record.is_target_maintainer?(user)
```

Same for the next line.

src/api/app/policies/bs_request_policy.rb
```diff
@@ -7,4 +7,10 @@ def create?
 
     false
   end
+
+  def handle_request?
```
hennevogel, line 11: This is not a policy. You are not deciding if someone can or can't do something.

dmarcoux: While it might be a bit stretching the definition of a policy, the alternative is to put this check in the controller and pass it around through multiple view levels. I prefer this approach with the policy.

danidoni: You can always push this to the model

ncounter: I liked the policy approach because it is self-contained and self-computed, no parameters to pass, no variables to carry on, etc.
I also agree it is not 100% a policy, it tells if someone can or cannot do something but it does not prevent anything in the end other than rendering a message.
Moving to the model could work but it does not really belong to the object model to be honest. I'm fine with the policy approach, but I don't want to enforce it.
I could see as an option to move it to the `user_helper.rb` class, because it is a user property more than a request one. The drawback still is to pass parameters around.
Opinions?

danidoni: I'm fine with having this in the policy

src/api/app/controllers/webui/request_controller.rb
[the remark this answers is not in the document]
ncounter: Addressed in 593b10c029f2ed62fce7d10b4c12fef287d715f5

I decided to keep them split into two different list, as it is a common practice we do across the whole project,

src/api/app/views/webui/request/_show_project_maintainers.html.haml
```diff
@@ -0,0 +1,13 @@
+.modal.fade#show-project-maintainers-modal{ tabindex: -1, role: 'dialog', aria: { labelledby: 'confirm-modal-label', hidden: true } }
```
krauselukas, line 1: @ncounter this is the same as the `_show_package_maintainers` partial besides the name `Package` and `Project`. Maybe you can adapt the package one to work for both and use it for both

ncounter: That makes sense, thanks for the pointer. I refactored it in https://github.com/openSUSE/open-build-service/pull/13383/commits/7cd65162f1964377a9fe8478d6a4faab2064a28f , the refactoring includes few changes outside of the beta because the former partial `_show_package_maintainers.html.haml` was used by both of the feature versions.

src/api/app/components/user_avatar_component.rb
```diff
@@ -0,0 +1,21 @@
...
+  def initialize(user)
+    super
+
+    @user = user
+  end
+
+  private
+
+  def avatar_object
+    @avatar_object = User.find_by_login(@user)
```
dmarcoux, line 11: This could be memoized, so it is not fetch over and over again.

```suggestion
    @avatar_object ||= User.find_by_login(@user)
```

ncounter:
> I do worry this is a lot to put there, could we move the project maintainer notice to the review dropdown instead? mt-5 is also quite a lot of spacing here, to maybe limit how much space this takes up, could we do:
> 
> ```gfm
> # Maintainers
>   * one
>   * two
>   * three
> ## Inherited from the project
>   * one
>   * two
>   * three
> ```
> 
> I think it would be possible to further limit how much space this takes up by combining package maintainers and project maintainers into a single list, and mentioning why they are a maintainer inline (sorting package maintainers first), but that may be too complex to address here


Yeah, I agree. For now I decreased the title size and the spacing, see below:


image



What do you think?

saraycp:
@ncounter can you please squash the commits before we merge this?

saraycp:
@ncounter I pushed a new commit to adjust some margins.

Without the "Show All" button, the warning was too close to the list of maintainers:

Screenshot 2022-11-22 at 10-03-07 Open Build Service

The "Show All" button wasn't aligned with the list of maintainers:
Screenshot 2022-11-22 at 10-10-29 Open Build Service

Adjusting the margins in those cases we have these results:

Screenshot 2022-11-22 at 10-12-33 Open Build Service

and

Screenshot 2022-11-22 at 10-09-52 Open Build Service

saraycp (approved):

PR #13393, opened 2022-11-17
Rename multiple_actions_request.rake to requests.rake

saraycp:
We are going to have one single file for all the rake tasks related to requests, so it's better to rename the file.

Renaming the file does not affect the usage of the task already declared in it. You still can run:

```
rails dev:requests:multiple_actions_request
```


dmarcoux (approved):
Sure, why not :+1:

danidoni (approved):

PR #13401, opened 2022-11-18
Add a multiaction submit request

danidoni:
This PR replaces the call to `request_with_submit_action_and_diff` during the data generation by the this new `request_with_multiple_submit_actions_and_diffs` because the new rake task kind of includes the previous one.

### DIFF
--- a/src/api/lib/tasks/dev/requests.rake
+++ b/src/api/lib/tasks/dev/requests.rake
@@ -67,8 +67,8 @@ namespace :dev do
       puts "* Request #{request.number} has been created."
     end
 
-    desc 'Creates a  request with submit action and diff'
-    task request_with_submit_action_and_diff: :environment do
+    desc 'Creates several requests with submit actions and diffs'
+    task request_with_multiple_submit_actions_and_diffs: :environment do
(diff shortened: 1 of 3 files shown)


src/api/lib/tasks/dev/requests.rake
```diff
@@ -79,22 +79,37 @@ namespace :dev do
...
       home_admin_project = RakeSupport.find_or_create_project(admin.home_project_name, admin)
 
-      request_action = create(:bs_request_action_submit_with_diff,
-                              creator: iggy,
-                              source_project_name: iggy_home_project.name,
-                              source_package_name: 'package_with_diff',
-                              target_project_name: home_admin_project.name,
-                              target_package_name: 'package_with_diff')
-
-      puts "* Request #{request_action.bs_request.number} has been created."
```
dmarcoux, line 97: It would be nice to keep this, so we can directly go to the request#show page of this new request.

danidoni: okay, I'll add it

dmarcoux (approved):

PR #13408, opened 2022-11-21
Add warning about request auto-acceptance

saraycp:
To make it clear that this is an "auto-accept" request, I add a badge next to the title.
The details are given in a warning box on top of the decline/accept box.

**Before:**

before_auto_accept

**After**

badge_and_warning_wo_icon

### DIFF
--- a/src/api/app/views/webui/request/beta_show.html.haml
+++ b/src/api/app/views/webui/request/beta_show.html.haml
@@ -33,6 +33,10 @@
           - if @bs_request.superseded_by.present?
             by
             = link_to(@bs_request.superseded_by, number: @bs_request.superseded_by)
+
+        - if @bs_request.accept_at.present?
+          %span.badge.alert-warning.ml-1 auto-accept
+
       %p.font-italic
         Created by
         = user_with_realname_and_icon(@bs_request.creator)
--- a/src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
+++ b/src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
@@ -72,6 +72,28 @@
     .comment_new.mt-3
       = render partial: 'webui/comment/new', locals: { commentable: bs_request }
     %hr
+
+    - if bs_request.accept_at.present?
+      .alert.alert-warning.p-2.mb-3
+        %i.fas.fa-exclamation-triangle
+        %span The current request
+        - if BsRequest::FINAL_REQUEST_STATES.include?(bs_request.state)
+          %span
+            was
+            %strong auto-accepted
+            at
+            %span.fuzzy-time{ title: "#{I18n.l bs_request.accept_at}" }
+            = succeed '.' do
+              = I18n.l bs_request.accept_at, format: :only_date
+        - elsif bs_request.accept_at.past?
+          %span
+            will be
+            %strong auto-accepted
+            after all the reviews are submitted.
+        - else
+          will be
+          %strong auto-accepted
+          in #{time_ago_in_words(bs_request.accept_at)}.
     = render RequestDecisionComponent.new(bs_request: bs_request, action: action,
                                           is_target_maintainer: is_target_maintainer,
                                           is_author: is_author)


src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
```diff
@@ -32,6 +32,20 @@
         - staging_project = review.project
         Is staged in
         = link_to(review.by_project, staging_workflow_staging_project_path(staging_project.staging_workflow.project, staging_project.name))
+
+    - if bs_request.accept_at.present?
+      %h4.mt-3 Request Auto-Acceptance
+      .pl-3
+        - if BsRequest::FINAL_REQUEST_STATES.include?(bs_request.state)
+          At
+          %span.fuzzy-time{ title: "#{l bs_request.accept_at}" }
```
dmarcoux, line 41: Could we prefix the `l` method with `I18n.`, so it's a bit clearer where this is coming from. This would result in:

```suggestion
          %span.fuzzy-time{ title: "#{I18n.l bs_request.accept_at}" }
```

This also applies to other lines with the `l` method.

eduardoj: There are already a couple of `l` methods in the views:

```
✗ rg -w l src/api/app/views/webui
src/api/app/views/webui/comment/_content.html.haml
8:      = link_to("#comment-#{comment.id}", title: l(comment.created_at.utc), name: "comment-#{comment.id}") do

src/api/app/views/webui/request/_request_history.html.haml
8:      = link_to('#request-creation', title: l(bs_request.created_at.utc), name: 'request-creation') do

src/api/app/views/webui/request/_requests_small.html.haml
15:              %u{ title: l(req.created_at.utc) }

src/api/app/views/webui/request/_request_history_element.html.haml
11:      = link_to("#status-history-#{element.id}", title: l(element.created_at.utc), name: "status-history-#{element.id}") do

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
46:            = link_to('#request-creation', title: l(bs_request.created_at.utc), name: 'request-creation') do

src/api/app/views/webui/request/_show_side_links.html.haml
31:        %span.fuzzy-time{ title: "#{l bs_request.accept_at}" }
33:          = l bs_request.accept_at, format: :only_date
```

For consistency, should we change them also?

saraycp:
> I'm not sure about that notice, it's not very noticeable, and it's not in proximity of the function it's performing. Would it maybe make more sense to use `alert-warning` somewhere around the "request accept/decline" box? I'm not sure how to make it more visible otherwise

@hellcp-work do you mean something like this?

alert_warning

src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
dmarcoux: Good point @eduardoj! This can happen in another PR :slightly_smiling_face:

saraycp:
We discussed offline and decided to move the warning closer to the decline/accept box.
But to make it clear at a first glance (without scrolling) we add a badge to the top.

src/api/app/assets/stylesheets/webui/requests.scss
```diff
@@ -13,6 +13,10 @@
   background-color: $cyan;
 }
 
+.badge-warning {
```
dmarcoux, line 16: This will overwrite the default `badge-warning` class from Bootstrap, see https://getbootstrap.com/docs/4.6/components/badge/#contextual-variations. We already use that class in other views, so we should avoid this.

Instead, we can combine `badge` with `alert-warning` and achieve the same.

dmarcoux (changes requested):

src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -33,6 +33,10 @@
           - if @bs_request.superseded_by.present?
             by
             = link_to(@bs_request.superseded_by, number: @bs_request.superseded_by)
+
+        - if @bs_request.accept_at.present?
+          %span.badge.badge-warning.ml-1 auto-accept
```
dmarcoux, line 38: Following up on my previous comment, this line will become:

```suggestion
          %span.badge.alert-warning.ml-1 auto-accept
```

saraycp: updated!

PR #13409, opened 2022-11-21
Add rake task to create request with build results

eduardoj:
Try it in the development environment with: `bin/rails dev:requests:request_with_build_results`

Example of running the new task in a Rails console:
```
frontend@45673968231d:/obs/src/api> bin/rails dev:requests:request_with_build_results
Creating a request which builds and produces build results...
* Request 17 has been created.
To start the builds confirm or perfom the following steps:
- Create the interconnect with openSUSE.org
- Create a couple of repositories in project home:Iggy or in project home:Iggy:branches:home:Admin
frontend@45673968231d:/obs/src/api> 
```

### DIFF
--- a/src/api/lib/tasks/dev/requests.rake
+++ b/src/api/lib/tasks/dev/requests.rake
@@ -96,5 +96,45 @@ namespace :dev do
 
       puts "* Request #{request_action.bs_request.number} has been created."
     end
+
+    desc 'Creates a request which builds and produces build results'
+    task request_with_build_results: :environment do
+      unless Rails.env.development?
+        puts "You are running this rake task in #{Rails.env} environment."
+        puts 'Please only run this task with RAILS_ENV=development'
+        puts 'otherwise it will destroy your database data.'
+        return
+      end
+
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+
+      puts 'Creating a request which builds and produces build results...'
+      admin = User.get_default_admin
+      home_admin_project = RakeSupport.find_or_create_project(admin.home_project_name, admin)
+
+      # Branch the hello_world package
+      iggy = User.find_by(login: 'Iggy') || create(:staff_user, login: 'Iggy')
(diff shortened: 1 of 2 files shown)


dmarcoux:
I ran the rake task and the generated request doesn't have build results due to missing build targets for the project.

no-builds

dmarcoux:
It's a bit confusing because the output of this new rake task does say that there will be build results, although there aren't any until we accomplish the steps you mentioned. Those steps could be in the output of the rake task.

eduardoj:
> It's a bit confusing because the output of this new rake task does say that there will be build results for the package in the newly created request, although there aren't any until we accomplish the steps you mentioned. Those steps could be in the output of the rake task.

Agree. I added a message, after the request is created. I also updated the message in the description of the PR. Please have a look again.

saraycp (changes requested):
I think we should call this task from the `dev:test_data:create` tasks defined in dev.rake.

src/api/lib/tasks/dev/requests.rake
```diff
@@ -96,5 +96,45 @@ namespace :dev do
...
+        creator: iggy,
+        target_project: home_admin_project,
+        target_package: 'hello_world',
+        source_project: branches_iggy,
+        source_package: hello_world_iggy
+      )
+      puts "* Request #{request.number} has been created."
+      puts 'To start the builds confirm or perfom the following steps:'
+      puts '- Create the interconnect with openSUSE.org'
+      puts '- Create a couple of repositories in project home:Iggy or in project home:Iggy:branches:home:Admin'
```
saraycp, line 137:
```suggestion
      puts '- Create a couple of repositories in project home:Iggy:branches:home:Admin'
```

As we do the branch before we add the repos to home:Iggy, those repos are not copied to the branched project. So it doesn't build anything.

saraycp (changes requested):

src/api/lib/tasks/dev/requests.rake
eduardoj: Ops, true. Changed.

eduardoj:
> I think we should call this task from the `dev:test_data:create` tasks defined in dev.rake.

Done. Please take a look again.

saraycp (approved):

PR #13418, opened 2022-11-23
Specify the group which is concerned by an event in notifications/emails

dmarcoux:
Before:
before

Now:
now

### DIFF
--- a/src/api/app/components/notification_action_description_component.rb
+++ b/src/api/app/components/notification_action_description_component.rb
@@ -5,6 +5,8 @@ class NotificationActionDescriptionComponent < ApplicationComponent
     @notification = notification
     @role = @notification.event_payload['role']
     @user = @notification.event_payload['who']
+    # If a notification is for a group, the notified user needs to know for which group. Otherwise, the user is simply referred to as 'you'.
+    @recipient = @notification.event_payload.fetch('group', 'you')
     @target_object = if @notification.event_payload['package']
                        "#{@notification.event_payload['project']} / #{@notification.event_payload['package']}"
                      else
@@ -23,9 +25,9 @@ class NotificationActionDescriptionComponent < ApplicationComponent
         commentable = @notification.notifiable.commentable
         "#{commentable.project.name} / #{commentable.name}"
       when 'Event::RelationshipCreate'
-        "#{@user} made you #{@role} of #{@target_object}"
+        "#{@user} made #{@recipient} #{@role} of #{@target_object}"
       when 'Event::RelationshipDelete'
-        "#{@user} removed you as #{@role} of #{@target_object}"
+        "#{@user} removed #{@recipient} as #{@role} of #{@target_object}"
       end
     end
   end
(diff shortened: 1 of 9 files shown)


dmarcoux:
https://trello.com/c/hQnGtTKu/2055-mention-the-group-title-when-the-relationship-is-for-a-group-you-are-in-m

src/api/spec/factories/notification.rb
```diff
@@ -6,6 +6,19 @@
...
+      originator { nil } # User login
+      recipient_group { nil } # Group title
+      role { nil } # Role title
+    end
+
+    after(:build) do |notification, evaluator|
+      notification.event_payload['who'] ||= evaluator.originator unless evaluator.originator.nil?
+      notification.event_payload['group'] ||= evaluator.recipient_group unless evaluator.recipient_group.nil?
+      notification.event_payload['role'] ||= evaluator.role unless evaluator.role.nil?
+      notification.event_payload['project'] ||= notification.notifiable.to_s unless notification.notifiable.nil?
```
eduardoj, line 19: This line fills the `project` key no matter what type `notifiable` is.

Could we add a condition... or modify the introduced code to prevent assigning other types of notifiable to the `project` key?

src/api/app/components/notification_action_description_component.rb
```diff
@@ -23,9 +25,9 @@ def call
         commentable = @notification.notifiable.commentable
         "#{commentable.project.name} / #{commentable.name}"
       when 'Event::RelationshipCreate'
-        "#{@user} made you #{@role} of #{@target_object}"
+        "#{@user} made #{@recipient} #{@role} of #{@target_object}"
```
ncounter, line 28: Probably out of this PR's scope but I was wondering if it would be worth the effort to have a link to the group here instead, just to let the user click and checkout the group members, reviews, requests, etc.

Orthogonal thinking it would also be nice to have a Group description text to define what is the scope or the purpose of that group, etc

src/api/spec/factories/notification.rb
dmarcoux: Good point, I'll adapt this.

src/api/app/components/notification_action_description_component.rb
dmarcoux: > Probably out of this PR's scope but I was wondering if it would be worth the effort to have a link to the group here instead, just to let the user click and checkout the group members, reviews, requests, etc.

It's a good idea, but I won't do this in this PR, since it could be done for the `@user` too and I want to keep the PR to the smallest set of changes possible.

src/api/spec/factories/notification.rb
dmarcoux: This is now only considering notifiables which are project and it won't crash even if the notifiable is `nil` since [`is_a?` is defined on Object](https://ruby-doc.org/core-3.1.2/Object.html#method-i-is_a-3F).

eduardoj (approved):

saraycp:
I pushed a rebase on master to include [these changes ](https://github.com/openSUSE/open-build-service/pull/13425)and make this PR build in OBS.

saraycp (approved):

Issue #256, opened 2013-04-30
file dependencies support for rpm


KostyaSha:
Please, add support for files in dependencies calculation. It used in fedora specs in number of packages. I think it could be also used in other rpm based distros http://rpm.org/wiki/PackagerDocs/Dependencies

coolo:
Can we close this again?

KostyaSha:
Well, i'm not working now on fedora full cycle build, but thats useful feature for companies that do this.

coolo:
well, rule of thumb: if for the last 6 years noone cared, the next 12 won't noone either.

PR #13419, opened 2022-11-23
Display file list in the binary view

### LINKED ISSUE #613: add file listing for rpms
Feature request. 
Like in koji when click on built rpm it make sense to show file list.
https://build.opensuse.org/package/binary/OBS:Server:2.5/rubygem-rdoc?arch=x86_64&filename=rubygem-rdoc-4.0.1-5.1.x86_64.rpm&repository=openSUSE_13.1
http://koji.fedoraproject.org/koji/rpminfo?rpmID=4292064

hellcp-work:
This adds displaying file list in the binary view

To verify:
1. Create a package that builds some binaries (I think obs-build only supports file lists in rpm and deb?, so build one of those)
2. Open binary view
3. See list of files in the package

image

Fixes #613

### DIFF
--- a/src/api/app/views/webui/package/_deps.html.haml
+++ b/src/api/app/views/webui/package/_deps.html.haml
@@ -85,3 +85,16 @@
           %tr
             %td{ colspan: '2' }
               %em No recommends
+
+  .table-responsive.col-sm-12.col-md-6.mt-4
+    %h4 Files
+    - if fileinfo['filelist']
+      %table.table.table-bordered.table-hover.table-sm
+        %thead
+          %tr
+            %th Filename
+        - fileinfo.elements('filelist') do |package_file|
+          %tr
+            %td.text-word-break-all= package_file
+    - else
+      %i No files
--- a/src/backend/BSXML.pm
+++ b/src/backend/BSXML.pm
@@ -476,6 +476,7 @@ our $fileinfo = [
       [ 'supplements' ],
       [ 'suggests' ],
       [ 'enhances' ],
+      [ 'filelist' ],
 
      [[ 'provides_ext' =>
 	    'dep',
--- a/src/backend/bs_repserver
+++ b/src/backend/bs_repserver
@@ -1283,7 +1283,7 @@ sub getbinary_info {
   }
   my @s = stat($path);
   die("404 $bin: $!\n") unless @s;
-  my $res = Build::query($path, 'evra' => 1, 'description' => 1, 'weakdeps' => 1) || {};
+  my $res = Build::query($path, 'evra' => 1, 'description' => 1, 'weakdeps' => 1, 'filelist' => 1) || {};
   if (!%$res && $path =~ /\/updateinfo\.xml$/) {
     my $updateinfos = readxml($path, $BSXML::updateinfo, 1);
     if ($updateinfos && @{$updateinfos->{'update'} || []} == 1) {


luc14n0:
This is something I consider quite useful for packager:

- Having a(n) (almost) full file list -- I'm not opposed to some throw out some garbage, though, otherwise the list would be really huge -- of all packages so **Zypper**, or perhaps something else, could find out what package provides the `/foo/bar/baz` file.

Although this PR is for the web UI, it has to touch the back-end too. And in the past there was a concern for generating a full list of files a package provides -- instead of a partial list we get nowadays --, because it would increase substantially the memory usage by the schedulers when dealing with repo data, see Michael Schroeder's comment in [issue 256](https://github.com/openSUSE/open-build-service/issues/256#issuecomment-17331993). for example.

That makes me wonder if things changed for the better meanwhile -- I really hope so, since that issue is from 2013 :smile: --, and whether that concern is still valid :thinking:

src/api/app/views/webui/package/_deps.html.haml
```diff
@@ -85,3 +85,16 @@
...
+  .table-responsive.col-sm-12.col-md-6.mt-4
+    %h4 Files
+    %table.table.table-bordered.table-hover.table-sm
+      %thead
+        %tr
+          %th Filename
+      - fileinfo.elements('filelist') do |package_file|
+        %tr
+          %td.text-word-break-all= package_file
+      - unless fileinfo['filelist']
```
dmarcoux, line 98: In this case, I would not even display a table, but only the heading and this _No files_ text in italic. What do you think?

dmarcoux (approved):

dmarcoux:
Nice work!

luc14n0:
> That concern doesn't really apply, because we aren't touching the scheduler here at all. The full file list is already available to us, it is generated in obs-build, but it was never exposed (I am not really aware why it was available, I couldn't really find usage of it anywhere, maybe it's used in other software that uses obs-build?). This does not enable file dependencies to be very clear, the full file list is not parsed by the scheduler still. What does change, and may be slightly concerning, is the response body size for `fileinfo_ext`, it's a couple times larger.
> 
> Keep in mind, all packages have to at some point have a full file list generated in the backend, because repo_md requires a filelists.xml file in repodata. This as far as I know is done by createrepo_c and not obs backend or obs-build, but that metadata has to already exist within the rpm file.

Oh, I got it. Now I have the (almost) full picture in my mind. It would be a concern if we were implementing file dependency support like mentioned in that issue. Good thing I don't care about that issue :smile:

I took a look at a filelists.xml from Tumbleweed's OSS and, indeed, it seems to be the full list of files -- in a 9M lines XML file :stuck_out_tongue_closed_eyes:. I can't believe I missed this before, this mean that **Zypper** could implement a search for virtually any file provided by packages in the repos, instead of only for some locations, like it does in the time being.

mlschroe:
The filelist of some packages is really big, so I'm not sure if the default should be to query and display it. How about making it a dynamic (AJAX) thing?

(As for filelists.xml, it's the reason why dnf is so slow when updating repo metadata...)

mlschroe:
Just as little data point: here are the top ten packages installed on my system with the biggest number of files:
```
  6910 texlive-tex4ht
  7383 gutenprint
  7466 oxygen5-icon-theme
  7775 go1.18
 12598 texlive-filesystem
 14944 libreoffice-l10n-en
 16001 kernel-default-devel
 22842 breeze5-icons
 35177 kernel-devel
111990 kernel-source
```

PR #13420, opened 2022-11-23
Correct the usage of list-group in notifications

hellcp-work:
This caused the list of filters to be displayed incorrectly:

Screenshot from 2022-11-23 13-48-33

A small adjustment in the way list-group is used corrects it:
Screenshot from 2022-11-23 13-48-39


src/api/app/components/notification_filter_component.html.haml
```diff
@@ -1,9 +1,9 @@
-.row.list-group-flush
+.list-group.list-group-flush.mt-2.mb-2
```
dmarcoux, line 1: This is really a nitpick, feel free to address this or not:

```suggestion
.list-group.list-group-flush.my-2
```

dmarcoux (approved):

PR #13421, opened 2022-11-23
Extend notifications primary and foreign keys capacity

saraycp:
We are adding new notification types which could increase the number of notifications considerably, so we extend the capacity of the primary key (from `int` to `bigint`) to avoid future overflow.

Not only do we extend the `Notification#id` and `NotifiedProject#id` but also the `NotifiedProject#notification_id`.

After running the migration these should be the results (from `int(11)` to `bigint(20)`):

```
MariaDB [api_development]> DESCRIBE notifications id;
+-------+------------+------+-----+---------+----------------+
| Field | Type       | Null | Key | Default | Extra          |
+-------+------------+------+-----+---------+----------------+
| id    | bigint(20) | NO   | PRI | NULL    | auto_increment |
+-------+------------+------+-----+---------+----------------+
```

```
MariaDB [api_development]> DESCRIBE notified_projects ;
+-----------------+-------------+------+-----+---------+----------------+
| Field           | Type        | Null | Key | Default | Extra          |
+-----------------+-------------+------+-----+---------+----------------+
| id              | bigint(20)  | NO   | PRI | NULL    | auto_increment |
| notification_id | bigint(20)  | NO   | MUL | NULL    |                |
...
+-----------------+-------------+------+-----+---------+----------------+
 ```
~**DO NOT MERGE**: this migration should be run during the next maintenance window.~

### DIFF
--- a/src/api/db/migrate/20221123163119_change_notifications_primary_and_foreign_keys_to_bigint.rb
+++ b/src/api/db/migrate/20221123163119_change_notifications_primary_and_foreign_keys_to_bigint.rb
@@ -1,0 +1,16 @@
+class ChangeNotificationsPrimaryAndForeignKeysToBigint < ActiveRecord::Migration[7.0]
+  def up
+    # This migration blocks the table, we are running it during Maintenance Window.
+    safety_assured do
+      # Foreign key
+      change_column :notified_projects, :notification_id, :bigint
+      # Primary key
+      change_column :notified_projects, :id, :bigint, auto_increment: true
+      change_column :notifications, :id, :bigint, auto_increment: true
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by dmarcoux** [comment]
We need to inform IBS people about this.

**Review by krauselukas — APPROVED** [review]

File: src/api/db/migrate/20221123163119_change_notifications_primary_and_foreign_keys_to_bigint.rb [hunk 1]
```diff
@@ -0,0 +1,13 @@
+class ChangeNotificationsPrimaryAndForeignKeysToBigint < ActiveRecord::Migration[7.0]
+  def up
+    # This migration blocks the table, we are running it during Maintenance Window.
+    safety_assured do
+      change_column :notified_projects, :notification_id, :bigint
```
saraycp, line 5: After talking with @eduardoj and testing the migration with real data, we think it's a good opportunity to also change the notified_projects id to bigint. Working on it...

eduardoj (approved):

PR #13424, opened 2022-11-24
Use the group subscription settings for Event::Relationship*

### LINKED ISSUE #12588: New notifications: maintainership added/removed does not follow group settings
A new notification for 'maintainership added/removed' was added and seems to have gone live on May 17 (the first time I noticed such a notification)

Generally, I think this is a good addition (it allows one to react at the time it happens and not be surprised months later why one is now the maintainer of something)

In the context of staging, though, this is a bit 'noisy', as every single adi project being created gets the group 'factory-staging' assigned as maintainer - and the new notification system informs all group members of this. So far 'as designed and 'ok' :)

# Issue Description
The issue though is that in my notifications, at the top, I have configured to not send emails for 'things relating to factory-staging; (email unchecked, web checked)

Nevertheless, the notifications are being sent by email

hellcp-work:
This enables filtering of the `Event::Relationship*` event types based on the type of subscription. This should have been included originally, because as it is, the notification avoids being filtered with the subscription settings for groups, and instead sends mails to all the members of the group.

Fixes #12588

### DIFF
--- a/src/api/app/models/event_subscription/find_for_event.rb
+++ b/src/api/app/models/event_subscription/find_for_event.rb
@@ -65,6 +65,8 @@ class EventSubscription
       receivers_and_subscriptions.values.flatten
     end
 
+    private
+
     def expand_receivers(receivers, channel)
       receivers.inject([]) do |new_receivers, receiver|
         case receiver
@@ -72,17 +74,14 @@ class EventSubscription
           new_receivers << receiver if receiver.is_active?
           puts "Skipped receiver #{receiver} because it's inactive" if @debug && !receiver.is_active?
         when Group
-          new_receivers += expand_receivers_for_groups(new_receivers, receiver, channel)
+          new_receivers += expand_receivers_for_groups(receiver, channel)
         end
 
         new_receivers
       end
     end
 
-    def expand_receivers_for_groups(_new_receivers, receiver, channel)
-      # We don't subscribe Groups so we have to get the group's users to get the subscriptions
-      return receiver.users if event.instance_of?(Event::RelationshipCreate) || event.instance_of?(Event::RelationshipDelete)
-
+    def expand_receivers_for_groups(receiver, channel)
       # We don't split events which come through the web channel, for a group subscriber.
       # They are split in the NotificationService::WebChannel service, if needed.
       return [receiver] if channel == :web || receiver.email.present?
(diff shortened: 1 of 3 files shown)


dmarcoux:
There are failing specs since the channel `:rss` isn't taken into account. This has to be addressed.

dmarcoux:
> But it's the mailer that is failing

The specs of the event mailer rely on `EventSubscription::FindForEvent#subscriptions` which is changed in this PR. This is why it fails.

See the line below

https://github.com/openSUSE/open-build-service/blob/e6acff36a1327e79e1f87bf60b6e5d72b904eb52/src/api/spec/mailers/event_mailer_spec.rb#L124

which then calls `EventSubscription::FindForEvent#subscriptions`

https://github.com/openSUSE/open-build-service/blob/e6acff36a1327e79e1f87bf60b6e5d72b904eb52/src/api/app/models/event/base.rb#L197-L203

dmarcoux:
> Should RSS just return an empty array? After all rss is not a channel that can be selected for group subscription

Yes, returning an empty array is fine.

src/api/app/models/event_subscription/find_for_event.rb
```diff
@@ -78,7 +78,10 @@ def expand_receivers(receivers, channel)
 
     def expand_receivers_for_groups(_new_receivers, receiver, channel)
       # We don't subscribe Groups so we have to get the group's users to get the subscriptions
-      return receiver.users if event.instance_of?(Event::RelationshipCreate) || event.instance_of?(Event::RelationshipDelete)
+      if event.instance_of?(Event::RelationshipCreate) || event.instance_of?(Event::RelationshipDelete)
+        return receiver.email_users if channel == :instant_email
```
dmarcoux, line 82: Once the `rss` channel is added here, this would be a use for a `case` statement instead of 3 `if` statements.

dmarcoux:
Should we close this PR and wait for #13453 to be merged? Am I understanding you correctly @hellcp-work?

eduardoj (approved):
The specs have been adapted to only consider groups without email.

Specs for groups with email would be welcome, but not having them should not stop merging this PR.

src/api/app/models/event_subscription/find_for_event.rb
```diff
@@ -72,17 +72,14 @@ def expand_receivers(receivers, channel)
...
 
         new_receivers
       end
     end
 
-    def expand_receivers_for_groups(_new_receivers, receiver, channel)
-      # We don't subscribe Groups so we have to get the group's users to get the subscriptions
-      return receiver.users if event.instance_of?(Event::RelationshipCreate) || event.instance_of?(Event::RelationshipDelete)
-
+    def expand_receivers_for_groups(receiver, channel)
```
danidoni, line 82: This is an internal method, it's better to make it private

eduardoj (approved):

PR #13430, opened 2022-11-25
Add withfilelist parameter to the binary query

hellcp-work:
This should avoid unnecessarily querying filelist if not required by the user.

To test this, compare the output of these two endpoints on a built package file
`/build/$project/$repository/$arch/$package/$filename?view=fileinfo_ext`
`/build/$project/$repository/$arch/$package/$filename?view=fileinfo_ext&withfilelist`

### DIFF
--- a/src/backend/BSXML.pm
+++ b/src/backend/BSXML.pm
@@ -479,6 +479,7 @@ our $fileinfo = [
       [ 'supplements' ],
       [ 'suggests' ],
       [ 'enhances' ],
+      [ 'filelist' ],
 
      [[ 'provides_ext' =>
 	    'dep',
--- a/src/backend/bs_repserver
+++ b/src/backend/bs_repserver
@@ -1302,7 +1302,7 @@ sub getbinary_info {
   }
   my @s = stat($path);
   die("404 $bin: $!\n") unless @s;
-  my $res = Build::query($path, 'evra' => 1, 'description' => 1, 'weakdeps' => 1) || {};
+  my $res = Build::query($path, 'evra' => 1, 'description' => 1, 'weakdeps' => 1, 'filelist' => $cgi->{'withfilelist'}) || {};
   if (!%$res && $path =~ /\/updateinfo\.xml$/) {
     my $updateinfos = readxml($path, $BSXML::updateinfo, 1);
     if ($updateinfos && @{$updateinfos->{'update'} || []} == 1) {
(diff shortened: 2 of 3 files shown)


src/api/app/lib/backend/api/build_results/binaries.rb
```diff
@@ -47,8 +47,8 @@ def self.rpmlint_log(project_name, package_name, repository_name, architecture_n
 
         # special view on a binary file for details display
         # @return [Hash]
-        def self.fileinfo_ext(project_name, package_name, repository, arch, filename)
-          fileinfo = http_get(['/build/:project/:repository/:arch/:package/:filename?view=fileinfo_ext', project_name, repository, arch, package_name, filename])
+        def self.fileinfo_ext(project_name, package_name, repository, arch, filename, filelist=0)
```
eduardoj: `self.fileinfo_ext` could be modified to accept `options` as last parameter, instead of `filelist`. You can take this other method as inspiration: https://github.com/openSUSE/open-build-service/blob/33c5a02429dd80905608397759aadbef265a5b63/src/api/app/lib/backend/api/sources/package.rb#L27-L32

I will also make clearer what parameter are we setting to `1` or `0` in the calls. And there is also an opportunity to refactor the `view` query parameter of the line below.

eduardoj (dismissed):

eduardoj:
The code is touching the backend. Could any of the developers of the @openSUSE/obs-backend developers leave a review, please? Thanks.

adrianschroeter:
except that you missed the [backend] prefix it should be fine.

But are you aware that there are rpms with more then 100k files inside? the current web interface would not be able to handle it nicely IMHO. Try kernel-source for example.

dmarcoux:
> there are rpms with more then 100k files inside? the current web interface would not be able to handle it

We could use the jQuery datatables for this. So the page loads, the data is fetched through an AJAX query. The table has pagination, search and sorting (maybe more, I don't remember now on top of my head).

danidoni:
> Tables aren't that great at displaying trees tbh, I would love this to be a widget that can display the information in a structured way instead

We can start with the tables and improve later...

danidoni:
@hellcp-work can we review this PR using the frontend?

danidoni:
> You should be able to use withfilelist parameter in the api, frontend doesn't currently have any way to use it though

Okay, can you write a small guide to help reviewers test this PR?

adrianschroeter (approved):

adrianschroeter:
I already wrote that it is fine from backend side, but did you also test with packages like kernel-source meanwhile?

PR #13431, opened 2022-11-25
Replace old watchlist code by code under new_watchlist feature in API

saraycp:
The `new_watchlist` feature has been enabled for all the users for a while now. It's time to replace the old code by the new one.

This PR only replaces the API's watchlist code.

**How to test?**

Play around with these endpoints, they should keep working:
- https://build.opensuse.org/apidocs-new/#/Person/put_person__login__watchlist
- https://build.opensuse.org/apidocs-new/#/Person/get_person__login__watchlist

### DIFF
--- a/src/api/app/controllers/person_controller.rb
+++ b/src/api/app/controllers/person_controller.rb
@@ -1,6 +1,7 @@
 require 'xmlhash'
 
 class PersonController < ApplicationController
+  # TODO: action userinfo no longer exists, remove these validations
   validate_action userinfo: { method: :get, response: :user }
   validate_action userinfo: { method: :put, request: :user, response: :status }
   validate_action grouplist: { method: :get, response: :directory }
@@ -11,7 +12,6 @@ class PersonController < ApplicationController
   skip_before_action :require_login, only: [:command, :register]
 
   before_action :set_user, only: [:post_userinfo, :change_my_password, :get_watchlist, :put_watchlist]
-  before_action :check_user_belongs_feature_flag, only: [:get_watchlist, :put_watchlist]
 
   def show
     @list = if params[:prefix]
@@ -160,7 +160,7 @@ class PersonController < ApplicationController
 
     xml = Xmlhash.parse(request.raw_post)
     ActiveRecord::Base.transaction do
-      update_new_watchlist(@user, xml)
+      update_watchlist(@user, xml)
     end
     render_ok
   end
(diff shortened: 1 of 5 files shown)


dmarcoux:
2 tests are failing in the minitest test suite, something is broken.

src/api/app/controllers/person_controller.rb
```diff
@@ -224,22 +223,9 @@ def internal_register
   end
 
   def update_watchlist(user, xml)
```
dmarcoux, line 225: Is something speaking against overwriting `update_watchlist` with the code from `update_new_watchlist`? `update_new_watchlist` could be removed since it would be then unneeded.

src/api/test/functional/person_controller_test.rb
```diff
@@ -129,8 +142,17 @@ def test_update_watchlist
     assert_select 'status', code: 'ok' do
       assert_select 'summary', 'Ok'
     end
+
     assert_equal ['Apache', 'BaseDistro3', 'home:tom'],
-                 User.find_by(login: 'tom').watched_project_names.sort
+                 User.find_by(login: 'tom')
```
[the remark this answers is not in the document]
saraycp: Done

src/api/spec/controllers/person_controller_spec.rb
```diff
@@ -106,6 +106,7 @@
           <watchlist>
             <project name="test-proj"/>
             <package name="test-pkg" project="test-proj"/>
+            <request number="1"/>
```
krauselukas, line 109: @saraycp maybe I overlook something, but for the `PUT` testcase you added the request to the xml but you dont check if its included in the end in the watchlist in line 125

krauselukas: shouldn't it also fail in line 126? I think it should include 3 elements in the watchlist, the package, request and project

saraycp: True! I updated the specs to make it more complete.

src/api/app/controllers/person_controller.rb
```diff
@@ -1,6 +1,7 @@
 require 'xmlhash'
 
 class PersonController < ApplicationController
+  # TODO: action userinfo no longer exists, remove these validations
```
danidoni, line 4: When are we going to remove this validations?

krauselukas (approved):

PR #13441, opened 2022-11-28
[api] support listing of deleted packages for remote projects

adrianschroeter:
There was a missing handling discovered in

 https://github.com/openSUSE/osc/issues/1197


saraycp:
Rebased on master and Rubocop complaint ignored (we can postpone the refactoring).

PR #13442, opened 2022-11-28
Stop notification creation if subscriber is same as the event initiator

rubhanazeem:
The current behavior is that if users change their role for one of their packages, they receive a notification mentioning their own name. Something like this `Admin made you maintainer of home:Admin / test`.

This pr tries to prevent such notifications


rubhanazeem:
[Avoid sending relationship notifications to the user creating the project (M)](https://trello.com/c/UfGVSUz4/2056-avoid-sending-relationship-notifications-to-the-user-creating-the-project-m)

src/api/app/services/notification_service/web_channel.rb
```diff
@@ -38,6 +38,8 @@ def call
     private
 
     def renew_notification(finder:)
+      return if @event.payload["user"] == @event.payload["who"]
```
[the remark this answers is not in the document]
rubhanazeem: This change has been removed

src/api/app/models/event/relationship_create.rb
```diff
@@ -4,6 +4,12 @@ class RelationshipCreate < Relationship
     self.description = 'Relationship was created'
 
     receiver_roles :any_role
+
+    def originator
```
dmarcoux, line 8: This method could go in the parent class `Relationship`, so the duplication would not be needed.

src/api/app/models/event/relationship_create.rb
```diff
@@ -4,6 +4,12 @@ class RelationshipCreate < Relationship
...
+      return User.find_by(login: payload['who']) if payload['who']
+
+      nil
```
danidoni, line 11: I believe this `nil` is redundant, as if the condition above is not met it will return nil automatically. I'm not 100% sure if this is more or less readable... Thoughts?

dmarcoux: You are right, it will return `nil` anyway whenever `if payload['who']` evaluates to `false`.

src/api/app/models/event/relationship.rb
```diff
@@ -25,5 +25,9 @@ def notifiable_id
 
       nil
     end
+
+    def originator
+      payload_address('who')
```
dmarcoux, line 30: This is the same [Event::Request](https://github.com/openSUSE/open-build-service/blob/842c83e03dc0e48516d68595a296ae2dfae0ee0c/src/api/app/models/event/request.rb#L18-L20), overriding what is defined in [Event::Base#originator](https://github.com/openSUSE/open-build-service/blob/842c83e03dc0e48516d68595a296ae2dfae0ee0c/src/api/app/models/event/base.rb#L210-L214).

krauselukas (approved):

danidoni (approved):

dmarcoux:
I fixed more specs which I somehow overlooked in the previous push.

PR #13444, opened 2022-11-28
Replace old watchlist view code by the new_watchlist one in WebUI

saraycp:
We are replacing the old watchlist code with the new one step by step.

In this case, we focus on the WebUI and 
- remove the view code we used to render for the old watchlist in favour of the new watchlist one,
- remove checks of the feature flag in the WatchedItemsController.
- adapt tests

**How to test?**

Play around adding and removing items from the Watchlist.

PR #13446, opened 2022-11-28
Drop Event::Request#find_watchers

saraycp:
We are replacing the old watchlist code with the new one step by step.

In this case, we remove `Event::Request#find_watchers` which was the old way to retrieve watchers of a project.


dmarcoux:
I believe this PR can be rebased now that #13444 has been merged.

dmarcoux (approved):

saraycp:
I pushed a spec fix.

danidoni (approved):

PR #13453, opened 2022-11-29
Deliver web notification to group members

hellcp-work:
`receivers_and_subscriptions` hash did not end up with any groups whatsoever, so the delivery of web notifications about groups wasn't working. This should fix that

# To verify:
1. Create a group and join a user to it
2. Have the group have a role in a project
3. Perform an action that would result in a notification in that project (except for `Event::Relationship*` one, that's addressed in another PR)
4. See if the user got notified about it

### DIFF
--- a/src/api/app/models/event_subscription/find_for_event.rb
+++ b/src/api/app/models/event_subscription/find_for_event.rb
@@ -52,6 +52,9 @@ class EventSubscription
               channel: default_subscription.channel,
               subscriber: receiver
             )
+          elsif channel == :web && receiver.instance_of?(Group) && receiver.web_users.any? { |u| EventSubscription.for_subscriber(u).find_by(options).present? }
+            # There is no default subscription for groups, so we are using the existing details
+            receivers_and_subscriptions[receiver] = EventSubscription.new(options.merge({ subscriber: receiver }))
           elsif @debug && default_subscription.present? && !default_subscription.enabled?
             puts "Skipped receiver #{receiver} because of a disabled default subscription"
           end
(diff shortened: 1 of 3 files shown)


krauselukas (commented):
@hellcp-work changes look good to me, how about a tiny spec that proves that this is working correctly? adding something to the `find_for_event_spec.rb`

krauselukas:
looks good now, but would still be great to add something to the `web_channel_spec.rb` to cover this

src/api/app/models/event_subscription/find_for_event.rb
```diff
@@ -52,6 +52,9 @@ def subscriptions(channel = :instant_email)
               channel: default_subscription.channel,
               subscriber: receiver
             )
+          elsif channel == :web && receiver.instance_of?(Group) && receiver.web_users.any? { |u| EventSubscription.for_subscriber(u).find_by(options).present? }
```
hennevogel, line 55: How can this be a group if we expand them above?

krauselukas: we dont expand them when its for the web channel

krauselukas: but maybe we should instead of the step we doing here? hmmm

src/api/spec/services/notification_service/web_channel_spec.rb
```diff
@@ -190,6 +191,21 @@
...
+        before do
+          event_subscription_group
+          create(:comment_request, commentable: new_bs_request, user: requester, updated_at: 1.hour.ago)
+        end
+
+        subject do
+          described_class.new(event_subscription_group, event).call
+        end
+
+        it 'only creates no notifications' do
```
danidoni, line 204:
```suggestion
        it 'creates no notifications' do
```

danidoni:
Squash'em all!

krauselukas (approved):
@hellcp-work with the follow up card in the backlog I'm fine with this :)

Issue #13443, opened 2022-11-28
Changes tab isn't working anymore


perlpunk:
# Issue Description

In this request https://build.opensuse.org/request/show/1038718#tab-pane-changes the Changes tab doesn't work, it says "Crunching the latest data" forever...

And it seems to happen for other requests as well

dmarcoux:
This happens inside or outside the beta. Somehow the `loadChanges()` JS function should be called whenever clicking on the button. That function was removed in 2f23bedd1f100e6dcd7ac1ab23f3153982755b29 as part of #13029.

![load-changes](https://user-images.githubusercontent.com/1102934/204356672-cc87ee67-d813-48db-b5f6-dfc397607363.gif)

eduardoj:
We have found that in this specific request there is an error when trying to uncompress the `.tar.xz` file. This is what we get from the API call:

```
The diff call for home:cdywan:branches:devel:languages:python/python-openqa_client failed: tar: exit status 512
```

This is not correctly shown in the UI. We are working on a fix on this.

hennevogel:
So this is "just" a problem with this particular request, dropped the priority label.

eduardoj:
> The 404 response could still be reproduced, with the same API call (`POST /request/1038718?cmd=diff`). It sounds to me that the backend should answer with another HTTP status, different to 404, for a `failed: tar: exit status 512` error.
> 
> An alternative could be to rescue from `Backend::NotFoundError` exceptions, in the [method source_diff](https://github.com/openSUSE/open-build-service/blob/81831ae612b29cc091367a7484716a8f802f8594/src/api/app/models/bs_request_action/differ/for_source.rb#L59-L65). But I still don't think this would be the right approach.
> 
> Could this be checked in the backend? CC: @openSUSE/obs-backend

I was wrong. A `Backend::NotFoundError` inherits from `Backend::Error`: https://github.com/openSUSE/open-build-service/blob/8cad6c83f9a462b94d4e4fdd426c05fb7c167a08/src/api/app/lib/backend/not_found_error.rb#L1-L3

This is not a problem in the backend.

rubhanazeem:
This issue is no longer occurring. Closing it now! If it happens anywhere else, please feel free to reopen it.

eduardoj:
Just for the record, the answer from the backend returns a different error:
```
> osc api -X POST "/request/1038718?cmd=diff"
Server returned an error: HTTP Error 404: Not Found
[...]

PR #13454, opened 2022-11-29
Bring back AJAX calls to request changes in beta

eduardoj:
Fixes issue #13443 only for beta, more concretely what refers to this sentence of [this comment](https://github.com/openSUSE/open-build-service/issues/13443#issuecomment-1329586992): 

> With the beta, the [button](https://github.com/openSUSE/open-build-service/blob/b07306d87baffd72a1ba7e51a64b1978a348ddfc/src/api/app/components/sourcediff_component.html.haml#L5) calls the loadChanges() JS function, but this function was removed in https://github.com/openSUSE/open-build-service/commit/2f23bedd1f100e6dcd7ac1ab23f3153982755b29 as part of https://github.com/openSUSE/open-build-service/pull/13029.

### DIFF
--- a/src/api/app/assets/javascripts/webui/request.js
+++ b/src/api/app/assets/javascripts/webui/request.js
@@ -166,3 +166,17 @@ function loadDiffs(element){
     }
   });
 }
+
+function loadChanges() { // jshint ignore:line
+  $('.loading-diff').removeClass('invisible');
+  var element = $('#changes-item');
+  // Always retrieve the changes of the first request action, by now
+  // TODO: request-action-id should be retrieved from the select component, when it is introduced
+  var url = '/request/' + element.data('request-number') + '/request_action/' + element.data('request-action-id') + '/changes';
+  $.ajax({
+    url: url,
+    success: function() {
+      $('.loading-diff').addClass('invisible');
+    }
+  });
+}
--- a/src/api/app/assets/javascripts/webui/request.js
+++ b/src/api/app/assets/javascripts/webui/request.js
@@ -169,9 +169,7 @@ function loadDiffs(element){
 
 function loadChanges() { // jshint ignore:line
   $('.loading-diff').removeClass('invisible');
-  var element = $('#changes-item');
-  // Always retrieve the changes of the first request action, by now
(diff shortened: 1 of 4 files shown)


rubhanazeem (approved):

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -212,6 +212,30 @@ def request_action
     end
   end
 
+  def request_action_changes
+    # TODO: Change @diff_limit to a local variable
+    @diff_limit = params[:full_diff] ? 0 : nil
+    # TODO: Change @actions to a local variable
+    @actions = @bs_request.webui_actions(filelimit: @diff_limit, tarlimit: @diff_limit, diff_to_superseded: @diff_to_superseded, diffs: true,
```
krauselukas, line 219: you already specify the `action_id` here, there should be no need to do `@actions.find { |action| action[:id] == params['id'].to_i }` again in the next line, you can simply call `.first` on it

src/api/app/assets/javascripts/webui/request.js
```diff
@@ -135,6 +135,9 @@ $(document).ready(function(){
     }
   });
   // TODO: Remove the enclosing code when the request_show_redesign feature is finished - END
+  $('#changes-item').on('shown.bs.tab', function () {
+    loadChanges();
+  });
```
ncounter, line 140: It seems this is a leftover now, `#changes-item` is gone since 933a14202dd262343223398f0cb8fab0f64b97d4

eduardoj: Removed. Thanks!

src/api/app/controllers/webui/request_controller.rb
eduardoj: Yes. The code can be refactored and adapted to what we already have in the show action of the request controller, regarding retrieving actions and a specific action.

But we opted for fixing the code, reverting it to what was already working. and modifying it as less as possible. We wrote down all those TODO's not to forget where we could improve the code. Basically, we first make it work, and then we refactor. It could be confusing to mix fixing and refactoring in the same PR.

danidoni:
Codecov is missing, merging

PR #13457, opened 2022-11-30
[dist] Fix obs-api %pre not to fail in chroots on disabling obsapidelayed.service

dmach:


dist/obs-server.spec
```diff
@@ -767,10 +767,15 @@ if [ "$1" == 2 ]; then
   if [ -e /etc/init.d/rc3.d/S50obsapidelayed ];then
     touch %{_rundir}/enable_obs-api-support.target
   fi
-  if systemctl --quiet is-active  obsapidelayed.service;then
+  if systemctl --quiet is-active obsapidelayed.service; then
     touch %{_rundir}/start_obs-api-support.target
     systemctl stop    obsapidelayed.service
-    systemctl disable obsapidelayed.service
+    if systemd-detect-virt --chroot; then
+      # systemctl is-active always exits with 0 in chroots, we don't know if the service runs or even exists
```
hennevogel, line 774:
```suggestion
      # If we are in a chroot, we don't know if the service runs or even exists. In that case ignore if disabling fails.
```

hennevogel (approved):

hennevogel:
@dmach ping

PR #13447, opened 2022-11-29
Receive notifications for build failures

danidoni:
This PR allows subscribers to receive notifications when a package fails to build.

This is how a notification for a package build failure looks like:

image

This is a follow up of #13428 

Edit: Update screenshot

PR #13460, opened 2022-11-30
Implement filters for build fail events

danidoni:
Now that we can receive notifications for build failure events, we
need a way to filter them from the rest.

This is how it looks like:

image

This is a follow up of #13447

### DIFF
--- a/src/api/app/components/notification_action_description_component.rb
+++ b/src/api/app/components/notification_action_description_component.rb
@@ -27,7 +27,7 @@ class NotificationActionDescriptionComponent < ApplicationComponent
       when 'Event::RelationshipDelete'
         "#{@user} removed #{@recipient} as #{@role} of #{@target_object}"
       when 'Event::BuildFail'
-        "Reason for the build: #{@notification.event_payload['reason']}"
+        "Build was triggered because of #{@notification.event_payload['reason']}"
       end
     end
   end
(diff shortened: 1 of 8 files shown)


src/api/app/components/notification_action_description_component.rb
```diff
@@ -28,6 +26,8 @@ def call
         "#{@user} made #{@recipient} #{@role} of #{@target_object}"
       when 'Event::RelationshipDelete'
         "#{@user} removed #{@recipient} as #{@role} of #{@target_object}"
+      when 'Event::BuildFail'
+        "Reason for the build: #{@notification.event_payload['reason']}"
```
hennevogel, line 30: What's the significance of the reason?

danidoni: I'm going to address this in #13447

danidoni: See https://github.com/openSUSE/open-build-service/pull/13447/commits/cdfcd6674ec7b6c8540242f75e5214c2973802bc

src/api/app/components/notification_component.rb
```diff
@@ -13,7 +13,11 @@ def notification_icon
     when 'Comment'
       tag.i(class: ['fas', 'fa-comments'], title: 'Comment notification')
     else
-      tag.i(class: ['fas', 'fa-user-tag'], title: 'Relationship notification')
+      if @notification.event_type == 'Event::BuildFail'
```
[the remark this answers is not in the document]
danidoni: It's ok for now because we only have notifiable_types `Package` for build failures, but that might change in the future. We'll change that then.

src/api/app/models/event/build.rb
```diff
@@ -35,6 +35,11 @@ def metric_fields
       }
     end
 
+    def parameters_for_notification
+      super.merge(notifiable_type: 'Package',
```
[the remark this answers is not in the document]
danidoni: Yessir! o7

dmarcoux (approved):

PR #13462, opened 2022-12-01
Use new receiver role name `project_watcher` for newly created event subscriptions

krauselukas:
In order to migrate the receiver role names from `watcher` to `project_watcher` we have to first use the new receiver role names for all newly created subscriptions. For now we still support the old naming by keeping the old receiver role names for the events, so we can still match them against each other. So we going to support both until we migrated all existing event subscriptions in a further step.

In order to support both the new and old naming for the receiver_role
`watcher` and `project_watcher` we need to update the old existing
records by just switching the `enabled` field while keeping the old
naming scheme for the receiver_role `watcher`. For non existing
subscriptions we create a new record with the new naming scheme
`project_watcher`

In order to render the checkboxes correctly for the old role names, we have to check for the old role names
when rendering the event subscription form and have to take them in to account. Otherwise we simply going with the 
new role names.

### DIFF
--- a/src/api/app/models/event/request.rb
+++ b/src/api/app/models/event/request.rb
@@ -84,11 +84,21 @@ module Event
       action_maintainers('sourceproject', 'sourcepackage')
     end
 
+    # TODO: Remove this method after all event subscriptions are migrated to use the new receiver role `source_project_watchers`
     def source_watchers
+      source_project_watchers
+    end
+
+    # TODO: Remove this method after all event subscriptions are migrated to use the new receiver role `target_project_watchers`
+    def target_watchers
+      target_project_watchers
+    end
+
+    def source_project_watchers
       source_or_target_project_watchers(project_type: 'sourceproject')
     end
 
-    def target_watchers
+    def target_project_watchers
       source_or_target_project_watchers(project_type: 'targetproject')
     end
 
--- a/src/api/app/models/event/base.rb
+++ b/src/api/app/models/event/base.rb
@@ -242,7 +242,12 @@ module Event
       ret
     end
 
+    # TODO: Remove this method after all event subscriptions are migrated to use the receiver role `project_watchers`
     def watchers
+      project_watchers
+    end
+
+    def project_watchers
       project = ::Project.find_by_name(payload['project'])
       return [] if project.blank?
 
--- a/src/api/app/models/event/build_fail.rb
+++ b/src/api/app/models/event/build_fail.rb
@@ -4,7 +4,7 @@ module Event
 
     self.message_bus_routing_key = 'package.build_fail'
     self.description = 'Package has failed to build'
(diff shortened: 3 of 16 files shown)


krauselukas:
[Rename receiver roles for projects (L)](https://trello.com/c/dkWUboHd/2053-rename-receiver-roles-for-projects-l)

krauselukas:
[WIP] Specs are failing, have to check what I have to adapt or what might be wrong

krauselukas:
> [WIP] Specs are failing, have to check what I have to adapt or what might be wrong

Alright, specs adapte to use the new receiver role name `project_watcher`, not WIP anymore

danidoni (approved):

src/api/app/models/event/request.rb
```diff
@@ -84,10 +84,20 @@ def source_maintainers
       action_maintainers('sourceproject', 'sourcepackage')
     end
 
+    def source_project_watchers
+      source_watchers
+    end
+
+    def target_project_watchers
+      target_watchers
+    end
```
saraycp, line 93: Shouldn't this be the opposite? So the old method (`source_watchers`) calls the new one ( `source_project_watchers`)? As you did before with `watchers`.

The same would apply for `target_watcher`.

saraycp (changes requested):

src/api/app/models/event/request.rb
krauselukas: i could do it the other way around as well, works both. but yeah then we just have to remove the old method instead of transfering the code. I will adapt

krauselukas: done :)

saraycp (changes requested):
There is a corner case that is not working:

1. I'm subscribed as `watcher` to an event. For example: comment created on project I watch.
1. I receive a notification when someone comments.
1. Your changes are applied.
1. I unsubscribe from the event described in 1).
1. I still receive notifications when someone adds comments to the project I watch.

After these steps, I see in database that there are two `EventSubscription`s for my user:

The old one still enabled:

```
  eventtype: "Event::CommentForProject",
  receiver_role: "watcher",
  user_id: 1,
  created_at: Mon, 05 Dec 2022 09:00:58.000000000 UTC +00:00,
  updated_at: Mon, 05 Dec 2022 09:00:58.371047000 UTC +00:00,
  enabled: true,
...
```

The latest one disabled:

```
  eventtype: "Event::CommentForProject",
  receiver_role: "project_watcher",
  user_id: 1,
  created_at: Mon, 05 Dec 2022 09:09:47.000000000 UTC +00:00,
  updated_at: Mon, 05 Dec 2022 09:09:47.585702000 UTC +00:00,
  enabled: false,
...
```

krauselukas:
Ah wait, something is broken with the form. Its does not render correctly

PR #13465, opened 2022-12-01
Drop old methods to handle watchlist

saraycp:
We are replacing the old watchlist code with the new one step by step.

In this case, we remove the code related to the old class `WatchedProject` and the methods defined in the User model to handle the watchlist.

Remove task to migrate from old to new watchlist as we can't longer reference the old model and the table will be removed soonish.


saraycp:
I've decided to push my remaining commits to this PR instead of opening a new PR. After all, they all consist of removing the WatchedProject-related code.

@hellcp-work can you have another look, please?

PR #13469, opened 2022-12-02
[api] speedup binary search database

adrianschroeter:
Also rename functional test, it basically overwrote the unit tests


src/api/spec/models/binary_release_spec.rb
```diff
@@ -85,7 +85,12 @@
 
       subject { described_class.update_binary_releases_via_json(repository, [repeated_binary_hash]) }
 
-      it { expect { subject }.to change(BinaryRelease, :count).by(-1) }
+      it 'sets the obsolete_time on the duplicated binary release record' do
```
krauselukas, line 88: @adrianschroeter can you confirm that this change in behavior is intended?

adrianschroeter: I am not sure which data set get removed. Is there a way to see the datasets of the repository and see which one disappears?

krauselukas:
I talked offline with Adrian, I removed the test case from the `binary_release_spec` which checked that repeated binary release records are removed when calling `update_binary_releases_via_json`. This does not apply anymore, the code was to handle those cases is removed in this PR https://github.com/openSUSE/open-build-service/pull/13469/files#diff-66b13f83703fa7445d86fa05460876542faf1729c2fcc8a63fa1f92e6ec189f3L46 Therefore the test case does not apply anymore. Repeated binary release records shouldn't exist in the first place, this was caused by buggy code in the past and need to be cleaned up.

krauselukas (approved):
review by commit

PR #13470, opened 2022-12-02
[api] fix whitespace in error message

bugfinder:
example: The current source revision in SUSE:Maintenance:26696/tar.SUSE_SLE-15_Updateare not on revision e56083b71a6732bddb9cf68337b8bdcd anymore.


adrianschroeter:
a better english would be using "is" instead of "are" ...

PR #13472, opened 2022-12-02
Point to the list of comments of a request from a notification link

### LINKED ISSUE #13459: The comments thread should appear by default when I click on a comment notification's link
# Issue Description
Whenever you receive a notification regarding a comment from a request, there is a link pointing to the comment thread. If you click on it, it should bring you to the comment thread, but instead we get the request page with nothing under the tabs. 

Expected Result
============
I expect the comments thread to appear by default.

How to Reproduce
===============
1. User Iggy is participating in the beta program (http://localhost:3000/my/beta_features) and has the `Request Show Redesign` feature on.
2. Iggy creates a Submit Request.
3. Iggy is subscribed to comments created on requests where Iggy is the creator (http://localhost:3000/my/subscriptions).
4. User Admin goes to that request and leaves a comment.
5. Iggy receives a notification saying there is a comment in that request.
6. Iggy clicks on the link in that notification and the request page appears, but no comments thread appear by default. Iggy has to click on the first tab to see the comments. 

Further Information
===============
Peek 2022-11-30 11-40

eduardoj:
Fixes #13459.

For reviewers: follow the steps described in the linked issue. You can use the review-app.

### DIFF
--- a/src/api/app/views/webui/request/beta_show.html.haml
+++ b/src/api/app/views/webui/request/beta_show.html.haml
@@ -108,4 +108,9 @@
 
       document.location.hash = anchor;
     }
+    else if (anchor == '#comments-list') {
+      $('#conversation-tab').tab('show');
+
+      document.location.hash = anchor;
+    }
 -# haml-lint:enable ViewLength
--- a/src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
+++ b/src/api/app/views/webui/request/beta_show_tabs/_conversation.html.haml
@@ -18,7 +18,7 @@
              project_maintainers: project_maintainers,
              show_project_maintainer_hint: show_project_maintainer_hint }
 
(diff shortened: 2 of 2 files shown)


dmarcoux (commented):
It's fixed for the beta request#show, but for the non-beta request#show, it's still somewhat weird. Is there something we can do about this?

view-shift

eduardoj:
> It's fixed for the beta request#show, but for the non-beta request#show, it's still somewhat weird. Is there something we can do about this? The viewport is shifted once the diffs are loaded.

For the non-beta we could add similar JavaScript code to point to the comment list, once the diffs are loaded:
```
if (anchor == '#comments-list') {
  document.location.hash = anchor;
}
```
Do you find this should be required in this PR? I don't have a strong opinion.

dmarcoux (approved):

PR #13473, opened 2022-12-02
Adapt code under new_watchlist feature flag

saraycp:
Adapt CSS and JavaScript.
This PR don't refactor the code related to `receiver_roles` fully so it doesn't collide with PR #13462.

Some refactoring affecting build results is postponed, see this [Trello card](https://trello.com/c/GGUO6mja/2839-stop-using-togglebuildinfo-in-favor-of-the-generic-togglecollapsibletooltip).


src/api/app/models/event_subscription.rb
```diff
@@ -60,11 +60,10 @@ class EventSubscription < ApplicationRecord
 
   after_save :measure_changes
 
-  def self.receiver_roles_to_display(user)
+  # TODO: unify both lists after renaming and migrating the *watchers methods
```
dmarcoux, line 63: Beside this method, both `RECEIVER_ROLE_TEXTS` and `BETA_RECEIVER_ROLE_TEXTS` are used only in this partial:

https://github.com/openSUSE/open-build-service/blob/1c240a0101f63e08c63784949b233736e5a6fac6/src/api/app/views/webui/subscriptions/_subscriptions_form.html.haml#L10

There's nothing preventing us from already merging `BETA_RECEIVER_ROLE_TEXTS` into `RECEIVER_ROLE_TEXTS`. This would simplify the view linked in this comment while also the method `self.receiver_roles_to_display` here.

What do you think?

src/api/app/models/event_subscription/for_event_form.rb
```diff
@@ -15,8 +15,9 @@ def call
 
     private
 
+    # TODO: use back receiver_roles after the renaming of *watchers methods
```
dmarcoux, line 18: I don't really understand what needs to be done here. Aren't we already using `event_class.receiver_roles`?

src/api/app/models/event_subscription.rb
saraycp: I finally merged them. Please have another look, @dmarcoux

dmarcoux (approved):

PR #13482, opened 2022-12-06
Document the superseded_by parameter to POST /request/{id}

dcermak:


src/api/public/apidocs-new/paths/request_id.yaml
```diff
@@ -73,6 +73,9 @@ post:
           after the last review.
 
         - `cancelapproval`: Reset the approval of a request
+
+        - `superseded_by`: Id of the request superseding this one. Required for
```
dmarcoux, line 77: `superseded_by` is not a command for this endpoint. This option should be documented under the `changestate` command.

dmarcoux (dismissed):
`superseded_by` is not a command for this endpoint. This option should be documented under the `changestate` command.

dmarcoux:
@dcermak: I've pushed a suggestion, let me know what you think of this:

suggestion

eduardoj:
We should also add the `superseded_by` parameter, and also make it clear there where can it be used.

dmarcoux:
> We should also add the `superseded_by` parameter, and also make it clear there where can it be used.

Good point, I'm on it.

dmarcoux:
@eduardoj Here's the parameter:

parameter

src/api/public/apidocs-new/paths/request_id.yaml
```diff
@@ -101,6 +106,11 @@ post:
       schema:
         type: string
       description: Define the new state
+    - in: query
+      name: superseded_by
+      schema:
+        type: string
+      description: Only in combination with the `changestate` command - Supersede the given request with another request by passing its ID in this parameter
```
eduardoj, line 113:
```suggestion
      description: Required and used only with the combination of the `changestate` command and `newstate=superseded` - Supersede the given request with another request by passing its ID in this parameter
```
I missed that this paremeter is used and required only with the combination of parameters shown above.

dmarcoux: I have applied the suggestion.

PR #13483, opened 2022-12-06
Log reason for request decline during auto_accept

hennevogel:
If there is an error at least log it so we can think of how we can get out of this situation.


src/api/app/models/bs_request.rb
```diff
@@ -888,7 +888,8 @@ def auto_accept
         rescue PostRequestNoPermission
           change_state(newstate: 'revoked', comment: 'Permission problem')
         rescue APIError
```
eduardoj, line 890:
```suggestion
        rescue APIError => e
```
Missing error handler...

eduardoj (changes requested):

eduardoj (approved):

PR #13489, opened 2022-12-06
Use `logger` instead `Rails.logger`

eduardoj:
From the [official guide](https://guides.rubyonrails.org/debugging_rails_applications.html#sending-messages):

_To write in the current log use the logger.(debug|info|warn|error|fatal|unknown) method from within a controller, model, or mailer:_


dmarcoux:
Enforcing this through a custom RuboCop cop would be nice, otherwise I expect us to have to redo this one day. What do you think?

As an inspiration, we already have some custom RuboCop cops [here](https://github.com/openSUSE/open-build-service/tree/master/src/api/lib/rubocop/cop).

eduardoj:
> Enforcing this through a custom RuboCop cop would be nice, otherwise I expect us to have to redo this one day. What do you think?
> 
> As an inspiration, we already have some custom RuboCop cops [here](https://github.com/openSUSE/open-build-service/tree/master/src/api/lib/rubocop/cop).

Yes, enforcing this through a custom RuboCop cop should be the way to go. But it seems that we have to adapt our test suites first. They are not ready/configured for this change.

PR #13491, opened 2022-12-06
[backend] support network during (kvm) VM builds

jjacque:
Now that https://github.com/openSUSE/obs-build allows to support network for VMs with `--vm-network` param, offer the possibility to enable this in the worker config using the option:
`OBS_VM_ENABLE_NET=true` in `/etc/sysconfig/obs-server`

Needs an `obs-build` version including https://github.com/openSUSE/obs-build/pull/725

It can be needed to build package relying for example on Maven (as its dependency management is pretty broken), i think that many private instances owners would find it useful, i'm personally running similar patches for years:
- https://github.com/algolia/open-build-service/tree/2.10.1-kvm-net
- https://github.com/algolia/obs-build/tree/20191114-kvm-net

Includes also a small fix of the spec file to prevent build failure if some files do not exists.

### DIFF
--- a/dist/obs-server.spec
+++ b/dist/obs-server.spec
@@ -522,10 +522,10 @@ fi
 # drop testcases for now
 rm -rf %{buildroot}%{__obs_api_prefix}/spec
 # only config for CI
-rm %{buildroot}%{__obs_api_prefix}/config/brakeman.ignore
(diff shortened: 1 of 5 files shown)


adrianschroeter:
fine, but can you please squash the commits together and prefix it with [backend] ?

Also please add a warning to the sysconfig template that it creates a security issue and also makes builds not reproducible.

PR #13495, opened 2022-12-08
Revert "Update rdoc to version 6.5.0"

eduardoj:
This reverts commit d71c494968d488f383ea5634a731c4ee4b6a3cd1.

Deploying this commit into production made the application unable to boot, throwing this message in the `/var/log/apache2/passenger_log` file:
```
[ E 2022-12-07 22:36:55.7711 2883/Tb age/Cor/Con/CheckoutSession.cpp:283 ]: [Client 2-6] Cannot checkout session because a spawning error occurred. The identifier of the error is f12ae16f. Please see earlier logs for details about the error.
[ E 2022-12-07 22:37:08.4944 2883/T3l age/Cor/App/Implementation.cpp:221 ]: Could not spawn process for application /srv/www/obs/api: The application encountered the following error: You have already activated stringio 3.0.1, but your Gemfile requires stringio 3.0.2. Since stringio is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports stringio as a default gem. (Gem::LoadError)
  Error ID: c61fa560
  Error details saved to: /var/run/passenger/passenger-error-Ylv6ce.html

[ E 2022-12-07 22:37:08.4966 2883/Tc age/Cor/Con/CheckoutSession.cpp:283 ]: [Client 3-6] Cannot checkout session because a spawning error occurred. The identifier of the error is c61fa560. Please see earlier logs for details about the error.
```

stringio 3.0.1 is installed with the `ruby3.1` rpm package.

stringio 3.0.2 is installed with the `obs-bundled-gems` rpm package.

There seems to be an issue for passenger when picking the correct gem. Adding `SetEnv GEM_PATH /usr/lib64/obs-api/ruby/3.1.0` to the apache configuration file workarounds the problem.

We prefer to revert the upgrade of the stringio rubygem until we find a better solution.


dmarcoux:
We should also pause the Depfu updates for this gem.

dmarcoux (approved):

PR #13497, opened 2022-12-08
Support the GitHub ping event

hellcp-work:
This should make GitHub display a green status next to ping events when first creating webhooks in repos.

# To verify:
1. Set up testing according to [instructions on the wiki](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration-Testing)
2. Send out a ping request
3. See a green response in OBS (and/or GitHub depending on the setup)

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -25,6 +25,13 @@ class Token::Workflow < Token
     raise Token::Errors::MissingPayload, 'A payload is required' if @scm_webhook.payload.blank?
 
     workflow_run.update(response_url: @scm_webhook.payload[:api_endpoint])
+
+    # We return early with a ping event, since it doesn't make sense to perform payload checks with it, just respond
+    if @scm_webhook.ping_event?
+      SCMStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, workflow_run, 'success', initial_report: true).call
+      return []
+    end
+
     yaml_file = Workflows::YAMLDownloader.new(@scm_webhook.payload, token: self).call
     @workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: @scm_webhook, token: self, workflow_run: workflow_run).call
 
--- a/src/api/spec/models/token/workflow_spec.rb
+++ b/src/api/spec/models/token/workflow_spec.rb
@@ -159,5 +159,47 @@ RSpec.describe Token::Workflow do
 
       it { expect { subject }.to change(workflow_token, :triggered_at) & change(workflow_run, :response_url).to('https://api.github.com') }
     end
+
+    context 'with a ping event' do
+      let(:scm) { 'github' }
+      let(:event) { 'ping' }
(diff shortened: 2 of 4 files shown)


src/api/app/models/token/workflow.rb
```diff
@@ -53,13 +56,17 @@ def validation_errors
...
-      @workflows.each do |workflow|
+      @workflows.try(:each) do |workflow|
         error_messages << workflow.errors.full_messages unless workflow.valid?
       end
 
       error_messages.flatten
     end
   end
+
+  def ping?
```
dmarcoux, line 67: This should be a method in the [SCMWebhook](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/scm_webhook.rb) class. It doesn't belong here.

dmarcoux (dismissed):
This should be covered with specs.

What is the payload for this webhook event like? Do you have an example?

src/api/app/models/token/workflow.rb
```diff
@@ -25,14 +25,17 @@ def call(options)
...
     workflow_run.update(response_url: @scm_webhook.payload[:api_endpoint])
-    yaml_file = Workflows::YAMLDownloader.new(@scm_webhook.payload, token: self).call
-    @workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: @scm_webhook, token: self, workflow_run: workflow_run).call
+    # We can't get the yaml from a ping event, because the payload doesn't contain all the necessary information
+    unless ping?
+      yaml_file = Workflows::YAMLDownloader.new(@scm_webhook.payload, token: self).call
+      @workflows = Workflows::YAMLToWorkflowsService.new(yaml_file: yaml_file, scm_webhook: @scm_webhook, token: self, workflow_run: workflow_run).call
+    end
 
     return validation_errors unless validation_errors.none?
```
dmarcoux, line 34: `validation_errors` is checking for validation errors in workflows, so this part of the validation should be skipped.

dmarcoux (approved):

src/api/spec/models/token/workflow_spec.rb
```diff
@@ -159,5 +159,47 @@
...
+        allow(SCMStatusReporter).to receive(:new).and_return(proc { true })
+      end
+
+      subject { workflow_token.call(workflow_run: workflow_run, scm_webhook: scm_extractor.call) }
+
+      it 'returns before checking for validation errors' do
+        expect(subject).to be_empty
+      end
+
+      it { expect { subject }.to change(workflow_token, :triggered_at) & change(workflow_run, :response_url).to('https://api.github.com') }
```
eduardoj, line 197:
```suggestion
      it { expect { subject }.to change(workflow_token, :triggered_at).and change(workflow_run, :response_url).to('https://api.github.com') }
```

There are some other parts of our code where we use the `&` operator. But we should make use of `and` to chain expectations. See https://relishapp.com/rspec/rspec-expectations/v/3-12/docs/compound-expectations#use-%60and%60-to-chain-expectations

This suggestion, still not tested, should also work.

eduardoj (approved):

PR #13500, opened 2022-12-08
Use correct commit sha on tag push in scm/ci integration for github

krauselukas:
The json payload key `:after` in the webhook of a push event from GitHub contains the sha of the tag itself. Instead we need to use the commit sha the tag is refering to. This is part of `:head_commit => :id`.

Fixes #13206

image

PR #13531, opened 2022-12-14
Fix scm integration for tag push to gitea

krauselukas:
Similar to github, we have to use the `:head_commit => :id` attribute
from the webhook payload on a tag push in order to point to
commit the tag is refering to, not the sha of the tag itself.
Otherwise we are not able to pull the workflow configuration and the obs_scm service
won't pull the correct sources since the wrong commit_sha is used in the branch_request
file.
The download url for a tag push is different to the one for
a regular push/pull_request event. In order to pull the
workflow configuration file the tag version is refering to
we have to adapt the url. Example of a valid download url for a tag: https://gitea.opensuse.org/krauselukas/hello_world/raw/tag/4.0/.obs/workflows.yml

Related to https://github.com/openSUSE/open-build-service/pull/13500

How to test:
Follow our testing guide here https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration-Testing and simulate a tag push to a Gitea repo. You could use the openSUSE instance https://gitea.opensuse.org/
See my test repo for reference https://gitea.opensuse.org/krauselukas/hello_world

### DIFF
--- a/src/api/app/services/trigger_controller_service/scm_extractor.rb
+++ b/src/api/app/services/trigger_controller_service/scm_extractor.rb
@@ -188,7 +188,8 @@ module TriggerControllerService
       # We need this for Workflow::Step#target_package_name
       # 'target_branch' will contain a commit SHA
       payload.merge!({ tag_name: payload_ref.sub('refs/tags/', ''),
-                       target_branch: @payload[:after] })
+                       target_branch: @payload.dig(:head_commit, :id),
+                       commit_sha: @payload.dig(:head_commit, :id) })
     end
   end
 end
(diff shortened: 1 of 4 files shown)


krauselukas:
[Fix issue for GitHub's tag_push event (M)](https://trello.com/c/2fdAo9zJ/2066-fix-issue-for-githubs-tagpush-event-m)

krauselukas:
I also realized we can do some clean up on the `@scm_payload` content, in some cases we have redundant values, but I would make this part of another PR...

danidoni (approved):

PR #13532, opened 2022-12-14
Hide empty blocks if content value is empty

ncounter:
Fix https://trello.com/c/tFUkVvrt

### Before
image
image

### After
image
image


ncounter:
[We should not show an empty request body on the workflow run show page when it runs successfully (S)](https://trello.com/c/tFUkVvrt/2069-we-should-not-show-an-empty-request-body-on-the-workflow-run-show-page-when-it-runs-successfully-s)

danidoni (approved):

danidoni (approved):

PR #13536, opened 2022-12-15
Revert "[src/api] Update selenium-webdriver: 4.5.0 → 4.7.1 (minor)"

hennevogel:
Reverts openSUSE/open-build-service#13474

ships a pre-build x86_64 only selenium-manager (https://github.com/SeleniumHQ/selenium/issues/11357) binary that will break builds on other architectures. We will pin this until this is solved.


danidoni (approved):

titusfortner:
Curious what the issue was with Selenium Manager. Selenium is shipped with Selenium Manager support for 3 architectures right now. They are only executed if drivers are not otherwise found (i.e. it would otherwise error), and any exception should be rescued and the original error provided (to put drivers on PATH).

hennevogel:
hey @titusfortner thanks for stopping by 🤝 

The problem is that we do not allow shipping foreign arch ELF binaries (especially on our Linux distribution) and we build for more than the x86 and arm variants. I guess you should make use of [platformed gems](https://guides.rubygems.org/specification-reference/#platform=) and only ship the pre-build binaries that will run on the platform you build for.

I would also find it nice if I would be able to disable prebuild binaries completely (building them during gem install) as they pose a security risk because I can't verify where they come from and what they do.

PR #13538, opened 2022-12-15
Reject set_flags step when deleting projects after a PR was closed

### LINKED ISSUE #12637: [production] missing keyword: :project_name
[See this exception on Errbit](https://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/6289fb9584bf460539e6592e)

krauselukas:
There is no project that should be deleted for the set_flags step in a workflow, when a PR is closed on the SCM. There for it should be one of the steps that gets rejected when we clean up. This will also fix an issue, since the `target_project_name` method of the `SetFlags` expects to get passed a keyword argument, which all other steps don't.

Fixes #12637

### DIFF
--- a/src/api/app/models/workflow.rb
+++ b/src/api/app/models/workflow.rb
@@ -114,7 +114,9 @@ class Workflow
   # TODO: Extract this into a service
   def destroy_target_projects
     # Do not process steps for which there's nothing to do
-    processable_steps = steps.reject { |step| step.instance_of?(::Workflow::Step::ConfigureRepositories) || step.instance_of?(::Workflow::Step::RebuildPackage) }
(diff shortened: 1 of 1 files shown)


krauselukas:
[Fix issue in set_flags step (M)](https://trello.com/c/azqxF1tx/2067-fix-issue-in-setflags-step-m)

hennevogel:
I'm beginning to think we should have a matrix somewhere about things that happen/don't happen in the code. But also in the documentation...

hennevogel:
Created a card https://trello.com/c/GbwYPuBv/207-step-support-action-matrix

hennevogel (approved):

PR #13551, opened 2022-12-19
Take workflow configuration file from external URL

saraycp:
From now on, the workflows configuration file doesn't need to be inside the repository. We can specify any other accessible URL.

Screenshot 2022-12-20 at 09-18-44 Open Build Service

### DIFF
--- a/src/api/app/views/webui/users/tokens/edit.html.haml
+++ b/src/api/app/views/webui/users/tokens/edit.html.haml
@@ -49,12 +49,18 @@
                     Leaving this text field empty will not update your SCM token.
                 .form-group#workflow-configuration-path-group
                   = f.label(:workflow_configuration_path, 'Path for Workflows Configuration File:')
-                  %abbr.text-danger{ title: 'required' } *
                   .input-group
-                    = f.text_field(:workflow_configuration_path, class: 'form-control', required: true,
+                    = f.text_field(:workflow_configuration_path, class: 'form-control',
                                                                  placeholder: 'Eg: .my_subdir/my_workflows_file.yml')
                   .form-text.text-muted
-                    The default path is '.obs/workflows.yml'
+                    The default path is '.obs/workflows.yml'.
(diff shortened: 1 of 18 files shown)


src/api/app/views/webui/users/tokens/edit.html.haml
```diff
@@ -49,12 +49,18 @@
...
                   .form-text.text-muted
-                    The default path is '.obs/workflows.yml'
+                    The default path is '.obs/workflows.yml'.
+                .form-group#workflow-configuration-url-group
+                  = f.label(:workflow_configuration_url, 'URL to Workflows Configuration File:')
+                  .input-group
+                    = f.text_field(:workflow_configuration_url, class: 'form-control',
+                                                                placeholder: 'Eg: https://example.com/.my_subdir/my_workflows_file.yml')
+                  .form-text.text-muted
+                    When the URL is given, the path will be ignored. It should be a public URL.
```
dmarcoux, line 63: I think we should rephrase this. The URL has to be accessible for the OBS instance, which might be public or not, this we don't know.

src/api/app/views/webui/users/tokens/show.html.haml
```diff
@@ -1,3 +1,4 @@
+-# haml-lint:disable ViewLength
```
dmarcoux, line 1: Refactoring this view is tracked in our backlog [here](https://trello.com/c/6Oe3vBnC/214-split-up-the-users-tokens-showhtmlhaml-view).

danidoni (approved):

danidoni:
I would appreciate a screenshot with the new changes in the form

saraycp:
> I would appreciate a screenshot with the new changes in the form

Screenshot and description added.

dmarcoux:
I will handle the Brakeman false positive. It needs to be ignored.

dmarcoux (approved):

PR #13553, opened 2022-12-19
Add monitoring for Event::ServiceFail

bmwiedemann:
Add monitoring for Event::ServiceFail

Because we need to see in monitoring / be alerted about excessive ServiceFail events
to fix issues before too many users are affected.

### DIFF
--- a/src/api/app/models/event/service_fail.rb
+++ b/src/api/app/models/event/service_fail.rb
@@ -21,6 +21,32 @@ module Event
       attribs['error'] = attribs['error'][0..800]
       super(attribs, keys)
     end
+
+    def metric_measurement
+      'service'
+    end
+
+    def metric_tags
+      error = case payload['error']
+              when start_with?('bad link:')
+                'bad_link'
+              when /^ 400 remote error:.*.service  No such file or directory/
+                'service_missing'
+              when /^ 400 remote error:.*service parameter.*is not defined/
+                'unknown_service_parameter'
+              else
+                'unknown'
+              end
+
+      {
+        status: 'fail',
+        error: error
+      }
+    end
+
+    def metric_fields
+      { value: 1 }
+    end
   end
 end
 
--- a/src/api/app/models/event/service_success.rb
+++ b/src/api/app/models/event/service_success.rb
@@ -15,6 +15,20 @@ module Event
       h['X-OBS-Package'] = "#{payload['project']}/#{payload['package']}"
       h
     end
+
+    def metric_measurement
+      'service'
+    end
+
+    def metric_tags
+      {
+        status: 'success'
+      }
+    end
+
+    def metric_fields
+      { value: 1 }
+    end
   end
 end
 


src/api/app/models/event/service_fail.rb
```diff
@@ -21,6 +21,31 @@ def set_payload(attribs, keys)
       attribs['error'] = attribs['error'][0..800]
       super(attribs, keys)
     end
+
+    def metric_tags
+      error = case payload['error']
+              when /cannot be expanded/
+                700
```
krauselukas, line 28: @bmwiedemann status code 700 & 701 are not part of http code world as far as I know, I cannot come up with an alternative quickly. Maybe we don't include those two case for now an only go with the others

src/api/app/models/event/service_fail.rb
```diff
@@ -21,6 +21,31 @@ def set_payload(attribs, keys)
...
+              when /bad link/
```
krauselukas, line 29: or for bad link we could maybe go for status 409 which indicated a conflict

src/api/app/models/event/service_fail.rb
bmwiedemann: The 7xx codes are meant to signify that it does not map to a standard 4xx or 5xx code. And 500 is already used below for unclassified errors.
The idea was to add an explanation in Grafana.

src/api/app/models/event/service_fail.rb
bmwiedemann: Hmm. Wouldn't a conflict cause a "cannot be expanded" error? Then we could use 409 there.

src/api/app/models/event/service_fail.rb
bmwiedemann: I mapped both to 409 now.

src/api/app/models/event/service_fail.rb
```diff
@@ -21,6 +21,40 @@ def set_payload(attribs, keys)
...
+      error = case payload['error']
+              when start_with?('bad link:')
+                'bad_link'
+              when /400 remote error:.*.service  No such file or directory/
+                'service_missing'
+              when start_with?('404 remote error:'), start_with?('FileNotFoundError:'), /ERROR 404: Not Found/, /No such file or directory/
+                'not_found'
+              when start_with?(' 400 remote error:')
+                'remote_error_400'
+              when /ERROR 403: Forbidden/
```
hennevogel, line 39: I'm not sure this will be the same forbidden every time we see this. I would count this as `unknown`

hennevogel (changes requested):

src/api/app/models/event/service_fail.rb
```diff
@@ -21,6 +21,40 @@ def set_payload(attribs, keys)
...
+
+    def metric_measurement
+      'service'
+    end
+
+    def metric_tags
+      error = case payload['error']
+              when start_with?('bad link:')
+                'bad_link'
+              when /400 remote error:.*.service  No such file or directory/
```
hennevogel:
```suggestion
              when /400 remote error:.*.service  No such file or directory$/
```

hennevogel (approved):

PR #13558, opened 2022-12-20
Update brakeman to 5.4.0

dmarcoux:
This is the latest version.

### DIFF
--- a/.github/workflows/brakeman.yml
+++ b/.github/workflows/brakeman.yml
@@ -16,4 +16,4 @@ jobs:
         gem install --no-format-executable brakeman -v 5.4.0
     - name: Brakeman
       run: |
-        brakeman --rails6 -p src/api
+        brakeman -p src/api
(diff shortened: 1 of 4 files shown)


.github/workflows/brakeman.yml
```diff
@@ -16,4 +16,4 @@ jobs:
         gem install --no-format-executable brakeman -v 5.4.0
     - name: Brakeman
       run: |
-        brakeman --rails6 -p src/api
+        brakeman -p src/api
```
saraycp, line 19: Shouldn't we force to Rails 7, so it doesn't use a 7+ version before we adapt our code?

dmarcoux: There's no Rails 7 mode, see this screenshot from Brakeman's help:

no-rails-7-mode

danidoni (approved):

dmarcoux:
I pushed again to add a FIXME comment regarding a Brakeman warning.

saraycp (approved):

PR #13564, opened 2022-12-20
Split out binary methods out of PackageController

hellcp-work:
I started working on this because of filelist changes. I will need to add some ajax methods, for handling data from the backend there, but package controller is already overloaded, so I decided to split it up and move all of the relevant methods into another controller. There are still a few failing tests, because of concerns handling them for some reason, but otherwise things should be working


dmarcoux:
Could you split this PR in smaller ones? This would help with reviewing the changes. Like the changes from `Webui::Packages::BinariesHelper` could be in their own PR. For the rest, I don't know since I didn't look at all the changes in details, 39 changed files in a PR is simply too much to review.

PR #13571, opened 2022-12-21
Correctly merge default options in policies

dmarcoux:
Before, options passed to the policies were simply ignored and the default values were always taken. Only `Staging::WorkflowPolicy` was correct.

I realized this was wrong while pairing with @ncounter as we worked on a feature.

### DIFF
--- a/src/api/app/views/webui/main/_status_messages.html.haml
+++ b/src/api/app/views/webui/main/_status_messages.html.haml
@@ -1,3 +1,7 @@
+-# This is needed since inferring the policy class with `policy(StatusMessage.new)` would use StatusMessagePolicy which is for the API
+  Anonymous users can see this view, so we need to set `ensure_logged_in` to `false`
+- status_message_policy = Webui::StatusMessagePolicy.new(User.possibly_nobody, StatusMessage.new, ensure_logged_in: false)
+
 .card.mb-3
   .card-header.d-flex.justify-content-between
     %h5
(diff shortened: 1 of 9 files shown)


dmarcoux:
[Correctly merge default options in policies](https://trello.com/c/6HOZmtde/2081-correctly-merge-default-options-in-policies)

ncounter (approved):
:+1:

src/api/app/views/webui/main/_status_messages.html.haml
```diff
@@ -1,3 +1,7 @@
+-# This is needed since inferring the policy class with `policy(StatusMessage.new)` would use StatusMessagePolicy which is for the API
+  Anonymous users can see this view, so we need to set `ensure_logged_in` to `false`
+- status_message_policy = Webui::StatusMessagePolicy.new(User.possibly_nobody, StatusMessage.new, ensure_logged_in: false)
```
dmarcoux, line 3: Without the changes in the policy, `ensure_logged_in` would still be `true`.

See:
```ruby
[4] pry(main)> opts = { ensure_logged_in: false }
=> {:ensure_logged_in=>false}
[5] pry(main)> opts.merge(ensure_logged_in: true)
=> {:ensure_logged_in=>true} # Wrong! It should be false, the default value has precedence over what I'm passing
[6] pry(main)> { ensure_logged_in: true }.merge(opts)
=> {:ensure_logged_in=>false}
```

saraycp (approved):

PR #13572, opened 2022-12-21
Remove unneeded variable @namespace in groups#index

dmarcoux:
This instance variable is not set in Webui::GroupsController and there are no partials beside `_tabs.html.haml` and
`_breadcrumb_items.html.haml` under `app/views/webui/configuration/`, so this can be safely removed.

Co-authored-by: Dario Leidi <dleidi@suse.com>


dmarcoux:
@ncounter and I found this while working on something else.

ncounter (approved):

PR #13573, opened 2022-12-21
Manage News in a dedicated page

ncounter:
`Manage News` is a new page to you guessed it, manage news / status messages!!! We went with `News` to be consistent with the UI on the homepage which refers to status messages as news. This confusion naming will be aligned in #13574.

The page works with AJAX, we didn't implement a full-text search since this would have increased the scope of the changes. It's accessible from the homepage by clicking on `News` and also from a tab under the `Configuration` page.

Preview:
preview

Closes #13458

PR #13590, opened 2022-12-22
Pass severity directly to icon helper

dmarcoux:
It makes the helper a tiny bit simpler.


dmarcoux:
We stumbled on this code while working on #13573 and thought about using it to generate the options of a `<select>` tag. With these changes, it's possible to reuse the helper. In the end, we didn't need the helper, but still thought this small refactoring is useful.

PR #13594, opened 2022-12-23
Filter out the irrelevant states in the badge

### LINKED ISSUE #13582: Badge status calculation should ignore Excluded results
@hellcp  First, I would like to thank a log for the feature!

As mentioned in the title, the following should not present `50%` (and `unknown` status):
https://build.opensuse.org/package/show/devel:openSUSE:Factory:rpmlint/rpmlint

[build result](https://build.opensuse.org/package/show/devel:openSUSE:Factory:rpmlint/rpmlint)
[build result](https://build.opensuse.org/package/show/devel:openSUSE:Factory:rpmlint/rpmlint)

hellcp-work:
States that aren't succeeded or failed don't really factor into the final badge anyway

Fixes #13582

### DIFF
--- a/src/api/app/models/badge.rb
+++ b/src/api/app/models/badge.rb
@@ -1,10 +1,11 @@
 class Badge
   RED = '#e05d44'.freeze
   GREEN = '#4c1'.freeze
+  FAILED = %w[failed unresolvable broken].freeze
 
   def initialize(type, states)
     @type = 'unknown'
-    finalstates = states&.select { |r| ['succeeded', 'failed'].any?(r.code) }
+    finalstates = states&.select { |r| ['succeeded', *FAILED].any?(r.code) }
     return if finalstates.blank?
 
     # Ratio of the successful to all results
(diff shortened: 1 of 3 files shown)


src/api/app/models/badge.rb
```diff
@@ -2,8 +2,9 @@ class Badge
   RED = '#e05d44'.freeze
   GREEN = '#4c1'.freeze
 
-  def initialize(type, results)
+  def initialize(type, buildresults)
     @type = 'unknown'
+    results = buildresults&.select { |r| ['succeeded', 'failed'].any?(r.code) }
```
danidoni, line 7: I would say this is a `state`. And `succeed` and `failed` are `finalstates`.

saraycp: According to [this line of code](https://github.com/openSUSE/open-build-service/blob/72e9bb9349a89b50815afb73d42221aba30bcb11/src/api/app/models/buildresult.rb#L68), these are considered final status: 'succeeded', 'failed', 'unresolvable', 'broken', 'disabled', 'excluded'

I can understand that we need to ignore "excluded" and "disabled" ones. However, I wonder if we should consider "unresolvable" or "broken" meaning "something wrong happened" as some action from the user is required to make it build. :thinking:

saraycp (approved):

PR #13595, opened 2022-12-23
Add item to watchlist from show view

saraycp:
To make it easier to add a project/package/request to the watchlist, we offer an icon next to its title.

Replace bookmark icon with eye icon for watchlist:

Screenshot 2022-12-27 at 14-32-51 Tender Is the Night

When the item is not on the watchlist yet:

Screenshot 2022-12-27 at 14-31-22 Tender Is the Night

When the item is already on the watchlist:

Screenshot 2022-12-27 at 14-31-05 Open Build Service

### DIFF
--- a/src/api/app/components/watchlist_component.rb
+++ b/src/api/app/components/watchlist_component.rb
@@ -15,6 +15,9 @@ class WatchlistComponent < ApplicationComponent
 
   private
 
+  # Some of the current_objects are not even accesible as they are remote or a multibuild flavor of a package.
+  # This methods return the item which is really accesible.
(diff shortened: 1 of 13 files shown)


saraycp:
> I don't think bookmark is the best way to represent watching, maybe let's go for an eye instead? I would also like to see some non hover text for mobile

I used the same icon we already have in the navigation bar. But I agree, the eye is better, I'll change it.

saraycp:
@hellcp-work I already replaced the bookmark with the eye. I also decided to display "Watch"/"Unwatch" always, no matter the viewport size.

PR #13602, opened 2022-12-28
Update to Bootstrap 5

hellcp-work:
This is quite big, but also not really possible to break up into smaller pieces. I have made sure to split this up into reasonable commits, so that you can check out the individual changes in order, hopefully making this easier to review

Depends on #13617


dmarcoux:
I had only a quick look at a few pages and found some issues. Overall, the UI is slightly different. The tone of colors, the spacing, the font... I don't think we can make everything look just like it did with Bootstrap 4.6. However, some issues will have to be fixed. Here's what I found so far:

Extra spacing in announcements:
extra-space

Whitespace character after icons. I don't know if it's only in the news section:
whitespace

Missing style for datatables:
missing-style-datatables

dmarcoux:
`a.nav-link` previously had padding (`padding: 0.5rem 1rem;`), but it's not the case anymore and this results in:

Bootstrap 4:
before

Bootstrap 5:
now

A bit more spacing between the links would be nice in the Bootstrap 5 version. I've seen this issue on the project#show page, but it's probably on other pages as well.

danidoni:
Looks like the delete file button in the Package show page is not working:
image

danidoni:
> Looks like the delete file button in the Package show page is not working: 
Looks like it's fixed

dmarcoux (approved):

PR #13611, opened 2022-12-30
Avoid errors when description_extension is nil

### LINKED ISSUE #13610: Can't find review when description_extension is nil
Old requests are no longer accessible (in beta and non-beta) because at [this point](https://github.com/openSUSE/open-build-service/commit/904813d7184dc7f049334f390ebce72fd1e6eb4a#diff-3b6c3a1b4733e92a049b76927e330e07932e318e2df919896c1cc17d1c0cbf0eR25)  the code assumes that the `description_extension` contains the ID of a review (as String). But that's not true for old requests, where that field is nil in the DB because their requests' history elements were created before [this line was introduced](https://github.com/openSUSE/open-build-service/commit/02f2764d93e5179333abe26c8a12b44d71faadb0#diff-4dd354307e5161407a24287136777333e1900263ff52e321507a662f60a5ca7bR546).

saraycp:
Fixes #13610

The code assumed the `HistoryElement::RequestReviewAdded#description_extension` always contained a review ID, but that's not true for old requests/history_elements.

When the review ID is present:

Screenshot 2023-01-02 at 11-41-44 Open Build Service

When the review ID is NOT present:

Screenshot 2023-01-02 at 11-42-16 Open Build Service

### DIFF
--- a/src/api/app/models/history_element/request_review_added.rb
+++ b/src/api/app/models/history_element/request_review_added.rb
@@ -5,6 +5,8 @@ module HistoryElement
     end
 
     def user_action
+      return 'added a reviewer' unless review
+
       "#{user_action_prefix} #{action_target} #{user_action_suffix}"
     end
 
@@ -20,8 +22,10 @@ module HistoryElement
       'as a reviewer'
     end
 
-    # self.description_extension is review id
+    # self.description_extension is review id, but it's not present in old history elements
     def review
+      return if description_extension.blank?
+
       ::Review.find(description_extension)
     end
   end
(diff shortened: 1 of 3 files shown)


src/api/app/models/history_element/request_review_added.rb
```diff
@@ -5,6 +5,8 @@ def description
     end
 
     def user_action
+      return 'added a reviewer' if review.nil?
```
saraycp, line 8: This is needed for the old request view which calls `user_action` independently. If review is not present this would have crashed in the old view too.

eduardoj (approved):

PR #13624, opened 2023-01-03
Refactor removing a user

### LINKED ISSUE #13604: Users#index user delete modal doesn't work with low breakpoints
# Issue Description
On lower breakpoints, when datatable collapses on users#index, the user delete modal stops functioning
remove_when_collapsed

Expected Result
============
You can still delete users when the viewport is narrow

How to Reproduce
===============
1. Make your viewport narrow
2. Open the actions dropdown
3. Click delete user button

eduardoj:
Instead of generating html code for each modal that asks for removing a user, only one modal is created. Arguments for the modal are passed with "data-*" attributes.

Fixes #13604.


eduardoj:
> Seems that user-login doesn't get set?

Argh, true. Fixed. Thanks! Please take a look again.

PR #13626, opened 2023-01-04
Badge shim for bootstrap 5 migration

hellcp-work:
This is a part of https://github.com/openSUSE/open-build-service/pull/13602, updating badge classes to bg classes

### DIFF
--- a/src/api/app/assets/stylesheets/webui/patchinfo.scss
+++ b/src/api/app/assets/stylesheets/webui/patchinfo.scss
@@ -22,19 +22,19 @@
       @extend .align-middle;
     }
     &.badge-category {
-      @extend .badge-secondary;
+      @extend .bg-secondary;
 
-      &.recommended { @extend .badge-success; }
-      &.security { @extend .badge-danger; }
-      &.optional { @extend .badge-info; }
+      &.recommended { @extend .bg-success; }
+      &.security { @extend .bg-danger; }
+      &.optional { @extend .bg-info; }
     }
     &.badge-rating {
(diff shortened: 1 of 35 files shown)


src/api/app/assets/stylesheets/webui/patchinfo.scss
```diff
@@ -16,25 +16,25 @@
 
 .patchinfo {
   .badge {
-    &.badge-rating, &.badge-category, &.badge-blocked {
+    &.bg-rating, &.bg-category, &.bg-blocked {
```
eduardoj, line 19:
```suggestion
    &.badge-rating, &.badge-category, &.badge-blocked {
```
I would leave this three `.badge-` classes as they are already defined:
- They don't only modify the background (with the exception of `.badge-blocked`).
- They are only applied to badges.
- They could get confused with the default `.bg-` Bootstrap 5 classes.

src/api/app/assets/stylesheets/webui/requests.scss
```diff
@@ -1,14 +1,14 @@
-.badge-added {
+.bg-added {
```
eduardoj, line 1: For the cases of `.badge-added`, `.badge-deleted` and `.badge-changed`, kind of the same as pointed out before:
- They don't only modify the background.
- They are only applied to badges.
- They could get confused with the default .bg- Bootstrap 5 classes.

eduardoj (approved):

dmarcoux:
Following the merge of this PR, the contrast between the badge background color and its text is quite bad. See:

contrast

PR #13629, opened 2023-01-04
Add table shim for bootstrap 5 migration

hellcp-work:
This is a part of https://github.com/openSUSE/open-build-service/pull/13602, updating table classes

### DIFF
--- a/src/api/app/assets/stylesheets/webui/bs5-shims.scss
+++ b/src/api/app/assets/stylesheets/webui/bs5-shims.scss
@@ -1,3 +1,5 @@
 .visually-hidden {
   @extend .sr-only;
 }
+
+.table-light { @extend .thead-light; }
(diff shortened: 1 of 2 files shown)


eduardoj (approved):

dmarcoux (approved):

dmarcoux:
Oh, this will have to rebased. I think to avoid conflicts, we could have a directory `src/api/app/assets/stylesheets/webui/bs5-shims/` which would contain various files for all the Bootstrap 5 shims. This way, we don't have to constantly rebase PRs whenever one is merged.

dmarcoux:
> That would just move what would have to be rebased, it would either be an index.scss file in that directory or this file we have now

Oh right, I didn't take into consideration that Sprockets always need to import files one by one...

PR #13637, opened 2023-01-04
Allow users to remove themselves from a group

### LINKED ISSUE #13406: Users should be able to remove themselves from a group
Users cannot remove themselves from a group they belong to, unless they are a maintainer of this group. They then need to ask one of the maintainers to remove them from the group.

Any user should be able to remove themselves from a group, even without the maintainer role.

Nothing... the current state is simply counter-intuitive.

**Additional context**
I created this following a question from @tserong in SUSE's internal chat.

dmarcoux:
Even if they aren't an admin or a maintainer of the group.

Closes #13406

Example - While logged in as _Iggy_, I can remove myself from the group. The alignment is a bit off in this case, but I didn't want to tackle this in this PR. 
preview

### DIFF
--- a/src/api/app/policies/groups_user_policy.rb
+++ b/src/api/app/policies/groups_user_policy.rb
@@ -1,0 +1,6 @@
+class GroupsUserPolicy < ApplicationPolicy
+  def destroy?
+    # An admin, a group maintainer or the user themselves can remove themselves from a group
+    user.is_admin? || record.group.group_maintainers.exists?(user: user) || user.id == record.user_id
+  end
+end
--- a/src/api/app/views/webui/groups/_group_members.html.haml
+++ b/src/api/app/views/webui/groups/_group_members.html.haml
@@ -8,13 +8,13 @@
     .d-flex.flex-wrap
       - group.users.each do |user|
         .card.m-1.p-2.group-user
-          - if write_access
+          - if policy(GroupsUser.new(group: group, user: user)).destroy?
             = link_to('#',
-                       title: 'Remove member from group',
-                       data: { toggle: 'modal',
-                               target: '#delete-group-members-modal',
-                               confirmation_text: "Please confirm deletion of '#{user}' from this group",
-                               action: group_user_path(group_title: group.title, user_login: user.login) }) do
+                      title: 'Remove member from group',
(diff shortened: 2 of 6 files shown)


src/api/app/views/webui/groups/_group_members.html.haml
```diff
@@ -42,9 +42,9 @@
           Add Member
   = render partial: 'add_group_user_modal', locals: { group: group }
 
-  = render DeleteConfirmationDialogComponent.new(modal_id: 'delete-group-members-modal',
-                                                 method: :delete,
-                                                 options: { modal_title: 'Do you want to remove this group member?' })
+= render DeleteConfirmationDialogComponent.new(modal_id: 'delete-group-members-modal',
```
dmarcoux, line 45: This change was needed for the modal to work when the logged-in user isn't a group member or an admin, so `write_access` on line 36 is `false`.

dmarcoux:
[Allow users to remove themselves from a group](https://trello.com/c/so37LwmY/2089-allow-users-to-remove-themselves-from-a-group)

src/api/app/policies/groups_user_policy.rb
```diff
@@ -0,0 +1,6 @@
+class GroupsUserPolicy < ApplicationPolicy
```
eduardoj, line 1: Do we really need to create a GroupsUserPolicy? I wonder if we could modify the `destroy?` method of GroupPolicy, here: https://github.com/openSUSE/open-build-service/blob/e96f6256a1de2ea10ccf58ffaa3f52173e1e8e32/src/api/app/policies/group_policy.rb#L15-L17

dmarcoux: `GroupPolicy#destroy?` is about deleting groups. `GroupsUserPolicy#destroy?` is about removing users from a group. Those are two different actions, so this is why we have this new `GroupsUserPolicy` now.

eduardoj: I see. Thanks for the explanation.

eduardoj (approved):

dmarcoux:
I had to rebase to fix a conflict following a merge on master.

eduardoj (approved):

Issue #13574, opened 2022-12-21
Align naming around Status Messages and News


dmarcoux:
Right now, we either use Status Message or News to refer to the messages under the `News` section on the homepage. This should always be using the same name. I suggest going with `News`. This involves renaming routes, models, controllers, etc...

PR #13645, opened 2023-01-05
Consistently name news accross the UI

dmarcoux:
Part of #13574

Renaming the internals like the model, controller and more is intentionally left out of this.


dmarcoux:
> Maybe a better way to express it would be a `news item` instead of `news` when referring to an instance?

I don't really have a preference for this. Do you mean replacing something like `Create News` with `Create News Item`, and so on?

dmarcoux:
Please have another look @hellcp-work

dmarcoux:
[Consistently name news accross the UI](https://trello.com/c/sjw20p23/2090-consistently-name-news-accross-the-ui)

PR #13647, opened 2023-01-05
Correct the badge shim

hellcp-work:
I missed a character and the style wasn't applied correctly, this should fix the issue


dmarcoux:
This looks better now :slightly_smiling_face:

badges

eduardoj (approved):

PR #13653, opened 2023-01-05
Replace btn-block with d-block for bootstrap 5 migration

hellcp-work:
https://getbootstrap.com/docs/5.0/migration/#buttons


eduardoj:
It seems `d-block` in Bootstrap 4 doesn't apply `width: 100%`.

Here you have an example comparing `btn-block` and `d-block`.

Screenshot from 2023-01-05 16-04-51

Will `d-block` apply `width: 100%` in Bootstrap 5?

eduardoj (approved):

PR #13655, opened 2023-01-05
Remove check of database in CircleCI

eduardoj:
This check is not waiting for the database to be up.

It has not been an issue that the database is not up and not ready to run the tests.


dmarcoux:
That doesn't seem to have even worked before like in this CI run here:

ci

dmarcoux (approved):

PR #13660, opened 2023-01-06
Remove the media component for bootstrap 5 migration

hellcp-work:
Removes media component entirely from the application


src/api/app/views/webui/request/_request_history.html.haml
```diff
@@ -1,7 +1,8 @@
-.media
```
dmarcoux, line 1: I only looked at the changes in this view, but I would assume they are all the same since the same pattern was applied over and over again.

In the new version, the gap between the image and the text is bigger. See below.

Before:
before

Now:
after

dmarcoux (approved):

PR #13661, opened 2023-01-06
Remove accordion class from workflow_runs

hellcp-work:
This class doesn't really make sense to be there


dmarcoux (approved):

dmarcoux:
> This class doesn't really make sense to be there

True, it doesn't change anything since the `accordion` element doesn't have any children elements with the `.collapse` class.

PR #13667, opened 2023-01-09
Overwrite the badge colours in staging workflows

### LINKED ISSUE #13665: Painful new color scheme on staging dashboard
On Jan 9 2023, the color scheme of the staging dashboard, for requests 'with pending reviews' has been changed from 'blue background' (review) to 'yellow background' (building)

Let alone that the blue background matched better the state of the request, the new palette is very unfriendly on the eye:

image

hellcp-work:
This fixes #13665, the background colour was overwritten by the .bg-warning class extends otherwise

### DIFF
--- a/src/api/app/assets/stylesheets/webui/staging-workflow.scss
+++ b/src/api/app/assets/stylesheets/webui/staging-workflow.scss
@@ -134,17 +134,17 @@ ul .table-list-group-item {
       }
 
       &.state-ready {
-        @extend .bg-success;
+        color: white;
         background-color: $request-ready-color;
       }
 
       &.state-review {
-        @extend .bg-warning;
+        color: white;
         background-color: $request-review-color;
       }
 
       &.state-obsolete {
-        @extend .bg-info;
+        color: white;
         background-color: $request-obsolete-color;
       }
 


src/api/app/assets/stylesheets/webui/staging-workflow.scss
```diff
@@ -135,17 +135,17 @@ ul .table-list-group-item {
...
       &.state-ready {
         @extend .bg-success;
-        background-color: $request-ready-color;
+        background-color: $request-ready-color !important;
       }
 
       &.state-review {
         @extend .bg-warning;
-        background-color: $request-review-color;
+        background-color: $request-review-color !important;
```
dmarcoux, line 143: `!important` is effectively reverting what the previous line `@extend .bg-warning` does, so we can instead remove the previous line and avoid using `!important`.

dmarcoux, line 143: `!important` is effectively reverting what the previous line `@extend .bg-warning` does, so we can instead remove the previous line and avoid using `!important`.

dmarcoux (changes requested):

src/api/app/assets/stylesheets/webui/staging-workflow.scss
dmarcoux: Okay, I would rather have an explicit font color and background color than extending a class. This is much clearer. It's also what Bootstrap's documentation suggests here: https://getbootstrap.com/docs/5.0/components/badge/#background-colors

PR #13669, opened 2023-01-10
Revert "[backend] forbid _project as package name in verify_pack()"

hellcp-work:
This reverts commit bd705eb378855c231d7901b70274f9258c1293cf.

We found out that it breaks workflows code in https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/workflow/step/link_package_step.rb#L70-L75 and we don't currently have a better workaround

### DIFF
--- a/src/backend/BSVerify.pm
+++ b/src/backend/BSVerify.pm
@@ -306,7 +306,6 @@ sub verify_proj {
 sub verify_pack {
   my ($pack, $packid) = @_;
   if (defined($packid)) {
-    die("illegal package name '$packid'\n") if $packid eq '_project';
     die("name does not match data\n") unless $packid eq $pack->{'name'};
   }
   verify_projid($pack->{'project'}) if exists $pack->{'project'};


danidoni (approved):

mlschroe:
You must not send PUT _project.xml requests to the backend. Please fix this!

mlschroe:
I'm also not sure what you're trying to do. You can write the _service file without "creating" a _project package.

adrianschroeter:
creating a _project package leads also invalid database entries which we keep need to fix manually in our production systems.

better disable the workflow code instead as temporary solution,.

mlschroe:
You could directly call Backend::Api::Sources::Package.write_file instead of creating a _project package.

(The correct way would be to add a method to the project model that does the call.)

mlschroe:
And maybe a write_service method in app/lib/backend/api/sources/project.rb

hennevogel:
We are doing this because you need "project services" defined to be able to run services on linked packages. We would need to be able to do this otherwise then, any suggestions?

hennevogel:
I guess we can just `Backend::Api::Sources::Package.write_file` a _service file to a project @mlschroe ?

mlschroe:
Yes, that's what I suggested in my last comment (except that there is no write_file)

PR #13674, opened 2023-01-11
Allow project restoring when package name is invalid

### LINKED ISSUE #13544: [production][project#restore] 500 error when restoring home project
# Issue Description

My home project was [auto-deleted](https://build.opensuse.org/request/show/1013494) a couple of months ago, and I only just noticed.  I was relieved to be given an option to restore it from [the page for (re)creating a home project](https://build.opensuse.org/project/new?name=home%3Aaspiers&restore_option=true) (thanks for being kind enough to implement this very considerate feature!).  But unfortunately when I tried it, I got a 500 error page.

I tried again a few times in case this was a glitch, and after roughly the third attempt, the error changed, and now I'm getting an inline notification instead of the separate 500 page:

image

Expected Result
============
I expected to get my original `home:aspiers` project back with all the old packages.

How to Reproduce
===============
1. Login to an account with a home project which was auto-deleted
2. Click on "Create Your Home Project"
3. Click on "[Click here](https://build.opensuse.org/project/restore?project=home%3Aaspiers) if you want to restore the project"
4. (Probably) observe the same 500 error page I got.

Further Information
===============

https://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/639cff2184bf460521649a7a

saraycp:
In [this issue](https://github.com/openSUSE/open-build-service/issues/13544), the project we are trying to restore contains one package called `_meta` that makes the restore code crash because `package.meta.content` is nil in [this line](https://github.com/openSUSE/open-build-service/blob/1989d07ba15697ad561fc53a3d66956d0c555ef7/src/api/app/models/project.rb#L145).

Package names like `_meta` are now forbidden in frontend and backend code. The one in the issue was probably created before we introduced the validations.

We should be able to restore the rest of the packages and the project by ignoring packages whose names are invalid.

Fixes #13544

*NOTE:* it's not worth adding a spec as it requires skipping validations from both [the model](https://github.com/openSUSE/open-build-service/blob/ff93a673a346ad5ad9f3e40061031425f4dc59cb/src/api/app/models/package.rb#L108) and [the backend](https://github.com/openSUSE/open-build-service/blob/ff93a673a346ad5ad9f3e40061031425f4dc59cb/src/backend/BSVerify.pm#L43), and then recording a cassette with invalid data. I did it manually and got the expected behaviour (the project was restored with only the valid packages on it).

### DIFF
--- a/src/api/app/models/project.rb
+++ b/src/api/app/models/project.rb
@@ -141,6 +141,9 @@ class Project < ApplicationRecord
         # restore all package meta data objects in DB
         backend_packages = Xmlhash.parse(Backend::Api::Search.packages_for_project(project_name))
         backend_packages.elements('package') do |package|
+          # Restoring packages with invalid names can cause issues, we ignore them.
+          next unless Package.valid_name?(package['name'])
+
           package = project.packages.new(name: package['name'])
           package_meta = Xmlhash.parse(package.meta.content)
 


dmarcoux:
There's a small typo in the commit message: re**s**toring instead of retoring.

src/api/app/models/project.rb
```diff
@@ -142,6 +142,9 @@ def restore(project_name, backend_opts = {})
         backend_packages = Xmlhash.parse(Backend::Api::Search.packages_for_project(project_name))
         backend_packages.elements('package') do |package|
           package = project.packages.new(name: package['name'])
+          # Some packages don't have a proper meta content and prevent the project to be restored
```
dmarcoux, line 145:
```suggestion
          # Some packages don't have a proper meta content and prevent the project from being restored
```

src/api/app/models/project.rb
```diff
@@ -142,6 +142,9 @@ def restore(project_name, backend_opts = {})
...
+          next if package.meta.content.blank?
```
hennevogel, line 146: You should probably check more like if the `Package.valid_name?`

saraycp:
@dmarcoux @hennevogel I pushed the changes. I tried to add a test as well, but I think it's not worth it. (Note also added to the description).

> *NOTE:* it's not worth adding a spec as it requires skipping validations from both the model and the backend, and then recording a cassette with invalid data. I did it manually and got the expected behaviour (the project was restored with only the valid packages on it).

PR #13678, opened 2023-01-12
Fix duplicated monitor package headers

### LINKED ISSUE #13677: Monitor page is misrendered
https://build.opensuse.org/project/monitor/openSUSE:Factory

In the first column all table entries are rendered twice.

hellcp-work:
This fixes #13677, displaying the fixed row correctly

Screenshot from 2023-01-12 12-31-41


saraycp (approved):
Works for me. Tested in different viewport sizes and using scroll as well

PR #13681, opened 2023-01-12
Improve UX of new request#show

dmarcoux:
For reviewers, see https://obs-reviewlab.opensuse.org/dmarcoux-request-multiple-actions-hint/request/show/9

Before:
before

After:
after

### DIFF
--- a/src/api/app/views/webui/request/_action_text.html.haml
+++ b/src/api/app/views/webui/request/_action_text.html.haml
@@ -4,7 +4,7 @@
     %p.mt-0.mb-0.fw-normal
       %span.me-2 ##{action_index + 1}
       = details['comment'].truncate(50)
-    %p.mb-0.mt-0.text-muted.fw-light.small
+    %p.mb-0.mt-0.small
       = action.name
   - else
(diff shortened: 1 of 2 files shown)


dmarcoux:
[Request action dropdown improvements (L)](https://trello.com/c/usNPBwRc/2097-request-action-dropdown-improvements-l)

danidoni:
I think the dropdown should display the index of the action, otherwise telling the user is looking at action number 12 without any other reference leaves the user to count the actions from the top. For requests with a couple of actions there is no problem, but with actions with more than 10 actions it can be disorienting...

dmarcoux:
@hellcp-work @danidoni I pushed changes to address the issues you mentioned in your comments. Here's a preview of the changes. Once with some revision comments and once without:

changes
changes-2

dmarcoux:
> Revision looks missing in those

It's because the revisions aren't there. I faked it in the other screenshot. I guess removing the `Revision` header could be a solution, but it involves checking all details of every action...

danidoni (approved):

PR #13682, opened 2023-01-12
Display information about the staging workflow in request

hellcp-work:
Currently when a request gets a review for a staging workflow project, a note is displayed saying that the request is staged in that staging workflow project. This PR expands that a little bit, by displaying that the request is unstaged at first, when submitted to a staging project, staged when the staging manager group accepts the review and staged to a staging workflow project when there's a review for that project.

Staging
Staged
Unstaged

### DIFF
--- a/src/api/app/assets/stylesheets/webui/requests.scss
+++ b/src/api/app/assets/stylesheets/webui/requests.scss
@@ -13,6 +13,14 @@
   background-color: $cyan;
 }
 
+.badge-staging {
+  color: $white;
+  background-color: $pink;
+  &:hover {
+    color: $gray-100;
+  }
+}
+
 .req-description-box {
   word-break: break-all;
   white-space: pre-line;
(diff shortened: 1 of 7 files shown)


saraycp (commented):
I like the texts you propose, they are more meaningful.
I also like how you try to make the three different cases clear: Staged, Unstaged, Staging:A.

However I'm not sure if it's a good idea to have an extra badge there, where we could have a total of 4 badges: [state] [staging info] [auto-accept] [watch].
If there isn't another better place, at least, I wouldn't use such a common color (green), maybe some color that represents staging stuff only?

saraycp:
I see. And as it's unlikely that both [auto-accept] and [staging info] coexist, I'm fine with keeping the badge.
What about using another color?

From the [color variables](https://github.com/openSUSE/open-build-service/blob/92df2e8b8c5feb626fc6015b5ed5d14165c6b62e/src/api/app/assets/stylesheets/webui/bootstrap_variables/colors.scss#L11) we already have, I'd use pink (which looks more purple than pink, TBH). But I like how it looks like:

Screenshot 2023-01-13 at 13-34-57 Open Build Service

PR #13683, opened 2023-01-12
Update rake task to create many request actions

saraycp:
In order to improve the request show page we need to deal with many actions in the `Select Action` dropdown.

Run `rails dev:requests:multiple_actions_request` locally to create a request with 8 submit actions.

### DIFF
--- a/src/api/lib/tasks/dev/requests.rake
+++ b/src/api/lib/tasks/dev/requests.rake
@@ -44,15 +44,20 @@ namespace :dev do
         source_package: source_package_a
       )
 
-      # Create an action to submit a new file to package B
-      action_attributes = {
-        source_package: source_package_b,
-        source_project: source_project,
-        target_project: target_project,
-        target_package: target_package_b
-      }
-      bs_req_action = build(:bs_request_action, action_attributes.merge(type: 'submit', bs_request: request))
-      bs_req_action.save! if bs_req_action.valid?
+      # Create more actions to submit a new file to different packages
+      (1..30).each do |i|
+        target_package = Package.where(name: "package_#{i}", project: target_project).first || create(:package, name: "package_#{i}", project: target_project)
+
+        # Create an action to submit a new file to package C
+        action_attributes = {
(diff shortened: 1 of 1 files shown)


dmarcoux:
Would you be against increasing the number of actions? Like 60? This would reproduce what we're seeing in production sometimes with, for example, requests in the KDE project.

saraycp:
> Would you be against increasing the number of actions? Like 60? This would reproduce what we're seeing in production sometimes with, for example, requests in the KDE project.

Is 30 a good tradeoff?

dmarcoux:
> > Would you be against increasing the number of actions? Like 60? This would reproduce what we're seeing in production sometimes with, for example, requests in the KDE project.
> 
> Is 30 a good tradeoff?

It's fine :+1:.

src/api/lib/tasks/dev/requests.rake
```diff
@@ -44,15 +44,20 @@ namespace :dev do
...
-        source_package: source_package_b,
-        source_project: source_project,
-        target_project: target_project,
-        target_package: target_package_b
-      }
-      bs_req_action = build(:bs_request_action, action_attributes.merge(type: 'submit', bs_request: request))
-      bs_req_action.save! if bs_req_action.valid?
+      # Create more actions to submit a new file to different packages
+      30.times.each do |i|
+        target_package = Package.where(name: "package_#{i}", project: target_project).first || create(:package, name: "package_#{i}", project: target_project)
```
dmarcoux, line 49: This can be a bit simplified:

```suggestion
        target_package = Package.find_or_create_by(name: "package_#{i}", project: target_project)
```

saraycp: There is a difference, in the current code we use the Factory to create the package if it didn't exist.

dmarcoux: Oh right, that's a key difference I overlooked.

dmarcoux (approved):

PR #13693, opened 2023-01-13
Don't underline urls

llyyr:
For consistency with how it used to be prior to the move to Bootstrap 5.


dmarcoux (approved):
Thank you for the PR. It's true that we used to not underline URLs before the update to Bootstrap 5. Let's stay consistent with how it was. Links are already emphasized with colors, a different and cursor on hovering.

dmarcoux:
@llyyr The changes from this PR are now live on build.opensuse.org.

PR #12563, opened 2022-05-11
[api] fix release behaviour when using token

adrianschroeter:
The token should also support the automatic time extension when
releasing into maintenance_release projects

PR #13701, opened 2023-01-16
Fix and extend release mechanics

adrianschroeter:
This is replacing https://github.com/openSUSE/open-build-service/pull/12563, had to skip a commit and needed a rebase

### DIFF
--- a/src/api/app/models/token/release.rb
+++ b/src/api/app/models/token/release.rb
@@ -47,7 +47,8 @@ class Token::Release < Token
     # releasing ...
     manual_release_targets.each do |release_target|
       next if options[:filter_source_repository].present? && options[:filter_source_repository] == release_target.repository.name
-      release_target.repository.check_valid_release_target!(release_target.target_repository, opts[:filter_architecture])
+
+      release_target.repository.check_valid_release_target!(release_target.target_repository, options[:arch])
       release(package_to_release, release_target.repository, release_target.target_repository, time_now, options)
     end
   end
--- a/src/api/app/models/token/release.rb
+++ b/src/api/app/models/token/release.rb
@@ -5,26 +5,49 @@ class Token::Release < Token
     set_triggered_at
     return unless options[:package]
 
+    # uniq timestring for all targets
+    time_now = Time.now.utc
+
     # FIXME: Take repository and arch into account
     package_to_release = options[:package]
+    if options[:targetproject].present? && options[:targetrepository].present? && options[:filter_source_repository].present?
+      source_repository = Repository.find_by_project_and_name(options[:project].name, options[:filter_source_repository])
+      target_repository = Repository.find_by_project_and_name(options[:targetproject], options[:targetrepository])
+      raise InsufficientPermissionOnTargetRepository, "no permission to write in project #{target_repository.project.name}" unless User.session!.can_modify?(target_repository.project)
+
+      release(package_to_release, source_repository, target_repository, time_now, options)
+      return
+    end
+
+    release_manually_triggered_targets(package_to_release, time_now, options)
+  end
+
+  def package_find_options
+    { use_source: true, follow_project_links: false, follow_multibuild: true }
+  end
+
+  private
+
+  def release(package_to_release, source_repository, target_repository, time_now, options)
+    opts = { filter_source_repository: source_repository,
+             manual: true,
+             comment: 'Releasing via trigger event' }
+    opts[:multibuild_container] = options[:multibuild_flavor] if options[:multibuild_flavor].present?
+    opts[:filter_architecture] = options[:arch] if options[:arch].present?
+    release_package(package_to_release,
+                    target_repository,
+                    package_to_release.release_target_name(target_repository, time_now),
+                    opts)
+  end
+
+  def release_manually_triggered_targets(package_to_release, time_now, options)
     manual_release_targets = package_to_release.project.release_targets.where(trigger: 'manual')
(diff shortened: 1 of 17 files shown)


src/api/public/apidocs-new/paths/trigger_operation.yaml
```diff
@@ -116,6 +116,65 @@ post:
           value: ppc64le
         x86_64:
           value: x86_64
+    - in: query
+      name: targetproject
+      description: |
+        Only used when the operation is 'release'.
+
+        Release binaries to the provided target project. Only works in combination
```
hennevogel, line 124:
```suggestion
        Release binaries only to the provided target project. Only works in combination
```

src/api/public/apidocs-new/paths/trigger_operation.yaml
```diff
@@ -116,6 +116,65 @@ post:
...
+          summary: '- Empty -'
+          value:
+        home:user1:
+          value: home:user1
+    - in: query
+      name: targetrepository
+      description: |
+        Only used when the operation is 'release'.
+
+        Release binaries to the provided target repository. Only works in combination
```
hennevogel, line 139:
```suggestion
        Release binaries only to the provided target repository. Only works in combination
```

src/api/public/apidocs-new/paths/trigger_operation.yaml
```diff
@@ -116,6 +116,65 @@ post:
...
+        openSUSE_Factory:
+          value: openSUSE_Factory
+        SLE_15_SP3:
+          value: SLE_15_SP3
+    - in: query
+      name: filter_source_repository
+      description: |
+        Only used when the operation is 'release'.
+
+        Release binaries from the provided source repository. Only works in combination
```
hennevogel, line 156:
```suggestion
        Release binaries only from the provided source repository. Only works in combination
```

hennevogel (changes requested):

src/api/app/models/token/release.rb
```diff
@@ -5,27 +5,50 @@ def call(options)
...
+    time_now = Time.now.utc
+
     # FIXME: Take repository and arch into account
     package_to_release = options[:package]
-    manual_release_targets = package_to_release.project.release_targets.where(trigger: 'manual')
-    raise NoReleaseTargetFound, "#{package_to_release.project} has no release targets that are triggered manually" unless manual_release_targets.any?
+    if options[:targetproject].present? && options[:targetrepository].present? && options[:filter_source_repository].present?
+      source_repository = Repository.find_by_project_and_name(options[:project].name, options[:filter_source_repository])
+      target_repository = Repository.find_by_project_and_name(options[:targetproject], options[:targetrepository])
+      raise CmdExecutionNoPermission, "no permission to write in project #{target_repository.project.name}" unless User.session!.can_modify?(target_repository.project)
```
krauselukas, line 16: `CmdExecutionNoPermission` is not available in this class, it throws an `uninitialized constant Token::Release::CmdExecutionNoPermission` error

krauselukas (changes requested):

src/api/app/models/token/release.rb
krauselukas: I can fix it

src/api/app/models/token/release.rb
```diff
@@ -5,27 +5,50 @@ def call(options)
...
     return unless options[:package]
 
+    # uniq timestring for all targets
+    time_now = Time.now.utc
+
     # FIXME: Take repository and arch into account
     package_to_release = options[:package]
-    manual_release_targets = package_to_release.project.release_targets.where(trigger: 'manual')
-    raise NoReleaseTargetFound, "#{package_to_release.project} has no release targets that are triggered manually" unless manual_release_targets.any?
+    if options[:targetproject].present? && options[:targetrepository].present? && options[:filter_source_repository].present?
```
krauselukas, line 13: @adrianschroeter I realized that if one of the params is missing we skip this `if` block and execute `release_manually_triggered_targets`. I guess we should return in this case, because a user might just have forgotten to provide one of those params and doesn't want to default to execute the other set manuel release targets

krauselukas: I will create a extra commit for this, so you can decide if it makes sense

src/api/app/models/token/release.rb
krauselukas: Fixed

src/api/app/models/token/release.rb
adrianschroeter: the options also be used to filter pre-defined release definitions from the project... I have not checked, but that might break that (so there may be a usecase to have only some of them defined)

krauselukas: @adrianschroeter okay, if thats intended I will remove the commit and adapt the apidocs and specs to reflect that.

SchoolGuy:
@hennevogel & @adrianschroeter could we progress this PR, please? The CLI is waiting for the server side implementation. The automations started by @lethliel are dependent in the mid-term on the merge of this PR. Without it, the authentication on our Botmaster instances for openSUSE and SUSE will not satisfy the desired level of security in these environments. Sadly my exchange ended and I am not able to invest work time but I am willing to sponsor the CLI implementation in my free time.

hennevogel:
> This is in production since a long time (on IBS where it is needed), @lethliel can just continue, no matter of the state of this pull request.

That doesn't mean you don't have to address feedback or CI right? If you need help, let us know...

hennevogel (approved):

krauselukas (approved):

PR #13702, opened 2023-01-16
Mark bs_request_action as seen by user

ncounter:
https://trello.com/c/OVu6ybaT

Let the user mark an action request as seen or not in order to be able to not load it back if he already viewed it.

action-as-seen-in-select-checkbox-improved

### DIFF
--- a/src/api/app/components/action_seen_by_user_component.html.haml
+++ b/src/api/app/components/action_seen_by_user_component.html.haml
@@ -1,0 +1,12 @@
+- button_text = 'Action seen'
+
+= link_to(toggle_request_path(action_id: @action.id),
+    class: 'btn btn-sm btn-outline-info',
+    title: 'Wether the action has been marked "as seen" or not yet',
+    method: :put, remote: true,
+    'data-disable-with': tag.i(class: 'fa-solid fa-spinner fa-spin') + button_text) do
+  - if seen_by_user
+    %i.fa-solid.fa-square-check
+  - else
+    %i.fa-regular.fa-square
+  = button_text
--- a/src/api/app/components/action_seen_by_user_component.html.haml
+++ b/src/api/app/components/action_seen_by_user_component.html.haml
@@ -1,12 +1,13 @@
-- button_text = 'Action seen'
+- if @render_only
+  = render_icon_status
 
-= link_to(toggle_request_path(action_id: @action.id),
-    class: 'btn btn-sm btn-outline-info',
-    title: 'Wether the action has been marked "as seen" or not yet',
-    method: :put, remote: true,
-    'data-disable-with': tag.i(class: 'fa-solid fa-spinner fa-spin') + button_text) do
-  - if seen_by_user
-    %i.fa-solid.fa-square-check
-  - else
-    %i.fa-regular.fa-square
-  = button_text
+- else
+  - button_text = 'Action seen'
+
+  = link_to(toggle_request_path(action_id: @action.id),
+      class: 'btn btn-sm btn-outline-info',
+      title: 'Whether the action has been marked "as seen" or not yet',
+      method: :put, remote: true,
+      'data-disable-with': tag.i(class: 'fa-solid fa-spinner fa-spin') + button_text) do
+    = render_icon_status
+    = button_text
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -16,6 +16,8 @@ class BsRequestAction < ApplicationRecord
   belongs_to :target_package_object, class_name: 'Package', foreign_key: 'target_package_id', optional: true
   belongs_to :target_project_object, class_name: 'Project', foreign_key: 'target_project_id', optional: true
 
+  has_and_belongs_to_many :seen_by_users, class_name: 'User', join_table: :bs_request_actions_seen_by_users
+
   scope :bs_request_ids_of_involved_projects, ->(project_ids) { where(target_project_id: project_ids).select(:bs_request_id) }
   scope :bs_request_ids_of_involved_packages, ->(package_ids) { where(target_package_id: package_ids).select(:bs_request_id) }
   scope :bs_request_ids_by_source_projects, ->(project_name) { where(source_project: project_name).select(:bs_request_id) }
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -838,7 +838,7 @@ class BsRequestAction < ApplicationRecord
   end
 
   def toggle_seen_by(user)
-    return unless user
+    return unless user.is_a?(User)
 
     if seen_by_users.exists?({ id: user.id })
       seen_by_users.destroy(user)
--- a/src/api/app/models/bs_request_action.rb
+++ b/src/api/app/models/bs_request_action.rb
@@ -837,6 +837,16 @@ class BsRequestAction < ApplicationRecord
     package.commit(source_rev) || package.commit
   end
 
+  def toggle_seen_by(user)
+    return unless user
+
+    if seen_by_users.exists?({ id: user.id })
+      seen_by_users.destroy(user)
+    else
+      seen_by_users << user
+    end
+  end
+
   private
 
   def cache_diffs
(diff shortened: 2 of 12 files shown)


src/api/db/schema.rb
```diff
@@ -190,6 +190,12 @@
     t.index ["target_project_id"], name: "index_bs_request_actions_on_target_project_id"
   end
 
+  create_table "bs_request_actions_seen_by_users", id: false, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
```
dmarcoux, line 193: There is a change missing in the schema. The line `ActiveRecord::Schema[7.0].define(version: ...)` should have also changed when you ran the migration locally. This needs to be committed.

src/api/db/schema.rb
```diff
@@ -190,6 +190,12 @@
...
+    t.bigint "bs_request_action_id", null: false
+    t.bigint "user_id", null: false
```
dmarcoux, line 195: Those columns should be integer, not bigint. This implies a change in the migration creating this table.

src/api/app/views/webui/request/_actions_details.html.haml
```diff
@@ -23,6 +23,9 @@
                                                                         diff_to_superseded: diff_to_superseded_id),
                                                       class: "dropdown-item #{action_item.id == active_action.id ? 'active' : ''}",
                                                       'aria-current': 'true')
+    - if User.session
```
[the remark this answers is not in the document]
ncounter: Ideally yes, but there is no real good place at the moment from my POV.
Placing it on a new line after the action description is something that looks really weird.

Actually I'd better refactor the `Select Action` dropdown to display the initial part of the title of the selected one. With that, having the `Action seen` check right after the dropdown would be perfectly related. But this is for another PR probably. What do you think?

src/api/db/schema.rb
ncounter: Done.

src/api/app/views/webui/request/_actions_details.html.haml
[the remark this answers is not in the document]
ncounter: Yeah, I agree. I'll start by moving it for now. As soon as some better options comes up, we can always change it.

Btw yes it can be related to the card you linked above, and also to this one https://trello.com/c/1Pg8ZhVO

danidoni:
Hmmm, the placing of that checkbox is a bit weird in my opinion... Would it be better to place it next to the action name? At the end of the `Showing action ...`?

saraycp (commented):
I know the position is temporary but it happens to look even worst when some information about staging is displayed:

Screenshot 2023-01-20 at 17-21-48 Open Build Service

Moreover, the button doesn't follow the style of other elements we have. What about getting inspiration from this button on the notifications page?
Screenshot 2023-01-20 at 17-28-50 Open Build Service

Maybe something like this? (alignment, spaces etc must be adapted, of course):
Request-9-Submit-package_a-Open-Build-Service(4)

saraycp:
I wonder if something like this can help us to display all that information:

Screenshot 2023-01-20 at 18-21-44 Button group
https://getbootstrap.com/docs/5.0/components/button-group/#nesting

Screenshot 2023-01-20 at 18-21-29 Input group
https://getbootstrap.com/docs/5.0/forms/input-group/#segmented-buttons

ncounter:
> I wonder if something like this can help us to display all that information:
> 
> 
> 
> Screenshot 2023-01-20 at 18-21-44 Button group
> 
> https://getbootstrap.com/docs/5.0/components/button-group/#nesting
> 
> 
> 
> Screenshot 2023-01-20 at 18-21-29 Input group
> 
> https://getbootstrap.com/docs/5.0/forms/input-group/#segmented-buttons

I'd really love something like the latest one: a select box with the search integrated and possibly the selected one visible as a value.

Thanks also for the pointers to the existing similar objects for the button, I'll take them for inspiration to look similar as much as possible.

Btw, bottom line consideration for this was: I didn't want to refactor or change anything in this PR context, I'd try to just add the check in the less intrusive and ugly way and that's it. Any improvement will go in a separate PR and iteration, and we already have a lot of cool ideas here. Thank you very much for that.

ncounter:
> ...
> Maybe something like this? (alignment, spaces etc must be adapted, of course): Request-9-Submit-package_a-Open-Build-Service(4)

@saraycp I addressed your suggestion and updated the PR description with a short recording, could you please take another look at it? Thank you very much

saraycp:
> I'd try to just add the check in the less intrusive and ugly way and that's it. Any improvement will go in a separate PR and iteration

I'm totally fine with this :+1:

src/api/app/views/webui/request/_actions_details.html.haml
```diff
@@ -25,6 +27,10 @@
                                                       'aria-current': 'true')
 %p
   - if submit_actions_count > 1
+    - if User.session
+      .d-inline#action-seen-by-user-wrapper
+        = render ActionSeenByUserComponent.new(action: active_action, user: User.session!)
+
     - active_action_index = submit_actions.find_index(active_action) + 1
     Showing
     %em action ##{active_action_index}
```
saraycp, line 36: I'd propose a little change here to improve alignment and spacing:

```   
    %span.ms-2.align-middle
      Showing
      %em action ##{active_action_index}
      \-
  %span.font-italic.align-middle= request_action_header(action, bs_request.creator)    
```

(sorry, GH doesn't allow me to add a "Suggestion" beyond the lines you modified.) 

What I propose is:
- Wrap the "Showing action #N" part with a `span` which adds `.ms-2.align-middle`
- Add `.align-middle` to the `request_action_header` as well.


**Before:**

Screenshot 2023-01-23 at 10-42-41 Open Build Service

**After:**

Screenshot 2023-01-23 at 10-45-59 Open Build Service

src/api/app/components/action_seen_by_user_component.html.haml
```diff
@@ -0,0 +1,19 @@
...
+    %i.fa-solid.fa-square-check
+  - else
+    %i.fa-regular.fa-square
+
+- else
+  - button_text = 'Action seen'
+
+  = link_to(toggle_request_path(action_id: @action.id),
+      class: 'btn btn-sm btn-outline-info',
+      title: 'Wether the action has been marked "as seen" or not yet',
```
dmarcoux, line 12:
```suggestion
      title: 'Whether the action has been marked "as seen" or not yet',
```

src/api/app/components/action_seen_by_user_component.html.haml
```diff
@@ -0,0 +1,19 @@
+- if @render_only
+  - if seen_by_user
+    %i.fa-solid.fa-square-check
+  - else
+    %i.fa-regular.fa-square
```
dmarcoux, line 5: This could be extracted to a method in the component, especially since the exact same code is present on lines 15-18.

src/api/app/views/webui/request/_action_text.html.haml
```diff
@@ -1,31 +1,37 @@
...
-      = action.source_package
-- if details
-  %p.mb-0.mt-0.small
-    by
-    = details['user']
-    #{ApplicationController.helpers.time_ago_in_words(details['time'].to_i)} ago
+.d-flex.align-items-center
+  - if User.session
+    .ms-1.me-3{ id: "action-seen-by-user-wrapper-in-select-#{action.id}" }
+      = render ActionSeenByUserComponent.new(action: action, user: User.session!, render_only: true)
```
dmarcoux, line 4: This would call `User.session/session!` only once:

```suggestion
  - if (user = User.session)
    .ms-1.me-3{ id: "action-seen-by-user-wrapper-in-select-#{action.id}" }
      = render ActionSeenByUserComponent.new(action: action, user: user, render_only: true)
```

src/api/app/models/bs_request_action.rb
```diff
@@ -835,6 +837,16 @@ def commit_details
     package.commit(source_rev) || package.commit
   end
 
+  def toggle_seen_by(user)
+    return unless user
```
dmarcoux, line 841: This code accepts everything which isn't `nil` or `false`, although it could be a bit more robust by checking for the class of `user`. Maybe something like:

```suggestion
    return unless user.is_a?(User)
```

saraycp (approved):

dmarcoux:
Nice work @ncounter

PR #13704, opened 2023-01-17
Fix overflow in revision diff view

hellcp-work:
Aftermath from bs5 migration
|Before|After|
|:---|:---|
|Screenshot from 2023-01-17 13-25-53|Screenshot from 2023-01-17 13-26-17|

This happens with files that don't have a diff, like binaries


danidoni (approved):

eduardoj (approved):

danidoni:
You can check it here: https://obs-reviewlab.opensuse.org/hellcp-work-diff-overflow/request/show/11#tab-pane-changes

PR #13714, opened 2023-01-17
Make index of the diff fit within the column

hellcp-work:
This could technically support any number of digits, but I didn't want to use `style` in haml, to use the exact number, so this PR only supports up to 10 digits. That should be more than enough for most cases.
|Before|After|
|:---|:---|
|Screenshot from 2023-01-18 15-35-33|Screenshot from 2023-01-18 15-36-21|

### DIFF
--- a/src/api/app/assets/stylesheets/webui/diff.scss
+++ b/src/api/app/assets/stylesheets/webui/diff.scss
@@ -11,6 +11,13 @@
 }
 
 .pre.scroll {
+  --line-index-digits: 6ch;
+  @for $i from 1 through 10 {
+    .digits-#{$i} {
+      --line-index-digits: #{$i}ch;
+    }
+    
+  }
   overflow: auto;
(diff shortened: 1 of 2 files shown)


src/api/app/assets/stylesheets/webui/diff.scss
```diff
@@ -31,7 +38,8 @@
         left: 0;
         .number {
           color: $text-muted;
-          width: 4rem;
+          font-family: monospace;
+          width: calc(var(--line-index-digits) + 2rem + #{$card-border-width});
```
dmarcoux, line 42: Would it not be simpler to set `max-width` to a certain amount of characters? The content would then take the space it needs, up to that `max-width` value.

dmarcoux: We talked and it is not possible.

danidoni (approved):

PR #13739, opened 2023-01-19
Run ViewComponent::AvoidGlobalState in view component templates

dmarcoux:
Offense messages are also now more descriptive.

### DIFF
--- a/src/api/lib/rubocop/cop/view_component/avoid_global_state.rb
+++ b/src/api/lib/rubocop/cop/view_component/avoid_global_state.rb
@@ -2,20 +2,31 @@ module RuboCop
   module Cop
     module ViewComponent
       class AvoidGlobalState < RuboCop::Cop::Base
-        # https://docs.rubocop.org/rubocop-ast/node_pattern.html#using-node-matcher-macros
-        def_node_search :params, <<~PATTERN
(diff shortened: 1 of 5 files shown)


src/api/lib/rubocop/cop/view_component/avoid_global_state.rb
```diff
@@ -2,20 +2,31 @@ module RuboCop
...
         PATTERN
 
-        MSG = 'View components should not rely on global state by using params. Instead, pass the required data to the initialize method.'.freeze
+        MESSAGE = 'View components should not rely on global state by using %<content>s. Instead, pass the required data to the initialize method.'.freeze
 
-        # Add an offense for every line using params
+        # Add an offense for using params
         #
         # @param [RuboCop::AST::ClassNode]
-        def on_class(node)
```
dmarcoux, line 15: Before with `on_class`, RuboCop would only run on those checks on Ruby classes. Since view component templates don't have Ruby classes, but potentially only a tiny bit of Ruby code sprinkled here and there, the templates would simply be ignored. So using `params[:something]` directly in a view component template was not resulting in RuboCop offenses.

danidoni (approved):

PR #13740, opened 2023-01-19
Lint view components to mark usage of User class methods as offenses

dmarcoux:
This relies on global state and it should be avoided.


dmarcoux:
This PR should be rebased once #13739 is merged.

dmarcoux:
Fixing all RuboCop offenses found by the new version of `ViewComponent/AvoidGlobalState` would be too much in a single PR, so I decided to regenerate the TODOs. Those can be addressed later in other PRs.

src/api/spec/lib/rubocop/cop/view_component/avoid_global_state_spec.rb
```diff
@@ -15,14 +15,18 @@ def initialize
...
+                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ View components should not rely on global state by calling User.find_by(login: 'someone'). Instead, pass the required data to the initialize method.
 
           end
         end
       RUBY
     end
   end
 
-  context 'when a view component does not use params' do
+  context 'when a view component does not use params or call a User class method' do
```
eduardoj, line 29:
```suggestion
  context 'when a view component does not use params and does not call a User class method' do
```

dmarcoux: The meaning is the same, but if you prefer this version, no worries!

src/api/spec/lib/rubocop/cop/view_component/avoid_global_state_spec.rb
```diff
@@ -34,11 +38,12 @@ def initialize
     end
   end
 
-  context 'when a class which is not a view component uses params' do
+  context 'when a class which is not a view component uses params or call a User class method' do
```
eduardoj, line 41:
```suggestion
  context 'when a class which is not a view component uses params or calls a User class method' do
```

eduardoj (approved):

dmarcoux:
The failing check `OBS: obs-server:obs-bundled-gems - 15.4/x86_64` is completely unrelated to the changes in this PR, it's about a rack version mismatch.

PR #13742, opened 2023-01-19
Show action_seen_by_user flag state in the Select Action dropdown

ncounter:
Follow up of https://github.com/openSUSE/open-build-service/pull/13702
action-as-seen-in-select-checkbox-improved


src/api/app/views/webui/request/_actions_details.html.haml
```diff
@@ -12,6 +12,7 @@
       %ul.dropdown-menu.dropdown-menu-start.scrollable-dropdown
         %li
           %h6.dropdown-header
+            %span Seen
```
dmarcoux, line 15: This should only be rendered when there is a logged-in user.

src/api/app/views/webui/action_seen_by_users/toggle_action_seen_by_user.js.erb
```diff
@@ -0,0 +1,7 @@
+$('#action-seen-by-user-wrapper').html(
+  "<%= escape_javascript(render ActionSeenByUserComponent.new(action: @action, user: @user)) %>"
+);
+
+$('#action-seen-by-user-wrapper-in-select-<%= @action.id %>').html(
+  "<%= escape_javascript(render ActionSeenByUserComponent.new(action: @action, user: @user, render_only: true)) %>"
+);
```
dmarcoux, line 7: Something is wrong with your editor, you're missing a newline at the end of this file, see https://thoughtbot.com/blog/no-newline-at-end-of-file

ncounter: Fixed both, the file and my editor's rule. Thanks for the pointer :+1:

src/api/app/components/action_seen_by_user_component.html.haml
```diff
@@ -0,0 +1,19 @@
+- if @render_only
+  - if seen_by_user
+    %i.fa-regular.fa-eye
```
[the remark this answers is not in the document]
ncounter: That sounds reasonable. I gave it a try with the eye but it makes more sense to keep it consistent.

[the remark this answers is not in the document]
ncounter: The only drawback here is that a checkbox looks like to be clickable but actually it is not, and I would not make it clickable in the dropdown, otherwise one can toggle it `as seen` even without loading it, just marking it from the dropdown. What do you think? is it clear enough? In the end if someone tries to click on the checkbox within the dropdown, it will loads the action and then he can marks it `as seen` from the page just loaded.

ncounter:
Closing this PR an squashing changes into its former one https://github.com/openSUSE/open-build-service/pull/13702

PR #13769, opened 2023-01-26
[backend] fix caching when using with obs_scm_bridge

adrianschroeter:
It is always the same cache independ of any used subdir cgi parameter


M0ses (approved):
LGTM

PR #13774, opened 2023-01-27
[dist] fix some systemd service dependencies

adrianschroeter:
Esp. obssigner did start before storage was mounted when running without obsstoragesetup service.


adrianschroeter:
@M0ses we may want to test this in kanku with the appliance tests before?

M0ses (approved):
LGTM

PR #13783, opened 2023-01-30
Introduces rails-erd

hennevogel:
For an overview diagram of how models are related


hennevogel:
no time to finish this...

PR #13809, opened 2023-02-08
Create a new component for a Card

ncounter:
The new **Card** is a [View Component](https://viewcomponent.org/) meant to unify usages of the [Boostrap card](https://getbootstrap.com/docs/5.0/components/card/) when it is a simple element with a title, few actions and a custom content. The implementation uses advantages of [View Component Slots](https://viewcomponent.org/guide/slots.html) and it comes with some [View Component Previews](https://viewcomponent.org/guide/previews.html).

The previews to check out how the component can look like are available [here](https://obs-reviewlab.opensuse.org/ncounter-card-component/rails/view_components/card_component/)

### DIFF
--- a/src/api/app/components/card_component.html.haml
+++ b/src/api/app/components/card_component.html.haml
@@ -1,0 +1,14 @@
+.card
+  .card-header.d-flex.justify-content-between
+    %div
+      = header
+    - if delete_button?
+      %div
+        = delete_button
+  .card-body
+    = content
+    - if actions?
+      %ul.nav.justify-content-end
+        - actions.each do |action|
+          %li.nav-item
+            = action
--- a/src/api/app/components/card_component.html.haml
+++ b/src/api/app/components/card_component.html.haml
@@ -3,7 +3,7 @@
     %div
       = header
     - if delete_button?
-      %div
+      .ms-4
         = delete_button
   .card-body
     = content
--- a/src/api/app/components/card_component.rb
+++ b/src/api/app/components/card_component.rb
@@ -1,0 +1,8 @@
+# The card component is shaping the card style only, it does not set anything about the width
+# of the card itself because it depends on the occurrence where it is used for.
+# I.e: look at the src/api/app/views/webui/repositories/_repository_entry.html.haml
+class CardComponent < ApplicationComponent
+  renders_one :header
+  renders_one :delete_button
+  renders_many :actions
+end
--- a/src/api/spec/components/previews/card_component_preview.rb
+++ b/src/api/spec/components/previews/card_component_preview.rb
@@ -1,0 +1,88 @@
+class CardComponentPreview < ViewComponent::Preview
+  def with_header
+    render(CardComponent.new) do |component|
+      component.with_header do
+        tag.a('openSUSE_Factory', href: '#').concat(
+          tag.span('tag-1', class: 'badge bg-primary ms-1')
+        ).concat(
+          tag.span('tag-2', class: 'badge bg-primary ms-1')
+        ).concat(
+          tag.span('tag-3', class: 'badge bg-primary ms-1')
+        ).concat(
+          tag.span('tag-4', class: 'badge bg-primary ms-1')
+        )
+      end
+      tag.strong('Repository paths:').concat(
+        content_tag(
+          :ol,
+          content_tag(
+            :li,
+            tag.span('SUSE:SLE-15-SP4:GA/standard').concat(
+              content_tag(
+                :a,
+                tag.i(class: 'fas fa-sm fa-times-circle text-danger'),
+                href: '#',
+                title: 'Delete',
+                class: 'ms-2'
+              )
+            )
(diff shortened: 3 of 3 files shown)


ncounter:
[Unify Card Index Pattern (L)](https://trello.com/c/Ym4kvbUL/1872-unify-card-index-pattern-l)

src/api/app/components/card_component.html.haml
```diff
@@ -0,0 +1,17 @@
+.col-12.col-md-6.col-lg-4
+  .card.mb-3.repository-card
+    .card-header.d-flex.justify-content-between
+      %div
+        = title
+        - tags.each do |tag|
```
dmarcoux, line 6: This is specific to how the view component would be used in the context of repositories. It doesn't apply to other contexts. You should rely on [slots](https://viewcomponent.org/guide/slots.html). You could have a slot for the card header, body and footer for example.

danidoni: Oh, nice tip... We'll take a look at slots for sure

src/api/app/components/card_component.html.haml
```diff
@@ -0,0 +1,13 @@
+.col-12.col-md-6.col-lg-4
+  .card.mb-3.repository-card
+    .card-header.d-flex.justify-content-between
+      %div
+        = header
+      = delete_button
+    .card-body
+      = content
+      - if actions?
+        .d-flex.justify-content-end.mt-2
```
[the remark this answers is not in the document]
ncounter: Sounds a good idea. I addressed that. There is one "drawback" about it, that one should remember to pass action links with the `nav-link` class, but we can live with it. :+1:

src/api/app/components/card_component.html.haml
```diff
@@ -0,0 +1,14 @@
+.col-12.col-md-6.col-lg-4
+  .card.mb-3.repository-card
+    .card-header.d-flex.justify-content-between
+      %div
+        = header
+      %div
+        = delete_button
```
saraycp, line 7: Is it possible to check the presence of the `delete_button` so we don't introduce an empty `div`?

ncounter: Good point, done. :+1:

saraycp (approved):

src/api/app/components/card_component.html.haml
```diff
@@ -0,0 +1,15 @@
+.col-12.col-md-6.col-lg-4
+  .card.mb-3.repository-card
+    .card-header.d-flex.justify-content-between
+      %div
+        = header
+      - if delete_button?
+        %div
+          = delete_button
```
dmarcoux, line 8: With this approach, we will have to duplicate the delete button over and over again. This could be streamlined to ensure the delete buttons are always consistent. This could be a component.

src/api/app/components/card_component.html.haml
```diff
@@ -0,0 +1,15 @@
+.col-12.col-md-6.col-lg-4
```
dmarcoux, line 1: What is the reason for having a fixed width? Having the component adjust to its content would make it more flexible. Here's an example from a PR based on the card component, there's clearly too much whitespace. It could easily have half the width and still be fine.

whitespace

src/api/app/components/card_component.html.haml
ncounter: That's the plan, to have a root component to reuse and do not _reinvent the wheel_ every time. I felt it was for a separate refactoring PR (as we do already have duplication around about it). Would you feel comfortable to keep it as it is for now, or do you prefer to add it here and now before merging the card component?

src/api/app/components/card_component.html.haml
```diff
@@ -0,0 +1,15 @@
...
+  .card.mb-3.repository-card
```
dmarcoux, line 2: `.repository-card` should not be here, right? It shouldn't apply to all card components.

src/api/app/components/card_component.html.haml
dmarcoux: Could we not merge this line with the `.card` line below? Having an container for the `.card` container isn't needed.

src/api/app/components/card_component.html.haml
dmarcoux: Handling this where the card is used, that would be a good first step. It doesn't have to happen in this PR though.

src/api/app/components/card_component.html.haml
ncounter: Fixed width is fine I'd say, but not in a hard way. I mean, the card component should not have fixed width, I agree, I will adjust it. But it'd still be good if every usage of the card keeps in mind to set the width for the occurrences of that type, i.e.: all the repository cards in the repository page should have the same width, and all the token cards in the token page should have the same width as well, but the two width for repository and token can be different, that one can.

Anyway, I'm going to adapt here the card component to not set the width, :+1:

PR #13811, opened 2023-02-08
Rescue format errors when parsing workflows yaml

hellcp-work:
This shows a nicer error when formatting of the workflows yaml fails due to non existent placeholder variable

In order to verify this:
1. Create a workflow
2. Set a workflows.yml containing random invalid placeholder variables (starting with `%`) around the file
3. Trigger a workflow run
4. See it show a nice error telling you what's wrong

### DIFF
--- a/src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
+++ b/src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
@@ -83,5 +83,14 @@ RSpec.describe Workflows::YAMLToWorkflowsService, type: :service do
         expect { subject }.to raise_error(Token::Errors::WorkflowsYamlNotParsable)
       end
     end
+
+    context 'with invalid placeholder variables' do
+      let(:workflows_yml_file) { Rails.root.join('spec/support/files/unparsable_workflows_placeholders.yml').expand_path }
+      let(:payload) { github_extractor_payload }
+
+      it 'raises a user-friendly error' do
+        expect { subject }.to raise_error(Token::Errors::WorkflowsYamlNotParsable, 'Unable to parse .obs/workflows.yml: malformed format string - %S')
+      end
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
```diff
@@ -83,5 +83,14 @@
         expect { subject }.to raise_error(Token::Errors::WorkflowsYamlNotParsable)
       end
     end
+
+    context 'with invalid placeholder variables' do
+      let(:workflows_yml_file) { Rails.root.join('spec/support/files/unparsable_workflows_placeholders.yml').expand_path }
+      let(:payload) { github_extractor_payload }
+
+      it 'raises a user-friendly error' do
+        expect { subject }.to raise_error(Token::Errors::WorkflowsYamlNotParsable)
```
dmarcoux, line 92: I don't think this is user friendly if it's simply the exception message, which ends up being something like `ArgumentError (malformed format string - %;):`.

dmarcoux: Oh right, we're re-raising a different exception. I should have paid closer attention :+1:

src/api/spec/services/workflows/yaml_to_workflows_service_spec.rb
```diff
@@ -83,5 +83,14 @@
         expect { subject }.to raise_error(Token::Errors::WorkflowsYamlNotParsable)
       end
     end
+
+    context 'with invalid placeholder variables' do
+      let(:workflows_yml_file) { Rails.root.join('spec/support/files/unparsable_workflows_placeholders.yml').expand_path }
+      let(:payload) { github_extractor_payload }
+
+      it 'raises a user-friendly error' do
+        expect { subject }.to raise_error(Token::Errors::WorkflowsYamlNotParsable, 'Unable to parse .obs/workflows.yml: malformed format string - %S')
```
dmarcoux, line 92: This is definitely better now, although I find the error message still a bit cryptic. Especially since `%S` refers to a string in `format(..)`, although the actual error is happening due to `%u`.

dmarcoux (approved):

rubhanazeem (approved):

PR #13814, opened 2023-02-08
Performance of the workflow runs UI 

rubhanazeem:
WorkflowRun finder was using an iterator to count the number of workflow runs grouped by generic event type. The iterator logic was increasing the page loading time.

With this PR the loading time is reduced. These calculations are based on 20000 records
**Before:** loading time was > 6 seconds
**After:** Loading time is < 1 second (~0.17 seconds)

### DIFF
--- a/src/api/app/queries/workflow_runs_finder.rb
+++ b/src/api/app/queries/workflow_runs_finder.rb
@@ -14,8 +14,8 @@ class WorkflowRunsFinder
   end
 
   def group_by_generic_event_type
-    @relation.all.each_with_object(Hash.new(0)) do |workflow_run, grouped_workflows|
-      grouped_workflows[workflow_run.generic_event_type] += 1
+    EVENT_TYPE_MAPPING.to_h do |key, _value|
+      [key, with_generic_event_type(key).count]
     end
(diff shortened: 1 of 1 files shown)


dmarcoux:
The lines below are surely also causing performance issues, since we sometimes fetch all workflow runs on line 11, before paginating the returned array on line 14:

https://github.com/openSUSE/open-build-service/blob/052a75cc9f2b393aeba151aadc811f33ed228e87/src/api/app/controllers/webui/workflow_runs_controller.rb#L6-L14

rubhanazeem:
> The lines below are surely also causing performance issues, since we sometimes fetch all workflow runs for a token on line 11, before paginating the returned array on line 14:
> 
> https://github.com/openSUSE/open-build-service/blob/052a75cc9f2b393aeba151aadc811f33ed228e87/src/api/app/controllers/webui/workflow_runs_controller.rb#L6-L14

True. There is an extra query. I've made some changes to avoid that.

dmarcoux (approved):
Tested locally with 20 000 workflow runs, it's definitely much faster. Nice job!

PR #13824, opened 2023-02-10
Use card components for group members

danidoni:
We just use the card component developed in https://github.com/openSUSE/open-build-service/pull/13809 (it's a follow up)
This is how it looks like:

image

### DIFF
--- a/src/api/app/views/webui/groups/_group_members.html.haml
+++ b/src/api/app/views/webui/groups/_group_members.html.haml
@@ -6,7 +6,7 @@
     This group does not have members.
 - else
   #group-users
-    .d-flex.flex-wrap.gap-2
+    .d-flex.flex-wrap.gap-4
       - group.users.each do |user|
         .group-user
           = render(CardComponent.new) do |component|
--- a/src/api/app/views/webui/groups/_group_members.html.haml
+++ b/src/api/app/views/webui/groups/_group_members.html.haml
@@ -35,8 +35,7 @@
                                   'bs-target': '#delete-group-members-modal',
                                   confirmation_text: "Please confirm deletion of '#{user}' from this group",
                                   action: group_user_path(group_title: group.title, user_login: user.login) }) do
-                  .float-end
-                    %i.fas.fa-sm.fa-times-circle.text-danger
+                  %i.fas.fa-sm.fa-times-circle.text-danger
 
 - if write_access || display_group_email
   .pt-4
--- a/src/api/app/views/webui/groups/_group_members.html.haml
+++ b/src/api/app/views/webui/groups/_group_members.html.haml
@@ -6,33 +6,37 @@
     This group does not have members.
 - else
   #group-users
-    .d-flex.flex-wrap
+    .d-flex.flex-wrap.gap-2
       - group.users.each do |user|
-        .card.m-1.p-2.group-user
-          - if policy(GroupsUser.new(group: group, user: user)).destroy?
-            = link_to('#',
-                      title: 'Remove member from group',
-                      data: { 'bs-toggle': 'modal',
-                              'bs-target': '#delete-group-members-modal',
-                              confirmation_text: "Please confirm deletion of '#{user}' from this group",
-                              action: group_user_path(group_title: group.title, user_login: user.login) }) do
-              .float-end
-                %i.fas.fa-sm.fa-times-circle.text-danger
-          .row.g-0
-            .col-md-3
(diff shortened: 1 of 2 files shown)


ncounter:
[Unify Card Index Pattern (L)](https://trello.com/c/Ym4kvbUL/1872-unify-card-index-pattern-l)

dmarcoux:
As I mentioned in a [comment from another PR](https://github.com/openSUSE/open-build-service/pull/13809/files#r1104166625), the cards should adjust to their content. Beside this, the cards definitely need some left/right margin. See the screenshot below:

group-members

src/api/app/views/webui/groups/_group_members.html.haml
```diff
@@ -34,6 +22,21 @@
                     Maintainer
                   %i.fas.fa-spinner.fa-spin.d-none
 
+          - component.with_header do
+            = link_to(user_path(user)) do
+              = display_name(user)
```
dmarcoux, line 27: We need to handle longer usernames to fix issues like in the screenshot below:

long-strings

ncounter: That's a good catch, but it should also be handled in the [`card-component`](https://github.com/openSUSE/open-build-service/pull/13809) implementation/improvement I'd say.
I'll take care of it.

ncounter: https://github.com/openSUSE/open-build-service/pull/13809/commits/78e11f25bb3e4f96221634aa04cfa0cf4c54a179

src/api/spec/features/webui/groups_spec.rb
```diff
@@ -58,7 +58,13 @@ def group_in_datatable(page, group)
     visit group_path(group_1)
 
     within(find('div.group-user', text: admin.login)) do
-      click_link('Remove member from group')
+      # HACK: Click on a button that has zero size. The delete button is a
```
danidoni, line 61: I had to add this because I was getting failures on the feature tests regarding the delete button not being clickable:
```
Selenium::WebDriver::Error::ElementNotInteractableError:                                                                        
       element not interactable: element has zero size
```
The problem was that the button element is a floating one, and the Capybara driver cannot click on it.
I'm not 100% about using this hack, @ncounter can we somehow change the delete button so it's not a floating element?

src/api/app/views/webui/groups/_group_members.html.haml
```diff
@@ -6,33 +6,37 @@
...
+
+            - component.with_delete_button do
+              - if policy(GroupsUser.new(group: group, user: user)).destroy?
+                = link_to('#',
+                          title: 'Remove member from group',
+                          data: { 'bs-toggle': 'modal',
+                                  'bs-target': '#delete-group-members-modal',
+                                  confirmation_text: "Please confirm deletion of '#{user}' from this group",
+                                  action: group_user_path(group_title: group.title, user_login: user.login) }) do
+                  .float-end
```
ncounter, line 38: This can be safely removed, there is no need for this class and this wrapper.

It will became from
- `<a><div class="float-end"><i/></div></a>`
to
- `<a><i/></a>`
and by this the link `<a>` will take the computed width and height that the `<i>` icon needs.

src/api/app/views/webui/groups/_group_members.html.haml
```diff
@@ -6,33 +6,37 @@
     This group does not have members.
 - else
   #group-users
-    .d-flex.flex-wrap
+    .d-flex.flex-wrap.gap-2
```
ncounter, line 9: In the name of consistency I'd align it to `gap-4` (which I personally prefer), or the other way around (which I don't mind either btw) I'd change to `gap-2` also the [repository](https://github.com/openSUSE/open-build-service/pull/13826) and the [token](https://github.com/openSUSE/open-build-service/pull/13829) usage of the card component

ncounter (approved):
I left few comments, other than them the rest LGTM :+1:

src/api/spec/features/webui/groups_spec.rb
dmarcoux: Cool, so this will be changed? It sounds like it should.

danidoni: I'm fixing it now

src/api/app/views/webui/groups/_group_members.html.haml
danidoni: I'll with `gap-4`

PR #13825, opened 2023-02-10
[dist] create also a 4k rsa key on the appliance

adrianschroeter:
fixing cleanup of temporary config file as well

This requires sign 2.7.3 or newer, otherwise a buffer limit will be reached.


M0ses (approved):
LGTM

PR #13840, opened 2023-02-13
Download yaml file

saraycp:


krauselukas (changes requested):
Oh wait, I might have found something...

src/api/app/services/workflows/yaml_downloader.rb
```diff
@@ -31,27 +30,26 @@ def download_url
       case @scm_payload[:scm]
       when 'github'
         client = Octokit::Client.new(access_token: @token.scm_token, api_endpoint: @scm_payload[:api_endpoint])
-        # :ref can be the name of the commit, branch or tag.
-        client.content("#{@scm_payload[:target_repository_full_name]}", path: "/#{@token.workflow_configuration_path}", ref: @scm_payload[:target_branch])[:download_url]
+        # :ref can be the name of the commit, branch or tag. We use the commit sha, which outlast the branch most of the time.
+        client.content("#{@scm_payload[:target_repository_full_name]}", path: "/#{@token.workflow_configuration_path}", ref: @scm_payload[:commit_sha])[:download_url]
```
krauselukas, line 34: yeah, so the `:commit_sha` for a pull_request is this part from the webhook payload `:pull_request, :head, :sha`. This contains the commit sha of the file after the changes from the PR are applied. If we use that in the download url and a user modifies the `workflow.yml` file inside the PR, we would download the modified version. Not the one from the target branch.

krauselukas (changes requested):

src/api/app/services/workflows/yaml_downloader.rb
krauselukas: This is a problem, because it basically allows everyone to run whatever workflow they want

krauselukas: We need to use something else here, need to check which field from the payload would be the correct one. To test this yourself, open a test PR with some changes on the `workflow.yml` and use the check the resulting download url. You will see the workflow.yml with the changes the PR introduces

saraycp: I think you are right, let me check something...

dmarcoux: Before, we downloaded `workflows.yml` from the target branch, which is more often than not, the default branch. This is what we want, as it prevents anybody from opening a PR with a modified `workflows.yml` and running it.

As for tag pushes, it's safe to assume that anybody able to push a tag is also able to change the `workflows.yml`, so downloading from the commit referenced by the tag should be considered safe.

So the way we download `workflows.yml` should change depending on the webhook event.

saraycp:
I close this PR, will open new ones

PR #13847, opened 2023-02-15
Extract the correct reference to the YAML file for GitHub/Gitea events

saraycp:
**NOTE:** This PR focuses on GitHub/Gitea events only.

The YAML file should be downloaded from the target branch or commit, not the source one.

We want to rely on the head commit SHA of the target branch. In case of a push event that deletes the ref, we need to take the commit SHA from the "after" key comming from the original payload.

A new key is added to the SCMWebhook payload: yaml_reference.

PR #13852, opened 2023-02-16
Ignore github push events for deleted refs

krauselukas:
When the push event is triggered when a branch/commit got deleted, there is nothing to run the workflow for. In those cases we should simply return and do nothing.

We realized that while working on https://github.com/openSUSE/open-build-service/pull/13847

### DIFF
--- a/src/api/app/controllers/trigger_workflow_controller.rb
+++ b/src/api/app/controllers/trigger_workflow_controller.rb
@@ -81,8 +81,8 @@ class TriggerWorkflowController < TriggerController
   def extract_scm_webhook
     @scm_webhook = TriggerControllerService::SCMExtractor.new(scm, event, payload).call
 
-    # There are plenty of different pull/merge request action which we don't support.
+    # There are plenty of different pull/merge request and push events which we don't support.
     # Those should not cause an error, we simply ignore them.
-    render_ok if @scm_webhook && @scm_webhook.ignored_pull_request_action?
+    render_ok if @scm_webhook && (@scm_webhook.ignored_pull_request_action? || @scm_webhook.ignored_push_event?)
   end
 end
--- a/src/api/app/models/scm_webhook.rb
+++ b/src/api/app/models/scm_webhook.rb
@@ -62,6 +62,10 @@ class SCMWebhook
     github_ping? || gitea_ping?
   end
 
+  def ignored_push_event?
+    ignored_github_push_event?
+  end
+
   private
 
   def github_push_event?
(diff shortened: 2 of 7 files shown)


src/api/app/controllers/trigger_workflow_controller.rb
```diff
@@ -83,6 +83,6 @@ def extract_scm_webhook
 
     # There are plenty of different pull/merge request action which we don't support.
```
dmarcoux, line 84: This comment should be updated to reflect the changes.

krauselukas: Updated the comment

hennevogel:
Hm, how can this even happen?

krauselukas:
@hennevogel this happens when a branch gets deleted on the repository where the webhook is setup. I triggers a push event that is referring to a commit with the sha '00000...' and the `deleted` field in the payload is set to `true`. So basically referring to nothing. Thats why we would ignore it, since there is nothing to do. We have some workflow runs that failed with push events triggered after depfu deletes the branch when the PR gets merged

dmarcoux:
> Hm, how can this even happen?

This is documented by GitHub [here](https://docs.github.com/webhooks-and-events/webhooks/webhook-events-and-payloads#push), see `deleted` under the _Body parameters_:

github

saraycp:
> @hennevogel this happens when a branch gets deleted on the repository where the webhook is setup

In other words:

When we have a branch created upstream (i.e. a depfu branch) and then we remove it (i.e. after merging or closing the PR), a push event with `deleted: true` is triggered.
OBS doesn't need to run a workflow for a branch that no longer exists.

krauselukas:
[Reliably download workflow configuration files (M)](https://trello.com/c/oVGjk512/2109-reliably-download-workflow-configuration-files-m)

src/api/app/models/scm_webhook.rb
```diff
@@ -72,6 +76,10 @@ def gitlab_push_event?
     @payload[:scm] == 'gitlab' && @payload[:event] == 'Push Hook'
   end
 
+  def github_ignored_push_event?
+    github_push_event? && @payload[:deleted]
+  end
+
```
saraycp, line 82: [nitpick] Please move this method to the bottom, where the rest of `*ignored*` methods are.

saraycp: And let's follow the same name pattern:
```suggestion
  def github_ignored_github_push_event?
    github_push_event? && @payload[:deleted]
  end

```

saraycp (changes requested):

src/api/app/models/scm_webhook.rb
krauselukas: done

saraycp (approved):

PR #13858, opened 2023-02-17
Use Marcel:Magic instead of hardcoded binary extensions

### LINKED ISSUE #13613: Zstd tarballs appear as clickable items in web UI
Hello good folks,

**Zstd** tarballs are appearing as _clickable_ links in the web UI while it shouldn't - just like any other tarball/archive. But not always they show as _clickable_, as seen in the pictures below.

How my [nix-user-chroot](https://build.opensuse.org/package/show/home:luc14n0/nix-user-chroot) looks like:

image

And if clicked:

image

How my [ion](https://build.opensuse.org/package/show/home:luc14n0/ion) package looks like:

image

---

The same issue happens for some other type of files, like icons, from the [OBS:Flatpak:Templates/FlatpakTemplate](https://build.opensuse.org/package/show/OBS:Flatpak:Templates/FlatpakTemplate) package:

image

hellcp-work:
Since we don't get enough information from the backend about the files, we don't know what their mime types are, but we can use Marcel, which is built into Rails nowadays, to guess the mimetypes from the extensions. We could use file content, but that seems inefficient. Sadly due to its license, Marcel uses Apache Tika database for mime information, instead of Freedesktop one, which is more complete, so we need to append some mime configuration in the initializer

This is intended to fix #13613

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -1186,12 +1186,6 @@ class Package < ApplicationRecord
     store if update_needed
   end
 
-  def self.is_binary_file?(filename)
-    return false unless (mime = Marcel::Magic.by_path(filename))
-
-    !mime.text?
-  end
-
   def serviceinfo
     begin
       dir = Directory.hashed(project: project.name, package: name)
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -13,13 +13,6 @@ class Package < ApplicationRecord
   include PackageSphinx
   include MultibuildPackage
 
-  BINARY_EXTENSIONS = ['.0', '.bin', '.bin_mid', '.bz', '.bz2', '.ccf', '.cert',
-                       '.chk', '.der', '.dll', '.exe', '.fw', '.gem', '.gif', '.gz',
-                       '.jar', '.jpeg', '.jpg', '.lzma', '.ogg', '.otf', '.oxt',
-                       '.pdf', '.pk3', '.png', '.ps', '.rpm', '.sig', '.svgz', '.tar',
-                       '.taz', '.tb2', '.tbz', '.tbz2', '.tgz', '.tlz', '.txz', '.ucode',
-                       '.xpm', '.xz', '.z', '.zip', '.ttf'].freeze
-
   has_many :relationships, dependent: :destroy, inverse_of: :package
   belongs_to :kiwi_image, class_name: 'Kiwi::Image', inverse_of: :package, optional: true
   accepts_nested_attributes_for :kiwi_image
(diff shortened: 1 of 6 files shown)


src/api/config/initializers/marcel.rb
```diff
@@ -0,0 +1,11 @@
+require 'marcel'
+require 'marcel/mime_type'
+
+# Marcel uses Apache Tika mime database, which is less complete than Freedesktop one
+# Marcel::Magic is based on MimeMagic which uses the Freedesktop mime database, so it's
+# worth considering using it in order to avoid the below configuration
```
hennevogel, line 6: Not sure what you want to say to me with this comment? That we shouldn't do this but something else? :)

src/api/app/models/package.rb
```diff
@@ -1194,7 +1187,9 @@ def enable_for_repository(repo_name)
   end
 
   def self.is_binary_file?(filename)
-    BINARY_EXTENSIONS.include?(File.extname(filename).downcase)
+    return false unless (mime = Marcel::Magic.by_path(filename))
+
+    !mime.text?
```
hennevogel, line 1192: This needs specs before changing. As we have no idea what is changing otherwise.

hennevogel (changes requested):

src/api/config/initializers/marcel.rb
hennevogel: I would rather have you explain why you are going down this route and not the other in the commit message :)

src/api/test/unit/package_test.rb
```diff
@@ -312,32 +312,6 @@ def test_activity
...
-    end
-
-    # these aren't binary
-    generate_suffixes(['diff', 'txt', 'csv', 'pm', 'c', 'rb', 'h']).each do |suffix|
-      file_paths.each do |file_path|
-        filename = file_path + '.' + suffix
-        assert_not Package.is_binary_file?(filename), "File #{filename} should not be treated as binary"
-      end
-    end
-  end
```
hennevogel, line 339: You can now also get rid of `generate_suffixes`...

src/api/spec/helpers/webui/package_helper_spec.rb
```diff
@@ -260,4 +260,12 @@
     it { expect(viewable_file?('some_file.rb')).to be(true) }
     it { expect(viewable_file?('some_file.bin')).to be(false) }
   end
+
+  describe '#binary_file?' do
+    it { expect(binary_file?('some/file.tar.gz')).to be(true) }
+    it { expect(binary_file?('some_file.rb')).to be(false) }
+    it { expect(binary_file?('some_file.spec')).to be(false) }
+    it { expect(binary_file?('some_file.tar')).to be(true) }
+    it { expect(binary_file?('some_file.tar.bz2')).to be(true) }
```
hennevogel, line 269: What about the other suffixes from the old test?

hennevogel (changes requested):

hennevogel:
Hello World?

hennevogel (approved):

PR #13859, opened 2023-02-17
Ignore GitLab push events for deleted references

saraycp:
When an upstream branch/commit is deleted, GitLab triggers a Push Hook event.
It makes no sense to try to run a workflow on something removed (i.e. because a merge request was merged or closed and then the branch was removed).

As we do with GitHub [here](https://github.com/openSUSE/open-build-service/pull/13852), we will ignore these kinds of push events.
Unlike GitHub, GitLab doesn't provide a `deleted` field or similar in the payload, so we rely on the `after` field which always contains a commit SHA like this: `0000000000000000000000000000000000000000`.

### DIFF
--- a/src/api/app/models/scm_webhook.rb
+++ b/src/api/app/models/scm_webhook.rb
@@ -63,7 +63,7 @@ class SCMWebhook
   end
 
   def ignored_push_event?
-    ignored_github_push_event?
+    ignored_github_push_event? || ignored_gitlab_push_event?
   end
 
   private
@@ -126,5 +126,10 @@ class SCMWebhook
 
   def ignored_github_push_event?
     github_push_event? && @payload[:deleted]
+  end
+
+  def ignored_gitlab_push_event?
+    # In Push Hook events to delete a branch, the after field is '0000000000000000000000000000000000000000'
+    gitlab_push_event? && @payload[:commit_sha].match?(/\A0+\z/)
   end
 end
--- a/src/api/spec/models/scm_webhook_spec.rb
+++ b/src/api/spec/models/scm_webhook_spec.rb
@@ -451,5 +451,17 @@ RSpec.describe SCMWebhook do
 
       it { is_expected.to be false }
     end
+
+    context 'with a Push Hook event from Gitlab for a deleted commit reference' do
+      let(:payload) { { scm: 'gitlab', event: 'Push Hook', ref: 'refs/heads/branch_123', commit_sha: '0000000000000000000000000000000000000000' } }
+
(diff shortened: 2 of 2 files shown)


src/api/spec/services/trigger_controller_service/scm_extractor_spec.rb
```diff
@@ -229,6 +229,42 @@
         end
       end
 
+      context 'with a push event for a commit that is removed' do
```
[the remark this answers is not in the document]
saraycp: True, I'll remove it.

src/api/app/models/scm_webhook.rb
```diff
@@ -112,11 +116,20 @@ def ignored_github_pull_request_action?
...
     gitlab_merge_request? && ALLOWED_MERGE_REQUEST_ACTIONS.exclude?(@payload[:action])
   end
 
   def ignored_gitea_pull_request_action?
     gitea_pull_request? && ALLOWED_PULL_REQUEST_ACTIONS.exclude?(@payload[:action])
   end
+
+  def ignored_gitlab_push_event?
+    # In Push Hook events to delete a branch, the after field is '0000000000000000000000000000000000000000'
+    gitlab_push_event? && @payload[:commit_sha] == '0000000000000000000000000000000000000000' # '0' * 40
```
krauselukas, line 133: I just thought that this comparison is a bit risky, if the amount of zeros is just off by one it will return someting else

krauselukas, line 133:
```suggestion
    gitlab_push_event? && @payload[:commit_sha].include?('0000000000')
```

krauselukas: I think checking with `include?` might be a bit more forgiving

krauselukas: sure, was more thinking about our zeros :D but yes, we code this only once, so I'm fine with it

krauselukas: it fine as it is

saraycp: I'm going to use a regular expression, wait for it...

saraycp:
@krauselukas @hellcp-work I rebased on master and check the commit SHA with a regular expression `@payload[:commit_sha].match?(/\A0+\z/)`

krauselukas:
@saraycp please just add your description also to the commit message :)

saraycp:
> @saraycp please just add your description also to the commit message :)

done

krauselukas (approved):

PR #13869, opened 2023-02-21
Fix broken diffs

llyyr:
We correctly calculate the max digits but for some reason the line index digits needs a +1 to correctly display diffs

Example of broken diffs: image


llyyr:
To be honest, it's harder to find a font where it doesn't happen. Happens with Roboto and Iosevka which are completely different. Would it be better to add 1px to https://github.com/openSUSE/open-build-service/blob/master/src/api/app/assets/stylesheets/webui/diff.scss#L42 ?

llyyr:
I realized that they only break when you zoom out, and figured this was the solution that felt the least arbitrary and also happened to be the least amount of change. Nothing changes visually at 100% zoom or higher, but it doesn't break at 90% zoom or lower anymore.

PR #13872, opened 2023-02-21
Only fetch one commit for the latest revision

### LINKED ISSUE #13871: Performance problems with Webui::PackageController#show and large amount of revisions
On the reference server we have a couple of packages that have a huge amount (>5K) of revisions. In that case viewing them is very slow. 

- [home:repo-checker/rebuilds](https://obs-measure.opensuse.org/d/fzuOfHQik/measurements-per-request?orgId=1&var-request_id=d30b5a23-a268-41c3-b75f-c821de61ad57&from=1676992094380&to=1676992152319&var-method=Webui%3A%3APackageController%23show)
- [openSUSE:Factory:Staging/dashboard](https://obs-measure.opensuse.org/d/fzuOfHQik/measurements-per-request?orgId=1&var-request_id=f95bd60b-4eef-4022-b0fd-69401d08f6c5&from=1676256611958&to=1676256643639&var-method=Webui%3A%3APackageController%23show)

This is because of `@package.commit` from `app/views/webui/package/_files_view.html.haml` fetching all of the history to display a commit...

https://github.com/openSUSE/open-build-service/blob/35f244db49a806762723c4a2e7d6dbbd064ce564/src/api/app/models/package.rb#L1222-L1237

hellcp-work:
This is intended to fix #13871, by performing a caching operation of just the latest commit if it wasn't cached already, and returning it quickly

### DIFF
--- a/src/api/app/models/package.rb
+++ b/src/api/app/models/package.rb
@@ -1207,10 +1207,9 @@ class Package < ApplicationRecord
     {}
   end
 
-  def parse_all_history
-    answer = source_file('_history')
-
-    doc = Xmlhash.parse(answer)
+  def cache_revisions(revision = nil)
+    opts = revision ? { rev: revision } : {}
+    doc = Xmlhash.parse(Backend::Api::Sources::Package.revisions(project.name, name, opts))
     doc.elements('revision') do |s|
       Rails.cache.write(['history', self, s['rev']], s)
       Rails.cache.write(['history_md5', self, s.get('srcmd5')], s)
@@ -1231,7 +1230,7 @@ class Package < ApplicationRecord
     commit = fetch_rev_from_history_cache(rev)
     return commit if commit
 
-    parse_all_history
+    cache_revisions
     # now it has to be in cache
     fetch_rev_from_history_cache(rev)
   end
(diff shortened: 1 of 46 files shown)


src/api/app/models/package.rb
```diff
@@ -1231,6 +1235,12 @@ def commit(rev = nil)
     commit = fetch_rev_from_history_cache(rev)
     return commit if commit
 
+    if rev&.to_i == self.rev&.to_i
+      doc = Xmlhash.parse(source_file('_history', limit: 1))
+      cache_revision_commits(doc)
```
eduardoj, line 1240: This call will store in cache only one element. Is that what we want?

eduardoj: I see.

I hoped we could paginate through the `_history` using this endpoint of the backend API : https://github.com/openSUSE/open-build-service/blob/b637d1118d0c9b3a6937959a281fb01cd13e19df/src/backend/bs_srcserver#L7610

But it looks like it is not supported.

src/api/app/models/package.rb
```diff
@@ -1210,6 +1210,15 @@ def serviceinfo
   def parse_all_history
     answer = source_file('_history')
 
+    cache_revision_commits(doc)
+  end
+
+  def cache_single_commit(revision)
```
hennevogel, line 1216: All 3 methods can now be `cache_revisions(rev: nil)`, the only difference is how you fetch the `_history`.

hennevogel (changes requested):

hennevogel (approved):

PR #13886, opened 2023-02-23
Increase `scroll-padding-top` value by height of notification toolbar

krauselukas:
Right now selected anchor tags are covered when the notification toolbar is present. Increasing the value by the height of the notification toolbar prevents it from happening.

Now:


krauselukas:
[Line number anchors don't account for notifications bar (S)](https://trello.com/c/7XE1AVg6/2128-line-number-anchors-dont-account-for-notifications-bar-s)

dmarcoux (approved):
This doesn't hurt either when the notification toolbar isn't displayed :+1:

PR #13891, opened 2023-02-24
Add BsRequestHistoryElementComponent specs

saraycp:


saraycp:
Rebased on master

danidoni (approved):

PR #13892, opened 2023-02-24
Add preview for HistoryElement::RequestReviewAdded

saraycp:
Both `RequestReviewAdded` and `RequestSuperseded` history elements look different from the rest of the history elements' types.  So it's worth having a preview of the first one too.

### DIFF
--- a/src/api/spec/components/previews/bs_request_history_element_component_preview.rb
+++ b/src/api/spec/components/previews/bs_request_history_element_component_preview.rb
@@ -1,9 +1,16 @@
 class BsRequestHistoryElementComponentPreview < ViewComponent::Preview
+  # /rails/view_components/bs_request_history_element_component/with_history_element_request_superseded
   def with_history_element_request_superseded
     render(BsRequestHistoryElementComponent.new(element: HistoryElement::RequestSuperseded.last))
   end
 
+  # /rails/view_components/bs_request_history_element_component/with_history_element_request_accepted
   def with_history_element_request_accepted
     render(BsRequestHistoryElementComponent.new(element: HistoryElement::RequestAccepted.last))
+  end
+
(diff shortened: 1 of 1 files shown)


src/api/spec/components/previews/bs_request_history_element_component_preview.rb
```diff
@@ -1,9 +1,16 @@
...
   end
 
+  # /rails/view_components/bs_request_history_element_component/with_history_element_request_accepted
   def with_history_element_request_accepted
     render(BsRequestHistoryElementComponent.new(element: HistoryElement::RequestAccepted.last))
   end
+
+  # /rails/view_components/bs_request_history_element_component/with_history_element_request_review_added
+  def with_history_element_request_review_added
+    render(BsRequestHistoryElementComponent.new(element: HistoryElement::RequestReviewAdded.last))
```
krauselukas, line 14: @saraycp hmm, I'm not sure if its good to depend pre existing database entries for the previews. Sometimes we create the records in the preview itself, sometime we depend on the existence of them in the DB. Anyway you don't introduce this, there are other previews that do the same. But I think I will create a card to start talking about this.

krauselukas (approved):

src/api/spec/components/previews/bs_request_history_element_component_preview.rb
krauselukas: card is in the backlog now

saraycp: I totally agree.

saraycp: I created a Free Style [card](https://trello.com/c/wGPrnopV/2860-set-data-for-components-previews).

PR #13907, opened 2023-02-27
[dist] clouduploader: Time.zone is no more

adrianschroeter:
It got dropped, but we calculate anyone only a time difference here, so the zone does not matter.


.rubocop_todo.yml
```diff
@@ -59,6 +66,64 @@ Metrics/ParameterLists:
   Exclude:
     - 'dist/clouduploader.rb'
 
+# Offense count: 6
+RSpec/BeforeAfterAll:
+  Exclude:
+    - 'spec/spec_helper.rb'
+    - 'spec/rails_helper.rb'
```
eduardoj, line 73:
```suggestion
```
This line could be removed, if I'm not wrong. There is no `rails_helper.rb` file in `dist/t/spec/`.

.rubocop_todo.yml
```diff
@@ -59,6 +66,64 @@ Metrics/ParameterLists:
...
+    - 'spec/support/**/*.rb'
```
eduardoj, line 74:
```suggestion
    - 'spec/support/capybara.rb'
```
There is only one file, inside `dist/t/spec/support/`. I guess if you modified this `.rubocop_todo.yml` manually, or was automatically generated...

.rubocop_todo.yml
hennevogel: This is automatically generated with `rake dev:lint:rubocop:auto_gen_config:root`

.rubocop_todo.yml
hennevogel: Same here...

.rubocop_todo.yml
eduardoj: Hm. I guess it doesn't harm. LGTM (let's get this merged).

eduardoj (approved):

PR #13910, opened 2023-02-27
Move new diffs to a component

hellcp-work:
While working on adding comments into the diffs, I started running into issues while trying to make my code work with PackageController#rdiff, since the partial was also used there. To avoid those issues I thought I could use encapsulation and make the code for the new diffs as reusable as possible for the future. That's what components are for! So here is me moving that code into components

### DIFF
--- a/src/api/spec/cassettes/DiffComponent/with_a_request_having_a_submit_action/renders_the_diff.yml
+++ b/src/api/spec/cassettes/DiffComponent/with_a_request_having_a_submit_action/renders_the_diff.yml
@@ -1,0 +1,569 @@
+---
+http_interactions:
+- request:
+    method: put
+    uri: http://backend:5352/source/target_project/_meta?user=user_6
+    body:
+      encoding: UTF-8
+      string: |
+        <project name="target_project">
+          <title>Dance Dance Dance</title>
+          <description/>
+        </project>
+    headers:
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+      Accept:
+      - "*/*"
+      User-Agent:
+      - Ruby
+  response:
+    status:
+      code: 200
+      message: OK
+    headers:
+      Content-Type:
+      - text/xml
+      Cache-Control:
+      - no-cache
+      Connection:
+      - close
+      Content-Length:
+      - '108'
+    body:
+      encoding: UTF-8
+      string: |
+        <project name="target_project">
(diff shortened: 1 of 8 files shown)


src/api/app/components/diff_list_component.html.haml
```diff
@@ -0,0 +1,16 @@
...
+        %button.accordion-button.text-bg-light{ type: 'button', data: { 'bs-toggle': 'collapse',
+                                                                        'bs-target': "#diff-item-#{name.parameterize}" },
+                                                aria: { expanded: 'true', controls: "diff-item-#{name.parameterize}" } }
+          - if (old_file = file_info['old'])
+            = changed_filename(old_file['name'], name, state)
+          - else
+            = name
+          %span.badge.ms-1{ class: badge_for_state(state) }= state.capitalize
+      .accordion-collapse.collapse{ class: expand?(name, state) ? 'show' : '', id: "diff-item-#{name.parameterize}" }
+        - if (diff = file_info['diff'])
```
dmarcoux, line 15: So if there's a diff, we should render DiffComponent. It sounds like this belongs in [DiffComponent#render?](https://github.com/openSUSE/open-build-service/wiki/View-Components#conditional-rendering) since I don't see a case where that component would render when there is no actual diff.

src/api/app/views/webui/package/_revision_diff_detail.html.haml
```diff
@@ -8,36 +8,7 @@
         .diff-card-header
           = calculate_filename(filename, file)
           %span.badge{ class: "badge-#{file['state']}" }= file['state'].capitalize
-      -# TODO: Remove this `if` condition, and the `else` clause once request_show_redesign is rolled out
```
dmarcoux, line 11: This comment doesn't apply anymore?

src/api/app/components/diff_component.html.haml
```diff
@@ -0,0 +1,21 @@
+.pre.scroll
+  .diff{ class: "digits-#{diff.max_digits}" }>
+    - diff.lines&.each do |line|
+      - id = "diff_#{file_index}_n#{line.index}"
+      %div{ class: "line #{line.state}", id: }>
```
dmarcoux, line 5: The `id` is missing here.

dmarcoux: Oh right, that syntax again. It always look like a bug.

danidoni (approved):

PR #13913, opened 2023-02-28
Add specs for BsRequestCommentComponent

saraycp:

### DIFF
--- a/src/api/spec/components/bs_request_comment_component_spec.rb
+++ b/src/api/spec/components/bs_request_comment_component_spec.rb
@@ -1,20 +1,77 @@
 require 'rails_helper'
 
 RSpec.describe BsRequestCommentComponent, type: :component do
-  context 'when we have a comment' do
-    context 'when we have a commentable' do
-      context 'when rendering the first level of a comment thread' do
-        it 'renders the comment'
-        it 'renders the child comments underneath'
-      end
-    end
+  let(:commentable) { create(:bs_request_with_submit_action) }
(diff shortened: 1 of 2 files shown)


src/api/spec/components/bs_request_comment_component_spec.rb
```diff
@@ -1,20 +1,77 @@
...
+
+  subject { render_inline(described_class.new(comment: comment_a, commentable: commentable, level: 1)) }
+
+  context 'when user is not logged in' do
+    it 'is not possible to reply a comment' do
+      expect(subject).not_to have_text('Reply')
+    end
+  end
+
+  context 'when user is logged in and the comment belongs to them' do
```
danidoni, line 15:
```suggestion
  context 'when user is logged in and is the author' do
```

src/api/spec/components/bs_request_comment_component_spec.rb
```diff
@@ -1,20 +1,77 @@
...
+    it 'is possible to edit a comment' do
+      expect(subject).to have_selector('.dropdown-menu', text: 'Edit')
+    end
+
+    it 'is possible to remove a comment' do
+      expect(subject).to have_selector('.dropdown-menu', text: 'Delete')
+    end
+  end
+
+  context 'when user is logged in and the comment does not belong to them' do
```
danidoni, line 33:
```suggestion
  context 'when user is logged in and it is not the author' do
```

danidoni (approved):

PR #13914, opened 2023-02-28
Run against the database consistency todo file

hennevogel:
No sense in having on if we don't run against it. It's not done by default.


hennevogel:
sorry @hellcp-work

PR #13916, opened 2023-02-28
[api] add service type definition schema file

adrianschroeter:


M0ses (approved):
tested with xmllint and obs-service-example service file

LGTM - approved

adrianschroeter:
@M0ses JFYI, this is a newer version which is a bit more relaxed then the old one (parameter order does not matter)

PR #13918, opened 2023-02-28
Add specs for BsRequestActivityTimelineComponent

saraycp:

### DIFF
--- a/src/api/spec/components/bs_request_activity_timeline_component_spec.rb
+++ b/src/api/spec/components/bs_request_activity_timeline_component_spec.rb
@@ -1,13 +1,16 @@
 require 'rails_helper'
 
 RSpec.describe BsRequestActivityTimelineComponent, type: :component do
-  context 'when we provide a bs_request' do
-    it 'shows an element telling who created the request and when'
-    it 'shows the comments threads'
-    it 'shows the history elements'
+  let(:bs_request) { create(:bs_request_with_submit_action) }
+  let!(:history_element) { create(:history_element_request_accepted, op_object_id: bs_request.id) }
+  let!(:comment) { create(:comment_request, commentable: bs_request, created_at: Time.zone.yesterday) }
+
+  it 'shows the comment first, as it is an older timeline item' do
+    expect(render_inline(described_class.new(bs_request: bs_request))).to have_selector('.timeline-item:first-child', text: 'wrote')
+    expect(render_inline(described_class.new(bs_request: bs_request))).to have_selector('.timeline-item:first-child', text: '1 day ago')
   end
 
-  context 'when we do not provide a bs_request' do
-    it 'raises an error warning the user to provide a request'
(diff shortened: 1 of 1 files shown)


src/api/spec/components/bs_request_activity_timeline_component_spec.rb
```diff
@@ -1,13 +1,28 @@
...
+    it 'shows the comments threads' do
+      expect(render_inline(described_class.new(bs_request: bs_request))).to have_text('accepted request')
+    end
+
+    it 'shows the history elements' do
+      expect(render_inline(described_class.new(bs_request: bs_request))).to have_text('wrote')
+      expect(render_inline(described_class.new(bs_request: bs_request))).to have_text('1 day ago')
+    end
+
+    it 'shows the comment first, as it is older timeline item' do
```
danidoni, line 24:
```suggestion
    it 'shows the comment first, as it is an older timeline item' do
```

src/api/spec/components/bs_request_activity_timeline_component_spec.rb
```diff
@@ -1,13 +1,24 @@
 require 'rails_helper'
 
 RSpec.describe BsRequestActivityTimelineComponent, type: :component do
-  context 'when we provide a bs_request' do
-    it 'shows an element telling who created the request and when'
-    it 'shows the comments threads'
-    it 'shows the history elements'
+  context 'when we do not provide a bs_request' do
+    it do
+      expect { render_inline(described_class.new) }.to raise_error(ArgumentError, 'missing keyword: :bs_request')
```
dmarcoux, line 6: I don't really see the benefit in testing how keyword arguments work. This is a core Ruby feature. I suggest removing this test.

saraycp: You are right. Changes pushed. I'm going to delete it from the other specs I added too.

danidoni:
Don't we test the presence of the avatar and the creation date?

saraycp:
> Don't we test the presence of the avatar and the creation date?

I don't think it should be part of this spec, as this component is in charge of displaying a comment or a history element in the right order. That's why I check the minimum info possible to simply know if a comment or a history element is displayed and in the right order.

Checking the details of the comments and history elements should be done in the specific components' specs. As I already did with the icon [in the specs for the history element](https://github.com/openSUSE/open-build-service/blob/bce6593e23921f66f54f8aedeb2fa15e882edf38/src/api/spec/components/bs_request_history_element_component_spec.rb#L17) I could create new PRs to check for more details...

danidoni:
> > Don't we test the presence of the avatar and the creation date?
> 
> I don't think it should be part of this spec, as this component is in charge of displaying a comment or a history element in the right order. That's why I check the minimum info possible to simply know if a comment or a history element is displayed and in the right order.

Okay!

danidoni (approved):

PR #13923, opened 2023-03-01
Fix arrangement of request show view HTML elements

eduardoj:
Fix including `<p>` elements inside other `<p>` elements in the requests and actions views. That results in an invalid HTML, and the HAML parser was preventing this from happening creating `<p></p>` groups of HTML elements.

A series of changes have been made in this pull request:
- Rearrange the title of the card and the text of the card, adjusting margins. Before, all the text belonged to the  `card-title` CSS class.
- Substitute `font-italic` with `fst-italic`. This should have been changed in the migration to Bootstrap 5.
- Wrap simple text not including `<p>` elements into `<p>` elements.
- Wrap the rest of the text groups into `<div>` elements with the following classes: `.d-flex.flex-wrap.mb-3.align-items-center`. That allowed to remove the use of `.align-middle`.

### DIFF
--- a/src/api/app/views/webui/request/_actions_details.html.haml
+++ b/src/api/app/views/webui/request/_actions_details.html.haml
@@ -3,8 +3,8 @@
   submit_actions_count = submit_actions.count
 
 - if submit_actions_count > 1
-  .d-flex.align-items-center
-    %p.mb-0 This request contains #{submit_actions_count} actions
+  .d-flex.flex-wrap.mb-3.align-items-center
+    This request contains #{submit_actions_count} actions
     .input-group.ms-2.w-auto
       = link_to('Previous', request_show_path(number: bs_request.number, request_action_id: prev_action&.id,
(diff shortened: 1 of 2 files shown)


src/api/app/views/webui/request/_actions_details.html.haml
```diff
@@ -3,8 +3,8 @@
   submit_actions_count = submit_actions.count
 
 - if submit_actions_count > 1
-  .d-flex.align-items-center
-    %p.mb-0 This request contains #{submit_actions_count} actions
+  .d-flex.flex-wrap.mb-3.align-items-center
```
eduardoj, line 6: @danidoni, I forgot to add this class `mb-3`. Please take a look again.

danidoni: LGTM

saraycp (approved):

PR #13929, opened 2023-03-02
Extend specs for BsRequestCommentComponent

saraycp:
I decided to test, at least once, the elements that are always present after this [conversation](https://github.com/openSUSE/open-build-service/pull/13918#issuecomment-1451664915).

So I extend the specs to do so and apply DRY rule too.

**Tip for reviewers:** don't compare with the previous version, review it as it was a new spec.

### DIFF
--- a/src/api/spec/components/bs_request_comment_component_spec.rb
+++ b/src/api/spec/components/bs_request_comment_component_spec.rb
@@ -2,49 +2,73 @@ require 'rails_helper'
 
 RSpec.describe BsRequestCommentComponent, type: :component do
   let(:commentable) { create(:bs_request_with_submit_action) }
-  let(:comment_a) { create(:comment_request, commentable: commentable, body: 'Comment A') }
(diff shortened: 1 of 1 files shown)


src/api/spec/components/bs_request_comment_component_spec.rb
```diff
@@ -2,49 +2,73 @@
...
+    expect(rendered_content).to have_text("#{comment_a.user.realname} (#{comment_a.user.login})\nwrote")
+  end
+
+  it 'displays the time when the comment happened in words' do
+    expect(rendered_content).to have_text('1 day ago')
+  end
+
+  it 'displays the comment in a comment bubble' do
+    expect(rendered_content).to have_selector('.comment-bubble', text: 'Comment A')
+  end
```
danidoni, line 29: I wouldn't go that far as to test this. The bubble is some styling that can change in the future without affecting the feature.

saraycp: :+1:  change pushed

danidoni (approved):

PR #13934, opened 2023-03-02
Enable autorefresh in .repo files

bmwiedemann:
`zypper addrepo` defaults to disabled autorefresh,
but in most OBS repos, rebuilds let old files vanish,
so often zypper install will try to fetch old non-existing files
and fail with a message such as

```
File './x86_64/libLLVM15-15.0.7-lp154.19.3.x86_64.rpm' not found on medium 'https://download.opensuse.org/repositories/devel:/tools:/compiler/openSUSE_Leap_15.4/'
Abort, retry, ignore? [a/r/i/...? shows all options] (a): r
```

This change would avoid user-frustration in these cases
by having zypper refresh repos before install.

As a downside, some zypper operations would take longer to start.


adrianschroeter:
people also complained about always refreshing before ....

In any case this is a behaviour change, this can not happen without a config option and documentation in any case.
(and the [backend] prefix is missing)

bmwiedemann:
In the obs-docu, I only found
https://github.com/openSUSE/obs-docu/blob/26c43838e9/xml/obs_osc.xml#L35
https://github.com/openSUSE/obs-docu/blob/26c43838e9/xml/art-obs-beginners-guide.xml#L1222

These seem to not be good places to go into details about autorefresh.

Conan-Kudo:
People complain about DNF doing auto-refresh, you think they're going to be happy when Zypper does it too? I don't think it'll go over well if you do this.

PR #13935, opened 2023-03-02
Watch button to the top-right

ncounter:
The `watch` button floats right after the title of the context (package | project | request) and this adds noise to the focus (the title) and takes space where sometimes there are badges or other information. Better to reserve a dedicated place for this action button, and detach it a little bit from the center of the page. This PR move it always to the top-right, still in line with the title of the context

### Package
package-watch-left
### Project
project-watch-left
### Request
request-watch-left

### Package
package-watch-right
### Project
project-watch-right
### Request
request-watch-right

### DIFF
--- a/src/api/app/components/watchlist_icon_component.html.haml
+++ b/src/api/app/components/watchlist_icon_component.html.haml
@@ -1,5 +1,5 @@
 - if object_to_be_watched_in_watchlist?
-  = link_to('#', class: 'btn btn-outline-secondary',
+  = link_to('#', class: 'btn btn-outline-secondary text-nowrap',
             title: "Remove #{watchable_type_text} from watchlist",
             data: { 'bs-toggle': 'modal',
                     'bs-target': '#delete-item-from-watchlist-modal',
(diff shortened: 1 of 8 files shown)


hennevogel:
Now we have action buttons and side links...

ncounter:
> Now we have action buttons and side links...

Is there maybe something I am missing about those side links?
If they are not properly placed, I'd agree. But side links are something that can go somewhere else or revisited, or they can even become something all together with the button.
Fact is, such an action button should not really stay there as it is now, and the drawback of having action button and side links there is a price I'd say we can pay for the pros we gain taking the button away from the title.
The main issue is living in the request case where there are badges and more next to the title. But even without that case, a button should not really stay there in the center of the page with the title. For consistency we want to move all the watch buttons to the top-right.

ncounter:
JFYI: I just find out these changes are breaking the `.in-place-editing` feature for the package and project case. I'll take a look

ncounter:
> JFYI: I just find out these changes are breaking the `.in-place-editing` feature for the package and project case. I'll take a look

Fixed now

danidoni (approved):

rubhanazeem (approved):

PR #13938, opened 2023-03-03
[api] support attribute mass update using PUT route

adrianschroeter:
this allows

 # osc api -e /source/$PROJECT/_attribute

for admins at least.

Also fixing multiple commit issues for the POST route.


eduardoj:
Would you mind modifying or adding the documentation to the OpenAPI specification? Thanks in advance.

PR #13940, opened 2023-03-03
Adjust colors after Bootstrap 5 migration

saraycp:
Before these changes, the background colors of the navigation areas were inconsistent. Something that started to happen after the migration to Bootstrap 5.
Compare the Watchlist bar and the top bar, for example.

Screenshot 2023-03-03 at 13-37-04 Open Build Service

After these changes, I bring back the dark gray color for all of the navigation areas:

Screenshot 2023-03-03 at 13-41-53 Open Build Service

**What happened with the migration?**

We were using `.bg-dark` for the background color of the navigation areas. However, `bg-dark` was `#343a40` in the past ([see v4.6 docu](https://getbootstrap.com/docs/4.6/utilities/colors/#background-color)), but now it is `#212529`, a darker color ([see v5.0 docu](https://getbootstrap.com/docs/5.0/customize/color/#theme-colors)).

Checking the new Sass variables I see Bootstrap still has both colors set in variables.
The new `bg-dark` (`bg-dark = $gray-900 = #212529`) and the color we used before  (`$gray-800 = #343a40`).

~That's why I now set `$gray-800` as the background color in `.navbar-dark` which is a class used by all the navigation areas and stop relying on `bg-dark`.~

That's why I now set a new Sass variable `.bg-gray-800` and use it, together with `.navbar-dark`, in all the navigation areas.

### DIFF
--- a/src/api/app/assets/stylesheets/webui/application.scss
+++ b/src/api/app/assets/stylesheets/webui/application.scss
@@ -74,6 +74,7 @@
 @import 'request_show_redesign/build_results';
 @import 'accordion-reviews-component';
 @import 'tokens';
+@import 'colors';
 
 html {
     overflow-y: scroll !important;
--- a/src/api/app/assets/stylesheets/webui/colors.scss
+++ b/src/api/app/assets/stylesheets/webui/colors.scss
@@ -1,0 +1,3 @@
+.bg-gray-800 {
+  background-color: $gray-800;
+}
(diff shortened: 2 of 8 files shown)


saraycp:
Good catch @hellcp-work. Please let me know what you think of the new changes.
I now set a new Sass variable `.bg-gray-800` and use it, together with `.navbar-dark`, in all the navigation areas.

PR #13941, opened 2023-03-03
 Request header section: reorganize information

ncounter:
Address https://trello.com/c/1Pg8ZhVO

## Before
image

## After
**New or Review**
new-or-review

**Revoke or declined**
revoke-or-declined

**Accepted**
accepted

**Multi action**
multi-action

**Superseeding**
superseding

**Staged**
staged

**Auto accept**
auto-accept

**Superseded**
superseded

### DIFF
--- a/src/api/app/components/action_seen_by_user_component.html.haml
+++ b/src/api/app/components/action_seen_by_user_component.html.haml
@@ -2,7 +2,7 @@
   = render_icon_status
 
 - else
-  - button_text = 'Action seen'
+  - button_text = 'Seen'
 
   = link_to(toggle_request_path(action_id: @action.id),
       class: 'btn btn-sm btn-outline-info',
--- a/src/api/app/components/fuzzy_time_component.rb
+++ b/src/api/app/components/fuzzy_time_component.rb
@@ -9,8 +9,13 @@ class FuzzyTimeComponent < ApplicationComponent
   end
 
   def human_time_ago
-    return 'now' if (Time.now.utc - time) < 60
+    now = Time.now.utc
+    diff = now - time
 
-    "#{time_ago_in_words(time)} ago"
+    if diff.positive?
+      "#{distance_of_time_in_words(now, time)} ago"
+    else
+      "in #{distance_of_time_in_words(now, time)}"
+    end
   end
 end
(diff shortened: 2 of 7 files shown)


src/api/app/components/action_seen_by_user_component.html.haml
```diff
@@ -2,12 +2,14 @@
...
-  - button_text = 'Action seen'
+  - button_text = 'Seen'
 
   = link_to(toggle_request_path(action_id: @action.id),
-      class: 'btn btn-sm btn-outline-info',
+      class: 'btn btn-sm text-info',
       title: 'Whether the action has been marked "as seen" or not yet',
       method: :put, remote: true,
       'data-disable-with': tag.i(class: 'fa-solid fa-spinner fa-spin') + button_text) do
+    (
```
krauselukas, line 12: @ncounter to me the brackets around the seen check mark make it look like a 'side' info. but it is an action. I would remove them.

krauselukas: maybe reintroducing the outline, but very light? Since the `Action details` header is very strong the `Seen` checkmark is to much in the background. If I don't know its there I have to look twice in order to see it

src/api/app/components/fuzzy_time_component.rb
```diff
@@ -9,8 +9,14 @@ def initialize(time:)
...
+    now = Time.now.utc
+    diff = now - time
+    return 'now' if diff.abs < 60
 
-    "#{time_ago_in_words(time)} ago"
+    if diff.positive?
+      "#{time_ago_in_words(time)} ago"
+    else
+      "in #{distance_of_time_in_words(now, time)}"
+    end
```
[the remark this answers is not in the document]
ncounter: I think you are correct actually. The only difference is a matter of what the method returns.
After trying it out with different options I'd say we can make a single use of `distance_of_time_in_words` for everything: it does not have the `now` option anymore, it will be `less than a minute` but I think it is even better. I'll align the method accordingly. :+1:

src/api/app/components/action_seen_by_user_component.html.haml
ncounter: I see, thanks for the feedback. I think I agree with you it still misses some highlight. I'll bring back the outline (as it is a button in the end). :+1:

ncounter: I addressed it and updated the screenshot in the PR description

danidoni (approved):

danidoni (approved):
LGTM

PR #13942, opened 2023-03-03
Show all nokogiri error messages, instead of only one

### LINKED ISSUE #13925: project meta: insufficient validation error message
# Issue Description
Project meta validation gives no hint on what's wrong with the data.

Expected Result
============
Inform about which field failed to validate and what is the invalid value.

How to Reproduce
===============
Send a PUT request to `source/<project>/_meta` with an invalid architecture:
```
  <repository name="repo">
    <path project="project" repository="standard" />
    <arch>foo</arch>
  </repository>
```
project validation error: 15:0: ERROR: Error validating value 

^^^ No clue what's wrong.

eduardoj:
Fixes #13925.


danidoni (approved):

src/api/app/lib/suse/validator.rb
```diff
@@ -98,18 +98,22 @@ def validate(opt, content)
...
         end
-        true
+
+        nokogiri_xml_syntaxerrors = schema.validate(doc)
+
+        return true if nokogiri_xml_syntaxerrors.empty?
+
+        error_string = nokogiri_xml_syntaxerrors.join("\n")
+
+        logger.error "validation error: #{error_string}"
```
hennevogel, line 114: For very wrong files this might blow up logging now. I think this should be debug logging too.

hennevogel (changes requested):

src/api/app/lib/suse/validator.rb
eduardoj: You mean "instead of" `logger.error`, or "also" `logger.debug`?

A code suggestion could have clarified what exactly do you mean.

hennevogel, line 114:
```suggestion
        logger.debug "validation error: #{error_string}"
```

eduardoj: Applied. Thanks!

hennevogel:
I don't think this fixes 13925 as for this case it's just the same information in long form...

PR #13912, opened 2023-02-28
Allow commenting on submit request diff lines

hellcp-work:
Depends on #13910 

Allows users to comment on submit request action diffs.

Screenshot from 2023-02-28 13-06-33

# How to test

1. Create a submit request with a diff
2. Open changes tab
3. Hover over a line
4. Click a button that shows up
5. Write a comment and save it

PR #13953, opened 2023-03-06
Instrument the diff comments feature

rubhanazeem:
Instrumenting the feature introduced in #13912

### DIFF
--- a/src/api/app/models/comment.rb
+++ b/src/api/app/models/comment.rb
@@ -13,8 +13,8 @@ class Comment < ApplicationRecord
 
   validate :validate_parent_id
 
-  after_create :create_event
-  after_destroy :delete_parent_if_unused
+  after_create :create_event, :instrument_diff_comments_on_create
+  after_destroy :delete_parent_if_unused, :instrument_diff_comments_on_destroy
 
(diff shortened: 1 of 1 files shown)


src/api/app/models/comment.rb
```diff
@@ -13,8 +13,8 @@ class Comment < ApplicationRecord
 
   validate :validate_parent_id
 
-  after_create :create_event
-  after_destroy :delete_parent_if_unused
+  after_create :create_event, :instrument_diff_comments_on_create
+  after_destroy :delete_parent_if_unused, :instrument_diff_comments_on_destroy
```
hennevogel, line 17: Why distinguish by state?

rubhanazeem: To get the total number of times the feature has been used. `diff_comment` values are different for both states `(add/remove)`.

hennevogel (changes requested):
We usually instrument the events (like `Event::CommentEvent`) and not some single operation

krauselukas:
@hennevogel @rubhanazeem I took this over, I went the route through the event system. Going to close this PR. Replaced by https://github.com/openSUSE/open-build-service/pull/13969

PR #13964, opened 2023-03-08
Introduce limiting results for /search/... endpoints

eduardoj:
WIP

## TODO
- [ ] Add documentation for the new error message


danidoni:
What about adding some specs?

bmwiedemann:
For spec: set limit=2 in config, test that 
* returning 1 result works
* returning 3 results causes 403
* returning 3 result ids works

danidoni (approved):

src/api/app/controllers/search_controller.rb
```diff
@@ -196,6 +196,19 @@ def search(what, render_all)
     items = find_items(what, predicate)
 
     matches = items.size
+    if render_all && search_results_exceed_configured_limit?(matches)
```
hennevogel, line 199: Sounds backwards to me somehow. Passing around `matches` instead of something like

```
if matches > search_results_limit
....
```

src/api/app/controllers/search_controller.rb
```diff
@@ -275,4 +288,16 @@ def find_items(what, predicate)
...
+    config_limit = CONFIG['limit_for_search_results']
+    return false if config_limit.blank?
+
+    params_limit = params[:limit].present? && params[:limit] =~ /\A\d+\z/ ? params[:limit].to_i : nil
+
+    returned_results = params_limit.present? && params_limit < matches ? params_limit : matches
+    return false if returned_results <= config_limit
+
+    true
+  end
```
hennevogel, line 302: I think you need to untangle `params_limit` from here and do this somewhere else. It has little to do with the configured limit IMHO.

hennevogel (changes requested):

src/api/app/controllers/search_controller.rb
eduardoj: `params_limit` has to be taken in account with matches and the configured limit. I'll try to illustrate that with two examples. First example:
- matches: 100
- configured limit: 50
- `params_limit`: 10

In this case, the search results (not returned by matches, but by `params_limit`) don't exceed the configured limit.

In this other example:
- matches: 100
- configured limit: 50
- `params_limit`: 70

In this case, the search results (not returned by matches, but by `params_limit`) **do** exceed the configured limit. That's why `params_limit` is included here, and has to do with the configured limit.

src/api/app/controllers/search_controller.rb
eduardoj: I agree it sounds somehow backwards. Unfortunately, the logic involved is not as straightforward as your snippet, and it is all abstracted in the call to the `search_results_exceed_configured_limit?` method.

Do you have a better/more readable suggestion, which performs the same amount of checks?

src/api/app/controllers/search_controller.rb
hennevogel: Sure in the case of `params_limit` existing, `params_limit` becomes `matches` right? That doesn't mean you have to do it in this method though....

eduardoj: > Sure in the case of `params_limit` existing, `params_limit` becomes `matches` right? That doesn't mean you have to do it in this method though....

Not always. I still think we have to do it in this method. If the config limit is not defined, it is irrelevant if `params_limit` is defined or not. In other words: `params_limit` becomes `matches` only when config limit is defined. We should not make `matches` becoming `params_limit` if config limit is not defined.

We could extract some logic, from inside the method to some other method. But I don't think the result would be more clear. I would prefer to leave the method as it is.

Obviously, I could be missing something. Please provide an alternative if you have something better in mind.

src/api/app/controllers/search_controller.rb
eduardoj: This line:

```
records_to_show = limit || items.size
```
looks wrong to me. `limit` could be defined and be greater than `items.size`, but we still should assign `items.size` to `records_to_show`.

hennevogel: We talked offline and in the controllers current form there isn't a way to do this that isn't weird. So let's take on some technical dept for the refactoring of this controller...

hennevogel (approved):

PR #13966, opened 2023-03-08
Document /source/{project_name}/{package_name} endpoints with OpenAPI

eduardoj:


eduardoj:
I'll split this PR in two.

PR #13967, opened 2023-03-08
Correct commentable type in `create_event` callback on comment creation

krauselukas:
The commentable type is `BsRequestAction` not `BsRequestActionSubmit`. Since it wasn't matching in the `create_event` method, no events got created for comments that have a relation to a request action.

This was introduced in https://github.com/openSUSE/open-build-service/pull/13912 for the comments on request diffs.

### DIFF
--- a/src/api/app/models/comment.rb
+++ b/src/api/app/models/comment.rb
@@ -60,7 +60,7 @@ class Comment < ApplicationRecord
       Event::CommentForProject.create(event_parameters)
     when 'BsRequest'
       Event::CommentForRequest.create(event_parameters)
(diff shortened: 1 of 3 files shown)


krauselukas:
https://github.com/openSUSE/open-build-service/blob/5955a57c0f1de863fd240cbf0799b4ad8f9d655a/src/api/app/models/bs_request_action.rb#L15

danidoni:
Can we add some specs to this?

krauselukas:
> Can we add some specs to this?

@danidoni done, please have another look :)

danidoni (approved):

PR #13973, opened 2023-03-10
Document package commitfilelist command with OpenAPI

bmwiedemann:
### For reviewers

Access to the documented enpoint in the review-app through [this link](https://obs-reviewlab.opensuse.org/bmwiedemann-dod1byy0/apidocs-new/index.html#/Sources%20-%20Packages/post_source__project_name___package_name__cmd_commitfilelist).

### DIFF
--- a/src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_commitfilelist.yaml
+++ b/src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_commitfilelist.yaml
@@ -1,0 +1,112 @@
+post:
+  summary: Commit package changes from a filelist.
+  description: |
+    Create a new commit using defined files already uploaded to the repository.
+    
+    This requires a directory xml providing file name and md5 as body.
+    It is recommended to use the `withvalidate` parameter and add the
+    sha256 sum including "sha256:" prefix in "hash" attribute for security reasons.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - $ref: '../components/parameters/package_name.yaml'
+    - in: query
+      name: comment
+      schema:
+        type: string
+      description: Provide a comment to the revision
+      example: This was changed to improve foobar.
+    - in: query
+      name: linkrev
+      schema:
+        type: string
+      description: |
+        Link revision in base package. Set to `base` to use the commit revision.
+
+        Used together with the `keeplink` parameter.
+      example: base
+    - in: query
+      name: rev
+      schema:
+        type: string
+      description: Revision of the package.
+    - in: query
+      name: keeplink
+      schema:
+        type: string
+        enum:
+          - 1
+          - 0
+      default: 0
+      description: Set to `1` to preserve the link (`_link` file) on the source.
+      example: 1
+    - in: query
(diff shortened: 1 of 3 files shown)


src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_commitfilelist.yaml
```diff
@@ -0,0 +1,114 @@
+post:
+  summary: Commit package changes.
```
eduardoj, line 2: `Commit package changes.` is the same description as for the `POST
/source/{project_name}/{package_name}?cmd=commit` endpoint. Do they both perform the same task? Can we modify the summary to make it easier to know what is each endpoint for?

Screenshot from 2023-03-10 14-00-47

src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_commitfilelist.yaml
```diff
@@ -0,0 +1,113 @@
+post:
+  summary: Commit package changes from a filelist.
+  description: |
+    Create a new commit using defined files already uploaded to the repository.
+    This can be done via `PUT` using `rev=repository` which avoids a commit.
```
eduardoj, line 5: I find this line confusing, even if it copied from /apidocs, because there is only a mention to the action (`PUT`) but not to the path.

Could we clarify what API endpoint (action with path) are we referring to here?

It would also be fine for me if this line is removed.

src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_commitfilelist.yaml
```diff
@@ -0,0 +1,112 @@
...
+          - 1
+          - 0
+      default: 0
+      description: |
+        Set to `1` to solve conflicts.
+
+        Used together with `keeplink` parameter.
+      example: 1
+    - in: withvalidate
+      name: keeplink
```
eduardoj, line 58:
```suggestion
    - in: query
      name: withvalidate
```
Typos.

eduardoj:
For the parameters `keeplink`, `repairlink` and `withvalidate` I would remove these sections:
```
        enum:
          - 1
          - 0
      default: 0
```
They are parameters which are treated as `:bool?` in the [backend](https://github.com/openSUSE/open-build-service/blob/a3bd82a38dfba6ebef167057bf4da8868cab9a94/src/backend/bs_srcserver#L7620) , and most probably (not checked, though) they would accept a value of `0` as thruty. We shouldn't document them as `enum` with `0` and `1` values. What do you think?

bmwiedemann:
https://github.com/openSUSE/open-build-service/blob/a3bd82a38dfba6ebef167057bf4da8868cab9a94/src/backend/bs_srcserver#L3039 shows how it is used. That would interpret a `"0"` or `""` as false and everything else as true. Still makes sense to declare `0` and `1` as the only supported inputs.

eduardoj:
> That would interpret a `"0"` or `""` as false and everything else as true. Still makes sense to declare `0` and `1` as the only supported inputs.

I see. Fine to keep the ` enum:` section then.

eduardoj (approved):

PR #13974, opened 2023-03-10
Document GET `/source/{project_name}/_attribute` endpoint

saraycp:
We already documented the `GET /source/{project_name}/_attribute/{attribute_name}` endpoint. But `GET /source/{project_name}/_attribute` is also accepted but wasn't documented until now.

The endpoint documented in this PR is in charge of listing all the attributes of a project.

**Testing Tips**

- Go to the Attributes tab on the project show page.
- Add some attributes
- Run `curl -u foo:bar -H 'Content-Type: application/xml; charset=utf-8' -X GET 'http://localhost:3000/source/home:Admin/_attribute`

### DIFF
--- a/src/api/public/apidocs-new/paths/source_project_name_attribute.yaml
+++ b/src/api/public/apidocs-new/paths/source_project_name_attribute.yaml
@@ -1,0 +1,62 @@
+get:
+  summary: Get all the project's attributes
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    # Query Strings:
+    # `rev`: doesn't make sense in project level and makes the endpoint crash.
+    # `meta`: either you pass `meta` or not, it's always replaced by meta: 1.
+    # `with_project`: does not make sense in project level and makes the endpoint crash.
+    - name: view
+      in: query
+      schema:
+        type: string
+        enum:
+          - blame
+      description: With view=blame, information about who changed each XML tag and when is displayed.
+      example: blame
+    - name: with_default
+      in: query
+      schema:
+        type: string
+      description: |
+        If the attribute doesn't contain any value and `with_default` is present,
+        the default values will be displayed, if any.
+      example: 1
+  responses:
+    '200':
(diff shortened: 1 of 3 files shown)


src/api/public/apidocs-new/paths/source_project_name_attribute.yaml
```diff
@@ -0,0 +1,54 @@
+get:
+  summary: Get all the project's attributes
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - name: meta
+      in: query
+      schema:
+        type: string
```
eduardoj, line 8: I miss for these parameters (meta, rev, view, with_default, and with_project):
- to fill out the `description:` field
- a definition of values they could have. For example, if they are booleans
- an example value

src/api/public/apidocs-new/paths/source_project_name_attribute.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - name: rev
+      in: query
+      schema:
+        type: string
+    - name: view
+      in: query
+      schema:
+        type: string
+      description: The value of this parameter should only be `blame`
```
eduardoj, line 17: When `view=blame` is passed, is the response of the endpoint using the same `source/attributes.yaml` schema?

If not, it could help for completeness to add an example in the `responses:` section.

saraycp: The result is not pure XML but the content type we receive is still `Content-Type: application/xml; charset=utf-8`, so I add an example so the user knows the structure thanks to the value of the example.

src/api/public/apidocs-new/paths/source_project_name_attribute.yaml
saraycp: Done.
I documented only the query parameters that can be used at this project level. Some others don't make sense or make the code crash, they are commented on but not documented.

eduardoj (approved):

PR #13976, opened 2023-03-10
Use separate controller method for deprecated search endpoint routes

krauselukas:
The routes '/search/package_id' and '/search/project_id' are deprecated but still using the same controller methods as the none deprecated ones. In order to track the usage through influxdb-rails and for later removal we can simply point those routes to separated controller methods.


krauselukas:
[Instrument deprecated API endpoints (M)](https://trello.com/c/DInrA61z/2142-instrument-deprecated-api-endpoints-m)

eduardoj (approved):

PR #13977, opened 2023-03-10
Document `DELETE /source/{project_name}/{package_name}` with OpenAPI

eduardoj:


src/api/public/apidocs-new/paths/source_project_name_package_name.yaml
```diff
@@ -0,0 +1,57 @@
+delete:
+  summary: Deletes a specified package including all its source files.
+  description: Deletes a specified package including all its source files, if no other packages use this package as devel package.
```
krauselukas, line 3:
```suggestion
  description: Deletes a specified package including all its source files. By default only if no other packages use this package as a devel package.
```

krauselukas: I would maybe say by default, since you can force it

krauselukas (approved):

PR #13995, opened 2023-03-14
Document POST `/source/{project_name}/{package_name}?cmd=collectbuildenv` endpoints

hellcp-work:
Adds documentation for 2 out of 3 endpoints from my card. I found out that the third endpoint, `?cmd=wipe` isn't even properly defined, so there's no way to use it, not to mention documenting it.

`?cmd=rebuild` has broken repository fetching, so setting a repo value causes an error 500, it works just fine if you don't set repo value. It should be considered for deprecation as mentioned in https://github.com/openSUSE/open-build-service/blob/ba63db9bb6e0bdcbe44205725a6daa5544007912/src/api/app/controllers/source_controller.rb#L849

For reviewers:
https://obs-reviewlab.opensuse.org/hellcp-work-apidocs-sources-package-builds/apidocs-new/index#/Sources%20-%20Packages/post_source__project_name___package_name__cmd_collectbuildenv

### DIFF
--- a/src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_collectbuildenv.yaml
+++ b/src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_collectbuildenv.yaml
@@ -1,0 +1,83 @@
+post:
+  summary: Collect build environment information.
+  description: |
+    Creates _buildenv files based on origin package builds. This can be used
+    to re-use exact older build enviroment even when further new binary packages
+    got added. For example to re-build an old maintenance update in the same way.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+    - $ref: '../components/parameters/package_name.yaml'
+    - in: query
+      name: oproject
+      required: true
+      schema:
+        type: string
+      description: Origin project. The project that the Origin Package will have the build information copied from.
+    - in: query
+      name: opackage
+      schema:
+        type: string
+      description: Origin package. Build environment information of the package that will be copied.
+    - in: query
+      name: orev
+      schema:
+        type: string
+      description: Origin package revision. Revision of the Origin Package on which to base the build environment collection.
+    - in: query
+      name: comment
+      description: Comment for the new revision.
+  responses:
+    '200':
+      description: OK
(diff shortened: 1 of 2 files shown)


src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_collectbuildenv.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - in: query
+      name: comment
+      description: Comment for the new revision.
+  responses:
+    '200':
+      description: OK
+      content:
+        application/xml; charset=utf-8:
+          schema:
+            $ref: '../components/schemas/revision.yaml'
```
eduardoj, line 38: The current apidocs states that the answer follows the status.xsd schema, [here](https://build.opensuse.org/apidocs/index#64). I haven't done any test of this enpoint, but, does it really answer with a revision element?

I have checked we don't have a revision schema, only the revisionlist schema. In case the response follows the status schema, we should mention the schema used in the `description:` field.

src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_collectbuildenv.yaml
```diff
@@ -0,0 +1,56 @@
...
+                rev: 3
+                vrev: 3
+                srcmd5: d41d8cd98f00b204e9800998ecf8427e
+                version: 20220902.37b45c2
+                time: 1678785078
+                user: Admin
+                comment: Copying the build environment from origin
+                requestid:
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
```
eduardoj, line 51: I miss the '400' (with a missing parameter, for example) and '403' (not enough permissions) errors. If an endpoint returns with an specific error, we should list them here, and provide at least an example.

src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_collectbuildenv.yaml
eduardoj: I see. Fine.

eduardoj (approved):

PR #13998, opened 2023-03-14
Deprecate broken endpoints source rebuild and wipe

hellcp-work:
The following two endpoints are broken and have working alternatives on the build side of the api, so we are deprecating them.

### DIFF
--- a/src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_rebuild.yaml
+++ b/src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_rebuild.yaml
@@ -1,0 +1,9 @@
+post:
+  deprecated: true
+  summary: Trigger a package rebuild.
+  description: |
+    Triggers a rebuild of the package. This endpoint
+    is broken and will be removed.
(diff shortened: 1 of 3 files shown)


src/api/public/apidocs-new/paths/source_project_name_package_name_cmd_rebuild.yaml
```diff
@@ -0,0 +1,9 @@
+post:
+  deprecated: true
+  summary: Trigger a package rebuild.
+  description: |
+    Triggers a rebuild of the package. This endpoint
+    is broken and will be removed.
+    See [https://github.com/openSUSE/open-build-service/issues/13997](https://github.com/openSUSE/open-build-service/issues/13997).
```
hennevogel, line 7: I don't think we need to point to github issues here. Instead you should tell people what to use instead 😸

hennevogel (changes requested):

hennevogel (approved):

eduardoj (approved):

PR #14012, opened 2023-03-15
Update bundled gem rack version

danidoni:
We need to point obs-server to the updated rack version


danidoni:
The obs-server build failures are expected because the spec file is pulling the rack version from the 2.10 branch, which is the one we're updating with this PR

hennevogel (approved):

PR #14021, opened 2023-03-16
Drop broken parameter `limit` in `source` `get_file`

ncounter:
Code line https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/source_controller.rb#L315 seems to recognize `limit` as a parameter, but hitting the endpoint with something like
```
http://localhost:3000/source/home:Admin/hello_world/new_file?limit=10
```
returns
```
<status code="400">
  <summary>unknown parameter 'limit'</summary>
</status>
```

This PR drops the parameter as it looks like a stale useless one.

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -312,7 +312,7 @@ class SourceController < ApplicationController
     end
 
     path = Package.source_path(project_name, package_name, file)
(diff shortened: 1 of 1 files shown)


ncounter:
Eventually, the `limit` parameter is used by other code path, i.e. it get hit by `/source/{project}/{package}/_history?limit=1`.
Closing this PR.

PR #14053, opened 2023-03-22
Phase out Timecop

eduardoj:
Replace calls to timecop rubygem methods with calls to ActiveSupport::Testing::TimeHelpers methods.

There were tests where these calls could be removed. Freezing or not freezing the time didn't make a difference.

Remove the timecop rubygem.


danidoni (approved):

ncounter (approved):
:+1:

PR #14055, opened 2023-03-23
Fix sorting of files in the changes tab

hellcp-work:
The sorting there was incorrect compared to the old diffs, this makes the sorting correct.

### DIFF
--- a/src/api/app/components/sourcediff_component.html.haml
+++ b/src/api/app/components/sourcediff_component.html.haml
@@ -23,7 +23,8 @@
             %h4
               #{diff_label(sourcediff['new'])} – #{diff_label(sourcediff['old'])}
           - if sourcediff['filenames'].present?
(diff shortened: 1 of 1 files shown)


src/api/app/components/sourcediff_component.html.haml
```diff
@@ -23,7 +23,8 @@
             %h4
               #{diff_label(sourcediff['new'])} – #{diff_label(sourcediff['old'])}
           - if sourcediff['filenames'].present?
-            = render(DiffListComponent.new(diff_list: sourcediff['files'], view_id: @action[:spkg], commentable:))
+            - diff_list = sourcediff['files'].sort_by { |k, _v| sourcediff['filenames'].find_index(k) }.to_h
+            = render(DiffListComponent.new(diff_list:, view_id: @action[:spkg], commentable:))
```
eduardoj, line 27: What about including the `sort_by` line into the `DIffListComponent` component. Does it make sense?

PR #14058, opened 2023-03-23
Fix diff comments not working in multi action requests

hellcp-work:
Found out that the wrong param was used for multi action requests, causing the comments to be created on the first request instead of the one that was visible. This fixes that issue

### DIFF
--- a/src/api/app/lib/routes_helper/routes_constraints.rb
+++ b/src/api/app/lib/routes_helper/routes_constraints.rb
@@ -20,7 +20,9 @@ module RoutesHelper
       repository_publish_build_id: %r{[^/]*},
       workflow_project: %r{[^/]*},
       staging_project_name: %r{[^/]*},
-      staging_project_copy_name: %r{[^/]*}
+      staging_project_copy_name: %r{[^/]*},
+      request_action_id: /\d*/,
+      line: /diff_\d*_n\d*/
     }.freeze
   end
 end


src/api/config/routes/webui_routes.rb
```diff
@@ -297,7 +297,7 @@
       post 'request/set_bugowner_request' => :set_bugowner_request
       get 'request/:number/request_action/:id' => :request_action, as: 'request_action'
       get 'request/:number/request_action/:id/changes' => :request_action_changes, as: 'request_action_changes'
-      get 'request/:number/request_action/:id/inline_comment/:line' => :inline_comment, constraints: cons, as: 'request_inline_comment'
+      get 'request/:number/request_action/:request_action_id/inline_comment/:line' => :inline_comment, constraints: cons, as: 'request_inline_comment'
```
eduardoj, line 300: Maybe a little late for the `:line` parameter, but for the new `:request_action_id` we could add a constraint, similar to the one used for `:id`, here: https://github.com/openSUSE/open-build-service/blob/6daa0b5f3f17db4015e408054a016a4d69f884d0/src/api/app/lib/routes_helper/routes_constraints.rb#L7

eduardoj: I saw the new commit, for the `:line` parameter. Fine.

What about the new `:request_action_id` parameter? I was expecting to see something like this added:
```
  request_action_id: /\d*/,
```

saraycp (approved):

src/api/config/routes/webui_routes.rb
eduardoj: If I'm not wrong `/\d*/` accepts an empty `request_action_id`.

eduardoj (approved):

PR #14059, opened 2023-03-23
Split out the diff subject into its own component

hellcp-work:
I want to use this component for displaying diff comments in the timeline


danidoni:
Codecov is stalled

danidoni (approved):

PR #14070, opened 2023-03-27
Show supported action types

eduardoj:
Before, in a multiaction submit request, a supported action was not shown if it was the only action supported and it was not the first action.

Now, in those conditions above mentioned, show the supported action.

For example, for a multiaction submit request where a Delete action is the only supported action, and also the last action (I was doing tests in my development environment, and the Delete action will be soonish supported), this is how it looked like before:

Screenshot from 2023-03-27 11-25-20

And this is after:

Screenshot from 2023-03-27 11-24-49

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -5,6 +5,10 @@ class Webui::RequestController < Webui::WebuiController
   # requests do not really add much value for our page rank :)
   before_action :lockout_spiders
   before_action :require_request, only: [:changerequest, :show, :request_action, :request_action_changes, :inline_comment]
+  before_action :set_actions, only: [:show], if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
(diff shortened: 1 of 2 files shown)


eduardoj:
> I have a feeling we should move setting action id to before_action, because these changes break
> 
> https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/request_controller.rb#L332-L341
> 
> which is pretty much the same code

Ops, I see. Working on it...

eduardoj:
> > I have a feeling we should move setting action id to before_action, because these changes break
> > https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/webui/request_controller.rb#L332-L341
> > which is pretty much the same code
> 
> Ops, I see. Working on it...

I applied the same solution with `before_action` methods and using them in the `inline_comment` action. Please have a look again.

rubhanazeem (approved):

danidoni (approved):

PR #14072, opened 2023-03-27
Performance improvements for beta request show 

rubhanazeem:
Removing the partial improves performance significantly:

With partial, the rendering time is ~8.6 seconds
Without partial, the rendering time is ~2.8 seconds

This is the package we have used to test: https://build.opensuse.org/package/show/openSUSE:Factory/python-django-axes

### DIFF
--- a/src/api/app/components/diff_component.html.haml
+++ b/src/api/app/components/diff_component.html.haml
@@ -29,4 +29,11 @@
                            locals: { comment: Comment.new, commentable: commentable, diff_ref: id })
             - else
               .comments-list
-                = render(partial: 'webui/request/add_inline_comment', locals: { commentable: commentable, diff_ref: id })
+                -# rubocop:disable ViewComponent/AvoidGlobalState
+                - if User.session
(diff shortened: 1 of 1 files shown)


src/api/app/components/diff_component.html.haml
```diff
@@ -29,4 +29,8 @@
                            locals: { comment: Comment.new, commentable: commentable, diff_ref: id })
             - else
               .comments-list
-                = render(partial: 'webui/request/add_inline_comment', locals: { commentable: commentable, diff_ref: id })
+                = link_to(request_inline_comment_path(commentable.bs_request, commentable, id),
```
eduardoj, line 32: In the partial there is a check for logged in users, and that check got removed in the changes above:

https://github.com/openSUSE/open-build-service/blob/aec566d2ec27ee6f2e3c97d8ff0ab88035a81c19/src/api/app/views/webui/request/_add_inline_comment.html.haml#L1

I guess this is for showing diffs only for authenticated users. Could you please check if the former behaviour hasn't changed?

eduardoj (approved):

PR #14076, opened 2023-03-27
Remove POST /build/{project_name}/{repository_name} API route

eduardoj:
... from the frontend router.

It always fails with a 404 error and it is not defined in the backend. See: https://github.com/openSUSE/open-build-service/blob/f8f8c2331da85ea844f0ec81f8b883e3ae123491/src/backend/bs_srcserver#L7725

### DIFF
--- a/src/api/test/functional/source_controller_test.rb
+++ b/src/api/test/functional/source_controller_test.rb
@@ -785,7 +785,7 @@ class SourceControllerTest < ActionDispatch::IntegrationTest
     assert_response 403
     post '/build/home:Iggy?cmd=wipe'
     assert_response 403
-    post '/build/home:Iggy/TestLinkPack?cmd=wipe'
+    post '/source/home:Iggy/TestLinkPack?cmd=wipe'
     assert_response 403
 
     # check branching from a locked project
--- a/src/api/config/routes/api_routes.rb
+++ b/src/api/config/routes/api_routes.rb
@@ -376,7 +376,7 @@ OBSApi::Application.routes.draw do
   match 'build/:project/:repository/:arch(/:package)' => 'build#index', constraints: cons, via: [:get, :post]
   get 'build/_result' => 'build#scmresult', constraints: cons
   get 'build/:project/_result' => 'build#result', constraints: cons
-  match 'build/:project/:repository' => 'build#index', constraints: cons, via: [:get, :post]
+  get 'build/:project/:repository' => 'build#index', constraints: cons
   match 'build/:project' => 'build#project_index', constraints: cons, via: [:get, :post, :put]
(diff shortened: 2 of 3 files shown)


danidoni (approved):

src/api/test/functional/source_controller_test.rb
```diff
@@ -785,7 +785,7 @@ def test_lock_project
     assert_response 403
     post '/build/home:Iggy?cmd=wipe'
     assert_response 403
-    post '/build/home:Iggy/TestLinkPack?cmd=wipe'
+    post '/source/home:Iggy/TestLinkPack?cmd=wipe'
```
adrianschroeter, line 788: these are here for backward compability reasons. Old remote OBS instance still may use them

eduardoj: > these are here for backward compability reasons. Old remote OBS instance still may use them

AFAIK, the backend never supported the `cmd` parameter in this GET line: https://github.com/openSUSE/open-build-service/blob/7ffd60ad56d87ee52d6816503eb23fae36109c3b/src/backend/bs_srcserver#L7725

Neither has supported paths like `POST:/build/$project/$repository cmd:`.

If this route ever existed in the backend and later on was removed, keeping a test for compability only for the permission case doesn't make much sense. Either tests are provided to test full compability, or they are removed.

adrianschroeter:
On Dienstag, 28. März 2023, 14:26:13 CEST Eduardo J. wrote:
> @eduardoj commented on this pull request.
> 
> 
> 
> > @@ -785,7 +785,7 @@ def test_lock_project
>      assert_response 403
>      post '/build/home:Iggy?cmd=wipe'
>      assert_response 403
> -    post '/build/home:Iggy/TestLinkPack?cmd=wipe'
> +    post '/source/home:Iggy/TestLinkPack?cmd=wipe'
> 
> > these are here for backward compability reasons. Old remote OBS instance still may use them
> 
> AFAIK, the backend never supported the `cmd` parameter in this GET line: https://github.com/openSUSE/open-build-service/blob/7ffd60ad56d87ee52d6816503eb23fae36109c3b/src/backend/bs_srcserver#L7725
> 
> Neither has supported paths like `POST:/build/$project/$repository cmd:`.
> 
> If this route ever existed in the backend and later on was removed, keeping a test for compability only for the permission case doesn't make much sense. Either tests are provided to test full compability, or they are removed.

yes, the wipe command indeed makes no sense on interconnect topics.

on the other side, why do you make use possible incompatible now?
Wouldn't it make more sense to mark this as to be cleaned up for 3.0 ?

-- 

Adrian Schroeter ***@***.***>

PR #14077, opened 2023-03-28
[api] show review creation time again

adrianschroeter:
OBS-215

### DIFF
--- a/src/api/app/models/review.rb
+++ b/src/api/app/models/review.rb
@@ -160,6 +160,11 @@ class Review < ApplicationRecord
     rescue TypeError
       # no valid time -> ignore
     end
+    begin
+      r.created_at = Time.zone.parse(hash.delete('created'))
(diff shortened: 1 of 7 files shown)


src/api/app/models/review.rb
```diff
@@ -159,6 +159,11 @@ def self.new_from_xml_hash(hash)
     rescue TypeError
       # no valid time -> ignore
     end
+    begin
+      r.created_at = Time.zone.parse(hash.delete('created'))
```
hennevogel, line 163: Why would we allow to change this?

hennevogel (changes requested):

src/api/app/models/review.rb
bugfinder: for details probably see jira OBS-215 which was not solved after the original PR 13090 was merged

PR #14083, opened 2023-03-29
Show a warning when a build is broken

danidoni:
When sending several request actions together, we display a warning in
the request actions dropdown to tell the user some of the builds are broken.


src/api/app/helpers/webui/request_helper.rb
```diff
@@ -174,4 +174,23 @@ def request_type_of_action(bs_request)
 
     bs_request.bs_request_actions.first.type.titleize
   end
+
+  def quick_build_results_status(buildresults)
+    return unless buildresults.reject { |b| b == 'succeeded' }.any?
```
ncounter, line 179: Wouldn't it be faster computation with
`return unless buildresults.any? { |b| b != 'succeeded' }` ?

src/api/app/views/webui/request/_actions_details.html.haml
```diff
@@ -10,6 +19,7 @@
         %button.btn.btn-outline-primary.btn-sm.dropdown-toggle.rounded-0{ 'aria-expanded' => 'false', 'data-bs-toggle' => 'dropdown',
                                                                           :type => 'button', 'data-boundary' => 'viewport' }
           Select Action
+          %i.fas{ class: quick_build_results_status(buildresults) }
```
ncounter, line 22: I wouldn't display it here in the dropdown header, it does not belong the the whole Request and this is only to select one action. I'd rather move it under the `Action details` part, as it belongs to that one. Something roughly like the following:
image

src/api/app/views/webui/request/_action_text.html.haml
```diff
@@ -39,3 +39,5 @@
         by
         = details['user']
         = render TimeComponent.new(time: details['time'].to_i)
+
+  %i.fas{ class: build_status_icon_from_action(action) }
```
ncounter, line 43: I think there is a missing header column for this, it should be added [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/views/webui/request/_actions_details.html.haml#L19)

src/api/app/helpers/webui/request_helper.rb
danidoni: Definitely!

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -174,4 +174,23 @@ def request_type_of_action(bs_request)
...
+  end
+
+  def quick_build_results_status(buildresults)
+    return unless request_with_broken_build(buildresults)
+
+    'fa-exclamation-triangle repository-state-broken'
+  end
+
+  def build_status_icon_from_action(action)
+    source_project = Project.find_by_name(action.source_project)
```
hennevogel, line 189: I'm not sure if it must but this doesn't take any [links](https://github.com/openSUSE/open-build-service/wiki/Links) into account...

hennevogel (changes requested):

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -174,4 +174,23 @@ def request_type_of_action(bs_request)
 
     bs_request.bs_request_actions.first.type.titleize
   end
+
+  def request_with_broken_build(buildresults)
+    buildresults.any? { |b| b != 'succeeded' && b != 'disabled' && b != 'excluded' && b != 'building' }
```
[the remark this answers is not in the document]
saraycp: There are a few methods that might be useful for this. This [unsuccessful_final_status?](https://github.com/openSUSE/open-build-service/blob/6daa0b5f3f17db4015e408054a016a4d69f884d0/src/api/app/models/buildresult.rb#L115) for example.

src/api/app/helpers/webui/request_helper.rb
danidoni: Let me think about that

danidoni: As far as I could test, linked projects work perfectly fine

src/api/app/helpers/webui/request_helper.rb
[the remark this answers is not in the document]
danidoni: > There are a few methods that might be useful for this. This [unsuccessful_final_status?](https://github.com/openSUSE/open-build-service/blob/6daa0b5f3f17db4015e408054a016a4d69f884d0/src/api/app/models/buildresult.rb#L115) for example.

Huh, I didn't see that message before. Let me take a look

src/api/app/helpers/webui/request_helper.rb
hennevogel: Which can only mean  you did not test hard enough...

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -182,4 +182,23 @@ def request_type_of_action(bs_request)
...
+
+  def build_status_attention_icon(buildresults)
+    return unless build_status_needs_attention?(buildresults)
+
+    'fa-exclamation-triangle repository-state-broken'
+  end
+
+  def build_status_icon_from_action(action)
+    source_project = Project.find_by_name(action.source_project)
+    source_package = Package.find_by_project_and_name(source_project.name, action.source_package)
```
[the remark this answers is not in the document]
danidoni: Thanks to @rubhanazeem latest changes in #14244, when there are no source projects, and therefore no build results, the tab does not get rendered and then this code is not called.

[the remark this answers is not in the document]
hennevogel: It's decided by the `BsRequestAction.source_....` which is a string. But if there is a string in there, the project might not exist because it's deleted for instance or because it's read protected etc.

src/api/app/helpers/webui/request_helper.rb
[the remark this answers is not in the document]
danidoni: I refactored the code to reuse `Buildresult#unsuccessful_final_status`. Please take a look.

src/api/app/helpers/webui/request_helper.rb
```diff
@@ -199,4 +200,18 @@ def next_prev_path(**opts)
       request_show_path(parameters)
     end
   end
+
+  def build_status_needs_attention?(buildresults)
+    buildresults.any?(&:unsuccessful_final_status?)
+  end
```
saraycp, line 206: Could this be moved to the BsRequestAction model?

It seems more natural for me to ask the action if any of its build results need attention... 
having the answer, you use the helpers to shape the output in the view.

src/api/app/views/webui/request/_request_header.html.haml
```diff
@@ -51,6 +59,11 @@
         %mark.text-dark.alert-warning.ps-2.pe-2.text-nowrap{ title: bs_request.accept_at }
           auto accepted
           = render TimeComponent.new(time: bs_request.accept_at)
+
+  - if build_status_needs_attention?(buildresults) && supported_actions.count > 1
+    %i.fas{ class: build_status_attention_icon(buildresults) }
+    Some action need attention
```
saraycp, line 65: If you accept my previous comment, this could be simplified. There is no need to pass the `buildresults` to the helpers.

src/api/app/helpers/webui/request_helper.rb
danidoni: Yeah, makes sense. Let me change it

src/api/app/views/webui/request/_request_header.html.haml
danidoni: You'll need to send the model anyway, but I see your point.

src/api/app/models/bs_request_action.rb
```diff
@@ -866,6 +866,16 @@ def event_parameters
     params
   end
 
+  def build_results
+    return unless source_project
+
+    source_project_object = Project.find_by_name(source_project)
+    source_package_object = Package.find_by_project_and_name(source_project_object.name, source_package)
```
hennevogel, line 873: Again, a `Project` with the name that is in `source_project` might note exist anymore or be not readable by `User.session` in which case this whole thing will barf.

hennevogel (changes requested):

src/api/app/models/bs_request_action.rb
danidoni: This is fixed, now

saraycp (changes requested):

danidoni:
We agreed offline to not go into this direction. Closing

PR #14087, opened 2023-03-30
Do not redirect, use render and reinitialize

ncounter:
Follow-up for https://github.com/openSUSE/open-build-service/pull/14080

### DIFF
--- a/src/api/.rubocop_todo.yml
+++ b/src/api/.rubocop_todo.yml
@@ -168,7 +168,6 @@ Metrics/ClassLength:
     - 'app/controllers/webui/search_controller.rb'
     - 'app/controllers/webui/staging/workflows_controller.rb'
     - 'app/controllers/webui/users_controller.rb'
-    - 'app/controllers/webui/users/tokens_controller.rb'
     - 'app/controllers/webui/webui_controller.rb'
     - 'app/lib/authenticator.rb'
     - 'app/lib/backend/connection.rb'
--- a/src/api/.rubocop_todo.yml
+++ b/src/api/.rubocop_todo.yml
@@ -168,6 +168,7 @@ Metrics/ClassLength:
     - 'app/controllers/webui/search_controller.rb'
     - 'app/controllers/webui/staging/workflows_controller.rb'
     - 'app/controllers/webui/users_controller.rb'
+    - 'app/controllers/webui/users/tokens_controller.rb'
     - 'app/controllers/webui/webui_controller.rb'
     - 'app/lib/authenticator.rb'
     - 'app/lib/backend/connection.rb'
--- a/src/api/app/controllers/webui/users/tokens_controller.rb
+++ b/src/api/app/controllers/webui/users/tokens_controller.rb
@@ -102,10 +102,12 @@ class Webui::Users::TokensController < Webui::WebuiController
     # Prevent setting a package for a workflow token
     return if @params[:type] == 'workflow'
 
+    @token = Token.token_type(@params[:type]).new(description: @params[:description])
+
     # Check if only project_name or only package_name are present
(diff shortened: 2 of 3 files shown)


src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/ClassLength
```
danidoni, line 1: I believe it's better to disable the cops with the [.rubocop_todo.yml](https://github.com/openSUSE/open-build-service/blob/9f99e2263a478c1d4a9a2d736f746ebdde61267c/.rubocop_todo.yml#L2)

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -105,19 +106,28 @@ def set_package
     # Check if only project_name or only package_name are present
     if @extra_params[:project_name].present? ^ @extra_params[:package_name].present?
       flash[:error] = 'When providing an optional package, both Project name and Package name must be provided.'
-      redirect_to new_token_path and return
+      @token = reinitialize_token(@params)
```
danidoni, line 109: You can move this `@token` assignment and the one below in just one line above the line 105

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -105,19 +106,28 @@ def set_package
...
     rescue Package::UnknownObjectError
       flash[:error] = 'When providing an optional package, the package must exist. ' \
                       "Package '#{elide(@extra_params[:project_name])}/#{elide(@extra_params[:package_name])}' does not exist."
-      redirect_to new_token_path
+      @token = reinitialize_token(@params)
+      render :new
     end
   end
+
+  def reinitialize_token(params)
```
hennevogel, line 128: I know what `Token.new` does, I have to look up what `reinitialize_token` does. I don't think it's worth turning this into a method...

hennevogel (changes requested):

ncounter:
Changes applied according to all the comments. Please could you have another look?

src/api/app/controllers/webui/users/tokens_controller.rb
```diff
@@ -102,22 +102,25 @@ def set_package
     # Prevent setting a package for a workflow token
     return if @params[:type] == 'workflow'
 
+    # Make sure token in initialized before running into initialization errors and return
```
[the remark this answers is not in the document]
ncounter: I rephrased a little bit. The point is making sense of the _why are we initializing the variable `@token` here?_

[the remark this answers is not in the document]
hennevogel: I don't think you clarify anything with this. Settings objects up before the controller action is standard practice. Usually even in a `before_action`....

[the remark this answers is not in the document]
ncounter: I see. I'll just drop the comment then.

src/api/.rubocop_todo.yml
```diff
@@ -168,6 +168,7 @@ Metrics/ClassLength:
     - 'app/controllers/webui/search_controller.rb'
     - 'app/controllers/webui/staging/workflows_controller.rb'
     - 'app/controllers/webui/users_controller.rb'
+    - 'app/controllers/webui/users/tokens_controller.rb'
```
hennevogel, line 171: This is needed because of one extra line?

ncounter: Apparently not. It was before with few more lines, but now it doesn't. Good catch. Dropped

hennevogel (approved):

PR #14088, opened 2023-03-30
[backend] bs_worker: make file writing fatal

adrianschroeter:
To catch such failures:

fetching sources, getsources: syswrite: No space left on device


mlschroe (changes requested):
Nope.

PR #14092, opened 2023-03-31
 Remove enforcing "vcr: true" for feature specs

eduardoj:
VCR was enforced for feature specs. This happened trying to make VCR the default for feature specs. See commit b652ae2.

Instead of fixing the code and making VCR the default for feature specs, remove enforcing of VCR for feature specs. This way, the `vcr:` flag is used like in the rest of the tests, and for those tests which don't need VCR there is no need to use `vcr: false`. We simply don't define any `vcr:` flag.

Also, adapt feature specs to use VCR only when needed, and remove VCR cassettes no loner needed.


hennevogel (approved):

hennevogel:
Did not merge yet so people can catch up on the new default...

danidoni (approved):

PR #14100, opened 2023-04-03
Use the comment components for rendering comments

hellcp-work:
This uses comment components for rendering comments in the changes tab, in order to avoid issues when rendering comments in the timeline and changes tab at the same time. It migrates the comment components to using delete modal component too.

Screenshot from 2023-04-03 09-31-48

### DIFF
--- a/src/api/app/assets/javascripts/webui/comment.js
+++ b/src/api/app/assets/javascripts/webui/comment.js
@@ -21,11 +21,11 @@ function validateForm(e) {
 
 $(document).ready(function(){
   // Disable submit button if textarea is empty and enable otherwise
-  $('.comments-list,.comment_new,.timeline').on('keyup', '.comment-field', function(e) {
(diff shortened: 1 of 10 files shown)


saraycp (changes requested):
These are a few cases that are not working fine:

- You edit a comment, but after clicking the Update button the change is saved but the UI doesn't change. We still see the textarea.
- Once you try to do the previous step (editing) then the Reply stops working: the Add Comment button keeps as disabled.

I think this GIF clarifies what I'm trying to explain:

inline_comments

saraycp (changes requested):
Now Edit and Reply work fine. However, if I add a comment and refresh the page, the comments are no longer displayed in the Changes tab.

saraycp:
> Now Edit and Reply work fine. However, if I add a comment and refresh the page, the comments are no longer displayed in the Changes tab.

The reason is that I tried with an action that is not the first one on the list. This is going to be fixed in: https://github.com/openSUSE/open-build-service/pull/14058

saraycp (approved):

PR #14107, opened 2023-04-04
Make sure the regex finds matches for diff_ref

hellcp-work:
Before if the diff_ref didn't match the regex, the page would not render, now the regex is tried first so we know if the string we get in diff_ref matches.

https://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/642bf86284bf4604fe227a8e

### DIFF
--- a/src/api/app/components/bs_request_activity_timeline_component.rb
+++ b/src/api/app/components/bs_request_activity_timeline_component.rb
@@ -23,9 +23,9 @@ class BsRequestActivityTimelineComponent < ApplicationComponent
   end
 
   def diff_for_ref(comment)
-    return unless comment.diff_ref
+    return unless (ref = comment.diff_ref&.match(/diff_([0-9]+)/))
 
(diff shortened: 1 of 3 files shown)


ncounter (approved):
LGTM

src/api/app/components/bs_request_activity_timeline_component.rb
```diff
@@ -23,9 +23,9 @@ def initialize(bs_request:)
   end
 
   def diff_for_ref(comment)
-    return unless comment.diff_ref
+    return unless (ref = comment.diff_ref&.match(/diff_([0-9]*)/))
```
eduardoj, line 26:
```suggestion
    return unless (ref = comment.diff_ref&.match(/diff_([0-9]+)/))
```
We can replace here the `*` by the `+`. We don't need to take in account the case of not having a digit.

eduardoj (approved):

PR #13736, opened 2023-01-19
Split out PackageController#binary into its own controller

hellcp-work:
This is a part of #13564, splitting out binary view into its own controller

PR #14112, opened 2023-04-05
Move url retrieval to binaries helper

hellcp-work:
This is a part of #13736, splitting this up into smaller pieces to make it easier to review


hennevogel:
Why?

hennevogel:
I think you have this backwards then. Because for I would for instance object to do something like `include Webui::Packages::BinariesHelper` in a refactored controller. As this helper includes `Webui::WebuiHelper` which includes everything else etc...

hennevogel:
> When this refactoring is done

You mean all of it not this PR right?

hennevogel:
But it still includes the helpers. I'm confused now...

PR #14116, opened 2023-04-05
Add missing database indexes

danidoni:


danidoni:
See #14150

danidoni:
This is so old I don't remember what was the issue...

PR #14119, opened 2023-04-06
Improve request state badge component

ncounter:
Follow-up of https://github.com/openSUSE/open-build-service/pull/14118 

Make use of the badge component everywhere is possible
Add a different `code-pull-request` icon next to the `bs_request` status text in the badge

### Review
after-component-beta

### Accepted
merge-icon

### Superseded
superseded-badge

### Watchlist

full-set

### DIFF
--- a/src/api/app/components/bs_request_state_badge_component.rb
+++ b/src/api/app/components/bs_request_state_badge_component.rb
@@ -9,7 +9,7 @@ class BsRequestStateBadgeComponent < ApplicationComponent
   def call
     content_tag(
       :span,
-      tag.i(class: 'fas fa-code-pull-request me-1').concat(@bs_request.state),
+      icon_state_tag.concat(@bs_request.state),
       class: ['badge', "bg-#{decode_state_color(@bs_request.state)}", @css_class]
     )
   end
@@ -18,14 +18,45 @@ class BsRequestStateBadgeComponent < ApplicationComponent
     case state
     when :review, :new
       'secondary'
-    when :declined, :revoke
+    when :declined
       'danger'
     when :superseded
       'warning'
     when :accepted
       'success'
+    when :revoked
+      'dismissed'
     else
       'dark'
+    end
+  end
+
+  def decode_state_icon(state)
+    case state
+    when :new
+      'code-branch'
+    when :review, :declined, :revoked
+      'code-pull-request'
+    when :superseded
+      'code-compare'
+    when :accepted
+      'code-merge'
+    else
+      'code-fork'
+    end
+  end
+
+  def icon_state_tag
+    if [:declined, :revoked].include?(@bs_request.state)
+      content_tag(
+        :span,
+        tag.i(class: "fas fa-#{decode_state_icon(@bs_request.state)}").concat(
+          tag.i(class: 'fas fa-slash fa-stack-1x fa-stack-slash top-icon')
+        ),
+        class: 'position-relative me-1'
+      )
+    else
+      tag.i(class: "fas fa-#{decode_state_icon(@bs_request.state)} me-1")
     end
   end
 end
--- a/src/api/app/components/bs_request_state_badge_component.rb
+++ b/src/api/app/components/bs_request_state_badge_component.rb
@@ -10,7 +10,22 @@ class BsRequestStateBadgeComponent < ApplicationComponent
     content_tag(
       :span,
       tag.i(class: 'fas fa-code-pull-request me-1').concat(@bs_request.state),
-      class: ['badge', "bg-#{helpers.request_badge_color(@bs_request.state)}", @css_class]
+      class: ['badge', "bg-#{decode_state_color(@bs_request.state)}", @css_class]
     )
+  end
+
+  def decode_state_color(state)
+    case state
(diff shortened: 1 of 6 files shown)


saraycp (changes requested):

src/api/app/components/bs_request_state_badge_component.rb
```diff
@@ -7,7 +7,40 @@ def initialize(bs_request:, css_class: nil)
...
+    else
+      'dark'
+    end
+  end
+
+  def decode_state_icon(state)
+    case state
+    when :new
+      'code-branch'
+    when :review, :declined, :revoke
```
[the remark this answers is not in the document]
ncounter: Done, thanks for the feedback. I agree

src/api/app/components/bs_request_state_badge_component.rb
```diff
@@ -7,7 +7,40 @@ def initialize(bs_request:, css_class: nil)
...
+      tag.i(class: "fas fa-#{decode_state_icon(@bs_request.state)} me-1").concat(@bs_request.state),
+      class: ['badge', "bg-#{decode_state_color(@bs_request.state)}", @css_class]
+    )
+  end
+
+  def decode_state_color(state)
+    case state
+    when :review, :new
+      'secondary'
+    when :declined, :revoke
```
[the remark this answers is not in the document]
ncounter: I like the option to distinguish the `revoked` one actually, because it is not a real `danger` situation.
Maybe the dark one is too dark, I tried with something lighter and i like it so far. It looks like the following, what do you think?

Screenshot from 2023-04-11 12-17-39
Screenshot from 2023-04-11 12-17-20

src/api/app/components/bs_request_state_badge_component.rb
```diff
@@ -9,8 +9,40 @@ def initialize(bs_request:, css_class: nil)
...
+    else
+      'dark'
+    end
+  end
+
+  def decode_state_icon(state)
+    case state
+    when :new
+      'code-branch'
+    when :review, :declined, :revoked
```
krauselukas, line 38: @ncounter for a decline review we show the `fa-times` icon in other places 
image

the `code-pull-request` indicates to me it is still open somehow. Maybe we can use the cross as well?

ncounter: Yeah, @krauselukas I agree it is not the best fit the `code-pull-request`. Problem is, we don't have the `code-pull-request-closed` icon in our font-awesome set because it is not in the _free_ one. So either we stay with some `pull-request`-ish icon, or we use the `fa-times` only. I'd like to avoid adding a custom icon in a different set.
I'd be fine with the `fa-times` only, even if it is no longer clear it is about a code-request by itself. What do you think? Is there any other [font-awesome](https://fontawesome.com/search?o=r&m=free) icon we could use for this special case?

DavidKang: @ncounter, I was thinking the same as @krauselukas. And not sure if adding the icon + text in this case make sense or is better to only have the icon or just the text. It would make sense to use both if you have the intention to use them without the "text" in some places.

I will use Github as example:
Screenshot from 2023-04-11 15-02-09
Screenshot from 2023-04-11 15-02-25

But is my opinion :smile_cat:. BTW, nice work :+1:

src/api/app/components/bs_request_state_badge_component.rb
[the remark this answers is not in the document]
saraycp: I like the gray-ish one :+1:

src/api/app/components/bs_request_state_badge_component.rb
krauselukas: @ncounter  IMO the `fa-times` would be fine, we use it for a declining review on a request, which is what the declined label represent in the end on the request. Someone declined in the review process. Through a quick search on the font awesome library I couldn't spot anything better...

saraycp: TBH, I was not a big fan of the `fa-times` for this. It recalls a delete action and it's also used for failed build results... so I wouldn't use it for request states too.

Taking into account we don't have better free icons, I'd go with your last proposal, yes. So `code-pull-request` + `slash` for `declined` and `revoked` requests.

danidoni (approved):

src/api/app/components/bs_request_state_badge_component.rb
DavidKang: Agree with @saraycp

ncounter: Great, thank you for your feedback.
I implemented the combination of two icons for both `declined` and `revoked`. 

In the PR description I attached the full set of different badge states.

Feel free to give a final review then, Thanks

saraycp (approved):

ncounter:
`codecov/project` is stale, everything else is passing https://app.circleci.com/pipelines/github/openSUSE/open-build-service/11431/workflows/47a86c0d-25cc-406b-a754-cdf52530f102 . Merging

PR #14120, opened 2023-04-06
BugFix: collapsible text for request beta

ncounter:
In the beta version of a `BsRequest` page the `collapsible-text` is buggy: the culprit is the fact that content is visible only after tabs are loaded and the one selected is rendered. The logic of the `collapsible-text` rely on the `height` of the element where the possible long text is contained so, if the tab is not yet rendered the `height` is computed to `0` and this makes the logic to fail adding the `show-more`/`show-less` button.

This PR fixes first the bug by making sure the `setCollapsible();` is called after the tab content is visible.
Later on the PR refactors the whole `setCollapsible()` logic because it still has corner cases (such as the `click` event that is added more than once to all the elements every time the `setCollapsible()` is called, for instance).

### DIFF
--- a/src/api/app/assets/javascripts/webui/application.js
+++ b/src/api/app/assets/javascripts/webui/application.js
@@ -44,7 +44,7 @@
 //= require webui/staging_workflow.js
 //= require webui/project.js
 //= require webui/project_monitor.js
-//= require webui/collapsible_text
+//= require webui/collapsible_text.js
 //= require webui/patchinfo.js
 //= require webui/image_templates.js
 //= require webui/kiwi_editor.js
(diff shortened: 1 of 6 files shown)


saraycp:
LGTM. Can you please squash the last five commits?

ncounter:
> LGTM. Can you please squash the last five commits?

Thanks. And done!

danidoni (approved):

ncounter:
`codecov/project` is stale, all the checks [are passing](https://app.circleci.com/pipelines/github/openSUSE/open-build-service/11383/workflows/f3627302-f4b3-4645-bae5-0f457d632c06/jobs/123713). Merging

PR #14130, opened 2023-04-11
Turn request tabs into controller actions

rubhanazeem:


src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -25,6 +25,7 @@
 
 .card
   .card-body.p-0
+<<<<<<< HEAD
```
krauselukas: @rubhanazeem seems some left over line from rebasing

krauselukas: I see, present in one commit and removed in another one. Maybe you can remove it in the commit where it got overlooked. There is also a `binding.pry` added in one commit, and removed in another one.

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -521,4 +495,61 @@ def staging_status(request, target_project)
       target_project_staging_url: staging_workflow_path(request.target_project_name)
     }
   end
+
+  def prepare_request_data
```
krauselukas, line 499: @rubhanazeem Do you plan to break this method up? So that only the data that is required for the individual views is prepared

rubhanazeem: There is a plan to break this method into smaller methods depending on the active tab. Will do it in a following PR

src/api/app/views/webui/request/beta_show.html.haml
rubhanazeem: I'll split the PR into smaller chunks.

rubhanazeem:
Closed because I'm moving the implementation into smaller PR's

PR #14135, opened 2023-04-12
Regenerate the database_consistency todo file

hellcp-work:
The checks involving database_consistency are failing now, we need to update this file to avoid it

### DIFF
--- a/.circleci/conditional_config.yml
+++ b/.circleci/conditional_config.yml
@@ -162,6 +162,7 @@ jobs:
   linters:
     docker:
       - <<: *frontend_base
+      - <<: *mariadb
     steps:
       - attach_workspace:
          at: .
@@ -203,6 +204,12 @@ jobs:
                 command: |
                   cd src/api
                   bundle exec rake dev:lint:apidocs
+      - run:
+          name: Run database consistency linter
+          working_directory: src/api
+          command: |
+            bundle exec rake db:setup
+            bundle exec database_consistency -c .database_consistency.todo.yml
 
   rspec:
     parallelism: 3
@@ -302,7 +309,6 @@ jobs:
             bundle exec rake db:migrate
             bundle exec rake db:seed
             bundle exec rake data:migrate
-            bundle exec database_consistency -c .database_consistency.todo.yml
 
   passenger_tests:
     docker:


ncounter (approved):

.circleci/config.yml
```diff
@@ -19,6 +19,7 @@ workflows:
             src/backend/.* run-backend-tests true
             src/api/app/(views|components)/.* run-haml-linter true
             src/api/db/.* run-migrations-tests true
+            src/api/app/models/.* run-migrations-tests true
```
eduardoj, line 22:
```suggestion
```
We can remove these two lines. `run-migrations-tests` boolean got removed.

.circleci/conditional_config.yml
```diff
@@ -512,6 +499,9 @@ workflows:
     jobs:
       - pipeline_settings
       - checkout_code
+      - migrations_tests:
```
hennevogel, line 502: That's a rather big hammer, running all the migrations tests.

hennevogel, line 502: That's a rather big hammer, running all the migration tests. How about just moving out `bundle exec database_consistency...`

.circleci/config.yml
hennevogel, line 22: You can drop this now...

hennevogel (approved):

.circleci/conditional_config.yml
```diff
@@ -203,6 +204,12 @@ jobs:
                 command: |
                   cd src/api
                   bundle exec rake dev:lint:apidocs
+      - run:
+          name: Run migration tests
```
hennevogel, line 208:
```suggestion
          name: Run database consistency linter
```

.circleci/conditional_config.yml
```diff
@@ -203,6 +204,12 @@ jobs:
...
+          command: |
+            cd src/api
```
eduardoj, line 210: We could use the `working_directory` key, instead of adding a `cd ...` line in the command. See https://circleci.com/docs/configuration-reference/#run

eduardoj (changes requested):
Don't forget to remove this line: https://github.com/openSUSE/open-build-service/pull/14135/files#diff-8685bb66c118bbbacdf2bde0a63996d186c0b336baff04ceb50e70b5cf3fb23eR312

.circleci/conditional_config.yml
eduardoj: You applied it.

eduardoj (approved):

PR #14144, opened 2023-04-12
Introduce templates for new packages

hellcp-work:
Introduces a simple template support to new project creation. Currently only supports one type of a package, SUSE style RPM, but it can be easily expanded to support everything that may be needed.

To verify:
1. Create a new project for templates
1. Set the OBS:PackageTemplates attribute on the project
1. Create a package in the template project
1. Create a file with name like: `{{ package.name }}.extension.liquid` and contents like `{{ user.name }}` (other options are in `src/api/app/services/package_service/templater/{project,package,user}_drop.rb files`)
1. Go to any other project
1. Click a create new package button
1. Write metadata about the new package
1. Select the template you created in the template project
1. Observe the package be initialized with a set of files

TODO:

- [x] Create test data
- [x] Create tests
- [x] Write documentation for the feature (being worked on in a follow up PR)

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -8,20 +8,22 @@ class Webui::PackageController < Webui::WebuiController
   # rubocop:disable Rails/LexicallyScopedActionFilter
   before_action :require_login, except: %i[show index branch_diff_info
                                            users requests statistics revisions view_file
-                                           devel_project buildresult rpmlint_result rpmlint_log rpmlint_summary files]
+                                           devel_project buildresult rpmlint_result rpmlint_log
+                                           rpmlint_summary files template_data]
 
   # The methods save_person, save_group and remove_role are defined in Webui::ManageRelationships
   before_action :set_project, only: %i[show edit update index users requests statistics revisions
                                        new branch_diff_info rdiff create remove
                                        save_person save_group remove_role view_file
-                                       buildresult rpmlint_result rpmlint_log rpmlint_summary files]
+                                       buildresult rpmlint_result rpmlint_log rpmlint_summary files template_data]
 
   before_action :check_scmsync, only: %i[requests revisions statistics users]
 
   before_action :set_package, only: %i[edit update show requests statistics revisions
                                        branch_diff_info rdiff remove
                                        save_person save_group remove_role view_file
-                                       buildresult rpmlint_result rpmlint_log rpmlint_summary files users]
+                                       buildresult rpmlint_result rpmlint_log rpmlint_summary files
+                                       users template_data]
   # rubocop:enable Rails/LexicallyScopedActionFilter
   before_action :lints_list, only: %i[rpmlint_summary]
 
@@ -348,6 +350,21 @@ class Webui::PackageController < Webui::WebuiController
 
   def files; end
   def view_file; end
+
+  def template_data
+    data = []
+    begin
+      xml = Nokogiri::XML(@package.source_file('_template'))
+      xml.css('template').children.each do |child|
+        next if child.name == 'text'
+
+        data << { name: child.name, description: child.content }
+      end
+    rescue StandardError
+      logger.debug("No template information provided for #{@project.name}/#{@package.name}")
+    end
+    render json: data.to_json, layout: false
+  end
 
   private
 
--- a/src/api/app/services/package_service/templater.rb
+++ b/src/api/app/services/package_service/templater.rb
@@ -20,6 +20,8 @@ module PackageService
 
     def files
       @template.dir_hash.elements('entry').map do |file|
+        next if file['name'] == '_template'
+
         template_contents = @template.source_file(file['name'])
         return uploaded_file(file['name'], template_contents) unless file['name'].end_with?('.liquid')
 
--- a/src/api/.jshintrc
+++ b/src/api/.jshintrc
@@ -3,7 +3,7 @@
   "eqeqeq"    : true, // true: Require triple equals (===) for comparison
   "undef"     : true, // true: Require all non-global variables to be declared (prevents global leaks)
   "unused"    : true, // Unused variables: true : all variables, last function parameter
-  "esversion" : 6,
+  "esversion" : 8,
 
   "globals" : {
     "$" :        false,
@@ -18,6 +18,8 @@
     "isOutdatedUrl" : false,
     "mainProjects"  : false,
     "exclProjects"  : false,
-    "sessionStorage": true
+    "sessionStorage": true,
+    "fetch": false,
+    "console": false
   }
 }
--- a/src/api/app/assets/javascripts/webui/packages.js
+++ b/src/api/app/assets/javascripts/webui/packages.js
@@ -1,3 +1,5 @@
+/* exported setTemplateData */
+
 /* global setupDropdownFilters */
 
 // Remove this after PackageController#rdiff moves to DiffListComponent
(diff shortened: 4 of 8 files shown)


src/api/app/models/project.rb
```diff
@@ -197,6 +197,39 @@ def remote_image_template_from_xml(remote_project, image_template_project)
       project
     end
 
+    def package_templates
+      Project.with_package_templates + remote_package_templates
+    end
+
+    def remote_package_templates
+      result = []
+      Project.remote.each do |project|
```
danidoni, line 206: We're inside `Project` right? You don't need the class name...

hellcp-work: Removed

src/api/app/controllers/webui/package_controller.rb
```diff
@@ -450,4 +452,11 @@ def get_diff(project, package, options = {})
     end
     true
   end
+
+  def set_template
+    return unless params['template']
+
+    template_project, template_package = params['template'].split('/')
```
danidoni, line 459: What happens if `template` does not have a '/' inside?

src/api/app/models/project.rb
```diff
@@ -105,6 +105,7 @@ def staging_history
   scope :related_to_user, ->(user_id) { joins(:relationships).where(relationships: { user_id: user_id }) }
   scope :for_group, ->(group_id) { joins(:relationships).where(relationships: { group_id: group_id, role_id: Role.hashed['maintainer'] }) }
   scope :related_to_group, ->(group_id) { joins(:relationships).where(relationships: { group_id: group_id }) }
+  scope :with_package_templates, -> { joins(:attribs).where(attribs: { attrib_type_id: AttribType.joins(:attrib_namespace).where(attrib_namespace: { name: 'OBS' }, attrib_types: { name: 'PackageTemplates' }) }) }
```
danidoni, line 108: I think you can do the same without subqueries by adding `:attrib_types` into the join, right?

src/api/spec/features/webui/package_templates_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'browser_helper'
+
+RSpec.describe 'Package Templates', :vcr do
+  let(:user) { create(:confirmed_user, :with_home, login: 'tom') }
+  let(:admin) { create(:admin_user) }
+  let(:project) { user.home_project }
+  let(:template_project) do
+    login admin
```
danidoni, line 8: Why do you need logging in?

hellcp-work:
> You should do that in `lib/tasks/dev/package-templates.rake` and `dev:test_data`

It's done in `lib/tasks/dev/templates.rake` and `dev:test_data`

src/api/spec/features/webui/package_templates_spec.rb
hellcp-work: Required to create attributes

src/api/app/models/project.rb
hellcp-work: Not sure, let me check

hennevogel:
I don't get it, is replacing Package/Project/User data the "feature" of those templates? Do we need to code and commit `Liquid::Drop` for each new object we want to fetch data from? What about data not from some object but from the form?

hellcp-work:
> I don't get it, is replacing Package/Project/User data the "feature" of those templates? Do we need to code and commit `Liquid::Drop` for each new object we want to fetch data from? What about data not from some object but from the form?

Yes, all data accessible from the template is specified by those drops. I wanted to avoid presenting data that we don't want to present, like User#password. The things in a form will be available via PackageDrop, since that's the package the user created with that metadata. Templates are unfortunately code, and keeping a minimal amount of access to the database is a good idea, since this feature if not done carefully, is an easy attack vector.

src/api/app/controllers/webui/package_controller.rb
hellcp-work: Resolved

src/api/app/models/project.rb
danidoni: You can join more than one table

hellcp-work: I tried doing the following, but it seems rails doesn't like this:
```ruby
  scope :with_package_templates, -> { joins(attribs: [ attrib_type: [:attrib_namespace]]).where(attribs: { attrib_type: { name: 'PackageTemplates', attrib_namespace: { name: 'OBS' }}}) }
  ```
erroring with:
```
Mysql2::Error: Unknown column 'attrib_type.name' in 'where clause'
```

So it does join, but then `where` seems to fail

Comment (danidoni) [reply src/api/app/models/project.rb:115]: > Mysql2::Error: Unknown column 'attrib_type.name' in 'where clause'
> ```
> 
> So it does join, but then `where` seems to fail

I think you need to use the table names, not the association names, try with `attrib_types`...

Comment (hellcp-work) [reply src/api/app/models/project.rb]: I figured it out now

File: src/api/app/controllers/public_controller.rb [hunk 5]
```diff
@@ -204,6 +204,10 @@ def image_templates
     render 'webui/image_templates/index'
   end
 
+  def package_templates
+    @projects = Project.package_templates
```
danidoni, line 208: It's a bit confusing that asking for package templates fills projects...

hellcp-work: This is the same issue as `image_templates` method

hellcp-work: The name
https://github.com/openSUSE/open-build-service/blob/fc378c6dafe23dbbb1ffeebb332b03ec386d0773/src/api/app/controllers/public_controller.rb#L202-L205

danidoni: Alright

src/api/Gemfile.lock
```diff
@@ -273,6 +273,9 @@ GEM
       childprocess (~> 5.0)
       logger (~> 1.6)
     lint_roller (1.1.0)
+    liquid (5.11.0)
```
krauselukas, line 276: @hellcp-work the gem file in "vendor/cache" and the version set in the Gemfile don't match, I guess maybe happened during rebasing. also the version of the strscan gem shouldn't be updated since its part of the ruby std gems (we need to keep the same version that is used by the ruby interpreter in our bundle)

src/api/db/seeds.rb
```diff
@@ -177,6 +177,9 @@
 at = ans.attrib_types.where(name: 'LimitReleaseSourceProject').first_or_create
 at.attrib_type_modifiable_bies.where(role_id: maintainer_role.id).first_or_create
 
+at = ans.attrib_types.where(name: 'PackageTemplates').first_or_create
```
krauselukas:
```suggestion
at = ans.attrib_types.where(name: 'PackageTemplates').first_or_create(value_count: 0)
```

krauselukas: @hellcp-work This attribute type will never have a value right? If you set the "value_count" to 0 it wont show the option to add a value in the webui etc.

src/api/db/data/20251127095605_add_package_templates_to_attrib_types.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class AddPackageTemplatesToAttribTypes < ActiveRecord::Migration[7.0]
+  def up
```
krauselukas, line 4: same as with the seeds, would add "value_count: 0" since the attrib doesnt use values

krauselukas (changes requested):

src/api/db/data/20251127095605_add_package_templates_to_attrib_types.rb
hellcp-work: Done

src/api/db/seeds.rb
hellcp-work: Done

src/api/Gemfile.lock
hellcp-work: Done

krauselukas: @hellcp-work there is still the update of strscan, this needs to be excluded. Otherwise we will have a problem in production because of a version mismatch between the bundle and the strscan std gem from ruby 3.4

src/api/app/services/package_service/templater.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def render_liquid_string(string)
+      template = Liquid::Template.parse(string)
+      template.render(template_options)
+    end
+
+    def template_options
+      { 'project' => PackageService::Templater::ProjectDrop.new(@package.project),
+        'package' => PackageService::Templater::PackageDrop.new(@package),
+        'user' => PackageService::Templater::UserDrop.new(User.session!) }
```
krauselukas: @hellcp-work I would suggest to not leak the user session into the service class, maybe you can pass the user down from the controller

hellcp-work: Done

PR #14147, opened 2023-04-12
Use instantiated data from previews

saraycp:


saraycp:
Many of the components need refactoring to not retrieve data from the database, it isn't just a matter of refactoring the preview. I close this PR and will open one per refactored component.

PR #14153, opened 2023-04-13
Refactoring card component

ncounter:
This PR fixes the `CardComponent` in order to align to same behavior for all the occurrences:
- cards have the same width for the same instance type
- cards on the same row have the same height
- make use of bootstrap footer to align action buttons to the bottom of the card, but keep the style transparent
- drop HTML card wrappers like `.repository` or `.group-user`, add the classes to the `.card` element instead

## Repositories
**Before**
repositories-before

**After**
repositories-after

## Tokens
**Before**
tokens-before

**After**
tokens-after

## Users
**Before**
users-before

**After**
users-after


src/api/app/components/card_component.rb
```diff
@@ -6,9 +6,10 @@ class CardComponent < ApplicationComponent
   renders_one :delete_button
   renders_many :actions
 
-  def initialize(extra_classes_for_body: [])
+  def initialize(extra_classes_for_card: [], extra_classes_for_body: [])
```
hennevogel, line 9: 💭 I don't dig this change. Passing around a string into a component every time to render it in the same place. Why?

ncounter: Former reason why I need it is that there are some `js` and `specs` relying on those classes (`repository-card` and `group-user`) so I cannot simply drop those classes.
Easiest solution would be to wrap every card by a `div` with the needed class, but that is something I want to avoid, and it would imply some `css` change anyway.
As we already have a typed wrapper container (`#tokens`, `#repositories`, `#group-users`) for all the groups of cards, ideally the best option would be to refactor `js` and `specs` in order to no longer rely on those classes. At this point I am not sure if it is in the scope of this PR or it is more a refactoring for later.

ncounter: I took the chance to refactor the code and drop the dependencies to those custom additional classes. Now there are no longer strings passed around: the component and the usages are is even more standardized and unified.

Thanks for raising it up.

hennevogel (approved):

PR #14156, opened 2023-04-13
Rearrange CircleCI linters steps

eduardoj:
RuboCop linter tests take longer than other linters.

Move them to the last position of the sequence of linters. This way, linters fail sooner because the linters that take less time are run first, and therefore, CI steps fail sooner on average when we encounter a linter offense and we use less test cycles.


hennevogel (approved):

hennevogel:
Codecov hangs. Codeclimate is fine though. See the checks. Merging!

PR #14157, opened 2023-04-13
Use the `working_directory` key instead of `cd` commands

eduardoj:
Improve the legibility of the `run` sections of the CircleCI configuration.


hennevogel (approved):

eduardoj:
Codecov hangs. Merging.

PR #14158, opened 2023-04-13
Move bugzilla_args into its own function

hellcp-work:
We can't set args in the proxy because it's overriding the args we provide later, so instead we have a common set of args that gets set to the appropriate user and api_key values


hennevogel (changes requested):

bmwiedemann (approved):

hennevogel (approved):

hennevogel:
Codecov hangs. Coverage is fine though. See the checks. Merging!

PR #14159, opened 2023-04-13
Split github action workflows into two separate workflows

krauselukas:
In order to allow the creation of comments on PR's that are
coming from forks, we would need to use the `:pull_request_target`
event over the `:pull_request` event. Since the `:pull_request_target`
event comes with more permissions on the target repo, its not recommended to
checkout the code coming from an untrusted fork.
Therefore the gh action workflow is now split into two separate
workflows.
One workflow keeps working on the `:pull_request` event and checksout
the code to perform the diff operation in order to receive gem
name, version etc. Afterwards the informations are saved as artifacts.
The second workflow listens for the `workflow_run` event and gets
triggered after the first one finished successfully, then downloads
the artifacts and finally creates the comment (the `:workflow_run`
event has permission to comment on the PR).

For reference https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

### DIFF
--- a/.github/workflows/comment_diffend_io_links_to_pr.yml
+++ b/.github/workflows/comment_diffend_io_links_to_pr.yml
@@ -1,0 +1,24 @@
+name: Add Comment with Diffend.io Links to PR
+on:
+  workflow_run:
+    workflows: ['Create Diffend.io Links and Comment Text']
+    types:
+      - completed
+permissions:
+  contents: read
+jobs:
+  comment_diffend_io_links_to_pr:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Download artifacts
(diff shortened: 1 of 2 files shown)


.github/workflows/comment_diffend_io_links_to_pr.yml
```diff
@@ -0,0 +1,24 @@
+name: Add Comment with Diffend.io binary diff links to PR
+on:
+  workflow_run:
+    workflows: ['Create Diffend.io gem binary diff links and comment text']
```
hennevogel, line 4:
```suggestion
    workflows: ['Create Diffend.io Links and Comment Text']
```

hennevogel (approved):

hennevogel:
Codecov hangs. Coverage is fine though. See the checks. Merging!

PR #14160, opened 2023-04-13
[api] fix bugzilla fetching

adrianschroeter:
overwriting args would ommit ids argument so the xmlrpc call failed


hennevogel:
Done differently in #14158

PR #5401, opened 2018-07-18
[frontend] Report review validation details

coolo:
'Reviews invalid' is little helpful, 'User not found' makes
it more obvious you had a typo

Fixes #5152

PR #14161, opened 2023-04-13
Return a "Not found" error instead of "Invalid review"

### LINKED ISSUE #14148: Wrong "Project can't be blank" error for add_review
When the given project does not exist (e.g. because of a typo), a wrong error message is returned:

```
> osc --debug review add -P X11:Xorg 1078779
makeurl: https://api.opensuse.org ['request', '1078779'] {'cmd': 'addreview', 'by_project': 'X11:Xorg'}
HTTPHeaderDict(...)
b'<status code="invalid_review">\n  <summary>Review invalid: Project can\'t be blank</summary>\n</status>\n'
Review invalid: Project can't be blank
```

It needs to state that the project does not exist.

eduardoj:
Fixes #14148.

Related to #5401.


src/api/app/models/bs_request.rb
```diff
@@ -736,6 +736,10 @@ def create_new_review(opts)
     )
     return newreview if newreview.valid?
 
+    if (error_text = newreview.errors.select { |error| error.attribute.in?([:user, :group, :project, :package]) }).any?
```
danidoni, line 739: This logic should go inside `Review` class, as `BsRequest` [is asking for information](https://thoughtbot.com/blog/tell-dont-ask) that is perfectly known by `Review`

src/api/app/models/bs_request.rb
```diff
@@ -736,6 +736,10 @@ def create_new_review(opts)
...
+      raise NotFoundError, error_text.map { |error| "#{error.attribute.capitalize} not found" }.to_sentence
```
danidoni, line 740: Same here

src/api/app/models/bs_request.rb
eduardoj: I created a method in the Review model and moved the login there. Please take a look again.

src/api/app/models/review.rb
```diff
@@ -302,6 +302,12 @@ def staging_project?
     for_project? && !project&.staging_workflow_id.nil?
   end
 
+  def check_reviewer!
+    return unless (selected_errors = errors.select { |error| error.attribute.in?([:user, :group, :project, :package]) }).any?
+
+    raise ::NotFoundError, selected_errors.map { |error| "#{error.attribute.capitalize} not found" }.to_sentence
```
bmwiedemann, line 308: Do we often use this style?
Normally, I would expect `if errors => raise` with the non-error case as fallthrough.

eduardoj: Changing the code to:
```ruby
  def check_reviewer!
    if (selected_errors = errors.select { |error| error.attribute.in?([:user, :group, :project, :package]) }).any?
      raise ::NotFoundError, selected_errors.map { |error| "#{error.attribute.capitalize} not found" }.to_sentence
    end
  end
```

... makes RuboCop complain:
```ruby
app/models/review.rb:306:5: C: [Correctable] Style/GuardClause: Use a guard clause (return unless (selected_errors = errors.select { |error| error.attribute.in?([:user, :group, :project, :package]) }).any?) instead of wrapping the code inside a conditional expression.
    if (selected_errors = errors.select { |error| error.attribute.in?([:user, :group, :project, :package]) }).any?
    ^^
```

Next time you could enter `bash` in the frontend container of the development environment (with `docker-compose run --rm --service-ports frontend bash`) and run `bin/rubocop --ignore_parent_exclusion app/models/review.rb`.

For reference see: https://github.com/openSUSE/open-build-service/wiki/Rubocop#running-rubocop-in-your-own

bmwiedemann: looking at https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause , reformulating it as
`raise if errors`
should also be accepted.

eduardoj: Changed. Please take a look again.

bmwiedemann (approved):

PR #14166, opened 2023-04-14
Simplify NotificationExcerptComponent

saraycp:
So the object is passed to the component and not retrieved inside. That makes it possible to not persist any object to use its preview.


danidoni (approved):

saraycp:
Rebase on master...

PR #14179, opened 2023-04-17
Add avatar and link component

saraycp:
Uses AvatarComponent to display an avatar next to a link to the user or group.

Screenshot 2023-04-18 at 00-48-27 Screenshot

**Reviewing Tips**

Use the view component previews to see how it looks.

### DIFF
--- a/src/api/app/components/avatar_and_link_component.html.haml
+++ b/src/api/app/components/avatar_and_link_component.html.haml
@@ -1,0 +1,2 @@
+= render(AvatarComponent.new(name: avatar_object.name, email: avatar_object.email, shape: shape, size: size, custom_css: avatar_css))
+= link_to(link_text, url)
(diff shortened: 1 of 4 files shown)


ncounter:
@saraycp if I get the aim of this PR correctly, you might want to check out the [user_avatar_component](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/components/user_avatar_component.rb) first and see if there is something to reuse or even better to create one and only one component to render an `icon/avatar + text/link` for all the instances.

saraycp:
> @saraycp if I get the aim of this PR correctly, you might want to check out the [user_avatar_component](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/components/user_avatar_component.rb) first and see if there is something to reuse or even better to create one and only one component to render an `icon/avatar + text/link` for all the instances.

Yes, I think we can get rid of the user_avatar_component in favor of this one which has a more generic name (should work for groups and users) and make use of the AvatarComponent. I'll do that in another PR.

saraycp:
I'm not sure if this is still helpful, I'm going to move it to Draft and will have a look first.

PR #14182, opened 2023-04-18
Improve rspec output in CI/dev-env

hennevogel:
In CI we are interested in way more info than dots so we can see which example is currently running.

In the dev-env we are interested in slightly more info than dots. Use Fuubar which is an instafailing formatter that uses a progress bar.

## CI Before
Screenshot from 2023-04-18 13-56-52

## CI After
Screenshot from 2023-04-18 13-56-31

## Dev-Env Before

......

## Dev-Env After

Fuubar


danidoni (approved):

danidoni:
The diffend step looks like it crashed

danidoni:
> Looks like the diffend step crashed, re-triggering

All good, merging

PR #14183, opened 2023-04-18
Darix/client cert support

darix:
This is not meant as a final implementation.  all those TODOs need solving. This is just a rough untested sketch to see how one might be able to do this.

the basic idea is to talk to the back end via TLS and use client certificate for authentication.

### DIFF
--- a/src/api/app/lib/backend/connection.rb
+++ b/src/api/app/lib/backend/connection.rb
@@ -28,6 +28,10 @@ module Backend
       unless @@ssl_options
         ## TODO: check if this really works. we should cache this in a class variable
         ##       so we do not have to do this over and over
+        ##       if we do caching here we also need to clarify
+        ##       how we handle certificate rotation
+        ##       e.g. signal handler to avoid having to restart the whole rails app
+        ##       but still reload the certs
 
         # the old default value for ssl_options
         @@ssl_options ||= { use_ssl: use_ssl, verify_mode: verify_mode }
--- a/src/api/app/lib/backend/connection.rb
+++ b/src/api/app/lib/backend/connection.rb
@@ -32,6 +32,8 @@ module Backend
         ##       how we handle certificate rotation
         ##       e.g. signal handler to avoid having to restart the whole rails app
         ##       but still reload the certs
+        ##       of course we could cache the mtime of the files and only the work
+        ##       if the file changed.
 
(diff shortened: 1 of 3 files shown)


hennevogel:
> the basic idea is to talk to the back end via TLS and use client certificate for authentication.

JFYI: The reason to go the client certificate route and not SSL + Auth is because socket passing is not implemented for some parts of the backend.

hennevogel:
Just some general architecturla thoughts: 

- I do like the simplicity on the frontend and how low level this is ❤️ 
- I don't like the reliance on just another tool (ha-proxy on front of the backend) to be able to do this. Especially for releases.
- I don't like the reliance of specially crafted certificates which will mean SSL documentation (probably distro specific and stuff)

src/api/app/lib/backend/connection.rb
```diff
@@ -22,13 +24,52 @@ class Connection
       end
     end
 
+    def self.ssl_options()
+      unless @@ssl_options
```
hennevogel, line 28: It'll probably be better to do this in an initializer once and store this in a [class accessor](https://api.rubyonrails.org/classes/Module.html#method-i-mattr_accessor). so this is per rails process (which we have many of) instead of per instantiated class (we do this often just for one call).

src/api/app/lib/backend/connection.rb
```diff
@@ -22,13 +24,52 @@ class Connection
...
+        @@ssl_options ||= { use_ssl: use_ssl, verify_mode: verify_mode }
+
+        # Handle client certificates
+        if CONFIG['source_host_client_certificate']
+
+          cert_data = File.read(CONFIG['source_host_client_certificate'])
+          ssl_options[:cert] = OpenSSL::X509::Certificate.new(cert_data)
+
+          # if you concatenate key and cert into one file you do not need to pass in source_client_key
+          # and we can reuse the cert_data here.
```
hennevogel, line 48: Or we can just settle and document for one of those options...

darix:
1. we want everything to use SSL for talking to other parts.
2. some parts of the backend can not easily implement this natively due to the use of socket passing. for those parts we will move the TLS part for now to haproxy. all other parts implement it natively.
3. this does not change the fact that the rails part needs to talk with TLS and client cert to the other bits and pieces.

PR #14185, opened 2023-04-18
Fix wrong version of migration

### LINKED ISSUE #14184: Migration with wrong filename
[The migration AddDiffRefToComments](https://github.com/openSUSE/open-build-service/commit/7efef8a2acc6b8dcc4bae10bd0dec4ca337dd665#diff-2e619db1439c44a3b08810c707040b550e4ddfe44becac57141d7d327a7346a6), which is already on master and deployed in production, has a wrong numerical prefix on the migration's filename (version).

As the prefix, `202302224114624`, is higher than any other, when we try to run new migrations later on, the changes in the new migrations are applied into the database but the `db/schema.rb` is never updated and `rails db:version` always return  `202302224114624`.

This was confirmed after applying the most recent migration,  [`AddApiKeyToIssueTracker`](https://github.com/openSUSE/open-build-service/pull/14149), which didn't update the version number inside `db/schema.rb`  or  `rails db:version`. Moreover, whatever changes we had in the `db/schema.rb` should have been included in the PR.

saraycp:
Fixes #14184 

A wrong migration's version number was introduced in the `schema_migrations` table.

```
| 20230116143207  |
| 202302224114624 | <=
| 20230413080503  |
+-----------------+
```
Why? Because the numeric prefix of the migration file was wrong: `db/migrate/202302224114624_add_diff_ref_to_comments.rb`

With this PR we:
  - Rename the migration filename to `db/migrate/20230224114624_add_diff_ref_to_comments.rb` (correct prefix with `yyyymmddhhmmss` format).
  - Set the real last version in `db/schema.rb` which is: `20230413080503` not `202302224114624`.
  - Reflect the addition of the new column in the `db/schema.rb` and annotations.

  
**\*** It's important we correct the version  in the `schema_migrations` table before running any migration. We finally decided to not use a data migration to do so but do it manually. How? In the production server, we update the `version` column of the `schema_migrations` table by replacing  `202302224114624`  with `20230224114624`.


src/api/db/data/20230221000000_fix_migration_version.rb
```diff
@@ -0,0 +1,20 @@
...
+end
+
+# This data migration fixes a wrong migration's version number introduced in the `schema_migrations` table.
+# Assuming that `rails db:migrate:with_data` will alternate schema and data migrations following the date order,
+# we name this file with an old date. This way we force this to run before any other schema migration.
+
+class FixMigrationVersion < ActiveRecord::Migration[7.0]
+  def up
+    wrong_version = DummySchemaMigration.find_by(version: '202302224114624')
+    wrong_version&.update(version: '20230224114624')
```
saraycp, line 14: In production, the version will be there. In new instances it won't, that's why we do the update only if it's present.

src/api/db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 202302224114624) do
+ActiveRecord::Schema[7.0].define(version: 2023_04_13_080503) do
```
krauselukas, line 13:
```suggestion
ActiveRecord::Schema[7.0].define(version: 20230413080503) do
```

krauselukas: The version format looks wrong, or does it work with the underscores? @saraycp

krauselukas (changes requested):

src/api/db/schema.rb
saraycp: In fact, I did set `20230413080503` but running `rails db:migrate:with_data` afterwards changed it to `2023_04_13_080503` (even if it didn't migrate anything).

You can see here it's a correct format: https://api.rubyonrails.org/classes/ActiveRecord/Schema/Definition/ClassMethods.html#method-i-define

krauselukas:
@saraycp after running the data migration locally I receive a diff in the schema 
```
--- a/src/api/db/schema.rb
+++ b/src/api/db/schema.rb
@@ -554,7 +554,6 @@ ActiveRecord::Schema[7.0].define(version: 2023_04_13_080503) do
     t.datetime "issues_updated", precision: nil, null: false
     t.boolean "enable_fetch", default: false
     t.boolean "publish_issues", default: true
-    t.string "api_key"
   end
```

saraycp:
> 1. Setup DB based on current master branch

Did you start from a clean DB or did you reset the DB? If so, that doesn't reflect what we have in production.

It's important you first run the migrations except the last one, then run the last migration an finally you apply our changes and run 'rails db:migrate:with_data'

saraycp:
@krauselukas I'll describe in detail the steps to test it. There are two possible scenarios: an instance set from scratch (like in CircleCI) and a production-like one.

saraycp:
We agreed on not using a data migration, but manually modifying the affected version in production servers before we deploy this PR.

krauselukas (approved):

danidoni (approved):

PR #14231, opened 2023-04-25
OpenAPI specifications GET/DELETE status/message/{id}

rubhanazeem:
specifications for GET/DELETE status/message/{id}

### DIFF
--- a/src/api/public/apidocs-new/components/responses/not_found.yaml
+++ b/src/api/public/apidocs-new/components/responses/not_found.yaml
@@ -1,0 +1,13 @@
+description: |
+  Not Found.
+
+  XML Schema used for body validation: [status.xsd](../schema/status.xsd)
+content:
+  application/xml; charset=utf-8:
+    schema:
+      $ref: '../schemas/api_response.yaml'
+    examples:
+      Not Found:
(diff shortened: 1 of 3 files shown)


src/api/public/apidocs-new/components/responses/not_found.yaml
krauselukas, line 1: @rubhanazeem I would just include the line about the schema validation, like here https://github.com/openSUSE/open-build-service/blob/e71cb3c5d350f2545cbd2a89bdea6e6beee816eb/src/api/public/apidocs-new/components/responses/person/not_found.yaml#L4 or maybe even better, not duplicating the response, but making a generic one, and only provide the correct example

rubhanazeem: @krauselukas I've added the line about schema validation. I haven't used shared response from the person namespace because it is placed in the person namespace and I'd have to move it out and edit the reference in person endpoints. We can do that in a follow-up but for now, I've added the missing validation line

krauselukas (approved):

saraycp:
The build failure is no related to this PR, so merging.

PR #14244, opened 2023-04-26
Convert build results tab to a separate action

rubhanazeem:
This PR converts the build results tab to a separate action. Please only review the last two commits because this PR is rebased from https://github.com/openSUSE/open-build-service/pull/14275


krauselukas (commented):
Besides the comments from @saraycp and the failing feature test I dont have more to add @rubhanazeem

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -292,6 +292,18 @@ def conversation
     @active = 'conversation'
   end
 
+  def build_results
+    redirect_to request_show_path(params[:number], params[:request_action_id]) unless @action[:sprj] || @action[:spkg]
```
saraycp, line 296: This is a temporary patch to fix this misbehavour:

content_of_build_result_when_no_tab

rubhanazeem: Build results are not available for certain actions such as `add_role`. That's why we have this redirect. Previously we were just rendering the conversation tab in such cases. 

@saraycp, @danidoni , and I have discussed that a better fix would be a refactoring to decouple requests and actions logic.

src/api/app/views/webui/request/build_results.html.haml
```diff
@@ -0,0 +1,37 @@
+:ruby
+  @pagetitle = "Request #{@bs_request.number}"
+  @pagetitle += ": #{@action[:name]}"
```
danidoni, line 3: Why those two lines are split?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -292,6 +292,18 @@ def conversation
...
+
+    @active = 'build_results'
+    @project = @staging_project || @action[:sprj]
+    @buildable = @action[:spkg] || @project
+
+    @ajax_data = {}
+    @ajax_data['project'] = @project if @project
+    @ajax_data['package'] = @action[:spkg] if @action[:spkg]
```
ncounter, line 304: Can this be squeezed into something like

```
@ajax_data = {
  project: @project if @project,
  package: @action[:spkg] if @action[:spkg]
}
```

Could you also explain why do you check `if @project` and `if @action[:spkg]` in the assignment? I mean, if the condition would be false it means there is no `@project`, but there is no `else` so I don't really see the difference to assign directly the parameter no matter what the value is. It would result into something like:
```
@ajax_data = {
  project: @project,
  package: @action[:spkg]
}
```
Am I missing something?
If the expectation is something different maybe you want to use a ternary operator and assign something else, just in case?

danidoni: The difference is that using the if will create the `project` key in `@ajax_data` only when there is something to fill the key with, otherwise you'll get a hash like `{ 'project': nil }`

rubhanazeem: These values are used in initializing some HTML data attributes. I guess the checks are important.

src/api/app/views/webui/request/build_results.html.haml
rubhanazeem: No specific reason. Other pages use the same style of initializing the title.

src/api/app/controllers/webui/request_controller.rb
krauselukas: @rubhanazeem just one thing, could you please make this a separate commit with some description. I know your reasoning's for adding this since we talked, but it gets hard to keep track of those changes in the PR.

src/api/app/controllers/webui/request_controller.rb
ncounter: I prefer the collapsed way actually, but at this point it is more about a personal style feeling so, as long as the majority prefer this way, I am totally fine. Feel free to keep it as it is. Thanks for double checking btw. :+1:

src/api/app/controllers/webui/request_controller.rb
rubhanazeem: This redirect will be part of other follow-up PRs as well in a similar pattern. The change we are talking about is not a small refactoring. I'd leave the change as part of the main commit because with that refactoring we might as well remove the whole action

PR #14257, opened 2023-04-28
Convert request#conversations to request#show

rubhanazeem:
This pr is a cleanup that converts request#conversation to request#beta_show. After converting tabs to separate pages we are not using `beta_show`. `beta_show` renders the `conversation` page. It's better if we replace the conversation page with beta_show


danidoni (approved):

src/api/app/views/webui/request/beta_show.html.haml
krauselukas, line 1: @rubhanazeem is the call to the javascript function in line 110 still needed? Looks like this is already called in the `_collapsible_text.html.haml` partial

krauselukas:
@rubhanazeem there is now a merge commit in the PR, could you remove it and rebase instead to keep master branch clean :)

src/api/app/views/webui/request/beta_show.html.haml
rubhanazeem: Thanks for checking it out. The javascript function is not needed anymore. I've removed it :)

krauselukas (approved):

PR #14253, opened 2023-04-27
Document staged requests endpoints

saraycp:
**Reviewing Tips:**

- Open requests towards the original project (which has a staging workflow). You'll see those requests in the backlog.
- Use this endpoint to assign any of those requests to one of the staging projects.
- Play around with listing all the staged requests and "unstage" some of them too.

Controller: Staging::StagedRequestsController < Staging::StagingController

PR #14259, opened 2023-04-28
Document DELETE staged requests endpoint

saraycp:
Endpoint: DELETE `/staging/{project_name}/staged_requests`

Please, review and merge https://github.com/openSUSE/open-build-service/pull/14253 first.

### DIFF
--- a/src/api/public/apidocs-new/paths/staging_project_name_staged_requests.yaml
+++ b/src/api/public/apidocs-new/paths/staging_project_name_staged_requests.yaml
@@ -1,0 +1,75 @@
+delete:
+  summary: Delete requests from any of the staging projects.
+  description: Delete the given requests from any of the staging projects they were staged in.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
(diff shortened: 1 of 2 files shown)


src/api/public/apidocs-new/paths/staging_project_name_staged_requests.yaml
```diff
@@ -0,0 +1,75 @@
+delete:
+  summary: Delete requests from any of the staging projects.
+  description: Delete the given requests from any of the staging projects they were staged in.
+  security:
+    - basic_authentication: []
+  parameters:
+    - $ref: '../components/parameters/project_name.yaml'
+  requestBody:
+    description: List of requests to be removed from the staging projects.
```
ncounter, line 9: I'd rather use `deleted` instead of `removed` for consistency with the rest of the text of this endpoint.
And I'd actually also omit the `from the staging projects` here, it has already been said in the endpoint description, here instead it's only a matter of defining the `requestBody` data.

ncounter (approved):

danidoni:
The build failure is not related to this PR, so merging.

PR #14260, opened 2023-04-28
Fix pathname and shebang when calling `resolve_swagger_yaml` script during build

krauselukas:


dist/obs-server.spec
```diff
@@ -498,8 +498,8 @@ make
 
 # combine swagger yaml files to one big yaml file by resolving all references
 # and replace the development version
-pushd /dist
-ruby resolve_swagger_yaml.rb -i %{__obs_api_prefix}/public/apidocs-new/OBS-v2.10.50.yaml -o %{__obs_api_prefix}/public/apidocs-new/OBS-v2.10.50.yaml -f
+pushd dist/
+./resolve_swagger_yaml.rb -i $RPM_BUILD_ROOT%{__obs_api_prefix}/public/apidocs-new/OBS-v2.10.50.yaml -o $RPM_BUILD_ROOT%{__obs_api_prefix}/public/apidocs-new/OBS-v2.10.50.yaml -f
```
hennevogel, line 502:
```suggestion
./resolve_swagger_yaml.rb -i ../src/api/public/apidocs-new/OBS-v2.10.50.yaml -o ../src/api/public/apidocs-new/OBS-v2.10.50.yaml -f
```

hennevogel (changes requested):

dist/obs-server.spec
hennevogel: `$RPM_BUILD_ROOT` is where the next section of the spec file will install files too. You are still preparing your source for this.

krauselukas: yes thats better, pushed

hennevogel (approved):

PR #14294, opened 2023-05-09
Refactor maintenance related dev tasks

eduardoj:
Move maintenance related development Rails tasks to its own task namespace: `dev:maintenance:...`.

Make it easier to add tasks related to maintenance, such as creating different types of requests related to maintenance for the development environtment.

Now we can call:
- `rake dev:test_data:create`

as we did before, and it will add a basic maintenance environment.

### DIFF
--- a/src/api/lib/tasks/dev.rake
+++ b/src/api/lib/tasks/dev.rake
@@ -127,15 +127,6 @@ namespace :dev do
 
       # we need to set the user again because some factories set the user back to nil :(
       User.session = admin
-      update_project = create(:update_project, target_project: leap, name: "#{leap.name}:Update")
-      create(
-        :maintenance_project,
-        name: 'MaintenanceProject',
-        title: 'official maintenance space',
-        target_project: update_project,
-        maintainer: admin
-      )
-
       # Create factory dashboard projects
       factory = create(:project, name: 'openSUSE:Factory')
       sworkflow = create(:staging_workflow, project: factory)
@@ -234,8 +225,8 @@ namespace :dev do
       # Create a request with multiple submit actions and diffs
       Rake::Task['dev:requests:request_with_multiple_submit_actions_builds_and_diffs'].invoke
 
-      # Create a request with a couple of maintenance incident request actions
-      Rake::Task['dev:requests:request_with_maintenance_incident_actions'].invoke
+      # Create a maintenance environment with maintenance requests
+      Rake::Task['dev:test_data:maintenance'].invoke
 
       # Create a request with a delete request action
       Rake::Task['dev:requests:request_with_delete_action'].invoke
--- a/src/api/lib/tasks/dev/requests.rake
+++ b/src/api/lib/tasks/dev/requests.rake
@@ -162,45 +162,6 @@ namespace :dev do
       puts "  - Create a couple of repositories in project #{iggy_home_project.name}"
     end
 
-    desc 'Creates a request with maintenance incident actions'
-    task request_with_maintenance_incident_actions: :development_environment do
-      require 'factory_bot'
-      include FactoryBot::Syntax::Methods
-
-      iggy = User.find_by(login: 'Iggy') || create(:staff_user, login: 'Iggy')
-      request = build(
-        :bs_request_with_maintenance_incident_action,
-        creator: iggy
-      )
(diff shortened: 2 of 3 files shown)


src/api/lib/tasks/dev/maintenance.rake
```diff
@@ -0,0 +1,72 @@
...
+        title: 'official maintenance space',
+        target_project: update_project,
+        maintainer: admin,
+        commit_user: admin
+      )
+    end
+
+    desc 'Create a request with maintenance incident actions'
+    task request_with_incident_actions: :development_environment do
+      require 'factory_bot'
```
krauselukas, line 28: I think you have to require this also for the maintenance `:project` task, its also using the factories

krauselukas (changes requested):
I would say if you extract the maintenance part, it should work on its own (so no dependencies to run other rake dev task upfront)

src/api/lib/tasks/dev/maintenance.rake
```diff
@@ -0,0 +1,72 @@
...
+    desc 'Set a maintenance environment'
+    task all: :development_environment do
+      Rake::Task['dev:maintenance:project'].invoke
+      Rake::Task['dev:maintenance:request_with_incident_actions'].invoke
+    end
+
+    desc 'Create a maintenance project'
+    task project: :development_environment do
+      admin = User.get_default_admin
+      leap = Project.find_by(name: 'openSUSE:Leap:15.0')
```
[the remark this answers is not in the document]
eduardoj: @krauselukas, I created the task `dev:test_data:create_leap_project` and made the dev:maintenance:project task dependent on the `dev:test_data:create_leap_project` one. This should fix the dependency problem.

src/api/lib/tasks/dev/maintenance.rake
eduardoj: True, done.

eduardoj:
> I would say if you extract the maintenance part, it should work on its own (so no dependencies to run other rake dev task upfront)

I created another commit to make the maintenance part work on its own, and also updated de description of the pull request. I tried both rake tasks (`dev:test_data:create` and `dev:maintenance:all`) after the changes and worked. Please have a look again.

src/api/lib/tasks/dev.rake
```diff
@@ -234,14 +225,42 @@ namespace :dev do
...
+      # The interconnect doesn't work unless we set the distributions
+      FetchRemoteDistributionsJob.perform_now
+    end
+
+    desc 'Create Tumbleweed Repository'
+    task create_tw_repository: :create_interconnect do
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+      interconnect = Project.find_by(name: 'openSUSE.org', remoteurl: 'https://api.opensuse.org/public')
+      create(:repository, name: 'snapshot', project: interconnect, remote_project_name: 'openSUSE:Factory')
```
hennevogel, line 252: Setting a repository on an interconnect makes little sense. Are we using this somewhere?

src/api/lib/tasks/dev.rake
```diff
@@ -234,14 +225,42 @@ namespace :dev do
...
 
       # Create news
       Rake::Task['dev:news:data'].invoke
     end
+
+    desc 'Create Interconnect'
+    task create_interconnect: :development_environment do
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+      create(:remote_project, name: 'openSUSE.org', remoteurl: 'https://api.opensuse.org/public')
```
hennevogel, line 242: Hm, do you expect people to call this outside of the `test_data:create` task? If so you probably should make sure that it works if there is already a project by that name...

hennevogel:
I don't get it. Why would I want *only* test data for maintenance?

hennevogel (changes requested):

src/api/lib/tasks/dev/maintenance.rake
```diff
@@ -0,0 +1,68 @@
+namespace :dev do
```
hennevogel, line 1: I think this should still be below `test_data` and not `dev`

hennevogel (approved):

PR #14299, opened 2023-05-09
[WIP] Refactoring prepare_request_data method: split and use only useful chunks for different pages

ncounter:
This is a follow-up of the split into different pages of a `Request` page tabs: it was one big page before, now pages are:
- `show` (aka `conversation`)
- `build results`
- `rpm lint`
- `changes`
- `mentioned issues`

The refactoring is about to split up all the data of the former single page to differentiate what is used in which page only.

The refactoring starts from a giant function `prepare_request_data`, and it results into:
- `header_data` : the method that prepares the data for the Request header page, which is always the same for all the pages
- `tabs_data` : the method that prepares the data for the partial that is responsible of showing the right tab, select the active one, etc
- the related and not shared data for each page are instantiated into the proper action method
- few other methods to handle few things left not really bounded to the action method

### Reviewers
- **commit-by-commit** is suggested
- no behavior differences are expected

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -23,6 +23,8 @@ class Webui::RequestController < Webui::WebuiController
                             if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :cache_diff_data, only: [:request_action, :request_action_changes, :show, :build_results, :rpm_lint, :changes, :mentioned_issues]
   before_action :check_beta_user_redirect, only: [:build_results, :rpm_lint, :changes, :mentioned_issues]
+  before_action :check_notifications, only: [:show, :build_results, :rpm_lint, :changes, :mentioned_issues],
+                                      if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
 
   after_action :verify_authorized, only: [:create]
 
@@ -596,7 +598,7 @@ class Webui::RequestController < Webui::WebuiController
     # provided by the action the code goes through
   end
 
-  def prepare_request_data
+  def check_notifications
     return unless User.session && params[:notification_id]
 
     @current_notification = Notification.find(params[:notification_id])
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -17,6 +17,8 @@ class Webui::RequestController < Webui::WebuiController
                                     if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :set_superseded_request, only: [:show, :request_action, :request_action_changes, :build_results, :rpm_lint, :changes, :mentioned_issues]
   before_action :check_ajax, only: :sourcediff
+  before_action :header_data, only: [:show, :build_results, :rpm_lint, :changes, :mentioned_issues],
+                              if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :prepare_request_data, only: [:show, :build_results, :rpm_lint, :changes, :mentioned_issues],
                                        if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :cache_diff_data, only: [:request_action, :request_action_changes, :show, :build_results, :rpm_lint, :changes, :mentioned_issues]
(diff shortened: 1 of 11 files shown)


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -515,53 +540,39 @@ def staging_status(request, target_project)
     }
   end
 
-  def prepare_request_data
-    @is_author = @bs_request.creator == User.possibly_nobody.login
-    @is_target_maintainer = @bs_request.is_target_maintainer?(User.session)
-    reviews = @bs_request.reviews.where(state: 'new')
-    @my_open_reviews = reviews.select { |review| review.matches_user?(User.session) }
-    @can_add_reviews = @bs_request.state.in?([:new, :review]) && (@is_author || @is_target_maintainer || @my_open_reviews.present?)
+  def should_refresh(action)
```
krauselukas, line 543: @ncounter hmm, maybe this is only required for the `changes` view/tab. I think this job is responsible for fetching and caching the source diff. Have you checked if this is really needed for the views other then `changes`?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -515,53 +540,39 @@ def staging_status(request, target_project)
...
-    # search for a project, where the user is not a package maintainer but a project maintainer and show
-    # a hint if that package has some package maintainers (issue#1970)
-    @show_project_maintainer_hint = !@package_maintainers.empty? && @package_maintainers.exclude?(User.session) && any_project_maintained_by_current_user?
+  def tabs_data
+    # Collecting all issues in a hash. Each key is the issue name and the value is a hash containing all the issue details.
 
-    # Handling build results
-    @staging_project = @bs_request.staging_project.name unless @bs_request.staging_project_id.nil?
+    @issues = @action.fetch(:sourcediff, []).reduce({}) { |accumulator, sourcediff| accumulator.merge(sourcediff.fetch('issues', {})) }
     @actions_for_diff = [:submit, :delete, :maintenance_incident, :maintenance_release]
```
krauselukas, line 572: @ncounter initializing this variable `@actions_for_diff` in `tabs_data` , and then using it in some of the other views/controller methods is IMO very implicit. Like I get the impression its just needed for tabs, but its not. Maybe we find a way to improve it

krauselukas:
@ncounter can we try to split this PR up a bit? For me its hard to review, and I'm not very confident that I will spot eventual missing pieces. Even If I go commit by commit. I already saw something that could happen independently. I'm happy to join you on this

src/api/app/controllers/webui/request_controller.rb
ncounter: Well, actually `@actions_for_diff` is used by other views but only to pass it to the `request_tabs` partial, therefore I defined a method where to instantiate variables required by that partial view. It is more odd to me (and I am with you on this) about the `@issues` variable which is initialized in `tabs_data` but it also used in the `mentioned_issues` not only for tabs but also for the content. In this case I agree, it is a bit implicit. But I'd also say it is a common pattern with instance variables (think about `@bs_request`, `@action`, etc.).

Btw, overall I agree and I am open to adapt it.
Do you think something like a _singleton pattern_ would work better? Something like:
```ruby
def get_issues
  return @issues if defined?(@issues)

  @issues = @action.fetch(:sourcediff, []).reduce({}) { |accumulator, sourcediff| accumulator.merge(sourcediff.fetch('issues', {})) }
end
```
and then use the method wherever the instance variable is meant to be fetched.

src/api/app/controllers/webui/request_controller.rb
ncounter: You're right!

On the other hand, now that are we thinking about it, I wonder if it would be nice to trigger it from the `header_data`, just to prepare earlier the diff data whenever someone enters for the first time the request page. But that's something we can adjust later, I'll start dropping it from `build_results` and `rpm_lint`.

src/api/app/controllers/webui/request_controller.rb
hennevogel: What about `[:submit, :delete, :maintenance_incident, :maintenance_release]` is variable? :)

ncounter:
Closing this PR as per offline discussion: single changes have already been cherry picked and merged from separate PRs (https://github.com/openSUSE/open-build-service/pull/14313, https://github.com/openSUSE/open-build-service/pull/14312, https://github.com/openSUSE/open-build-service/pull/14315, https://github.com/openSUSE/open-build-service/pull/14321, https://github.com/openSUSE/open-build-service/pull/14335, https://github.com/openSUSE/open-build-service/pull/14337), the rest is dropped for now as it just modify the code without bringing too much value/improvement. Spin-off refactorings will continue to come from here eventually, but we agreed this PR in too big to digest in one shot.

PR #14306, opened 2023-05-10
Split creation of a request with a delete action in its own rake task

eduardoj:
It doesn't make sense to create a request with two `maintenance_incident` actions and also a `delete` action.


eduardoj:
> Besides a maintenance package, could you extend this to also include patchinfo?

Most probably, yes. If it happens, it will be in other pull request, for sure.

PR #14309, opened 2023-05-11
Reduce diff HTML size and fix delete comment issue

rubhanazeem:
TODO:

- [ ] Check why the delete and edit options are not visible on comments
- [ ] Re-render diff component on comment delete


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -324,6 +325,16 @@ def mentioned_issues
     @active = 'mentioned_issues'
   end
 
+  def render_diff
```
rubhanazeem, line 328: It should be removed because the action has been moved to the comments controller.

src/api/app/views/webui/comment/_comment_field.html.haml
```diff
@@ -23,7 +24,7 @@ class: "#{form_method}-comment-form" }, data: { preview_comment_url: preview_com
       .comment-controls.mb-3
         - case form_method
         - when :post
-          = f.submit 'Add comment', class: 'btn btn-primary me-2', data: { disable_with: 'Creating comment...' }, disabled: true
+          = f.submit 'Add comment', class: 'btn btn-primary me-2', data: { disable_with: 'Creating comment...' }
```
rubhanazeem, line 27: Bring back the `disabled: true` value and check why the js connected to this button is not working

src/api/app/views/webui/comment/create.js.erb
```diff
@@ -0,0 +1 @@
+$('#<%= file_name %>').html("<%= escape_javascript(render(DiffComponent.new(diff: diff, file_index: file_index, commentable: commentable, commented_lines: commented_lines, file_name: file_name))) %>");
```
rubhanazeem, line 1: Remove this file. It's not needed anymore

danidoni:
The PR title aims to accomplish two different things, why don't you split the PR in two, then?

PR #14311, opened 2023-05-11
Fix another place where live log is crashing

adrianschroeter:
when building remote package sources.

Also pull request #12351 is required to complete it, but that code there is currently not in a working state anymore, so submitting this independend of it.

### DIFF
--- a/src/api/app/views/webui/package/_tabs.html.haml
+++ b/src/api/app/views/webui/package/_tabs.html.haml
@@ -1,7 +1,7 @@
 .card-header.p-0
   .scrollable-tabs
     = tab_link('Overview', package_show_path(project, package), false, 'scrollable-tab-link')
-    - if package.name == 'patchinfo'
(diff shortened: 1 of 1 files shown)


adrianschroeter:
this is a crashing when in the webui independ of any mails.

Also not limited to scmsync, this is also broken for any setup using a remote project source for building.

hennevogel:
So my question stands, are you trying to fix links included in build failed mails for `setup using a remote project`? Or what are you trying to do here?

adrianschroeter:
The mentioned crashes are fixed by exactly this change and not introduced.

It is not suddenly, I keep fixing these regressions since years in our production systems, there is also https://github.com/openSUSE/open-build-service/pull/12351 for example.

This has nothing todo with any mail, you just can not click on any failed build result in any package where no local package object exists. No matter if scmsync or simple remote object setup. Yes, the links in mails are most likely also not working in addition.

PR #14312, opened 2023-05-11
Refactor request refresh logic

ncounter:
The logic of that piece of code is to check if the `diff` changes are already computed and cached, or it triggers the async job to fetch them.
The refactoring makes all the access to the request page (no matter which tab) to hit that method so the `diff` will be fetched in advance. The refactoring also renames the `@refresh` variable into some more explicit flag with the name of `@diff_not_cached`, which is in line with the id of the hash retrieved from the backend job.

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -19,6 +19,8 @@ class Webui::RequestController < Webui::WebuiController
   before_action :check_ajax, only: :sourcediff
   before_action :prepare_request_data, only: [:show, :build_results, :rpm_lint, :changes, :mentioned_issues],
                                        if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
+  before_action :cache_diff_data, only: [:show, :build_results, :rpm_lint, :changes, :mentioned_issues],
+                                  if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :check_beta_user_redirect, only: [:build_results, :rpm_lint, :changes, :mentioned_issues]
 
   after_action :verify_authorized, only: [:create]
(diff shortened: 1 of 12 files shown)


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -503,7 +503,9 @@ def staging_status(request, target_project)
   end
 
   def cache_diff_data
-    return unless (@refresh = @action[:diff_not_cached])
+    return @diff_not_cached = true unless defined?(@action)
+
+    return unless (@diff_not_cached = @action[:diff_not_cached])
```
eduardoj: I guess that checking for `defined?(@action)` is now needed because of the previous commit, which moves the assignment to the `@refresh` variable into the `cache_diff_data before_action` method.

Currently, without introducing the `cache_diff_data before_action` method, the variable `@action` gets assigned before `@refresh` gets assigned, [here](https://github.com/openSUSE/open-build-service/blob/8f5ab39aea4552297a35bd9fa4e4882625aea43a/src/api/app/controllers/webui/request_controller.rb#L159-L163), for example.

In these cases, shouldn't we keep assigning the `@refresh` variable after `@action` gets assigned?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -515,6 +502,16 @@ def staging_status(request, target_project)
     }
   end
 
+  def cache_diff_data
+    return @diff_not_cached = true unless defined?(@action)
```
hennevogel, line 506: Hm, how can this happen?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -515,6 +502,16 @@ def staging_status(request, target_project)
...
+
+    return unless (@diff_not_cached = @action[:diff_not_cached])
```
hennevogel, line 508: I'm beginning to wonder why we need this extra variable if we have `@action[:diff_not_cached]`...

hennevogel (changes requested):

src/api/app/controllers/webui/request_controller.rb
ncounter: You triggered one of my _hot buttons_ so ----> https://github.com/openSUSE/open-build-service/pull/14312/commits/4d7b597629c5fbdad0708dcd5b1c557be012ae23

hennevogel (approved):

PR #14319, opened 2023-05-12
Bug fix: make sure loadChanges receives parameters to build url properly

ncounter:
Since the conversion from tabs to separate pages, in the request page the `#changes-tab` id got dropped. Therefore the `loadChanges` is no longer able to fetch the parameters (`request number` and `action id`) properly.

The bug fix passes those parameters on calling the function instead of delegating to the function the search and read of those parameters relying on some other element.

### DIFF
--- a/src/api/app/assets/javascripts/webui/request.js
+++ b/src/api/app/assets/javascripts/webui/request.js
@@ -185,10 +185,9 @@ function loadDiffs(element){
   });
 }
 
-function loadChanges() { // jshint ignore:line
+function loadChanges(requestNumber, requestActionId) { // jshint ignore:line
   $('.loading-diff').removeClass('invisible');
-  var element = $('#changes-tab');
-  var url = '/request/' + element.data('request-number') + '/request_action/' + element.data('request-action-id') + '/changes';
+  var url = '/request/' + requestNumber + '/request_action/' + requestActionId + '/changes';
   $.ajax({
     url: url,
     success: function() {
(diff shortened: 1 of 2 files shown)


src/api/app/assets/javascripts/webui/request.js
```diff
@@ -185,10 +185,9 @@ function loadDiffs(element){
   });
 }
 
-function loadChanges() { // jshint ignore:line
+function loadChanges(requestId, requestActionId) { // jshint ignore:line
   $('.loading-diff').removeClass('invisible');
-  var element = $('#changes-tab');
-  var url = '/request/' + element.data('request-number') + '/request_action/' + element.data('request-action-id') + '/changes';
+  var url = '/request/' + requestId + '/request_action/' + requestActionId + '/changes';
```
krauselukas, line 190: @ncounter this route is using the bs_request.number, not the bs_request.id. Most of the time its the same, but I'm not sure if we can rely on it. So better use the number as intended for this route https://github.com/openSUSE/open-build-service/blob/809d4f0930b8048e730c0dc5ceef3e97e1eda7ec/src/api/config/routes/webui_routes.rb#L303

krauselukas (changes requested):
See my comment

src/api/app/assets/javascripts/webui/request.js
ncounter: A-ha!! That's a good catch. Thank you. Fixing..

krauselukas (approved):

PR #14335, opened 2023-05-15
Create `can_add_reviews?` pundit policy and use in request beta views

krauselukas:
This code is for authorization and therefore should be in a
pundit policy.

Extracted out of https://github.com/openSUSE/open-build-service/pull/14318 to ease review

### DIFF
--- a/src/api/app/policies/bs_request_policy.rb
+++ b/src/api/app/policies/bs_request_policy.rb
@@ -13,4 +13,11 @@ class BsRequestPolicy < ApplicationPolicy
     is_author = record.creator == user.login
     record.state.in?([:new, :review, :declined]) && (is_target_maintainer || is_author)
   end
+
+  def add_reviews?
+    is_target_maintainer = record.is_target_maintainer?(user)
+    is_author = record.creator == user.login
+    has_open_reviews = record.reviews.where(state: 'new').select { |review| review.matches_user?(user) }.present?
+    record.state.in?([:new, :review]) && (is_author || is_target_maintainer || has_open_reviews)
+  end
 end
--- a/src/api/app/views/webui/request/beta_show_tabs/_conversation_aside.html.haml
+++ b/src/api/app/views/webui/request/beta_show_tabs/_conversation_aside.html.haml
@@ -3,16 +3,15 @@
   - if request_reviews.present?
     .mt-4
       .text-end
-        = render AddReviewDropdownComponent.new(bs_request: bs_request, user: User.session,
-                                              can_add_reviews: can_add_reviews,
-                                              my_open_reviews: my_open_reviews)
+        = render AddReviewDropdownComponent.new(bs_request: bs_request, user: User.session, my_open_reviews: my_open_reviews)
       = render AddReviewCollapsibleComponent.new
       .mb-2
         %h6
           Reviewers
       .mb-4
         = render partial: 'webui/request/beta_show_tabs/review_summary', collection: request_reviews, as: :review
-  = render partial: 'webui/request/beta_show_tabs/ask_for_review', locals: { can_add_reviews: can_add_reviews }
+  - if policy(bs_request).add_reviews?
+    = render partial: 'webui/request/beta_show_tabs/ask_for_review'
 
   -# PACKAGE MAINTAINERS
   - unless package_maintainers.empty?
--- a/src/api/spec/policies/bs_request_policy_spec.rb
+++ b/src/api/spec/policies/bs_request_policy_spec.rb
@@ -1,0 +1,75 @@
+require 'rails_helper'
+
+RSpec.describe BsRequestPolicy, type: :policy do
+  subject { BsRequestPolicy }
+
+  permissions :add_reviews? do
+    let(:user) { create(:confirmed_user, login: 'iggy') }
+    let(:author) { create(:confirmed_user, login: 'foo') }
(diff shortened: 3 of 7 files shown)


src/api/app/policies/bs_request_policy.rb
```diff
@@ -13,4 +13,11 @@ def handle_request?
     is_author = record.creator == user.login
     record.state.in?([:new, :review, :declined]) && (is_target_maintainer || is_author)
   end
+
+  def can_add_reviews?
```
hennevogel, line 17: 💭  Drop the `can_`  prefix. A policy is always about can and can't....

src/api/spec/policies/bs_request_policy_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      context 'and the user is the author of the record' do
+        it { expect(subject).to permit(author, bs_request) }
+      end
+
+      context 'and the user is not the author of the record' do
+        it { expect(subject).not_to permit(user, bs_request) }
+      end
+    end
+
+    context 'when the record is in state review' do
```
hennevogel, line 58: This and `'when the record is in state new' ` is the same no?

hennevogel (changes requested):

src/api/spec/policies/bs_request_policy_spec.rb
krauselukas: yes, somehow dropped changing the state of the request to `review` for this context. Its basically the same, but I thought its one of the conditions of the policy. But I will DRY it

src/api/app/policies/bs_request_policy.rb
krauselukas: true

src/api/app/views/webui/request/beta_show_tabs/_conversation_aside.html.haml
```diff
@@ -3,16 +3,14 @@
...
-                                              my_open_reviews: my_open_reviews)
+        = render AddReviewDropdownComponent.new(bs_request: bs_request, user: User.session, my_open_reviews: my_open_reviews)
       = render AddReviewCollapsibleComponent.new
       .mb-2
         %h6
           Reviewers
       .mb-4
         = render partial: 'webui/request/beta_show_tabs/review_summary', collection: request_reviews, as: :review
-  = render partial: 'webui/request/beta_show_tabs/ask_for_review', locals: { can_add_reviews: can_add_reviews }
+  = render partial: 'webui/request/beta_show_tabs/ask_for_review', locals: { can_add_reviews: policy(bs_request).add_reviews? }
```
ncounter, line 13: What about wrapping the `render partial` with `if policy(...` and get rid of the `locals` completely here and inside the `ask_for_review` partial? If it was a view component that can compute its own policy it would be better (like the `AddReviewDropdownComponent`), but this is a simple stateless partial I'd rather keep it _ignorant_ about the condition.

src/api/app/policies/bs_request_policy.rb
```diff
@@ -13,4 +13,11 @@ def handle_request?
     is_author = record.creator == user.login
     record.state.in?([:new, :review, :declined]) && (is_target_maintainer || is_author)
   end
+
+  def add_reviews?
+    is_target_maintainer = record.is_target_maintainer?(user)
+    is_author = record.creator == user.login
+    open_reviews = record.reviews.where(state: 'new').select { |review| review.matches_user?(user) }
```
ncounter, line 20: IIUC it is used only as a `.present?` boolean condition check. Maybe it could be more straight forward to save the boolean only: `has_open_reviews = record.reviews.where............present?` only

src/api/app/views/webui/request/beta_show_tabs/_conversation_aside.html.haml
krauselukas: done

src/api/spec/policies/bs_request_policy_spec.rb
krauselukas: I moved the tests to a shared example and running it with the two states `review` and `new`

src/api/app/policies/bs_request_policy.rb
krauselukas: done

src/api/app/policies/bs_request_policy.rb
```diff
@@ -13,4 +13,11 @@ def handle_request?
     is_author = record.creator == user.login
     record.state.in?([:new, :review, :declined]) && (is_target_maintainer || is_author)
   end
+
+  def add_reviews?
+    is_target_maintainer = record.is_target_maintainer?(user)
+    is_author = record.creator == user.login
+    has_open_reviews = record.reviews.where(state: 'new').select { |review| review.matches_user?(user) }.present?
+    record.state.in?([:new, :review]) && (is_author || is_target_maintainer || has_open_reviews.present?)
```
ncounter, line 21: Due to [this comment](https://github.com/openSUSE/open-build-service/pull/14335#pullrequestreview-1429987687) and the change you made, I think `.present?` should be dropped from `has_open_reviews` here now

krauselukas: True, overlooked this. Done :)

danidoni (approved):

hennevogel (approved):

PR #14339, opened 2023-05-16
Fix build on tumbleweed

darix:
without consistently using the ruby suffix it tried to build the gems for ruby 3.2 while only 3.1 was available.

### DIFF
--- a/dist/obs-bundled-gems.spec
+++ b/dist/obs-bundled-gems.spec
@@ -16,7 +16,8 @@
 #
 
 %if 0%{?suse_version}
-%define __obs_ruby_interpreter /usr/bin/ruby.ruby3.1
+%define __obs_ruby_suffix      ruby3.1
+%define __obs_ruby_interpreter /usr/bin/ruby.%{__obs_ruby_suffix}
 %define rack_version %(%{__obs_ruby_interpreter} -r rack -e "puts Rack::RELEASE")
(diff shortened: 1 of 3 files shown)


dist/obs-bundled-gems.spec
```diff
@@ -16,7 +16,8 @@
 #
 
 %if 0%{?suse_version}
-%define __obs_ruby_interpreter /usr/bin/ruby.ruby3.1
+%define __obs_ruby_suffix      ruby3.1
+%define __obs_ruby_interpreter /usr/bin/ruby.%{__obs_ruby_suffix}
```
eduardoj, line 20:
```suggestion
%define __obs_ruby_suffix      .ruby3.1
%define __obs_ruby_interpreter /usr/bin/ruby%{__obs_ruby_suffix}
```

Could it be possible to use the same suffix for both `__obs_ruby_suffix` and `OBS_RUBY_SUFFIX`, starting both with a dot (`.`)?

darix: the current solution has the advantage that if you do not want to override the setting you could use

```
%define __obs_ruby_suffix %rb_default_ruby_suffix
```

though i can change the code in the make file to use the dot-less version.

danidoni:
Is this still relevant?

darix:
probably?

PR #14345, opened 2023-05-17
Display links to source and target files in diffs

hellcp-work:
Screenshot from 2023-05-18 11-29-31

Displays links to the source and target files inline with the git range information. I wanted to include anchors for the lines that the range is referring to, but as far as I can tell codemirror doesn't have a simple way of having line ids in html.

To verify this feature:
- Create a request with a diff
- Open changes tab in the beta view
- Click the new icons in the range lines

### DIFF
--- a/src/api/app/components/diff_component.html.haml
+++ b/src/api/app/components/diff_component.html.haml
@@ -5,7 +5,25 @@
       .line-container{ id: }
         %div{ class: "line #{line.state}" }>
           - case line.state
-          - when 'comment', 'range'
+          - when 'range'
+            - if source_file
+              = link_to(source_file, title: 'See original file') do
+                .number.p-0.text-center.link-danger<
+                  %i.fas.fa-file-arrow-down
+            - else
+              %a
+                .number
+            - if target_file
+              = link_to(target_file, title: 'See changed file') do
+                .number.p-0.text-center.link-success<
+                  %i.fas.fa-file-arrow-up
+            - else
+              %a
+                .number
+            .value<
+              .text-muted.ms-3
+                = line.content
+          - when 'comment'
             .offset
               .text-muted.text
                 = line.content
--- a/src/api/app/components/diff_component.html.haml
+++ b/src/api/app/components/diff_component.html.haml
@@ -14,7 +14,7 @@
               %a
                 .number
             - if target_file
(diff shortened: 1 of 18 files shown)


src/api/app/components/diff_component.html.haml
```diff
@@ -5,7 +5,25 @@
       .line-container{ id: }
         %div{ class: "line #{line.state}" }>
           - case line.state
-          - when 'comment', 'range'
+          - when 'range'
+            - if source_file
+              = link_to(source_file) do
+                .number<
+                  %i.fas.fa-arrows-up-to-line
```
hennevogel, line 12: I understand what you want to imply here but I don't think this is clear. You will have to get the reference to PRs which many people will not have.

hennevogel (changes requested):

src/api/app/components/diff_component.html.haml
hennevogel: Yeah I think those are way better communicating what they do...

src/api/app/components/diff_component.html.haml
```diff
@@ -5,7 +5,25 @@
...
+          - when 'range'
+            - if source_file
+              = link_to(source_file, title: 'See original file') do
+                .number.p-0.text-center.link-danger<
+                  %i.fas.fa-file-arrow-down
+            - else
+              %a
+                .number
+            - if target_file
+              = link_to(target_file, title: 'See changed file') do
```
hennevogel, line 17:
```suggestion
              = link_to(target_file, title: 'See submitted file') do
```

hennevogel (changes requested):

src/api/app/components/diff_list_component.html.haml
```diff
@@ -1,13 +1,16 @@
 - diff_list.each_with_index do |(name, file_info), file_index|
-  - state = file_info['state']
-  - expanded = expand?(name, state, file_index)
+  :ruby
```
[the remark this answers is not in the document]
hennevogel: What would be enough?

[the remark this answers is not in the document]
hennevogel: alright, this is a bit much then...

hennevogel (approved):

PR #14346, opened 2023-05-17
Request tab visibility helper

ncounter:
Instead of passing around an array of `action[:type]`s only to decide whether or not to show a request action tab, let an helper compute the tab visibility instead.
This PR is a refactoring in order to centralize visibility checks, then adding more complexity to the conditions as per https://github.com/openSUSE/open-build-service/pull/14347


ncounter:
Superseded by https://github.com/openSUSE/open-build-service/pull/14349

PR #14347, opened 2023-05-17
Refine request tabs visibility conditions

ncounter:
Request action tabs visibility are too generic. In case of a `:delete` action, the following conditions are missing:
- `changes` tab should be visible only if the `:delete` action's target is a `package`. If it is a `project` no `changes` tab instead
- `mentioned issues` tab should never be visible in case of a `:delete` action

This PR, based on https://github.com/openSUSE/open-build-service/pull/14346, refine the visibility conditions according to the above statements.

### Reviewers
Do not consider the first three commits, they are included in https://github.com/openSUSE/open-build-service/pull/14346 instead. This PR is rebased on top of that one

### DIFF
--- a/src/api/spec/helpers/webui/request_helper_spec.rb
+++ b/src/api/spec/helpers/webui/request_helper_spec.rb
@@ -151,6 +151,9 @@ RSpec.describe Webui::RequestHelper do
 
         it { is_expected.to match(expected_regex) }
       end
+
+      it { expect(request_action_tab_visibility(action, 'conversation')).to be(true) }
+      it { expect(request_action_tab_visibility(action, 'mentioned_issues')).to be(false) }
     end
 
     context 'when action is :add_role' do


src/api/app/helpers/webui/request_helper.rb
```diff
@@ -199,4 +199,18 @@ def next_prev_path(**opts)
       request_show_path(parameters)
     end
   end
+
+  # Handle tabs visibility
+  def request_action_tab_visibility(action, tab_name)
```
danidoni, line 204: Instead of extracting the data from the action and then computing the logic, we should ask the `action` if I should render that tab or not. So ideally each bs request action class should have a method telling the view if it should render the tab or not.

ncounter: I see your point but, to be honest, I prefer to avoid _view_ conditions in the _model_. It if was about a _policy_ or _specific property_ of the action itself I would agree 100%. In this case instead, the model is ignorant about how the elements are rendered and presented, it is just a view condition method that belongs to the view, and I just wrapped and centralized in a method to be more handy. If the webui page will change in the future (like no tabs, or anything different), the model should be loose from that change.

danidoni: I'm okay in avoiding view conditions in the model, but I think we can live with that in order to achieve better code structure. Also the view helpers are tightly coupled to the `bs_request_action` details right now, and I think that's worse

And the view knows too much: we're breaking the [Tell, don't ask](https://thoughtbot.com/blog/tell-dont-ask) principle.

If you are not comfortable with that, then maybe we need a [decorator](https://thoughtbot.com/blog/tidy-views-and-beyond-with-decorators), a class sitting between the model and the view that encapsulates the  "view logic" (as we do with [Finders](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/search_finder.rb)).

PR #14349, opened 2023-05-18
Refactoring tab visibility - use decorator instead of helper

ncounter:
Instead of passing around an array of `action[:type]`s only to decide whether or not to show a request action tab, let the object itself tell about its tabs visibility instead.

This PR refactors the way the request page tabs visibility is computed by moving the logic from an helper to a decorator class, based also on [this discussion](https://github.com/openSUSE/open-build-service/pull/14347#pullrequestreview-1431216663).

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -13,6 +13,8 @@ class Webui::RequestController < Webui::WebuiController
                                         if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :set_action_id, only: [:inline_comment, :show, :build_results, :rpm_lint, :changes, :mentioned_issues],
                                 if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
+  before_action :set_bs_request_action, only: [:show, :build_results, :rpm_lint, :changes, :mentioned_issues],
+                                        if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :set_active_action, only: [:inline_comment, :show, :build_results, :rpm_lint, :changes, :mentioned_issues],
                                     if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :set_superseded_request, only: [:show, :request_action, :request_action_changes, :build_results, :rpm_lint, :changes, :mentioned_issues]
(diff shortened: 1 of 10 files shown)


src/api/app/models/bs_request_action_tab_visibility.rb
```diff
@@ -0,0 +1,20 @@
+class BsRequestActionTabVisibility < ApplicationComponent
```
hennevogel, line 1: This is not a ViewComponent is it?

hennevogel (changes requested):

src/api/app/models/bs_request_action_tab_visibility.rb
danidoni: Is this a [Decorator](https://thoughtbot.com/blog/evaluating-alternative-decorator-implementations-in#the-quotplain-old-ruby-objectquot-poro-decorator)?

danidoni (approved):

danidoni:
@hennevogel do you miss something for this PR to be approved?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -488,6 +490,10 @@ def set_action_id
     @action_id = params[:request_action_id] || @supported_actions.first&.id || @actions.first.id
   end
 
+  def set_bs_request_action
+    @bs_request_action = BsRequestAction.find(@action_id)
```
hennevogel, line 494: This will allow me to load *any* BsRequestAction onto the page that I want just by passing the parameter...

hennevogel (changes requested):

src/api/app/controllers/webui/request_controller.rb
ncounter: Addressed in the last changes

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -488,6 +490,16 @@ def set_action_id
     @action_id = params[:request_action_id] || @supported_actions.first&.id || @actions.first.id
   end
 
+  def set_bs_request_action
+    # Make sure the action belongs to the request
+    if @action_id.in?(@bs_request.bs_request_actions.ids)
+      @bs_request_action = BsRequestAction.find(@action_id)
```
danidoni, line 496: I believe you can just do `@bs_request.bs_request_actions.find(@action_id)` (untested)

ncounter: works like a charm, pushed

danidoni (approved):

hennevogel (approved):

PR #14343, opened 2023-05-17
Set up maintenance incident request in development environment

saraycp:
From the `test_data` rake task we are calling `create_maintenance_setup` which intends to set all the data needed to reproduce the maintenance workflow.

So far it sets three maintenance incident requests and all the elements needed before the request.

- maintained projects
- update project
- maintenance project
- packages
- incident request with one action not accepted
- incident request with multiple actions not accepted
- incident request with multiple actions accepted

**Reviewing Tips**

- Run `bundle exec rake dev:test_data:create` locally
- You should see three incident requests:
  - Number 11: one action,  no patchinfo, not accepted
  - Number 12: many actions, with patchinfo, not accepted
  - Number 13: many actions, with patchinfo, accepted

PR #14361, opened 2023-05-22
Create patchinfo factory

eduardoj:
~Depends on #14343.~

### DIFF
--- a/src/api/spec/factories/patchinfo.rb
+++ b/src/api/spec/factories/patchinfo.rb
@@ -1,0 +1,6 @@
+FactoryBot.define do
+  factory :patchinfo, class: 'Patchinfo' do
+    # Patchinfo is an ActiveModel::Model. Override FactoryBot `create` method
+    skip_create
+  end
+end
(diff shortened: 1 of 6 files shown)


src/api/spec/factories/patchinfo.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :patchinfo, class: 'Patchinfo' do
+    # Patchinfo is an ActiveModel::Model. Override FactoryBot `create` method
+    to_create { nil }
```
krauselukas, line 4: @eduardoj maybe you can use `skip_create` here? https://github.com/thoughtbot/factory_bot/blob/main/GETTING_STARTED.md#custom-methods-to-persist-objects

eduardoj: @krauselukas yes. Thanks!

saraycp (commented):
Most of the time you end up calling `create_patchinfo`. Why don't you simply do that internally in the factory?

Calling the factory could look like this: `create(:patchinfo, project: project, package: package, comment: comment, data: data)`

eduardoj:
> Most of the time you end up calling `create_patchinfo`. Why don't you simply do that internally in the factory?
> 
> Calling the factory could look like this: `create(:patchinfo, project: project, package: package, comment: comment, data: data)`

Good idea! I added a commit which creates two factories. Please take a look.

krauselukas (approved):

PR #14363, opened 2023-05-22
Cover another case of incident request

saraycp:
The request doesn't come from a user's branched package but from a package which is "officially" developed on a specific project.

**Reviewing Tips**

Run `bundle exec rake dev:test_data:create` locally. You should see an incident request with number 14: many actions, with patchinfo, accepted

### DIFF
--- a/src/api/lib/tasks/dev/test_data/maintenance.rb
+++ b/src/api/lib/tasks/dev/test_data/maintenance.rb
@@ -49,26 +49,31 @@ module TestData
              commit_user: admin)
     end
 
-    def create_request_with_incident_actions(update_project_branch_name:, maintenance_project:, package_names:, update_project_names:, patchinfo: false)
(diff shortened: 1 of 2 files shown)


saraycp:
We're going to ignore codecov/patch.

src/api/lib/tasks/dev/test_data/maintenance.rb
```diff
@@ -131,6 +131,26 @@ def mimic_patchinfo(project, comment = 'test comment')
       Patchinfo.new.create_patchinfo(project, 'patchinfo', comment: comment)
     end
 
+    # Simulate `osc maintenancerequest servers apache2 openSUSE:Leap:15.4:Update`
+    def mimic_maintenancerequest(source_project_name:, source_package_name:, target_project_name:, target_releaseproject_name:)
```
saraycp, line 135: We've realized this method can be merged with `create_request_with_incident_actions`. We are going to do it, so this PR will be moved to Draft for a while

saraycp:
The code has been refactored. This is ready for review again.

eduardoj (approved):

PR #14366, opened 2023-05-23
Disallow deleting and replying to deleted comments

hellcp-work:
To verify:
- Create a comment
- Reply to the comment
- Delete the root comment
- You will no longer be able to reply to or delete the root comment even as admin

### DIFF
--- a/src/api/app/components/bs_request_comment_component.html.haml
+++ b/src/api/app/components/bs_request_comment_component.html.haml
@@ -42,7 +42,7 @@
          comment: comment, commentable: commentable, element_suffix: "edit_#{comment.id}", has_cancel: true })
 
   -# This should be refactored to avoid relying on global state
-  - if User.session && !comment.user.is_nobody? # rubocop:disable ViewComponent/AvoidGlobalState
+  - if policy(comment).reply?
     .mb-2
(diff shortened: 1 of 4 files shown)


src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -42,7 +42,7 @@
          comment: comment, commentable: commentable, element_suffix: "edit_#{comment.id}", has_cancel: true })
 
   -# This should be refactored to avoid relying on global state
-  - if User.session # rubocop:disable ViewComponent/AvoidGlobalState
+  - if User.session && !comment.user.is_nobody? # rubocop:disable ViewComponent/AvoidGlobalState
```
hennevogel, line 45: You should ask the policy and not replicate it...

hennevogel (changes requested):

src/api/app/policies/comment_policy.rb
```diff
@@ -27,4 +27,10 @@ def update?
 
     user == record.user
   end
+
+  def reply?
+    return false if user.blank? || user.is_nobody? || record.user.is_nobody?
+
+    true
```
eduardoj, line 34:
```suggestion
    !(user.blank? || user.is_nobody? || record.user.is_nobody?)
```
`return false if ...` and in the next sentence return `true` looks redundant to me. What do you think?

hennevogel (approved):

PR #14371, opened 2023-05-23
[api] improve cleanup on user removal

adrianschroeter:
Also remove Event Subscriptions.

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -610,6 +610,7 @@ class User < ApplicationRecord
     self.realname = ''
     self.state = 'deleted'
     comments.destroy_all
+    event_subscriptions.destroy_all
     save!
 
     # wipe also all home projects


bmwiedemann (approved):

src/api/app/models/user.rb
```diff
@@ -674,6 +674,11 @@ def delete!
       project.destroy
     end
 
+    # comments are non-critical personal content, drop them as well. Even though they shouldn't be visiable anymore
```
eduardoj, line 677:
```suggestion
    # comments are non-critical personal content, drop them as well. Even though they shouldn't be visible anymore
```
Typo.

src/api/app/models/user.rb
```diff
@@ -674,6 +674,11 @@ def delete!
       project.destroy
     end
 
+    # comments are non-critical personal content, drop them as well. Even though they shouldn't be visible anymore
+    Comment.where(user: self).delete_all
```
danidoni, line 678: `delete_all` will skip the callbacks defined [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/comment.rb#L17) is that what you want to do?

hennevogel (approved):

PR #14378, opened 2023-05-24
[backend] Make use of projectscmsync functionality in scm bridge

adrianschroeter:
Use package url meta element for now to transfer the information.

Ideally we should found a new element, but it would break external OBS instances


adrianschroeter:
still untested

src/backend/BSSrcServer/Service.pm
```diff
@@ -256,6 +256,7 @@ sub generate_obs_scm_bridge_service {
   my @params;
   push @params, { 'name' => 'url', '_content' => $data->{'url'} };
   push @params, { 'name' => 'projectmode', '_content' => '1' } if $data->{'projectmode'};
+  push @params, { 'name' => 'projectscmsync', '_content' => $data->{'projectscmsync'} } if $data->{'projectscmsync'};
```
mlschroe, line 259: `&& !$data->{'projectmode'}`?

mlschroe (approved):

PR #14383, opened 2023-05-24
Add `project_id` to Gitlab Tag Push payload

krauselukas:
The `:project_id` is also required for the `tag_push` payload since we always need it to in order to report back the status to the correct project through the  Gitlab API in the `gitlab_status_reporter` class. Right now it fails.


krauselukas:
the `project_id` is required here https://github.com/openSUSE/open-build-service/blob/ef6b0ae409576d1cfdb76561a3757509e7745c20/src/api/app/services/gitlab_status_reporter.rb#L15

PR #14384, opened 2023-05-24
Validate supported filter for event on workflow execution

krauselukas:
Right now we have an intermix we the validations for a SCM
workflow. When we intialize the `Workflow` class we run
the `WorkflowFiltersValidator`. This validator makes sure that
all filters in the `Workflow` are valid (so basically that every
filter definition in the `workflow.yml` file is correct).

The problem right now is, we cannot check in this validator that
the current webhook event is supported by the filters (in this
case the branches filter for the `tag_push`) since the class is
always intialized no matter if it is meant for the currently incoming
webhook.

If a certain webhook event is meant for the workflow is checked
when we execute the workflow (`Workflow.call`). Here we check
if the event matches the event filter etc.
Therefore we also have to check here if the current event supports
the filters or not.

An issue occurs currently when there a multiple workflow definitions
in the same `workflow.yml` file. For example:

```
branch:
  steps:
    - branch_package:
        source_project: home:krauselukas
        source_package: hello_world
        target_project: home:krauselukas:test
  filters:
    event: push
    branches:
      only:
        - main
release:
  steps:
    - trigger_services:
        project: home:krauselukas
        package: hello_world
  filters:
    event: tag_push
```

The first workflow is for the `push` event, which supports
the `branches` filter. The second for `tag_push` doesn't.
This will fail, when the event is a `tag_push` since the
`WorkflowFiltersValidator` is called when intializing the workflow
and checks the current webhook event against all filters (which in
this example also includes a branches filter).

With the current architecture we have to validate for this
when the workflow is executed.

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -43,9 +43,12 @@ class Token::Workflow < Token
 
     # This is just an initial generic report to give a feedback asap. Initial status pending
     SCMStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, workflow_run, initial_report: true).call
-    @workflows.each(&:call)
-    SCMStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, workflow_run, 'success', initial_report: true).call
+    @workflows.each do |workflow|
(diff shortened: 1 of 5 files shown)


bluca:
Thank you!

PR #14372, opened 2023-05-23
Set up maintenance release request in development environment

saraycp:
**Reviewing Tips**

Run `bundle exec rake dev:test_data:create` locally. It should create a maintenance **release** request with Number 15, which contains actions from 2 projects, 2 packages and 2 patchinfos.

PR #14386, opened 2023-05-25
Modify request with maintenance release actions factory

eduardoj:
~This pull request depends/"is based" on #14372.~

### DIFF
--- a/src/api/spec/factories/bs_requests.rb
+++ b/src/api/spec/factories/bs_requests.rb
@@ -160,7 +160,7 @@ FactoryBot.define do
       end
     end
 
-    factory :bs_request_with_maintenance_release_action do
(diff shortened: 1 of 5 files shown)


saraycp (approved):

src/api/spec/factories/bs_requests.rb
```diff
@@ -162,6 +162,38 @@
 
     factory :bs_request_with_maintenance_release_actions do
       type { :maintenance_release }
+
+      transient do
+        source_project_name { '' }
```
saraycp, line 167: Why `''` and not `nil`?

saraycp: alright, then nothing to say :smiley_cat:

PR #14398, opened 2023-05-30
[dist]: Add obs config option for obs worker

hudeng-go:
Resolve the problem of possible conflicts between the directory created by the chroot build method and the disk name of QEMU, resulting in build failures


adrianschroeter:
Is that only needed when switching between chroot and kvm?

hudeng-go:
> Is that only needed when switching between chroot and kvm?

yes,when nodes support both chroot and kvm, an error will be reported due to file name conflicts

PR #14403, opened 2023-05-31
Cover keyinfo special cases

hennevogel:
Getting the project attribute from the backend may depend on the config in the signer wrapper script, better not expect it to be there always.

### DIFF
--- a/src/api/app/models/project/signing_key_gpg.rb
+++ b/src/api/app/models/project/signing_key_gpg.rb
@@ -15,10 +15,9 @@ class Project
 
     def initialize(project_name)
       keyinfo = Xmlhash.parse(key_info_for_project(project_name))
-      @origin = keyinfo['project']
-
       return if keyinfo['pubkey'].blank?
 
+      @origin = keyinfo['project'] || project_name
       @id = keyinfo['pubkey']['keyid']
       @user_id = keyinfo['pubkey']['userid']
       @algorithm = keyinfo['pubkey']['algo']
--- a/src/api/app/models/project/signing_key_ssl.rb
+++ b/src/api/app/models/project/signing_key_ssl.rb
@@ -18,10 +18,9 @@ class Project
 
     def initialize(project_name)
       keyinfo = Xmlhash.parse(key_info_for_project(project_name))
-      @origin = keyinfo['project']
-
       return if keyinfo['sslcert'].blank?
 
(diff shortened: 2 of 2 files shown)


krauselukas (approved):

src/api/app/models/project/signing_key_gpg.rb
```diff
@@ -15,7 +15,7 @@ class SigningKeyGPG
 
     def initialize(project_name)
       keyinfo = Xmlhash.parse(key_info_for_project(project_name))
-      @origin = keyinfo['project']
+      @origin = keyinfo['project'] || project_name
 
       return if keyinfo['pubkey'].blank?
```
eduardoj, line 20:
```suggestion
      return if keyinfo['pubkey'].blank?

      @origin = keyinfo['project'] || project_name
```

We could switch the order of the sentences. There is no need to do the assignment, if the condition is fullfiled.

src/api/app/models/project/signing_key_ssl.rb
```diff
@@ -18,7 +18,7 @@ class SigningKeySSL
 
     def initialize(project_name)
       keyinfo = Xmlhash.parse(key_info_for_project(project_name))
-      @origin = keyinfo['project']
+      @origin = keyinfo['project'] || project_name
 
       return if keyinfo['sslcert'].blank?
```
eduardoj, line 23:
```suggestion
      return if keyinfo['sslcert'].blank?

      @origin = keyinfo['project'] || project_name
```

Same case here.

hennevogel: done

src/api/app/models/project/signing_key_gpg.rb
hennevogel: done

eduardoj (approved):

adrianschroeter:
On Mittwoch, 31. Mai 2023, 13:33:11 CEST Henne Vogelsang wrote:
> 
> 
> 
> > +      @origin = keyinfo['project'] || project_name

IMHO you should not report a fantasy project_name, if that
project is not defining the key.

This can be very misleading.

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> you should not report a fantasy project_name

We don't, we just show something if this is different to the project we display this for.

PR #14423, opened 2023-06-02
Start rolling out the new API documentation

eduardoj:
This is the first step on rolling out the new API documentation.

In this PR:
- create `apidocs-old` routes
- create redirections from `/apidocs` to `/apidocs-old`

## For reviewers

1. On a running development environment set up the apidocs:
   - `docker-compose exec frontend bash`
   - `cd /obs/`
   -  `make -C docs/api/api apidocs`
   - `cd /obs/src/`
   - `ln -s ../doc`
1. Access the new path: http://localhost:3000/apidocs-old/
1. Access the old path. It should redirect to the new path: http://localhost:3000/apidocs/

### DIFF
--- a/src/api/config/routes/webui_routes.rb
+++ b/src/api/config/routes/webui_routes.rb
@@ -403,7 +403,10 @@ OBSApi::Application.routes.draw do
 
     ### /apidocs
     get 'apidocs', to: redirect('/apidocs/index')
-    get 'apidocs/(index)' => 'webui/apidocs#index', as: 'apidocs_index'
+    get 'apidocs/(index)' => 'webui/apidocs#index'
+    ### /apidocs-old
+    get 'apidocs-old', to: redirect('/apidocs-old/index')
+    get 'apidocs-old/(index)' => 'webui/apidocs#index', as: 'apidocs_index'
   end
 
   resources :staging_workflows, except: :index, controller: 'webui/staging/workflows', param: :workflow_project, constraints: cons do
--- a/src/api/config/routes/webui_routes.rb
+++ b/src/api/config/routes/webui_routes.rb
@@ -401,12 +401,14 @@ OBSApi::Application.routes.draw do
       end
     end
 
-    ### /apidocs
-    get 'apidocs', to: redirect('/apidocs/index')
-    get 'apidocs/(index)' => 'webui/apidocs#index'
     ### /apidocs-old
     get 'apidocs-old', to: redirect('/apidocs-old/index')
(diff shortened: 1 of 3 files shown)


krauselukas (approved):

src/api/config/routes/webui_routes.rb
```diff
@@ -401,9 +401,14 @@
...
+    ### /apidocs-old
+    get 'apidocs-old', to: redirect('/apidocs-old/index')
+    get 'apidocs-old/(index)' => 'webui/apidocs#index', as: 'apidocs_index'
+
+    # Redirections to be removed, just to keep the legacy routes reachable
     ### /apidocs
-    get 'apidocs', to: redirect('/apidocs/index')
-    get 'apidocs/(index)' => 'webui/apidocs#index', as: 'apidocs_index'
+    get 'apidocs', to: redirect('/apidocs-old/index')
+    get 'apidocs/(index)', to: redirect('apidocs-old/index')
```
rubhanazeem, line 411: If I understand correctly, `/apidocs` will represent the new swagger-ui? If that is not the case then what will be the URL to reach swagger-ui? Maybe I'm missing something?

eduardoj: `/apidocs` will represent the new swagger-ui, yes. But not yet. This plan doesn't perform all the tasks at once. This is the plan:
- First we create `/apidocs-old`, with the same contents of `/apidocs`
- Then we redirect `/apidocs` to `/apidocs-old`
At this point every user trying to access `/apidocs` will be redirected to `/apidocs-old`. No harm here. They will see the same page, with another path.

When we roll out the new documentation `/apidocs` will render the new documentation, with swagger-ui. Those redirections (from `/apidocs` to `/apidocs-old`) will be removed.

This workflow provides a constrained interval of time where users will be redirected to the definitive path. Later on, they will be presented the new interface. That is a less invasive option than to migrate everything at the same time.

rubhanazeem: Ah okay, I see. This is an intermediate step.

rubhanazeem (approved):

danidoni:
Shouldn't we also change the url [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/public/apidocs-new/OBS-v2.10.50.yaml#L77)?

danidoni:
> Shouldn't we also change the url [here](https://github.com/openSUSE/open-build-service/blob/master/src/api/public/apidocs-new/OBS-v2.10.50.yaml#L77)?

Ah, now I see the linked PR :+1:

Issue #8199, opened 2019-08-29
[production][] <status code="400">
  <summary>conflict in file R2spec.spec</summary>
</status>



danidoni:
threw this 1 times
<pre>
&lt;status code=&quot;400&quot;&gt;
  &lt;summary&gt;conflict in file R2spec.spec&lt;/summary&gt;
&lt;/status&gt;

</pre>

~~[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5d673dd0d6bec200085871e1)~~

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5cdea6fe4e8f8d0016bf0438)

PR #14444, opened 2023-06-05
Ignore more backend errors

bmwiedemann:
Ignore more backend errors - similar to #8199

except now the message is

    <status code="400"> <summary>home:regataos/gcc12: conflict in file gcc12.spec</summary> </status>

### DIFF
--- a/src/api/config/initializers/airbrake.rb
+++ b/src/api/config/initializers/airbrake.rb
@@ -84,7 +84,7 @@ def ignore_by_class_and_message?(notice)
 end
 
 def ignore_by_backend_400_message?(message)
-  messages_to_ignore = ['<summary>conflict in file', '<summary>unknown request:', '<summary>bad link',
+  messages_to_ignore = ['<summary>unknown request:', '<summary>bad link',
                         '<summary>broken link in', '<summary>bad files', 'does not exist</summary>',
                         'is illegal</summary>', '<summary>service in progress</summary>', '<summary>service error',
                         '<summary>could not apply patch', '<summary>illegal characters</summary>',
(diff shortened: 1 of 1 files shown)


src/api/config/initializers/airbrake.rb
```diff
@@ -95,6 +95,10 @@ def ignore_by_backend_400_message?(message)
   messages_to_ignore.each do |ignored_error_message|
     return true if message.include?(ignored_error_message)
   end
+  messages_to_ignore_re = ['<summary>[^<>]*: conflict in file [^<>]*</summary>'].freeze
```
danidoni, line 98: What about merging the previous `messages_to_ignore` list and this one? I have the impression that maybe this regexp would cover some of the previous errors also...

hennevogel: It's already included above: `message.include?('<summary>conflict in file')` right?

hennevogel (changes requested):

src/api/config/initializers/airbrake.rb
bmwiedemann: I omitted the `: ` so that now the Regexp matches both.

hennevogel: I don't get it, the message changed or what happened?

bmwiedemann: yes the part with `prj/pkg: ` got added.

danidoni: See [here](https://github.com/openSUSE/open-build-service/pull/14444/files#diff-567a729df5a343260e81707d6c050fb07cfd16878914491a892f24e866f3161aR14-R15) to get more details

hennevogel (changes requested):

hennevogel (approved):

PR #14461, opened 2023-06-07
Modify footer link to apidocs

saraycp:
After recent changes, the URL was pointing to `apidocs-old`.

Screenshot from 2023-06-07 12-37-59

Now that the new API doc is rolled out, it should point to `apidocs`.


src/api/app/views/layouts/webui/_footer.html.haml
```diff
@@ -21,7 +21,7 @@
     %ul
       %li= link_to('OpenBuildService.org', 'https://openbuildservice.org/')
       %li= link_to('Documentation', 'https://openbuildservice.org/help/manuals/')
-      %li= link_to('API Documentation', apidocs_index_path)
+      %li= link_to('API Documentation', '/apidocs/')
```
hennevogel, line 24: This will break if OBS is served with some prefix like we do in the review apps with `RAILS_RELATIVE_URL_ROOT` for instance.

hennevogel (changes requested):

src/api/app/views/layouts/webui/_footer.html.haml
saraycp: I replaced it with the absolute path. Is that better?

saraycp: @hellcp-work  please have another look.

PR #14464, opened 2023-06-07
Document GET POST /status_reports/* endpoints

hellcp-work:
Most of the implementation happens in the first commit, the rest are copies for the most part.

To verify:
- Use GET endpoint to verify they are there

### DIFF
--- a/src/api/app/controllers/status/checks_controller.rb
+++ b/src/api/app/controllers/status/checks_controller.rb
@@ -7,7 +7,7 @@ class Status::ChecksController < ApplicationController
 
   # POST /status_reports/published/:project_name/:repository_name/reports/:uuid
   # POST /status_reports/built/:project_name/:repository_name/:arch/reports/:uuid
-  # PUT /status_reports/requests/:number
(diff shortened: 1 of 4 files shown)


src/api/public/apidocs/paths/status_reports_built_project_name_repository_name_architecture_name_reports_build_id.yaml
```diff
@@ -0,0 +1,75 @@
...
+    '401':
+      $ref: '../components/responses/unauthorized.yaml'
+    '404':
+      $ref: '../components/responses/not_found.yaml'
+  tags:
+    - Status Reports
+
+post:
+  summary: Submit a check to a built project repository's status report
+  description: Submits a new or modified check to the status report for the built project repository
```
eduardoj, line 39:
```suggestion
  description: Create or update a check to the status report for the built project repository
```
"Submits a new or modified check..." doesn't sound to me accurate enough. What do you think?

Same for the two other cases (`published/` and `requests/`).

eduardoj (approved):

PR #14470, opened 2023-06-08
Document GET POST required checks api endpoints

hellcp-work:
This is documentation related to required checks part of the status reports api. The first commit has the most content, the rest are copying what is done there across other endpoints

To verify:
- Use GET endpoint to verify they are there


saraycp (approved):
LGTM

PR #14473, opened 2023-06-08
[backend] avoid projectsmcsync on single packages

adrianschroeter:


mlschroe:
This commit does not make sense.

PR #14474, opened 2023-06-08
Document DELETE required checks api endpoints 

hellcp-work:
This is a continuation of #14470 and so it depends on it. The first commit (which is the 4th after the previous PR's commits) has most of the content, the remaining commits are copies with most of the same content.

To verify:
- Use DELETE endpoint to delete checks
- Use GET endpoint to verify they aren't there


saraycp:
Reviewing the code of this endpoint I found a bug.

The DELETE endpoints should return an error when we pass a `check_name` that doesn't exist, however, it's always returning a 200 OK response. That's happening because [this code](https://github.com/openSUSE/open-build-service/blob/744cac47272e8ef24e06581ae506ad0c771ee4e5/src/api/app/controllers/status/required_checks_controller.rb#L97) is wrong.

I opened this issue https://github.com/openSUSE/open-build-service/issues/14475.
Once the bug is fixed, we can add another response to the docu. Meanwhile, this PR is ok.

PR #14493, opened 2023-06-13
Apidocs swagger initially collapsed

ncounter:
According to official documentation, https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/, this PR sets the parameter to initially loads the index page of the apidocs endpoints all collapsed.

The reason behind it: the index list is already long, and it separates concepts and topics by "chapters". Having them exploded generates instead a bit of fuss that is somehow useless. The user comes to the apidocs index to look for a topic and its endpoints, not to browse them all.

Note: it is important to notice that the topic name is also exactly the first slice of the endpoint name (i.e.: **Staging** --> "/**staging**/{project_name}/excluded_requests"). The user can always reach the topic by using the browser search function with that text slice and then expand it.

Additionally: now that the loading time is fast with the latest changes (this will be noticed in production only btw), the page can load the index and once the user expand a topic, it will display immediately the content of that topic only, without taking the focus away with all the other topics content.

You can check it at https://obs-reviewlab.opensuse.org/ncounter-apidocs-collapsed/apidocs/index.html

### DIFF
--- a/src/api/public/apidocs/swagger-initializer.js
+++ b/src/api/public/apidocs/swagger-initializer.js
@@ -15,6 +15,7 @@ window.onload = function() {
       SwaggerUIBundle.plugins.DownloadUrl
     ],
     showExplorer: false,
+    docExpansion : "none",
   });
 
   //</editor-fold>


eduardoj:
A couple of points here:
- This change prevents from finding an specific endpoint using the browser search. This is something users of the previous version of the documentation would miss. Users could like to search by `/status/` or by `/status_reports`. With this change, it would not be possible.
- There is no possibility yet to collapse or to expand all sections with a single click in a button, for example.

Basically, this change would make previous users of the documentation harder to adapt to the Swagger-UI interface, and new users harder to find and specific endpoint. I prefer not to introduce this change, at least until we receive more feedback from the users in this direction.

PR #14509, opened 2023-06-15
Handle config file URL or path easily from workflow run

saraycp:
We are extending the WorkflowRun model to make it easier to query by relevant data that helps us debug failed integrations.

Two of the relevant values are stored in  `Token#workflow_configuration_(url|path)`. As we want to know the values stored on those fields when the workflow ran, I added new columns to the `workflow_runs` table.

In the UI, I added a new tab, `Workflow configuration`, where we are going to display all the information related to the workflow configuration file: URL, path, content etc.

Screenshot 2023-06-15 at 16-04-37 Open Build Service

### DIFF
--- a/src/api/app/components/workflow_run_detail_component.rb
+++ b/src/api/app/components/workflow_run_detail_component.rb
@@ -26,14 +26,11 @@ class WorkflowRunDetailComponent < ApplicationComponent
 
   # Old workflow run entries didn't store the configuration-related information
   def workflow_configuration_data
-    configuration_source = [workflow_run.workflow_configuration_url,
-                            workflow_run.workflow_configuration_path].filter_map(&:presence).first
-
-    return content_tag(:p, 'This information is not available.') unless configuration_source
+    return content_tag(:p, 'This information is not available.') unless workflow_run.configuration_source
 
     content_tag(:h5, "Workflow Configuration File #{workflow_run.workflow_configuration_url.present? ? 'URL' : 'Path'}").concat(
       content_tag(:pre,
-                  configuration_source,
+                  workflow_run.configuration_source,
                   class: 'border p-2')
     )
   end
--- a/src/api/app/components/workflow_run_detail_component.rb
+++ b/src/api/app/components/workflow_run_detail_component.rb
@@ -1,11 +1,12 @@
 class WorkflowRunDetailComponent < ApplicationComponent
-  attr_reader :id, :request_headers, :pretty_request_payload,
+  attr_reader :id, :workflow_run, :request_headers, :pretty_request_payload,
               :response_url, :response_body, :artifacts,
               :scm_vendor, :status_reports
 
   def initialize(workflow_run:)
     super
     @id = workflow_run.id
+    @workflow_run = workflow_run
     @request_headers = workflow_run.request_headers
     @pretty_request_payload = parse_payload(workflow_run)
     @response_url = workflow_run.response_url
@@ -21,5 +22,19 @@ class WorkflowRunDetailComponent < ApplicationComponent
     JSON.pretty_generate(JSON.parse(workflow_run.request_payload))
   rescue JSON::ParserError
     workflow_run.request_payload
+  end
+
+  # Old workflow run entries didn't store the configuration-related information
+  def workflow_configuration_data
+    configuration_source = [workflow_run.workflow_configuration_url,
+                            workflow_run.workflow_configuration_path].filter_map(&:presence).first
+
+    return content_tag(:p, 'This information is not available.') unless configuration_source
+
+    content_tag(:h5, "Workflow Configuration File #{workflow_run.workflow_configuration_url.present? ? 'URL' : 'Path'}").concat(
+      content_tag(:pre,
+                  configuration_source,
+                  class: 'border p-2')
+    )
   end
 end
--- a/src/api/app/models/workflow_run.rb
+++ b/src/api/app/models/workflow_run.rb
@@ -125,6 +125,10 @@ class WorkflowRun < ApplicationRecord
     scm_status_reports.last&.response_body
   end
 
+  def configuration_source
+    [workflow_configuration_url, workflow_configuration_path].filter_map(&:presence).first
+  end
+
   private
 
   def parsed_request_headers
--- a/src/api/app/models/workflow_run.rb
+++ b/src/api/app/models/workflow_run.rb
@@ -23,7 +23,7 @@ class WorkflowRun < ApplicationRecord
     :state, :status_options
   ].freeze
 
-  validates :response_url, length: { maximum: 255 }
+  validates :response_url, :workflow_configuration_path, :workflow_configuration_url, length: { maximum: 255 }
   validates :request_headers, :status, presence: true
 
   belongs_to :token, class_name: 'Token::Workflow', optional: true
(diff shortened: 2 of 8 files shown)


src/api/app/models/workflow_run.rb
```diff
@@ -31,6 +31,9 @@ class WorkflowRun < ApplicationRecord
   has_many :scm_status_reports, class_name: 'SCMStatusReport', dependent: :destroy
   has_many :event_subscriptions, dependent: :destroy
 
+  delegate :workflow_configuration_path, to: :token
+  delegate :workflow_configuration_url, to: :token
```
hennevogel, line 35: A `WorkflowRun` is happening at a specific time. Tokens are always *now*.

hennevogel (changes requested):

src/api/app/models/workflow_run.rb
saraycp: True! I have to save the value those fields had when the workflow ran. Changing...

src/api/db/migrate/20230615111252_add_colums_to_workflow_run.rb
```diff
@@ -0,0 +1,6 @@
+class AddColumsToWorkflowRun < ActiveRecord::Migration[7.0]
```
saraycp, line 1: FYI: adding a column is not considered "unsafe" by `strong_migrations`.

src/api/app/components/workflow_run_detail_component.rb
```diff
@@ -22,4 +23,21 @@ def parse_payload(workflow_run)
   rescue JSON::ParserError
     workflow_run.request_payload
   end
+
+  def workflow_configuration_available?
+    workflow_run.workflow_configuration_url || workflow_run.workflow_configuration_path
+  end
+
+  # Old workflow run entries didn't store the configuration-related information
+  def workflow_configuration_data
```
krauselukas, line 32: @saraycp why not doing this in the `haml` file of the view component?

krauselukas: IMO it would be nice to read

saraycp: TBH, because haml-lint complained about every change I tried.

Do you remember the error you had some time ago? It happened to me again.

```
rails dev:lint:all                                                                                                                                                             
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem                                                                                                                                                                                                                                                      
Running haml-lint in src/api...                                                                                                                                                                                    
/usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:606:in `dump': can't dump hash with default proc (TypeError)                                                                                       
        from /usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:606:in `process_incoming_jobs'                                                                                                        
        from /usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:577:in `block in worker'                                                                                                              
        from /usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:568:in `fork'                                                                                                                         
        from /usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:568:in `worker'                                                                                                                                                                                                                                 
...                                                                               
        from /usr/lib64/ruby/gems/3.1.0/gems/railties-7.0.4.3/lib/rails/commands/rake/rake_command.rb:18:in `perform'                                                                         
        from /usr/lib64/ruby/gems/3.1.0/gems/railties-7.0.4.3/lib/rails/command.rb:51:in `invoke'                                                                                             
        from /usr/lib64/ruby/gems/3.1.0/gems/railties-7.0.4.3/lib/rails/commands.rb:18:in `<top (required)>'                                                                                  
        from bin/rails:4:in `require'                                                          
        from bin/rails:4:in `<main>'                                                           
marshal data too short                                                                         
<internal:marshal>:34:in `load'                                                                
/usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:81:in `work'                   
/usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:522:in `block (4 levels) in work_in_processes'                                                                                
/usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:637:in `with_instrumentation'  
/usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:521:in `block (3 levels) in work_in_processes'                                                                                
/usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:509:in `loop'                  
/usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:509:in `block (2 levels) in work_in_processes'                                                                                
/usr/lib64/ruby/gems/3.1.0/gems/parallel-1.23.0/lib/parallel.rb:219:in `block (4 levels) in in_threads' 
```

saraycp: Moreover, writing it in Haml implies a lot of embedded Ruby code in the view (`- if`, `- else`, `= <ruby code>`...).
Something that ViewComponents are meant to avoid: https://viewcomponent.org/#why-use-viewcomponents

danidoni (approved):

src/api/app/components/workflow_run_detail_component.rb
```diff
@@ -22,4 +23,18 @@ def parse_payload(workflow_run)
   rescue JSON::ParserError
     workflow_run.request_payload
   end
+
+  # Old workflow run entries didn't store the configuration-related information
+  def workflow_configuration_data
+    configuration_source = workflow_run.workflow_configuration_url if workflow_run.workflow_configuration_url.present?
+    configuration_source ||= workflow_run.workflow_configuration_path if workflow_run.workflow_configuration_path.present?
```
saraycp, line 30: Sometimes the value is an empty string (""), so I need to check the presence and not the existence of a value, as
` "".present? => false`.

src/api/app/components/workflow_run_detail_component.rb
```diff
@@ -1,11 +1,12 @@
 class WorkflowRunDetailComponent < ApplicationComponent
-  attr_reader :id, :request_headers, :pretty_request_payload,
+  attr_reader :id, :workflow_run, :request_headers, :pretty_request_payload,
               :response_url, :response_body, :artifacts,
               :scm_vendor, :status_reports
 
   def initialize(workflow_run:)
     super
     @id = workflow_run.id
+    @workflow_run = workflow_run
```
saraycp, line 9: Bring back the `workflow_run` attribute that I removed by mistake.

src/api/app/components/workflow_run_detail_component.rb
danidoni: You can also use
```
[workflow_run.workflow_configuration_url, workflow_run.workflow_configuration_path].presence.compact.first
```

saraycp:
> @saraycp what about backfilling the data?

@hennevogel if we introduce these two fields now, we don't have a way to know which values they had when the WorkflowRun was created.

I can only backfill them with the current value in `Token#workflow_configuration_(url|path) `, which might not be realistic.

src/api/app/components/workflow_run_detail_component.rb
saraycp: I assume this is what you tried to suggest:

```
[workflow_run.workflow_configuration_url, workflow_run.workflow_configuration_path].map(&:presence).compact.first
```

As your code could return an empty array as a result, because:

```
[workflow_run.workflow_configuration_url, workflow_run.workflow_configuration_path].presence
=> ["", ".obs/workflows.yml"]
```

saraycp: Pushed the one with the `map`.

danidoni: Nice

src/api/app/components/workflow_run_detail_component.rb
```diff
@@ -22,4 +23,18 @@ def parse_payload(workflow_run)
   rescue JSON::ParserError
     workflow_run.request_payload
   end
+
+  # Old workflow run entries didn't store the configuration-related information
+  def workflow_configuration_data
+    configuration_source = [workflow_run.workflow_configuration_url,
+                            workflow_run.workflow_configuration_path].filter_map(&:presence).first
```
danidoni, line 30: Also, this part could be moved into `WorkflowRun`.

hennevogel: pushed

danidoni (approved):

PR #14522, opened 2023-06-19
Allow versioning for SCM/CI workflow features and add support for event aliases (v.1.1)

krauselukas:
This PR introduces a small implementation to allow versioning for SCM/CI workflow features and changes. The version scheme I would suggest would be `major.minor`. Major versions for non-backwards compatible features and changes. Minor versions for backwards compatible features and changes (non breaking features). For now I would not enforce to specify a version and would fallback to the latest minor version.

The version is specified on the top level and matches the whole `workflow.yml` file against a specific version:
```
version: '1.1'
workflow:
  steps:                                                                                                                                                                                                                                           
    - trigger_services:                                                                                                                                                                                                                            
        project: home:krauselukas
        package: hello_world
  filters:
    event: merge_request
```


**Documentation will happen as follow up PR**

### DIFF
--- a/src/api/app/models/concerns/workflow_version_matcher.rb
+++ b/src/api/app/models/concerns/workflow_version_matcher.rb
@@ -1,0 +1,16 @@
+module WorkflowVersionMatcher
+  # In case no version is specified in the workflow yaml, we fallback
+  # to the current highest minor version, since we don't introduce
+  # breaking changes with those
+  FALLBACK_VERSION = '1.1'.freeze
+  FEATURES_FOR_VERSION = { '1.1': ['event_aliases'] }.freeze
+
+  def feature_available_for_workflow_version?(workflow_version:, feature_name:)
+    workflow_version = FALLBACK_VERSION if workflow_version.blank?
+
+    FEATURES_FOR_VERSION.each do |feature_version, features|
+      return true if Gem::Version.new(workflow_version) >= Gem::Version.new(feature_version) && features.include?(feature_name)
+    end
+    false
+  end
+end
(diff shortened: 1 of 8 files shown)


krauselukas:
I use the `Gem::Version` class for easy version number comparison, its part of the ruby stdlib.

krauselukas:
[Introduce skeleton for versioning (L)](https://trello.com/c/pcfSkttS/2281-introduce-skeleton-for-versioning-l)

krauselukas:
[Allow GitLab terms like merge_request for event filters (M)](https://trello.com/c/6sPGD85I/2282-allow-gitlab-terms-like-mergerequest-for-event-filters-m)

danidoni:
I would go for the second way of specifying versions... The first one could imply that you could add different versioning to another workflow in the same file...

And I would prevent from anyone adding a workflow named `version`... We can error if someone tries that returning something like `version is a reserved word, please use another name for your workflow name`.

krauselukas:
> I would go for the second way of specifying versions... The first one could imply that you could add different versioning to another workflow in the same file...
> 
> And I would prevent from anyone adding a workflow named `version`... We can error if someone tries that returning something like `version is a reserved word, please use another name for your workflow name`.
> 
> I agree on everything else!

Alright adapted the PR to go with the second option.

src/api/app/models/concerns/workflow_version_matcher.rb
```diff
@@ -0,0 +1,16 @@
+module WorkflowVersionMatcher
+  # Incase no version is specified in the workflow yaml, we fallback
```
danidoni, line 2:
```suggestion
  # In case no version is specified in the workflow yaml, we fallback
```

krauselukas: Done

hennevogel:
> This breaks workflows that are named version (very unlikely) 

Famous last words...

krauselukas:
> > This breaks workflows that are named version (very unlikely)
> 
> Famous last words...

let me see if I can check if a workflow is meant or a version, should be able to check for that

krauselukas:
> > This breaks workflows that are named version (very unlikely)
> 
> Famous last words...

@hennevogel alright, I included a check for that to avoid breaking workflows in that case

danidoni (approved):

PR #14523, opened 2023-06-20
Expand workflow run with headers info

saraycp:
We are extending the WorkflowRun model to make it easier to query by relevant data that helps us debug failed integrations.

The SCM vendor and the hook event are key for us to query. They are now stored in the `workflow_runs` table.

This PR includes a data migration to backfill those values taking the data from `WorkflowRun#request_headers`. 

Nothing changed in the UI as we were displaying that information already:

Workflow-Run-23-Open-Build-Service

Based on #14509

### DIFF
--- a/src/api/app/components/workflow_run_detail_component.rb
+++ b/src/api/app/components/workflow_run_detail_component.rb
@@ -12,7 +12,7 @@ class WorkflowRunDetailComponent < ApplicationComponent
     @response_url = workflow_run.response_url
     @response_body = workflow_run.response_body
     @artifacts = workflow_run.artifacts # collection of WorkflowArtifactsPerStep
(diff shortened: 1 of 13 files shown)


src/api/app/components/workflow_run_detail_component.rb
```diff
@@ -22,4 +23,18 @@ def parse_payload(workflow_run)
   rescue JSON::ParserError
     workflow_run.request_payload
   end
+
+  # Old workflow run entries didn't store the configuration-related information
+  def workflow_configuration_data
+    configuration_source = [workflow_run.workflow_configuration_url,
+                            workflow_run.workflow_configuration_path].filter_map(&:presence).first
```
danidoni, line 30: Also, this part could be moved into `WorkflowRun`.

hennevogel (approved):

PR #14526, opened 2023-06-21
[dist] obsapisetup: create openssl key with 4096

M0ses:
This is required since opensuse 15.4. Otherwise apache will fail with a config error


bmwiedemann (approved):

M0ses:
merging because circle ci failures cannot be caused be this change

PR #14530, opened 2023-06-21
[backend] add a way to specify related projects in BSConfig

mlschroe:
We currently only use the project namespace. Sometimes we must define two distinct projects to be related.


hennevogel:
Hm, how can the other parts of the application know about this?

PR #14537, opened 2023-06-22
Drop specs for the WorkflowRun components

hennevogel:
1007 lines of spec for code that has not a single line of logic that goes beyond a simple if or loop.

The only thing this was is a view spec, which we don't do...

Check it out yourself: 

```
cat app/components/workflow_run_*.html*|less
cat app/components/workflow_run_*.rb|less
```


danidoni (approved):

danidoni:
The total coverage is declining, but that's expected for this PR. Merging

PR #14541, opened 2023-06-23
Introduces workflow run payload fields

hennevogel:
Extend workflow_run DB table to store data from the payload. Extracted from #14527 so we can run this online.

1. We create the fields in the database
2. Getters are still using the `request_payload` field to fetch the data
3. Setters are the default from AR
4. Data migration to backfill data from payload field
5. Getters will [be removed](https://github.com/openSUSE/open-build-service/pull/14527/commits/114aaccc8b9ab2d4333af88fbbc8ed21e3f96629) once the second PR is merged


hennevogel:
we will have to life with some duplicated code and a controller concern that is only used in the data migration until the second PR is merged...

rubhanazeem (approved):

PR #14549, opened 2023-06-25
feat: workflow build job add pending status support

hudeng-go:
When the repositories configuration is clearly defined in the workflow, the job list can be determined, so the pending status of the job can be added.

### DIFF
--- a/src/api/app/models/token/workflow.rb
+++ b/src/api/app/models/token/workflow.rb
@@ -49,6 +49,9 @@ class Token::Workflow < Token
       workflow.call
     end
     SCMStatusReporter.new(@scm_webhook.payload, @scm_webhook.payload, scm_token, workflow_run, 'success', initial_report: true).call
+    # Report build job pending when evnt is pr or mr
+    report_job_pending_status(workflow_run)
+
     # Always returning validation errors to report them back to the SCM in order to help users debug their workflows
     validation_errors
   rescue Octokit::Unauthorized, Gitlab::Error::Unauthorized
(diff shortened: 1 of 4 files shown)


hennevogel (changes requested):
I'm afraid it's not that simple... You would also need to figure out which packages will build, including wild things like multibuild, linked builds etc. As later we only report for packages. 

Once you figured that out you would need to take [`set_flags`](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html#sec.obs.obs_scm_ci_workflow_integration.obs_workflows.steps.set_flags) into account too. 

I guess you can see why we are just sending a general initial report...

hudeng-go:
> I'm afraid it's not that simple... You would also need to figure out which packages will build, including wild things like multibuild, linked builds etc. As later we only report for packages.
> 
> Once you figured that out you would need to take [`set_flags`](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html#sec.obs.obs_scm_ci_workflow_integration.obs_workflows.steps.set_flags) into account too.
> 
> I guess you can see why we are just sending a general initial report...

Yes, but jobs that include repositories configuration can clearly add a pending status,suitable for a specific scene in complex scenarios.

hennevogel:
> Yes, but jobs that include repositories configuration can clearly add a pending status,suitable for a specific scene in complex scenarios.

A pending status that no on will update later on? What use would that be?

hudeng-go:
> > Yes, but jobs that include repositories configuration can clearly add a pending status,suitable for a specific scene in complex scenarios.
> 
> A pending status that no on will update later on? What use would that be?

update by job status later,https://github.com/openSUSE/open-build-service/blob/master/src/api/app/jobs/report_to_scm_job.rb#L10

hennevogel:
> update by job status later

Not if there never will be a build for a package that is named like your repository it won't...

hudeng-go:
> > update by job status later
> 
> Not if there never will be a build for a package that is named like your repository it won't...

If so, modify workflows. yml to remove this build, It can be understood that build jobs are controlled by workflows. yml, which basically meets the requirements in the PR usage scenario.

hennevogel:
> If so, modify workflows.yml to remove this build

I wish it would be that easy and everybody could just adopt to the most simplest case. But sadly that is not what people will be able to do.

hennevogel:
no feedback closing

PR #14553, opened 2023-06-26
Introduces new ContainerPublished event

hennevogel:
An event if a container image is published to a registry.

Jira feature [OBS-243](https://jira.suse.com/browse/OBS-243)


hennevogel:
@mlschroe guess we need to coordinate deployments

eduardoj (approved):

ncounter (approved):

PR #14554, opened 2023-06-26
Save the workflow yaml content into the workflow run

danidoni:
Having the actual Workflow configuration stored may help debug what happened when a Workflow Run fails.

### DIFF
--- a/src/api/app/services/workflows/yaml_to_workflows_service.rb
+++ b/src/api/app/services/workflows/yaml_to_workflows_service.rb
@@ -19,21 +19,22 @@ module Workflows
     private
 
     def create_workflows
+      @workflow_run.update(workflow_configuration: File.read(@yaml_file))
       begin
-        parsed_workflows_yaml = YAML.safe_load(parse_workflows_file(@yaml_file))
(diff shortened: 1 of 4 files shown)


rubhanazeem (approved):

src/api/app/services/workflows/yaml_to_workflows_service.rb
```diff
@@ -33,7 +35,7 @@ def create_workflows
       end
     end
 
-    def parse_workflows_file(file_path)
+    def parse_workflows_content(workflows_file_content)
```
hennevogel, line 38: You now use *4* different names for the same thing

- workflows_file_content
- workflow_configuration_yaml
- workflows_content
- (parsed_)workflows_yaml

Can we reduce this to 1 please?

hennevogel (changes requested):

src/api/app/services/workflows/yaml_to_workflows_service.rb
danidoni: Yes, of course

danidoni: DoneDone

PR #14555, opened 2023-06-26
Save more header content to WorkflowRun

hennevogel:
event_uuid: Set by all vendors, helps to identify the exact webhook event that caused this workflow to run

webhook_id: Set only be github so far, help to identify the webhook that caused this workflow to run


src/api/db/data/20230626141554_backfill_workflow_run_payload_data.rb
```diff
@@ -0,0 +1,24 @@
...
+        workflow_run.webhook_id = extract_webhook_id
+        workflow_run.save!
+      end
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+
```
eduardoj, line 20: Should this data migration be irreversible? We could empty those fields, at least.

hennevogel:
will follow up later...

PR #14565, opened 2023-06-29
[api] fix delivering of delete package meta data

adrianschroeter:
api blocked requests for deleted packages always.

### DIFF
--- a/src/api/public/apidocs/paths/source_project_name_package_name_meta.yaml
+++ b/src/api/public/apidocs/paths/source_project_name_package_name_meta.yaml
@@ -7,6 +7,13 @@ get:
     - $ref: '../components/parameters/project_name.yaml'
     - $ref: '../components/parameters/package_name.yaml'
     - in: query
+      name: deleted
+      schema:
+        type: string
+        enum:
+          - null
+      description: Set to retrieve the package metadata of a deleted package.
+    - in: query
       name: meta
       schema:
         type: string
--- a/src/api/app/controllers/source_project_package_meta_controller.rb
+++ b/src/api/app/controllers/source_project_package_meta_controller.rb
@@ -20,13 +20,16 @@ class SourceProjectPackageMetaController < SourceController
 
   # GET /source/:project/:package/_meta
   def show
-    pack = Package.get_by_project_and_name(@project_name, @package_name, use_source: false)
+    if params.key?(:deleted)
+      validate_read_access_of_deleted_package(@project_name, @package_name)
+    else
+      pack = Package.get_by_project_and_name(@project_name, @package_name, use_source: false)
+    end
 
-    if params.key?(:meta) || params.key?(:rev) || params.key?(:view) || pack.nil?
-      # check if this comes from a remote project, also true for _project package
-      # or if meta is specified we need to fetch the meta from the backend
+    # Let the backend answer for deleted or remote packages. For specific revisions or the blame view. Or if the meta parameter is used.
+    if params.key?(:deleted) || params.key?(:meta) || params.key?(:rev) || params.key?(:view) || pack.nil?
       path = request.path_info
-      path += build_query_from_hash(params, %i[meta rev view])
+      path += build_query_from_hash(params, %i[deleted meta rev view])
       pass_to_backend(path)
       return
     end


src/api/public/apidocs/paths/source_project_name_package_name_meta.yaml
```diff
@@ -6,6 +6,12 @@ get:
   parameters:
     - $ref: '../components/parameters/project_name.yaml'
     - $ref: '../components/parameters/package_name.yaml'
+    - in: query
+      name: deleted
+      schema:
+        type: string
+      description: Set to `1` to retrieve the package metadata of a deleted package.
+      example: 1
```
adrianschroeter, line 14: no, it is a boolean, just the existens is enough

eduardoj: Checking locally with `&deleted=a` returns a `not boolean` error:
```
> curl -i -u Admin:XXXXXXXX -H 'Content-Type: application/xml; charset=utf-8' -X GET "http://localhost:3000/source/home:Admin/hey/_meta?deleted=a"
<status code="400">
  <summary>not boolean</summary>
</status>
>
```
The backend also defines the `deleted` parameter as bool: https://github.com/openSUSE/open-build-service/blob/d5f0c54b8ad3163115cf8c986302b482e4934dc2/src/backend/bs_srcserver#L7675

eduardoj: I tested locally with a deleted `hey` package.

These are some other tests:

- With `deleted=1`:
   ```
   > curl -i -u Admin:XXXXXXXX -H 'Content-Type: application/xml; charset=utf-8' -X GET "http://localhost:3000/source/home:Admin/hey/_meta?deleted=1"
   <package name="hey" project="home:Admin">
     <title></title>
     <description></description>
   </package>
   ```
- With `deleted=0`:
   ```
   > curl -i -u Admin:XXXXXXXX -H 'Content-Type: application/xml; charset=utf-8' -X GET "http://localhost:3000/source/home:Admin/hey/_meta?deleted=0"
   <status code="404">
     <summary>package 'hey' does not exist in project 'home:Admin'</summary>
     <details>404 package 'hey' does not exist in project 'home:Admin'</details>
   </status>
   ```
- Without the `deleted` parameter:
   ```
   > curl -i -u Admin:XXXXXXXX -H 'Content-Type: application/xml; charset=utf-8' -X GET "http://localhost:3000/source/home:Admin/hey/_meta"
   <status code="unknown_package">
     <summary>Package not found: home:Admin/hey</summary>
   </status>
   ```

eduardoj: And lastly, if we are talking about `type:` being `string` and not `boolean`, I chose `string` because the `boolean` type is only for representing the values `true` and `false`.

See the reference documentation: https://swagger.io/docs/specification/data-models/data-types/#boolean

eduardoj: You can see in the review-app how it will look like, [here](https://obs-reviewlab.opensuse.org/adrianschroeter-fix_deleted_meta/apidocs/index.html#/Sources%20-%20Packages/get_source__project_name___package_name___meta).

adrianschroeter: http://localhost:3000/source/home:Admin/hey/_meta?deleted

is supposed to be used and should work.

adrianschroeter: deleted=0 is not really defined what will happen though, please do not mention this anywhere.

(yes, some need it as hacks to workaround bugs in some cases, but really nothing what should be documented)

eduardoj: > http://localhost:3000/source/home:Admin/hey/_meta?deleted
> 
> is supposed to be used and should work.

Confirmed. - With `deleted`:
   ```
   > curl -i -u Admin:XXXXXXXX -H 'Content-Type: application/xml; charset=utf-8' -X GET "http://localhost:3000/source/home:Admin/hey/_meta?deleted"
   <package name="hey" project="home:Admin">
     <title></title>
     <description></description>
   </package>
   ```

eduardoj: > deleted=0 is not really defined what will happen though, please do not mention this anywhere.

I removed the `0` option from the `enum`. Now it is not mentioned.

> (yes, some need it as hacks to workaround bugs in some cases, but really nothing what should be documented)

I prefer to document all the endpoints with all their parameters in the reference documentation. We can always decide what to include in the long form API documentation, yet to be written.

I have modified the documentation commit I added to adapt it to the requirements of this boolean `deleted` parameter. Please change the commit as you wish, or even delete it if you think it is better that way.

adrianschroeter:
what about with this solution?

(delete=1 is not something what we should recommend in any way, we may even mark it as not working in future maybe. After all it was just a workaround, because the rails framework once was not able to deal with the simple backend api)

src/api/public/apidocs/paths/source_project_name_package_name_meta.yaml
```diff
@@ -6,6 +6,13 @@ get:
   parameters:
     - $ref: '../components/parameters/project_name.yaml'
     - $ref: '../components/parameters/package_name.yaml'
+    - in: query
+      name: deleted
+      schema:
+        type: string
+        enum:
+          - null
```
eduardoj, line 14:
```suggestion
        enum:
          - null
```
This translates into passing the string `'null'` instead of passing the parameter without a value. I would prefer not to define the `enum` in this case.

eduardoj:
```suggestion
      allowEmptyValue: true
```
Forget my previous proposal: adding `allowEmptyValue: true` will make it. :sweat_smile:

Peek 2023-07-03 16-12-swagger-ui-deleted-parameter-allow-empty-value

eduardoj:
> what about with this solution?
> 
> (delete=1 is not something what we should recommend in any way, we may even mark it as not working in future maybe. After all it was just a workaround, because the rails framework once was not able to deal with the simple backend api)

Fine for me. :+1:

Just see [this comment](https://github.com/openSUSE/open-build-service/pull/14565#discussion_r1250952503) to document properly a parameter which allows empty values.

hennevogel (approved):

PR #14575, opened 2023-06-29
Update the dist spec to the new file upload workflow

hellcp-work:
The way to upload files changed a while back to support uploading multiple files. This is meant to update the dist specs to catch up to that change.

### DIFF
--- a/dist/t/spec/features/0040_package_spec.rb
+++ b/dist/t/spec/features/0040_package_spec.rb
@@ -26,10 +26,8 @@ RSpec.describe 'Package', type: :feature do
       click_link('Your Home Project')
     end
     click_link('hello_world')
-    click_link('Add File')
-    attach_file('file', File.expand_path('../fixtures/hello_world.spec', __dir__), make_visible: true)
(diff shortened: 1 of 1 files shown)


dist/t/spec/features/0040_package_spec.rb
```diff
@@ -26,10 +26,8 @@
       click_link('Your Home Project')
     end
     click_link('hello_world')
-    click_link('Add File')
-    attach_file('file', File.expand_path('../fixtures/hello_world.spec', __dir__), make_visible: true)
-    click_button('Add File')
-    expect(page).to have_content("The file 'hello_world.spec' has been successfully saved.")
+    attach_file('files', File.expand_path('../fixtures/hello_world.spec', __dir__), make_visible: true)
```
eduardoj, line 29:
```suggestion
    attach_file('files[]', File.expand_path('../fixtures/hello_world.spec', __dir__), make_visible: true)
```
Does it also work without the brackets?

eduardoj (approved):

PR #5870, opened 2018-09-18
Fix xml output  for comments

eduardoj:

PR #14582, opened 2023-06-30
Show diff comments in the API endpoint `GET /comments/request/id`

eduardoj:
After diff comments were introduced within the beta program, users of the API weren't able to list diff comments.

With the changes proposed, these could be examples of responses of calling the `GET /comments/request/id` API endpoint:

1) Example of a response performed in the development environment. The text of the comment is modified for inline comments:
```xml
<comments request="10">
  <comment who="Admin" when="2023-07-03 10:56:52 UTC" id="9">Inline comment for target: 'home:Admin/another_package_with_diff_1688381711', file: 'README.txt', and line: 4:

In this README.txt file I would prefer `# Content` instead of `# New content`.</comment>
  <comment who="Admin" when="2023-07-03 10:57:11 UTC" id="10">Inline comment for target: 'home:Admin/another_package_with_diff_1688381711', file: 'source_package_with_multiple_submit_request_and_diff
_1688381712.spec', and line: 14:

Line 13 of this .spec file... `%build`... ok.</comment>
  <comment who="Admin" when="2023-07-03 10:57:29 UTC" id="11">Well, thats nice.</comment>
  <comment who="Admin" when="2023-07-03 10:57:41 UTC" id="12" parent="11">Sure that's nice!</comment>
</comments>
```

2) Example of a response of a request with an inline comment where the target package is removed. The text of comment is not modified:

```xml
<comments request="17">
  <comment who="Admin" when="2023-07-03 12:25:58 UTC" id="13">Makes things more clear.</comment>
</comments>
```

### DIFF
--- a/src/api/app/components/comment_component.rb
+++ b/src/api/app/components/comment_component.rb
@@ -22,10 +22,24 @@ class CommentComponent < ApplicationComponent
   end
 
   def body
+    return inline_comment_body if comment.commentable.is_a?(BsRequestAction)
+
     comment.body.delete("\u0000")
   end
 
   private
 
   attr_accessor :comment
+
+  def inline_comment_body
+    target = "#{comment.commentable.target_project}/#{comment.commentable.target_package}"
+    file_index, line_number = comment.diff_ref.match(/diff_([0-9]+)_n([0-9]+)/).captures
+    sourcediff = comment.commentable.bs_request.webui_actions(action_id: comment.commentable, diffs: true, cacheonly: 1).first[:sourcediff].first
+
+    return comment.body if sourcediff[:error]
+
+    filename = sourcediff['filenames'][file_index.to_i]
+
+    "Inline comment for target: '#{target}', file: '#{filename}', and line: #{line_number}:\n\n#{comment.body}"
+  end
 end
--- a/src/api/app/controllers/comments_controller.rb
+++ b/src/api/app/controllers/comments_controller.rb
@@ -2,7 +2,9 @@ class CommentsController < ApplicationController
   before_action :find_obj, only: [:index, :create]
 
   def index
-    @comments = @obj.comments.includes(:user).order(:id)
+    comments = @obj.comments.includes(:user)
+    comments += @obj.bs_request_actions.flat_map { |a| a.comments.includes(:user) } if @obj.is_a?(BsRequest)
+    @comments = comments.sort_by(&:id)
   end
 
   def create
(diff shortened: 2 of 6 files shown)


src/api/app/components/comment_component.rb
```diff
@@ -17,6 +17,10 @@ def attributes
       attrs[comment.commentable.class.name.downcase.to_sym] = comment.commentable.to_param
       attrs[:project] = comment.commentable.project if comment.commentable.is_a?(Package)
     end
+    if comment.commentable.is_a?(BsRequestAction)
```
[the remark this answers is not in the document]
eduardoj: What do you mean by "break API clients"? We are adding the `request_action_id` and `diff_ref` attributes for `comment` XML elements. Adding attributes to XML elements in a known response shouldn't break API clients.

In what way should we provide that information not to break API clients?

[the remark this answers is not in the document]
hennevogel: I said "possibly". For instance if they expect this to be a static list. 

> In what way should we provide that information not to break API clients?

The question is more, what could they possibly do with this?

src/api/app/components/comment_component.rb
```diff
@@ -22,10 +22,21 @@ def attributes
...
 
   attr_accessor :comment
+
+  def inline_comment_body
+    target = "#{comment.commentable.target_project}/#{comment.commentable.target_package}"
+    file_index, line_number = comment.diff_ref.match(/diff_([0-9]+)_n([0-9]+)/).captures
+    diffs = comment.commentable.bs_request.webui_actions(action_id: comment.commentable, diffs: true, cacheonly: 1)
+    filename = diffs.first[:sourcediff].first['filenames'][file_index.to_i]
+
+    "Inline comment for target: '#{target}', file: '#{filename}', and line: #{line_number}:\n\n#{comment.body}"
```
hennevogel, line 40: This creates a `.body` -> `.inline_comment_body`  loop no?

hennevogel: True. I guess should should remove the unicode null byte then too? I have no idea why we do this though...

eduardoj:
The target package could be deleted. I modified the code to take this case into account.

src/api/app/components/comment_component.rb
eduardoj: > True. I guess should should remove the unicode null byte then too? I have no idea why we do this though...

I remember that one: https://github.com/openSUSE/open-build-service/pull/5870

hennevogel: 2018, those where the days!

eduardoj:
I added controller specs. @hennevogel please take another look.

hennevogel (approved):

PR #14593, opened 2023-07-03
[api] avoid services in flight on request creation

adrianschroeter:
We wait to finish here for any running service.

Also avoiding not necessary backend calls when no addrevision mechanic is used.


src/api/app/models/bs_request_action.rb
```diff
@@ -787,18 +787,22 @@ def check_for_expand_errors!(add_revision)
...
-        # fix the revision to the expanded sources at the time of submission
-        self.source_rev = dir['srcmd5']
+        # enforce the revision update
+        add_revision = true
+        source_rev = nil
       end
 
       if add_revision && !source_rev
+        # ensure no changes are in flight when using scripts to create and submit content
+        Backend::Api::Sources::Package.wait_service(source_project, source_package)
```
hennevogel, line 803: Phew that might be a *very* long time right and be error prone right? AFAIR this part mostly runs during creation of the requests which happens in a (manual) transaction. Sounds to me like another likely source of deadlocks...

hennevogel: How about doing this only for `scmsync` packages? I highly doubt that other people submit requests *before* services are finished. I'm not even sure it's possible with UI or API workflows....

dcermak:
Henne Vogelsang ***@***.***> writes:

>
>
>
>>        end
>  
>        if add_revision && !source_rev
> +        # ensure no changes are in flight when using scripts to create and submit content
> +        Backend::Api::Sources::Package.wait_service(source_project, source_package)
>
> How about doing this only for `scmsync` packages? I highly doubt that other people submit requests *before* services are finished. I'm not even sure it's possible with UI or API workflows....

It is most certainly possible via the API. How else would I have managed
to create this SR?

adrianschroeter:
On Montag, 3. Juli 2023, 12:30:47 CEST Henne Vogelsang wrote:
> 
> 
> 
> >        end
>  
>        if add_revision && !source_rev
> +        # ensure no changes are in flight when using scripts to create and submit content
> +        Backend::Api::Sources::Package.wait_service(source_project, source_package)
> 
> How about doing this only for `scmsync` packages? I highly doubt that other people submit requests *before* services are finished. I'm not even sure it's possible with UI or API workflows....

this is not special for scmsync, it is a generic issue for any service package.

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> this is not special for scmsync, it is a generic issue for any service package.

Not from the user interfaces. Both the UI and `osc` don't let you submit packages with running services...

adrianschroeter:
> Not from the user interfaces. Both the UI and `osc` don't let you submit packages with running services...

For one this is not true in all usecases and secondly I see no point to add additional code, just to keep buggy behaviour and not necessary backend calls in some cases. It is a generic issue and should be kept simple therefore IMHO.

hennevogel:
>  It is a generic issue and should be kept simple therefore IMHO.

But surely not at the cost of tons of new deadlocks and the problems they will bring with them right?

PR #14597, opened 2023-07-03
Migrate the remaining database tables and columns to utf8mb4

### LINKED ISSUE #14411: Bad handling of unicode characters in /person?cmd=register
# Issue Description
OBS returns error 500 when `<realname>` or `<login>` contain unicode characters.
`<password>` with unicode characters works fine.

Expected Result
============
OBS either accepts the unicode characters or returns a sensible error other than 500.
Login should probably stick with ascii, while realname should support unicode/utf-8.

How to Reproduce
===============
1. Write test-user.xml:
```
 <unregisteredperson>
  <login>🚀</login>
  <email>unicode@example.com</email>
  <realname>🚀</realname>
  <state>confirmed_user</state>
  <password>Password with unicode characters 🚀🚀🚀</password>
</unregisteredperson>
```
2. Run `osc api -X POST '/person?cmd=register' --file=test-user.xml`
3. See error

hellcp-work:
The current database is a mixture of utf8mb3 and utf8mb4, which happens to bring up some issues with encoding characters from time to time. In order to actually fix this and avoid any encoding issues in the future (until we move to utf1024 or something) we need to change database charset and collation of everything that isn't currently utf8mb4.

These migrations will require downtime.

Fixes #14411

### DIFF
--- a/src/api/lib/xpath_engine.rb
+++ b/src/api/lib/xpath_engine.rb
@@ -348,7 +348,7 @@ class XpathEngine
 
       @joins = ['LEFT JOIN repositories AS release_repositories ON binary_releases.repository_id = release_repositories.id',
                 'LEFT JOIN projects AS release_projects ON release_repositories.db_project_id = release_projects.id',
-                'LEFT join product_media on (product_media.repository_id=release_repositories.id AND product_media.name=binary_releases.medium COLLATE utf8_unicode_ci)',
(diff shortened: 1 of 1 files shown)


eduardoj (approved):

eduardoj (approved):

jengelh:
>until we move to utf1024 or something

(Won't happen. 6 bytes is max in utf8, but Unicode in 2003 declared 21 bits to be _enough for everyone_ and every codepoint, so it's 4 bytes max in practical utf8.)

jengelh:
🤦. Let's leave error correction and cryptography to real cryptographers.

PR #14599, opened 2023-07-04
Add spec test for incident request

bmwiedemann:
This PR adds a feature spec to check the Maintenance Incident Workflow when the "Request Show Redesign" beta flag is on


danidoni:
We're focusing on another approach

PR #14600, opened 2023-07-04
Send metrics about daemon availability

bmwiedemann:
Send metrics about daemon availability via RabbitMQ+telegraf into influxdb+grafana

This patch allows us to do alerting when a required backend service becomes unavailable.

### DIFF
--- a/src/api/app/jobs/worker_measurements_job.rb
+++ b/src/api/app/jobs/worker_measurements_job.rb
@@ -12,6 +12,7 @@ class WorkerMeasurementsJob < ApplicationJob
     send_worker_metrics
     send_job_metrics
     send_scheduler_metrics
+    send_daemon_metrics
   end
 
   private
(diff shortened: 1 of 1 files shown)


src/api/app/jobs/worker_measurements_job.rb
```diff
@@ -50,4 +51,17 @@ def send_scheduler_metrics
...
+    running=0
+    nonrunning=0
+    @workerstatus.xpath('//partition/daemon').each do |daemon|
+      if daemon.attributes['state'].value == 'running'
+        running += 1
+      else
+        nonrunning += 1
+      end
+    end
+    RabbitmqBus.send_to_bus('metrics', "worker_status running=#{running},nonrunning=#{nonrunning}")
```
hennevogel, line 65: Why summarize it in the metric already? Instead of individual measurements. Makes it hard to get more information, like *what* is `nonrunning` for instance.

hennevogel (changes requested):

src/api/app/jobs/worker_measurements_job.rb
bmwiedemann: I rewrote it to collect more details

hennevogel (approved):

PR #14602, opened 2023-07-04
Show diff comments in the Web UI, non-beta

eduardoj:
After diff comments were introduced within the beta program, users not in the beta program weren't able to see diff comments in the web interface.

### Before

Screenshot from 2023-07-05 13-18-45

### After

Screenshot from 2023-07-05 13-17-25

### DIFF
--- a/src/api/app/views/webui/comment/_comment_list.html.haml
+++ b/src/api/app/views/webui/comment/_comment_list.html.haml
@@ -1,3 +1,6 @@
+- if commentable.is_a?(BsRequest)
+  - Comment.without_parent.where(commentable: BsRequestAction.where(bs_request: commentable)).includes(:user).each do |comment|
+    = render partial: 'webui/comment/content', locals: { comment: comment, commentable: commentable, level: 1 }
 - commentable.comments.without_parent.includes(:user).each do |comment|
   = render partial: 'webui/comment/content', locals: { comment: comment, commentable: commentable, level: 1 }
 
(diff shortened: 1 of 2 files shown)


src/api/app/views/webui/comment/_comment_list.html.haml
```diff
@@ -1,3 +1,5 @@
+- commentable.bs_request_actions.flat_map { |a| a.comments.without_parent.includes(:user) }.each do |comment|
```
danidoni, line 1: I believe this line is doing several queries, right? One for the `commentable.bs_request_actions.flat_map { ... }` and another for each of the `a.comments.without_parent.includes(:user)`. Can we rewrite it? I think querying for `Comments.where(...)` will fix that.

eduardoj: I "borrowed" that line from here: https://github.com/openSUSE/open-build-service/blob/e0ffa8e1d498e43a3e5806f30592e97be8fdf380/src/api/app/components/bs_request_activity_timeline_component.rb#L13

And I used it again here: https://github.com/openSUSE/open-build-service/blob/e0ffa8e1d498e43a3e5806f30592e97be8fdf380/src/api/app/controllers/comments_controller.rb#L6

Tests are not passing. After I fix them, I will take a look if it can be improved.

eduardoj: I tried with `Comment.without_parent.where(commentable: BsRequestAction.where(bs_request: commentable))`. It seems to work. Please take a look again.

danidoni: Nice

danidoni (approved):

PR #14607, opened 2023-07-05
Request description as part of the conversation history

ncounter:
In the scope of highlighting more the `BsRequest` description, it is a fact that the description itself is the _initial comment_ of the request, so it is the root from where the history starts.

With this PR the description has its own title and box, and the history timeline starts from this box.
By doing this, the aside column with all the other information (reviewers, maintainers, etc) are aligned with the top of the Conversation tab panel.

# Before
image

# After
image

### DIFF
--- a/src/api/app/views/webui/request/beta_show.html.haml
+++ b/src/api/app/views/webui/request/beta_show.html.haml
@@ -16,91 +16,93 @@
                   actions_count: @supported_actions.count, active_tab: @active_tab }
     .container.p-4
       .row
-        .col-md-12
-          .d-flex.justify-content-between.mb-4
-            #description-text
-              - if @bs_request.description.present?
-                = render partial: 'webui/shared/collapsible_text', locals: { text: @bs_request.description }
-              - else
-                %i No description set
-      .row
-        = render partial: 'webui/request/beta_show_tabs/conversation_aside',
-                locals: { bs_request: @bs_request,
-                  my_open_reviews: @my_open_reviews,
-                  request_reviews: @request_reviews,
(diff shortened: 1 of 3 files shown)


src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -16,91 +16,93 @@
...
-                = link_to('#collapse-superseding', class: 'd-flex align-items-center',
-                    data: { 'bs-toggle': 'collapse' },
-                    aria: { expanded: false, controls: 'collapse-superseding' }) do
-                  %i.timeline-break.expander>
-                  %i.timeline-offset.collapser>
+                %h4.list-group.mb-4.ms-3 Comments & History
+                .timeline-item
+                  .d-inline-flex
+                    %i.fas.fa-lg.fa-code-commit.text-dark
+                    - creator = User.find_by_login(@bs_request.creator) || User.nobody
```
danidoni, line 43: This can be pushed as a `User` method

ncounter: It is true, definitely, it's a good point, but I wouldn't change the business logic here.

Unfortunately the changes comparison reports much more than the real changes, indeed that is something that was there already before. What this PR does is just to move few things around in the UI and decorate the request description with a border box. 

The scope of the PR is to improve the look&feel so I'd leave it for a later refactoring maybe. Do you agree?

danidoni: Yes, a follow-up with that small refactoring would be nice!

danidoni (approved):

PR #14608, opened 2023-07-05
Improve query to retrieve comments for request actions

eduardoj:
Follow up to #14602. See this comment: https://github.com/openSUSE/open-build-service/pull/14602#pullrequestreview-1514045146

### DIFF
--- a/src/api/app/components/bs_request_activity_timeline_component.rb
+++ b/src/api/app/components/bs_request_activity_timeline_component.rb
@@ -10,7 +10,7 @@ class BsRequestActivityTimelineComponent < ApplicationComponent
     super
     @bs_request = bs_request
     @creator = User.find_by_login(bs_request.creator) || User.nobody
-    action_comments = @bs_request.bs_request_actions.flat_map { |a| a.comments.without_parent.includes(:user) }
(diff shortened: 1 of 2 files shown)


src/api/app/components/bs_request_activity_timeline_component.rb
```diff
@@ -10,7 +10,7 @@ def initialize(bs_request:, request_reviews_for_non_staging_projects: [])
     super
     @bs_request = bs_request
     @creator = User.find_by_login(bs_request.creator) || User.nobody
-    action_comments = @bs_request.bs_request_actions.flat_map { |a| a.comments.without_parent.includes(:user) }
+    action_comments = Comment.without_parent.where(commentable: BsRequestAction.where(bs_request: @bs_request)).includes(:user)
```
danidoni, line 13: That could be a scope, right? It would make this very compact

eduardoj: True. Creating this scope would affect this same PR and also #14602. That could be a follow up after both PRs get merged.

danidoni: Alright!

danidoni (approved):

PR #14611, opened 2023-07-06
Fix rendered dates when revision param is present

### LINKED ISSUE #12674: package/show sometimes shows wrong rev

bmwiedemann:
Fix rendered dates when revision param is present but expand param is not

Fixes #12674


hennevogel (changes requested):
https://github.com/openSUSE/open-build-service/blob/master/CONTRIBUTING.md#proper-commit-messages

PR #14620, opened 2023-07-07
Instrument bs_request_actions by type

rubhanazeem:
New tag for bs_request_action type to instrument BsRequestActions per type for request_workflow_redesign beta feature


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -569,4 +571,10 @@ def prepare_request_data
 
     handle_notification
   end
+
+  def set_influxdb_data_request_actions
+    InfluxDB::Rails.current.tags = {
```
hennevogel, line 576: This will overwrite all other tags no?

hennevogel (changes requested):

src/api/app/controllers/webui/request_controller.rb
rubhanazeem: No, it is just adding a new tag

hennevogel: ok, guess I don't remember my own code then ;)

hennevogel (approved):

PR #14621, opened 2023-07-07
Adapt request's view for request action "Set Bugowner"

eduardoj:
## TODO

- [x] Set bugowner for a user
- [ ] Set bugowner for a group


ncounter (approved):
LGTM :+1:

PR #14626, opened 2023-07-10
Fix filter order for set_influxdb_data

rubhanazeem:
In 253fce629160e96a74381230e93e0a0a10d0966c @hennevogel  moved setting influx tags into set_influxdb_data. But in WebuiController the filter was running *before* check_user so User.session was *always* nil.
    
That means for webui requests, both beta and anonymous tags have been wrong since that time. oops... :-(

### DIFF
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -14,10 +14,10 @@ class Webui::WebuiController < ActionController::Base
   include Webui::ElisionsHelper
   protect_from_forgery
 
-  before_action :set_influxdb_data
   before_action :setup_view_path
   before_action :check_user
   before_action :check_anonymous
+  before_action :set_influxdb_data
   before_action :require_configuration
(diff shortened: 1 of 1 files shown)


src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -231,7 +231,8 @@ def set_influxdb_data
     InfluxDB::Rails.current.tags = {
       beta: User.possibly_nobody.in_beta?,
       anonymous: !User.session,
-      interface: :webui
+      interface: :webui,
+      bs_request_action_type: nil
```
eduardoj, line 235:
```suggestion
      interface: :webui
```
I would leave this lines unchanged. Why is it needed to assign `bs_request_action_type` to `nil` here?

hennevogel: we completely changed this PR. please re-review...

eduardoj (approved):

Issue #13866, opened 2023-02-20
Revisions Page improvement


hennevogel:
I was wondering why [some requests](https://obs-measure.opensuse.org/d/fzuOfHQik/measurements-per-request?orgId=1&var-request_id=f95bd60b-4eef-4022-b0fd-69401d08f6c5&from=1676256611958&to=1676256643639&var-method=Webui%3A%3APackageController%23show) to `PackageController#revision` are so slow.

It's because we paginate an array of numbers

https://github.com/openSUSE/open-build-service/blob/022f041aab945711a3d148f82f6ec38cf41bd9c0/src/api/app/controllers/webui/package_controller.rb#L270

to load & render all revisions on the page

https://github.com/openSUSE/open-build-service/blob/022f041aab945711a3d148f82f6ec38cf41bd9c0/src/api/app/views/webui/package/revisions.html.haml#L12

this is a bad idea with a package with 7K+ revisions

dmarcoux:
I turned the links into permalinks, this is possible by pressing `Y` and copying the URL when looking at a file. This is explained [here](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files).

PR #14630, opened 2023-07-11
Show the request creator in package revisions

danidoni:
If you were looking for a specific revision from someone in the package revisions list, there was no way of seeing who created it.

Screenshot from 2023-07-11 16-09-07


src/api/app/views/webui/package/_commit_item.html.haml
```diff
@@ -5,13 +5,17 @@
...
     = link_to(user_path(user)) do
-      = user.realname.present? ? "#{user.realname} (#{user})" : user
+      = realname_with_login(user)
   - else
     %i= commit['user'] || '-'
 
   - if commit['requestid']
     accepted
     = link_to("request #{commit['requestid']}", request_show_path(commit['requestid']))
+    - creator = BsRequest.find(commit['requestid']).creator
```
hennevogel, line 15: This will barf an `ActiveRecord::RecordNotFound` if the request can't be found...

hennevogel (changes requested):

src/api/app/views/webui/package/_commit_item.html.haml
danidoni: True, fixed

hennevogel (approved):

hennevogel:
I fear it will make #13866 even slower but I guess we have to live with this

PR #14633, opened 2023-07-12
Drop unused `user_registrations` table

eduardoj:
Also drop never reachable code.


danidoni:
What about [this](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/configuration.rb#L20) configuration option?

eduardoj:
> What about [this](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/configuration.rb#L20) configuration option?

I don't think it is related...

PR #14637, opened 2023-07-13
Add a TODO file for HAML-Lint

eduardoj:
After HAML-Lint was updated to 0.48.0, running haml-lint resulted in a lot of errors and warnings popping up.

Include a TODO file after running `haml-lint --auto-gen-config` and remove the disabled linters.

This will fix `master` being broken because of the linters.

Further pull requests will address the new errors and warnings.


danidoni:
Wouldn't this be `Disable custom SetPagetitleInView linter` instead?

eduardoj:
> Wouldn't this be `Disable custom SetPagetitleInView linter` instead? It's true that you are adding the todo file, but the important thing is that we disable the custom linter until it's fixed, right?

Hm... nope.

- We disable two linters: SetPagetitleInView and RuboCop.
- We add them in the TODO list because we want to fix those linter complaints. We will address them in the near future. We disable linters in the `.haml-lint.yml` file only when we are sure we don't want to use them.

danidoni (approved):

PR #14638, opened 2023-07-13
Fix custom linter SetPagetitleInView

danidoni:
This is due to [the latest](https://github.com/sds/haml-lint/pull/415) `haml-lint` upgrade.

### DIFF
--- a/src/api/.haml-lint_todo.yml
+++ b/src/api/.haml-lint_todo.yml
@@ -11,7 +11,3 @@ linters:
   # Offense count: 47
   RuboCop:
     enabled: false
-
-  # Offense count: 105
-  SetPagetitleInView:
-    enabled: false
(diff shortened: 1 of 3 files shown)


src/api/.haml-lint_todo.yml
```diff
@@ -14,4 +14,6 @@ linters:
 
   # Offense count: 105
   SetPagetitleInView:
-    enabled: false
+    enabled: true
+    exclude:
+      - 'app/views/webui/users/new.html.haml'
```
eduardoj, line 19: This `SetPagetitleInView` should be removed from this file, and inserted into the `src/api/.haml-lint.yml` file.

eduardoj (changes requested):

eduardoj (approved):

PR #14641, opened 2023-07-13
Download workflow file for private GitLab repositories

rubhanazeem:
Previously we have been using Down gem to download workflows. Down is not able to download workflows from private repos. This PR is using the GitLab client to download private workflow.yaml files.

### DIFF
--- a/src/api/app/services/workflows/yaml_downloader.rb
+++ b/src/api/app/services/workflows/yaml_downloader.rb
@@ -15,7 +15,10 @@ module Workflows
     private
 
     def download_yaml_file
-      Down.download(download_url, max_size: MAX_FILE_SIZE)
+      url = download_url
+      return download_gitlab_yaml_file if url.nil? && @scm_payload[:scm] == 'gitlab'
+
+      Down.download(url, max_size: MAX_FILE_SIZE)
     rescue Down::Error => e
       raise Token::Errors::NonExistentWorkflowsFile, "#{@token.workflow_configuration_url} could not be downloaded.\n#{e.message}" if @token.workflow_configuration_url.present?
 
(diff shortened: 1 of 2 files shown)


src/api/app/services/workflows/yaml_downloader.rb
```diff
@@ -54,5 +54,18 @@ def gitea_download_url
...
+    def download_gitlab_yaml_file
+      gitlab_client = Gitlab.client(endpoint: "#{@scm_payload[:api_endpoint]}/api/v4", private_token: @token.scm_token)
+      gitlab_file = gitlab_client.file_contents(@scm_payload[:project_id], @token.workflow_configuration_path, @scm_payload[:target_branch])
+
+      tempfile = Tempfile.new(["#{Time.zone.now}", '.yaml'])
+      tempfile.write(gitlab_file)
+      tempfile.rewind
+      tempfile
+    rescue Gitlab::Error::NotFound => e
+      raise Token::Errors::NonExistentRepository, e.message
```
eduardoj, line 68:
```suggestion
      begin
        gitlab_client = Gitlab.client(endpoint: "#{@scm_payload[:api_endpoint]}/api/v4", private_token: @token.scm_token)
        gitlab_file = gitlab_client.file_contents(@scm_payload[:project_id], @token.workflow_configuration_path, @scm_payload[:target_branch])
      rescue Gitlab::Error::NotFound => e
        raise Token::Errors::NonExistentRepository, e.message
      end

      tempfile = Tempfile.new(["#{Time.zone.now}", '.yaml'])
      tempfile.write(gitlab_file)
      tempfile.rewind
      tempfile
```
I would be more precise about where to rescue from the Gitlab exception. It would help to understand better where the exception can occur.

rubhanazeem: Okay, It's updated

eduardoj: > Okay, It's updated

Hm... not the way I suggested. In fact, I prefer the original that this last version you propose.

To make more clear what I wanted to express in my first comment, I wanted the added `begin` block to rescue only the lines of code where the exception could be raised up.

In other words, it doesn't make much sense to try to rescue a `Gitlab::Error::NotFound` exception from lines where it cannot occur (those which start with `tempfile`). So we can take out from the `begin` block the lines which start with `tempfile`.

eduardoj (changes requested):
See the comment above.

src/api/app/services/workflows/yaml_downloader.rb
rubhanazeem: I understand what you are saying. I reverted the `begin` statement because Rubocop is complaining about that. I'd instead leave it like this, without `begin`, because it's more clear to read

eduardoj: > I understand what you are saying. I reverted the `begin` statement because Rubocop is complaining about that.

Sure. That what not what I suggested.

> I'd instead leave it like this, without `begin`, because it's more clear to read

Not enough to convince me...

eduardoj (approved):

PR #14655, opened 2023-07-18
Create event subscription for trigger service workflow step

danidoni:


danidoni:
Closing in favor of #14666

PR #14664, opened 2023-07-18
Introduces User.mark_as_spammer!

hennevogel:
Just a little helper method to deal with spam accounts.

1. Enable admin rights for your user with
  ```
   run_in_api rake user:toggle_admin_rights YOUR_LOGIN
  ```
2. Delete the account in the rails console
  ```
  User.find_by(login: 'YOUR_LOGIN').run_as { User.find_by(login: 'SPAMMER_LOGIN').mark_as_spammer! }
  ```
3. Remove admin rights for your user again with
  ```
  run_in_api rake user:toggle_admin_rights YOUR_LOGIN
  ```

### DIFF
--- a/src/api/app/models/user.rb
+++ b/src/api/app/models/user.rb
@@ -116,11 +116,8 @@ class User < ApplicationRecord
   validates :password, confirmation: true, allow_blank: true
   validates :biography, length: { maximum: MAX_BIOGRAPHY_LENGTH_ALLOWED }
 
-  after_create :create_home_project, :track_create
-
-  def track_create
-    RabbitmqBus.send_to_bus('metrics', 'user.create value=1')
-  end
+  after_create :create_home_project, :measure_create
+  after_update :measure_delete
 
   def create_home_project
     # avoid errors during seeding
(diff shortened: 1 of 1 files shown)


danidoni:
Why not using a rake task wrapped with the `user:toggle_admin_rights` instead of a method? That way you don't have to spawn the rails console, just run a task.

hennevogel:
> That way you don't have to spawn the rails console, just run a task.

Does that matter?

danidoni:
> Does that matter?

I think so. Now you need to remember this syntax `User.find_by(login: 'YOUR_LOGIN').run_as { User.find_by(login: 'SPAMMER_LOGIN').mark_as_spammer! }`. And remember to have admin right before and release them after. With a task you don't need to remember to acquire and release the right nor the syntax. You need to remember only the name of the task, and if you don't, you have `rake -T`.

hennevogel:
> With a task you don't need to remember to acquire and release the right nor the syntax

You will have to remember the syntax of the task as it would require *two* arguments...

danidoni (approved):

PR #14679, opened 2023-07-21
Switch deprecated API route to retrieve worker statuses

eduardoj:
See deprecated API route in the documentation: https://api.opensuse.org/apidocs/#/Build/get_build__workerstatus

Found while researching a fix for #14677.


hennevogel (approved):

eduardoj:
@hennevogel, I tried to change the API call to the backend. The backend doesn't support the `/worker/_status` API route. We should not change this call.

"Fortunatelly", I found another call to this deprecated API route, so the purpose and the description of this pull request doesn't change. Please take another look.

hennevogel:
Did you look at the history of the file, I bet i can be removed...

PR #14683, opened 2023-07-21
Tell people about Pundit errors in WorkflowRun

hennevogel:
We started to use Pundit a lot in the various `Workflow::Step`. Authorization failures that happen inside the `Workflow` should be reflected in the `WorkflowRun` and not in the request cycle.

Also: As the `TriggerWorkflowController` takes care about APIErrors and Pundit now, we don't need to handle those exceptions in `Workflow.destroy_target_projects`.

### DIFF
--- a/src/api/app/controllers/trigger_workflow_controller.rb
+++ b/src/api/app/controllers/trigger_workflow_controller.rb
@@ -23,8 +23,12 @@ class TriggerWorkflowController < TriggerController
           @workflow_run.update_as_failed(render_error(status: 400, message: validation_errors.to_sentence))
         end
       end
+    # We always want to return 200 in the request. Because a lot of things in `Workflow`` can go wrong outside this request cycle.
+    # People need to have a look at their `WorkflowRun` to see how `Workflow` went.
     rescue APIError => e
       @workflow_run.update_as_failed(render_error(status: e.status, errorcode: e.errorcode, message: e.message))
+    rescue Pundit::NotAuthorizedError => e
+      @workflow_run.update_as_failed(e.message)
     end
   end
 
--- a/src/api/app/models/workflow.rb
+++ b/src/api/app/models/workflow.rb
@@ -136,12 +136,7 @@ class Workflow
 
     target_project_names = processable_steps.map(&:target_project_name).uniq.compact
 
-    begin
-      Project.where(name: target_project_names).destroy_all
-    rescue Project::Errors::UnknownObjectError, Pundit::NotAuthorizedError => e
-      workflow_run.update_as_failed(e.message)
-      raise e
-    end
+    Project.where(name: target_project_names).destroy_all
   end
 
   # TODO: Extract this into a service


src/api/app/models/workflow.rb
```diff
@@ -136,12 +136,7 @@ def destroy_target_projects
 
     target_project_names = processable_steps.map(&:target_project_name).uniq.compact
 
-    begin
-      Project.where(name: target_project_names).destroy_all
-    rescue Project::Errors::UnknownObjectError, Pundit::NotAuthorizedError => e
-      workflow_run.update_as_failed(e.message)
-      raise e
-    end
+    Project.where(name: target_project_names).destroy_all
```
ncounter, line 139: We are no longer covering the `Project::Errors::UnknownObjectError` now. Am I missing it somewhere?

src/api/app/controllers/trigger_workflow_controller.rb
```diff
@@ -23,8 +23,12 @@ def create
           @workflow_run.update_as_failed(render_error(status: 400, message: validation_errors.to_sentence))
         end
       end
+    # We always want to return 200 in the request. Because a lot of things in `Workflow`` can go wrong outside this request cycle.
+    # People need to have a look at their `WorkflowRun` to see how `Workflow` went.
     rescue APIError => e
       @workflow_run.update_as_failed(render_error(status: e.status, errorcode: e.errorcode, message: e.message))
+    rescue Pundit::NotAuthorizedError => e
+      @workflow_run.update_as_failed(e.message)
```
eduardoj, line 31: This, the `TriggerWorkflowController` controller, is a better place to move the handling of the exceptions than the one before (the `Workflow` model).

But I'm not sure if it would be yet better to handle these exceptions (yes, the `APIError` one too) inside the `Token::Workflow` model, here: https://github.com/openSUSE/open-build-service/blob/1ddf0d702af0f20ae4f9b53fbdede4b39962f7c2/src/api/app/models/token/workflow.rb#L54-L55

src/api/app/models/workflow.rb
hennevogel: Yeah, it's an `APIError`.

src/api/app/controllers/trigger_workflow_controller.rb
hennevogel: Not sure as I don't know where they come from. `Workflows::YAMLDownloader`?

eduardoj: My point is to rescue the `APIError` and the `Pundit::NotAuthorizedError` exceptions into the most nested place possible, and also where we can call the `workflow_run.update_as_failed` method.

That's why I suggested to move the `rescue` lines to the `Token::Workflow` model. There we have the `workflow_run` variable, and we can call its `update_as_failed` method.

Regarding rescuing the `Octokit::Unauthorized` and `Gitlab::Error::Unauthorized` exceptions, I would leave that lines untouched. But yes, I don't know why we don't call the `update_as_failed` method when we rescue these exceptions too. Anyway, that is not related to this PR.

hennevogel: > My point is to rescue the APIError and the Pundit::NotAuthorizedError exceptions into the most nested place possible

That would mean we duplicate the same `rescue` block all over no?

> That's why I suggested to move the rescue lines to the Token::Workflow model. There we have the workflow_run variable, and we can call its update_as_failed method.

Hm but we have `WorkflowRun` in this controller too and also call `update_as_failed` here. What's the thing we gain by moving this down into the token?

eduardoj: > > My point is to rescue the APIError and the Pundit::NotAuthorizedError exceptions into the most nested place possible
> 
> That would mean we duplicate the same `rescue` block all over no?

Not necessarily. We discuss where to `rescue` the exceptions. If we rescue them in the controller, or we rescue them in the model.

I was not clear enough in my previous comments. I suggest to move the new `rescue` block, not to duplicate it.

I wouldn't modify the code in a way we rescue some exceptions in the controller and some other in the model... or duplicate some `rescue` blocks and some other not..., or duplicate all the `rescue` blocks in both places. I would like to have the exception handling not spread in different parts of the codebase.

> > That's why I suggested to move the rescue lines to the Token::Workflow model. There we have the workflow_run variable, and we can call its update_as_failed method.
> 
> Hm but we have `WorkflowRun` in this controller too and also call `update_as_failed` here. What's the thing we gain by moving this down into the token?

1. We have all the `rescue` blocks only in one place. We lower coupling and increase cohesion. See https://www.educative.io/blog/coding-best-practices#lowhigh
1. We abstract handling the exceptions from the controller. From my point of view, we also don't gain from rescuing the exceptions in the controller.

hennevogel: Okay if you want to move all rescues into the model then this will be a way larger refactoring. I suggest we do this later. I've added a card: https://trello.com/c/NqfGRmsJ

PR #14666, opened 2023-07-18
Create event subscription for trigger service and rebuild package workflow steps

danidoni:
Until now we weren't notifying the SCM of the build status triggered by the Trigger Service or the Rebuild Package steps.

How to test this:
- [Set up your environment](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration-Testing) to set up your environment, but don't trigger the workflow yet.
- Tweak the package so the building fails (I use an `exit(1)` in the `%build` phase)
- Use a `payload.json` like this one:
```
{
  "object_kind": "push",
  "before": "60cd89df0215ab64cff9b8076a4a3334dce71ba7",
  "after": "1873d25e0fb0c245e3ad75aadca1f69fdf5fd70b",
  "ref": "refs/heads/master",
  "project_id": 16871458,
  "project": {
    "id": 16871458,
    "name": "hello_world",
    "web_url": "https://github.com/danidoni/hello_world",
    "path_with_namespace": "danidoni/hello_world",
    "http_url": "https://github.com/danidoni/hello_world.git"
  },
  "commits": [
    {
      "id": "1873d25e0fb0c245e3ad75aadca1f69fdf5fd70b",
      "url": "https://github.com/danidoni/hello_world/-/commit/1873d25e0fb0c245e3ad75aadca1f69fdf5fd70b"
    }
  ],
  "repository": {
    "name": "hello_world",
    "url": "git@github.com:danidoni/hello_world.git",
    "homepage": "https://github.com/danidoni/hello_world",
    "full_name": "danidoni/hello_world"
 }
}
```
- Use a workflows.yml like [this one](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration-Workflows#rebuild-a-package) to trigger a Rebuild.
- Set up a Github repository publicly reachable, e.g.: `https://github.com/danidoni/hello_world`. Replace the repository path in the payload above to match.
- Push some changes and get the commit of the change. Replace the commit id and url of the payload above to match.
- Trigger the workflow.
- Let the package build. It should end with a failed build.
- Go to the repository, in the commit list you should see an indication that the commit has failures (on GitHub there should be a red cross next to the commit id).

PR #14690, opened 2023-07-24
Reuse existing event subscription to notify SCMs back

danidoni:
When including the workflow run in the query, we do not get any EventSubscription back, then the `create` part of the `find_or_create_by` kicks in and we end up having tons of notifications.

Follows up https://github.com/openSUSE/open-build-service/pull/14666

How to test this:
- [Set up your environment](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration-Testing), but don't trigger the workflow yet.
- Use a `workflows.yml` like [this one](https://github.com/openSUSE/open-build-service/wiki/Better-SCM-CI-Integration-Workflows#rebuild-a-package) to trigger a Rebuild.
- Tweak the package so the building fails (I use an `exit(1)` in the `%build` phase)
- Push some changes against your default branch, let the workflow finish and you should get a :x: next to the commit telling you the build failed.
- Fix the package spec so it builds successfully.
- Push another change, let the workflow finish and you should get a :heavy_check_mark:  next to the commit telling you the build succeeded.
- Before this change the older commit would change to :heavy_check_mark: also. Now it stays as failed :x:.

### DIFF
--- a/src/api/app/services/workflows/scm_event_subscription_creator.rb
+++ b/src/api/app/services/workflows/scm_event_subscription_creator.rb
@@ -16,9 +16,9 @@ module Workflows
                                              channel: 'scm',
                                              enabled: true,
                                              token: @token,
-                                             package: @package,
-                                             workflow_run: @workflow_run).tap do |subscription|
-          subscription.update!(payload: @scm_webhook.payload) # The payload is updated regardless of whether the subscription already existed or not.
+                                             package: @package).tap do |subscription|
+          # Set payload and workflow_run regardless of whether the subscription already existed or not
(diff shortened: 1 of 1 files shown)


src/api/app/services/workflows/scm_event_subscription_creator.rb
```diff
@@ -16,8 +16,7 @@ def call
                                              channel: 'scm',
                                              enabled: true,
                                              token: @token,
-                                             package: @package,
-                                             workflow_run: @workflow_run).tap do |subscription|
+                                             package: @package).tap do |subscription|
```
hennevogel, line 19: You will have to update the `workflow_run` association afterward...

PR #14693, opened 2023-07-24
Never report on Tag Push events

hennevogel:
SCMs don't support statuses for tags, so we don't need to report back in ANY case


hennevogel:
[Handle reporting in rebuild_package/trigger_services steps (M)](https://trello.com/c/ypR24apr/2300-handle-reporting-in-rebuildpackage-triggerservices-steps-m)

rubhanazeem (approved):

PR #14694, opened 2023-07-24
Revisions pagination info

ncounter:
The badge counter in the Revisions page title is buggy:
- it contains the number of the elements visible in the page (i.e. if we are at page 2 of a list of 25 elements with 10 items per page, it reports `10` because it counts the paginated slice only)
- how many items are shown in the page (paginated or not) is not a property to deliver in the title as a badge
- all about pagination should be forwarded and managed by Kaminari

Before
image

After
(1) 
image

(2)
image

### DIFF
--- a/src/api/app/views/webui/package/revisions.html.haml
+++ b/src/api/app/views/webui/package/revisions.html.haml
@@ -23,11 +23,10 @@
         .col-auto
           = page_entries_info @revisions, entry_name: 'revision'
 
-      - unless params['show_all']
-        - unless @revisions.total_pages == 1 || !User.session
-          .text-center.mt-4= link_to('Show all',
-                                package_view_revisions_path(project: @project, package: @package, show_all: 1),
-                                class: 'btn btn-sm btn-secondary')
+      - unless params['show_all'] || @revisions.total_pages == 1 || !User.session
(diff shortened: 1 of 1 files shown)


src/api/app/views/webui/package/revisions.html.haml
```diff
@@ -4,24 +4,29 @@
...
-          .text-center= link_to('Show all',
-                                package_view_revisions_path(project: @project, package: @package, show_all: 1),
-                                class: 'btn btn-sm btn-secondary')
+
+      .row.justify-content-center.mt-2
+        .col-auto#revisions-pagination
+          = paginate @revisions, views_prefix: 'webui'
+
+      .row.justify-content-center.mt-2
+        .col-auto#revisions-page-entries-info
```
eduardoj, line 23:
```suggestion
        .col-auto
          = paginate @revisions, views_prefix: 'webui'

      .row.justify-content-center.mt-2
        .col-auto
```
No need to specify a unique id for these HTML elements if we are not going to use them, I guess.

ncounter: You're right. Fixed, thanks

eduardoj (approved):

PR #14696, opened 2023-07-25
README.md: use Leap 15.4 instead of 15.1

johanneskastl:
- use Leap 15.4 instead of 15.1
- use new repository naming pattern (`15.4` instead of `openSUSE_15.1`)


eduardoj (approved):
Thanks for the fix!

PR #14717, opened 2023-07-27
Disable partial inserts in active record

hellcp-work:
This being enabled causes strong migrations to complain about a dangerous operation while migrating our older migration, when linting the migrations or generating schema.rb

### DIFF
--- a/src/api/config/application.rb
+++ b/src/api/config/application.rb
@@ -109,6 +109,10 @@ module OBSApi
     config.active_record.cache_versioning = true
     config.active_record.collection_cache_versioning = false
 
+    # Disable partial writes to avoid causing incorrect values
+    # to be inserted when changing the default value of a column.
+    config.active_record.partial_inserts = false
+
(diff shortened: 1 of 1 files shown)


src/api/config/application.rb
```diff
@@ -109,6 +109,10 @@ class Application < Rails::Application
     config.active_record.cache_versioning = true
     config.active_record.collection_cache_versioning = false
 
+    # Disable partial writes to avoid causing incorrect values
+    # to be inserted when changing the default value of a column.
+    config.active_record.partial_inserts = false
```
hennevogel, line 114: Hm isn't this the default? :)

hennevogel: I guess fixing this is the thing you want to do then

https://github.com/openSUSE/open-build-service/blob/6655095aadacd0557c389828fcc4e8c187ab6f5d/src/api/config/application.rb#L41

danidoni (approved):

PR #14720, opened 2023-07-27
Workflow run menu enhancements

ncounter:
Little improvements on the workflow_run filters menu:
- use the icon of the status type (already in use in the list of entries) in the related filter link
- do not use a badge for the counter of each type, just show a lighter number
- align the counter of each type to the right

Before:
image

After:
image

### DIFF
--- a/src/api/app/components/workflow_run_filter_link_component.html.haml
+++ b/src/api/app/components/workflow_run_filter_link_component.html.haml
@@ -1,4 +1,8 @@
-= link_to(token_workflow_runs_path(@token, @filter_item), class: "list-group-item list-group-item-action #{css_for_link}") do
-  = @text
+= link_to(token_workflow_runs_path(@token, @filter_item),
+          class: "list-group-item list-group-item-action #{css_for_link} d-flex justify-content-between") do
+  %span
+    - if @icon != ''
+      %i.me-1{ class: "#{@icon} #{css_for_icon}" }
+    = @text
   - if @amount.positive?
     %span.badge.align-text-top.ms-2{ class: css_for_badge_color }>= @amount
--- a/src/api/app/components/workflow_run_filter_link_component.html.haml
+++ b/src/api/app/components/workflow_run_filter_link_component.html.haml
@@ -5,4 +5,4 @@
       %i.me-1{ class: "#{@icon} #{css_for_icon}" }
     = @text
   - if @amount.positive?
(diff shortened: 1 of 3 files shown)


src/api/app/components/workflow_run_filter_link_component.html.haml
```diff
@@ -1,4 +1,8 @@
...
-  = @text
+= link_to(token_workflow_runs_path(@token, @filter_item), class: "list-group-item list-group-item-action #{css_for_link} d-flex justify-content-between") do
+  %span
+    - if @icon != ''
+      %i.me-1{class: "#{@icon} #{css_for_icon}"}
+    = @text
   - if @amount.positive?
-    %span.badge.align-text-top.ms-2{ class: css_for_badge_color }>= @amount
+    %span
+    = "(#{@amount})"
```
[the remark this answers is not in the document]
ncounter: True, it's even lighter. I'll update the screenshot too :+1:

ncounter:
> Is this a pattern we should also repeat for other views with similar design? Notifications in particular cone to mind

@hellcp-work yes, I'd really like to do so. I don't Unfortunately the two features does not share the same code, we should probably deduplicate it with a core view component or something to reuse. But for now, I can quickly go through the notification filters menu as well and apply the same changes, just to prevent inconsistencies between the two pages.

ncounter:
@hellcp-work this is for the notification view https://github.com/openSUSE/open-build-service/pull/14721 , same changes to the parallel implementation.

src/api/app/components/workflow_run_filter_link_component.html.haml
```diff
@@ -1,4 +1,8 @@
-= link_to(token_workflow_runs_path(@token, @filter_item), class: "list-group-item list-group-item-action #{css_for_link}") do
-  = @text
+= link_to(token_workflow_runs_path(@token, @filter_item),
+          class: "list-group-item list-group-item-action #{css_for_link} d-flex justify-content-between") do
+  %span
+    - if @icon != ''
+      %i.me-1{ class: "#{@icon} #{css_for_icon}" }
+    = @text
```
danidoni, line 6: This can be pushed into the `WorkflowRunFilterLinkComponent`

ncounter: done

danidoni (approved):

PR #14730, opened 2023-07-28
WIP: Filter workflow runs by event source name

saraycp:


src/api/app/queries/workflow_runs_finder.rb
```diff
@@ -40,29 +40,11 @@ def with_generic_event_type(generic_event_type, request_action = nil)
...
-    #           end.join(' OR ')
-    #         end
 
-    workflow_runs = @relation.where(query)
-    if request_action && generic_event_type == 'pull_request'
-      workflow_runs = workflow_runs.where("JSON_EXTRACT(request_payload, '$.action') = (?) OR JSON_EXTRACT(request_payload, '$.object_attributes.action') = (?)", request_action,
-                                          request_action)
+    if event_source_name
+      workflow_runs = workflow_runs.where(event_source_name: event_source_name)
     end
```
danidoni:
```suggestion
    return workflow_runs.where(event_source_name: event_source_name) if event_source_name
```

saraycp: This code has been moved to an independent method. Please have another look.

danidoni (approved):

saraycp:
I've just rebased on master

src/api/app/components/workflow_run_filter_input_component.html.haml
```diff
@@ -0,0 +1,7 @@
+= form_tag(token_workflow_runs_path(token_id), method: :get, class: 'd-flex px-3 py-2') do
```
ncounter, line 1: I'd rather stick to the same appearance.
Maybe you can still give the `.list-group-item` CSS class to this element to be consistent with the other filters, and give the `.active` CSS class in case it is selected. It will behave the same way of the others by having a green background when in use.
For the button instead keeping the white background should be enough in both active/non active cases.

src/api/spec/queries/workflow_runs_finder_spec.rb
```diff
@@ -3,13 +3,13 @@
 RSpec.describe WorkflowRunsFinder do
   let(:workflow_token) { create(:workflow_token) }
   # GitHub
-  let!(:workflow_run_github_push) { create(:workflow_run, :push, :succeeded, token: workflow_token) }
-  let!(:workflow_run_github_tag_push) { create(:workflow_run, :tag_push, :succeeded, token: workflow_token) }
+  let!(:workflow_run_github_push) { create(:workflow_run, :push, :succeeded, token: workflow_token, event_source_name: '94561cd') }
```
ncounter, line 6: Maybe better to add the test case when the filter is active (button search is clicked) but the input value is empty. It will fail (AKA the `with_event_source_name` needs to be adapted I guess) because the result I'd expect is to show them all, but it shows nothing instead.

An alternative is to prevent submitting the filter if the input value is empty.

hennevogel:
Here https://obs-reviewlab.opensuse.org/saraycp-event_source_name_filter/my/tokens/1/workflow_runs 

I see a couple of things that don't make sense:

- all `#1` links point to the workflow run index again?
- if I filter per PR/MR `1` there is no way to get rid of the filter again

saraycp:
> * if I filter per PR/MR `1` there is no way to get rid of the filter again

@hennevogel,  the user can always click on another filter or `All`, but with my last changes, once the user makes the text input empty and click on the button, the whole list of workflow runs is displayed. Is that enough?

filter_input_empty

src/api/spec/queries/workflow_runs_finder_spec.rb
ncounter: addressed in [Empty filter input displays all the results](https://github.com/openSUSE/open-build-service/pull/14730/commits/de042cbbb969fbe31e9483975b970fa626b6814f)

hennevogel (approved):

PR #14732, opened 2023-07-28
Handle services in progress while branching a package

ncounter:
https://trello.com/c/8LUmlVIr

If `service` is running on a package while trying to branch it, SCM Workflow avoids the branching returning a message that `service is running`. This PR adds a `wait_service` call for the target package, so the branching can happen successfully right after service ends.

### DIFF
--- a/src/api/app/models/workflow/step/branch_package_step.rb
+++ b/src/api/app/models/workflow/step/branch_package_step.rb
@@ -70,6 +70,9 @@ class Workflow::Step::BranchPackageStep < Workflow::Step
     end
 
     begin
+      # Service running on package avoids branching it: wait until services finish
+      Backend::Api::Sources::Package.wait_service(source_project_name, source_package_name)
+
       BranchPackage.new({ project: source_project_name, package: source_package_name,
                           target_project: target_project_name,
                           target_package: target_package_name }).branch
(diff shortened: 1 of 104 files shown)


src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -70,9 +70,14 @@ def create_branched_package
...
 
     begin
+      # Service running on package avoids branching it
+      Backend::Api::Sources::Package.wait_service(source_project_name, source_package_name)
+
       BranchPackage.new({ project: source_project_name, package: source_package_name,
                           target_project: target_project_name,
                           target_package: target_package_name }).branch
+
+      Backend::Api::Sources::Package.runservice(source_project_name, source_package_name)
```
hennevogel, line 80: Branching a package creates files which triggers services. No need to do this manually...

ncounter: Ah! I missed it triggers the source ones. Thanks for the hint. :+1:

src/api/app/models/workflow/step/branch_package_step.rb
```diff
@@ -70,6 +70,9 @@ def create_branched_package
     end
 
     begin
+      # Service running on package avoids branching it: wait until services finish
+      Backend::Api::Sources::Package.wait_service(source_project_name, source_package_name) unless scm_synced?
```
hennevogel, line 74: Why would you not wait if we source package is scmsync?

danidoni (approved):

hennevogel (approved):

PR #14750, opened 2023-08-01
Support private GitHub repos for SCM/CI integration

rubhanazeem:
This PR ensures the support of private GitHub repos for SCM/CI integration.

PR #14754, opened 2023-08-02
Refactor yaml downloader

rubhanazeem:
This pr is rebased with PR #14750 and it only refactors the `YAMLDownloader`


src/api/app/services/workflows/yaml_downloader.rb
```diff
@@ -15,46 +15,42 @@ def call
     private
 
     def download_yaml_file
```
hennevogel, line 17: This is just `.call`...

src/api/app/services/workflows/yaml_downloader.rb
```diff
@@ -15,46 +15,42 @@ def call
...
-    rescue Down::Error => e
-      raise Token::Errors::NonExistentWorkflowsFile, "#{@token.workflow_configuration_url} could not be downloaded.\n#{e.message}" if @token.workflow_configuration_url.present?
-
-      # 'target_branch' can contain a commit sha (when tag push) instead of a branch name
-      raise Token::Errors::NonExistentWorkflowsFile, "#{@token.workflow_configuration_path} could not be downloaded from the SCM branch/commit #{@scm_payload[:target_branch]}." \
-                                                     "\nIs the configuration file in the expected place? Check #{DOCUMENTATION_LINK}\n#{e.message}"
-    end
-
-    def download_url
       # When an external URL is given, it prevails over the path.
```
hennevogel, line 18: This comment doesn't provide any extra context anymore. We can remove it IHMO

src/api/app/services/workflows/yaml_downloader.rb
```diff
@@ -15,46 +15,42 @@ def call
...
+      return download_from_url(@token.workflow_configuration_url) if @token.workflow_configuration_url.present?
 
       case @scm_payload[:scm]
-      when 'github'
-        client = Octokit::Client.new(access_token: @token.scm_token, api_endpoint: @scm_payload[:api_endpoint])
-        # :ref can be the name of the commit, branch or tag.
-        client.content("#{@scm_payload[:target_repository_full_name]}", path: "/#{@token.workflow_configuration_path}", ref: @scm_payload[:target_branch])[:download_url]
       when 'gitea'
-        gitea_download_url
+        download_from_url(gitea_configuration_file_url)
```
hennevogel, line 23: I would do this inside  `gitea_configuration_file_url` and call that method `download_gitea_yaml_file` for consistency.

src/api/app/services/workflows/yaml_downloader.rb
```diff
@@ -15,46 +15,42 @@ def call
...
+      begin
+        content = client.content("#{@scm_payload[:target_repository_full_name]}", path: "/#{@token.workflow_configuration_path}", ref: @scm_payload[:target_branch])[:content]
+      rescue Octokit::InvalidRepository => e
+        raise Token::Errors::NonExistentRepository, e.message
+      rescue Octokit::NotFound => e
+        # 'target_branch' can contain a commit sha (when tag push) instead of a branch name
+        raise Token::Errors::NonExistentWorkflowsFile,
+              "#{@token.workflow_configuration_path} could not be downloaded from the SCM branch/commit #{@scm_payload[:target_branch]}: #{e.message}"
+      end
+      create_temp_file(Base64.decode64(content))
```
hennevogel, line 51: Hm, why do we need to create a temp file?

src/api/app/services/workflows/yaml_downloader.rb
```diff
@@ -63,11 +59,24 @@ def download_gitlab_yaml_file
...
     end
+
+    def download_from_url(url)
+      Down.download(url, max_size: MAX_FILE_SIZE)
+    rescue Down::Error => e
+      raise Token::Errors::NonExistentWorkflowsFile, "#{@token.workflow_configuration_url} could not be downloaded.\n#{e.message}" if @token.workflow_configuration_url.present?
+
+      # 'target_branch' can contain a commit sha (when tag push) instead of a branch name
+      raise Token::Errors::NonExistentWorkflowsFile, "#{@token.workflow_configuration_path} could not be downloaded from the SCM branch/commit #{@scm_payload[:target_branch]}." \
+                                                     "\nIs the configuration file in the expected place? Check #{DOCUMENTATION_LINK}\n#{e.message}"
```
hennevogel, line 79: 💭  How do you know that the `Down:Error` means it's not in the right place? It could be any reason...

hennevogel (changes requested):

src/api/app/services/workflows/yaml_downloader.rb
rubhanazeem: Yeah, removing this method and adding it to `call` makes sense. Done

src/api/app/services/workflows/yaml_downloader.rb
rubhanazeem: This has already been discussed. It's part of #14750

src/api/app/services/workflows/yaml_downloader.rb
rubhanazeem: Done.

src/api/app/services/workflows/yaml_downloader.rb
rubhanazeem: This is not new. It has been the same in master. 
The reasoning is, `Down` is just taking the URL of the file and then just downloads it. If Down is not able to download it then the file does not exist on the given URL. 

It's also important to note that the `Down` gem is only being used for `Gitea` or when the `workflow_configuration_url` is provided in workflow token.

hennevogel (approved):

PR #14772, opened 2023-08-07
Fix initial reporting

hennevogel:
The first payload is the OBS event payload, not the webhook payload.

### DIFF
--- a/src/api/app/models/workflow/step/submit_request.rb
+++ b/src/api/app/models/workflow/step/submit_request.rb
@@ -43,7 +43,7 @@ class Workflow::Step::SubmitRequest < Workflow::Step
     @bs_request.save!
 
     Workflows::ScmEventSubscriptionCreator.new(token, workflow_run, scm_webhook, @bs_request).call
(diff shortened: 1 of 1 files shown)


src/api/app/models/workflow/step/submit_request.rb
```diff
@@ -43,7 +43,7 @@ def submit_package
     @bs_request.save!
 
     Workflows::ScmEventSubscriptionCreator.new(token, workflow_run, scm_webhook, @bs_request).call
-    SCMStatusReporter.new(scm_webhook.payload, scm_webhook.payload, @token.scm_token, workflow_run, 'Event::RequestStatechange').call
+    SCMStatusReporter.new({number: @bs_request.number, state: @bs_request.state}, scm_webhook.payload, @token.scm_token, workflow_run, 'Event::RequestStatechange').call
```
danidoni, line 46:
```suggestion
    SCMStatusReporter.new({ number: @bs_request.number, state: @bs_request.state }, scm_webhook.payload, @token.scm_token, workflow_run, 'Event::RequestStatechange').call
```

PR #14780, opened 2023-08-08
Add "Update of the Base Distributions" section to the release notes

eduardoj:


hennevogel (approved):

ReleaseNotes-2.10.21
```diff
@@ -11,6 +11,24 @@ The dist/README.UPDATERS file has information for people updating
 from a previous OBS release.
 
 
+Update of the Base Distributions
+================================
+
+We have updated the base distribution of the appliance to Leap 15.4. We also
```
hennevogel, line 17:
```suggestion
We have updated the base distribution of the appliance to Leap 15.5. We also
```

hennevogel (changes requested):

ReleaseNotes-2.10.21
eduardoj: Applied. Thanks!

PR #14786, opened 2023-08-09
Avoid duplicated gem diff comments

krauselukas:
By using the `comment_tag` the github action `actions-comment-pull-request` will update an existing comment over creating a new one. This avoids flooding the depfu PR's with comment when they are updated multiple times.

For reference https://github.com/thollander/actions-comment-pull-request#update-a-comment


krauselukas:
[Avoid Duplicated Gem Diff Comments](https://trello.com/c/cUQyrN7U/2319-avoid-duplicated-gem-diff-comments)

eduardoj (approved):

PR #14791, opened 2023-08-10
Extract logic accessing the database from a component

danidoni:
As a best practice, we should not be accessing the database from inside a View Component. The WatchlistComponent now accepts the logic accessing the database as parameters in the constructor.


src/api/app/models/watchlist_items.rb
```diff
@@ -0,0 +1,9 @@
+class WatchlistItems
```
hennevogel, line 1: Two more objects around `*ForWatchlistFinder`?

danidoni: What do you suggest?

hennevogel: scope `watched_items` in `User`

src/api/app/models/user.rb
```diff
@@ -75,6 +75,13 @@ class User < ApplicationRecord
   scope :in_beta, -> { where(in_beta: true) }
   scope :in_rollout, -> { where(in_rollout: true) }
 
+  scope :watchlisted_requests, ->(user) { BsRequest.joins(:watched_items).where(watched_items: { user: user }).order('number DESC') }
```
hennevogel, line 78: Hm no need to join monstrous tables if you can just fetch the IDs of the `BsRequest`s from `watched_items`?

hennevogel (changes requested):

src/api/app/components/watchlist_component.rb
```diff
@@ -5,12 +5,17 @@ class WatchlistComponent < ApplicationComponent
     'BsRequest' => 'request'
   }.freeze
 
-  def initialize(user:, current_object:, bs_request: nil, package: nil, project: nil)
+  attr_reader :projects, :packages, :bs_requests
+
+  def initialize(user:, current_object:, bs_request: nil, package: nil, project: nil, bs_requests: [], packages: [], projects: [])
```
hennevogel, line 10: What do you need the singular `bs_request`, `package` and `project` for?

danidoni: Those contain the object we are currently viewing in the app. I think they collide somehow with `current_object`, but I didn't want to fix that in this PR

src/api/app/models/user.rb
rubhanazeem: This is updated. We are using `watched_items` association here

src/api/app/models/user.rb
```diff
@@ -75,6 +75,10 @@ class User < ApplicationRecord
   scope :in_beta, -> { where(in_beta: true) }
   scope :in_rollout, -> { where(in_rollout: true) }
 
+  scope :watchlisted_requests, ->(user) { BsRequest.where(id: user.watched_items.where(watchable_type: 'BsRequest').pluck(:watchable_id)).order('number DESC') }
+  scope :watchlisted_packages, ->(user) { Package.where(id: user.watched_items.where(watchable_type: 'Package').pluck(:watchable_id)).order('LOWER(name), name') }
+  scope :watchlisted_projects, ->(user) { Project.where(id: user.watched_items.where(watchable_type: 'Project').pluck(:watchable_id)).order('LOWER(name), name') }
```
hennevogel, line 80:
```suggestion
  scope :watched_requests, -> { BsRequest.where(id: watched_items.where(watchable_type: 'BsRequest').pluck(:watchable_id)).order('number DESC') }
  scope :watched_packages, -> { Package.where(id: watched_items.where(watchable_type: 'Package').pluck(:watchable_id)).order('LOWER(name), name') }
  scope :watched_projects, -> { Project.where(id: watched_items.where(watchable_type: 'Project').pluck(:watchable_id)).order('LOWER(name), name') }
```

hennevogel (changes requested):

src/api/app/models/user.rb
rubhanazeem: Scopes are probably not the best idea to fetch these records. Instead I've added instance methods because these are user specific querirs

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -209,6 +210,13 @@ def current_announcement
     @current_announcement = StatusMessage.latest_for_current_user
   end
 
+  def fetch_watchlist_items
+    user = User.possibly_nobody
+    @watchlist_bs_requests = user.watched_requests
+    @watchlist_packages = user.watched_packages
+    @watchlist_projects = user.watched_projects
```
hennevogel, line 217:
```suggestion
    @watched_requests =  User.possibly_nobody.watched_requests
    @watched_packages =  User.possibly_nobody.watched_packages
    @watched_projects =  User.possibly_nobody.watched_projects
```

hennevogel (changes requested):

src/api/app/controllers/webui/webui_controller.rb
hennevogel: Please do not rename thing in every layer...

PR #14803, opened 2023-08-14
Transfer scmsync while branching

hennevogel:
instead of only creating an empty package.

### DIFF
--- a/src/api/app/models/branch_package.rb
+++ b/src/api/app/models/branch_package.rb
@@ -147,7 +147,7 @@ class BranchPackage
         raise DoubleBranchPackageError.new(tprj.name, tpkg.name), "branch target package already exists: #{tprj.name}/#{tpkg.name}" unless params[:force]
       else
         if pac.is_a?(Package)
-          tpkg = tprj.packages.new(name: pack_name, title: pac.title, description: pac.description, url: pac.url)
+          tpkg = tprj.packages.new(name: pack_name, title: pac.title, description: pac.description, url: pac.url, scmsync: pac.scmsync)
           tpkg.bcntsynctag = pac.bcntsynctag
         else
           tpkg = tprj.packages.new(name: pack_name)


adrianschroeter:
no, that is pointless. a fork needs to be used from a different place.

This needs to be handled by external tooling.

hennevogel:
> no, that is pointless. a fork needs to be used from a different place.

Would you mind to elaborate what "different place" you are talking about?

adrianschroeter:
On Montag, 14. August 2023, 13:25:03 CEST Henne Vogelsang wrote:
> > no, that is pointless. a fork needs to be used from a different place.
> 
> Would you mind to elaborate what "different place" you are talking about?

a different repo or branch at least.

Just discussed that with daniel, it is more likely that we will need
a different "branch" cmd, one that is not executing source handling, but
takes a new scm repo url instead.

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> Just discussed that with daniel, it is more likely that we will need a different "branch" cmd, one that is not executing source handling, but takes a new scm repo url instead.

I'm sure that would be nice to have (we already do that in SCM/CI workflows with the scmsync attribute) but why would that mean we do not support transferring this attribute as is when someone wants to branch the package?

adrianschroeter:
On Dienstag, 22. August 2023, 15:58:34 CEST Henne Vogelsang wrote:
> @adrianschroeter why would we do not transfer this attribute, as is, when someone branches a package? Just because somewhere else we might use `scmsync` in a different setup?

It makes little sense to have two projects always building the same source
with same repository setup.

Also creating first a source branch in the backend and removing it briefly
afterwards makes no sense.

And third, keep in mind that the result is different, so also follow up 
steps, like creating a submit request won't work as expected.

Therefore, we need in any case a propper support in client tooling here.
And then this tooling can also create a propper request to setup
a project based on another git fork or branch.

I can follow up on this with Daniel.

good morning
adrian

-- 

Adrian Schroeter ***@***.***>

hennevogel:
> It makes little sense to have two projects always building the same source with same repository setup.
> Also creating first a source branch in the backend and removing it briefly afterwards makes no sense.

Hm, but this isn't this exactly what we are *always* doing when branching without `scmsync`?

adrianschroeter:
Here is a fork implementation to be used for scmsync packages: https://github.com/openSUSE/open-build-service/pull/14830

please note that we may let the source fail on branch command in future when the source exists and has an scmsync configured to avoid non supported setups.

hennevogel:
> No, the branch (_link file with branch element plus full copy of sources) remains after the branch command. Here it would be removed when the scmsync job got executed (hopefully).

Sounds like you did not try what is happening now. No files are transferred, just the package object and the repositories and *NOT* the scmsync url. Try it...

```
osc branch home:hennevogel:test/aaa_base
```

adrianschroeter:
On Donnerstag, 24. August 2023, 13:21:20 CEST Henne Vogelsang wrote:
> > No, the branch (_link file with branch element plus full copy of sources) remains after the branch command. Here it would be removed when the scmsync job got executed (hopefully).
> 
> Sounds like you did not try what is happening now. No files are transferred, just the package object and the repositories and *NOT* the scmsync url. Try it...
> 
> ```
> osc branch home:hennevogel:test/aaa_base

yes, the result is broken. And it will remain broken, just in a different way.

Tnis call should not work at all.

We will never support mixing old-style source handling mechanics with scmsync.

And in any case, it makes no sense to build always the same source in two package instances.

What you want is to test your git fork with a specific source change instead.

-- 

Adrian Schroeter ***@***.***>

PR #14814, opened 2023-08-18
Move diffend io github action to ruby script

krauselukas:


krauselukas:
[Make Gem Diff Github Action work when there is only a diff for a `.gem` file](https://trello.com/c/L5rZacKj/2318-make-gem-diff-github-action-work-when-there-is-only-a-diff-for-a-gem-file)

PR #14834, opened 2023-08-24
Make BranchPackageStep/LinkPackageStep.call readable

hennevogel:
.call is where the action happens, no need to move some of it into a separate methods. Only makes this harder to follow.


krauselukas (approved):
LGTM

hennevogel:
depends on #14847

hennevogel:
@krauselukas I guess you should have a look again

krauselukas (approved):

PR #14847, opened 2023-08-28
Stop supporting scmsync for some operations

hennevogel:
Creating a link is creating a _link file but creating files (outside git) is not supported in scmsync'ed packages.


rubhanazeem (approved):

src/api/spec/models/workflow/step/link_package_step_spec.rb
```diff
@@ -380,51 +380,6 @@
         it_behaves_like 'insufficient permission on target project'
         it_behaves_like 'insufficient permission to create new target project'
       end
-
-      context 'when scmsync is active' do
```
eduardoj, line 384: Removing this context make is unnecessary to keep its cassettes under `spec/cassettes/Workflow_Step_LinkPackageStep/_call/when_the_SCM_is_GitHub/when_scmsync_is_active/`. Same thing for the other two removed contexts.

eduardoj (changes requested):

src/api/spec/models/workflow/step/link_package_step_spec.rb
hennevogel: Which other two contexts?

hennevogel: Ah, the nested ones. Yeah they are gone..

eduardoj (approved):

PR #14850, opened 2023-08-29
Request page: build results chart overview

ncounter:
Make use of https://chartkick.com/ to present a full overview of all the build results of all the actions of a request.

### Request build results chart
image
image

### Request build results simplified view (threshold: <= 6 distinct build results repositories)
image


hennevogel:
I guess it would make sense to stick to integers for the x-axis

Screenshot from 2023-08-31 11-33-13

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -299,6 +299,66 @@ def build_results
...
+    @actions.where(type: [:submit, :maintenance_incident, :maintenance_release]).each do |action|
+      source_project_object = Project.find_by_name(action.source_project)
+      source_package_object = Package.find_by_project_and_name(source_project_object.name, action.source_package)
+
+      # fetch all build results for the source project/package
+      action_build_results_data = LocalBuildResult::ForPackage.new(
+        package: source_package_object,
+        project: source_project_object,
+        show_all: true,
+        lastbuild: true
```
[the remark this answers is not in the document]
ncounter: I guess you're right. I overlooked about the real meaning of it. Adjusting it now....

rubhanazeem:
When the number of repositories is smaller then the graph should be different. We are creating the same graph for 1 repository or 100 repositories. There is no threshold defined for the smaller number of repositories. It's up to you to choose a suitable number.

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -299,6 +299,66 @@ def build_results
     @ajax_data['package'] = @action[:spkg] if @action[:spkg]
   end
 
+  def build_results_chart
```
rubhanazeem, line 302: Is it possible to take out all this logic to a view component?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -299,6 +299,66 @@ def build_results
...
+    failed_data = {}
+    building_data = {}
+
+    # take all the build results for all the actions (only actions where it makes sense to have a build status)
+    @actions.where(type: [:submit, :maintenance_incident, :maintenance_release]).each do |action|
+      source_project_object = Project.find_by_name(action.source_project)
+      source_package_object = Package.find_by_project_and_name(source_project_object.name, action.source_package)
+
+      # fetch all build results for the source project/package
+      action_build_results_data = LocalBuildResult::ForPackage.new(
```
krauselukas, line 313: @ncounter reusing this https://github.com/openSUSE/open-build-service/blob/2b928d8e8374de93addeb4d2d30da986dcdcf54c/src/api/app/models/package.rb#L856 would might save a few lines of code :)

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -299,6 +299,66 @@ def build_results
...
   end
 
+  def build_results_chart
+    success_data = {}
+    failed_data = {}
+    building_data = {}
+
+    # take all the build results for all the actions (only actions where it makes sense to have a build status)
+    @actions.where(type: [:submit, :maintenance_incident, :maintenance_release]).each do |action|
+      source_project_object = Project.find_by_name(action.source_project)
```
krauselukas, line 309: @ncounter what I meant this morning regarding the build results for a request to a staging project. I think we have to consider it here as well. Basically whats done here https://github.com/openSUSE/open-build-service/blob/2b928d8e8374de93addeb4d2d30da986dcdcf54c/src/api/app/controllers/webui/request_controller.rb#L294
 So showing the build results of the staging project (the `bs_request` and `bs_request_action` then contains a `:staging_project_id`).

src/api/app/views/webui/request/beta_show.html.haml
```diff
@@ -100,6 +100,12 @@
                     %strong auto-accepted
                     = render TimeComponent.new(time: @bs_request.accept_at)
 
+              .m-4
+                %h4
+                  Build results summary
+                = column_chart(request_build_results_chart_path(number: @bs_request.number),
+                    width: "100%", height: "300px", stacked: true)
```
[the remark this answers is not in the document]
ncounter: Done, the whole box is hidden in case of `no_data`

src/api/app/controllers/webui/request_controller.rb
ncounter: Addressed, thank you for the pointer :+1:

ncounter:
> @ncounter the architectures are overflowing on mobile view image

Good catch @krauselukas , thank you. I pushed the fix for that right now.

src/api/app/components/chart_component.rb
```diff
@@ -0,0 +1,102 @@
...
+            repository: result_entry.repository,
+            status: result_entry.code,
+            package_name: src_pkg_obj.name,
+            project_name: src_prj_obj.name
+          }
+        end
+      ).flatten
+    end
+
+    raw_data
```
krauselukas, line 43:
```suggestion
      raw_data.push(
        action_build_results.map do |result_entry|
          {
            architecture: result_entry.architecture,
            repository: result_entry.repository,
            status: result_entry.code,
            package_name: src_pkg_obj.name,
            project_name: src_prj_obj.name
          }
        end
      )
    end

    raw_data.flatten
```

krauselukas: @ncounter I think calling `flatten` once after everything is collected in the `raw_data` array would be enough right?

ncounter: That is true, fixing it :+1: thanks

rubhanazeem:
To address code climate complaints I've added this card https://trello.com/c/heYrm34S/164-code-climate-issues-caused-by-build-results-summary

rubhanazeem (approved):

PR #14860, opened 2023-09-01
[api] fix for token remained active when user was not in confirmed st…

adrianschroeter:
…ate.


adrianschroeter:
fortunatly not an issue in 2.10, so no security anouncement needed.

adrianschroeter:
nah, got tricked by the owner...

PR #14866, opened 2023-09-04
Remove no longer needed cassette file

eduardoj:


eduardoj:
This file is still needed for non-beta tests. Closing.

PR #14878, opened 2023-09-07
Avoid sending empty partition metric

saraycp:
Telegraf can't parse it when partition is empty for 'backend_daemon_status'. It will be 'main' in such cases.


bmwiedemann (approved):

saraycp:
Ignoring codecov/patch  and merging...

PR #14880, opened 2023-09-07
Initial implementation of reports feature

rubhanazeem:
This PR covers the implementation of the following things:

- [x]  Implementation of the Report model and its associations
- [x]  Implementation of flipper flag
- [x] Implementation of reporting a comment
- [x] Implementation of reporting a package
- [x] Implementation of reporting a user

There will be a follow-up PR #14898 for the remaining features i.e., 
- [Specs for report policy](https://github.com/openSUSE/open-build-service/pull/14909)
- [Implementation of reporting a project](https://github.com/openSUSE/open-build-service/pull/14898)
- ~Implementation of report listing view~

### DIFF
--- a/src/api/app/controllers/webui/reports_controller.rb
+++ b/src/api/app/controllers/webui/reports_controller.rb
@@ -1,0 +1,29 @@
+class Webui::ReportsController < Webui::WebuiController
+  before_action :require_login
+  after_action :verify_authorized
+
+  def create
+    user = User.session!
+    report = user.submitted_reports.new(report_params)
+    authorize report
+
+    @link_id = params[:link_id]
+    @modal_id = params[:modal_id]
+
+    if report.save
+      flash[:success] = "#{report.reportable_type} reported successfully"
+    else
+      flash[:error] = report.errors.full_messages.to_sentence
+    end
+
+    respond_to do |format|
+      format.js
+    end
+  end
+
+  private
+
+  def report_params
+    params.require(:report).permit(:reason, :reportable_id, :reportable_type)
+  end
+end
--- a/src/api/app/models/report.rb
+++ b/src/api/app/models/report.rb
@@ -1,0 +1,30 @@
+# Report class flags abusive content, be it projects, packages, users or comments
+class Report < ApplicationRecord
+  validates :reason, length: { maximum: 65_535 }
+  validates :reportable_type, length: { maximum: 255 }
+
+  belongs_to :user, optional: false
+  belongs_to :reportable, polymorphic: true, optional: false
+end
+
+# == Schema Information
+#
+# Table name: reports
+#
+#  id              :bigint           not null, primary key
+#  reason          :text(65535)
+#  reportable_type :string(255)      not null, indexed => [reportable_id]
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  reportable_id   :integer          not null, indexed => [reportable_type]
+#  user_id         :integer          not null, indexed
+#
+# Indexes
+#
+#  index_reports_on_reportable  (reportable_type,reportable_id)
+#  index_reports_on_user_id     (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
--- a/src/api/app/policies/report_policy.rb
+++ b/src/api/app/policies/report_policy.rb
@@ -1,0 +1,34 @@
+class ReportPolicy < ApplicationPolicy
+  attr_reader :user, :record, :reportable
+
+  def initialize(user, record)
+    super(user, record)
+
+    @user = user
+    @record = record
+    @reportable = record.reportable
+  end
+
+  def show?
+    user.is_admin? || record.user == user
+  end
+
+  def create?
+    return false unless Flipper.enabled?(:content_moderation, user)
+
+    # We don't want reports twice...
+    return false if user.submitted_reports.where(reportable: reportable).any?
+
+    # We don't want reports for things you can change yourself...
+    case reportable.class.name
+    when 'Package'
+      !PackagePolicy.new(user, reportable).update?
+    when 'Project'
+      !ProjectPolicy.new(user, reportable).update?
+    when 'Comment'
+      !CommentPolicy.new(user, reportable).update?
+    when 'User'
+      !UserPolicy.new(user, reportable).update?
+    end
+  end
+end
--- a/src/api/config/initializers/flipper.rb
+++ b/src/api/config/initializers/flipper.rb
@@ -2,7 +2,8 @@
 # environment and review apps to test changes for those features. Rolled-out feature toggles aren't displayed anyway
 # to users in the web UI for beta features.
 ENABLED_FEATURE_TOGGLES = [
-  { name: :request_show_redesign, description: 'Redesign of the request pages to improve the collaboration workflow' }
+  { name: :request_show_redesign, description: 'Redesign of the request pages to improve the collaboration workflow' },
+  { name: :content_moderation, description: 'Reporting inappropriate content' }
 ].freeze
 
 Flipper.configure do
--- a/src/api/config/routes/webui_routes.rb
+++ b/src/api/config/routes/webui_routes.rb
@@ -428,4 +428,6 @@ OBSApi::Application.routes.draw do
       end
     end
   end
+
+  resources :reports, only: [:create], controller: 'webui/reports'
 end
--- a/src/api/db/schema.rb
+++ b/src/api/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_07_27_110437) do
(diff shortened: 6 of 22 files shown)


src/api/config/initializers/flipper.rb
```diff
@@ -2,7 +2,8 @@
 # environment and review apps to test changes for those features. Rolled-out feature toggles aren't displayed anyway
 # to users in the web UI for beta features.
 ENABLED_FEATURE_TOGGLES = [
-  { name: :request_show_redesign, description: 'Redesign of the request pages to improve the collaboration workflow' }
+  { name: :request_show_redesign, description: 'Redesign of the request pages to improve the collaboration workflow' },
+  { name: :content_moderation, description: 'Moderating the different types of content submitted to the OBS' }
```
hennevogel, line 6:
```suggestion
  { name: :content_moderation, description: 'Reporting abusive content' }
```

hennevogel (changes requested):

src/api/app/models/report.rb
```diff
@@ -0,0 +1,27 @@
+# TODO: Please overwrite this comment with something explaining the model target
```
danidoni, line 1:
```suggestion
# Report is used to flag abusive content, be it Projects, Packages or Comments
```

src/api/app/views/webui/comment/_content.html.haml
```diff
@@ -8,6 +8,15 @@
       wrote
       = link_to("#comment-#{comment.id}", title: I18n.l(comment.created_at.utc), name: "comment-#{comment.id}") do
         = render TimeComponent.new(time: comment.created_at)
+
+      - if User.session && User.session.submitted_reports.where(reportable: comment).empty?
```
danidoni, line 12: We can encapsulate this line in a helper because we're using it for packages also

rubhanazeem: Done. This logic is now part of ReportPolicy

src/api/app/policies/report_policy.rb
```diff
@@ -0,0 +1,28 @@
...
+    user.is_admin? || record.user == user
+  end
+
+  def submit_report?
+    not_self = if record.instance_of?(::Comment)
+                 record.user != user
+               elsif record.instance_of?(::User)
+                 record != user
+               elsif record.instance_of?(::Package)
+                 record.project.users.exclude?(user)
```
hennevogel, line 21: This should be limited to the maintainer role, same for project.

src/api/app/views/webui/comment/_content.html.haml
```diff
@@ -8,6 +8,15 @@
...
+        = link_to('#', class: 'badge badge-pill badge-light text-muted', id: "js-comment-#{comment.id}",
```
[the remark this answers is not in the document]
rubhanazeem: No specific reason for that. I've addressed this issue and now it's like other clickables

src/api/.database_consistency.todo.yml
```diff
@@ -2104,3 +2104,13 @@ WorkflowRun:
...
+Report:
+  id:
+    PrimaryKeyTypeChecker:
+      enabled: false
+  reportable_type:
+    LengthConstraintChecker:
+      enabled: false
+  reason:
+    LengthConstraintChecker:
+      enabled: false
```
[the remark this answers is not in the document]
rubhanazeem: Done. This is addressed

src/api/app/policies/report_policy.rb
```diff
@@ -0,0 +1,28 @@
...
+               elsif record.instance_of?(::Project)
+                 record.users.exclude?(user)
+               end
```
hennevogel, line 24: 💭  I also think you should move this into it's own private method...

src/api/app/policies/report_policy.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    @user = user
+    @record = record
+  end
+
+  def show?
+    user.is_admin? || record.user == user
+  end
+
+  def submit_report?
```
hennevogel, line 15: This is `create?`...

src/api/app/controllers/webui/reports_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Webui::ReportsController < Webui::WebuiController
+  before_action :require_login
+  after_action :verify_authorized, except: [:create]
+
+  def create
+    user = User.session!
+    report = user.submitted_reports.new(report_params)
```
hennevogel, line 7: You should authorize here.

hennevogel (changes requested):

src/api/app/policies/report_policy.rb
rubhanazeem: Do you mean maintainers should be able to report the projects?

hennevogel: Ah sorry I've read `include?`....

Why would a a user of the package/project not be able to report it?

rubhanazeem: Same reason as `comments` and `users`. I think it's not logical to add a comment and then report it yourself Or report your own `user` profile.

hennevogel: For comments and users sure, but *inside* a group of people that work together (on a package or project) there can be abuse for instance.

rubhanazeem: That makes sense. I got it :+1:

src/api/app/controllers/webui/reports_controller.rb
rubhanazeem: done

src/api/app/policies/report_policy.rb
rubhanazeem: Just to check inclusion in maintainers? This is just a one-liner and not used anywhere else.  I've updated this method and there are new changes. Please have a look again

src/api/app/policies/report_policy.rb
rubhanazeem: I've updated this policy. Please check again

src/api/app/policies/report_policy.rb
hennevogel: What do you mean by "just"? Instead of 

```
 not_self = if record.instance_of?(::Comment)
                 record.user != user
               elsif record.instance_of?(::User)
                 record != user
               end
...
``` 

just create a private method by that name. Much more readable.

```
private

def not_self
  if record.instance_of?(::Comment)
    return record.user != user
  elsif record.instance_of?(::User)
    record != user
  end
end
```

which is much more readable IMHO

src/api/app/models/report.rb
```diff
@@ -0,0 +1,30 @@
+# Report class flags abusive content, be it projects, packages, users or comments
+class Report < ApplicationRecord
+  validates :reason, length: { maximum: 65_535 }
+  validates :reportable_type, length: { maximum: 255 }
+
+  belongs_to :user, optional: false
```
danidoni, line 6: A better name would be `reporter`. What do you think?

src/api/db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_07_27_110437) do
+ActiveRecord::Schema[7.0].define(version: 202302224114624) do
```
danidoni, line 13: This timestamp is a bit weird, February 2023?

src/api/app/models/report.rb
danidoni: We agreed offline to do it some other day

src/api/app/policies/report_policy.rb
rubhanazeem: We don't need it anymore. This code has been replaced by other policy classes

src/api/db/schema.rb
rubhanazeem: Ah nice catch. This is reverted

src/api/config/routes/webui_routes.rb
```diff
@@ -428,4 +428,6 @@
       end
     end
   end
+
+  resources :reports, except: [:new], controller: 'webui/reports'
```
hennevogel, line 432: You also don't have any other of the default actions. You `only: [:create]`

rubhanazeem: Right. We won't be using other routes for reports. This is updated

src/api/app/controllers/webui/reports_controller.rb
```diff
@@ -0,0 +1,29 @@
+class Webui::ReportsController < Webui::WebuiController
+  before_action :require_login
+  after_action :verify_authorized, except: [:create]
```
hennevogel, line 3: Hm, why?

rubhanazeem: This is not correct. Fallout from the mob session and didn't realize it was still skipped

hennevogel (approved):

danidoni (approved):

Issue #14881, opened 2023-09-07
Invalid assumptions in package name finder for diffs


darix:
for a package in a maintenance project with the name `python-ruamel.yaml.clib.SUSE_SLE-15-SP4_Update` it runs the diff against `python-ruamel.yaml`

for `python-backports.entry_points_selectable.SUSE_SLE-15-SP4_Update`   against `python-backports`.

searching just from the back is also invalid as `python-ruamel.yaml.clib.openSUSE_Leap_15.5`

it should be noted that we had a similar bug in the backend not that long ago. there mls added code that would just strip dot elements from the back until he finds a valid file.

darix:
one easy solution might be that we store the clean package name somewhere in the DB when creating the special maintenance package names. that saves us guessing later.

neither the package name nor the repository name disallow dots. so this really just becomes guessing unless we store the clean name.

darix:
backend commit was bff1ab8a879

hennevogel:
We pass `python-ruamel.yaml.clib.SUSE_SLE-15-SP4_Update` into

https://github.com/openSUSE/open-build-service/blob/be43f75f8af5aeed90d452a7315c0f41f99029f8/src/api/app/models/bs_request_action/differ/query_builder.rb#L61

which looks up the linkinfo of that package. The package links to `SOME:SP:Update / python-ruamel.yaml.clib`. So we then run `python-ruamel.yaml.clib` through

https://github.com/openSUSE/open-build-service/blob/be43f75f8af5aeed90d452a7315c0f41f99029f8/src/api/app/models/bs_request_action/differ/query_builder.rb#L20

which turns `python-ruamel.yaml.clib` into `python-ruamel.yaml`. 

`python-ruamel.yaml` by change also exists in SUSE:SLE-15-SP4:Update so we diff two unrelated packages.

hennevogel:
I don't really get the why `check_for_local_linked_packages` sets `target_package` here

https://github.com/openSUSE/open-build-service/blob/be43f75f8af5aeed90d452a7315c0f41f99029f8/src/api/app/models/bs_request_action/differ/query_builder.rb#L74-L76

@adrianschroeter can you imagine why?

adrianschroeter:
well, it is the default target usually.

The naming of the method makes indeed no sense, that was introduced by some refactoring.

But the base problem is that the releasename is ignored at all. You can not guess the target of open maintenance_incident actions neither by link nor by package name substrings.

[...]

PR #14885, opened 2023-09-08
[api] request action diffing

adrianschroeter:
Take possible defined release name into account as fallback target name.

This is important for open (not accepted) maintenance_incident request actions where no target_package is defined.

The package or the abstracted project name suffix both may contain dots and can not be used therefore as fallback name directly.

Reported by @darix and as @dmach rightfully pointed to the releasename support to solve it correctly.


hennevogel:
Is this about #14881 or about something else?

PR #14888, opened 2023-09-11
Improve token creation error message

### LINKED ISSUE #14828: Please improve the error message when creating a token with a project parameter but without a package parameter

The following API call leads to a suboptimal error message:
```ShellSession
❯ osc api -X POST "/person/dancermak/token?cmd=create&project=home:dancermak"
Package not found: home:dancermak/
```

An error message like:
```
The package parameter must be provided in conjunction with the project parameter
```

mfriedrichgh:
Fixes #14828

### DIFF
--- a/src/api/app/controllers/person/token_controller.rb
+++ b/src/api/app/controllers/person/token_controller.rb
@@ -14,9 +14,9 @@ module Person
     def create
       authorize @user, :update?
 
-      pkg = (Package.get_by_project_and_name(params[:project], params[:package]) if params[:project] || params[:package])
+      set_package
 
-      @token = Token.token_type(params[:operation]).create(description: params[:description], executor: @user, package: pkg, scm_token: params[:scm_token])
+      @token = Token.token_type(params[:operation]).create(description: params[:description], executor: @user, package: @package, scm_token: params[:scm_token])
       return if @token.valid?
 
       render_error status: 400,
@@ -40,6 +40,15 @@ module Person
 
     def set_user
       @user = User.find_by(login: params[:login]) || User.find_nobody!
+    end
+
+    def set_package
+      @package = nil
+      return unless (params[:project] || params[:package]) && params[:operation] != 'workflow'
+
+      raise MissingParameterError, 'The package and project parameters must be provided together.' unless params[:project] && params[:package]
+
+      @package = Package.get_by_project_and_name(params[:project], params[:package])
     end
   end
 end
--- a/src/api/spec/controllers/person/token_controller_spec.rb
+++ b/src/api/spec/controllers/person/token_controller_spec.rb
@@ -53,6 +53,19 @@ RSpec.describe Person::TokenController, vcr: false do
       end
     end
 
+    context 'called with only project parameter' do
+      before do
+        login user
+      end
+
+      subject { post :create, params: { login: user.login, package: user.home_project, operation: 'rebuild' }, format: :xml }
+
+      it 'does not create a token' do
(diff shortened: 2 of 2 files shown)


src/api/app/controllers/person/token_controller.rb
```diff
@@ -14,7 +14,12 @@ def index
     def create
       authorize @user, :update?
 
-      pkg = (Package.get_by_project_and_name(params[:project], params[:package]) if params[:project] || params[:package])
+      unless params[:project] && params[:package]
+        render_error status: 400, errorcode: 'missing_parameters', message: 'The package and project parameters must be provided.'
```
krauselukas, line 18: Hey @mfriedrichgh we have a exception class for this :) so you can use it here https://github.com/openSUSE/open-build-service/blob/ee8f92cfc8134dafe3d541bec26a72588ccd9a91/src/api/lib/api_error.rb#L84

krauselukas, line 18:
```suggestion
       raise MissingParameterError, 'The package and project parameters must be provided.'
```

krauselukas: Let me know if something is not clear

mfriedrichgh: Everything clear, thanks for the suggestion.

krauselukas (changes requested):
@mfriedrichgh I just realized that you can create a token without a package or project assigned. Thats why you have a failing test now https://app.circleci.com/pipelines/github/openSUSE/open-build-service/14039/workflows/7444e3f0-8114-483f-8464-40b35203a22c/jobs/145324?invite=true#step-104-182983_103
So basically you would need to check if `project && package` params are given or non of them.

src/api/spec/controllers/person/token_controller_spec.rb
```diff
@@ -53,6 +53,19 @@
...
     end
 
+    context 'called with only project parameter' do
+      before do
+        login user
+      end
+
+      subject { post :create, params: { login: user.login, package: user.home_project, operation: 'rebuild' }, format: :xml }
+
+      it 'creates a global token' do
```
krauselukas, line 63:
```suggestion
      it 'does not create a token' do
```

src/api/public/apidocs/paths/person_login_token.yaml
```diff
@@ -68,19 +68,19 @@ post:
       example: Rebuild gchz
     - in: query
       name: project
+      required: true
       schema:
         type: string
       description: |
-        Project name. Should be provided, together with the package query parameter,
-        to limit the token to a specific package.
+        Project name. Limits the token to a specific project.
```
krauselukas, line 75: Hmm, I would say we should not mark the project and package parameters as required. You can create a global token without providing those parameters. The description was correct before I would say

mfriedrichgh: Agreed, the edit made before doesn't consider the ability to create token without parameters.

src/api/app/controllers/person/token_controller.rb
```diff
@@ -2,6 +2,7 @@
 module Person
   class TokenController < ApplicationController
     before_action :set_user
+    before_action :set_package, only: [:create]
```
eduardoj, line 5: Adding this `before_action` here would call `set_package` even the user is not authorized.

Before the changes of this pull request, checking for the permissions of the user was done before searching for the package. Could we keep this behaviour?

danidoni (approved):

PR #14892, opened 2023-09-11
Use `DeleteConfirmationDialogComponent` in users index view

krauselukas:
We created this view component to avoid having multiple dialog partials for the same thing. Lets use it here as well.


src/api/app/views/webui/users/index.html.haml
```diff
@@ -23,7 +23,11 @@
       = link_to(new_user_path(pagetitle: 'create user', submit_btn_text: 'create', title: 'Create User')) do
         %i.fas.fa-plus-circle.text-primary
         Create User
-    = render partial: 'webui/users/delete_dialog'
+    - confirmation_text = "Please confirm deletion of '#{content_tag(:span, '', class: 'user-login')}'"
+    = render DeleteConfirmationDialogComponent.new(modal_id: 'delete-user-modal',
+                                                   method: :delete,
+                                                   options: { modal_title: 'Delete user?', remote: false,
+                                                              confirmation_text: sanitize(confirmation_text) })
```
eduardoj, line 30:
```suggestion
                                                              confirmation_text: sanitize(confirmation_text, tags: ['span']) })
```
The `confirmation_text` variable will never vary. And we only want to sanitize the `span` tag.

eduardoj (changes requested):

src/api/app/views/webui/users/index.html.haml
krauselukas: @eduardoj thanks, done

eduardoj (approved):

PR #14898, opened 2023-09-13
Report projects

danidoni:
Follow-up on #14880 

Adds support for reporting projects from the left sidebar

image

### DIFF
--- a/src/api/app/views/webui/project/show.html.haml
+++ b/src/api/app/views/webui/project/show.html.haml
@@ -86,5 +86,8 @@
                                                               confirmation_text: "Please confirm deletion of the project '#{@project}'",
                                                               action: project_destroy_path(project: @project) })
 
+- if policy(Report.new(reportable: @project)).create?
+  = render ReportComponent.new(modal_id: 'report-project-modal', options: { object_type: 'Project', user: User.session })
+
 - content_for :ready_function do
   initializeDataTable('#inherited-packages-table');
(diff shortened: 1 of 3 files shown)


src/api/app/views/webui/project/show.html.haml
```diff
@@ -86,5 +86,8 @@
                                                               confirmation_text: "Please confirm deletion of the project '#{@project}'",
                                                               action: project_destroy_path(project: @project) })
 
+- if policy(Report.new(reportable: @project).create?
```
hennevogel, line 89: I think we should move this into `ReportComponent.render`..

src/api/.haml-lint.yml
```diff
@@ -18,6 +18,8 @@ linters:
     enabled: true
     exclude:
       - 'app/views/webui/users/new.html.haml'
+      # Workaround bug: Couldn't process the file for linting. NoMethodError: undefined method `each_node' for nil:NilClass
```
hennevogel, line 21: How are you tracking this "bug"?

danidoni: In the end it was not a bug. There was a syntax error somewhere else and the linter was choking on it

src/api/app/views/webui/project/show.html.haml
danidoni: Hm, yeah, I'll write a PR later to move the policy checks into the component. Works for you?

hennevogel: sure, where do you track this?

danidoni: Here: https://trello.com/c/lRN12PZK/31-move-the-policy-checks-inside-the-report-component

danidoni: Well, that check is controlling if the "button" to report content shows or not, so either we create a reporting button component and move the check there, or moving it inside the `ReportComponent.render` won't work...

danidoni: Alright, we'll move the one affecting the component

hennevogel (approved):

PR #14899, opened 2023-09-13
[backend] handle empty debian repo case

bugfinder:


bugfinder:
dropped for other solution from mls

PR #14900, opened 2023-09-13
[dist] fix can not get full hostname in functions such as abc.site.com

YangMame:
`hostname -f` didn't show full hostname with "." such as abc.site.com, it'll shows abc, this will case vist problem

### DIFF
--- a/dist/functions.setup-appliance.sh
+++ b/dist/functions.setup-appliance.sh
@@ -143,12 +143,21 @@ function get_hostname {
   else
     TIMEOUT=30
     while [ -z "$FQHOSTNAME" -o "$FQHOSTNAME" = "localhost" ];do
-      FQHOSTNAME=`hostname -f 2>/dev/null`
+      # Try to get the FQHN via hostname
+      HN=`hostname -f 2>/dev/null`
+      if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]];then
+        FQHOSTNAME=$HN
+      fi
       TIMEOUT=$(($TIMEOUT-1))
       [ "$TIMEOUT" -le 0 ] && break
       echo "Waiting for FQHOSTNAME ($TIMEOUT)"
       sleep 1
     done
+    if [ -z "$FQHOSTNAME" -o "$FQHOSTNAME" = "localhost" ] && [ -x "$(command -v hostnamectl)" ];then
+      FQHOSTNAME=`hostnamectl --static 2>/dev/null`
+    elif [ -s /etc/hostname ];then
+      FQHOSTNAME=`cat /etc/hostname`
+    fi
   fi
 
   if type -p ec2-public-hostname; then


YangMame:
wicked cant get full hostname, NetworkManager is works fines

YangMame:
ping

M0ses:
Unfortunatly this might cause regressions for other users. 

```
obs-server:~ # hostname -f
obs-server.kanku.site
obs-server:~ # hostnamectl --static
obs-server
obs-server:~ # 
```

YangMame:
> Unfortunatly this might cause regressions for other users.
> 
> ```
> obs-server:~ # hostname -f
> obs-server.kanku.site
> obs-server:~ # hostnamectl --static
> obs-server
> obs-server:~ # 
> ```

how about check if exist `/etc/hostname` and read it, then use `hostname -f`?

M0ses:
This all depends on your lokal network settings and configured services like your dns or dhcp server.
My guess - you are using dnsmasq with dhcp and no "domain ..." option?

YangMame:
> This all depends on your lokal network settings and configured services like your dns or dhcp server. My guess - you are using dnsmasq with dhcp and no "domain ..." option?

the environment was destoryed, it seems should not be no "domain" option and other else.
but the network is unsable and not under control by me. is it fine to just read `/etc/hostname`?

M0ses:
Your patch would break 99% of our appliance tests as we normally set only the shortname in /etc/hostname and let dhcp handle the rest.
Here my (untested) approach. If it also fullfills also your needs, please feel free to include in the PR.


```bash
# Set default to localhost
FQHOSTNAME="localhost"

# Try to get the FQHN via hostname
HN=`hostname -f 2>/dev/null`

if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]]
then
  FQHOSTNAME=$HN
else
  HN=`hostnamectl --static 2>/dev/null`
  if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]]
  then
    FQHOSTNAME=$HN
  else
    HN=`cat /etc/hostname 2>/dev/null`
    [[ -n "$HN" ]] && FQHOSTNAME=$HN
  fi
fi
```

YangMame:
> Your patch would break 99% of our appliance tests as we normally set only the shortname in /etc/hostname and let dhcp handle the rest. Here my (untested) approach. If it also fullfills also your needs, please feel free to include in the PR.
> 
> ```shell
> # Set default to localhost
> FQHOSTNAME="localhost"
> 
> # Try to get the FQHN via hostname
> HN=`hostname -f 2>/dev/null`
> 
> if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]]
> then
>   FQHOSTNAME=$HN
> else
>   HN=`hostnamectl --static 2>/dev/null`
>   if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]]
>   then
>     FQHOSTNAME=$HN
>   else
>     HN=`cat /etc/hostname 2>/dev/null`
>     [[ -n "$HN" ]] && FQHOSTNAME=$HN
>   fi
> fi
> ```

wow! this one will works fine, thanks a lot!

YangMame:
> Your patch would break 99% of our appliance tests as we normally set only the shortname in /etc/hostname and let dhcp handle the rest. Here my (untested) approach. If it also fullfills also your needs, please feel free to include in the PR.
> 
> ```shell
> # Set default to localhost
> FQHOSTNAME="localhost"
> 
> # Try to get the FQHN via hostname
> HN=`hostname -f 2>/dev/null`
> 
> if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]]
> then
>   FQHOSTNAME=$HN
> else
>   HN=`hostnamectl --static 2>/dev/null`
>   if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]]
>   then
>     FQHOSTNAME=$HN
>   else
>     HN=`cat /etc/hostname 2>/dev/null`
>     [[ -n "$HN" ]] && FQHOSTNAME=$HN
>   fi
> fi
> ```

I've changed it to get hostname by `hostname -f`, `hostnamectl --static` or read form `/etc/hostname`, if failed, use localhost as default

dist/functions.setup-appliance.sh
```diff
@@ -143,12 +143,27 @@ function get_hostname {
...
+        fi
+      fi
       TIMEOUT=$(($TIMEOUT-1))
       [ "$TIMEOUT" -le 0 ] && break
       echo "Waiting for FQHOSTNAME ($TIMEOUT)"
       sleep 1
     done
+
+    # Set default to localhost
+    FQHOSTNAME="localhost"
```
M0ses, line 166: This would break all functions below which are checking if FQHOSTNAME is already set.
"localhost" is really the worst choice and should only be taken if we could not find anything else.

dist/functions.setup-appliance.sh
```diff
@@ -143,12 +143,27 @@ function get_hostname {
...
-      FQHOSTNAME=`hostname -f 2>/dev/null`
+      # Try to get the FQHN via hostname
+      HN=`hostname -f 2>/dev/null`
+      if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]];then
+        FQHOSTNAME=$HN
+      else
+        HN=`hostnamectl --static 2>/dev/null`
+        if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]];then
+          FQHOSTNAME=$HN
+        else
```
M0ses, line 154: /etc/hostname should be not checked in the loop - only if we run into a timeout.

I`m sorry this was my fault in my suggestion.

dist/functions.setup-appliance.sh
M0ses: See line 173/188 - Here it gets set to localhost if nothing else could be found. Its better to use an IP-Address than "localhost" - then the appliance could be reached at least by IP from the network and has a valid self-signed cert.

dist/functions.setup-appliance.sh
```diff
@@ -143,12 +143,19 @@ function get_hostname {
...
+      if [[ $HN =~ ^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)+$ ]];then
+        FQHOSTNAME=$HN
+      fi
       TIMEOUT=$(($TIMEOUT-1))
       [ "$TIMEOUT" -le 0 ] && break
       echo "Waiting for FQHOSTNAME ($TIMEOUT)"
       sleep 1
     done
+    if [ -z "$FQHOSTNAME" -o "$FQHOSTNAME" = "localhost" ];then
+      FQHOSTNAME=`hostnamectl --static 2>/dev/null`
```
M0ses, line 157: Are you sure that you want to accept the return value of `hostnamectl --static` even if it does not match a FQHN?
And I do not see a fallback to /etc/hostname
Might be useful  if hostnamectl is not installed.

M0ses (changes requested):
Please see my inline comments

dist/functions.setup-appliance.sh
YangMame: i remove `cat /etc/hostname` was because i found hostnamectl is reading it and it's systemd's func, but systemd may not be included in container image or not using systemd as init after you remind, so you're right

M0ses:
LGTM - I will build an OBS-Appliance containing a obs-server package based on your patches - After passing our tests, I will merge it.
(https://build.opensuse.org/project/show/home:M0ses:ToTest:PR14900)

M0ses (approved):
passed relevant kanku tests

M0ses:
Thanks for your contribution

YangMame:
> Thanks for your contribution

thank you for your patience :)

PR #14907, opened 2023-09-14
Handle moderation on comments

saraycp:
We are introducing moderation on comments.

Moderators will be able to flag those comments, so a text like _"this comment was considered problematic ..."_ is going to be displayed instead of the original message.

moderate

### DIFF
--- a/src/api/app/components/bs_request_comment_component.html.haml
+++ b/src/api/app/components/bs_request_comment_component.html.haml
@@ -17,6 +17,9 @@
             = link_to("#edit_form_of_#{comment.id}", id: "edit_button_of_#{comment.id}", 'data-bs-toggle': 'collapse',
                       class: 'dropdown-item collapsed') do
               Edit
+          - if policy(comment).moderate?
+            = button_to(comment.moderated? ? 'Unmoderate' : 'Moderate', comment_moderate_path(comment),
+                        class: 'dropdown-item', remote: true, form_class: 'moderate-form', params: { moderation_state: !comment.moderated? })
           - if policy(comment).destroy?
             = link_to('#', data: { 'bs-toggle': 'modal',
                                    'bs-target': '#delete-comment-modal',
--- a/src/api/app/components/bs_request_comment_component.html.haml
+++ b/src/api/app/components/bs_request_comment_component.html.haml
@@ -17,9 +17,10 @@
             = link_to("#edit_form_of_#{comment.id}", id: "edit_button_of_#{comment.id}", 'data-bs-toggle': 'collapse',
                       class: 'dropdown-item collapsed') do
               Edit
-          - if policy(comment).moderate?
-            = button_to(comment.moderated? ? 'Unmoderate' : 'Moderate', comment_moderate_path(comment),
-                        class: 'dropdown-item', remote: true, form_class: 'moderate-form', params: { moderation_state: !comment.moderated? })
+          - if Flipper.enabled?(:content_moderation, User.session)
+            - if policy(comment).moderate?
+              = button_to(comment.moderated? ? 'Unmoderate' : 'Moderate', comment_moderate_path(comment),
+                          class: 'dropdown-item', remote: true, form_class: 'moderate-form', params: { moderation_state: !comment.moderated? })
           - if policy(comment).destroy?
             = link_to('#', data: { 'bs-toggle': 'modal',
                                    'bs-target': '#delete-comment-modal',
--- a/src/api/app/components/bs_request_comment_component.html.haml
+++ b/src/api/app/components/bs_request_comment_component.html.haml
@@ -19,7 +19,7 @@
               Edit
           - if Flipper.enabled?(:content_moderation, User.session)
             - if policy(comment).moderate?
-              = button_to(comment.moderated? ? 'Permit' : 'Moderate', comment_moderate_path(comment),
+              = button_to(comment.moderated? ? 'Permit' : 'Moderate', moderate_comment_path(comment),
                           class: 'dropdown-item', remote: true, form_class: 'moderate-form', params: { moderation_state: !comment.moderated? })
           - if policy(comment).destroy?
             = link_to('#', data: { 'bs-toggle': 'modal',
--- a/src/api/app/components/bs_request_comment_component.html.haml
+++ b/src/api/app/components/bs_request_comment_component.html.haml
@@ -19,7 +19,7 @@
               Edit
           - if Flipper.enabled?(:content_moderation, User.session)
             - if policy(comment).moderate?
-              = button_to(comment.moderated? ? 'Unmoderate' : 'Moderate', comment_moderate_path(comment),
+              = button_to(comment.moderated? ? 'Permit' : 'Moderate', comment_moderate_path(comment),
                           class: 'dropdown-item', remote: true, form_class: 'moderate-form', params: { moderation_state: !comment.moderated? })
           - if policy(comment).destroy?
             = link_to('#', data: { 'bs-toggle': 'modal',
--- a/src/api/app/controllers/webui/comments_controller.rb
+++ b/src/api/app/controllers/webui/comments_controller.rb
@@ -137,6 +137,7 @@ class Webui::CommentsController < Webui::WebuiController
     params.require(:comment).permit(:body, :parent_id, :diff_ref)
   end
 
+  # FIXME: Use this function for the rest of the actions
   def set_comment
     @comment = Comment.find(params[:id] || params[:comment_id])
   rescue ActiveRecord::RecordNotFound => e
--- a/src/api/app/controllers/webui/comments_controller.rb
+++ b/src/api/app/controllers/webui/comments_controller.rb
@@ -139,9 +139,8 @@ class Webui::CommentsController < Webui::WebuiController
 
   def set_comment
     @comment = Comment.find(params[:id] || params[:comment_id])
-    return if @comment.present?
-
-    flash[:error] = "Comment with id '#{params[:id]}' doesn't exist."
+  rescue ActiveRecord::RecordNotFound => e
+    flash[:error] = e.message
     render partial: 'layouts/webui/flash'
   end
 
--- a/src/api/app/controllers/webui/comments_controller.rb
+++ b/src/api/app/controllers/webui/comments_controller.rb
@@ -1,6 +1,7 @@
 class Webui::CommentsController < Webui::WebuiController
   before_action :require_login
   before_action :set_commented, only: :create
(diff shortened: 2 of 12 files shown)


src/api/app/models/comment.rb
```diff
@@ -51,6 +51,16 @@ def unused_parent?
...
 
+  def moderated?
+    moderated_at || moderated_by
+  end
+
+  def moderate(state)
+    moderated_at = state ? Time.now : nil
+    moderated_by = state ? User.session : nil
+    save!
+  end
```
hennevogel, line 62: https://github.com/pienkowb/active_record-events

src/api/app/policies/comment_policy.rb
```diff
@@ -31,4 +31,11 @@ def update?
   def reply?
     !(user.blank? || user.is_nobody? || record.user.is_nobody?)
   end
+
+  # Only logged-in Admins can moderate non-deleted comments
+  def moderate?
+    return false if user.blank? || user.is_nobody? || record.user.is_nobody?
+
+    user.is_admin?
```
hennevogel, line 39:
```suggestion
    return true if user.try(:is_admin?)
    
    false
```

hennevogel (changes requested):

src/api/app/policies/comment_policy.rb
saraycp: Alright, I applied your changes but added `&& !record.user.is_nobody?` to the condition to avoid moderating on comments that were soft-deleted already.

src/api/app/models/comment.rb
saraycp: @hennevogel we wonder if it's worth introducing the gem to handle `moderated_at` as we still have to handle `moderator_id` manually anyway...

src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -105,12 +106,45 @@ def preview
...
 
   def permitted_params
     params.require(:comment).permit(:body, :parent_id, :diff_ref)
   end
 
+  def set_comment
+    @comment = Comment.find(params[:id] || params[:comment_id])
+    return if @comment.present?
+
+    flash[:error] = "Comment with id '#{params[:id]}' doesn't exist."
```
danidoni, line 144: Why don't we use the rails way of dealing with nonexistent resources in the controller?

src/api/app/models/comment.rb
hennevogel: I just wanted to point it out, you decide.

src/api/app/models/comment.rb
```diff
@@ -3,6 +3,7 @@
 class Comment < ApplicationRecord
   belongs_to :commentable, polymorphic: true # belongs to a Project, Package, BsRequest or BsRequestActionSubmit
   belongs_to :user, inverse_of: :comments
+  belongs_to :moderator, class_name: 'User', optional: true
```
hennevogel, line 6: Hm, why is this optional?

src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -1,6 +1,7 @@
 class Webui::CommentsController < Webui::WebuiController
   before_action :require_login
   before_action :set_commented, only: :create
+  before_action :set_comment, only: :moderate
```
hennevogel, line 4: Why not use it on the other comment related operations too?

src/api/app/views/webui/comment/_reply.html.haml
```diff
@@ -15,6 +15,10 @@
       = link_to('#', data: { 'bs-toggle': 'modal', 'bs-target': "#delete-comment-modal-#{comment.id}" },
                 class: 'delete_link btn btn-outline-danger', title: 'Delete comment') do
         Delete
+  - if policy(comment).moderate?
```
hennevogel, line 18: You need to put this behind the feature flag.

hennevogel (changes requested):

src/api/app/models/comment.rb
saraycp: Because it's not compulsory to always have a moderator associated with the comment.

saraycp: Double-checking... 

Indeed. If I remove `optional: true` and try to create a comment without specifying a moderator I get:

```Validation failed: Moderator must exist
```

src/api/app/controllers/webui/comments_controller.rb
danidoni: Yes

src/api/app/controllers/webui/comments_controller.rb
hennevogel: If it's out of scope you need a follow up card or at least a FIXME...

src/api/app/policies/comment_policy.rb
```diff
@@ -31,4 +31,11 @@ def update?
   def reply?
     !(user.blank? || user.is_nobody? || record.user.is_nobody?)
   end
+
+  # Only logged-in Admins can moderate non-deleted comments
```
hennevogel, line 35: This makes little sense I guess. No can should be able to moderate soft-deleted comments...

src/api/app/policies/comment_policy.rb
```diff
@@ -31,4 +31,11 @@ def update?
...
+  def moderate?
+    return true if user.try(:is_admin?) && !record.user.is_nobody?
```
hennevogel, line 37: What about `Staff`?

saraycp: Added

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -17,6 +17,10 @@
             = link_to("#edit_form_of_#{comment.id}", id: "edit_button_of_#{comment.id}", 'data-bs-toggle': 'collapse',
                       class: 'dropdown-item collapsed') do
               Edit
+          - if Flipper.enabled?(:content_moderation, User.session)
+            - if policy(comment).moderate?
+              = button_to(comment.moderated? ? 'Unmoderate' : 'Moderate', comment_moderate_path(comment),
```
hennevogel, line 22:
```suggestion
              = button_to(comment.moderated? ? 'Release' : 'Moderate', comment_moderate_path(comment),
```

src/api/app/policies/comment_policy.rb
```diff
@@ -31,4 +31,11 @@ def update?
   def reply?
     !(user.blank? || user.is_nobody? || record.user.is_nobody?)
   end
+
+  # Only logged-in Admins of Staff members can moderate comments
+  def moderate?
+    return true if user.try(:is_admin?) || user.try(:is_staff?)
+
+    false
+  end
```
hennevogel, line 40:
```suggestion
  def moderate?
    return false if record.user == User.nobody # soft-deleted comments
    return true if user.try(:is_admin?) || user.try(:is_staff?)

    false
  end
```

src/api/app/views/webui/comment/_reply.html.haml
```diff
@@ -15,6 +15,11 @@
       = link_to('#', data: { 'bs-toggle': 'modal', 'bs-target': "#delete-comment-modal-#{comment.id}" },
                 class: 'delete_link btn btn-outline-danger', title: 'Delete comment') do
         Delete
+  - if Flipper.enabled?(:content_moderation, User.session)
+    - if policy(comment).moderate?
+      %li.list-inline-item
+        = button_to(comment.moderated? ? 'Unmoderate' : 'Moderate', comment_moderate_path(comment),
```
hennevogel, line 21:
```suggestion
        = button_to(comment.moderated? ? 'Release' : 'Moderate', comment_moderate_path(comment),
```

src/api/app/components/bs_request_comment_component.html.haml
saraycp: Relase can be confusing on our context. What about synonyms like: exempt, write off, reprieve... ?

hennevogel (approved):

src/api/app/views/webui/comment/_reply.html.haml
```diff
@@ -15,6 +15,11 @@
       = link_to('#', data: { 'bs-toggle': 'modal', 'bs-target': "#delete-comment-modal-#{comment.id}" },
                 class: 'delete_link btn btn-outline-danger', title: 'Delete comment') do
         Delete
+  - if Flipper.enabled?(:content_moderation, User.session)
+    - if policy(comment).moderate?
```
danidoni, line 19: What about moving the flipper check inside the policy?

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -17,6 +17,9 @@
             = link_to("#edit_form_of_#{comment.id}", id: "edit_button_of_#{comment.id}", 'data-bs-toggle': 'collapse',
                       class: 'dropdown-item collapsed') do
               Edit
+          - if policy(comment).moderate?
+            = button_to(comment.moderated? ? 'Unmoderate' : 'Moderate', comment_moderate_path(comment),
```
hennevogel, line 21:
```suggestion
            = button_to(comment.moderated? ? 'Permit' : 'Moderate', comment_moderate_path(comment),
```

src/api/spec/policies/comment_policy_spec.rb
```diff
@@ -144,4 +144,42 @@
...
+
+    context 'with a deleted comment' do
+      it 'an admin user is unable to moderate a deleted comment' do
+        expect(subject).not_to permit(admin_user, comment_deleted)
+      end
+
+      it 'an non-admin user is unable to moderate a deleted comment' do
+        expect(subject).not_to permit(other_user, comment_deleted)
+      end
+    end
```
hennevogel, line 174:
```suggestion
    context 'with a deleted comment' do
      it 'no one is able to moderate a deleted comment' do
        expect(subject).not_to permit(admin_user, comment_deleted)
      end
    end
```

src/api/app/views/webui/comment/_reply.html.haml
saraycp: I finally left it as it was.

saraycp:
Rebase on master and resolve conflicts on schema.rb

saraycp:
Again. Rebase on master and resolve conflicts on schema.rb

src/api/app/views/webui/comment/_reply.html.haml
danidoni: Alright

danidoni (approved):

PR #14908, opened 2023-09-14
Improve minimum button size of requests link in tables

dirkmueller:
The current request listing requires to hit *exactly* the 15x18 pixel region to the icon to get into the detailed listing. nothing else in the whole line links to details. there is no sensible keyboard navigation either.

This improves it to roughly 30x36 which is still far away from the minimum recommended sizes for user interaction elements (44x44) but slightly less insane.


dirkmueller:
image

this is new vs old / current

dirkmueller:
another option could be to move the `<a>` link to the first column and remove the last column alltogether. together with some keyboard navigation fixes that could be even better.

hennevogel:
The real solution is to make an interface and not render a database table in HTML 😃

PR #14909, opened 2023-09-14
Add report specs

danidoni:
This PR follows-up #14880


src/api/app/policies/report_policy.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def show?
+    user.is_admin? || record.user == user
+  end
+
+  def create?
+    return false unless Flipper.enabled?(:content_moderation, user)
+
+    # We don't want reports twice...
+    return false if user.submitted_reports.where(reportable: record.reportable).any?
```
rubhanazeem, line 20: @danidoni this is not correct. I've fixed this in PR #14880.

danidoni: Ah, I messed up while rebasing. Let me change it

src/api/spec/factories/reports.rb
```diff
@@ -0,0 +1,21 @@
...
+  factory :report_for_project, class: 'Report' do
+    user { association :confirmed_user }
+    reportable { association :project }
+  end
+
+  factory :report_for_package, class: 'Report' do
+    user { association :confirmed_user }
+    reportable { association :package }
+  end
+end
```
hennevogel, line 21: https://thoughtbot.github.io/factory_bot/traits/summary.html

src/api/spec/policies/report_policy_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+      it { is_expected.not_to permit(user, record) }
+    end
+  end
+
+  permissions :create? do
+    before do
+      Flipper.enable(:content_moderation, user)
+    end
+
+    context 'when the user does not have the content_moderation active' do
```
hennevogel, line 34: I don't think we need to spec this. There is no logic involved, just makes it harder to rip out once we roll this out...

src/api/spec/factories/reports.rb
danidoni: Fixed

src/api/spec/factories/reports.rb
```diff
@@ -0,0 +1,21 @@
+FactoryBot.define do
+  factory :report do
+    user
```
hennevogel, line 3: I think you should have a default for reportable

src/api/spec/policies/report_policy_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+    end
+
+    context 'when the current user is an admin' do
+      let(:user) { create(:admin_user) }
+      let(:record) { create(:report, :for_comment) }
+
+      it { is_expected.to permit(user, record) }
+    end
+
+    context 'when the current user is a regular user' do
```
hennevogel, line 21:
```suggestion
    context 'when the current user is not the owner of the report' do
```

hennevogel (changes requested):

src/api/spec/policies/report_policy_spec.rb
danidoni: Okay

hennevogel (approved):

PR #14920, opened 2023-09-15
Allow object deletion despite of reports

saraycp:
When a moderator reports any misconduct on any object like comment, project, package or user and then decides to delete that object, we don't want to delete the associated report but nullify it. That's only possible if the database allows it.

### DIFF
--- a/src/api/app/models/report.rb
+++ b/src/api/app/models/report.rb
@@ -13,10 +13,10 @@ end
 #
 #  id              :bigint           not null, primary key
 #  reason          :text(65535)
-#  reportable_type :string(255)      not null, indexed => [reportable_id]
+#  reportable_type :string(255)      indexed => [reportable_id]
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
-#  reportable_id   :integer          not null, indexed => [reportable_type]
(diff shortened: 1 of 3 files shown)


src/api/.database_consistency.todo.yml
```diff
@@ -1718,6 +1718,10 @@ ReleaseTarget:
   id:
     PrimaryKeyTypeChecker:
       enabled: false
+Report:
+  reportable:
+    ColumnPresenceChecker:
+      enabled: false
```
hennevogel, line 1724: Why disabling instead of validating?

saraycp: Thinking twice... the issue is a different one. I don't have to fix or ignore the lack of consistency in the DB.

The problem is that we were making `Report#reportable` compulsory, but we shouldn't as we can nullify it at some point.

So I'm going to modify this line:
https://github.com/openSUSE/open-build-service/blob/eccd4b1163b1b6040f62bb6bb3cc533320428fd7/src/api/app/models/report.rb#L7

saraycp: done, @hennevogel @hellcp-work

danidoni (approved):

PR #14934, opened 2023-09-20
Show information about existing reports on the associated objects

hellcp-work:
Screenshot from 2023-09-20 17-43-00
Screenshot from 2023-09-20 17-42-41
Screenshot from 2023-09-20 17-42-30
Screenshot from 2023-09-20 17-42-10

### DIFF
--- a/src/api/spec/components/previews/reports_notice_component_preview.rb
+++ b/src/api/spec/components/previews/reports_notice_component_preview.rb
@@ -2,8 +2,8 @@ class ReportsNoticeComponentPreview < ViewComponent::Preview
   # Preview at http://HOST:PORT/rails/view_components/reports_notice_component/user_reportable_as_admin
   def user_reportable_as_admin
     reportable = User.last
-    User.take(3).each do |reporter|
-      Report.new(reportable:, user: reporter).save!
+    User.take(3).each do |user|
+      Report.new(reportable:, user:).save!
     end
     render(ReportsNoticeComponent.new(reportable:, user: User.admins.first))
   end
@@ -12,8 +12,8 @@ class ReportsNoticeComponentPreview < ViewComponent::Preview
   def user_reportable_as_reporter
     reportable = User.last
     reporters = User.take(3)
-    reporters.each do |reporter|
-      Report.new(reportable:, user: reporter).save!
+    reporters.each do |user|
+      Report.new(reportable:, user:).save!
     end
     render(ReportsNoticeComponent.new(reportable:, user: reporters.last))
   end


src/api/app/components/reports_notice_component.html.haml
```diff
@@ -0,0 +1,6 @@
+.text-warning
+  %i.fas.fa-flag
+  - if by_user
+    You reported this #{reportable.class.name.downcase}.
```
danidoni, line 4: Maybe we can push this `reportable.class.name.downcase` into the `Report` class, as we use it in several places now.

src/api/app/components/reports_notice_component.html.haml
```diff
@@ -0,0 +1,6 @@
...
+  - else
+    This #{reportable.class.name.downcase} has #{pluralize(reports.count, 'report')}.
```
danidoni, line 6: This `pluralize(...)` can go inside the `reports` method in the component class. WDYT?

src/api/app/components/reports_notice_component.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def initialize(reportable:, user:)
+    super
+
+    @reportable = reportable
+    @user = user
+  end
+
+  def by_user
+    !!Report.find_by(user:, reportable:)
```
eduardoj, line 12:
```suggestion
    Report.exists?(user:, reportable:)
```
I haven't tested this, but it should work.

src/api/app/components/reports_notice_component.rb
```diff
@@ -0,0 +1,23 @@
...
+  attr_reader :reportable, :user
+
+  def initialize(reportable:, user:)
+    super
+
+    @reportable = reportable
+    @user = user
+  end
+
+  def by_user
```
eduardoj, line 11:
```suggestion
  def by_user?
```
This should make clearer that this is a method that returns `true` or `false`...

src/api/app/components/reports_notice_component.html.haml
danidoni: What about making `reports` method an instance variable then?

src/api/spec/components/previews/reports_notice_component_preview.rb
```diff
@@ -0,0 +1,20 @@
+class ReportsNoticeComponentPreview < ViewComponent::Preview
+  # Preview at http://HOST:PORT/rails/view_components/reports_notice_component/user_reportable_as_admin
+  def user_reportable_as_admin
+    reportable = User.last
+    User.take(3).each do |reporter|
+      Report.new(reportable:, user: reporter).save!
+    end
```
eduardoj, line 7: I think this could work:

```suggestion
    User.take(3).each do |user|
      Report.new(reportable:, user:).save!
    end
```

... but also to change the name of the parameter from `user` to `reporter`:

```suggestion
    User.take(3).each do |reporter|
      Report.new(reportable:, reporter:).save!
    end
```

Now we have kind of a mix of both. What do you think?

danidoni (approved):

PR #14935, opened 2023-09-20
Refactor report component

eduardoj:


krauselukas (dismissed):
The report modal does not open anymore when I try to report a comment (on request, packages and projects). Clicking the report button has no effect.

danidoni:
> The report modal does not open anymore when I try to report a comment (on request, packages and projects). Clicking the report button has no effect.

After rebasing it, it somehow works, but the modal does not close itself after submitting. Looking at it...

danidoni:
> > The report modal does not open anymore when I try to report a comment (on request, packages and projects). Clicking the report button has no effect.
> 
> After rebasing it, it somehow works, but the modal does not close itself after submitting. Looking at it...

Fixed. @krauselukas can you take a look again? Thanks!

krauselukas:
There is an issue now with this refactoring. We render the report component with the modal only once one the package, project and request and set the attributes on it with javascript. We are not rendering it for the comments anymore, since the modal is already rendered in the view. At the same time we added the `render?` condition that checks with the report policy if the user has permission to report. But since its not rendered for the individual comments, it can happen that no modal is rendered just because the user cant report the project... I will fix this

eduardoj:
LGTM.

PR #14949, opened 2023-09-21
Do not use action as "local" variable 

hennevogel:
buildresult is a controller action, re-using it as variable is a desaster bound
to happen...


eduardoj:
> buildresult is a controller action, re-using it as variable is a desaster bound to happen...

For example, to try to refer to `buildresult` with `buildresults`:

https://github.com/openSUSE/open-build-service/blob/e41741acb2d43707b219e05e88e250a2dfea5f99/src/api/app/controllers/webui/project_controller.rb#L7

danidoni (approved):

PR #14921, opened 2023-09-15
Provide some test Reports for development

danidoni:

PR #14954, opened 2023-09-22
Add test data for decisions

saraycp:
Based on #14921

**Testing tips**

1. bundle exec rake dev:test_data:create
2. rails c
3. Decision.count # There should be 4 decisions
4. Decision.first.reports # The first decision applies to two reports

### DIFF
--- a/src/api/lib/tasks/dev/reports.rake
+++ b/src/api/lib/tasks/dev/reports.rake
@@ -19,5 +19,28 @@ namespace :dev do
         Report.create!(reportable: reportable, user: iggy, reason: 'Watch your language, please')
       end
     end
+
+    # Run `rake dev:reports:decisions` (always after running `rake dev:reports:data`)
+    desc 'Create decisions related to existing reports'
+    task decisions: :development_environment do
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+
+      puts 'Taking decisions regarding some reports'
+
+      admin = User.get_default_admin
+
+      Report.find_each do |report|
+        # Reports with even id will be 'cleared' (0). Those with odd id will be 'favor' (1).
+        decision = Decision.create!(reason: "Just because! #{report.id}", moderator: admin, kind: (report.id % 2))
+        decision.reports << report
+      end
+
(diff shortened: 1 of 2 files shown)


src/api/lib/tasks/dev/reports.rake
```diff
@@ -19,5 +19,28 @@ namespace :dev do
...
+
+      Report.find_each do |report|
+        # Reports with even id will be 'cleared' (0). Those with odd id will be 'favor' (1).
+        decision = Decision.create!(reason: "Just because! #{report.id}", moderator: admin, kind: (report.id % 2))
+        decision.reports << report
+      end
+
+      # The same decision applies to more than one report about the same object/reportable.
+      reportable = Decision.first.reports.first.reportable
+      another_user = User.find_by(login: 'Requestor')
```
danidoni, line 41: This is created later, in the `notifications` rake task. It is not finding that user and crashing the creation of the test data.

danidoni (changes requested):

src/api/lib/tasks/dev/reports.rake
ncounter, line 41: There is a `race condition` with this User: actually it gets created in [`src/api/lib/tasks/dev/notifications.rake`](https://github.com/openSUSE/open-build-service/blob/master/src/api/lib/tasks/dev/notifications.rake#L16) but this PR introduces `Rake::Task['dev:reports:decisions'].invoke` in `src/api/lib/tasks/dev.rake` before that user gets created. A fast-forward solution would be to swap position between `Rake::Task['dev:reports:decisions'].invoke` and `Rake::Task['dev:notifications:data'].invoke(2)`, but to be honest, I'd rather rely on another user (like `_nobody_` that is created in `seeds.db` or even better move the `Requestor` user creation there too.

saraycp: I could have sworn I ran the task after moving the code...  maybe I didn't save it before :facepalm: 

Thanks for catching this @ncounter @danidoni!  I pushed the changes already (creating Requestor if it's not there yet).

ncounter (approved):

danidoni (approved):

PR #14955, opened 2023-09-22
Implement UI for creating decision objects

hellcp-work:
Implements a ui for submitting decision to non decided reports.

Screenshot from 2023-09-22 14-37-13


src/api/app/models/user.rb
```diff
@@ -404,6 +404,12 @@ def is_staff?
     @is_staff = roles.exists?(title: 'Staff')
   end
 
+  def is_moderator?
+    return @is_moderator unless @is_moderator.nil?
+
+    @is_moderator = roles.exists?(title: 'Moderator')
```
danidoni, line 410:
```suggestion
    @is_moderator ||= roles.exists?(title: 'Moderator')
```
Would this be the same?

krauselukas (approved):

PR #14964, opened 2023-09-25
Align `is_moderator?` method and report pundit policies

krauselukas:
In the beginning of the content_moderation feature we considered
user that have the admin or stuff roles to be moderator. But now
we have the dedicated moderator role. So we only should check
for that in the `is_moderator?` method and leave the rest to
the pundit policies.

The authorization logic about if a user should receive notifications
for reports is spread over several places.
Lets move it to a policy to avoid duplication and confusion.

Also in the moderate comment policy we didn't considered the moderator role, 
only staff and admin users.

### DIFF
--- a/src/api/app/policies/comment_policy.rb
+++ b/src/api/app/policies/comment_policy.rb
@@ -32,10 +32,10 @@ class CommentPolicy < ApplicationPolicy
     !(user.blank? || user.is_nobody? || record.user.is_nobody?)
   end
 
-  # Only logged-in Admins of Staff members can moderate comments
+  # Only logged-in Admins/Staff members or user with moderator role can moderate comments
   def moderate?
     return false if record.user.is_nobody? # soft-deleted comments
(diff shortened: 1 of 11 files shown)


src/api/app/policies/comment_policy.rb
```diff
@@ -32,10 +32,10 @@ def reply?
     !(user.blank? || user.is_nobody? || record.user.is_nobody?)
   end
 
-  # Only logged-in Admins of Staff members can moderate comments
+  # Only logged-in Admins/Staff members or user with moderator role can moderate comments
```
ncounter, line 35: I would also not show the `Moderate` button in case the user is the author

`return false if user == record.user`

in that case the user should edit/delete its own comment.

ncounter: nice , thank you

danidoni (approved):

danidoni:
Okay, It just needs to be rebased with master. I'll rebase it.

danidoni (approved):

PR #14967, opened 2023-09-26
Introduce notifications for decisions on moderated objects

danidoni:
This PR wires up the events and subscriptions required to get a notification when a Moderator makes a Decision about some Report.

It features very basic notifications, the actual notification is fleshed out in #14976 

The filters required for the notification screen are going to be implemented in #14975 

offender_and_reporters_receive_notifications

### DIFF
--- a/src/api/app/components/notification_avatars_component.rb
+++ b/src/api/app/components/notification_avatars_component.rb
@@ -9,6 +9,7 @@ class NotificationAvatarsComponent < ApplicationComponent
 
   private
 
+  # rubocop:disable Metrics/CyclomaticComplexity
   def avatar_objects
     @avatar_objects ||= case @notification.notifiable_type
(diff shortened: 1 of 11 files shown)


saraycp (changes requested):

saraycp:
I wonder if it's better to create one event per decision-report. My last commit is a WIP of the possible implementation.
We can discuss whether to continue or drop it.

saraycp:
> I wonder if it's better to create one event per decision-report. My last commit is a WIP of the possible implementation.
> We can discuss whether to continue or drop it.

We already discussed it and concluded this is not needed. We are not going to mention specific reports but the reportable, which is the same for all the reports related to the decision that generates the event.

saraycp (approved):

ncounter (approved):

PR #14968, opened 2023-09-26
No need to catch and re-raise the same exception...

hennevogel:


bmwiedemann (approved):
https://github.com/openSUSE/open-build-service/blob/a1fc5ff1132b/src/api/app/models/project.rb#L243 should cover it.

It was probably added here before 4e3b261b3ddeae35718563a98c20743106c44db6 added the human-readable message upstream.

PR #14975, opened 2023-09-27
Implement filters for decision events

danidoni:
Follows up #14967

### DIFF
--- a/src/api/app/queries/notifications_finder.rb
+++ b/src/api/app/queries/notifications_finder.rb
@@ -1,6 +1,10 @@
 class NotificationsFinder
   def initialize(relation = Notification.all)
-    @relation = relation.order(created_at: :desc)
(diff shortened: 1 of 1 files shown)


ncounter (approved):

saraycp (changes requested):
I know the idea is to display both reports-related and decision-related notifications under the Report filter.

But there is a corner case that is not working. Look at the screenshot. Iggy only have decision-related notifications, not report-related ones and the filter Report is not displayed.

That's what I see in review-up. Having a look locally...

Screenshot 2023-09-29 at 11-21-17 Open Build Service

danidoni:
> I know now... Iggy is not in beta. We control that people who are not in beta don't see the Report filter. However, we don't control that the notifications themselves are not displayed.

Fixed. Please take a look again!

saraycp (approved):
LGTM now

PR #14976, opened 2023-09-27
Implement better notifications for report decisions

danidoni:
The original PR started with a bare-bones implementation of the notification for Decisions. This PR is an improved view for those notifications

report_and_decision_notifications

Notifications about cleared and favor decisions look like this:

Screenshot 2023-09-29 at 10-41-49 Open Build Service

Follows up https://github.com/openSUSE/open-build-service/pull/14967

## How to test
1. The default test data has some decision notifications, so if you log in with Iggy (the reporter) and go into [the notification area](https://obs-reviewlab.opensuse.org/danidoni-implement-better-notifications-for-report-decisions/my/notifications?type=unread) and scroll down, you'll see some.

### DIFF
--- a/src/api/app/components/notification_action_description_component.rb
+++ b/src/api/app/components/notification_action_description_component.rb
@@ -30,11 +30,11 @@ class NotificationActionDescriptionComponent < ApplicationComponent
       when 'Event::BuildFail'
         "Build was triggered because of #{@notification.event_payload['reason']}"
       when 'Event::CreateReport'
-        "User '#{@notification.notifiable.user.login}' created a report for a #{@notification.event_payload['reportable_type']} for the following reason:"
+        "'#{@notification.notifiable.user.login}' created a report for a #{@notification.event_payload['reportable_type']} for the following reason:"
       when 'Event::ClearedDecision'
-        "User '#{@notification.notifiable.moderator.login}' cleared a report for #{@notification.notifiable.reports.first.reportable} for the following reason:"
+        "'#{@notification.notifiable.moderator.login}' cleared a report for #{@notification.notifiable.reports.first.reportable} for the following reason:"
       when 'Event::FavoredDecision'
-        "User '#{@notification.notifiable.moderator.login}' favored a report for #{@notification.notifiable.reports.first.reportable} for the following reason:"
+        "'#{@notification.notifiable.moderator.login}' favored a report for #{@notification.notifiable.reports.first.reportable} for the following reason:"
       end
     end
   end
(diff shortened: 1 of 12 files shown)


ncounter:
From a UI consistency perspective it would be nice to have an alignment of the notification content like the other types do:
image

Unfortunately, while it seems quite trivial, from the code perspective I think it is a bit tricky, because the second and aligned line (like in a Comment notification body) it comes from the `notification_component.html.haml` using the `NotificationExcerptComponent` who extract some content, but in the Report case is not used and it might need collateral changes to do so.
Maybe you can give it a try to see if it is still easy enough to achieve right away, otherwise I'd keep track with a card for this: the notification component seems a bit convoluted and might need refactoring/reshaping.

saraycp (changes requested):
The link on top is wrong, doesn't point to a comment/project/etc.

Debugging the code I see that at some point `reportable_type` is set to `Report`. It should be Comment/Project/etc.

I'll try to push a commit with the solution

src/api/app/components/notification_action_description_component.rb
```diff
@@ -34,9 +35,20 @@ def call
           concat(tag.p("User '#{@notification.notifiable.user.login}' created a report for a #{@notification.event_payload['reportable_type']} for the following reason:"))
           concat(tag.p(@notification.event_payload['reason']))
         end
+      when 'Event::ClearedDecision'
+        capture do
+          concat(tag.p("User '#{@notification.notifiable.moderator.login}' cleared a report for #{@notification.notifiable.reports.first.reportable} for the following reason:"))
```
ncounter, line 40: I'd omit the word `User` (here and in the other similar cases) and leave just the login value in the sentence:
`Iggy cleared a report........`

even better if it could be the either in bold or the link to the user page.

saraycp:
@ncounter I decided to move the second paragraph (the reason) to the excerpt component. In consequence, the left margin is no longer there but I think the difference between the info and the reason is clear enough. Mostly after rephrasing the text. What do you think @hellcp-work?

Please have a look at the screenshot, you can compare different types of notifications now.

ncounter:
> @ncounter I decided to move the second paragraph (the reason) to the excerpt component. In consequence, the left margin is no longer there but I think the difference between the info and the reason is clear enough. Mostly after rephrasing the text. What do you think @hellcp-work?
> 
> Please have a look at the screenshot, you can compare different types of notifications now.

Perfect, thank you :+1:

saraycp (approved):

ncounter (approved):

Issue #14810, opened 2023-08-16
cannot create rss type token via POST /person/$login/token


dcermak:
When trying to create a rss token via `POST /person/$login/token?cmd=create&operation=rss` you get a `service` type token instead.

PR #14979, opened 2023-09-28
Handle invalid request token operation values

ncounter:
Fix https://github.com/openSUSE/open-build-service/issues/14810

tl;dr
Handle and reject requests when `token`'s `operation` values are not valid.

Long version:
The former issue #14810 hits this scenario:
- user wants to create an `rss_token`
- he finds `token`'s endpoints to use
- he submit a `create token` request with an `operation` parameter value which is meant to be for an `rss` token
- eventually the `create token` allows any `operation` value in a fallback creating a `service token`, which is far away from the user's goal

The problem can be solved by two different actions:
1. document the `rss_token` creation (because at the moment I couldn't find it)
2. prevent creating a `token` if the `operation` value has no meaningful value (AKA only if it matches one of the possible values or it is completely omitted, and it fails in case of _any other value_ otherwise), reject the request and hints the user with a message about what happened

This PR addresses the second point.

### How to test this PR
- checkout this branch
- hit the `POST /person/$login/token?cmd=create&operation=XYZ` multiple times trying to change the `operation=` parameter value with: 
  - one of the possible API values `['rebuild', 'release', 'runservice', 'workflow']` --> it creates a token accordingly
  - omitting the `operation` parameter completely --> it creates a `runservice` token
  - using `operation=rss` --> it fails with an error message
  - using `operation=blabla` --> it fails with an error message

### DIFF
--- a/src/api/app/controllers/person/token_controller.rb
+++ b/src/api/app/controllers/person/token_controller.rb
@@ -4,6 +4,7 @@ module Person
     rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
 
     before_action :set_user
+    before_action :validate_operation, only: [:create]
 
     # GET /person/<login>/token
     def index
@@ -51,6 +52,19 @@ module Person
       raise MissingParameterError, 'The package and project parameters must be provided together.' unless params[:project] && params[:package]
 
       @package = Package.get_by_project_and_name(params[:project], params[:package])
+    end
+
+    def validate_operation
+      operation_param = params[:operation]
+      # TODO: align operation parameter allowed values
+      # - webUI: https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/token.rb#L27
+      # - API: https://github.com/openSUSE/open-build-service/blob/master/src/api/public/apidocs/paths/person_login_token.yaml#L89
+      return if operation_param.nil? ||
+                ['runservice', 'rebuild', 'release', 'workflow'].include?(operation_param) # possible API parameter values
+
+      render_error status: 400,
+                   errorcode: 'invalid_token_type',
+                   message: "'#{operation_param}' is not a valid operation type for a token."
     end
   end
 end


src/api/app/controllers/person/token_controller.rb
```diff
@@ -50,5 +51,17 @@ def set_package
 
       @package = Package.get_by_project_and_name(params[:project], params[:package])
     end
+
+    def validate_operation
+      operation_param = params[:operation]
+      return if operation_param.nil? || Token::OPERATIONS.include?(operation_param.capitalize)
```
eduardoj, line 57: Is this line changing from the allowed token type `runservice` to `service`? :thinking:

This is already documented [here](https://github.com/openSUSE/open-build-service/blob/221811444256005c46b48af45f7f6313496c8499/src/api/public/apidocs/paths/person_login_token.yaml#L89).

ncounter: Ah, I think I misused https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/token.rb#L27 set of possible values instead. But now I'm puzzled: why the API uses a different set than the webUI?

ncounter: I added a change to handle and accept both webUI and API values keeping the same behavior otherwise.

https://github.com/openSUSE/open-build-service/pull/14979/files#diff-67a22ba8cf70b37462f7ef3e9fe0370283ffae0f2ace759387591b1bbb06324dR57-R62

But it still feels dirty and hacked somehow, so we should rather align the two usages eventually.

src/api/app/controllers/person/token_controller.rb
```diff
@@ -50,5 +51,18 @@ def set_package
...
       @package = Package.get_by_project_and_name(params[:project], params[:package])
     end
+
+    def validate_operation
+      operation_param = params[:operation]
+      # TODO: align operation parameter allowed values
+      # - webUI: https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/token.rb#L27
+      # - API: https://github.com/openSUSE/open-build-service/blob/master/src/api/public/apidocs/paths/person_login_token.yaml#L89
+      return if operation_param.nil? ||
+                ['runservice', 'rebuild', 'release'].include?(operation_param) # possible API parameter values
```
eduardoj, line 61:
```suggestion
                ['runservice', 'rebuild', 'release', 'workflow'].include?(operation_param) # possible API parameter values
```
We also support the `workflow` value for the `operation` parameter, as the failing test pointed out. It is not yet documented, but we support it.

eduardoj (approved):

PR #14980, opened 2023-09-28
Revert "[src/api] Update selenium-webdriver 4.5.0 → 4.13.1 (minor)"

bmwiedemann:
Reverts openSUSE/open-build-service#14973

This probably broke package build in https://build.opensuse.org/package/show/OBS:Server:Unstable/obs-server


hennevogel:
Still the same problem as #13536

PR #14985, opened 2023-09-29
[backend] improve requestid logging

mlschroe:


hennevogel:
How about you allow us to pass it to your API? The we can connect frontend and backend requests.

mlschroe:
We already do this via "X-Request-ID".

PR #14988, opened 2023-09-29
Fetch bugowners from the devel package in the controller as well

### LINKED ISSUE #14987: Show a "report bug" button on packages in openSUSE:Factory and other distro projects
Especially recently, users use comments on OBS packages as a place to report bugs. This is not the right way and will hopefully be disabled soon (https://github.com/openSUSE/open-build-service/issues/4192). However, this is a sign that somehow users gravitate towards those pages on OBS instead of bugzilla and we need a way to get them onto the right track again.

Like for other packages on OBS where there is a bugowner defined, distro packages should also have a "report bug" button that ends up on bugzilla, with the appropriate assignee/cc (bugowner/maintainer from devel pkg/prj) and subject/text (project and package name) filled out as much as possible.

In the case that comments are disabled, the report bug button could be placed next to where the comments would be.

None.

**Additional context**
Most of https://github.com/openSUSE/open-build-service/issues/4192

hellcp-work:
This is intended to enable both package's and devel package's bugowners to be mentioned in bugzilla link, so that report bug link shows when there's a bugowner in a devel package only, and to make the developers of the package aware of the issue.

Fixes #14987

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -56,7 +56,7 @@ class Webui::PackageController < Webui::WebuiController
     @srcmd5 = params[:srcmd5]
     @revision_parameter = params[:rev]
 
-    @bugowners_mail = (@package.bugowner_emails + @project.bugowner_emails).uniq
+    @bugowners_mail = bugowners_emails
     @revision = params[:rev]
     @failures = 0
 
@@ -700,5 +700,10 @@ class Webui::PackageController < Webui::WebuiController
       opts[k] = params[k] if params[k].present?
     end
     opts
+  end
+
+  def bugowners_emails
+    @package.develpackage&.bugowner_emails.presence || @package.bugowner_emails.presence ||
+      @package.develpackage&.project&.bugowner_emails.presence || @project.bugowner_emails.presence || []
   end
 end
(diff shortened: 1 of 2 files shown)


src/api/app/controllers/webui/package_controller.rb
```diff
@@ -56,7 +56,7 @@ def show
     @srcmd5 = params[:srcmd5]
     @revision_parameter = params[:rev]
 
-    @bugowners_mail = (@package.bugowner_emails + @project.bugowner_emails).uniq
+    @bugowners_mail = (@package.bugowner_emails + @project.bugowner_emails + @package.develpackage.try(:bugowner_emails) + @package.develpackage.try(:project).try(:bugowner_emails)).uniq
```
vpereira, line 59: wouldn't be better to use a safe navigation operator (&.) ?

i.e 

```

def your_controller
  ....
  @bugowner_emails = maintaners_emails
end

private 

def devel_emails
 @package.develpackage&.bugowner_emails || []
end

bug devel_project_emails 
  @package.develpackage&.project&.bugowner_emails || []
end

def maintaners_emails
 (devel_emails + devel_project_emails).uniq
end
```

hennevogel:
This is not what `osc` does though...

PR #14999, opened 2023-10-04
Create a default subscription for Decision-related events

danidoni:
We need to have everyone subscribed automatically to FavoredDecision and ClearedDecision events.

### DIFF
--- a/src/api/db/data/20231004134538_create_a_default_subscription_for_decision_related_events.rb
+++ b/src/api/db/data/20231004134538_create_a_default_subscription_for_decision_related_events.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateADefaultSubscriptionForDecisionRelatedEvents < ActiveRecord::Migration[7.0]
+  def up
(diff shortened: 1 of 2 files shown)


krauselukas (dismissed):

krauselukas (changes requested):
@danidoni I think you forgot to commit the timestamp change in the `schema.rb` file

saraycp (changes requested):
Nobody is subscribed to Event::CreateReport in production right now.
I think we should do that in this data migration as well.
I double-checked if an admin can do it through the UI but they can't yet (Configuration > Notification).

danidoni:
> @danidoni I think you forgot to commit the timestamp change in the `schema.rb` file

Yeah, I completely forgot!

danidoni:
Depends on #15007

danidoni:
> > @danidoni I think you forgot to commit the timestamp change in the `schema.rb` file
> 
> Yeah, I completely forgot!

Schema file updated. Please take a look again.

krauselukas (approved):

PR #15004, opened 2023-10-05
Implement report categories

hellcp-work:
This implements model changes and UI for reporting with categories
Screenshot from 2023-10-05 10-59-48
Screenshot from 2023-10-05 10-59-37

### DIFF
--- a/src/api/app/models/report.rb
+++ b/src/api/app/models/report.rb
@@ -8,9 +8,23 @@ class Report < ApplicationRecord
 
   belongs_to :decision, optional: true
 
+  enum category: {
+    spam: 10,
+    scam: 20,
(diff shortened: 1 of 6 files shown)


src/api/app/models/report.rb
```diff
@@ -8,10 +8,24 @@ class Report < ApplicationRecord
 
   belongs_to :decision, optional: true
 
+  enum category: {
+    spam: 10,
```
krauselukas, line 12: @hellcp-work why starting at `10` not at `1`?

krauselukas: alright

krauselukas (approved):

danidoni (approved):

PR #15007, opened 2023-10-05
Notify only projects not already notified

danidoni:
Otherwise we get errors like this one:

    ActiveRecord::RecordInvalid: Validation failed: Notification These notification and project are already associated
    from /usr/lib64/ruby/gems/3.1.0/gems/activerecord-7.0.8/lib/active_record/validations.rb:80:in `raise_validation_error'

### DIFF
--- a/src/api/db/data/20200421121610_backfill_notified_projects.rb
+++ b/src/api/db/data/20200421121610_backfill_notified_projects.rb
@@ -1,7 +1,10 @@
 class BackfillNotifiedProjects < ActiveRecord::Migration[6.0]
   def up
     Notification.where.not(notifiable_id: nil).where.not(notifiable_type: nil).find_each do |notification|
-      notification.projects << NotifiedProjects.new(notification).call
+      # We cannot add the projects to be notified twice, so we need to calculate
+      # the remainder of the projects that are not yet notified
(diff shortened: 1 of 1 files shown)


src/api/db/data/20200421121610_backfill_notified_projects.rb
```diff
@@ -1,7 +1,10 @@
 class BackfillNotifiedProjects < ActiveRecord::Migration[6.0]
   def up
     Notification.where.not(notifiable_id: nil).where.not(notifiable_type: nil).find_each do |notification|
-      notification.projects << NotifiedProjects.new(notification).call
+      # We cannot add the projects to be notified twice, so we need to calculate
+      # the remainder of the projects that are not yet notified
+      projects_to_notify = NotifiedProjects.new(notification).call
```
krauselukas, line 6: @danidoni I think we have to do this here https://github.com/openSUSE/open-build-service/blob/b92d06f15f00110ca2b987e2e91becac9cf4fac3/src/api/app/services/notified_projects.rb#L29 I guess the issue occurs since we introduced the reports. So I guess we have to assign the projects there

krauselukas (approved):

PR #15011, opened 2023-10-06
Improve error message for request not found

### LINKED ISSUE #14248: GET /comment/request endpoint return raw query result
# Issue Description
When documenting the comment endpoints, I noticed that the request one returns a rather crude results, showing raw query response we would get from activerecord:

```
Couldn&apos;t find BsRequest with [WHERE `bs_requests`.`number` = ?]
```

Expected Result
============
The error message is more direct and references the request id.

How to Reproduce
===============
`osc api -X GET /comments/request/1`

eduardoj:
Fixes #14248.

### DIFF
--- a/src/api/app/controllers/comments_controller.rb
+++ b/src/api/app/controllers/comments_controller.rb
@@ -31,7 +31,9 @@ class CommentsController < ApplicationController
         @header = { project: @obj.name }
       end
     elsif params[:request_number]
-      @obj = BsRequest.find_by_number!(params[:request_number])
+      @obj = BsRequest.find_by(number: params[:request_number])
+      raise ActiveRecord::RecordNotFound, "Couldn't find Request with number '#{params[:request_number]}'" if @obj.nil?
+
       @header = { request: @obj.number }
     else
       @obj = User.session!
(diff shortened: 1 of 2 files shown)


src/api/app/controllers/comments_controller.rb
```diff
@@ -31,7 +31,9 @@ def find_obj
         @header = { project: @obj.name }
       end
     elsif params[:request_number]
-      @obj = BsRequest.find_by_number!(params[:request_number])
+      @obj = BsRequest.find_by(number: params[:request_number])
+      raise ActiveRecord::RecordNotFound, "Couldn't find Request with number '#{params[:request_number]}'" if @obj.nil?
```
ncounter, line 35: I'd even drop (here and in the other places) the `with` word, as `number` is not a property or a detail but it is the identifier of the `Request`

`Couldn't find Request number 1234` looks fine by me. What do you think?

eduardoj: I didn't want to deviate much from the by default message `"Couldn't find #{name} with '#{primary_key}'=#{id}"`. See https://github.com/rails/rails/blob/869459ed0c9eb368ad88225d056f6d42adffc9d4/activerecord/lib/active_record/core.rb#L252

I would prefer to keep the `with`.

ncounter: Oh, I see. Consistency for the win!!
Fine by me then :+1:

ncounter (approved):

PR #15017, opened 2023-10-06
Allow maintainers to view reports on their packages, projects and requests

hellcp-work:
Displays the reports modal without the decision form and displays the reports notice for them in order to trigger it.


saraycp (approved):
LGTM

PR #15028, opened 2023-10-10
[api] add missing revert parameter to assignreview command in docu

adrianschroeter:


src/api/public/apidocs/paths/request_id.yaml
```diff
@@ -85,6 +85,8 @@ post:
           value: cmd=addreview&by_group=heroes
         assignreview:
           value: cmd=assignreview&by_group=heroes&reviewer=foo
+        assignreview:
```
eduardoj, line 88:
```suggestion
        "assignreview with revert":
```
We can change the name of the example, just to not use the same example name as the previous one.

eduardoj: I added a commit with the changes. Feel free to use it however you prefer.

danidoni (approved):

PR #15039, opened 2023-10-11
Fix content-type header returned by GET /build/_... API endpoints

eduardoj:


mfriedrichgh (approved):
Tested on my local instance and works as described.

PR #15040, opened 2023-10-11
Stop setting options that are already default

hennevogel:
.exists_by_project_and_name and .get_by_project_and_name have defaults. Stop setting those options all over the app to their defaults again.


mfriedrichgh (approved):

src/api/app/models/bs_request_action_maintenance_incident.rb
```diff
@@ -185,7 +185,7 @@ def _merge_pkg_into_maintenance_incident(incident_project)
       new_pkg = Package.get_by_project_and_name(ret[:data][:targetproject], ret[:data][:targetpackage])
     elsif linkinfo && linkinfo['package'] # a new package for all targets
       if Package.exists_by_project_and_name(incident_project.name, source_package, follow_project_links: false)
-        new_pkg = Package.get_by_project_and_name(incident_project.name, source_package, use_source: false, follow_project_links: false)
```
dmarcoux, line 188: This change should be reverted since `follow_project_links: false` isn't a default option, so it should still be passed.

src/api/app/models/package.rb
```diff
@@ -234,7 +234,8 @@ def self.get_by_project_and_name!(project, package, opts = {})
 
   # to check existens of a project (local or remote)
   def self.exists_by_project_and_name(project, package, opts = {})
```
dmarcoux, line 236: One more instance of `allow_remote_packages: false` should be removed here:

https://github.com/openSUSE/open-build-service/blob/e9a6fa9c0e7ab5993a0be9d23812b0b6d1e8c84e/src/api/app/models/channel_binary.rb#L39

dmarcoux (changes requested):
See my 2 previous comments

src/api/app/models/bs_request_action_maintenance_incident.rb
hennevogel: ups

dmarcoux (approved):

PR #15044, opened 2023-10-11
Comment history dropdown

hellcp-work:
Displays a history of modifications to the comment

Screenshot from 2023-10-11 14-44-43

Screenshot from 2023-10-11 18-13-25

### DIFF
--- a/src/api/app/policies/comment_policy.rb
+++ b/src/api/app/policies/comment_policy.rb
@@ -69,7 +69,10 @@ class CommentPolicy < ApplicationPolicy
 
   def history?
     return false unless Flipper.enabled?(:content_moderation, user)
+    # Always display the comment history if the user is admin or moderator
+    return true if user.is_admin? || user.is_staff? || user.is_moderator?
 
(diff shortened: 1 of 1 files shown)


dmarcoux:
The modal display the _Before_ and _After_ states of a comment should take into consideration that a comment's text can be quite long. Right now, it's going to take the whole screen, like in this example below:

comments

We could definitely have a larger modal on bigger screens. An accordion or an expandable ellipsis for long comments would also help.

src/api/app/policies/comment_policy.rb
```diff
@@ -66,4 +66,10 @@ def locked?
       record.commentable.comment_lock.present?
     end
   end
+
+  def history?
+    return false unless Flipper.enabled?(:content_moderation, user)
+
+    !(record.moderated? || record.user.is_nobody?) || (user.is_admin? || user.is_staff? || user.is_moderator?)
```
danidoni, line 73: I find this a bit difficult to understand. Maybe rework the booleans or add a comment...

danidoni: Better

danidoni (approved):

PR #14991, opened 2023-09-29
Introduces remote package support for Package.get_by_project_and_name

hennevogel:
Instead of returning nil for packages that might be on remote or scmsync, check if they are known to the backend and return an in-memory, readonly package.

In the project [ProjectLinks:RemoteLinkedBuild](https://obs-reviewlab.opensuse.org/hennevogel-refactoringget_by_project_and_name/project/show/ProjectLinks:RemoteLinkedBuild) you should

- see `ctris` on the [monitor page](https://obs-reviewlab.opensuse.org/hennevogel-refactoringget_by_project_and_name/project/monitor/ProjectLinks:RemoteLinkedBuild)
- be able to [show `ctris`](https://obs-reviewlab.opensuse.org/hennevogel-refactoringget_by_project_and_name/package/show/ProjectLinks:RemoteLinkedBuild/ctris)
- and not be able to do anything else to the package

There are of course rough edges to the UI that would need to be hammered out...

PR #15045, opened 2023-10-11
Introduces scmsync package support for Package.get_by_project_and_name

hennevogel:
Depends on #14991 

Instead of returning nil for packages that might be on the SCM, check if they
are known to the backend and return an in-memory, readonly package.

In the project ProjectLinks:Scmsynced you should

- see some packages on the monitor page
- be able to show those packages
- and not be able to do anything else to the packages

There are of course rough edges to the UI that would need to be hammered out...


hennevogel:
need more work

PR #15047, opened 2023-10-11
Split up report event class

krauselukas:
This is the first PR to split up the events created by reports. The code still supports the old implementation of the `Event::CreateReport` class. But after this PR is merged we won't produce new records with this anymore, but instead the new STI event classes.
This enables us to safely deploy those changes, then migrate existing records first and finally remove the old implementation.

HOW TO TEST THIS:

1. Create some test data with `bundle exec rake dev:test_data:create`
2. The rake task will produce some reports and corresponding notifications
3. Check that notifications are still produced and the links on those are still working
4. Check the subscription form, it won't show the subscription option for the `Event::CreateReport` anymore, but should show the new ones


krauselukas:
Need to work on some specs

krauselukas:
> Need to work on some specs

Done

PR #15048, opened 2023-10-11
Fix "not_found" API responses

eduardoj:
For "Not Found" exceptions happening in the backend, return the `summary` section of that XML response.

Also, adapt the documentation accordingly.

## Before

Screenshot from 2023-10-11 17-00-54

## After

Screenshot from 2023-10-12 15-59-58

## For reviewers

Take a look at the 404 responses for the modified API endpoints in the review-app:
- [POST /source/{project_name}?cmd=undelete](https://obs-reviewlab.opensuse.org/eduardoj-fixapidocs_not_found_responses/apidocs/index#/Sources%20-%20Projects/post_source__project_name__cmd_undelete)
- [POST /worker?cmd=checkconstraints](https://obs-reviewlab.opensuse.org/eduardoj-fixapidocs_not_found_responses/apidocs/index#/Workers/post_worker_cmd_checkconstraints)

### DIFF
--- a/src/api/app/controllers/concerns/rescue_handler.rb
+++ b/src/api/app/controllers/concerns/rescue_handler.rb
@@ -42,8 +42,20 @@ module RescueHandler
       render_error status: 403, errorcode: 'modify_package_no_permission', message: exception.message
     end
 
-    rescue_from Backend::NotFoundError, ActiveRecord::RecordNotFound do |exception|
-      render_error message: exception.message, status: 404, errorcode: 'not_found'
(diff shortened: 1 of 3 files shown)


mfriedrichgh (changes requested):
Tested on my local instance, but results in the responses documented previously for both endpoints.

eduardoj:
> Tested on my local instance, but results in the responses documented previously for both endpoints.

This is fixed now. I added another commit. Please take another look.

mfriedrichgh (approved):

PR #15052, opened 2023-10-12
Instrument the content moderation feature

dmarcoux:
Nothing fancy, just the usual callbacks in models to track what happened. With this information, I plan on having panels in our Grafana instance to answer the following questions:

### Questions which can be answered with the current instrumentation
- [x] How many reports have been submitted?
- [x] Is there a certain report category which is more common than others?
- [x] Are the reports for comments, packages, projects, etc?
- [x] How long does it take for a decision to be taken after a report has been submitted?
- [x] Did the decisions favor or clear reports?

### DIFF
--- a/src/api/app/models/decision.rb
+++ b/src/api/app/models/decision.rb
@@ -11,6 +11,9 @@ class Decision < ApplicationRecord
   }
 
   after_create :create_event
+  after_create :track_decision
+
+  private
 
   def create_event
     case kind
@@ -23,6 +26,16 @@ class Decision < ApplicationRecord
 
   def event_parameters
     { id: id, moderator_id: moderator.id, reason: reason, reportable_type: reports.first.reportable.class.name }
+  end
+
+  def track_decision
+    RabbitmqBus.send_to_bus('metrics', "decision,kind=#{kind} hours_before_decision=#{hours_before_decision},count=1")
+  end
+
+  def hours_before_decision
+    first_report_created_at = reports.order(created_at: :asc).first.created_at
+
+    ((created_at - first_report_created_at) / 1.hour).floor
   end
 end
 
--- a/src/api/app/models/report.rb
+++ b/src/api/app/models/report.rb
@@ -17,6 +17,7 @@ class Report < ApplicationRecord
   }
 
   after_create :create_event
+  after_create :track_report
 
   scope :without_decision, -> { where(decision: nil) }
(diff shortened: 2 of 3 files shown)


dmarcoux:
[Instrument report and decisions notifications (M)](https://trello.com/c/I34Vsd0A/2351-instrument-report-and-decisions-notifications-m)

src/api/app/models/decision.rb
```diff
@@ -24,6 +27,10 @@ def create_event
   def event_parameters
     { id: id, moderator_id: moderator.id, reason: reason, reportable_type: reports.first.reportable.class.name }
   end
+
+  def track_decision
+    RabbitmqBus.send_to_bus('metrics', "decision,kind=#{kind},hours_before_decision=#{((updated_at - created_at) / 1.hour).floor} count=1")
```
bmwiedemann, line 32: `hours_before_decision` is a tag here - and tags don't like to have a large number of different values. Better add it as a value after the space.

src/api/app/models/decision.rb
```diff
@@ -24,6 +27,10 @@ def create_event
...
+  end
```
hennevogel, line 33:
```suggestion
  def track_decision
    hours_before_decision = ((updated_at - created_at) / 1.hour).floor
    RabbitmqBus.send_to_bus('metrics', "decision,kind=#{kind},hours_before_decision=#{hours_before_decision} count=1")
  end
```

src/api/app/models/report.rb
```diff
@@ -37,6 +38,10 @@ def create_event
   def event_parameters
     { id: id, user_id: user_id, reportable_id: reportable_id, reportable_type: reportable_type, reason: reason }
   end
+
+  def track_report
+    RabbitmqBus.send_to_bus('metrics', "report,category=#{category},type=#{reportable_type} count=1")
```
hennevogel, line 43: Maybe also track the number of sibling reports?

hennevogel (changes requested):

krauselukas (dismissed):

src/api/app/models/decision.rb
dmarcoux: Yeah, I will take this out and put it in a function. I was somewhat divided about this when submitting the PR.

hennevogel (approved):

PR #15053, opened 2023-10-12
Don't display the comment field when the comments are locked

hellcp-work:
Displays an alert instead of a comment submission field, and disables reply fields
Screenshot from 2023-10-12 13-03-25

### DIFF
--- a/src/api/app/models/comment_locking_validator.rb
+++ b/src/api/app/models/comment_locking_validator.rb
@@ -2,7 +2,7 @@ class CommentLockingValidator < ActiveModel::Validator
   def validate(record)
     commentable = record.commentable
     return unless commentable
-    return if comment_policy(record)
+    return if maintainer_or_unlocked(record)
 
     commentable_name = case commentable
                        when Package, Project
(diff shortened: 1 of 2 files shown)


src/api/app/models/comment_locking_validator.rb
```diff
@@ -14,4 +14,16 @@ def validate(record)
...
+  private
+
+  def comment_policy(record)
+    user = User.session
+    policy = CommentPolicy.new(user, record)
+    # Allow maintainers and admins, moderators and staff to create comments despite the lock
+    return true if policy.maintainer? || user.is_admin? || user.is_moderator? || user.is_staff?
+
+    # Check if there is a lock, including the parents (project for package and bs_request for bs_request_action)
+    !policy.locked?
```
krauselukas, line 27: @hellcp-work you will never reach this check if line 24 is truthy

src/api/app/policies/comment_policy.rb
```diff
@@ -3,6 +3,13 @@ def initialize(user, record)
     super(user, record, user_optional: true)
   end
 
+  def create?
+    return false if user.blank? || user.is_nobody?
+    return true if maintainer? || user.is_admin? || user.is_moderator? || user.is_staff?
+
+    !locked?
```
krauselukas, line 10: same here if the previous line is truthy you wont reach the check for the lock

src/api/app/models/comment_locking_validator.rb
krauselukas: upps, alright, there is even the comment about it...

krauselukas (approved):

PR #15056, opened 2023-10-13
Stop setting options that are already default, II

eduardoj:
`errorcode` has a default value when `status` is `404`. Stop setting those options all over the app to their defaults again.

See [here](https://github.com/openSUSE/open-build-service/blob/004d2ff1f9ce1331a01e266163522d6da2e2591d/src/api/app/controllers/application_controller.rb#L151), where errorcode takes its default value.


hennevogel (approved):
hehe 👍🏿

eduardoj:
Some checks failed:
```
codeclimate/total-coverage — 84% (-2.7% change)
codecov/patch Failing after 1s — 42.85% of diff hit (target 86.94%)
codecov/project — 84.23% (-2.71%) compared to 004d2ff
```

Merging.

PR #15061, opened 2023-10-13
Collect all linked project validations

hennevogel:
There are a couple of places throughout the app that validate LinkedProject.

- LinkedProject (duh!)
- Project.validate_link_xml_attribute
- Project.find_package

This just gathers all of them into LinkedProject so the validations in Project can be dropped some day...


hennevogel:
Depends on #15095

src/api/app/models/linked_project.rb
```diff
@@ -2,20 +2,52 @@ class LinkedProject < ApplicationRecord
...
-    elsif LinkedProject.find_by(project: project, linked_db_project: linked_db_project)
-      errors.add(:project, 'Already linked with that project')
-    end
+  def validate_target
+    return unless linked_db_project && linked_remote_project_name
+
+    errors.add(:base, 'can not have both linked_db_project and linked_remote_project_name')
+  end
+
+  def validate_cycles
```
danidoni, line 32:
```suggestion
  def validate_circular_dependency
```
I think that's the [standard name](https://en.wikipedia.org/wiki/Circular_dependency)

danidoni (approved):

src/api/app/models/linked_project.rb
hennevogel: not sure what you mean...

danidoni: It was about the `cycles` it took me a while to understand the code was ensuring no circular dependencies were set, so I thought that changing the `validate_cycles` to `validate_circular_dependency` was a better name.

PR #15071, opened 2023-10-17
Comment locking tests

ncounter:
Add comment locking tests:
- on a request with comments enabled:
  - [x] moderator can lock comments :heavy_check_mark: 
  - [x] non-moderator cannot lock comments but still can comment :heavy_check_mark: 
- on a request with comments disabled:
  - [x] moderator can unlock comments :heavy_check_mark: 
  - [x] non-moderator cannot unlock comments and cannot comment too :heavy_check_mark:

### DIFF
--- a/src/api/spec/features/webui/comment_locks_spec.rb
+++ b/src/api/spec/features/webui/comment_locks_spec.rb
@@ -1,0 +1,97 @@
+require 'browser_helper'
+
+RSpec.describe 'CommentLocks' do
+  let!(:moderator_user) { create(:moderator) }
+
+  before do
+    Flipper.enable(:content_moderation)
+  end
+
+  context 'on a request' do
+    let(:submitter) { create(:confirmed_user, :with_home, login: 'kugelblitz') }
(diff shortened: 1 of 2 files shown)


src/api/spec/features/webui/comment_locks_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        it 'can comment' do
+          expect(page).not_to have_text('Commenting on this is locked')
+          fill_in 'new_comment_body', with: 'Comment Body'
+          find_button('Add comment')
+        end
+      end
+    end
+
+    describe 'with comments disabled' do
```
danidoni, line 57: I think when you say "comments disabled" you mean "comments locked". If that's the case better use just one naming for all so lock/unlock instead of lock/unlock/enable/disable. WDYT?

ncounter: You're right. I was carried away by the fact in the end you cannot comment, but it's still about `comment lock`. Fixing...

danidoni (approved):

PR #15076, opened 2023-10-17
Create Isolated Migrations GitHub action

saraycp:
Check if migrations and their related code changes are not included on the same pull request.
Check if the [schema.rb](url) and data_schema.rb files are included in the pull request that introduces a migration or data migration.

Screenshot 2023-10-18 at 10-02-09 No name for job by saraycp · Pull Request #27 · saraycp_hello_world

Screenshot 2023-10-17 at 22-03-47 data_migration_only · saraycp_hello_world@c1e7860

You can see this GitHub action working on my [sandbox repository](https://github.com/saraycp/hello_world/pulls).
You will see a few pull requests representing the different scenarios we could have. For example, ["migration only"](https://github.com/saraycp/hello_world/pull/19) fails because the PR contains a migration but not the `schema.rb` changes.

### DIFF
--- a/.github/workflows/isolated_migrations.yml
+++ b/.github/workflows/isolated_migrations.yml
@@ -1,0 +1,67 @@
+# GitHub action that checks:
+# - that the migrations and the related code changes are not in the same pull request
+# - that the schema.rb and data_schema.rb files are included in the pull request that introduces the corresponding migration
+# Based on https://github.com/marketplace/actions/changed-files
+
+name: Isolated Migrations
+
+on:
+  pull_request
+
+jobs:
+  check_changed_files:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Group files that have changed
+        id: changed-files-yaml
+        uses: tj-actions/changed-files@v39
+        with:
+          files_yaml: |
+            migrations:
+              - src/api/db/**
+            not_migrations:
+              - '!src/api/db/**'
+            schema_migrations:
+              - src/api/db/migrate/**
+            data_migrations:
+              - src/api/db/data/**
+            schema:
+              - src/api/db/schema.rb
+            data_schema:
+              - src/api/db/data_schema.rb
+
+      - name: List the files related to migrations that this PR changes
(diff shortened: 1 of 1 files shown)


.github/workflows/isolated_migrations.yml
```diff
@@ -0,0 +1,69 @@
+# GitHub action that checks:
+# - if the migrations and the related code changes are not in the same pull request
```
danidoni, line 2:
```suggestion
# - that the migrations and the related code changes are not in the same pull request
```

.github/workflows/isolated_migrations.yml
```diff
@@ -0,0 +1,69 @@
...
+            schema_migrations:
+              - src/api/db/migrate/**
+            data_migrations:
+              - src/api/db/data/**
+            schema:
+              - src/api/db/schema.rb
+            data_schema:
+              - src/api/db/data_schema.rb
+
+      - name: List the files related to migrations that this PR changes
```
dmarcoux, line 37: Is there a reasoning for having this step here? Same for the other steps listing files. What is the benefit in having such steps?

saraycp: If the check fails, it's helpful to see what are the files included in the PR to understand why. I thought it was better to see them in the output than forcing the user to move back to the PR to see the _Files Changed_.

Moreover, while we fine-tune this GH action is good to see what is the action considering as migration-related files or non-migration-related files.

dmarcoux: Good points :+1:

.github/workflows/isolated_migrations.yml
```diff
@@ -0,0 +1,69 @@
...
+        if: steps.changed-files-yaml.outputs.migrations_any_changed == 'true'
+        run: |
+          echo "Related to migrations:"
```
dmarcoux, line 40: This `echo` and the others in the steps below doesn't provide value when looking at this in the GitHub UI. With the step name, it's already clear what the listed files are. See this example below:

ui

saraycp: Yes, we can omit them.

danidoni (approved):

PR #15084, opened 2023-10-19
Get notified about failed WorkflowRuns

eduardoj:
Until now, users who set the SCM/CI integration with OBS had to check the workflow runs list, from time to time, to discover if some workflow runs failed.
From now on, they are going to receive notifications about the failed ones.

The subscriptions page includes the new type of event:

Screenshot 2023-10-24 at 17-13-59 Open Build Service

The new type of notifications looks like this:

Screenshot from 2023-10-24 18-00-40

**NOTES:**

- This PR covers the web channel only. The email channel will be introduced in another PR.
- There is no need to subscribe everyone to the new type of event (`Event::WorkflowRunFail`). We'll use the blog post to encourage people to subscribe so we don't overload their Notifications page.
- At the moment, only the token creator (a.k.a. executor) will receive notifications. The people the owner shared the token with won't receive notifications yet.

### DIFF
--- a/src/api/app/models/event/workflow_run_fail.rb
+++ b/src/api/app/models/event/workflow_run_fail.rb
@@ -9,5 +9,9 @@ module Event
     def token_executors
       [Token.find_by(id: payload['token_id'], type: 'Token::Workflow')&.executor].compact
     end
+
+    def parameters_for_notification
+      super.merge(notifiable_type: 'WorkflowRun', notifiable_id: payload['id'])
+    end
   end
 end
--- a/src/api/app/models/event/workflow_run_fail.rb
+++ b/src/api/app/models/event/workflow_run_fail.rb
@@ -1,0 +1,13 @@
+module Event
+  class WorkflowRunFail < Base
+    self.message_bus_routing_key = 'workflow_run.fail'
+    self.description = 'Workflow run has failed'
+    payload_keys :id, :token_id
+
+    receiver_roles :token_executor
+
+    def token_executors
+      [Token.find_by(id: payload['token_id'], type: 'Token::Workflow')&.executor].compact
+    end
+  end
+end
--- a/src/api/app/models/workflow_run.rb
+++ b/src/api/app/models/workflow_run.rb
@@ -99,6 +99,26 @@ class WorkflowRun < ApplicationRecord
     [workflow_configuration_url, workflow_configuration_path].filter_map(&:presence).first
   end
 
+  def formatted_event_source_name
+    case hook_event
+    when 'pull_request', 'Merge Request Hook'
+      "##{event_source_name}"
+    else
+      event_source_name
+    end
+  end
+
+  # Examples of summary:
+  #   Pull request #234, opened
+  #   Merge request hook #234, open
+  #   Push 0940857924387654354986745938675645365436
+  #   Tag push hook Unknown source
+  def summary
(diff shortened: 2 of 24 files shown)


src/api/app/models/event/workflow_run_fail.rb
```diff
@@ -0,0 +1,17 @@
+module Event
+  class WorkflowRunFail < Base
+    self.message_bus_routing_key = 'workflow_run.fail'
+    self.description = 'Workflow run has failed'
+    payload_keys :id, :token_id # TODO: :generic_event_type, :hook_action, :hook_event, :repository_name, :repository_owner, :scm_vendor
+
+    receiver_roles :token_executor
+
+    def token_executor
```
danidoni, line 9: I think the receiver role finder method should be in plural form. See [here](https://github.com/openSUSE/open-build-service/blob/a7112563a0ae9b57ce18e668ff955c0410da003b/src/api/app/models/event_subscription/find_for_event.rb#L16)

saraycp: True! We realized later on.

src/api/app/components/notification_action_description_component.rb
```diff
@@ -54,4 +56,16 @@ def bs_request
                       @notification.notifiable.commentable
                     end
   end
+
+  def event_workflow_run_fail
```
danidoni, line 60: We're asking `WorkflowRun` a lot of data. I think this method should be moved there.

src/api/app/models/workflow_run.rb
```diff
@@ -27,6 +29,9 @@ class WorkflowRun < ApplicationRecord
   has_many :scm_status_reports, class_name: 'SCMStatusReport', dependent: :destroy
   has_many :event_subscriptions, dependent: :destroy
 
+  # Create the event only if the status was updated and changed from any value to 'fail'
+  after_save :create_event, if: -> { saved_change_to_attribute(:status)&.second == 'fail' }
```
danidoni, line 33:
```suggestion
  after_save :create_event, if: -> { has_failed? }
  
  def has_failed?
    saved_change_to_attribute(:status)&.second == 'fail'
  end
```
I find this easier to follow. What do you think?

eduardoj: Sure. We created `status_changed_to_fail?`. Please take a look again.

src/api/app/models/event/workflow_run_fail.rb
```diff
@@ -0,0 +1,17 @@
+module Event
+  class WorkflowRunFail < Base
+    self.message_bus_routing_key = 'workflow_run.fail'
+    self.description = 'Workflow run has failed'
+    payload_keys :id, :token_id
+
+    receiver_roles :token_executor
+
+    def token_executors
+      [Token.find_by(id: payload['token_id'], type: 'Token::Workflow')&.executor]
```
danidoni, line 10:
```suggestion
      [Token.find_by(id: payload['token_id'], type: 'Token::Workflow')&.executor.compact]
```
In order to remove the nil that will appear if there is no executor or no token

saraycp: Applied in two places

danidoni (approved):

saraycp:
@danidoni we have just added another spec. Please have a look at the last commit.

danidoni (approved):

danidoni (approved):

PR #15085, opened 2023-10-19
Collect artifacts for set flag step

rubhanazeem:
Prior to this PR, we were not collecting artifacts for the `set_flag` steps. This PR collects these missing artifacts.

image

### DIFF
--- a/src/api/app/components/workflow_artifacts_per_step_component.rb
+++ b/src/api/app/components/workflow_artifacts_per_step_component.rb
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/ClassLength
 class WorkflowArtifactsPerStepComponent < ApplicationComponent
   with_collection_parameter :artifacts_per_step
 
@@ -26,6 +27,8 @@ class WorkflowArtifactsPerStepComponent < ApplicationComponent
       artifacts_for_configure_repositories(parsed_artifacts)
     when 'Workflow::Step::SubmitRequest'
       artifacts_for_submit_request(parsed_artifacts)
+    when 'Workflow::Step::SetFlags'
+      artifacts_for_set_flag(parsed_artifacts)
     end
   rescue JSON::ParserError, ActionController::UrlGenerationError => e
     Airbrake.notify(e, artifacts_per_step_id: artifacts_per_step.id)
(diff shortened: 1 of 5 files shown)


danidoni (approved):

src/api/app/components/workflow_artifacts_per_step_component.rb
```diff
@@ -34,6 +37,37 @@ def call
...
+
+  def list_of_flags(flags)
+    flags.each do |flag|
+      concat(flag_step_sentence(flag))
+    end
+  end
+
+  def flag_step_sentence(flag)
+    if flag[:package]
+      path = helpers.package_show_path(project: flag[:project], package: flag[:package])
```
danidoni, line 54: :thought_balloon: The only thing I would change (but it's a matter of personal style maybe) is that extract a function to get the `path` and another to get the `text`. That way, `flag_step_sentence` will be smaller and simpler.

rubhanazeem: Yeah will do that. Codeclimate is also complaining about that

src/api/app/components/workflow_artifacts_per_step_component.rb
```diff
@@ -34,6 +37,37 @@ def call
...
+      text = "#{flag[:project]}/#{flag[:package]}"
+    else
+      path = helpers.project_show_path(project: flag[:project])
+      text = "#{flag[:project]}"
+    end
+
+    tag.li do
```
dmarcoux, line 61: This will quickly become hard to parse/read once there are a few lines, especially if most of them are for a repository and architecture. At the very least, we should work with text hierarchy to put emphasis on the important information. Something even better would be to have a proper interface for this. It doesn't have to be exactly like what we have in the repository reviews of a project/package, but it is a good inspiration:

interface

rubhanazeem: Do users have a large number of flags in their set_flag step? Because the table makes sense when the data is big. The current solution is easily readable now even with 5 flags. I can bold the tags etc but table seems extra if the information about set flags is not too large

dmarcoux: I don't have statistics on how many flags users set with the `set_flags` step. I agree that a table is overkill for cases with only a few flags. As long as the text hierarchy is improved, it should be good enough for now.

dmarcoux:
Would it be possible to add one example of artifacts for `set_flags` here:

https://github.com/openSUSE/open-build-service/blob/cd6c23ca2549cd307488bda903079dec8e6f863c/src/api/lib/tasks/dev/workflows.rake#L38-L43

This would make it easier to see how it renders would having to go through the whole process of running a workflow locally.

danidoni:
:thought_balloon: The only thing missing for me would be a link in the generated artifact pointing directly to its [repository page](https://obs-reviewlab.opensuse.org/rubhanazeem-artifacts-set-flag/repositories/home:Admin/pariatur)

src/api/app/components/workflow_artifacts_per_step_component.rb
```diff
@@ -34,6 +37,33 @@ def call
...
+  def list_of_flags(flags)
+    flags.each do |flag|
+      concat(flag_step_sentence(flag))
+    end
+  end
+
+  def flag_step_sentence(flag)
+    path_details = package_or_project_path(flag)
+
+    tag.li do
```
saraycp, line 55: Comparing with other artifacts, this is how it looks like:

Screenshot 2023-10-26 at 10-14-47 Open Build Service

I'd say there are too many **bold** words in the flag sentence.

It's also a bit hard for me to understand the sentence:
> Set flag tag: build, status: enable for [home:Admin/et](http://localhost:3000/package/show/home:Admin/et), repository: openSUSE_Tumbleweed, architecture: x86_64.

What about?

- Set flag **build** enabled on project/package  [home:Admin/et](http://localhost:3000/repositories/home:Admin/et) for repository openSUSE_Tumbleweed and architecture x86_64.

- Set flag **publish** disabled on project  [home:Admin](http://localhost:3000/repositories/home:Admin) for repository openSUSE_Tumbleweed and architecture x86_64.

I also propose that the links don't go to the project or package overview but to their repositories tab.

rubhanazeem: This is done. @danidoni now the project/package link is pointing towards the repositories page.

src/api/app/components/workflow_artifacts_per_step_component.rb
```diff
@@ -34,6 +37,33 @@ def call
...
+      concat('Set flag ')
+      concat(tag.b("#{flag[:type]} "))
+      concat("#{flag[:status]}")
```
saraycp, line 58:
```suggestion
      concat("#{flag[:status]}d")
```

To have enable**d** and desable**d**

saraycp (approved):

PR #15087, opened 2023-10-20
Ignore errors early, to avoid creating extra WorkflowRuns

hellcp-work:
This makes sure to weed out the events and actions that shouldn't have WorkflowRuns created for them.

To test this:
1. Create a token and create a webhook on the scm side associated with it
2. Trigger an action that falls outside of the accepted events and actions
3. See it return a status 200 with a note about the event and/or action not being supported

Additionally, the extractor error can be triggered earlier, because it indicates missing headers for events, which needs to be detected before we can ok the action

### DIFF
--- a/src/api/app/controllers/trigger_workflow_controller.rb
+++ b/src/api/app/controllers/trigger_workflow_controller.rb
@@ -8,6 +8,7 @@ class TriggerWorkflowController < TriggerController
   before_action :set_scm_event
   before_action :set_scm_extractor
   before_action :extract_scm_webhook
+  before_action :verify_event_and_action
   before_action :create_workflow_run
   before_action :validate_scm_event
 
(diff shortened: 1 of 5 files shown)


src/api/app/controllers/trigger_workflow_controller.rb
```diff
@@ -40,10 +41,9 @@ def set_scm_extractor
 
   def extract_scm_webhook
     @scm_webhook = @scm_extractor.call
+    return @scm_webhook if @scm_webhook && @scm_extractor.error_message.blank?
```
eduardoj, line 44:
```suggestion
    return @scm_webhook if @scm_webhook && @scm_extractor.valid?
```
What do you think?

eduardoj (changes requested):

eduardoj (approved):

PR #15092, opened 2023-10-23
Remove the WorkflowRuns with unsupported events

hellcp-work:
This removes all of the WorkflowRuns with unsupported events from both GitHub/Gitea and Gitlab.

In order to test:
1. Make sure you have any unsupported event in WorkflowRuns (for example with `WorkflowRun.where.not(hook_event: [*SCMWebhookEventValidator::ALLOWED_GITHUB_AND_GITEA_EVENTS, *SCMWebhookEventValidator::ALLOWED_GITLAB_EVENTS])`)
2. Run the data migration
3. Check the unsupported events again

This should be merged after #15087

### DIFF
--- a/src/api/db/data/20231023110428_remove_unsupported_hook_events.rb
+++ b/src/api/db/data/20231023110428_remove_unsupported_hook_events.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class RemoveUnsupportedHookEvents < ActiveRecord::Migration[7.0]
+  def up
+    WorkflowRun.where.not(hook_event: [*SCMWebhookEventValidator::ALLOWED_GITHUB_AND_GITEA_EVENTS, *SCMWebhookEventValidator::ALLOWED_GITLAB_EVENTS]).destroy_all
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/src/api/db/data_schema.rb
+++ b/src/api/db/data_schema.rb
@@ -1,1 +1,1 @@
-DataMigrate::Data.define(version: 20231016113740)
+DataMigrate::Data.define(version: 20231023110428)


src/api/db/data/20231023110428_remove_unsupported_hook_events.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class RemoveUnsupportedHookEvents < ActiveRecord::Migration[7.0]
+  def up
+    WorkflowRun.where.not(hook_event: [*SCMWebhookEventValidator::ALLOWED_GITHUB_AND_GITEA_EVENTS, *SCMWebhookEventValidator::ALLOWED_GITLAB_EVENTS]).destroy_all
```
eduardoj, line 5:
```suggestion
    WorkflowRun.where.not(hook_event: [*SCMWebhookEventValidator::ALLOWED_GITHUB_AND_GITEA_EVENTS, *SCMWebhookEventValidator::ALLOWED_GITLAB_EVENTS]).delete_all
```
`delete_all` will trigger the deletion strategy defined in the associations. See: https://github.com/openSUSE/open-build-service/blob/0778664120b4fde2a3e4d43c868afe5aa7e147aa/src/api/app/models/workflow_run.rb#L26-L28

`destroy_all`  won't. Any reason to not trigger them?

eduardoj (changes requested):

src/api/db/data/20231023110428_remove_unsupported_hook_events.rb
dmarcoux: Yes and we want the callbacks, so we should stick with `destroy_all`.

eduardoj: Ups. My fault. Thanks for double checking!

eduardoj (approved):

PR #15093, opened 2023-10-23
Unify description text for decision events

eduardoj:
Seen while working on #15084. This will help to decrease the lines of code of the method.

### DIFF
--- a/src/api/app/components/notification_action_description_component.rb
+++ b/src/api/app/components/notification_action_description_component.rb
@@ -32,12 +32,8 @@ class NotificationActionDescriptionComponent < ApplicationComponent
       # TODO: Remove `Event::CreateReport` after all existing records are migrated to the new STI classes
       when 'Event::CreateReport', 'Event::ReportForProject', 'Event::ReportForPackage', 'Event::ReportForComment', 'Event::ReportForUser'
         "'#{@notification.notifiable.user.login}' created a report for a #{@notification.event_payload['reportable_type'].downcase}. This is the reason:"
-      when 'Event::ClearedDecision'
-        class_name = @notification.notifiable.reports.first.reportable.class.name.downcase
-        "'#{@notification.notifiable.moderator.login}' decided to clear the report about the #{class_name}. This is the reason:"
-      when 'Event::FavoredDecision'
-        class_name = @notification.notifiable.reports.first.reportable.class.name.downcase
-        "'#{@notification.notifiable.moderator.login}' decided to favor the report about the #{class_name}. This is the reason:"
+      when 'Event::ClearedDecision', 'Event::FavoredDecision'
+        event_decision_description
       end
     end
   end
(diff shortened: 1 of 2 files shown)


src/api/app/components/notification_action_description_component.rb
```diff
@@ -54,4 +50,10 @@ def bs_request
                       @notification.notifiable.commentable
                     end
   end
+
+  def event_decision_description
+    decision_text = @notification.event_type.match(/Event::(\w+?)edDecision/)[1].downcase
```
danidoni, line 55: :thought_balloon: I find this code harder to follow than the previous one, it was dead easy to understand. In order of reducing the number of lines of that method, I think the right change is to get rid of the event case (I'm not saying I know how to do it, though...).

dmarcoux: OOP to the rescue. Each event is an object, so let's have a method on the event which is generating this notification action description. The downside would be to put some view logic in the models. I have to admit that I don't have a problem with this huge `case` statement, since it's quite simple to follow what happens.

eduardoj: We had OOP in mind, but wanted to give the regular expression a try.

Changed to OOP. Please take a look again.

eduardoj: On the first approach tests failed with:

```
  1) NotificationActionDescriptionComponent when the notification is for an Event::FavoredDecision renders the information about the favored decision
     Failure/Error: raise "Unexpected payload_keys :#{na.sort.join(', :')} (#{attribs.inspect}) provided during '#{self.class.name}' event creation. "
     
     RuntimeError:
       Unexpected payload_keys :fake ({"fake"=>"payload"}) provided during 'Event::FavoredDecision' event creation. 
     # ./app/models/event/base.rb:138:in `check_left_attribs'
     # ./app/models/event/base.rb:152:in `set_payload'
     # ./app/models/event/base.rb:124:in `initialize'
     # ./app/models/notification.rb:23:in `event'
     # ./app/components/notification_action_description_component.rb:56:in `event_decision_description'
     # ./app/components/notification_action_description_component.rb:36:in `block in call'
     # ./app/components/notification_action_description_component.rb:17:in `call'
     # ./spec/components/notification_action_description_component_spec.rb:213:in `block (3 levels) in <top (required)>'
```

I found we could use `@notification.notifiable.kind`, and adapt it. Please take a look again.

src/api/app/components/notification_action_description_component.rb
```diff
@@ -54,4 +50,10 @@ def bs_request
                       @notification.notifiable.commentable
                     end
   end
+
+  def event_decision_description
+    decision_text = @notification.notifiable.kind.sub(/ed$/, '')
```
danidoni, line 55: I find this even less clear than  the previous one, sorry... To me `OOP` to the rescue meant something else...

eduardoj: I see. Is it now better?

danidoni: My point is: in order to save a case branch, we're getting more complex code in exchange. I think this change is not worth it, sorry. Those two `when` branches were easy to understand as they were before...

eduardoj:
This approach is not convincing. Trying another approach.

PR #15094, opened 2023-10-23
Workflow_configuration: show the file content

ncounter:
Show all the information available about the `Workflows Configuration File` for a `workflow_run`.
**Existing information**
- Path
- URL

**Additional information**
- File Content

If the workflow configuration file was ~received, parsed and stored by the scm service~ downloadable, it renders now the content of the file in the dedicated workflow run configuration sub-tab.

## Before
image

## After

Screenshot 2023-10-26 at 18-15-43 Open Build Service

## Testing tips

- Go to review-app or run `rake dev:test_data:create` in dev. env.
- Open any workflow run to see the content of the _Workflow Configuration_ tab (Profile > Manage Tokens > Workflow runs > click on one of the workflow runs)
- [Workflow run with ID 7](https://obs-reviewlab.opensuse.org/ncounter-workflow-config-ui/my/tokens/1/workflow_runs/7#tab-pane-workflow-configuration-tab-content7) has URL and configuration content stored.
- [Workflow run with ID 8](https://obs-reviewlab.opensuse.org/ncounter-workflow-config-ui/my/tokens/1/workflow_runs/8#tab-pane-workflow-configuration-tab-content8) doesn't have URL, PATH or configuration content stored. Emulates the old workflow runs.
- [The rest of the workflow runs](https://obs-reviewlab.opensuse.org/ncounter-workflow-config-ui/my/tokens/1/workflow_runs/1#tab-pane-workflow-configuration-tab-content1) have PATH and configuration content stored.

### DIFF
--- a/src/api/app/components/workflow_run_detail_component.rb
+++ b/src/api/app/components/workflow_run_detail_component.rb
@@ -28,10 +28,18 @@ class WorkflowRunDetailComponent < ApplicationComponent
   def workflow_configuration_data
     return content_tag(:p, 'This information is not available.') unless workflow_run.configuration_source
 
-    content_tag(:h5, "Workflow Configuration File #{workflow_run.workflow_configuration_url.present? ? 'URL' : 'Path'}").concat(
-      content_tag(:pre,
-                  workflow_run.configuration_source,
-                  class: 'border p-2')
+    output = content_tag(:h5, "Workflow Configuration File #{workflow_run.workflow_configuration_url.present? ? 'URL' : 'Path'}").concat(
+      content_tag(:pre, workflow_run.configuration_source, class: 'border p-2')
     )
+
+    if workflow_run.workflow_configuration
+      output.concat(
+        content_tag(:h5, 'Workflow Configuration').concat(
+          content_tag(:pre, workflow_run.workflow_configuration, class: 'border p-2')
+        )
+      )
+    end
+
+    output
   end
 end
--- a/src/api/lib/tasks/dev/workflows.rake
+++ b/src/api/lib/tasks/dev/workflows.rake
@@ -24,6 +24,8 @@ namespace :dev do
       create(:workflow_run, :succeeded, :tag_push, token: workflow_token)
       create(:workflow_run, :succeeded, token: workflow_token)
       create(:workflow_run, :succeeded, :pull_request_closed, token: workflow_token)
+      create(:workflow_run, :with_url, token: workflow_token)
+      create(:workflow_run, :without_configuration_data, token: workflow_token)
 
       # GitLab
       create(:workflow_run_gitlab, token: workflow_token)
--- a/src/api/spec/components/workflow_run_detail_component_spec.rb
+++ b/src/api/spec/components/workflow_run_detail_component_spec.rb
@@ -1,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.describe WorkflowRunDetailComponent, type: :component do
+  context 'with a workflow run with configuration data path' do
+    let(:workflow_run_with_path) { create(:workflow_run) }
+
+    before do
+      render_inline(described_class.new(workflow_run: workflow_run_with_path))
+    end
+
+    it { expect(rendered_content).to have_text('Workflow Configuration File Path') }
+    it { expect(rendered_content).to have_text('Workflow Configuration') }
+  end
+
+  context 'with a workflow run with configuration data URL' do
+    let(:workflow_run_with_url) { create(:workflow_run, :with_url) }
+
+    before do
+      render_inline(described_class.new(workflow_run: workflow_run_with_url))
+    end
+
+    it { expect(rendered_content).to have_text('Workflow Configuration File URL') }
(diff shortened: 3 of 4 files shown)


src/api/app/components/workflow_run_detail_component.rb
```diff
@@ -24,14 +24,33 @@ def parse_payload(workflow_run)
...
-                  workflow_run.configuration_source,
-                  class: 'border p-2')
-    )
+  def workflow_configuration_path
+    workflow_run.workflow_configuration_path
+  end
+
+  def workflow_configuration_url
+    workflow_run.workflow_configuration_url
+  end
```
danidoni, line 33: :thought_balloon: Those two functions don't provide much value, I would drop them

ncounter: done

src/api/app/components/workflow_run_detail_component.rb
```diff
@@ -24,14 +24,25 @@ def parse_payload(workflow_run)
...
+    [workflow_run.workflow_configuration, workflow_run.workflow_configuration_url, workflow_run.workflow_configuration_path].all?(&:blank?)
+  end
+
+  def parse_workflow_configuration_yaml(blob_content)
+    return if blob_content.blank?
+
+    # parse the workflow_configuration file content previously stored
+    yaml_parsed = YAML.safe_load(blob_content)
+    # extract only the slice of interest
+    yaml_parsed.dig('payload', 'blob', 'rawLines') if yaml_parsed.is_a?(Hash)
```
danidoni, line 46: In my tests, this step is not necessary, there is no `payload -> blob -> rawlines` nodes if it's a Hash...

src/api/app/components/workflow_run_detail_component.html.haml
```diff
@@ -35,7 +35,20 @@
         = pretty_request_payload
   .tab-pane.fade{ id: "workflow-configuration-tab-content#{id}", role: 'tabpanel',
                   aria: { labelledby: "workflow-configuration-tab#{id}" } }
-    = workflow_configuration_data
+    - if configuration_source_not_available
```
hennevogel, line 38: How can you have a `WorkflowRun` from running a `Workflow` without a configuration? :)

saraycp: This check is necessary for those old workflow runs that didn't store any of those values: `workflow_configuration`, `workflow_configuration_path` or `workflow_configuration_url`. When we introduced them it wasn't possible to backfill those values.

From that moment on, those values are no longer empty unless they aren't downloadable (we create the workflow run before we check). See this line and other similar ones in the same file:
https://github.com/openSUSE/open-build-service/blob/cfe9121d0dbee5e2643f88278e6ad0f226a1898f/src/api/app/services/workflows/yaml_downloader.rb#L74

src/api/app/components/workflow_run_detail_component.rb
saraycp: Using the value as it is (WorkflowRun#workflow_configuration) in the view, it's displayed correctly.
I don't know why parsing is needed.

saraycp: I also get nil with `payload -> blob -> rawlines`.

I'll go without parsing until we have a good reason for that.

saraycp:
@ncounter I explain what I did:

- I simplified the code as I didn't find a reason for parsing the file content.
- I keep the methods we had before and simply concatenate the new part (the configuration file content).
- As the headers get the eyes attention (big and green) compared to the text inside the tab, I prefer to use describing headers in case the  text in the tab goes unnoticed.

src/api/app/components/workflow_run_detail_component.rb
ncounter: Interesting, in my live test (against GitHub for instance), I was receiving the whole structure with that tree and I had to extract the part I needed. 🤔 😕 
Anyway, if the two of you had different output, the majority is set. 👍

danidoni:
Could we add some test data having a workflow configuration with an url?

src/api/app/components/workflow_run_detail_component.rb
saraycp: I picked up 10+ random samples from production and all of them look quite "normal":

```
"---\npr:\n  steps:\n  - branch_package:\n      source_project: [...]
```
```
"version: '1.1'\ntestbuild:\n  steps:\n    - branch_package:\n        source_project: [...]
```
```
"factory-rebuild:\n  steps:\n    - trigger_services:\n       [...]
```

src/api/app/components/workflow_run_detail_component.html.haml
saraycp: The code has been refactored, but it still checks the existence of those values upfront.

danidoni (approved):

saraycp:
I added a spec for the WorkflowRunDetails component.

**codecov/path** still complains about `lib/tasks/dev/workflows.rake` but I don't see the point of testing that. I'd say we can ignore it.

danidoni (approved):

PR #15096, opened 2023-10-23
Switch allow_remote_projects default

hennevogel:
There is only one place in the app where we look at remote project links. All the others turned it off, let's make this the default.


hennevogel:
> There is also this one occurrence:

This is a test, which still works right?

eduardoj (approved):

PR #15097, opened 2023-10-23
Make includeallpackages a boolean keyword argument

hennevogel:
Instead of a single option opts hash...

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -564,7 +564,7 @@ class SourceController < ApplicationController
   def project_command_release
     params[:user] = User.session!.login
 
-    @project = Project.get_by_name(params[:project], includeallpackages: 1)
+    @project = Project.get_by_name(params[:project], include_all_packages: true)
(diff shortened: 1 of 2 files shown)


src/api/app/controllers/source_controller.rb
```diff
@@ -564,7 +564,7 @@ def project_command_undelete
   def project_command_release
     params[:user] = User.session!.login
 
-    @project = Project.get_by_name(params[:project], includeallpackages: 1)
+    @project = Project.get_by_name(params[:project], includeallpackages: true)
```
eduardoj, line 567:
```suggestion
    @project = Project.get_by_name(params[:project], include_all_packages: true)
```
What about to make it more clear the name of the variable?

hennevogel: not sure if this makes anything more clear, but why not...

eduardoj: Yes, it makes things more clear. Thanks!

eduardoj (approved):

PR #15101, opened 2023-10-23
Simplify Package.get_by_project_and_name

hennevogel:
Just use speaking variables, return early, simpler nil checking etc.


hennevogel:
more to refactor with the `Project.find_package` logic...

PR #15102, opened 2023-10-23
Drop check_update_project option from Project.find_package

hennevogel:
And make this it's own method.  We only use it once throughout the app  (BranchPackage::CheckForUpdate).
 
With this we can simplify the logic in Package.get_by_project_and_name


hennevogel:
Depends on #15099

hennevogel:
apparently there is even more loops in `Project.update_instance`, I give up for today...

PR #15105, opened 2023-10-24
Make the token naming consistent

### LINKED ISSUE #15078: Inconsistent naming for service/runservice tokens
# Issue Description

The runservice/service type tokens are named inconsistently in the API:
1. When creating a token via `POST /person/$name/token?type=runservice` you have to specify `runservice` as the type
2. When querying tokens via `GET /person/$name/token` they appear as `kind="service"`:

This difference makes it inconvenient to use in API wrappers as you have to handle both values as the same one.

Expected Result
============

The token kind name should be the same.

danidoni:
Because when we create the token, we have to tell the token is of type=runservice, but when we ask about the tokens a user has, we get that same token as kind=service instead.

Fixes #15078

### DIFF
--- a/src/api/app/views/person/token/index.xml.builder
+++ b/src/api/app/views/person/token/index.xml.builder
@@ -1,6 +1,7 @@
 xml.directory(count: @list.length) do |dir|
   @list.each do |token|
-    p = { id: token.id, string: token.string, kind: token.token_name, description: token.description, triggered_at: token.triggered_at }
+    token_name = token.token_name.gsub('service', 'runservice') # To make token naming consistent: we create the token as runservice
+    p = { id: token.id, string: token.string, kind: token_name, description: token.description, triggered_at: token.triggered_at }
     if token.package
       p[:project] = token.package.project.name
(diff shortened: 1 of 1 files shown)


src/api/app/views/person/token/index.xml.builder
```diff
@@ -1,6 +1,7 @@
 xml.directory(count: @list.length) do |dir|
   @list.each do |token|
-    p = { id: token.id, string: token.string, kind: token.token_name, description: token.description, triggered_at: token.triggered_at }
+    token_name = token.token_name.gsub('service', 'runservice') # To make token naming consistent: we create the token as runservice
```
eduardoj, line 3:
```suggestion
    token_name = token.token_name.sub('service', 'runservice') # To make token naming consistent: we create the token as runservice
```
I would use `sub` instead of `gsub`. We don't expect to substitute more ocurrences.

danidoni:
Hm, I created the PR against the wrong remote. I opened a new one https://github.com/openSUSE/open-build-service/pull/15109

PR #15110, opened 2023-10-24
Reduce number of lines of `call` method

eduardoj:
This way we comply with the requirements of codeclimate.

Related to #15084.


dmarcoux (approved):
The real issue here is the cyclomatic complexity. So if the changes fix the CodeClimate issue about the size of the `call` method, it's at least a start somehow. Let's move on to other things.

PR #15113, opened 2023-10-24
Send email notifications for failed workflow runs

eduardoj:
Enables the email channel and sets up everything to allow users to receive email notifications regarding failed workflow runs.

### DIFF
--- a/src/api/app/models/event/workflow_run_fail.rb
+++ b/src/api/app/models/event/workflow_run_fail.rb
@@ -2,9 +2,15 @@ module Event
   class WorkflowRunFail < Base
     self.message_bus_routing_key = 'workflow_run.fail'
     self.description = 'Workflow run has failed'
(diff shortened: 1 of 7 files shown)


src/api/app/components/notification_avatars_component.rb
```diff
@@ -20,6 +20,8 @@ def avatar_objects
                           [User.find(@notification.event_payload['user_id'])]
                         when 'Decision'
                           [User.find(@notification.event_payload['moderator_id'])]
+                        when 'WorkflowRun'
+                          [Token.find(@notification.event_payload['token_id'])&.executor]
```
danidoni, line 24:
```suggestion
                          [Token.find(@notification.event_payload['token_id'])&.executor].compact
```
In order to remove the nil that will appear if there is no executor or no token

danidoni (approved):

Issue #12813, opened 2022-07-15
[production][trigger#create] undefined method `start_with?' for nil:NilClass

      return name if name.start_with?('_patchinf...


danidoni:
[See this exception on Errbit](http://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/6277675584bf46053ee62ba3)
# undefined method `start_with?&#39; for nil:NilClass

      return name if name.start_with?(&#39;_patchinfo:&#39;, &#39;_product:&#39;)
                         ^^^^^^^^^^^^ #
### URL ###
[https://api.opensuse.org/trigger/rebuild?project=KDE%3AUnstable%3AApplications&amp;repository=KDE_Unstable_Frameworks_openSUSE_Factory](https://api.opensuse.org/trigger/rebuild?project=KDE%3AUnstable%3AApplications&amp;repository=KDE_Unstable_Frameworks_openSUSE_Factory)"
### Where ###
trigger#create

### Occured ###
May 08  6:46am

### Similar ###
4

## Params ##
~~~
{
  &quot;action&quot; =&gt; &quot;create&quot;,
  &quot;controller&quot; =&gt; &quot;trigger&quot;,
  &quot;project&quot; =&gt; &quot;KDE:Unstable:Applications&quot;,
  &quot;repository&quot; =&gt; &quot;KDE_Unstable_Frameworks_openSUSE_Factory&quot;
}
~~~

## Session ##
~~~
{}
~~~

## Backtrace ##
~~~
11:  app/models/concerns/multibuild_package.rb -> **striping_multibuild_suffix**
159:  app/models/package.rb -> **get_by_project_and_name**
19:  app/controllers/concerns/triggerable.rb -> **set_package**
427:  /GEM_ROOT/gems/activesupport-6.1.5.1/lib/active_support/callbacks.rb -> **block in make_lambda**
198:  /GEM_ROOT/gems/activesupport-6.1.5.1/lib/active_support/callbacks.rb -> **block (2 levels) in halting**
34:  /GEM_ROOT/gems/actionpack-6.1.5.1/lib/abstract_controller/callbacks.rb -> **block (2 levels) in &lt;module:Callbacks&gt;**
199:  /GEM_ROOT/gems/activesupport-6.1.5.1/lib/active_support/callbacks.rb -> **block in halting**
512:  /GEM_ROOT/gems/activesupport-6.1.5.1/lib/active_support/callbacks.rb -> **block in invoke_before**
[...]

Issue #2094, opened 2016-08-30
create new role that allows to trigger rebuilds


lnussel:
OBS all too often needs somebody who can manually trigger rebuilds. That role should not necessarily have full access to modify packages or review requests. Therefore separate from maintainer and reviewer.
Examples where this would be useful are the distro projects, staging and ports.

hennevogel:
Can you explain why only triggering builds is needed?

lnussel:
random build failures spoil the fun. I check in stuff, obs works on it over night or weekend and produces random build failures. If I have to review and trigger those myself Monday morning we lose time. If there was a role for triggering rebuilds more community person could help.

adrianschroeter:
Having random build failures means you have a not maintainable distribution IMHO. I am not sure if we want to have a role for that. In first place we should find the reason for random failures and work on that.

Has anything changed lately that this is needed suddenly? Any new errors which do appear?

lnussel:
you are welcome to play staging master for a week or so

DimStar77:
random build failures include: syswrite: no disk space

coolo:
those are actually supposed to be handled as bad jobs. build greps for errors like that.

adrianschroeter:
yep, do you have examples where it did not work?

However, you should add constraints to avoid wasting our build power by too many attempts just by accident finding a proper build host.


(seems my mail answer is lost .... )

lnussel:
Result of tonight's Leap stagings are failures of gdb, graphviz, libkolab-qt5, java-1_7_0-openjdk-bootstrap. Only in the graphviz looks like a worker related issue, the other are random build failures due to races in make -j or whatever.

In graphviz case:
[29380s] ### WATCHDOG MARKER END ###
[29380s] No buildstatus set, either the base system is broken (kernel/initrd/udev/glibc/bash/perl)
[29380s] or the build host has a kernel or hardware problem...


Job seems to be stuck here, killed. (after 28800 seconds of inactivity)

adrianschroeter:
can you please store the entire build log the next time when it failed? it is gone now...

adrianschroeter:
On Montag, 24. Oktober 2016, 08:04:07 CEST wrote Dominique Leuenberger:
> random build failures include: syswrite: no disk space

this means you lack _constraints in first place.

But IMHO we have also code which should mark this job as badhost result.
Do you have an example where it did not work?


-- 

Adrian Schroeter
email: adrian@suse.de

[...]

PR #15119, opened 2023-10-26
[api] fix trigger route behaviour

adrianschroeter:
- fail when wrong operation type is used (instead of executing the token operation)
- allow to use /trigger without specifing the operation type (is defined by the token)
- verify given project/package parameter with possible token package
- fix triggering rebuilds and releases of entire projects
- improve test coverage and api docu
   - A preview of the changes in the api documentation can be seen with the review-app [here](https://obs-reviewlab.opensuse.org/adrianschroeter-fix_trigger_operations/apidocs/index#/Trigger/post_trigger__operation_).


eduardoj (changes requested):
This would require changes in the current API documentation...

dirkmueller:
@eduardoj are you saying `src/api/public/apidocs/paths/trigger_operation.yaml` needs to be modified?

adrianschroeter:
I just renamed it now, given that the new endpoint is just a different name without the so far missing operation type check.

src/api/app/controllers/trigger_controller.rb
```diff
@@ -66,10 +67,12 @@ def pundit_user
...
   def set_project_name
     @project_name = params[:project]
+    raise InvalidParameterError if @project_name.present? && @token.package && @token.package.project.name != @project_name
   end
 
   def set_package_name
     @package_name = params[:package]
-    raise MissingPackage if @package_name.blank? && @token.package_id.nil? && @token.token_name.in?(['rebuild', 'release', 'service'])
+    raise InvalidParameterError if @package_name.present? && @token.package && @token.package.name != @project_name
+    raise MissingPackage if @package_name.blank? && @token.package_id.nil? && @token.token_name.in?(['service'])
```
dmarcoux, line 76: What is the reason behind removing `release` and `rebuild` in the `in?(...)` method call?

dirkmueller: because the purpose here is to implement rebuild for projects, so bug #12813 was not fixing it, it was forbidding the feature from being used.

eduardoj:
> I just renamed it now, given that the new endpoint is just a different name without the so far missing operation type check.

@adrianschroeter, I see you added changes to the current documentation. Thanks.

But renaming the `src/api/public/apidocs/paths/trigger_operation.yaml` to the `src/api/public/apidocs/paths/trigger.yaml` file makes Swagger-UI not able to find that file. See the `OBS-v2.10.50.yaml` file: https://github.com/openSUSE/open-build-service/blob/929e9674138cf91ed6d83a591e7f35d438c01534/src/api/public/apidocs/OBS-v2.10.50.yaml#L503-L504

I can imagine solving this issue using one of these two options:
1. Renaming back the file to `src/api/public/apidocs/paths/trigger_operation.yaml`. This is fine because we define the `{operation}` path parameter as "not required".
1. Keep the file renaming. We should change the reference to the file from `$ref: 'paths/trigger_operation.yaml'` to `$ref: 'paths/trigger.yaml'`

I prefer the first option, renaming back the file to `src/api/public/apidocs/paths/trigger_operation.yaml`.. I recognize this contradicts with my first comment: _Not only that, a new endpoint has been introduced: POST /trigger. This should also be documented, most likely in a new file named src/api/public/apidocs/paths/trigger.yaml._. This solution you propose convinces my. Sorry if my first comment was misleading.

src/api/app/controllers/trigger_controller.rb
dmarcoux: > it was forbidding the feature from being used.

This is false.

Before, whenever a rebuild token without a package was triggered to a rebuild a project, the endpoint crashed with a 500. Passing a package to the endpoint in the query string would make it work. So whenever a package wasn't passed, the change recently introduced made the endpoint respond with a 400 instead of a 500. The behavior I am describing here is also what was already included in the API documentation.

api-documentation

adrianschroeter: the endpoint just crashed because of the introduced concern code. It should have worked before.

adrianschroeter:
okay, will rename the docu file back to origin name.

src/api/app/controllers/trigger_controller.rb
dirkmueller: > > it was forbidding the feature from being used.
> This is false.

> Before, whenever a rebuild token without a package was triggered to a rebuild a project, the endpoint crashed with a 500.

Yes, previously it was failing due to a bug, now its forbidding to use that. Thats what I said. 


> Passing a package to the endpoint in the query string would make it work.

we don't want to pass a package, we want to pass a project. There is a 7 year old feature request for it: #2094 

> So the feature can change, sure, but this gives you the reasoning behind the previous change.

You asked why the code got removed, not me ;)

eduardoj:
I added a commit to fix the RuboCop offenses. Let's see if the previous changes make some other fail.

eduardoj (changes requested):
The following specs must be adapted:
- `spec/controllers/person/token_controller_spec.rb`
- `spec/controllers/trigger_controller_spec.rb`
- `spec/controllers/webui/users/token_triggers_controller_spec.rb`

adrianschroeter:
@eduardoj can you update the specs please? I do not have currently an environment to run these.

eduardoj:
I added a commit to get rid of the codeclimate complaint.

I'll try to adapt the failing tests:
- [ ] `spec/controllers/person/token_controller_spec.rb`
- [ ] `spec/controllers/trigger_controller_spec.rb`
- [ ] `spec/controllers/webui/users/token_triggers_controller_spec.rb`
- [ ] `spec/controllers/trigger_workflow_controller_spec.rb`

hennevogel:
Superseded by #15261

PR #15120, opened 2023-10-26
Move check_access into instantiation callback

hennevogel:
We already have a default_scope to check for the access Flag hack while finding Projects. And then we had Project.check_access? all over the app where we maybe instantiate a Project through some other means (associations, arel etc.)

Instead let's do this in a after_initialize callback.


hennevogel:
no time to follow up on this. I'll keep the branch around...

PR #15128, opened 2023-10-27
Fix expression to exclude dev rake tasks from codecov

eduardoj:
Fixes the previous PR https://github.com/openSUSE/open-build-service/pull/15124

According to the [documentation](https://docs.codecov.com/docs/ignoring-paths), the key is `ignore:`.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,9 +5,8 @@ coverage:
         target: auto
         base: auto
         threshold: 1
-        paths:
-        - !src/api/lib/tasks/dev/
-
+ignore:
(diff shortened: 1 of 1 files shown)


saraycp:
According to the [documentation](https://docs.codecov.com/docs/ignoring-paths), the key is `ignore:`.
Trying with that...

saraycp:
All the checks passed using
```
ignore:
  - "src/api/lib/tasks/dev/"
 ```
I removed all the commits from the other PR and made it ready for review.

danidoni (approved):

PR #15129, opened 2023-10-27
[backend][api] Fix fetching of remote /distributions.xml and image te…

adrianschroeter:
…mplates

The frontend code can not do this correctly itself, since it does not know anything about network detailed setup (eg. credentials and routeing).

Therefore route these via the source server.

Remaining issue is the remote support for remote maintenance statistics. We may just drop this code and drop app/models/project/remote_url.rb ?


hennevogel:
> maintenance statistics... We may just drop this code...

For sure. No one has accessed this in the last 2 years on both instances...

eduardoj:
These changes would require to document the two new API endpoints created:
- `GET /source/:project/_remotedistributions`
- `GET /source/:project/_remoteimagetemplates`

eduardoj (dismissed):
See comment above.

hennevogel:
> These changes would require to document the two new API endpoints introduced:

This is just backend API, the OBS API stays the same.

PR #15153, opened 2023-11-02
Autocorrect some RuboCop offenses (III)

eduardoj:


danidoni:
Can we do something about the issues CodeClimate is complaining about?

eduardoj:
> Can we do something about the issues CodeClimate is complaining about?

codeclimate reports 9 issues fixed and 2 new issues to fix. Those 2 new issues have poped up after refactoring the code. I would say they are in fact the same.

danidoni (approved):

eduardoj:
@danidoni, I need to rebase on master.

After that, new `Lint/RedundantCopDisableDirective` appeared. I added a commit to fix them.

I also added the las commit, to improve the definition of migration for the "Isolated Migrations" GitHub action.

Please take a look again.

danidoni (approved):

eduardoj:
I approved the two codeclime issues. Merging.

PR #15160, opened 2023-11-03
Address some RuboCop offenses from the main `.rubocop.yml` file

eduardoj:
These changes contain a renaming of a `binaries.spec` file to a `package_controller_binaries_spec.rb` file. The `binaries.spec` file was introduced in 77dd5dc28bc5a84bdb44860a68b7c3fa6b5ada64. Fixing the extension of the file made the tests run.

The tests were adapted and their missing cassettes added.

Two actions of these tests (`binary` and `binaries`) were moved out from the controller in aa8b2582db157ae8bd15215abea4207cbada0897 and 00fe3d193233690afb0a7d31c1f42622848c2816. Tests for these two actions were not added.


src/api/spec/controllers/webui/package_controller_binaries_spec.rb
```diff
@@ -0,0 +1,104 @@
+require 'webmock/rspec'
+require 'rails_helper'
+
+RSpec.describe Webui::PackageController, :vcr do
```
krauselukas, line 4: @eduardoj hmm, but this is testing the `webui/package_controller.rb`, also the endpoints `#dependency` and `#wipe_binaries` are not part of the `webui/packages/binaries_controller.rb`.

src/api/spec/controllers/webui/package_controller_binaries_spec.rb
krauselukas, line 1: @eduardoj shouldn't that be `webui/packages/binaries_controller_spec.rb`?

src/api/spec/controllers/webui/package_controller_binaries_spec.rb
krauselukas: something looks wrong here, i think those test belong into the `webui/package_controller_spec.rb`

eduardoj: @krauselukas, I just wanted to keep separated the file with the tests of `#dependency` and `#wipe_binaries`, and to rename it to something valid.

I also followed the naming convention described by the [RSpec/SpecFilePathFormat](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/SpecFilePathFormat) cop. In case we had lots of methods to test in a class, instead of having only one test file (`my_class_spec.rb`) we could have several spec files (`my_class_file1_spec.rb`, `my_class_file2_spec.rb`, ...). That's the reason I decided to rename the `binaries.spec` file to `package_controller_binaries_spec.rb`.

krauselukas: @eduardoj but since this is not testing endpoints defined in the `webui/packages/binaries_controller.rb` but the `webui/package_controller.rb` I would say its wrong. Like if I ever change those endpoints in the `package_controller.rb` I would not search for the corresponding tests in the `package_controller_binaries_spec.rb`. Also from what I understand from the cop it talking about the method name. `my_class_method_spec.rb  # describe MyClass, '#method'`

krauselukas: I know, not introduced by you, but since you touching it, I would move those tests to the `package_controller_spec.rb` instead.

eduardoj: @krauselukas I don't know why you wouldn't search for tests for the `Webui::PackageController` in files with the format `package_controller_*_spec.rb`. It solves the problem introduced when the tests were splitted into the `spec/controllers/webui/package_controller/binaries.spec` file. I don't see a less intrusive alternative.

I don't have any issue with leaving the tests for the `dependency` and `wipe_binaries` in the `package_controller_binaries_spec.rb` file.

From the last sentence you said about cop and the method name, it is still not clear to me if you are suggesting to split the tests into more files or not, like these ones:
- `package_controller_spec.rb`
- `package_controller_dependency_spec.rb`
- `package_controller_wipe_binaries_spec.rb`

I'm also fine with this approach.

It is clear you feel something is wrong with the current approach. Sorry for that. What option are you really suggesting?
1. to keep the tests in one file
2. to split the tests in three files, like above

krauselukas: I'm fine with either moving the tests  `#wipe_binaries` and `#dependency` to `package_controller_spec.rb` or creating `package_controller_dependency_spec.rb` and `package_controller_wipe_binaries_spec.rb`.

krauselukas: Just not as it is right now with `package_controller_binaries_spec.rb`, because there is no method/endpoint in `package_controller.rb` called `binaries`

eduardoj: I created the `package_controller_dependency_spec.rb` and `package_controller_wipe_binaries_spec.rb` files and splitted the tests. Please take a look again.

krauselukas (approved):

PR #15165, opened 2023-11-03
Move `can_decline_request?` logic to a pundit policy

krauselukas:
This code is for authorization and therefore should be in a pundit policy.


src/api/app/policies/bs_request_policy.rb
```diff
@@ -23,6 +23,10 @@ def revoke_request?
     author? && record.state.in?([:new, :review, :declined])
   end
 
+  def decline_request?
+    !author?
```
hennevogel, line 27: This sounds wrong. Everyone except the author can decline a request?

krauselukas: hmm, thats because its used nested under the `can_accept_request?` condition (which should also be a policy) https://github.com/openSUSE/open-build-service/pull/15165/files#diff-79f2aa701b809591100770e467ebdc17cd00bd45c6d900cf0c8f764694122562L22
but maybe that should not be achieved through nesting the two conditions... I will change this

hennevogel: I think you are getting carried away with "validations" in a policy. Policies should be about permission not about correctness.

hennevogel: Follow up PRs are nice, where do you track it?

krauselukas: @hennevogel tracking it here now https://trello.com/c/w1pwXR6l/2439-move-istargetmaintainer-authorization-logic-to-a-pundit-policy

hennevogel (approved):

PR #15174, opened 2023-11-06
Correct manually `RSpec/RepeatedExampleGroupDescription` RuboCop offenses

eduardoj:


src/api/spec/controllers/webui/package_controller_spec.rb
```diff
@@ -589,7 +589,7 @@ def remove_file_post
       login(user)
     end
 
-    context 'when triggering a rebuild fails' do
+    context 'non existant repository' do
```
krauselukas, line 592:
```suggestion
    context 'non existent repository' do
```

krauselukas: this is misspelled in other places of this context as well, maybe you can correct those as well :)

krauselukas:
Rest looks good to me

src/api/spec/controllers/webui/package_controller_spec.rb
eduardoj: That's not the purpose of this pull request, sorry. That can be addressed in a following pull request.

eduardoj: I applied your suggestion. This misspelling was introduced in this pull request. Please take a look again.

The rest of the misspellings could be addressed later on in other pull requests.

krauselukas (approved):

PR #15175, opened 2023-11-06
Fix interconnect

adrianschroeter:


adrianschroeter:
put it into #15129 instead

PR #15180, opened 2023-11-06
Make attrib tables static for RSpec tests

eduardoj:
Now that the creation of several attribute table records are included in the seeds (see #15181), make use of them instead of creating those records before every RSpec example.

Making the attribute tables static, prevents from deleting their records before running every RSpec example.

Some initial measurements of time consumed by the RSpec tests in the CI showed a reduction of ~25%.


bmwiedemann:
It would be good to have the commit message explain why this change is wanted. E.g. what is bad before / good after applying the change.

eduardoj:
> It would be good to have the commit message explain why this change is wanted. E.g. what is bad before / good after applying the change.

I updated the descriptions of the commits and the description of the pull request. @bmwiedemann please take a look again. @dmarcoux, you agreed with Bernhard's comment. Please take a look also.

dmarcoux:
The new commit description is good :+1:

saraycp (approved):

PR #15198, opened 2023-11-14
Fix `GET /configuration` to show default_tracker

eduardoj:
The default_tracker configuration parameter is not shown. It must be shown.

### Before

```bash
> osc -A http://localhost:3000 api '/configuration' | grep default_tracker
>
```

### After

```bash
> osc -A http://localhost:3000 api '/configuration' | grep default_tracker
  <default_tracker>bnc</default_tracker>
>
```


dmarcoux (approved):

eduardoj:
I forgot to fix the API documentation. I modified the last commit to include that change. @dmarcoux please take a look again.

PR #15199, opened 2023-11-14
Remove unnecesssary configuration create

eduardoj:
This should have been created with the seeds. There is no way of removing the configuration record.


eduardoj:
#15200 got merged. This means we still get rid of a database query.

### Before

```
[1] pry(main)> Configuration.gravatar
  Configuration Load (0.6ms)  SELECT `configurations`.* FROM `configurations` ORDER BY `configurations`.`id` ASC LIMIT 1
  Configuration Load (0.7ms)  SELECT `configurations`.* FROM `configurations` ORDER BY `configurations`.`id` ASC LIMIT 1
=> true
[2] pry(main)>
```

### After

```
[1] pry(main)> Configuration.gravatar
  Configuration Load (0.7ms)  SELECT `configurations`.* FROM `configurations` ORDER BY `configurations`.`id` ASC LIMIT 1
=> true
[2] pry(main)>
```

dmarcoux:
Some specs are broken now.

eduardoj:
@dmarcoux, the conditional containing `Configuration.instance_methods` worked before because of the line containing `... if first`. This was done just before checking for `Configuration.instance_methods`.

When the line containing `... if first` is removed, `Configuration.instance_methods` returns empty for some tests.

I made the tests pass with `Configuration.new.methods(...)`. Please take a look again.

dmarcoux (approved):

PR #15201, opened 2023-11-14
Add importmaps rails and turbo rails

krauselukas:
DO NOT MERGE: I would like to test this on our test instance first, to see if it works in production environment.

https://github.com/rails/importmap-rails
https://github.com/hotwired/turbo-rails

Import maps let us import Javascript modules without the need of Webpack
or similar. This is one step towards a more modern Javascript stack
in our project and required in order to use Turbo Rails.

[Turbo](https://turbo.hotwired.dev/) gives you the speed of a single-page web application without having to write any JavaScript. Turbo accelerates links and form submissions without requiring you to change your server-side generated HTML.

As an example, I included a little refactoring in the PR, that moves the inline editing of the project information from using Javascript to Turbo. 

*TODO:*

- [ ] Fix obs-server package
- [ ] Test in test production instance


krauselukas:
> Would it be possible to go further, so that we can test if we can include action cable with turbo. That way we could use the full functionality of it?

Maybe, I dont want to make the scope of the PR to big. I have to fix the obs-server rpm first.

hennevogel:
done differently now

PR #14164, opened 2023-04-14
Convert conversation tab to a separate action

rubhanazeem:
This PR converts conversation tab to a separate action

PR #15202, opened 2023-11-14
Clean up JavaScript leftovers on request redesign

saraycp:
We stopped calling these JS functions in commit [a14e43e](https://github.com/openSUSE/open-build-service/pull/14164/commits/a14e43e442e8e87618c2e521ce51a1acc38d0268), so we can safely remove them.

NOTE: I'm not removing the file because I'm using it on the new PR I'm about to open.


dmarcoux:
Good catch! Is there anything else to do? I see the PR is a draft. Otherwise I would approve and merge.

dmarcoux (approved):

PR #15220, opened 2023-11-16
Distinguish conversation from other history elements

saraycp:
These are the changes so far:

- Nested comments don't show the comment icon. It's displayed only once per comment thread on top of the timeline.
- The name of the commenter and the date are moved inside the comment bubble.
- The comment bubble is moved to the right side of the avatar.
- The comment bubble is more rounded and its arrow is moved to the left.
- The Report and Reply links are moved to the right side, but this comes from PR #15218.
- Any comment/message which is not part of a conversation i.e. the message sent when asking someone to review.
  - doesn't have an avatar
  - uses a smaller font size
  -  is not displayed but can be read when hovering the file icon next to the date.
- Having `wrote` everywhere is redundant now, so it's replaced by a dash.

Please compare the following images. Look as the second one is more compacted.

**Before:**

before

**After:**

Screenshot 2023-11-21 at 17-32-17 Open Build Service

**TODO:**
- [x] Align the comment icon and the first avatar in the comment thread properly
- [x] Collapse/hide non-conversation comments (from HistoryElement#comment).

**NOTE:** I tried to add one commit per change, so it'll be easy to remove/modify any of them independently

### DIFF
--- a/src/api/app/components/bs_request_history_element_component.rb
+++ b/src/api/app/components/bs_request_history_element_component.rb
@@ -27,15 +27,6 @@ class BsRequestHistoryElementComponent < ApplicationComponent
     end
   end
 
-  # While all history elements possibly have a comment, not all of them are from an actual human...
-  def element_with_comment_from_human?
-    ['RequestReviewAdded', 'ReviewAccepted', 'ReviewDeclined', 'RequestAccepted', 'RequestDeclined'].include?(@element.type.demodulize)
-  end
-
-  def css_for_comment
-    element_with_comment_from_human? ? 'border-start border-3 ps-3 ms-1' : ''
-  end
-
   def pending_reviews?
     request_reviews_for_non_staging_projects.any?(&:new?)
   end
--- a/src/api/app/components/bs_request_history_element_component.rb
+++ b/src/api/app/components/bs_request_history_element_component.rb
@@ -33,7 +33,7 @@ class BsRequestHistoryElementComponent < ApplicationComponent
   end
 
   def css_for_comment
-    element_with_comment_from_human? ? 'border-start border-3 ps-3 ms-4' : ''
+    element_with_comment_from_human? ? 'border-start border-3 ps-3 ms-1' : ''
   end
 
   def pending_reviews?
--- a/src/api/app/components/bs_request_history_element_component.rb
+++ b/src/api/app/components/bs_request_history_element_component.rb
@@ -33,7 +33,7 @@ class BsRequestHistoryElementComponent < ApplicationComponent
   end
 
   def css_for_comment
-    element_with_comment_from_human? ? 'comment-bubble comment-bubble-content' : ''
+    element_with_comment_from_human? ? 'border-start border-3 ps-3 ms-4' : ''
   end
 
   def pending_reviews?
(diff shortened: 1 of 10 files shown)


hennevogel:
Maybe we can even collapse review.reason by default?

hennevogel:
Would turn this

review reason uncollapsed

into this

hennevogel:
Or even collapsed by default and smaller

reviews collapsed-and-smaller

saraycp:
> Maybe we can even collapse review.reason by default?

I'll do that as well. I'll hide those types of messages (which come from `HistoryElement#comment`) and will display an arrow to uncollapse them.

Do you like the idea of displaying them as a quotation when they are uncollapsed? Something like this:

uncollapsed_history_element_comment

ncounter:
> > Maybe we can even collapse review.reason by default?
> 
> I'll do that as well. I'll hide those types of messages (which come from `HistoryElement#comment`) and will display an arrow to uncollapse them.
> 
> Do you like the idea of displaying them as a quotation when they are uncollapsed? Something like this:
> 
> uncollapsed_history_element_comment

It could even be just a tooltip popup visible on-hover of an icon representing the `review.reason` message. In the end nobody is going to reply to that message and it's just an additional text to the history element description text. But maybe it is too much of a change for the scope of this PR?

saraycp:
With the last commits, I stop displaying all the `HistoryElement#comment` but the user can hover the file icon next to the date to read. I also reduce the font size of history elements that are not real conversarion.

Screenshot 2023-11-17 at 17-14-37 Open Build Service

At first glance, only the real conversation is displayed.

@hennevogel and @ncounter please let me know if you like it.

src/api/app/components/bs_request_history_element_component.rb
```diff
@@ -27,15 +27,6 @@ def icon
     end
   end
 
-  # While all history elements possibly have a comment, not all of them are from an actual human...
-  def element_with_comment_from_human?
-    ['RequestReviewAdded', 'ReviewAccepted', 'ReviewDeclined', 'RequestAccepted', 'RequestDeclined'].include?(@element.type.demodulize)
```
ncounter, line 32: While for the most of the types it is fine to hide the message behind an icon and have it displayed only on hover with the tooltip, maybe it would still be important to have it always visible in case it is a `RequestAccepted` and `RequestDeclined`. For all  the other cases I'm all for it to hide it, but for the last two I'd still keep them. What do you think?

saraycp: I agree, they are the only ones that might be relevant. Especially the Decline, which might contain some relevant reasoning.

I'll try something else for them

saraycp: We can consider these three as "final states":

- RequestAccepted
- RequestRevoked
- RequestDeclined

But the reason for a re-opening might be relevant as well... :thinking: 

Too many exceptions maybe? I better go for all collapsed and the hover reveals the reason/comment.

I also think an info icon has a more clear meaning than a file. Please compare:

Screenshot 2023-11-21 at 14-22-43 Open Build Service

saraycp: Finally, we are going with collapse.

Screenshot 2023-11-21 at 17-31-17 Open Build Service

saraycp:
I've just rebased on master

danidoni (approved):

ncounter (approved):
Nice job :+1:

PR #15223, opened 2023-11-17
Refresh button for request chart build results

ncounter:
- [x] Add a button in order to async refresh the request chart build results without reloading the whole page.
- [x] Autorefresh the request chart build results

**Behavior**:
- if there is no build results, the chart box is completely hidden, and the Refresh button too
- since the beginning of the page load the autorefresh returns the build results every 1 minute
- until there is no data, nothing is shown
- once there is some data, the chart **and** the Refresh button are both visible
- once the Refresh button is clicked, an on-demand refresh is triggered, meanwhile the autorefresh still continues to reload every 1 minute.
 
## After
image

### DIFF
--- a/src/api/app/components/chart_component.html.haml
+++ b/src/api/app/components/chart_component.html.haml
@@ -5,7 +5,7 @@
       %i.fas.fa-sync-alt{ id: 'chart-build-reload' }
     %h4
       Build results summary
-    - if distinct_repositories.size > 5 || raw_data.size > 12
+    - if distinct_repositories.size > ChartComponent::MINIMUM_DISTINCT_BUILD_REPOSITORIES || raw_data.size > ChartComponent::MINIMUM_BUILD_RESULTS
       = column_chart( chart_data,
                       width: '100%',
                       height: '300px',
--- a/src/api/app/components/chart_component.html.haml
+++ b/src/api/app/components/chart_component.html.haml
@@ -1,5 +1,8 @@
 - if raw_data.size.positive?
   .m-4
+    .btn.btn-outline-primary.float-end{ onclick: 'updateChartBuildResults()', accesskey: 'r', title: 'Refresh Chart Build Results' }
+      Refresh
+      %i.fas.fa-sync-alt{ id: 'chart-build-reload' }
     %h4
       Build results summary
     - if distinct_repositories.size > 5 || raw_data.size > 12
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -10,7 +10,7 @@ class Webui::RequestController < Webui::WebuiController
                        :changes, :mentioned_issues, :chart_build_results]
   before_action :set_actions, only: [:inline_comment, :show, :build_results, :rpm_lint, :changes, :mentioned_issues, :chart_build_results],
                               if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
-  before_action :set_build_results_data, only: [:show], if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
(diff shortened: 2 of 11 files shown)


src/api/app/components/chart_component.html.haml
```diff
@@ -1,10 +1,12 @@
...
 - if raw_data.size.positive?
   .m-4
+    .btn.btn-outline-primary.float-end{ onclick: 'updateChartBuildResults()', accesskey: 'r', title: 'Refresh Chart Build Results' }
+      Refresh
+      %i.fas.fa-sync-alt{ id: 'chart-build-reload' }
     %h4
       Build results summary
-    - if distinct_repositories(raw_data).size > 5 || raw_data.size > 12
-      = column_chart( chart_data(raw_data),
+    - if distinct_repositories.size > 5 || raw_data.size > 12
```
danidoni, line 8: Can we transform those numbers into constants with a meaningful name?

danidoni: Magic numbers are a code smell https://stackoverflow.com/a/47902

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -473,6 +479,10 @@ def set_actions
     @actions = @bs_request.bs_request_actions
   end
 
+  def set_build_results_data
+    @build_results_data = ActionBuildResultsService::ChartDataExtractor.new(actions: @actions).call
```
danidoni, line 483: You are using this function as a "getter" and a "setter" (to use JAVA parlance) at the same time. You should set the instance variable or use the return value of the function to set the locals above, but not both, please.

danidoni: I also thing `@build_results_data` is not used anywhere...

src/api/app/components/chart_component.html.haml
ncounter: done :+1:

src/api/app/controllers/webui/request_controller.rb
ncounter: Good catch. It's not used anymore now. I'm refactoring it :+1:

ncounter: done

src/api/app/components/chart_component.html.haml
```diff
@@ -5,7 +5,7 @@
       %i.fas.fa-sync-alt{ id: 'chart-build-reload' }
     %h4
       Build results summary
-    - if distinct_repositories.size > 5 || raw_data.size > 12
+    - if distinct_repositories.size > ChartComponent::MINIMUM_DISTINCT_BUILD_REPOSITORIES || raw_data.size > ChartComponent::MINIMUM_BUILD_RESULTS
```
saraycp, line 8: This change corresponds to changes coming from #, the should be added there and rebase this branch
https://github.com/openSUSE/open-build-service/pull/15221/files#diff-bd413799cfd90c78b4726ba5d29235fe39542e36ad8b79165986e8523c0664d0R5

ncounter: Well, yes and no. This line was changed in PR#15221, but the logic was already present even before. In the meanwhile that PR got merged :confused: .
I understand this change is out-of-scope of the current PR. I can extract this commit and open a separate PR if you prefer.

danidoni: LGTM

danidoni (approved):

src/api/app/components/chart_component.html.haml
saraycp: No! Now that the other is merged, we can leave the change here. No worries

PR #15235, opened 2023-11-20
Load the pry-byebug gem by default in dev and test envs

danidoni:
We disabled this gem before[^1], but doing so forces us to explicitly require the gem just before using `binding.pry` like this:

```
require 'pry-byebug'
binding.pry
```

The gains we get by not explicitly loading it by default under dev and test envs are not worth it.

This PR loads the gem again under `development` and `test` environments.

[^1]: https://github.com/openSUSE/open-build-service/pull/15140


src/api/Gemfile
```diff
@@ -184,7 +184,7 @@ group :development, :test do
   # as alternative to the standard IRB shell
   gem 'pry', require: false
   # add step-by-step debugging and stack navigation capabilities to pry
-  gem 'pry-byebug', require: false
+  gem 'pry-byebug', require: true
```
hennevogel, line 187: This is the default no?

danidoni: Oh, yes, I can omit it

hennevogel (approved):

PR #15238, opened 2023-11-20
Replace Token::Rss with User.rss_secret 

hennevogel:
We jump through a lot of hoops in dealing with Token because there
is one Token that does not behave like every other Token: Token:Rss

And all Token:Rss does is provide a ActiveRecord::SecureToken. We can
use an attribute in User for this instead.

Depends on #15237


src/api/app/controllers/webui/users_controller.rb
```diff
@@ -140,6 +140,20 @@ def change_password
     end
   end
 
+  def rss_secret
+    user = User.session!
+
+    verb_prefix = user.rss_secret.present? ? 're-' : ''
```
dmarcoux, line 146:
```suggestion
    verb_prefix = user.rss_secret.present? ? 're' : ''
```

src/api/app/controllers/webui/users_controller.rb
```diff
@@ -140,6 +140,20 @@ def change_password
...
+
+    respond_to do |format|
+      if user.regenerate_rss_secret
```
danidoni, line 149: `User.regenerate_rss_secret` will raise an exception, instead of returning `false`...

hennevogel: removed

src/api/app/models/user.rb
```diff
@@ -121,8 +119,7 @@ class User < ApplicationRecord
   validates :password, length: { minimum: 6, maximum: ActiveModel::SecurePassword::MAX_PASSWORD_LENGTH_ALLOWED }, allow_nil: true
   validates :password, confirmation: true, allow_blank: true
   validates :biography, length: { maximum: MAX_BIOGRAPHY_LENGTH_ALLOWED }
-  # TODO: Remove if once migration has been executed. This is to prevent issues whenever a user is validated, but the rss_secret column isn't there yet
-  validates :rss_secret, uniqueness: true, length: { maximum: 200 }, allow_blank: true, if: -> { has_attribute?(:rss_secret) }
+  validates :rss_secret, uniqueness: true, length: { maximum: 200 }, allow_blank: true
```
dmarcoux, line 122: IBS was deployed yesterday, so this change can safely be merged and deployed.

danidoni (approved):

src/api/app/models/user.rb
```diff
@@ -877,6 +874,10 @@ def watched_projects
     Project.where(id: watched_items.where(watchable_type: 'Project').pluck(:watchable_id)).order('LOWER(name), name')
   end
 
+  def regenerate_rss_secret
+    update!(rss_secret: SecureRandom.base58(24))
```
dmarcoux, line 878: We should explain why we didn't go with [has_secure_token](https://api.rubyonrails.org/classes/ActiveRecord/SecureToken/ClassMethods.html#method-i-has_secure_token). Otherwise, we definitely don't know in X months/years why this isn't the case.

hennevogel: pushed a comment

PR #15246, opened 2023-11-21
Revert "Change User.rss_secret to default to ''"

hennevogel:
Doesn't work either. Have to think about it more...


danidoni (approved):

hennevogel:
sorry for the noise...

PR #15247, opened 2023-11-21
Add List-Id for better filtering with GMail

dirkmueller:
GMail is not able to filter emails by arbitrary headers, but only by a few selected ones. the easiest one to use is "List-ID", which is also set by e.g. GitHub for notifications in the notifications.

Setting such a header allowes in combination with optionally parts of the subject very powerful and reliable filtering of notifications when a user would be using GMail or Google Workspace.


src/api/app/mailers/event_mailer.rb
```diff
@@ -58,6 +58,7 @@ def set_event_headers
     return unless @event
 
     headers['X-OBS-event-type'] = @event.template_name
+    headers['List-ID'] = "OBS Notification System <#{@event.template_name}.#{URI.parse(Configuration.obs_url).host.downcase}>}"
```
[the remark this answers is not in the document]
dirkmueller: we can always prepend I think, we just can't change the suffix anymore. I wasn't able to find a "repository" equivalent like GitHub does, because OBS requests (which are the bulk of my notifications) can have many source projects and potentially only one target project, but it's optional. 

do you want me to add a target project, if available? I think it would only be another prefix, so it could also be done later. wdyt?

i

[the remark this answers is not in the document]
hennevogel: Yeah in general this belongs into `Event::Base.custom_headers` (and the subclasses) I think. It should give you a good hint on what's there to "identify" events.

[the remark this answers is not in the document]
dirkmueller: I'm not convinced that it should be an event specific custom header, but I'll explore that and update the PR when I have the time.

src/api/test/fixtures/event_mailer/request_event
```diff
@@ -36,6 +37,6 @@ To DECLINE the request:
 To REVOKE the request:
    osc request revoke REQUESTID --message="retracted because ..., sorry / thx / see better version ..."
 
--- 
+--
```
hennevogel, line 40: https://en.wikipedia.org/wiki/Signature_block#Standard_delimiter

src/api/test/fixtures/event_mailer/set_bugowner_event
```diff
@@ -36,6 +37,6 @@ To DECLINE the request:
 To REVOKE the request:
    osc request revoke REQUESTID --message="retracted because ..., sorry / thx / see better version ..."
 
--- 
+--
```
hennevogel, line 40: https://en.wikipedia.org/wiki/Signature_block#Standard_delimiter

src/api/test/fixtures/event_mailer/tom_gets_mail_too
```diff
@@ -38,6 +39,6 @@ To DECLINE the request:
 To REVOKE the request:
    osc request revoke REQUESTID --message="retracted because ..., sorry / thx / see better version ..."
 
--- 
+--
```
hennevogel, line 42: https://en.wikipedia.org/wiki/Signature_block#Standard_delimiter

hennevogel (changes requested):

src/api/app/mailers/event_mailer.rb
```diff
@@ -58,6 +58,7 @@ def set_event_headers
     return unless @event
 
     headers['X-OBS-event-type'] = @event.template_name
+    headers['List-ID'] = "OBS Notification System <#{@event.template_name}.#{URI.parse(@configuration.obs_url).host.downcase}>"
```
Conan-Kudo, line 61: Instead of `List-ID`, you can use `X-Loop` instead. That's [what the Red Hat Bugzilla does](https://bugzilla.redhat.com/show_bug.cgi?id=1689640).

Conan-Kudo (changes requested):

src/api/app/mailers/event_mailer.rb
dirkmueller: What's the advantage over List-ID? for List-ID we can make more specific dynamic lists, that would be the same for X-Loop (depending on the exact RFC definition of it) or more coarse (like simply the instance hostname).

hennevogel: You will never implement a functioning filter system in a mail header. If someone wants a functioning mail filter, get off GMail...

dirkmueller: so X-BeenThere: $projectname? I'm not sure that will work (I haven't tried).

dirkmueller:
@hellcp-work I did get your point and I agree to it. However, requests have not a clear "organisation/project", as they can have a set of sources and a set of destinations and types. 

for everything else (build failure notifications etc) the project/package structure makes a ton of sense to be reflected on the List-Id and I'll work on that shortly after catching up with the backlog.

PR #15257, opened 2023-11-22
Add update information to the conversations tab

hellcp-work:
This PR adds update information to the description box in the conversations tab, making it easier to read the patchinfo update details

image

You can check it out under https://obs-reviewlab.opensuse.org/hellcp-work-patchinfo_summary/request/show/15


saraycp (commented):
I'm afraid that the Patchinfo information to be displayed in that "description" box might be quite long. That will move the most important part (Conversation & History) a lot down. As an alternative, I'd propose to move all the patchinfo information to a new tab. Is that an option?

Anyway, If we finally decide to leave it in the Description box. I'd propose to reduce the font size of all the headings in it. For example use `<h4>`  instead of `<h3>` for Description and Patches. Then `<h5>` for the patchinfo summary and badge; and so on. This is a matter of taste, but the patchinfo badges look almost as big as the Request state badge on top of the page and headings are all `<h3>`.

src/api/app/components/patchinfo_component.html.haml
```diff
@@ -0,0 +1,71 @@
...
+      = @patchinfo['packager']
+    - if @patchinfo.key?('incident')
+      in the maintenance incident request
+      - if (incident = BsRequest.find_by(number: @patchinfo['incident']))
+        = link_to(request_show_path(incident)) do
+          \##{@patchinfo['incident']}
+      - else
+        \##{@patchinfo['incident']}
+
+  %h5 Patch description
```
dmarcoux, line 23: If there is no patch description, I suggest not displaying this heading and the `No description specified.` text at all. There's already possibly a lot of information included in a patchinfo, let's not make it harder to process with such text.

src/api/app/components/patchinfo_component.html.haml
```diff
@@ -0,0 +1,71 @@
...
+  %h4
+    = link_to(@path) do
+      = @patchinfo['summary']
+    = category
+    = rating
+    = stopped
+    = retracted
+  %p
+    Patch created by
+    - if (packager = User.find_by_login(@patchinfo['packager']))
```
dmarcoux, line 11: This `if/else` statement could be simplified to `= render UserAvatarComponent.new(packager.login)` if we slightly adapt the view component to deal with inexistent users. Right now, it would crash here in this case since `avatar_object` is not referring to a user: https://github.com/openSUSE/open-build-service/blob/9190df59ecb80b20d9fba3efec29f7f4cf136a5f/src/api/app/components/user_avatar_component.rb#L15

If we don't adapt the view component, the code in this view has to change since we're fetching twice the user. Once in this view and once in the view component here: https://github.com/openSUSE/open-build-service/blob/9190df59ecb80b20d9fba3efec29f7f4cf136a5f/src/api/app/components/user_avatar_component.rb#L11

src/api/spec/features/webui/requests_spec.rb
```diff
@@ -436,4 +436,44 @@
...
+      visit request_show_path(maintenance_request)
+    end
+
+    it 'shows patch information' do
+      expect(page).to have_text('Patches')
+      expect(page).to have_css('#patchinfo-details')
+    end
+
+    it 'shows category badge' do
+      patchinfo_css = find_by_id('patchinfo-details')
```
dmarcoux, line 468: I believe this example could be all combine in one line like so:

`expect(page).to have_css('#patchinfo-details .badge.text-bg-info', text: 'Recommended')`

I didn't give it a try locally.

src/api/spec/features/webui/requests_spec.rb
```diff
@@ -436,4 +436,44 @@
...
+      expect(patchinfo_css).to have_text('Recommended')
+      expect(patchinfo_css).to have_css('.badge.text-bg-info')
+    end
+
+    it 'shows rating badge' do
+      patchinfo_css = find_by_id('patchinfo-details')
```
dmarcoux, line 474: Same comment as the other example above.

src/api/spec/features/webui/requests_spec.rb
```diff
@@ -436,4 +436,44 @@
...
+
+    before do
+      Flipper.enable(:request_show_redesign)
+      login submitter
+      create(:patchinfo, project_name: source_project.name, package_name: 'patchinfo')
+      visit request_show_path(maintenance_request)
+    end
+
+    it 'shows patch information' do
+      expect(page).to have_text('Patches')
```
dmarcoux, line 463: Did you try to combine those? I believe it should work, you can try this:

`expect(page).to have_css('#patchinfo-details', text: 'Patches')`

src/api/app/components/patchinfo_component.html.haml
dmarcoux: Displaying nothing, so an empty description, doesn't bring any value to the user. I still suggest hiding it in this case.

saraycp (approved):

saraycp:
The feedback from @dmarcoux was addressed so we are good to go and merge. The missing specs are going to be done later on and it's tracked in [this card](https://trello.com/c/YMqqeEbj/178-add-patchinfo-component-spec).

PR #15262, opened 2023-11-23
Decouple the two /trigger controllers

hennevogel:
We inherit from TriggerController in TriggerWorkflowController just for one `before_action`. We don't use anything else from TriggerController. This does not make much sense anymore.

### DIFF
--- a/src/api/app/controllers/trigger_workflow_controller.rb
+++ b/src/api/app/controllers/trigger_workflow_controller.rb
@@ -1,10 +1,19 @@
-class TriggerWorkflowController < TriggerController
+class TriggerWorkflowController < ApplicationController
   include ScmWebhookHeadersDataExtractor
   include ScmWebhookPayloadDataExtractor
+  include Trigger::Errors
 
-  # We don't need to validate that the body of the request is XML. We receive JSON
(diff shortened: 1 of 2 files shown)


danidoni:
https://trello.com/c/s3rXELVj/316-stop-using-inheritance-on-workflowtriggercontroller

src/api/app/controllers/trigger_workflow_controller.rb
```diff
@@ -35,6 +44,15 @@ def create
...
   private
 
+  def set_token
+    @token = ::TriggerControllerService::TokenExtractor.new(request).call
+    raise InvalidToken, 'No valid token found' unless @token
+  end
+
+  def pundit_user
+    @token.executor
+  end
```
danidoni, line 54: We can extract those two methods into a module and share it with `TriggerController` also

danidoni: To achieve DRYness. The next time someone needs to change those methods they will start to diverge. Having them in a single place will make people think about all the use cases.

hennevogel: If that "next time" comes we can optimize it. Or in other words: f**k the future :)

danidoni: All right, then, f**ck the future.

danidoni (approved):

PR #15266, opened 2023-11-24
[api] fix handling of architecture filter for release command

adrianschroeter:
it would filter on release before, but fail if not filtered architectures are missing in target

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -674,6 +674,7 @@ POST /source/<project>/<package>?cmd=release
 
 Parameters:
   comment: description for the history
+  arch: limit the release to the specified architecture
   repository: limit the release to the specified repository
   target_repository: specify the target repository name (together with target_project)
   target_project: overwrites the target definition from project meta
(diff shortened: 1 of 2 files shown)


docs/api/api/api.txt
```diff
@@ -671,6 +671,7 @@ POST /source/<project>/<package>?cmd=release
 
 Parameters:
   comment: description for the history
+  arch: limit the release to the specified architecture
```
dmarcoux, line 674: The API documentation are now under https://github.com/openSUSE/open-build-service/tree/master/src/api/public/apidocs, as this is what OBS points to when clicking on the [`API Documentation` on the homepage](https://api.opensuse.org/apidocs/index). Could you reflect this change there? I think it's in this [file](https://github.com/openSUSE/open-build-service/blob/master/src/api/public/apidocs/paths/source_project_name_package_name_cmd_release.yaml).

adrianschroeter: it does exist there already, it is just not working properly.

eduardoj: That's right. The `arch` parameter was documented in 8b3848bbc9bb16de9420260531b98c827beb6f57.

hennevogel (approved):

Issue #10085, opened 2020-08-27
Replace xmlhash with Nokogiri 


dmarcoux:
We currently have 2 gems for parsing XML. We either use `xmlhash` or `Nokogiri`. We should only use `Nokogiri` since, unlike `xmlhash`, it is widely adopted by the Ruby community. It's the de facto gem for parsing XML.

dmarcoux:
We sometimes need to convert a XML document to a hash. In order to do this, I see two possible solutions:

1. Set the backend of `ActiveSupport::XmlMini` to use Nokogiri and rely on the `to_hash` method on a Nokogiri document:
```ruby
# Setup in an initializer under `src/api/app/config/initializers`
ActiveSupport::XmlMini.backend = 'Nokogiri'

# Usage example
my_hash = Nokogiri::XML("<some><thing></thing></some>").to_hash
```

2. Install the [`libxml-ruby` gem](https://rubygems.org/gems/libxml-ruby),set the backend of `ActiveSupport::XmlMini` to use libxml-ruby and rely on `Hash.from_xml`:
```ruby
# Setup in an initializer under `src/api/app/config/initializers`
ActiveSupport::XmlMini.backend = 'LibXML'

# Usage example
my_hash = Hash.from_xml("<some><thing></thing></some>")
```

dmarcoux:
Here's a benchmark:

```ruby
# Put this file under `src/api/` and run the benchmarks with `ruby src/api/this_file.rb`
require 'benchmark'
require_relative './config/environment'

xml = <<XML
  <root>
    <under>divide</under>
    <factory>
      <whole>
        <struggle>2037830316</struggle>
        <badly>-828494737</badly>
        <leaf>speak</leaf>
        <across>-1282250434</across>
        <shinning>262024442</shinning>
        <ran>potatoes</ran>
      </whole>
      <mistake>2131212417</mistake>
      <remain>opportunity</remain>
      <milk>boat</milk>
      <lunch>moon</lunch>
      <species>-1502423854.7647963</species>
    </factory>
    <hunter>199823558</hunter>
    <palace>-1968452430</palace>
    <command>slept</command>
    <animal>1502711410.04618</animal>
  </root>
XML

Benchmark.bm do |benchmark|
  benchmark.report("Nokogiri") do
    ActiveSupport::XmlMini.backend = 'Nokogiri'

    10_000.times do
      Nokogiri::XML(xml).to_hash
    end
  end

  benchmark.report("LibXML") do
[...]

PR #15274, opened 2023-11-27
Introduce happymapper with an example usage

hellcp-work:
This introduces a nice way to parse xml documents into objects, using `nokogiri-happymapper`. This way, we have an easier way to consistently and reliably parse xml into a static structure, with a possibility to add custom functions (see `Backend::Xml::Patchinfo#packager_object`) which will make it easier for us to use the parsed results without having to do `Object.find_by` calls elsewhere.

I also added TrueClass type to happymapper, because the backend tends to use `<element/>` as `true` and ` ` as `false` (in our case it will be `nil` instead, but it's falsy so it's fine).

### DIFF
--- a/src/api/app/lib/backend/xml/patchinfo.rb
+++ b/src/api/app/lib/backend/xml/patchinfo.rb
@@ -43,7 +43,7 @@ module Backend::Xml
       # check the packager field
       User.find_by_login!(packager) if packager
       # valid issues?
-      issues.map { |i| i.object }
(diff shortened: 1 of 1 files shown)


dmarcoux:
We had discussions in the past regarding XML parsing and performance. This happened [here](https://github.com/openSUSE/open-build-service/issues/10085).

So how is the performance of this? I would be interested in knowing how this adds up to the execution time of Nokogiri.

PR #15275, opened 2023-11-27
Make sure that Mail::Address is available in the patchinfo component

hellcp-work:
This doesn't seem to be included by rails by default, so let's include it so that it's available


hennevogel (changes requested):
Please explain *why* this is needed. 

https://github.com/openSUSE/open-build-service/blob/master/CONTRIBUTING.md#proper-commit-messages

PR #15276, opened 2023-11-27
Fix issue when creating notification with group and RSS

danidoni:
Fixes: #15273

It's weird because there is no subscription for a group + RSS channel in the production database.
That kind of subscription is not even possible through the WebUI.
Screenshot 2023-11-27 at 17-48-41 Open Build Service

### DIFF
--- a/src/api/app/services/notification_service/notifier.rb
+++ b/src/api/app/services/notification_service/notifier.rb
@@ -67,7 +67,7 @@ module NotificationService
     end
 
     def create_notification?(subscriber, channel)
-      return false if subscriber.nil? || subscriber.away? || (channel == :rss && subscriber.rss_secret.blank?)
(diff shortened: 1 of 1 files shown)


src/api/spec/services/notification_service/notifier_spec.rb
```diff
@@ -135,6 +135,25 @@
...
 
+          context 'and a group is subscribed to the event via RSS' do
+            before do
+              event_subscription = create(
+                :event_subscription,
+                eventtype: 'Event::RelationshipCreate',
+                receiver_role: 'any_role',
+                user: nil,
+                group: group,
+                channel: :rss
```
saraycp, line 146: As this scenario shouldn't be possible (I don't know yet what happened in production). I prefer to remove the spec. It was helpful to fix the issue but makes little sense to leave it.

eduardoj (approved):

PR #15277, opened 2023-11-27
[api] fix project kind removal

adrianschroeter:
...again....

### DIFF
--- a/src/api/app/models/project/update_from_xml_command.rb
+++ b/src/api/app/models/project/update_from_xml_command.rb
@@ -28,7 +28,8 @@ class Project
       project.remoteurl = xmlhash.value('remoteurl')
       project.remoteproject = xmlhash.value('remoteproject')
       project.scmsync = xmlhash.value('scmsync')
-      project.kind = xmlhash.value('kind') if xmlhash.value('kind').present?
+      project.kind = xmlhash.value('kind').presence || 'standard'
+
       #--- update flag group ---#
       project.update_all_flags(xmlhash)
       if ::Configuration.default_access_disabled == true && new_record && xmlhash.elements('access').empty?
--- a/src/api/test/unit/project_test.rb
+++ b/src/api/test/unit/project_test.rb
@@ -372,6 +372,36 @@ class ProjectTest < ActiveSupport::TestCase
     project_b.destroy
   end
 
+  def test_setting_project_kind
+    User.session = users(:king)
+    prj = Project.create!(name: 'project_1')
+    prj.update_from_xml!(Xmlhash.parse(
+                           "<project name='project_1' kind='maintenance_release'>
+                             <title/>
+                             <description/>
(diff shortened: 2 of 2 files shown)


src/api/app/models/project/update_from_xml_command.rb
```diff
@@ -28,7 +28,8 @@ def run(xmlhash, force = nil)
       project.remoteurl = xmlhash.value('remoteurl')
       project.remoteproject = xmlhash.value('remoteproject')
       project.scmsync = xmlhash.value('scmsync')
-      project.kind = xmlhash.value('kind') if xmlhash.value('kind').present?
+      project.kind = xmlhash.value('kind').presence || 'standard'
```
eduardoj, line 31: I've searched backwards in the commit history about this change. The title of the pull request, the commit title and the description of the commit gave me a hint about fixing something, or reverting some change. I found these changes:

https://github.com/openSUSE/open-build-service/commit/c03f3e0ebbc2e97714ba684a71b8dd14a00e9578#diff-40f55459121e262fa08c43869c11eb281f30a3b5e0624fa8aee3f3e970e103f5L441-R451

Shouldn't we describe better why `'standard'` should be the default value for the `kind` of a project?

src/api/test/unit/project_test.rb
```diff
@@ -372,6 +372,29 @@ def test_handle_project_links
...
+    xml = prj.to_axml
+    assert_xml_tag xml, tag: :project, attributes: { kind: 'maintenance_release' }
+
+    prj.update_from_xml!(Xmlhash.parse(
+                           "<project name='kind'>
+                             <title/>
+                             <description/>
+                           </project>"
+                         ))
+    assert_no_xml_tag xml, tag: :project, attributes: { kind: 'maintenance_release' }
```
eduardoj, line 395:
```suggestion
    prj.update_from_xml!(Xmlhash.parse(
                           "<project name='kind'>
                             <title/>
                             <description/>
                           </project>"
                         ))
    xml = prj.to_axml
    assert_no_xml_tag xml, tag: :project, attributes: { kind: 'maintenance_release' }
    assert_xml_tag xml, tag: :project, attributes: { kind: 'standard' }
```

If removing the kind of a project makes its new kind to be `'standard'`, shouldn't we check for standard too?

eduardoj: The minitests fail. I think there is an assignment missing in the test:

```
xml = prj.to_axml
```

hennevogel:
`standard` is the default value in the database since 2015 

https://github.com/openSUSE/open-build-service/commit/c03f3e0ebbc2e97714ba684a71b8dd14a00e9578

adrianschroeter:
and what has this to do with the issue?

this is about a different set kind and when the user removes the attribute from xml.

eduardoj (approved):

PR #15280, opened 2023-11-28
Refactor methods to run backend test tasks

eduardoj:
Reduce duplication offenses reported by codeclimate.

Also, remove a method which was never used.

### DIFF
--- a/src/api/app/lib/backend/test/tasks.rb
+++ b/src/api/app/lib/backend/test/tasks.rb
@@ -11,15 +11,6 @@ module Backend
         end
       end
 
-      def run_dispatcher
-        Rails.logger.debug 'run dispatcher'
-        perlopts = "-I#{Rails.root}/../backend -I#{Rails.root}/../backend/build"
-        IO.popen("cd #{backend_config}; exec perl #{perlopts} ./bs_dispatch --testmode") do |io|
-          # just for waiting until dispatcher finishes
-          io.each { |line| Rails.logger.debug { "dispatcher: #{line.strip.chomp}" } if line.present? }
-        end
-      end
-
       def run_publisher
         Rails.logger.debug 'run publisher'
         perlopts = "-I#{Rails.root}/../backend -I#{Rails.root}/../backend/build"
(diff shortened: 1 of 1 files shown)


src/api/app/lib/backend/test/tasks.rb
```diff
@@ -3,39 +3,15 @@ class Test
...
       def run_scheduler(arch)
-        Rails.logger.debug { "RUN_SCHEDULER #{arch}" }
-        perlopts = "-I#{Rails.root}/../backend -I#{Rails.root}/../backend/build"
-        IO.popen("cd #{backend_config}; exec perl #{perlopts} ./bs_sched --testmode #{arch}") do |io|
-          # just for waiting until scheduler finishes
-          io.each { |line| Rails.logger.debug { "scheduler(#{arch}): #{line.strip.chomp}" } if line.present? }
-        end
-      end
-
-      def run_dispatcher
```
dmarcoux, line 14: This method is now gone, is this what we want? I don't think so, could you confirm?

src/api/app/lib/backend/test/tasks.rb
```diff
@@ -49,6 +25,16 @@ def run_admin(args)
         end
         ret
       end
+
+      private
+
+      def run_backend_server(name:, command:)
```
adrianschroeter, line 31: this name is highly confusing, could you use something like "single_scheduler_evaluation_run" or something like that?

adrianschroeter: hm, or something like "execute_backend_daemon".

but "backend_server" are many processes for me, not a single temporary execution.

eduardoj: "single_scheduler_evaluation_run" could be more confusing. This method not only runs a scheduler. I'm open to more ideas.

eduardoj: I renamed the method to `execute_backend_daemon`.

adrianschroeter:
On Dienstag, 28. November 2023, 11:31:35 CET Eduardo J. wrote:
> @eduardoj commented on this pull request.
> 
> 
> 
> > @@ -49,6 +25,16 @@ def run_admin(args)
>          end
>          ret
>        end
> +
> +      private
> +
> +      def run_backend_server(name:, command:)
> 
> I renamed the method to `execute_backend_daemon`.

sounds good

-- 

Adrian Schroeter ***@***.***>

PR #15281, opened 2023-11-28
Patchinfo component fix

saraycp:
Handle patchinfo tags as arrays. Otherwise, the loop crashes when there is only one element (not an array).
Include component specs.

Follow up https://github.com/openSUSE/open-build-service/pull/15257

### DIFF
--- a/src/api/app/components/patchinfo_component.html.haml
+++ b/src/api/app/components/patchinfo_component.html.haml
@@ -36,16 +36,16 @@
     - if @patchinfo['releasetarget'].present?
       %h6 Targeted for release in the following projects
       %ul
-        - release_targets.each do |releasetarget|
+        - release_targets.each do |release_target|
           %li
-            - if (project = Project.find_by_name(releasetarget['project']))
+            - if (project = Project.find_by_name(release_target['project']))
               = link_to(project_show_path(project)) do
-                = releasetarget['project']
+                = release_target['project']
             - else
-              = releasetarget['project']
-            - if releasetarget.key?('repository')
+              = release_target['project']
+            - if release_target.key?('repository')
               Repository:
-              = releasetarget['repository']
+              = release_target['repository']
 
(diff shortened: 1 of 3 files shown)


src/api/app/components/patchinfo_component.html.haml
```diff
@@ -36,7 +36,7 @@
     - if @patchinfo['releasetarget'].present?
       %h6 Targeted for release in the following projects
       %ul
-        - @patchinfo['releasetarget'].each do |releasetarget|
+        - [@patchinfo['releasetarget']].flatten.each do |releasetarget|
```
danidoni, line 39: I would move all those `[@something].flatten` into the component class, expose it with a method or an instance variable

saraycp: @danidoni  please have another look

src/api/spec/components/patchinfo_component_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'rails_helper'
+
+RSpec.describe PatchinfoComponent, :vcr, type: :component do
```
eduardoj, line 3:
```suggestion
RSpec.describe PatchinfoComponent, type: :component do
```
These specs seem not to need VCR.

saraycp: True! Removed!

src/api/app/components/patchinfo_component.rb
```diff
@@ -20,6 +20,22 @@ def initialize(patchinfo, path)
...
+    [@patchinfo['binary']].flatten
+  end
+
+  def packages
+    [@patchinfo['package']].flatten
+  end
+
+  def issues
+    [@patchinfo['issue']].flatten
+  end
```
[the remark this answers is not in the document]
saraycp: Please have another look

eduardoj (approved):

PR #15293, opened 2023-11-30
Add canned responses table to the database

hellcp-work:
This adds a table for canned responses, that will be used for storing canned responses


dmarcoux (approved):

src/api/db/migrate/20231130094135_create_canned_responses.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCannedResponses < ActiveRecord::Migration[7.0]
+  def change
+    create_table :canned_responses, id: :integer do |t|
```
eduardoj, line 3:
```suggestion
    create_table :canned_responses, id: :bigint do |t|
```
Any reason not to use the now default `bigint`?

I haven't tested it, but `create_table :canned_responses do |t|` should also work.

PR #15295, opened 2023-11-30
Remove not needed `try` for no longer existing RSS group subscribers

eduardoj:
This is a follow up to #15290.

Now that subscription of groups to RSSs are prevented, we can now safely rely on the fact that when we deal with RSS channels, the receiver will always be a user.

### DIFF
--- a/src/api/app/services/notification_service/notifier.rb
+++ b/src/api/app/services/notification_service/notifier.rb
@@ -67,7 +67,7 @@ module NotificationService
     end
 
     def create_notification?(subscriber, channel)
-      return false if subscriber.nil? || subscriber.away? || (channel == :rss && subscriber.try(:rss_secret).blank?)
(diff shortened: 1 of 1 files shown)


bmwiedemann:
Does it hurt to keep the `try` there? I wonder what would happen if groups were subscribed to RSS already.

eduardoj:
> Does it hurt to keep the `try` there? I wonder what would happen if groups were subscribed to RSS already.

Groups couldn't have been subscribed to RSS and cannot be subscribed to RSS, ever.

This PR is a follow up to #15290. From that PR:

_At some point, the code returned the default subscription for RSS as a valid subscription for that receiver. That's wrong because RSS + group is not a valid combination._

The `try` was only introduced as a monkey patch in #15276, when the reason of the issue #15273 was still not well known.

Now that the issue is correctly fixed in #15290, we no longer need to use `try`.

ncounter (approved):

PR #15296, opened 2023-11-30
Implement UI for canned responses feature

hellcp-work:
Depends on #15293 

To test:
1. Go to your profile
2. Click on canned responses in the sidebar
3. Create a new canned response
4. Open anything with an ability to comment on
5. See the comment box have an ability to use canned responses

### DIFF
--- a/src/api/app/assets/javascripts/webui/canned_responses.js
+++ b/src/api/app/assets/javascripts/webui/canned_responses.js
@@ -1,5 +1,7 @@
 function setupCannedResponses() { // jshint ignore:line
   $('[data-canned-controller]').on('click', '[data-canned-response]', function(e) {
     $(e.target).closest('[data-canned-controller]').find('textarea').val(e.target.dataset.cannedResponse);
+    // we have to enable the submit button for the comments form
+    $(e.target).closest('[class*="-comment-form"]').find('input[type="submit"]').prop('disabled', false);
   });
 }
--- a/src/api/app/assets/javascripts/webui/canned_responses.js
+++ b/src/api/app/assets/javascripts/webui/canned_responses.js
@@ -1,0 +1,5 @@
+function setupCannedResponses() { // jshint ignore:line
+  $('[data-canned-controller]').on('click', '[data-canned-response]', function(e) {
+    $(e.target).closest('[data-canned-controller]').find('textarea').val(e.target.dataset.cannedResponse);
+  });
+}
--- a/src/api/app/components/canned_responses_dropdown_component.html.haml
+++ b/src/api/app/components/canned_responses_dropdown_component.html.haml
@@ -1,0 +1,15 @@
+.dropdown
+  %button.btn.btn-link.text-decoration-none.dropdown-toggle{ type: 'button', 'data-bs-toggle': 'dropdown', 'aria-expanded': false }
+    %i.fas.fa-comment-dots
+    Canned Responses
+  %ul.dropdown-menu.dropdown-menu-end
+    - @canned_responses.each do |canned_response|
+      %li.dropdown-item{ data: { 'canned-response': canned_response.content }, role: 'button' }= canned_response.title
+    %li
+      %hr.dropdown-divider/
+    %li
+      = link_to('Create and modify' , canned_responses_path, class: 'dropdown-item')
+:javascript
+  $(document).ready(function(){
+    setupCannedResponses();
+  });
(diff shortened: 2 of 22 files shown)


src/api/app/models/user.rb
```diff
@@ -64,6 +64,7 @@ class User < ApplicationRecord
 
   has_many :moderated_comments, class_name: 'Comment', foreign_key: 'moderator_id'
   has_many :decisions, foreign_key: 'moderator_id'
+  has_many :canned_responses
```
dmarcoux, line 67: We should destroy all canned responses of a user if the user gets deleted, so:

```suggestion
  has_many :canned_responses, dependent: :destroy
```

dmarcoux: The default strategy is to do nothing. This is what the documentation says in the link you shared:

> For has_many, (...) if no :dependent option is given, then it will follow the default strategy. The default strategy is to do nothing (leave the foreign keys with the parent ids set)

src/api/app/assets/javascripts/webui/canned_responses.js
```diff
@@ -0,0 +1,5 @@
+$(document).ready(function(){
```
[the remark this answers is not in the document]
dmarcoux: I updated my comment to show an example of what we do right now.

danidoni:
Looks like applying a canned response here[^1] does not trigger the enabling of the button.

[^1]: https://obs-reviewlab.opensuse.org/hellcp-work-canned-responses-ui/project/show/openSUSE:Factory

src/api/app/policies/canned_response_policy.rb
krauselukas, line 1: I guess we should also check for the beta flag and that the user is a moderator. Right now this shown to every logged in user.

krauselukas:
Took me a bit time to realize that those are the saved canned responses, IMO they are not well visible and through the green color can be confused with the flash message right after the creation.

src/api/app/policies/canned_response_policy.rb
krauselukas: I added the beta flag to the links and policies

krauselukas:
>  Took me a bit time to realize that those are the saved canned responses, IMO they are not well visible and through the green color can be confused with the flash message right after the creation.

I worked a bit on the UI, this is the result. To me this makes it a bit more visible and clear.
Screenshot 2023-12-06 at 17-07-59 Open Build Service

krauselukas:
> Looks like applying a canned response [here](https://obs-reviewlab.opensuse.org/hellcp-work-canned-responses-ui/project/show/openSUSE:Factory) keeps the button disabled. It's not until you type something in the box, that the button is enabled.

Fixed, the javascript code for enabling the submit button on the comments form only listens for key events, I had to add an extra line to enable the button after inserting the value through the canned responses

src/api/app/models/canned_response.rb
```diff
@@ -0,0 +1,50 @@
+# TODO: Please overwrite this comment with something explaining the model target
```
dmarcoux, line 1: This should be removed if we don't have anything useful to write here.

src/api/app/views/webui/users/canned_responses/index.html.haml
```diff
@@ -0,0 +1,39 @@
...
+                = canned_response.title
+            .accordion-collapse.collapse{ id: "#{canned_response.title.parameterize}-collapse" }
+              .accordion-body
+                .d-flex.justify-content-between
+                  = canned_response.content
+                  .d-flex
+                    = link_to(edit_canned_response_path(canned_response), class: 'btn py-0 px-1 btn-link') do
+                      %i.fas.fa-edit
+                      Edit
+                    = button_to(canned_response_path(canned_response), method: :delete, class: 'btn py-0 px-1 text-danger') do
```
dmarcoux, line 35: We should ask the user to confirm before deleting a canned response. Otherwise, we're one click away from deleting by mistake a canned response.

src/api/app/components/canned_responses_dropdown_component.html.haml
```diff
@@ -0,0 +1,15 @@
+.dropdown
+  %button.btn.btn-link.text-decoration-none.dropdown-toggle{ type: 'button', 'data-bs-toggle': 'dropdown', 'aria-expanded': false }
+    %i.fas.fa-comment-dots
+    Canned Responses
+  %ul.dropdown-menu.dropdown-menu-end
+    - @canned_responses.each do |canned_response|
+      %li.dropdown-item{ data: { 'canned-response': canned_response.content } }= canned_response.title
```
dmarcoux, line 7: It would be nice to have on these elements the usual cursor for clickable elements with `cursor: pointer`.

dmarcoux:
> I guess the feature test need some adaptation after the recent changes

The policies were changed to require the `content_moderation` feature to be enabled and the specs didn't take this change in consideration.

PR #15299, opened 2023-11-30
Instantly display the 'You reported the X' message

dmarcoux:
A page refresh isn't needed anymore for the message to appear.

For comments, the implementation isn't really robust, but considering how the comments are rendered, this is what we can do without refactoring a lot of code. In the best scenario, rendering a comment would be as simple as rendering a view component. It would then be unneccessary to fiddle much with JavaScript, re-rendering the comment view component would be the solution.

As for projects and packages, it's not possible to re-render the side_links partials since they depend upon so many objects and we don't have those in webui/reports_controller.


dmarcoux:
This is functional, but I am not done yet.

dmarcoux:
[Display text right after reporting (M)](https://trello.com/c/9wgmgCF7/2378-display-text-right-after-reporting-m)

src/api/spec/features/beta/webui/reports_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    let(:project) { create(:project, name: 'some_random_project') }
+    let!(:comment) { create(:comment, commentable: project) }
+
+    before do
+      login user
+      visit project_show_path(project)
+    end
+
+    it 'displays the "You reported this comment." message instantly' do
+      expect(page).not_to have_text('You reported this comment.')
```
eduardoj, line 19:
```suggestion
```

I would not test for a page not having the `'You reported this comment'` text just after loading a project page. It looks obvious to me, that the project page won't have this text every time a project page is loaded.

The same applies in the rest of the `it` clauses below, but for packages, comments and users.

saraycp:
There is one case missing: reporting a comment on a Request (redesigned page) requires a refresh to see the yellow text about the report inside the bubble.

Compare these two comment bubbles. This is what I did:
- I reported the first comment and then refreshed the page => yellow text is there
- I reported the second comment but didn't refresh afterwards => no yellow text.

Screenshot 2023-12-01 at 15-02-24 Open Build Service

src/api/spec/features/beta/webui/reports_spec.rb
dmarcoux: Without checking for the absence of `'You reported this comment'` message when the page loads, how would we prove that the action of reporting a comment really triggered the message to appear?

dmarcoux:
> There is one case missing: reporting a comment on a Request (redesigned page) requires a refresh to see the yellow text about the report inside the bubble.

Oh right, I forgot that. The comments are different there... I'll have a look.

src/api/spec/features/beta/webui/reports_spec.rb
eduardoj: > Without checking for the absence of 'You reported this comment' message when the page loads, how would we prove that the action of reporting a comment really triggered the message to appear?

Showing the `'You reported this comment'` message should not happen when the page loads, ever. We shouldn't check for things that doesn't happen in a controlled environment.

> It could have been there already before if you already reported a comment on a project/package. Sure, in the specs, it doesn't because there aren't other comments.

Yes, we could write tests switching from page with notifications to other pages. But this is not the case. And I don't find it necessary in this pull request.

These tests first load a page in an environment where no other report has been submitted before. That is what I meant with a controlled environment. That should be enough for not including the `expect(page).not_to have_text('You reported this comment.')` check.

dmarcoux: It's true that since we fully set the page how we want it to be, it's redundant to check for the absence of the message. I'll remove those lines.

src/api/app/assets/javascripts/webui/report.js
```diff
@@ -40,6 +40,28 @@ function hideReportButton(element) {
   $(element).addClass('d-none');
 }
 
+/* exported showYouReportedMessage */
+function showYouReportedMessage(reportLinkId, reportableType, reportableId, message) {
+  switch(reportableType) {
+    case 'Comment':
+      // Comments differ depending on where they are, so this is why we have two ways. If an element isn't found, nothing will happen...
```
dmarcoux, line 47: I won't say this is great. As I wrote in the PR description, given how the comments are rendered, I decided to do it this way.

saraycp:
@dmarcoux could you please get rid of the white background behind the flag icon? (in the yellow text).

It seems it's somehow inherited but not needed in the places where the text is displayed.

saraycp (approved):
I leave my approval. Only missing the change of the background color.

eduardoj (approved):

PR #15303, opened 2023-12-01
Include more information in notifications for deleted reportables

eduardoj:
Include more information in notifications for deleted projects and packages.

## Before

Screenshot from 2023-12-01 13-51-09

## After

Screenshot from 2023-12-06 13-55-18

### DIFF
--- a/src/api/app/models/event/report_for_package.rb
+++ b/src/api/app/models/event/report_for_package.rb
@@ -8,8 +8,7 @@ module Event
 
       Rails.application.routes.url_helpers.package_show_path(package: notification.event_payload['package_name'],
                                                              project: notification.event_payload['project_name'],
-                                                             notification_id: notification.id,
-                                                             anchor: 'comments-list')
+                                                             notification_id: notification.id)
     end
 
     def self.notification_link_text(payload)
--- a/src/api/app/models/event/report_for_package.rb
+++ b/src/api/app/models/event/report_for_package.rb
@@ -3,6 +3,15 @@ module Event
     self.description = 'Report for a package has been created'
     payload_keys :package_name, :project_name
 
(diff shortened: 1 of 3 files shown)


krauselukas (commented):
@eduardoj could we disable the link or make it just text when the reportable got deleted?

hennevogel:
What about comments?

eduardoj:
> @eduardoj could we disable the link or make it just text when the reportable got deleted?

@krauselukas, I added a commit to show a text instead of a link. Please take a look again.

eduardoj:
> What about comments?

This PR doesn't deal with comments. Only with projects and packages.

src/api/app/models/event/report_for_package.rb
```diff
@@ -3,6 +3,15 @@ class ReportForPackage < Report
     self.description = 'Report for a package has been created'
     payload_keys :package_name, :project_name
 
+    def self.notification_link_path(notification)
+      return unless Package.exists_by_project_and_name(notification.event_payload['project_name'], notification.event_payload['package_name'])
+
+      Rails.application.routes.url_helpers.package_show_path(package: notification.event_payload['package_name'],
+                                                             project: notification.event_payload['project_name'],
+                                                             notification_id: notification.id,
+                                                             anchor: 'comments-list')
```
krauselukas, line 12: @eduardoj just saw the `anchor: 'comments-list'` here, this was already there, but seems wrong. Dont makes sense to jump to the comment section for a report on a package. Maybe you can remove it

eduardoj: @krauselukas, sure thing. I added another commit to fix that. Thanks!

krauselukas (approved):

PR #15308, opened 2023-12-04
Email notifications for events related to decisions

saraycp:
Depends on #15328.

### DIFF
--- a/src/api/app/models/event/cleared_decision.rb
+++ b/src/api/app/models/event/cleared_decision.rb
@@ -5,6 +5,11 @@ module Event
 
     payload_keys :id, :reason, :moderator_id, :reportable_type
 
+    def subject
+      decision = Decision.find(payload['id'])
+      "Cleared #{decision.reports.first.reportable&.class&.name} Report".squish
+    end
+
     def parameters_for_notification
       super.merge(notifiable_type: 'Decision')
     end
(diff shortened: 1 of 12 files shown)


krauselukas:
Maybe squash the commit about "Fix interpolation..." with the previous one.

src/api/app/models/event/cleared_decision.rb
```diff
@@ -3,7 +3,12 @@ class ClearedDecision < Base
     receiver_roles :reporter
     self.description = 'Reported content has been cleared'
 
-    payload_keys :id, :reason, :moderator_id, :reportable_type
+    payload_keys :id, :reason, :moderator_id, :report_last_id, :reportable_type
+
+    def subject
+      decision = Decision.find(payload['id'])
+      "Cleared #{decision.reports.first.reportable&.class&.name} Report".squish
```
krauselukas, line 10: @eduardoj just wondering why the `squish` is required?

eduardoj:
> Maybe squash the commit about "Fix interpolation..." with the previous one.

Sure. Done.

src/api/app/models/event/cleared_decision.rb
eduardoj: @krauselukas just for the case the interpolation of the variable returns `nil`: instead of returning `'Cleared  Report'` (with two spaces in between), only one space is returned: `'Cleared Report'`.

The same code can be found here: https://github.com/openSUSE/open-build-service/blob/91debd1dcbc8ea8be3c964d5ab2acd7ba549933b/src/api/app/components/notification_notifiable_link_component.rb#L54-L63

I would leave as it is.

krauselukas (approved):

PR #15314, opened 2023-12-05
Create the appeals table

danidoni:
This will support the implementation of the Appeal model.

### DIFF
--- a/src/api/db/migrate/20231204093136_create_appeals.rb
+++ b/src/api/db/migrate/20231204093136_create_appeals.rb
@@ -1,0 +1,11 @@
+class CreateAppeals < ActiveRecord::Migration[7.0]
+  def change
+    create_table :appeals, id: :bigint do |t|
+      t.text :reason, null: false
+      t.integer :appellant_id, null: false, foreign_key: { to_table: :users }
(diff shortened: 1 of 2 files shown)


src/api/db/migrate/20231204093136_create_appeals.rb
```diff
@@ -0,0 +1,11 @@
+class CreateAppeals < ActiveRecord::Migration[7.0]
+  def change
+    create_table :appeals, id: :bigint do |t|
+      t.text :reason, null: false
+      t.integer :appellant_id, null: false, foreign_key: { to_table: :users }
+      t.integer :decision_id, null: false, foreign_key: true
```
dmarcoux, line 6: This should be a `bigint` since this is what the decisions table's primary key is using, see https://github.com/openSUSE/open-build-service/blob/9b50289ba1a59378493f6209e9218b25c7f6c28c/src/api/app/models/decision.rb#L44-L46

dmarcoux:
I applied the suggestion I wrote in my previous comment. Please have a look. This is ready to be reviewed/merged.

krauselukas (approved):

eduardoj (approved):

PR #15318, opened 2023-12-05
Redirect to login page or root if anonymous

### LINKED ISSUE #15315: Redirection to login portal broken when authentication cookie expired
Since anomymous access is disabled on build service on IBS, 

Firefox is complaining every morning on pages already loaded in tabs ( https://build.suse.de/staging_workflows/SUSE:ALP:Source:Standard:1.0 or https://build.suse.de/project/show/SUSE:ALP:Source:Standard:Core:1.0:Build for instance) with the following message:

"can't redirect properly, problem might be disabled or refusing cookies"

and doesn't redirect properly to login page.

I need to open https://build.suse.de/ manually and login from there.

hennevogel:
Everything but root_path requires auth if Configuration.anonymous is true. No need to send people to :back if it will not work. Send them to the correct login page in proxy mode or directly to root_path instead.
    
Fixes #15315

### DIFF
--- a/src/api/app/controllers/webui/webui_controller.rb
+++ b/src/api/app/controllers/webui/webui_controller.rb
@@ -172,14 +172,20 @@ class Webui::WebuiController < ActionController::Base
 
   # before filter to only show the frontpage to anonymous users
   def check_anonymous
-    if User.session
-      false
-    else
-      unless ::Configuration.anonymous
-        flash[:error] = 'No anonymous access. Please log in!'
-        redirect_back(fallback_location: root_path)
-      end
-    end
+    return if User.session.present?
+    return if ::Configuration.anonymous
+
+    login_page = case CONFIG['proxy_auth_mode']
+                 when :mellon
+                   add_return_to_parameter_to_query(url: CONFIG['proxy_auth_login_page'], parameter_name: 'ReturnTo')
+                 when :ichain
+                   add_return_to_parameter_to_query(url: CONFIG['proxy_auth_login_page'], parameter_name: 'url')
+                 else
+                   root_path
+                 end
+
+    flash[:error] = 'No anonymous access. Please log in!' unless ::Configuration.proxy_auth_mode_enabled?
+    redirect_to login_page
   end
 
   # After filter to clean up caches
@@ -239,5 +245,16 @@ class Webui::WebuiController < ActionController::Base
       anonymous: !User.session,
       interface: :webui
     }
+  end
+
+  def add_return_to_parameter_to_query(url:, parameter_name:)
+    uri = URI(url)
+    return_to = {}
+    return_to[parameter_name] = request.fullpath
+    query_array = uri.query.to_s.split('&')
+    query_array << return_to.to_query # for URL encoding
+    uri.query = query_array.join('&')
+
+    uri.to_s
   end
 end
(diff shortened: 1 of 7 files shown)


adrianschroeter:
The request is to return to the page. Going back to main page is the same as closing and reopening the tab.

hennevogel:
> Going back to main page is the same as closing and reopening the tab.

Which is exactly what you have to do now manually....

bmwiedemann:
The whole UX is still improvable, because currently, 
* I follow a link to https://build.suse.de/request/show/314640
* get redirected to /
* log in
* get sent to / again and
* now need to find the original link again, because it is not in the browser history.

One idea of improvement would be to allow a `osc login` to re-use the ssh-key auth for the browser.
Another idea would be to keep the original requested URL and send the user there after login.

hennevogel:
Depends on #15321

eliroca:
> @bmwiedemann:  Another idea would be to keep the original requested URL and send the user there after login.

I've been testing, and found a working patch for this, as mentioned in #15315 :
```
CONFIG['proxy_auth_login_page'] + "?ReturnTo=" + URI.encode_www_form_component(request.fullpath)
```
Then we can work with that `ReturnTo` to redirect the user there.

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -172,14 +172,20 @@ def require_admin
...
-      end
-    end
+    return if User.session.present?
+    return if ::Configuration.anonymous
+
+    login_page = case CONFIG['proxy_auth_mode']
+                 when :mellon
+                   add_return_to_parameter_to_query(url: CONFIG['proxy_auth_login_page'], paramter_name: 'ReturnTo')
+                 when :ichain
+                   add_return_to_parameter_to_query(url: CONFIG['proxy_auth_login_page'], paramter_name: 'url')
```
saraycp, line 182:
```suggestion
                   add_return_to_parameter_to_query(url: CONFIG['proxy_auth_login_page'], parameter_name: 'ReturnTo')
                 when :ichain
                   add_return_to_parameter_to_query(url: CONFIG['proxy_auth_login_page'], parameter_name: 'url')
```

src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -240,4 +246,15 @@ def set_influxdb_data
       interface: :webui
     }
   end
+
+  def add_return_to_parameter_to_query(url:, paramter_name:)
+    uri = URI(url)
+    return_to = {}
+    return_to[paramter_name] = request.fullpath
```
saraycp, line 253:
```suggestion
  def add_return_to_parameter_to_query(url:, parameter_name:)
    uri = URI(url)
    return_to = {}
    return_to[parameter_name] = request.fullpath
```

eduardoj:
There is a typo in the description of the second commit: `Configurtion.anonymous`. It should be `Configuration.anonymous`.

eduardoj (changes requested):
Only a small typo in the description commit. See comment below...

saraycp (approved):
I did some debugging and ended up with a URL like `http://idp.com?ReturnTo=%2Fproject%2Fshow%2Fhome%3AAdmin` when `Configuration.anonymous` is false and `:mellon` is set.

LGTM

hennevogel:
@eduardoj pushed the typo fix

eduardoj (approved):

PR #15319, opened 2023-12-05
Implement appealing of decisions

danidoni:
Depends on #15314 

This PR implements process of appealing a report decision.


krauselukas:
[Allow users to appeal the moderators' decision (L)](https://trello.com/c/Do5xvurq/2374-allow-users-to-appeal-the-moderators-decision-l)

krauselukas:
Appeal through notification:
appeal_in_notification

New Appeal:
Screenshot 2023-12-08 at 17-11-59 Open Build Service

Show Appeal:
Screenshot 2023-12-08 at 17-12-33 Open Build Service

src/api/app/controllers/webui/appeals_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  after_action :verify_authorized
+
+  def show
+    @appeal = Appeal.find(params[:id])
+
+    authorize @appeal
+  end
+
+  def new
+    @decision = Decision.find(params[:decision_id])
```
dmarcoux, line 11: This could also use `decision_params`.

src/api/app/policies/decision_policy.rb
```diff
@@ -4,4 +4,14 @@ def create?
 
     user.is_moderator? || user.is_admin? || user.is_staff?
   end
+
+  def appeal?
```
dmarcoux, line 8: So this needs to be done here since to create an appeal, we first need to check if a decision was in favor of the report. Then, we check if the first report can be updated by the current user. Why only the first report? What if this report wasn't from the current user?

src/api/app/components/notification_mark_button_component.html.haml
```diff
@@ -1,3 +1,7 @@
+- if @notification.notifiable_type == 'Decision' && DecisionPolicy.new(@notification.subscriber, @notification.notifiable).appeal?
+  = link_to(new_decision_appeal_path(@notification.notifiable), class: 'btn btn-sm btn-outline-warning px-3 me-2',
+            title: 'Appeal Decision') do
+    %i.fas{ class: 'fa-hand' }
```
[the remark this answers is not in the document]
dmarcoux: Yes, only having an icon is often not enough on mobile... unless it's very very obvious, like the green checkmark to mark the notification as read.

src/api/app/policies/decision_policy.rb
krauselukas: So from my understanding, a decision can have many reports. Those reports should always belong to the same reportable. Thats why I only check permission based on the first report. I check if the user has permission to update the reportable, because that would mean IMO that the user should also be able to appeal the decision. I think we don't have yet a validation in place to make sure the reports of a decision always point to the same reportable, but technically there is not really a way to do that (other then having access to the console and doing it by hand). Until we have the validation, I would also be fine to go through all reports and do the permission check. What do you think? This is the card https://trello.com/c/xrjOZGa7/45-ensure-all-reports-of-a-decision-point-to-the-same-reportable-m

krauselukas: This is handled here now https://github.com/openSUSE/open-build-service/pull/15342

src/api/app/components/notification_mark_button_component.html.haml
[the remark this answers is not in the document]
krauselukas: alright I added some text 
image

src/api/app/policies/appeal_policy.rb
```diff
@@ -0,0 +1,36 @@
...
+    # 1. I reported something, but the moderator cleared my report
+    # 2. Something I did/created was reported (comment, project, etc...), then the moderator favored the report
+    #
+    # If a report was created, we can safely assume the reporter had the rights to do so, otherwise the report wouldn't even exist in the first place
+    return false unless Flipper.enabled?(:content_moderation, user)
+
+    cleared_report_on_something_user_reported? || favored_report_on_something_user_can_update?
+  end
+
+  private
```
danidoni, line 26:
```suggestion
  private

```

krauselukas: Well, those changes are extracted to this https://github.com/openSUSE/open-build-service/pull/15342 and will go away once its merged and this PR rebased

krauselukas:
I would create a card for the refactoring of the last codeclimate complain, the method was already quite long, and I would say this is not scope to refactor here right now...

src/api/spec/components/notification_filter_component_spec.rb
```diff
@@ -58,14 +59,4 @@
       expect(rendered_content).to have_link(group.title)
     end
   end
-
-  context 'when show reports is true' do
```
krauselukas, line 62: @danidoni let's enable the `:content_moderation` beta here instead

src/api/spec/components/notification_filter_component_spec.rb
```diff
@@ -58,14 +59,4 @@
...
-    before do
-      render_inline(described_class.new(selected_filter: { type: 'unread' }, show_reports: true))
```
krauselukas, line 64: `render_inline(described_class.new(selected_filter: { type: 'unread' }, user: user))`

src/api/app/components/notification_component.rb
```diff
@@ -9,19 +23,8 @@ def initialize(notification:, selected_filter:, page:, show_more:)
...
-      tag.i(class: ['fas', 'fa-comments'], title: 'Comment notification')
-    when 'Package'
-      tag.i(class: ['fas', helpers.build_status_icon(:failed)], title: 'Package notification')
-    when 'Report'
-      tag.i(class: ['fas', 'fa-flag'], title: 'Report notification')
-    when 'Decision'
-      tag.i(class: ['fas', 'fa-clipboard-check'], title: 'Report decision')
-    when 'WorkflowRun'
-      tag.i(class: ['fas', 'fa-book-open'], title: 'Workflow run')
+    if NOTIFICATION_ICON[@notification.notifiable_type].present?
```
dmarcoux, line 26: I like this refactoring. :+1:

krauselukas (approved):

PR #15322, opened 2023-12-05
Extend reports feature spec with more user roles

krauselukas:
Test that user cannot report themself or the comments, projects and
packages they own.


krauselukas:
[Add feature spec for the report+decision functionality (M)](https://trello.com/c/zGs7I0Ew/2379-add-feature-spec-for-the-reportdecision-functionality-m)

PR #15327, opened 2023-12-07
Remove more occurences of CONFIG['proxy_auth_mode']

hennevogel:
Configuration.proxy_auth_mode_enabled? is the authorative place...


src/api/spec/services/webui_controller_service/user_checker_spec.rb
```diff
@@ -28,7 +28,9 @@
     end
 
     context 'when the proxy authentication is enabled' do
-      let(:config) { { 'proxy_auth_mode' => :on } }
+      before do
+        stub_const('CONFIG', { proxy_auth_mode: :on }.with_indifferent_access)
```
bmwiedemann, line 32: This looks like we test the old `CONFIG['proxy_auth_mode']` that we want to replace. Is that intended? Should we also test the new stuff?

hennevogel: We don't want to replace the setting, we want to replace checking it in the code.

hennevogel: applied your suggestion @dmarcoux

dmarcoux (approved):

PR #15329, opened 2023-12-08
Include more information for notifications about deleted comments

dmarcoux:
For deleted comments, we don't link to the project/package/request since the comment is gone. It also clearly states that the notification is for a deleted comment.

For all comments, we show in the notification's details who wrote the comment with the `from ABC` text.

example

### DIFF
--- a/src/api/app/components/notification_notifiable_link_component.rb
+++ b/src/api/app/components/notification_notifiable_link_component.rb
@@ -47,8 +47,14 @@ class NotificationNotifiableLinkComponent < ApplicationComponent
       arch = @notification.event_payload['arch']
       "Package #{package} on #{project} project failed to build against #{repository} / #{arch}"
     # TODO: Remove `Event::CreateReport` after all existing records are migrated to the new STI classes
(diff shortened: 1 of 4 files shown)


hennevogel:
The commenter is for sure the most important information. Especially as many people delete comments after reporting them...

dmarcoux:
[Include more information about the report notifications for deleted reportables (M)](https://trello.com/c/u81KB9iX/2375-include-more-information-about-the-report-notifications-for-deleted-reportables-m)

src/api/app/components/notification_notifiable_link_component.rb
```diff
@@ -69,6 +75,7 @@ def notifiable_link_text
   # rubocop:enable Metrics/CyclomaticComplexity
 
   # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
```
dmarcoux, line 78: This method is already huge... and I'm adding more, but refactoring this method is out of scope for this PR.

eduardoj (approved):

PR #15340, opened 2023-12-11
Fix CSS class for italic font style

eduardoj:
This should have been included when migrating to Bootstrap 5.0.


ncounter (approved):

dmarcoux (approved):

ncounter:
> As a reference, this is what the [Bootstrap migration guide for the version 5.3](https://getbootstrap.com/docs/5.3/migration/) wrote:
> 
> `[BREAKING] Renamed .font-italic utility to .fst-italic for brevity and consistency with new .fst-normal utility.`

`<rant>`
Just one single letter less, an acronym you now need to remember, and a tons of find/replace to align with that :man_facepalming: 
`</rant>`

PR #15342, opened 2023-12-11
Create appeals controller, routes and views

krauselukas:
This is extracted from https://github.com/openSUSE/open-build-service/pull/15319

*TODO*

- [x] Write specs for pundit policies

### DIFF
--- a/src/api/app/policies/appeal_policy.rb
+++ b/src/api/app/policies/appeal_policy.rb
@@ -1,0 +1,40 @@
+class AppealPolicy < ApplicationPolicy
+  def new?
+    create?
+  end
+
+  def show?
+    return false unless Flipper.enabled?(:content_moderation, user)
+    return true if record.appellant == user
+
+    user.is_admin? || user.is_moderator? || user.is_staff?
+  end
+
+  # A user can create an appeal for a decision which is against them
+  # A decision is against them if:
+  # 1. They reported a project/package/etc, but a moderator took a decision which cleared the report
+  #   OR
+  # 2. They did something was reported (a spam comment, etc...) and a moderator took a decision which favored the report
+  def create?
+    return false unless Flipper.enabled?(:content_moderation, user)
+
+    decision_cleared_report_from_user? || decision_favored_report_of_action_from_user? ||
+      user.is_admin? || user.is_staff?
+  end
+
+  private
+
+  def decision_cleared_report_from_user?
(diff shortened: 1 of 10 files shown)


src/api/app/policies/appeal_policy.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def show?
+    return false unless Flipper.enabled?(:content_moderation, user)
+    return true if record.appellant == user
+
+    user.is_admin? || user.is_moderator? || user.is_staff?
+  end
+
+  def create?
+    DecisionPolicy.new(user, record.decision).appeal?
```
dmarcoux, line 14: I wouldn't delegate this to `DecisionPolicy` which then delegate to another policy depending on the reportable. We can trust that if a user reported something, it's because the user had the right to create the report. This authorization check was done at that point in time. So in this case, the user can appeal a decision which cleared the report.

dmarcoux:
The changes should be covered by specs.

src/api/app/policies/appeal_policy.rb
dmarcoux: @krauselukas and I talked. We pushed some changes.

dmarcoux:
I pushed changes for the `AppealPolicy` and specs for it, Let me know what you think.

src/api/spec/policies/appeal_policy_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    context 'when the decision cleared a report for something the appellant did' do
+      let(:report) { create(:report, reportable: create(:comment_package, user: appellant)) }
+      let(:decision) { create(:decision, kind: 'cleared', reports: [report]) }
+      let(:appeal) { create(:appeal, decision: decision, appellant: appellant) }
+
+      permissions :create? do
+        it { is_expected.not_to permit(anonymous_user, appeal) }
+        it { is_expected.not_to permit(user, appeal) }
+        it { is_expected.not_to permit(appellant, appeal) }
```
krauselukas, line 82: I think `appelant` is not available in this context

krauselukas: oh sorry, miss read

krauselukas:
Can't approve, since I'm the original author of the PR, but LGTM

danidoni (approved):

PR #15351, opened 2023-12-11
Fix the use of the appeal policy

danidoni:
While writing the feature specs the policy use for the
AppealsController#new action was not allowing the creation of the
appeals because it was missing the appellant.

### DIFF
--- a/src/api/app/controllers/webui/appeals_controller.rb
+++ b/src/api/app/controllers/webui/appeals_controller.rb
@@ -9,7 +9,7 @@ class Webui::AppealsController < Webui::WebuiController
 
   def new
     @decision = Decision.find(decision_params)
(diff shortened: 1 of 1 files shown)


dmarcoux:
This should be merged without having to wait for #15319, since the `appeals_controller` is already on master.

src/api/app/controllers/webui/appeals_controller.rb
```diff
@@ -18,7 +18,7 @@ def create
     @decision = Decision.find(decision_params)
     @appeal = Appeal.new(appeal_params)
     @appeal.decision = @decision
-    @appeal.appellant = User.session!
```
dmarcoux, line 21: If there is no logged-in user, then the authorization check will fail. This is what we want.

PR #15352, opened 2023-12-11
Decision appealing feature specs

danidoni:
Depends on ##15351


krauselukas:
This is done in another PR, closing

PR #15360, opened 2023-12-13
Enforce the presence of a reportable on a report's creation

dmarcoux:
Creating a report without a reportable is pointless since what would even be reported in this case? Nothing!

However after its creation, a report doesn't necessarily have a reportable if this reportable is deleted. This is why the validation is only on creation.

### DIFF
--- a/src/api/app/models/report.rb
+++ b/src/api/app/models/report.rb
@@ -2,6 +2,7 @@
 class Report < ApplicationRecord
   validates :reason, length: { maximum: 65_535 }
   validates :reportable_type, length: { maximum: 255 }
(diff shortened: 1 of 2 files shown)


dmarcoux:
[Prevent the creation of a report without reportable (M)](https://trello.com/c/vPDiLjIQ/2382-prevent-the-creation-of-a-report-without-reportable-m)

dmarcoux:
> You should add _why_ you only want to do this on creation in the commit message...

I thought it's obvious, but yeah, I can explain why.

dmarcoux:

dmarcoux:
I mentioned now why this validation is only on creation.

PR #15362, opened 2023-12-13
Report the author of a comment as well as their comment if wanted

dmarcoux:
This is an improvement over having to report a comment, then its author by going to their profile. It's simply quicker to do so now, all at once.

### DIFF
--- a/src/api/app/controllers/webui/reports_controller.rb
+++ b/src/api/app/controllers/webui/reports_controller.rb
@@ -10,7 +10,14 @@ class Webui::ReportsController < Webui::WebuiController
     @link_id = params[:link_id]
 
     if @report.save
-      flash[:success] = "#{@report.reportable_type} reported successfully"
+      if @report.reportable_type == 'Comment' && params[:report_comment_author].present?
+        @user.submitted_reports.create!(report_params.merge(reportable_id: @report.reportable.user_id,
+                                                            reportable_type: 'User',
+                                                            reason: "This user has been reported together with a comment they wrote. Report reason for the comment: #{@report.reason}"))
+        flash[:success] = 'Comment and its author both reported successfully'
+      else
+        flash[:success] = "#{@report.reportable_type} reported successfully"
+      end
     else
       flash[:error] = @report.errors.full_messages.to_sentence
     end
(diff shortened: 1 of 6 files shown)


src/api/app/controllers/webui/reports_controller.rb
```diff
@@ -10,7 +10,14 @@ def create
     @link_id = params[:link_id]
 
     if @report.save
-      flash[:success] = "#{@report.reportable_type} reported successfully"
+      if @report.reportable_type == 'Comment' && params[:report_comment_author].present?
+        @user.submitted_reports.create!(report_params.merge(reportable_id: @report.reportable.user_id,
```
dmarcoux, line 14: This can create multiple reports for the same user, as someone can report multiple comments of a user, while reporting this user as the same time over and over again.

Is it an issue? I don't think so. What do you think?

hennevogel: I wouldn't think about this from the can/can not angle. What do you think make sense?

krauselukas: So lets you delete and report 5 comments of a user. That would trigger the creation of 5 Event::ReportForComment and generates 5 notifications. The created reports are not an issue I would say, but the flood of notifications might be annoying. So maybe the notifications should somehow be merged

krauselukas: Ah, I guess this is already done by the `OutdatedNotificationsFinder::Report`

dmarcoux: > I wouldn't think about this from the can/can not angle. What do you think make sense?

I would go with this implementation. The UX for reporters is much better.

src/api/app/views/webui/shared/_report_modal.html.haml
```diff
@@ -28,6 +28,9 @@
               = form.text_area(:reason, class: 'form-control')
               .text-muted
                 Please include any additional information here.
+          .form-check#report-comment-author-container
+            = check_box_tag(:report_comment_author, '1', false, class: 'form-check-input')
+            = label_tag(:report_comment_author, 'Report as well the author of the comment', class: 'form-check-label')
```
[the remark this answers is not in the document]
dmarcoux: Yes, simple is better.

dmarcoux:
[Report a commenter of a comment (M)](https://trello.com/c/TMAxjN1j/2384-report-a-commenter-of-a-comment-m)

danidoni (approved):

PR #15370, opened 2023-12-14
Report a request

ncounter:
## How to test
- https://obs-reviewlab.opensuse.org/ncounter-report-a-request/request/show/1
- Click on the left side action link to `Report a Request`
- A modal dialog pops up
- Fill the report form at your wishes
- Submit the report
- Reload the Request page (to be improved by an autorefresh/autoupdate)
- Check at the Request header: a message about `You reported this request` is visible now

---

image
image
image

PR #15374, opened 2023-12-14
Support request report event

danidoni:
This PR adds the support for the event to detect a reported Request, enables the subscription of the event and the notification of that event.

This is how it looks like in the notifications page:

image

Depends on #15370


danidoni:
Splitted this into #15375, #15375 and #15378

PR #15376, opened 2023-12-15
Allow users to subscribe to reported requests events

danidoni:
You can now go into [your subscriptions](https://build.opensuse.org/my/subscriptions) page and add yourself to receive events for reported requests. 

You have to be a moderator to be able to subscribe yourself, though.

This is how it looks like:
image

## How Can I Test This?

1. Ensure your user has a "moderator" role.
2. Ensure your user is enrolled into the [beta program](http://localhost:3000/my/beta_features).
3. You can see it in the [review app instance](https://obs-reviewlab.opensuse.org/danidoni-allow-users-to-subscribe-to-reported-requests-events/my/subscriptions)

Depends on #15375

### DIFF
--- a/src/api/app/models/event/base.rb
+++ b/src/api/app/models/event/base.rb
@@ -22,6 +22,7 @@ module Event
       'Event::ReportForPackage' => 'Receive notifications for reported packages.',
       'Event::ReportForProject' => 'Receive notifications for reported projects.',
       'Event::ReportForUser' => 'Receive notifications for reported users.',
+      'Event::ReportForRequest' => 'Receive notifications for reported requests',
       'Event::ClearedDecision' => 'Receive notifications for cleared report decisions.',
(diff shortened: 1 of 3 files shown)


krauselukas:
@danidoni just the commit message of the first commit seems to not apply. Its about the notifications filter and finder, not subscriptions.

danidoni:
> @danidoni just the commit message of the first commit seems to not apply. Its about the notifications, not subscriptions.

Alright, that commit was in fact part of the following PR, the one about the notifications. Moved it there.

ncounter:
@danidoni it seems the `rss` subscription option is enabled too, is that expected?

danidoni:
> @danidoni it seems the `rss` subscription option is enabled too, is that expected?

No, that's an error. I'll fix it now.

danidoni:
> @danidoni it seems the `rss` subscription option is enabled too, is that expected?

Fixed!

PR #15377, opened 2023-12-15
 Use file_fixture for file fixtures 

hennevogel:
No need to re-invent the standard...

https://api.rubyonrails.org/classes/ActiveSupport/Testing/FileFixtures.html#method-i-file_fixture


src/api/spec/factories/packages.rb
```diff
@@ -62,8 +62,8 @@
         file_name { 'README.txt' }
         file_content { Faker::Lorem.paragraph }
         spec_file_name { "#{name}.spec" }
-        spec_file_content { Pathname.new(File.join('spec', 'fixtures', 'files', 'factory_package.spec')).read.gsub('factory_package', name) }
-        changes_file_content { Pathname.new(File.join('spec', 'fixtures', 'files', 'factory_package.changes')).read }
+        spec_file_content { file_fixture('factory_package.spec').read.gsub('factory_package', name) }
```
danidoni, line 65: Looks like the factories do not have `file_fixture` defined...

danidoni (approved):

PR #15379, opened 2023-12-15
Add decision events to subscription configuration form 

krauselukas:
Users that are participating in the content_moderation beta
program should be able to configure those subscriptions.
Admin should always be able to configure them for the defaults.

image


krauselukas:

PR #15381, opened 2023-12-15
Revert "Add email subscriptions for decision related events"

krauselukas:
This reverts commit 9eceba620ec52909eb924130f6b7f2c64f400843.

We should not create default subscriptions through a data migration, this is up to the OBS admin to configure. We now have those events integrated in the subscription configuration UI (https://github.com/openSUSE/open-build-service/pull/15379). Lets drop this data migration again.

NOTE: There is another data migration which got created after this one. This revert just drops the data migration but leaves the timestamp of the `data_schema` untouched https://github.com/openSUSE/open-build-service/blob/6e28671843dc549dbc32d34701b4c81f41224011/src/api/db/data_schema.rb#L1


krauselukas:

danidoni (approved):

PR #15384, opened 2023-12-15
Drop cassette rewinder

hennevogel:
No one is using this anymore...


danidoni:
> No one is using this anymore...

I didn't know about this! It is not useful then to clean unused cassettes?

eduardoj (approved):

dmarcoux:
> I didn't know about this! Isn't it useful then to clean up unused cassettes?

Same thought from me.

hennevogel:
> Isn't it useful then to clean up unused cassettes?

Things are only useful if they are used :)

dmarcoux:
> > Isn't it useful then to clean up unused cassettes?
> 
> Things are only useful if they are used :)

Wise words, we didn't use this. I don't think this will change even if we now know that this script exists.

dmarcoux (approved):

PR #15399, opened 2023-12-19
Fix database_consistency TODO file

eduardoj:
This fix could not be included in the pull request which updated the database_consistency rubygem.


saraycp (approved):
Indeed, the index is in the database already: https://github.com/openSUSE/open-build-service/blob/1d9c198e4e05097efbbe0791286b6bff7f13f4b2/src/api/db/schema.rb#L1132

PR #15400, opened 2023-12-19
Address code climate complain for method for notifiable type

danidoni:


src/api/app/queries/notifications_finder.rb
```diff
@@ -63,39 +79,9 @@ def for_appealed_decisions
...
-    when 'workflow_runs'
-      notifications.for_workflow_runs
-    when 'appealed_decisions'
-      notifications.for_appealed_decisions
-    else
-      notifications.unread
-    end
-  end
-  # rubocop:enable Metrics/CyclomaticComplexity
+    NOTIFIABLE_TYPE_MAP[type].call(notifications)
```
hennevogel, line 83: *You* find this less complex than a `case`?

src/api/spec/queries/notifications_finder_spec.rb
```diff
@@ -0,0 +1,131 @@
+RSpec.describe NotificationsFinder do
```
hennevogel, line 1: What are you testing here? `Hash.map`?

src/api/spec/queries/notifications_finder_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+  describe '#for_notificable_type' do
```
saraycp, line 2:
```suggestion
  describe '#for_notifiable_type' do
```

saraycp (changes requested):

src/api/app/queries/notifications_finder.rb
danidoni: Not really. I was not convinced by how it played out, but I pushed the PR to get feedback. I'm happy to see that someone else thinks the case was not that card to read.

src/api/spec/queries/notifications_finder_spec.rb
danidoni: The idea was to test that I didn't break anything by refactoring. That's how it's done, right? Test, refactor.

hennevogel: And how will this not break if you refactor and introduce the test together? And how will this break in the future?

src/api/app/queries/notifications_finder.rb
hennevogel: Yeah we are not slave to the numbers :)

src/api/spec/queries/notifications_finder_spec.rb
danidoni: The first commit is the test, passing in green. Then the refactor, that keeps passing in green. Successfully refactored!
Another thing is: do we need that test in the CI? No, once that refactor is done, that test is useless, but I left it there, because:
1) Was not sure about the whole PR
2) Prove to the reviewers that the refactoring didn't break anything

danidoni:
Alright, I'm going to close this as it's no better than the previous one.

PR #15406, opened 2023-12-20
Refactoring source controller

hennevogel:
Just moving code around to make more clear what an action is and what not.

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -22,6 +22,8 @@ class SourceController < ApplicationController
 
   before_action :require_scmsync_host_check, only: [:global_command_triggerscmsync]
 
+  before_action :require_package, only: [:show_package, :delete_package, :package_command]
+
   # GET /source
   #########
   def index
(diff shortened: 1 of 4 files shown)


src/api/app/controllers/source_controller.rb
```diff
@@ -445,6 +324,75 @@ def global_command_triggerscmsync
 
   private
 
+  # before_action for show_package, delete_package and package_command
+  def require_package
+    # init and validation
+    #--------------------
+    @deleted_package = params.key?(:deleted)
+
+    # FIXME: for OBS 3, api of branch and copy calls have target and source in the opossite place
```
danidoni, line 333:
```suggestion
    # FIXME: for OBS 3, api of branch and copy calls have target and source in the opposite place
```

src/api/app/controllers/source_controller.rb
```diff
@@ -51,21 +51,6 @@ def index
     end
   end
 
-  def set_issues_default
-    @filter_changes = @states = nil
-    @filter_changes = params[:changes].split(',') if params[:changes]
-    @states = params[:states].split(',') if params[:states]
-    @login = params[:login]
-  end
```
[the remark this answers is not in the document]
eduardoj: And I guess that the method `set_issues_default` should be renamed to `set_issues_defaults`, and not be removed...

src/api/app/controllers/source_controller.rb
hennevogel: I'm just moving this code around...

eduardoj (approved):

PR #15414, opened 2023-12-22
Re-apply token changes

saraycp:
Recently we reverted the commits related to PR #15261 as it was causing problems for some users.

After digging into the issue, we conclude the problem is on the user side. So far, they were using the `osc` token command as follows:

```
osc token --trigger <TOKEN> <PROJECT> <PACKAGE>
```

But after the code changes, it's necessary to specify the operation:

```
osc token --operation  release --trigger <TOKEN> <PROJECT> <PACKAGE>
```

This wasn't a problem before because the OBS code ended up performing the operation given by the token object and ignored the operation given in the command. But now there is [a validation](https://github.com/openSUSE/open-build-service/pull/15261/commits/25a2f0ec3a45cb13befff385cc902a7ab771c939#diff-77ed5d907b983934be02351099c20880d01bca6aab8366863839fc0721b3bb5fR59) that requires they match.

This PR re-introduces the reverted changes, adapts the error message to make it more descriptive and reduces some cognitive complexity.


eduardoj (approved):

saraycp:
@eduardoj I reduced the cognitive complexity of one of the methods, please look at the last push.

Helpful reading: https://docs.codeclimate.com/docs/cognitive-complexity#breaks-in-flow

eduardoj (approved):

adrianschroeter:
The token is supposed to be able to define the entire operation in the original design.

Yes, it is also possible to have a global token, but the original purpose was to avoid that.

It is realtive pointless and error prone to duplicate all informations in token and operation.

Also this is an incompatible change, it can not happen without announcement in former release.

So please drop this check again.

saraycp:
@adrianschroeter  I agree it will be a better practice to make an announcement. We could do it on the usual channels (blog post, IRC, mailing list, status message and Slack).

The purpose of the check is to avoid an "inconsistency" we had before. When we triggered a **release token** without specifying the operation like in `osc token --trigger <RELEASE TOKEN> <PROJECT> <PACKAGE>`, with the former code, `osc` called the `POST /trigger/runservice` endpoint (the default operation). 

```
$ osc token -h
usage: osc [global opts] token [-h] [-c] [-d TOKENID] [-o OPERATION] [-t TOKENSTRING] [--scm-token SCM_TOKEN]
...
  -o OPERATION, --operation OPERATION
                        Default is "runservice", but "branch", "release", "rebuild", or "workflow" can also be used
```

However, [the code](https://github.com/openSUSE/open-build-service/blame/e1be7fbf5fb545a610d9bac2988e2e218ed533a4/src/api/app/controllers/trigger_controller.rb#L34) ended up triggering a **release token** because it only took the `Token#type` into account, not the operation.

The questions are:

- What's the point of passing the operation to `osc` if the code ends up triggering whatever is given by the `Token#type` and not the `operation`?
- What do you mean by `global token`? The purpose of this PR is not to introduce global tokens and all the tokens in production have an operation defined (a.k.a type).

adrianschroeter:
On Dienstag, 2. Januar 2024, 14:06:26 CET Saray Cabrera Padrón wrote:
> @adrianschroeter  I agree it will be a better practice to make an announcement. We could do it on the usual channels (blog post, IRC, mailing list, status message and Slack).

well, not enough for an incompatible change, we used to annouce it before hand in a former OBS release.

> The purpose of the check is to avoid an "inconsistency" we had before. When we triggered a **release token** without specifying the operation like in `osc token --trigger <RELEASE TOKEN> <PROJECT> <PACKAGE>`, with the former code, `osc` called the `POST /trigger/runservice` endpoint (the default operation). 
> 
> ```
> $ osc token -h
> usage: osc [global opts] token [-h] [-c] [-d TOKENID] [-o OPERATION] [-t TOKENSTRING] [--scm-token SCM_TOKEN]
> ...
>   -o OPERATION, --operation OPERATION
>                         Default is "runservice", but "branch", "release", "rebuild", or "workflow" can also be used
> ```
> 
> However, [the code](https://github.com/openSUSE/open-build-service/blame/e1be7fbf5fb545a610d9bac2988e2e218ed533a4/src/api/app/controllers/trigger_controller.rb#L34) ended up triggering a **release token** because it only took the `Token#type` into account, not the operation.
> 
> The questions are:
> 
> - What's the point of passing the operation to `osc` if the code ends up triggering whatever is given by the `Token#type` and not the `operation`?

IIRC this was only intended for creating a token (this is done with the same token command in osc).

> - What do you mean by `global token`? The purpose of this PR is not to introduce global tokens and all the tokens in production have an operation defined (a.k.a type).

A global token is a token which is not bound to a package.

The original/default idea of the token is/was that the token defines exactly the operation type and 
location where it should be executed. The global token (undefined package) was a later add-on.

However, it still makes a lot of sense to support this, so it is enough for the user just to 
provide the token and the OBS instance and nothing more in external tools.

So I would not like to loose this at all for no good reason.

There still needs to be a check of course which validates that a possible manual
specified operation is not conflicting with the token registered operation.

And we may can consider if we want to support multiple operation types with one 
token in future. Not sure if that is a good idea though.

happy new year
adrian

-- 

Adrian Schroeter ***@***.***>

saraycp:
Thanks for the clarifications @adrianschroeter

saraycp:
@adrianschroeter, this is the plan for now:

- remove the validation temporarily: https://github.com/openSUSE/open-build-service/pull/15434
- request `osc` to start using the new `POST /trigger` endpoint when the `operation` is not given (instead of using `trigger/runservice` by default)
- bring the validation back

After that, we could consider getting rid of `operation` for `osc trigger` and always use the `POST /trigger` endpoint and rely on `Token#type`.

PR #15419, opened 2023-12-25
add loongarch64 architecture

hitmoon:
support loongarch64 target architecture

Since more and more distributions targets for loongarch64, it's time to add this new architecture


adrianschroeter:
we need also a migration to add the architecture to the database for existing installations.

hennevogel:
@hitmoon if you need help with the migration, holla

PR #15423, opened 2023-12-28
Migrations reminder

saraycp:
Instead of having a GitHub check that needs to pass, we better have a comment in the PR as a reminder.
The purpose is to warn that the migrations and related code changes should be shipped separately. We don't need to complicate things much.


danidoni:
Could be that you need to run the migrations locally and upload the new schema file?

saraycp:
> Could be that you need to run the migrations locally and upload the new schema file?

The check "Migrations Reminder" is failing due to permissions stuff. The exact same GitHub action works in another repo I have for testing. But it doesn't work in this repository. Still trying to figure out...

saraycp:
This is already covered in https://github.com/openSUSE/open-build-service/pull/15824

PR #15453, opened 2024-01-08
Make basic Developer Strategy work

danidoni:
Things like redirecting back still don't work though


src/api/app/controllers/webui/session_controller.rb
```diff
@@ -26,7 +26,19 @@ def destroy
...
 
   def session_creator
-    @session_creator = SessionControllerService::SessionCreator.new(params.slice(:username, :password))
+    if request.env['omniauth.auth']
+      info = request.env['omniauth.auth'].info
+      user = User.find_or_create_by(login: info['nickname'])
+      user.state = 'confirmed'
+      # We cannot create users without password, so for now I'll just create one
+      # dummy password so the validation shuts up.
+      user.password = '123456'
```
danidoni, line 35: I'm thinking into setting a Random password here

PR #15457, opened 2024-01-09
Organize the simplified chart_build_result_summary UI

ncounter:
# Before
image

# After
image

### DIFF
--- a/src/api/app/components/chart_component.html.haml
+++ b/src/api/app/components/chart_component.html.haml
@@ -5,7 +5,7 @@
         Refresh
         %i.fas.fa-sync-alt{ id: 'chart-build-reload' }
       %h4
-        Build results summary
+        Build Results Summary
     - if distinct_repositories.size > ChartComponent::MINIMUM_DISTINCT_BUILD_REPOSITORIES || raw_data.size > ChartComponent::MINIMUM_BUILD_RESULTS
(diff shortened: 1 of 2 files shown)


saraycp:
We discussed offline that we are going to improve this design for color-blind people after the detailed view of the build results is finished. The idea is to use a different icon per status.

src/api/app/components/chart_component.html.haml
```diff
@@ -1,10 +1,11 @@
...
-      Refresh
-      %i.fas.fa-sync-alt{ id: 'chart-build-reload' }
-    %h4
-      Build results summary
+    .mb-4
+      .btn.btn-outline-primary.float-end{ onclick: 'updateChartBuildResults()', accesskey: 'r', title: 'Refresh Chart Build Results' }
+        Refresh
+        %i.fas.fa-sync-alt{ id: 'chart-build-reload' }
+      %h4
+        Build results summary
```
saraycp, line 8:
```suggestion
        Build Results Summary
```

saraycp (changes requested):

src/api/app/components/chart_component.html.haml
ncounter: good catch, done

saraycp (approved):

PR #15461, opened 2024-01-10
Dark theme for the UI

mdeniz:
Added a Feature toggle on the Beta Program to activate the Dark Theme.

imagen

Some fixes were needed on the dark theme because we had light backgrounds in some cases. Changed the top navigation to black on the dark theme, I think it works better than the lighter gray it had in comparison with the body background.


src/api/app/views/webui/package/_buildstatus.html.haml
```diff
@@ -22,7 +22,7 @@
           repository_name = result.repository.tr('.', '_')
           expanded = repository_expanded?(collapsed_repositories, repository_name, package_name)
         - if result.repository != previous_repo
-          .d-flex.flex-row.py-1.bg-light.ps-1.ps-sm-2
```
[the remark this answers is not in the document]
mdeniz: I created a specific one

src/api/app/views/webui/shared/_editor.html.haml
```diff
@@ -7,7 +7,7 @@
 
 %div{ class: "card #{'editable' unless style[:read_only]}" }
   - unless style[:read_only]
-    .sticky-top.bg-light{ id: "top_#{uid}" }
```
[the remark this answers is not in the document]
mdeniz: I added a specific one

src/api/app/views/webui/shared/_repositories_flag_table.html.haml
```diff
@@ -1,6 +1,6 @@
 .table-responsive
   %table.table.table-hover.table-sm{ id: table_id }
-    %thead.table-light
```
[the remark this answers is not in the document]
mdeniz: I added a specific one

src/api/app/views/webui/subscriptions/_subscriptions_form.html.haml
```diff
@@ -1,6 +1,6 @@
 - i = 0
 - subscriptions_form.subscriptions_by_event.each do |subscription|
-  .card.bg-light.mb-2
```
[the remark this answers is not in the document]
mdeniz: I added a specific one

mdeniz:
Hey @hellcp-work, thanks for the feedback. I changed many things:

- comments and changes requested
- fixed more coloring issues I didn't spot before
- removed some unused CSS class
- dropped the feature toggle for now as this is WIP. I will implement a new field for the profile to choose the theme, dark mode is forced temporarily at this moment

This is a very tailored approach to implement a dark mode, to be honest, I think that the next step after having this merged is to provide a better way of having many themes. Nowadays colors are scattered among all the SCSS files and also in HAML templates based on classes applied directly to elements. We need to leverage more on the modern version of Bootstrap using the secondary and tertiary variations for instance.

src/api/app/assets/stylesheets/webui/build-results.scss
```diff
@@ -18,6 +18,18 @@
   }
 }
 
+[data-bs-theme="light"] {
```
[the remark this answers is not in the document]
mdeniz: Using this I see no difference between the header and the background in both themes. Not sure if this is what we want...
imagen
imagen

src/api/app/models/user.rb
```diff
@@ -10,6 +10,9 @@ class User < ApplicationRecord
   NOBODY_LOGIN = '_nobody_'.freeze
   MAX_BIOGRAPHY_LENGTH_ALLOWED = 250
 
+  COLOR_THEMES = ['light', 'dark'].freeze
```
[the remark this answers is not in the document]
mdeniz: I'll replace "light" with "system" instead because we will only have two options and we won't craft a "light" theme right away.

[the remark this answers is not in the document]
mdeniz: So, in the case of "system" you want to use the one coming from the OS setting, right?  We could check with `@media (prefers-color-scheme: dark)` to add the overrides. But in the code, I used `@include color-mode(dark)` to define the overrides which are based on the `data-bs-theme` attribute from the HTML. 

I'm not sure how to use one or the other depending on the value of the column and point to the same dark theme without repeating all the overrides inside the `prefers-color-schema: dark` section. We need to specify `$color-mode-type`, which controls the if you are based on the `media-query` or the bootstrap color scheme system... and you can only choose one or the other.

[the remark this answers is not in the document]
mdeniz: Ok, I added a small javascript snippet to handle the values coming from DB. If the user selects `system` in the profile, it uses the `prefers-color-scheme` from the browser, otherwise, it goes straight to that value set in the profile.

src/api/app/assets/javascripts/webui/color_themes.js
```diff
@@ -0,0 +1,27 @@
...
+ * Color mode toggler for Bootstrap's docs (https://getbootstrap.com/)
+ * Copyright 2011-2023 The Bootstrap Authors
+ * Licensed under the Creative Commons Attribution 3.0 Unported License.
+ */
+
+(() => {
+  'use strict'
+  
+  const getPreferredTheme = () => {
+    const userTheme = document.documentElement.getAttribute('data-bs-theme-from-user')
```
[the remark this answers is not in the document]
mdeniz: Much more elegant :+1:

src/api/app/assets/stylesheets/webui/colors.scss
```diff
@@ -1,8 +1,14 @@
...
   background-color: $gray-500;
 }
+
+$navbar-themed-bg: $gray-800;
+$breadcrumbs-item-text-color: $gray-700;
+
+// Dark theme overrides
+@include color-mode(dark)  {
+  $breadcrumbs-item-text-color: $gray-400;
+  $navbar-themed-bg: #121212;
```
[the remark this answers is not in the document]
mdeniz: Done, I followed the documentation and its a bit different, the color-mode part needs to be out, see the code

mdeniz:
@hellcp-work let me know if you think there is anything that needs to be done in this and the other PR #15474

BTW, not sure why CircleCI and Codeclimate aren't running.

src/api/app/assets/stylesheets/webui/build-results.scss
```diff
@@ -66,6 +70,8 @@
   .build-info-content {
     @extend .py-2;
     @extend .px-3;
+    padding-top: 0.4rem !important;
+    padding-bottom: 0.4rem !important;
```
hellcp, line 74: You already have py-2 above, it's one or the other

src/api/app/assets/stylesheets/webui/colors.scss
```diff
@@ -1,8 +1,16 @@
...
+
+:root {
+  --navbar-themed-bg: #{$gray-800};
+  --breadcrumbs-item-text-color: #{$gray-700};
+}
+
+@include color-mode(dark, true) {
+  --navbar-themed-bg: #{$themed-dark-nav-bg};
+  --breadcrumbs-item-text-color: #{$gray-400};
+}
```
hellcp, line 16: Don't want to have variables assigned without some kind of a parent (even if it's `:root`)
```suggestion

  @include color-mode(dark, true) {
    --navbar-themed-bg: #{$themed-dark-nav-bg};
    --breadcrumbs-item-text-color: #{$gray-400};
  }
}
```

src/api/app/views/webui/user/_basic_info.html.haml
```diff
@@ -43,6 +43,11 @@
       = title.upcase
   %p= render_as_markdown(user.biography)
 
+  .mt-4
+    - if User.session
+      Color theme:
+      = user.color_theme
+
```
hellcp, line 50: I don't think other people have to know what color theme I'm using
```suggestion
```

src/api/app/models/user.rb
```diff
@@ -10,6 +10,9 @@ class User < ApplicationRecord
   NOBODY_LOGIN = '_nobody_'.freeze
   MAX_BIOGRAPHY_LENGTH_ALLOWED = 250
 
+  COLOR_THEMES = ['system', 'light', 'dark'].freeze
+  enum :color_theme, COLOR_THEMES
```
hellcp, line 14:
```suggestion
  enum :color_theme, ['system', 'light', 'dark']
```

src/api/app/views/webui/user/_edit_account_form.html.haml
```diff
@@ -18,6 +18,9 @@
       %small.form-text#bio-chars-counter
   .mb-3
     = f.text_field(:email, required: true, email: true, placeholder: 'Email', readonly: behind_proxy, class: 'form-control')
+  .mb-3
+    = f.label :color_theme, "Theme"
+    = f.select(:color_theme, options_for_select(User::COLOR_THEMES), {}, class: 'form-select')
```
hellcp, line 23:
```suggestion
    = f.select(:color_theme, options_for_select(User.color_themes.keys), {}, class: 'form-select')
```

src/api/app/models/user.rb
```diff
@@ -121,6 +124,7 @@ class User < ApplicationRecord
   validates :password, confirmation: true, allow_blank: true
   validates :biography, length: { maximum: MAX_BIOGRAPHY_LENGTH_ALLOWED }
   validates :rss_secret, uniqueness: true, length: { maximum: 200 }, allow_blank: true
+  validates :color_theme, inclusion: { in: COLOR_THEMES }
```
hellcp, line 127:
```suggestion
  validates :color_theme, inclusion: { in: color_themes.keys }
```

src/api/app/views/layouts/webui/webui.html.haml
```diff
@@ -1,5 +1,5 @@
 !!!
-%html{ lang: 'en' }
+%html{ lang: 'en', "data-bs-theme-from-user": User.session&.color_theme || "system" }
```
hellcp, line 2: I am thinking that we should actually keep the beta flag for this specifically, only display this if a person has explicitly enabled a beta of the feature. The reason being is that I think there are going to be some issues still and I don't really want to risk interrupting the work of people who aren't willing to try things out. If we only have this data attribute for beta users, and only display the dropdown in user settings for beta users then we won't have this issue.

src/api/app/assets/stylesheets/webui/build-results.scss
mdeniz: I'll remove the `@extend .py-2`, this is the culprit of an issue, visually it shows little dots below each row for the archs results, those little dots just aren't dots... are parts of the text that are hidden in the question mark icon :smile:

src/api/app/assets/stylesheets/webui/colors.scss
mdeniz: Did you try that on your side? 
I did it and the navbar stopped having the proper background in the dark mode, it defaults to the light one.
imagen
imagen

Can it be that the `@include color-mode(dark, true)` part does not work there?

src/api/app/views/layouts/webui/webui.html.haml
mdeniz: OMG, I just removed all that work hahaha.
Ok, it makes sense to me... that's why I started with that in the first place.
I'll add that after all the rest is settled, WDYT?

src/api/app/assets/stylesheets/webui/build-results.scss
hellcp: I would probably just move the text from that collapse lower then, but I guess it doesn't matter either way

src/api/app/views/layouts/webui/webui.html.haml
hellcp: Yeah, I want it to one day move it out of beta, but I think that it's not now and we are better safe than to disturb the software so many people use for work :P

src/api/app/assets/stylesheets/webui/colors.scss
mdeniz: I think I saw it somewhere... nevermind, even without the `true` parameter it doesn't work...
I'm not sure what to do TBH

hellcp: Hm, guess you will have to invert it then and place `:root` on the inside of `@include`  to encapsulate those variables

mdeniz:
Ok, @hellcp-work is there anything left?

mdeniz:
Thanks @hellcp-work for all your feedback and care :wink:

uncomfyhalomacro:
@mdeniz LOVE YOUR WORK <3

PR #15470, opened 2024-01-11
Fix for code snippet disappearance issue for CRUD operations on comment 

### LINKED ISSUE #15241: Replying to comment on a code change needs to be fixed
# Issue Description

If you are on the Conversation tab and try to reply to a comment on code changes you'll see something like this:

Screenshot 2023-11-20 at 12-17-08 Open Build Service

Right after you submit the form, the code snippet disappears, as you can see here:

Screenshot 2023-11-20 at 12-17-25 Open Build Service

Then, if you refresh the page, the reply is duplicated and displayed in the wrong position, as shown in this image:

Screenshot 2023-11-20 at 12-17-52 Open Build Service

Expected Result
============

The reply is displayed only once inside the conversation tree.

How to Reproduce
===============
1. Join beta
2. Go to a request Changes tab
3. Add a comment in a line of code
4. Go to the Conversation tab
5. Try to reply to the comment you created in 3.
6. See the error

rubhanazeem:
Fixes #15241

In addition to the issue mentioned above, there is the same behavior in the changes tab and they both are directly related to the same piece of code.  One can not be fixed without another. This PR covers:
- [x] Fix CRUD operations on comments on the comments timeline
- [x] Fix CRUD operations on comments in the changes tab

Notes for test:
1. Test all the comments features(CRUD and also report/moderation) on timeline. If there is a comment on a line of code, it should not disappear.
2. Test everything mentioned in 1 for the changes tab. Previously there has been a blank line when you Delete or update a comment. Now, it should be fixed. There was a [separate card](https://trello.com/c/gHn4spEf/116-handle-deleting-comments-in-changes-tab-better-m) for this but this PR also fixed that issue.
3. Test comments on projects/packages

### DIFF
--- a/src/api/app/controllers/webui/comments_controller.rb
+++ b/src/api/app/controllers/webui/comments_controller.rb
@@ -4,58 +4,52 @@ class Webui::CommentsController < Webui::WebuiController
   before_action :set_comment, only: [:moderate, :history]
 
   def create
-    if @commented.nil?
-      flash.now[:error] = "Failed to create comment: This #{@commentable_type.name.downcase} does not exist anymore."
-      render partial: 'layouts/webui/flash' and return
-    end
+    return commented_unavailable if @commented.nil?
 
-    comment = @commented.comments.new(permitted_params)
-    authorize comment, :create?
-    User.session!.comments << comment
-    @commentable = comment.commentable
+    @comment = @commented.comments.new(permitted_params)
+    authorize @comment, :create?
+    User.session!.comments << @comment
+    @commentable = @comment.commentable
 
-    status = if comment.save
+    status = if @comment.save
                flash.now[:success] = 'Comment created successfully.'
                :ok
              else
-               flash.now[:error] = "Failed to create comment: #{comment.errors.full_messages.to_sentence}."
+               flash.now[:error] = "Failed to create comment: #{@comment.errors.full_messages.to_sentence}."
                :unprocessable_entity
              end
 
-    if Flipper.enabled?(:request_show_redesign, User.session) && ['BsRequest', 'BsRequestAction'].include?(comment.commentable_type)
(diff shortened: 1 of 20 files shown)


src/api/app/views/webui/comment/_render_timeline.js.erb
krauselukas, line 1: @rubhanazeem I wouldnt call this view `render_timeline`, a view implies that it renders something

rubhanazeem: It is actually re-rendering the timeline.  What do you suggest?

src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -141,6 +119,28 @@ def history
...
   end
 
+  def render_timeline
+    @commentable ||= @comment.commentable
+    bs_request = @comment.commentable_type == 'BsRequestAction' ? @commentable.bs_request : @commentable
+    if params[:file_name].present?
+      action = bs_request.webui_actions(diffs: true, action_id: @commentable.id, cacheonly: 1).first
+      sourcediff = action[:sourcediff].first
+      file_name = params[:file_name]
+      file_index = sourcediff['filenames'].index(file_name)
```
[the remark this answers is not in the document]
rubhanazeem: We discussed that on the call and indeed file_name can be obtained through comments. But file_name is also used as an identifier for rendering of changes. So, we are currently trying to find a way if we can achieve this in a simpler way

danidoni:
I find this PR hard to follow... Is everything really needed for the fix? Can we split this PR in two or more?

rubhanazeem:
> I find this PR hard to follow... Is everything really needed for the fix? Can we split this PR in two or more?

I'll try to simply it

rubhanazeem:
Closing it in favor of #15501.

PR #15473, opened 2024-01-11
Introduce rpmlint log parser

eduardoj:


src/api/app/services/rpmlint_log_parser.rb
```diff
@@ -0,0 +1,51 @@
+class RpmlintLogParser
+  attr_reader :errors, :badness, :warnings, :info
+
+  def initialize(content:)
+    @content = content || ''
```
danidoni, line 5:
```suggestion
    @content = content || []
```

danidoni: We're iterating over it below, so it should be initialized to an enumerable

eduardoj: We're iterating over a string below, using [each_line](https://ruby-doc.org/3.1.4/String.html#method-i-each_line).

I would agree with you if we were iterating over an enumerable, which doesn't have the `each_line` method.

danidoni: Alright, it understand new lines so it works... Dismissing

src/api/app/services/rpmlint_log_parser.rb
```diff
@@ -0,0 +1,53 @@
...
```
rubhanazeem, line 5: You can just initialize the default value in params. 

```
def initialize(content: '')
   @content = content
   ...

```

src/api/.rubocop_todo.yml
```diff
@@ -1252,6 +1252,7 @@ ViewComponent/MissingPreviewFile:
     - 'app/components/delete_confirmation_dialog_component.rb'
     - 'app/components/notification_component.rb'
     - 'app/components/request_decision_component.rb'
+    - 'app/components/rpm_lint_component.rb'
```
rubhanazeem, line 1255: Any specific reason to ignore this? This is a new component, so it should be simple to fix the rubocop complaints. No?

src/api/app/components/rpm_lint_component.rb
```diff
@@ -0,0 +1,24 @@
...
+    warnings = rpmlint_log_parser.warnings.select(&threshold_filter)
+    info = rpmlint_log_parser.info.select(&threshold_filter)
+    badness = rpmlint_log_parser.badness.select(&threshold_filter).sort(&badness_sorter).reverse
+
+    @issues_chart_data = [{ name: 'Errors', data: @errors },
+                          { name: 'Warnings', data: warnings },
+                          { name: 'Info', data: info }]
+    @badness_chart_data = [{ name: 'Badness', data: badness }]
+  end
+end
```
rubhanazeem, line 24: Are we planning to cover specs for this component?

src/api/app/services/rpmlint_log_parser.rb
eduardoj: True! Applied.

src/api/.rubocop_todo.yml
eduardoj: In this particular case we only need a preview file. I'll add it.

src/api/app/components/rpm_lint_component.rb
danidoni: Let me add it

src/api/.rubocop_todo.yml
eduardoj: Done.

src/api/app/components/rpm_lint_component.rb
danidoni: Done

rubhanazeem: Thank you

krauselukas (changes requested):
@eduardoj this is not behind the beta flag, I would say we should set one for this. And maybe only use it for the requests view for now if possible. In the rpm lint tab on the package view it looks like it deserves a bit UI work first

krauselukas:
image

eduardoj:
> @eduardoj this is not behind the beta flag, I would say we should set one for this. And maybe only use it for the requests view for now if possible. In the rpm lint tab on the package view it looks like it deserves a bit UI work first

True. I'll have a look. Thanks!

Also, @rubhanazeem and I realized the graphic doesn't get plotted in some cases.

I will switch the PR to Draft until the issue is fixed.

src/api/app/views/webui/package/_rpmlint_log.html.haml
```diff
@@ -1,2 +1,4 @@
+- if plot_graphic
```
danidoni, line 1: What about using the word `render_chart`? We already use the word `chart` it in the `RpmLintComponent`, and we use `render` everywhere else. Using new words to describe the same will increase the cognitive load...

eduardoj: Agree. Changed. Thanks!

krauselukas (approved):

PR #15474, opened 2024-01-12
Add Color Theme column to users

mdeniz:
This PR needs to be in isolation as it has a migration. This is related to the work for adding color themes.
Schema migration is mandatory to add a column to the users' table. 
This new field is needed to provide a selector among multiple themes on the profile of the users.


mdeniz:
@hellcp this is needed to be merged before #15461

src/api/db/migrate/20240111231639_add_theme_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddThemeToUser < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :color_theme, :string, limit: 30, default: "light"
```
hellcp, line 3: Make it an `integer` and then enum in the model in the next PR. I would also vote the default to be `system` and not `light` (though for integer just set it to 0 and think about what values in the enum will mean)
```suggestion
    add_column :users, :color_theme, :integer, default: 0, null: false
```

mdeniz: Done here, please review again

eduardoj:
Just don't forget the changes to be made by the annotate gem to files `app/models/unregistered_user.rb` and `app/models/user.rb`. They don't show up with `bin/rails db:migrate:with_data`, but with `bin/rails db:migrate`.

mdeniz:
If I add the annotates in this PR then the `Isolated Migrations / check_changed_files (pull_request)`  fails because I "touched" code at the same time as the migrations :shrug: 
I had them included before, but it failed and then I removed them to let it pass :smile: 
I added the annotations in the #15461 PR which should be based on this changes after merged

eduardoj:
> If I add the annotations in this PR then the `Isolated Migrations / check_changed_files (pull_request)` fails because I "touched" code at the same time as the migrations 🤷 I had them included before, but it failed and then I removed them to let it pass 😄 I added the annotations in the #15461 PR which should be based on this changes after merged

Then we have an issue with the linter, unfortunately... :thinking:

mdeniz:
@hellcp-work and @eduardoj let me know if you think there is anything that needs to be done in this and the other PR #15461 

BTW, not sure why CircleCI and Codeclimate aren't running.

PR #15476, opened 2024-01-12
[api] fix project update policy

adrianschroeter:
It makes no difference anymore between a project meta change and any other project level changes.

Eg. a project maintainer, who is no admin should still be able to change any other project level changes, except the project meta. eg. updating build config, key managemend, releases or package creation.


adrianschroeter:
This has nothing to-do with scmsync, it is about a regression when using admin only elements in project meta, like DoD.
Maintainers still need to be able to write to build config

PR #15477, opened 2024-01-12
Extract rpmlint log from build _log files

eduardoj:
Some failed builds don't generate the `rpmlint.log` file.

Create a service to retrieve information about rpmlint for failed builds from the `_log` internal build file.

Introduce this change under the `request_show_redesign` beta feature.

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -24,8 +24,6 @@ class Webui::PackageController < Webui::WebuiController
   # FIXME: Remove this before_action, it's doing validation and authorization at the same time
   before_action :check_package_name_for_new, only: [:create]
 
-  before_action :handle_parameters_for_rpmlint_log, only: [:rpmlint_log]
-
   prepend_before_action :lockout_spiders, only: [:revisions, :dependency, :rdiff, :requests]
 
   after_action :verify_authorized, only: [:new, :create, :remove_file, :remove, :abort_build, :trigger_rebuild, :wipe_binaries, :save_meta, :save, :abort_build]
@@ -470,11 +468,16 @@ class Webui::PackageController < Webui::WebuiController
     end
   end
 
+  def rpmlint_log_params
+    params.require([:project, :package, :repository, :architecture])
+    params.slice(:project, :package, :repository, :architecture).permit!
+  end
+
   def rpmlint_log
-    rpmlint_log_file = Backend::Api::BuildResults::Binaries.rpmlint_log(params[:project], params[:package], params[:repository], params[:architecture])
+    rpmlint_log_file = RpmlintLogExtractor.new(rpmlint_log_params).call
+    render plain: 'No rpmlint log' and return if rpmlint_log_file.blank?
+
     render partial: 'rpmlint_log', locals: { rpmlint_log_file: rpmlint_log_file }
-  rescue Backend::NotFoundError
-    render plain: 'No rpmlint log'
   end
 
   def meta
@@ -570,10 +573,6 @@ class Webui::PackageController < Webui::WebuiController
 
     flash[:error] = "Couldn't find repository '#{params[:repository]}'"
     redirect_to package_show_path(project: @project, package: @package)
-  end
-
-  def handle_parameters_for_rpmlint_log
-    params.require([:project, :package, :repository, :architecture])
   end
 
   def set_file_details
--- a/src/api/app/services/rpmlint_log_extractor.rb
+++ b/src/api/app/services/rpmlint_log_extractor.rb
@@ -1,0 +1,56 @@
+class RpmlintLogExtractor
+  attr_reader :project, :package, :repository, :architecture, :log_content
+
+  def initialize(attributes = {})
+    @project = attributes[:project]
+    @package = attributes[:package]
+    @repository = attributes[:repository]
+    @architecture = attributes[:architecture]
+
+    @log_content = ''
+  end
+
+  def call
+    # Retrieve the content of the 'rpmlint.log' file
+    Backend::Api::BuildResults::Binaries.rpmlint_log(project, package, repository, architecture)
+  rescue Backend::NotFoundError => e
+    # TODO: Remove this `unless` condition once request_show_redesign is rolled out
+    return unless Flipper.enabled?(:request_show_redesign, User.session)
+
+    # Case of removed project, package, repository or architecture:
+    return unless e.summary.include?('rpmlint.log: No such file or directory')
+
+    # Case of debian builds:
+    return unless rpms_exist?
+
+    retrieve_rpmlint_log_from_log
+  end
+
+  private
+
+  def rpms_exist?
+    binaries = Backend::Api::BuildResults::Binaries.files(project, repository, architecture, package)
+    binaries.match?(/<binary filename="\S+\.rpm"/)
+  end
+
+  def retrieve_rpmlint_log_from_log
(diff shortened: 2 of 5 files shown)


src/api/app/controllers/webui/package_controller.rb
```diff
@@ -471,10 +471,10 @@ def rpmlint_result
   end
 
   def rpmlint_log
-    rpmlint_log_file = Backend::Api::BuildResults::Binaries.rpmlint_log(params[:project], params[:package], params[:repository], params[:architecture])
+    rpmlint_log_file = RpmlintLogExtractor.new(params.slice(:project, :package, :repository, :architecture).permit!).call
```
danidoni, line 474: It's a Rails convention to have the strong params declaration in a private method, we should follow that.

src/api/app/services/rpmlint_log_extractor.rb
```diff
@@ -0,0 +1,56 @@
+class RpmlintLogExtractor
+  attr_reader :project, :package, :repository, :architecture, :log_content
+
+  def initialize(params)
```
danidoni, line 4: Better use regular arguments or positional arguments, instead of hashes

src/api/app/services/rpmlint_log_extractor.rb
```diff
@@ -0,0 +1,56 @@
...
+    @package = params[:package]
+    @repository = params[:repository]
+    @architecture = params[:architecture]
+
+    @log_content = ''
+  end
+
+  def call
+    # Retrieve the content of the 'rpmlint.log' file
+    Backend::Api::BuildResults::Binaries.rpmlint_log(project, package, repository, architecture)
```
danidoni, line 15: Can we check for the existence of the `rpmlint.log` file instead of using an exception for flow control?

eduardoj: Using an exception for flow control is how we check for the existence of the `rpmlint.log` file, now. See the implementation of the `rpmlint_log` method: https://github.com/openSUSE/open-build-service/blob/75a7d0df472b3532a732ec3a7fe79f175b719bd1/src/api/app/lib/backend/api/build_results/binaries.rb#L47-L51

Do you have another approach in mind?

eduardoj:
> The parser would benefit from some specs checking it works as expected...

I added a commit with a spec. Please take a look.

src/api/app/services/rpmlint_log_extractor.rb
eduardoj: I renamed `params` to `attributes`, and keep using it as a hash parameter. We use this convention in other models and services:
```
✗ grep -r -E "def.+initialize.+attributes" src/api/app/{models,services}
src/api/app/models/workflow.rb:  def initialize(attributes = {})
src/api/app/models/workflow/step.rb:  def initialize(attributes = {})
src/api/app/models/scm_webhook.rb:  def initialize(attributes = {})
src/api/app/models/package_service_error_file.rb:  def initialize(attributes = {})
src/api/app/models/project_meta_file.rb:  def initialize(attributes = {})
src/api/app/models/local_build_result/for_package.rb:    def initialize(attributes = {})
src/api/app/models/project_service_file.rb:  def initialize(attributes = {})
src/api/app/models/patchinfo.rb:  def initialize(attributes = {})
src/api/app/models/project_config_file.rb:  def initialize(attributes = {})
src/api/app/models/local_build_statistic/for_package.rb:    def initialize(attributes = {})
src/api/app/models/package_meta_file.rb:  def initialize(attributes = {})
src/api/app/services/rpmlint_log_extractor.rb:  def initialize(attributes = {})
```

Would you say this is a bad practice?

In that case, what kind of arguments should we use in a method when we call it with a permitted slice of `params`?

src/api/app/controllers/webui/package_controller.rb
eduardoj: True. I created the `rpmlint_log_params` method. Please take a look again.

src/api/app/services/rpmlint_log_extractor.rb
```diff
@@ -0,0 +1,56 @@
...
+  rescue Backend::NotFoundError => e
+    # TODO: Remove this `unless` condition once request_show_redesign is rolled out
+    return unless Flipper.enabled?(:request_show_redesign, User.session)
+
+    # Case of removed project, package, repository or architecture:
+    return unless e.summary.include?('rpmlint.log: No such file or directory')
```
krauselukas, line 21: @eduardoj so incase we cannot find the rpmlint.log file and rescue, will this message not always be included in the summary? I havent tested it myself yet, first wanted to ask you

krauselukas: ah sorry, its `unless` get it now :)

eduardoj: @krauselukas, this backend call could return with a 404 also when the project, package, repository or architecture don't exist. That's why I put the comment just before:
```
# Case of removed project, package, repository or architecture:
```
In those cases, no, the message returned is not related to the `rpmlint.log` file.

src/api/app/services/rpmlint_log_extractor.rb
eduardoj: I created another commit _"Use positional arguments"_ that refactors a before_action method and uses positional arguments.

We can merge this commit with the previous one. Or remove it, depending on what we decide.

eduardoj: After talking with @krauselukas, we agreed on using named arguments. The `rpmlint_log_params` method still checks for required parameters and return an `ActionController::Parameters` instance later on.

src/api/spec/services/rpmlint_log_extractor_spec.rb
```diff
@@ -0,0 +1,24 @@
+RSpec.describe RpmlintLogExtractor, type: :service do
+  let(:parameters) { ActionController::Parameters.new(project: 'home:user1', package: 'package1', repository: 'repo1', architecture: 'arch1') }
+
+  subject { described_class.new(parameters).call }
+
+  describe '#call' do
+    context 'no rpmlint.log file found, beta program, rpm files built, `_log` file with rpmlint log output' do
```
krauselukas, line 7:
```suggestion
    context 'no rpmlint.log file available since error exceedes allowed badness level' do
```

src/api/spec/services/rpmlint_log_extractor_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+          .and_raise(Backend::NotFoundError, 'rpmlint.log: No such file or directory')
+        allow(Backend::Api::BuildResults::Binaries).to receive(:files)
+          .with('home:user1', 'repo1', 'arch1', 'package1')
+          .and_return("<binarylist><binary filename=\"ctris-0.42.1-8.1.src.rpm\" size=\"26772\" mtime=\"1704975490\"/></binarylist>\n")
+        allow(Backend::Api::BuildResults::Binaries).to receive(:file)
+          .with('home:user1', 'repo1', 'arch1', 'package1', '_log')
+          .and_return(file_fixture('rpmlint_log_extractor_log').read)
+      end
+
+      it { expect(subject).to eq(file_fixture('rpmlint_log_extractor_expected').read) }
```
krauselukas, line 21:
```suggestion
      it 'extracts the summary from the build log file' do
        expect(subject).to eq(file_fixture('rpmlint_log_extractor_expected').read)
      end
```

krauselukas (approved):

PR #15497, opened 2024-01-17
Render a chart with the summary of the RPM lint issues

danidoni:


danidoni:
Closing in favor of #15473

PR #15501, opened 2024-01-17
Make diffs on comments render consistently

hellcp-work:
This removes the issue of diffs disappearing on actions on comments. Depends on #15500

### DIFF
--- a/src/api/app/controllers/webui/comments_controller.rb
+++ b/src/api/app/controllers/webui/comments_controller.rb
@@ -21,7 +21,7 @@ class Webui::CommentsController < Webui::WebuiController
 
     if Flipper.enabled?(:request_show_redesign, User.session) && ['BsRequest', 'BsRequestAction'].include?(@comment.commentable_type)
       render(partial: 'webui/comment/beta/comments_thread',
-             locals: { comment: @comment.root, commentable: @commentable, level: 1 },
+             locals: { comment: @comment.root, commentable: @commentable, level: 1, diff: diff },
              status: status)
     else
       render(partial: 'webui/comment/comment_list',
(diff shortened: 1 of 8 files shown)


src/api/app/controllers/webui/comments_controller.rb
```diff
@@ -165,4 +165,15 @@ def commented_unavailable
     flash.now[:error] = "Failed to create comment: This #{@commentable_type.name.downcase} does not exist anymore."
     render partial: 'layouts/webui/flash'
   end
+
+  def diff
```
rubhanazeem, line 169: @hellcp-work can you please add some specs for this new method?

danidoni:
> This is the related card that explains the issue https://trello.com/c/u1uBm8jt/2397-code-snippet-disappears-when-replying-to-a-comment-on-code-changes-m

Alright, but it's a best practice to include the "why" in the PR's description...

src/api/app/controllers/webui/comments_controller.rb
rubhanazeem: I'm taking care of the specs. I will add it in a follow-up pr so this PR is not blocked

src/api/spec/features/webui/requests/diff_comments_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      before do
+        Flipper.enable(:request_show_redesign, admin)
+        login admin
+        visit request_show_path(bs_request)
+      end
+
+      it 'displays the comment in the timeline' do
+        expect(page).to have_text(comment.body)
+        expect(page).to have_text('target_project/target_package > package_a.changes')
+      end
```
danidoni, line 33: I think this context is trying to do too much: either you move this into it's own context (like "rendering comments") so you can free the before block to prepare for this test (that's why you have to cram everything un the `it` block below) or you remove this expectation and move the clicking in the before block above

danidoni (approved):

PR #15502, opened 2024-01-17
Send metrics related to the rpmlint log extractor

eduardoj:


src/api/app/services/rpmlint_log_extractor.rb
```diff
@@ -8,19 +8,31 @@ def initialize(attributes = {})
...
+      # TODO: Remove the Flipper.enable? part of the condition once request_show_redesign is rolled out
+      if (@beta_feature = Flipper.enabled?(:request_show_redesign, User.session)) && (@rpmlint_log_file_not_found = e.summary.include?('rpmlint.log: No such file or directory'))
+        # Case of removed project, package, repository or architecture:
+        retrieve_rpmlint_log_from_log
+      end
+    end
+
+    time_delta = Time.now.to_i - time_start
+    metrics = "beta=#{@beta_feature},rpmlint_log_file_not_found=#{@rpmlint_log_file_not_found},mark_not_found=#{@mark_not_found}"
+    RabbitmqBus.send_to_bus('rpmlint_log', "extractor,#{metrics} value=#{time_delta}")
```
danidoni, line 33: I would refactor this into a method that accepts a block, something like:

```ruby
time_it do
   # Retrieve the content of the 'rpmlint.log' file
   ...
end
```

Inside the `time_it` method lives the code to time the execution, capturing the bloc of code with a `yield` or `&block`

```ruby
def time_it(&block)
  time_start = Time.now.to_i
  block.call
  time_delta = Time.now.to_i - time_start
  metrics = "beta=#{@beta_feature},rpmlint_log_file_not_found=#{@rpmlint_log_file_not_found},mark_not_found=#{@mark_not_found}"
  RabbitmqBus.send_to_bus('rpmlint_log', "extractor,#{metrics} value=#{time_delta}")
end
```

eduardoj: > This way we decouple the code responsible of timing the execution from the code responsible of fetching the log

I would argue that:
- the `call` method isn't so big, so we can understand the code
- we will not reuse the `time_it` method somewhere else

I don't feel we should do this refactoring...

danidoni: Refactoring is about making the next change simpler, it's a small change and it will make everything simpler and easier to understand. I can do it for you if you don't feel like changing it.

eduardoj: Let's try the refactoring in other pull request, if you don't mind.

src/api/app/services/rpmlint_log_extractor.rb
```diff
@@ -8,24 +8,38 @@ def initialize(attributes = {})
...
-    retrieve_rpmlint_log_from_log
+    time_start = Time.now.to_f
+    content = begin
+      # Retrieve the content of the 'rpmlint.log' file
+      Backend::Api::BuildResults::Binaries.rpmlint_log(project, package, repository, architecture)
+    rescue Backend::NotFoundError => e
+      @rpmlint_log_file_found = false
+
+      # Condition: Flipper.enabled?(...   : TODO: Remove this `unless` condition once request_show_redesign is rolled out
+      # Condition: e.summary.include?(... : Case of removed project, package, repository or architecture
```
danidoni, line 27: Are those comments valid?

eduardoj: Yes. Sorry, I don't get what do you mean with your question...

eduardoj: Yes. Sorry, I don't get what do you mean with your question...

danidoni: Well I don't understand those comments at all. I thought maybe it was something you forgot to remove...

eduardoj: Hm...

The comment that starts with `Condition: Flipper.enabled?(...` refers to the first part of the condition.

The comment that starts with `Condition: e.summary.include?(...` refers to the rest of the condition.

I hope this clarifies it.

danidoni: Ah! Now I get it, it's for the line below... Alright!

danidoni (approved):

PR #15503, opened 2024-01-18
Fix small cluttering

danidoni:
When there are excluded results, the text appears too close to the actual results. This PR adds a bit of spacing between those elements.

_Before_
Screenshot from 2024-01-18 10-24-58

_After_


saraycp (approved):
This whole view is going to disappear as soon as we finish the Build Status Monitor page. But it's not bad to have it nice until then.

PR #15504, opened 2024-01-18
Allow commenting on the same line after a comment was deleted from it

hellcp-work:
Depends on #15501 

After deleting a comment in the changes tab, there was no way to add a comment to the same line without reloading. This aims to fix the issue


rubhanazeem:
Delete comment is not working after this change 
image

danidoni (approved):

Issue #15104, opened 2023-10-24
[production][submissions#create] Mysql2::Error: Data too long for column 'payload' at row 1


danidoni:
[See this exception on Errbit](http://errbit.opensuse.org/apps/5620ca2bdc71fa00b1000000/problems/627f865a84bf460539e62c3d)
# Mysql2::Error: Data too long for column &#39;payload&#39; at row 1 #
### URL ###
[https://build.opensuse.org/projects/home:wkazubski:test:science/packages/fldigi/submissions](https://build.opensuse.org/projects/home:wkazubski:test:science/packages/fldigi/submissions)"
### Where ###
submissions#create

### Occured ###
May 14 10:37am

### Similar ###
24

## Params ##
~~~
{
  &quot;action&quot; =&gt; &quot;create&quot;,
  &quot;add_revision&quot; =&gt; &quot;yes&quot;,
  &quot;authenticity_token&quot; =&gt; &quot;9W1oCp0X6R0u0o6WYHGdJekwrBk95Bd0qnvgmfFg2GgLSvLRXFKnMp7eg07XxIwqoFjAMNxXadrgjmUT0O1LBA==&quot;,
  &quot;bs_request&quot; =&gt; {
    &quot;bs_request_actions_attributes&quot; =&gt; {
      &quot;0&quot; =&gt; {
        &quot;source_package&quot; =&gt; &quot;fldigi&quot;,
        &quot;source_project&quot; =&gt; &quot;home:wkazubski:test:science&quot;,
        &quot;source_rev&quot; =&gt; &quot;5&quot;,
        &quot;sourceupdate&quot; =&gt; &quot;cleanup&quot;,
        &quot;target_package&quot; =&gt; &quot;fldigi&quot;,
        &quot;target_project&quot; =&gt; &quot;hardware:sdr&quot;,
        &quot;type&quot; =&gt; &quot;submit&quot;
      }
    },
[...]

PR #15513, opened 2024-01-19
change events payload column size from TEXT to MEDIUMTEXT

FiloSpaTeam:
Hey Friends :)

Basically this solves the #15104 and may others issues related to notify system. 

**How to reproduce**
When a description hit the char limit (65535 ~ 64kb), the event payload that has the same limit of the description, receive instead a modified payload (description + event_parameters) making the after_create notify crash due to oversized payload (> 65535 chars).
Try for example with the submissions_controller.rb and a description that is very very long (like 65535 chars).

This PR introduces a new migration that change the `payload` column size for the `events` table from TEXT to MEDIUMTEXT.
Actually i didn't provide a `self.down` because a rollback for the column size would result in a DATALOSS

Cheers <3

### DIFF
--- a/src/api/db/migrate/20240119114738_change_events_payload_from_text_to_mediumtext.rb
+++ b/src/api/db/migrate/20240119114738_change_events_payload_from_text_to_mediumtext.rb
@@ -1,0 +1,11 @@
+class ChangeEventsPayloadFromTextToMediumtext < ActiveRecord::Migration[7.0]
+  def self.up
+    safety_assured { change_column :events, :payload, :text, limit: 16.megabytes - 1 }
+  end
+
+  def self.down
+    # NOTE: change back to TEXT could cut some payloads that could be more than 65535 characters.
+    # From TEXT to MEDIUMTEXT we could store from ~64kb to ~16Mb of data but basically nothing has
+    # a payload more than ~64kb so the event payload nearly should be always around 64kb.
+  end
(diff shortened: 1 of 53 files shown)


src/api/db/migrate/20240119114738_change_events_payload_from_text_to_mediumtext.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeEventsPayloadFromTextToMediumtext < ActiveRecord::Migration[7.0]
+  def self.up
+    safety_assured { change_column :events, :payload, :text, limit: 16.megabytes - 1 }
```
danidoni, line 3: I think this triggered a [Strong Migrations](https://github.com/ankane/strong_migrations) warning, right? We may need to run this migration during [our maintenance window](https://github.com/openSUSE/ansible-obs#deployment-with-migration-with-downtime)

FiloSpaTeam: yeah

krauselukas:
DO NOT MERGE: We have to wait for the maintenance window for this migration

rubhanazeem:
@FiloSpaTeam I had to open a new PR to address conflicts in this branch. Your changes have already been deployed here #15649. Thank you!

PR #15520, opened 2024-01-22
Specs for diff comments on request timeline

rubhanazeem:
This PR covers specs to check the behavior of CRUD diff comments on the request timeline. Follow-up of #15501


krauselukas:
@rubhanazeem I would include the commit with the specs in https://github.com/openSUSE/open-build-service/pull/15501

rubhanazeem:
> @rubhanazeem I would include the commit with the specs in #15501

Yes, I've pushed it to this PR

PR #15523, opened 2024-01-22
Make the footer contrast better

hellcp-work:
Screenshot from 2024-01-22 13-11-35


dmarcoux:
I had a look in the review app and somehow, the spacing between the notifications and the color scheme is quite small. I'm on Firefox 120.0.

example

Beside this, I wonder where we could put this on mobile. Right now, it's only displayed for devices with bigger viewports.

dmarcoux:
> > Beside this, I wonder where we could put this on mobile. Right now, it's only displayed for devices with bigger viewports.
> 
> I have not found a good place this far

Me neither, we don't have enough room on mobile. So it would have to be in another menu, but it doesn't fit in _Actions_ or _Places_. We don't have to solve this now.

eduardoj:
Shouldn't we show the "color scheme toggle" in the top bar only if we have the "Color themes" beta feature enabled?

hennevogel:
Please put it wherever else people configure their accounts which is not he top level links. How often do you think people will change this? :)

src/api/app/assets/stylesheets/webui/layout.scss
```diff
@@ -112,17 +112,17 @@
 
 #footer, #footer-legal {
   & a {
-    color: $gray-700;
-    &:hover { color: $gray-900; }
+    color: var(--bs-secondary-color);
+    &:hover { color: var(--bs-body-color); }
   }
 }
```
eduardoj, line 118: I have just realized, that we redefined here the styles of the `#footer `element defined in the `sticky-footer.scss` file: https://github.com/openSUSE/open-build-service/blob/9d0b803b6dfbdd80811c92b503c649db6e48a342/src/api/app/assets/stylesheets/webui/sticky-footer.scss#L9-L32

 Could be this PR followed by another PR refactoring those classes, at least?

eduardoj (approved):

PR #15505, opened 2024-01-18
Fix broken YAML multi-line documentation strings

mfriedrichgh:
The current OpenAPI schema (https://api.opensuse.org/apidocs/OBS-v2.10.50.yaml) doesn't contain valid YAML on multi-line description strings. 
This PR changes the `>` to `|` which results in valid YAML multi-line strings when building the documentation.

## Before
```yaml
...
  "/architectures":
    get:
      summary: List all known architectures.
      description: 'Get a list of all known architectures known to OBS in general.
        This is not the list of architectures provided by this instance. Check the
        schedulers element from the `/configuration` route for this.

'
      security:
      - basic_authentication: []
...
```
## After
```yaml
...
  "/architectures":
    get:
      summary: List all known architectures.
      description: |
        Get a list of all known architectures known to OBS in general.
        This is not the list of architectures provided by this instance. Check the
        schedulers element from the `/configuration` route for this.
      security:
      - basic_authentication: []
...
```

PR #15530, opened 2024-01-23
Fix broken single line documentation strings

mfriedrichgh:
This is related to PR #15505.
This PR addresses another issue found with the `|` string literal symbol:
If used on single line strings, the formatter of the documentation generator will cause otherwise legal YAML strings to become invalid by moving terminating quotes to the very start of a new line. Since these strings aren't multi-line in the first place, we should be able to simply move them into the same line.

## Before
```yaml
...
      responses:
        '200':
          description: 'OK. The request has succeeded.

'
          content:
...
```
## After
```yaml
...
      responses:
        '200':
          description: OK. The request has succeeded.
          content:
...
```

### DIFF
--- a/src/api/public/apidocs/paths/build_project_name_repository_name_architecture_name_package_name_file_name_view_fileinfo.yaml
+++ b/src/api/public/apidocs/paths/build_project_name_repository_name_architecture_name_package_name_file_name_view_fileinfo.yaml
@@ -1,6 +1,6 @@
 get:
   summary: This endpoint returns details about an specific artifact
-  description: |
+  description: 
     Given a project, repository, architecture and package, retrieve the given artifact's details.
   security:
     - basic_authentication: []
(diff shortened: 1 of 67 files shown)


src/api/public/apidocs/paths/build_project_name_repository_name_architecture_name_package_name_file_name_view_fileinfo.yaml
```diff
@@ -1,6 +1,6 @@
 get:
   summary: This endpoint returns details about an specific artifact
-  description: |
+  description: 
     Given a project, repository, architecture and package, retrieve the given artifact's details.
```
eduardoj, line 4:
```suggestion
  description: Given a project, repository, architecture and package, retrieve the given artifact's details.
```
This was not moved up...

eduardoj (changes requested):

eduardoj (approved):

PR #15540, opened 2024-01-24
[backend] change default compression for createrepo to gz

M0ses:
According to the documentation, the correct option for createrepo_c to compress also primary/filelists/other xml files with gz compression is `--general-compress-type=gz`

Taken from `createrepo_c --help`:

```
--general-compress-type=COMPRESSION_TYPE     Which compression type to use (even for primary, filelists and other xml).
```

### DIFF
--- a/dist/functions.setup-appliance.sh
+++ b/dist/functions.setup-appliance.sh
@@ -155,7 +155,8 @@ function get_hostname {
     done
     if [ -z "$FQHOSTNAME" -o "$FQHOSTNAME" = "localhost" ] && [ -x "$(command -v hostnamectl)" ];then
       FQHOSTNAME=`hostnamectl --static 2>/dev/null`
-    elif [ -s /etc/hostname ];then
+    fi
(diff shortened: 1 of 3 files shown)


M0ses:
Please only review - I will merge myself after some proper testing

adrianschroeter:
good finding, the minitest is most likely failing because we do not have a new enough createrepo/modifyrepo in OSU.

adrianschroeter (approved):

adrianschroeter:
modifyrepo only knows compress-type, so we must handling the parameters different now

adrianschroeter (changes requested):
modifyrepo parameters are different now as createrepo ones. --compress-type must stay there.

You should maybe also require a recent version of createrepo in the obs-server package now.

M0ses:
> modifyrepo parameters are different now as createrepo ones. --compress-type must stay there.

Next commit should take care

> You should maybe also require a recent version of createrepo in the obs-server package now.

`--general-compression-type` is part of createrepo_c since version 0.9, so I don`t see any need to require a specific version.

adrianschroeter (approved):

PR #15541, opened 2024-01-24
Add link to create a New Image to the main page

eduardoj:
Some users aren't aware OBS can be used to create images. They only see this link when they enter a project page.

Screenshot from 2024-01-24 16-31-05_with_box


danidoni:
Is this something you detected from a private conversation, the IRC channel...? Just to add a bit of context where this comes from ...

ncounter (approved):

PR #15542, opened 2024-01-24
List all comments for a specific user

danidoni:
We're rendering the comments in a new tab under the user profile, alongside the involvement and the activity.

image

### DIFF
--- a/src/api/app/components/bs_request_comment_component.html.haml
+++ b/src/api/app/components/bs_request_comment_component.html.haml
@@ -1,6 +1,7 @@
 .d-flex.mt-3.align-items-start
-  %i{ title: 'Comment', class: "fas fa-lg fa-comment #{ level > 1 ? 'd-none' : '' }" }
-  = render(AvatarComponent.new(name: comment.user.name, email: comment.user.email, size: 35, shape: :circle, custom_css: 'avatars-counter'))
+  - unless @nodecoration
+    %i{ title: 'Comment', class: "fas fa-lg fa-comment #{ level > 1 ? 'd-none' : '' }" }
+    = render(AvatarComponent.new(name: comment.user.name, email: comment.user.email, size: 35, shape: :circle, custom_css: 'avatars-counter'))
   .timeline-item-comment.ms-0.flex-fill.overflow-auto
     .comment-bubble.ms-3
       - if policy(comment).update? || policy(comment).destroy?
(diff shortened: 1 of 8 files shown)


src/api/app/views/webui/user/_comments.html.haml
```diff
@@ -0,0 +1,4 @@
+.card.mb-3
+  .card-body
+    - comments.each do |comment|
+      = render BsRequestCommentComponent.new(comment: comment, level: 1, commentable: comment.commentable, nodecoration: true)
```
danidoni, line 4: I wanted to refactor this component and use only the part that renders the comment bubble, but it's rapidly becoming a rabbit hole, so I'll do it in a follow up

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -1,6 +1,7 @@
 .d-flex.mt-3.align-items-start
-  %i{ title: 'Comment', class: "fas fa-lg fa-comment #{ level > 1 ? 'd-none' : '' }" }
-  = render(AvatarComponent.new(name: comment.user.name, email: comment.user.email, size: 35, shape: :circle, custom_css: 'avatars-counter'))
+  - unless @nodecoration
```
danidoni, line 2: I don't like the idea of customizing view components using specific corner-case parameters, and the View Components documentation states that [using inheritance](https://viewcomponent.org/viewcomponents-at-github.html#avoid-inheritance) (to develop a version of this View Component for a specific case) is not recommended, so any other ideas would be very welcome...

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -1,6 +1,7 @@
...
+    %i{ title: 'Comment', class: "fas fa-lg fa-comment #{ level > 1 ? 'd-none' : '' }" }
+    = render(AvatarComponent.new(name: comment.user.name, email: comment.user.email, size: 35, shape: :circle, custom_css: 'avatars-counter'))
   .timeline-item-comment.ms-0.flex-fill.overflow-auto
     .comment-bubble.ms-3
```
danidoni, line 6: I was playing with the idea of extracting the "comment bubble" in #15553

src/api/app/views/webui/user/_comments.html.haml
rubhanazeem: We really need to do something about this partial. This one place is used for all the comments (bs_request/request timeline/project/packages). Some of these places have different use-case and can't keep using the same partial everywhere

src/api/app/components/bs_request_comment_component.html.haml
rubhanazeem: What is the purpose of this `@nodecoration`?

src/api/app/views/webui/user/_comments.html.haml
danidoni: Let's see if I can do something with https://github.com/openSUSE/open-build-service/pull/15542#pullrequestreview-1853186851

src/api/app/components/bs_request_comment_component.html.haml
danidoni: The `@nodecoration` removes the user avatar and the username, because we already know who the user is. That way we can reclaim some space

danidoni: This are the things that are removed 
image

rubhanazeem (approved):

PR #15547, opened 2024-01-25
[api] Fix crash on package removal

adrianschroeter:
Any token still might exist and they need to be cleaned up. Otherwise we run into a database constraint error.

Discovered by osc test suite.


adrianschroeter:
@dmach this should fix your issue, but it is untested atm.

hennevogel (approved):
review by commit

PR #15548, opened 2024-01-25
Fix colouring for footer text

mdeniz:
Before it was coloured based on the light theme but we need it to adapt to each theme. On the dark theme the text was too dark and made difficult to read.

Before:
image

After:
image


mdeniz:
I started doing the same... but I realized that there's no need to force those exact colors, if we remove them from the SCSS it will go with the same values coming from the theme

mdeniz:
Ok, no worries then. I will close this one, happy to see the changes on your PR whenever ready :+1:

PR #15552, opened 2024-01-26
Improve event email for reportable

ncounter:
In case the `reportable` got already deleted, in the email for `Appeal|Cleared Decision|Favored Decision` event two issues occurres:
- the subject of the email does not contain the `reportable type` due to the lack of the reported object
- the link to the reported object is not available (#15551) but there is no clue about the fact the reported object is gone

This PR addresses those two points from above:
- fetch the `reportable type` from the event `payload_key` value instead
- add a simple text message in the email body about the missing reported object

### DIFF
--- a/src/api/spec/mailers/event_mailer_spec.rb
+++ b/src/api/spec/mailers/event_mailer_spec.rb
@@ -585,5 +585,50 @@ RSpec.describe EventMailer, :vcr do
         expect(mail.body.encoded).to include("<a href=\"https://build.example.com/project/show/#{project}#comments-list\">#{project}</a>")
       end
     end
+
+    context 'for an event of type Event::AppealCreated of a no longer existing reportable object' do
+      let(:moderator) { create(:admin_user) }
+      let(:offender) { create(:confirmed_user, login: 'offender') } # user who wrote the offensive comment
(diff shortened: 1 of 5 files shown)


src/api/spec/mailers/event_mailer_spec.rb
```diff
@@ -585,5 +585,50 @@
...
+      end
+
+      it 'contains the correct text' do
+        expect(mail.body.encoded).to include("'#{appellant}' decided to appeal to the decision '#{decision.reason}'. This is the reason:")
+        expect(mail.body.encoded).to include('I strongly disagree!')
+      end
+
+      it 'renders link to the page of the comment' do
+        expect(mail.body.encoded).to include("The reported #{decision.reports.first.reportable&.class&.name&.downcase} does not exist anymore.")
+      end
```
eduardoj, line 631: I would change the `it` description, given that a link is not rendered.

ncounter: ups, good catch, thanks :+1:

ncounter: done

eduardoj (approved):

PR #15553, opened 2024-01-26
Extract a CommentBubbleComponent

danidoni:


src/api/app/components/comment_bubble_component.html.haml
```diff
@@ -0,0 +1,63 @@
...
+                        class: 'dropdown-item', remote: true, form_class: 'moderate-form', params: { moderation_state: !comment.moderated? })
+        - if policy(comment).destroy?
+          = link_to('#', data: { 'bs-toggle': 'modal',
+                                'bs-target': '#delete-comment-modal',
+                                confirmation_text: 'Please confirm deletion of comment',
+                                action: comment_path(comment) },
+                        class: 'dropdown-item delete_link', title: 'Delete comment') do
+            Delete
+  .px-3.py-2.border-bottom
+    - unless hideusername
```
danidoni, line 23: I think this can be solved in a more elegant manner with [template variants](https://viewcomponent.org/guide/templates.html#templates)

danidoni:
This is not the way

PR #15559, opened 2024-01-29
Show user state on user's page, only for admin users

eduardoj:
Before:

Screenshot from 2024-01-29 14-19-06

After:

Screenshot from 2024-01-29 14-18-42

### DIFF
--- a/src/api/app/views/webui/user/_basic_info.html.haml
+++ b/src/api/app/views/webui/user/_basic_info.html.haml
@@ -35,7 +35,7 @@
     .col-8.col-sm-10.col-md-12.ps-0.p-md-3
       %h4#home-realname
         = user.realname
-      %h5.text-muted#home-login
+      %h5.text-reset#home-login
         = user.login
       - if User.possibly_nobody.is_admin?
         .text-reset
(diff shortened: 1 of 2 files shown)


danidoni:
What about using [badges](https://getbootstrap.com/docs/5.3/components/badge/) for the user state?

eduardoj:
> What about using [badges](https://getbootstrap.com/docs/5.3/components/badge/) for the user state?

I thought they could be mistaken with the user roles. So I didn't use them.

danidoni:
> I thought they could be mistaken with the user roles. So I didn't use them.

But it says "State: " so it's kind of very clear what this is about, so I think it won't be confusing...

eduardoj:
> But it says "State: " so it's kind of very clear what this is about, so I think it won't be confusing...

I particularly don't find any special benefit from using a badge for showing the value of a field when this field will have only one value. Badges could be used to show one value, but tend to be used to show multiple values. The page will also look more loaded, without that being need. This will be also a field that only administrators will see.

I still see more reasons against using a badge than in favor of using it. I hope I convinced you...

danidoni:
> I still see more reasons against using a badge than in favor of using it. I hope I convinced you...

Alright, let it be like it is now and we can improve it later

danidoni (approved):

PR #15560, opened 2024-01-29
Point out which commentable is locked in the locked alert

rubhanazeem:
Prior to this PR, the comment lock alert on the package page showed the wrong message `You can remove the lock by clicking on the button below` even when a package is not locked. This PR fixes this message and points out which commendable is locked.

Notes for test: 
1. lock a project
2. go into a package in the project
3. look at the comments
4. the alert should say the project is locked

### DIFF
--- a/src/api/app/helpers/webui/comment_locks_helper.rb
+++ b/src/api/app/helpers/webui/comment_locks_helper.rb
@@ -1,0 +1,14 @@
+module Webui::CommentLocksHelper
+  def comment_lock_alert(commentable)
+    alert = 'You can remove the lock by clicking on the button below.'
+    return alert if commentable.comment_lock
+
+    if commentable.is_a?(Package) && commentable.project.comment_lock
+      text = 'You can remove the lock by visiting'
+      link = link_to(commentable.project.name, project_show_path(commentable.project))
+      safe_join([text, ' ', content_tag(:a, link)])
+    else
+      alert
+    end
+  end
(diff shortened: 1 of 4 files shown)


src/api/app/helpers/webui/comment_locks_helper.rb
```diff
@@ -0,0 +1,14 @@
+module Webui::CommentLocksHelper
+  # rubocop:disable Rails/OutputSafety
+  def comment_lock_alert(commentable)
+    alert = 'You can remove the lock by clicking on the button below.'
+    return alert if commentable.comment_lock
+
+    if commentable.is_a?(Package) && commentable.project.comment_lock
+      "You can remove the lock by visiting #{link_to(commentable.project.name, project_show_path(commentable.project))}.".html_safe
```
danidoni, line 8:
```suggestion
      message = "You can remove the lock by visiting ".html_safe
      message.concat(link_to(commentable.project.name, project_show_path(commentable.project)))
```
I think you can do it like this and avoid hitting the Rails/OutputSafety cop (seen [here](https://www.rubydoc.info/gems/rubocop-rails/2.21.1/RuboCop/Cop/Rails/OutputSafety))

rubhanazeem: Thank you for looking into this. I've addressed it

danidoni (approved):

PR #15564, opened 2024-01-29
Don't show delete links for deleted users in admin pages

eduardoj:
Related to: #15559.

### Before

Screenshot from 2024-01-29 15-32-51

For a deleted user:

Screenshot from 2024-01-29 15-35-03

### After

Screenshot from 2024-01-29 15-33-08

For a deleted user:

Screenshot from 2024-01-29 15-34-52

### DIFF
--- a/src/api/app/helpers/webui/user_helper.rb
+++ b/src/api/app/helpers/webui/user_helper.rb
@@ -25,9 +25,11 @@ module Webui::UserHelper
         mail_to(user.email) do
           tag.i(nil, class: 'fas fa-envelope text-secondary pe-1', title: 'Send Email to User')
         end,
-        link_to('#', title: 'Delete User', data: { 'bs-toggle': 'modal',
-                                                   'bs-target': '#delete-user-modal', 'user-login': user.login, action: user_path(user.login) }) do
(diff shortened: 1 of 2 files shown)


src/api/app/helpers/webui/user_helper.rb
```diff
@@ -25,9 +25,13 @@ def user_actions(user)
         mail_to(user.email) do
           tag.i(nil, class: 'fas fa-envelope text-secondary pe-1', title: 'Send Email to User')
         end,
-        link_to('#', title: 'Delete User', data: { 'bs-toggle': 'modal',
-                                                   'bs-target': '#delete-user-modal', 'user-login': user.login, action: user_path(user.login) }) do
-          tag.i(nil, class: 'fas fa-times-circle text-danger pe-1')
+        if user.state == 'deleted'
+          ''
+        else
```
ncounter, line 30: I think you can omit this slice and do just `if user.state != 'delete'`

eduardoj: > I think you can omit this slice and do just `if user.state != 'delete'`

Yes, it works. I only had to invert the logic of the conditional, to pass RuboCop. Thanks!

ncounter (approved):

PR #15570, opened 2024-01-30
Show the user roles alongside the comment

danidoni:
In order to provide more context to who wrote a comment, we want to render the user role also.

This is how it looks like in the comments on a request:
image

And this is how it looks like in the comments on a package for example:
image

## Tips for reviewing:
1. Go to https://obs-reviewlab.opensuse.org/danidoni-label-user-roles-in-comments/
2. Log in as Iggy, password 'opensuse'
3. Visit https://obs-reviewlab.opensuse.org/danidoni-label-user-roles-in-comments/request/show/1 and check how the badges look like

### DIFF
--- a/src/api/app/components/bs_request_comment_component.html.haml
+++ b/src/api/app/components/bs_request_comment_component.html.haml
@@ -24,12 +24,17 @@
                                     action: comment_path(comment) },
                             class: 'dropdown-item delete_link', title: 'Delete comment') do
                 Delete
-      .px-3.py-2.border-bottom
-        - unless @nodecoration
-          = link_to(helpers.realname_with_login(comment.user), user_path(comment.user))
-          \-
-        = link_to_comment(comment)
-        = render CommentHistoryComponent.new(comment)
+      .px-3.py-2.border-bottom.d-flex.flex-column.flex-lg-row.justify-content-between.align-items-start
+        .d-inline-flex.justify-content-start
+          - unless @nodecoration
+            = link_to(helpers.realname_with_login(comment.user), user_path(comment.user))
+            \-
+          = link_to_comment(comment)
+          = render CommentHistoryComponent.new(comment)
+        .d-inline-flex.justify-content-end
+          - helpers.comment_user_role_titles(comment).each do |title|
+            %span.badge.text-bg-secondary.mx-1
+              = title.upcase
       - if comment.diff_ref && diff
         - action = comment.commentable
         .comment-diff
--- a/src/api/app/components/bs_request_comment_component.html.haml
+++ b/src/api/app/components/bs_request_comment_component.html.haml
@@ -24,17 +24,17 @@
                                     action: comment_path(comment) },
                             class: 'dropdown-item delete_link', title: 'Delete comment') do
                 Delete
-      .px-3.py-2.border-bottom.d-flex.flex-column.flex-lg-row.justify-content-between.align-items-start
+      .px-3.py-2.border-bottom.d-flex.flex-column.flex-lg-row.justify-content-between.align-items-start.gap-2
         .d-inline-flex.justify-content-start
           - unless @nodecoration
             = link_to(helpers.realname_with_login(comment.user), user_path(comment.user))
             \-
           = link_to_comment(comment)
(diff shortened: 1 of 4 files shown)


src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -24,7 +24,7 @@
                                     action: comment_path(comment) },
                             class: 'dropdown-item delete_link', title: 'Delete comment') do
                 Delete
-      .px-3.py-2.border-bottom
+      .px-3.py-2.border-bottom.d-flex.flex-column.flex-lg-row.justify-content-between.align-items-start
```
danidoni, line 27: This makes the badges appear below the username and the timestamp for smaller screens

hennevogel:
Design wise I find the badges a bit too heavy. Can you make them smaller and lighter?

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -24,12 +24,17 @@
                                     action: comment_path(comment) },
                             class: 'dropdown-item delete_link', title: 'Delete comment') do
                 Delete
-      .px-3.py-2.border-bottom
-        - unless @nodecoration
-          = link_to(helpers.realname_with_login(comment.user), user_path(comment.user))
-          \-
-        = link_to_comment(comment)
-        = render CommentHistoryComponent.new(comment)
+      .px-3.py-2.border-bottom.d-flex.flex-column.flex-lg-row.justify-content-between.align-items-start
```
ncounter, line 27: I would use `align-items-center` to align vertically instead, because if one of the element has a bigger or smaller height, the alignment would not be to the center but to the top.

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -24,12 +24,17 @@
...
+        .d-inline-flex.justify-content-start
+          - unless @nodecoration
+            = link_to(helpers.realname_with_login(comment.user), user_path(comment.user))
+            \-
+          = link_to_comment(comment)
+          = render CommentHistoryComponent.new(comment)
+        .d-inline-flex.justify-content-end
```
ncounter, line 34: Adding a `small` class here will make the badges a bit smaller.

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -24,12 +24,17 @@
...
+          - helpers.comment_user_role_titles(comment).each do |title|
+            %span.badge.text-bg-secondary.mx-1
+              = title.upcase
```
ncounter, line 37: What about using just the capitalization? Having the whole word in uppercase feels a bit too strong

danidoni: I like it. Let me change it

src/api/app/components/bs_request_comment_component.html.haml
```diff
@@ -24,12 +24,17 @@
...
+      .px-3.py-2.border-bottom.d-flex.flex-column.flex-lg-row.justify-content-between.align-items-start
+        .d-inline-flex.justify-content-start
+          - unless @nodecoration
+            = link_to(helpers.realname_with_login(comment.user), user_path(comment.user))
+            \-
+          = link_to_comment(comment)
+          = render CommentHistoryComponent.new(comment)
+        .d-inline-flex.justify-content-end
+          - helpers.comment_user_role_titles(comment).each do |title|
+            %span.badge.text-bg-secondary.mx-1
```
ncounter, line 36: Unfortunately Bootstrap `badge` by default doesn't have an `outline` version of the color pattern, so we might want to create it by removing the `.text-bg-secondary` and adding `.border.border-secondary.text-secondary` to make it look the opposite of the current color style, like the following:

image

src/api/app/components/bs_request_comment_component.html.haml
danidoni: Done

src/api/app/components/bs_request_comment_component.html.haml
danidoni: Fixed

ncounter (approved):

PR #15573, opened 2024-01-31
Show icons for image templates using source paths instead of public paths

eduardoj:
Access to the Image Templates page is restricted to authenticated users.


hennevogel:
> Access to the Image Templates page is restricted to authenticated users.

Not really no? 

https://github.com/openSUSE/open-build-service/blob/3692aaffe07a1783276aab9a557de7bc792e9620/src/api/app/controllers/webui/image_templates_controller.rb#L1-L10

eduardoj:
> > Access to the Image Templates page is restricted to authenticated users.
> 
> Not really no?
> 
> https://github.com/openSUSE/open-build-service/blob/3692aaffe07a1783276aab9a557de7bc792e9620/src/api/app/controllers/webui/image_templates_controller.rb#L1-L10

Ups, that should be fixed, true.

In #15541 we made sure the link to the image templates page should be shown only for authenticated users.

eduardoj:
#15575 will address the authentication issue.

PR #15578, opened 2024-02-01
Add column `blocked_from_commenting` to table users

eduardoj:
This field will be used to define if a user is blocked from creating comments.

### DIFF
--- a/src/api/db/migrate/20240201082303_add_blocked_from_commenting_to_users.rb
+++ b/src/api/db/migrate/20240201082303_add_blocked_from_commenting_to_users.rb
@@ -1,0 +1,5 @@
+class AddBlockedFromCommentingToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :blocked_from_commenting, :boolean, default: false, null: false
+  end
+end
--- a/src/api/db/schema.rb
+++ b/src/api/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_01_11_231639) do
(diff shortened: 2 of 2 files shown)


src/api/db/migrate/20240201082303_add_blocked_commenting_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddBlockedCommentingToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :blocked_commenting, :boolean, default: false, null: false
```
danidoni, line 3: I believe it should be `commenting_blocked`

eduardoj: I prefered `blocked_commenting`  for consistency with other fields in the same table, where the adjective goes before the noun. See: https://github.com/openSUSE/open-build-service/blob/10295dfe69bfa4425a658ad1a4457bdc4ea8ce74/src/api/app/models/user.rb#L959-L961

I also tought about `blocked_from_commenting`, too...

Do you believe this is a must?

danidoni: `blocked_commenting` is something you did in the past: "I blocked commenting in the page". Its adjective+verb.

`deprecated_password` is adjective+noun,  If you want to follow the same rule, then it should be `blocked_comment`.

`blocked_from_commenting` is a state of things: "The table is blocked from commenting", so it works for me.

eduardoj: Good. I renamed it to `blocked_from_commenting`. Please take a look again.

PR #15582, opened 2024-02-01
[api] add some missing api parameters

adrianschroeter:
Reviewing all places where a CGI parameter can be used multiple times _and_ is exposed via the api.

Some of the /build/_workerstatus parameters are currently not support, but used to be in the past IIRC. However, not documented yet, but we may want to bring these back.

### DIFF
--- a/docs/api/api/api.txt
+++ b/docs/api/api/api.txt
@@ -505,6 +505,7 @@ Parameters:
 
   deleted: bool, show deleted package instead of existing, optional
   expand: bool, include also packages from linked projects, optional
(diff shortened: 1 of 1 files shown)


adrianschroeter:
for @eduardoj

adrianschroeter:
@eduardoj just scan for "multiple" string in entire file, some of them were in before.

eduardoj:
> @eduardoj just scan for "multiple" string in entire file, some of them were in before.

I'll check that. Thanks!

hennevogel (approved):

PR #15587, opened 2024-02-02
Don't fetch unneeded information for JSON requests

eduardoj:
The current announcement and the watchlist items are not needed to be fetched in JSON requests.


src/api/app/controllers/webui/webui_controller.rb
```diff
@@ -216,9 +216,15 @@ def current_announcement
   end
 
   def fetch_watchlist_items
-    @watched_requests = User.possibly_nobody.watched_requests
-    @watched_packages = User.possibly_nobody.watched_packages
-    @watched_projects = User.possibly_nobody.watched_projects
+    if request.xhr? && params[:action] != 'toggle_watched_item'
```
hennevogel, line 219: https://api.rubyonrails.org/classes/AbstractController/Base.html#method-i-action_name

hennevogel (approved):

PR #15603, opened 2024-02-06
List all comments for a specific user

danidoni:
We're rendering the comments in a new tab under the user profile, alongside the involvement and the activity.

See #15542 and #15596


src/api/app/views/webui/user/_involvement_and_activity.html.haml
```diff
@@ -3,35 +3,46 @@
...
+              = link_to('#comments', id: 'comments-tab', class: 'nav-link text-nowrap',
+                        data: { 'bs-toggle': 'tab' }, role: 'tab', aria: { controls: 'comments', selected: false }) do
+                Comments
 
   .card-body.p-0#involvement-and-activity
     = render partial: 'webui/user/involvement', locals: { involved_items_service: involved_items_service }
 
     -# Displaying tabs in medium-large viewports when the Contribution Graph is enabled
-    - if contribution_graph?
+    - if contribution_graph? || Flipper.enabled?(:content_moderation, User.session) && policy(Report.new(reportable: displayed_user)).create?
```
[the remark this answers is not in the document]
danidoni: > I'm not sure I understand the use of `policy(Report.new(reportable: displayed_user)).create?` here. Why does everyone that can create a report (logged in users which haven't submitted a report against that user yet) see the comments?

I see now that this reasoning is flawed... I it would be better that only Moderators could review it, right?

src/api/app/policies/user_policy.rb
```diff
@@ -10,4 +10,8 @@ def show?
   def check_watchlist?
     record.login == User.session!.login || User.admin_session?
   end
+
+  def comments?
```
krauselukas, line 14: @hellcp-work just curious, why shouldn't we display the comments to the user that owns them and is in the beta program?

krauselukas: and a suggestion, just `comments?` is a little confusing for me. By the name I dont really get what it is suppose to do, maybe `list_comments?`

src/api/app/controllers/webui/users_controller.rb
```diff
@@ -19,6 +19,7 @@ def index
   def show
     @groups = @displayed_user.groups
     @involved_items_service = UserService::Involved.new(user: @displayed_user, filters: extract_filter_params, page: params[:page])
+    @comments = paged_comments
```
krauselukas, line 22: @hellcp-work maybe we can already check the policy here, and only fetch the comments if we in the end really going to render them. Might save a query here and there.

krauselukas (approved):

PR #15614, opened 2024-02-06
Index blocking_from_commenting field in users table

eduardoj:
Queries for this field will be performed in new users' measurements. See #15613.


danidoni (approved):
Fix the schema conflicts and you are good to go!

PR #15617, opened 2024-02-07
Add annotate comments after indexing state in users

eduardoj:
This could not be included in #15615 because of the checks we perform in `check_changed_files`.


danidoni (approved):

danidoni:
Resolve the conflicts and you are good to go!

PR #15622, opened 2024-02-07
Filter out reviews by staging project

### LINKED ISSUE #15471: Remove staging review from the review menu
**Problem**
Miss-clicking on the GUI and "approving" staging moves the package directly to GA. This leads ti a lot of complications and added effort. Simply put, The cost for clicks is low with a substantial probability, but the price of the repair is very high. 

**Solution description**
Forcing the package to GA is not "approving", and it should be removed from the menu. Move this functionality to a separate place/button.  

Completly removing this from GUI and relying on the command line interface. 

**Additional context**
Screenshot from 2024-01-11 15-07-10

ncounter:
Fix #15471


src/api/app/controllers/webui/request_controller.rb
```diff
@@ -573,6 +573,10 @@ def prepare_request_data
     handle_notification
   end
 
+  def review_by_staging_project(review)
+    review.by_project && Project.find_by(id: review.project_id).staging_workflow_id
+  end
```
eduardoj, line 578: I haven't checked, but it looks like we could use `staging_project?`...: https://github.com/openSUSE/open-build-service/blob/0e6498f3c6fe3d1ace26eb8f8f3a929689c24f7b/src/api/app/models/review.rb#L302-L304

ncounter: Nice catch, thanks. I changed to use it

PR #15630, opened 2024-02-09
Fix display of the bs_request monitor with more architectures

hellcp-work:
Displays the monitor with a larger amount of architectures in a less broken way

### DIFF
--- a/src/api/app/components/build_results_monitor_component.html.haml
+++ b/src/api/app/components/build_results_monitor_component.html.haml
@@ -35,7 +35,7 @@
                         project_package_repository_binaries_path(project_name: project_name,
                                                                  package_name: package_name,
                                                                  repository_name: repository_name),
-                        title: "Binaries for #{repository_name}", class: 'build-result-name text-break')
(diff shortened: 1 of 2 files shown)


rubhanazeem (approved):

eduardoj (approved):

rubhanazeem:
This is failing some minitests

src/api/app/components/build_results_monitor_component.html.haml
```diff
@@ -35,7 +35,7 @@
                         project_package_repository_binaries_path(project_name: project_name,
                                                                  package_name: package_name,
                                                                  repository_name: repository_name),
-                        title: "Binaries for #{repository_name}", class: 'build-result-name text-break')
+                        title: "Binaries for #{repository_name}", class: 'build-result-name')
```
ncounter, line 38: We'll might meet a different corner case where the repo name is very long we might want to handle differently, but I think it is fine for now and we can wait until that happens, I don't have a real case in mind for now.

ncounter (approved):

PR #15641, opened 2024-02-13
Refactor package cmd router out of SourceController

hennevogel:
There are several API routes that have a `cmd` parameter that is used to to call different methods from one controller action. Instead of having multiple routes for multiple actions.

```
/source/<project>/<package>?cmd=diff -> SourceController#package_command
/source/<project>/<package>?cmd=release  -> SourceController#package_command
```

And `package_command` then acts as a router. If `cmd` is diff then we call `SourceController#package_command_diff`. If `cmd` is `release`, then `SourceController#package_command_release` is called.

All of this can also just be done in the Rails router...


hennevogel:
changing source branch....

PR #15652, opened 2024-02-15
[api] fix release of package where source is remote or scmsync'd

adrianschroeter:
This creates a temporary package object which is not stored. We may track remote package objects in future but for now this temporary object seems to be the way to go.


hennevogel:
Please let's do this once (#16226) and not in 100 different places...

adrianschroeter:
there are high chances that we may need this this or next week on build.opensuse.org as well.

I would need to merge this then and you can rework it later.

hennevogel:
handled differently in #16506

adrianschroeter:
still crashing without this

hennevogel:
> still crashing without this

What is still crashing and where?

PR #15656, opened 2024-02-15
Link request build summary chart to build result tab

krauselukas:
*How to test this:*

1. Create a request that includes a few build results
2. You can patch the following line if the amount of repositories is to low to show the chart summary https://github.com/openSUSE/open-build-service/blob/8fcea11ebc44bc9e0a8000cebdb85d509dd35010/src/api/app/components/chart_component.html.haml#L9
3. Click on the bars in the chart and see that it links to the build result tab and applies the corresponding repository and build state filters

*Note:*

There is currently no way to customize click events directly through Chartkick (and the underlying Chart.js library). I can also not generate the url's server side in ruby and provide them through the data hash that we pass to chartkick (additional keys are simply ignored and thrown away). So I have to generate the URL on the Javascript level and based on the data that is clicked on in the chart.     

Peek 2024-02-15 17-55

### DIFF
--- a/src/api/app/assets/javascripts/webui/request_show_redesign/chart_build_results.js
+++ b/src/api/app/assets/javascripts/webui/request_show_redesign/chart_build_results.js
@@ -13,3 +13,37 @@ function updateChartBuildResults() {
     }
   });
 }
+
+function linkBuildSummaryChartToBuildResultsTab() { // jshint ignore:line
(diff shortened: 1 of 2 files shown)


krauselukas:
[Build Results Graph to Details Link (M)](https://trello.com/c/94ihVERA/2414-build-results-graph-to-details-link-m)

danidoni (approved):
Works like a charm!

PR #15667, opened 2024-02-21
Authorize comment `#create` api endpoint

krauselukas:
Currently there is not authorization going on, in some cases the `CommentLockingValidator` still prevents comments from being created if permission is insufficient in certain cases. But some things are not handled, for example when a user is blocked for commenting.

This should be handled on the controller level using our pundit policy.

@a-chris thanks bringing this up

### DIFF
--- a/src/api/app/controllers/comments_controller.rb
+++ b/src/api/app/controllers/comments_controller.rb
@@ -8,7 +8,9 @@ class CommentsController < ApplicationController
   end
 
   def create
-    @obj.comments.create!(body: request.raw_post, user: User.session!, parent_id: params[:parent_id])
+    comment = @obj.comments.new(body: request.raw_post, user: User.session!, parent_id: params[:parent_id])
+    authorize comment, :create?
+    comment.save!
     render_ok
   end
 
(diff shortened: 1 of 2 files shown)


src/api/app/controllers/comments_controller.rb
```diff
@@ -8,7 +8,9 @@ def index
   end
 
   def create
-    @obj.comments.create!(body: request.raw_post, user: User.session!, parent_id: params[:parent_id])
+    comment = @obj.comments.new(body: request.raw_post, user: User.session!, parent_id: params[:parent_id])
+    authorize comment, :create?
```
danidoni, line 12: I believe we should test the policy `:create` method

krauselukas: I haven't checked if this policy is tested... just wanted to fix the endpoint. I will come up with a little test for this

danidoni: It is not. Maybe in another PR that this will depend on?

krauselukas: I create the specs in this PR @danidoni

danidoni (approved):

PR #15679, opened 2024-02-22
[api] relax package meta

adrianschroeter:
OBS is not always delivering title and description elements itself anymore (when they come from pure backend projects). So accept package meta also without it to avoid that tooling can not read and write same meta from OBS instances.

Also the order of title and description does not matter to us.

Rest is just indenting fixes.

### DIFF
--- a/docs/api/api/package.rng
+++ b/docs/api/api/package.rng
@@ -45,90 +45,108 @@
         <attribute name="project"/>
       </optional>
 
-      <element ns="" name="title">
-        <text/>
-      </element>
+      <interleave>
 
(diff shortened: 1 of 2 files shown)


hennevogel:
> OBS is not always delivering title and description elements itself anymore (when they come from pure backend projects). 

What are pure backend projects?

hennevogel:
> s,projects,packages,

So we are talking scmsync'ed projects?

adrianschroeter:
yes, also. But maybe also other remote packages in future of newer OBS instances.

hennevogel (approved):

PR #15683, opened 2024-02-23
Badge for user state

saraycp:
I used a discrete light background for the badge as using colorful ones made the area look overloaded.

Large viewports:

Screenshot 2024-02-23 at 17-40-59 Your Profile

Small viewports:

Screenshot 2024-02-23 at 17-41-30 Your Profile


ncounter:
Because we already have badges for roles, what do you think about not having another badge, and moving the state maybe above, right after the user image? Something that could look like the following ideally:

image

saraycp:
@ncounter can you have another look? New screenshots in the description

src/api/app/views/webui/user/_basic_info.html.haml
```diff
@@ -29,18 +29,17 @@
       = link_to('javascript:void(0);', id: 'toggle-in-place-editing', remote: true, title: 'Edit Your Account') do
         %i.fas.fa-user-edit
         %span.nav-item-name Edit
-  .row.mb-3.mb-md-0
+  .row.mb-3
     .col-4.col-sm-2.col-md-12.text-center
       = render(AvatarComponent.new(name: user.name, email: user.email, size: 200))
-    .col-8.col-sm-10.col-md-12.ps-0.p-md-3
+      - if User.possibly_nobody.is_admin?
+        .badge.text-bg-light.d-block.my-2
```
ncounter, line 36: I'd add `border` too just to give a bit of contrast

ncounter:
> @ncounter can you have another look? New screenshots in the description

Nice, I like it :+1:

danidoni:
> @danidoni these are the possible states `STATES = ['unconfirmed', 'confirmed', 'locked', 'deleted', 'subaccount']` At least I would display `locked` and `deleted`. What do you think?

I would skip only confirmed, looking what @ncounter propses, I think I like his idea better.

saraycp:
@ncounter @danidoni  your feedback is addressed:

- State not displayed when "Confirmed"
- State displayed as a wide "badge" right under the avatar.

ncounter (approved):

PR #15702, opened 2024-02-27
Move user update authorization to pundit

krauselukas:
This moves the authorization of updating a user to pundit and
separates the authorization logic for blocking users from commenting.

The reasoning is that it got a little confusing.
The previous authorization condition in the the case of submitting
the form to block a user from commenting, was checking for the presence
of the `:blocked_for_commenting` parameter and the that the user
is a moderator.

The problem is, you could potentially sneak in more parameters and
the authorization would still be truthy and therefore update other user
attributes. To prevent this from happening we strictly should only
toggle the `:blocked_from_commenting` column and ignore all other
attributes.


krauselukas:
[Move Webui::UserController#update auth code into a Pundit policy (M)](https://trello.com/c/XblOS04j/2419-move-webuiusercontrollerupdate-auth-code-into-a-pundit-policy-m)

danidoni (approved):

PR #15703, opened 2024-02-27
Create migration to add decision kind to canned responses

saraycp:
We'll need to group the canned responses by decision kind later on. The default is null for those canned responses not related to decisions.


danidoni (approved):

saraycp:
I move it to draft because I have to double check if it's worth to make it nil by default for non-decision comments.

PR #15706, opened 2024-02-28
Upgrade rack to 2.2.8.1

danidoni:
See #15688

### DIFF
--- a/dist/obs-bundled-gems.spec
+++ b/dist/obs-bundled-gems.spec
@@ -15,7 +15,7 @@
 # Please submit bugfixes or comments via http://bugs.opensuse.org/
 #
 
-%define rack_version 2.2.6.4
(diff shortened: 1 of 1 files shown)


danidoni:
The OBS checks are going to fail because we to remove the rack dependency on 2.2.6.4, which is what this PR is trying to do.

krauselukas (commented):
@danidoni shouldn't this happen together with the version bump in the Gemfile? (so basically the depfu PR)

danidoni:
> @danidoni shouldn't this happen together with the version bump in the Gemfile (so basically the Depfu PR)? Or is there a reason to do this separate?

I need this to make OBS builds to succeed, then I can go back to the referenced PR and rerun everything green.

danidoni:
Closing in favor of doing this step in the referenced PR

PR #15708, opened 2024-02-28
Show only relevant user's roles badges in comments

ncounter:
Roles that matter in a comment:

<table>
<thead>
<tr><th>Commentable Entity</td><th>User Roles</th></tr>
</thead>
<tbody>
<tr><td>Project</td><td>Maintainer</td></tr>
<tr><td>Package</td><td>Maintainer, Project Maintainer</td></tr>
<tr><td>BsRequest</td><td>Author, Reviewer, Source Maintainer, Target Maintainer</td></tr>
</tbody>
</table>

Note: due to the multiaction request option, there is no reason to distinguish if the Source|Target Maintainer is related to the project or the package because multiaction requests affect multiple packages, and conversation comments instead are attached to the global request.

Note 2: it might be questionable for _what about all the other roles?_  As long as they are not as relevant as the current ones in this PR, adding them all would result in an overflooded UI of badges and roles, and the value of this information would be hard to grasp, therefore lost

image
image
image


src/api/app/helpers/webui/comments_helper.rb
```diff
@@ -1,14 +1,20 @@
...
+    roles = []
 
-    roles += roles_for_request(comment) if comment.commentable.is_a?(BsRequest)
+    # binding.pry
+    case
+    when comment.commentable.is_a?(BsRequest)
+      roles = roles_for_request(comment)
+    when comment.commentable.is_a?(Project)
+      # roles.push('maintainer') if comment.user.can_modify?(comment.commentable)
+      roles.push('maintainer') if roles_for_project(comment).any?
```
danidoni, line 11: Is there any reason why you don't use `roles << 'maintainer'`?

src/api/app/helpers/webui/comments_helper.rb
```diff
@@ -1,14 +1,20 @@
...
   def comment_user_role_titles(comment)
-    roles = comment.user.roles.global.pluck(:title)
-
-    roles += roles_for_project(comment) if comment.commentable.is_a?(Project)
-
-    roles += roles_for_package(comment) if comment.commentable.is_a?(Package)
+    roles = []
 
-    roles += roles_for_request(comment) if comment.commentable.is_a?(BsRequest)
+    # binding.pry
```
danidoni, line 5: :thinking:

src/api/app/helpers/webui/comments_helper.rb
ncounter: I personally prefer the `push` format, it makes it visually more clear to me what is on the left and on the right of the `if` due to the `.push(...` concatenation:

`roles.push('maintainer') if roles_for_project(comment).any?`
VS
`roles << 'maintainer' if roles_for_project(comment).any?`

Is there any style preference for that or any other reason I am missing to prefer one or the other?

ncounter: I was missing the _speed_ diff, thanks for pointing it out. I'm adapting to that then :+1:

ncounter: Mmmm..do you have any pointers to read about the performance difference? I've been searching for it and I can only find the fact that `push` accept multiple and `<<` doesn't, and `<<` can have lower or higher precedence because it's an operator, `push` it's more _linear_ instead.

Ruby's doc doesn't mention anything about it [1]

[1] https://ruby-doc.org/3.2.2/Array.html#class-Array-label-Adding+Items+to+Arrays

danidoni: No, I don't. `<<` is an operator. `push` is a method. I had the impression that operators are faster than methods, but I could not find anything to back this up. So then my reasoning falls back to `<<` is the convention used everywhere when concatenating only one object.

src/api/app/helpers/webui/comments_helper.rb
```diff
@@ -27,21 +31,40 @@ def roles_for_package(comment)
...
-             .pluck('roles.title')
-
-      acc + Relationship
-            .joins(:role)
-            .joins(:package)
-            .where(packages: { name: action.target_package })
-            .where(user: comment.user)
-            .pluck('roles.title')
+    roles << 'author' if comment.commentable.creator == comment.user.login
+    roles << 'reviewer' if comment.commentable.reviews.pluck(:user_id).include?(comment.user.id) ||
```
danidoni, line 35: I think you can just do `comment.commentable.reviews.map(&:user).include(comment.user)`.

src/api/app/helpers/webui/comments_helper.rb
```diff
@@ -27,21 +31,40 @@ def roles_for_package(comment)
...
+                                  .where(projects: { name: action.target_project })
+                                  .where(user: comment.user)
+                                  .where('roles.title': 'maintainer')
+                                  .pluck('roles.title')
+                                  .union(Relationship.joins(:role).joins(:package).joins('LEFT JOIN projects ON projects.id = packages.project_id')
+                                                     .where(packages: { name: action.target_package })
+                                                     .where(projects: { name: action.target_project })
+                                                     .where(user: comment.user)
+                                                     .where('roles.title': 'maintainer')
+                                                     .pluck('roles.title'))
```
danidoni, line 62: All of this should go into `queries/relationships_finder.rb`

src/api/app/helpers/webui/comments_helper.rb
ncounter: done

src/api/app/helpers/webui/comments_helper.rb
ncounter: done

danidoni (approved):

PR #15712, opened 2024-02-29
Release notes for 2.10.22

danidoni:


danidoni:
The obs-server 15.4 is not present in the workflow configuration so there is a bug going on, so don't expect that check to go green soon

hennevogel:
> there is a bug going on,

The bug is that it's marked as required check on github. Required checks are expected to happen...

krauselukas (approved):

PR #15713, opened 2024-02-29
Add rake task to generate test data for bs_requests

rubhanazeem:
Run the task `rake dev:requests:copy_requests_from_opensuse_factory`
Before you run the task, please enter your username and password to `make_api_request` method.

This task involves copying 10 `new,review` requests from `openSUSE:Factory`. However, this implementation has certain limitations; it will disregard requests associated with staging projects and will solely focus on requests targeting `openSUSE:Factory`. Additionally, users are required to input login details to facilitate API requests.
 
TODO:
- [x] Setup interconnect
- [x] Copy openSUSE:Factory using interconnect including (proj configs and _meta)
- [x] Get list of 10 requests in state `new,review`
- [x] Copy source projects from requests including (proj configs and _meta)
- [x] Branch package from requests to `openSUSE:Factory` and `source projects`
- [x] Copy reviews and history elements
- [x] Change users login

### DIFF
--- a/src/api/lib/tasks/dev/requests.rake
+++ b/src/api/lib/tasks/dev/requests.rake
@@ -224,5 +224,57 @@ namespace :dev do
         puts "* Request with delete action #{request.number} has been created."
       end
     end
+
+    desc 'Copy 10 requests from openSUSE:Factory'
+    task requests_from_opensuse_factory: :development_environment do
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+
+      admin = User.get_default_admin
+      admin.run_as do
+        # Setup interconnect
+        remote_proj = Project.find_or_create_by(name: 'openSUSE.org', remoteurl: 'https://api.opensuse.org/public')
+        remote_proj.store
+        FetchRemoteDistributionsJob.perform_now
+
+        clone_project(project_name: 'openSUSE:Factory')
+      end
+    end
+
+    def clone_project(project_name:)
+      project = Project.find_or_create_by(name: project_name)
+      config = make_api_request(url: "#{base_api_url}/source/#{project.name}/_config")
+      clone_prj_configs(config: config, project: project, comment: "Cloned from #{project.name}")
+
+      request = make_api_request(url: "#{base_api_url}/source/#{project.name}/_meta")
+      request_data = Xmlhash.parse(request)
+    end
+
+    def clone_prj_configs(config:, comment:, project:)
+      project.config.save({ user: User.session!.login, comment: comment }, config)
+    end
+
+    def make_api_request(url:, params: {}, headers: { 'Content-Type' => 'application/xml' })
+      username = ''
+      password = ''
+      conn = Faraday.new(
+        url: url,
+        params: params,
+        headers: headers
+      )
+      conn.set_basic_auth(username, password)
+      request = conn.get
+
+      request.body
+    end
+
+    def base_api_url
+      'https://api.opensuse.org'
+    end
+
+    def print_message(message)
+      puts '=' * 50
+      puts message
+    end
   end
 end
--- a/src/api/lib/tasks/dev/requests.rake
+++ b/src/api/lib/tasks/dev/requests.rake
@@ -248,6 +248,129 @@ namespace :dev do
 
       request = make_api_request(url: "#{base_api_url}/source/#{project.name}/_meta")
       request_data = Xmlhash.parse(request)
+
+      create_users_and_groups_from_meta(request_data: request_data)
+      create_repositories_from_meta(request_data: request_data, target_project: project)
+      clone_meta(meta: request, comment: "Cloned _meta from #{project.name}", project: project)
+    end
+
+    def clone_meta(meta:, comment:, project:)
+      params = ActionController::Parameters.new({ meta: meta, comment: comment, project: project.name })
+      meta_validator = MetaControllerService::MetaXMLValidator.new(params)
+      meta_validator.call
+      updater = MetaControllerService::ProjectUpdater.new(project: project, request_data: meta_validator.request_data).call
+      print_message(updater.errors) if updater.errors
+    end
+
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/PerceivedComplexity
+    def create_repositories_from_meta(request_data:, target_project:)
+      [request_data['repository']].flatten.each do |repository|
+        if repository['releasetarget'] && repository['releasetarget']['project']
+          project = Project.find_or_create_by(name: repository['releasetarget']['project'])
+          proj_repo = project.repositories.find_or_initialize_by(name: repository['releasetarget']['repository'])
+        elsif repository['path'].is_a?(Array)
+          repository['path'].each do |path|
+            project = Project.find_or_create_by(name: path['project'])
+            proj_repo = project.repositories.find_or_initialize_by(name: path['repository'])
+
+            target_repo = repository['name']
+            target_repository = Repository.find_by_project_and_name(target_project.name, target_repo)
+            target_repository = create(:repository, project: target_project, name: target_repo) if target_repository.nil?
+            proj_repo.path_elements.find_or_initialize_by(link: target_repository)
+
+            if repository['arch'].is_a?(Array)
+              repository['arch'].each do |architecture|
+                proj_repo.repository_architectures.find_or_initialize_by(architecture: Architecture.find_by(name: architecture))
+              end
+            else
+              proj_repo.repository_architectures.find_or_initialize_by(architecture: Architecture.find_by(name: repository['arch']))
+            end
+            proj_repo.save!
+            project.store(comment: "Added #{proj_repo.name} repository")
+          end
+
+          next
+        elsif repository['path']
+          project = Project.find_or_create_by(name: repository['path']['project'])
+          proj_repo = project.repositories.find_or_initialize_by(name: repository['path']['repository'])
+        else
+          next
+        end
+        target_repo = repository['name']
+        target_repository = Repository.find_by_project_and_name(target_project.name, target_repo)
+        target_repository = create(:repository, project: target_project, name: target_repo) if target_repository.nil?
+        proj_repo.path_elements.find_or_initialize_by(link: target_repository)
+
+        if repository['arch'].is_a?(Array)
+          repository['arch'].each do |architecture|
+            proj_repo.repository_architectures.find_or_initialize_by(architecture: Architecture.find_by(name: architecture))
+          end
+        elsif repository['arch']
+          proj_repo.repository_architectures.find_or_initialize_by(architecture: Architecture.find_by(name: repository['arch']))
+        end
+
+        proj_repo.save!
+        project.store(comment: "Added #{proj_repo.name} repository")
+      end
+    end
+    # rubocop:enable Metrics/CyclomaticComplexity
+    # rubocop:enable Metrics/PerceivedComplexity
+
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/PerceivedComplexity
+    def create_users_and_groups_from_meta(request_data:)
+      # create users
+      if request_data['person'].present?
+        [request_data['person']].flatten.each do |person|
+          find_user(person['userid'])
+        end
+      end
+
+      return if request_data['group'].nil?
+
+      if request_data['group'].is_a?(Array)
+        request_data['group'].each do |group|
+          next if Group.find_by(title: group['groupid'])
+
+          create(:group, title: group['groupid'])
+        end
+      else
+        group = Group.find_by(title: request_data['group']['groupid'])
+
+        create(:group, title: request_data['group']['groupid']) if group.nil?
+      end
+    end
+    # rubocop:enable Metrics/CyclomaticComplexity
+    # rubocop:enable Metrics/PerceivedComplexity
+
+    def find_user(login)
+      user = User.find_by_login(login)
+      return user if user
+
+      create(:confirmed_user, login: login)
+    end
+
+    def find_group(title)
+      return if title.nil?
+
+      Group.find_or_create_by!(title: title)
+    end
+
+    def find_project(project_name)
+      return if project_name.nil?
+
+      Project.find_or_create_by(name: project_name)
+    end
+
+    def find_package(project_name, package_name)
+      return if project_name.nil? || package_name.nil?
+
+      package = Package.find_by_project_and_name(project_name, package_name)
+      return package if package
+
+      project = Project.find_or_create_by(name: project_name)
+      create(:package, name: package_name, project: project)
     end
 
     def clone_prj_configs(config:, comment:, project:)
--- a/src/api/lib/tasks/dev/requests.rake
+++ b/src/api/lib/tasks/dev/requests.rake
@@ -1,3 +1,5 @@
+BASE_API_URL = 'https://api.opensuse.org'.freeze
+
 namespace :dev do
   namespace :requests do
     # Run this task with: rails dev:requests:multiple_actions_request
@@ -242,7 +244,7 @@ namespace :dev do
         clone_project(project_name: 'openSUSE:Factory')
 
         # Get the list of requests from openSUSE:Factory
-        url = "#{base_api_url}/search/request"
+        url = "#{BASE_API_URL}/search/request"
         params = { match: "target/@project='openSUSE:Factory' and state/@name='review' and action/@type='submit'", project: 'openSUSE:Factory', limit: '10', withhistory: '1', withfullhistory: '1' }
         # Get the total number of submit requests
         temp_request = make_api_request(url: url, params: params.merge(limit: '1'))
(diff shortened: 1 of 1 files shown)


src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,357 @@ namespace :dev do
...
+      # Setup interconnect
+      remote_proj = Project.find_by_name('openSUSE.org')
+      if remote_proj.nil?
+        print_message 'Creating the remote project for interconnect'
+
+        remote_proj = RemoteProject.new(name: 'openSUSE.org',
+                                        remoteurl: 'https://api.opensuse.org/public',
+                                        title: 'Standard OBS instance at build.opensuse.org',
+                                        description: 'This instance delivers the default build targets for OBS.')
+        remote_proj.store
```
hennevogel, line 221:
```suggestion
        remote_proj = RemoteProject.find_or_create_by(name: 'openSUSE.org',
                                        remoteurl: 'https://api.opensuse.org/public')
        remote_proj.store
```

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,357 @@ namespace :dev do
...
+
+      bs_request.save!
+      print_message("Successfully created request##{bs_request.number}")
+
+      bs_request
+    end
+
+    def clone_project(project_name:, users: [])
+      project = Project.find_by(name: project_name)
+      project = create(:project, name: project_name) if project.nil?
```
hennevogel, line 346:
```suggestion
      project = Project.find_or_create_by(name: project_name)
```

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,357 @@ namespace :dev do
...
+        remote_proj.store
+        FetchRemoteDistributionsJob.perform_now
+      end
+
+      clone_project(project_name: 'openSUSE:Factory', users: users)
+
+      # Get the list of requests from openSUSE:Factory
+      print_message 'Getting the list of requests'
+      url = "#{base_api_url}/request"
+      params = { view: 'collection', project: 'openSUSE:Factory', limit: '10', states: 'new,review' }
```
hennevogel, line 230: Not sure if the "first 10" is the best decision. They are probably not very collaborated on.

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,357 @@ namespace :dev do
...
+      request = make_api_request(url: url, params: params)
+      requests_list = Xmlhash.parse(request)
+      print_message 'Successfully got the requests list'
+
+      requests_list['request'].each do |req|
+        # target project not openSUSE:Factory
+        next if req['action']['target']['project'] != 'openSUSE:Factory'
+        next if req['action']['target']['project'].match?('openSUSE:Factory:Staging')
+        next if req['action']['source']['project'].match?('openSUSE:Factory:Staging')
```
hennevogel, line 239: How would any of this be possible?

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,357 @@ namespace :dev do
...
+        package: source_package_name
+      }
+
+      branch_params[:target_project] = target_project
+
+      branched_package = BranchPackage.new(branch_params).branch
+      created_project_name = branched_package[:data][:targetproject]
+      created_package_name = branched_package[:data][:targetpackage]
+
+      Event::BranchCommand.create(project: source_project_name, package: source_package_name,
```
hennevogel, line 377: Hm, why would you do that?

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,357 @@ namespace :dev do
...
+        create(:group, title: request_data['group']['groupid']) if group.nil?
+      end
+    end
+    # rubocop:enable Metrics/CyclomaticComplexity
+    # rubocop:enable Metrics/PerceivedComplexity
+
+    def create_user(login:, users: [])
+      return if User.find_by(login: login).present?
+
+      users << create(:staff_user, login: login)
```
hennevogel, line 487: Why staff users? And why assign this to this empty array?

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,357 @@ namespace :dev do
...
+    end
+
+    def clone_project(project_name:, users: [])
+      project = Project.find_by(name: project_name)
+      project = create(:project, name: project_name) if project.nil?
+      print_message("#{project.name} successfully created")
+
+      print_message("Cloning prjconfigs for #{project.name}")
+      request = make_api_request(url: "#{base_api_url}/source/#{project.name}/_config")
+      clone_prj_configs(config: request, project: project, comment: "Cloned from #{project.name}")
```
hennevogel, line 351:
```suggestion
      config = make_api_request(url: "#{base_api_url}/source/#{project.name}/_config")
      clone_prj_configs(config: config, project: project, comment: "Cloned from #{project.name}")
```

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,357 @@ namespace :dev do
...
+
+      project = Project.find_by_name(project_name)
+      project = create(:project, name: project_name) if project.nil?
+
+      create(:package, name: package_name, project: project)
+    end
+
+    def clone_prj_configs(config:, comment:, project:)
+      params = ActionController::Parameters.new({ config: config, comment: comment })
+      ProjectConfigurationService::ProjectConfigurationUpdater.new(project, User.session!, params).call
```
hennevogel, line 528: This is a 🐄 💩 service. It's just doing `project.config.save({user: 'Admin', comment: comment}, config)`

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,357 @@ namespace :dev do
...
+      request_data = Xmlhash.parse(request)
+
+      create_users_and_groups_from_meta(request_data: request_data, users: users)
+      create_repositories_from_meta(request_data: request_data, target_project: project)
+      clone_meta(meta: request, comment: "Cloned _meta from #{project.name}", project: project)
+    end
+
+    def branch_package(source_project_name:, source_package_name:, target_project:)
+      package = Package.find_by_project_and_name(target_project, source_package_name)
+      package.destroy if package.present?
```
hennevogel, line 364: You can use `force: 1` as `BranchPackage` parameter to do this.

hennevogel (changes requested):

src/api/lib/tasks/dev/requests.rake
rubhanazeem: I'm intentionally not using `find_or_create_by` because some fields are required and `find_or_create_by` will break if I don't provide all the required fields. In this case this statement is already under a check to find existing remote project.

src/api/lib/tasks/dev/requests.rake
rubhanazeem: This requires modification. Currently, it only retrieves the oldest 10 requests in `review` state, which are dormant with minimal activity. I will explore methods to fetch the latest requests instead of the oldest ones.

src/api/lib/tasks/dev/requests.rake
rubhanazeem: This requires exploring some more request data from the API. This would even make this PR hard to review and would even delay this PR. For now, we have good enough data to use in dev env. If we really need this then doing it in a follow-up pr would be the best way to move forward because staging requests can't use the logic we have in this PR for handling normal bs_requests.

src/api/lib/tasks/dev/requests.rake
rubhanazeem: Ugh, thanks for pointing it out. It's removed

src/api/lib/tasks/dev/requests.rake
rubhanazeem: No specific reason. Just to make sure nothing breaks. Converted it to `confirmed_users`.
This empty users list was to keep track of the newly created users so they could be anonymized later on

src/api/lib/tasks/dev/requests.rake
rubhanazeem: Thanks for the suggestion. You are correct. I've updated it

src/api/lib/tasks/dev/requests.rake
rubhanazeem: True. It works

src/api/lib/tasks/dev/requests.rake
hennevogel: description and title are for sure not required...

src/api/lib/tasks/dev/requests.rake
hennevogel: 10 from the middle sounds good to me.

src/api/lib/tasks/dev/requests.rake
hennevogel: Ah that API returns also outgoing requests from `openSUSE:Factory` I see. That might mean you don't get *any* request that matches your criteria...

I guess you really want to use https://api.opensuse.org/apidocs/#/Search/get_search_request

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,333 @@ namespace :dev do
...
+        group = Group.find_by(title: request_data['group']['groupid'])
+
+        create(:group, title: request_data['group']['groupid']) if group.nil?
+      end
+    end
+    # rubocop:enable Metrics/CyclomaticComplexity
+    # rubocop:enable Metrics/PerceivedComplexity
+
+    def find_user(login, users = [])
+      return if login.nil?
```
hennevogel, line 471: I don't think this is possible...

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,333 @@ namespace :dev do
 
       puts "* Request with delete action #{request.number} has been created."
     end
+
+    desc 'Copy last 10 requests from openSUSE:Factory'
+    task requests_from_open_suse_factory: :development_environment do
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+
+      users = []
```
hennevogel, line 208: I don't like that you pass this "global" variable around 3 levels deep. Why are you doing this?

src/api/lib/tasks/dev/requests.rake
rubhanazeem: The API you mentioned also doesn't return what we are expecting. It just returns the "total number" of requests matched with criteria passed in parameters

src/api/lib/tasks/dev/requests.rake
rubhanazeem: They are required unfortunately https://github.com/openSUSE/open-build-service/blob/master/src/api/app/models/remote_project.rb#L3

src/api/lib/tasks/dev/requests.rake
rubhanazeem: This condition is to prevent executing this method because in the creation of the review, I'm using this method and a review can be by_user, by_project, by_package etc. That's why I have this check in case by_user is nil

src/api/lib/tasks/dev/requests.rake
rubhanazeem: It's removed. Not needed anymore

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,325 @@ namespace :dev do
 
       puts "* Request with delete action #{request.number} has been created."
     end
+
+    desc 'Copy last 10 requests from openSUSE:Factory'
+    task requests_from_open_suse_factory: :development_environment do
```
hennevogel, line 204: `opensuse` is one word

src/api/lib/tasks/dev/requests.rake
hennevogel: Not if you use `Project`...

src/api/lib/tasks/dev/requests.rake
hennevogel: To me it sounds like you should handle this there and not here...

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,325 @@ namespace :dev do
...
+      require 'factory_bot'
+      include FactoryBot::Syntax::Methods
+
+      admin = User.get_default_admin
+      User.session = admin
```
hennevogel, line 209: I'm not sure if rake tasks run in their own thread (probably?) but if you never reset this it stays around for at least all following rake tasks executed.

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,325 @@ namespace :dev do
...
+        remote_proj.store
+        FetchRemoteDistributionsJob.perform_now
+      end
+
+      clone_project(project_name: 'openSUSE:Factory')
+
+      # Get the list of requests from openSUSE:Factory
+      print_message 'Getting the list of requests'
+      url = "#{base_api_url}/search/request"
+      params = { match: "target/@project='openSUSE:Factory' and state/@name='review'", project: 'openSUSE:Factory', limit: '10', offset: '50', withhistory: '1', withfullhistory: '1' }
```
hennevogel, line 229: `offset: '50'` sounds like a magic number that can go wrong.

src/api/lib/tasks/dev/requests.rake
rubhanazeem: Okay, I've updated this

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,315 @@ namespace :dev do
...
+      create_users_and_groups_from_meta(request_data: request_data)
+      create_repositories_from_meta(request_data: request_data, target_project: project)
+      clone_meta(meta: request, comment: "Cloned _meta from #{project.name}", project: project)
+    end
+
+    def clone_meta(meta:, comment:, project:)
+      params = ActionController::Parameters.new({ meta: meta, comment: comment, project: project.name })
+      meta_validator = MetaControllerService::MetaXMLValidator.new(params)
+      meta_validator.call
+      updater = MetaControllerService::ProjectUpdater.new(project: project, request_data: meta_validator.request_data).call
```
hennevogel, line 371: Why do you validate this? You don't trust your own production instance? :)

hennevogel: Also, this will update the repositories you crafted in `create_repositories_from_meta` no?

src/api/lib/tasks/dev/requests.rake
```diff
@@ -199,5 +199,315 @@ namespace :dev do
...
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/PerceivedComplexity
+    def create_repositories_from_meta(request_data:, target_project:)
+      [request_data['repository']].flatten.each do |repository|
+        if repository['releasetarget'] && repository['releasetarget']['project']
+          project = Project.find_or_create_by(name: repository['releasetarget']['project'])
+          proj_repo = project.repositories.find_or_initialize_by(name: repository['releasetarget']['repository'])
+        elsif repository['path'].is_a?(Array)
+          repository['path'].each do |path|
+            project = Project.find_or_create_by(name: path['project'])
```
hennevogel, line 384: It makes little sense to create local projects for this. You should use the projects over the interconnect. So 

```
  <repository name="15.5">
    <path project="openSUSE:Leap:15.5" repository="standard"/>
    <arch>x86_64</arch>
  </repository>
```

becomes

```
  <repository name="15.5">
    <path project="openSUSE.org:openSUSE:Leap:15.5" repository="standard"/>
    <arch>x86_64</arch>
  </repository>
```

src/api/lib/tasks/dev/requests.rake
hennevogel: Ah is this the reason you validate this? To make sure the other methods went through okay?

src/api/lib/tasks/dev/requests.rake
```diff
@@ -224,5 +224,341 @@ namespace :dev do
...
+        params: params,
+        headers: headers
+      )
+      conn.set_basic_auth(username, password)
+      request = conn.get
+
+      request.body
+    end
+
+    def base_api_url
```
danidoni, line 555: This is a constant

src/api/lib/tasks/dev/requests.rake
```diff
@@ -224,5 +224,341 @@ namespace :dev do
...
+      'https://api.opensuse.org'
+    end
+
+    def print_message(message)
```
danidoni, line 559: You can remove this

rubhanazeem: I think we should keep it because it works like a logger to inform about newly created requests. Any reason why you would like to drop this?

danidoni: I think we can find out the generated data in the UI already. Having it as console output is just noise, in my opinion

danidoni (approved):

src/api/lib/tasks/dev/requests.rake
rubhanazeem: After discussion we have decided to keep this

danidoni (approved):

PR #15714, opened 2024-02-29
Drop stale user scopes

ncounter:
Drop stale user scopes


hennevogel:
define stale

ncounter:
> define stale

Something that was used before but it is no longer used now.

danidoni:
> > define stale
> 
> Something that was used before but it is no longer used now.

Unused?

hennevogel (approved):

PR #15726, opened 2024-03-01
[api] fix corner case in approver permission handling.

adrianschroeter:
An approve state in a request should only be taken into account when the operation is acctually a request approve.

But not on any other state change like decline, revoke or supersede

OBS-302


adrianschroeter:
we should have a test case for this ...

adrianschroeter:
added a test case now

src/api/app/models/bs_request_permission_check.rb
```diff
@@ -352,14 +352,19 @@ def set_permissions_for_action(action, new_state = nil)
     # meta data change shall also be allowed after freezing a project using force:
     ignore_lock = (new_state == 'declined') ||
                   (opts[:force] && action.action_type == :set_bugowner)
+    target_user = if new_state == 'accepted'
```
hennevogel, line 355: What is it you try to express with the name `target_user` here? The user assigned to this variable is the target of what exactly?

hennevogel: A maintainer in the target?

danidoni:
I think all this BsRequestPermissionCheck should be a policy

src/api/test/functional/request_controller_test.rb
```diff
@@ -857,6 +857,11 @@ def test_create_request_approve_and_review
     assert_response :success
     assert_xml_tag(tag: 'state', attributes: { name: 'review', approver: 'fred' })
 
+    # random guy must not be allowed to drop the request after approval
+    login_adrian
+    post "/request/#{id}?cmd=changestate&newstate=declined"
+    assert_response 403
+
```
danidoni, line 864: I think it's better to test this in the model, that is cheaper on the CI, instead of a feature spec, this instantiates the whole stack

adrianschroeter: well, it is anyway instantiate when running this test and we don't need to manually set this situation up a second time there. 
But feel free to create a unit test.

PR #15727, opened 2024-03-01
Add spec for the comment `history?` pundit policy 

krauselukas:
~~We have to make sure the policy returns false unless it is
a admin/staff/moderator user that attempts to access the history.
Otherwise, in case of a regular user it simply doesn't
match the condition and returns true in the next line if the comment
is not soft deleted.~~

The policy was missing specs.

### DIFF
--- a/src/api/spec/policies/comment_policy_spec.rb
+++ b/src/api/spec/policies/comment_policy_spec.rb
@@ -173,4 +173,26 @@ RSpec.describe CommentPolicy do
       end
     end
   end
+
+  permissions :history? do
+    let(:staff_user) { create(:staff_user) }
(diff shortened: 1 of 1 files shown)


bmwiedemann (approved):

krauselukas:
> > Hmm, so everyone suppose to see the history of a comment? not only admins/moderators/staff?
> 
> Unless the comment is deleted or moderated, yes

Alright, than I got that wrong :) well I will extract the specs then (after adapting them)

krauselukas:
@hellcp-work @bmwiedemann alright, extracted the spec and adapted it. Dropped the changes I made on the policy

PR #15739, opened 2024-03-04
Move racc to 1.7.2

rubhanazeem:
Nokogiri `1.16.2` resolves to racc `1.7.2` and it's causing issues in openQA


danidoni:
Shouldn't be this patched against 2.10?

hennevogel:
and which issues? :)

danidoni:
> and which issues? :)

I'm sure @rubhanazeem wants to fix [this](https://openqa.opensuse.org/tests/3984770#step/rspec_webui_tests/7), isn't it?

hennevogel:
> I'm sure @rubhanazeem wants to fix [this](https://openqa.opensuse.org/tests/3984770#step/rspec_webui_tests/7), isn't it?

Sure but what is *this*? :)

rubhanazeem:
The issue is bundler can't continue. This has been happening for a while since we have updated the Nokogiri version. 
@danidoni you are correct it should only be against 2.10. I thought openQA was also failing for unstable

danidoni:
> > Sure but what is _this_? :)
> 
> The issue is bundler can't continue. This has been happening for a while since we have updated the Nokogiri version. @danidoni you are correct it should only be against 2.10. I thought openQA was also failing for unstable

Then I think we need another PR targeting 2.10 only. Because this fix won't land on 2.10...

rubhanazeem:
> > The issue is bundler can't continue. This has been happening for a while since we have updated the Nokogiri version. @danidoni you are correct it should only be against 2.10. I thought openQA was also failing for unstable
> 
> Then I think we need another PR targeting 2.10 only. Because this fix won't land on 2.10...

Okay, I'll be creating a new PR for that. Closing this one

hennevogel:
> The issue is bundler can't continue

This is not the issue this is the result of some issue...

PR #15746, opened 2024-03-06
Split out the dependency method out of PackageController

hellcp-work:
This also moves the remaining views that only pertain to those views out of the `views/package` directory


src/api/app/controllers/webui/packages/binaries_controller.rb
```diff
@@ -60,6 +60,41 @@ def show
...
+        @dependant_repository = @dependant_project.repositories.find_by(name: @dependant_repository_name) if @dependant_project.remoteurl.blank?
+        return @dependant_repository if @dependant_repository
+
+        flash[:error] = "Repository '#{@dependant_repository_name}' is invalid."
+        redirect_back(fallback_location: project_show_path(project: @project.name))
+      end
+
+      def set_filename
+        # Ensure it really is just a file name, no '/..', etc.
+        @filename = File.basename(params[:binary_filename] || params[:filename])
```
eduardoj, line 95:
```suggestion
        @filename = File.basename(params[:filename])
```
I would not rename some ocurrences of the `filename` parameter to `binary_filename` in this commit. Or even include the renaming in this pull request.

Is it really needed this renaming to split out the `dependency` method?

eduardoj: I guess we should add a parameter constraint , for `binary_filename` like this one: https://github.com/openSUSE/open-build-service/blob/296c43015231da84fdfd24871f9b61b957b72505/src/api/app/lib/routes_helper/routes_constraints.rb#L6

src/api/app/lib/routes_helper/routes_constraints.rb
```diff
@@ -4,6 +4,7 @@ module RoutesConstraints
       arch: %r{[^/]*},
       binary: %r{[^/]*},
       filename: %r{[^/]*},
+      binary_filename: %r{[^/]*},
```
danidoni, line 7: Is this trying to match everything before the `/`?

danidoni (approved):

PR #15749, opened 2024-03-06
Autocorrect Style/WordArray RuboCop offenses

eduardoj:


eduardoj:
The isolated migrations check reports a failure. In this case, this is a refactoring change and it makes sense to have all the refactoring in the same commit.

codecov/patch also fails: `Failing after 1s — 82.81% of diff hit (target 86.02%) `. Ignoring because this refactoring shouldn't affect test code coverage.

Merging.

PR #15750, opened 2024-03-06
Restore binary controller tests

hellcp-work:
These tests fell through the cracks in #15160

### DIFF
--- a/src/api/spec/controllers/webui/packages/binaries_controller_spec.rb
+++ b/src/api/spec/controllers/webui/packages/binaries_controller_spec.rb
@@ -1,0 +1,119 @@
+require 'webmock/rspec'
+
+RSpec.describe Webui::Packages::BinariesController, :vcr do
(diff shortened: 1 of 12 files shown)


src/api/spec/controllers/webui/packages/binaries_controller_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+        # We want to use the backend path here
+        allow(Backend::Api::BuildResults::Binaries).to receive_messages(fileinfo_ext: fake_fileinfo, download_url_for_file: nil)
+      end
+
+      context 'and normal html request' do
+        before do
+          get :show, params: { package_name: toms_package, project_name: home_tom, repository_name: repo_for_home_tom, arch: 'x86_64', filename: 'filename.txt', format: :html }
+        end
+
+        it { expect(response).to have_http_status(:success) }
```
danidoni, line 105: Don't we check that it returns the url of the binary?

danidoni (approved):

PR #15751, opened 2024-03-06
Autocorrect Style/SymbolArray RuboCop offenses

eduardoj:
Depends on #15749.


danidoni (approved):

eduardoj:
The isolated migrations check reports a failure. In this case, this is a refactoring change and it makes sense to have all the refactoring in the same commit.

Merging.

PR #15753, opened 2024-03-07
Remove package_files method from Webui::PackageController

hellcp-work:
This method recreated a hash that already exists, so this PR just uses the hash directly

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -83,7 +83,7 @@ class Webui::PackageController < Webui::WebuiController
       authorize @current_notification, :update?, policy_class: NotificationPolicy
     end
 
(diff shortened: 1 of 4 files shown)


danidoni:
Where this hash is being created instead?

danidoni:
> > Where this hash is being created instead?
> 
> https://github.com/openSUSE/open-build-service/pull/15753/files#diff-bbc200e2cd3f375997c657b55d73038ffecc3ba009b022d402118f0c3187ac1aR349 https://github.com/openSUSE/open-build-service/pull/15753/files#diff-bbc200e2cd3f375997c657b55d73038ffecc3ba009b022d402118f0c3187ac1aR569

I don't understand, could those links be outdated?

danidoni (approved):

PR #15754, opened 2024-03-07
Use resources instead of resource for file controller actions

hellcp-work:
Changes all the actions to use resources routes, so that it aligns more with how it's used. In practice that means that in update action, there is a file name parameter. Depends on #15753

### DIFF
--- a/src/api/app/controllers/webui/packages/files_controller.rb
+++ b/src/api/app/controllers/webui/packages/files_controller.rb
@@ -3,6 +3,7 @@ module Webui
     class FilesController < Packages::MainController
       before_action :set_project
       before_action :set_package
+      before_action :set_filename, only: :update
       after_action :verify_authorized
 
       def new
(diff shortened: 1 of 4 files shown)


src/api/app/controllers/webui/packages/files_controller.rb
```diff
@@ -64,15 +65,21 @@ def update
...
 
         status ||= 200
         render layout: false, status: status, partial: 'layouts/webui/flash', object: flash
       end
+
+      private
+
+      def set_filename
+        @filename = params[:filename]
+      end
```
danidoni, line 82: We don't gain anything by doing this... And we're increasing amount of code... I would just use the param as it was before...

danidoni: Oh I see you wanna use that instance variable below... NVM

danidoni (approved):

PR #15760, opened 2024-03-07
Comment for isolated migrations

saraycp:
The Isolated Migrations check doesn't cover all the project's necessities:

- it fails when we commit the models' annotations together with the migrations,
- it fails when we commit validations and their corresponding database consistency fix (which should go together),
- ...

There is no need to have a check that fails or passes. It is enough with a comment warning about the presence of migrations together with other changes.

With these two new GitHub actions, the PR can get any of the comments  in the following image when:

- there are migrations together with other changes (except schema changes)
- there are migrations without their corresponding schema

Screenshot 2024-03-11 at 23-16-05 WIP warn migration by saraycp · Pull Request #29 · saraycp_hello_world

Pull requests that come from a fork don't have write permissions, so we can't write comments to a PR from one single GitHub action. That's why this is split into two actions to make it work and be safe, according to [this article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

Moreover, these actions should be in production to work. Otherwise, the second action/workflow will never run.


saraycp:
Closing in favor of: https://github.com/openSUSE/open-build-service/pull/15824

PR #15761, opened 2024-03-07
Exclude comments without a commentable from the list

ncounter:
Fix https://github.com/openSUSE/open-build-service/issues/15748

For comments created before the `commentable` object feature, the latter is `nil`. This is an issue for the user's page in the `Comments` tab where all comments are rendered with related commentable link and much more behind the scene (edit form, delete the comment, reply to the comment).

This is a PR to workaround the issue due to the complexity of the current comment rendering chain:
- `BsRequestCommentComponent` --> partial `webui/comment/comment_field`
both the above files make use of the `commentable` object to create links, buttons, forms etc.

A `ReadonlyCommentComponent` should be implemented in order to render a vanilla comment without all the possible interactions instead.


ncounter:
> I worry that this fixes the issue here, but not future issues. Instead we should prepare a data migration that gets rid of existing comments with nil commentable, and validate that no future comments end up without one.

I agree a deeper and wider fix should be implemented, either a migration or a component to render all the corner cases, but that will take time, meanwhile users are hitting the error without a workaround, which is the target of this PR.
Once the proper fix is implemented, this PR can be reverted.

The minor problem I have with the migration is: do we really want to delete them all?

saraycp:
If `default_scope` didn't have so many [side effects](https://stackoverflow.com/questions/25087336/why-is-using-the-rails-default-scope-often-recommend-against)... would be the temporary solution :thinking:

PR #15762, opened 2024-03-07
Autocorrect RSpec/LeadingSubject RuboCop offenses, part I

eduardoj:
We have almost 400 offenses of this kind. I split fixing the offenses in several pull requests, to make it easier to review the changes proposed.

### DIFF
--- a/src/api/spec/policies/appeal_policy_spec.rb
+++ b/src/api/spec/policies/appeal_policy_spec.rb
@@ -1,11 +1,11 @@
 RSpec.describe AppealPolicy do
+  subject { AppealPolicy }
+
   let(:anonymous_user) { create(:user_nobody) }
   let(:user) { create(:confirmed_user) }
   let(:moderator) { create(:moderator) }
   let(:staff_user) { create(:staff_user) }
   let(:admin_user) { create(:admin_user) }
-
-  subject { AppealPolicy }
 
   context 'with the content_moderation feature disabled' do
     let(:appeal) { create(:appeal) }
--- a/src/api/spec/policies/appeal_policy_spec.rb
+++ b/src/api/spec/policies/appeal_policy_spec.rb
@@ -1,5 +1,5 @@
 RSpec.describe AppealPolicy do
-  subject { AppealPolicy }
+  subject { described_class }
 
   let(:anonymous_user) { create(:user_nobody) }
   let(:user) { create(:confirmed_user) }
--- a/src/api/spec/policies/cloud/azure/configuration_policy_spec.rb
+++ b/src/api/spec/policies/cloud/azure/configuration_policy_spec.rb
@@ -1,10 +1,10 @@
 RSpec.describe Cloud::Azure::ConfigurationPolicy do
+  subject { described_class }
+
   let(:user) { create(:user) }
   let(:other_user) { build(:user) }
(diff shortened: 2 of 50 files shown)


src/api/spec/policies/event_subscription/form_policy_spec.rb
```diff
@@ -1,12 +1,12 @@
 RSpec.describe EventSubscription::FormPolicy do
+  subject { described_class }
```
hennevogel, line 2: This is even already there implicitly

src/api/spec/policies/group_policy_spec.rb
```diff
@@ -1,13 +1,13 @@
 RSpec.describe GroupPolicy do
+  subject { GroupPolicy }
```
hennevogel, line 2: This is even already there implicitly

src/api/spec/policies/package_policy_spec.rb
```diff
@@ -1,13 +1,13 @@
 RSpec.describe PackagePolicy do
+  subject { PackagePolicy }
```
hennevogel, line 2: This is even already there implicitly

src/api/spec/policies/staging/staged_request_policy_spec.rb
```diff
@@ -1,12 +1,12 @@
 RSpec.describe Staging::StagedRequestPolicy do
+  subject { Staging::StagedRequestPolicy }
```
hennevogel, line 2: This is even already there implicitly

src/api/spec/policies/staging/workflow_policy_spec.rb
```diff
@@ -1,11 +1,11 @@
 RSpec.describe Staging::WorkflowPolicy do
+  subject { Staging::WorkflowPolicy }
```
hennevogel, line 2: This is even already there implicitly

hennevogel (changes requested):

src/api/spec/policies/appeal_policy_spec.rb
```diff
@@ -1,12 +1,12 @@
 RSpec.describe AppealPolicy do
+  subject { AppealPolicy }
```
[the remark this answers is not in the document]
hennevogel: There are a lot of those, you get the idea...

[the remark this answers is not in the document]
eduardoj: Sure, thanks! I changed it on all your comments and found a couple of them more. Please take a look again.

src/api/spec/policies/cloud/azure/configuration_policy_spec.rb
```diff
@@ -1,11 +1,11 @@
 RSpec.describe Cloud::Azure::ConfigurationPolicy do
+  subject { described_class }
```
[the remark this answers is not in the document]
eduardoj: This is the right way of defining the test.

From the link of implicity you shared: _"If the first argument to an example group is a class, an instance of that class is exposed to each example in that example group via the subject method."_

The implicity subject takes an instance of a class. To take the class, we should use  `subject { described_class }`. This applies to all Pundit specs.

hennevogel (approved):

PR #15763, opened 2024-03-07
Skip email notifications for hidden projects

hennevogel:
Email is notoriously unsafe in transport and access disabled projects are often used in distribution workflows about embargoed software updates. Let's skip emails if they are about access protected projects.


src/api/app/models/event/request.rb
```diff
@@ -100,6 +100,13 @@ def target_package_watchers
       source_or_target_package_watchers(project_type: 'targetproject', package_type: 'targetpackage')
     end
 
+    def involves_hidden_project?
+      bs_request = BsRequest.find_by(number: payload[:number])
+      return false unless bs_request
+
+      bs_request.bs_request_actions.any?(&:invovles_hidden_project?)
```
eduardoj, line 107:
```suggestion
      bs_request.bs_request_actions.any?(&:involves_hidden_project?)
```
Typo.

eduardoj (approved):

PR #15764, opened 2024-03-07
Reduce max line length to a more sane setting

eduardoj:


danidoni:
I'm not sure about this change... How did you came into the conclusion that 220 is better?

eduardoj:
> I'm not sure about this change... How did you came into the conclusion that
> 
> ```
> Layout/LineLength:
>   Max: 220
> ```
> 
> is better?

The default max length configuration value for this cop is 120. The effort to reduce from any high value to something near to the default value is worth it, for the same reasons explained here: https://rubystyle.guide/#max-line-length .

danidoni:
> The default max length configuration value for this cop is 120. The effort to reduce from any high value to something near to the default value is worth it, for the same reasons explained here: https://rubystyle.guide/#max-line-length .

Then, why not lowering to 120?

eduardoj:
> Then, why not lowering to 120?

Basically, I prefer to approach reaching the 120 value gradually:
- we address the cases where the length of the lines are bigger
- The pull request will be more complicated to create and to review

danidoni:
> > Then, why not lowering to 120?
> 
> Basically, I prefer to approach reaching the 120 value gradually:
> 
>     * we address the cases where the length of the lines are bigger
> 
>     * if we present a PR lowering the value to 120, the pull request will be more complicated to create and to review

I don't understand, [the documentation](https://docs.rubocop.org/rubocop/cops_layout.html#layoutlinelength) says this cop is full aucorrectable. What's the difference between doing it in phases and doing it all at once?

eduardoj:
> I don't understand, [the documentation](https://docs.rubocop.org/rubocop/cops_layout.html#layoutlinelength) says this cop is full aucorrectable. What's the difference between doing it in phases and doing it all at once?

Our codebase is big enough to be worth it to split the changes.

I created a draft PR with all the changes at once: #15800.

PR #15766, opened 2024-03-08
Move view_file into files controller

hellcp-work:
Depends on #15754


src/api/app/views/webui/package/_breadcrumb_items.html.haml
```diff
@@ -25,6 +25,3 @@
       = link_to 'Repository State', project_package_repository_binaries_path(@project, @package_name, @repository)
     %li.breadcrumb-item.active{ 'aria-current' => 'page' }
       Statistics
-  - if current_page?(package_view_file_path(@project, @package))
```
danidoni, line 28: Why this is being removed?

danidoni: I'm sorry but I don't see the same code in there...

danidoni: Alright

danidoni (approved):

PR #15767, opened 2024-03-08
[api] tune fork command error handling

adrianschroeter:
* Do not insist on write permissions in the source (adapting test case for this as well)
* Complain about existing target package by default. Allowing overwrite via force parameter as the branch command behaves
* Also adding missing force parameter to branch api docu

### DIFF
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -1138,10 +1138,8 @@ class SourceController < ApplicationController
 
   # POST /source/<project>/<package>?cmd=fork&scmsync="url"&target_project="optional_project"
   def package_command_fork
-    @target_project_name ||= User.session!.branch_project_name(params[:project])
-
     # The branch command may be used just for simulation
-    verify_can_modify_target!
+    verify_can_modify_target! if @target_project_name
 
     raise MissingParameterError, 'scmsync url is not specified' if params[:scmsync].blank?
 
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -337,7 +337,7 @@ class SourceController < ApplicationController
     @deleted_package = params.key?(:deleted)
 
     # FIXME: for OBS 3, api of branch and copy calls have target and source in the opposite place
-    if params[:cmd].in?(%w[branch release])
+    if params[:cmd].in?(%w[branch fork release])
       @target_package_name = params[:package]
       @target_project_name = params[:target_project] # might be nil
       @target_package_name = params[:target_package] if params[:target_package]
@@ -1138,6 +1138,8 @@ class SourceController < ApplicationController
 
   # POST /source/<project>/<package>?cmd=fork&scmsync="url"&target_project="optional_project"
   def package_command_fork
+    @target_project_name ||= User.session!.branch_project_name(params[:project])
+
     # The branch command may be used just for simulation
     verify_can_modify_target!
 
--- a/src/api/app/controllers/source_controller.rb
+++ b/src/api/app/controllers/source_controller.rb
@@ -336,14 +336,14 @@ class SourceController < ApplicationController
     #--------------------
     @deleted_package = params.key?(:deleted)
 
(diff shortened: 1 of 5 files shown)


dmach:
What I've tested so far:

- [ ] Forks a package that doesn't contain `scmsync` node - shouldn't it error out instead?
- [x] Forking to an existing target errors out as expected
- [ ]  Forking to an existing target with `&force=1` overwrites only `scmsync` node, everything else remains - is it expected?
- [x] Works fine for both admin and an unprivileged user that has no permissions in the source
- [ ] `&force=0` still forces the fork :exclamation:

adrianschroeter:
"Forks a package that doesn't contain scmsync node - shouldn't it error out instead?"

I do not think so, you could use any project as template when you are converting to git.

adrianschroeter:
"Forking to an existing target with &force=1 overwrites only scmsync node, everything else remains - is it expected?" 

I have no opinion about that. Title and description are highly unimportent at build time at least.

adrianschroeter:
"&force=0 still forces the fork"

well, we should document it only as boolean "&fork" without any parameter. The =1/0 was only a workaround for old rails behaviour unable to deal with it.

dmach:
Yes, document `&force=0` better please. I consider that the only showstopper in this patch.
I'm perfectly fine with the rest. Mainly wanted to ask if the behavior is expected.

adrianschroeter:
urgs ... I wasn't aware that this one was not merged yet ... so fixed it again and pushed over it...

saraycp:
@adrianschroeter I pushed two commits with little refactorings.

Instead of setting the default `target_project_name` again I propose skipping the call to `verify_can_modify_target!` when `target_project_name`  is nil and relying on the `BranchPackage#branch` to set the default name. That's something we already do in [package_command_branch](https://github.com/openSUSE/open-build-service/blob/005d398343d3e9d4d1904136e6f13b861f2f5234/src/api/app/controllers/source_controller.rb#L1134) and [package_command_release](https://github.com/openSUSE/open-build-service/blob/005d398343d3e9d4d1904136e6f13b861f2f5234/src/api/app/controllers/source_controller.rb#L1005) (which calls `_package_command_release_manual_target` which calls  `verify_can_modify_target!`).

Feel free to handle my commits as you consider.

Apart from that, I think we should revisit the documentation. The endpoint should look like this [ /source/{project_name}/{package_name}?cmd=branch](https://api.opensuse.org/apidocs/index#/Sources%20-%20Packages/post_source__project_name___package_name__cmd_branch) and not like this [/source?cmd=fork](https://api.opensuse.org/apidocs/index#/Sources/post_source_cmd_fork) where the path parameters are missing. But that's something we can do in another PR as many other endpoints are also affected.

adrianschroeter:
I personally think that it increases the complexity for testing when the same permission checking happens in two different places depending on a detail (default project name versus user specified project name). I would changed the checking in branch case to simplify that there instead.

But in the end it is more important to get it fixed for now, it can be improved later.

hennevogel (approved):

src/api/app/controllers/source_controller.rb
```diff
@@ -336,14 +336,14 @@ def require_package
...
     @deleted_package = params.key?(:deleted)
 
+    @target_package_name = params[:package]
+
     # FIXME: for OBS 3, api of branch and copy calls have target and source in the opposite place
-    if params[:cmd].in?(%w[branch release])
-      @target_package_name = params[:package]
+    if params[:cmd].in?(%w[branch fork release])
       @target_project_name = params[:target_project] # might be nil
       @target_package_name = params[:target_package] if params[:target_package]
```
eduardoj, line 344: Adding the `fork` command to this `if` condition will make it possible to read the `target_package` query parameter and overwrite the `package` parameter from the path.

If this is what we really want, the `target_package` query parameter should added to the `...?cmd=fork` endpoint API documentation.

saraycp: pushed

eduardoj (approved):

PR #15772, opened 2024-03-08
Fix maintenance workflow flaky test

danidoni:
Accepting the request may take a bit longer than 6 seconds, that is the
default timeout we have configured for Capybara operations.

We introduce a new helper to temporarily execute some code with a different timeout.

See https://github.com/teamcapybara/capybara?tab=readme-ov-file#asynchronous-javascript-ajax-and-friends for more details.

### DIFF
--- a/src/api/spec/features/webui/maintenance_workflow_spec.rb
+++ b/src/api/spec/features/webui/maintenance_workflow_spec.rb
@@ -71,7 +71,10 @@ RSpec.describe 'MaintenanceWorkflow', :js, :vcr do
     fill_in('reason', with: 'really? ok')
 
     click_button('Accept request')
-    expect(page).to have_css('#overview h3', text: "Request #{bs_request.number} accepted")
+    # Looks like accepting the request takes some time, so we allow it to take a bit more than usual
+    wait_up_to(12.seconds) do
+      expect(page).to have_css('#overview h3', text: "Request #{bs_request.number} accepted")
+    end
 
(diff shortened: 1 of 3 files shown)


eduardoj (approved):

src/api/spec/features/webui/maintenance_workflow_spec.rb
```diff
@@ -71,7 +71,10 @@
     fill_in('reason', with: 'really? ok')
 
     click_button('Accept request')
-    expect(page).to have_css('#overview h3', text: "Request #{bs_request.number} accepted")
+    # Looks like accepting the request takes some time, so we allow it to take a bit more than usual
```
hennevogel, line 74: Isn't this a page load? What does this have to do with `Capybara.default_max_wait_time`?

danidoni: When using the `have_*` matchers (and the finders also), Capybara waits some time before saying "I didn't find that thing you were expecting" and failing with an error. That is the `Capybara.default_max_wait_time` (that we have set to 6).

See https://github.com/teamcapybara/capybara?tab=readme-ov-file#asynchronous-javascript-ajax-and-friends (bear in mind that the waiting happens even if there is no AJAX involved)

hennevogel: Which explicitly says it's about asynchronous javascript, but feel free to ignore me :)

PR #15773, opened 2024-03-08
Move remove_file to files controller

hellcp-work:
Depends on #15754


eduardoj (approved):
It's curious that moving an action from one controller to another, the cassettes now are more, and are bigger: `+4,781 −644`. Approving anyway.

PR #15774, opened 2024-03-08
Refactor LinkPackageStep spec

hennevogel:
Remove some superflous specs, test the things once in the right spec.


src/api/spec/models/workflow/step/link_package_step_spec.rb
```diff
@@ -9,24 +9,6 @@
                         token: token)
   end
 
-  RSpec.shared_context 'failed without link permissions' do
```
eduardoj, line 48: I'm concerned about removing this shared context and the lines that use it. I'm not aware we check that an error is raised for disabled access packages in some other place of the test suite.

Do you think we are safe removing it?

eduardoj:
Typos in one commit title: _"Adopt casettes..."_ should be _"Ad**a**pt cas**s**ettes..."_. `o` -> `a`, and missing `s`.

src/api/spec/models/workflow/step/link_package_step_spec.rb
hennevogel: This is "just" `Package.get_by_project_and_name` -> `Package.check_source_access?`. If this is broken we have other problems than `LinkPackageStep`...

eduardoj (approved):

PR #5182, opened 2018-06-21
Fix rubocop setup

eduardoj:
Splitting the configuration up makes the setup simpler and better to maintain.

Co-authored-by: Björn Geuken <bgeuken@suse.de>

PR #15776, opened 2024-03-08
Refine exclusions for Style/MultilineBlockChain

eduardoj:
Prevent from overlooking Rubocop offenses from this kind in other specs.

### DIFF
--- a/src/api/.rubocop.yml
+++ b/src/api/.rubocop.yml
@@ -195,7 +195,8 @@ Style/MethodCallWithArgsParentheses:
 # We disabled this cop because of Rantly.
 Style/MultilineBlockChain:
   Exclude:
-    - 'spec/**/*'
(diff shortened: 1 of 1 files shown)


danidoni:
Why do we make rubocop skip that?

eduardoj:
> Why do we make rubocop skip that?

The coment a couple of lines before should explain it... _"# We disabled this cop because of Rantly."_ 

See: https://github.com/openSUSE/open-build-service/blob/368c6f09dd2beff9220f19ddd1b8d7038b63c771/src/api/.rubocop.yml#L195

It comes from this pull request: #5182.

danidoni:
> The comment a couple of lines before should explain it... "# We disabled this cop because of Rantly."

Alright

danidoni (approved):

PR #15778, opened 2024-03-11
Wait in test 0030_project_spec

bmwiedemann:
Wait in test `0030_project_spec.rb`

Without this change, tests would randomly fail with

    Failures:

      1) Project is able to add repositories
         Failure/Error: check('openSUSE Leap 15.5')

         Capybara::ElementNotFound:
           Unable to find checkbox "openSUSE Leap 15.5" that is not disabled
         # ./spec/features/0030_project_spec.rb:28:in `block (2 levels) in <top (required)>'

    Finished in 22.38 seconds (files took 0.16822 seconds to load)
    5 examples, 1 failure

    Failed examples:

    rspec ./spec/features/0030_project_spec.rb:20 # Project is able to add repositories

### DIFF
--- a/dist/t/spec/features/0030_project_spec.rb
+++ b/dist/t/spec/features/0030_project_spec.rb
@@ -22,6 +22,7 @@ RSpec.describe 'Project', type: :feature do
       click_link('Your Home Project')
     end
     click_link('Repositories')
+    sleep(5)
     click_link('Add from a Distribution')
     check('openSUSE Leap 15.5')
     visit current_path


dist/t/spec/features/0030_project_spec.rb
```diff
@@ -22,6 +22,7 @@
       click_link('Your Home Project')
     end
     click_link('Repositories')
+    sleep(5)
```
danidoni, line 25: See [here](https://github.com/openSUSE/open-build-service/pull/15772/files#diff-5046c54ca51db7f1b66601407e62497dddd91065d2cba142ccb91dae3a8b6402R13) how it can be done more efficiently. Also, you can use [find](https://www.rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Finders#find-instance_method) to check for the existence of some element and wait for it to appear until the default timeout kicks in.

bmwiedemann: The tricky thing here is, that the "Leap 15.5" element is only displayed on the next page and I'm not even sure why waiting here helps. Maybe the backend syncs data over the interconnect from build.o.o?
So we would probably need to reload the 'Add from a Distribution' page until the checkbox appears.

danidoni: If "Leap 15.5" only is displayed after clicking on "Add from a Distribution", why don't you wait there?

bmwiedemann: I tested that only waiting here helped. Maybe we could go to the next page, wait and then reload it, but I don't think that is better.

PR #15781, opened 2024-03-11
Simplify Workflow::Step

hennevogel:


hennevogel:
Regarding codeclimate, I rather have this explicit and a big duplicated than implicit and short.

PR #15782, opened 2024-03-11
Move shared target object code into concern 

hennevogel:
LinkPackageStep and BranchPackageStep share a lot of code that deals with
projects / packages that are going to be created (targets). So far this code
has been in Workflow::Step. But it's useless for the other steps, they don't
need to inherit this.


hennevogel:
I need to make this smaller...

PR #15789, opened 2024-03-13
[api] fix package unlock behaviour

adrianschroeter:
it must not be allowed when project is still locked.

Regression which happened in 2.10 in a refactoring. (2.9 and before had the correct behaviour)


hennevogel:
As you can see now, this changes behavior. I'm not sure it's of any significance if this `:forbidden` and not `:bad_request`. Please clarify...

PR #15794, opened 2024-03-14
Deprecate action details

ncounter:
Depends on 
- https://github.com/openSUSE/open-build-service/pull/15810
- https://github.com/openSUSE/open-build-service/pull/15811
- https://github.com/openSUSE/open-build-service/pull/15817
- ~~https://github.com/openSUSE/open-build-service/pull/15827~~

`webui_actions` and `action_details` are old fashioned functions they return `Hash {}` objects
The object is a core typified one of the app, which is a `BsRequestAction`.

This PR forks the code-path differenciating the Beta version from the old one where the first one benefits of the `BsRequestAction` object with all its methods to retrieve all the properties at the time of the need, the former one still uses the old fashioned functions instead.

For now the old stack is just deprecated by a comment. Once Beta version will be rolled out, those deprecated functions can be dropped.

## Review:
Commit-by-commit is suggested


src/api/app/components/bs_request_action_description_component.rb
```diff
@@ -0,0 +1,55 @@
+# This component renders the request action description based on the type of the action
+
+class BsRequestActionDescriptionComponent < ApplicationComponent
```
danidoni, line 3: Maybe this component should go into it's own PR?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -519,9 +519,14 @@ def staging_status(request, target_project)
   end
 
   def cache_diff_data
-    return unless @action[:diff_not_cached]
+    bs_request_action = @action
+
+    # [DEPRECATED] TODO: drop the first part of the conditional after the request_workflow_redesign rollout
+    # we need this because before the beta version @action was still an Hash. With the beta it is a pure BsRequestAction instead
+    bs_request_action = BsRequestAction.find(@action[:id]) if @action.instance_of?(Hash)
```
danidoni, line 526:
```suggestion
    bs_request_action = BsRequestAction.find(@action.id) if @action.instance_of?(Hash)
```

src/api/app/components/bs_request_action_description_component.rb
ncounter: The problem is that without this component the full code-path doesn't work because only the beta version is using it and it starts to use it at the time the switch from the `Hash` to the `BsRequestAction` happens.

A separate PR to refactor the component and get rid of the `html_safe` will happen for sure though

src/api/app/controllers/webui/request_controller.rb
ncounter: Nope, the point is exactly that if the `@action` is an `Hash` we still need to fetch the `id` with `@action[:id]` because this method is used by both beta and non beta version, as the comment above describes it

src/api/app/components/bs_request_action_description_component.rb
danidoni: Then why don't pass first the component using the old hash version, and then switch it to the OOP version?

src/api/app/controllers/webui/request_controller.rb
ncounter: Yeah, I'm fine with that. I mean it's one line marked by the comment, but isolating the code path sounds safer.

ncounter: Done

ncounter: Okay, sorry for the noise, I made a mistake: `cache_diff_data` is used only in the beta version. Refactoring....... :sweat_smile:

src/api/app/components/bs_request_activity_timeline_component.rb
```diff
@@ -12,6 +12,7 @@ def initialize(bs_request:, request_reviews_for_non_staging_projects: [])
     @creator = User.find_by_login(bs_request.creator) || User.nobody
     action_comments = Comment.on_actions_for_request(@bs_request).without_parent.includes(:user)
     commented_actions = action_comments.map { |c| c.commentable.id }.uniq.compact
+    # TODO: webui_actions is DEPRECATED, get rid of it and use @bs_request.bs_request_actions instead
     @diffs = commented_actions.flat_map { |a| @bs_request.webui_actions(action_id: a, diffs: true, cacheonly: 1) }
```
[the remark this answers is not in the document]
ncounter: I added the approximate comment to make sure to come back later. And _later_ is now :wink: 
Thanks for double checking and making it clear in advance.

src/api/app/components/bs_request_action_description_component.rb
ncounter: https://github.com/openSUSE/open-build-service/pull/15817

src/api/app/components/bs_request_activity_timeline_component.rb
```diff
@@ -27,7 +26,7 @@ def diff_for_ref(comment)
     return unless (ref = comment.diff_ref&.match(/diff_([0-9]+)/))
 
     file_index = ref.captures.first
-    sourcediff = @diffs.find { |d| d[:id] == comment.commentable.id }[:sourcediff].first
+    sourcediff = @commented_actions.find { |d| d.id == comment.commentable.id }.webui_sourcediff.first
```
danidoni, line 29:
```suggestion
    sourcediff = @commented_actions.find(comment.commentable).webui_sourcediff.first
```
I belive they are equivalent but it needs to be tested

PR #15797, opened 2024-03-14
Removing diff_limit variable

danidoni:
It's not set in [Webui::RequestController#build_results](https://github.com/openSUSE/open-build-service/blob/b0ca06d47b8715eada4f689e1865e60c2c8bb8f7/src/api/app/controllers/webui/request_controller.rb#L292), so it's getting nil all the way down.


ncounter (commented):
Nice, JFYI I am addressing this already and widely in https://github.com/openSUSE/open-build-service/pull/15794 at the commit 02a31afd1c0b17ec9ac67aa2deafa82b608072f9

ncounter (approved):

PR #15806, opened 2024-03-15
Autocorrect RSpec/LeadingSubject RuboCop offenses, part III

eduardoj:
Follow up to #15798.

### DIFF
--- a/src/api/spec/controllers/distributions_controller_spec.rb
+++ b/src/api/spec/controllers/distributions_controller_spec.rb
@@ -45,11 +45,11 @@ RSpec.describe DistributionsController do
   end
 
   describe '#create' do
+    subject { post :create, body: distribution_xml, format: :xml }
+
     before do
       login admin
     end
-
-    subject { post :create, body: distribution_xml, format: :xml }
 
     it { is_expected.to have_http_status(:ok) }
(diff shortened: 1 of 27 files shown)


danidoni:
Why the change from `subject!/it is_expected` to `before/expect(response)`?

eduardoj:
> Why the change from `subject!/it is_expected` to `before/expect(response)`?

A spec over a controller action has an HTTP request just before the checks. Most of the checks over a controller action use `response` as the implicit subject. Using `subject` only in the cases where it is really needed and `response` in the rest of the cases helps to understand better this pattern used in controller specs.

I could see this pattern in the documentation: https://rspec.info/features/6-1/rspec-rails/controller-specs/controller-spec/

danidoni:
> A spec over a controller action has an HTTP request just before the checks. Most of the checks over a controller action use `response` as the implicit subject. Using `subject` only in the cases where it is really needed and `response` in the rest of the cases helps to understand better this pattern used in controller specs.
> 
> I could see this pattern in the documentation: https://rspec.info/features/6-1/rspec-rails/controller-specs/controller-spec/

Great! That info could into the commit description...

rubhanazeem (approved):

PR #15822, opened 2024-03-18
Remove useless index parameter from SourcediffComponent

danidoni:
It's not really a parameter, as we always use 0 as index...

### DIFF
--- a/src/api/app/components/sourcediff_component.html.haml
+++ b/src/api/app/components/sourcediff_component.html.haml
@@ -4,7 +4,7 @@
       .clearfix.mb-2.text-center
         .btn.btn-outline-primary.cache-refresh{ title: 'Refresh results', onclick: "loadChanges(#{@bs_request.number}, #{@action[:id]});" }
           Crunching the latest data. Refresh again in a few seconds
-          %i.fas.fa-sync-alt{ id: "cache#{@index}-reload" }
+          %i.fas.fa-sync-alt{ id: "cache#0-reload" }
     - else
       - (@action[:sourcediff] || []).each do |sourcediff|
         .clearfix.mb-2
--- a/src/api/app/components/sourcediff_component.rb
+++ b/src/api/app/components/sourcediff_component.rb
@@ -1,15 +1,14 @@
 class SourcediffComponent < ApplicationComponent
-  attr_accessor :bs_request, :action, :index, :refresh
+  attr_accessor :bs_request, :action, :refresh
 
   delegate :diff_label, to: :helpers
   delegate :diff_data, to: :helpers
 
-  def initialize(bs_request:, action:, index:)
+  def initialize(bs_request:, action:)
     super
 
     @bs_request = bs_request
     @action = action
-    @index = index
   end
 
   def commentable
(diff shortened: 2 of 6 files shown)


src/api/app/components/sourcediff_component.html.haml
```diff
@@ -4,7 +4,7 @@
       .clearfix.mb-2.text-center
         .btn.btn-outline-primary.cache-refresh{ title: 'Refresh results', onclick: "loadChanges(#{@bs_request.number}, #{@action[:id]});" }
           Crunching the latest data. Refresh again in a few seconds
-          %i.fas.fa-sync-alt{ id: "cache#{@index}-reload" }
+          %i.fas.fa-sync-alt{ id: "cache#0-reload" }
```
eduardoj, line 7:
```suggestion
          %i.fas.fa-sync-alt{ id: "cache-reload" }
```
I find the `#0` confusing. Do we really need it?

danidoni: Yes, it's the handler of the first item to be reloaded when the data crunching ends, so 0. If we had many more items to reload they would be `cache#0-reload`, `cache#1-reload`, `cached#2-reload`, etc... So, yes, the zero is needed.

danidoni: See https://github.com/openSUSE/open-build-service/blob/ed980b18d481c36b5e1b420449284c884dea9bae/src/api/app/assets/javascripts/webui/buildresult.js#L6

eduardoj: I see we need the index for `'#rpm'+index+'-reload'` elements. But we are now removing the `index` variable from `'#cache'+index+'-reload'` elements.

Is there some other place in the codebase where `'#cache'+index+'-reload'` elements with the `index` variable different than `0` are created or used?

danidoni: [No](https://github.com/search?q=repo%3AopenSUSE%2Fopen-build-service+cache+-reload&type=code)

eduardoj: > [No](https://github.com/search?q=repo%3AopenSUSE%2Fopen-build-service+cache+-reload&type=code)

Interesting.

In this PR we are removing the `index` variable from one component: `app/components/sourcediff_component.html.haml`. And we are leaving it on the other one: `app/components/sourcediff_tab_component.html.haml`.

Isn't it safe to remove it from the `sourcediff_tab_component` component too? Is this planned in a follow up pull request?

If there isn't some other place in the codebase where '#cache'+index+'-reload' elements with the `index` variable different than `0` are created or used, I would prefer to use `cache-reload` instead of `cache#0-reload`.

danidoni: > Isn't it safe to remove it from the `sourcediff_tab_component` component too? Is this planned in a follow up pull request?

Probably, right now I'm focused on this component. I will deal with `SourcediffTabComponent` elsewhere.

eduardoj (approved):

src/api/app/components/sourcediff_component.html.haml
eduardoj: I hope that later on, when you deal with `SourcediffTabComponent` the `cache#0-reload` string could be renamed to something more meaningful.

PR #15823, opened 2024-03-18
[backend] mkosi: publish images to per-arch subdirectory

bluca:
Images do not have the architecture in the name, so an image recipe building for multiple architectures would have clashing filenames, which means a single mkosi image build cannot currently enable multiple architectures. Change the publisher so that if we detect that we are publishing mkosi artifacts, they are stored in a per-architecture sub-folder.

### DIFF
--- a/src/backend/bs_publish
+++ b/src/backend/bs_publish
@@ -2418,6 +2418,10 @@ sub publish {
 	  $kiwimedium = "$arch/$1" if $bin =~ /(.+)ONIE\.bin$/;
 	} elsif ($bin =~ /(.*)\.raw(?:\.install)?(?:\.(?:gz|bz2|xz|zst|zstd))?(?:\.sha256(?:\.asc)?)?$/) {
 	  $p = $bin;
+	  if (my @files = glob("$r/mkosi.*")) {
+		# mkosi image names don't include the arch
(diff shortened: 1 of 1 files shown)


bmwiedemann:
FYI: https://build.opensuse.org/project/show/home:bluca:mkosi has example packages.

Vogtinator:
Or maybe document that the image id should include `%a`?

bluca:
It's a bit of a footgun, so it seems better to be safe I think

Vogtinator:
mkosi also doesn't use version numbers by default. I'm actually not sure whether all of the mirroring infrastructure is ready to use mtime or checksum based comparisions, but if there's some code left that relies on file name differences that would break too.

Vogtinator:
Does the subdir still work together with `Repotype: checksumsfile`?

bluca:
> Does the subdir still work together with `Repotype: checksumsfile`?

I have not tried that combination

bluca:
> > Does the subdir still work together with `Repotype: checksumsfile`?
> 
> I have not tried that combination

Yeah it turns out it doesn't work - I've opened a PR to also move other mkosi artifacts out of the per-arch subdir: https://github.com/openSUSE/open-build-service/pull/17309

PR #15831, opened 2024-03-18
Refactoring action forward logic

ncounter:
Depends on https://github.com/openSUSE/open-build-service/pull/15794#

Refactoring action forward logic

### DIFF
--- a/src/api/app/models/bs_request_action_submit.rb
+++ b/src/api/app/models/bs_request_action_submit.rb
@@ -152,38 +152,23 @@ class BsRequestActionSubmit < BsRequestAction
     request_creator.has_local_role?(Role.hashed['maintainer'], target_package_object)
   end
 
-  # rubocop:disable Metrics/BlockNesting
   def forward
-    forward_object = nil
+    return [] unless target_package_object
 
-    if target_package_object
-      linkinfo = target_package_object.linkinfo
-      target_package_object.developed_packages.each do |dev_pkg|
-        forward_object ||= []
-        forward_object << { project: dev_pkg.project.name, package: dev_pkg.name, type: 'devel' }
-      end
-      if linkinfo
-        lprj = linkinfo['project']
-        lpkg = linkinfo['package']
-        link_is_already_devel = false
-        if forward_object
-          forward_object.each do |forward|
-            if forward[:project] == lprj && forward[:package] == lpkg
-              link_is_already_devel = true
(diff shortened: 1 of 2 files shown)


eduardoj (approved):

src/api/app/models/bs_request_action_submit.rb
```diff
@@ -149,39 +149,21 @@ def creator_is_target_maintainer
...
-          end
-        end
-        unless link_is_already_devel
-          forward_object ||= []
-          forward_object << { project: linkinfo['project'], package: linkinfo['package'], type: 'link' }
-        end
-      end
+    linkinfo = target_package_object.linkinfo
+    # add all the devel packages into the forwards
+    forward_object = target_package_object.developed_packages.map do |dev_pkg|
```
danidoni, line 157: I think `map` will return `[]` so the `forward_object || []` is reduntant. I haven't tested it though

ncounter: I agree, thanks for the catch. It's a leftover after the IDE auto-fix.

src/api/app/models/bs_request_action_submit.rb
```diff
@@ -149,39 +149,21 @@ def creator_is_target_maintainer
...
+      { project: dev_pkg.project.name, package: dev_pkg.name, type: 'devel' }
+    end
+    # check if the link is already in the forwards, add it otherwise
+    if linkinfo && (forward_object.none? { |forward| forward[:project] == linkinfo['project'] && forward[:package] == linkinfo['package'] })
```
danidoni, line 161: So, we could do this, right?

```suggestion
    return forward_object unless linkinfo
    
    if forward_object.none? { |forward| forward[:project] == linkinfo['project'] && forward[:package] == linkinfo['package'] }
```

ncounter: We could, sure, but why splitting a condition logic into two pieces?

hennevogel: because return early makes code more readable for humans.

danidoni: The main idea here is that when you are reading code, processing the first return you are removing cognitive load from the second condition. Whatever we do that removes cognitive load from code, the better.

danidoni (approved):

PR #15843, opened 2024-03-20
Handle ActionMailer producing nil

hennevogel:
We can't use assume `email` is a `Mail::Message` when we rescue `StandardError`. `ActionMailer` might fail to produce the `Mail::Message` object somehow already.

- `EventMailer` produces `nil`
- `nil.deliver_now` produces `NoMethodError`
- `email.body` in `Airbrake.notify` will produce `NoMethodError` again


hennevogel:
This is probably only the first part to fix #15833

eduardoj (approved):

bmwiedemann (approved):

PR #15844, opened 2024-03-20
Fix spec warning

danidoni:
`bs_request.bs_request_actions.first[:name]` was returning `nil`, and that raised a warning like this one:

```
Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead. /obs/src/api/spec/features/webui/requests/submissions_spec.rb:142
```

Obviously that expectation was wrong.


ncounter (approved):
Nice catch, :+1:

PR #15847, opened 2024-03-21
Expand the message about why the request was declined

### LINKED ISSUE #15287: Beta view of SubmitRequests don't show message why it was declined
# Issue Description
The beta view doesn't show any Request History like in the old page. Which means when a Request gets declined, there is no way to see the message why it got declined. Here is a screenshot of an example. https://paste.opensuse.org/pastes/25984a24eee8 (sorry for the small font, but I wanted to get the page in one whole screenshot)

# Expected Result
See the reason why some Requests get declined

# How to reproduce
Create a SR which you decline and add a message to it why it got declined

### LINKED ISSUE #15711: Decline reason should be more prominent in new UI
Here it is virtually impossible to guess that there is a reason to be expanded from the decline entry.
image

Expanded entry:
image

Declined entries should always be expanded by default.
After all this is the most important information in existence regarding a SR.

rubhanazeem:
closes #15287, closes #15711

### DIFF
--- a/src/api/app/components/bs_request_history_element_component.html.haml
+++ b/src/api/app/components/bs_request_history_element_component.html.haml
@@ -20,9 +20,9 @@
       - if element.comment.present?
         = link_to("#collapse-history-element-comment-#{element.id}",
                     data: { 'bs-toggle': 'collapse' },
-                    aria: { expanded: false, controls: "collapse-history-element-comment" }) do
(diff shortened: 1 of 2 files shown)


danidoni:
Tested [here](https://obs-reviewlab.opensuse.org/rubhanazeem-uncollapse-declined/request/show/17?notification_id=33), the reason is expanded by default.

danidoni (approved):

Issue #15768, opened 2024-03-08
OBS tries to load rpmlint_log from disabled repo


jcronenberg:
# Issue Description
When going to https://build.opensuse.org/package/show/server:dns/unbound it shows an error that it can't load rpmlint_log from 15.4. The 15.4 repo is disabled so likely there isn't a rpmlint_log.

Expected Result
============
No Error

How to Reproduce
===============
1. Be logged in with beta features enabled
1. Go to https://build.opensuse.org/package/show/server:dns/unbound
1. See error toast at the top

Further Information
===============
![image](https://github.com/openSUSE/open-build-service/assets/54934253/7e5463cf-fe09-4c75-8651-72ceeb654e50)

cryptomilk:
Same for me on my home repo

PR #15852, opened 2024-03-21
Do not crash if no _log file found

ncounter:
Fix https://github.com/openSUSE/open-build-service/issues/15768

`Backend::Api::BuildResults::Binaries.file(project, repository, architecture, package, '_log')` crashes in case of `_log` file is missing. This PR handle the exception and returns a nicer message.


src/api/app/services/rpmlint_log_extractor.rb
```diff
@@ -40,7 +40,13 @@ def call
   def retrieve_rpmlint_log_from_log
     @parse_internal_log_file = true
 
-    log_content = Backend::Api::BuildResults::Binaries.file(project, repository, architecture, package, '_log')
+    log_content = begin
+      Backend::Api::BuildResults::Binaries.file(project, repository, architecture, package, '_log')
+    rescue Backend::NotFoundError => e # do not crash if no `_log` file found, just return empty log_content
+      @parse_internal_log_file = false
+      '' if e.summary.include?("package '#{package}' has no logfile")
```
eduardoj, line 47:
```suggestion
      ''
```
Why do we have to check for this error message? Shouldn't we return `''` for every `Backend::NotFoundError`?

ncounter: We could do that as well: I wonder what if the endpoint would return different messages for different scenarios. In my case we catch only the error when it is exactly with that wording, in yours we catch them all. But I think it is okay if we handle them all for now. Applying..

eduardoj: Hm... The error is catched always. Before, instead of returning `''` it would have returned `nil`...

PR #15863, opened 2024-03-22
Introduces dependency review action

hennevogel:
https://github.com/actions/dependency-review-action

### DIFF
--- a/.github/workflows/dependency-review-action.yml
+++ b/.github/workflows/dependency-review-action.yml
@@ -1,0 +1,18 @@
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  dependency-review:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review-action.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v4
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@v4
+        with:
+	  license-check: false
```
eduardoj, line 17: Shouldn't we enable the license check? Why is the check switched off?

hennevogel: Because you will have to manually setup an allow or deny list of license names...

eduardoj (approved):

PR #15865, opened 2024-03-22
Scroll to the specific comment on conversation

saraycp:
The notifications about comments on requests used to point to the `Changes`  tab, something which was not working since we started to load a new page per tab.

Now, the link goes to the specific comment in the conversation tab.


src/api/app/components/notification_notifiable_link_component.rb
```diff
@@ -88,8 +88,8 @@ def notifiable_link_path
     when 'Event::CommentForRequest'
       # TODO: It would be better to eager load the commentable association with `includes(...)`,
       #      but it's complicated since this isn't for all notifications and it's nested 2 levels deep.
```
eduardoj, line 90:
```suggestion
```
With the changes proposed in the line below, we could get rid of these comments. Couldn't we?

saraycp: I don't think the comment is related to the code I modified, which only sets the anchor of the link.

krauselukas: @eduardoj I agree, we are not loading the commentable anymore. The comment is outdated, I will remove it

eduardoj (approved):

PR #15873, opened 2024-03-23
Compute actions count instead of using a var of the value

ncounter:
Depends on https://github.com/openSUSE/open-build-service/pull/15872


danidoni:
Awesome

danidoni (approved):

PR #15881, opened 2024-03-25
Use logarithmic scale for build result summary chart

krauselukas:
Currently the click target is really small on the bars in the chart when there is a big margin between the values of the individual results (e.g. 100 Succeeding Builds and 1 Failed Build).
Using a logarithmic scale on the y-axis balances the difference on the bars in the chart.

In order to stay consistent with the ordering of the individual build results types and to distinguish the click target a little more I disabled stacking of the bars.

stacked
unstacked

**With a few more repositories**

image

### DIFF
--- a/src/api/app/components/chart_component.html.haml
+++ b/src/api/app/components/chart_component.html.haml
@@ -11,10 +11,10 @@
                       id: 'build-summary-chart',
                       width: '100%',
                       height: '300px',
-                      stacked: true,
+                      stacked: false,
                       responsive: true,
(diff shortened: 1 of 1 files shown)


krauselukas:
[Too small clickable area in the build summary chart (L)](https://trello.com/c/puHMUo6W/2445-too-small-clickable-area-in-the-build-summary-chart-l)

krauselukas:
Currently there is an issue https://github.com/chartjs/Chart.js/issues/9629 after enabling the logarithmic scale the y-axis doesn't start at 0, its starts at 0.1 There are some workarounds, but I would prefer to either do this in a follow up or not at all.

PR #15884, opened 2024-03-25
Fix deprecation of Socket.gethostbyname

danidoni:
It appears on the output of OpenQA runs, for example this one:

https://openqa.opensuse.org/tests/4038465#step/rspec_webui_tests/11


dist/t/spec/support/capybara.rb
```diff
@@ -24,7 +24,7 @@
 
 # Set hostname
 begin
-  hostname = Socket.gethostbyname(Socket.gethostname).first
+  hostname = Addrinfo.getaddrinfo(Socket.gethostname).first
```
eduardoj, line 27: With that change, we could also get rid of these lines in the RuboCop TODO file: https://github.com/openSUSE/open-build-service/blob/0a61c6e93b22936bf36864e2f999d84a29318ba5/.rubocop_todo.yml#L27-L32

dist/t/spec/support/capybara.rb
```diff
@@ -24,7 +24,7 @@
...
 rescue SocketError
```
eduardoj, line 28: Does still make sense to rescue from a SocketError?

bmwiedemann: https://ruby-doc.org/stdlib-3.0.0/libdoc/socket/rdoc/Addrinfo.html#method-c-getaddrinfo does not list it, but a quick test showed that it still throws a
```
SocketError (getaddrinfo: No address associated with hostname)
```

dist/t/spec/support/capybara.rb
bmwiedemann: In irb, this results in `ArgumentError (wrong number of arguments (given 1, expected 2..6))`
and the output format is different, too.

`Addrinfo.getaddrinfo(Socket.gethostname, 443, nil, :STREAM).first.getnameinfo[0]` also returns a name.

dist/t/spec/support/capybara.rb
eduardoj: > https://ruby-doc.org/stdlib-3.0.0/libdoc/socket/rdoc/Addrinfo.html#method-c-getaddrinfo does not list it, but a quick test showed that it still throws a
> 
> ```
> SocketError (getaddrinfo: No address associated with hostname)
> ```

Sorry for not understanding it. What do you mean by "a quick test"?

danidoni: If you go into an IRB session and try to resolve a hostname that does not exist, you'll see that it throws a `SocketError` error

src/api/.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config --auto-gen-only-exclude --exclude-limit 100`
-# on 2024-03-18 14:23:06 UTC using RuboCop version 1.62.1.
+# on 2024-03-26 09:01:51 UTC using RuboCop version 1.62.1.
```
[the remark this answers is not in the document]
danidoni: Alright, fixed!

eduardoj (approved):

PR #15886, opened 2024-03-25
Fix OpenQA due to a bug in Apparition gem

danidoni:
Our OpenQA tests use the [Apparition](https://github.com/twalpole/apparition) Capybara driver. 

There is a [bug](https://github.com/twalpole/apparition/issues/80) that is not fixed  in the latest version (0.6.0). 

This PR pins the gem to a specific commit id to overcome that bug.

### DIFF
--- a/dist/t/Gemfile
+++ b/dist/t/Gemfile
@@ -4,4 +4,7 @@ gem 'capybara'
 gem 'rspec-core'
 gem 'rspec-expectations'
 # as driver for capybara
-gem 'apparition'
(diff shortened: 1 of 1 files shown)


eduardoj:
Typo in the commit title: `but` -> `bug`

danidoni:
> Typo in the commit title: `but` -> `bug`

Fixed

danidoni:
We're not sure this is the reason OpenQA fails...

PR #15887, opened 2024-03-25
Defer Diff Loading on Changes Tab

ncounter:
Applying the common pattern: load the static/text part of the page within the main page request, send an async call right after the page load to fetch the heavy content of the page while a loading message is visible.

## Test
Visit https://obs-reviewlab.opensuse.org/ncounter-async-loading-changes/request/show/1/changes

## Before: the page is loaded fully sync - spot the loading icon at the top of the browser tab
before-async-loading

## After: the page is quickly loaded in the header and the rest but the content, that starts loading async

after-async-loading

### DIFF
--- a/src/api/app/controllers/webui/request_controller.rb
+++ b/src/api/app/controllers/webui/request_controller.rb
@@ -16,6 +16,7 @@ class Webui::RequestController < Webui::WebuiController
                              if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :set_influxdb_data_request_actions, only: %i[show build_results rpm_lint changes mentioned_issues],
                                                     if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
+  before_action :set_diff_to_superseded_id, only: %i[show request_action request_action_changes build_results rpm_lint changes mentioned_issues]
   before_action :set_superseded_request, only: %i[show request_action request_action_changes build_results rpm_lint changes mentioned_issues]
   before_action :check_ajax, only: :sourcediff
   before_action :prepare_request_data, only: %i[show build_results rpm_lint changes mentioned_issues],
@@ -34,7 +35,6 @@ class Webui::RequestController < Webui::WebuiController
       render :beta_show
     else
       @diff_limit = params[:full_diff] ? 0 : nil
-      @diff_to_superseded_id = params[:diff_to_superseded]
       @is_author = @bs_request.creator == User.possibly_nobody.login
 
       @is_target_maintainer = @bs_request.is_target_maintainer?(User.session)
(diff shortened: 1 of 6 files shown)


src/api/app/views/webui/request/_changes_content.html.haml
```diff
@@ -0,0 +1,12 @@
+- if diff_to_superseded_id
+  You're reviewing changes from
+  = link_to("superseded request ##{diff_to_superseded_id}.", request_show_path(number: diff_to_superseded_id))
```
krauselukas, line 3: @ncounter we should stick to the bs_request number here. `id !=number`. Usually its the same, but its still two different things, and could lead to problems if we exchange them.

krauselukas: ah its just the naming, you are using the number. but we should stay consistent with the naming here

ncounter: Exactly, it's only a variable name to the loop to keep the queryString parameter alive. The parameter is populated in [set_diff_to_superseded_id](https://github.com/openSUSE/open-build-service/pull/15887/files/950fb4d3a2202168031e5dd0bfab29e4a3ed87cc#diff-4240c3b55deedb4b829fef429bfc8fd6832885a1ed59928cabccd613a1494e18R356). 

I'm going to rename it. Thinking about it: before it was `diff_to_superseded`, now it is `diff_to_superseded_id`, I'm going for `diff_to_superseded_number` but I also wonder if I could get rid of the `diff_to` prefix and finalize it to `superseded_number` only :thinking: The fact that we use it for the `diff` is a detail, we could use it for something else too. What do you think?

src/api/app/controllers/webui/request_controller.rb
```diff
@@ -16,6 +16,7 @@ class Webui::RequestController < Webui::WebuiController
                              if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
   before_action :set_influxdb_data_request_actions, only: %i[show build_results rpm_lint changes mentioned_issues],
                                                     if: -> { Flipper.enabled?(:request_show_redesign, User.session) }
+  before_action :set_diff_to_superseded_id, only: %i[show request_action request_action_changes build_results rpm_lint changes mentioned_issues]
```
krauselukas, line 19: @ncounter one more thing, couldn't this also be done in `:set_superseded_request`? IMO its not worth to create another callback and method just to set this variable.

src/api/app/views/webui/request/_changes_content.html.haml
ncounter: As discussed and decided offline, the renaming change will happen in a separate PR due to the fact the changes involves much more than the scope of this PR.

https://github.com/openSUSE/open-build-service/pull/15906

PR #15890, opened 2024-03-26
apidocs: Add missing params for branch

bmwiedemann:
Add missing parameters for `/source/{project_name}/{package_name}?cmd=branch`
as used by `osc branch -f -r 123 ...`

https://github.com/openSUSE/osc/blob/ec363bb/osc/core.py#L6138

### DIFF
--- a/src/api/public/apidocs/paths/source_project_name_package_name_cmd_branch.yaml
+++ b/src/api/public/apidocs/paths/source_project_name_package_name_cmd_branch.yaml
@@ -77,6 +77,19 @@ post:
           - 1
       description: Set to ignore validation and resolving of devel package or devel project definition.
     - in: query
+      name: force
+      schema:
+        type: string
+        enum:
+          - 1
+      allowEmptyValue: true
+      description: Set to allow overwriting of a pre-existing package.
+    - in: query
+      name: rev
+      schema:
+        type: string
+      description: Set to branch against a specific revision.
+    - in: query
       name: maintenance
       schema:
         type: string


src/api/public/apidocs/paths/source_project_name_package_name_cmd_branch.yaml
```diff
@@ -76,6 +76,18 @@ post:
         enum:
           - 1
       description: Set to ignore validation and resolving of devel package or devel project definition.
+    - in: query
+      name: force
+      schema:
+        type: string
+        enum:
+          - 1
+      description: Set to allow overwrite of a pre-existing package.
```
eduardoj, line 85: I would use `allow overwriting` or `allow to overwrite`. See this [link](https://www.grammaring.com/to-infinitive-or-gerund-advise-recommend-allow-permit-forbid-require).

adrianschroeter:
That is fixed here in a bit more correct way:  https://github.com/openSUSE/open-build-service/pull/15767

eduardoj:
> That is fixed here in a bit more correct way: #15767

Ops, yes. That's true for the case of the `force` parameter. Not for the case of the `rev` parameter.

bmwiedemann:
PR 15767 updates other files (API endpoints) so we can have both. I added your `allowEmptyValue: true` here.

src/api/public/apidocs/paths/source_project_name_package_name_cmd_branch.yaml
bmwiedemann: Updated

PR #15895, opened 2024-03-27
Loading partial

ncounter:
The simple ` Loading <something> ...` with the spinning icon is a pattern to be reused: this PR makes it a reusable partial.

PS: this depends on https://github.com/openSUSE/open-build-service/pull/15887 , because that PR introduces two occurrences of this partial. Only the last commit is the target of this current PR

### DIFF
--- a/src/api/app/views/webui/shared/_loading.haml
+++ b/src/api/app/views/webui/shared/_loading.haml
@@ -1,3 +1,3 @@
-.text-center.p-4{ class: wrapper_css ||= '' }
+.text-center.p-4{ class: wrapper_css || '' }
   %i.fas.fa-spinner.fa-spin.me-1
   = text
--- a/src/api/app/views/webui/shared/_loading.haml
+++ b/src/api/app/views/webui/shared/_loading.haml
@@ -1,0 +1,3 @@
+.text-center.p-4{ class: wrapper_css ||= '' }
+  %i.fas.fa-spinner.fa-spin.me-1
+  = text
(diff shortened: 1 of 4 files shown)


src/api/app/views/webui/shared/_loading.haml
```diff
@@ -0,0 +1,3 @@
+.text-center.p-4{ class: wrapper_css ||= '' }
```
saraycp: I'd say this is what we need:

```suggestion
.text-center.p-4{ class: wrapper_css || '' }
```

saraycp (commented):
Is there a reason to use a partial and not a ViewComponent? I pushed a commit with the ViewComponent alternative and the fix I commented on inline.
Feel free to discard my commit if it doesn't convince you.

danidoni:
> Is there a reason to use a partial and not a ViewComponent? I pushed a commit with the ViewComponent alternative and the fix I commented on inline. Feel free to discard my commit if it doesn't convince you.

I believe the logic behind using a ViewComponent should be encapsulating complexity to allow reuse and testing. In this case there is almost no logic involved. Using a view component here introduces a bit of overhead that it's not worth it.

saraycp:
> > Is there a reason to use a partial and not a ViewComponent? I pushed a commit with the ViewComponent alternative and the fix I commented on inline. Feel free to discard my commit if it doesn't convince you.
> 
> I believe the logic behind using a ViewComponent should be encapsulating complexity to allow reuse and testing. In this case there is almost no logic involved, so there will be no testing as well. Using a view component here introduces a bit of overhead that it's not worth it.

From the [documentation](https://viewcomponent.org/#tldr): _"ViewComponents work best for templates that are reused  [...]"_  But I understand what you explain.

saraycp:
@danidoni @ncounter Understood! I removed the ViewComponent commit and pushed a new one with the fix commented on [this line](https://github.com/openSUSE/open-build-service/pull/15895#discussion_r1546817776).

ncounter:
> @danidoni @ncounter Understood! I removed the ViewComponent commit and pushed a new one with only the fix commented on [this line](https://github.com/openSUSE/open-build-service/pull/15895#discussion_r1546817776).

Awesome, and thank you for the fix :slightly_smiling_face:

saraycp (approved):

PR #15898, opened 2024-03-27
Set X-Opensuse-Errorcode for pundit errors

hennevogel:
As we do in ApplicationController#render_error


hennevogel:
https://github.com/openSUSE/open-build-service/blob/599e101fd2a74e9022a175e683236d398a02b72a/src/api/app/controllers/application_controller.rb#L161-L175

eduardoj (approved):

PR #15899, opened 2024-03-27
Spec SourceController package_command

hennevogel:
Some day I'm going to refactor this. Before that day, this should be spec'ed


hennevogel:
If you are looking for a birthday/easter/eid present for me, help to reduce the `skip`'s

src/api/spec/controllers/source_controller_package_command_spec.rb
```diff
@@ -0,0 +1,312 @@
...
+  end
+
+  describe 'POST #set_flag' do
+    subject { post :package_command, params: { project: project, package: package, cmd: 'set_flag' } }
+
+    it 'adds the flag'
+
+    context 'without flag param' do
+      it 'MissingParameterError' do
+        skip('to be tested')
```
danidoni, line 286:
```suggestion
      xit 'MissingParameterError'
```
JFYI. You can shortcut the skips like this.

hennevogel:
Depends on #17485

PR #13748, opened 2023-01-20
Show the reviewer why their review was requested

eduardoj:
**Before:**

Screenshot from 2023-01-20 14-58-14

**After:**

Screenshot from 2023-01-20 14-57-49

### For reviewers

- Branch a package of another user.
- Add yourself as reviewer, clicking on the "Add reviewer" button. If you have logged in as user `user_1`, add `user_1` as reviewer.
  - Introduce different texts for the "Comment for reviewer" textarea.
- Click on the "Review" dopdown button.
- See that the different reasons are displayed.

PR #15900, opened 2024-03-27
Show the reviewer why their review was requested

krauselukas:
Based on https://github.com/openSUSE/open-build-service/pull/13748 + a fix for the N+1 queries. We already fetch the History Elements for the request timeline. Instead of querying them again for the reviews it makes more sense to filter them in memory.


krauselukas:
[Explain why a review was requested from me (M)](https://trello.com/c/QZWCW9BR/2451-explain-why-a-review-was-requested-from-me-m)

krauselukas:
Codeclimate complains because I added one line of code to the `show` endpoint which is already huge and needs refactoring. I would ignore this for this PR...

danidoni (approved):

PR #15901, opened 2024-03-27
Drop generated backend lib docu

hennevogel:
If you want to have it, use rdoc to generate it...


eduardoj:
There are other lines that could be removed:
- https://github.com/openSUSE/open-build-service/blob/a68cb1e62a25c01614f7cf5c1e66512cda832d8a/.codeclimate.yml#L57
- https://github.com/openSUSE/open-build-service/blob/a68cb1e62a25c01614f7cf5c1e66512cda832d8a/.jshintignore#L4

eduardoj (approved):

PR #15904, opened 2024-03-28
[backend] scmsync: allow to add build config lines via meta

adrianschroeter:
A possible usecase is to build only a small set of packages from a large project. We need to have another place, beside the git source to define this subset for temporary test builds. This is because the git source must be always in a mergable state.

This implementation allows to add any build config line via project meta description. Any #! prefixed line of project meta will be put on top of the _config file.

Please note that this is not the final place, but avoids any compat breakage for now and allows people to play with it.

An example description may look like this:
...
 <description>
 This is my test build for Id X
 #!BuildFlags: onlybuild:acl
 #!BuildFlags: onlybuild:bc
 Some other text describing the project
 </description

### DIFF
--- a/src/backend/BSSrcServer/Scmsync.pm
+++ b/src/backend/BSSrcServer/Scmsync.pm
@@ -249,16 +249,22 @@ sub sync_project {
 
   # update the project config
   my $config = '';
+  # TEMPORARY HACK: checking for additional project config in project description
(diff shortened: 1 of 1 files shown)


adrianschroeter:
solved via https://github.com/openSUSE/obs-scm-bridge/pull/25 instead.

PR #15931, opened 2024-04-04
Less global variables, more methods call

ncounter:
_[Targeting Beta RequestWorkflowRedesign only]_

Use the Model method instead of storing data into a global variable and carry it around.

### DIFF
--- a/src/api/app/components/add_review_dropdown_component.rb
+++ b/src/api/app/components/add_review_dropdown_component.rb
@@ -1,11 +1,10 @@
 class AddReviewDropdownComponent < ApplicationComponent
-  def initialize(bs_request:, user:, my_open_reviews:, history_elements:)
+  def initialize(bs_request:, user:, my_open_reviews:)
     super
 
     @bs_request = bs_request
     @user = user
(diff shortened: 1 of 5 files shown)


src/api/app/components/add_review_dropdown_component.rb
```diff
@@ -26,7 +25,8 @@ def reviewer_icon_and_text(review:)
   end
 
   def reason_when_review_was_requested(review:)
-    reason = @history_elements.reverse.find { |history_element| history_element.type == 'HistoryElement::RequestReviewAdded' && history_element.description_extension == review.id.to_s }&.comment
+    history_elements = bs_request.history_elements.includes(user)
```
hennevogel, line 28: https://github.com/openSUSE/open-build-service/wiki/View-Components#avoid-database-queries

ncounter: Argh, there is a query behind the scenes. Thanks for the pointer

ncounter:
Closing --> https://github.com/openSUSE/open-build-service/pull/15931#discussion_r1551400258

PR #15933, opened 2024-04-04
Include revision in the file url

### LINKED ISSUE #15857: Missing rev parameter for URL display in web interface
# Issue Description
When viewing and old revision the link to file display is missing the revision parameter.

Expected Result
============
Link should work

How to Reproduce
===============
1. Go to 'https://build.opensuse.org/package/show/Application:Geo/postgis?rev=96'
2. Click on 'patch-tests-results.patch' or the download button
3. See error

Further Information
===============
* The resulting link is https://build.opensuse.org/projects/Application:Geo/packages/postgis/files/patch-tests-results.patch?expand=1, but the correct address is https://build.opensuse.org/projects/Application:Geo/packages/postgis/files/patch-tests-results.patch?expand=1&rev=96

For old revisions all links also must have the revision in the URL.

P.S. For display of old revisions the "Delete file" action makes no sense and should be removed. The link contains the rev parameter but it seems it tries to delete the file in the current revision nonetheless, which is plan wrong as well.

hellcp-work:
We were missing `srcmd5` from the views in case we are displaying a revision. We need it to appear in links in order to link to the right file version. This also removes the delete file link from older revisions.

Fixes #15857

### DIFF
--- a/src/api/app/controllers/webui/package_controller.rb
+++ b/src/api/app/controllers/webui/package_controller.rb
@@ -466,7 +466,9 @@ class Webui::PackageController < Webui::WebuiController
       @revision = @current_rev if !@revision && !@srcmd5 # on very first page load only
 
       files_xml = @package.source_file(nil, { rev: @srcmd5 || @revision, expand: @expand }.compact)
-      @files = Xmlhash.parse(files_xml).elements('entry')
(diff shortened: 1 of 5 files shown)


hennevogel (changes requested):
You should spec this out...

hennevogel (approved):

PR #15934, opened 2024-04-04
Don't display the login page for logged in users

### LINKED ISSUE #6365: Logging in via [BS instance]/session/new is weird
# Issue Description
I tried logging in via the login form at /session/new. After logging in, I still saw the same login form. That is confusing.

Expected Result
============
I'd expect to not see a log in form after logging in. I'd expect to be taken back to the SR (or other page) that I came from, and if that is not possible, I should be forwarded to /user/show/[myname]. 

How to Reproduce
===============
1. Make sure you're logged out of OBS
2. Go to a random SR, e.g.: https://build.opensuse.org/request/show/178397
3. Scroll down to the "login" link in the comments section and click that
4. You are now on a login form, login through that.

Result:
* The page header shows that you are logged in
* You are still on the same login form that asks you to "Please Log In"

Further Information
===============

hellcp-work:
Avoids displaying the login page if you are already logged in. Meant to fix #6365

### DIFF
--- a/src/api/app/controllers/webui/session_controller.rb
+++ b/src/api/app/controllers/webui/session_controller.rb
@@ -1,5 +1,6 @@
 class Webui::SessionController < Webui::WebuiController
   before_action :kerberos_auth, only: [:new]
+  before_action :already_logged_in, only: [:new]
 
   before_action :session_creator, only: [:create]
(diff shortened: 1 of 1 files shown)


saraycp:
@hellcp-work you have a few feature specs failing...

PR #15937, opened 2024-04-04
Multiple workflowrun filters - UI

ncounter:
Move from filtering `WorkflowRun` results by a single link to multiple filter options selection.

# Test:
Visit https://obs-reviewlab.opensuse.org/ncounter-multiple-workflowrun-filters/my/tokens/1/workflow_runs and play with filters

# Note - DO NOT MERGE :warning: 
The current PR is changing only the UI part, so the filter mechanism **is broken** at the moment and the implementation is happening in https://github.com/openSUSE/open-build-service/pull/15950

# Before
image

# After
image


src/api/app/components/workflow_run_filter_checkbox_component.html.haml
```diff
@@ -0,0 +1,9 @@
+- sanitized_key = @text.parameterize.underscore
```
danidoni, line 1: This should go into the class

ncounter: Done

src/api/app/components/workflow_run_filter_component.html.haml
```diff
@@ -8,40 +8,52 @@
...
-                                              filter_item: { generic_event_type: 'tag_push' }, selected_filter: @selected_filter)
-.list-group.list-group-flush.mt-5.mb-2
-  %h5.mx-3
-    Event Source
-    = render partial: 'webui/shared/info_popover', locals: { position: 'bottom', text: event_source_info_text }
-  = render WorkflowRunFilterInputComponent.new(token_id: @token.id, text: 'PR/MR', selected_input_filter: @selected_filter[:pr_number],
-                                               placeholder: '12345')
-  = render WorkflowRunFilterInputComponent.new(token_id: @token.id, text: 'Commit', selected_input_filter: @selected_filter[:commit_sha],
-                                               placeholder: '97561db8664eaf86a1e4c7b77d5fb5d5bff6681e')
+-# TODO: fix the form parameters
```
danidoni, line 19: What do we need to fix?

src/api/app/components/workflow_run_filter_component.html.haml
```diff
@@ -8,40 +8,52 @@
...
+= form_for(@token, url: token_workflow_runs_path(@token, @selected_filter), method: :get) do |form|
```
danidoni, line 20: In my tests, the `@selected_filter` was not needed because the inputs were sent via query strings, and those selected filters were just adding noise

ncounter: True. It's no longer needed now. Dropped
(It will make much more sense for the whole picture when frontend and backend changes will be put together)

danidoni (approved):

ncounter:
Promoting https://github.com/openSUSE/open-build-service/pull/15950 as the **main** PR to include this one too. Closing this.

Issue #11300, opened 2021-06-28
Erroneous return of HTTP 200 code when accessing login-restricted items


jengelh:
```
$ curl -D- https://build.opensuse.org/package/binary/GNOME:Next/at-spi2-core/openSUSE_Factory/x86_64/typelib-1_0-Atspi-2_0-2.40.2-360.3.x86_64.rpm
HTTP/2 200 
status: 200 OK
```

When opening this URL with a web browser, it gives a red box `Please login to access the requested page.` The HTTP request should therefore not respond with 200, because that erroneously suggests to automated tools that there was success, when it fact, there was not. It should return 403 as a HTTP status code (and the same HTML content with the main page/red box as it currently does.)

eduardoj:
This happens on purpose for this and other Web-UI URLs. The web client is detected as a bot and the response returned is a 200 HTTP code before performing any user login authentication. Closing.

PR #15944, opened 2024-04-08
Update list of crawler user agents

eduardoj:
I noticed this was missing after working on: #11300.

Related to: #10648.


saraycp:
For the record: this was updated using `rake voight_kampff:import_user_agents`.

PR #15949, opened 2024-04-09
Restore constraints to the Webui::Packages::BinariesController#destroy route

hellcp-work:
This causes certain paths to report 404 instead of triggering the route


hennevogel (approved):

rhabacker:
This problem was reported at https://bugzilla.suse.com/show_bug.cgi?id=1221600

PR #15950, opened 2024-04-09
Multiple workflowrun filters

danidoni:
Move from filtering `WorkflowRun` results from a single link to multiple filter options selection.

# Test:
Visit  http://obs-reviewlab.opensuse.org/danidoni-backend-support-for-multiple-workflowrun-filters/my/tokens/1/workflow_runs and play with filters

# Before
image

# After
image


src/api/app/queries/workflow_runs_finder.rb
```diff
@@ -6,42 +6,83 @@ class WorkflowRunsFinder
...
+                                  request_action)
+    end
+
+    return self if other_event_types.nil?
+
+    other_event_types = [other_event_types] unless other_event_types.is_a?(Array)
+
+    # The rest of the event types need to be OR'ed against the first event type
+    @relation = other_event_types.inject(@relation) do |local_relation, other_event_type|
+      query = case other_event_type
```
hennevogel, line 60: By now there is `WorkflowRun.generic_event_type`

src/api/app/queries/workflow_runs_finder.rb
```diff
@@ -6,42 +6,83 @@ class WorkflowRunsFinder
...
             end
 
-    workflow_runs = @relation.where(query)
-    if request_action && generic_event_type == 'pull_request'
-      workflow_runs = workflow_runs.where("JSON_EXTRACT(request_payload, '$.action') = (?) OR JSON_EXTRACT(request_payload, '$.object_attributes.action') = (?)", request_action,
-                                          request_action)
+    @relation = @relation.where(query)
+
+    if request_action.any? && first_event_type == 'pull_request'
+      @relation = @relation.where("JSON_EXTRACT(request_payload, '$.action') = (?) OR JSON_EXTRACT(request_payload, '$.object_attributes.action') = (?)", request_action,
```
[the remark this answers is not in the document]
danidoni: Fixed

ncounter (approved):

src/api/.rubocop_todo.yml
```diff
@@ -373,6 +373,7 @@ Metrics/CyclomaticComplexity:
     - 'app/models/update_notification_events.rb'
     - 'app/models/user.rb'
     - 'app/models/user_ldap_strategy.rb'
+    - 'app/models/workflow_run_filter.rb'
```
hennevogel, line 376: Hm, ignoring that I don't really like... Why is that so?

src/api/.rubocop_todo.yml
```diff
@@ -696,6 +697,7 @@ RSpec/ExampleLength:
     - 'spec/controllers/webui/users/notifications_controller_spec.rb'
     - 'spec/db/data/backfill_notified_projects_spec.rb'
     - 'spec/db/data/regenerate_notifications_spec.rb'
+    - 'spec/features/beta/webui/workflow_runs_spec.rb'
```
hennevogel, line 700: How this still in "beta"?

src/api/.rubocop_todo.yml
danidoni: Because that class is not that complex to understand as rubocop says

ncounter: I agree, it's a simple list of single conditions for the parameters.

src/api/.rubocop_todo.yml
danidoni: It is not, let me change it

src/api/.rubocop_todo.yml
hennevogel: Then please create follow up cards. We are supposed to have a plan how to pay back tech-debt we take on.

src/api/app/components/workflow_run_filter_checkbox_component.rb
```diff
@@ -0,0 +1,28 @@
...
+    @icon = icon
+  end
+
+  def icon_tag
+    tag.i(class: ['me-1', @icon]) if @icon != ''
+  end
+
+  private
+
+  def workflow_run_filter_matches?
```
hennevogel, line 19: Hm, why is this a component?

src/api/.rubocop_todo.yml
danidoni: Done

src/api/app/components/workflow_run_filter_checkbox_component.rb
danidoni: Instead of a partial?

ncounter: Definitely a partial. So the other filters actually.

ncounter: They were components, they still are. Refactoring to be tackled later. A card exists already for that

src/api/.rubocop_todo.yml
ncounter: https://trello.com/c/W2QqLX6H

src/api/app/models/workflow_run_filter.rb
```diff
@@ -0,0 +1,28 @@
+class WorkflowRunFilter
```
hennevogel, line 1: Phew.... A PORO to rename params instead of renaming the params?

ncounter: https://github.com/openSUSE/open-build-service/pull/15979/commits/e21d15c73456017604f3c759f7ba7615e4ebacf4 one step toward dismissing it

src/api/app/queries/workflow_runs_finder.rb
```diff
@@ -6,42 +6,22 @@ class WorkflowRunsFinder
...
 
   def initialize(relation = WorkflowRun.all)
+    @initial_relation = relation
     @relation = relation.includes([:token]).order(created_at: :desc)
   end
 
-  def all
-    @relation.all
-  end
+  def reset
```
hennevogel, line 13: You would not need to do this if this would not be in a scope/where wrapper class...

src/api/spec/features/webui/workflow_runs_filter_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'browser_helper'
+
+RSpec.describe 'Workflow runs filters' do
```
hennevogel, line 3: You should test the view components and filter class directly instead of firing up a browser to do this...

hennevogel: Don't get me wrong, I have nothing against testing the happy path of this feature but how it behaves should be tested as close to where the code is...

src/api/app/models/workflow_run_filter.rb
danidoni: Renaming the params won't change anything... This object is adapting the way we use the params: our filter form sets parameters like `succeeded=1&running=1` and the way we consume the filters expects something like `status=['success','fail']`. We need some code to translate it.

src/api/spec/features/webui/workflow_runs_filter_spec.rb
ncounter: Maybe testing simple components here is a bit of an effort that vanishes soon, as discussed in https://github.com/openSUSE/open-build-service/pull/15950#discussion_r1562364224 . No?

src/api/app/queries/workflow_runs_finder.rb
danidoni: What is a where wrapper class?

hennevogel: `WorkflowRunsFinder` wraps a couple of `WorkflowRun.where...` calls

src/api/spec/features/webui/workflow_runs_filter_spec.rb
hennevogel: I did not mean "test every component". I mean test that this feature behaves correctly as close to the code as you can. That is usually *not* the browser. My guess is this is currently: `WorkflowRunFinder`, `WorkflowRunController` and `WorkflowRunFilterComponent`.

src/api/app/models/workflow_run_filter.rb
hennevogel: > We need some code to translate it.

Okay but for sure not a PORO you instantiate...

danidoni: I got rid of this class

src/api/app/controllers/webui/workflow_runs_controller.rb
```diff
@@ -1,23 +1,20 @@
...
-                     elsif params[:generic_event_type]
-                       @workflow_runs_finder.with_generic_event_type(params[:generic_event_type], @request_action)
-                     else
-                       @workflow_runs_finder.all
-                     end
-
-    @workflow_runs = @workflow_runs.page(params[:page])
-
-    @selected_filter = selected_filter
+    # TODO: The pull/merge request dropdown should accept multiple selections
```
danidoni, line 3: We are tracking this [here](https://trello.com/c/DxI7jUxT)

src/api/app/components/workflow_run_filter_component.rb
```diff
@@ -1,17 +1,16 @@
 class WorkflowRunFilterComponent < ApplicationComponent
-  def initialize(token:, selected_filter:, finder:)
+  def initialize(token:, selected_filter:)
     super
 
-    @count = workflow_runs_count(finder)
+    @count = {
+      'success' => WorkflowRun.success.count,
```
hennevogel, line 6: This would be all of the `WorkflowRun` in the database no?

src/api/spec/components/workflow_run_filter_component_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(rendered_content).to(have_no_checked_field('fail'))
+      expect(rendered_content).to(have_no_checked_field('Pull/Merge Request'))
+      expect(rendered_content).to(have_no_checked_field('push'))
+      expect(rendered_content).to(have_no_checked_field('tag_push'))
+    end
+    # rubocop:enable RSpec/MultipleExpectations
+    # rubocop:enable RSpec/ExampleLength
+  end
+
+  context 'when having multiple filters selected' do
```
hennevogel, line 25: I guess the point here is not no vs. multiple but that those filters work right? I would prefer individual examples...

src/api/spec/controllers/webui/workflow_runs_controller_spec.rb
```diff
@@ -18,17 +18,51 @@
 
     context 'when action_filter is available' do
       it 'finds workflow runs when the action is available' do
-        get :index, params: { token_id: workflow_token.id, request_action: 'opened', generic_event_type: 'pull_request' }
+        get :index, params: { token_id: workflow_token.id, request_action: 'opened', pull_merge_request: 1 }
 
         expect(assigns(:workflow_runs).count).to eq(1)
```
hennevogel, line 23: What's the difference to the context above? A count of `1` sound like a magic number to me... How about expecting the data your really expect in both?

src/api/spec/controllers/webui/workflow_runs_controller_spec.rb
```diff
@@ -18,17 +18,51 @@
...
-        get :index, params: { token_id: workflow_token.id, request_action: 'closed', generic_event_type: 'pull_request' }
+        # TODO: Check this pull_merge_requet parameter...
+        get :index, params: { token_id: workflow_token.id, request_action: 'closed', pull_merge_request: 1 }
 
         expect(assigns(:workflow_runs).count).to eq(0)
-        expect(assigns(:request_action)).to eq('closed')
+      end
+    end
+
+    context 'multiple filters' do
```
hennevogel, line 34: Again I don't think multiple vs. none is the point here. I would test one of the `status_params` filters and one of the `event_type_params` maybe?

hennevogel (changes requested):

src/api/app/components/workflow_run_filter_component.rb
danidoni: No, this would be only the 'success' ones, the failing and the running are not included here

hennevogel: Yeah but the whole UI is about the ones of the token, not all of the ones in the database no?

krauselukas: should be addressed now by https://github.com/openSUSE/open-build-service/pull/15950/commits/bcbf780c63ce276dc8d53b8b47057de760d14f82

src/api/spec/controllers/webui/workflow_runs_controller_spec.rb
danidoni: Fixed

src/api/spec/controllers/webui/workflow_runs_controller_spec.rb
danidoni: Done

src/api/spec/controllers/webui/workflow_runs_controller_spec.rb
```diff
@@ -13,22 +13,74 @@
...
+
+      before do
+        login pull_request_1_workflow_run.token.executor
+      end
+
+      # rubocop:disable RSpec/ExampleLength
+      it 'shows all the workflow runs' do
+        get :index, params: { token_id: token.id, pr_mr: '', commit_sha: '' }
+
+        expect(assigns[:workflow_runs]).to contain_exactly(pull_request_1_workflow_run,
```
hennevogel, line 50: Maybe create one more WorkflowRun that doesn't match?

src/api/spec/controllers/webui/workflow_runs_controller_spec.rb
```diff
@@ -13,22 +13,74 @@
...
+    end
+
+    context 'when setting a status filter of success' do
+      let!(:succeeded_workflow_run) { create(:workflow_run, :succeeded) }
+
+      before do
+        login succeeded_workflow_run.token.executor
+      end
+
+      it 'shows all the succeeded workflow runs' do
```
hennevogel, line 66: Here too. Create also a WorkflowRun that wouldn't match the filter.

src/api/spec/controllers/webui/workflow_runs_controller_spec.rb
danidoni: There is already a workflow run at the top of the spec that also applies here but does not match because it gets filtered out by the token.

hennevogel: A bit implicit but okay good.

hennevogel (approved):

PR #15955, opened 2024-04-10
Enhance the visual difference from placeholder to real input value

ncounter:
https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder#color_contrast
https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder#change_placeholder_appearance

## Before
before-placeholder-2
before-placeholder
before-placeholder-3
before-placeholder-4

## After
after-placeholder-2
after-placeholder
after-placeholder-3
after-placeholder-4

### DIFF
--- a/src/api/app/assets/stylesheets/webui/colors.scss
+++ b/src/api/app/assets/stylesheets/webui/colors.scss
@@ -8,9 +8,11 @@ $themed-dark-nav-bg: #121212;
 :root {
   --navbar-themed-bg: #{$gray-800};
   --breadcrumbs-item-text-color: #{$gray-700};
+  --placeholder-text: #{$gray-600};
 }
 
 @include color-mode(dark) {
   --navbar-themed-bg: #{$themed-dark-nav-bg};
   --breadcrumbs-item-text-color: #{$gray-400};
+  --placeholder-text: #{$gray-500};
 }
--- a/src/api/app/assets/stylesheets/webui/mixin.scss
+++ b/src/api/app/assets/stylesheets/webui/mixin.scss
@@ -5,3 +5,8 @@
 .text-bg-hover {
   box-shadow: 0px 0px 5px $hover;
 }
+
+::placeholder {
+  color: var(--placeholder-text)!important;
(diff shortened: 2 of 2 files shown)


ncounter:
> > https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder#color_contrast
> 
> Thanks for providing the link! It states the following:
> 
> > Color contrast ratio is determined by comparing the luminosity of the placeholder text and the input background color values. In order to meet current [Web Content Accessibility Guidelines (WCAG)](https://www.w3.org/WAI/standards-guidelines/wcag/), a ratio of 4.5:1 is required for text content and 3:1 for larger text such as headings. Large text is defined as 18.66px and bold or larger, or 24px or larger.
> 
> `$gray-500` we [currently have defined](https://github.com/openSUSE/open-build-service/blob/b9d1c5cbb0cf92942f1e8db068ed4b1b3a59e8e8/src/api/app/assets/stylesheets/webui/bootstrap_variables/colors.scss#L6) as `#878C92`, which with pure white background (the best case scenario) has contrast of 3.38:1.

:rofl: my bad typo, I meant to use `$gray-600` which means `6c757d`, it provides 4.68:1 contrast which should be good. I'm fixing with it, would you agree?

ncounter:
> > 🤣 my bad typo, I meant to use `$gray-600` which means `6c757d`, it provides 4.68:1 contrast which should be good. I'm fixing with it, would you agree?
> 
> I have bad news once again, our body background in dark mode is `#212529` which with `$gray-600` will provide us with 3.28:1 contrast.

Yeah, the rule needs to be adapted for the dark mode.

ncounter:
@hellcp-work 
- light mode uses a `#FFFFFF` background, with a `$gray-600` (`#6c757d`) text color the contrast is `4.68:1`
- dark mode uses a `#212529` background, with a `$gray-500` (`#878c92`) text color the contrast is `4.55:1`
I also updated the screenshots. Feel free to review it once again. Thank you

